PR #955, opened 2026-07-22
fix(runtime): isolate synchronous database work from async task execution

AlexLiu190625:
## Summary

This change prevents synchronous database work from blocking the asyncio event loop when the SQLAlchemy connection pool is under pressure.

It introduces a shared cancellation-safe database boundary, moves task-runtime persistence into short-lived worker-owned sessions, and returns detached snapshots instead of carrying SQLAlchemy sessions or ORM objects across async/thread boundaries.

The same invariant is applied to task execution, A2A, `/v1` task APIs, authentication, personal-key agent management, and upload workflows.

Fixes #935.

## Confirmed incident behavior addressed

During the SG incident, the configured pool reached its limit:

```text
pool_size=10
max_overflow=20
pool_timeout=10s
```

The confirmed code amplification was:

1. task execution retained an outer session and then requested another connection while resolving execution scope;
2. after that checkout timed out, failure handling immediately requested another connection before releasing the outer session;
3. lease heartbeat, tracker, and cleanup paths performed synchronous database work from async code;
4. each synchronous pool wait blocked the main event loop, affecting unrelated requests such as login and health checks.

This change addresses those nested/repeated checkouts and event-loop stalls. It does not claim that execution-scope resolution initiated the original pool exhaustion or that every overlapping task continuously held a connection.

### Shared async database boundary

- Add cancellation-safe helpers for running synchronous database operations in worker threads.
- Require worker operations to create, use, and close their own sessions.
- Return immutable or detached snapshots instead of ORM objects.
- Drain in-flight worker operations before propagating cancellation so late commits cannot race cleanup.
- Classify SQLAlchemy pool timeouts through the full exception chain.

### Task runtime and lifecycle

- Load task setup, execution scope, runtime identity, model configuration, reconstruction state, and execution recovery state through worker-owned snapshot boundaries.
- Reuse the same setup snapshot for normal construction and reconstruction instead of probing history and reloading runtime configuration through separate sessions.
- Resolve execution scope before the main execution session is opened and pass it into the runtime.
- Move task claim, lease acquisition, heartbeat, failure persistence, final settlement, pause/resume, command handling, and tracker writes off the event loop.
- Refactor `TaskTracker` to retain primitive identifiers and counters rather than a caller-owned session and ORM task.
- Avoid an immediate second checkout after a pool-timeout failure; retain the fenced lease for TTL recovery when capacity is still unavailable.
[...]

Issue #935, opened 2026-07-20
bug: Task runtime blocks the asyncio event loop during database pool exhaustion


AlexLiu190625:
## Summary

During the SG production incident on 2026-07-20, the SQLAlchemy connection pool reached its configured limit. Task execution then performed synchronous connection-pool waits from async code paths, blocking the backend event loop in repeated 10-second intervals.

The initiating pool exhaustion and the code-path amplification are separate facts:

- production logs confirm that the pool was exhausted;
- current code can synchronously request a second connection while an execution session is already open;
- after that checkout times out, the exception handler immediately performs another synchronous checkout before the outer session is released;
- lease-heartbeat, tracker, and some orchestrator cleanup paths can also wait for the pool directly on the event-loop thread.

This coincided with unrelated API failures, including the login health check.

## Environment

- Region: SG production
- Incident date: 2026-07-20
- Backend process: one `python -m xagent_saas` process
- Deployed `xagent-saas` commit: `4f2093d2dde20e7ec841e707879b475749e23ca7`
- Deployed `xagent` submodule: `3a1b69f5640f0fbd99f09c9b68f48d79e32c827b`
- Current `xagent/main`: `217d915f6052b9b14b41b8462229321386654382`
- Current downstream `xagent-saas/main`: `e645ba9287dc65491d3e8129cf64edaedf621d4a`
- Runtime pool configuration:

```text
XAGENT_DB_POOL_SIZE=10
XAGENT_DB_MAX_OVERFLOW=20
XAGENT_DB_POOL_TIMEOUT_SECONDS=10
```

The configured maximum was therefore 30 pooled plus overflow connections, with a 10-second checkout timeout.

The relevant websocket, execution-scope, task-orchestrator, lease-heartbeat, and task-tracker files are unchanged between deployed `xagent` revision `3a1b69f` and current `main`. The related `chat.py` changes on current main are token-accounting changes and do not alter this execution path.

## Production evidence

The backend logged:

```text
QueuePool limit of size 10 overflow 20 reached, connection timed out, timeout 10.00
```

Observed facts from the affected window:

1. There were 43 overlapping task lifetimes. This exceeds the configured maximum of 30 and coincides with confirmed pool exhaustion, but it does **not** prove that all 43 tasks held database connections continuously.
2. Task `73983` entered execution at `06:20:49 UTC` and its failure path ended at approximately `06:21:09 UTC`.
[...]

PR #964, opened 2026-07-22
refactor(tools): prefetch runtime database state with worker-owned sessions

AlexLiu190625:
## Summary

This PR introduces the worker-owned database prefetch foundation used by the task-runtime isolation work. Synchronous database access is executed outside the asyncio event loop, each worker owns and closes its short-lived SQLAlchemy session, and async consumers receive detached data instead of ORM objects.

## Changes

- Add a cancellation-safe boundary for synchronous database work.
- Prefetch detached ToolFactory inputs, including tool policy, credentials, SQL connections, model selections, custom API configuration, and published-agent records.
- Run snapshot and policy inputs through one checked-out-session ownership boundary while preserving their distinct fallback policies.
- Make creator-scoped fail-closed inputs explicit and give each audio input a distinct diagnostic label.
- Close owned embedding and rerank default-model sessions without closing caller-owned sessions.
- Build published-agent tools from detached records.
- Release clean request-session read transactions before worker pool checkout.
- Keep per-turn policy refresh and factory preparation under one shared ownership boundary.

## Compatibility

- Existing `BaseToolConfig` implementations without optional prepare/release hooks retain the existing construction path.
- Existing `WebToolConfig` callers may continue providing a live session; factory-backed callers can create short-lived sessions.
- No database migration or public API/schema change is included.

## Scope boundary

The standalone `model_service.get_default_model`, `get_fast_model`, and `get_compact_model` helpers remain unchanged. Repository-wide call-site search found no in-repository production caller for these helpers; live task LLM resolution is owned by `llm_utils` and the separate core task-runtime work rather than the ToolFactory snapshot.

This split does not include the remaining #935 task-runtime changes: task setup snapshots, agent creation and reconstruction isolation, task lease and status persistence, tracker writes, or `/v1` authentication SQL and bcrypt isolation. Those changes remain in the separate core runtime split.

## Verification

- The six affected test modules passed locally: 174 tests.
- `uv run pre-commit run --files ...`: passed for the four changed Python files, including Ruff, formatting, mypy, whitespace, import sorting, and codespell checks.
- `git diff --check`: passed.
- GitHub Actions run `29979345388`: all required pre-commit, frontend, migration, fast, slow, integration, Deepdoc, and end-to-end checks passed.

Related to #935.

Split from #955.


src/xagent/web/services/db_runtime.py
```diff
@@ -0,0 +1,79 @@
...
+            result = await asyncio.shield(task)
+        except asyncio.CancelledError as exc:
+            current = asyncio.current_task()
+            caller_is_cancelling = current is not None and current.cancelling() > 0
+            if not caller_is_cancelling:
+                raise
+            if cancellation is None:
+                cancellation = exc
+            if not task.done():
+                continue
```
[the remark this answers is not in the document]
AlexLiu190625: The busy-loop conclusion does not reproduce on the supported runtimes. With the actual helper on Python 3.12.13, one cancellation produced two shield calls total; that count remained unchanged while the child stayed blocked for 50 ms. A second later cancel produced exactly one additional caught cancellation and one additional shield call. Equivalent cancellation control flow on Python 3.11.15 behaved the same way. Task.cancelling() retains the request count, but it does not automatically inject CancelledError at every subsequent await. The focused database-runtime suite also passes all six tests, including the case that must resume the event loop to release the blocked worker. uncancel() is intended when cancellation is suppressed so a task can continue indefinitely; this helper instead drains the owned worker and then re-raises the original cancellation. Decrementing the cancellation count would therefore change the intended accounting. No code change is needed.

src/xagent/core/tools/adapters/vibe/factory.py
```diff
@@ -282,6 +282,26 @@ class ToolFactory:
...
+        if callable(prepare_factory_runtime):
+            await prepare_factory_runtime(config)
+        try:
+            return await ToolFactory._create_all_tools_prepared(
+                config,
+                apply_user_override_filter=apply_user_override_filter,
+            )
+        finally:
+            if callable(release_factory_runtime):
+                release_factory_runtime(config)
```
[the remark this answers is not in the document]
AlexLiu190625: The class-level lookup is deliberate for this optional lifecycle capability. Repository-wide search found no instance-only lifecycle overrides; the production implementation defines both hooks on `WebToolConfig`. Several existing factory tests use unspecced `MagicMock` configs, where instance-level `getattr` synthesizes a callable `MagicMock` for the absent hook and `await hook()` raises `TypeError`. Subclass overrides remain supported by the class lookup. Switching to instance lookup would regress existing config shapes, so no code change is appropriate.

src/xagent/web/services/db_runtime.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    cancellation: asyncio.CancelledError | None = None
+    while True:
+        try:
+            result = await asyncio.shield(task)
+        except asyncio.CancelledError as exc:
+            current = asyncio.current_task()
+            caller_is_cancelling = current is not None and current.cancelling() > 0
+            if not caller_is_cancelling:
+                raise
```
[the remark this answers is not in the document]
AlexLiu190625: Python 3.10 is outside this repository supported runtime contract. `pyproject.toml` declares `requires-python = ">=3.11"`, mypy targets 3.11, the backend image defaults to 3.11, and CI uses 3.11/3.12. `Task.cancelling()` is present in both supported versions, verified on 3.11.15 and 3.12.13. The proposed child-state fallback would also weaken the explicit distinction between caller cancellation and child cancellation. No compatibility fallback is needed.

rogercloud (commented):
## Summary

This PR extracts the synchronous DB-session prefetch for `ToolFactory` inputs (tool policy, credentials, SQL connections, model selections, custom API config, and published-agent records) off the event loop and into short-lived worker threads, returning **detached snapshots** to the async consumers. The mechanism is new (`db_runtime.py` async/cancellation boundary; a plan/snapshot/prepare/release lifecycle in `WebToolConfig`; a session-owning refactor of `model_service`; and published-agent tool building from detached ORM records in the vibe factory). It directly targets issue #935 (SG production incident where sync DB calls on the loop exhausted the SQLAlchemy pool and caused ~10s stalls that took down unrelated requests such as login). It is split from umbrella PR #955, with task-setup snapshots, agent-creation isolation, task lease/tracker writes, and /v1 auth SQL/bcrypt isolation explicitly deferred to the core-runtime split.

The worker-owned-session + detached-snapshot pattern is a coherent and correct answer to "get sync DB off the event loop," and the drain-before-cancel semantics in `db_runtime.py` are the right call for resource safety (they prevent orphaned worker threads from writing into a session that's already being torn down, and prevent ORM objects from leaking across the thread boundary). The design is heavier than the incident strictly requires, but that heaviness buys real safety, and it is **not** over-engineered for the problem.

Two reservations keep this from an unqualified approval:

- **Incomplete delivery within the chosen scope.** Several DB-touching model getters in the very file this PR refactors were left on the old on-loop `next(get_db())` path (see F2). The PR doesn't fully deliver its own stated goal even within `model_service.py`.
- **D1 — Shared default-executor blast radius.** `run_db_io_cancellation_safe` offloads via bare `asyncio.to_thread(operation)`, i.e. `loop.run_in_executor(None, ...)` — the process-wide default `ThreadPoolExecutor` (unsized default `min(32, os.cpu_count()+4)`; no `set_default_executor` override exists in the repo). That same default executor is heavily used for unrelated blocking work: websocket event persistence (10+ sites), task status/CRUD, KB ingestion (~40 sites), Docker sandbox calls, and task orchestration. Under a #935-style incident, many concurrently-blocked DB-prefetch workers — each waiting up to the pool timeout (~10s) — would occupy slots in that same small shared pool and could starve those unrelated operations. This PR correctly gets the DB waits off the event loop, but does not contain the blast radius to a dedicated executor. **Not a blocker for this PR**, but worth a forward-looking follow-up: route DB-prefetch offloads through a dedicated bounded executor.

## Findings

### F1 — MAJOR — Shared worker session has no `rollback()`, so one read's failure cascades into unrelated reads on Postgres
`src/xagent/web/tools/config.py`, `_load_tool_factory_runtime_snapshot` (~L687–L863)

The function opens **one** worker-thread `Session` (`db = session_factory()` at L687, closed in `finally` at ~L855) and reuses it across every prefetch read: policy, credentials (~L694), SQL connections (~L713), custom API (~L724), published-agent records (~L742), and all `model_service.get_*` calls (~L756–L843). Each read is independently `try/except`-wrapped and falls back to a default on failure, but there is **no `db.rollback()` anywhere in the function** (confirmed by grep — the only `rollback` occurrences in `config.py` belong to an unrelated `oauth_db` object at L2591/L2605; `db_runtime.py` has none).

On PostgreSQL (the incident database), an unhandled SQL error inside one read leaves the shared transaction in the *aborted* state; every subsequent read sharing that same session/transaction then also fails with `current transaction is aborted, commands ignored until end of transaction block` and silently degrades to its own default — even though it had nothing to do with the original failure. This is a genuine regression: pre-PR, each model getter opened its **own** `next(get_db())` session, so one getter's DB error could not poison another's read. The test suite runs on SQLite only, which has no equivalent abort-until-rollback mode, so this cascade is entirely untested and would surface only on Postgres — precisely the #935 environment.

**Suggested fix:** call `db.rollback()` in each `except` block (or wrap each read in a nested transaction / `SAVEPOINT` via `db.begin_nested()`) before falling through to the default, so one input's failure cannot cascade into unrelated inputs' reads.

### F2 — MAJOR / MINOR (mixed) — Several live model getters were not migrated off the on-loop DB path
`src/xagent/web/services/model_service.py`: `get_default_model` (~L209), `get_fast_model` (~L272), `get_compact_model` (~L335), `get_default_embedding_model` (~L961), `get_default_rerank_model` (~L1013)

The PR migrated `get_default_vision_model`, `get_default_image_generate_model`, `get_default_image_edit_model`, `get_default_video_model`, `get_default_asr_model`, `get_default_tts_model`, `get_default_sound_effect_model`, and `get_default_music_model` to accept an optional `db: Session | None = None` and participate in the worker-thread/snapshot pattern. But the five getters above were **not** migrated — no diff hunk touches them, they still call `next(get_db())` directly, and `_ToolFactoryRuntimeSnapshot` (config.py:249–273) has no fields for default/fast/compact/embedding/rerank models.

- **MAJOR half:** `get_default_model`, `get_fast_model`, `get_compact_model` are called directly by chat/agent code paths on the event loop, un-gated by any snapshot — i.e. still doing exactly the synchronous on-loop DB access this PR exists to eliminate, for a different subset of models.
- **MINOR half:** the embedding/rerank getters are effectively dead on the tool-build path today — no production caller of `WebToolConfig.get_embedding_model()`/`get_rerank_model()` was found, and `test_factory_does_not_inject_global_default_embedding_model` asserts they aren't consulted during tool building. Low real-world impact today, but still an untouched on-loop pattern.

**Suggested fix:** either migrate `get_default_model`/`get_fast_model`/`get_compact_model` now (they're in the same file and pattern as everything else here), or explicitly document them in an "Out of scope" note the way task-runtime/lease/auth work is already called out — so the remaining on-loop DB access is an acknowledged decision, not an oversight.

### F3 — MINOR — Three-way asymmetry in per-input failure handling
`src/xagent/web/tools/config.py`, `_load_tool_factory_runtime_snapshot` (~L646–L764)

Failures are handled three different ways in the same function:
- The `load_optional_input` helper (L665–L680, used for model-selection reads at L769–L860) catches exceptions, re-raises only pool timeouts, logs, and returns a default — never touching `failed_inputs`.
- The `basic` credentials (~L706) and `database`/SQL (~L718) blocks hand-roll their own try/except **and** add to `failed_inputs`; downstream, `get_tool_credential` (~L1882) and `get_sql_connections` (~L1890) check `failed_inputs` and raise `RuntimeError`.
- But `custom_api` (~L724–L741) and `published_agent` (~L743–L764) also hand-roll try/except yet **never** touch `failed_inputs` — so their failures silently degrade to defaults with no way for a caller to detect the degradation, while basic/database failures correctly surface as hard errors.

This reads as an unintentional inconsistency rather than a deliberate "required vs optional inputs" split — custom API config and published-agent records seem at least as consequential as SQL connections. Worth either aligning the behavior or adding a comment documenting why these two are intentionally soft-fail. Not a merge-blocker.

### F4 — MINOR (test gap) — No test for cross-read failure isolation in a single snapshot build
`tests/web/tools/test_web_tool_config_session_factory.py`

No test exercises F1's scenario: one prefetch read raising a plain (non-timeout) exception, with an assertion on what happens to a **different, subsequent** read in the same snapshot-build call. Existing tests cover only pool-timeout propagation (a different exception type) or all-success paths; the closest analogues (`test_mcp_config_scan_failure_raises_safe_typed_error`, `test_failed_mcp_config_refresh_never_reuses_stale_cache`) repeat the *same* failing loader, not a different downstream loader sharing the same session.

**Recommend:** add a test that forces one loader to raise a generic `Exception` and asserts a later loader in the same snapshot build still succeeds (this should pass once F1's `rollback()` fix lands; on Postgres semantics it would currently fail — worth documenting as a known gap while CI is SQLite-only).

### F5 — MINOR (test gap) — Un-migrated getters have no session-lifecycle coverage
`tests/web/test_model_service.py`

No test imports or exercises `get_default_embedding_model`/`get_default_rerank_model` at all (they are not in the file's imports; the existing embedding tests target `DynamicMemoryStoreManager._get_embedding_model_from_db`, a different code path). The session-close test added for the migrated getters (`test_audio_generation_default_closes_owned_session`) has no equivalent for the un-migrated ones, so the pre-existing `next(get_db())`-without-close pattern on those getters remains untested.

## Prior review findings (gemini-code-assist) — re-verified, all WAIVED

All three earlier automated findings were independently re-verified against asyncio semantics, Mock behavior, and repo config. The author's rebuttals hold; no code change is needed:

1. **"Busy-loop from missing `uncancel()`" (`await_task_settlement`, HIGH)** — WAIVED. `Task.cancelling()` is a bookkeeping counter for `uncancel()` accounting; it isn't polled by the loop and doesn't make a subsequent `await asyncio.shield(task)` re-raise. CancelledError delivery is event-driven (one-shot per `cancel()`), not counter-driven, so there's no spin. Not calling `uncancel()` is correct here since the cancellation is being re-raised, not suppressed.
2. **"`type(config)` lookup breaks Mock/instance overrides" (`create_all_tools`, MEDIUM)** — WAIVED. `BaseToolConfig` defines no default for these hooks and no instance-level override exists in the codebase; the actual fixtures use `MagicMock(spec=BaseToolConfig)`, under which both class- and instance-level lookup return `None`. Class-level lookup is the safer choice for these fixtures. (See the simplification note below for a cleaner base-class-default alternative that is also compatible with this.)
3. **"Python 3.10 compat for `Task.cancelling()`" (MEDIUM)** — WAIVED. `pyproject.toml` sets `requires-python = ">=3.11"`; mypy, CI, and the Docker base image all target 3.11/3.12 only. `Task.cancelling()` was added in 3.11, so the compatibility floor makes the fallback unnecessary.

## Simplification opportunities (all confirmed)

- **`src/xagent/web/services/db_runtime.py:74` — delete:** `drain_async_task_cancellation_safe` has zero production callers in the repo — only its own dedicated test exercises it. Delete the function and its dedicated test (`tests/web/services/test_db_runtime.py:129`), or fold its logic into a test helper if the settlement-only coverage is still wanted.
- **`src/xagent/web/services/db_runtime.py:25` — yagni (optional):** `await_task_settlement` was split out to back two callers; with `drain_async_task_cancellation_safe` gone it has one real production caller (`run_db_io_cancellation_safe`). Could be folded back inline — though it currently has useful standalone unit coverage of the settlement/cancellation-recording logic in isolation, which is a legitimate reason to keep it separate.
- **`factory.py:287`, `service.py:98`, `service.py:647`, `service.py:661` — yagni:** the `getattr(obj_or_type, "hook_name", None) + callable()` probing for optional lifecycle hooks (`prepare_factory_runtime` / `release_prepared_factory_runtime` / `refresh_runtime_policy`) is duplicated across 4 sites. `BaseToolConfig` already sets a no-op-default precedent (`get_video_models`, `emit_mcp_load_summary`, `release_db_connection` are concrete no-op methods, not probed). Add no-op defaults for these three hooks directly on `BaseToolConfig` (each keeping its own sync/async signature) and call them directly (e.g. `await config.prepare_factory_runtime()`). This stays compatible with class-level lookup — a base-class default is found by class-level lookup too.
- **`src/xagent/web/tools/config.py` — shrink:** ~14 near-identical accessors (`get_image_models` ~L1192, `get_video_models` ~L1201, `get_asr_models` ~L1985, `get_tts_model` ~L2048, `get_music_model` ~L2114; spanning ~L1180–L2131) all follow the same shape: "if snapshot present and plan flag set → return snapshot field; else lazily load via legacy path and cache." A `_snapshot_or_legacy(plan_flag, snapshot_attr, cache_attr, loader)` helper would collapse most of these to one-liners. Exclude `get_tool_credential`/`get_sql_connections` (no caching, raise on `failed_inputs`) and `get_custom_api_configs` (inline query, no delegated loader) — they don't fit the uniform shape.

**net: roughly -15 to -25 lines possible** (one dead function + its test removed, ~4 probing sites collapsed to base-class defaults), plus the optional fold-back and the larger accessor consolidation.

## Testing

Independently re-ran the 6 affected modules:

```
pytest tests/core/agent/test_agent_service_tool_refresh.py \
       tests/core/tools/test_agent_tool_publish_status.py \
       tests/core/tools/test_vision_tool.py \
       tests/web/services/test_db_runtime.py \
       tests/web/test_model_service.py \
       tests/web/tools/test_web_tool_config_session_factory.py
```

**147 passed, 0 failed** — matches the PR's claim exactly.

## Recommendation

Approve-with-changes. F1 (Postgres transaction-abort cascade) is the one finding I'd want addressed before merge, since it is a real regression against the exact database and failure class from issue #935 and is invisible to the SQLite-only tests. F2 needs at minimum an explicit scope note. Everything else (F3–F5, D1, simplifications) can be follow-ups. Nice work engaging substantively with the earlier automated review — the three rebuttals all check out.

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,428 @@ async def refresh_oauth_token_if_needed(
...
+                exc_info=True,
+            )
+            return default
+
+    requires_direct_db = plan.requires_direct_db_without_policy or (
+        plan.load_policy and policy_snapshot is None
+    )
+    db = None
+    if requires_direct_db:
+        db = session_factory()
```
rogercloud: **MAJOR — shared worker session has no `rollback()`; one read's failure cascades into unrelated reads on Postgres.**

This function opens a single worker-thread `Session` here and reuses it across every prefetch read (policy, credentials ~L694, SQL ~L713, custom API ~L724, published agents ~L742, and all `model_service.get_*` calls ~L756–L843). Each read falls back to a default on failure, but there is no `db.rollback()` anywhere in the function.

On PostgreSQL (the issue #935 database), an unhandled SQL error in one read leaves the shared transaction *aborted*; every subsequent read on the same session then fails with `current transaction is aborted, commands ignored until end of transaction block` and silently degrades to its default — even though it was unrelated to the original failure. Pre-PR each getter opened its own `next(get_db())` session, so this cascade could not happen — it's a genuine regression. The SQLite-only test suite has no equivalent abort-until-rollback mode, so this is untested and would surface only on Postgres.

Suggested fix: call `db.rollback()` in each `except` block before falling through to the default, or isolate each read with `db.begin_nested()` (SAVEPOINT).

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,428 @@ async def refresh_oauth_token_if_needed(
...
+                except Exception as exc:
+                    if is_database_pool_timeout(exc):
+                        raise
+                    failed_inputs.add("database")
+                    logger.warning(
+                        "Failed to prefetch SQL connections",
+                        exc_info=True,
+                    )
+
+            if plan.load_custom_api:
```
rogercloud: **MINOR — inconsistent per-input failure handling.**

The `basic` (~L706) and `database` (~L718) blocks add to `failed_inputs` on failure, which downstream causes `get_tool_credential` (~L1882) / `get_sql_connections` (~L1890) to raise `RuntimeError` — a hard, detectable failure. But `custom_api` here (and `published_agent` ~L743) hand-roll try/except yet never touch `failed_inputs`, so their failures silently degrade to defaults with no way for a caller to detect the degradation. (`load_optional_input` for model reads is a third variant that also never touches `failed_inputs`.)

Custom API config and published-agent records seem at least as consequential as SQL connections, so this reads as an unintentional asymmetry rather than a deliberate required-vs-optional split. Either align the behavior or add a comment documenting why these are intentionally soft-fail.

tests/web/tools/test_web_tool_config_session_factory.py
```diff
@@ -109,6 +126,692 @@ def test_live_db_path_unchanged():
...
+    with pytest.raises(RuntimeError, match="registered tool build failed"):
+        await ToolFactory.create_all_tools(cfg)
+
+    assert len(sessions) == 1
+    assert sessions[0].closed
+    assert cfg._factory_runtime_snapshot is None
+
+
+@pytest.mark.asyncio
+async def test_factory_prepare_snapshots_selected_sync_factory_inputs(
```
rogercloud, line 594: **MINOR (test gap) — no cross-read failure-isolation test.**

No test covers the F1 scenario: one prefetch read raising a plain (non-timeout) exception, with an assertion on a *different, subsequent* read in the same snapshot build. Existing tests cover only pool-timeout propagation or all-success; the closest analogues (`test_mcp_config_scan_failure_raises_safe_typed_error`, `test_failed_mcp_config_refresh_never_reuses_stale_cache`) repeat the *same* failing loader.

Recommend adding a test that forces one loader to raise a generic `Exception` and asserts a later loader in the same snapshot build still succeeds. This should pass once the F1 `rollback()` fix lands; on Postgres semantics it would currently fail, so it also guards against the regression.

tests/web/test_model_service.py
```diff
@@ -661,19 +661,18 @@ def test_get_music_models_uses_independent_category(self):
         "get_default",
         [get_default_sound_effect_model, get_default_music_model],
     )
-    def test_audio_generation_default_closes_database_generator(self, get_default):
+    def test_audio_generation_default_closes_owned_session(self, get_default):
```
rogercloud, line 678: **MINOR (test gap) — un-migrated getters have no session-lifecycle coverage.**

Neither `get_default_embedding_model` nor `get_default_rerank_model` is imported or exercised here (the existing embedding tests target `DynamicMemoryStoreManager._get_embedding_model_from_db`, a different path). The session-close test added for the migrated getters (`test_audio_generation_default_closes_owned_session`) has no equivalent for the un-migrated ones, so their pre-existing `next(get_db())`-without-close pattern is untested. If F2 leaves them unmigrated, add at least a close/lifecycle test to document current behavior.

src/xagent/web/services/db_runtime.py
```diff
@@ -0,0 +1,79 @@
...
+    cancellation is propagated so a late transaction cannot race cleanup.
+    """
+    worker = asyncio.get_running_loop().create_task(asyncio.to_thread(operation))
+    result, cancellation = await await_task_settlement(worker)
+    if cancellation is not None:
+        raise cancellation
+    return result
+
+
+async def drain_async_task_cancellation_safe(task: asyncio.Task[_T]) -> _T:
```
rogercloud: **delete — `drain_async_task_cancellation_safe` has zero production callers.** Only its own dedicated test (`tests/web/services/test_db_runtime.py:129`) exercises it. Delete the function and that test, or fold its logic into a test helper if the settlement-only coverage is still wanted.

src/xagent/web/services/db_runtime.py
```diff
@@ -0,0 +1,79 @@
...
+    seen: set[int] = set()
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        if isinstance(current, SQLAlchemyTimeoutError):
+            return True
+        current = current.__cause__ or current.__context__
+    return False
+
+
+async def await_task_settlement(
```
rogercloud, line 25: **yagni (optional) — `await_task_settlement` was split out to back two callers; with `drain_async_task_cancellation_safe` removed (see above) it has one real production caller (`run_db_io_cancellation_safe`).** Could be folded back inline. Note it currently has standalone unit coverage of the settlement/cancellation-recording logic in isolation from the thread-offload wrapper, which is a legitimate reason to keep it separate — your call.

src/xagent/core/tools/adapters/vibe/factory.py
```diff
@@ -282,6 +282,26 @@ class ToolFactory:
     @staticmethod
     async def create_all_tools(
         config: BaseToolConfig, apply_user_override_filter: bool = True
+    ) -> List[Tool]:
+        """Create tools within the config's optional prepared-runtime boundary."""
+        prepare_factory_runtime = getattr(type(config), "prepare_factory_runtime", None)
```
rogercloud, line 290: **yagni — the `getattr(obj_or_type, "hook_name", None) + callable()` probing for optional lifecycle hooks is duplicated across 4 sites** (here, plus `agent/service.py:98`, `:647`, `:661`) for `prepare_factory_runtime` / `release_prepared_factory_runtime` / `refresh_runtime_policy`.

`BaseToolConfig` already sets a no-op-default precedent (`get_video_models`, `emit_mcp_load_summary`, `release_db_connection` are concrete no-op methods, not probed). Add no-op defaults for these three hooks on `BaseToolConfig` (each keeping its own sync/async signature) and call them directly (e.g. `await config.prepare_factory_runtime()`). This stays compatible with class-level lookup, since a base-class default is found by class-level lookup too — so it doesn't reintroduce the Mock concern from the earlier automated review.

src/xagent/web/tools/config.py
```diff
@@ -684,36 +1182,54 @@ def get_vision_model(self) -> Optional[Any]:
...
             return self._explicit_vision_model
 
+        snapshot = self._factory_runtime_snapshot
+        if snapshot is not None and snapshot.plan.load_vision:
+            return snapshot.vision_model
         if self._cached_vision_config is None:
             self._cached_vision_config = self._load_vision_model()
         return self._cached_vision_config
 
     def get_image_models(self) -> Dict[str, Any]:
```
rogercloud, line 1200: **shrink — ~14 near-identical snapshot-or-legacy accessors** (`get_image_models` here, `get_video_models` ~L1201, `get_asr_models` ~L1985, `get_tts_model` ~L2048, `get_music_model` ~L2114; spanning ~L1180–L2131) all follow the same shape: "if snapshot present and plan flag set → return snapshot field; else lazily load via legacy path and cache."

A `_snapshot_or_legacy(plan_flag, snapshot_attr, cache_attr, loader)` helper would collapse most of these to one-liners. Exclude `get_tool_credential` / `get_sql_connections` (no caching, raise on `failed_inputs`) and `get_custom_api_configs` (inline query, no delegated loader) — they don't fit the uniform shape.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed in `f37e0a1f`. Each logical snapshot input now creates and closes its own short-lived session. The worker forces checkout before entering legacy broad catches, so pool timeouts propagate, and an aborted transaction from one loader cannot affect a later loader. Session isolation is used instead of relying on rollback or savepoints because the loaders have heterogeneous transaction behavior. The regression suite covers cross-read failure isolation, policy-hook isolation, later-checkout timeout propagation, cleanup, and constrained-pool behavior.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed in `f37e0a1f`. Snapshot inputs now go through one `load_snapshot_input` owner with an explicit policy: `basic` and `database` record hard failures in `failed_inputs`; Custom API, published-agent, and model inputs preserve their legacy soft-fail defaults; `ConnectorRuntimeError` and pool checkout timeouts propagate. This removes the ad hoc branch asymmetry while preserving existing product behavior.

tests/web/tools/test_web_tool_config_session_factory.py
AlexLiu190625: Covered in `f37e0a1f`. Added deterministic regressions for a swallowed PostgreSQL-style aborted transaction followed by a successful unrelated read, a required-input failure followed by successful SQL loading, policy-hook isolation, later-input checkout-timeout propagation and cleanup, and sequential high-water with a pool that permits more than one checkout.

src/xagent/web/services/db_runtime.py
AlexLiu190625: Fixed in `e241e973`. Removed `drain_async_task_cancellation_safe` and its dedicated test from this PR. The helper belongs with the core runtime cleanup callers introduced by the separate runtime change, so #964 no longer carries an unused API.

rogercloud (commented):
## Follow-up Review

*Re-review at head `e241e973`, base `06e3b456`. Prior review posted at `77e082df`.*

Two commits landed. `f37e0a1f` (*isolate tool snapshot database reads*) rewrites `_load_tool_factory_runtime_snapshot` and `_load_tool_runtime_policy_snapshot` in `src/xagent/web/tools/config.py` so that **every** prefetch input — credentials, SQL, custom API, published-agent records, each model type, and each policy override/allowlist read — now opens and closes its **own** SQLAlchemy `Session` via new `load_snapshot_input` / `load_policy_input` closures, replacing the previous single shared worker session. `e241e973` (*move task draining to runtime callers*) deletes the now-dead `drain_async_task_cancellation_safe` function and its test from `src/xagent/web/services/db_runtime.py`. Five new regression tests were added in `tests/web/tools/test_web_tool_config_session_factory.py`, backed by a new `_PostgresAbortSession` double that models Postgres' abort-until-rollback transaction semantics.

### Design verdict: Sound *(upgraded from acceptable-with-reservations)*

The reservation in the prior review was driven entirely by the F1 regression (a shared worker session with no rollback, which under Postgres poisons every subsequent read in the snapshot once one read hits a caught SQL error). Per-input session isolation structurally eliminates that cascade — each read runs on its own connection/transaction, so an aborted transaction can no longer contaminate a later one. The added per-input checkout/checkin cost is acceptable: reads are sequential within a single worker thread, so peak concurrent pool usage is unchanged (still 1); only the total checkout *count* per snapshot rises. Correctness-for-cost tradeoff worth making.

### Status of every prior finding

| # | Finding (severity) | Status | Note |
|---|---|---|---|
| **F1** | Shared worker session had no rollback → cascading Postgres failures (MAJOR) | **FIXED** | Every prefetch and policy input now opens/closes its own session. Verified against the diff; 5 new tests exercise the isolation guarantee with non-tautological assertions and all pass. |
| **F2** | `get_default_model` / `get_fast_model` / `get_compact_model` un-migrated, still on the event-loop path (MAJOR/MINOR) | **NOT FIXED** | `model_service.py` byte-for-byte unchanged; still `next(get_db())` directly, not in the snapshot. PR "Out of scope" section not updated to acknowledge the gap. |
| **F3** | Three-way asymmetry in `failed_inputs` tracking (MINOR) | **REFACTORED** | Code-quality/duplication half resolved (all three styles unified into `load_snapshot_input`). Semantic half unchanged: `custom_api` / `published_agent` / model-type loaders still soft-fail without `unavailable_input=` while `basic`/`database` hard-fail — but this is now explicitly documented as intentional in the closure docstring. No longer block-worthy; worth a second look at whether custom_api/published-agent *should* be required. |
| **F4** | No test for cross-read failure isolation (MINOR) | **FIXED** | 5 new tests cover both snapshot and policy loaders, both "isolation holds when a loader swallows an error" and "a later loader's pool-timeout still propagates." |
| **F5** | No session-lifecycle test for embedding/rerank getters (MINOR) | **NOT FIXED** | `tests/web/test_model_service.py` unchanged; `get_default_embedding_model` / `get_default_rerank_model` still have zero test references. |
| **D1** | Shared default-executor blast radius (design, non-blocking) | **Not addressed** (expected) | Flagged as forward-looking; `run_db_io_cancellation_safe` still uses bare `asyncio.to_thread`. No action required for this PR. |
| Simpl. | Delete `drain_async_task_cancellation_safe` | **FIXED** | Function + its test + now-unused import fully removed; zero remaining references. |
| Simpl. | Fold `await_task_settlement` back inline (optional) | Not done — fine | Retains valuable standalone unit-test coverage. |
| Simpl. | Consolidate 4 hook-probing sites (optional) | Not done | `factory.py`/`service.py`/`vibe/config.py` unchanged; still stands for a future pass. |
| Simpl. | Consolidate ~14 snapshot-or-legacy accessors (optional) | Not done | Outside these commits' diff scope; still stands. |
| Gemini | busy-loop/uncancel, `type(config)` mock lookup, Py3.10 compat | **Still WAIVED** | All three concern code byte-for-byte unchanged since the last review. |

### New findings (all Low/Info — optional polish, no blockers)

**N1 (Low, code quality)** — `config.py`, `load_snapshot_input` (~665-708) and `load_policy_input` (~907-930) duplicate the same "open own session → try loader → classify exception → close in `finally`" skeleton, differing only in the loader signature and whether a `unavailable_input` / `propagated_exceptions` / `log_level` classification layer is present. Could be unified into one shared isolated-session-loader helper. Not a bug.

**N2 (Low, consistency)** — The required-vs-optional failure split (credentials/SQL hard-fail via `failed_inputs`; custom_api/published-agent/all model types soft-fail to a default) is intentional and documented, but the classification is expressed only *implicitly*, per-call-site, by which keyword args are passed. There is no single declarative list of "which inputs are required," so a new required input could be miswired as optional (or vice versa) without it being obvious. Suggest a short comment or a central set/enum documenting the classification.

**N3 (Info, diagnosability)** — The four audio loader calls (`asr_models`, `tts_models`, `sound_effect_models`, `music_models`, ~config.py:797-864) all pass `input_name="audio"`, so a failure log for any of the four is indistinguishable from the others. Consider `"audio:asr"`, `"audio:tts"`, etc.

**N4 (Info, scope note)** — Confirms the un-migrated getters from F2 remain out of scope for this refactor; no *new* instance found beyond what F2 already covers.

### New simplification opportunities

- **`log_level` dead flexibility** — `load_snapshot_input`'s `log_level: int = logging.WARNING` is passed a non-default value (`logging.ERROR`) at exactly one of its ~14 call sites (the Custom API loader). Could be dropped in favor of a hardcoded `logger.error(...)` at that one site, unless a second non-default use is anticipated.
- **`propagated_exceptions` dead flexibility** — the `propagated_exceptions: tuple[type[Exception], ...] = ()` parameter is only ever passed a single-element tuple `(ConnectorRuntimeError,)` at one call site. Could simplify to `propagated_exception: type[Exception] | None = None` with an `isinstance` check guarded by `is not None`.
- **N1 duplication** (same finding as above) — unifying `load_snapshot_input` / `load_policy_input` removes one copy of the skeleton.

Net: these are small subtractive cleanups (roughly a net reduction once the two params are narrowed and the two closures share a helper); all optional, none affect correctness.

### Testing

Independently re-ran the full suite across the 6 affected modules at the new head: **151 passed, 0 failed** (`tests/core/agent/test_agent_service_tool_refresh.py`, `tests/core/tools/test_agent_tool_publish_status.py`, `tests/core/tools/test_vision_tool.py`, `tests/web/services/test_db_runtime.py`, `tests/web/test_model_service.py`, `tests/web/tools/test_web_tool_config_session_factory.py`). The new `_PostgresAbortSession` double is justified — SQLite has no abort-until-rollback mode, so this behavior genuinely cannot be reproduced without a hand-rolled double; not over-engineering.

Nit (not a problem): the PR description still says "147 tests." That count is now stale (+5 isolation tests, −1 dead-code test → net +4). Documentation staleness only — everything passes.

### Recommendation

The MAJOR issue (F1) is fixed and the design is now sound. Two genuinely open items remain, both pre-existing and both MINOR:

- **F2** — either migrate `get_default_model` / `get_fast_model` / `get_compact_model` into the snapshot, or add an explicit scope note (and update the PR body's "Out of scope" section) acknowledging they still run `next(get_db())` on the event-loop path.
- **F5** — add a session-lifecycle test for `get_default_embedding_model` / `get_default_rerank_model`.

Everything else is either fixed or optional/non-blocking (N1-N3, the two simplification items, D1, and the remaining optional simplifications). I'd treat F2 and F5 as the only things worth resolving — or explicitly deferring with a note — before merge; N1-N3 and the simplifications can land in a follow-up.

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,435 @@ async def refresh_oauth_token_if_needed(
...
+    asr_model: Any | None = None
+    tts_models: dict[str, Any] = {}
+    tts_model: Any | None = None
+    sound_effect_models: dict[str, Any] = {}
+    sound_effect_model: Any | None = None
+    music_models: dict[str, Any] = {}
+    music_model: Any | None = None
+    published_agent_records: list[Any] = []
+
+    def load_snapshot_input(
```
rogercloud, line 674: **N1 — Low / code quality (optional).** `load_snapshot_input` here and `load_policy_input` in `_load_tool_runtime_policy_snapshot` (~line 907) share the same skeleton — open a dedicated session, call `db.connection()` to fail fast, run the loader, classify the exception, and close in `finally`. They differ only in the loader signature and whether the `unavailable_input` / `propagated_exceptions` / `log_level` classification layer is present. Consider extracting a single shared isolated-session-loader helper so the isolation invariant lives in one place. Not a bug — the current form is correct.

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,435 @@ async def refresh_oauth_token_if_needed(
...
+        loader: Callable[[Any], Any],
+        default: Any,
+        *,
+        unavailable_input: str | None = None,
+        propagated_exceptions: tuple[type[Exception], ...] = (),
+        log_level: int = logging.WARNING,
+        log_message: str | None = None,
+    ) -> Any:
+        """Load one logical input through an isolated Session boundary.
+
```
rogercloud, line 685: **N2 — Low / consistency (optional).** The required-vs-optional split — credentials and SQL hard-fail via `failed_inputs`, while custom_api, published-agent, and all model-type loaders soft-fail to a default — is intentional and documented here, but "required" is expressed only implicitly per call site by whether `unavailable_input=` is passed. There's no single declarative list of which inputs are required, so a future required input could easily be miswired as optional (or vice versa) without it being obvious. A short comment or a central set/enum naming the required inputs would make the classification harder to get wrong.

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,435 @@ async def refresh_oauth_token_if_needed(
...
+        )
+    if plan.load_video:
+        video_models = load_snapshot_input(
+            "video",
+            lambda db: model_service.get_video_models(db, plan.user_id),
+            {},
+        )
+    if plan.load_audio:
+        asr_models = load_snapshot_input(
+            "audio",
```
rogercloud: **N3 — Info / diagnosability (optional).** All four audio loaders (`asr_models`, `tts_models`, `sound_effect_models`, `music_models`) pass `input_name="audio"`, so a failure log for any one of them is indistinguishable from the others. Consider distinct names like `"audio:asr"`, `"audio:tts"`, `"audio:sound_effect"`, `"audio:music"` to make failures traceable to the specific model type.

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,435 @@ async def refresh_oauth_token_if_needed(
...
+    published_agent_records: list[Any] = []
+
+    def load_snapshot_input(
+        input_name: str,
+        loader: Callable[[Any], Any],
+        default: Any,
+        *,
+        unavailable_input: str | None = None,
+        propagated_exceptions: tuple[type[Exception], ...] = (),
+        log_level: int = logging.WARNING,
```
rogercloud, line 681: **Simplification (optional) — `log_level` dead flexibility.** This is only ever passed a non-default value (`logging.ERROR`) at a single call site (the Custom API loader) out of ~14. Unless a second non-default use is anticipated, the parameter could be dropped and that one site could call `logger.error(...)` directly, removing a rarely-used knob from the signature.

src/xagent/web/tools/config.py
```diff
@@ -443,6 +517,435 @@ async def refresh_oauth_token_if_needed(
...
+    music_model: Any | None = None
+    published_agent_records: list[Any] = []
+
+    def load_snapshot_input(
+        input_name: str,
+        loader: Callable[[Any], Any],
+        default: Any,
+        *,
+        unavailable_input: str | None = None,
+        propagated_exceptions: tuple[type[Exception], ...] = (),
```
rogercloud, line 680: **Simplification (optional) — `propagated_exceptions` dead flexibility.** This is only ever passed a single-element tuple `(ConnectorRuntimeError,)` at one call site. Since no caller needs to propagate multiple types, this could be narrowed to `propagated_exception: type[Exception] | None = None` with an `isinstance(exc, propagated_exception)` check guarded by `is not None` — slightly simpler than iterating a tuple that is almost always empty.

tests/web/test_model_service.py
AlexLiu190625: Fixed in `243110f2`. `get_default_embedding_model` and `get_default_rerank_model` now use the shared `_default_model_session` boundary. Owned sessions close after normal returns, early returns, and query failures; caller-owned sessions are never closed. Query fallback semantics remain unchanged, and lifecycle tests also verify that session-factory and close failures are not swallowed.

src/xagent/web/services/db_runtime.py
AlexLiu190625: I’m keeping this helper. The dependent task-runtime split already uses the same settlement contract in database-worker draining, cancellation-safe lease acquisition, and heartbeat shutdown. Those paths all need to shield an owned task, record caller cancellation, wait for settlement, inspect late results for cleanup or compensation, and preserve a child failure as the cancellation cause. The standalone test isolates that contract from `asyncio.to_thread`; inlining it here would immediately require re-extraction or duplicate the cancellation logic.

src/xagent/core/tools/adapters/vibe/factory.py
AlexLiu190625: I’m keeping these hooks as optional capabilities. `AgentService` uses the presence of `refresh_runtime_policy` to decide whether to run the legacy synchronous refresh; adding a no-op method to `BaseToolConfig` would make every subclass appear async-capable and could suppress that legacy refresh. The codebase also accepts duck-typed tool configs that do not inherit `BaseToolConfig`, so unconditional calls would not be semantics-preserving. A broader interface migration can be evaluated separately.

src/xagent/web/tools/config.py
AlexLiu190625: I’m deferring this refactor. The accessors share a superficial shape, but the candidate subset still has different snapshot-plan conditions, cache defaults, and loader signatures. A string-based attribute helper would replace visible typed accesses with dynamic lookup without strengthening the session-isolation invariant addressed here. This can be evaluated separately with a typed design rather than expanding this PR’s scope.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed in `243110f2`. Snapshot and policy inputs now share `_run_with_checked_out_session`, which reuses `tool_session_scope`, eagerly checks out the connection, and always closes the per-input Session. Their distinct failure-classification policies remain in their owning loaders, so factory, checkout, and close failures stay outside soft fallback. Focused tests cover success, operation failure, close failure, cross-read isolation, and later checkout timeouts.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed in `243110f2`. The implicit `unavailable_input` marker is now the typed `failed_input_key: Literal["basic", "database"]`, and the loader contract documents the complete policy: those two keys fail only their matching creator closed; Custom API, published-agent, and model discovery keep their legacy soft defaults; pool timeouts and explicitly propagated runtime errors still escape.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed in `243110f2`. All eight audio collection and default-model reads now have distinct input names (`audio:asr-models` through `audio:default-music`), with a parameterized regression test that verifies every emitted failure label.

src/xagent/web/tools/config.py
AlexLiu190625: I’m keeping `log_level` because the shared loader owns the exception boundary and therefore the failure log. The Custom API path intentionally preserves its existing ERROR severity while other optional inputs use WARNING. Removing the parameter would require duplicating catch-and-log logic or replacing it with an equivalent callback or flag, so it would not simplify the flow.

src/xagent/web/tools/config.py
AlexLiu190625: I’m keeping the tuple form because it is the loader’s exception-propagation allowlist and maps directly to `isinstance`’s native tuple contract. The Custom API path currently supplies `ConnectorRuntimeError`; narrowing it to one optional type would add a `None` guard without removing the branch or changing behavior. The helper is local, so this creates no public API surface.

rogercloud (approved):
## PR #964 Re-review — `refactor(tools): prefetch runtime database state with worker-owned sessions`

### 1. Summary

This slice of #935 moves tool-factory setup off the "live-session-crossing-into-async-code" pattern that caused the SG pool-exhaustion incident. Tool credentials, SQL connections, tool policy, custom API config, published-agent records, and default model selections are now prefetched through worker-owned SQLAlchemy sessions that are always closed, and only detached data (dataclasses / dicts / detached ORM instances) crosses the async/thread boundary. The large diffs in `model_service.py` and `agent_tool.py` are the mechanical consequence of that goal, not bundled scope-creep.

### 2. What changed since the last review

Three follow-up commits landed on top of the previously reviewed head:

- **f37e0a1f** — introduced per-input session isolation: each prefetch read gets its own dedicated session so one read's failure can no longer cascade into unrelated reads (the base for finding D's fix); added the cross-read failure-isolation tests.
- **e241e973** — removed the dead `drain_async_task_cancellation_safe` helper (zero production callers) and its test.
- **243110f2** — tightened tool-snapshot session ownership: a shared `_run_with_checked_out_session` / `tool_session_scope` checkout skeleton, a typed `failed_input_key: Literal["basic", "database"]` with documented required-vs-optional policy, migration of `get_default_embedding_model` / `get_default_rerank_model` onto the shared `_default_model_session` boundary, and distinct per-loader failure labels.

### 3. Approach verdict — sound / acceptable

The design is the simplest correct approach that fits the codebase. Session lifecycle (open/close on every path including exceptions) is correct and is a net improvement over pre-PR, which leaked in places (e.g. `get_default_embedding_model` previously never closed its session). No thread-crossing live sessions or ORM objects remain. The backward-compat claims in the PR description hold against the code. Test coverage is strong: constrained-pool tests (`pool_size=1, max_overflow=0`) assert the event loop stays responsive during blocking work rather than only asserting mocked call counts.

### 4. New findings this round

All minor/info — none blocking.

**N1 — [minor, correctness/design] Non-timeout DB error during published-agent prefetch silently drops all delegation tools.**
`src/xagent/web/tools/config.py` (~795-803, inside `_load_tool_factory_runtime_snapshot`): the published-agent loader is called with no `failed_input_key`, so any non-pool-timeout exception silently degrades to `[]` — zero delegation/sub-agent tools — with only a WARNING and no caller-visible signal. This is preserved legacy behavior (pre-PR `get_published_agents_tools` had the identical try/except-then-`[]` shape), not a regression. But centralizing it is the natural moment to decide consciously: given the impact on agent workflows, consider whether losing all delegation tools deserves fail-closed treatment (a `failed_input_key`) like `basic`/`database`, rather than silent degradation.

**N2 — [minor, correctness] Credential/SQL fail-closed now has whole-category blast radius.**
`src/xagent/web/tools/config.py`: `load_tool_credentials` (~743-750) loops over every `TOOL_CREDENTIAL_SPECS` entry under a single shared try/except, so one entry's DB error trips `failed_inputs.add("basic")` for the whole batch; `get_tool_credential` (~1905-1910) then raises `RuntimeError` for every tool needing any credential, not just the one that failed. This is a genuine widening vs. pre-PR (base did per-tool, per-call lookups, isolating failures). Mitigated by `_decrypt` in `tool_credentials.py` swallowing `InvalidToken`/`ValueError` → `None`, so only a real DB error during the batch query triggers it (rare) — hence minor. Suggestion: track credential failures per-spec so the blast radius matches the actual failure.

**N3 — [minor, robustness/nit] `release_prepared_factory_runtime` is skipped if `prepare_factory_runtime` raises.**
`src/xagent/core/tools/adapters/vibe/factory.py` `create_all_tools` (~283-300): `await prepare_factory_runtime(config)` sits before the `try/finally` that runs the release hook, so a raise in prepare skips release. Not an active leak today — `_pending_runtime_policy` is cleared unconditionally at the top of `prepare_factory_runtime` (`config.py` ~1732-1733), the only state-mutating call (`_apply_factory_runtime_snapshot`) is the last statement of prepare (~1746), and every per-input worker session closes via `tool_session_scope`'s `finally` regardless. Recommend moving `await prepare_factory_runtime(config)` inside the `try` for self-evident symmetry and to protect the invariant against future edits to `prepare`.

**N4 — [minor, info/design trade-off] Per-input session multiplication under a saturated pool.**
`src/xagent/web/tools/config.py`: `_run_with_checked_out_session` opens a dedicated session per logical input — up to ~18-20 sequential checkout/close cycles per full build (18 in `_load_tool_factory_runtime_snapshot` + 2 in `_load_tool_runtime_policy_snapshot`), up from a single shared session pre-PR. This is a deliberate, tested isolation choice (`test_factory_prefetch_isolates_later_read_from_swallowed_sql_failure` — the fix for prior MAJOR finding D): a swallowed abort in one loader must not poison a later one. The cost is bounded (failure path caps at ~1 pool_timeout since timeouts propagate immediately; only the success-but-slow path can approach N× latency). Not a blocker — flag as a known trade-off worth watching in production pool metrics.

**N5 — [info, out of scope] Shared default executor for DB-runtime offload.**
`src/xagent/web/services/db_runtime.py`: `asyncio.to_thread` uses the shared default `ThreadPoolExecutor` with no dedicated/sized executor for this DB-runtime work. Event-loop responsiveness is still met (confirmed by the ticker test). Thread-pool sizing/isolation belongs to the separate task-runtime-isolation follow-up (#935's remaining scope), not this PR.

### 5. Prior findings — resolution status

All 16 prior findings (A-P) re-verified against head `243110f2`.

|---|---|---|---|
| A | gemini: `db_runtime.py:47` busy-loop on repeated cancellation | **WAIVED-justified** | `Task.cancelling()>0` is an informational counter, not a re-trigger; `shield` doesn't re-raise `CancelledError` per subsequent await. Gemini's premise was a mistaken cancellation model. |
| B | gemini: `factory.py:300` `type(config)` bypasses instance overrides | **WAIVED-justified** | Instance-level `getattr` on an unspecced `MagicMock` synthesizes a callable → `await hook(config)` raises `TypeError`, breaking 2 tests in `test_selection_spec.py`. No instance-only override pattern exists in repo. |
| C | gemini: `db_runtime.py:43` `Task.cancelling()` needs 3.10 fallback | **WAIVED-justified** | `pyproject.toml` requires `>=3.11`; Docker + all CI pin 3.11/3.12. No 3.10 path exists. |
| D | rogercloud MAJOR: shared worker session, no rollback, cross-read cascade | **FIXED** | Every ~18-20 prefetch read now opens/closes its own session; 2 regression tests confirm isolation and recovery after a swallowed/required failure. |
| E | rogercloud MINOR: `custom_api`/`published_agent` never touch `failed_inputs` | **FIXED** | Shared loader now takes typed `failed_input_key` with documented policy; asymmetry is now explicit and intentional (soft-fail behavior itself unchanged by design). |
| F | rogercloud MINOR (test gap): no cross-read failure-isolation test | **FIXED** | 2 tests confirm a different loader in the same build still succeeds with distinct sessions. |
| G | rogercloud MINOR (test gap): `get_default_embedding/rerank_model` lifecycle | **FIXED** | Both now use `_default_model_session`; 14 parametrized tests cover owned-close on success/early-return/error and non-close of caller-owned sessions. |
| H | rogercloud "delete": `drain_async_task_cancellation_safe` has zero callers | **FIXED** | Removed in e241e973 (grep-confirmed zero hits, function + test deleted). |
| I | rogercloud "yagni": `await_task_settlement` foldable into 1 caller | **WAIVED-plausible** | Exactly 1 in-repo caller with genuine standalone cancellation-contract coverage; forward-reuse rests on an unmerged PR but keep-cost is low and the isolated test is a real benefit. Accepted nit. |
| J | rogercloud "yagni": hook-probing via `getattr` across 4 sites; suggest no-op defaults | **WAIVED-justified** | `AgentService` (`service.py` ~98-100, 647-661) gates the legacy sync refresh on `callable(getattr(..., "refresh_runtime_policy", None))`; a no-op default would permanently suppress that path. `test_agent_service_tool_refresh.py` actively asserts on this presence-check. |
| K | rogercloud "shrink": ~14 snapshot-or-legacy accessors → 1 helper | **WAIVED-questionable** | The ~13 accessors are structurally identical modulo 3 flag-name strings — the "different conditions/defaults/signatures" rationale doesn't hold. But it's a pure DRY call with zero correctness risk; defer to a focused follow-up PR. |
| L | rogercloud N1: `load_snapshot_input`/`load_policy_input` share a skeleton | **FIXED** | Both delegate to `_run_with_checked_out_session`/`tool_session_scope` while keeping their distinct classification logic. |
| M | rogercloud N2: no central declarative "required" list | **FIXED** | Typed `Literal` alias + docstring on `load_snapshot_input` enumerating the full required-vs-optional policy. |
| N | rogercloud N3: 4 audio loaders share generic `input_name="audio"` | **FIXED** | 8 distinct labels (`audio:asr-models` … `audio:default-music`), with a parametrized test asserting each label in the failure log. |
| O | rogercloud: `log_level` param is dead flexibility | **NOT FIXED** | Signature still `log_level: int = logging.WARNING`; 1 of 18 call sites passes non-default (`logging.ERROR`). Low-priority. |
| P | rogercloud: `propagated_exceptions` dead flexibility | **NOT FIXED** | Still `tuple[type[Exception], ...] = ()`; 1 of 18 call sites passes a single-element `(ConnectorRuntimeError,)`. Low-priority. |

### 6. Simplification opportunities

```
config.py:233  delete  _ToolFactoryRuntimeLoadPlan.requires_direct_db_without_policy — zero call sites (grep-confirmed). Drop the ~17-line property.
config.py:659  shrink  _load_tool_factory_runtime_snapshot hand-declares ~19-20 locals re-passed by keyword to _ToolFactoryRuntimeSnapshot(...). Use a dict accumulator + **unpack (names map 1:1); wrap failed_inputs in frozenset(...) and handle tool_overrides/tool_allowlist from runtime_policy explicitly.
config.py:259  shrink  _ToolFactoryRuntimeSnapshot re-declares tool_overrides/tool_allowlist already modeled by _ToolRuntimePolicySnapshot. Embed policy: _ToolRuntimePolicySnapshot; 4 read sites (config.py:1708-1709, test_web_tool_config_session_factory.py:966-967) become snapshot.policy.*.
config.py       shrink  load_snapshot_input log_level param is dead flexibility — 1 of 18 sites non-default. Drop the param; log at the fixed level, special-case the one ERROR site. (prior finding O, still open)
config.py       shrink  propagated_exceptions is always () or a single-element tuple. Narrow to propagated_exception: type[Exception] | None = None. (prior finding P, still open)

net: ~-40 lines possible
```

### 7. Recommendation

**Approve with minor, non-blocking suggestions.** Zero critical or major issues survive across both the fresh review and the re-verification of all 16 prior findings — the one prior MAJOR (D) is fixed and confirmed by regression tests. The 5 new findings are minor/info, the remaining waivers are well-founded, and the simplification items are pure follow-up polish. N1 (silent loss of all delegation tools) is the one item worth an explicit conscious sign-off before merge, but it preserves existing behavior and does not block.

PR #951, opened 2026-07-21
fix(triggers): exclude workforce-generated manager agents from trigger ownership resolution

codeacme17:
## What

Add the workforce-generated manager agent exclusion to `get_owned_agent` in `src/xagent/web/services/triggers.py`, which resolves agent ownership for all trigger management operations (CRUD, listing, runs, test-fire).

## Why

Manager agents are private implementation details of a workforce. Every other externally-reachable channel already excludes them — `api/share.py`, `api/widget.py`, `api/v1/deps.py`, and `services/api_keys.py` (`_owned_agents_query`) — but trigger management did not, so triggers could be attached to a workforce manager agent. A trigger firing on a manager agent would build a plain `Task` without `workforce_run_id`, silently losing all worker delegation.

The fix applies the same SQL-level `Agent.origin != WORKFORCE_GENERATED_MANAGER` filter used by `AgentApiKeyService._owned_agents_query`, so all trigger routes now return 404 for manager agents.

## Testing

- New regression test `test_trigger_routes_reject_workforce_manager_agent` (list + create both 404).
- Full trigger suites pass: `test_agent_triggers.py`, `test_gmail_auto_triggers.py`, `test_trigger_payload_retention.py`, `test_trigger_rate_limits.py`, `test_trigger_audit.py`, `test_trigger_provider_pipeline.py`, `test_trigger_dispatcher_gmail_gate.py` (68 passed, 1 skipped requiring Postgres).
- `ruff format --check`, `ruff check`, `mypy` clean on changed files.

Closes #945. Part of #805.

Issue #805, opened 2026-07-08
feat(workforce): Add deployment options (Widget/REST API/SDK/Webhook) for Workforce


yiboyasss:
## Description

Workforce currently cannot be used outside of Xagent. Requesting the same deployment options that individual agents have (Widget, REST API/SDK, Shareable Link, Webhook).

## Expected result

-

## Actual result

-

codeacme17:
I looked into how the four deployment channels are implemented for standalone agents and how Workforce executes today. The runtime path is already shared — a workforce run is just a `Task` executed by the generated manager agent (`workforce_runs.py` → `TaskTurnOrchestrator`), so all four channels can reuse the existing public routers and chat runtime. The open questions are all about **where the deployment config lives and how external entry maps to a `WorkforceRun`**. Listing them here for alignment before implementation:

## Key design decision: which layer owns the deployment config

- **Option A — unblock the manager agent**: let the generated manager agent use the existing `widget_key` / `share_token` / `AgentApiKey` / `AgentTrigger` directly. Smallest diff, but the manager agent is a private implementation detail (`WORKFORCE_GENERATED_MANAGER`, explicitly rejected today in `api/share.py` and `api/widget.py`), and hanging external credentials on it makes manager rebuild/replace operations dangerous. It also bypasses `WorkforceRun` bookkeeping (snapshot, runs history).
- **Option B — workforce-level config (recommended)**: add deployment fields/tables on `Workforce` (`widget_key`, `share_token`, allowed domains, a `WorkforceApiKey` table, workforce-level triggers). Public entry points resolve the workforce token → validate `status == active` → go through `create_workforce_run`, so external invocations produce proper `WorkforceRun` records with snapshots, and runs history stays meaningful.

The manager-agent blocks in `share.py` / `widget.py` should stay — external access should only enter via workforce-level credentials, never by addressing the manager agent directly.

## Open questions

1. **Conversation ↔ run mapping.** Widget/Share are multi-turn conversations, but `WorkforceRun` currently models "one run = one task". Proposal: one run per external conversation (append messages route to the run's task), with `sync_workforce_run_status` covering multi-turn lifecycles. Should runs history distinguish the entry channel (widget / share / api / webhook) with a `source` field?
2. **Snapshot semantics for long-lived external sessions.** Snapshots are pinned at run creation. If the workforce is edited while an external conversation is live, does the conversation keep the old snapshot (my assumption: yes) — and does a re-publish invalidate open share/widget sessions?
[...]

PR #966, opened 2026-07-22
feat(workforce): webhook/trigger deployment channel (#950)

### LINKED ISSUE #950: feat(workforce): webhook/trigger deployment
## Description

Webhook / trigger channel for Workforce. Last channel — has a hard dependency on `idempotency_key` from the foundation issue and benefits from validation already exercised by the earlier channels.

Scope (per the design discussion in #805, Q6):

- Add nullable `workforce_id` to `AgentTrigger`; workforce triggers fire via `create_workforce_run(source="trigger", idempotency_key=...)`, never by binding to the manager agent (a plain Task without `workforce_run_id` silently loses all delegation ability).
- New route group `/api/workforces/{workforce_id}/triggers` (existing routes embed `agent_id` in the path and can't be reused).
- `TriggerRun` status bookkeeping: trigger execution is a two-phase state machine (prepare → dispatch `begin_turn`) while `create_workforce_run` is one-shot — needs a dedicated branch.
- Ensure the trigger path routes through `create_workforce_run` so `validate_workforce_for_run` applies (the existing trigger code builds a plain `Task` directly — easy to copy that pattern and accidentally bypass validation).
- Frontend: trigger management for workforce in the deploy/triggers UI.

## Expected result

- A webhook (and other trigger providers where applicable) can start workforce runs with signature verification, idempotent retries, and `TriggerRun` history intact.

Part of #805. Blocked by #946. The `get_owned_agent` security fix ships separately beforehand.

## Additional scope (from #951 review)

Pre-existing `AgentTrigger` rows pointing at workforce manager agents (created before the #951 fix) become invisible/unmanageable via every API endpoint yet continue to fire through the dispatcher and webhook receivers — an "orphaned but still live" state. Fold the cleanup into this issue:

- Data migration to disable (or delete) `AgentTrigger` rows whose agent has `origin = workforce_generated_manager`.
- Defense-in-depth origin check in the firing path (`_attach_task_to_trigger_run` in `services/triggers.py`), so a manager-agent trigger can never construct a plain Task even if a row slips through.

codeacme17:
## What

Adds the webhook / scheduled / Gmail **trigger channel for Workforce** — the last external deployment channel from #805. Closes #950.

A workforce can now be started by an external event (signed webhook, cron schedule, or Gmail) with signature verification, idempotent retries, and full `TriggerRun` history — the same guarantees agent triggers already have.

## Backend

- **Model / migration**: `AgentTrigger` gains a nullable `workforce_id`; `agent_id` becomes nullable. Exactly one owner is set (service-layer invariant, documented on the model and in the migration). Migration also disables pre-existing "orphaned but still live" trigger rows bound to `workforce_generated_manager` agents (the #951 review follow-up).
- **Validated firing path**: workforce triggers fire through a new `create_workforce_run_record` (the synchronous half of `create_workforce_run`), so `validate_workforce_for_run`, the snapshot, and the config fingerprint all apply. A plain `Task` bound to the manager agent would silently lose all delegation ability — the dedicated `_attach_workforce_task_to_trigger_run` branch avoids that and preserves the two-phase `prepare → dispatch(begin_turn)` state machine.
- **Defense in depth**: the firing path refuses any trigger bound to a workforce-generated manager agent, even if a row slips past ownership resolution and the migration cleanup.
- **New route group** `/api/workforces/{workforce_id}/triggers` (CRUD + runs + test), scoped through `ensure_workforce_access`. The existing agent routes embed `agent_id` in the path and can't be reused.

## Frontend

- `agent-triggers-api` parameterized by a `TriggerOwnerRef`; the agent-scoped wrappers are kept for back-compat.
- `AgentTriggersDialog` takes an optional `owner`; a **Triggers** action on the workforce detail page opens it for the workforce.

## Tests

- Backend: workforce trigger CRUD / ownership / firing / dedup / scheduled-dispatch, the manager-agent defense guard, and migration up/down/idempotency.
- Frontend: owner-based routing in the api client.
- Full backend suite green (only the usual local PostgreSQL/KB skips).

## Manual E2E

Verified end-to-end against an isolated DB + live backend/frontend: enabling a webhook in the workforce Triggers dialog persists `workforce_id`-owned trigger, and a signature-verified callback creates a `WorkforceRun` whose hidden task carries `workforce_run_id` (delegation intact, `source=trigger`) — routed through the validated path, not a plain Task.


src/xagent/web/services/triggers.py
```diff
@@ -762,6 +888,66 @@ def _trigger_execution_context(
...
+    # Local import: workforce_runs pulls in the orchestrator stack, which
+    # would otherwise risk an import cycle with this module.
+    from .workforce_runs import create_workforce_run_record
+
+    user = db.query(User).filter(User.id == int(trigger.user_id)).first()
+    if user is None:
+        raise TriggerServiceError("Trigger owner not found")
+    workforce = (
+        db.query(Workforce).filter(Workforce.id == int(trigger.workforce_id)).first()
+    )
```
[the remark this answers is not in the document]
codeacme17: Addressed in 98fb26cd: the User and Workforce primary-key lookups now use Session.get().

src/xagent/web/services/triggers.py
```diff
@@ -780,9 +966,20 @@ def _attach_task_to_trigger_run(
         source_event_id=source_event_id,
         test=test,
     )
+    if trigger.workforce_id is not None:
+        return _attach_workforce_task_to_trigger_run(
+            db, trigger=trigger, run=run, prompt=prompt, test=test
+        )
     agent = db.query(Agent).filter(Agent.id == trigger.agent_id).first()
```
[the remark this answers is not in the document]
codeacme17: Addressed in 98fb26cd: the Agent primary-key lookup now uses Session.get().

src/xagent/web/api/triggers.py
```diff
@@ -447,6 +476,241 @@ async def test_trigger(
...
+
+
+def _workforce_or_404(
+    db: Session,
+    *,
+    user: User,
+    workforce_id: int,
+    action: str,
+) -> Workforce:
+    workforce = db.query(Workforce).filter(Workforce.id == workforce_id).first()
```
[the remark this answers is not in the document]
codeacme17: Addressed in 98fb26cd: the Workforce primary-key lookup now uses Session.get().

src/xagent/web/services/triggers.py
```diff
@@ -762,6 +888,64 @@ def _trigger_execution_context(
...
+    """
+    # Local import: workforce_runs pulls in the orchestrator stack, which
+    # would otherwise risk an import cycle with this module.
+    from .workforce_runs import create_workforce_run_record
+
+    user = db.get(User, int(trigger.user_id))
+    if user is None:
+        raise TriggerServiceError("Trigger owner not found")
+    workforce = db.get(Workforce, int(trigger.workforce_id))
+    record = create_workforce_run_record(
```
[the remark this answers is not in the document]
codeacme17: Fixed in 2661e936. The background trigger path now raises TriggerServiceError("Workforce not found") before entering the workforce run service, with a regression test covering the error type.

src/xagent/web/api/triggers.py
```diff
@@ -447,6 +476,241 @@ async def test_trigger(
...
+@router.get(
+    "/api/workforces/{workforce_id}/triggers/{trigger_id}/runs",
+    response_model=list[TriggerRunResponse],
+)
+async def list_workforce_trigger_runs_route(
+    workforce_id: int,
+    trigger_id: int,
+    current_user: User = Depends(get_current_user),
+    db: Session = Depends(get_db),
+) -> list[TriggerRunResponse]:
```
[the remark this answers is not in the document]
codeacme17: Fixed in 2661e936. This route is now a synchronous def so FastAPI runs its blocking SQLAlchemy work in the thread pool; a regression test locks the route execution model.

src/xagent/web/api/triggers.py
```diff
@@ -447,6 +476,241 @@ async def test_trigger(
...
+)
+async def get_workforce_trigger_run_route(
+    workforce_id: int,
+    trigger_id: int,
+    run_id: int,
+    request: Request,
+    include_payload: bool = False,
+    current_user: User = Depends(get_current_user),
+    db: Session = Depends(get_db),
+) -> TriggerRunResponse:
```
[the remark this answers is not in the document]
codeacme17: Fixed in 2661e936. The run-detail route is now a synchronous def so its blocking database work is offloaded by FastAPI; covered by the same regression test.

src/xagent/web/services/triggers.py
```diff
@@ -762,6 +888,66 @@ def _trigger_execution_context(
     }
 
 
+def _attach_workforce_task_to_trigger_run(
```
rogercloud, line 921: **Major — connector-runtime config silently dropped for workforce triggers.** This branch builds the run/task via `create_workforce_run_record` but never binds connector runtime to `record.task`. The agent-trigger branch (~990) calls `prepare_create_connector_runtime(..., payload_items=_trigger_connector_runtime_payload(trigger.config))`. Because the trigger `config` schema is shared (generic `config: dict[str, Any]`, same `_validate_config` incl. `_validate_persisted_connector_runtime_config`), a workforce trigger can define connector-runtime selections that validate successfully but are silently discarded at fire time. Please propagate `_trigger_connector_runtime_payload(trigger.config)` here to mirror the agent path, and add a test asserting the binding.

src/xagent/web/services/triggers.py
```diff
@@ -475,6 +499,62 @@ def create_agent_trigger(
...
+    workforce_id: int | None = None,
+    trigger_type: str,
+    name: str | None = None,
+    enabled: bool = True,
+    config: dict[str, Any] | None = None,
+    prompt_template: str | None = None,
+    secret: str | None = None,
+) -> tuple[AgentTrigger, str | None]:
+    if (agent_id is None) == (workforce_id is None):
+        raise TriggerServiceError(
```
rogercloud, line 581: **Minor — exactly-one-owner invariant enforced only in app code.** With both `agent_id` and `workforce_id` now nullable, this guard is the only thing preventing a both-set/neither-set row. A direct/buggy insert bypassing the service layer could violate it. Recommend a DB-level backstop `CHECK (num_nonnulls(agent_id, workforce_id) = 1)` in the migration, plus a unit test exercising both the both-set and neither-set rejection branches (currently uncovered).

frontend/src/app/workforces/[id]/page-client.tsx
```diff
@@ -434,6 +447,14 @@ export default function WorkforceDetailPage() {
                     }
                 />
             </div>
+
+            <AgentTriggersDialog
+                agentId={null}
+                owner={{ kind: "workforce", id: id ?? workforce.id }}
```
Comment (rogercloud) [inline frontend/src/app/workforces/[id]/page-client.tsx:467]: **Minor — no test coverage.** The `owner={{ kind: "workforce", id: id ?? workforce.id }}` prop construction/passing is untested; only the api-client routing (`agent-triggers-api.test.ts`) is covered. Suggest a render test asserting `AgentTriggersDialog` receives a correctly constructed `owner`.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -222,7 +228,15 @@ export function AgentTriggersDialog({
 
   const stagedTriggersProp = staged?.triggers ?? null
   const isStaging = !isValidAgentId(agentId) && stagedTriggersProp !== null
-  const canOperate = isValidAgentId(agentId) || isStaging
+  // The live-CRUD target. Explicit owner (e.g. a workforce) wins; otherwise
+  // fall back to the agent. Memoized on primitives so it stays referentially
+  // stable across renders and can safely sit in effect/callback deps.
+  const resolvedOwner = useMemo<TriggerOwnerRef | null>(() => {
```
rogercloud, line 240: **Minor/nit — memoization assumption.** `resolvedOwner`'s deps are the primitives `[owner?.kind, owner?.id, agentId]` (with eslint-disable), which is correct only because the sole caller passes an inline object literal. A future caller passing a stable `owner` reference expecting identity-based memoization would silently lose referential stability. No change required; please expand the comment to make this assumption explicit.

rogercloud (commented):
## Re-review: feat(workforce): webhook/trigger deployment channel (#950)

**What changed.** This PR extends the existing agent-trigger channel to Workforce. `AgentTrigger` now carries a nullable `workforce_id` alongside a now-nullable `agent_id`, with an exactly-one-owner rule enforced in the service layer. Workforce triggers fire through `create_workforce_run_record` (via a dedicated `_attach_workforce_task_to_trigger_run` branch), which preserves the delegation/validation/snapshot/fingerprint guarantees rather than fabricating a plain `Task`. A new `/api/workforces/{workforce_id}/triggers` route group and a `TriggerOwnerRef`-parameterized frontend client/dialog complete the channel. A migration disables pre-existing orphaned triggers bound to workforce-generated manager agents, backed by a defense-in-depth guard in the firing path.

This is a **re-review**. All 6 prior findings from the earlier passes are confirmed FIXED.

### Prior findings — all resolved

1. `src/xagent/web/services/triggers.py:912` — User lookup now uses `db.get(User, ...)`. FIXED (98fb26cd).
2. `src/xagent/web/services/triggers.py:973` — Agent lookup now uses `db.get(Agent, ...)`. FIXED (98fb26cd).
3. `src/xagent/web/api/triggers.py:494` (`_workforce_or_404`) — Workforce lookup now uses `db.get(Workforce, ...)`. FIXED (98fb26cd).
4. `src/xagent/web/services/triggers.py:916-918` — Explicit `if workforce is None: raise TriggerServiceError(...)` guard added before `create_workforce_run_record`, preventing an HTTPException layer-violation leak in a background trigger context. FIXED (2661e936); regression test `test_missing_workforce_raises_trigger_service_error` passes.
5. `src/xagent/web/api/triggers.py:625` (`list_workforce_trigger_runs_route`) — Converted `async def` → `def` (blocking sync DB queries only). FIXED (2661e936); regression test `test_workforce_trigger_run_read_routes_are_sync` passes.
6. `src/xagent/web/api/triggers.py:652` (`get_workforce_trigger_run_route`) — Same conversion, same test. FIXED (2661e936).

### Design verdict

**Sound (acceptable with minor reservations).** The channel faithfully mirrors the agent-trigger architecture, and the central decision — reusing `create_workforce_run_record` as the synchronous half of `create_workforce_run` rather than building a bare `Task` — is correct and avoids silently dropping delegation. Duplication is well-factored through shared `_create_trigger`/`_apply_trigger_updates`/`_delete_trigger` helpers, and the two-phase (prepare→dispatch) vs one-shot reconciliation resolves to a single clean branch.

Two design-level concerns survived verification:

- **Connector-runtime config is silently dropped for workforce triggers (most significant finding this pass).** The trigger `config` schema is not owner-specific — both agent and workforce triggers accept a generic `config: dict[str, Any]` and run through the same `_validate_config` (including `_validate_persisted_connector_runtime_config`). A user can therefore set connector-runtime selections on a workforce trigger, have them validate successfully, and see them silently discarded at fire time, because `_attach_workforce_task_to_trigger_run` never calls `prepare_create_connector_runtime(...)` the way the agent branch does. This is a real, user-visible behavior gap, not a hypothetical.
- **Exactly-one-owner invariant is enforced only in application code.** With both `agent_id` and `workforce_id` now nullable, the invariant lives solely in `_create_trigger`'s guard. There is no DB-level backstop (e.g. `CHECK (num_nonnulls(agent_id, workforce_id) = 1)`), so a direct or buggy insert bypassing the service layer could persist a both-set/neither-set row.

### Line-level findings

**Major**

- **`src/xagent/web/services/triggers.py` (workforce branch ~891-948)** — `_attach_workforce_task_to_trigger_run` never binds connector runtime to `record.task`, whereas the agent branch (~990) calls `prepare_create_connector_runtime(..., payload_items=_trigger_connector_runtime_payload(trigger.config))`. Since the config schema and validation are shared, connector selections on a workforce trigger validate but are dropped at fire time. **Fix:** propagate `_trigger_connector_runtime_payload(trigger.config)` in the workforce branch (binding to `record.task`), mirroring the agent path; add a test asserting a workforce trigger with connector-runtime config actually binds it.

**Minor**

- **`src/xagent/web/models/trigger.py` + `src/xagent/web/services/triggers.py:554-556`** — The exactly-one-owner invariant has no DB constraint and no unit test exercising the both-set/neither-set branches. `_apply_trigger_updates` never touches these columns, so post-creation drift isn't a concern, but a direct insert could violate it. **Fix:** add a DB-level `CHECK (num_nonnulls(agent_id, workforce_id) = 1)` (via the migration) and a unit test covering the guard's rejection of both-set and neither-set inputs.
- **`frontend/src/app/workforces/[id]/page-client.tsx:55`** — `owner={{ kind: "workforce", id: id ?? workforce.id }}` passed to `AgentTriggersDialog` has no test coverage; only the api-client routing (`agent-triggers-api.test.ts`) is tested. **Fix:** add a render test asserting the dialog receives a correctly constructed `owner` prop.
- **`frontend/src/components/build/agent-triggers-dialog.tsx:234-238`** — The `resolvedOwner` useMemo depends on primitives `[owner?.kind, owner?.id, agentId]` with an eslint-disable. Correct today (the only caller passes an inline object literal), but a future caller passing a stable `owner` reference expecting identity-based memoization would silently lose referential stability. **Fix:** none required; strengthen the existing comment to call out this assumption explicitly.

### Tests

Backend tests for the changed area (`tests/web/api/test_workforce_triggers.py`, `tests/alembic/test_20260722_add_workforce_triggers.py`) pass in an isolated venv, including the two regression tests above run individually. No test covers the connector-runtime drop or the owner-invariant guard branches (see findings).

Adjacent, non-blocking: the agent-side `list_trigger_runs`/`get_trigger_run` routes (`api/triggers.py:334,360`) are still `async def` over sync DB calls — the same pattern this PR fixed for the workforce routes. Pre-existing and out of scope for #966; noting only as a cleanup opportunity.

Lean already.

codeacme17:
Thanks for the thorough re-review, @rogercloud. All findings addressed in 12419819.

**Major — connector-runtime silently dropped (workforce branch).**
Fixed, but by **rejecting** rather than propagating. The workforce run path goes through `create_workforce_run_record` → `prepare_connector_runtime_selection_snapshot`, which is *selection-only* by design — its docstring is explicit that "non-/v1 task creation paths do not accept per-invocation runtime payloads in this phase." So mirroring the agent branch's `prepare_create_connector_runtime(payload_items=...)` would actually contradict that documented contract. Instead, `_reject_workforce_connector_runtime` now rejects a `connector_runtime_context` payload on workforce triggers at create/update time, turning the validate-then-drop into a clear 400. If per-trigger connector runtime for workforces is wanted later, it's a deliberate follow-up on the workforce run path, not something to smuggle in through the shared trigger config. Tests cover both the create and update rejection.

**Minor — exactly-one-owner has no DB backstop.**
Added a portable `CHECK ((agent_id IS NULL) <> (workforce_id IS NULL))` in the migration (works on both SQLite and PostgreSQL), created inside the initial column-add batch (idempotent) and dropped in `downgrade`. Added a service-layer unit test for the guard (both-set / neither-set) and a migration test asserting the constraint rejects illegal inserts.

**Minor — frontend `owner` prop has no coverage.**
Attempted a `render()` test of `AgentTriggersDialog` asserting the owner flows to the API layer, but mounting the full dialog (its Radix dependency tree) reliably crashes the vitest worker ("Worker exited unexpectedly", ~200s, test bodies never run). A worker-crashing test is worse than none, so I dropped it. The owner→endpoint routing stays covered at two levels: `agent-triggers-api.test.ts` (owner-scoped URL routing) and the backend `test_workforce_triggers.py` (full CRUD + firing through the workforce routes).

**Minor — `resolvedOwner` useMemo deps.**
No code change (correct today); strengthened the comment to spell out the by-value-owner assumption behind the primitive deps.

Also folded in a small mypy fix in the migration (`_columns` return annotation). Full pre-commit (ruff / mypy / isort / codespell / npm lint / npm type-check) passes locally, single alembic head, and the trigger + migration + frontend api-client tests are green.

src/xagent/web/services/triggers.py
codeacme17: Fixed in 12419819 — by **rejecting** rather than propagating. The workforce run path (`create_workforce_run_record` → `prepare_connector_runtime_selection_snapshot`) is selection-only *by design*: its docstring states non-/v1 creation paths do not accept per-invocation runtime payloads. Mirroring the agent branch's `prepare_create_connector_runtime(payload_items=...)` would contradict that contract. `_reject_workforce_connector_runtime` now rejects a `connector_runtime_context` payload on workforce triggers at create/update time (400) instead of validate-then-drop. Per-trigger connector runtime for workforces can be a deliberate follow-up on the run path. Tests cover both create and update rejection.

src/xagent/web/services/triggers.py
codeacme17: Fixed in 12419819 — added a portable DB backstop: `CHECK ((agent_id IS NULL) <> (workforce_id IS NULL))` (works on both SQLite and PostgreSQL), created inside the initial column-add batch (idempotent) and dropped in `downgrade`. Also added a service-layer unit test for the guard (both-set / neither-set) and a migration test asserting the constraint rejects illegal inserts.

frontend/src/app/workforces/[id]/page-client.tsx
Comment (codeacme17) [reply frontend/src/app/workforces/[id]/page-client.tsx:467]: Partly addressed in 12419819. I attempted a `render()` test of `AgentTriggersDialog` asserting the owner flows to the API layer, but mounting the full dialog (its Radix dependency tree) reliably crashes the vitest worker ("Worker exited unexpectedly", ~200s, test bodies never run), so I dropped it — a worker-crashing test is worse than none. The owner→endpoint routing stays covered at two levels: `agent-triggers-api.test.ts` (owner-scoped URL routing) and the backend `test_workforce_triggers.py` (full CRUD + firing through the workforce routes).

frontend/src/components/build/agent-triggers-dialog.tsx
codeacme17: Addressed in 12419819 — no code change (correct today); strengthened the comment to spell out the by-value-owner assumption behind the primitive deps, so a future caller relying on stable-reference identity is warned to revisit them.

src/xagent/web/models/trigger.py
```diff
@@ -57,7 +57,13 @@ class TriggerAuditOutcome(str, enum.Enum):
...
-    """Reusable automatic entry point for an agent."""
+    """Reusable automatic entry point for an agent or a workforce.
+
+    Exactly one of ``agent_id`` / ``workforce_id`` is set (enforced by the
+    service layer). Workforce triggers fire through ``create_workforce_run``
+    and are never bound to the workforce's generated manager agent: a plain
+    Task without a workforce run silently loses all delegation ability.
+    """
 
     __tablename__ = "agent_triggers"
```
rogercloud, line 69: **F1 (significant):** The owner-invariant CHECK constraint lives only in the migration's `upgrade()`, not on this model. `try_upgrade_db` stamps an empty DB straight to head and `init_db` then builds the schema from `Base.metadata.create_all()` — so every fresh install (new dev setups, new deployments, and the API test DB) gets no DB-level enforcement, only the Python service guard. Verified: the generated DDL on a fresh `init_db()` DB has no CHECK. This undercuts the DB-backstop goal of the owner-invariant fix.

Add the same constraint to the model so model-driven and migration-driven schemas agree:

```python
__table_args__ = (
    CheckConstraint(
        "(agent_id IS NULL) <> (workforce_id IS NULL)",
        name="ck_agent_triggers_single_owner",
    ),
)
```

src/xagent/migrations/versions/20260722_add_workforce_triggers.py
```diff
@@ -0,0 +1,152 @@
...
+            sa.column("agent_id", sa.Integer),
+            sa.column(WORKFORCE_ID_COLUMN, sa.Integer),
+        )
+        bind.execute(agent_triggers.delete().where(agent_triggers.c.agent_id.is_(None)))
+
+        if WORKFORCE_ID_INDEX in _index_names(inspector, TRIGGERS_TABLE):
+            op.drop_index(WORKFORCE_ID_INDEX, table_name=TRIGGERS_TABLE)
+        with op.batch_alter_table(TRIGGERS_TABLE) as batch:
+            # Drop the check before the column it references disappears.
+            batch.drop_constraint(SINGLE_OWNER_CHECK, type_="check")
```
rogercloud: **F3 (significant):** This drop is unconditional, but per F1 every fresh install is stamped to head without the constraint actually present. Running `alembic downgrade` on such a DB raises `ValueError: No such constraint` (reproduced locally in batch mode). Fixing F1 (constraint created via `create_all`) resolves this; otherwise guard the drop behind an existence/reflection check symmetric to the `needs_workforce_column` guard used in `upgrade()`.

src/xagent/web/services/triggers.py
```diff
@@ -762,6 +909,66 @@ def _trigger_execution_context(
...
+    task = record.task
+    setattr(
+        task,
+        "agent_config",
+        {
+            **dict(task.agent_config or {}),
+            **_trigger_execution_context(trigger=trigger, run=run, test=test),
+        },
+    )
+    db.add(task)
```
rogercloud: **F4 (minor, reachable edge case):** `create_workforce_run_record` commits the run+task, then this block does a second commit setting `TriggerRun.task_id`/status. The dedup guard in `prepare_trigger_run` (`:1077`) only blocks replay once `task_id` is persisted. If the process dies between the two commits, a retry replays via the idempotency key (`record.created=False`) and unconditionally re-overwrites `task.agent_config` and forces `run.status=PENDING`, regressing progress made in between. Narrow window, but confirmed reachable and untested. Make the re-attach idempotent — skip the config overwrite / status reset when the task is already attached, or persist `task_id` in the same commit as the run record.

frontend/src/app/workforces/[id]/page-client.tsx
Comment (rogercloud) [inline frontend/src/app/workforces/[id]/page-client.tsx:467]: Re: my prior finding on missing owner-prop test coverage. The reply states a render test was attempted but crashes the vitest worker via the Radix dependency tree — this doesn't hold up under verification. Commit 12419819 touched no test files here, and `frontend/src/components/build/agent-triggers-dialog.test.tsx` (which predates this PR) already does a full `render()` of `AgentTriggersDialog` through the real `Dialog`/`DialogPortal`/`DialogContent` tree: 13 tests, 256ms, no crash. Same component, same dependency tree.

Could you either add a render test asserting the resulting fetch URL/payload for an `owner={{ kind: "workforce", ... }}` prop (following that existing file's pattern), or acknowledge the coverage gap directly? The stated technical blocker isn't the reason.

rogercloud (commented):
## Follow-up review (head 8889033)

Re-reviewed after 12419819 ("fix: address workforce trigger re-review"), which addresses my 4 prior findings from 2661e93. Three of the four are genuinely fixed; the connector-runtime reject is architecturally the right call. The DB-constraint fix for the owner invariant is real but only reaches upgraded databases, not fresh installs — see F1/F3 below. The frontend test finding is not fixed, and its stated justification does not hold up. The merge with main pulled in unrelated commits (websocket hardening, agent-builder fixes); those are out of scope and excluded here.

### Prior findings

1. Connector-runtime config silently dropped for workforce triggers (Major) — **FIXED.** `_reject_workforce_connector_runtime` (`src/xagent/web/services/triggers.py:371-386`) is invoked from both `_create_trigger` (~579) and `_apply_trigger_updates` (~677), raising `TriggerServiceError` → 400. Reject (not propagate) is correct: `prepare_connector_runtime_selection_snapshot` (`connector_runtime.py:375-390`) takes no payload arg, so the workforce run path is structurally selection-only. Covered by `test_workforce_trigger_rejects_connector_runtime_context` (`tests/web/api/test_workforce_triggers.py:234-262`).

2. Owner invariant had no DB constraint / no test (Minor) — **FIXED WITH NEW GAP.** Migration adds `CHECK ((agent_id IS NULL) <> (workforce_id IS NULL))` correctly (valid XOR, portable, idempotent, correct downgrade ordering), plus raw-INSERT and service-layer tests. But the constraint lives only in the migration `upgrade()`, not the ORM model, so fresh installs never get it — see F1.

3. `page-client.tsx` owner prop has no test coverage (Minor) — **NOT FIXED.** Commit 12419819 touched zero test files for this. The reply states a render test was attempted but crashes the vitest worker via the Radix dependency tree. The stated technical blocker does not hold up under verification: `frontend/src/components/build/agent-triggers-dialog.test.tsx` (pre-dating this PR) already does a full `render()` of `AgentTriggersDialog` through the real `Dialog`/`DialogPortal`/`DialogContent` tree — 13 tests, 256ms, no crash. Same component, same tree. A render test asserting the resulting fetch URL/payload for an `owner={{ kind: "workforce", ... }}` prop is a normal, low-risk addition following that existing file's pattern.

4. `resolvedOwner` useMemo comment (nit) — **FIXED.** `agent-triggers-dialog.tsx:230-244` now documents the by-value-owner assumption and warns future callers about identity-based memoization. Complete.

### New findings

**F1 (significant) — Owner-invariant CHECK constraint is absent on all fresh installs.**
`src/xagent/web/models/trigger.py` — `AgentTrigger` has no `__table_args__`/`CheckConstraint`. The constraint exists only inside the migration's `upgrade()`. `try_upgrade_db` (`src/xagent/db/migration.py:285-309`) `stamp`s an empty DB straight to head, then `init_db` builds the schema via `Base.metadata.create_all()` from the ORM models — so fresh SQLite/PostgreSQL installs (new dev setups, new deployments, and the API test DB itself) get no DB-level enforcement, only the Python service guard. Verified by inspecting the generated DDL on a fresh `init_db()` DB: no CHECK. The migration's own unit test masks this because it hand-builds a legacy schema and calls `upgrade()` directly, bypassing the stamp-on-empty path. This undercuts the stated goal of finding #2's fix.
*Fix:* add the same `CheckConstraint((agent_id IS NULL) <> (workforce_id IS NULL))` to `AgentTrigger.__table_args__` so model-driven and migration-driven schemas agree.

**F3 (significant, follows from F1) — `downgrade()` will error on fresh-install DBs.**
`src/xagent/migrations/versions/20260722_add_workforce_triggers.py` `downgrade()` unconditionally calls `batch.drop_constraint(SINGLE_OWNER_CHECK, type_="check")`. Since every fresh install is stamped to head without the constraint actually present (F1), `alembic downgrade` on such a DB raises `ValueError: No such constraint` (reproduced locally in batch mode).
*Fix:* fixing F1 (constraint present via `create_all`) makes downgrade consistent; otherwise guard the drop behind an existence/reflection check symmetric to the `needs_workforce_column` guard in `upgrade()`.

**F2 (minor, follows from F1) — No test exercises the constraint via the real app-init path.**
The only tests hitting the constraint are the migration's hand-built-schema unit test and the service-layer guard (`test_create_trigger_requires_exactly_one_owner`, which calls `_create_trigger` and never reaches the DB). Nothing does a raw insert against the `init_db`-built test DB, so F1 goes uncaught.
*Fix:* add a raw-INSERT-rejection test against the real `init_db` fixture DB (will fail today, then pass once F1 is fixed).

**F4 (minor, real edge case) — Replay can regress trigger-run state in a narrow retry window.**
`_attach_workforce_task_to_trigger_run` (`src/xagent/web/services/triggers.py:912-969`): `create_workforce_run_record` commits the WorkforceRun+Task, then this function does a second commit setting `TriggerRun.task_id`/status (954-968). The dedup guard in `prepare_trigger_run` (`:1077`, `if not created and run.task_id is not None`) only blocks replay once `task_id` is persisted. If the process dies between the two commits, a retry replays via idempotency key (`record.created=False`) and unconditionally re-overwrites `task.agent_config` and forces `run.status=PENDING` again, regressing any progress since. Narrow but reachable; untested.
*Fix:* make the re-attach idempotent — skip the config overwrite / status reset when the task is already attached, or persist `task_id` in the same commit as the run record.

**F5 — no finding.** The update guard re-validating only when `config` is present is airtight: no post-creation path can reassign ownership or smuggle connector-runtime config onto an existing workforce trigger.

### Tests
Re-ran the changed area in an isolated venv against the new head: `tests/web/api/test_workforce_triggers.py` + `tests/alembic/test_20260722_add_workforce_triggers.py` — all pass (17, up from 14).

### Simplification
Lean already — the update diff is a two-line guard, a single CHECK statement with matching downgrade, and corresponding tests. No over-engineering.

codeacme17:
Thanks @rogercloud — follow-up review addressed in ba08ffe0.

First, a correction to my previous reply: you're right that the frontend-test justification did not hold up. My standalone test crashed the vitest worker because **it over-mocked** — it replaced the whole `@/lib/agent-triggers-api` module rather than the low-level `apiRequest`, which is what the pre-existing `agent-triggers-dialog.test.tsx` does. I misattributed the crash to the dialog's render tree. Corrected below.

**Prior finding 3 (frontend owner prop) — now fixed.** Added two owner-routing cases to the existing `agent-triggers-dialog.test.tsx`, following its `apiRequest`-mock pattern: `owner={{ kind: "workforce", id: 5 }}` hits `/api/workforces/5/triggers` (and asserts it never falls through to `/api/agents/`), and the no-owner path hits `/api/agents/42/triggers`. Full file now 15 tests, 1.69s, no crash.

**F1 (significant) — CHECK absent on fresh installs.** Fixed. Added `CheckConstraint("(agent_id IS NULL) <> (workforce_id IS NULL)", name="ck_agent_triggers_single_owner")` to `AgentTrigger.__table_args__`, so the model-driven (`create_all`) schema and the migration-driven schema agree and both enforce the invariant. Name matches the migration so downgrade can still drop it.

**F3 (significant) — downgrade errors when the constraint is absent.** Fixed. F1 makes it consistent for all post-merge DBs; additionally the downgrade now reflects check-constraint names and only drops when present, so it can't raise "No such constraint".

**F2 (minor) — no test via the real init_db path.** Added `test_single_owner_check_constraint_on_init_db_schema`: a raw INSERT against the `init_db`-built test DB, rejecting both-set and neither-set. (With SQLite FK enforcement off in tests, a missing CHECK would let both inserts succeed — so this genuinely exercises F1.)

**F4 (minor) — replay can regress trigger-run state.** Fixed. `_attach_workforce_task_to_trigger_run` is now idempotent: on a replay it only sets `task_id`/merges config when missing, and never resets a `RUNNING`/`COMPLETED` run back to `PENDING`.

Verified against the rebased head: backend `test_workforce_triggers.py` + `test_20260722_add_workforce_triggers.py` = 17 passing (up from 14); frontend dialog test = 15 passing; mypy / ruff / tsc / eslint clean.

src/xagent/web/models/trigger.py
codeacme17: Fixed in ba08ffe0. Added `CheckConstraint("(agent_id IS NULL) <> (workforce_id IS NULL)", name="ck_agent_triggers_single_owner")` to `AgentTrigger.__table_args__`, so the ORM-built (`create_all`) schema fresh installs use now carries the same constraint as the migration-upgraded schema. Name matches the migration so downgrade still finds it. Covered by a new raw-INSERT test against the `init_db` test DB (`test_single_owner_check_constraint_on_init_db_schema`).

src/xagent/migrations/versions/20260722_add_workforce_triggers.py
codeacme17: Fixed in ba08ffe0. Two parts: F1 (constraint now on the model) makes every post-merge DB carry it, and this drop is now guarded — the downgrade reflects check-constraint names via a new `_check_constraint_names` helper and only calls `drop_constraint` when present, so it can no longer raise "No such constraint" on a DB that lacks it.

src/xagent/web/services/triggers.py
codeacme17: Fixed in ba08ffe0. `_attach_workforce_task_to_trigger_run` is now idempotent on replay: it only writes `task_id` / merges the trigger context when they are missing, and it never resets a run already in `RUNNING`/`COMPLETED` back to `PENDING`. So a retry after a crash between the run-record commit and the attach commit completes the attach without regressing state.

frontend/src/app/workforces/[id]/page-client.tsx
Comment (codeacme17) [reply frontend/src/app/workforces/[id]/page-client.tsx:467]: You are right, and my earlier justification was wrong — thanks for pushing on it. The crash came from my standalone test over-mocking (`@/lib/agent-triggers-api` replaced wholesale) rather than the dialog render tree. Fixed in ba08ffe0 by following this file's existing pattern: two owner-routing cases added to `agent-triggers-dialog.test.tsx` mocking `apiRequest`, asserting the workforce owner hits `/api/workforces/5/triggers` (never `/api/agents/`) and the no-owner path hits `/api/agents/42/triggers`. Full file: 15 tests, 1.69s, no crash.

src/xagent/web/services/triggers.py
```diff
@@ -475,9 +516,67 @@ def create_agent_trigger(
...
+    user_id: int,
+    workforce_id: int,
+    trigger_type: str,
+    name: str | None = None,
+    enabled: bool = True,
+    config: dict[str, Any] | None = None,
+    prompt_template: str | None = None,
+    secret: str | None = None,
+) -> tuple[AgentTrigger, str | None]:
+    """Create a workforce-owned trigger (workforce access checked by caller)."""
```
rogercloud: N2 (low, design observation): execution-identity asymmetry with agent triggers. An agent trigger can only be created by the agent's owner (`get_owned_agent`), but a workforce trigger can be created by anyone with workforce `edit` access — and the trigger persists that creator's `user_id`, which becomes the identity the workforce run later executes as. This is not a privilege-escalation issue: `ensure_workforce_access(action="run")` is re-checked at fire time in `create_workforce_run_record`, so a creator who later loses access is rejected then. But it is a real behavioral difference from the agent-trigger model. Worth a short code comment here making the choice explicit rather than leaving it implicit.

src/xagent/web/services/workforce_runs.py
```diff
@@ -294,7 +320,7 @@ async def create_workforce_run(
             task_owner_user_id=int(user.id),
             # Workforce runs as the requesting user; actor == owner here.
             actor_user_id=int(user.id),
-            payload=TaskTurnPayload(transcript_message=normalized_message),
+            payload=TaskTurnPayload(transcript_message=str(task.description or "")),
```
rogercloud, line 348: N3 (very-low, code quality): this now passes `str(task.description or "")` where it previously used the local `normalized_message`. Behavior-preserving today because `task.description` is set to `normalized_message` at record-creation time, but that equivalence is undocumented. A future change to how `description` is derived/transformed could silently diverge the turn transcript from the original message. Suggest a one-line comment noting the intentional equivalence (`# task.description == normalized_message at creation`).

src/xagent/web/services/triggers.py
```diff
@@ -762,6 +909,79 @@ def _trigger_execution_context(
...
+    if run.task_id is None or int(run.task_id) != task_id:
+        setattr(run, "task_id", task_id)
+        db.add(run)
+    # Arm only a run that has not moved past preparation; a replay must not
+    # regress a RUNNING/COMPLETED run.
+    if str(run.status) not in (
+        TriggerRunStatus.RUNNING.value,
+        TriggerRunStatus.COMPLETED.value,
+    ):
+        if str(run.status) != TriggerRunStatus.PENDING.value:
```
rogercloud: shrink (simplification): the nested `if status not in (RUNNING, COMPLETED): if != PENDING: set PENDING` is two conditions doing the work of one — the inner check only guards against a redundant identical write. Collapse to a single guard:

```python
if str(run.status) not in (
    TriggerRunStatus.RUNNING.value,
    TriggerRunStatus.COMPLETED.value,
    TriggerRunStatus.PENDING.value,
):
    setattr(run, "status", TriggerRunStatus.PENDING.value)
    db.add(run)
```

net: -3 lines.

rogercloud (approved):
## Follow-up review — commit ba08ffe0

**Update summary.** Since my last review at `8889033c`, the author pushed `ba08ffe0` to close all 4 findings from that round. It adds the owner-invariant `CheckConstraint` to the `AgentTrigger` model (so fresh `init_db()` installs get the DB-level guard, not just the migration path), makes the migration `downgrade()` reflect the constraint before dropping it (no longer erroring on schemas that never had it), hardens the workforce-run re-attach against replay (no redundant writes, no regressing a RUNNING/COMPLETED run back to PENDING), and adds the frontend owner-routing tests plus the model/migration test coverage. Net: 4 fixed, 1 new test added to the changed-area suite (18 passed, up from 17).

### Prior findings

- **F1 — owner-invariant CHECK absent from model → FIXED.** `src/xagent/web/models/trigger.py:70-79` now declares `CheckConstraint("(agent_id IS NULL) <> (workforce_id IS NULL)", name="ck_agent_triggers_single_owner")`, name matching the migration exactly. Verified against a fresh `init_db()` schema: the CHECK is present in the generated DDL and raw both-set / neither-set INSERTs are now rejected at the DB layer. New test `test_single_owner_check_constraint_on_init_db_schema` (`tests/web/api/test_workforce_triggers.py:233`) pins it.
- **F3 — migration downgrade errors on fresh-install DBs → FIXED.** New `_check_constraint_names()` helper reflects existing check-constraint names (mirrors the existing `_index_names` pattern, `NotImplementedError` fallback for unsupported dialects); `downgrade()` now drops the constraint only `if has_check`. Verified both paths: downgrade of a `create_all` schema (present) and of a hand-built schema lacking it (skipped, no `ValueError`). *Residual (non-blocking):* no dedicated regression test exercises the constraint-absent downgrade — the existing migration test always runs `upgrade()` first, so the constraint is always present when its downgrade runs.
- **F4 — replay idempotency gap in `_attach_workforce_task_to_trigger_run` → FIXED.** `src/xagent/web/services/triggers.py:959-980` now only rewrites `task.agent_config` when the merged config actually differs, only sets `run.task_id` when missing/different, and never resets a RUNNING/COMPLETED run back to PENDING (advances only non-terminal, non-pending states). Traced the crash-and-retry scenario against the new code — no regression of an already-advanced run, first-time-success path unchanged. *Residual (non-blocking):* no dedicated test simulates the crash-and-retry / RUNNING-preserved path; correctness is established by code trace + existing passing tests, but that specific branch isn't pinned (see N1).
- **Frontend — page-client owner prop test → FIXED, with acknowledgment.** New `describe("AgentTriggersDialog owner routing", ...)` block in `frontend/src/components/build/agent-triggers-dialog.test.tsx` asserts a workforce owner routes to `/api/workforces/{id}/triggers` and an agent owner to `/api/agents/{id}/triggers`. Ran directly: 15/15 pass, ~1.1s, no crash — confirming the earlier "Radix tree crashes vitest" excuse was false. Worth noting: the author directly conceded the earlier justification was wrong ("the crash came from my standalone test over-mocking, not the dialog render tree") — that's good-faith engagement, appreciated. *Residual (non-blocking):* the tests drive the dialog with `owner={{ kind: "workforce", id: 5 }}` directly rather than exercising `page-client.tsx:453`'s own `id ?? workforce.id` fallback expression; landing coverage at the dialog's owner-routing behavior is a reasonable place to put it, but the original fallback expression still isn't directly tested.

### New findings

This PR is in good shape — the items below are all low / very-low, none blocking.

- **N1 (low, test coverage).** No test directly exercises the idempotent re-attach branch added in `ba08ffe0` (the `record.created is False` replay where `run.task_id` is `None` from a prior crash). Existing tests either short-circuit before re-entering the attach logic (`task_id` already set) or never create a `WorkforceRun`. Suggest a test that calls `_attach_workforce_task_to_trigger_run` twice with the same trigger/run to pin the replay behavior (config-skip, status-preserve, task_id-fill-in). Same gap noted under F4.
- **N2 (low, design observation).** Execution-identity asymmetry: agent triggers can only be created by the agent's owner (`get_owned_agent`), but workforce triggers can be created by anyone with workforce `edit` access, and the trigger persists that creator's `user_id` as the identity the workforce run later executes as. Not a privilege-escalation issue — `ensure_workforce_access(action="run")` is re-checked at fire time in `create_workforce_run_record`, so a creator who later loses access is rejected then. But it's a real behavioral difference from the agent-trigger model and worth a conscious note (a code comment) rather than leaving implicit. Inline.
- **N3 (very-low, code quality).** `src/xagent/web/services/workforce_runs.py:323`: the turn-start payload now uses `str(task.description or "")` instead of the local `normalized_message`. Behavior-preserving today (`task.description` is set to `normalized_message` at record creation), but the equivalence is undocumented — a future change to how `description` is derived could silently diverge the turn transcript from the original message. Suggest a one-line comment noting the intentional equivalence. Inline.

- `src/xagent/web/services/triggers.py` L977: `shrink` — nested `if status not in (RUNNING, COMPLETED): if != PENDING: set PENDING` is two conditions doing one job; the inner check just guards a redundant identical write. Collapse to `if str(run.status) not in (RUNNING.value, COMPLETED.value, PENDING.value): set PENDING`.
- net: -3 lines possible.

### Tests

Re-ran the full changed-area suite in an isolated venv against `ba08ffe0`:
- `tests/web/api/test_workforce_triggers.py` + `tests/alembic/test_20260722_add_workforce_triggers.py`: **18 passed** (up from 17).
- `tests/web/api/test_agent_triggers.py`: **40 passed** (confirms the new model-level CHECK doesn't regress the existing agent-trigger path).
- Frontend `agent-triggers-dialog.test.tsx`: **15 passed**, no crash.

All 4 prior findings fixed; new items are minor / nice-to-have. This looks close to mergeable pending only those optional follow-ups.

codeacme17:
Thanks for the approval, @rogercloud 🙏 — knocked out the optional follow-ups in 0e8932fd:

- **N1** — added `test_workforce_reattach_is_idempotent_on_replay`: calls `_attach_workforce_task_to_trigger_run` twice with the same trigger/run, simulating the crash-and-retry window (task_id lost, run already RUNNING). Asserts the same task is re-linked, no duplicate WorkforceRun, config unchanged, and RUNNING is not knocked back to PENDING.
- **N2** — documented the execution-identity difference on `create_workforce_trigger`: the creator's `user_id` is the identity the run executes as, and it's re-checked via `ensure_workforce_access(action="run")` at fire time (not an escalation).
- **N3** — added a comment at the turn-start payload noting the intentional `task.description == normalized_message` equivalence and to keep them in sync.
- **Simplification** — collapsed the nested status guard into a single membership check (`RUNNING`/`COMPLETED`/`PENDING`).

Left the constraint-absent-downgrade regression test out (your F3 residual): after the model-level constraint, no head schema legitimately lacks it, so a test would have to hand-build an artificial schema to hit that branch — happy to add it if you'd prefer it pinned.

`test_workforce_triggers.py` now 19 → and `test_agent_triggers.py` 40 still green; ruff/mypy clean.

src/xagent/web/services/triggers.py
codeacme17: Documented in 0e8932fd — added a docstring note on `create_workforce_trigger` spelling out that the creator's `user_id` is the run's execution identity and that `ensure_workforce_access(action="run")` re-checks at fire time, so it is not an escalation. Thanks for calling out the asymmetry explicitly.

src/xagent/web/services/workforce_runs.py
codeacme17: Fixed in 0e8932fd — added a comment at the payload noting the intentional `task.description == normalize_text(message)` equivalence and to keep the two in sync if description derivation ever changes.

src/xagent/web/services/triggers.py
codeacme17: Done in 0e8932fd — collapsed to a single `if str(run.status) not in (RUNNING, COMPLETED, PENDING): set PENDING`, dropping the redundant inner guard.

PR #959, opened 2026-07-22
feat(models): split OpenAI provider into official OpenAI and OpenAI-Compatible

kejhz653-stack:
Rename the OpenAI provider description and add a dedicated OpenAI-Compatible provider for third-party endpoints that implement the OpenAI API shape. Its Connect flow shows Base URL directly instead of hiding it behind Advanced settings, since the field is required.#958

PR #967, opened 2026-07-22
feat: add workforce shareable link deployment

### LINKED ISSUE #947: feat(workforce): shareable link deployment
## Description

First external channel for Workforce — the thinnest slice validating the full `external entry → create_workforce_run → begin_turn` path. Token-based auth; does not touch `AgentApiKey` / `v1/deps.py`.

Scope:

- Workforce-level `share_enabled` / `share_token` stored in the `deployments` table (from the foundation issue).
- Share management endpoints for workforce (enable / rotate / disable), gated on workforce `status == active` + `share_enabled` (per-channel opt-in, default off — mirroring Agent's pattern).
- `api/share.py` / `public_chat_access.py`: workforce-token resolution path → `create_workforce_run(source="shared_link")`. The existing manager-agent rejection stays; external access enters only via workforce credentials.
- Multi-turn: append messages route to the run's task; the APPEND path must explicitly call `sync_workforce_run_status` after `begin_turn` to flip the run back to `running`.
- Frontend: public share chat page reuse (`public-agent-chat-page.tsx`), share management in the workforce detail page.

## Expected result

- A workforce with share enabled is usable through a public share link end-to-end (multi-turn), producing proper `WorkforceRun` records visible in runs history with `source = shared_link`.
- Archived workforce / disabled share / stale fingerprint all reject cleanly.

Part of #805. Blocked by #946.

codeacme17:
## Summary

- add workforce share-link management backed by deployment records, including enable, rotate, and disable flows
- authenticate public workforce share tokens and create workforce runs with `source=shared_link`
- support multi-turn public conversations while keeping workforce run status synchronized
- add the workforce share dialog and reuse the existing public chat experience
- cover share lifecycle, authentication, task creation, multi-turn behavior, and rejection paths with focused tests

## Testing

- `pytest tests/web/api/test_workforce_share_link.py` — 17 passed
- `pytest tests/web` — 2483 passed, 9 skipped
- backend mypy and Ruff checks passed
- frontend TypeScript and i18n checks passed
- manual frontend E2E verified:
  - generate and rotate a workforce share link
  - open the anonymous public chat page
  - complete a real two-turn model conversation
  - confirm the task records `source=shared_link` and the workforce run completes
  - confirm rotated and disabled links are rejected

Closes #947


frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -159,14 +163,34 @@ function PublicConversationContent({
...
+        for (const file of files) {
+          const formData = new FormData()
+          formData.append("file", file)
+          formData.append("task_type", "task")
+          formData.append("task_id", newTaskId.toString())
+          await fetch(`${getApiUrl()}${publicApiPrefix}/files/upload`, {
+            method: "POST",
+            headers: { "Authorization": `Bearer ${accessToken}` },
+            body: formData,
+          }).catch((err) => console.error(err))
```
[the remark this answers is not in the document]
codeacme17: Fixed in 6cf275e1. I extracted a shared upload helper used by both first-turn workforce uploads and the normal transport path. It now checks response.ok, validates success and file_id, surfaces the backend detail/message, and throws on failure so the outer send flow does not clear the draft files after a failed upload. I also added regression coverage for an HTTP 413 response and successful metadata normalization.

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -323,35 +356,17 @@ export function PublicAgentChatPage({
...
-        })
+        uploadedFiles.push(await uploadPublicChatFile({
+          url: `${getApiUrl()}/${authMode === "share" ? "api/share" : "api/widget"}/files/upload`,
+          accessToken: publicAccessToken,
+          file,
+          taskType: params.taskType,
+          taskId: params.taskId,
+          fallbackError: t("files.uploadFailed"),
+        }))
       }
```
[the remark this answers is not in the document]
codeacme17: Fixed in 5d596db5. The shared public chat transport now returns Promise.all for the upload batch, which uploads files concurrently and preserves the input ordering of the returned metadata.

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -159,14 +173,33 @@ function PublicConversationContent({
...
+        for (const file of files) {
+          await uploadPublicChatFile({
+            url: `${getApiUrl()}${publicApiPrefix}/files/upload`,
+            accessToken,
+            file,
+            taskType: "task",
+            taskId: newTaskId,
+            fallbackError: t("files.uploadFailed"),
+          })
+        }
```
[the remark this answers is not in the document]
codeacme17: Fixed in 5d596db5. The first-turn workforce file uploads now run concurrently with Promise.all while preserving rejection propagation, so the draft is still retained if any upload fails.

src/xagent/web/api/websocket.py
```diff
@@ -115,6 +115,7 @@
 )
 from ..services.uploaded_file_store import UploadedFileStore
 from ..services.workforce_runtime import (
+    is_workforce_task,
```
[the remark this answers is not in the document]
codeacme17: Thanks for flagging this. sync_workforce_run_status is already imported from services.workforce_runtime at websocket.py line 122, and that import predates this PR and is present on the base branch. This PR only adds is_workforce_task to the existing import block, so no additional import change is needed.

src/xagent/web/api/public_chat_access.py
```diff
@@ -435,13 +535,64 @@ async def create_public_chat_task(
...
+    through ``create_workforce_run``: it pins the config snapshot, creates the
+    ``WorkforceRun`` record, and begins the first turn with the guest's
+    message — so ``request.description`` doubles as the opening message.
+    """
+    workforce = access_context.workforce
+    assert workforce is not None
+    if request.agent_id is not None:
+        raise HTTPException(status_code=403, detail="Share link is unavailable")
+
+    result = await create_workforce_run(
```
rogercloud, line 603: **[Major] M1 — First-turn attachments silently dropped for workforce share sessions.**

`_create_workforce_share_chat_task` calls `create_workforce_run(...)` without `selected_file_ids`, even though `create_workforce_run` accepts it (`src/xagent/web/services/workforce_runs.py:166`) and the authenticated owner endpoint threads it through (`src/xagent/web/api/workforces.py:1111`). The frontend uploads first-message files only after the turn has already begun (`frontend/src/components/widget/public-agent-chat-page.tsx:172`), so first-turn attachments are silently dropped for workforce share sessions. This is newly introduced here (the agent-share flow threads files into the first turn) and has no test coverage.

Suggestion: upload files first and pass their ids as `selected_file_ids` into this call, or explicitly document/disable first-message attachments for workforce shares until supported.

src/xagent/web/api/workforces.py
```diff
@@ -852,6 +861,130 @@ async def unpublish_workforce(
...
+    """Rotate the public share link token for an active workforce."""
+    workforce = ensure_workforce_access(
+        db, user, _load_workforce(db, workforce_id), action="edit"
+    )
+    _ensure_shareable_workforce(workforce)
+    with _commit_or_rollback(db):
+        deployment = get_or_create_deployment(
+            db, DeploymentOwnerType.WORKFORCE, int(workforce.id)
+        )
+        cast(Any, deployment).share_enabled = True
```
rogercloud: **[Minor] m1 — `rotate_workforce_share_link` unconditionally sets `share_enabled = True`.**

Rotating a previously-disabled link silently re-enables public access. This is currently unreachable via the shipped UI (rotate is only offered when already enabled), but it is surprising semantics for any direct API caller.

Suggestion: preserve the existing `share_enabled` value on rotate, or document that rotate implies enable.

src/xagent/web/api/workforces.py
```diff
@@ -852,6 +861,130 @@ async def unpublish_workforce(
...
+async def get_workforce_share_link(
+    workforce_id: int,
+    db: Session = Depends(get_db),
+    user: User = Depends(get_current_user),
+) -> WorkforceShareLinkResponse:
+    """Return the current owner-only share link state for a workforce."""
+    workforce = _load_workforce(db, workforce_id)
+    if workforce is None:
+        raise HTTPException(status_code=404, detail="Workforce not found")
+    # The raw token is a credential: gate reads on edit permission (owner /
```
rogercloud, line 915: **[Minor] m2 — `disable_workforce_share_link` rejects archived workforces, inconsistent with `get_workforce_share_link`.**

Disable routes through `ensure_workforce_access(action="edit")`, which 409s on archived workforces (`src/xagent/web/services/workforce_access.py:183`), while `get_workforce_share_link` here uses `can_edit_workforce` directly and explicitly allows reads on archived workforces. The asymmetry is confusing. It is near-moot in practice since `_authenticate_workforce_share_link` already rejects non-active workforces regardless of `share_enabled` (`src/xagent/web/api/share.py:101`).

Suggestion: let disable succeed as an idempotent revoke on archived workforces for symmetry, or add a comment explaining the intentional divergence.

src/xagent/web/services/deployments.py
```diff
@@ -0,0 +1,61 @@
...
+        db.query(Deployment)
+        .filter(
+            Deployment.owner_type == owner_type.value,
+            Deployment.owner_id == int(owner_id),
+        )
+        .first()
+    )
+
+
+def get_or_create_deployment(
```
rogercloud, line 36: **[Minor] m3 — Concurrent enable/rotate can 500 instead of resolving idempotently.**

`get_or_create_deployment` does a check-then-`flush()` insert against the `uq_deployment_owner` unique constraint with no `IntegrityError` handling, and the caller's `_commit_or_rollback` (`src/xagent/web/api/workforces.py:895`) is a bare `except Exception: rollback; raise`. Two near-simultaneous enable/rotate calls (double-click, two tabs) leave the loser with an unhandled `IntegrityError` → 500. Impact is low (owner-only action).

Suggestion: catch `IntegrityError`, roll back, and re-fetch the existing row so the operation resolves idempotently.

src/xagent/web/api/websocket.py
```diff
@@ -4173,6 +4174,24 @@ async def finish_existing_delivery(
                             context=context,
                         )
                         logger.info(f"Task {task_id} started in background")
+                        # ``begin_turn`` flips the Task to RUNNING but does
+                        # not project that onto the WorkforceRun; without an
```
rogercloud, line 4203: **[Minor] m4 — WS running-projection is racy; `db.refresh` implies a stronger guarantee than exists.**

After `begin_turn` (which commits in a genuinely separate DB session, per its own docstring at `task_orchestrator.py:271`) returns, `db.refresh(task)` may not yet observe that commit, so `sync_workforce_run_status` can no-op on a stale terminal status until the next completion re-sync. This is already guarded by `try/except Exception: logger.warning(...)` and is non-fatal by design, so risk is low.

Suggestion: add a one-line comment noting that the running projection is best-effort, not guaranteed-fresh, since `db.refresh` here implies a stronger guarantee than actually holds.

rogercloud (commented):
## Summary

This PR adds a public, token-based share-link channel for Workforces, mirroring the existing Agent share-link pattern. A Workforce owner can enable, rotate, and disable a share link (state persisted in the `deployments` table); anonymous visitors authenticate with the share token to obtain a guest JWT and drive a multi-turn conversation that creates a real `WorkforceRun` tagged `source="shared_link"`. Rejection logic (archived / disabled / rotated) is re-validated consistently at auth time, per-WS-message, and again under lock inside `create_workforce_run`. Adds 586 lines of new backend tests plus new frontend upload-helper tests.

## Design verdict (non-blocking commentary)

**Verdict: sound — acceptable with reservations.** The slice is appropriately thin: it reuses the existing public-chat page and share dependency, and the `deployments.py` helper plus the `ShareChatAccessContext` agent/workforce discriminated union are clean, proportional abstractions. The rejection/re-validation flow is coherent.

Design-level reservations, all non-blocking, none introduced as regressions by this PR:

- **No per-guest isolation within a single workforce share link** (see M2 below). Inherited from the agent-share pattern rather than introduced here, but the workforce case carries substantially more private multi-turn content, so this deserves an explicit product decision rather than silent inheritance. Recommend a follow-up issue.
- **No rate limiting / abuse control on the public auth and task-create endpoints.** Each anonymous request can spawn a full workforce run against the owner's model/billing budget. Inherited from the agent pattern and arguably outside the "thinnest slice" scope, but the blast radius is larger for workforces. Recommend a follow-up issue.
- **`sync_workforce_run_status` now fires for all workforce tasks, not only shared ones.** This broadens existing behavior beyond the share feature's scope. It is non-fatal by design (guarded), so risk is low, but flagging that the change is not scoped strictly to sharing.

## Findings

### Major

**M1 — First-turn attachments silently dropped for workforce share sessions.** `src/xagent/web/api/public_chat_access.py:557` — `_create_workforce_share_chat_task` calls `create_workforce_run(...)` without `selected_file_ids`, even though the function accepts it (`src/xagent/web/services/workforce_runs.py:166`) and the authenticated owner endpoint threads it through (`src/xagent/web/api/workforces.py:1111`). The frontend uploads first-message files only *after* the turn has already begun (`frontend/src/components/widget/public-agent-chat-page.tsx:172`), so first-turn attachments never reach the run. This is newly introduced by this PR (the agent-share flow does thread files into the first turn) and has zero test coverage. Fix: upload files first and pass their ids as `selected_file_ids` into `_create_workforce_share_chat_task` → `create_workforce_run`, or explicitly document/disable first-message attachments for workforce shares until supported.

**M2 — Cross-guest task access within a workforce share link (inherited, pre-existing pattern; follow-up, not a blocker for this PR).** `_get_task_for_workforce_share_context` (`src/xagent/web/api/public_chat_access.py:309`) authorizes solely by `Task.id == task_id` (attacker-controllable sequential integer PK, `src/xagent/models/task.py:76`) plus owner and workforce_id match — with no per-guest claim. Contrast the widget path (`src/xagent/web/api/public_chat_access.py:296`), which additionally requires `agent_config.get("guest_id") == access_context.guest_id`. This is the same architecture already present in the pre-existing agent-share flow (base commit `06e3b456`), which this PR faithfully mirrors per issue #947's explicit "mirror Agent's pattern" scope — so it is not a regression. It remains a real, currently-unaddressed privacy gap affecting both flows and should be tracked as a follow-up: add a per-guest claim (e.g. a `guest_id`/session token) checked in both `get_task_for_share_context` branches.

### Minor

**m1 — `rotate_workforce_share_link` unconditionally sets `share_enabled = True`.** `src/xagent/web/api/workforces.py:963`. Rotating a previously-disabled link silently re-enables public access. Currently unreachable via the shipped UI (`frontend/.../workforce-share-dialog.tsx:127` only offers rotate when already enabled), but this is surprising API semantics for any direct caller. Fix: preserve the existing `share_enabled` value on rotate, or document that rotate implies enable.

**m2 — `disable_workforce_share_link` rejects archived workforces, inconsistent with `get_workforce_share_link`.** Disable routes through `ensure_workforce_access(action="edit")`, which 409s on archived workforces (`src/xagent/web/services/workforce_access.py:183`), while get uses `can_edit_workforce` directly and explicitly allows reads on archived workforces (`src/xagent/web/api/workforces.py:915`). Near-moot functionally, since `_authenticate_workforce_share_link` already rejects non-active workforces regardless of `share_enabled` (`src/xagent/web/api/share.py:101`), but the asymmetry is confusing API design. Fix: let disable succeed (idempotent revoke) on archived workforces for symmetry, or add a comment explaining the intentional divergence.

**m3 — Concurrent enable/rotate can 500 instead of resolving idempotently.** `get_or_create_deployment` (`src/xagent/web/services/deployments.py:35`) does a check-then-`flush()` insert against the `uq_deployment_owner` unique constraint with no `IntegrityError` handling, and `_commit_or_rollback` (`src/xagent/web/api/workforces.py:895`) is a bare `except Exception: rollback; raise`. Two near-simultaneous enable/rotate calls (double-click, two tabs) leave the loser with an unhandled `IntegrityError` → 500. Low impact (owner-only action). Fix: on `IntegrityError`, rollback and re-fetch the existing row.

**m4 — WS running-projection is racy; `db.refresh` implies a stronger guarantee than exists.** `src/xagent/web/api/websocket.py:4178`. After `begin_turn` (which commits in a genuinely separate DB session per its own docstring — `task_orchestrator.py:271`) returns, `db.refresh(task)` may not yet observe that commit, so `sync_workforce_run_status` can no-op on a stale terminal status until the next completion re-sync. Already guarded by `try/except Exception: logger.warning(...)` (non-fatal by design), so risk is low. Fix: add a one-line comment noting the running projection is best-effort, not guaranteed-fresh.

## Test coverage gaps

Not bugs, but worth flagging:
- No test that a *disabled* link invalidates an already-issued guest JWT on a subsequent request (only the rotate-invalidates-JWT case is covered).
- No test for the M1 first-turn-attachment gap.
- No test asserting cross-guest task isolation (M2), either direction.
- The WS "sync failure is non-fatal" except-branch (m4) is untested.
- No concurrency test for enable/rotate (m3).

## Prior findings (gemini-code-assist bot) — re-review recap

1. **File upload silently ignored HTTP errors** (`fetch(...).catch(...)` without checking `response.ok`) — author claimed fixed in `6cf275e1`. **VERIFIED FIXED**: `frontend/src/lib/public-chat-file-upload.ts:44` now checks `!response.ok || data?.success !== true || !fileId` and throws with the backend's error detail; both call sites use the shared helper. Regression tests cover a 413 response and successful metadata normalization.
2. **`sync_workforce_run_status` NameError (not imported) — flagged critical** — author claimed a false positive. **VERIFIED — FALSE POSITIVE**: the import exists at `src/xagent/web/api/websocket.py:113` in both current head and base commit `06e3b456` (this PR only added `is_workforce_task` to that import line); all call sites are in scope. No code change needed.
3. **Sequential file uploads (`for...of`) should use `Promise.all` — two instances** — author claimed fixed in `5d596db5`. **VERIFIED FIXED for both**: the shared-transport path (`public-agent-chat-page.tsx:353`) and the first-turn workforce path (`public-agent-chat-page.tsx:182`) now use `Promise.all(files.map(...))`; returned-metadata ordering is preserved via positional mapping, and rejection propagation still retains the draft on any single failure.

Lean already. A dedicated simplification pass over the full diff found no over-engineering, dead flexibility, or reinventable logic. `public-chat-file-upload.ts` is a genuine dedup, `_commit_or_rollback` and `_merge_agent_config` are proportional non-speculative helpers, and the `ShareChatAccessContext` union is the minimum shape for a real two-kind discriminated union.

## Test results

- `pytest tests/web/api/test_workforce_share_link.py` — 17 passed (verified in the review worktree).
- Full `pytest tests/web` suite — re-run in the review worktree: exit code 0, no failures (a handful of tests skipped due to missing Postgres/Pub-Sub emulator environment, consistent with the author's reported 2483 passed / 9 skipped).
- Backend mypy/Ruff and frontend TypeScript/i18n: not independently re-run; author reports passing.

codeacme17:
Thanks @rogercloud for the thorough review. Addressed the newly-introduced Major and all four minors in `2e0ba457`; the two inherited design reservations are tracked as a follow-up.

**M1 — first-turn attachments dropped** ✅ Fixed. Opening-message files are now uploaded task-lessly (there is no task yet — this relaxation is scoped to the *workforce* share context only; the agent path keeps its `task_id`-required contract, so `test_share_upload_requires_task_id` stays valid) and threaded into `create_workforce_run(selected_file_ids=...)`, so the first turn sees them. New test: `test_workforce_share_first_turn_attachments_reach_run` + `test_agent_share_taskless_upload_still_requires_task_id`.

**m1 — rotate force-enables** ✅ Fixed. `rotate` now only replaces the token and preserves the existing `share_enabled`, so resetting a disabled link can't silently re-expose the workforce. Test: `test_rotate_preserves_disabled_state`.

**m2 — disable rejects archived, inconsistent with get** ✅ Fixed. `disable` is now idempotent and allowed on archived workforces (mirrors the read endpoint's `can_edit_workforce` check) since it only ever removes access. Test: `test_share_link_disable_allowed_for_archived_workforce` (and enable/rotate stay 409 on archive).

**m3 — concurrent enable/rotate 500** ✅ Fixed. `get_or_create_deployment` now catches the `uq_deployment_owner` `IntegrityError`, rolls back, and re-fetches the winner. Test: `test_get_or_create_deployment_recovers_from_insert_race`.

**m4 — WS running-projection racy** ✅ Comment added noting the projection is best-effort, not a guaranteed-fresh read of `begin_turn`'s separate-session commit.

Also closed a couple of the flagged coverage gaps: `test_disabled_workforce_share_invalidates_existing_guest_tokens`.

**M2 (cross-guest task access) + rate limiting** — agreed these are inherited from the agent-share pattern and not regressions here. Filed #973 to track per-guest isolation (a `guest_id` claim checked in both branches of `get_task_for_share_context`) and public-endpoint rate limiting, linked to #805.

src/xagent/web/api/public_chat_access.py
codeacme17: Fixed in 2e0ba457. Opening-message files are now uploaded task-lessly (no task exists yet — the relaxation in `upload_share_chat_files` is scoped to the *workforce* share context; the agent path keeps `task_id` required, so `test_share_upload_requires_task_id` stays valid) and their ids are threaded into `create_workforce_run(selected_file_ids=request.files)`, so the first turn sees them. Frontend now uploads first, then passes the ids in the task-create body. Tests: `test_workforce_share_first_turn_attachments_reach_run`, `test_agent_share_taskless_upload_still_requires_task_id`.

src/xagent/web/api/workforces.py
codeacme17: Fixed in 2e0ba457. `rotate` now only replaces the token and preserves the existing `share_enabled`, so resetting a disabled link cannot silently re-expose the workforce. Test: `test_rotate_preserves_disabled_state`.

src/xagent/web/api/workforces.py
codeacme17: Fixed in 2e0ba457. `disable` now mirrors the read endpoint (`can_edit_workforce`, no archived-edit 409) and succeeds as an idempotent revoke on archived workforces, with a docstring explaining the intentional divergence from enable/rotate (which stay 409 on archive since they re-expose access). Test: `test_share_link_disable_allowed_for_archived_workforce`.

src/xagent/web/services/deployments.py
codeacme17: Fixed in 2e0ba457. `get_or_create_deployment` now catches the `uq_deployment_owner` `IntegrityError`, rolls back, and re-fetches the winner row so concurrent enable/rotate resolves idempotently instead of 500. Test: `test_get_or_create_deployment_recovers_from_insert_race`.

src/xagent/web/api/websocket.py
codeacme17: Done in 2e0ba457 — added a comment at the WS running-projection site noting it is best-effort, not a guaranteed-fresh read of `begin_turn`s separate-session commit (a stale/failed projection only delays the running flip and re-syncs on completion).

src/xagent/web/api/public_chat_access.py
```diff
@@ -340,7 +440,23 @@ async def upload_share_chat_files(
         raise HTTPException(status_code=422, detail="No files provided")
 
     if not task_id:
```
rogercloud, line 453: **R1-1 (major) — Task-less guest uploads are an unbounded storage-abuse surface (introduced by this fix).** Allowing `task_id=None` uploads here (needed so first-turn files can land before the task exists) is now reachable by anyone holding the public share link, with no per-request file-count cap, no owner storage quota (`check_storage_gate` is only called from KB ingestion, never this path), and no GC for orphaned `task_id IS NULL` rows if a guest never completes task creation. This trades the dropped-attachment bug (M1) for a new anonymous-write surface. #973 tracks per-guest isolation + rate limiting; please call out this specific task-less orphan/quota exposure explicitly against #973 (or a dedicated issue). Suggestion: add a per-request file-count cap now (cheap, blocks the worst case), defer quota + orphan-GC to the tracked follow-up. Not a merge blocker given this feature's thinnest-slice scope and the existing pattern of tracking hardening as follow-ups, but worth an explicit decision before shipping.

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -118,6 +129,21 @@ function PublicConversationContent({
 
       setCreateTaskError(null)
 
+      // Workforce shares start their first turn inside task creation, so any
```
rogercloud, line 140: **R1-2 (minor) — First-turn upload orphans if task creation later fails.** Files are uploaded task-lessly here, then `create_workforce_run` validates the message (`normalize_text(..., required=True)`, 400 on empty) at `workforce_runs.py:176` *before* `_bind_selected_files_to_task` runs at line 242 — so a validation or earlier error leaves the already-uploaded files unbound and uncleaned. Same root cause as R1-1. Suggestion: validate the message client-side before uploading, or bind-or-delete on the task-creation failure path.

rogercloud (commented):
## Re-review: feat(workforce): shareable link deployment

Follow-up to my prior review at `5d596db510fd11bc9d9876d8773b1b4c5bc7b15c`. Current head is `c4c5a850f052a22d218798e09206bbdf2aa02d1f`. The author's `2e0ba457` ("fix(workforce): address share-link review") addresses all 5 actionable prior findings; head also merged `main` (PRs #961/#963, both unrelated to share links) — those are **out of scope** for this review.

### Prior findings — verified fix status

All 5 re-verified against current code + tests (independently of the author's claims), each backed by a real, non-tautological test.

- **M1 (major) — first-turn attachments dropped — FIXED.** `_create_workforce_share_chat_task` now passes `selected_file_ids=request.files` into `create_workforce_run`, and `_bind_selected_files_to_task` rebinds task-less (`task_id IS NULL`) uploads before the run is created; the agent-share path still requires `task_id` (unweakened). Covered by `test_workforce_share_first_turn_attachments_reach_run` + `test_agent_share_taskless_upload_still_requires_task_id`.
- **m1 (minor) — rotate force-enabled a disabled link — FIXED.** `rotate_workforce_share_link` now only touches `share_token`/`share_updated_at`, never `share_enabled`. Covered by `test_rotate_preserves_disabled_state`.
- **m2 (minor) — disable inconsistent with get on archived workforces — FIXED.** `disable_workforce_share_link` now uses `can_edit_workforce` (matching the read path) and succeeds on archived workforces; enable/rotate still 409 on archived (intentional, documented). Covered by `test_share_link_disable_allowed_for_archived_workforce`.
- **m3 (minor) — concurrent enable/rotate race → 500 — FIXED.** `get_or_create_deployment` now catches `IntegrityError`, rolls back, and re-fetches the winner. Covered by `test_get_or_create_deployment_recovers_from_insert_race` (drives the real constraint, not a mocked exception).
- **m4 (minor) — WS running-projection comment — FIXED.** Clarifying best-effort comment added at the projection site.

(M2 from the prior review was a pre-existing architectural issue tracked separately, not in scope for this PR — no action expected.)

### Design verdict

**Acceptable-with-reservations** — this is an independent blank-slate re-evaluation of the current code, not a corollary of "all prior findings fixed." The core design remains sound: token-in-deployments mirrors the Agent pattern, entering via `create_workforce_run` reuses the snapshot/run/turn machinery, and per-message WS token revalidation (rotation/disable/archive all invalidate outstanding guest tokens on every request) is the right security posture. The reservation is R1-1 below.

### New findings

#### Major

**R1-1 — Task-less guest uploads are an unbounded storage-abuse surface (introduced by the fix commit).** Fixing M1 required allowing `task_id=None` uploads in the workforce-share context so first-turn files can land before the task exists. That path is now reachable by anyone holding the public share link, with no per-request file-count cap, no owner storage quota (`check_storage_gate` is never called here — only from KB ingestion), and no GC for orphaned `task_id IS NULL` rows if a guest never completes task creation. This trades the dropped-attachment bug for a new anonymous-write surface. The `2e0ba457` message notes per-guest isolation + rate limiting are tracked in #973; the specific task-less orphan/quota exposure should be called out explicitly against #973 (or a dedicated issue). Suggested: add a per-request file-count cap now (cheap, blocks the worst case), defer quota + orphan-GC to the tracked follow-up. Not treated as a merge blocker given the "thinnest slice" scope of this feature and the existing pattern of tracking hardening as follow-ups (e.g. M2 in the prior review), but worth an explicit decision before shipping.

#### Minor

**R1-2 — First-turn upload orphans if task creation later fails.** The frontend uploads task-lessly, then `create_workforce_run` validates the message (`normalize_text(..., required=True)`, 400 on empty) *before* `_bind_selected_files_to_task` runs — so a validation/earlier error leaves already-uploaded files unbound and uncleaned. Same root cause as R1-1. Suggested: validate the message client-side before upload, or bind-or-delete on the failure path.

**R1-3 — `rotate` before any `enable` creates a disabled-but-tokened deployment row.** Calling rotate first materializes a `Deployment` with `share_enabled` at its `False` column default but a live `share_token`. Harmless at rest (`find_enabled_share_deployment` filters `share_enabled=True`); `test_rotate_preserves_disabled_state` only covers rotate-after-disable. Low priority — add a rotate-before-enable test, or document that rotate implicitly requires a prior enable.

**R1-4 — `assert workforce is not None` used for control flow.** Both sites sit behind an `if access_context.workforce is not None` guard at their single call site, so unreachable today, but `python -O` strips asserts (repo confirmed never to run `-O`). Defensive-coding nit: replace with an explicit `if workforce is None: raise HTTPException(403, ...)`. Not blocking.

**R1-6 — WS `db.commit()` after `sync_workforce_run_status` commits the whole session (observation, no action).** Verified no other uncommitted writes are pending on the session at that point, so it's safe today, and it matches the existing pattern in `workforce_runs.py`. Noting only because a future change adding earlier uncommitted work in the same handler would get flushed here inadvertently.

- `shrink: src/xagent/web/services/deployments.py:19` — `new_share_token()` duplicates `_new_share_token()` in `src/xagent/web/api/agents.py:248` (both `secrets.token_urlsafe(24)`); the same one-liner recurs 4x total (`triggers.py:173`, `gmail_provisioning.py:104`). Extract one shared helper.
- `yagni: src/xagent/web/api/workforces.py:895` — `_commit_or_rollback` contextmanager is used 3x while 7+ sites in the same file inline `try/db.commit()/except/db.rollback()`; inline it to match the file's own convention.
- `yagni: frontend/src/lib/public-chat-file-upload.ts:13` — `taskId?: number | string | null` accepts a `string` variant no caller passes; narrow to `number | null`.

Net: ~-10 to -15 lines possible, mostly readability/consistency rather than volume.

### Test results

- `tests/web/api/test_workforce_share_link.py` — 24 tests pass (grew 17 → 24 since `2e0ba457`, covering the 5 fixes).
- Full `tests/web` suite — exit 0, all pass (only environment-gated tests skipped, e.g. missing Postgres/Pub-Sub emulator).
- WS tests (`test_websocket_task_control_state.py`, `test_websocket_preview.py`, added/grown via the merged #961/#963) pass and don't interact badly with the share-link changes.

**Nothing here blocks merge.** All 5 prior findings are genuinely fixed. R1-1 is worth an explicit decision (add a cheap file-count cap, or consciously defer to #973) but is consistent with this feature's already-accepted "thinnest slice, harden later" risk posture; the rest are minors/cleanups.

src/xagent/web/api/workforces.py
```diff
@@ -852,6 +861,143 @@ async def unpublish_workforce(
...
+        cast(Any, deployment).share_updated_at = datetime.now(timezone.utc)
+        if not deployment.share_token:
+            cast(Any, deployment).share_token = new_share_token()
+    return _serialize_workforce_share_link(workforce, deployment)
+
+
+@router.post(
+    "/{workforce_id}/share-link/rotate", response_model=WorkforceShareLinkResponse
+)
+async def rotate_workforce_share_link(
```
rogercloud, line 949: **R1-3 (minor) — `rotate` before any `enable` creates a disabled-but-tokened deployment row.** Calling rotate before enable materializes a `Deployment` via `get_or_create_deployment` with `share_enabled` at its `False` column default (`models/deployment.py:56`) but a live `share_token`. Harmless at rest (`find_enabled_share_deployment` filters `share_enabled=True`), and `test_rotate_preserves_disabled_state` only covers rotate-after-disable. Low priority — add a rotate-before-ever-enabled test, or document that rotate implicitly requires a prior enable.

src/xagent/web/api/public_chat_access.py
```diff
@@ -250,15 +306,61 @@ def get_task_for_public_context(
     return task
 
 
+def _get_task_for_workforce_share_context(
+    db: Session, task_id: int, access_context: ShareChatAccessContext
```
rogercloud, line 316: **R1-4 (minor, non-blocking) — `assert workforce is not None` used for control flow.** Both this site and line 572 sit behind an `if access_context.workforce is not None` guard at their single call site, so unreachable today — but `python -O` strips asserts (repo confirmed never to run `-O`). Defensive-coding nit: replace with an explicit `if workforce is None: raise HTTPException(403, ...)` to stay robust against a future unguarded caller.

src/xagent/web/api/websocket.py
```diff
@@ -4198,6 +4199,30 @@ async def finish_existing_delivery(
...
+                        # that commit yet and the sync can no-op on a stale
+                        # terminal status. That is acceptable — terminal
+                        # statuses re-sync on completion, so a stale/failed
+                        # projection here only delays the "running" flip and
+                        # is non-fatal.
+                        if is_workforce_task(task):
+                            try:
+                                db.refresh(task)
+                                if sync_workforce_run_status(db, task, task.status):
+                                    db.commit()
```
rogercloud, line 4218: **R1-6 (observation, no action) — `db.commit()` after `sync_workforce_run_status` commits the whole pending session.** Verified no other uncommitted writes are pending on this session at this point (checked from session acquisition through this call), so it's safe today, and it matches the existing pattern in `workforce_runs.py:314-316`. Noting only for awareness: a future change adding earlier uncommitted work in the same handler would get flushed here inadvertently.

src/xagent/web/api/public_chat_access.py
codeacme17: Addressed in 8cfdb4b5. Took the recommended split: added a cheap per-request file-count cap now (`MAX_TASKLESS_SHARE_UPLOAD_FILES = 10`) on the task-less branch, returning 422 past the limit (test `test_taskless_share_upload_enforces_file_count_cap`). Owner storage quota (`check_storage_gate` is never called on this path) and orphan GC for `task_id IS NULL` rows are explicitly called out against #973 — see https://github.com/xorbitsai/xagent/issues/973#issuecomment-5056888718 — scoped to cover both the workforce-share and agent-share upload surfaces. Kept within the thinnest-slice posture as you suggested; not treating quota/GC as a merge blocker.

frontend/src/components/widget/public-agent-chat-page.tsx
codeacme17: Fixed in 8cfdb4b5. Went with the client-side guard: `handleSend` now returns early on an empty/whitespace first message for the workforce path, before any upload runs — so a turn that `create_workforce_run` would reject with 400 can never leave orphaned uploads behind. (The deeper bind-or-delete-on-failure cleanup for the general orphan case is folded into the #973 orphan-GC note.)

src/xagent/web/api/workforces.py
codeacme17: Fixed in 8cfdb4b5. Added `test_rotate_before_enable_leaves_link_disabled`: rotating before any enable materializes the row with `share_enabled=False` (column default) and a token, and `/api/share/auth` rejects that tokened-but-disabled link with 404 — so the semantics are now pinned by test, and rotate never implies enable. Complements `test_rotate_preserves_disabled_state` (rotate-after-disable).

src/xagent/web/api/public_chat_access.py
codeacme17: Fixed in 8cfdb4b5. Replaced both `assert workforce is not None` sites (`_get_task_for_workforce_share_context` and `_create_workforce_share_chat_task`) with explicit `if workforce is None: raise HTTPException(403, ...)`, so the share auth boundary fails closed even under `python -O` / a future unguarded caller.

src/xagent/web/api/websocket.py
codeacme17: Thanks — agreed, no action. Leaving the `db.commit()` as-is since it matches the `workforce_runs.py` pattern and no earlier uncommitted writes exist on the session at that point; the awareness note is well taken for any future handler change.

codeacme17:
## Re-review addressed (8cfdb4b5)

Thanks for the thorough independent re-verification. All new findings handled:

- **R1-1 (major)** — added a per-request file-count cap (`MAX_TASKLESS_SHARE_UPLOAD_FILES = 10`) on the task-less share upload path; owner quota + orphan-GC explicitly scoped to #973 ([note](https://github.com/xorbitsai/xagent/issues/973#issuecomment-5056888718)), covering both workforce- and agent-share surfaces. Kept the thinnest-slice posture.
- **R1-2** — empty first-turn messages now guarded client-side before upload, so a 400 cannot orphan uploaded files.
- **R1-3** — pinned rotate-before-enable semantics with a test (row stays `share_enabled=False`; auth 404s).
- **R1-4** — both control-flow `assert`s on the auth boundary replaced with explicit `HTTPException(403)`.
- **R1-6** — no action (matches the `workforce_runs.py` pattern; awareness noted).

- `new_share_token` dedup (4x across `agents.py`/`triggers.py`/`gmail_provisioning.py`) — **deferring**: consolidating those touches auth-critical code outside this slice; better as its own cleanup.
- `_commit_or_rollback` — **keeping**: it was extracted per the first-round review to collapse the three near-identical share mutations; happy to inline later if the file settles on one convention.
- `taskId?: number | string | null` → I left the `string` variant as harmless defensive typing; can narrow if you feel strongly.

New tests: `test_taskless_share_upload_enforces_file_count_cap`, `test_rotate_before_enable_leaves_link_disabled` (share suite now 26).

rogercloud (approved):
## Re-review (round 3) — follow-up

Third review round. Prior (round 2) review was posted against commit `c4c5a850f052a22d218798e09206bbdf2aa02d1f`; current PR head is `c686d476b737d5cee1fe25a50d055d3dfe094df1`. Since the last round the author pushed `8cfdb4b5` ("fix(workforce): address share-link re-review (R1-1..R1-4)"). The head also merged two unrelated PRs from main (#959, #974) — those are out of scope for this review and were not evaluated here.

This pass was a blank-slate re-evaluation of the current code (not just a diff of the fixes), and every factual claim below was independently verified against the code and tests.

### Prior findings — verified fix status

**R1-1 (was major) — task-less upload abuse surface — PARTIALLY FIXED.**
`MAX_TASKLESS_SHARE_UPLOAD_FILES = 10` now caps the task-less upload path in `public_chat_access.py`, returning 422 over the limit, covered by `test_taskless_share_upload_enforces_file_count_cap`. The cap is real and correctly enforced server-side.
Still open (per-request cap raises the bar but does not close the surface): no cross-request rate limiting on `/api/share/files/upload` or `/api/share/auth` (the codebase's only limiter, `trigger_rate_limit.py`, is wired solely into `triggers.py`); no owner-level storage quota on this path (`check_storage_gate` is called only from KB ingestion); and no GC for orphaned `task_id IS NULL` `UploadedFile` rows. A guest can therefore still call the endpoint repeatedly, each call under the 10-file cap, and accumulate unbounded orphaned storage against the owner. **This is explicitly tracked**: the author's own code comments reference #973, and — checked directly — issue #973 ("Harden public share channels: per-guest isolation + rate limiting") already carries a scoping comment (added 2026-07-23, same day as this fix) that names exactly these two residual gaps (owner storage quota, orphan GC) and states they should be handled alongside per-guest isolation + rate limiting, covering both the workforce-share and agent-share upload surfaces.

**R1-2 (was minor) — first-turn upload orphans on task-creation failure — PARTIALLY FIXED.**
The client-side guard `if (workforceId && !message.trim()) return` is correctly placed before the upload, and its `!message.trim()` semantics match the server's `normalize_text(required=True)` validation, so it genuinely closes the "empty message → 400 → orphaned upload" path. It remains partial: files are still uploaded task-lessly before `create_workforce_run` begins, and other failure points between upload and the final commit can still orphan files — `ensure_workforce_access`, `build_workforce_snapshot`, `policy.before_workforce_run`, `normalize_execution_mode`, connector-runtime binding, the task-insert flush, `_bind_selected_files_to_task` raising 404 on a missing file, and the `WorkforceRun` insert `IntegrityError` path. It is also purely client-side and bypassed by a guest calling the API directly. Same underlying gap as R1-1, covered by the same #973 tracking described above.

**R1-3 (was minor) — rotate-before-enable leaves a disabled-but-tokened row — FIXED.**
`test_rotate_before_enable_leaves_link_disabled` asserts `share_enabled=False` with a live token after rotate-before-enable and that `/api/share/auth` 404s for it; `find_enabled_share_deployment` filters on `share_enabled=True` and the column defaults to `False`.

**R1-4 (was minor) — assert used for control flow — FIXED.**
Both `assert workforce is not None` sites are replaced with explicit `if workforce is None: raise HTTPException(403, "Share link is unavailable")`. Both call sites remain pre-guarded (real-caller behavior unchanged) and no other unguarded caller exists.

**R1-6 (was minor) — WS commit scope — no action needed.** Agreed in the prior round; not re-litigated.

### Design verdict (independent re-evaluation)

**Acceptable-with-reservations** (unchanged). The design direction remains correct — deployments-table token storage mirroring the Agent pattern, `create_workforce_run` reuse, fail-closed auth boundaries, and now-solid rotate/disable semantics. The standing reservation is the R1-1/R1-2 gap: the per-request 10-file cap raises the bar but does not close the cross-request storage-abuse surface. This is a consciously-tracked follow-up (#973), not an open loose end.

### New findings this round

**F-6 (minor, low severity) — task-less files bound only by owner identity, not per-guest.** Inline comment below. Real but low-severity, and falls squarely under #973's "per-guest isolation" item — no standalone fix needed.

**F-7 — WS best-effort running-projection: confirmed working as intended, no action.** The stale-read risk is real but non-fatal (terminal statuses re-sync on completion), documented in an inline comment, and the happy path is tested (`test_ws_append_syncs_workforce_run_back_to_running`).

**F-8 — get/disable vs enable/rotate archived-workforce permission asymmetry: confirmed intentional, no action.** `get`/`disable` use `can_edit_workforce` (allowing archived) while `enable`/`rotate` use `ensure_workforce_access` (409 on archived); documented in-code and both directions have test coverage. Working as designed.

### Merge-readiness

**Nothing here blocks merge.** R1-3 and R1-4 are fully fixed. R1-1 and R1-2 are partially fixed, and — verified directly against #973 — their residual gaps (per-request-only cap with no cross-request throttling, no owner storage quota, no orphan GC) are already explicitly named in that tracking issue's latest comment, alongside per-guest isolation (which also covers F-6) and rate limiting, scoped to both the workforce-share and agent-share surfaces. This is exactly the level of specificity a follow-up issue needs to not get lost, and it's already there — so there's no outstanding ask beyond what's already tracked.

### Test results

- `tests/web/api/test_workforce_share_link.py` — grown again since the last round, all passing.
- Full `pytest tests/web` suite — exit code 0, all pass (only environment-gated tests skipped).

PR #952, opened 2026-07-21
feat(workforce): deployment foundation — deployments table, run source & idempotency, archive termination, config fingerprint

codeacme17:
## What

Foundation layer for exposing Workforce through external channels (Share / Widget / REST API / Webhook). No external endpoint is opened by this PR; it provides the shared model, run semantics, and access-control guards the channel PRs (#947–#950) build on.

### 1. Shared `deployments` table

New `Deployment` model (`models/deployment.py`): one row per (`owner_type` ∈ {agent, workforce}, `owner_id`), unique on the pair, holding `widget_enabled` / `allowed_domains` / `widget_key` / `share_enabled` / `share_token` / `share_updated_at`. Workforce adopts it from day one; migrating Agent's existing flattened columns is explicitly out of scope (touches the auth-critical `public_chat_access.py` path).

### 2. `create_workforce_run`: `source` + `idempotency_key`

- `source` param threaded to `Task.source` (previously hardcoded `"internal"`); surfaced in the runs-history serializer so external runs are attributable per channel.
- Optional `idempotency_key` with a unique `(workforce_id, idempotency_key)` index. A replayed key returns the existing run (`created=False`, no new turn) instead of creating a duplicate; the pre-insert lookup is backed by an IntegrityError fallback for the concurrent race.

### 3. Archive terminates active runs

`DELETE /api/workforces/{id}` now calls `terminate_active_workforce_runs`: every in-flight run (`pending`/`running`/`paused`) is marked terminal `cancelled`, and currently RUNNING tasks receive a durable PAUSE through the existing task-command pipeline. `sync_workforce_run_status` treats `cancelled` as non-overwritable so the PAUSE landing later cannot resurrect the run.

### 4. Config fingerprint + turn-entry guard

The run snapshot only freezes prompt-building data; worker execution re-reads live Agent rows per call. `build_workforce_snapshot` now pins a `config_fingerprint` (sha256 over manager/worker execution config: instructions, execution_mode, models, knowledge_bases, skills, tool_categories, assignment instructions, membership). At the shared APPEND turn entry (`_begin_turn_atomic_sync`), workforce tasks are re-checked:

- owning workforce archived (or gone) → turn rejected with `workforce_archived`
- live fingerprint ≠ pinned fingerprint → rejected with `workforce_config_changed` ("start a new session")

[...]

PR #968, opened 2026-07-22
feat: add Workforce REST API and SDK deployment channel

### LINKED ISSUE #949: feat(workforce): REST API/SDK deployment
## Description

REST API / SDK channel for Workforce. The heaviest channel — blocked on the workforce API-key auth model.

Scope (per the design discussion in #805, Q4/Q5):

- Generalize `AgentApiKey` with a nullable `workforce_id` (no new table; no existing unique constraint depends on `agent_id`). `xag_*` key format unchanged.
- Generalize the `_owned_agents_query` family (~7 call sites) to an owned-agents-OR-owned-workforces branch; API Keys page supports workforce keys.
- `v1/deps.py`: resolve workforce keys (currently 401s anything resolving to a manager agent — keep that; workforce keys resolve to the workforce, not the manager).
- `POST /v1/workforces/{id}/runs` for run creation; polling / steps / append reuse the existing `/v1/chat/tasks/{task_id}` endpoints as-is (`WorkforceRun.task_id` is a 1:1 unique binding). Requires `source="sdk"` threading from the foundation issue (the task-query scope filter checks `Task.source == "sdk"`).
- Usage metering: count external calls only (run creation + append); worker delegations excluded — consistent with existing agent-as-tool behavior, and LLM cost is already captured by the quota/token tracking system.
- Update `xagent-sdk` (Python) with workforce support + docs/snippets in the deploy dialog.

## Expected result

- A workforce API key can create runs and drive multi-turn conversations through `/v1`, with key usage stats visible on the API Keys page.

Part of #805. Blocked by #946.

codeacme17:
## Summary

Add a REST API and Python SDK deployment channel for published Workforces.

Closes #949.

## What changed

- Add `POST /v1/workforces/{workforce_id}/runs`.
- Support Workforce-bound API keys with owner isolation.
- Reuse `/v1/chat/tasks/{task_id}` for status polling, steps, and follow-up messages.
- Support idempotent run creation and API usage metering.
- Add a Workforce Deploy dialog with cURL and Python SDK examples.
- Support inline API key creation with one-time secret reveal.
- Display Workforce ownership on the API Keys page.
- Add the database migration and API schemas required for Workforce keys.
- Add integration tests covering authentication, isolation, idempotency, polling, and usage tracking.

## Validation

- `tests/web/api/v1/test_workforces.py`: 12/12 passed
- `tests/web/api/test_agent_api_keys.py`: passed
- Frontend TypeScript type-check: passed
- API snippet tests: 18/18 passed
- Manual browser E2E:
  - created and published a Workforce
  - created a Workforce API key
  - created a real run and received `202 Accepted`
  - verified idempotent replay reused the same run
  - polled the task from `running` to `completed`
  - received the expected output: `E2E-949-OK`
  - verified steps and usage metering
  - revoked the temporary key and confirmed subsequent requests returned `401`
  - confirmed the completed run and output in the Runs UI


frontend/src/components/workforce/deploy-workforce-dialog.tsx
```diff
@@ -0,0 +1,269 @@
...
+      setRevealedKey(created.full_key)
+      setNewLabel("")
+      const rows = await listAgentApiKeys({ workforceId })
+      setKeys(rows)
+    } catch {
+      toast.error(t("apiKeysPage.messages.createFailed"))
+    } finally {
+      setCreating(false)
+    }
+  }
```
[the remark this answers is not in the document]
codeacme17: Fixed in 25f273dc. The key-list refresh now has its own error boundary, so a successful key creation keeps the one-shot secret visible and reports only the secondary load failure if the refresh fails.

I also added regression coverage for the create-success/refresh-failure scenario. The related 19 frontend tests, TypeScript type-check, and targeted ESLint checks all pass. Thanks for catching this!

src/xagent/web/api/v1/workforces.py
```diff
@@ -0,0 +1,191 @@
...
+idempotency all behave identically across channels. External metering
+counts only run creation (worker delegations are excluded, consistent
+with agent-as-tool behavior; LLM cost is captured by the quota system).
+"""
+
+import logging
+from typing import NoReturn, Tuple
+
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy.orm import Session, selectinload
```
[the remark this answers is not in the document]
codeacme17: Addressed in 77fa5289. I imported joinedload alongside Session and selectinload to support eager loading of the scalar relationships. Thanks!

src/xagent/web/api/v1/workforces.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    return (
+        db.query(Workforce)
+        .options(
+            selectinload(Workforce.manager_agent),
+            selectinload(Workforce.workers).selectinload(WorkforceAgent.agent),
+        )
+        .filter(Workforce.id == workforce_id)
+        .first()
+    )
```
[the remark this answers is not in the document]
codeacme17: Addressed in 77fa5289. Workforce.manager_agent now uses joinedload, while the workers collection keeps selectinload and chains joinedload for WorkforceAgent.agent. This preserves the collection-friendly loading strategy while avoiding additional round trips for the many-to-one relationships. The 12 targeted Workforce API integration tests pass.

src/xagent/web/api/v1/workforces.py
```diff
@@ -0,0 +1,191 @@
...
+        raise V1ApiError(V1ErrorCode.WORKFORCE_NOT_FOUND, 404)
+
+    workforce = _load_workforce_for_run(db, workforce_id)
+    if workforce is None:
+        raise V1ApiError(V1ErrorCode.WORKFORCE_NOT_FOUND, 404)
+
+    # ``create_workforce_run`` acts as the workforce owner. Resolve the
+    # user from the workforce's owner_user_id (the key carries no user
+    # session of its own).
+    owner = db.query(User).filter(User.id == workforce.owner_user_id).first()
```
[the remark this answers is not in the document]
codeacme17: Addressed in 77fa5289. The Workforce owner is now fetched with db.get(User, int(workforce.owner_user_id)), while preserving the existing missing-owner error handling. The targeted integration tests and Ruff checks pass. Thanks!

src/xagent/web/api/v1/workforces.py
```diff
@@ -0,0 +1,191 @@
...
+    clients switch on ``body.error.code``. Detail-string matching is
+    used only to split the several 400/409 cases the service overloads
+    onto one status code.
+    """
+    status = exc.status_code
+    detail = str(exc.detail or "")
+    lowered = detail.lower()
+
+    if status == 404:
+        raise V1ApiError(V1ErrorCode.WORKFORCE_NOT_FOUND, 404) from exc
```
rogercloud: This 404 handling collapses every 404-status `HTTPException` into `WORKFORCE_NOT_FOUND`, but `_bind_selected_files_to_task` (`src/xagent/web/services/workforce_runs.py:141`) raises `HTTPException(404, "Selected file not found")` for an invalid `message.files` id. A client passing a bad file id therefore gets `error.code = "workforce_not_found"` with a workforce-not-found message — wrong and confusing for SDK clients switching on `error.code`. Unlike the 400/409 cases here, the 404 branch does no detail differentiation. Consider distinguishing this case (e.g. a distinct `FILE_NOT_FOUND` code, or keying off the exception's origin rather than its status alone). No test currently covers this path.

rogercloud (commented):
This PR generalizes `AgentApiKey` into a shared credential that can be bound to either an agent or a workforce (nullable `agent_id`/`workforce_id`, with exactly-one-owner enforced at both the schema and service layers), then adds a REST/SDK deployment channel for workforces: `POST /v1/workforces/{id}/runs` for run creation and reuse of `/v1/chat/tasks/{task_id}` for polling and append, scoped via `source="sdk"`. Run creation is made idempotent through a unique `(workforce_id, idempotency_key)` index with IntegrityError-catch-and-replay, and usage metering is applied to external calls only (creation + append), guarded against double-billing on idempotent replay. A frontend Deploy dialog plus SDK snippets round it out. Closes #949, part of #805, building on #952.

**Approach: sound.** The `ApiKeyPrincipal` unification for auth resolution and the `_resolve_task_or_404` branch-by-principal-type reuse of the existing polling endpoint are clean and avoid net-new surface; the idempotency and metering scoping are race-safe and correct.

## Update since last review

All 4 prior findings were re-verified against the current HEAD (`77fa5289`) and are confirmed fixed:

1. **Key-refresh masking successful creation** (`deploy-workforce-dialog.tsx`) — fixed in `25f273dc`; the list-refresh now has its own nested try-catch, so a refresh failure no longer clears the revealed secret or fires a misleading "createFailed" toast. New `deploy-workforce-dialog.test.tsx` covers it and passes.
2. **Missing `joinedload` import** (`workforces.py`) — fixed in `77fa5289`.
3. **`selectinload` → `joinedload` for the many-to-one `manager_agent`** (`workforces.py:50-58`) — fixed; the `selectinload` on the `workers` collection is correctly retained to avoid join fan-out.
4. **`db.query(User).filter(...).first()` → `db.get(User, ...)`** (`workforces.py:146`) — fixed; verified behavior-equivalent (the original query had no filters beyond the primary key).

## Findings

### [Major / design] Error mapping relies on free-text substring matching

`_raise_v1_for_workforce_http_error` (`src/xagent/web/api/v1/workforces.py:61-97`) distinguishes error cases by lowercasing `HTTPException.detail` prose and substring-matching `"idempotency"`, `"archived"`, and `"must be active"` to select the `IDEMPOTENCY_CONFLICT` / `WORKFORCE_ARCHIVED` / `WORKFORCE_NOT_ACTIVE` codes. The underlying exceptions (`workforce_snapshot.py:83,85`, `workforce_runs.py:106-109`) are bare `HTTPException`s carrying only free-text `detail` — there is no stable type or code to key off. Rewording any of those messages would silently reclassify the response into the generic `INVALID_INPUT` / `TASK_BUSY` fallback, which SDK clients switch on.

This is not how the codebase handles the analogous problem elsewhere: `raise_for_turn_rejection` (`src/xagent/web/api/v1/errors.py:196-221`) switches on a stable typed `reason` via dict lookup rather than prose. Notably, `errors.py`'s own docstring documents `detail`/`message` as "free text that may be updated for clarity without breaking clients" — directly contradicting using it as a discriminant here. Compounding the fragility, no test in `test_workforces.py` exercises the archived / not-active / idempotency-conflict branches through this endpoint while asserting `error.code`, so an accidental reclassification would go unnoticed.

**Suggestion:** give the service-layer exceptions a stable discriminant (a custom `HTTPException` subclass or an explicit error-code attribute) and switch on that, following the `raise_for_turn_rejection` pattern already in this codebase. This is non-blocking but worth a fast-follow before more clients depend on these codes.

### [Minor] Bad file id maps to a misleading `workforce_not_found`

See inline comment on `workforces.py:75`. A bad `message.files` id raises a 404 in `_bind_selected_files_to_task`, which the endpoint funnels into `WORKFORCE_NOT_FOUND` (all 404s collapse to one code, unlike the differentiated 400/409 cases), giving SDK clients a wrong/confusing `error.code`.

### [Minor] Test-coverage gaps on the error branches

`tests/web/api/v1/test_workforces.py` has solid coverage for happy-path, auth/isolation, idempotency-replay, usage metering, and append — but nothing for: archived-workforce creation (409 `WORKFORCE_ARCHIVED`), draft/not-active workforce (409 `WORKFORCE_NOT_ACTIVE`), config-drift on append (409 `WORKFORCE_CONFIG_CHANGED`), idempotency-key reuse against a deleted task (409 `IDEMPOTENCY_CONFLICT`), and invalid `execution_mode` (422). These are precisely the branches guarded by the string-matching above — the least-tested code is also the most breakage-prone. Adding these tests would also serve as a regression guard for the major finding.

- `frontend/src/app/workforces/[id]/page-client.tsx:370` — `disabled={!!isArchived}` on the Deploy button is dead: the button only renders inside the `workforce.status === "active"` branch where `isArchived` is always `false`. Remove it.
- `src/xagent/web/schemas/v1.py:392` — `CreateWorkforceRunRequest.metadata` is accepted but never read anywhere in the endpoint or service layer (confirmed via grep, zero consumers). Drop it until a caller needs it, or wire it through.

_net: -2 lines possible._

## Test results

- Backend: `tests/web/api/v1/test_workforces.py` (12/12) + `tests/web/api/test_agent_api_keys.py` (44/44) — 56/56 passed.
- Frontend: the regression test for the key-refresh fix passed via vitest.

## Recommendation

**Approve with follow-up.** The design is sound and the core paths are well-tested; all prior findings are fixed. The three new findings are 1 major-design + 2 minor and none of them block merge — the major one is a robustness/maintainability concern rather than a current bug. Recommend merging and addressing the error-mapping discriminant plus the missing error-branch tests as a fast-follow.

frontend/src/app/workforces/[id]/page-client.tsx
```diff
@@ -360,6 +362,17 @@ export default function WorkforceDetailPage() {
                             {t("workforces.actions.archive")}
                         </Button>
                     )}
+                    {workforce.status === "active" && (
+                        <Button
+                            variant="outline"
+                            size="sm"
+                            onClick={() => setDeployOpen(true)}
+                            disabled={!!isArchived}
```
Comment (rogercloud) [inline frontend/src/app/workforces/[id]/page-client.tsx]: `disabled={!!isArchived}` is dead code here: this button only renders inside the `workforce.status === "active"` branch, where `isArchived` is always `false`. Safe to remove the prop.

src/xagent/web/schemas/v1.py
```diff
@@ -330,6 +358,84 @@ class AppendMessageResponse(BaseModel):
...
+    )
+    idempotency_key: Optional[str] = Field(
+        default=None,
+        max_length=128,
+        description=(
+            "Caller-supplied dedup token. A retry with the same key returns "
+            "the original run rather than creating a new one."
+        ),
+    )
+    metadata: Optional[Dict[str, Any]] = Field(
```
rogercloud: `metadata` is accepted by the request schema but never read anywhere in `workforces.py` or the service layer (confirmed via grep — zero consumers). Suggest dropping it until a caller actually needs it, or wiring it through to persistence/correlation now. Shipping an unread field risks clients relying on a no-op.

codeacme17:
Thanks Roger — all three findings + both simplifications addressed in `24cc652a`.

### [Major] Error mapping now switches on a stable discriminant
Replaced the free-text substring matching with a machine code. New `WorkforceRunError(HTTPException)` (in `services/workforce_errors.py`) carries a stable `code`; the run service raises it at the archived / not-active / idempotency-conflict / invalid-execution-mode / invalid-idempotency-key / file-not-found sites, and `_raise_v1_for_workforce_http_error` now switches on `exc.code` via a dict — same shape as `raise_for_turn_rejection`. Rewording a `detail` string can no longer silently reclassify a response. `WorkforceRunError` subclasses `HTTPException`, so the JWT web channel keeps surfacing it unchanged through FastAPI's default handler; plain codeless `HTTPException`s (e.g. `ensure_workforce_access`) still fall back to the conservative status-based mapping.

### [Minor] Bad file id no longer maps to `workforce_not_found`
`_bind_selected_files_to_task` now raises `WorkforceRunError(..., code=FILE_NOT_FOUND)`, and I added a distinct `FILE_NOT_FOUND` v1 code (404). A bad `message.files` id now returns `error.code = "file_not_found"` instead of the misleading `workforce_not_found`.

### [Minor] Error-branch test coverage
Added `test_workforces.py` cases asserting the exact `error.code` for: archived (409 `workforce_archived`), draft/not-active (409 `workforce_not_active`), invalid `execution_mode` (422 `invalid_input`), bad file id (404 `file_not_found`), idempotency-key reuse against a deleted task (409 `idempotency_conflict`), and append turn-rejection covering `workforce_config_changed` / `workforce_archived` / `workforce_run_not_found` / `busy`. These now guard the mapping against accidental reclassification.

### Simplifications
- Removed the dead `disabled={!!isArchived}` on the Deploy button.
- Dropped the unread `CreateWorkforceRunRequest.metadata` field.

Verification: `tests/web/api/v1/` + `test_agent_api_keys.py` (195) and the workforce service/foundation/API suites all pass; ruff + mypy clean on the changed files; frontend `type-check` + eslint clean. Thanks for the thorough review!

src/xagent/web/api/v1/workforces.py
codeacme17: Fixed in `24cc652a`. `_bind_selected_files_to_task` now raises `WorkforceRunError(code=FILE_NOT_FOUND)` and the endpoint maps it to a distinct `file_not_found` (404) instead of collapsing into `workforce_not_found`. Added `test_bad_file_id_returns_file_not_found_not_workforce_not_found` to cover it.

frontend/src/app/workforces/[id]/page-client.tsx
Comment (codeacme17) [reply frontend/src/app/workforces/[id]/page-client.tsx]: Fixed in `24cc652a` — removed the dead `disabled={!!isArchived}` prop.

src/xagent/web/schemas/v1.py
codeacme17: Fixed in `24cc652a` — dropped the unread `metadata` field from `CreateWorkforceRunRequest`.

rogercloud (approved):
Re-reviewed at commit `24cc652a`, which replaced the fragile free-text `HTTPException.detail` substring matching with a typed error-code mechanism and added the previously-missing error-branch tests.

## Verification of prior findings

| # | Prior finding | Status | Justification |
|---|---------------|--------|---------------|
| 1 (major) | Error-code mapping via fragile substring matching on `HTTPException.detail`, untested branches | **FIXED** | New `src/xagent/web/services/workforce_errors.py` adds `WorkforceRunError(HTTPException)` with a mandatory typed `.code`; `_raise_v1_for_workforce_http_error` now dispatches via `getattr(exc, "code")` + a `_SERVICE_CODE_TO_V1` dict lookup. `.detail` only feeds the human-readable `message`, never code selection — renaming a service message no longer changes the wire code. |
| 2 (minor) | Bad file id incorrectly mapped to `workforce_not_found` | **FIXED** | `_bind_selected_files_to_task` raises `WorkforceRunError(404, FILE_NOT_FOUND)`, mapped to a distinct `V1ErrorCode.FILE_NOT_FOUND`; the codeless 404→`workforce_not_found` fallback is never reached because `.code` takes precedence. |
| 3 (minor) | Missing coverage for several error branches | **FIXED** | All 6 scenarios now have dedicated tests in `tests/web/api/v1/test_workforces.py` (archived, draft/not-active, config-drift-on-append, idempotency-conflict-vs-deleted-task, invalid execution_mode, bad file id), each asserting the exact `error.code`. |
| 4 (simplification) | Dead `disabled` prop on Deploy button | **FIXED** | Clean single-line removal in `page-client.tsx`; nearby `disabled={saving || !!isArchived}` on publish/unpublish is legitimate and left intact. |
| 5 (simplification) | Unused `metadata` field on `CreateWorkforceRunRequest` | **FIXED** | Removed; repo-wide grep confirms no remaining references (backend, frontend, tests, SDK snippet generators). |

## Fresh pass on the full diff

A blank-slate re-review of the whole PR (not just the delta) leaves the design verdict **sound**. The new `workforce_errors.py` mechanism is clean and non-speculative: service-layer codes are decoupled from the `/v1` wire vocabulary, so other channels aren't forced into it, and it correctly generalizes the codebase's existing typed-discriminant pattern (`raise_for_turn_rejection`) rather than introducing a parallel one. A fresh simplification-lens scan of the full diff — including the new error handling, the `ApiKeyPrincipal`/`deps.py` auth code, and the `AgentApiKeyService` dual-owner refactor — found **zero new issues**. Two sub-minor observations (FYI, not action items): invalid execution mode returns 400 on the JWT web channel vs 422 on the SDK/v1 channel (intentional per-channel adapter behavior), and `WorkforceRunErrorCode` is a plain string-constants class rather than a `str, Enum` like `V1ErrorCode` (minor style inconsistency; both work correctly as dict keys).

## Test results

- `tests/web/api/v1/test_workforces.py` — 18/18 passing
- `tests/web/api/test_agent_api_keys.py` — 44/44 passing

## Recommendation

**Approve.** All previously identified issues are resolved and independently verified, and no new issues were found. Ready to merge.

rogercloud:
please resolve the conflicts

PR #972, opened 2026-07-23
fix: optimize trigger switch state submission logic

### LINKED ISSUE #929: feat: Auto-save Trigger enable/disable toggle changes
**`Deadline: July 23, 2026`**

### Description

Improve the Trigger enable/disable experience by automatically applying toggle changes without requiring users to click "Save Changes".

yiboyasss:
## Summary

- The overview switch no longer auto-navigates into the trigger config view — it just flips or creates the trigger in place.
- The detail switch persists enabled/disabled immediately; the Save button is removed — pending field edits now commit on Done, Back, switching pills, Add, and dismissing the dialog.
- New trigger forms default to disabled, matching the overview switch's default state.
- Per-pill delete confirmation moved to an inline popover on the pill itself instead of a footer button.
- Fix: dismissing the dialog (Escape / overlay click / header X) no longer silently drops unsaved edits or an unrevealed one-time webhook secret — the secret alert now requires an explicit dismiss before the dialog is allowed to close.
- Fix: reversing the Gmail quick-toggle intent (turning the switch back off before picking an account) no longer leaves `formDirty` stuck, which previously blocked Done from closing.
- Fix: a failed enable/disable toggle no longer rolls back onto whatever form the user has since navigated to (stale-closure race between an in-flight request and navigation).
- Fix: closing the dialog after committing an edit no longer issues a redundant second trigger-list refetch.
- Extracted `disableAgentTriggersOfType`/`mergeUpdatedTriggers` helpers shared by the builder summary cards and the triggers dialog.

## Test plan

- [x] `vitest run` on `agent-triggers-dialog.test.tsx` and `agent-builder-triggers.test.tsx` — all passing
- [x] `tsc --noEmit` clean
- [x] `eslint` clean on changed files (no new warnings/errors)
- [x] Manually verified in the browser preview: quick-toggle create/disable, detail switch immediate persist, delete via pill popover, Done/Back/Escape commit paths, Gmail no-account flow, webhook secret reveal + dismiss gating close

Closes #929

### DIFF
--- a/frontend/src/components/build/agent-builder.tsx
+++ b/frontend/src/components/build/agent-builder.tsx
@@ -57,7 +57,9 @@ import {
   StagedTrigger,
   createAgentTrigger,
   createStagedTriggers,
+  disableAgentTriggersOfType,
   listAgentTriggers,
+  mergeUpdatedTriggers,
   stagedToCreatePayload,
   stagedToPseudoTrigger,
 } from "@/lib/agent-triggers-api"
@@ -425,6 +427,40 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
   const dismissWebhookSecrets = () => {
     setCreatedWebhookSecrets([])
     reconcileDeferredNavigation(failedStagedTriggers, [])
+  }
+
+  // Builder card switch: toggling it off disables every enabled trigger of
+  // that type in place — no dialog. The card disappears afterwards because
+  // the summary section only lists types with enabled > 0. A Set (not a
+  // single value) so two types disabling concurrently keep their own guard.
+  const [disablingTriggerTypes, setDisablingTriggerTypes] = useState<ReadonlySet<AgentTriggerType>>(
+    new Set(),
+  )
+  const disableTriggerType = async (type: AgentTriggerType) => {
+    if (!localAgentId) {
+      setStagedTriggers((prev) =>
+        prev.map((item) => (item.type === type ? { ...item, enabled: false } : item)),
+      )
+      toast.success(t("triggers.messages.disabled"))
+      return
+    }
+    setDisablingTriggerTypes((prev) => new Set(prev).add(type))
+    try {
+      const updated = await disableAgentTriggersOfType(localAgentId, triggerSummary, type)
+      setTriggerSummary((prev) => mergeUpdatedTriggers(prev, updated))
+      toast.success(t("triggers.messages.disabled"))
+    } catch (error) {
+      console.error(error)
+      toast.error(error instanceof Error ? error.message : t("triggers.messages.saveFailed"))
+      // A partial failure may have disabled some triggers; resync the summary.
+      void refreshTriggerSummary()
+    } finally {
+      setDisablingTriggerTypes((prev) => {
+        const next = new Set(prev)
+        next.delete(type)
+        return next
+      })
+    }
   }
 
   // During creation the agent has no server-side triggers yet; the summary
@@ -2438,7 +2474,14 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
                           </div>
                         </div>
                       </button>
-                      <Switch checked onCheckedChange={openDialog} className="scale-75" />
+                      <Switch
+                        checked
+                        disabled={disablingTriggerTypes.has(item.type)}
+                        onCheckedChange={(checked) => {
+                          if (!checked) void disableTriggerType(item.type)
+                        }}
+                        className="scale-75"
+                      />
                       <Button
                         type="button"
                         variant="ghost"
@@ -2857,10 +2900,11 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
         open={isTriggersDialogOpen}
         onOpenChange={(dialogOpen) => {
           setIsTriggersDialogOpen(dialogOpen)
-          if (!dialogOpen) {
-            setTriggerDialogInitialType(null)
-            void refreshTriggerSummary()
-          }
+          // No refetch here: every mutation inside the dialog already calls
+          // onChanged (refreshTriggerSummary) as it happens, so closing the
+          // dialog has nothing left to resync — doing it anyway doubled the
+          // request for every Done/dismiss that just committed an edit.
+          if (!dialogOpen) setTriggerDialogInitialType(null)
         }}
         onChanged={refreshTriggerSummary}
         initialType={triggerDialogInitialType}
--- a/frontend/src/components/build/agent-triggers-dialog.test.tsx
+++ b/frontend/src/components/build/agent-triggers-dialog.test.tsx
@@ -44,7 +44,7 @@ vi.mock("@/lib/clipboard", () => ({
 }))
 
 import { AgentTriggersDialog } from "./agent-triggers-dialog"
-import type { StagedTrigger } from "@/lib/agent-triggers-api"
+import type { AgentTrigger, StagedTrigger } from "@/lib/agent-triggers-api"
 
 function jsonResponse(body: unknown, init?: ResponseInit): Response {
   return new Response(JSON.stringify(body), {
@@ -54,25 +54,15 @@ function jsonResponse(body: unknown, init?: ResponseInit): Response {
   })
 }
 
-const GMAIL_ACCOUNTS_URL = "http://api.local/api/cloud/accounts?provider=gmail"
-
-describe("AgentTriggersDialog", () => {
-  let gmailAccounts: Array<{ id: number; provider: string; email: string | null }>
-
-  const baseTrigger9 = {
-    id: 9,
+function makeTrigger(overrides: Partial<AgentTrigger> & { id: number }): AgentTrigger {
+  return {
     user_id: 1,
     agent_id: 42,
-    type: "gmail" as const,
-    name: "Support inbox",
+    type: "webhook",
+    name: "Trigger",
     enabled: true,
-    config: {
-      watch_label: "INBOX",
-      sender_filter: "boss@company.com",
-      subject_keyword: "urgent",
-      oauth_account_id: 7,
-    },
-    prompt_template: "Reply to {{payload}}",
+    config: {},
+    prompt_template: null,
     webhook_token: null,
     webhook_secret: null,
     next_run_at: null,
@@ -80,7 +70,27 @@ describe("AgentTriggersDialog", () => {
     last_error: null,
     created_at: null,
     updated_at: null,
+    ...overrides,
   }
+}
+
+const GMAIL_ACCOUNTS_URL = "http://api.local/api/cloud/accounts?provider=gmail"
+
+describe("AgentTriggersDialog", () => {
+  let gmailAccounts: Array<{ id: number; provider: string; email: string | null }>
+
+  const baseTrigger9 = makeTrigger({
+    id: 9,
+    type: "gmail",
+    name: "Support inbox",
+    config: {
+      watch_label: "INBOX",
+      sender_filter: "boss@company.com",
+      subject_keyword: "urgent",
+      oauth_account_id: 7,
+    },
+    prompt_template: "Reply to {{payload}}",
+  })
 
   beforeEach(() => {
     apiRequestMock.mockReset()
@@ -352,23 +362,14 @@ describe("AgentTriggersDialog", () => {
       }
       if (url === "http://api.local/api/agents/42/triggers" && options?.method === "POST") {
         return Promise.resolve(
-          jsonResponse({
-            id: 11,
-            user_id: 1,
-            agent_id: 42,
-            type: "webhook",
-            name: "API / Webhook",
-            enabled: true,
-            config: {},
-            prompt_template: null,
-            webhook_token: "tok",
-            webhook_secret: "wh_secret_once",
-            next_run_at: null,
-            last_run_at: null,
-            last_error: null,
-            created_at: null,
-            updated_at: null,
-          }),
+          jsonResponse(
+            makeTrigger({
+              id: 11,
+              name: "API / Webhook",
+              webhook_token: "tok",
+              webhook_secret: "wh_secret_once",
+            }),
+          ),
         )
       }
       return Promise.resolve(jsonResponse([]))
@@ -454,23 +455,14 @@ describe("AgentTriggersDialog", () => {
       }
       if (url === "http://api.local/api/agents/42/triggers" && options?.method === "POST") {
         return Promise.resolve(
-          jsonResponse({
-            id: 12,
-            user_id: 1,
-            agent_id: 42,
-            type: "webhook",
-            name: "API / Webhook",
-            enabled: true,
-            config: {},
-            prompt_template: null,
-            webhook_token: "tok",
-            webhook_secret: "wh_escape_secret",
-            next_run_at: null,
-            last_run_at: null,
-            last_error: null,
-            created_at: null,
-            updated_at: null,
-          }),
+          jsonResponse(
+            makeTrigger({
+              id: 12,
+              name: "API / Webhook",
+              webhook_token: "tok",
+              webhook_secret: "wh_escape_secret",
+            }),
+          ),
         )
       }
       return Promise.resolve(jsonResponse([]))
@@ -556,40 +548,8 @@ describe("AgentTriggersDialog", () => {
       if (url === TRIGGERS_URL) {
         return Promise.resolve(
           jsonResponse([
-            {
-              id: 20,
-              user_id: 1,
-              agent_id: 42,
-              type: "webhook",
-              name: "Backup hook",
-              enabled: true,
-              config: {},
-              prompt_template: null,
-              webhook_token: null,
-              webhook_secret: null,
-              next_run_at: null,
-              last_run_at: null,
-              last_error: null,
-              created_at: null,
-              updated_at: null,
-            },
-            {
-              id: 21,
-              user_id: 1,
-              agent_id: 42,
-              type: "webhook",
-              name: "Primary hook",
-              enabled: true,
-              config: {},
-              prompt_template: null,
-              webhook_token: null,
-              webhook_secret: null,
-              next_run_at: null,
-              last_run_at: null,
-              last_error: null,
-              created_at: null,
-              updated_at: null,
-            },
+            makeTrigger({ id: 20, name: "Backup hook" }),
+            makeTrigger({ id: 21, name: "Primary hook" }),
           ]),
         )
       }
@@ -647,40 +607,8 @@ describe("AgentTriggersDialog", () => {
   it("resyncs the trigger list when a batch disable partially fails", async () => {
     const TRIGGERS_URL = "http://api.local/api/agents/42/triggers"
     let triggers = [
-      {
-        id: 30,
-        user_id: 1,
-        agent_id: 42,
-        type: "webhook",
-        name: "Hook A",
-        enabled: true,
-        config: {},
-        prompt_template: null,
-        webhook_token: null,
-        webhook_secret: null,
-        next_run_at: null,
-        last_run_at: null,
-        last_error: null,
-        created_at: null,
-        updated_at: null,
-      },
-      {
-        id: 31,
-        user_id: 1,
-        agent_id: 42,
-        type: "webhook",
-        name: "Hook B",
-        enabled: true,
-        config: {},
-        prompt_template: null,
-        webhook_token: null,
-        webhook_secret: null,
-        next_run_at: null,
-        last_run_at: null,
-        last_error: null,
-        created_at: null,
-        updated_at: null,
-      },
+      makeTrigger({ id: 30, name: "Hook A" }),
+      makeTrigger({ id: 31, name: "Hook B" }),
     ]
     let getCallsAfterFailure = 0
     let patchAttempted = false
@@ -730,40 +658,8 @@ describe("AgentTriggersDialog", () => {
   it("keeps each overview switch's busy guard independent across two types toggled back-to-back", async () => {
     const TRIGGERS_URL = "http://api.local/api/agents/42/triggers"
     const triggers = [
-      {
-        id: 40,
-        user_id: 1,
-        agent_id: 42,
-        type: "webhook",
-        name: "Hook",
-        enabled: true,
-        config: {},
-        prompt_template: null,
-        webhook_token: null,
-        webhook_secret: null,
-        next_run_at: null,
-        last_run_at: null,
-        last_error: null,
-        created_at: null,
-        updated_at: null,
-      },
-      {
-        id: 41,
-        user_id: 1,
-        agent_id: 42,
-        type: "scheduled",
-        name: "Schedule",
-        enabled: true,
-        config: { interval_seconds: 3600 },
-        prompt_template: null,
-        webhook_token: null,
-        webhook_secret: null,
-        next_run_at: null,
-        last_run_at: null,
-        last_error: null,
-        created_at: null,
-        updated_at: null,
-      },
+      makeTrigger({ id: 40, name: "Hook" }),
+      makeTrigger({ id: 41, type: "scheduled", name: "Schedule", config: { interval_seconds: 3600 } }),
     ]
 
     let resolveWebhookPatch: ((value: Response) => void) | undefined
--- a/frontend/src/components/build/agent-triggers-dialog.test.tsx
+++ b/frontend/src/components/build/agent-triggers-dialog.test.tsx
@@ -59,46 +59,51 @@ const GMAIL_ACCOUNTS_URL = "http://api.local/api/cloud/accounts?provider=gmail"
 describe("AgentTriggersDialog", () => {
   let gmailAccounts: Array<{ id: number; provider: string; email: string | null }>
 
+  const baseTrigger9 = {
+    id: 9,
+    user_id: 1,
+    agent_id: 42,
+    type: "gmail" as const,
+    name: "Support inbox",
+    enabled: true,
+    config: {
+      watch_label: "INBOX",
+      sender_filter: "boss@company.com",
+      subject_keyword: "urgent",
+      oauth_account_id: 7,
+    },
+    prompt_template: "Reply to {{payload}}",
+    webhook_token: null,
+    webhook_secret: null,
+    next_run_at: null,
+    last_run_at: null,
+    last_error: null,
+    created_at: null,
+    updated_at: null,
+  }
+
   beforeEach(() => {
     apiRequestMock.mockReset()
     routerPushMock.mockReset()
     gmailAccounts = [
       { id: 7, provider: "gmail", email: "gerard.santos@gmail.com" },
     ]
-    apiRequestMock.mockImplementation((url: string) => {
+    apiRequestMock.mockImplementation((url: string, init?: { method?: string; body?: string }) => {
       if (url === GMAIL_ACCOUNTS_URL) {
         return Promise.resolve(jsonResponse(gmailAccounts))
       }
       if (url === "http://api.local/api/agents/42/triggers") {
-        return Promise.resolve(
-          jsonResponse([
-            {
-              id: 9,
-              user_id: 1,
-              agent_id: 42,
-              type: "gmail",
-              name: "Support inbox",
-              enabled: true,
-              config: {
-                watch_label: "INBOX",
-                sender_filter: "boss@company.com",
-                subject_keyword: "urgent",
-                oauth_account_id: 7,
-              },
-              prompt_template: "Reply to {{payload}}",
-              webhook_token: null,
-              webhook_secret: null,
-              next_run_at: null,
-              last_run_at: null,
-              last_error: null,
-              created_at: null,
-              updated_at: null,
-            },
-          ]),
-        )
+        return Promise.resolve(jsonResponse([baseTrigger9]))
       }
       if (url === "http://api.local/api/agents/42/triggers/9/runs") {
         return Promise.resolve(jsonResponse([]))
+      }
+      if (url === "http://api.local/api/agents/42/triggers/9" && init?.method === "PATCH") {
+        // Echo the base trigger merged with the PATCH body, like a real
+        // backend would — a bare `[]` fallback here would make `updated`
+        // shapeless for any code that reads fields off the response.
+        const patch = init.body ? JSON.parse(init.body) : {}
+        return Promise.resolve(jsonResponse({ ...baseTrigger9, ...patch }))
       }
       return Promise.resolve(jsonResponse([]))
     })
@@ -296,6 +301,48 @@ describe("AgentTriggersDialog", () => {
       )
     })
     expect(detailSwitch).toHaveAttribute("aria-checked", "false")
+  })
+
+  it("reconciles the detail switch from the PATCH response, not just the requested value", async () => {
+    apiRequestMock.mockImplementation((url: string, init?: { method?: string }) => {
+      if (url === GMAIL_ACCOUNTS_URL) return Promise.resolve(jsonResponse(gmailAccounts))
+      if (url === "http://api.local/api/agents/42/triggers") {
+        return Promise.resolve(jsonResponse([baseTrigger9]))
+      }
+      if (url === "http://api.local/api/agents/42/triggers/9/runs") {
+        return Promise.resolve(jsonResponse([]))
+      }
+      if (url === "http://api.local/api/agents/42/triggers/9" && init?.method === "PATCH") {
+        // A backend that (hypothetically) overrides the requested value —
+        // the switch must reflect this, not the optimistic `checked` it was
+        // set to before the request resolved.
+        return Promise.resolve(jsonResponse({ ...baseTrigger9, enabled: true }))
+      }
+      return Promise.resolve(jsonResponse([]))
+    })
+
+    render(
+      <AgentTriggersDialog
+        agentId={42}
+        open
+        onOpenChange={vi.fn()}
+        gmailConnection={{ isConnected: true, connectedAccount: null }}
+      />,
+    )
+
+    fireEvent.click(await screen.findByText("triggers.cards.gmail.title"))
+    await screen.findByLabelText("triggers.form.watchLabel")
+
+    const [detailSwitch] = screen.getAllByRole("switch")
+    expect(detailSwitch).toHaveAttribute("aria-checked", "true")
+    fireEvent.click(detailSwitch)
+
+    // Optimistic: flips to false immediately.
+    expect(detailSwitch).toHaveAttribute("aria-checked", "false")
+    // Reconciled: the response said `enabled: true`, so it flips back.
+    await waitFor(() => {
+      expect(detailSwitch).toHaveAttribute("aria-checked", "true")
+    })
   })
 
   it("shows the one-time webhook secret on the overview after a quick-toggle create", async () => {
@@ -677,6 +724,105 @@ describe("AgentTriggersDialog", () => {
     })
     await waitFor(() => {
       expect(getCallsAfterFailure).toBeGreaterThan(0)
+    })
+  })
+
+  it("keeps each overview switch's busy guard independent across two types toggled back-to-back", async () => {
+    const TRIGGERS_URL = "http://api.local/api/agents/42/triggers"
+    const triggers = [
+      {
+        id: 40,
+        user_id: 1,
+        agent_id: 42,
+        type: "webhook",
+        name: "Hook",
+        enabled: true,
+        config: {},
+        prompt_template: null,
+        webhook_token: null,
+        webhook_secret: null,
+        next_run_at: null,
+        last_run_at: null,
+        last_error: null,
+        created_at: null,
+        updated_at: null,
+      },
+      {
+        id: 41,
+        user_id: 1,
+        agent_id: 42,
+        type: "scheduled",
+        name: "Schedule",
+        enabled: true,
+        config: { interval_seconds: 3600 },
+        prompt_template: null,
+        webhook_token: null,
+        webhook_secret: null,
+        next_run_at: null,
+        last_run_at: null,
+        last_error: null,
+        created_at: null,
+        updated_at: null,
+      },
+    ]
+
+    let resolveWebhookPatch: ((value: Response) => void) | undefined
+    const webhookPatchPromise = new Promise<Response>((resolve) => {
+      resolveWebhookPatch = resolve
+    })
+    let resolveScheduledPatch: ((value: Response) => void) | undefined
+    const scheduledPatchPromise = new Promise<Response>((resolve) => {
+      resolveScheduledPatch = resolve
+    })
+
+    apiRequestMock.mockImplementation((url: string, init?: { method?: string }) => {
+      if (url === GMAIL_ACCOUNTS_URL) return Promise.resolve(jsonResponse([]))
+      if (url === TRIGGERS_URL && (!init?.method || init.method === "GET")) {
+        return Promise.resolve(jsonResponse(triggers))
+      }
+      if (url === `${TRIGGERS_URL}/40` && init?.method === "PATCH") return webhookPatchPromise
+      if (url === `${TRIGGERS_URL}/41` && init?.method === "PATCH") return scheduledPatchPromise
+      return Promise.resolve(jsonResponse([]))
+    })
+
+    render(
+      <AgentTriggersDialog
+        agentId={42}
+        open
+        onOpenChange={vi.fn()}
+        gmailConnection={{ isConnected: false, connectedAccount: null }}
+      />,
+    )
+
+    // TRIGGER_TYPES order is webhook, scheduled, gmail.
+    const [webhookSwitch, scheduledSwitch] = await screen.findAllByRole("switch")
+    expect(webhookSwitch).toHaveAttribute("aria-checked", "true")
+    expect(scheduledSwitch).toHaveAttribute("aria-checked", "true")
+
+    fireEvent.click(webhookSwitch)
+    await waitFor(() => {
+      expect(webhookSwitch).toBeDisabled()
+    })
+    expect(scheduledSwitch).not.toBeDisabled()
+
+    fireEvent.click(scheduledSwitch)
+    await waitFor(() => {
+      expect(scheduledSwitch).toBeDisabled()
+    })
+
+    // Resolve the scheduled toggle first. A scalar busy-guard would have
+    // cleared entirely here and wrongly re-enabled webhook's switch while
+    // its own PATCH was still in flight — the Set-based guard keeps them
+    // independent.
+    resolveScheduledPatch?.(jsonResponse({ ...triggers[1], enabled: false }))
+    await waitFor(() => {
+      expect(scheduledSwitch).not.toBeDisabled()
+    })
+    expect(webhookSwitch).toBeDisabled()
+
+    resolveWebhookPatch?.(jsonResponse({ ...triggers[0], enabled: false }))
+    await waitFor(() => {
+      expect(webhookSwitch).not.toBeDisabled()
     })
   })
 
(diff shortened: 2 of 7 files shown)


frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -582,47 +651,126 @@ export function AgentTriggersDialog({
...
+      )
+      notifyChanged()
+      toast.success(checked ? t("triggers.messages.enabled") : t("triggers.messages.disabled"))
+      return
+    }
+
+    if (!isValidAgentId(agentId)) return
+    setBusy(true)
+    try {
+      await updateAgentTrigger(agentId, selectedTrigger.id, { enabled: checked })
```
rogercloud: **minor — stale server fields after detail-view toggle**

This branch discards the return value of `await updateAgentTrigger(...)` and patches local state with `{ ...item, enabled: checked }` a few lines below. The overview path (`handleTypeToggle` enable-primary, line 656) correctly uses the response object instead. Since `AgentTrigger` carries server-computed fields (`next_run_at`, `last_run_at`, `updated_at`) that a scheduled trigger's enable/disable can change server-side, toggling from the detail view leaves those fields stale until the next full refresh — inconsistent with the overview path.

Suggestion: patch local state from the awaited response object instead of hand-setting `enabled`.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -582,47 +651,126 @@ export function AgentTriggersDialog({
...
+  const handleDetailToggle = async (checked: boolean) => {
+    setFormValue("enabled", checked)
     if (!canOperate) return
+
+    // Identity of the form this toggle started on. If the user navigates
+    // away (Back/pill switch/Add) before an in-flight request settles, the
+    // failure handlers below check this before touching `form` — otherwise a
+    // late rejection would silently mutate whatever unrelated form is now on
+    // screen.
+    const formKeyAtStart = syncedFormKeyRef.current
```
rogercloud, line 731: **minor — untested toggle-rollback race**

This is the PR's most novel fix: capturing `formKeyAtStart` so a *failed* enable/disable no longer rolls back onto whatever form the user navigated to meanwhile (guards at lines 710 and 743). Neither test file simulates a rejected toggle combined with navigating away before it resolves, so this core race-condition fix currently has zero regression coverage.

Suggestion: add a test that rejects the toggle request, navigates to a different trigger before it resolves, and asserts the failed toggle doesn't roll back onto the new selection.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -582,47 +651,126 @@ export function AgentTriggersDialog({
...
       }
-      await loadTriggers(preferredId)
       notifyChanged()
       toast.success(checked ? t("triggers.messages.enabled") : t("triggers.messages.disabled"))
     } catch (err) {
       console.error(err)
       toast.error(err instanceof Error ? err.message : t("triggers.messages.saveFailed"))
+      // A batch disable is not atomic: some triggers may already be disabled
+      // server-side, so resync rather than trusting the local list.
+      if (isValidAgentId(agentId)) void loadTriggers(selectedTriggerIdRef.current)
```
rogercloud: **minor — untested non-atomic bulk-disable resync**

This resync after a partially-failed `Promise.all` bulk disable (see the comment on `disableAgentTriggersOfType`) has no test where one of the batched `updateAgentTrigger` calls rejects to verify the resync actually fires. Same gap exists in `agent-builder.tsx`'s `disableTriggerType` catch (line 425).

Suggestion: add a test where one batched call rejects and assert `loadTriggers`/`refreshTriggerSummary` runs.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -396,6 +398,40 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
+    setDisablingTriggerTypes((prev) => new Set(prev).add(type))
+    try {
+      const updated = await disableAgentTriggersOfType(localAgentId, triggerSummary, type)
+      setTriggerSummary((prev) => mergeUpdatedTriggers(prev, updated))
+      toast.success(t("triggers.messages.disabled"))
+    } catch (error) {
+      console.error(error)
+      toast.error(error instanceof Error ? error.message : t("triggers.messages.saveFailed"))
+      // A partial failure may have disabled some triggers; resync the summary.
+      void refreshTriggerSummary()
```
rogercloud, line 463: **minor — untested non-atomic bulk-disable resync (summary-card path)**

Same gap as `agent-triggers-dialog.tsx:677`: no test rejects one of the batched `disableAgentTriggersOfType` calls to verify `refreshTriggerSummary()` actually fires on partial failure.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -791,44 +1017,64 @@ export function AgentTriggersDialog({
...
   }
 
+  // Dismissal (Esc, overlay click, header X) commits pending edits like every
+  // other exit path. A failed commit is surfaced via its toast but does not
+  // trap the user — dismissal still closes rather than forcing them to fix
+  // an invalid form just to leave. A freshly generated webhook secret is the
+  // one thing dismissal must not drop: unlike a form validation error, it
+  // already exists server-side and is unrecoverable once unseen, so — like
+  // handleDone — closing waits until it has been shown.
+  const handleDismiss = (nextOpen: boolean) => {
```
rogercloud, line 1063: **minor — untested dismiss-with-failing-commit path**

`handleDone` (line 870) keeps the dialog open when `commitPendingEdits()` fails, while `handleDismiss` here only checks `secretPending` and always closes even when the commit failed — this divergence is intentional and documented in the comment above (Escape/overlay/X must never trap the user on an invalid form, only an un-revealed one-time secret blocks close), so it is **not a bug**. But only a passing-commit Escape-dismiss test and a secret-reveal-blocks-Escape test exist; nothing exercises Escape/overlay/X while a pending edit *fails* to commit.

Suggestion: add a test asserting the dialog still closes (dropping the edit) on a failing non-secret commit during dismiss, to lock in the documented behavior.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -384,50 +426,78 @@ export function AgentTriggersDialog({
...
     }
-  }, [activeType, loadRuns, open, selectedTrigger])
+  }, [activeType, loadRunsFor, open, selectedTrigger])
 
+  // Navigation helpers set the form synchronously (no flash of stale values)
+  // and stamp syncedFormKeyRef so the form-sync effect treats the target as
+  // already synced. secretReveal deliberately survives navigation: it is a
+  // one-time value the user must copy, so only closing the dialog or deleting
+  // a trigger clears it.
   const openType = (type: AgentTriggerType) => {
```
rogercloud, line 483: **minor — cross-navigation persistence of the one-time secret reveal (UX note, not a defect)**

This PR intentionally removed `setSecretReveal(null)` from `openType`/`beginEdit`/`beginCreateForType` (present in the base commit, removed here), so a one-time webhook secret reveal now persists across navigation to unrelated trigger types/detail views until explicit dismiss/delete/close. This matches the documented intent (comments near line 447 and 861) — the one-time secret must never be silently lost — so it's working as designed.

Light suggestion only: consider scoping the reveal banner to the trigger/type that generated it to reduce confusion, while keeping the "can't close without dismissing it" safety guarantee.

frontend/src/lib/agent-triggers-api.ts
```diff
@@ -247,6 +247,32 @@ export async function updateAgentTrigger(
...
 }
 
+// Disables every enabled trigger of one type. Shared by the builder summary
+// card and the triggers dialog so both switches keep identical semantics.
+// Returns the updated triggers so callers can patch local state without a
+// full list reload. Promise.all is not atomic: on rejection some triggers may
+// already be disabled server-side, so callers should resync in their catch.
+export async function disableAgentTriggersOfType(
+  agentId: number | string,
+  triggers: Pick<AgentTrigger, "id" | "type" | "enabled">[],
```
rogercloud: **simplification (yagni)**: `triggers` is typed as `Pick<AgentTrigger, "id" | "type" | "enabled">[]`, but both call sites (`agent-builder.tsx:418`, `agent-triggers-dialog.tsx:667`) pass full `AgentTrigger[]`. No caller needs the narrower shape — restore `AgentTrigger[]`.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -702,16 +911,24 @@ export function AgentTriggersDialog({
...
+      const remaining = staged.triggers.filter((item) => item.clientId !== trigger.id)
+      staged.onChange(remaining)
+      // Deleting a pill other than the selected one must not reset the form
+      // being edited; when the selected one goes, fall back to the next
+      // trigger of the type (newest first), mirroring the live flow.
+      if (selectedTriggerIdRef.current === trigger.id) {
+        const next = remaining
+          .map(stagedToPseudoTrigger)
+          .filter((item) => item.type === trigger.type)
+          .sort(newestFirst)[0]
```
rogercloud: **simplification (shrink)**: this "pick next trigger of same type, newest first, after removal" logic is duplicated below (line ~949) in the live branch, identical apart from the `.map(stagedToPseudoTrigger)` mapping. Extract `pickNextAfterDelete(list: AgentTrigger[], type: AgentTriggerType): number | null` and call it from both branches. net: -15 lines possible (combined with the `agent-triggers-api.ts` type fix).

rogercloud (commented):
## Summary

This PR reworks the agent triggers dialog so trigger enable/disable toggles auto-save (closing issue #929, P0): the overview switch now flips or creates a trigger in place without forcing the user into the config view, and the detail-view enabled switch persists immediately (the explicit "Save Changes" button is gone). Pending field edits now commit implicitly on every exit path — Done, Back, pill-switch, Add, and dialog dismissal. New triggers default to disabled, delete confirmation moves to an inline pill popover, and `disableAgentTriggersOfType` / `mergeUpdatedTriggers` are extracted for reuse between `agent-triggers-dialog.tsx` and `agent-builder.tsx`. The change also carries several race / stale-closure fixes around dismissal, the Gmail quick-toggle intent, and toggle rollback.

**Verdict: acceptable-with-reservations.** The core fix is correct and the P0 behavior is delivered; the reservations are design-complexity and test-coverage, not correctness blockers.

## Design-level findings (no single line owns these)

**D1 — Implicit "commit on every exit" contract (design concern, not a blocker).** The strongest part of this PR is the state-identity model: `syncedFormKeyRef` keyed by `type:id` correctly prevents a list refresh's object-reference churn from wiping unsaved edits, and it's the most important fix here. However, this PR *replaces* the base's simple explicit-Save-button model with an implicit "commit pending edits on every exit path" contract. That persistence-timing behavior is now spread across `formDirty`, `syncedFormKeyRef`, `formMatchesEmptyIgnoringEnabled`, stale-closure reasoning in `secretPending`, four refs, and five separate `commitPendingEdits()` call sites (Done / Back / pill-switch / Add / dismiss). This is net-new complexity introduced by this PR, not inherited ref-heavy style. It's a legitimate, traceable implementation of the required auto-save behavior — not a wrong-direction rewrite — so it does not block merge. A reducer or explicit state machine (`overview` / `editing(id)` / `creating(type)`) would consolidate this timing contract more cleanly and is worth considering as a follow-up.

**D2 — `formMatchesEmptyIgnoringEnabled` (passing note).** This helper (lines 100-108, single call site at line 704) exists only so the Gmail quick-toggle's synthetic dirty preset can be silently un-dirtied if reversed before submit. This is a reasonable, well-isolated reuse of the existing create-form/validation machinery for a genuinely constrained UX case (you can't auto-enable Gmail without picking an account). No action needed.

- `frontend/src/lib/agent-triggers-api.ts` L257: `disableAgentTriggersOfType`'s `triggers` param is typed as `Pick<AgentTrigger, "id" | "type" | "enabled">[]`, but both call sites (`agent-builder.tsx:418`, `agent-triggers-dialog.tsx:667`) pass full `AgentTrigger[]` — restore `AgentTrigger[]`.
- `frontend/src/components/build/agent-triggers-dialog.tsx` L925-931 / L946-951: `handleDelete`'s staged and live branches duplicate identical "pick next trigger of same type, newest first, after removal" logic (byte-identical apart from the staged branch's `.map(stagedToPseudoTrigger)`). Extract `pickNextAfterDelete(list: AgentTrigger[], type: AgentTriggerType): number | null`, called with `remaining.map(stagedToPseudoTrigger)` (staged) and `remaining` (live).

net: -15 lines possible

## No issues found (verified)

- i18n: `en.ts` / `zh.ts` additions and removals are symmetric — no dead or missing keys.
- Removed exports/props: `Trash2` import removed cleanly; `disableAgentTriggersOfType` / `mergeUpdatedTriggers` are correctly shared by both call sites — no broken callers.
- Core form-sync identity-key logic (list refresh preserving unsaved edits) is correct and covered by a staging round-trip test.
- Concurrent multi-type disable guard uses a `Set` correctly, so two cards disabling concurrently don't clobber each other.

Out of scope (aside, not counted as a finding): the overview quick-toggle "create" path calls `notifyChanged()` twice (once inside `createDefaultTrigger`, once in the caller), causing a duplicate refetch — but this exact double-call pre-exists in the base commit in both branches and is neither introduced nor worsened here.

---
*Testing note: tests were not re-run in this review pass — this is a static review. Test files were read directly to confirm the coverage gaps below but vitest was not executed.*

rogercloud (commented):
## Re-review — update commit `82964b9` (on top of `11c1038`)

Scope: follow-up pass on the changes pushed since the prior review. Assumes the reader has the original review in hand — this only covers what moved.

### What changed in `82964b9`

The update commit is a focused pass at the earlier findings, not a redesign:

- **Detail-toggle now trusts the server.** `handleDetailToggle`'s existing-trigger branch (`agent-triggers-dialog.tsx:740-747`) captures the `updateAgentTrigger(...)` return value and patches `liveTriggers` from it, instead of hand-setting `enabled`. A new comment notes `next_run_at`/`last_run_at` can shift server-side — this now matches the overview path.
- **Toggle-rollback race is fenced off at the UI.** A new test proves that while a detail toggle is in flight, the Back button, sibling trigger switches, and pill navigation are all disabled, and rollback on rejection lands on the still-selected trigger.
- **Partial batch-disable resync is covered on both sides.** Two new tests exercise the resync/`refreshTriggerSummary` catch blocks (`agent-triggers-dialog.tsx:675`, `agent-builder.tsx:418`).
- **Dismiss-with-failing-commit is locked in.** A new Escape test forces validation failure and asserts the dialog still closes, the error toast fires, and the invalid edit is never persisted.
- **`disableAgentTriggersOfType`'s param type widened back** to `AgentTrigger[]` (`agent-triggers-api.ts:257`), matching both callers.
- **`pickNextAfterDelete` extracted** (`agent-triggers-dialog.tsx:~209-215`) and called from both the staged (`:934`) and live (`:952`) delete branches, removing the duplicated "pick next after delete" logic.

### Design verdict (current state): acceptable-with-reservations

Independent read of the code as it stands at `82964b9`, not a deduction from "the findings got fixed." The direction is sound — removing explicit Save, persisting immediately, and sharing helpers addresses the root cause rather than papering over it. The reservation is unchanged: the implicit "commit on every exit path" contract is coordinated across a sizable state surface (`formDirty` / `syncedFormKeyRef` / `selectedTriggerIdRef` / `activeTypeRef` / `busy` / `busyType` / `deleteConfirmId` / `secretReveal` / `formKeyAtStart`). Most of that is the inherent cost of "no Save + immediate persistence + preserve unsaved edits," but the ref-mirroring + key-stamping technique is still a moderately high bar for a future maintainer. Not a blocker.

### Prior findings — status

| ID | Status | Reason |
|----|--------|--------|
| F3 (detail-toggle discarded server response) | **FIXED** | Now patches `liveTriggers` from the `updateAgentTrigger` return value (`agent-triggers-dialog.tsx:740-747`). No dedicated regression test, but the fix is correct and complete. |
| F4 (toggle-rollback race untested) | **PARTIAL** | New test proves the race is unreachable via the UI (navigation disabled in-flight) — good defense-in-depth, arguably better than a targeted race test. But the `formKeyAtStart !== syncedFormKeyRef.current` guard branch itself remains formally uncovered; it's now effectively dead unless a future change re-enables navigation during the window. |
| F5 (bulk-disable resync untested — dialog) | **FIXED** | New "resyncs the trigger list when a batch disable partially fails" test; executed and passing, stack trace hits `agent-triggers-dialog.tsx:675`. |
| F5 (bulk-disable resync untested — builder) | **FIXED** | New "resyncs the summary via refreshTriggerSummary…" test; executed and passing, hits `agent-builder.tsx:418`. |
| F2/F6 (dismiss-with-failing-commit untested) | **FIXED** | New Escape test covers exactly the flagged gap; underlying behavior/comment byte-for-byte unchanged (no regression, just coverage). |
| F7 (one-time secret reveal persists across navigation) | **NOT ADDRESSED (acceptable)** | `secretReveal` still unscoped; this was a light UX suggestion, not a required fix. Fine to leave unless the author wants it. |
| S1 (`Pick<...>` over-generalized) | **FIXED** | Param typed `AgentTrigger[]` (`agent-triggers-api.ts:257`). |
| S2 (duplicated delete-selection logic) | **FIXED** | `pickNextAfterDelete` extracted and shared across both branches. |

- **Extract a `makeTrigger(overrides)` test fixture factory.** Several new cases in `frontend/src/components/build/agent-triggers-dialog.test.tsx` and `agent-builder-triggers.test.tsx` inline a full ~15-line trigger object that differs only in id/name/webhook_token (dialog ids 20/21/30/31, builder ids 9/10). A shared factory collapses each block to a one-line call. Production-side simplifications (S1/S2) are already done and lean, so this is test-only.

  net: -60 lines possible

---

*Note: unlike the first-pass static consolidation, this pass ran the tests directly — multiple subagents executed `npx vitest run` (including `-t "partially fails"`) and confirmed the suite passing, 36/36, with stack traces validating the exact catch blocks under test.*

*Closing aside (not a finding): the "Missing Description or aria-describedby for DialogContent" a11y warning in test output was reproduced identically on the pre-PR base commit — pre-existing and unrelated to this PR, so it's not flagged.*

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -210,6 +240,11 @@ export function AgentTriggersDialog({
   const [busy, setBusy] = useState(false)
```
rogercloud, line 246: **minor, informational — busyType guards concurrent type-toggles with a scalar, unlike the builder's Set**

This scalar `busyType` is pre-existing and unchanged by this PR, but this PR's new summary cards in `agent-builder.tsx:401-403` deliberately use a `ReadonlySet<AgentTriggerType>` (comment: "a Set so two types disabling concurrently keep their own guard"). Toggling two different types back-to-back here makes each `setBusyType(newType)` overwrite the prior guard, briefly re-enabling the first type's switch/navigation while its own PATCH is still in flight (and the unconditional `finally { setBusyType(null) }` opens a second such window on resolve). No data corruption — the two PATCH calls are independent — but it's a real, reachable consistency gap now that this same PR introduced the more correct pattern next to it.

Suggestion: key `busyType` by type (a `Set`, like the builder). Deferrable, not a blocker.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -396,6 +398,40 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
   }
 
+  // Builder card switch: toggling it off disables every enabled trigger of
+  // that type in place — no dialog. The card disappears afterwards because
+  // the summary section only lists types with enabled > 0. A Set (not a
+  // single value) so two types disabling concurrently keep their own guard.
+  const [disablingTriggerTypes, setDisablingTriggerTypes] = useState<ReadonlySet<AgentTriggerType>>(
+    new Set(),
+  )
+  const disableTriggerType = async (type: AgentTriggerType) => {
```
rogercloud, line 446: **minor — untested staging-mode branch**

The `if (!localAgentId) { setStagedTriggers(...) }` branch here has no test coverage — both tests in `agent-builder-triggers.test.tsx` render `AgentBuilder` with a live `agentId`, never the pre-creation staging state. (The dialog's own staged batch-disable path is tested at `agent-triggers-dialog.test.tsx:1124`, but that's a different component/function and doesn't reach this path.)

Suggestion: add one staging-state test for this branch.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -582,47 +659,127 @@ export function AgentTriggersDialog({
...
     }
   }
 
-  const handleSubmit = async () => {
+  // The enabled switch in the detail view applies immediately — no Save
+  // needed. Toggling on in the creation state creates the trigger right away
+  // from the current form values (delegated to handleSubmit so the two create
+  // paths cannot drift), mirroring the overview quick toggle.
+  const handleDetailToggle = async (checked: boolean) => {
+    setFormValue("enabled", checked)
```
rogercloud, line 723: **minor, informational — form.enabled isn't reconciled from the response (latent, not currently reachable)**

`form.enabled` is set optimistically here before the PATCH, and the success path (~740-747) only reconciles `liveTriggers` from the response, never `form`. If the backend ever returned a 200 with an `enabled` differing from the request, the detail switch (`form.enabled`) and the list (response-derived) would disagree until the next full resync. Verified against `src/xagent/web/services/triggers.py`: `update_trigger` sets `enabled` directly and unconditionally from the request with no overriding rule, so this can't happen today — purely a defensive note for the future, not an active bug.

frontend/src/components/build/agent-triggers-dialog.test.tsx
```diff
@@ -265,6 +273,581 @@ describe("AgentTriggersDialog", () => {
...
+      expect(toastMocks.error).toHaveBeenCalled()
+    })
+    await waitFor(() => {
+      expect(getCallsAfterFailure).toBeGreaterThan(0)
+    })
+  })
+
+  it("keeps each overview switch's busy guard independent across two types toggled back-to-back", async () => {
+    const TRIGGERS_URL = "http://api.local/api/agents/42/triggers"
+    const triggers = [
```
rogercloud, line 761: **simplification (yagni), second round unaddressed**: the suggested `makeTrigger(overrides)` test-fixture factory still doesn't exist, and duplication ticked up rather than down — inline full-trigger literals went from 7 to 9 across this update. This commit added a small local win (`baseTrigger9`, deduping two prior copies in one describe block), but this new busy-guard test then hand-inlines two more ~15-line trigger objects (id 40/41) right next to it — exactly the pattern the factory would absorb.

Not a blocker, but worth adopting next time a test in these files needs a trigger literal — it would have made this commit's own new tests shorter.

rogercloud (commented):
## Re-review — update commit `fd2bab4b`

### Update summary

This commit (`fd2bab4b`, on top of `82964b94`) addresses all three findings from the previous round:

- **Busy-guard keyed by type.** The dialog's scalar guard is replaced with a `busyTypes` Set (`useState<ReadonlySet<AgentTriggerType>>(new Set())`), matching the pattern the builder already used. Types are added before async work and removed in `finally`, and the overview switch's `disabled` now reads `busyTypes.has(type)`. A dedicated regression test toggles two different types back-to-back, resolves the second first, and asserts the first switch stays disabled until its own request settles.
- **Detail switch reconciled from the PATCH response.** `handleDetailToggle`'s success path now sets `form.enabled` from `updated.enabled` (not just the optimistic value), gated by the same stale-closure guard used elsewhere in the handler. A new test simulates a response that diverges from the requested value and asserts the switch reflects the response.
- **Builder staging-mode coverage.** A new test renders `<AgentBuilder />` with no agentId, toggles a staged trigger off via its summary card, and asserts the UI updates with no `/triggers` network call — exercising the previously-untested `!localAgentId` branch.

### Design verdict — acceptable-with-reservations

This is a genuine improvement over the previous round, not just "still fine." The Set-based guards on both the dialog and builder, response-driven reconciliation, the stale-closure guard, and the shared helpers are now all correctly implemented and consistent with each other.

Two standing design-level observations carry forward (neither is a blocker, neither owns a single line):

- **Scope.** #929 asked for one behavior (auto-apply the toggle). This PR delivers that inside a much larger dialog rewrite (Save button removed, delete UX changed, secret-gating added). It's internally coherent, but the large blast radius is why each round has kept surfacing small items; a narrower PR would have converged sooner.
- **State-coordination surface.** The set of coordinating pieces (`busy`/`busyTypes`, `formDirty`, `syncedFormKeyRef`, `formKeyAtStart`, and several refs) grew this round rather than shrank. Each addition is independently justified and well-commented, but a reader must hold all of them at once to reason about any single handler. It's near the ceiling of what's maintainable before extracting a reducer or a custom hook would pay off. Worth keeping in mind for a future cleanup, not something to act on here.

### Carried-over items

- Finding 1 (scalar busy-guard re-enabling the wrong switch) — **FIXED** (Set-keyed guard + regression test).
- Finding 2 (builder staging-mode branch untested) — **FIXED** (new no-agent staging test).
- Finding 3 (`form.enabled` not reconciled from the response) — **FIXED** (reconcile added under the stale-closure guard + divergent-response test).
- F4 (toggle-rollback race — `formKeyAtStart` mismatch branch formally uncovered) — **STILL PARTIAL.** Guard behavior is unchanged and correct; no test in this update touches it. The UI-level test still shows the race is practically unreachable (navigation is disabled in-flight), so the branch stays formally uncovered but low-risk. Unchanged from last round, not a regression.
- F7 (one-time secret reveal persists across navigation to unrelated trigger types) — **UNCHANGED, acceptable.** `secretReveal` is untouched this round; still a light UX suggestion, not a defect.

### Closing

Two candidate findings surfaced during discovery this round were investigated and dropped — a "double `notifyChanged()` on quick-toggle-create" in `handleTypeToggle`/`createDefaultTrigger`, and an `exhaustive-deps` warning on `activeTypeTriggers` — both confirmed byte-for-byte pre-existing in the base commit and untouched by this PR, so they're out of scope and not listed as findings.

Verification this round was run directly, not inferred: `vitest run` on both test files passes 39/39, `tsc --noEmit` is clean on the changed files, and `eslint` introduces no new errors or warnings (the one dialog-file warning and the `agent-builder.tsx` errors are all pre-existing, outside this PR's hunks).

Net: all three prior findings resolved with targeted tests, no regressions, design unchanged in substance and improved in implementation. Converging well — the only open loop worth acting on is the test-fixture factory (see inline comment).

frontend/src/lib/agent-triggers-api.ts
```diff
@@ -275,6 +275,42 @@ export async function deleteOwnerTrigger(
...
+  triggers: AgentTrigger[],
+  type: AgentTriggerType,
+): Promise<AgentTrigger[]> {
+  const enabled = triggers.filter((trigger) => trigger.type === type && trigger.enabled)
+  return Promise.all(
+    enabled.map((trigger) => updateOwnerTrigger(owner, trigger.id, { enabled: false })),
+  )
+}
+
+export async function disableAgentTriggersOfType(
```
rogercloud, line 296: **nit, low priority — thin pass-through wrapper**

`disableAgentTriggersOfType` is a one-line pass-through to `disableOwnerTriggersOfType({kind:"agent",id})`, used only by `agent-builder.tsx:449`. The builder could call the owner-based version directly instead. Not worth acting on in isolation — it matches the file's existing convention of keeping an agent-scoped alias next to each owner-based function.

rogercloud (commented):
## Re-review — PR #972 (4th round, at commit `8ce5020`)

### Update summary (read this first)

The important thing to state plainly about this update: it is dominated by a **rebase onto a new main**, not by new work in this PR. In the interim, main merged an unrelated *workforce-triggers* feature (issue #950) that generalized the trigger API from agent-specific to **owner**-generic — `TriggerOwnerRef`, `createOwnerTrigger`/`updateOwnerTrigger`/etc., supporting both `{kind:"agent"}` and `{kind:"workforce"}` owners — and added an "owner routing" test describe block. **That owner-generic layer, including the owner-routing tests, was not built by this PR; it already existed in the new base commit.**

On top of the rebase, this PR's own commits only:
- **(a)** added 3 helper functions (`disableOwnerTriggersOfType`, `disableAgentTriggersOfType`, `mergeUpdatedTriggers`) that plug this PR's auto-save logic into the pre-existing owner layer — confirmed this is the ENTIRE diff this PR carries in `agent-triggers-api.ts` relative to the new base (a single hunk, nothing else in the file was touched). Integration verified clean: every write path this PR added (`handleDetailToggle`, `handleTypeToggle`, `handleDelete`, `handleSubmit`, `createDefaultTrigger`) uniformly threads `resolvedOwner`, with no agent-only hardcoding left behind, and `agent-builder.tsx` correctly stays agent-only (only agents have builder pages).
- **(b)** finally extracted the `makeTrigger(overrides)` test fixture factory that had been suggested and skipped in the previous two rounds.

### Design verdict: sound (upgrade from "acceptable-with-reservations")

This is a genuine upgrade, and the reason is worth being explicit about: the owner-generic complexity I flagged as a "standing observation" last round turns out to be **inherited from main, not introduced here**. Stripped of that, this PR's contribution reads as a clean, small integration into an already-abstracted layer rather than something bolted on. The core redesign — auto-save via commit-on-navigation — remains a real root-cause fix for #929, uniformly applied across all write paths.

### Status of carried-over items

- **busyTypes Set / form.enabled reconciliation / builder staging test (all fixed in round 3) — STILL INTACT after the rebase.** Verified line-by-line: `busyTypes` remains a `ReadonlySet<AgentTriggerType>` (`agent-triggers-dialog.tsx:250`, mutated 658/709-715, consumed 1129); `form.enabled` reconciliation still runs under the `formKeyAtStart` guard (`handleDetailToggle`, 768-780), now calling the renamed `updateOwnerTrigger` — a mechanical adaptation, not a behavior change; builder staging test present and passing (`agent-builder-triggers.test.tsx:371`).
- **makeTrigger fixture factory — FINALLY FIXED.** Present in both `agent-triggers-dialog.test.tsx:57-75` and `agent-builder-triggers.test.tsx:153-171` (independent per-file factories, differing only in the default `webhook_token`: `null` vs `"tok"`). Adoption is complete — every previously-flagged inline literal now routes through the factory, zero raw trigger literals remain in either file. This closes out a suggestion raised twice before.
- **F4 (toggle-rollback race, `formKeyAtStart` mismatch branch) — STILL PARTIAL** (3rd round for this one). Guard code intact (lines 731/745/778/785); no test exercises the actual mismatch branch — still only "practically unreachable via the UI," not directly tested. No regression, just still open.
- **F7 (one-time secret reveal not scoped to type/id) — unchanged, still acceptable.** `secretReveal` still has no scoping (`agent-triggers-dialog.tsx:259`), comment at 480-482 confirms this remains deliberate. Always a light suggestion, never a defect.

### New finding — workforce write-path coverage gap (minor, no single line in this PR's diff owns it, so noting here rather than inline)

`agent-triggers-dialog.tsx` takes an explicit `owner?: TriggerOwnerRef | null` prop (line 68), and this is a live production configuration: `frontend/src/app/workforces/[id]/page-client.tsx:499-505` renders `<AgentTriggersDialog agentId={null} owner={{kind:"workforce", id}} .../>` on the workforce detail page. The pre-existing "owner routing" tests (`agent-triggers-dialog.test.tsx:1187-1227`, inherited from base, not part of this PR's diff) only assert the initial GET route differs for a workforce owner — no test anywhere exercises this PR's actual feature (PATCH-on-toggle, POST-on-create, DELETE-on-delete) against a workforce owner. The code is correct by construction (every write path this PR added threads `resolvedOwner` uniformly), so this isn't a known bug — but it's a real, currently-unverified gap in a feature otherwise covered thoroughly for the agent case. Suggest adding at least one toggle/create/delete test using a workforce owner, mirroring the existing GET-only coverage.

### Wrap-up

This PR has converged well across 4 rounds. The design verdict is now unreserved, and the only things still open are the coverage-gap suggestion above and the two long-standing items (F4/F7) that were never blockers. Nice progress on the fixture factory — that one had been outstanding for a while.

One thing I looked into and dropped: a set of 4 legacy `*AgentTrigger*` wrapper functions in `agent-triggers-api.ts` (`updateAgentTrigger`/`deleteAgentTrigger`/`listAgentTriggerRuns`/`testAgentTrigger`) look like dead code with no production callers — but they're already present unchanged in the new base commit, not introduced or touched by this PR, so they're out of scope here.

Tests were run directly this round: `vitest run` on both component test files → **41/41 pass**; `tsc --noEmit` clean.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -678,33 +847,91 @@ export function AgentTriggersDialog({
...
+  const secretPending = (result: SubmitResult) => Boolean(result.secret || secretReveal)
+
+  const handleDone = async () => {
+    const result = await commitPendingEdits()
+    if (!result.ok) return
+    if (secretPending(result)) return
+    closeDialog(false)
+  }
+
+  const handleBack = async () => {
```
rogercloud, line 917: These three navigation handlers gate only on `result.ok` after `commitPendingEdits()`, unlike `handleDone` (~912) and `handleDismiss` (~1062) which also gate on `secretPending`. If the commit here creates a new webhook trigger with a one-time secret and the user immediately clicks Back / another pill / Add, the unscoped `renderSecretReveal` alert stays mounted and visually attaches to the newly-navigated view rather than the trigger that owns the secret. Not data loss (the secret is still shown), but the alert is mis-attributed. Consider gating on `secretPending` here too, or scoping the reveal to its owning trigger.

frontend/src/components/build/agent-triggers-dialog.test.tsx
```diff
@@ -456,4 +1224,124 @@ describe("AgentTriggersDialog owner routing", () => {
...
+        expect.objectContaining({ method: "POST" }),
+      )
+    })
+    expect(apiRequestMock).not.toHaveBeenCalledWith(
+      expect.stringContaining("/api/agents/"),
+      expect.anything(),
+    )
+  })
+
+  it("deletes a workforce-owned trigger via DELETE on the workforce route", async () => {
```
rogercloud, line 1453: This commit adds workforce-owner routing tests for overview-card disable (PATCH), create (POST), and delete (DELETE), but not for toggling an existing trigger's `enabled` state from the detail/form view (`handleDetailToggle`'s existing-trigger branch, ~754-790) under `owner={{kind:"workforce"}}` — that combination is only covered under the agent-owner route. The path shares the same `updateOwnerTrigger`/`ownerTriggerUrl` plumbing the other 3 workforce tests already prove, so risk is low. Nice-to-have to close the matrix.

frontend/src/components/build/agent-triggers-dialog.tsx
```diff
@@ -189,6 +211,14 @@ function newestFirst(a: AgentTrigger, b: AgentTrigger): number {
   return Math.abs(b.id) - Math.abs(a.id)
 }
 
+// After deleting the selected trigger, the next one to show — same type,
+// newest first — or null if none remain. Shared by handleDelete's staged and
+// live branches, which otherwise differ only in whether the list still needs
+// stagedToPseudoTrigger mapping before this runs.
+function pickNextAfterDelete(remaining: AgentTrigger[], type: AgentTriggerType): number | null {
```
rogercloud, line 218: `pickNextAfterDelete` returns the newest-first next same-type trigger after a deletion, but its real-return branch is never exercised: existing delete tests either delete the only item of a type (falls back to empty) or delete a non-selected pill (selection preserved). No test creates 2+ same-type triggers, deletes the currently-selected one, and asserts the newest-first survivor becomes selected — so a broken sort/filter here would go uncaught. Worth one test covering that branch.

rogercloud (approved):
## Re-review (round 5) — Update summary (read this first)

Since the last review round (commit `8ce5020`), exactly **one** commit was pushed:
`e84ffea` *"test: cover workforce-owned trigger writes (toggle/create/delete)"*. It is
**test-only** (~120 lines added to `frontend/src/components/build/agent-triggers-dialog.test.tsx`
covering workforce-owner routing: toggle via PATCH, create via POST, delete via DELETE) with
**no production code changes**. Current head: `e84ffea`.

Beyond re-verifying every prior-round finding against the current code, this round **also**
includes a full independent blank-slate design/discovery pass over the whole diff (not just
prior-finding verification). That pass surfaced a few new **minor** findings (B/C/D below);
none is blocking.

All 44 relevant tests pass (`vitest run agent-triggers-dialog.test.tsx agent-builder-triggers.test.tsx`).

---

## What this PR changes

Issue #929 (P0) asked for one thing: make the trigger enable/disable toggle **auto-save** instead
of requiring a "Save Changes" click. This PR delivers that and considerably more — it removes the
Save button entirely and converts **all** field edits to implicit commit-on-navigate
(Done / Back / switch-pill / Add / dismiss each commit pending edits), moves delete confirmation to
an inline pill popover, changes the default state for new triggers, and fixes several races
(stale-closure toggle rollback, Gmail quick-toggle `formDirty` stuck, redundant refetch on close).
The busy-guard and resync helpers were extracted and shared at the API layer.

---

## Round 0 approach verdict: **acceptable-with-reservations**

Three design-level reservations (not tied to a single line, not blocking):

1. **Scope creep beyond issue #929.** The issue authorized only enable/disable auto-save. This PR
   silently expanded that into implicit-commit-on-navigate for *all* form fields, plus a UI reflow
   (delete-via-popover) and a default-state change for new triggers. Implicit save-on-navigate whose
   failures surface only via a toast — potentially after the user has already navigated away — is a
   real UX/product decision beyond what the linked issue authorized. Worth an explicit product/design
   sign-off, not a defect.
2. **Component-level duplication of the busy-guard pattern.** The `Set<AgentTriggerType>` busy-guard
   is implemented once in `agent-triggers-dialog.tsx` (`busyTypes`) and again, in parallel, in
   `agent-builder.tsx` (`disablingTriggerTypes`) for the summary-card path. The API-layer helpers
   (`disableOwnerTriggersOfType`, `mergeUpdatedTriggers`) were properly shared; this component-level
   orchestration was not. Acceptable-for-now DRY gap.
3. **Fragile fresh/stale-closure dual-read pattern.** `secretPending` and the
   `formKeyAtStart`/`syncedFormKeyRef` mechanism are correct but rely on the reader grasping a subtle
   stale-vs-fresh-closure distinction. Verified correct by testing; flagged as a readability/maintenance
   risk.

---

## Prior-round findings — status against current code

| # | Finding | Status |
|---|---------|--------|
| 1 | Stale server fields after detail-view toggle (`handleDetailToggle`) | **FIXED** — now patches `liveTriggers`/`form.enabled` from the `updated` server response, guarded by `formKeyAtStart` |
| 2 | Untested toggle-rollback race (`formKeyAtStart` guard) | **NOT FIXED** — no test starts a toggle, navigates away, fails it, and asserts no clobber. Guard remains unexercised by normal interaction |
| 3 | Untested non-atomic bulk-disable resync (dialog) | **PARTIAL** — a test now asserts resync is *triggered* (toast + follow-up GET) but not the *reconciled* state |
| 4 | Untested bulk-disable resync (builder/summary-card) | **FIXED** — covered by "resyncs the summary via refreshTriggerSummary when a batch disable partially fails" |
| 5 | Untested dismiss-with-failing-commit path | **FIXED** — now intentional/documented, with a dedicated test |
| 6 | Cross-navigation persistence of one-time secret reveal (UX note) | **UNCHANGED** — now documented in-code as intentional; no action (see finding B for a related edge) |
| 7 | yagni: `triggers` typed as `Pick<AgentTrigger,…>` | **RESOLVED** — full `AgentTrigger[]` at both call sites; thread can be resolved |
| 8 | shrink: duplicated next-trigger-pick logic | **RESOLVED** — consolidated into shared `pickNextAfterDelete`; thread can be resolved |
| 9 | busyType scalar vs builder's Set | **FIXED** — now a `Set<AgentTriggerType>` with explanatory comment; thread can be resolved |
| 10 | Untested staging-mode branch (`agent-builder.tsx`) | **FIXED** — covered by the staged-in-place test |
| 11 | `form.enabled` not reconciled from response (latent) | **FIXED** — reconciled from `updated.enabled`, same `formKeyAtStart` guard |
| 12 | yagni: `makeTrigger` fixture factory "still doesn't exist" | **FIXED / finding was stale** — factory exists (introduced in `8ce5020`, predating the round that flagged it), used at 13 sites incl. all 3 new tests; duplication went *down*. The "second round unaddressed" label was itself based on a stale diff view |
| 13 | nit: `disableAgentTriggersOfType` thin pass-through wrapper | **STILL STANDS** — carried forward as a simplification nit (see below) |

---

## New findings this round

**B — [minor, PR-introduced] Secret-reveal mis-attribution on quick-navigate.**
`handleBack`, `handleSelectTrigger`, and `handleAddAnother` call `commitPendingEdits()` and gate only
on `result.ok` before navigating — unlike `handleDone` and `handleDismiss`, which also gate on
`secretPending`. If `commitPendingEdits()` creates a new webhook trigger with a one-time secret and the
user immediately clicks Back / another pill / Add, the unscoped secret-reveal alert
(`renderSecretReveal`) ends up visually attached to the newly-navigated view rather than the trigger
that owns the secret. Not data loss (the secret is still shown), but a real UI mis-attribution.

**C — [low-priority test gap] Detail-view existing-trigger toggle under workforce owner is untested.**
The new commit added workforce-owner routing tests for overview-card disable (PATCH), create (POST),
and delete (DELETE), but not for toggling an *existing* trigger's `enabled` state from the detail/form
view (`handleDetailToggle`'s existing-trigger branch) under `owner={{kind:"workforce"}}` — that
combination is only covered under the agent-owner route. The path shares the identical
`updateOwnerTrigger`/`ownerTriggerUrl` plumbing already proven by the 3 new workforce tests, so risk is
low. Nice-to-have coverage completion.

**D — [minor test gap] `pickNextAfterDelete` "returns a real next id" branch is untested.**
`pickNextAfterDelete` picks the next same-type trigger (newest-first) after a deletion. Every existing
delete test either deletes the only item of a type (falls back to empty) or deletes a *non-selected*
pill (selection preserved, helper not exercised in its real-return branch). No test creates 2+
same-type triggers, deletes the currently-selected one, and asserts the newest-first remaining one
becomes selected — so a broken sort/filter here would go uncaught.

**A — [minor, informational, NOT a regression — pre-existing] Double `notifyChanged()` on quick-toggle create.**
`handleTypeToggle`'s quick-toggle create sub-path fires the parent `notifyChanged()` callback twice —
once inside `createDefaultTrigger` and once unconditionally after it returns — whereas the
enable-existing and disable sub-paths fire it once. Verified this exact double-call already exists at
the base commit (`7586f51d`); it predates this PR and was carried forward, not introduced. Impact: one
extra idempotent refetch on create — not a correctness bug. Noted only because it ironically
contradicts this PR's own goal of removing redundant refetches. Pre-existing / optional.

*(Considered and dropped: staging-mode trigger delete not clearing `secretReveal` — confirmed
unreachable, since staged triggers never populate `secretReveal` without a server round-trip.)*

---

- `agent-triggers-api.ts:296` — **yagni: thin one-call-site wrapper.** `disableAgentTriggersOfType`
  only forwards to `disableOwnerTriggersOfType({kind:"agent",id})`, and `agent-builder.tsx:449` is its
  sole caller. Call `disableOwnerTriggersOfType` directly and drop the wrapper.
- `agent-triggers-dialog.tsx:806` — **overly generic override type.** `handleSubmit`'s
  `overrides?: Partial<Pick<TriggerFormState,"enabled">>` is only ever `{enabled: true}` or omitted; a
  plain `enabled?: boolean` would do.
- `agent-triggers-dialog.tsx:510` — **verbose truthiness check.**
  `setFormDirty(Boolean(initial && Object.keys(initial).length > 0))` can be `Boolean(initial)`, since
  `beginCreateForType` is only ever called with `undefined` or a non-empty object — never `{}`.

net: ~-4 lines possible (type/expression simplifications, not deletions).

---

Approving: production behavior is unchanged since the last round, the one new commit is sound
test-only coverage, and no critical/major issues remain open. Recommend a product/design sign-off on
reservation #1 (scope) and optional follow-up on findings B–D at your discretion.

PR #942, opened 2026-07-21
enh: improve branded static visual generation

qinxuye:
> [!IMPORTANT]
> Depends on #943. Merge #943 first, then rebase this branch onto main to remove the duplicated general execution-reliability changes and leave the ad-creative-specific diff.

## Summary

- add a static-visual-design skill for art-directed posters and ad creatives, including concept selection, campaign-device ideation, variant generation, and visual QA
- prefer direct image generation for modern static creatives while treating logos, QR codes, and exact brand assets as references instead of rebuilding the whole design in HTML
- add official website asset discovery/download plus direct SVG-source understanding, and remove the brittle logo-overlay workflow
- harden auto/DAG/ReAct execution around skill loading, unavailable tools, malformed DeepSeek tool arguments, child LLM failures, and generated-artifact completion
- report the actual image model used and make checkpoint blob persistence safe under parallel DAG execution

## Why

Static ad-generation tasks could fail before producing an artifact, select stale composition workflows, lose brand context, or stop on recoverable provider/tool-protocol errors. Even successful runs often optimized for “an image exists” rather than for a coherent, campaign-grade design.

This change moves the workflow toward a single art-directed creative concept, direct high-quality generation, reliable use of official brand references, and explicit inspection/revision before delivery.

## Validation

- 107 passed in focused tests covering the static visual skill, web/SVG asset handling, DeepSeek/OpenRouter tool protocols, trace storage, and auto/DAG/ReAct recovery/completion behavior
- Ruff passed on all changed Python files
- git diff --check upstream/main...HEAD

PR #977, opened 2026-07-23
feat: add safe SVG web asset support

qinxuye:
## Summary

- discover and download exact static assets from official web pages
- validate and rasterize SVG inputs for media workflows
- share bounded public-HTTP fetching with redirect and private-host validation

## Why

This isolates the web-asset and SVG capability from #942 so the static visual skill change can remain focused on creative generation behavior.

## Validation

- `pytest -q tests/core/utils/test_svg.py tests/core/model/test_security.py tests/core/tools/test_download_web_asset.py tests/core/tools/test_fetch_web_content.py tests/core/tools/test_vision_tool.py tests/core/tools/test_web_tool_categories.py tests/core/tools/adapters/vibe/test_basic_tools_web_search_provider.py` (123 passed)
- `ruff check` on all changed Python files
- commit hooks: ruff, formatting, mypy, codespell, and migration checks

### DIFF
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -129,6 +129,7 @@ SESSION_SECRET = "XAGENT_SESSION_SECRET"
 OPENROUTER_OFFICIAL_PROVIDERS_ONLY = "XAGENT_OPENROUTER_OFFICIAL_PROVIDERS_ONLY"
 MCP_OAUTH_ALLOW_PRIVATE_HOSTS = "XAGENT_MCP_OAUTH_ALLOW_PRIVATE_HOSTS"
 MCP_OAUTH_PROXY_URL = "XAGENT_MCP_OAUTH_PROXY_URL"
+TRUSTED_EGRESS_PROXY = "XAGENT_TRUSTED_EGRESS_PROXY"
 
 TOOL_MAX_OUTPUT_LENGTH = "XAGENT_TOOL_MAX_OUTPUT_LENGTH"
 TOOL_MAX_RECURSION_DEPTH = "XAGENT_TOOL_MAX_RECURSION_DEPTH"
@@ -341,6 +342,19 @@ def get_mcp_oauth_allow_private_hosts() -> bool:
     servers. Production deployments should leave it disabled.
     """
     return _get_bool_env(MCP_OAUTH_ALLOW_PRIVATE_HOSTS, False)
+
+
+def get_trusted_egress_proxy_enabled() -> bool:
+    """Return whether the ambient HTTP(S)_PROXY may be used for public fetches.
+
+    Outbound SSRF guarding pins the DNS resolution used at validation time
+    to the one used at connect time. Routing through a proxy breaks that
+    guarantee, since the proxy performs its own, independent DNS resolution
+    of the target host. This flag is an explicit opt-in acknowledging that
+    the configured proxy is trusted to enforce its own private-range egress
+    policy; leave it disabled unless that is true.
+    """
+    return _get_bool_env(TRUSTED_EGRESS_PROXY, False)
 
 
 def get_mcp_oauth_proxy_url() -> str | None:
--- a/src/xagent/core/tools/adapters/vibe/download_web_asset.py
+++ b/src/xagent/core/tools/adapters/vibe/download_web_asset.py
@@ -142,6 +142,7 @@ class DownloadWebAssetTool(AbstractBaseTool):
                 max_content_bytes=self._max_content_bytes,
                 resource_name="remote asset",
                 require_non_empty=True,
+                via_proxy=bool(proxy_url),
             )
         return response.content, response.url, response.content_type
 
--- a/src/xagent/core/tools/adapters/vibe/download_web_asset.py
+++ b/src/xagent/core/tools/adapters/vibe/download_web_asset.py
@@ -13,6 +13,7 @@ from pydantic import BaseModel, Field
 
 from ....file_ref import build_workspace_file_ref
 from ....utils.security import fetch_public_http_bytes
+from ....utils.svg import validate_svg_bytes
 from ....workspace import TaskWorkspace
 from ...core.web_content import (
     DEFAULT_MAX_CONTENT_BYTES,
@@ -140,7 +141,6 @@ class DownloadWebAssetTool(AbstractBaseTool):
                 timeout=30,
                 max_content_bytes=self._max_content_bytes,
                 resource_name="remote asset",
-                content_length_name="declared",
                 require_non_empty=True,
             )
         return response.content, response.url, response.content_type
@@ -157,9 +157,7 @@ class DownloadWebAssetTool(AbstractBaseTool):
         if media_type and not media_type.startswith("image/"):
             raise ValueError(f"Unsupported web asset content type: {content_type}")
         if media_type == "image/svg+xml" or url_suffix == ".svg":
-            prefix = content[:4096].decode("utf-8", errors="ignore").lower()
-            if "<svg" not in prefix:
-                raise ValueError("Remote SVG asset does not contain an <svg> root")
+            validate_svg_bytes(content)
             return ".svg"
         try:
             with Image.open(io.BytesIO(content)) as image:
@@ -193,8 +191,14 @@ class DownloadWebAssetTool(AbstractBaseTool):
             filename = "web_asset"
         if "\x00" in filename:
             raise ValueError("filename contains an invalid null byte")
-        if not Path(filename).suffix:
+        existing = Path(filename).suffix.lower()
+        jpeg_aliases = {".jpg", ".jpeg"}
+        if not existing:
             filename = f"{filename}{detected_extension}"
+        elif existing != detected_extension and not (
+            existing in jpeg_aliases and detected_extension in jpeg_aliases
+        ):
+            filename = f"{Path(filename).stem}{detected_extension}"
         return filename
 
     def _write_unique_output(self, filename: str, content: bytes) -> Path:
--- a/src/xagent/core/tools/adapters/vibe/download_web_asset.py
+++ b/src/xagent/core/tools/adapters/vibe/download_web_asset.py
@@ -18,7 +18,7 @@ from ....workspace import TaskWorkspace
 from ...core.web_content import (
     DEFAULT_MAX_CONTENT_BYTES,
     DEFAULT_USER_AGENT,
-    get_proxy_url,
+    get_trusted_proxy_url,
 )
 from .base import AbstractBaseTool, ToolCategory, ToolVisibility
 
@@ -129,7 +129,7 @@ class DownloadWebAssetTool(AbstractBaseTool):
 
     async def _download(self, url: str) -> tuple[bytes, str, str]:
         client_kwargs: dict[str, Any] = {}
-        proxy_url = get_proxy_url()
+        proxy_url = get_trusted_proxy_url()
         if proxy_url:
             client_kwargs["proxy"] = proxy_url
 
--- a/src/xagent/core/tools/adapters/vibe/download_web_asset.py
+++ b/src/xagent/core/tools/adapters/vibe/download_web_asset.py
@@ -1,0 +1,228 @@
+"""Download an exact web image asset into the current task workspace."""
+
+from __future__ import annotations
+
+import io
+from pathlib import Path
+from typing import Any, Mapping, Type
+from urllib.parse import unquote, urlsplit
+
+import httpx
+from PIL import Image
+from pydantic import BaseModel, Field
+
+from ....file_ref import build_workspace_file_ref
+from ....utils.security import fetch_public_http_bytes
+from ....workspace import TaskWorkspace
+from ...core.web_content import (
+    DEFAULT_MAX_CONTENT_BYTES,
+    DEFAULT_USER_AGENT,
+    get_proxy_url,
+)
+from .base import AbstractBaseTool, ToolCategory, ToolVisibility
+
+
+class DownloadWebAssetArgs(BaseModel):
+    url: str = Field(
+        description=(
+            "Exact HTTP or HTTPS image URL discovered on an authoritative page. "
+            "Prefer the official brand domain for logos."
+        )
+    )
+    filename: str | None = Field(
+        default=None,
+        description=(
+            "Optional output basename. The remote filename and MIME type are used "
+            "when omitted."
+        ),
+    )
+
+
+class DownloadWebAssetResult(BaseModel):
+    success: bool
+    source_url: str
+    content_type: str = ""
+    size: int | None = None
+    filename: str | None = None
+    file_id: str | None = None
+    file_ref: dict[str, Any] | None = None
+    markdown_link: str | None = None
+    error: str | None = None
+
+
+class DownloadWebAssetTool(AbstractBaseTool):
+    """Download and register an exact remote image without model reconstruction."""
+
+    category = ToolCategory.WEB_SEARCH
+
+    def __init__(
+        self,
+        workspace: TaskWorkspace,
+        *,
+        max_content_bytes: int = DEFAULT_MAX_CONTENT_BYTES,
+    ) -> None:
+        self._visibility = ToolVisibility.PUBLIC
+        self._workspace = workspace
+        self._max_content_bytes = max_content_bytes
+
+    @property
+    def name(self) -> str:
+        return "download_web_asset"
+
+    @property
+    def description(self) -> str:
+        return (
+            "Download an exact PNG, JPEG, WebP, GIF, BMP, or SVG asset from a "
+            "known URL and register it in the current workspace. Use this after "
+            "fetch_web_content(include_assets=true) discovers an official logo or "
+            "brand image. It returns a trusted FileRef, so do not reproduce the "
+            "asset through api_call plus write_file."
+        )
+
+    @property
+    def tags(self) -> list[str]:
+        return ["web", "image", "asset", "download", "logo"]
+
+    def args_type(self) -> Type[BaseModel]:
+        return DownloadWebAssetArgs
+
+    def return_type(self) -> Type[BaseModel]:
+        return DownloadWebAssetResult
+
+    def run_json_sync(self, args: Mapping[str, Any]) -> Any:
+        raise NotImplementedError("DownloadWebAssetTool only supports async execution.")
+
+    async def run_json_async(self, args: Mapping[str, Any]) -> Any:
+        download_args = DownloadWebAssetArgs.model_validate(args)
+        try:
+            content, final_url, content_type = await self._download(download_args.url)
+            detected_extension = self._validate_image(content, content_type, final_url)
+            filename = self._resolve_filename(
+                download_args.filename,
+                final_url,
+                detected_extension,
+            )
+            with self._workspace.auto_register_files():
+                target = self._write_unique_output(filename, content)
+            file_ref = build_workspace_file_ref(
+                workspace=self._workspace,
+                file_path=target,
+                mime_type=self._mime_type_for_extension(detected_extension),
+            )
+            return DownloadWebAssetResult(
+                success=True,
+                source_url=final_url,
+                content_type=content_type,
+                size=len(content),
+                filename=target.name,
+                file_id=file_ref["file_id"],
+                file_ref=file_ref,
+                markdown_link=file_ref["markdown_link"],
+            ).model_dump()
+        except Exception as exc:
+            return DownloadWebAssetResult(
+                success=False,
+                source_url=download_args.url,
+                error=str(exc),
+            ).model_dump()
+
+    async def _download(self, url: str) -> tuple[bytes, str, str]:
+        client_kwargs: dict[str, Any] = {}
+        proxy_url = get_proxy_url()
+        if proxy_url:
+            client_kwargs["proxy"] = proxy_url
+
+        async with httpx.AsyncClient(**client_kwargs) as client:
+            response = await fetch_public_http_bytes(
+                client,
+                url,
+                headers={"User-Agent": DEFAULT_USER_AGENT},
+                timeout=30,
+                max_content_bytes=self._max_content_bytes,
+                resource_name="remote asset",
+                content_length_name="declared",
+                require_non_empty=True,
+            )
+        return response.content, response.url, response.content_type
+
+    @classmethod
+    def _validate_image(
+        cls,
+        content: bytes,
+        content_type: str,
+        url: str,
+    ) -> str:
+        media_type = cls._media_type(content_type)
+        url_suffix = Path(urlsplit(url).path).suffix.lower()
+        if media_type and not media_type.startswith("image/"):
+            raise ValueError(f"Unsupported web asset content type: {content_type}")
+        if media_type == "image/svg+xml" or url_suffix == ".svg":
+            prefix = content[:4096].decode("utf-8", errors="ignore").lower()
+            if "<svg" not in prefix:
+                raise ValueError("Remote SVG asset does not contain an <svg> root")
+            return ".svg"
+        try:
+            with Image.open(io.BytesIO(content)) as image:
+                image_format = image.format
+                image.verify()
+        except Exception as exc:
+            raise ValueError("Remote asset is not a valid supported image") from exc
+        registered_extensions = Image.registered_extensions()
+        for extension, registered_format in registered_extensions.items():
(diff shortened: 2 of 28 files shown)


src/xagent/core/utils/security.py
```diff
@@ -50,10 +74,161 @@ def reject_private_network_host(hostname: str) -> None:
...
+
+    logical_url = url
+    display_name = resource_name[:1].upper() + resource_name[1:]
+    for redirect_count in range(_MAX_REDIRECTS + 1):
+        resolved_ips = await validate_public_http_url(logical_url)
+        parsed = urlsplit(logical_url)
+        connect_url, host_header, sni = _pin_url_to_ip(parsed, resolved_ips[0])
+        request_headers = dict(headers or {})
+        request_headers["Host"] = host_header
+        async with client.stream(
```
qinxuye, line 185: [P1] Preserve HTTPS proxy compatibility when pinning the target

Rewriting the request URL to the validated IP works for a direct httpcore connection, but not for an HTTP CONNECT proxy: that path ignores the request's `sni_hostname` extension and derives the TLS server name from the rewritten remote origin, so it sends the IP as SNI. I reproduced this with `HTTPS_PROXY` set: a normal `httpx` GET to `https://example.com` returns 200, while this PR's `fetch_web_content` path fails with `SSLV3_ALERT_HANDSHAKE_FAILURE`. All three callers construct proxied clients via `get_proxy_url`, so proxy deployments lose HTTPS page, SVG, and asset fetching. Please use a proxy-compatible pinning transport (or explicitly delegate DNS policy to a trusted proxy) and cover a real CONNECT-proxy request rather than only `MockTransport`.

src/xagent/web/api/files.py
```diff
@@ -117,6 +120,74 @@ def _preview_can_redirect(path: Path, media_type: str) -> bool:
...
+) -> FileResponse:
+    """Build the final inline preview FileResponse, rasterizing SVG to PNG.
+
+    Raw SVG bytes are never served inline — an embedded ``<script>`` would
+    execute on direct top-level navigation to this response. Every other
+    media type is served as-is.
+    """
+    if media_type == "image/svg+xml":
+        try:
+            png_path = _rasterize_svg_preview(path, file_id=file_id)
```
qinxuye: [P1] Move SVG rasterization off the async request loop

`preview_file` and `public_preview_file` are async endpoints, but this helper synchronously reads the SVG and runs CairoSVG before returning the `FileResponse`. A bounded 5 MB SVG can still contain expensive paths/filters and render up to the 4096×4096 pixel budget, so one first-time preview can block every request handled by that event loop; the public preview path makes this externally triggerable. Please perform the cache miss/rasterization through `asyncio.to_thread` or a bounded worker (and keep cache writes coordinated) rather than running it inline in the async route.

src/xagent/web/api/files.py
```diff
@@ -117,6 +120,74 @@ def _preview_can_redirect(path: Path, media_type: str) -> bool:
...
+    """Return the on-disk cache path for a rasterized SVG preview.
+
+    Mirrors ``_pptx_pdf_cache_path``'s cache-dir-outside-uploads pattern.
+    """
+    import hashlib
+
+    cache_dir = get_storage_root() / "svg_png_cache"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    if file_id:
+        return cache_dir / f"{file_id}.preview.png"
```
qinxuye: [P1] Include the preview target in the SVG cache key

`public_preview_file` accepts a base `file_id` plus different `relative_path` assets, but every one of those SVGs reaches this helper with the same base `file_id`. The first rendered asset therefore populates `<file_id>.preview.png`, and later assets reuse it whenever their mtime predates the cache. I reproduced two different SVG paths with one file ID and the second preview returned the first SVG's cached bytes. Key the cache by the resolved target/asset identity as well as the base file ID (or pass the registered asset's own ID) so relative previews cannot alias each other.

src/xagent/web/api/files.py
```diff
@@ -117,6 +120,74 @@ def _preview_can_redirect(path: Path, media_type: str) -> bool:
...
 
 
+def _svg_png_cache_path(svg_path: Path, file_id: Optional[str] = None) -> Path:
+    """Return the on-disk cache path for a rasterized SVG preview.
+
+    Mirrors ``_pptx_pdf_cache_path``'s cache-dir-outside-uploads pattern.
+    """
+    import hashlib
+
+    cache_dir = get_storage_root() / "svg_png_cache"
```
qinxuye, line 144: [P2] Remove derived SVG previews when the source is deleted

This cache lives outside the uploads tree, but unlike the adjacent PPTX preview cache it has no lifecycle cleanup. `UploadedFileStore.delete()` and the HTTP delete route only call `delete_pptx_pdf_cache`, so rasterized user content remains on disk indefinitely after its source row, local file, and durable object are removed. Please add an SVG cache deletion helper and invoke it from the same centralized deletion paths (or generalize the existing helper to remove all derived previews).

src/xagent/core/tools/adapters/vibe/fetch_web_content.py
```diff
@@ -12,6 +12,28 @@
...
+        description=(
+            "Also discover image, icon, manifest, script, and stylesheet URLs. "
+            "Enable this when looking for logos, brand assets, or images."
+        ),
+    )
+    asset_query: str | None = Field(
+        default=None,
+        description=(
+            "Optional substring such as 'logo' used to filter discovered assets. "
+            "For React SPAs this also searches the official asset manifest."
```
qinxuye: [P2] Keep the LLM-facing asset contract aligned with the implementation

The follow-up commit removed SPA asset-manifest fetching, but this argument still tells the agent that `asset_query` searches the official manifest, and the result schema below makes the same promise. The current implementation only filters references extracted from the fetched HTML (it may return a `<link rel=manifest>` URL, but it never fetches or searches that manifest). This can make the agent plan around behavior that no longer exists. Please remove the manifest-search claims from both descriptions, or restore the documented behavior.

src/xagent/core/utils/svg.py
```diff
@@ -0,0 +1,88 @@
...
+    if not svg_bytes:
+        raise ValueError("SVG content is empty")
+    if len(svg_bytes) > MAX_SVG_BYTES:
+        raise ValueError(f"SVG exceeds maximum size of {MAX_SVG_BYTES} bytes")
+    lowered_all = svg_bytes.lower()
+    lowered_prefix = lowered_all[:4096]
+    if b"<svg" not in lowered_prefix:
+        raise ValueError("SVG content does not contain an <svg> root")
+    if b"<!doctype" in lowered_all or b"<!entity" in lowered_all:
+        raise ValueError("SVG declarations and entities are not supported")
```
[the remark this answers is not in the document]
rogercloud: Fixed in ba610ec9: replaced the literal substring check with `re.search(rb"<!\s*(?:doctype|entity)\b", lowered_all)`, which also rejects whitespace-obfuscated variants like `<! DOCTYPE` or `<!\nENTITY`. Added regression tests in `tests/core/utils/test_svg.py` covering whitespace/newline/mixed-case bypass attempts plus the original exact-match cases.

src/xagent/core/utils/security.py
```diff
@@ -50,10 +67,128 @@ def reject_private_network_host(hostname: str) -> None:
...
+    content_length_name: str | None = None,
+    require_non_empty: bool = False,
+) -> PublicHttpResponse:
+    """Fetch a bounded HTTP body, validating every redirect target before I/O."""
+
+    current_url = url
+    display_name = resource_name[:1].upper() + resource_name[1:]
+    length_name = content_length_name or resource_name
+    for redirect_count in range(max_redirects + 1):
+        await validate_public_http_url(current_url)
```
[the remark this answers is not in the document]
rogercloud: For direct (non-proxied) requests this is already closed: `_pin_url_to_ip` connects to the exact IP that `validate_public_http_url` resolved and validated, with TLS SNI/Host pinned to the original hostname (see `fetch_public_http_bytes` in `security.py`), so there's only one DNS resolution in the request path.

For proxied requests (see the `via_proxy` mode added in 85f1171d, discussed on the security.py:185 thread), we deliberately do not rewrite the connect URL to the pinned IP: httpcore's HTTP CONNECT path ignores the `sni_hostname` extension, so pinning there would send the resolved IP as TLS SNI and break SNI-strict origins. In that mode we rely on the upfront `validate_public_http_url` check as a best-effort gate; the actual outbound connection is made by the configured proxy, so egress-level SSRF policy for private ranges is inherently the proxy's responsibility once a proxy is in the path. This is a deliberate scope limit rather than an oversight — open to revisiting if we want to require a DNS-pinning-capable proxy transport instead.

src/xagent/core/tools/adapters/vibe/download_web_asset.py
```diff
@@ -0,0 +1,228 @@
...
+    ) -> str:
+        media_type = cls._media_type(content_type)
+        url_suffix = Path(urlsplit(url).path).suffix.lower()
+        if media_type and not media_type.startswith("image/"):
+            raise ValueError(f"Unsupported web asset content type: {content_type}")
+        if media_type == "image/svg+xml" or url_suffix == ".svg":
+            prefix = content[:4096].decode("utf-8", errors="ignore").lower()
+            if "<svg" not in prefix:
+                raise ValueError("Remote SVG asset does not contain an <svg> root")
+            return ".svg"
```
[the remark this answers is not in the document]
rogercloud: This path already calls `validate_svg_bytes` (see the `_validate_image` helper in this file), which rejects `<!doctype`/`<!entity` declarations — and after ba610ec9 that check is regex-based and no longer bypassable via whitespace obfuscation (see the sibling thread on svg.py:31). No separate check is needed in this file.

src/xagent/core/tools/core/vision_tool.py
```diff
@@ -160,11 +165,113 @@ def _convert_image_to_base64(self, image_path: str) -> str:
...
+                mime_type = mimetypes.guess_type(image_path)[0]
+                if mime_type != "image/svg+xml" and not image_path.lower().endswith(
+                    ".svg"
+                ):
+                    return None, None
+                source = await asyncio.to_thread(
+                    Path(image_path).read_text,
+                    encoding="utf-8-sig",
+                    errors="replace",
+                )
```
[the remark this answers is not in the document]
rogercloud: Fixed in 2154be6c: the local-file branch of `_svg_source_content` now checks `stat().st_size` against `MAX_SVG_BYTES` before reading, and verifies the first 4096 bytes contain an `<svg` root, matching the guards already in place on the remote/data-URL branches. Added regression tests in `tests/core/tools/test_vision_tool.py` for the oversized-file, missing-root, and normal-file cases.

src/xagent/core/tools/adapters/vibe/download_web_asset.py
```diff
@@ -0,0 +1,228 @@
...
+        detected_extension: str,
+    ) -> str:
+        raw_name = requested or Path(unquote(urlsplit(final_url).path)).name
+        filename = Path(str(raw_name)).name.strip()
+        if not filename or filename in {".", ".."}:
+            filename = "web_asset"
+        if "\x00" in filename:
+            raise ValueError("filename contains an invalid null byte")
+        if not Path(filename).suffix:
+            filename = f"{filename}{detected_extension}"
```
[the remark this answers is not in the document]
rogercloud: Keeping the current silent-repair behavior here rather than switching to a hard reject. The file type actually used downstream (`_mime_type_for_extension`) is derived from the content-validated `detected_extension`, not from the caller-supplied filename — the extension is a display/consistency concern, not a security boundary. A caller guessing the wrong extension is a common, benign situation, and hard-rejecting would turn an easily-recoverable mismatch into a tool-call failure with no real safety upside; the existing JPEG-alias handling reflects the same reasoning. Open to revisiting if there's a concrete case where the silent correction causes a problem.

src/xagent/core/tools/core/web_content.py
```diff
@@ -186,34 +232,186 @@ async def fetch_text(self, url: str) -> str:
...
+                    name=name,
+                    alt=alt,
+                )
+            )
+
+        for image in soup.find_all("img"):
+            add(
+                image.get("src") or image.get("data-src"),
+                kind="image",
+                name=str(image.get("id") or image.get("class") or ""),
```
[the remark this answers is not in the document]
rogercloud: Fixed in 2154be6c: joins the multi-valued `class` attribute with spaces instead of stringifying the list, so discovered assets get a clean `name` like `"logo primary"` instead of `"['logo', 'primary']"`. Added a regression test in `tests/core/tools/test_fetch_web_content.py`.

src/xagent/core/utils/security.py
rogercloud: Fixed in 85f1171d: added a `via_proxy` mode to `fetch_public_http_bytes` — when routing through a proxy we no longer rewrite the connect URL to the pinned IP (which would send that IP as SNI over httpcore's CONNECT path); instead we keep the original hostname and rely on the upfront `validate_public_http_url` check. All three callers (`web_content.py`, `download_web_asset.py`, `vision_tool.py`) now pass `via_proxy=bool(proxy_url)`. Added coverage in `tests/core/model/test_security.py` and updated the vision_tool proxy test that was previously asserting the old (buggy) rewrite-to-IP behavior.

src/xagent/web/api/files.py
rogercloud: Fixed in 85f1171d: `_inline_preview_response` is now async and offloads the read+rasterize call via `asyncio.to_thread`, so a cache-miss SVG preview no longer blocks the event loop. Added a regression test in `tests/web/test_svg_preview_cache.py` that asserts a concurrent ticker isn't stalled during rasterization.

src/xagent/web/api/files.py
rogercloud: Fixed in 85f1171d: `_svg_png_cache_path` now includes a hash of the resolved source path in the cache filename even when `file_id` is given, so distinct relative-path assets registered under the same base `file_id` no longer collide on one cached PNG. Added regression tests in `tests/web/test_svg_preview_cache.py` covering two different SVGs under one file_id.

src/xagent/web/api/files.py
rogercloud: Fixed in 85f1171d: added `delete_svg_png_cache(file_id)` (globs and removes every `<file_id>.*.preview.png`) and wired it into both `UploadedFileStore.delete()` and the HTTP delete route, alongside the existing `delete_pptx_pdf_cache` call. Added regression tests in `tests/web/services/test_uploaded_file_store.py`.

src/xagent/core/tools/adapters/vibe/fetch_web_content.py
rogercloud: Fixed in 85f1171d: removed the manifest-search claims from both the `asset_query` parameter description and the `assets` result field description, since the implementation only extracts references from the fetched HTML (which may include a `<link rel=manifest>` URL) and never fetches or searches the manifest itself.

qinxuye (commented):
Follow-up review on the latest head found three remaining correctness and security gaps.

src/xagent/core/utils/security.py
```diff
@@ -50,10 +74,178 @@ def reject_private_network_host(hostname: str) -> None:
...
+
+    logical_url = url
+    display_name = resource_name[:1].upper() + resource_name[1:]
+    for redirect_count in range(_MAX_REDIRECTS + 1):
+        resolved_ips = await validate_public_http_url(logical_url)
+        parsed = urlsplit(logical_url)
+        request_headers = dict(headers or {})
+        stream_extensions: dict[str, str] = {}
+        if via_proxy:
+            connect_url = logical_url
```
qinxuye, line 180: [P1] Do not silently drop DNS pinning for ordinary proxies

In this branch the application validates one DNS answer locally, but the proxy subsequently resolves `logical_url` again. A rebinding or split-horizon hostname can therefore be public during validation and private when the proxy connects, reopening the SSRF TOCTOU window this helper is meant to close. Because every ambient `HTTP(S)_PROXY` automatically enables `via_proxy`, an ordinary proxy silently weakens the helper's security contract, and the `MockTransport` tests do not exercise that second resolver. Please either require an explicitly trusted proxy that enforces private-range egress policy, or use a proxy transport that connects to the validated IP while preserving Host/SNI, and cover the behavior with a real CONNECT proxy.

src/xagent/web/api/files.py
```diff
@@ -117,6 +124,88 @@ def _preview_can_redirect(path: Path, media_type: str) -> bool:
...
+    try:
+        if (
+            cache_path.exists()
+            and cache_path.stat().st_mtime >= svg_path.stat().st_mtime
+        ):
+            return cache_path
+    except OSError:
+        pass
+    png = rasterize_svg_bytes(svg_path.read_bytes())
+    tmp = cache_path.with_suffix(".png.tmp")
```
qinxuye: [P1] Coordinate same-key SVG cache misses

`asyncio.to_thread` allows multiple preview requests to enter this function concurrently, but every request for the same cache key writes the same `.png.tmp` path. With two requests for one uncached SVG, both can finish writing before replacement; the first replacement removes the temp path and the second raises `FileNotFoundError`, which `_inline_preview_response` turns into a 422. I reproduced this deterministically with two concurrent calls. Please use a unique temporary file plus atomic replacement, or a per-key lock, and add a concurrent same-key regression test.

src/xagent/web/api/files.py
```diff
@@ -117,6 +124,88 @@ def _preview_can_redirect(path: Path, media_type: str) -> bool:
...
+    preview request would read back the first asset's PNG. The ``file_id``
+    prefix is kept so ``delete_svg_png_cache`` can still glob-remove every
+    derived preview for a deleted upload.
+    """
+    import hashlib
+
+    cache_dir = get_storage_root() / "svg_png_cache"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    path_key = hashlib.sha256(str(svg_path.resolve()).encode()).hexdigest()[:24]
+    if file_id:
```
qinxuye, line 147: [P2] Clean up legacy path-keyed SVG previews

For a non-UUID legacy file, the preview callers pass `file_id=None`, so this branch creates `<path-hash>.preview.png`. The deletion paths later call `delete_svg_png_cache(file_id)`, which only globs `<file_id>.*.preview.png`; consequently a deleted legacy file's derived cache never matches and remains on disk. Please give legacy entries a stable cleanup prefix or delete their cache by resolved source path as part of the same lifecycle cleanup.

src/xagent/core/utils/security.py
rogercloud: Fixed in 8bc24ef3: added an explicit `XAGENT_TRUSTED_EGRESS_PROXY` opt-in (see `config.py`/`web_content.py::get_trusted_proxy_url()`). Previously any ambient `HTTP(S)_PROXY` was used automatically; now, unless that flag is set, a configured proxy causes `get_trusted_proxy_url()` to raise `PrivateNetworkHostError` instead of silently routing the SSRF-sensitive fetch through it and losing client-side IP pinning. All three call sites (`web_content.py`, `vision_tool.py`, `download_web_asset.py`) now go through this gate. We evaluated a transport that preserves Host/SNI while connecting to the pinned IP through a CONNECT tunnel, but httpcore's tunnel implementation derives SNI from the remote origin and ignores the `sni_hostname` extension, so that would require forking/subclassing httpcore's proxy transport for a fairly fragile gain. The explicit-trust opt-in mirrors the existing `XAGENT_MCP_OAUTH_ALLOW_PRIVATE_HOSTS` pattern in this codebase and keeps the default posture fail-closed. Added coverage in `tests/core/test_config.py` and `tests/core/tools/test_fetch_web_content.py` for both the untrusted-proxy-rejects and trusted-proxy-passes cases.

src/xagent/web/api/files.py
rogercloud: Fixed in 8bc24ef3: reproduced this deterministically (fixed `.png.tmp` name raced under `asyncio.gather`, second `replace()` raising `FileNotFoundError`). `_rasterize_svg_preview` now writes to a per-call unique temp name (pid + uuid4) before `os.replace()` onto the shared cache path — renaming distinct temp files onto the same destination is atomic and needs no extra locking. Added a concurrent same-key regression test in `tests/web/test_svg_preview_cache.py` (20 rounds of two-way `asyncio.gather`, fresh SVG per round to force a real cache miss each time) that reliably reproduced the failure before the fix and passes after.

src/xagent/web/api/files.py
rogercloud: Fixed in 8bc24ef3: `delete_svg_png_cache` now accepts an optional resolved `source_path`; for legacy (non-UUID) files the HTTP delete route passes the resolved source path so the bare `<path-hash>.preview.png` entry (no `file_id` prefix to glob on) gets removed too. `UploadedFileStore.delete()` only ever operates on DB-registered records that always have a `file_id`, so it doesn't need this — legacy files can only reach deletion through the HTTP route, which is covered. Added regression tests in `tests/web/services/test_uploaded_file_store.py` for both the legacy path-hash cleanup and the existing UUID-keyed behavior (unaffected).

qinxuye (commented):
LGTM

tests/core/tools/test_fetch_web_content.py
```diff
@@ -234,3 +367,59 @@ async def test_fetch_webpage_http_error(self, fetch_tool):
     def test_args_validation(self):
         args = FetchWebContentArgs(url="https://example.com")
         assert args.url == "https://example.com"
+        assert args.include_assets is False
+        assert args.asset_query is None
+
+
+class TestGetTrustedProxyUrl:
+    """An ambient HTTP(S)_PROXY must not silently reopen the DNS-rebinding
```
qinxuye, line 375: Non-blocking: please keep these proxy tests isolated from the host environment. The existing fetch tests inherit ambient upper- or lowercase `HTTP_PROXY`/`HTTPS_PROXY` values, so on a proxied developer or CI host they now fail at the new trust gate before reaching their mocked request; I reproduced nine such failures locally. An autouse fixture that clears both proxy casings and `XAGENT_TRUSTED_EGRESS_PROXY`, with individual proxy tests opting back in explicitly, would keep the suite deterministic.

src/xagent/config.py
```diff
@@ -343,6 +344,19 @@ def get_mcp_oauth_allow_private_hosts() -> bool:
...
+    """Return whether the ambient HTTP(S)_PROXY may be used for public fetches.
+
+    Outbound SSRF guarding pins the DNS resolution used at validation time
+    to the one used at connect time. Routing through a proxy breaks that
+    guarantee, since the proxy performs its own, independent DNS resolution
+    of the target host. This flag is an explicit opt-in acknowledging that
+    the configured proxy is trusted to enforce its own private-range egress
+    policy; leave it disabled unless that is true.
+    """
+    return _get_bool_env(TRUSTED_EGRESS_PROXY, False)
```
qinxuye, line 357: Non-blocking: please add `XAGENT_TRUSTED_EGRESS_PROXY` to `example.env` with the same warning described here. Proxy deployments now need this opt-in before webpage fetching, remote SVG inspection, and asset downloads work, and the repository's configuration conventions require new environment variables to be documented in the template.

rogercloud (approved):

tests/core/tools/test_fetch_web_content.py
rogercloud: Fixed in 40113b00 (now fea2aba3 after a rebase onto main): added an autouse `isolate_proxy_env` fixture in `tests/conftest.py` that clears `HTTP_PROXY`/`http_proxy`/`HTTPS_PROXY`/`https_proxy`/`XAGENT_TRUSTED_EGRESS_PROXY` before every test via `monkeypatch.delenv(..., raising=False)` (auto-restored per test). Reproduced the leak locally with a host `HTTP_PROXY` set (failures across `test_fetch_web_content.py`/`test_download_web_asset.py`), confirmed the fixture fixes it, and confirmed no regression with a clean env. Tests that specifically exercise proxy behavior opt back in via `monkeypatch.setenv(...)`.

src/xagent/config.py
rogercloud: Fixed in 40113b00 (now fea2aba3 after a rebase onto main): documented `XAGENT_TRUSTED_EGRESS_PROXY` in `example.env` alongside the other MCP OAuth proxy settings, with the same DNS-rebinding/TOCTOU risk explanation as the `config.py` docstring, defaulting to `false`.

PR #979, opened 2026-07-23
fix: resolve agent logo update caching bug and add logo removal

### LINKED ISSUE #975: bug: Agent logo update does not persist after saving
### Description

Updating an existing agent's logo does not persist after saving.

When a user uploads a new logo for an existing agent and saves the changes, the new logo is not retained.

### Steps to Reproduce

1. Open an existing Agent.
2. Upload a new logo.
3. Save the changes.
4. Refresh the page or reopen the Agent.
5. Check the displayed logo.

### Actual Behavior

- The updated logo is not persisted.
- The previous logo remains displayed after reopening the Agent.

### Expected Behavior

- The newly uploaded logo should be saved successfully.
- The updated logo should remain after refreshing or revisiting the Agent configuration page.

### LINKED ISSUE #976: feat: Allow users to remove an agent logo
### Description

Add the ability for users to remove an existing Agent logo and revert the Agent back to the default state.

### Problem Statement

Currently, users can upload or update an Agent logo, but there is no option to remove an existing logo.

Users should be able to clear the logo if they no longer want to use a custom image.

### Expected Behavior

Users should be able to:

- Remove the current Agent logo.
- Save the Agent without a custom logo.
- See the default logo/avatar after removal.

yiboyasss:
## Summary
- Agent logo uploads were saved to a deterministic filename (`agent_<id>.<ext>`), so replacing a logo produced the exact same URL and browsers kept serving the cached image — the builder also never synced `logo_url` back into state after a successful update, and `_delete_logo` resolved paths against the wrong working directory so old files were never actually cleaned up from disk.
- Added a remove (X) button on the logo preview in the agent builder so a logo can be cleared without uploading a replacement.

## Test plan
- [x] Verified two sequential logo uploads for the same agent now return distinct URLs, and the previous file is deleted from disk
- [x] Verified the builder preview reflects the new logo immediately after clicking Update, including after a full page reload
- [x] Verified clicking the new remove button clears the logo, enables Update, and persists `logo_url: null` after saving
- [x] `pytest tests/web/api/test_agents_management.py` passes
- [x] `tsc --noEmit` passes with no new errors

Closes #975 
Closes #976


rogercloud (commented):
# PR #979 Review — fix: resolve agent logo update caching bug and add logo removal

## Summary

This PR fixes a user-visible bug where replacing an existing agent's logo silently did nothing: uploads were written to a deterministic path (`agent_<id>.<ext>`), so the served URL never changed and browsers kept the stale cached image. The fix makes saved logo filenames unique via a `uuid.uuid4()` suffix so the URL changes on every update, syncs the server-returned `logo_url` back into the builder's local state after save, and resolves logo deletes against the configured uploads root instead of the process CWD so old files are actually removed. It also implements #976 by adding an X button that clears a logo entirely — sending `logo_base64: ""` through the existing `PUT /api/agents/{id}` path so the avatar falls back to the default. A follow-up commit adds a path-containment guard to `_delete_logo`.

Each of the three sub-fixes is the simplest correct approach, and the remove feature correctly reuses the existing update endpoint with an empty-string sentinel rather than introducing a parallel delete endpoint. Two design-level reservations:

- **The `""` sentinel contract is implicit and split across endpoints.** `undefined` = no change, `""` = clear, data-URL = set. `create_agent` uses a truthy check (so `""` is a no-op), while `update_agent` uses `is not None` (so `""` clears). This is intentional and correct for the two flows, but the contract lives only in a frontend code comment (`agent-builder.tsx` ~line 1369) — it is neither documented nor validated server-side.
- **Delete-then-save cleanup logic is duplicated** between `update_agent` and the separate `upload_agent_logo` endpoint in `agents.py`. This is pre-existing (not introduced here), but the PR touches both without consolidating.

## Findings

### MAJOR — No test coverage for anything this PR changes
`tests/web/api/test_agents_management.py` (untouched — confirmed via `git diff --stat`)

`_delete_logo` is fully mocked/monkeypatched in every relevant test (e.g. line 1495 `monkeypatch.setattr("xagent.web.api.agents._delete_logo", logo_calls.append)`; similar at line 1998), so the new path-resolution and traversal-guard logic never executes under test. The one test that references `logo_url` (`test_logo_and_cache_cleanup_observe_committed_delete`, ~lines 1971–2018) hardcodes the *old* deterministic filename pattern `f"/uploads/agent_logos/agent_{target_id}.png"` (line 1977) — it would pass identically with or without the uuid-suffix fix, so it does not exercise the caching fix at all. No test drives `logo_base64=""` (backend or frontend), and no frontend test covers the remove button (`logoRemoved` state, `handleRemoveLogo`) — grep across `*.test.tsx` finds zero matches. The "58 passed" result is real but proves nothing about this PR's new behavior, and this is exactly the class of gap (untested path-traversal-guard + untested cache-fix + untested new UI flow) that's worth blocking on until covered.

Suggested additions:
- Assert two sequential saves for the same agent id produce distinct filenames/URLs (the caching fix).
- Call the real (non-mocked) `_delete_logo` with a crafted traversal-style path and assert it is refused, and that a legitimate in-root path is removed.
- Drive `PUT /api/agents/{id}` with `logo_base64=""` and assert `logo_url` becomes null and the old file is deleted.
- A frontend test for the remove-button state reset.

### MINOR — Security commit framing overstated (not a code defect)
`src/xagent/web/api/agents.py:378-398` (`_delete_logo`), commit 785fa44f

The added guard (resolve against uploads root + `relative_to()` containment check) is correct and good defense-in-depth. However, all four call sites of `_delete_logo` pass a value that always originates from the server-generated `Agent.logo_url` DB column, which is only ever set from `_save_logo`'s return value — `AgentCreateRequest`/`AgentUpdateRequest` expose only `logo_base64`, never `logo_url`, as a settable field. There is therefore no currently-reachable path-traversal vector, and the commit title "prevent path traversal" reads as fixing an externally-triggerable exploit that the codebase already prevented structurally. Recommendation: keep the guard (still valuable against future changes that might loosen this invariant), but consider softening the commit/PR wording so it doesn't read as a CVE-class fix.

### MINOR — Concurrent updates can orphan logo files on disk
`src/xagent/web/api/agents.py`, `update_agent` (~lines 739-747)

`agent.logo_url` is read once at the top of the request (via `get_owned_agent`, called without `for_update=True` — no row lock, no version column on the Agent model) and reused for the later delete, with no re-fetch. Two concurrent PUTs updating the same agent's logo: request A saves fileA and commits `logo_url=A`; request B (which read the pre-A value) deletes the *old* file, saves fileB, commits `logo_url=B` — fileA is now orphaned, referenced by nothing and never cleaned up. This is a genuine behavior change from this PR: at base, deterministic filenames meant concurrent updates raced to overwrite the *same* file (a worse race — mid-write corruption — but no disk leak). The correct caching fix (unique filenames) trades that overwrite race for an orphan-accumulation race. Low severity: disk leak only, the DB always ends consistent with whichever request commits last. Realistic triggers: double-click save, two tabs editing the same agent, naive retry-on-timeout. Non-blocking; worth a follow-up (row locking via `for_update=True` on this path, or a periodic orphan sweep).

### MINOR — Cosmetic false-positive "unsaved changes" state
`frontend/src/components/build/agent-builder.tsx`, `hasChanges` (~lines 1200-1201)

`hasChanges` returns true whenever `logoRemoved` is set, even when net logo state is unchanged — e.g. an agent that never had a logo, where the user uploads a file then clicks remove (`logoFile=null`, `logoUrl=null`, `logoRemoved=true`). Traced end-to-end: saving in this state sends `logo_base64: ""`; `_save_logo("")` returns `None` (falsy short-circuit), and since `agent.logo_url` was already `None` the delete branch is skipped — a harmless no-op DB write (`None -> None`), no data loss or unintended deletion. Purely a UI false-positive that enables the Update button with nothing meaningful to persist. Fine to fix opportunistically (track the original `logoUrl` and compare against the final derived state) or leave as-is.

Lean already. The dedicated over-engineering scan found nothing: `logoRemoved` is a necessary boolean (disambiguates "never had a logo" from "user cleared it"), the filename fix uses `uuid.uuid4()` directly, and the `_delete_logo` rewrite uses `Path.resolve()`/`Path.relative_to()` — the stdlib-idiomatic containment check, not a hand-rolled string-prefix reinvention.

## Test results

`58 passed in 5.32s` in `tests/web/api/test_agents_management.py`. Flagging explicitly (per the MAJOR finding): this green result is misleading as a correctness signal for *this* PR, because none of the logic it introduces — uuid-suffixed filenames, real `_delete_logo` path resolution + traversal guard, or the `logo_base64=""` clear flow — is actually exercised. `_delete_logo` is mocked everywhere, and the one logo test still asserts the old deterministic filename.

## Prior review context

One prior review exists, from the Gemini Code Assist bot (state: COMMENTED), which flagged the path-traversal concern in logo deletion. The PR's third commit (785fa44f) addresses it, and our review confirms the guard is implemented correctly as defense-in-depth. Note the caveat in the MINOR framing finding above: the vulnerability was not actually reachable from client input in the current codebase, so the guard is a sound hardening measure rather than a fix for a live exploit.

## Recommendation

The three sub-fixes are correct and appropriately minimal, and no correctness-blocking bug exists in the new code itself (the one pre-existing issue in this area — delete-before-save with no rollback in `update_agent` — pre-dates this PR and is out of scope here). That said, the MAJOR finding — zero test coverage for security-sensitive path-resolution logic and the actual bug fix this PR claims to make — is a real blocker on its own: please add the tests outlined above before merging. The three MINOR items are non-blocking follow-ups.

rogercloud (approved):
# Follow-up Review — commit 5c9b981d (re-review of PR #979)

## Update since last review

Since the previous round (785fa44f), two commits were pushed: `98295192` adds ~99 lines of tests to `tests/web/api/test_agents_management.py` covering logo re-upload cache-busting, the empty-string clear flow, and a real (non-mocked) path-traversal guard; `5c9b981d` is a ~5-line fix to `agent-builder.tsx`'s `hasChanges` logic. Headline outcome: backend test coverage — the gap that drove last round's MAJOR — is now substantially closed, and the cosmetic false-positive "unsaved changes" bug is fixed. The test suite is at `62 passed` (up from 58).

## Design verdict: SOUND

Upgraded from "acceptable-with-reservations." The reservation last round was driven entirely by the absence of test coverage, not by the design itself; that coverage now exists for the backend (cache-busting, clear flow, and traversal containment all have real end-to-end assertions). The uuid-suffix filename remains the most direct fix for the browser-cache problem (#975), the `""`-vs-`undefined` clear/no-change distinction matches cleanly between frontend and backend (`agent_data.logo_base64 is not None`), and `_delete_logo`'s resolution against `get_uploads_dir().resolve()` matches how `/uploads` is actually mounted.

Two minor design notes (non-blocking): unique-per-upload filenames remove the old "overwrite self-cleans" property, so cleanup now depends entirely on `_delete_logo` succeeding (ties into the new finding below); and the path-traversal guard is more accurately defense-in-depth than a fix for an externally-exploitable bug, since the value passed to `_delete_logo` always originates server-side.

## Prior findings — status

| # | Finding | Status | Reason |
|---|---------|--------|--------|
| 1 | [Gemini, inline `agents.py:396`] Path traversal in `_delete_logo` | **FIXED** | Guard implemented and now has a genuine test: `test_delete_logo_refuses_a_path_traversal_attempt` (test_agents_management.py:1504) calls the real `_delete_logo` with a crafted `/uploads/../` path and asserts the outside file survives; positive companion at :1490. |
| 2 | [MAJOR] No test coverage for anything this PR changes | **PARTIAL (substantially addressed)** | 3 of 4 sub-items now covered with real assertions — see breakdown below. Only the frontend remove-button test is still missing. Downgraded to non-blocking. |
| 3 | [MINOR] Security-commit framing overstated | **WAIVED** | Re-confirmed: `AgentCreate/UpdateRequest` still expose only `logo_base64`, never `logo_url` (agents.py ~85-112). Documentation/framing observation, no code fix warranted. |
| 4 | [MINOR] Concurrent updates can orphan logo files | **NOT FIXED** | `git diff` shows zero changes to `agents.py`/`agent_store.py` in the 2 new commits; `get_owned_agent` still called without `for_update=True` (agents.py:677). Low severity, non-blocking, future follow-up. |
| 5 | [MINOR] Cosmetic false-positive "unsaved changes" | **FIXED** | `5c9b981d` changed `agent-builder.tsx:1201` to `if (logoRemoved && originalData.logo_url) return true`. Code-traced across all 3 scenarios (never-had-logo upload-then-remove now `false`; existing-logo remove still `true`; replace-then-remove still `true`) — no regression. No automated test locks it in (same gap as 2d). |

Finding 2 breakdown:
- (a) Distinct filenames/URLs on re-upload — **FIXED**: `test_logo_reupload_gets_a_fresh_url_and_removes_the_old_file` (:1427) drives two real sequential `PUT`s, asserts URLs differ and the old file is deleted.
- (b) Real non-mocked traversal test — **FIXED**: see finding 1; only `get_uploads_dir` is patched, the real path-resolution logic executes.
- (c) `logo_base64=""` clear flow — **FIXED**: `test_clearing_logo_with_empty_string_nulls_url_and_deletes_file` (:1461) asserts `logo_url is None` and the file is deleted.
- (d) Frontend remove-button/`hasChanges` test — **STILL MISSING**: grep finds zero `*.test.tsx` references to `logoRemoved` or the remove button. Low-effort, non-blocking gap.

## New finding from this pass

**[MINOR — introduced by this PR] `_delete_logo` silently swallows all failures, now permanently orphaning files.** `agents.py:378-398` catches `except Exception as e: logger.error(...)` with no re-raise and no return signal, and none of its ~4 call sites check for failure. At the base commit this was benign: deterministic filenames (`agent_<id>.<ext>`) meant the next upload overwrote the same path (self-cleaning). This PR moves to unique-per-upload names (`agent_{id}_{uuid4().hex[:8]}.{ext}`), so a failed delete now orphans that specific file *permanently* — no future upload reuses that path. Realistic triggers: FS permission errors, read-only/offline storage, disk I/O errors. This is mechanically distinct from the concurrent-update orphan (finding 4) — exception-swallowing vs. lost-update race — though both fall in the "orphan logo files" bucket; keep them separate. Suggested remediation (optional, non-blocking): a periodic sweep of `agent_logos/` for files not referenced by any `agent.logo_url`, and/or surfacing delete failures rather than swallowing them.

_(FYI footnote, rechecked and out of scope: in `update_agent` (~739-746) an invalid replacement `logo_base64` deletes the old file before `_save_logo` fails, silently nulling `logo_url` with a 200. Base-vs-head diff confirms this is byte-for-byte pre-existing and untouched by this PR — same class already excluded last round. Not a blocking finding.)_

## Simplification (delta scope)

`monkeypatch.setattr(agents_api, "get_uploads_dir", lambda: tmp_path)` is repeated verbatim across all 4 new tests (test_agents_management.py:1433, 1466, 1493, 1509). No existing fixture is reusable (the similarly-named `temp_uploads_dir` patches a different module target). Extracting a local `uploads_dir` fixture would remove the duplication. Net line count is roughly neutral — the value is DRY/maintainability (one place to update if `get_uploads_dir`'s import path changes), not a line-count win. Non-blocking.

## Test results

`62 passed` (up from 58; 4 new tests, all passing). Backend logo handling — cache-busting on re-upload, empty-string clear flow, and real traversal containment — is now solidly covered with meaningful (non-vacuous) assertions. The only remaining coverage gap is a frontend regression test for the remove-button / `hasChanges` state (findings 2d and 5), which is verified-correct-by-code-trace but not locked in by an automated test.

## Recommendation: Approve, with non-blocking follow-ups

The MAJOR blocker from last round is resolved for the security- and correctness-critical backend half; what remains (a frontend remove-button test) is low-effort and non-blocking. The one new finding is MINOR, and the NOT-FIXED/WAIVED items are pre-existing low-severity or accepted framing observations. Tracking the following as follow-ups rather than gating this PR:

1. Add the frontend `hasChanges`/remove-button regression test (2d, 5).
2. Address orphaned-logo cleanup holistically — the silent `_delete_logo` failure (new finding) and the concurrent-update race (finding 4) together — e.g. a referenced-file sweep of `agent_logos/`.
3. Optionally extract the `uploads_dir` test fixture (DRY).

Issue #973, opened 2026-07-23
Harden public share channels: per-guest isolation + rate limiting (follow-up to #947/#967)


codeacme17:
## Context

Follow-up to the public share-link channels (Agent share + Workforce share, #947 / #967). These are pre-existing / inherited design gaps in the public share path, surfaced during review of #967. None is a regression introduced by #967 — filing so they get an explicit product decision rather than silent inheritance.

## 1. No per-guest isolation within a single share link (privacy)

`get_task_for_share_context` authorizes a task solely by `Task.id` (a sequential, attacker-guessable integer PK) + owner + agent/workforce id — with **no per-guest claim**. Any visitor holding a valid share JWT for the same shared entity can address another visitor's task id and read/continue that conversation.

Contrast the widget path, which additionally requires `agent_config["guest_id"] == access_context.guest_id`.

- Affects **both** the pre-existing agent-share flow and the new workforce-share flow (workforce mirrors Agent per #947's explicit "mirror Agent's pattern" scope).
- Workforce sessions carry substantially more private multi-turn content, so the blast radius is larger.

**Proposed fix:** issue a per-guest id/session token at share-auth time, stamp it into `agent_config`, and require it in both branches of `get_task_for_share_context` (agent + workforce).

## 2. No rate limiting / abuse control on public auth & task-create

`POST /api/share/auth` and `POST /api/share/chat/task/create` are unauthenticated-visitor endpoints. Each anonymous task-create spawns a full run against the **owner's** model/billing budget, with no rate limit or abuse control. Larger blast radius for workforces (multi-agent runs).

**Proposed fix:** per-share-token and/or per-IP rate limiting on the public auth + task-create endpoints; consider a per-share run quota.

## Notes

- `sync_workforce_run_status` now fires for all workforce tasks on the WS turn path, not only shared ones (guarded/non-fatal). Flagging for awareness; not necessarily action.

Part of #805.

codeacme17:
Scoping note from PR #967 re-review (R1-1): the workforce-share **first-turn attachment fix** introduced a task-less guest upload path (`upload_share_chat_files` with `task_id=None`, workforce-share only). PR #967 lands a cheap mitigation — a per-request file-count cap (`MAX_TASKLESS_SHARE_UPLOAD_FILES = 10`) — but the deeper exposures belong here:

[...]

Issue #946, opened 2026-07-21
feat(workforce): deployment foundation — deployments table, run source & idempotency, archive termination, config fingerprint


codeacme17:
## Description

Foundation layer for exposing Workforce through external channels (Share / Widget / REST API / Webhook). Blocks all channel issues.

Scope (per the design discussion in #805):

1. **Shared `deployments` table** — `owner_type` ∈ {`agent`, `workforce`} + `owner_id` (unique pair), holding `widget_enabled` / `allowed_domains` / `widget_key` / `share_enabled` / `share_token`. Workforce adopts it from day one. Migrating Agent's existing flattened columns onto this table is explicitly **out of scope** (separate follow-up — touches the auth-critical `public_chat_access.py` path).
2. **`create_workforce_run` parameters** — add `source` (threaded to `Task.source`, currently hardcoded `"internal"` at `workforce_runs.py:133`; existing free-form values: `internal/sdk/trigger/widget/shared_link/a2a`) and optional `idempotency_key` with a unique constraint (workforce path only). Surface `run.task.source` in the runs-history serializer.
3. **Archive terminates external sessions** — `archive_workforce` currently only flips status; in-flight runs keep running. Terminate active runs via the existing `PAUSE` task-command infrastructure (`enqueue_task_command` + `mark_workforce_task_status`) and reject new turns for archived workforces at the shared turn-entry point.
4. **Config fingerprint** — snapshot only freezes prompt-building data; worker execution re-reads the live Agent per call. Compute a config fingerprint at run creation, verify at each turn entry, reject on mismatch ("config changed, please start a new session").

Items 3 and 4 live at the same turn-entry code location and are an access-control requirement from day one of external exposure, not a UX nice-to-have.

## Expected result

- Alembic migration for `deployments` + `workforce_runs.idempotency_key`.
- `create_workforce_run(source=..., idempotency_key=...)` with tests (dedup, source threading).
- Archive terminates active runs; archived workforce rejects new turns.
- Fingerprint mismatch rejects the turn with a clear error.

Part of #805.

PR #985, opened 2026-07-23
feat(workforce): widget deployment

### LINKED ISSUE #948: feat(workforce): widget deployment
## Description

Embeddable widget channel for Workforce, building on the Share channel's token pattern plus the existing embed-ticket / allowed-domains machinery in `api/widget.py`.

Scope:

- Workforce-level `widget_enabled` / `widget_key` / `allowed_domains` in the `deployments` table.
- Widget key management endpoints for workforce (get / rotate), gated on `status == active` + `widget_enabled`.
- `api/widget.py`: workforce-key resolution path (embed ticket, origin/domain checks) → `create_workforce_run(source="widget")`. Existing manager-agent rejection stays.
- Frontend: `widget.js` + widget chat page support for workforce keys; deploy dialog generalization for workforce.

## Expected result

- A workforce is embeddable via the widget snippet end-to-end (multi-turn), with domain allowlist enforced and runs recorded with `source = widget`.

Part of #805. Blocked by #946; builds on the Share channel patterns.

codeacme17:
## Description

Embeddable **widget** channel for Workforce, building on the share-link channel (#947) plus the existing embed-ticket / allowed-domains machinery in `api/widget.py`. Workforce widget state lives in the shared `deployments` table (from #946) — **no new migration**.

Part of #805. Closes #948.

## What's included

**Backend**
- `services/deployments.py`: `new_widget_key` (now the single source of truth, re-exported from `agent_store`) and `find_enabled_widget_deployment`.
- `api/workforces.py`: owner-only widget management endpoints — `GET /{id}/widget-key` (read, no lazy mint), `PUT /{id}/widget` (enable flag + allowed_domains, mints key on enable, requires active), `POST /{id}/widget-key/rotate` (preserves enabled). Gated via `can_edit_workforce` / `ensure_workforce_access`, mirroring the share-link endpoints.
- `api/widget.py`: resolve a widget key to an agent **or** a workforce; embed ticket carries an `owner_type` discriminator (agent tickets stay backward compatible); workforce auth mints a guest JWT carrying `widget_workforce_id` + `widget_key` so rotate/disable invalidates live guests. Manager-agent rejection stays.
- `api/public_chat_access.py`: widget guest path now supports workforces — `ensure_widget_workforce_available` (re-checks active + widget_enabled + key), per-guest workforce task scoping, task-less opening-message upload (with the same `MAX_TASKLESS_SHARE_UPLOAD_FILES` cap as the share path, #973), and guest task creation entering through `create_workforce_run(source="widget")`.

**Frontend**
- `WorkforceWidgetDialog` on the workforce detail page (reuses the shared widget-config helpers), `workforces-api` widget calls, types, en/zh i18n. The public chat page already supports a `workforce_id` auth result.

## Result

A workforce is embeddable via the widget snippet end-to-end (multi-turn), with the domain allowlist enforced and runs recorded with `source = widget`.

## Tests

New `tests/web/api/test_workforce_widget.py` (18 cases): management lifecycle & gating, embed-ticket + auth resolution (direct visit and ticket), domain allowlist, manager-agent rejection, rotate/disable invalidating existing guest tokens, task creation via `create_workforce_run(source="widget")`, task scoping, and the task-less upload cap. Full web-api suite passes; ruff / mypy / tsc / eslint clean.

## Notes

Stacked on #947 (merged). Rebased onto the squashed share-link change on `main`, so this diff is the widget delta only.


rogercloud:
please resolve the conflicts

src/xagent/web/api/public_chat_access.py
```diff
@@ -283,9 +338,60 @@ def dependency(
...
+) -> Task:
+    """Scope a widget guest to a workforce-widget task.
+
+    Mirrors :func:`_get_task_for_workforce_share_context` but retains the
+    widget channel's per-guest isolation: the task must belong to this guest
+    (``guest_id``), target this workforce, and be backed by a real
+    ``WorkforceRun`` row (so a guest cannot address an arbitrary task id).
+    """
+    widget_workforce_id = access_context.widget_workforce_id
+    assert widget_workforce_id is not None
```
rogercloud: Minor: this uses `assert widget_workforce_id is not None` at an auth boundary. The sibling share-channel function in this same file, `_get_task_for_workforce_share_context`, deliberately uses `raise HTTPException(403, ...)` instead, with a comment explaining that `python -O` strips asserts and that an auth boundary must fail closed for any future unguarded caller. The current call site is guarded upstream (`access_context.widget_workforce_id is not None`), so this isn't exploitable today, but it silently drops the fail-closed convention this file just established one function away. Suggest replacing with `raise HTTPException(status_code=403, detail=...)` to match the sibling pattern (same applies to the `assert` at line 653).

src/xagent/web/api/workforces.py
```diff
@@ -998,6 +999,128 @@ async def disable_workforce_share_link(
...
+async def update_workforce_widget(
+    workforce_id: int,
+    request: WorkforceWidgetUpdateRequest,
+    db: Session = Depends(get_db),
+    user: User = Depends(get_current_user),
+) -> WorkforceWidgetResponse:
+    """Update a workforce's widget configuration (enable flag / allowed
+    domains). Enabling requires an active workforce and lazily mints a widget
+    key so the embed snippet is immediately usable."""
+    workforce = ensure_workforce_access(
```
rogercloud: Minor: this endpoint is the only way to disable the widget, and `ensure_workforce_access(action="edit")` raises 409 for archived workforces — so an owner can't turn `widget_enabled` off once a workforce is archived. `DELETE /{id}/share-link` deliberately uses the softer `can_edit_workforce` gate instead, specifically so archived workforces can still have their share link revoked. Not a security gap (`ensure_widget_workforce_available` independently blocks guest access once `status != "active"`, regardless of the stale flag), but worth considering the same archived-safe gate here for consistency with the share-link disable path.

rogercloud (approved):
## Summary

This PR adds an embeddable widget channel for Workforce, closely paralleling the existing agent-widget embed machinery and the recently merged workforce share-link channel (#947). Widget state (enabled flag, key, allowed domains) is stored in the pre-existing shared `deployments` table, so no migration is needed. On the backend it adds owner-only widget management endpoints (get key, enable+configure, rotate) in `api/workforces.py`, extends `api/widget.py`'s embed-ticket/auth flow with an `owner_type` discriminator that resolves either an agent or a workforce, and extends `api/public_chat_access.py` so widget guests can create and scope tasks against a workforce (`create_workforce_run(source="widget")`), including the task-less opening-message upload path. The frontend adds a `WorkforceWidgetDialog` component, API client calls, types, and i18n strings.

## Design verdict

**Sound.** This is a clean, well-factored parallel of the existing agent-widget path and the share-link channel. It reuses shared machinery — `new_widget_key` consolidated into `deployments.py` and re-exported from `agent_store.py` for backward compatibility — rather than duplicating it, and the `owner_type` discriminator / `ResolvedWidgetOwner` abstraction unifies agent-vs-workforce resolution without breaking existing agent embeds (the `owner_type = claims.get("owner_type") or EMBED_TICKET_OWNER_AGENT` fallback plus relaxing the `agent_id` response field to `int | None` keep legacy agent tickets working).

**Design note (not a blocker):** the workforce widget's guest JWT is re-validated on every request against the live deployment (`ensure_widget_workforce_available`), so rotating or disabling the widget immediately cuts off live guests. The pre-existing **agent** widget path does not re-validate per request, so disabling/rotating an agent widget does not invalidate its live guests today. This PR does not regress that behavior — it just makes the inconsistency visible by doing the safer thing on the new path. Worth a follow-up ticket to bring the agent path in line, but out of scope here.

`src/xagent/web/api/workforces.py`:
- L874: shrink: `_ensure_shareable_workforce` and `_ensure_widget_configurable_workforce` (L1023) are identical except for the `HTTPException` detail string. Merge into one `_ensure_active_workforce(workforce, detail)` helper.
- L1023: shrink: same as above — reuse the shared `_ensure_active_workforce` helper at all 4 call sites (share ~L935, ~L964; widget ~L1089, ~L1115).

net: -7 lines possible

## Verification

- `tests/web/api/test_workforce_widget.py`: 18/18 passing. Coverage is strong on the management and auth surface — owner-only gating, archived/inactive rejection, rotate-preserves-disabled-flag, domain allowlist enforcement, rotate-and-disable both invalidating live guest JWTs, `source="widget"` on created tasks, and the task-less upload cap.
- `ruff check`: clean on all changed backend files.
- `mypy`: clean on all changed backend files.

Minor test gaps (not blocking): no direct test for cross-guest / cross-workforce task-access rejection in the widget context specifically (only exercised transitively via the invalidation tests), and no test for the legacy agent-embed-ticket-without-`owner_type` backward-compat fallback path.

Both inline findings below are minor and non-blocking given the current call sites are already guarded; flagging for consistency with this file's own conventions.

codeacme17:
Thanks for the thorough review @rogercloud — all of your non-blocking feedback is addressed in `8e19f36c`:

**Inline findings**
- **`assert` at the auth boundary** (`public_chat_access.py`): replaced the `assert widget_workforce_id is not None` at both widget sites (`_get_task_for_workforce_widget_context` and `_create_workforce_widget_chat_task`) with an explicit `raise HTTPException(403, ...)`, so they fail closed under `python -O` — matching the fail-closed convention `_get_task_for_workforce_share_context` established one function away.
- **Archived-safe widget disable** (`workforces.py`): `PUT /{id}/widget` now gates on `can_edit_workforce` (mirroring `DELETE /{id}/share-link`) instead of `ensure_workforce_access(action="edit")`, so an owner can still turn the widget off on an archived workforce. Re-enabling stays blocked by `_ensure_active_workforce`, and guest access remains independently blocked once `status != "active"`.

**Simplification**
- Merged `_ensure_shareable_workforce` and `_ensure_widget_configurable_workforce` into a single `_ensure_active_workforce(workforce, detail)` helper, reused at all four share/widget enable+rotate call sites.

**Test gaps**
- Added: widget disable-on-archived, cross-guest / cross-workforce task-access rejection in the widget context, and the legacy agent embed-ticket-without-`owner_type` backward-compat fallback.

Verification: `test_workforce_widget.py` (21) + `test_workforce_share_link.py` = 46 passing; `ruff` + `mypy` clean.

On the **design note** (agent widget path not re-validating per request, so disabling/rotating an agent widget doesn't cut off live guests): agreed it's out of scope here — I'll open a follow-up issue to bring the agent path in line with the per-request revalidation the workforce path now does.

src/xagent/web/api/public_chat_access.py
codeacme17: Fixed in `8e19f36c`. Replaced the `assert` here (and the twin in `_create_workforce_widget_chat_task`) with `raise HTTPException(status_code=403, detail="Widget is unavailable")` so both widget auth boundaries fail closed under `python -O`, matching the fail-closed pattern in `_get_task_for_workforce_share_context`.

src/xagent/web/api/workforces.py
codeacme17: Fixed in `8e19f36c`. `PUT /{id}/widget` now gates on the archived-safe `can_edit_workforce` (like `DELETE /{id}/share-link`) instead of `ensure_workforce_access(action="edit")`, so an owner can turn the widget off on an archived workforce. Re-enabling stays blocked by `_ensure_active_workforce`, and `ensure_widget_workforce_available` still blocks guests once `status != "active"`. Added a `test_widget_disable_allowed_for_archived_workforce` case.

PR #990, opened 2026-07-23
feat: managed SSH MCP tools with sandboxed execution

bluefish-08:
## Summary

Adds a managed SSH MCP capability: agents can run commands and transfer files
(SFTP) on authorized remote hosts through first-class SSH tools, without the
private key ever touching tool arguments, environment, serialization, or logs.
SSH credentials and targets are treated as owned, versioned, authorizable
resources. This includes the tool schemas, the execution engine, the security
policy, and the injection seams for credential resolution and audit.

## What's included

- **Managed tools**: `ssh_list_targets`, `ssh_execute`, `ssh_upload`,
  `ssh_download`. Emitted only when a provider is installed and the executing
  agent has at least one bound target.
- **Execution engine** (`core/ssh/`): `SshExecutor` orchestrates one call —
  resolve binding → capability check → fail-closed egress pre-flight → decrypt
  one credential version → materialize for this call only → run → cap output →
  clean up (on success, error, timeout, cancellation).
- **Two runners behind one seam**:
  - `AsyncsshRunner` — in-process (self-hosted), strict host-key verification,
    public-key-only, no agent/forwarding/user-config, peer-IP re-check.
  - `SandboxSshRunner` — runs the OpenSSH binary inside a leased, per-task
    sandbox; connects to the egress-authorized IP with `HostKeyAlias` (DNS
    rebinding defense), clamps runtime with `timeout`, stages SFTP through the
    sandbox. The private key is materialized into the sandbox and never touches
    the backend disk; capacity exhaustion fails closed (no host fallback).
- **Injection seams (Protocols)** installed via process hooks:
  `SshTargetProvider`, `SshSecretStore`, `SandboxSecretMaterializer`,
  `SshAuditSink`.
- **Egress policy**: denies loopback, link-local, private, and cloud-metadata
  ranges by default; `XAGENT_SSH_ALLOW_CIDRS` allowlists networks. Resolves all
  A/AAAA addresses and authorizes each before connecting.
- **Host-key verification**: strict, against a pinned `known_hosts`; no TOFU.
- **Audit**: an injectable, best-effort sink records one event per operation
  (success and failure) with no secret material.
- **Agent builder**: an inline SSH-bindings section, shown only in deployments
  that expose the team/SSH API.

## Security properties

- Private key / passphrase never in env, argv, tool serialization, or logs.
- Exactly one credential version materialized per call.
- Host-key mismatch fails before authentication.
- Egress and sandbox-capacity both fail closed.
- Secret files cleaned on every exit path; sandbox teardown is the backstop.

## Dependencies / deployment notes

- `asyncssh` is an optional dependency, imported lazily — tools and environments
  that don't run SSH are unaffected.
- The sandbox image must include `openssh-client` (added to
  `docker/Dockerfile.sandbox`); the sandbox runner execs `ssh`/`sftp`/`timeout`.

## Testing

Unit + integration coverage under `tests/core/ssh/` (against a local asyncssh
server), including sandbox-path tests: strict host-key failure before command,
SFTP round-trip, argv security contract, exit-code mapping, and secret cleanup.


rogercloud (commented):
## Summary

This PR adds a managed SSH MCP capability: four new tools (`ssh_list_targets`,
`ssh_execute`, `ssh_upload`, `ssh_download`) backed by a new `core/ssh/` execution
engine. `SshExecutor` orchestrates a clear pipeline
(resolve → capability-check → egress-preflight → credential-decrypt → materialize →
run → cap-output → cleanup) behind a single runner seam, with two implementations:
`AsyncsshRunner` for in-process/self-hosted execution and `SandboxSshRunner` for
sandboxed/cloud execution over the OpenSSH binary. It also ships an IP egress policy
(deny loopback/link-local/private/metadata by default, with an allowlist override),
strict host-key verification, an injectable audit sink, and a frontend
agent-builder SSH-bindings section. ~4850 lines across 39 files, mostly new code.

_This is a re-review: one prior review already existed on this PR; a fresh full pass
was performed over the current head, and all 293 SSH-related tests pass in the review
worktree._

## Overall verdict: Acceptable with reservations

The layering is sound and testable — pure egress decision logic is cleanly separated
from I/O, the runner seam is expressed as `@runtime_checkable` Protocols, materialize-
and-shred is a per-call context manager, and the sandbox lease correctly spans both
materialize and run. The two-runner split is coherent rather than duplicative.

That said, several concrete security gaps undercut the PR's own stated security
properties and should be addressed before merge:

- **The strongest confinement claim (remote-path scoping under `remote_root`) is
  defeatable.** The sandbox runner builds its SFTP batch file by unescaped string
  interpolation — a crafted `remote_path` injects a second, independent SFTP command,
  enabling arbitrary remote read/write outside the intended root (Critical #1).
- **The two runners do not fail the same way.** `AsyncsshRunner` lets auth/connection
  failures escape as raw exceptions, breaking both the "one audit event per operation"
  claim and the tool error contract (Major #2, #6).
- **Output is fully buffered before it is capped**, a host-memory DoS vector (Major #3).
- **The documented egress defenses are weaker than described** — the metadata deny is
  overridable by a broad allowlist, and the tunnel-address backstop doesn't actually
  decode what it claims to (Minor #4, #5).

None of these are architectural — they are localized fixes — but #1 is a genuine
confinement bypass and should block merge.

---

## Critical

### 1. SFTP batch-file command injection via `remote_path`
`src/xagent/core/ssh/sandbox_runner.py` (upload ~L195-198, download ~L233)

Unescaped f-string batch construction; `_constrain_remote_path` (executor.py
~L307-326) rejects neither `"` nor `\n`, and `TransferArgs` has no validator. PoC:
`/allowed/root/x"\nget "/etc/shadow" "/tmp/exfil\n#` passes validation and becomes a
second SFTP command once embedded in the batch file. Only the sandboxed/production
runner is exposed — `AsyncsshRunner` uses asyncssh's programmatic sftp client.

**Fix:** reject/escape `"` and `\n` in `staged`/`remote_path` (ideally in a
`TransferArgs` validator), or drive SFTP non-textually with proper quoting.

---

## Major

### 2. In-process runner leaves auth/connection failures unwrapped
`src/xagent/core/ssh/runner.py` (`_connect`, ~L121-148)

Only `HostKeyNotVerifiable` is caught; `PermissionDenied` etc. are sibling
`DisconnectError` subclasses and propagate raw, skipping the failure audit
(executor.py) and escaping the tool wrapper (ssh_tools.py) instead of becoming
`SshOpResult(ok=False)`. `SandboxSshRunner` maps these deterministically — the
runners diverge on an ordinary path (real auth failures).

**Fix:** catch `asyncssh.Error`/`OSError`/`DisconnectError` broadly and map to the
right `SshErrorCode`.

### 3. Unbounded output buffering before the cap (host-memory DoS)
`src/xagent/core/ssh/runner.py` (~L172-175), `sandbox_runner.py` (~L152)

Both runners fully buffer stdout/stderr before the 1 MiB post-hoc cap; asyncssh has
no native byte limit, and the sandbox exec buffers in the docker-py host client, not
the container (so container cgroup limits don't help). The 10-minute timeout doesn't
bound volume.

**Fix:** enforce the byte cap during the read, not after full buffering.

---

## Minor — Design & Security Hardening

### 4. `XAGENT_SSH_ALLOW_CIDRS` can override the cloud-metadata deny
`src/xagent/core/ssh/egress.py` (~L99 allow-check before ~L102-109 deny-checks)

Reasonable for private/loopback overrides, but the metadata address
(169.254.169.254) is uniquely dangerous and arguably shouldn't be overridable.
`test_sandbox_runner.py` itself uses `allow_cidrs=("0.0.0.0/0",)`.

**Fix:** move the metadata deny ahead of the allowlist check.

### 5. Tunnel-address (NAT64/6to4) backstop is ineffective and single-runner
`egress.py` comment ~L92-94; `runner.py` `_authorize_peer` ~L256-267

The re-check exists only in `AsyncsshRunner` and re-invokes the same non-decoding
`check_ip` — it never actually decodes the tunnel-embedded address it claims to guard
against, and `SandboxSshRunner` (production path) has no equivalent at all.

**Fix:** decode NAT64/6to4 addresses in `check_ip` itself; add an equivalent control
to `SandboxSshRunner`.

### 6. Audit skipped on cancellation and non-`SshError` exceptions
`executor.py` (execute/upload/download)

Same root cause as #2 — audits only fire inside `except SshError`; `CancelledError`
is a `BaseException` and would never match anyway.

**Fix:** guarantee a failure audit as part of #2's remediation.

### 7. Sandbox runner mislabels transport failures as `CONNECTION_TIMEOUT`
`sandbox_runner.py` (`_raise_for_transport`, ~L105-116)

Exit 255 → `CONNECTION_TIMEOUT` for any non-host-key failure, including
connection-refused/unreachable/auth failures that didn't time out.

**Fix:** use a generic `CONNECTION_FAILED` or sub-classify by stderr.

### 8. Silent SSH-target load failure drops the "ssh" tool category on save
`frontend/src/components/build/agent-ssh-bindings.tsx` (`load()` catch, ~L73-85)

Fetch failures are swallowed without calling `onCount`; saving during a transient
failure silently omits "ssh" from tool categories even though bindings exist
server-side.

**Fix:** surface the error (or retry) instead of treating failure as "no bindings."

### 9. Unguarded `os.rmdir` in cleanup `finally` can mask the real exception
`src/xagent/core/ssh/materializer.py` (~L43-47)

`unlink` is wrapped in `suppress(OSError)`, `rmdir` is not — can replace a real
propagating exception with a confusing `ENOTEMPTY`.

**Fix:** wrap `os.rmdir` in `suppress(OSError)` too.

### 10. Injection-seam `cast()` gives no runtime guardrail
`ssh_tools.py` (~L426); hook setters in `ssh_runtime.py`

No `isinstance` check even though the Protocols are `@runtime_checkable`. Low
severity (trusted first-party wiring), but cheap to harden.

---

`src/xagent/core/tools/adapters/vibe/ssh_tools.py`
- L83-203: shrink `_SshToolBase`/`_SshTransferTool`/`SshExecuteTool` duplicate
  `category`/`run_json_sync`/`return_type` boilerplate across 3 classes. Slim
  `_SshToolBase.__init__` to take only `context` and have all SSH tool classes extend it.

net: ~20-30 lines possible

---

## Testing

All 293 SSH-related tests pass (python3.11 venv, review worktree). Coverage is
otherwise strong but has gaps tied directly to the findings above: no test for a
quote/newline `remote_path` (#1), no test for in-process auth-failure → clean
`SshOpResult` (#2), no output-memory-bound test (#3).

---

_Also considered and ruled out:_ the "empty wrapper div" claim in `agent-builder.tsx`
is a false positive; the known-hosts cross-runner format concern isn't a bug (asyncssh
falls back to bare-hostname matching) though the docstring could specify the expected
format; the `preferred_auth` naming and `sandbox_materializer.py` UTF-8 decode
assumption were checked and are fine.

src/xagent/core/ssh/sandbox_runner.py
```diff
@@ -0,0 +1,320 @@
...
+                username=username,
+                private_key_path=private_key_path,
+                known_hosts_path=known_hosts_path,
+                remote_path=remote_path,
+            ):
+                raise SshError(
+                    SshErrorCode.OPERATION_NOT_ALLOWED,
+                    "remote destination already exists",
+                )
+            # sftp batch mode isn't shell-parsed: it quotes paths with double
```
rogercloud, line 203: **Critical: SFTP batch-file command injection.** This batch line is built by unescaped f-string interpolation, and `_constrain_remote_path` (executor.py ~L307-326) rejects neither `"` nor newline (nor does the `TransferArgs` model). A crafted `remote_path` like `/allowed/root/x"\nget "/etc/shadow" "/tmp/exfil\n#` becomes a second independent SFTP command, enabling arbitrary remote read/write outside `remote_root`. The download path at ~L233 has the same flaw; the in-process `AsyncsshRunner` is unaffected because it uses programmatic `sftp.put`/`sftp.get`. Fix: reject/escape `"` and `\n` (ideally in a `TransferArgs` validator so both runners are covered), or drive SFTP via a non-textual invocation with proper quoting.

src/xagent/core/ssh/runner.py
```diff
@@ -0,0 +1,275 @@
...
+        private_key_path: str,
+        known_hosts_path: str,
+        egress_config: EgressPolicyConfig,
+    ) -> AsyncIterator[asyncssh.SSHClientConnection]:
+        # Imported lazily so merely loading this module (and thus the SSH MCP
+        # tools) never hard-requires asyncssh: agents that don't execute SSH,
+        # and environments without the dep, still load every other tool.
+        import asyncssh
+
+        try:
```
rogercloud, line 125: **Major: unwrapped auth/connection errors.** `_connect` only catches `HostKeyNotVerifiable`. In asyncssh, `PermissionDenied` and `HostKeyNotVerifiable` are sibling subclasses of `DisconnectError`, so auth/connection-refused/unreachable errors escape as raw asyncssh exceptions. Combined with the `except SshError`-only handling in executor.py (~L133/185/226) and ssh_tools.py (~L192/236/271), this skips the failure audit and escapes the tool wrapper instead of returning `SshOpResult(ok=False)` — on an ordinary path (real auth failures). `SandboxSshRunner` maps these deterministically, so the runners diverge. Fix: catch `asyncssh.Error`/`OSError`/`DisconnectError` and map to the right `SshErrorCode`.

src/xagent/core/ssh/runner.py
```diff
@@ -0,0 +1,275 @@
...
+    ) -> SshRunResult:
+        async with self._connect(
+            hostname=hostname,
+            port=port,
+            username=username,
+            private_key_path=private_key_path,
+            known_hosts_path=known_hosts_path,
+            egress_config=egress_config,
+        ) as conn:
+            try:
```
rogercloud, line 196: **Major (DoS): unbounded output buffering before the cap.** `conn.run` fully buffers stdout/stderr before `_cap_outputs` (1 MiB) truncates it — asyncssh has no native byte limit. A cheap unbounded-output command (`yes`, `cat /dev/zero | base64`) can exhaust host memory well within the 10-minute timeout. The sandbox path (sandbox_runner.py ~L152) has the same issue since `exec_run` buffers in the docker-py host client, not inside the container. Fix: enforce the byte cap during the read rather than after full buffering.

src/xagent/core/ssh/egress.py
```diff
@@ -0,0 +1,115 @@
...
+    # Normalize IPv4-mapped IPv6 (e.g. ::ffff:127.0.0.1) to its IPv4 form so
+    # dual-stack representations classify identically to the bare IPv4 address.
+    # NOTE: tunnel-embedded IPv4 (NAT64 64:ff9b::/96, 6to4 2002::/16) is NOT
+    # decoded here; the executor (Phase 3) re-checks the actual connected peer
+    # IP, which closes that class of gap.
+    if isinstance(addr, ipaddress.IPv6Address) and addr.ipv4_mapped is not None:
+        addr = addr.ipv4_mapped
+
+    # Explicit allowlist wins over all deny rules.
+    if _in_any(addr, config._allow_networks):
```
rogercloud: **Minor (defense-in-depth): allowlist overrides the metadata deny.** This allow check runs before the metadata deny (~L102-109), so `XAGENT_SSH_ALLOW_CIDRS` can override the cloud-metadata deny (169.254.169.254 / fd00:ec2::254). `test_sandbox_runner.py` uses `allow_cidrs=("0.0.0.0/0",)`, showing how easily this is misconfigured. Recommend moving the metadata deny ahead of the allowlist, or excluding metadata addresses from allow matching.

src/xagent/core/ssh/egress.py
```diff
@@ -0,0 +1,115 @@
...
+def check_ip(ip: str, config: EgressPolicyConfig) -> EgressDecision:
+    """Decide whether ``ip`` may be connected to under ``config``."""
+    try:
+        addr = ipaddress.ip_address(ip)
+    except ValueError:
+        return EgressDecision(False, "invalid ip address")
+
+    # Normalize IPv4-mapped IPv6 (e.g. ::ffff:127.0.0.1) to its IPv4 form so
+    # dual-stack representations classify identically to the bare IPv4 address.
+    # NOTE: tunnel-embedded IPv4 (NAT64 64:ff9b::/96, 6to4 2002::/16) is NOT
```
rogercloud: **Minor: tunnel-address backstop is ineffective and single-runner.** This comment relies on the executor re-checking the peer IP to close the NAT64/6to4 tunnel-address gap, but that re-check (`_authorize_peer`, runner.py ~L256-267) exists only in `AsyncsshRunner` (not `SandboxSshRunner`) and re-invokes the same non-decoding `check_ip` — so it never actually decodes the tunnel-embedded address this comment names. Fix: decode NAT64/6to4 addresses in `check_ip` itself, and add an equivalent egress control to `SandboxSshRunner`.

src/xagent/core/ssh/sandbox_runner.py
```diff
@@ -0,0 +1,320 @@
...
+            known_hosts_path=known_hosts_path,
+        ),
+        f"{username}@{connect_ip}",
+    ]
+    if command is not None:
+        argv.append(command)
+    return argv
+
+
+def _raise_for_transport(exit_code: int, stderr: str) -> None:
```
rogercloud, line 105: **Minor: mislabeled transport failure.** `_raise_for_transport` maps exit 255 to `CONNECTION_TIMEOUT` for any non-host-key transport failure, including connection-refused/unreachable/auth failures that did not time out. Use a generic code (e.g. `CONNECTION_FAILED`) or sub-classify by stderr pattern.

frontend/src/components/build/agent-ssh-bindings.tsx
```diff
@@ -0,0 +1,289 @@
...
+  const [bindings, setBindings] = useState<SshBinding[]>([])
+  const [targets, setTargets] = useState<TargetOption[]>([])
+  const [dialogOpen, setDialogOpen] = useState(false)
+  const [targetPublicId, setTargetPublicId] = useState("")
+  const [capabilities, setCapabilities] = useState<string[]>(["execute"])
+  const [approvalPolicy, setApprovalPolicy] = useState<string>("always")
+  const [submitting, setSubmitting] = useState(false)
+  const { inTeam } = useAuth()
+
+  const load = useCallback(async () => {
```
rogercloud, line 73: **Minor: silent load failure drops the "ssh" category.** This `load()` catch swallows fetch failures without calling `onCount`, so `hasSshBindings` in agent-builder.tsx (driven solely by this callback) stays false. Saving during a transient load failure silently omits the "ssh" tool category even though bindings exist server-side, so SSH tools silently fail to emit. Surface the error (or retry) instead of treating failure as "no bindings."

src/xagent/core/ssh/materializer.py
```diff
@@ -0,0 +1,72 @@
...
+        os.chmod(directory, 0o700)
+        key_path = os.path.join(directory, "id_key")
+        known_hosts_path = os.path.join(directory, "known_hosts")
+        try:
+            _write_private(key_path, credential.private_key)
+            _write_private(known_hosts_path, known_hosts.encode("utf-8"))
+            yield MaterializedSshPaths(
+                private_key_path=key_path, known_hosts_path=known_hosts_path
+            )
+        finally:
```
rogercloud, line 43: **Minor: unguarded `os.rmdir` in cleanup.** `os.unlink` is wrapped in `suppress(OSError)` but the following `os.rmdir` is not. If unlink silently failed (dir non-empty) while an original exception was propagating, rmdir's `ENOTEMPTY` replaces it under `finally` semantics, masking the real error. Wrap `os.rmdir` in `suppress(OSError)` too.

src/xagent/core/tools/adapters/vibe/ssh_tools.py
```diff
@@ -0,0 +1,441 @@
...
+    sandbox_lease = _make_ssh_sandbox_lease(numeric_task_id, agent_id)
+    if sandbox_lease is not None:
+        materializer: Any = SandboxTmpfsSecretMaterializer()
+        runner: Any = SandboxSshRunner()
+    else:
+        materializer = LocalTmpSecretMaterializer()
+        runner = AsyncsshRunner()
+    executor = SshExecutor(
+        provider=provider,
+        secret_store=cast(SshSecretStore, provider),
```
rogercloud: **Minor (hardening): `cast()` gives no runtime guardrail.** `cast(SshSecretStore, provider)` and the `ssh_runtime` hook setters do no `isinstance` check, though the relevant Protocols are `@runtime_checkable`. Add an `isinstance` assert at registration time to fail fast with a clear error instead of an `AttributeError` on the first real SSH call.

src/xagent/core/tools/adapters/vibe/ssh_tools.py
```diff
@@ -0,0 +1,441 @@
...
+    error_code: str | None = None
+    message: str = ""
+    exit_code: int | None = None
+    stdout: str | None = None
+    stderr: str | None = None
+    truncated: bool | None = None
+    duration_ms: int | None = None
+
+
+class _SshToolBase(AbstractBaseTool):
```
rogercloud, line 83: **Simplification:** `_SshToolBase`/`_SshTransferTool`/`SshExecuteTool` (L83-203) re-declare byte-identical `category`, `run_json_sync` (same `NotImplementedError`), and `return_type()`. Slim `_SshToolBase.__init__` to take only `context` and have all SSH tool classes extend it. net: ~20-30 lines possible.

bluefish-08:
Thanks for the thorough re-review. All findings are addressed in `c3a80470`.

**Critical**
1. **SFTP batch injection** — `_constrain_remote_path` now rejects `"` / `\n` / `\r` / `\0` on the shared transfer prologue, closing the bypass at the root for both runners.

**Major**
2. **In-process runner failures** — `_connect` now catches `asyncssh.Error`/`OSError` broadly and maps to a new `CONNECTION_FAILED` code (`SshError` still propagates), so both runners fail identically and the failure audit fires.
3. **Unbounded output** — output is now capped as it is read. The asyncssh runner drives a process and drains stdout/stderr concurrently up to the cap; `Sandbox.exec` gained an optional `max_output_bytes` (default off, zero-regression), with the Docker path using the low-level streaming exec API and Boxlite trimming post-read.

**Minor**
4. Cloud-metadata deny moved ahead of the allowlist — a broad `allow_cidrs` can no longer reopen it.
5. `check_ip` now decodes NAT64 (`64:ff9b::/96`) and 6to4 (`2002::/16`) embedded IPv4, so the pre-flight covers it and the sandbox path is protected through the same check.
6. The failure audit now fires on any escaping error (incl. non-`SshError` / cancellation), not only `SshError`.
7. Non-host-key transport failures map to `CONNECTION_FAILED` instead of `CONNECTION_TIMEOUT`.
8. The agent builder now surfaces a load failure (toast) instead of silently treating it as "no bindings".
9. `os.rmdir` cleanup is wrapped in `suppress(OSError)`.
10. The provider/sink injection seams now validate the structural Protocol via `isinstance` at the hook boundary; the `cast` in `ssh_tools` is replaced by a runtime guard.

**Simplification** — the SSH tool classes now share `_SshToolBase`, dropping the duplicated `category` / `run_json_sync` / `return_type`.

Added targeted tests for the injection chars, egress metadata/tunnel decoding, the streaming cap (both runners), and the non-`SshError` audit path.

One caveat on #3: the Docker streaming path and the Boxlite post-read trim have no CI coverage (both runtimes skip locally); the asyncssh cap and the sandbox `truncated` propagation are covered.

src/xagent/core/ssh/runner.py
```diff
@@ -0,0 +1,314 @@
...
+                ) from exc
+
+        return SshRunResult(
+            exit_code=exit_status if exit_status is not None else -1,
+            stdout=out.decode("utf-8", "replace"),
+            stderr=err.decode("utf-8", "replace"),
+            truncated=truncated,
+        )
+
+    async def upload(
```
rogercloud, line 215: **N1 (Medium) — no timeout on in-process SFTP transfers.** Unlike `execute()`, which now has both a timeout and a streaming output cap, `upload`/`download` have no `asyncio.wait_for` and no timeout parameter. A stalled remote peer during an in-process transfer hangs the worker indefinitely. Recommend wrapping the transfer in `asyncio.wait_for` with a configurable timeout mirroring `execute()`.

src/xagent/core/ssh/errors.py
```diff
@@ -0,0 +1,61 @@
...
+    CREDENTIAL_IN_USE = "ssh_credential_in_use"
+    HOST_KEY_UNVERIFIED = "ssh_host_key_unverified"
+    HOST_KEY_MISMATCH = "ssh_host_key_mismatch"
+    EGRESS_DENIED = "ssh_egress_denied"
+    APPROVAL_REQUIRED = "ssh_approval_required"
+    CONNECTION_TIMEOUT = "ssh_connection_timeout"
+    CONNECTION_FAILED = "ssh_connection_failed"
+    COMMAND_TIMEOUT = "ssh_command_timeout"
+    OUTPUT_LIMIT_EXCEEDED = "ssh_output_limit_exceeded"
+    TRANSFER_LIMIT_EXCEEDED = "ssh_transfer_limit_exceeded"
```
rogercloud: **N2 (Medium/Low) — no transfer size cap; dead error codes.** Nothing in the upload/download path enforces a byte-size limit, and `TRANSFER_LIMIT_EXCEEDED` / `QUOTA_EXCEEDED` are defined but never raised anywhere in the tree (verified by full-tree grep). Either enforce a size cap that raises these codes, or drop them from the public error contract until they're wired up.

src/xagent/core/ssh/executor.py
```diff
@@ -0,0 +1,363 @@
...
+
+def _cap_outputs(stdout: str, stderr: str, budget: int) -> tuple[str, str, bool]:
+    """Cap combined stdout+stderr to ``budget`` bytes; flag if anything was cut.
+    Decodes with errors='ignore' so a byte-boundary cut can't raise."""
+    out_b = stdout.encode("utf-8")
+    err_b = stderr.encode("utf-8")
+    truncated = False
+    if len(out_b) > budget:
+        out_b = out_b[:budget]
+        err_b = b""
```
rogercloud: **N3 (Low) — stderr starvation in combined-budget truncation.** The new streaming layer caps stdout/stderr independently, but `execute()` still passes the result through this unchanged `_cap_outputs`, where stdout exceeding the combined 1 MiB budget hardcodes stderr to `b""`. Consider reserving a minimum stderr slice within the combined budget so error output isn't fully dropped.

src/xagent/core/tools/adapters/vibe/ssh_tools.py
```diff
@@ -0,0 +1,439 @@
...
+        return []
+    numeric_task_id = _numeric_task_id(task_id)
+    agent_id = _agent_id_for_task(session_factory, numeric_task_id)
+    if agent_id is None:
+        logger.info("ssh tools: skip (unresolved agent_id for task_id=%r)", task_id)
+        return []
+
+    context = SshExecutionContext(
+        actor=ActorRef(actor_type="user", actor_id=str(user_id)),
+        execution_principal=PrincipalRef(
```
rogercloud, line 388: **N4 (Low) — hardcoded user principal.** `execution_principal` is pinned to `principal_type="user"` with no team-principal branch. If team-scoped execution is possible, derive the principal type from context instead of hardcoding.

src/xagent/core/tools/adapters/vibe/ssh_tools.py
```diff
@@ -0,0 +1,439 @@
...
+    if task is None:
+        return None
+    if task.agent_id is not None:
+        return int(task.agent_id)
+    cfg = task.agent_config or {}
+    pid = cfg.get("preview_agent_id") if isinstance(cfg, dict) else None
+    return int(pid) if pid is not None else None
+
+
+def _numeric_task_id(task_id: Any) -> int | None:
```
rogercloud, line 280: **N5 (Low/cosmetic) — brittle task-id parse.** `re.search(r"(\d+)$", ...)` only matches trailing digits and silently depends on task-id formatting. Document the assumption or make the extraction explicit (e.g. `re.fullmatch(r"web_task_(\d+)$", ...)`) so a format change fails loudly rather than mis-parsing.

rogercloud (commented):
## Re-Review — Head `c3a80470`
(Previous review at `d26ea353`)

### Update summary
A single, deliberate fix commit (`c3a80470`, "fix: address SSH MCP review feedback", 22 files, +425/-82) landed since the last review and squarely targets findings F1–F10 from our previous review round. This is a high-quality response: the Critical merge blocker (F1) and both Major findings (F2, F3) are addressed, most with dedicated regression tests, and the simplification (F10) matches the exact shape we suggested. All 112 SSH-related tests pass at this head.

### Verdict

Notably more positive than the last round. The Critical SFTP injection blocker (**F1**) is fully fixed and covered by a regression test; **F2** (unwrapped connection/auth failures) is fully fixed; **F3** (unbounded output buffering) is fixed for the in-process and Docker paths. Six of the remaining findings are cleanly resolved.

Two items remain open:

- **F3 — Boxlite backend still unbounded.** The Docker path now streams and caps memory, but the Boxlite backend buffers the entire output before trimming (the author's own code comment acknowledges "host memory isn't actually bounded here"). If Boxlite is a supported/used sandbox backend in production, I'd recommend closing this gap before merge. I can't tell from here how central Boxlite is to your deployment relative to Docker (which is fully fixed) — please treat this as a question rather than an assumed hard blocker.
- **F7 — SSH-bindings load failure is now visible (toast) but not structurally safe.** The user can still ignore the toast and save, silently dropping the "ssh" category, because `onCount` is not called on failure. Acceptable as a minimal mitigation; noting it stays partially open.

Everything else below is resolved. Recommendation: resolve the Boxlite F3 case (or confirm it's out of scope for this deployment), and this is mergeable.

### Resolved findings

- **F1 (Critical) — SFTP command injection via `remote_path` — FIXED.** `_constrain_remote_path` (executor.py) now rejects `"`, `\n`, `\r`, `\x00` unconditionally, ahead of the absolute-path/root-containment checks, so it runs even when `remote_root` is unset. It's the sole shared prologue for both runners. Four traced injection payloads all raise `SshError(OPERATION_NOT_ALLOWED)` before sftp batch construction. Covered by `test_constrain_remote_path_rejects_injection_chars`.
- **F2 (Major) — unwrapped in-process auth/connection failures — FIXED.** `AsyncsshRunner._connect` adds `except (asyncssh.Error, OSError)` after the existing `HostKeyNotVerifiable` clause (correct ordering; `SshError` from `_authorize_peer` still propagates untouched). `executor.py` execute/upload/download now use `except BaseException` → `_audit_failure` (covers non-`SshError` and `CancelledError`) and always re-raise. Covered by `test_execute_audits_unexpected_non_ssh_error`.
- **F4 (Minor) — allowlist overriding metadata deny — FIXED.** `check_ip` (egress.py) now evaluates `deny_metadata` before the allowlist. Covered by `test_metadata_denied_even_when_broadly_allowlisted`.
- **F5 (Minor) — NAT64/6to4 backstop ineffective — FIXED.** New `_decode_tunneled_ipv4` extracts embedded IPv4 from `64:ff9b::/96` and `2002::/16`, applied inside the shared `check_ip` pre-flight before metadata/allow/deny. Byte-slicing independently verified. Covered by 3 new tests.
- **F6 (Minor) — transport failures mislabeled CONNECTION_TIMEOUT — FIXED.** New `SshErrorCode.CONNECTION_FAILED`; `_raise_for_transport`'s generic branch now raises it.
- **F8 (Minor) — unguarded `os.rmdir` — FIXED.** Now `with suppress(OSError): os.rmdir(directory)`.
- **F9 (Minor) — injection-seam `cast()` without runtime guard — FIXED.** `ssh_runtime.py`'s hook getters now raise `TypeError` on failed `isinstance` against the `@runtime_checkable` Protocols; `create_ssh_tools` replaces the bare `cast` with an `isinstance` check before any sensitive use.
- **F10 (Simplification) — triplicated tool-class boilerplate — FIXED.** `_SshToolBase.__init__(context)` is now the single base; all three tool classes extend it. No duplicated `category`/`run_json_sync`/`return_type`.

### Partially resolved

- **F3 (Major) — output buffered before cap — PARTIAL (Boxlite backend still open).** In-process (`runner.py`'s `_run_capped`/`_drain`) and Docker (`docker_sandbox.py`'s `_exec_capped`) are genuinely stream-bounded now. **Boxlite (`boxlite_sandbox.py`) is not:** `exec()` accepts `max_output_bytes` but calls the underlying box unbounded first and trims after full buffering — cosmetic for the DoS vector, and the code's own comment says as much. Keep open for the Boxlite path only.
- **F7 (Minor) — silent SSH-bindings load failure — PARTIAL (toast mitigation only).** `load()` now surfaces `toast.error(...)`, making the failure visible, but `onCount` is still not called on failure, so a user can ignore the toast and save, silently dropping the "ssh" category. Reasonable minimal response; structurally the drop is still possible.

### Newly discovered, still open (N1–N5)

These come from our own internal notes from the last review round and were not yet posted as inline comments; this commit's scope did not touch them. Posting them now.

- **N1 (Medium) — no timeout on in-process SFTP transfers.** `AsyncsshRunner.upload`/`download` (runner.py) have no `asyncio.wait_for` and no timeout parameter, unlike `execute()` which now has both a timeout and the streaming cap. A stalled remote peer hangs the worker indefinitely.
- **N2 (Medium/Low) — no byte-size cap on transfers, and dead error codes.** Nothing in the upload/download path enforces a size limit, and `TRANSFER_LIMIT_EXCEEDED`/`QUOTA_EXCEEDED` are defined but never raised anywhere (verified by full-tree grep).
- **N3 (Low) — stderr starvation in `_cap_outputs`.** The new streaming layer caps stdout/stderr independently, but `execute()` still routes the result through the unchanged `_cap_outputs` for the final combined-budget truncation, where stdout alone exceeding budget zeroes stderr.
- **N4 (Low) — hardcoded user principal.** `create_ssh_tools`'s `execution_principal=PrincipalRef(principal_type="user", ...)` has no team-principal branch.
- **N5 (Low/cosmetic) — brittle task-id parse.** `_numeric_task_id`'s `re.search(r"(\d+)$", ...)` matches only trailing digits.

### Testing
Full SSH suite re-run at `c3a80470` — `tests/core/ssh/`, `tests/core/tools/test_ssh_tools.py`, `tests/web/services/test_ssh_runtime.py` — **112/112 pass.** Sandbox-implementation-level streaming tests for `docker_sandbox.py`/`boxlite_sandbox.py` could not be exercised in this environment (likely needs a live Docker daemon); the Boxlite F3 gap was confirmed by source review, including the implementation's own acknowledging comment.

### Bottom line
This PR is now much closer to mergeable. The Critical blocker and both Major findings are resolved, most with regression coverage. Before merge I'd like the author to close **F3's Boxlite case** — or confirm Boxlite is out of scope for this deployment. **N1/N2** are worth doing but read as reasonable follow-ups rather than blockers. F7 can stay a tracked UX follow-up.

bluefish-08:
Thanks for the thorough re-review. Follow-up commit addresses N1/N3/N5 plus the remaining items:

- **N1**: in-process SFTP transfers now bounded by a timeout, mirroring `execute()`.
- **N3**: combined-output cap now reserves a stderr floor so a large stdout can't zero out diagnostics.
- **N5**: documented the trailing-digits task-id parse assumption.
- **N2**: dropped the dead `TRANSFER_LIMIT_EXCEEDED`/`QUOTA_EXCEEDED` codes; leaving the byte-size cap as the follow-up you flagged.
- **F7**: `ssh` is an assignable category, so it round-trips through `selectedToolCategories` — a failed bindings-load won't drop it at save. Seeded `hasSshBindings` from the saved config on load to make that explicit.
- **F3 (Boxlite)**: intentional — Boxlite is the local low-volume dev runner; production runs on the stream-bounded Docker backend, and Boxlite's box exec has no streaming read API (documented in the code comment).
- **N4**: no team context at this call site yet (`config` only exposes `get_user_id()`); a team-principal branch is deferred to the upcoming team tool-ownership work.

rogercloud (commented):
## Summary

This PR adds a managed SSH-over-MCP capability: four tools (`ssh_list_targets`, `ssh_execute`, `ssh_upload`, `ssh_download`) backed by a new `core/ssh/` execution engine. `SshExecutor` drives a single pipeline — resolve → capability-check → egress-preflight → credential-decrypt → materialize → run → cap-output → cleanup — behind one runner seam with two implementations: `AsyncsshRunner` (in-process) and `SandboxSshRunner` (OpenSSH binary inside a leased per-task sandbox). It also ships an IP egress policy, strict host-key verification, an injectable audit sink, and a frontend agent-builder SSH-bindings section. ~5.3k lines / 43 files. No linked issue found.

## Approach — acceptable with reservations

The seam design is clean and testable: pure egress decision logic is separated from I/O (`egress.py` vs `egress_io.py`), the runner seam is a `@runtime_checkable` Protocol, materialize-and-shred is a per-call context manager, and the sandbox lease correctly spans both materialize and run. The two-runner split is coherent rather than duplicative. Design-level reservations (not tied to a single line):

- **Connect-phase timeout gap** — the in-process runner does not time-bound the TCP-connect + SSH handshake, unlike the sandbox runner which wraps the whole invocation. Tracked as **M1** below.
- **"Deny-by-default" oversells the shipped posture** — `EgressPolicyConfig.default_allow_public=True` is the actual default and nothing in `_egress_from_env` flips it, so by default any *public* IP is reachable; only loopback/private/link-local/metadata are denied. Not a bug, but the PR description should be corrected.
- **Minor constant scatter** — duplicated `_SECRET_ROOT="/tmp"`, transfer-timeout, and `_TIMEOUT_EXIT` constants across modules; cosmetic.

## Prior findings — re-verified

Re-read fresh this round; the following are confirmed **FIXED**:

- SFTP batch-file command injection via `remote_path` (prior Critical) — `_constrain_remote_path` now rejects `"`, `\n`, `\r`, `\x00` unconditionally before the path/root checks, covering upload+download and rooted+unrooted; regression test `test_constrain_remote_path_rejects_injection_chars` exists.
- Unwrapped in-process auth/connection failures (prior Major) — `_connect` catches `(asyncssh.Error, OSError)` after `HostKeyNotVerifiable` in the correct order so `_authorize_peer`'s `SshError` still propagates; executor catches `BaseException` around the audit call.
- Unbounded output buffering / host-memory DoS (prior Major) — in-process runner and Docker sandbox now stream-cap incrementally. *(Boxlite still open — see M2.)*
- Egress allowlist could override metadata deny — `deny_metadata` now evaluated before the allowlist.
- NAT64/6to4 backstop — `_decode_tunneled_ipv4` extracts embedded IPv4 from `64:ff9b::/96` and `2002::/16` before allow/deny.
- Transport failures mislabeled as `CONNECTION_TIMEOUT` — new `CONNECTION_FAILED` code.
- Unguarded `os.rmdir` in cleanup — now under `suppress(OSError)`.
- Injection-seam `cast()` with no runtime guard — explicit `isinstance` checks against the Protocols.
- Triplicated tool-class boilerplate — collapsed into `_SshToolBase.__init__`.
- Silent SSH-bindings load could drop the "ssh" category on save — `agent-builder.tsx` seeds `hasSshBindings` from the saved `tool_categories`.
- No timeout on in-process SFTP transfers (prior N1) — now `asyncio.wait_for` wrapped *(but see m3 for a related new gap in the same code)*.
- Dead `TRANSFER_LIMIT_EXCEEDED`/`QUOTA_EXCEEDED` codes removed (N2); stderr starvation fixed with a `budget//4` floor (N3); brittle task-id parse documented (N5).
- Gemini bot (4): private `ipaddress._BaseAddress` typing → public aliases; per-call CIDR reparse → pre-parsed in `__post_init__`; missing `os.fsync` before shred → present and correctly ordered; bare `object` sandbox typing → `SandboxLike` Protocol (zero `type: ignore[attr-defined]` remain in `core/ssh/`).

**Still open (low severity):** hardcoded `principal_type="user"` (prior N4) — cross-referenced as **m7** below.

Residual test gap on the auth fix: no test drives a real `asyncssh.PermissionDenied` through `_connect`, and only `execute` (not `upload`/`download`) exercises the generic-exception audit path.

## Findings

### Major

**M1 — in-process connect/handshake is not time-bounded** (`src/xagent/core/ssh/runner.py`, `_connect` ~L108-162)
`asyncssh.connect(...)` is called with no `connect_timeout=`/`login_timeout=`, and each `asyncio.wait_for(timeout_seconds)` in `execute`/`upload`/`download` wraps only the post-connect operation, never the `_connect()` context-manager entry. A slow or malicious target can stall the handshake indefinitely regardless of the requested `timeout_seconds`, hanging the worker — a real divergence from `SandboxSshRunner`, which bounds the entire invocation via the `timeout` coreutil. No test covers a hanging connect. Wrap the `_connect()` entry in `asyncio.wait_for(timeout_seconds)` (or pass `connect_timeout=`/`login_timeout=`) at all three call sites.

**M2 — Boxlite backend still buffers unbounded SSH output** (`src/xagent/sandbox/boxlite_sandbox.py`, `exec()` ~L165-199)
Carried over, still open. The Docker path and in-process runner now stream-cap, but Boxlite calls `self._box.exec(...)` unbounded and only trims via `_truncate_utf8` after the full output is buffered — the code comment itself acknowledges host memory isn't bounded. This is reachable from the SSH feature: `SandboxSshRunner.execute` is backend-agnostic and `SANDBOX_BACKEND=boxlite` (`web/sandbox_manager.py`) constructs `BoxliteSandboxService`, so any boxlite deployment retains a host-memory DoS via `ssh_execute` on unbounded remote output. Please either add a stream-level cap to Boxlite's exec, or gate SSH tool availability on the Docker backend and confirm Boxlite is out of scope for SSH-enabled deployments.

### Minor

- **m1 — cancellation-audit race** (`core/ssh/executor.py` ~L135-137, 186-188, 226-228): the `except BaseException` → audit-then-reraise path isn't shielded; a second `CancelledError` during `await self._audit_sink.record(...)` can silently drop the failure audit, undermining the "audit fires even on cancellation" guarantee. Wrap the audit call in `asyncio.shield(...)`. Untested.
- **m2 — truncation forces `exit_code=-1`** (`core/ssh/runner.py`, `_run_capped` ~L293-311): on truncation `proc.wait()` is skipped and `exit_code` is hardcoded `-1`, so any command exceeding the 1 MiB cap reports `-1` even on real success, with no cross-reference to the `truncated` flag. Recover the status with a short bounded `wait()`/`kill()`+`wait()`, or document that `exit_code` is meaningless when `truncated=True`. Untested.
- **m3 — SFTP transfer timeout bypasses the configured clamp** (`core/ssh/executor.py` ~L184, 224; `runner.py` `upload` ~L206-246): transfers use a fixed `DEFAULT_TRANSFER_TIMEOUT_SECONDS=300` that ignores the executor's `max_timeout_seconds` clamp (unlike `execute()`). Worse, `upload()`'s existence-check and `put()` use two separate `wait_for(timeout_seconds)` calls, so one upload's worst-case wall-clock approaches 2×300s even when the deployment configures a tighter max. Clamp to `max_timeout_seconds` and share one budget across the two waits.
- **m4 — double-slash false rejection** (`core/ssh/executor.py`, `_constrain_remote_path` ~L340-344): `posixpath.normpath` preserves two leading slashes, so `//root/x` under `remote_root="/root"` normalizes to `//root/x` and fails the `"/root/"` prefix check — a valid in-root path is wrongly rejected. Not a security hole (nothing wrongly allowed), just over-strict; collapse leading slashes before the check. Untested.
- **m5 — silent SSH-target load failure** (`frontend/src/components/build/agent-ssh-bindings.tsx` ~L93-103): the target-picker load-failure catch is silent (no toast, no error state), inconsistent with the sibling bindings-`load()` catch in the same file which toasts the same failure class. Leaves the dropdown empty with no feedback.
- **m6 (nit) — multicast not checked** (`core/ssh/egress.py`, `check_ip` ~L102-141): `.is_multicast` (e.g. `224.0.0.1`) falls through to `default_allow_public`. `.is_reserved`/`.is_unspecified` are already covered via `.is_private`. Narrow and not practically exploitable for a TCP/SSH connect; mention only.
- **m7 (carryover, low) — hardcoded principal type** (`core/tools/adapters/vibe/ssh_tools.py` ~L388-390): `execution_principal` hardcodes `principal_type="user"` with no team branch, though `PrincipalType` already includes `"team"` and is tested in `test_types.py`. Looks like an unfinished gap, low impact today.

- `core/ssh/types.py` L11, L78: **dead-feature** `ApprovalPolicy`/`approval_policy` is threaded through `ResolvedSshTarget` and the UI but never read; `APPROVAL_REQUIRED` is never raised. Drop the field and UI control until enforcement exists.
- `core/ssh/errors.py` L14-26: **dead-code** 7 of 18 `SshErrorCode` members are never raised (`TARGET_IN_USE`, `EXTERNAL_POLICY_DENIED`, `CREDENTIAL_REVOKED`, `CREDENTIAL_IN_USE`, `HOST_KEY_UNVERIFIED`, `CONNECTION_TIMEOUT`, `OUTPUT_LIMIT_EXCEEDED`). Remove until a caller needs them.
- `core/ssh/egress.py` L72, L84: **dead-code** `extra_denied_cidrs`/`_denied_networks` is only set in tests, no production wiring. Drop until a deployment path uses it.
- `core/ssh/sandbox_runner.py` L141, L180, L233: **unused-param** `egress_config: object = None` is never referenced in any body. Remove; it only exists for Protocol parity.
- `core/ssh/runner.py` (Protocol + both impls): **duplication** the ~10-keyword signature is repeated across 6 method defs. Replace with one `SshConnectionParams` dataclass.
- `web/services/ssh_runtime.py` L30-64: **duplication** `get_ssh_target_provider`/`get_ssh_audit_sink` are structurally identical. Factor into `_build_hook(factory, protocol, label)`.

net: -70 lines possible

## Testing

`pytest tests/core/ssh/ -q` in the review worktree: full suite (106+ tests — egress, runner, executor, sandbox_runner, materializers, in-memory adapters, live asyncssh integration) passes, 100%, no failures. Coverage is otherwise strong; the per-finding gaps noted above (M1 hanging connect, m1 cancellation-audit race, m2 truncation exit-code, m3 transfer timeout, m4 double-slash) have no regression tests.

## Also considered and ruled out

- Non-UTF-8 private-key materialization in `sandbox_materializer.py` — not a live bug; asyncssh `client_keys=` requires PEM/OpenSSH text (always valid UTF-8/ASCII).
- Concurrent SSH-tool lease sharing one sandbox instance — each call materializes into a fresh `token_hex(16)`/`mkdtemp` path, so no file collision.
- `_numeric_task_id` trailing-digit regex — limitation documented and all task-ids produced here (`web_task_{id}`) are safe.
- Unguarded audit-sink hook in `create_ssh_tools` — the outer `ToolFactory` creator loop wraps every creator in try/except, so a bad sink only drops SSH tools for that agent.
- `docker_sandbox.py` `_exec_capped` draining to EOF — every call site wraps the command with the `timeout` coreutil, so it's wall-clock bounded in practice.
- Frontend alias-dedup relying on client state — the backend uniqueness endpoint isn't in this checkout; flagging as a question for the author rather than a finding.

src/xagent/core/ssh/runner.py
```diff
@@ -0,0 +1,337 @@
...
+
+class AsyncsshRunner:
+    """Runs commands and SFTP transfers over SSH with strict, non-interactive
+    security settings, in-process (self-hosted, no sandbox subsystem).
+
+    Accepts ``sandbox`` and ``connect_ip`` for seam parity but ignores them: it
+    connects by hostname and re-checks the actual peer IP (``_authorize_peer``)
+    as its own DNS-rebinding backstop."""
+
+    @asynccontextmanager
```
rogercloud, line 108: M1 (Major): the connect/handshake phase here is never time-bounded. `asyncssh.connect(...)` is called with no `connect_timeout=`/`login_timeout=`, and the `asyncio.wait_for(timeout_seconds)` calls in execute()/upload()/download() wrap only the post-connect operation, never this `_connect()` entry. A slow or malicious target can stall the TCP-connect + banner/kex/auth indefinitely regardless of the requested timeout_seconds, hanging the worker — diverging from SandboxSshRunner, which bounds the whole invocation via the `timeout` coreutil. Please wrap the `_connect()` entry in `asyncio.wait_for(timeout_seconds)` (or pass connect_timeout/login_timeout) at all three call sites. No test currently covers a hanging connect.

src/xagent/sandbox/boxlite_sandbox.py
```diff
@@ -170,10 +179,23 @@ async def exec(
                 # Remove first line, keep the rest
                 stderr = lines[1] if len(lines) > 1 else ""
 
+        stdout = res.stdout
+        truncated = False
+        if max_output_bytes is not None:
```
rogercloud, line 184: M2 (Major, carryover — still open): this path calls `self._box.exec(...)` unbounded and only trims via `_truncate_utf8` after the full output is buffered in host memory. The Docker path and in-process runner were fixed to stream-cap incrementally, but Boxlite still has the host-memory DoS. This is reachable from the SSH feature: `SandboxSshRunner.execute` is backend-agnostic and `SANDBOX_BACKEND=boxlite` constructs `BoxliteSandboxService`, so any boxlite deployment retains the DoS via `ssh_execute` on unbounded remote output. Please add a stream-level cap here, or gate SSH tool availability on the Docker backend and confirm Boxlite is out of scope for SSH-enabled deployments.

src/xagent/core/ssh/executor.py
```diff
@@ -0,0 +1,370 @@
...
+                    username=resolved.username,
+                    private_key_path=paths.private_key_path,
+                    known_hosts_path=paths.known_hosts_path,
+                    command=command,
+                    timeout_seconds=timeout,
+                    egress_config=self._egress_config,
+                    max_output_bytes=self._max_output_bytes,
+                )
+            duration_ms = int((time.monotonic() - start) * 1000)
+        except BaseException as exc:
```
rogercloud, line 142: m1 (Minor): this `except BaseException` -> audit-then-reraise path isn't shielded from a second cancellation. If `CancelledError` lands again while `await self._audit_sink.record(...)` is in flight, the failure audit is silently dropped, undermining the 'audit fires even on cancellation' guarantee. Wrap the audit call in `asyncio.shield(...)`. Same pattern in upload()/download(). No test cancels an in-flight operation to verify this.

src/xagent/core/ssh/runner.py
```diff
@@ -0,0 +1,337 @@
...
+                await asyncio.wait_for(
+                    sftp.get(remote_path, local_path), timeout_seconds
+                )
+            except TimeoutError as exc:
+                raise SshError(
+                    SshErrorCode.COMMAND_TIMEOUT, "transfer timed out", cause=exc
+                ) from exc
+
+
+async def _run_capped(
```
rogercloud, line 304: m2 (Minor): on truncation `proc.wait()` is skipped and `exit_code` is hardcoded to -1, so any command whose output exceeds the 1 MiB cap reports exit_code=-1 even on real success, with no cross-reference to the `truncated` flag for callers. Either recover the real status with a short bounded `wait()` (or `kill()`+`wait()`), or document that exit_code is meaningless when truncated=True. Untested.

src/xagent/core/ssh/executor.py
```diff
@@ -0,0 +1,370 @@
...
+                    connect_ip=connect_ip,
+                    port=resolved.port,
+                    username=resolved.username,
+                    private_key_path=paths.private_key_path,
+                    known_hosts_path=paths.known_hosts_path,
+                    local_path=local_path,
+                    remote_path=remote_path,
+                    overwrite=overwrite,
+                    egress_config=self._egress_config,
+                    timeout_seconds=DEFAULT_TRANSFER_TIMEOUT_SECONDS,
```
rogercloud: m3 (Minor): SFTP transfers use a fixed DEFAULT_TRANSFER_TIMEOUT_SECONDS=300 that bypasses the executor's `max_timeout_seconds` clamp entirely (unlike execute(), which clamps). Additionally, upload()'s existence-check and the actual put() are wrapped in two separate `wait_for(timeout_seconds)` calls rather than one shared budget, so worst-case wall-clock for a single upload approaches 2x300s even when the deployment configures a tighter max. Please clamp the transfer timeout to max_timeout_seconds and share one budget across the two waits.

src/xagent/core/ssh/executor.py
```diff
@@ -0,0 +1,370 @@
...
+            SshErrorCode.OPERATION_NOT_ALLOWED,
+            "remote path contains forbidden characters",
+        )
+    if not remote_path.startswith("/"):
+        raise SshError(
+            SshErrorCode.OPERATION_NOT_ALLOWED, "remote path must be absolute"
+        )
+    if remote_root is None:
+        return
+    normalized = posixpath.normpath(remote_path)
```
rogercloud, line 354: m4 (Minor): `posixpath.normpath` preserves exactly two leading slashes per POSIX, so a remote_path like `//root/x` under remote_root="/root" normalizes to `//root/x` and fails the `"/root/"` prefix check — a valid in-root path is wrongly rejected. Not a security hole (nothing wrongly allowed), just over-strict; collapse leading slashes before the prefix check. No test covers this edge case.

frontend/src/components/build/agent-ssh-bindings.tsx
```diff
@@ -0,0 +1,291 @@
...
+      // bindings exist server-side, since onCount never fires.
+      toast.error(t("ssh.bindings.loadFailed"))
+    }
+  }, [agentId, onCount, t])
+
+  useEffect(() => {
+    load()
+  }, [load])
+
+  useEffect(() => {
```
rogercloud, line 93: m5 (Minor): the SSH-target picker's load-failure catch is silent (no toast, no error state), inconsistent with the sibling bindings-load() catch in this same file, which toasts the same class of failure. The dropdown is left empty with no user feedback — please surface the error the same way.

bluefish-08:
All of M1, M2, and m1–m6 are fixed, plus the dead error codes and duplicate hook getters cleaned up. Regression tests added for m3/m4/m6.

- **M2** gates SSH tools off the Boxlite backend only (`SANDBOX_IMPLEMENTATION=boxlite` → `create_ssh_tools` emits nothing); the Docker default and the self-hosted in-process path are unaffected, so this doesn't disable SSH broadly — it just removes the one unbounded-buffer path.

Deferred, with reasoning:
- **`egress_config` on `SandboxSshRunner`** stays — it's required by the `SshRunner` Protocol because `AsyncsshRunner` uses it for the DNS-rebinding peer re-check, so removing it breaks the in-process path.
- **`extra_denied_cidrs`** stays too — it's wired into `check_ip` and covered by a test, so it's an unwired-but-implemented deny hook, not dead code.
- The signature→`SshConnectionParams` dataclass refactor and m7 (hardcoded principal) are tracked as follow-ups; `ApprovalPolicy` kept for now.

rogercloud (approved):
## Re-review of `7f7baf2f` — "fix: address SSH MCP re-review (M1, M2, m1-m6)"

### Update summary

This commit responds point-by-point to the prior review at `d880b2d9` and lands cleanly: both Majors and all six minors are resolved, and the two simplification cleanups that were in scope are done. The logic on every item is correct; the only residuals are two missing regression tests (M1 timeout kwargs, M2 Boxlite gate) and one non-blocking design-coupling note on m6. No new regressions were found in a fresh blind pass over the diff, and the full SSH suite is green.

### Resolved findings

- **M1 — FIXED.** `_connect` now threads a required `timeout_seconds` into all three call sites and passes `connect_timeout`/`login_timeout` to `asyncssh.connect(...)`, with `TimeoutError` added to the `CONNECTION_FAILED` mapping; `_authorize_peer`'s `SshError` still propagates. (Minor note: the code comment attributes `login_timeout` expiry to `TimeoutError`, but it actually surfaces as `ConnectionLost`, already covered by the pre-existing `asyncssh.Error` branch — functionally harmless.)
- **M2 — FIXED (logic), PARTIAL (test gap only).** `create_ssh_tools` now short-circuits to `[]` when `SANDBOX_IMPLEMENTATION == "boxlite"`, before any runner/materializer is built. The env var name and `"docker"` default match `sandbox_manager.py` exactly; the in-process (`sandbox_lease is None`) path is untouched; the short-circuit releases no resources (lease is a lazy closure). This is a test gap, not a logic bug.
- **m1 — FIXED.** `_audit_failure` wraps the audit call in `asyncio.shield(...)`, the correct primitive to survive a second cancellation; no side effect on the normal path. Untested in practice.
- **m2 — FIXED as documentation.** Comment-only change documenting that `exit_code == -1` on truncation is meaningless and must be read together with `truncated`. Acceptable for a minor finding; the caveat remains unenforced convention (no caller jointly checks the two fields).
- **m3 — FIXED + test.** `SshExecutor` now clamps transfer timeout to `min(DEFAULT_TRANSFER_TIMEOUT_SECONDS, max_timeout_seconds)`, and `AsyncsshRunner.upload` folds the existence-check + `put()` into one `wait_for`, closing the ~2x worst case. `test_transfer_timeout_clamped_to_max` covers it; the "destination exists" `SshError` still propagates distinctly from a real timeout.
- **m4 — FIXED + test.** `_constrain_remote_path` collapses a preserved `//root/x` → `/root/x` before containment; injection-char check and 3+-slash normpath behavior unaffected. `test_constrain_remote_path_allows_leading_double_slash`.
- **m5 — FIXED.** The target-picker load path now throws on non-ok and toasts `ssh.bindings.targetsLoadFailed` (key present in both `en.ts`/`zh.ts`), mirroring the sibling loader; `t` is a stable memoized reference, so the added dep introduces no refetch loop.
- **m6 — FIXED + test.** `check_ip` now denies `is_multicast` alongside `is_private` under `deny_private`, with correct `allow_cidrs` precedence. `test_multicast_denied`. (See residual note below.)
- **Simplification cleanups — FIXED.** All 7 dead `SshErrorCode` members removed (full-repo grep clean); the two duplicated hook getters in `ssh_runtime.py` collapse into one `_resolve_hook(...)` helper with per-protocol/label TypeError messages preserved. `test_ssh_runtime.py` + `test_errors.py`: 9 passed.

### Non-blocking residual notes

1. **Missing regression tests (M1, M2).** Both are logic-verified correct but untested: nothing asserts `connect_timeout`/`login_timeout` reach `asyncssh.connect` or that a connect-timeout maps to `CONNECTION_FAILED`, and nothing exercises the Boxlite `[]` gate (no `boxlite`/`SANDBOX_IMPLEMENTATION` hits in `test_ssh_tools.py`). Recommend adding both before merge — nice-to-have, not a blocker.
2. **m6 `deny_private`/multicast coupling.** Multicast-deny is gated behind `deny_private`, but the two are orthogonal: a deployment setting `deny_private=False` (e.g. to allow private-network SSH targets) silently loses multicast protection. Not a regression from the default config. Recommend a follow-up to decouple (independent always-on check or its own flag) if a `deny_private=False` deployment is ever expected.

### Still open (unchanged, out of scope)

- **m7** — hardcoded `principal_type="user"` remains, as noted previously. Low severity, correctly out of scope for this commit; no action needed here. (The four un-addressed simplification items — ApprovalPolicy dead field, unused `egress_config` param, repeated runner signatures, unused `extra_denied_cidrs` — likewise remain, none in scope for this commit.)

### Testing

Full SSH suite re-run at head `7f7baf2f` (`tests/core/ssh/`, `tests/core/tools/test_ssh_tools.py`, `tests/web/services/test_ssh_runtime.py`): **118 passed, 0 failed**.

### Bottom line

**Mergeable.** Every Major and minor is resolved with correct logic, the in-scope cleanups are done, and no new regressions were introduced. The two missing regression tests and the m6 coupling are worth addressing but do not block — treat them as nice-to-have follow-ups.

bluefish-08:
Addressed the two residual notes: added regression tests for the M1 `connect_timeout`/`login_timeout` kwargs + the `CONNECTION_FAILED` mapping, and for the M2 Boxlite gate; and decoupled the multicast deny from `deny_private` (now unconditional, since the two are orthogonal). Also rebased onto latest `main`.

PR #994, opened 2026-07-24
feat(mcp): add Google Ads connector via OAuth + platform developer token

yiboyasss:
## Summary
- Add a Google Ads MCP connector (`google_ads.py`) with two tools:
  `google_ads_list_accessible_customers` and `google_ads_search` (GAQL),
  built on the Google Ads REST API v23 with plain `requests` (no new
  dependency, matches the existing HubSpot connector's style).
- Register the connector in `builtin_mcp_registry.py` + a seed migration,
  reusing the existing "google" OAuth provider (scope: `.../auth/adwords`).
  No per-user API key form needed — same OAuth "Connect" flow as Gmail.
- Add a generic `static_env` mechanism to the OAuth launch-config builder
  in `tools/config.py`, so an OAuth-transport connector can forward one
  platform-wide secret (here, `GOOGLE_ADS_DEVELOPER_TOKEN`) into its MCP
  subprocess alongside the per-user OAuth token. MCP subprocesses don't
  inherit the parent process's env, so this was previously not possible
  without a per-app special case.
- Document `GOOGLE_ADS_DEVELOPER_TOKEN` in `example.env` as a
  platform-wide secret (analogous to `GOOGLE_CLIENT_ID`), not a per-user
  key.

## Test plan
- [x] `pytest tests/web/tools/test_google_ads_mcp.py
      tests/alembic/test_20260724_seed_google_ads_mcp_app.py
      tests/web/tools/test_oauth_launch_config_static_env.py -v`
- [x] `pytest tests/web/api/test_public_mcp_connector_visibility.py -v`
      (regression: new app doesn't break existing catalog assertions)
- [x] `ruff check` / `ruff format --check` on all changed files
- [x] Manual: connect Google Ads via the OAuth popup in the connector UI,
      confirm `google_ads_list_accessible_customers` returns real
      customer IDs, then run a GAQL query with `google_ads_search`
      against one of them

### DIFF
--- a/src/xagent/web/tools/mcp/google_ads.py
+++ b/src/xagent/web/tools/mcp/google_ads.py
@@ -65,6 +65,57 @@ def _headers(login_customer_id: str | None = None) -> dict[str, str]:
     return headers
 
 
+def _extract_error_detail(response: requests.Response) -> str | None:
+    """Pull the human-readable message(s) out of a Google Ads error body.
+
+    Google Ads error responses are a large structured JSON payload
+    (``{"error": {"message": ..., "details": [{"errors": [...]}]}}``);
+    returning that whole blob to the LLM wastes tokens and is harder for it
+    to act on than the plain message(s) it actually needs to fix a bad GAQL
+    query. Returns None if the body isn't in the expected shape, so the
+    caller can fall back to the raw response text.
+    """
+    try:
+        payload = response.json()
+    except ValueError:
+        return None
+
+    if not isinstance(payload, dict):
+        return None
+    error = payload.get("error")
+    if not isinstance(error, dict):
+        return None
+
+    messages: list[str] = []
+    top_message = error.get("message")
+    if isinstance(top_message, str) and top_message:
+        messages.append(top_message)
+
+    for detail in error.get("details") or []:
+        if not isinstance(detail, dict):
+            continue
+        for sub_error in detail.get("errors") or []:
+            if not isinstance(sub_error, dict):
+                continue
+            sub_message = sub_error.get("message")
+            if isinstance(sub_message, str) and sub_message:
+                messages.append(sub_message)
+
+    if not messages:
+        return None
+    return "; ".join(dict.fromkeys(messages))
+
+
+def _require_dict_result(result: Any) -> dict[str, Any]:
+    """Guard against a non-dict payload (e.g. a list or string) before
+    calling dict methods on it. A dict that's merely empty (zero
+    accessible customers, zero GAQL rows) is a normal, valid response and
+    must not be rejected here."""
+    if not isinstance(result, dict):
+        raise ValueError("Unexpected response format from Google Ads API")
+    return result
+
+
 def _request(
     method: str,
     path: str,
@@ -82,10 +133,12 @@ def _request(
     try:
         response.raise_for_status()
     except requests.HTTPError as exc:
-        response_text = response.text.strip()
         message = str(exc)
-        if response_text:
-            message = f"{message} - {response_text}"
+        detail = _extract_error_detail(response)
+        if detail is None:
+            detail = response.text.strip()
+        if detail:
+            message = f"{message} - {detail}"
         raise RuntimeError(message) from exc
 
     if response.status_code == 204 or not response.content:
@@ -100,9 +153,13 @@ def google_ads_list_accessible_customers() -> str:
     Use this first to discover which customer_id values are available for google_ads_search.
     """
     try:
-        result = _request("GET", "/customers:listAccessibleCustomers")
+        result = _require_dict_result(
+            _request("GET", "/customers:listAccessibleCustomers")
+        )
         resource_names = result.get("resourceNames", [])
-        customer_ids = [name.rsplit("/", 1)[-1] for name in resource_names]
+        customer_ids = [
+            name.rsplit("/", 1)[-1] for name in resource_names if isinstance(name, str)
+        ]
         return _success(customer_ids=customer_ids)
     except Exception as e:
         logger.error(f"Error listing accessible customers: {e}")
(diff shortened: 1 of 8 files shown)


src/xagent/web/tools/mcp/google_ads.py
```diff
@@ -0,0 +1,199 @@
...
+
+# Ensure standard proxy environment variables are set to prevent hanging requests
+setup_proxy_env()
+
+mcp = FastMCP("google-ads-mcp")
+
+GOOGLE_ADS_BASE_URL = "https://googleads.googleapis.com/v23"
+DEFAULT_TIMEOUT_SECONDS = 30
+
+_CUSTOMER_ID_PATTERN = re.compile(r"^[0-9-]+\Z")
```
rogercloud: [Minor] `^[0-9-]+\Z` matches an all-dash string like `"---"`, which `_normalize_customer_id` (lines 34-44) then strips to `""`. Two consequences: `google_ads_search(customer_id="---")` builds the collapsed URL `/customers//googleAds:search` and issues a real request instead of failing fast locally, and `_headers(login_customer_id="---")` passes the `if login_customer_id:` check and sends `login-customer-id: ""` as an empty header. Not a security/injection issue — purely an input-validation gap. Suggest requiring at least one digit (`^[0-9][0-9-]*\Z`) or asserting the post-strip value is non-empty. No test currently covers this.

rogercloud (approved):
## Summary

This PR adds a Google Ads MCP connector (`google_ads_list_accessible_customers`, `google_ads_search`) that reuses the existing "google" OAuth provider rather than introducing a new auth flow. To carry the platform-wide `GOOGLE_ADS_DEVELOPER_TOKEN` into the MCP subprocess (which does not inherit the parent process env), it introduces a new `static_env` mechanism in the OAuth launch-config builder that forwards a named host env var into the subprocess env alongside the per-user OAuth token. The connector is registered through `builtin_mcp_registry.py` plus a seed migration.

## Approach — Sound

The `static_env` design is a clean, minimal extension: it is symmetric with the existing `env_mapping` (same `launch_config` dict, same builder function), rather than a per-app special case. It is safer than the alternatives — having the subprocess inherit all host env would leak every platform secret, and a separate "platform secrets registry" would be over-engineered. The builder constructs `env` from scratch (it does not splat `os.environ`), so `static_env` is genuinely necessary and not speculative; connectors without it are unaffected (the loop no-ops on an empty dict), and only `server.name`/`transport` are logged, so no secret leaks into logs.

One design-level note (suggestion, not a blocker): a `static_env` value can name *any* host env var to forward. This is currently safe only because `launch_config` is written exclusively by migrations, the builtin registry, and admin-only API routes (`src/xagent/web/api/admin_mcp.py`, gated by `Depends(verify_admin)`) — there is no end-user-writable path. A one-line comment recording this trust assumption would keep it from being silently invalidated if `launch_config` editing is ever exposed more broadly.

## Findings

### [Minor] All-dash `customer_id` / `login_customer_id` bypasses validation into an empty string
`src/xagent/web/tools/mcp/google_ads.py:23` — `_CUSTOMER_ID_PATTERN = re.compile(r"^[0-9-]+\Z")` matches a string of only dashes (e.g. `"---"`). `_normalize_customer_id` (lines 34-44) then does `.replace("-", "")`, yielding `""`. Two concrete consequences:
1. `google_ads_search(customer_id="---")` builds the collapsed URL `/customers//googleAds:search` and issues a real HTTP request instead of failing fast locally — the bad input surfaces only as an opaque remote 404/400.
2. `_headers(login_customer_id="---")` passes the `if login_customer_id:` truthiness check, normalizes to `""`, and sets `headers["login-customer-id"] = ""` — an empty header value is sent silently rather than rejected.

Not a security/injection issue (no header injection, no path traversal outside the API's own customer namespace) — purely an input-validation gap. Suggest requiring at least one digit, e.g. `^[0-9][0-9-]*\Z`, or asserting the post-strip result is non-empty. No existing test covers this edge case.

## Prior-review reconciliation

All items raised across earlier review passes are now **fixed and independently re-verified** against the current code and passing tests:
- **customer_id / login_customer_id validation** (reject anything not digits+dashes) — FIXED: `_normalize_customer_id` (`google_ads.py:34-44`) is applied to both and raises `ValueError` instead of silently sanitizing.
- **`$` → `\Z` anchor** (closes the trailing-newline/CRLF bypass) — FIXED: `google_ads.py:23` now uses `\Z`.
- **Missing test for trailing-newline rejection** — FIXED: `tests/web/tools/test_google_ads_mcp.py:59-63` covers both `\r\n` and `\n`.
- **Empty-string `static_env` value dropped by truthiness check** — FIXED: `config.py:2514-2519` now uses `if value is not None:` + `str(value)` coercion, with a dedicated passing test.
- **Structured JSON error parsing** — FIXED (latest commit `26ee99a0`): `_extract_error_detail` now parses `error.message`/`error.details[].errors[].message` with a safe fallback to raw text, covered by 2 new tests.
- **`isinstance(result, dict)` guard before `.get(...)`** — FIXED (latest commit `26ee99a0`): `_require_dict_result` now guards both tool functions, with new tests for non-dict responses and for non-string entries in `resourceNames`.

Only the all-dash `customer_id` gap above remains outstanding, and it is minor/non-blocking.

- `src/xagent/web/tools/config.py` L2514: yagni — `static_env` typed as `Mapping[str, str]` (`{subprocess_var: host_var}`) to allow renaming between host and subprocess var names, but every current caller (builtin registry entry, seed migration `ROW`, all relevant tests) uses an identity mapping — no caller ever renames. A `list[str]` (e.g. `static_env: ["GOOGLE_ADS_DEVELOPER_TOKEN"]`) would cover every real case with less surface area.

  **Constraint:** don't change the shape in the seed migration file — repo convention treats merged migrations as immutable, and `test_seed_row_matches_registry` guards that the migration `ROW` and the registry `ROW` have the *same* shape, so changing one without the other would break that test. Net benefit is small (~-2 to -4 lines) once that constraint is respected — worth weighing, not a clear win.

## Tests

All 65 tests pass, run from a fresh venv (Python 3.11) via `.venv/bin/python -m pytest -v`:
- `tests/web/tools/test_google_ads_mcp.py`
- `tests/alembic/test_20260724_seed_google_ads_mcp_app.py`
- `tests/web/tools/test_oauth_launch_config_static_env.py`
- `tests/web/api/test_public_mcp_connector_visibility.py` (regression check)

65 passed, 0 failed.

PR #993, opened 2026-07-24
fix: enforce per-guest isolation on public share channels (#973)

codeacme17:
## What

**PR1 of the public-share-channel hardening pass (#973): per-guest isolation.**

A public share link is opened by many anonymous visitors who all share the same owner + agent/workforce id. `get_task_for_share_context` authorized a task solely by `Task.id` (a guessable sequential PK) + owner + agent/workforce id, with **no per-guest claim** — so any visitor holding a valid share JWT for the same shared entity could address another visitor's task id and read/continue that conversation. This affects both the agent-share and workforce-share paths; workforce sessions carry more private multi-turn content, so the blast radius is larger.

## How

**Backend**
- Mint a server-owned, high-entropy `guest_id` at `POST /api/share/auth` (both agent- and workforce-share paths) and sign it into the guest JWT. The endpoint **never** accepts a client-supplied `guest_id`: share has no secondary credential (unlike widget's embed ticket / widget key), so `guest_id` is the sole isolation credential and must be server-minted.
- `get_share_chat_user` **fails closed** on tokens that carry no `guest_id` (those minted before this change) rather than falling back to the old no-isolation behavior.
- `get_task_for_share_context` requires `task.agent_config["guest_id"]` to match the caller's, on top of the existing checks, in **both** branches (shared helper `_require_share_guest_owns_task`). This automatically covers the WS connect + per-message revalidation, since `share_chat_websocket_endpoint` routes through the same function.
- Stamp `guest_id` into the task at create time: the agent-share path assigns server keys **last** so a client-supplied value can't override; the workforce-share path threads it via `extra_agent_config` (survives the snapshot merge, which never sets `guest_id`).

**Frontend** (share only; widget behavior unchanged)
- Persist and reuse the guest JWT in `localStorage` so `guest_id` is stable across reloads, instead of re-authing on every mount (which would mint a new guest each time and make a returning visitor's own tasks fail the check).
- Recover a returning visitor whose persisted `taskId` belongs to a different guest (e.g. a pre-migration task): catch the `4003`/`403` access denial, drop the stale `taskId`, and start a fresh session.

## Migration / scope

- No schema change and no migration — `guest_id` lives in the existing `agent_config` JSON.
- **Accepted cost:** pre-existing anonymous share conversations become unreachable (old tasks have no `guest_id`), which is acceptable for anonymous public share (no accounts, no persistence promise).
[...]

PR #996, opened 2026-07-24
feat: harden public share uploads — storage gate + orphan GC (#973)

### LINKED ISSUE #973: Harden public share channels: per-guest isolation + rate limiting (follow-up to #947/#967)

codeacme17:
## What

**PR3 of the public-share-channel hardening pass (#973): public upload hardening.**

The public share/widget upload paths bypassed the owner's storage accounting, and the task-less first-turn upload path (workforce share) creates `UploadedFile` rows **before any task binding** — orphaned forever if the guest never completes task creation. This adds the missing owner storage gate and a scoped orphan GC, completing the three-layer defense with PR2: limiter (rate) / storage gate (total) / GC (residue).

> **No longer stacked** — #993 (PR1) and #995 (PR2) have merged; this branch is rebased onto current `main` and the diff is PR3-only. The rebase also adapted to two things that landed on `main` meanwhile:
> - The upload-path refactor (worker-owned sessions): `upload_source` is now threaded through the two-phase registration pipeline (`LocalUploadRegistration` → `stage_uploaded_file_from_local_path` → `StagedUploadedFile`) instead of the old inline `create_from_local_path` call.
> - The new **workforce-widget task-less upload branch** (same orphan exposure as the share one) now also stamps the marker, so GC covers it too.
> - The migration is re-parented onto `main`'s new chain tail (`20260725_add_uploaded_file_recovery_index`) — single head, verified with a from-scratch `alembic upgrade head`.

## How

**Storage gate (3a)** — call `check_storage_gate(db, owner)` before `store_uploaded_files` on every public upload entry (share, both branches + widget), mirroring the KB ingest gate: fail-open, raise 402 on a truthy reason. Charged to the share/widget entity **owner** (`access_context.user`) — the same attribution as the run gate and the on-disk write.

**Orphan GC (3b)** — a coarse `task_id IS NULL` sweep can't be used: that state is a normal system-wide intermediate (plain uploads + turn-time binding across every channel), so it would reap logged-in users' un-sent draft attachments. Instead:

- New nullable `UploadedFile.upload_source` marker column (the **one** migration in the whole #973 pass), stamped **only** on the task-less public uploads (workforce share + workforce widget). The same migration adds `ix_uploaded_files_orphan_gc` on `(upload_source, task_id, created_at)` — the sweep's exact predicate — built `CONCURRENTLY` on PostgreSQL.
- An **in-app background loop** (`run_orphan_upload_gc_loop`, started from the app lifespan like the uploaded-file recovery loop — runs in every supported deployment, no Celery required) reaps rows that carry the marker, still have no `task_id`, and are older than the TTL: one bounded keyset page per tick, with the cursor carried across ticks (a short page resets it), so a repeatedly failing oldest page can't starve newer orphans. The index also lives in `UploadedFile.__table_args__`, since fresh installs stamp Alembic head before `create_all()`. Deletion **rides the existing uploaded-file compensation protocol** end to end: the local staging file goes first (while the row is still `available` and durable-backed, so a winning bind re-materializes from the durable object); then the exact `compensate_registered_uploads_sync` CAS claim (`status='available' AND task_id IS NULL` → `compensating` + `updated_at` generation token — overlapping sweeps are mutually exclusive, binders lose cleanly); then the session-less `delete_uploaded_file_compensation_object` + token-fenced `settle_uploaded_file_compensation_no_commit`. Every post-claim crash window is owned by the shipped stale-compensation recovery loop, with nothing left it cannot clean.
- Both binders interlock with the claim: `bind_turn_files_no_commit` already did, and `_bind_selected_files_to_task` — the binder the task-less workforce path actually uses — now claims with a conditional UPDATE + post-verify instead of a stale read-then-assign.
- The widget upload route gets its own throttle keyed per widget entity + caller IP (the widget `guest_id` is client-supplied, so a guest-keyed bucket would be rotatable) — closing the orphan-production faucet on the task-less workforce-widget branch, which had no request-rate limit.

**Scope (3c)** — quota/gate covers all uploads; GC targets task-less only (bound-at-create uploads can't orphan; CASCADE cleans them on task delete).

## Config

`XAGENT_TASKLESS_UPLOAD_TTL_SECONDS` (default 48h, 1h floor so a slow real first turn is never reaped mid-flow), `XAGENT_ORPHAN_UPLOAD_SWEEP_INTERVAL_SECONDS` (default hourly), plus the widget upload throttle `XAGENT_WIDGET_UPLOAD_RATE_LIMIT` (per widget entity, default 240/minute) and `XAGENT_WIDGET_UPLOAD_IP_RATE_LIMIT` (per caller IP, default 60/minute) — documented in `example.env`, tested in `tests/core/test_config.py`.

## Tests

- GC unit tests: reaps aged+marked+unbound; spares marked-but-recent, unmarked (draft), and marked-but-bound rows; verifies on-disk file removal. Claim-protocol tests pin both halves of the GC↔bind interlock, the bound-between-fetch-and-claim race, batch bounding, and failed-delete retry.
- Widget upload throttle: limiter bucket tests (IP trips across entities/guests; entity ceiling trips across IPs) and an endpoint test pinning that rotating the client-supplied guest_id does not reset the budget (429).
- Integration: task-less share upload stamps the marker; storage-gate hook → 402 and no orphan row created.
- Config default / override / floor tests.
- Full `tests/web` suite passes; ruff / mypy clean.

Part of #973.


src/xagent/web/api/public_chat_access.py
```diff
@@ -789,11 +881,35 @@ async def share_chat_websocket_endpoint(
...
+            # limiter and each starts an owner-billed run, so rate-limit the
+            # run-starting turn types per guest here. Reject the turn (the
+            # client surfaces it and can retry) rather than closing — a rate
+            # limit is transient. Interventions are control messages, not new
+            # runs, so they are not gated.
+            if message_type in ("chat", "execute_task") and (
+                not get_share_rate_limiter().allow_ws_turn(
+                    current_access_context.guest_id
+                )
+            ):
```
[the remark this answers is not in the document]
codeacme17: No `NameError` here. `current_access_context` is assigned within the `try` block of the same `while` loop iteration (line 873). Python has function-level scoping, not block-level, so a name bound inside a `try` remains visible in the rest of the function. Also, the `except HTTPException` branch `return`s on failure, so by the time execution reaches this line `current_access_context` is guaranteed to be bound.

This is also intentional: `current_access_context` is the context re-validated per message, which is more appropriate than `access_context` (bound once at connect). Both are `ShareChatAccessContext` and carry `guest_id`, so they are functionally equivalent. Keeping as is.

src/xagent/web/api/chat.py
```diff
@@ -2289,6 +2289,49 @@ async def execute_task(
...
+        # per-link + per-guest rolling ceiling on top, keyed off the share
+        # markers PR1 stamped into agent_config. Only share tasks are gated;
+        # fails open (availability) like the run gate above.
+        if gate_task is not None:
+            try:
+                share_config = getattr(gate_task, "agent_config", None)
+                if (
+                    isinstance(share_config, Mapping)
+                    and share_config.get("auth_mode") == "share"
+                ):
```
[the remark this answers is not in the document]
codeacme17: `Mapping` is already imported in this file (line 9: `from typing import Any, Dict, List, Mapping, Optional, cast`) and is used in several places (e.g. lines 254, 260, 804, 2268), so there is no `NameError`. Using `Mapping` is intentional: it is broader than `dict` (also covers dict subclasses / mapping types) and is consistent with the existing style in this file. Keeping as is.

rogercloud:
ci is failing

codeacme17:
Rebased onto current `main` now that #993 and #995 have merged — the diff has collapsed to the PR3-only changes (13 files, +632).

Re CI: the previous run had two failures, both addressed:
- **pre-commit / isort** — import order in `public_chat_access.py`; fixed in the rebase.
- **Test SQLite Migrations** — runner ran out of disk (`No space left on device`), infra flake, not code.

Rebase adaptations (called out in the updated PR description): `upload_source` is threaded through main's new two-phase upload registration pipeline; the new workforce-widget task-less upload branch is now also stamped (same orphan exposure); the migration is re-parented onto `20260725_add_uploaded_file_recovery_index` (single head, verified with a from-scratch `alembic upgrade head`). Full `tests/web` passes locally; isort / ruff / mypy clean.

codeacme17:
Correction to my previous comment: the second CI round showed the SQLite migrations failure was **not** (only) an infra flake — the `downgrade` test caught a real bug in this PR's migration.

`uploaded_files` is migration-created (`20260225`) with FKs to `users`/`tasks`, which are `create_all()`-owned and absent in the migrations-only database the downgrade test builds (SQLite tolerates dangling FK targets at create time). The upgrade's ADD COLUMN takes the native no-recreate path, but the downgrade's DROP COLUMN forces a SQLite batch table recreate, which reflects the table's FKs and raised `NoSuchTableError: users`.

Fixed by passing `reflect_kwargs={"resolve_fks": False}` to `batch_alter_table` (both directions): reflection no longer tries to load the dangling FK target tables, while the FK DDL itself is still carried over by name. Reproduced and verified locally with the exact CI command — `tests/migrations/test_migration_integration.py --db sqlite all` (upgrade + idempotence + downgrade + re-upgrade) now passes, as does the `tests/migrations` pytest suite.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            # Force a fresh read so a bind committed by a concurrent run-start
+            # (between the query above and now) is visible — a stale identity-
+            # map snapshot would otherwise let GC delete a now-bound
+            # attachment. refresh raises if the row was deleted meanwhile, in
+            # which case there is nothing to GC.
+            db.refresh(row)
+            if row.task_id is not None:
+                continue
+            store.delete(row)
```
rogercloud: **[P2] Make the orphan claim atomic with deletion**

`refresh()` only narrows the race: run-start can bind and commit immediately after the `task_id` check, before this PK-based deletion. GC would then remove the newly bound row and its storage bytes. Use an atomic conditional/CAS claim—or a locking protocol shared with binding—so cleanup proceeds only while `task_id IS NULL`.

src/xagent/web/api/public_chat_access.py
```diff
@@ -676,13 +697,17 @@ async def upload_public_chat_files(
...
+        # these rows if the guest never completes task creation, without a
+        # coarse task_id-IS-NULL sweep touching other paths' unbound drafts.
         return await store_uploaded_files(
             upload_items=upload_items,
             task_type=task_type,
             task_id=None,
             folder=folder,
             user_id=user_id,
             single_file_mode=file is not None and (not files),
+            upload_source=TASKLESS_SHARE_UPLOAD_SOURCE,
```
rogercloud, line 710: **[P2] Rate-limit task-less widget uploads**

Unlike the share upload route, this newly marked workforce-widget path has no request-rate limiter. A valid widget guest can repeatedly create orphan uploads until GC catches up. Apply a widget/workforce/IP-keyed limit before storage; `guest_id` alone is unsuitable because the client supplies it.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,83 @@
...
+    reference = now or datetime.now(timezone.utc)
+    cutoff = reference - timedelta(seconds=older_than_seconds)
+    rows = (
+        db.query(UploadedFile)
+        .filter(
+            UploadedFile.upload_source == TASKLESS_SHARE_UPLOAD_SOURCE,
+            UploadedFile.task_id.is_(None),
+            UploadedFile.created_at < cutoff,
+        )
+        .all()
```
rogercloud: **[P2] Process the GC backlog in bounded batches**

`.all()` materializes every eligible row, followed by durable-file I/O and a commit per row. With no singleton lease or continuation, a large backlog can exhaust worker memory and overlap later hourly sweeps. Process deterministic bounded batches under a claim or lease.

src/xagent/migrations/versions/20260724_add_upload_source_to_uploaded_files.py
```diff
@@ -0,0 +1,68 @@
...
+    bind = context.get_bind()
+    inspector = Inspector.from_engine(bind)
+    if TABLE not in inspector.get_table_names():
+        return
+    # Guarded so the migration is re-runnable on a partially-applied DB.
+    if COLUMN not in _existing_columns(inspector, TABLE):
+        with op.batch_alter_table(
+            TABLE, reflect_kwargs=BATCH_REFLECT_KWARGS
+        ) as batch_op:
+            batch_op.add_column(sa.Column(COLUMN, sa.String(length=64), nullable=True))
```
rogercloud, line 139: **[P2] Index the hourly orphan-GC predicate**

The migration adds `upload_source` without an index supporting the hourly `upload_source + task_id + created_at` query. Existing indexes cannot serve it, and marked rows remain marked after binding, so full-table scans grow with upload history. Add a composite or partial index and remove it on downgrade.

rogercloud (changes requested):
The provenance marker, nullable migration/configuration, owner storage gate, and TTL cleanup form a coherent solution, but the inline correctness and production-scaling issues should be addressed before relying on this GC path. Focused tests, SQLite migration round-trip, Ruff, and diff checks passed; current GitHub CI is green.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 52bcb3e2. GC now claims each row with a conditional UPDATE — `SET storage_status='compensating' WHERE id=? AND task_id IS NULL`, committed before any storage I/O — which is the exact counterpart of `bind_turn_files_no_commit`'s conditional bind (`SET task_id WHERE task_id IS NULL AND storage_status != 'compensating'`). The database serializes the two row-level updates, so whichever commits first wins: bind-first → the claim's predicate no longer matches and the row is spared; claim-first → the bind skips the now-`compensating` row. This reuses the existing claim protocol from `compensate_registered_uploads_sync` rather than inventing a new locking scheme. Tests pin both halves of the interlock plus the bound-between-fetch-and-claim race (`test_claim_fails_on_bound_row…`, `test_claimed_row_is_excluded_from_run_start_binding`, `test_row_bound_between_fetch_and_claim_is_spared`).

src/xagent/web/api/public_chat_access.py
codeacme17: Fixed in 52bcb3e2. The widget upload route now carries its own throttle, checked before any authorization/storage work — keyed per widget entity (`agent:<id>` / `workforce:<id>`) **plus** caller IP, per your point that the widget `guest_id` is client-supplied and therefore rotatable (unlike the server-minted share `guest_id`, which is why the share route can key on it). Per-IP is the tight abuser bucket (default 60/minute), per-entity the loose backstop bounding total production through one widget across many callers (default 240/minute); both are `XAGENT_WIDGET_UPLOAD_*` ops-knobs documented in example.env. The endpoint test authenticates two different guest_ids and asserts the second request still 429s — pinning that guest rotation does not reset the budget.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 52bcb3e2. The sweep now processes oldest-first batches (`LIMIT 500`, at most 20 batches per tick) instead of `.all()`, so a backlog can neither materialize wholesale into worker memory nor run unbounded into the next tick; whatever a capped sweep leaves behind still matches the predicate and is drained by later ticks. A full batch that makes zero progress (all rows spared or failing) stops the sweep instead of refetching the same stuck set. On overlap: with the per-row CAS claim from the other thread, two overlapping sweeps are already correctness-safe (each row is claimed exactly once — the loser's claim matches zero rows), so I bounded the work per tick rather than adding a separate singleton lease; happy to add a Redis lease on top if you'd prefer sweeps to be strictly serialized.

src/xagent/migrations/versions/20260724_add_upload_source_to_uploaded_files.py
codeacme17: Fixed in 52bcb3e2. The same migration now creates `ix_uploaded_files_orphan_gc` on `(upload_source, task_id, created_at)` — exactly the sweep's predicate shape — and drops it on downgrade (before the column drop). On PostgreSQL it is built inside an `autocommit_block` with `postgresql_concurrently=True` (mirroring the `20260725` recovery-index precedent) so the hot uploaded_files table is not write-locked during the build; guarded/`if_not_exists` so the migration stays re-runnable. Verified via the SQLite `all` round-trip (upgrade + idempotence + downgrade + re-upgrade) and a from-scratch upgrade asserting the index exists.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _claim_orphan(db: Session, row_id: int) -> bool:
+    """Atomically claim one still-unbound row for deletion.
+
+    The conditional UPDATE is the counterpart of run-start binding
+    (``bind_turn_files_no_commit``: ``WHERE task_id IS NULL AND
+    storage_status != 'compensating'``): if binding committed first the
+    ``task_id IS NULL`` predicate no longer matches and the claim returns
+    False (row spared); if the claim commits first, binding skips the row
+    because it is now ``compensating``. Committed immediately so the claim is
```
rogercloud: **[P2] Interlock with the binder used by workforce runs**

The claimed counterpart, `bind_turn_files_no_commit`, is not used by the task-less workforce share/widget path. That path reaches `_bind_selected_files_to_task`, which selects the row and later performs a normal PK-based ORM update. It can therefore select an available row, let this GC claim commit, then commit a stale `task_id` assignment before GC deletes the now-bound upload. Make the actual workforce binder use a conditional update/CAS sharing this claim protocol.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,141 @@
...
+    claimed = (
+        db.query(UploadedFile)
+        .filter(
+            UploadedFile.id == row_id,
+            UploadedFile.task_id.is_(None),
+        )
+        .update(
+            {UploadedFile.storage_status: "compensating"},
+            synchronize_session=False,
+        )
```
rogercloud, line 206: **[P2] Preserve the durable object identity during deletion**

Setting `storage_status` to `compensating` before calling `UploadedFileStore.delete()` prevents durable deletion: `ManagedFileRef.has_durable_object` only returns true for `storage_status == "available"`. The sweep then removes the local file and metadata while silently leaving the durable object permanently untracked. Delete through a detached exact storage claim, or otherwise ensure the durable object is removed before settling the claimed row.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    if claimed != 1:
+        db.rollback()
+        return False
+    db.commit()
```
rogercloud, line 263: **[P2] Recover the local file after a crashed GC claim**

Once this claim is committed, a worker crash hands the row to the generic stale-compensation recovery path. That recovery can delete the durable object and metadata, but its candidate snapshot has no `storage_path` and never removes the GC-owned local file. Add GC-aware recovery that cleans both resources, or use a distinct claim protocol carrying the local path.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,141 @@
...
+    False (row spared); if the claim commits first, binding skips the row
+    because it is now ``compensating``. Committed immediately so the claim is
+    visible to concurrent binders before any storage I/O starts.
+    """
+    claimed = (
+        db.query(UploadedFile)
+        .filter(
+            UploadedFile.id == row_id,
+            UploadedFile.task_id.is_(None),
+        )
```
rogercloud, line 199: **[P2] Make the claim exclusive across overlapping sweeps**

The claim predicates only on `id` and `task_id IS NULL`, so a row already marked `compensating` still matches. Two overlapping sweeps can both receive `rowcount == 1` and both proceed as owners of the same row. Require the exact expected status and a generation token, then CAS-settle that claimed version.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,141 @@
...
+                    getattr(row, "id", "?"),
+                    exc_info=True,
+                )
+        if len(rows) < batch_size:
+            break
+        if batch_deleted == 0:
+            # A full batch produced no deletions (all spared or failing):
+            # refetching would return the same stuck set, so yield to the
+            # next scheduled sweep instead of spinning.
+            break
```
rogercloud: **[P2] Advance past a permanently failing oldest batch**

When the oldest full batch produces no deletions, every scheduled sweep stops on that same set. A batch of undeletable rows therefore permanently starves all newer eligible orphans. Continue with a keyset cursor, skip/back off failed claims, or quarantine them so later rows remain reachable.

src/xagent/migrations/versions/20260724_add_upload_source_to_uploaded_files.py
```diff
@@ -0,0 +1,96 @@
...
+    if INDEX not in _existing_indexes(Inspector.from_engine(bind), TABLE):
```
rogercloud: **[P2] Repair invalid concurrent indexes on rerun**

A failed PostgreSQL `CREATE INDEX CONCURRENTLY` can leave an index entry with `indisvalid = false`. Inspector still reports its name, so this guard—and `IF NOT EXISTS`—treats the broken index as complete forever. Reuse the following recovery-index migration's `pg_index.indisvalid` check and concurrent drop/rebuild logic.

rogercloud (changes requested):
The update fixes the widget limiter, bounds each sweep, and adds the missing GC index, but the cleanup ownership protocol still permits data loss and storage leaks. Focused tests, SQLite migration round-trip, Ruff, and diff checks pass; current GitHub CI is green.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 66aac9de — you were right that I interlocked with the wrong binder. `_bind_selected_files_to_task` now claims rows with a conditional UPDATE (`SET task_id WHERE task_id IS NULL AND storage_status != 'compensating'`) followed by a post-UPDATE ownership verification, the same protocol as `bind_turn_files_no_commit` — the stale read-then-assign is gone, so a bind can no longer overwrite a committed GC/compensation claim. The 404 contract is preserved (rows bound to THIS task pass idempotently; rows bound elsewhere, claimed, or nonexistent surface as missing, and the caller's rollback discards partial binds). Test: `test_claimed_row_is_excluded_from_workforce_run_binding` pins that a claimed row raises FILE_NOT_FOUND with the claim intact, alongside the existing WS-turn binder pin.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 66aac9de. GC no longer deletes through the status-gated `UploadedFileStore.delete` path at all: the durable object is removed via `delete_uploaded_file_compensation_object` (session-less, keyed by owner + exact storage key, presence-probed) — the same detached storage claim request compensation uses — and metadata settles via the token-fenced `settle_uploaded_file_compensation_no_commit`. An unresolved presence (`exists`/`unknown`) defers the claimed row to the stale-compensation recovery loop instead of settling, so a live object can never be stranded untracked. `test_reap_deletes_durable_object_via_compensation_helper` pins that the delete is invoked with the owner + exact key regardless of the claimed status.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 66aac9de by ordering rather than by widening the shared recovery: GC deletes the local staging file FIRST, before the claim, while the row is still `available` and durable-backed — a recoverable state (a consumer that wins the bind re-materializes the bytes from the durable object via `ensure_local`; a crash there just leaves a fully consistent row for the next sweep). By the time any claim exists, the local file is already gone, so every post-claim crash window lands on the generic stale-compensation recovery with nothing left that it cannot clean. This keeps the recovery candidate shape (and its no-`storage_path` contract) untouched. `test_unresolved_durable_delete_hands_off_to_stale_recovery` pins the handoff: deferred row is claimed, local file already removed, and it surfaces in `get_stale_uploaded_file_compensation_candidates`.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 66aac9de. The claim is now the exact `compensate_registered_uploads_sync` CAS: `WHERE id + user_id + file_id + storage_key match AND storage_status = 'available' AND task_id IS NULL`, setting `storage_status='compensating'` and `updated_at` as the generation token (read back as persisted, since the DB may round the datetime). Requiring the exact prior status makes overlapping sweeps mutually exclusive — the loser matches zero rows — and the settlement is fenced on the exact token via `settle_uploaded_file_compensation_no_commit`, so even a resurrected claim cannot settle someone else's version. The candidate scan is also scoped to `storage_status='available'`, so claimed rows never re-enter a sweep; stale claims belong to the recovery loop's takeover protocol. `test_claim_requires_exact_available_status` pins bound-row rejection, token persistence, and second-claim exclusion.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 66aac9de. The sweep now advances a keyset cursor `(created_at, id)` across batches — the zero-progress break is gone — so spared or failing rows are passed over within the sweep and every newer orphan is still reached. Rows that fail *after* a successful claim leave the sweep's predicate entirely (the scan is scoped to `storage_status='available'`) and belong to the stale-compensation recovery loop, so they cannot re-poison later pages either. `test_failing_oldest_rows_do_not_starve_newer_orphans` pins it with a poisoned oldest row and `batch_size=1`.

src/xagent/migrations/versions/20260724_add_upload_source_to_uploaded_files.py
codeacme17: Fixed in 66aac9de. The PostgreSQL path now mirrors `20260725_add_uploaded_file_recovery_index`: it checks `pg_index.indisvalid` (via `to_regclass`) plus the on-line column shape, returns early only when the index is valid AND matches, and otherwise drops the broken/mismatched entry concurrently inside the `autocommit_block` before rebuilding concurrently. Downgrade likewise drops concurrently with `if_exists`, covering a valid, invalid, or absent index alike. Verified via the SQLite `all` round-trip locally; the PostgreSQL migration job exercises the PG path in CI.

src/xagent/web/jobs/celery_app.py
```diff
@@ -38,12 +40,19 @@ def create_celery_app() -> Any:
...
+            "xagent.web.jobs.maintenance_tasks.sweep_orphaned_uploads": {
+                "queue": "default",
+            },
         },
         beat_schedule={
             "scan-due-triggers-and-stale-jobs": {
                 "task": "xagent.web.jobs.trigger_tasks.scan_due_triggers",
                 "schedule": float(sweep_interval),
             },
+            "sweep-orphaned-taskless-uploads": {
```
rogercloud: **[P1] Run orphan GC in every supported deployment**

This Beat entry is the only scheduler for the cleanup task, but Celery defaults to disabled while supported Gunicorn-only deployments still accept taskless uploads. Those installations never reap abandoned metadata or stored bytes. Provide a supported non-Celery scheduler, or gate this upload path on an active maintenance runner.

src/xagent/migrations/versions/20260724_add_upload_source_to_uploaded_files.py
```diff
@@ -0,0 +1,152 @@
...
+            if_not_exists=True,
+            postgresql_concurrently=True,
+        )
+
+
+def upgrade() -> None:
+    from alembic import context
+
+    bind = context.get_bind()
+    inspector = Inspector.from_engine(bind)
```
rogercloud, line 131: **[P2] Keep offline migrations reflection-free**

Offline `alembic upgrade … --sql` supplies a `MockConnection`, which `Inspector.from_engine()` cannot reflect; downgrade fails identically. Add explicit offline upgrade/downgrade branches that emit deterministic DDL without inspecting a live database.

src/xagent/web/models/uploaded_file.py
```diff
@@ -41,6 +41,12 @@ class UploadedFile(Base):  # type: ignore
     etag = Column(String(255), nullable=True)
     workspace_relative_path = Column(String(2048), nullable=True)
     workspace_category = Column(String(64), nullable=True)
+    # Provenance marker for uploads created before any task/owner binding
+    # exists (currently the task-less public-share path, #973). NULL for the
+    # overwhelming majority of rows (task-bound / logged-in draft uploads);
+    # only orphan GC keys off this so a coarse "task_id IS NULL" sweep can't
+    # reap a logged-in user's un-sent draft attachments.
+    upload_source = Column(String(64), nullable=True)
```
rogercloud, line 59: **[P2] Add the GC index to ORM metadata**

Fresh startup stamps an empty database to Alembic head before `Base.metadata.create_all()`, so this migration never creates `ix_uploaded_files_orphan_gc` there. Because the model omits that composite index, fresh installations run the growing GC query without it. Declare the index in `UploadedFile.__table_args__` and test the stamp-plus-create-all path.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,291 @@
...
+    of ``batch_size`` (at most ``max_batches`` per sweep). The cursor always
+    advances past spared or failing rows, so a bad oldest page can never
+    starve newer orphans; whatever one sweep defers is retried by the next.
+    Per-row failures are logged and skipped so one bad row does not abort the
+    sweep. Returns the number of metadata rows deleted.
+    """
+    reference = now or datetime.now(timezone.utc)
+    cutoff = reference - timedelta(seconds=older_than_seconds)
+    deleted = 0
+    cursor: _OrphanCursor | None = None
```
rogercloud: **[P2] Carry progress across capped sweeps**

The keyset cursor advances only within one invocation and resets on every scheduled run. If the oldest `max_batches * batch_size` rows repeatedly fail before being claimed, every sweep consumes its cap on that same prefix and permanently starves newer orphans. Persist progress across runs or move repeated failures into an excluded retry/backoff state.

src/xagent/web/services/workforce_runs.py
```diff
@@ -235,19 +234,36 @@ def _bind_selected_files_to_task(
...
 
-    uploaded_files = (
-        db.query(UploadedFile)
+    # Claim every currently-unbound row with a conditional UPDATE (the same
+    # protocol as ``bind_turn_files_no_commit``), NOT a read-then-assign: a
+    # stale ORM assignment would overwrite a compensation/GC claim that
+    # committed between the SELECT and the flush, binding a row whose bytes
+    # are already being deleted (#973). The database serializes this against
+    # the claimers' own conditional updates, so whichever commits first wins.
+    db.query(UploadedFile).filter(
```
rogercloud: **[native] Reuse the existing turn-file binder**

`bind_turn_files_no_commit()` already implements this conditional update, ownership verification, same-task idempotence, and ordered missing-ID result without committing. Delegating to it and translating non-empty missing IDs to `WorkforceRunError` preserves behavior while removing roughly 27 net lines of duplicated concurrency protocol.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,291 @@
...
+            },
+            synchronize_session=False,
+        )
+    )
+    if claimed != 1:
+        db.rollback()
+        return None
+    # Read the token back as persisted: the database may round the datetime,
+    # and the settlement fences on exact equality.
+    token = cast(
```
rogercloud: **[native] Reuse the compensation-token loader**

`_load_uploaded_file_compensation_token_no_commit()` performs the same row-id plus `compensating`-status token query without committing and introduces no import cycle. Reusing it here removes the duplicated token-read implementation.

rogercloud (changes requested):
The compensation/CAS design is now coherent and the previous ownership-protocol issues are largely fixed, but the remaining deployment, migration, and bounded-progress gaps still make this not merge-ready. Focused tests, SQLite migration round-trip, Ruff, diff checks, and current GitHub CI pass; offline Alembic generation fails as noted inline.

src/xagent/web/jobs/celery_app.py
codeacme17: Fixed in 0c4d6807. The GC is now an **in-app background loop** — `run_orphan_upload_gc_loop` started/stopped from the app lifespan (`start_orphan_upload_gc_task` / `stop_orphan_upload_gc_task` in app.py), mirroring the uploaded-file recovery loop that already runs this way — so every supported deployment reaps task-less orphans, Celery or not. The Beat entry, task route, and `maintenance_tasks` module are removed (single scheduler, no double-run), and the existing `XAGENT_TASKLESS_UPLOAD_TTL_SECONDS` / `XAGENT_ORPHAN_UPLOAD_SWEEP_INTERVAL_SECONDS` knobs now feed the loop. Overlap across Gunicorn workers stays safe under the exact-status claim (the loser matches zero rows).

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 0c4d6807, folded into the in-app loop from the P1 thread: the sweep is now **one bounded keyset page per tick**, and the loop carries the cursor ACROSS ticks exactly like `run_uploaded_file_compensation_recovery_loop` (`cursor = result.next_cursor if result.scanned == batch_size else None`; a short page resets it for the next bounded pass). `next_cursor` always points past every processed row — spared and failing included — so a repeatedly failing oldest page is passed over on the following tick instead of consuming every sweep's cap. Rows that fail *after* a successful claim additionally leave the scan predicate entirely (`status='available'` scope) and belong to the stale-compensation recovery. `test_cursor_advances_past_failing_rows` pins the poison-page case; `test_pages_thread_the_cursor_and_drain_the_backlog` pins the paging contract.

src/xagent/web/models/uploaded_file.py
codeacme17: Fixed in 0c4d6807. `ix_uploaded_files_orphan_gc` is now declared in `UploadedFile.__table_args__` (next to the existing recovery-index declaration) with a comment explaining why it must live in both places — fresh installations stamp Alembic head before `Base.metadata.create_all()`, so the migration never runs there. `test_create_all_declares_the_gc_index` pins the stamp-plus-create-all path by inspecting the engine after `init_db()`; the migration's guarded index creation stays a no-op on databases where create_all already produced it.

src/xagent/migrations/versions/20260724_add_upload_source_to_uploaded_files.py
codeacme17: Fixed in 0c4d6807. Both `upgrade()` and `downgrade()` now start with an explicit `context.is_offline_mode()` branch that emits deterministic DDL with no inspection — plain `op.add_column` / `op.create_index` (PG: inside `autocommit_block` with `CONCURRENTLY`), and the mirror-image drops on downgrade. Verified locally: `alembic upgrade …:head --sql` and `alembic downgrade head:… --sql` both emit the expected statements (they previously crashed with `MockConnection has no attribute 'close'` in both directions).

src/xagent/web/services/workforce_runs.py
codeacme17: Done in 0c4d6807 — `_bind_selected_files_to_task` now delegates to `bind_turn_files_no_commit` and translates a non-empty missing list into the `FILE_NOT_FOUND` `WorkforceRunError`; the hand-copied conditional-update block is gone (the `UploadedFile`/`or_` imports went with it). No import cycle: `file_turn` depends only on models/core.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Done in 0c4d6807 — the claim now calls `_load_uploaded_file_compensation_token_no_commit` instead of the duplicated inline token query. (I had originally avoided it only because of the underscore prefix; agreed it's the right reuse.)

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,355 @@
...
+        except asyncio.CancelledError:
+            raise
+        except Exception as exc:
+            if is_database_pool_timeout(exc):
+                logger.warning(
+                    "Orphan task-less upload GC skipped after database pool timeout"
+                )
+            else:
+                logger.exception("Orphan task-less upload GC tick failed")
+        await asyncio.sleep(poll_interval_seconds)
```
rogercloud: **[P1] Drain a bounded set of pages before sleeping**

With the defaults (`poll_interval_seconds=3600`, `batch_size=500`), this unconditional sleep caps a supported single-process deployment at 500 reclaimed objects per hour, while the public task-less upload endpoints permit sustained ingress well above that rate. Expired metadata and stored bytes can therefore grow indefinitely, defeating this PR's resource-exhaustion protection. Keep the isolated per-page sessions and cursor fairness, but continue through full pages up to a bounded page/time budget, and apply the long sleep only after a short page or that budget is exhausted; the previous 20-page/10,000-row per-tick cap is a useful capacity reference.

rogercloud (changes requested):
The latest update fixes the previously reported application scheduling, cross-tick cursor, fresh-install index, offline migration, and helper-reuse gaps. One remaining production-capacity regression is called out inline and should be addressed before merge. Focused tests, SQLite migration integration, PostgreSQL/SQLite offline SQL generation, Ruff, diff checks, and the current GitHub CI pass; no live PostgreSQL concurrency test was run.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 5d407730. A tick (`_run_gc_tick`) now keeps consuming **full** pages — each still in its own short Session, cursor fairness unchanged — until either a short page (backlog drained; cursor resets for the next bounded pass) or a `GC_MAX_PAGES_PER_TICK = 20` page budget (10,000 rows at the defaults, matching the Beat-era capacity you referenced), and only then takes the long sleep. Exhausting the budget hands the next tick a live cursor so it resumes where this one stopped. `test_tick_drains_full_pages_before_sleeping` pins both halves: a single tick drains a multi-page backlog to a reset cursor, and a budget-capped tick returns a live cursor that the following tick finishes from.

rogercloud (commented):
The latest update restores bounded multi-page draining and preserves cursor fairness, but the previous capacity finding is only partially addressed; the remaining operational and documentation issues are inline. The overall design remains acceptable with reservations. In an isolated environment, 418 focused tests passed; Ruff, diff checks, migration/offline checks, and the current GitHub CI are green. No live PostgreSQL concurrency test was run.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,393 @@
...
+        if result.scanned:
+            logger.info(
+                "Orphan task-less upload GC: scanned=%s deleted=%s",
+                result.scanned,
+                result.deleted,
+            )
+        if result.scanned < batch_size:
+            return None
+        cursor = result.next_cursor
+    return cursor
```
rogercloud, line 364: **[P1] Continue promptly while the cursor is still live**

This raises the default capacity to 20 × 500 = 10,000 rows per hour/process, but the caller still takes the full 3,600-second sleep after budget exhaustion. One task-less widget IP can admit 60 requests/minute × 10 files = 36,000 orphan rows/hour (and an entity up to 144,000/hour), so supported traffic can still grow the expired backlog indefinitely. When this return value is non-null, continue after only a short yield/delay and reserve the long poll sleep for a short page that proves the eligible backlog was drained; alternatively, globally bound admitted task-less file rows below guaranteed GC throughput. This is a partial fix of the earlier capacity finding.

example.env
```diff
@@ -97,6 +97,13 @@ PORT="80"
 # XAGENT_BACKGROUND_JOB_STALE_SECONDS="7200"
 # Scheduler interval for scanning stale background jobs.
 # XAGENT_BACKGROUND_JOB_SWEEP_INTERVAL_SECONDS="300"
+# Public share-channel upload orphan GC (#973). A task-less public-share upload
+# is bound to its task at run start; if the guest never finishes task creation
+# it stays orphaned. An in-app background loop (runs in every deployment, no
+# Celery required) reaps rows older than the TTL that still carry the
+# task-less-share marker and have no task_id, one bounded page per tick.
```
rogercloud: **[P3] Document the multi-page tick budget**

This still says one bounded page per tick, but the updated loop can process up to 20 pages (10,000 rows by default) before sleeping. Update the operator-facing comment so it reflects the actual per-tick database and storage-delete load.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 386b29b2 — your arithmetic is right and the long sleep was the real cap, not the page budget. The loop now reserves the long `poll_interval_seconds` sleep for a **short page**, which is the only proof that the eligible backlog is drained; while the tick returns a live cursor the next page budget follows after a ~1s breather (`GC_BACKLOG_CONTINUE_DELAY_SECONDS`), so drain throughput is set by page cost rather than by the interval. A failed tick still takes the long sleep, so a persistent error can't hot-loop. `test_loop_defers_long_sleep_until_backlog_is_drained` pins the cadence (short, short, long-after-failure, long-after-drain) and that the live cursor is threaded into the immediately following tick. The page budget now only bounds work between breathers (~500 rows/s worst case), which stays well above the ingress figures you cite.

example.env
codeacme17: Fixed in 386b29b2. The operator comment now states the real per-tick load — up to 20 keyset pages x 500 rows (10,000 rows, each with a durable-object + local-file delete), one short DB session per page — and explains that while a backlog remains the loop continues after a ~1s breather rather than sleeping, so `XAGENT_ORPHAN_UPLOAD_SWEEP_INTERVAL_SECONDS` only sets how often an *already-drained* backlog is re-checked (renamed in the comment to "idle re-check interval").

codeacme17:
Rebased onto current `main` (`f2713f16`) to clear the conflict that appeared after #1044 and the other 10 commits landed. The only conflict was one import block in `app.py` (the new `skill_runtime` handler alongside the GC loop import) — no logic changes, and the migration chain is untouched (still a single head off `20260725_add_uploaded_file_recovery_index`).

Post-rebase verification: app import smoke, the GC / share-upload / widget / rate-limit / config suites, SQLite migration `all` round-trip, offline `--sql` generation both directions, and isort / ruff / mypy are all clean. Full `tests/web` shows only the known order-dependent `test_task_tracker` flake, which passes in isolation and in its own directory, and my changes touch nothing under `tracking/`.

src/xagent/web/services/orphan_upload_gc.py
```diff
@@ -0,0 +1,412 @@
...
+# independent of the poll interval: the rate-limited task-less upload
+# surfaces admit far more rows per hour than any single fixed-interval
+# page budget could reclaim, so a live cursor must be followed promptly
+# rather than parked until the next tick. The cursor also carries ACROSS
+# ticks (mirroring ``run_uploaded_file_compensation_recovery_loop``), so a
+# repeatedly failing oldest page cannot starve newer orphans.
+GC_BATCH_SIZE = 500
+GC_MAX_PAGES_PER_TICK = 20
+# Breather between page budgets while a backlog remains: bounds sustained
+# database/storage-delete load (~GC_BATCH_SIZE rows per this interval in the
```
rogercloud: [P3] This says each breather interval bounds the worst-case load to about `GC_BATCH_SIZE` rows, but `_run_gc_tick` can process `GC_MAX_PAGES_PER_TICK` full pages before the breather. With the defaults that is 20 × 500 = 10,000 rows, so this understates the sustained backlog load by 20×. Please describe the bound as `GC_MAX_PAGES_PER_TICK * GC_BATCH_SIZE`; `example.env` already documents it correctly.

rogercloud (approved):
Design is sound and the previously blocking GC capacity issue is fixed. The remaining inline P3 is a non-blocking documentation correction. Focused verification passed (419 tests in the main pass and 65 in an independent pass), along with Ruff and `git diff --check`; GitHub CI currently has no failures, with a few test jobs still running. A real-PostgreSQL GC/bind concurrency test was not run locally.

src/xagent/web/services/orphan_upload_gc.py
codeacme17: Fixed in 2dfaa195 — you're right, the constant's comment understated the bound by 20x. It now reads "at most GC_MAX_PAGES_PER_TICK * GC_BATCH_SIZE rows (10,000 at the defaults) per this interval", matching what example.env already documented. Thanks for the review throughout this one.

PR #998, opened 2026-07-24
fix(ssh): restore cross-seam SshErrorCode members removed as false dead code

bluefish-08:
The SSH MCP re-review removed four `SshErrorCode` members as unused, based on a grep of this repo only:

- `TARGET_IN_USE` / `CREDENTIAL_REVOKED` / `CREDENTIAL_IN_USE` / `HOST_KEY_UNVERIFIED`

These are **not raised inside core**, but they are part of the cross-seam error contract: out-of-repo consumers on the other side of the provider / secret-store / audit seam raise them, and core only *records* them via the audit sink. An in-repo grep therefore wrongly flags them as dead code, and removing them broke consumers — every consumer call site now hits `AttributeError` on the missing enum member.

This restores the four members with their original (contract) string values and documents that consumer-raised codes must not be pruned based on in-repo usage.

The other three removed at the same time — `EXTERNAL_POLICY_DENIED`, `CONNECTION_TIMEOUT`, `OUTPUT_LIMIT_EXCEEDED` — have no consumer and remain removed.

### DIFF
--- a/src/xagent/core/ssh/errors.py
+++ b/src/xagent/core/ssh/errors.py
@@ -7,11 +7,24 @@ from typing import Any
 
 
 class SshErrorCode(str, Enum):
-    """Stable, public error codes. Values are a contract; do not rename."""
+    """Stable, public error codes. Values are a contract; do not rename.
+
+    Several codes are part of the cross-seam contract: they are raised by
+    out-of-repo consumers on the other side of the provider / secret-store /
+    audit seam and only *recorded* by core's audit sink, never raised inside
+    this repo. A grep for raise-sites within this repo alone will find them
+    unused — do NOT remove them on that basis; that breaks the consumer
+    contract. TARGET_IN_USE / CREDENTIAL_REVOKED / CREDENTIAL_IN_USE /
+    HOST_KEY_UNVERIFIED are such consumer-raised codes.
+    """
 
     TARGET_NOT_FOUND = "ssh_target_not_found"
     TARGET_DISABLED = "ssh_target_disabled"
+    TARGET_IN_USE = "ssh_target_in_use"
     OPERATION_NOT_ALLOWED = "ssh_operation_not_allowed"
+    CREDENTIAL_REVOKED = "ssh_credential_revoked"
+    CREDENTIAL_IN_USE = "ssh_credential_in_use"
+    HOST_KEY_UNVERIFIED = "ssh_host_key_unverified"
     HOST_KEY_MISMATCH = "ssh_host_key_mismatch"
     EGRESS_DENIED = "ssh_egress_denied"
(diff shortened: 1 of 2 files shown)


rogercloud (approved):
## Review: `fix(ssh): restore cross-seam SshErrorCode members`

### Summary
A prior commit (`7f7baf2f`) deleted 7 `SshErrorCode` members after an in-repo grep showed them "never raised." That grep missed a cross-seam contract: 4 of the 7 (`TARGET_IN_USE`, `CREDENTIAL_REVOKED`, `CREDENTIAL_IN_USE`, `HOST_KEY_UNVERIFIED`) are raised by out-of-repo consumers via `SshErrorCode.X` attribute access, and are only *recorded* — not raised — inside core (the audit sink types `error_code` as `str | None`, so it never depended on the enum). Their removal broke consumer access with `AttributeError`. This PR restores exactly those 4 members with byte-identical contract values and expands the class docstring to warn against pruning consumer-raised codes on in-repo grep evidence alone.

### Approach: acceptable with reservations
The surgical restoration is correct — exactly 4 members, not a blanket revert of all 7 (the other 3 have no known consumer and correctly stay removed). Three non-blocking concerns:
1. **No enforcement** — the "don't prune" rule lives only in prose; nothing in CI would catch a re-deletion.
2. **Warning is class-level only** — not inline next to the 4 specific members.
3. **No durable in-tree pointer to the consumer** — the cross-seam evidence is real (a sibling package references exactly these 4 members), but it lives in an unmerged/unpushed branch, invisible to an in-repo grep.

### Finding — medium: test suite still can't catch this regression
`tests/core/ssh/test_errors.py` (unmodified by this PR, so no inline anchor available in this diff) still doesn't guard the 4 restored codes. `test_error_codes_have_stable_string_values` pins string values for 5 other codes but none of `TARGET_IN_USE` / `CREDENTIAL_REVOKED` / `CREDENTIAL_IN_USE` / `HOST_KEY_UNVERIFIED`, and `test_all_codes_are_lowercase_ssh_prefixed` only checks format, not membership. A repo-wide grep for these 4 names/values across `tests/` returns zero hits — so deleting or renaming them again would pass CI silently, the exact failure mode this PR exists to prevent. Suggest adding explicit value assertions for the 4 restored codes (or a dedicated test pinning the full cross-seam set as a frozen collection).

### Testing
Verified that the enum imports cleanly and all 4 restored values match their pre-removal originals exactly (`ssh_target_in_use`, `ssh_credential_revoked`, `ssh_credential_in_use`, `ssh_host_key_unverified`). Full pytest run wasn't possible in this environment; given this is a purely additive enum/docstring change with no logic, import verification is sufficient confidence.

PR #904, opened 2026-07-16
feat(connectors,agents,knowledge-base): opt-in team ownership via application hooks

bluefish-08:
## What

Adds an opt-in extension surface so a multi-tenant application can overlay team
ownership on connectors, knowledge bases, and agents without teaching xagent
about the application's team tables. Standalone xagent is unchanged: with no
hooks installed, connectors and collections stay scoped to the owning user, and
no team UI is shown.

## Hooks

- `connector_team_scope`: visibility, delete, rename, and credential-policy
  hooks for team-owned MCP servers and Custom APIs. `ConnectorDeleteDecision`
  reports whether a connector was team-owned, who is authorized, whether the
  shared definition should be dropped, and `blocked_reason` when a delete is
  refused.
- `knowledge_base_team_scope`: visibility and access-resolution hooks that map a
  logical team collection to the physical storage user, plus rename/delete
  lifecycle notifications.
- `agent_team_scope`: team-scope resolution and validators that reject promoting
  a team agent whose selected connectors / knowledge bases are not shared with
  the team.

## Delete safety

The MCP and Custom API delete endpoints honor
`ConnectorDeleteDecision.blocked_reason`, returning HTTP 409 before any mutation
when the application reports the connector is still selected by a team agent —
mirroring the existing unshare guard, so deleting a definition can no longer
orphan a team agent's selection.

## Frontend

Agent builder, MCP connect/settings dialogs, and the knowledge-base page gain
personal-vs-team ownership controls (with en/zh strings). All team UI is gated
behind the `inTeam` auth flag, so standalone xagent (no team, `my-team` 404)
shows none of it.

## Backward compatibility

All hooks default to the standalone (user-owned) behavior when unset; existing
single-tenant deployments and tests are unaffected.

## Testing

`agent_team_scope`, `agent_visibility`, and the MCP / Custom API delete
endpoints are covered by unit tests; the multi-tenant overlay is exercised in
the downstream application's suite. Frontend components have vitest coverage.

PR #1003, opened 2026-07-25
fix(kb): run collection scans off the event loop, allow downstream job handlers

bluefish-08:
## Why

`_list_collections_impl` is `async def` but scans the documents table and
aggregates stats synchronously, so every caller stalls the event loop for the
scan's duration. Callers include `GET /api/kb/collections` and, through
`document_search`, the `search_knowledge_base` / `list_knowledge_bases` agent
tools — so one agent's knowledge-base lookup freezes the whole process. Since
`web/__main__.py` runs uvicorn without `workers`, that is the entire API,
`/health` included.

Separately, long-running work belongs in the Celery worker, but
`execute_background_job` — the single entry point that retries and the
scheduler's stale-job requeue funnel through — only dispatches job types defined
in this package.

## Changes

- `_scan_document_rows` extracted and dispatched via `asyncio.to_thread`, same
  for `aggregate_collection_stats`. Regression test counts event-loop heartbeats
  during a slow scan.
- `register_background_job_handler()`: a downstream package can claim a job type
  at worker import time and keep the shared retry/requeue path. Unknown types
  still raise.
- Frontend: when an ownership endpoint answers `202` with a job (a downstream
  distribution provides that endpoint), poll to a terminal status before
  reporting success; `204` keeps the old synchronous path. Also repairs
  `knowledge-base.test.tsx`, red on `main` because its `@/lib/utils` mock
  stripped `cn` — the widget CI job never runs that file.

## Verification

`pre-commit run --all-files`, `pytest tests/core/tools/core/RAG_tools
tests/web/test_background_jobs.py`, the frontend tests and `npm run build` all
pass.

Real single-worker server, 400k-row `documents` table, one listing while probing
`/health` every 20ms:

|        | listing | probes | peak `/health`      |
| ------ | ------- | ------ | ------------------- |
| before | 5.28s   | 12     | 5222ms              |
| after  | 4.02s   | 24     | 2346ms (median 2ms) |

The residual spike is GIL contention from the per-row `as_py()` loop, not loop
starvation; vectorising it is a separate change.

### DIFF
--- a/frontend/src/components/build/agent-builder.tsx
+++ b/frontend/src/components/build/agent-builder.tsx
@@ -1364,13 +1364,17 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
         // the job instead of racing it. Backends without a job queue send 204.
         if (res.status === 202) {
           const job = await res.json().catch(() => null)
-          if (isBackgroundJobResponse(job)) {
-            const finished = await waitForBackgroundJob(getApiUrl(), job)
-            if (finished.status !== "succeeded") {
-              throw new Error(
-                getBackgroundJobFailureMessage(finished, t("builds.editor.error.unknown")),
-              )
-            }
+          // An unreadable 202 body leaves the transfer untrackable, so the KB
+          // may still be personal. Continuing would race the agent promotion
+          // that this wait exists to order.
+          if (!isBackgroundJobResponse(job)) {
+            throw new Error(t("builds.editor.error.unknown"))
+          }
+          const finished = await waitForBackgroundJob(getApiUrl(), job)
+          if (finished.status !== "succeeded") {
+            throw new Error(
+              getBackgroundJobFailureMessage(finished, t("builds.editor.error.unknown")),
+            )
           }
         }
       }
--- a/frontend/src/components/build/agent-builder.tsx
+++ b/frontend/src/components/build/agent-builder.tsx
@@ -9,6 +9,11 @@ import { Button } from "@/components/ui/button"
 import { Label } from "@/components/ui/label"
 import { Switch } from "@/components/ui/switch"
 import { apiRequest, isJsonRecord } from "@/lib/api-wrapper"
+import {
+  getBackgroundJobFailureMessage,
+  isBackgroundJobResponse,
+  waitForBackgroundJob,
+} from "@/lib/background-jobs"
 import { getApiUrl } from "@/lib/utils"
 import { isBuiltinModel, hostnameFromUrl } from "@/lib/models"
 import { PlusCircle, MessageSquare, Upload, Settings2, Check, Zap, BookOpen, Gauge, Sparkles, Loader2, X, XCircle, Trash2, Bot, Brain, Webhook, CalendarClock, Mail, Eye, Workflow, AlertCircle, Copy } from "lucide-react"
@@ -1353,6 +1358,20 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
         if (!res.ok) {
           const body = await res.json().catch(() => ({}))
           throw new Error(body?.detail?.message ?? body?.detail ?? t("builds.editor.error.unknown"))
+        }
+        // 202 means the transfer runs as a durable background job. Agent
+        // promotion below requires the KB to already be team-owned, so wait for
+        // the job instead of racing it. Backends without a job queue send 204.
+        if (res.status === 202) {
+          const job = await res.json().catch(() => null)
+          if (isBackgroundJobResponse(job)) {
+            const finished = await waitForBackgroundJob(getApiUrl(), job)
+            if (finished.status !== "succeeded") {
+              throw new Error(
+                getBackgroundJobFailureMessage(finished, t("builds.editor.error.unknown")),
+              )
+            }
+          }
         }
       }
       setUnsharedConnectors([])
--- a/frontend/src/components/pages/knowledge-base.tsx
+++ b/frontend/src/components/pages/knowledge-base.tsx
@@ -11,6 +11,11 @@ import { useI18n } from "@/contexts/i18n-context"
 import { useAuth } from "@/contexts/auth-context"
 import { apiRequest } from "@/lib/api-wrapper"
 import {
+  getBackgroundJobFailureMessage,
+  isBackgroundJobResponse,
+  waitForBackgroundJob,
+} from "@/lib/background-jobs"
+import {
   Plus,
   FileText,
   FolderOpen,
@@ -183,6 +188,18 @@ export function KnowledgeBasePage() {
       if (!response.ok) {
         const body = await response.json().catch(() => ({}))
         throw new Error(typeof body.detail === "string" ? body.detail : t("kb.ownership.failed"))
+      }
+      // 202 means the transfer runs as a durable background job: the physical
+      // move can take minutes, so wait for it instead of reporting success on
+      // acceptance. Backends without a job queue still answer 204 synchronously.
+      if (response.status === 202) {
+        const job = await response.json().catch(() => null)
+        if (isBackgroundJobResponse(job)) {
+          const finished = await waitForBackgroundJob(getApiUrl(), job)
+          if (finished.status !== "succeeded") {
+            throw new Error(getBackgroundJobFailureMessage(finished, t("kb.ownership.failed")))
+          }
+        }
       }
       toast.success(
         collection.ownership === "team"
--- a/frontend/src/components/pages/knowledge-base.tsx
+++ b/frontend/src/components/pages/knowledge-base.tsx
@@ -194,11 +194,15 @@ export function KnowledgeBasePage() {
       // acceptance. Backends without a job queue still answer 204 synchronously.
       if (response.status === 202) {
         const job = await response.json().catch(() => null)
-        if (isBackgroundJobResponse(job)) {
-          const finished = await waitForBackgroundJob(getApiUrl(), job)
-          if (finished.status !== "succeeded") {
-            throw new Error(getBackgroundJobFailureMessage(finished, t("kb.ownership.failed")))
-          }
+        // An unreadable 202 body leaves the transfer untrackable. Falling
+        // through here would report success for work that may still be running,
+        // which is the very race this wait exists to close.
+        if (!isBackgroundJobResponse(job)) {
+          throw new Error(t("kb.ownership.failed"))
+        }
+        const finished = await waitForBackgroundJob(getApiUrl(), job)
+        if (finished.status !== "succeeded") {
+          throw new Error(getBackgroundJobFailureMessage(finished, t("kb.ownership.failed")))
         }
       }
       toast.success(
--- a/src/xagent/core/tools/core/RAG_tools/management/collections.py
+++ b/src/xagent/core/tools/core/RAG_tools/management/collections.py
@@ -6,6 +6,7 @@ system, including listing collections, managing documents, and handling deletion
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 import os
@@ -569,6 +570,128 @@ async def _load_collection_ingestion_configs(
     return collection_configs
 
 
+def _scan_document_rows(
+    vector_store: Any,
+    *,
+    user_id: Optional[int],
+    is_admin: Optional[bool],
+) -> tuple[
+    Dict[str, Set[str]],
+    Dict[str, Set[int]],
+    Dict[str, List[CollectionDocumentMetadata]],
+]:
+    """Scan the documents table for collection names, owners and metadata.
+
+    Pure blocking LanceDB work. It must run in a worker thread: on the event
+    loop it stalls every other coroutine in the process, which on a
+    single-worker deployment means the whole API, ``/health`` included.
+    """
+    document_names: Dict[str, Set[str]] = defaultdict(set)
+    owners: Dict[str, Set[int]] = defaultdict(set)
+    document_metadata: Dict[str, List[CollectionDocumentMetadata]] = defaultdict(list)
+    document_metadata_seen: Dict[str, Set[tuple[str, str, str]]] = defaultdict(set)
+
+    def _normalize_optional_identifier(value: Any) -> Optional[str]:
+        if not isinstance(value, str):
+            return None
+        normalized = value.strip()
+        return normalized or None
+
+    def _add_document_entry(
+        collection_key: str,
+        source_value: Any,
+        doc_id_value: Any,
+        file_id_value: Any,
+    ) -> None:
+        normalized_doc_id = _normalize_optional_identifier(doc_id_value)
+        normalized_file_id = _normalize_optional_identifier(file_id_value)
+        display_name = None
+        if source_value:
+            display_name = os.path.basename(str(source_value))
+        display_name = (display_name or normalized_doc_id or "").strip()
+        if not display_name:
+            return
+
+        document_names[collection_key].add(display_name)
+
+        dedupe_key = (
+            display_name,
+            normalized_file_id or "",
+            normalized_doc_id or "",
+        )
+        seen_keys = document_metadata_seen[collection_key]
+        if dedupe_key in seen_keys:
+            return
+        seen_keys.add(dedupe_key)
+        document_metadata[collection_key].append(
+            CollectionDocumentMetadata(
+                filename=display_name,
+                file_id=normalized_file_id,
+                doc_id=normalized_doc_id,
+            )
+        )
+
+    for batch in vector_store.iter_batches(
+        table_name="documents",
+        columns=[
+            "collection",
+            "source_path",
+            "doc_id",
+            "file_id",
+            "user_id",
+        ],
+        user_id=user_id,
+        is_admin=is_admin,
+    ):
+        collection_idx = batch.schema.get_field_index("collection")
+        source_idx = batch.schema.get_field_index("source_path")
+        doc_id_idx = batch.schema.get_field_index("doc_id")
+        file_id_idx = batch.schema.get_field_index("file_id")
+        user_idx = batch.schema.get_field_index("user_id")
+        if collection_idx == -1:
+            continue
+        collection_array = batch.column(collection_idx)
+        source_array = (
+            batch.column(source_idx)
+            if source_idx != -1
+            else pa.array([None] * batch.num_rows)
+        )
+        doc_id_array = (
+            batch.column(doc_id_idx)
+            if doc_id_idx != -1
+            else pa.array([None] * batch.num_rows)
+        )
+        file_id_array = (
+            batch.column(file_id_idx)
+            if file_id_idx != -1
+            else pa.array([None] * batch.num_rows)
+        )
+        user_array = (
+            batch.column(user_idx)
+            if user_idx != -1
+            else pa.array([None] * batch.num_rows)
+        )
+        for idx in range(batch.num_rows):
+            collection_raw = collection_array[idx].as_py()
+            if not collection_raw:
+                continue
+            collection_key = str(collection_raw)
+            _add_document_entry(
+                collection_key,
+                source_array[idx].as_py(),
+                doc_id_array[idx].as_py(),
+                file_id_array[idx].as_py(),
+            )
+            user_val = user_array[idx].as_py()
+            if user_val is not None:
+                try:
+                    owners[collection_key].add(int(user_val))
+                except (TypeError, ValueError):
+                    pass
+
+    return document_names, owners, document_metadata
+
+
 async def list_collections(
     user_id: Optional[int] = None,
     is_admin: Optional[bool] = None,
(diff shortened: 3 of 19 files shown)


rogercloud:
ci is failing

src/xagent/core/tools/core/RAG_tools/management/collections.py
```diff
@@ -652,112 +775,14 @@ async def _list_collections_impl(
...
-        for batch in vector_store.iter_batches(
-            table_name="documents",
-            columns=[
-                "collection",
-                "source_path",
-                "doc_id",
-                "file_id",
-                "user_id",
-            ],
+        document_names, owners, document_metadata = await asyncio.to_thread(
```
rogercloud, line 790: This `to_thread` dispatch is correct, but it makes a previously-unreachable timeout reachable. `src/xagent/web/api/kb.py:4739-4745` wraps `list_collections(...)` in `asyncio.wait_for(..., timeout=15)`; before this change the scan monopolized the loop so that deadline's timer callback could never fire mid-scan — slow scans just returned successfully. Now a deployment whose table takes ~20s gets a hard 503 after 15s instead. Also note `wait_for` cancels the awaiting coroutine, not the `to_thread` worker: the scan runs to completion in the shared default executor after the request already failed, so repeated timeouts accumulate orphaned scans competing for GIL time with the ~233 other `to_thread` sites. Please raise or make `kb_collections_timeout_seconds` configurable here, and consider a dedicated bounded executor for scans. (Related, but pre-existing/out-of-diff code not commentable inline here: `metadata_store.list_collections`/`get_collection_config`/`save_collections` in this same function are also unwrapped sync LanceDB calls — see review body.)

src/xagent/web/jobs/tasks.py
```diff
@@ -49,6 +65,10 @@ def _execute_job_handler(db: Session, job: BackgroundJob) -> dict[str, Any]:
 
         return handle_trigger_scan(db, job)
 
+    handler = _EXTRA_HANDLERS.get(str(job.job_type))
+    if handler is not None:
+        return handler(db, job)
+
     raise ValueError(f"Unsupported background job type: {job.job_type}")
```
rogercloud: An unknown `job.job_type` raises a plain `ValueError`, which is unrelated to `BackgroundJobHandlerError` (a `RuntimeError` subclass with a `retryable` flag), so it falls into `execute_background_job`'s generic `except Exception` at `:108-117` and retries unconditionally up to `max_attempts` with backoff. "This worker doesn't know this job type" is not transient — and because nothing guarantees every worker has the downstream `_EXTRA_HANDLERS` registration (see review body), this is a realistic path. Please raise `BackgroundJobHandlerError(..., retryable=False)` here so it fails fast.

src/xagent/web/jobs/tasks.py
```diff
@@ -21,6 +22,21 @@
...
+BackgroundJobHandler = Callable[[Session, BackgroundJob], dict[str, Any]]
+
+# Downstream distributions own job types this package must not import. They
+# register a handler when the worker imports their task module, so their work
+# still flows through execute_background_job and keeps the shared retry and
+# stale-requeue behaviour instead of needing a parallel Celery task.
+_EXTRA_HANDLERS: dict[str, BackgroundJobHandler] = {}
+
+
+def register_background_job_handler(
```
rogercloud, line 38: Two issues: (1) `_EXTRA_HANDLERS[str(job_type)] = handler` silently overwrites on duplicate registration, unlike the repo's sibling `register_trigger_provider` (`services/trigger_providers/registry.py:15-21`), which raises `ValueError` unless `replace=True`; (2) `_execute_job_handler` checks the built-in if-chain *before* `_EXTRA_HANDLERS`, so a handler registered under a string colliding with one of the 4 built-in `BackgroundJobType` values silently never fires, with no error at registration time. Mirroring the existing registry pattern would close both.

src/xagent/web/jobs/tasks.py
```diff
@@ -21,6 +22,21 @@
 
 logger = logging.getLogger(__name__)
 
+BackgroundJobHandler = Callable[[Session, BackgroundJob], dict[str, Any]]
+
```
rogercloud, line 26: This registry has no in-repo consumer and no way to be reached: the comment says downstream distributions register "when the worker imports their task module," but `register_celery_tasks()` (`jobs/celery_app.py:59-65`) imports only a hardcoded tuple, there is no env var / entry-points / config-driven module list (contrast `XAGENT_EXTERNAL_SKILLS_LIBRARY_DIRS`), and the compose worker command has no `--include`. Suggest either landing the module-import mechanism in this same PR, or deferring the registry until a real caller exists. Separately, if this is meant to be a public extension point, export it from `src/xagent/web/jobs/__init__.py` (currently just a docstring) rather than making callers import from a module full of private names.

frontend/src/components/pages/knowledge-base.tsx
```diff
@@ -184,6 +189,18 @@ export function KnowledgeBasePage() {
         const body = await response.json().catch(() => ({}))
         throw new Error(typeof body.detail === "string" ? body.detail : t("kb.ownership.failed"))
       }
+      // 202 means the transfer runs as a durable background job: the physical
+      // move can take minutes, so wait for it instead of reporting success on
+      // acceptance. Backends without a job queue still answer 204 synchronously.
+      if (response.status === 202) {
+        const job = await response.json().catch(() => null)
+        if (isBackgroundJobResponse(job)) {
+          const finished = await waitForBackgroundJob(getApiUrl(), job)
```
rogercloud: Two things here. (1) This new `waitForBackgroundJob` call omits the optional `onUpdate` callback that all 4 pre-existing call sites pass (`knowledge-base-detail.tsx`, `knowledge-base-creation-dialog.tsx`), so the user gets no progress feedback during what the PR describes as a potentially multi-minute physical move — and `requeue_stale_background_jobs` can reset a crashed job back to non-terminal `pending`, extending the wait. (2) The `if (response.status === 202)` / `if (isBackgroundJobResponse(job))` nesting at `:195-209` has no `else` and no throw: if the 202 body fails the 4-field structural check (`background-jobs.ts:25-34`) — version skew, gateway stripping the body — execution falls through to `toast.success(...)` + `fetchCollections()`, reproducing on the degradation path the exact bug this PR fixes on the happy path. Please pass `onUpdate` and throw on an unrecognized 202 body. (Separate, pre-existing issue also worth flagging: the `/api/knowledge-bases/{name}/promote-team` route this calls does not exist anywhere in this backend — see review body.)

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1354,6 +1359,20 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
           const body = await res.json().catch(() => ({}))
           throw new Error(body?.detail?.message ?? body?.detail ?? t("builds.editor.error.unknown"))
         }
+        // 202 means the transfer runs as a durable background job. Agent
+        // promotion below requires the KB to already be team-owned, so wait for
+        // the job instead of racing it. Backends without a job queue send 204.
+        if (res.status === 202) {
+          const job = await res.json().catch(() => null)
+          if (isBackgroundJobResponse(job)) {
+            const finished = await waitForBackgroundJob(getApiUrl(), job)
```
rogercloud: Same pair as in `knowledge-base.tsx`: no `onUpdate` progress callback (unlike the 4 pre-existing call sites), and the `202` / `isBackgroundJobResponse` block at `:1362-1384` has no `else`/throw — an unparseable 202 body falls through and the code proceeds as if the KB is already team-owned, continuing agent promotion on an unverified transfer. Please pass `onUpdate` and throw on an unrecognized body.

tests/core/tools/core/RAG_tools/kb/test_list_collections_event_loop.py
```diff
@@ -0,0 +1,109 @@
...
+            user_id=1, is_admin=False
+        )
+    finally:
+        beat.cancel()
+
+    assert result.status == "success"
+    assert [collection.name for collection in result.collections] == ["kb1"]
+    # Two 0.5s blocking calls: a responsive loop ticks tens of times. Held on
+    # the event loop thread it manages one or two.
+    assert ticks >= 20, f"event loop was blocked, only {ticks} ticks"
```
rogercloud: `assert ticks >= 20` cannot catch a partial reversion. Verified experimentally: reverting *only* the `_scan_document_rows` dispatch (`collections.py:~780`) to a direct blocking call, while leaving `aggregate_collection_stats`'s dispatch (`~885`) intact, still passes at ~24 ticks — the remaining call's 0.5s of yielded time alone clears the bar. Suggest asserting each phase's heartbeat count separately, or raising the threshold toward ~100 (roughly 2× the ~48-tick both-blocking baseline).

tests/web/test_background_jobs.py
```diff
@@ -1427,3 +1427,59 @@ def test_requeue_stale_background_jobs_marks_old_running_pending(tmp_path, monke
         assert job.progress["message"] == "Requeued stale background job"
     finally:
         db.close()
+
+
+def test_registered_external_handler_receives_job(tmp_path):
```
rogercloud, line 1436: These two tests call the private `_execute_job_handler` directly and never go through the Celery task `execute_background_job` — and no test in this file does (only a mocked `.apply_async`). So nothing verifies that a registered handler raising `BackgroundJobHandlerError(retryable=True)` flows through the retry path (job reset to `ENQUEUED`, `self.retry(...)` called). Pre-existing gap that also affects the built-in handlers, but the new feature is a good place to close it: one end-to-end test driving `execute_background_job` with a retryable-failing registered handler.

rogercloud (commented):
## Summary

This PR bundles three unrelated changes. **(1) Event-loop fix:** `_list_collections_impl` in `src/xagent/core/tools/core/RAG_tools/management/collections.py` was declared `async def` but executed synchronous blocking LanceDB scans directly on the event loop, stalling the entire single-worker API process (including `/health`) for the duration of a scan; the fix extracts `_scan_document_rows` and dispatches it plus `aggregate_collection_stats` through `asyncio.to_thread`. **(2) Job-handler extensibility:** a new `register_background_job_handler()` in `src/xagent/web/jobs/tasks.py` lets a downstream distribution register a handler for a job type this package does not define, so it still flows through the shared `execute_background_job` Celery task and its retry/stale-requeue machinery. **(3) Frontend polling:** when a KB-ownership-change endpoint answers `202` with a job body, `knowledge-base.tsx` and `agent-builder.tsx` now poll to a terminal status via the pre-existing `waitForBackgroundJob` helper before reporting success, instead of declaring success on mere acceptance; the PR also repairs a pre-broken `knowledge-base.test.tsx` whose `vi.mock("@/lib/utils")` replaced the whole module and stripped the `cn` export needed by UI primitives.

## Approach: acceptable with reservations

The `asyncio.to_thread` approach is correct in kind, not a new invention — it is this repo's established convention for sync LanceDB work (`RAG_tools/kb/coordinator.py` alone wraps 34 call sites this way), and it is the right altitude: switching to a fully async DB driver or pushing this interactive read onto Celery would cost far more than the problem is worth. The job-handler registry and the frontend reuse of `waitForBackgroundJob` are likewise built on existing patterns rather than new machinery. The frontend test fix (`importOriginal`-based partial mock) is a genuine fix — reproduced red on base, green on HEAD.

Three design-level reservations, none of which belongs to a single line:

**The event-loop fix stops short of its own boundary (D1).** Several synchronous LanceDB calls remain directly in `_list_collections_impl`'s async body, with no `to_thread` and no genuine internal `await` of real I/O (all three are "async in name only"):
- `collections.py:744-749` → `metadata_store.list_collections(...)` (`storage/lancedb_stores.py:189-227`, synchronous `table.search().to_arrow()`)
- `collections.py:552-556` → `get_collection_config(...)` (`lancedb_stores.py:381-435`), called **serially inside the `for collection in collection_keys` loop** in `_load_collection_ingestion_configs`. This is the worst of the three: its blocking cost scales linearly with collection count, so in aggregate it is arguably a bigger event-loop hazard than either call this PR did fix.
- `collections.py:925` → `save_collections(...)` (`lancedb_stores.py:161-187`, synchronous `merge_insert(...).execute()`)

Given the 34-site convention in `coordinator.py`, the partial coverage here reads as an inconsistency rather than a deliberate scope line. At minimum, wrap `get_collection_config`'s loop.

**The job-handler registry has no consumer in this repository (D5).** `register_background_job_handler`/`_EXTRA_HANDLERS` (`src/xagent/web/jobs/tasks.py:26-38`) is documented as something downstream distributions use "when the worker imports their task module" — but `register_celery_tasks()` (`src/xagent/web/jobs/celery_app.py:59-65`) imports only a hardcoded tuple (`xagent.web.jobs.tasks`, `xagent.web.jobs.trigger_tasks`). There is no env var, no entry-points discovery, no config-driven extra-module list (contrast `XAGENT_EXTERNAL_SKILLS_LIBRARY_DIRS` in `config.py`), and the `docker-compose.yml` worker command carries no `--include`. `BackgroundJobType` (`models/background_job.py:21-25`) has no team-transfer values, and there is no first-party caller. As merged, the registry is inert scaffolding: reachable only from a fork that also patches `register_celery_tasks` or relies on import side effects. Either land it together with the module-import mechanism that makes it usable, or defer it until a real second consumer exists.

**The frontend 202 path targets a route that does not exist here (D6).** `frontend/src/components/pages/knowledge-base.tsx:185` and `frontend/src/components/build/agent-builder.tsx:1355` both `POST /api/knowledge-bases/{name}/promote-team` (and `demote-personal`). Across all 28 `APIRouter(prefix=...)` declarations in `src/xagent/web/api/*.py` there is no `/api/knowledge-bases` prefix; the only `promote-team`/`demote-personal` endpoints belong to a different resource, `/api/agents/{agent_id}/promote-team` (`agents.py:783,822`), which returns a synchronous `AgentResponse` with no 202 branch. **This broken call is pre-existing** — confirmed present in the base commit, so this PR did not break it. But the PR builds new 202/job-polling logic directly on top of it, so that logic is unreachable and unverifiable against any backend in this repo; the 3 new frontend tests mock `apiRequest` directly and never exercise a route.

**Recommendation: split this into separate PRs.** The three changes have zero technical coupling and materially different risk profiles. Item 1 is independently tested and mergeable today (modulo F1/D1 below). Items 2 and 3 cannot be exercised against any real backend in this repository and only make sense paired with a downstream fork that supplies the missing worker-import mechanism and the missing `/api/knowledge-bases/*` route. Reviewing them under one approval flattens confidence levels that should stay very different.

## Findings

### Major

**1. The 15s collections timeout was structurally unreachable and is now reachable — undiscussed 503 regression, plus orphaned scan threads.** `src/xagent/web/api/kb.py:4739-4745` (pre-existing, unmodified) sets `kb_collections_timeout_seconds = 15` and wraps `list_collections(...)` in `asyncio.wait_for(..., timeout=15)`, returning HTTP 503 "Knowledge base is temporarily unavailable. Please retry." on expiry (`kb.py:~5023-5031`). Before this PR the scan and `aggregate_collection_stats` ran as bare synchronous code inside an `async def` with no `await`, monopolizing the single-threaded loop so the deadline's timer callback could never fire mid-scan — the timeout was effectively dead during scans, and slow scans simply returned successfully after however long they took. Wrapping those calls in `asyncio.to_thread` yields control back to the loop, so the 15s deadline can now fire mid-scan for the first time. Failure scenario: a deployment whose table previously took ~20s and returned data now gets a hard 503 after exactly 15s. Neither the PR's benchmark (400k rows = 4.02s, well inside the budget) nor its new tests cover this. Compounding it: `asyncio.wait_for` cancels the *awaiting coroutine*, not the `to_thread` worker — `run_in_executor`/`ThreadPoolExecutor` has no cancellation, so the scan runs to completion in the shared default executor (used by ~233 other `to_thread` sites repo-wide) after the request already failed, and repeated timeouts/retries accumulate orphaned scans competing for GIL time. Suggestion: raise or make `kb_collections_timeout_seconds` configurable in this PR, and state the intended behavior for large tables; consider a dedicated bounded executor for scans so orphaned work cannot starve unrelated `to_thread` callers.

**2. Unknown job types burn through the full retry budget instead of failing fast.** `src/xagent/web/jobs/tasks.py:72` raises a plain `ValueError` when `job.job_type` matches no built-in branch and is absent from `_EXTRA_HANDLERS` (pre-existing fallback, now reachable via a new path). `ValueError` is unrelated to `BackgroundJobHandlerError` (a `RuntimeError` subclass carrying a `retryable` flag), so it lands in `execute_background_job`'s generic `except Exception` branch (`:108-117`), which retries unconditionally up to `max_attempts` with backoff before failing. Failure scenario: worker A (without the downstream module imported) picks up a `kb.*` job enqueued by worker B (with it) — because per D5 nothing guarantees the registration is present in every worker — and the job cycles through every retry attempt and its backoff before dying, when "this worker doesn't know this job type" is not a transient condition retrying can fix. Suggestion: raise `BackgroundJobHandlerError(..., retryable=False)` for the unknown-job-type case so it fails fast.

**3. `knowledge-base.test.tsx` — and most of `frontend/src` — is invisible to CI, so this class of breakage recurs immediately.** `.github/workflows/ci.yml:640-642` runs `npm run test:widget:coverage`, driven by `frontend/vitest.widget.config.ts` with a hardcoded `include` covering exactly 2 widget files. No CI job anywhere runs the broader `frontend/vitest.config.ts` (`include: ["src/**/*.test.ts", "src/**/*.test.tsx"]`) that would pick up `knowledge-base.test.tsx`. This PR's repair of that file is real and correct, but the reason it was silently broken in the first place — a shared-mock change breaking an un-run test file — is left fully in place. Failure scenario: the next `@/lib/utils` or shared-mock edit re-breaks these tests and CI stays green until someone runs vitest by hand. Suggestion: in this PR or an immediate follow-up, add a CI-invoked vitest command covering `src/components/pages/**` (or the full default config).

### Minor

**4. `waitForBackgroundJob`'s two new call sites give the user no progress feedback during a potentially multi-minute move.** `frontend/src/lib/background-jobs.ts:82-105` (pre-existing, byte-identical to base) polls `GET /api/jobs/{id}` every 1000ms in a `while (!isBackgroundJobTerminal(job))` loop with no attempt cap, no overall timeout, no `AbortController`, and throws immediately on a single non-ok response with no retry or backoff. The new call sites (`knowledge-base.tsx:198`, `agent-builder.tsx:1368`) omit the optional `onUpdate` callback that all 4 pre-existing call sites pass (`knowledge-base-detail.tsx`, `knowledge-base-creation-dialog.tsx`). Failure scenario: `requeue_stale_background_jobs` resets a crashed job back to non-terminal `pending`, and the user stares at an unlabeled spinner through repeated cycles of what the PR description itself calls a "physical move." Not unbounded in the worst case (a server-side `max_attempts` backstop exists) and no data is corrupted — KB ownership is durable server-side regardless of client behavior — so this is a UX gap, not a correctness one. Suggestion: pass `onUpdate` at both new sites; consider a bounded client-side timeout and a cancel affordance in `waitForBackgroundJob` itself.

**5. Silent overwrite on duplicate handler registration, and built-in job types shadow registered handlers.** `src/xagent/web/jobs/tasks.py:34-38` does `_EXTRA_HANDLERS[str(job_type)] = handler` with no duplicate check — inconsistent with the repo's own sibling pattern `register_trigger_provider` (`services/trigger_providers/registry.py:15-21`), which raises `ValueError` on duplicates unless `replace=True`. Separately, `_execute_job_handler` (`:50-72`) checks the built-in if-chain *before* consulting `_EXTRA_HANDLERS`, so a handler registered under a string colliding with one of the 4 built-in `BackgroundJobType` values silently never fires, with no error at registration time. Low likelihood under today's single-registrant-at-import usage, but a silent footgun with an established fix already in-tree. Suggestion: mirror `register_trigger_provider` — reject duplicates unless `replace=True`, and reject job-type strings that collide with built-ins.

**6. The 202 path silently falls through to success reporting when the job body doesn't parse.** `frontend/src/components/pages/knowledge-base.tsx:195-209` and `frontend/src/components/build/agent-builder.tsx:1362-1384` wrap the poll in `if (response.status === 202) { ...; if (isBackgroundJobResponse(job)) {...} }` with no `else` and no throw. Failure scenario: the backend returns 202 with a body that fails `isBackgroundJobResponse`'s 4-field structural check (`background-jobs.ts:25-34`) — version skew, a gateway stripping the body, a field-name mismatch — and execution falls straight through to `toast.success(...)` + `fetchCollections()`, or in `agent-builder`'s case proceeds as if the KB is already team-owned and continues agent promotion. That reproduces, on the degradation path, exactly the "success declared before work is done" bug this PR fixes on the happy path. Suggestion: throw on an unrecognized 202 body rather than falling through.

**7. The event-loop regression test cannot catch a partial reversion.** `tests/core/tools/core/RAG_tools/kb/test_list_collections_event_loop.py:81`, `test_list_collections_keeps_event_loop_responsive`, asserts `ticks >= 20`. Verified experimentally: reverting *only* the `_scan_document_rows` dispatch (`collections.py:~780`) to a direct blocking call while leaving `aggregate_collection_stats`'s dispatch (`~885`) intact still passes at ~24 ticks, because the remaining call's 0.5s of yielded time alone clears the threshold. Failure scenario: a future refactor drops one of the two `to_thread` calls and the regression test stays green. Suggestion: assert each phase's heartbeat count separately, or raise the threshold toward ~100 (about 2× the ~48-tick both-blocking baseline).

**8. No test exercises the new registry through the actual Celery task or its retry path.** `tests/web/test_background_jobs.py:1432,1469` (`test_registered_external_handler_receives_job`, `test_unregistered_job_type_still_raises`) call the private `_execute_job_handler` directly, bypassing `execute_background_job` entirely; no test anywhere in that file calls `execute_background_job` (only a mocked `.apply_async`). So nothing verifies that a registered handler raising `BackgroundJobHandlerError(retryable=True)` flows correctly through the retry path (job reset to `ENQUEUED`, `self.retry(...)` invoked). This is a pre-existing coverage gap that also affects the built-in handlers, not a regression — but the new feature was the natural place to close it. Suggestion: add one test that drives `execute_background_job` end-to-end with a registered handler raising a retryable error.

**Also noticed (follow-up, not blocking):** `src/xagent/core/tools/core/RAG_tools/management/collection_manager.py:1202`, `_rebuild_collection_stats_impl` (`async def`), calls `get_vector_index_store().aggregate_collection_stats(...)` with no `to_thread` — the identical bug class this PR fixes elsewhere. Confirmed by full-repo grep to have zero production callers (reachable only via `rebuild_collection_stats`/`rebuild_collection_stats_sync`, themselves referenced only from their own test file), so it is dead code today. Worth a follow-up note so it doesn't become live with the bug intact.

**9. `register_background_job_handler` is not exported from the package.** `src/xagent/web/jobs/__init__.py` contains only a one-line docstring, so external callers must import from `xagent.web.jobs.tasks` — a module that also holds genuinely private names (`_EXTRA_HANDLERS`, `_execute_job_handler`). For something intended as a public extension point (per D5), export it from `__init__` and document its stability guarantee.

**10. The collections endpoint issues N+1 full scans, each with its own 15s budget — multiplying finding 1.** `src/xagent/web/api/kb.py:4767-4789`: `GET /api/kb/collections` performs one full `list_collections` for the personal scope plus one more per distinct team-KB owner (`for storage_user_id, refs in refs_by_owner.items(): owner_result = await asyncio.wait_for(list_collections(...), timeout=15)`), serially, each with an independent 15s deadline. Not introduced here — the N+1 structure predates this PR — but it directly multiplies finding 1's blast radius: a user belonging to N team-KB owners gets N+1 independent chances per request to hit the newly-reachable timeout and leave an orphaned scan thread behind. Suggestion: consider collapsing these into a single admin-scoped scan.

## Simplification

`src/xagent/web/jobs/tasks.py:26-38`: **yagni** — speculative `register_background_job_handler()`/`_EXTRA_HANDLERS` registry with no in-repo consumer and no mechanism to import a downstream module into the worker. Either delete it and add a real branch to `_execute_job_handler()` when an actual second job type appears, or land it as a config-driven module-import mechanism (mirroring `XAGENT_EXTERNAL_SKILLS_LIBRARY_DIRS`) in the same PR that adds a real caller. *net: -13 lines possible*

## Review limitations

- The backend pytest suite and frontend vitest were not run locally (worktree Python/Node setup issue unrelated to this PR). GitHub Actions CI for this PR is fully green across all jobs (pytest, frontend-build, pre-commit, e2e, migrations), and every pytest/vitest file cited above was read in full with its logic traced or partially reproduced.
- **Open question for the author:** LanceDB thread-safety of the shared `DBConnection`/`Table` under concurrent `to_thread` calls. Multiple worker threads can now hit the single shared `self._conn`, and the sync path has no internal locking (unlike the async path's `self._async_lock`). This could not be settled from the repo alone — it is plausible the underlying Rust connection is thread-safe. Please confirm against upstream LanceDB documentation.
- The performance claims in the PR description (5.28s → 4.02s, 12 → 24 probes) were not independently reproduced; doing so requires a real 400k-row table plus a single-worker server.

src/xagent/core/tools/core/RAG_tools/management/collections.py
rogercloud: Fixed in 81621c52 — `kb_collections_timeout_seconds` is now configurable via `XAGENT_KB_COLLECTIONS_TIMEOUT_SECONDS` (default raised to 30s), and the per-owner scans in 9311c87a/81621c52 now run concurrently via `asyncio.gather` instead of serially, which also reduces the number of sequential timeout opportunities per request.

src/xagent/web/jobs/tasks.py
rogercloud: Fixed in b2979069 — unknown job types now raise `BackgroundJobHandlerError(..., retryable=False)` instead of a plain `ValueError`, so they fail fast instead of burning through the full retry budget. Covered by `test_unknown_job_type_fails_fast_without_retry`.

src/xagent/web/jobs/tasks.py
rogercloud: Fixed in b2979069 — `register_background_job_handler` now takes a `replace: bool = False` keyword-only param and rejects duplicate registrations and built-in-type collisions with a `ValueError` (mirroring `register_trigger_provider`). Covered by new tests in `test_background_jobs.py`.

src/xagent/web/jobs/tasks.py
rogercloud: Correcting my own earlier read here: I dug into this further and this premise is wrong — the registry is not an unreached extension point. A downstream repo (xagent-saas) has an open PR that imports `register_background_job_handler` and registers a `kb.team.transfer` handler at import time, and solves the worker-import problem by pointing its own Celery `-A` entrypoint at a module that imports the handler module for its side effect — no config-driven mechanism needed on this side. So no action needed here; withdrawing this finding. (I did keep the separate, still-valid hardening from the sibling comment on line 38 — duplicate-registration and built-in-collision checks — since xagent-saas's own test suite was reaching into the private `_EXTRA_HANDLERS` dict directly, which b2979069 also now exposes a public accessor for.)

frontend/src/components/pages/knowledge-base.tsx
rogercloud: Fixed in 6c6df76a — the 202 branch now throws when the body doesn't satisfy `isBackgroundJobResponse` instead of silently falling through to the success path. Covered by new tests in `knowledge-base.test.tsx`.

Also correcting the aside in my original comment: I claimed the `/api/knowledge-bases/{name}/promote-team` route doesn't exist anywhere and this code is unreachable — that's wrong. It's a real route implemented in a downstream repo (xagent-saas) that ships this exact OSS frontend to its users; the OSS-frontend-calls-SaaS-backend split is an established, intentional pattern here (predating this PR, from OSS PR #904), not a gap.

frontend/src/components/build/agent-builder.tsx
rogercloud: Fixed in 6c6df76a — same fix as the sibling comment on `knowledge-base.tsx`: throws on an unrecognized 202 body instead of falling through. Added test coverage in the new `agent-builder-kb-share.test.tsx`.

tests/core/tools/core/RAG_tools/kb/test_list_collections_event_loop.py
rogercloud: Fixed in 9311c87a — the heartbeat assertion now scores each dispatched call over its own window (plus a thread-identity check) instead of one cumulative total, so a regression in any single `to_thread` call is caught individually. Verified via mutation testing (removing each dispatch one at a time) that all are caught.

tests/web/test_background_jobs.py
rogercloud: Fixed in 0fdd43ab — added `test_registered_handler_retryable_error_flows_through_execute_background_job`, which drives a registered handler raising `BackgroundJobHandlerError(retryable=True)` through the real `execute_background_job` task and asserts the retry path (job reset to `ENQUEUED`, `self.retry` invoked) is exercised, not just the private dispatch function.

rogercloud (approved):

Issue #1004, opened 2026-07-25
test(rag): real_rag suite has never executed and cannot fail as written


rogercloud:
## Summary

The `real_rag` test suite (34 tests, 5 files) has never executed a single test in CI history. Fixing the obvious cause alone would not make it useful, because the tests as written cannot fail in the ways they appear to check.

## Why it never runs

1. `DASHSCOPE_API_KEY` is not configured at repository or organization level. `.github/workflows/ci-real-rag.yml` self-skips via its "Check real embedding provider config" step. Verified on both `pull_request` and `push: main` runs — the `push: main` case, where secrets would be available, also skipped every step from `Checkout code` onward and completed in 9s.
2. **Second, independent gate:** `tests/web_integration/test_real_url_ingestion_e2e.py:23-28` also carries `requires_network`, which `tests/conftest.py` gates behind `--run-special` or `XAGENT_TESTS_ALLOW_NETWORK=1`. Neither is set in `ci-real-rag.yml`. So even after configuring the API key, 11 of the 34 tests would continue to skip.

## Most of the coverage is already duplicated

`tests/web_integration/conftest.py:128-146` installs an autouse `_StubEmbeddingAdapter` for every e2e test *except* those marked `real_rag`. The equivalent `contract_stub` suites run on every PR and already cover:

- Cross-tenant 403 authorization (9 tests) — duplicated by `tests/web_integration/test_kb_access_control_e2e.py`, whose own docstring at line 9 documents the intended split
- Schema/migration CRUD field stability (12 tests) — structurally duplicated
- ingest to search HTTP contract — duplicated

Roughly 21 of the 34 tests add no coverage beyond the mocked suites.

## The genuinely unique coverage is not actually asserted

Only two things are unique to this suite: real DashScope API contract drift, and real embedding retrieval quality. Neither is effectively tested:

- Most relevance assertions are wrapped in `if results:`. A real embedding path returning zero results would fail nothing.
- The strongest semantic assertion is substring presence (`token_phrase in joined`); ordering and ranking are unchecked.
- No test anywhere — mocked or real — asserts the embedding dimension against the configured model, or against the LanceDB schema vector width.
- `tests/core/model/embedding/test_dashscope.py` and the rerank equivalent hand-mock the response JSON, so they validate the client's parser against a frozen contract, not against the live API.
- `.github/workflows/ci-real-rag.yml` sets no `timeout-minutes`, so it falls back to the platform default.

## Recommended order

[...]

PR #1005, opened 2026-07-25
ci: prepare main for a merge queue and harden the alembic head check

rogercloud:
Groundwork for replacing the global serial-merge rule on `main` with a merge
queue. Nothing here changes branch protection or enables a queue -- every change
is inert until those settings move, and can be reviewed and landed on its own.

## Why

`main` has `required_status_checks.strict: true` ("require branches to be up to
date"). That setting is a per-branch boolean, so it cannot be scoped to the PRs
that actually need it, and the practical effect is that every PR merges one at a
time: each merge makes every other open PR stale and someone has to re-sync
`main` by hand.

The rule exists for a real reason. Two migrations branched off the same Alembic
revision each declare the same `down_revision`; individually valid, merged
together they leave `main` with two heads. The repository carries 13 hand-written
`*_merge_*.py` revisions from cleaning that up. A merge queue reaches the same
guarantee -- it validates the exact tree that will land, at the moment it lands --
while letting unrelated PRs merge freely.

## What is in here

**`fix(migrations)`** -- `scripts/check_alembic_heads.sh` detected all three
defects it cares about but described two of them wrongly: a duplicate revision ID
was reported as "found 2 heads", and an unresolvable `down_revision` as "found 0
heads". Both send the reader looking for a branch that does not exist. Now each
case says which one it is.

Nothing proved this hook could fail before now. `tests/migrations/test_check_alembic_heads.py`
injects each defect into a throwaway Alembic environment and asserts rejection,
plus a healthy control so a failure is attributable to the defect rather than to
the fixture harness. Verified by crippling the script to always exit 0 and
confirming all three rejection tests fail while the control still passes.

**`ci`** -- `merge_group:` triggers on `test-migrations.yml` (owns the two
currently required contexts) and `ci.yml` (whose `CI Summary` is about to become
required). A required check that does not report on a merge group times out and
stalls every merge, so this has to be in place before a queue exists. Both are
no-ops until then.

`test-migrations.yml` also needs to recognise the event when deciding whether to
run -- left alone it fell through to the non-`pull_request` branch and ran the
full 5-6 minute suite on every merge, including for changes touching no
migrations. Verified across all six event/diff combinations against a real
migration commit.

**`docs`** -- `docs/branch-protection.md` records the configuration, the
reasoning, and the break-glass procedure, including why it uses the dedicated
`enforce_admins` endpoint rather than a full protection `PUT` (which would
silently drop the empty-allowlist push restriction).

The `.gitignore` change is a fix in its own right: `*.md` has no leading slash,
so it matched at every depth and silently ignored any doc added in a
subdirectory unless it happened to be named `README.md`. The comment above it
already said "Root directory markdown files"; now it is scoped to that. Verified
that no other file in the tree changes tracking status.

## Deliberately not here

* No branch protection or ruleset changes.
* No new implementation of the Alembic graph check -- the existing hook is the
  single implementation, and this only fixes its diagnostics and covers it.
* No `merge_group:` on `ci-real-rag.yml`. That suite has never executed once, and
  `merge_group` runs with secret access, so adding it would let a never-run suite
  start gating merges. See #1004.

## Checks

`ruff check`, `ruff format`, and `alembic-check` all pass. `mypy` is unaffected
(no `src/xagent/**` changes) and the npm hooks do not trigger (no JS/TS/JSON
changes). `tests/migrations` is already in the `pytest-fast-deepdoc` core matrix,
so the new tests run in CI without further wiring.

### DIFF
--- a/docs/branch-protection.md
+++ b/docs/branch-protection.md
@@ -1,0 +1,113 @@
+# Branch protection and the merge queue
+
+How `main` is protected, why it is set up this way, and what to do when it
+blocks something urgent.
+
+## What is being protected against
+
+Two migrations branched off the same Alembic revision will each declare the same
+`down_revision`. Individually both are valid; merged together they leave `main`
+with two heads, and `alembic upgrade head` can no longer resolve a single
+target. This has happened repeatedly -- the repository carries 13 hand-written
+`*_merge_*.py` revisions and a trail of `fix(migrations): merge alembic heads`
+commits -- and each occurrence is a fire drill that blocks everyone.
+
+Alembic does not prevent any of this. A branched graph is legal to Alembic, and
+`ScriptDirectory.get_heads()` returns every head with no warning and no error.
+The single-head invariant is a policy this repository enforces, not something
+the tool gives us.
+
+## How it is enforced
+
+| Layer | What it catches | Where |
+| --- | --- | --- |
+| `alembic-check` pre-commit hook | multiple heads, duplicate revision IDs, unresolvable `down_revision` | locally on commit, and on every PR via the `pre-commit` job |
+| Merge queue | the same defects, evaluated against the tree that will actually land | at merge time |
+| `Test SQLite/PostgreSQL Migrations` | a real `alembic upgrade head` on both backends | required status check |
+| `ci.yml` on `push: main` | post-merge detection if something slipped through | red build on `main`, emailed to the pusher |
+
+`scripts/check_alembic_heads.sh` is the single implementation of the graph
+check. Its regression coverage lives in
+`tests/migrations/test_check_alembic_heads.py`, which injects each defect into a
+throwaway Alembic environment and asserts the script rejects it. Those negative
+tests exist because a check that cannot fail is worse than no check: it reads as
+protection that is not there.
+
+## Why a merge queue instead of "require branches to be up to date"
+
+`required_status_checks.strict` is a per-branch boolean -- it cannot be scoped to
+PRs that touch migrations. Turning it on therefore serialized *every* PR: after
+each merge, every other open PR went stale and had to re-sync `main` by hand.
+
+The merge queue reaches the same guarantee without that cost. For each entry it
+builds `gh-readonly-queue/main/pr-N-<sha>` containing `main` plus the entries
+ahead of it plus that PR, and runs the required checks on that branch. Two
+conflicting migrations produce two heads there, so the second entry fails and is
+ejected while the first merges. This is evaluated at merge time against the exact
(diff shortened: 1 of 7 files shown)


docs/branch-protection.md
```diff
@@ -0,0 +1,123 @@
+# Branch protection and the merge queue
+
+How `main` is protected, why it is set up this way, and what to do when it
+blocks something urgent.
+
+> **Status.** Live today: the `alembic-check` hook, the required
+> `Test SQLite/PostgreSQL Migrations` checks, and the `merge_group:` triggers
+> (inert until a queue exists). Still pending: `CI Summary` as a required check,
```
qinxuye: [P1] Include disabling strict status checks in the queue rollout

The live `main` protection still has `required_status_checks.strict=true`, but this status block does not list turning it off. GitHub treats the strict up-to-date requirement separately from requiring a merge queue, and a PR is only admitted after its existing branch-protection requirements are satisfied. If the queue is enabled while strict remains true, every PR still has to sync `main` and rerun checks before it can enter the queue, preserving the serialization this change is intended to remove. Please make `strict=false` an explicit pending activation step and record it in the live/pending status.

qinxuye (commented):
One architecture-level rollout gap remains on the latest head.

qinxuye (commented):
One rollout-order issue remains after the latest fix.

docs/branch-protection.md
```diff
@@ -0,0 +1,139 @@
...
+blocks something urgent.
+
+> **Status.** Live today: the `alembic-check` hook, the required
+> `Test SQLite/PostgreSQL Migrations` checks, and the `merge_group:` triggers
+> (inert until a queue exists). Still pending, in this order:
+>
+> 1. `CI Summary` as a required status check -- independent of the queue.
+> 2. Exclude `refs/heads/gh-readonly-queue/**` from the branch-creation ruleset,
+>    or the queue cannot build anything.
+> 3. **Set `required_status_checks.strict` to `false`.** It is still `true` on
```
qinxuye: [P1] Enable the merge queue before relaxing strict checks

This ordered rollout disables `required_status_checks.strict` before the merge queue is active. Between steps 3 and 4, `main` would therefore accept an approved stale PR whose old checks are green without validating the combined tree, reopening the migration race this queue is intended to prevent; if queue activation fails, the repository remains in that weaker state. Please keep strict enabled while enabling and verifying the merge queue, then set `strict=false` only after the queue is confirmed active (or make the two changes atomically if the configuration surface supports it).

qinxuye (approved):
LGTM

PR #1008, opened 2026-07-26
docs: correct branch protection to match the shipped merge queue

rogercloud:
The merge queue is now live on `main` and `required_status_checks.strict` is off.
`docs/branch-protection.md` landed in #1005 describing both as pending, so it is
now wrong about the current configuration — and one of the things it is wrong
about is a security control.

## The urgent part

The page states:

> Pushes to `main` are restricted with an empty allowlist, so nobody can push
> directly.

That restriction was removed during rollout. A document that claims a protection
which no longer exists is worse than no document, so this corrects it.

## Why the restriction had to go

**A merge queue and "Restrict who can push to matching branches" cannot both be
enabled.** The queue merges by pushing to `main` as `github-merge-queue[bot]`,
which is not a repository admin, so a push restriction rejects it. The entry is
then ejected with `branch_protection_failure` — *after* every required check has
gone green, which makes it look like an unrelated fault.

The bot cannot be added to the allowlist: `github-merge-queue` is an internal
GitHub service, not an installable App, so it is not selectable as a user, team,
or app. GitHub's own position on this is that using a merge queue means granting
push access, and that the protection you keep is the ban on force pushes.

Removing the restriction also silently disabled `block_creations`, which depends
on it — the API reports `false` even when a request sets it `true`. Low impact
here (it only blocked creating a branch named `main`, which cannot happen while
`main` exists and `allow_deletions` is false), but it is recorded rather than
glossed over.

What still holds `main`: no force pushes, no deletions, both migration checks
required, one approval plus resolved conversations, and the queue validating the
combined tree before it lands. The exposure is that a write collaborator can push
an ordinary commit straight to `main`, bypassing review; nothing can rewrite or
delete history.

## Gotchas recorded

Each of these presents as a silent failure, and each cost real time to find:

* **`refs/heads/gh-readonly-queue/**` does not match the queue's refs.** The
  exclusion needs a trailing `/*`. GitHub's rules API reports zero applicable
  rules for the short spelling, so it looks like a fix while leaving every merge
  stalled. `~ALL` does match those refs, and the `creation` rule does reject them
  (HTTP 422) — the blocker is real.
* **A ruleset whose include and exclude are identical is rejected** and the
  previous value stays in place, with no error surfaced by `--jq` on the happy
  path. Check the response.
* **Removal reasons exist only in GraphQL.** The REST timeline event has no
  reason field; `RemovedFromMergeQueueEvent.reason` does. Without it the
  ejections are indistinguishable from a timeout.
* **Merge queue is organization-only.** A `merge_queue` ruleset on a personal
  repository is rejected with `Invalid rule 'merge_queue'` regardless of any
  other setting, so a personal fork cannot be used to rehearse queue changes.

## Verification

Every configuration value quoted here was checked against the live settings
rather than written from memory.

The conflict-detection chain was verified by reconstructing the exact tree the
queue builds: two branches each adding a migration off the same head are
individually valid and merge with no git conflict, but the combined tree fails
`alembic-check` ("expected exactly 1 head, found 2") and `alembic upgrade head`
(`CommandError: Multiple head revisions are present`), and
`detect-migration-changes` reports `should-test=true` for it. The one link not
directly exercised is the queue ejecting on that failure; it is documented
behaviour and the queue was observed ejecting entries three times during
rollout.

Docs only — no configuration changes in this PR.

### DIFF
--- a/docs/branch-protection.md
+++ b/docs/branch-protection.md
@@ -3,33 +3,16 @@
 How `main` is protected, why it is set up this way, and what to do when it
 blocks something urgent.
 
-> **Status.** Live today: the `alembic-check` hook, the required
-> `Test SQLite/PostgreSQL Migrations` checks, and the `merge_group:` triggers
-> (inert until a queue exists). Still pending, in this order:
+> **Status: the merge queue is live.** `required_status_checks.strict` is off, so
+> PRs no longer have to sync `main` by hand. Merging is now "Merge when ready",
+> which enqueues the PR; the queue validates it against the latest `main` and
+> merges it for you.
 >
-> 1. `CI Summary` as a required status check -- independent of the queue.
-> 2. Exclude `refs/heads/gh-readonly-queue/**` from the branch-creation ruleset,
->    or the queue cannot build anything.
-> 3. Enable the merge queue, **leaving `required_status_checks.strict` on**.
-> 4. Verify the queue end to end: a merge-group branch is created, and every
->    required context reports on it. An already-up-to-date PR satisfies strict, so
->    it can be enqueued for this without relaxing anything first.
-> 5. **Only now set `required_status_checks.strict` to `false`.** It is still
->    `true` on `main`, and it has to go: a PR must satisfy its branch-protection
->    requirements before it is admitted to the queue, so leaving strict on
->    permanently would keep every PR syncing `main` and re-running checks first --
->    exactly the serialization the queue exists to remove.
-> 6. `enforce_admins` -- deliberately last, once the queue is proven.
+> **The push restriction on `main` was removed to make this possible**, and
+> `block_creations` went with it. See "The cost: no push restriction" below --
+> that trade-off is the most important thing on this page.
 >
-> **The order of 3 and 5 matters.** Turning strict off first would leave a window
-> where neither protection applies: `main` would accept an approved but stale PR
-> on the strength of green checks that never saw the combined tree, which is the
-> migration race this whole change exists to prevent. If queue activation then
-> failed, the repository would be stuck in that weaker state. Keeping strict on
-> until the queue is confirmed working means the worst case is the status quo.
->
-> Sections below describe the mechanism; anything above that is still pending is
-> not in force yet. Update this block as each step lands.
+> Still pending: `CI Summary` as a required status check, and `enforce_admins`.
 
 ## What is being protected against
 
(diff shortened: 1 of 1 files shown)


docs/branch-protection.md
```diff
@@ -80,50 +73,107 @@ the same column produce a valid single-head graph that still fails on upgrade.
...
+
+| | |
+| --- | --- |
+| `allow_force_pushes: false` | history cannot be rewritten |
+| `allow_deletions: false` | `main` cannot be deleted |
+| required status checks | `Test SQLite/PostgreSQL Migrations` |
+| required reviews | 1 approval, conversations resolved |
+| merge queue | validates the combined tree before it lands |
+
+So the exposure is that a collaborator with write access can push an ordinary
```
qinxuye: [P1] Required PR reviews still prevent collaborator direct pushes

The live rule still has `required_pull_request_reviews` enabled with one approval. GitHub applies that requirement to branch updates: ordinary write collaborators can only update `main` through an approved PR; removing `restrictions` removes the additional actor allowlist, not the required-PR gate. Only administrators can currently bypass it because `enforce_admins=false`, and they could already do that before this rollout. `strict` controls whether the PR must be up to date, not whether a PR is required. Please rewrite this exposure and fallback accordingly: the queue does not need to be removed to prevent collaborator direct pushes, and enabling `enforce_admins` closes the remaining administrator bypass.

qinxuye (commented):
One security-semantics issue remains in the updated branch-protection documentation.

qinxuye (approved):
LGTM

docs/branch-protection.md
rogercloud: Correct, and the section was wrong in a second way you did not have the data to
see. Fixed in bd79db67.

You are right that `required_pull_request_reviews` is the gate that blocks
direct pushes and that `restrictions` is an actor allowlist layered on top.
Checked against `main`'s history: all 50 most recent commits have an associated
pull request, with no bare pushes.

The extra detail is what the removed `restrictions` actually contained -- an
**empty** allowlist (`users: []`, `teams: []`, `apps: []`). With
`enforce_admins: false` that meant only administrators could merge, which is why
every pre-rollout merge on the branch was performed by an admin. So the real
change is:

| | before | after |
| --- | --- | --- |
| direct push, write collaborator | rejected | rejected |
| direct push, administrator | allowed | allowed (unchanged) |
| merging an approved PR, write collaborator | rejected | allowed |
| merging an approved PR, administrator | allowed | allowed |

Merge permission widened from administrators to all write collaborators, still
via an approved PR. That is the intended effect of the rollout, not a security
regression -- the page had it as the latter, describing an exposure that does
not exist.

Also corrected the fallback: `strict` is not the lever. Narrowing merge
permission again means restoring `restrictions`, which costs the queue, and the
remaining administrator bypass closes with `enforce_admins` without touching the
queue.

PR #1011, opened 2026-07-26
chore: remove the migration PR refresh automation

rogercloud:
`refresh-migration-prs.yml` and `.github/scripts/refresh_migration_prs.py`
existed to work around `required_status_checks.strict: true`. That is now off
and `main` uses a merge queue, so everything they do is either obsolete or
already covered -- but they are still armed, and the next migration that lands
on `main` will set them off.

## What they currently do, and who covers it now

The workflow fires on `push: main` under `src/xagent/migrations/**`, holds
`contents: write` and `pull-requests: write`, and for every open same-repo PR
touching migrations it:

| | now covered by |
| --- | --- |
| `PUT /pulls/{n}/update-branch` -- merges `main` into the contributor's branch | nothing needed; `strict` is off and the queue builds the combined tree itself |
| polls until the refreshed head appears | same |
| re-derives the Alembic head graph by `ast`-parsing revision files | `scripts/check_alembic_heads.sh` (pre-commit and the `pre-commit` CI job), the two migration jobs running a real `alembic upgrade head`, and the queue evaluating both against the tree that will actually land |
| posts a `migration-refresh/alembic-heads` commit status | nothing -- see below |
| dispatches `ci.yml` and `test-migrations.yml` on the PR branch | those already run on `pull_request` and on `merge_group` |

## Why removing it is a net improvement, not just cleanup

**It writes to other people's branches.** `update-branch` puts a merge commit
into the contributor's branch without asking, then dispatches two workflows
against it. Under `strict` that was the price of keeping PRs mergeable. Now it
is unsolicited churn: it burns CI on PRs nobody asked to rebuild, and it moves a
branch out from under whoever is working on it.

**It is a fourth, divergent implementation of the single-head check.** The other
three go through Alembic. This one hand-parses `down_revision` out of the source
with `ast` and reimplements the graph walk, so it can disagree with the
authoritative check while looking equally official on the PR. A check that can
drift from the thing it is standing in for is worse than no check.

**Its status is decorative.** `migration-refresh/alembic-heads` is not among the
required contexts on `main` (those are `Test SQLite Migrations` and
`Test PostgreSQL Migrations`), and no open PR currently carries it. It cannot
block anything; it can only look like it might.

## Verification

* `grep` across the repository for `refresh_migration_prs`,
  `refresh-migration-prs` and `migration-refresh` finds no reference outside the
  two deleted files.
* No test references the script.
* `.github/scripts/` held only this file, so the directory goes with it.
* Confirmed against the live branch protection that
  `migration-refresh/alembic-heads` is not a required context, and against three
  open PRs that none of them carry that status.

The workflow has not run since 2026-07-24 only because no migration has landed
on `main` since. This is not dead code that is safely inert -- it is live code
waiting for a trigger.

## Not included

Making `Migrations Summary` and `CI Summary` the required contexts, and the
resulting update to `docs/branch-protection.md`, are separate steps. The doc is
deliberately left for last so it can be written from the settled live
configuration rather than from intent.


qinxuye (approved):
LGTM

PR #1013, opened 2026-07-26
docs: record the summary contexts and why job names cannot be required

rogercloud:
Two things changed on `main` after #1010 and #1011 landed, and the page does not
reflect either:

* the required contexts are now `Migrations Summary` and `CI Summary`, not the
  individual migration job names
* `.github/scripts/refresh_migration_prs.py` is gone

The page still says `CI Summary` is pending, which is the part that matters:
that line reads as "red tests can still merge", and that is no longer true.

## Why the contexts changed, recorded as its own section

This is the piece most worth keeping, because it is not obvious and it fails
silently.

**GitHub reports a skipped job to branch protection as a success.** Successful
statuses are `success`, `skipped` and `neutral`. GitHub's "Troubleshooting
required status checks" documents both routes: a job skipped by a conditional
"reports 'Success'", and a job skipped because something in its `needs` failed
"may not block merging", with `always()` given as the fix.

So requiring an individual job name is not fail-closed. If the job the migration
tests depend on fails, they are skipped, their contexts go green, and the merge
lands with nothing having run. The new section states this, names the two
summary jobs and what each one gathers, and flags that `always()` is
load-bearing and that adding a job to a workflow does not gate on it until it is
added to the summary's `needs` and `check_job` list.

It also records why `CI Summary` had to become required: before it did, three
open PRs with red CI were mergeable.

## Also recorded: the removed second head-check implementation

The page claimed `scripts/check_alembic_heads.sh` was "the single implementation
of the graph check" while `refresh_migration_prs.py` was still in the tree
hand-parsing `down_revision` with `ast` and re-deriving the head graph without
Alembic. That claim is true now, and the paragraph says what made it true --
worth keeping, because a second implementation that can disagree with the
authoritative one while posting an equally official-looking status is the
failure mode, not the duplication itself.

## Verification

Every value on the page was re-read from the live configuration rather than
written from intent, which is the specific mistake that produced #1008:

* branch protection: `strict: false`, contexts `Migrations Summary` +
  `CI Summary`, no `restrictions`, `enforce_admins: false`,
  `block_creations: false`, no force pushes, no deletions, 1 approval,
  conversation resolution required
* merge queue ruleset: all seven parameters match what the page quotes
* the two summary jobs: for each, every entry in `needs` has a matching
  `check_job` call -- no job is gathered but left unchecked

Docs only -- no configuration changes in this PR.

### DIFF
--- a/docs/branch-protection.md
+++ b/docs/branch-protection.md
@@ -199,6 +199,21 @@ the summary. **`always()` is load-bearing.** Removing it looks harmless and
 silently restores the hole, which is why it carries a comment in the workflow
 saying so.
 
+### Gate at the step, not at the job
+
+Failing on `skipped` only works because **no gathered job is ever skipped**.
+`Test SQLite Migrations` and `Test PostgreSQL Migrations` carry no job-level
+`if:` at all. When `detect-migration-changes` reports `should-test=false`, the
+jobs still run; it is their *steps* that are gated, and the job reports
+`success` in a few seconds. A PR that touches no migration still gets a green
+`Migrations Summary` -- this page's own PR is an example.
+
+That is a deliberate property and an easy one to destroy. Moving the condition
+up to the job level looks like an optimisation -- same outcome, one less job
+started -- but it turns every non-migration PR into a skipped required job,
+which reports success, which is the exact hole the summary exists to close.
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #1018, opened 2026-07-26
fix: stage uploaded files off the event loop

AlexLiu190625:
## Summary

- move local upload path reservation, source reads, size enforcement, and destination writes into one synchronous worker
- reserve same-name files atomically with exclusive creation
- make the request-level failure path the sole owner of completed staging cleanup

## Root cause

The shared upload service selected a unique filename with a check-then-open sequence and performed local destination writes from the async request flow. Concurrent same-name uploads could choose the same path. Cancellation also split cleanup between the staging helper and the outer request failure path, so the same late result could be unlinked twice.

## What changed

- keep path construction, exclusive `xb` reservation, bounded reads, and same-descriptor writes in one worker
- treat only an exclusive-open `FileExistsError` as a filename collision
- keep partial-copy cleanup inside the worker that owns the reserved descriptor
- drain a cancelled staging worker, publish its exact late path to `written_paths`, and let the outer `finally` clean it once
- reuse `_delete_staged_upload` for request-level cleanup so operational unlink handling has one implementation
- remove the unused copied-size return value while retaining maximum-size enforcement
- add real `get_upload_path`, exact cleanup-count, exception-narrowing, and non-broad `open` mock coverage

## Compatibility and scope

- no route, response schema, database schema, durable-registration, preview, or storage-CAS changes
- maximum file-size behavior and existing upload authorization remain unchanged
- collision retry policy and executor sizing remain unchanged; those require separate operational evidence and configuration ownership

## Verification

- `uv run pytest tests/web/api/test_upload_connection_boundary.py tests/web/api/test_kb_web_ingestion_uploaded_file.py -q`
- `git diff --check` passed

This is follow-up boundary hardening related to #935; it does not close the task-runtime acceptance criteria in that issue.

### DIFF
--- a/src/xagent/web/api/files.py
+++ b/src/xagent/web/api/files.py
@@ -301,7 +301,7 @@ def _delete_staged_upload(path: Path) -> None:
 
     try:
         path.unlink(missing_ok=True)
-    except BaseException:
+    except OSError:
         logger.warning("Failed to clean up local upload: %s", path)
 
 
--- a/src/xagent/web/api/files.py
+++ b/src/xagent/web/api/files.py
@@ -311,7 +311,7 @@ def _reserve_and_copy_upload(
     task_id: str | None,
     folder: str | None,
     user_id: int,
-) -> tuple[Path, int]:
+) -> Path:
     """Reserve one local upload path and copy bytes through its descriptor.
 
     This synchronous worker owns path construction, exclusive reservation, and
@@ -357,7 +357,7 @@ def _reserve_and_copy_upload(
         except BaseException:
             _delete_staged_upload(candidate)
             raise
-        return candidate, total_size
+        return candidate
 
 
 async def _stage_uploaded_file(
@@ -367,8 +367,8 @@ async def _stage_uploaded_file(
     folder: str | None,
     user_id: int,
     written_paths: list[Path],
-) -> tuple[Path, int]:
-    """Run local staging off-loop and drain exact cleanup on cancellation."""
+) -> Path:
+    """Run local staging off-loop and publish its path to request cleanup."""
 
     worker = asyncio.create_task(
         asyncio.to_thread(
@@ -379,17 +379,13 @@ async def _stage_uploaded_file(
             user_id=user_id,
         )
     )
-    (target_path, file_size), cancellation = await await_task_settlement(worker)
+    target_path, cancellation = await await_task_settlement(worker)
     # The request cleanup owner knows this exact path before any subsequent
     # preview, registration, or cancellation-cleanup await point.
     written_paths.append(target_path)
     if cancellation is not None:
-        cleanup_worker = asyncio.create_task(
-            asyncio.to_thread(_delete_staged_upload, target_path)
-        )
-        await drain_async_task_cancellation_safe(cleanup_worker)
         raise cancellation
-    return target_path, file_size
+    return target_path
 
 
 async def store_uploaded_files(
@@ -428,7 +424,7 @@ async def store_uploaded_files(
                 )
 
             try:
-                target_path, _file_size = await _stage_uploaded_file(
+                target_path = await _stage_uploaded_file(
                     uploaded,
                     task_id=task_id,
                     folder=folder,
@@ -526,13 +522,7 @@ async def store_uploaded_files(
 
                     def _delete_local_paths() -> None:
                         for path in written_paths:
-                            try:
-                                path.unlink(missing_ok=True)
-                            except OSError:
-                                logger.warning(
-                                    "Failed to clean up local upload: %s",
-                                    path,
-                                )
+                            _delete_staged_upload(path)
 
                     cleanup_worker = asyncio.create_task(
                         asyncio.to_thread(_delete_local_paths)
--- a/src/xagent/web/api/files.py
+++ b/src/xagent/web/api/files.py
@@ -43,6 +43,7 @@ from ..models.task import Task
 from ..models.uploaded_file import UploadedFile
 from ..models.user import User
 from ..services.db_runtime import (
+    await_task_settlement,
     drain_async_task_cancellation_safe,
     run_db_io_cancellation_safe,
 )
@@ -288,35 +289,107 @@ def _accel_redirect_response(
     )
 
 
-async def _write_upload_with_size_limit(uploaded: UploadFile, target_path: Path) -> int:
-    """Persist an uploaded file while enforcing the configured size limit."""
-    total_size = 0
-    read_buffer_size = 1024 * 1024  # 1MB chunks keep memory bounded for large files.
+_UPLOAD_COPY_BUFFER_SIZE = 1024 * 1024
+
+
+class _InvalidUploadPathError(ValueError):
+    """Path-construction validation failure for a request upload."""
+
+
+def _delete_staged_upload(path: Path) -> None:
+    """Best-effort removal which never replaces the staging failure."""
 
     try:
-        with open(target_path, "wb") as buffer:
-            while True:
-                chunk = await uploaded.read(read_buffer_size)
-                if not chunk:
-                    break
-                total_size += len(chunk)
-                if total_size > MAX_FILE_SIZE:
-                    raise HTTPException(
-                        status_code=413,
-                        detail=(
-                            f"File size exceeds maximum limit of {MAX_FILE_SIZE_LABEL}"
-                        ),
-                    )
-                buffer.write(chunk)
+        path.unlink(missing_ok=True)
     except BaseException:
-        try:
-            if target_path.exists():
-                target_path.unlink()
-        except OSError:
-            pass
-        raise
+        logger.warning("Failed to clean up local upload: %s", path)
 
-    return total_size
+
+def _reserve_and_copy_upload(
+    uploaded: UploadFile,
+    *,
+    task_id: str | None,
+    folder: str | None,
+    user_id: int,
+) -> tuple[Path, int]:
+    """Reserve one local upload path and copy bytes through its descriptor.
+
+    This synchronous worker owns path construction, exclusive reservation, and
+    bounded source reads. A collision is only a ``FileExistsError`` raised by
+    the exclusive create itself; errors from source reads or destination writes
+    remain I/O failures and clean the exact reserved candidate.
+    """
+
+    try:
+        path = get_upload_path(uploaded.filename or "", task_id, folder, user_id)
+    except ValueError as exc:
+        raise _InvalidUploadPathError(str(exc)) from exc
+    stem = path.stem
+    suffix = path.suffix
+    candidate = path
+    suffix_index = 1
+
+    while True:
+        try:
+            destination = open(candidate, "xb")
+        except FileExistsError:
+            candidate = path.parent / f"{stem}_{suffix_index}{suffix}"
+            suffix_index += 1
+            continue
+
+        total_size = 0
+        try:
+            with destination:
+                while True:
+                    chunk = uploaded.file.read(_UPLOAD_COPY_BUFFER_SIZE)
+                    if not chunk:
+                        break
+                    total_size += len(chunk)
+                    if total_size > MAX_FILE_SIZE:
+                        raise HTTPException(
+                            status_code=413,
+                            detail=(
+                                f"File size exceeds maximum limit of "
+                                f"{MAX_FILE_SIZE_LABEL}"
+                            ),
+                        )
+                    destination.write(chunk)
+        except BaseException:
+            _delete_staged_upload(candidate)
+            raise
+        return candidate, total_size
+
+
+async def _stage_uploaded_file(
+    uploaded: UploadFile,
+    *,
+    task_id: str | None,
+    folder: str | None,
+    user_id: int,
+    written_paths: list[Path],
+) -> tuple[Path, int]:
+    """Run local staging off-loop and drain exact cleanup on cancellation."""
+
+    worker = asyncio.create_task(
+        asyncio.to_thread(
+            _reserve_and_copy_upload,
+            uploaded,
+            task_id=task_id,
+            folder=folder,
+            user_id=user_id,
+        )
+    )
+    (target_path, file_size), cancellation = await await_task_settlement(worker)
+    # The request cleanup owner knows this exact path before any subsequent
+    # preview, registration, or cancellation-cleanup await point.
+    written_paths.append(target_path)
+    if cancellation is not None:
+        cleanup_worker = asyncio.create_task(
+            asyncio.to_thread(_delete_staged_upload, target_path)
+        )
+        await drain_async_task_cancellation_safe(cleanup_worker)
+        raise cancellation
+    return target_path, file_size
 
 
 async def store_uploaded_files(
@@ -355,18 +428,20 @@ async def store_uploaded_files(
                 )
 
             try:
-                target_path = _build_unique_file_path(
-                    get_upload_path(uploaded.filename, task_id, folder, user_id)
+                target_path, _file_size = await _stage_uploaded_file(
+                    uploaded,
+                    task_id=task_id,
+                    folder=folder,
+                    user_id=user_id,
+                    written_paths=written_paths,
                 )
-            except ValueError as e:
+            except _InvalidUploadPathError as e:
                 logger.warning(f"Invalid folder name rejected: {folder!r} - {e}")
                 raise HTTPException(
                     status_code=422, detail=f"Invalid folder name: {str(e)}"
                 ) from e
 
-            written_paths.append(target_path)
             file_id = str(uuid4())
-            await _write_upload_with_size_limit(uploaded, target_path)
 
             content_preview = ""
             file_extension = Path(uploaded.filename).suffix.lower()
@@ -513,20 +588,6 @@ def _parse_task_id(task_id: Optional[str]) -> Optional[int]:
         return int(task_id)
     except ValueError as exc:
         raise HTTPException(status_code=400, detail="Invalid task_id") from exc
-
-
-def _build_unique_file_path(path: Path) -> Path:
-    if not path.exists():
-        return path
-    stem = path.stem
-    suffix = path.suffix
-    parent = path.parent
-    i = 1
-    while True:
-        candidate = parent / f"{stem}_{i}{suffix}"
-        if not candidate.exists():
-            return candidate
-        i += 1
 
 
 def _ensure_under_uploads(path: Path, user_id: int) -> None:
--- a/tests/web/api/test_upload_connection_boundary.py
+++ b/tests/web/api/test_upload_connection_boundary.py
@@ -3,11 +3,14 @@
 from __future__ import annotations
 
 import asyncio
+import builtins
 import io
 import threading
+import time
 from pathlib import Path
 
 import pytest
+from fastapi import HTTPException
 from fastapi.datastructures import UploadFile
 
 from xagent.core.file_storage.factory import get_unscoped_file_storage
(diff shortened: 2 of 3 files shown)


src/xagent/web/api/files.py
```diff
@@ -288,35 +289,107 @@ def _accel_redirect_response(
...
-                    raise HTTPException(
-                        status_code=413,
-                        detail=(
-                            f"File size exceeds maximum limit of {MAX_FILE_SIZE_LABEL}"
-                        ),
-                    )
-                buffer.write(chunk)
+        path.unlink(missing_ok=True)
     except BaseException:
+        logger.warning("Failed to clean up local upload: %s", path)
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in 28211423. The cleanup helper now catches only OSError, so SystemExit and KeyboardInterrupt propagate while operational unlink failures remain best-effort. Added regression coverage for both process-control exceptions and the OSError cleanup path; the shared upload and V1 suites pass 99 tests.

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,476 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+        payload = await upload.read()
+        target_path.write_bytes(payload)
+        return len(payload)
+
+    monkeypatch.setattr(
+        files_api,
+        "_write_upload_with_size_limit",
+        interleaved_write,
+        raising=False,
+    )
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in eadc1cc7. The regression now synchronizes two threads at the actual `_reserve_and_copy_upload` boundary with `threading.Barrier`, asserts that both workers entered, and then invokes the original implementation. Replacing exclusive reservation with overwrite mode makes the test fail; the shared upload and V1 suites pass 99 tests.

src/xagent/web/api/files.py
```diff
@@ -355,18 +428,20 @@ async def store_uploaded_files(
                 )
 
             try:
-                target_path = _build_unique_file_path(
-                    get_upload_path(uploaded.filename, task_id, folder, user_id)
+                target_path, _file_size = await _stage_uploaded_file(
```
rogercloud: **Minor (dead value):** `total_size` is used only inside `_reserve_and_copy_upload` for the 413 check, which is self-contained before the return — and it's discarded here as `_file_size`, never referenced again. The authoritative size on the DB record/response comes from an independent `stat()` in `build_uploaded_file_record` (`src/xagent/web/services/uploaded_file_store.py:1276`), not from this value. Its only real consumer is a unit assertion in `tests/web/api/test_kb_web_ingestion_uploaded_file.py:111`, i.e. production code carrying a value that exists to satisfy a test. Consider dropping it from the return signature (returning just the path), or add a comment if it's kept intentionally for future use.

src/xagent/web/api/files.py
```diff
@@ -288,35 +289,107 @@ def _accel_redirect_response(
...
+    """
+
+    try:
+        path = get_upload_path(uploaded.filename or "", task_id, folder, user_id)
+    except ValueError as exc:
+        raise _InvalidUploadPathError(str(exc)) from exc
+    stem = path.stem
+    suffix = path.suffix
+    candidate = path
+    suffix_index = 1
```
rogercloud, line 330: **Minor (robustness, pre-existing shape):** this `while True` / `suffix_index += 1` probe is unbounded and O(N²) across N same-named uploads. The shape predates this PR (`_build_unique_file_path` was the same), so not a regression — but this PR is what made collision handling depend on `FileExistsError` from the exclusive `open`. If a later change ever widened the `except FileExistsError` clause to also cover errors raised from inside the copy body, this loop would spin forever, and the test guarding exactly that (`test_reserve_and_copy_does_not_mistake_source_file_exists_error_for_collision`) would hang rather than fail — there's no `pytest-timeout` marker on it and no global `--timeout` in the pytest addopts, so CI would only notice via a job-level timeout. Cheap insurance: cap the retries (raise after e.g. 10000 attempts) and/or add `@pytest.mark.timeout(...)` to that test.

src/xagent/web/api/files.py
```diff
@@ -288,35 +289,107 @@ def _accel_redirect_response(
...
+            task_id=task_id,
+            folder=folder,
+            user_id=user_id,
+        )
+    )
+    (target_path, file_size), cancellation = await await_task_settlement(worker)
+    # The request cleanup owner knows this exact path before any subsequent
+    # preview, registration, or cancellation-cleanup await point.
+    written_paths.append(target_path)
+    if cancellation is not None:
```
rogercloud, line 386: **Minor (duplicated ownership):** this internal delete-and-drain is redundant. `written_paths.append(target_path)` on the line above runs before this branch, and `store_uploaded_files`'s `finally` already unlinks every path in `written_paths` on any exception — including cancellation. So the caller's cleanup fires regardless of this block. Removing lines 387-391 leaves the entire suite green except one white-box unit test that asserts this exact internal behavior in isolation (`test_cancelled_staging_worker_deletes_its_exact_late_result`); no route-level test depends on it.

Suggest picking a single owner: let `store_uploaded_files`'s `finally` be the sole cleanup path, drop this block, and keep `written_paths` as the one cleanup contract.

src/xagent/web/api/files.py
```diff
@@ -288,35 +289,107 @@ def _accel_redirect_response(
...
+    uploaded: UploadFile,
+    *,
+    task_id: str | None,
+    folder: str | None,
+    user_id: int,
+    written_paths: list[Path],
+) -> tuple[Path, int]:
+    """Run local staging off-loop and drain exact cleanup on cancellation."""
+
+    worker = asyncio.create_task(
```
rogercloud, line 373: **Informational (not a regression, not blocking):** bare `asyncio.to_thread` puts upload staging on the event loop's default `ThreadPoolExecutor`, which is also what `run_db_io_cancellation_safe` in `src/xagent/web/services/db_runtime.py` uses — there's no dedicated executor anywhere under `src/xagent/web/`. Under many large concurrent uploads, staging work can queue ahead of DB-bound pool work.

To be clear this is a strict improvement over the base commit, where these writes were awaited directly on the loop (full loop stall rather than one pool slot). Just flagging as a forward-looking option: a dedicated executor for uploads, or a semaphore bounding concurrent staging.

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,475 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+            task_type="general",
+            task_id=None,
+            folder=None,
+            user_id=user_id,
+            single_file_mode=True,
+        )
+    )
+    try:
+        await asyncio.sleep(0.01)
+        assert asyncio.get_running_loop().time() - started_at < 0.04
```
rogercloud: **Minor (potential flake):** `await asyncio.sleep(0.01)` followed by `< 0.04` leaves only ~30ms of slack on a wall-clock assertion — plausible to trip on a loaded CI runner.

Worth noting this assertion is *not* redundant with the `read_threads` check below, so please don't just delete it: the thread check proves the read happened off-loop, while this one separately proves the loop wasn't blocked waiting on it (a bug that read off-thread but then synchronously blocked on `.result()` would pass the thread check and fail only this one). Suggest widening the margin instead — e.g. `< 0.2`.

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,475 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+
+    return get_upload_path
+
+
+class _SlowTrackingSource(io.BytesIO):
+    """A synchronous source that reveals an accidental event-loop read."""
+
+    def __init__(self, payload: bytes, delay_seconds: float) -> None:
+        super().__init__(payload)
+        self._rolled = False
```
rogercloud: **Minor (dead test state):** `self._rolled` is never read anywhere in this file. It looks like a leftover from drafting against the old async-read path (it was meant to drive Starlette's `UploadFile._in_memory`), but the new worker reads `uploaded.file` directly and bypasses `UploadFile.read/write/seek/close` entirely. Safe to delete.

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,475 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+        self.release = threading.Event()
+
+    def read(self, size: int = -1) -> bytes:
+        self.started.set()
+        assert self.release.wait(timeout=2)
+        return super().read(size)
+
+
+def _admin_user_id() -> int:
+    _admin_headers()
```
rogercloud, line 124: **Minor (clarity):** `_admin_headers()` is called purely for its bootstrap side effect and the returned headers dict is discarded. Correct, but not obvious without cross-referencing `_admin_headers`/`_setup_admin`. A one-line comment would help, e.g. `# Bootstraps the admin user; headers unused here.`

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,475 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+) -> None:
+    """Exclusive-open failures are I/O failures, never candidate collisions."""
+
+    user_id = 7
+    monkeypatch.setattr(files_api, "get_upload_path", _stage_path_in(tmp_path, user_id))
+
+    def failing_open(*_args, **_kwargs) -> None:  # type: ignore[no-untyped-def]
+        raise PermissionError("open denied")
+
+    monkeypatch.setattr(builtins, "open", failing_open)
```
rogercloud, line 351: **Minor (test isolation):** this replaces `builtins.open` wholesale with no passthrough for non-`"x"` modes, unlike the shim at line 306 which calls `original_open` unconditionally and only intercepts when `"x" in mode`. No flakiness was actually observed here, so this is a style nit — but matching the safer passthrough shape (or patching the specific call site) would be more surgical.

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,475 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
     return checked_out
 
 
+def _stage_path_in(root: Path, user_id: int):
```
rogercloud, line 76: **Minor (coverage gap):** the headline change in this PR is that path construction (`get_upload_path`, driven by `task_id`/`folder`) moved *into* the synchronous worker — but every new test in this file stubs `get_upload_path` out via this `_stage_path_in` helper, which explicitly discards `task_id` and `folder`. Searching `tests/web/api/`, `tests/web/api/v1/`, and `tests/web/test_file_upload.py` turned up no test anywhere that drives the **real** `get_upload_path` through this worker on the success path (mkdir + copy) with a non-trivial `folder`/`task_id`; existing tests either pass trivial/`None` values or only cover the real function's rejection path (422 on invalid folder).

So nothing currently proves `_reserve_and_copy_upload` forwards and uses `task_id`/`folder` correctly against real path-construction logic from inside a worker thread. Suggest one test that lets the real `get_upload_path` run with a genuine nested `folder` + `task_id` and asserts the file lands at the expected path with the expected contents.

tests/web/api/test_upload_connection_boundary.py
```diff
@@ -70,6 +73,475 @@ def recording_sync(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+        )
+    assert not list(tmp_path.rglob("open-error*"))
+
+
+@pytest.mark.parametrize(
+    "control_error",
+    [SystemExit("stop"), KeyboardInterrupt()],
+    ids=["system-exit", "keyboard-interrupt"],
+)
+def test_delete_staged_upload_preserves_process_control(
```
rogercloud, line 367: **Minor (test coverage of the PF1 fix):** narrowing `_delete_staged_upload` from `except BaseException` to `except OSError` was the right call, but the new regression tests only prove (a) `SystemExit`/`KeyboardInterrupt` propagate and (b) `OSError` is swallowed. Widening the handler back to `except Exception:` still leaves both tests passing — neither raises something that is an `Exception` but not an `OSError`. Suggest one more case that raises e.g. `ValueError` from `unlink` and asserts it propagates, so the narrowing itself is actually pinned.

rogercloud (approved):
## Summary

This PR consolidates local upload staging — path construction, exclusive path reservation, chunked source reads, size-limit enforcement, and destination writes — into a single synchronous worker (`_reserve_and_copy_upload`) executed via `asyncio.to_thread`, so none of that blocking file I/O runs on the event loop. Reserving the destination with exclusive creation (`open(path, "xb")`) inside the same worker also closes the TOCTOU window in the previous "probe for a free name, then write" sequence, making same-name concurrent uploads atomically distinct. The worker is wrapped by `_stage_uploaded_file`, which drains the thread on cancellation and hands the reserved path to the request-level cleanup owner. Shared across the JWT, V1, widget, and share upload routes.

Related in spirit (not a direct fix) to #935: same "keep blocking I/O off the loop" principle, applied to uploads.

## Approach verdict

**Sound in direction, acceptable-with-reservations in execution.** One synchronous owner doing path construction + `O_EXCL` reservation + same-descriptor writes off the loop is the right and simplest fix for both the loop-blocking problem and the filename race, and it matches the codebase's established `to_thread` + `await_task_settlement` idiom. Reservations are about duplicated ownership in the cleanup path and thread-pool coupling — details, not direction. No blocking issues.

## Design-level findings

**D1 — Shared thread pool for uploads (informational, not a regression, not blocking).**
`_reserve_and_copy_upload` runs on a bare `asyncio.to_thread`, i.e. the event loop's default `ThreadPoolExecutor`, which is also what `run_db_io_cancellation_safe` uses (no dedicated executor exists anywhere under `src/xagent/web/`). Large concurrent uploads can therefore queue ahead of DB-bound thread-pool work. Explicitly **not a regression**: at base, upload writes were awaited directly on the loop, which stalled the whole loop rather than one pool slot — this PR is a strict improvement. Forward-looking suggestion only: a dedicated executor, or a semaphore bounding concurrent upload staging.

**D2 — Cancellation cleanup is implemented twice; one copy is dead weight (minor).**
`_stage_uploaded_file` appends the reserved path to the caller's `written_paths` **and** spawns its own delete-in-a-thread + drain before re-raising on cancellation. But `store_uploaded_files`'s `finally` already unlinks every path in `written_paths` on any exception including cancellation, and the `append` happens *before* the cancellation branch — so the caller's cleanup fires either way. Removing files.py:387-391 leaves the whole suite green except one white-box unit test that asserts exactly this internal behavior; no route-level test depends on it. Recommend picking a single owner — most naturally, let `store_uploaded_files`'s `finally` be the sole cleanup path and drop the internal delete-and-drain.

**D3 — Unbounded collision-probe loop (minor, pre-existing shape).**
The `while True` / `suffix_index += 1` probe is O(N²) across N same-named uploads. This shape is unchanged from the pre-PR `_build_unique_file_path`, so not introduced here. The reason it's worth flagging *now* is that this PR is what made collision handling exclusive-creation-based: if a future change widened `except FileExistsError` to also catch errors raised from inside the copy body, the loop would spin forever — and the regression test that guards precisely that (`test_reserve_and_copy_does_not_mistake_source_file_exists_error_for_collision`) would **hang instead of failing**, since there's no `pytest-timeout` marker on it and no global `--timeout` in the pytest addopts. Cheap insurance: a retry cap (raise after e.g. 10000 attempts) and/or a timeout marker on that test.

**D4 — Cancellation is no longer prompt; undocumented behavior change (minor).**
Because the entire chunked copy is one `to_thread` call and `await_task_settlement` deliberately drains rather than abandons, a cancelled upload must let the worker finish copying the remaining bytes (up to `MAX_FILE_SIZE`, default 100MB) before cleanup runs. Previously `_write_upload_with_size_limit` awaited each chunk in async code, so cancellation landed at the next chunk boundary. Mitigated in practice — Starlette fully buffers the upload before the handler runs, so this is disk-to-disk only, worst case low hundreds of ms to a few seconds, not unbounded. Still a real change and not mentioned; consider adding it to the PR description's "Compatibility and scope".

## Line-level findings

All minor; details in the inline comments.

| # | Location | Issue |
|---|---|---|
| 1 | `src/xagent/web/api/files.py:527` (pre-existing code, unchanged by this diff — noted for awareness, not an inline comment) | `_delete_local_paths` duplicates `_delete_staged_upload` verbatim instead of calling it |
| 2 | `src/xagent/web/api/files.py:431` | Returned `total_size` is dead in production; only a test consumes it |
| 3 | `tests/web/api/test_upload_connection_boundary.py:364` | Narrowing to `except OSError` has no regression coverage for the narrowing itself |
| 4 | `tests/web/api/test_upload_connection_boundary.py:234` | ~30ms wall-clock slack, flake-prone on loaded CI |
| 5 | `tests/web/api/test_upload_connection_boundary.py:99` | `self._rolled` is never read — leftover dead test state |
| 6 | `tests/web/api/test_upload_connection_boundary.py:124` | `_admin_headers()` called for side effect only, not obvious |
| 7 | `tests/web/api/test_upload_connection_boundary.py:76` | Real `get_upload_path` never exercised through the new worker — genuine coverage gap |
| 8 | `tests/web/api/test_upload_connection_boundary.py:348` | Blanket `builtins.open` replacement with no passthrough fallback |

## Re-review: prior automated findings

Both prior Gemini Code Assist findings are **FIXED**, verified independently against HEAD:

- **PF1** — `_delete_staged_upload` catching `BaseException` (would swallow `KeyboardInterrupt`/`SystemExit`). Fixed in `28211423`, narrowed to `except OSError`, with passing regression tests (`test_delete_staged_upload_preserves_process_control`, parametrized over `SystemExit`/`KeyboardInterrupt`, plus `test_delete_staged_upload_suppresses_operational_unlink_failure`). See line-level finding #3 for a residual coverage gap in the narrowing itself.
- **PF2** — `test_concurrent_same_name_uploads_reserve_distinct_paths_and_contents` mocked the since-deleted `_write_upload_with_size_limit`, making it vacuous. Fixed in `eadc1cc7`, rewritten to synchronize two threads at the real `_reserve_and_copy_upload` boundary via `threading.Barrier`. Confirmed to have teeth: reverting `"xb"` to `"wb"` makes it fail with a `UNIQUE constraint failed` IntegrityError.

**Lean already.** The chunked read/write loop, the atomic-collision retry, the reused `db_runtime` helpers, and the test doubles are all proportional to what the change requires. Nothing to remove.

## Test results

96/96 shared upload-boundary + V1 upload tests pass, re-run independently 3x for stability.

## Recommendation

**Approve** — no critical or major findings survived verification. Everything above is minor or informational, and both prior automated findings are fixed with real regression coverage. D2 (single cleanup owner), D4 (document the cancellation-latency change), and line-level finding #7 (a test through the real `get_upload_path`) are the three most worth doing, but none needs to block merge.

src/xagent/web/api/files.py
AlexLiu190625: Fixed in c4507daf. `_reserve_and_copy_upload` and `_stage_uploaded_file` now return only the staged `Path`; the local byte count remains internal to maximum-size enforcement. The real response size continues to come from durable registration.

src/xagent/web/api/files.py
AlexLiu190625: Not changed. The suffix-probe policy predates this change, and the exclusive-open `FileExistsError` handler remains narrowly scoped outside the copy body. The regression test explicitly proves that a source-side `FileExistsError` escapes instead of advancing the loop. Adding a cap or UUID fallback would change filename behavior without incident evidence, so that policy is kept out of this boundary fix.

src/xagent/web/api/files.py
AlexLiu190625: Fixed in c4507daf. The staging helper now drains the worker, records the late path, and propagates cancellation. `store_uploaded_files` is the sole completed-staging cleanup owner, with a route-level regression asserting exactly one unlink.

src/xagent/web/api/files.py
AlexLiu190625: Acknowledged and intentionally not changed here. Upload staging and database workers still share the default executor. A dedicated executor or semaphore needs a process-level admission/configuration owner and workload evidence; this PR only removes event-loop file I/O and records the executor boundary in its scope.

tests/web/api/test_upload_connection_boundary.py
AlexLiu190625: Fixed in c4507daf. The blocking source delay is now 200 ms and the responsiveness threshold is 100 ms, preserving the separate event-loop assertion with substantially more scheduler margin.

tests/web/api/test_upload_connection_boundary.py
AlexLiu190625: Fixed in c4507daf. The unused `_rolled` test state was removed.

tests/web/api/test_upload_connection_boundary.py
AlexLiu190625: Fixed in c4507daf. `_admin_user_id` now states that `_admin_headers()` bootstraps the admin record when the test database is empty.

tests/web/api/test_upload_connection_boundary.py
AlexLiu190625: Fixed in c4507daf. The mock intercepts only exclusive-create modes and delegates every other mode to the original `open`.

tests/web/api/test_upload_connection_boundary.py
AlexLiu190625: Fixed in c4507daf. A success-path test now runs the real production `get_upload_path` with `task_id=42` and `folder=documents`, then verifies the nested destination and copied bytes.

tests/web/api/test_upload_connection_boundary.py
AlexLiu190625: Fixed in c4507daf. A new regression makes `Path.unlink` raise `ValueError` and verifies that it propagates, pinning the handler to `OSError` rather than all `Exception` subclasses.

PR #1019, opened 2026-07-26
fix: offload team knowledge base visibility lookup

AlexLiu190625:
## Summary

- move the synchronous Team Knowledge Base visibility hook off the asyncio event loop
- keep Session ownership inside the existing application hook
- harden the constrained-pool tests without changing unrelated process-global hooks

## Root cause

The async document-search collection path called the synchronous Team Knowledge Base visibility hook directly. When that hook waited for a database connection, the event-loop thread waited with it and delayed unrelated async work.

## What changed

- invoke `visible_team_knowledge_bases(None, user_id)` through the shared cancellation-safe database worker boundary
- document that a hook called with `db=None` may run in a worker thread and must create, use, and close a thread-confined Session
- isolate only the visibility hook in this test suite instead of clearing access and lifecycle hooks
- give the one-slot pool test a realistic two-second checkout window
- put the timeout inside `threading.Event.wait` so a failed test cannot leave a blocked default-executor worker behind
- preserve anonymous/admin bypasses, visibility errors, owner lookup order, and result shaping

## Compatibility and scope

- no API, schema, migration, collection-response, frontend, or SDK changes
- this PR changes only the document-search visibility lookup
- direct `kb.py` hook callers still use request-owned Sessions; moving those calls to a worker would cross Session thread ownership and requires a separate owner redesign
- no new hook registry or parallel ownership layer is introduced

## Verification

- `uv run pytest tests/core/tools/core/test_document_search_team_visibility_boundary.py tests/web/test_team_sharing_hooks.py -q`
- `git diff --check` passed

This is follow-up event-loop hardening related to #935; it does not close the broader task-runtime acceptance criteria in that issue.

### DIFF
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -30,6 +30,7 @@ async def _list_visible_collections(
     if user_id is None or is_admin:
         return result
 
+    from ....web.services.db_runtime import run_db_io_cancellation_safe
     from ....web.services.knowledge_base_team_scope import (
         visible_team_knowledge_bases,
     )
@@ -38,7 +39,10 @@ async def _list_visible_collections(
         collection.name: collection for collection in result.collections
     }
     refs_by_owner: dict[int, list] = {}
-    for ref in visible_team_knowledge_bases(None, int(user_id)):
+    team_refs = await run_db_io_cancellation_safe(
+        lambda: visible_team_knowledge_bases(None, int(user_id))
+    )
+    for ref in team_refs:
         refs_by_owner.setdefault(ref.storage_user_id, []).append(ref)
     for storage_user_id, refs in refs_by_owner.items():
         owner_result = await list_collections(user_id=storage_user_id, is_admin=False)
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -32,8 +32,12 @@ async def _list_visible_collections(
 
     from ....web.services.db_runtime import run_db_io_cancellation_safe
     from ....web.services.knowledge_base_team_scope import (
+        has_knowledge_base_visibility_hook,
         visible_team_knowledge_bases,
     )
+
+    if not has_knowledge_base_visibility_hook():
+        return result
 
     collections_by_name = {
         collection.name: collection for collection in result.collections
--- a/tests/core/tools/core/test_document_search_team_visibility_boundary.py
+++ b/tests/core/tools/core/test_document_search_team_visibility_boundary.py
@@ -1,0 +1,261 @@
+"""Regression coverage for the document-search team visibility boundary."""
+
+from __future__ import annotations
+
+import asyncio
+import threading
+from contextlib import suppress
+
+import pytest
+from sqlalchemy import create_engine, text
+from sqlalchemy.orm import Session
+from sqlalchemy.pool import QueuePool
+
+from xagent.core.tools.core import document_search
+from xagent.core.tools.core.RAG_tools.core.schemas import (
+    CollectionInfo,
+    ListCollectionsResult,
+)
+from xagent.web.services.knowledge_base_team_scope import (
+    KnowledgeBaseAccess,
+    set_knowledge_base_team_hooks,
+)
+
+
+@pytest.fixture(autouse=True)
+def _clear_team_knowledge_base_hooks():
+    """Keep process-global application hooks isolated between tests."""
+    set_knowledge_base_team_hooks()
+    yield
+    set_knowledge_base_team_hooks()
+
+
+def _collections_result(*collections: CollectionInfo) -> ListCollectionsResult:
+    return ListCollectionsResult(
+        status="success",
+        collections=list(collections),
+        total_count=len(collections),
+        message="ok",
+    )
+
+
+def _install_collection_listing(monkeypatch: pytest.MonkeyPatch) -> CollectionInfo:
+    personal = CollectionInfo(name="personal")
+    shared = CollectionInfo(name="shared")
+
+    async def list_for_user(
+        user_id: int | None = None, is_admin: bool = False
+    ) -> ListCollectionsResult:
+        del is_admin
+        if user_id == 2:
+            return _collections_result(shared)
+        return _collections_result(personal)
+
+    monkeypatch.setattr(document_search, "list_collections", list_for_user)
+    return personal
+
+
+def _one_slot_engine():
+    return create_engine(
+        "sqlite://",
+        connect_args={"check_same_thread": False},
+        poolclass=QueuePool,
+        pool_size=1,
+        max_overflow=0,
+        pool_timeout=0.25,
+    )
+
+
+async def _wait_for(event: threading.Event) -> None:
+    await asyncio.wait_for(asyncio.to_thread(event.wait), timeout=5)
+
+
+async def _await_without_leaking(task: asyncio.Task[object]) -> None:
+    if not task.done():
+        task.cancel()
+    with suppress(asyncio.CancelledError, Exception):
+        await task
+
+
+@pytest.mark.asyncio
+async def test_team_visibility_hook_waits_off_loop_and_releases_pool(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A blocked hook checkout must leave the loop responsive until release."""
+    _install_collection_listing(monkeypatch)
+    engine = _one_slot_engine()
+    held_connection = engine.connect()
+    hook_started = threading.Event()
+    hook_closed = threading.Event()
+    ticker_advanced = threading.Event()
+    observed: dict[str, int] = {}
+    loop_thread_id = threading.get_ident()
+
+    def visibility_hook(db: Session | None, user_id: int) -> list[KnowledgeBaseAccess]:
+        assert db is None
+        assert user_id == 1
+        observed["hook_thread_id"] = threading.get_ident()
+        hook_started.set()
+        try:
+            with Session(engine) as session:
+                observed["sql_thread_id"] = threading.get_ident()
+                session.execute(text("SELECT 1"))
+        finally:
+            hook_closed.set()
+        return [KnowledgeBaseAccess(name="shared", storage_user_id=2)]
+
+    async def tick_after_hook_starts() -> None:
+        await _wait_for(hook_started)
+        await asyncio.sleep(0)
+        ticker_advanced.set()
+
+    set_knowledge_base_team_hooks(visibility=visibility_hook)
+    ticker = asyncio.create_task(tick_after_hook_starts())
+    listing = asyncio.create_task(
+        document_search._list_visible_collections(user_id=1, is_admin=False)
+    )
+    try:
+        await _wait_for(hook_started)
+        await _wait_for(ticker_advanced)
+        assert not listing.done()
+        assert engine.pool.checkedout() == 1
+
+        held_connection.close()
+        result = await listing
+        await _wait_for(hook_closed)
+
+        assert observed["hook_thread_id"] != loop_thread_id
+        assert observed["sql_thread_id"] != loop_thread_id
+        assert [collection.name for collection in result.collections] == [
+            "personal",
+            "shared",
+        ]
+        shared = next(
+            collection
+            for collection in result.collections
+            if collection.name == "shared"
+        )
+        assert shared.ownership == "team"
+        assert shared.storage_user_id == 2
+        assert engine.pool.checkedout() == 0
+    finally:
+        if not held_connection.closed:
+            held_connection.close()
+        await _await_without_leaking(listing)
+        await _await_without_leaking(ticker)
+        engine.dispose()
+
+
+@pytest.mark.asyncio
+async def test_cancellation_drains_team_visibility_session_before_propagating(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Cancellation must wait for the hook-owned Session to close."""
+    _install_collection_listing(monkeypatch)
+    engine = _one_slot_engine()
+    held_connection = engine.connect()
+    hook_started = threading.Event()
+    hook_closed = threading.Event()
+
+    def visibility_hook(db: Session | None, user_id: int) -> list[KnowledgeBaseAccess]:
+        assert db is None
+        assert user_id == 1
+        hook_started.set()
+        try:
+            with Session(engine) as session:
+                session.execute(text("SELECT 1"))
+        finally:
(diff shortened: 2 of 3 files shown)


src/xagent/core/tools/core/document_search.py
```diff
@@ -38,7 +39,10 @@ async def _list_visible_collections(
         collection.name: collection for collection in result.collections
     }
     refs_by_owner: dict[int, list] = {}
-    for ref in visible_team_knowledge_bases(None, int(user_id)):
+    team_refs = await run_db_io_cancellation_safe(
+        lambda: visible_team_knowledge_bases(None, int(user_id))
+    )
```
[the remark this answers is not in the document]
AlexLiu190625: The call already passes None so the embedding hook can own its temporary Session. The established hook owner is xagent.web.services.knowledge_base_team_scope, and document_search imported that hook before this PR; xagent.core.hooks does not exist. Adding a parallel registry here would bypass the installed hook and expand the change beyond moving the existing synchronous invocation onto the cancellation-safe worker boundary, so this suggestion is not adopted in this PR.

tests/core/tools/core/test_document_search_team_visibility_boundary.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def _one_slot_engine():
+    return create_engine(
+        "sqlite://",
+        connect_args={"check_same_thread": False},
+        poolclass=QueuePool,
+        pool_size=1,
+        max_overflow=0,
+        pool_timeout=0.25,
```
rogercloud: `pool_timeout=0.25` is too tight and makes this module flaky as written. `test_team_visibility_hook_waits_off_loop_and_releases_pool` races three `asyncio.to_thread` scheduling round-trips against this 250ms checkout deadline; running it 20 times locally gave 4 failures, all `TimeoutError` from the pool rather than the assertion the test is meant to make. Bumping this to `1.0`-`2.0` removes the race without weakening anything: the test still proves the second checkout has to wait for the first to be released. Worth fixing before merge — otherwise this will fail intermittently in CI.

tests/core/tools/core/test_document_search_team_visibility_boundary.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+
+async def _wait_for(event: threading.Event) -> None:
+    await asyncio.wait_for(asyncio.to_thread(event.wait), timeout=5)
```
rogercloud: `asyncio.wait_for` cancels the await, not the worker thread — `event.wait()` has no timeout, so on a timeout here the thread stays blocked forever. Low impact in practice (each test gets its own loop and default executor), but `asyncio.to_thread(event.wait, 5)` makes the helper self-limiting. Optional.

tests/core/tools/core/test_document_search_team_visibility_boundary.py
```diff
@@ -0,0 +1,261 @@
...
+from xagent.web.services.knowledge_base_team_scope import (
+    KnowledgeBaseAccess,
+    set_knowledge_base_team_hooks,
+)
+
+
+@pytest.fixture(autouse=True)
+def _clear_team_knowledge_base_hooks():
+    """Keep process-global application hooks isolated between tests."""
+    set_knowledge_base_team_hooks()
```
rogercloud: Bare `set_knowledge_base_team_hooks()` resets all four hook globals (`visibility`, `access`, `renamed`, `deleted`), not just the one this module installs. Nothing breaks today since no test registers those outside function scope, but passing through / restoring only `visibility` would keep this fixture from clobbering a future module-scoped hook fixture. Non-blocking.

src/xagent/core/tools/core/document_search.py
```diff
@@ -38,7 +39,10 @@ async def _list_visible_collections(
         collection.name: collection for collection in result.collections
     }
     refs_by_owner: dict[int, list] = {}
-    for ref in visible_team_knowledge_bases(None, int(user_id)):
+    team_refs = await run_db_io_cancellation_safe(
```
rogercloud, line 46: This is the first place `run_db_io_cancellation_safe` wraps an application-installed callback rather than an in-repo helper, so the visibility hook can now run on a worker thread. That contract isn't stated anywhere the hook author would see it — `visible_team_knowledge_bases` only documents "hooks must open a session when `db` is None", and `KnowledgeBaseVisibilityHook` says nothing. Suggest a one-line note in `src/xagent/web/services/knowledge_base_team_scope.py` that the hook may be invoked off the event-loop thread and must not rely on thread-local state. Non-blocking (no in-tree hook does).

rogercloud (commented):
## Summary

This PR takes the one remaining synchronous DB call on the document-search tool's async path — `visible_team_knowledge_bases(None, int(user_id))` inside `_list_visible_collections` — and routes it through the existing `run_db_io_cancellation_safe` boundary, so the team-visibility hook's connection-pool checkout happens on a worker thread instead of blocking the event loop. It is a 6-line production change plus a new 261-line test module that exercises pool-wait behavior, cancellation draining, error identity, and the no-hook/admin/anonymous bypasses.

The mechanism is the right one and matches how the rest of the codebase crosses this boundary after #986. My reservations are about test reliability and about how the PR is framed, not about the fix itself.

## Approach and design notes

**Verdict: acceptable with reservations.** The chosen boundary is correct and idiomatic:

- `run_db_io_cancellation_safe` runs the callable via `asyncio.to_thread` and drains the worker before propagating cancellation, so a cancelled caller cannot leave an orphaned session holding a pooled connection. Exception identity from the hook is preserved through the boundary.
- Nothing in this repo relies on thread-affine session state (no `scoped_session`, no `threading.local` around DB access), and `contextvars` propagate into `to_thread`, so moving this call to a worker thread does not change observable behavior here.
- The function-local `core → web.services` import was raised in the earlier Gemini review, which suggested a `core.hooks` registry instead. The author's reply declining that is sound: `xagent.core.hooks` does not exist, the sibling `knowledge_base_team_scope` import at this same site predates the PR, and this exact deferred-import pattern is used at dozens of sites under `core/` specifically to avoid circular imports. **Resolved via prior discussion — no action needed.**

Three design-level notes that are not tied to a single changed line:

1. **The "hook must open its own session" precondition becomes an unenforceable contract.** This is the first place the `run_db_io_cancellation_safe` boundary wraps an *application-installed callback* rather than an in-repo helper. `src/xagent/web/services/knowledge_base_team_scope.py:57-63` documents only "hooks must open a session when `db` is None"; it does not say the hook may now run off the event-loop thread, and the `KnowledgeBaseVisibilityHook` alias at `src/xagent/web/services/knowledge_base_team_scope.py:29` carries no such note either. A downstream implementation using thread-local or thread-scoped state would break silently rather than loudly. Suggest a one-line docstring/alias comment stating the hook may be invoked on a worker thread. Not blocking — there is no such implementation in-tree.

2. **Scope is narrower than the title suggests.** `visible_team_knowledge_bases` has a second caller, `list_collections_api` (`src/xagent/web/api/kb.py:4786`), which still invokes it synchronously on the loop with a live request-scoped `Session`. Three sibling hooks are in the same shape: `resolve_knowledge_base_access` (`src/xagent/web/api/kb.py:3228`), `notify_knowledge_base_deleted` (`src/xagent/web/api/kb.py:6482`, `:6544`), `notify_knowledge_base_renamed` (`src/xagent/web/api/kb.py:7614`). All pre-existing and untouched here, and converting the `kb.py` call is genuinely not a drop-in change (a caller-owned `Session` cannot simply move to a worker thread). Keeping them out is a fine scope decision — please just say so explicitly in the PR body and/or file a follow-up issue, because "offload team knowledge base visibility lookup" reads as covering the whole hook class. Similarly, this PR does not touch any of #935's five confirmed incident paths; those landed in #986.

3. **Inherited properties of the boundary** (pre-existing, informational, not this PR's to fix): because the worker is drained before cancellation propagates, a caller-side timeout around this call is not promptly enforced — a 0.1s `asyncio.timeout` around a 1.0s blocking op surfaces at ~1.0s, with a worst-case ceiling of `XAGENT_DB_POOL_TIMEOUT_SECONDS` (default 30s). This is intentional and tested design shared by 20+ existing call sites; the document-search path now inherits it. Relatedly, when a caller is cancelled *and* the wrapped worker then raises, `run_db_io_cancellation_safe` attaches the worker error as `CancelledError.__cause__` and logs it nowhere — so that class of error is silently lost repo-wide across ~100 call sites. Worth knowing; out of scope here.

Also pre-existing and out of scope: `_list_visible_collections` and `list_collections_api` both implement personal+team collection merging, but `kb.py` fans out over owners concurrently under a timeout while `document_search.py` scans owners sequentially with none. Untouched by this diff.

## Findings

### 1. `pool_timeout=0.25` makes the new tests flaky today (major — will break CI)

`tests/core/tools/core/test_document_search_team_visibility_boundary.py:65`

`_one_slot_engine()` sets `pool_timeout=0.25` on a size-1 pool. `test_team_visibility_hook_waits_off_loop_and_releases_pool` then races three `asyncio.to_thread` scheduling round-trips against that 250ms SQLAlchemy checkout deadline. Running that test 20 times locally produced **4 failures out of 20 (20%)**: when thread scheduling is even slightly slow, the pool times out before the test's own teardown releases the held connection, and the test fails with `TimeoutError` instead of ever reaching the assertion it exists to make. This does not match the "24 tests passed" claim in the PR description.

Please raise `pool_timeout` to 1-2s. The assertion strength is unaffected — the test still proves the second checkout has to *wait*, which is the whole point; only the arbitrary deadline that the test itself must beat goes away. This is the one item I'd hold merge on.

### 2. PR description overstates what the new tests regression-guard (minor, description-only)

Reverting the production change locally, **2 of the 6 new tests fail**: `test_team_visibility_hook_waits_off_loop_and_releases_pool` and `test_cancellation_drains_team_visibility_session_before_propagating`. The other four — the error-identity test and the three bypass tests — pass identically with and without the fix. They are useful behavior-preservation tests, but describing them as "error-identity and bypass *regression* tests" for this change is not accurate. Suggest rewording the PR body; no code change needed.

### 3. Hook thread-affinity is undocumented (minor)

See design note 1 above. Concretely: add a sentence to the `visible_team_knowledge_bases` docstring (`src/xagent/web/services/knowledge_base_team_scope.py:57-63`) and/or beside `KnowledgeBaseVisibilityHook` (`:29`) saying the visibility hook may be called on a worker thread and must not depend on running on the event-loop thread.

### 4. `_wait_for` can leak a blocked thread on failure (minor, optional)

`tests/core/tools/core/test_document_search_team_visibility_boundary.py:70`

`asyncio.wait_for(asyncio.to_thread(event.wait), timeout=5)` cancels the *await*, not the thread — `event.wait()` has no timeout, so if the outer `wait_for` ever fires, the worker thread stays blocked forever. Impact here is low (this repo's pytest-asyncio config gives each test its own loop and default executor, so it cannot cascade into unrelated tests), but `event.wait(5)` gets you defense in depth for free.

### 5. Autouse fixture clears all four hooks, not just `visibility` (minor, future-proofing)

`tests/core/tools/core/test_document_search_team_visibility_boundary.py:26-30`

`_clear_team_knowledge_base_hooks` calls bare `set_knowledge_base_team_hooks()`, which resets `visibility`, `access`, `renamed` and `deleted` process-globals. No current test installs any of those outside function scope, so nothing breaks today — noting it only because a future module-scoped hook fixture would be silently clobbered by this file.

Lean already. The one candidate I looked at — `_one_slot_engine()` standing up a real SQLAlchemy engine rather than a fake — does not hold up: the genuine pool-exhaustion behavior *is* the thing under test, and it is exactly what makes tests 1 and 2 fail when the fix is reverted. No changes recommended.

## Test results

- `pytest tests/core/tools/core/test_document_search_team_visibility_boundary.py -p no:warnings` → **6 passed**
- Same test repeated 20x → **4/20 failures** from the `pool_timeout=0.25` race (finding 1), not from a code defect
- Related suite, `tests/core/tools/core/ -k "team or visibility or knowledge"` → **8 passed**
- Revert-the-fix check → 2 of 6 new tests fail (finding 2)
- `mypy` clean on the changed file

Overall: the fix is correct and I'm happy with the approach. Fix the `pool_timeout` value and this is good to merge; everything else is optional or informational.

tests/core/tools/core/test_document_search_team_visibility_boundary.py
AlexLiu190625: Fixed in 7b7f130a. The constrained QueuePool now uses a two-second checkout timeout, retaining the one-slot wait invariant without racing normal CI scheduling.

tests/core/tools/core/test_document_search_team_visibility_boundary.py
AlexLiu190625: Fixed in 7b7f130a. `_wait_for` now passes the timeout into `threading.Event.wait` and raises if it returns false, so a failed assertion cannot leave an unbounded default-executor worker.

tests/core/tools/core/test_document_search_team_visibility_boundary.py
AlexLiu190625: Fixed in 7b7f130a. The suite now installs and restores only `_visibility_hook` through `monkeypatch`; a regression verifies that an unrelated access hook remains intact.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed through 73bafe78. The hook owner now documents that visibility callbacks may run on a worker thread, must own a thread-confined Session when `db=None`, and must not depend on event-loop thread-local state.

rogercloud, line 46: Minor: when no visibility hook is configured, this still queues a guaranteed-empty lookup on the shared default executor. If that executor is saturated, collection listing—and cancellation—waits behind unrelated work, whereas the previous path returned immediately. Please consider exposing/checking whether a visibility callback is installed and calling `run_db_io_cancellation_safe` only for an actual callback; a saturated-executor no-hook test would cover this.

rogercloud (approved):
The event-loop fix and follow-up test hardening look good. Approving with one non-blocking P3 inline comment about avoiding executor dispatch in the default no-hook configuration.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed in 5d50afc6. The hook owner now exposes a capability predicate, and document search returns the personal collection result before scheduling worker work when no visibility hook is installed. The regression test saturates the sole default-executor worker and verifies that the no-hook path still completes. The focused 8-test suite and pre-commit checks pass.

PR #1020, opened 2026-07-26
fix: revoke undelivered runtime API keys

AlexLiu190625:
## Summary

- keep V1 one-shot runtime API key creation cancellation-safe after commit
- retain a fenced transition receipt and compensate only the transition that failed delivery
- restore the previously delivered active keys when a new rotation is not delivered
- serialize rotation and compensation on the same cross-database transition fence
- preserve caller cancellation, original worker errors, and process-control exception precedence

## Root cause

V1 agent creation and runtime-key rotation can commit a new credential before the runtime returns the plaintext key to the route. If cancellation or post-commit response mapping fails, revoking only the new row prevents exposure of an undelivered secret but also leaves every previously delivered key revoked.

The original transition fence also had two concurrency gaps: compensation did not acquire the same agent-row lock as rotation, and SQLite ignores `FOR UPDATE`. Those gaps could let overlapping transitions retain more than one active replacement key. Compensation also needed to yield to a later administrative pause.

## What changed

- use one runtime-key transition fence in rotation and compensation:
  - PostgreSQL and MySQL use `SELECT ... FOR UPDATE` on the agent row
  - SQLite uses a no-op parent-row write before the key snapshot
  - the SQLite statement bypasses the ORM `updated_at` handler
- snapshot the exact active key ids replaced by each transition
- record the new key id, agent id, prefix, replaced key ids, and rotation timestamp in a detached receipt
- centralize staging, commit ambiguity handling, and post-commit response construction in one generic service boundary
- let the worker Session boundary build successful detached outcomes, removing duplicate create/rotate envelope construction
- compensate in a fresh worker-owned Session with three fences:
  - acquire the same per-agent transition fence used by rotation
  - revoke the new key only while its exact id, owner, prefix, active state, and unpaused state still match
  - restore only receipt-listed prior rows whose `revoked_at` still equals this transition timestamp, and only when the new-key CAS changed one row
- leave all prior rows untouched when a later rotation or pause superseded the receipt, so later state wins
- log new-key and restored-row counts for both successful and no-op compensation
- restore deferred cancellation after the compensation worker settles
- keep compensation failures as notes/log evidence without replacing an existing worker error cause
- share the process-control classifier and keep runtime-specific tests with the runtime owner

## Compatibility and scope

- no API response, database schema, or migration changes
- the legacy `/api/agents/{agent_id}/api-key` shares the serialized snapshot/revoke primitive, but its response contract is unchanged and it does not gain post-commit delivery compensation
- covered paths are `POST /v1/agents`, `POST /v1/agents/from-template`, and `POST /v1/agents/{agent_id}/api-key`
- the compensation boundary ends when `AgentManagementRuntime` returns its detached result to the route; it does not claim to observe a later HTTP transport failure
- these other one-shot credential paths remain outside this PR:
  - `POST /api/agents/{agent_id}/api-key`
  - `POST /api/agent-api-keys`
  - `POST /api/agent-api-keys/{key_id}/regenerate`
  - `POST /api/me/personal-keys`
- extending the contract to those endpoints requires a shared pending-delivery state spanning their different ownership and rotation semantics; this PR intentionally does not partially retrofit them
- compensation still uses the configured database pool and timeout. A separate compensation engine or background retry owner is not introduced because it would change deterministic cleanup and deployment configuration

## Verification

- `uv run pytest -q tests/web/api/v1/test_agent_management_runtime.py tests/web/api/test_agents_management.py`
- `uv run pytest -q tests/web/api/test_agent_api_keys.py tests/web/api/v1/test_auth.py`
- `XAGENT_TEST_POSTGRES_URL=... uv run pytest -q tests/web/services/test_runtime_key_transition_postgres.py`
- `git diff --check` passed

This is V1 control-plane hardening related to #935; it does not close the task-runtime acceptance criteria in that issue.

### DIFF
--- a/src/xagent/web/services/agent_management.py
+++ b/src/xagent/web/services/agent_management.py
@@ -7,9 +7,10 @@ import logging
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from types import TracebackType
-from typing import Any, Callable, Literal, cast
+from typing import Any, Callable, Generic, Literal, TypeVar, cast
 
 from sqlalchemy import func, or_, select, update
+from sqlalchemy.engine import CursorResult
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session, aliased
 
@@ -35,11 +36,17 @@ from .api_keys import (
     RuntimeKeyDeliveryError,
     RuntimeKeyReceipt,
 )
-from .db_runtime import await_task_settlement, run_db_io_cancellation_safe
+from .db_runtime import (
+    await_task_settlement,
+    is_process_control_exception,
+    propagate_deferred_cancellation,
+    run_db_io_cancellation_safe,
+)
 from .workforce_access import can_edit_workforce, filter_visible_workforces
 from .workforce_lifecycle import is_workforce_manager_discard_safe
 
 logger = logging.getLogger(__name__)
+_RuntimeKeyResultT = TypeVar("_RuntimeKeyResultT")
 
 # Agent-builder tool category that gates knowledge-base access. A KB
 # selection is only valid when this category is also enabled.
@@ -231,13 +238,21 @@ class AgentDeleteResult:
 
 
 @dataclass(frozen=True)
-class _RuntimeKeyDeliveryOutcome:
+class _RuntimeKeyDeliveryOutcome(Generic[_RuntimeKeyResultT]):
     """Detached worker result that retains an exact key receipt on failure."""
 
-    result: object | None
+    result: _RuntimeKeyResultT | None
     receipt: RuntimeKeyReceipt | None
     error: BaseException | None
     traceback: TracebackType | None
+
+
+@dataclass(frozen=True)
+class _RuntimeKeyCompensationResult:
+    """Auditable row counts from one fenced compensation transaction."""
+
+    new_key_revoked: int
+    prior_keys_restored: int
 
 
 class AgentWorkforceConflictError(RuntimeError):
@@ -587,6 +602,7 @@ class AgentManagementService:
         the conflict translation here must be split by source (agent ->
         duplicate-name 400, key -> 409).
         """
+        self.runtime_key_receipt = None
         if self.store.agent_name_exists(user_id, name):
             raise DuplicateAgentNameError(name)
 
@@ -609,31 +625,19 @@ class AgentManagementService:
         # here (agent has no unique constraint), so the conflict-to-409
         # translation wraps key staging + commit together. See the
         # contract note in the docstring above.
-        staged_key = None
-        self.runtime_key_receipt = None
-        commit_attempted = False
-        try:
+        def stage_runtime_key() -> tuple[AgentApiKey, str] | None:
             if generate_runtime_key:
                 staged_key = self.key_service.stage_rotated_key(
                     int(agent.id),
                     candidate=runtime_key_candidate,
                 )
                 self.runtime_key_receipt = self.key_service.runtime_key_receipt
-            commit_attempted = True
-            self.db.commit()  # single transaction boundary for both writes
-        except IntegrityError as exc:
-            self.db.rollback()
-            raise KeyRotationConflict(str(exc)) from exc
-        except BaseException as exc:
-            if commit_attempted and self.runtime_key_receipt is not None:
-                raise RuntimeKeyDeliveryError(
-                    self.runtime_key_receipt,
-                    exc,
-                    exc.__traceback__,
-                ) from exc
-            raise
+                return staged_key
+            return None
 
-        try:
+        def build_response(
+            staged_key: tuple[AgentApiKey, str] | None,
+        ) -> tuple[Agent, APIKeyGenerateResponse | None]:
             key_resp: APIKeyGenerateResponse | None = None
             self.db.refresh(agent)
             agent_id = int(agent.id)
@@ -665,14 +669,11 @@ class AgentManagementService:
                     exc_info=True,
                 )
             return agent, key_resp
-        except BaseException as exc:
-            if self.runtime_key_receipt is not None:
-                raise RuntimeKeyDeliveryError(
-                    self.runtime_key_receipt,
-                    exc,
-                    exc.__traceback__,
-                ) from exc
-            raise
+
+        return self.key_service.complete_runtime_key_delivery(
+            stage=stage_runtime_key,
+            build_response=build_response,
+        )
 
     async def create_agent_from_template(
         self,
@@ -872,10 +873,6 @@ def _is_runtime_key_prefix_collision(error: BaseException) -> bool:
     return False
 
 
-def _is_process_control(error: BaseException) -> bool:
-    return not isinstance(error, (Exception, asyncio.CancelledError))
-
-
 async def _validate_agent_knowledge_bases(
     *,
     knowledge_bases: tuple[str, ...],
@@ -933,8 +930,11 @@ class AgentManagementRuntime:
 
     @staticmethod
     def _run_runtime_key_session_operation(
-        operation: Callable[[AgentManagementService], _RuntimeKeyDeliveryOutcome],
-    ) -> _RuntimeKeyDeliveryOutcome:
+        operation: Callable[
+            [AgentManagementService],
+            _RuntimeKeyDeliveryOutcome[_RuntimeKeyResultT],
+        ],
+    ) -> _RuntimeKeyDeliveryOutcome[_RuntimeKeyResultT]:
         """Detach runtime-key outcomes before closing their worker Session.
 
         A committed key receipt is needed even when the service reports a
@@ -945,7 +945,7 @@ class AgentManagementRuntime:
 
         db: Session | None = None
         service: AgentManagementService | None = None
-        outcome: _RuntimeKeyDeliveryOutcome | None = None
+        outcome: _RuntimeKeyDeliveryOutcome[_RuntimeKeyResultT] | None = None
         try:
             SessionLocal = get_session_local()
             db = SessionLocal()
@@ -964,7 +964,7 @@ class AgentManagementRuntime:
                     result=None,
                     receipt=(
                         None
-                        if isinstance(exc, KeyRotationConflict) or service is None
+                        if isinstance(exc, KeyRotationConflict)
                         else service.runtime_key_receipt
                     ),
                     error=exc,
@@ -989,7 +989,7 @@ class AgentManagementRuntime:
                             error=close_error,
                             traceback=close_error.__traceback__,
                         )
-                    elif _is_process_control(close_error):
+                    elif is_process_control_exception(close_error):
                         outcome = _RuntimeKeyDeliveryOutcome(
                             result=None,
                             receipt=outcome.receipt,
@@ -1035,7 +1035,7 @@ class AgentManagementRuntime:
         *,
         user_id: int,
         spec: AgentCreateSpec,
-    ) -> _RuntimeKeyDeliveryOutcome:
+    ) -> _RuntimeKeyDeliveryOutcome[AgentCreateSnapshot]:
         attempts = PREFIX_COLLISION_RETRIES if spec.generate_runtime_key else 1
         for attempt in range(attempts):
             # Candidate generation includes bcrypt. It deliberately precedes
@@ -1048,7 +1048,8 @@ class AgentManagementRuntime:
 
             def create_attempt(
                 service: AgentManagementService,
-            ) -> _RuntimeKeyDeliveryOutcome:
+                candidate: ApiKeyCandidate | None = candidate,
+            ) -> _RuntimeKeyDeliveryOutcome[AgentCreateSnapshot]:
                 agent, key_response = service.create_agent_with_optional_key(
                     user_id=user_id,
                     name=spec.name,
@@ -1182,13 +1183,14 @@ class AgentManagementRuntime:
         *,
         user_id: int,
         agent_id: int,
-    ) -> _RuntimeKeyDeliveryOutcome:
+    ) -> _RuntimeKeyDeliveryOutcome[RuntimeKeySnapshot]:
         for attempt in range(PREFIX_COLLISION_RETRIES):
             candidate = generate_api_key(None, kind=ApiKeyKind.AGENT)
 
             def rotate_attempt(
                 service: AgentManagementService,
-            ) -> _RuntimeKeyDeliveryOutcome:
+                candidate: ApiKeyCandidate = candidate,
+            ) -> _RuntimeKeyDeliveryOutcome[RuntimeKeySnapshot]:
                 response = service.generate_agent_runtime_key(
                     user_id=user_id,
                     agent_id=agent_id,
@@ -1223,41 +1225,46 @@ class AgentManagementRuntime:
 
     async def _run_runtime_key_delivery(
         self,
-        operation: Callable[[], _RuntimeKeyDeliveryOutcome],
-    ) -> object | None:
+        operation: Callable[[], _RuntimeKeyDeliveryOutcome[_RuntimeKeyResultT]],
+    ) -> _RuntimeKeyResultT | None:
         """Settle key delivery before exposing cancellation or worker failure."""
 
         worker = asyncio.create_task(asyncio.to_thread(operation))
         outcome, cancellation = await await_task_settlement(worker)
-        compensation_error: BaseException | None = None
-        if outcome.receipt is not None and (
-            cancellation is not None or outcome.error is not None
-        ):
-            compensation_error = await self._compensate_runtime_key(outcome.receipt)
+        with propagate_deferred_cancellation(cancellation):
+            compensation_error: BaseException | None = None
+            if outcome.receipt is not None and (
+                cancellation is not None or outcome.error is not None
+            ):
+                compensation_error = await self._compensate_runtime_key(outcome.receipt)
 
-        if outcome.error is not None:
-            error = outcome.error.with_traceback(outcome.traceback)
-            if _is_process_control(error):
+            if outcome.error is not None:
+                error = outcome.error.with_traceback(outcome.traceback)
+                if compensation_error is not None:
+                    error.add_note(
+                        "Runtime-key compensation also failed: "
+                        f"{type(compensation_error).__name__}: {compensation_error}"
+                    )
                 raise error
-            if cancellation is not None:
-                raise cancellation from error
             if compensation_error is not None:
-                raise error from compensation_error
-            raise error
-        if cancellation is not None:
-            if compensation_error is not None:
-                raise cancellation from compensation_error
-            raise cancellation
+                # Compensation only runs without a worker error when caller
+                # cancellation was captured. Raising here lets the shared
+                # boundary preserve that cancellation with this failure as its
+                # cause.
+                raise compensation_error
         return outcome.result
 
     @staticmethod
-    def _revoke_runtime_key_sync(receipt: RuntimeKeyReceipt) -> None:
-        """Idempotently revoke only the exact key produced by this request."""
+    def _compensate_runtime_key_sync(
+        receipt: RuntimeKeyReceipt,
+    ) -> _RuntimeKeyCompensationResult:
+        """Fence a failed transition and restore only its replaced key rows."""
 
         SessionLocal = get_session_local()
         with SessionLocal() as db:
             now = datetime.now(timezone.utc)
-            try:
+            revoke_result = cast(
+                "CursorResult[Any]",
                 db.execute(
                     update(AgentApiKey)
                     .where(
@@ -1267,11 +1274,45 @@ class AgentManagementRuntime:
                         AgentApiKey.revoked_at.is_(None),
                     )
                     .values(revoked_at=now, updated_at=now)
+                ),
+            )
+            new_key_revoked = max(int(revoke_result.rowcount or 0), 0)
+            prior_keys_restored = 0
+            if (
+                new_key_revoked == 1
+                and receipt.replaced_key_ids
+                and receipt.rotation_timestamp is not None
+            ):
+                restore_result = cast(
+                    "CursorResult[Any]",
+                    db.execute(
+                        update(AgentApiKey)
+                        .where(
+                            AgentApiKey.agent_id == receipt.agent_id,
+                            AgentApiKey.id.in_(receipt.replaced_key_ids),
+                            AgentApiKey.revoked_at == receipt.rotation_timestamp,
+                        )
+                        .values(revoked_at=None, updated_at=now)
+                    ),
                 )
-                db.commit()
-            except BaseException:
-                db.rollback()
-                raise
+                prior_keys_restored = max(int(restore_result.rowcount or 0), 0)
+            db.commit()
+
+        result = _RuntimeKeyCompensationResult(
+            new_key_revoked=new_key_revoked,
+            prior_keys_restored=prior_keys_restored,
+        )
+        logger.info(
+            "Compensated undelivered runtime key id=%s agent_id=%s prefix=%s "
+            "(new_key_revoked=%d, prior_keys_restored=%d, expected_prior=%d)",
+            receipt.key_id,
+            receipt.agent_id,
+            receipt.key_prefix,
+            result.new_key_revoked,
+            result.prior_keys_restored,
+            len(receipt.replaced_key_ids),
+        )
+        return result
 
     async def _compensate_runtime_key(
         self,
@@ -1280,19 +1321,24 @@ class AgentManagementRuntime:
         """Run exact revocation in a fresh worker-owned Session."""
 
         worker = asyncio.create_task(
-            asyncio.to_thread(self._revoke_runtime_key_sync, receipt)
+            asyncio.to_thread(self._compensate_runtime_key_sync, receipt)
         )
         try:
-            _result, _cancellation = await await_task_settlement(worker)
+            _result, cancellation = await await_task_settlement(worker)
+        except asyncio.CancelledError:
+            raise
         except BaseException as exc:
-            if not isinstance(exc, (Exception, asyncio.CancelledError)):
+            if is_process_control_exception(exc):
                 raise
             logger.warning(
-                "Failed to revoke undelivered runtime key id=%s agent_id=%s prefix=%s",
+                "Failed to compensate undelivered runtime key "
+                "id=%s agent_id=%s prefix=%s",
                 receipt.key_id,
                 receipt.agent_id,
                 receipt.key_prefix,
                 exc_info=True,
             )
             return exc
+        with propagate_deferred_cancellation(cancellation):
+            pass
         return None
--- a/src/xagent/web/services/agent_management.py
+++ b/src/xagent/web/services/agent_management.py
@@ -35,6 +35,7 @@ from .api_keys import (
     KeyRotationConflict,
     RuntimeKeyDeliveryError,
     RuntimeKeyReceipt,
+    acquire_runtime_key_transition_fence,
 )
 from .db_runtime import (
     await_task_settlement,
@@ -937,7 +938,7 @@ class AgentManagementRuntime:
     def _run_runtime_key_session_operation(
         operation: Callable[
             [AgentManagementService],
-            _RuntimeKeyDeliveryOutcome[_RuntimeKeyResultT],
+            _RuntimeKeyResultT | None,
         ],
     ) -> _RuntimeKeyDeliveryOutcome[_RuntimeKeyResultT]:
         """Detach runtime-key outcomes before closing their worker Session.
@@ -956,7 +957,13 @@ class AgentManagementRuntime:
             db = SessionLocal()
             service = AgentManagementService(db)
             try:
-                outcome = operation(service)
+                result = operation(service)
+                outcome = _RuntimeKeyDeliveryOutcome(
+                    result=result,
+                    receipt=service.runtime_key_receipt,
+                    error=None,
+                    traceback=None,
+                )
             except RuntimeKeyDeliveryError as exc:
                 outcome = _RuntimeKeyDeliveryOutcome(
                     result=None,
@@ -1055,7 +1062,7 @@ class AgentManagementRuntime:
             def create_attempt(
                 service: AgentManagementService,
                 candidate: ApiKeyCandidate | None = candidate,
-            ) -> _RuntimeKeyDeliveryOutcome[AgentCreateSnapshot]:
+            ) -> AgentCreateSnapshot:
                 agent, key_response = service.create_agent_with_optional_key(
                     user_id=user_id,
                     name=spec.name,
@@ -1073,14 +1080,9 @@ class AgentManagementRuntime:
                 agent_snapshot = _agent_response_snapshot(
                     service.store.agent_to_response_dict(agent)
                 )
-                return _RuntimeKeyDeliveryOutcome(
-                    result=AgentCreateSnapshot(
-                        agent=agent_snapshot,
-                        api_key=_runtime_key_snapshot(key_response),
-                    ),
-                    receipt=service.runtime_key_receipt,
-                    error=None,
-                    traceback=None,
+                return AgentCreateSnapshot(
+                    agent=agent_snapshot,
+                    api_key=_runtime_key_snapshot(key_response),
                 )
 
             outcome = AgentManagementRuntime._run_runtime_key_session_operation(
@@ -1195,18 +1197,13 @@ class AgentManagementRuntime:
             def rotate_attempt(
                 service: AgentManagementService,
                 candidate: ApiKeyCandidate = candidate,
-            ) -> _RuntimeKeyDeliveryOutcome[RuntimeKeySnapshot]:
+            ) -> RuntimeKeySnapshot | None:
                 response = service.generate_agent_runtime_key(
                     user_id=user_id,
                     agent_id=agent_id,
                     runtime_key_candidate=candidate,
                 )
-                return _RuntimeKeyDeliveryOutcome(
-                    result=_runtime_key_snapshot(response),
-                    receipt=service.runtime_key_receipt,
-                    error=None,
-                    traceback=None,
-                )
+                return _runtime_key_snapshot(response)
 
             outcome = AgentManagementRuntime._run_runtime_key_session_operation(
                 rotate_attempt
@@ -1268,41 +1265,51 @@ class AgentManagementRuntime:
         SessionLocal = get_session_local()
         with SessionLocal() as db:
             now = datetime.now(timezone.utc)
-            revoke_result = cast(
-                "CursorResult[Any]",
-                db.execute(
-                    update(AgentApiKey)
-                    .where(
-                        AgentApiKey.id == receipt.key_id,
-                        AgentApiKey.agent_id == receipt.agent_id,
-                        AgentApiKey.key_prefix == receipt.key_prefix,
-                        AgentApiKey.revoked_at.is_(None),
-                    )
-                    .values(revoked_at=now, updated_at=now)
-                    .execution_options(synchronize_session=False)
-                ),
+            agent_exists = acquire_runtime_key_transition_fence(
+                db,
+                receipt.agent_id,
             )
-            new_key_revoked = max(int(revoke_result.rowcount or 0), 0)
+            new_key_revoked = 0
             prior_keys_restored = 0
-            if (
-                new_key_revoked == 1
-                and receipt.replaced_key_ids
-                and receipt.rotation_timestamp is not None
-            ):
-                restore_result = cast(
+            if agent_exists:
+                revoke_result = cast(
                     "CursorResult[Any]",
                     db.execute(
                         update(AgentApiKey)
                         .where(
+                            AgentApiKey.id == receipt.key_id,
                             AgentApiKey.agent_id == receipt.agent_id,
-                            AgentApiKey.id.in_(receipt.replaced_key_ids),
-                            AgentApiKey.revoked_at == receipt.rotation_timestamp,
+                            AgentApiKey.key_prefix == receipt.key_prefix,
+                            AgentApiKey.revoked_at.is_(None),
+                            AgentApiKey.paused_at.is_(None),
                         )
-                        .values(revoked_at=None, updated_at=now)
+                        .values(revoked_at=now, updated_at=now)
                         .execution_options(synchronize_session=False)
                     ),
                 )
-                prior_keys_restored = max(int(restore_result.rowcount or 0), 0)
+                new_key_revoked = max(int(revoke_result.rowcount or 0), 0)
+                if (
+                    new_key_revoked == 1
+                    and receipt.replaced_key_ids
+                    and receipt.rotation_timestamp is not None
+                ):
+                    restore_result = cast(
+                        "CursorResult[Any]",
+                        db.execute(
+                            update(AgentApiKey)
+                            .where(
+                                AgentApiKey.agent_id == receipt.agent_id,
+                                AgentApiKey.id.in_(receipt.replaced_key_ids),
+                                AgentApiKey.revoked_at == receipt.rotation_timestamp,
+                            )
+                            .values(revoked_at=None, updated_at=now)
+                            .execution_options(synchronize_session=False)
+                        ),
+                    )
+                    prior_keys_restored = max(
+                        int(restore_result.rowcount or 0),
+                        0,
+                    )
             db.commit()
 
         result = _RuntimeKeyCompensationResult(
--- a/src/xagent/web/services/agent_management.py
+++ b/src/xagent/web/services/agent_management.py
@@ -2,12 +2,14 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 from dataclasses import dataclass
-from datetime import datetime
-from typing import Any, Literal, cast
+from datetime import datetime, timezone
+from types import TracebackType
+from typing import Any, Callable, Literal, cast
 
-from sqlalchemy import func, or_, select
+from sqlalchemy import func, or_, select, update
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session, aliased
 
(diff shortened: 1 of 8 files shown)


src/xagent/web/services/agent_management.py
```diff
@@ -1021,37 +1169,130 @@ async def rotate_agent_runtime_key(
...
+        self,
+        receipt: RuntimeKeyReceipt,
+    ) -> BaseException | None:
+        """Run exact revocation in a fresh worker-owned Session."""
+
+        worker = asyncio.create_task(
+            asyncio.to_thread(self._revoke_runtime_key_sync, receipt)
         )
+        try:
+            _result, _cancellation = await await_task_settlement(worker)
```
rogercloud: **Critical — the compensation path silently swallows cancellation, defeating this PR's central guarantee.**

`_result, _cancellation = await await_task_settlement(worker)` discards `_cancellation`. Per `await_task_settlement`'s documented contract (`db_runtime.py:30-65`), a cancellation that arrives while the *compensation* worker is running (as opposed to the original worker) causes it to return normally with the cancellation deferred into that second return value. Dropping it loses the cancellation entirely: `_run_runtime_key_delivery` then re-raises the original `RuntimeError` instead of `CancelledError`.

Confirmed with a runnable reproduction — after the cancel, `task.cancelled() == False` and `task.cancelling() == 1`, i.e. a live, unresolved cancellation request that never surfaces. That silently breaks `asyncio.TaskGroup`, `wait_for`, and Starlette's cancellation propagation, which is exactly the property the PR claims to establish.

The repo already has the right primitive: `propagate_deferred_cancellation` (`db_runtime.py:69-92`), paired with `await_task_settlement` correctly in `websocket.py:2485-2488` and `task_command_transport.py:871-875`. This module's import at `agent_management.py:38` pulls in only `await_task_settlement` and then hand-rolls the cancellation-priority logic incorrectly.

Fix: use `propagate_deferred_cancellation` here, matching the established pattern. Please also add the missing test — cancellation arriving during compensation *with the revoke succeeding*. The two existing compensation-interaction tests (`tests/web/services/test_agent_management_runtime.py:826`, `:874`) both mock the revoke to raise, which routes through a different, correct branch, so this scenario has zero coverage today.

src/xagent/web/services/agent_management.py
```diff
@@ -1021,37 +1169,130 @@ async def rotate_agent_runtime_key(
...
+            compensation_error = await self._compensate_runtime_key(outcome.receipt)
+
+        if outcome.error is not None:
+            error = outcome.error.with_traceback(outcome.traceback)
+            if _is_process_control(error):
+                raise error
+            if cancellation is not None:
+                raise cancellation from error
+            if compensation_error is not None:
+                raise error from compensation_error
```
rogercloud: **Major — exception chain is corrupted and causally backwards on the double-failure path.**

`raise error from compensation_error` unconditionally overwrites `error.__cause__`, discarding the original driver-level cause precisely in the scenario that most needs it: a DB outage that causes both the original failure *and* the compensation failure. Nothing logs the original chain before this line, so that context is gone for good.

The direction is also inverted — `error` happened first and is what triggered the compensation attempt, not the other way round.

Suggested shape: log `error` with its intact chain (`exc_info=True`) before re-raising, and attach the compensation failure as context rather than replacing the cause — e.g. re-raise `error` untouched and record `compensation_error` separately, or use `add_note` if you want them in one traceback.

This branch (`error is not None AND not cancelled AND compensation_error is not None`) also has zero test coverage — all six compensation-failure-adjacent tests go through either the plain success-revoke path or the cancellation path.

src/xagent/web/services/agent_management.py
```diff
@@ -1021,37 +1169,130 @@ async def rotate_agent_runtime_key(
...
+        SessionLocal = get_session_local()
+        with SessionLocal() as db:
+            now = datetime.now(timezone.utc)
             try:
-                with SessionLocal() as db:
-                    response = AgentManagementService(db).generate_agent_runtime_key(
-                        user_id=user_id,
-                        agent_id=agent_id,
-                        runtime_key_candidate=candidate,
+                db.execute(
```
rogercloud: **Major — successful compensation leaves no audit trail.**

`_revoke_runtime_key_sync` discards the `UPDATE`'s rowcount and logs nothing on success; only the failure case warns (`:1290`). Operators therefore cannot distinguish "we successfully revoked an orphaned live credential" from "the compare-and-set predicate matched nothing and this was a no-op" — on a security-sensitive cleanup path where that distinction is the whole point.

Sibling code added by this same PR already does the right thing: `api_keys.py:143` and `:272` log at INFO with row counts. Please match that here — capture `result.rowcount` and log at INFO with `key_id` / `agent_id` / `key_prefix` and whether a row was actually revoked.

Two smaller things on this statement:
- It omits `synchronize_session=False`, inconsistent with the sibling UPDATEs at `api_keys.py:124` and `:262` which pass it explicitly.
- Coverage gap: the `revoked_at IS NULL` predicate at `:1267` — the compare-and-set property the PR description names by name — has **zero** coverage. Removing it locally leaves all 46 relevant tests passing. Please add a test that runs compensation against an already-revoked key and asserts no row is touched.

src/xagent/web/services/api_keys.py
```diff
@@ -108,6 +135,11 @@ def stage_rotated_key(
         )
         self.db.add(new_row)
         self.db.flush()
+        self.runtime_key_receipt = RuntimeKeyReceipt(
```
rogercloud, line 198: **Major — rotation compensation can leave an agent with zero usable keys.**

`stage_rotated_key` bulk-revokes **all** of the agent's active keys before staging the new one, in the same transaction as the insert. If delivery of the new key subsequently fails or is cancelled, the compensation in `agent_management.py:1253-1274` revokes only the new key — it never restores the previously-active ones.

Reproduced directly: rotate an agent that already has an active key, cancel after the commit, and both the old and new rows end with `revoked_at IS NOT NULL`. The agent is silently locked out of its own API, from an ordinary user action (rotate, then client disconnect).

Every existing rotation/cancellation test creates the target agent with `generate_runtime_key=False`, so there is never a pre-existing key and this path has no coverage at all.

Options: capture the revoked key ids in the receipt and un-revoke them during compensation (still fail-open, but at least complete), or move to the stage-inert-then-activate shape discussed in the review body, which makes this scenario structurally impossible. Either way the PR description's "compensation is exact and cancellation-safe" needs amending.

Also: this method's docstring doesn't mention its new side effect of setting `self.runtime_key_receipt`.

src/xagent/web/services/agent_management.py
```diff
@@ -1021,37 +1169,130 @@ async def rotate_agent_runtime_key(
...
-        raise KeyRotationConflict(
-            "Failed to generate a unique runtime key prefix after retrying."
+
+    async def _compensate_runtime_key(
+        self,
+        receipt: RuntimeKeyReceipt,
+    ) -> BaseException | None:
+        """Run exact revocation in a fresh worker-owned Session."""
+
+        worker = asyncio.create_task(
```
rogercloud, line 1337: **Minor — the failure path now performs a second non-cancellable blocking DB-pool checkout.**

The compensation worker dispatches through the shared default thread pool and takes a `pool_timeout`-bound checkout from the shared connection pool, on top of the first worker's checkout at `:1230`. This is new behaviour — the base commit's rotate path was a single dispatch with no compensation concept.

Under pool exhaustion this doubles the serial blocking time per affected request, and it happens precisely when the pool is most likely to be saturated (the DB is already misbehaving, which is why we're compensating). Worth bounding the compensation checkout with a shorter timeout, or deferring it to a background task rather than blocking the request's failure path.

src/xagent/web/services/agent_management.py
```diff
@@ -590,41 +610,69 @@ def create_agent_with_optional_key(
         # translation wraps key staging + commit together. See the
         # contract note in the docstring above.
         staged_key = None
+        self.runtime_key_receipt = None
+        commit_attempted = False
```
rogercloud: **Minor (not a bug) — please document the deliberate over-triggering.**

`commit_attempted = True` is set *before* `self.db.commit()` here (and at `:622`, and in `api_keys.py:198`/`:205`), so any non-`IntegrityError` commit failure will trigger a compensation UPDATE against a key that was never committed.

Verified as intentional and correct: commit ambiguity is real, erring toward "attempt to compensate" is the right default, and the only cost is one harmless idempotent no-op UPDATE. But a reader will file this as a bug on first pass. A one-line comment stating the trade-off would save that.

src/xagent/web/services/agent_management.py
```diff
@@ -877,6 +931,84 @@ def _list_agents_sync(*, user_id: int) -> tuple[AgentSummarySnapshot, ...]:
...
+                        logger.warning(
+                            "Failed to close runtime-key worker session after a "
+                            "delivery failure",
+                            exc_info=(
+                                type(close_error),
+                                close_error,
+                                close_error.__traceback__,
+                            ),
+                        )
+        assert outcome is not None
```
rogercloud, line 1021: **Nit (style only).** `assert outcome is not None` is strippable under `python -O`. Verified unreachable through this function's own control flow regardless, and this codebase never runs with `-O`/`PYTHONOPTIMIZE` (checked all Docker and deploy configs); bare invariant asserts are established convention in this service layer (`task_orchestrator.py:1380`, `widget_domains.py:68`, `uploaded_file_store.py:1663-1665`). Flagging for consistency awareness only — no change needed.

src/xagent/web/services/agent_management.py
```diff
@@ -822,6 +872,10 @@ def _is_runtime_key_prefix_collision(error: BaseException) -> bool:
     return False
 
 
+def _is_process_control(error: BaseException) -> bool:
```
rogercloud: **Minor — duplicate definition.** `_is_process_control` here is byte-for-byte identical to `db_runtime.py:14-15`, and this module already imports other names from that module at `:38`. Add it to the existing import and delete the local copy.

src/xagent/web/services/agent_management.py
```diff
@@ -912,42 +1045,57 @@ def _create_agent_with_retry_sync(
...
+                    description=spec.description,
+                    instructions=spec.instructions,
+                    execution_mode=spec.execution_mode,
+                    models=spec.model_mapping(),
+                    knowledge_bases=list(spec.knowledge_bases),
+                    skills=list(spec.skills),
+                    tool_categories=list(spec.tool_categories),
+                    suggested_prompts=list(spec.suggested_prompts),
+                    generate_runtime_key=spec.generate_runtime_key,
+                    runtime_key_candidate=candidate,
```
rogercloud, line 1078: **Minor — late-binding closure over `candidate`.** `ruff check --select B023` flags this and `:1195` (`create_attempt` / `rotate_attempt`) as real B023 violations. Harmless today because both closures are invoked synchronously within the same loop iteration that defines them, but it's fragile to any future change that defers invocation.

Worth noting the project's ruff config currently has no `B` rules enabled (confirmed against `pyproject.toml` — a plain `ruff check` passes), so this is a genuine gap in lint coverage rather than a deliberate style choice. Binding via a default arg (`def create_attempt(candidate=candidate):`) fixes both sites.

src/xagent/web/services/agent_management.py
```diff
@@ -590,41 +610,69 @@ def create_agent_with_optional_key(
         # translation wraps key staging + commit together. See the
         # contract note in the docstring above.
         staged_key = None
+        self.runtime_key_receipt = None
```
rogercloud: **Minor (latent) — reset ordering.** `self.runtime_key_receipt = None` runs *after* two validation calls. Safe today only because a fresh service instance is constructed per attempt (verified — no reuse path exists), but if that ever changes, a validation failure would leave a stale receipt from a prior attempt and compensation would revoke the wrong key. Move the reset to the first line of the method.

src/xagent/web/services/agent_management.py
```diff
@@ -1021,37 +1169,130 @@ async def rotate_agent_runtime_key(
...
-            except KeyRotationConflict as exc:
-                if (
-                    _is_runtime_key_prefix_collision(exc)
-                    and attempt + 1 < PREFIX_COLLISION_RETRIES
-                ):
-                    continue
+                    .values(revoked_at=now, updated_at=now)
+                )
+                db.commit()
+            except BaseException:
```
rogercloud: **Minor — redundant rollback handler.** The `except BaseException: db.rollback(); raise` inside the `with SessionLocal() as db:` block is unnecessary — `Session.__exit__` / `close()` already rolls back an uncommitted transaction (confirmed in the SQLAlchemy source). Dropping it removes a level of nesting with no behaviour change.

src/xagent/web/services/agent_management.py
```diff
@@ -590,41 +610,69 @@ def create_agent_with_optional_key(
...
         try:
             if generate_runtime_key:
                 staged_key = self.key_service.stage_rotated_key(
                     int(agent.id),
                     candidate=runtime_key_candidate,
                 )
+                self.runtime_key_receipt = self.key_service.runtime_key_receipt
+            commit_attempted = True
             self.db.commit()  # single transaction boundary for both writes
         except IntegrityError as exc:
```
rogercloud: **Minor — an important invariant is undocumented and load-bearing across 5 sites.**

"`KeyRotationConflict` implies nothing was committed" is relied on at `agent_management.py:624-626`, `:967`, `:1090`, `~:1214-1219`, and `api_keys.py:207-209`, with no comment linking them. It holds by construction today, but a future change that adds a second write inside either commit block would silently defeat it and leak an orphaned active key — with no test failing. Please state the invariant once at the contract site and reference it from the guards.

Also here (low impact, no action strictly required): if `self.db.rollback()` inside this handler itself raises, the new exception is not caught by the sibling same-level `except BaseException` and propagates with a still-non-`None` receipt, causing a no-op compensation UPDATE.

src/xagent/web/services/agent_management.py
```diff
@@ -1021,37 +1169,130 @@ async def rotate_agent_runtime_key(
...
+        outcome, cancellation = await await_task_settlement(worker)
+        compensation_error: BaseException | None = None
+        if outcome.receipt is not None and (
+            cancellation is not None or outcome.error is not None
+        ):
+            compensation_error = await self._compensate_runtime_key(outcome.receipt)
+
+        if outcome.error is not None:
+            error = outcome.error.with_traceback(outcome.traceback)
+            if _is_process_control(error):
```
rogercloud: **Minor — `BaseException` handling during shutdown.** `SystemExit` / `KeyboardInterrupt` / `GeneratorExit` are re-raised as-is here, but only *after* compensation has been awaited at `:1233-1236`. Two consequences: these bypass FastAPI's `@app.exception_handler(Exception)` (there is no `BaseException` handler registered), and awaiting a DB round-trip on this path can delay process shutdown. Consider skipping compensation — or bounding it with a short timeout — for these three.

src/xagent/web/services/agent_management.py
```diff
@@ -221,6 +230,16 @@ class AgentDeleteResult:
     logo_url: str | None
 
 
+@dataclass(frozen=True)
+class _RuntimeKeyDeliveryOutcome:
```
rogercloud: **Minor (simplification) — make the outcome generic.** `result: object | None` forces unchecked `cast(...)` at `:1031` and `:1178`. Verified by actually applying it: making `_RuntimeKeyDeliveryOutcome` `Generic[_T]` and threading the parameter through eliminates the cast at `:1178` entirely and narrows `:1031` from "recovering an erased type" to a safe non-null assertion, with mypy clean. Low risk, meaningful type-safety gain on a security-relevant path.

src/xagent/web/services/agent_management.py
```diff
@@ -877,6 +931,84 @@ def _list_agents_sync(*, user_id: int) -> tuple[AgentSummarySnapshot, ...]:
...
+                    receipt=exc.receipt,
+                    error=exc.error,
+                    traceback=exc.traceback,
+                )
+            except BaseException as exc:
+                outcome = _RuntimeKeyDeliveryOutcome(
+                    result=None,
+                    receipt=(
+                        None
+                        if isinstance(exc, KeyRotationConflict) or service is None
```
rogercloud: **Minor — dead code.** The `or service is None` clause is unreachable: `service` is unconditionally assigned before the try block this handler guards. Drop the clause.

src/xagent/web/services/api_keys.py
```diff
@@ -153,6 +185,45 @@ def rotate_key(
             created_at=new_row.created_at,
         )
 
+    def rotate_key_for_runtime_delivery(
```
rogercloud, line 285: **Minor (simplification) — near-verbatim duplication.** `rotate_key_for_runtime_delivery` reproduces the `commit_attempted` / receipt / `RuntimeKeyDeliveryError`-wrapping skeleton from `agent_management.py:610` almost line for line (~35-40 lines). A shared helper parameterised by `stage` / `build_result` callables collapses this without awkward indirection — checked, it works cleanly.

Separately: this method and `rotate_key` (`:151`) don't cross-reference each other or explain when to use which, despite the overlap. Please add that to both docstrings.

src/xagent/web/services/agent_management.py
```diff
@@ -590,41 +610,69 @@ def create_agent_with_optional_key(
         # translation wraps key staging + commit together. See the
```
rogercloud, line 632: **Doc.** This docstring documents the IntegrityError conflict contract in detail but never mentions `RuntimeKeyDeliveryError`, which this method can now raise. Worth noting alongside it that the exception is raised from 4 sites and caught at only 1 (`:955`) — `create_agent` / `create_agent_from_template` would leak it uncaught, harmless only because they currently have no live callers.

tests/web/services/test_db_runtime.py
```diff
@@ -8,6 +8,10 @@
 import pytest
 from sqlalchemy.exc import TimeoutError as SQLAlchemyTimeoutError
 
+from xagent.web.services.agent_management import (
```
rogercloud: **Nit — misplaced tests.** This module imports the private `_RuntimeKeyDeliveryOutcome` from `agent_management` and exercises `AgentManagementRuntime._run_runtime_key_delivery` directly (around lines 237 and 271). Those belong in `tests/web/services/test_agent_management_runtime.py`; `test_db_runtime.py` should stay focused on the `db_runtime` primitives. Organizational only.

rogercloud (changes requested):
## Summary

This PR closes a window in the V1 agent-creation and runtime-key-rotation paths where a new API key credential could be committed to the database (bcrypt hash persisted, plaintext held only in memory) before the plaintext was handed back to the caller — leaving an active, permanently undeliverable credential if the request was cancelled or a post-commit step failed. The fix carries a detached delivery receipt (`key_id`, `agent_id`, `key_prefix`) out of the worker thread and, on failure or cancellation, compensates by revoking exactly that key through an idempotent compare-and-set `UPDATE` in a fresh session, while preserving cancellation / process-control / original-error precedence. Diff: `src/xagent/web/services/agent_management.py` (+322/-81), `src/xagent/web/services/api_keys.py` (+71), `src/xagent/web/services/db_runtime.py` (+8), plus extensions to two pre-existing test modules.

## Approach verdict: acceptable with reservations, leaning over-engineered

The problem is real and worth fixing, and the compare-and-set revoke primitive itself is clean. But the chosen shape — fail-open commit plus after-the-fact compensation — is the heavier and weaker of the two available designs, it is applied to only a third of the affected call sites, and it contains a confirmed critical bug in the exact cancellation-safety property the PR claims to guarantee (see C1 inline, `agent_management.py:1286`). That last point is blocking; the rest is design feedback worth resolving before or shortly after merge.

### D1 — A fail-closed alternative gives a strictly better failure mode and does not appear to have been considered

`AgentApiKey.paused_at` already exists (`src/xagent/web/models/agent_api_key.py:49`) and V1 auth already excludes paused keys (`src/xagent/web/api/v1/deps.py:163`, `:337`). The alternative direction is: commit the new key in an inert state, then flip it live with a small idempotent CAS only after delivery succeeds.

Compare the failure modes. Today, a DB outage *during compensation* leaves the credential active and undelivered — the PR description itself acknowledges compensation is only best-effort. Under a stage-inert-then-activate design, the failure path does nothing at all (already safe), and the residual risk moves to the success-path activation, where a failure means "delivered but not yet usable" (availability) rather than "never delivered but fully usable" (security). That is the trade you want on a credential path.

One correction to how this was initially framed: this is **not** schema-free. `paused_at` is a live user-facing feature — pause/resume endpoints at `src/xagent/web/api/agent_api_keys.py:106,118`, service methods at `src/xagent/web/services/api_keys.py:482-497`, and a `status` field surfaced to the frontend via `src/xagent/web/schemas/agent_api_key.py:78`. Reusing it directly would make an undelivered key appear in the user's key list as "paused" with a working Resume button before delivery is confirmed, which is a new race and a UX bug. A dedicated sentinel (new column or distinguishing marker) would be needed. The design point stands regardless; I'd like to see it explicitly weighed rather than skipped.

### D2 — This fixes 2 of 6 structurally identical call sites, and the other 4 are not disclosed as still vulnerable

All four remaining sites commit an active (or overwritten) key before plaintext delivery, with no compensation, and none use the new receipt machinery. All are live, reachable V1 endpoints, unchanged by this diff:

1. `api/agents.py:1282` → `AgentApiKeyService.rotate_key` (`services/api_keys.py:151-186`). The PR describes this as "the legacy transaction contract remains unchanged" — but it is an unfixed instance of the very bug being fixed elsewhere, not merely legacy.
2. `api/agent_api_keys.py:129` → `regenerate_key` (`services/api_keys.py:502-531`). **Worst of the four**: `key_prefix`/`key_hash` are overwritten in place rather than a new row being inserted, so a post-commit delivery failure destroys the old key with nothing to compensate back to. Total lockout for that agent.
3. `api/agent_api_keys.py:59` → `AgentApiKeyService.create_key` (`services/api_keys.py:309-375`).
4. `api/me.py:19` → `UserApiKeyService.create_key` (`services/api_keys.py:559-584`).

If the intent is to fix these in follow-ups, please say so in the PR description and file the issue. As written, the description implies the class of bug is closed when it is not.

### D4 — Rotation compensation can leave an agent with zero usable keys

`stage_rotated_key` (`services/api_keys.py:116-126`) bulk-revokes **all** of the agent's active keys before staging the new one, in the same transaction as the insert. If delivery of the new key then fails or is cancelled, compensation (`agent_management.py:1253-1274`) revokes only the new key and never restores the old ones. Reproduced: rotate an agent that has an existing active key, cancel after commit, and both old and new rows end with `revoked_at IS NOT NULL`. That is an agent silently locked out of its own API, reachable from an ordinary user action (rotate + client disconnect).

Every existing rotation/cancellation test creates the target agent with `generate_runtime_key=False`, so no pre-existing key exists and this path has zero coverage. The PR description's claim that "compensation is exact and cancellation-safe" should be amended — this case is neither.

### D6 / D7 — Weight and precedent

The mechanism is heavier than the problem warrants: +241 net lines in `agent_management.py` alone, a new outcome type there plus two more in `api_keys.py`, `rotate_key_for_runtime_delivery` as a near-duplicate of `rotate_key`, mutable receipt state added to two service classes, and a four-level-nested hand-rolled session lifecycle helper (`_run_runtime_key_session_operation`, `agent_management.py:934-1010`) standing in for the codebase's usual `with SessionLocal() as db:` — including a `finally` block guarding against `db.close()` raising on an already-committed session, which documents a fragile invariant rather than removing it.

Relatedly, the repo already has an established post-commit compensation pattern: `services/uploaded_file_store.py` (`compensation_claim`, `compensate_staged_uploaded_files`, a `storage_status == "compensating"` state machine) and `services/kb_file_service.py`. This PR builds a fully parallel implementation without referencing or reusing either. Even if the existing pattern doesn't fit, the divergence is worth a sentence of justification.

### D5 / D8 / D9 — Scope caveats worth recording

- **D5**: The compensation boundary closes when `_run_runtime_key_delivery` returns, which is *before* the HTTP response is built and written (`api/v1/agents.py:116-123, 201`). "Cancellation-safe up to delivery" means delivery to the route handler, not to the client; a failure during serialization or network write is not covered. Worth stating explicitly in the description.
- **D8** (latent): `RuntimeKeyDeliveryError` is raised from 4 sites but caught at only 1 (`agent_management.py:955`). `AgentManagementService.create_agent` / `create_agent_from_template` would leak it uncaught — harmless today only because those have no live callers. A trap for the next caller.
- **D9**: `agent_management.py:655-666` newly downgrades cache-invalidation failures to a warning (previously hard-failing on the create path). Correct in isolation — but it means the single most likely real-world post-commit failure now bypasses the compensation machinery this PR adds. That tension should be acknowledged rather than left implicit.

## Line-level findings

**Critical**
- **C1** — `agent_management.py:1286`: compensation-path cancellation is silently swallowed, defeating the PR's central claim. Details inline. **This one blocks.**

**Major**
- **C2** — `agent_management.py:1245`: `raise error from compensation_error` corrupts the exception chain and inverts causality on the double-failure path; branch is untested.
- **C5** — `agent_management.py:1261-1270`: successful compensation emits no audit trail at all, on a security-sensitive cleanup path.
- **D4** — `api_keys.py:116`: bulk-revoke-then-stage, with compensation that only unwinds half of it (see above).
- **D2** — `api_keys.py:502`: `regenerate_key` in-place overwrite, the highest-severity unfixed site.

**Minor** — C6, C7, C8, C9, C10, C12, C13, C14, C15, C16, C11: see inline. Of these, C7 and C8 are explicitly **not** bugs (verified conservative-by-design and unreachable respectively) and are flagged only for a clarifying comment or as style nits.

## Test coverage gaps

- **T1** — The compare-and-set predicate `revoked_at IS NULL` (`agent_management.py:1267`) has **zero** coverage: deleting it locally leaves all 46 relevant tests green. This is precisely the idempotency property the PR description names.
- **T2** — No test rotates an agent that already has an active key through a compensation-triggering failure, which is why D4 slipped through.
- **T3** — No test covers `error is not None AND not cancelled AND compensation_error is not None` (the C2 branch); all six compensation-failure-adjacent tests route through the success-revoke or cancellation paths.
- **T4** — No test covers cancellation arriving *during* compensation with the revoke succeeding (the C1 bug). The two existing compensation-interaction tests (`tests/web/services/test_agent_management_runtime.py:826`, `:874`) both mock the revoke to raise, which takes a different, correct branch.
- **T5** — `tests/web/services/test_db_runtime.py:11-14` imports the private `_RuntimeKeyDeliveryOutcome` from `agent_management` and exercises `AgentManagementRuntime._run_runtime_key_delivery` directly. Organizational nit — belongs in `test_agent_management_runtime.py`.

## Documentation gaps

- **Dc1** — `create_agent_with_optional_key`'s docstring documents the IntegrityError conflict contract in detail but never mentions the new `RuntimeKeyDeliveryError` it can now raise.
- **Dc2** — `stage_rotated_key`'s docstring omits its new side effect of setting `self.runtime_key_receipt`.
- **Dc3** — `rotate_key` and `rotate_key_for_runtime_delivery` don't cross-reference each other or say when to use which, despite substantially duplicated bodies.
- **Dc4** — PR description accuracy: (a) "Related to #935" links an unrelated production incident (event-loop blocking from DB pool exhaustion); (b) the two files listed as new tests were pre-existing (441 and 288 lines at the base commit) and only extended; (c) "compensation is exact and cancellation-safe" does not disclose D4.

- `src/xagent/web/services/agent_management.py:875`: **delete** — `_is_process_control` is a byte-for-byte duplicate of `db_runtime.py:14`, which this module already imports from. Import it instead.
- `src/xagent/web/services/agent_management.py:967`: **delete** — the `or service is None` clause is unreachable; `service` is unconditionally assigned before the guarded try block.
- `src/xagent/web/services/agent_management.py:234` (+ casts at `:1031`, `:1178`): **shrink** — make `_RuntimeKeyDeliveryOutcome` `Generic[_T]` instead of `result: object | None`; this fully removes the cast at `:1178` and narrows `:1031` to a safe non-null assertion, mypy clean.
- `src/xagent/web/services/api_keys.py:188` vs `src/xagent/web/services/agent_management.py:610`: **shrink** — `rotate_key_for_runtime_delivery` duplicates the `commit_attempted` / receipt / `RuntimeKeyDeliveryError`-wrapping skeleton nearly verbatim (~35-40 lines). A shared helper taking `stage` / `build_result` callables collapses it cleanly.

net: -40 lines possible

## Testing performed

163+ existing tests pass per the PR description; independently spot-checked 46 tests across the two modified test modules — all green. Mutation testing (deleting compensation calls, weakening the compare-and-set predicate, reverting the process-control guards) confirms the covered behavior is genuinely exercised where coverage exists; the gaps in T1-T4 are real absences rather than measurement artifacts. C1 and D4 were each confirmed with a runnable reproduction.

## Recommendation

Request changes. C1 must be fixed before merge — the correct primitive (`propagate_deferred_cancellation`) already exists in this repo and is used properly in two other places. C2, C5 and the D4 zero-keys scenario should also land here, each with the test that would have caught it. D1, D2 and D6/D7 are design conversations that can be resolved in the description and follow-up issues rather than in this diff.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. Compensation now restores the cancellation returned by `await_task_settlement` through `propagate_deferred_cancellation`. A regression cancels while a successful compensation worker is still running and verifies that `CancelledError` surfaces after settlement.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. The original worker exception is re-raised with its existing cause intact. A compensation failure is logged by the compensation boundary and attached with `add_note`; the double-failure regression verifies that the original driver/mapper cause is not overwritten.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed through 8b19f095. Compensation captures both UPDATE row counts, logs exact key identity plus revoked/restored counts at INFO, and uses `synchronize_session=False`. Coverage includes an already-revoked key and proves the active-state CAS is an audited no-op.

src/xagent/web/services/api_keys.py
AlexLiu190625: Fixed in 8575cee6. The receipt now contains the exact previously active key ids and the rotation timestamp. Compensation revokes the undelivered key by CAS and restores only receipt-listed rows still fenced by that timestamp. Cancellation and mapping-failure regressions verify previous delivered keys remain usable; a later-rotation regression verifies stale compensation restores nothing.

src/xagent/web/services/agent_management.py
AlexLiu190625: Not changed. Compensation intentionally completes before the request failure returns so cleanup is deterministic; moving it to a background task would weaken that guarantee. A separate engine or shorter pool timeout has no existing configuration/lifecycle owner and would introduce a second database policy. The PR body now states that compensation uses the configured shared pool and timeout.

src/xagent/web/services/agent_management.py
AlexLiu190625: Resolved structurally in 8575cee6. The duplicated `commit_attempted` flags were removed. The shared `complete_runtime_key_delivery` boundary captures the receipt before commit and explicitly documents why any ambiguous commit result is compensated with a harmless fenced no-op when nothing persisted.

src/xagent/web/services/agent_management.py
AlexLiu190625: No code change, as suggested. The assertion is an internal unreachable-state invariant and matches established repository style; production and deployment configurations do not strip assertions.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. The classifier now has one public owner, `db_runtime.is_process_control_exception`, used by both task settlement and runtime-key delivery.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. Both retry-loop closures bind `candidate` as a default argument, so later refactoring cannot observe a subsequent iteration value.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. The service receipt is cleared at the start of create, and the shared runtime-key transaction helper clears its receipt before any staging callback runs.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. The fresh compensation Session no longer has an explicit catch/rollback/re-raise block; context-managed close handles an uncommitted transaction.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed through 8b19f095. `KeyRotationConflict` now documents that the conflicting transaction was rolled back before delivery, while `RuntimeKeyDeliveryError` documents the only post-receipt/commit-attempt cases. The shared commit boundary centralizes that invariant instead of repeating it across call sites.

src/xagent/web/services/agent_management.py
AlexLiu190625: Not changed. A post-commit receipt represents an active plaintext credential that was not delivered, so skipping compensation during process control would intentionally strand that key. The process-control exception still wins after the configured pool-bounded compensation attempt; introducing a separate short-timeout engine is a broader deployment policy, not a local exception-handling fix.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed through 8b19f095. `_RuntimeKeyDeliveryOutcome` is generic end to end. Rotation returns the inferred optional snapshot directly, and create uses only a non-null invariant assertion before returning the inferred snapshot.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 8575cee6. The unreachable `service is None` clause was removed from the inner handler, where service construction is already guaranteed.

src/xagent/web/services/api_keys.py
AlexLiu190625: Fixed through 8b19f095. `AgentApiKeyService.complete_runtime_key_delivery` now owns receipt reset, staging, commit ambiguity, conflict translation, and post-commit response construction for both create and V1 rotate. The method docstrings distinguish ordinary legacy rotation from receipt-bearing V1 delivery.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed through 8b19f095. `create_agent_with_optional_key` now documents when `RuntimeKeyDeliveryError` is raised, why the receipt exists, and that the runtime boundary is its sole consumer and compensation owner.

tests/web/services/test_db_runtime.py
AlexLiu190625: Fixed in 8575cee6. The two runtime-delivery tests were moved out of `test_db_runtime.py` into the existing V1 agent-management runtime test module; the database-runtime module now tests only its own primitives.

src/xagent/web/services/agent_management.py
```diff
@@ -1033,25 +1188,164 @@ def _rotate_agent_runtime_key_with_retry_sync(
...
+        receipt: RuntimeKeyReceipt,
+    ) -> _RuntimeKeyCompensationResult:
+        """Fence a failed transition and restore only its replaced key rows."""
+
+        SessionLocal = get_session_local()
+        with SessionLocal() as db:
+            now = datetime.now(timezone.utc)
+            revoke_result = cast(
+                "CursorResult[Any]",
+                db.execute(
```
rogercloud: [P1] Serialize compensation with rotations

Compensation updates key rows without acquiring the same `Agent ... FOR UPDATE` lock used by rotation. A later rotation can snapshot K1, compensation can revoke K1 and restore K0, and the later rotation's conditional revoke can then affect zero rows before inserting K2, leaving both K0 and K2 active. Acquire the agent lock in this compensation transaction before applying the revoke/restore sequence, and cover the overlapping two-session ordering in a PostgreSQL test.

src/xagent/web/services/api_keys.py
```diff
@@ -79,8 +121,28 @@ def stage_rotated_key(
...
-        and refreshes.
+        and refreshes. ``runtime_key_receipt`` is replaced with the exact
+        staged transition after the flush succeeds.
         """
+        self.runtime_key_receipt = None
+        # Serialize legacy single-key rotations for this agent. Multi-key
+        # additions that commit after this snapshot are later state and must
+        # not be revoked or restored by this transition.
+        self.db.execute(
+            select(Agent.id).where(Agent.id == int(agent_id)).with_for_update()
```
rogercloud: [P2] Use serialization that works on SQLite

SQLite drops `FOR UPDATE`, so two rotations can both snapshot the same replaced-key IDs. After the first commits, the second scoped revoke updates zero rows but still inserts its key, leaving both replacement keys active. This regresses the base implementation's unscoped active-key update. Use an SQLite-effective serialization or atomic compare-and-swap mechanism.

src/xagent/web/services/agent_management.py
```diff
@@ -1033,25 +1188,164 @@ def _rotate_agent_runtime_key_with_retry_sync(
...
+                    update(AgentApiKey)
+                    .where(
+                        AgentApiKey.id == receipt.key_id,
+                        AgentApiKey.agent_id == receipt.agent_id,
+                        AgentApiKey.key_prefix == receipt.key_prefix,
+                        AgentApiKey.revoked_at.is_(None),
```
rogercloud: [P2] Preserve a later pause operation

This compensation fence ignores `paused_at`. If an administrator pauses the durable replacement key before compensation runs, this predicate still matches, compensation revokes that key and restores the old credential, undoing the newer administrative state. Include `AgentApiKey.paused_at.is_(None)` so compensation yields after a later pause.

src/xagent/web/services/agent_management.py
```diff
@@ -877,6 +933,87 @@ def _list_agents_sync(*, user_id: int) -> tuple[AgentSummarySnapshot, ...]:
             payloads = AgentManagementService(db).list_agents_for_user(user_id)
             return tuple(_agent_summary_snapshot(payload) for payload in payloads)
 
+    @staticmethod
+    def _run_runtime_key_session_operation(
```
rogercloud, line 938: [P3] Let this helper wrap successful results

This helper can accept callbacks returning the raw success value and construct `_RuntimeKeyDeliveryOutcome` itself using `service.runtime_key_receipt`. Its two callers can then drop the repeated result/receipt/no-error boilerplate, removing roughly 15–20 lines without changing exception, retry, or compensation behavior.

rogercloud (changes requested):
The delivery-receipt design is coherent, but the concurrency serialization and later-state fencing need correction before this is safe to merge. Focused runtime/database tests, Ruff, and `git diff --check` passed.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 1a624acb. Compensation now acquires the same per-agent transition fence before applying the receipt CAS. The PostgreSQL regression test holds a later rotation after its active-key snapshot, starts stale compensation in a second Session, verifies that compensation waits, then commits the rotation and asserts that only the later replacement remains active. The test passed against PostgreSQL 16.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 1a624acb. The compensation CAS now requires `paused_at IS NULL`. The regression test pauses the durable replacement before compensation and verifies a zero-row compensation result, the prior key remains revoked, and the paused replacement remains unrevoked.

src/xagent/web/services/agent_management.py
AlexLiu190625: Fixed in 1a624acb. `_run_runtime_key_session_operation` now accepts the raw success value and constructs the detached result/receipt/no-error envelope once. The create and rotate callbacks return only their snapshots, and a focused test pins receipt capture at the Session boundary.

src/xagent/web/services/api_keys.py
AlexLiu190625: Fixed in 1a624acb. The shared transition fence keeps `SELECT ... FOR UPDATE` on PostgreSQL/MySQL and uses a no-op parent-row write on SQLite before the active-key snapshot. The SQLite two-Session regression test verifies that the second transition blocks until the first commits and that the fence does not change `Agent.updated_at`.

tests/web/services/test_runtime_key_transition_postgres.py
```diff
@@ -0,0 +1,160 @@
...
+)
+
+
+@pytest.fixture()
+def postgres_runtime_keys():
+    """Isolated runtime-key rows in a real PostgreSQL test database."""
+
+    url = os.getenv("XAGENT_TEST_POSTGRES_URL")
+    if not url:
+        pytest.skip("XAGENT_TEST_POSTGRES_URL is not set")
```
rogercloud, line 35: [P2] Run this PostgreSQL regression test in CI

`XAGENT_TEST_POSTGRES_URL` is unset in the normal `tests/web` job, so this fixture skips there (the current CI log confirms it). The only workflow that sets the variable invokes only `test_gmail_provisioning.py -k contend`, and its change detector excludes this test and the affected runtime-key services. As written, this concurrency regression can break without any CI failure. Please add this file to a PostgreSQL-backed CI job and ensure changes to this test, `api_keys.py`, and `agent_management.py` trigger that job.

rogercloud (changes requested):
Re-review of the latest push: the four findings from the previous review are resolved, and the revised transition/compensation design now looks sound. Focused tests, Ruff, and `git diff --check` pass; however, the PostgreSQL-specific concurrency guarantee is still not enforced by CI.

tests/web/services/test_runtime_key_transition_postgres.py
AlexLiu190625: Fixed in 831296de. The PostgreSQL workflow detector now includes agent_management.py, api_keys.py, and test_runtime_key_transition_postgres.py, and the existing PostgreSQL 16 job runs the regression as a dedicated step with XAGENT_TEST_POSTGRES_URL. CI run 30274353301 executed the step without skipping and completed successfully: https://github.com/xorbitsai/xagent/actions/runs/30274353301/job/90004533283

rogercloud (approved):
Re-review of 831296de: no findings. The PostgreSQL runtime-key fencing regression test is now wired into the existing PostgreSQL CI job, relevant source/test changes trigger that job, and the GitHub Actions log confirms the test executed successfully. All current checks pass.

PR #934, opened 2026-07-20
feat: add SEO Brief Writer, Research & Enricher, KB Writer and Email Lead Response templates

yiboyasss:
## Summary

Adds four new built-in agent templates and updates one existing template, based on the latest template expansion spec. Also archives the old Inbound Agent, which is superseded by the new Email Lead Response Agent.

## New templates

| Template | Category | Connections |
|---|---|---|
| SEO Brief Writer | Marketing | Google Drive, Google Docs |
| Research & Enricher | Sales | HubSpot, Google Docs |
| KB Writer | Support | Google Drive, OneDrive, Google Docs |
| Email Lead Response Agent | Sales | HubSpot, Google Calendar, Gmail, Outlook |

All templates include bilingual (en/zh) descriptions and features, follow the existing YAML schema, and use `execution_mode: balanced`.

## Updated

- **Email Agent (Sales)** (`sales-email-agent.yaml`): rewritten per the new spec — prospect list can now come from an uploaded Excel/CSV or HubSpot CRM; drafts are grounded in CRM conversation history; after approval, sends via Gmail/Outlook and logs back to HubSpot (or updates the uploaded file). Connections expanded from Gmail to HubSpot + Gmail + Outlook.

## Archived

- **Inbound Agent** (`sales-inbound-agent.yaml`): moved from `built_in/` to `archived/` — its inbound qualification and booking flow is superseded by the new Email Lead Response Agent.

## Notes

- Templates that reference web search include `web_search` in `tool_categories` (required by `test_builtin_templates_that_use_web_search_select_web_search_category`).
- HubSpot CRM and Google Docs connectors are not shipped yet; the agent instructions include graceful fallbacks (e.g. deliver the brief as a file and note that CRM enrichment is unavailable). Templates need no further changes once the connectors land.

## Test plan

- `pytest tests/templates/ tests/web/api/test_templates_api.py` — 25 passed
- Verified via `TemplateManager` that all templates load, zh localization resolves, and connections auto-append `mcp:*` tool categories

Issue #802, opened 2026-07-08
feat(agent): Hide "Agent Invocation" tool from agent configuration


yiboyasss:
## Description

Agent Invocation tool overlaps with Workforce functionality. Allowing this tool in agent configuration may cause unnecessary agent calls within the workspace of the user; this should be configured in Workforce instead.

## Expected result

Agent Invocation is similar to Workforce. The user should be able to see past runs/conversations with workforce instead of exposing this as a generic agent tool.

## Actual result

Currently, the user can select this tool and have access to ALL of the agents created in the account.

PR #1023, opened 2026-07-27
feat: add Meeting Agent, GA Analyzer, Ads Recommendation and DevOps templates

yiboyasss:
## Summary
- Add 4 built-in agent templates to `src/xagent/templates/built_in/`: Meeting Agent (Sales), Google Analytics Analyzer (Marketing), Google Ads Recommendation (Marketing), DevOps AI Agent (Operations)
- Add a new "Operations" template category, with English/Chinese labels wired into the templates gallery
- Growth Marketing Workforce (multi-agent orchestration from the spec) is intentionally out of scope — no workforce-template mechanism exists yet, only the manual wizard

## Test plan
- [x] `pytest tests/templates/test_manager.py tests/web/api/test_templates_api.py` — 25/25 pass
- [x] `tsc --noEmit` on frontend — no errors
- [x] Verified all 4 templates load via `TemplateManager` with expected `mcp:*` tool-category merging
- [x] Visual check of the templates gallery in the browser (categories render, cards display correctly)

### DIFF
--- a/frontend/src/i18n/locales/zh.ts
+++ b/frontend/src/i18n/locales/zh.ts
@@ -1539,7 +1539,8 @@ const zh = {
       general_productivity: "通用与效率",
       customer_service: "客户服务",
       finance_lms_ops: "金融、LMS 与运营",
-      security: "安全"
+      security: "安全",
+      operations: "运维"
     },
     sections: {
       knowledge: "知识与研究",
--- a/src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
+++ b/src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
@@ -29,7 +29,7 @@ agent_config:
     # Responsibilities
 
     1. **Pull Ads data** — Fetch spend, CTR, CPA, conversions and search terms from the connected Google Ads account (read-only).
-    2. **Join GA signals** — If a 'Signals for Ads' table is available (from a Google Analytics Analyzer agent), join it by campaign / UTM to see on-site CVR, conversions and value per campaign.
+    2. **Join GA signals** — If the user has provided a 'Signals for Ads' table (pasted from a Google Analytics Analyzer agent, or passed along via a manually assembled Workforce — there is no automatic agent-to-agent handoff today), join it by campaign / UTM to see on-site CVR, conversions and value per campaign.
     3. **Prioritize** — Rank opportunities by estimated revenue impact, not just by ad-platform metrics alone.
     4. **Recommend** — Propose budget shifts, pauses, negative keywords, and keyword / copy changes, each with a clear rationale.
 
--- a/src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
+++ b/src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
@@ -12,7 +12,7 @@ features:
   - Ranks and recommends budget, keyword, negative-keyword and copy changes for you to approve
   zh:
   - 从 Google Ads 拉取花费、点击率、单次转化成本、转化数和搜索词数据
-  - 通过广告系列 / UTM 匹配，结合 Google Analytics Analyzer agent 输出的网站端"广告信号"
+  - 通过广告系列 / UTM 匹配，结合 Google Analytics Analyzer agent 输出的网站端「广告信号」
   - 按收入影响排序，推荐预算、关键词、否定关键词与文案调整，等待您审批
 connections:
 - name: Google Ads
--- a/src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
+++ b/src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
@@ -1,0 +1,57 @@
+id: marketing-google-ads-recommendation
+name: Google Ads Recommendation
+category: Marketing
+featured: false
+descriptions:
+  en: Reviews Google Ads performance, joins it with on-site conversion signals, and recommends budget, keyword and copy fixes ranked by revenue impact — recommend-and-approve, never auto-applied.
+  zh: 分析 Google Ads 投放表现，结合网站端转化信号，按收入影响排序推荐预算、关键词与文案优化方案——仅推荐待审批，绝不自动执行。
+features:
+  en:
+  - Pulls spend, CTR, CPA, conversions and search terms from Google Ads
+  - Joins ad metrics with on-site 'Signals for Ads' from a Google Analytics Analyzer agent, matched by campaign / UTM
+  - Ranks and recommends budget, keyword, negative-keyword and copy changes for you to approve
+  zh:
+  - 从 Google Ads 拉取花费、点击率、单次转化成本、转化数和搜索词数据
+  - 通过广告系列 / UTM 匹配，结合 Google Analytics Analyzer agent 输出的网站端"广告信号"
+  - 按收入影响排序，推荐预算、关键词、否定关键词与文案调整，等待您审批
+connections:
+- name: Google Ads
+  logo: https://www.google.com/s2/favicons?domain=ads.google.com&sz=64
+setup_time:
+  en: 5 min setup
+  zh: 5分钟配置
+agent_config:
+  instructions: |
+    # Role & Purpose
+
+    You are the Google Ads Recommendation agent. You recommend — you never auto-apply — ad optimizations. You combine ad-platform performance with on-site conversion signals to prioritize what matters.
+
+    # Responsibilities
+
+    1. **Pull Ads data** — Fetch spend, CTR, CPA, conversions and search terms from the connected Google Ads account (read-only).
+    2. **Join GA signals** — If a 'Signals for Ads' table is available (from a Google Analytics Analyzer agent), join it by campaign / UTM to see on-site CVR, conversions and value per campaign.
+    3. **Prioritize** — Rank opportunities by estimated revenue impact, not just by ad-platform metrics alone.
+    4. **Recommend** — Propose budget shifts, pauses, negative keywords, and keyword / copy changes, each with a clear rationale.
+
+    # Operating Guidelines
+
+    - **Recommend-and-approve only:** Never change spend, budgets, keywords, or ad copy automatically. Every change is a proposal awaiting explicit user approval.
+    - **Missing GA signals:** If no 'Signals for Ads' data is available, proceed using Ads data alone and explicitly flag that the recommendation is ads-only (no on-site conversion context).
+    - **Join carefully:** Match GA signals to Ads campaigns by campaign name or UTM parameter; if a match is ambiguous, say so rather than guessing.
+    - **Justify every recommendation:** State the expected impact (e.g. estimated CPA change, revenue delta) and your confidence level.
+
+    # Tone & Personality
+
+    Direct and numbers-first. Lead with the highest-impact recommendation. No hedging without reason — state confidence levels plainly instead.
+
+    # Reminders
+
+    - Output format: action · campaign · rationale · expected impact · confidence, ranked highest-impact first.
+    - Never touch spend or settings — you are recommend-only, always.
+    - Flag explicitly whenever a recommendation is based on Ads data alone (no GA join available).
+  execution_mode: balanced
+  skills: []
+  tool_categories:
+  - basic
+  - file
+  - knowledge
--- a/src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
+++ b/src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
@@ -3,17 +3,17 @@ name: Google Analytics Analyzer
 category: Marketing
 featured: false
 descriptions:
-  en: Connects to GA4, explains what changed in your traffic and conversions and why, then emits structured performance signals your Ads agent can act on.
-  zh: 连接 GA4，用通俗易懂的语言解释流量与转化的变化及其原因，并输出结构化的表现信号供广告投放 agent 使用。
+  en: Analyzes a GA4 export you upload or paste, explains what changed in your traffic and conversions and why, then emits a structured 'Signals for Ads' table you can hand to a Google Ads Recommendation agent.
+  zh: 分析您上传或粘贴的 GA4 导出数据，用通俗易懂的语言解释流量与转化的变化及其原因，并输出结构化的「广告信号」表格，可交由 Google Ads Recommendation agent 使用。
 features:
   en:
-  - Pulls GA4 metrics for any date range and property on request
+  - Reads a GA4 export you upload (CSV) or paste as a table — no live GA4 connection required
   - Compares against the prior period and flags anomalies by channel, campaign, landing page and segment
-  - Emits a machine-readable 'Signals for Ads' block that a Google Ads agent can consume
+  - Emits a 'Signals for Ads' table for a companion Ads agent, and can save the write-up to Google Docs or Slides on request
   zh:
-  - 按需拉取任意日期范围和媒体资源的 GA4 指标
+  - 读取您上传的 GA4 导出文件（CSV）或粘贴的表格——无需接入实时 GA4 连接
   - 与上一周期对比，按渠道、广告系列、落地页和细分维度标记异常
-  - 输出可供 Google Ads agent 消费的结构化「广告信号」数据块
+  - 输出供 Ads agent 使用的「广告信号」表格，并可按需将报告保存到 Google Docs 或 Slides
 connections:
 - name: Google Docs
   logo: https://www.google.com/s2/favicons?domain=docs.google.com&sz=64
@@ -26,21 +26,21 @@ agent_config:
   instructions: |
     # Role & Purpose
 
-    You are the Google Analytics Analyzer. You explain what changed in GA4 and why, in plain language, and you produce structured signals a companion Google Ads Recommendation agent can use — the user is responsible for carrying that output across, there is no automatic handoff between agents today.
+    You are the Google Analytics Analyzer. You explain what changed in a GA4 export and why, in plain language, and you produce structured signals a companion Google Ads Recommendation agent can use. You have no live connection to GA4 — you work from a CSV export or pasted table the user provides.
 
     # Responsibilities
 
-    1. **Pull metrics** — Given a GA4 property and date range, fetch traffic, engagement and conversion metrics.
-    2. **Compare & flag** — Compare against the prior period of equal length; identify anomalies.
+    1. **Get the data** — If the user hasn't already uploaded or pasted a GA4 export, ask for one covering the period to analyze, plus a prior period of equal length for comparison.
+    2. **Compare & flag** — Compare the current period against the prior one; identify anomalies.
     3. **Break down** — Quantify performance by channel, campaign, landing page and segment to find the drivers of any change.
-    4. **Report & hand off** — Explain the findings in plain language, then emit a 'Signals for Ads' table for handoff.
+    4. **Report** — Explain the findings in plain language, then emit a 'Signals for Ads' table. On request, save the write-up to a Google Doc, or the table to a Slide deck.
 
     # Operating Guidelines
 
-    - **Informational only:** You never change campaign settings, budgets, or site configuration — you analyze and report.
+    - **Work only from provided data:** You have no live GA4 connection — every number must come from the export or table the user gave you. If answering a question needs data that wasn't provided, say so and ask for it rather than estimating.
     - **Always cite the metric and date range** used for any claim; never state a percentage change without the underlying numbers.
-    - **Note caveats:** Flag sampling, attribution model, or data-freshness caveats when they could materially affect the conclusion.
-    - **Signals contract:** When emitting the handoff table, use exactly these columns: campaign / source · sessions · CVR · conversions · value · verdict (winning / underperforming) — do not change the column names. This table is designed to be read by a companion Google Ads Recommendation agent, joined by campaign / UTM, but there is no automatic agent-to-agent handoff today: pass it along yourself (paste it into that agent's chat, or assemble both agents into a Workforce).
+    - **Note caveats:** Flag sampling, attribution model, or data-freshness caveats the export itself notes.
+    - **Signals contract:** When emitting the handoff table, use exactly these columns: campaign / source · sessions · CVR · conversions · value · verdict (winning / underperforming) — do not change the column names. This table is designed to be read by a companion Google Ads Recommendation agent; there is no automatic agent-to-agent handoff today, so pass it along yourself (paste it into that agent's chat, or assemble both agents into a Workforce).
 
     # Tone & Personality
 
@@ -48,8 +48,8 @@ agent_config:
 
     # Reminders
 
-    - Always compare to a like-for-like prior period, not an arbitrary earlier date.
-    - Every insight must be traceable to a specific metric and date range — no vague claims.
+    - Never estimate a metric that wasn't in the provided data — ask for it instead.
+    - Every insight must be traceable to a specific metric and date range in the source export — no vague claims.
     - The 'Signals for Ads' table's structure is a contract a companion Ads agent depends on — keep it exact even if the surrounding commentary changes. You are not automatically connected to that agent; the user carries the table across.
   execution_mode: balanced
   skills: []
--- a/src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
+++ b/src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
@@ -1,0 +1,59 @@
+id: marketing-google-analytics-analyzer
+name: Google Analytics Analyzer
+category: Marketing
+featured: false
+descriptions:
+  en: Connects to GA4, explains what changed in your traffic and conversions and why, then emits structured performance signals your Ads agent can act on.
+  zh: 连接 GA4，用通俗易懂的语言解释流量与转化的变化及其原因，并输出结构化的表现信号供广告投放 agent 使用。
+features:
+  en:
+  - Pulls GA4 metrics for any date range and property on request
+  - Compares against the prior period and flags anomalies by channel, campaign, landing page and segment
+  - Emits a machine-readable 'Signals for Ads' block that a Google Ads agent can consume
+  zh:
+  - 按需拉取任意日期范围和媒体资源的 GA4 指标
+  - 与上一周期对比，按渠道、广告系列、落地页和细分维度标记异常
+  - 输出可供 Google Ads agent 消费的结构化"广告信号"数据块
+connections:
+- name: Google Analytics
+  logo: https://www.google.com/s2/favicons?domain=analytics.google.com&sz=64
+- name: Google Drive
+  logo: https://www.google.com/s2/favicons?domain=drive.google.com&sz=64
+setup_time:
+  en: 5 min setup
+  zh: 5分钟配置
+agent_config:
+  instructions: |
+    # Role & Purpose
+
+    You are the Google Analytics Analyzer. You explain what changed in GA4 and why, in plain language, and you feed a downstream Google Ads Recommendation agent with structured signals.
+
+    # Responsibilities
+
+    1. **Pull metrics** — Given a GA4 property and date range, fetch traffic, engagement and conversion metrics.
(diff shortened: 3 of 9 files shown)


src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -0,0 +1,61 @@
...
+  en:
+  - Wakes on an alert / webhook, a schedule, or a direct chat request
+  - Investigates via CloudWatch, DynamoDB and SQS, plus read-only SSH diagnostics
+  - Triages against a known-noise catalog and posts a root-cause hypothesis with a recommended fix to Slack
+  zh:
+  - 由告警 / webhook、定时任务或聊天请求触发
+  - 通过 CloudWatch、DynamoDB、SQS 及只读 SSH 诊断进行排查
+  - 对照已知噪声名录进行分诊，并将根因假设与修复建议发布到 Slack
+connections:
+- name: AWS
```
rogercloud: Neither `AWS` (line 18) nor `Slack` (line 20) is present in `get_builtin_public_mcp_app_rows` in `src/xagent/web/builtin_mcp_registry.py`. `TemplateManager._enrich_template` (`src/xagent/templates/manager.py:126-152`) turns each connection into a real `mcp:{name}` tool grant with no validation against the registry, and the build wizard's Configure step requires each app to resolve to a connected server or a registered official app. With both connectors unregistered, this template's Configure step cannot be completed via the built-in catalog — the user sees an "app not supported" badge with no connect affordance. (A manually created custom MCP server named exactly `AWS`/`Slack` does satisfy the check, but nothing here says so.) PR #934 hit the same class of issue and fixed it by removing the unbacked entries. Suggest either registering these connectors in `builtin_mcp_registry.py`, or dropping the `connections:` entries and describing the AWS/Slack integration in the instructions instead.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - Slack message structure: incident summary · evidence · root-cause hypothesis · recommended fix · severity.
+    - Never apply a fix yourself — always end with a recommendation awaiting approval.
+    - If SSH, CloudWatch, DynamoDB, or SQS access fails, report the access failure itself rather than silently skipping that data source.
+  execution_mode: balanced
+  skills: []
+  tool_categories:
+  - basic
+  - file
+  - knowledge
+  - ssh
```
rogercloud: The `ssh` category bundles `ssh_execute` (arbitrary non-interactive command), `ssh_upload` and `ssh_download` together — there is no read-only subset (`src/xagent/core/tools/adapters/vibe/ssh_tools.py`). So the "read-only by default" posture stated at lines 6, 11, 34 and 40 is enforced by the prompt alone. The platform does have real per-target scoping — `capabilities` is checked in `src/xagent/core/ssh/executor.py:103,315` — so please add setup guidance telling operators to bind the SSH target with `capabilities={"execute"}` only, excluding upload/download. Separately, worth being explicit in the prompt that the Slack approval step at line 40 is an agent-behavior convention, not a platform gate: `approval_policy` in `src/xagent/core/ssh/types.py` is declared but never read anywhere in the codebase. This is the first built-in template to use `ssh`, so whatever pattern lands here becomes the precedent.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,59 @@
...
+    2. **Compare & flag** — Compare against the prior period of equal length; identify anomalies.
+    3. **Break down** — Quantify performance by channel, campaign, landing page and segment to find the drivers of any change.
+    4. **Report & hand off** — Explain the findings in plain language, then emit a 'Signals for Ads' table for handoff.
+
+    # Operating Guidelines
+
+    - **Informational only:** You never change campaign settings, budgets, or site configuration — you analyze and report.
+    - **Always cite the metric and date range** used for any claim; never state a percentage change without the underlying numbers.
+    - **Note caveats:** Flag sampling, attribution model, or data-freshness caveats when they could materially affect the conclusion.
+    - **Signals contract:** When emitting the handoff table, use exactly these columns: campaign / source · sessions · CVR · conversions · value · verdict (winning / underperforming). This is consumed by the Google Ads Recommendation agent, joined by campaign / UTM — do not change the column names.
```
rogercloud: This describes a machine-to-machine contract ("consumed by the Google Ads Recommendation agent... do not change the column names", echoed at lines 12, 36 and 53) that cannot exist in a template today. This template's `agent_config` grants only `basic`/`file`/`knowledge` with no skills, and no subagent or handoff mechanism is expressible in the template schema at all — `AGENT_CONFIG_UNASSIGNABLE_CATEGORIES` (`src/xagent/core/tools/adapters/vibe/base.py:39-45`) explicitly excludes the `agent` category from template-grantable categories, noting multi-agent orchestration is a separate Workforce concern (#802). Suggest rewording to a user-mediated handoff — e.g. "emit this table so it can be pasted into the Google Ads Recommendation agent" — and dropping "a contract other agents rely on" until a real workforce mechanism ships.

src/xagent/templates/built_in/marketing-google-ads-recommendation.yaml
```diff
@@ -0,0 +1,57 @@
...
+agent_config:
+  instructions: |
+    # Role & Purpose
+
+    You are the Google Ads Recommendation agent. You recommend — you never auto-apply — ad optimizations. You combine ad-platform performance with on-site conversion signals to prioritize what matters.
+
+    # Responsibilities
+
+    1. **Pull Ads data** — Fetch spend, CTR, CPA, conversions and search terms from the connected Google Ads account (read-only).
+    2. **Join GA signals** — If a 'Signals for Ads' table is available (from a Google Analytics Analyzer agent), join it by campaign / UTM to see on-site CVR, conversions and value per campaign.
```
rogercloud: Same as the GA Analyzer side of this pair: "from a Google Analytics Analyzer agent" (also lines 11 and 39) implies an automatic agent-to-agent handoff that the template schema cannot express. To this template's credit line 39 degrades gracefully when no signals are present, so nothing breaks — but the phrasing promises a capability that does not exist. Suggest rewording to describe the user pasting the GA agent's output table into the chat.

src/xagent/templates/built_in/sales-meeting-agent.yaml
```diff
@@ -0,0 +1,64 @@
...
+  en:
+  - Connects to Granola (cloud) or Hermes (local) MCP to pull transcripts, summaries and action items
+  - Extracts decisions, owners and due dates from any meeting on request
+  - Drafts a follow-up email and can create tasks / CRM notes after your approval
+  zh:
+  - 通过 Granola（云端）或 Hermes（本地）MCP 连接，拉取会议转录、摘要和行动项
+  - 按需从任意会议中提取决策、负责人和截止日期
+  - 起草跟进邮件，并在您批准后创建任务或 CRM 记录
+connections:
+- name: Granola
```
rogercloud: `Granola` (line 18) and `Hermes` (line 20) are not in `get_builtin_public_mcp_app_rows` (`src/xagent/web/builtin_mcp_registry.py`), so their synthesized `mcp:*` grants can never resolve in the build wizard's Configure step. Note `HubSpot` (line 22) and `Gmail` (line 24) in this same file *are* registered — so 2 of 4 connections here are broken, and the template's core capability (reading meeting notes) is the part that does not work. See the AWS/Slack comment on `operations-devops-ai-agent.yaml` for the mechanism and the PR #934 precedent. Suggest either registering these connectors or removing the entries and describing the Granola/Hermes MCP setup in prose.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,59 @@
...
+  en:
+  - Pulls GA4 metrics for any date range and property on request
+  - Compares against the prior period and flags anomalies by channel, campaign, landing page and segment
+  - Emits a machine-readable 'Signals for Ads' block that a Google Ads agent can consume
+  zh:
+  - 按需拉取任意日期范围和媒体资源的 GA4 指标
+  - 与上一周期对比，按渠道、广告系列、落地页和细分维度标记异常
+  - 输出可供 Google Ads agent 消费的结构化"广告信号"数据块
+connections:
+- name: Google Analytics
```
rogercloud: `Google Analytics` is not registered in `get_builtin_public_mcp_app_rows` (`src/xagent/web/builtin_mcp_registry.py`), so this connection produces an `mcp:Google Analytics` grant the wizard can never satisfy — blocking the Configure step for the template's primary data source. (`Google Drive` on line 20 is registered.) Same mechanism and PR #934 precedent as the other templates in this PR.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,59 @@
...
+  logo: https://www.google.com/s2/favicons?domain=analytics.google.com&sz=64
+- name: Google Drive
```
rogercloud: The `Google Drive` connection is never used — the instructions (lines 27-53) contain no mention of Drive, export, files or documents. This costs a live `mcp:Google Drive` tool grant plus an extra "must connect" requirement in the setup card for no benefit. Compare `marketing-seo-brief-writer.yaml`, which justifies each of its Drive/Docs connections in-prompt (lines 37, 39, 47). Suggest dropping the connection, or adding an instruction step that actually exports the report to Drive.

src/xagent/templates/built_in/sales-meeting-agent.yaml
```diff
@@ -0,0 +1,64 @@
...
+  logo: https://www.google.com/s2/favicons?domain=granola.ai&sz=64
+- name: Hermes
+  logo: https://www.google.com/s2/favicons?domain=github.com&sz=64
```
rogercloud: The Hermes logo points at `domain=github.com`, so the card renders the GitHub mark rather than a Hermes logo. Every other connection across all built-in templates points at its own vendor domain — this is the only mismatch in the repo. Suggest pointing at the real project domain, or omitting `logo` entirely to fall back to the letter avatar.

frontend/src/i18n/locales/zh.ts
```diff
@@ -1539,7 +1539,8 @@ const zh = {
       general_productivity: "通用与效率",
       customer_service: "客户服务",
       finance_lms_ops: "金融、LMS 与运营",
-      security: "安全"
+      security: "安全",
+      operations: "运维"
```
rogercloud, line 1543: Minor consistency note, no change needed now: `operations: "运维"` uses the IT/infra sense of "ops", while the sibling `finance_lms_ops: "金融、LMS 与运营"` on line 1541 uses the business sense for its "Ops". Both are idiomatic, and 运维 is the better fit while DevOps is the category's only member. Worth revisiting if a non-IT operations template later lands in this category.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,59 @@
...
+  zh: 连接 GA4，用通俗易懂的语言解释流量与转化的变化及其原因，并输出结构化的表现信号供广告投放 agent 使用。
+features:
+  en:
+  - Pulls GA4 metrics for any date range and property on request
+  - Compares against the prior period and flags anomalies by channel, campaign, landing page and segment
+  - Emits a machine-readable 'Signals for Ads' block that a Google Ads agent can consume
+  zh:
+  - 按需拉取任意日期范围和媒体资源的 GA4 指标
+  - 与上一周期对比，按渠道、广告系列、落地页和细分维度标记异常
+  - 输出可供 Google Ads agent 消费的结构化"广告信号"数据块
```
rogercloud: The Chinese text wraps 广告信号 in ASCII double quotes; full-width quotes (“ ” or 「」) would be more conventional in CJK copy. Same on `marketing-google-ads-recommendation.yaml:15`. Note there is no existing convention to follow — this is the only place any built-in template's Chinese text uses quotation marks at all. Cosmetic; does not affect YAML parsing.

rogercloud (commented):
## Summary

This PR adds four new built-in agent templates as declarative YAML (`Meeting Agent` / Sales, `Google Analytics Analyzer` and `Google Ads Recommendation` / Marketing, `DevOps AI Agent` / Operations) and introduces a new **Operations** category by adding one key to `CATEGORY_LABEL_KEYS` in the templates page plus matching `en`/`zh` locale strings. No backend or schema changes were required — category rendering is already dynamic with a graceful fallback for unknown categories, so the frontend change is purely a nicer display label.

## Approach: Acceptable with reservations

The approach is right. Declarative YAML files that match the existing built-in schema exactly, plus a single hand-added label key, is the smallest change that does the job and it follows established convention precisely. Nothing here needed new abstraction.

The reservation is about `connections:`. That block is **not decorative metadata** — `TemplateManager._enrich_template` (`src/xagent/templates/manager.py:126-152`) synthesizes each entry directly into a real `mcp:{name}` tool-category grant, and the build wizard's Configure step requires each named app to resolve to a connected server or a registered official app. Three of the four new templates name connectors that do not exist in `src/xagent/web/builtin_mcp_registry.py` (`get_builtin_public_mcp_app_rows`, lines 118-393): AWS and Slack (DevOps), Granola and Hermes (Meeting), Google Analytics (GA Analyzer). Only `Google Ads` is fully clean. For those apps the wizard renders a "platform does not support this app yet" state with no connect affordance, so the Configure step cannot be completed through the built-in catalog. This is the same defect class that PR #934 had to correct in the opposite direction (removing `connections:` entries that pointed at nothing) — worth citing as precedent for the fix.

Two smaller design-level points that span more than one line:

- **The DevOps template's "read-only" posture is prompt-only.** It grants the `ssh` category, which bundles `ssh_execute`, `ssh_upload` and `ssh_download` with no partial grant. The platform does have real enforcement — per-target `SshCapability` is checked in `src/xagent/core/ssh/executor.py:103,315` — but the template never tells the operator to use it. Worth knowing: `approval_policy` in `src/xagent/core/ssh/types.py` is declared but never read anywhere in the codebase, so the Slack-approval step this template describes is a convention the agent is asked to follow, not a gate the platform enforces. That should be stated plainly rather than implied.
- **The two Marketing templates encode an agent-to-agent handoff that cannot exist today.** The GA Analyzer describes a "contract other agents rely on" consumed by the Ads agent; neither template has (or can have) any subagent mechanism — `AGENT_CONFIG_UNASSIGNABLE_CATEGORIES` (`src/xagent/core/tools/adapters/vibe/base.py:39-45`) explicitly excludes the `agent` category from template-grantable categories, with a comment noting multi-agent orchestration is a separate Workforce concern (#802). The Ads template does degrade gracefully when signals are absent, so this is a misleading promise rather than a hard failure.

Two informational notes, neither actionable: the template history now contains both the new active `Operations` category and a pre-existing archived `Finance, LMS & Ops` category (not user-facing, so not a conflict); and `CATEGORY_LABEL_KEYS` remains hand-maintained rather than derived — this PR follows that existing pattern correctly, it just keeps needing manual upkeep.

Two findings live in files this PR does not touch, so they are recorded here rather than as inline comments:

- `src/xagent/web/api/tools.py:41-56` — `CATEGORY_DISPLAY_NAMES` has no `"ssh"` key, so the fallback at line 216 renders **"Ssh"** in the tools UI. This is a pre-existing gap, but the DevOps template is the first built-in to select `ssh`, so this PR makes it user-visible for the first time. One-line fix: `"ssh": "SSH"` (and `"database": "Database"` while you are there).
- `frontend/src/components/templates/library-template-card.tsx` — `PILL_KNOWN` has no `operations` entry, so the pill color falls through to the hash-based palette; `"operations"` hashes to the same index as `"productivity"`. Deterministic and readable, purely a cosmetic collision. Optional: add an explicit palette entry.

Lean already.

## Tests

`pytest tests/templates/test_manager.py tests/web/api/test_templates_api.py` passes **25/25** locally — independently reproduced, matching the PR's own claim. No regressions.

Coverage gaps worth following up on (recommendations, not merge blockers):

1. No test loads the real `built_in/` directory through `TemplateManager._parse_yaml_file`'s actual validation path — both suites use synthetic YAML in temp dirs. A new built-in missing a required field would be silently skipped at runtime with no test failure.
2. Nothing asserts the new template ids or the `Operations` category exist after load.
3. Nothing cross-checks `connections[].name` against the registered MCP catalog — which is exactly the invariant finding #1 violates, and exactly what a PR #934-style regression test would have caught permanently.
4. There is no allow-list validation of `tool_categories` in the write path (bare `List[str]`/JSON), so a typo'd category is a silent zero-tool no-op end to end.

A cheap smoke test would close 1-3: instantiate `TemplateManager` against the real `built_in/` directory, assert `len(list_templates()) == len(glob("*.yaml"))`, and add a parametrized connections-vs-catalog consistency check.

## Recommendation

**Merge after fixing finding #1; the other Mediums can be fast-follow.**

Findings #2 and #3 are documentation-quality problems — the templates over-promise, but they still function. Finding #1 is different in kind: three of the four templates cannot complete the build wizard's Configure step through the built-in catalog, which means most of what this PR ships does not work end to end for a user who follows the happy path. There is a workaround (manually creating a custom MCP server whose name normalizes to match, e.g. one literally named "AWS"), which is why this is Medium rather than High, but nothing in the templates or setup copy mentions it.

The cheap fix is the one PR #934 used: drop the unregistered `connections:` entries and describe those integrations in prose. Registering the connectors properly in `builtin_mcp_registry.py` is the better long-term answer but is clearly out of scope for a templates PR. Either way, shipping templates whose Configure step dead-ends is the one thing I would not defer.

src/xagent/templates/built_in/sales-meeting-agent.yaml
```diff
@@ -0,0 +1,60 @@
+id: sales-meeting-agent
+name: Meeting Agent
+category: Sales
+featured: false
+descriptions:
+  en: Reads your meeting notes via MCP (Granola or Hermes) and turns them into summaries, decisions, action items and follow-up drafts — grounded in the transcript, nothing invented.
```
rogercloud: [P1] Configure the transcript source the Meeting Agent depends on

This template promises Granola/Hermes transcript retrieval, but the instantiated agent receives only `mcp:HubSpot` and `mcp:Gmail`. It therefore cannot perform its defining meeting-summary workflow. Add a supported transcript connector/tool, or change the template to accept pasted/uploaded notes and remove the retrieval claim.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,59 @@
...
+  en:
+  - Pulls GA4 metrics for any date range and property on request
+  - Compares against the prior period and flags anomalies by channel, campaign, landing page and segment
+  - Emits a machine-readable 'Signals for Ads' block that a Google Ads agent can consume
+  zh:
+  - 按需拉取任意日期范围和媒体资源的 GA4 指标
+  - 与上一周期对比，按渠道、广告系列、落地页和细分维度标记异常
+  - 输出可供 Google Ads agent 消费的结构化「广告信号」数据块
+connections:
+- name: Google Docs
```
rogercloud, line 18: [P1] Provide a GA4 data source instead of output-only integrations

Google Docs and Slides cannot query GA4, while this template's core instructions require fetching GA4 metrics. The created agent therefore cannot perform its primary analysis. Replace these connections with a supported GA4 integration/tool, or rewrite the workflow around user-supplied reports.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -0,0 +1,58 @@
...
+features:
+  en:
+  - Wakes on an alert / webhook, a schedule, or a direct chat request
+  - Investigates via CloudWatch, DynamoDB and SQS, plus read-only SSH diagnostics
+  - Triages against a known-noise catalog and posts a root-cause hypothesis with a recommended fix to Slack
+  zh:
+  - 由告警 / webhook、定时任务或聊天请求触发
+  - 通过 CloudWatch、DynamoDB、SQS 及只读 SSH 诊断进行排查
+  - 对照已知噪声名录进行分诊，并将根因假设与修复建议发布到 Slack
+connections: []
```
rogercloud, line 17: [P1] Provision the systems this DevOps template claims to monitor

With `connections: []`, the created agent has no CloudWatch, DynamoDB, SQS, or Slack tools. SSH does not supply those integrations. Add supported AWS and Slack connectors/selectors, or narrow the template to SSH-only diagnostics and in-chat reporting.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - Slack message structure: incident summary · evidence · root-cause hypothesis · recommended fix · severity.
+    - Never apply a fix yourself — always end with a recommendation awaiting approval.
+    - If SSH, CloudWatch, DynamoDB, or SQS access fails, report the access failure itself rather than silently skipping that data source.
+  execution_mode: balanced
+  skills: []
+  tool_categories:
+  - basic
+  - file
+  - knowledge
+  - ssh
```
rogercloud: [P1] Do not advertise enforced read-only/approval safety for unrestricted SSH

Selecting the full `ssh` category can expose arbitrary command execution plus upload/download for a bound target, while `approval_policy` is not enforced by `SshExecutor`. Prompt text alone cannot guarantee read-only, approval-gated operation. Enforce restrictions and approval in code, or remove SSH and the safety promise from this template.

src/xagent/templates/built_in/sales-meeting-agent.yaml
```diff
@@ -0,0 +1,60 @@
...
+  zh: 通过 MCP（Granola 或 Hermes）读取会议记录，生成摘要、决策、行动项和跟进邮件草稿——完全基于会议记录，不臆造内容。
+features:
+  en:
+  - Connects to Granola (cloud) or Hermes (local) MCP to pull transcripts, summaries and action items
+  - Extracts decisions, owners and due dates from any meeting on request
+  - Drafts a follow-up email and can create tasks / CRM notes after your approval
```
rogercloud: [P2] Remove the unsupported task-creation claim

The configured HubSpot connector supports CRM notes but not tasks, and Gmail only provides mail operations. Add a task-capable integration or limit this feature and the related instructions to CRM notes and email drafts.

rogercloud (changes requested):
Design verdict: acceptable-with-reservations. The declarative template/category integration fits the existing architecture, but several newly added templates cannot perform their defining workflow with the capabilities they provision. Verification: 27 focused backend tests passed; frontend TypeScript check passed; `git diff --check` passed. No linked issue was provided.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -0,0 +1,63 @@
...
+  zh: 10分钟配置
+agent_config:
+  instructions: |
+    # Role & Purpose
+
+    You are a DevOps diagnostics agent. You investigate a problem on a server by running commands over an SSH connection the operator binds to you, then report a root-cause hypothesis and recommended fix directly in this chat. You have no connection to AWS, CloudWatch, DynamoDB, SQS, or Slack — only to the SSH target(s) explicitly bound to you.
+
+    # Setup Required (read before relying on this agent)
+
+    - This agent needs an SSH target bound to it before it can do anything (Settings → SSH bindings).
```
rogercloud: [P1] Gate this SSH-only template to deployments that can bind SSH targets

This template is returned by the global built-in templates API, but standard/self-hosted Xagent cannot satisfy its only prerequisite: `inTeam` remains false, so `AgentSshBindings` renders nothing, and this repository installs no default SSH provider hook, causing `create_ssh_tools()` to return no tools. Gate/hide the template when SSH target management is unavailable, implement the provider/API/UI for standard Xagent, or provide a file/log-input fallback.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+    # Tone & Personality
+
+    Analytical and plain-spoken. Lead with the "so what" — explain what changed and why before diving into numbers. Avoid analytics jargon unless the user has used it first.
+
+    # Reminders
+
+    - Never estimate a metric that wasn't in the provided data — ask for it instead.
+    - Every insight must be traceable to a specific metric and date range in the source export — no vague claims.
+    - The 'Signals for Ads' table's structure is a contract a companion Ads agent depends on — keep it exact even if the surrounding commentary changes. You are not automatically connected to that agent; the user carries the table across.
```
rogercloud: [simplification: delete] Line 43 already defines the exact table schema and the manual handoff flow. This reminder adds no new constraint and can be removed.

src/xagent/templates/built_in/sales-meeting-agent.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+    # Tone & Personality
+
+    Precise, efficient, and neutral. Summaries should be scannable — short sentences, bullet points, no filler. Follow-up email drafts should sound like the user wrote them: professional but warm, not corporate.
+
+    # Reminders
+
+    - Structure output as: ## Summary · ## Decisions · ## Action items (owner · due date) · ## Draft email.
+    - Never fabricate an owner or due date — mark it "unassigned" / "no date given" if the transcript didn't specify one.
+    - Confirm before anything is sent or logged; nothing leaves the draft stage without approval.
```
rogercloud: [simplification: delete] Line 42 already states the stronger and more precise rule requiring explicit confirmation before sending email or logging a CRM note. This reminder is redundant.

rogercloud (changes requested):
Re-review of the update since e9beb83. All five findings from the previous review are fixed. Design verdict remains acceptable-with-reservations because one deployment-capability issue remains. Verification: 27 focused backend tests passed; frontend TypeScript check passed; `git diff --check` passed.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+    # Tone & Personality
+
+    Calm, precise, incident-report style. No alarmism. Lead with impact and root cause, not raw log dumps — summarize the evidence, then quote the specific line that supports it.
+
+    # Reminders
+
+    - Report structure: summary · evidence · root-cause hypothesis · recommended fix · severity.
+    - You have no live connection to any system — never imply you fetched or ran anything yourself.
+    - If the user's material is incomplete, ask for the specific missing piece rather than filling the gap with assumptions.
```
rogercloud: [non-blocking simplification: delete] Line 36 already requires asking for missing diagnostic material instead of guessing, so this reminder can be removed without changing behavior.

src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - **Note caveats:** Flag sampling, attribution model, or data-freshness caveats the export itself notes.
+    - **Signals contract:** When emitting the handoff table, use exactly these columns: campaign / source · sessions · CVR · conversions · value · verdict (winning / underperforming) — do not change the column names. This table is designed to be read by a companion Google Ads Recommendation agent; there is no automatic agent-to-agent handoff today, so pass it along yourself (paste it into that agent's chat, or assemble both agents into a Workforce).
+
+    # Tone & Personality
+
+    Analytical and plain-spoken. Lead with the "so what" — explain what changed and why before diving into numbers. Avoid analytics jargon unless the user has used it first.
+
+    # Reminders
+
+    - Never estimate a metric that wasn't in the provided data — ask for it instead.
```
rogercloud: [non-blocking simplification: delete] Line 40 already prohibits estimating unprovided metrics and requires asking for missing data, so this reminder is redundant.

rogercloud (approved):
Re-review complete. All blocking findings from the previous reviews are fixed. Design verdict: sound. Verification: 27 focused backend tests passed; frontend TypeScript check passed; `git diff --check` passed. Two optional prompt-deduplication suggestions are included inline and are non-blocking.

PR #1027, opened 2026-07-27
fix: make tool user interactions resumable

qinxuye:
## Summary

- let any tool suspend ReAct execution by returning `status="waiting_for_user"`
- checkpoint the interaction, stop later calls in the turn, and resume from only the new user response
- deliver the response once through an optional `resume_user_interaction` tool capability
- preserve interaction capabilities and control envelopes through output-filter wrappers
- propagate execution status to tool teardown so stateful tools can preserve resources across a pause
- distinguish semantic completion outcomes (`completed`, `partial`, `blocked`) from the execution lifecycle

## Why

`ask_user_question` already pauses the ReAct control flow, but ordinary tools could not request user input safely. Their result was treated as a tool failure, wrappers could hide optional runtime capabilities, and stateful tools could not tell a pause from final teardown.

This change adds a generic interaction contract without embedding any domain-specific approval policy. Tools own the meaning of the response; the runtime only transports, checkpoints, and delivers it.

## Behavior

- serial and concurrent tool results can request a pause
- multiple concurrent requests are presented in one deterministic message
- interaction responses survive checkpoint restoration and are consumed once by the next matching capable tool
- waiting tools close tracing spans with `waiting_for_user`, not an error
- output filtering cannot discard the interaction control envelope
- final answers expose `completion_outcome` while retaining `status="completed"` as the lifecycle state

## Validation

- `python -m pytest -q tests/core/agent tests/core/tools/test_user_interaction.py tests/core/tools/adapters/vibe/test_output_filter_passthrough.py`
- `ruff check` on all changed source and test files
- `mypy` on all changed source files
- repository pre-commit hooks
- private implementation keyword audit over the added diff

### DIFF
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -1207,12 +1207,15 @@ class ReActPattern(AgentPattern):
     def _queue_tool_interaction_responses(
         self,
         *,
-        waiting_request: dict[str, Any],
+        waiting_request: Any,
         response: str,
     ) -> None:
         """Persist a user's answer for each tool interaction in the pause."""
 
-        if waiting_request.get("kind") != "tool_waiting_for_user":
+        if (
+            not isinstance(waiting_request, dict)
+            or waiting_request.get("kind") != "tool_waiting_for_user"
+        ):
             return
         raw_requests = waiting_request.get("requests")
         requests = raw_requests if isinstance(raw_requests, list) else [waiting_request]
@@ -1810,6 +1813,7 @@ class ReActPattern(AgentPattern):
             request_interactions = _normalize_ask_user_interactions(
                 result.get("interactions", [])
             )
+            deduplicated_request_interactions: list[dict[str, Any]] = []
             for interaction in request_interactions:
                 item = dict(interaction)
                 base_field = str(item.get("field") or "response")
@@ -1821,6 +1825,7 @@ class ReActPattern(AgentPattern):
                 item["field"] = field
                 used_fields.add(field)
                 interactions.append(item)
+                deduplicated_request_interactions.append(item)
 
             requests.append(
                 {
@@ -1835,7 +1840,7 @@ class ReActPattern(AgentPattern):
                         or "This tool requires user input before it can continue."
                     ),
                     "message_type": str(result.get("message_type") or "question"),
-                    "interactions": request_interactions,
+                    "interactions": deduplicated_request_interactions,
                 }
             )
 
@@ -2928,10 +2933,10 @@ class ReActPattern(AgentPattern):
         for index, pending in enumerate(self.pending_tool_interaction_responses):
             if pending.get("tool_name") != tool_name:
                 continue
-            response = self.pending_tool_interaction_responses.pop(index)
+            pending_response = self.pending_tool_interaction_responses.pop(index)
             resumed = resume(
-                interaction_id=response.get("interaction_id", ""),
-                response=response.get("response", ""),
+                interaction_id=pending_response.get("interaction_id", ""),
+                response=pending_response.get("response", ""),
             )
             if inspect.isawaitable(resumed):
                 await resumed
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -252,6 +252,7 @@ class ReActPattern(AgentPattern):
         waiting_result = await self._resume_waiting_for_user_if_needed(
             context=context,
             runtime=runtime,
+            tools=tools,
         )
         if waiting_result is not None:
             await runtime.on_pattern_end(
@@ -1135,6 +1136,7 @@ class ReActPattern(AgentPattern):
         *,
         context: Any,
         runtime: PatternRuntime,
+        tools: list[Any],
     ) -> dict[str, Any] | None:
         if self.status != "waiting_for_user" or not self.waiting_for_user_request:
             return None
@@ -1167,6 +1169,7 @@ class ReActPattern(AgentPattern):
         self._queue_tool_interaction_responses(
             waiting_request=self.waiting_for_user_request,
             response=response or "",
+            tools=tools,
         )
         waiting_task = self.waiting_for_user_request.get("task_text")
         if waiting_task and self.task_text is None:
@@ -1221,8 +1224,9 @@ class ReActPattern(AgentPattern):
         *,
         waiting_request: Any,
         response: str,
+        tools: list[Any],
     ) -> None:
-        """Persist a user's answer for each tool interaction in the pause."""
+        """Queue replies only for tools that expose the optional resume callback."""
 
         if (
             not isinstance(waiting_request, dict)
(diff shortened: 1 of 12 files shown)


src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1151,28 +1179,60 @@ def _mark_latest_user_message_as_waiting_response(
...
+    def _queue_tool_interaction_responses(
+        self,
+        *,
+        waiting_request: dict[str, Any],
+        response: str,
+    ) -> None:
+        """Persist a user's answer for each tool interaction in the pause."""
+
+        if waiting_request.get("kind") != "tool_waiting_for_user":
             return
```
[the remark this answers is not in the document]
qinxuye: Added an explicit runtime type guard for malformed or missing checkpoint payloads, and added regression coverage for `None`, list, and string values.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1716,6 +1794,97 @@ def _backfill_result(
...
+                    ),
+                    "message": str(
+                        result.get("message")
+                        or result.get("error")
+                        or "This tool requires user input before it can continue."
+                    ),
+                    "message_type": str(result.get("message_type") or "question"),
+                    "interactions": request_interactions,
+                }
+            )
```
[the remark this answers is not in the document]
qinxuye: Fixed the request payload to reuse the per-request deduplicated interaction items, so nested request fields now match the global interactions. Added a regression assertion for duplicate fields across concurrent tools.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -2683,6 +2914,29 @@ async def _execute_tool(self, tool_call: dict[str, Any], tools: list[Any]) -> An
...
+        if resume is None:
+            return
+        for index, pending in enumerate(self.pending_tool_interaction_responses):
+            if pending.get("tool_name") != tool_name:
+                continue
+            response = self.pending_tool_interaction_responses.pop(index)
+            resumed = resume(
+                interaction_id=response.get("interaction_id", ""),
+                response=response.get("response", ""),
+            )
```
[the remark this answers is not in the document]
qinxuye: Renamed the popped value to `pending_response` to remove the shadowing and make the resume call clearer.

rogercloud (changes requested):
## Review summary

This review was performed against head `156974c1`. Five actionable resume-safety findings were confirmed at that revision. They were subsequently addressed by commit `53bdf36d`; see the follow-up review on the latest head for verification.

Original validation at `156974c1`: the focused suite passed 569 tests under the locked Python 3.11 environment; Ruff passed on all 11 changed files; mypy passed on the six changed source files; and `git diff --check` passed.

qinxuye:
@rogercloud Addressed all five findings in 53bdf36d:

- Discard and close every unexecuted tool-call suffix when any serial, concurrent, or control tool pauses, so resume always starts from a fresh LLM plan.
- Deliver queued replies before replanning, directly to each saved interaction_id; delivery no longer depends on the next same-name tool call.
- Keep a reply queued until resume_user_interaction returns successfully, with a checkpoint after each acknowledged delivery.
- Preserve interaction_id, message_type, field/type, and option routing values verbatim while filtering only display content.
- Initialize teardown status to failed before tool setup and before each pattern attempt.

Validation: the complete tests/core/agent suite plus the user-interaction and output-wrapper tests pass; Ruff, mypy, and all pre-commit hooks pass.

rogercloud (commented):
## Re-review summary

No findings.

Commit `53bdf36d` addresses all five findings from the review of `156974c1`: paused runs now discard the stale tool-call suffix and replan; interaction replies are delivered to the suspended interaction before replanning and remain retryable until the callback succeeds; routing identifiers bypass lossy output filtering; and tool setup failures reach teardown with `failed` status.

Validation on the latest head: the focused ReAct, runner, user-interaction, and output-filter test set passed under Python 3.11 with the latest worktree explicitly selected on `PYTHONPATH`.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1151,28 +1191,103 @@ def _mark_latest_user_message_as_waiting_response(
...
+
+        while self.pending_tool_interaction_responses:
+            pending = self.pending_tool_interaction_responses[0]
+            tool_name = pending.get("tool_name", "")
+            tool = self._find_tool(tool_name, tools)
+            resume = user_interaction_resume_callable(tool)
+            if resume is None:
+                raise ValueError(
+                    f"Tool {tool_name} cannot resume a pending user interaction."
+                )
```
rogercloud, line 1299: [P2] Validate or support callback-less pauses

The pause path accepts any `waiting_for_user` result, but this branch raises on resume when the tool lacks the documented optional `resume_user_interaction` capability. Such a tool can pause and collect a reply successfully, then deterministically dead-end. Either reject the pause before presenting it as resumable, or support callback-less pauses by replanning without queuing callback delivery.

src/xagent/core/tools/adapters/vibe/output_filter_wrapper.py
```diff
@@ -149,6 +208,65 @@ def _make_async_wrapper(self, original_func: Any) -> Any:
...
+        if "options" in item:
+            filtered["options"] = self._filter_interaction_options(item["options"])
+
+        display_fields = 0
+        for key, value in item.items():
+            if key in _INTERACTION_CONTROL_KEYS or key == "options":
+                continue
+            if display_fields >= self._filter.max_fields:
+                break
+            filtered[key] = self._filter.filter(value, self._target.name)
```
rogercloud: [P2] Preserve the complete interaction control envelope

This field budget treats control properties such as `accept`, `multiple`, and `multiline` as expendable display fields; the same cap also slices the interactions and options lists. With a low configured `max_fields`, a valid file question loses `multiple=false`/`accept` and can lose later questions or choices, changing what the UI permits and what the suspended tool receives. Preserve control/cardinality fields and collections verbatim, filtering only user-facing text.

rogercloud (commented):
## Final approval gate

Two actionable findings remain and are attached inline. I am not approving this revision yet.

Validation: all 14 GitHub CI checks pass; the focused ReAct/runner/user-interaction/output-filter tests pass locally on Python 3.11; Ruff, mypy, and `git diff --check` pass.

rogercloud (approved):
## Approval summary

Approved at head `53bdf36d`. The two remaining findings are recorded inline in the preceding review and are treated as non-blocking follow-ups.

Validation: all 14 GitHub CI checks pass; the focused local tests pass on Python 3.11; Ruff, mypy, and `git diff --check` pass.

src/xagent/core/agent/pattern/react/react.py
qinxuye: Fixed in 4a4e3db3. Reply delivery is now queued only when the suspended tool exposes resume_user_interaction. Callback-less tools keep the annotated user answer in context and resume through normal ReAct replanning; legacy pending deliveries with an unavailable callback are skipped with an explicit checkpoint instead of dead-ending. Added end-to-end and legacy-checkpoint regression coverage.

src/xagent/core/tools/adapters/vibe/output_filter_wrapper.py
qinxuye: Fixed in 4a4e3db3. Interaction and option collections are no longer sliced by max_fields, and every routing, cardinality, and submitted-value property is preserved verbatim. Filtering now applies only to user-facing display text such as label, description, and placeholder. The low-budget regression covers multiple interactions/options plus accept, multiple, multiline, default, field, and value.

AlexLiu190625:
Follow-up ownership:

[Issue 1077](https://github.com/xorbitsai/xagent/issues/1077) owns normalization of the clarification signals supported by this runtime substrate and their root-only atomic publication as a checkpoint, waiting transition, canonical interaction, and transcript projection.

This merged PR remains the runtime foundation. The follow-up does not reopen its scope or move lifecycle persistence into the runtime control implementation.

PR #298, opened 2026-04-15
feat(mcp): Support connecting to MCP applications

yiboyasss:

PR #1028, opened 2026-07-27
feat: add Facebook comment reading and post engagement metrics

yiboyasss:


rogercloud (changes requested):
Two issues need to be addressed before merge:

1. **[P1] Request the permission required to read user comments** (`src/xagent/web/tools/mcp/facebook.py:144`)

   `facebook_list_post_comments` reads user-generated Page content, but the Facebook connector requests only `pages_show_list`, `pages_read_engagement`, and `pages_manage_posts`. Meta documents comments by users or other Pages under `pages_read_user_content`, so connected accounts will receive an OAuth error. Add that scope to the Facebook registry and ensure existing connections reauthorize.

2. **[P2] Limit expanded engagement edges to summary-only data** (`src/xagent/web/tools/mcp/facebook.py:120`)

   `likes.summary(true)` and `comments.summary(true)` expand those edges and return their default data pages as well as totals. With up to 100 posts, this can unnecessarily return thousands of user/comment objects when the feature only promises metrics. Use `likes.limit(0).summary(true)` and `comments.limit(0).summary(true)`.

Focused verification: all 11 tests in `tests/web/tools/test_facebook_mcp.py` pass. The tests mock the Graph API and therefore do not exercise these OAuth and field-expansion contracts.

src/xagent/migrations/versions/20260728_add_facebook_pages_read_user_content_scope.py
```diff
@@ -0,0 +1,53 @@
...
+
+    bind.execute(
+        sa.update(PUBLIC_MCP_APPS_TABLE)
+        .where(PUBLIC_MCP_APPS_TABLE.c.app_id == APP_ID)
+        .values(oauth_scopes=scopes)
+    )
+
+
+def upgrade() -> None:
+    _set_scopes(op.get_bind(), CURRENT_SCOPES)
```
rogercloud: **[P1] Force reauthorization for existing Facebook grants**

This migration updates only `public_mcp_apps`. Existing `UserOAuth(provider="facebook")` rows retain tokens issued without `pages_read_user_content`, while `_oauth_account_can_connect` checks only token presence/expiry and the runtime injects that token unchanged. After deployment those users still appear connected, but comment reads fail. Mark affected grants as requiring reconnection, or compare granted versus required scopes before treating the connector as connected.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -332,6 +332,7 @@ def get_builtin_public_mcp_app_rows() -> list[dict[str, Any]]:
                 "pages_show_list",
                 "pages_read_engagement",
                 "pages_manage_posts",
+                "pages_read_user_content",
```
rogercloud, line 335: **[P2] Cover the `META_CONFIG_ID` authorization path**

This registry scope is ignored whenever `META_CONFIG_ID` is configured because that login path sends `config_id` and deliberately omits `scope`. Unless the external Meta Login configuration is also updated, even newly reconnected users will not grant `pages_read_user_content`. Document and validate the required configuration update, and add a Facebook-specific config-ID test.

rogercloud (changes requested):
Re-review of `deee0711`: the summary-only engagement query is fixed, and the requested Facebook permission is now present for new raw-scope OAuth flows. The authorization rollout remains incomplete in the two supported scenarios called out inline. Focused Facebook MCP and migration tests pass; the Alembic graph has a single valid head.

src/xagent/migrations/versions/20260728_add_facebook_pages_read_user_content_scope.py
```diff
@@ -0,0 +1,90 @@
...
+    if "user_oauth" not in set(inspector.get_table_names()):
+        return
+
+    columns = {c["name"] for c in inspector.get_columns("user_oauth")}
+    if not {"provider", "access_token"}.issubset(columns):
+        return
+
+    bind.execute(
+        sa.update(USER_OAUTH_TABLE)
+        .where(USER_OAUTH_TABLE.c.provider == APP_ID)
```
rogercloud, line 91: **[P1] Invalidate provider-wide Meta grants too**

This only clears `provider="facebook"`, but `generic_oauth_login` still supports `app_id=None` and stores those grants as `provider="meta"`; `_resolve_legacy_oauth_access_token` then checks both `meta` and `facebook`, so that stale token can still service the Facebook connector. Those users remain connected with a grant issued before `pages_read_user_content` and comment reads still fail. Invalidate the legacy `meta` grant as well (accounting for the shared Instagram connection), or prevent provider-wide grants from satisfying Facebook, and add a regression test for that row.

rogercloud (changes requested):
Re-review of `b75ce7d5`: four of the five prior issues are fixed, including the summary-only engagement query and the `META_CONFIG_ID` authorization documentation/test. The authorization rollout remains partial in the supported legacy path called out inline. Focused verification passed: 11 Facebook MCP tests, 6 migration tests, a single valid Alembic head, and `git diff --check`.

src/xagent/web/mcp_apps.py
```diff
@@ -13,6 +13,17 @@
...
+# Apps that must not be satisfied by a bare provider-level OAuth grant (one
+# created via the app_id-less connect flow, e.g. UserOAuth.provider == "meta").
+# That flow requests only the provider's default scopes (see
+# generic_oauth_login's app_scopes=None branch when app_id is absent), never
+# an app's own oauth_scopes, so it can't carry a permission such as
+# pages_read_user_content that was added after the bare flow already existed.
+# Only an app-scoped grant (UserOAuth.provider == the app_id) counts for these
+# apps; Instagram is deliberately excluded so its existing bare "meta" grants
+# keep working, since its required scopes haven't changed.
+APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT = frozenset({"facebook"})
```
rogercloud, line 25: **[P2] Apply the app-scoped policy to resolver-hook candidates**

This allowlist is currently consumed only by connected-state lookup and the legacy `UserOAuth` fallback. `_build_mcp_server_config` calls `_oauth_token_provider_candidates` first, which still returns `["meta", "facebook"]` for Facebook; a configured token-resolver hook that returns a generic `meta` token therefore wins immediately and is injected into the Facebook MCP process, bypassing the new policy. Filter hook candidates as well (so Facebook resolves only `facebook`) and add a regression test where the resolver offers a `meta` token.

rogercloud (changes requested):
Re-review of `8fef1e18`: all six prior findings are fixed, including the provider-wide `UserOAuth(provider="meta")` path. The centralized app-scoped policy is not yet applied to every supported token source; the remaining token-resolver hook gap is called out inline. Focused verification passed: 11 Facebook MCP tests, 6 migration tests, one valid Alembic head, and `git diff --check`. The new legacy OAuth test module could not be collected in the minimal isolated environment because the full project dependency chain (starting with `pyarrow`) was unavailable.

rogercloud (commented):
## Summary

This PR adds a `facebook_list_post_comments` MCP tool and surfaces like/comment/share counts on `facebook_list_page_posts` (`28c36e81`), then layers on the OAuth work that review feedback surfaced: hardened Graph-path id validation and `limit(0).summary(true)` on the new summary fields, plus the `pages_read_user_content` scope (`419964a4`); a migration that force-clears stored Facebook tokens so users reconnect and re-grant (`39a12007`, rebased in `4921159b`); and a new `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT` policy that stops bare `provider="meta"` grants from satisfying Facebook, applied first to the legacy/connector paths (`b7edae61`) and then to the token-resolver-hook path (`d03dde0b`).

## Update since last review

All seven previously-raised findings were independently re-verified against the current code: six are FIXED (several confirmed causally, by reverting the fix files and re-running the new regression tests to prove they are not tautological), and one — the `META_CONFIG_ID` scope gap — is appropriately WAIVED as unfixable in-repo, with documentation and a locking test added. Beyond that re-check, this review also ran a fresh design + discovery pass done blind to the prior reviews, which turned up five new MAJOR issues not previously flagged, including a downstream consequence of the waived `META_CONFIG_ID` gap and a normalization inconsistency that reopens exactly the bypass this PR exists to close.

## Round 0 — approach verdict

**acceptable-with-reservations.**

The *filtering* half of the design is sound and well-targeted: `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT` correctly encodes "a bare `meta` grant must not satisfy Facebook", and it is now applied at the source in `_oauth_token_provider_candidates` so both hook callers inherit it.

The *scope-guarantee* half is not actually a guarantee, and the responsibility for "which scopes / grant policy does app X need" is now spread across four places:

- **Guarantee does not hold under `META_CONFIG_ID`** — in that deployment mode the Meta App Dashboard Login Configuration determines granted scopes, not the app's requested-scopes list, so the reconnect migration cannot deliver `pages_read_user_content` (see MAJOR-1).
- **Four sources of truth** — the code registry (`builtin_mcp_registry.py`), the `public_mcp_apps` DB row, the migration's `_set_scopes`, and a hardcoded app-id allowlist. Only the first actually drives runtime behavior (see MINOR-3).
- **Allowlist predicate duplicated three times, with inconsistent normalization** — the same "narrow candidates to app_id" logic is re-implemented at three call sites (MAJOR-3), and two of the three skip normalization that the third performs, producing a reachable bypass (MAJOR-4).
- **Read/delete asymmetry** — making the read path app-scoped without touching the symmetric delete path creates a live cross-app disconnect bug (MAJOR-5).

None of these block the feature itself; MAJOR-3/4 are a single shared-helper fix, and MAJOR-1/2/5 are each small, targeted changes.

## Prior findings — re-review checklist

1. `_post_path` / `_page_path` crash on non-string/None ids — **FIXED** (`src/xagent/web/tools/mcp/facebook.py:28-37`; matches the suggested fix, `_page_path` hardened too).
2. Missing `pages_read_user_content` scope — **FIXED** (`src/xagent/web/builtin_mcp_registry.py:335`).
3. Unbounded `likes.summary(true)` / `comments.summary(true)` expansion — **FIXED** (`src/xagent/web/tools/mcp/facebook.py:118-121`, now `limit(0).summary(true)` on both).
4. Existing Facebook grants not force-reauthorized (migration only touched `public_mcp_apps`) — **FIXED** (`_invalidate_existing_facebook_grants` clears `user_oauth.access_token` for `provider='facebook'`; confirmed to causally force `_oauth_account_can_connect` to report disconnected).
5. `META_CONFIG_ID` authorize path ignores the registry scope list — **WAIVED-appropriately** (not fixable in-repo; operator docs at `example.env:415-424` plus `test_meta_login_uses_config_id_without_scope_for_facebook`). Its downstream consequence is escalated as MAJOR-1.
6. Bare `provider="meta"` grants satisfying Facebook via the legacy path — **FIXED** (`_resolve_legacy_oauth_access_token` and `_oauth_keys_for_app`; verified by reverting the fix files and confirming the new regression tests fail).
7. Resolver-hook candidates path leaking a bare `meta` token ahead of the app-scoped policy — **FIXED** (`_oauth_token_provider_candidates` filters at the source; regression test in `tests/web/tools/test_oauth_token_resolver_hook.py` fails on revert).

## New findings

### MAJOR-1 — the reconnect migration is a no-op under `META_CONFIG_ID`

`src/xagent/web/api/auth.py:1138-1142`

When `META_CONFIG_ID` is set, the authorize URL sends `config_id` and — because the branch is a strict `if/elif` — never sends `scope` at all. Granted scopes come solely from the static Meta Dashboard Login Configuration. That means the migration's entire remediation strategy (mass-clear Facebook tokens → users reconnect → users re-grant `pages_read_user_content`) accomplishes nothing in this mode: every affected user is forcibly disconnected and then reconnects with exactly the same scope set they had before.

The PR's only response is a documentation comment (`example.env:415-424`) and a test that documents the gap (`tests/web/test_generic_oauth_login.py:275-323`). There is no code-level mitigation. Suggested: either skip the mass-invalidation when `META_CONFIG_ID` is set (a forced disconnect with no possible remedy is strictly worse than no migration), or emit a startup/migration warning telling the operator that the Login Configuration must be updated in the Meta App Dashboard for the migration to have any effect.

### MAJOR-2 — bare Meta login still creates an unusable Facebook `UserMCPServer` row, and the agent runtime picks it up

`src/xagent/web/api/auth.py:1400-1418`

The bare-Meta-login batch-connect loop selects every app with `provider == "meta"` (`apps = [app for app in get_all_mcp_apps(db) if app.get("provider") == provider]`) and calls `_ensure_user_mcp_server` for each. It was never updated to skip apps in `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT`, so a bare Meta login still creates a Facebook `UserMCPServer` row whose token can never resolve.

This is worse than cosmetic. The connector catalog (`_oauth_keys_for_app`) correctly reports Facebook as not connected, but the agent-runtime tool loader `_load_mcp_server_configs` (`src/xagent/web/tools/config.py:3298-3316`) does not go through that connected-check — it queries active `UserMCPServer` rows directly, picks up the orphan row, and lands on `reason="oauth_token_required"` (`config.py:3069`) at execution time. An agent run can therefore hit a broken Facebook tool for a connector the UI told the user was never connected.

No test covers the bare-Meta-login path for Facebook specifically; existing tests either pass `app_id="facebook"` directly or exercise a different provider.

### MAJOR-3 — the app-scoped-grant predicate is implemented three times

- `src/xagent/web/api/mcp.py:1592-1597` (`_oauth_keys_for_app`)
- `src/xagent/web/tools/config.py:424-441` (`_oauth_token_provider_candidates`)
- `src/xagent/web/tools/config.py:2859-2881` (`_resolve_legacy_oauth_access_token`)

The same predicate — "if `app_id` requires an app-scoped grant, narrow candidates to just `app_id`" — is written out independently at each site rather than shared. Verified that no circular-import constraint blocks consolidation: hoisting the `mcp_apps` import to module scope in a scratch copy imports cleanly. A single helper such as `restrict_to_app_scoped(app_id, candidates)` in `mcp_apps.py` would serve all three sites, and would also close MAJOR-4 by construction.

### MAJOR-4 — inconsistent `app_id` normalization reopens the bypass this PR closes

Same three call sites as MAJOR-3. `_oauth_keys_for_app` (`src/xagent/web/api/mcp.py:1592`) normalizes via `_normalize_app_key` (lowercase, strip, collapse whitespace) before the membership test; `_oauth_token_provider_candidates` (`src/xagent/web/tools/config.py:432`) and `_resolve_legacy_oauth_access_token` (`src/xagent/web/tools/config.py:2872`) test the raw string.

`PublicMCPApp.app_id` is admin-settable with no format validation (`POST /admin/mcp/apps`, `src/xagent/web/api/admin_mcp.py:76-92`), and the builtin-app guard `is_builtin_public_mcp_app` does an exact, case-sensitive match. So an admin creating a custom app with `app_id="Facebook"` (or `" facebook "`) gets the app-scoped-grant filter enforced at the connector-display layer (normalizes to `facebook`, matches the allowlist) but silently skipped at *both* token-resolution layers (`"Facebook" != "facebook"`, filter never fires) — a bare `meta` grant would satisfy that app's token resolution even though the UI reports it as requiring an app-scoped Facebook grant. This is precisely the class of bug this PR exists to close, reintroduced through normalization drift. Folding this into the MAJOR-3 shared helper (which normalizes once, inside the helper) fixes both.

### MAJOR-5 — disconnecting Facebook still deletes the shared `meta` grant, silently disconnecting Instagram

`src/xagent/web/api/mcp.py:2824-2830` (`delete_mcp_server`, unchanged by this PR)

`UserOAuth` has no `app_id` column: a bare Meta grant and an app-scoped Facebook grant are simply two rows keyed by `provider` (`'meta'` vs `'facebook'`). The disconnect path deletes `UserOAuth WHERE provider IN (provider, app_id)` — for Facebook that is `IN ('meta', 'facebook')`.

This PR made the **read** path asymmetric (Instagram, absent from `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT`, still accepts the shared bare `provider='meta'` row; Facebook no longer does) while leaving the **delete** path symmetric. Result: disconnecting Facebook deletes the shared `provider='meta'` row that Instagram may still depend on, silently disconnecting Instagram too. The code path predates this PR and was latent while both sides were symmetric; the asymmetric read-side fix turns it into a live, user-visible inconsistency, so it belongs in this PR or an immediate follow-up.

### MINOR-2 — migration docstring overstates the reconnect guarantee

`src/xagent/migrations/versions/20260728_add_facebook_pages_read_user_content_scope.py:55-64`

"the connector UI prompts the user to reconnect and grant the new permission" — users can deselect individual permissions on Meta's consent screen, and under `META_CONFIG_ID` the scope is never requested at all (MAJOR-1). Worth softening to describe the intent rather than a guarantee.

### MINOR-3 — the migration's `_set_scopes` does not affect runtime OAuth behavior

`src/xagent/web/mcp_apps.py:44-66` (`_app_to_dict`)

`_set_scopes` updates `public_mcp_apps.oauth_scopes`, but for builtin apps `_app_to_dict` sources `oauth_scopes` from the code registry (`get_builtin_execution_fields`), not the DB row. The DB write exists solely to satisfy `validate_builtin_public_mcp_apps` drift detection. Nothing in the migration says so; a future reader could reasonably conclude it drives runtime scope requesting. A one-line comment in the migration would prevent that.

### MINOR-4 — `_post_path` and `_page_path` are near-identical

`src/xagent/web/tools/mcp/facebook.py:28-37`

The two helpers differ only in parameter name and error message. Collapsible into one `_graph_path(value, name, suffix)`.

### MINOR-5 — `paging` handling crashes on an explicit `null`

`src/xagent/web/tools/mcp/facebook.py:127` and `:154`

`result.get("paging", {}).get("next")` raises `AttributeError` if Graph API returns `"paging": null` — the `{}` default applies only when the key is absent, not when its value is `None` (confirmed by direct repro). It is swallowed by the surrounding bare `except Exception` and surfaces as an opaque error. The codebase already uses the defensive form elsewhere (`src/xagent/web/tools/mcp/hubspot.py:122`: `(page.get("paging") or {}).get("next")`). One-line fix, low urgency; affects both the new tool and the pre-existing one.

### MINOR-6 — `facebook_list_post_comments` docstring does not match `filter=stream`

`src/xagent/web/tools/mcp/facebook.py:148`

The docstring says "comments left by fans", but `filter=stream` returns the full flattened comment stream, including nested replies and the Page's own replies. No `parent` field is requested either, so callers cannot distinguish top-level comments from nested ones. Either correct the docstring or request `parent` and document the distinction.

### MINOR-7 — composite post-id format is undocumented

`src/xagent/web/tools/mcp/facebook.py`

The `facebook_list_post_comments` docstring does not mention that Facebook post IDs take the composite `{page_id}_{post_id}` form. Passing a `post_id` that does not belong to `page_id` yields a raw Graph API error with no guidance for the model or user.

### MINOR-9 — missing negative test at the resolver-hook layer

`tests/web/tools/test_oauth_token_resolver_hook.py`

Only the Facebook (filtered) case is covered at this layer. There is no counterpart asserting that Instagram (not in `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT`) still receives `["meta", "instagram"]` candidates, so an over-filtering regression in `_oauth_token_provider_candidates` would only be caught by the separate legacy-path tests.

- **shrink** — extract the app-scoped-filter predicate duplicated across `mcp.py:1592-1597`, `config.py:424-441`, and `config.py:2859-2881` into one `restrict_to_app_scoped()` helper in `mcp_apps.py` (also fixes MAJOR-3/MAJOR-4). Approx. −20 lines net.
- **delete** — collapse `_post_path` / `_page_path` (`facebook.py:28-37`) into a single `_graph_path(value, name, suffix)` (MINOR-4). Approx. −6 lines.
- **shrink** — the local `from ...web.mcp_apps import APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT` import is repeated inside two functions in `config.py`; once the shared helper lives in `mcp_apps.py`, this becomes a single module-scope import. Approx. −3 lines.

Net: roughly −30 lines, with the duplication-driven bypass (MAJOR-4) eliminated structurally rather than by patching each site.

## Test results

All tests in the touched files pass: 167 tests across the files modified by this PR, plus 89 tests in the two new OAuth-scoping test files (`tests/web/tools/test_oauth_token_resolver_hook.py` and the legacy-path scoping tests). The new regression tests were additionally verified as non-tautological — reverting the fix files causes them to fail as expected, for both the legacy-path fix (prior finding 6) and the resolver-hook fix (prior finding 7).

rogercloud (commented):
## PR Summary

This PR adds Facebook comment reading and post engagement metrics: a new `facebook_list_post_comments` tool, comment/like summary counts on `facebook_list_page_posts`, and the `pages_read_user_content` scope required to read them. Because that scope expansion invalidates existing Facebook tokens, the PR also introduces an app-scoped OAuth grant policy for Facebook — Facebook can no longer ride on a bare `meta` grant and must have its own `facebook` grant row — plus an alembic migration that blanks existing Facebook access tokens to force reconnection.

## Update Summary

Commit `aae586a5` closes all 12 findings from the previous round (5 MAJOR + 7 MINOR). The centerpiece is consolidating the previously triplicated app-scoped-grant predicate into `requires_app_scoped_oauth_grant()` / `restrict_to_app_scoped_oauth_grant()` in `src/xagent/web/mcp_apps.py`, built on a single shared `_normalize_oauth_grant_key()`; all five functional call sites now delegate to it, which fixes the normalization-bypass, the Instagram-grant-deletion bug, and the bare-Meta batch-connect bug in one stroke. The migration now detects `META_CONFIG_ID` and skips mass invalidation with an operator-facing warning instead of silently no-op'ing. A fresh independent discovery pass on the fix commit surfaced one MAJOR-severity risk that needs author verification against a real token (NEW-MAJOR-1) plus six minor items, none of which are security or data-integrity regressions.

## Round 0 Verdict: sound

Upgraded from *acceptable-with-reservations*. The app-scoped-grant policy is now expressed once and applied consistently on every path that matters — connect, token resolution, and deletion. Normalization happens exactly once, at a single choke point, so the admin-triggerable bypass class is closed by construction rather than by three parallel implementations agreeing. One display-only call site was missed (MINOR-A) and the migration leaves some harmless pre-existing data debris (MINOR-B), but neither has functional or security impact.

Forward-looking note (not blocking): `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT` remains a hardcoded name set rather than being derived from data (e.g. "app's `oauth_scopes` is not a subset of the provider's `default_scopes`"). That is acceptable for a single-app policy today, but will need revisiting the second time an app joins the set.

## Prior Findings — Verification Checklist

All 12 verified against `aae586a5` with file:line evidence; the four highest-risk fixes were additionally confirmed by reverting the fix and re-running the new tests to prove the guards are non-tautological.

| # | Finding | Status |
|---|---------|--------|
| MAJOR-1 | `META_CONFIG_ID` mass-invalidation is a no-op | **FIXED** — `upgrade()` now checks `os.environ.get("META_CONFIG_ID")`, skips `_invalidate_existing_facebook_grants`, and emits a `logger.warning` (confirmed to reach the console via alembic's root logger) directing the operator to update the Meta App Dashboard Login Configuration. Covered by `test_upgrade_leaves_grants_connected_under_meta_config_id`; 7/7 migration tests pass. |
| MAJOR-2 | Bare-Meta batch-connect creates an unusable Facebook server | **FIXED** — the batch-connect loop in `api/auth.py` now calls `requires_app_scoped_oauth_grant(app_info.get("id"))` and skips those apps during a bare Meta login. Verified by reverting the skip: `test_bare_meta_login_skips_facebook_but_still_connects_instagram` fails as expected. |
| MAJOR-3 | Triplicated app-scoped-grant predicate | **FIXED** — consolidated into `requires_app_scoped_oauth_grant()` / `restrict_to_app_scoped_oauth_grant()` in `web/mcp_apps.py`. Grep confirms no remaining inline reimplementation across all five call sites (`api/mcp.py:1592`, `api/mcp.py:2828`, `tools/config.py:427`, `tools/config.py:2864`, `api/auth.py:1416`). |
| MAJOR-4 | Inconsistent normalization → admin-triggerable bypass | **FIXED** — normalization is now performed exactly once inside `_normalize_oauth_grant_key` (case-fold + strip + whitespace-collapse) and reached uniformly via the shared helpers. Confirmed in a shell that `"Facebook"`, `"facebook"`, and `" facebook  "` all return `True`; regression test `test_restrict_to_app_scoped_oauth_grant_normalizes_admin_created_app_id` covers the `" Facebook "` case. |
| MAJOR-5 | Facebook disconnect deletes Instagram's shared `meta` grant | **FIXED** — `delete_mcp_server` computes `providers_to_delete` via `restrict_to_app_scoped_oauth_grant(app_id, [provider, app_id])`, so only the `facebook` row is removed. Verified by reverting the fix: `test_disconnecting_facebook_preserves_shared_bare_meta_grant_for_instagram` fails with `assert 0 == 1`. |
| MINOR-2 | Migration docstring overstated the reconnect guarantee | **FIXED** — docstring now states the reconnect "can't guarantee they grant the new permission (they can deselect it there), and under `META_CONFIG_ID` it can't even offer the chance." |
| MINOR-3 | `_set_scopes` DB write is a runtime no-op, undocumented | **FIXED** — new docstring states it does not drive authorize-time scope requests and exists solely to keep `validate_builtin_public_mcp_apps` from reporting drift. |
| MINOR-4 | `_post_path` / `_page_path` duplication | **FIXED** — collapsed into a single `_graph_path(value, name, suffix)` used at all four call sites. |
| MINOR-5 | `paging=null` crash | **FIXED** — both tools now use `(result.get("paging") or {}).get("next")`. |
| MINOR-6 | Docstring vs `filter=stream` mismatch | **FIXED** — the false claim is removed and the flattened-stream / no-`parent`-field limitation is documented (behavior unchanged; see MINOR-E). |
| MINOR-7 | Composite `post_id` format undocumented | **FIXED** — docstring now documents the `{page_id}_{post_id}` composite format. |
| MINOR-9 | Missing Instagram hook-path test | **FIXED** — `test_hook_still_accepts_bare_meta_token_for_instagram` added. |

## New Findings

### NEW-MAJOR-1 — Combined `fields` request may break `facebook_list_page_posts` on under-scoped tokens

**Confidence caveat, stated up front:** roughly 75% on the underlying Graph API behavior. This is based on documented community reports, *not* verified live against a real token in this session. Please verify before merging — if the behavior is not as described, this finding collapses to nothing.

In `src/xagent/web/tools/mcp/facebook.py`, `facebook_list_page_posts` — a tool that worked before this PR — now requests `likes.limit(0).summary(true),comments.limit(0).summary(true)` in the *same combined* `fields` string as the basic post fields, in a single `GET /{page_id}/feed` call. Meta's Graph API is reported to reject the *entire* request with an `OAuthException` when a requested field/edge's permission (here `pages_read_user_content`, needed for the comments summary) is absent from the token, rather than omitting just that field.

Combined with the MAJOR-1 outcome — `META_CONFIG_ID` deployments intentionally retain under-scoped tokens until an operator updates the Meta Dashboard Login Configuration — this means that on such a deployment, or on any token where the operator hasn't yet made that change, `facebook_list_page_posts` would start failing *entirely*, not just lose the new comment-count feature. That is a regression in a previously working tool, not a degradation of a new one. There is currently no test for a partial or permission-denied Graph response.

Suggested: verify against a real token lacking `pages_read_user_content`. If confirmed, either split the summary fields into a separate request, add a fallback that retries without the summary fields on a permission error, or at minimum document the hard coupling prominently.

### MINOR-A — `_enrich_oauth_server_info` is the one call site the consolidation missed

`src/xagent/web/api/mcp.py:1536` still does `oauth_emails.get(app_id) or oauth_emails.get(provider)` with no app-scoped-grant guard, feeding `GET /servers` and `GET /servers/{id}`. This was initially investigated as a possible MAJOR and downgraded during verification: `connected_account` flows only into a frontend display field and never into any token-resolution path, so the impact is purely cosmetic. A user holding only a bare `meta` grant would see Facebook's UI badge as "connected as user@example.com" while `list_mcp_apps` and the runtime correctly treat it as disconnected.

Separately, `oauth_emails` (`api/mcp.py:2049-2056`, `2165-2168`) has no liveness filter (`_oauth_account_can_connect`), so a user whose token this PR's migration just blanked would still see a stale "connected as ..." badge immediately afterwards.

Suggested: route `oauth_emails` lookups through `restrict_to_app_scoped_oauth_grant` as well, and/or filter by `_oauth_account_can_connect`.

### MINOR-B — Migration doesn't clean up pre-existing orphan `UserMCPServer` rows

The migration blanks `user_oauth.access_token` but leaves behind Facebook `UserMCPServer` rows created by past bare-Meta-login batch-connects. That is a real, reachable historical state: the batch-connect loop has existed since PR #298 with no Facebook-specific exclusion until this fix. Also downgraded from an initial MAJOR hypothesis — the orphans degrade gracefully at execution time with `oauth_token_required`, the same message any disconnected state produces, so this is residual data debris rather than a crash or security issue. Optionally clean up in the same migration or in a follow-up.

### MINOR-C — Facebook-only users leak a bare `meta` grant row on disconnect

`src/xagent/web/api/mcp.py:2828`: narrowing `providers_to_delete` is correct for the Instagram-sharing case MAJOR-5 targeted, but for a Facebook-only user (no Instagram connected) disconnecting Facebook now permanently leaves the bare `meta` `UserOAuth` row behind with no UI path to remove it. Previously it was deleted. Same shape as the edge case noted while verifying MAJOR-5: a user who historically connected Facebook via *only* a bare `meta` grant will orphan that row on disconnect.

### MINOR-D — `facebook_list_post_comments` doesn't validate that `post_id` belongs to `page_id`

In `src/xagent/web/tools/mcp/facebook.py`, `facebook_list_post_comments(page_id, post_id)` performs no consistency check between the two arguments, even though Facebook composite IDs are `{page_id}_{post_id}` and therefore carry the page ID. A mismatched pair silently issues the query with the wrong page's token instead of raising a clear error.

### MINOR-E — `parent` field still not requested, so replies remain indistinguishable

The MINOR-6 fix documents the `filter=stream` / missing-`parent`-field limitation but doesn't close it. `parent` still isn't in the requested `fields`, so replies can't be distinguished from top-level comments. Documenting is better than not documenting, but adding `parent` to `fields` would close the gap outright.

### MINOR-F — Test-coverage gaps

Three gaps: no test covers `_enrich_oauth_server_info` / `GET /servers` for Facebook specifically (the MINOR-A path); no test covers a pre-existing orphan `UserMCPServer` row (MINOR-B); and `_graph_path`'s empty/`None`-value `ValueError` branch lost its explicit coverage when `_post_path` and `_page_path` were merged — it is now only implicitly exercised.

**Investigated and dropped:** a hypothesis that `_resolve_legacy_oauth_access_token`'s candidate-list consolidation dropped non-string candidates was checked and found to have no practical behavior change for any real provider/app_id combination — the only value ever filtered is a `None` app_id, which would not have matched anyway.

## Simplification

One small item; the incremental diff is otherwise lean.

- **stdlib** — `src/xagent/web/mcp_apps.py` (~line 69): the new consolidated helper hand-rolls a seen-set/list dedup loop, reinventing `list(dict.fromkeys(...))` — which the very code being consolidated (`_oauth_token_provider_candidates`) already used elsewhere. Suggested: `deduped = list(dict.fromkeys(c for c in candidates if isinstance(c, str) and c))`. Net ~-4 lines.

## Test Results

180 tests pass across the seven touched test files:

- `tests/web/test_meta_oauth.py`
- `tests/web/test_generic_oauth_login.py`
- `tests/web/tools/test_facebook_mcp.py`
- `tests/web/tools/test_legacy_oauth_provider_scoping.py`
- `tests/web/tools/test_oauth_token_resolver_hook.py`
- `tests/web/api/test_public_mcp_connector_visibility.py`
- `tests/alembic/test_20260728_add_facebook_pages_read_user_content_scope.py`

Single alembic head confirmed, no branching.

yiboyasss:
Thanks for the thorough re-review. MINOR-B and MINOR-D from the latest round are intentionally left as-is:

**MINOR-B** (migration doesn't clean up pre-existing orphan `UserMCPServer` rows) — per your own analysis, these degrade gracefully to the same `oauth_token_required` state as any disconnected app, so there's no functional or security difference from "never connected". There's also no schema-based way to distinguish a legitimate app-scoped `UserMCPServer` row from a historical bare-login orphan, so any cleanup query would risk deleting rows that are actually fine. Leaving this as residual data debris rather than guessing.

**MINOR-D** (`facebook_list_post_comments` doesn't validate `post_id` belongs to `page_id`) — a mismatched pair already surfaces a Graph API error today, just a less friendly one; this is a UX polish item rather than a correctness/security gap, so I'm deprioritizing it in favor of not growing the diff further for this PR.

Everything else from this round (NEW-MAJOR-1, MINOR-A, C, E, and the simplification) is fixed in 8942890c.

rogercloud (approved):
## Summary

This PR adds Facebook comment reading and post engagement metrics to the Meta connector: a `facebook_list_post_comments` tool, engagement fields (likes/comments counts) on `facebook_list_page_posts`, and the surrounding OAuth/connector plumbing so Facebook can be granted, displayed, and disconnected under the app-scoped-grant policy without disturbing the shared bare `meta` login or Instagram.

## Update since last round

Commit `8942890c` addressed 4 of the 6 open findings from the previous round — **NEW-MAJOR-1**, **MINOR-A**, **MINOR-C**, and **MINOR-E** — and the author posted a comment explicitly declining two others (**MINOR-B**, **MINOR-D**) with stated reasoning. This round independently re-verified every fix against `file:line` evidence (with real test runs for the highest-risk items, each confirmed to fail on revert), independently evaluated both waiver justifications, and ran one more fully independent Round 0 design pass. The PR is in very good shape; this is a final polish pass, not a merge blocker.

## Round 0 design verdict: sound

Carried forward from the prior round's upgrade and further corroborated here by a third, blind design pass. The app-scoped-grant policy (`requires_app_scoped_oauth_grant()` / `restrict_to_app_scoped_oauth_grant()` in `mcp_apps.py`) is the single policy source and is wired into all five relevant paths — bare-login batch connect, catalog/apps display, `/servers` display, the runtime token-resolver hook (including the legacy fallback), and disconnect cleanup — with no remaining bypass and no Instagram regression. The migration's scoping (blanks only `provider='facebook'`, never the bare `meta` row) and its `META_CONFIG_ID` skip are well-reasoned and documented. The independent pass surfaced the same two residual items already noted below (fallback error-classification, disconnect-cleanup TOCTOU) and no new blocking concerns.

## Findings checklist

| ID | Severity | Status |
|----|----------|--------|
| NEW-MAJOR-1 | Major | ✅ Fixed (minor residual noted) |
| MINOR-A | Minor | ✅ Fixed |
| MINOR-C | Minor | ✅ Fixed (minor residual noted) |
| MINOR-E | Minor | ✅ Fixed |
| MINOR-D | Minor | ✅ Waiver confirmed appropriate |
| MINOR-B | Minor | ⚠️ Waiver reasoning contested — see below |

### NEW-MAJOR-1 — Fixed

`facebook_list_page_posts` now requests `_POST_FIELDS_WITH_ENGAGEMENT` first and, on a `GraphAPIError`, logs a warning, sets `engagement_available=False`, and retries with `_POST_FIELDS_BASE` — so an under-scoped token degrades gracefully instead of failing the whole call. Docstring updated and `engagement_available` is surfaced in the response. The new test `test_list_page_posts_falls_back_without_engagement_on_permission_error` passes and was confirmed to fail on revert.

Residual (low-risk, non-blocking — corroborated by two separate verification passes): the fallback catches any `GraphAPIError` rather than inspecting `error.details` for an OAuth/permission-specific code. In practice this is well-contained — network errors raise a different exception type and are not masked, and a genuinely bad `page_id` or a rate-limit fails both the primary and fallback calls so it still surfaces as an error. The only realistic false-positive is a transient HTTP error on the first call that would have succeeded, reported as `engagement_available=false` instead of a transient failure. Worth a follow-up (classify on `error.details` before falling back), but not blocking.

### MINOR-A — Fixed

`_enrich_oauth_server_info` now iterates `restrict_to_app_scoped_oauth_grant(app_id, [app_id, provider])` and only surfaces an email from an allowed key, and both `oauth_emails` construction sites (`GET /servers` and `GET /servers/{id}`) additionally filter with `_oauth_account_can_connect` for liveness. Instagram's bare-meta display is confirmed unaffected (`test_instagram_server_list_still_shows_bare_meta_email_as_connected` passes). The full `test_meta_oauth.py` suite (14 tests) passes; the new bare-grant and blanked-token tests were each confirmed to fail when the corresponding guard is reverted.

### MINOR-C — Fixed

`delete_mcp_server` now checks the user's other `MCPServer`/`UserMCPServer` rows and, only when none resolve to an app sharing the same provider (via `get_app_for_mcp_server`), also deletes the bare provider-level `UserOAuth` row. Verified on both branches: Facebook+Instagram connected → disconnecting Facebook preserves the shared `meta` row (existing test still passes); Facebook-only → disconnecting now removes the orphaned `meta` row (new `test_disconnecting_facebook_only_user_also_removes_orphaned_bare_meta_grant`, confirmed to fail on revert).

Residual (low priority, corroborated by two passes): this is a check-then-delete with no transaction lock, so a concurrent disconnect of a sibling app could in theory race. The worst outcome only regresses to the pre-fix "permanent orphan" state — not a data-integrity or security issue — it self-heals on reconnect, and a single user clicking disconnect makes a true same-account race very unlikely. Noted for completeness only.

### MINOR-E — Fixed

Confirmed addressed in `8942890c` and covered by the passing suite.

### MINOR-D — Waiver confirmed appropriate

The author declined to add explicit `post_id`/`page_id` cross-validation, reasoning that a mismatched pair already surfaces a Graph API error (just a less friendly one) and that this is UX polish rather than a correctness or security gap. Independently confirmed: Facebook Page Access Tokens are strictly page-scoped, so a `post_id` belonging to another Page is rejected by Meta with a `GraphAPIError` — which the tool already catches and returns as a structured error. There is no silent wrong-data path and no cross-page data leak, so the waiver holds. If cheap, a friendlier upfront message would be a nice touch someday, but it is genuinely optional.

### MINOR-B — Waiver reasoning contested (please reconsider the stated justification)

We fully agree this is **low severity** and reasonable to defer: an orphaned `UserMCPServer` row degrades gracefully and is functionally indistinguishable from never-connected, so deprioritizing it is fine.

What we'd push back on is the stated *technical* justification — that "there's no schema-based way to distinguish a legitimate app-scoped row from a historical bare-login orphan." A reliable distinguishing query does exist: for any `UserMCPServer` whose resolved app_id is in `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT` (today, only Facebook), the row is a historical orphan **iff** no `UserOAuth(user_id=<that user>, provider=<that app_id>)` row exists. This is exact rather than heuristic, because every path that could legitimately create such a row today either also creates the matching app-scoped `UserOAuth` row (the app_id-specific connect flow) or is now explicitly skipped (the bare-login batch-connect, fixed last round). With `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT` currently holding a single member, that is a narrow, well-bounded, low-risk cleanup — not the open-ended guessing the reasoning implies.

Net: leaving this for a future PR is a perfectly fine call on severity grounds, but we'd suggest correcting the "can't be done safely" framing — a targeted migration is available whenever the team wants to close it.

## New minor findings (informational, very low priority)

- **NIT-1** — In `src/xagent/web/api/mcp.py` (the disconnect-cleanup sibling check), `sibling_app.get("provider") == provider` compares raw strings, whereas the rest of the module normalizes app/provider keys via `_normalize_app_key`. Both operands are catalog-controlled literals (e.g. `"meta"`) today, so it's correct in practice — flagged only for consistency with the module's normalization discipline.
- **NOTE-1** (pre-existing, not a regression from this PR) — If a second `meta`-provider app that does *not* require an app-scoped grant were ever added alongside Instagram, disconnecting one of those two non-app-scoped siblings would still unconditionally delete the shared bare `meta` row (the orphan-guard branch only triggers for app-scoped-only apps like Facebook), breaking the other. Only Instagram sits under `meta` today, so there's no live impact and this behavior predates the PR. No action needed now — just worth remembering if a third `meta`-based app is ever added.

## Simplification

Lean already. One purely-cosmetic nit if you happen to touch the line: in the new sibling-check comprehension in `src/xagent/web/api/mcp.py`, the `(sibling_app := get_app_for_mcp_server(db, other_server)) is not None and sibling_app.get(...)` can drop the explicit `is not None`, since `get_app_for_mcp_server` only ever returns `None` or a populated dict — a plain `and` short-circuits identically. Entirely optional.

## Test results

68 tests pass — `tests/web/test_meta_oauth.py` (14, including all four new/updated tests for this round), `tests/web/tools/test_facebook_mcp.py`, and `tests/web/api/test_public_mcp_connector_visibility.py`. Every new test was confirmed to fail when its corresponding fix is reverted.

Nice work on this round — the fixes are clean and well-tested. The only thing that actually warrants a reply is MINOR-B's stated reasoning; everything else here is FYI-level. No blocking issues remain — approving.

PR #1029, opened 2026-07-27
feat: select custom agents for Telegram conversations via /agents

OliverBryant:
## What

Telegram bot users can now choose which Agent Builder agent handles their conversation, directly inside Telegram:

- A new `/agents` command shows an inline keyboard listing the channel owner's custom agents (paginated, 8 per page), with a "Default assistant" option and a ✓ on the current selection.
- Tapping an agent starts a fresh conversation bound to that agent; subsequent messages run with the agent's instructions, model slots, and execution mode (via the existing `Task.agent_id` runtime resolution).
- `/new` reverts to the default assistant. Selections persist per Telegram user across bot restarts.
- If the selected agent is deleted before the next message, the turn falls back to the default assistant with a notice instead of failing.
- The bot registers its command menu (`/new`, `/agents`, `/stop`) via `set_my_commands` on startup so the command is discoverable.

## How

- `channel_runtime.py`: new `ChannelAgentSnapshot` plus `list_channel_owner_agents` / `get_channel_owner_agent` (worker-owned short sessions, channel-sender authorization, `owned_agent_clause` scoping with workforce-generated managers excluded). `prepare_channel_task` gains a keyword-only `agent_id`: ownership is validated at task creation, `Task.agent_id` is set, and the connector runtime selection snapshot is bound following the widget-channel precedent. Feishu call sites are untouched (default `None` keeps behavior byte-identical).
- `telegram/bot.py`: `/agents` command handler, `agsel:` / `agpage:` callback-query handlers, per-user selection persisted to `data/telegram_selected_agents_{instance}.json`, and the selection wired into task preparation with a stale-selection fallback notice.

## Tests

- `tests/web/services/test_channel_runtime.py`: agent binding, foreign/nonexistent/workforce-manager fallback, listing order/scoping/authorization, single-agent lookup.
- `tests/web/test_telegram_message_queue.py`: keyboard building and pagination, selection/default/missing-agent callbacks, empty-list and unauthorized `/agents`, persistence round-trip, batch path passing `agent_id` and clearing stale selections.

All 94 tests across the Telegram/Feishu/channel-runtime suites pass; ruff and mypy are clean on the touched files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/xagent/web/channels/telegram/bot.py
+++ b/src/xagent/web/channels/telegram/bot.py
@@ -1,4 +1,5 @@
 import asyncio
+import hashlib
 import html
 import json
 import logging
@@ -101,6 +102,7 @@ class TelegramBotInstance:
         self.user_active_executions: Dict[int, tuple[int, object]] = {}
         self.user_preparing_executions: set[int] = set()
         self.user_stop_events: Dict[int, asyncio.Event] = {}
+        self.user_conversation_generations: Dict[int, int] = {}
         self._accepting = True
         self._ingress_stopped = False
         self._stop_lock: asyncio.Lock | None = None
@@ -111,9 +113,7 @@ class TelegramBotInstance:
         self.active_tasks = self._load_active_tasks()
 
         # Per-Telegram-user custom agent selection for the next conversation
-        self.selected_agents_file = Path(
-            f"data/telegram_selected_agents_{instance_id}.json"
-        )
+        self.selected_agents_file = self._selected_agents_store_path(channel_id, token)
         self.selected_agents: Dict[int, int] = self._load_selected_agents()
 
         default_props = DefaultBotProperties(parse_mode=ParseMode.HTML)
@@ -158,6 +158,23 @@ class TelegramBotInstance:
             logger.error(
                 f"Failed to save Telegram active tasks for {self.instance_id}: {e}"
             )
+
+    @staticmethod
+    def _selected_agents_store_path(channel_id: Optional[int], token: str) -> Path:
+        """Durable per-channel selection store.
+
+        Lives under the storage root (survives container recreation, unlike a
+        cwd-relative ``data/`` path) and is keyed by the stable channel id —
+        falling back to a full-token hash — so distinct bots whose tokens share
+        a prefix can never collide on one file.
+        """
+        from ....config import get_storage_root
+
+        if channel_id is not None:
+            key = f"channel_{int(channel_id)}"
+        else:
+            key = hashlib.sha256(token.encode()).hexdigest()[:16]
+        return Path(get_storage_root()) / "telegram" / f"selected_agents_{key}.json"
 
     def _load_selected_agents(self) -> Dict[int, int]:
         if self.selected_agents_file.exists():
@@ -305,7 +322,15 @@ class TelegramBotInstance:
         )
         return True
 
+    def _conversation_generation(self, user_id: int) -> int:
+        return self.user_conversation_generations.get(user_id, 0)
+
     def _start_new_conversation(self, user_id: int) -> bool:
+        # Bumping the generation fences out any in-flight preparation for the
+        # previous conversation: its result must not overwrite the new state.
+        self.user_conversation_generations[user_id] = (
+            self._conversation_generation(user_id) + 1
+        )
         stopped = self._request_current_conversation_stop(
             user_id, reason="new Telegram conversation requested"
         )
@@ -527,7 +552,11 @@ class TelegramBotInstance:
                     return
                 self._set_selected_agent(user_id, agent.agent_id)
                 self._start_new_conversation(user_id)
-                await callback.answer(f"{agent.name} selected")
+                # answerCallbackQuery rejects texts over 200 characters
+                toast = f"{agent.name} selected"
+                if len(toast) > 200:
+                    toast = f"{toast[:197]}..."
+                await callback.answer(toast)
                 confirmation = (
                     f"Agent selected: {html.escape(agent.name)}. "
                     "Send a message to start a fresh conversation. "
@@ -881,6 +910,7 @@ class TelegramBotInstance:
 
         self.user_preparing_executions.add(user_id)
         self._clear_user_stop_request(user_id)
+        conversation_generation = self._conversation_generation(user_id)
         claimed_task_id: int | None = None
         managed_lease: ManagedTaskLease | None = None
         voice_asr_model: Any | None = None
@@ -941,6 +971,17 @@ class TelegramBotInstance:
             claimed_task_id = task_id
             owner_user_id = prepared_task.user_id
             is_new_task = prepared_task.is_new_task
+
+            # The user switched conversations (agent selection or /new) while
+            # this turn was being prepared: settle the stale claim without
+            # touching active_tasks or selected_agents.
+            if self._conversation_generation(user_id) != conversation_generation:
+                await self._finalize_requested_stop(
+                    managed_lease,
+                    task_id=task_id,
+                )
+                return
+
             if is_new_task:
                 self.active_tasks[user_id] = task_id
                 self._save_active_tasks()
@@ -1529,6 +1570,7 @@ class TelegramBotInstance:
             self.user_active_executions.clear()
             self.user_preparing_executions.clear()
             self.user_stop_events.clear()
+            self.user_conversation_generations.clear()
 
     async def _stop_ingress(self) -> None:
         try:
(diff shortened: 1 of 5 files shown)


src/xagent/web/channels/telegram/bot.py
```diff
@@ -360,6 +425,171 @@ def _is_stop_request_text(self, text: str) -> bool:
...
+        page_agents = agents[page * page_size : (page + 1) * page_size]
+
+        default_label = "Default assistant"
+        if selected_agent_id is None:
+            default_label += " ✓"
+        rows = [
+            [InlineKeyboardButton(text=default_label, callback_data="agsel:default")]
+        ]
+        for agent in page_agents:
+            label = agent.name[:60]
```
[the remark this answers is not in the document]
OliverBryant: Done — truncation now appends an ellipsis: names over 60 chars render as the first 57 chars + `...`, with a unit test covering it (fd4baa00).

src/xagent/web/services/channel_runtime.py
```diff
@@ -282,13 +404,27 @@ def _prepare_channel_task_sync(
...
+                    db=db,
+                    agent=agent_row,
+                    connector_user_id=owner_id,
+                )
+                bind_connector_runtime_selection_snapshot(
+                    task=task, selected_refs=selected_refs
                 )
                 db.add(task)
                 db.flush()
+            else:
```
rogercloud: [P1] Revalidate the selection before reusing an active task

After agent A completes its first Telegram turn, deleting A nulls Task.agent_id but leaves the persisted Telegram selection pointing at A. This branch returns the existing task with requested_agent_missing=false, and AgentServiceManager can then reuse its task-id-keyed cached service, so the deleted or visibility-revoked agent instructions, models, KBs, and tools continue to execute. On a cache miss it instead falls back silently without clearing the selection or sending the promised notice. Revalidate the requested selection for existing tasks too; when it is unavailable, clear/notify and create or evict into a clean default task (including its connector snapshot), with regression coverage for deletion and visibility loss.

src/xagent/web/channels/telegram/bot.py
```diff
@@ -360,6 +425,171 @@ def _is_stop_request_text(self, text: str) -> bool:
...
+                    external_user_id=str(user_id),
+                    agent_id=int(payload),
+                )
+                if agent is None:
+                    await callback.answer(
+                        "That agent is no longer available.", show_alert=True
+                    )
+                    return
+                self._set_selected_agent(user_id, agent.agent_id)
+                self._start_new_conversation(user_id)
```
rogercloud, line 554: [P2] Fence selection changes against in-flight preparation

Aiogram runs callbacks concurrently with the per-user message queue. If an A message is awaiting prepare_channel_task when this callback selects B, the old preparation can return afterward and overwrite active_tasks with the A task; the stop event only pauses that lease, so the next message reclaims A and ignores B. A stale-agent result can likewise clear the newer B selection. Track a per-user conversation generation, capture it before preparation, and after the await finalize/discard any stale result without writing active_tasks or selected_agents; add a deterministic blocked-preparation regression test.

src/xagent/web/channels/telegram/bot.py
```diff
@@ -360,6 +425,171 @@ def _is_stop_request_text(self, text: str) -> bool:
...
+                await callback.answer(f"{agent.name} selected")
```
rogercloud: [P2] Keep the callback answer within Telegram’s 200-character limit

Agent names may be 200 characters, but answerCallbackQuery allows at most 200 characters. A valid 192-character name makes this notification 201 characters, so Telegram returns an uncaught 400 after the selection and fresh-task reset have already been persisted; the keyboard remains stale even though state changed. Use a fixed answer or truncate this notification independently, and cover the 192-character boundary.

src/xagent/web/channels/telegram/bot.py
```diff
@@ -101,6 +110,12 @@ def __init__(
         self.active_tasks_file = Path(f"data/telegram_active_tasks_{instance_id}.json")
         self.active_tasks = self._load_active_tasks()
 
+        # Per-Telegram-user custom agent selection for the next conversation
+        self.selected_agents_file = Path(
+            f"data/telegram_selected_agents_{instance_id}.json"
```
rogercloud: [P2] Persist selections under a durable, collision-free key

This relative data path resolves under /opt/xagent in the documented container, while the durable backend volume is mounted at /root/.xagent, so selected-agent preferences disappear when the container is recreated for an upgrade. The instance id is also only the first eight token characters, allowing distinct bots to share this file and overwrite each other’s user selections. Store the map under get_storage_root() (or in the database) and key it by the stable channel_id/full-token hash; test both container-style persistence and same-prefix bot isolation.

rogercloud (changes requested):
Design verdict: acceptable-with-reservations. The selection/task-binding boundary is coherent, but revocation and concurrent conversation transitions need stronger lifecycle fencing before merge. The sole prior finding (adding an ellipsis to truncated button labels) is fixed in fd4baa0 and covered by a unit test. Validation: the three focused Telegram/channel-runtime suites passed (80 tests), and Ruff passed on all five changed files. No linked issue was provided; SaaS team-scope behavior was statically traced but not executed under the overlay.

src/xagent/web/services/channel_runtime.py
OliverBryant: Fixed in a4a59dc0. The selection is now revalidated on every turn, not only for new tasks: when the requested agent no longer resolves (deleted or visibility revoked), or the existing task's binding no longer matches the selection, the turn evicts to a fresh task (with its own connector snapshot) instead of resuming — which also sidesteps the task-id-keyed AgentService cache — and `requested_agent_missing` drives the clear+notify path as promised. Regression tests: `test_prepare_channel_task_evicts_existing_task_when_selection_stale` (mirrors `stage_delete_agent` nulling `Task.agent_id`), `test_prepare_channel_task_continues_task_when_selection_still_valid`, `test_prepare_channel_task_starts_new_task_when_binding_drifts`.

src/xagent/web/channels/telegram/bot.py
OliverBryant: Fixed in a4a59dc0. Added a per-user conversation generation: `_start_new_conversation` (both `/new` and selection callbacks) bumps it, the batch path captures it before `prepare_channel_task`, and a mismatch after the claim settles the stale lease (PAUSED) and returns without writing `active_tasks` or `selected_agents` — including the stale `requested_agent_missing` clear. Deterministic regression: `test_batch_discards_prepared_task_after_conversation_switch` switches the selection from inside the mocked preparation and asserts the -1 marker and the newer selection survive.

src/xagent/web/channels/telegram/bot.py
OliverBryant: Fixed in a4a59dc0. The selection toast is truncated independently to 200 chars (197 + ellipsis). Boundary covered by `test_agent_selection_toast_stays_within_telegram_limit` with a 192-char name (raw toast = 201).

src/xagent/web/channels/telegram/bot.py
OliverBryant: Fixed in a4a59dc0. Selections now persist under `get_storage_root()/telegram/`, keyed by the stable `channel_id` (`selected_agents_channel_<id>.json`) with a full-token SHA-256 hash fallback when no channel id exists — durable across container recreation and collision-free for same-prefix tokens. Covered by `test_selected_agents_store_path_is_durable_and_collision_free`.

src/xagent/web/services/channel_runtime.py
```diff
@@ -66,6 +80,8 @@ class ClaimedChannelTask:
     task_id: int
     is_new_task: bool
     managed_lease: ManagedTaskLease
+    agent_id: int | None = None
```
rogercloud: [shrink] Remove the write-only agent_id snapshot plumbing

Neither production caller reads ClaimedChannelTask.agent_id or the matching private snapshot field; Task.agent_id is already the authoritative persisted binding, while Telegram only consumes requested_agent_missing. Removing both fields, the existing-task-only task_agent_id assignment, and the two transfer assignments eliminates about eight production lines plus redundant test assertions without changing behavior.

rogercloud (approved):
Re-review complete for a4a59dc. All five prior findings are verified fixed, and no new correctness findings survived validation. Design verdict: acceptable-with-reservations. Focused Telegram/channel-runtime validation passed (86 tests), and Ruff passed on all five changed files. I left one non-blocking shrink suggestion inline. No linked issue was provided; the SaaS team-scope overlay was statically traced but not executed.

src/xagent/web/services/channel_runtime.py
OliverBryant: Applied in the latest commit — dropped `agent_id` from both claim snapshots along with the transfer assignments and the existing-task-only backfill; tests now assert the persisted `Task.agent_id` directly.

PR #1015, opened 2026-07-26
ref: add command policy foundation

AlexLiu190625:
## Summary

- Add a pure command-policy contract that can validate shell text and argv before process creation.
- Add stable exit-126 policy rejection, fail-closed validator error handling, and path-redacted caller errors to `CommandExecutorCore`.
- Add `TaskWorkspace.resolve_authorized_path()` as the single explicit-base canonical authorization primitive for workspace and allowed external paths.

## Scope

This PR is the foundation of the ordered implementation for [xagent-saas#171](https://github.com/xorbitsai/xagent-saas/issues/171). It intentionally does not include a shell parser, command classifiers, `bashlex`, execution-scope activation, production tool wiring, or claims about coverage of specific commands.

No in-tree production tool attaches a policy guard in this PR. Existing command behavior is unchanged unless an explicit `CommandPolicyGuard` is injected.

## Stack

The remaining work will be released sequentially after this foundation merges:

1. Shell language boundary and mutable shell-state handling.
2. File-command operand policy.
3. Runtime and nested-execution hardening.
4. Production scope activation and Docker end-to-end coverage.

Each PR will be reviewed and closed only against code reachable from its own head. Production activation remains last.

## Security boundary

This PR defines an application-level policy seam, not a sandbox. It provides no isolation by itself. Even after activation, cooperative command inspection remains defense in depth; eliminating unknown executable or alternate-runtime bypasses requires operating-system or container isolation.

## Validation

- `uv run pytest tests/core/tools/test_command_executor.py tests/core/test_workspace_path_containment.py tests/core/test_workspace_external_dirs.py tests/core/test_execution_scope_workspace.py`
- `uv run ruff check src/xagent/core/tools/core/command_executor.py src/xagent/core/tools/core/command_policy.py src/xagent/core/workspace.py tests/core/tools/test_command_executor.py tests/core/test_workspace_path_containment.py`
- `uv run ruff format --check src/xagent/core/tools/core/command_executor.py src/xagent/core/tools/core/command_policy.py src/xagent/core/workspace.py tests/core/tools/test_command_executor.py tests/core/test_workspace_path_containment.py`
- `uv run mypy src/xagent/core/tools/core/command_executor.py src/xagent/core/tools/core/command_policy.py src/xagent/core/workspace.py`
- `git diff --check upstream/main`

Issue #13, opened 2026-02-24
Files page mixes same-name files across agents/workspaces due to non-unique frontend identity


bluefish-08:
Problem
In the Files management page, clicking around (filtering/selecting/deleting) can make the list look duplicated or mixed across different agents/workspaces when files share the same filename.

Expected
Each file row should remain stable and isolated by its true identity (path/file id), even if filenames are the same.

Actual
Rows can appear to “grow”/mix after interactions. Selection/delete can target the wrong entry when multiple files have the same `filename`.

Reproduction
1. Create/upload files with the same name under different agent/workspace folders.
2. Open `/files`.
3. Click filters/select/delete or refresh list interactions multiple times.
4. Observe unstable row reuse/mixed display.
Suspected Root Cause
Frontend list identity currently relies on `filename` in critical places (row key / selection / delete matching), which is not unique across folders.

Suggestion
Use a stable unique identity from backend (preferably `file_id`, or `file_path`/`relative_path`) consistently for:
- row `key`
- selection state
- delete target mapping

Notes
Backend list API already returns path-level fields, so this may be fixable with frontend-only changes first.

PR #1032, opened 2026-07-27
ref: add Bash command policy boundary

AlexLiu190625:
## Summary

- add an opt-in `bashlex`-backed workspace command guard for Bash syntax, shell state, nested execution, wrappers, scripts, and redirections
- keep file-command classification deliberately minimal to `cat`, `rm`, redirections, and directory state
- bind guarded shell execution to Bash and validate guarded Bash scripts without temporary files

## Boundary

- dormant unless a caller explicitly injects the command guard
- unknown executable commands retain the current cooperative fail-open contract
- reuses `TaskWorkspace.resolve_authorized_path` as the path-authorization owner
- does not add `ExecutionScope`, tool-factory, preview, service, or SaaS activation
- provides defense in depth and is not an OS or container isolation boundary

This is the shell-language layer for #171, following #1015. Broader file-command classification, runtime hardening, and production activation remain in subsequent PRs.

## Verification

- `uv run pytest -q tests/core/tools/test_command_executor.py tests/core/tools/test_command_path_guard_bash.py tests/core/test_workspace_path_containment.py`
- `uv run mypy src/xagent/core/tools/core/command_executor.py src/xagent/core/tools/core/command_path_guard.py`
- `uv run ruff check pyproject.toml src/xagent/core/tools/core/command_executor.py src/xagent/core/tools/core/command_path_guard.py tests/core/tools/test_command_executor.py tests/core/tools/test_command_path_guard_bash.py`
- `uv run ruff format --check src/xagent/core/tools/core/command_executor.py src/xagent/core/tools/core/command_path_guard.py tests/core/tools/test_command_executor.py tests/core/tools/test_command_path_guard_bash.py`
- `uv lock --check`
- `uv build`

### DIFF
--- a/src/xagent/core/tools/core/command_executor.py
+++ b/src/xagent/core/tools/core/command_executor.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from typing import Any, Dict, Optional, cast
 
 from .command_policy import (
+    CommandArgvExecutionPolicy,
     CommandPathViolation,
     CommandPolicyGuard,
     CommandPolicyViolation,
@@ -207,7 +208,7 @@ class CommandExecutorCore:
             working_directory: Directory to use as working directory during execution
             path_guard: Optional policy implementation invoked before process creation
         """
-        self.path_guard = path_guard
+        self._path_guard = path_guard
         self._working_directory = _bind_policy_working_directory(
             working_directory,
             path_guard,
@@ -220,6 +221,11 @@ class CommandExecutorCore:
         if self._working_directory is None:
             return None
         return os.fspath(self._working_directory)
+
+    @property
+    def path_guard(self) -> Optional[CommandPolicyGuard]:
+        """Return the captured policy guard without exposing a mutation surface."""
+        return self._path_guard
 
     def execute_command(
         self,
@@ -252,14 +258,22 @@ class CommandExecutorCore:
         if shell and not isinstance(command, str):
             return _command_rejected_result("shell=True requires a string command")
 
-        if self.path_guard is not None:
+        argv_prepared_by_policy = False
+        if self._path_guard is not None:
             try:
                 if shell:
-                    self.path_guard.validate(cast(str, command))
+                    self._path_guard.validate(cast(str, command))
                 else:
                     argv = [command] if isinstance(command, str) else list(command)
-                    self.path_guard.validate_argv(argv)
-                    command = argv
+                    if isinstance(
+                        self._path_guard,
+                        CommandArgvExecutionPolicy,
+                    ):
+                        command = self._path_guard.prepare_argv_for_execution(argv)
+                        argv_prepared_by_policy = True
+                    else:
+                        self._path_guard.validate_argv(argv)
+                        command = argv
             except CommandPolicyViolation as exc:
                 return _command_rejected_result(exc)
             except Exception:
@@ -285,7 +299,7 @@ class CommandExecutorCore:
             "timeout": timeout,
             "cwd": self._working_directory,
         }
-        if self.path_guard is not None and shell:
+        if self._path_guard is not None and shell:
             try:
                 run_options["executable"] = os.fspath(
                     resolve_trusted_executable("bash")
@@ -293,8 +307,9 @@ class CommandExecutorCore:
             except CommandPolicyViolation as exc:
                 return _command_rejected_result(exc)
         elif (
-            self.path_guard is not None
+            self._path_guard is not None
             and not shell
+            and not argv_prepared_by_policy
             and isinstance(command, list)
             and command
             and os.path.basename(command[0]) == "bash"
@@ -376,7 +391,7 @@ class CommandExecutorCore:
         """
         timeout = _validate_timeout(timeout, self.timeout)
 
-        if self.path_guard is not None:
+        if self._path_guard is not None:
             return _command_rejected_result(
                 CommandPolicyViolation(
                     "restricted execute_script requires race-safe file semantics"
--- a/src/xagent/core/tools/core/command_executor.py
+++ b/src/xagent/core/tools/core/command_executor.py
@@ -7,8 +7,6 @@ Execute shell commands and scripts with proper controls.
 import logging
 import os
 import re
-import shlex
-import shutil
 import subprocess
 import tempfile
 from pathlib import Path
@@ -18,6 +16,8 @@ from .command_policy import (
     CommandPathViolation,
     CommandPolicyGuard,
     CommandPolicyViolation,
+    CwdBoundCommandPolicy,
+    resolve_trusted_executable,
 )
 
 logger = logging.getLogger(__name__)
@@ -31,16 +31,6 @@ TIMEOUT_EXIT_CODE = -999
 
 # Conventional shell exit code for a command found but not permitted to run.
 COMMAND_REJECTED_EXIT_CODE = 126
-
-
-def _resolve_policy_shell_executable() -> str:
-    """Return the Bash executable whose grammar is owned by the policy parser."""
-    executable = shutil.which("bash")
-    if executable is None:
-        raise CommandPolicyViolation(
-            "restricted command execution requires a Bash executable"
-        )
-    return executable
 
 
 def _command_rejected_result(reason: object) -> Dict[str, Any]:
@@ -155,6 +145,37 @@ def _validate_working_directory(working_directory: Optional[str]) -> None:
         )
 
 
+def _bind_policy_working_directory(
+    working_directory: Optional[str],
+    path_guard: Optional[CommandPolicyGuard],
+) -> str | Path | None:
+    """Bind execution to a cwd-aware policy while preserving legacy guards."""
+    if path_guard is None or not isinstance(path_guard, CwdBoundCommandPolicy):
+        return working_directory
+
+    policy_cwd = path_guard.execution_cwd
+    if policy_cwd is None:
+        return working_directory
+    if not policy_cwd.is_absolute():
+        raise ValueError("command policy execution cwd must be canonical and absolute")
+    try:
+        canonical_policy_cwd = policy_cwd.resolve(strict=True)
+    except (OSError, RuntimeError) as exc:
+        raise ValueError("command policy execution cwd must be accessible") from exc
+    if canonical_policy_cwd != policy_cwd:
+        raise ValueError("command policy execution cwd must be canonical and absolute")
+    _validate_working_directory(str(policy_cwd))
+
+    if working_directory:
+        _validate_working_directory(working_directory)
+        caller_cwd = Path(working_directory).expanduser().resolve(strict=True)
+        if caller_cwd != policy_cwd:
+            raise ValueError(
+                "working_directory does not match command policy execution cwd"
+            )
+    return policy_cwd
+
+
 def _sanitize_interpreter_suffix(interpreter: str) -> str:
     """
     Sanitize interpreter name for use as temp file suffix.
@@ -186,9 +207,19 @@ class CommandExecutorCore:
             working_directory: Directory to use as working directory during execution
             path_guard: Optional policy implementation invoked before process creation
         """
-        self.working_directory = working_directory
         self.path_guard = path_guard
+        self._working_directory = _bind_policy_working_directory(
+            working_directory,
+            path_guard,
+        )
         self.timeout = 300  # 5 minutes default
+
+    @property
+    def working_directory(self) -> Optional[str]:
+        """Return the configured cwd without exposing a mutation surface."""
+        if self._working_directory is None:
+            return None
+        return os.fspath(self._working_directory)
 
     def execute_command(
         self,
@@ -252,11 +283,27 @@ class CommandExecutorCore:
             "capture_output": capture_output,
             "text": True,
             "timeout": timeout,
-            "cwd": self.working_directory,
+            "cwd": self._working_directory,
         }
         if self.path_guard is not None and shell:
             try:
-                run_options["executable"] = _resolve_policy_shell_executable()
+                run_options["executable"] = os.fspath(
+                    resolve_trusted_executable("bash")
+                )
+            except CommandPolicyViolation as exc:
+                return _command_rejected_result(exc)
+        elif (
+            self.path_guard is not None
+            and not shell
+            and isinstance(command, list)
+            and command
+            and os.path.basename(command[0]) == "bash"
+        ):
+            try:
+                command = [
+                    os.fspath(resolve_trusted_executable(command[0])),
+                    *command[1:],
+                ]
             except CommandPolicyViolation as exc:
                 return _command_rejected_result(exc)
 
@@ -311,6 +358,11 @@ class CommandExecutorCore:
         """
         Execute script content with specified interpreter.
 
+        Guarded execution is rejected because this method's real-file semantics
+        cannot yet be preserved through the command policy without introducing
+        a validation-to-execution race. Unguarded execution retains its legacy
+        temporary-file behavior.
+
         Args:
             script_content: Script content to execute
             interpreter: Interpreter to use (bash, python, node, sh, etc.)
@@ -325,23 +377,10 @@ class CommandExecutorCore:
         timeout = _validate_timeout(timeout, self.timeout)
 
         if self.path_guard is not None:
-            try:
-                interpreter_argv = shlex.split(interpreter)
-            except ValueError:
-                return _command_rejected_result("invalid script interpreter")
-            if not interpreter_argv:
-                return _command_rejected_result("script interpreter is required")
-            if (
-                len(interpreter_argv) != 1
-                or os.path.basename(interpreter_argv[0]) != "bash"
-            ):
-                return _command_rejected_result(
-                    "restricted execute_script requires the Bash policy shell"
+            return _command_rejected_result(
+                CommandPolicyViolation(
+                    "restricted execute_script requires race-safe file semantics"
                 )
-            return self.execute_command(
-                [*interpreter_argv, "-c", script_content],
-                timeout=timeout,
-                shell=False,
             )
 
         try:
--- a/src/xagent/core/tools/core/command_executor.py
+++ b/src/xagent/core/tools/core/command_executor.py
@@ -7,6 +7,8 @@ Execute shell commands and scripts with proper controls.
 import logging
 import os
 import re
+import shlex
+import shutil
 import subprocess
 import tempfile
 from pathlib import Path
@@ -29,6 +31,16 @@ TIMEOUT_EXIT_CODE = -999
 
 # Conventional shell exit code for a command found but not permitted to run.
 COMMAND_REJECTED_EXIT_CODE = 126
+
+
+def _resolve_policy_shell_executable() -> str:
+    """Return the Bash executable whose grammar is owned by the policy parser."""
+    executable = shutil.which("bash")
+    if executable is None:
+        raise CommandPolicyViolation(
+            "restricted command execution requires a Bash executable"
+        )
+    return executable
 
 
 def _command_rejected_result(reason: object) -> Dict[str, Any]:
@@ -235,14 +247,23 @@ class CommandExecutorCore:
                 f"CommandExecutor: Using working directory: {self.working_directory}"
             )
 
+        run_options: dict[str, Any] = {
+            "shell": shell,
+            "capture_output": capture_output,
+            "text": True,
+            "timeout": timeout,
+            "cwd": self.working_directory,
+        }
+        if self.path_guard is not None and shell:
+            try:
+                run_options["executable"] = _resolve_policy_shell_executable()
+            except CommandPolicyViolation as exc:
+                return _command_rejected_result(exc)
+
         try:
             result = subprocess.run(
                 command,
-                shell=shell,
-                capture_output=capture_output,
-                text=True,
-                timeout=timeout,
-                cwd=self.working_directory,  # Use cwd parameter instead of os.chdir()
+                **run_options,
             )
 
             output = result.stdout if capture_output else ""
@@ -304,8 +325,23 @@ class CommandExecutorCore:
         timeout = _validate_timeout(timeout, self.timeout)
 
         if self.path_guard is not None:
-            return _command_rejected_result(
-                "script execution is unavailable under an injected command policy"
+            try:
+                interpreter_argv = shlex.split(interpreter)
+            except ValueError:
+                return _command_rejected_result("invalid script interpreter")
+            if not interpreter_argv:
+                return _command_rejected_result("script interpreter is required")
+            if (
+                len(interpreter_argv) != 1
+                or os.path.basename(interpreter_argv[0]) != "bash"
+            ):
+                return _command_rejected_result(
+                    "restricted execute_script requires the Bash policy shell"
+                )
+            return self.execute_command(
+                [*interpreter_argv, "-c", script_content],
+                timeout=timeout,
+                shell=False,
             )
 
         try:
--- a/src/xagent/core/tools/core/command_path_guard.py
+++ b/src/xagent/core/tools/core/command_path_guard.py
@@ -26,7 +26,17 @@ from typing import (
     cast,
 )
 
-import bashlex
+try:
+    import bashlex
+except ModuleNotFoundError as _bashlex_import_error:  # pragma: no cover
+    # ``bashlex`` ships in the optional ``command-path-guard`` extra, not the
+    # base dependencies. Import lazily so importing this module (and collecting
+    # its tests) never fails on a plain install; construction fails closed with
+    # an actionable hint when the parser is genuinely required.
+    bashlex = None
+    _BASHLEX_IMPORT_ERROR: ModuleNotFoundError | None = _bashlex_import_error
+else:
+    _BASHLEX_IMPORT_ERROR = None
 
 from ...workspace import TaskWorkspace
 from .command_policy import (
@@ -39,7 +49,11 @@ from .command_policy import (
 logger = logging.getLogger(__name__)
 
 _MAX_COMMAND_POLICY_INPUT_CHARS = 64 * 1024
-_MAX_INSPECTED_SCRIPT_BYTES = _MAX_COMMAND_POLICY_INPUT_CHARS
+# Bounded budget for reading script/config files off disk before inspection.
+# Numerically equal to the character input cap but kept in its own byte domain
+# (compared against ``os.fstat().st_size`` and ``len(raw_bytes)``) so the two
+# limits can move independently instead of silently coupling chars to bytes.
+_MAX_INSPECTED_SCRIPT_BYTES = 64 * 1024
 _MAX_INSPECTED_SCRIPT_DEPTH = 8
 _MAX_COMMAND_WRAPPER_DEPTH = 32
 _MAX_POSSIBLE_SHELL_STATES = 16
@@ -636,6 +650,11 @@ class WorkspaceCommandPathGuard:
     """Validate Bash language boundaries against one task workspace."""
 
     def __init__(self, workspace: TaskWorkspace) -> None:
+        if bashlex is None:
+            raise ModuleNotFoundError(
+                "WorkspaceCommandPathGuard requires the 'bashlex' parser. Install "
+                "the optional extra: pip install 'xagent[command-path-guard]'."
+            ) from _BASHLEX_IMPORT_ERROR
         self._workspace = workspace
         self._initial_cwd = workspace.resolve_path("").resolve()
 
@@ -657,12 +676,19 @@ class WorkspaceCommandPathGuard:
         ):
             return []
         policy_source = _normalize_policy_shell_source(command)
+        if bashlex is None:  # pragma: no cover - guarded at construction
+            raise CommandPolicyViolation("shell command parser is unavailable")
         try:
             nodes = cast(list[Any], bashlex.parse(policy_source))
         except Exception as exc:
             # bashlex exposes several parser-internal exception types. Normalize
             # all of them at this boundary so malformed input never reaches the
             # executor through a parser-version-specific failure mode.
+            #
+            # Invariant: input bashlex cannot model (``$(( ))``, ``[[ ]]``,
+            # ``case``, ``select``, ``coproc``, array assignments, ...) MUST stay
+            # fail-closed here. Never relax this into a fail-open path to reduce
+            # false positives on unparsable input.
             logger.debug("Command path guard rejected unparsed shell input")
             raise CommandPolicyViolation("cannot safely parse shell command") from exc
         _mark_unmodeled_expansions(nodes, policy_source)
@@ -899,11 +925,11 @@ class WorkspaceCommandPathGuard:
                 f"cannot safely inspect privilege escalation via {command_name}"
             )
         if self._is_direct_command_path(command_word):
-            if not self._is_trusted_system_command(command_word, command_name):
+            if not self._is_trusted_system_command(command_word):
                 self._inspect_direct_shell_script(command_word, state)
                 return state
         elif command_name in _CLASSIFIED_EXECUTABLE_COMMANDS:
-            if not self._is_trusted_system_command(command_word, command_name):
+            if not self._is_trusted_system_command(command_word):
                 discovered = shutil.which(command_name)
                 if discovered is not None:
                     self._inspect_direct_shell_script(discovered, state)
@@ -980,6 +1006,53 @@ class WorkspaceCommandPathGuard:
             )
         return home
 
+    @staticmethod
+    def _reject_symlink_traversing_chdir(target_word: str, cwd: Path) -> None:
+        """Reject a directory change whose target path crosses a symlink.
+
+        The guard tracks a *physically* resolved cwd, but Bash's ``cd``/``pushd``
+        (without ``-P``) are *logical*: they collapse ``..`` textually against
+        the pre-symlink path instead of following symlinks. When a target both
+        crosses a symlink and carries ``..`` segments, the guard's resolved cwd
+        diverges from where Bash actually lands, so a later relative operand can
+        resolve inside the workspace while Bash reads/writes a sibling tenant's
+        files. Physical and logical resolution agree whenever no symlink is
+        traversed, so rejecting symlink-crossing directory changes keeps the
+        tracked cwd equal to Bash's logical cwd. Fail closed rather than emulate
+        logical ``cd``. ``cd -``/``popd``/stack rotation restore an
+        already-validated, symlink-free cwd and do not reach this check.
+        """
+        if isinstance(target_word, _CommandValue) and not target_word.is_static:
+            # Dynamic targets are rejected by the resolver below; do not probe
+            # an unexpanded literal against the filesystem.
+            return
+        target = Path(target_word).expanduser()
+        if target.is_absolute():
+            probe = Path(target.anchor)
+            parts = target.parts[1:]
+        else:
+            probe = cwd
+            parts = target.parts
+        for part in parts:
+            if part == ".":
+                continue
+            if part == "..":
+                probe = probe.parent
+                continue
+            probe = probe / part
+            try:
+                crosses_symlink = probe.is_symlink()
+            except OSError as exc:
+                raise CommandPolicyViolation(
+                    "cannot inspect directory change target"
+                ) from exc
+            if crosses_symlink:
+                raise CommandPolicyViolation(
+                    "cannot safely resolve a directory change that traverses a "
+                    "symlink; Bash resolves 'cd' logically while the guard "
+                    "resolves physically"
+                )
+
     def _change_directory(
         self,
         command_name: str,
@@ -998,6 +1071,7 @@ class WorkspaceCommandPathGuard:
                     )
                 target = state.previous_cwd
             else:
+                self._reject_symlink_traversing_chdir(target_word, state.cwd)
                 target = self._check_path(target_word, state.cwd, "read")
             return replace(state, cwd=target, previous_cwd=state.cwd)
 
@@ -1012,6 +1086,7 @@ class WorkspaceCommandPathGuard:
             if operands:
                 if operands[0] == "-":
                     raise CommandPolicyViolation("cannot safely resolve pushd target")
+                self._reject_symlink_traversing_chdir(operands[0], state.cwd)
                 target = self._check_path(operands[0], state.cwd, "read")
                 stack = (state.cwd, *state.directory_stack)
             else:
@@ -1243,7 +1318,7 @@ class WorkspaceCommandPathGuard:
         return "/" in command_word
 
     @staticmethod
-    def _is_trusted_system_command(command_word: str, command_name: str) -> bool:
+    def _is_trusted_system_command(command_word: str) -> bool:
         try:
             resolve_trusted_executable(command_word)
             return True
@@ -1613,7 +1688,7 @@ class WorkspaceCommandPathGuard:
                 "instead of using active globs or shell expansions"
             )
 
-        if raw_path in {"", "-", "{}"}:
+        if raw_path in {"", "-"}:
             return cwd
 
         candidate = Path(raw_path).expanduser()
--- a/src/xagent/core/tools/core/command_path_guard.py
+++ b/src/xagent/core/tools/core/command_path_guard.py
@@ -1067,6 +1067,8 @@ class WorkspaceCommandPathGuard:
             )
         command_option_index: int | None = None
         for index, value in enumerate(literals[:-1]):
+            if value == "--":
+                break
             if (
                 value.startswith("-")
                 and not value.startswith("--")
@@ -1111,6 +1113,9 @@ class WorkspaceCommandPathGuard:
         index = 0
         while index < len(values):
             value = values[index]
+            if value == "--":
+                remaining.extend(values[index:])
+                break
             if value in _BASH_FILE_OPTIONS:
                 if index + 1 < len(values):
                     initialization_files.append(values[index + 1])
--- a/src/xagent/core/tools/core/command_path_guard.py
+++ b/src/xagent/core/tools/core/command_path_guard.py
@@ -182,6 +182,11 @@ _NO_FILESYSTEM_EFFECT_COMMANDS = {
     "declare",
     "typeset",
 }
+# Options that consume a following token as their value, per classified command,
+# so the value is not misread as a path operand (e.g. `mkdir -m 0755 dir`).
+_COMMAND_VALUE_OPTIONS = {
+    "mkdir": frozenset({"-m", "--mode"}),
+}
 _BASH_FILE_OPTIONS = {"--init-file", "--rcfile"}
 _BASH_LONG_FLAG_OPTIONS = {
     "--debug",
@@ -966,7 +971,12 @@ class WorkspaceCommandPathGuard:
         if command_name in _READ_COMMANDS:
             self._check_operands(args, state.cwd, "read")
         elif command_name in _WRITE_COMMANDS:
-            self._check_operands(args, state.cwd, "write")
+            self._check_operands(
+                args,
+                state.cwd,
+                "write",
+                value_options=_COMMAND_VALUE_OPTIONS.get(command_name, frozenset()),
+            )
         elif command_name in _SHELL_COMMANDS:
             self._check_nested_shell(command_name, args, state)
         elif command_name in _UNSUPPORTED_SHELL_COMMANDS:
@@ -1268,8 +1278,10 @@ class WorkspaceCommandPathGuard:
         values: Sequence[str],
         cwd: Path,
         access: PathAccess,
+        *,
+        value_options: frozenset[str] = frozenset(),
     ) -> None:
-        for raw_path in self._operands(values):
+        for raw_path in self._operands(values, value_options=value_options):
             self._check_path(raw_path, cwd, access)
 
     def _read_policy_script(self, raw_path: str, cwd: Path) -> str:
@@ -1664,14 +1676,29 @@ class WorkspaceCommandPathGuard:
             raise CommandPolicyViolation(f"cannot resolve dynamic {context}")
 
     @staticmethod
-    def _operands(values: Sequence[str]) -> list[str]:
+    def _operands(
+        values: Sequence[str],
+        *,
+        value_options: frozenset[str] = frozenset(),
+    ) -> list[str]:
         operands: list[str] = []
         options_done = False
+        skip_next_value = False
         for value in values:
+            if skip_next_value:
+                # Separated value of a preceding value-consuming option.
+                skip_next_value = False
+                continue
             if not options_done and value == "--":
                 options_done = True
                 continue
             if not options_done and value.startswith("-") and value != "-":
+                # An option that consumes a following token must not let that
+                # value be misread as a path operand. Attached forms (`-m0755`,
+                # `--mode=0755`) are already dropped whole; only a separated
+                # value needs to be skipped explicitly.
+                if value in value_options:
+                    skip_next_value = True
                 continue
             operands.append(value)
         return operands
--- a/src/xagent/core/tools/core/command_path_guard.py
+++ b/src/xagent/core/tools/core/command_path_guard.py
@@ -33,6 +33,7 @@ from .command_policy import (
     CommandPathViolation,
     CommandPolicyViolation,
     PathAccess,
+    resolve_trusted_executable,
 )
 
 logger = logging.getLogger(__name__)
@@ -180,15 +181,6 @@ _SAFE_DEVICE_PATHS = {
     Path("/dev/stdout"),
     Path("/dev/stderr"),
 }
-_TRUSTED_EXECUTABLE_ROOTS = tuple(
-    path.resolve()
-    for path in (
-        Path("/bin"),
-        Path("/usr/bin"),
-        Path("/usr/local/bin"),
-        Path("/opt/homebrew/bin"),
-    )
-)
 _POLICY_TIME_WRAPPER = "__t_"
 
 
@@ -644,6 +636,11 @@ class WorkspaceCommandPathGuard:
         self._workspace = workspace
         self._initial_cwd = workspace.resolve_path("").resolve()
 
+    @property
+    def execution_cwd(self) -> Path:
+        """Return the canonical directory whose paths this guard validates."""
+        return self._initial_cwd
+
     @staticmethod
     def _parse_shell(command: str) -> list[Any]:
         if len(command) > _MAX_COMMAND_POLICY_INPUT_CHARS:
@@ -878,13 +875,11 @@ class WorkspaceCommandPathGuard:
                 self._inspect_direct_shell_script(command_word, state)
                 return state
         elif command_name in _CLASSIFIED_EXECUTABLE_COMMANDS:
-            discovered = shutil.which(command_name)
-            if discovered is not None and not self._is_trusted_system_command(
-                discovered,
-                command_name,
-            ):
-                self._inspect_direct_shell_script(discovered, state)
-                return state
+            if not self._is_trusted_system_command(command_word, command_name):
+                discovered = shutil.which(command_name)
+                if discovered is not None:
+                    self._inspect_direct_shell_script(discovered, state)
+                    return state
 
         if command_name in {"cd", "pushd", "popd"}:
             return self._change_directory(command_name, args, state)
@@ -1219,18 +1214,10 @@ class WorkspaceCommandPathGuard:
 
     @staticmethod
     def _is_trusted_system_command(command_word: str, command_name: str) -> bool:
-        candidate = Path(command_word).expanduser()
-        if not candidate.is_absolute():
-            return False
-        discovered = shutil.which(command_name)
-        if discovered is None:
-            return False
         try:
-            resolved = candidate.resolve()
-            return resolved == Path(discovered).resolve() and any(
-                resolved.is_relative_to(root) for root in _TRUSTED_EXECUTABLE_ROOTS
-            )
-        except (OSError, RuntimeError):
+            resolve_trusted_executable(command_word)
+            return True
+        except CommandPolicyViolation:
             return False
 
     def _inspect_direct_shell_script(
--- a/src/xagent/core/tools/core/command_path_guard.py
+++ b/src/xagent/core/tools/core/command_path_guard.py
@@ -11,7 +11,6 @@ from __future__ import annotations
 import errno
 import logging
 import os
-import re
 import shutil
 import stat
 from contextlib import contextmanager
@@ -38,10 +37,10 @@ from .command_policy import (
 
 logger = logging.getLogger(__name__)
 
-_MAX_INSPECTED_SCRIPT_BYTES = 1024 * 1024
+_MAX_COMMAND_POLICY_INPUT_CHARS = 64 * 1024
+_MAX_INSPECTED_SCRIPT_BYTES = _MAX_COMMAND_POLICY_INPUT_CHARS
 _MAX_INSPECTED_SCRIPT_DEPTH = 8
 _MAX_COMMAND_WRAPPER_DEPTH = 32
-_MAX_COMMAND_POLICY_INPUT_CHARS = 64 * 1024
 _MAX_POSSIBLE_SHELL_STATES = 16
 _MAX_POLICY_PARSE_ATTEMPTS = 64
 _MAX_POLICY_SOURCE_CHARS = 1024 * 1024
@@ -147,7 +146,27 @@ _READ_COMMANDS = {"cat"}
 _WRITE_COMMANDS = {"rm"}
 _SHELL_COMMANDS = {"bash"}
 _UNSUPPORTED_SHELL_COMMANDS = {"dash", "sh", "zsh"}
+_UNSUPPORTED_PRIVILEGE_COMMANDS = {"sudo"}
 _BASH_FILE_OPTIONS = {"--init-file", "--rcfile"}
+_BASH_LONG_FLAG_OPTIONS = {
+    "--debug",
+    "--debugger",
+    "--dump-po-strings",
+    "--dump-strings",
+    "--help",
+    "--login",
+    "--noediting",
+    "--noprofile",
+    "--norc",
+    "--posix",
+    "--pretty-print",
+    "--protected",
+    "--restricted",
+    "--verbose",
+    "--version",
+    "--wordexp",
+}
+_BASH_SHORT_FLAG_OPTIONS = frozenset("abefhiklmnprstuvxBCDHP")
 _IMPLICIT_SHELL_ENVIRONMENT = {"BASH_ENV", "CDPATH", "ENV", "ZDOTDIR"}
 _REJECTED_SHELL_ASSIGNMENTS = {
     *_IMPLICIT_SHELL_ENVIRONMENT,
@@ -170,14 +189,6 @@ _TRUSTED_EXECUTABLE_ROOTS = tuple(
         Path("/opt/homebrew/bin"),
     )
 )
-_QUOTED_HEREDOC_PATTERN = re.compile(
-    r"(?m)(?P<operator><<-?)(?P<space>[ \t]*)(?P<quote>['\"])"
-    r"(?P<delimiter>[A-Za-z_][A-Za-z0-9_]*)(?P=quote)(?=[ \t]*$)"
-)
-_TIME_KEYWORD_PATTERN = re.compile(
-    r"(?m)(?P<prefix>^|(?<=[;\n])|(?<=&&)|(?<=\|\|))"
-    r"(?P<space>[ \t]*)time(?=[ \t]+)"
-)
 _POLICY_TIME_WRAPPER = "__t_"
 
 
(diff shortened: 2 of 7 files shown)


src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1337 @@
...
+        first_line = script.splitlines()[0] if script.splitlines() else ""
+        if first_line.startswith("#!"):
+            interpreter = os.path.basename(first_line[2:].strip().split()[0])
+            if interpreter == "env":
+                interpreter_parts = first_line[2:].strip().split()
+                interpreter = (
+                    os.path.basename(interpreter_parts[1])
+                    if len(interpreter_parts) > 1
+                    else ""
+                )
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in 6ad0a416. The guard now tokenizes the shebang once and maps a missing interpreter to the existing rejected-dialect path, so validation raises CommandPolicyViolation instead of IndexError. Parameterized regression coverage for `#!` and `#!   ` now exercises the real validate_argv-to-executor path and verifies stable exit 126 without the generic validation error.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1339 @@
...
+        body_start += 1
+        delimiter = match.group("delimiter")
+        strip_tabs = match.group("operator") == "<<-"
+        cursor = body_start
+        while cursor <= len(command):
+            line_end = command.find("\n", cursor)
+            if line_end < 0:
+                line_end = len(command)
+            line = command[cursor:line_end]
+            comparable = line.lstrip("\t") if strip_tabs else line
```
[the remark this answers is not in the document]
AlexLiu190625: Verified against real Bash behavior. With all-CRLF input, Bash treats the carriage return as part of the delimiter while the policy parser rejects the command; with an LF opener and CRLF closing line, Bash treats the remainder as here-document data and the guard also rejects it. There is no path where a post-heredoc command executes after being masked. Commit 34302b1b adds parameterized fail-closed regressions for both forms. Stripping only the closing line would diverge from Bash delimiter semantics, so the parser behavior remains unchanged.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1339 @@
...
+        script_path = self._check_path(raw_path, cwd, "read")
+        written_paths = _COMMAND_WRITTEN_PATHS.get()
+        if written_paths is not None and script_path in written_paths:
+            raise CommandPolicyViolation(
+                "cannot inspect a script modified earlier in the same command"
+            )
+
+        descriptor: int | None = None
+        try:
+            flags = os.O_RDONLY | os.O_NONBLOCK | os.O_NOFOLLOW
```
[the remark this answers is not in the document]
AlexLiu190625: Addressed in 34302b1b without weakening the required open semantics. Secure direct-script inspection now requires typed O_NONBLOCK and O_NOFOLLOW flags; when either capability is unavailable, the guard raises CommandPolicyViolation and the executor returns the stable policy rejection code 126. The implementation intentionally does not fall back to zero, which would permit blocking FIFO opens and final-component symlink following. Parameterized regressions simulate each missing flag through the real guard-to-executor path.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1349 @@
...
+            )
+        command_option_index: int | None = None
+        for index, value in enumerate(literals[:-1]):
+            if (
+                value.startswith("-")
+                and not value.startswith("--")
+                and "c" in value[1:]
+            ):
+                command_option_index = index
+                break
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in bd646032. Nested-shell short-option scanning now stops at a literal double dash, and Bash file-option extraction preserves every later token for script and positional-argument parsing. Regression coverage verifies that an unsafe workspace script named -c is inspected and rejected with return code 126, while safe dash-prefixed scripts are not falsely rejected. The focused regressions and the full PR CI pass.

tests/core/tools/test_command_path_guard_bash.py
```diff
@@ -0,0 +1,1144 @@
...
+        ],
+    )
+    def test_rejects_dynamic_nested_command_names(
+        self, scoped_command_workspace, command_template
+    ):
+        workspace, _, _ = scoped_command_workspace
+        guard = WorkspaceCommandPathGuard(workspace)
+
+        with pytest.raises(CommandPolicyViolation):
+            guard.validate(command_template)
```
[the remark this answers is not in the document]
AlexLiu190625: Covered in bd646032. The regression is parameterized for -c and --rcfile script operands after the option terminator, verifies that safe scripts accept an out-of-workspace-looking positional argument without a false rejection, and separately verifies that an unsafe -c script is denied with return code 126 and cannot expose sibling data. The focused regressions and the full PR CI pass.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+                line_end = len(command)
+            line = command[cursor:line_end]
+            comparable = line.lstrip("\t") if strip_tabs else line
+            if comparable == delimiter:
+                break
+            for index in range(cursor, line_end):
+                normalized[index] = "x"
+            if line_end == len(command):
+                break
+            cursor = line_end + 1
```
rogercloud: **Unresolved security-critical thread — needs an explicit answer before merge.** I re-ran the reported nested-heredoc PoC and several variants (different delimiter names, quoted/unquoted, with/without trailing terminator, same-delimiter reuse) against `validate()`: every variant either failed closed with `CommandPolicyViolation: cannot safely parse shell command` or was correctly rejected as a path violation — I could not reproduce an exploitable bypass. But the smell is real: the nested-heredoc masking loop scans the unmodified original string index-independently and does corrupt the outer terminator; safety comes only from that corruption incidentally causing a parse error, which is not a designed invariant. Please (1) reply/resolve this thread explicitly, (2) apply the `last_heredoc_end` tracking fix anyway, and (3) add a regression test pinning the current fail-closed behavior for this exact input.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+            return state
+
+        command_word = self._command_value(words[0])
+        if not command_word.is_static:
+            raise CommandPolicyViolation("cannot resolve dynamic command name")
+        args = self._command_values(words[1:])
+
+        return self._validate_command_values(command_word, args, state)
+
+    def _validate_command_values(
```
rogercloud, line 936: **Major: script inspection defeated by any unclassified write command.** Only `cat`/`rm` operands and redirections are checked; every other executable returns early *and* never registers its destination in `_COMMAND_WRITTEN_PATHS`. `printf 'cat <sibling-secret>\n' > evil.sh && cp evil.sh run.sh && bash run.sh` reproducibly succeeds end-to-end through a real `CommandExecutorCore` and leaks the sibling file — `cp` is an untracked writer, so the TOCTOU protection built for redirections is bypassed. Track written paths for all write-capable commands, or treat any path a preceding command in the same state chain touched as potentially written.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+    @staticmethod
+    def _reject_implicit_shell_environment(values: Sequence[str]) -> None:
+        for value in values:
+            name, _, _ = str(value).partition("=")
+            if name in _REJECTED_SHELL_ASSIGNMENTS:
+                raise CommandPolicyViolation(
+                    f"cannot safely inspect shell execution environment via {name}"
+                )
+
+    def _check_nested_shell(
```
rogercloud, line 1402: **Major: `$VAR` in the option region bypasses `--rcfile`/`--init-file` checks.** `_check_command_wrapper` (~L814) calls `_reject_dynamic_values`; `_check_nested_shell` does not. `bash --rcfile <sibling>/x.rc -i -c exit` correctly raises `CommandPathViolation`, but `OPTS="--rcfile <sibling>/x.rc -i"; bash $OPTS -c exit` and `bash "$OPTS" -c exit` both pass — same out-of-workspace target, smuggled through a variable. Apply `_reject_dynamic_values` to the pre-`-c` option region here too.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+                    f"{active_environment}"
+                )
+            nodes = self._parse_shell(command)
+            states: tuple[_ShellState, ...] = (_ShellState(cwd=self._initial_cwd),)
+            for node in nodes:
+                states = self._validate_node_states(node, states)
+        finally:
+            _COMMAND_WRITTEN_PATHS.reset(write_token)
+
+    def _validate_node_states(
```
rogercloud, line 735: **Major: exponential validation cost with no work budget, evaluated before any timeout.** A 4-branch `cd a || cd b || cd c || cd d` isn't deduplicated (distinct cwd per branch) and fans out to ~4 states; nesting that through `bash <script>.sh` chains within `_MAX_INSPECTED_SCRIPT_DEPTH = 8` re-reads and re-parses each level once per state from the level above. Measured ~4.1-5.3x per added depth, 54s for one `validate()` call at depth 8 from an ~8-line script chain. `command_executor.py:227` runs this before `subprocess.run(..., timeout=...)` at L264, so nothing bounds it. Add a global node/parse-count or wall-clock budget — the four per-dimension limits compose multiplicatively.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+
+        script = self._shell_script_operand(file_values)
+        if script is None:
+            raise CommandPolicyViolation(
+                f"cannot inspect {command_name} input without command text or a script"
+            )
+        self._inspect_shell_script(script, state, propagate_state=False)
+
+    @staticmethod
+    def _extract_bash_file_options(
```
rogercloud: **Major: fail-open when `--rcfile`/`--init-file` is the last token.** The "has a following value" check is false, nothing is appended, and the loop silently advances instead of rejecting. Both `bash --rcfile -c 'printf safe'` and `bash --init-file -c 'printf safe' <sibling-secret-path>` are allowed — the second is a genuine bypass, since the sibling path goes unchecked because the malformed option list was accepted. Raise a policy violation for any recognized file option with no argument.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+from ...workspace import TaskWorkspace
+from .command_policy import (
+    CommandPathViolation,
+    CommandPolicyViolation,
+    PathAccess,
+)
+
+logger = logging.getLogger(__name__)
+
+_MAX_INSPECTED_SCRIPT_BYTES = 1024 * 1024
```
rogercloud: Minor: `_MAX_INSPECTED_SCRIPT_BYTES` (1 MiB) is unreachable: `_read_policy_script` passes content to `_parse_shell`, which rejects anything over `_MAX_COMMAND_POLICY_INPUT_CHARS` (64 KiB, L42) with a generic "command is too large to inspect" — misleading for a legitimate 100 KB script. Align the constants or give script inspection its own limit and message.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+        Path("/usr/bin"),
+        Path("/usr/local/bin"),
+        Path("/opt/homebrew/bin"),
+    )
+)
+_QUOTED_HEREDOC_PATTERN = re.compile(
+    r"(?m)(?P<operator><<-?)(?P<space>[ \t]*)(?P<quote>['\"])"
+    r"(?P<delimiter>[A-Za-z_][A-Za-z0-9_]*)(?P=quote)(?=[ \t]*$)"
+)
+_TIME_KEYWORD_PATTERN = re.compile(
```
rogercloud: Minor: `_TIME_KEYWORD_PATTERN` is not quote-aware, unlike the heredoc-masking branch which checks `_is_unquoted_on_current_line`. Confirmed: `_normalize_policy_shell_source("cat 'a\ntime b'")` -> `"cat 'a\n__t_ b'"`, so the guard validates a different literal than bash executes. Not exploitable today (filename-like literals only, same containment verdict either way), but it's the exact validate/execute divergence this module exists to remove. Reuse `_is_unquoted_on_current_line`.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+from pathlib import Path
+from typing import (
+    Any,
+    Iterable,
+    Sequence,
+    SupportsIndex,
+    cast,
+)
+
+import bashlex
```
rogercloud: Minor: `import bashlex` is unconditional, but `bashlex` is only in the `command-path-guard`/`all` extras and the `test` dependency group — not base dependencies — with no `try/except ImportError`. A plain `pip install -e .` makes `tests/core/tools/test_command_path_guard_bash.py` fail at collection time rather than skip. Fine while the guard is dormant, but the extra must be wired into the build/release image before any PR activates it.

src/xagent/core/tools/core/command_executor.py
```diff
@@ -304,8 +325,23 @@ def execute_script(
         timeout = _validate_timeout(timeout, self.timeout)
```
rogercloud, line 392: Minor: the guarded `execute_script` path does avoid temp files as claimed (`execute_command(["bash", "-c", script], shell=False)`, no `NamedTemporaryFile`), but it changes script semantics: `$0` becomes `"bash"` instead of the script path, and there are no positional args unless separately appended. Scripts relying on `$0` / `BASH_SOURCE` / `$1…` will behave differently depending on whether a guard is injected. Worth a one-line docstring note.

src/xagent/core/tools/core/command_executor.py
```diff
@@ -31,6 +33,16 @@
 COMMAND_REJECTED_EXIT_CODE = 126
 
 
+def _resolve_policy_shell_executable() -> str:
```
rogercloud: Minor: `_resolve_policy_shell_executable` (new here) resolves bash via `shutil.which("bash")` without the `_TRUSTED_EXECUTABLE_ROOTS` check that `command_path_guard.py`'s `_is_trusted_system_command` (~L1001-1015) applies to a `bash` argument found inside a command string. Real inconsistency introduced by this PR — previously `shell=True` used Python's hardcoded `/bin/sh`, not a PATH lookup. Exploiting it needs an already-poisoned ambient `PATH` (host-level compromise, outside what a guarded command string can achieve, and covered by the module's "not an OS security boundary" disclaimer), so this is hardening rather than an exploitable gap. Apply the same trusted-roots check for consistency.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1354 @@
...
+    script_depth: int = 0
+    wrapper_depth: int = 0
+
+
+@dataclass(frozen=True)
+class _CommandWrapperGrammar:
+    flag_options: frozenset[str] = frozenset()
+    value_options: frozenset[str] = frozenset()
+    attached_short_value_options: frozenset[str] = frozenset()
+    leading_scalars: int = 0
```
rogercloud: Simplification (`yagni`): `_CommandWrapperGrammar.leading_scalars: int` supports an arbitrary count of leading positional scalars, but only the `timeout` entry (L451) sets it, always to `1`; nothing else anticipates more. Replace the `int` field and the `for _ in range(...)` consumer loop (L848) with a `bool` (e.g. `has_duration_arg`) and a single conditional skip — same behavior, less machinery (~-5 lines).

rogercloud (changes requested):
## Summary

This PR adds `src/xagent/core/tools/core/command_path_guard.py` (~1354 lines), a `bashlex`-backed `WorkspaceCommandPathGuard` that layers shell-language understanding on top of the `CommandPolicyGuard` foundation from #1015, so that Bash command strings can be inspected for out-of-workspace path access before execution (shared-tenant isolation, xagent-saas#171). `command_executor.py` is changed so that guarded script execution is routed through `bash -c` instead of a temp file, and a 1183-line test module covers the new semantics. The guard is **dormant**: no production call site passes `path_guard=`, only tests do — activation is deferred to a later PR.

---

## Design verdict: acceptable with reservations, leaning over-engineered for what it currently buys

1354 lines of careful Bash-semantics emulation currently guard a two-command allowlist — `cat` (read) and `rm` (write) — while `cp`, `mv`, `head`, `tee`, `dd`, `grep`, `ln -s`, `find -exec`, `awk`, `python -c`, `perl -e` and everything else walk through unvalidated. That fall-through is deliberate and documented (the "cooperative fail-open contract" for unrecognized commands, pinned by `test_unknown_commands_keep_the_explicit_fail_open_contract`), so it is an accepted tradeoff rather than a hidden gap — but it is worth discussing explicitly, because it sets the ceiling on what all this machinery can actually deliver today. More concerning: two of the controls the module *does* implement are bypassable (Major 1 and Major 2 below), i.e. the enumerated allowlist itself is not airtight.

Given that complexity/benefit ratio, it is worth asking whether a much smaller design — a strict allowlist plus **fail-closed** on unknown commands, with no `bashlex` dependency and no Bash-semantics emulation — would deliver more real isolation for a fraction of the review surface and long-term maintenance cost. This is not a blocker given the explicitly staged rollout, but the question should get an answer on the record before the next PR turns the guard on.

Three design-level points that no single line owns:

**D2 — the guard's cwd and the executor's cwd are derived independently.** `WorkspaceCommandPathGuard.__init__` (`command_path_guard.py:474-476`) and `CommandExecutorCore.__init__` (`command_executor.py:176-191`) each compute a working directory from their own inputs; nothing at runtime ties them together. Every relative-path decision in the guard silently assumes they agree, and that agreement is enforced only by caller discipline — today, only test callers exist. Please make this a structural invariant (guard derives its cwd from the executor, or asserts equality at construction) before the guard is wired into production.

**D3 — `sudo` is modeled as a permitted wrapper.** At `command_path_guard.py:369` `sudo` uses the same `_CommandWrapperGrammar` mechanism as `env`/`timeout`/`nice`, i.e. the inner command is unwrapped and path-checked, and execution is allowed. That is qualitatively different from `env`: `sudo` changes the privilege context. `eval`, `chroot`, `sh`, `dash`, `zsh` are hard-rejected; `sudo` arguably belongs in that set. If the target deployment guarantees `sudo` is absent or inert, that assumption should be stated in a comment.

**Documentation gap.** Nothing in README, project CLAUDE.md, `example.env`, `docs/`, `.github/`, or the Dockerfiles mentions this guard, `CommandPolicyGuard`, or the new `bashlex` dependency. For a security-boundary module of this size, that is worth fixing — it also ties into Minor 7 below.

---

## Findings

### Unresolved prior thread (security-critical label) — needs an explicit resolution before merge

`src/xagent/core/tools/core/command_path_guard.py:231`

There is an open, never-replied-to review thread (gemini-code-assist, security-critical) claiming that `_normalize_policy_shell_source` mishandles nested heredoc delimiters, so that a command masking `rm -rf /` behind a nested `<<'EOF2'` pattern could hide it from bashlex while real bash executes it.

I re-investigated this specifically for this review and could not reproduce it as an exploitable bypass. I ran the exact PoC plus several variants directly against `WorkspaceCommandPathGuard.validate()`. In every variant the guard either (a) failed closed with `CommandPolicyViolation: cannot safely parse shell command` — the corrupted terminator makes bashlex's parse of the normalized text fail, and parse failures are converted into a policy rejection — or (b) still saw the live command and rejected it as a path violation. No variant produced "validate passes and the dangerous command is hidden".

That said, this should not be closed as a false alarm: the underlying code smell is real. The nested-heredoc masking loop scans the unmodified original string index-independently and does corrupt the outer terminator. The only reason this ends up safe is that the corruption happens to cause a parse error, which happens to fail closed — that is incidental, not a designed invariant.

Recommended actions:
1. The author or a maintainer should explicitly verify and reply/resolve this thread before merge.
2. Apply the suggested `last_heredoc_end` tracking fix anyway, for defense in depth and clarity.
3. Add a regression test pinning the current fail-closed behavior for this exact input, so a future change cannot silently turn it into a fail-open.

### Major 1 — Script inspection is defeated by any unclassified write command (validate-then-execute content divergence)

`src/xagent/core/tools/core/command_path_guard.py:655` (`_validate_command_values`), with `_read_policy_script` (~936-974) and `_COMMAND_WRITTEN_PATHS` (~1319-1354)

`_validate_command_values` only path-checks operands when `command_name` is in `_READ_COMMANDS = {"cat"}` or `_WRITE_COMMANDS = {"rm"}` (plus redirections, which do register into `_COMMAND_WRITTEN_PATHS`). Every other executable — `cp`, `mv`, `tee`, … — falls through with zero path checking, and never registers its destination as a written path. So the guard's script-inspection TOCTOU protection is bypassed by any writer it doesn't enumerate:

```
printf 'cat <sibling-secret-path>\n' > evil.sh && cp evil.sh run.sh && bash run.sh
```

End-to-end through a real `CommandExecutorCore` this reproducibly succeeds and leaks the sibling secret. `run.sh` is a pre-existing benign script; `cp` (an untracked writer) overwrites it; `bash run.sh` then inspects and executes the new content, and the guard never rejects.

Fix: register written paths for all write-capable commands — or treat any path touched as an argument by a preceding command in the same shell-state chain as potentially written, rather than only `rm` targets and redirect targets.

### Major 2 — Dynamic `$VAR` content in a `bash` option region bypasses `--rcfile`/`--init-file` inspection

`src/xagent/core/tools/core/command_path_guard.py:1056` (`_check_nested_shell`); compare `_check_command_wrapper` at ~814

`_check_command_wrapper` calls `_reject_dynamic_values` on its arguments; `_check_nested_shell` never does so for the pre-`-c` option region, and `_extract_bash_file_options` only recognizes literal `--rcfile`/`--init-file` tokens, silently dropping anything else. Reproduced:

- `bash --rcfile <sibling>/x.rc -i -c exit` → correctly raises `CommandPathViolation`
- `OPTS="--rcfile <sibling>/x.rc -i"; bash $OPTS -c exit` → no violation raised
- `bash "$OPTS" -c exit` (double-quoted variant) → no violation raised

Same out-of-workspace `--rcfile` target, merely smuggled through a shell variable. This is a working, directly reproducible evasion of a control the guard explicitly enforces in literal form.

Fix: apply `_reject_dynamic_values` to the pre-`-c` option region in `_check_nested_shell`, exactly as the wrapper-command path already does.

### Major 3 — Validation cost grows exponentially with attacker-controlled input, with no work budget, before any timeout applies

`src/xagent/core/tools/core/command_path_guard.py:518` (`_validate_node_states`/`_validate_list_states`), interacting with per-state script re-inspection in `_check_nested_shell`/`_read_policy_script` (~936-995); `command_executor.py:227` calls `path_guard.validate()` before `subprocess.run(..., timeout=...)` at line 264.

A `||`-chained `cd` with four branches (`cd d0 || cd d1 || cd d2 || cd d3`) is not collapsed by state deduplication — each branch has a distinct `cwd` — so it fans out to ~4 live shell states. Wrapping that in a chain of nested `bash <next-script>.sh` calls, well inside the documented `_MAX_INSPECTED_SCRIPT_DEPTH = 8`, makes each level's script independently re-read and re-parsed once per active state from the level above. Measured: ~4.1-5.3x wall-clock growth per added depth, reaching 54 seconds for a single `guard.validate()` call at depth 8 — from an ~8-line, hand-writable script chain. Textbook `O(branch_factor^depth)`.

Because validation is synchronous and runs before the subprocess timeout is armed, this hangs command execution with no bound at all.

Fix: add a global work budget — node/parse count or wall-clock — independent of the per-dimension bounds, since they compose multiplicatively.

### Major 4 — `_extract_bash_file_options` fails open when `--rcfile`/`--init-file` is the last token

`src/xagent/core/tools/core/command_path_guard.py:1108` (~1108-1131)

When `--rcfile`/`--init-file` is the final value in the option list, the "has a following value" check is false, nothing is appended, and the loop silently advances past it rather than rejecting the malformed input. Reproduced against the guard:

- `bash --rcfile -c 'printf safe'` → allowed
- `bash --init-file -c 'printf safe' <sibling-secret-path>` → allowed

The second case is a genuine fail-open: the sibling path argument, which the guard's own path-authorization logic should catch, goes unchecked because the malformed option list was silently accepted. Raise a policy violation for any recognized file option that has no argument.

### Minor 5 — `_MAX_INSPECTED_SCRIPT_BYTES` is unreachable, and the resulting error message is misleading

`src/xagent/core/tools/core/command_path_guard.py:39`

`_MAX_INSPECTED_SCRIPT_BYTES` is 1 MiB, but `_read_policy_script` hands content straight to `_parse_shell`, which rejects anything over `_MAX_COMMAND_POLICY_INPUT_CHARS` (64 KiB, line 42) with a generic "command is too large to inspect". A legitimate 100 KB script therefore fails with a message about a "command". Align the two constants, or give script inspection its own limit and message.

### Minor 6 — `_TIME_KEYWORD_PATTERN` is not quote-aware

`src/xagent/core/tools/core/command_path_guard.py:80` (~80-84)

Unlike the heredoc-masking branch, which checks `_is_unquoted_on_current_line`, the `time` keyword rewrite has no quote awareness. Confirmed: `_normalize_policy_shell_source("cat 'a\ntime b'")` yields `"cat 'a\n__t_ b'"` — the guard validates a different literal than the one real bash will use. Today this only affects filename-like literal content and both readings get the same containment verdict, so it is not exploitable, but it is exactly the validate/execute divergence class this module exists to eliminate. Reuse `_is_unquoted_on_current_line` here.

### Minor 7 — unconditional `import bashlex` against an optional extra

`src/xagent/core/tools/core/command_path_guard.py:28`

`bashlex` appears only in the `command-path-guard` and `all` extras and the `test` dependency group in `pyproject.toml`, not in base dependencies, and the import has no `try/except ImportError`. A plain `pip install -e .` therefore breaks `tests/core/tools/test_command_path_guard_bash.py` at collection time rather than skipping per test. Not urgent while the guard is dormant, but the extra must be wired into the build/release image before any PR activates the guard.

### Minor 8 — the guarded `execute_script` path changes script semantics

`src/xagent/core/tools/core/command_executor.py:325` (~325-345)

The guarded path genuinely avoids temp files as claimed — it routes to `execute_command(["bash", "-c", script], shell=False)`, with no `NamedTemporaryFile` — but that also changes semantics versus the unguarded temp-file path: `$0` becomes `"bash"` rather than the script path, and there are no positional args unless separately appended. Worth a one-line docstring note.

### Minor 9 — guarded interpreter resolved via `shutil.which` without the trusted-roots check

`src/xagent/core/tools/core/command_executor.py:36` (~36-43)

`_resolve_policy_shell_executable` (new in this PR) resolves the guarded path's bash via `shutil.which("bash")` with no check against `_TRUSTED_EXECUTABLE_ROOTS`, the allowlist that `command_path_guard.py`'s `_is_trusted_system_command` (~1001-1015) applies when validating a `bash` argument inside a command string. Before this PR, `shell=True` with no override used Python's hardcoded `/bin/sh`, not a PATH lookup, so this PR does add a new PATH-trust dependency for the interpreter itself. Exploiting it requires an already-poisoned ambient `PATH`, which is outside what a guarded command string can bring about on its own — so this is hardening rather than an exploitable gap. Apply the same `_TRUSTED_EXECUTABLE_ROOTS` check for consistency.

---

- `src/xagent/core/tools/core/command_path_guard.py` L285, L451, L848: `yagni` `_CommandWrapperGrammar.leading_scalars: int` generalizes to an arbitrary count of leading positional scalars, but only the `timeout` grammar entry (L451) ever sets it, always to `1`; no other entry, test, or comment anticipates more. Replace the `int` field and the `for _ in range(...)` consumer loop (L848) with a `bool` (e.g. `has_duration_arg`) plus a single conditional skip.

net: -5 lines possible

---

## Test results

- `pytest -q tests/core/tools/test_command_path_guard_bash.py` → all passing (151 in one run, 132 in another — the discrepancy is most likely parametrized-case counting differing between runs; both runs were fully green, no failures either time).
- `pytest -q tests/core/tools/test_command_executor.py tests/core/tools/test_command_path_guard_bash.py tests/core/test_workspace_path_containment.py` → 2 pre-existing failures unrelated to this PR (both invoke the bare `python` interpreter, unavailable as `python` on this machine; neither test touches the diff or constructs a `path_guard`), 1 skip (Linux-only), everything else passed.

src/xagent/core/tools/core/command_path_guard.py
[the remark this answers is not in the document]
AlexLiu190625: Fixed in 4f930ed6. Heredoc normalization now uses one quote-aware lexical pass that queues declarations and consumes each body range exactly once, so delimiter-like text inside an active body is never scanned as a new opener. The reported nested-delimiter case now preserves the real outer terminator, and regressions cover the normalized output plus the live command after the terminator.

AlexLiu190625: Fixed in 4f930ed6. The regex rescan was replaced with a single lexical heredoc pass, removing the incidental parse-failure dependency. Tests cover nested-looking body text, multiple declarations in order, tab-stripping heredocs, preserved outer terminators, and rejection of the live command after the body.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in ff0618d0. One root validation session now tracks exact writes, unknown command effects, and inspected scripts. Earlier unknown effects reject later script inspection, and pipelines pre-collect concurrent effects so both writer/script directions reject. The real unclassified-copy overwrite case now returns 126 without exposing sibling data; propagation and reset controls are covered.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 4f930ed6. Bash invocation parsing is now a single sequential grammar. Dynamic values in the option region fail closed, file options and their arguments are consumed once, and dynamic positional values after command text remain outside the option region. The literal and variable-smuggling cases are covered by focused regressions.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in ff0618d0. Every root validation owns cumulative deterministic budgets for parse attempts, source characters, script bytes, repeated node-state evaluations, and argv tokens. Nested scripts and public re-entry share the same counters, while fresh and concurrent validations are isolated. Independent exhaustion and near-boundary controls replace timing-dependent assertions.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 4f930ed6. The former independent file-option scan and command-text scan were replaced by one structured Bash invocation parser. Recognized file options consume exactly one following argument, missing arguments reject explicitly, and consumed dash-prefixed values cannot be reinterpreted as command options. Real Bash differential cases and executor regressions cover the contract.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 4f930ed6. The per-script byte limit now shares the reachable 64 KiB policy boundary and oversized scripts raise a script-specific policy reason before parsing. Boundary tests cover limit minus one, limit, limit plus one, multibyte UTF-8, Bash file and source entry points, while oversized outer command text retains the generic command-size error.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 4f930ed6. Time-keyword normalization is part of the same cross-line quote-aware lexical pass used for heredocs. Multiline single- and double-quoted literals remain byte-for-byte unchanged, while executable unquoted time tokens are still normalized and validated.

src/xagent/core/tools/core/command_executor.py
AlexLiu190625: Fixed with a stronger contract in 6ca264ec. Guarded execute_script now fails closed with return code 126 before validation, tempfile creation, or process spawn, so it cannot silently switch from real-file semantics to bash -c semantics. The unguarded path retains its existing tempfile, $0, and BASH_SOURCE behavior, with both paths covered by regressions.

src/xagent/core/tools/core/command_executor.py
AlexLiu190625: Fixed in 6ca264ec and ba6fde72. Trusted executable roots and resolution now have one owner in command_policy.py, and the resolved canonical absolute Bash path is passed directly to process creation. Direct shell-free Bash is prepared by the policy owner; wrapped shell-free Bash is rejected before spawn. PATH-poisoning, exact argv/executable identity, and no-spawn regressions cover both entry forms.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 4f930ed6. The unused arbitrary count was replaced with the boolean consumes_leading_scalar invariant, and the wrapper parser now performs one conditional skip. Existing timeout controls and the full wrapper regression matrix remain green.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1636 @@
...
+        access: PathAccess,
+    ) -> Path:
+        return self._resolve_policy_path(
+            raw_path,
+            cwd,
+            access,
+            include_external_dirs=False,
+        )
+
+    def _resolve_policy_path(
```
rogercloud, line 1727: **CRITICAL: physical path resolution here diverges from Bash's logical `cd`, allowing a workspace escape.**

`_resolve_policy_path` (and the `cd` handling at ~982) always resolves to a *physically* resolved directory -- `resolve_authorized_path` in `src/xagent/core/workspace.py:1087-1122` calls `candidate.resolve()`, which follows symlinks, then checks `is_relative_to()`. Real Bash `cd` without `-P` is *logical*: it treats a symlink as an opaque segment and cancels `..` textually against the pre-symlink path. When a `cd` target crosses a symlink and carries enough `..` segments to round-trip past that symlink's real depth, the simulated `state.cwd` no longer matches where Bash actually lands.

Reproduced through a real guarded `CommandExecutorCore` (three separate calls, fresh workspace, sibling tenant dir holding `secret.txt`):

```
mkdir -p a/b/c/d/e/f
ln -s a/b/c/d/e/f s
cd s/../../../../../.. && cat <sibling-path>/secret.txt
```

The third call returns `success: True` with the sibling's file contents; `guard.validate()` on that string does not raise. The `rm` variant deletes the sibling's file. Verified twice independently, and it generalises to other symlink depths.

This is not the documented fail-open surface -- `cd`, `cat` and `rm` are all commands this module claims to fully classify. There is zero coverage: no test in `tests/core/tools/test_command_path_guard_bash.py` mentions "symlink".

Fix by either (a) tracking a **logical** cwd via string-based path collapsing that never touches the filesystem, mirroring Bash's own algorithm, or (b) rejecting any `cd` argument whose path traverses a symlink at all -- simpler and probably sufficient given the guard's narrow scope. Either way please add a `cd`/symlink differential regression test.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1636 @@
...
+    written_paths: set[Path] = field(default_factory=set)
+    unknown_effect: bool = False
+    inspected_scripts: set[Path] = field(default_factory=set)
+
+    @property
+    def visible_written_paths(self) -> frozenset[Path]:
+        return self.prior_written_paths | self.written_paths
+
+    @property
+    def has_visible_unknown_effect(self) -> bool:
```
rogercloud, line 80: **Major: the unclassified-effect flag is session-wide with no path scoping, which over-rejects ordinary command lines.**

`has_visible_unknown_effect` is a single boolean, so `inspect_script` refuses after *any* unclassified command anywhere earlier in the session, regardless of whether it could have touched the script. Reproduced as rejected: `echo hi && bash run.sh`, `mkdir -p out && bash run.sh`, and even `bash run.sh && bash run.sh` (the first script's own internal `echo` poisons the flag for the second). Only `bash run.sh && echo hi` passes.

I can see this is deliberate -- it's pinned by `test_effect_before_script_rejects_but_script_before_effect_is_allowed` -- and it fixes the previous round's `cp`/`mv`/`tee` hole in the safe direction. But almost any realistic agent-composed line that does setup before invoking a script now fails. Suggest tracking effects per-path (or at least per-script-target) rather than session-wide. Non-blocking; worth a follow-up issue.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1636 @@
...
+        *,
+        include_external_dirs: bool,
+    ) -> Path:
+        if isinstance(raw_path, _CommandValue) and not raw_path.is_static:
+            raise CommandPolicyViolation(
+                "cannot resolve dynamic path operand; enumerate concrete paths "
+                "instead of using active globs or shell expansions"
+            )
+
+        if raw_path in {"", "-", "{}"}:
```
rogercloud: Minor: the `{}` case here was presumably added for the xargs placeholder, but `_check_xargs` (~1498) rejects xargs unconditionally, so that path is unreachable. The remaining observable effect is that a literal file named `{}` in the workspace (`cat '{}'`) gets authorized against the *directory* rather than the file -- still inside the boundary, so harmless today, but it silently skips the per-file check. Suggest dropping `{}` from this set, or scoping it to the context it was intended for.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1636 @@
...
+                "cannot propagate multiple shell directory states from source"
+            )
+        validated_state = validated_states[0]
+        return replace(validated_state, script_depth=state.script_depth)
+
+    @staticmethod
+    def _is_direct_command_path(command_word: str) -> bool:
+        return "/" in command_word
+
+    @staticmethod
```
rogercloud, line 1355: Minor: `command_name` is unused in this function -- `resolve_trusted_executable(command_word)` is the only thing consulted. Call sites at ~902 and ~906 still pass it. Suggest dropping the parameter.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1636 @@
...
+from .command_policy import (
+    CommandPathViolation,
+    CommandPolicyViolation,
+    PathAccess,
+    resolve_trusted_executable,
+)
+
+logger = logging.getLogger(__name__)
+
+_MAX_COMMAND_POLICY_INPUT_CHARS = 64 * 1024
```
rogercloud, line 51: Minor (naming): `_MAX_INSPECTED_SCRIPT_BYTES` aliases `_MAX_COMMAND_POLICY_INPUT_CHARS`, i.e. a *character* limit reused as a *byte* limit. The byte-domain usage is correct in practice (`os.fstat().st_size` and `len(raw_script)` over bytes read in `"rb"` mode -- a multibyte-UTF-8 test confirms it), so this is purely a naming issue, but the alias makes the two domains look interchangeable when they aren't. Consider defining it as an independent constant.

rogercloud (commented):
## Re-review — `ref: add Bash command policy boundary` @ `ba6fde72`

Five new commits (`ff0618d0`, `4f930ed6`, `6ca264ec`, `049bb5a8`, `ba6fde72`) since the reviewed `bd646032`, and this is a substantial rewrite rather than a patch-up: `command_path_guard.py` grew ~280 lines, `command_executor.py` gained a guarded-execution path, and a new `command_policy.py` centralises trusted-executable resolution. I re-verified all 16 prior findings by reproduction rather than by reading the author's replies — 15 are genuinely fixed, several via real structural fixes (the heredoc lexer rewrite and the cross-call validation budget in particular are the right shape, not band-aids). However, an independent blind pass over the new head surfaced **one new CRITICAL defect that currently voids the boundary entirely**, so this cannot be approved as-is.

### Design verdict (updated)

The approach remains *acceptable-with-reservations*, but this update reveals a critical structural defect: the guard's `cd`/path-authorization model resolves paths **physically** (following symlinks), while real Bash's `cd` (without `-P`) is **logical** — it does not follow symlinks and collapses `..` textually against the pre-symlink path. That mismatch lets an ordinary, fully-classified command sequence escape the workspace using only `cd`/`cat`/`rm`, i.e. not via the documented fail-open surface. This is a regression-class gap in the sense that the `cd`-state tracking was largely untouched by this round's fixes, and the hardening everywhere else throws it into relief as the one live hole.

Two lower-priority design observations, both confirmed:

- The **cooperative fail-open contract** for unrecognized commands (`head`, `cp`, `tee`, `dd`, `python -c`, …) is unchanged and remains a disclosed, tested tradeoff (`test_unknown_commands_keep_the_explicit_fail_open_contract`). Already accepted in the prior round; noting it only for continuity, not as a new finding.
- `bashlex` does not parse a meaningful slice of Bash grammar — `$(( ))`, `[[ ]]`, `case`, `select`, `coproc`, array assignments. All of these currently **fail closed**, which I verified, and that is the correct behaviour. But it is a structural ceiling worth naming: any future "reduce false positives on unparseable input" change risks converting a `NotImplementedError` fail-closed path into a fail-open one. bashlex 0.18 (2023) has no active upstream maintenance.

---

## CRITICAL — physical vs. logical `cd` lets a fully-classified command sequence escape the workspace

**Files:** `src/xagent/core/tools/core/command_path_guard.py` — `_resolve_policy_path` (~1602-1636) and `_change_directory` / `cd` handling (~982-1042). Root cause reaches into `src/xagent/core/workspace.py:1087-1122`, where `resolve_authorized_path` calls `candidate.resolve()` (physical, follows symlinks) and then checks `is_relative_to(workspace_dir.resolve())`.

**Mechanism.** The guard's simulated shell state (`state.cwd`) is always the *physically resolved* directory. Real Bash's `cd` is *logical* by default: it never consults the filesystem to resolve intermediate symlinks, treats a symlink as an opaque path segment, and cancels `..` against the textual path. Whenever a `cd` target both crosses a symlink **and** carries enough `..` segments to round-trip past that symlink's real depth, the guard's belief about "where we are" diverges from where Bash actually lands.

**Reproduction** — three separate calls through a real, guarded `CommandExecutorCore`, fresh temp workspace, with a sibling tenant directory containing `secret.txt`:

```
1. mkdir -p a/b/c/d/e/f
2. ln -s a/b/c/d/e/f s
3. cd s/../../../../../.. && cat <sibling-tenant-path>/secret.txt
```

Step 3 returns `{'success': True, 'output': 'SIBLING SECRET\n', 'return_code': 0}` — a full, correct read of another tenant's file, through the exact guarded executor this PR ships. Calling `guard.validate()` directly on that string does not raise. The `rm` variant (`cd s/../../../../../.. && rm <sibling-path>/secret.txt`) also succeeds and deletes the sibling's file. Both were reproduced twice, independently, with byte-for-byte matching output. The bypass generalises — verified with the symlink one level deeper in a subdirectory (different depth, different dotdot count), same result.

**Why this matters.** `cd`, `cat`, and `rm` are precisely the three operations this module claims to fully classify and protect, so this is *not* the documented fail-open surface. It defeats the module's purpose for the exact threat model in xagent-saas#171 (one end user reaching a sibling end user's files via the shell). There is no live exploitation today — the guard is still dormant, with no production call site injecting it (re-confirmed this round) — but it must be fixed before any PR wires the guard into production. There is currently **zero** test coverage: no test in `tests/core/tools/test_command_path_guard_bash.py` references "symlink".

**Recommended fixes** (either is acceptable):

- (a) Track a **logical** cwd — string-based path collapsing that never touches the filesystem, mirroring Bash's own algorithm — alongside or instead of the physical one; or
- (b) Reject any `cd` argument whose path traverses a symlink at all. Simpler and more conservative, and likely sufficient given the guard's already-narrow scope.

Plus a `cd`/symlink differential regression test in either case.

---

## Prior findings — status (all 16 independently re-verified by reproduction)

|---|---|---|---|
| 1 | (gemini) shebang `#!` IndexError | **FIXED** | Still guards empty `interpreter_parts`; no regression from the later rewrites. |
| 2 | (gemini) CRLF heredoc | **FIXED** (fails closed) | Re-verified after the heredoc rewrite; still correctly fails closed on CRLF. |
| 3 | (gemini) `O_NONBLOCK`/`O_NOFOLLOW` Windows `AttributeError` | **FIXED** | Unchanged fail-closed design (not a 0-fallback). |
| 4 | (gemini) nested-shell `--` double-dash bypass | **FIXED** | Live-tested; no regression. |
| 5 | (gemini) regression test for `--` | **FIXED** | Covered. |
| 6 | (gemini + prior review) **critical** nested-heredoc-delimiter bypass | **FIXED — genuine structural fix** | The regex rescan is replaced by a real single-pass lexer (`_parse_heredoc_delimiter` + `_consume_heredoc_bodies`) that queues declarations and consumes each body exactly once. Live-reproduced the original PoC: the guard now specifically detects and rejects the dangerous command rather than merely failing to parse, and a heredoc-only safe case is still allowed (no false positive). |
| 7 | Major: script inspection defeated by unclassified write commands (`cp`/`mv`/`tee`) | **FIXED, but over-corrected** — see new Major below | A session-wide `unknown_effect` flag now blocks script inspection after *any* unclassified command, not just writers to the specific script. |
| 8 | Major: `$VAR` in the `bash` option region bypasses `--rcfile` checks | **FIXED** | `_parse_bash_invocation` now rejects any dynamic value anywhere in the option region; both previously-bypassing cases live-reproduced as rejected; no regression on legitimate dynamic content inside `-c` script text. |
| 9 | Major: exponential validation-cost DoS | **FIXED** | A real cross-call `_ValidationSession` (ContextVar-based) shares parse / node-state / byte / token budgets across the whole nested-script tree. Re-measured the original depth-8 chain: flat ~29 ms at every depth (was 54 s at depth 8), clean `CommandPolicyViolation` on budget exhaustion, no hang. |
| 10 | Major: `--rcfile`/`--init-file` as last token fails open | **FIXED** | Now explicitly raises "missing bash argument for …"; both bypassing cases rejected; complete pairs still allowed. |
| 11 | Minor: script byte limit unreachable / misleading message | **FIXED** | Limit is now shared and reachable; script-specific error message is distinct from the generic one. |
| 12 | Minor: `time` keyword rewrite not quote-aware | **FIXED** | Folded into the same quote-aware lexical pass as the heredoc fix; quoted content verified byte-for-byte unchanged. |
| 13 | Minor: unconditional `import bashlex` vs. optional extra | **NOT FIXED** | Still a bare unconditional import at `command_path_guard.py:29`; `bashlex` is still absent from base `[project] dependencies` in `pyproject.toml`. No author reply on this one either. |
| 14 | Minor: guarded `execute_script` changes `$0` semantics | **FIXED, via a different approach than requested** | Rather than preserving `$0`/positional-arg semantics, the guarded path now unconditionally fails closed (126) before any tempfile or spawn — it never executes a script via this method when a guard is present, so the divergence concern is moot (there is no guarded successful-execution path left to diverge). |
| 15 | Minor: guarded bash executable resolved via unchecked `shutil.which` | **FIXED** | Trusted-root resolution is centralised in `command_policy.py`'s `resolve_trusted_executable`, used by both the shell and argv-exec entry points. A live PATH-poisoning repro (fake `bash` earlier in `PATH`) is rejected before spawn; normal operation unaffected. |
| 16 | Simplification: `leading_scalars: int` → bool | **FIXED** | Now `consumes_leading_scalar: bool` with a single conditional skip; timeout-wrapper tests still pass. |

---

## New findings from this round

### Major — the unclassified-command "poison" flag is session-wide, not script- or path-scoped

`src/xagent/core/tools/core/command_path_guard.py` — `_EffectScope` / `unknown_effect` (~53-79, 942).

Confirmed: this is a single boolean with zero path specificity. Reproduced all of the following as **rejected**: `echo hi && bash run.sh`; `mkdir -p out && bash run.sh`; and even `bash run.sh && bash run.sh` (the first script's own internal `echo` trips the flag for the second invocation). Only `bash run.sh && echo hi` — effect strictly *after* the script — is allowed.

This is deliberate and pinned by tests (`test_effect_before_script_rejects_but_script_before_effect_is_allowed`), so it is not an accidental regression. But the practical impact is real: almost any agent-composed command line that does setup (`mkdir`, a non-`declare` `export`, any ordinary CLI tool) before invoking a script will now be rejected, even though the setup never touched the script. Fail-closed-but-annoying is the safe direction to err in, so this is not blocking — but effects should be tracked per-path, or at minimum per-script-target, rather than session-wide. Worth filing as a follow-up now instead of letting it surprise users later.

### Minor — `_TRUSTED_EXECUTABLE_ROOTS` is broader than deployment needs

`src/xagent/core/tools/core/command_policy.py:11-19` includes `/usr/local/bin` and `/opt/homebrew/bin`. Technically these can be non-root-writable on general Linux/macOS setups. Confirmed low-impact for this codebase's actual deployment: the `xagent-sandbox` image (`docker/Dockerfile.sandbox`) creates a dedicated unprivileged `sandbox` user, `/usr/local/bin` stays root-owned in the resulting image (nothing chowns it), and `/opt/homebrew/bin` does not exist in the Linux container — so exploiting this would need privileges the guard's own attacker model does not have. Narrowing to `/bin` and `/usr/bin` would remove the reliance on base-image ownership assumptions. Defense-in-depth, not urgent.

### Minor — `{}` is special-cased for a dead xargs code path

`src/xagent/core/tools/core/command_path_guard.py:1616`. `_check_xargs` rejects xargs unconditionally (~1498), so the `{}`-as-xargs-placeholder handling in `_resolve_policy_path` is unreachable for its intended purpose. Confirmed side effect: a literal file named `{}` in the workspace (e.g. `cat '{}'`) is authorized against the *directory* (cwd) instead of the file — harmless today since it stays inside the boundary, but it masks the real per-file check. Remove the dead special-case, or scope it correctly.

### Minor (code quality) — dead parameter

`src/xagent/core/tools/core/command_path_guard.py:1245` — `_is_trusted_system_command(command_word, command_name)` never uses `command_name`; call sites at ~902 and ~906 still pass it. No behavioural impact.

### Minor (naming) — `_MAX_INSPECTED_SCRIPT_BYTES` aliases a character-limit constant

`src/xagent/core/tools/core/command_path_guard.py:41-42`. The byte-domain usage is actually correct in practice (`os.fstat().st_size`, `len(raw_script)` over bytes read in `"rb"` mode), and a dedicated multibyte-UTF-8 test passes. This is a misleading name, not a functional bug.

---

## Tests

`pytest -q tests/core/tools/test_command_executor.py tests/core/tools/test_command_path_guard_bash.py tests/core/test_workspace_path_containment.py` → **352 passed, 1 skipped** (Linux-only), 0 failures.

---

## Recommendation

**Do not approve yet.** Everything raised in the previous round is resolved apart from #13 (unconditional `bashlex` import vs. optional extra), and several of the fixes are genuinely well-built — the heredoc lexer and the shared validation budget in particular. But the physical-vs-logical `cd` defect lets `cd` + `cat`/`rm` walk straight out of the workspace, which is the precise scenario this module exists to prevent, and it has no test coverage at all.

Blocking condition: fix the `cd`/symlink resolution mismatch and add a differential regression test. If the guard genuinely stays dormant and unwired for another release cycle, the minimum acceptable alternative is to file this as a **launch-blocking issue against whichever PR activates the guard** — it must not be allowed to go live in this state. The new Major (effect-ledger scoping) and the three Minors are follow-up material and need not block this merge.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 5b48070a. `cd`/`pushd` now reject any target that traverses a symlink (`_reject_symlink_traversing_chdir`), so the guard's tracked cwd can no longer diverge from Bash's logical cwd — physical and logical resolution agree whenever no symlink is crossed. The same-command variant (`ln -s a/b/c/d/e/f s && cd s/../../../../../..`) is also closed: when the link is absent at validation time, `resolve()` collapses `..` lexically and the target fails the existing containment check. Added cd/pushd symlink differential regression tests through the guarded executor (cat and rm variants) in `TestCdSymlinkTraversal`.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 5b48070a. The top-level `import bashlex` is now guarded, so the module imports (and its tests collect) on a base install; guard construction fails closed with an actionable hint (`pip install 'xagent[command-path-guard]'`) when the parser is genuinely required. bashlex stays an optional extra to keep the dependency staged until activation.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 5b48070a. Removed the dead `{}` cwd sentinel; a literal `{}` operand is now resolved and authorized as a file path instead of the directory. Added a regression test.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 5b48070a. Dropped the unused `command_name` parameter from `_is_trusted_system_command` and updated both call sites.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in 5b48070a. `_MAX_INSPECTED_SCRIPT_BYTES` now has its own byte-domain definition instead of aliasing the character input cap, with a comment noting the two limits are independent.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Deferring this as a follow-up rather than changing behavior here. The two-tier effect model is intentional: known writes are tracked per-path (`written_paths`), while a command with *unknown* effects has an unknowable target set, so it must poison a session-wide fail-closed flag — scoping that flag per-path would assume a path bound the guard cannot prove, i.e. it would trade a fail-closed property for a guess. The over-rejection is best reduced by classifying more commands (turning unknown effects into known per-path ones), which is the broader file-command classification this series explicitly defers to a later change, not by loosening the flag. Leaving the thread open and tracking it against the file-command coverage work.

AlexLiu190625: Addressed in 5585c23a via command classification rather than by scoping the unknown-effect flag. The flag stays session-wide for genuinely unknown commands (their target set is unknowable), but commands that cannot write the filesystem (`echo`, `printf`, `pwd`, `true`, `false`, `:`, `test`, `[`, `export`, `declare`, `typeset`) no longer poison it, and `mkdir` is now a path-scoped writer whose operands are validated as workspace writes and recorded per-path. So `mkdir out && bash run.sh`, `echo hi && bash run.sh` and `export X=1 && bash run.sh` are accepted, while a redirect from a no-effect command still registers its write (`echo hi > run.sh; bash run.sh` is rejected) and an unknown command still blocks a later script inspection. Regression tests in `TestNoEffectCommandClassification`.

src/xagent/core/tools/core/command_path_guard.py
```diff
@@ -0,0 +1,1734 @@
...
+        )
+
+    @staticmethod
+    def _reject_dynamic_values(context: str, values: Sequence[str]) -> None:
+        if any(
+            isinstance(value, _CommandValue) and not value.is_static for value in values
+        ):
+            raise CommandPolicyViolation(f"cannot resolve dynamic {context}")
+
+    @staticmethod
```
rogercloud, line 1678: **Minor:** `_operands()` has no concept of an option that consumes a following value -- it drops any token starting with `-` and treats everything else as an operand. Now that `mkdir` was added to `_WRITE_COMMANDS` and is routed through `_check_operands`, `mkdir -m 0755 newdir` drops `-m` correctly but then treats the mode string `0755` as a second path operand and runs it through the workspace containment check.

Not a security issue: nothing is checked less strictly, so this cannot produce a false ALLOW -- `0755` just resolves as a harmless relative segment under cwd. The only visible effect is a potentially confusing or duplicated rejection for a legitimate `mkdir -m <mode> <path-outside-workspace>`.

The imprecision is pre-existing (it applies to any option-with-value across every command using this helper); it only becomes realistic now that `mkdir`, which commonly takes `-m <mode>`, flows through here. Suggest a fast-follow that teaches this splitter which options take a value -- the wrapper-option handling elsewhere in this module already tracks per-option values with a real grammar, so converging on that would remove the discrepancy.

rogercloud (approved):
## Re-review (round 3) — head `5585c23a`

Two new commits since the last review (`ba6fde72`): `5b48070a` "fix: close cd symlink workspace escape and harden guard boundaries" and `5585c23a` "fix: classify no-effect and path-scoped commands for script inspection". Together they are a small, targeted 273-line diff (`command_path_guard.py` +112/-7 plus 168 lines of new tests) that addresses all 5 findings from the previous round, and additionally fixes the long-standing unconditional `bashlex` import that had been carried over unresolved.

All 6 items were independently verified as FIXED by reproduction against the current code — not by reading the change description. The critical `cd`/symlink workspace escape in particular was adversarially re-tested across 6 distinct bypass categories (chained `cd`, `pushd`/`popd`/`cd -` state, symlink-without-`..`, TOCTOU, absolute-path symlinks, non-existent/permission-denied components) plus double-symlink indirection and relative symlink targets; none succeeded.

### Status of previously reported findings

| 1 | **Critical** — `cd`/`pushd` resolved paths physically (following symlinks) while Bash's `cd` is logical, letting `mkdir`+`ln -s`+`cd`+`cat`/`rm` escape the workspace into a sibling tenant | **FIXED** | New `_reject_symlink_traversing_chdir()` (`src/xagent/core/tools/core/command_path_guard.py:1033`) is called from both the `cd` and `pushd` handlers before containment checking; it walks the target component-by-component and rejects if *any* traversed component is a symlink — fail-closed rather than emulating logical resolution. `cd -`/`popd` are correctly exempt (they restore an already-validated cwd). Original PoC now rejected both via `guard.validate()` and end-to-end through the guarded executor; 6 adversarial variant categories all rejected, no regression on symlink-free `cd sub/inner/.. && cat file.txt`. New `TestCdSymlinkTraversal` (6 tests) passes. |
| 2 | **Major** — a single unclassified command set a session-wide `unknown_effect` flag that poisoned *all* subsequent script inspection | **FIXED** | Resolved by reclassification rather than per-path scoping: new `_NO_FILESYSTEM_EFFECT_COMMANDS` (`command_path_guard.py:172`, covering `echo`/`printf`/`pwd`/`true`/`false`/`:`/`test`/`[`/`export`/`declare`/`typeset`) no longer poisons the flag, and `mkdir` is promoted to a path-scoped writer (`_WRITE_COMMANDS = {"rm", "mkdir"}`, line 164) validated and recorded per path. All previously blocked cases now allowed (`echo hi && bash run.sh`, `mkdir -p out && bash run.sh`, `printf`/`pwd`/`true`/`export`/`declare` variants, repeated `bash run.sh`); fail-closed preserved for genuinely unknown commands (`python -c pass && bash run.sh` still rejected); `mkdir <sibling-path>/newdir` still raises `CommandPathViolation`; and the subtle case is right — `echo hi > run.sh; bash run.sh` is still rejected because the redirect write is tracked independently of the declassified command. `TestNoEffectCommandClassification` (11 tests) passes. |
| 3 | **Minor** — dead `{}` xargs sentinel in the path-authorization check | **FIXED** | `raw_path in {"", "-", "{}"}` reduced to `raw_path in {"", "-"}`; a literal file named `{}` is now authorized as itself rather than silently substituted with cwd. Regression test passes. |
| 4 | **Minor** — unused `command_name` parameter on `_is_trusted_system_command` | **FIXED** | Signature reduced to `command_word` only; both call sites updated, no stale two-argument calls remain. |
| 5 | **Minor** — byte-limit constant aliased a character-limit constant | **FIXED** | `_MAX_INSPECTED_SCRIPT_BYTES` is now an independent literal `64 * 1024` with an explanatory comment (`command_path_guard.py:56`) instead of aliasing `_MAX_COMMAND_POLICY_INPUT_CHARS`. Value unchanged; pure definition cleanup. Multibyte test still passes. |
| 6 | **Minor** (carried over) — unconditional `import bashlex` in an optional module | **FIXED** | Import is now wrapped in `try/except ModuleNotFoundError` with `bashlex = None` fallback; `WorkspaceCommandPathGuard.__init__` raises an actionable `ModuleNotFoundError` pointing at `pip install 'xagent[command-path-guard]'`. Verified by monkeypatching `bashlex = None`: the error fires at construction, not at import/collection. With bashlex installed, collection is clean (254 tests, no errors). Keeping bashlex out of the base dependencies and failing closed at construction is the right call for a module meant to stay dormant. |

### New finding this round

**Minor — `_operands()` does not know that `mkdir -m <mode>` takes a separate value argument**

`src/xagent/core/tools/core/command_path_guard.py:1667` (`_operands`), now reached by `mkdir` through `_check_operands` (line ~969) since `mkdir` joined `_WRITE_COMMANDS` in this round. `_operands()` is a generic option/operand splitter: it drops any token starting with `-` but has no concept of an option that consumes a following value. For `mkdir -m 0755 newdir` it correctly drops `-m` but then treats `0755` as a second path operand and runs it through the workspace containment check alongside the real `newdir`.

Impact is low and is not a security gap: nothing is checked *less* strictly, so no false ALLOW is possible. `0755` resolves as a harmless relative segment under cwd and passes containment trivially; at worst, a legitimate `mkdir -m <mode> <dir-outside-workspace>` produces a slightly confusing or doubly-reported rejection. This is a pre-existing imprecision in `_operands()` — it understands no option-with-value pattern for any command — newly exposed to realistic input only because `mkdir`, a command commonly used with `-m <mode>`, was routed through `_check_operands` for the first time in this round. Worth a fast-follow to teach `_operands()` (or a `mkdir`-specific variant) about `-m`'s argument, ideally converging on the per-option value tracking the wrapper-option grammar elsewhere in this file already does.

### Tests

`tests/core/tools/test_command_executor.py`, `tests/core/tools/test_command_path_guard_bash.py`, `tests/core/test_workspace_path_containment.py` → **369 passed, 1 skipped** (Linux-only), **0 failed**.

### Recommendation

**Approve.** The critical blocking issue from the last round is genuinely and robustly fixed — the fail-closed symlink rejection holds up under broad adversarial probing, not merely against the single reported PoC — and every other finding from all three rounds is now resolved, each verified by reproduction. The one new minor finding (`mkdir -m` operand-parsing precision) is not blocking and can be handled as a fast-follow.

src/xagent/core/tools/core/command_path_guard.py
AlexLiu190625: Fixed in abaf71a1. `_operands()` now takes a per-command set of value-consuming options and skips the separated value token, wired via `_COMMAND_VALUE_OPTIONS` (`mkdir`: `-m`/`--mode`). So `mkdir -m 0755 dir` no longer runs the mode string through the containment check, while the real directory operand is still validated and `mkdir -m 0755 <path-outside-workspace>` is still rejected. Attached forms (`-m0755`, `--mode=0755`) were already dropped whole. Regression tests cover separated/attached forms and the outside-path case in `TestNoEffectCommandClassification`.

PR #1037, opened 2026-07-28
feat: add resolvable context references

qinxuye:
## Summary

- add a durable `ContextReference` protocol backed by registered, sanitized `FileRef` values
- preserve references across messages, checkpoints, compaction, and tool-result boundaries without persisting image bytes or host paths
- materialize image references only at the LLM call boundary for vision-capable models; degrade unresolved or non-vision inputs to safe text/FileRef context
- teach the auto router to prefer matching input modalities and expose the selected model's resolved abilities
- keep resolver memory bounded with size, TTL, and LRU limits

## Scope

This is the generic context/runtime layer only. It intentionally contains no browser, desktop, relay, target-selection, screenshot-frame, or Computer Use product implementation.

## Validation

- `PYTHONPATH=src pytest -q tests/core/agent tests/core/test_context_ref.py tests/core/test_context_materializer.py tests/core/test_file_ref.py tests/core/model/chat/basic/test_router.py`
- `pre-commit run --all-files`
- diff boundary audit for Computer Use-specific concepts

### DIFF
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -40,6 +40,8 @@ from .skill_tool import LOADED_SKILLS_METADATA_KEY, SKILL_INDEX_METADATA_KEY
 READ_FILE_CONTEXT_LIMIT = 12_000
 COMPACT_SUMMARY_MAX_TOKENS = 1024
 COMPACT_SUMMARY_MIN_TOKENS = 256
+COMPACT_CONTEXT_REF_MAX_TOKENS = 2048
+COMPACT_DROPPED_REF_NOTICE_MAX_CHARS = 2048
 
 
 def _utcnow() -> datetime:
@@ -993,8 +995,11 @@ class ExecutionContext:
             )
 
         latest_user = self._latest_visible_user_message()
-        compacted_context_refs = self._context_refs_removed_by_compaction(latest_user)
-        summary_message = Message.role_system(
+        compacted_context_refs, dropped_context_refs = (
+            self._context_refs_removed_by_compaction(latest_user)
+        )
+        dropped_refs_notice = self._dropped_context_refs_notice(dropped_context_refs)
+        summary_content = (
             "Compacted conversation summary:\n"
             f"{summary}\n\n"
             "Use this summary as the current execution state. Continue from the "
@@ -1002,7 +1007,12 @@ class ExecutionContext:
             "regenerate completed artifacts unless the latest user request "
             "explicitly asks to restart, revise, or regenerate them. Current system "
             "instructions still take precedence, and the latest user request remains "
-            "the overall goal.",
+            "the overall goal."
+        )
+        if dropped_refs_notice:
+            summary_content = f"{summary_content}\n\n{dropped_refs_notice}"
+        summary_message = Message.role_system(
+            summary_content,
             metadata={"compacted_context": True},
             context_refs=compacted_context_refs,
         )
@@ -1019,6 +1029,8 @@ class ExecutionContext:
                 "removed_count": max(0, original_count - len(self.messages)),
                 "summary_chars": len(summary),
                 "compact_model": getattr(llm, "model_name", None),
+                "retained_context_ref_count": len(compacted_context_refs),
+                "dropped_context_ref_count": len(dropped_context_refs),
             },
         )
         if original_tokens is not None:
@@ -1027,23 +1039,69 @@ class ExecutionContext:
 
     def _context_refs_removed_by_compaction(
         self, latest_user: Message | None
-    ) -> tuple[ContextReference, ...]:
+    ) -> tuple[tuple[ContextReference, ...], tuple[ContextReference, ...]]:
         seen = (
             {reference.identity_key() for reference in latest_user.context_refs}
             if latest_user is not None
             else set()
         )
+        latest_user_tokens = (
+            sum(reference.estimated_tokens() for reference in latest_user.context_refs)
+            if latest_user is not None
+            else 0
+        )
+        remaining_tokens = max(
+            0,
+            min(
+                COMPACT_CONTEXT_REF_MAX_TOKENS,
+                max(0, self.compact_config.threshold // 8),
+            )
+            - latest_user_tokens,
+        )
         retained: list[ContextReference] = []
-        for message in self.messages:
+        dropped: list[ContextReference] = []
+        for message in reversed(self.messages):
             if message.hidden or message is latest_user:
                 continue
-            for reference in message.context_refs:
+            for reference in reversed(message.context_refs):
                 identity = reference.identity_key()
                 if identity in seen:
                     continue
                 seen.add(identity)
-                retained.append(reference)
-        return tuple(retained)
+                reference_tokens = reference.estimated_tokens()
+                if reference_tokens <= remaining_tokens:
+                    retained.append(reference)
+                    remaining_tokens -= reference_tokens
+                else:
+                    dropped.append(reference)
+        retained.reverse()
+        return tuple(retained), tuple(dropped)
+
+    @staticmethod
+    def _dropped_context_refs_notice(
+        references: tuple[ContextReference, ...],
+    ) -> str:
+        if not references:
+            return ""
+        prefix = (
+            "Older image references exceeded the structured-reference budget and "
+            "will not be automatically rematerialized after compaction. Durable "
+            "handles retained in this summary:\n"
+        )
+        lines: list[str] = []
+        current_chars = len(prefix)
+        omitted = 0
+        for reference in references:
+            filename = reference.safe_file_ref.get("filename") or "image"
+            line = f"- [image: {filename}, file_id={reference.file_id}]"
+            if current_chars + len(line) + 1 > COMPACT_DROPPED_REF_NOTICE_MAX_CHARS:
+                omitted += 1
+                continue
+            lines.append(line)
+            current_chars += len(line) + 1
+        if omitted:
+            lines.append(f"- ... {omitted} additional older reference(s) omitted")
+        return prefix + "\n".join(lines)
 
     def _latest_visible_user_message(self) -> Message | None:
         for message in reversed(self.messages):
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -8,6 +8,8 @@ from typing import Any
 from uuid import uuid4
 
 from ...context_ref import (
+    CONTEXT_REFS_KEY,
+    ContextReference,
     normalize_context_references,
     split_tool_result_context_references,
 )
@@ -345,17 +347,20 @@ class ExecutionContext:
             if include_system and message_dict.get("role") == "system":
                 content = str(message_dict.get("content") or "").strip()
                 if content:
-                    messages.append(
-                        {
-                            "role": "user",
-                            "content": (
-                                "Previous system-context message retained for "
-                                "continuity. Current system instructions above "
-                                "take precedence:\n"
-                                f"{content}"
-                            ),
-                        }
-                    )
+                    continuity_message: dict[str, Any] = {
+                        "role": "user",
+                        "content": (
+                            "Previous system-context message retained for "
+                            "continuity. Current system instructions above "
+                            "take precedence:\n"
+                            f"{content}"
+                        ),
+                    }
+                    if CONTEXT_REFS_KEY in message_dict:
+                        continuity_message[CONTEXT_REFS_KEY] = message_dict[
+                            CONTEXT_REFS_KEY
+                        ]
+                    messages.append(continuity_message)
                 continue
             messages.append(message_dict)
 
@@ -988,6 +993,7 @@ class ExecutionContext:
             )
 
         latest_user = self._latest_visible_user_message()
+        compacted_context_refs = self._context_refs_removed_by_compaction(latest_user)
         summary_message = Message.role_system(
             "Compacted conversation summary:\n"
             f"{summary}\n\n"
@@ -998,6 +1004,7 @@ class ExecutionContext:
             "instructions still take precedence, and the latest user request remains "
             "the overall goal.",
             metadata={"compacted_context": True},
+            context_refs=compacted_context_refs,
         )
         next_messages = [summary_message]
         if latest_user is not None:
@@ -1017,6 +1024,26 @@ class ExecutionContext:
         if original_tokens is not None:
             return self._annotate_compact_result(result, original_tokens)
         return result
+
+    def _context_refs_removed_by_compaction(
+        self, latest_user: Message | None
+    ) -> tuple[ContextReference, ...]:
+        seen = (
+            {reference.identity_key() for reference in latest_user.context_refs}
+            if latest_user is not None
+            else set()
+        )
+        retained: list[ContextReference] = []
+        for message in self.messages:
+            if message.hidden or message is latest_user:
+                continue
+            for reference in message.context_refs:
+                identity = reference.identity_key()
+                if identity in seen:
+                    continue
+                seen.add(identity)
+                retained.append(reference)
+        return tuple(retained)
 
     def _latest_visible_user_message(self) -> Message | None:
         for message in reversed(self.messages):
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -7,6 +7,10 @@ from enum import Enum
 from typing import Any
 from uuid import uuid4
 
+from ...context_ref import (
+    normalize_context_references,
+    split_tool_result_context_references,
+)
 from ...file_ref import FILE_REF_MODEL_INSTRUCTIONS
 from ...tools.artifacts import (
     format_tool_result_for_observation,
@@ -155,8 +159,23 @@ class ExecutionContext:
         tool_name: str,
         result: Any,
         tool_call_id: str | None = None,
+        *,
+        context_refs: Any = (),
     ) -> Message:
-        context_result = self._sanitize_tool_result_for_context(tool_name, result)
+        public_result, embedded_refs = split_tool_result_context_references(result)
+        explicit_refs = normalize_context_references(context_refs)
+        all_context_refs = []
+        seen_context_refs: set[str] = set()
+        for reference in (*explicit_refs, *embedded_refs):
+            identity = reference.identity_key()
+            if identity in seen_context_refs:
+                continue
+            seen_context_refs.add(identity)
+            all_context_refs.append(reference)
+
+        context_result = self._sanitize_tool_result_for_context(
+            tool_name, public_result
+        )
         content = self._format_tool_result(tool_name, context_result)
         metadata = {
             "tool_name": tool_name,
@@ -171,6 +190,7 @@ class ExecutionContext:
             content,
             tool_call_id=tool_call_id,
             metadata=metadata,
+            context_refs=tuple(all_context_refs),
         )
 
     def attach_workspace(
@@ -541,6 +561,7 @@ class ExecutionContext:
             tool_call_id=response_message.tool_call_id,
             hidden=response_message.hidden,
             output_tokens=output_tokens,
+            context_refs=response_message.context_refs,
         )
         self.messages.append(updated_response)
         self.llm_calls.append(
@@ -752,6 +773,9 @@ class ExecutionContext:
                     "tool_call_id": message.tool_call_id,
                     "hidden": message.hidden,
                     "output_tokens": message.output_tokens,
+                    "context_refs": [
+                        reference.durable_dict() for reference in message.context_refs
+                    ],
                 }
                 for message in self.messages
             ],
@@ -797,6 +821,7 @@ class ExecutionContext:
                 tool_call_id=item.get("tool_call_id"),
                 hidden=item.get("hidden", False),
                 output_tokens=item.get("output_tokens"),
+                context_refs=item.get("context_refs", ()),
             )
             for item in data.get("messages", [])
         ]
@@ -1057,6 +1082,8 @@ class ExecutionContext:
                     + json.dumps(message.tool_calls, ensure_ascii=False, default=str)
                 )
             chunks.append(message.content)
+            if message.context_refs:
+                chunks.append(message.context_refs_text())
         return "\n".join(chunks)
 
     def _compact_response_text(self, response: Any) -> str:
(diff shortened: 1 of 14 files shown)


src/xagent/core/agent/context/execution.py
```diff
@@ -1057,6 +1082,8 @@ def _compact_transcript(self, messages: list[Message]) -> str:
                     + json.dumps(message.tool_calls, ensure_ascii=False, default=str)
                 )
             chunks.append(message.content)
+            if message.context_refs:
+                chunks.append(message.context_refs_text())
```
rogercloud, line 1171: [P2] Preserve structured context references through LLM compaction

This only renders each reference as plain text for the summarizer. `compact_with_llm_response()` subsequently replaces the history with a new ref-less summary plus the latest user message, so references on older user, assistant, or tool messages can no longer reach runtime image materialization. Carry deduplicated structured references onto a retained compacted message, and add an end-to-end post-compaction vision test.

src/xagent/core/context_materializer.py
```diff
@@ -0,0 +1,258 @@
...
+
+    async def resolve_image(self, reference: ContextReference) -> str:
+        cache_key = self._cache_key(reference)
+        cached = self._cache.get(cache_key)
+        now = time.monotonic()
+        if cached is not None:
+            created_at, data_url = cached
+            if now - created_at <= self.cache_ttl_seconds:
+                self._cache.move_to_end(cache_key)
+                return data_url
```
rogercloud: [P2] Check the current file generation before serving cached bytes

A cache hit returns solely by `file_id` before re-resolving the workspace entry. Workspace re-registration can retain the same ID while replacing the underlying generation, and normal references contain no `sha256`; this therefore returns stale image bytes for up to 300 seconds. Resolve and key by an immutable generation/digest, or invalidate this cache whenever registration changes the file behind an ID.

src/xagent/core/context_materializer.py
```diff
@@ -0,0 +1,258 @@
...
+            self._cache.pop(cache_key, None)
+
+        resolve_file_id = getattr(self.workspace, "resolve_file_id", None)
+        if not callable(resolve_file_id):
+            raise ContextReferenceResolutionError(
+                "workspace must expose resolve_file_id"
+            )
+        path = resolve_file_id(reference.file_id)
```
rogercloud: [P2] Move workspace resolution off the async model-call path

`TaskWorkspace.resolve_file_id()` is synchronous and can execute a database query plus durable object-store materialization. Calling it directly here can block the event loop and delay unrelated requests. Expose an asynchronous resolver or safely offload the complete resolution operation, including appropriate DB-session ownership.

src/xagent/core/context_ref.py
```diff
@@ -0,0 +1,149 @@
...
+        result: dict[str, Any] = {
+            "type": self.type,
+            "file_ref": self.safe_file_ref,
+            "detail": self.detail.value,
+            "metadata": _validated_metadata(self.metadata),
+        }
+        if self.text_fallback is not None:
+            result["text_fallback"] = self.text_fallback
+        return result
+
```
rogercloud, line 192: [P2] Sanitize every persisted `ContextReference` field

Only `file_ref` receives path stripping. Arbitrary `metadata` can still persist host paths, while `text_fallback` can contain a `data:image/...;base64,...` payload; both then enter checkpoints and traces. Restrict metadata to a typed/allowlisted durable schema (or recursively reject private path fields) and apply the materialized-image rejection to `text_fallback` before serialization.

rogercloud (commented):
Design verdict: acceptable with reservations. The protocol/runtime split is coherent, but the inline P2 findings should be addressed before relying on durable multimodal continuity in long-running or concurrent sessions. All current GitHub checks pass; Python 3.11 byte-compilation and `git diff --check` also passed locally. Focused pytest was not run locally because no prepared project environment was available.

src/xagent/core/context_ref.py
qinxuye: Fixed in efd3d504. ContextReference metadata is now recursively validated: path-like fields, absolute/private filesystem paths, non-string keys, and materialized image data are rejected. text_fallback now rejects materialized image data too, and durable_dict revalidates post-construction mutations. Added coverage for nested paths, direct paths, fallback base64, and late mutations.

src/xagent/core/context_materializer.py
qinxuye: Fixed in efd3d504. Workspace resolution and filesystem metadata/read work now run via asyncio.to_thread. TaskWorkspace exposes resolve_file_id_detached(), which keeps the same scope checks but opens/closes a worker-owned DB session rather than reusing the caller session. Added tests for worker-thread execution and detached session ownership.

src/xagent/core/agent/context/execution.py
qinxuye: Fixed in efd3d504. LLM compaction now deduplicates structured refs from removed visible messages onto the retained summary message, and system-continuity conversion preserves the reserved ref envelope. Added an end-to-end test that compacts an older image ref and verifies it is still materialized for a vision model without persisting base64.

src/xagent/core/context_materializer.py
qinxuye: Fixed in efd3d504. Cache lookup now happens only after re-resolving the current file generation, and the key includes the resolved immutable generation path plus device/inode/size/mtime (and sha256 when supplied). Reads verify that the generation remains stable. Added a same-file_id generation-swap regression test.

rogercloud (commented):
Re-review of efd3d504. Three prior findings are fixed (structured refs across compaction, generation-aware cache invalidation, and event-loop-safe workspace resolution). The ContextReference sanitization finding is only partially fixed, and the remaining issue is included inline. The current implementation still has three P2 issues that should be addressed before merge, plus one P3 portability issue.

All current GitHub checks pass. Local Python 3.11 byte-compilation and `git diff --check` pass. Focused pytest was not run locally because the required dependencies were unavailable; the relevant tests were inspected.

src/xagent/core/model/chat/basic/router.py
```diff
@@ -512,11 +582,54 @@ async def _select_model(self, prompt: str) -> str:
...
+            route_parameters = {}
+        supports_modality_preferences = (
+            "preferred_input_modalities" in route_parameters
+            or any(
+                parameter.kind is inspect.Parameter.VAR_KEYWORD
+                for parameter in route_parameters.values()
+            )
+        )
+        if preferred_input_modalities and supports_modality_preferences:
+            route_kwargs["preferred_input_modalities"] = preferred_input_modalities
```
rogercloud, line 610: [P2] Do not silently route image requests without modality support

The locked `xrouter-llm` 0.1.2 API does not accept `preferred_input_modalities`, so this branch silently omits the image requirement and allows the auto-router to choose a text-only model. The later materializer then degrades image refs to text fallback, making modality-aware routing ineffective for the supported configuration. If modality preferences are present but unsupported, select a known vision-capable fallback or fail clearly; alternatively integrate and lock a router API/profile that actually enforces the capability.

src/xagent/core/agent/context/execution.py
```diff
@@ -993,6 +1025,26 @@ def compact_with_llm_response(
...
+        retained: list[ContextReference] = []
+        for message in self.messages:
+            if message.hidden or message is latest_user:
+                continue
+            for reference in message.context_refs:
+                identity = reference.identity_key()
+                if identity in seen:
+                    continue
+                seen.add(identity)
+                retained.append(reference)
```
rogercloud: [P2] Bound the refs retained by compaction

This retains every distinct historical context ref indefinitely. Each AUTO image costs at least 255 estimated tokens, so roughly 126 refs can leave the newly compacted context above the default 32k threshold; subsequent calls then compact again and rematerialize old, irrelevant images. Apply a deterministic ref/token budget during compaction, retaining only recent or still-needed refs while preserving dropped handles as text.

src/xagent/core/context_ref.py
```diff
@@ -0,0 +1,214 @@
...
+        if value not in (None, "", [], {}):
+            raise ValueError(f"metadata field {key!r} must not contain a path")
+
+    if isinstance(value, dict):
+        for child_key, child_value in value.items():
+            if not isinstance(child_key, str):
+                raise ValueError("metadata keys must be strings")
+            _validate_metadata_tree(child_value, key=child_key)
+        return
+    if isinstance(value, list):
```
rogercloud, line 86: [P2] Validate the JSON-normalized metadata tree

The recursive validator descends into lists but not tuples. A tuple-wrapped host path or image data URI therefore passes validation and is persisted; JSON serialization converts the tuple to a list, after which checkpoint restoration rejects the same value. Recurse through both lists and tuples, or validate the `json.loads(serialized)` representation so accepted state is guaranteed to round-trip.

src/xagent/core/file_ref.py
```diff
@@ -160,6 +160,41 @@ def build_workspace_file_ref(
...
+        file_id=file_id,
+        filename=filename,
+        mime_type=mime_type,
+        size=size,
+    )
+    relative_path = file_ref.get("relative_path")
+    if relative_path is not None:
+        relative = Path(str(relative_path))
+        if not relative.is_absolute() and ".." not in relative.parts:
+            result["relative_path"] = str(relative)
```
rogercloud: [P3] Treat Windows-shaped paths as paths on POSIX hosts

On POSIX, `Path(r"C:\\Users\\name\\file.png")` and `Path(r"..\\secret.png")` are neither absolute nor split into `..` parts, so `relative_path` persists host-path/traversal forms the sanitizer intends to exclude. Normalize separators or use `PureWindowsPath` checks before retaining this field.

src/xagent/core/model/chat/basic/router.py
qinxuye: Fixed in 0f0a63f8. Modality-bearing requests now fail clearly when the installed xrouter API cannot enforce preferred_input_modalities, and this dedicated error is not hidden by the generic router fallback. Routers that expose the supported argument still receive the requested modalities; regression coverage includes both paths.

src/xagent/core/agent/context/execution.py
qinxuye: Fixed in 0f0a63f8. Compaction now applies a deterministic recent-first token budget, accounts for refs already present in the latest user turn, and preserves dropped durable handles in a bounded text notice. Tests cover retained/dropped ordering and budget metadata.

src/xagent/core/context_ref.py
qinxuye: Fixed in 0f0a63f8. Metadata is now validated before serialization and again on the JSON-normalized tree, so tuple values normalize to lists and tuple-wrapped private paths are rejected before persistence. Round-trip and rejection tests were added.

src/xagent/core/file_ref.py
qinxuye: Fixed in 0f0a63f8. The sanitizer now detects Windows drives and roots with PureWindowsPath, normalizes backslashes, and rejects traversal after normalization. Tests cover drive paths, UNC/rooted paths, traversal, and safe relative Windows-shaped input.

src/xagent/core/context_ref.py
```diff
@@ -0,0 +1,219 @@
...
+        for child in value:
+            _validate_metadata_tree(child)
+        return
+    if isinstance(value, str):
+        if _contains_materialized_image(value):
+            raise ValueError("metadata must not contain materialized image data")
+        stripped = value.strip()
+        if _WINDOWS_ABSOLUTE_PATH_RE.match(stripped) or (
```
rogercloud: [P2] Reject UNC and Windows-rooted metadata paths

This still accepts host paths such as `\\\\server\\share\\file.png`, `\\rooted\\file.png`, and `//server/share/file.png` when they appear under a non-path-like metadata key, so they can enter persisted checkpoints and traces. Use a cross-platform absolute-path check that covers UNC and Windows-rooted forms, and add tests for these values in ordinary and nested metadata.

src/xagent/core/context_materializer.py
```diff
@@ -0,0 +1,322 @@
...
+                    "context image changed while it was being read"
+                ) from exc
+
+            if len(image_bytes) > self.max_image_bytes:
+                raise ContextReferenceResolutionError(
+                    f"context image exceeds {self.max_image_bytes} bytes"
+                )
+            encoded = base64.b64encode(image_bytes).decode("ascii")
+            data_url = f"data:{mime_type};base64,{encoded}"
+            self._cache[cache_key] = (time.monotonic(), data_url)
```
rogercloud: [P2] Bound the cache by encoded bytes, not only entry count

A permitted 20 MiB image becomes roughly a 26.7 MiB base64 data URL. With eight entries, one active resolver can retain more than 213 MiB, and memory scales linearly with concurrent runs. Track a total encoded-byte budget and evict or skip entries that would exceed it; entry count alone is not a meaningful memory bound.

src/xagent/core/agent/context/execution.py
```diff
@@ -987,12 +1029,80 @@ def compact_with_llm_response(
...
 
+    def _context_refs_removed_by_compaction(
+        self, latest_user: Message | None
+    ) -> tuple[tuple[ContextReference, ...], tuple[ContextReference, ...]]:
+        seen = (
+            {reference.identity_key() for reference in latest_user.context_refs}
+            if latest_user is not None
+            else set()
+        )
+        latest_user_tokens = (
```
rogercloud, line 1048: [P2] Enforce an aggregate image budget on the current request

Historical refs are now budgeted, but the entire latest user message remains outside that cap. For example, 43 HIGH-detail refs already exceed the default 32k estimate, so the compacted context stays over threshold and all images are rematerialized. Enforce a request-level ref-count, image-byte, or token budget before materialization and explicitly reject or degrade over-budget input rather than silently trimming the current request.

src/xagent/core/context_materializer.py
```diff
@@ -0,0 +1,322 @@
...
+            while len(self._cache) > self.cache_size:
+                self._cache.popitem(last=False)
+            return data_url
+
+        raise ContextReferenceResolutionError("unable to read a stable context image")
+
+    def clear(self) -> None:
```
rogercloud: [simplification:delete] Remove the unused cache-clear API

`clear()` has no production caller and its sole test call is unnecessary because every lookup recomputes the generation-bearing cache key. Removing the method and that test-only call preserves behavior and eliminates an unused public surface.

src/xagent/core/agent/context/execution.py
```diff
@@ -987,12 +1029,80 @@ def compact_with_llm_response(
...
+            sum(reference.estimated_tokens() for reference in latest_user.context_refs)
+            if latest_user is not None
+            else 0
+        )
+        remaining_tokens = max(
+            0,
+            min(
+                COMPACT_CONTEXT_REF_MAX_TOKENS,
+                max(0, self.compact_config.threshold // 8),
```
rogercloud: [simplification:shrink] Remove the redundant inner clamp

The outer `max(0, ... - latest_user_tokens)` already clamps negative results, and `latest_user_tokens` is non-negative. The inner `max(0, threshold // 8)` can therefore be reduced to `threshold // 8` without changing behavior for the declared integer configuration.

rogercloud (commented):
Re-review of 0f0a63f8. The latest update fixes the tuple-metadata bypass and Windows-shaped `FileRef.relative_path` handling, adds a bounded policy for historical compacted refs, and replaces silent modality degradation with an explicit compatibility error.

Three P2 action items remain across two areas: complete cross-platform metadata path sanitization, and enforce aggregate image/reference budgets for both cached bytes and individual requests. Two non-blocking simplification opportunities are also included inline.

All current GitHub checks pass. Local Python 3.11 byte-compilation and `git diff --check` pass. Focused pytest was not run locally because the available Python 3.11 environment lacks pytest and the other environments have incompatible versions or dependencies.

src/xagent/core/context_materializer.py
qinxuye: Fixed in 3d00cf6d. Removed the unused clear() method and the test-only call; the generation-bearing cache key continues to invalidate rewritten files without manual clearing.

src/xagent/core/context_ref.py
qinxuye: Fixed in 3d00cf6d. Metadata strings now use a cross-platform absolute-path check: POSIX-leading paths plus Windows drive/root/UNC forms detected through PureWindowsPath are rejected. Added ordinary, nested-object, and nested-list regressions for UNC, Windows-rooted, and //server paths.

src/xagent/core/agent/context/execution.py
qinxuye: Fixed in 3d00cf6d. Vision materialization now preflights the aggregate reference token estimate against 25% of the resolved model context window (8,192 fallback) and raises a clear error before resolution when exceeded. It also enforces a 32 MiB aggregate materialized data-URL byte cap before the provider call. Tests cover both token and byte overflow.

src/xagent/core/agent/context/execution.py
qinxuye: Fixed in 3d00cf6d. Removed the redundant inner max(0, ...) clamp while retaining the outer non-negative clamp.

src/xagent/core/context_materializer.py
qinxuye: Fixed in 3d00cf6d. Resolver cache entries now record encoded data-URL bytes and enforce a 32 MiB aggregate byte budget in addition to entry count. LRU entries are evicted until both bounds fit, while a single entry larger than the byte budget is returned but not cached. Added an eviction/read-count regression.

src/xagent/core/context_ref.py
```diff
@@ -0,0 +1,224 @@
...
+
+def _validate_metadata_tree(value: Any, *, key: str | None = None) -> None:
+    if key is not None and _metadata_key_is_path_like(key):
+        if value not in (None, "", [], {}):
+            raise ValueError(f"metadata field {key!r} must not contain a path")
+
+    if isinstance(value, dict):
+        for child_key, child_value in value.items():
+            if not isinstance(child_key, str):
+                raise ValueError("metadata keys must be strings")
```
rogercloud, line 82: [P2] Validate sensitive content in metadata keys as well as values

This loop only verifies that each key is a string; the data-URI, absolute-path, and private-path checks are applied only to `child_value`. Consequently metadata keys containing `data:image/...;base64,...`, `/private/...`, or a UNC path survive `durable_dict()` and checkpoint/trace JSON round-trips. Apply the same sensitive-string validation to each key and add persistence tests for unsafe keys.

src/xagent/core/context_materializer.py
```diff
@@ -0,0 +1,382 @@
...
+            data_url = f"data:{mime_type};base64,{encoded}"
+            encoded_bytes = len(data_url)
+            if encoded_bytes <= self.max_cache_bytes:
+                while self._cache and (
+                    len(self._cache) >= self.cache_size
+                    or self._cache_bytes + encoded_bytes > self.max_cache_bytes
+                ):
+                    _, evicted = self._cache.popitem(last=False)
+                    self._cache_bytes -= evicted.encoded_bytes
+                self._cache[cache_key] = _CacheEntry(
```
rogercloud, line 129: [P3] Subtract a replaced same-key entry before updating cache bytes

Two concurrent misses for the same generation can both await the read, then assign this same key while each increments `_cache_bytes`. A reproduced 158-byte entry was therefore reported as 316 bytes, causing later false eviction. Before assignment, remove and subtract any existing same-key entry, or serialize/double-check per-key population; add an `asyncio.gather` regression asserting the counter equals the stored entries.

src/xagent/core/context_materializer.py
```diff
@@ -0,0 +1,382 @@
...
+        if self._generation(path) != expected:
+            raise _FileGenerationChanged
+        return image_bytes
+
+    @staticmethod
+    def _cache_key(
+        reference: ContextReference,
+        generation: _FileGeneration,
+    ) -> str:
+        digest = reference.metadata.get("sha256")
```
rogercloud: [simplification:shrink] Remove the unused metadata sha256 cache-key branch

No production producer or protocol contract supplies `metadata["sha256"]`, and the value is never checked against the resolved bytes. The validated file generation already participates in the cache key, so this untrusted optional string only creates undocumented cache misses without providing integrity. Always key by `file_id` plus generation and remove this branch.

rogercloud (commented):
Re-review of 3d00cf6d. The update fixes the previously reported cross-platform metadata-value paths, aggregate cache-byte bound, request-level image budgets, and both simplification items.

One P2 persistence-boundary issue remains, together with one non-blocking P3 cache-accounting race. A separate non-blocking simplification opportunity is included inline.

All current GitHub checks pass. Local Python compilation and `git diff --check` pass. Focused context-reference/materializer tests, FileRef tests, and context-reference tests also pass.

src/xagent/core/context_ref.py
qinxuye: Fixed in 8a83161c. Metadata keys and values now share the same sensitive-string validator, covering materialized image data plus POSIX, Windows-rooted, and UNC paths. Added durable_dict mutation/persistence-boundary regressions for data-URI, /private, and UNC keys.

src/xagent/core/context_materializer.py
qinxuye: Fixed in 8a83161c. Cache insertion now removes and subtracts an existing same-key entry before applying byte/count eviction and replacing it. Added a barrier-synchronized asyncio.gather regression that forces two same-generation misses and verifies _cache_bytes equals the stored entries.

src/xagent/core/context_materializer.py
qinxuye: Fixed in 8a83161c. Removed the undocumented metadata sha256 branch; cache keys now always use the registered file_id plus the validated immutable file generation.

src/xagent/core/context_ref.py
```diff
@@ -0,0 +1,229 @@
...
+    parts = {part for part in re.split(r"[^a-z0-9]+", key.lower()) if part}
+    return bool(parts & _PATH_METADATA_PARTS)
+
+
+def _looks_like_absolute_filesystem_path(value: str) -> bool:
+    stripped = value.strip()
+    if not stripped:
+        return False
+    windows_path = PureWindowsPath(stripped)
+    return stripped.startswith("/") or bool(windows_path.drive or windows_path.root)
```
rogercloud: [P2] Detect host paths embedded inside metadata strings

The current check recognizes strings that are themselves absolute paths, while the private-path regex requires the path to follow the beginning of the string or whitespace. Values such as `saved at (/private/frame.png)`, `file:///private/frame.png`, and `source=C:\\Users\\name\\frame.png` therefore still survive `durable_dict()` and checkpoint/trace persistence. Please detect URI/path tokens behind punctuation or prefixes and add nested plus post-construction-mutation regressions.

rogercloud (approved):
Approved. The durable-reference design is coherent, the previously reported persistence, cache-accounting, and resource-budget issues are addressed, and all current CI checks pass.

One P2 persistence-hardening follow-up remains and is documented inline. It should be fixed, but I am approving with the expectation that the author follows up.

src/xagent/core/context_ref.py
qinxuye: Fixed in e602a200. Metadata string sanitization now detects embedded file URIs, private POSIX paths behind punctuation, Windows drive paths, UNC paths, and Windows-rooted paths while preserving normal remote URLs. Added nested metadata and post-construction durable_dict mutation regressions for the reported bypasses. The focused context-reference/agent test suite and full pre-commit run pass.

PR #1042, opened 2026-07-28
fix(frontend): strip debug console logs from production builds

### LINKED ISSUE #1040: fix(frontend): strip debug console logs from production builds
## Description

Production builds keep every debug `console.*` call. In the embedded widget the
whole WebSocket traffic — user messages, agent output, tool arguments and tool
results — is therefore printed to the browser console of the third-party site
that hosts the widget.

## Root cause

- `next.config.mjs:33` — `compiler: { removeConsole: false }`
- `src/contexts/app-context-chat.tsx`, the context the widget chat runs on, holds
  ~50 debug statements. Several dump entire frames:
  lines 3872, 3876, 4001, 4244, 4272, 4310, 4323, 4367, 4394
  (`console.trace('Original message:', JSON.stringify(message), …)`),
  plus 1693, 1729, 2262, 4434, 4665.

## Fix

    compiler: { removeConsole: { exclude: ["error", "warn"] } },

SWC drops `log` / `debug` / `info` / `trace` at build time; `next dev` is
untouched, so local debugging is unaffected.

Then review the remaining `console.error` / `console.warn` calls on the widget
path so they log a message rather than a serialized frame.

## Tests

The built bundle contains none of the debug markers (e.g. `Original message:`).

bluefish-08:
Closes #1040

## Problem

Production builds keep every debug `console.*` call. In the embedded widget the whole WebSocket traffic — user messages, agent output, tool arguments and tool results — is printed to the browser console of the third-party site hosting the widget. `src/contexts/app-context-chat.tsx` alone holds ~50 debug statements, several of which dump entire frames (`console.trace('Original message:', JSON.stringify(message), …)`).

## Change

`next.config.mjs`: `removeConsole: false` → `removeConsole: { exclude: ["error", "warn"] }`.

SWC drops `log` / `debug` / `info` / `trace` at build time. `next dev` is untouched, so local debugging is unaffected.

## On the remaining error/warn calls

The issue also asks to review the `console.error` / `console.warn` calls left on the widget path. Checked, no change needed: `app-context-chat.tsx` has 3 (lines 4546, 4654, 4658) and `use-websocket.ts` has 5 — all of them log a message plus an `Error` object, none serializes a frame.

## Verification

`npm run build`, then over `out/`:

| marker | before | after |
|---|---|---|
| `Original message:` | present | 0 |
| `console.trace` | present | 0 |
| `console.error` | present | 328 (kept) |

The 9 remaining `console.log` occurrences all come from third-party bundles (xlsx, katex, diff); SWC does not transform `node_modules`. No project source is among them.

### DIFF
--- a/frontend/next.config.mjs
+++ b/frontend/next.config.mjs
@@ -31,9 +31,11 @@ const nextConfig = {
   // Strip debug logging from production bundles. The embedded widget runs on
   // third-party sites, and the chat context logs whole WebSocket frames (user
   // messages, agent output, tool arguments and results) into their console.
-  // SWC drops log/debug/info/trace at build time only — `next dev` keeps them.
+  // Next hands this straight to SWC with no development guard of its own
+  // (build/swc/options.js), so `next dev` has to be excluded explicitly for
+  // local debugging to keep its log/debug/info/trace output.
   compiler: {
-    removeConsole: { exclude: ["error", "warn"] },
+    removeConsole: isDev ? false : { exclude: ["error", "warn"] },
   },
   // 解决开发模式错误
   reactStrictMode: true,
--- a/frontend/next.config.mjs
+++ b/frontend/next.config.mjs
@@ -28,9 +28,12 @@ const nextConfig = {
   experimental: {
     optimizeCss: false,
   },
-  // 确保CSS正确处理
+  // Strip debug logging from production bundles. The embedded widget runs on
+  // third-party sites, and the chat context logs whole WebSocket frames (user
+  // messages, agent output, tool arguments and results) into their console.
+  // SWC drops log/debug/info/trace at build time only — `next dev` keeps them.
   compiler: {
-    removeConsole: false,
+    removeConsole: { exclude: ["error", "warn"] },
   },
   // 解决开发模式错误
   reactStrictMode: true,


frontend/next.config.mjs
```diff
@@ -28,9 +28,12 @@ const nextConfig = {
...
     optimizeCss: false,
   },
-  // 确保CSS正确处理
+  // Strip debug logging from production bundles. The embedded widget runs on
+  // third-party sites, and the chat context logs whole WebSocket frames (user
+  // messages, agent output, tool arguments and results) into their console.
+  // SWC drops log/debug/info/trace at build time only — `next dev` keeps them.
   compiler: {
-    removeConsole: false,
+    removeConsole: { exclude: ["error", "warn"] },
```
rogercloud: [P2][Non-blocking] This unconditional setting also strips `log`, `debug`, `info`, and `trace` during `next dev`, contrary to the stated intent that local debugging remain unaffected. Consider gating it with `removeConsole: isDev ? false : { exclude: ["error", "warn"] }`.

rogercloud (approved):
The production privacy fix works as intended. Approving with one non-blocking developer-experience follow-up.

frontend/next.config.mjs
bluefish-08: Good catch, and my claim in the description was wrong — I took "next dev is untouched" from the issue text without checking. Next passes `compiler.removeConsole` straight through to SWC in `build/swc/options.js:132` with no development guard of its own (unlike `emotion`/`styledComponents`, which do receive `development`), so dev really was losing its debug output.

Applied your suggestion in b0017b4f: `removeConsole: isDev ? false : { exclude: ["error", "warn"] }`, reusing the `isDev` already defined above for the output mode. Rebuilt to confirm production is unchanged — `Original message:` 0, `console.trace` 0, `console.error` 328.

PR #1043, opened 2026-07-28
fix(widget): remove mocked demo agents from the chat start screen

### LINKED ISSUE #1038: fix(widget): remove mocked demo agents from the chat start screen
## Description

The chat start screen shows three agents — "Researcher Agent", "Poster Maker",
"LinkedIn Post Creator" — that exist in no deployment. They are hardcoded design
placeholders, they are not clickable, and in the embedded widget they look like
agents the tenant configured.

## Root cause

`src/components/chat/ChatStartScreen.tsx:170` renders the "chat with agents"
block even when no list was passed:

    {(agents && agents.length > 0) || !agents ? (

and then falls through to the mocked markup at lines 213-235
(`chatPage.agents.researcher | poster | linkedin`, `src/i18n/locales/en.ts:218-220`).

Both callers that omit `agents` are affected:

- `src/components/widget/public-agent-chat-page.tsx:374` — widget and share chat
- `src/app/agent/[id]/page-client.tsx:152` — agent detail page

The block is only reachable when the agent has suggested prompts, because the
outer condition at line 130 requires a non-empty `prompts` array.

## Fix

Require a real list and delete the fallback:

    {agents && agents.length > 0 ? (

Remove the mocked markup (213-235) and the now-unused `chatPage.agents.*` keys
from every locale.

## Tests

`ChatStartScreen` renders no agents section when `agents` is `undefined` or
empty, and still renders the list when one is passed.

bluefish-08:
Closes #1038

## Problem

The chat start screen shows three agents — "Researcher Agent", "Poster Maker",
"LinkedIn Post Creator" — that exist in no deployment. They are hardcoded design
placeholders, they are not clickable, and in the embedded widget they look like
agents the tenant configured.

`ChatStartScreen.tsx:170` rendered the "chat with agents" block even when no list
was passed (`(agents && agents.length > 0) || !agents`) and then fell through to
mocked markup. Both callers that omit `agents` were affected:

- `src/components/widget/public-agent-chat-page.tsx:377` — widget and share chat
- `src/app/agent/[id]/page-client.tsx:152` — agent detail page

`src/app/task/page.tsx:243` passes a real list, so it never hit the fallback.

## Change

- Require a real list: `{agents && agents.length > 0 ? (`
- Delete the mocked markup and the now-unused `Smartphone` import
- Drop the `chatPage.agents.{researcher,poster,linkedin}` keys from `en.ts` / `zh.ts`
  (`cards.linkedin` and `cards.design_poster` are prompt-card copy and stay)

No behaviour change for the internal task page.

## Tests

New `ChatStartScreen.test.tsx`: the agents section is absent when `agents` is
`undefined` and when it is empty, and renders the passed agents otherwise.

`vitest run src/components/chat src/components/widget src/components/task` — 96 passed.

### DIFF
--- a/frontend/src/components/chat/ChatStartScreen.test.tsx
+++ b/frontend/src/components/chat/ChatStartScreen.test.tsx
@@ -1,0 +1,65 @@
+import React from "react"
+import { cleanup, render, screen } from "@testing-library/react"
+import { afterEach, describe, expect, it, vi } from "vitest"
+
+vi.mock("@/contexts/i18n-context", () => ({
+  useI18n: () => ({
+    t: (key: string) => key,
+  }),
+}))
+
+vi.mock("@/lib/utils", () => ({
+  getApiUrl: () => "http://api.local",
+}))
+
+vi.mock("@/components/chat/ChatInput", () => ({
+  ChatInput: () => <div data-testid="chat-input" />,
+}))
+
+vi.mock("@/components/ui/tooltip", () => ({
+  Tooltip: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipContent: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipProvider: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipTrigger: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+}))
+
+import { ChatStartScreen } from "@/components/chat/ChatStartScreen"
+
+const AGENTS_SECTION = "chatPage.sections.chatWithAgents"
+
+// The agents block only renders inside the prompts branch, so every case needs
+// a non-empty prompts list to reach it at all.
+const renderScreen = (agents?: React.ComponentProps<typeof ChatStartScreen>["agents"]) =>
+  render(
+    <ChatStartScreen
+      title="Support Agent"
+      prompts={["Summarize this page"]}
+      agents={agents}
+      onSend={vi.fn()}
+    />
+  )
+
+afterEach(() => {
+  cleanup()
+})
+
+describe("ChatStartScreen agents section", () => {
+  it("renders nothing when no agent list is passed", () => {
+    renderScreen(undefined)
+
+    expect(screen.queryByText(AGENTS_SECTION)).toBeNull()
+  })
+
+  it("renders nothing for an empty agent list", () => {
+    renderScreen([])
+
+    expect(screen.queryByText(AGENTS_SECTION)).toBeNull()
+  })
+
+  it("renders the agents that were passed", () => {
+    renderScreen([{ id: 7, name: "Billing Agent" }])
+
+    expect(screen.getByText(AGENTS_SECTION)).toBeTruthy()
+    expect(screen.getByText("Billing Agent")).toBeTruthy()
+  })
+})
--- a/frontend/src/components/chat/ChatStartScreen.test.tsx
+++ b/frontend/src/components/chat/ChatStartScreen.test.tsx
@@ -8,10 +8,6 @@ vi.mock("@/contexts/i18n-context", () => ({
   }),
 }))
 
-vi.mock("@/lib/utils", () => ({
-  getApiUrl: () => "http://api.local",
-}))
-
 vi.mock("@/components/chat/ChatInput", () => ({
   ChatInput: () => <div data-testid="chat-input" />,
 }))
@@ -26,6 +22,10 @@ vi.mock("@/components/ui/tooltip", () => ({
 import { ChatStartScreen } from "@/components/chat/ChatStartScreen"
 
 const AGENTS_SECTION = "chatPage.sections.chatWithAgents"
+// t() is mocked as identity, so the deleted demo cards would surface as their
+// literal keys — assert on those directly rather than only on the header they
+// happened to share a branch with.
+const MOCKED_AGENT_KEYS = /chatPage\.agents\./
 
 // The agents block only renders inside the prompts branch, so every case needs
 // a non-empty prompts list to reach it at all.
@@ -48,12 +48,14 @@ describe("ChatStartScreen agents section", () => {
     renderScreen(undefined)
 
     expect(screen.queryByText(AGENTS_SECTION)).toBeNull()
+    expect(screen.queryByText(MOCKED_AGENT_KEYS)).toBeNull()
   })
 
   it("renders nothing for an empty agent list", () => {
     renderScreen([])
 
     expect(screen.queryByText(AGENTS_SECTION)).toBeNull()
+    expect(screen.queryByText(MOCKED_AGENT_KEYS)).toBeNull()
   })
 
   it("renders the agents that were passed", () => {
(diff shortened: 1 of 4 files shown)


rogercloud (approved):
## Summary

`ChatStartScreen` gated its "chat with agents" block on `(agents && agents.length > 0) || !agents`, so callers that pass no `agents` prop fell through to a hardcoded fallback rendering three fake agent cards ("Researcher Agent", "Poster Maker", "LinkedIn Post Creator"). Those cards carried hover affordances but no click behavior, and the two public-facing surfaces (embedded/share widget chat and the public agent detail page) both omit `agents` — so tenants saw mock agents indistinguishable from their own configured ones. This PR tightens the gate to `agents && agents.length > 0`, deletes the fallback markup, drops the now-unused `Smartphone` import, removes the orphaned `chatPage.agents.{researcher,poster,linkedin}` keys from `en.ts`/`zh.ts`, and adds a unit test covering the undefined / empty / populated states. The internal task page always passes a real list and was never affected.

## Approach

The approach is sound. This is the minimal correct fix: it targets the actual root cause (a gating condition that treated "no data" as "show demo data") and resolves it by deletion rather than by adding a flag or an abstraction. The two alternatives are both worse for this changeset — wiring real tenant agents into the widget is a product decision well beyond a bug fix, and hiding the fallback behind a demo/storybook flag would add machinery purely to preserve dead code. The reformatted `agents.map(...)` body is content-identical to before (verified with `git diff -w --ignore-blank-lines`): only the wrapping ternary was removed and the block dedented, so `onAgentClick`, `isSelected` highlighting, `logo_url` construction, and the description tooltip are untouched.

Two design-level observations, neither blocking:

1. **The agents block remains nested inside the outer `prompts` gate.** `src/components/chat/ChatStartScreen.tsx:130` opens `{prompts && prompts.length > 0 && (...)}` and closes around line 217, with the agents section inside it. A caller passing non-empty `agents` but empty/undefined `prompts` would silently render no agents section at all. This is pre-existing — the same nesting is in base commit `3cb7aa7` — and this PR only changed the inner condition. It is also currently unreachable: all three callers (`src/app/task/page.tsx`, `src/app/agent/[id]/page-client.tsx`, `src/components/widget/public-agent-chat-page.tsx`) either always pass non-empty prompts or never pass agents. Worth a follow-up to hoist the agents block out from under the prompts gate, but out of scope here.

2. **Style: `src/components/chat/ChatStartScreen.tsx:170`** now reads `{agents && agents.length > 0 ? (...) : null}`. Since the else-branch content was deleted, the `: null` no longer carries meaning; `&&` would match the sibling `prompts` condition at line 130. This one *is* PR-introduced (the base had a real else-branch), but it is purely cosmetic with zero behavioral difference.

## Findings

No critical or major issues found. Everything below is nit-level and optional.

**A. Nit — `src/components/chat/ChatStartScreen.test.tsx`: test matrix and assertions leave a gap.** All three cases share a hardcoded `prompts={["Summarize this page"]}`, so the "non-empty `agents` + empty `prompts`" combination (design concern #1) is never exercised. Separately, the negative assertions only check that the section header is absent via `queryByText(AGENTS_SECTION)` rather than asserting the removed fallback strings are gone. Because `t` is mocked as identity, the old fallback would have rendered the literal keys `chatPage.agents.researcher` etc. — the current assertion catches that only because the fallback happened to live in the same conditional branch as the header, which is coincidental coupling rather than a direct regression guard. Consider adding a `prompts={[]}` + non-empty `agents` case, and asserting `queryByText(/chatPage\.agents\./)` is `null`.

**B. Nit — `src/components/chat/ChatStartScreen.test.tsx`: the reformatted `agents.map` body has no direct coverage.** The new tests verify the three-state gating only; `onAgentClick`, `isSelected` highlighting, absolute/relative `logo_url` construction, and the description tooltip remain untested. This is pre-existing untested code, not a regression — raised only because this PR moved and reindented it. The `git diff -w` equivalence noted above already establishes no behavior change, so this is optional.

**C. Nit — `src/components/widget/public-agent-chat-page.test.tsx` (untouched by this PR):** it fully mocks `ChatStartScreen`, so nothing at the widget-integration level observes that the mock agents are gone from the widget surface, which is where the bug was most visible. Only the new unit test exercises the real component. Observation, non-blocking.

**D. Nit — `src/components/chat/ChatStartScreen.test.tsx`: `vi.mock("@/lib/utils", ...)`** replaces the module wholesale instead of spreading `vi.importActual`. Safe today (the component only imports `getApiUrl`) and consistent with existing files such as `src/components/chat/ChatInput.test.tsx`, though other tests in the repo use the safer importActual-spread style. Not a new risk introduced here.

### Verified clean

- i18n cleanup is complete: zero remaining references to `chatPage.agents.*` repo-wide (including `docs/`); `en.ts` and `zh.ts` are the only locales, and both the `TranslationKey` type and the en/zh parity check in `translations.test.ts` would have caught an orphaned or mismatched key.
- `Smartphone` removal is safe — no other uses in the file; `Sparkles` and `Bot` remain imported and used.
- No TypeScript narrowing concern: `agents` is a destructured prop (`AgentCard[] | undefined`), never reassigned, and `agents.map` is called synchronously inside the truthy branch.
- No documentation references the removed keys or the three fake agent names.

Lean already. The new test file's mocking boilerplate matches the convention already used across ~25-30 component test files; factoring out a shared mock-utils helper for a single new file would be scope creep.

## Testing

Independently re-ran in the review worktree (not just taken from the PR description):
- `vitest run src/components/chat src/components/widget src/components/task` — **9 test files, 96 tests, all passed.**
- `tsc --noEmit` — clean, no errors.
- `npm run lint` (eslint) — clean, no errors or warnings.

All match the PR description's reported results.

Recommend merge, with the agents/prompts nesting (design concern #1) filed as a follow-up.

bluefish-08:
Thanks for the thorough pass. Took three of the points in ee5c513b, and left the rest with reasoning below.

**Applied**

- *Design observation 2* — `? (...) : null` is now `&&`, matching the sibling `prompts` gate at line 130.
- *Nit A (second half)* — both negative cases now also assert `queryByText(/chatPage\.agents\./)` is null. You're right that keying only off the section header was coincidental coupling; with `t` mocked as identity the deleted cards would have surfaced as their literal keys, so that is the direct regression guard.
- *Nit D* — dropped the `@/lib/utils` mock entirely. The component's only use is `getApiUrl`, which no test case reaches (none of the fixtures carry a `logo_url`), so the real module is both safer and shorter than an importActual spread.

**Not applied**

- *Nit A (first half)* — a `prompts={[]}` + non-empty `agents` case would pin the current nesting as a contract, and that nesting is exactly what design observation 1 says should change. I'd rather not have to delete the test in the same PR that fixes the behaviour.
- *Nit B* — `onAgentClick` / `isSelected` / `logo_url` / tooltip are pre-existing untested code, and your `git diff -w` check already establishes the move was behaviour-preserving. Adding coverage for them is worth its own PR, not this one.
- *Nit C* — agreed as an observation; the widget test mocks `ChatStartScreen`, so the unit test is the right level for this fix.

**Follow-up**

Design observation 1 (the agents block sitting under the `prompts` gate) — I'll file it as a separate issue rather than widening this PR.

PR #11, opened 2026-02-24
fix(nginx): avoid proxying /files page route to backend

bluefish-08:
**Summary**
- Remove the `location /files/` proxy rule from `docker/nginx.conf`.
- Keep `/api/files/...` traffic routed through the existing `/api/` backend proxy.

**Why**
In Docker deployment, `/files` is a frontend page route, but it was being forwarded to backend due to the `/files/` nginx rule, causing:
`{"detail":"Not Found"}`

**Result**
- `/files` now resolves to the Next.js frontend page as expected.
- File APIs (download/preview/upload) continue to work via `/api/files/...`.

PR #1044, opened 2026-07-28
fix: isolate Skill runtime database sessions

### LINKED ISSUE #1033: bug: Personal skill loading retains DB sessions across waiting agent runtimes
## Summary

Long-lived agent runtimes can retain a SQLAlchemy session after loading personal database-backed skills. Executions that enter `WAITING_FOR_USER` then keep one PostgreSQL connection `idle in transaction` for the lifetime of the waiting runtime.

Repeated scheduled executions can therefore consume the entire `QueuePool` (`pool_size=10`, `max_overflow=20`) and make unrelated authentication, task, and health endpoints fail with pool checkout timeouts.

This is a distinct session-lifetime path not covered by #889. That issue addressed request/runtime sessions held across MCP network awaits; this issue occurs after the earlier release boundary, when personal skill loading re-acquires a connection through a retained skill context.

## Owning path

1. `WebToolConfig.get_skill_scope_context()` returns a `SkillScopeContext` containing `db=self.db`.
2. `AgentService` copies that context into the long-lived service and execution adapter.
3. `XagentPersonalDbSkillProvider.list_records()` uses `context.db` to query `UserSkill` rows.
4. The query can occur after `release_db_connection_if_clean()` has already released the construction-time transaction, so it re-acquires a pooled connection.
5. No later ownership boundary releases or closes that transaction after skill initialization.
6. If the execution calls `ask_user_question`, the runtime remains `WAITING_FOR_USER` and retains the context and session.

## Reproduction

1. Use the default skill provider chain with personal database-backed skills enabled.
2. Configure a scheduled agent that calls `ask_user_question` and enters `WAITING_FOR_USER`.
3. Start more executions than the configured pool capacity without responding to the interaction.
4. Inspect PostgreSQL `pg_stat_activity`.
5. Call an unrelated database-backed endpoint such as login.

Observed result:

- One `idle in transaction` session accumulates per waiting runtime.
- The last statement is the `user_skills` query issued by the personal skill provider.
- At 30 retained sessions, the `10 + 20` pool is exhausted.
- Authentication and task persistence fail with `sqlalchemy.exc.TimeoutError`.
- Restart recovery can replay pending executions and rebuild the same retained session state.

## Required invariant

An `AgentService`, execution adapter, skill manager, or `WAITING_FOR_USER` runtime must never retain a live SQLAlchemy `Session` or pooled connection.

## Suggested fix boundary

- Do not store a live request/runtime `Session` in `SkillScopeContext` that escapes into a long-lived runtime.
- Resolve personal skill records through a short-lived, provider-owned session or a detached snapshot.
- Fully materialize skill metadata and file content before closing the session.
- Ensure waiting, paused, completed, failed, and recovery paths retain only detached skill data.
- Keep #889's release boundary as defense in depth, but do not rely on an early release when later lazy providers can re-acquire the same session.

## Verification criteria

- PostgreSQL regression test with more concurrent `WAITING_FOR_USER` executions than the configured pool capacity.
- Personal database-backed skills remain visible and readable.
- `pg_stat_activity` shows no accumulated `idle in transaction` sessions after skill loading.
- Login and health endpoints remain responsive while executions wait for user input.
- Restart/recovery of waiting executions does not retain database sessions.
- Tests cover normal completion, failure, cancellation, pause, and `WAITING_FOR_USER` lifecycle paths.

AlexLiu190625:
## Summary

- add a verified final handoff for ToolFactory and prebuilt AgentService paths so retained tool configuration keeps only detached identity and a session factory
- define detached read/write Skill provider contexts and a single provider invocation boundary with typed public failures and sanitized unknown failures
- keep database-backed Skill providers responsible for their own short-lived sessions and transactions
- cover HTTP routes and retained AgentService `waiting_for_user` execution with a real one-slot PostgreSQL pool

## Root cause

Tool construction could release a clean request session, then re-acquire its connection while loading later factory inputs such as MCP configuration. The resulting `WebToolConfig` and `AgentService` could outlive the request while still retaining the request Session, request object, and ORM User. A database-backed Skill provider then opened its own session and blocked against the retained connection when the pool was constrained.

## Architecture

Tool construction now ends at one verified ownership boundary. It consumes all construction-time database inputs first, releases clean caller-owned sessions, terminally closes factory-owned lazy sessions, detaches request and ORM objects, and retains only a session factory plus scalar identity. Cache-hit and failure paths use the same boundary.

Read and write Skill providers receive frozen contexts containing only `user_id` and copied scalar metadata. Database-backed providers re-resolve authorization and own their complete operation: open, authorize, materialize, commit or roll back, and close. Team write routes share one typed invocation boundary; unrecognized or malformed provider failures produce a stable sanitized response.

## Provider migration

- `SkillScopeContext` and `SkillWriteContext` contain only `user_id` and exact scalar metadata values; request objects, ORM entities, Sessions, and session factories are not accepted.
- Metadata is non-authoritative. Providers must resolve current scope and authorization from `user_id` inside their own dependency boundary.
- Database-backed providers must own their sessions and transactions and return fully materialized results.
- Expected public write failures use `SkillWriteProviderError` with `FORBIDDEN` or `INVALID_REQUEST`; unexpected failures are logged and returned as a stable HTTP 500.

## Impact

There is no database schema, migration, or successful API response-shape change. Downstream provider implementations that used request-owned fields from the old context shape must adopt the detached provider contract above.

## Verification

- all tests mapped to the changed Skill, ToolFactory, WebToolConfig, and AgentService owners passed; one optional environment-variable-driven provider test was skipped
- Docker PostgreSQL E2E passed with `pool_size=1`, `max_overflow=0`, two retained AgentService instances, explicit Skill and MCP selection, and exact `waiting_for_user` results
- final handoff, detached-context, and exception-redaction mutation checks failed at their intended assertions and were restored
- Ruff, formatting, mypy, isort, codespell, diff checks, and pre-commit passed across every changed file

Fixes #1033.


rogercloud (changes requested):
## Summary

This PR fixes #1033, where an agent runtime parked in `WAITING_FOR_USER` held a live SQLAlchemy session (and its pooled connection) for the entire wait, eventually starving the pool and breaking unrelated endpoints such as login with `TimeoutError`. The core change is an ownership split in the skill layer: `SkillScopeContext` (`src/xagent/skills/library.py:20-39`) becomes a frozen, detached, read-only identity carrier (`user_id` + scalar-only frozen metadata, rejecting non-scalars in `__post_init__`), while request-owned write resources move to a separate `SkillWriteContext`. The personal-skill provider (`src/xagent/skills/personal_db.py:48-57`) now opens its own short-lived session via `run_db_io_cancellation_safe`, eager-loading everything it needs (`selectinload(UserSkill.files)`, `bytes(file.content)`, `deepcopy` of metadata) before closing. A new `src/xagent/web/services/skill_runtime.py` adds an explicit release/handoff boundary that the HTTP skill routes (`skills.py`, `skill_hub.py`) depend on. Diff is 14 files, +872/-115, including new unit tests, a docker-gated e2e Postgres pool test, and a CI workflow change.

## Approach verdict

**Acceptable with reservations.**

The read/write context split is a genuine ownership model rather than a point patch, and the detachment pattern in `personal_db.py` correctly mirrors the existing `WebToolConfig.prepare_factory_runtime` precedent, with no `DetachedInstanceError` exposure. Design-level reservations:

- **The new release boundary is wired only into the HTTP route path, not the agent-runtime path the issue is actually about.** `get_skill_runtime_scope` in `src/xagent/web/services/skill_runtime.py:41-54` performs the handoff for FastAPI routes, but the runtime-path equivalent `WebToolConfig.get_skill_scope_context` (`src/xagent/web/tools/config.py:1540-1544`) performs no release or handoff at all. See Critical #1 — the reported bug plausibly survives on its own reported path.
- **No migration or compatibility path for the narrowed read context.** `SkillScopeContext` dropped `db`/`user`/`request` with no alias or deprecation shim, and the overlay-provider hook is a documented out-of-tree extension point (see Major #4).
- **The two scope builders are asymmetric.** The route dependency populates `metadata["team_id"]`; the runtime builder populates nothing, silently degrading team-scoped skills on the agent path (see Major #5).
- The fail-closed handoff error is never handled anywhere, so its "never fail toward data loss" intent does not hold for external write providers (see Major #3).

## Findings

### Critical

**1. Pool-exhaustion gap remains on the `AgentService` runtime path — the path #1033 describes**

`src/xagent/core/tools/adapters/vibe/factory.py:277-311`, `src/xagent/web/tools/config.py:1715-1749` and `:2989-3033`, `src/xagent/skills/personal_db.py:48-57`

Traced call chain: `AgentService._ensure_tools_initialized` → `ToolFactory.create_all_tools` → `WebToolConfig.prepare_factory_runtime()`, which releases the request-scoped connection at `src/xagent/web/tools/config.py:1741` — this is the #1033 fix. But within the *same* `create_all_tools()` call, MCP tool creation then runs unconditionally and calls `_load_mcp_server_configs()`, which issues `self.db.query(MCPServer)...` directly against `self._live_db` (`src/xagent/web/tools/config.py:2989-3033`), re-checking out a pooled connection on the session that was just released. Nothing releases it a second time before skill tools are built or invoked, and `get_skill_scope_context` (`src/xagent/web/tools/config.py:1540-1544`) — unlike the HTTP dependency — never calls any release/handoff helper. When a skill tool subsequently runs, `personal_db.py` opens an *independent* second session via `get_session_local()`, so two pool slots are held concurrently on exactly the runtime path the issue reports. This needs no unusual timing: MCP tools are enabled by default.

Compounding this: the PR's own e2e test only drives the HTTP route path, so this gap is entirely unexercised by the PR's verification (see Major #8). I'd want either the release/handoff boundary threaded into `get_skill_scope_context`/`create_all_tools` after MCP config loading, or an explicit rationale for why the runtime path is considered out of scope for this fix.

### Major

**2. Lost graceful degradation when the DB is not initialized takes down the entire skill list**

`src/xagent/skills/personal_db.py:48-57`, `src/xagent/skills/library.py:130-134`, `src/xagent/skills/manager.py:76-97`

The old code short-circuited with `if db is None or user_id is None: return []`, so an unavailable DB silently yielded no personal skills. The new code calls `get_session_local()` unconditionally whenever `user_id` is set, which raises `RuntimeError("Session Local is not initialized. Call init_db() first.")`. That exception is caught nowhere: `CompositeSkillLibraryProvider.list_records` (`src/xagent/skills/library.py:130-134`) has no per-provider guard, and the only `try/except` in `SkillManager.reload()` wraps per-record parsing, not the `list_records()` call itself at `src/xagent/skills/manager.py:81`. So a personal-provider failure now nukes the whole skill list, filesystem skills included.

Mitigating: every current production entry point (web server, background workers) calls `init_db()` at startup before any request carrying a real `user_id`, so this is not reachable through any existing production path — it was only reproduced by constructing `WebToolConfig` directly in a unit test. Still worth fixing as a robustness regression: either catch `RuntimeError` around the session-local lookup, or restore a per-provider `try/except` in the composite provider so one layer can't take out the others.

**3. `SkillRuntimeSessionBoundaryError` is unhandled, yielding an opaque 500 and still losing the write**

`src/xagent/web/services/skill_runtime.py:16-26`, `src/xagent/web/api/skill_hub.py:708`, `:942`

`handoff_skill_runtime_session` raises `SkillRuntimeSessionBoundaryError` whenever the caller's session still carries dirty/new/deleted ORM state at the boundary, and nothing in the codebase catches it — `app.py`'s catch-all re-raises for non-`/v1/` paths, so it surfaces as a generic 500 with no boundary-specific log line and no `HTTPException` mapping. Worse, `get_db`'s `finally: db.close()` then rolls the uncommitted work back anyway, so the documented "fail closed, never toward data loss" intent doesn't actually hold: the data is still lost, just behind an unhelpful error.

Reachable only via an externally-registered team/SaaS `SkillWriteProvider` that performs DML without committing; the in-repo personal-scope write path commits before the boundary and is unaffected. Suggest catching the error at the route layer, logging the offending boundary, and mapping it to a 5xx with a diagnosable message.

**4. `SkillScopeContext` narrowing is a breaking change for a documented out-of-tree extension point**

`src/xagent/skills/library.py:20-39`, `src/xagent/skills/utils.py:62-86`, `src/xagent/skills/manager.py:81`

`db`/`user`/`request` were removed outright with no alias or deprecation shim. Confirmed no in-repo provider still reads them (the one first-party DB provider was migrated in this same PR), so this is not a live bug here. But `skills/utils.py` and `skill_hub.py` both carry comments describing the overlay hook as a deliberate SaaS extension point ("SaaS uses this to add team skills below personal skills"), and because `manager.reload()`'s `list_records()` call is unguarded (same mechanism as #2), an out-of-tree provider on the old shape will raise an unhandled `AttributeError` that takes down the whole skill surface on upgrade. At minimum this needs a migration note in the PR description for the downstream maintainer.

**5. Agent-runtime skill scope silently drops team identity that the HTTP path retains**

`src/xagent/web/tools/config.py:1540-1544` (vs. `src/xagent/web/services/skill_runtime.py:41-54`)

`get_skill_scope_context()` previously passed the full `user` object, from which an overlay could derive the team. It now returns a bare `SkillScopeContext(user_id=self._user_id)` with no metadata, while the sibling HTTP dependency added in this same PR explicitly sets `metadata["team_id"]` from `current_user._saas_team_id`. `self._user` is still available on `WebToolConfig` (`src/xagent/web/tools/config.py:1103`), so the data is present and simply not threaded through.

Net effect for a team-scoped overlay provider filtering on `context.metadata["team_id"]`: team skills appear in `GET /api/skill-hub/installed` but are absent from the agent runtime's tool list — the agent silently runs without its team's skills, no error raised. Note this is a *silent omission*, not the `AttributeError` one might first assume; the class was narrowed uniformly and nothing in-repo touches `.db`/`.user`. Fix by populating `team_id` symmetrically in the runtime builder, ideally via a single shared helper so the two builders can't drift again.

**8. No test drives the actual `AgentService` / `WAITING_FOR_USER` scenario from the issue**

`tests/e2e/test_skill_runtime_session_postgres.py`, `tests/web/services/test_skill_runtime.py`

Credit where due: these tests make real behavioral assertions (`pool.checkedout() == 0`), not merely structural ones, and `SkillScopeContext` being a two-field frozen dataclass is a stronger guarantee than `test_webtoolconfig_identity.py`'s `assert not hasattr(context, "db")` suggests on its own. But coverage stops at the HTTP route dependency: nothing exercises `AgentService`/the execution adapter, and nothing simulates a `WAITING_FOR_USER` pause across multiple concurrent waiting executions — which is precisely the reproduction in #1033. Given Critical #1, this is the gap that matters most: the one path still holding two slots is the one path with no test.

### Minor

**6. `SkillScopeContext` is no longer deepcopy/pickle-able**

`src/xagent/skills/library.py:39`

Wrapping metadata in `MappingProxyType` makes the context unpicklable — verified by execution: `copy.deepcopy(SkillScopeContext(user_id=7, metadata={"team_id": 1}))` raises `TypeError: cannot pickle 'mappingproxy' object`. No current caller deepcopies or pickles `AgentService.skill_scope_context` or `execution_adapter.config.skill_scope_context`, so this is purely latent — but it will bite the first time agent-state checkpointing or snapshotting includes this context. A `dict`-copying `__deepcopy__`, or a frozen mapping type that supports pickling, would close it cheaply.

**7. Half of the new e2e regression test doesn't guard against the bug it claims to**

`tests/e2e/test_skill_runtime_session_postgres.py:147-179`

The `/api/skills/` assertions are genuine: the pre-fix route minted an uncached second session via `get_session_local()()` while the `get_current_user`-derived session held the only slot, so these assertions do fail against the old code at `pool_size=1`. The `/api/skill-hub/installed` assertions do not: that route already resolved `db` through the same cached `Depends(get_db)` (one session per request) and passed it straight into `context.db`, so it never opened a second connection before or after this PR — those assertions pass against the old, buggy code too. Either drop them or point them at something that actually regressed.

**9. Redundant double handoff calls, plus one dead parameter**

`src/xagent/web/services/skill_runtime.py:41-54`, `src/xagent/web/api/skills.py:59-69`, `src/xagent/web/api/skill_hub.py:291`

`get_skill_runtime_scope` already calls `handoff_skill_runtime_session` during dependency resolution, then both `skills.py::_request_skill_manager` and `skill_hub.py::_get_scoped_manager` call it again on the same cached session. Harmless (the repeat hits the `not db.in_transaction()` no-op branch) but it obscures which call site is the authoritative boundary — worth collapsing to one. Separately, in `skills.py` the `_request` parameter of `_request_skill_manager` is genuinely unused in the body, yet all three routes still declare `request: Request` purely to pass it through; that plumbing can go. (This does *not* apply to `skill_hub.py`, where `request` is used elsewhere in `_get_scoped_manager`.)

**10. Scalar-only metadata enforcement is undocumented**

`src/xagent/skills/library.py:31-39`

`__post_init__` rejects any non-`str`/`int`/`float`/`bool`/`None` metadata value with `TypeError`, unconditionally and with no opt-out. I think this is the right call for a security boundary — failing loud beats silently dropping team scoping, and it structurally guarantees no live `Session` can hide in the context. It's just not stated anywhere outside the type alias. One line on the class docstring would save the next contributor a confused debugging session. Not currently triggerable: the only production caller supplies a lone scalar `team_id`.

**11. `Mapping` import inconsistency between the two files this PR touches**

`src/xagent/skills/library.py:15` imports `Mapping` from `typing`; the new sibling `src/xagent/web/services/skill_runtime.py:5` imports it from `collections.abc`. The repo has no enforced convention (roughly 40 files one way, 14 the other), so this isn't a style violation — just an inconsistency introduced inside a single PR. Cosmetic.

Lean already. A dedicated mechanical over-engineering pass found nothing worth changing: the `SkillScopeContext`/`SkillWriteContext` split, the detachment pattern in `personal_db.py`, the `skill_hub.py` read/write context refactor, and the CI multi-image retry loop are all non-speculative and appropriately sized.

## Testing

Run in a prepared review worktree with a fresh venv, all passing:

- `tests/skills/`
- `tests/web/services/test_skill_runtime.py`
- `tests/web/tools/test_webtoolconfig_identity.py`
- `tests/skills/test_skill_hub_security.py`

One test skipped because `XAGENT_TEST_POSTGRES_URL` is not configured in this environment.

Not run, and worth flagging:

- The docker-gated e2e Postgres test (`tests/e2e/test_skill_runtime_session_postgres.py`) could not be executed here — no Docker/Postgres available. This matters more than usual: that test is the PR's primary behavioral evidence, and per Critical #1 the remaining pool-exhaustion gap lives on the `AgentService` path that this test does not cover in the first place.
- Ruff/mypy results were not comparable in this environment (version and config mismatches), so I make no claim about lint or type-check status.

**Recommendation:** request changes, primarily on Critical #1 — either extend the release boundary to the `AgentService`/tool-factory path with a test that reproduces the `WAITING_FOR_USER` scenario from #1033, or explicitly scope that path out with a follow-up issue. Major #2, #3 and #5 are worth addressing in this PR; #4 needs at least a migration note in the description.

AlexLiu190625:
Implemented the review follow-up:

- added a verified final ToolFactory/AgentService handoff after all construction-time database reads, including explicit MCP selection with zero active servers
- detached read and write provider contexts, centralized typed/sanitized provider failures, registered the session-boundary handler, and documented the downstream provider migration
- added a real one-slot PostgreSQL regression covering two retained AgentService executions through `waiting_for_user`
- removed dead request plumbing while retaining the two handoffs that protect different re-acquisition windows

Intentionally not adopted:

- silent fallback when a registered database provider has no initialized database: production initializes the database before registration, while core-only mode should omit that provider; swallowing the failure would hide a broken Skill layer
- serialization support for the frozen context: it is an in-process boundary object with no current checkpoint or transport consumer
- removal of the Skill Hub PostgreSQL assertion: it still protects the public route's provider-owned session invariant
- collapsing dependency-time and pre-manager handoffs: the first releases the authenticated dependency transaction, while the second protects against later route work re-acquiring it
- an import-only style change: the repository has no single enforced import source for that type

Team scope was addressed by removing request-derived tenant metadata from both paths. Metadata is non-authoritative; downstream providers must re-resolve current authorization from `user_id` in their own short-lived session.

rogercloud (commented):
## Re-review: update summary

Re-reviewed at `113eaa50` (previous review was at `3a3f4872`). This revision adds 4 commits — *release tool factory database sessions*, *isolate skill provider runtime contracts*, *cover retained skill runtime pool handoff*, *preserve detached tool policy handoff* — totalling **+2011/-180 across 21 files**, with most of the growth in new tests.

**Headline: the original critical pool-exhaustion gap is genuinely fixed, with strong evidence — but the fix itself introduces a new pooled-connection leak on a *different* route (`GET /api/tools/available`).** The connection-handoff mechanism added for the AgentService runtime path clears the snapshot state that this other route depends on, causing it to mint a second pooled session that nothing ever closes. That is the same class of failure this PR exists to eliminate, just relocated, so it needs to be resolved before merge rather than tracked as follow-up.

## New findings in this revision

### HIGH — the connection-handoff fix leaks a second pooled connection on `GET /api/tools/available`

The route's flow is now self-conflicting:

1. `src/xagent/web/api/tools.py:302-306` calls `ToolFactory.create_all_tools`.
2. Per the Critical #1 fix, that method's `finally` (`src/xagent/core/tools/adapters/vibe/factory.py:277-321`) calls `handoff_factory_runtime()` (`src/xagent/web/tools/config.py:1827-1876`), which points `self._db_factory` at a session factory bound to the same engine, nulls `_live_db`, and clears `_factory_runtime_snapshot`.
3. Immediately after, `src/xagent/web/api/tools.py:332-338` calls seven getters: `get_vision_model`, `get_image_models`, `get_video_models`, `get_asr_models`, `get_tts_models`, `get_sound_effect_models`, `get_music_models`. Each one consults `self._factory_runtime_snapshot` first — now `None` — so all seven fall through to `_load_*_models()`, which reads `self.db`.
4. The `db` property sees `_live_db is None` with `_db_factory` set, so it **mints a fresh `_lazy_db` session** on first access and reuses it for the remaining getters.

Nothing closes that session: `grep -n "tool_config.close" src/xagent/web/api/tools.py` returns no hits. Meanwhile the handler keeps using its original FastAPI-injected `db` from `Depends(get_db)` — a different object — later in the same function for `db.query(ToolUsage)` (`src/xagent/web/api/tools.py:361`) and `db.query(ToolConfig)` (`src/xagent/web/api/tools.py:374`).

Net effect: this route now checks out two separate pooled connections per request and only ever returns one, relying on GC to eventually reclaim the orphaned `_lazy_db`. Under a constrained pool — precisely the `pool_size=1, max_overflow=0` configuration this PR's own e2e test uses elsewhere — the route can self-deadlock waiting for a slot its own earlier code is still holding. Under normal production pool settings it silently doubles the route's per-request connection cost.

This is also a real coverage gap. The one test that drives these getters against a real snapshot (`tests/web/tools/test_web_tool_config_session_factory.py:982-1073`) calls them right after `prepare_factory_runtime()` and never calls `handoff_factory_runtime()`, so it never reaches the post-handoff state the live route hits. The route's own API tests (`tests/web/api/test_tools_api.py`) monkeypatch `create_all_tools` wholesale, bypassing the real prepare/handoff logic entirely. No test asserts `pool.checkedout()` for this route.

Three viable fix directions, any one of which closes it:
- Have `_apply_factory_runtime_snapshot` populate the six `_cached_*` media-model fields from the snapshot, so post-handoff getters serve from cache instead of touching `self.db`.
- Have the route call `tool_config.close()` after the getters run.
- Have the getters cache-and-close on the fallback path.

Please also add a test that pins the post-handoff getter path and asserts `pool.checkedout()` for this route, so the regression can't reappear.

### MEDIUM — `WebToolConfig.close()` can raise from a bare `finally` and shadow the real exception

`close()` (`src/xagent/web/tools/config.py:1995-2003`) raises a bare, message-less `ToolFactoryRuntimeSessionBoundaryError()` when both `lazy_db.close()` and `lazy_db.invalidate()` fail. `src/xagent/core/tools/adapters/vibe/agent_tool.py:2022-2081` calls `tool_config.close()` from a bare `finally:` with no protective `except`, and that `try` body wraps both `AgentService(...)` construction and `execute_task(...)`.

By standard Python semantics, if the try body raises *and* the finally's `close()` also raises, the finally's exception replaces the original — so a diagnostically useful error gets swapped for a message-less `ToolFactoryRuntimeSessionBoundaryError`. Triggering it requires both `close()` and `invalidate()` to fail on the lazy session, which realistically means connection-level failure (DB down, network partition, pool exhaustion at the connection level). That makes it comparatively rare, but note it is *more* likely to co-occur with genuine DB-related failures in the try body — i.e. it destroys the original error in exactly the cases where that error matters most for debugging. Wrapping the `close()` call in a `try/except` that logs rather than propagates (and giving the exception a real message) would resolve it.

### Low-priority robustness note (not a production bug)

`handoff_factory_runtime` hard-codes `live_released = True` for a non-`Session` `db` object with no actual verification, nulls `_live_db` anyway, and never populates `_db_factory` (which requires `isinstance(live_db, Session)`) — leaving the config permanently DB-less with `self._user` as `None`, so later tool-override/allowlist refreshes silently degrade to empty rather than raising. I verified this is **not reachable in production**: every real `WebToolConfig(...)` construction site (`src/xagent/web/api/tools.py`, `src/xagent/web/api/agents.py`, `src/xagent/web/api/chat.py`, `src/xagent/core/tools/adapters/vibe/agent_tool.py`) passes either a genuine SQLAlchemy `Session` or explicit `None`. Only test-harness fakes (e.g. `PendingSession` in `tests/web/tools/test_web_tool_config_session_factory.py`) hit this path. A defensive log/warning when `handoff_factory_runtime` sees a non-Session, non-None `live_db` would be nice, but nothing here blocks merge.

## Prior findings — verified status

### Fixed
- **Critical #1 — pool exhaustion on the AgentService runtime path.** `handoff_factory_runtime()` (`src/xagent/web/tools/config.py:1827-1876`) now performs a real, verified release of both `_live_db` and `_lazy_db`, fails closed with `ToolFactoryRuntimeSessionBoundaryError` when it cannot verify a clean release, and `ToolFactory.create_all_tools`'s `finally` (`src/xagent/core/tools/adapters/vibe/factory.py:277-321`) invokes it *after* MCP config loading, which closes the original gap. Verified by running `tests/core/agent/test_agent_service_tool_refresh.py` plus `tests/core/tools/adapters/vibe/test_factory_sandbox_release.py` — 21 passed, with real constrained-pool assertions. The new e2e test `test_retained_agent_services_wait_without_holding_postgres_pool` reproduces the actual `WAITING_FOR_USER` scenario from #1033 end-to-end against a real one-slot Postgres pool. (See the HIGH finding above for the new problem this same mechanism creates on another route.)
- **Major #3 — unhandled `SkillRuntimeSessionBoundaryError`.** Now has a dedicated handler at `src/xagent/web/app.py:738` that logs with `exc_info` and returns a deliberate 500.
- **Major #4 — undocumented breaking change to the SaaS extension point.** `src/xagent/skills/README.md:84-105` documents the new context contract, and the PR description carries a migration note.
- **Major #8 — no test drove the real WAITING_FOR_USER scenario.** Closed by the new e2e test above.
- **Minor #10 — undocumented scalar-only metadata enforcement.** Now documented in both docstrings and the README.

### Partial
- **Minor #9 — redundant double handoff plus dead parameter.** The dead `_request` parameter is gone from `skills.py`, but the redundant `handoff_skill_runtime_session` calls remain in both `skills.py::_request_skill_manager` and `skill_hub.py::_get_scoped_manager:286`. Harmless because the call is idempotent, but it still leaves the authoritative call site ambiguous.

### Not fixed
- **Major #2 — lost graceful degradation when the DB is not initialized.** No changes to `personal_db.py` or `manager.py` in this range. `get_session_local()` still raises unconditionally when `user_id` is set, and neither `CompositeSkillLibraryProvider.list_records` nor `SkillManager.reload()` catches it — so the entire skill list dies, filesystem skills included, instead of degrading. Not reachable in production today since every real entry point calls `init_db()` at startup, but cheap to fix.
- **Minor #6 — `SkillScopeContext` is not deepcopy/pickle-able.** Re-verified by execution: still raises `TypeError: cannot pickle 'mappingproxy' object`. Latent; no current caller depends on it.
- **Minor #7 — half the e2e test does not guard the bug it claims to.** The `/api/skill-hub/installed` assertions are unchanged and still pass against pre-fix code. Effectively superseded by the much stronger new e2e test.
- **Minor #11 — `Mapping` import inconsistency.** Still split between `typing.Mapping` and `collections.abc.Mapping`.

## Needs an explicit answer before merge: the `team_id` removal (Major #5)

The HTTP-route/runtime asymmetry is gone, but not by making the two paths symmetric — `_saas_team_id`/`team_id` handling was **deleted from both paths** in `src/xagent/web/services/skill_runtime.py`, which a new test confirms by explicitly asserting that both builders now ignore `_saas_team_id`. If any SaaS or overlay skill provider relied on `metadata['team_id']` for team-scoped filtering, that capability is now gone entirely rather than repaired on the deficient path. Please confirm whether team-scoped overlay skill filtering is still required, and if so, where it is now expected to happen.

- `tests/e2e/test_skill_runtime_session_postgres.py:206-233` and `tests/web/tools/test_web_tool_config_session_factory.py:699-726` each hand-roll an object-graph walker. Replace both with one shared BFS over `gc.get_referents()`.
- `src/xagent/core/agent/service.py:74-79` and `src/xagent/core/agent/service.py:668-682` duplicate the same finalizer-lookup pattern. Extract a single helper.

## Recommendation

The core fix is solid and well-evidenced, but I can't approve this as-is: **the HIGH finding above is a new bug shipped by this very fix, on a different route than the one being repaired, and it is the same class of pool-exhaustion problem this PR exists to solve.** Merging a pool-exhaustion fix that opens a new pooled-connection leak elsewhere would undercut the whole change. Please resolve F1 — plus a test that pins the post-handoff getter path and asserts `pool.checkedout()` for `GET /api/tools/available` — before merge.

The MEDIUM `close()`-in-`finally` shadowing issue is worth fixing in the same pass: it lives in the same file and shares the theme of "the fix's own cleanup path has a bug," though on its own it would not block merge.

The two earlier follow-up asks still stand, but are secondary to the HIGH finding:
1. **Major #2** — restore graceful degradation when the DB is not initialized, so a missing DB does not take out filesystem skills too.
2. **`team_id` removal** — confirm whether team-scoped overlay skill filtering is still a requirement, and where it now happens.

AlexLiu190625:
Follow-up to the latest review:

- Fixed the post-handoff database re-acquisition on `GET /api/tools/available`: all retained model reads now use detached state, the route owns terminal cleanup, and the real one-slot PostgreSQL regression proves both the pre-fix second-checkout timeout and the fixed zero-checkout result.
- Fixed cleanup exception shadowing through one shared runtime-cleanup boundary: an active primary exception is preserved, cleanup failures are logged, and a standalone cleanup failure has a stable typed error.

The remaining suggestions are intentionally not included. Non-SQLAlchemy database objects occur only in test doubles, while production constructors pass a real `Session` or `None`. Database initialization is guaranteed by current production entry points; broadly swallowing provider failures would hide database and authorization faults. `SkillScopeContext` has no serialization consumer. The two HTTP handoffs guard distinct re-acquisition windows. The two object-graph walkers verify different ownership boundaries, and the duplicated finalizer lookup is intentionally local to its two lifecycle owners. Team authorization remains downstream-owned: SaaS providers must re-resolve current membership from `user_id` inside their own short-lived session.

rogercloud (approved):
## Re-review (round 3) — `23cc9137`

### Update summary

This round is tightly scoped to the two blocking findings from round 2: **F1 (HIGH — connection leak on `GET /api/tools/available`)** and **F3 (MEDIUM — `WebToolConfig.close()` shadowing the real exception from a bare `finally`)**. Seven commits landed (+1761/−268 across 10 files), all in the tool-factory / media-model connection-lifecycle code and its tests.

**Both blocking findings are fixed**, and not merely plausibly fixed — each has targeted regression tests that fail on the pre-fix behavior and assert the exact invariant that was broken. From a correctness standpoint this PR is now in a mergeable state.

The skills-layer items (A, B, C, D, E) are untouched this round, which is expected: none of `src/xagent/skills/personal_db.py`, `library.py`, `manager.py`, `src/xagent/web/services/skill_runtime.py`, `skill_hub.py`, or `skills.py` appear in this update's changed-file set. The author has posted explicit positions on all of them, so these are deliberate decisions rather than missed items.

---

### This round's fixes — verified status

#### F1 — connection leak on `GET /api/tools/available` — FIXED

Two independent mechanisms now close this hole, and both are needed:

1. **Terminal cleanup owned by the route.** The handler body is now wrapped in the new shared helper `run_with_tool_runtime_cleanup(build_tool_listing_response, tool_config.close, ...)` (`src/xagent/core/tools/adapters/vibe/config.py:38-63`). The wrapped body covers the *entire* route body — all seven media-model getters **and** the later `db.query(ToolUsage)` / `db.query(ToolConfig)` calls — so `tool_config.close()` is guaranteed to run exactly once in a `finally`, closing any lazily-opened session before the handler returns, on both the success and failure paths.

2. **No post-handoff DB reads are possible at all.** `handoff_factory_runtime()` (`src/xagent/web/tools/config.py:1951-1965`) now freezes a `_RetainedFactoryModelState` snapshot of every media-model value at handoff time — including categories whose load-plan flag was `False`, which are retained as `None`/empty. Every getter (`get_vision_model`, `get_image_models`, …) routes through `_get_factory_model_value` / `_get_factory_model_mapping` (`src/xagent/web/tools/config.py:1327-1370`), which resolves live snapshot → retained state → and only falls back to `self.db` while `_factory_runtime_handed_off` is still `False`. Once handed off, getters return `None`/`{}` rather than re-acquiring a connection. This closes the class of bug, not just the one observed call path.

Test evidence:

- `tests/web/tools/test_web_tool_config_session_factory.py:1006` (`test_handoff_retains_loaded_model_values_without_database_fallback`) builds a config whose `db_factory` raises `AssertionError("db fallback")` if it is ever called, then performs the handoff and invokes **all fourteen getters** (7 value + 6 mapping, across both `True` and `False` load-plan flags). None trip the assertion — direct proof that no getter touches `self.db` post-handoff.
- `tests/web/api/test_tools_api.py:210` (`test_available_tools_cleanup_failure_follows_complete_route_body`) and `tests/web/api/test_tools_api.py:253` (`test_available_tools_primary_failure_wins_over_terminal_cleanup`) confirm `close()` fires exactly once and as the last event, and that a primary body failure still wins over a cleanup failure.
- Combined run of `tests/web/api/test_tools_api.py`, `tests/web/tools/test_web_tool_config_session_factory.py`, and `tests/core/tools/adapters/vibe/test_factory_sandbox_release.py`: **112 passed, 0 failed.**

I consider F1 closed with confidence.

#### F3 — cleanup exception shadowing the primary exception — FIXED

The bare `try/finally` previously at `src/xagent/core/tools/adapters/vibe/agent_tool.py:2020-2089` is gone, replaced by the same shared boundary:

```python
result, agent_service = await run_with_tool_runtime_cleanup(
    execute_delegated_runtime, tool_config.close, logger=logger, cleanup_error_message=...
)
```

The helper (`src/xagent/core/tools/adapters/vibe/config.py:38-63`) implements exactly the two-case contract that was asked for:

- **Body raised, cleanup also raised** → the cleanup failure is logged with `exc_info=True` and the *original* exception propagates unchanged, by identity. Asserted directly: `caught.value is primary` in `test_shared_cleanup_preserves_primary_base_exception_identity` (`tests/core/tools/adapters/vibe/test_factory_sandbox_release.py:150-171`).
- **Body succeeded, cleanup raised** → raises `ToolFactoryRuntimeSessionBoundaryError() from cleanup_error`, so the cleanup fault is still surfaced as a stable typed error and is properly chained. Asserted directly: `caught.value.__cause__ is cleanup_fault` in `test_shared_cleanup_surfaces_a_stable_boundary_error_after_success` (same file, lines 112-126).

`tests/core/tools/adapters/vibe/test_factory_sandbox_release.py`: **15 passed**, including both of the above regressions. Consolidating this into one shared helper used by both the runtime path and the listing route is also the right call — it means there is a single place where this invariant lives.

F3 closed with confidence.

---

### Remaining open items — author's stated positions

- **A (Major — lost graceful degradation when the DB is not initialized).** Author's position: production initializes the database before registration, and core-only mode omits that provider entirely; broadly swallowing provider failures would risk hiding a broken skill layer or a real authorization fault. The code is confirmed unchanged — the provider still raises unconditionally, which still fails the whole skill list rather than degrading. I accept the reasoning about current entry points and am **not blocking on it**, but I'd still like this picked up later: the failure mode (one uninitialized provider taking out the entire skill listing) is worse than the cost of a scoped, provider-local degradation that logs loudly — a narrow guard is not the same as "broadly swallowing provider failures." Suggest a small follow-up issue.
- **B (team_id removal) — WAIVED, by design.** Author's stated position: team scope was addressed by removing request-derived tenant metadata from both paths; team authorization remains downstream-owned, and SaaS providers must re-resolve current membership from `user_id`. That directly answers the round-2 question; the metadata removal is intentional and the contract for overlay/SaaS providers is now explicit. No further action.
- **C (`SkillScopeContext` not deepcopy/pickle-able) — WAIVED.** Author's position: it is an in-process boundary object with no current checkpoint or transport consumer, so serialization support is intentionally not adopted. Reasonable given no consumer exists.
- **D (double session-handoff in `skills.py` / `skill_hub.py`) — WAIVED.** Author's position: collapsing dependency-time and pre-manager handoffs is intentionally not adopted — the first releases the authenticated dependency transaction, the second guards later route work from re-acquiring it. Distinct windows; accepted.
- **E (`Mapping` import source) — WAIVED.** Author's position: this is an import-only style choice, and the repository has no single enforced import source for that type. Accepted as a non-issue.

Not re-litigating B–E.

---

### Simplification opportunities (optional, non-blocking)

- `src/xagent/web/tools/config.py:313-347`: `_RetainedFactoryModelState.__post_init__` wraps every mapping field in `MappingProxyType`, but `_get_factory_model_mapping` (`src/xagent/web/tools/config.py:1355-1378`) already returns a fresh `dict(...)` copy on every read, so no caller ever holds a live reference to the frozen mapping. Suggest dropping the proxy wrapping (keep the plain `dict()` copy in `from_factory_snapshot`) and removing the now-unused `types.MappingProxyType` import.
- `src/xagent/web/tools/config.py:1355-1378`: `_get_factory_model_mapping` duplicates `_get_factory_model_value` line-for-line except that each return is wrapped in `dict(...)`. Suggest a one-line delegation: `return dict(self._get_factory_model_value(load_flag=load_flag, field_name=field_name, cache_name=cache_name, loader=loader) or {})`.

On the object-graph reachability walker in `tests/web/tools/test_web_tool_config_session_factory.py:113-138` (`_graph_reaches_identity` / `_assert_identities_not_reachable`, plus the positive-control self-test at lines 1109-1123): this could be simpler as direct `assert retained.<field> is not <forbidden>` checks against the five known fields, rather than a generic dataclass/dict/mappingproxy/list/tuple/set/frozenset walker with a cycle guard. Non-blocking.

One forward-looking aside, not a finding and not actionable here: factory-runtime lifecycle is now tracked across three separately-toggled fields (`_factory_runtime_snapshot`, `_retained_factory_model_state`, `_factory_runtime_handed_off`). Each carries genuinely distinct and exercised information today — e.g. `abort_factory_runtime` preserves a prior retained generation while still setting `handed_off=True` — so this is fine as-is; just worth keeping an eye on if a fourth state flag ever shows up. The remainder of the +330 lines in `config.py` (the getter-dispatch helpers replacing ~12 duplicated getter bodies, and `_RetainedFactoryModelState` itself) is necessary complexity backing the "no DB reads after handoff" contract, and is covered by tests.

---

### Additional observations (non-blocking)

A deeper design pass over the `WebToolConfig` connection-lifecycle changes surfaced two more points, both verified against the current code and tests:

- **`close()` permanently disables all 14 model getters after it's called, undocumented.** `src/xagent/web/tools/config.py:2137-2149`'s `finally` unconditionally sets `_retained_factory_model_state = None` and `_factory_runtime_handed_off = True`, but the docstring still only says "Close the lazily-opened factory session, if any." After `close()`, every media-model getter short-circuits to `None`/`{}` — even ignoring a populated `_cached_*` value — with no log line. Not reachable today (both call sites invoke `close()` only in a `finally` after the body that reads the getters has already run), but it's a landmine for any future caller that reorders these calls. Worth a docstring update and maybe a debug log; not worth blocking on.
- **A successful run whose `close()` then fails discards the completed result.** This is a real, deliberate trade-off of the F1/F3 fix, not an oversight: `run_with_tool_runtime_cleanup` (`src/xagent/core/tools/adapters/vibe/config.py:38-63`) raises when cleanup fails even if the body succeeded, so a fully-computed `/api/tools/available` response becomes a 500, and a successful sub-agent run in `agent_tool.py` gets converted to an error string instead of returning its real output. Both behaviors are directly locked in by tests (`tests/core/tools/test_agent_tool_session_isolation.py:99-111`, `tests/web/api/test_tools_api.py:210-227`), so this is intentional. It's also gated behind a narrow condition — both `Session.close()` and `Session.invalidate()` failing together, which realistically only happens when the underlying DB connection is already broken well below the ORM. Fail-loud here is defensible given this PR exists specifically to stop broken/leaked connections from re-entering the pool. One refinement worth raising, not a defect: in `agent_tool.py`, the sub-agent's already-completed output could still be preserved and returned alongside a logged cleanup warning, rather than discarded outright — but this is a suggestion for a future pass, not something to hold up this PR.

---

### Recommendation: Approve

Both blocking findings from round 2 are fixed with strong, targeted test evidence — F1 with a `db_factory`-that-explodes test covering all fourteen getters plus route-level cleanup-ordering tests, F3 with two regression tests asserting exception identity and `__cause__` chaining. 112 tests pass across the three relevant suites.

The one item I'd still like addressed (Major — degradation when the DB is not initialized) is low-risk and not reachable in production today, since every real entry point calls `init_db()` at startup. It is cheap to fix later and does not justify a fourth round on a PR that has already been through multiple rounds fixing the actual pool-exhaustion problem it exists to solve. Suggest filing a lightweight follow-up issue for it. The two simplification suggestions are optional cleanup and are not blockers.

AlexLiu190625:
Implemented the remaining actionable feedback in `68a2a634` through `346f7fa7`:

- The personal skill provider now treats only an uninitialized database capability as unavailable, logs that condition, and leaves filesystem/overlay skills available. An installed factory's session, query, materialization, and cleanup failures still propagate unchanged; the composite provider does not return partial results.
- Media-model value and mapping getters now share one resolution owner while preserving snapshot → retained generation → handoff neutral → cache/loader order. Mapping getters avoid truthiness normalization, keep invalid pre-handoff `None` fail-loud, and continue returning fresh dictionaries.
- `WebToolConfig.close()` now documents its actual generation-level contract: it finalizes the current prepared/retained factory generation and owned lazy session; a later public prepare may install a new generation, while an explicitly supplied vision model remains independent.

The optional suggestions to remove `MappingProxyType` or replace the transitive reachability check were not adopted because they would weaken internal immutability and resource-detachment evidence. Cleanup remains fail-loud after a successful body because returning success with a cleanup fault would require a separate partial-success/idempotency contract.

Verification: all focused skill, factory-runtime, AgentTool, tools API, AgentService refresh, skill-runtime, identity, and security suites passed; pre-commit (including mypy) passed; and the real Docker/PostgreSQL pool-handoff E2E passed 3/3.

PR #1048, opened 2026-07-28
feat(connectors): expose selected visible connector resolver

AlexLiu190625:
## Summary

- expose one shared resolver for the MCP and Custom API connectors that are both visible to a user and selected by an Agent
- keep runtime-declaration projection in one helper shared by task snapshot and create-plan paths while allowing consumers to resolve selected connectors without that restriction
- preserve scoped-name selection, owner/team visibility, normalization, and deterministic `(connector_type, connector_id)` ordering
- document the public resolver contract and cover persisted `None`/`[]` selection sentinels at the ORM boundary

## Verification

- `uv run pytest -q tests/web/services/test_connector_runtime_snapshot.py` — 26 passed
- `uv run pytest -q tests/web/services/test_connector_runtime_snapshot.py tests/core/tools/adapters/vibe/test_selection_spec.py` — 127 passed
- `uv run mypy src/xagent/web/services/connector_runtime.py` — passed
- `uv run pre-commit run --files src/xagent/web/services/connector_runtime.py tests/web/services/test_connector_runtime_snapshot.py` — passed
- `git diff --check` — passed

### DIFF
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -607,6 +607,10 @@ def resolve_agent_selected_connectors(
 ) -> dict[ConnectorRef, Any]:
     """Return visible connectors selected by ``agent`` for the supplied owner.
 
+    Selection derives only from ``Agent.tool_categories``; task-level runtime
+    overlays are not applied. Callers that need the effective per-task tool set
+    must apply those policies at the task-selection boundary.
+
     Runtime-declaration filtering is intentionally left to consumers. Results
     iterate deterministically by ``(connector_type, connector_id)``.
     """
@@ -640,6 +644,8 @@ def _runtime_declared_refs(
 def _select_agent_visible_connectors(
     agent: Agent, visible: dict[ConnectorRef, Any]
 ) -> dict[ConnectorRef, Any]:
+    """Select visible connectors and establish canonical ref iteration order."""
+
     tool_categories = (
         list(agent.tool_categories) if isinstance(agent.tool_categories, list) else None
     )
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -634,11 +634,10 @@ def _select_agent_visible_connectors(
     )
     spec = ToolSelectionSpec.from_raw(tool_categories=tool_categories)
     selected: dict[ConnectorRef, Any] = {}
-    for ref in _sort_connector_refs(visible):
-        connector = visible[ref]
+    for ref, connector in visible.items():
         if _is_agent_selected_connector(spec, ref, connector):
             selected[ref] = connector
-    return selected
+    return {ref: selected[ref] for ref in _sort_connector_refs(selected)}
 
 
 def _is_agent_selected_connector(
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -645,7 +645,7 @@ def _is_agent_selected_connector(
     spec: ToolSelectionSpec, ref: ConnectorRef, connector: Any
 ) -> bool:
     scoped_mcp_servers = spec.scoped_mcp_servers()
-    name_key = normalize_mcp_server_name(str(connector.name or ""))
+    name_key = normalize_mcp_server_name(connector.name)
     if ref.connector_type == CONNECTOR_TYPE_MCP:
         if not spec.includes_mcp():
             return False
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -644,17 +644,20 @@ def _select_agent_visible_connectors(
 def _is_agent_selected_connector(
     spec: ToolSelectionSpec, ref: ConnectorRef, connector: Any
 ) -> bool:
-    scoped_mcp_servers = spec.scoped_mcp_servers()
-    name_key = normalize_mcp_server_name(connector.name)
     if ref.connector_type == CONNECTOR_TYPE_MCP:
         if not spec.includes_mcp():
             return False
-        return scoped_mcp_servers is None or name_key in scoped_mcp_servers
-    if ref.connector_type == CONNECTOR_TYPE_CUSTOM_API:
+    elif ref.connector_type == CONNECTOR_TYPE_CUSTOM_API:
         if not spec.includes_custom_api():
             return False
-        return scoped_mcp_servers is None or name_key in scoped_mcp_servers
-    return False
+    else:
+        return False
+
+    scoped_mcp_servers = spec.scoped_mcp_servers()
+    if scoped_mcp_servers is None:
+        return True
+    name_key = normalize_mcp_server_name(connector.name)
+    return name_key in scoped_mcp_servers
 
 
 def _sort_connector_refs(refs: Iterable[ConnectorRef]) -> tuple[ConnectorRef, ...]:
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -391,8 +391,16 @@ def prepare_connector_runtime_selection_snapshot(
 
     if agent is None or connector_user_id is None:
         return ()
-    visible = _load_visible_runtime_connectors(db, user_id=int(connector_user_id))
-    return _plan_selected_refs(agent, visible)
+    selected = resolve_agent_selected_connectors(
+        db=db,
+        agent=agent,
+        connector_user_id=int(connector_user_id),
+    )
+    return _sort_connector_refs(
+        ref
+        for ref, connector in selected.items()
+        if _has_runtime_declaration(connector)
+    )
 
 
 def bind_connector_runtime_selection_snapshot(
@@ -598,31 +606,55 @@ def _load_visible_runtime_connectors(
     return visible
 
 
+def resolve_agent_selected_connectors(
+    *, db: Session, agent: Agent, connector_user_id: int
+) -> dict[ConnectorRef, Any]:
+    """Return visible connectors selected by ``agent`` for the supplied owner."""
+
+    visible = _load_visible_runtime_connectors(db, user_id=int(connector_user_id))
+    return _select_agent_visible_connectors(agent, visible)
+
+
 def _plan_selected_refs(
     agent: Agent, visible: dict[ConnectorRef, Any]
 ) -> tuple[ConnectorRef, ...]:
+    selected = _select_agent_visible_connectors(agent, visible)
+    return _sort_connector_refs(
+        ref
+        for ref, connector in selected.items()
+        if _has_runtime_declaration(connector)
+    )
+
+
+def _select_agent_visible_connectors(
+    agent: Agent, visible: dict[ConnectorRef, Any]
+) -> dict[ConnectorRef, Any]:
     tool_categories = (
         list(agent.tool_categories) if isinstance(agent.tool_categories, list) else None
     )
     spec = ToolSelectionSpec.from_raw(tool_categories=tool_categories)
-    selected: list[ConnectorRef] = []
+    selected: dict[ConnectorRef, Any] = {}
+    for ref in _sort_connector_refs(visible):
+        connector = visible[ref]
+        if _is_agent_selected_connector(spec, ref, connector):
+            selected[ref] = connector
+    return selected
+
+
+def _is_agent_selected_connector(
+    spec: ToolSelectionSpec, ref: ConnectorRef, connector: Any
+) -> bool:
     scoped_mcp_servers = spec.scoped_mcp_servers()
-    for ref, connector in visible.items():
-        if not _has_runtime_declaration(connector):
-            continue
-        name_key = normalize_mcp_server_name(str(connector.name or ""))
-        if ref.connector_type == CONNECTOR_TYPE_MCP:
-            if not spec.includes_mcp():
-                continue
-            if scoped_mcp_servers is not None and name_key not in scoped_mcp_servers:
-                continue
-        elif ref.connector_type == CONNECTOR_TYPE_CUSTOM_API:
-            if not spec.includes_custom_api():
-                continue
-            if scoped_mcp_servers is not None and name_key not in scoped_mcp_servers:
-                continue
-        selected.append(ref)
-    return _sort_connector_refs(selected)
+    name_key = normalize_mcp_server_name(str(connector.name or ""))
+    if ref.connector_type == CONNECTOR_TYPE_MCP:
+        if not spec.includes_mcp():
+            return False
+        return scoped_mcp_servers is None or name_key in scoped_mcp_servers
+    if ref.connector_type == CONNECTOR_TYPE_CUSTOM_API:
+        if not spec.includes_custom_api():
+            return False
+        return scoped_mcp_servers is None or name_key in scoped_mcp_servers
+    return False
 
 
 def _sort_connector_refs(refs: Iterable[ConnectorRef]) -> tuple[ConnectorRef, ...]:
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -366,7 +366,7 @@ def prepare_create_connector_runtime(
     return ConnectorRuntimeCreatePlan(
         task_source=task_source,
         connector_user_id=int(connector_user_id),
-        selected_refs=_sort_connector_refs(selected_refs),
+        selected_refs=selected_refs,
         context_by_ref=context_by_ref,
         ephemeral_by_ref=ephemeral_by_ref,
     )
@@ -396,11 +396,7 @@ def prepare_connector_runtime_selection_snapshot(
         agent=agent,
         connector_user_id=int(connector_user_id),
     )
-    return _sort_connector_refs(
-        ref
-        for ref, connector in selected.items()
-        if _has_runtime_declaration(connector)
-    )
+    return _runtime_declared_refs(selected)
 
 
 def bind_connector_runtime_selection_snapshot(
@@ -609,7 +605,11 @@ def _load_visible_runtime_connectors(
 def resolve_agent_selected_connectors(
     *, db: Session, agent: Agent, connector_user_id: int
 ) -> dict[ConnectorRef, Any]:
-    """Return visible connectors selected by ``agent`` for the supplied owner."""
+    """Return visible connectors selected by ``agent`` for the supplied owner.
+
+    Runtime-declaration filtering is intentionally left to consumers. Results
+    iterate deterministically by ``(connector_type, connector_id)``.
+    """
 
     visible = _load_visible_runtime_connectors(db, user_id=int(connector_user_id))
     return _select_agent_visible_connectors(agent, visible)
@@ -619,7 +619,18 @@ def _plan_selected_refs(
     agent: Agent, visible: dict[ConnectorRef, Any]
 ) -> tuple[ConnectorRef, ...]:
     selected = _select_agent_visible_connectors(agent, visible)
-    return _sort_connector_refs(
+    return _runtime_declared_refs(selected)
+
+
+def _runtime_declared_refs(
+    selected: dict[ConnectorRef, Any],
+) -> tuple[ConnectorRef, ...]:
+    """Project declared refs while preserving canonical selection order.
+
+    Callers provide the mapping ordered by the connector selection owner.
+    """
+
+    return tuple(
         ref
         for ref, connector in selected.items()
         if _has_runtime_declaration(connector)
--- a/tests/web/services/test_connector_runtime_snapshot.py
+++ b/tests/web/services/test_connector_runtime_snapshot.py
@@ -469,6 +469,7 @@ def test_create_plan_preserves_canonical_cross_type_connector_order(
     db_session.flush()
     selected_mcp = _create_runtime_mcp(db_session, user, "Records")
     selected_api = _create_runtime_custom_api(db_session, user, "records")
+    unselected_mcp = _create_runtime_mcp(db_session, user, "Billing")
 
     plan = prepare_create_connector_runtime(
         db=db_session,
@@ -482,6 +483,7 @@ def test_create_plan_preserves_canonical_cross_type_connector_order(
         ConnectorRef("custom_api", int(selected_api.id)),
         ConnectorRef("mcp", int(selected_mcp.id)),
     )
+    assert ConnectorRef("mcp", int(unselected_mcp.id)) not in plan.selected_refs
 
 
 def test_selection_snapshot_uses_public_resolver_and_filters_runtime_declarations(
--- a/tests/web/services/test_connector_runtime_snapshot.py
+++ b/tests/web/services/test_connector_runtime_snapshot.py
@@ -7,12 +7,16 @@ from sqlalchemy import create_engine
 from sqlalchemy.orm import Session, sessionmaker
 from sqlalchemy.pool import StaticPool
 
-from xagent.core.tools.adapters.vibe.connector_runtime import ConnectorRuntimeError
+from xagent.core.tools.adapters.vibe.connector_runtime import (
+    ConnectorRef,
+    ConnectorRuntimeError,
+)
 from xagent.web.models import Agent, Base, MCPServer, Task, User, UserMCPServer
 from xagent.web.models.agent import AgentStatus
 from xagent.web.models.custom_api import CustomApi, UserCustomApi
 from xagent.web.models.task import TaskStatus
 from xagent.web.services import connector_runtime as connector_runtime_service
+from xagent.web.services import connector_team_scope
 from xagent.web.services.connector_runtime import (
     ConnectorRuntimeValues,
     bind_connector_runtime_selection_snapshot,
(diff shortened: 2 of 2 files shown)


src/xagent/web/services/connector_runtime.py
```diff
@@ -598,31 +606,55 @@ def _load_visible_runtime_connectors(
...
-            if scoped_mcp_servers is not None and name_key not in scoped_mcp_servers:
-                continue
-        elif ref.connector_type == CONNECTOR_TYPE_CUSTOM_API:
-            if not spec.includes_custom_api():
-                continue
-            if scoped_mcp_servers is not None and name_key not in scoped_mcp_servers:
-                continue
-        selected.append(ref)
-    return _sort_connector_refs(selected)
+    name_key = normalize_mcp_server_name(str(connector.name or ""))
```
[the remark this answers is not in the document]
AlexLiu190625: Updated in 8bc9a98c. The selection helper now passes `connector.name` directly to `normalize_mcp_server_name`, relying on the non-nullable `MCPServer.name` and `CustomApi.name` schema contract. The connector-runtime snapshot suite (22 tests), Ruff, formatting, mypy, and pre-commit all pass.

src/xagent/web/services/connector_runtime.py
```diff
@@ -598,31 +606,55 @@ def _load_visible_runtime_connectors(
...
+    name_key = normalize_mcp_server_name(connector.name)
+    if ref.connector_type == CONNECTOR_TYPE_MCP:
+        if not spec.includes_mcp():
+            return False
+        return scoped_mcp_servers is None or name_key in scoped_mcp_servers
+    if ref.connector_type == CONNECTOR_TYPE_CUSTOM_API:
+        if not spec.includes_custom_api():
+            return False
+        return scoped_mcp_servers is None or name_key in scoped_mcp_servers
+    return False
```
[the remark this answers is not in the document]
AlexLiu190625: Implemented in f21ca611. Connector type and dispatch checks now short-circuit before resolving the optional server scope, and unscoped selections return without normalizing connector names. A public-resolver regression test covers both MCP and Custom API connectors and verifies that unscoped selection skips normalization. The connector snapshot and selection-spec suites, Ruff, formatting, mypy, and pre-commit all pass.

src/xagent/web/services/connector_runtime.py
```diff
@@ -598,31 +606,58 @@ def _load_visible_runtime_connectors(
...
         list(agent.tool_categories) if isinstance(agent.tool_categories, list) else None
     )
     spec = ToolSelectionSpec.from_raw(tool_categories=tool_categories)
-    selected: list[ConnectorRef] = []
+    selected: dict[ConnectorRef, Any] = {}
+    for ref in _sort_connector_refs(visible):
+        connector = visible[ref]
+        if _is_agent_selected_connector(spec, ref, connector):
+            selected[ref] = connector
+    return selected
```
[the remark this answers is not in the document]
AlexLiu190625: Implemented in 463b5cc9. `_select_agent_visible_connectors` now filters visible MCP and Custom API candidates first, then rebuilds the result in `_sort_connector_refs` order. This preserves deterministic output while sorting only the selected subset. A resolver-level regression test verifies that unselected refs do not reach the sort boundary and that the returned mapping remains ordered. The connector snapshot suite (24 tests), selection-spec suite (101 tests), Ruff, formatting, mypy, and pre-commit pass.

src/xagent/web/services/connector_runtime.py
```diff
@@ -391,8 +391,16 @@ def prepare_connector_runtime_selection_snapshot(
...
     if agent is None or connector_user_id is None:
         return ()
-    visible = _load_visible_runtime_connectors(db, user_id=int(connector_user_id))
-    return _plan_selected_refs(agent, visible)
+    selected = resolve_agent_selected_connectors(
+        db=db,
+        agent=agent,
+        connector_user_id=int(connector_user_id),
+    )
+    return _sort_connector_refs(
```
rogercloud: This filter-and-sort block is now byte-identical to the one still inside `_plan_selected_refs` at L622-626. Before this PR it lived in exactly one place, which both entry points shared; extracting the resolver duplicated it. Suggest a small shared helper — `_runtime_declared_refs(selected) -> tuple[ConnectorRef, ...]` — called from both sites.

Separately (pre-existing, not introduced here): `_select_agent_visible_connectors` already returns its dict in `_sort_connector_refs` order at L640, and filtering preserves order, so this outer sort is a no-op. If the deterministic-ordering contract is documented on `resolve_agent_selected_connectors`, the helper can be a plain generator with no sort at all. Note this is a different issue from the sort fix in 463b5cc9, which correctly addressed the sort *inside* `_select_agent_visible_connectors`.

src/xagent/web/services/connector_runtime.py
```diff
@@ -598,31 +606,57 @@ def _load_visible_runtime_connectors(
     return visible
 
 
+def resolve_agent_selected_connectors(
+    *, db: Session, agent: Agent, connector_user_id: int
+) -> dict[ConnectorRef, Any]:
+    """Return visible connectors selected by ``agent`` for the supplied owner."""
```
rogercloud: This docstring doesn't state either of the two properties callers actually depend on: (a) the result is deliberately **not** filtered by `_has_runtime_declaration` — the whole reason this function exists separately from `_plan_selected_refs`, and the easiest thing for a future caller to get wrong; and (b) iteration order is deterministic (`_sort_connector_refs` order, by `(connector_type, connector_id)`). Worth spelling out both on a newly public function — (b) in particular is what would let the call sites stop re-sorting.

tests/web/services/test_connector_runtime_snapshot.py
```diff
@@ -254,6 +258,297 @@ def test_selection_snapshot_scopes_custom_api_by_source_server_name(
     } not in (task.connector_runtime_selected_refs or [])
 
 
+def test_selected_connector_resolver_returns_selected_mcp_without_runtime_declaration(
```
rogercloud, line 261: All seven new tests — and in fact every test in this file — construct the agent with a non-empty `tool_categories`, so only the `_SpecByCategories` branch of `ToolSelectionSpec` reaches `_is_agent_selected_connector`. Neither `tool_categories=None` (`_SpecAll`) nor `[]` / explicit-none (`_SpecNone`) is exercised through the newly public `resolve_agent_selected_connectors`, and there's no indirect coverage elsewhere. The logic is equivalent to the pre-refactor path for those cases, so this isn't a suspected bug — but since this function is now public API, two short cases here would lock in the "select everything" and "select nothing" behaviors cheaply.

tests/web/services/test_connector_runtime_snapshot.py
```diff
@@ -254,6 +258,297 @@ def test_selection_snapshot_scopes_custom_api_by_source_server_name(
...
+        connector_user_id=int(user.id),
+    )
+
+    assert list(resolved) == [
+        ConnectorRef("custom_api", int(custom_api.id)),
+        ConnectorRef("mcp", int(mcp_server.id)),
+    ]
+
+
+def test_selected_connector_resolver_filters_before_sorting_selected_refs(
```
rogercloud: This test monkeypatches the private `_sort_connector_refs` to assert on its call count and argument set, which pins an implementation detail rather than observable behavior. Its one behavioral assertion (`list(resolved) == sorted(selected_refs)`) is already covered by `test_selected_connector_resolver_normalizes_names_and_orders_refs` above. Suggest dropping this test, or folding any assertion worth keeping into that one — it would also make the sort-related cleanup in `prepare_connector_runtime_selection_snapshot` / `_plan_selected_refs` possible without a test rewrite.

tests/web/services/test_connector_runtime_snapshot.py
```diff
@@ -254,6 +258,297 @@ def test_selection_snapshot_scopes_custom_api_by_source_server_name(
...
+        return {
+            ConnectorRef("mcp", int(declared.id)): declared,
+            ConnectorRef("mcp", int(undeclared.id)): undeclared,
+        }
+
+    monkeypatch.setattr(
+        connector_runtime_service,
+        "resolve_agent_selected_connectors",
+        resolve_selected,
+        raising=False,
```
rogercloud: `raising=False` isn't needed here — `resolve_agent_selected_connectors` exists on the module at patch time. Worse, it defeats what this test is for: if the resolver is ever renamed, the patch silently becomes a no-op and the test would keep passing while no longer verifying that the snapshot path routes through the public resolver. Please drop the flag so a rename fails loudly. The test itself is a good contract test; only this argument needs to go.

rogercloud (commented):
## PR summary

`prepare_connector_runtime_selection_snapshot` needs the set of connectors that an agent has selected *and* that are visible to a given user. Before this PR that logic existed only inside the private `_plan_selected_refs`, where it was entangled with an unrelated `_has_runtime_declaration` filter, so no other caller could reuse it. This PR extracts the selection predicate into a new public `resolve_agent_selected_connectors` and pushes the runtime-declaration filter up into the callers that actually need it. Selection semantics — scoped MCP server names, owner/team visibility, name normalization, and deterministic ordering — are preserved; two files change (`src/xagent/web/services/connector_runtime.py`, `tests/web/services/test_connector_runtime_snapshot.py`).

## Update summary

This PR had already been through three rounds of automated review by `gemini-code-assist[bot]` plus author fixes, all landing as commits inside this same PR and all before this re-review began. All three of gemini's inline findings are confirmed fixed at HEAD (see the checklist below). There are no commits after the last-reviewed commit — `463b5cc9` is both the last commit reviewed and current HEAD — so there is no separate "update-only" diff to describe; the findings below come from an independent fresh review of the full PR diff against base `a1bd6717`.

The refactor is a genuine decomposition rather than a behavior change. Tracing both entry points confirms the selection predicate produces the same result set as base for the two real connector types, with ordering and normalization intact. The extraction of `_is_agent_selected_connector` is a net readability win: the pre-PR nested conditionals repeated the `scoped_mcp_servers` check in both the MCP and custom-API branches, and that duplication is now gone.

Four design-level reservations, none blocking:

**D1 — the new public API has exactly one consumer.** `resolve_agent_selected_connectors` is called from exactly one production site, `prepare_connector_runtime_selection_snapshot` in the same module. The PR description's motivation ("allowing other callers to resolve selected connectors without that extra restriction") is forward-looking rather than realized in this diff. Landing a public seam ahead of its second consumer is a defensible call, but the contract is currently unverified by any second caller.

**D2 + D3 — redundant sort and a newly duplicated filter block.** One finding, two layers — and it is *not* the same thing as gemini's prior finding #3 below.

- *New in this PR (D2):* the block
  ```python
  _sort_connector_refs(
      ref for ref, connector in selected.items()
      if _has_runtime_declaration(connector)
  )
  ```
  now appears verbatim in two places — inlined into `prepare_connector_runtime_selection_snapshot` (`src/xagent/web/services/connector_runtime.py:399-403`) and still inside `_plan_selected_refs` (`:622-626`). Before this PR it existed in exactly one place, inside the old `_plan_selected_refs`, which both entry points called. A one-line helper (e.g. `_runtime_declared_refs(selected) -> tuple[ConnectorRef, ...]`) collapses both sites.
- *Pre-existing, not caused by this PR (D3):* both of those sites re-sort a dict that `_select_agent_visible_connectors` already returns in `_sort_connector_refs` order (`:640`), and filtering preserves order, so the second sort is a no-op. The same sort-then-sort-again shape existed in the pre-PR `_plan_selected_refs`. This PR neither introduced nor worsened the redundant sort; it's an optimization opportunity that survives the refactor.

**D4 — `dict[ConnectorRef, Any]` return type.** `resolve_agent_selected_connectors` (`:611`) returns values that are always `MCPServer | CustomApi` ORM rows, and `Any` forfeits mypy checking on downstream attribute access. Real, but this extends an existing module-wide `Any` idiom (20+ uses in the same file) rather than introducing a new anomaly. A `Protocol` or a union would be a nice-to-have on a public function; not a regression.

## Prior findings checklist (gemini-code-assist[bot])

| # | Finding | Status | Fix |
|---|---------|--------|-----|
| 1 | Missing `str(x or "")` guard before `normalize_mcp_server_name(connector.name)` | **FIXED** | `8bc9a98c` — the redundant guard was dropped, on the grounds that `MCPServer.name` / `CustomApi.name` are DB `nullable=False`. Verified sound: every connector reaching this code is DB-loaded via `_load_visible_runtime_connectors`, so no path can pass `name=None`. |
| 2 | Predicate did expensive work before cheap inclusion checks | **FIXED** | `f21ca611` — `_is_agent_selected_connector` now runs type/inclusion checks with early `return False`, calls `scoped_mcp_servers()` / `normalize_mcp_server_name()` only after they pass, and returns early when there is no server-level restriction. |
| 3 | `_select_agent_visible_connectors` sorted the full `visible` dict before filtering | **FIXED** | `463b5cc9` — now filters `visible` into `selected` first and sorts only `selected` once at `:640`. **Note:** this is distinct from D2/D3 above, which concerns a second, redundant sort at this function's two *call sites*, not inside it. |

## Line-level findings

Severities: nothing here is critical or major.

### Informational / no action

- **F1 — `src/xagent/web/services/connector_runtime.py:658`** — `normalize_mcp_server_name(connector.name)` has no `str(x or "")` guard. Confirmed safe: both `MCPServer.name` and `CustomApi.name` are DB non-nullable and every connector here is DB-loaded. Same concern as prior finding #1, deliberately resolved by removing the guard; recorded only to confirm the fix holds.
- **F2 — `:652-653`** — the `else: return False` for a `connector_type` outside `{mcp, custom_api}` is unreachable: `ConnectorRef.__post_init__` rejects other types at construction and `_load_visible_runtime_connectors` only builds those two. It is a semantic tightening vs. the old implicit-fallthrough-select behavior, but since unreachable this is purely informational.
- **F3 — `:655`** — `spec.scoped_mcp_servers()` is now called once per connector inside the predicate rather than once per selection pass. All implementations are pure dict/set lookups over pre-built data with no I/O, so not worth changing.

### Minor

- **F4 — `tests/web/services/test_connector_runtime_snapshot.py:408-456`** — `test_selected_connector_resolver_filters_before_sorting_selected_refs` monkeypatches the private `_sort_connector_refs` purely to assert on call count/argument, an implementation detail. Its only behavioral assertion duplicates `test_selected_connector_resolver_normalizes_names_and_orders_refs` (`:378`). Consider merging and dropping this one.
- **F5 — `tests/web/services/test_connector_runtime_snapshot.py:493`** — `monkeypatch.setattr(..., "resolve_agent_selected_connectors", ..., raising=False)`. The attribute already exists at patch time, so `raising=False` is unnecessary and actively harmful: if the function is ever renamed, the patch silently becomes a no-op instead of failing. Drop the flag; the test itself is valuable.
- **F6 — `tests/web/services/test_connector_runtime_snapshot.py`** — every test in the file exercises only the `_SpecByCategories` branch of `ToolSelectionSpec` (non-empty `tool_categories`). Neither `tool_categories=None` (`_SpecAll`) nor `[]`/explicit-none (`_SpecNone`) is exercised through the refactored path, with no indirect coverage elsewhere. Logic is provably equivalent to base for those cases, so this is a coverage gap rather than a suspected bug — but it's cheap to close on a newly public function.

### Trivial

- **F7 — `src/xagent/web/services/connector_runtime.py:612`** — the docstring omits both load-bearing contracts of this new public function: that results are **not** filtered by `_has_runtime_declaration` (the reason it exists separately from `_plan_selected_refs`), and that iteration order is deterministic.
- **F8 — PR description** — claims "22 passed" for the test file; actual count at HEAD is 24. Accurate as of `642d3016`, but two later commits added tests. Please refresh the description.

```
connector_runtime.py L399: delete duplicated filter+sort block, identical to L622-626. Extract `_runtime_declared_refs(selected)` and call it from both sites.
connector_runtime.py L399: shrink redundant re-sort of an already-sorted dict (`_select_agent_visible_connectors` sorts at L640; filtering preserves order). Drop the outer `_sort_connector_refs` and document the ordering contract (see F7). Pre-existing, same pattern at L622.
test_connector_runtime_snapshot.py L408: shrink `test_selected_connector_resolver_filters_before_sorting_selected_refs`, redundant with `test_selected_connector_resolver_normalizes_names_and_orders_refs` (L378) apart from a monkeypatched call-count assertion on a private helper.

net: -54 lines possible
```

One soft, unconfirmed aside: `test_selected_connector_resolver_skips_name_normalization_without_server_scope` (`:336`) overlaps substantially with `test_selected_connector_resolver_returns_selected_mcp_without_runtime_declaration` (`:261`), but it does cover a distinct scenario (a `custom_api` decoy alongside the mcp connector, plus the skip-normalization-when-unscoped behavior) — not a confirmed redundancy, only mentioning in case a clean merge is easy.

## Verification

- `.venv/bin/python -m pytest -q tests/web/services/test_connector_runtime_snapshot.py` → **24 passed** at HEAD `463b5cc9`.
- `ruff` / `mypy` reported passing by the author; not independently re-run in this review.
- Full PR diff reviewed against base `a1bd6717`; all four commits traced.

## Closing

No critical or major issues. This is a small, low-risk internal refactor with good test coverage and correct behavior preservation, and the three prior automated findings were each addressed properly. The design reservations — the single-consumer public seam (D1), the duplicated filter block and redundant sort (D2/D3), and the docstring gaps (F7) — are worth addressing, ideally in this PR since D2 is newly introduced here, but none of them blocks merge.

src/xagent/web/services/connector_runtime.py
AlexLiu190625: Fixed in 51cdbd2f. Both selection-snapshot and create-plan paths now use `_runtime_declared_refs`, which preserves the canonical order established by the selection owner. The caller-level sorts and the final create-plan re-sort were removed. Snapshot and create-plan tests cover declaration filtering and canonical ordering.

src/xagent/web/services/connector_runtime.py
AlexLiu190625: Updated in 51cdbd2f. The public resolver now states that runtime-declaration filtering is intentionally left to consumers and that results iterate deterministically by `(connector_type, connector_id)`.

tests/web/services/test_connector_runtime_snapshot.py
AlexLiu190625: Added in 51cdbd2f. The public resolver now has ORM round-trip cases for `tool_categories=None` and `tool_categories=[]`, covering select-all and select-none behavior with visible MCP and Custom API connectors.

tests/web/services/test_connector_runtime_snapshot.py
AlexLiu190625: Fixed in 51cdbd2f. The patch no longer uses `raising=False`, so the seam test fails immediately if `resolve_agent_selected_connectors` is absent or renamed.

tests/web/services/test_connector_runtime_snapshot.py
AlexLiu190625: Removed in 51cdbd2f. Canonical ordering is now tested through observable behavior across connector types and the same-type ID tie-break. The filter-before-sort complexity shape is intentionally not treated as a public contract.

AlexLiu190625:
Two design notes remain intentionally unchanged. The resolver stays public because an existing downstream embedding consumer needs the selected-visible mapping before runtime-declaration filtering. The ORM value type remains `Any` in this change: a sound replacement requires a dedicated row `Protocol` or SQLAlchemy `Mapped[...]` boundary, while a local union or cast would only hide the current ORM typing limitation. The existing fail-closed unknown-type branch and lightweight scope lookup are also retained.

src/xagent/web/services/connector_runtime.py
```diff
@@ -598,31 +602,72 @@ def _load_visible_runtime_connectors(
...
+    """
+
+    return tuple(
+        ref
+        for ref, connector in selected.items()
+        if _has_runtime_declaration(connector)
+    )
+
+
+def _select_agent_visible_connectors(
```
rogercloud, line 644: After this refactor, this is the only function in the module that performs the canonical sort — `_runtime_declared_refs` is now a pure order-preserving projection, and `resolve_agent_selected_connectors`'s docstring advertises the `(connector_type, connector_id)` ordering that originates right here. But neither of those docstrings names this function, and this one has no docstring at all, so the invariant has to be reconstructed by reading three functions. Worth a one-liner, e.g. "Establishes the canonical `(connector_type, connector_id)` order that all downstream projections rely on." Not blocking — I confirmed the invariant is test-enforced (removing the sort here fails 4 tests), so this is about discoverability for the next reader, not correctness.

src/xagent/web/services/connector_runtime.py
```diff
@@ -598,31 +602,72 @@ def _load_visible_runtime_connectors(
     return visible
 
 
+def resolve_agent_selected_connectors(
+    *, db: Session, agent: Agent, connector_user_id: int
+) -> dict[ConnectorRef, Any]:
+    """Return visible connectors selected by ``agent`` for the supplied owner.
```
rogercloud, line 608: Not introduced by this PR, and not something to fix here — but this docstring is the first place this contract gets written down publicly, so it's a good spot to record a caveat. The spec built below (around line 646) calls `ToolSelectionSpec.from_raw(tool_categories=...)` without the `extras_only_when_unconfigured` / `exclude_custom_api_when_unconfigured` flags that the real tool-loading path passes for workforce-manager tasks (`src/xagent/web/api/chat.py`, `_build_tool_selection_spec_for_task`). For a workforce-manager agent with unconfigured `tool_categories`, this therefore returns *all* visible connectors — a superset of what would actually be loaded for that turn. Confirmed the same omission exists in the pre-PR code at the base commit, and the impact is bounded (the result feeds payload validation, not tool instantiation/authorization). Suggest a one-line "may be a superset of the effective tool-loading spec for workforce-manager agents" note here plus a follow-up issue.

tests/web/services/test_connector_runtime_snapshot.py
```diff
@@ -254,6 +258,324 @@ def test_selection_snapshot_scopes_custom_api_by_source_server_name(
...
+        connector_user_id=int(user.id),
+    )
+
+    assert list(resolved) == [
+        ConnectorRef("custom_api", int(custom_api.id)),
+        ConnectorRef("mcp", int(mcp_server.id)),
+    ]
+
+
+def test_create_plan_preserves_canonical_cross_type_connector_order(
```
rogercloud, line 455: Good trade replacing the monkeypatch-based test with a real end-to-end one. One small piece of coverage went with it: the deleted test also created two connectors outside the tested category scope ("Billing" mcp, "Invoices" custom_api) and so implicitly proved they were excluded, whereas this one creates only the connectors that should be selected. That means "a visible, runtime-declared connector excluded purely by category/name scope doesn't leak into `plan.selected_refs`" is no longer asserted at the `prepare_create_connector_runtime` entry point — it is still covered at the snapshot entry point and directly against the resolver. Since all three share `_select_agent_visible_connectors`, the residual risk is wiring-level rather than logic-level, so this is optional: one extra out-of-scope fixture plus a `not in plan.selected_refs` assertion would close it, or leave it on the shared-logic argument.

rogercloud (approved):
## Follow-up review 2 — commit 51cdbd2f

### PR summary

This PR extracts the agent → connector selection logic in `src/xagent/web/services/connector_runtime.py` into a public resolver, `resolve_agent_selected_connectors`, so callers outside the runtime-plan path can ask "which visible connectors does this agent select?" without going through plan construction. The resolver returns the visible connectors an agent selects for a given owner, in a deterministic `(connector_type, connector_id)` order, and deliberately leaves runtime-declaration filtering to consumers. The existing snapshot and create-plan paths are rewired on top of it, and the module's ordering behaviour is pinned by tests. Scope is narrow: one production file and one test file.

### Update summary

Commit 51cdbd2f ("ref: centralize connector declaration projection") responds to all five actionable findings from the previous round (463b5cc9). Concretely it: extracts the duplicated filter-and-sort block into a shared `_runtime_declared_refs` helper called from both `prepare_connector_runtime_selection_snapshot` and `_plan_selected_refs`; drops the redundant no-op `_sort_connector_refs(selected_refs)` call in `prepare_create_connector_runtime`; expands the `resolve_agent_selected_connectors` docstring to state both contracts (no runtime-declaration filtering, deterministic ordering); adds a parametrized sentinel test covering `tool_categories=None` vs `tool_categories=[]` through the real resolver; and replaces the monkeypatch-based implementation-detail test with an end-to-end ordering test, also dropping the `raising=False` from the remaining monkeypatch call.

### Approach verdict: sound

Upgraded from *acceptable with reservations* in the first review. The upgrade is not because the comments were answered — it is because the update collapsed two of the original reservations ("the filter+sort logic is duplicated across two projections" and "the canonical ordering is an implicit, unstated invariant") into a single, well-scoped, test-enforced invariant. `_select_agent_visible_connectors` (`src/xagent/web/services/connector_runtime.py:640`) is now the only place in the module that sorts; `_runtime_declared_refs` is a pure order-preserving projection over its output.

I verified that invariant rather than assuming it: temporarily removing the `_sort_connector_refs` call at the end of `_select_agent_visible_connectors` and running the suite fails 4 tests. So the ordering guarantee the two new docstrings advertise is genuinely enforced by tests, not an incidental artifact of dict insertion order that would silently rot. That is what moves this from "reasonable but load-bearing on convention" to "sound".

The remaining first-round reservations about *where this resolver is consumed* are unchanged and still non-blocking (see D2).

### Prior findings — all 5 fixed

1. **FIXED** — duplicated filter+sort block. `_runtime_declared_refs` (`connector_runtime.py:625`) is now the single projection, called from both the snapshot path and `_plan_selected_refs`. Its correctness depends on receiving canonically ordered input, and that precondition is real and test-enforced (mutation check above).
2. **FIXED** — missing contract documentation. The `resolve_agent_selected_connectors` docstring (`connector_runtime.py:608-612`) now states both contracts explicitly: runtime-declaration filtering is left to consumers, and results iterate deterministically by `(connector_type, connector_id)`.
3. **FIXED** — no coverage for the select-all / select-none sentinels. `test_selected_connector_resolver_honors_all_and_none_sentinels` drives the real DB-backed resolver (no stub) and correctly distinguishes `tool_categories=None` (select all) from `tool_categories=[]` (select none), matching `ToolSelectionSpec.from_raw`'s actual branching.
4. **FIXED** — `raising=False` removed from the monkeypatch in `test_selection_snapshot_uses_public_resolver_and_filters_runtime_declarations`, so a renamed target now fails loudly instead of silently patching nothing.
5. **FIXED** — implementation-detail test replaced. `test_selected_connector_resolver_filters_before_sorting_selected_refs` (which monkeypatched a private function to assert call ordering) is deleted, and `test_create_plan_preserves_canonical_cross_type_connector_order` exercises the real `prepare_create_connector_runtime` end-to-end with no internal monkeypatching.

**Two things worth highlighting** — both show the fixes went past surface-level compliance:

- **The fixture creation-order swap in `test_selected_connector_resolver_includes_owner_and_team_visible_connectors_only` is a real test-quality fix, not churn.** With the previous fixture order, the expected `(connector_type, connector_id)`-sorted order happened to coincide exactly with plain dict insertion order — so the assertion passed whether or not the sort ran, i.e. it was a tautology. I confirmed this with an independent probe that varied fixture order while toggling the sort on and off. The new order makes the two diverge, so the test now actually exercises the sort. Easy to mistake for cosmetic reordering in review; it isn't.
- **The `db_session.expire(agent, ["tool_categories"])` call in the new sentinel test is a legitimate, non-obvious safeguard.** `Agent.tool_categories` carries an ORM-level `default=list`, which only fires when the field is never set. Expiring and re-reading after flush is what actually proves a `None` sentinel survived the persistence round-trip rather than being quietly coerced to `[]` — confirmed with a live SQLAlchemy check. Not defensive noise; without it the `None` case could pass for the wrong reason.

Two more points I checked and want to record as *not* problems:

- Keeping `prepare_create_connector_runtime` on `_plan_selected_refs` (which reuses the already-loaded `visible` mapping) instead of routing it through the public `resolve_agent_selected_connectors` (which would reload `visible` with a second query) is the right call. This is a deliberate avoidance of a redundant DB round-trip, not a missed consolidation.
- `_is_agent_selected_connector` now returns `False` explicitly for an unrecognised connector type (`connector_runtime.py:663-664`), where the old code fell through and effectively selected it. Unreachable given current type constraints, and untested either way, but it is a strictly safer default and worth noting.

### New findings this round

All low or nit severity. None blocking.

**D1 (low, new) — the sole ordering invariant site is undocumented.** `_select_agent_visible_connectors` (`src/xagent/web/services/connector_runtime.py:640`) is now the only function in the module that establishes the canonical order, and both `resolve_agent_selected_connectors`'s and `_runtime_declared_refs`'s docstrings describe or depend on that order — but neither names the function that produces it, and `_select_agent_visible_connectors` has no docstring at all. A one-liner there would make the invariant discoverable to the next person who touches this chain instead of something you reconstruct by reading three functions.

**D2 (low, carried forward, unchanged) — the public resolver still has no non-test production consumer.** Same as D1 in the first review. This commit was not expected to add one, so this is a status note rather than a new concern. Still fine as a prepared seam; no action in this PR.

**D3 (low, pre-existing — not introduced by this PR, informational).** The connector-selection spec built at `connector_runtime.py:646` calls `ToolSelectionSpec.from_raw(tool_categories=tool_categories)` without the `extras_only_when_unconfigured` / `exclude_custom_api_when_unconfigured` flags that the real tool-loading path passes for workforce-manager tasks (`src/xagent/web/api/chat.py`, `_build_tool_selection_spec_for_task`, around line 449). For a workforce-manager agent with unconfigured `tool_categories`, the connector-runtime snapshot's closed set therefore selects *all* visible connectors, i.e. a superset of what the real tool-loading path would select for the same agent. I confirmed this omission is identical in the pre-PR `_plan_selected_refs` at base commit `a1bd6717`, so this PR neither introduces nor worsens it. Practical impact is limited: the closed set is consumed as a validation gate for client-supplied runtime payloads (`_validate_payload_refs`), not as an authorization boundary for which tools get instantiated — so the superset means "a payload ref may validate even though it wouldn't correspond to a live tool that turn," not privilege escalation. Raising it here only because this PR's docstring is the first time this contract has been written down publicly, which makes it the natural place to record the caveat. Suggest a one-line docstring note and/or a follow-up issue; not for this PR to fix.

**D4 (nit) — double construction in `_select_agent_visible_connectors`.** The function accumulates `selected`, then rebuilds it as `{ref: selected[ref] for ref in _sort_connector_refs(selected)}` (`connector_runtime.py:647-651`). It could be built once. Pure style, take it or leave it. Explicitly *not* suggesting you sort `visible` before filtering instead — that would undo the "sort only the selected subset" optimization from an earlier commit in this same PR.

**6.1 (low, new) — the replacement plan test dropped one piece of coverage the deleted test had.** The deleted `test_selected_connector_resolver_filters_before_sorting_selected_refs` created two extra connectors outside the tested category scope ("Billing" mcp, "Invoices" custom_api) and so implicitly proved they were excluded. Its replacement, `test_create_plan_preserves_canonical_cross_type_connector_order` (`tests/web/services/test_connector_runtime_snapshot.py:455`), creates only the two connectors that *should* be selected. So "a visible, runtime-declared connector excluded purely by category/name scope does not leak into `plan.selected_refs`" is no longer asserted at the `prepare_create_connector_runtime` entry point — though equivalent scope-exclusion coverage does exist at the snapshot entry point and directly against the resolver. Since all three paths share the same `_select_agent_visible_connectors` filtering, the residual risk is wiring-level, not logic-level, which caps it. One extra fixture plus one assertion would close it cheaply; leaving it as-is is defensible on the shared-logic argument.

Lean already. I ran the simplification pass over the update-only diff (463b5cc9..51cdbd2f) and nothing survives it. `_runtime_declared_refs` is precisely the deduplication the previous round asked for — a five-line comprehension shared by two call sites, with no speculative parameters or extension points. The docstring additions state real, non-obvious contracts rather than restating signatures.

### Verification

- Full suite at HEAD: **26 passed**.
- `ruff check`, `ruff format --check`, and `mypy`: clean. Run independently, not taken from the PR description.
- Ordering invariant verified by mutation: removing the sort in `_select_agent_visible_connectors` fails 4 tests. The experiment was reverted and the worktree left pristine.

### Closing

No blocking issues. This round is a clean, well-targeted response to the previous review: the shared helper removes the duplication, the docstrings pin the contracts, and the test changes swap two weak assertions for real ones — including one that was silently tautological before. Approving; ready to merge.

D1, D3, D4, and 6.1 are all optional polish and none of them need to land before merge. D3 in particular is a pre-existing gap unrelated to this change and is better handled as a follow-up issue than as scope creep here.

src/xagent/web/services/connector_runtime.py
AlexLiu190625: Updated in 780d3138. `_select_agent_visible_connectors` now documents that it establishes the canonical `ConnectorRef` iteration order consumed by downstream selection projections.

tests/web/services/test_connector_runtime_snapshot.py
AlexLiu190625: Added in 780d3138. The create-plan test now includes a visible, runtime-declared MCP connector outside the selected name scope and asserts that it is absent from `plan.selected_refs`. The targeted test fails if create-plan bypasses the shared selection owner.

src/xagent/web/services/connector_runtime.py
AlexLiu190625: Documented in 780d3138. The public resolver now states that it derives selection from `Agent.tool_categories` only and does not apply task-level runtime overlays. Runtime behavior remains unchanged here: effective per-task selection requires the task-owned `ToolSelectionSpec` and cannot be inferred by this Agent-only API; the actual tool-loading path continues to enforce its task-level policy.

AlexLiu190625:
Closing the two non-inline suggestions:

- `resolve_agent_selected_connectors` remains public intentionally as an integration seam for callers that need the selected-visible mapping before runtime-declaration projection.
- The two-stage selected mapping is retained because it makes filter-before-sort explicit and keeps the canonical key in `_sort_connector_refs`; one-expression alternatives still materialize an intermediate sequence or duplicate the sort key without changing complexity.

Commit `780d3138` also documents the task-overlay boundary and adds create-plan exclusion coverage.

PR #893, opened 2026-07-15
fix(task): surface failed task reason live and add a coded-error seam

bluefish-08:
## Problem

Creating a task whose first run is refused before execution starts (e.g. an
app-layer quota gate) degraded badly in the live view, and only a page reload
showed the truth:

1. The chat showed a generic red "unknown error" instead of the refusal reason.
2. The sender's own message was invisible.
3. After a reload both rendered correctly, so the live view and the replayed
   view disagreed.

## Root causes

- **Reason dropped by the conversation panel.** The failed turn's reason rides
  in the terminal `task_completed` event and the chat context did add a failed
  assistant bubble — but without `isResult`, and the conversation panel only
  renders user / result / system-notice messages. The invisible bubble left an
  empty failed turn, which the panel rendered as a virtual "unknown error"
  placeholder.
- **No user_message event before tracing.** The sender's bubble normally comes
  from the `user_message` trace event, emitted once the agent run starts. A run
  refused at the run-start gate returns before tracing, so the event never
  fires. The new-task send path (unlike the existing-task path) never added an
  optimistic copy, so nothing rendered until a reload replayed the persisted
  transcript row.

## Changes

Frontend:
- Flag the terminal failure bubble as the turn's result (`isResult: true`) so
  the conversation panel renders it, and use the reason verbatim (no live-only
  "execution error" prefix) — a reload replays the persisted transcript row
  with the exact same text, so the two views now match.
- Add the optimistic sender bubble on the new-task and queued send paths once
  delivery is acknowledged, reusing the existing reconciliation machinery
  (turn id / content matching), so a run that dies before tracing still shows
  the sender's message. The existing-task path already did this.
- Coded-error seam: when a terminal event carries an `error_code`, the chat
  context emits an `xagent:task-error` CustomEvent with the code, structured
  details and fallback message. A no-op `TaskErrorController` is mounted as the
  reserved seam; an app layer can replace it to present product-specific UI
  (e.g. a plan-aware quota dialog). Stock xagent never sets a code, so behavior
  is unchanged there.

Backend:
- The run-start gate hook may now return a structured reason
  (`{code, metric, limit, plan, message}`) instead of a plain string. Core
  forwards it opaquely into the run result and the `task_completed` broadcast
  as `error_code` / `error_details`; the quota-hook seam stays quota-agnostic.

## Tests

- `task_completed` failure payloads: reason surfaces live, bubble is
  result-flagged and verbatim, coded failures emit the `xagent:task-error`
  event with code + details.
[...]

PR #1054, opened 2026-07-29
feat(widget): hide execution trace and add new-conversation control in guest chat

### LINKED ISSUE #1039: feat(widget): add a "new conversation" control to guest chat
## Description

A visitor of the embedded widget (and of a share link) cannot end the current
conversation. The task id is persisted in localStorage and restored on every
load, so every turn keeps accumulating in one task and its context grows without
bound. There is no way to start over short of clearing site data.

## Root cause

`src/components/widget/public-agent-chat-page.tsx:149-192` persists `taskId`
under `storageKey` and restores it at mount. The header (342-364) carries only
the agent name and status, and `TaskConversationPanel` is mounted with
`showTaskActions={false}` (line 396), so no reset control is rendered.

## Fix

Add a "New conversation" button to the widget header that:

- `safeRemoveItem(storageKey)`
- `setTaskId(null, { navigate: false })`
- clears `draftMessage`, `draftFiles`, `createTaskError`

The next message creates a fresh task through the existing `handleSend` path, so
no backend change is needed. Applies to both `authMode` values.

## Notes

#984 (XA-3) specifies a `new_conversation` message for the delegated-session
mode. That mode is dormant and does not cover guest chat; this is the same UI
affordance and should be reused when the session branch lands.

## Tests

Clicking the control drops the persisted task id and returns the view to the
start screen; the following message starts a new task.

### LINKED ISSUE #1041: feat(widget): hide thinking process and raw tool output in embedded chat
## Description

The embedded widget renders the model's reasoning and the raw JSON of every tool
call (arguments and output) above the answer. An end user on a customer site
should see the answer, not the execution trace.

## Root cause

`src/components/task/task-conversation-panel.tsx:697,721` hardcodes
`showProcessView={true}`. `ChatMessage.tsx:372` then mounts `TraceEventRenderer`,
which renders the LLM `reasoning` as markdown (`TraceEventRenderer.tsx:1242`) and
tool `args`/`output` as pretty-printed JSON (891, 903, 1048).

## Fix

Add `showProcessView?: boolean` (default `true`) to `TaskConversationPanelProps`
and pass `showProcessView={false}` from `public-agent-chat-page.tsx:394`.
Internal pages keep their current behaviour.

With the trace hidden, `ChatMessage.isProcessOnlyMessage` (line 328) turns
process-only turns into empty bubbles, so the panel must keep a single status
line ("Thinking…") until the answer streams in.

## Tests

A widget conversation renders no trace block and no raw tool JSON; a running
task still shows a status line; the internal task page is unchanged.

bluefish-08:
Closes #1041
Closes #1039

## Problem

The embedded widget renders the model's reasoning and the raw JSON of every tool
call (arguments and output) above the answer. An end user on a customer site
should see the answer, not the execution trace.

`task-conversation-panel.tsx:697,721` hardcoded `showProcessView={true}`, so
`ChatMessage.tsx:372` always mounted `TraceEventRenderer`.

## Change

- `TaskConversationPanelProps` gains `showProcessView?: boolean`, default `true`;
  `public-agent-chat-page.tsx` passes `authMode === "share"` — the widget hides
  the trace, share links keep it (#1041 is scoped to the embedded widget).
  Internal pages are unchanged.
- `ChatMessage.isProcessOnlyMessage` no longer keys off `shouldShowProcess` but
  off the trace events themselves. Without this, hiding the trace turns every
  past trace-only turn (`content: null`, `showEmptyStatus: false`) into a bubble
  holding nothing but the assistant avatar. The component now returns `null` for
  those turns — the timeline separates children with `space-y-*`, so a childless
  wrapper still contributes its gap.
- The status line falls back to a neutral `common.thinking` when the trace is
  hidden. `GeneratingIndicator`'s `latestTitle` is the running step's name
  ("LLM Call Start", "calling web_search"), i.e. the very trace detail the status
  line replaces. A terminal turn uses a dedicated `common.statusDone` so it can't
  sit on a frozen "Thinking".
- Failure text: the trace carries the backend's raw error string, so the mining
  loop is gated on `showProcessView` and the widget falls back to a generic
  `common.errors.taskFailed`.
- A past turn that stopped without an answer (`failed`, `paused`,
  `waiting_for_user`) keeps a visible mark when the trace is hidden — dropping
  both the trace and the bubble would leave the visitor's question followed by
  nothing at all. Known trade-off: in the common `ask_user` flow the historical
  waiting turn now shows a status line above the clarification message;
  redundant, but preferred over the blank-turn failure mode.

## New-conversation control (#1039)

The widget header gains a "New conversation" button, shown only while a task is
active. Clicking it drops the persisted task id (`safeRemoveItem(storageKey)`),
resets `taskId` (`setTaskId(null, { navigate: false })`), and clears the draft
message, draft files, and any task-create error. The visitor lands back on the
start screen and the next message creates a fresh task through the existing
`handleSend` path — no backend change. Applies to both `authMode` values; for
share links only the task pointer is dropped, the persisted guest auth survives
so no re-auth round-trip happens.

## Self-review round

- Ending a conversation mid-run now resets `isProcessing`/`currentTask`
  explicitly: nulling the taskId closes the socket, so no terminal WS event
  would ever clear them, leaving the start screen's composer disabled and the
  header stuck on "Connecting..." until a reload.
- Failed-turn `content` is redacted like mined trace errors when the trace is
  hidden: `final_answer_error` streams `str(exc)` into the answer message and
  the terminal handler stores the failure reason verbatim (#893), so the
  widget swaps both for the generic `common.errors.taskFailed`; internal pages
  and share links still show the original text. The panel flags failed result
  messages as failed for `ChatMessage` only when the trace is hidden.

## Follow-up review round

- The failed-turn redaction now also covers the copy button: `copyableContent`
  derives from `failedMessageText` on failed assistant turns, so copy hands out
  exactly what the bubble shows.
- A queued message whose target task is switched away before its socket
  connects (e.g. via "New conversation") now rejects immediately at the context
  level instead of sitting out the 30s timeout as a silent unhandled rejection.
- The mid-run reset also clears `isHistoryLoading` (its onConnect timer may
  never have been scheduled), so the header can't stick on "Initializing".
- Simplifications: factored the duplicated `item.status === "failed"` test in
  the panel's taskStatus wiring; dropped the dead `"failed"` entry from the
  dot-animation exclusion list.
- Known accepted trade-off (per review): "New conversation" does not stop the
  abandoned run server-side; the widget has no stop affordance and #1039 scopes
  the fix to the client.

## Review round

- Scoped `showProcessView` to the widget only; share links keep the trace
  (pinned by tests for both `authMode` values).
- Extended the keep-the-bubble exception from `failed` to every
  stopped-but-not-completed status via `isStoppedTraceProcessStatus`.
- `GeneratingIndicator` now reads the same normalized status as `statusTitle`;
  its `failed` branch was unreachable (the call site renders `failedMessageText`
  first) and has been removed along with the `errorMessage` prop.

## Tests

`ChatMessage.test.tsx` (15 cases), `public-agent-chat-page.test.tsx` pins
`data-show-process-view` per auth mode, `task-conversation-panel.test.tsx`
covers the prop default and forwarding. Every regression case was verified red
against the pre-fix logic before being fixed.

`vitest run src/components/chat src/components/widget src/components/task src/contexts/app-context-chat.test.tsx` — 138 passed.


frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -360,10 +372,28 @@ export function ChatMessage({
...
+    ? latestTitle
+    : resolvedProcessStatus === "completed"
+      ? t("common.done")
+      : t("common.thinking");
+
+  // An empty wrapper is not free: the timeline separates children with
+  // space-y-*, so a childless div still contributes its gap.
+  if (isProcessOnlyMessage && !copyableContent) {
+    return null;
+  }
```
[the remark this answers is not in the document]
bluefish-08: Applied in e4dbe2b6 — the suggested condition is the right one.

`!copyableContent` was expressing the wrong idea: once `isProcessOnlyMessage` holds, the bubble is not rendering regardless, so `rawContent` can only ever leave the action row floating on its own. Keying the bail-out on `!shouldShowProcess` says what actually matters: neither the trace nor the bubble will render, so there is nothing to wrap.

Worth noting the case is unreachable through the current callers — `task-conversation-panel.tsx` builds process items without `rawContent` (line 376-388) and message items without `traceEvents` (line 280), so `hasTraceEvents && rawContent` cannot co-occur. Took it anyway because the new condition is the clearer statement of intent, and added a regression test that pins it (`ChatMessage.test.tsx`: a trace-only turn carrying `rawContent` still renders nothing). Verified it fails against the old guard before applying the fix.

104 tests pass, lint and type-check clean.

rogercloud:
ci is failing

bluefish-08:
> ci is failing

it ok

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -360,10 +372,30 @@ export function ChatMessage({
...
+    ? latestTitle
+    : resolvedProcessStatus === "completed"
+      ? t("common.done")
+      : t("common.thinking");
+
+  // Neither the trace nor the bubble is going to render, so there is nothing
+  // left to wrap. Bail out rather than emit an empty div: the timeline
+  // separates children with space-y-*, so a childless wrapper still takes its
+  // gap, and a stray rawContent would leave the copy button floating alone.
+  if (isProcessOnlyMessage && !shouldShowProcess) {
```
rogercloud, line 432: **major** — the keep-the-bubble exception in `isProcessOnlyMessage` (lines 349-361) only excludes `resolvedProcessStatus === "failed"`: `(showProcessView || resolvedProcessStatus !== "failed")`. But a past trace-only turn's inferred status can also be `paused` or `waiting_for_user` (see `STOPPED_TASK_STATUSES` in `frontend/src/lib/task-status.ts`, and `isStoppedTraceProcessStatus` in `frontend/src/lib/trace-process-status.ts`, which already exports the right helper).

In widget mode (`showProcessView={false}`), such a turn hits the new `if (isProcessOnlyMessage && !shouldShowProcess) return null;` and disappears entirely — reproducing the exact "visitor's question followed by nothing" bug this PR fixes for the `failed` case, just for `paused`/`waiting_for_user`. This is reachable: a historical turn group closed by a later user message before it received a terminal (`completed`/`failed`) trace event lands on `inferredProcessStatus === "waiting_for_user"`. No test exercises `paused`/`waiting_for_user` combined with `showEmptyStatus={false}`.

Suggested fix — use the exported helper instead of the narrower check:

```ts
(showProcessView ||
  !(isStoppedTraceProcessStatus(resolvedProcessStatus) &&
    resolvedProcessStatus !== "completed"))
```

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -400,6 +400,9 @@ function PublicConversationContent({
             showTokenUsage={false}
             showDagPreview={false}
             showTaskFiles={false}
+            // A visitor on a customer's site gets the answer, not the run: no
+            // reasoning, no tool arguments, no raw tool output.
+            showProcessView={false}
```
rogercloud: **major** — `showProcessView={false}` is unconditional, but this component backs both `/widget/chat/[token]` (`authMode="widget"`) and `/share/[token]` (`authMode="share"`). So share links also lose the execution trace, which is outside the scope of #1041 and unmentioned in the PR body. `authMode` is already destructured here and used for branching elsewhere in this component (`publicApiPrefix`, `authPath`), so scoping this is a one-liner:

```tsx
showProcessView={authMode !== "share"}
```

If hiding the trace on share links is intentional, please say so in the PR body and add a test pinning that behavior for `authMode="share"` — right now nothing covers it either way.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -360,10 +372,30 @@ export function ChatMessage({
...
+      : failureText;
+
+  // latestTitle names the running step ("calling web_search"), which is part of
+  // the trace: with the trace hidden it would leak the very detail the status
+  // line replaces.
+  const statusTitle = showProcessView
+    ? latestTitle
+    : resolvedProcessStatus === "completed"
+      ? t("common.done")
+      : t("common.thinking");
```
rogercloud, line 426: **minor** (edge case) — this new `statusTitle` branch keys off `resolvedProcessStatus === "completed"`, but `GeneratingIndicator` receives `resolvedProcessStatus || taskStatus` (line 461) as its own status prop. If `resolvedProcessStatus` is falsy and the raw `taskStatus` is an off-contract string that `normalizeTaskStatus` can't map, `statusTitle` renders "thinking" copy *and* the indicator's dot animation keeps running for an already-finished turn.

Narrow — it needs a non-canonical backend status value with no terminal trace event — but it's a new branch interacting with a pre-existing fallback expression, so worth aligning the two to read from the same normalized value. Low priority.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -360,10 +372,30 @@ export function ChatMessage({
...
-      : errorMessage || t("common.errors.unknown");
+      : failureText;
+
+  // latestTitle names the running step ("calling web_search"), which is part of
+  // the trace: with the trace hidden it would leak the very detail the status
+  // line replaces.
+  const statusTitle = showProcessView
+    ? latestTitle
+    : resolvedProcessStatus === "completed"
+      ? t("common.done")
```
rogercloud: **nit** — the terminal-turn status line reuses the i18n key `common.done` ("Done", localized in zh as well), which everywhere else in the codebase is exclusively a `<Button>` label (`agent-triggers-dialog.tsx`, `pages/api-keys.tsx`, `pages/personal-api-keys-panel.tsx`, `workforce/workforce-config-panel.tsx`). Reusing an actionable-label string as a passive status descriptor is a mild semantic mismatch, not a functional issue. A dedicated key (e.g. `common.statusDone`) would read more precisely and avoid future translation drift. Optional.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -426,7 +458,7 @@ export function ChatMessage({
                   <div className="text-sm leading-relaxed break-words [overflow-wrap:anywhere]">{content}</div>
                 )
               ) : (
-                !isUser && showEmptyStatus && <GeneratingIndicator latestTitle={latestTitle} taskStatus={resolvedProcessStatus || taskStatus} errorMessage={errorMessage} />
+                !isUser && showEmptyStatus && <GeneratingIndicator latestTitle={statusTitle} taskStatus={resolvedProcessStatus || taskStatus} errorMessage={failureText} />
```
rogercloud: **simplification** — `GeneratingIndicator`'s own fallback (`errorMessage || t("common.errors.unknown")` at line 87) is now dead code. This call site always passes `failureText`, which the new logic at lines 375-377 guarantees is non-empty:

```ts
errorMessage || (showProcessView ? t("common.errors.unknown") : t("common.errors.taskFailed"))
```

Before this PR the call site passed raw `errorMessage` directly (which could be empty), so this inner fallback was live; routing through `failureText` is what makes it unreachable now. Render `{errorMessage}` directly, or drop the prop's optionality so the type reflects that it's always provided.

rogercloud (commented):
## Summary

This PR threads a new `showProcessView?: boolean` (default `true`) from `TaskConversationPanel` down into `ChatMessage`, and the public widget page passes `false` so embedded chat renders only the final answer. When the trace is hidden, pure-trace turns collapse to `null` instead of an empty bubble, the status line falls back to neutral `common.thinking`/`common.done` copy instead of the raw running-step title, and failures fall back to a generic `common.errors.taskFailed` instead of the raw backend error string. This addresses #1041 for the widget surface.

The approach is the simplest change that fits the existing architecture — the widget already threads sibling display flags (`showTokenUsage`, `showTaskFiles`, `showDagPreview`, …) through the same props, so adding one more is consistent, not over-engineered. Three design-level concerns are worth surfacing:

**1. Scope leak onto share links (actionable).** `PublicAgentChatPage` / `PublicConversationContent` backs both `/widget/chat/[token]` (`authMode="widget"`) and `/share/[token]` (`authMode="share"`). `showProcessView={false}` is passed unconditionally, so share links also lose the execution trace — a surface neither #1041 nor the PR body mentions, and no test covers `showProcessView` under `authMode="share"`. Since the prop is new, this is entirely new behavior rather than a pre-existing quirk. `authMode` is already destructured at the call site, so `showProcessView={authMode !== "share"}` is a one-line scoping fix if the intent was widget-only. Flagged inline as well.

**2. `showProcessView` conflates three concerns.** It currently controls (a) whether `TraceEventRenderer` mounts, (b) whether raw backend error text may be shown, and (c) whether the status line may show the real running-step name. Any future caller that treats "hide the trace" and "mask error text" as separable axes will be surprised they are coupled. Not blocking. A `displayMode`/`trusted` concept derived from `authMode` would separate the UI-display axis from the trust/redaction axis more explicitly.

**3. The fix is UI-only; the backend still streams the same raw data (follow-up, not a blocker).** This PR is frontend-only. The public widget websocket path still sends full LLM reasoning and tool call args/output: it reuses the same `normalize_public_trace_event` / public task trace filter used for authenticated task streams, with only credential-shaped keys redacted and a byte-size truncation — no widget-specific server-side stripping and no content filtering. So anyone opening devtools on a public widget page still sees the full reasoning and tool I/O. Given the widget is explicitly meant for external/customer embedding, this is a real exposure gap rather than an aesthetic one, but it is properly a backend follow-up. Please consider opening a tracking issue.

## Test coverage gaps

None of these block merge, but each is a real uncovered branch:

- No test covers the **internal** (non-widget) path of the rewritten `isProcessOnlyMessage`: `showProcessView={true}` + `content={null}` + `showEmptyStatus={false}` should still render the trace and produce no bubble. This is precisely the invariant that keying off trace events (instead of `shouldShowProcess`) could break for internal pages, and nothing locks it down.
- No test covers `showProcessView={true}` with a failed trace that has **no** extractable error string, i.e. the `common.errors.unknown` fallback firing while the trace is still shown. Only the trace-hidden (`taskFailed`) and trace-shown-with-error paths are covered.
- `task-conversation-panel.test.tsx` asserts prop forwarding for the page/widget cases only. `showProcessView` defaults flatly to `true` in `TaskConversationPanelProps`, while every sibling display flag is derived from `mode === "page"`. That asymmetry is untested for `mode="embedded-preview"` — likely fine, since only the widget opts out, but a test would pin the intent.
- No test asserts `showProcessView` behavior under `authMode="share"` (see design concern 1).

- **`frontend/src/components/chat/ChatMessage.tsx:87` — delete dead fallback.** `GeneratingIndicator`'s own `errorMessage || t("common.errors.unknown")` is now unreachable. Its only call site always passes `failureText`, which the new logic guarantees is non-empty (`errorMessage || (showProcessView ? t("common.errors.unknown") : t("common.errors.taskFailed"))`). Before this PR the call site passed raw `errorMessage` directly (possibly empty), so the inner fallback used to be live; routing through `failureText` is what kills it. Render `{errorMessage}` directly, or drop the prop's optionality.

net: -1 line possible.

## Re-review status

One substantive prior finding from the automated reviewer has been addressed in the current head commit:

- **Prior finding:** the guard checked `!copyableContent` instead of `!shouldShowProcess`, which could leave a floating copy button / action row with no visible bubble when `rawContent` was truthy and the trace was hidden.
- **Status: FIXED.** The current guard in `frontend/src/components/chat/ChatMessage.tsx:396-398` is exactly `if (isProcessOnlyMessage && !shouldShowProcess) { return null; }`, matching the suggested fix. A regression test in `frontend/src/components/chat/ChatMessage.test.tsx` ("drops it even when the turn carries copyable rawContent") renders with truthy `rawContent` and asserts the container renders nothing; it was confirmed to fail against the old `!copyableContent` guard when reverted temporarily. This is a genuine, test-locked fix.

## Test results

`vitest run src/components/chat src/components/widget src/components/task` — 104 tests passing (the PR body says 103; actual is 104). `tsc --noEmit` clean. An unrelated pre-existing OOM in the `src/components/build` suite is not attributable to this PR.

frontend/src/components/chat/ChatMessage.tsx
bluefish-08: Fixed in 526bcee7. `isProcessOnlyMessage` now keys off `isStoppedTraceProcessStatus(resolvedProcessStatus) && resolvedProcessStatus !== "completed"` (extracted as `isStoppedWithoutAnswer`). The condition change alone was not enough, though: a kept paused/waiting bubble would have rendered as a bare avatar, because the status line was gated on `showEmptyStatus`. The `GeneratingIndicator` gate is now `showEmptyStatus || (!showProcessView && isStoppedWithoutAnswer)`, so the turn shows the generic `taskPaused`/`waitingForUser` copy (no trace leak, no dots — both statuses are in the indicator's terminal list).

Tests added: paused and waiting_for_user with `showEmptyStatus={false}` and the trace hidden (both red against the old guard), plus the internal-page invariants (`showProcessView={true}` trace-only turn renders the trace and no bubble, with and without a stopped status).

One deliberate trade-off worth noting: in the common `ask_user` flow the clarification question itself is a visible assistant message, so the historical waiting turn now shows a redundant status line above it. Preferred redundant over the blank-turn failure mode this comment describes (which is real whenever the clarification text is filtered out as a placeholder); can switch to past-tense copy in a follow-up if it bothers.

frontend/src/components/widget/public-agent-chat-page.tsx
bluefish-08: Fixed in 0317d1a7, with one correction to the suggested snippet: `authMode !== "share"` evaluates to `true` for the widget, which would have shown the trace exactly where #1041 wants it hidden. Applied `showProcessView={authMode === "share"}` instead — share links keep the trace, the widget hides it. Both directions are pinned in `public-agent-chat-page.test.tsx` via a `data-show-process-view` attribute on the mocked panel, and `task-conversation-panel.test.tsx` additionally pins the `true` default for `mode="embedded-preview"`. PR body updated to state the scoping.

frontend/src/components/chat/ChatMessage.tsx
bluefish-08: Fixed in 526bcee7: the indicator now receives `resolvedProcessStatus` directly, so both it and `statusTitle` read the same normalized value. Behaviorally this is equivalent — the old `|| taskStatus` fallback could only ever inject a string that `normalizeTaskStatus` rejected, and any such off-contract string hits the same indicator branches as `undefined` (it matches none of the four terminal literals) — but the two reads no longer diverge.

frontend/src/components/chat/ChatMessage.tsx
bluefish-08: Fixed in 526bcee7: added a dedicated `common.statusDone` ("Done" / "已完成") and left `common.done` to its four button call sites.

frontend/src/components/chat/ChatMessage.tsx
bluefish-08: Taken one step further in 29ee3399. The whole `taskStatus === 'failed'` branch was unreachable, not just the fallback: the call site's outer ternary renders `failedMessageText` whenever `resolvedProcessStatus === "failed"`, so the indicator never mounts for a failed turn. Dropped the branch and the `errorMessage` prop entirely (net -8 lines); `failureText` remains in use for `failedMessageText`.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -360,10 +370,33 @@ export function ChatMessage({
       }
     }
   }
+  const failureText =
```
rogercloud, line 390: **Major — the failed-turn redaction above is display-only; the raw backend error still leaks through the copy button.**

`copyableContent` (line 287: `typeof content === "string" ? content : rawContent`) is not touched by this redaction. On a failed widget turn the bubble correctly renders `failedMessageText`, but the copy-to-clipboard button (line 479, rendered whenever `copyableContent` is truthy — true here, since `content` is a string) copies the unredacted backend error/exception text.

This is reachable on every failed widget turn: both producer paths (`UPSERT_STREAMING_FINAL_ANSWER` and the terminal `task_failed` handler in `app-context-chat.tsx`) write the raw failure text directly into `content`.

Suggested fix: have `copyableContent` fall back to `failedMessageText` when `resolvedProcessStatus === "failed"`, mirroring the display logic above — and add a test asserting copy behavior for a failed widget turn, which is currently uncovered.

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -215,6 +215,23 @@ function PublicConversationContent({
     setTaskId(null, { navigate: false })
   }, [authMode, connectionError, state.taskId, storageKey, setTaskId])
 
+  // Ending a conversation is purely client-side: drop the persisted id and the
+  // active taskId, and the visitor is back on the start screen; the next
+  // message creates a fresh task through handleSend. #1039
+  const handleNewConversation = useCallback(() => {
```
rogercloud, line 226: **Moderate — `handleNewConversation` doesn't clear `pendingMessage`, so a message can silently vanish.**

The handler clears `taskId`/localStorage/`isProcessing`/`currentTask`/draft state, but `pendingMessage` lives separately in `app-context-chat.tsx` and is left behind.

This is not a rare edge case. `handleSend` calls `setTaskId(newTaskId)` and then synchronously `sendMessage(...)` in the same tick, so **every** first message on a fresh task goes through the `pendingMessage` queue while the websocket connects — and this button has no `disabled` gating during that connect window. Click it there and the queued message's `targetTaskId` can never match (the new task's socket is torn down before connecting); it waits out the full 30s timeout and then rejects into an unhandled promise rejection. No toast, no visible error, message just disappears.

Fix either side: clear/cancel any pending message here, or have the pending-message matching effect bail out once `state.taskId` no longer matches `targetTaskId`.

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -215,6 +215,23 @@ function PublicConversationContent({
...
+    safeRemoveItem(storageKey)
+    setTaskId(null, { navigate: false })
+    // Nulling taskId closes the socket, so no terminal WS event will ever
+    // reset these. Left stale mid-run, isProcessing keeps the start screen's
+    // composer disabled forever and currentTask pins the header on
+    // "Connecting..." — a reload would be the only way out.
```
rogercloud: **Minor — same handler: `isHistoryLoading` isn't cleared and can stick `true`.**

If the reset happens before a task's websocket `onConnect` fires, the 2-second safety timeout that normally clears this flag was never scheduled, so it stays set. This is user-visible: the header status text (`showStatus`) is gated only on `!createTaskError`, not on having a task, so the start screen can show "initializing" indefinitely until the user starts a new task or reloads.

Together with the `pendingMessage` gap above, this argues for a single `resetConversation()` action owned by the chat context (which already knows its full state shape) rather than an enumerated reset list at this call site.

frontend/src/components/task/task-conversation-panel.tsx
```diff
@@ -694,10 +699,15 @@ export function TaskConversationPanel({
...
+                        showProcessView={showProcessView}
                         processStatus={item.processStatus}
                         taskStatus={
-                          isFailedFinalAnswerStream
+                          // With the trace hidden, a failed result message
+                          // (final_answer_error stream or the verbatim terminal
+                          // reason, #893) must reach ChatMessage as failed so
+                          // its raw error text gets the generic replacement.
+                          isFailedFinalAnswerStream ||
+                          (!showProcessView && item.status === "failed")
```
rogercloud: **Simplification — duplicated status test.** `isFailedFinalAnswerStream || (!showProcessView && item.status === "failed")` checks `item.status === "failed"` on both sides: it's `(A && S) || (P && S)`, which is `S && (A || P)`. Factoring the shared `item.status === "failed"` test out once reads better and is algebraically identical.

rogercloud (commented):
## Follow-up review (e4dbe2b6 → 95bc6c8c)

Six new commits landed since the last review: `526bcee7` and `0317d1a7` address the two prior major findings (stopped/paused/waiting_for_user turns disappearing in widget mode; share links losing the trace), `29ee3399` drops the dead `errorMessage` fallback, `68957054` resets processing state when a conversation ends mid-run, and `95bc6c8c` redacts failed-turn content when the trace is hidden. `c65bed54` adds a **new feature** — a "New conversation" control for guest/widget chat — so this PR now closes both #1041 (original scope) and #1039 (newly linked).

All five prior findings are genuinely fixed and test-locked. Three new confirmed findings below, one of them major.

### Prior findings status

| 1 | Paused/`waiting_for_user` turns dropped in widget mode (major) | **FIXED** | `frontend/src/components/chat/ChatMessage.tsx:341-358` now computes `isStoppedWithoutAnswer = isStoppedTraceProcessStatus(resolvedProcessStatus) && resolvedProcessStatus !== "completed"` (`526bcee7`); new tests cover `paused` and `waiting_for_user` explicitly. |
| 2 | Share links losing the trace (major) | **FIXED** | `frontend/src/components/widget/public-agent-chat-page.tsx:434` now passes `showProcessView={authMode === "share"}` (`0317d1a7`); tests assert `data-show-process-view` for both widget (`"false"`) and share (`"true"`). |
| 3 | `statusTitle`/`GeneratingIndicator` mismatch on off-contract status strings (minor) | **FIXED** | Call site now passes `taskStatus={resolvedProcessStatus}` only; the `|| taskStatus` fallback is gone and `resolvedProcessStatus` already folds in the raw status in `trace-process-status.ts`, so no coverage was lost. |
| 4 | `common.done` reused as a status-line string (nit) | **FIXED** | Dedicated `common.statusDone` key added (distinct zh string from `common.done`); remaining `common.done` uses are genuinely button labels. |
| 5 | Dead `errorMessage` fallback in `GeneratingIndicator` (simplification) | **FIXED** | Prop removed from the signature entirely in `29ee3399`; failed-turn text now renders via a mutually-exclusive `resolvedProcessStatus === "failed"` branch — cleaner than either option originally suggested. |

### Design verdict for the new code: acceptable with reservations

The core #1041 fix (threading `showProcessView`) remains minimal and idiomatic — it mirrors sibling flags like `showTokenUsage`. Bundling #1039 into the same PR is a process nit rather than a blocker: the two scopes are genuinely disjoint (render-gating vs. widget-local state reset) and share only one file. Two design-level reservations on the newly added code:

**1. The "New conversation" reset is manually enumerated state clearing, not a single owned action.** `handleNewConversation` in `frontend/src/components/widget/public-agent-chat-page.tsx` has to independently know that nulling `taskId` tears down the socket — and therefore that no terminal WS event will ever arrive to reset `isProcessing`/`currentTask` — so it dispatches those resets by hand alongside clearing localStorage and the draft fields. That enumeration is already incomplete: it misses `pendingMessage` (a real, reachable bug — see below) and `isHistoryLoading` (a real cosmetic bug — see below). This is the characteristic failure mode of enumerate-everything resets: every future piece of chat-context state must be remembered again at this call site. A `resetConversation()` action owned by the chat context — which already knows its own full state shape — would be materially more robust. Not a blocker for this PR, but worth doing before a third caller appears.

**2. Clicking "New conversation" does not ask the backend to stop the in-flight task.** The old task keeps running server-side with nobody listening. This matches the issue's stated scope ("no backend change is needed") and is presumably an accepted trade-off; naming it explicitly as a design constraint rather than a defect. No action needed unless the team wants to revisit.

### New findings

**Major**

- **`frontend/src/components/chat/ChatMessage.tsx` — redaction is display-only; the raw backend error still leaks via the copy button.** The entire point of `95bc6c8c` is to keep raw backend failure text out of the widget, but `copyableContent` (line 287: `typeof content === "string" ? content : rawContent`) is untouched by the redaction. On a failed widget turn the bubble correctly shows the generic `failedMessageText`, while the copy button (line 479, rendered whenever `copyableContent` is truthy — true here since `content` is a string) still copies the unredacted backend error/exception text. Both message-construction paths (`UPSERT_STREAMING_FINAL_ANSWER` and the terminal `task_failed` handler in `app-context-chat.tsx`) put raw failure text straight into `content`, so this is reachable on every failed widget turn. No test covers copy behavior for a failed widget turn. Suggested fix: have `copyableContent` fall back to `failedMessageText` when `resolvedProcessStatus === "failed"`, mirroring the display logic.

**Moderate**

- **`frontend/src/components/widget/public-agent-chat-page.tsx:221` (`handleNewConversation`) — doesn't clear `pendingMessage`, silently swallowing a message.** `handleSend` calls `setTaskId(newTaskId)` and then synchronously `sendMessage(...)` in the same tick, so **every** first message on a fresh task routes through the `pendingMessage` queue while the websocket connects — this is the normal path, not an edge case. The "New conversation" button has no `disabled` gating during that connect window. Click it there and the queued message's `targetTaskId` can never match (the new task's socket is torn down before connecting); it sits for the full 30-second timeout and then rejects into an unhandled promise rejection. No toast, no error — the message just vanishes.

**Minor**

- **Same handler — `isHistoryLoading` isn't cleared and can stick `true`.** If the reset happens before a task's websocket `onConnect` fires, the 2-second safety timeout that normally clears this flag was never scheduled, so it stays set. This is user-visible: the header status text (`showStatus`) is gated only on `!createTaskError`, not on having a task, so the start screen can show "initializing" indefinitely until the user starts a new task or reloads.

_Two candidate findings were investigated and dropped: "partial legitimate answer discarded on failure" (`content` on a failed turn is backend failure text by construction on every producer path, so there is no real answer to lose — and the trade-off is documented in the PR body), and "task-level failed status leaking onto an earlier successful message" (`item.status` is derived per-message from its own event history; no task-level fallback exists). A third candidate — "unrelated i18n keys added in this PR" — was also investigated and dropped: those keys already exist at this PR's actual merge-base and are untouched by this diff; an earlier check of that claim used the wrong base commit._

- `frontend/src/components/chat/ChatMessage.tsx:93` — the dot-animation exclusion array still lists `"failed"`, but `GeneratingIndicator`'s only call site (line 465) is reachable only via the `else` branch of a `resolvedProcessStatus === "failed"` ternary, so `taskStatus` can never be `"failed"` there. Dead entry; drop it, no behavior change.
- `frontend/src/components/task/task-conversation-panel.tsx:709-710` — `isFailedFinalAnswerStream || (!showProcessView && item.status === "failed")` tests `item.status === "failed"` twice; it's `(A && S) || (P && S)`, algebraically `S && (A || P)`. Factor the shared test once.

net: -6 lines possible.

### Tests

`ChatMessage.test.tsx` 17/17 pass, `public-agent-chat-page.test.tsx` 21/21 pass (spot-checked directly). Full-suite green plus clean lint/type-check per the PR author; not independently re-run in full for this update.

frontend/src/components/chat/ChatMessage.tsx
bluefish-08: Fixed in c44a4da3. `copyableContent` now derives from `failedMessageText` on failed assistant turns, so the copy button hands out exactly what the bubble shows — redacted in the widget, verbatim on internal pages and share links. Test added: on a failed hidden-trace turn the copy button writes `common.errors.taskFailed` to the clipboard, not the raw backend error.

frontend/src/components/widget/public-agent-chat-page.tsx
bluefish-08: Fixed in 96271d3e, taking the second suggested direction: a context-level effect rejects a queued message as soon as `state.taskId` no longer matches its `targetTaskId`, rather than a widget-local cleanup — the 30s-silent-vanish failure mode is gone for every caller, not just this button. False-positive safety was checked per queue site: every one of them sets the task id in the same tick it queues (`setTaskId` is fully synchronous), so the effect only fires on a genuine switch-away, and the rejection goes through the queue's wrapped reject so the 30s timer is cleared too. Test: a queued opening message rejects immediately once `setTaskId(null)` lands, and `sendChatMessage` is never called.

frontend/src/components/widget/public-agent-chat-page.tsx
bluefish-08: Fixed in 96271d3e — the reset also dispatches `SET_HISTORY_LOADING: false` now, and the mid-run test asserts all three resets. On `resetConversation()`: agreed on the direction, held off while this is the only caller — and with the pendingMessage gap now handled inside the context itself, what future callers must remember shrinks to display flags. Will extract the owned action when a second caller appears.

frontend/src/components/task/task-conversation-panel.tsx
bluefish-08: Applied in 9b134dfb as `isFailedResultMessage = item.status === "failed" && (item.isStreamingFinalAnswer || !showProcessView)`. The dead `"failed"` entry in the dot-animation array went in c44a4da3 along with the copy-button fix.

rogercloud (approved):
## Follow-up review (round 3) — since commit `95bc6c8c`, now at `9b134dfb`

### What changed
Three targeted fix commits landed, each closing one finding from the previous round: `c44a4da3` (redact failed-turn copy button text + drop dead dot-animation entry), `96271d3e` (fail queued messages fast on conversation reset), and `9b134dfb` (factor the duplicated failed-status test in `taskStatus` wiring).

### Prior findings status — all resolved

| # | Round | Finding (severity) | Status | Evidence |
|---|-------|--------------------|--------|----------|
| 1 | R1 | Paused/`waiting_for_user` trace-only turns dropped in widget mode (major) | FIXED | `isStoppedWithoutAnswer` logic intact — `frontend/src/components/chat/ChatMessage.tsx:334-336` |
| 2 | R1 | Share links losing the trace (major) | FIXED | `showProcessView={authMode === "share"}` intact — `frontend/src/components/widget/public-agent-chat-page.tsx:436` |
| 3 | R1 | `statusTitle`/`GeneratingIndicator` status mismatch (minor) | FIXED | Call site still passes `taskStatus={resolvedProcessStatus}` only; no reintroduced fallback |
| 4 | R1 | `common.done` reused as status-line string (nit) | FIXED | Dedicated `common.statusDone` key still in place and used |
| 5 | R1 | Dead `errorMessage` fallback in `GeneratingIndicator` (simplification) | FIXED | Prop still absent from the component signature |
| 6 | R2 | Copy button leaked raw backend error on failed widget turns (major) | FIXED by `c44a4da3` | `copyableContent` now branches to `failedMessageText` for failed non-user turns — `frontend/src/components/chat/ChatMessage.tsx:376-379`; internal pages still copy the real error |
| 7 | R2 | `handleNewConversation` didn't clear `pendingMessage`, silently losing a queued message (moderate) | FIXED by `96271d3e` | New watcher rejects any pending message whose `targetTaskId` no longer matches `state.taskId` — `frontend/src/contexts/app-context-chat.tsx:1380-1391` |
| 8 | R2 | `handleNewConversation` didn't clear `isHistoryLoading`, sticking header on "Initializing" (minor) | FIXED by `96271d3e` | Now dispatches `SET_HISTORY_LOADING: false` on reset — `frontend/src/components/widget/public-agent-chat-page.tsx:233` |
| 9 | R2 | Duplicated `item.status === "failed"` test in `taskStatus` wiring (simplification) | FIXED by `9b134dfb` | `isFailedResultMessage = item.status === "failed" && (item.isStreamingFinalAnswer || !showProcessView)` — algebraically identical, `frontend/src/components/task/task-conversation-panel.tsx:697-699` |

### Design verdict — sound

An independent blank-slate pass over all three commits found no new correctness issues. The `taskStatus` refactor is behaviorally identical to the prior duplicated form (the old `isFailedFinalAnswerStream` had no other consumer), and the copy-button redaction makes `copyableContent` mirror `failedMessageText` exactly on failed turns without regressing internal pages.

One detail from the fail-fast fix (`96271d3e`) is worth stating plainly, as it corrects an ambiguity from earlier reasoning: the reject-fast rejection is **not** a silent unhandled rejection. It propagates through `ChatInput.tsx`'s `await`/`catch` — the path both widget composers funnel through — which calls `toast.error(...)`. So clicking "New conversation" mid-connect now produces an immediate, visible "message not sent" toast instead of the message silently disappearing after the 30s timeout. That is a genuine UX improvement, not just faster cleanup. The effect fires only on a real `state.taskId !== targetTaskId` divergence and never on an in-flight message that would still succeed (both queuing sites set `taskId` and queue in the same synchronous tick).

### Residual note (non-blocking)
The new regression test in `app-context-chat.test.tsx` exercises the reducer/effect mechanism directly, so the underlying reject-fast behavior is test-locked, but the toast-surfacing itself (the full `ChatInput` → `toast.error` chain) is not. Worth a mention only; not blocking.

### Tests
- `ChatMessage.test.tsx`: 18/18 pass
- `task-conversation-panel.test.tsx`: 16/16 pass
- `app-context-chat.test.tsx` + `public-agent-chat-page.test.tsx`: 44/44 pass combined

All directly re-run this round.

### Recommendation
Every finding raised across all three review rounds is now resolved, test-locked, and confirmed non-regressed, with no new findings.

PR #1058, opened 2026-07-29
fix(widget): require a published agent on every widget entry point (#1055)

### LINKED ISSUE #1055: fix(widget): unpublishing an agent does not revoke its widget (no PUBLISHED check)
Found during the re-review of #1049 (@rogercloud) and verified against `main`.

## Problem

`ensure_widget_agent_available` is the only one of the three availability helpers that does not gate on publication state:

| helper | status gate |
| --- | --- |
| `ensure_share_agent_available` | `agent.status != AgentStatus.PUBLISHED` -> 403 |
| `ensure_widget_workforce_available` | `workforce.status != "active"` -> 403 |
| **`ensure_widget_agent_available`** | **none** |

`AgentStore.unpublish_agent` (`src/xagent/web/services/agent_store.py:590`) only flips the status:

```python
agent.status = AgentStatus.DRAFT
```

It never touches `widget_enabled` or `widget_key`. So an unpublished agent stays fully reachable through its existing widget key: `/api/widget/auth` still mints guest tokens for it, and outstanding guest JWTs keep working on both the HTTP and WS paths.

An owner who unpublishes an agent to take it out of service would reasonably expect its embedded widget to stop serving. It does not.

## Expected

Unpublishing revokes widget access the same way disabling the widget or rotating the key does — i.e. `ensure_widget_agent_available` gains a `status != AgentStatus.PUBLISHED` check, so it is enforced on every request and drops live WS guests on their next message (the revalidation loop covered by #992).

Worth deciding explicitly whether that is the right lever, or whether `unpublish_agent` should instead clear `widget_enabled`. The helper-side check is preferable: it is re-derived per request, so it also covers agents unpublished by any other code path.

## Scope note

Pre-existing — `git blame` shows the function unchanged by #1049, whose tests scope themselves to the two revocation levers that do work (disable + key rotation). A fix here should add the third case to `tests/web/api/test_widget_ws_revalidation.py`, which already has the fixtures for it.

codeacme17:
Closes #1055.

## Problem

The agent widget channel was never gated on publication state. `AgentStore.unpublish_agent` only flips `status` — it deliberately leaves `widget_enabled` and `widget_key` intact — so an unpublished agent stayed fully reachable through its existing widget key: `/api/widget/auth` kept minting guest tokens, and outstanding guest JWTs kept working on both the HTTP and WS paths.

The workforce widget channel already gates on `status == "active"` at two layers, and the agent *share* channel gates on `PUBLISHED`. Only the agent widget channel did not — at any layer.

## Fix

All three agent-widget entry points now require `AgentStatus.PUBLISHED`:

| entry point | endpoints it guards |
| --- | --- |
| `ensure_widget_agent_available` | `/chat/task/create`, `/files/upload`, `/chat/ws/{task_id}` (per request, and per inbound WS message), plus the public file routes (see below) |
| `_resolve_widget_agent_by_key` | `/api/widget/auth` (direct visit), `/api/widget/embed-ticket` |
| `_get_widget_enabled_agent` → renamed `_get_live_widget_agent` | `/api/widget/auth` (embed-ticket redemption) |

The issue only called for the first. The other two matter because fixing the revalidation helper alone would still let `/api/widget/auth` mint a guest token for an unpublished agent that then 403s on its first real message — and would leave a live embed ticket redeemable inside its 60s TTL.

`_get_live_widget_agent` also gained a non-null `widget_key` check, for parity with its three sibling gates (`_resolve_widget_agent_by_key`, `ensure_widget_agent_available`, `_workforce_owner_from_ticket`), which all had one. Nothing produces `widget_enabled=True, widget_key=None` today and no schema constraint forbids it either; without the check such a row would redeem a ticket into a guest token carrying `widget_key: null` that then 403s on its first real request instead of failing cleanly at redemption.

### Why the helper-side lever

Publication is checked in the availability helpers rather than by having `unpublish_agent` clear `widget_enabled`:

- it is re-derived per request, so it covers every code path that takes an agent out of `PUBLISHED`, not just this one function;
- a state-clearing approach structurally cannot cover a never-published draft, which never passes through `unpublish_agent` at all (see below);
- it does not destroy the owner's widget configuration — re-publishing restores the same embed snippet with no re-deploy;
- it matches how the share channel already does it.

Denials reuse the existing `"Invalid widget key"` / `"Widget is disabled for this agent"` details, so publication state stays unprobeable.

## Behaviour change beyond the issue text

**Never-published drafts.** Agents are created `widget_enabled=True` with a key already minted (`agent_store.py`), so a never-published draft also carried a live widget credential. Those keys stop authenticating too. No supported flow distributed such a key — the embed snippet is only reachable from the Deploy dialog, which opens only for published agents (`frontend/src/app/build/page.tsx`) — but it is a real change in what the API accepts, so calling it out explicitly.

**Public file routes.** `ensure_widget_agent_available` is also reached transitively from the public file preview/download/open routes, via `_validate_public_task_file_access` (`files.py`) → `get_public_chat_user`. So unpublishing additionally cuts a widget guest off from that task's file attachments. Correct and fail-safe — the attachments belong to a channel that is no longer serving — but it is part of the shipped blast radius, so it is now called out here and covered by a test.

**WS timing, unchanged.** For an already-connected WS guest, revocation lands on the guest's next *inbound* message: an unpublished agent's socket keeps receiving outbound/broadcast frames and in-flight runs keep executing until the guest speaks. That is the contract the existing disable/rotate levers already have, so this is consistent with prior behaviour rather than a new gap — but "owner clicks Unpublish" reads as *stop now*, and closing that delta for all three levers at once is worth a follow-up issue.

## Tests

- Unpublish revokes an existing guest token on its next HTTP request; re-publishing revives the same token (pins the reversibility the helper-side lever buys).
- `/api/widget/auth` and `/api/widget/embed-ticket` reject an unpublished agent's key, and a never-published agent's key — parametrized over `DRAFT` and `ARCHIVED` to pin the gate as `!= PUBLISHED` rather than `== DRAFT`.
- An embed ticket minted while published no longer redeems after an unpublish.
- `ARCHIVED` variants for the other two gates as well (`ensure_widget_agent_available` and `_get_live_widget_agent`). Both are otherwise only reached through the real `/unpublish` endpoint, which hardcodes `DRAFT` and can never produce `ARCHIVED`, so a narrowed `== DRAFT` gate would have passed. Written via a direct status write, since no route assigns `ARCHIVED`.
- A `widget_enabled=True, widget_key=None` agent fails embed-ticket redemption rather than minting a keyless guest token.
- A widget guest's file download 403s after the agent is unpublished — the first widget/guest file-access assertion in the suite.
- `unpublish` added as a third case to the WS `4003` revalidation matrix in `test_widget_ws_revalidation.py`, for both the agent and workforce channels (the workforce case exercises pre-existing behaviour and keeps the `channel × revocation` matrix square).

Each of the four source-side checks was verified to be independently pinned: removing any one of them fails at least one test.

## Not in this PR

The Deploy dialog gates its **Share link** tab on publication but not its **Embed** tab. That state is unreachable today (the dialog only opens for published agents), so the missing notice is defence-in-depth rather than a live gap — left as a follow-up.

## Verification

`tests/web` passes in full. Whole-repo runs on this machine show pre-existing order-dependent `caplog` failures (including four widget-allowlist tests whose behavioural assertions pass and only whose log capture is empty); all pass in isolation. `ruff`, `ruff format`, `mypy`, `isort`, `codespell`, `npm lint`, `npm type-check` all pass.

### DIFF
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -166,14 +166,19 @@ def ensure_widget_agent_available(
     Mirrors :func:`ensure_share_agent_available` for the widget channel. The
     guest JWT is long-lived (30-day TTL), so access is re-derived from the
     agent's *current* widget state on every request rather than trusted from
-    the token alone: disabling the widget (``widget_enabled = False``) or
-    rotating ``widget_key`` cuts off outstanding guest tokens on their next
-    request.
+    the token alone. Three owner-side levers cut off outstanding guest tokens
+    on their next request: disabling the widget (``widget_enabled = False``),
+    rotating ``widget_key``, and unpublishing the agent (#1055).
+
+    Publication is checked here rather than by having ``unpublish_agent`` clear
+    ``widget_enabled``, so that the gate is re-derived per request and covers
+    every code path that takes an agent out of ``PUBLISHED`` — and so that
+    re-publishing restores the widget without the owner re-deploying it.
 
     ``expected_widget_key`` is the key carried by the guest JWT; when present
     it must still match the agent's live key. Tokens minted before the key was
     embedded carry no key and skip that comparison — they stay gated on
-    ``widget_enabled`` so the disable path still revokes them.
+    ``widget_enabled`` and publication so those paths still revoke them.
     """
     agent = db.query(Agent).filter(Agent.id == widget_agent_id).first()
     if (
@@ -182,6 +187,7 @@ def ensure_widget_agent_available(
         or agent.user_id != user_id
         or not agent.widget_enabled
         or not agent.widget_key
+        or agent.status != AgentStatus.PUBLISHED
         or (expected_widget_key is not None and agent.widget_key != expected_widget_key)
     ):
         raise HTTPException(status_code=403, detail="Widget is unavailable")
--- a/src/xagent/web/api/widget.py
+++ b/src/xagent/web/api/widget.py
@@ -20,7 +20,7 @@ from pydantic import BaseModel, Field
 from sqlalchemy.orm import Session
 
 from ..auth_config import JWT_ALGORITHM, JWT_SECRET_KEY
-from ..models.agent import Agent, is_workforce_generated_manager_agent
+from ..models.agent import Agent, AgentStatus, is_workforce_generated_manager_agent
 from ..models.database import get_db
 from ..models.deployment import DeploymentOwnerType
 from ..models.user import User
@@ -117,21 +117,38 @@ class ResolvedWidgetOwner:
     workforce: Workforce | None = None
 
 
-def _get_widget_enabled_agent(db: Session, agent_id: int) -> Agent:
-    """Load a widget-enabled agent or raise the matching HTTP error."""
+def _get_live_widget_agent(db: Session, agent_id: int) -> Agent:
+    """Load an agent whose widget is actually serving, or raise the matching
+    HTTP error.
+
+    "Live" is three conditions, not just ``widget_enabled``: a real non-manager
+    agent, its widget enabled, and the agent still published.
+
+    External access requires the agent to still be published, mirroring
+    ``_workforce_owner_from_ticket``'s ``status == "active"`` check: a ticket
+    minted before an unpublish must not still redeem inside its TTL (#1055).
+    Unpublished collapses into the same "disabled" 403 so the ticket path does
+    not leak an agent's publication state.
+    """
     agent = db.query(Agent).filter(Agent.id == agent_id).first()
     if agent is None or is_workforce_generated_manager_agent(agent):
         raise HTTPException(
             status_code=401, detail="Widget owner not found or invalid agent_id"
         )
-    if not agent.widget_enabled:
+    if not agent.widget_enabled or agent.status != AgentStatus.PUBLISHED:
         raise HTTPException(status_code=403, detail="Widget is disabled for this agent")
     return agent
 
 
 def _resolve_widget_agent_by_key(db: Session, widget_key: str) -> Agent | None:
     """Return a widget-enabled agent for the key, or ``None`` if no eligible
-    agent matches (unknown key, disabled widget, or a workforce-manager agent).
+    agent matches (unknown key, disabled widget, unpublished agent, or a
+    workforce-manager agent).
+
+    External access requires the agent to still be published, mirroring the
+    workforce branch of :func:`_resolve_widget_owner_by_key` (#1055). Agents are
+    created ``widget_enabled=True`` with a key already minted, so without this
+    a never-published draft would serve its widget too.
 
     Returning ``None`` — rather than raising — lets the caller fall through to
     the workforce lookup; every real failure still collapses to a single 403
@@ -143,6 +160,7 @@ def _resolve_widget_agent_by_key(db: Session, widget_key: str) -> Agent | None:
         or not agent.widget_key
         or is_workforce_generated_manager_agent(agent)
         or not agent.widget_enabled
+        or agent.status != AgentStatus.PUBLISHED
     ):
         return None
     return agent
@@ -317,7 +335,7 @@ def _owner_from_embed_ticket(db: Session, embed_ticket: str) -> ResolvedWidgetOw
     ticket_agent_id = claims.get("agent_id")
     if not isinstance(ticket_agent_id, int):
         raise HTTPException(status_code=403, detail="Invalid or expired embed ticket")
-    agent = _get_widget_enabled_agent(db, ticket_agent_id)
+    agent = _get_live_widget_agent(db, ticket_agent_id)
     allowed_domains = agent.allowed_domains
     # Re-check so tickets die immediately if the allowlist shrinks.
     require_domain_allowed(
--- a/src/xagent/web/api/widget.py
+++ b/src/xagent/web/api/widget.py
@@ -121,21 +121,34 @@ def _get_live_widget_agent(db: Session, agent_id: int) -> Agent:
     """Load an agent whose widget is actually serving, or raise the matching
     HTTP error.
 
-    "Live" is three conditions, not just ``widget_enabled``: a real non-manager
-    agent, its widget enabled, and the agent still published.
+    "Live" is four conditions, not just ``widget_enabled``: a real non-manager
+    agent, its widget enabled, a widget key still minted, and the agent still
+    published.
 
     External access requires the agent to still be published, mirroring
     ``_workforce_owner_from_ticket``'s ``status == "active"`` check: a ticket
     minted before an unpublish must not still redeem inside its TTL (#1055).
     Unpublished collapses into the same "disabled" 403 so the ticket path does
     not leak an agent's publication state.
+
+    ``widget_key`` is checked for parity with the three sibling gates
+    (:func:`_resolve_widget_agent_by_key`, ``ensure_widget_agent_available``,
+    :func:`_workforce_owner_from_ticket`). No path produces
+    ``widget_enabled=True`` with no key today, but nothing at the schema level
+    forbids it either; without this check such a row would redeem a ticket into
+    a guest token that then 403s on its first real request instead of failing
+    here.
     """
     agent = db.query(Agent).filter(Agent.id == agent_id).first()
     if agent is None or is_workforce_generated_manager_agent(agent):
         raise HTTPException(
             status_code=401, detail="Widget owner not found or invalid agent_id"
         )
-    if not agent.widget_enabled or agent.status != AgentStatus.PUBLISHED:
+    if (
+        not agent.widget_enabled
+        or not agent.widget_key
+        or agent.status != AgentStatus.PUBLISHED
+    ):
         raise HTTPException(status_code=403, detail="Widget is disabled for this agent")
     return agent
 
--- a/tests/web/api/test_agents_management.py
+++ b/tests/web/api/test_agents_management.py
@@ -1203,6 +1203,174 @@ def test_rotated_widget_key_invalidates_existing_guest_tokens() -> None:
     assert response.json()["detail"] == "Widget is unavailable"
 
 
+def test_unpublishing_agent_invalidates_existing_widget_guest_tokens() -> None:
+    """Unpublishing an agent must invalidate already-issued guest JWTs on their
+    next request — the third revocation lever alongside disable and rotate.
+
+    ``unpublish_agent`` only flips ``status``; it leaves ``widget_enabled`` and
+    ``widget_key`` intact on purpose, so that re-publishing restores the same
+    embed snippet. That makes the per-request check in
+    ``ensure_widget_agent_available`` the thing that has to enforce this (#1055).
+    """
+    headers = _admin_headers()
+    owner_id = _user_id("admin")
+    agent_id = _create_agent_row(
+        user_id=owner_id,
+        name="Unpublish Invalidate Widget Agent",
+        status=AgentStatus.PUBLISHED,
+        widget_enabled=True,
+    )
+    guest_headers = _authenticate_widget_guest(agent_id=agent_id)
+
+    # Sanity: the guest token works while the agent is published.
+    assert _create_widget_task(guest_headers, agent_id).status_code == 200
+
+    unpublished = client.post(f"/api/agents/{agent_id}/unpublish", headers=headers)
+    assert unpublished.status_code == 200, unpublished.text
+
+    response = _create_widget_task(guest_headers, agent_id)
+    assert response.status_code == 403, response.text
+    assert response.json()["detail"] == "Widget is unavailable"
+
+
+def test_republishing_agent_restores_widget_access_for_existing_tokens() -> None:
+    """Unpublish is a reversible lever: re-publishing revives the same guest
+    token, because the widget key it pins was never rotated. Locks in why the
+    status check lives in the availability helper rather than in
+    ``unpublish_agent`` clearing ``widget_enabled`` (#1055)."""
+    headers = _admin_headers()
+    owner_id = _user_id("admin")
+    agent_id = _create_agent_row(
+        user_id=owner_id,
+        name="Republish Widget Agent",
+        status=AgentStatus.PUBLISHED,
+        widget_enabled=True,
+    )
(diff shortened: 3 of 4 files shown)


src/xagent/web/api/widget.py
```diff
@@ -117,21 +117,38 @@ class ResolvedWidgetOwner:
...
+    Unpublished collapses into the same "disabled" 403 so the ticket path does
+    not leak an agent's publication state.
+    """
     agent = db.query(Agent).filter(Agent.id == agent_id).first()
     if agent is None or is_workforce_generated_manager_agent(agent):
         raise HTTPException(
             status_code=401, detail="Widget owner not found or invalid agent_id"
         )
-    if not agent.widget_enabled:
+    if not agent.widget_enabled or agent.status != AgentStatus.PUBLISHED:
```
rogercloud: [Low] This is the only one of the four sibling gate helpers that doesn't also check `widget_key` is non-null — `_resolve_widget_agent_by_key` (widget.py:160), `ensure_widget_agent_available` (public_chat_access.py:189), and `_workforce_owner_from_ticket` (widget.py:287) all check it. The docstring above frames "live" as exactly three conditions, so the omission reads as deliberate.

It's unreachable today: `add_agent` always mints a key when `widget_enabled=True` (core/agent_store.py:388), the update path self-heals by minting one if the flag is toggled on without a key (api/agents.py:744-749), rotation never nulls the key, and legacy rows were backfilled. But there's no DB-level NOT-NULL-when-enabled constraint, so the invariant is latent rather than structural. If a future path ever creates `widget_enabled=True, widget_key=None`, this helper mints a guest token that 403s on the very next real request instead of failing at redemption.

Suggest `or not agent.widget_key` here for consistency with the three siblings, or a docstring line explaining why the key is intentionally excluded.

tests/web/api/test_agents_management.py
```diff
@@ -1203,6 +1203,174 @@ def test_rotated_widget_key_invalidates_existing_guest_tokens() -> None:
...
+    response = client.post(
+        "/api/widget/auth",
+        json={"widget_key": widget_key, "guest_id": "guest-1"},
+    )
+    assert response.status_code == 403, response.text
+    # Same detail as an unknown key: publication state must not be probeable.
+    assert response.json()["detail"] == "Invalid widget key"
+
+
+@pytest.mark.parametrize("status", [AgentStatus.DRAFT, AgentStatus.ARCHIVED])
```
rogercloud, line 1305: [Low] Nice reasoning in the docstring for parametrizing over both non-published states — a DRAFT-only test wouldn't catch a narrowing bug like `== AgentStatus.DRAFT` instead of `!= AgentStatus.PUBLISHED`.

That argument applies equally to the other two gates, but they don't get an ARCHIVED variant. `_get_live_widget_agent` (web/api/widget.py:138) and `ensure_widget_agent_available` (web/api/public_chat_access.py:190) are only reached through the real `/unpublish` endpoint, which hardcodes `AgentStatus.DRAFT` (core/agent_store.py:590) and can never produce ARCHIVED — so a narrowing bug in either of those two lines would still pass.

Low severity: `AgentStatus.ARCHIVED` isn't assigned anywhere in production code today, so it's a dead path.

src/xagent/web/api/public_chat_access.py
```diff
@@ -182,6 +187,7 @@ def ensure_widget_agent_available(
         or agent.user_id != user_id
         or not agent.widget_enabled
         or not agent.widget_key
+        or agent.status != AgentStatus.PUBLISHED
```
rogercloud, line 190: [Low / informational] Worth calling out that this check has a wider blast radius than the PR description covers: `ensure_widget_agent_available` is also reached from the public file preview/download/open routes via `_validate_public_task_file_access` (web/api/files.py:624) -> `get_public_chat_user` (web/api/files.py:653). So unpublishing now also cuts widget guests off from that task's file attachments.

That's correct and fail-safe, but it's untested — no test in tests/web/test_file_upload.py or tests/web/api/test_admin_file_access.py uses a widget/guest token, and nothing combines unpublish with a file download. Suggest a one-line mention in the PR description, or a small test asserting a guest's file download 403s after unpublish.

tests/web/api/test_widget_ws_revalidation.py
```diff
@@ -177,9 +178,11 @@ class _WidgetGuest:
 
     token: str
     task_id: int
-    # Owner-side revocations, each of which must invalidate ``token``.
-    disable: Callable[[], None]
-    rotate: Callable[[], None]
+    # Owner-side revocation levers keyed by the ``revocation`` parametrize id,
```
rogercloud: L181: yagni — `Mapping[str, Callable[[], None]]` for a closed 3-key set with a single consumer. Both builders (L223, L263) populate the same hardcoded disable/rotate/unpublish keys, and there's exactly one read site (L334), so the comment's rationale ("not every call site in between") doesn't apply — there is only one.

A plain `unpublish: Callable[[], None]` field next to the existing ones, dispatched with `getattr(guest, revocation)()` or a small if/elif, gives identical coverage and failure clarity with one less abstraction layer and two fewer typing imports. Roughly line-count neutral.

rogercloud (approved):
## Summary

This closes the gap where the agent-widget channel was the only external entry point that never re-derived publication state: `widget_enabled` + a live `widget_key` were sufficient on their own, so unpublishing an agent left its embed key and every outstanding guest JWT fully functional. The fix adds an `AgentStatus.PUBLISHED` condition to all three agent-widget gate helpers — `ensure_widget_agent_available` (chat/files/WS), `_resolve_widget_agent_by_key` (direct key auth + embed-ticket minting), and `_get_live_widget_agent` (embed-ticket redemption) — and folds every denial into the pre-existing "Invalid widget key" / "Widget is disabled for this agent" 403s so publication state stays unprobeable. Tests cover the new gate at each of the three helpers plus the WS per-message revalidation path.

## Approach: sound

Gating in the helpers rather than clearing `widget_enabled`/`widget_key` inside `unpublish_agent` is the right lever, and for a reason the PR description undersells: a state-clearing approach structurally cannot cover a **never-published draft**. `add_agent` mints every new agent with `widget_enabled=True` and a live `widget_key` immediately, with `status` defaulting to `DRAFT` — a draft never passes through `unpublish_agent`, so its key would still serve. Re-deriving the gate per request closes that case for free. All five routes in `src/xagent/web/api/widget.py` are funnelled through one of the three helpers, so there is no fourth unguarded path. The "unknown key" vs "unpublished" collapse also holds up — same status, same detail string, same query shape.

One forward-looking design note, not a finding: for an already-connected WS guest, revocation only lands on the guest's next inbound message — an unpublished agent's socket keeps receiving outbound/broadcast frames and in-flight runs keep executing until the guest speaks. That is exactly the contract the existing disable/rotate levers already have, so this PR is consistent with prior behavior rather than introducing a new gap. Still, "owner clicks Unpublish" reads as *stop now* to a user, and the delta between that expectation and per-message revalidation is a reasonable follow-up issue for all three levers at once — out of scope here.

## Findings

### 1. [Low] `_get_live_widget_agent` is the only sibling gate that doesn't check `widget_key`
`src/xagent/web/api/widget.py:138`

Of the four sibling gate helpers, this is the only one that omits a non-null `widget_key` check — compare `_resolve_widget_agent_by_key` (widget.py:160), `ensure_widget_agent_available` (public_chat_access.py:189), and `_workforce_owner_from_ticket` (widget.py:287), all of which check it. The docstring at widget.py:123-124 frames "live" as exactly three conditions with the key deliberately excluded, so this reads as intentional rather than an oversight.

It's unreachable today: `add_agent` always mints a key when `widget_enabled=True`; the update path self-heals by minting one if the flag is toggled on without a key; rotation never nulls the key; legacy rows were backfilled. But there's no DB-level NOT-NULL-when-enabled constraint, so the invariant is latent, not structural. If a future path ever produces `widget_enabled=True, widget_key=None`, this helper mints a guest token that 403s on the very next real request instead of failing cleanly at ticket redemption.

Suggest adding `or not agent.widget_key` for consistency with the three siblings, or documenting why the key is intentionally excluded here.

### 2. [Low] ARCHIVED coverage is uneven across the three gated entry points
`tests/web/api/test_agents_management.py:1289`

`test_widget_auth_rejects_agent_that_is_not_published` parametrizes over `[AgentStatus.DRAFT, AgentStatus.ARCHIVED]` for the `_resolve_widget_agent_by_key` gate, reasoning that a DRAFT-only test wouldn't catch a narrowing bug such as `== AgentStatus.DRAFT` in place of `!= AgentStatus.PUBLISHED`. That argument applies equally to the other two gates, but neither gets an ARCHIVED variant — `_get_live_widget_agent` (widget.py:138) and `ensure_widget_agent_available` (public_chat_access.py:190) are only exercised through the real `/unpublish` endpoint, which hardcodes `AgentStatus.DRAFT` and can never produce ARCHIVED.

Kept at Low because `AgentStatus.ARCHIVED` is never assigned anywhere in current production code, so this is a dead path today.

### 3. [Low / informational] Unpublish also revokes public file access — undocumented and untested
`src/xagent/web/api/public_chat_access.py:190`

The new PUBLISHED check in `ensure_widget_agent_available` is also reached transitively from the public file preview/download/open routes, via `_validate_public_task_file_access` (files.py:624) → `get_public_chat_user` (files.py:653). So unpublishing now additionally cuts a widget guest off from that task's file attachments. Correct and fail-safe, but it's part of the shipped blast radius and is neither mentioned in the PR description nor covered by a test — no test in test_file_upload.py or test_admin_file_access.py uses a widget/guest token at all.

Suggest a one-line note in the PR description, or a small test asserting a widget guest's file download 403s after unpublish.

`tests/web/api/test_widget_ws_revalidation.py`

- `L181: yagni` — `Mapping[str, Callable[[], None]]` used for a closed 3-key set consumed at one call site. Only two builders populate it (:223, :263), both with the same hardcoded disable/rotate/unpublish keys, and there is exactly one consumer (:334). Replace with a plain `unpublish: Callable[[], None]` field alongside the existing ones and dispatch via `getattr(guest, revocation)()` or an if/elif.

Net: roughly line-count neutral, but drops an abstraction layer and two typing imports for a closed, single-consumer case.

## Testing

Ran in a clean checkout of the PR head:

- `pytest tests/web/api/test_widget_ws_revalidation.py` — 11 passed
- `pytest tests/web/api/test_agents_management.py` — 80 passed

Also ran a revert experiment, removing all three status checks one at a time and rerunning the suite: each of the three source lines is independently pinned by a failing test. No source line in this change is only accidentally covered.

## Verdict

Approve with minor suggestions. All three findings are Low — one defence-in-depth consistency nit, one dead-path test-coverage gap, one documentation/coverage note on a correct side effect. None blocks merge.

src/xagent/web/api/widget.py
codeacme17: Added `or not agent.widget_key` in 06b375d9, and updated the docstring to say "four conditions" plus a paragraph on why the key is checked.

Went with the check rather than documenting the omission because the failure mode you describe is the worse of the two: minting a guest token that dies one request later is strictly less debuggable than failing at redemption. The revert experiment makes the difference concrete — with the check removed, `/api/widget/auth` returns 200 and the token payload carries `"widget_key":null`:

```
E   AssertionError: {"access_token":"...eyJ...d2lkZ2V0X2tleSI6bnVsbCw...","agent_id":1,...}
E   assert 200 == 403
```

Pinned by `test_embed_ticket_rejects_widget_enabled_agent_with_no_widget_key`, which writes `widget_key = None` directly since (as you note) no code path produces the row.

tests/web/api/test_agents_management.py
codeacme17: Right — the argument does generalize, so both gates now get an ARCHIVED variant in 06b375d9:

- `test_widget_guest_token_rejected_when_agent_is_archived` (`ensure_widget_agent_available`)
- `test_embed_ticket_stops_authenticating_once_agent_is_archived` (`_get_live_widget_agent`)

Both use a new `_set_agent_status` helper that writes the status straight to the row, since as you say `/unpublish` hardcodes DRAFT and nothing else assigns ARCHIVED. I kept them ARCHIVED-only rather than parametrizing over both states — the DRAFT half is already covered through the real endpoint by the sibling unpublish tests, and each new test docstring cross-references that so the split is visible.

Confirmed the gap was real: narrowing both `!= AgentStatus.PUBLISHED` to `== AgentStatus.DRAFT` leaves the whole pre-existing suite green and fails exactly these two new tests.

src/xagent/web/api/public_chat_access.py
codeacme17: Did both. 06b375d9 adds `test_unpublishing_agent_revokes_widget_guest_access_to_task_files`, and the PR description now has a **Public file routes** subsection under "Behaviour change beyond the issue text" naming the `_validate_public_task_file_access` -> `get_public_chat_user` path.

The test drives the real chain rather than hand-building the task row: widget auth -> `/api/widget/chat/task/create` -> `/api/widget/files/upload` -> `GET /api/files/public/download/{file_id}?token=...`, asserting 200 with the right bytes before the unpublish and 403 `"Widget is unavailable"` after. Going through the real upload is what makes the pre-unpublish 200 meaningful — `_validate_public_task_file_access` returns early when the task carries no `guest_id`, so a hand-written row could have made the whole assertion vacuous.

Also confirmed there was nothing to fix behaviourally: the guest is cut off from the attachments of a channel that has stopped serving, which is the fail-safe direction.

tests/web/api/test_widget_ws_revalidation.py
codeacme17: Agreed — the "not every call site in between" rationale was written for a call-site count I never actually had. Reverted to plain fields in 06b375d9:

```python
    disable: Callable[[], None]
    rotate: Callable[[], None]
    unpublish: Callable[[], None]
```

and the single consumer dispatches by parametrize id:

```python
    revoke: Callable[[], None] = getattr(guest, revocation)
    revoke()
```

Went with `getattr` over if/elif so adding a fourth lever still only touches the dataclass, the two builders, and the parametrize list — which was the property I wanted from the mapping, without the mapping. Field names now have to match the parametrize ids, so there is a comment on the dataclass saying so. `Mapping` import dropped; net -1 line.

Issue #983, opened 2026-07-23
XA-2: widget.js — delegated-context session mode (top-level session ownership)


rogercloud:
## Dependencies
- Blocked by: XA-0 (hard — regression net merges first)
- Blocks: XA-3 (defines the postMessage protocol it consumes)
- Runtime (not build) dependency: deployment-provided session endpoints — mode is dormant unless the page carries the attribute; deploy order guarantees endpoints exist first.

## Summary
Add an opt-in mode to `widget.js`: when the script tag carries `data-encrypted-context` (a signed-then-encrypted delegated context grant minted by the embedding application's backend), the widget establishes a chat session against the deployment's session API instead of the guest flow. When the attribute is absent, behavior is unchanged.

### Exchange
```
POST /v1/external/chat/sessions
Content-Type: application/json          (fetch with credentials: 'omit'; no Authorization)

{"encrypted_context": "<the attribute value, verbatim>"}
```
Success **200**:
```json
{
  "session_token": "st_…",
  "session_token_expires_at": "2026-07-22T05:15:00Z",
  "reconnect_token": "rt_…",
  "session": {
    "absolute_expires_at": "2026-07-22T13:00:00Z",
    "agent": {
      "id": 42,
      "name": "…",
      "description": "…",
      "logo_url": "…",
      "suggested_prompts": ["…"]
    }
  }
}
```

### Reconnect
```
POST /v1/external/chat/sessions/reconnect
Content-Type: application/json

{"reconnect_token": "rt_…", "encrypted_context": "<page grant, attach whenever present>"}
```
Success **200**: identical shape to Exchange (a fresh token pair; the old reconnect token is void).

### Error envelope (all endpoints)
```json
{"error": {"code": "<code>", "reason": "<optional>", "message": "<human-readable>"}}
```

### Failure codes → widget behavior
| code | HTTP | class | widget behavior |
|---|---|---|---|
| `grant_malformed` | 400 | integrity | fail closed; suppress reconnect; unavailable state |
| `signature_invalid` | 401 | integrity | fail closed; suppress reconnect |
| `grant_expired` | 401 | stale | fall back to in-memory reconnect token (attach the stale grant) |
| `grant_already_used` | 401 | stale | same as above |
| `agent_not_granted` | 403 | authorization | terminal unavailable |
| `agent_not_available` | 409 | resource state | terminal unavailable |
| `widget_disabled` | 409 | resource state | terminal unavailable |
[...]

PR #1059, opened 2026-07-29
fix(widget): include session error reason in diagnostics

AlexLiu190625:
## Summary

Adds safe error-reason diagnostics to delegated Widget Session exchange and reconnect failures.

- Parses the documented error envelope through one shared helper.
- Logs `agent_not_granted/origin_not_allowed` as `[agent_not_granted/origin_not_allowed]` so Allowed Domains configuration failures are distinguishable from missing agent grants.
- Keeps retry decisions, terminal/degraded state, and iframe `postMessage` payloads code-only.
- Rejects inherited error fields and synthetic failure codes, so host-page prototype pollution cannot alter retry or terminal classification.
- Ignores malformed, non-string, inherited, unknown, and unregistered reasons without logging the remote error message.

## Why

The session API intentionally uses `agent_not_granted` for both missing grants and Origin × `allowed_domains` denials. Without the optional `origin_not_allowed` diagnostic, partner developers are directed toward agent registration even when the actual problem is the Allowed Domains configuration.

The reason catalog is fail-closed: new server reasons must be registered deliberately and the hosted `widget.js` redeployed before they can appear in browser diagnostics.

## Verification

- `npm run test:widget:coverage` — 705/705 tests passed across 31 files
- `public/widget.js` coverage — 97.23% statements/lines, 93.03% branches, 98.30% functions
- `npx eslint src/components/widget/widget-session.test.ts public/widget.js`
- `npx tsc --noEmit`
- `node --check public/widget.js`
- `pre-commit run --files frontend/public/widget.js frontend/src/components/widget/widget-session.test.ts`

Refs #983

### DIFF
--- a/frontend/public/widget.js
+++ b/frontend/public/widget.js
@@ -28,6 +28,7 @@
     identity_mismatch: true,
     rate_limited: true
   };
+  // Only diagnostic reasons defined by public issue #983 may be logged.
   var SESSION_ERROR_REASONS = {
     agent_not_granted: {
       origin_not_allowed: true
@@ -57,7 +58,7 @@
   }
 
   function isPlainSessionErrorObject(value) {
-    if (!value || Object.prototype.toString.call(value) !== '[object Object]') return false;
+    if (!value || typeof value !== 'object') return false;
     var prototype = Object.getPrototypeOf(value);
     return prototype === Object.prototype || prototype === null;
   }
@@ -69,9 +70,7 @@
   }
 
   function sessionErrorEnvelope(result) {
-    if (!isPlainSessionErrorObject(result) ||
-        !Object.prototype.hasOwnProperty.call(result, 'data') ||
-        !isPlainSessionErrorObject(result.data) ||
+    if (!isPlainSessionErrorObject(result.data) ||
         !Object.prototype.hasOwnProperty.call(result.data, 'error') ||
         !isPlainSessionErrorObject(result.data.error)) {
       return { code: null, reason: null };
@@ -754,7 +753,8 @@
     }
 
     function classifySessionFailure(result) {
-      if (result && result.syntheticCode) return { code: result.syntheticCode, reason: null };
+      var syntheticCode = ownSessionErrorString(result, 'syntheticCode');
+      if (syntheticCode) return { code: syntheticCode, reason: null };
       var envelope = sessionErrorEnvelope(result);
       if (isKnownSessionErrorCode(envelope.code)) return envelope;
       if (result.status >= 500) return { code: 'network_unavailable', reason: null };
--- a/frontend/public/widget.js
+++ b/frontend/public/widget.js
@@ -59,8 +59,9 @@
 
   function isPlainSessionErrorObject(value) {
     if (!value || typeof value !== 'object') return false;
+    // Prototype identity is not affected by host-page Symbol.toStringTag pollution.
     var prototype = Object.getPrototypeOf(value);
-    return prototype === Object.prototype || prototype === null;
+    return prototype === Object.prototype;
   }
 
   function ownSessionErrorString(value, key) {
--- a/frontend/public/widget.js
+++ b/frontend/public/widget.js
@@ -28,7 +28,7 @@
     identity_mismatch: true,
     rate_limited: true
   };
-  // Only diagnostic reasons defined by public issue #983 may be logged.
+  // Only diagnostic reasons registered for their session error code may be logged.
   var SESSION_ERROR_REASONS = {
     agent_not_granted: {
       origin_not_allowed: true
--- a/frontend/public/widget.js
+++ b/frontend/public/widget.js
@@ -28,6 +28,11 @@
     identity_mismatch: true,
     rate_limited: true
   };
+  var SESSION_ERROR_REASONS = {
+    agent_not_granted: {
+      origin_not_allowed: true
+    }
+  };
   function sessionAbortError() {
     return new DOMException('session request superseded', 'AbortError');
   }
@@ -51,9 +56,38 @@
     });
   }
 
-  function sessionErrorCode(result) {
-    var code = result && result.data && result.data.error && result.data.error.code;
-    return typeof code === 'string' ? code : null;
+  function isPlainSessionErrorObject(value) {
+    if (!value || Object.prototype.toString.call(value) !== '[object Object]') return false;
+    var prototype = Object.getPrototypeOf(value);
+    return prototype === Object.prototype || prototype === null;
+  }
+
+  function ownSessionErrorString(value, key) {
+    if (!Object.prototype.hasOwnProperty.call(value, key)) return null;
+    var candidate = value[key];
+    return typeof candidate === 'string' ? candidate : null;
+  }
+
+  function sessionErrorEnvelope(result) {
+    if (!isPlainSessionErrorObject(result) ||
+        !Object.prototype.hasOwnProperty.call(result, 'data') ||
+        !isPlainSessionErrorObject(result.data) ||
+        !Object.prototype.hasOwnProperty.call(result.data, 'error') ||
+        !isPlainSessionErrorObject(result.data.error)) {
+      return { code: null, reason: null };
+    }
+    var error = result.data.error;
+    var code = ownSessionErrorString(error, 'code');
+    var reason = ownSessionErrorString(error, 'reason');
+    var allowedReasons = code && Object.prototype.hasOwnProperty.call(SESSION_ERROR_REASONS, code)
+      ? SESSION_ERROR_REASONS[code]
+      : null;
+    return {
+      code: code,
+      reason: allowedReasons && reason && Object.prototype.hasOwnProperty.call(allowedReasons, reason)
+        ? reason
+        : null
+    };
   }
 
   function isKnownSessionErrorCode(code) {
@@ -140,7 +174,7 @@
       retry.attempts += 1;
       return makeRequest(requestTimeoutMs).then(function (result) {
         if (signal.aborted) throw sessionAbortError();
-        var code = sessionErrorCode(result);
+        var code = sessionErrorEnvelope(result).code;
         if (code === 'rate_limited') {
           if (retry.rateLimited >= policy.maxRateLimitRetries) return result;
           retry.rateLimited += 1;
@@ -386,9 +420,10 @@
     return 'g' + hash.toString(16);
   }
 
-  function logSession(level, text, code, status) {
+  function logSession(level, text, code, status, reason) {
     var suffix = status ? ' (HTTP ' + status + ')' : '';
-    console[level]('Xagent Widget: ' + text + ' [' + code + ']' + suffix + '.');
+    var diagnostic = reason ? code + '/' + reason : code;
+    console[level]('Xagent Widget: ' + text + ' [' + diagnostic + ']' + suffix + '.');
   }
 
   function createSessionMode(scriptTag, host) {
@@ -710,20 +745,20 @@
       flush();
     }
 
-    function recordFailure(code, status) {
+    function recordFailure(code, status, reason) {
       if (state.terminalCode) return;
       state.terminalCode = code;
       state.recoverableCode = null;
-      logSession('error', 'chat unavailable', code, status);
+      logSession('error', 'chat unavailable', code, status, reason);
       flush();
     }
 
     function classifySessionFailure(result) {
-      if (result && result.syntheticCode) return result.syntheticCode;
-      var code = sessionErrorCode(result);
-      if (isKnownSessionErrorCode(code)) return code;
-      if (result.status >= 500) return 'network_unavailable';
-      return 'unexpected_error';
+      if (result && result.syntheticCode) return { code: result.syntheticCode, reason: null };
+      var envelope = sessionErrorEnvelope(result);
+      if (isKnownSessionErrorCode(envelope.code)) return envelope;
+      if (result.status >= 500) return { code: 'network_unavailable', reason: null };
+      return { code: 'unexpected_error', reason: null };
     }
 
     function handleResult(result, phase) {
@@ -739,8 +774,8 @@
         return;
       }
 
-      var code = classifySessionFailure(result);
-      recordFailure(code, result.status);
+      var failure = classifySessionFailure(result);
+      recordFailure(failure.code, result.status, failure.reason);
     }
 
     function postJson(url, body, timeoutMs, operationSignal) {
--- a/frontend/src/components/widget/widget-session.test.ts
+++ b/frontend/src/components/widget/widget-session.test.ts
@@ -17,12 +17,7 @@ const RECONNECT_URL = `${HOST}/v1/external/chat/sessions/reconnect`
 
 const fetchMock = vi.fn()
 
-type SessionFailureTestSeam = {
-  parse: (result: unknown) => { code: string | null; reason: string | null }
-  classify: (result: unknown) => { code: string; reason: string | null }
-}
-
-function runWidget(attributes: Record<string, string>, exposeClassifier = false) {
+function runWidget(attributes: Record<string, string>) {
   const script = document.createElement("script")
   script.src = `${HOST}/widget.js`
   for (const [name, value] of Object.entries(attributes)) {
@@ -30,13 +25,7 @@ function runWidget(attributes: Record<string, string>, exposeClassifier = false)
   }
   document.body.appendChild(script)
   Object.defineProperty(document, "currentScript", { configurable: true, value: script })
-  const executable = exposeClassifier
-    ? widgetScript.replace(
-      "    return { attach: attach };",
-      "    window.__xagentWidgetTestSessionFailure = { parse: sessionErrorEnvelope, classify: classifySessionFailure };\n    return { attach: attach };",
-    )
-    : widgetScript
-  window.eval(`${executable}\n//# sourceURL=${widgetScriptUrl}`)
+  window.eval(`${widgetScript}\n//# sourceURL=${widgetScriptUrl}`)
   return script
 }
 
@@ -104,14 +93,6 @@ function directJsonResponse(status: number, data: unknown): Response {
     headers: new Headers(),
     json: () => Promise.resolve(data),
   } as Response
-}
-
-function exposedSessionFailureTestSeam(): SessionFailureTestSeam {
-  const seam = (window as unknown as {
-    __xagentWidgetTestSessionFailure?: SessionFailureTestSeam
-  }).__xagentWidgetTestSessionFailure
-  if (!seam) throw new Error("session failure test seam was not exposed")
-  return seam
 }
 
 function proxyWithInheritedValue(target: Record<string, unknown>, key: string, value: unknown) {
@@ -194,7 +175,6 @@ describe("widget session mode", () => {
     localStorage.clear()
     // The grant dedupe registry lives on window and survives between tests in a file.
     Reflect.deleteProperty(window as unknown as Record<string, unknown>, "__xagentWidgetGrants")
-    Reflect.deleteProperty(window as unknown as Record<string, unknown>, "__xagentWidgetTestSessionFailure")
     vi.stubGlobal("fetch", fetchMock)
     fetchMock.mockReset()
 
@@ -219,8 +199,6 @@ describe("widget session mode", () => {
       window.removeEventListener(type, listener)
     }
     windowListeners = []
-    Reflect.deleteProperty(window as unknown as Record<string, unknown>, "__xagentWidgetTestSessionFailure")
-
     if (currentScriptDescriptor) {
       Object.defineProperty(document, "currentScript", currentScriptDescriptor)
     } else {
@@ -818,6 +796,70 @@ describe("widget session mode", () => {
     })
   })
 
+  it("ignores an inherited synthetic session failure code", async () => {
+    const syntheticCodeDescriptor = Object.getOwnPropertyDescriptor(Object.prototype, "syntheticCode")
+    Object.defineProperty(Object.prototype, "syntheticCode", {
+      configurable: true,
+      value: "network_unavailable",
+    })
+    try {
+      const errorSpy = vi.spyOn(console, "error").mockImplementation(() => undefined)
+      fetchMock.mockResolvedValueOnce(jsonResponse(403, {
+        error: {
+          code: "agent_not_granted",
+          reason: "origin_not_allowed",
+        },
+      }))
+      runWidget({ "data-encrypted-context": GRANT })
+      const post = spyOnIframePostMessage()
+
+      await vi.waitFor(() => expect(errorSpy).toHaveBeenCalledWith(
+        "Xagent Widget: chat unavailable [agent_not_granted/origin_not_allowed] (HTTP 403).",
+      ))
+      fromIframe("ready")
+
+      expect(post.mock.calls[0][0]).toMatchObject({ type: "session_terminal", code: "agent_not_granted" })
+    } finally {
+      if (syntheticCodeDescriptor) {
+        Object.defineProperty(Object.prototype, "syntheticCode", syntheticCodeDescriptor)
+      } else {
+        Reflect.deleteProperty(Object.prototype, "syntheticCode")
+      }
+    }
+  })
+
+  it("keeps a registered diagnostic intact when Object.prototype has a toString tag", async () => {
+    const toStringTagDescriptor = Object.getOwnPropertyDescriptor(Object.prototype, Symbol.toStringTag)
+    Object.defineProperty(Object.prototype, Symbol.toStringTag, {
+      configurable: true,
+      value: "polluted",
+    })
+    try {
+      const errorSpy = vi.spyOn(console, "error").mockImplementation(() => undefined)
+      fetchMock.mockResolvedValueOnce(jsonResponse(403, {
+        error: {
+          code: "agent_not_granted",
+          reason: "origin_not_allowed",
+        },
+      }))
+      runWidget({ "data-encrypted-context": GRANT })
+      const post = spyOnIframePostMessage()
+
+      await vi.waitFor(() => expect(errorSpy).toHaveBeenCalledWith(
+        "Xagent Widget: chat unavailable [agent_not_granted/origin_not_allowed] (HTTP 403).",
+      ))
+      fromIframe("ready")
+
+      expect(post.mock.calls[0][0]).toMatchObject({ type: "session_terminal", code: "agent_not_granted" })
+    } finally {
+      if (toStringTagDescriptor) {
+        Object.defineProperty(Object.prototype, Symbol.toStringTag, toStringTagDescriptor)
+      } else {
+        Reflect.deleteProperty(Object.prototype, Symbol.toStringTag)
+      }
+    }
+  })
+
   it("keeps the existing exchange diagnostic unchanged when the reason is absent", async () => {
     const errorSpy = vi.spyOn(console, "error").mockImplementation(() => undefined)
     fetchMock.mockResolvedValueOnce(errorResponse(403, "agent_not_granted"))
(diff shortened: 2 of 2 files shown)


frontend/public/widget.js
```diff
@@ -56,9 +61,27 @@
...
+        typeof error.code !== 'string') return empty;
+    var code = error.code;
+    var reasons = Object.prototype.hasOwnProperty.call(SESSION_ERROR_REASONS, code)
+      ? SESSION_ERROR_REASONS[code]
+      : null;
+    var reason = Object.prototype.hasOwnProperty.call(error, 'reason') &&
+      typeof error.reason === 'string' && reasons &&
+      Object.prototype.hasOwnProperty.call(reasons, error.reason)
+      ? error.reason
+      : null;
```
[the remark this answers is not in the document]
AlexLiu190625: Thanks. `SESSION_ERROR_REASONS` is a code-owned constant inside the widget IIFE, not runtime configuration; its only assignment defines nested objects. Remote response data can select a key but cannot change the catalog value type. If this catalog ever becomes dynamic, its shape should be validated once at that configuration boundary rather than at each lookup, so the current contract is kept here.

frontend/public/widget.js
```diff
@@ -591,23 +615,23 @@
...
     }
 
     function classifySessionFailure(result) {
-      var code = sessionErrorCode(result);
-      if (isKnownSessionErrorCode(code)) return code;
-      if (result.status >= 500) return 'network_unavailable';
-      return 'unexpected_error';
+      var envelope = sessionErrorEnvelope(result);
+      if (isKnownSessionErrorCode(envelope.code)) return envelope;
+      if (result.status >= 500) return { code: 'network_unavailable', reason: null };
```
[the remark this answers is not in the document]
AlexLiu190625: `result` cannot be null on this path. `postJson()` resolves only a structured `{ ok, status, data, retryAfter }` object; fetch, timeout, abort, and body-stream failures reject instead. `withRetry()` also reads `result.status` before `handleResult()` is reached, so a local guard here would not provide the proposed protection. Any future change to the request-result contract should be validated at the shared `postJson`/`withRetry` boundary.

frontend/public/widget.js
```diff
@@ -627,8 +651,8 @@
         return;
       }
 
-      var code = classifySessionFailure(result);
-      recordFailure(code, result.status);
+      var failure = classifySessionFailure(result);
+      recordFailure(failure.code, result.status, failure.reason);
```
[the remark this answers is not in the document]
AlexLiu190625: The same fulfilled-result contract applies here. `handleResult()` has already read `result.ok`, and `classifySessionFailure()` has already read `result.status`, before this call. Adding `result &&` only at the final argument would not guard a null result and would duplicate a downstream check, so the shared request-result contract is kept.

rogercloud:
please resolve the conflicts

frontend/public/widget.js
```diff
@@ -28,6 +28,11 @@
     identity_mismatch: true,
     rate_limited: true
   };
+  var SESSION_ERROR_REASONS = {
```
rogercloud, line 32: **major:** `SESSION_ERROR_REASONS` hardcodes `origin_not_allowed` under `agent_not_granted` with no producer, spec, or documentation anywhere in this repo. A repo-wide search finds this literal only in `widget.js` and its test file — no backend code, contract doc, or spec references it, and the linked issue #983 doesn't mention this reason at all (it's introduced fresh by this PR).

Because the catalog fails closed, a server-side literal that differs even slightly (casing, hyphen vs. underscore, etc.) would silently degrade the diagnostic to logging nothing, and nothing in this repo could catch the mismatch.

Suggestion: land the reason literal in a spec doc (e.g. alongside the #983 failure-code table, or in the server-side contract) and reference it in a comment above the catalog. A cross-repo contract test would be ideal.

frontend/public/widget.js
```diff
@@ -51,9 +56,38 @@
...
+
+  function ownSessionErrorString(value, key) {
+    if (!Object.prototype.hasOwnProperty.call(value, key)) return null;
+    var candidate = value[key];
+    return typeof candidate === 'string' ? candidate : null;
+  }
+
+  function sessionErrorEnvelope(result) {
+    if (!isPlainSessionErrorObject(result) ||
+        !Object.prototype.hasOwnProperty.call(result, 'data') ||
```
rogercloud: **minor:** These two `result`-level clauses (`!isPlainSessionErrorObject(result)` and `!hasOwnProperty.call(result, 'data')`) are unreachable given the current code. Every producer of `result` reaching this function (`postJson()`'s object literal and `exhaustedRetryResult()`'s object literal) always yields a plain object with an own `data` key.

Suggestion: drop these two clauses and start the check at `isPlainSessionErrorObject(result.data)`.

frontend/public/widget.js
```diff
@@ -51,9 +56,38 @@
     });
   }
 
-  function sessionErrorCode(result) {
-    var code = result && result.data && result.data.error && result.data.error.code;
-    return typeof code === 'string' ? code : null;
+  function isPlainSessionErrorObject(value) {
+    if (!value || Object.prototype.toString.call(value) !== '[object Object]') return false;
```
rogercloud: **minor:** `Object.prototype.toString.call(value) !== '[object Object]'` is redundant with the following `Object.getPrototypeOf` check for every input `res.json()` (native `JSON.parse`) can actually produce — the two checks only diverge for class instances, which `JSON.parse` cannot produce.

Suggestion: drop the `toString` clause and keep only the prototype check.

frontend/public/widget.js
```diff
@@ -140,7 +174,7 @@
       retry.attempts += 1;
       return makeRequest(requestTimeoutMs).then(function (result) {
         if (signal.aborted) throw sessionAbortError();
-        var code = sessionErrorCode(result);
+        var code = sessionErrorEnvelope(result).code;
```
rogercloud, line 177: **minor (PR description accuracy):** Swapping the retry decision's error-code extraction to `sessionErrorEnvelope(result).code` is a real behavior change on the retry path, not just a logging change. Verified: under `Object.prototype` pollution (e.g. another script on the host page setting `Object.prototype.code = 'rate_limited'`), the old direct-property-access code would incorrectly honor the polluted code on an uncoded 500 and enter the rate-limited retry branch, while this new code correctly fails closed.

This is a good hardening improvement, but the PR description's "keeps retry decisions ... code-only" framing hides that the extraction mechanism itself changed. Suggestion: add a line to the PR description calling this out as an intentional side benefit.

frontend/public/widget.js
```diff
@@ -710,20 +745,20 @@
...
 
     function classifySessionFailure(result) {
-      if (result && result.syntheticCode) return result.syntheticCode;
-      var code = sessionErrorCode(result);
-      if (isKnownSessionErrorCode(code)) return code;
-      if (result.status >= 500) return 'network_unavailable';
-      return 'unexpected_error';
+      if (result && result.syntheticCode) return { code: result.syntheticCode, reason: null };
+      var envelope = sessionErrorEnvelope(result);
+      if (isKnownSessionErrorCode(envelope.code)) return envelope;
```
rogercloud, line 760: **minor (cosmetic):** Three of `classifySessionFailure`'s four return branches construct a fresh `{code, reason}` literal, while this one returns the shared `envelope` object directly. Purely a style inconsistency — `sessionErrorEnvelope`'s two return statements both produce exactly `{code, reason}` by construction, so there's no actual extra-field risk today.

Optional suggestion: `return { code: envelope.code, reason: envelope.reason };` for uniformity.

frontend/src/components/widget/widget-session.test.ts
```diff
@@ -755,6 +793,185 @@ describe("widget session mode", () => {
...
+    }, HOST)
+  })
+
+  it.each([
+    ["missing reason", () => ({ error: { code: "agent_not_granted", message: "sentinel" } }), "agent_not_granted"],
+    ["null reason", () => ({ error: { code: "agent_not_granted", reason: null, message: "sentinel" } }), "agent_not_granted"],
+    ["non-string reason", () => ({ error: { code: "agent_not_granted", reason: 42, message: "sentinel" } }), "agent_not_granted"],
+    ["unregistered reason", () => ({ error: { code: "agent_not_granted", reason: "future_reason", message: "sentinel" } }), "agent_not_granted"],
+    ["prototype-key reason", () => ({ error: { code: "agent_not_granted", reason: "toString", message: "sentinel" } }), "agent_not_granted"],
+    ["malformed error", () => ({ error: ["agent_not_granted", "origin_not_allowed"] }), "unexpected_error"],
```
rogercloud: **minor (test quality):** This "malformed error" row (`error: ["agent_not_granted", "origin_not_allowed"]`, an array) produces the same `unexpected_error` outcome under both the pre-PR parsing code and the new parsing code — it doesn't actually exercise any new behavior this PR added (both reject the array, just via different mechanisms: `.code` being `undefined` vs. the `[object Object]` tag check).

Suggestion: either keep it as an explicit regression guard with a comment noting that, or replace it with a case that only the new guard catches (e.g. a prototype-polluted `error` object).

frontend/src/components/widget/widget-session.test.ts
```diff
@@ -17,15 +17,26 @@ const RECONNECT_URL = `${HOST}/v1/external/chat/sessions/reconnect`
...
   const script = document.createElement("script")
   script.src = `${HOST}/widget.js`
   for (const [name, value] of Object.entries(attributes)) {
     script.setAttribute(name, value)
   }
   document.body.appendChild(script)
   Object.defineProperty(document, "currentScript", { configurable: true, value: script })
-  window.eval(`${widgetScript}\n//# sourceURL=${widgetScriptUrl}`)
+  const executable = exposeClassifier
+    ? widgetScript.replace(
```
rogercloud: **minor (test design):** This source-mutation "test seam" (string-replacing an exact literal snippet inside the loaded widget source) is a pattern unique to this file — there's no precedent elsewhere in the suite, including the sibling `widget-bootstrap.test.ts`. It's fragile to any reformatting of that exact line in `widget.js` (though breakage is loud via `exposedSessionFailureTestSeam()`'s explicit throw, not silent).

On scope: only 2 of the 8 seam-dependent test cases genuinely need direct-call access (the `result.data` inherited-proxy row, and the "synthetic failure precedence" direct `classify()` test) — both are structurally unreachable via the ordinary `directJsonResponse` fetch-mock path. The other ~6 (the `data.error`/`error.code`/`error.reason` proxy rows and the malformed-envelope rows) could plausibly be rewritten through the normal mock path.

Suggestion: keep the seam only for the 2 cases that require it; migrate the rest to `directJsonResponse`-based mocking to shrink the fragile surface.

rogercloud (commented):
## Summary

This PR extends the widget's session-error diagnostics by introducing a single shared envelope parser (`sessionErrorEnvelope`) and a fail-closed "reason catalog" (`SESSION_ERROR_REASONS`) that whitelists an optional `reason` sub-field under the `agent_not_granted` failure code (currently just `origin_not_allowed`). The parsed `reason` is used for console diagnostics only. A full data-flow trace confirms `reason` never reaches `state.terminalCode` / `state.recoverableCode`, so it structurally cannot leak into the `postMessage` frames sent to the iframe — the PR's "code-only postMessage" claim holds.

## Round 0 — approach verdict: acceptable-with-reservations

The overall shape is right: one shared parser replacing ad-hoc property reads, a whitelist that fails closed, and the guest-mode path diff stays zero. Reservations are about contract provenance and defensive over-scaffolding, not the design direction.

- **D1 (major, folds into the inline finding below):** No spec/producer for `origin_not_allowed` exists anywhere in the repo.
- **D2 (informational):** The fail-closed catalog means every future server reason needs a widget.js deploy — an accepted cost since widget.js is host-served, but worth stating explicitly in the PR description.
- **D3 (informational, ties to the retry-path finding below):** The own-property-only hardening is a real prototype-pollution defense improvement, but it also changes retry-decision behavior, which the PR description doesn't mention.

## Findings

Line-level findings are posted as inline comments below. Summary:

- **Major** — `origin_not_allowed` has no producer/spec/doc anywhere in this repo; a server literal mismatch would silently degrade to logging nothing, undetectable by any test here.
- **Minor** — dead/unreachable guard clauses on `result` itself in `sessionErrorEnvelope`; a redundant `toString` brand check; a test row that doesn't discriminate old vs. new behavior; a fragile source-mutation test seam (partially necessary); an undocumented retry-path behavior change; a cosmetic return-shape inconsistency in `classifySessionFailure`.
- **Informational** — decorative catalog boolean values (matches pre-existing style, no action needed).

## Simplification opportunities (net ~-30 to -40 lines possible)

The Proxy/inherited-prototype hardening on `result`/`result.data`/`result.data.error` guards a threat model unreachable via native `JSON.parse` (browser `res.json()` can never produce a Proxy or non-`Object.prototype` object). `isPlainSessionErrorObject`/`ownSessionErrorString` plus the matching test scaffologies (the `proxyWithInheritedValue` helper and its dependent tests, and the null-prototype-envelope acceptance test) could be dropped in favor of plain `typeof x === 'object' && x !== null` checks with no behavior change against real inputs.

**Exception — keep this part:** the `hasOwnProperty` check on the `SESSION_ERROR_REASONS` catalog lookup itself is legitimate and should stay. It's the one piece of hardening that guards genuinely attacker-influenced data (a host page polluting `Object.prototype`), not the widget's own literals.

## Re-review: prior gemini-code-assist findings

Three inline comments from `gemini-code-assist[bot]` were disputed by the author. All three were independently re-verified against the current code and the author's rebuttals are factually correct — **WAIVED, justified** in all three cases:

1. `hasOwnProperty` on the reasons catalog — `SESSION_ERROR_REASONS` is a static literal, never reassigned; no path makes it a primitive string.
2. Null-guard on `result.status` (line 760) — `postJson()` never resolves null (all failure paths reject); `withRetry()`'s `run()` already dereferences `result.status` earlier, so a null result would have thrown upstream first.
3. Null-guard before `recordFailure` (line 778) — `handleResult()` dereferences `result.ok` as its very first statement, before this call is reached.

No unresolved prior-review items remain.

## Testing performed

- `npx vitest run --config vitest.widget.config.ts src/components/widget/widget-session.test.ts` — **111/111 passed**
- `npx eslint src/components/widget/widget-session.test.ts public/widget.js` — clean
- `npx tsc --noEmit` — clean, exit 0
- Not independently re-verified: the full `npm run test:widget:coverage` (702/702 across 31 files) and coverage percentages quoted in the PR description — only the single changed test file was run in isolation here.

## Bottom line

**Approve, with the `origin_not_allowed` contract-documentation gap as a strongly recommended follow-up rather than a merge blocker.** The feature is diagnostics-only and fails closed: a wrong or missing reason literal degrades to logging nothing, never to a wrong decision or a leaked value. The retry-path PR-description update (see inline comment) is worth doing before merge since it's free. The dead-code/redundant-check/test-seam cleanup and the simplification pass are good candidates for a fast follow-up PR.

frontend/public/widget.js
AlexLiu190625: The public contract is defined in #983's July 29 contract amendment, which explicitly specifies `agent_not_granted` with `reason: origin_not_allowed` for Allowed Domains denials: https://github.com/xorbitsai/xagent/issues/983#issuecomment-5115705305. I added that public contract provenance above the fail-closed catalog. The server producer is outside this repository, so duplicating a producer or a cross-repository test here would create a second contract owner.

frontend/public/widget.js
AlexLiu190625: Removed the result-level clauses. `postJson()` and `exhaustedRetryResult()` own the fulfilled-result shape, so the envelope parser now begins at `result.data` instead of duplicating that producer contract.

frontend/public/widget.js
AlexLiu190625: Removed the `Object.prototype.toString` brand check and retained the prototype boundary. A production-path regression test now verifies that inherited `Symbol.toStringTag` cannot make a valid JSON error envelope fail classification.

frontend/public/widget.js
AlexLiu190625: Updated the PR description to call out the own-property hardening on retry and terminal classification. The shared classifier now also accepts `syntheticCode` only as an own string property, preventing inherited host-page values from overriding a server failure.

frontend/src/components/widget/widget-session.test.ts
AlexLiu190625: Kept this as an explicit malformed-envelope contract guard and moved it through native `Response.json()`. The matrix now covers root, error, code, and reason array shapes and verifies both the console classification and the code-only iframe frame.

frontend/src/components/widget/widget-session.test.ts
AlexLiu190625: Removed the source-mutation seam completely. Reachable inherited-property and malformed-envelope cases now run through the normal request, parser, classification, console, and iframe path; the unreachable direct parser/classifier cases were removed.

frontend/public/widget.js
AlexLiu190625: Keeping the direct return here. `sessionErrorEnvelope()` creates a fresh, fixed `{ code, reason }` value on every call, so there is no shared mutable state or extra-field exposure to normalize. Reconstructing the same shape in `classifySessionFailure()` would duplicate the envelope owner's mapping solely for stylistic consistency without changing the contract.

rogercloud (commented):
## Follow-up review — `fix(widget): simplify session error parsing` (head `65115e13`)

This is a follow-up pass checking the new commit against my prior review at `433ce2e7`. Overall this is a good round of cleanup: the parsing helpers are genuinely simpler, one change fixes a real (previously unnoticed) bug rather than merely restating the old logic, and the fragile test seam is gone with net-better coverage in its place. One prior finding remains open and has moved slightly in the wrong direction.

### Status of prior findings

| # | Finding | Status |
|---|---|---|
| F1 | `origin_not_allowed` has no producer or spec anywhere in the repo (major) | **NOT FIXED** — and slightly regressed; see inline comment |
| F2 | Dead `result`-level guard clauses in `sessionErrorEnvelope` | **FIXED** — checks now start at `result.data`, correct since `result` is always an internally-constructed literal |
| F3 | Redundant `Object.prototype.toString.call` brand check | **FIXED**, and better than asked — see note below |
| F4 | Non-discriminating "malformed error" test row | **FIXED** — replaced with a discriminating set (root array, missing/null `error`, `error` array, non-string `code` variants) |
| F5 | Fragile source-mutation test seam | **FIXED** — seam fully removed, dependents migrated to fetch mocking with iframe-payload assertions; one dropped case has no replacement (see inline comment on the null-prototype branch) |
| F6 | Decorative catalog boolean values (informational) | Unchanged — no fix expected |
| F7 | Retry-path hardening missing from PR description | **FIXED** — description now covers inherited fields / synthetic codes |
| F8 | `classifySessionFailure` returns a shared envelope in one branch, fresh literals elsewhere (cosmetic) | Unchanged — no fix expected; envelope is provably always `{code, reason}` by construction |
| D2 | PR description should note new reasons require a `widget.js` redeploy | **FIXED** — now stated in "Why" |

**On F3 — credit where it's due.** Replacing `Object.prototype.toString.call(...)` with `typeof value !== 'object'` + an explicit `Object.getPrototypeOf` check actually fixes a live bug: `toString.call` is spoofable through `Object.prototype[Symbol.toStringTag]`, so a single host-page script setting that tag would have made the old check reject *every* legitimate envelope, silently downgrading a real `agent_not_granted` / `origin_not_allowed` to `unexpected_error`. The new `Symbol.toStringTag` pollution test pins exactly this — confirmed it fails if the old check is restored. Good catch.

### Testing performed

Run against `65115e13`:

- `npx vitest run --config vitest.widget.config.ts src/components/widget/widget-session.test.ts` — **113/113 passed** (up from 111)
- `npx eslint src/components/widget/widget-session.test.ts public/widget.js` — clean
- `npx tsc --noEmit` — clean, exit 0
- Additionally: reverted the `toString.call` → `getPrototypeOf` change locally to confirm the new `Symbol.toStringTag` test genuinely fails without it, and wrote a small Node repro confirming the existing `hasOwnProperty` guards defeat direct `Object.prototype` pollution (relevant to an inline comment below).
- Not re-verified: the PR description's 704/704 across 31 files and the stated coverage percentages — only the changed test file was run in isolation. Same caveat as my prior review.

### Bottom line

Eight of the nine actionable prior findings are resolved, and F3 turned out to fix a real bug beyond what was asked. The code is in good shape and runtime behavior is safe in every case examined.

**Requesting one change before merge:** correct the `widget.js:31` comment so it doesn't attribute `origin_not_allowed` to issue #983 (see inline comment) — either cite the real authority or mark it as anticipatory with no producer yet. It's a one-line edit, but shipping a comment that misstates the feature's own contract compounds later.

The test-coverage items (inline comments below) I'd be happy to see land as a fast follow-up rather than blocking this PR. Once the `widget.js:31` comment is corrected, this is an approve from me.

frontend/public/widget.js
```diff
@@ -28,6 +28,12 @@
     identity_mismatch: true,
     rate_limited: true
   };
+  // Only diagnostic reasons defined by public issue #983 may be logged.
```
rogercloud: **major, not fixed from prior review:** This comment attributes `origin_not_allowed` to issue #983, but #983's failure-code table only lists `agent_not_granted` (403, terminal-unavailable) — it defines no reason sub-classification and never mentions `origin_not_allowed`. A repo-wide search still finds this literal only in `widget.js` and its test file; no backend producer, spec, or doc anywhere confirms the server actually emits it.

This comment now asserts a provenance that doesn't exist, which is worse than having no comment: a future maintainer will read it and believe the catalog is validated against a real contract, then extend it on that assumption. Runtime behavior stays safe either way (fail-closed — an unknown reason is simply not logged), but please correct the comment before merge. Either:
1. Cite the actual authority for `origin_not_allowed` (a backend PR, a design doc, a different issue), or
2. Rephrase to not attribute it to #983 — e.g. "Allow-list of diagnostic reasons; entries must correspond to a reason the backend actually emits. `origin_not_allowed` is anticipatory and has no producer yet."

frontend/public/widget.js
```diff
@@ -51,9 +57,36 @@
     });
   }
 
-  function sessionErrorCode(result) {
-    var code = result && result.data && result.data.error && result.data.error.code;
-    return typeof code === 'string' ? code : null;
+  function isPlainSessionErrorObject(value) {
+    if (!value || typeof value !== 'object') return false;
+    var prototype = Object.getPrototypeOf(value);
+    return prototype === Object.prototype || prototype === null;
```
rogercloud: **low, new test-coverage gap:** The `prototype === Object.prototype || prototype === null` branch accepts `Object.create(null)`-based envelope objects. The only test that exercised this exact case (`"accepts the registered diagnostic pair from null-prototype envelope objects"`) was deleted along with the test seam in this commit, and nothing replaced it — `Object.create(null)` now has zero hits in the test file. The branch is still reachable code but is completely untested.

Suggestion: either add back a small test building a null-prototype `error`/`data` object through the normal `jsonResponse` mock path, or drop `|| prototype === null` — native `JSON.parse` never produces null-prototype objects, so it's unclear the laxness buys anything real.

frontend/public/widget.js
```diff
@@ -51,9 +57,36 @@
...
+  }
+
+  function ownSessionErrorString(value, key) {
+    if (!Object.prototype.hasOwnProperty.call(value, key)) return null;
+    var candidate = value[key];
+    return typeof candidate === 'string' ? candidate : null;
+  }
+
+  function sessionErrorEnvelope(result) {
+    if (!isPlainSessionErrorObject(result.data) ||
```
rogercloud, line 74: **low, nit:** This now dereferences `result.data` without the previous `result &&` precondition (the original pre-PR `sessionErrorCode` had `result && result.data && ...`). Not a live crash path — both call sites always pass a `postJson`/`exhaustedRetryResult` object — but the invariant that `result` is always non-null is now implicit rather than documented. Optional: a one-line comment naming the precondition.

frontend/src/components/widget/widget-session.test.ts
```diff
@@ -755,6 +771,211 @@ describe("widget session mode", () => {
...
+      () => ({
+        error: proxyWithInheritedValue(
+          { code: "agent_not_granted" },
+          "reason",
+          "origin_not_allowed",
+        ),
+      }),
+      "agent_not_granted",
+    ],
+  ])("suppresses an inherited %s diagnostic component", async (_boundary, makeData, expectedCode) => {
```
rogercloud: **medium, test-realism / coverage gap:** This `it.each` block uses `proxyWithInheritedValue` routed through `directJsonResponse`, whose `json()` is just `Promise.resolve(data)` — it never goes through real serialization, so these three cases depend on `res.json()` returning a Proxy, which native `JSON.parse` can never produce. This tests a test-harness artifact, not a scenario reachable via a real browser `fetch`.

Meanwhile the file *does* use the real technique (`Object.defineProperty(Object.prototype, ...)`) for `syntheticCode` (line ~801) and `Symbol.toStringTag` (line ~833), but never applies it to `error`/`code`/`reason` — so the genuinely reachable threat (a host-page script polluting `Object.prototype.error`/`.code`/`.reason` directly) has zero coverage, even though the unreachable Proxy variant has three tests here.

To be clear on severity: **not exploitable** — I confirmed via a Node repro that the existing `hasOwnProperty.call(result.data, 'error')` guard (and its counterpart in `ownSessionErrorString`) already fully defends against direct `Object.prototype` pollution, since `hasOwnProperty` ignores the prototype chain. This is a test-quality issue only.

Suggestion: swap the three Proxy cases for `Object.defineProperty(Object.prototype, 'error' | 'code' | 'reason', {...})` around a plain `jsonResponse` mock, restoring descriptors in `finally` — matching what the file already does elsewhere. `proxyWithInheritedValue` can then go; `directJsonResponse` itself can stay for cases that need it (e.g. a raw 503).

frontend/src/components/widget/widget-session.test.ts
```diff
@@ -755,6 +771,211 @@ describe("widget session mode", () => {
...
+
+    expect(post.mock.calls[0][0]).toEqual({
+      xagent: true,
+      v: 1,
+      type: "session_terminal",
+      code: "agent_not_granted",
+    })
+  })
+
+  it("ignores an inherited synthetic session failure code", async () => {
```
rogercloud, line 816: **low, nit:** This asserts only that the correct code is eventually logged. Adding `expect(errorSpy).not.toHaveBeenCalledWith(expect.stringContaining("network_unavailable"))` would also catch a partial regression where both codes get logged.

frontend/src/components/widget/widget-session.test.ts
```diff
@@ -755,6 +771,211 @@ describe("widget session mode", () => {
...
+    } finally {
+      if (syntheticCodeDescriptor) {
+        Object.defineProperty(Object.prototype, "syntheticCode", syntheticCodeDescriptor)
+      } else {
+        Reflect.deleteProperty(Object.prototype, "syntheticCode")
+      }
+    }
+  })
+
+  it("keeps a registered diagnostic intact when Object.prototype has a toString tag", async () => {
```
rogercloud, line 848: **low, nit:** This is the most valuable new test here (it pins a real bug fix — see the note on F3 in the review body), but nothing explains why it exists. One line noting that reverting to `toString.call` would silently downgrade a real `agent_not_granted`/`origin_not_allowed` to `unexpected_error` under `Symbol.toStringTag` pollution would keep it from being "simplified" away later.

frontend/public/widget.js
AlexLiu190625: Fixed in 69fd8006. The parser now accepts only objects with Object.prototype, so null-prototype response objects fail closed as unexpected_error. Added a regression test for that boundary.

frontend/src/components/widget/widget-session.test.ts
AlexLiu190625: Fixed in 69fd8006. Replaced the Proxy/direct-response cases with Object.prototype pollution cases that pass through a real Response.json() path, covering inherited error, code, and reason fields with descriptor restoration.

frontend/src/components/widget/widget-session.test.ts
AlexLiu190625: Fixed in 69fd8006. The synthetic-code pollution regression now asserts the complete console.error call matrix, ensuring no network_unavailable diagnostic is emitted alongside the expected code.

frontend/src/components/widget/widget-session.test.ts
AlexLiu190625: Fixed in 69fd8006. Documented at the parser owner that prototype identity is intentional because host-page Symbol.toStringTag pollution must not affect valid JSON object classification.

frontend/public/widget.js
AlexLiu190625: `sessionErrorEnvelope` is private and both call paths receive only an object produced by `postJson` or `exhaustedRetryResult`. `postJson` rejects fetch, abort, timeout, and body-stream failures, while every fulfilled response is mapped to `{ ok, status, data, retryAfter }`; exhaustion produces `{ ok, status, data, syntheticCode }`. `handleResult` also reads `result.ok` before classification. A null guard or local precondition comment here would duplicate the request-boundary contract rather than protect a reachable path, so this is retained unchanged.

frontend/public/widget.js
AlexLiu190625: The source is the public contract amendment on #983: https://github.com/xorbitsai/xagent/issues/983#issuecomment-5114685343. Section 3 explicitly defines the first diagnostic case as `code: agent_not_granted` with `reason: origin_not_allowed`. The producer is outside this frontend repository, but the catalog entry is not anticipatory and the existing contract provenance is accurate, so the comment is retained unchanged.

frontend/public/widget.js
```diff
@@ -28,6 +28,12 @@
     identity_mismatch: true,
     rate_limited: true
   };
+  // Only diagnostic reasons defined by public issue #983 may be logged.
+  var SESSION_ERROR_REASONS = {
+    agent_not_granted: {
+      origin_not_allowed: true
+    }
+  };
```
rogercloud, line 36: **medium, suggested follow-up (non-blocking):** This catalog nests reasons per-code (`origin_not_allowed` is only valid under `agent_not_granted`), but no test verifies that binding is enforced. Confirmed via mutation testing: flattening this to an unconditional allowlist (checking `reason` regardless of `code`) still left all 114 tests passing.

Suggestion: add one test pairing `reason: "origin_not_allowed"` with a different **known terminal** code (e.g. `session_expired`, `widget_disabled`, or `identity_mismatch` — not `rate_limited` or an unknown/5xx code, since those exit through paths that never read `reason`), asserting the console line becomes `[session_expired]` with the reason correctly suppressed.

frontend/public/widget.js
rogercloud: **nit, optional:** Now that the amendment on #983 (the comment defining `origin_not_allowed`) is confirmed genuine and pre-existing, this is no longer misleading in substance — just imprecise in wording. Citing the specific amendment rather than "#983" generically would save a future reader who checks only the issue body (where this reason doesn't appear) a few minutes of confusion.

rogercloud (approved):
## Round 3 Review — `fix(widget): tighten session error object handling` (head `69fd8006`)

Third review round. Re-checked this commit against round 2's findings, plus a fresh full pass over the change as it now stands.

**Short version: this round resolves essentially everything outstanding, and the one finding that was blocking in round 2 turns out to have real spec backing that I had missed. One new, concrete test-coverage gap remains, noted below as a non-blocking suggestion.**

### Status of prior findings

| Finding | Round 2 severity | Status this round |
| --- | --- | --- |
| F1 — comment attributes `origin_not_allowed` to issue #983 | major | **Substance resolved / waived-justified** — see note below |
| NEW-1 — Proxy-based inherited-property tests exercised an unreachable scenario | medium | **Fixed** — replaced with real `Object.prototype` pollution tests through a real `jsonResponse` |
| NEW-2 — null-prototype acceptance branch untested | low | **Fixed** — the `|| prototype === null` branch was removed entirely; a new test covers the fail-closed behavior |
| NEW-3 — lost `result &&` precondition | low (nit) | **Waived — justified**, independently verified unreachable |
| NEW-4 — inherited-synthetic-code test only asserted the positive case | low (nit) | **Fixed** — now an exact-match assertion on the full call list |
| NEW-5 — `Symbol.toStringTag` test lacked an explanatory comment | low (nit) | **Fixed** |

**On F1 — correction to my earlier reading.** My round-1/2 position was that `origin_not_allowed` had no spec provenance. That was wrong. The author's cited issue #983 comment (id `5114685343`, `2026-07-29T07:49:10Z`) — confirmed via the GitHub API to have been posted a full day *before* my first review (`2026-07-30T09:14:24Z`) — is a standalone, pre-existing contract amendment, not a reactive citation. The reason genuinely is specified, and the implementation matches it byte-for-byte. This is no longer a blocker; only the code comment's wording (citing "#983" generically rather than the specific amendment) remains as an optional precision nit.

### New finding this round (non-blocking, suggested follow-up)

**NEW-6 (medium, test coverage) — the reason catalog's per-code scoping is untested.** `SESSION_ERROR_REASONS` nests reasons under their specific code (`{ agent_not_granted: { origin_not_allowed: true } }`), but no test verifies this binding is enforced. Confirmed via live mutation testing: flattening the catalog to an unconditional check left all 114 tests passing. See inline comment for the suggested test case. Happy for this to land as a fast follow-up rather than block this PR.

**Minor/nits (no action needed):** the `Symbol.toStringTag` test can't currently fail against this implementation (it guards a hypothetical future regression, which is a defensible reason to keep it); the 503 test's use of `directJsonResponse` could use a one-line comment explaining why; one unrelated blank-line diff in `afterEach`.

### Testing performed (at `69fd8006`)

- `npx vitest run --config vitest.widget.config.ts src/components/widget/widget-session.test.ts` — **114/114 passed**
- `npx eslint src/components/widget/widget-session.test.ts public/widget.js` — clean
- `npx tsc --noEmit` — clean, exit 0
- Live mutation testing on the reason-catalog scoping (see NEW-6), reverted afterward

### Bottom line

**Approving.** This round replaced tests of an impossible scenario with tests of the actual threat (NEW-1), removed a risky branch rather than just testing around it (NEW-2), and the F1 blocker dissolves once the amendment's timeline is accounted for — my error, not the author's. NEW-6 is worth picking up as a quick follow-up but shouldn't hold this up.

frontend/public/widget.js
AlexLiu190625: Fixed in d953b3ae. Replaced the generic issue reference with a stable contract-oriented invariant: only diagnostic reasons registered for their session error code may be logged. The implementation comment now explains the enforced behavior directly, while the PR and issue retain the provenance.

frontend/public/widget.js
AlexLiu190625: Fixed in d953b3ae. Added a regression case pairing `session_expired` with `origin_not_allowed`; it asserts the exact code-only console diagnostic and iframe terminal frame. An unconditional reason allowlist makes the test fail, while the current per-code lookup passes.

PR #1062, opened 2026-07-30
ref: add Home and Build extension seams

AlexLiu190625:
## Summary

- add neutral component contracts for Home and Build page extensions
- render Home and per-Agent Build extensions through real React component boundaries
- keep a stable Build page Provider mounted across loading, refresh, empty, and populated states
- demonstrate one Provider-owned shared load across multiple `agentId`-only card extensions
- keep general Build page behavior tests separate from extension-boundary tests
- run Home, Build page, and Build extension suites in focused frontend CI

## Why

Embedding distributions need to add small page-specific UI without copying and maintaining complete route implementations. These no-op-by-default seams keep the canonical Home and Build pages owned by Xagent while allowing replacement modules to supply components that can safely use React hooks.

The Build contract separates a page-level Provider from an `agentId`-only card component. The Provider and card exports form one replaceable module pair: the Provider can load shared page data once, while each card joins that data by the stable Agent ID without coupling the extension contract to the page-local Agent model.

## Validation

- `npm run test:home-build-pages` — 3 files, 14 tests passed
- `npm run type-check`
- targeted ESLint — 0 errors
- `pre-commit run --files ...`
- mutations confirm the focused scope fails when the canonical Build suite is removed, card/value joins are mismatched, or the deletion suite loses its default ASR input
- `git diff --check 4b42e977...2e4c00ff`

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,7 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
-    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page.test.tsx src/app/build/page-extension.test.tsx",
+    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/build/ src/app/page.test.tsx",
     "test:widget": "vitest run --config vitest.widget.config.ts",
     "test:widget:coverage": "vitest run --config vitest.widget.config.ts --coverage"
   },
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,7 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
-    "test:app-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page.test.tsx",
+    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page-extension.test.tsx",
     "test:widget": "vitest run --config vitest.widget.config.ts",
     "test:widget:coverage": "vitest run --config vitest.widget.config.ts --coverage"
   },
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,7 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
-    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page-extension.test.tsx",
+    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page.test.tsx src/app/build/page-extension.test.tsx",
     "test:widget": "vitest run --config vitest.widget.config.ts",
     "test:widget:coverage": "vitest run --config vitest.widget.config.ts --coverage"
   },
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,6 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
+    "test:app-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page.test.tsx",
     "test:widget": "vitest run --config vitest.widget.config.ts",
     "test:widget:coverage": "vitest run --config vitest.widget.config.ts --coverage"
   },
--- a/frontend/src/app/build/page-extension.test.tsx
+++ b/frontend/src/app/build/page-extension.test.tsx
@@ -6,6 +6,8 @@ const apiRequestMock = vi.hoisted(() => vi.fn())
 const routerPushMock = vi.hoisted(() => vi.fn())
 const routerReplaceMock = vi.hoisted(() => vi.fn())
 const cardRenderMock = vi.hoisted(() => vi.fn())
+const cardPropsMock = vi.hoisted(() => vi.fn())
+const providerLoaderMock = vi.hoisted(() => vi.fn())
 const providerLifetime = vi.hoisted(() => ({ mounts: 0, unmounts: 0 }))
 
 vi.mock("@/lib/api-wrapper", async () => {
@@ -64,28 +66,57 @@ vi.mock("@/components/voice-input-controller", () => ({
 
 vi.mock("@/lib/build-page-extension", async () => {
   const ReactModule = await vi.importActual<typeof import("react")>("react")
+  const missingProvider = Symbol("missing Build page extension Provider")
+  const BuildPageExtensionContext = ReactModule.createContext<
+    Record<string, string> | null | typeof missingProvider
+  >(missingProvider)
   const BuildPageExtensionProvider = ReactModule.memo(
     ({ children }: { children: React.ReactNode }) => {
+      const [sharedData, setSharedData] = ReactModule.useState<
+        Record<string, string> | null
+      >(null)
       ReactModule.useEffect(() => {
+        let active = true
         providerLifetime.mounts += 1
+        void providerLoaderMock().then((data: Record<string, string>) => {
+          if (active) {
+            setSharedData(data)
+          }
+        })
         return () => {
+          active = false
           providerLifetime.unmounts += 1
         }
       }, [])
       return ReactModule.createElement(
         "div",
         { "data-testid": "build-extension-provider" },
-        children,
+        ReactModule.createElement(
+          BuildPageExtensionContext.Provider,
+          { value: sharedData },
+          children,
+        ),
       )
     },
   )
   const BuildAgentCardExtension = ReactModule.memo(
-    ({ agentId }: { agentId: number }) => {
+    (props: { agentId: number }) => {
       ReactModule.useState(null)
+      const sharedData = ReactModule.useContext(BuildPageExtensionContext)
+      if (sharedData === missingProvider) {
+        throw new Error("Build Agent card extension requires its Provider")
+      }
+      const { agentId } = props
       cardRenderMock(agentId)
-      return ReactModule.createElement("div", {
-        "data-testid": `agent-card-supplement-${agentId}`,
-      })
+      cardPropsMock(props)
+      const supplement = sharedData?.[String(agentId)]
+      return supplement
+        ? ReactModule.createElement(
+          "div",
+          { "data-testid": `agent-card-supplement-${agentId}` },
+          supplement,
+        )
+        : null
     },
   )
   return {
@@ -151,6 +182,9 @@ describe("BuildsPage extension boundaries", () => {
     routerPushMock.mockReset()
     routerReplaceMock.mockReset()
     cardRenderMock.mockClear()
+    cardPropsMock.mockClear()
+    providerLoaderMock.mockReset()
+    providerLoaderMock.mockResolvedValue({ "42": "supplement-42" })
     providerLifetime.mounts = 0
     providerLifetime.unmounts = 0
   })
@@ -163,8 +197,50 @@ describe("BuildsPage extension boundaries", () => {
     render(<BuildsPage />)
 
     await screen.findByText("Research Agent")
-    expect(screen.getByTestId("agent-card-supplement-42")).toBeInTheDocument()
+    expect(await screen.findByTestId("agent-card-supplement-42"))
+      .toBeInTheDocument()
     expect(cardRenderMock).toHaveBeenCalledWith(42)
+    expect(cardPropsMock).toHaveBeenCalledWith({ agentId: 42 })
+  })
+
+  it("shares one provider load across multiple Agent card extensions", async () => {
+    const agents = [1, 2, 3].map((id) => ({
+      ...agent,
+      id,
+      name: `Agent ${id}`,
+    }))
+    const providerLoad = deferred<Record<string, string>>()
+    providerLoaderMock.mockReturnValue(providerLoad.promise)
+    apiRequestMock.mockResolvedValue(jsonResponse(agents))
+
+    render(<BuildsPage />)
+    await screen.findByText("Agent 3")
+
+    for (const id of [1, 2, 3]) {
+      expect(screen.queryByTestId(
+        `agent-card-supplement-${id}`,
+      )).not.toBeInTheDocument()
+    }
+    expect(providerLoaderMock).toHaveBeenCalledTimes(1)
+
+    await act(async () => {
+      providerLoad.resolve({
+        "1": "shared-value-1",
+        "2": "shared-value-2",
+        "3": "shared-value-3",
+      })
+      await providerLoad.promise
+    })
+
+    for (const id of [1, 2, 3]) {
+      expect(await screen.findByText(`shared-value-${id}`)).toBeInTheDocument()
+    }
+    expect(providerLoaderMock).toHaveBeenCalledTimes(1)
+    expect(new Set(cardPropsMock.mock.calls.map(([props]) => props.agentId)))
+      .toEqual(new Set([1, 2, 3]))
+    for (const [props] of cardPropsMock.mock.calls) {
+      expect(props).toEqual({ agentId: props.agentId })
+    }
   })
 
   it("keeps the page provider mounted across loading and a real publish refresh", async () => {
@@ -211,73 +287,10 @@ describe("BuildsPage extension boundaries", () => {
     })
     await screen.findByText("builds.emptyState.title")
     expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+    expect(providerLoaderMock).toHaveBeenCalledTimes(1)
 
     view.unmount()
     expect(providerLifetime).toEqual({ mounts: 1, unmounts: 1 })
   })
 
-  it("renders voice input in the create dialog", async () => {
-    apiRequestMock.mockResolvedValue(jsonResponse([]))
-
-    render(<BuildsPage />)
-    await waitFor(() => {
-      expect(screen.queryByText("common.loading")).not.toBeInTheDocument()
-    })
-    fireEvent.click(screen.getByRole("button", {
-      name: "builds.list.header.create",
-    }))
-
-    expect(screen.getByRole("button", {
-      name: "voiceInput.start",
-    })).toBeInTheDocument()
-  })
-
-  it("renders privileged actions for an editable Agent", async () => {
-    apiRequestMock.mockResolvedValue(jsonResponse([agent]))
-
-    render(<BuildsPage />)
-    await screen.findByText("Research Agent")
-
-    for (const name of [
-      "builds.list.actions.apiKey",
-      "builds.list.actions.triggers",
-      "builds.list.actions.publish",
-      "builds.list.actions.delete",
-      "builds.list.actions.edit",
-    ]) {
-      expect(screen.getByRole("button", { name })).toBeInTheDocument()
-    }
-    expect(screen.queryByRole("button", {
-      name: "builds.list.actions.viewConfig",
-    })).not.toBeInTheDocument()
-  })
-
-  it("limits a published read-only Agent to run and view actions", async () => {
-    apiRequestMock.mockResolvedValue(jsonResponse([{
-      ...agent,
-      status: "published",
-      can_edit: false,
-      can_publish: false,
-      can_delete: false,
-    }]))
-
-    render(<BuildsPage />)
-    await screen.findByText("Research Agent")
-
-    expect(screen.getByRole("button", {
-      name: "builds.list.actions.chat",
-    })).toBeInTheDocument()
-    expect(screen.getByRole("button", {
-      name: "builds.list.actions.viewConfig",
-    })).toBeInTheDocument()
-    for (const name of [
-      "builds.list.actions.apiKey",
-      "builds.list.actions.triggers",
-      "builds.list.actions.publish",
-      "builds.list.actions.delete",
-      "builds.list.actions.edit",
-    ]) {
-      expect(screen.queryByRole("button", { name })).not.toBeInTheDocument()
-    }
-  })
 })
--- a/frontend/src/app/build/page-extension.test.tsx
+++ b/frontend/src/app/build/page-extension.test.tsx
@@ -1,0 +1,283 @@
+import React from "react"
+import { act, cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+
+const apiRequestMock = vi.hoisted(() => vi.fn())
+const routerPushMock = vi.hoisted(() => vi.fn())
+const routerReplaceMock = vi.hoisted(() => vi.fn())
+const cardRenderMock = vi.hoisted(() => vi.fn())
+const providerLifetime = vi.hoisted(() => ({ mounts: 0, unmounts: 0 }))
+
+vi.mock("@/lib/api-wrapper", async () => {
+  const actual = await vi.importActual<typeof import("@/lib/api-wrapper")>(
+    "@/lib/api-wrapper",
+  )
+  return { ...actual, apiRequest: apiRequestMock }
+})
+
+vi.mock("@/lib/utils", async () => {
+  const actual = await vi.importActual<typeof import("@/lib/utils")>("@/lib/utils")
+  return { ...actual, getApiUrl: () => "http://api.local" }
+})
+
+vi.mock("next/navigation", () => ({
+  useRouter: () => ({ push: routerPushMock, replace: routerReplaceMock }),
+  useSearchParams: () => ({ get: () => null }),
+}))
+
+vi.mock("next/link", () => ({
+  default: ({
+    children,
+    href,
+    ...props
+  }: React.AnchorHTMLAttributes<HTMLAnchorElement> & { href: string }) => (
+    <a href={href} {...props}>{children}</a>
+  ),
+}))
+
+vi.mock("@/contexts/i18n-context", () => ({
+  useI18n: () => ({
+    t: (key: string) => key,
+  }),
+}))
+
+vi.mock("@/contexts/app-context-chat", () => ({
+  useApp: () => ({
+    dispatch: vi.fn(),
+    setTaskId: vi.fn(),
+    setPendingMessage: vi.fn(),
+  }),
+}))
+
+vi.mock("@/lib/branding", () => ({
+  getBrandingFromEnv: () => ({ appName: "Xagent" }),
+}))
+
+vi.mock("@/components/voice-input-controller", () => ({
+  useVoiceInputControls: () => ({
+    status: "idle",
+    hasAsrModel: true,
+    startRecording: vi.fn(),
+    stopRecording: vi.fn(),
+  }),
+}))
+
+vi.mock("@/lib/build-page-extension", async () => {
+  const ReactModule = await vi.importActual<typeof import("react")>("react")
+  const BuildPageExtensionProvider = ReactModule.memo(
+    ({ children }: { children: React.ReactNode }) => {
+      ReactModule.useEffect(() => {
+        providerLifetime.mounts += 1
+        return () => {
+          providerLifetime.unmounts += 1
+        }
+      }, [])
+      return ReactModule.createElement(
+        "div",
+        { "data-testid": "build-extension-provider" },
+        children,
+      )
+    },
+  )
+  const BuildAgentCardExtension = ReactModule.memo(
+    ({ agentId }: { agentId: number }) => {
+      ReactModule.useState(null)
+      cardRenderMock(agentId)
+      return ReactModule.createElement("div", {
+        "data-testid": `agent-card-supplement-${agentId}`,
+      })
+    },
+  )
+  return {
+    BuildPageExtensionProvider,
+    BuildAgentCardExtension,
+  }
+})
+
+vi.mock("@/components/build/deploy-agent-dialog", () => ({
+  DeployAgentDialog: () => null,
+}))
+
+vi.mock("@/components/build/agent-triggers-dialog", () => ({
+  AgentTriggersDialog: () => null,
+}))
+
+vi.mock("@/components/ui/popover", () => ({
+  Popover: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+  PopoverTrigger: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  PopoverContent: ({ children }: { children: React.ReactNode }) => <div>{children}</div>,
+}))
+
+vi.mock("@/components/ui/sonner", () => ({
+  toast: { error: vi.fn(), success: vi.fn() },
+}))
+
+import BuildsPage from "./page"
+
+const agent = {
+  id: 42,
+  name: "Research Agent",
+  description: "Researches launch topics",
+  logo_url: null,
+  status: "draft",
+  created_at: "2026-07-01T00:00:00Z",
+  updated_at: "2026-07-02T00:00:00Z",
+  widget_enabled: false,
+  allowed_domains: [],
+  can_edit: true,
+  can_publish: true,
+  can_delete: true,
+}
+
+function jsonResponse(data: unknown, init?: ResponseInit) {
+  return new Response(JSON.stringify(data), {
+    status: 200,
+    headers: { "Content-Type": "application/json" },
+    ...init,
+  })
+}
+
+function deferred<T>() {
+  let resolve!: (value: T | PromiseLike<T>) => void
+  const promise = new Promise<T>((resolvePromise) => {
+    resolve = resolvePromise
+  })
+  return { promise, resolve }
+}
+
+describe("BuildsPage extension boundaries", () => {
+  beforeEach(() => {
+    apiRequestMock.mockReset()
+    routerPushMock.mockReset()
+    routerReplaceMock.mockReset()
+    cardRenderMock.mockClear()
+    providerLifetime.mounts = 0
+    providerLifetime.unmounts = 0
+  })
+
+  afterEach(() => cleanup())
+
+  it("renders a hook-bearing supplement for a real Agent card", async () => {
+    apiRequestMock.mockResolvedValue(jsonResponse([agent]))
+
+    render(<BuildsPage />)
+
+    await screen.findByText("Research Agent")
+    expect(screen.getByTestId("agent-card-supplement-42")).toBeInTheDocument()
+    expect(cardRenderMock).toHaveBeenCalledWith(42)
+  })
+
+  it("keeps the page provider mounted across loading and a real publish refresh", async () => {
+    const initialAgents = deferred<Response>()
+    const refreshedAgents = deferred<Response>()
+    let agentListRequests = 0
+
+    apiRequestMock.mockImplementation((url: string, options?: RequestInit) => {
+      if (url === "http://api.local/api/agents" && !options?.method) {
+        agentListRequests += 1
+        return agentListRequests === 1
+          ? initialAgents.promise
+          : refreshedAgents.promise
+      }
+      if (
+        url === "http://api.local/api/agents/42/publish" &&
+        options?.method === "POST"
+      ) {
+        return Promise.resolve(new Response(null, { status: 204 }))
+      }
+      throw new Error(`Unhandled apiRequest: ${url}`)
+    })
+
+    const view = render(<BuildsPage />)
+    expect(screen.getByTestId("build-extension-provider")).toBeInTheDocument()
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+
+    await act(async () => {
+      initialAgents.resolve(jsonResponse([agent]))
+      await initialAgents.promise
+    })
+    await screen.findByText("Research Agent")
+
+    fireEvent.click(screen.getByRole("button", {
+      name: "builds.list.actions.publish",
+    }))
+    await waitFor(() => expect(agentListRequests).toBe(2))
+    expect(screen.getByText("common.loading")).toBeInTheDocument()
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+
+    await act(async () => {
+      refreshedAgents.resolve(jsonResponse([]))
+      await refreshedAgents.promise
+    })
+    await screen.findByText("builds.emptyState.title")
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+
+    view.unmount()
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 1 })
+  })
+
+  it("renders voice input in the create dialog", async () => {
+    apiRequestMock.mockResolvedValue(jsonResponse([]))
+
+    render(<BuildsPage />)
+    await waitFor(() => {
+      expect(screen.queryByText("common.loading")).not.toBeInTheDocument()
+    })
+    fireEvent.click(screen.getByRole("button", {
+      name: "builds.list.header.create",
+    }))
+
+    expect(screen.getByRole("button", {
+      name: "voiceInput.start",
+    })).toBeInTheDocument()
+  })
+
+  it("renders privileged actions for an editable Agent", async () => {
+    apiRequestMock.mockResolvedValue(jsonResponse([agent]))
+
+    render(<BuildsPage />)
+    await screen.findByText("Research Agent")
+
+    for (const name of [
+      "builds.list.actions.apiKey",
+      "builds.list.actions.triggers",
+      "builds.list.actions.publish",
+      "builds.list.actions.delete",
+      "builds.list.actions.edit",
+    ]) {
+      expect(screen.getByRole("button", { name })).toBeInTheDocument()
+    }
+    expect(screen.queryByRole("button", {
+      name: "builds.list.actions.viewConfig",
+    })).not.toBeInTheDocument()
+  })
+
+  it("limits a published read-only Agent to run and view actions", async () => {
+    apiRequestMock.mockResolvedValue(jsonResponse([{
+      ...agent,
+      status: "published",
+      can_edit: false,
+      can_publish: false,
+      can_delete: false,
+    }]))
+
+    render(<BuildsPage />)
+    await screen.findByText("Research Agent")
+
+    expect(screen.getByRole("button", {
+      name: "builds.list.actions.chat",
+    })).toBeInTheDocument()
+    expect(screen.getByRole("button", {
+      name: "builds.list.actions.viewConfig",
+    })).toBeInTheDocument()
+    for (const name of [
+      "builds.list.actions.apiKey",
+      "builds.list.actions.triggers",
+      "builds.list.actions.publish",
+      "builds.list.actions.delete",
+      "builds.list.actions.edit",
+    ]) {
+      expect(screen.queryByRole("button", { name })).not.toBeInTheDocument()
+    }
+  })
+})
--- a/frontend/src/app/build/page-extension.test.tsx
+++ b/frontend/src/app/build/page-extension.test.tsx
@@ -332,10 +332,8 @@ describe("BuildsPage extension boundaries", () => {
     })
     await screen.findByText("builds.emptyState.title")
     expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
-    expect(providerLoaderMock).toHaveBeenCalledTimes(1)
 
     view.unmount()
     expect(providerLifetime).toEqual({ mounts: 1, unmounts: 1 })
   })
-
 })
--- a/frontend/src/app/build/page-extension.test.tsx
+++ b/frontend/src/app/build/page-extension.test.tsx
@@ -1,5 +1,13 @@
 import React from "react"
-import { act, cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
+import {
+  act,
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 
 const apiRequestMock = vi.hoisted(() => vi.fn())
@@ -233,7 +241,11 @@ describe("BuildsPage extension boundaries", () => {
     })
 
     for (const id of [1, 2, 3]) {
-      expect(await screen.findByText(`shared-value-${id}`)).toBeInTheDocument()
+      const supplement = await screen.findByTestId(
+        `agent-card-supplement-${id}`,
+      )
+      expect(within(supplement).getByText(`shared-value-${id}`))
+        .toBeInTheDocument()
     }
     expect(providerLoaderMock).toHaveBeenCalledTimes(1)
     expect(new Set(cardPropsMock.mock.calls.map(([props]) => props.agentId)))
--- a/frontend/src/app/build/page-extension.test.tsx
+++ b/frontend/src/app/build/page-extension.test.tsx
@@ -15,6 +15,7 @@ const routerPushMock = vi.hoisted(() => vi.fn())
 const routerReplaceMock = vi.hoisted(() => vi.fn())
 const cardRenderMock = vi.hoisted(() => vi.fn())
 const cardPropsMock = vi.hoisted(() => vi.fn())
+const cardActionMock = vi.hoisted(() => vi.fn())
 const providerLoaderMock = vi.hoisted(() => vi.fn())
 const providerLifetime = vi.hoisted(() => ({ mounts: 0, unmounts: 0 }))
 
@@ -118,13 +119,25 @@ vi.mock("@/lib/build-page-extension", async () => {
       cardRenderMock(agentId)
       cardPropsMock(props)
       const supplement = sharedData?.[String(agentId)]
-      return supplement
-        ? ReactModule.createElement(
-          "div",
-          { "data-testid": `agent-card-supplement-${agentId}` },
-          supplement,
-        )
-        : null
+      return ReactModule.createElement(
+        ReactModule.Fragment,
+        null,
+        supplement
+          ? ReactModule.createElement(
+            "div",
+            { "data-testid": `agent-card-supplement-${agentId}` },
+            supplement,
+          )
+          : null,
+        ReactModule.createElement(
+          "button",
+          {
+            type: "button",
+            onClick: () => cardActionMock(agentId),
+          },
+          `extension-action-${agentId}`,
+        ),
+      )
     },
   )
   return {
@@ -191,6 +204,7 @@ describe("BuildsPage extension boundaries", () => {
     routerReplaceMock.mockReset()
     cardRenderMock.mockClear()
     cardPropsMock.mockClear()
+    cardActionMock.mockClear()
     providerLoaderMock.mockReset()
     providerLoaderMock.mockResolvedValue({ "42": "supplement-42" })
     providerLifetime.mounts = 0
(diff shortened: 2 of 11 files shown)


frontend/src/app/build/page.test.tsx
```diff
@@ -133,10 +149,103 @@ describe("BuildsPage Agent deletion", () => {
...
+    expect(screen.queryByRole("button", {
+      name: "builds.list.actions.publish",
+    })).not.toBeInTheDocument()
+    expect(screen.queryByRole("button", {
+      name: "builds.list.actions.delete",
+    })).not.toBeInTheDocument()
+    expect(screen.queryByRole("button", {
+      name: "builds.list.actions.edit",
+    })).not.toBeInTheDocument()
+  })
```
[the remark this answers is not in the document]
AlexLiu190625: Reorganized the file into separate top-level `BuildsPage rendering` and `BuildsPage Agent deletion` suites. Shared mock reset remains in a file-local helper so both suites use the same setup. Verified with `npm run test:app-pages` (12/12) and `npm run type-check`.

frontend/src/app/page.tsx
```diff
@@ -1,5 +1,6 @@
 "use client";
 
+import React from "react";
```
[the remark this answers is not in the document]
AlexLiu190625: Keeping this import. The repository’s current Vitest/Vite configuration uses the classic JSX transform for this test path; removing the import emits `React.createElement` without a `React` binding and makes `src/app/page.test.tsx` fail at runtime. Next.js production compilation can omit it, but the current test toolchain cannot without a broader configuration change.

frontend/src/app/build/page.tsx
```diff
@@ -651,6 +653,9 @@ export default function BuildsPage() {
                           {t('builds.card.updatedAt')}: {formatDate(agent.updated_at || agent.created_at)}
                         </div>
                       </div>
+                      {buildPageExtension.renderAgentCardSupplement({
+                        agentId: agent.id,
+                      })}
```
rogercloud: **major:** This extension point is invoked as a plain function call inside `.map()`, not rendered as a React element — it gets no fiber of its own. A downstream implementation needing `useState`/`useEffect`/a data-fetching hook cannot be written here without violating the Rules of Hooks (hook count would vary with list length). Same issue at `frontend/src/app/page.tsx:541`.

Suggestion: change the contract to expose a component type and render it as JSX:
```tsx
// page-extension-contracts.ts
export interface BuildPageExtension {
  AgentCardSupplement: React.ComponentType<BuildAgentCardExtensionProps>;
}

// build/page.tsx
<ext.AgentCardSupplement agentId={agent.id} />
```
This also removes the need for `useBuildPageExtension` to be hook-shaped — see `build-page-extension.tsx`.

frontend/src/app/page.tsx
```diff
@@ -536,6 +538,7 @@ export default function Home() {
 
         </div>
       </div>
+      {homePageExtension()}
```
rogercloud: **major:** Same issue as `build/page.tsx:658` — called as a bare function (`{homePageExtension()}`) rather than rendered as `<HomePageExtension />`, so a downstream implementation gets no independent component boundary and cannot safely use hooks. Recommend switching both Home and Build to component-based contracts (see the other comment for a sketch).

frontend/src/lib/page-extension-contracts.ts
```diff
@@ -0,0 +1,14 @@
+import type { ReactNode } from "react"
+
+export type HomePageExtension = () => ReactNode
+
+export interface BuildAgentCardExtensionProps {
+  agentId: number
```
rogercloud, line 10: **minor:** `BuildAgentCardExtensionProps` carries only `agentId`, but the call site (`build/page.tsx:658`) already has the full `Agent` object in hand. Any downstream supplement needing name/status/timestamp must re-fetch by id, duplicating a request the page already made. Consider passing the `Agent` object through, or add a comment noting the id-only narrowness is a deliberate trade-off.

frontend/package.json
```diff
@@ -11,6 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
+    "test:app-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page.test.tsx",
```
rogercloud: **minor:** `test:app-pages` hardcodes two file paths instead of globbing. `src/app/layout.test.tsx`, `src/app/settings/page.test.tsx`, and `src/app/widget/chat/[token]/page-client.test.tsx` all pass today but are run by no CI script at all — verified by running them directly. A glob over `src/app/**/*.test.tsx` would pick these up automatically along with future page tests; it would need to carry over the existing exclusion of `src/app/workforces/workforces-routes.test.tsx`, which has a known failing assertion already acknowledged via TODO in `ci.yml`.

frontend/src/lib/build-page-extension.tsx
```diff
@@ -0,0 +1,11 @@
+import type {
+  BuildAgentCardExtensionProps,
+  UseBuildPageExtension,
+} from "@/lib/page-extension-contracts"
+
+export const useBuildPageExtension = (() => ({
```
rogercloud: **minor:** `useBuildPageExtension` returns a fresh object and a fresh function identity on every call, with no memoization. Harmless today (the call site uses the result directly in the render body, never in a dependency array or passed to a memoized child), but a latent footgun for future consumers. Moot if this moves to a component-based contract per the comment on `build/page.tsx:658`.

frontend/src/lib/home-page-extension.tsx
```diff
@@ -0,0 +1,3 @@
+import type { HomePageExtension } from "@/lib/page-extension-contracts"
+
+export const homePageExtension = (() => null) satisfies HomePageExtension
```
rogercloud: **minor:** `satisfies` does not widen the exported symbol to the contract type, so the public type of this module is the narrow literal shape rather than `HomePageExtension` (same for `useBuildPageExtension` in `build-page-extension.tsx:6`). No assignability problems today, but for a module whose whole purpose is advertising a contract, an explicit annotation communicates intent better: `const homePageExtension: HomePageExtension = () => null;`.

frontend/src/app/build/page.test.tsx
```diff
@@ -52,12 +64,16 @@ vi.mock("@/lib/branding", () => ({
 vi.mock("@/components/voice-input-controller", () => ({
   useVoiceInputControls: () => ({
     status: "idle",
-    hasAsrModel: false,
+    hasAsrModel: true,
```
rogercloud: **minor:** This `useVoiceInputControls` mock is file-scoped, so flipping `hasAsrModel` from `false` to `true` here also silently changes the input for the pre-existing `"BuildsPage Agent deletion"` suite below. Verified harmless — no test in that suite asserts on `hasAsrModel`, and the `false` branch is still covered in `components/voice-input-controller.test.tsx` — but a one-line comment noting the shared-mock change is intentional would help future readers.

frontend/src/lib/build-page-extension.tsx
```diff
@@ -0,0 +1,11 @@
...
+  renderAgentCardSupplement: (props: BuildAgentCardExtensionProps) => {
+    void props
```
rogercloud: shrink: `void props` exists only to silence an unused-parameter lint warning. Drop the parameter entirely — `renderAgentCardSupplement: () => null` structurally satisfies the wider contract type. Note the usual `_props` workaround does *not* silence this repo's `@typescript-eslint/no-unused-vars` (no `argsIgnorePattern` configured — verified `_props` still errors).

frontend/src/lib/page-extension-contracts.ts
```diff
@@ -0,0 +1,14 @@
+import type { ReactNode } from "react"
```
rogercloud: indirection (low priority): each exported type here has exactly one consumer — `home-page-extension.tsx` and `build-page-extension.tsx` each import one type, and neither page component imports from this file directly. Inlining these into their respective impl files would remove a layer. Not blocking, and moot if the contracts are reshaped per the comment on `build/page.tsx:658`.

rogercloud (commented):
## Summary

This PR introduces two typed no-op "extension seam" points into the canonical Home and Build pages, backed by a new shared contracts file, so downstream distributions can inject small pieces of UI without forking whole page components. The Home page gains a bare `homePageExtension()` call in JSX; the Build page gains a `useBuildPageExtension()` hook whose `renderAgentCardSupplement` is invoked per agent card inside a `.map()`. Both ship trivial `() => null` defaults, plus rendering tests and a new `test:app-pages` CI step.

The motivation is legitimate and the diff footprint is genuinely small, but three design-level concerns should be settled before this becomes the template for future seams.

**No substitution mechanism exists.** Nothing in the repo lets a downstream fork actually swap the default implementations: there is no `resolve.alias` in `next.config.mjs`, no path mapping in `tsconfig.json` beyond `@/*`, and no documentation describing an intended override path. In practice a downstream consumer's only option today is to patch the three-line default files in their fork — the same merge-conflict problem this PR sets out to avoid, just at smaller scale. That may be intentional, but it should be written down explicitly (a comment in `page-extension-contracts.ts` or a short doc note), ideally backed by a real alias so seam files are never edited downstream at all.

**The seams are called as functions, not rendered as components — the most important actionable item** (see inline comments below). Neither call site gives the extension its own React fiber, so a downstream implementation needing `useState`, `useEffect`, or a data-fetching hook cannot be written without violating the Rules of Hooks. For the Build seam this is worse than a lint technicality: the call happens inside a `.map()`, so hook call count would vary with list length. Since the most likely downstream use of a per-agent-card supplement is exactly "render something with its own state or its own fetch", the seam as shipped blocks its primary use case, and there is no `<Component />` boundary for `eslint-plugin-react-hooks` to flag it statically — a downstream author would only discover this at runtime.

**Shape asymmetry.** Home uses a bare function while Build uses a hook returning a slot object, with no stated rationale and no guidance for what a third page should do. The resolution should not be to collapse Build down to Home's bare-function shape — that would remove the only headroom the design currently has. The clear recommendation is to move both seams to component-based contracts (`React.ComponentType<Props>` slots rendered as JSX elements). That fixes the hooks problem, gives Home the same room to grow that Build has, and makes the answer for a third page obvious.

## Prior findings (re-review)

- **FIXED** — gemini-code-assist: new rendering tests nested inside the pre-existing deletion suite. The file now has two top-level `describe` blocks (`"BuildsPage rendering"` at line 155, `"BuildsPage Agent deletion"` at line 254) with shared setup factored into a common `resetMocks()` helper.
- **WAIVED** — gemini-code-assist: remove `import React from "react"` in `frontend/src/app/page.tsx:3`. The author's justification is correct and verified: `vitest.config.ts` registers no React plugin/automatic-runtime esbuild setting, `tsconfig.json` sets `"jsx": "preserve"`, and removing the import reproduces `ReferenceError: React is not defined` at test runtime. Keep the import.

## Testing performed

- `npm run test:app-pages` — 12/12 passed
- `npm run type-check` — clean
- Targeted ESLint — 0 errors, 4 pre-existing `no-img-element` warnings
- Additionally verified: `src/app/layout.test.tsx`, `src/app/settings/page.test.tsx`, and `src/app/widget/chat/[token]/page-client.test.tsx` all pass but are wired into no CI script; `src/app/workforces/workforces-routes.test.tsx` fails 1/21 on a pre-existing stale width assertion (unrelated to this PR, already TODO-acknowledged in `ci.yml`).

frontend/src/app/build/page.tsx
AlexLiu190625: Changed this seam to a real `BuildAgentCardExtension` component rendered through JSX, with a stable page-level `BuildPageExtensionProvider`. The focused tests use hook-bearing memo components and cover Provider lifetime across initial loading and a publish-triggered refresh.

frontend/src/app/page.tsx
AlexLiu190625: Changed the Home seam to an explicitly typed `HomePageExtension` rendered as `<HomePageExtension />`. Its focused test uses a hook-bearing memo component, so the former bare-function invocation fails the regression test.

frontend/src/lib/build-page-extension.tsx
AlexLiu190625: Removed `useBuildPageExtension` entirely. The default implementation now exports stable, explicitly typed Provider and card components, so there is no per-render object or callback identity.

frontend/src/lib/home-page-extension.tsx
AlexLiu190625: Replaced `satisfies` with explicit `ComponentType`-based annotations for the Home component and both Build components. The exported symbols now advertise the shared contract directly.

frontend/src/app/build/page.test.tsx
AlexLiu190625: Moved the extension-boundary coverage into a dedicated `page-extension.test.tsx` file and restored the original Agent deletion test file byte-for-byte, including its `hasAsrModel: false` input.

frontend/src/lib/build-page-extension.tsx
AlexLiu190625: Removed the render callback and its unused parameter entirely. The no-op default is now an explicitly typed `BuildAgentCardExtension` component defined as `() => null`.

frontend/src/lib/page-extension-contracts.ts
AlexLiu190625: Keeping the `agentId`-only boundary intentionally. `Agent` is a page-local representation, while the extension contract should remain independent of its fields and evolution. `agentId` is the stable join key for data owned by a page-level Provider, so this shape does not require one request per card and avoids widening the public seam.

frontend/package.json
AlexLiu190625: Keeping this command focused and renamed it to `test:home-build-pages` so its scope is explicit. Globbing all app tests would expand this PR's CI responsibility and require an exclusion for the known-failing workforces suite. Broad app-test coverage should be introduced separately rather than coupled to this extension-boundary check.

frontend/src/lib/page-extension-contracts.ts
AlexLiu190625: Keeping the shared contracts file. It is the single compatibility boundary used by canonical page call sites, default implementations, and embedding replacements. Inlining those types into the implementation modules would duplicate the seam shape and allow replacements to drift from the page-owned contract.

frontend/package.json
```diff
@@ -11,6 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
+    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page-extension.test.tsx",
```
rogercloud: **major (new regression, CI coverage):** This rename dropped coverage. Before this commit, `test:app-pages` ran both `src/app/page.test.tsx` and `src/app/build/page.test.tsx` (the latter then containing both the extension tests and the pre-existing "Agent deletion" suite). After the split, `test:home-build-pages` only runs `src/app/page.test.tsx` and the new `src/app/build/page-extension.test.tsx` — `src/app/build/page.test.tsx` is in no CI-invoked script (`test:pages` is scoped to `src/components/pages`; `test:run` is commented out behind a TODO). Those 7 Agent-deletion tests pass in isolation and are unrelated to this PR's own changes, but they were covered by CI before this commit and are silently orphaned now.

Fix: add `src/app/build/page.test.tsx` back to the `test:home-build-pages` file list, or give it its own CI step.

frontend/src/app/build/page-extension.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      await refreshedAgents.promise
+    })
+    await screen.findByText("builds.emptyState.title")
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+
+    view.unmount()
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 1 })
+  })
+
+  it("renders voice input in the create dialog", async () => {
```
rogercloud: **minor, not new:** These three tests ("renders voice input in the create dialog", "renders privileged actions for an editable Agent", "limits a published read-only Agent to run and view actions") exercise `BuildsPage` button/dialog logic and never touch `BuildPageExtensionProvider`/`BuildAgentCardExtension`. They're carried over verbatim from the prior version of `build/page.test.tsx` — this file split was a natural moment to relocate them to a more fitting file (or drop if redundant) but that wasn't done. Not a regression, just an opportunity missed.

rogercloud (commented):
## PR summary

This PR carves out three product-neutral extension seams in the frontend — one on the Home page and two on the Build page — so embedding distributions can inject their own UI at those points without forking the canonical pages. The seams ship with no-op default implementations plus a shared contracts module defining their types, and a scoped CI script runs the affected page test suites.

## Update summary

The new commit ("ref: use component boundaries for page extensions") reworks the mechanism from plain function/fake-hook calls into real React components: `HomePageExtension`, `BuildPageExtensionProvider`, and `BuildAgentCardExtension`, typed as `ComponentType` in `frontend/src/lib/page-extension-contracts.ts` and rendered as JSX at the call sites. It also splits the Build tests — `frontend/src/app/build/page.test.tsx` is restored byte-for-byte to the pre-PR baseline and all extension tests move to a new `frontend/src/app/build/page-extension.test.tsx` — and renames the CI script `test:app-pages` to `test:home-build-pages` with an updated file list and CI step.

This is a genuine architectural fix rather than a patch: both major findings from the last round are resolved, and the Home/Build shape asymmetry now reads as justified (Build needs a page-level Provider to anchor N per-card extensions; Home has no repeated structure).

## Prior findings

1. **FIXED** — Build hooks-blocking (`frontend/src/app/build/page.tsx:658`): `BuildAgentCardExtension` is a real `ComponentType` rendered as JSX inside the card map; a hook-bearing `React.memo` mock proves hooks work, and a mount-counter test proves the Provider survives loading → publish → refresh without remounting.
2. **FIXED** — Home hooks-blocking (`frontend/src/app/page.tsx:541`): `HomePageExtension` is a typed `ComponentType` rendered as `<HomePageExtension />`, verified with a hook-bearing mock.
3. **NOT FIXED** — agentId-only props (`frontend/src/lib/page-extension-contracts.ts:6`): the waiver rests on `agentId` being "the stable join key for data owned by a page-level Provider," but `BuildPageExtensionProvider` is currently `({children}) => <>{children}</>` — no `createContext`/`useContext`, no state, no fetch. There is no data-sharing channel backing the claim, so a downstream per-card extension needing more than an id still fetches per card. Please either wire the Provider to actually hold and share that data, or drop the "avoids one request per card" framing.
4. **WAIVED** — Hardcoded test paths in the CI script (`frontend/package.json:14`): keeping the list narrow rather than globbing (which would pull in the known-failing `workforces-routes.test.tsx`) is a reasonable deferral, and the rename was applied consistently with no dangling references to the old name. See new finding #1 for a side effect of the same change.
5. **FIXED** — Per-render identity (`frontend/src/lib/build-page-extension.tsx:6`): `useBuildPageExtension` is gone entirely (zero remaining references), replaced by two module-scope component exports.
6. **FIXED** — `satisfies` narrowing exported types (`frontend/src/lib/home-page-extension.tsx:3`, `frontend/src/lib/build-page-extension.tsx:6`): both now use explicit type annotations against the shared contract types; `satisfies` is gone from both.
7. **FIXED** — Shared `hasAsrModel` mock flip (`frontend/src/app/build/page.test.tsx:67`): the file is byte-for-byte identical to the pre-PR baseline again, and the `hasAsrModel: true` case lives in the new isolated file with its own file-scoped mock.
8. **FIXED** — `void props` (`frontend/src/lib/build-page-extension.tsx:8`): the default takes no parameter at all; ESLint clean.
9. **WAIVED** — Contracts file inlining (`frontend/src/lib/page-extension-contracts.ts:1`): the stated rationale isn't literally accurate (the page components import the implementations, not the contract types), but the structural point holds — the contracts file is the one shared type-checking surface between each default implementation and any downstream replacement, so inlining would duplicate that shape per replacement. Accepted; this was always low priority.

## Design note

The one PR-level reservation still open across both rounds is the **absence of an actual substitution mechanism**. The contract shapes are now clean and swappable in principle — one module per seam, stable named exports, product-neutral types — but there is no implemented path mapping (`next.config.mjs` has no `resolve.alias`; `tsconfig.json` has only the standard `@/*` mapping) and no documentation beyond one inline comment in `page-extension-contracts.ts` stating that embedding distributions "may replace default implementation modules during frontend composition." A downstream consumer therefore has no supported, verifiable way to perform that replacement, and nothing in CI would catch it if the seam contract drifted out from under them. This doesn't need to block the PR, but it should be resolved (alias config plus a short doc, ideally with one test that swaps an implementation) before the seams are advertised as usable.

Lean already on this update — no simplification opportunities to report.

## Testing performed

- `frontend/src/app/page.test.tsx` and the new `frontend/src/app/build/page-extension.test.tsx` pass locally (6/6 relevant new tests observed passing).
- `frontend/src/app/build/page.test.tsx` passes in isolation (7/7) but is not invoked by CI — see new finding #1 below.
- ESLint clean on `frontend/src/lib/build-page-extension.tsx` (zero errors/warnings).

frontend/package.json
AlexLiu190625: Fixed. `test:home-build-pages` now runs the Home test, the canonical Build page test, and the Build extension test. The focused command covers 3 files and 14 tests, including the existing Agent deletion suite.

frontend/src/app/build/page-extension.test.tsx
AlexLiu190625: Fixed. The voice-input and Agent-action tests now live in `build/page.test.tsx`; `page-extension.test.tsx` is limited to component-boundary, shared-Provider, and Provider-lifecycle coverage. The ASR mock resets to the deletion suite default, and only the voice-input test opts in.

frontend/src/lib/page-extension-contracts.ts
AlexLiu190625: Kept the `agentId`-only boundary and made the ownership contract explicit. The canonical replacement test now uses a Provider-owned asynchronous load with three cards joining distinct values by ID, and verifies each card receives only `{ agentId }`. This demonstrates shared page data without coupling the extension contract to the page-local `Agent` shape.

AlexLiu190625:
The supported substitution boundary is the exact implementation modules imported by the canonical pages: frontend composition replaces `home-page-extension.tsx` and `build-page-extension.tsx` as a module pair while `page-extension-contracts.ts` remains the shared type-checking owner. The canonical page tests substitute those exact modules and verify the Home component boundary plus the Build Provider, one shared asynchronous load, and per-card ID joins. Adding a separate runtime alias would create a second configuration contract without changing the page data flow or strengthening the replacement contract.

frontend/package.json
AlexLiu190625: Tracked in #1090. The durable fix is to make frontend CI discover every test under `src/app` after repairing the existing Workforce width assertion, rather than add an exclusion or keep extending a per-file allowlist. This PR keeps the focused Home/Build lane so the known failure is not hidden; #1090 owns replacing it with the broad app suite.

frontend/src/app/build/page.tsx
```diff
@@ -651,6 +656,7 @@ export default function BuildsPage() {
                           {t('builds.card.updatedAt')}: {formatDate(agent.updated_at || agent.created_at)}
                         </div>
                       </div>
+                      <BuildAgentCardExtension agentId={agent.id} />
```
rogercloud: **major (new, correctness):** This extension slot sits outside the card's click-propagation boundary. The card root has `onClick={() => router.push(\`/build/${agent.id}\`)}` (around line 536-539). The `<div className="flex gap-2" onClick={(e) => e.stopPropagation()}>` wrapper right below this (line 660) exists precisely to stop the action buttons from triggering that navigation — but `<BuildAgentCardExtension>` is a plain sibling, not inside any stopPropagation boundary. React synthetic clicks bubble through the real DOM tree, so any downstream supplement containing an interactive element (button, link, checkbox) will also navigate the card when clicked. This is a bug in the extension-point contract itself, not in any downstream implementation, and there's currently zero test coverage for it (the mocked extension in `page-extension.test.tsx` only renders a plain non-interactive `<div>`).

Suggested fix:
```tsx
<div onClick={(e) => e.stopPropagation()}>
  <BuildAgentCardExtension agentId={agent.id} />
</div>
```
Wrapping is strictly better than documenting "consumers must stop propagation themselves" — it doesn't rely on every future implementer remembering.

frontend/src/app/page.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    startRecording: vi.fn(),
+    stopRecording: vi.fn(),
+  }),
+}))
+
+vi.mock("@/components/welcome-modal", () => ({
+  WelcomeModal: () => null,
+}))
+
+vi.mock("@/lib/home-page-extension", async () => {
```
rogercloud: **minor:** This mock means the shipped `HomePageExtension` default (`() => null`) is never rendered by any test — a regression making it throw or emit stray DOM would go uncaught. `src/app/build/page.test.tsx` has no equivalent mock for `@/lib/build-page-extension`, so the Build defaults are exercised there; consider one unmocked smoke assertion here too for parity.

frontend/src/app/build/page.tsx
```diff
@@ -462,7 +466,8 @@ export default function BuildsPage() {
   }
 
   return (
-    <div className="flex flex-col h-full bg-background">
+    <BuildPageExtensionProvider>
```
rogercloud: **minor (code quality):** The ~425-line JSX subtree was wrapped in `<BuildPageExtensionProvider>` without re-indenting; `npx prettier --check` flags this file, and the closing tags around line 896-898 show inverted indentation (child tag less indented than parent). Please run the formatter on this file.

frontend/package.json
```diff
@@ -11,6 +11,7 @@
     "test": "vitest",
     "test:run": "vitest run",
     "test:pages": "vitest run --config vitest.config.ts src/components/pages",
+    "test:home-build-pages": "vitest run --config vitest.config.ts src/app/page.test.tsx src/app/build/page.test.tsx src/app/build/page-extension.test.tsx",
```
rogercloud: **minor, optional:** Still a hardcoded three-path allowlist. Verified a directory-scoped alternative works cleanly: `vitest run --config vitest.config.ts src/app/build src/app/page.test.tsx` discovers all 3 current files (14/14 passing) while correctly excluding the sibling `src/app/workforces` suite. This would auto-cover any new file added under `src/app/build/` going forward. Narrower than the full `src/app` glob owned by #1090, so it could be adopted now independently — not blocking either way.

frontend/src/app/build/page-extension.test.tsx
```diff
@@ -0,0 +1,308 @@
...
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react"
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+
+const apiRequestMock = vi.hoisted(() => vi.fn())
```
rogercloud, line 13: **minor, nice-to-have:** This file and `page.test.tsx` share roughly 85-100 lines of near-identical `vi.mock` setup and fixture data. Note: a clean extraction isn't a verbatim move — Vitest hoists `vi.mock` per file and these factories close over `vi.hoisted()` locals, so the real fix is extracting shared *factory functions and fixture data*, with each file keeping its own local `vi.mock(...)` wrapper around them. Not required, given that caveat.

frontend/src/app/page.tsx
```diff
@@ -536,6 +538,7 @@ export default function Home() {
 
         </div>
       </div>
+      <HomePageExtension />
```
rogercloud: **minor, low priority:** `<HomePageExtension />` is a direct child of a `flex flex-col overflow-hidden ... overflow-y-auto` root, with no wrapper, no `shrink-0`, and no comment about layout expectations. Downstream has no stated guarantee about available space or shrink behavior in this slot.

rogercloud (commented):
## PR summary

This PR introduces extension seams for the Home and Build pages: a `HomePageExtension` slot, a `BuildPageExtensionProvider` wrapper around the Build page tree, and a per-card `BuildAgentCardExtension` slot, with default no-op implementations shipped in `frontend/src/lib/`. The intent is that downstream builds substitute their own implementations without forking the page components. The extension contract is defined in `frontend/src/lib/page-extension-contracts.ts`.

## What changed in this update

Two commits: `c35565e7` ("fix: preserve Build page test coverage") restores `src/app/build/page.test.tsx` to the CI script and moves three non-seam tests back out of `page-extension.test.tsx` into `page.test.tsx`. `2e4c00ff` ("fix: pin Build extension data joins") rewrites `page-extension.test.tsx` around a Provider-owned async load with three cards joining distinct values by id, and adds a one-line comment to `page-extension-contracts.ts` describing `agentId` as the join key.

## Prior findings

| Finding | Status |
| --- | --- |
| CI coverage regression: `test:home-build-pages` dropped `src/app/build/page.test.tsx` | **FIXED** — script now lists all 3 files; 14 tests run and pass; no cross-file mock leakage |
| Three non-seam tests living in `page-extension.test.tsx` | **FIXED** — relocated to `page.test.tsx:144,161,181`; `page-extension.test.tsx` now holds only the 3 genuine seam tests |
| `BuildAgentCardExtensionProps` is `agentId`-only with no shipped data-sharing mechanism | **NOT FIXED (third round raising this)** — see design note below |
| CI script uses a hardcoded file allowlist rather than a glob | **WAIVED** — deferral is now backed by a real, open, on-topic issue (#1090) |

On the NOT FIXED item: the framing has improved — the claim is no longer that the shipped Provider already shares data, but that the test is a "canonical replacement" example. However, `BuildPageExtensionProvider` in `frontend/src/lib/build-page-extension.tsx` is still literally `({ children }) => <>{children}</>` — no `createContext`, no `useContext`, no state, no fetch (repo-wide grep for `createContext` under `frontend/src/lib/` returns 0 hits). The `createContext` + async-load + id-join machinery lives entirely inside the `vi.mock` factory in `page-extension.test.tsx:75-134`, which replaces the real module wholesale. That proves the contract shape can support a Context-based downstream implementation; it does not change what ships. Nothing in the contracts file enforces or type-checks the Provider/card pairing — the coupling exists only in a comment and a test.

## Design note

The core architectural work here is genuinely solid now — the component boundaries are clean, hooks work inside the extensions, and the tests are real rather than trivially green. But two things have gone three full rounds without any movement in shipped code. First, the substitution mechanism: there is still no alias or path-mapping implemented anywhere in the repo, and no documentation beyond a single inline comment, so how a downstream build actually swaps these modules remains unspecified. Second, the `agentId`-only/Provider data-sharing question has now been discussed in all three rounds, and each round has added a comment or a test-only example rather than changing what ships.

The reason this one keeps recurring is that it's ambiguous which of two designs is intended. Either the shipped `BuildPageExtensionProvider`/`BuildAgentCardExtension` pair should carry an actual data-sharing primitive — a minimal exported `createContext` plus a hook, roughly ten lines, would settle it concretely — or the PR should state outright that the Provider is a placeholder with no guarantees and that downstream is expected to replace the Provider and its data-sharing mechanism together. Either is a legitimate answer. Picking one explicitly is what would close the item; adding another comment or test-side example will not.

## Testing performed

- `npm run test:home-build-pages` → 3 files, 14/14 passing (`src/app/page.test.tsx` 1, `src/app/build/page-extension.test.tsx` 3, `src/app/build/page.test.tsx` 10)
- ESLint on the changed files → clean, 0 errors
- Proof-of-concept for the CI directory-filter suggestion: `vitest run --config vitest.config.ts src/app/build src/app/page.test.tsx` → 14/14 passing, correctly excludes unrelated suites
- `npx prettier --check` on `src/app/build/page.tsx` → reports the file needs formatting

frontend/src/app/build/page.tsx
AlexLiu190625: Fixed at the canonical card boundary. The extension and the existing action controls now share a local propagation boundary, while the card root still owns navigation. The regression test clicks an interactive extension control and proves it does not navigate, then clicks the card title and proves `/build/42` still does.

frontend/src/app/page.test.tsx
AlexLiu190625: Fixed. The Home test now renders the actual shipped default implementation inside the canonical slot and verifies that the slot remains empty and inert. The configured-extension test separately verifies that the replacement component renders inside that slot. The temporary module unmock is restored in `finally` so later tests cannot inherit it.

frontend/src/app/build/page.tsx
AlexLiu190625: Fixed structurally. The page body is now a module-level `BuildsPageContent`, and the default export only composes `BuildPageExtensionProvider` around it. This removes the unindented large wrapper without reformatting unrelated legacy code, while preserving Provider lifetime.

frontend/package.json
AlexLiu190625: Applied the directory-scoped form. `test:home-build-pages` now runs `src/app/build/` plus the Home page test, so future Build test files are discovered automatically without pulling in the unrelated Workforce suite.

frontend/src/app/page.tsx
AlexLiu190625: Fixed. The Home page now owns a canonical `shrink-0` slot wrapper, and tests cover both a configured extension and the empty shipped default so replacement components do not need to infer the parent flex behavior.

frontend/src/lib/page-extension-contracts.ts
AlexLiu190625: The purpose of this seam is to keep the canonical Home and Build pages single-sourced: embedding builds replace small implementation modules instead of maintaining copies of the full pages. Passing the page-local `Agent` shape would move page ownership back into the extension contract and make it evolve with the page. The paired Provider/card implementation can perform one page-level batch load and join by `agentId`, so the narrow contract does not require a request per card. Keeping this boundary intentionally.

frontend/src/app/build/page-extension.test.tsx
AlexLiu190625: Keeping this duplication intentionally local. The single-source goal applies to the product pages; these test modules have independent `vi.hoisted()` state and lifecycle assumptions. A shared parameterized factory would couple those tests and add more indirection than it removes. The Home extension factory is extracted only within its own file where it is needed to restore the module mock after the real-default probe.

frontend/src/lib/build-page-extension.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+
+import type {
+  BuildAgentCardExtensionComponent,
+  BuildPageExtensionProviderComponent,
+} from "@/lib/page-extension-contracts"
+
+export const BuildPageExtensionProvider: BuildPageExtensionProviderComponent = ({
+  children,
+}) => <>{children}</>
```
rogercloud, line 10: NOT FIXED (fourth round). This is still `({ children }) => <>{children}</>` — no `createContext`, no state, no invalidation path. Combined with `BuildPageExtensionProviderProps` being `{ children }` only, a downstream Provider that loads agent-keyed data at mount has no way to learn that the Agent list changed (create/delete/publish/rename are all invisible to it), because the Provider is mounted once at `frontend/src/app/build/page.tsx:902-908`, outside every refetch path.

Please close this in this PR by picking one explicitly:
(a) add a minimal `createContext` + hook and put something in the props that changes when the Agent list changes (a refresh token, or the agent ID list), so a downstream Provider can invalidate; or
(b) document that this Provider is a placeholder with no staleness guarantees and that a downstream distribution must replace the Provider and supply its own data-sharing/invalidation mechanism.

Either is acceptable; the current state is neither, and it has now been raised in rounds 2, 3, and 4.

frontend/src/app/build/page-extension.test.tsx
```diff
@@ -0,0 +1,341 @@
...
+    expect(screen.getByText("common.loading")).toBeInTheDocument()
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+
+    await act(async () => {
+      refreshedAgents.resolve(jsonResponse([]))
+      await refreshedAgents.promise
+    })
+    await screen.findByText("builds.emptyState.title")
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 0 })
+    expect(providerLoaderMock).toHaveBeenCalledTimes(1)
```
rogercloud: This assertion is what turns the Provider staleness gap from a limitation into an asserted contract: `providerLoaderMock` called exactly once, inside a test named "keeps the page provider mounted across loading and a real publish refresh". As written, the suite documents "the Provider never reloads across a publish-triggered list refresh" as intended, canonical behavior.

If you take option (b) on the Provider (placeholder, no staleness guarantees), please also re-word this test name so it reads as a documented limitation rather than a guarantee. If you take option (a), this assertion should change to expect a reload after the refresh.

frontend/src/app/build/page.test.tsx
```diff
@@ -6,6 +6,7 @@ const apiRequestMock = vi.hoisted(() => vi.fn())
 const routerPushMock = vi.hoisted(() => vi.fn())
 const routerReplaceMock = vi.hoisted(() => vi.fn())
 const toastErrorMock = vi.hoisted(() => vi.fn())
+const voiceInputState = vi.hoisted(() => ({ hasAsrModel: false }))
```
rogercloud, line 9: PARTIAL on the "shipped defaults are inert" finding. Home's side is now covered by `frontend/src/app/page.test.tsx:122-143`, which unmocks `@/lib/home-page-extension` and asserts the real default renders an empty slot. The Build side has no equivalent: `frontend/src/app/build/page-extension.test.tsx` mocks `@/lib/build-page-extension` for all of its tests, and this file never mocks it but also never asserts anything about the defaults' DOM footprint.

Suggest mirroring the Home test here: assert the real `BuildPageExtensionProvider` renders its children unwrapped, and that the real `BuildAgentCardExtension` contributes no DOM to a rendered card. Otherwise the asymmetry has just moved from Home to Build.

frontend/src/lib/page-extension-contracts.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// A Build replacement must supply the Provider and card extension together from one implementation module.
+export interface BuildAgentCardExtensionProps {
+  // Stable key for joining Provider-owned page data to an Agent card.
+  agentId: number
+}
+
+export interface BuildPageExtensionProviderProps {
+  children: ReactNode
+}
```
rogercloud, line 15: Minor, docs: no StrictMode guidance for downstream Provider implementers. `frontend/next.config.mjs` sets `reactStrictMode: true`, so a Provider's mount effect double-invokes in dev. The suite's own mock provider guards against exactly this with an `active` flag (`frontend/src/app/build/page-extension.test.tsx:87-99`), but nothing in the contracts tells an implementer they need the same guard — and the `mounts: 1` assertions only hold because React Testing Library does not wrap `render()` in `<React.StrictMode>` by default, so a downstream Provider missing the guard would double-fire in dev with no test catching it.

Suggest a one-line note on `BuildPageExtensionProviderProps`, e.g. "Implementations must tolerate StrictMode double-invocation of mount effects (guard async loads with a cancellation flag)."

frontend/src/app/page.tsx
```diff
@@ -536,6 +538,9 @@ export default function Home() {
 
         </div>
       </div>
+      <div data-slot="home-page-extension" className="shrink-0">
```
rogercloud, line 541: Minor: this wrapper renders unconditionally even though the shipped default child renders `null`, so every non-extension consumer of the Home page now gets a new empty `<div>` that did not exist before this PR. It is harmless today — no layout-affecting classes, and `frontend/src/app/page.test.tsx:122-143` pins that — but it is a permanent DOM footprint for an otherwise no-op feature. Worth recording as a deliberate trade-off (stable slot for downstream) rather than leaving it implicit.

frontend/src/app/page.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+      const { container } = render(<DefaultHome />)
+      const slot = container.querySelector('[data-slot="home-page-extension"]')
+
+      expect(slot).toBeInTheDocument()
+      expect(slot).toHaveClass("shrink-0")
+      expect(slot).toBeEmptyDOMElement()
+      expect(
+        Array.from(slot?.classList ?? []).some((className) =>
+          /(?:^|:)(?:p[trblxy]?|m[trblxy]?|min-h)-/.test(className),
```
rogercloud: Test-quality nit: the inertness check is a class-name regex proxy rather than a layout assertion. It would miss an added inline `style={{ padding }}` or a `gap-`/`h-` utility class, and can false-fail on an unrelated future class that happens to share one of these prefixes. Since you already assert `toBeEmptyDOMElement()`, an exact `expect(slot).toHaveAttribute("class", "shrink-0")` would be both stricter and less brittle than the regex.

frontend/src/app/page.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+      "data-slot",
+      "home-page-extension",
+    )
+    expect(extension.parentElement).toHaveClass("shrink-0")
+    expect(screen.getAllByTestId("home-extension")).toHaveLength(1)
+    expect(homeExtensionRenderMock).toHaveBeenCalled()
+    await waitFor(() => expect(apiRequestMock).toHaveBeenCalledTimes(2))
+  })
+
+  it("renders the shipped default extension in an inert canonical slot", async () => {
```
rogercloud, line 122: Test-quality nit: `vi.doUnmock` + `vi.resetModules()` + dynamic `import("./page")` makes this suite order-dependent. It works only because this is the last test in the file — any test appended after it runs against the stale static `Home` binding with the mock registry reset out from under it. The `finally` block re-mocks but cannot restore the top-level import. Moving this into its own `describe` with scoped `beforeEach`/`afterEach`, or into a separate file, would remove the ordering trap.

frontend/src/app/build/page-extension.test.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+    view.unmount()
+    expect(providerLifetime).toEqual({ mounts: 1, unmounts: 1 })
+  })
+
```
rogercloud: Nit: stray blank line between the last `it` block's closing `})` and the `describe`'s closing `})`.

rogercloud (approved):
## PR summary

This PR adds two "no-op by default" extension seams to the frontend so downstream embedding distributions can inject small UI pieces without forking the canonical pages: `HomePageExtension` for the Home page, and `BuildPageExtensionProvider` + `BuildAgentCardExtension` for the Build page (the Provider wraps the page, the card extension renders once per Agent with `agentId` as the join key). Both are backed by a shared `frontend/src/lib/page-extension-contracts.ts` types file, and all shipped defaults render nothing. A new CI step `test:home-build-pages` runs the three affected suites.

## Update summary

One new commit since the last review (`f523f170`, "fix: harden page extension boundaries"). It moves `<BuildAgentCardExtension>` inside the card's `stopPropagation` wrapper in `frontend/src/app/build/page.tsx` so interactive extension content no longer bubbles a click into card navigation — the correctness issue raised in round 3 — and adds a regression test for it in `frontend/src/app/build/page-extension.test.tsx`. It also adds one clarifying comment to `page-extension-contracts.ts`. No other files were touched.

## Approach verdict

**Acceptable-with-reservations** — unchanged from the prior rounds. The seam shape is sound: the Home/Build asymmetry (bare component vs. Provider + per-card component) is justified by Build's repeated per-card structure, and shipping the Provider and card as one replaceable module pair correctly means a distribution's own Provider can hold arbitrary state beyond `agentId`. The reservations are the two long-standing ones below, of which one is now on its fourth round unresolved.

## Prior findings

| # | Finding | Status |
|---|---------|--------|
| 1 | Card extension outside the `stopPropagation` event boundary | **FIXED** |
| 2 | Provider has no data-sharing primitive / no staleness signal | **NOT FIXED** (4th round) |
| 3 | No unmocked smoke test that shipped defaults are inert | **PARTIAL** (Home fixed, Build still missing) |
| 4 | No substitution mechanism for downstream replacement | **NOT FIXED** (open design note, non-blocking) |
| 5 | `page-extension-contracts.ts` indirection value | **WAIVED** (round 2) |
| 6 | `test:home-build-pages` hardcoded file list vs. glob | **WAIVED** (tracked in open issue #1090) |

### 1. Event boundary — FIXED

`frontend/src/app/build/page.tsx:658-659` now renders `<BuildAgentCardExtension agentId={agent.id} />` as the first child of `<div className="space-y-4" onClick={(e) => e.stopPropagation()}>`, the same wrapper that holds the action-button row. `frontend/src/app/build/page-extension.test.tsx:228-245` ("keeps interactive extension controls inside the card event boundary") is a real regression guard: clicking the extension's button fires the extension handler and does *not* call `router.push`, while clicking the card title does call `router.push("/build/42")`. Good fix, correctly tested.

### 2. Provider data-sharing / staleness — NOT FIXED (fourth round raising this)

`frontend/src/lib/build-page-extension.tsx:8-10` is still literally:

```tsx
export const BuildPageExtensionProvider: BuildPageExtensionProviderComponent = ({
  children,
}) => <>{children}</>
```

No `createContext`, no hook, no state. `BuildPageExtensionProviderProps` (`page-extension-contracts.ts:12-14`) is still `{ children }` only — no agent IDs, no refresh signal, no callback. The Provider is mounted once at `frontend/src/app/build/page.tsx:902-908`, outside all list-refetch logic, so a downstream Provider that loads agent-keyed data at mount cannot learn when the Agent list changes: create, delete, publish, and rename are all invisible to it.

What makes this more than a documentation gap is that the test suite now asserts the staleness as intended behavior. `frontend/src/app/build/page-extension.test.tsx:335` asserts `expect(providerLoaderMock).toHaveBeenCalledTimes(1)` inside a test named "keeps the page provider mounted across loading and a real publish refresh" — i.e. "the Provider loads exactly once, even across a publish-triggered list refresh" is now a passing, asserted contract, presented as the canonical pattern.

This has been raised in rounds 2, 3, and now 4, and each round has answered it with more comments and tests around the ambiguity rather than a decision. Two acceptable resolutions, pick one explicitly:

- **(a)** Give the Provider a real primitive — a minimal `createContext` + hook, plus something in the props that changes when the Agent list changes (a refresh token or the agent ID list), so a downstream Provider can invalidate.
- **(b)** State in `page-extension-contracts.ts` that the Provider is a placeholder with **no** staleness guarantees, and that a downstream distribution must replace the Provider *and* supply its own data-sharing/invalidation mechanism. Then re-word the test name so it documents a limitation rather than a guarantee.

Not blocking the merge, but four rounds without a decision is overdue — please close it one way or the other in this PR rather than carrying it further.

### 3. Default-extension inertness smoke test — PARTIAL

Home's gap is closed: `frontend/src/app/page.test.tsx:122-143` adds "renders the shipped default extension in an inert canonical slot", which `vi.doUnmock`s `@/lib/home-page-extension` and asserts the real `() => null` default produces an empty, layout-neutral slot.

The identical gap now exists on the Build side and was never covered. `frontend/src/app/build/page-extension.test.tsx` mocks `@/lib/build-page-extension` for every test in the file, and `frontend/src/app/build/page.test.tsx` never mocks it but also never asserts anything about the defaults' absence or inertness in the DOM. Net effect: the asymmetry moved from Home to Build rather than being resolved. Suggest mirroring the Home test — assert the real `BuildPageExtensionProvider` renders children unwrapped and the real `BuildAgentCardExtension` contributes no DOM to a card.

### 4-6

- **No substitution mechanism (still open, non-blocking).** There is still no `resolve.alias`, no path mapping beyond `tsconfig.json`'s standard `@/*`, and no documentation of how a distribution swaps in its own implementations — a downstream's only option today is to patch the default files in a fork. Restating once because it has never moved; the seam is still useful without it.
- **`page-extension-contracts.ts` indirection — remains WAIVED** per round 2 (single shared compatibility boundary between the defaults and downstream replacements). `f523f170` only adds a comment line; no new evidence, not re-litigating.
- **`test:home-build-pages` hardcoded file list — remains WAIVED**, tracked by open issue #1090 ("Run all App Router tests in frontend CI", explicit follow-up to this PR). `f523f170` does not touch `ci.yml`.

## New findings

**Minor — no StrictMode guidance for downstream Provider implementers.** `frontend/next.config.mjs` sets `reactStrictMode: true`, so in dev a Provider's mount effect double-invokes. The suite's own mock provider defends against exactly this with an `active` flag (`frontend/src/app/build/page-extension.test.tsx:87-99`), but neither `page-extension-contracts.ts` nor `build-page-extension.tsx` tells a downstream implementer they need the same guard. The `mounts: 1` assertions only hold because React Testing Library does not wrap `render()` in `<React.StrictMode>` by default, so a real downstream Provider missing the guard would double-fire its mount effect in dev with nothing catching it. Suggest a one-line note on `BuildPageExtensionProviderProps` in `page-extension-contracts.ts`.

**Minor — unconditional empty DOM node on Home.** `frontend/src/app/page.tsx:541-543` always renders `<div data-slot="home-page-extension" className="shrink-0">` even though the default child renders `null`. Against the pre-PR base this is a genuinely new empty node for every non-extension consumer. Harmless today (no layout-affecting classes, and the new test pins that), but it is a permanent DOM footprint for an otherwise no-op feature. Worth a deliberate "yes, we accept this" rather than leaving it implicit.

**Minor — test-quality nits (batched):**
- `frontend/src/app/page.test.tsx:134-137`: the inertness check is a class-name regex proxy (`/(?:^|:)(?:p[trblxy]?|m[trblxy]?|min-h)-/`) rather than a layout assertion. It misses an inline `style={{ padding }}` or a `gap-`/`h-` utility, and can false-fail on an unrelated future class sharing a prefix. `toBeEmptyDOMElement()` plus an exact `className` equality check would be both stricter and less brittle.
- `frontend/src/app/page.test.tsx:122-143`: this test uses `vi.doUnmock` + `vi.resetModules()` + a dynamic `import("./page")`, which makes the suite order-dependent. It works only because it is the last test in the file — anything appended after it runs against the stale static `Home` binding with the mock registry reset out from under it. The `finally` block re-mocks but cannot restore the top-level import. Moving it to its own `describe` with a scoped `beforeEach`/`afterEach`, or its own file, would remove the ordering trap.
- `frontend/src/app/build/page-extension.test.tsx:340`: stray blank line between the last `it` block's `})` and the `describe`'s `})`.

## Testing performed

- `npm run test:home-build-pages` — 3 files, **16/16 passed** (up from 14 last round: the new event-boundary test and the new Home inertness test).
- `npm run type-check` — clean.
- ESLint on changed files — 0 errors (pre-existing unrelated warnings only).

Lean already.

frontend/src/lib/build-page-extension.tsx
AlexLiu190625: Clarified the contract at its owning boundary. The canonical page guarantees stable Provider lifetime and the agentId join key; the paired replacement implementation owns its data loading, sharing, and invalidation. This keeps product data and refresh policy out of the neutral page contract.

frontend/src/app/build/page-extension.test.tsx
AlexLiu190625: Separated the two invariants. The lifecycle test now asserts only Provider mount and unmount stability across loading and publish refresh. The independent multi-card test continues to prove one shared Provider load and per-ID joins, without turning a fixture-specific load count into a page lifecycle contract.

frontend/src/app/build/page.test.tsx
AlexLiu190625: Added direct tests for the shipped Build defaults. The real Provider is pinned to render its child as the sole direct element with no wrapper, and the real card extension is pinned to contribute no DOM. The directory-scoped focused test lane discovers the new file automatically.

frontend/src/app/page.test.tsx
AlexLiu190625: Simplified the inert slot assertion to its exact contract: empty DOM, exact shrink-0 class, and no inline style. This removes the prefix-regex proxy and fails on any added layout class or style.

frontend/src/app/build/page-extension.test.tsx
AlexLiu190625: Removed the stray blank line.

frontend/src/lib/page-extension-contracts.ts
AlexLiu190625: Not adding seam-specific StrictMode guidance here. The active flag prevents a late state update after cleanup, but it does not prevent StrictMode from invoking the effect again or issuing another request, so the suggested wording would imply a guarantee it cannot provide. StrictMode-safe effects remain the replacement component implementation's normal React responsibility; this contract stays limited to the page-owned lifetime and join-key guarantees.

frontend/src/app/page.tsx
AlexLiu190625: Keeping the unconditional slot intentionally. The page owns the stable insertion point and its parent flex behavior; making the slot conditional would require a second presence contract or move layout responsibility into the extension implementation. The data-slot marker plus the exact shrink-0, no-inline-style, and empty-DOM assertions record and protect this trade-off.

frontend/src/app/page.test.tsx
AlexLiu190625: Keeping the current test isolation. The top-level Home binding is already evaluated with the configured mock and resetModules does not retroactively change that binding; the finally block restores the mock registry for future module loads. This was also verified with shuffled seed 3, where the real-default probe ran first and the configured-extension test ran afterward, with both passing. Splitting the file would duplicate the page mock setup without addressing an observed ordering failure.

Issue #1065, opened 2026-07-30
refactor(channels): three diverging copies of the turn-orchestration machinery (telegram/feishu/slack)


OliverBryant:
Telegram, Feishu and Slack now carry three structurally independent copies of the same turn-orchestration machinery (~4.7k lines total across `src/xagent/web/channels/`). Only the DB/lease boundary in `services/channel_runtime.py` is shared; the queue handling, active-task persistence, stop-locking, mention gating, progress-message editing and Markdown conversion are duplicated per transport.

This was defensible for each transport in isolation, but the copies have already diverged in ways that are bugs rather than intentional differences:

- **`_extract_local_file_id` diverged silently.** The Slack copy added `html.unescape()` and dropped `unquote()`, so percent-encoded internal file ids stopped resolving and silently fell back to a text-reference message. Fixed in #1050 by porting `unquote()` back, but nothing prevents the next drift.
- **Table rendering exists in one copy only.** Telegram has `_format_markdown_tables_for_telegram` / `_render_markdown_table_as_list` / `_split_markdown_table_row`; Slack has no table handling at all (see #1064). Same agent output, different fidelity depending on which channel the user connected.
- **The sync lock exists in one copy only.** `SlackChannelManager.__init__` holds a `_sync_lock` guarding concurrent mutation of shared dicts across `await` points, because channel CRUD endpoints fire background sync tasks. `TelegramChannelManager` and `FeishuChannelManager` have the identical race and no lock — the fix here is to port the lock outward, not to remove it.
- **Dedup is reimplemented per transport**, all three in-memory (see #1063).

Suggested direction, smallest-useful-first rather than one big refactor:

1. Extract the genuinely identical helpers into a shared module (`channels/_shared/` or extend `channel_runtime`): `_extract_local_file_id`, `_clean_stripped_markdown_refs`, and the file-ref stripping skeleton.
2. Port `_sync_lock` to the Telegram and Feishu managers.
3. Extract only the "compute add/remove/update key sets" diff from the per-manager sync methods; leave the start/stop/update semantics separate, since their idempotency caches and lifecycle hooks genuinely differ.
4. Leave the turn pipeline itself alone for now — a shared base class would have to abstract over three different progress-update models and is not obviously a win.

Raised during the #1050 review, where this became the third copy.

PR #1050, opened 2026-07-29
feat: add Slack channel OAuth integration

qinxuye:
## Summary

- add Slack as a first-class user channel alongside Telegram and Feishu, including mentions, direct messages, threaded replies, progress updates, and file transfer
- add a real Slack OAuth installation flow with an `Add to Slack` popup, encrypted workspace token persistence, API response redaction, and manual Socket Mode setup as a fallback
- route OAuth-installed workspaces through one shared Socket Mode connection keyed by Slack `team_id`
- wire Slack into application startup/shutdown, channel display, configuration, dependencies, and lifecycle tests

## Why

The Channels page previously had no production Slack transport. Users need a workspace authorization flow that does not require each installer to paste bot and app tokens, while self-hosted deployments still need a manual setup option.

## Operator impact

The deployment owner configures one Slack app with `XAGENT_SLACK_CLIENT_ID`, `XAGENT_SLACK_CLIENT_SECRET`, `XAGENT_SLACK_APP_TOKEN`, and a public HTTPS callback URL. Workspace users then only click **Add to Slack**.

## Validation

- pre-commit hooks: ruff check/format, mypy, isort, codespell, npm lint, npm type-check
- `pytest -q tests/core/test_config.py tests/web/test_slack_channel.py tests/web/test_user_channel_model.py tests/web/services/test_channel_runtime.py tests/integration/test_auto_migration_startup.py tests/web/test_task_lease_recovery_lifecycle.py`
- `npm run test:run -- src/lib/channel-display.test.ts`
- `npm run type-check`
- targeted ESLint for changed frontend files
- `uv lock --check`

PR #1066, opened 2026-07-30
fix(channels): serialize telegram and feishu bot sync

OliverBryant:
## Summary

- serialize `_sync_bots_async` in `TelegramChannelManager` and `FeishuChannelManager` with an `asyncio.Lock`, matching the guard `SlackChannelManager` already has
- add regression coverage for concurrent sync on both managers

## Why

Channel CRUD endpoints schedule bot sync as a background task (`trigger_telegram_sync` / `trigger_feishu_sync`), so two syncs can overlap. Each one reads `self.bots`, then awaits while starting or stopping bots:

```python
current_tokens = set(self.bots.keys())
for token in current_tokens - active_tokens:
    await self._stop_bot_for_token(token)      # <-- await
for token in active_tokens - current_tokens:
    await self._start_bot_for_token(...)       # <-- await
```

The second sync resumes on the pre-start view and starts a duplicate bot for the same credential — two polling loops on one token, and the shared `self.bots` / `_bot_stop_tasks` dicts get mutated from both.

`SlackChannelManager` already holds a `_sync_lock` for exactly this reason. This ports it to the other two managers rather than removing it from Slack. Noted in #1065 as one of the divergences between the three channel implementations.

## Validation

- new `tests/web/test_channel_manager_sync_lock.py`: two concurrent `_sync_bots_async()` calls must start each bot exactly once. Verified the test **fails** without the lock (`started == ['tg-token', 'tg-token']`) and passes with it, so it exercises the real race rather than passing trivially.
- `pytest tests/web/test_channel_manager_sync_lock.py tests/web/test_telegram_message_queue.py tests/web/test_feishu_message_queue.py` — 69 passed
- ruff check / ruff format / mypy clean on the changed files

Refs #1065.

### DIFF
--- a/src/xagent/web/channels/feishu/bot.py
+++ b/src/xagent/web/channels/feishu/bot.py
@@ -735,6 +735,10 @@ class FeishuChannelManager:
     def __init__(self) -> None:
         self.bots: Dict[str, FeishuBotInstance] = {}
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
+        # Channel CRUD endpoints fire sync as a background task, so two syncs
+        # can interleave at the awaits below and act on a stale view of
+        # self.bots -- starting a duplicate bot or stopping a live one.
+        self._sync_lock = asyncio.Lock()
 
     async def start(self) -> None:
         await self._sync_bots_async()
@@ -744,38 +748,39 @@ class FeishuChannelManager:
             await self._stop_bot_for_appid(app_id)
 
     async def _sync_bots_async(self) -> None:
-        active_app_ids = set()
-        channel_info_by_appid: Dict[str, Dict] = {}
+        async with self._sync_lock:
+            active_app_ids = set()
+            channel_info_by_appid: Dict[str, Dict] = {}
 
-        try:
-            channels = await load_active_channel_configs(
-                channel_type="feishu",
-                required_config_keys=("app_id", "app_secret"),
-            )
-            for ch in channels:
-                app_id = ch.config_value("app_id")
-                app_secret = ch.config_value("app_secret")
-                if app_id and app_secret:
-                    active_app_ids.add(app_id)
-                    channel_info_by_appid[app_id] = {
-                        "app_secret": app_secret,
-                        "id": ch.channel_id,
-                        "name": ch.channel_name,
-                    }
-        except Exception as e:
-            logger.error(f"Failed to load feishu channels for sync: {e}")
-            return
+            try:
+                channels = await load_active_channel_configs(
+                    channel_type="feishu",
+                    required_config_keys=("app_id", "app_secret"),
+                )
+                for ch in channels:
+                    app_id = ch.config_value("app_id")
+                    app_secret = ch.config_value("app_secret")
+                    if app_id and app_secret:
+                        active_app_ids.add(app_id)
+                        channel_info_by_appid[app_id] = {
+                            "app_secret": app_secret,
+                            "id": ch.channel_id,
+                            "name": ch.channel_name,
+                        }
+            except Exception as e:
+                logger.error(f"Failed to load feishu channels for sync: {e}")
+                return
 
-        current_app_ids = set(self.bots.keys())
+            current_app_ids = set(self.bots.keys())
 
-        for app_id in current_app_ids - active_app_ids:
-            await self._stop_bot_for_appid(app_id)
+            for app_id in current_app_ids - active_app_ids:
+                await self._stop_bot_for_appid(app_id)
 
-        for app_id in active_app_ids - current_app_ids:
-            info = channel_info_by_appid[app_id]
-            await self._start_bot_for_appid(
-                app_id, info["app_secret"], info["id"], info["name"]
-            )
+            for app_id in active_app_ids - current_app_ids:
+                info = channel_info_by_appid[app_id]
+                await self._start_bot_for_appid(
+                    app_id, info["app_secret"], info["id"], info["name"]
+                )
 
     async def _start_bot_for_appid(
         self, app_id: str, app_secret: str, channel_id: int, channel_name: str
--- a/src/xagent/web/channels/feishu/bot.py
+++ b/src/xagent/web/channels/feishu/bot.py
@@ -736,8 +736,12 @@ class FeishuChannelManager:
         self.bots: Dict[str, FeishuBotInstance] = {}
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
         # Channel CRUD endpoints fire sync as a background task, so two syncs
-        # can interleave at the awaits below and act on a stale view of
-        # self.bots -- starting a duplicate bot or stopping a live one.
+        # can interleave. _stop_bot_for_appid awaits the shutdown drain and
+        # only removes the bot from self.bots afterwards, so a second sync
+        # entering that window sees an app_id that is still present but
+        # already being torn down: it skips starting it, the first sync
+        # completes the removal, and a re-enabled channel ends up neither
+        # running nor tracked until some later sync happens to run.
         self._sync_lock = asyncio.Lock()
 
     async def start(self) -> None:
--- a/src/xagent/web/channels/telegram/bot.py
+++ b/src/xagent/web/channels/telegram/bot.py
@@ -1603,8 +1603,12 @@ class TelegramChannelManager:
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
         self.enabled = True  # Always enabled, we load dynamically
         # Channel CRUD endpoints fire sync as a background task, so two syncs
-        # can interleave at the awaits below and act on a stale view of
-        # self.bots -- starting a duplicate bot or stopping a live one.
+        # can interleave. _stop_bot_for_token awaits the shutdown drain and
+        # only removes the bot from self.bots afterwards, so a second sync
+        # entering that window sees a token that is still present but already
+        # being torn down: it skips starting it, the first sync completes the
+        # removal, and a re-enabled channel ends up neither running nor
+        # tracked until some later sync happens to run.
         self._sync_lock = asyncio.Lock()
 
     async def start(self) -> None:
--- a/src/xagent/web/channels/telegram/bot.py
+++ b/src/xagent/web/channels/telegram/bot.py
@@ -1602,6 +1602,10 @@ class TelegramChannelManager:
         self.bots: Dict[str, TelegramBotInstance] = {}
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
         self.enabled = True  # Always enabled, we load dynamically
+        # Channel CRUD endpoints fire sync as a background task, so two syncs
+        # can interleave at the awaits below and act on a stale view of
+        # self.bots -- starting a duplicate bot or stopping a live one.
+        self._sync_lock = asyncio.Lock()
 
     async def start(self) -> None:
         await self._sync_bots_async()
@@ -1612,46 +1616,47 @@ class TelegramChannelManager:
             await self._stop_bot_for_token(token)
 
     async def _sync_bots_async(self) -> None:
-        active_tokens = set()
-        channel_info_by_token: Dict[str, Dict] = {}
+        async with self._sync_lock:
+            active_tokens = set()
+            channel_info_by_token: Dict[str, Dict] = {}
 
-        try:
-            channels = await load_active_channel_configs(
-                channel_type="telegram",
-                required_config_keys=("bot_token",),
+            try:
+                channels = await load_active_channel_configs(
+                    channel_type="telegram",
+                    required_config_keys=("bot_token",),
+                )
+                for ch in channels:
+                    token = ch.config_value("bot_token")
+                    if token:
+                        active_tokens.add(token)
+                        channel_info_by_token[token] = {
+                            "id": ch.channel_id,
+                            "name": ch.channel_name,
+                        }
+            except Exception as e:
+                logger.error(f"Failed to load user channels for sync: {e}")
+                return  # Don't try to sync if we failed to load from db
+
+            current_tokens = set(self.bots.keys())
+
+            logger.info(
+                f"Syncing telegram bots. Current active in db: {len(active_tokens)}, currently running: {len(current_tokens)}"
             )
-            for ch in channels:
-                token = ch.config_value("bot_token")
-                if token:
-                    active_tokens.add(token)
-                    channel_info_by_token[token] = {
-                        "id": ch.channel_id,
-                        "name": ch.channel_name,
-                    }
-        except Exception as e:
-            logger.error(f"Failed to load user channels for sync: {e}")
-            return  # Don't try to sync if we failed to load from db
 
-        current_tokens = set(self.bots.keys())
+            # Stop bots that are no longer active
+            for token in current_tokens - active_tokens:
+                await self._stop_bot_for_token(token)
 
-        logger.info(
-            f"Syncing telegram bots. Current active in db: {len(active_tokens)}, currently running: {len(current_tokens)}"
-        )
-
-        # Stop bots that are no longer active
-        for token in current_tokens - active_tokens:
-            await self._stop_bot_for_token(token)
-
-        # Start bots that are newly active
-        for token in active_tokens - current_tokens:
-            channel_info = channel_info_by_token.get(token, {})
-            ch_id = channel_info.get("id")
-            ch_name = channel_info.get("name")
-            await self._start_bot_for_token(
-                token,
-                int(ch_id) if ch_id is not None else None,
-                str(ch_name) if ch_name is not None else None,
-            )
+            # Start bots that are newly active
+            for token in active_tokens - current_tokens:
+                channel_info = channel_info_by_token.get(token, {})
+                ch_id = channel_info.get("id")
+                ch_name = channel_info.get("name")
+                await self._start_bot_for_token(
+                    token,
+                    int(ch_id) if ch_id is not None else None,
+                    str(ch_name) if ch_name is not None else None,
+                )
 
     async def _start_bot_for_token(
         self,
--- a/tests/web/test_channel_manager_sync_lock.py
+++ b/tests/web/test_channel_manager_sync_lock.py
@@ -1,9 +1,17 @@
-"""Concurrent channel-manager syncs must not act on a stale bot view.
+"""Concurrent channel-manager syncs must not interleave.
 
-Channel CRUD endpoints schedule ``_sync_bots_async`` as a background task, so
-two syncs can overlap. Each one reads ``self.bots`` and then awaits while
-starting or stopping bots; without serialization the second sync observes the
-pre-start view and starts a duplicate bot for the same credential.
+Channel CRUD endpoints schedule ``_sync_bots_async`` as a background task
+(``trigger_telegram_sync`` / ``trigger_feishu_sync``), so two syncs can
+overlap. The real suspension point is inside ``_stop_bot_for_token`` /
+``_stop_bot_for_appid``, which awaits the bot's shutdown drain and only then
+removes it from ``self.bots``.
+
+Without serialization that window loses a channel: sync A decides to stop a
+token that has disappeared from the database and suspends in the drain; sync B
+runs, sees the token still present in ``self.bots`` (A has not removed it yet)
+and therefore does not start it, even though the database now lists it as
+active again; A resumes and completes the removal. The token ends up neither
+in ``self.bots`` nor running, and nothing starts it until the next sync.
 """
 
 from __future__ import annotations
@@ -19,67 +27,119 @@ from xagent.web.services.channel_runtime import ChannelConfigSnapshot
 
 
 @pytest.mark.asyncio
-async def test_telegram_concurrent_sync_starts_each_bot_once(
-    monkeypatch: pytest.MonkeyPatch,
-) -> None:
+async def test_telegram_sync_does_not_interleave_with_a_pending_stop() -> None:
     manager = TelegramChannelManager()
-    snapshot = ChannelConfigSnapshot(
-        channel_id=1,
-        channel_name="Telegram Bot",
-        config_items=(("bot_token", "tg-token"),),
-    )
+    token = "tg-token"
+    # A bot is running for a token the database no longer lists.
+    manager.bots[token] = object()  # type: ignore[assignment]
+
+    active_tokens: list[str] = []
+    drain_entered = asyncio.Event()
+    release_drain = asyncio.Event()
 
     async def load_configs(**_kwargs: Any) -> tuple[ChannelConfigSnapshot, ...]:
-        # Yield so a second sync can interleave right after the DB read.
-        await asyncio.sleep(0)
-        return (snapshot,)
+        return tuple(
+            ChannelConfigSnapshot(
+                channel_id=1,
+                channel_name="Telegram Bot",
+                config_items=(("bot_token", value),),
+            )
+            for value in active_tokens
+        )
+
+    async def stop_bot(stopped_token: str) -> None:
+        # Mirrors the real method: suspend on the shutdown drain, then remove.
+        drain_entered.set()
+        await release_drain.wait()
+        manager.bots.pop(stopped_token, None)
 
     started: list[str] = []
 
-    async def start_bot(token: str, *_args: Any, **_kwargs: Any) -> None:
-        started.append(token)
-        await asyncio.sleep(0)
-        manager.bots[token] = object()  # type: ignore[assignment]
+    async def start_bot(started_token: str, *_args: Any, **_kwargs: Any) -> None:
+        if started_token in manager.bots:
+            return
+        started.append(started_token)
+        manager.bots[started_token] = object()  # type: ignore[assignment]
 
-    monkeypatch.setattr(
-        "xagent.web.channels.telegram.bot.load_active_channel_configs",
-        load_configs,
-    )
+    manager._stop_bot_for_token = stop_bot  # type: ignore[method-assign]
     manager._start_bot_for_token = start_bot  # type: ignore[method-assign]
 
-    await asyncio.gather(manager._sync_bots_async(), manager._sync_bots_async())
+    import xagent.web.channels.telegram.bot as telegram_bot
 
-    assert started == ["tg-token"]
+    original_loader = telegram_bot.load_active_channel_configs
+    telegram_bot.load_active_channel_configs = load_configs  # type: ignore[assignment]
+    try:
+        # Sync A: token absent from the DB, so it stops the running bot.
+        first = asyncio.create_task(manager._sync_bots_async())
+        await drain_entered.wait()
+
+        # The token is re-enabled while A is suspended mid-stop.
+        active_tokens.append(token)
+        second = asyncio.create_task(manager._sync_bots_async())
+        await asyncio.sleep(0)
+
+        release_drain.set()
+        await asyncio.gather(first, second)
+    finally:
+        telegram_bot.load_active_channel_configs = original_loader  # type: ignore[assignment]
+
+    # B must observe the completed stop and start the re-enabled token.
+    assert started == [token]
+    assert token in manager.bots
 
 
 @pytest.mark.asyncio
-async def test_feishu_concurrent_sync_starts_each_bot_once(
-    monkeypatch: pytest.MonkeyPatch,
-) -> None:
+async def test_feishu_sync_does_not_interleave_with_a_pending_stop() -> None:
     manager = FeishuChannelManager()
-    snapshot = ChannelConfigSnapshot(
-        channel_id=2,
-        channel_name="Feishu Bot",
-        config_items=(("app_id", "cli_123"), ("app_secret", "secret")),
-    )
+    app_id = "cli_123"
+    manager.bots[app_id] = object()  # type: ignore[assignment]
+
+    active_ids: list[str] = []
+    drain_entered = asyncio.Event()
+    release_drain = asyncio.Event()
 
     async def load_configs(**_kwargs: Any) -> tuple[ChannelConfigSnapshot, ...]:
-        await asyncio.sleep(0)
-        return (snapshot,)
+        return tuple(
+            ChannelConfigSnapshot(
+                channel_id=2,
+                channel_name="Feishu Bot",
+                config_items=(("app_id", value), ("app_secret", "secret")),
+            )
+            for value in active_ids
+        )
+
+    async def stop_bot(stopped_id: str) -> None:
+        drain_entered.set()
+        await release_drain.wait()
+        manager.bots.pop(stopped_id, None)
 
     started: list[str] = []
 
-    async def start_bot(app_id: str, *_args: Any, **_kwargs: Any) -> None:
-        started.append(app_id)
-        await asyncio.sleep(0)
-        manager.bots[app_id] = object()  # type: ignore[assignment]
+    async def start_bot(started_id: str, *_args: Any, **_kwargs: Any) -> None:
+        if started_id in manager.bots:
+            return
+        started.append(started_id)
+        manager.bots[started_id] = object()  # type: ignore[assignment]
 
-    monkeypatch.setattr(
-        "xagent.web.channels.feishu.bot.load_active_channel_configs",
-        load_configs,
-    )
+    manager._stop_bot_for_appid = stop_bot  # type: ignore[method-assign]
     manager._start_bot_for_appid = start_bot  # type: ignore[method-assign]
 
-    await asyncio.gather(manager._sync_bots_async(), manager._sync_bots_async())
+    import xagent.web.channels.feishu.bot as feishu_bot
 
-    assert started == ["cli_123"]
+    original_loader = feishu_bot.load_active_channel_configs
+    feishu_bot.load_active_channel_configs = load_configs  # type: ignore[assignment]
+    try:
+        first = asyncio.create_task(manager._sync_bots_async())
+        await drain_entered.wait()
+
+        active_ids.append(app_id)
+        second = asyncio.create_task(manager._sync_bots_async())
+        await asyncio.sleep(0)
+
+        release_drain.set()
+        await asyncio.gather(first, second)
+    finally:
+        feishu_bot.load_active_channel_configs = original_loader  # type: ignore[assignment]
+
+    assert started == [app_id]
+    assert app_id in manager.bots
--- a/tests/web/test_channel_manager_sync_lock.py
+++ b/tests/web/test_channel_manager_sync_lock.py
@@ -1,0 +1,85 @@
+"""Concurrent channel-manager syncs must not act on a stale bot view.
+
+Channel CRUD endpoints schedule ``_sync_bots_async`` as a background task, so
+two syncs can overlap. Each one reads ``self.bots`` and then awaits while
+starting or stopping bots; without serialization the second sync observes the
+pre-start view and starts a duplicate bot for the same credential.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+import pytest
+
+from xagent.web.channels.feishu.bot import FeishuChannelManager
+from xagent.web.channels.telegram.bot import TelegramChannelManager
+from xagent.web.services.channel_runtime import ChannelConfigSnapshot
+
+
+@pytest.mark.asyncio
+async def test_telegram_concurrent_sync_starts_each_bot_once(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    manager = TelegramChannelManager()
+    snapshot = ChannelConfigSnapshot(
+        channel_id=1,
+        channel_name="Telegram Bot",
+        config_items=(("bot_token", "tg-token"),),
+    )
+
+    async def load_configs(**_kwargs: Any) -> tuple[ChannelConfigSnapshot, ...]:
+        # Yield so a second sync can interleave right after the DB read.
+        await asyncio.sleep(0)
+        return (snapshot,)
+
+    started: list[str] = []
+
+    async def start_bot(token: str, *_args: Any, **_kwargs: Any) -> None:
+        started.append(token)
+        await asyncio.sleep(0)
+        manager.bots[token] = object()  # type: ignore[assignment]
+
+    monkeypatch.setattr(
+        "xagent.web.channels.telegram.bot.load_active_channel_configs",
+        load_configs,
+    )
+    manager._start_bot_for_token = start_bot  # type: ignore[method-assign]
+
+    await asyncio.gather(manager._sync_bots_async(), manager._sync_bots_async())
+
+    assert started == ["tg-token"]
+
+
+@pytest.mark.asyncio
+async def test_feishu_concurrent_sync_starts_each_bot_once(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    manager = FeishuChannelManager()
+    snapshot = ChannelConfigSnapshot(
+        channel_id=2,
+        channel_name="Feishu Bot",
+        config_items=(("app_id", "cli_123"), ("app_secret", "secret")),
+    )
+
+    async def load_configs(**_kwargs: Any) -> tuple[ChannelConfigSnapshot, ...]:
+        await asyncio.sleep(0)
+        return (snapshot,)
+
+    started: list[str] = []
+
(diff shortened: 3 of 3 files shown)


rogercloud (commented):
## Summary

This PR serializes concurrent channel-sync work by wrapping the entire body of `_sync_bots_async()` in a new per-manager `asyncio.Lock` (`self._sync_lock`) in both `TelegramChannelManager` (`src/xagent/web/channels/telegram/bot.py:1608`) and `FeishuChannelManager` (`src/xagent/web/channels/feishu/bot.py:741`). The production diff is a pure structural change — an `async with self._sync_lock:` wrap plus re-indentation, with no logic modified. A new regression test file `tests/web/test_channel_manager_sync_lock.py` fires two overlapping syncs per manager and asserts each bot is started only once.

## Approach verdict: sound (acceptable with reservations)

Wrapping the whole sync body in a single lock is the correct minimal fix. At channel-CRUD-admin-action frequency, full serialization costs nothing, and per-token locks or a generation/coalescing scheme would be unjustified complexity. Lock-release correctness was checked on every exit path: the early `return` on DB-load failure and any raised exception both unwind inside the `async with`, so there is no lock leak.

Two design-level notes that don't map to a single line:

**1. The "port the Slack pattern" rationale is unverifiable from `main`.** The PR description and issue #1065 both cite `SlackChannelManager._sync_bots_async` as the pattern being mirrored, but `SlackChannelManager` does not exist on `main` — it only exists on the still-open, unmerged #1050. The lock is correct on its own technical merits regardless of #1050's fate, so this isn't a blocker. But there is no shared base class and no shared test coupling the three channel managers, so they can silently re-diverge — which is precisely the risk #1065 tracks. Worth keeping that issue open until a structural fix (shared mixin/base, or a parametrized test over all managers) lands.

**2. The PR narrative describes the wrong failure mode.** The stated race ("two polling loops on one token") is not reachable. `_start_bot_for_token` (`src/xagent/web/channels/telegram/bot.py:1661`) and `_start_bot_for_appid` (`src/xagent/web/channels/feishu/bot.py:785`) perform a *synchronous* `if token not in self.bots: ... self.bots[token] = bot` with no `await` between the membership check and the dict write, so that check-and-set is already atomic with respect to the single-threaded event loop. Confirmed empirically: bypassing the lock but driving the real start method produces exactly one bot instance, never two.

The race the lock actually fixes is different and real: two overlapping `_sync_bots_async()` calls can load *different point-in-time DB snapshots* and interleave their stop/start decisions, leaving `self.bots` in a final state inconsistent with the DB — a bot stopped when it should be running, or vice versa. That's a lost-update / inconsistent-final-state bug, not a duplicate-instance bug. This matters beyond wording, because it's what the new test should be exercising (see Finding A).

## Findings

**A — major: the regression test proves a race in its own test double, not in production code.**
`tests/web/test_channel_manager_sync_lock.py:39-42` and `:72-75`. Both tests replace the real start method with a fake that does `started.append(...)` → `await asyncio.sleep(0)` → `manager.bots[...] = object()`. That inserted await point between "decide to start" and "record in `self.bots`" does not exist in the real methods. So without the lock the fake double-starts (verified: `['tg-token', 'tg-token']`), but the real code does not. The test therefore neither reproduces a production bug nor covers the actual race. See the inline comment for the suggested rewrite.

**B — minor: inaccurate comments in both managers (and the test docstring).**
`src/xagent/web/channels/telegram/bot.py:1607` and `src/xagent/web/channels/feishu/bot.py:739` claim the race risks "starting a duplicate bot or stopping a live one"; `tests/web/test_channel_manager_sync_lock.py:3-6` repeats the same claim. Suggest rewording to the real failure mode so a future maintainer isn't led to believe there's a missing start-side dedup guard.

**C — informational, pre-existing, not blocking: `stop()` is outside the lock's scope.**
`TelegramChannelManager.stop()` (`src/xagent/web/channels/telegram/bot.py:1613`) and `FeishuChannelManager.stop()` (`src/xagent/web/channels/feishu/bot.py:746`) snapshot and iterate `self.bots` without taking `_sync_lock`, so a bot started by an in-flight sync after the snapshot survives shutdown. Both `stop()` bodies are unchanged context in this diff, and `SlackChannelManager.stop()` behaves the same way — so this is not a regression and not this PR's job to fix. Noting it because the gap is reachable rather than academic: `trigger_telegram_sync` / `trigger_feishu_sync` tasks are not tracked or cancelled at shutdown the way other background tasks are in `src/xagent/web/app.py:1388-1407`. Worth a follow-up issue, not a change here.

## Prior review: Gemini Code Assist — WAIVED

Gemini left 4 inline comments (`feishu/bot.py:741,751`; `telegram/bot.py:1608,1619`) recommending lazy `asyncio.Lock()` construction to avoid `RuntimeError: Lock is bound to a different event loop`. **All 4 are waived as refuted — no change needed.**

- This project requires Python >= 3.11. Since 3.10, `asyncio.Lock.__init__` does nothing loop-related (it sets `self._waiters = None; self._locked = False`); the loop-binding-at-construction behavior Gemini describes was removed. Confirmed by inspecting the source of `asyncio.Lock.__init__` on the project's own interpreter.
- Empirically non-reproducible: a `Lock` constructed outside any running loop was acquired successfully across two independent `asyncio.run()` calls with no error.
- The cross-loop singleton concern also doesn't apply here — production touches the managers only from the app's single event loop, and the new tests construct a fresh manager per test.

## Verification

- `pytest tests/web/test_channel_manager_sync_lock.py tests/web/test_telegram_message_queue.py tests/web/test_feishu_message_queue.py` → **69 passed**.
- Diff reviewed as a pure structural wrap: no behavioral change to the sync logic itself.
- A dedicated simplification pass found nothing to cut — the lock usage, test structure, and comments are already minimal for the scope of the fix.

## Verdict

Ship the lock. Finding A is the one thing I'd like addressed before merge — the test as written would not catch a regression through the real code path. Findings B and C are cheap wording/follow-up items.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,85 @@
...
+    async def load_configs(**_kwargs: Any) -> tuple[ChannelConfigSnapshot, ...]:
+        # Yield so a second sync can interleave right after the DB read.
+        await asyncio.sleep(0)
+        return (snapshot,)
+
+    started: list[str] = []
+
+    async def start_bot(token: str, *_args: Any, **_kwargs: Any) -> None:
+        started.append(token)
+        await asyncio.sleep(0)
```
rogercloud: **Major - this test validates the test double, not production code.**

The fake `start_bot` inserts an `await asyncio.sleep(0)` between `started.append(token)` and `manager.bots[token] = ...`. The real `_start_bot_for_token` (`src/xagent/web/channels/telegram/bot.py:1661`) has **no** await between its `if token not in self.bots` check and the `self.bots[token] = bot` write, so that check-and-set is already atomic on the single-threaded event loop.

Verified empirically both ways: with this fake and the lock bypassed, two concurrent syncs produce `started == ['tg-token', 'tg-token']` (the race this test claims); with the **real** `_start_bot_for_token` and the lock bypassed, exactly one bot instance is created. So the duplicate-start race exists only inside this fake.

The race the lock actually fixes is different: two overlapping syncs load different point-in-time DB snapshots and interleave their stop/start decisions, leaving `self.bots` inconsistent with the DB. This test doesn't exercise that at all, and would not catch a future regression that removes the lock or introduces a genuine await point in the real start/stop path.

Suggested rewrite: drive the **real** `_start_bot_for_token` (mock only the underlying bot construction/polling), and have the patched `load_active_channel_configs` return **different snapshots** to the two concurrent `_sync_bots_async()` calls - e.g. the first sees the channel active, the second sees it removed. Then assert the final `self.bots` matches the latest snapshot rather than an interleaved intermediate state.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,85 @@
...
+
+    async def load_configs(**_kwargs: Any) -> tuple[ChannelConfigSnapshot, ...]:
+        await asyncio.sleep(0)
+        return (snapshot,)
+
+    started: list[str] = []
+
+    async def start_bot(app_id: str, *_args: Any, **_kwargs: Any) -> None:
+        started.append(app_id)
+        await asyncio.sleep(0)
```
rogercloud: Same issue as the telegram test above (Finding A): this fake inserts an artificial await between "decide to start" and "record in `self.bots`" that the real `_start_bot_for_appid` (`src/xagent/web/channels/feishu/bot.py:785`) does not have, so the double-start it guards against isn't reachable in production. Please apply the same rewrite - drive the real start method and give the two concurrent syncs divergent snapshots, asserting the final `self.bots` state converges to the latest DB view.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,85 @@
+"""Concurrent channel-manager syncs must not act on a stale bot view.
+
+Channel CRUD endpoints schedule ``_sync_bots_async`` as a background task, so
+two syncs can overlap. Each one reads ``self.bots`` and then awaits while
+starting or stopping bots; without serialization the second sync observes the
+pre-start view and starts a duplicate bot for the same credential.
```
rogercloud: **Minor (Finding B) - inaccurate failure-mode description.** "the second sync observes the pre-start view and starts a duplicate bot for the same credential" is not reachable: both real start methods do a synchronous `if key not in self.bots: ... self.bots[key] = bot` with no await in between.

Suggest describing the real race instead: two overlapping syncs can read *different point-in-time DB snapshots* and interleave their stop/start decisions, leaving `self.bots` inconsistent with the DB (a bot stopped when it should be running, or vice versa).

src/xagent/web/channels/telegram/bot.py
```diff
@@ -1602,6 +1602,10 @@ def __init__(self) -> None:
         self.bots: Dict[str, TelegramBotInstance] = {}
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
         self.enabled = True  # Always enabled, we load dynamically
+        # Channel CRUD endpoints fire sync as a background task, so two syncs
+        # can interleave at the awaits below and act on a stale view of
+        # self.bots -- starting a duplicate bot or stopping a live one.
```
rogercloud: **Minor (Finding B) - the comment names a failure mode that can't happen.** "starting a duplicate bot" is not reachable here: `_start_bot_for_token` (line 1661) does `if token not in self.bots: ... self.bots[token] = bot` with zero awaits between the check and the write, so it is atomic w.r.t. the single-threaded event loop. Verified empirically - bypassing this lock and driving the real start method yields exactly one bot instance.

Suggested wording:
```suggestion
        # Channel CRUD endpoints fire sync as a background task, so two syncs
        # can overlap, read different point-in-time DB snapshots, and interleave
        # their stop/start decisions -- leaving self.bots inconsistent with the
        # DB (a bot stopped when it should be running, or vice versa).
```
Without this, a future maintainer may conclude there's a missing start-side dedup guard and add redundant protection.

src/xagent/web/channels/feishu/bot.py
```diff
@@ -735,6 +735,10 @@ class FeishuChannelManager:
     def __init__(self) -> None:
         self.bots: Dict[str, FeishuBotInstance] = {}
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
+        # Channel CRUD endpoints fire sync as a background task, so two syncs
+        # can interleave at the awaits below and act on a stale view of
```
rogercloud: **Minor (Finding B)** - same inaccurate wording as the telegram copy. `_start_bot_for_appid` (line 785) does a synchronous `if app_id not in self.bots: ... self.bots[app_id] = bot`, so "starting a duplicate bot" is not reachable.

Suggested wording:
```suggestion
        # Channel CRUD endpoints fire sync as a background task, so two syncs
        # can overlap, read different point-in-time DB snapshots, and interleave
        # their stop/start decisions -- leaving self.bots inconsistent with the
        # DB (a bot stopped when it should be running, or vice versa).
```

src/xagent/web/channels/telegram/bot.py
```diff
@@ -1612,47 +1616,48 @@ async def stop(self) -> None:
             await self._stop_bot_for_token(token)
```
rogercloud, line 1620: **Informational (Finding C) - pre-existing, not this PR's job to fix.** `stop()` (line 1613; the diff hunk containing this comment starts at the `for token in tokens:` loop) snapshots and iterates `self.bots` without acquiring `_sync_lock`, so it can still race a concurrent `_sync_bots_async()`: a bot started by an in-flight sync *after* this snapshot is taken will survive shutdown. `FeishuChannelManager.stop()` (`src/xagent/web/channels/feishu/bot.py:746`) has the same gap, as does `SlackChannelManager.stop()`, so this is consistent, not a regression - the new lock's scope is intentionally sync-vs-sync only.

Flagging it because it's reachable rather than theoretical: `trigger_telegram_sync` / `trigger_feishu_sync` background tasks are not tracked or cancelled at app shutdown the way other tasks are in `src/xagent/web/app.py:1388-1407`. Suggest a follow-up issue rather than expanding this PR.

src/xagent/web/channels/feishu/bot.py
```diff
@@ -735,6 +735,10 @@ class FeishuChannelManager:
...
+        # self.bots -- starting a duplicate bot or stopping a live one.
+        self._sync_lock = asyncio.Lock()
```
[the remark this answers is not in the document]
OliverBryant: Not changing this — the `RuntimeError: Lock is bound to a different event loop` failure mode described here was removed in Python 3.10, and this project requires >= 3.11.

Verified independently on the project's own interpreter rather than taking it on trust:

```
python: 3.12.4
--- Lock.__init__ source ---
    def __init__(self):
        self._waiters = None
        self._locked = False

run1: True
run2 (different loop): True
```

`__init__` touches no loop state, and a `Lock` constructed outside any running loop was acquired successfully from two separate `asyncio.run()` calls. The lazy-construction guard would add a branch to every call site to defend against something that cannot happen here.

Worth noting the managers are module-level singletons, so if loop binding *did* happen at construction it would be a pre-existing problem for every `asyncio` primitive they hold, not something this PR introduces.

src/xagent/web/channels/feishu/bot.py
```diff
@@ -744,38 +748,39 @@ async def stop(self) -> None:
             await self._stop_bot_for_appid(app_id)
 
     async def _sync_bots_async(self) -> None:
-        active_app_ids = set()
-        channel_info_by_appid: Dict[str, Dict] = {}
+        async with self._sync_lock:
```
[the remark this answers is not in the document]
OliverBryant: Not changing this — the `RuntimeError: Lock is bound to a different event loop` failure mode described here was removed in Python 3.10, and this project requires >= 3.11.

Verified independently on the project's own interpreter rather than taking it on trust:

```
python: 3.12.4
--- Lock.__init__ source ---
    def __init__(self):
        self._waiters = None
        self._locked = False

run1: True
run2 (different loop): True
```

`__init__` touches no loop state, and a `Lock` constructed outside any running loop was acquired successfully from two separate `asyncio.run()` calls. The lazy-construction guard would add a branch to every call site to defend against something that cannot happen here.

Worth noting the managers are module-level singletons, so if loop binding *did* happen at construction it would be a pre-existing problem for every `asyncio` primitive they hold, not something this PR introduces.

src/xagent/web/channels/telegram/bot.py
```diff
@@ -1602,6 +1602,10 @@ def __init__(self) -> None:
         self.bots: Dict[str, TelegramBotInstance] = {}
         self._bot_stop_tasks: Dict[str, asyncio.Task[None]] = {}
         self.enabled = True  # Always enabled, we load dynamically
+        # Channel CRUD endpoints fire sync as a background task, so two syncs
+        # can interleave at the awaits below and act on a stale view of
+        # self.bots -- starting a duplicate bot or stopping a live one.
+        self._sync_lock = asyncio.Lock()
```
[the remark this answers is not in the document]
OliverBryant: Not changing this — the `RuntimeError: Lock is bound to a different event loop` failure mode described here was removed in Python 3.10, and this project requires >= 3.11.

Verified independently on the project's own interpreter rather than taking it on trust:

```
python: 3.12.4
--- Lock.__init__ source ---
    def __init__(self):
        self._waiters = None
        self._locked = False

run1: True
run2 (different loop): True
```

`__init__` touches no loop state, and a `Lock` constructed outside any running loop was acquired successfully from two separate `asyncio.run()` calls. The lazy-construction guard would add a branch to every call site to defend against something that cannot happen here.

Worth noting the managers are module-level singletons, so if loop binding *did* happen at construction it would be a pre-existing problem for every `asyncio` primitive they hold, not something this PR introduces.

src/xagent/web/channels/telegram/bot.py
```diff
@@ -1612,47 +1616,48 @@ async def stop(self) -> None:
             await self._stop_bot_for_token(token)
 
     async def _sync_bots_async(self) -> None:
-        active_tokens = set()
-        channel_info_by_token: Dict[str, Dict] = {}
+        async with self._sync_lock:
```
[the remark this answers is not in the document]
OliverBryant: Not changing this — the `RuntimeError: Lock is bound to a different event loop` failure mode described here was removed in Python 3.10, and this project requires >= 3.11.

Verified independently on the project's own interpreter rather than taking it on trust:

```
python: 3.12.4
--- Lock.__init__ source ---
    def __init__(self):
        self._waiters = None
        self._locked = False

run1: True
run2 (different loop): True
```

`__init__` touches no loop state, and a `Lock` constructed outside any running loop was acquired successfully from two separate `asyncio.run()` calls. The lazy-construction guard would add a branch to every call site to defend against something that cannot happen here.

Worth noting the managers are module-level singletons, so if loop binding *did* happen at construction it would be a pre-existing problem for every `asyncio` primitive they hold, not something this PR introduces.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: You're right, and this was a real methodology failure on my part — I verified that removing the lock made the test fail, but never checked *why* it failed. It failed inside my own double.

Confirmed your analysis by reading the real method: `_start_bot_for_token` has no await between `if token not in self.bots` and `self.bots[token] = bot`, and it carries its own existence guard, so the window my fake created does not exist in production.

The real suspension point is in `_stop_bot_for_token`:

```python
stop_task = asyncio.create_task(self._shutdown_bot_for_token(token, bot))
self._bot_stop_tasks[token] = stop_task
await drain_async_task_cancellation_safe(stop_task)   # <-- real await
```

with the `self.bots` removal happening inside `_shutdown_bot_for_token`, i.e. *after* that await. So the actual failure mode is a lost channel, not a duplicate bot: sync A stops a token that has disappeared from the DB and suspends in the drain; the channel is re-enabled; sync B runs, sees the token still in `self.bots` and skips starting it; A resumes and completes the removal. The token ends up neither tracked nor running.

Rewrote both tests around that path — they gate on an `asyncio.Event` set inside the stop double at the drain point, so the interleaving is deterministic rather than sleep-timed, and the start double keeps the real method's `if token in self.bots: return` guard so it cannot double-start on its own. Reverse-verified again, this time checking the failure reason: with the lock removed the telegram test fails on `assert started == ['tg-token']` getting `[]` — the lost-channel outcome, through the real code path.

src/xagent/web/channels/telegram/bot.py
OliverBryant: Reworded in `c19aa71e`. Both manager comments and the test docstring now describe the lost-channel outcome instead of "starting a duplicate bot or stopping a live one" — you were right that the old wording would send a future maintainer looking for a missing start-side dedup guard that is in fact already there.

src/xagent/web/channels/feishu/bot.py
OliverBryant: Reworded in `c19aa71e`. Both manager comments and the test docstring now describe the lost-channel outcome instead of "starting a duplicate bot or stopping a live one" — you were right that the old wording would send a future maintainer looking for a missing start-side dedup guard that is in fact already there.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: Reworded in `c19aa71e`. Both manager comments and the test docstring now describe the lost-channel outcome instead of "starting a duplicate bot or stopping a live one" — you were right that the old wording would send a future maintainer looking for a missing start-side dedup guard that is in fact already there.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: Rewritten alongside the telegram one in `c19aa71e` — same structure, gated on the drain window in `_stop_bot_for_appid`.

src/xagent/web/channels/telegram/bot.py
OliverBryant: Agreed on scope — filed as #1110 rather than expanded here.

I verified the reachability claim before writing it up: the `trigger_*_sync` tasks are scheduled via `background_tasks.add_task(...)` with the handle discarded, while the sandbox idle sweep / uploaded-file reconcile / metadata rebuild loops in `app.py` all keep a handle and `.cancel()` it on shutdown. So the ordering really is unconstrained.

The issue proposes both halves: take `_sync_lock` in `stop()` plus an accepting-flag under the same lock so a sync starting afterwards is a no-op, and track the sync tasks so they are cancelled with the other background work. Noted there that it wants doing in all three managers at once, so it pairs with #1065. Also called out that the Slack case is the sharper one — a leaked Socket Mode connection races the next process's gateway for the same app token.

src/xagent/web/channels/telegram/bot.py
```diff
@@ -1612,47 +1620,48 @@ async def stop(self) -> None:
...
```
rogercloud, line 1623: Minor: the lock is now held across an unbounded await chain with no timeout — `_sync_lock` → `_stop_bot_for_token` → `_shutdown_bot_for_token` → `bot.stop()` → `_stop_once` → `_drain_user_message_tasks` → `drain_async_task_cancellation_safe` / `cancel_and_drain_async_task` in `src/xagent/web/services/db_runtime.py`, none of which bounds its wait.

A hung user-message task (a swallowed `CancelledError` in a `finally`, a slow sandbox teardown) would pin `_sync_lock` indefinitely, and because of this change every subsequent CRUD-triggered sync then queues behind it forever, where previously each sync hung independently. Same structure already ships in `SlackChannelManager`, so this isn't a new class of risk and I wouldn't change it here — but worth a follow-up that wraps the drain in `asyncio.wait_for(...)` across all three managers.

src/xagent/web/channels/feishu/bot.py
```diff
@@ -744,38 +752,39 @@ async def stop(self) -> None:
...
```
rogercloud, line 755: Same as the Telegram side: `_sync_lock` is held across the unbounded `_stop_bot_for_appid` → `_shutdown_bot_for_appid` → drain chain with no timeout, so one stuck drain now serializes every later sync behind it. Not a blocker (Slack has the same shape) — flagging for the follow-up that adds `asyncio.wait_for(...)` to the drain in all three managers.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,145 @@
...
+    manager._start_bot_for_token = start_bot  # type: ignore[method-assign]
+
+    import xagent.web.channels.telegram.bot as telegram_bot
+
+    original_loader = telegram_bot.load_active_channel_configs
+    telegram_bot.load_active_channel_configs = load_configs  # type: ignore[assignment]
+    try:
+        # Sync A: token absent from the DB, so it stops the running bot.
+        first = asyncio.create_task(manager._sync_bots_async())
+        await drain_entered.wait()
```
rogercloud: Minor: this wait and the `asyncio.gather(first, second)` below (and the equivalent pair in the Feishu test) have no timeout, and there's no global pytest timeout configured for the project (`pytest-timeout` is a dev dependency but unconfigured). If a future regression stops the first sync from ever reaching the drain point, this hangs the CI job instead of failing cleanly.

`tests/web/test_telegram_message_queue.py` wraps every comparable wait in `asyncio.wait_for(..., timeout=...)`; suggest matching that here, e.g. `await asyncio.wait_for(drain_entered.wait(), timeout=5)` and `await asyncio.wait_for(asyncio.gather(first, second), timeout=5)`.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,145 @@
...
+
+    started: list[str] = []
+
+    async def start_bot(started_id: str, *_args: Any, **_kwargs: Any) -> None:
+        if started_id in manager.bots:
+            return
+        started.append(started_id)
+        manager.bots[started_id] = object()  # type: ignore[assignment]
+
+    manager._stop_bot_for_appid = stop_bot  # type: ignore[method-assign]
```
rogercloud, line 133: Minor, test coverage (pre-existing gap, not introduced here): these tests swap out `_stop_bot_for_*` / `_start_bot_for_*` wholesale. The fakes correctly reproduce the real await-then-remove *timing* — which is exactly what makes this a valid regression test now — but they never reach the real `_shutdown_bot_for_appid` or the real `_bot_stop_tasks` single-flight bookkeeping.

Telegram has separate coverage for that path (`test_manager_stop_cancellation_drains_telegram_polling_cleanup` and `test_manager_stop_is_single_flight_per_telegram_bot` in `tests/web/test_telegram_message_queue.py`, which drive the real method against a fake bot). There's no Feishu equivalent for `_shutdown_bot_for_appid`. Worth a follow-up issue rather than expanding this PR.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,145 @@
...
+    telegram_bot.load_active_channel_configs = load_configs  # type: ignore[assignment]
+    try:
+        # Sync A: token absent from the DB, so it stops the running bot.
+        first = asyncio.create_task(manager._sync_bots_async())
+        await drain_entered.wait()
+
+        # The token is re-enabled while A is suspended mid-stop.
+        active_tokens.append(token)
+        second = asyncio.create_task(manager._sync_bots_async())
+        await asyncio.sleep(0)
```
rogercloud: Nit: a single `sleep(0)` to hand control to the second sync task (here and in the Feishu test) is a fragile scheduling assumption — it only works because the `load_active_channel_configs` double has no real internal await, so the second task reaches the lock in exactly one step. Consider pinning the intent explicitly, e.g. `assert manager._sync_lock.locked()` right after this line, so the test doesn't depend on that nuance.

tests/web/test_channel_manager_sync_lock.py
```diff
@@ -0,0 +1,145 @@
...
+        started.append(started_token)
+        manager.bots[started_token] = object()  # type: ignore[assignment]
+
+    manager._stop_bot_for_token = stop_bot  # type: ignore[method-assign]
+    manager._start_bot_for_token = start_bot  # type: ignore[method-assign]
+
+    import xagent.web.channels.telegram.bot as telegram_bot
+
+    original_loader = telegram_bot.load_active_channel_configs
+    telegram_bot.load_active_channel_configs = load_configs  # type: ignore[assignment]
```
rogercloud: Nit: this hand-rolls the patch-and-restore for a module global (here and in the Feishu test). pytest's `monkeypatch.setattr(...)` fixture handles the restore for you and is the dominant idiom in this directory (107 of 192 files under `tests/web/` use it). Functionally equivalent as written — just an outlier style choice.

rogercloud (approved):
## Summary

Both `TelegramChannelManager` and `FeishuChannelManager` reconcile their running bots against the DB via `_sync_bots_async()`, which the channel CRUD endpoints in `src/xagent/web/api/channel.py` schedule as FastAPI background tasks (three call sites per manager). Two concurrent CRUD requests can therefore run two overlapping syncs that interleave their stop/start decisions over the shared `self.bots` dict. This PR adds an `asyncio.Lock` (`_sync_lock`) to each manager and wraps the whole body of `_sync_bots_async()` in it, so the load-then-reconcile sequence is atomic — a direct port of the `_sync_lock` pattern already shipped in `SlackChannelManager`. It also adds `tests/web/test_channel_manager_sync_lock.py` with one regression test per manager.

## Update summary (since last review)

Commit `c19aa71e` addresses both findings from the previous round. The regression tests were rewritten: instead of gating on an artificial `await asyncio.sleep(0)` inserted into the start-side test double (which only proved a race in the double itself), they now gate on the **real** suspension point — the drain-then-remove ordering inside `_stop_bot_for_token` / `_stop_bot_for_appid` — using `asyncio.Event`s, and the start-side fake retains the real method's existence guard so it cannot self-double-start. The in-code comments and the test module docstring were also corrected to describe the actual failure mode. No production-code changes beyond the original lock wrap.

## Approach verdict: sound

- **The race is real, but not the one the PR description states.** The description claims the sync race "starts a duplicate bot for the same credential — two polling loops on one token." That is not reachable: `_start_bot_for_token` / `_start_bot_for_appid` do a synchronous existence-check-then-insert with no `await` in between, which is already atomic on a single-threaded event loop. The reachable failure is a **lost channel**: `_stop_bot_for_token` / `_stop_bot_for_appid` `await` a shutdown drain and only pop the entry from `self.bots` *afterwards*; a second sync entering that window still sees the entry, skips starting it, and once the first sync completes the pop, the channel is neither tracked nor running. The lock fixes exactly this. The shipped comments and test docstring now describe this correctly — only the PR description is still stale (see F3).
- **Faithful port.** Lock placement matches `SlackChannelManager`: it wraps the entire method *including* the config load, which is correct — the read-then-act sequence is what must be atomic, not just the mutation. Nothing executed under the lock re-enters `_sync_bots_async`, so no deadlock. The `stop()`-outside-the-lock gap is identical in Slack and is pre-existing, not a regression.
- **Correctly scoped.** Issue #1065 lays out a smallest-useful-first plan and names "port the Slack `_sync_lock` to Telegram/Feishu" as step 2. This PR does that and nothing more — no speculative merge of the three turn-orchestration copies.

## Prior findings

| # | Finding | Status | Note |
|---|---|---|---|
| 1 | gemini-code-assist: lazy-init the `asyncio.Lock` to avoid cross-event-loop `RuntimeError` | **Waived** | Correct outcome. Both managers are module-level singletons driven by one persistent production loop, and the tests construct fresh instances per test rather than reusing a singleton across loops, so the failure mode is unreachable here. One correction to the author's reply: the cross-loop `RuntimeError` mechanism was *narrowed* in 3.10, not removed — `_LoopBoundMixin` still raises in 3.11 when the **same** lock instance is *contended* (not merely acquired) from two loops. Doesn't change the verdict; noting it so the reasoning isn't reused elsewhere as a general guarantee. |
| 2 | The regression test proved a race in its own test double, not in production code | **Fixed** | Tests now gate on the real drain-then-remove suspension point via `asyncio.Event`s, and the start-side fake keeps the real existence guard. Reverse-verified: with the lock temporarily removed, both tests fail with the correct lost-channel signature (`assert [] == ['tg-token']`), not a duplicate-start signature. |
| 3 | Comments/docstring inaccurately claimed "starting a duplicate bot or stopping a live one" | **Fixed** | The comments in `telegram/bot.py` and `feishu/bot.py` and the test module docstring now describe the lost-channel mechanism, and the description matches the actual drain/pop code — reworded *and* factually correct. |
| 4 | `stop()` sits outside `_sync_lock`, so a bot started by an in-flight sync can survive shutdown | **Waived / correctly deferred** | Filed as #1110 by the author, with matching scope: take `_sync_lock` in `stop()`, add an accepting-flag, and track/cancel the sync background tasks, across all three managers. This PR's diff does not touch any `stop()` body. Confirmed the gap still exists and that `trigger_telegram_sync` / `trigger_feishu_sync` background tasks are not registered for cancellation at shutdown, unlike the other tracked tasks in `app.py`. Right call to keep it out of this PR. |

## New findings

All findings this round are **minor** or **nit** — none block merge.

**F2 — minor — `src/xagent/web/channels/telegram/bot.py:1623` and `src/xagent/web/channels/feishu/bot.py:755`.** The lock is now held across an unbounded await chain with no timeout: `_sync_lock` → `_stop_bot_for_token` → `_shutdown_bot_for_token` → `bot.stop()` → `_stop_once` → `_drain_user_message_tasks` → `drain_async_task_cancellation_safe` / `cancel_and_drain_async_task` in `src/xagent/web/services/db_runtime.py`. None of those has a timeout. A hung user-message task (a `CancelledError` swallowed in a `finally`, a slow sandbox teardown) would pin `_sync_lock` indefinitely, and *because of this PR* every subsequent CRUD-triggered sync would then queue behind it forever — previously each sync hung independently. This is a faithful port of the same structure already shipped in `SlackChannelManager`, so it's not a new class of risk. Suggest a follow-up that wraps the drain in `asyncio.wait_for(...)` in all three managers rather than changing it here.

**F5 — minor — `tests/web/test_channel_manager_sync_lock.py:74` (and the Feishu equivalent at :133).** `await drain_entered.wait()` and `await asyncio.gather(first, second)` have no timeout, and the project has no global pytest timeout configured (`pytest-timeout` is a dev dependency but unconfigured). If a future regression prevents the first sync from reaching the drain point, these tests hang the CI job instead of failing. The sibling file `tests/web/test_telegram_message_queue.py` already wraps every comparable wait in `asyncio.wait_for(..., timeout=...)`; suggest matching that.

**F4 — minor — `tests/web/test_channel_manager_sync_lock.py:124`.** The tests replace `_stop_bot_for_token` / `_start_bot_for_token` (and the Feishu equivalents) wholesale via attribute assignment. The fakes correctly mirror the real await-then-remove *timing* — that's precisely what makes the fix to prior finding #2 valid — but they never reach the real `_shutdown_bot_for_token` / `_shutdown_bot_for_appid` or the real `_bot_stop_tasks` single-flight bookkeeping, so a regression in that bookkeeping wouldn't be caught here. Telegram is covered elsewhere (`test_manager_stop_cancellation_drains_telegram_polling_cleanup`, `test_manager_stop_is_single_flight_per_telegram_bot` in `tests/web/test_telegram_message_queue.py`, which drive the real method against a fake bot), but **there is no equivalent single-flight / cancellation test for the Feishu manager's `_shutdown_bot_for_appid`**. Pre-existing gap, not introduced here — worth a follow-up issue.

**F3 — minor (documentation) — PR description.** The PR body still says the race "starts a duplicate bot for the same credential — two polling loops on one token." That mechanism isn't reachable (see the approach verdict above), and the PR's own latest commit already corrected the in-code comments and the test docstring to describe the lost-channel mechanism — so the description now contradicts the code it ships. Please update the PR description before merge; it's what future readers land on from `git log` / blame archaeology.

**F1 — minor (informational, no action) — `src/xagent/web/channels/telegram/bot.py:1617-1620` and `src/xagent/web/channels/feishu/bot.py:750-752`.** A blank-slate pass this round independently rediscovered that `stop()` sits outside `_sync_lock` in both managers. This is the same gap as prior finding #4, already filed as #1110 — recording it only as corroboration that the issue is real and reproducible from a fresh reading, and that its shape is identical in Slack's `stop()`. No separate action item.

**F6 — nit — `tests/web/test_channel_manager_sync_lock.py:79` (and the Feishu equivalent at :137).** A single `await asyncio.sleep(0)` used to interleave the two sync tasks is a fragile scheduling assumption; it only works because the `load_active_channel_configs` double has no real internal await. Consider pinning the intent explicitly with `assert manager._sync_lock.locked()` right after scheduling the second task, so the test doesn't silently depend on scheduling nuance. Low priority.

**F7 — nit — `tests/web/test_channel_manager_sync_lock.py:70` (and the Feishu equivalent at :130).** The test monkeypatches a module global with a hand-rolled `try/finally` restore instead of pytest's `monkeypatch.setattr(...)` fixture, which is the dominant idiom in this directory (107 of 192 files under `tests/web/` use it). Functionally fine, just an outlier style choice.

```
L1: shrink tests/web/test_channel_manager_sync_lock.py — the two test functions duplicate the same harness (load_configs stub, drain/release asyncio.Events, stop/start stubs, asyncio.gather orchestration) almost line-for-line, differing only in id attribute (token vs app_id), manager type, and module path. Replace with a single @pytest.mark.parametrize over a per-manager config struct (manager factory, id value, stop/start attribute names, module path, config-item builder), ids=["telegram","feishu"] to preserve per-manager failure attribution.
net: -55 lines possible
```

## Test results

- `pytest tests/web/test_channel_manager_sync_lock.py tests/web/test_telegram_message_queue.py tests/web/test_feishu_message_queue.py` — **69 passed**, no failures, no warnings-as-errors. Independently reran in a clean review worktree at `c19aa71e`.
- Reverse-verification: temporarily removing the lock makes both new tests fail with the lost-channel signature, confirming they are not vacuous.
- `ruff check` on the changed files reports only pre-existing findings unrelated to this PR's changed lines.

## Verdict

**Approve.** The design is sound, the fix is a correctly-scoped port of an already-proven pattern, and both substantive findings from the previous round (test validity, comment accuracy) are genuinely fixed rather than reworded. Everything remaining is minor or a nit: F2 and F4 are follow-ups that should be filed against all three managers, F3 is a PR-description edit worth doing before merge, and F5/F6/F7 are optional test-hygiene improvements.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: Fixed in `ecab395e`. Both waits in both tests are now `asyncio.wait_for(..., timeout=5.0)` via a module constant, with a comment noting the reason (no global pytest timeout configured, so an unbounded wait hangs the job rather than failing). Matched the pattern in `test_telegram_message_queue.py` as you suggested.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: Fixed in `ecab395e` — added `assert manager._sync_lock.locked()` right after the `sleep(0)` in both tests. You're right that the single scheduler step only works because the loader double has no internal await; the assertion now states the required condition directly, so if that nuance ever changes the test says so instead of silently testing nothing.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: Fixed in `ecab395e` — switched both tests to `monkeypatch.setattr("...load_active_channel_configs", load_configs)`, which also let the `try/finally` and the `import ... as` go away. Cleaner and consistent with the rest of the directory.

src/xagent/web/channels/telegram/bot.py
OliverBryant: Agreed, not changing it here — filed as #1111.

Worth recording the point you made about why the lock raises the stakes: before this PR each hung sync hung independently, whereas now every later CRUD-triggered sync queues behind the stuck one, so one wedged teardown silently stops all channel configuration changes from taking effect. The lock is still the right fix (it closes the channel-loss race), but that consequence is exactly why bounding the drain is worth doing.

The issue proposes `asyncio.wait_for(...)` around the drain in all three managers, and on timeout logging loudly and proceeding with the `self.bots` removal rather than blocking — an unresponsive bot should not be able to hold the manager hostage. Noted that it pairs with #1110 and #1065 since all three touch the same teardown path across the same three managers.

src/xagent/web/channels/feishu/bot.py
OliverBryant: Agreed, not changing it here — filed as #1111.

Worth recording the point you made about why the lock raises the stakes: before this PR each hung sync hung independently, whereas now every later CRUD-triggered sync queues behind the stuck one, so one wedged teardown silently stops all channel configuration changes from taking effect. The lock is still the right fix (it closes the channel-loss race), but that consequence is exactly why bounding the drain is worth doing.

The issue proposes `asyncio.wait_for(...)` around the drain in all three managers, and on timeout logging loudly and proceeding with the `self.bots` removal rather than blocking — an unresponsive bot should not be able to hold the manager hostage. Noted that it pairs with #1110 and #1065 since all three touch the same teardown path across the same three managers.

tests/web/test_channel_manager_sync_lock.py
OliverBryant: Agreed — filed as #1112.

Confirmed the asymmetry you describe: Telegram has `test_manager_stop_cancellation_drains_telegram_polling_cleanup` and `test_manager_stop_is_single_flight_per_telegram_bot` driving the real method against a fake bot, and Feishu has no equivalent for `_shutdown_bot_for_appid` or its `_bot_stop_tasks` single-flight bookkeeping. The issue notes it is mostly a port of the two Telegram tests, and flags that it wants doing *before* #1110 and #1111 change that teardown path across all three managers.

PR #1067, opened 2026-07-30
feat: add Google Analytics (GA4) MCP connector

yiboyasss:
## Summary
- Register `google-analytics` as a built-in OAuth MCP connector (`builtin_mcp_registry.py` + seed migration), reusing the shared `google` OAuth provider with the `analytics.readonly` scope — no new client credentials needed
- New `xagent.web.tools.mcp.google_analytics` MCP module with three tools:
  - `google_analytics_list_properties` — list accessible GA4 properties via the Admin API (handles pagination)
  - `google_analytics_get_metadata` — list a property's available dimensions/metrics
  - `google_analytics_run_report` — Data API `runReport` with multi-date-range comparison, dimension filters, ordering and row limits
- Deployment prerequisite: enable the Google Analytics Data API and Admin API in the Google Cloud project, and add the `analytics.readonly` scope to the OAuth consent screen

## Test plan
- [x] `pytest tests/web/tools/test_google_analytics_mcp.py tests/alembic/test_20260730_seed_google_analytics_mcp_app.py tests/templates/test_manager.py` — all passing
- [x] Manual end-to-end against a real GA4 property: OAuth connect, list properties, metadata lookup, and a period-over-period channel report all working

### DIFF
--- a/src/xagent/web/tools/mcp/google_analytics.py
+++ b/src/xagent/web/tools/mcp/google_analytics.py
@@ -147,14 +147,18 @@ def google_analytics_list_properties() -> str:
                     params=params,
                 )
             )
-            for account in result.get("accountSummaries", []):
+            for account in result.get("accountSummaries") or []:
                 if not isinstance(account, dict):
                     continue
                 account_name = account.get("displayName")
-                for prop in account.get("propertySummaries", []) or []:
+                for prop in account.get("propertySummaries") or []:
                     if not isinstance(prop, dict):
                         continue
-                    resource_name = prop.get("property", "")
+                    resource_name = prop.get("property")
+                    if not isinstance(resource_name, str) or not resource_name:
+                        # No usable resource name -> no property_id to offer;
+                        # skip rather than emitting a garbage empty id.
+                        continue
                     properties.append(
                         {
                             "account_name": account_name,
@@ -187,8 +191,8 @@ def google_analytics_get_metadata(property_id: str) -> str:
             )
         )
         return _success(
-            dimensions=result.get("dimensions", []),
-            metrics=result.get("metrics", []),
+            dimensions=result.get("dimensions") or [],
+            metrics=result.get("metrics") or [],
         )
     except Exception as e:
         logger.error(f"Error getting Google Analytics metadata for {property_id}: {e}")
@@ -247,10 +251,10 @@ def google_analytics_run_report(
             )
         )
         return _success(
-            dimension_headers=result.get("dimensionHeaders", []),
-            metric_headers=result.get("metricHeaders", []),
-            rows=result.get("rows", []),
-            row_count=result.get("rowCount", 0),
+            dimension_headers=result.get("dimensionHeaders") or [],
+            metric_headers=result.get("metricHeaders") or [],
+            rows=result.get("rows") or [],
+            row_count=result.get("rowCount") or 0,
         )
     except Exception as e:
         logger.error(f"Error running Google Analytics report for {property_id}: {e}")
--- a/src/xagent/web/tools/mcp/google_analytics.py
+++ b/src/xagent/web/tools/mcp/google_analytics.py
@@ -21,6 +21,11 @@ DATA_API_BASE_URL = "https://analyticsdata.googleapis.com/v1beta"
 ADMIN_API_BASE_URL = "https://analyticsadmin.googleapis.com/v1beta"
 DEFAULT_TIMEOUT_SECONDS = 30
 MAX_ACCOUNT_SUMMARY_PAGES = 20
+# A 5-dimension + 5-metric row serializes to ~275 chars through _success(); at
+# RUN_REPORT_MAX_LIMIT rows that's comfortably under the platform's 51200-char
+# MCP output truncation threshold even for wide reports.
+RUN_REPORT_DEFAULT_LIMIT = 100
+RUN_REPORT_MAX_LIMIT = 150
 
 _PROPERTY_ID_PATTERN = re.compile(r"^[0-9]+\Z")
 
@@ -204,7 +209,11 @@ def _project_metadata_entries(entries: list[Any], search: str) -> list[dict[str,
     for entry in entries:
         if not isinstance(entry, dict):
             continue
-        api_name = str(entry.get("apiName") or "")
+        api_name = entry.get("apiName")
+        if not isinstance(api_name, str) or not api_name:
+            # No usable apiName -> nothing for the LLM to pass back into
+            # run_report; skip rather than emitting a garbage empty name.
+            continue
         ui_name = str(entry.get("uiName") or "")
         if needle and needle not in api_name.lower() and needle not in ui_name.lower():
             continue
@@ -257,7 +266,7 @@ def google_analytics_run_report(
     dimension_filter: dict[str, Any] | None = None,
     metric_filter: dict[str, Any] | None = None,
     order_bys: list[dict[str, Any]] | None = None,
-    limit: int = 1000,
+    limit: int = RUN_REPORT_DEFAULT_LIMIT,
     offset: int = 0,
 ) -> str:
     """
@@ -280,22 +289,29 @@ def google_analytics_run_report(
     metric_filter: a raw GA4 FilterExpression dict applied to metric values
       (e.g. "sessions > 100"). Optional.
     order_bys: a raw GA4 OrderBy list, e.g. to sort by a metric descending.
-    limit: max rows to return (default 1000).
+    limit: max rows to return (default 100, max 150 — keeps a wide report's
+      serialized response under the MCP output size limit).
     offset: row offset for paging — if row_count in the response exceeds
       limit, call again with offset=limit, then offset=2*limit, and so on.
     """
     try:
         normalized_property_id = _normalize_property_id(property_id)
+        if not metrics:
+            raise ValueError("metrics must contain at least one metric name")
         if not date_ranges:
             raise ValueError("date_ranges must contain at least one date range")
         if len(date_ranges) > 4:
             raise ValueError("GA4 allows at most 4 date_ranges per report")
+        if limit < 1 or limit > RUN_REPORT_MAX_LIMIT:
+            raise ValueError(f"limit must be between 1 and {RUN_REPORT_MAX_LIMIT}")
+        if offset < 0:
+            raise ValueError("offset must be >= 0")
         body: dict[str, Any] = {
             "metrics": [{"name": m} for m in metrics],
             "dateRanges": [_date_range_body(dr) for dr in date_ranges],
             # Always sent explicitly: GA4's default would otherwise return up
             # to 10k rows in one response.
-            "limit": str(max(1, limit)),
+            "limit": str(limit),
         }
         if dimensions:
             body["dimensions"] = [{"name": d} for d in dimensions]
--- a/src/xagent/web/tools/mcp/google_analytics.py
+++ b/src/xagent/web/tools/mcp/google_analytics.py
@@ -117,9 +117,18 @@ def _require_dict_result(result: Any) -> dict[str, Any]:
 
 
 def _date_range_body(date_range: dict[str, Any]) -> dict[str, Any]:
+    # FastMCP's signature validation guarantees a dict, but not its keys — an
+    # empty dict or misnamed keys (e.g. camelCase "startDate") would otherwise
+    # sail through and reach the API as two nulls. Fail here with a message
+    # the LLM can self-correct from.
+    if not date_range.get("start_date") or not date_range.get("end_date"):
+        raise ValueError(
+            'Each date range needs "start_date" and "end_date" (YYYY-MM-DD or '
+            'GA4 relative terms like "7daysAgo"/"today")'
+        )
     body = {
-        "startDate": date_range.get("start_date"),
-        "endDate": date_range.get("end_date"),
+        "startDate": date_range["start_date"],
+        "endDate": date_range["end_date"],
     }
     if date_range.get("name"):
         body["name"] = date_range["name"]
--- a/src/xagent/web/tools/mcp/google_analytics.py
+++ b/src/xagent/web/tools/mcp/google_analytics.py
@@ -1,0 +1,261 @@
+import json
+import logging
+import os
+import re
+from typing import Any
+
+import requests
+from mcp.server.fastmcp import FastMCP
+
+from .utils import setup_proxy_env
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger("google-analytics-mcp")
+
+# Ensure standard proxy environment variables are set to prevent hanging requests
+setup_proxy_env()
+
+mcp = FastMCP("google-analytics-mcp")
+
+DATA_API_BASE_URL = "https://analyticsdata.googleapis.com/v1beta"
+ADMIN_API_BASE_URL = "https://analyticsadmin.googleapis.com/v1beta"
+DEFAULT_TIMEOUT_SECONDS = 30
+MAX_ACCOUNT_SUMMARY_PAGES = 20
+
+_PROPERTY_ID_PATTERN = re.compile(r"^[0-9]+\Z")
+
+
+def _success(**payload: Any) -> str:
+    return json.dumps({"status": "success", **payload}, ensure_ascii=False)
+
+
+def _error(message: str) -> str:
+    return json.dumps({"status": "error", "message": message}, ensure_ascii=False)
+
+
+def _normalize_property_id(property_id: str) -> str:
+    """Accept either a bare numeric id ("123456") or a resource name
+    ("properties/123456") and return the bare numeric id.
+
+    Rejects anything else rather than silently sanitizing it, since this
+    value is interpolated directly into a URL path.
+    """
+    raw = str(property_id)
+    if raw.startswith("properties/"):
+        raw = raw[len("properties/") :]
+    if not _PROPERTY_ID_PATTERN.match(raw):
+        raise ValueError(
+            "property_id must be numeric (optionally prefixed 'properties/')"
+        )
+    return raw
+
+
+def _headers() -> dict[str, str]:
+    access_token = os.environ.get("GOOGLE_ACCESS_TOKEN")
+    if not access_token:
+        raise ValueError("GOOGLE_ACCESS_TOKEN environment variable is missing")
+    return {
+        "Authorization": f"Bearer {access_token}",
+        "Content-Type": "application/json",
+    }
+
+
+def _extract_error_detail(response: requests.Response) -> str | None:
+    """Pull the human-readable message out of a Google API error body
+    (``{"error": {"code": ..., "message": ..., "status": ...}}``).
+    Returns None if the body isn't in the expected shape, so the caller can
+    fall back to the raw response text.
+    """
+    try:
+        payload = response.json()
+    except ValueError:
+        return None
+    if not isinstance(payload, dict):
+        return None
+    error = payload.get("error")
+    if not isinstance(error, dict):
+        return None
+    message = error.get("message")
+    return message if isinstance(message, str) and message else None
+
+
+def _request(
+    method: str,
+    url: str,
+    *,
+    params: dict[str, Any] | None = None,
+    body: dict[str, Any] | None = None,
+) -> Any:
+    response = requests.request(
+        method=method,
+        url=url,
+        headers=_headers(),
+        params=params,
+        json=body,
+        timeout=DEFAULT_TIMEOUT_SECONDS,
+    )
+    try:
+        response.raise_for_status()
+    except requests.HTTPError as exc:
+        message = str(exc)
+        detail = _extract_error_detail(response)
+        if detail is None:
+            detail = response.text.strip()
+        if detail:
+            message = f"{message} - {detail}"
+        raise RuntimeError(message) from exc
+
+    if response.status_code == 204 or not response.content:
+        return {}
+    return response.json()
+
+
+def _require_dict_result(result: Any) -> dict[str, Any]:
+    if not isinstance(result, dict):
+        raise ValueError("Unexpected response format from Google Analytics API")
+    return result
+
+
+def _date_range_body(date_range: dict[str, Any]) -> dict[str, Any]:
+    body = {
+        "startDate": date_range.get("start_date"),
+        "endDate": date_range.get("end_date"),
+    }
+    if date_range.get("name"):
+        body["name"] = date_range["name"]
+    return body
+
+
+@mcp.tool()
+def google_analytics_list_properties() -> str:
+    """
+    List the GA4 accounts and properties accessible to the connected account.
+    Use this first to discover which property_id values are available for
+    google_analytics_run_report / google_analytics_get_metadata.
+    """
+    try:
+        properties: list[dict[str, Any]] = []
+        page_token: str | None = None
+        for _ in range(MAX_ACCOUNT_SUMMARY_PAGES):
+            params: dict[str, Any] = {"pageSize": 200}
+            if page_token:
+                params["pageToken"] = page_token
+            result = _require_dict_result(
+                _request(
+                    "GET",
+                    f"{ADMIN_API_BASE_URL}/accountSummaries",
+                    params=params,
+                )
+            )
+            for account in result.get("accountSummaries", []):
+                if not isinstance(account, dict):
+                    continue
+                account_name = account.get("displayName")
+                for prop in account.get("propertySummaries", []) or []:
+                    if not isinstance(prop, dict):
+                        continue
+                    resource_name = prop.get("property", "")
+                    properties.append(
+                        {
+                            "account_name": account_name,
+                            "property_id": resource_name.rsplit("/", 1)[-1],
+                            "property_display_name": prop.get("displayName"),
+                        }
+                    )
+            page_token = result.get("nextPageToken")
+            if not page_token:
+                break
+        return _success(properties=properties)
+    except Exception as e:
+        logger.error(f"Error listing Google Analytics properties: {e}")
+        return _error(str(e))
+
+
+@mcp.tool()
+def google_analytics_get_metadata(property_id: str) -> str:
+    """
+    List the dimensions and metrics available to query for a GA4 property.
+    Use this before google_analytics_run_report if you're unsure which
+    dimension/metric names are valid for this property.
+    """
+    try:
+        normalized_property_id = _normalize_property_id(property_id)
+        result = _require_dict_result(
+            _request(
+                "GET",
+                f"{DATA_API_BASE_URL}/properties/{normalized_property_id}/metadata",
+            )
+        )
+        return _success(
+            dimensions=result.get("dimensions", []),
+            metrics=result.get("metrics", []),
+        )
+    except Exception as e:
+        logger.error(f"Error getting Google Analytics metadata for {property_id}: {e}")
+        return _error(str(e))
+
+
+@mcp.tool()
+def google_analytics_run_report(
+    property_id: str,
+    metrics: list[str],
+    date_ranges: list[dict[str, str]],
+    dimensions: list[str] | None = None,
+    dimension_filter: dict[str, Any] | None = None,
+    order_bys: list[dict[str, Any]] | None = None,
+    limit: int | None = None,
+) -> str:
+    """
+    Run a GA4 report: quantify performance by any combination of dimensions
+    (e.g. channel, campaign, landing page, segment) over one or more date
+    ranges.
+
+    metrics: metric names, e.g. ["sessions", "conversions", "totalRevenue"].
+    date_ranges: one dict per period to compare, each with "start_date" and
+      "end_date" (YYYY-MM-DD, or GA4 relative terms like "7daysAgo"/"today"),
+      and an optional "name" to label the period (e.g. "current", "previous")
+      — pass two date_ranges to compare a period against a prior one in a
+      single call; the response's rows are tagged with the matching name.
+    dimensions: dimension names, e.g. ["sessionDefaultChannelGroup",
+      "landingPagePlusQueryString"]. Use google_analytics_get_metadata to
+      discover valid names for this property.
+    dimension_filter: a raw GA4 FilterExpression dict, for narrowing results
+      (e.g. to one campaign or landing page). Optional.
+    order_bys: a raw GA4 OrderBy list, e.g. to sort by a metric descending.
+    limit: max rows to return.
+    """
+    try:
+        normalized_property_id = _normalize_property_id(property_id)
+        body: dict[str, Any] = {
+            "metrics": [{"name": m} for m in metrics],
+            "dateRanges": [_date_range_body(dr) for dr in date_ranges],
+        }
+        if dimensions:
+            body["dimensions"] = [{"name": d} for d in dimensions]
+        if dimension_filter:
+            body["dimensionFilter"] = dimension_filter
+        if order_bys:
+            body["orderBys"] = order_bys
+        if limit is not None:
+            body["limit"] = str(limit)
+
+        result = _require_dict_result(
+            _request(
+                "POST",
+                f"{DATA_API_BASE_URL}/properties/{normalized_property_id}:runReport",
+                body=body,
+            )
+        )
+        return _success(
+            dimension_headers=result.get("dimensionHeaders", []),
+            metric_headers=result.get("metricHeaders", []),
+            rows=result.get("rows", []),
+            row_count=result.get("rowCount", 0),
+        )
+    except Exception as e:
+        logger.error(f"Error running Google Analytics report for {property_id}: {e}")
+        return _error(str(e))
+
+
+if __name__ == "__main__":
+    mcp.run()
--- a/src/xagent/web/tools/mcp/google_analytics.py
+++ b/src/xagent/web/tools/mcp/google_analytics.py
@@ -7,6 +7,7 @@ from typing import Any
 import requests
 from mcp.server.fastmcp import FastMCP
 
+from ....config import get_tool_max_output_length
 from .utils import setup_proxy_env
 
 logging.basicConfig(level=logging.INFO)
@@ -21,9 +22,13 @@ DATA_API_BASE_URL = "https://analyticsdata.googleapis.com/v1beta"
 ADMIN_API_BASE_URL = "https://analyticsadmin.googleapis.com/v1beta"
 DEFAULT_TIMEOUT_SECONDS = 30
 MAX_ACCOUNT_SUMMARY_PAGES = 20
-# A 5-dimension + 5-metric row serializes to ~275 chars through _success(); at
-# RUN_REPORT_MAX_LIMIT rows that's comfortably under the platform's 51200-char
-# MCP output truncation threshold even for wide reports.
+# The limit bounds row *count*, not bytes: a 5-dimension + 5-metric row with
+# short (~11-char) values serializes to ~275 chars through _success(), which
+# keeps RUN_REPORT_MAX_LIMIT rows comfortably under the platform's 51200-char
+# MCP output truncation threshold — but several long-valued dimensions (e.g.
+# full URLs) at once can still cross it. google_analytics_run_report guards
+# the actual serialized size and trims rows if needed, same as
+# google_analytics_list_properties does for its own response.
 RUN_REPORT_DEFAULT_LIMIT = 100
 RUN_REPORT_MAX_LIMIT = 150
 
@@ -331,12 +336,41 @@ def google_analytics_run_report(
                 body=body,
             )
         )
-        return _success(
-            dimension_headers=result.get("dimensionHeaders") or [],
-            metric_headers=result.get("metricHeaders") or [],
-            rows=result.get("rows") or [],
-            row_count=result.get("rowCount") or 0,
+        dimension_headers = result.get("dimensionHeaders") or []
+        metric_headers = result.get("metricHeaders") or []
+        rows = result.get("rows") or []
+        row_count = result.get("rowCount") or 0
+
+        # RUN_REPORT_MAX_LIMIT bounds row count, not bytes: several long
+        # dimension values (e.g. full URLs) at once can still cross the
+        # platform's output truncation threshold. Halve the returned rows
+        # until the serialized response fits, mirroring the truncated-flag
+        # pattern in google_analytics_list_properties.
+        max_output_length = get_tool_max_output_length()
+        original_row_count = len(rows)
+        response = _success(
+            dimension_headers=dimension_headers,
+            metric_headers=metric_headers,
+            rows=rows,
+            row_count=row_count,
+            truncated=False,
         )
+        while len(response) > max_output_length and len(rows) > 1:
+            rows = rows[: len(rows) // 2]
+            response = _success(
+                dimension_headers=dimension_headers,
+                metric_headers=metric_headers,
+                rows=rows,
+                row_count=row_count,
+                truncated=True,
+            )
+        if len(rows) < original_row_count:
+            logger.warning(
+                f"Google Analytics run_report response trimmed from "
+                f"{original_row_count} to {len(rows)} rows to stay under "
+                f"the {max_output_length}-char output limit"
+            )
+        return response
     except Exception as e:
         # !r escapes embedded newlines, keeping a crafted property_id from
         # injecting fake log lines.
--- a/src/xagent/web/tools/mcp/google_analytics.py
+++ b/src/xagent/web/tools/mcp/google_analytics.py
@@ -100,7 +100,10 @@ def _request(
         message = str(exc)
         detail = _extract_error_detail(response)
         if detail is None:
-            detail = response.text.strip()
+            # Cap the raw-body fallback: an upstream HTML error page can be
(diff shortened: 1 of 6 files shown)


src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,265 @@
...
+      (e.g. to one campaign or landing page). Optional.
+    order_bys: a raw GA4 OrderBy list, e.g. to sort by a metric descending.
+    limit: max rows to return.
+    """
+    try:
+        normalized_property_id = _normalize_property_id(property_id)
+        body: dict[str, Any] = {
+            "metrics": [{"name": m} for m in metrics],
+            "dateRanges": [_date_range_body(dr) for dr in date_ranges],
+        }
```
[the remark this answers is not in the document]
yiboyasss: Verified empirically rather than assumed: invoking this tool through FastMCP's tool manager with `metrics="sessions"` (a scalar) or a bare-dict `date_ranges` is rejected with a `ToolError` **before the function body runs** — FastMCP validates arguments against the signature via pydantic (`list[str]` / `list[dict[str, str]]`), and pydantic v2 does not iterate strings into lists, so the char-by-char scenario cannot reach this code. Silent coercion (e.g. `True` → `["True"]`) would send garbage metric names to the API instead of surfacing a clean validation error, so leaving this to the framework layer is intentional.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def _date_range_body(date_range: dict[str, Any]) -> dict[str, Any]:
+    body = {
+        "startDate": date_range.get("start_date"),
+        "endDate": date_range.get("end_date"),
+    }
+    if date_range.get("name"):
+        body["name"] = date_range["name"]
+    return body
```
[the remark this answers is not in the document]
yiboyasss: Partially applied in 3c3a5ef3. The `isinstance` half is unreachable — FastMCP's signature validation (pydantic, `list[dict[str, str]]`) already rejects non-dict elements before the body runs. But the suggestion did point at a real gap the signature can't catch: an **empty dict** or **misnamed keys** (an LLM plausibly sends camelCase `startDate`/`endDate`, since that's what the GA API itself uses) passes `dict[str, str]` validation and previously sailed through as two nulls for the API to reject remotely. `_date_range_body` now validates `start_date`/`end_date` up front and fails with a message the LLM can self-correct from, before any network call — covered by a test asserting no request is issued for `{}` and camelCase-keyed ranges.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+    except Exception as e:
+        logger.error(f"Error listing Google Analytics properties: {e}")
+        return _error(str(e))
+
+
+@mcp.tool()
+def google_analytics_get_metadata(property_id: str) -> str:
+    """
+    List the dimensions and metrics available to query for a GA4 property.
+    Use this before google_analytics_run_report if you're unsure which
```
rogercloud: **Major:** `google_analytics_get_metadata` returns the entire GA4 metadata payload (every dimension and metric, typically 200-350+ entries each carrying `apiName`/`uiName`/`description`/`category`) as a single JSON string. MCP tool results pass through the platform output filter, and `src/xagent/core/tools/adapters/vibe/output_filter.py:235-239` truncates a single oversized string leaf by hard-cutting it at the default 51200-char threshold and appending a non-JSON suffix. A typical GA4 metadata document serializes to roughly 50-140KB, so the realistic outcome is that the LLM receives syntactically broken JSON.

Suggest projecting each entry down to `apiName`/`uiName`/`category` and/or adding an optional search/category filter parameter so the response stays comfortably under the threshold.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+      — pass two date_ranges to compare a period against a prior one in a
+      single call; the response's rows are tagged with the matching name.
+    dimensions: dimension names, e.g. ["sessionDefaultChannelGroup",
+      "landingPagePlusQueryString"]. Use google_analytics_get_metadata to
+      discover valid names for this property.
+    dimension_filter: a raw GA4 FilterExpression dict, for narrowing results
+      (e.g. to one campaign or landing page). Optional.
+    order_bys: a raw GA4 OrderBy list, e.g. to sort by a metric descending.
+    limit: max rows to return.
+    """
```
rogercloud, line 301: **Major:** when `limit` is `None`, the key is never added to the request body, so GA4's `runReport` default applies and can return up to 10,000 rows in one response. There is no `offset` parameter anywhere in the tool, so even though `row_count` is returned (implying more rows may exist), the caller has no mechanism to retrieve them.

Suggest defaulting `limit` to a sane value (100-1000) when omitted, and adding an `offset` parameter wired to GA4's request `offset` field.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+def _require_dict_result(result: Any) -> dict[str, Any]:
+    if not isinstance(result, dict):
+        raise ValueError("Unexpected response format from Google Analytics API")
+    return result
+
+
+def _date_range_body(date_range: dict[str, Any]) -> dict[str, Any]:
+    # FastMCP's signature validation guarantees a dict, but not its keys — an
+    # empty dict or misnamed keys (e.g. camelCase "startDate") would otherwise
+    # sail through and reach the API as two nulls. Fail here with a message
```
rogercloud, line 139: **Minor:** `_date_range_body` reads `start_date`/`end_date` with no shape or key checks, so a caller passing camelCase (`startDate`/`endDate`) or an empty dict silently produces `{"startDate": None, "endDate": None}`, which GA4 rejects with an opaque 400 instead of a clear local validation error.

Suggest guarding with something like `if not dr or not isinstance(dr, dict)` plus a check that both dates are present, and rejecting an empty `date_ranges` list. GA4 also caps `date_ranges` at 4 entries — worth enforcing locally.

This overlaps with the existing gemini-code-assist review on this commit, which flagged the same underlying validation gap — independent confirmation, not a duplicate.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+    List the GA4 accounts and properties accessible to the connected account.
+    Use this first to discover which property_id values are available for
+    google_analytics_run_report / google_analytics_get_metadata.
+    """
+    try:
+        properties: list[dict[str, Any]] = []
+        page_token: str | None = None
+        for _ in range(MAX_ACCOUNT_SUMMARY_PAGES):
+            params: dict[str, Any] = {"pageSize": 200}
+            if page_token:
```
rogercloud, line 167: **Minor:** `result.get("accountSummaries", [])` lacks the `or []` fallback that the inner `propertySummaries` access at line 154 correctly has. An explicit `"accountSummaries": null` in the response therefore raises a `TypeError` that the broad `except Exception` below (lines 169-171) masks into a generic error message.

Fix: `result.get("accountSummaries") or []`.

This matches the existing gemini-code-assist review comment on this exact line — independent confirmation.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+            result = _require_dict_result(
+                _request(
+                    "GET",
+                    f"{ADMIN_API_BASE_URL}/accountSummaries",
+                    params=params,
+                )
+            )
+            for account in result.get("accountSummaries") or []:
+                if not isinstance(account, dict):
+                    continue
```
rogercloud, line 178: **Minor:** `prop.get("property", "").rsplit("/", 1)[-1]` assumes the default kicks in, but `dict.get` only applies the default when the key is *absent*; an explicit `"property": null` returns `None` and `.rsplit()` then raises `AttributeError`, masked by the same broad handler above.

The sibling connector already guards this exact pattern with an `isinstance(name, str)` check at `src/xagent/web/tools/mcp/google_ads.py:164-166` — please mirror it.

This matches the existing gemini-code-assist review comment on this exact line — independent confirmation.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+        "startDate": date_range["start_date"],
+        "endDate": date_range["end_date"],
+    }
+    if date_range.get("name"):
+        body["name"] = date_range["name"]
+    return body
+
+
+@mcp.tool()
+def google_analytics_list_properties() -> str:
```
rogercloud, line 156: **Minor:** when `MAX_ACCOUNT_SUMMARY_PAGES` (defined at line 23) is exhausted, the function returns whatever it accumulated with no signal that more pages existed, so the caller cannot distinguish "that's all your properties" from "truncated".

Suggest returning a `truncated: true` / `has_more` flag (and ideally the next page token) alongside the results.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+        raise RuntimeError(message) from exc
+
+    if response.status_code == 204 or not response.content:
+        return {}
+    return response.json()
+
+
+def _require_dict_result(result: Any) -> dict[str, Any]:
+    if not isinstance(result, dict):
+        raise ValueError("Unexpected response format from Google Analytics API")
```
rogercloud, line 132: **Minor:** `_require_dict_result` is missing its explanatory docstring. The sibling implementation at `src/xagent/web/tools/mcp/google_ads.py:112-119` documents that an *empty* dict is a legitimate, valid API response and must not be treated as an error. Carrying that comment over is cheap insurance against someone "fixing" this into a falsy check later.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+    date_ranges: one dict per period to compare, each with "start_date" and
+      "end_date" (YYYY-MM-DD, or GA4 relative terms like "7daysAgo"/"today"),
+      and an optional "name" to label the period (e.g. "current", "previous")
+      — pass two date_ranges to compare a period against a prior one in a
+      single call; the response's rows are tagged with the matching name.
+    dimensions: dimension names, e.g. ["sessionDefaultChannelGroup",
+      "landingPagePlusQueryString"]. Use google_analytics_get_metadata to
+      discover valid names for this property.
+    dimension_filter: a raw GA4 FilterExpression dict, for narrowing results
+      (e.g. to one campaign or landing page). Optional.
```
rogercloud, line 293: **Minor:** `dimensionFilter` is supported but there's no `metricFilter` counterpart, so a metric-side predicate such as "sessions > 100" cannot be expressed even though the PR description mentions filtering generally. GA4's `runReport` supports `metricFilter` with the same `FilterExpression` shape; adding it is a small, symmetric addition.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+        json=body,
+        timeout=DEFAULT_TIMEOUT_SECONDS,
+    )
+    try:
+        response.raise_for_status()
+    except requests.HTTPError as exc:
+        message = str(exc)
+        detail = _extract_error_detail(response)
+        if detail is None:
+            detail = response.text.strip()
```
rogercloud: **Nit:** uncapped `response.text.strip()` in the error path — a large upstream HTML error page gets embedded whole into the error message and context. This same unbounded pattern already exists at `src/xagent/web/tools/mcp/google_ads.py:143`, so this PR is continuing rather than introducing it — but `src/xagent/web/tools/mcp/meta_graph.py:88-92` shows the better precedent of capping at 1000 chars.

rogercloud (commented):
## Summary

Adds a built-in Google Analytics (GA4) MCP connector. The new module `src/xagent/web/tools/mcp/google_analytics.py` exposes three tools — `google_analytics_list_properties` (paginated `accountSummaries` walk over the Admin API), `google_analytics_get_metadata` (Data API `getMetadata`), and `google_analytics_run_report` (Data API `runReport` with metrics/dimensions/multi-date-range/dimensionFilter/orderBys support) — all authenticated through the existing shared `google` OAuth provider with the added `analytics.readonly` scope, so no new client credentials are introduced. Supporting changes: a `google-analytics` entry in `builtin_mcp_registry.py` and a seed Alembic migration modeled on the google-ads one, plus unit tests for the tools and the migration.

## Approach review: sound

The core design decision — reusing the shared `google` OAuth provider rather than standing up a new provider — is correct and safe. Tokens are keyed by `(user, app_id)` rather than `(user, provider)`, so connecting `google-analytics` cannot collide with the existing gmail/drive/calendar tokens; scopes are merged through `_merge_oauth_scopes`; and expired tokens are refreshed before MCP launch. The seed migration follows the sibling `google_ads` migration pattern exactly, with idempotency guards, and only executes on a fresh empty database, so there is no double-seed or clobber risk. No design blockers.

Three design-level notes that no single line owns:

**D1 — Undocumented operational prerequisites.** The PR description itself calls out real setup steps (enabling `analyticsadmin.googleapis.com` and `analyticsdata.googleapis.com` in GCP, and registering `analytics.readonly` as a sensitive scope on the OAuth consent screen), but none of this is written down anywhere in the repo — not `example.env`, not the README, not docs. The sibling google-ads connector documents its equivalent prerequisite inline at `example.env:374-380`; please follow that precedent so operators aren't left to rediscover it from a stale PR description.

**D2 — Second copy of the same Google HTTP helpers.** `_success`/`_error` in `src/xagent/web/tools/mcp/google_analytics.py:28-33` are byte-for-byte identical to `src/xagent/web/tools/mcp/google_ads.py:26-31`, and the other four helpers (`_headers`, `_extract_error_detail`, `_require_dict_result`, `_request`) are independently duplicated as well. The repo already has a working precedent for sharing exactly this kind of thing: `src/xagent/web/tools/mcp/meta_graph.py`, imported by both `facebook.py` and `instagram.py`. Not blocking this PR, but this is now copy #2 of the same ~6 helpers; extracting a shared `google_api.py` before the next Google connector arrives would stop it becoming copy #3.

**D3 — Inconsistent parameter convention in `run_report`.** At `src/xagent/web/tools/mcp/google_analytics.py:199-207`, `date_ranges` takes snake_case keys (`start_date`/`end_date`) that are translated internally, while `dimension_filter` and `order_bys` are documented as raw GA4 `FilterExpression`/`OrderBy` structures passed through untouched in camelCase. Mixing "we translate this one" with "send us raw API JSON for these" in one signature is a footgun for the LLM caller, and it is the direct cause of one of the inline findings below. Either translate consistently or accept raw GA4 shapes consistently (and say so unambiguously in the docstring).

Line-level findings (major/minor/nit) are posted as inline comments on the diff.

One additional minor finding that falls outside the diff and so isn't anchorable inline:

**`tests/web/api/test_public_mcp_connector_visibility.py:733` — `expected_python_apps` was not updated with `google-analytics`.** That test iterates only over its own hardcoded dict rather than the full registry, so the omission is silently harmless and the test still passes. `google-ads` has the same gap from an earlier PR, so this is a pre-existing pattern rather than a new regression — but the fix is one line and keeps the fixture honest.

## Test coverage notes

Informational, no severity attached:

- The new module lands at 94% line coverage. The uncovered lines are the non-JSON error body path, the non-dict error payload path, and the two `isinstance` guard branches called out in the inline comments below (which don't exist yet).
- No test exercises the tool-level behavior when the access token is missing — only `_headers()` is tested directly.
- **The headline feature is under-tested.** Multi-date-range comparison is only verified at the request-construction level (asserting two `dateRanges` entries appear in the body); nothing verifies that the *response* is attributable per range. Please double-check against a real GA4 property whether GA4 automatically adds the `dateRange` pseudo-dimension to report rows when multiple ranges are requested, or whether the tool must append it to `dimensions` explicitly. Without that, the LLM cannot tell which row belongs to which period — which would make the marquee feature quietly wrong rather than merely untested.
- Positive note: the migration test suite (5 cases) is more thorough than the sibling google-ads migration test (4 cases).

Lean already. No dead flexibility, no reinvented stdlib, no unnecessary dependencies, no speculative abstractions.

## Verification performed

- 35 relevant tests pass.
- `ruff format` and `ruff check` clean on the touched files.
- `mypy` could not be independently verified in this environment — please confirm via CI.
- Checked and found clean: OAuth scope handling end-to-end; injection surface (`property_id` is regex-validated, filter/order structures are passed as JSON body rather than string-interpolated); no token leakage in errors or logs; the pagination loop is bounded with no infinite-loop risk; the HTTP error handling chain is complete and internally consistent.

## Note on overlap with the existing bot review

Some of the inline findings below overlap with the `gemini-code-assist` review already posted on this same commit (`5f82413`), which flagged the `accountSummaries`/`property` null-safety gaps and weak validation on `metrics`/`dimensions`/`date_ranges`. These were reached independently here by reading the code, so treat the overlap as corroborating evidence that those are real and worth fixing — not as a duplicate flag. The two major findings (metadata truncation and unbounded/unpageable report rows) are additional to what the bot reported.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+@mcp.tool()
+def google_analytics_get_metadata(property_id: str) -> str:
+    """
+    List the dimensions and metrics available to query for a GA4 property.
+    Use this before google_analytics_run_report if you're unsure which
+    dimension/metric names are valid for this property.
+    """
+    try:
```
rogercloud, line 244: **Nit:** raw, un-normalized `property_id` (not the regex-validated numeric form) is interpolated into the log message, a small log-injection surface via embedded newlines. This pattern is widespread across other connectors in the repo (`facebook.py`, `teams.py`, `onedrive.py`, `outlook.py`), so it is not something to gate this PR on.

src/xagent/web/tools/mcp/google_analytics.py
yiboyasss: The key/shape validation was already addressed in 3c3a5ef3 (before this review round — it appears to have been taken against the initial commit): `_date_range_body` validates `start_date`/`end_date` up front, which also covers the empty-dict and camelCase-key cases, before any network call. The two increments this comment adds on top — rejecting an empty `date_ranges` list and enforcing GA4's 4-entry cap locally — were applied in 6c533f0c.

src/xagent/web/tools/mcp/google_analytics.py
yiboyasss: Already addressed in 794de855, before this review round (the review appears to be against the initial commit): the line now reads `result.get("accountSummaries") or []`, consistent with the inner `propertySummaries` access, and is covered by `test_list_properties_tolerates_null_values_in_response`.

src/xagent/web/tools/mcp/google_analytics.py
yiboyasss: Already addressed in 794de855, before this review round: the code now checks `isinstance(resource_name, str)` (mirroring google_ads.py as suggested) and additionally skips entries with a missing/empty resource name entirely, instead of emitting a record with an empty property_id. Covered by the same null-tolerance test.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,332 @@
...
+    name (apiName/uiName/category), optionally filtered by a case-insensitive
+    substring. The full GA4 metadata document (200-350+ entries with
+    descriptions) serializes to 50-140KB, which would be hard-truncated by
+    the platform output filter into broken JSON."""
+    needle = search.strip().lower()
+    projected = []
+    for entry in entries:
+        if not isinstance(entry, dict):
+            continue
+        api_name = str(entry.get("apiName") or "")
```
rogercloud: **N3 (minor):** entries missing `apiName` are kept — `str(entry.get("apiName") or "")` just produces an empty-string name and the entry is still appended below. This is the opposite of the guard added in this same PR at `google_analytics_list_properties` (skipping entries with an unusable resource name rather than emitting a garbage empty id). Suggest skipping entries without a usable `apiName` here too, and add a missing-`apiName` case to the metadata tests.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,332 @@
...
+        if needle and needle not in api_name.lower() and needle not in ui_name.lower():
+            continue
+        projected.append(
+            {
+                "apiName": api_name,
+                "uiName": ui_name,
+                "category": entry.get("category"),
+            }
+        )
+    return projected
```
rogercloud, line 232: **Residual note on finding #1 (metadata truncation, now largely fixed):** typical responses land around ~29.6KB after this projection, safely under the 51200-char MCP output truncation threshold. But a pathological case (~375 entries, all GA4-allowed custom dimensions/metrics with long names) can still reach ~57.3KB and would truncate into invalid JSON — not clamped, not tested. Worth a hard cap (e.g. truncate to first N entries + a `truncated` flag, mirroring `google_analytics_list_properties`) if this needs to be fully closed rather than just much-improved.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,332 @@
...
+        if not date_ranges:
+            raise ValueError("date_ranges must contain at least one date range")
+        if len(date_ranges) > 4:
+            raise ValueError("GA4 allows at most 4 date_ranges per report")
+        body: dict[str, Any] = {
+            "metrics": [{"name": m} for m in metrics],
+            "dateRanges": [_date_range_body(dr) for dr in date_ranges],
+            # Always sent explicitly: GA4's default would otherwise return up
+            # to 10k rows in one response.
+            "limit": str(max(1, limit)),
```
rogercloud: **Finding #2 — NOT FIXED (was major): this default reproduces the exact truncation bug it was meant to prevent.**

This only clamps the floor (`max(1, limit)`); there's no ceiling. A realistic 5-dimension + 5-metric GA4 row serializes to ~275 chars through `_success()`; ~186 rows already crosses the 51200-char MCP output truncation threshold enforced in `src/xagent/core/tools/adapters/vibe/output_filter.py:224-244`. The new default of `limit=1000` therefore produces ~277,097 chars — **~5.4x over the threshold** — which gets hard-truncated into invalid JSON, reachable with zero unusual input since it's just the tool's own default. A caller can also still pass `limit=250000` (GA4-legal) for an ~85MB response, since there's no upper bound at all.

Compare `src/xagent/web/tools/mcp/hubspot.py:145`, which double-clamps: `max(1, min(limit, 100))`.

Suggest: clamp both ends (e.g. `max(1, min(limit, PAGE_MAX))`) and lower the default so a full page of wide rows stays under the threshold — given ~275 chars/row for a 5x5 report, a default/ceiling in the low hundreds (~100-150) leaves real headroom. Also add a test that builds a multi-row response at the default `limit` and asserts the serialized `_success()` payload stays under 51200 chars — no such test exists today.

rogercloud: **N2 (minor):** `limit <= 0` is silently clamped to 1 here (no error), and negative `offset` is silently dropped a few lines below (`if offset > 0: ...`) — both inconsistent with the rest of this function, where `property_id`, empty/oversized `date_ranges`, and malformed `date_range` keys all raise `ValueError`. Suggest raising `ValueError` for `limit < 1` and `offset < 0` instead, so the model gets a correctable error rather than silently-wrong behavior. Neither path is tested today.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,332 @@
...
+      limit, call again with offset=limit, then offset=2*limit, and so on.
+    """
+    try:
+        normalized_property_id = _normalize_property_id(property_id)
+        if not date_ranges:
+            raise ValueError("date_ranges must contain at least one date range")
+        if len(date_ranges) > 4:
+            raise ValueError("GA4 allows at most 4 date_ranges per report")
+        body: dict[str, Any] = {
+            "metrics": [{"name": m} for m in metrics],
```
rogercloud, line 315: **N4 (minor):** `metrics` is required with no default, but an empty list `metrics=[]` passes straight through to `"metrics": []` here with no local rejection — relying on the remote API to error. `date_ranges` gained an explicit non-empty check in this same round of fixes; `metrics` should probably get the equivalent. Untested.

rogercloud (commented):
## What changed since the last review

Three fixup commits (`794de855`, `3c3a5ef3`, `6c533f0c`) landed on top of `5f824131`. They address the metadata-response size problem (field projection + a `search` filter), add `date_ranges` validation, harden the null-safety of `google_analytics_list_properties`, surface a `truncated` flag, cap error-detail text, add `metric_filter`, and rework `run_report` sizing/paging by introducing a `limit` default plus a new `offset` parameter. 8 new tests came with them (35 -> 43 passing).

Every fix below was re-verified by reading the code at the current HEAD rather than taken from the inline replies — including the three findings marked as fixed in the author's comments.

## Prior findings — status

| # | Finding | Status |
|---|---|---|
| 1 | Major — metadata response truncation | FIXED (residual minor) |
| 2 | Major — `run_report` unbounded rows / no pagination | **NOT FIXED — see below** |
| 3 | Minor — `date_ranges` validation | FIXED |
| 4 | Minor — `accountSummaries` None-guard | FIXED |
| 5 | Minor — property resource-name None-guard | FIXED |
| 6 | Minor — pagination cap silently truncates | FIXED |
| 7 | Minor — missing docstring on `_require_dict_result` | FIXED |
| 8 | Minor — missing `metricFilter` support | FIXED |
| 9 | Nit — uncapped error-detail text | FIXED |
| 10 | Nit — raw `property_id` in log messages | PARTIAL / resolved in spirit |
| 11 | Body nit — `expected_python_apps` missing `google-analytics` | STILL OPEN |

Details:

**1 — FIXED, with a residual note.** `_project_metadata_entries` now projects each entry down to `apiName`/`uiName`/`category`, dropping the heavy `description`, and adds an optional `search` filter. Typical responses land around 29.6KB, comfortably under the 51200-char MCP output truncation threshold. A pathological case (~375 entries, all GA4-allowed custom dimensions/metrics with long names) can still reach ~57.3KB and would truncate; that worst case is neither clamped nor tested. Good enough to close the major, worth a follow-up note (see inline comment).

**3 — FIXED, verified negatively.** All five sub-checks are present: non-dict `date_range`, missing required keys, `> 4` ranges, and empty list all raise `ValueError`. To confirm the new tests are not shell tests, the fix was reverted locally and `test_run_report_rejects_date_range_missing_required_keys` genuinely fails without it.

**4 / 5 — FIXED.** `result.get("accountSummaries") or []`, and `isinstance(resource_name, str) and resource_name` before `.rsplit()` (skipping unusable entries, matching the `google_ads.py` pattern that was suggested). Covered by `test_list_properties_tolerates_null_values_in_response`, which includes an explicit `"accountSummaries": None` case.

**6 — FIXED.** `google_analytics_list_properties` now returns `truncated=bool(page_token)`, with a test asserting `truncated=True` at the cap.

**7 / 8 / 9 — FIXED.** Docstring on `_require_dict_result` now matches the `google_ads.py` sibling; `metric_filter` is accepted and wired; error detail is `response.text.strip()[:1000]` with a rationale comment, matching the `meta_graph.py` precedent.

**10 — PARTIAL, treating as closed.** Both log sites now use `{property_id!r}`, which closes the log-injection angle (repr escapes embedded newlines). The value logged is still the raw parameter rather than `normalized_property_id`, so the original wording technically still applies, but the practical risk is gone. Not re-flagging.

**11 — STILL OPEN.** None of the three fixup commits touched `tests/web/api/test_public_mcp_connector_visibility.py:733`; `google-analytics` is still absent from `expected_python_apps`. Same pre-existing gap as `google-ads`, so still non-blocking — but it is a one-line addition that has been available since the first review.

## Blocking: finding #2 is not actually closed

The fixup commit titled "address PR review on the GA connector output sizes and paging" is easy to read as having resolved this. It did not. It changed shape: previously `limit` was omitted and GA4 applied its 10,000-row default; now `limit: int = 1000` is hardcoded. `offset: int = 0` was added and is correctly wired into the request body, and the docstring guides the model on how to page — that part is a genuine improvement. The sizing problem is worse in one important sense, though: **the failure is now reachable with the tool's own default arguments, no unusual input required.**

Measured against the real serialization path (`_success()` JSON output, fed through `src/xagent/core/tools/adapters/vibe/output_filter.py:224-244`, which treats the tool's entire JSON return value as a single string leaf and hard-truncates at 51200 chars):

- A realistic 5-dimension + 5-metric GA4 row serializes to ~275 chars.
- ~186 rows already crosses the truncation threshold.
- `limit=1000` — the new default — produces ~277,097 chars, roughly **5.4x over the threshold**. The result is truncated mid-JSON, i.e. exactly the invalid-JSON failure mode this change was meant to eliminate.

There is also still no upper clamp. `max(1, limit)` is a floor only, so a caller can legitimately pass `limit=250000` (GA4 permits it) for an ~85MB response. Compare the sibling connector `src/xagent/web/tools/mcp/hubspot.py:145`, which double-clamps: `max(1, min(limit, 100))`.

See inline comment for the suggested fix.

## New findings from this round

Detailed as inline comments below: `limit`/`offset` silently correcting invalid input instead of rejecting it (N2), `_project_metadata_entries` keeping entries with no `apiName` instead of skipping them (N3), and `metrics` having no empty-list validation (N4).

## Design-level re-assessment

**Verdict: acceptable with reservations** (previously "sound"). Downgraded because the fixup work introduced two internal inconsistencies in validation and failure philosophy, on top of the unresolved `run_report` sizing issue above.

**New — inconsistent failure philosophy within one file.** After the fixups, `google_analytics.py` has a split personality: `property_id`, `date_ranges`, and `date_range` keys all fail loud with `ValueError` (the right call for an LLM-driven tool — the model can retry with corrected arguments), while `limit` and `offset` fail quiet by clamping or dropping. Given the direction the rest of the file now leans, the silent clamps are the outlier. This is not merely stylistic: an explicit ceiling with a clear error is precisely the mechanism that would have surfaced the oversized default before it shipped. Ties directly to N2 and to the still-open finding #2.

**New — defensive-guard pileup with untested branches.** The null-safety hardening now stacks `_require_dict_result` + several `isinstance(..., dict)` checks + `or []` fallbacks + `isinstance(..., str)` checks across the three tools. Defensible given how loosely typed the GA4 responses are, but several of these branches have zero coverage: `_require_dict_result`'s raise path on any of the three tools, and the metadata endpoint's `or []` and non-dict-entry guards. The guards exist but the suite does not verify they behave as intended.

**D1 (undocumented operational prerequisites) — downgraded to a nit.** Still accurate, but this turns out to be a repo-wide documentation gap: no document anywhere in the repo describes GCP Console API enablement or OAuth consent-screen scope registration for any connector, including the sibling `google-ads`. Better handled as a separate repo-level documentation issue than as a change request on this PR.

**D2 (helper duplication with `google_ads.py`) — narrower than originally stated, with one correction.** Only `_success`/`_error` are byte-identical duplicates. `_headers`, `_extract_error_detail`, and `_request` genuinely differ — Ads carries developer-token and login-customer-id handling, GA carries the 1000-char detail truncation. Correction to the earlier note: the pagination loop is not duplicated — `google_ads.py` has no pagination code at all. Per-connector `_success`/`_error`/`_request` is also the prevailing pattern in `hubspot.py`, `onedrive.py`, `outlook.py`, and `teams.py`; only `facebook.py`/`instagram.py` share via `meta_graph.py`. Keeping as a nit: the ~10-15 lines of `_success`/`_error` could move to `utils.py` someday, not urgent.

**D3 (snake_case params with camelCase provider passthrough) — nit, not a blocker.** The mixed convention stands as described, but "normalized snake_case params plus raw provider-shaped passthrough for filter/dimension structures" already exists in `outlook.py` and `hubspot.py`, and the docstrings here clearly label which parameters are raw GA4 passthrough.

## Test coverage gaps (informational)

Well covered by this round: empty and `>4` `date_ranges` (parametrized), `date_range` missing keys and camelCase forms (parametrized), null `accountSummaries`/property entries (dedicated regression test), the `truncated` flag, error-detail truncation, and metadata projection plus `search`.

Remaining gaps, all tied to the findings above:

- No test for `run_report` with a realistic multi-row response asserting total serialized output size (blocking finding #2).
- No test for `limit <= 0` or negative `offset` (N2).
- No test for a metadata entry missing `apiName` (N3).
- No test for `metrics=[]` (N4).
- No test exercising `_require_dict_result`'s raise branch on any of the three tools.

## Verification performed

- 43 tests pass in the current checkout at `6c533f0c`, up from 35 at the time of the first review.
- Finding #3's fix was verified negatively by reverting it locally and confirming the new test fails, then restoring.
- Output-size figures for findings #1 and #2 were computed by serializing constructed responses through the tool's actual `_success()` path and comparing against the 51200-char threshold enforced in `src/xagent/core/tools/adapters/vibe/output_filter.py:224-244`.
- `ruff`/`mypy` were not re-run this round — please confirm they are green in CI.

---

All statuses above come from reading the code at the current HEAD, including the three findings the inline replies reported as fixed; none of it relies on those replies. Two of those three check out; the third (`run_report` output sizing) does not, and is the one item worth resolving before merge.

src/xagent/web/tools/mcp/google_analytics.py
```diff
@@ -0,0 +1,348 @@
...
+
+mcp = FastMCP("google-analytics-mcp")
+
+DATA_API_BASE_URL = "https://analyticsdata.googleapis.com/v1beta"
+ADMIN_API_BASE_URL = "https://analyticsadmin.googleapis.com/v1beta"
+DEFAULT_TIMEOUT_SECONDS = 30
+MAX_ACCOUNT_SUMMARY_PAGES = 20
+# A 5-dimension + 5-metric row serializes to ~275 chars through _success(); at
+# RUN_REPORT_MAX_LIMIT rows that's comfortably under the platform's 51200-char
+# MCP output truncation threshold even for wide reports.
```
rogercloud: **N5 (minor):** this justification overclaims — "even for wide reports" doesn't hold. The limit bounds row *count*, not bytes: with 5 dimensions at `limit=150`, once average dimension-value length reaches roughly 26 characters, the serialized response crosses the 51200-char threshold and truncates into invalid JSON again. Nothing caps the number of dimensions/metrics or the length of dimension values, and the docstring below uses `landingPagePlusQueryString` (full URL-shaped values) as an example dimension. A realistic case (1 long page-path + 2 short dimensions + 5 metrics) at 150 rows measures ~47,978 chars — still fine — so the everyday case is safe; the exposure is specifically several long-valued dimensions at once.

Suggest either tightening this comment to state its actual bounds, or (better) adding a final `len(serialized) > threshold` check before returning that trims rows and sets a `truncated: true` flag — `google_analytics_list_properties` already implements exactly this pattern (line ~195), so the precedent exists in this same file.

tests/web/tools/test_google_analytics_mcp.py
```diff
@@ -0,0 +1,704 @@
...
+            date_ranges=[{"start_date": "7daysAgo", "end_date": "today"}],
+            offset=-1,
+        )
+    )
+
+    assert result["status"] == "error"
+    assert "offset" in result["message"]
+    mock_request.assert_not_called()
+
+
```
rogercloud, line 516: **N6 (minor):** this regression test only runs at `RUN_REPORT_DEFAULT_LIMIT` (100), never at `RUN_REPORT_MAX_LIMIT` (150) — the value that actually determines worst-case output size. It also uses 11-character dimension values (`f"dim-value-{i}"`), the narrowest realistic case, leaving a ~24KB margin. As written, it wouldn't catch a regression like raising `RUN_REPORT_MAX_LIMIT` to 1000 while leaving the default at 100.

Suggest parametrizing over both `RUN_REPORT_DEFAULT_LIMIT` and `RUN_REPORT_MAX_LIMIT`, and widening the dimension-value length to ~30-40 chars to actually stress the boundary described in the comment at `google_analytics.py:24-26`.

tests/web/tools/test_google_analytics_mcp.py
```diff
@@ -0,0 +1,704 @@
...
+    response = google_analytics.google_analytics_run_report(
+        "42",
+        metrics=[f"metric{i}" for i in range(5)],
+        date_ranges=[{"start_date": "7daysAgo", "end_date": "today"}],
+        dimensions=[f"dim{i}" for i in range(5)],
+    )
+
+    result = json.loads(response)
+    assert result["status"] == "success"
+    assert len(response) < 51200
```
rogercloud: **N7 (nit):** `51200` is hardcoded here instead of referencing its source of truth, `get_tool_max_output_length()` in `src/xagent/config.py`. If that default ever changes, this test silently stops guarding the real threshold.

rogercloud (approved):
## Round 3 follow-up review

Commit `e8f48942` clamps the `run_report` limit, converts the remaining silent-correction paths into explicit `ValueError`s, and drops metadata entries with unusable `apiName`. All four open items from round 2 are addressed, and I re-verified each one by reading the current code and running the tests rather than relying on the commit message.

### Previously-blocking finding #2 (run_report output size) — RESOLVED

This was the blocking MAJOR from round 2, and it is genuinely fixed.

- `src/xagent/web/tools/mcp/google_analytics.py:27-28` introduces `RUN_REPORT_DEFAULT_LIMIT = 100` and `RUN_REPORT_MAX_LIMIT = 150`, and the `limit` parameter default moves from `1000` to `RUN_REPORT_DEFAULT_LIMIT`.
- `src/xagent/web/tools/mcp/google_analytics.py:305-308` now raises on out-of-range input: `limit < 1 or limit > RUN_REPORT_MAX_LIMIT` and `offset < 0`, replacing the old floor-only `str(max(1, limit))` clamp.

Measured on real serialized output built through the tool's actual `_success()` structure (not estimated), at the reference scenario documented in the code (5 dimensions + 5 metrics, ~11-char dimension values):

| | serialized size | vs. 51200-char threshold |
|---|---|---|
| before (`limit=1000` default) | ~277,000 chars | ~5.4x over — truncated into invalid JSON |
| after (`limit=100` default) | ~29,417 chars | comfortably under |
| after (`limit=150`, the max) | ~43,967 chars | under, ~14% margin |

The regression tests are non-trivial — each was confirmed to fail when the fix is reverted: `test_run_report_rejects_out_of_range_limit` (parametrized over `[0, -1, RUN_REPORT_MAX_LIMIT + 1]`), `test_run_report_rejects_negative_offset`, and `test_run_report_default_limit_keeps_serialized_response_under_truncation_threshold` (builds a real 100-row response and asserts the serialized length stays under 51200).

### Status of everything from round 2

| Item | Status |
|---|---|
| #2 — run_report unbounded/unpageable output (blocking MAJOR) | RESOLVED (residual minor N5 below) |
| N2 — limit/offset silently corrected instead of raising | FIXED — `google_analytics.py:305-308`, covered by tests asserting `mock_request.assert_not_called()` |
| N3 — metadata entries missing `apiName` kept instead of skipped | FIXED — `google_analytics.py:212-216` skips entries whose `apiName` is not a non-empty `str`; `test_get_metadata_skips_entries_without_usable_api_name` covers `None`, missing key, and valid |
| N4 — `metrics=[]` not validated | FIXED — `google_analytics.py:299-300` raises `ValueError`; covered by `test_run_report_rejects_empty_metrics` |
| #11 — `expected_python_apps` missing `google-analytics` | FIXED and verified correct — `tests/web/api/test_public_mcp_connector_visibility.py:718-721`; module name and `env_mapping` match the registry entry at `src/xagent/web/builtin_mcp_registry.py:241-256` exactly |
| Design note — inconsistent failure philosophy | RESOLVED — all input validation now raises; silent skipping is confined to response parsing, which is the right split |

Caveat on #11: the consumer test (`test_builtin_registry_uses_runtime_available_launch_commands`) iterates the dict's own keys rather than asserting `set(expected) == set(actual_registered_apps)`, so the dict remains a silently-droppable pattern. That is pre-existing and out of scope here — noting it, not pressing it on this PR.

### New findings this round — all minor/nit, none blocking

Posted as inline comments below (N5, N6, N7). Summary: the comment justifying `RUN_REPORT_MAX_LIMIT = 150` overclaims ("even for wide reports") and there's no final byte-size guard for the case where several dimension values are simultaneously long; the new sizing regression test only exercises the default limit, not the max; and the 51200 threshold is hardcoded in the test rather than sourced from config.

### Simplification opportunity (scoped to `e8f48942`)

`tests/web/tools/test_google_analytics_mcp.py:478: shrink: parametrize bad_limit=[0, -1, MAX+1] tests two values (0 and -1) that both hit the identical limit < 1 branch. Drop one, keep bad_limit=[0, RUN_REPORT_MAX_LIMIT + 1].`

Net: one redundant parametrization value removed; no coverage lost.

### Tests run

- `tests/web/tools/test_google_analytics_mcp.py` — 34 passed
- `tests/web/api/test_public_mcp_connector_visibility.py` — 40 passed

Both were actually executed against `e8f48942`, not inferred.

### Verdict

Design: acceptable-with-reservations — unchanged from round 2, but for a much smaller reservation. This commit finished converting silent correction into explicit validation, bringing `run_report` in line with the rest of the file's error-handling philosophy, and no new bugs were introduced. The one remaining reservation is architectural rather than a defect: the connector still uses row count as a proxy for a byte budget with no final safety net before returning, when the same file already has a working `truncated`-flag precedent (in `google_analytics_list_properties`) that was not reused here.

This is the third independently verified round, and no blocking issues remain. Approving — N5-N7 are polish and can land now or as a fast-follow.

PR #15, opened 2026-02-25
feat(files): migrate to file_id-only flow and harden preview/upload security

bluefish-08:
**Summary**
- Migrate file handling to file_id-only across backend, websocket output, DAG result formatting, and frontend preview flows.
- Add UploadedFile metadata model + migration, and switch file APIs to file_id semantics.
- Fix security and reliability issues discovered during review.

**Key Changes**
- Backend/API
  - Added `uploaded_files` metadata table and file_id-based endpoints.
  - Updated upload/list/download/preview/delete flows to use `file_id`.
  - Hardened public preview path resolution to block directory traversal via `relative_path`.
- WebSocket / DAG
  - Normalize `file_outputs` from both string/dict formats into file_id-based objects.
  - Rewrite markdown `file:` links to final file_id references.
  - Use randomized output file_id (removed predictable path-based generation).
  - Deduplicate websocket upload target paths for same-name files.
  - Build preview uploaded metadata now stores `task_id=None` (no invalid FK type).
  - Removed remaining path-wording residue in result analyzer prompt text.
- Frontend
  - File preview/title behavior aligned to file_id flow.
  - Markdown file-link preview title now prefers link text / alt / title (instead of UUID fallback when possible).

**Validation**
- `pytest tests/web/test_admin_file_access.py tests/web/test_file_upload.py tests/integration/test_websocket.py tests/core/agent/pattern/dag_plan_execute/test_result_analyzer_language.py -q`
- `npm run type-check` (frontend)

PR #1069, opened 2026-07-30
feat: add Slack MCP connector

yiboyasss:
## Summary
- Register `slack` as a built-in OAuth MCP connector (`builtin_mcp_registry.py` + seed migration) — new provider (own client_id/secret), since Slack isn't a workspace shared with any existing OAuth provider
- New `xagent.web.tools.mcp.slack` module, two tools:
  - `slack_list_channels` — paginated `conversations.list`, resolves a channel name to an id
  - `slack_post_message` — `chat.postMessage`, accepts either a channel id or `#name`
- Scope: `chat:write` + `chat:write.public` (post to any public channel without a manual bot invite) + `channels:read` — matches the PDF DevOps Agent spec, which only requires posting incident findings (summary, evidence, root-cause hypothesis, recommended fix, severity), not reading or reacting to messages
- No changes needed in `auth.py`/`config.py`: confirmed via Slack's official docs that `oauth.v2.access` accepts `client_id`/`client_secret` in the POST body (Basic Auth is only recommended, not enforced, unlike Zoom), and bot tokens don't expire unless token rotation is explicitly enabled on the Slack app (left off) — the existing generic exchange/refresh code handles both cases already
- Slack's Web API always returns HTTP 200 even on failure, signalling errors via `{"ok": false, "error": "..."}` in the body — the connector's `_request` helper checks `ok`/`error` explicitly rather than relying on `raise_for_status()`

## Test plan
- [x] `pytest tests/web/tools/test_slack_mcp.py tests/alembic/test_20260801_seed_slack_mcp_app.py tests/templates/test_manager.py` — all passing
- [x] `alembic upgrade head` / `downgrade -1` round-trip verified against a scratch DB — only the `slack` rows are added/removed, single head
- [x] Manual end-to-end via a real Slack workspace (OAuth connect through an ngrok tunnel, since the Slack app requires a public HTTPS redirect): connected successfully, `slack_list_channels` and `slack_post_message` both verified working, including posting a structured incident message (summary/evidence/root-cause/fix/severity) matching the DevOps Agent's target output format

### DIFF
--- a/example.env
+++ b/example.env
@@ -460,6 +460,8 @@ META_CONFIG_ID=""
 # Required for the Slack integration in MCP (list channels, post messages)
 # Create an app at https://api.slack.com/apps, add these scopes under
 # "OAuth & Permissions": chat:write, chat:write.public, channels:read
+# Note: chat:write.public also lets the bot post into any public channel
+# without being invited — it bypasses Slack's invite gate.
 # Leave "Token Rotation" off — bot tokens don't expire, so no refresh is needed.
 SLACK_CLIENT_ID=""
 SLACK_CLIENT_SECRET=""
--- a/example.env
+++ b/example.env
@@ -454,6 +454,20 @@ META_REDIRECT_URI=""
 # users who authorize via config_id will not grant it.
 META_CONFIG_ID=""
 
+# ===========================================
+# Slack OAuth Configuration (Optional)
+# ===========================================
+# Required for the Slack integration in MCP (list channels, post messages)
+# Create an app at https://api.slack.com/apps, add these scopes under
+# "OAuth & Permissions": chat:write, chat:write.public, channels:read
+# Leave "Token Rotation" off — bot tokens don't expire, so no refresh is needed.
+SLACK_CLIENT_ID=""
+SLACK_CLIENT_SECRET=""
+# Redirect URI for Slack OAuth callback — must also be added under the app's
+# "OAuth & Permissions" > "Redirect URLs".
+# Example: http://localhost:8000/api/auth/slack/callback
+SLACK_REDIRECT_URI=""
+
 # Note:
 # - If embedding API keys are configured, LanceDB will be used automatically
 # - LanceDB storage path: <project_root>/memory_store/
--- a/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
+++ b/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
@@ -1,0 +1,174 @@
+"""seed built-in Slack (OAuth) MCP connector
+
+Revision ID: 20260801_seed_slack_mcp_app
+Revises: 20260724_add_upload_source_to_uploaded_files
+Create Date: 2026-08-01 00:00:00.000000
+
+"""
+
+import os
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "20260801_seed_slack_mcp_app"
+down_revision: Union[str, None] = "20260724_add_upload_source_to_uploaded_files"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+FULL_OAUTH_PROVIDERS_TABLE = sa.table(
+    "oauth_providers",
+    sa.column("provider_name", sa.String),
+    sa.column("name", sa.String),
+    sa.column("client_id", sa.String),
+    sa.column("client_secret", sa.String),
+    sa.column("auth_url", sa.String),
+    sa.column("token_url", sa.String),
+    sa.column("redirect_uri", sa.String),
+    sa.column("userinfo_url", sa.String),
+    sa.column("user_id_path", sa.String),
+    sa.column("email_path", sa.String),
+    sa.column("default_scopes", sa.JSON),
+)
+
+OAUTH_PROVIDERS_TABLE = sa.table(
+    "oauth_providers",
+    sa.column("provider_name", sa.String),
+)
+
+PUBLIC_MCP_APPS_TABLE = sa.table(
+    "public_mcp_apps",
+    sa.column("app_id", sa.String),
+    sa.column("name", sa.String),
+    sa.column("description", sa.Text),
+    sa.column("icon", sa.String),
+    sa.column("transport", sa.String),
+    sa.column("provider_name", sa.String),
+    sa.column("category", sa.String),
+    sa.column("oauth_scopes", sa.JSON),
+    sa.column("is_visible_in_connector", sa.Boolean),
+    sa.column("launch_config", sa.JSON),
+)
+
+APP_ID = "slack"
+
+SLACK_SCOPES = ["chat:write", "chat:write.public", "channels:read"]
+
+
+def _filter_row(row: dict[str, object], allowed_columns: set[str]) -> dict[str, object]:
+    return {key: value for key, value in row.items() if key in allowed_columns}
+
+
+def _slack_provider_row() -> dict[str, object]:
+    return {
+        "provider_name": "slack",
+        "name": "Slack",
+        "client_id": os.environ.get("SLACK_CLIENT_ID", ""),
+        "client_secret": os.environ.get("SLACK_CLIENT_SECRET", ""),
+        "auth_url": "https://slack.com/oauth/v2/authorize",
+        "token_url": "https://slack.com/api/oauth.v2.access",
+        "redirect_uri": os.environ.get("SLACK_REDIRECT_URI", ""),
+        "userinfo_url": "https://slack.com/api/auth.test",
+        "user_id_path": "team_id",
+        "email_path": "team",
+        "default_scopes": SLACK_SCOPES,
+    }
+
+
+def _slack_app_row() -> dict[str, object]:
+    return {
+        "app_id": APP_ID,
+        "name": "Slack",
+        "description": "Connect to Slack to list channels and post messages, e.g. incident summaries and recommended fixes.",
+        "icon": "https://www.google.com/s2/favicons?domain=slack.com&sz=128",
+        "transport": "oauth",
+        "provider_name": "slack",
+        "category": "Communication",
+        "oauth_scopes": SLACK_SCOPES,
+        "is_visible_in_connector": True,
+        "launch_config": {
+            "command": "python",
+            "args": ["-m", "xagent.web.tools.mcp.slack"],
+            "env_mapping": {"SLACK_ACCESS_TOKEN": "access_token"},
+        },
+    }
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    existing_tables = set(inspector.get_table_names())
+
+    if "oauth_providers" in existing_tables:
+        oauth_columns = {
+            column["name"] for column in inspector.get_columns("oauth_providers")
+        }
+        existing_provider_names = set(
+            bind.execute(sa.select(OAUTH_PROVIDERS_TABLE.c.provider_name)).scalars()
+        )
+        if "slack" not in existing_provider_names:
+            bind.execute(
+                sa.insert(FULL_OAUTH_PROVIDERS_TABLE),
+                [_filter_row(_slack_provider_row(), oauth_columns)],
+            )
+
+    if "public_mcp_apps" in existing_tables:
+        app_columns = {
+            column["name"] for column in inspector.get_columns("public_mcp_apps")
+        }
+        existing_app_ids = set(
+            bind.execute(sa.select(PUBLIC_MCP_APPS_TABLE.c.app_id)).scalars()
+        )
+        if APP_ID not in existing_app_ids:
+            bind.execute(
+                sa.insert(PUBLIC_MCP_APPS_TABLE),
+                [_filter_row(_slack_app_row(), app_columns)],
+            )
+
+
+def downgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    existing_tables = set(inspector.get_table_names())
+
+    if "public_mcp_apps" in existing_tables:
+        # Only the catalog entry is removed. Any user OAuth connections created
+        # against this app are not owned by this migration and are cleaned up
+        # through the normal disconnect path.
+        bind.execute(
+            sa.delete(PUBLIC_MCP_APPS_TABLE).where(
+                PUBLIC_MCP_APPS_TABLE.c.app_id == APP_ID
+            )
+        )
+
+    if "oauth_providers" not in existing_tables:
+        return
+
+    if "public_mcp_apps" in existing_tables:
+        remaining_slack_apps = bind.execute(
+            sa.select(sa.func.count())
+            .select_from(PUBLIC_MCP_APPS_TABLE)
+            .where(PUBLIC_MCP_APPS_TABLE.c.provider_name == "slack")
+        ).scalar()
+        if remaining_slack_apps:
+            return
+
+    # Only delete the provider row when it still matches the static shape this
+    # migration seeded, so an admin-created "slack" provider (via
+    # POST /admin/mcp/providers) is preserved. client_id/client_secret are
+    # env-dependent and intentionally not part of the guard.
+    provider_columns = {
+        column["name"] for column in inspector.get_columns("oauth_providers")
+    }
+    seeded_provider = _slack_provider_row()
+    delete_stmt = sa.delete(FULL_OAUTH_PROVIDERS_TABLE).where(
+        FULL_OAUTH_PROVIDERS_TABLE.c.provider_name == "slack"
+    )
+    for column in ("name", "auth_url", "token_url"):
+        if column in provider_columns:
+            delete_stmt = delete_stmt.where(
+                FULL_OAUTH_PROVIDERS_TABLE.c[column] == seeded_provider[column]
+            )
+    bind.execute(delete_stmt)
--- a/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
+++ b/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
@@ -1,7 +1,7 @@
 """seed built-in Slack (OAuth) MCP connector
 
 Revision ID: 20260801_seed_slack_mcp_app
-Revises: 20260724_add_upload_source_to_uploaded_files
+Revises: 20260728_add_facebook_pages_read_user_content_scope
 Create Date: 2026-08-01 00:00:00.000000
 
 """
@@ -14,7 +14,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "20260801_seed_slack_mcp_app"
-down_revision: Union[str, None] = "20260724_add_upload_source_to_uploaded_files"
+down_revision: Union[str, None] = "20260728_add_facebook_pages_read_user_content_scope"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
+++ b/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
@@ -1,7 +1,7 @@
 """seed built-in Slack (OAuth) MCP connector
 
 Revision ID: 20260801_seed_slack_mcp_app
-Revises: 20260728_add_facebook_pages_read_user_content_scope
+Revises: 20260730_add_slack_oauth_flow_states
 Create Date: 2026-08-01 00:00:00.000000
 
 """
@@ -14,7 +14,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "20260801_seed_slack_mcp_app"
-down_revision: Union[str, None] = "20260728_add_facebook_pages_read_user_content_scope"
+down_revision: Union[str, None] = "20260730_add_slack_oauth_flow_states"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
+++ b/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
@@ -71,6 +71,11 @@ def _slack_provider_row() -> dict[str, object]:
         "token_url": "https://slack.com/api/oauth.v2.access",
         "redirect_uri": os.environ.get("SLACK_REDIRECT_URI", ""),
         "userinfo_url": "https://slack.com/api/auth.test",
+        # auth.test never returns an email for a bot token; the workspace
+        # name ("team") is deliberately stored in the email slot because
+        # UserOAuth.email is only consumed as the "connected account" display
+        # label for non-gmail providers — without it the Slack connection
+        # would show up unlabeled in the connector UI.
         "user_id_path": "team_id",
         "email_path": "team",
         "default_scopes": SLACK_SCOPES,
--- a/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
+++ b/src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
@@ -33,11 +33,6 @@ FULL_OAUTH_PROVIDERS_TABLE = sa.table(
     sa.column("default_scopes", sa.JSON),
 )
 
-OAUTH_PROVIDERS_TABLE = sa.table(
-    "oauth_providers",
-    sa.column("provider_name", sa.String),
-)
-
 PUBLIC_MCP_APPS_TABLE = sa.table(
     "public_mcp_apps",
     sa.column("app_id", sa.String),
@@ -111,7 +106,9 @@ def upgrade() -> None:
             column["name"] for column in inspector.get_columns("oauth_providers")
         }
         existing_provider_names = set(
-            bind.execute(sa.select(OAUTH_PROVIDERS_TABLE.c.provider_name)).scalars()
+            bind.execute(
+                sa.select(FULL_OAUTH_PROVIDERS_TABLE.c.provider_name)
+            ).scalars()
         )
         if "slack" not in existing_provider_names:
             bind.execute(
@@ -160,20 +157,13 @@ def downgrade() -> None:
         if remaining_slack_apps:
             return
 
-    # Only delete the provider row when it still matches the static shape this
-    # migration seeded, so an admin-created "slack" provider (via
-    # POST /admin/mcp/providers) is preserved. client_id/client_secret are
-    # env-dependent and intentionally not part of the guard.
-    provider_columns = {
-        column["name"] for column in inspector.get_columns("oauth_providers")
-    }
-    seeded_provider = _slack_provider_row()
-    delete_stmt = sa.delete(FULL_OAUTH_PROVIDERS_TABLE).where(
-        FULL_OAUTH_PROVIDERS_TABLE.c.provider_name == "slack"
+    # Delete unconditionally by provider_name, matching the sibling seed
+    # migrations (meta, google-maps). A shape-matching guard would protect
+    # the wrong thing: an admin recreating the *real* Slack provider enters
+    # Slack's canonical URLs, which would match the guard and be deleted
+    # anyway — only a differently-shaped row would survive.
+    bind.execute(
+        sa.delete(FULL_OAUTH_PROVIDERS_TABLE).where(
+            FULL_OAUTH_PROVIDERS_TABLE.c.provider_name == "slack"
+        )
     )
-    for column in ("name", "auth_url", "token_url"):
-        if column in provider_columns:
-            delete_stmt = delete_stmt.where(
-                FULL_OAUTH_PROVIDERS_TABLE.c[column] == seeded_provider[column]
-            )
-    bind.execute(delete_stmt)
--- a/src/xagent/web/builtin_mcp_registry.py
+++ b/src/xagent/web/builtin_mcp_registry.py
@@ -121,6 +121,11 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
             "token_url": "https://slack.com/api/oauth.v2.access",
             "redirect_uri": os.environ.get("SLACK_REDIRECT_URI", ""),
             "userinfo_url": "https://slack.com/api/auth.test",
+            # auth.test never returns an email for a bot token; the workspace
+            # name ("team") is deliberately stored in the email slot because
+            # UserOAuth.email is only consumed as the "connected account"
+            # display label for non-gmail providers — without it the Slack
+            # connection would show up unlabeled in the connector UI.
             "user_id_path": "team_id",
             "email_path": "team",
             "default_scopes": ["chat:write", "chat:write.public", "channels:read"],
--- a/src/xagent/web/builtin_mcp_registry.py
+++ b/src/xagent/web/builtin_mcp_registry.py
@@ -112,6 +112,19 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
             "email_path": "email",
             "default_scopes": ["public_profile"],
         },
+        {
+            "provider_name": "slack",
+            "name": "Slack",
+            "client_id": os.environ.get("SLACK_CLIENT_ID", ""),
+            "client_secret": os.environ.get("SLACK_CLIENT_SECRET", ""),
+            "auth_url": "https://slack.com/oauth/v2/authorize",
+            "token_url": "https://slack.com/api/oauth.v2.access",
+            "redirect_uri": os.environ.get("SLACK_REDIRECT_URI", ""),
+            "userinfo_url": "https://slack.com/api/auth.test",
+            "user_id_path": "team_id",
+            "email_path": "team",
+            "default_scopes": ["chat:write", "chat:write.public", "channels:read"],
+        },
     ]
 
 
@@ -378,6 +391,22 @@ def get_builtin_public_mcp_app_rows() -> list[dict[str, Any]]:
                 "command": "npx",
                 "args": ["-y", "@cablate/mcp-google-map", "--stdio"],
                 "required_env": ["GOOGLE_MAPS_API_KEY"],
+            },
+        },
+        {
+            "app_id": "slack",
+            "name": "Slack",
+            "description": "Connect to Slack to list channels and post messages, e.g. incident summaries and recommended fixes.",
+            "icon": "https://www.google.com/s2/favicons?domain=slack.com&sz=128",
+            "transport": "oauth",
+            "provider_name": "slack",
+            "category": "Communication",
+            "oauth_scopes": ["chat:write", "chat:write.public", "channels:read"],
+            "is_visible_in_connector": True,
+            "launch_config": {
+                "command": "python",
+                "args": ["-m", "xagent.web.tools.mcp.slack"],
+                "env_mapping": {"SLACK_ACCESS_TOKEN": "access_token"},
             },
         },
     ]
--- a/src/xagent/web/tools/mcp/slack.py
+++ b/src/xagent/web/tools/mcp/slack.py
@@ -41,18 +41,24 @@ def _request(
     path: str,
     *,
     params: dict[str, Any] | None = None,
+    json_data: dict[str, Any] | None = None,
 ) -> dict[str, Any]:
     """Call a Slack Web API method.
 
     Slack always answers with HTTP 200, even on failure — success/failure is
     only signalled by the body's "ok" boolean plus an "error" code, so error
     detection here is on that field, not on raise_for_status().
+
+    Write calls pass json_data rather than params: message text can exceed
+    URL length limits, and query strings are commonly logged in plaintext by
+    proxies/load balancers, which would leak message content.
     """
     response = requests.request(
         method=method,
         url=f"{SLACK_BASE_URL}/{path}",
         headers=_headers(),
         params=params,
+        json=json_data,
         timeout=DEFAULT_TIMEOUT_SECONDS,
     )
     response.raise_for_status()
@@ -75,7 +81,10 @@ def slack_list_channels(exclude_archived: bool = True) -> str:
         for _ in range(MAX_PAGES):
             params: dict[str, Any] = {
                 "types": "public_channel",
-                "exclude_archived": exclude_archived,
+                # Slack expects a lowercase "true"/"false" string; requests
+                # would otherwise serialize the Python bool as "True"/"False",
+                # which Slack's parser doesn't recognize as a boolean.
+                "exclude_archived": "true" if exclude_archived else "false",
                 "limit": 200,
             }
             if cursor:
@@ -109,7 +118,7 @@ def slack_post_message(channel: str, text: str) -> str:
         result = _request(
             "POST",
             "chat.postMessage",
-            params={"channel": channel, "text": text},
+            json_data={"channel": channel, "text": text},
         )
         return _success(channel=result.get("channel"), ts=result.get("ts"))
     except Exception as e:
--- a/src/xagent/web/tools/mcp/slack.py
+++ b/src/xagent/web/tools/mcp/slack.py
@@ -1,0 +1,121 @@
+import json
+import logging
+import os
+from typing import Any
+
+import requests
+from mcp.server.fastmcp import FastMCP
+
+from .utils import setup_proxy_env
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger("slack-mcp")
+
+# Ensure standard proxy environment variables are set to prevent hanging requests
+setup_proxy_env()
+
+mcp = FastMCP("slack-mcp")
+
+SLACK_BASE_URL = "https://slack.com/api"
+DEFAULT_TIMEOUT_SECONDS = 30
+MAX_PAGES = 20
+
+
+def _success(**payload: Any) -> str:
+    return json.dumps({"status": "success", **payload}, ensure_ascii=False)
+
+
+def _error(message: str) -> str:
+    return json.dumps({"status": "error", "message": message}, ensure_ascii=False)
+
+
+def _headers() -> dict[str, str]:
+    access_token = os.environ.get("SLACK_ACCESS_TOKEN")
+    if not access_token:
+        raise ValueError("SLACK_ACCESS_TOKEN environment variable is missing")
+    return {"Authorization": f"Bearer {access_token}"}
+
+
+def _request(
+    method: str,
+    path: str,
+    *,
+    params: dict[str, Any] | None = None,
+) -> dict[str, Any]:
+    """Call a Slack Web API method.
+
+    Slack always answers with HTTP 200, even on failure — success/failure is
+    only signalled by the body's "ok" boolean plus an "error" code, so error
+    detection here is on that field, not on raise_for_status().
+    """
+    response = requests.request(
+        method=method,
+        url=f"{SLACK_BASE_URL}/{path}",
+        headers=_headers(),
+        params=params,
+        timeout=DEFAULT_TIMEOUT_SECONDS,
+    )
+    response.raise_for_status()
+    payload: dict[str, Any] = response.json()
+    if not payload.get("ok"):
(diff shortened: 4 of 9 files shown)


src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,130 @@
...
+                "types": "public_channel",
+                # Slack expects a lowercase "true"/"false" string; requests
+                # would otherwise serialize the Python bool as "True"/"False",
+                # which Slack's parser doesn't recognize as a boolean.
+                "exclude_archived": "true" if exclude_archived else "false",
+                "limit": 200,
+            }
+            if cursor:
+                params["cursor"] = cursor
+            result = _request("GET", "conversations.list", params=params)
```
rogercloud: **MAJOR — a mid-pagination failure discards every page fetched so far.**

There's no per-page `try`/`except` around this call, so if page N fails the exception unwinds to the outer `except Exception` at line 105 and all channels accumulated from pages 1..N-1 are thrown away in favour of a bare error string.

This isn't hypothetical: `conversations.list` is Tier-2 rate-limited at roughly 20 requests/minute, and this loop issues up to `MAX_PAGES` (20) requests back to back with no delay, no backoff, and no `Retry-After` handling — so hitting a 429 partway through a large workspace is the expected outcome, not an edge case.

Suggested shape: catch per page, and on failure return what was collected plus a marker, e.g.
```python
try:
    result = _request("GET", "conversations.list", params=params)
except Exception as e:
    logger.warning(f"Slack pagination stopped early: {e}")
    return _success(channels=channels, truncated=True, error=str(e))
```
plus honouring `Retry-After` on 429. Worth a test too — the current pagination test only covers all-success, and the failure test fails on the first page where there's nothing to lose.

rogercloud (commented):
## Summary

Registers Slack as a built-in OAuth MCP connector. The change adds an `oauth_providers` row (own `client_id`/`client_secret`, `auth_url`/`token_url`/`userinfo_url`, `default_scopes`) plus a `public_mcp_apps` row to `src/xagent/web/builtin_mcp_registry.py`, mirrors both as a frozen snapshot in a new seed migration `src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py`, and introduces `src/xagent/web/tools/mcp/slack.py` — a FastMCP stdio module exposing `slack_list_channels` and `slack_post_message`. `example.env` documents the three new env vars. 774 added lines across 6 files, no existing lines modified.

Notably, `src/xagent/web/api/auth.py` and `src/xagent/web/tools/config.py` are untouched: the PR relies entirely on the existing generic OAuth exchange/refresh path, on the premise that Slack's quirks (credentials in the POST body, non-expiring bot tokens) are already covered there. I verified that premise for token exchange, non-expiry, and the `env_mapping` launch path — it holds. It does *not* fully hold for the userinfo-fetch path (see MAJOR 2).

## What changed since the previous review

Three commits, the latter two responding to the earlier (now-resolved) bot review:

- `c62e9b3c` — `feat`: initial connector (registry rows, migration, `slack.py`, tests, env docs).
- `da0d0e4f` — `fix`: write payloads (`chat.postMessage`) now go in a JSON body via a new `json_data` parameter on `_request()` instead of query params; `exclude_archived` is explicitly serialized as lowercase `"true"`/`"false"`.
- `7aabfa36` — `docs`: added a comment justifying `email_path: "team"` rather than changing it.

All 8 previously-opened inline threads are resolved. I re-verified each below rather than taking the resolution at face value.

The connector's shape faithfully mirrors the established house pattern — registry provider+app rows, a seed migration duplicating them as a frozen snapshot guarded by a registry-equality test, `python -m` stdio launch with `env_mapping`, and local `_success`/`_error`/`_request` helpers matching `src/xagent/web/tools/mcp/hubspot.py`. Two non-obvious design choices I specifically checked and consider sound:

- **`env_mapping: {"SLACK_ACCESS_TOKEN": "access_token"}`** is consumed correctly by `src/xagent/web/tools/config.py`; the launch path works as written.
- **`email_path: "team"`** (storing the Slack workspace name in the email column) is safe. `UserOAuth.email` in `src/xagent/web/models/user_oauth.py` has no format validation and no uniqueness constraint (the real constraint is `(user_id, provider, provider_user_id)`), it is consumed only as a display label in `src/xagent/web/api/mcp.py`'s connected-account enrichment, and Gmail-specific logic is provider-gated elsewhere, so a Slack "email" cannot leak into Gmail provisioning. Worst case is a cosmetic display oddity.

Reservations:

1. **The "no `auth.py` changes needed" claim is only partially true.** Reusing the generic userinfo path for a provider that encodes errors inside HTTP 200 is the one place the reuse assumption breaks — see MAJOR 2.
2. **Nothing in-repo consumes this connector yet.** `src/xagent/templates/built_in/operations-devops-ai-agent.yaml` — the template this PR is described as being for — still has `connections: []`, and its prompt explicitly tells the agent it has no Slack connection. Scope observation, not a blocker: the connector is independently usable and testable, but the PR's stated motivation isn't realized until a follow-up wires it in. Worth confirming that follow-up is tracked.

## Prior findings

| # | Prior finding | Status |
|---|---|---|
| 1 | Write ops sent parameters as query params instead of a JSON body | **FIXED** (`da0d0e4f`) — `_request()` takes `json_data`, `slack_post_message` uses it, tests assert on `call_kwargs["json"]`. `slack_list_channels` correctly still uses `params` for its GET; no regression. |
| 2 | `exclude_archived` passed as a raw Python bool (serializes to `"True"`) | **FIXED** (`da0d0e4f`) — explicit `"true"`/`"false"`, with an explanatory comment. Verified by running the specific test. |
| 3 | `email_path` should be `"email"`, not `"team"` | **WAIVED** (`7aabfa36`) — kept as `"team"` with a justifying comment. I independently checked the justification (see Design verdict) and found no functional risk. A legitimate waive. |

## Findings

### Major

**1. Mid-pagination failure discards every already-fetched page; no rate-limit handling.** `src/xagent/web/tools/mcp/slack.py:81-107`. The `for _ in range(MAX_PAGES)` loop calls `_request` per page with no per-page `try`/`except`, so a failure on page N propagates to the outer `except Exception` at line 105, which throws away all channels accumulated from pages 1..N-1 and returns a bare error. This is plausible rather than theoretical: `conversations.list` is Tier-2 rate-limited (~20 req/min) and the loop fires up to 20 back-to-back requests with zero delay, no backoff, and no `Retry-After` handling. Existing tests don't cover this — pagination is only tested all-success (2 pages), and the failure test only fails on page 1, where there is nothing to lose.

**2. The reused `auth.test` userinfo path checks HTTP status only, silently swallowing Slack's HTTP-200-encoded errors.** `src/xagent/web/builtin_mcp_registry.py:123` points `userinfo_url` at `https://slack.com/api/auth.test`. The generic (pre-existing, unmodified) userinfo fetch in `src/xagent/web/api/auth.py:1355-1358` decides success purely on `info_response.status_code == 200` — but Slack's Web API always returns 200 and signals failure only via `{"ok": false, "error": ...}` in the body. A failed `auth.test` (bad or revoked token) is therefore treated as success: `provider_user_id` and `email` both resolve to `None`, the connection is persisted with no error, and the UI reports "connected" with no visible sign of a problem.

To be clear about scope: this is a **pre-existing gap in the generic path, newly exposed by this PR's choice to reuse it for a provider with non-standard error semantics** — `auth.py` was not edited here, so this isn't a bug introduced by the diff. It's still worth addressing in this PR because this PR is what makes it reachable. What makes it stand out is that `slack.py`'s own `_request()` checks the `ok` field for exactly this reason, so the quirk was understood — it just wasn't carried over to the path the registry change now depends on.

### Minor

**3. `slack_list_channels` has no filter or limit parameter despite advertising itself for name resolution, and `MAX_PAGES` truncates silently.** `src/xagent/web/tools/mcp/slack.py:71-107`. The docstring says "resolve a channel name to an id... find the right channel when the user names one imprecisely," yet the only parameter is `exclude_archived`. In a workspace with more than a couple hundred channels, resolving a single name means dumping up to 4000 entries into the model's context. Separately, if the public-channel count exceeds the pagination cap, line 104 returns `_success(channels=channels)` with a partial list and zero signal — no `truncated` flag, no warning log. Both are graceful-degradation issues affecting larger workspaces rather than crashes; not blocking.

**4. `slack_post_message`'s docstring claims a bare channel name works, but that's untested and unverified.** `src/xagent/web/tools/mcp/slack.py:114` documents `"incidents"` (no `#`) as acceptable; the code does no normalization and simply forwards the string. Tests cover `"C0123"` and `"#incidents"` only.

**5. The migration's `downgrade()` provenance guard doesn't do what its comment claims, and is disproportionately complex for what it does protect.** `src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py:163-179`. The guard matches on `name`/`auth_url`/`token_url` before deleting, with a comment saying this preserves an admin-created Slack provider. But an admin setting up a *real* Slack integration via `POST /admin/mcp/providers` would enter Slack's canonical name and URLs — matching the guard exactly and being deleted anyway, the opposite of the comment's claim. The only case actually protected is a differently-named or differently-URLed row, which is the less realistic one. `tests/alembic/test_20260801_seed_slack_mcp_app.py:174`'s `test_downgrade_preserves_admin_created_slack_provider` only exercises that differently-shaped row, never the canonical-match case. Sibling migrations (`20260627_seed_meta_connectors.py`, `20260703_seed_google_maps_mcp_app.py`) delete unconditionally by `provider_name`/`app_id`.

**6. Test coverage gaps.** In `tests/web/tools/test_slack_mcp.py`: no test exercises the default `exclude_archived=True` → `"true"` path (only `False` → `"false"` is asserted); no test covers a payload missing the `"channels"` key; no test verifies that the *tool functions* (as opposed to `_headers()`/`_request()` directly) convert a missing-token `ValueError` or an `HTTPError` into the JSON error envelope; and `test_post_message_returns_error_payload_on_channel_not_found` / `..._on_not_in_channel` are near-duplicates of the same path differing only in the error string. In `tests/alembic/test_20260801_seed_slack_mcp_app.py`: no test exercises a reduced/partial-column schema (so `_filter_row`'s filtering behaviour is never actually tested) or a "one table present, one absent" mix, and `test_seed_rows_match_registry` takes an unused `tmp_path` fixture.

**7. `tests/web/api/test_public_mcp_connector_visibility.py`'s `expected_python_apps` dict wasn't extended with `slack`.** Because that test iterates `expected_python_apps.keys()`, it silently skips validating slack's `launch_config`. A real safety net does exist — `tests/alembic/test_20260801_seed_slack_mcp_app.py:122`'s `test_seed_rows_match_registry` does full dict equality between the live registry row and the migration's frozen copy, covering `launch_config`/`args`/`env_mapping` — so a typo would still fail CI there (unless identically copy-pasted into both files). Adding the entry is a consistency fix, not a coverage hole.

**8. `chat:write.public`'s effect isn't called out.** `example.env:426` and the connector description at `src/xagent/web/builtin_mcp_registry.py:403` list the scope without noting that `chat:write.public` lets the bot post into *any* public channel without being invited. This matches codebase convention (Drive/Outlook/Teams entries are equally terse), but this scope bypasses a platform-side admission control rather than merely widening data access, and no other connector in the repo has an analogous scope — a one-line callout is worth more here specifically. Slack's own consent screen does disclose it before granting, so this is product clarity, not an authorization gap.

**9. Slack scope separator is unpinned by any test.** `src/xagent/web/api/auth.py:123-126`'s `_oauth_scope_separator()` special-cases only `meta` → comma; Slack falls through to space-joined, while Slack's docs specify comma-separated. The manual end-to-end verification described in the PR (real OAuth connect plus working `slack_list_channels`/`slack_post_message` against a live workspace) is direct runtime evidence that the current form works — a scope-string parse failure would surface immediately as a broken or degraded grant, not a silent partial success. So this is not a functional defect, but nothing pins it: a future refactor of the separator logic could silently break Slack with no test signal.

### Informational

**10. The migration writes `SLACK_CLIENT_SECRET` into `oauth_providers.client_secret` in plaintext,** whereas the admin API's `create_provider` path encrypts it. This is a systemic, pre-existing pattern across every seed migration in the repo (meta, hubspot, microsoft-graph seeds all do the identical plaintext write) — this PR does not deviate or make it worse. `decrypt_value` tolerates plaintext without erroring, and `_resolve_oauth_secret` falls back to the env var when the column is empty, so there's no functional break. Raising it as a systemic follow-up against the shared seeding convention, explicitly not against this PR.

- `yagni:` `src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py:36-39` — `OAUTH_PROVIDERS_TABLE` exists only to expose `provider_name`, which `FULL_OAUTH_PROVIDERS_TABLE` already has. Use `FULL_OAUTH_PROVIDERS_TABLE.c.provider_name` at the single call site (line 114) and delete the second `sa.table()`. Zero-risk, −3 lines.
- `shrink:` `20260801_seed_slack_mcp_app.py:163-179` — the column-by-column provenance guard in `downgrade()`; folded into Minor 5 above, same code region.
- `yagni:` `20260801_seed_slack_mcp_app.py:60-61, 109-133` — `_filter_row` plus the `inspector.get_columns(...)` guards are unreachable on any real `alembic upgrade head` path. Cross-cutting — better as a follow-up than a change here: the same pattern is copy-pasted across three sibling seed migrations.
- `yagni:` `20260801_seed_slack_mcp_app.py:107-109, 122` — the `existing_tables` guards in `upgrade()` are likewise unreachable. Cross-cutting — better as a follow-up: removing them also requires deleting `test_upgrade_and_downgrade_no_op_without_tables`, which only 2 of 5 comparable seed-migration test files even have.

Net: −3 lines from the first item alone; roughly −40 if the shared convention is revisited, but that belongs in its own cleanup ticket rather than a single-connector PR.

## Test results

All 33 relevant tests pass — `tests/web/tools/test_slack_mcp.py`, `tests/alembic/test_20260801_seed_slack_mcp_app.py`, `tests/templates/test_manager.py` — verified independently several times during this review. A coverage run shows **98% line coverage on `src/xagent/web/tools/mcp/slack.py`**, with only the `if __name__ == "__main__"` guard uncovered. The gaps noted in Minor 6 are behavioural (untested branches reachable via different inputs), not line-coverage gaps.

## Verdict

No blocking defects. MAJOR 2 is the one I'd want addressed or explicitly triaged before merge, since it makes a failed OAuth grant indistinguishable from a successful one in the UI; MAJOR 1 is a robustness fix that's cheap now and annoying to retrofit later. Everything else is fine as follow-up.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -112,6 +112,24 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
...
         },
+        {
+            "provider_name": "slack",
+            "name": "Slack",
+            "client_id": os.environ.get("SLACK_CLIENT_ID", ""),
+            "client_secret": os.environ.get("SLACK_CLIENT_SECRET", ""),
+            "auth_url": "https://slack.com/oauth/v2/authorize",
+            "token_url": "https://slack.com/api/oauth.v2.access",
+            "redirect_uri": os.environ.get("SLACK_REDIRECT_URI", ""),
+            "userinfo_url": "https://slack.com/api/auth.test",
```
rogercloud, line 123: **MAJOR — pointing `userinfo_url` at `auth.test` reuses a generic path that can't detect Slack's failures.**

The generic userinfo fetch in `src/xagent/web/api/auth.py:1355-1358` decides success on `info_response.status_code == 200` alone. Slack's Web API always returns HTTP 200 and signals failure only via `{"ok": false, "error": ...}` in the body — so a failed `auth.test` (bad or revoked token) is treated as success: `provider_user_id` and `email` both resolve to `None`, the connection is persisted with no error, and the UI shows "connected" with nothing indicating a problem.

Scope note: this is a pre-existing gap in the generic path (`auth.py` isn't touched by this PR), but this line is what makes it reachable. Worth flagging that `slack.py`'s own `_request()` checks the `ok` field for precisely this reason — the quirk was understood, it just wasn't carried across to the path this registry entry depends on.

Two options: teach the generic handler to treat a JSON body with `ok: false` as a failure (helps every future provider with these semantics), or gate connection persistence on a non-`None` `provider_user_id`.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+    response.raise_for_status()
+    payload: dict[str, Any] = response.json()
+    if not payload.get("ok"):
+        raise RuntimeError(payload.get("error") or "Unknown Slack API error")
+    return payload
+
+
+@mcp.tool()
+def slack_list_channels(exclude_archived: bool = True) -> str:
```
rogercloud: **MINOR — the signature can't support what the docstring advertises.**

The docstring below says this is for resolving a channel name to an id and finding the right channel "when the user names one imprecisely," but `exclude_archived` is the only parameter. In a workspace with a few hundred-plus channels, resolving one name means dumping up to 4000 entries into the model's context.

Suggest adding `name_contains: str | None = None` and a `limit: int` cap. Related: when a workspace exceeds `MAX_PAGES`, line 104 returns a partial list with no `truncated` flag and no warning log — a caller can't distinguish "these are all the channels" from "these are the first 4000."

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,130 @@
...
+    except Exception as e:
+        logger.error(f"Error listing Slack channels: {e}")
+        return _error(str(e))
+
+
+@mcp.tool()
+def slack_post_message(channel: str, text: str) -> str:
+    """
+    Post a message to a Slack channel.
+    channel: a channel id (e.g. "C0123456789") or name (e.g. "#incidents" or "incidents").
```
rogercloud: **MINOR — the bare-name form is documented but neither normalized nor tested.**

The docstring offers `"incidents"` (no `#`) as valid, but the code performs no normalization and forwards the string as-is. Tests cover `"C0123"` and `"#incidents"` only — nothing exercises the bare form against real API semantics. Either add a test pinning it down, or remove that form from the docstring.

src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
```diff
@@ -0,0 +1,179 @@
...
+    if "public_mcp_apps" in existing_tables:
+        remaining_slack_apps = bind.execute(
+            sa.select(sa.func.count())
+            .select_from(PUBLIC_MCP_APPS_TABLE)
+            .where(PUBLIC_MCP_APPS_TABLE.c.provider_name == "slack")
+        ).scalar()
+        if remaining_slack_apps:
+            return
+
+    # Only delete the provider row when it still matches the static shape this
```
rogercloud: **MINOR — this guard protects the opposite of what the comment claims.**

The comment says matching `name`/`auth_url`/`token_url` preserves an admin-created Slack provider. But an admin setting up a *real* Slack integration via `POST /admin/mcp/providers` would enter Slack's canonical name and OAuth URLs — matching this guard exactly, and getting deleted. The only row actually preserved is one with a different name or different URLs, which is the less realistic scenario. `test_downgrade_preserves_admin_created_slack_provider` reflects this: it only ever tests the differently-shaped row, never the canonical-match case the comment describes.

Sibling seed migrations (`20260627_seed_meta_connectors.py`, `20260703_seed_google_maps_mcp_app.py`) just delete unconditionally by `provider_name`. Suggest matching them, accepting the same trade-off every other seed migration already accepts, and dropping ~15 lines of `inspector`/loop ceremony. If the protection really is wanted, please reword the comment to state what it actually protects.

src/xagent/migrations/versions/20260801_seed_slack_mcp_app.py
```diff
@@ -0,0 +1,179 @@
...
+    sa.column("auth_url", sa.String),
+    sa.column("token_url", sa.String),
+    sa.column("redirect_uri", sa.String),
+    sa.column("userinfo_url", sa.String),
+    sa.column("user_id_path", sa.String),
+    sa.column("email_path", sa.String),
+    sa.column("default_scopes", sa.JSON),
+)
+
+OAUTH_PROVIDERS_TABLE = sa.table(
```
rogercloud: `yagni:` This second `sa.table()` exists only to expose `provider_name`, which `FULL_OAUTH_PROVIDERS_TABLE` already declares. Use `FULL_OAUTH_PROVIDERS_TABLE.c.provider_name` at the single call site (line 114) and delete lines 36-39. Zero behaviour change, −3 lines.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -112,6 +112,24 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
...
+            # auth.test never returns an email for a bot token; the workspace
+            # name ("team") is deliberately stored in the email slot because
+            # UserOAuth.email is only consumed as the "connected account"
+            # display label for non-gmail providers — without it the Slack
+            # connection would show up unlabeled in the connector UI.
+            "user_id_path": "team_id",
+            "email_path": "team",
+            "default_scopes": ["chat:write", "chat:write.public", "channels:read"],
```
rogercloud, line 131: **MINOR — worth a regression test pinning the Slack scope-string form.**

`_oauth_scope_separator()` in `src/xagent/web/api/auth.py:123-126` special-cases only `meta` → comma, so these scopes are joined with spaces, while Slack's docs specify comma-separated. The manual end-to-end verification in the PR description is good evidence the current form works in practice, so I'm not flagging this as a defect.

The gap is that nothing pins it: a future refactor of the separator logic could silently break Slack with no test signal. A one-line assertion on the scope string actually sent for `slack` would lock in what was manually verified.

example.env
```diff
@@ -418,6 +418,20 @@ META_REDIRECT_URI=""
 # configurations for Pages and Instagram permissions.
 META_CONFIG_ID=""
 
+# ===========================================
+# Slack OAuth Configuration (Optional)
+# ===========================================
+# Required for the Slack integration in MCP (list channels, post messages)
+# Create an app at https://api.slack.com/apps, add these scopes under
+# "OAuth & Permissions": chat:write, chat:write.public, channels:read
```
rogercloud, line 462: `chat:write.public` deserves one more line here. Unlike the other two scopes it lets the bot post into *any* public channel **without being invited** — it bypasses Slack's invite gate rather than just widening data access.

This matches convention elsewhere in the registry (Drive/Outlook/Teams entries are equally terse), but no other connector here has a scope that bypasses a platform-side admission control, so a callout is worth more in this one case. Slack's own consent screen does disclose it, so this is operator clarity rather than a security gap. Suggest something like: `# chat:write.public also lets the bot post into any public channel without being invited.`

tests/web/tools/test_slack_mcp.py
```diff
@@ -0,0 +1,207 @@
...
+    result = json.loads(slack.slack_list_channels())
+
+    assert result["status"] == "success"
+    assert result["channels"] == [
+        {"id": "C1", "name": "general", "is_archived": False},
+        {"id": "C2", "name": "incidents", "is_archived": False},
+    ]
+
+
+def test_list_channels_passes_exclude_archived_and_types(monkeypatch):
```
rogercloud: `nit:` This asserts the `exclude_archived=False` → `"false"` path, but the **default** (`True` → `"true"`) is never exercised — and the default is what production traffic actually hits. Given `da0d0e4f` was specifically about this serialization, pinning both values (a `@pytest.mark.parametrize` over `(True, "true")` / `(False, "false")`) would be a better guard.

tests/web/tools/test_slack_mcp.py
```diff
@@ -0,0 +1,207 @@
...
+        return_value=MockResponse({"ok": True, "channel": "C0123", "ts": "1.1"})
+    )
+    monkeypatch.setattr(slack.requests, "request", mock_request)
+
+    json.loads(slack.slack_post_message("#incidents", "hello"))
+
+    assert mock_request.call_args.kwargs["json"]["channel"] == "#incidents"
+
+
+def test_post_message_returns_error_payload_on_channel_not_found(monkeypatch):
```
rogercloud: `nit:` This and `test_post_message_returns_error_payload_on_not_in_channel` below exercise the identical code path, differing only in the error string Slack returns. Good candidate for a single parametrized test over `["channel_not_found", "not_in_channel"]`.

While here — two behaviours reachable from these tool entry points are untested: a missing `SLACK_ACCESS_TOKEN` (the `ValueError` from `_headers()`) and an `HTTPError`, converted to the JSON error envelope. Both are tested against `_headers()`/`_request()` directly, but not through `slack_list_channels()`/`slack_post_message()`, which is what the model actually calls. A payload missing the `"channels"` key is also uncovered.

tests/alembic/test_20260801_seed_slack_mcp_app.py
```diff
@@ -0,0 +1,210 @@
...
+            text("SELECT COUNT(*) FROM public_mcp_apps WHERE app_id='slack'")
+        ).scalar()
+        assert app_count == 1
+        provider_count = connection.execute(
+            text("SELECT COUNT(*) FROM oauth_providers WHERE provider_name='slack'")
+        ).scalar()
+        assert provider_count == 1
+
+
+def test_seed_rows_match_registry(tmp_path):
```
rogercloud: `nit:` `tmp_path` is unused here — this test compares the registry dict against the migration's frozen copy without touching a database. Safe to drop the parameter.

Credit where due: this test is the real safety net for the slack `launch_config`, since `tests/web/api/test_public_mcp_connector_visibility.py`'s `expected_python_apps` dict wasn't extended with a `slack` entry and its loop therefore skips validating it. Adding that entry would still be worthwhile for consistency, but it's not a coverage hole given this check exists.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,195 @@
...
+                # which Slack's parser doesn't recognize as a boolean.
+                "exclude_archived": "true" if exclude_archived else "false",
+                "limit": 200,
+            }
+            if cursor:
+                params["cursor"] = cursor
+            try:
+                result = _request("GET", "conversations.list", params=params)
+            except Exception as page_exc:
+                if not channels:
```
rogercloud: **MINOR (new in this commit).** `if not channels: raise` conflates two different states: "we haven't successfully fetched anything yet" and "we scanned page 1 fine, but `name_contains` filtered every result out." In the second case a page-2 failure re-raises and the caller gets a hard error, when the accurate answer is "no matches so far, then a rate limit." Consider tracking pages-scanned (or raw results seen) separately from the filtered `channels` list and keying the re-raise on that instead.

Separately, this rescue path returns `"status": "success"` for what is really a partial, rate-limited scan. `truncated` and `error` are both present so a careful caller can tell, but a caller that branches only on `status` will treat it as a complete channel list. A distinct `"partial"` status (or at least a docstring note) would make that harder to get wrong.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,195 @@
...
+                if needle and needle not in name.lower():
+                    continue
+                channels.append(
+                    {
+                        "id": channel.get("id"),
+                        "name": channel.get("name"),
+                        "is_archived": channel.get("is_archived", False),
+                    }
+                )
+                if len(channels) >= max_channels:
```
rogercloud, line 165: **MINOR (new in this commit).** `truncated=True` here is a false positive when the limit is reached on exactly the last available page — the check is purely `len(channels) >= max_channels` with no look-ahead at whether `next_cursor` was actually returned, so a workspace with exactly `limit` matching channels reports truncation that doesn't exist.

Related: none of the four `_success()` call sites include a `next_cursor`/resume token, so even when truncation is genuine the caller has no way to continue from where the scan stopped. Both are ergonomics gaps rather than correctness bugs, but they're cheap to close together.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,195 @@
...
+MAX_PAGES = 20
+MAX_CHANNELS = 1000
+# conversations.list is Tier-2 rate-limited (~20 req/min); on a 429 with a
+# small Retry-After we wait once and retry rather than failing the page.
+MAX_RETRY_AFTER_SECONDS = 30
+
+# Slack channel/user/group ids are uppercase alphanumerics with a letter
+# prefix; channel *names* are forced lowercase by Slack, so this can't
+# misclassify a name as an id.
+_SLACK_ID_PATTERN = re.compile(r"^[A-Z][A-Z0-9]{5,}$")
```
rogercloud, line 32: **MINOR, test-only.** `_SLACK_ID_PATTERN` requires at least 6 characters (`^[A-Z][A-Z0-9]{5,}$`), but three tests in `tests/web/tools/test_slack_mcp.py` pass `"C0123"` (5 chars) as the channel: `test_post_message_returns_error_payload_on_slack_error`, `test_post_message_reports_http_error_through_the_tool`, and `test_post_message_reports_missing_token_through_the_tool`. Those inputs get silently rewritten to `"#C0123"` by `_normalize_channel`, and it's invisible because none of the three assert the outgoing channel value.

Real Slack channel IDs are always ≥9 characters, so the pattern itself is fine — this is a fixture inconsistency. Switching those three to `"C0123456789"` (as the other tests already use) keeps the fixtures honest about which branch they're exercising.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,195 @@
...
+    returned (more pages, the limit was hit, or a page failed mid-way — a
+    partial list is returned rather than discarded).
+    """
+    channels: list[dict[str, Any]] = []
+    needle = name_contains.strip().lower()
+    max_channels = max(1, min(int(limit), MAX_CHANNELS))
+    cursor: str | None = None
+    truncated = False
+    try:
+        for page_index in range(MAX_PAGES):
```
rogercloud: `shrink:` `page_index` is bound but never used in the loop body — `for _ in range(MAX_PAGES):` as before. Zero-risk.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -0,0 +1,195 @@
...
+            # MAX_PAGES exhausted with a cursor still pending.
+            truncated = bool(cursor)
+        return _success(channels=channels, truncated=truncated)
+    except Exception as e:
+        logger.error(f"Error listing Slack channels: {e}")
+        return _error(str(e))
+
+
+@mcp.tool()
+def slack_post_message(channel: str, text: str) -> str:
```
rogercloud, line 190: **MINOR, informational — matches existing project convention, optional hardening.** With `chat:write.public` in the requested scopes, `slack_post_message` can post to any public channel without an invite, and the tool accepts an arbitrary channel and arbitrary text with no allow-list, rate limit, or confirmation step. There is no enforced tool-approval layer in the agent framework either (the ReAct pattern's `send_message`/`ask_user_question` are voluntary, model-invoked affordances, not an interception point), so a prompt-injection payload reaching this tool can post anywhere in the workspace.

To be clear, this is not a Slack-specific regression: `teams.py`'s `send_channel_message`/`send_chat_message` and `outlook.py`'s `outlook_send_message` are equally unrestricted, so this PR is following house convention. The one asymmetry worth noting is that Teams/Outlook require the caller to already have access to the target, whereas `chat:write.public` deliberately discards the gate Slack itself designed in — which makes the blast radius here wider than its siblings. An opt-in `SLACK_ALLOWED_CHANNELS` env guard would cap it cheaply, but I'd treat that as optional rather than a condition of merge.

tests/web/test_slack_oauth.py
```diff
@@ -0,0 +1,175 @@
...
+            "user_id": user.id,
+            "provider": "slack",
+            "app_id": "slack",
+        },
+        expires_delta=timedelta(minutes=10),
+    )
+    return SimpleNamespace(query_params={"code": "slack-code", "state": state})
+
+
+def test_slack_callback_persists_workspace_identity(db_session, monkeypatch):
```
rogercloud, line 90: **MINOR.** This test mocks `requests.post` but never inspects `.call_args`/`.call_args.kwargs`, so the actual token-exchange request shape goes unasserted. That matters here specifically because the PR's stated reason for not modifying the token-exchange code is "Slack accepts `client_id`/`client_secret` in the POST body" — that claim is currently pinned by nothing, and a future refactor of the shared exchange path (e.g. moving credentials to a Basic auth header for another provider) would break Slack silently.

Adding an assertion on the posted `data` payload would turn the PR's justification into a regression test. Note the sibling `test_meta_oauth.py` has the same gap on the same shared path, so this isn't Slack-specific — but it's the PR that rests an argument on it.

tests/web/test_slack_oauth.py
```diff
@@ -0,0 +1,175 @@
...
+    create_access_token,
+    generic_oauth_callback,
+)
+from xagent.web.models.database import Base
+from xagent.web.models.public_mcp import PublicMCPApp
+from xagent.web.models.user import User
+from xagent.web.models.user_oauth import UserOAuth
+
+
+class MockResponse:
```
rogercloud, line 25: `shrink:` This `MockResponse` (with `text`) and the one added in `tests/web/tools/test_slack_mcp.py:10` (with `headers` + `raise_for_status()`) are near-duplicates introduced in the same commit. The repo already has a shared test-mock factory at `tests/utils/mock_helpers.py` that other suites use for exactly this. Consolidating into a single `MockResponse`/`create_mock_response()` there — `text` and `headers` both optional, `raise_for_status()` included — and importing it from both files would drop the duplication.

rogercloud (approved):
## Follow-up review (2nd pass) — commit `ca129605`

### PR summary

This PR adds a Slack MCP connector: an Alembic seed migration registering the Slack OAuth app and its scopes, a `builtin_mcp_registry` entry, and a `slack.py` MCP tool module exposing `slack_list_channels` and `slack_post_message` against Slack's Web API, plus `example.env` documentation for the new client credentials. It follows the same shape as the existing Meta / Google Maps / Teams connectors, reusing the shared generic OAuth callback rather than adding a Slack-specific auth path.

### Update summary

One new commit, `ca129605` ("fix: address PR review on the Slack connector"), responds point-by-point to the previous review. Both MAJOR findings are fixed, along with nearly all MINOR/nit items. Notably the author went beyond the connector itself and hardened `src/xagent/web/api/auth.py` (previously untouched by this PR) so the shared userinfo step no longer silently accepts Slack's HTTP-200-encoded errors, and added a new `tests/web/test_slack_oauth.py` (+175) plus a substantial expansion of `tests/web/tools/test_slack_mcp.py` (207 → 370 lines).

### Design verdict: acceptable-with-reservations

The connector continues to match the established house pattern. The `auth.py` change is well-scoped and, in my view, the right call: it is provider-agnostic rather than Slack-special-cased, guarded by `isinstance(...)` plus a strict `is False` (so it cannot false-positive on the five other providers sharing that path, none of which return an `ok` key), it fires before any `UserOAuth` row is persisted, and it does not add another `provider.lower() == "..."` branch to the existing conditionals. That is a generic hardening of a shared path, done correctly.

Reservations are narrower than last round and **none are blocking**:

- Two sibling silent-success paths remain in the same `generic_oauth_callback` function (pre-existing, untouched by this PR) — see New findings #1/#2, filed as follow-up material only.
- The new pagination-rescue and `limit` logic in `slack_list_channels` has two small edge cases (#3, #4).
- `chat:write.public` plus an unrestricted `slack_post_message` is a real but project-wide pattern, not Slack-specific (#6).

### Prior-findings checklist

| # | Prior finding | Status |
|---|---|---|
| 1 | MAJOR — mid-pagination failure discarded all fetched channels; no rate-limit handling | **FIXED** — per-page `try/except` returns `_success(channels=..., truncated=True, error=...)`; first-page failure still raises (correct). `_request` gained a bounded single 429 retry honoring `Retry-After` (capped at `MAX_RETRY_AFTER_SECONDS=30`). Tests added. |
| 2 | MAJOR — reused `auth.test` userinfo path checked HTTP status only, swallowing Slack's HTTP-200 errors | **FIXED** — provider-agnostic `ok is False` guard in `src/xagent/web/api/auth.py` returns HTTP 400 with the escaped provider error before persisting. `test_slack_callback_fails_when_auth_test_reports_ok_false` asserts no row is written; no regressions across the broader OAuth suite. |
| 3 | MINOR — `slack_list_channels` had no filter/limit despite its docstring; silent truncation | **FIXED** — `name_contains` and `limit` params added; `truncated` now set on both `limit` and `MAX_PAGES`-with-pending-cursor. |
| 4 | MINOR — `slack_post_message` docstring claimed bare-name support, untested | **FIXED** — `_normalize_channel()` added, docstring updated, parametrized tests cover the bare-name case. |
| 5 | MINOR — migration `downgrade()` provenance guard didn't match its comment | **FIXED** — unconditional delete by `provider_name == "slack"`, matching sibling migrations; comment now accurate; test inverted accordingly. |
| 6 | yagni — duplicate `OAUTH_PROVIDERS_TABLE` definition | **FIXED** — removed; call site uses `FULL_OAUTH_PROVIDERS_TABLE.c.provider_name`. |
| 7 | MINOR — Slack scope separator unpinned by any test | **FIXED** — `test_slack_scopes_join_with_spaces` pins `_oauth_scope_separator("slack") == " "` and the joined string, documented as an intentional regression pin. |
| 8 | MINOR (docs) — `chat:write.public` invite-bypass undocumented | **FIXED** — noted in `example.env`. (The registry description text still omits it; the finding only asked for `example.env`.) |
| 9 | nit — `exclude_archived` `True`→`"true"` path untested | **FIXED** — parametrized over both values. |
| 10 | nit — duplicate `channel_not_found`/`not_in_channel` tests | **FIXED** — consolidated into one parametrized test. |
| 11 | nit — unused `tmp_path` fixture | **FIXED** |
| 12a | missing `"channels"` key untested | **FIXED** — `test_list_channels_tolerates_missing_channels_key`. |
| 12b | tool-level `ValueError` → error envelope untested | **FIXED** — covered for both tools. |
| 12c | tool-level `HTTPError` → error envelope untested | **PARTIAL** — covered for `slack_post_message` only; no equivalent test for `slack_list_channels` converting an `HTTPError`/5xx through the tool function (only `_request()` directly). Small residual gap. |
| 12d | migration `_filter_row` partial-column / one-table-present cases untested | **NOT FIXED** — only the both-tables-absent case exists. Low priority: those branches are unreachable on any real `alembic upgrade head` path. |
| 13 | `expected_python_apps` missing a `slack` entry | **FIXED** — entry added with correct module path and `env_mapping`. |
| 14 | informational — migration writes `SLACK_CLIENT_SECRET` in plaintext | **WAIVED** — confirmed untouched, as expected; systemic pre-existing pattern across all seed migrations. |
| 15 | design observation — no in-repo consumer | **STILL OPEN** — `src/xagent/templates/built_in/operations-devops-ai-agent.yaml:19` still has `connections: []` and its prompt still says the agent has no Slack connection. The connector is seeded and independently testable but not wired to any consumer. Not blocking — is a follow-up PR planned? |

### New findings

No MAJOR findings this round.

**1. (Informational — pre-existing, NOT introduced by this PR, NOT blocking) Two sibling silent-success paths in `generic_oauth_callback` remain asymmetric with the new guard.** In `src/xagent/web/api/auth.py`: (a) `if info_response.status_code == 200:` has no `else`, so a non-200 userinfo response skips identity extraction and the code below still persists a "connected" `UserOAuth` row with `provider_user_id=None`/`email=None`; (b) `if userinfo_url and access_token:` — a 200 token-exchange response that omits `access_token` without an `error` key still persists a row with `access_token=None`. I verified both are byte-identical before and after `ca129605`; they predate Slack support and affect every generic-OAuth provider. Suggest a separate systemic hardening ticket (add `else` branches failing the callback, plus tests) rather than expanding this PR.

**2. (Informational — pre-existing, NOT blocking) Token-exchange uses a weaker Slack-error idiom than the new userinfo check, in the same function.** `src/xagent/web/api/auth.py:1334` detects the same Slack HTTP-200 quirk via `"error" in token_data` rather than an `ok is False` check. It works today (Slack always populates `error` alongside `ok:false`) but is inconsistent with the more precise check added a few lines below, and no test exercises the token-exchange failure path at all. Same follow-up ticket as #1.

**3. (MINOR) The pagination rescue re-raises when a filter legitimately matched nothing on page 1.** `src/xagent/web/tools/mcp/slack.py:142-143` — the `if not channels: raise` guard cannot distinguish "nothing fetched yet" from "pages scanned, but `name_contains` filtered everything out". With an active filter that matches nothing on page 1 and a failure on page 2, the caller gets a hard error instead of the accurate "no matches so far, then a rate limit". Track pages-scanned (or raw-results-seen) separately from filtered `channels`. Separately, the rescue labels a rate-limited partial scan as `"status": "success"`; `truncated`+`error` are both present for a careful caller, but a caller branching only on `status` could treat it as a complete list. Newly introduced by this commit (only reachable via the just-added filter).

**4. (MINOR) `truncated=True` can be a false positive at the `limit` boundary, and no resume token is ever returned.** `src/xagent/web/tools/mcp/slack.py:160-161` — the check is purely `len(channels) >= max_channels` with no look-ahead at whether a `next_cursor` actually exists, so hitting the limit on exactly the last available page reports truncation that isn't real. Relatedly, none of the four `_success()` call sites include a `next_cursor`/resume token, so even genuine truncation leaves the caller unable to continue. Both are ergonomics gaps in the new code, not correctness bugs.

**5. (MINOR, test-only) Three tests use a channel fixture that the new ID pattern rejects.** `_SLACK_ID_PATTERN` (`^[A-Z][A-Z0-9]{5,}$`, min 6 chars) does not match `"C0123"` (5 chars), so `_normalize_channel` silently rewrites it to `"#C0123"` in `test_post_message_returns_error_payload_on_slack_error`, `test_post_message_reports_http_error_through_the_tool`, and `test_post_message_reports_missing_token_through_the_tool`. None of them assert the outgoing channel value, so it is invisible. Real Slack IDs are ≥9 chars, so this is a fixture inconsistency, not a functional bug — use `"C0123456789"` like the other tests.

**6. (MINOR, informational — matches existing project convention) `chat:write.public` plus an unrestricted `slack_post_message` has a real prompt-injection blast radius.** There is no allow-list, rate limit, or confirmation step, and no enforced tool-approval/HITL interception layer in the agent framework (`send_message`/`ask_user_question` in the ReAct pattern are voluntary, model-invoked affordances). I confirmed the identical unrestricted pattern in `teams.py`'s `send_channel_message`/`send_chat_message` and `outlook.py`'s `outlook_send_message`, so this is project-wide and Slack should not be singled out. That said, Slack's case is meaningfully wider because `chat:write.public` discards a safety gate Slack itself designed in, while Teams/Outlook require the caller to already have channel access. Optional hardening: an opt-in `SLACK_ALLOWED_CHANNELS` env guard.

**7. (MINOR) The new happy-path OAuth test doesn't pin the token-exchange request shape.** `tests/web/test_slack_oauth.py:90` (`test_slack_callback_persists_workspace_identity`) mocks `requests.post` but never inspects `.call_args`. The PR's headline justification for not modifying the token-exchange code is that Slack accepts `client_id`/`client_secret` in the POST body — that claim is currently unpinned by any regression test. The sibling `test_meta_oauth.py` has the same gap for the same shared path, so this isn't Slack-specific, but it's worth closing here given the PR explicitly rests its argument on this request shape.

- `shrink:` `src/xagent/web/tools/mcp/slack.py:128` — `for page_index in range(MAX_PAGES):` never uses `page_index`; revert to `for _ in range(MAX_PAGES):`. Zero-risk.
- `shrink:` `tests/web/test_slack_oauth.py:25` and `tests/web/tools/test_slack_mcp.py:10` — two near-identical `MockResponse` helpers were added independently (one with `text`, one with `headers` + `raise_for_status()`). The repo already has a shared mock factory at `tests/utils/mock_helpers.py` used by other suites for exactly this; consolidate into one `MockResponse`/`create_mock_response()` there with `text`/`headers` both optional and `raise_for_status()` included.

Net: ~5-10 lines, not blocking.

### Test results

All 91 tests across the five relevant files pass:
- `tests/web/tools/test_slack_mcp.py`
- `tests/web/test_slack_oauth.py` (new)
- `tests/alembic/test_20260801_seed_slack_mcp_app.py`
- `tests/web/api/test_public_mcp_connector_visibility.py`
- (plus the broader `tests/web/ -k oauth` sweep, 262 tests, with no regressions from the `auth.py` change)

### Verdict

**Good to merge.** No blocking issues remain. Both MAJOR findings from the previous review are properly fixed, the `auth.py` hardening is a genuine improvement to a shared path rather than a Slack-specific patch, and test coverage is now solid. The two `auth.py` observations (#1, #2) are pre-existing and belong in a separate systemic ticket; #3-#7 and the two simplification items are worth a small cleanup pass but shouldn't hold this up. The one thing I'd like a word on before/after merge is prior-finding #15 — whether wiring the DevOps agent template to this connector is coming in a follow-up.

Issue #764, opened 2026-07-05
ref(mcp): single source of truth for OAuth vs key-based app classification


bluefish-08:
## Context

Follow-up from PR #753 review. Not a bug today — a latent consistency risk.

Whether a catalog app connects via OAuth or a static API key is decided
independently in (at least) four places, each re-deriving the same
classification from a different field, with no schema-level constraint keeping
them in sync:

| Location | Classifies by |
|---|---|
| Backend connect rejection — `_ensure_catalog_app_server` (`src/xagent/web/api/mcp.py:1980`) | `app_info["transport"] == "oauth"` |
| `ConnectMcpDialog` (`frontend/src/components/mcp/connect-mcp-dialog.tsx:319-323`) | `app.provider` + `launch_config.required_env` |
| `OfficialMcpSettingsDialog` (`frontend/src/components/mcp/official-mcp-settings-dialog.tsx:180`) | `launch_config.required_env.length > 0` |
| Runtime injector — `_is_mcp_oauth_http_server` (`src/xagent/web/services/mcp_runtime.py:283`) | `server.transport` + decrypted `auth.type == "mcp_oauth"` |

## Problem

`transport`, `provider_name`, and `required_env` are all internally authored on
the catalog entry with no constraint tying them together. Nothing prevents an
entry from being set inconsistently (e.g. `transport` not `"oauth"` while
`required_env` is empty), which would route the same app through different flows
on the frontend vs the backend. It works today only because existing entries
happen to be authored consistently — there is no mechanism enforcing it.

## Proposal

Introduce a single canonical `auth_type` enum on the catalog entry
(`builtin_oauth` / `mcp_oauth` / `api_key`) as the one source of truth, and have
all four call sites read that field instead of re-deriving the classification.

**Scope is classification only.** The three underlying connect implementations
are genuinely different credential lifecycles (redirect + refresh /
dynamic-PKCE + refresh / static key with no refresh and no identity semantics)
and should stay separate as implementations — this change only unifies how each
app is *labeled*, not how it is *connected*.

## Touch points

- Catalog schema / data: add `auth_type`, backfill existing entries, migration
- Backend: `src/xagent/web/api/mcp.py`, `src/xagent/web/services/mcp_runtime.py`
- Frontend: `connect-mcp-dialog.tsx`, `official-mcp-settings-dialog.tsx`

PR #1096, opened 2026-07-31
feat: support remote MCP servers with Dynamic Client Registration in the built-in catalog

yiboyasss:
## Summary
- Add a third catalog connection shape (`mcp_oauth`) alongside the existing `builtin_oauth` (static-provider redirect) and `api_key` types — a genuine remote MCP server (streamable_http/sse/websocket transport) authenticated via per-user OAuth Authorization Code + PKCE, using Dynamic Client Registration (RFC 7591) when no static client_id is configured
- `classify_app_auth()` classifies a catalog entry as `mcp_oauth` when its `launch_config` declares both a `url` and `auth.type == "mcp_oauth"` on a remote transport
- Admin catalog validator (`PublicMCPAppCreate`) rejects a partial declaration of this shape (url-without-auth.type / auth.type-without-url), mirroring the existing `api_key` asymmetric-shape guard from #764
- New `POST /api/mcp/apps/{app_id}/oauth/connect` endpoint idempotently ensures the shared server row and the caller's association exist (with the same name-hijack guards as `_ensure_catalog_app_server`), then delegates to the existing `connect_mcp_oauth` flow — the per-user DCR/token machinery is fully reused, no duplicated logic
- Frontend: the connector dialog gains the third connect branch — an `mcp_oauth` app's Connect button POSTs to `/api/mcp/apps/{id}/oauth/connect` (`Accept: application/json`) and drives the returned `authorization_url` in a popup, refreshing connection state when the popup closes; `parseMcpOAuthErrorMessage`/`McpOAuthConnectResponse` move from `custom-mcp-form` to `lib/mcp-utils` so both flows share them
- No registry entry is added yet — this PR is the platform capability only; a follow-up PR adds the first connector using it (Granola, whose remote MCP server at mcp.granola.ai is DCR-only with no static client credentials)

## Why
The built-in catalog currently supports exactly two connection shapes: static-provider OAuth wrapping our own stdio Python modules (Zoom/Slack/Google-style) and API-key stdio apps (google-maps/AWS-style). A remote MCP server that only offers browser OAuth with Dynamic Client Registration fits neither — the DCR runtime already exists (`register_mcp_oauth_public_client`, `resolve_mcp_oauth_runtime_auth`) but was only reachable through the self-service "custom MCP server" path, which templates cannot reference (template `connections:` entries must resolve to catalog apps).

## Test plan
- [x] `pytest tests/web/test_classify_app_auth.py tests/web/api/test_public_mcp_connector_visibility.py tests/web/api/test_mcp_oauth_flow.py tests/web/api/test_mcp_apps_connect.py` — all passing
- [x] New tests cover: classification (positive shapes for all three remote transports, both partial shapes, stdio-transport misrouting guard), admin validator (partial-shape 422s, full-shape 200), and the new connect endpoint (server + association creation with DCR registration, idempotency across repeated connects, reconnect-after-disconnect reactivation, 400 on a non-mcp_oauth app, 404 on an unknown app id)
- [x] Frontend: `vitest` — connect-mcp-dialog (6, incl. new happy-path and error-path tests for the mcp_oauth branch) and custom-mcp-form (13, unchanged behavior after the helper move) all passing; `npm run type-check` clean

- [x] Manual end-to-end against a real DCR provider (Granola, via the follow-up connector branch): catalog Connect opens the centered OAuth popup, Dynamic Client Registration + Authorization Code/PKCE complete against mcp.granola.ai, the popup self-closes on success, the connector shows connected, and the server's remote tools (account info, meeting list/search/read) work from agent chat

### DIFF
--- a/frontend/src/app/tools/page.tsx
+++ b/frontend/src/app/tools/page.tsx
@@ -195,6 +195,22 @@ function ToolsPageContent() {
   }, [router, searchParams])
 
   useEffect(() => {
+    // The MCP OAuth callback redirects the connect popup here after a
+    // successful authorization. The popup was opened under this window name
+    // by the connect flows; self-close it so the opener's popup-closed poll
+    // refreshes connection state, instead of showing the whole app in the
+    // popup. On error, stay open so the toast above is readable.
+    if (
+      typeof window !== "undefined" &&
+      window.name === "mcp-oauth" &&
+      !searchParams.get("mcp_oauth_error") &&
+      !searchParams.get("mcp_oauth_error_message")
+    ) {
+      window.close()
+    }
+  }, [searchParams])
+
+  useEffect(() => {
     loadTools()
     loadMCPServers()
   }, [])
--- a/frontend/src/app/tools/page.tsx
+++ b/frontend/src/app/tools/page.tsx
@@ -57,6 +57,8 @@ import {
   mcpServerDetailToEditState,
   parseCustomApiDetail,
   parseMcpServerDetail,
+  shouldSelfCloseMcpOauthPopup,
+  MCP_OAUTH_SUCCESS_PARAM,
   type CustomApiDetail,
   type McpServerDetail,
 } from "@/lib/mcp-utils"
@@ -196,19 +198,24 @@ function ToolsPageContent() {
 
   useEffect(() => {
     // The MCP OAuth callback redirects the connect popup here after a
-    // successful authorization. The popup was opened under this window name
-    // by the connect flows; self-close it so the opener's popup-closed poll
-    // refreshes connection state, instead of showing the whole app in the
-    // popup. On error, stay open so the toast above is readable.
-    if (
-      typeof window !== "undefined" &&
-      window.name === "mcp-oauth" &&
-      !searchParams.get("mcp_oauth_error") &&
-      !searchParams.get("mcp_oauth_error_message")
-    ) {
+    // successful authorization, appending mcp_oauth_success=1. The popup was
+    // opened under this window name by the connect flows; self-close it so
+    // the opener's popup-closed poll refreshes connection state, instead of
+    // showing the whole app in the popup. On error the popup stays open so
+    // the toast above is readable. shouldSelfCloseMcpOauthPopup is keyed on
+    // the explicit success param, not the absence of error params, so a
+    // future error param added to the callback can't be mistaken for success.
+    if (!searchParams.get(MCP_OAUTH_SUCCESS_PARAM)) return
+    if (typeof window !== "undefined" && shouldSelfCloseMcpOauthPopup(window.name, searchParams)) {
       window.close()
+      return
     }
-  }, [searchParams])
+    // Not the popup (e.g. the flow ran in a full tab): just tidy the URL.
+    const nextParams = new URLSearchParams(searchParams.toString())
+    nextParams.delete(MCP_OAUTH_SUCCESS_PARAM)
+    const nextQuery = nextParams.toString()
+    router.replace(nextQuery ? `/tools?${nextQuery}` : "/tools", { scroll: false })
+  }, [router, searchParams])
 
   useEffect(() => {
     loadTools()
--- a/frontend/src/app/tools/page.tsx
+++ b/frontend/src/app/tools/page.tsx
@@ -184,16 +184,26 @@ function ToolsPageContent() {
   const { getAppIcon } = useMcpApps()
   const isAdmin = Boolean(user?.is_admin)
 
+  // F8: strip all three mcp_oauth redirect params in both effects below, not
+  // just the one each effect owns. The two backend redirects are mutually
+  // exclusive today (never both present at once), but if that ever changed,
+  // one effect's router.replace could otherwise undo the other's cleanup and
+  // cause a duplicate-toast loop.
+  const stripMcpOauthRedirectParams = (params: URLSearchParams) => {
+    const next = new URLSearchParams(params.toString())
+    next.delete("mcp_oauth_error")
+    next.delete("mcp_oauth_error_message")
+    next.delete(MCP_OAUTH_SUCCESS_PARAM)
+    const nextQuery = next.toString()
+    router.replace(nextQuery ? `/tools?${nextQuery}` : "/tools", { scroll: false })
+  }
+
   useEffect(() => {
     const oauthErrorMessage = searchParams.get("mcp_oauth_error_message")
     if (!oauthErrorMessage) return
 
     toast.error(oauthErrorMessage)
-    const nextParams = new URLSearchParams(searchParams.toString())
-    nextParams.delete("mcp_oauth_error")
-    nextParams.delete("mcp_oauth_error_message")
-    const nextQuery = nextParams.toString()
-    router.replace(nextQuery ? `/tools?${nextQuery}` : "/tools", { scroll: false })
+    stripMcpOauthRedirectParams(searchParams)
   }, [router, searchParams])
 
   useEffect(() => {
@@ -211,10 +221,7 @@ function ToolsPageContent() {
       return
     }
     // Not the popup (e.g. the flow ran in a full tab): just tidy the URL.
-    const nextParams = new URLSearchParams(searchParams.toString())
-    nextParams.delete(MCP_OAUTH_SUCCESS_PARAM)
-    const nextQuery = nextParams.toString()
-    router.replace(nextQuery ? `/tools?${nextQuery}` : "/tools", { scroll: false })
+    stripMcpOauthRedirectParams(searchParams)
   }, [router, searchParams])
 
   useEffect(() => {
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -132,6 +132,13 @@ export function ConnectMcpDialog({
   // leaking (N3: the poll otherwise has no unmount cleanup, unlike
   // custom-mcp-form's equivalent).
   const mcpOauthPollTimersRef = React.useRef<Set<number>>(new Set())
+  // The builtin_oauth flow's postMessage listener, tracked the same way so a
+  // dialog-close or unmount removes it too — otherwise a genuinely
+  // successful auth completed after the user closed the dialog would still
+  // fire onSuccess/auto-select via a listener nothing else was going to
+  // remove (its own interval, the only thing that used to remove it, is
+  // itself cleared by the same close/unmount cleanup).
+  const mcpOauthMessageListenersRef = React.useRef<Set<(event: MessageEvent) => void>>(new Set())
   // F5: this component instance is commonly kept mounted by the parent
   // across dialog open/close (only the Radix Dialog's own content
   // unmounts), so the unmount-only cleanup above never fires on an ordinary
@@ -231,6 +238,22 @@ export function ConnectMcpDialog({
     }
   }
 
+  // Shared by the dialog-close branch below and the unmount cleanup effect:
+  // stop any in-flight OAuth popup poll and remove the builtin_oauth flow's
+  // postMessage listener (F6) — leaving the listener attached would let a
+  // genuinely successful auth, completed after the user closed the dialog,
+  // still fire onSuccess/auto-select via a listener the (now-cleared)
+  // interval was the only thing that used to remove.
+  const clearMcpOauthPollState = () => {
+    mcpOauthPollTimersRef.current.forEach((timer) => window.clearInterval(timer))
+    mcpOauthPollTimersRef.current.clear()
+    mcpOauthMessageListenersRef.current.forEach((listener) =>
+      window.removeEventListener('message', listener)
+    )
+    mcpOauthMessageListenersRef.current.clear()
+    setLoadingApp(null)
+  }
+
   useEffect(() => {
     if (open) {
       setMcpFormData({
@@ -254,9 +277,7 @@ export function ConnectMcpDialog({
       // in-flight OAuth popup poll — otherwise it can later fire
       // onSuccess/auto-select against a closed dialog, and leaves a stale
       // spinner on the card if the dialog is reopened.
-      mcpOauthPollTimersRef.current.forEach((timer) => window.clearInterval(timer))
-      mcpOauthPollTimersRef.current.clear()
-      setLoadingApp(null)
+      clearMcpOauthPollState()
     }
   }, [open, t, selectedMcpServers])
 
@@ -272,8 +293,7 @@ export function ConnectMcpDialog({
   }, [])
 
   useEffect(() => () => {
-    mcpOauthPollTimersRef.current.forEach((timer) => window.clearInterval(timer))
-    mcpOauthPollTimersRef.current.clear()
+    clearMcpOauthPollState()
   }, [])
 
   useEffect(() => {
@@ -708,6 +728,7 @@ export function ConnectMcpDialog({
       if (event.data?.type === 'oauth-success') {
         setLoadingApp(null)
         window.removeEventListener('message', handleMessage)
+        mcpOauthMessageListenersRef.current.delete(handleMessage)
         window.clearInterval(checkPopup)
         mcpOauthPollTimersRef.current.delete(checkPopup)
 
@@ -724,6 +745,7 @@ export function ConnectMcpDialog({
     };
 
     window.addEventListener('message', handleMessage);
+    mcpOauthMessageListenersRef.current.add(handleMessage)
 
     // Fallback: check if popup was closed without success message, or give up
     // after a timeout (N3: this poll previously had neither an unmount-safe
@@ -737,6 +759,7 @@ export function ConnectMcpDialog({
       window.clearInterval(checkPopup);
       mcpOauthPollTimersRef.current.delete(checkPopup)
       window.removeEventListener('message', handleMessage);
+      mcpOauthMessageListenersRef.current.delete(handleMessage)
       setLoadingApp(null);
     }, 500);
     mcpOauthPollTimersRef.current.add(checkPopup)
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -527,7 +527,17 @@ export function ConnectMcpDialog({
     setLoadingApp(app.id)
     // Open the popup synchronously on the click, before any await — popup
     // blockers reject windows opened outside direct user-gesture handling.
-    const popup = window.open("about:blank", "_blank")
+    // The window-features string matters: without it, browsers open a full
+    // new tab instead of the small centered popup the builtin OAuth flow uses.
+    const width = 600
+    const height = 700
+    const left = window.screenX + (window.outerWidth - width) / 2
+    const top = window.screenY + (window.outerHeight - height) / 2
+    const popup = window.open(
+      "about:blank",
+      "mcp-oauth",
+      `width=${width},height=${height},left=${left},top=${top},scrollbars=yes`,
+    )
     if (!popup) {
       toast.error("Popup blocked. Please allow popups for this site to connect.")
       setLoadingApp(null)
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -43,6 +43,7 @@ import { sanitizeAppIntegrations } from "@/lib/team-sharing-sanitizers"
 import {
   isValidMcpName,
   parseMcpOAuthErrorMessage,
+  MCP_OAUTH_POPUP_WINDOW_NAME,
   type McpOAuthConnectResponse,
   buildCustomApiPayload,
   buildMcpServerPayload,
@@ -126,6 +127,10 @@ export function ConnectMcpDialog({
   const [customApiEditBaseline, setCustomApiEditBaseline] = useState<CustomApiDetail | null>(null)
   const [mcpEditBaseline, setMcpEditBaseline] = useState<McpServerDetail | null>(null)
   const connectorEditRequestRef = React.useRef(0)
+  // Popup-closed poll intervals for in-flight mcp_oauth connects, so they can
+  // be cleared on unmount instead of leaking (N3: the poll otherwise has no
+  // unmount cleanup, unlike custom-mcp-form's equivalent).
+  const mcpOauthPollTimersRef = React.useRef<Set<number>>(new Set())
 
   // Custom MCP Server state
   const [isSavingCustom, setIsSavingCustom] = useState(false)
@@ -242,6 +247,11 @@ export function ConnectMcpDialog({
 
   useEffect(() => () => {
     connectorEditRequestRef.current += 1
+  }, [])
+
+  useEffect(() => () => {
+    mcpOauthPollTimersRef.current.forEach((timer) => window.clearInterval(timer))
+    mcpOauthPollTimersRef.current.clear()
   }, [])
 
   useEffect(() => {
@@ -535,7 +545,7 @@ export function ConnectMcpDialog({
     const top = window.screenY + (window.outerHeight - height) / 2
     const popup = window.open(
       "about:blank",
-      "mcp-oauth",
+      MCP_OAUTH_POPUP_WINDOW_NAME,
       `width=${width},height=${height},left=${left},top=${top},scrollbars=yes`,
     )
     if (!popup) {
@@ -577,20 +587,49 @@ export function ConnectMcpDialog({
       return
     }
 
-    // The OAuth callback lands the popup on the app's own /tools page (no
-    // postMessage like the builtin flow) — so refresh once the popup closes.
-    const checkPopup = setInterval(() => {
-      if (popup.closed) {
-        clearInterval(checkPopup)
-        setLoadingApp(null)
+    // The popup's opener link is severed (popup.opener = null), so unlike the
+    // builtin flow there is no postMessage channel — and a closed popup can
+    // mean success OR a cancelled/denied/failed authorization (the error
+    // redirect leaves the popup open for the user to read, then they close it
+    // by hand). So on close, ask the backend which one actually happened and
+    // gate the success actions on the app really being connected; is_connected
+    // for mcp_oauth apps requires a completed grant, not just the association.
+    const startedAt = Date.now()
+    const maxWaitMs = 5 * 60 * 1000
+    const checkPopup = window.setInterval(() => {
+      const expired = Date.now() - startedAt >= maxWaitMs
+      if (!popup.closed && !expired) return
+      window.clearInterval(checkPopup)
+      mcpOauthPollTimersRef.current.delete(checkPopup)
+      setLoadingApp(null)
+      if (!popup.closed) {
+        // Timed out with the popup still open: stop the spinner. If the user
+        // eventually finishes, the next apps refresh shows the connection.
+        return
+      }
+      void (async () => {
+        let connected = false
+        try {
+          const response = await apiRequest(`${getApiUrl()}/api/mcp/apps?location=remote`)
+          if (response.ok) {
+            const data = sanitizeAppIntegrations(await response.json())
+            connected = data.some(
+              (candidate) => candidate.id === app.id && candidate.is_connected,
+            )
+          }
+        } catch (error) {
+          console.error("Failed to refresh apps after the OAuth popup closed:", error)
+        }
         loadApps()
+        if (!connected) return
         if (onSuccess) onSuccess()
         if (autoSelect && onConnectSelected) {
           setLocalSelectedServers(prev => prev.includes(app.name) ? prev : [...prev, app.name])
         }
         setSelectedApp(null)
-      }
+      })()
     }, 500)
+    mcpOauthPollTimersRef.current.add(checkPopup)
   }
 
   const handleConnectApp = (app: AppIntegration, autoSelect: boolean = false) => {
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -42,6 +42,8 @@ import { sanitizeAppIntegrations } from "@/lib/team-sharing-sanitizers"
 
 import {
   isValidMcpName,
+  parseMcpOAuthErrorMessage,
+  type McpOAuthConnectResponse,
   buildCustomApiPayload,
   buildMcpServerPayload,
   customApiDetailToEditState,
@@ -516,12 +518,79 @@ export function ConnectMcpDialog({
     }
   }
 
+  // Remote-MCP OAuth catalog app (e.g. Granola): the backend ensures the
+  // shared server row + this user's association, runs Dynamic Client
+  // Registration when the provider has no static client, and returns the
+  // authorization URL. Mirrors custom-mcp-form's handleConnectMcpOAuth,
+  // but keyed by catalog app_id instead of a server id.
+  const handleConnectMcpOAuthApp = async (app: AppIntegration, autoSelect: boolean) => {
+    setLoadingApp(app.id)
+    // Open the popup synchronously on the click, before any await — popup
+    // blockers reject windows opened outside direct user-gesture handling.
+    const popup = window.open("about:blank", "_blank")
+    if (!popup) {
+      toast.error("Popup blocked. Please allow popups for this site to connect.")
+      setLoadingApp(null)
+      return
+    }
+    popup.opener = null
+    try {
+      const response = await apiRequest(`${getApiUrl()}/api/mcp/apps/${app.id}/oauth/connect`, {
+        method: "POST",
+        credentials: "include",
+        headers: {
+          "Accept": "application/json",
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({ redirect_after: "/tools?tab=mcp" }),
+      })
+      if (!response.ok) {
+        popup.close()
+        const message = await parseMcpOAuthErrorMessage(response, t('tools.mcp.dialog.oauthConnectFailed'))
+        toast.error(message)
+        setLoadingApp(null)
+        return
+      }
+      const data = await response.json() as McpOAuthConnectResponse
+      if (!data.authorization_url) {
+        popup.close()
+        toast.error(t('tools.mcp.dialog.oauthConnectFailed'))
+        setLoadingApp(null)
+        return
+      }
+      popup.location.href = data.authorization_url
+    } catch (error) {
+      console.error("Failed to start MCP OAuth connect:", error)
+      popup.close()
+      toast.error(t('tools.mcp.dialog.oauthConnectFailed'))
+      setLoadingApp(null)
+      return
+    }
+
+    // The OAuth callback lands the popup on the app's own /tools page (no
+    // postMessage like the builtin flow) — so refresh once the popup closes.
+    const checkPopup = setInterval(() => {
+      if (popup.closed) {
+        clearInterval(checkPopup)
+        setLoadingApp(null)
+        loadApps()
+        if (onSuccess) onSuccess()
+        if (autoSelect && onConnectSelected) {
+          setLocalSelectedServers(prev => prev.includes(app.name) ? prev : [...prev, app.name])
+        }
+        setSelectedApp(null)
+      }
+    }, 500)
+  }
+
   const handleConnectApp = (app: AppIntegration, autoSelect: boolean = false) => {
     if (app.auth_type !== "builtin_oauth") {
-      // Key-based catalog app: collect the key. Anything else is a mis-authored
-      // entry (neither OAuth nor a launchable key-based command).
+      // Key-based catalog app: collect the key; remote-MCP OAuth app: start
+      // the per-user OAuth (DCR) flow. Anything else is a mis-authored entry.
       if (app.auth_type === "api_key") {
         openKeyConnect(app);
+      } else if (app.auth_type === "mcp_oauth") {
+        handleConnectMcpOAuthApp(app, autoSelect);
       } else {
         toast.error(t('tools.mcp.alerts.notConfigured'));
       }
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -127,10 +127,18 @@ export function ConnectMcpDialog({
   const [customApiEditBaseline, setCustomApiEditBaseline] = useState<CustomApiDetail | null>(null)
   const [mcpEditBaseline, setMcpEditBaseline] = useState<McpServerDetail | null>(null)
   const connectorEditRequestRef = React.useRef(0)
-  // Popup-closed poll intervals for in-flight mcp_oauth connects, so they can
-  // be cleared on unmount instead of leaking (N3: the poll otherwise has no
-  // unmount cleanup, unlike custom-mcp-form's equivalent).
+  // Popup-closed poll intervals for in-flight mcp_oauth/builtin_oauth
+  // connects, so they can be cleared on unmount or dialog-close instead of
+  // leaking (N3: the poll otherwise has no unmount cleanup, unlike
+  // custom-mcp-form's equivalent).
   const mcpOauthPollTimersRef = React.useRef<Set<number>>(new Set())
+  // F5: this component instance is commonly kept mounted by the parent
+  // across dialog open/close (only the Radix Dialog's own content
+  // unmounts), so the unmount-only cleanup above never fires on an ordinary
+  // close. Checked after every await in handleConnectMcpOAuthApp so an
+  // in-flight connect POST that resolves after the dialog (or the whole
+  // component) is gone doesn't register a poll nothing will ever clear.
+  const isMountedRef = React.useRef(true)
 
   // Custom MCP Server state
   const [isSavingCustom, setIsSavingCustom] = useState(false)
@@ -242,11 +250,25 @@ export function ConnectMcpDialog({
       setShareChoice("private")
     } else {
       connectorEditRequestRef.current += 1
+      // F6: closing the dialog (not just unmounting it) must also stop any
+      // in-flight OAuth popup poll — otherwise it can later fire
+      // onSuccess/auto-select against a closed dialog, and leaves a stale
+      // spinner on the card if the dialog is reopened.
+      mcpOauthPollTimersRef.current.forEach((timer) => window.clearInterval(timer))
+      mcpOauthPollTimersRef.current.clear()
+      setLoadingApp(null)
     }
   }, [open, t, selectedMcpServers])
 
   useEffect(() => () => {
     connectorEditRequestRef.current += 1
+  }, [])
+
+  useEffect(() => {
+    isMountedRef.current = true
+    return () => {
+      isMountedRef.current = false
+    }
   }, [])
 
   useEffect(() => () => {
@@ -586,6 +608,13 @@ export function ConnectMcpDialog({
       setLoadingApp(null)
       return
     }
+    // F5: if the dialog closed (or this component unmounted) while the
+    // connect POST was in flight, don't register a poll after the cleanup
+    // effects have already run — it would never get cleared. The popup
+    // itself is left navigating to the authorization URL either way; the
+    // user can still complete auth in it, and the next apps refresh (e.g.
+    // reopening the dialog) will reflect it.
+    if (!isMountedRef.current) return
 
     // The popup's opener link is severed (popup.opener = null), so unlike the
     // builtin flow there is no postMessage channel — and a closed popup can
@@ -679,6 +708,8 @@ export function ConnectMcpDialog({
       if (event.data?.type === 'oauth-success') {
         setLoadingApp(null)
         window.removeEventListener('message', handleMessage)
+        window.clearInterval(checkPopup)
+        mcpOauthPollTimersRef.current.delete(checkPopup)
 
         loadApps();
         if (onSuccess) onSuccess();
@@ -694,14 +725,21 @@ export function ConnectMcpDialog({
 
     window.addEventListener('message', handleMessage);
 
-    // Fallback: check if popup was closed without success message
-    const checkPopup = setInterval(() => {
-      if (popup?.closed) {
-        clearInterval(checkPopup);
-        window.removeEventListener('message', handleMessage);
-        setLoadingApp(null);
-      }
+    // Fallback: check if popup was closed without success message, or give up
+    // after a timeout (N3: this poll previously had neither an unmount-safe
+    // cleanup nor a timeout cap, unlike the mcp_oauth handler above and
+    // custom-mcp-form.tsx's equivalent flow).
+    const startedAt = Date.now()
+    const maxWaitMs = 5 * 60 * 1000
+    const checkPopup: number = window.setInterval(() => {
+      const expired = Date.now() - startedAt >= maxWaitMs
+      if (!popup?.closed && !expired) return
+      window.clearInterval(checkPopup);
+      mcpOauthPollTimersRef.current.delete(checkPopup)
+      window.removeEventListener('message', handleMessage);
+      setLoadingApp(null);
     }, 500);
+    mcpOauthPollTimersRef.current.add(checkPopup)
   }
 
   const handleCardClick = (app: AppIntegration, isGloballyConnected: boolean) => {
--- a/frontend/src/lib/mcp-utils.ts
+++ b/frontend/src/lib/mcp-utils.ts
@@ -540,3 +540,23 @@ export function buildMcpServerPayload(
   })
   return payload
 }
+
+// --- MCP OAuth (remote server) connect helpers -----------------------------
+
+export interface McpOAuthConnectResponse {
+  authorization_url: string
+}
+
+/** Extract a human-readable message from an MCP OAuth endpoint error body.
+ * Shared by the custom-server form and the catalog connect dialog. */
+export async function parseMcpOAuthErrorMessage(response: Response, fallback: string): Promise<string> {
+  try {
+    const payload = await response.json()
+    if (typeof payload?.detail === "string") return payload.detail
+    if (payload?.detail?.message) return payload.detail.message
+    if (payload?.detail?.code) return payload.detail.code
+  } catch {
+    // Keep the provided fallback.
+  }
+  return fallback
+}
--- a/frontend/src/lib/mcp-utils.ts
+++ b/frontend/src/lib/mcp-utils.ts
@@ -560,3 +560,27 @@ export async function parseMcpOAuthErrorMessage(response: Response, fallback: st
   }
   return fallback
 }
+
+/** The window name the mcp_oauth connect popup opens under (connect-mcp-dialog
+ * and custom-mcp-form both use this), and the query param the callback
+ * appends on success. Shared so the popup-open side and the tools page's
+ * self-close guard can't drift apart on either literal. */
+export const MCP_OAUTH_POPUP_WINDOW_NAME = "mcp-oauth"
+export const MCP_OAUTH_SUCCESS_PARAM = "mcp_oauth_success"
+
+/** Whether the tools page, loaded with these search params under this window
+ * name, is the mcp_oauth connect popup landing on a successful callback and
+ * should self-close. Keyed on the explicit success param (not on the absence
+ * of error params) so a future error param added to the callback redirect
+ * can't be mistaken for success here — see the PR review that flagged this
+ * coupling (N5). Extracted as a pure function so it's unit-testable without
+ * mounting the full tools page. */
+export function shouldSelfCloseMcpOauthPopup(
+  windowName: string,
+  searchParams: URLSearchParams,
+): boolean {
+  return (
+    searchParams.get(MCP_OAUTH_SUCCESS_PARAM) !== null &&
+    windowName === MCP_OAUTH_POPUP_WINDOW_NAME
+  )
+}
--- a/frontend/src/lib/mcp-utils.ts
+++ b/frontend/src/lib/mcp-utils.ts
@@ -561,10 +561,15 @@ export async function parseMcpOAuthErrorMessage(response: Response, fallback: st
   return fallback
 }
 
-/** The window name the mcp_oauth connect popup opens under (connect-mcp-dialog
- * and custom-mcp-form both use this), and the query param the callback
- * appends on success. Shared so the popup-open side and the tools page's
- * self-close guard can't drift apart on either literal. */
+/** The window name the mcp_oauth catalog connect popup opens under
+ * (connect-mcp-dialog.tsx), and the query param the callback appends on
+ * success. Shared so that popup-open side and the tools page's self-close
+ * guard can't drift apart on either literal.
+ *
+ * Not (yet) used by custom-mcp-form.tsx's own OAuth flow: that popup still
+ * opens under the reserved target "_blank", so window.name is empty there
+ * and this self-close mechanism never fires for it — a pre-existing gap,
+ * not something this constant's naming implies is already covered. */
 export const MCP_OAUTH_POPUP_WINDOW_NAME = "mcp-oauth"
 export const MCP_OAUTH_SUCCESS_PARAM = "mcp_oauth_success"
 
--- a/src/xagent/web/api/admin_mcp.py
+++ b/src/xagent/web/api/admin_mcp.py
@@ -222,9 +222,20 @@ def _public_mcp_app_response(app: PublicMCPApp) -> Dict[str, Any]:
     }
 
 
-def _validate_public_mcp_app_values(values: Dict[str, Any]) -> None:
+def _validate_public_mcp_app_values(
+    values: Dict[str, Any], *, enforce_connect_shape: bool = True
+) -> None:
+    # F4: PublicMCPAppCreate's _enforce_auth_classification rejects a
+    # connect-shape that classifies as "unconnectable" (see that validator's
+    # docstring). Applying it unconditionally on every edit would let a shape
+    # rule added after a row was created (or a row otherwise grandfathered
+    # in) permanently block that row's unrelated fields (e.g. icon) from ever
+    # being edited again. Skip that one check — via PublicMCPAppBase, which
+    # PublicMCPAppCreate extends with no other override — when this edit
+    # doesn't touch the fields the shape check reads (transport/launch_config).
+    model = PublicMCPAppCreate if enforce_connect_shape else PublicMCPAppBase
     try:
-        PublicMCPAppCreate.model_validate(values)
+        model.model_validate(values)
     except ValidationError:
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
@@ -235,6 +246,7 @@ def _validate_public_mcp_app_values(values: Dict[str, Any]) -> None:
 def _apply_public_mcp_app_update(db_app: PublicMCPApp, changes: Dict[str, Any]) -> None:
     canonical = get_builtin_public_mcp_app(db_app.app_id)
     persisted = _public_mcp_app_values(db_app)
+    enforce_connect_shape = bool({"transport", "launch_config"} & changes.keys())
 
     if canonical is not None:
         for field in _BUILTIN_PROTECTED_FIELDS.intersection(changes):
@@ -253,7 +265,9 @@ def _apply_public_mcp_app_update(db_app: PublicMCPApp, changes: Dict[str, Any]) 
             **writable_changes,
             **{field: canonical[field] for field in _BUILTIN_PROTECTED_FIELDS},
         }
-        _validate_public_mcp_app_values(merged)
+        _validate_public_mcp_app_values(
+            merged, enforce_connect_shape=enforce_connect_shape
+        )
     else:
         if "app_id" in changes and changes["app_id"] != db_app.app_id:
             raise HTTPException(
@@ -261,7 +275,9 @@ def _apply_public_mcp_app_update(db_app: PublicMCPApp, changes: Dict[str, Any]) 
                 detail="MCP app ID is immutable",
             )
         merged = {**persisted, **changes}
-        _validate_public_mcp_app_values(merged)
+        _validate_public_mcp_app_values(
+            merged, enforce_connect_shape=enforce_connect_shape
+        )
         writable_changes = {
             field: value for field, value in changes.items() if field != "app_id"
         }
--- a/src/xagent/web/api/admin_mcp.py
+++ b/src/xagent/web/api/admin_mcp.py
@@ -97,19 +97,28 @@ class PublicMCPAppCreate(PublicMCPAppBase):
     def _enforce_auth_classification(self) -> "PublicMCPAppCreate":
         # Reuse the single source of truth (classify_app_auth) rather than
         # re-deriving the rule here. Reject an entry that declares a partial
-        # launch_config (command or required_env) yet still classifies as
-        # "unconnectable" — the write-time constraint issue #764 asked for. This
-        # covers both asymmetric shapes: command-without-required_env and
-        # required_env-without-command.
+        # launch_config (either the key-based or the remote-OAuth shape) yet
+        # still classifies as "unconnectable" — the write-time constraint
+        # issue #764 asked for. Covers four asymmetric shapes:
+        # command-without-required_env, required_env-without-command,
+        # url-without-auth.type=mcp_oauth, and auth.type=mcp_oauth-without-url.
         from ..mcp_apps import classify_app_auth
 
         launch = self.launch_config or {}
-        if (launch.get("command") or launch.get("required_env")) and (
+        declares_key_shape = bool(launch.get("command") or launch.get("required_env"))
+        auth = launch.get("auth")
+        declares_remote_oauth_shape = bool(launch.get("url")) or (
+            isinstance(auth, dict) and auth.get("type") == "mcp_oauth"
+        )
+        if (declares_key_shape or declares_remote_oauth_shape) and (
             classify_app_auth(self.transport, self.launch_config) == "unconnectable"
         ):
             raise ValueError(
                 "A key-based catalog app must declare both launch_config.command "
-                "and launch_config.required_env, otherwise it cannot be connected."
+                "and launch_config.required_env; a remote-OAuth catalog app must "
+                "declare both launch_config.url and launch_config.auth.type == "
+                "'mcp_oauth' (with transport one of sse/websocket/streamable_http), "
+                "otherwise it cannot be connected."
             )
         return self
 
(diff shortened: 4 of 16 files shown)


src/xagent/web/api/mcp.py
```diff
@@ -2302,6 +2302,97 @@ def _ensure_catalog_app_server(db: Session, app_id: str) -> tuple[MCPServer, dic
...
+                UserMCPServer.mcpserver_id == server.id,
+                UserMCPServer.is_owner.is_(True),
+            )
+            .first()
+        )
+        if owned is not None:
+            raise HTTPException(
+                status_code=status.HTTP_409_CONFLICT,
+                detail="A user-owned server already exists under this catalog id",
+            )
```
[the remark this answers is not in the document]
yiboyasss: Partially adopted in a6f7d27e. The owned-row and transport/url hijack guards described in the first half were already present in this function (the suggested block keeps them unchanged). The genuinely new point — syncing the reused row's auth config when the catalog entry's auth changes — is valid and now implemented, with one correction to the suggested code: `server.auth != auth` compares the stored (encrypted-at-rest) value against the catalog's plaintext, which would spuriously differ on every connect once any sensitive field (e.g. `client_secret`) is configured, rewriting the row each time. The implemented sync compares on `server._decrypt_auth_config(server.auth)` instead, and re-encrypts sensitive fields before storing. Two regression tests added: one verifying a registry auth change propagates on the next connect, and one pinning the decrypted-comparison behavior (unchanged Fernet ciphertext across two connects proves no spurious rewrite).

src/xagent/web/api/mcp.py
```diff
@@ -2436,6 +2543,72 @@ def _apply_updates(a: Any) -> None:
...
+    assoc: Any = (
+        db.query(UserMCPServer)
+        .filter(
+            UserMCPServer.user_id == current_user.id,
+            UserMCPServer.mcpserver_id == server.id,
+        )
+        .first()
+    )
+    if assoc is None:
+        assoc = UserMCPServer(
```
rogercloud, line 2621: **[Major] M1: an abandoned/failed OAuth authorization leaves the app permanently "Connected" with no working credential.**

This `UserMCPServer` association is created with `is_active=True` and committed here — before the user ever reaches the consent screen (the DCR/PKCE redirect happens later, in `connect_mcp_oauth`). Meanwhile `is_connected` for `mcp_oauth` apps (see `list_mcp_apps`, ~line 1879, via `_build_active_non_oauth_server_lookup`) is derived purely from this active association — it never checks `MCPOAuthGrant`. Contrast `builtin_oauth`, which correctly requires a real completed OAuth account via `_connected_oauth_server_for_app`.

Concretely: a user clicks Connect, sees the consent screen, and closes the popup without finishing (or denies consent, or the callback errors). The card renders "Connected" — and in select mode may be auto-added to an agent's selected servers — with zero credential behind it. Every subsequent tool call fails at runtime, and there's currently no re-authorize affordance to recover (see D2 in the review body).

Suggested fix: don't create/activate this association until the OAuth callback actually succeeds (mirroring how `api_key` only associates after the key validates), or alternatively make `mcp_oauth`'s `is_connected` also require a valid, non-revoked `MCPOAuthGrant` for that user+server.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -516,12 +518,89 @@ export function ConnectMcpDialog({
...
+      console.error("Failed to start MCP OAuth connect:", error)
+      popup.close()
+      toast.error(t('tools.mcp.dialog.oauthConnectFailed'))
+      setLoadingApp(null)
+      return
+    }
+
+    // The OAuth callback lands the popup on the app's own /tools page (no
+    // postMessage like the builtin flow) — so refresh once the popup closes.
+    const checkPopup = setInterval(() => {
```
rogercloud: **[Major] M2 (compounds M1): popup-closed is treated as unconditional success — [Minor] N3: no unmount cleanup or timeout cap.**

M2: this `setInterval` checks only `popup.closed`; when true it unconditionally calls `loadApps()`, `onSuccess?.()`, and pushes the app into the auto-selected list, without reading any signal about *why* the popup closed. On an OAuth failure, the backend redirects the popup to `/tools?mcp_oauth_error=...`, and `tools/page.tsx`'s self-close effect deliberately leaves the popup open in that case so the user can read the error toast — but once the user manually closes that error popup, this poll fires the identical success path anyway. The sibling `builtin_oauth` handler in this same file (~line 659) avoids this correctly via a `postMessage({type: 'oauth-success'})` listener, using `popup.closed` only as a loading-state fallback. Suggest the backend's success redirect also signal success explicitly (postMessage or a distinguishing query param) rather than conflating "closed" with "succeeded".

N3: this loop also has no `useEffect`-based cleanup tied to unmount and no elapsed-time cap, unlike `custom-mcp-form.tsx`'s equivalent (`clearOAuthPolling` + 5-minute `maxWaitMs`), which this code's own comment says it mirrors. Impact is bounded (a stuck spinner on one card until navigation away from `/tools`), and the sibling `builtin_oauth` handler here has the same gap already — but worth aligning while touching this code.

src/xagent/web/mcp_apps.py
```diff
@@ -78,20 +78,39 @@ def restrict_to_app_scoped_oauth_grant(
...
+        - "unconnectable": none of the above
     """
     if str(transport or "").lower() == "oauth":
         return "builtin_oauth"
     launch = launch_config if isinstance(launch_config, dict) else {}
     if launch.get("required_env") and launch.get("command"):
         return "api_key"
+    auth = launch.get("auth")
+    if (
+        str(transport or "") in _REMOTE_MCP_TRANSPORTS
```
rogercloud: **[Minor] N1: transport-string case handling is inconsistent across this feature.**

This remote-transport membership check is case-sensitive, but the `builtin_oauth` check two lines above (`str(transport or "").lower() == "oauth"`) lowercases. The other half of the same feature disagrees too: `_ensure_catalog_mcp_oauth_server`'s hijack-guard comparison (`mcp.py:2338`-ish) *does* lowercase both sides, and `_get_mcp_oauth_config` (`mcp.py:586`) is case-sensitive like this one. A stored `"Streamable_HTTP"` would classify as `unconnectable` here while matching case-insensitively in the mcp.py hijack guard.

No shipped registry entry hits this today, but `PATCH /admin/mcp/apps/{app_id}` (via `PublicMCPAppUpdate`, which has no classification validator) can write a mixed-case `transport` onto an existing row with no guard at all — at which point this classifier and the hijack-guard genuinely disagree about the same stored row. Suggest normalizing transport casing once at write time, or `.lower()` consistently at every comparison site.

src/xagent/web/api/mcp.py
```diff
@@ -2302,6 +2302,113 @@ def _ensure_catalog_app_server(db: Session, app_id: str) -> tuple[MCPServer, dic
...
+) -> tuple[MCPServer, dict]:
+    """Idempotently ensure the shared server row for a remote-MCP OAuth
+    (DCR-capable) catalog app exists, without creating any per-user
+    association. Returns (server, app_info). Mirrors
+    _ensure_catalog_app_server's hijack guards, but for a streamable_http/
+    sse/websocket server row instead of a stdio one.
+    """
+    from ..mcp_apps import get_app_by_id
+
+    app_info = get_app_by_id(db, app_id)
```
rogercloud, line 2389: **[Minor] N2 (pre-existing, replicated — not blocking): this endpoint doesn't check `is_visible_in_connector` before provisioning.**

`get_app_by_id` does a raw id lookup with no visibility filter; `is_visible_in_connector` is enforced only inside `list_mcp_apps`'s listing loop (~line 1909). This means any user who knows a hidden connector's `app_id` can still provision the shared server row via this endpoint. Confirmed this identical gap already exists at the base commit for the pre-existing stdio `/apps/{app_id}/connect` endpoint, so it isn't introduced by this PR — it's just now replicated into a second path that additionally mints DCR client registrations, modestly widening blast radius. Worth a follow-up covering both endpoints; not a reason to hold this PR.

src/xagent/web/api/admin_mcp.py
```diff
@@ -97,19 +97,28 @@ class PublicMCPAppCreate(PublicMCPAppBase):
...
+        # issue #764 asked for. Covers four asymmetric shapes:
+        # command-without-required_env, required_env-without-command,
+        # url-without-auth.type=mcp_oauth, and auth.type=mcp_oauth-without-url.
         from ..mcp_apps import classify_app_auth
 
         launch = self.launch_config or {}
-        if (launch.get("command") or launch.get("required_env")) and (
+        declares_key_shape = bool(launch.get("command") or launch.get("required_env"))
+        auth = launch.get("auth")
+        declares_remote_oauth_shape = bool(launch.get("url")) or (
```
rogercloud, line 110: **[Minor] N4 (follow-up, zero current blast radius — not blocking): `bool(launch.get("url"))` is broader than the shape this guard is meant to enforce.**

Any `launch_config` carrying a `url` key that isn't a complete `mcp_oauth` shape gets rejected here — including a hypothetical legitimately no-auth or statically-authenticated remote catalog entry. That said, `classify_app_auth` has no representation for such a shape at all (remote transports classify only as `mcp_oauth` or `unconnectable`), and that limitation predates and is independent of this validator — so this is arguably surfacing a pre-existing scope limit rather than adding a new bug. No registry entry today carries a `url` key, so nothing currently writable is affected; applies on PUT (full replacement) but not PATCH. Worth revisiting if/when a no-auth or static-auth remote catalog shape is wanted.

frontend/src/app/tools/page.tsx
```diff
@@ -194,6 +194,22 @@ function ToolsPageContent() {
...
 
+  useEffect(() => {
+    // The MCP OAuth callback redirects the connect popup here after a
+    // successful authorization. The popup was opened under this window name
+    // by the connect flows; self-close it so the opener's popup-closed poll
+    // refreshes connection state, instead of showing the whole app in the
+    // popup. On error, stay open so the toast above is readable.
+    if (
+      typeof window !== "undefined" &&
+      window.name === "mcp-oauth" &&
```
rogercloud: **[Minor] N5: hardcoded coupling between this self-close guard and the error-redirect param names, two files apart.**

This closes any window named `"mcp-oauth"` unless exactly `mcp_oauth_error` / `mcp_oauth_error_message` are present — and those two param names are produced far away in `mcp.py`'s `_mcp_oauth_callback_error_redirect` (~line 455). Confirmed the window name `"mcp-oauth"` is used nowhere else today, so there's no collision currently. But a future third error-signaling param added on the backend without a matching update here would make the popup silently self-close and swallow the error. A shared constant, or a positive "success" marker instead of an error denylist, would remove the coupling.

src/xagent/web/api/mcp.py
```diff
@@ -2302,6 +2302,113 @@ def _ensure_catalog_app_server(db: Session, app_id: str) -> tuple[MCPServer, dic
...
+    if server:
+        if (
+            str(server.transport or "").lower() != transport.lower()
+            or server.url != url
+        ):
+            raise HTTPException(
+                status_code=status.HTTP_409_CONFLICT,
+                detail="A server with this name already exists with a different configuration",
+            )
+        owned = (
```
rogercloud: **[Design note] D1: this `is_owner` squat guard, and the add-server/race-recovery tail further below, are byte-for-byte duplicates of `_ensure_catalog_app_server`'s equivalents.**

A security-relevant guard living in two copies risks a future hardening fix landing in only one of them. Suggest extracting just these two specific blocks into small shared helpers (e.g. `_reject_user_owned_squat(db, server)` and `_add_server_with_race_recovery(manager, db, config, server_name)`). To be clear about scope: the rest of this function (404 check, auth-type gate, the hijack/config-mismatch comparison just above) genuinely compares different fields per shape (`command`/`args`/`stdio` vs `url`/`transport`/`auth`) and shouldn't be forced into one generic comparator — only these two verbatim blocks are worth extracting.

src/xagent/web/mcp_apps.py
```diff
@@ -78,20 +78,39 @@ def restrict_to_app_scoped_oauth_grant(
     ]
 
 
+# Transports for a genuine remote MCP server the platform talks to directly
+# (as opposed to "oauth", which here means a static-provider redirect wrapping
+# our own stdio module — see classify_app_auth). Mirrors HTTP_MCP_TRANSPORTS
+# in services/mcp_runtime.py and the runtime check in
+# core/tools/core/mcp/manager/db.py's _requires_runtime_mcp_oauth.
+_REMOTE_MCP_TRANSPORTS = frozenset({"sse", "websocket", "streamable_http"})
```
rogercloud: **[Simplification] stdlib:** `_REMOTE_MCP_TRANSPORTS` duplicates `HTTP_MCP_TRANSPORTS` already exported from `src/xagent/web/services/mcp_runtime.py:8` (byte-identical set; no import-cycle risk — checked). Import and reuse the existing constant instead of maintaining two copies that have to be kept in sync by hand.

rogercloud (commented):
## Review: `feat: support remote MCP servers with Dynamic Client Registration in the built-in catalog`

### Summary

The built-in MCP catalog previously understood only two connection shapes: `builtin_oauth` (a static-provider redirect wrapping one of our own stdio modules) and `api_key` (stdio + command + required env). Remote MCP servers that offer only browser OAuth with Dynamic Client Registration (RFC 7591 — no static client credentials to pre-provision) fit neither. This PR adds a third shape, `mcp_oauth`: `classify_app_auth()` recognizes remote transport + `url` + `auth.type == "mcp_oauth"`, the admin validator rejects partial declarations of that shape, and a new `POST /api/mcp/apps/{app_id}/oauth/connect` endpoint provisions a shared `MCPServer` row plus a per-user `UserMCPServer` association before delegating to the existing `connect_mcp_oauth` DCR/PKCE machinery. The frontend connector dialog gains a third connect branch that drives a popup through the returned `authorization_url`. No registry entry ships in this PR — this is platform capability only, with the first real connector to follow.

Note on prior reviews: the earlier Gemini Code Assist suggestion (sync a reused catalog server row's auth config from the registry when it drifts) was already implemented within this PR's own history by commit `a6f7d27e`, so it is resolved and not carried forward here.

### Approach verdict

**Acceptable with reservations** — the reuse instinct is right (no new tables, no duplicated PKCE/DCR/token code, one classification source of truth), but wrapping a *shared* catalog server row around machinery designed for *per-user* server rows leaks in three places this PR does not close: connection-state truthfulness (M1/M2) and credential lifecycle on disconnect (M3).

**D1 — two verbatim blocks should be extracted from `_ensure_catalog_mcp_oauth_server`.** The new function duplicates, byte-for-byte, the `is_owner` catalog-squat guard (~13 lines) and the add-server / `IntegrityError` race-recovery / error-mapping tail (~20 lines) from `_ensure_catalog_app_server`. A *security-relevant* guard living in two copies is the specific risk here: a future hardening fix will very plausibly land in only one of them. Two small helpers — roughly `_reject_user_owned_squat(db, server)` and `_add_server_with_race_recovery(manager, db, config, server_name)` — would close that. To be explicit about what I am *not* asking for: the rest of the two functions (404 check, auth-type gate, hijack/config-mismatch comparison) genuinely compares different fields per shape (`command`/`args`/`stdio` vs `url`/`transport`/`auth`), and folding those into one generic field-comparator would be over-abstraction. Extract the two identical blocks; leave the shape-specific logic separate.

**D2 — there is no manage / revoke / re-authorize surface for a connected `mcp_oauth` catalog app.** `official-mcp-settings-dialog.tsx` is untouched and has no `mcp_oauth` branch (`isKeyBased` only tests `api_key`), so a connected app of the new shape gets no grants list, no per-grant revoke, and no explicit re-authorize — only full Disconnect. That is a regression in capability relative to `custom-mcp-form.tsx`'s own `mcp_oauth` flow, which has all three. More sharply: when `OfficialMcpSettingsDialog` is mounted from `tools/page.tsx` rather than from `connect-mcp-dialog.tsx`, no `onConnectStart` prop is wired, so its internal fallback `handleConnectApp` would build a `builtin_oauth`-style `/api/auth/{provider}/login` URL for an `mcp_oauth` app — the wrong endpoint entirely, defaulting `provider` to `"linkedin"` when unset. This is unreachable today (no live `mcp_oauth` catalog entry exists), so it is not a blocker on *this* PR, but it will actively break the moment the follow-up connector ships and should be a required prerequisite for that PR.

One pre-existing limitation the new shape inherits rather than introduces: if a registry entry's `url` or `transport` ever changes, `_ensure_catalog_mcp_oauth_server` returns a permanent 409 with no in-code migration path — exactly mirroring the existing `_ensure_catalog_app_server` behavior for `command`/`args`, which has historically been resolved by direct DB migration. Faithful mirroring, not a new defect.

---

## Findings

### Major

See inline comments M1/M2 below for file:line anchors (M3 targets `delete_mcp_server`, which this PR does not touch, so GitHub won't anchor an inline comment there — noting it here instead). Summary:

- **M1** — An abandoned or failed OAuth authorization leaves the app permanently displayed as "Connected" with no working credential. The `UserMCPServer` association is committed before the user reaches the consent screen, and `is_connected` for `mcp_oauth` apps never checks `MCPOAuthGrant` (unlike `builtin_oauth`).
- **M2** (compounds M1) — The frontend's popup-closed poll treats *any* popup closure as success (`loadApps()`, `onSuccess()`, auto-select all fire unconditionally), with no signal distinguishing a completed connection from a cancelled/failed one — unlike the sibling `builtin_oauth` handler in the same file, which uses `postMessage`.
- **M3** — Disconnecting a shared `mcp_oauth` catalog server never revokes the disconnecting user's `MCPOAuthGrant` (refresh token); it only cascades away once the *last* associated user disconnects. The correct revoke logic (`delete_mcp_oauth_grant`) exists but isn't wired into `delete_mcp_server`.

**M1 and M2 compound each other** — either alone produces a wrong "Connected" state; together, the backend records a phantom connection *and* the frontend affirmatively reports success for it, so there is no layer where the failure is caught.

### Minor

See inline comments N1/N2/N4/N5 below for file:line anchors. Summary:

- **N1** — Transport-string case handling is inconsistent: `classify_app_auth`'s remote-transport check is case-sensitive while its `builtin_oauth` check and `_ensure_catalog_mcp_oauth_server`'s hijack-guard are both case-insensitive. `PATCH /admin/mcp/apps/{app_id}` (via `PublicMCPAppUpdate`) can write a mixed-case `transport` with no validator at all, creating a live disagreement between the two halves of the feature on the same stored row.
- **N2** *(pre-existing, replicated — not blocking)* — The new `/oauth/connect` endpoint doesn't check `is_visible_in_connector` before provisioning, same as the existing pre-PR stdio `/apps/{id}/connect` endpoint. Not introduced here, but now replicated into a path that also mints DCR client registrations.
- **N3** — The popup-closed poll (same code as M2) has no unmount cleanup and no timeout cap, unlike `custom-mcp-form.tsx`'s equivalent flow which has both. Impact is bounded (stuck spinner until navigation), and the sibling `builtin_oauth` handler in the same file has the identical gap already.
- **N4** *(follow-up, zero current blast radius — not blocking)* — The admin validator's `bool(launch.get("url"))` gate rejects any `url`-bearing launch_config that isn't a full `mcp_oauth` shape, including a hypothetical legitimately no-auth remote entry — but `classify_app_auth` has no representation for that shape at all today, so this is arguably surfacing a pre-existing scope limit rather than a new bug. No registry entry currently carries `url`.
- **N5** — The popup self-close guard in `tools/page.tsx` is hardcoded to two specific error param names set far away in `mcp.py`'s `_mcp_oauth_callback_error_redirect`. Not exploitable today (no window-name collision exists), but a future third error param added without updating this guard would cause the popup to silently self-close and swallow the error.

---

## Test coverage assessment

**Genuinely well covered:** first-connect provisioning including DCR and association fields; idempotency across repeated connects; reconnect-after-disconnect re-activation; registry auth-change propagation — including the encrypted-comparison subtlety (`test_connect_app_does_not_rewrite_unchanged_auth_with_secret` correctly asserts ciphertext stability); 400 for a non-`mcp_oauth` app; 404 for an unknown id. `classify_app_auth` coverage is solid across all three remote transports, both partial shapes, wrong `auth.type`, and the stdio-misrouting guard. The admin validator is covered for both asymmetric partial shapes plus the accepting case.

**Gaps worth naming:**
- The hand-copied name-hijack guards (is_owner squat, url/transport mismatch) are **entirely untested** — combined with D1 (duplicated blocks), untested duplicated security guards is the combination I'd most want closed.
- The `Accept: application/json` branch — the one the actual frontend uses — is never exercised by a backend test; every backend test defaults `accept=None` and asserts the 303 redirect branch instead.
- No test asserts that a cancelled/failed authorization leaves the app **not** connected — there is no test asserting `is_connected` for an `mcp_oauth` app at all, which is precisely why M1 shipped.
- No test for the association-insert `IntegrityError` recovery path, nor the server-create race recovery.
- No test for `is_visible_in_connector=False` against the new endpoint (N2).
- No precedence test for an entry declaring both `api_key` and `mcp_oauth` shapes simultaneously.
- No test for mixed-case transports (N1).
- Frontend: no test for the popup-closed poll's success path, the popup-blocked branch, or the `tools/page.tsx` self-close effect in its entirety.

**Limitation of this review:** no Python interpreter was available in the review environment, so the test suite was **not executed**. All coverage claims above are from reading the test files against the diff, not from observed results — treat "covered" as *coverage exists*, not *coverage passes*.

---

- `src/xagent/web/mcp_apps.py:86` — `stdlib:` `_REMOTE_MCP_TRANSPORTS` duplicates `HTTP_MCP_TRANSPORTS` already exported from `src/xagent/web/services/mcp_runtime.py:8` (byte-identical set; no import-cycle risk). Import and reuse the existing constant instead.

`net: -3 lines` — the value isn't the line count, it's removing a second copy of a set that has to stay in sync with the runtime's notion of which transports are remote.

I also looked at extracting a shared popup/poll helper between `connect-mcp-dialog.tsx` and `custom-mcp-form.tsx` and concluded against it: the two polls are different algorithms (a plain `popup.closed` check vs. a status-polling loop with timeout and mount guards), and the genuinely shared popup-open boilerplate is only ~8-10 lines — not worth an abstraction.

src/xagent/web/api/mcp.py
yiboyasss: Acknowledged as pre-existing/replicated, not blocking, per your own note — no change made here to keep this PR's diff scoped to the mcp_oauth shape itself. Worth a follow-up covering both the stdio and mcp_oauth connect endpoints together if you'd like it tracked separately.

src/xagent/web/api/admin_mcp.py
yiboyasss: Acknowledged as a pre-existing scope limit (classify_app_auth has no representation for a no-auth/static-auth remote shape today), not blocking per your note, and no registry entry currently carries a bare url without full mcp_oauth auth — left as-is. Happy to revisit if/when that shape is actually wanted.

src/xagent/web/api/mcp.py
```diff
@@ -2436,6 +2558,72 @@ def _apply_updates(a: Any) -> None:
...
+    assoc: Any = (
+        db.query(UserMCPServer)
+        .filter(
+            UserMCPServer.user_id == current_user.id,
+            UserMCPServer.mcpserver_id == server.id,
+        )
+        .first()
+    )
+    if assoc is None:
+        assoc = UserMCPServer(
```
rogercloud, line 2621: **[M1 — PARTIAL] The root cause of M1 is still here: this association is created and committed before the user ever reaches OAuth consent.**

The display-layer symptom is now fixed (`is_connected` correctly nulls out for a grant-less `mcp_oauth` app in the default view, see the gate added around line 1914). But this line itself — the actual premature commit — is unchanged: nothing deactivates or deletes this `UserMCPServer` row if the user abandons/denies consent. That's why the same underlying bug resurfaces through side doors this round: the `location=local|all` branch of `list_mcp_apps` doesn't consult the new grant gate at all (see the comment on line ~1914), and once `is_connected` correctly goes false, there's no UI path left to clean up this now-dormant row (see the comment on line ~1919) — it just keeps getting loaded into every agent run as an `authorization_required` unavailable tool.

Suggest closing this at the root: don't create/commit this association until the OAuth callback actually succeeds (create it alongside the grant), rather than gating visibility after the fact.

src/xagent/web/api/mcp.py
```diff
@@ -2874,6 +3062,27 @@ def delete_mcp_server(
...
 
+        # Revoke this user's MCP OAuth grants for the server. On a shared
+        # (multi-user) row the server outlives this disconnect, so without
+        # this the grant's refresh token would stay usable until the LAST
+        # user disconnects and the cascade finally removes it. Local
+        # revocation is sufficient for the runtime (it only resolves
+        # status == "active" grants); best-effort revocation at the external
+        # provider stays with the per-grant DELETE endpoint, which this sync
+        # handler cannot await.
+        revoked_at = _utc_now()
```
rogercloud: **[M3 — PARTIAL] This revocation is local-only — the external provider is never told.**

This block correctly flips the disconnecting user's active grant(s) to `status="revoked"`, scoped correctly to `(user_id, mcp_server_id)` — verified another user's grant on the same shared server is untouched. That's sufficient to stop this platform from reusing the token (every runtime resolution path filters `status=="active"`). But `_revoke_mcp_oauth_grant_externally` (the function `delete_mcp_oauth_grant` uses, which actually POSTs to the provider's `revocation_endpoint`) is never called here — because this function is a synchronous `def` and can't await it. The encrypted access/refresh tokens remain stored and stay live at the provider until they expire naturally.

Also worth closing alongside this: when the shared row is *kept* (team-owned or platform-key rows, a few lines below), the now-revoked grant, its `MCPOAuthClient`, and any `MCPOAuthFlowState` rows are never purged — there's no sweep job for any of them, so they accumulate indefinitely as inert but secret-bearing rows.

Suggest making this handler (or a background task it enqueues) able to await the real external revoke, and purging the now-revoked local rows in the kept-row branches too.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -516,12 +528,118 @@ export function ConnectMcpDialog({
...
+        if (autoSelect && onConnectSelected) {
+          setLocalSelectedServers(prev => prev.includes(app.name) ? prev : [...prev, app.name])
+        }
+        setSelectedApp(null)
+      })()
+    }, 500)
+    mcpOauthPollTimersRef.current.add(checkPopup)
+  }
+
   const handleConnectApp = (app: AppIntegration, autoSelect: boolean = false) => {
```
rogercloud, line 684: **[N3 — PARTIAL] This sibling `builtin_oauth` handler still has the exact gap N3 flagged, unlike the `mcp_oauth` handler above which is now fixed.**

`handleConnectMcpOAuthApp` above (line 536) now has a mount-tracking ref, an unmount cleanup effect, and a 5-minute timeout cap. This handler's own `checkPopup` poll (~line 698) still has neither: no unmount-safe cleanup, no timeout cap. N3 explicitly named this sibling as sharing the same weakness — worth aligning it with the new pattern while this file is already being touched.

src/xagent/web/api/mcp.py
```diff
@@ -1890,6 +1911,12 @@ def list_mcp_apps(
                 server_id = _connected_non_oauth_server_for_app(
                     app, non_oauth_server_lookup
                 )
+                if (
```
rogercloud: **[F1 — Major] This grant-based gate only applies to the default/`remote` location branch — the `location in ("local", "all")` branch further down (search `if location in ["local", "all"]:`) computes its own connected/server_id state via a separate string comparison (`app["name"].lower()` vs `server.name.lower()`, where `server.name` holds the raw `app_id`) and never consults this gate at all.**

Concretely: if an admin-authored catalog app's `app_id` differs from its display `name` — an entirely ordinary pattern, e.g. `app_id="google-ads"` with `name="Google Ads"` — that branch's match misses, and it reports the exact abandoned-auth "Connected" state this gate was added to eliminate, just reached via the dialog's user-switchable Local/All tab instead of the default tab. No live catalog entry triggers this today, but it only takes a normal admin authoring action to trigger, not an edge case.

Suggest deriving connected state for `mcp_oauth` apps from one shared helper both branches call, keyed on `app_id` rather than display name.

rogercloud (commented):
## Re-review: update summary

Since the last review the head moved from `06c9e88a` to `62dceffe`, adding three genuine fix commits: `99d7a767` (extract the shared catalog-server guards, add a positive OAuth success signal — targets D1 and N5), `f88d61ec` (require an active grant for `mcp_oauth` `is_connected`, revoke the grant on disconnect — targets M1 and M3), and `62dceffe` (distinguish a cancelled OAuth popup from a completed one — targets M2 and N3). The author also replied inline on N2 and N4, confirming both were intentionally left as-is to keep the diff scoped; both were explicitly non-blocking and I agree with leaving them.

Overall: the direction of these three commits is right and non-superficial — M2 in particular ended up more robust than what was originally suggested. Two of the three Major findings are, however, only partially closed, and a fresh independent pass over the current code found the same root causes resurfacing through paths that weren't in scope for the specific fixes.

---

## Status of prior findings

| Finding | Status | Notes |
| --- | --- | --- |
| **M1** — abandoned OAuth leaves app permanently "Connected" | **PARTIAL** | The reported mechanism is genuinely fixed: `list_mcp_apps` (`src/xagent/web/api/mcp.py:1884-1919`) now pre-loads the user's active `MCPOAuthGrant` server-ids in a single query and nulls `server_id` (hence `is_connected`) for an `mcp_oauth` app without one. Traced end to end: abandon consent → no grant → `is_connected=False`; well tested. But the root cause is untouched — the `UserMCPServer` association is still created and committed (`mcp.py:2588`) before the user ever reaches consent, and nothing deactivates or deletes it on an abandoned flow. That dormant row resurfaces as F1 and F3 below. |
| **M2** — popup closure treated as unconditional success | **FIXED** | Solved more robustly than suggested: instead of a postMessage signal (which can race or be missed), `popup.closed` is now only a *trigger* to re-query the authoritative `GET /api/mcp/apps?location=remote`, and `onSuccess`/auto-select only fire `if (connected)` (`frontend/src/components/mcp/connect-mcp-dialog.tsx:536-633`). Traced the abandon-then-close-error-popup scenario: correctly not marked connected. The sibling `builtin_oauth` handler is untouched, so no regression there. Tested in both directions. |
| **M3** — disconnect never revoked the user's OAuth refresh token | **PARTIAL** | `delete_mcp_server` (`mcp.py:3065-3086`) now flips the disconnecting user's active grant(s) for that server to `status="revoked"` on every disconnect, correctly scoped to `(user_id, mcp_server_id)` — verified a second user's grant on the same shared server is untouched, and the missing-grant edge case is handled gracefully. This *is* sufficient to stop in-platform reuse (every runtime token-resolution path filters `status == "active"`). But it is a local status flip only: `_revoke_mcp_oauth_grant_externally` is never called, because `delete_mcp_server` is still a synchronous `def` and cannot await it. The encrypted access/refresh tokens stay in the DB and stay live at the provider until natural expiry — the code's own comment acknowledges this. |
| **N1** — transport-string case sensitivity | **FIXED** | Both the `classify_app_auth` remote-transport check and the hijack guard now lowercase consistently, and the PATCH path now runs through the same case-insensitive classification validator. Storage is still not normalized (only comparisons are), which matches the alternative fix the finding itself offered — fine. |
| **N2** — new endpoint doesn't check `is_visible_in_connector` | **CONFIRMED UNCHANGED (accepted)** | Exactly as the author's inline reply stated. Explicitly flagged as pre-existing and non-blocking; no action needed in this PR. |
| **N3** — popup poll lacked unmount cleanup / timeout cap | **PARTIAL** | The `mcp_oauth` flow — the finding's main subject — is fully fixed: mount-tracking ref plus cleanup effect (`connect-mcp-dialog.tsx:252-255`) and a 5-minute `maxWaitMs` cap. The finding also explicitly named the sibling `builtin_oauth` handler in the same file (`handleConnectApp`, ~635-705) as having the identical gap, and that handler is untouched: still no unmount-safe cleanup and no timeout cap. |
| **N4** — admin `bool(url)` validator gate broader than intended | **CONFIRMED UNCHANGED (accepted)** | Exactly as the author's inline reply stated. Pre-existing and explicitly non-blocking; no action needed. (See F4 below for a *new* edge this PR's own validator change introduces — that one is separate and does warrant a look.) |
| **D1** — duplicated blocks between the two catalog-provisioning functions | **FIXED** | Both blocks are now shared helpers — `_reject_user_owned_catalog_squat` and `_add_catalog_server_with_race_recovery` (`mcp.py:2227-2285`) — called from both `_ensure_catalog_app_server` and `_ensure_catalog_mcp_oauth_server`. Behavior verified identical against the pre-refactor commit, and the genuinely shape-specific logic was correctly left separate rather than force-unified. Tests for both paths pass. |
| **D2** — no manage/revoke/re-authorize UI for a connected `mcp_oauth` app | **NOT ADDRESSED (expected)** | `official-mcp-settings-dialog.tsx` is untouched by this diff, which is correct: D2 was scoped as a prerequisite for the later connector PR, not a blocker on this platform-capability PR. Carrying the reminder forward. Note F3 below makes this gap sharper than before. |
| **Simplification** — duplicate `frozenset` constant | **FIXED** | `mcp_apps.py` now imports and reuses `HTTP_MCP_TRANSPORTS` from `src/xagent/web/services/mcp_runtime.py`; the duplicate `_REMOTE_MCP_TRANSPORTS` definition is gone. |

---

## New findings — Major

See inline comments F1 and F3 below.

- **F1** — The new active-grant gate only covers the default/`remote` location branch of `list_mcp_apps`; the `location in ("local", "all")` branch computes connection state via a separate, ungated string comparison on display name, so an admin-authored app whose `app_id` differs from its display `name` (an ordinary authoring pattern) would still show the exact abandoned-auth "Connected" state, just via the dialog's Local/All tab.
- **F3** — The M1 fix removed the only reachable cleanup path for a grant-less dormant connection. The Disconnect UI is now gated behind the same `is_connected` flag M1 correctly nulls, so a user who abandoned consent can rediscover the dormant row only via the generic Tools server list, where the only actions are Reconnect or a moot per-grant delete — there is no delete-server action reachable anywhere. Meanwhile the dormant row keeps being loaded into every agent run as an `authorization_required` unavailable tool.

---

## New findings — Minor

See inline comments F4, F7/F9, F5, F6, F8, F10, F11, F13 below. F4 and F7 were independently re-verified; F5/F6/F8/F9/F10/F11/F13 come from direct code reading with concrete anchors but weren't independently re-verified a second time — treat those as "worth a look" rather than fully confirmed.

---

## Design-level observation (the main takeaway)

M1 and M3 share one pattern, worth naming explicitly: both fixes correct the primary-path *symptom* (`is_connected` display for M1, local grant status for M3) while leaving the *root cause* in place (the premature `UserMCPServer` commit before consent; no external token revocation). This round's independent pass found that root cause resurfacing through whichever path wasn't in the specific fix's scope: F1 (the alternate `local`/`all` location branch recomputes connection state without the gate), F3 (the cleanup path for the dormant row is now unreachable precisely *because* `is_connected` is correctly false), and F10 (orphaned secret-bearing rows in the kept-row disconnect branches).

For state-truthfulness bugs like these, a display-layer or single-path fix tends to reproduce the same bug through the next code path. The two highest-leverage remaining changes: (a) don't commit the association until consent completes (or reap grant-less associations), which collapses F1 and F3 and finishes M1; and (b) make the disconnect path able to await the external revoke and purge the local grant, which collapses F10 and finishes M3.

Secondary, no action needed: the two `_ensure_catalog_*_server` functions still exist as two functions plus two shared helpers, rather than one function parameterized by shape — D1's actual ask (extract the two verbatim blocks) was met; the remaining skeleton duplication was never part of that ask.

---

## Tests actually run this pass

Unlike the initial review pass (no interpreter available then), both suites were executed against the current head:

- `uv run pytest tests/web/api/test_mcp_oauth_flow.py tests/web/test_classify_app_auth.py tests/web/api/test_public_mcp_connector_visibility.py -q` → **126 passed**
- `npx vitest run src/components/mcp/connect-mcp-dialog.test.tsx src/lib/mcp-utils.test.ts` → **39 passed**

New tests are well-targeted: idempotency, reactivation, auth-sync ciphertext stability, the M1 grant gate, the M3 sibling-grant isolation, `classify_app_auth` case-insensitivity, both asymmetric admin shapes, and both the cancelled and timeout popup paths. Also clean this round: delegation to the existing `connect_mcp_oauth` machinery (no route collisions), race handling in `_add_catalog_server_with_race_recovery`, and the hijack guards (now tested for the `mcp_oauth` path too).

src/xagent/web/api/mcp.py
```diff
@@ -1890,6 +1911,12 @@ def list_mcp_apps(
...
+                    app.get("auth_type") == "mcp_oauth"
+                    and server_id is not None
+                    and server_id not in active_grant_server_ids
+                ):
+                    server_id = None
```
rogercloud: **[F3 — Major] Nulling `server_id` here (correctly) also removes the only reachable path to clean up this now-dormant association.**

Once this fires, the connector card shows a plain "Connect" button, and the Disconnect/settings UI in `official-mcp-settings-dialog.tsx` is entirely gated behind the same `is_connected` flag this line now correctly sets to false — so it's unreachable there. The dormant `UserMCPServer` row is still discoverable via the generic Tools page's server list (`GET /api/mcp/servers`, which has no grant filter), but that editor's only actions are Reconnect (retry OAuth) and a per-grant delete that's moot at zero grants — there is no delete-server action reachable for this row anywhere in the UI. Meanwhile it keeps being loaded into every agent run as an `authorization_required` unavailable tool entry.

Suggest either not creating the association until consent succeeds (the root fix, see the comment on the `assoc = UserMCPServer(...)` call above), or reaping/offering cleanup for grant-less associations directly from this same query path.

src/xagent/web/api/admin_mcp.py
rogercloud, line 110: **[F4 — Minor] This PR's own validator tightening has an edit-path regression: a pre-existing admin-authored row can become permanently uneditable via PUT.**

Before this PR, an admin could freely author (via the admin form's free-form `launch_config` JSON textarea) a catalog row with a `url` key and non-`mcp_oauth` auth (or no auth) — it classified as `unconnectable` but wasn't rejected. This line's `declares_remote_oauth_shape` check now correctly rejects that shape on POST. But since the same validator also backs PUT (full replacement), and `_apply_public_mcp_app_update` re-validates the whole merged row unconditionally, any such pre-existing row becomes permanently uneditable — even for an unrelated field change like the icon. No built-in catalog entry is affected (none use `launch_config.url`), so this is limited to admin-authored custom rows, but those are realistically authorable today.

Consider only enforcing this shape rule when the request actually touches `launch_config`/`auth`, or grandfathering rows that already violate it.

frontend/src/lib/mcp-utils.ts
```diff
@@ -540,3 +540,47 @@ export function buildMcpServerPayload(
...
+    // Keep the provided fallback.
+  }
+  return fallback
+}
+
+/** The window name the mcp_oauth connect popup opens under (connect-mcp-dialog
+ * and custom-mcp-form both use this), and the query param the callback
+ * appends on success. Shared so the popup-open side and the tools page's
+ * self-close guard can't drift apart on either literal. */
+export const MCP_OAUTH_POPUP_WINDOW_NAME = "mcp-oauth"
```
rogercloud, line 573: **[F7/F9 — Minor] This constant's doc comment is inaccurate, and its reuse mechanism has a durability caveat plus a newly-introduced leak.**

The doc comment above claims both `connect-mcp-dialog.tsx` and `custom-mcp-form.tsx` use this constant — `custom-mcp-form.tsx` (line 260) still opens its popup via the reserved target `"_blank"`, not this name, so `window.name` is empty there and this self-close mechanism can never fire for that flow (pre-existing behavior; the comment is new and now misleading).

Genuinely new regression: `ConnectMcpDialog` (and thus `custom-mcp-form`'s OAuth flow) is also mounted from `frontend/src/components/build/agent-builder.tsx:3009`, a `/build…` route. That flow's `redirect_after` is `window.location.pathname + search` — whatever page currently hosts it — so completing a custom MCP server's OAuth from inside the agent builder now redirects back to that `/build…` URL with a stray `?mcp_oauth_success=1` appended, and nothing in the build route reads or strips it (zero references there). Not a crash, but a permanently visible junk query param on the builder URL after an ordinary user action.

Separately (F9): this whole mechanism depends on `window.name` surviving a cross-origin round trip to the OAuth provider and back — reliable in Chromium/Firefox, but not guaranteed on WebKit's cross-site state clearing, nor for a popup-blocker fallback or a `noopener`-opened window. On those paths the popup won't self-close, reproducing the exact "renders the whole app in the popup" state this change was meant to remove. Also, since a *failed* popup is deliberately left open under this reused name, a later unrelated success elsewhere could `window.close()` a window the user has since started using normally.

Suggest: fix `custom-mcp-form.tsx:260` to use `MCP_OAUTH_POPUP_WINDOW_NAME` (or correct the doc comment), and strip `mcp_oauth_success` in the build route too (or make the redirect-after target success-param-agnostic).

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -516,12 +528,118 @@ export function ConnectMcpDialog({
     }
   }
 
+  // Remote-MCP OAuth catalog app (e.g. Granola): the backend ensures the
+  // shared server row + this user's association, runs Dynamic Client
+  // Registration when the provider has no static client, and returns the
+  // authorization URL. Mirrors custom-mcp-form's handleConnectMcpOAuth,
+  // but keyed by catalog app_id instead of a server id.
+  const handleConnectMcpOAuthApp = async (app: AppIntegration, autoSelect: boolean) => {
```
rogercloud, line 578: **[F5 — Minor] No `isMounted` guard after this handler's initial `await`, unlike the sibling `custom-mcp-form.tsx` flow.**

`custom-mcp-form.tsx` checks `isMountedRef.current` after every `await` in its equivalent flow. Here, if the dialog unmounts while the initial connect POST is in flight, the poll interval created afterward gets registered into the cleanup set (line ~252) after that cleanup effect has already run on unmount — so it's never drained, and later fires state updates (`setLoadingApp`, `loadApps`, etc.) against an unmounted tree.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -242,6 +249,11 @@ export function ConnectMcpDialog({
     connectorEditRequestRef.current += 1
   }, [])
 
+  useEffect(() => () => {
```
rogercloud, line 295: **[F6 — Minor] This cleanup only runs on unmount — closing (not unmounting) the dialog mid-authorization doesn't clear the poll or reset `loadingApp`.**

A user who closes the dialog while a connect is in flight can have the poll later fire `onSuccess`/auto-select against a closed dialog, and reopening the dialog resets other state in a way that silently discards the "connect and auto-select" intent while leaving a stale spinner on that row.

frontend/src/app/tools/page.tsx
```diff
@@ -194,6 +196,27 @@ function ToolsPageContent() {
...
+  useEffect(() => {
+    // The MCP OAuth callback redirects the connect popup here after a
+    // successful authorization, appending mcp_oauth_success=1. The popup was
+    // opened under this window name by the connect flows; self-close it so
+    // the opener's popup-closed poll refreshes connection state, instead of
+    // showing the whole app in the popup. On error the popup stays open so
+    // the toast above is readable. shouldSelfCloseMcpOauthPopup is keyed on
+    // the explicit success param, not the absence of error params, so a
+    // future error param added to the callback can't be mistaken for success.
+    if (!searchParams.get(MCP_OAUTH_SUCCESS_PARAM)) return
```
rogercloud, line 218: **[F8 — Minor] This effect's cleanup only strips its own success param, and the sibling error-toast effect above only strips the error params — if both were ever present simultaneously, one `router.replace` could undo the other's cleanup and cause a duplicate-toast loop.**

Not reachable today (the two backend redirects are mutually exclusive), but cheap to harden by stripping all three params (`mcp_oauth_error`, `mcp_oauth_error_message`, `mcp_oauth_success`) in both effects rather than just the one each currently owns.

src/xagent/web/api/mcp.py
```diff
@@ -2874,6 +3062,27 @@ def delete_mcp_server(
...
+                MCPOAuthGrant.mcp_server_id == server_id,
+                MCPOAuthGrant.user_id == user_id,
+                MCPOAuthGrant.status == "active",
+            )
+            .all()
+        ):
+            setattr(grant, "status", "revoked")
+            setattr(grant, "revoked_at", revoked_at)
+
         # Remove user-server association
```
rogercloud, line 3131: **[F10 — Minor, relates to M3] When the shared row is kept alive here (team-owned or platform-key rows, just below), the grant/client/flow-state rows for the disconnecting user are never purged.**

The grant was just marked `revoked` a few lines up, but its row (with both encrypted tokens still physically stored), any associated `MCPOAuthClient`, and any `MCPOAuthFlowState` rows have no sweep/cleanup job anywhere in the codebase. Inert at runtime, but an unbounded accumulation of secret-bearing rows over time. Doing the external revoke (M3's remaining half) plus purging the local rows here would close both at once.

src/xagent/web/mcp_apps.py
```diff
@@ -78,20 +78,39 @@ def restrict_to_app_scoped_oauth_grant(
     ]
 
 
+# Transports for a genuine remote MCP server the platform talks to directly
```
rogercloud: **[F11 — Minor, code quality] Dangling comment: this describes the `_REMOTE_MCP_TRANSPORTS` constant that no longer exists here — it was replaced by an in-function import of `HTTP_MCP_TRANSPORTS`.**

Misleading to the next reader; one-line move or delete.

src/xagent/web/api/mcp.py
```diff
@@ -2276,29 +2344,83 @@ def _ensure_catalog_app_server(db: Session, app_id: str) -> tuple[MCPServer, dic
...
+        # Compare on the decrypted form — sensitive auth fields are encrypted
+        # at rest, so comparing the raw stored value against the catalog's
+        # plaintext would spuriously differ on every connect once any secret
+        # is configured. Runs only after the owned-check above: a user-owned
+        # row is rejected, never mutated.
+        if server._decrypt_auth_config(server.auth) != auth:
+            encrypted_auth = dict(auth)
+            for key in SENSITIVE_AUTH_FIELDS:
+                if key in encrypted_auth and encrypted_auth[key]:
+                    encrypted_auth[key] = encrypt_value(encrypted_auth[key])
```
rogercloud: **[F13 — Minor, edge case] This calls `encrypt_value()` on whatever the catalog stored for a sensitive auth field with no type check first.**

A mis-authored non-string value for a sensitive field (e.g. a nested object where a string is expected) would raise inside a user-facing connect request instead of being caught/rejected at admin write time.

src/xagent/web/api/mcp.py
```diff
@@ -1876,6 +1900,22 @@ def list_mcp_apps(
...
+    # screen, so an abandoned/denied/failed authorization would otherwise
+    # render as "Connected" forever with no credential behind it. Require an
+    # active grant too, mirroring how builtin_oauth requires a completed
+    # OAuth account. One query for all apps to avoid an N+1.
+    active_grant_server_ids = {
+        row[0]
+        for row in db.query(MCPOAuthGrant.mcp_server_id)
+        .filter(
+            MCPOAuthGrant.user_id == current_user.id,
+            MCPOAuthGrant.status == "active",
```
rogercloud, line 1914: **[F3 — PARTIAL / NEW-1 scenario A] This active-grant gate is correctly the single source of truth for `is_connected` here, but nothing downstream consumes it to offer a cleanup action for the apps it now (correctly) reports as disconnected.**

`frontend/src/components/mcp/official-mcp-settings-dialog.tsx` (untouched by this PR's diff) still gates its Disconnect/"Delete service" button behind `isGloballyConnected = Boolean(app.is_connected)` — the very flag this gate now correctly nulls for a grant-less `mcp_oauth` app. So a user who abandoned consent sees only "Connect" again in that dialog, with no way to remove the dormant association from that surface. A workaround exists via the separate, ungated `GET /api/mcp/servers` Tools-page listing, but that's a different screen than the one showing the problem.

See the companion comment below (`mcp.py:2024`) for a sharper, newly-introduced version of this same gap.

src/xagent/web/api/mcp.py
```diff
@@ -1969,7 +2017,13 @@ def list_mcp_apps(
                     "icon": "",
                     "users": "1",
                     "transport": server.transport,
-                    "is_connected": True,
+                    # F1: this loop's own membership check above (name-based)
+                    # doesn't gate on a real grant, so a custom mcp_oauth
+                    # server the user abandoned mid-consent must not be
+                    # reported connected just because the row exists.
+                    "is_connected": _mcp_oauth_server_is_actually_connected(
```
rogercloud, line 2024: **[NEW-1 scenario B — Major] Correctly gating `is_connected` here turns a user's own custom mcp_oauth server into a complete dead end.**

This line previously hardcoded `"is_connected": True` unconditionally (confirmed via diff against base) — now it's correctly grant-gated. But the dict built in this loop has no `auth_type` field at all. For a user-created custom `streamable_http` server with `auth.type=mcp_oauth` and no grant, that means: (1) `is_connected` is false, so the Delete button in `official-mcp-settings-dialog.tsx` disappears (same gating as the comment above); and (2) clicking "Connect" reaches `handleConnectApp`'s type switch in `frontend/src/components/mcp/connect-mcp-dialog.tsx` (~line 684) with `app.auth_type === undefined`, which falls through to the generic `else` branch that just toasts `notConfigured` and does nothing.

Net effect: this card cannot be connected (does nothing useful) and cannot be deleted — a genuine dead end for the user's own, previously-deletable server. Suggest including `auth_type` in this branch's dict, and/or keying the Delete affordance on "an association row exists" rather than on `is_connected`.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -516,12 +570,125 @@ export function ConnectMcpDialog({
...
+        if (autoSelect && onConnectSelected) {
+          setLocalSelectedServers(prev => prev.includes(app.name) ? prev : [...prev, app.name])
+        }
+        setSelectedApp(null)
+      })()
+    }, 500)
+    mcpOauthPollTimersRef.current.add(checkPopup)
+  }
+
   const handleConnectApp = (app: AppIntegration, autoSelect: boolean = false) => {
```
rogercloud, line 684: **[NEW-1 scenario B, companion] This type switch has no fallback for a genuine `mcp_oauth` server whose `auth_type` field is missing from the payload (see `mcp.py:2024`) — it silently toasts `notConfigured` instead of routing to the `mcp_oauth` connect flow.**

Worth either back-filling `auth_type` at the source, or having this switch treat a `transport` in the remote-OAuth set as `mcp_oauth` even when `auth_type` is absent, so a custom server in this state isn't a dead end.

src/xagent/web/api/mcp.py
```diff
@@ -3405,8 +3659,17 @@ async def mcp_oauth_callback(
             message="MCP OAuth authorization failed",
         )
 
+    # Positive success signal: the connect popup's self-close logic keys on
+    # this param instead of inferring success from "no error params", so a
+    # future error param added to the error redirect can't be mistaken for
+    # success by an out-of-date guard.
     response = RedirectResponse(
```
rogercloud, line 3666: **[NEW-2 / F9 — Minor] This success param is appended unconditionally to every completed MCP OAuth flow, including the custom-server flow that can originate from `/build/<agent-id>` (`custom-mcp-form.tsx`'s `redirect_after = window.location.pathname + search`), and nothing on that route strips or reads it.**

Confirmed newly introduced by this PR (the base redirect appended no params at all) and confirmed as the custom-server OAuth popup already never self-closes regardless of this PR (a separate, pre-existing gap — it opens under `"_blank"`, not the shared window name), so the practical impact is limited to a stray, permanent `?mcp_oauth_success=1` on an already-stuck popup's URL rather than a functional break. Still worth a follow-up: either scope this param's addition to the catalog-dialog flow specifically, or add stripping logic to the `/build` route too.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -516,12 +570,125 @@ export function ConnectMcpDialog({
...
+    setLoadingApp(app.id)
+    // Open the popup synchronously on the click, before any await — popup
+    // blockers reject windows opened outside direct user-gesture handling.
+    // The window-features string matters: without it, browsers open a full
+    // new tab instead of the small centered popup the builtin OAuth flow uses.
+    const width = 600
+    const height = 700
+    const left = window.screenX + (window.outerWidth - width) / 2
+    const top = window.screenY + (window.outerHeight - height) / 2
+    const popup = window.open(
```
rogercloud, line 588: **[NEW-3 — Minor] This popup always opens under the same fixed window name, with no guard against a second concurrent connect.**

`loadingApp` only disables the initiating app's own card, and `OfficialMcpSettingsDialog` has no connecting-lock, so a user can start app B's connect while app A's popup is still open. Standard `window.open` name-reuse then navigates app A's popup to app B's auth flow. App A's poll correctly avoids reporting a false success (it re-queries the backend, which correctly shows no grant), but there's no error message telling the user that app A's authorization attempt was silently killed. Suggest a per-app-unique window name, or a lock that requires finishing (or cancelling) an in-progress connect before starting another.

src/xagent/web/api/mcp.py
```diff
@@ -2276,29 +2371,89 @@ def _ensure_catalog_app_server(db: Session, app_id: str) -> tuple[MCPServer, dic
...
+        if server._decrypt_auth_config(server.auth) != auth:
+            encrypted_auth = dict(auth)
+            for key in SENSITIVE_AUTH_FIELDS:
+                value = encrypted_auth.get(key)
+                # A mis-authored non-string sensitive field (e.g. a nested
+                # object where a string is expected) must not crash this
+                # user-facing connect request — encrypt_value() calls
+                # .encode() unconditionally. Leave it as-is; that's an admin
+                # authoring bug to catch at write time, not here (F13).
+                if value and isinstance(value, str):
```
rogercloud, line 2437: **[F13 residual — Minor] This guard is correct here for the auth-sync/reuse path, but the parallel first-provision path (`MCPServer.from_config`, `src/xagent/core/tools/core/mcp/model.py`, not touched by this PR) has no equivalent guard.**

That path calls `encrypt_value` on a sensitive field with only a truthiness check — a non-string value there reaches `cipher.encrypt(value.encode())` and raises an uncaught `AttributeError`, which `_add_catalog_server_with_race_recovery`'s `except (ValueError, IntegrityError)` does not catch. So a mis-authored non-string sensitive field in a catalog entry's `launch_config` still crashes the *first* user's connect attempt as an unhandled 500 instead of a clean 400. Same one-line `isinstance(value, str)` guard would close it there too.

rogercloud (commented):
## Third Review Round — Follow-up

The six new commits (`d8bac44f` → `38f937fd`) targeted the items left open after the second round: gating `mcp_oauth` connection state across the remaining listing branch (M1/F1), external token revocation on disconnect (M3), the admin validator PUT/PATCH regression (F4), and the frontend popup-poll lifecycle issues (F5/F6/F7/F8). Most of those are now genuinely fixed, and the M3 external-revoke work in particular is a proper root-cause fix rather than a symptom patch. What remains open is the same theme as round 2 — the fix is applied where the state is *computed*, but not everywhere that state is *consumed* — and this round that gap produced one confirmed new regression (NEW-1 scenario B).

### Status of items open after the second review

| Item | Status | Notes |
|---|---|---|
| **M1** — abandoned OAuth shows "Connected" (default/remote branch of `list_mcp_apps`) | **FIXED** | `src/xagent/web/api/mcp.py:1909-1917` prefetches active grants, and `:1936-1943` nulls `server_id` for a grant-less `mcp_oauth` app. Covered by `test_mcp_oauth_app_not_connected_until_grant_completes`. |
| **F1** — `location in ("local","all")` branch bypassed the gate | **FIXED within `list_mcp_apps`** | `src/xagent/web/api/mcp.py:2024-2026` replaces the previously hardcoded `"is_connected": True` with the shared helper `_mcp_oauth_server_is_actually_connected(...)`; the diff against base confirms this is a real behavior change. New test `tests/web/api/test_mcp_oauth_flow.py:2252`. **However**, the gate is still local to `list_mcp_apps` — `GET /api/mcp/servers` (the generic Tools-page server list) has no grant gate at all, so a dormant server still shows as present/usable there. |
| **F3** — no way to clean up a dormant grant-less association | **PARTIAL** | The Disconnect/Delete button in `official-mcp-settings-dialog.tsx` is still gated behind `isGloballyConnected` (`Boolean(app.is_connected)`) — the very flag the F1 fix now correctly nulls — so the dialog offers no cleanup action. This round it also produces a confirmed new regression; see **NEW-1** below. |
| **M3** — disconnect never revoked the refresh token externally | **FIXED** (two new caveats) | `delete_mcp_server` (`src/xagent/web/api/mcp.py:2987`) is now `async def` and at `:3116-3119` genuinely `await`s `_revoke_mcp_oauth_grant_externally(...)` per active grant, placed *before* all the "keep the shared row" early returns. `test_delete_mcp_server_revokes_external_tokens_when_endpoint_is_advertised` mocks the provider via `httpx.MockTransport` and asserts the real POST. Two caveats written up as Minor findings below (Caveat 1 subsumes the former F10). |
| **F4** — admin validator PUT regression | **FIXED** | `src/xagent/web/api/admin_mcp.py:248-249` now picks the backing model based on whether the request actually touches `{"transport","launch_config"}`. PUT still enforces unconditionally, but PATCH only enforces when the shape-relevant fields change, so a pre-existing non-conforming row can still have its icon/description edited. Test: `test_admin_update_grandfathers_a_preexisting_bad_shape_row_for_unrelated_edits`. |
| **F5** — missing `isMounted` guard | **FIXED** | `connect-mcp-dialog.tsx:148` ref, `:288-292` effect, and a guard at `:637` before the poll interval is registered. |
| **F6** — dialog close doesn't clear poll/listeners | **FIXED**, and extended | New `clearMcpOauthPollState` helper (`connect-mcp-dialog.tsx:247-257`) clears every tracked interval *and* removes every tracked message listener; called from both the dialog-close branch and the unmount effect. The sibling `builtin_oauth` listener is now tracked and removed the same way. Tests at `connect-mcp-dialog.test.tsx:645,705`. |
| **F7** — misleading doc comment | **FIXED** | `frontend/src/lib/mcp-utils.ts:565-572` now states explicitly that `custom-mcp-form.tsx`'s flow is *not* covered by the shared window-name constant, instead of claiming both flows use it. |
| **F8** — duplicate-toast edge case | **FIXED** | A single `stripMcpOauthRedirectParams` (`frontend/src/app/tools/page.tsx:190-198`) strips all three params and is called from both effects, so neither can undo the other's cleanup. |
| **F9** — stray query param leaks onto non-`/tools` routes | **NOT FIXED** | The F8 fix only unified stripping *within* the tools page. `custom-mcp-form.tsx` (reachable from `/build/<agent-id>` via `agent-builder.tsx`) still sends `redirect_after = window.location.pathname + search`, and the backend's success redirect — confirmed new in this PR, the base appended no params at all — now appends `mcp_oauth_success=1` to every successful completion, including that one. `agent-builder.tsx` has zero references to any `mcp_oauth_*` param. Impact assessment in **NEW-2** below. |
| **F11** — dangling comment referring to removed code | **FIXED** | No residual comment remains in the touched hunks; confirmed by grep plus clean `ruff`/`mypy`. |
| **F13** — auth-sync crash on malformed sensitive field | **PARTIAL** | The sync/reuse path (`src/xagent/web/api/mcp.py:2434-2438`) now guards with `if value and isinstance(value, str)` before `encrypt_value`. The first-provision path has the same gap — see the Minor finding below. |

*(F10 is no longer listed separately: it is now the same underlying issue as M3's Caveat 1.)*

---

### New findings this round

**NEW-1 (Major)** — the connection gate fixed the read side but not the actions keyed off it. Scenario A (catalog app): a grant-less mcp_oauth catalog app's card shows only "Connect"; Disconnect is gated behind the same `is_connected` flag F1 now correctly nulls — a workaround exists via the separate, ungated `GET /api/mcp/servers` listing, but that's a different screen. Scenario B (confirmed NEW regression, the sharper case): a user's own custom (non-catalog) `streamable_http` server with `auth.type=mcp_oauth` and no grant now renders `is_connected: false` in the local/all branch (this line previously hardcoded `is_connected: True`, confirmed via diff) — and that dict carries no `auth_type` field at all, so clicking "Connect" falls through to a generic `notConfigured` toast that does nothing, while the Delete button is also gone since `is_connected` is now false. This card is a confirmed, complete dead end for the user's own previously-deletable server.

**NEW-2 (Minor)** — `mcp_oauth_success=1` now permanently leaks onto `/build` routes. Confirmed real and newly introduced by this PR's success-redirect change, but assessed as cosmetic: `custom-mcp-form.tsx`'s OAuth popup already never self-closes regardless of this PR (a pre-existing gap the new doc comment on `MCP_OAUTH_POPUP_WINDOW_NAME` explicitly acknowledges), so that popup was already going to sit stuck on `/build/<agent-id>`. This PR just adds a stray, unread query param to that already-stuck popup's URL.

**NEW-3 (Minor)** — concurrent connects silently hijack each other's popup. The mcp_oauth connect popup always opens under the same fixed window name, with no per-app uniqueness and no guard against starting a second connect while a first is still in flight (`loadingApp` only disables the initiating app's own card). `window.open`'s standard name-reuse behavior then navigates the first popup to the second app's auth flow; the first app's poll correctly avoids reporting false success, but the user gets no error message explaining the silent failure.

**Minor — M3 Caveat 1**: individually-revoked grants escape the disconnect-time purge. The purge query filters `status == "active"`, but the per-grant DELETE endpoint only flips status to `"revoked"` without deleting the row (still holding encrypted tokens) — so an already-individually-revoked grant is invisible to the purge and survives indefinitely on any "keep the shared row" branch. This subsumes the former F10.

**Minor — M3 Caveat 2**: external HTTP I/O now runs inside the open write transaction — up to ~20s (2 tokens × N grants × a 10s timeout) held open before the final commit. If that commit then fails and rolls back, the grant rows are restored locally even though the tokens are already dead at the provider.

**Minor — F13 residual**: the first-provision path (`MCPServer.from_config`, not touched by this PR's diff) still calls `encrypt_value` on a sensitive field with only a truthiness check, no `isinstance(str)` guard — a non-string value raises an uncaught `AttributeError` that escapes as an unhandled 500 on the very first connect attempt, instead of a clean 400. Same one-line guard as the now-fixed sync path would close it.

---

### Design observation

Across all three rounds the trajectory is genuinely convergent, and this round's fixes are real engineering — M3's external-revoke work (async conversion, real HTTP calls, correctly placed before the early returns, tested against a mocked provider) is a proper root-cause fix. But the remaining gaps have narrowed to one consistent pattern: each round fixes the *read* side of `is_connected` without updating every consumer whose *action* depends on it. F1 was gated in `list_mcp_apps` but not in `GET /api/mcp/servers`. F3's root problem — committing state before it's real — keeps resurfacing wherever `is_connected` feeds a UI action, because the actual fix (don't create the association, or expose a manage surface, until there's something real to manage) was never done, only the display-layer symptom. NEW-1 scenario B is the sharpest version yet: correcting the read side turned a previously-working-if-slightly-wrong UI into a genuine dead end, because the Connect/Delete button logic keying off `auth_type`/`is_connected` wasn't updated in the same pass.

---

- There are now three near-identical popup-open-and-poll implementations — the mcp_oauth handler and the builtin_oauth handler in `connect-mcp-dialog.tsx` (both duplicating the `startedAt`/`maxWaitMs`/5-minute-expiry/`clearInterval`/ref-cleanup pattern), plus an independent third in `custom-mcp-form.tsx`. Extract a single `pollPopupClosed(popup, {timers, onClose, timeoutMs})` helper in `mcp-utils.ts` that all three call sites use — roughly 15-20 duplicated lines per copy in `connect-mcp-dialog.tsx` alone.

The backend is lean by contrast: `_redirect_after_with_params`, `_add_catalog_server_with_race_recovery`, `_reject_user_owned_catalog_squat`, and `_mcp_oauth_server_is_actually_connected` are all genuine multi-call-site extractions, and the external-revoke wiring cleanly reuses `_revoke_mcp_oauth_grant_externally` rather than reinventing it.

---

### Tests

Ran this round: `uv run pytest` across the five touched backend suites — **172 passed**; frontend — **42 passed**; `ruff` and `mypy` clean on touched Python files.

New tests are well-placed: `test_mcp_oauth_local_listing_also_requires_a_grant`, `test_delete_mcp_server_revokes_external_tokens_when_endpoint_is_advertised` (mocks the provider, asserts real POSTs), `test_delete_mcp_server_purges_the_users_own_flow_state_rows`, `test_admin_update_grandfathers_a_preexisting_bad_shape_row_for_unrelated_edits`, `test_connect_app_auth_sync_tolerates_a_malformed_sensitive_field`, plus eight new frontend tests for popup success/cancel/timeout/cleanup.

Coverage gaps: no test for NEW-1 scenario B (custom-server dead end); no test asserting an already-revoked grant survives disconnect untouched (M3 Caveat 1); no test for the create-path `encrypt_value` crash (F13 residual); no test for the concurrent-connect popup hijack (NEW-3).

src/xagent/web/api/mcp.py
```diff
@@ -2874,6 +3095,39 @@ def delete_mcp_server(
...
+        # (now awaitable, unlike the previous local-only status flip), then a
+        # hard delete rather than a "revoked" status flip: a revoked grant
+        # row is otherwise never swept, accumulating indefinitely as an
+        # inert but secret-bearing row (F10).
+        for grant in (
+            db.query(MCPOAuthGrant)
+            .filter(
+                MCPOAuthGrant.mcp_server_id == server_id,
+                MCPOAuthGrant.user_id == user_id,
+                MCPOAuthGrant.status == "active",
```
rogercloud, line 3112: **[M3 Caveat 1 — Minor, subsumes the former F10] This filter misses grants that were already individually revoked before this disconnect.**

`DELETE /api/mcp/{server_id}/oauth/grants/{grant_id}` (~line 3904) only flips a grant's `status` to `"revoked"` — it does not delete the row, which still holds the encrypted access/refresh tokens. This purge query, filtering only `status == "active"`, never sees those already-revoked rows, so on any "keep the shared row" branch (other users still connected, platform-key row, team-owned row) they survive indefinitely — the exact orphaned-secret-row problem this fix's own comment above says it's eliminating, just for a sub-case the filter doesn't cover. Suggest also purging (or at least also hard-deleting) `status == "revoked"` rows for this user/server here.

src/xagent/web/api/mcp.py
```diff
@@ -2874,6 +3095,39 @@ def delete_mcp_server(
...
+            )
+            .all()
+        ):
+            if isinstance(grant.oauth_client, MCPOAuthClient):
+                await _revoke_mcp_oauth_grant_externally(
```
rogercloud, line 3117: **[M3 Caveat 2 — Minor] This external HTTP call (and its siblings, one per active grant) now runs inside the open write transaction, before the commit at line 3133.**

With the provider's advertised timeout (`MCP_OAUTH_HTTP_TIMEOUT_SECONDS`) applied per token per grant, a single disconnect can hold this DB connection open for a nontrivial stretch under an unresponsive provider. More importantly: if the transaction's final commit then fails and rolls back, these grant rows are restored locally even though the tokens are already dead at the provider — an app that reports "connected" but can never actually work again. Consider moving the external revoke calls outside the transaction (best-effort, possibly queued/retried) and committing the local state change first.

PR #1097, opened 2026-07-31
feat: add Granola as a built-in remote-MCP connector for the Meeting Agent

yiboyasss:
> **Depends on #1096** — this PR is stacked on `feat/remote-mcp-oauth-catalog-connector` and must merge after it. The diff will shrink to the Granola-only commit (`0ed99e03`) once #1096 lands.

## Summary
- Register `granola` in the built-in catalog as the first connector using the remote-MCP OAuth (`mcp_oauth`) shape from #1096: `transport: streamable_http` pointing at Granola's hosted MCP server (`https://mcp.granola.ai/mcp`), `auth.type: mcp_oauth` — Granola issues no static client credentials, so connecting runs the per-user Dynamic Client Registration flow
- No local tool module (unlike the Zoom/AWS-style connectors): Granola's hosted server exposes its own meeting-notes tools (search/list/read meetings, transcripts), so the connector is a catalog row + seed migration only
- Seed migration inserts the catalog row only — no `oauth_providers` row exists for a DCR connector; downgrade removes just the row, leaving connect-driven `MCPServer`/`MCPOAuth*` rows to the normal disconnect path
- Wire the Meeting Agent template (`sales-meeting-agent.yaml`) to Granola per the Advanced Templates spec: `connections` entry, feature copy (EN/ZH), and instructions steering the agent to look meetings up in Granola notes first, falling back to a pasted/uploaded transcript
- Move the oauth-flow test fixture's synthetic app_id from `granola` to `remote-notes`: `granola` is now a real builtin registry entry, and the builtin execution overlay replaces a DB row's execution fields for matching app_ids, which would silently override the fixture's url/auth

## Test plan
- [x] `pytest tests/alembic/test_20260731_seed_granola_mcp_app.py tests/web/api/test_public_mcp_connector_visibility.py tests/templates/test_manager.py tests/web/test_classify_app_auth.py tests/web/api/test_mcp_oauth_flow.py tests/web/api/test_mcp_apps_connect.py` — 161 passed
- [x] New tests: migration upgrade/downgrade round-trip, idempotency, registry-drift guard (migration ROW == registry row), and a classification guard asserting the seeded shape classifies as `mcp_oauth` (an `unconnectable` classification would make the entry dead on arrival)
- [x] Template-connections resolution test passes (Granola resolves against the registry)
- [x] `alembic upgrade head` / `downgrade -1` round-trip verified against a local dev DB — only the `granola` row is added/removed
- [x] `ruff format` / `ruff check` / `mypy` / pre-commit (incl. alembic-check single-head) all clean

### DIFF
--- a/src/xagent/templates/built_in/sales-meeting-agent.yaml
+++ b/src/xagent/templates/built_in/sales-meeting-agent.yaml
@@ -3,22 +3,26 @@ name: Meeting Agent
 category: Sales
 featured: false
 descriptions:
-  en: Turns a Zoom meeting — or a transcript you paste or upload — into a summary, decisions, action items and a follow-up email draft, grounded in the transcript, nothing invented.
-  zh: 将 Zoom 会议——或您粘贴/上传的会议记录——转化为摘要、决策、行动项和跟进邮件草稿，完全基于记录内容，不臆造。
+  en: Turns a meeting — from Zoom, your Granola notes, or a transcript you paste or upload — into a summary, decisions, action items and a follow-up email draft, grounded in the transcript, nothing invented.
+  zh: 将会议——来自 Zoom、您的 Granola 笔记，或您粘贴/上传的会议记录——转化为摘要、决策、行动项和跟进邮件草稿，完全基于记录内容，不臆造。
 features:
   en:
   - Looks up a Zoom meeting by its meeting id and reads its cloud-recording transcript directly, once Zoom is connected
-  - Also accepts a pasted or uploaded transcript (.txt, .docx or .pdf) if Zoom isn't connected or the meeting wasn't recorded
+  - Or finds a meeting in your Granola notes by topic or date and reads its notes and transcript directly, once Granola is connected
+  - Also accepts a pasted or uploaded transcript (.txt, .docx or .pdf) if neither is connected or the meeting isn't available there
   - Extracts decisions, owners and due dates from the transcript on request
   - Drafts a follow-up email in Gmail and can log a CRM note to HubSpot after your approval
   zh:
   - 连接 Zoom 后，可通过会议 ID 查找会议，直接读取其云录制转录文本
-  - 若未连接 Zoom 或该场会议没有录制，也可接受您粘贴/上传的会议记录（.txt、.docx 或 .pdf）
+  - 或连接 Granola 后，可按主题或日期在您的 Granola 笔记中查找会议，直接读取其笔记与转录文本
+  - 若两者均未连接，或该会议在其中不可用，也可接受您粘贴/上传的会议记录（.txt、.docx 或 .pdf）
   - 按需从记录中提取决策、负责人和截止日期
   - 在 Gmail 中起草跟进邮件，并在您批准后于 HubSpot 中记录 CRM 备注
 connections:
 - name: Zoom
   logo: https://www.google.com/s2/favicons?domain=zoom.us&sz=64
+- name: Granola
+  logo: https://www.google.com/s2/favicons?domain=granola.ai&sz=64
 - name: HubSpot
   logo: https://www.google.com/s2/favicons?domain=hubspot.com&sz=64
 - name: Gmail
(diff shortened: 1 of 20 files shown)


rogercloud:
ci is failing

src/xagent/templates/built_in/sales-meeting-agent.yaml
```diff
@@ -51,6 +55,7 @@ agent_config:
 
     - Structure output as: ## Summary · ## Decisions · ## Action items (owner · due date) · ## Draft email.
     - Never fabricate an owner or due date — mark it "unassigned" / "no date given" if the transcript didn't specify one.
+    - Always cite the meeting's title and date in your output.
```
rogercloud: **Minor:** this new reminder has no fallback for the unknown case, unlike its immediate sibling one line above (`Never fabricate an owner or due date — mark it "unassigned" / "no date given" ...`). The pasted/uploaded-transcript path can supply neither a title nor a date, so an unconditional "always cite the title and date" nudges the model to invent them — in tension with the rest of the prompt's grounded, nothing-invented framing.

Suggested wording:

```suggestion
    - Always cite the meeting's title and date if known — say "untitled" / "date not given" if the source doesn't provide one.
```

tests/alembic/test_20260731_seed_granola_mcp_app.py
```diff
@@ -0,0 +1,122 @@
...
+    from xagent.web.builtin_mcp_registry import get_builtin_public_mcp_app_rows
+
+    migration = _load_migration_module()
+    registry_row = next(
+        r for r in get_builtin_public_mcp_app_rows() if r["app_id"] == "granola"
+    )
+    assert migration.ROW == registry_row
+
+
+def test_seed_row_classifies_as_mcp_oauth(tmp_path):
```
rogercloud, line 101: **Nit / optional simplification:** this assertion is transitively implied. `test_seed_row_matches_registry` above already asserts full equality (`migration.ROW == registry_row`), `classify_app_auth` is a pure function, and `tests/web/api/test_public_mcp_connector_visibility.py` already classifies the registry row as `mcp_oauth`. So this test can only fail in tandem with one of those two.

Fine to keep as belt-and-suspenders in case the drift guard is ever weakened — just flagging it as a candidate to trim (~3-8 lines). Also note `tmp_path` is unused here and in `test_seed_row_matches_registry`.

src/xagent/web/mcp_apps.py
```diff
@@ -78,20 +78,39 @@ def restrict_to_app_scoped_oauth_grant(
...
+        - "unconnectable": none of the above
     """
     if str(transport or "").lower() == "oauth":
         return "builtin_oauth"
     launch = launch_config if isinstance(launch_config, dict) else {}
     if launch.get("required_env") and launch.get("command"):
         return "api_key"
+    auth = launch.get("auth")
+    if (
+        str(transport or "") in _REMOTE_MCP_TRANSPORTS
```
rogercloud: Re-verified on this PR's tip: still unresolved (referencing the existing thread on this line). Noting for the record that this file is **not** touched by this PR's own commit (`40141538..2709b5d1`) — it comes from the stacked #1096, so the fix belongs there.

Confirming this is reachable rather than purely defensive: `str(transport or "")` is compared without `.lower()`, while the `"oauth"` check two lines above does normalize. The admin write path (`PublicMCPAppCreate` in `admin_mcp.py`) does not normalize `transport` case either, so a mixed-case value stored via the admin API would fail this check on every read and be silently reported as `unconnectable`. Worth fixing on #1096 before it merges, since Granola is the first connector to depend on this classification in production.

src/xagent/web/api/mcp.py
```diff
@@ -2436,6 +2543,72 @@ def _apply_updates(a: Any) -> None:
...
+            UserMCPServer.user_id == current_user.id,
+            UserMCPServer.mcpserver_id == server.id,
+        )
+        .first()
+    )
+    if assoc is None:
+        assoc = UserMCPServer(
+            user_id=current_user.id,
+            mcpserver_id=server.id,
+            is_active=True,
```
rogercloud, line 2624: Re-verified on this PR's tip: still unresolved (referencing the existing thread near this line). Again, this file is not touched by this PR's own commit — it belongs to the stacked #1096.

The `UserMCPServer` row is created with `is_active=True` before the DCR/OAuth popup flow completes, and `list_mcp_apps` derives "connected" for `mcp_oauth` apps purely from `UserMCPServer.is_active` with no `MCPOAuthGrant` existence check. Traced this through the frontend popup-close handler in `connect-mcp-dialog.tsx` as well — there is no reconciliation step anywhere, so abandoning the popup leaves the connector displayed as "connected" indefinitely. Recommend either gating `is_active` on grant creation or having the connected-status query join against `MCPOAuthGrant`, on #1096.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -380,6 +380,26 @@ def get_builtin_public_mcp_app_rows() -> list[dict[str, Any]]:
                 "required_env": ["GOOGLE_MAPS_API_KEY"],
             },
         },
+        {
+            "app_id": "granola",
```
rogercloud, line 479: **FYI, no change required.** The connect path keys `MCPServer.name` off `app_id`, while template `connections:` entries are authored as this display `name` — the two only line up after `normalize_mcp_server_name`'s case/space/hyphen folding. Verified it resolves correctly for `Granola` ↔ `granola` (as it does for `Google Maps` ↔ `google-maps`).

A future `mcp_oauth` connector whose display name diverges from its `app_id` by more than case/space/hyphen would silently produce a template selector matching nothing, with no error raised. A one-line comment here noting the constraint would help the next connector author.

rogercloud (approved):
## Summary

Reviewed **only PR #1097's own commit** (`40141538..2709b5d1`), i.e. the diff on top of the stacked, not-yet-merged #1096. GitHub's diff-vs-`main` view also shows #1096's platform commits; those were deliberately **not** re-reviewed here.

Within that scope the change adds Granola as a built-in remote-MCP (`mcp_oauth`) connector: a new seed migration (`src/xagent/migrations/versions/20260731_seed_granola_mcp_app.py`), the matching runtime entry in `src/xagent/web/builtin_mcp_registry.py`, a `connections:` entry plus a prompt tweak in the `sales-meeting-agent` built-in template, a test-fixture rename to free up the `granola` app_id, and new tests covering the migration (insert / idempotency / downgrade / registry-drift / auth classification) and registry-level visibility and classification.

## Round 0 — approach verdict: **sound**

Minimal and precedent-consistent — effectively a line-for-line clone of the existing `google-maps` / `google-ads` seed-migration pattern. Specifically verified:

- `classify_app_auth("streamable_http", {"url": ..., "auth": {"type": "mcp_oauth"}})` returns `mcp_oauth`, so the catalog row is genuinely connectable rather than dead on arrival.
- The migration/registry duplication is not new duplication: the builtin execution overlay (`get_builtin_execution_fields`) means the **registry** copy governs runtime behavior, with the migration acting as a frozen seed. `test_seed_row_matches_registry` guards the drift. This is the same convention every other builtin connector seed already uses.
- Downgrade semantics (drop only the catalog row, leave connect-created rows in place) match precedent and are correct here — Granola is DCR-only, so there is no `oauth_providers` row to unwind.
- The template `connections:` entry resolves end-to-end: display name `Granola` → `tool_category` `mcp:Granola` → normalized to `granola` → matches `MCPServer.name` / `app_id` `granola` (traced through `templates/manager.py`, `selection_spec.py`, and the connect endpoint in `web/api/mcp.py`).

### Design-level FYI (not a required change)

The remote-MCP-OAuth connect path keys `MCPServer.name` off `app_id`, whereas template `connections:` entries are authored as the human **display name**. The two are only equivalent after `normalize_mcp_server_name`'s case/space/hyphen folding. It works for `Granola` ↔ `granola` (and `Google Maps` ↔ `google-maps`), but a future `mcp_oauth` connector whose display name diverges from its `app_id` by more than case/space/hyphen would silently yield a template selector that matches nothing, with no error anywhere. A one-line note next to the registry entry would save the next connector author a debugging session.

## Findings

### Minor

1. **`src/xagent/templates/built_in/sales-meeting-agent.yaml:58`** — the new reminder `Always cite the meeting's title and date in your output.` has no unknown-value fallback, unlike its immediately preceding sibling (`Never fabricate an owner or due date — mark it "unassigned" / "no date given" ...`). The pasted/uploaded-transcript path can supply neither a title nor a date, so this line nudges the model toward inventing them, in tension with the rest of the prompt's grounded/anti-fabrication framing.

### Nits (optional, non-blocking)

2. **`tests/alembic/test_20260731_seed_granola_mcp_app.py:71-72`** — `assert "https://mcp.granola.ai/mcp" in str(row[2])` / `assert "mcp_oauth" in str(row[2])` are substring checks on the serialized `launch_config`, so a malformed `auth` (e.g. a bare string) would still pass. Practically compensated by `test_seed_row_classifies_as_mcp_oauth` in the same file, which runs the real `classify_app_auth` and requires `isinstance(auth, dict)`. Also a pre-existing pattern copied from the google-maps seed test, not something introduced here — flagging for awareness only.
3. **`tests/alembic/test_20260731_seed_granola_mcp_app.py:89,101`** — `test_seed_row_matches_registry` and `test_seed_row_classifies_as_mcp_oauth` take an unused `tmp_path` parameter. Mirrors the same pattern in the existing google-maps seed test; trivial cleanup if you're touching the file anyway.

No correctness, security, or migration-safety issues found in this PR's own diff.

- `tests/alembic/test_20260731_seed_granola_mcp_app.py` **L101-111: shrink**: `test_seed_row_classifies_as_mcp_oauth` is transitively implied — `test_seed_row_matches_registry` (L89-98) already asserts full equality `migration.ROW == registry_row`, `classify_app_auth` is pure, and `tests/web/api/test_public_mcp_connector_visibility.py:767-775` already classifies the registry row. Replacement: drop the alembic-side classify test and rely on the equality guard plus the registry-side classification.

Note this is a genuine judgment call: keeping it is defensible belt-and-suspenders in case someone later weakens the drift guard. Net: **-3 to -8 lines possible, optional.**

## Prior review findings (gemini-code-assist) — still unresolved, but belong to #1096

Both prior inline comments were re-verified against this PR's tip and are **still not fixed** — but `git log -p 40141538..2709b5d1` shows #1097's commit touches **neither** file, so both belong to the stacked platform PR #1096, not here:

1. **`src/xagent/web/mcp_apps.py:108` (HIGH)** — the remote-transport check `str(transport or "") in _REMOTE_MCP_TRANSPORTS` is not case-normalized, while the sibling `"oauth"` check two lines above does call `.lower()`. This is reachable, not merely defensive: the admin write path (`PublicMCPAppCreate` in `admin_mcp.py`) does not normalize `transport` case either, so a mixed-case value stored via the admin API would fail classification on every read and silently surface as `unconnectable`.
2. **`src/xagent/web/api/mcp.py:2576` (MEDIUM)** — `connect_mcp_oauth_app` creates the `UserMCPServer` association with `is_active=True` **before** the DCR/OAuth popup flow completes, and `list_mcp_apps` derives "connected" for `mcp_oauth` apps purely from `UserMCPServer.is_active` with no `MCPOAuthGrant` existence check. Traced end-to-end including the frontend popup-close handler in `connect-mcp-dialog.tsx`: there is no reconciliation anywhere. A user who abandons the popup mid-flow sees the connector as "connected" indefinitely.

**Recommendation:** fix both on #1096 before it merges, or immediately after. Granola is the first connector to exercise this `mcp_oauth` path in production and inherits both behaviors the moment #1096 lands.

Also out of scope but worth a future consistency pass: `frontend/.../connect-mcp-dialog.test.tsx` still uses `granola` as a mock `app_id`, which the backend fixture rename in this PR was moving away from. Harmless here — that test fully mocks the API layer, so no real registry lookup or execution overlay is involved — and the file is untouched by this commit.

## Testing

I was **not** able to execute the test suite in the review environment (no project virtualenv available, and the system Python is too old for some test-support stub syntax). All findings above were verified statically by reading the current file contents and by hand-evaluating the real `classify_app_auth` / registry / template-resolution logic — not inferred from test names. A CI green run should be treated as the authority on test correctness.

## Verdict

No blockers for #1097 itself. The one minor prompt-wording finding is worth taking; the nits and the simplification are optional. The two upstream issues above are the real risk carried by this feature and should be tracked on #1096.

rogercloud:
please resolve the conflicts

PR #1100, opened 2026-08-01
ci: pre-pull the sandbox image before docker-backed fast tests

### LINKED ISSUE #1099: ci: pre-pull the sandbox image before docker-backed fast tests
## Problem

`Pytest Fast (web)` failed on main with 6 failures, all in
`tests/web/test_sandbox_reconcile_e2e.py`, all with the same root cause:

```
docker.errors.APIError: 500 Server Error for
.../images/create?tag=latest&fromImage=xprobe%2Fxagent-sandbox:
Internal Server Error ("Get "https://registry-1.docker.io/v2/": context deadline exceeded")
```

Run: https://github.com/xorbitsai/xagent/actions/runs/30644180191/job/91202272011
(`6 failed, 3802 passed`; every other job in the run was green, and the same
commit passed in run 30644180165.)

These tests drive a real Docker daemon. Before creating a container,
`_ensure_image()` in `src/xagent/sandbox/docker_sandbox.py:974` pulls
`xprobe/xagent-sandbox:latest` when it is not present locally. The runner has no
local copy, so every run reaches out to Docker Hub mid-test and is exposed to
registry timeouts and anonymous pull rate limits. The failure is transient
infrastructure, unrelated to the code under test -- but it is structurally
guaranteed to recur.

The `e2e` job already avoids this: it has a "Pull e2e service images" step that
pulls with 3 retries before pytest starts, precisely so "registry failures
surface here instead of as confusing Docker daemon errors inside e2e fixtures"
(`.github/workflows/ci.yml:603`). The fast pytest jobs have no equivalent.

## Proposal

Add a pre-pull step, mirroring the existing e2e retry loop, to every fast pytest
job leg that drives a real Docker daemon:

- `pytest-fast` / `web` -- `tests/web/test_sandbox_reconcile_e2e.py` and the
  other sandbox tests under `tests/web`.
- `pytest-fast-deepdoc` / `core` -- `tests/sandbox` has the same exposure
  (`test_docker_lifecycle_api.py` and friends also drive a real daemon).

The image is `get_sandbox_image()`'s default, `xprobe/xagent-sandbox:latest`.

Effects:
- Registry flakes fail in a clearly-named setup step instead of masquerading as
  test failures.
- The retry loop absorbs single transient timeouts, which is the common case.
- Tests find the image locally, so `_ensure_image()` never hits the network.

rogercloud:
Closes #1099.

## Problem

`Pytest Fast (web)` failed on main with `6 failed, 3802 passed`, every failure in
`tests/web/test_sandbox_reconcile_e2e.py` and every one with the same cause:

```
docker.errors.APIError: 500 Server Error for
.../images/create?tag=latest&fromImage=xprobe%2Fxagent-sandbox:
Internal Server Error ("Get "https://registry-1.docker.io/v2/": context deadline exceeded")
```

Those tests drive a real Docker daemon. `_ensure_image()`
(`src/xagent/sandbox/docker_sandbox.py:974`) pulls `xprobe/xagent-sandbox:latest`
when the runner has no local copy, so the pull happens mid-test and a Docker Hub
timeout or an anonymous pull rate limit surfaces as a pile of `APIError` test
failures instead of a clearly-named setup failure. Transient infrastructure, but
structurally guaranteed to recur.

The `e2e` job already guards against this with a retrying pre-pull step; the fast
pytest jobs had no equivalent.

## Change

- Add a "Pre-pull the sandbox image" step to the two legs that drive a real
  daemon: `pytest-fast` / `web` (`tests/web`) and `pytest-fast-deepdoc` / `core`
  (`tests/sandbox`). Both run before pytest.
- Resolve the tag from `get_sandbox_image()` rather than hardcoding it, so the
  pre-pull cannot drift from what `_ensure_image()` actually requests.
- Factor the retry loop the `e2e` job already had into
  `.github/scripts/docker-pull-retry.sh` and share it across all three call
  sites, instead of a third copy of the same twelve lines.

The script takes one or more images, retries `PULL_ATTEMPTS` times (default 3)
with `PULL_BACKOFF_SECONDS` between attempts (default 15), and exits non-zero if
any image cannot be pulled.

## Verification

- `ci.yml` parses; step ordering confirmed — the pre-pull lands between
  "Install pptxgenjs globally" and "Run tests" in both jobs.
- `bash -n` clean. Script exercised against a stub `docker` covering all four
  paths: fail-then-succeed (retry) → 0, multiple images → 0, attempts exhausted
  → 1, no arguments → 2 with usage.
- The image resolver one-liner prints `xprobe/xagent-sandbox:latest` locally.

The real-pull success path is only exercised by CI on this PR — locally it was
covered with a stub daemon.

### DIFF
--- a/.github/scripts/docker-pull-retry.sh
+++ b/.github/scripts/docker-pull-retry.sh
@@ -20,7 +20,7 @@ fi
 
 for image in "$@"; do
   pulled=false
-  for i in $(seq 1 "$ATTEMPTS"); do
+  for ((i = 1; i <= ATTEMPTS; i++)); do
     if docker pull "$image"; then
       pulled=true
       break
--- a/.github/scripts/docker-pull-retry.sh
+++ b/.github/scripts/docker-pull-retry.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# Pull one or more images, retrying transient registry failures.
+#
+# CI jobs that drive a real Docker daemon must pre-pull every image their tests
+# need. Otherwise the pull happens inside a test fixture, and a Docker Hub
+# timeout or an anonymous pull rate limit surfaces as a pile of confusing
+# APIError test failures instead of a clearly-named setup failure.
+#
+# Usage: docker-pull-retry.sh IMAGE [IMAGE...]
+
+set -euo pipefail
+
+ATTEMPTS="${PULL_ATTEMPTS:-3}"
+BACKOFF_SECONDS="${PULL_BACKOFF_SECONDS:-15}"
+
+if [ "$#" -eq 0 ]; then
+  echo "usage: $0 IMAGE [IMAGE...]" >&2
+  exit 2
+fi
+
+for image in "$@"; do
+  pulled=false
+  for i in $(seq 1 "$ATTEMPTS"); do
+    if docker pull "$image"; then
+      pulled=true
+      break
+    fi
+    if [ "$i" -lt "$ATTEMPTS" ]; then
+      echo "Pull attempt $i for $image failed; retrying in ${BACKOFF_SECONDS}s..."
+      sleep "$BACKOFF_SECONDS"
+    fi
+  done
+  if [ "$pulled" != true ]; then
+    echo "Failed to pull $image after $ATTEMPTS attempts" >&2
+    exit 1
+  fi
+done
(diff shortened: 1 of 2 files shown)


qinxuye (approved):
LGTM

PR #1068, opened 2026-07-30
feat: add AWS MCP connector

yiboyasss:
## Summary
- Register `aws` as a built-in key-based MCP connector (`builtin_mcp_registry.py` + seed migration) — access-key auth via the connect dialog (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_REGION`), mirroring the `google-maps` pattern since AWS isn't OAuth
- Add `boto3` as a core dependency (`pyproject.toml` + `uv.lock` regenerated) — no existing SigV4 code in the repo, and the SDK covers CloudWatch/Logs/DynamoDB/SQS's differing wire protocols
- New `xagent.web.tools.mcp.aws` module, nine read-only tools:
  - STS: `aws_get_caller_identity`
  - CloudWatch: `aws_cloudwatch_describe_alarms`, `aws_cloudwatch_get_metric_data`, `aws_cloudwatch_describe_log_groups`, `aws_cloudwatch_filter_log_events`
  - DynamoDB: `aws_dynamodb_list_tables`, `aws_dynamodb_describe_table`
  - SQS: `aws_sqs_list_queues`, `aws_sqs_get_queue_attributes`
- Read-only is enforced by construction — only read/describe/list APIs are implemented, no write calls exist in the module
- Cross-account access via an optional `role_arn` parameter on every tool (STS AssumeRole exchanged fresh on every call — each MCP tool call runs in its own short-lived subprocess, so there's no in-process layer where caching credentials would help) rather than a connection-time field, since the platform's `required_env` has no optional-key concept

## Test plan
- [x] `pytest tests/web/tools/test_aws_mcp.py tests/alembic/test_20260731_seed_aws_mcp_app.py tests/templates/test_manager.py` — all passing
- [x] `ruff format` / `ruff check` clean, `mypy` clean on touched files
- [x] `uv lock` / `uv sync --locked` succeed (Docker build parity)
- [x] `alembic upgrade head` / `downgrade -1` round-trip verified locally — only the `aws` row is added/removed
- [x] Manual end-to-end with a real read-only IAM key: connected via the connector UI, verified `aws_get_caller_identity`, CloudWatch alarms, log search, DynamoDB table lookup, and SQS queue attributes all working

PR #1102, opened 2026-08-01
feat(templates): update DevOps AI Agent template

yiboyasss:


src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -33,41 +33,48 @@ sample_prompts:
     prompt: 我粘贴了一份监控导出数据——请将其与错误日志关联，告诉我究竟是什么导致了这个问题。
     highlights:
     - 监控导出数据
-connections: []
+connections:
+- name: AWS
```
rogercloud, line 37: **Critical — this line breaks `tests/templates/test_manager.py::TestTemplateManager::test_builtin_template_connections_resolve_to_a_registered_mcp_app`, and CI should be red.**

At PR head this test fails with `AssertionError: assert not ['operations-devops-ai-agent.yaml: AWS']` (base commit: 18 passed; head: 1 failed, 17 passed). No `AWS` app is registered in `src/xagent/web/builtin_mcp_registry.py`; the AWS MCP connector lives in a separate, still-open, unmerged PR (#1068). On this codebase, a search for cloudwatch/dynamodb/boto3/sqs returns zero hits outside this file's own prose.

This is not cosmetic metadata: `src/xagent/templates/manager.py:193-201` appends `mcp:<name>` for each connection to the agent's served `tool_categories`, so this template ships a dangling `mcp:AWS` category.

Two concrete consequences:
1. **Trigger-fired runs fail outright.** Trigger-sourced runs use STRICT MCP failure policy (`src/xagent/web/api/chat.py:490-494`); an unresolvable scoped category is "requested but unavailable" (`mcp_tools.py:218-226`, `no_tools_returned` at line 125) and `enforce_mcp_failure_policy` (`config.py:176-183`) raises `RequiredMCPUnavailableError`. That is exactly the webhook/alert/schedule behavior this template advertises as its primary feature.
2. **Plain chat fails silently.** No warning is logged; only an internal `__audit_only__` `mcp_load_summary` trace event (`src/xagent/web/tools/config.py:1782-1801`). The agent will fabricate CloudWatch/DynamoDB/SQS findings instead of reporting missing access.

Also note `frontend/src/components/build/agent-builder.tsx`: `templateMissingMcp` (lines 1736-1745) is permanently true for this template, so the wizard's Configure step never shows complete (lines 1749-1758), with no skip path. It does not block agent creation, but it is a persistent "incomplete" badge.

**Suggested fix:** remove the `AWS` connection entry and the instructions that depend on AWS data sources until an AWS MCP app is actually registered on `main`. Keep `Slack`, which does resolve to a real registered app.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,17 +3,17 @@ name: DevOps AI Agent
...
-  en: Triages a diagnostic bundle you paste or upload (logs, command output, monitoring exports) and reports a root-cause hypothesis and recommended fix directly in chat.
-  zh: 对您粘贴或上传的诊断信息（日志、命令输出、监控导出）进行分诊，并直接在对话中汇报根因假设与修复建议。
+  en: Monitors your AWS environment, triages alerts and incidents, runs read-only diagnostics over SSH, and posts findings plus a recommended fix to Slack — read-only by default, action gated by human approval.
+  zh: 监控 AWS 环境，对告警与事件进行分诊，通过 SSH 执行只读诊断，并将排查结果和修复建议发布到 Slack——默认只读，任何操作变更均需人工审批。
 features:
   en:
-  - Reads logs, command output or monitoring exports you paste or upload — no live server connection required
-  - Correlates findings into a root-cause hypothesis and a recommended fix
-  - Reports directly in chat — no external monitoring or messaging integration required
+  - Wakes on an alert / webhook, a schedule, or a direct chat request
```
rogercloud, line 10: **Critical (same root cause as the `AWS` connection at line 37).** This line — and lines 52 and 73 — instruct the agent to pull CloudWatch/DynamoDB/SQS data that no registered tool can provide on this codebase. Line 73 in particular tells the agent to "report the access failure itself rather than silently skipping that data source," but the failure mode here never reaches the model: in chat, unresolved MCP categories are recorded only in an `__audit_only__` trace event, so the agent has no signal and will confabulate findings. In trigger runs the whole execution raises `RequiredMCPUnavailableError` before the model is asked anything.

**Suggested fix:** drop the AWS/CloudWatch/DynamoDB/SQS data-source claims from the instruction block until the connector exists, or gate this template behind deployments where those tools are actually bound.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,17 +3,17 @@ name: DevOps AI Agent
...
+  - Investigates via CloudWatch, DynamoDB and SQS, plus read-only SSH diagnostics
+  - Triages against a known-noise catalog and posts a root-cause hypothesis with a recommended fix to Slack
   zh:
-  - 读取您粘贴或上传的日志、命令输出或监控导出——无需连接到实时服务器
-  - 将排查结果关联为根因假设与修复建议
-  - 直接在对话中汇报——无需外部监控或消息集成
+  - 由告警 / webhook、定时任务或聊天请求触发
+  - 通过 CloudWatch、DynamoDB、SQS 及只读 SSH 诊断进行排查
+  - 对照已知噪声名录进行分诊，并将根因假设与修复建议发布到 Slack
 sample_prompts:
```
rogercloud, line 17: **Major — `sample_prompts` now contradict the rewritten framing.** These prompts (unchanged by this PR) put the user in charge of supplying data: "Here are the logs and command output from the incident…", "I'm pasting a monitoring export below…". Lines 6, 11, and 53 now describe an agent that pulls CloudWatch/DynamoDB/SQS/SSH data itself. The gallery card, description, and starter prompts should tell one consistent story — either update these prompts to match the active-monitoring framing, or keep the paste-driven framing throughout.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -33,41 +33,48 @@ sample_prompts:
...
     # Responsibilities
 
-    1. **Get the material** — If the user hasn't already pasted or attached diagnostic output, ask for the specific logs or command output relevant to the problem.
-    2. **Correlate** — Cross-reference what was provided; rule out likely red herrings before settling on a hypothesis.
-    3. **Report** — Post a summary, the evidence, a root-cause hypothesis, a recommended fix, and a severity assessment, directly in this chat.
-    4. **Recommend, don't apply** — You have no way to run anything on the user's systems. Always phrase the fix as a recommendation for the user (or their own tooling) to carry out.
+    1. **Receive the trigger** — An alert / webhook, a scheduled check, or a direct chat request.
+    2. **Investigate, read-only** — Pull relevant CloudWatch metrics/logs, DynamoDB and SQS state, and run read-only SSH diagnostics where needed.
+    3. **Triage** — Correlate signals, filter out entries matching the known-noise catalog, and identify the most likely root cause.
+    4. **Report & recommend** — Post an incident summary, evidence, root-cause hypothesis, recommended fix and severity to Slack. Wait for approval before any change is made.
```
rogercloud: **Major — unenforced safety promise, and an unresolved self-contradiction previously flagged in review.**

Contradiction: lines 48 and 72 state absolutely that this agent recommends fixes and does not apply them, while this line and line 59 say to wait for Slack approval "before any change is made" / never mutate "without explicit human approval in Slack first" — implying mutation *is* available with approval. The earlier bot review raised this and it is still open; both phrasings remain verbatim.

Enforcement: nothing in the platform backs the read-only/approval-gated claim. `SshExecuteTool` (`src/xagent/core/tools/adapters/vibe/ssh_tools.py:61,175-186`) forwards `command: str` verbatim to `SshExecutor.execute` (`src/xagent/core/ssh/executor.py:92-156`) and on to `asyncssh.create_process` (`src/xagent/core/ssh/runner.py:172-213,312`). There is no allowlist, denylist, or command classification on that path — only a capability-membership check and an egress policy, neither of which inspects the command. The `ssh` category also grants `SshUploadTool`/`SshDownloadTool` (`ssh_tools.py:197-264`), i.e. mutating file transfer. `ResolvedSshTarget.approval_policy` (`src/xagent/core/ssh/types.py:11,78`) is never read or branched on outside test fixtures — dead config. This is the same P1 raised on PR #1023 and it was never fixed, only removed from the template.

**Suggested fix:** pick one model and make the copy honest. Advisory-only: delete the approval-gating language and state the prohibition unconditionally. Approval-gated mutation: implement real enforcement in `SshExecutor` honoring `approval_policy` before the template claims it.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -33,41 +33,48 @@ sample_prompts:
...
+    - Slack message structure: incident summary · evidence · root-cause hypothesis · recommended fix · severity.
+    - Never apply a fix yourself — always end with a recommendation awaiting approval.
+    - If SSH, CloudWatch, DynamoDB, or SQS access fails, report the access failure itself rather than silently skipping that data source.
   execution_mode: balanced
   skills: []
   tool_categories:
   - basic
   - file
   - knowledge
+  - ssh
```
rogercloud, line 80: **Major — `ssh` will very likely resolve to zero tools for most deployments, and this is the only built-in template that requests it.**

`src/xagent/web/services/ssh_runtime.py` only supplies SSH tools when an out-of-tree provider hook is installed (per its docstring the DB-backed provider is in closed-source xagent-cloud; no `SshTarget`/`SshCredential` model exists under `src/xagent/web/models/` here). `create_ssh_tools` returns `[]` with no provider (`ssh_tools.py:369-372`), returns `[]` unless the agent has at least one bound SSH target (`ssh_tools.py:404-406`), and is hard-disabled under `SANDBOX_IMPLEMENTATION == "boxlite"` (`ssh_tools.py:427-429`).

No other file under `src/xagent/templates/built_in/` lists `ssh`, so this sets an unvetted precedent. Self-hosted and standard users picking this template get an agent whose instructions reference SSH diagnostics while having zero SSH tools bound — silently, with no user-facing explanation.

**Suggested fix:** if this template is cloud-only, gate its availability explicitly rather than shipping it as a plain built-in; otherwise remove `ssh` and the SSH-dependent instructions.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -33,41 +33,48 @@ sample_prompts:
...
-connections: []
+connections:
+- name: AWS
+  logo: https://www.google.com/s2/favicons?domain=aws.amazon.com&sz=64
+- name: Slack
+  logo: https://www.google.com/s2/favicons?domain=slack.com&sz=64
 setup_time:
-  en: 5 min setup
-  zh: 5分钟配置
+  en: 10 min setup
```
rogercloud: **Minor —** `10 min` is the only such value across the 18 built-in templates; every other is `5 min setup` (16) or `2 min setup` (1). Plausible given the added connections, but other multi-connection templates still use 5 min — worth a brief rationale or aligning with the existing convention.

rogercloud (changes requested):
## Summary

This PR rewrites a single built-in template, `src/xagent/templates/built_in/operations-devops-ai-agent.yaml`, from a "paste-only diagnostic triage" agent into one that claims to actively monitor AWS (CloudWatch/DynamoDB/SQS), run SSH diagnostics on hosts, post to Slack, and wake on webhook/alert/schedule triggers. It adds `ssh` to `tool_categories` and adds `AWS` + `Slack` `connections` entries, and rewrites the description and instruction block accordingly. The PR body is empty and there is no linked issue, so there is no stated rationale for the change of direction.

## Approach verdict: WRONG DIRECTION

**This PR is, byte-for-byte, a revert of a version that review already rejected on this same file.**

PR #1023 originally proposed exactly this AWS/Slack-monitoring template. It received P1 review findings and was rewritten to the current paste-only diagnostics form before merging; the approval on #1023 explicitly recorded that all findings were fixed. Diffing #1023's pre-review commit against this PR's head shows the two are identical apart from the `sample_prompts` block added later. So this PR restores the rejected version verbatim — with an empty PR body that neither acknowledges that history nor explains which of the original blockers are now resolved.

Of the four original P1 concerns, **only one has since been resolved**:

| Original P1 concern | Status today |
| --- | --- |
| Slack is not a registered MCP app | **RESOLVED** — a real `Slack` app now exists in the built-in MCP registry backed by a working `slack_post_message` tool |
| AWS is not a registered MCP app | **NOT RESOLVED** — no AWS app is registered anywhere; the connector lives in a still-open, unmerged PR (#1068) |
| Read-only / approval-gated SSH safety is advertised but not enforced | **NOT RESOLVED** — see M3 |
| SSH-only template needs gating to deployments that can bind SSH targets | **NOT RESOLVED** — see M4 |

Three of four blockers were never actually fixed; they were only *removed from the template* in #1023, and this PR reintroduces them.

Importantly, `connections` here is **not decorative UI metadata**: `src/xagent/templates/manager.py:193-201` deterministically appends `mcp:<name>` for each entry to the agent's served `tool_categories`. This template therefore ships `tool_categories = [basic, file, knowledge, ssh, mcp:AWS, mcp:Slack]`, where `mcp:AWS` is a dangling reference to nothing that exists on `main`.

**Recommendation:** do not land this as-is. Either (a) hold it until the AWS MCP connector (#1068) merges and the SSH enforcement gap is closed, then re-scope; or (b) land a version that only claims capabilities the platform actually provides today (Slack + paste-driven analysis), leaving AWS/SSH out. If some form of this must land sooner, at minimum the `AWS` connection and the instruction lines that promise CloudWatch/DynamoDB/SQS access have to go, and the safety copy needs to stop asserting enforcement that does not exist.

## Findings

### Critical

**C1 — This PR breaks an existing test; CI should be red.**
`tests/templates/test_manager.py::TestTemplateManager::test_builtin_template_connections_resolve_to_a_registered_mcp_app` exists precisely to catch templates whose `connections[].name` does not resolve to a registered MCP app. It fails on this PR's head with `AssertionError: assert not ['operations-devops-ai-agent.yaml: AWS']`. Root cause: the `- name: AWS` entry at line 37. Fix: remove the `AWS` connection (and the instructions that depend on it) until an AWS MCP app is actually registered.

**C2 — The headline advertised feature (webhook/alert/schedule triggering) is the one path guaranteed to fail; plain chat instead fails silently and will hallucinate.**
Trigger-sourced runs use a STRICT MCP failure policy (`src/xagent/web/api/chat.py:490-494`). A scoped MCP category with no registered app is treated as "requested but unavailable" (`src/xagent/core/tools/adapters/vibe/mcp_tools.py:218-226`, `no_tools_returned` at line 125), and `enforce_mcp_failure_policy` (`src/xagent/core/tools/adapters/vibe/config.py:176-183`) raises `RequiredMCPUnavailableError` under STRICT. So **every trigger-fired run of this agent fails outright** — that is the "wakes on alert/webhook/schedule" behavior this PR advertises as feature #1.

In plain chat the same condition degrades silently: no warning is logged on this path, only an internal `__audit_only__` `mcp_load_summary` trace event (`src/xagent/web/tools/config.py:1782-1801`). The agent will therefore fabricate CloudWatch/DynamoDB/SQS findings rather than surface the missing access — directly violating its own instruction at line 73 ("report the access failure itself rather than silently skipping that data source").

### Major

**M1 — The agent-builder Configure step is permanently marked incomplete for this template.**
In `frontend/src/components/build/agent-builder.tsx`, `requiredMcpServers` is derived from template `connections` (lines 966-969); `templateMissingMcp` (lines 1736-1745) is true unless every required server resolves to a connected server or a registered+connected official app. `AWS` matches neither, so it is always true here, and `configStepCompleted` (lines 1749-1758) is consequently always false, with no skip or override path. Caveat, verified: this does **not** block creating/saving the agent (the create button's disabled state does not consult `configStepCompleted`) — it is a persistent "incomplete" badge, not a hard block. Hence major rather than critical.

**M2 — Internal inconsistency: `sample_prompts` still describe the old paste-only workflow.**
Lines 17-35 (unchanged by this PR) frame the user as manually supplying data ("Here are the logs and command output from the incident…", "I'm pasting a monitoring export below…"), which contradicts the new active-pull framing at lines 6, 11, and 53. The gallery card and the actual chat starter prompts now tell users two different stories about how this agent works. Whichever direction wins, both need to agree.

**M3 — The "read-only by default, human-approval-gated" safety claim is prompt-only; nothing enforces it.**
Lines 6-7, 11, 53, 59, and 72 promise read-only SSH diagnostics with Slack-approval gating for mutating actions. Verified: `SshExecuteTool` (`src/xagent/core/tools/adapters/vibe/ssh_tools.py:61,175-186`) passes `command: str` verbatim to `SshExecutor.execute` (`src/xagent/core/ssh/executor.py:92-156`) and on to `asyncssh`'s `create_process` (`src/xagent/core/ssh/runner.py:172-213,312`). There is no allowlist, denylist, or command classification anywhere on that path — only a capability-membership check (`"execute"` in granted capabilities) and an egress policy, neither of which inspects command content. The `ssh` category additionally grants `SshUploadTool`/`SshDownloadTool` (`ssh_tools.py:197-264`), i.e. mutating file transfer, gated the same capability-only way. `ResolvedSshTarget.approval_policy` (`src/xagent/core/ssh/types.py:11,78`) is never read or branched on outside test fixtures — dead configuration.

This also subsumes the bot's earlier contradiction finding (see Prior-review verification below): lines 48 and 72 state absolutely that the agent recommends but never applies fixes, while lines 55 and 59 say to wait for Slack approval "before any change is made" / never mutate "without explicit human approval in Slack first" — which implies mutation *is* possible with approval. Both readings are still present. Fix: pick one model. If the agent is genuinely advisory-only, drop the approval-gating language entirely and say so unconditionally. If approval-gated mutation is the intent, it needs real enforcement in `SshExecutor` (command classification honoring `approval_policy`) before the template may claim it.

**M4 — The `ssh` category will very likely resolve to zero tools for most deployments, and this is the only built-in template that uses it.**
`src/xagent/web/services/ssh_runtime.py` only returns SSH tools when an out-of-tree provider hook is installed (per its docstring the DB-backed provider is in closed-source xagent-cloud; no `SshTarget`/`SshCredential` model exists under `src/xagent/web/models/` in this tree). `create_ssh_tools` (`ssh_tools.py:369-372`) returns `[]` with no provider, also returns `[]` unless the agent has at least one bound SSH target (`ssh_tools.py:404-406`), and is hard-disabled when `SANDBOX_IMPLEMENTATION == "boxlite"` (`ssh_tools.py:427-429`). This is the only file under `src/xagent/templates/built_in/` listing `ssh` in `tool_categories`, so it sets an unvetted precedent: self-hosted and standard users who pick this template get an agent whose own instructions reference SSH diagnostics while having literally zero SSH tools. If this template is cloud-only, it needs an explicit availability gate rather than shipping as a plain built-in.

### Minor

**m1 — `setup_time: 10 min` is an unexplained outlier.**
Across the 18 built-in templates, every other one is `2 min setup` (1) or `5 min setup` (16). Plausibly justified given the extra connections, but worth a one-line rationale, since other multi-connection templates still use 5 min.

## Prior-review verification (gemini-code-assist passes)

- **"Corrupted connection name" (stray artifact text in the `connections` block): FIXED.** Line 37 now reads cleanly `- name: AWS`, with no artifact text.
- **"Read-only vs. approval-gated contradiction" (lines 48/72 vs. 55/59): NOT FIXED.** Both phrasings are still present verbatim in the current file. Folded into M3 above, since the root cause is the same: the instructions describe an approval-gated mutation capability the platform neither implements nor enforces.

## Test results

Independently executed `tests/templates/test_manager.py`:

- Base commit (`main`, pre-PR): **18 passed**
- PR head: **1 failed, 17 passed**

```
FAILED tests/templates/test_manager.py::TestTemplateManager::
       test_builtin_template_connections_resolve_to_a_registered_mcp_app
AssertionError: assert not ['operations-devops-ai-agent.yaml: AWS']
```

This is a pre-existing guard test (lines 422-456) doing exactly its job. Fixing C1 fixes the test.

## Simplification lens

Lean already. No over-engineering findings on this diff.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,71 +3,78 @@ name: DevOps AI Agent
 category: Operations
 featured: false
 descriptions:
-  en: Triages a diagnostic bundle you paste or upload (logs, command output, monitoring exports) and reports a root-cause hypothesis and recommended fix directly in chat.
-  zh: 对您粘贴或上传的诊断信息（日志、命令输出、监控导出）进行分诊，并直接在对话中汇报根因假设与修复建议。
+  en: Monitors your AWS environment, triages alerts and incidents, runs read-only diagnostics over SSH, and posts findings plus a recommended fix to Slack for a human to action — it never makes changes itself.
```
rogercloud: F1 (minor-to-moderate, non-blocking): this states SSH diagnostics as an unconditional capability, but the instructions correctly hedge it — line 53 says "where a target is available" and line 73 says "SSH diagnostics depend on the deployment having a bound target". SSH requires an out-of-tree provider hook (`set_ssh_target_provider_hook` in `src/xagent/web/services/ssh_runtime.py` is never called anywhere under `src/`) plus manually bound targets, and `create_ssh_tools` returns `[]` without them — it's also hard-disabled under the boxlite sandbox. Most deployments picking this template from the gallery will only discover at runtime that this third of the description is inert. Suggest matching the hedge already present in the instructions, e.g. "…runs read-only diagnostics over SSH where a target is configured…".

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,71 +3,78 @@ name: DevOps AI Agent
...
+  zh: 监控 AWS 环境，对告警与事件进行分诊，通过 SSH 执行只读诊断，并将排查结果和修复建议发布到 Slack 供人工处理——该 Agent 本身不会执行任何变更操作。
 features:
   en:
-  - Reads logs, command output or monitoring exports you paste or upload — no live server connection required
-  - Correlates findings into a root-cause hypothesis and a recommended fix
-  - Reports directly in chat — no external monitoring or messaging integration required
+  - Wakes on an alert / webhook, a schedule, or a direct chat request
+  - Investigates via CloudWatch, DynamoDB and SQS, plus read-only SSH diagnostics
```
rogercloud: Same as the note on line 6 (non-blocking): "plus read-only SSH diagnostics" reads as unconditional, while the instructions (lines 53, 73) correctly treat SSH as conditional on a bound target that most deployments won't have. Suggest hedging this bullet to match.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,71 +3,78 @@ name: DevOps AI Agent
...
 
-    - **Work only from provided material:** Every claim must be traceable to something the user pasted or uploaded. If a detail you need wasn't provided, ask for it rather than guessing.
-    - **No secrets in output:** Never repeat credentials, tokens, keys, or connection strings that appear in the provided logs — redact them.
-    - **Escalate uncertainty:** If the root cause isn't clear from what was provided, say so and ask for the next piece of diagnostic output rather than guessing at a fix.
-    - **Severity discipline:** Only call something high-severity if it maps to defined criteria (customer-facing impact, data risk, SLA breach) — don't inflate severity.
+    - **Diagnostics only, always:** Never run a mutating command, deploy, restart, or scale a resource — under any circumstance, including after Slack approval. Implementing the fix is always a human's job.
+    - **No secrets in output:** Never print or forward credentials, tokens, keys, or connection strings, even if they appear in logs — redact them.
+    - **Known-noise catalog:** Suppress or de-prioritize alerts matching known benign patterns; note when you've done so instead of silently dropping them.
```
rogercloud: F2 (moderate, non-blocking): the "known-noise catalog" is referenced six times across this file (lines 12, 16, 24, 33, 54, 61) but is never defined or sourced. `skills: []` (line 75) is empty, so no skill supplies it. `knowledge` is in `tool_categories` (line 79), so a KB lookup is possible if the user populates one — but nothing tells the agent what document to look for or what to do when it's missing.

Because this template's output feeds a human incident decision, this is a correctness risk rather than a wording nit: the agent will either silently skip the noise-filtering step it advertises, or invent a plausible-sounding "known-noise" ruleset from general knowledge instead of the user's real environment — which can mask or misclassify an actual incident.

Suggest naming the expected artifact explicitly, e.g. "search the knowledge base for a document describing known-noise patterns for this environment; if none exists, state that no catalog was found rather than assuming which alerts are benign." Alternatively, soften the feature copy until that grounding exists.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,71 +3,78 @@ name: DevOps AI Agent
...
     # Responsibilities
 
-    1. **Get the material** — If the user hasn't already pasted or attached diagnostic output, ask for the specific logs or command output relevant to the problem.
-    2. **Correlate** — Cross-reference what was provided; rule out likely red herrings before settling on a hypothesis.
-    3. **Report** — Post a summary, the evidence, a root-cause hypothesis, a recommended fix, and a severity assessment, directly in this chat.
-    4. **Recommend, don't apply** — You have no way to run anything on the user's systems. Always phrase the fix as a recommendation for the user (or their own tooling) to carry out.
+    1. **Receive the trigger** — An alert / webhook, a scheduled check, or a direct chat request.
+    2. **Investigate, read-only** — Pull relevant CloudWatch metrics/logs, DynamoDB and SQS state, and run read-only SSH diagnostics where a target is available.
+    3. **Triage** — Correlate signals, filter out entries matching the known-noise catalog, and identify the most likely root cause.
+    4. **Report & recommend** — Post an incident summary, evidence, root-cause hypothesis, recommended fix and severity to Slack for a human to act on.
```
rogercloud: F3 (moderate, UX, non-blocking): lines 10 and 52 both advertise "a direct chat request" as a trigger mode, but this responsibility (and the reminder on line 71) route the entire output — summary, evidence, root cause, fix, severity — to Slack, with no chat-reply branch described anywhere. The second sample prompt (lines 23-26) doesn't mention Slack at all and reads exactly like a request expecting an in-conversation answer.

So a user who triggers this agent from chat, exactly as the template invites them to, will plausibly get back "Posted findings to #incidents" and nothing substantive where they're actually looking.

Suggest adding an explicit branch — e.g. "if the trigger was a direct chat request, reply with the full report in the chat as well as posting to Slack" — or making the trigger copy state that output always goes to Slack regardless of source.

src/xagent/templates/built_in/operations-devops-ai-agent.yaml
```diff
@@ -3,71 +3,78 @@ name: DevOps AI Agent
...
+    - Slack message structure: incident summary · evidence · root-cause hypothesis · recommended fix · severity.
+    - Never apply a fix yourself, even a diagnostic one flagged as low-risk — always end with a recommendation for a human to carry out.
+    - SSH diagnostics depend on the deployment having a bound target; if SSH, CloudWatch, DynamoDB, or SQS access is unavailable or fails, say so explicitly rather than silently skipping that data source.
   execution_mode: balanced
   skills: []
   tool_categories:
   - basic
   - file
   - knowledge
+  - ssh
```
rogercloud, line 80: F4 (minor, pre-existing frontend gap this template exposes, non-blocking): hardcoding `ssh` here makes it a sticky, invisible selected category in the agent builder.

Normally `ssh` only enters `tool_categories` once the user has real SSH bindings (`hasSshBindings` in `frontend/src/components/build/agent-builder.tsx`, ~lines 280-284, 1116-1118, 1431-1433 — the save path only ever adds it, never removes it). But template-apply (lines 931-932) copies `tool_categories` almost verbatim (`isAssignableToolCategory` strips only `agent`/`other`), so `ssh` lands in `selectedToolCategories` without `hasSshBindings` ever being set.

Two confirmed effects: (1) the picker's options come from `/api/tools`, which won't offer `ssh` when no SSH tools exist, so the multi-select renders an orphaned unlocalized `"ssh"` chip via the raw-label fallback; (2) on reload, `setHasSshBindings(rawToolCategories.includes("ssh"))` (line 860) flips the flag true purely because the persisted config contains `ssh`, regardless of whether a target was ever bound.

The bindings panel itself (`agent-ssh-bindings.tsx:190`) still correctly reports "No SSH targets bound to this agent", so the most user-visible surface is honest. Not asking this PR to fix it — flagging because this is the only built-in template that triggers it.

rogercloud (approved):
## Follow-up review (head `b19d24b9`)

Two things landed since my previous `REQUEST_CHANGES`. First, the template itself was rewritten: the internal contradictions between the "read-only" framing and the "wait for approval before any change" language are gone, `sample_prompts` were re-pointed at the active-investigation model the descriptions promise, and `setup_time` went back to `5 min setup`. Second — and independently of this PR — `main` merged #1068, which adds a real AWS MCP connector. That removes the core objection from last round: `mcp:AWS` is no longer an unbacked connection name.

### Updated approach verdict: ACCEPTABLE-WITH-RESERVATIONS

Upgraded from the prior **WRONG DIRECTION**. With AWS and Slack both genuinely backed by registered MCP apps, and the instructions now describing one coherent model (strictly advisory, never mutates), the design holds together. My remaining reservation is narrower: the SSH third of this template still isn't practically usable in most deployments — especially self-hosted / open-source ones — and the gallery-facing copy doesn't yet reflect the hedge the instructions themselves correctly apply.

### Status of prior findings

**C1 (critical — test failure): FIXED.**
`tests/templates/test_manager.py::TestTemplateManager::test_builtin_template_connections_resolve_to_a_registered_mcp_app` now passes. AWS is registered in `src/xagent/web/builtin_mcp_registry.py:499-501` as `{"app_id": "aws", "name": "AWS"}`, which matches the template's `connections: - name: AWS` at `src/xagent/templates/built_in/operations-devops-ai-agent.yaml:37` exactly.

**C2 (critical — trigger-fired runs guaranteed to fail / chat silently hallucinating): FIXED.**
`src/xagent/web/tools/mcp/aws.py:322-751` registers a real `FastMCP` server exposing nine concrete tools: `aws_get_caller_identity`, `aws_cloudwatch_describe_alarms`, `aws_cloudwatch_get_metric_data`, `aws_cloudwatch_describe_log_groups`, `aws_cloudwatch_filter_log_events`, `aws_dynamodb_list_tables`, `aws_dynamodb_describe_table`, `aws_sqs_list_queues`, `aws_sqs_get_queue_attributes`. I checked for mutating calls (Put/Send/Delete/Update) and found none — the template's "read-only" claim for AWS is genuinely true at the tool layer, not just aspirational. `mcp:AWS` now resolves to real tools, so neither the STRICT-policy trigger-failure path nor the silent-degradation path applies to AWS anymore. Runs still require the user to supply their own AWS credentials at connect time, but that's the same by-design behavior as every other MCP-gated template, not a defect.

**M1 (major — Configure step permanently stuck): FIXED, with an expected nuance.**
`templateMissingMcp` in `frontend/src/components/build/agent-builder.tsx` will now find a matching official app for AWS. The Configure step still reads "incomplete" until the user actually connects their AWS credentials — but that is the correct, shared behavior of Slack, Gmail and every other MCP-gated template. The old bug (permanently unsatisfiable regardless of user action, because no matching app existed at all) is gone.

**M2 (major — `sample_prompts` contradicted the framing): FIXED.**
Lines 17-35 were rewritten. Both English prompts now describe pulling data directly, e.g. line 20 "Pull the CloudWatch metrics and logs for the last hour, run read-only diagnostics on the affected hosts…" and line 24 "Check the SQS queues and DynamoDB tables for this service for anomalies…". These are consistent with `descriptions` and `features`.

**M3 (major — read-only vs. conditional-mutation contradiction; also gemini-code-assist's open finding): FIXED.**
Line 59 now reads "**Diagnostics only, always:** Never run a mutating command, deploy, restart, or scale a resource — under any circumstance, including after Slack approval. Implementing the fix is always a human's job." Line 72 reinforces it: "Never apply a fix yourself, even a diagnostic one flagged as low-risk — always end with a recommendation for a human to carry out." The old "Wait for approval before any change is made" language, which implied conditional mutation, is gone entirely. The loophole is closed.

**M4 (major — `ssh` tool_category likely yields zero tools): STILL VALID, but downgraded to a documented caveat.**
Re-verified against current code: `src/xagent/web/services/ssh_runtime.py` still has no in-tree SSH target provider — `set_ssh_target_provider_hook` is called from nowhere under `src/`, only from tests. `create_ssh_tools` still returns `[]` when there is no provider or no bound target, and remains hard-disabled under `SANDBOX_IMPLEMENTATION == "boxlite"`. This is still the only built-in template listing `ssh`. It is a real, still-open gap — but it is a pre-existing platform limitation, not a regression this diff introduces, so I'm no longer treating it as blocking. It does, however, pair with F1 below: the copy shouldn't promise unconditionally what the platform can't unconditionally deliver.

**m1 (minor — `setup_time: 10 min` outlier): FIXED.** Line 42 is back to `5 min setup`, matching 16 of the other 17 built-in templates.

**gemini-code-assist (corrupted "AWS" connection name artifact): still FIXED.**

### New findings (non-blocking)

**F1 (minor-to-moderate) — gallery copy over-promises SSH, while the instructions correctly hedge it.**
`descriptions.en` (line 6, "runs read-only diagnostics over SSH") and `features.en` (line 11, "plus read-only SSH diagnostics") state SSH as an unconditional capability. The instructions are more honest: line 53 says "run read-only SSH diagnostics **where a target is available**", and line 73 says "SSH diagnostics **depend on the deployment having a bound target**; if … access is unavailable or fails, say so explicitly." Given SSH requires an out-of-tree provider hook plus manually bound targets (see M4), users who pick this template based on the gallery card will only discover at runtime that a third of the advertised capability is inert. Suggested fix: soften lines 6 and 11 to match the hedge already present in the instructions — e.g. "…runs read-only diagnostics over SSH where a target is configured…".

**F2 (moderate) — the "known-noise catalog" is referenced five times but never defined or sourced.**
It appears at lines 12, 16, 24, 33, 54 and 61 (line 61: "**Known-noise catalog:** Suppress or de-prioritize alerts matching known benign patterns…"). `skills: []` (line 75) is empty, so no skill supplies it. `tool_categories` includes `knowledge` (line 79), so a knowledge-base lookup is *possible* if the user populates one — but nothing in the instructions tells the agent where to look, what artifact to look for, or what to do when it isn't there. Because the template's whole purpose is triage feeding a human decision, this is a correctness risk rather than a documentation nit: the agent will either silently skip the noise-filtering step it promises, or fabricate a plausible-sounding "known-noise" ruleset from general knowledge instead of the user's actual environment — which can mask or misclassify a real incident. Suggested fix: name the expected knowledge-base artifact explicitly ("search the knowledge base for a document describing known-noise patterns; if none exists, say so rather than assuming"), or soften the copy until that grounding exists.

**F3 (moderate, UX inconsistency) — "direct chat request" is advertised as a trigger, but the only described output path is Slack.**
Lines 10 and 52 both list "a direct chat request" as one of three trigger modes. But Responsibility 4 (line 55) and the Reminders (line 71) both mandate posting the summary, evidence, root cause, fix and severity **to Slack**, and nothing anywhere describes a chat-reply variant. The second sample prompt (lines 23-26, "…report anything that doesn't match a known benign pattern") doesn't mention Slack at all — it reads exactly like a request expecting an in-conversation answer. A user who triggers this agent from chat, precisely as the template invites, will plausibly get back "Posted findings to #incidents" and nothing substantive in the conversation they're looking at. Suggested fix: add an explicit branch ("if triggered by a direct chat request, also reply with the full summary in the chat"), or make the copy state that output always goes to Slack regardless of trigger source.

**F4 (minor) — hardcoding `ssh` in `tool_categories` makes it a sticky, invisible selected category in the builder.**
Normally `ssh` enters an agent's `tool_categories` only once the user has real SSH bindings (`hasSshBindings` in `frontend/src/components/build/agent-builder.tsx`, around lines 280-284, 1116-1118 and 1431-1433; the save path only ever *adds* `ssh`, never removes it). But template-apply (lines 931-932) copies `tool_categories` almost verbatim — `isAssignableToolCategory` only strips `agent`/`other` — so `ssh` lands in `selectedToolCategories` without `hasSshBindings` ever becoming true. Two confirmed consequences: (1) the category picker's options come from `/api/tools`, which won't offer `ssh` when no SSH tools are available, so the multi-select renders an orphaned, unlocalized `"ssh"` chip via the raw-label fallback; (2) on reload, `setHasSshBindings(rawToolCategories.includes("ssh"))` (line 860) flips the flag to `true` purely because the persisted config contains `ssh`, regardless of whether any target was ever bound. Mitigating: the bindings panel itself (`agent-ssh-bindings.tsx:190`) still correctly reports "No SSH targets bound to this agent", so the most visible surface doesn't lie. This is arguably a pre-existing frontend gap that any template could trigger; this is simply the only one that does. Flagging it rather than asking this PR to fix it.

### Test results

Re-ran `tests/templates/test_manager.py` against the current head: **18 passed** (previously 1 failed, 17 passed). The connection-registry test that blocked the last review now passes.

### Overall recommendation

Every critical and major blocker from the previous review is resolved, and the one still-valid item (M4) is a pre-existing platform limitation rather than something this diff introduces. What remains — F1, F2, F3 — are copy and instruction-wording issues: quick edits to this one YAML file, no architecture change, nothing test-breaking. F4 needs no action here.

Approving on that basis. F1-F3 are worth a quick follow-up pass (touching lines 6, 11, 55 and 61) but are not blockers for this PR.

PR #1098, opened 2026-07-31
feat: add computer use foundation

qinxuye:
## Summary

- add provider-neutral Computer Use action, observation, target, viewport, and policy contracts
- add a serialized environment lifecycle with stale frame, session, and element validation
- persist screenshot observations as bounded workspace-internal FileRefs without exposing scratch files as user uploads

## Scope

This is the foundation-only slice of the Computer Use work. It intentionally does not add a browser backend, ComputerTool/ReAct integration, Local browser, browser extension relay, or desktop companion.

## Validation

- `PYTHONPATH="$PWD/src" python -m pytest -q tests/core/computer tests/core/test_file_ref.py tests/core/test_workspace_durable_registration.py` (115 passed)
- `pre-commit run --files <changed files>` (all hooks passed)

PR #1104, opened 2026-08-01
feat: add browser computer use loop

qinxuye:
## Scope

This PR is the **ephemeral Playwright browser loop** built on top of the provider-neutral Computer Use foundation merged in #1098. It is deliberately a focused implementation slice, not the complete Browser Use or Computer Use product.

## What this PR does

- implements `BrowserComputerEnvironment` with a new isolated Playwright browser session
- captures each viewport as an internal screenshot FileRef and returns the current URL, title, viewport, and bounded DOM hints
- supports one action per observed frame, followed by a mandatory fresh observation
- preserves the foundation stale-frame and task/session isolation guarantees
- exposes one `computer` tool through the existing browser tool category
- connects the tool to the existing ReAct loop and task/step-scoped browser lifecycle
- supersedes stale browser observations in the live model prompt while retaining their durable FileRefs in execution state
- redacts sensitive input values from DOM hints

## Explicitly out of scope

This PR does **not** implement or change:

- persistent Playwright profiles or reused login state
- control of an existing local Chrome window through `cua-driver`
- Chrome extension takeover of a user-authorized signed-in tab
- WebSocket or Redis browser relay, pairing, or distributed routing
- macOS desktop relay, window control, or whole-computer control
- product UI, settings, target selection, pairing flows, or confirmation UX
- native OpenAI, Claude, or Gemini computer-use model protocols
- router/model-selection policy beyond exposing the ordinary `computer` tool
- iframe-wide or closed-shadow-root DOM extraction; screenshots remain authoritative when structural hints are incomplete

Those capabilities should be reviewed in their own follow-up PRs with their own security and lifecycle boundaries. They should not be required to merge this foundational browser-loop slice.

## Review focus

Review in this PR should stay focused on:

1. conformance with the #1098 Computer Use schema and environment contracts
2. stale-frame enforcement and one-action-per-observation behavior
3. task/step session isolation and deterministic teardown
4. Playwright action and screenshot correctness
5. safe FileRef/context handling and bounded prompt growth
6. regressions to the existing browser tools and ReAct loop

Suggestions that require persistent profiles, local-browser ownership, relays, desktop control, product authorization, or new provider-specific protocols are intentionally follow-up work rather than blockers for this PR.

## Validation

- `pre-commit run --files ...`
- `pytest -q tests/core/computer tests/core/agent/test_browser_tools.py`
- `pytest -q tests/core/test_context_ref.py tests/core/test_context_materializer.py tests/core/agent/test_context.py tests/core/agent/test_react.py`
- real headless Chromium smoke: screenshot, navigate to Example Domain, capture a fresh observation, and teardown

### DIFF
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -12,6 +12,7 @@ from ...context_ref import (
     ContextReference,
     normalize_context_references,
     split_tool_result_context_references,
+    split_tool_result_supersedes_scope,
 )
 from ...file_ref import FILE_REF_MODEL_INSTRUCTIONS
 from ...tools.artifacts import (
@@ -166,7 +167,10 @@ class ExecutionContext:
         *,
         context_refs: Any = (),
     ) -> Message:
-        public_result, embedded_refs = split_tool_result_context_references(result)
+        public_result, supersedes_scope = split_tool_result_supersedes_scope(result)
+        public_result, embedded_refs = split_tool_result_context_references(
+            public_result
+        )
         explicit_refs = normalize_context_references(context_refs)
         all_context_refs = []
         seen_context_refs: set[str] = set()
@@ -181,7 +185,7 @@ class ExecutionContext:
             tool_name, public_result
         )
         content = self._format_tool_result(tool_name, context_result)
-        metadata = {
+        metadata: dict[str, Any] = {
             "tool_name": tool_name,
             "raw_result": context_result,
             "workspace_id": self.workspace_id,
@@ -189,12 +193,58 @@ class ExecutionContext:
             "cwd": self.cwd,
             "memory_session_id": self.memory_session_id,
         }
+        if supersedes_scope:
+            metadata["supersedes_scope"] = supersedes_scope
+            self._compact_superseded_tool_messages(supersedes_scope)
         return self.add_message(
             "tool",
             content,
             tool_call_id=tool_call_id,
             metadata=metadata,
             context_refs=tuple(all_context_refs),
+        )
+
+    def _compact_superseded_tool_messages(self, scope: str) -> None:
+        """Keep stale observations durable while removing them from live prompts."""
+
+        for index, message in enumerate(self.messages):
+            metadata = message.metadata or {}
+            if metadata.get("supersedes_scope") != scope:
+                continue
+            if metadata.get("superseded"):
+                continue
+            self.messages[index] = replace(
+                message,
+                content=self._superseded_tool_summary(message),
+                metadata={
+                    **metadata,
+                    "superseded": True,
+                    "raw_result": None,
+                },
+            )
+
+    @staticmethod
+    def _superseded_tool_summary(message: Message) -> str:
+        metadata = message.metadata or {}
+        tool_name = str(metadata.get("tool_name") or "tool")
+        raw_result = metadata.get("raw_result")
+        details: list[str] = []
+        if isinstance(raw_result, dict):
+            observation = raw_result.get("observation")
+            for key in ("frame_id", "active_url", "title"):
+                value = raw_result.get(key)
+                if not value and isinstance(observation, dict):
+                    value = observation.get(key)
+                if value:
+                    details.append(f"{key}={value}")
+        if message.context_refs:
+            details.append(
+                "file_id=" + ",".join(ref.file_id for ref in message.context_refs)
+            )
+        suffix = f" ({', '.join(details)})" if details else ""
+        return (
+            f"[{tool_name} result superseded by a later observation{suffix}. "
+            "The page state it described no longer applies.]"
         )
 
     def attach_workspace(
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -219,7 +219,7 @@ class ExecutionContext:
                 metadata={
                     **metadata,
                     "superseded": True,
-                    "raw_result": None,
+                    "raw_result": {"success": True, "superseded": True},
                 },
             )
 
@@ -1218,7 +1218,9 @@ class ExecutionContext:
                 )
             chunks.append(message.content)
             if message.context_refs:
-                chunks.append(message.context_refs_text())
+                context_refs_text = message.context_refs_text()
+                if context_refs_text:
+                    chunks.append(context_refs_text)
         return "\n".join(chunks)
 
     def _compact_response_text(self, response: Any) -> str:
--- a/src/xagent/core/computer/browser.py
+++ b/src/xagent/core/computer/browser.py
@@ -6,8 +6,14 @@ from typing import Any, cast
 from urllib.parse import urlsplit
 from uuid import uuid4
 
+from pydantic import TypeAdapter, ValidationError
+
 from ..tools.core.browser_use import BrowserSessionManager, get_browser_manager
-from .environment import ComputerEnvironment, ComputerTargetNotFoundError
+from .environment import (
+    ComputerEnvironment,
+    ComputerFrameMismatchError,
+    ComputerTargetNotFoundError,
+)
 from .schema import (
     ELEMENT_EXTRACTION_FAILED_KEY,
     ELEMENT_EXTRACTION_INCOMPLETE_KEY,
@@ -21,11 +27,14 @@ from .schema import (
     ComputerEnvironmentType,
     ComputerObservation,
     NormalizedPoint,
+    ObservedUrl,
     Viewport,
 )
 from .store import ObservationStore
 
 logger = logging.getLogger(__name__)
+
+_OBSERVED_URL_ADAPTER = TypeAdapter(ObservedUrl)
 
 _INTERACTIVE_ELEMENTS_SCRIPT = """
 (limit) => {
@@ -77,8 +86,7 @@ _INTERACTIVE_ELEMENTS_SCRIPT = """
       autocomplete === "webauthn" ||
       autocomplete.startsWith("cc-")
     );
-    const safeValue = sensitive ? "" : String(node.value || "");
-    const text = String(node.innerText || safeValue).trim().slice(0, 240);
+    const text = String(node.innerText || "").trim().slice(0, 240);
     const label = String(
       node.getAttribute("aria-label") ||
       node.getAttribute("title") ||
@@ -133,12 +141,25 @@ class BrowserComputerEnvironment(ComputerEnvironment):
         self.headless = headless
         self.viewport_width = viewport_width
         self.viewport_height = viewport_height
+        self._manager_session_id = f"computer:{self.session_id}"
+        self._browser_session: Any | None = None
 
-    async def _get_page(self) -> Any:
+    async def _get_page(self, *, reject_recreated_session: bool) -> Any:
         session = await self.manager.get_or_create(
-            self.session_id,
+            self._manager_session_id,
             headless=self.headless,
         )
+        recreated = (
+            self._browser_session is not None and session is not self._browser_session
+        )
+        self._browser_session = session
+        if recreated:
+            self._invalidate_observation()
+            if reject_recreated_session:
+                raise ComputerFrameMismatchError(
+                    "browser session changed after the expected frame was captured; "
+                    "request a fresh screenshot before another action"
+                )
         page = await session.get_page()
         requested = {
             "width": self.viewport_width,
@@ -149,13 +170,16 @@ class BrowserComputerEnvironment(ComputerEnvironment):
         return page
 
     async def _close(self) -> None:
-        await self.manager.close(self.session_id)
+        await self.manager.close(self._manager_session_id)
+        self._browser_session = None
 
     async def _observe(self) -> ComputerObservation:
-        return await self._capture_observation(await self._get_page())
+        return await self._capture_observation(
+            await self._get_page(reject_recreated_session=False)
+        )
 
     async def _execute(self, batch: ComputerActionBatch) -> ComputerObservation:
-        page = await self._get_page()
+        page = await self._get_page(reject_recreated_session=True)
         action = batch.actions[0]
         await self._execute_action(page, action)
         if action.type not in {
@@ -205,7 +229,13 @@ class BrowserComputerEnvironment(ComputerEnvironment):
             # screenshot remains authoritative for cross-origin child frames.
             metadata[ELEMENT_EXTRACTION_INCOMPLETE_KEY] = True
 
-        active_url = str(getattr(page, "url", "") or "").strip() or None
+        raw_active_url = str(getattr(page, "url", "") or "").strip() or None
+        active_url: str | None = None
+        if raw_active_url is not None:
+            try:
+                active_url = _OBSERVED_URL_ADAPTER.validate_python(raw_active_url)
+            except ValidationError:
+                metadata["active_url_unavailable"] = True
         return ComputerObservation(
             session_id=self.session_id,
             frame_id=frame_id,
@@ -296,6 +326,18 @@ class BrowserComputerEnvironment(ComputerEnvironment):
         if action.type is ComputerActionType.REPLACE_TEXT:
             x, y = self._target_pixels(action)
             await page.mouse.click(x, y)
+            editable = await page.evaluate(
+                """() => {
+                  const node = document.activeElement;
+                  if (!node) return false;
+                  if (node.isContentEditable) return true;
+                  const tag = String(node.tagName || "").toLowerCase();
+                  return (tag === "input" || tag === "textarea") &&
+                    !node.disabled && !node.readOnly;
+                }"""
+            )
+            if not editable:
+                raise ValueError("replace_text target is not an editable element")
             modifier = "Meta" if sys.platform == "darwin" else "Control"
             await page.keyboard.press(f"{modifier}+A")
             await page.keyboard.insert_text(action.text or "")
@@ -404,4 +446,10 @@ class BrowserComputerEnvironment(ComputerEnvironment):
             "SPACE": "Space",
             "TAB": "Tab",
         }
-        return "+".join(aliases.get(key.upper(), key) for key in keys)
+        normalized = [aliases.get(key.upper(), key) for key in keys]
+        modifiers = {"Alt", "Control", "Meta", "Shift"}
+        if len(normalized) > 1 and any(key not in modifiers for key in normalized[:-1]):
+            raise ValueError(
+                "keypress accepts one key chord; send sequential keys as separate actions"
+            )
+        return "+".join(normalized)
--- a/src/xagent/core/computer/browser.py
+++ b/src/xagent/core/computer/browser.py
@@ -36,6 +36,28 @@ logger = logging.getLogger(__name__)
 
 _OBSERVED_URL_ADAPTER = TypeAdapter(ObservedUrl)
 
+_EDITABLE_ACTIVE_ELEMENT_SCRIPT = """() => {
+  let currentDocument = document;
+  const visited = new Set();
+  while (currentDocument && !visited.has(currentDocument)) {
+    visited.add(currentDocument);
+    const node = currentDocument.activeElement;
+    if (!node) return false;
+    if (node.isContentEditable) return true;
+    const tag = String(node.tagName || "").toLowerCase();
+    if (tag === "input" || tag === "textarea") {
+      return !node.disabled && !node.readOnly;
+    }
+    if (tag !== "iframe" && tag !== "frame") return false;
+    try {
+      currentDocument = node.contentDocument;
+    } catch (_) {
+      return false;
+    }
+  }
+  return false;
+}"""
+
 _INTERACTIVE_ELEMENTS_SCRIPT = """
 (limit) => {
   const selector = [
@@ -141,7 +163,7 @@ class BrowserComputerEnvironment(ComputerEnvironment):
         self.headless = headless
         self.viewport_width = viewport_width
         self.viewport_height = viewport_height
-        self._manager_session_id = f"computer:{self.session_id}"
+        self._manager_session_id = f"{self.session_id}:computer"
         self._browser_session: Any | None = None
 
     async def _get_page(self, *, reject_recreated_session: bool) -> Any:
@@ -326,16 +348,7 @@ class BrowserComputerEnvironment(ComputerEnvironment):
         if action.type is ComputerActionType.REPLACE_TEXT:
             x, y = self._target_pixels(action)
             await page.mouse.click(x, y)
-            editable = await page.evaluate(
-                """() => {
-                  const node = document.activeElement;
-                  if (!node) return false;
-                  if (node.isContentEditable) return true;
-                  const tag = String(node.tagName || "").toLowerCase();
-                  return (tag === "input" || tag === "textarea") &&
-                    !node.disabled && !node.readOnly;
-                }"""
-            )
+            editable = await page.evaluate(_EDITABLE_ACTIVE_ELEMENT_SCRIPT)
             if not editable:
                 raise ValueError("replace_text target is not an editable element")
             modifier = "Meta" if sys.platform == "darwin" else "Control"
--- a/src/xagent/core/computer/browser.py
+++ b/src/xagent/core/computer/browser.py
@@ -1,0 +1,398 @@
+from __future__ import annotations
+
+import logging
+import sys
+from typing import Any, cast
+from urllib.parse import urlsplit
+from uuid import uuid4
+
+from ..tools.core.browser_use import BrowserSessionManager, get_browser_manager
+from .environment import ComputerEnvironment
+from .schema import (
+    ELEMENT_EXTRACTION_FAILED_KEY,
+    ELEMENT_EXTRACTION_INCOMPLETE_KEY,
+    ELEMENTS_TRUNCATED_KEY,
+    MAX_OBSERVATION_ELEMENTS,
+    ComputerAction,
+    ComputerActionBatch,
+    ComputerActionType,
+    ComputerElement,
+    ComputerElementSource,
+    ComputerEnvironmentType,
+    ComputerObservation,
+    NormalizedPoint,
+    Viewport,
+)
+from .store import ObservationStore
+
+logger = logging.getLogger(__name__)
+
+_INTERACTIVE_ELEMENTS_SCRIPT = """
+(limit) => {
+  const selector = [
+    "a[href]", "button", "input", "textarea", "select", "summary",
+    "[role='button']", "[role='link']", "[role='checkbox']", "[role='radio']",
+    "[role='tab']", "[role='menuitem']", "[role='textbox']", "[onclick]",
+    "[tabindex]", "[contenteditable='true']"
+  ].join(",");
+  const width = Math.max(1, window.innerWidth);
+  const height = Math.max(1, window.innerHeight);
+  const elements = [];
+  let truncated = false;
+  for (const node of document.querySelectorAll(selector)) {
+    if (elements.length >= limit) {
+      truncated = true;
+      break;
+    }
+    const rect = node.getBoundingClientRect();
+    const style = window.getComputedStyle(node);
+    if (
+      style.visibility === "hidden" ||
+      style.visibility === "collapse" ||
+      style.display === "none" ||
+      Number(style.opacity) === 0 ||
+      rect.width < 2 ||
+      rect.height < 2 ||
+      rect.right <= 0 ||
+      rect.bottom <= 0 ||
+      rect.left >= width ||
+      rect.top >= height
+    ) {
+      continue;
+    }
+    const left = Math.max(0, rect.left);
+    const top = Math.max(0, rect.top);
+    const right = Math.min(width, rect.right);
+    const bottom = Math.min(height, rect.bottom);
+    const inputType = String(node.getAttribute("type") || "").toLowerCase();
+    const autocomplete = String(
+      node.getAttribute("autocomplete") || ""
+    ).toLowerCase();
+    const sensitive = (
+      inputType === "password" ||
+      inputType === "hidden" ||
+      autocomplete === "current-password" ||
+      autocomplete === "new-password" ||
+      autocomplete === "one-time-code" ||
+      autocomplete === "webauthn" ||
+      autocomplete.startsWith("cc-")
+    );
+    const safeValue = sensitive ? "" : String(node.value || "");
+    const text = String(node.innerText || safeValue).trim().slice(0, 240);
+    const label = String(
+      node.getAttribute("aria-label") ||
+      node.getAttribute("title") ||
+      node.getAttribute("placeholder") ||
+      text
+    ).trim().slice(0, 240);
+    elements.push({
+      bounds: {
+        x: left / width,
+        y: top / height,
+        width: Math.max(0.000001, (right - left) / width),
+        height: Math.max(0.000001, (bottom - top) / height)
+      },
+      label: label || null,
+      role: node.getAttribute("role") || node.tagName.toLowerCase(),
+      text: text || null,
+      metadata: {
+        tag: node.tagName.toLowerCase(),
+        input_type: inputType || null,
+        autocomplete: autocomplete || null,
+        sensitive,
+        disabled: Boolean(node.disabled),
+        focused: node === document.activeElement
+      }
+    });
+  }
+  return {elements, truncated};
+}
+"""
+
+
+class BrowserComputerEnvironment(ComputerEnvironment):
+    """Ephemeral Playwright implementation of a computer environment."""
+
+    def __init__(
+        self,
+        *,
+        session_id: str,
+        workspace: Any,
+        manager: BrowserSessionManager | None = None,
+        observation_store: ObservationStore | None = None,
+        headless: bool = True,
+        viewport_width: int = 1280,
+        viewport_height: int = 720,
+    ) -> None:
+        super().__init__(session_id)
+        if viewport_width <= 0 or viewport_height <= 0:
+            raise ValueError("viewport dimensions must be positive")
+        self.workspace = workspace
+        self.manager = manager or get_browser_manager()
+        self.observation_store = observation_store or ObservationStore(workspace)
+        self.headless = headless
+        self.viewport_width = viewport_width
+        self.viewport_height = viewport_height
+
+    async def _get_page(self) -> Any:
+        session = await self.manager.get_or_create(
+            self.session_id,
+            headless=self.headless,
+        )
+        page = await session.get_page()
+        requested = {
+            "width": self.viewport_width,
+            "height": self.viewport_height,
+        }
+        if page.viewport_size != requested:
+            await page.set_viewport_size(cast(Any, requested))
+        return page
+
+    async def _close(self) -> None:
+        await self.manager.close(self.session_id)
+
+    async def _observe(self) -> ComputerObservation:
+        return await self._capture_observation(await self._get_page())
+
+    async def _execute(self, batch: ComputerActionBatch) -> ComputerObservation:
+        page = await self._get_page()
+        action = batch.actions[0]
+        await self._execute_action(page, action)
+        if action.type not in {
+            ComputerActionType.SCREENSHOT,
+            ComputerActionType.WAIT,
+        }:
+            await page.wait_for_timeout(250)
+        return await self._capture_observation(page)
+
+    async def _capture_observation(self, page: Any) -> ComputerObservation:
+        viewport = await self._read_viewport(page)
+        frame_id = f"frame-{uuid4().hex}"
+        screenshot_bytes = await page.screenshot(full_page=False, type="png")
+        screenshot = await self.observation_store.save_screenshot(
+            session_id=self.session_id,
+            frame_id=frame_id,
+            image_bytes=screenshot_bytes,
+            mime_type="image/png",
+            viewport=viewport,
+            text_fallback="Current browser viewport screenshot.",
+        )
+        try:
+            title = await page.title()
+        except Exception:  # noqa: BLE001 - title is optional observation metadata.
+            title = None
+
+        metadata: dict[str, Any] = {
+            "browser_runtime_kind": "ephemeral_playwright",
+            "supported_actions": [action.value for action in ComputerActionType],
+        }
+        try:
+            elements, truncated = await self._read_interactive_elements(page)
+        except Exception:  # noqa: BLE001 - screenshots remain useful without DOM hints.
+            logger.warning(
+                "Failed to collect interactive browser elements for %s",
+                self.session_id,
+                exc_info=True,
+            )
+            elements = []
+            truncated = False
+            metadata[ELEMENT_EXTRACTION_FAILED_KEY] = True
+        if truncated:
+            metadata[ELEMENTS_TRUNCATED_KEY] = True
+        frames = getattr(page, "frames", ())
+        if isinstance(frames, (list, tuple)) and len(frames) > 1:
+            # This adapter intentionally reports only top-level DOM hints. The
+            # screenshot remains authoritative for cross-origin child frames.
+            metadata[ELEMENT_EXTRACTION_INCOMPLETE_KEY] = True
+
+        active_url = str(getattr(page, "url", "") or "").strip() or None
+        return ComputerObservation(
+            session_id=self.session_id,
+            frame_id=frame_id,
+            environment=ComputerEnvironmentType.BROWSER,
+            viewport=viewport,
+            screenshot=screenshot,
+            elements=elements,
+            active_url=active_url,
+            title=str(title) if title else None,
+            metadata=metadata,
+        )
+
+    async def _read_viewport(self, page: Any) -> Viewport:
+        size = page.viewport_size
+        if not size:
+            size = await page.evaluate(
+                "() => ({width: window.innerWidth, height: window.innerHeight})"
+            )
+        try:
+            device_pixel_ratio = float(
+                await page.evaluate("() => window.devicePixelRatio || 1")
+            )
+        except Exception:  # noqa: BLE001 - DPR defaults safely.
+            device_pixel_ratio = 1.0
+        return Viewport(
+            width=int(size["width"]),
+            height=int(size["height"]),
+            device_pixel_ratio=device_pixel_ratio,
+        )
+
+    async def _read_interactive_elements(
+        self,
+        page: Any,
+    ) -> tuple[list[ComputerElement], bool]:
+        payload = await page.evaluate(
+            _INTERACTIVE_ELEMENTS_SCRIPT,
+            MAX_OBSERVATION_ELEMENTS,
+        )
+        if not isinstance(payload, dict):
+            return [], False
+        raw_elements = payload.get("elements")
+        if not isinstance(raw_elements, list):
+            return [], bool(payload.get("truncated"))
+        elements: list[ComputerElement] = []
+        for index, raw_element in enumerate(
+            raw_elements[:MAX_OBSERVATION_ELEMENTS],
+            start=1,
+        ):
+            if not isinstance(raw_element, dict):
+                continue
+            try:
+                elements.append(
+                    ComputerElement(
+                        element_id=f"dom-{index}",
+                        source=ComputerElementSource.DOM,
+                        bounds=raw_element["bounds"],
+                        label=raw_element.get("label"),
+                        role=raw_element.get("role"),
+                        text=raw_element.get("text"),
+                        metadata=raw_element.get("metadata") or {},
+                    )
+                )
+            except (KeyError, TypeError, ValueError):
+                logger.debug("Skipping invalid DOM element payload", exc_info=True)
+        return elements, bool(payload.get("truncated"))
+
+    async def _execute_action(self, page: Any, action: ComputerAction) -> None:
+        if action.type is ComputerActionType.SCREENSHOT:
+            return
+        if action.type is ComputerActionType.NAVIGATE:
+            assert action.url is not None
+            await page.goto(
+                self._resolve_navigation_url(action.url),
+                wait_until="domcontentloaded",
+                timeout=30_000,
+            )
+            return
+        if action.type is ComputerActionType.WAIT:
+            await page.wait_for_timeout(action.duration_ms)
+            return
+        if action.type is ComputerActionType.KEYPRESS:
+            await page.keyboard.press(self._playwright_key_chord(action.keys))
+            return
+        if action.type is ComputerActionType.TYPE:
+            await page.keyboard.insert_text(action.text or "")
+            return
+        if action.type is ComputerActionType.REPLACE_TEXT:
+            x, y = self._target_pixels(action)
+            await page.mouse.click(x, y)
+            modifier = "Meta" if sys.platform == "darwin" else "Control"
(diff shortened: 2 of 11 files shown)


src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,398 @@
...
+                )
+            except (KeyError, TypeError, ValueError):
+                logger.debug("Skipping invalid DOM element payload", exc_info=True)
+        return elements, bool(payload.get("truncated"))
+
+    async def _execute_action(self, page: Any, action: ComputerAction) -> None:
+        if action.type is ComputerActionType.SCREENSHOT:
+            return
+        if action.type is ComputerActionType.NAVIGATE:
+            assert action.url is not None
```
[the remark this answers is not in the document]
qinxuye: Fixed in 23f5dec6. The NAVIGATE branch now performs an explicit null check and raises a descriptive ValueError instead of relying on an assert. Pydantic still validates the action at the schema boundary; this is the adapter-side defensive check.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,398 @@
...
+            return
+        if action.type is ComputerActionType.SCROLL:
+            viewport = self._current_viewport()
+            await page.mouse.wheel(
+                action.delta_x * viewport.width,
+                action.delta_y * viewport.height,
+            )
+            return
+        if action.type is ComputerActionType.DRAG:
+            assert action.start is not None and action.end is not None
```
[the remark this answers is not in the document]
qinxuye: Fixed in 23f5dec6. The DRAG branch now checks both endpoints explicitly and raises a descriptive ValueError instead of relying on an assert.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,398 @@
...
+            return self._point_pixels(action.target.point)
+        target_id = action.target.element_id
+        observation = self.current_observation
+        if target_id is None or observation is None:
+            raise ValueError("element target requires a current observation")
+        element = next(
+            element
+            for element in observation.elements
+            if element.element_id == target_id
+        )
```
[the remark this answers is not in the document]
qinxuye: Fixed in 23f5dec6. Element lookup now uses a None default and raises ComputerTargetNotFoundError with the missing element ID. The base environment remains the primary frame/target validator, while the adapter now fails clearly if its invariant is violated.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+        observation_store: ObservationStore | None = None,
+        headless: bool = True,
+        viewport_width: int = 1280,
+        viewport_height: int = 720,
+    ) -> None:
+        super().__init__(session_id)
+        if viewport_width <= 0 or viewport_height <= 0:
+            raise ValueError("viewport dimensions must be positive")
+        self.workspace = workspace
+        self.manager = manager or get_browser_manager()
```
rogercloud, line 161: **Major — shared session breaks the advertised "isolated browser" guarantee.**

Session resolution here goes through the same process-global `get_browser_manager()` singleton that the legacy browser tools use, and the id comes from the same `BrowserTaskSessionMixin._default_session_id` (`src/xagent/core/tools/adapters/vibe/computer.py:160`). Consequences:

- A legacy tool mutating the page invalidates nothing here — no invalidation hook fires, so the cached frame can silently drift from reality.
- Lines 143-148 unconditionally resize the shared page to 1280x720; the legacy default is 1920x1080 (`src/xagent/core/tools/core/browser_use.py:96,109`). That's a live behavior change for existing browser tools on the same session.
- `BrowserSessionManager.get_or_create` deliberately ignores `headless` for a pre-existing session (`src/xagent/core/tools/core/browser_use.py:230-236`), and `browser_navigate` defaults `headless=False` (`:350`) — so `ComputerTool(headless=True)` is silently dropped whenever a legacy tool touched the session first.

Since `create_browser_tools()` now registers both families by default, the tool's model-facing description ("an isolated temporary browser… does not inherit the user's existing browser profile or signed-in sessions") isn't accurate in the default config.

Suggestion: give the computer environment its own session-id namespace (e.g. a suffix on the derived id) so it never aliases a legacy session. If sharing is deliberate, calling the existing `invalidate_observation()` hook from the legacy mutating paths and dropping the unconditional resize would make it safe.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+        self.observation_store = observation_store or ObservationStore(workspace)
+        self.headless = headless
+        self.viewport_width = viewport_width
+        self.viewport_height = viewport_height
+
+    async def _get_page(self) -> Any:
```
rogercloud: **Major — no liveness check means a recreated session leaves a stale frame over a blank page.**

`_get_page` calls `get_or_create` without verifying the returned session is the one the cached observation was captured from. If the session is evicted out-of-band — the `browser_close` tool, or the shared manager's 30-minute idle sweep (`src/xagent/core/tools/core/browser_use.py:217-286`) — `get_or_create` quietly launches a fresh `about:blank` browser under the same key.

`validate_action_batch` (`src/xagent/core/computer/environment.py:91-123`) only compares a frame-id string; it never checks page/session identity. So the next model action passes validation and executes stale coordinates against a blank page, with nothing surfaced as an error.

Suggestion: track a session identity or generation counter in `_get_page`, and invalidate the current observation (or raise a frame-mismatch error) when the underlying session object no longer matches the one the current frame came from.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+            metadata[ELEMENT_EXTRACTION_FAILED_KEY] = True
+        if truncated:
+            metadata[ELEMENTS_TRUNCATED_KEY] = True
+        frames = getattr(page, "frames", ())
+        if isinstance(frames, (list, tuple)) and len(frames) > 1:
+            # This adapter intentionally reports only top-level DOM hints. The
+            # screenshot remains authoritative for cross-origin child frames.
+            metadata[ELEMENT_EXTRACTION_INCOMPLETE_KEY] = True
+
+        active_url = str(getattr(page, "url", "") or "").strip() or None
```
rogercloud: **Major — an out-of-allowlist page URL wedges the environment with no recovery path.**

`page.url` goes straight into `ComputerObservation.active_url`, typed `ObservedUrl` (`src/xagent/core/computer/schema.py`), which enforces a scheme allowlist, an 8192-char cap, and rejects embedded credentials. Executing the model directly confirms `data:`, `file:`, `view-source:`, `user:pass@` URLs, and over-long URLs all raise `ValidationError`.

That propagates through the observe/execute wrapper in `environment.py`, which invalidates the frame and re-raises; `ComputerTool.run_json_async` catches it as a generic `ValueError` and returns an error with `current_observation` now `None`. A screenshot request is the only way to get a fresh frame — and it re-runs `_capture_observation` on the same still-bad URL. So the environment stays wedged for the rest of its lifetime while the page remains there. `file://` is reachable in practice, since the shared session's legacy `browser_navigate` explicitly supports it.

Suggestion: wrap the `active_url` assignment in try/except and fall back to `None` plus a metadata flag (e.g. `active_url_unavailable`), so an unusual URL degrades to "screenshot only" rather than bricking the loop.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -0,0 +1,274 @@
...
+                session_id="",
+                error="Computer tool requires a task-scoped browser session.",
+            )
+        if self._workspace is None:
+            return self._error_result(
+                session_id=session_id,
+                error="Computer tool requires a task workspace for observations.",
+            )
+
+        environment = self._environments.get(session_id)
```
rogercloud, line 181: **Major — environments (and their Chromium processes) are never closed per DAG step.**

`_environments` is keyed by a session id embedding the DAG step id. For a plan where several steps each call `computer`, every step mints its own environment and its own headless Chromium — and nothing closes any of them until task-level `teardown()`. There's no per-step cleanup call in `dag.py` or `react.py`.

Each environment's `ObservationStore` also retains up to 40 screenshots (`src/xagent/core/computer/store.py:31`). The 30-minute idle sweep on the shared browser-session manager is a different layer and doesn't bound this dict.

Suggestion: close and evict an environment when its owning DAG step completes, or cap `_environments` with LRU eviction that calls `close()`.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+            min(point.y * viewport.height, viewport.height - 1),
+        )
+
+    def _current_viewport(self) -> Viewport:
+        observation = self.current_observation
+        if observation is None:
+            raise RuntimeError("computer action requires a current observation")
+        return observation.viewport
+
+    @staticmethod
```
rogercloud, line 429: **Major — navigation has no SSRF protection, though a house helper already exists.**

`_resolve_navigation_url` validates only that the scheme is http/https and a hostname is present. Nothing rejects localhost, RFC1918 ranges, or link-local addresses like the `169.254.169.254` cloud metadata endpoint.

The repo already has `reject_private_network_host` / `validate_public_http_url` in `src/xagent/core/utils/security.py`, used by at least five other network-facing tools (`a2a_agent_tool.py`, `mcp_oauth.py`, `web_content.py`, `download_web_asset.py`, `vision_tool.py`).

This isn't a regression — legacy `browser_navigate` has no URL validation at all — but the resulting screenshot and DOM text are returned straight to the model, which makes it a reachable SSRF/information-disclosure primitive, either via a direct `navigate` action or via a prompt-injected link in page content. New navigation code seems like a natural place to start closing this.

Suggestion: call the existing helper here, with a config escape hatch for local development if needed.

---

Separately (informational, no action needed): the scheme/hostname raise on lines 374-380 is unreachable — `NavigationUrl`'s pydantic validator already rejects invalid schemes at `ComputerAction` construction time, confirmed by direct construction. Same for the `ComputerTargetNotFoundError` path at line 351, which `validate_action_batch` pre-empts (its test has to call the private `_target_pixels` directly). Both document intent rather than live code paths.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+      rect.left >= width ||
+      rect.top >= height
+    ) {
+      continue;
+    }
+    const left = Math.max(0, rect.left);
+    const top = Math.max(0, rect.top);
+    const right = Math.min(width, rect.right);
+    const bottom = Math.min(height, rect.bottom);
+    const inputType = String(node.getAttribute("type") || "").toLowerCase();
```
rogercloud, line 98: **Major — sensitive-value redaction has confirmed bypasses, and the injected script has no real test coverage.**

Redaction here is gated purely on the live `type`/`autocomplete` attributes. Confirmed gaps:

1. A page-side "show password" toggle flipping `type="password"` to `type="text"` causes the plaintext value to be emitted verbatim — the type is read live at extraction time and the original isn't cached.
2. Fields named or id'd `password`/`otp`/`ssn`/`cvv`/`card` with `autocomplete="off"` — a common real-world pattern — aren't caught, since `name`/`id`/`placeholder` are never consulted.
3. Any ordinary text input's value is emitted verbatim up to 240 chars, with no PII heuristic and no total-output cap.

The injected script — this redaction logic included — also never actually executes in any test: `tests/core/computer/test_browser_environment.py`'s `FakePage.evaluate` pattern-matches on script substrings and returns a canned dict.

Suggestion: add name/id/placeholder heuristics (a regex over `pass|pwd|otp|cvv|ssn|secret|token|card`), and consider emitting only `has_value` plus a length for all input values, relying on the screenshot as the authoritative view. One real-browser test exercising this path end to end would make future edits here much safer.

src/xagent/core/tools/adapters/vibe/browser_use.py
```diff
@@ -1135,7 +1135,10 @@ def create_browser_tools(
     Returns:
         List of browser tool instances
     """
+    from .computer import ComputerTool
```
rogercloud, line 1138: **Major — on by default with no config gate, and the #1098 policy contract shipped alongside it is never invoked.**

Two separate, narrower points, neither about confirmation UX or product authorization flows (which this PR's description already scopes out):

- `ComputerTool` is prepended unconditionally here, in contrast to the `if include_debug_tools:` gate a few lines below. `browser_tools_enabled` is hardcoded `True` on every production call path (`config.py:491`, `web/tools/config.py:1128`, `web/api/tools.py:303`, `web/api/chat.py:644`, `agent/service.py:644-645`), so the one conditional consumer never actually sees `False`.
- Separately — and more squarely a #1098-conformance question, per this PR's own review-focus item on conformance with the #1098 schema/environment contracts — the foundation's `ComputerActionPolicy`/`ComputerPolicyDecision` protocol (`src/xagent/core/computer/policy.py`) has zero production call sites in `src/`: fully defined, never invoked before an action executes, even as a no-op allow-all.

Suggestion: a config flag for this slice so `computer` can be disabled without a code change, and calling `ComputerActionPolicy.evaluate` (even with an allow-all default) from `ComputerEnvironment.execute` so the hook the foundation defines is actually exercised — neither of these requires any UI or approval flow.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+            )
+            return
+        if action.type is ComputerActionType.WAIT:
+            await page.wait_for_timeout(action.duration_ms)
+            return
+        if action.type is ComputerActionType.KEYPRESS:
+            await page.keyboard.press(self._playwright_key_chord(action.keys))
+            return
+        if action.type is ComputerActionType.TYPE:
+            await page.keyboard.insert_text(action.text or "")
```
rogercloud, line 346: **Minor — `insert_text` dispatches no keydown/keyup events.**

Tracing Playwright's driver, `page.keyboard.insert_text` fires only an `input` event. Keydown-driven autocomplete, search-as-you-type, and `maxlength` enforcement won't react, even though the resulting screenshot will look correct — which makes this a failure mode the model can't detect. Consider `page.keyboard.type()` as the default, keeping `insert_text` for long text where the per-key cost matters.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+            "ESCAPE": "Escape",
+            "HOME": "Home",
+            "META": "Meta",
+            "PAGEDOWN": "PageDown",
+            "PAGEUP": "PageUp",
+            "SHIFT": "Shift",
+            "SPACE": "Space",
+            "TAB": "Tab",
+        }
+        return "+".join(aliases.get(key.upper(), key) for key in keys)
```
rogercloud: **Minor — joining all keys with `+` holds modifiers instead of pressing sequentially.**

`keys=["ArrowDown","Enter"]` becomes `"ArrowDown+Enter"`. Tracing Playwright's driver, this doesn't raise — it silently holds "ArrowDown" down as a pseudo-modifier while pressing "Enter", so it's a real behavioral bug that surfaces as nothing at all.

Consider validating that all but the final key are genuine modifiers (Alt/Control/Meta/Shift), or interpreting the list as sequential presses.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+        if action.type is ComputerActionType.WAIT:
+            await page.wait_for_timeout(action.duration_ms)
+            return
+        if action.type is ComputerActionType.KEYPRESS:
+            await page.keyboard.press(self._playwright_key_chord(action.keys))
+            return
+        if action.type is ComputerActionType.TYPE:
+            await page.keyboard.insert_text(action.text or "")
+            return
+        if action.type is ComputerActionType.REPLACE_TEXT:
```
rogercloud, line 348: **Minor — `replace_text` doesn't verify the click landed on an editable element.**

Lines 296-302 click, then select-all + insert, without confirming focus landed on an editable target. If it didn't, `Ctrl/Cmd+A` selects the whole page and the subsequent insert goes somewhere unintended. A focus/editability check after the click would make the failure explicit.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -0,0 +1,274 @@
...
+    def run_json_sync(self, args: Mapping[str, Any]) -> Any:
+        raise NotImplementedError("computer is async-only")
+
+    async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
+        raw_args = dict(args)
+        step_id = raw_args.pop(_STEP_SESSION_ARG, None)
+        # The session is an execution-scoped resource. Never let model output
+        # select another task's browser, even if it invents a session_id field.
+        raw_args.pop("session_id", None)
+        parsed = ComputerToolArgs.model_validate(raw_args)
```
rogercloud: **Minor — argument validation runs outside the tool's own error handling.**

`ComputerToolArgs.model_validate` is called before the `try:` block that starts at line 183. A validation error therefore bypasses the structured `ComputerToolResult(success=False, ...)` and is caught instead by the generic exception handler in the ReAct loop's tool-execution wrapper, producing a plain error dict with no frame or session context — so the model loses the frame id it needs to recover. Moving the validate call inside the try would keep the error shape consistent with every other failure here.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,407 @@
...
+    async def _observe(self) -> ComputerObservation:
+        return await self._capture_observation(await self._get_page())
+
+    async def _execute(self, batch: ComputerActionBatch) -> ComputerObservation:
+        page = await self._get_page()
+        action = batch.actions[0]
+        await self._execute_action(page, action)
+        if action.type not in {
+            ComputerActionType.SCREENSHOT,
+            ComputerActionType.WAIT,
```
rogercloud, line 209: **Minor — fixed 250ms settle delay, no load-state wait.**

Lines 157-166: the only post-action settle for click/keypress/type is a flat 250ms `wait_for_timeout`, with no `wait_for_load_state`. SPA route changes or XHR updates slower than that produce an observation labeled fresh that actually shows the pre-update view — and since the frame contract treats it as authoritative, the model will act on it. A `wait_for_load_state` (or a network-idle bound) alongside the timeout would tighten this.

src/xagent/core/agent/context/execution.py
```diff
@@ -197,6 +204,49 @@ def add_tool_result(
...
+    def _compact_superseded_tool_messages(self, scope: str) -> None:
+        """Keep stale observations durable while removing them from live prompts."""
+
+        for index, message in enumerate(self.messages):
+            metadata = message.metadata or {}
+            if metadata.get("supersedes_scope") != scope:
+                continue
+            if metadata.get("superseded"):
+                continue
+            self.messages[index] = replace(
```
rogercloud, line 216: **Minor — supersession invalidates the token-accounting fingerprint on every call.**

This `replace()` call rewrites the message's content and clears its context refs. Downstream, `_get_total_tokens`'s cache-validity fingerprint (lines 1246-1265) changes for the superseded message on both terms it checks, so the equality check fails immediately after any supersede and the rest of the run falls back to the `char/4` heuristic rather than the provider-accurate count. The error direction is toward underestimating, so the practical impact is mild — but the mechanism looks accidental rather than chosen.

Cosmetic companion: `_compact_transcript` appends an empty-string chunk for superseded messages, since `context_refs` stays truthy while the text becomes empty.

src/xagent/core/agent/context/execution.py
```diff
@@ -197,6 +204,49 @@ def add_tool_result(
...
+                message,
+                content=self._superseded_tool_summary(message),
+                metadata={
+                    **metadata,
+                    "superseded": True,
+                    "raw_result": None,
```
rogercloud: **Minor (defensive) — `raw_result=None` on a superseded tool message would read as "succeeded".**

Currently unreachable: the newest tool message is never the one superseded, and tail-preserving compaction keeps the newest messages. But the invariant is incidental rather than enforced, so it'd quietly break if compaction changes later. A `{"superseded": True}` sentinel instead of `None`, or having the success check skip superseded messages, would be cheap insurance.

rogercloud (commented):
## Summary

This PR layers a real browser onto the provider-neutral Computer Use foundation from #1098: a `BrowserComputerEnvironment` that drives an ephemeral headless Playwright page, plus a `ComputerTool` that exposes it to the agent. Each turn captures the viewport as a screenshot `FileRef` alongside a bounded set of DOM hints, and the one-action-per-observed-frame contract is enforced through frame ids so the model can never act on a view it hasn't seen. It also wires browser lifecycle into the task/DAG step boundaries and adds a supersede mechanism that drops stale browser observations from the live prompt while keeping their `FileRef`s in durable execution state.

The frame/action contract, the observation store, and the supersede mechanism are the strongest parts here — they're cleanly built and fit the #1098 foundation well. Overall verdict: **acceptable with reservations**, and the reservations all cluster at one seam: the interaction between this new environment and the pre-existing browser tooling. Specifically, the "isolated session" property the tool advertises to the model doesn't actually hold (it shares a Playwright page and session with the legacy browser tools), and the foundation's own frame-invalidation hook is left unwired. Those aren't just structural preferences; they have concrete correctness and security consequences, so they're written up below.

Everything below is offered as feedback rather than gating — happy to be told any of it is intentional for this slice.

## Major

**1. The shared Playwright page means the "isolated browser" guarantee doesn't hold** — `src/xagent/core/computer/browser.py:131`

`ComputerTool` derives its session id via the same `BrowserTaskSessionMixin._default_session_id` the legacy tools use (`src/xagent/core/tools/adapters/vibe/computer.py:160`), and resolution goes through the same process-global `get_browser_manager()` singleton. Three follow-on effects:

- A legacy tool mutating the page invalidates nothing on the computer-use side — no invalidation hook fires, so the cached frame can silently desynchronize from reality.
- `browser.py:143-148` unconditionally resizes the shared page to 1280x720, while the legacy default is 1920x1080 (`src/xagent/core/tools/core/browser_use.py:96,109`). That's a live behavioral change for existing browser tools sharing the session.
- `BrowserSessionManager.get_or_create` deliberately ignores `headless` for a pre-existing session (`src/xagent/core/tools/core/browser_use.py:230-236`), and `browser_navigate` defaults `headless=False` (`:350`). So a `ComputerTool(headless=True)` request is silently dropped whenever a legacy tool touched the session first.

Because `create_browser_tools()` now registers both tool families into the same task by default, the tool's own model-facing description ("an isolated temporary browser… does not inherit the user's existing browser profile or signed-in sessions") is inaccurate in the default configuration. Suggestion: give the computer environment its own session-id namespace (e.g. a suffix on the derived id) so it never aliases a legacy session. If sharing is intentional, then calling the existing `invalidate_observation()` hook from the legacy mutating paths and dropping the unconditional resize would make the sharing safe.

**2. No liveness check on the underlying session — a recreated session leaves a stale frame pointing at a blank page** — `src/xagent/core/computer/browser.py:137`

`_get_page` (`browser.py:137-149`) calls `get_or_create` without checking that the returned session is the same one the cached observation came from. If the session is evicted out-of-band — via the `browser_close` tool, or the shared manager's 30-minute idle sweep (`src/xagent/core/tools/core/browser_use.py:217-286`) — `get_or_create` quietly launches a fresh `about:blank` browser under the same key. `validate_action_batch` (`src/xagent/core/computer/environment.py:91-123`) only compares a frame-id string and never checks page/session identity, so the next action passes validation and executes stale coordinates against a blank page, with no error surfaced to the model. Suggestion: track a session identity or generation counter in `_get_page` and invalidate the current observation (or raise a frame-mismatch error) when the underlying session object no longer matches.

**3. An out-of-allowlist page URL wedges the environment with no recovery path** — `src/xagent/core/computer/browser.py:208`

`browser.py:208-217` passes `page.url` straight into `ComputerObservation.active_url`, typed `ObservedUrl` (`src/xagent/core/computer/schema.py`), which enforces a scheme allowlist, an 8192-char cap, and rejects embedded credentials. Executing the model directly confirms `data:`, `file:`, and `view-source:` schemes, URLs containing `user:pass@`, and over-long URLs all raise `ValidationError`. That propagates through the observe/execute wrapper in `environment.py`, which invalidates the frame and re-raises; `ComputerTool.run_json_async` catches it as a generic `ValueError` and returns an error with `current_observation` now `None`. Since a screenshot request is the only way to obtain a fresh frame, and it re-runs `_capture_observation` against the same still-bad URL, the environment stays wedged for the rest of its lifetime as long as the page remains there. `file://` is reachable in practice because the shared session's legacy `browser_navigate` explicitly supports it. Suggestion: wrap the `active_url` assignment in try/except and degrade to `None` plus a metadata flag (e.g. `active_url_unavailable`) so an unusual URL means "screenshot only" rather than a bricked loop.

**4. Environments and their Chromium processes are never closed per DAG step** — `src/xagent/core/tools/adapters/vibe/computer.py:172`

`ComputerTool._environments` (`computer.py:116,172-179`) is keyed by a session id embedding the DAG step id. For a plan where several steps each call `computer`, each step mints its own environment and its own Chromium process, and nothing closes any of them until task-level `teardown()` — there's no per-step cleanup anywhere in `dag.py`/`react.py`. Each environment's `ObservationStore` also holds up to 40 screenshots (`src/xagent/core/computer/store.py:31`). The 30-minute idle sweep on the shared session manager is a different layer and doesn't bound this dict. Suggestion: close and evict the environment when its owning step completes, or cap `_environments` with LRU eviction that calls `close()`.

**5. Navigation has no SSRF protection, though the repo already has a helper for exactly this** — `src/xagent/core/computer/browser.py:374`

`_resolve_navigation_url` (`browser.py:374-380`, new here) only checks that the scheme is http/https and a hostname is present — nothing rejects localhost, RFC1918 ranges, or link-local addresses such as the `169.254.169.254` metadata endpoint. `src/xagent/core/utils/security.py` already provides `reject_private_network_host` / `validate_public_http_url`, used by at least five other network-facing tools (`a2a_agent_tool.py`, `mcp_oauth.py`, `web_content.py`, `download_web_asset.py`, `vision_tool.py`). This isn't a regression — legacy `browser_navigate` has no URL validation at all — but since the resulting screenshot and DOM text go straight back to the model, it's a reachable SSRF/information-disclosure primitive, either via a direct `navigate` action or via a prompt-injected link in page content. New navigation code seems like a good place to start closing the gap. Suggestion: call the existing helper here, with a config escape hatch for local development if that's needed.

**6. DOM-hint redaction has bypasses, and the injected script has no real test coverage** — `src/xagent/core/computer/browser.py:67`

`browser.py:67-81` gates redaction purely on the live `type`/`autocomplete` attributes. Confirmed gaps:

- A page-side "show password" toggle flipping `type="password"` to `type="text"` causes the plaintext value to be emitted verbatim, since the type is read live at extraction time and the original isn't cached.
- Fields named or id'd `password`/`otp`/`ssn`/`cvv`/`card` with `autocomplete="off"` — a common real-world pattern — aren't caught, because `name`/`id`/`placeholder` are never consulted.
- Any ordinary text input's value is emitted verbatim up to 240 chars, with no PII heuristic and no total-output cap.

Separately, the injected script (including this redaction logic) never actually executes in any test: `tests/core/computer/test_browser_environment.py`'s `FakePage.evaluate` pattern-matches on script substrings and returns a canned dict. Suggestion: add name/id/placeholder heuristics (a regex over `pass|pwd|otp|cvv|ssn|secret|token|card`), and consider emitting only `has_value` plus a length for all input values, leaning on the screenshot as the authoritative view. One real-browser test exercising the redaction path end to end would also make future changes here much safer.

**7. The `computer` tool is on by default with no config gate, and the #1098 policy contract it ships alongside is never invoked** — `src/xagent/core/tools/adapters/vibe/browser_use.py:1138`

Two separate, narrower points here — neither is about building confirmation UX or product authorization flows, which the PR description already scopes out for this slice:

- `create_browser_tools` (`browser_use.py:1136-1150`) prepends `ComputerTool` unconditionally, in contrast to the `if include_debug_tools:` gate a few lines below for other tools. `browser_tools_enabled` is hardcoded `True` on every production call path (`config.py:491`, `web/tools/config.py:1128`, `web/api/tools.py:303`, `web/api/chat.py:644`, `agent/service.py:644-645`), so the one conditional consumer never sees `False` in practice.
- Separately, and more squarely a #1098-conformance question (per this PR's own "conformance with the #1098 Computer Use schema and environment contracts" review focus): the foundation's `ComputerActionPolicy`/`ComputerPolicyDecision` protocol (`src/xagent/core/computer/policy.py`) has zero production call sites anywhere in `src/` — it's fully defined but never invoked before an action executes, even as a no-op allow-all. That's a gap against the contract this PR is meant to conform to, independent of any confirmation UI question.

Suggestion: consider a config flag for this slice so `computer` can be disabled without a code change, and calling `ComputerActionPolicy.evaluate` (even with an allow-all default policy) from `ComputerEnvironment.execute` so the hook the foundation defines is actually exercised — no UI or approval flow required for either.

## Minor

**8. `type` uses `insert_text`, which fires no keydown/keyup** — `src/xagent/core/computer/browser.py:294`. Tracing Playwright's driver, `page.keyboard.insert_text` emits only an `input` event, so keydown-driven autocomplete, search-as-you-type, and `maxlength` enforcement won't react — while the screenshot looks correct. `page.keyboard.type()` might be the better default, keeping `insert_text` for long text.

**9. `keypress` joins keys with `+`, holding modifiers instead of pressing sequentially** — `src/xagent/core/computer/browser.py:407`. `keys=["ArrowDown","Enter"]` becomes `"ArrowDown+Enter"`. This doesn't error — Playwright silently holds "ArrowDown" as a pseudo-modifier while pressing "Enter", so it's a quiet behavioral bug. Consider validating that all but the last key are real modifiers (Alt/Control/Meta/Shift), or treating the list as sequential presses.

**10. `replace_text` doesn't verify the click landed on an editable element** — `src/xagent/core/computer/browser.py:296`. `browser.py:296-302` clicks then does select-all + insert without confirming focus is on an editable target; if it isn't, `Ctrl/Cmd+A` selects the whole page.

**11. Argument validation runs outside the tool's own error handling** — `src/xagent/core/tools/adapters/vibe/computer.py:159`. `ComputerToolArgs.model_validate` is called before the `try:` at line 183, so a validation error bypasses the structured `ComputerToolResult(success=False, ...)` and instead lands in the ReAct loop's generic tool-execution handler, producing a plain error dict with no frame/session context. Moving the validate call inside the try would keep the error shape consistent.

**12. Fixed 250ms settle delay, no load-state wait** — `src/xagent/core/computer/browser.py:163`. `browser.py:157-166` uses a flat 250ms `wait_for_timeout` as the only post-action settle for click/keypress/type, with no `wait_for_load_state`. SPA route changes or XHR updates slower than that yield a stale "fresh" observation.

**13. Two defensive branches are unreachable (informational only)** — `src/xagent/core/computer/browser.py:374`. `_resolve_navigation_url`'s scheme/hostname raise can't fire, since `NavigationUrl`'s pydantic validator already rejects bad schemes at `ComputerAction` construction. Likewise the `ComputerTargetNotFoundError` path at `browser.py:351` is unreachable through `execute()`, because `validate_action_batch` rejects unknown element ids first — the test has to call the private `_target_pixels` directly to reach it. No action needed; just noting these document intent rather than live paths.

**14. Supersession invalidates the token-accounting fingerprint on every call** — `src/xagent/core/agent/context/execution.py:1256`. `_get_total_tokens`'s cache-validity fingerprint (`execution.py:1246-1265`) changes on both terms it checks for the superseded message, so the equality check fails right after any supersede and the remainder of the run falls back to the `char/4` heuristic instead of the provider-accurate count. The error direction is toward underestimating, so it's fairly benign, but it looks accidental rather than intended. Cosmetic companion: `_compact_transcript` appends an empty-string chunk for superseded messages, since `context_refs` stays truthy while its text becomes empty.

**15. `raw_result=None` on a superseded tool message would read as "succeeded"** — `src/xagent/core/agent/context/execution.py:222`. Currently unreachable — the newest tool message is never the one superseded, and tail-preserving compaction keeps the newest messages — but a `{"superseded": True}` sentinel (or having the success check skip superseded messages) would be cheap insurance if compaction logic shifts later.

## Also checked, no issue found

- **Coordinate-basis mismatch**: the injected script normalizes against `window.innerWidth` while Playwright de-normalizes against `viewport_size`. Verified live in a real Chromium instance — `innerWidth` already accounts for the scrollbar's footprint and matches `viewport_size`, so no drift occurs. Dropped.
- **Replacing the DOM-extraction script with `page.accessibility.snapshot()`**: the API doesn't exist in the pinned Playwright version, and it never provided the geometry, redaction, or stable-id capabilities this code depends on. Dropped.
- **Inlining `supersedes_scope`**: it's a small, natural extension of an existing same-file convention (`CONTEXT_REFS_KEY`), and inlining wouldn't meaningfully reduce complexity. Dropped.

## Prior review findings

The three findings from the earlier automated `gemini-code-assist` review — the unsafe `assert` in the NAVIGATE branch, the same in the DRAG branch, and the unhandled `StopIteration` risk in element lookup — were each re-verified against the current code and are confirmed fixed in 23f5dec6. The NAVIGATE and DRAG branches now raise explicit `ValueError`s, and element lookup defaults to `None` and raises a named `ComputerTargetNotFoundError`.

## Tests

Ran every suite the PR description lists as validation, in a clean environment — all pass, no failures:

- `tests/core/computer`
- `tests/core/agent/test_browser_tools.py`
- `tests/core/test_context_ref.py`
- `tests/core/test_context_materializer.py`
- `tests/core/agent/test_context.py`
- `tests/core/agent/test_react.py`

src/xagent/core/computer/browser.py
qinxuye: Fixed in 7f66d84f. `BrowserComputerEnvironment` now uses the dedicated manager key `computer:<session_id>`, so it cannot alias the legacy browser-tool session, inherit its headless mode, or resize its page. The focused test also asserts the namespaced manager key.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 7f66d84f. The environment now tracks the concrete `BrowserSession` object. If the manager recreates it, observe rebinds and captures a fresh frame; execute invalidates the old observation and raises `ComputerFrameMismatchError` before dispatching the stale action. Covered by a session-replacement test.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 7f66d84f. The adapter validates `page.url` independently before constructing `ComputerObservation`; unsupported or malformed URLs now degrade to `active_url=None` plus `active_url_unavailable=true`, while screenshot and DOM observation remain usable.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Not changing this in this PR. The claim that these environments are never closed is incorrect: `ComputerTool.teardown()` deterministically closes every environment at task teardown. Step-scoped environments are deliberate so parallel or resumable DAG steps do not share one mutable page/frame stream. The runtime currently exposes no step-finalization lifecycle hook, and adding an adapter-local LRU would be unsafe because it could evict a live or resumable step. If the runtime later adds an explicit step-finalization contract, this tool can consume it; inventing that lifecycle inside this focused adapter is neither a correctness fix nor part of this PR.

src/xagent/core/computer/browser.py
qinxuye: Not applying a localized URL check here. Blocking only `computer.navigate` is not an SSRF boundary: clicks, redirects, page JavaScript, and the existing `browser_navigate` path would still reach the same destinations. That would create the appearance of protection without enforcing it. A real fix belongs at `BrowserSession` network interception, covering every request and redirect (including DNS-rebinding behavior) with explicit local-development semantics. That is a cross-cutting security change and should be reviewed as such; this PR does not claim browser network isolation.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 7f66d84f without adding another incomplete secret-name heuristic. The injected script no longer reads `node.value` at all, so input values—sensitive or otherwise—cannot enter DOM hints. A contract test asserts that the shipped script contains no `node.value` access. The screenshot remains the authoritative visual observation.

src/xagent/core/tools/adapters/vibe/browser_use.py
qinxuye: Not changing this. `computer` is the ordinary browser interaction tool introduced by this slice, not a debug tool, so hiding it behind a second feature flag would contradict the PR's intended contract. Likewise, forcing an allow-all `ComputerActionPolicy` call into the provider-neutral base class would be ceremony, not enforcement: the foundation defines a policy boundary for concrete/product adapters where authorization semantics exist; it does not require the mechanism layer to invent a no-op policy. Confirmation and product authorization remain explicitly outside this PR. We should not expand this precursor slice merely to manufacture a production call site.

src/xagent/core/computer/browser.py
qinxuye: Not changing this. `TYPE` intentionally means text insertion at the current caret; `KEYPRESS` represents physical keyboard-event semantics. `keyboard.type()` is materially worse for Unicode/IME and non-US text and adds per-character overhead. Conflating the two actions to satisfy keydown-driven widgets would regress the primary text-input path; callers that require keyboard events should use `KEYPRESS`.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 7f66d84f. `keypress` now accepts exactly one chord and validates that every key before the final key is a real modifier (`Alt`, `Control`, `Meta`, or `Shift`). Non-modifier sequences such as `ArrowDown+Enter` fail clearly and must be sent as separate actions. Covered by tests.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 7f66d84f. After clicking, `replace_text` now checks `document.activeElement` and proceeds only for an enabled, writable input/textarea or contenteditable element. It raises before select-all or insertion otherwise; the failure path is covered by a focused test.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in 7f66d84f. `ComputerToolArgs` validation now returns the normal structured `ComputerToolResult` error and preserves the existing session/frame id when available, without creating an environment solely for invalid input. Covered by a recovery-context test.

src/xagent/core/computer/browser.py
qinxuye: Not changing this. `wait_for_load_state` is not a correctness condition for arbitrary UI actions: it commonly resolves immediately for SPA/XHR updates, while `networkidle` can hang or add large nondeterministic latency. Explicit navigation already waits for `domcontentloaded`; asynchronous application settling is represented by the existing `WAIT` action. The returned frame is a real observation at capture time, not a promise that the application has reached global quiescence.

src/xagent/core/agent/context/execution.py
qinxuye: The cosmetic empty transcript chunk is fixed in 7f66d84f. The token-cache invalidation is deliberate and remains unchanged: once supersession rewrites message content and removes live image refs, the provider token count describes a different prompt and must not be reused. Preserving that cached count would be knowingly inaccurate; fallback estimation is the correct behavior until a later provider measurement refreshes it.

src/xagent/core/agent/context/execution.py
qinxuye: Fixed in 7f66d84f. Superseded tool messages now carry the explicit structured sentinel `{"success": true, "superseded": true}` instead of `raw_result=None`, so future success classification does not depend on the current compaction invariant. Covered by the supersession test.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,455 @@
...
+        super().__init__(session_id)
+        if viewport_width <= 0 or viewport_height <= 0:
+            raise ValueError("viewport dimensions must be positive")
+        self.workspace = workspace
+        self.manager = manager or get_browser_manager()
+        self.observation_store = observation_store or ObservationStore(workspace)
+        self.headless = headless
+        self.viewport_width = viewport_width
+        self.viewport_height = viewport_height
+        self._manager_session_id = f"computer:{self.session_id}"
```
rogercloud: **Likely regression: the new namespace prefix falls outside the task-level cleanup sweep.**

The old key was the bare `session_id`, which the task-level cleanup path in `browser_use.py` matches via `session_id == task_id or session_id.startswith(f"{task_id}:")`. `f"computer:{session_id}"` no longer matches that prefix check, so this session now depends solely on `teardown()` (already covered) and the shared manager's 30-minute idle sweep as its only remaining safety net — the task-level sweep no longer sees it at all.

Suggestion: keep the namespace suffix inside the task-scoped prefix instead, e.g. `f"{session_id}:computer"`, so it stays a prefix match while still being distinct from any legacy browser-tool key.

src/xagent/core/computer/browser.py
```diff
@@ -0,0 +1,455 @@
...
+            await page.mouse.click(x, y)
+            editable = await page.evaluate(
+                """() => {
+                  const node = document.activeElement;
+                  if (!node) return false;
+                  if (node.isContentEditable) return true;
+                  const tag = String(node.tagName || "").toLowerCase();
+                  return (tag === "input" || tag === "textarea") &&
+                    !node.disabled && !node.readOnly;
+                }"""
```
rogercloud: **`replace_text`'s editable check doesn't look inside iframes.**

This only reads `document.activeElement` on the top-level document. If the click lands on an input inside an iframe, the top-level `activeElement` is the `<iframe>` element itself — not an input/textarea/contenteditable — so the action would reject a perfectly valid editable target any time it's inside a frame. This file already acknowledges multi-frame pages elsewhere (`ELEMENT_EXTRACTION_INCOMPLETE_KEY`), so this isn't just a hypothetical.

Suggestion: when `activeElement` is an iframe, drill into `contentDocument.activeElement` for same-origin frames, and fall back to the original click-only behavior for cross-origin frames rather than hard-failing.

rogercloud (approved):
## Follow-up review

Re-checked the fix commit (7f66d84f, "fix: harden browser computer use loop") against the prior review. All test suites pass (116 tests in the computer/browser area, no failures).

### Confirmed fixed (independently re-verified against the current code, not just taken on the reply's word)

- Session now uses a dedicated manager key (`computer:<session_id>`), so it can no longer alias a legacy browser-tool session, inherit its headless mode, or resize its page.
- `_get_page` now tracks the concrete `BrowserSession` object; `execute()` raises `ComputerFrameMismatchError` *before* dispatching any action if the session was recreated underneath it, and `observe()` transparently rebinds. Verified the ordering — the stale action genuinely never reaches `page.mouse`/`page.keyboard`.
- `active_url` is now pre-validated independently and degrades to `None` + `active_url_unavailable` on an unsupported scheme/length/credentials, instead of blowing up `ComputerObservation` construction and wedging the environment.
- `replace_text` now checks `document.activeElement` for an editable target before select-all + insert.
- `keypress` now rejects non-modifier chords instead of silently mis-executing them.
- Argument validation is now inside the tool's own try/except and preserves the current frame id on failure, instead of bypassing the tool's structured error result.
- The DOM-extraction script no longer reads `node.value` anywhere, closing all three previously-confirmed redaction bypasses categorically (not just narrowing them with another name/id heuristic). Tradeoff: legitimate non-sensitive input values are no longer visible via DOM hints either — the screenshot is the only remaining source for that, which seems like the right call given the alternative.
- Superseded tool messages now carry an explicit `{"success": true, "superseded": true}` sentinel instead of `raw_result=None`, and the empty-string transcript chunk for superseded messages is gone.

Each of the above has a real, passing test behind it (not just a code read) — I ran them individually.

### On the items left unchanged

Five points from the last review were explicitly left as-is with reasoning. Went back through each independently:

- **SSRF on `computer.navigate`, no config gate + unwired policy hook, `insert_text` vs `type()`, fixed 250ms settle** — the reasoning holds up. Worth noting on the SSRF point specifically: the existing `browser_navigate`/`browser_evaluate` tools already allow arbitrary URLs (including `file://`) and arbitrary JS execution with zero validation, so a scheme+host check scoped only to `computer.navigate` really would be more decorative than protective while those sibling tools stay wide open — a real fix belongs at the shared `BrowserSession` layer, not duplicated per-tool. Agreed this isn't a blocker here, but it'd be worth a tracking issue so the metadata-endpoint case (a single, cheap, synchronous check) doesn't fall through the cracks indefinitely.
- **DAG-step environment lifecycle** — this is the one where the reasoning doesn't fully hold up. `teardown()` does deterministically close every environment, but that wasn't the concern — the concern was unbounded accumulation *during* a long-running task. Confirmed: DAG plans have a concurrency cap (default 4) but no cap on total step count, and completed steps have no cleanup hook of their own — only `teardown()` (task-level) and the shared manager's 30-minute idle sweep remove anything. So a wide/long DAG plan can hold several live Chromium processes concurrently for the task's duration, which the reply doesn't really address. Not a merge blocker given the idle-sweep fallback, but worth a tracking issue rather than closing as resolved.

### New, from this fix commit itself

- **A likely regression in the new session-key.** The old key was the bare `session_id`, which the task-level cleanup path in `browser_use.py` matches via `session_id == task_id or session_id.startswith(f"{task_id}:")`. The new `f"computer:{session_id}"` key no longer matches that prefix check, so this session now falls outside the task-level cleanup sweep and depends solely on `teardown()` (which is already covered) and the 30-minute idle timeout as its second line of defense. Simple fix: keep the namespace suffix inside the task-scoped prefix, e.g. `f"{session_id}:computer"`, rather than prefixing the whole thing.
- **`replace_text`'s new editable check doesn't look inside iframes.** It only reads `document.activeElement` on the top-level document. If the click lands on an input inside an iframe, the top-level `activeElement` is the `<iframe>` element itself, which isn't an input/textarea/contenteditable — so the action would incorrectly reject a perfectly valid editable target any time it's inside a frame. Given this file already acknowledges multi-frame pages elsewhere (`ELEMENT_EXTRACTION_INCOMPLETE_KEY`), this isn't just a hypothetical.
- A couple of minor/informational items, not blocking: a keypress/target validation failure that happens *before* any page interaction still invalidates the current frame and costs a re-screenshot (harmless but slightly wasteful); `active_url_unavailable` is a bare string literal rather than a module-level constant like its sibling metadata keys; the new redaction contract test only asserts a `node.value` substring is absent from the script rather than exercising the script's actual output, so an equivalent-but-differently-named read (e.g. `node.defaultValue`) wouldn't be caught.

### Overall

This is a meaningfully improved iteration — the fixes are substantive, not cosmetic, and each is backed by a real test. Approving. Two of the notes above (the session-key regression and the iframe editable-check gap) would be worth a quick follow-up before or shortly after merge, and the DAG-step lifecycle point deserves a tracking issue rather than being considered closed.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 1f291ace. The manager key is now `<task[:step]>:computer`, which remains distinct from legacy browser sessions while preserving the task-id prefix consumed by `_close_task_sessions`. The focused test asserts the exact suffix form.

src/xagent/core/computer/browser.py
qinxuye: Fixed in 1f291ace. The editability check now follows the focused `iframe`/`frame` chain through `contentDocument`, so same-origin nested inputs are verified correctly. I deliberately did not adopt the suggested cross-origin fallback: continuing with select-all + insertion when editability cannot be verified would reintroduce the accidental-input risk this check was added to prevent. Cross-origin frames therefore remain fail-closed instead of silently bypassing validation.

PR #1106, opened 2026-08-02
feat: add local computer use via cua-driver

qinxuye:
## Review boundary

This PR adds one narrowly scoped capability: **Local computer**, meaning an Xagent backend directly controls one explicitly selected application window on the **same host** through `cua-driver mcp`. It builds on the generic Computer Use foundation in #1098 and browser loop in #1104.

Please evaluate this PR as a same-host, window-scoped computer runtime. Browser automation remains a separate runtime. Remote user-device takeover and whole-desktop control are separate transports/scopes and intentionally excluded.

## In scope: what this PR does

- adds a task-scoped `cua-driver` MCP client and a provider-neutral local computer environment
- enumerates visible host application windows in the composer and requires the user to select one exact `(pid, window_id)` before task creation
- persists that exact target with the task and never silently switches applications or windows
- controls browser windows or other desktop application windows through the shared Computer Use action/observation contract
- reports desktop observations, FileRef-backed screenshots, and accessibility elements through the shared schema
- exposes a flat single-action tool contract to avoid nested one-element action-array serialization failures while accepting the preview shape for compatibility
- contributes Local computer through the generic Task Runtime Extension hook, with `image` as a routing preference rather than a hard model requirement
- suppresses the colliding Playwright `computer` family only for tasks that bind Local computer, so the model sees one unambiguous `computer` tool
- adds authenticated readiness/window discovery and a compact composer `+` menu entry
- keeps the feature disabled by default and restricts task creation/readiness probing to administrators
- uses canonical `XAGENT_LOCAL_COMPUTER_*` / `XAGENT_CUA_DRIVER_*` configuration while reading the preview Local browser names for compatibility
- handles the specific provider response where reasoning is mandatory by retrying the same selected model once with thinking enabled

## Browser separation

- High-level web navigation remains part of the Browser runtime.
- Local computer does not expose a browser-specific `navigate` action; it operates the selected application GUI.
- A selected local browser window may still use its existing signed-in state through normal GUI actions, but it is not treated as a tab-scoped browser session.

## Out of scope: what this PR does not do

- no Chrome extension for taking over a tab on a user's laptop
- no WebSocket/Redis browser relay, pairing flow, or other remote-user-device transport
- no macOS Desktop Relay or companion application
- no whole-display capture/control mode in the Xagent product surface
- no native provider-specific Computer Use model API integration
- no automatic `cua-driver` installation, packaging, permission granting, or host lifecycle management
- no attempt to make a cloud Xagent directly control an end user's computer

These excluded transports can integrate later through the generic runtime-extension boundary without being added to this implementation.

## Deployment behavior

- **Local/self-hosted Xagent:** can operate an explicitly selected application window on that same machine, including an existing signed-in browser window.
- **Cloud/server Xagent:** can only operate a window on the cloud/server host. It cannot see or control an end user's laptop through this PR.

## Validation

- focused backend suites for Computer Use, Task Runtime Extension, readiness, routing, and config: all passed
- frontend TypeScript type-check passed
- frontend ChatInput and AppProvider tests: 99 passed
- pre-commit over every changed file passed, including ruff, mypy, frontend lint, and frontend type-check
- real local `cua-driver 0.16.0` readiness probe passed with permissions healthy and 17 visible windows enumerated

### DIFF
--- a/frontend/src/i18n/locales/en.ts
+++ b/frontend/src/i18n/locales/en.ts
@@ -238,10 +238,11 @@ const en = {
         config: "Configure Model",
         upload: "Upload",
       },
-      localBrowser: {
-        label: "Local browser",
+      localComputer: {
+        label: "Local computer",
         chipLabel: "Computer use",
         checking: "Checking the Xagent host...",
+        chooseWindow: "Choose an application window",
         unavailable: "Not available on this Xagent host",
       },
       noModel: "No Model",
--- a/frontend/src/i18n/locales/en.ts
+++ b/frontend/src/i18n/locales/en.ts
@@ -234,8 +234,15 @@ const en = {
       processing: "Processing",
       usingAgentLabel: "Using",
       actions: {
+        add: "Add",
         config: "Configure Model",
         upload: "Upload",
+      },
+      localBrowser: {
+        label: "Local browser",
+        chipLabel: "Computer use",
+        checking: "Checking the Xagent host...",
+        unavailable: "Not available on this Xagent host",
       },
       noModel: "No Model",
       noModelAlert: "No model configured. Please configure a model first.",
--- a/src/xagent/core/computer/cua_driver.py
+++ b/src/xagent/core/computer/cua_driver.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import base64
 import binascii
-import io
 import os
 from collections.abc import Mapping
 from dataclasses import dataclass, field
@@ -15,9 +14,9 @@ from mcp.client.stdio import stdio_client
 from mcp.types import CallToolResult, ImageContent, TextContent
 
 from ...config import (
-    get_browser_cua_driver_command,
-    get_browser_cua_driver_socket,
-    get_browser_cua_driver_timeout_seconds,
+    get_cua_driver_command,
+    get_cua_driver_socket,
+    get_cua_driver_timeout_seconds,
 )
 
 
@@ -62,14 +61,12 @@ class CuaDriverMCPClient:
         socket: str | None = None,
         timeout_seconds: float | None = None,
     ) -> None:
-        self.command = command or get_browser_cua_driver_command()
+        self.command = command or get_cua_driver_command()
         self.socket = (
-            get_browser_cua_driver_socket()
-            if socket is None
-            else socket.strip() or None
+            get_cua_driver_socket() if socket is None else socket.strip() or None
         )
         self.timeout_seconds = (
-            get_browser_cua_driver_timeout_seconds()
+            get_cua_driver_timeout_seconds()
             if timeout_seconds is None
             else timeout_seconds
         )
@@ -220,37 +217,38 @@ class CuaDriverMCPClient:
         )
         failure: BaseException | None = None
         try:
-            async with (
-                stdio_client(parameters, errlog=io.StringIO()) as (
-                    read_stream,
-                    write_stream,
-                ),
-                ClientSession(
-                    read_stream,
-                    write_stream,
-                    read_timeout_seconds=timedelta(seconds=self.timeout_seconds),
-                ) as session,
-            ):
-                await session.initialize()
-                if not ready.done():
-                    ready.set_result(None)
-                while True:
-                    request = await queue.get()
-                    if request is None:
-                        break
-                    if request.future.done():
-                        continue
-                    try:
-                        response = await session.call_tool(
-                            request.name,
-                            request.arguments,
-                        )
-                    except Exception as exc:
-                        if not request.future.done():
-                            request.future.set_exception(exc)
-                    else:
-                        if not request.future.done():
-                            request.future.set_result(response)
+            with open(os.devnull, "w", encoding="utf-8") as errlog:
+                async with (
+                    stdio_client(parameters, errlog=errlog) as (
+                        read_stream,
+                        write_stream,
+                    ),
+                    ClientSession(
+                        read_stream,
+                        write_stream,
+                        read_timeout_seconds=timedelta(seconds=self.timeout_seconds),
+                    ) as session,
+                ):
+                    await session.initialize()
+                    if not ready.done():
+                        ready.set_result(None)
+                    while True:
+                        request = await queue.get()
+                        if request is None:
+                            break
+                        if request.future.done():
+                            continue
+                        try:
+                            response = await session.call_tool(
+                                request.name,
+                                request.arguments,
+                            )
+                        except Exception as exc:
+                            if not request.future.done():
+                                request.future.set_exception(exc)
+                        else:
+                            if not request.future.done():
+                                request.future.set_result(response)
         except FileNotFoundError as exc:
             failure = CuaDriverError(
                 f"cua-driver executable was not found: {self.command!r}"
--- a/src/xagent/core/computer/cua_driver.py
+++ b/src/xagent/core/computer/cua_driver.py
@@ -1,0 +1,273 @@
+from __future__ import annotations
+
+import asyncio
+import base64
+import binascii
+import io
+import os
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from datetime import timedelta
+from typing import Any, Protocol
+
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.stdio import stdio_client
+from mcp.types import CallToolResult, ImageContent, TextContent
+
+from ...config import (
+    get_browser_cua_driver_command,
+    get_browser_cua_driver_socket,
+    get_browser_cua_driver_timeout_seconds,
+)
+
+
+class CuaDriverError(RuntimeError):
+    """Raised when the local cua-driver process cannot satisfy a tool call."""
+
+
+@dataclass(frozen=True)
+class CuaDriverResult:
+    """Provider-neutral subset of one cua-driver MCP tool response."""
+
+    structured: dict[str, Any] = field(default_factory=dict)
+    text: str = ""
+    image_bytes: bytes | None = None
+    image_mime_type: str | None = None
+
+
+class CuaDriverClientProtocol(Protocol):
+    async def call_tool(
+        self,
+        name: str,
+        arguments: Mapping[str, Any] | None = None,
+    ) -> CuaDriverResult: ...
+
+    async def close(self) -> None: ...
+
+
+@dataclass(frozen=True)
+class _ToolCallRequest:
+    name: str
+    arguments: dict[str, Any]
+    future: asyncio.Future[CallToolResult]
+
+
+class CuaDriverMCPClient:
+    """Lazy, task-scoped stdio MCP client for the native cua-driver runtime."""
+
+    def __init__(
+        self,
+        *,
+        command: str | None = None,
+        socket: str | None = None,
+        timeout_seconds: float | None = None,
+    ) -> None:
+        self.command = command or get_browser_cua_driver_command()
+        self.socket = (
+            get_browser_cua_driver_socket()
+            if socket is None
+            else socket.strip() or None
+        )
+        self.timeout_seconds = (
+            get_browser_cua_driver_timeout_seconds()
+            if timeout_seconds is None
+            else timeout_seconds
+        )
+        if not self.command.strip():
+            raise ValueError("cua-driver command must not be empty")
+        if self.timeout_seconds <= 0:
+            raise ValueError("cua-driver timeout must be positive")
+        self._worker: asyncio.Task[None] | None = None
+        self._queue: asyncio.Queue[_ToolCallRequest | None] | None = None
+        self._ready: asyncio.Future[None] | None = None
+        self._lifecycle_lock = asyncio.Lock()
+
+    async def call_tool(
+        self,
+        name: str,
+        arguments: Mapping[str, Any] | None = None,
+    ) -> CuaDriverResult:
+        if not name.strip():
+            raise ValueError("cua-driver tool name must not be empty")
+        queue = await self._ensure_worker()
+        future: asyncio.Future[CallToolResult] = (
+            asyncio.get_running_loop().create_future()
+        )
+        await queue.put(
+            _ToolCallRequest(
+                name=name,
+                arguments=dict(arguments or {}),
+                future=future,
+            )
+        )
+        try:
+            response = await asyncio.wait_for(
+                asyncio.shield(future),
+                timeout=self.timeout_seconds,
+            )
+        except TimeoutError as exc:
+            raise CuaDriverError(
+                f"cua-driver tool {name!r} timed out after "
+                f"{self.timeout_seconds:g} seconds"
+            ) from exc
+        except Exception as exc:
+            raise CuaDriverError(f"cua-driver tool {name!r} failed: {exc}") from exc
+        finally:
+            if not future.done():
+                future.cancel()
+
+        text_parts = [
+            item.text for item in response.content if isinstance(item, TextContent)
+        ]
+        image_parts = [
+            item for item in response.content if isinstance(item, ImageContent)
+        ]
+        message = "\n".join(text_parts).strip()
+        if response.isError:
+            raise CuaDriverError(message or f"cua-driver tool {name!r} failed")
+
+        image_bytes: bytes | None = None
+        image_mime_type: str | None = None
+        if image_parts:
+            image = image_parts[0]
+            try:
+                image_bytes = base64.b64decode(image.data, validate=True)
+            except (binascii.Error, ValueError) as exc:
+                raise CuaDriverError(
+                    f"cua-driver tool {name!r} returned invalid image data"
+                ) from exc
+            image_mime_type = image.mimeType
+
+        structured = response.structuredContent
+        return CuaDriverResult(
+            structured=dict(structured) if isinstance(structured, dict) else {},
+            text=message,
+            image_bytes=image_bytes,
+            image_mime_type=image_mime_type,
+        )
+
+    async def close(self) -> None:
+        async with self._lifecycle_lock:
+            worker = self._worker
+            queue = self._queue
+            self._worker = None
+            self._queue = None
+            self._ready = None
+        if worker is None:
+            return
+        if queue is not None and not worker.done():
+            await queue.put(None)
+        try:
+            await asyncio.wait_for(worker, timeout=self.timeout_seconds)
+        except TimeoutError:
+            worker.cancel()
+            await asyncio.gather(worker, return_exceptions=True)
+        except Exception:
+            # Callers already receive the specific transport failure from their
+            # request future. Teardown remains best-effort and idempotent.
+            pass
+
+    async def _ensure_worker(
+        self,
+    ) -> asyncio.Queue[_ToolCallRequest | None]:
+        async with self._lifecycle_lock:
+            if self._worker is None or self._worker.done():
+                loop = asyncio.get_running_loop()
+                self._queue = asyncio.Queue()
+                self._ready = loop.create_future()
+                self._worker = asyncio.create_task(
+                    self._run_worker(self._queue, self._ready),
+                    name="xagent-cua-driver-mcp",
+                )
+            queue = self._queue
+            ready = self._ready
+        assert queue is not None and ready is not None
+        try:
+            await asyncio.wait_for(
+                asyncio.shield(ready),
+                timeout=self.timeout_seconds,
+            )
+        except TimeoutError as exc:
+            raise CuaDriverError(
+                "cua-driver MCP initialization timed out after "
+                f"{self.timeout_seconds:g} seconds"
+            ) from exc
+        except Exception as exc:
+            if isinstance(exc, CuaDriverError):
+                raise
+            raise CuaDriverError(f"could not initialize cua-driver MCP: {exc}") from exc
+        return queue
+
+    async def _run_worker(
+        self,
+        queue: asyncio.Queue[_ToolCallRequest | None],
+        ready: asyncio.Future[None],
+    ) -> None:
+        args = ["mcp"]
+        if self.socket is not None:
+            args.extend(["--socket", self.socket])
+        parameters = StdioServerParameters(
+            command=self.command,
+            args=args,
+            env={
+                # The MCP SDK inherits only a safe environment allowlist. Keep
+                # driver telemetry disabled unless the host explicitly opted in.
+                "CUA_DRIVER_RS_TELEMETRY_ENABLED": os.getenv(
+                    "CUA_DRIVER_RS_TELEMETRY_ENABLED",
+                    "0",
+                ),
+            },
+        )
+        failure: BaseException | None = None
+        try:
+            async with (
+                stdio_client(parameters, errlog=io.StringIO()) as (
+                    read_stream,
+                    write_stream,
+                ),
+                ClientSession(
+                    read_stream,
+                    write_stream,
+                    read_timeout_seconds=timedelta(seconds=self.timeout_seconds),
+                ) as session,
+            ):
+                await session.initialize()
+                if not ready.done():
+                    ready.set_result(None)
+                while True:
+                    request = await queue.get()
+                    if request is None:
+                        break
+                    if request.future.done():
+                        continue
+                    try:
+                        response = await session.call_tool(
+                            request.name,
+                            request.arguments,
+                        )
+                    except Exception as exc:
+                        if not request.future.done():
+                            request.future.set_exception(exc)
+                    else:
+                        if not request.future.done():
+                            request.future.set_result(response)
+        except FileNotFoundError as exc:
+            failure = CuaDriverError(
+                f"cua-driver executable was not found: {self.command!r}"
+            )
+            failure.__cause__ = exc
+        except asyncio.CancelledError:
+            failure = CuaDriverError("cua-driver MCP worker was cancelled")
+            raise
+        except Exception as exc:
+            failure = CuaDriverError(f"could not initialize cua-driver MCP: {exc}")
+            failure.__cause__ = exc
+        finally:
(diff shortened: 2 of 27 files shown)


src/xagent/core/computer/native_browser.py
```diff
@@ -0,0 +1,831 @@
...
+                "direction": direction,
+                "amount": max(1, min(20, math.ceil(abs(delta) * 10))),
+                "by": "line",
+            }
+            if action.target is not None:
+                arguments.update(self._action_target_arguments(action))
+            await self._call_action("scroll", arguments)
+            return
+        if action.type is ComputerActionType.DRAG:
+            assert action.start is not None and action.end is not None
```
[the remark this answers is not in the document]
qinxuye: Fixed in `39d1d0a4`. The drag path now raises an explicit `ValueError` when either endpoint is missing. I also added a regression test using `model_construct` to bypass the normal Pydantic schema validation, so the defensive runtime path is exercised directly.

src/xagent/core/computer/cua_driver.py
```diff
@@ -0,0 +1,269 @@
...
+                timeout=self.timeout_seconds,
+            )
+        except TimeoutError as exc:
+            future.cancel()
+            raise CuaDriverError(
+                f"cua-driver tool {name!r} timed out after "
+                f"{self.timeout_seconds:g} seconds"
+            ) from exc
+        except Exception as exc:
+            raise CuaDriverError(f"cua-driver tool {name!r} failed: {exc}") from exc
```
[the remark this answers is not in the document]
qinxuye: Fixed in `a23d5c4b`. `call_tool` now cancels the underlying request future from a `finally` block, covering both tool timeouts and cancellation of the caller while retaining `shield` so the worker can observe the cancelled request cleanly.

src/xagent/core/computer/cua_driver.py
```diff
@@ -0,0 +1,269 @@
...
+                    read_timeout_seconds=timedelta(seconds=self.timeout_seconds),
+                ) as session,
+            ):
+                await session.initialize()
+                if not ready.done():
+                    ready.set_result(None)
+                while True:
+                    request = await queue.get()
+                    if request is None:
+                        break
```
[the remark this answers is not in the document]
qinxuye: Fixed in `a23d5c4b`. The worker now skips completed or cancelled request futures before invoking MCP. I also added a deterministic queued-cancellation regression test proving that a stale call never reaches the session.

src/xagent/web/services/local_browser_runtime.py
```diff
@@ -0,0 +1,227 @@
...
+                status_code=403,
+            )
+        if target is not None:
+            _store_task_target(context, target)
+
+    def build_runtime(
+        self,
+        context: TaskRuntimeContext,
+    ) -> TaskRuntimeContribution | None:
+        if not _task_is_bound(context, self.extension_name):
```
rogercloud, line 61: `build_runtime` runs on every execution but only checks task binding — it never re-checks `get_local_computer_enabled()` or that the owner is still an admin, both of which are enforced only in `on_task_created`. A demoted admin keeps getting the computer tool, and flipping the feature flag off is caught only later in `NativeComputerEnvironment.__init__`, after Playwright tools have already been suppressed. Suggest re-running both checks here.

src/xagent/web/services/local_browser_runtime.py
```diff
@@ -0,0 +1,227 @@
...
+        session.close()
+
+
+def _task_target(context: TaskRuntimeContext) -> dict[str, Any] | None:
+    session = context.session_factory()
+    try:
+        task = session.query(Task).filter(Task.id == context.task_id).first()
+        if task is None or int(task.user_id) != context.user_id:
+            return None
+        raw = (task.agent_config or {}).get(_TARGET_AGENT_CONFIG_KEY)
```
rogercloud, line 218: `local_computer_target` is read back as authoritative but is not in `CLIENT_RESERVED_AGENT_CONFIG_KEYS` (task_runtime.py), so a client can pre-seed it and bypass `_validate_target_configuration` entirely. Suggest adding the key to the reserved set and re-validating the payload here before it reaches `NativeComputerEnvironment`.

src/xagent/web/services/local_browser_runtime.py
```diff
@@ -0,0 +1,227 @@
...
+        user = session.query(User).filter(User.id == context.user_id).first()
+        return bool(user is not None and user.is_admin)
+    finally:
+        session.close()
+
+
+def _validate_target_configuration(
+    configuration: Mapping[str, Any],
+) -> dict[str, Any] | None:
+    if not configuration:
```
rogercloud, line 167: An empty configuration returns `None` with no error, so `on_task_created` stores no target and `_select_target` later falls back to the frontmost visible window. That contradicts the PR description and the tool's own "never switch windows silently" instruction. Suggest making the target required rather than degrading silently.

src/xagent/core/computer/native_browser.py
```diff
@@ -0,0 +1,792 @@
...
+            "computer_runtime_kind": "local_computer",
+            "native_driver": "cua-driver",
+            "application": target.app_name,
+            "pid": target.pid,
+            "window_id": target.window_id,
+            "user_takeover_available": True,
+            "delivery_mode": "background",
+            **computer_input_metadata(host_computer_input_platform()),
+            "supported_actions": [action.value for action in _SUPPORTED_ACTIONS],
+            "available_windows": [
```
rogercloud, line 342: This exposes the app name and title of up to 20 unrelated host windows to the model on every observation frame, even though the task is locked to one window it cannot switch away from. Window titles commonly contain document names and private URLs. Also, `_visible_windows` is only populated during the first `_select_target()`, so this list is stale after the first frame. Suggest dropping it once a target is bound.

src/xagent/core/computer/native_browser.py
```diff
@@ -0,0 +1,792 @@
...
+            )
+        mime_type = result.image_mime_type or str(
+            result.structured.get("screenshot_mime_type") or "image/png"
+        )
+        if mime_type not in {"image/png", "image/jpeg", "image/webp"}:
+            raise CuaDriverError(
+                f"cua-driver returned unsupported screenshot type {mime_type!r}"
+            )
+        width, height = self._screenshot_size(result)
+        viewport = Viewport(width=width, height=height, device_pixel_ratio=1.0)
```
rogercloud, line 311: The viewport is built from screenshot pixel dimensions with `device_pixel_ratio=1.0`, while element bounds are normalized against the window's point dimensions from `list_windows`. Point-based actions are then denormalized against this pixel viewport, so on HiDPI displays the round-trip is not unit-consistent. Worth confirming cua-driver's coordinate contract and matching one space.

src/xagent/core/tools/adapters/vibe/browser_tools.py
```diff
@@ -11,12 +11,46 @@
...
     """Create browser automation tools."""
     if not config.get_browser_tools_enabled():
         return []
 
+    # A bound Local computer task contributes its own ``computer`` instance.
+    # Suppress the Playwright browser family as one unit so the runtime tool
+    # does not collide with the core tool or accidentally expose a second,
+    # unrelated browser to the model.
+    contribution = config.get_task_runtime_contribution()
+    if _has_local_computer_runtime(contribution):
```
rogercloud, line 51: This fails open: `build_task_runtime` swallows provider `build_runtime` exceptions and drops the contribution, so a failure here silently gives the task a throwaway Playwright browser instead of the user's selected window, with no user-visible signal. Suggest suppression fail closed when the local-computer extension is bound but its contribution is missing.

src/xagent/core/computer/native_browser.py
```diff
@@ -0,0 +1,792 @@
...
+            if isinstance(raw, Mapping)
+            and (parsed := self._parse_window(raw)) is not None
+        ]
+        visible = [
+            window
+            for window in windows
+            if window.is_on_screen and window.on_current_space is not False
+        ]
+        self._visible_windows = visible
+        if self.target_pid is not None and self.target_window_id is not None:
```
rogercloud, line 221: This exact-match branch compares only `pid`/`window_id`; the `application` name stored by `_store_task_target` is never threaded through here. PIDs and `CGWindowID`s are both recyclable, so a stale pair on a resumed task can bind to a different application — the exact outcome the class docstring says it prevents. Suggest verifying `window.app_name` against the stored `application` and raising `ComputerTargetNotFoundError` on mismatch.

src/xagent/web/services/local_browser_runtime.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def __init__(self, extension_name: str = LOCAL_COMPUTER_TASK_EXTENSION) -> None:
+        self.extension_name = extension_name
+
+    def on_task_created(
+        self,
+        context: TaskRuntimeContext,
+        configuration: Mapping[str, Any],
+    ) -> None:
+        target = _validate_target_configuration(configuration)
```
rogercloud, line 42: Validation runs before the enablement/admin checks, so an unauthorized caller with a malformed config gets a 400 instead of a 403. Suggest authorizing first, then validating.

src/xagent/web/api/computer.py
```diff
@@ -0,0 +1,64 @@
...
+            message="Local computer is disabled on this Xagent host.",
+        )
+        return LocalComputerReadiness(
+            ready=False,
+            connected=False,
+            attached=False,
+            issues=[issue],
+            message=issue.message,
+        )
+    if not bool(user.is_admin):
```
rogercloud, line 45: The not-admin branch returns HTTP 200 with `issues[].code = "not_authorized"` rather than a 403, which lets an authenticated non-admin enumerate whether the feature is enabled on this host from the response shape. Suggest returning 403 here.

src/xagent/core/computer/native_browser_readiness.py
```diff
@@ -0,0 +1,241 @@
...
+_cache = _ReadinessCache()
+
+
+async def get_local_computer_readiness() -> LocalComputerReadiness:
+    """Probe cua-driver and visible application windows with a short cache."""
+
+    now = time.monotonic()
+    if _cache.value is not None and _cache.expires_at > now:
+        return _cache.value.model_copy(deep=True)
+    async with _cache.lock:
```
rogercloud, line 62: The cache lock is held for the full probe duration (bounded by the driver timeout, default 30s), so all concurrent readiness requests serialize behind one slow probe. A shared cache is reasonable here — just flagging the worst-case latency within the 10s TTL window.

src/xagent/core/computer/native_browser_readiness.py
```diff
@@ -0,0 +1,241 @@
...
+
+def reset_local_computer_readiness_cache() -> None:
+    _cache.value = None
+    _cache.expires_at = 0
+
+
+async def _probe_local_computer_readiness() -> LocalComputerReadiness:
+    client = CuaDriverMCPClient()
+    try:
+        health, windows_result = await asyncio.gather(
```
rogercloud, line 80: `asyncio.gather` without `return_exceptions=True`: if one call raises first, the sibling keeps running unobserved and can log "Task exception was never retrieved" at GC time. Suggest `return_exceptions=True` with explicit handling of both results.

src/xagent/core/computer/native_browser.py
```diff
@@ -0,0 +1,792 @@
...
+            y=top / target.height,
+            width=(right - left) / target.width,
+            height=(bottom - top) / target.height,
+        )
+
+    async def _execute_action(self, action: ComputerAction) -> None:
+        if action.type is ComputerActionType.SCREENSHOT:
+            return
+        if action.type is ComputerActionType.WAIT:
+            duration_ms = action.duration_ms or 1_000
```
rogercloud, line 530: This fallback is dead — the schema already requires `duration_ms > 0` for WAIT. Same pattern at :622/:625 (schema forbids `duration_ms` for DRAG, so the literal 500 always governs), :605 (schema forbids `target` for SCROLL), and :169 (the base `execute()` guarantees a non-None observation before `_execute`). Worth trimming for clarity.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -31,59 +33,92 @@
...
     )
-    actions: list[ComputerAction] = Field(
-        default_factory=_initial_screenshot_actions,
-        min_length=1,
-        max_length=1,
-        description=(
-            "One browser action. Coordinates are normalized from 0 to 1. Every "
-            "call returns a fresh observation before another action is planned."
-        ),
+    action: ComputerActionType = Field(
```
rogercloud, line 54: Typing this as the full `ComputerActionType` enum means `navigate` and `move` — which the PR says are deliberately not exposed for local computer — still show up as valid values in the JSON schema the model sees, and are only rejected at runtime. Consider narrowing the type per runtime.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -31,59 +33,92 @@
...
+            for key, item in nested.items():
+                normalized.setdefault("action" if key == "type" else key, item)
+        raw_action = normalized.get("action")
+        if isinstance(raw_action, Mapping):
+            nested = dict(raw_action)
+            normalized.pop("action")
+            for key, item in nested.items():
+                normalized.setdefault("action" if key == "type" else key, item)
+        if "type" in normalized:
+            normalized.setdefault("action", normalized.pop("type"))
```
rogercloud, line 104: `.pop("type")` is evaluated before the `setdefault` check, so if both `action` and a conflicting `type` are supplied, `type`'s value is silently discarded. The `expected_frame_id` conflict path just above raises on mismatch — worth aligning the two.

src/xagent/core/computer/cua_driver.py
```diff
@@ -0,0 +1,271 @@
...
+            text=message,
+            image_bytes=image_bytes,
+            image_mime_type=image_mime_type,
+        )
+
+    async def close(self) -> None:
+        async with self._lifecycle_lock:
+            worker = self._worker
+            queue = self._queue
+            self._worker = None
```
rogercloud, line 150: `_worker`/`_queue` are nulled and the lock released before the old worker is awaited, so a concurrent `call_tool()` in that window would spawn a fresh subprocess rather than wait on the in-flight shutdown. Related: native_browser.py:139 clears `self._driver` before `await driver.close()`, losing the reference if close raises. Neither is reachable today (callers serialize through `_operation_lock`), so this is a latent note.

frontend/src/i18n/locales/en.ts
```diff
@@ -234,9 +234,17 @@ const en = {
       processing: "Processing",
       usingAgentLabel: "Using",
       actions: {
+        add: "Add",
         config: "Configure Model",
         upload: "Upload",
       },
+      localComputer: {
+        label: "Local computer",
+        chipLabel: "Computer use",
```
rogercloud, line 243: `localComputer.chipLabel` is never referenced anywhere in the frontend — the UI uses `localComputer.label`, and `ChatInput.test.tsx` even asserts this raw key does not render. Same key in zh.ts:243. Suggest removing both.

rogercloud (commented):
## Summary

This PR adds a second Computer Use runtime alongside the existing Playwright browser one: a `local_computer` Task Runtime Extension that drives one explicitly-selected application window on the same host as the Xagent backend, through a locally-spawned `cua-driver` MCP subprocess (screenshot, accessibility tree, mouse/keyboard). The composer gains a window picker backed by a new readiness endpoint (`GET /api/computer/local-computer/readiness`); the chosen `(pid, window_id)` is persisted on the task's `agent_config` and re-resolved on every execution. When the extension is bound, the Playwright browser tool family is suppressed so only one browser surface is exposed to the model.

## Re-review status

This is a re-review of the previously-flagged items from the automated review round:

- **Resolved** — drag validation now raises `ValueError` instead of asserting, with regression coverage.
- **Resolved** — queued-but-undispatched request futures are now cancelled on shutdown/timeout. One nuance: this cancels calls still sitting in the queue, not a call already in flight when it times out; that request still runs to completion driver-side. Not counting this as unresolved, just noting the boundary.
- **Resolved** — the worker now skips stale requests before dispatch rather than after.
- **Not applicable** — the earlier "env replacement strips PATH/HOME" concern on `src/xagent/core/computer/cua_driver.py` does not hold: the `mcp` SDK's stdio client merges `get_default_environment()` with the supplied `env` before spawning, so the existing code comment there is accurate. No action needed.

Everything below is new from this pass.

## Findings

### Major

- **1. `build_runtime` never re-checks enablement or admin status** — `src/xagent/web/services/local_browser_runtime.py:57-71`. It only calls `_task_is_bound` (owner + binding exists), while `get_local_computer_enabled()` and the admin check run only in `on_task_created`. Consequences: (a) an admin demoted after task creation keeps receiving `ComputerTool` on every subsequent execution; (b) turning `XAGENT_LOCAL_COMPUTER_ENABLED` off does not stop the tool being offered — the flag is only caught later inside `NativeComputerEnvironment.__init__` (`src/xagent/core/computer/native_browser.py:108`), by which point Playwright tools are already suppressed, leaving the task with no browser capability at all. Suggest re-running both checks inside `build_runtime`.

- **2. `local_computer_target` is not in the client-reserved agent-config key set** — the target is written to `agent_config["local_computer_target"]` (`src/xagent/web/services/local_browser_runtime.py:205`) and read back at `:218`, but that key is absent from `CLIENT_RESERVED_AGENT_CONFIG_KEYS` in `src/xagent/web/services/task_runtime.py`, so `sanitize_client_agent_config` does not strip it from client-supplied config. A client can pre-seed an unvalidated payload, bypassing `_validate_target_configuration`'s positive-integer/allowed-field checks entirely, and `_task_target` performs no validation before the values reach `NativeComputerEnvironment`. Admin-only access limits real exploitability, but it breaks the "anything read back as authoritative must be stripped from the client copy" invariant. Suggest adding the key to `CLIENT_RESERVED_AGENT_CONFIG_KEYS` and re-validating in `_task_target`.

- **3. Empty target configuration silently binds to the frontmost window** — `_validate_target_configuration({})` returns `None` with no error (`local_browser_runtime.py:167-168`), `on_task_created` then stores nothing, `NativeComputerEnvironment.__init__` only rejects a *partial* pair (`native_browser.py:116`), and `_select_target` falls through to `return max(visible, key=... z_index)` (`native_browser.py:249`). `tests/web/services/test_local_browser_runtime.py::test_local_computer_create_requires_enablement_admin_and_valid_target` asserts this succeeds. This contradicts both the PR description ("requires the user to select one exact (pid, window_id) before task creation") and the tool's own prompt text ("never switch windows silently"), while `example.env` quietly documents the actual fallback. Not reachable via the composer UI, but reachable by any admin hitting the task-creation API directly. Suggest rejecting empty/partial configuration in `on_task_created`.

- **4. Titles of up to 20 unrelated host windows are sent to the LLM on every frame** — `src/xagent/core/computer/native_browser.py:342-350`. `metadata["available_windows"]` includes app name and title for `self._visible_windows[:20]`, even though the task is locked to one window the model cannot switch away from. Window titles routinely carry document names, email subjects, and private URLs — an unmitigated leak with no found functional consumer (there's no window-switch action). Compounding it, `_visible_windows` is populated once during the first `_select_target()` (`:220`) and never refreshed, so long tasks report a stale snapshot anyway. Suggest dropping `available_windows` from observation metadata once a target is bound, or at minimum omitting titles.

- **5. Coordinate-space mismatch between element bounds and point actions** — `native_browser.py:311` builds `Viewport` from the screenshot's *pixel* size with `device_pixel_ratio=1.0` hardcoded, while accessibility element bounds are normalized against `target.width/height` from `list_windows` — window *point* dimensions. Point-based coordinates are later denormalized against `observation.viewport.width/height` (`:694`), i.e. back into pixel space. On any HiDPI display where screenshot resolution differs from window points (typically 2x), this round-trip is not unit-consistent, and nothing corrects for it. Element-token clicks (the tool's preferred path) are unaffected; point-based click/drag/replace-text are affected. Test fixtures use non-proportional sizes (1000x800 window vs 1200x800 screenshot), which masks rather than proves correctness for the uniform-scale case. Suggest confirming cua-driver's actual coordinate contract (points vs pixels) and normalizing through one space.

- **6. Playwright suppression fails open on provider errors** — `build_task_runtime` (`src/xagent/web/services/task_runtime.py`) catches any exception from a provider's `build_runtime` — including `RuntimeError("Local computer requires a task workspace")` raised at `local_browser_runtime.py:64` — logs it, and drops that provider's contribution from the merge with nothing propagated to the caller. Since Playwright suppression (`src/xagent/core/tools/adapters/vibe/browser_tools.py:51`) is decided by inspecting the merged contribution, a provider failure means suppression never triggers: the task silently falls back to a temporary Playwright browser instead of the user's selected window, with no user-visible signal anything went wrong. Suggest failing closed instead — if the local-computer extension is bound to the task but its contribution is absent, block browser tools and surface the error rather than silently substituting a different browser.

- **7. Stored `application` name is never used to guard against pid/window_id reuse** — `local_browser_runtime.py:187-191` persists `application`, but it's dropped in `build_runtime` (`:67-71`) and `_select_target`'s exact-match branch (`native_browser.py:221-236`) compares only `pid`/`window_id`. OS PIDs and macOS `CGWindowID`s are both recyclable after the original process/window closes — plausible for a task resumed after idling — so a stale pair can resolve to a different application and the code binds to it silently, violating the class's own documented invariant ("if the window closes, the environment fails instead of silently taking over another application"). Low likelihood, but the consequence is exactly the safety property this feature promises. Suggest verifying `window.app_name` against the stored `application` after the exact match and raising `ComputerTargetNotFoundError` on mismatch.

### Minor

- **8. Validation runs before authorization** — `local_browser_runtime.py:42-53`. `_validate_target_configuration` runs before the enablement/admin checks, so a non-admin sending a malformed config gets a 400 instead of a 403. Only config shape leaks, but suggest authorizing before validating.

- **9. Readiness endpoint returns 200 for unauthorized callers** — `src/xagent/web/api/computer.py:33-59`. Both "feature disabled" and "not admin" return HTTP 200, distinguished only by `issues[].code` in the body, letting an authenticated non-admin enumerate whether the feature is enabled from the response shape. Suggest 403 for the not-admin branch.

- **10. Readiness probe serializes all callers behind one lock** — `native_browser_readiness.py:62-69`. The cache lock is held for the whole probe (bounded by the driver timeout, default 30s). A shared cache is reasonable here since the data isn't user-specific — flagging only the bounded worst-case latency within the 10s TTL window.

- **11. `asyncio.gather` without `return_exceptions=True`** — `native_browser_readiness.py:80-83`. If one call raises first, the sibling task keeps running unobserved and can produce a "Task exception was never retrieved" warning at GC time. Suggest `return_exceptions=True` with explicit handling of both results.

- **12. Dead defensive branches given schema-layer invariants** — `native_browser.py`: `duration_ms or 1_000` at `:530` (schema requires `duration_ms > 0` for WAIT), `duration_ms or 500` at `:622`/`:625` (schema forbids `duration_ms` for DRAG, so the literal always governs), `if action.target is not None` at `:605` (schema forbids `target` for SCROLL), and the `current_observation is None` guard at `:169-172` (the base `execute()` already guarantees non-None). Harmless, but worth trimming.

- **13. `navigate`/`move` still appear in the tool's JSON schema** — `src/xagent/core/tools/adapters/vibe/computer.py:54`. `ComputerToolArgs.action` uses the full `ComputerActionType` enum, so actions the PR says are deliberately not exposed for local computer are still advertised to the model and only rejected at runtime.

- **14. Silent conflict resolution in `_normalize_preview_shapes`** — `computer.py:104`. `.pop("type")` evaluates before the `setdefault` check, so when both `action` and a conflicting `type` are present, `type` is discarded with no error — unlike the `expected_frame_id` conflict path just above (`:88-93`), which raises. Worth aligning.

- **15. Two latent teardown races** — (a) `cua_driver.py:146-152`: `close()` nulls `_worker`/`_queue` and releases its lock before awaiting the old worker's shutdown, so a concurrent `call_tool()` in that window would spawn a fresh subprocess. (b) `native_browser.py:139`: `_close()` clears `self._driver` before `await driver.close()`, so a raise there loses the reference needed to retry cleanup. Neither is reachable today — current callers serialize through `_operation_lock` — so these are latent notes rather than active bugs.

- **16. Unused i18n key** — `frontend/src/i18n/locales/en.ts:243` / `zh.ts:243` add `localComputer.chipLabel`, never referenced (the UI uses `localComputer.label`); `ChatInput.test.tsx` even asserts the raw key string is absent from the DOM. Suggest removing it.

## Testing performed

- Backend focused suites (`tests/core/computer/`, `tests/web/api/test_computer_readiness.py`, `tests/web/services/test_local_browser_runtime.py`, `tests/core/model/chat/basic/test_router.py`, `tests/core/test_config.py`): **165 passed, exit 0**.
- Frontend: `ChatInput.test.tsx` + `app-context-chat.test.tsx`: **99 passed**. TypeScript `tsc --noEmit`: clean.

Coverage gaps worth naming:
- No test exercises route-level authentication (401 for anonymous callers) — the readiness tests call the handler function directly with a stubbed user, bypassing FastAPI dependency injection.
- No test covers a closed/recycled window mid-task (Finding 7).
- No test covers a client-supplied malformed `agent_config.local_computer_target` (Finding 2).
- No test exercises the coordinate math end-to-end (Finding 5); existing fixtures' non-proportional dimensions would mask a uniform-scale bug.
- No test covers the fail-open Playwright fallback path (Finding 6).

## Documentation

`example.env`'s new section describes the frontmost-window fallback but doesn't mention the feature is admin-only. Nothing in the README/docs tells an operator that `cua-driver` needs separate installation or macOS Screen Recording + Accessibility permissions.

qinxuye:
Closing this version because the Local computer implementation has been reworked around explicit perception modes and native-accessibility control, without an implicit browser debugging/CDP path. The replacement branch is still undergoing additional testing and will be proposed separately when ready.

PR #1107, opened 2026-08-02
fix: preserve Home and Build behavior without route overrides

### LINKED ISSUE #1105: Preserve Home and Build behavior without downstream route overrides
## Problem

Downstream distributions can compose the canonical Home and Build pages through the existing extension seams, but several page behaviors still require replacing a complete route or accepting user-visible regressions.

On current `main`, Home couples locale-dependent templates with locale-independent recent tasks, lacks stale-completion cleanup, silently handles task-create failures, keeps the submitted textarea value and expanded height, and owns permissive task-response/title shaping. Build silently handles publish/unpublish failures, prefixes absolute Agent logo URLs with the API base, and formats dates without an invalid-value fallback or the selected locale. Home also hardcodes its Get Started destinations, so an embedding product cannot override or disable them through the existing Home extension seam.

## Desired architecture

- Keep the canonical Home and Build routes as the only page owners.
- Make Home and Build consume one shared task-response validator and prompt-title normalizer, designed for reuse by other task-creation surfaces.
- Build the validator on the existing `parseApiResponse` body boundary. Validate the current required response core before publishing task state.
- Let the frontend title helper trim and collapse whitespace only; persisted title length remains owned by the task-create backend.
- Give templates a locale-generation-scoped loader and recent tasks a mount-scoped loader. Each source owns its request, parsing, empty state, diagnostics, and cleanup independently.
- Preserve localized user-visible errors for Home task creation and Build publish/unpublish failures.
- Put Agent logo URL shaping in one reusable helper and add the strict locale-aware display formatter to the existing `time-utils` ownership area.
- Export typed optional Get Started destination overrides from the existing replaceable Home extension module. Canonical Home owns the OSS defaults and consumes the overrides.
- Do not require downstream route copies, DOM patching, or build-time source rewriting.

### Home and Build task creation

- Both pages parse task-create bodies through `parseApiResponse` and the same validator.
- A valid response requires a positive safe-integer numeric `task_id` plus string `title`, `status`, and `created_at`. `id` aliases, numeric strings, zero, negative, fractional, unsafe, missing, or mistyped required fields are rejected before any task-state action or publication.
- Prompt-derived titles are trimmed and internal whitespace is collapsed. No independent frontend length cap is introduced.
- Home non-OK, rejected, unreadable, empty, or malformed responses show a localized error, publish neither pending-message nor task-ID state, and preserve the textarea value and height.
- After a validated Home success, pending-message and task-ID state are published, then the textarea value is cleared and its computed height is reset.
- Build does not dispatch `RESET_STATE`, `TRIGGER_TASK_UPDATE`, or any other task-state action until model resolution, transport, body parsing, and shared response validation succeed. After validation, it retains the existing successful reset/update flow, modal cleanup, and visible task-create failure behavior.

### Home data lifecycle

- Locale changes re-request templates but do not re-request recent tasks.
- A transport rejection or body-parse failure in either source does not suppress successful data from the other source.
- An old locale generation cannot overwrite a newer template result.
- Cleanup before a response is ready prevents body parsing from starting. If cleanup occurs after parsing has started, parsing may finish but cannot publish state, diagnostics, or other page-owned UI effects.

### Build mutations

- Publish and unpublish each show their matching localized error for non-OK and rejected requests while the mounted page still owns the action.
- A successful mutation starts the existing generation-fenced Agent-list refresh without showing a mutation-failure error.
- A later list-refresh failure is not reclassified as a publish/unpublish failure.
- A mutation completion after page cleanup produces no page-owned toast or state update.

### Display shaping

- `http://` and `https://` Agent logo URLs pass through unchanged. Empty or whitespace-only values use the fallback icon. Relative paths are joined to the API base with exactly one separator for every base/path slash combination.
- The shared display-date formatter returns an empty string for missing or invalid input, accepts the selected locale and explicit `Intl.DateTimeFormatOptions`, preserves Home recent tasks' existing month/day/hour/minute presentation, and supports Build's date-only presentation.
- The formatter is added to the existing `time-utils` ownership area; the unrelated hard-coded `utils.formatDate` is not used for this contract.

### Get Started destinations

- Each destination override is typed as `string | null | undefined`.
- An omitted value uses the canonical OSS default, a string renders that destination, and `null` renders a non-link card without an anchor, tab stop, pointer cursor, or link-only hover styling.
- OSS tests cover defaults, string overrides, nullable overrides, and canonical Home consumption. Embedding distributions verify their concrete values in their own composition tests.

## Test matrix

- Home loaders: transport rejection in each direction; body-parse failure in each direction; `en -> zh -> old en resolves last`; unmount before transport; cleanup after parse starts.
- Home task creation: non-OK, rejected, empty/malformed body, invalid required core, and valid success with task-state/input assertions.
- Build task creation: model-resolution failure, transport failure, unreadable/empty/malformed body, and malformed or mistyped required-core responses dispatch no task-state action, including `RESET_STATE` and `TRIGGER_TASK_UPDATE`, and publish no pending message or task ID; a valid response publishes the validator-returned task ID and performs the existing reset/update workflow only after validation; the request body contains the shared normalized title.
- Publish/unpublish: non-OK, rejection, success-refresh, refresh-failure separation, and late completion after cleanup.
- Logo/date: absolute and relative URL matrix, fallback icon, missing/invalid dates, Build date-only options, Home month/day/hour/minute options, and explicit selected locale.
- Get Started: omitted/default, string override, and `null` non-link behavior.

## Scope

This issue owns canonical Home/Build behavior and the embedder-facing Home extension contract. Other task-creation surfaces, product-specific destination values, and downstream dependency-pin updates remain outside this issue.

## Authoritative implementation invariants

The constraints below refine the acceptance criteria above and take precedence where earlier wording is less specific.

### Task action ownership

- Home and Build allocate a unique monotonic attempt token synchronously before starting model I/O. React pending state is presentation only and is not the single-flight lock.
- Page cleanup invalidates the active attempt. Closing the Build modal also cancels the frontend attempt, clears pending presentation state, and preserves the current draft. A backend task already accepted is not rolled back.
- Every page-owned continuation checks the active attempt after awaited model resolution, task transport, and task-body parsing before starting the next operation or publishing diagnostics, state, toast, or navigation.
- Draft cleanup is governed by a monotonic revision incremented by every edit, including voice-input events. Success clears only the unchanged submitted revision, so an A -> B -> A edit is preserved.
- Operational prepare failures remain separate from the synchronous commit section. A commit programming error is diagnosed but is not reclassified as backend task-create failure, and no transactional rollback is promised.

### Shared model-read contract

- `frontend/src/lib/models.ts` owns the model-list reader, user-default reader, and `resolveTaskLlmSelection()`. Home and Build do not call model endpoints or parse model bodies locally.
- Each create attempt starts exactly two concurrent collection-reader operations through `apiRequest`: `/api/models/?category=llm` and `/api/models/user-default`. This preserves the shared authentication/retry boundary and never performs one request per model or card.
- The user-default response is decoded from its actual array payload into a typed map. The model-list reader validates the complete consumed response shape and enforces that an LLM-filtered response contains only `category: "llm"` records.
- Selection priority is a non-empty `user-default.general` model ID, then the first non-empty model ID in server response order. The current list producer does not emit `is_default`, so no such branch is used and stable backend ordering is not promised.
- The resolver returns one of `success`, `no_model`, or `operational_error`. A valid general candidate wins even if the sibling source fails; no candidates from two valid sources is `no_model`; no candidate plus an observable source failure is `operational_error`.
- The backend currently converts an internal user-default failure to HTTP 200 with an empty array. The frontend cannot distinguish that from a genuine empty source and does not claim otherwise.
- `DefaultModelsSettings` receives only the compile-level signature/field adaptation required by the corrected reader. It remains unmounted. Other direct model readers and deterministic backend ordering remain outside this issue.

### Lifecycle, presentation, and extension boundaries

- Home template and recent-task loaders are independent. Templates use locale-generation ownership; recent tasks use mount-lifetime ownership and are not re-requested on locale changes. Both use pre-parse and post-parse ownership checks and are exercised under React StrictMode.
- Build publication uses one page-owned publish/unpublish operation. Mutation failure remains separate from the generation-fenced Agent-list refresh. Publish/unpublish error copy moves with all current consumers to the neutral `builds.publication` namespace.
- Logo shaping rejects surrounding whitespace, backslashes, ASCII control characters, unsupported schemes, protocol-relative/network-path forms, malformed HTTP(S) authorities, and invalid API bases. Valid absolute HTTP(S) URLs pass through; valid relative paths join the API base with one separator.
- Date shaping accepts only a non-blank valid string and always receives the selected application locale plus explicit Home or Build display options. Invalid Home metadata has no dangling separator; invalid Build dates omit the row.
- The replaceable Home extension module must export typed Get Started destination overrides. `undefined` uses the canonical default, a non-empty string creates a visibly focusable external link, and `null` or invalid blank input renders a non-interactive card without link-only styling.

### Verification boundary

- Tests cover synchronous double-submit, A -> B -> A, page cleanup, all three Build modal-cancel phases, stale finally behavior, model-source resolution order, malformed mixed arrays, filtered URL/category enforcement, real `parseApiResponse` behavior, StrictMode loader replay, publication/refresh separation, URL mutation cases, exact date locale/options, and Get Started accessibility semantics.
- The focused Home/Build CI command explicitly includes every new helper, page, and locale-parity test. Type-check, lint, both regular and standalone Next builds, and repository pre-commit checks are required before delivery.

AlexLiu190625:
## Summary

- make Home and Build use the same model-selection, task-response validation, and prompt-title normalization owners while preserving the two batched model collection requests
- separate Home template/recent-task lifecycles and Build publication/refresh lifecycles so stale or late work cannot publish page-owned state
- centralize Agent logo and locale-aware display-date shaping across the canonical pages
- expose typed Home Get Started destination overrides so embedding distributions can configure or disable links without replacing the route

## Root cause

The canonical Home and Build routes owned parallel task-creation and presentation paths, while several lifecycle and destination behaviors had no reusable contract. That made equivalent behavior drift between the pages and forced embedding distributions to replace a complete route for small configuration differences.

## Impact

Home and Build now retain one canonical page implementation with shared behavioral owners. Task state is published only after validated responses, stale lifecycle completions stay silent, display shaping is consistent, and Get Started links can be overridden or rendered inert through the existing Home extension seam.

## Validation

- `npm run test:home-build-pages -- --reporter=verbose` — 9 files, 406 tests
- `npm run type-check`
- scoped ESLint — 0 errors; 3 existing `no-img-element` warnings
- `npm run build`
- `NEXT_OUTPUT=standalone npm run build`
- repository pre-commit hooks on the exact changed files

Both Next builds completed successfully. They continue to print the repository-level Next/ESLint invalid-options diagnostic that is independently covered by the passing scoped ESLint and type-check commands.

Closes #1105

### DIFF
--- a/frontend/src/app/build/page.tsx
+++ b/frontend/src/app/build/page.tsx
@@ -136,29 +136,50 @@ function BuildsPageContent() {
     }
   }, [])
 
-  const handlePublish = async (agentId: number) => {
+  const publicationOperations = {
+    publish: {
+      suffix: "publish",
+      errorKey: "builds.publication.publishFailed",
+      diagnostic: "Failed to publish agent:",
+    },
+    unpublish: {
+      suffix: "unpublish",
+      errorKey: "builds.publication.unpublishFailed",
+      diagnostic: "Failed to unpublish agent:",
+    },
+  } as const
+
+  const performPublicationMutation = async (
+    agentId: number,
+    kind: keyof typeof publicationOperations,
+  ): Promise<"success" | "failed" | "stale"> => {
+    const operation = publicationOperations[kind]
     try {
-      const response = await apiRequest(`${getApiUrl()}/api/agents/${agentId}/publish`, {
+      const response = await apiRequest(`${getApiUrl()}/api/agents/${agentId}/${operation.suffix}`, {
         method: "POST",
       })
-      if (response.ok) {
-        fetchAgents() // Refresh list
+      if (!isMountedRef.current) return "stale"
+      if (!response.ok) {
+        console.error(operation.diagnostic, response)
+        toast.error(t(operation.errorKey))
+        return "failed"
       }
+      return "success"
     } catch (error) {
-      console.error("Failed to publish agent:", error)
+      if (!isMountedRef.current) return "stale"
+      console.error(operation.diagnostic, error)
+      toast.error(t(operation.errorKey))
+      return "failed"
     }
   }
 
-  const handleUnpublish = async (agentId: number) => {
-    try {
-      const response = await apiRequest(`${getApiUrl()}/api/agents/${agentId}/unpublish`, {
-        method: "POST",
-      })
-      if (response.ok) {
-        fetchAgents() // Refresh list
-      }
-    } catch (error) {
-      console.error("Failed to unpublish agent:", error)
+  const handlePublication = async (
+    agentId: number,
+    kind: keyof typeof publicationOperations,
+  ) => {
+    const outcome = await performPublicationMutation(agentId, kind)
+    if (outcome === "success" && isMountedRef.current) {
+      void fetchAgents()
     }
   }
 
@@ -574,9 +595,9 @@ function BuildsPageContent() {
                                       onClick={(e) => {
                                         e.stopPropagation()
                                         if (agent.status === 'published') {
-                                          handleUnpublish(agent.id)
+                                          void handlePublication(agent.id, "unpublish")
                                         } else {
-                                          handlePublish(agent.id)
+                                          void handlePublication(agent.id, "publish")
                                         }
                                       }}
                                     >
--- a/frontend/src/app/build/page.tsx
+++ b/frontend/src/app/build/page.tsx
@@ -19,7 +19,8 @@ import { useI18n } from "@/contexts/i18n-context"
 import { useApp } from "@/contexts/app-context-chat"
 import { useRouter, useSearchParams } from "next/navigation"
 import { apiRequest, parseApiResponse } from "@/lib/api-wrapper"
-import { getApiUrl } from "@/lib/utils"
+import { getApiUrl, resolveAgentLogoUrl } from "@/lib/utils"
+import { formatDisplayDate } from "@/lib/time-utils"
 import { resolveTaskLlmSelection } from "@/lib/models"
 import { normalizeTaskPromptTitle, parseTaskCreateCore } from "@/lib/task-create"
 import {
@@ -47,7 +48,7 @@ import {
 } from "@/lib/build-page-extension"
 
 function BuildsPageContent() {
-  const { t } = useI18n()
+  const { t, locale } = useI18n()
   const { dispatch, setTaskId, setPendingMessage } = useApp()
   const router = useRouter()
   const searchParams = useSearchParams()
@@ -449,11 +450,6 @@ function BuildsPageContent() {
     }
   }
 
-  const formatDate = (dateString: string) => {
-    const date = new Date(dateString)
-    return date.toLocaleDateString()
-  }
-
   return (
     <div className="flex flex-col h-full bg-background">
       {/* Header */}
@@ -519,7 +515,15 @@ function BuildsPageContent() {
             {/* List */}
             {filteredAgents.length > 0 ? (
               <div className="grid grid-cols-1 md:grid-cols-2 xl:grid-cols-3 gap-4">
-                {filteredAgents.map((agent) => (
+                {filteredAgents.map((agent) => {
+                  const resolvedLogoUrl = resolveAgentLogoUrl(agent.logo_url, getApiUrl())
+                  const createdDate = formatDisplayDate(agent.created_at, locale, {
+                    year: "numeric", month: "numeric", day: "numeric",
+                  })
+                  const updatedDate = formatDisplayDate(agent.updated_at, locale, {
+                    year: "numeric", month: "numeric", day: "numeric",
+                  })
+                  return (
                   <div
                     key={agent.id}
                     className="group relative flex flex-col justify-between space-y-4 rounded-xl border bg-card p-6 shadow-sm transition-all cursor-pointer hover:shadow-md hover:border-primary/50"
@@ -529,8 +533,8 @@ function BuildsPageContent() {
                       <div className="space-y-4">
                         <div className="flex items-start gap-4">
                           <div className="h-10 w-10 shrink-0 rounded-lg bg-primary/10 flex items-center justify-center text-primary overflow-hidden">
-                            {agent.logo_url ? (
-                              <img src={`${getApiUrl()}${agent.logo_url}`} alt={agent.name} className="h-full w-full object-cover" />
+                            {resolvedLogoUrl ? (
+                              <img src={resolvedLogoUrl} alt={agent.name} className="h-full w-full object-cover" />
                             ) : (
                               <Bot className="h-6 w-6" />
                             )}
@@ -635,14 +639,18 @@ function BuildsPageContent() {
 
                     <div className="space-y-4 pt-2">
                       <div className="space-y-1.5">
-                        <div className="flex items-center text-xs text-muted-foreground">
-                          <Calendar className="h-3.5 w-3.5 mr-1.5" />
-                          {t('builds.card.createdAt')}: {formatDate(agent.created_at)}
-                        </div>
-                        <div className="flex items-center text-xs text-muted-foreground">
-                          <Clock className="h-3.5 w-3.5 mr-1.5" />
-                          {t('builds.card.updatedAt')}: {formatDate(agent.updated_at || agent.created_at)}
-                        </div>
+                        {createdDate && (
+                          <div className="flex items-center text-xs text-muted-foreground">
+                            <Calendar className="h-3.5 w-3.5 mr-1.5" />
+                            {t('builds.card.createdAt')}: {createdDate}
+                          </div>
+                        )}
+                        {updatedDate && (
+                          <div className="flex items-center text-xs text-muted-foreground">
+                            <Clock className="h-3.5 w-3.5 mr-1.5" />
+                            {t('builds.card.updatedAt')}: {updatedDate}
+                          </div>
+                        )}
                       </div>
                       <div className="space-y-4" onClick={(e) => e.stopPropagation()}>
                         <BuildAgentCardExtension agentId={agent.id} />
@@ -714,7 +722,8 @@ function BuildsPageContent() {
                       </div>
                     </div>
                   </div>
-                ))}
+                  )
+                })}
               </div>
             ) : (
               <div className="flex flex-col items-center justify-center h-[400px] text-center space-y-4 border rounded-lg bg-muted/10 border-dashed">
--- a/frontend/src/app/build/page.tsx
+++ b/frontend/src/app/build/page.tsx
@@ -18,8 +18,10 @@ import { FeatureEmptyState } from "@/components/ui/feature-empty-state"
 import { useI18n } from "@/contexts/i18n-context"
 import { useApp } from "@/contexts/app-context-chat"
 import { useRouter, useSearchParams } from "next/navigation"
-import { apiRequest } from "@/lib/api-wrapper"
+import { apiRequest, parseApiResponse } from "@/lib/api-wrapper"
 import { getApiUrl } from "@/lib/utils"
+import { resolveTaskLlmSelection } from "@/lib/models"
+import { normalizeTaskPromptTitle, parseTaskCreateCore } from "@/lib/task-create"
 import {
   canDeleteAgent,
   canEditAgent,
@@ -44,43 +46,6 @@ import {
   BuildPageExtensionProvider,
 } from "@/lib/build-page-extension"
 
-interface LlmModel {
-  model_id: string
-  is_default?: boolean
-}
-
-interface DefaultModelRecord {
-  config_type?: "general" | "small_fast" | "visual" | "compact"
-  model?: {
-    model_id?: string
-  } | null
-}
-
-interface TaskCreateResponse {
-  task_id: number
-}
-
-const isLlmModel = (value: unknown): value is LlmModel => {
-  if (!value || typeof value !== "object") {
-    return false
-  }
-
-  const candidate = value as Record<string, unknown>
-  return (
-    typeof candidate.model_id === "string" &&
-    (candidate.is_default === undefined || typeof candidate.is_default === "boolean")
-  )
-}
-
-const isTaskCreateResponse = (value: unknown): value is TaskCreateResponse => {
-  if (!value || typeof value !== "object") {
-    return false
-  }
-
-  const candidate = value as Record<string, unknown>
-  return typeof candidate.task_id === "number"
-}
-
 function BuildsPageContent() {
   const { t } = useI18n()
   const { dispatch, setTaskId, setPendingMessage } = useApp()
@@ -91,6 +56,9 @@ function BuildsPageContent() {
   const isMountedRef = useRef(false)
   const agentListRequestGenerationRef = useRef(0)
   const agentDeleteActionGenerationRef = useRef(0)
+  const activeTaskCreateAttemptRef = useRef<number | null>(null)
+  const taskCreateCounterRef = useRef(0)
+  const draftRevisionRef = useRef(0)
   const [searchTerm, setSearchTerm] = useState("")
   const [agents, setAgents] = useState<Agent[]>([])
   const [loading, setLoading] = useState(true)
@@ -162,6 +130,7 @@ function BuildsPageContent() {
     void fetchAgents()
     return () => {
       isMountedRef.current = false
+      activeTaskCreateAttemptRef.current = null
       agentListRequestGenerationRef.current += 1
       agentDeleteActionGenerationRef.current += 1
     }
@@ -339,106 +308,105 @@ function BuildsPageContent() {
     setIsCreateModalOpen(true)
   }
 
-  const resolveTaskLlmIds = async (): Promise<[string, string | null, string | null, string | null] | null> => {
-    const apiUrl = getApiUrl()
-    const [modelsResponse, defaultResponse] = await Promise.all([
-      apiRequest(`${apiUrl}/api/models/?category=llm`, { headers: {} }),
-      apiRequest(`${apiUrl}/api/models/user-default`, { headers: {} }),
-    ])
-
-    let allModels: LlmModel[] = []
-    if (modelsResponse.ok) {
-      const modelsData = await modelsResponse.json()
-      if (Array.isArray(modelsData)) {
-        allModels = modelsData.filter(isLlmModel)
-      }
-    }
-
-    const defaultModels: Record<string, string | undefined> = {}
-    if (defaultResponse.ok) {
-      const defaultsData = await defaultResponse.json()
-      if (Array.isArray(defaultsData)) {
-        defaultsData.forEach((defaultConfig: DefaultModelRecord) => {
-          if (defaultConfig?.config_type && defaultConfig.model?.model_id) {
-            defaultModels[defaultConfig.config_type] = defaultConfig.model.model_id
-          }
-        })
-      }
-    }
-
-    const generalModelId =
-      defaultModels.general ||
-      allModels.find((model) => model.is_default)?.model_id ||
-      allModels[0]?.model_id
-
-    if (!generalModelId) {
-      return null
-    }
-
-    return [
-      generalModelId,
-      defaultModels.small_fast ?? null,
-      defaultModels.visual ?? null,
-      defaultModels.compact ?? null,
-    ]
-  }
-
   const handleBuildWithPrompt = async () => {
-    const prompt = createPrompt.trim()
-    if (!prompt || isStartingTask) return
+    const rawPrompt = createPrompt
+    const prompt = rawPrompt.trim()
+    if (!prompt || activeTaskCreateAttemptRef.current !== null) return
+
+    const attempt = ++taskCreateCounterRef.current
+    activeTaskCreateAttemptRef.current = attempt
+    const revision = draftRevisionRef.current
+    const isCurrent = () => isMountedRef.current && activeTaskCreateAttemptRef.current === attempt
 
     setIsStartingTask(true)
     try {
-      dispatch({ type: "RESET_STATE" })
-      const llmIds = await resolveTaskLlmIds()
-      if (!llmIds) {
-        toast.error(t("chatPage.input.noModelAlert"))
+      let task = null
+
+      try {
+        const selection = await resolveTaskLlmSelection()
+        if (!isCurrent()) return
+
+        if (selection.kind === "no_model") {
+          toast.error(t("chatPage.input.noModelAlert"))
+          return
+        }
+        if (selection.kind === "operational_error") throw selection.error
+
+        const taskResponse = await apiRequest(`${getApiUrl()}/api/chat/task/create`, {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify({
+            title: normalizeTaskPromptTitle(rawPrompt),
+            description: prompt,
+            llm_ids: selection.llmIds,
+          }),
+        })
+
+        if (!isCurrent()) return
+        const parsed = await parseApiResponse(taskResponse)
+        if (!isCurrent()) return
+
+        task = taskResponse.ok ? parseTaskCreateCore(parsed.data) : null
+        if (!task) throw new Error("Task create response is invalid")
+      } catch (error) {
+        if (isCurrent()) {
+          console.error("Failed to start task from build modal:", error)
+          toast.error(t("builds.list.createModal.startTaskFailed"))
+        }
         return
       }
 
-      const taskResponse = await apiRequest(`${getApiUrl()}/api/chat/task/create`, {
-        method: "POST",
-        headers: {
-          "Content-Type": "application/json",
-        },
-        body: JSON.stringify({
-          title: prompt,
-          description: prompt,
-          llm_ids: llmIds,
-        }),
-      })
+      if (!isCurrent() || !task) return
 
-      if (!taskResponse.ok) {
-        throw new Error("Failed to create task")
+      try {
+        if (!isCurrent()) return
+        dispatch({ type: "RESET_STATE" })
+
+        if (!isCurrent()) return
+        setPendingMessage({
+          message: prompt,
+          files: [],
+          targetTaskId: task.taskId,
+        })
+
+        if (!isCurrent()) return
+        dispatch({ type: "TRIGGER_TASK_UPDATE" })
+
+        if (!isCurrent()) return
+        setTaskId(task.taskId)
+
+        if (!isCurrent()) return
+        setIsCreateModalOpen(false)
+
+        if (!isCurrent()) return
+        if (draftRevisionRef.current === revision) {
+          draftRevisionRef.current += 1
+          setCreatePrompt("")
+        }
+      } catch (error) {
+        if (isCurrent()) console.error("Failed to commit task creation:", error)
       }
-
-      const taskData = await taskResponse.json()
-      if (!isTaskCreateResponse(taskData)) {
-        throw new Error("Task create response is missing task_id")
-      }
-
-      setPendingMessage({
-        message: prompt,
-        files: [],
-        targetTaskId: taskData.task_id,
-      })
-      dispatch({ type: "TRIGGER_TASK_UPDATE" })
-      setTaskId(taskData.task_id)
-      setIsCreateModalOpen(false)
-      setCreatePrompt("")
-    } catch (error) {
-      console.error("Failed to start task from build modal:", error)
-      toast.error(t("builds.list.createModal.startTaskFailed"))
-      setIsCreateModalOpen(true)
-      setCreatePrompt(prompt)
     } finally {
-      setIsStartingTask(false)
+      if (isCurrent()) {
+        activeTaskCreateAttemptRef.current = null
+        setIsStartingTask(false)
+      }
     }
   }
 
+  const handleCreateModalOpenChange = (open: boolean) => {
+    if (!open) {
+      activeTaskCreateAttemptRef.current = null
+      setIsStartingTask(false)
+    }
+    setIsCreateModalOpen(open)
+  }
+
   const handleManualCreate = () => {
+    handleCreateModalOpenChange(false)
     router.push("/build/new")
-    setIsCreateModalOpen(false)
   }
 
   const createPromptVoiceInputLabel =
@@ -778,7 +746,7 @@ function BuildsPageContent() {
         onOpenChange={(open) => { if (!open) setTriggersAgent(null) }}
       />
 
-      <Dialog open={isCreateModalOpen} onOpenChange={setIsCreateModalOpen}>
+      <Dialog open={isCreateModalOpen} onOpenChange={handleCreateModalOpenChange}>
         <DialogContent className="sm:max-w-[550px] gap-0 p-0 overflow-hidden bg-background shadow-lg rounded-xl">
           <DialogHeader className="px-6 py-5 border-b pr-10">
             <DialogTitle className="flex items-start sm:items-center gap-2 text-xl font-semibold">
@@ -809,7 +777,7 @@ function BuildsPageContent() {
                   ref={createPromptRef}
                   data-voice-input="false"
                   value={createPrompt}
-                  onChange={(e) => setCreatePrompt(e.target.value)}
+                  onChange={(e) => { draftRevisionRef.current += 1; setCreatePrompt(e.target.value) }}
                   placeholder={t("builds.list.createModal.placeholder")}
                   className="min-h-[100px] flex-1 resize-none border-0 shadow-none focus-visible:ring-0"
                   onKeyDown={(e) => {
--- a/frontend/src/app/page.test.tsx
+++ b/frontend/src/app/page.test.tsx
@@ -1,9 +1,14 @@
 import React from "react"
-import { cleanup, render, screen, waitFor } from "@testing-library/react"
+import { act, cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 
 const apiRequestMock = vi.hoisted(() => vi.fn())
+const resolveTaskLlmSelectionMock = vi.hoisted(() => vi.fn())
 const homeExtensionRenderMock = vi.hoisted(() => vi.fn())
+const setPendingMessageMock = vi.hoisted(() => vi.fn())
+const setTaskIdMock = vi.hoisted(() => vi.fn())
+const routerPushMock = vi.hoisted(() => vi.fn())
+const toastErrorMock = vi.hoisted(() => vi.fn())
 
 async function createHomeExtensionMock() {
   const ReactModule = await vi.importActual<typeof import("react")>("react")
@@ -22,13 +27,21 @@ vi.mock("@/lib/api-wrapper", async () => {
   return { ...actual, apiRequest: apiRequestMock }
 })
 
+vi.mock("@/lib/models", () => ({
+  resolveTaskLlmSelection: resolveTaskLlmSelectionMock,
+}))
+
 vi.mock("@/lib/utils", async () => {
   const actual = await vi.importActual<typeof import("@/lib/utils")>("@/lib/utils")
   return { ...actual, getApiUrl: () => "http://api.local" }
 })
 
+vi.mock("@/components/ui/sonner", () => ({
+  toast: { error: toastErrorMock },
+}))
+
 vi.mock("next/navigation", () => ({
-  useRouter: () => ({ push: vi.fn() }),
+  useRouter: () => ({ push: routerPushMock }),
 }))
 
 vi.mock("next/link", () => ({
@@ -50,8 +63,8 @@ vi.mock("@/contexts/i18n-context", () => ({
 
 vi.mock("@/contexts/app-context-chat", () => ({
   useApp: () => ({
-    setTaskId: vi.fn(),
-    setPendingMessage: vi.fn(),
+    setTaskId: setTaskIdMock,
+    setPendingMessage: setPendingMessageMock,
   }),
 }))
 
@@ -79,17 +92,77 @@ vi.mock("@/lib/home-page-extension", createHomeExtensionMock)
 
 import Home from "./page"
 
-function jsonResponse(data: unknown) {
+const successfulSelection = {
+  kind: "success" as const,
+  llmIds: ["general", null, null, null] as [string, null, null, null],
+}
+
+function jsonResponse(data: unknown, init?: ResponseInit) {
   return new Response(JSON.stringify(data), {
     status: 200,
     headers: { "Content-Type": "application/json" },
+    ...init,
   })
 }
 
+function deferred<T>() {
+  let resolve!: (value: T | PromiseLike<T>) => void
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((resolvePromise, rejectPromise) => {
+    resolve = resolvePromise
+    reject = rejectPromise
+  })
+  return { promise, resolve, reject }
+}
+
+function input(): HTMLTextAreaElement {
+  return screen.getByPlaceholderText("home.hero.searchPlaceholder")
+}
+
+function submitButton(): HTMLButtonElement {
+  const button = input().parentElement?.querySelector("button:not([aria-label])")
+  if (!(button instanceof HTMLButtonElement)) throw new Error("Home submit button not found")
+  return button
+}
+
+function typePrompt(value: string) {
+  fireEvent.input(input(), { target: { value } })
+}
+
+function submitWithEnter() {
+  fireEvent.keyDown(input(), { key: "Enter" })
+}
+
+function taskCore(taskId = 7) {
+  return {
+    task_id: taskId,
+    title: "created task",
+    status: "running",
+    created_at: "2026-01-01T00:00:00Z",
+  }
+}
+
 describe("Home", () => {
+  let consoleErrorMock: ReturnType<typeof vi.spyOn>
+
+  function expectNoTaskCreatePublication() {
+    expect(setPendingMessageMock).not.toHaveBeenCalled()
+    expect(setTaskIdMock).not.toHaveBeenCalled()
+    expect(routerPushMock).not.toHaveBeenCalled()
+    expect(toastErrorMock).not.toHaveBeenCalled()
+    expect(consoleErrorMock).not.toHaveBeenCalled()
+  }
+
   beforeEach(() => {
     apiRequestMock.mockReset()
+    resolveTaskLlmSelectionMock.mockReset()
     homeExtensionRenderMock.mockClear()
+    setPendingMessageMock.mockReset()
+    setTaskIdMock.mockReset()
+    routerPushMock.mockReset()
+    toastErrorMock.mockReset()
+    consoleErrorMock = vi.spyOn(console, "error").mockImplementation(() => undefined)
+    resolveTaskLlmSelectionMock.mockResolvedValue(successfulSelection)
     apiRequestMock.mockImplementation((url: string) => {
       if (url.startsWith("http://api.local/api/templates/")) {
         return Promise.resolve(jsonResponse([]))
@@ -101,7 +174,10 @@ describe("Home", () => {
     })
   })
 
-  afterEach(() => cleanup())
+  afterEach(() => {
+    consoleErrorMock.mockRestore()
+    cleanup()
+  })
 
   it("renders a hook-bearing configured extension as one component", async () => {
     render(<Home />)
@@ -109,10 +185,7 @@ describe("Home", () => {
     expect(screen.getByRole("button", { name: "voiceInput.start" })).toBeInTheDocument()
     const extension = await screen.findByTestId("home-extension")
     expect(extension).toBeInTheDocument()
-    expect(extension.parentElement).toHaveAttribute(
-      "data-slot",
-      "home-page-extension",
-    )
+    expect(extension.parentElement).toHaveAttribute("data-slot", "home-page-extension")
     expect(extension.parentElement).toHaveClass("shrink-0")
     expect(screen.getAllByTestId("home-extension")).toHaveLength(1)
     expect(homeExtensionRenderMock).toHaveBeenCalled()
@@ -124,7 +197,6 @@ describe("Home", () => {
     vi.resetModules()
     try {
       const { default: DefaultHome } = await import("./page")
-
       const { container } = render(<DefaultHome />)
       const slot = container.querySelector('[data-slot="home-page-extension"]')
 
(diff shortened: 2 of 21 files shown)


frontend/src/lib/models.ts
```diff
@@ -91,32 +95,254 @@ export interface DefaultModelConfig {
...
+  typeof value === 'number' && Number.isSafeInteger(value) && value > 0;
+
+const nullableString = (value: unknown): value is string | null =>
+  value === null || typeof value === 'string';
+
+const nullableFinite = (value: unknown): value is number | null =>
+  value === null || (typeof value === 'number' && Number.isFinite(value));
+
+const nullableInt = (value: unknown): value is number | null =>
+  value === null || (typeof value === 'number' && Number.isInteger(value));
```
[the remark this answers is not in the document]
AlexLiu190625: Both producers of these payloads always emit every key, so `undefined` is not a reachable value here. `/api/models/?category=llm` serializes through `response_model=List[ModelWithAccessInfo]` (`src/xagent/web/schemas/model.py:352`) with FastAPI defaults, which emit all fields and encode `None` as JSON `null`. `/api/models/user-default` builds its dicts by hand in `ModelStore.serialize_model_with_access`, `_default_model_payload`, and `_shared_fallback_model_payload` (`src/xagent/web/services/model_store.py`), each of which writes every key unconditionally.

Validating the complete consumed response shape is an acceptance criterion of #1105, and `models.test.ts` ("rejects every missing or wrong consumed ModelWithAccessInfo field") pins omitted-field rejection per field on purpose: an omitted key from these producers means the contract drifted, and the reader should fail observably through the resolver's `operational_error` path rather than silently default it.

frontend/src/lib/models.ts
```diff
@@ -91,32 +95,254 @@ export interface DefaultModelConfig {
...
+      abilities: abilities === null ? null : [...abilities],
+      description,
+      created_at,
+      updated_at,
+      is_active,
+      is_owner,
+      can_edit,
+      can_delete,
+      is_shared,
+    });
```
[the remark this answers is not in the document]
AlexLiu190625: `abilities` cannot arrive omitted from either producer: `ModelWithAccessInfo` serializes all fields (`None` becomes JSON `null`), and the hand-built dicts in `ModelStore.serialize_model_with_access` / `_shared_fallback_model_payload` always write the key. Accepting `undefined` would loosen the complete-shape validation that #1105 requires and that `models.test.ts` pins with explicit per-field omitted/`undefined` rejection vectors, turning a producer contract drift into a silent default instead of an observable failure.

frontend/src/lib/models.ts
```diff
@@ -91,32 +95,254 @@ export interface DefaultModelConfig {
...
+
+      result[item.config_type] = {
+        id: item.id,
+        user_id: item.user_id,
+        model_id: item.model_id,
+        config_type: item.config_type,
+        created_at: item.created_at,
+        updated_at: item.updated_at,
+        model: models[0],
+      };
```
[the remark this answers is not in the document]
AlexLiu190625: `_default_model_payload` (`src/xagent/web/services/model_store.py`) writes `created_at` and `updated_at` unconditionally, emitting `null` when the column is empty, so the omitted case is unreachable from the actual producer. Defaulting `undefined` to `null` here would mask a contract drift; the reader intentionally validates the complete payload shape (#1105 acceptance criteria) and the omitted-field rejection is pinned by tests.

frontend/src/app/page.tsx
```diff
@@ -20,74 +20,227 @@ import {
...
+
+  const connections: HomeTemplateConnection[] = [];
+  for (const connection of value.connections) {
+    if (
+      !isJsonRecord(connection) ||
+      typeof connection.name !== "string" ||
+      (typeof connection.logo !== "string" && connection.logo !== null)
+    ) return null;
+    connections.push({ name: connection.name, logo: connection.logo });
+  }
```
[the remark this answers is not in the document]
AlexLiu190625: The templates producer cannot omit `logo`: `/api/templates/` serializes through `response_model=list[TemplateInfo]`, and `ConnectionInfo.logo` is `Optional[str] = Field(default=None)` (`src/xagent/web/api/templates.py:75`), which FastAPI emits as `"logo": null`. The decoder pins that contract deliberately — where a producer genuinely can omit fields, the corresponding decoder accepts omission (see the producer-omitted Agent field handling in `page.test.tsx`), so the strict/lenient split here is per-producer by design, not an oversight.

frontend/src/app/page.tsx
```diff
@@ -20,74 +20,227 @@ import {
...
+import { normalizeTaskPromptTitle, parseTaskCreateCore } from "@/lib/task-create";
 import { useI18n } from "@/contexts/i18n-context";
 import { useApp } from "@/contexts/app-context-chat";
 import { WelcomeModal } from "@/components/welcome-modal";
 import { getBrandingFromEnv } from "@/lib/branding";
 import { useVoiceInputControls } from "@/components/voice-input-controller";
-import { HomePageExtension } from "@/lib/home-page-extension";
+import {
+  HomePageExtension,
+  homeGetStartedDestinationOverrides,
```
rogercloud: **Major:** this static import of the new named value export `homeGetStartedDestinationOverrides` breaks the documented module-replacement contract. The pre-existing contract (base 4e73692) was that this module exports ONLY `HomePageExtension` — the base page imported only the component. An embedding distribution that replaces this module with a file exporting only the old surface now fails: ESM-style replacement → Next/webpack "export not found" build error; CJS-style → undefined import → TypeError at the `.docs/.guides/.whatsNew` reads (~:517-519) on render. No test covers an old-surface replacement (page.test.tsx:21-29 always mocks both exports). Recommendation: make the config an optional capability (e.g. namespace read with a `{}` fallback when the export is missing), or add a compatibility test where a replacement exporting only the old surface still renders.

frontend/src/app/page.test.tsx
```diff
@@ -124,17 +393,1313 @@ describe("Home", () => {
...
+    view.unmount()
+    await act(async () => body.resolve(JSON.stringify(taskCore())))
+
+    expect(setPendingMessageMock).not.toHaveBeenCalled()
+    expect(setTaskIdMock).not.toHaveBeenCalled()
+    expect(toastErrorMock).not.toHaveBeenCalled()
+    expect(consoleErrorMock).not.toHaveBeenCalled()
+  })
+
+  it.each([
```
rogercloud, line 1042: **Minor:** the task-create failure matrix (non-OK/empty/malformed/invalid-core/unreadable) has no mounted transport-rejection case — the only create-fetch rejection in the suite is the unmounted silent path (:851-868). Failure cases also don't per-case assert textarea value+height preservation (adjacent tests at :800-801 and :1023-1024 do cover draft preservation on other failure paths, which mitigates this). Suggest adding a mounted rejection case and per-case value/height assertions.

frontend/src/app/build/page.tsx
```diff
@@ -809,7 +807,7 @@ function BuildsPageContent() {
                   ref={createPromptRef}
                   data-voice-input="false"
                   value={createPrompt}
-                  onChange={(e) => setCreatePrompt(e.target.value)}
+                  onChange={(e) => { draftRevisionRef.current += 1; setCreatePrompt(e.target.value) }}
```
rogercloud, line 810: **Minor:** the new draft-revision invariant ("every edit including voice transcription increments the revision") is not covered through the voice path. Both page tests stub the voice controller with no DOM write/event dispatch (page.test.tsx:102-109, build/page.test.tsx:82-89), while production relies on the real controller's native setter + bubbled input/change (voice-input-controller.tsx:194-208) reaching this handler. If that dispatch contract regresses, a successful submit would clear a voice-modified draft while the A→B→A revision tests still pass. Latent coverage gap, not a live bug; one test driving transcription through the real event path would pin it.

frontend/src/lib/page-extension-contracts.ts
```diff
@@ -3,6 +3,12 @@ import type { ComponentType, ReactNode } from "react"
 // Embedding distributions may replace default implementation modules during frontend composition.
 export type HomePageExtensionComponent = ComponentType
 
+export interface HomeGetStartedDestinationOverrides {
```
rogercloud, line 26: **Minor:** this is the only contract in the file without a doc comment, and the semantics of `docs?: string | null` are not discoverable from the type: `undefined` keeps the OSS default destination, `null` hides the card, a non-empty string passes verbatim (resolution at page.tsx:158-165, behavior pinned by page.test.tsx). Embedding distributions reading the public contract cannot know these three states without reverse-engineering the resolver. Add JSDoc covering them.

rogercloud (approved):
## Follow-up review — PR #1107 (head 6e27bf4a)

### Summary
This PR preserves Home/Build behavior without requiring downstream route overrides. It unifies model-selection and task-response validation behind strict per-producer decoders, fences attempt/revision lifecycle state on both pages (frontend/src/app/page.tsx:290-369, frontend/src/app/build/page.tsx:333-427), scopes the template loader by locale while the recent-task loader stays mount-scoped, separates mutations from refresh on Build (140-185), and adds display shaping for connection logos and dates plus typed Get Started destination overrides (resolved at page.tsx:158-165, 515-579). Backend producers were verified to always emit every consumed key, so the strict frontend decoders match the actual wire contract.

### Update since last review
One commit landed since 967e3990: `6e27bf4a fix(frontend): keep display-date locale map open for widened Locale unions`. It widens the locale map in frontend/src/lib/time-utils.ts to `Partial<Record<Locale, string>>` with a `??` fallback (+6/-2) and pins the behavior with 14 new test lines (time-utils.test.ts). The change is consistent with the earlier display-shaping work and closes the type friction the widened `Locale` union introduced; it does not affect any previously reviewed finding.

### Prior findings (gemini-code-assist bot) — all four WAIVED
All four bot findings remain unfixed in code (the author deliberately declined), but every factual claim in the author's rebuttals was verified accurate against backend sources at head. Verdict: **WAIVED** for all four — the strictness is the intended, test-pinned contract, and the suggested leniencies would be dead flexibility given what the producers actually emit.

- **P1 — frontend/src/lib/models.ts:143 (nullableString/nullableFinite/nullableInt reject undefined): WAIVED.** All three producers emit every key unconditionally: `serialize_model_with_access` writes all 19 keys (src/xagent/web/services/model_store.py:60-86), `_default_model_payload` writes 7 keys (:467-478), `_shared_fallback_model_payload` writes 19 keys (:484-505); every response is validated through `ModelWithAccessInfo` (src/xagent/web/api/model.py:595, 641-645, 1378, 1447; the `response_model` is at :598). In Pydantic v1, non-defaulted `Optional` fields are required input keys, so omission fails server-side before the frontend ever sees the payload. models.test.ts:178-228 pins per-field undefined rejection deliberately. No reachable failure path.
- **P2 — frontend/src/lib/models.ts:221 (validAbilities undefined): WAIVED.** `abilities` is a required `Optional` field in the schema (schemas/model.py:364), the store dicts always include it (model_store.py:72, 495; api/model.py:585), and there is no `exclude_none` anywhere in the web layer. JSON cannot represent `undefined` at all, so the suggestion contradicts the PR's own tested contract.
- **P3 — frontend/src/lib/models.ts:258 (created_at/updated_at `?? null`): WAIVED.** `_default_model_payload` writes both keys unconditionally (model_store.py:467-482), and the suggested `?? null` would be a no-op anyway: the `nullableString` guard already narrows the value to `string | null` before construction. Drift detection lives in the guard, not the assignment.
- **P4 — frontend/src/app/page.tsx:92 (connection logo undefined): WAIVED.** `ConnectionInfo.logo` is `Optional[str] = Field(default=None)` (src/xagent/web/api/templates.py:75) with `response_model=list[TemplateInfo]` (:232) and no `exclude_none` anywhere in src/xagent/web, so the wire always carries `"logo": string | null`. Leniency exists exactly where producers omit keys (chat.py:4126-4128 agent fields), and the strict branch is test-pinned (page.test.tsx:1309, 1327, 1379-1391, 1666).

### New findings

**A1 (major) — frontend/src/app/page.tsx:33-36, frontend/src/lib/home-page-extension.tsx:8 — new required export breaks the documented replacement contract.** Home now statically imports the new named value export `homeGetStartedDestinationOverrides`. The pre-existing replacement contract (base 4e73692) was that the module exports ONLY `HomePageExtension`, and the header comment in page-extension-contracts.ts ("Embedding distributions may replace default implementation modules during frontend composition") is unchanged and does not document the new required export; no migration note exists anywhere. Failure modes for old-contract replacements: ESM-style → Next/webpack "export not found" build error; CJS-style → `undefined` import → TypeError at page.tsx:517-519 on render. Tests always mock both exports (page.test.tsx:21-29), so no test covers an old-surface replacement. Recommendation: make the config an optional capability (namespace read with `{}` fallback when the export is missing), or add a compatibility test where a replacement exporting only the old surface still renders.

**A2 (minor) — frontend/src/app/page.test.tsx:946-986 — mounted transport-rejection case missing from the create-failure matrix.** The task-create failure matrix has no mounted transport-rejection case; the only create-fetch rejection in the suite is the unmounted silent path (:851-868). Failure cases assert toast/no-publication but not per-case textarea value+height preservation (adjacent tests at :800-801 and :1023-1024 do cover draft preservation on other failure paths). The gap is real but low-risk: a mounted non-OK response exercises the same branch. Suggest adding a mounted rejection case and per-case value/height assertions.

**A3 (minor) — frontend/src/app/page.test.tsx:102-109, frontend/src/app/build/page.test.tsx:82-89 — voice transcription path excluded from the draft-revision invariant.** The voice controllers are stubbed with no DOM write/event dispatch, so the new invariant "every edit including voice transcription increments draftRevisionRef" is not covered. Production relies on the real controller's native setter + bubbled input/change (voice-input-controller.tsx:194-208) reaching the page handlers (page.tsx:440-441 onInput, build/page.tsx:809-810 onChange). If that dispatch contract regresses, a successful submit would clear a voice-modified draft while the A→B→A revision tests still pass. Latent coverage gap, not a live bug; one test driving transcription through the real event path would pin it.

**A4 (minor) — six test ranges read production source text and lock implementation shape.** page.test.tsx:688-728 (sourceSlice + variable-name/regex/occurrence-count assertions on contracts/page/home-page-extension/package.json), page.test.tsx:1642-1703 (loader statement order, `return {` count, negative regexes), build/page.test.tsx:348-360 (exact statements + `formatDisplayDate` x1), build/page.test.tsx:921-937 (`void fetchAgents()` present, `catch`/`await fetchAgents()` absent), models.test.ts:550-563 (import-shape/identifier-absence), translations.test.ts:126-144 (key occurrence counts), plus time-utils.test.ts:82-83 (readFileSync/toContain). False-green is real (e.g. a `.then(onRejected)` rewrite passes `not.toContain('catch')`); false-red on equivalent refactors. The tests are deliberately named source-locked and defend real invariants, with extensive behavioral coverage coexisting — hence minor. Suggest keeping only the narrow behavioral locks or moving the shape checks to an import-boundary lint.

**A5 (minor) — frontend/src/lib/page-extension-contracts.ts:6-10 — public contract lacks JSDoc.** `HomeGetStartedDestinationOverrides` is the only contract in this file without a doc comment, and the semantics of `docs?: string | null` are not obvious: `undefined` keeps the OSS default destination, `null` hides the card, and a non-empty untrimmed string passes verbatim (resolution at page.tsx:158-165). Embedders reading the public contract cannot know these states without reverse-engineering the resolver. Add JSDoc covering the three states.

- `delete:` page.tsx:149-150 — `configuredHomeGetStartedDestinationOverrides` re-binds the imported `homeGetStartedDestinationOverrides` with the same type and no transformation; use the import directly.
- `shrink:` page.tsx:291-292 — `const rawPrompt = content;` is a dead alias; use `const submittedPrompt = content.trim()`. (build/page.tsx:334 is a different pattern with genuine reuse — leave it.)
- `shrink:` page.tsx:192-194, 236 — `templateGenerationRef` is observably equivalent to the effect-local `active` flag (React guarantees cleanup before the next setup; the ref is only incremented in setup, with no interleaving where the counter rejects what `active` accepts). Drop the counter and keep `const isCurrent = () => active`; the source-pattern assertions at page.test.tsx:1655-1672 must be updated in the same change.
- `shrink:` frontend/src/components/settings/default-models.tsx:95-109 — `getModelCategory`/`getModelAbilities`/`getModelDisplayName`/`getModelProviderLabel` are pure module-local field forwards (22 call sites); inline as `model.category` / `model.abilities || []` / `model.model_name` / `model.model_provider` (~16 lines + shorter expressions).

net: ~25-30 lines possible.

### Tests
`npm run test:home-build-pages`: 9 files, 407 tests, all passed.

### Verdict
Acceptable with reservations: A1 is the reservation — the new required export should either be made optional (namespace read with `{}` fallback) or covered by an old-surface compatibility test before downstream embedders adopt the contract.

frontend/src/app/page.tsx
AlexLiu190625: Fixed in 4e365a05 with the optional-capability shape: the page now reads the overrides through the module namespace with an empty-object fallback (`homePageExtensionModule.homeGetStartedDestinationOverrides ?? {}`), and the contract header documents `HomePageExtension` as the only required export. Verified empirically with a real `next build` against a replacement module exporting only the old surface: webpack downgrades the missing namespace member to an "Attempted import error" warning (documented in the contract comment for warnings-as-errors distributions), the page prerenders statically without throwing, and wrong-typed replacement values still fail `tsc` because the widening cast is a plain `as`, not `as unknown as`. An old-surface compatibility test renders Home with every canonical default destination; one note there: vitest's mock proxy cannot represent a genuinely absent export (it throws on unknown keys), so the test pins the equivalent present-but-`undefined` shape, while the genuinely-absent case is covered by the build-level behavior above.

frontend/src/app/page.test.tsx
AlexLiu190625: Fixed in 4e365a05: added a mounted create transport-rejection case asserting the failure toast, zero navigation/publication, and textarea value + height preservation, and extended the existing mounted failure matrix (non-OK/empty/malformed/invalid-core/unreadable) with the same per-case value/height assertions. Mutation-checked: reintroducing a height reset in the failure catch turns all the extended cases red.

frontend/src/app/build/page.tsx
AlexLiu190625: Fixed in 4e365a05: both pages now have a test that drives transcription through the production mechanism — the native `HTMLTextAreaElement.prototype.value` setter plus the same bubbled `InputEvent("input")`/`Event("change")` dispatch `voice-input-controller.tsx` performs — then completes a stale successful submit and asserts the transcribed draft survives. Mutation-checked: removing the draft-revision increment from either page's handler turns the new tests red. The helper deliberately reproduces only the write+dispatch mechanism, not caret splicing, and says so in its comment.

frontend/src/lib/page-extension-contracts.ts
AlexLiu190625: Fixed in 4e365a05: `HomeGetStartedDestinationOverrides` now documents the three states against the resolver's exact behavior — `undefined`/omitted key/omitted export uses the canonical default; `null`, a non-string, or a string that is blank after trimming renders a non-interactive card without an anchor, tab stop, pointer cursor, or link-only hover styling; a string that is non-empty after trimming is used verbatim, including surrounding whitespace, which is deliberately not trimmed from the emitted href.

frontend/src/lib/models.ts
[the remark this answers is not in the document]
AlexLiu190625: This file is unchanged since this exact finding was triaged in the previous round (`git diff 967e3990..4e365a05 -- frontend/src/lib/models.ts` is empty). The producers cannot omit these keys: `/api/models/` serializes through `response_model=List[ModelWithAccessInfo]` with FastAPI defaults (every field emitted, `None` → JSON `null`), and `/api/models/user-default` builds its dicts by hand with every key written unconditionally (`model_store.py`: `serialize_model_with_access`, `_default_model_payload`, `_shared_fallback_model_payload`). Strict complete-shape validation is an acceptance criterion of #1105 and is deliberately pinned per-field by `models.test.ts`. The maintainer review verified these facts against backend sources and waived this finding; the strictness stands.

frontend/src/lib/models.ts
```diff
@@ -91,32 +95,254 @@ export interface DefaultModelConfig {
...
+      created_at,
+      updated_at,
+      is_active,
+      is_owner,
+      can_edit,
+      can_delete,
+      is_shared,
+    } = item;
+    const validAbilities = abilities === null
+      || (Array.isArray(abilities) && abilities.every((ability) => typeof ability === 'string'));
```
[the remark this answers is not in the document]
AlexLiu190625: Unchanged code, previously triaged and waived: `abilities` is written unconditionally by every producer path (`model_store.py:72,495`; `ModelWithAccessInfo` has no `exclude_none` anywhere in the web layer), so an omitted key is unreachable from the real wire contract, and rejecting it is the intended drift detector.

frontend/src/lib/models.ts
```diff
@@ -91,32 +95,254 @@ export interface DefaultModelConfig {
...
+      id,
+      model_id,
+      category,
+      model_provider,
+      model_name,
+      base_url,
+      temperature,
+      context_window,
+      dimension,
+      abilities: abilities === null ? null : [...abilities],
```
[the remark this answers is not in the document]
AlexLiu190625: The spread is unreachable with `undefined`: `validAbilities` runs before construction and rejects anything that is not `null` or a string array, returning `null` for the whole payload. Combined with the producer facts above (the key is always emitted), there is no runtime `TypeError` path. Unchanged code, same verdict as the prior round.

frontend/src/app/build/page.tsx
```diff
@@ -339,106 +330,105 @@ function BuildsPageContent() {
...
+        activeTaskCreateAttemptRef.current = null
+        setIsStartingTask(false)
+      }
+    }
+  }
+
+  const handleCreateModalOpenChange = (open: boolean) => {
+    if (!open) {
+      activeTaskCreateAttemptRef.current = null
       setIsStartingTask(false)
```
rogercloud, line 424: **Major (new regression):** closing the create modal while the create request is in flight nulls `activeTaskCreateAttemptRef`, so a successful response is silently discarded (no toast/nav/state update) and the re-entry guard is also removed, allowing a duplicate concurrent create on reopen. The pre-PR `isStartingTask` guard did not have this gap. Suggested fix: only fence state writes on modal close, keep the attempt token alive until the request settles (or explicitly abort it).

frontend/src/app/build/page.tsx
```diff
@@ -162,34 +131,56 @@ function BuildsPageContent() {
...
     try {
-      const response = await apiRequest(`${getApiUrl()}/api/agents/${agentId}/publish`, {
+      const response = await apiRequest(`${getApiUrl()}/api/agents/${agentId}/${operation.suffix}`, {
         method: "POST",
       })
-      if (response.ok) {
-        fetchAgents() // Refresh list
+      if (!isMountedRef.current) return "stale"
+      if (!response.ok) {
+        console.error(operation.diagnostic, response)
```
rogercloud, line 164: **Major (PR-introduced inconsistency):** this branch never reads the response body (`error.detail`) before showing the generic `builds.publication.publishFailed`/`unpublishFailed` toast, and logs the raw `Response` object rather than status/detail. `agent-builder.tsx`'s publish/unpublish handlers use the same i18n keys (renamed by this PR) but do read `error.detail` — same user-facing text, different diagnostics depending on which surface fails.

frontend/src/lib/models.ts
```diff
@@ -91,32 +95,254 @@ export interface DefaultModelConfig {
...
+  | { kind: 'success'; llmIds: [string, string | null, string | null, string | null] }
+  | { kind: 'no_model' }
+  | { kind: 'operational_error'; error: Error }
+
+export async function resolveTaskLlmSelection(): Promise<TaskLlmSelection> {
+  const [models, defaults] = await Promise.allSettled([
+    getUserModels({ category: 'llm' }),
+    getUserDefaultModels(),
+  ]);
+  const defaultMap = defaults.status === 'fulfilled' ? defaults.value : null;
```
rogercloud, line 313: Minor, pre-existing (the prior `resolveTaskLlmIds` had the same gap — not a regression): if `getUserDefaultModels()` rejects here while the model list succeeds, `defaultMap` silently becomes `null`, so `small_fast`/`visual`/`compact` resolve to `null` in the success path below and the rejection reason is never logged anywhere. Worth closing now that this is a shared, reusable resolver.

frontend/src/lib/time-utils.ts
```diff
@@ -2,6 +2,28 @@
...
+  value: unknown,
+  locale: Locale,
+  options: Intl.DateTimeFormatOptions,
+): string {
+  if (typeof value !== "string" || value.trim() === "") return ""
+
+  const date = new Date(value)
+  if (!Number.isFinite(date.getTime())) return ""
+
+  return new Intl.DateTimeFormat(displayDateLocales[locale] ?? locale, options).format(date)
```
rogercloud, line 25: Minor, not currently reachable: `new Intl.DateTimeFormat(displayDateLocales[locale] ?? locale, options)` throws `RangeError` for a malformed BCP-47 tag (verified with `'ja_JP'`), with no try/catch — inconsistent with this function's own "return `''` for invalid input" contract (line 14). Not reachable with the current narrow `Locale` union (both `en`/`zh` map to valid tags), but the code comment above already anticipates a future widened `Locale` hitting this.

frontend/src/app/page.tsx
```diff
@@ -20,74 +20,228 @@ import {
...
+}
+
+function decodeRecentTask(value: unknown): RecentTask | null {
+  if (
+    !isJsonRecord(value) ||
+    !isSafeInteger(value.task_id) ||
+    value.task_id <= 0 ||
+    typeof value.title !== "string" ||
+    (value.created_at !== undefined && value.created_at !== null && typeof value.created_at !== "string") ||
+    (value.agent_name !== undefined && typeof value.agent_name !== "string") ||
```
rogercloud, line 123: Minor, latent: `agent_name` only accepts `undefined` here while `agent_logo_url` (next line) also accepts `null`, and one rejected element voids the entire recent-tasks array (line 125). Not exploitable today (`GET /api/chat/tasks` only ever omits both fields together), but `POST /api/chat/task/create`'s response model already emits `agent_name: null` for the same no-agent case, so the two backend conventions already disagree.

frontend/src/app/page.tsx
```diff
@@ -20,74 +20,228 @@ import {
...
+        if (!response.ok) throw new Error(`Recent task request failed: ${response.status}`);
+
+        const parsed = await parseApiResponse(response);
+        if (!active) return;
+        const decoded = decodeRecentTasks(parsed.data);
+        if (!decoded) throw new Error("Invalid recent task response");
+
+        setRecentTasks(decoded);
+      } catch (error) {
+        if (active) console.error("Failed to fetch recent tasks:", error);
```
rogercloud, line 237: Minor: this catch only logs and never resets `recentTasks` to `[]`, unlike the sibling templates loader's catch just above (line 211: `setTemplates([])`). Benign today (mount-once, starts empty) but the two loaders diverge for no stated reason.

frontend/src/app/page.tsx
```diff
@@ -20,74 +20,228 @@ import {
...
+  guides: "https://docs.xagent.co/models/overview",
+  whatsNew: "https://docs.xagent.co/release-notes",
+}
+
+function resolveHomeGetStartedDestination(
+  configured: unknown,
+  defaultDestination: string,
+): string | null {
+  if (configured === undefined) return defaultDestination
+  if (typeof configured !== "string" || configured.trim().length === 0) return null
```
rogercloud, line 166: Minor: no scheme/protocol validation here, unlike `resolveAgentLogoUrl` in `lib/utils.ts` which explicitly rejects `javascript:` and friends. The resolved value is used directly as `href={card.link!}` a few hundred lines down; `target="_blank" rel="noopener noreferrer"` does not mitigate a `javascript:` href. Low exploitability since this is build-time-trusted config, but worth the same rigor as the logo helper, or at least documenting in the contract that it's unvalidated.

frontend/src/lib/time-utils.ts
```diff
@@ -2,6 +2,28 @@
...
+/** Formats a non-blank, parseable display-date string, returning empty text for invalid input. */
+export function formatDisplayDate(
+  value: unknown,
+  locale: Locale,
+  options: Intl.DateTimeFormatOptions,
+): string {
+  if (typeof value !== "string" || value.trim() === "") return ""
+
+  const date = new Date(value)
+  if (!Number.isFinite(date.getTime())) return ""
```
rogercloud, line 23: Minor, latent: `Number.isFinite(new Date(value).getTime())` accepts loose JS date parses — verified `formatDisplayDate("5", "en", {...})` returns `"5/1/2001"` instead of `""`. The acceptance criteria says only valid non-blank strings should format; a stricter guard (e.g. requiring an ISO-ish shape) would match that more precisely. Latent only since the backend always sends ISO-8601.

frontend/src/app/page.tsx
```diff
@@ -20,74 +20,228 @@ import {
...
+  }
+  return tasks;
+}
+
+// `homeGetStartedDestinationOverrides` is an OPTIONAL export of the replaceable
+// home-page-extension module: a replacement that only implements the required
+// `HomePageExtension` export must still build, so this is read through the
+// module namespace with a fallback rather than a static named import.
+const homeGetStartedDestinationOverrides: HomeGetStartedDestinationOverrides =
+  (homePageExtensionModule as { homeGetStartedDestinationOverrides?: HomeGetStartedDestinationOverrides })
```
rogercloud, line 152: Minor: this `as {...}` cast bypasses structural type-checking of the replacement module's export shape — a replacement exporting the wrong type (e.g. `{docs: 42}`) would compile silently. Only the runtime `typeof` check inside `resolveHomeGetStartedDestination` (below) prevents an actually wrong-shaped replacement from breaking. A narrow runtime type guard would give the same flexibility with real checking.

frontend/src/components/settings/default-models.tsx
```diff
@@ -89,64 +89,47 @@ const modelTypeConfig = {
   },
 }
 
-const defaultModelTypes = Object.keys(modelTypeConfig) as DefaultModelType[]
+type SettingsDefaultModelType = Exclude<DefaultModelType, 'rerank'>
```
rogercloud, line 92: Minor: the backend's `DEFAULT_MODEL_CONFIG_TYPES` includes `rerank` and the frontend's own `currentBackendDefaultModelTypes` mirrors it, so a rerank default is fetched, validated, and stored into the map — but this `Exclude<DefaultModelType,'rerank'>` + cast makes it silently unrenderable in the Settings UI. Pre-existing gap, newly type-visible; the cast papers over it rather than resolving it.

frontend/src/app/build/page.tsx
```diff
@@ -339,106 +330,105 @@ function BuildsPageContent() {
...
-          llm_ids: llmIds,
-        }),
-      })
+      if (!isCurrent() || !task) return
 
-      if (!taskResponse.ok) {
-        throw new Error("Failed to create task")
-      }
+      try {
+        if (!isCurrent()) return
```
rogercloud, line 386: shrink: five of the six `if (!isCurrent()) return` checks in this block (386, 389, 396, 399, 402, 405) guard a run of purely synchronous statements (dispatch / setPendingMessage / dispatch / setTaskId / setIsCreateModalOpen) with zero `await` between them, so the ref can't change mid-run — only the first check (line 383) is load-bearing. Suggest collapsing to that single check and dropping the rest (~-6 lines).

frontend/src/app/page.tsx
```diff
@@ -134,103 +288,91 @@ export default function Home() {
...
-            message: content,
-            files: [],
-            targetTaskId: parsedTaskId
-          });
-
-          setTaskId(parsedTaskId);
+      if (!isCurrent() || !task) return;
+
+      try {
+        if (!isCurrent()) return;
```
rogercloud, line 345: shrink: three `if (!isCurrent()) return` checks here (345, 352, 355) guard synchronous setPendingMessage/setTaskId/draft-clear calls with zero `await` between them. Suggest collapsing to the single check already at line 342 and dropping the rest (~-3 lines).

rogercloud (changes requested):
## PR Summary

This PR replaces the ad-hoc, per-route request/response handling on the Home (`frontend/src/app/page.tsx`) and Build (`frontend/src/app/build/page.tsx`) screens with shared, validated primitives: task-creation model selection and model-list/user-default parsing move into `frontend/src/lib/models.ts` (`parseModelList`, `parseUserDefaultModels`, `resolveTaskLlmSelection`), display-date shaping into `frontend/src/lib/time-utils.ts` (`formatDisplayDate`), and agent logo URL shaping into `resolveAgentLogoUrl` in `frontend/src/lib/utils.ts`. It also adds structural decoding at both route boundaries (`decodeHomeTemplates`, `decodeRecentTasks`, `parseApiResponse`, `parseTaskCreateCore`) so malformed payloads fail closed instead of rendering a misleading empty state, adds generation/lifecycle fencing to the async loaders so late responses from unmounted or superseded effects can't write state, and introduces an optional `homeGetStartedDestinationOverrides` hook on the Home page-extension contract so downstream distributions can retarget Get Started cards without patching the route. Closes #1105.

## Update Summary

One commit (`4e365a05`) landed since the last review at `6e27bf4a`. It fixes A1 by making `homeGetStartedDestinationOverrides` an *optional* export read through a namespace import, restoring the module-replacement contract for old-surface replacement modules (with a new compatibility test and contract-header documentation); adds the missing mounted create-transport-rejection test (A2) and voice-transcription draft-revision tests on both pages (A3); and adds JSDoc for `HomeGetStartedDestinationOverrides` (A5). Unclaimed but confirmed: all four simplification items from the prior round were also applied. A4 (test source-text/regex assertions) was not addressed.

## Approach Verdict

**acceptable-with-reservations.** The shared-lib extraction itself is correct, correctly typed, and consistent with the existing `frontend/src/lib/` architecture. The reservations are about scope and idiom, not correctness:

- **The "unification" stops at the two route files while the real duplicates live elsewhere.** `frontend/src/components/build/agent-builder.tsx:730-795,1127-1132` (the Build editor's own preview task-create path) still has independent inline two-request model selection, and it diverges *semantically* — it keys on numeric `model.id` instead of `model_id`, and takes `availableModels[0]` unconditionally instead of the first entry with a non-empty `model_id`. This PR touched that exact file (three i18n key renames) and left the duplicate in place. Same story for `frontend/src/components/chat/ChatInput.tsx:481-528` and `frontend/src/components/config-dialog.tsx:80-125` (both still branch on a non-existent `is_default` field that isn't on `ModelWithAccessInfo`) and `frontend/src/components/pages/models.tsx:304-330`.
- **`resolveTaskLlmSelection`/`parseUserDefaultModels` is the fix for a live bug it wasn't applied to.** `frontend/src/components/kb/knowledge-base-creation-dialog.tsx:232-243` and `frontend/src/components/kb/knowledge-base-detail.tsx:316-338` read `/api/models/user-default` as an object map (`defaultData.embedding?.model?.model_id`) while the endpoint returns a bare list — so the user's embedding default is *never* honored on those two KB screens. Pre-existing, not introduced here, but this PR authored exactly the parser that fixes it.
- **"Logo shaping in one helper" / "date shaping in one formatter" holds only for the two touched routes.** Five other logo-URL builders remain hand-rolled and vulnerable to the same `javascript:`/protocol-relative inputs the new helper defends against: `agent-builder.tsx:2008` (no scheme check at all — again, a file this PR touched), `frontend/src/components/layout/sidebar.tsx:906`, `frontend/src/components/chat/ChatStartScreen.tsx:260`, `frontend/src/app/task/[id]/page-client.tsx:54`, `frontend/src/components/widget/public-agent-chat-page.tsx:374`. Nine other display-date call sites still do bare `new Date(x).toLocaleString()` with no invalid-date guard. And `frontend/src/lib/utils.ts:166-176` `formatDate()` (hardcoded `zh-CN`) now has zero importers — dead code a centralization change should have removed.
- **Home moved off a shared type rather than onto one.** It now uses a page-private `HomeTemplateCard` instead of `Template` from `@/types/template` — which is still consumed unvalidated by `frontend/src/app/templates/page.tsx`, `ChatStartScreen.tsx`, and `TemplateQuickAccess.tsx`. That diverges from every other template consumer instead of converging.
- **Lifecycle fencing is hand-rolled four different ways.** No `AbortController` anywhere; the four loaders use `let active`, `isMountedRef` + generation counter, and an attempt counter respectively. One small shared helper would collapse four shapes into one.
- **Fail-closed is applied at whole-collection granularity.** `decodeHomeTemplates`/`decodeRecentTasks`/`parseModelList`/`parseUserDefaultModels` each return `null` if *any single element* is malformed, so one bad row hides an entire section. Deliberate per the acceptance criteria, but a forward-compat trap worth revisiting.

Two acceptance criteria are only partially met as a result: "reusable by other task-creation surfaces" is true in principle but zero other surfaces consume it; "Home/Build don't call model endpoints or parse model bodies locally" is true for the two route files and false for `agent-builder.tsx` if "Build" includes the editor. A third, "invalid Build dates omit the row", holds for the `formatDisplayDate` → `""` path but not for the `RangeError` path (see finding below), where nothing renders at all.

## Prior Findings Status

- **P1-P4 (gemini-bot, `models.ts` nullable-parsing helpers rejecting `undefined`) — WAIVED, unchanged, still valid.** The waiver rationale (intentional strictness; all backend producers emit every key unconditionally) was re-verified. `4e365a05` did not touch `models.ts` or the connection-logo validation in `page.tsx`. No action needed.
- **A1 (major) — new required export broke the module-replacement contract — FIXED.** `page.tsx` now uses `import * as homePageExtensionModule` and reads the field optionally (`?? {}`); an old-surface replacement module exporting only `HomePageExtension` builds and renders all canonical defaults. New compatibility test at `app/page.test.tsx:443-481`; contract headers documented in both contract files.
- **A2 (minor) — missing mounted create-transport-rejection test — FIXED.** New test at `app/page.test.tsx:1015-1040` asserts toast, zero navigation/publication, and textarea value+height preservation. Mutation check confirmed: reintroducing a height reset in the failure catch turns the new tests red.
- **A3 (minor) — voice-transcription path excluded from the draft-revision invariant — FIXED.** Both page test files now have a `simulateVoiceTranscription` helper using the real native-setter + bubbled-event mechanism, plus a per-page test asserting a stale successful submit doesn't clear a voice-modified draft. Mutation check confirmed: removing the `draftRevisionRef` increment turns them red.
- **A5 (minor) — public contract type lacked JSDoc — FIXED.** `HomeGetStartedDestinationOverrides` now documents all three states, verified to match the resolver including the "href is not trimmed" detail.
- **A4 (minor) — tests lock implementation shape via source-text/regex assertions — NOT FIXED, still open.** All cited locations remain, unchanged in kind. Notably `app/build/page.test.tsx:921-944` added a comment acknowledging the exact false-green risk, then responded by tightening the regex rather than removing the source-text dependency.

## New Findings

### Major

1. **`frontend/src/app/build/page.tsx:421-427` — `handleCreateModalOpenChange` orphans an in-flight task create, and drops the re-entry guard. Genuine new regression introduced by this PR.** Closing the create modal while `/api/chat/task/create` is in flight nulls `activeTaskCreateAttemptRef`, so the subsequent `isCurrent()` check fails and the successful response is discarded: no toast, no navigation, no state update — the server-side task exists but the user never learns about it. Nulling the ref also removes the re-entry guard, so reopening the modal and submitting again fires a *second* concurrent create. The pre-PR code did not have this: `isStartingTask` was never reset on modal close and correctly blocked re-entry until the in-flight request settled. Suggested fix: keep the attempt token alive until the request settles (only fence state writes on modal close, not the request-completion path), or explicitly abort the request.

2. **`frontend/src/app/build/page.tsx:159-175` vs `frontend/src/components/build/agent-builder.tsx:1603-1663` — publish/unpublish failures now share an i18n key but not behavior. Genuine PR-introduced inconsistency.** This PR renamed `agent-builder.tsx`'s keys to `builds.publication.publishFailed`/`unpublishFailed` so both surfaces show the same string, but only `agent-builder.tsx` reads `error.detail` out of the response body. The new `performPublicationMutation` in `build/page.tsx` never reads the body and always shows the generic message, and it logs the raw `Response` object rather than status/detail — so identical user-facing text now hides materially different diagnostics depending on which surface failed.

### Minor / nits

3. `frontend/src/lib/models.ts:308-346` (`resolveTaskLlmSelection`) — pre-existing, not a regression. If the user-default fetch fails while the model list succeeds, `small_fast`/`visual`/`compact` defaults are silently dropped and `defaults.reason` is never logged. Verified the pre-PR `resolveTaskLlmIds` had the identical gap — not new, but the refactor was a natural place to close it.
4. `frontend/src/app/build/page.tsx:94-127` (agent-list loader) — pre-existing, outside this PR's diff. Unvalidated `any` flows into state; `agent.name.toLowerCase()` with no null-check would throw during render on a malformed item; a non-OK/failed fetch renders the same "no agents yet" empty state as a genuinely empty list. Verified untouched by the diff, so out of scope, but this PR added `parseApiResponse`/`parseTaskCreateCore`-style validation a few hundred lines away that this loader could reuse.
5. `frontend/src/lib/time-utils.ts:25` (`formatDisplayDate`) — real robustness gap, not currently reachable. `new Intl.DateTimeFormat(tag)` throws `RangeError: Invalid language tag` for a malformed tag (verified with `'ja_JP'`), no try/catch — inconsistent with the function's own documented "return `''` for invalid input" contract. Today `Locale` is the narrow `"en"|"zh"` union so both map to valid tags; the code carries a comment acknowledging this risk for a future widened `Locale`.
6. `frontend/src/app/page.tsx:122-124` (`decodeRecentTask`) — latent forward-compat risk, not currently reachable. `agent_logo_url` accepts `null` but `agent_name` accepts only `undefined`, and one rejected element voids the entire recent-tasks list. `GET /api/chat/tasks` only ever omits both fields together today, so it isn't exploitable — but the sibling `POST /api/chat/task/create` response model already uses "always emit, possibly null" for the same fields.
7. `frontend/src/app/page.tsx:236-238` (recent-tasks loader) — the templates loader resets `setTemplates([])` on failure; the recent-tasks loader only logs and never resets state. Benign today, but the two sibling loaders differ for no stated reason.
8. `frontend/src/app/page.tsx:161-168` (`resolveHomeGetStartedDestination`) — no scheme validation, unlike `resolveAgentLogoUrl` which explicitly rejects `javascript:`. Resolved value flows straight into `href={card.link!}`. Build-time-trusted config, so low exploitability, but inconsistent rigor and undocumented in the contract.
9. `frontend/src/lib/time-utils.ts:22-23` — loose date parsing fabricates dates: `formatDisplayDate("5", "en", {...})` returns `"5/1/2001"` rather than `""` (verified). Latent only, since the backend always sends ISO-8601.
10. `frontend/src/app/page.tsx:151-153` — the `as {...}` cast on the optional extension export bypasses structural type-checking of a replacement module's export shape; only the runtime `typeof` check in `resolveHomeGetStartedDestination` prevents an actually wrong-shaped replacement from breaking.
11. `frontend/src/lib/models.ts:82,393` and `frontend/src/lib/utils.ts:166-176` — dead code: `DefaultModelConfig`, `getSystemDefaultModels` (still carrying the removed `_token` parameter), and `formatDate` (hardcoded `zh-CN`) all have zero importers repo-wide.
12. `frontend/src/components/settings/default-models.tsx:92-93` — the backend's `DEFAULT_MODEL_CONFIG_TYPES` includes `rerank`, so a rerank default is fetched/validated/stored, but `Exclude<DefaultModelType,'rerank'>` plus a cast makes it silently unrenderable in the Settings UI. Pre-existing gap, newly type-visible.

## Simplification Opportunities

All four items from the prior round were already addressed in `4e365a05` (dead `configuredHomeGetStartedDestinationOverrides` re-bind removed, `rawPrompt` alias collapsed, `templateGenerationRef` replaced with `active`/`isCurrent()`, and the four pure field-forwarding helpers in `default-models.tsx` inlined). Two new items:

- `frontend/src/app/build/page.tsx:386`: shrink: five of the six `if (!isCurrent()) return` checks (~386, 389, 396, 399, 402, 405) guard a run of purely synchronous statements (dispatch / setPendingMessage / dispatch / setTaskId / setIsCreateModalOpen) with zero `await` between them, so only the first check is load-bearing. Keep the single check already at ~L383 and delete the rest.
- `frontend/src/app/page.tsx:345`: shrink: three `if (!isCurrent()) return` checks (~345, 352, 355) guard synchronous setPendingMessage/setTaskId/draft-clear calls with zero `await` between them. Keep the single check already at ~L342 and delete the rest.

net: -8 to -10 lines possible.

## Tests

`npm run test:home-build-pages`: 9 files, 411 tests, all passed. `npx tsc --noEmit`: clean, no errors. The new A2/A3 tests were mutation-checked and do assert real behavior. The A4 caveat still applies to the source-text/regex assertions elsewhere in these files.

## Overall Verdict

**Request changes** — narrowly. The update commit cleanly resolved every actionable item from the last round, and the design-level reservations in the approach verdict are accepted trade-offs that shouldn't block a merge. What blocks it is the two new majors above, both attributable to this PR rather than pre-existing. Fix those two and this is an approve. Everything under "Minor / nits" is follow-up material — I'd single out A4 (source-text test assertions) and the unapplied `parseUserDefaultModels` fix on the two KB screens as the highest-value follow-ups, since the latter is a live user-visible bug this PR already built the fix for.

Issue #1108, opened 2026-08-03
enh(widget): widget public endpoints missing remaining share-path abuse controls (auth, task-create, run quota)


codeacme17:
Spun out of the analysis on #1056. That issue covers the two missing **websocket** limiters on the widget path; while verifying it, three more share-path abuse controls turned out to have no widget counterpart. They are grouped here because they share the same root cause (widget surfaces were added after the #973/#993 share hardening and never picked up the controls) and the same fix substrate (`get_share_rate_limiter()`).

### 1. The per-link/per-guest run quota does not cover widget tasks

`_load_task_share_quota_config_isolated` (`src/xagent/web/api/chat.py:992`) only returns the quota markers when `agent_config["auth_mode"] == "share"`, so the `allow_run` gate at the `execute_task` chokepoint (`src/xagent/web/api/chat.py:2922`) — the rolling per-link 500/day + per-guest 60/hour ceiling that bounds owner-billed runs — is skipped entirely for `auth_mode == "widget"` tasks. This is the deepest billing backstop on the share path and arguably matters more than the WS throttles in #1056: even with those added, a widget link has no rolling run ceiling at all.

Widget tasks already stamp `widget_agent_id` / `widget_workforce_id` into `agent_config` (`src/xagent/web/api/public_chat_access.py:883`), so the entity key (`agent:<id>` / `workforce:<id>`) is available at the chokepoint. Note the guest-bucket caveat below.

### 2. `/api/widget/auth` has no rate limit

The share auth endpoint is throttled via `allow_auth` — per-IP 300/min plus per-token 60/min (`src/xagent/web/api/share.py:57`). The widget equivalent `authenticate_widget` (`src/xagent/web/api/widget.py:386`) has no gate: every call does DB lookups (key/ticket resolution, owner load) and mints a JWT. The widget key is public by design, so this endpoint is reachable by anyone who can view the hosting page.

### 3. Widget HTTP task-create has no rate limit

Share task-create is throttled via `allow_task_create` — per-token then per-guest (`src/xagent/web/api/share.py:185`). The widget equivalent `create_widget_task` (`src/xagent/web/api/widget.py:493`) passes straight through to `create_public_chat_task` with no gate. (#1056 argues WS turns matter because they *bypass* the HTTP task-create limiter; on the widget path the limiter being bypassed does not exist either.)

## Design constraint: widget guest_id is not a usable limiter key

[...]

Issue #1057, opened 2026-07-29
enh(widget): connect-time revocation reports a bare 4001, losing the real close reason


codeacme17:
Raised in the re-review of #1049 (@rogercloud). Pre-existing and currently pinned by a test, so it reads as intentional — filing to decide whether it should stay that way.

## Problem

The two public-chat WS endpoints disagree about how an auth failure at connect is reported.

`share_chat_websocket_endpoint` accepts the handshake **before** authorizing, deliberately:

```python
# uvicorn only preserves a close code + reason on the wire for a *post-accept*
# close: a pre-accept ``websocket.close()`` is collapsed into a bare HTTP 403
# with an empty body ...
await websocket.accept()
```

so a denial arrives as `4003` carrying the real reason, and the frontend can tell an access denial from a transport failure and recover.

`public_chat_websocket_endpoint` (widget) does the opposite — it authorizes first and closes pre-accept on any exception:

```python
except Exception:
    await websocket.close(code=4001, reason="Authentication required")
```

Two consequences:

1. The real code/reason (`4003` / "Widget is unavailable") is discarded, and uvicorn collapses the pre-accept close into a bare HTTP 403 — the browser sees `code=1006, reason=""`. The widget client cannot distinguish "your widget was revoked" from "the network dropped".
2. It catches bare `Exception`, so an `HTTPException` from the revocation checks is flattened into the same generic `4001` as a malformed token.

The result is an asymmetry *within the widget path itself*: revocation mid-session gives a clean `4003` + reason (covered by #992), but a guest who reconnects after revocation — the far more common sequence, since a revoked guest reloads the page — gets nothing actionable.

## Notes

- `test_public_chat_websocket_db_boundary.py::test_public_access_websocket_initial_auth_failure_stays_4001` pins the current behaviour and documents the share/widget split, so changing this means updating that test intentionally.
- Reconnect-after-revocation is untested on the widget path today; a fix should add that case.
- Aligning widget with share (accept first, map `HTTPException` -> `4003` + reason, keep `4001` for non-HTTP failures) would make the two paths consistent — but it also makes every widget connect attempt complete a 101 upgrade before rejection, which is exactly why the share path needed a per-IP connect budget. So this should probably land together with, or after, the widget rate-limiting issue.

PR #1113, opened 2026-08-03
enh(widget): add rate limiting to the widget chat websocket (#1056)

### LINKED ISSUE #1056: enh(widget): widget WS/chat endpoints have no rate limiting (share path does)
Raised in the re-review of #1049 (@rogercloud). Pre-existing; #992 only added test coverage on this path, which is what made the asymmetry visible.

## Problem

The share and widget public-chat paths carry very different abuse budgets. `share_chat_websocket_endpoint` has two limiters (added in #973/#993):

- **per-IP connect budget**, checked *pre-accept*, closing `4008` — because accept-before-auth means every attempt completes a 101 upgrade even with a garbage token.
- **per-guest turn budget** on `chat` / `execute_task` frames — because follow-up turns bypass the HTTP task-create limiter and each starts an owner-billed run.

`public_chat_websocket_endpoint` (the widget path) has **neither**. A widget guest holding one valid token can open sockets and send run-starting turns without any ceiling, each turn billed to the widget owner.

## Why it matters

A widget key is embedded in a public page by design, so anyone who can view the page can mint a guest token via `/api/widget/auth`. The economics are the same as the share path — owner-billed runs from anonymous callers — but the controls are not.

## Suggested shape

Reuse `get_share_rate_limiter()` rather than inventing a second mechanism; the limiter is already generic (`allow_ws_connect(ip)` / `allow_ws_turn(guest_id)`) and is only named for the share path. Points to settle:

- Whether widget and share should share one budget or get separate env-configurable limits (`XAGENT_SHARE_WS_CONNECT_IP_RATE_LIMIT` has no widget counterpart today).
- The widget endpoint closes *pre-accept* on auth failure, unlike share — so a pre-accept rate-limit close there is consistent with its existing behaviour, but see the sibling issue about that asymmetry.
- Turn rejection should follow the share precedent: reject the turn via `send_message_delivery` and keep the socket open, since a rate limit is transient.

`tests/web/api/test_share_rate_limit_endpoints.py` is the model for the tests.

codeacme17:
Closes #1056.

The widget public chat websocket (`public_chat_websocket_endpoint`, served at `/api/widget/chat/ws/{task_id}`) had neither of the two abuse gates the share websocket gained in #973/#993, despite identical economics: a widget key is embedded in a public page by design, so anyone who can view the page can mint a guest token and start owner-billed runs. This PR adds both gates, following the design settled in the issue discussion.

## Connect gate (pre-accept, per IP)

`allow_widget_ws_connect(remote_ip)` runs before the handshake is accepted; an over-limit attempt is closed with `4008` and never pays the upgrade cost. This matches the endpoint's existing pre-accept close on auth failure (see #1057 for that asymmetry with share) and the share gate's precedent.

## Turn gate (entity + IP, in-band rejection)

Run-starting frames (`chat` / `execute_task`) are throttled via a new `allow_widget_ws_turn(entity_key, remote_ip)`. Per the correction recorded on the issue, this does **not** reuse the share path's `allow_ws_turn(guest_id)`: the widget `guest_id` is client-supplied and rotatable at will, so the gate keys on the widget entity (`agent:<id>` / `workforce:<id>`) plus caller IP — the same reasoning and shape as `allow_widget_upload`. The per-IP bucket is the tight per-abuser bound; the per-entity bucket is the loose backstop across all callers of one widget.

Rejection follows the share precedent: the turn is refused via `send_message_delivery` (generic error fallback when the client didn't tag the turn) and the socket stays open, since a rate limit is transient. Interventions are control messages, not new runs, and stay ungated. A missing entity key (unreachable for a validated widget token) degrades to the limiter's shared "unknown" bucket rather than skipping the gate, so the throttle fails closed on the key.

## Configuration

Separate buckets and env vars from the share limits, so probes against one public channel cannot consume the other's budget (settled on the issue):

| Env var | Default | Bucket |
|---|---|---|
| `XAGENT_WIDGET_WS_CONNECT_IP_RATE_LIMIT` | `120/minute` | connection attempts per caller IP |
| `XAGENT_WIDGET_WS_TURN_IP_RATE_LIMIT` | `60/minute` | run-starting turns per caller IP |
| `XAGENT_WIDGET_WS_TURN_RATE_LIMIT` | `240/minute` | run-starting turns per widget entity |

Defaults mirror the share values (connect, turn-IP) and the widget upload entity backstop. Documented in `example.env`; getters follow the `config.py` pattern.

## Plumbing

- `WebSocketPrincipal` gains an optional `widget_entity_key`, derived at authorize time from the validated access context (`_widget_entity_key`), so the turn gate never trusts client input for its key.
- The share endpoint's inline turn-rejection block is extracted into `_reject_rate_limited_turn` and shared by both endpoints (behavior-preserving; existing share WS tests unchanged and passing).

## Tests

Modeled on `tests/web/api/test_share_rate_limit_endpoints.py` as the issue prescribed:

- Limiter unit tests: per-IP connect bucket isolation from the share bucket, turn per-IP bucket trips across entities (entity rotation cannot escape it), entity ceiling trips across IPs, fail-open coverage for both new gates.
- Endpoint tests: pre-accept `4008` refusal (with the admitted-then-denied `4001`/`4008` sequence pinning the ordering), rate-limited turn rejected without dispatch while the socket stays open, interventions dispatch even with the turn budget exhausted.
- Config: three parametrized default/override/blank-fallback cases.
- `test_public_chat_websocket_db_boundary.py` updated for the new principal field and now pins the entity-key derivation on the widget authorize path.

The remaining non-WS widget gaps found during the issue analysis (run quota, auth throttle, task-create throttle) are deliberately out of scope here and tracked in #1108.

### DIFF
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -1119,10 +1119,25 @@ async def create_share_chat_task(
     )
 
 
+def _widget_entity_key(context: PublicChatAccessContext) -> str | None:
+    """Rate-limit key for the widget entity a guest is scoped to (#1056).
+
+    Matches the ``allow_widget_upload`` keying: the widget ``guest_id`` is
+    client-supplied (rotatable at will), so widget throttles key on the
+    embedded agent/workforce instead.
+    """
+    if context.widget_workforce_id is not None:
+        return f"workforce:{context.widget_workforce_id}"
+    if context.widget_agent_id is not None:
+        return f"agent:{context.widget_agent_id}"
+    return None
+
+
 def _authorize_chat_websocket_sync(
     *,
     load_access_context: Callable[[Session], _ChatAccessContextT],
     authorize_task: Callable[[Session, _ChatAccessContextT], object],
+    widget_entity_key: Callable[[_ChatAccessContextT], str | None] | None = None,
 ) -> WebSocketPrincipal:
     """Authorize one public WebSocket operation in a worker-owned Session."""
 
@@ -1137,6 +1152,9 @@ def _authorize_chat_websocket_sync(
             id=int(user_id),
             is_admin=bool(access_context.user.is_admin),
             guest_id=access_context.guest_id,
+            widget_entity_key=widget_entity_key(access_context)
+            if widget_entity_key is not None
+            else None,
         )
 
 
@@ -1162,6 +1180,7 @@ async def _authorize_public_chat_websocket(
         lambda: _authorize_chat_websocket_sync(
             load_access_context=load_access_context,
             authorize_task=authorize_task,
+            widget_entity_key=_widget_entity_key,
         )
     )
 
@@ -1206,6 +1225,35 @@ def _ws_close_reason(detail: object) -> str:
     return encoded[:_WS_CLOSE_REASON_MAX_BYTES].decode("utf-8", errors="ignore")
 
 
+async def _reject_rate_limited_turn(websocket: WebSocket, message_data: dict) -> None:
+    """Reject one rate-limited run-starting turn, keeping the socket open.
+
+    A rate limit is transient, so the turn is refused in-band (the client
+    surfaces it and can retry) rather than closing the connection.
+    """
+    client_message_id = message_data.get("client_message_id")
+    rate_limited_message = (
+        "You're sending messages too quickly. Please wait a moment and try again."
+    )
+    await send_message_delivery(
+        websocket,
+        client_message_id=client_message_id,
+        turn_id=str(client_message_id or ""),
+        accepted=False,
+        message=rate_limited_message,
+        rejection_outcome="not_accepted",
+    )
+    # send_message_delivery no-ops without a client_message_id, so a client
+    # that didn't tag the turn would otherwise get zero feedback and see the
+    # message silently dropped. Fall back to a generic error so the throttle
+    # is always surfaced.
+    if client_message_id is None:
+        await manager.send_personal_message(
+            {"type": "error", "message": rate_limited_message},
+            websocket,
+        )
+
+
 async def public_chat_websocket_endpoint(
     *,
     websocket: WebSocket,
@@ -1213,7 +1261,20 @@ async def public_chat_websocket_endpoint(
     token: str = Query(..., description="Authentication token"),
     expected_auth_mode: str,
 ) -> None:
-    """Serve widget/share websocket chat with per-message revalidation."""
+    """Serve widget websocket chat with per-message revalidation."""
+    # Handshake budget (#1056): mirror of the share connect gate — a widget
+    # key is embedded in a public page, so connection attempts are anonymous
+    # and need a per-IP ceiling. This endpoint closes pre-accept on auth
+    # failure below (uvicorn collapses that into a plain HTTP 403), so the
+    # pre-accept refusal here is consistent with its existing behaviour and
+    # is the cheapest rejection: no upgrade is performed at all. The IP is
+    # captured once — it cannot change for the lifetime of the socket — and
+    # reused by the per-turn gate in the loop below.
+    remote_ip = remote_ip_from_request(websocket)
+    if not get_share_rate_limiter().allow_widget_ws_connect(remote_ip):
+        await websocket.close(code=4008, reason="Too many connection attempts")
+        return
+
     try:
         principal = await _authorize_public_chat_websocket(
             token=token,
@@ -1246,11 +1307,32 @@ async def public_chat_websocket_endpoint(
             message_data["user_id"] = current_principal.id
             message_data["user"] = current_principal
 
-            if message_data.get("type") == "chat":
+            message_type = message_data.get("type")
+            # Abuse control (#1056): follow-up turns bypass any HTTP throttle
+            # and each starts an owner-billed run, so rate-limit the
+            # run-starting turn types here — the widget mirror of the share
+            # turn gate. Keyed on the widget entity + caller IP, NOT the
+            # guest: the widget guest_id is client-supplied (rotatable at
+            # will). Interventions are control messages, not new runs, so
+            # they are not gated. A missing entity key (unreachable for a
+            # validated widget token, which always carries exactly one entity
+            # id) degrades to the limiter's shared "unknown" bucket rather
+            # than skipping the gate — the per-IP bucket needs no entity, so
+            # the throttle must not fail open on the key.
+            if message_type in (
+                "chat",
+                "execute_task",
+            ) and not get_share_rate_limiter().allow_widget_ws_turn(
+                current_principal.widget_entity_key or "", remote_ip
+            ):
+                await _reject_rate_limited_turn(websocket, message_data)
+                continue
+
+            if message_type == "chat":
                 await handle_chat_message(websocket, task_id, message_data)
-            elif message_data.get("type") == "execute_task":
+            elif message_type == "execute_task":
                 await handle_execute_task(websocket, task_id, message_data)
-            elif message_data.get("type") == "intervention":
+            elif message_type == "intervention":
                 await handle_intervention(websocket, task_id, message_data)
     except Exception as exc:
         from fastapi import WebSocketDisconnect
@@ -1340,28 +1422,7 @@ async def share_chat_websocket_endpoint(
                     current_principal.guest_id
                 )
             ):
-                client_message_id = message_data.get("client_message_id")
-                rate_limited_message = (
-                    "You're sending messages too quickly. "
-                    "Please wait a moment and try again."
-                )
-                await send_message_delivery(
-                    websocket,
-                    client_message_id=client_message_id,
-                    turn_id=str(client_message_id or ""),
-                    accepted=False,
-                    message=rate_limited_message,
-                    rejection_outcome="not_accepted",
-                )
-                # send_message_delivery no-ops without a client_message_id, so
-                # a client that didn't tag the turn would otherwise get zero
-                # feedback and see the message silently dropped. Fall back to a
-                # generic error so the throttle is always surfaced.
-                if client_message_id is None:
-                    await manager.send_personal_message(
-                        {"type": "error", "message": rate_limited_message},
-                        websocket,
-                    )
+                await _reject_rate_limited_turn(websocket, message_data)
                 continue
 
             if message_type == "chat":
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -33,6 +33,7 @@ from ..services.db_runtime import run_db_io_cancellation_safe
 from ..services.deployments import get_deployment
 from ..services.orphan_upload_gc import TASKLESS_SHARE_UPLOAD_SOURCE
 from ..services.share_rate_limit import (
+    entity_rate_limit_key,
     get_share_rate_limiter,
     remote_ip_from_request,
 )
@@ -1126,11 +1127,7 @@ def _widget_entity_key(context: PublicChatAccessContext) -> str | None:
     client-supplied (rotatable at will), so widget throttles key on the
     embedded agent/workforce instead.
     """
-    if context.widget_workforce_id is not None:
-        return f"workforce:{context.widget_workforce_id}"
-    if context.widget_agent_id is not None:
-        return f"agent:{context.widget_agent_id}"
-    return None
+    return entity_rate_limit_key(context.widget_agent_id, context.widget_workforce_id)
 
 
 def _authorize_chat_websocket_sync(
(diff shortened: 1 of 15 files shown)


src/xagent/web/services/share_rate_limit.py
```diff
@@ -211,6 +226,44 @@ def allow_ws_connect(self, remote_ip: str | None) -> bool:
...
+        known yet).
+        """
+        return self._limiter.hit(
+            self._widget_ws_connect_ip_limit,
+            _WIDGET_WS_CONNECT_IP_NAMESPACE,
+            remote_ip or "unknown",
+        )
+
+    @_fail_open
+    def allow_widget_ws_turn(self, entity_key: str, remote_ip: str | None) -> bool:
```
rogercloud: Minor design note, not blocking and not new to this PR: this hits the per-IP bucket before testing the per-entity bucket, so a request that is ultimately denied by the entity ceiling still burns an IP-bucket slot. The pre-existing `allow_widget_upload` has exactly this ordering and the docstring here says it is mirroring it deliberately, so this is consistent. It also doesn't manifest under the shipped defaults (entity 240/min > IP 60/min, so in normal operation the IP bucket denies first). Worth noting that `allow_run` in this same file takes the other approach — non-destructive `test()` on both buckets before committing either — which avoids the wasted slot entirely. If the two methods ever get consolidated (see the simplification note), the `test()`-first shape would be the better base.

tests/web/api/test_share_rate_limit_endpoints.py
```diff
@@ -321,6 +323,132 @@ def test_ws_connect_over_limit_is_refused_pre_accept(
     assert denied.value.code == 4008
 
 
+@pytest.mark.asyncio
+async def test_widget_ws_turn_rate_limited_rejects_without_dispatch(
```
rogercloud, line 332: Test-quality nit: this asserts the dispatch/delivery call counts but never asserts that the websocket was *not* closed after a throttled turn. Since the shared fake websocket's `close()` is a true no-op (it doesn't block subsequent `receive_text()` calls), a future regression that closes the socket on a throttled turn would still pass this test silently. Suggest adding an explicit assertion that `close` was not called, so the "socket stays open, turn is just rejected" contract is actually pinned.

src/xagent/web/api/public_chat_access.py
```diff
@@ -1246,11 +1307,32 @@ async def public_chat_websocket_endpoint(
...
+            # run-starting turn types here — the widget mirror of the share
+            # turn gate. Keyed on the widget entity + caller IP, NOT the
+            # guest: the widget guest_id is client-supplied (rotatable at
+            # will). Interventions are control messages, not new runs, so
+            # they are not gated. A missing entity key (unreachable for a
+            # validated widget token, which always carries exactly one entity
+            # id) degrades to the limiter's shared "unknown" bucket rather
+            # than skipping the gate — the per-IP bucket needs no entity, so
+            # the throttle must not fail open on the key.
+            if message_type in (
```
rogercloud, line 1319: Coverage gap: this gate condition covers both `chat` and `execute_task` frames, but no test in `tests/web/api/test_share_rate_limit_endpoints.py` or `tests/web/api/test_public_chat_websocket_db_boundary.py` drives an `execute_task` frame through the new widget turn gate — only `chat` frames. Suggest parametrizing (or duplicating) one of the widget turn-gate tests over both frame types so the `execute_task` branch is exercised.

tests/web/api/test_share_rate_limit_endpoints.py
```diff
@@ -321,6 +323,132 @@ def test_ws_connect_over_limit_is_refused_pre_accept(
...
+        task_id=1,
+        token="jwt",
+        expected_auth_mode="widget",
+    )
+
+    assert chat_dispatch.await_count == 1
+    assert intervention_dispatch.await_count == 1
+
+
+def test_widget_ws_connect_over_limit_is_refused_pre_accept(
```
rogercloud, line 438: Docstring accuracy: this says a garbage token lets the two refusals be distinguished by close code (4001 vs 4008), which reads as a claim about what a real client can observe. Both closes here happen pre-accept, and the endpoint's own comment at `src/xagent/web/api/public_chat_access.py:1265-1276` notes that uvicorn collapses a pre-accept close into a bare HTTP 403 / `code=1006, reason=''` — so on the wire the two refusals are indistinguishable in production. The test only passes because Starlette's `TestClient` preserves close codes exactly. This differs from the analogous share-path test, where the two closes genuinely do differ (one is post-accept). Suggest rewording the docstring to say the distinction is TestClient-only and does not survive to a real client.

rogercloud (approved):
## Summary

This PR closes the abuse-gate gap between the widget public-chat websocket and the share websocket, which received a per-IP connect gate and a per-guest turn gate back in #973/#993. Two analogous gates are added to the widget endpoint: a pre-accept per-IP connect gate (`allow_widget_ws_connect`) and an in-band turn gate (`allow_widget_ws_turn`) keyed on **widget entity + caller IP** rather than `guest_id` — the right call, since the widget's `guest_id` is client-supplied and freely rotatable. Three new env vars give the widget path its own rate-limit namespaces so probes against one channel can't drain the other's budget. Supporting plumbing: `WebSocketPrincipal` gains a server-derived `widget_entity_key`, and the share endpoint's turn-rejection logic is extracted into a shared `_reject_rate_limited_turn` helper.

## Design verdict: Sound

Every load-bearing claim in the PR description holds up against the code and tests. The entity+IP keying can't be escaped by rotating `guest_id`, rotating the entity doesn't escape the IP bucket, and rotating the IP doesn't escape the entity ceiling — each is covered by a dedicated test. The "fail closed on a missing entity key" behavior and the separate-namespace claim are both true in the code, and the `_reject_rate_limited_turn` extraction is genuinely behavior-preserving: the three pre-existing share turn-rejection tests are unmodified and still pass.

**Test run: 406 passed, 0 failed.**

Everything below is minor or optional. Nothing blocks merge.

## Design-level notes

**Duplicated entity-key formatting (refactor suggestion, not a correctness issue).** The new `_widget_entity_key` in `src/xagent/web/api/public_chat_access.py:1122-1133` and the existing formatting in `src/xagent/web/api/widget.py:472-476` independently build the same `workforce:<id>` / `agent:<id>` string, with a third copy in `src/xagent/web/api/chat.py:2910-2917`. The one visible divergence — what happens when both ids are `None` — is not reachable in practice: both call sites only ever see a context where exactly one id is guaranteed set by `get_public_chat_user`, as your own comment at `public_chat_access.py:1317-1319` notes. So this is purely a "one helper instead of three" cleanup (see Simplification below), not a bug.

**`X-Forwarded-For` trust — awareness only, not something to fix here.** `remote_ip_from_request` in `src/xagent/web/services/trigger_rate_limit.py` (untouched by this PR, pre-existing) will clamp to the leftmost, client-controlled `X-Forwarded-For` entry if `XAGENT_TRUSTED_PROXY_HOPS` is configured larger than the real hop count, which would let a caller rotate a fake IP per request and bypass any IP-keyed gate. The share path already depended on this before this PR; the only thing that changes here is that a widget-path, explicitly billing-relevant gate now depends on it too, so a misconfiguration has a slightly wider blast radius. Flagging for awareness rather than as a required change.

**Docstring wording on the per-IP turn default.** In `src/xagent/config.py`, `get_widget_ws_turn_ip_rate_limit`'s docstring describes the 60/min default as "kept loose enough for enterprise NAT (matches the share per-guest turn rate)." The docstring does correctly disclose the IP keying, so it isn't misleading — but "matches" overstates the parity: the share gate is per-guest, so N guests behind one NAT egress get N x 60/min there, while on widget they share a single 60/min. That compounds with `XAGENT_TRUSTED_PROXY_HOPS` defaulting to 0, which can collapse an entire proxied deployment onto one apparent IP. The defaults are a reasonable starting point; a one-line tightening of the wording ("numerically matches the share per-guest default, though this bucket is per-IP") would avoid the implication of behavioral parity. Would be interested in your read on whether 60/min per shared egress IP is comfortable for the NAT cases you have in mind.

- `src/xagent/web/services/share_rate_limit.py` L245: `native:` `allow_widget_ws_turn` reimplements the "hit IP bucket, then hit entity bucket, both with `or "unknown"` fallback" sequence already present byte-for-byte in `allow_widget_upload` (L274-294). Extract a private `_hit_ip_then_entity(ip_limit, ip_ns, remote_ip, entity_limit, entity_ns, entity_key)` and call it from both; each keeps its own `@_fail_open`.
- `src/xagent/web/api/public_chat_access.py` L1122: `native:` `_widget_entity_key`, `src/xagent/web/api/widget.py:472-476`, and `src/xagent/web/api/chat.py:2910-2917` all format the same `workforce:<id>`/`agent:<id>` key with the same workforce-first precedence. Extract one shared `widget_entity_key(agent_id, workforce_id) -> str | None` (e.g. in `share_rate_limit.py`) and use it at all three sites; as a side effect this also removes the `widget.py` `"agent:None"` fallback divergence.

Net: ~13 lines removed across both, no behavior change.

src/xagent/web/services/share_rate_limit.py
codeacme17: Done in 8636acee — took your suggestion one step further and consolidated now rather than later: `allow_widget_upload` and `allow_widget_ws_turn` both delegate to a shared `_admit_ip_and_entity`, built on the `test()`-first shape you pointed at. Both buckets are tested non-destructively and only consumed when both admit, so the wasted-slot asymmetry this thread describes is gone entirely (for the upload gate too, which previously burned an IP slot on an entity denial). The non-atomic test→hit overshoot is documented on the helper, same as on `allow_run`.

tests/web/api/test_share_rate_limit_endpoints.py
codeacme17: Done in 8636acee. `_FakeWebSocket.close()` now records into `self.closed`, and the widget turn test, the share turn test, and the share no-client-id test all assert `websocket.closed is False` after the throttled turn, so the "socket stays open" contract is pinned rather than implied.

src/xagent/web/api/public_chat_access.py
codeacme17: Done in 8636acee — `test_widget_ws_turn_rate_limited_rejects_without_dispatch` is now parametrized over `["chat", "execute_task"]`, with one mock behind both run-starting handlers so the admitted/throttled dispatch counts hold for either frame type.

tests/web/api/test_share_rate_limit_endpoints.py
codeacme17: Good catch — reworded in 8636acee. The docstring now states explicitly that the 4001/4008 distinction is TestClient-only, that in production uvicorn collapses both pre-accept closes into a bare HTTP 403 (code=1006, reason="" client-side), and that the codes pin server-side gate ordering rather than client-observable behaviour — with the contrast to the share test, whose admitted attempt genuinely fails post-accept.

codeacme17:
Thanks for the thorough pass @rogercloud — all actionable items are addressed in 8636acee:

- **Entity-key duplication**: extracted `entity_rate_limit_key(agent_id, workforce_id)` into `share_rate_limit.py` and pointed all three sites at it (widget WS authorize, widget upload, share run-quota gate in `chat.py`). Named it `entity_rate_limit_key` rather than `widget_entity_key` since the third caller keys the *share* run quota — same format, not a widget. As you predicted, this also removes the `widget.py` `"agent:None"` fallback divergence: the helper returns `None` there and the gate degrades to its `"unknown"` bucket.
- **Dual-bucket consolidation**: done now rather than deferred — `_admit_ip_and_entity` backs both `allow_widget_upload` and `allow_widget_ws_turn`, on the `test()`-first shape per your inline note.
- **Turn-IP docstring**: reworded along the lines you suggested, plus a pointer to the env var and `XAGENT_TRUSTED_PROXY_HOPS` for proxied deployments.
- **X-Forwarded-For trust**: acknowledged. Pre-existing and shared with every IP-keyed gate; `example.env` already carries a stern warning on `XAGENT_TRUSTED_PROXY_HOPS` (both directions of misconfiguration). Agreed it's not this PR's to fix.

On your question — **is 60/min per shared egress IP comfortable for NAT populations?** My read: yes as a default, for two reasons. First, these are *run-starting* turns, each of which triggers an LLM-backed run that takes seconds to minutes; a NAT population legitimately sustaining >1 run-start per second through one widget for a full minute is already an unusual deployment, and it's exactly the deployment that should be tuning `XAGENT_WIDGET_WS_TURN_IP_RATE_LIMIT` consciously. Second, the failure mode is gentle: the turn is rejected in-band with a retry message and the socket stays open — a briefly throttled NAT user loses seconds, not their session. The sharper operational risk is the one you flagged: `XAGENT_TRUSTED_PROXY_HOPS=0` behind a proxy collapsing *all* traffic onto one IP. That one bites every IP-keyed gate at once (share connect included), so it's a deployment-checklist item rather than something a larger default here would meaningfully absorb.

All affected suites green after the refactor (limiter, endpoint, revalidation, db-boundary, workforce widget, config); mypy/ruff clean.

PR #1114, opened 2026-08-03
refactor(tasks): stage continuation commands in caller transactions

### LINKED ISSUE #1073: refactor(tasks): stage continuation commands in caller transactions
## Summary

Let the lifecycle owner insert a durable continuation command inside the same transaction that accepts an interaction response.

## Current behavior

The command transport currently commits and notifies from inside `enqueue_task_command()`. A response could therefore be accepted without its lifecycle rows, or lifecycle rows could commit without the command.

## Acceptance criteria

- [ ] Add a caller-transaction command-staging path that validates and adds/flushes only; it must not commit, roll back, or notify.
- [ ] Make the existing `enqueue_task_command()` wrapper a production consumer of the staging primitive in the same PR, preserving commit, post-commit notify, and conflict re-read behavior.
- [ ] Keep a convenience wrapper for existing callers if needed, implemented on top of the staging primitive.
- [ ] Emit notification only after the owner transaction commits. Lost notification must converge through durable command polling.
- [ ] An outer rollback leaves no command and no partial response mutation.
- [ ] A uniqueness conflict rolls back the entire owner transaction and is classified by the lifecycle owner.
- [ ] Cover SQLite, PostgreSQL, concurrent writers, rollback, duplicate correlation, and post-commit notification loss.

## Dependencies

- None; this can land independently.

## Non-goals

- Changing worker command execution semantics.
- Adding a second response queue.
- Adding an interaction-response command kind or handler before its lifecycle consumer exists.

AlexLiu190625:
## Summary

Extract a caller-transaction command-staging primitive from `enqueue_task_command` so a lifecycle owner can insert a durable continuation command inside the same transaction that accepts an interaction response.

- `stage_task_command(db, ...) -> StagedTaskCommand`: validates the command id, flushes the caller's pending state so later failures are attributable to the command insert, resolves the task snapshot (`status`/`runner_id`/`run_id`) with a single Core select that bypasses the identity map and sees the caller's own uncommitted writes, performs the idempotent precheck, then adds and flushes the command row. It never commits, rolls back, or notifies; the staged row id (`staged_db_id`) is not dispatchable until the owner commits.
- `classify_task_command_conflict(...)`: post-rollback conflict classification (`RACED_DUPLICATE` with a frozen projection of the raced row computed in one query, `TASK_MISSING` with task-FK vs actor-FK disambiguation, `UNRELATED`).
- `TaskCommandOwnerStateError`: raised when the entry flush fails on the caller's own pending state; deliberately not an `IntegrityError` subclass so command-conflict handling never misattributes it. Callers must roll back before issuing further statements.
- `enqueue_task_command` is rewritten as the production consumer of both primitives. Its signature, return type, exception surface, idempotency semantics, and commit/notify ordering (commit and post-commit notify only when a row was created) are unchanged; the three existing call sites have a zero diff.

## Behavior changes

None for existing callers. The transport module's external contract is byte-identical; all pre-existing tests in `tests/web/services/test_task_command_transport.py` pass unmodified except two whose flush monkeypatches were extended to pass through the new entry flush.

## Verification

- `tests/web/services/test_task_command_transport.py`: 53 baseline tests all green and retained; 16 new tests cover staging visibility (a second session cannot see a staged row before the owner commits), outer-rollback zero-residue, the interleaved duplicate-enqueue race on both backends, a PostgreSQL-only truly-concurrent insert conflict (executed against a real `postgres:16` service), owner-state error routing, snapshot correctness under both bulk-update and ORM-attribute mutations pending in the caller session, staged-id non-dispatchability with poll-based convergence, SQLite lock-window behavior, and notify-after-commit ordering.
- Mutation checks: re-adding a commit or a notify inside the staging primitive, breaking the conflict classifier, and reordering the entry flush each turn specific tests red.
- Full suite (`-n 4 --dist=loadscope`) shows no failure attributable to this change: the touched module's tests are fully green, and the remaining failures reproduce identically on the base commit.

Fixes #1073.

### DIFF
--- a/src/xagent/web/services/task_command_transport.py
+++ b/src/xagent/web/services/task_command_transport.py
@@ -88,6 +88,55 @@ class EnqueuedTaskCommand:
 
 
 @dataclass(frozen=True)
+class StagedTaskCommand:
+    """A command row added to a caller-owned session, not yet committed.
+
+    ``staged_db_id`` is deliberately not named ``command_id`` -- unlike
+    ``EnqueuedTaskCommand.command_id``, this id is only guaranteed to exist in
+    the database once the caller commits the session it was staged on. Code
+    that dispatches or notifies by id must never accept a StagedTaskCommand.
+    """
+
+    staged_db_id: int
+    client_command_id: str
+    created: bool
+    payload_matches: bool
+    status: str
+
+
+class TaskCommandOwnerStateError(RuntimeError):
+    """The caller's own pending writes failed to flush before staging began.
+
+    Deliberately not an IntegrityError subclass: the wrapper's
+    ``except IntegrityError`` classifies conflicts on the command insert
+    itself, and must not also catch a failure that has nothing to do with
+    that insert. After this is raised, the session is unusable -- the caller
+    must roll back before issuing any further statement on it.
+    """
+
+
+class TaskCommandConflictKind(enum.Enum):
+    RACED_DUPLICATE = "raced_duplicate"
+    TASK_MISSING = "task_missing"
+    UNRELATED = "unrelated"
+
+
+@dataclass(frozen=True)
+class RacedTaskCommandProjection:
+    """Immutable snapshot of the row that won a duplicate-insert race."""
+
+    command_db_id: int
+    status: str
+    payload_matches: bool
+
+
+@dataclass(frozen=True)
+class TaskCommandConflictClassification:
+    kind: TaskCommandConflictKind
+    raced: RacedTaskCommandProjection | None = None
+
+
+@dataclass(frozen=True)
 class ClaimedTaskCommand:
     id: int
     task_id: int
@@ -160,7 +209,7 @@ def _matches_existing(
     )
 
 
-def enqueue_task_command(
+def stage_task_command(
     db: Session,
     *,
     task_id: int,
@@ -168,18 +217,51 @@ def enqueue_task_command(
     command_id: str,
     kind: TaskCommandKind,
     payload: dict[str, Any],
-) -> EnqueuedTaskCommand:
-    """Commit an idempotent command and return only after it is durable."""
+) -> StagedTaskCommand:
+    """Add an idempotent command row to the session without ending its transaction.
+
+    Never calls ``db.commit()``, ``db.rollback()``, or the dispatcher notify --
+    the caller owns the transaction boundary and decides what durability means
+    for its own writes alongside this command. Statement order is load-bearing:
+
+    1. Reject a malformed command_id before touching the database at all.
+    2. Flush the caller's own pending writes first, so a conflict there is
+       reported as TaskCommandOwnerStateError rather than folded into the
+       IntegrityError this function raises for a duplicate command insert.
+    3. Read the task through a single Core select rather than the ORM. This
+       both closes the window a separate existence check and a separate
+       snapshot read would leave between them, and -- unlike an ORM query --
+       is not served from the identity map, so it observes a caller's own
+       uncommitted CAS-style update to the same row instead of a stale cached
+       instance.
+    4. Check for an existing command with this (task_id, command_id) before
+       adding a new row, so a repeated call is idempotent without relying on
+       the unique constraint to reject it.
+    5. Add and flush the new row last, so any remaining IntegrityError is
+       attributable to the command insert itself.
+    """
 
     normalized_id = command_id.strip()
     if COMMAND_ID_PATTERN.fullmatch(normalized_id) is None:
         raise ValueError("command_id must be 1-64 URL-safe characters")
     resolved_task_id = int(task_id)
+
+    try:
+        db.flush()
+    except IntegrityError as exc:
+        raise TaskCommandOwnerStateError(
+            "caller's pending writes failed to flush before command staging"
+        ) from exc
+
     # A task snapshot a caller loaded earlier would widen the concurrent-delete
     # window across everything it did in between, so existence is re-checked
     # here, by query, before inserting a command that references the row.
-    task = db.query(Task).filter(Task.id == resolved_task_id).first()
-    if task is None:
+    snapshot = db.execute(
+        select(Task.status, Task.runner_id, Task.run_id).where(
+            Task.id == resolved_task_id
+        )
+    ).one_or_none()
+    if snapshot is None:
         raise TaskCommandTaskMissing(f"Task {task_id} not found")
 
     existing = (
@@ -197,8 +279,8 @@ def enqueue_task_command(
             kind=kind,
             payload=payload,
         )
-        return EnqueuedTaskCommand(
-            command_id=int(existing.id),
+        return StagedTaskCommand(
+            staged_db_id=int(existing.id),
             client_command_id=normalized_id,
             created=False,
             payload_matches=matches,
@@ -206,8 +288,8 @@ def enqueue_task_command(
         )
 
     active_runner_id = (
-        task.runner_id
-        if task.status == TaskStatus.RUNNING and task.runner_id is not None
+        snapshot.runner_id
+        if snapshot.status == TaskStatus.RUNNING and snapshot.runner_id is not None
         else None
     )
     command = TaskExecutionCommand(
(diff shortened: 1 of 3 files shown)


tests/web/services/test_task_command_transport.py
```diff
@@ -1995,3 +2078,622 @@ def flush_raising_fk_error(*args, **kwargs):
...
+        assert observed_task.title != "rolled-back-sibling-write"
+        assert (
+            observer.query(TaskExecutionCommand)
+            .filter(TaskExecutionCommand.task_id == task_id)
+            .count()
+            == 0
+        )
+
+
+def test_enqueue_idempotent_hit_issues_no_commit(db_session) -> None:
```
rogercloud, line 2232: The docstring says this pins that a duplicate enqueue leaves the accompanying sibling write uncommitted (still pending) on the owner session, but the only assertion checks that a second, independent session cannot see it. Those are different properties: inserting a `db.rollback()` into the idempotent-hit path -- which destroys the sibling write rather than leaving it pending -- still makes this test pass. Consider asserting the sibling write is still dirty/pending on the owner session (e.g. inspect `db_session.dirty`, or commit the owner session afterward and assert the write persists), so the test actually distinguishes "still pending" from "discarded."

src/xagent/web/services/task_command_transport.py
```diff
@@ -160,26 +209,68 @@ def _matches_existing(
...
     *,
     task_id: int,
     actor_user_id: int | None,
     command_id: str,
     kind: TaskCommandKind,
     payload: dict[str, Any],
-) -> EnqueuedTaskCommand:
-    """Commit an idempotent command and return only after it is durable."""
+) -> StagedTaskCommand:
+    """Add an idempotent command row to the session without ending its transaction.
```
rogercloud, line 253: This docstring is the whole interface contract for future direct callers, and it's missing two obligations. First, it doesn't say the caller should invoke the dispatcher-notify function after its own commit; skipping that degrades silently to up to `DISPATCHER_IDLE_SECONDS` (0.5s) of polling latency, which is low-impact but hard to attribute after the fact. Second, the "you must roll back the whole transaction on `IntegrityError`" obligation is documented only on `classify_task_command_conflict`, whereas this function is where the caller first sees the exception. Worth also adding a line that a caller staying in a long transaction after staging holds SQLite's database-wide writer lock for the entire transaction, not the microseconds the wrapper's immediate commit implies.

src/xagent/web/services/task_command_transport.py
```diff
@@ -160,26 +209,68 @@ def _matches_existing(
...
+    TaskCommandTaskMissing -- this call must be the last write a caller
+    issues before it commits, with no I/O or other slow work in between.
+    Returning leaves the caller holding a write lock (database-wide on
+    SQLite) over its pre-flushed writes, and over the new command row on the
+    created path, until its transaction ends; anything slow placed after
+    this call stalls every concurrent writer, the dispatcher's claim
+    included.
+    """
 
     normalized_id = command_id.strip()
```
rogercloud: `command_id` normalization now happens in three places: here (with `COMMAND_ID_PATTERN` format validation), at ~line 355 in `classify_task_command_conflict` (`.strip()` only, no format validation), and again at ~line 426. The missing validation in the middle one is harmless today because its only caller validated upstream, but that asymmetry is exactly what drifts once the function is reused directly. A shared `_normalize_command_id()` helper would collapse all three into one behavior.

src/xagent/web/services/task_command_transport.py
```diff
@@ -216,54 +307,144 @@ def enqueue_task_command(
...
+                created=False,
+                payload_matches=raced.payload_matches,
+                status=raced.status,
+            )
+        if classification.kind is TaskCommandConflictKind.TASK_MISSING:
+            raise TaskCommandTaskMissing(f"Task {task_id} not found") from None
+        raise
+
+    if not staged.created:
         return EnqueuedTaskCommand(
```
rogercloud: `StagedTaskCommand` and `EnqueuedTaskCommand` are structurally identical apart from the deliberately renamed id field -- keeping them as distinct types is a worthwhile guard and should stay. But the field-by-field copy with the identical id mapping is written out twice here (~436-443 and ~445-452 below). A `StagedTaskCommand.to_enqueued(...)` method would keep the mapping in one place and make the two branches symmetric.

src/xagent/web/services/task_command_transport.py
```diff
@@ -160,26 +209,68 @@ def _matches_existing(
...
+    Returning leaves the caller holding a write lock (database-wide on
+    SQLite) over its pre-flushed writes, and over the new command row on the
+    created path, until its transaction ends; anything slow placed after
+    this call stalls every concurrent writer, the dispatcher's claim
+    included.
+    """
 
     normalized_id = command_id.strip()
     if COMMAND_ID_PATTERN.fullmatch(normalized_id) is None:
         raise ValueError("command_id must be 1-64 URL-safe characters")
```
rogercloud: No test asserts this `ValueError` for a malformed `command_id`, nor that whitespace-padded ids are normalized. The validation itself predates this PR (it was inline in `enqueue_task_command`), but the PR promotes it into a new public entry point, which makes this a good moment to add those two cases.

tests/web/services/test_task_command_transport.py
```diff
@@ -92,6 +101,64 @@ def queue_pool_command_db(tmp_path):
...
+    Base.metadata.create_all(bind=engine)
+    SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+    try:
+        yield engine, SessionLocal
+    finally:
+        engine.dispose()
+
+
+@pytest.fixture()
+def postgres_task_command_sessions():
```
rogercloud, line 130: This fixture calls `init_db(db_url=url)`, rebinding the process-global engine/sessionmaker, but teardown only calls `drop_all` and never restores the previous binding. It also runs `drop_all` destructively at setup time against whatever `XAGENT_TEST_POSTGRES_URL` points at. It's latent today -- every other test in this file re-inits or bypasses the global, and it's gated behind an opt-in env var -- but worth restoring the prior binding in teardown before more Postgres tests land here.

tests/web/services/test_task_command_transport.py
```diff
@@ -1942,8 +2009,16 @@ def test_task_foreign_key_violation_is_reported_as_a_missing_task(
...
     real_flush = db_session.flush
+    flush_calls = 0
 
     def flush_raising_fk_error(*args, **kwargs):
+        # stage_task_command's owner pre-flush is a real, harmless no-op on
+        # this clean session -- let it through so only the attributable
+        # command-insert flush is intercepted below.
+        nonlocal flush_calls
+        flush_calls += 1
+        if flush_calls == 1:
```
rogercloud, line 2031: Gating the fabricated `IntegrityError` on `flush_calls == 1` couples this test to the exact number and order of flushes inside `stage_task_command` (owner pre-flush first, command insert second). If that ever changes, the injection lands on the wrong flush and surfaces as a confusing wrong-exception-type failure rather than a clear one. A short comment explaining the counter, or gating on flush context instead of a raw count, would make the failure mode legible. (Same pattern repeats at line ~2065.)

src/xagent/web/services/task_command_transport.py
```diff
@@ -216,54 +307,144 @@ def enqueue_task_command(
...
+    *,
+    task_id: int,
+    actor_user_id: int | None,
+    command_id: str,
+    kind: TaskCommandKind,
+    payload: dict[str, Any],
+) -> EnqueuedTaskCommand:
+    """Commit an idempotent command and return only after it is durable."""
+
+    try:
```
rogercloud, line 448: Two small notes on this block. (1) The pre-flush at ~258-263 exists specifically to keep the caller's own constraint failures out of command-conflict classification, but `db.commit()` is inside the same `try/except IntegrityError`, so a constraint firing only at COMMIT time (e.g. a deferrable FK) on the caller's own row would still be misrouted into `classify_task_command_conflict` -- and with a coincidental raced duplicate present, could be silently swallowed. No deferrable constraint exists in the current schema, so this is latent, and the commit message shows the tradeoff was weighed; a one-line comment acknowledging the residual conflation would preserve that reasoning in the code. (2) This function's docstring doesn't mention that `TaskCommandOwnerStateError` can propagate out uncaught. All three current call sites are safe (each uses `with SessionLocal() as db:`, whose exception-path close is an effective rollback), so no caller needs changing -- but documenting the escape would help a future caller that manages its own session lifetime.

tests/web/services/test_task_command_transport.py
```diff
@@ -1995,3 +2078,622 @@ def flush_raising_fk_error(*args, **kwargs):
...
+
+    assert winner.created is True
+    assert loser.created is False
+    assert loser.command_id == winner.command_id
+    assert loser.payload_matches is True
+    assert loser.status == winner.status
+
+
+@pytest.mark.postgresql
+def test_postgres_concurrent_insert_blocks_until_the_winner_resolves(
```
rogercloud, line 2496: This test is gated on `XAGENT_TEST_POSTGRES_URL`, which is unset in CI, so it only ever runs locally. That's a pre-existing repo-wide pattern rather than something this PR introduces -- two other Postgres-only test files have the same gap. Still, this test covers precisely the concurrent-insert race the refactor is meant to make safe, so it would be worth wiring into `.github/workflows/test-migrations.yml` alongside the other Postgres-only tests.

rogercloud (approved):
## Summary

This PR decomposes `enqueue_task_command()` into two composable primitives. `stage_task_command()` validates the request, resolves a task snapshot through a Core `select` (preceded by a pre-flush so the caller's own uncommitted writes are visible), runs an idempotency precheck, and adds+flushes the command row — but never commits, rolls back, or notifies the dispatcher. `classify_task_command_conflict()` handles post-rollback classification of an `IntegrityError` into `RACED_DUPLICATE` / `TASK_MISSING` / `UNRELATED`. `enqueue_task_command()` is rewritten as the production consumer of both primitives and preserves its exact external behavior — all three existing call sites (`a2a.py`, `websocket.py`, `workforce_runtime.py`) are zero-diff. This closes #1073, whose acceptance criteria require a caller-transaction staging primitive so a future lifecycle owner can insert a durable continuation command atomically with its own writes; that future consumer is explicitly out of scope here per the issue's own non-goals.

## Approach verdict: Sound, with reservations

The decomposition is the right shape for the stated problem, and the three unusual-looking design choices are each load-bearing rather than cargo-culted:

- **The Core `select` (bypassing the ORM identity map)** is necessary for callers that performed a bulk `.update(synchronize_session=False)` CAS — an ORM query would hand back the stale identity-mapped instance and the snapshot would be wrong.
- **The pre-flush before that Core select** is necessary for callers that mutated an ORM attribute directly. With `autoflush=False` (the production sessionmaker setting), a raw Core select would not otherwise observe the pending change.
- **`TaskCommandOwnerStateError` deliberately not subclassing `IntegrityError`** is necessary so that `enqueue_task_command`'s `except IntegrityError` does not misroute the caller's *own* constraint failure through duplicate/missing-task conflict classification.

Reservations, all design-level rather than defects:

1. **Surface area vs. current consumer count.** `classify_task_command_conflict`, `TaskCommandConflictKind`, `RacedTaskCommandProjection`, and `TaskCommandConflictClassification` are, as of this PR, only used by `enqueue_task_command` in the same module — no production caller outside this file invokes `stage_task_command` or `classify_task_command_conflict` directly. This is by design: #1073 explicitly names the wrapper as the required consumer in this PR and defers the real consumer. Noting it only as a cost/scope observation — nothing here should be removed. The counterweight is that the primitives' contracts get their first *external* exercise in a later PR, so the docstrings are effectively the whole interface spec until then, which raises their importance.
2. **Unenforceable ordering contract.** The `stage_task_command` docstring's rule — "this call must be the last write a caller issues before it commits, with no I/O or slow work in between" — is the single most important correctness invariant for any future direct caller, and there is no way to detect a violation at runtime. That's an inherent cost of the chosen design (prose-only contract), not something to fix in this PR, but worth the author's awareness for when the real consumer lands: a violation will manifest as lock contention or a stale snapshot, both hard to attribute back to call ordering.
3. **Forward-looking note on the staging lock window.** The PR's own tests demonstrate that holding a staged-but-uncommitted command row open causes SQLite's database-wide writer lock to block unrelated concurrent claims until `busy_timeout` expires. Today this window is microseconds because `enqueue_task_command` commits immediately after staging, and nothing propagates unhandled — `dispatch_task_command_promptly`'s shield + 50ms timeout + detached task swallows any resulting `OperationalError` through a done-callback that only logs, under the real 5000ms default `busy_timeout`. This is **not** a current bug. It is, however, worth one line in the `stage_task_command` docstring: a caller that stays in a long transaction after staging holds that writer lock for the entire transaction, not microseconds.

## Findings

No critical or major findings survived verification. Everything below is minor, documentation, or test hygiene — see inline comments for specifics on:

- Test assertion that doesn't distinguish "still pending" from "discarded by rollback" (`test_enqueue_idempotent_hit_issues_no_commit`)
- Two missing caller obligations in `stage_task_command`'s docstring (notify-after-commit, rollback-on-IntegrityError)
- Module docstring at line 1-6 is stale: "Ingress commits a command before acknowledging it" no longer describes the module's full public surface now that `stage_task_command` deliberately does not commit. Worth a sentence distinguishing the committing wrapper from the staging primitive (no single diff line owns this, since the docstring itself is unchanged pre-existing text)
- `command_id` normalization duplicated three ways with an unvalidated middle copy
- Duplicated field-mapping between `StagedTaskCommand` and `EnqueuedTaskCommand`
- No test for `command_id` format validation (pre-existing gap, but a good moment to add it)
- `postgres_task_command_sessions` fixture leaks global engine/sessionmaker state
- Flush-count-coupled monkeypatch fragility
- Residual commit-time `IntegrityError` conflation (latent — no deferrable constraint exists today)
- `TaskCommandOwnerStateError` propagation undocumented on `enqueue_task_command` (all current call sites are safe; docstring-only suggestion)
- New PostgreSQL concurrent-insert test never runs in CI (pre-existing repo-wide pattern, not a regression)

## Test results

Ran the full test file locally: **68 passed, 1 skipped, 0 failures**. The single skip is the PostgreSQL concurrent-insert test (no `XAGENT_TEST_POSTGRES_URL` configured). This matches the PR's own claim.

## Recommendation

**Approve.** No correctness defects survived verification. The decomposition is well-motivated, each of the three unusual design choices was confirmed load-bearing via mutation testing, and the existing call sites are genuinely zero-diff. Every remaining finding is minor, documentation, or test hygiene — the docstring items are the ones most worth picking up before the real external consumer lands, since those docstrings are currently the entire interface contract for the new primitives.

tests/web/services/test_task_command_transport.py
AlexLiu190625: Fixed in 066b314e. The owner pre-flush inside stage_task_command flushes any pre-existing pending change, so asserting the original sibling write's dirty/new state after the call is vacuous by construction. The test now adds a fresh pending write after the idempotent hit, asserts it is still in session.new, then commits and verifies both writes persisted — which does distinguish still-pending from discarded-by-rollback.

src/xagent/web/services/task_command_transport.py
AlexLiu190625: Fixed in 066b314e: the docstring now states all three obligations — notify after commit (else polling-latency degradation), full rollback before any further statement on IntegrityError, and the writer-lock span for a caller that stays in a long transaction after staging.

src/xagent/web/services/task_command_transport.py
AlexLiu190625: Fixed in 066b314e: extracted _normalize_command_id (strip + full-match validation + ValueError) and routed all three sites through it; classify_task_command_conflict's docstring now notes it validates format.

src/xagent/web/services/task_command_transport.py
AlexLiu190625: Fixed in 066b314e: added StagedTaskCommand.to_enqueued() and replaced both hand-written copies.

src/xagent/web/services/task_command_transport.py
AlexLiu190625: Fixed in 066b314e: added a malformed-id ValueError test (asserting no DB write) and a whitespace-normalization test.

tests/web/services/test_task_command_transport.py
AlexLiu190625: Fixed in 066b314e: the fixture captures the prior engine/sessionmaker binding and restores it in teardown.

tests/web/services/test_task_command_transport.py
AlexLiu190625: Fixed in 066b314e: both sites now carry a comment explaining the coupling to the two-flush sequence and the failure mode if the sequence changes.

src/xagent/web/services/task_command_transport.py
AlexLiu190625: Fixed in 066b314e: enqueue_task_command's docstring documents the uncaught TaskCommandOwnerStateError propagation, and the except block carries a comment acknowledging the residual commit-time conflation (no deferrable constraint exists today; parity with the original single-function behavior is deliberate).

tests/web/services/test_task_command_transport.py
AlexLiu190625: Agreed on the value. Deferred only for mechanical reasons: an open sibling change is adding the named-step + path-filter scaffolding to that same workflow block, and editing it here too would conflict. Will wire this file in immediately after that lands.

AlexLiu190625: Reconsidered: wired in this PR after all (the sibling change can absorb the workflow-file conflict on its side). Commit adds the transport module and test file to both path filters and a marker-selected step in the PostgreSQL migrations job; verified locally that -m postgresql collects exactly the concurrent-insert test.

PR #1117, opened 2026-08-03
fix(checkpoint): distinguish transient load failures from missing checkpoints

### LINKED ISSUE #978: fix(agent): transient checkpoint load failures are indistinguishable from missing checkpoints, silently corrupting resume
## Summary

A transient database error while loading the latest checkpoint is indistinguishable from "no checkpoint exists". During resume of a `waiting_for_user` task, this causes the runtime to silently discard the checkpointed execution state and continue on a fresh empty context (web/external path) or to fall back to PAUSED + APPEND replan (A2A path), instead of failing loudly or staying in the waiting state. Neither outcome is observable: no error is raised to the caller and no degradation signal is registered.

## Root cause

`DatabaseTraceHandler.load_latest_checkpoint` (`src/xagent/web/api/trace_handlers.py:101-117`) catches any exception from the whole query and returns `None`:

- A connection failure / statement timeout and "no matching checkpoint rows" both surface as `None` to every caller.
- Only a single warning log line is emitted. Per-row decode failures nearby (`trace_handlers.py:148-182`) do call `register_degradation(...)` and fall back to an older readable checkpoint, but the whole-query failure path has no such signal.

## Consequence A — web/external resume silently loses execution state

`execute_resume_background` performs two independent checkpoint loads:

1. The first load happens inside `post_user_message` (message injection). This step is guarded: if injection fails, a `RuntimeError` is raised and the task is marked FAILED (`src/xagent/web/api/websocket.py:2092-2106`). Safe.
2. Regardless of step 1, `resume_execution_by_id` then calls `AgentRunner.resume()`, which performs a second, independent `_load_latest_checkpoint` (`src/xagent/core/agent/runner.py:282`). If this one fails transiently:
   - `run()` takes the `else` branch and builds a fresh empty context (`runner.py:83-88` → `_build_context`). In this path `task` is `None`, so not even the original task text is re-injected.
   - `run()` always returns a normal result dict, so the `result is None` guard in `execute_resume_background` never fires.

Net effect: execution position, DAG node state, and the completed-tool ledger are silently discarded; the task keeps running on an empty context with no error, no retry, and no user-visible signal. If the pattern subsequently re-plans a step with side effects (send email, place order, write files), that side effect can be executed twice.

## Consequence B — A2A falls back to replan on transient errors

The comment at `src/xagent/web/api/a2a.py:426-428` states the invariant "Only a missing checkpoint falls back to the durable APPEND/replan path". This invariant does not hold: `_resume_waiting_a2a_task` receives `posted=False` for both "checkpoint genuinely missing" and "transient DB error swallowed to None" (`runner.py:316-323`), and converts the task to PAUSED + APPEND replan (`a2a.py:244-256`, `a2a.py:425-438`), discarding the checkpointed execution position.

## Proposed fix

1. Make the checkpoint load API tri-state instead of collapsing to `None`: `found` / `authoritatively absent` (query succeeded, zero rows) / raise a dedicated error (e.g. `CheckpointUnavailableError`) on query failure, and register a degradation signal for the error case (mirroring the per-row decode fallback).
2. `AgentRunner.run(resume=True)` / `resume()` must not silently build a fresh context on the error state; propagate the error so callers hit the same guard that already exists for message-injection failures (task FAILED with a visible error, or retry).
3. A2A: only take the PAUSED + APPEND fallback on authoritative absence; on the error state, keep `WAITING_FOR_USER` and surface a retryable error to the client.

## Additional context

The upcoming HITL interaction design (`InteractionStateRepairService`) assumes exactly this "authoritatively absent vs transient failure" distinction from the checkpoint store, so this fix is a prerequisite investment for that work as well. Since checkpoints live in the primary database (`trace_events`, same session factory as `tasks`), a successful query returning zero rows is already an authoritative absence proof — the change is API shape, not storage capability.

AlexLiu190625:
## Summary

A transient database error while loading the latest checkpoint is currently indistinguishable from "no checkpoint exists": the loader collapses every failure to `None`, and resume paths interpret that as authoritative absence — silently rebuilding an empty context (discarding execution position, DAG state, and the completed-tool ledger, with the risk of re-executing side effects) or reporting a permanent condition for a retryable one.

This change makes checkpoint reads a typed contract and threads it through every consumer:

- `CheckpointReadError(RuntimeError)` with `CheckpointUnavailableError` (transient: session checkout, query, or partition-resolution failure; registers a self-clearing degradation signal), `CheckpointCorruptError` (matching rows exist but none decodes; malformed or snapshot-less payloads), and `CheckpointAccessRefusedError` (the run-partition policy refuses the read — including when a run-tagged checkpoint provably exists — which is not absence). `None` now means exactly one thing: the query succeeded and the reader's own partition has no matching row. The matching predicate is pushed into SQL so the row window bounds matching rows, and a window full of undecodable rows classifies as unavailable rather than corrupt.
- Pass-through layers (`Tracer`, `TraceCheckpointStore`, `read_latest_checkpoint_payload`, the agent-tool delegate) propagate the contract verbatim; the tracer returns the first capable reader's result instead of probing later handlers. Reader-less tracer stacks keep their existing absent semantics.
- `AgentRunner` never builds a fresh context on a failed read, and the message-injection baseline read happens before any context mutation, so a rejected injection leaves no dedupe-matching residue.
- Web resume: the lease acquisition path captures the pre-acquisition status, and a transient read failure restores the task to that status (instead of settling it as FAILED) through the exact-lease fence, including when A2A hands over a pre-acquired lease. Database-pool exhaustion keeps its existing no-write TTL-recovery path, in which case the task recovers as PAUSED. The live-lease injection point degrades unavailable reads to the existing deferred-delivery path and rejects corrupt/refused ones without orphaning the delivery claim.
- A2A input-required: unavailable maps to a retryable 503 while the task stays `waiting_for_user`; corrupt and refused map to distinct terminal 400s; all three restore the exact prelease first.

## Behavior changes

- Resume paths no longer continue on an empty context after a transient read failure; they surface a retryable error and preserve the waiting state (or PAUSED via TTL recovery under pool exhaustion).
- A2A distinguishes retryable unavailability (503) from absence and corruption (400s) instead of one uniform message.
- `/health` gains a `CHECKPOINT_LOAD_UNAVAILABLE` degradation signal that self-clears on the next successful read.
- Authoritative-absence behavior is unchanged: legitimate absent checkpoints keep today's semantics everywhere.

## Verification

- 134 tests across the checkpoint/runner/websocket/A2A surfaces (18 net-new plus updated fakes), including: four-state translation at the loader (query failure, zero rows, all-rows-undecodable, partition refusal vs absence), pass-through pins at each intermediate layer, no-fresh-context and zero-residue runner pins, restore-to-prior-status for both self-acquired and handed-over leases, pool-timeout ordering with exception-chain preservation, deferred-delivery degradation and claim non-orphaning, and per-state A2A error codes.
- Mutation checks: restoring the original catch-all, conflating refusal with absence, swapping the window-full classification, reordering the baseline read, and disabling the restore branch each turn specific tests red.
- `tests/web` fully green; full-suite failures are a strict subset of the base commit's pre-existing set (verified against a clean checkout at the same base).

Fixes #978.

### DIFF
--- a/src/xagent/core/agent/checkpoint.py
+++ b/src/xagent/core/agent/checkpoint.py
@@ -58,9 +58,10 @@ class CheckpointReadError(RuntimeError):
 class CheckpointUnavailableError(CheckpointReadError):
     """Raised when a checkpoint read could not be completed.
 
-    Covers query failures (session checkout, transient DB errors) and
-    decode failures that cannot be proven to be permanent corruption
-    because the candidate row window was exhausted before ruling it out.
+    Covers infrastructure failures only: session checkout, query
+    execution, and generic per-row decode errors such as a failed blob
+    prefetch. Rows that decode as permanently unreadable are classified
+    by the corrupt error once the matching set is exhausted.
     """
 
 
--- a/src/xagent/core/agent/checkpoint.py
+++ b/src/xagent/core/agent/checkpoint.py
@@ -232,6 +232,20 @@ class TraceCheckpointStore:
             "snapshot": dict(payload),
         }
 
+    def _validate_snapshot(self, container: dict[str, Any]) -> dict[str, Any]:
+        """Return the readable checkpoint's snapshot, or raise if absent.
+
+        A container whose ``checkpoint_type`` is readable but that carries no
+        ``snapshot`` dict claims to be a checkpoint while holding no payload
+        -- corrupt, not absent.
+        """
+        snapshot = container.get("snapshot")
+        if isinstance(snapshot, dict):
+            return dict(snapshot)
+        raise CheckpointCorruptError(
+            "Checkpoint payload has a readable checkpoint_type but no snapshot."
+        )
+
     def _unwrap_checkpoint_payload(self, payload: Any) -> dict[str, Any] | None:
         # ``None`` is the reader's authoritative "no checkpoint" -- pass it
         # through unchanged. Anything else that isn't a recognized shape is
@@ -244,23 +258,13 @@ class TraceCheckpointStore:
                 "Checkpoint reader returned a non-dict, non-None payload."
             )
         if payload.get("checkpoint_type") in READABLE_CHECKPOINT_TYPES:
-            snapshot = payload.get("snapshot")
-            if isinstance(snapshot, dict):
-                return dict(snapshot)
-            raise CheckpointCorruptError(
-                "Checkpoint payload has a readable checkpoint_type but no snapshot."
-            )
+            return self._validate_snapshot(payload)
         data = payload.get("data")
         if (
             isinstance(data, dict)
             and data.get("checkpoint_type") in READABLE_CHECKPOINT_TYPES
         ):
-            snapshot = data.get("snapshot")
-            if isinstance(snapshot, dict):
-                return dict(snapshot)
-            raise CheckpointCorruptError(
-                "Checkpoint payload has a readable checkpoint_type but no snapshot."
-            )
+            return self._validate_snapshot(data)
         if payload.get("type") == "checkpoint" or "context" in payload:
             return dict(payload)
         raise CheckpointCorruptError(
--- a/src/xagent/core/agent/checkpoint.py
+++ b/src/xagent/core/agent/checkpoint.py
@@ -44,6 +44,46 @@ class CheckpointPersistenceError(RuntimeError):
     """Raised when a checkpoint cannot be durably persisted."""
 
 
+class CheckpointReadError(RuntimeError):
+    """Base for checkpoint read failures that must not collapse to absence.
+
+    ``None`` from a checkpoint reader means "queried successfully, nothing
+    found" — an authoritative fact callers may act on (e.g. build a fresh
+    context). Anything that prevented that determination raises one of the
+    subclasses below instead, so a transient or refused read can never be
+    mistaken for a checkpoint that genuinely does not exist.
+    """
+
+
+class CheckpointUnavailableError(CheckpointReadError):
+    """Raised when a checkpoint read could not be completed.
+
+    Covers query failures (session checkout, transient DB errors) and
+    decode failures that cannot be proven to be permanent corruption
+    because the candidate row window was exhausted before ruling it out.
+    """
+
+
+class CheckpointCorruptError(CheckpointReadError):
+    """Raised when matching checkpoint rows exist but none are usable.
+
+    Distinct from ``CheckpointUnavailableError``: the read completed and
+    the candidate set was proven exhausted, but every row is permanently
+    undecodable or shaped without a payload. This is a terminal state, not
+    a retryable one.
+    """
+
+
+class CheckpointAccessRefusedError(CheckpointReadError):
+    """Raised when a reader is not authoritative for the requested partition.
+
+    The checkpoint may exist, but this reader's partition (run binding,
+    build scope) is not the one allowed to observe it right now. Distinct
+    from absence: callers must not treat a refusal as "no checkpoint" and
+    fall back to building fresh state.
+    """
+
+
 async def read_latest_checkpoint_payload(
     reader: Any,
     execution_id: str,
--- a/src/xagent/core/agent/checkpoint.py
+++ b/src/xagent/core/agent/checkpoint.py
@@ -81,7 +81,20 @@ class CheckpointAccessRefusedError(CheckpointReadError):
     build scope) is not the one allowed to observe it right now. Distinct
     from absence: callers must not treat a refusal as "no checkpoint" and
     fall back to building fresh state.
+
+    ``reason`` discriminates *why* the read was refused, so consumers can
+    report an accurate message instead of one generic sentence for every
+    case: ``"lease_mismatch"`` (an active lease exists but is not bound to
+    this reader), ``"active_run"`` (a different run is in progress under
+    its own lease), or ``"superseded_legacy"`` (a tagged run has already
+    superseded the untagged/legacy partition this reader is confined to).
+    Defaults to ``"active_run"``, the most common case, so existing call
+    sites that do not pass ``reason`` keep working unchanged.
     """
+
+    def __init__(self, message: str, *, reason: str = "active_run") -> None:
+        super().__init__(message)
+        self.reason = reason
 
 
 async def read_latest_checkpoint_payload(
--- a/src/xagent/core/agent/checkpoint.py
+++ b/src/xagent/core/agent/checkpoint.py
@@ -220,21 +220,39 @@ class TraceCheckpointStore:
         }
 
     def _unwrap_checkpoint_payload(self, payload: Any) -> dict[str, Any] | None:
-        if not isinstance(payload, dict):
+        # ``None`` is the reader's authoritative "no checkpoint" -- pass it
+        # through unchanged. Anything else that isn't a recognized shape is
+        # corrupt, not absent: a caller must not treat a malformed payload
+        # as "build fresh state".
+        if payload is None:
             return None
+        if not isinstance(payload, dict):
+            raise CheckpointCorruptError(
+                "Checkpoint reader returned a non-dict, non-None payload."
+            )
         if payload.get("checkpoint_type") in READABLE_CHECKPOINT_TYPES:
             snapshot = payload.get("snapshot")
-            return dict(snapshot) if isinstance(snapshot, dict) else None
+            if isinstance(snapshot, dict):
+                return dict(snapshot)
+            raise CheckpointCorruptError(
+                "Checkpoint payload has a readable checkpoint_type but no snapshot."
+            )
         data = payload.get("data")
         if (
             isinstance(data, dict)
             and data.get("checkpoint_type") in READABLE_CHECKPOINT_TYPES
         ):
             snapshot = data.get("snapshot")
-            return dict(snapshot) if isinstance(snapshot, dict) else None
+            if isinstance(snapshot, dict):
+                return dict(snapshot)
+            raise CheckpointCorruptError(
+                "Checkpoint payload has a readable checkpoint_type but no snapshot."
+            )
         if payload.get("type") == "checkpoint" or "context" in payload:
             return dict(payload)
-        return None
+        raise CheckpointCorruptError(
+            "Checkpoint payload shape is not recognized by any reader."
+        )
 
     def _execution_id(self, payload: dict[str, Any]) -> str:
         execution_id = payload.get("execution_id")
--- a/src/xagent/web/api/a2a.py
+++ b/src/xagent/web/api/a2a.py
@@ -157,6 +157,7 @@ async def _schedule_waiting_a2a_resume(
     task_lease: TaskLease,
     heartbeat_stop: asyncio.Event,
     heartbeat_task: asyncio.Task[TaskLeaseHeartbeatOutcome],
+    resumable_status: TaskStatus,
 ) -> None:
     from .websocket import background_task_manager, execute_resume_background
 
@@ -177,6 +178,10 @@ async def _schedule_waiting_a2a_resume(
                 preacquired_lease=task_lease,
                 preacquired_heartbeat_stop=heartbeat_stop,
                 preacquired_heartbeat_task=heartbeat_task,
+                # The prelease claimed this task out of an input-required
+                # status; hand that over so a checkpoint the resume cannot
+                # read restores it instead of failing it terminally.
+                preacquired_prior_status=resumable_status,
             )
         )
         background_task_manager.register_reserved_resume(task_id, bg_task)
@@ -444,6 +449,7 @@ async def _resume_input_required_a2a_task(
             task_lease=task_lease,
             heartbeat_stop=heartbeat_stop,
             heartbeat_task=heartbeat_task,
+            resumable_status=resumable_status,
         )
         ownership_transferred = True
     except (
--- a/src/xagent/web/api/a2a.py
+++ b/src/xagent/web/api/a2a.py
@@ -456,12 +456,15 @@ async def _resume_input_required_a2a_task(
         # Ownership was never transferred, so restore the exact prelease
         # to the prior input-required status exactly like the absent-
         # checkpoint fallback above, then translate the failure instead of
-        # letting it escape as a raw exception.
-        cleanup_task = asyncio.create_task(stop_and_restore_prelease())
-        if not await drain_async_task_cancellation_safe(cleanup_task):
-            raise TaskLeaseLostError(
-                f"Task {task_id} lease changed before A2A checkpoint-failure fallback"
-            ) from exc
+        # letting it escape as a raw exception. Same ownership_transferred/
+        # prelease_cleanup_done guard as the sibling except BaseException
+        # handler below, for symmetry.
+        if not ownership_transferred and not prelease_cleanup_done:
+            cleanup_task = asyncio.create_task(stop_and_restore_prelease())
+            if not await drain_async_task_cancellation_safe(cleanup_task):
+                raise TaskLeaseLostError(
+                    f"Task {task_id} lease changed before A2A checkpoint-failure fallback"
+                ) from exc
         if isinstance(exc, CheckpointCorruptError):
             raise a2a_error(
                 "unsupported_operation",
@@ -470,18 +473,19 @@ async def _resume_input_required_a2a_task(
                 details={"taskId": task_id},
             ) from exc
         if isinstance(exc, CheckpointAccessRefusedError):
-            if exc.reason == "lease_mismatch":
-                message = (
+            message = {
+                "lease_mismatch": (
                     "This task is currently owned by a different execution "
                     "and cannot accept a new message."
-                )
-            elif exc.reason == "superseded_legacy":
-                message = (
+                ),
+                "superseded_legacy": (
                     "This task's checkpoint history has been superseded by "
                     "a newer run and cannot accept a new message."
-                )
-            else:
-                message = "Task is currently running and cannot accept a new message."
+                ),
+            }.get(
+                exc.reason,
+                "Task is currently running and cannot accept a new message.",
+            )
             raise a2a_error(
                 "unsupported_operation",
                 message,
--- a/src/xagent/web/api/a2a.py
+++ b/src/xagent/web/api/a2a.py
@@ -12,6 +12,11 @@ from fastapi.responses import StreamingResponse
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from sqlalchemy import String, and_, cast, func, or_, update
 
+from ...core.agent.checkpoint import (
+    CheckpointAccessRefusedError,
+    CheckpointCorruptError,
+    CheckpointUnavailableError,
+)
 from ..models.agent import Agent
 from ..models.database import get_session_local
 from ..models.task import Task, TaskStatus
@@ -441,6 +446,40 @@ async def _resume_input_required_a2a_task(
             heartbeat_task=heartbeat_task,
         )
         ownership_transferred = True
+    except (
+        CheckpointUnavailableError,
+        CheckpointCorruptError,
+        CheckpointAccessRefusedError,
+    ) as exc:
+        # Ownership was never transferred, so restore the exact prelease
+        # to the prior input-required status exactly like the absent-
+        # checkpoint fallback above, then translate the failure instead of
+        # letting it escape as a raw exception.
+        cleanup_task = asyncio.create_task(stop_and_restore_prelease())
+        if not await drain_async_task_cancellation_safe(cleanup_task):
+            raise TaskLeaseLostError(
+                f"Task {task_id} lease changed before A2A checkpoint-failure fallback"
+            ) from exc
+        if isinstance(exc, CheckpointUnavailableError):
+            raise a2a_error(
+                "temporarily_unavailable",
+                "The task's saved progress could not be read. Please retry.",
+                status_code=503,
+                details={"taskId": task_id},
+            ) from exc
+        if isinstance(exc, CheckpointCorruptError):
+            raise a2a_error(
+                "unsupported_operation",
+                "The task's saved progress is unreadable.",
+                status_code=400,
+                details={"taskId": task_id},
+            ) from exc
+        raise a2a_error(
+            "unsupported_operation",
+            "Task is currently running and cannot accept a new message.",
+            status_code=400,
+            details={"taskId": task_id},
+        ) from exc
     except BaseException:
         if not ownership_transferred and not prelease_cleanup_done:
             cleanup_task = asyncio.create_task(stop_and_restore_prelease())
--- a/src/xagent/web/api/a2a.py
+++ b/src/xagent/web/api/a2a.py
@@ -15,7 +15,7 @@ from sqlalchemy import String, and_, cast, func, or_, update
 from ...core.agent.checkpoint import (
     CheckpointAccessRefusedError,
     CheckpointCorruptError,
-    CheckpointUnavailableError,
+    CheckpointReadError,
 )
 from ..models.agent import Agent
 from ..models.database import get_session_local
@@ -452,11 +452,7 @@ async def _resume_input_required_a2a_task(
             resumable_status=resumable_status,
         )
         ownership_transferred = True
-    except (
-        CheckpointUnavailableError,
-        CheckpointCorruptError,
-        CheckpointAccessRefusedError,
-    ) as exc:
+    except CheckpointReadError as exc:
         # Ownership was never transferred, so restore the exact prelease
         # to the prior input-required status exactly like the absent-
         # checkpoint fallback above, then translate the failure instead of
@@ -466,13 +462,6 @@ async def _resume_input_required_a2a_task(
             raise TaskLeaseLostError(
                 f"Task {task_id} lease changed before A2A checkpoint-failure fallback"
             ) from exc
-        if isinstance(exc, CheckpointUnavailableError):
-            raise a2a_error(
-                "temporarily_unavailable",
-                "The task's saved progress could not be read. Please retry.",
-                status_code=503,
-                details={"taskId": task_id},
-            ) from exc
         if isinstance(exc, CheckpointCorruptError):
             raise a2a_error(
                 "unsupported_operation",
@@ -480,10 +469,34 @@ async def _resume_input_required_a2a_task(
                 status_code=400,
                 details={"taskId": task_id},
             ) from exc
+        if isinstance(exc, CheckpointAccessRefusedError):
+            if exc.reason == "lease_mismatch":
+                message = (
+                    "This task is currently owned by a different execution "
+                    "and cannot accept a new message."
+                )
+            elif exc.reason == "superseded_legacy":
+                message = (
+                    "This task's checkpoint history has been superseded by "
+                    "a newer run and cannot accept a new message."
+                )
+            else:
+                message = "Task is currently running and cannot accept a new message."
+            raise a2a_error(
+                "unsupported_operation",
+                message,
+                status_code=400,
+                details={"taskId": task_id},
+            ) from exc
+        # CheckpointUnavailableError, or any future CheckpointReadError
+        # subclass this dispatch does not yet know about: treat it
+        # conservatively as retryable rather than assuming a terminal or
+        # policy failure, so an unrecognized failure mode never silently
+        # collapses into a data-losing branch.
         raise a2a_error(
-            "unsupported_operation",
-            "Task is currently running and cannot accept a new message.",
-            status_code=400,
+            "temporarily_unavailable",
+            "The task's saved progress could not be read. Please retry.",
+            status_code=503,
             details={"taskId": task_id},
         ) from exc
     except BaseException:
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -57,6 +57,31 @@ logger = logging.getLogger(__name__)
 CHECKPOINT_ROW_SCAN_LIMIT = 100
 
 
+def _checkpoint_execution_id_predicate(execution_id: str) -> Any:
+    """SQL mirror of ``checkpoint_execution_id()``: root wins, then the flat
+    field, then the snapshot's own id -- so legacy rows that only set one of
+    them are not skipped. The read query and history pruning must agree on
+    which rows belong to one execution, or pruning could drop a row the read
+    path still considers current (or vice versa); both consume this one
+    predicate rather than keeping independently maintained copies in sync by
+    hand.
+    """
+    return (
+        func.coalesce(
+            func.nullif(
+                DatabaseTraceEvent.data["root_execution_id"].as_string(),
+                "",
+            ),
+            func.nullif(
+                DatabaseTraceEvent.data["execution_id"].as_string(),
+                "",
+            ),
+            DatabaseTraceEvent.data["snapshot"]["execution_id"].as_string(),
+        )
+        == execution_id
+    )
+
+
 def _convert_float_to_datetime(timestamp: Any) -> datetime:
     """Convert float timestamp to datetime for database storage."""
     if isinstance(timestamp, (int, float)):
@@ -155,22 +180,11 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 DatabaseTraceEvent.data["checkpoint_type"]
                 .as_string()
                 .in_(sorted(READABLE_CHECKPOINT_TYPES)),
-                # SQL mirror of checkpoint_execution_id(): root wins, then
-                # the flat field, then the snapshot's own id. The LIMIT
-                # below bounds this predicate's matching set, not an
-                # unfiltered row scan, so a zero-row result is authoritative.
-                func.coalesce(
-                    func.nullif(
-                        DatabaseTraceEvent.data["root_execution_id"].as_string(),
-                        "",
-                    ),
-                    func.nullif(
-                        DatabaseTraceEvent.data["execution_id"].as_string(),
-                        "",
-                    ),
-                    DatabaseTraceEvent.data["snapshot"]["execution_id"].as_string(),
-                )
-                == str(execution_id),
+                # The page size below bounds this predicate's matching set,
+                # not an unfiltered row scan, so a page shorter than it
+                # proves the matching set exhausted (see the scan loop
+                # below), and a zero-row first page is authoritative.
+                _checkpoint_execution_id_predicate(str(execution_id)),
             )
             if self.build_id is None:
                 try:
@@ -586,21 +600,7 @@ class DatabaseTraceHandler(BaseTraceHandler):
                     DatabaseTraceEvent.data["checkpoint_type"]
                     .as_string()
                     .in_(sorted(READABLE_CHECKPOINT_TYPES)),
-                    # SQL mirror of checkpoint_execution_id(): root wins,
-                    # then the flat field, then the snapshot's own id, so
-                    # legacy rows that only set one of them are not skipped.
-                    func.coalesce(
-                        func.nullif(
-                            DatabaseTraceEvent.data["root_execution_id"].as_string(),
-                            "",
-                        ),
-                        func.nullif(
-                            DatabaseTraceEvent.data["execution_id"].as_string(),
-                            "",
-                        ),
-                        DatabaseTraceEvent.data["snapshot"]["execution_id"].as_string(),
-                    )
-                    == execution_id,
+                    _checkpoint_execution_id_predicate(execution_id),
                 )
                 .order_by(
                     DatabaseTraceEvent.timestamp.desc(),
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -305,6 +305,10 @@ class DatabaseTraceHandler(BaseTraceHandler):
                     # A short page proves the matching set is exhausted --
                     # no further rows exist beyond this one.
                     break
+                # OFFSET paging over (timestamp DESC, id DESC): a row inserted
+                # mid-scan shifts later pages by one. Root reads are fenced to a
+                # single run partition, which excludes concurrent writers there;
+                # build-scoped histories are append-only per build.
                 offset += CHECKPOINT_ROW_SCAN_LIMIT
 
             if not saw_any_row:
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -13,6 +13,9 @@ from ...config import get_checkpoint_history_limit
 from ...core.agent.checkpoint import (
     CHECKPOINT_TYPE,
     READABLE_CHECKPOINT_TYPES,
+    CheckpointAccessRefusedError,
+    CheckpointCorruptError,
+    CheckpointUnavailableError,
     checkpoint_execution_id,
 )
 from ...core.agent.trace import BaseTraceHandler
@@ -26,6 +29,7 @@ from ...web.models.task import TraceEvent as DatabaseTraceEvent
 from ...web.models.tool_config import ToolUsage
 from ...web.services.ops_signals import (
     CHECKPOINT_DECODE_FALLBACK,
+    CHECKPOINT_LOAD_UNAVAILABLE,
     clear_degradation,
     register_degradation,
 )
@@ -42,6 +46,12 @@ from ...web.services.trace_message_storage import (
 )
 
 logger = logging.getLogger(__name__)
+
+# Candidate rows scanned per checkpoint read. A window this size lets a few
+# unreadable rows fall back to an older readable one without a second query;
+# whether it was large enough to prove the whole matching set was scanned
+# governs the unavailable-vs-corrupt classification below.
+CHECKPOINT_ROW_SCAN_LIMIT = 100
 
 
 def _convert_float_to_datetime(timestamp: Any) -> datetime:
@@ -105,35 +115,62 @@ class DatabaseTraceHandler(BaseTraceHandler):
     async def load_latest_checkpoint(
         self, execution_id: str
     ) -> Optional[Dict[str, Any]]:
-        """Load the latest agent checkpoint persisted as a trace event."""
-        try:
-            return await asyncio.to_thread(
-                self._sync_load_latest_checkpoint,
-                execution_id,
-            )
-        except Exception as e:
-            logger.warning(
-                "Failed to load latest checkpoint for task %s execution %s: %s",
-                self.task_id,
-                execution_id,
-                e,
-            )
-            return None
+        """Load the latest agent checkpoint persisted as a trace event.
+
+        ``None`` means the query completed and found nothing -- an
+        authoritative fact. Anything that prevented that determination
+        (query failure, refused partition, undecodable rows) is translated
+        to a ``CheckpointReadError`` subclass inside the sync worker below
+        and propagates through here unchanged; it must never collapse back
+        to ``None``, or a transient failure would be indistinguishable from
+        "no checkpoint" to every caller up the stack.
+        """
+        return await asyncio.to_thread(
+            self._sync_load_latest_checkpoint,
+            execution_id,
+        )
 
     def _sync_load_latest_checkpoint(
         self,
         execution_id: str,
     ) -> Optional[Dict[str, Any]]:
-        db = next(get_db())
+        try:
+            db = next(get_db())
+        except Exception as exc:
+            register_degradation(
+                CHECKPOINT_LOAD_UNAVAILABLE,
+                f"task {self.task_id}: checkpoint session checkout failed",
+            )
+            raise CheckpointUnavailableError(
+                f"task {self.task_id}: could not open a database session "
+                "to read the checkpoint"
+            ) from exc
         try:
             query = db.query(DatabaseTraceEvent).filter(
                 DatabaseTraceEvent.task_id == self.task_id,
                 DatabaseTraceEvent.event_type == "system_update_general",
+                DatabaseTraceEvent.data["checkpoint_type"]
+                .as_string()
+                .in_(sorted(READABLE_CHECKPOINT_TYPES)),
+                # SQL mirror of checkpoint_execution_id(): root wins, then
+                # the flat field, then the snapshot's own id. The LIMIT
+                # below bounds this predicate's matching set, not an
+                # unfiltered row scan, so a zero-row result is authoritative.
+                func.coalesce(
+                    func.nullif(
+                        DatabaseTraceEvent.data["root_execution_id"].as_string(),
+                        "",
+                    ),
+                    func.nullif(
+                        DatabaseTraceEvent.data["execution_id"].as_string(),
+                        "",
+                    ),
+                    DatabaseTraceEvent.data["snapshot"]["execution_id"].as_string(),
+                )
+                == str(execution_id),
             )
             if self.build_id is None:
-                allowed, run_id = self._root_checkpoint_read_partition(db)
-                if not allowed:
-                    return None
+                run_id = self._root_checkpoint_read_partition(db)
                 query = query.filter(
                     DatabaseTraceEvent.build_id.is_(None),
                     self._checkpoint_run_partition_filter(run_id),
@@ -141,20 +178,34 @@ class DatabaseTraceHandler(BaseTraceHandler):
             else:
                 query = query.filter(DatabaseTraceEvent.build_id == self.build_id)
 
-            rows = (
-                query.order_by(
-                    DatabaseTraceEvent.timestamp.desc(),
-                    DatabaseTraceEvent.id.desc(),
+            try:
+                rows = (
+                    query.order_by(
+                        DatabaseTraceEvent.timestamp.desc(),
+                        DatabaseTraceEvent.id.desc(),
+                    )
+                    .limit(CHECKPOINT_ROW_SCAN_LIMIT)
+                    .all()
                 )
-                .limit(100)
-                .all()
-            )
+            except Exception as exc:
+                register_degradation(
+                    CHECKPOINT_LOAD_UNAVAILABLE,
+                    f"task {self.task_id}: checkpoint query failed",
+                )
+                raise CheckpointUnavailableError(
+                    f"task {self.task_id}: checkpoint query failed"
+                ) from exc
+            # The query itself succeeded -- whatever the decode loop below
+            # concludes about the rows it found, the read infrastructure is
+            # healthy again.
+            clear_degradation(CHECKPOINT_LOAD_UNAVAILABLE)
+            if not rows:
+                return None
+
+            saw_generic_failure = False
+            saw_undecodable_row = False
             for row in rows:
                 data: Dict[str, Any] = row.data if isinstance(row.data, dict) else {}
-                if data.get("checkpoint_type") not in READABLE_CHECKPOINT_TYPES:
-                    continue
-                if checkpoint_execution_id(data) != str(execution_id):
-                    continue
                 try:
                     data = decode_trace_event_data(
                         db,
@@ -163,6 +214,7 @@ class DatabaseTraceHandler(BaseTraceHandler):
                         strict=True,
                     )
                 except CheckpointMessageDecodeError as exc:
+                    saw_undecodable_row = True
                     logger.warning(
                         "Skipping unreadable checkpoint trace event %s for task %s: %s",
                         row.event_id,
@@ -177,6 +229,7 @@ class DatabaseTraceHandler(BaseTraceHandler):
                     # degradation on /health so a systemic decode failure is
                     # observable instead of only a per-row warning log; the
                     # signal self-clears on the next successful decode.
+                    saw_generic_failure = True
                     register_degradation(
                         CHECKPOINT_DECODE_FALLBACK,
                         f"task {self.task_id}: checkpoint decode failed, "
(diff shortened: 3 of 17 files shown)


src/xagent/core/agent/checkpoint.py
```diff
@@ -180,21 +220,39 @@ def _event_payload(
...
         ):
             snapshot = data.get("snapshot")
-            return dict(snapshot) if isinstance(snapshot, dict) else None
+            if isinstance(snapshot, dict):
+                return dict(snapshot)
+            raise CheckpointCorruptError(
+                "Checkpoint payload has a readable checkpoint_type but no snapshot."
+            )
         if payload.get("type") == "checkpoint" or "context" in payload:
             return dict(payload)
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in 58f5943b. Every checkpoint writer persists a context dict, so a recognized payload without one is malformed rather than absent. The resume path now raises CheckpointCorruptError instead of falling through to a fresh context, and message injection raises instead of returning the absent-style None. The write-baseline read is deliberately unchanged: a malformed stored baseline there is overwritten by the next full persist, which is self-healing rather than a resume decision. Pinned by two tests with a contextless fake store.

src/xagent/web/api/websocket.py
```diff
@@ -2768,6 +2786,27 @@ def _acquire_resume_task_lease(
         return lease
 
 
+def _restore_resumed_task_lease_to_prior_status(
```
rogercloud, line 2800: **MAJOR:** `prior_status_out`/the captured prior status here can be `RUNNING`. A TTL-expiry lease reclaim (`task_lease_service.py:514`) can steal a lease while `Task.status` is still `RUNNING` (e.g. a crashed worker whose row was never transitioned off RUNNING). If that reclaimed resume then hits `CheckpointUnavailableError`/`CheckpointAccessRefusedError`, `_restore_resumed_task_lease_to_prior_status` calls `release_task_lease_no_commit(db, lease, status=RUNNING)`, which raises `ValueError("Cannot release a task lease with RUNNING status")` (`task_lease_service.py:687-688`, pre-existing). That exception is swallowed by the broad `except Exception` in the finally block (see `:3656-3677`), which logs a misleading "retaining lease for TTL recovery" message — the lease is never released and now carries a freshly extended TTL, with no `task_error` broadcast either. The PR's headline safety property ("restore prior status instead of failing") silently doesn't hold for exactly the reclaim case it's meant to handle.

Suggestion: branch explicitly on `RUNNING` before calling `release_task_lease_no_commit` — route to the settle/FAILED path (which does broadcast) instead of relying on the catch-all to absorb the ValueError.

src/xagent/web/api/websocket.py
```diff
@@ -3449,6 +3506,36 @@ async def mark_deferred_delivery_failed() -> bool:
             # otherwise left reclaimable when no lease was acquired). Do not
             # emit the generic FAILED/task_error payload below.
             return
+        elif (
```
rogercloud, line 3522: This is the call site that feeds a `RUNNING` prior status into the restore path — see the detailed note on `_restore_resumed_task_lease_to_prior_status` (`:2789`). When `resume_prior_status` is `RUNNING`, the restore silently no-ops (lease retained, no broadcast) instead of actually restoring or failing loudly.

src/xagent/web/api/websocket.py
```diff
@@ -3570,6 +3657,28 @@ async def finalize_resume_resources() -> None:
                     lease is not None
                     and not lease_released
                     and not defer_db_cleanup_to_ttl_recovery
+                    and restore_lease_to_prior_status is not None
```
rogercloud, line 3677: **MINOR:** a successful restore-to-prior-status broadcasts nothing to connected clients, while the sibling settle/FAILED path (`:3679-3707`) broadcasts `create_terminal_task_error_event`. Since `task_resumed` was already sent earlier (telling clients the task is RUNNING), a client on this path sees no corrective event and its UI keeps showing running/resumed while the DB has actually moved the task back to `WAITING_FOR_USER`/`PAUSED`, until a manual refresh. Consider broadcasting a status-correction event (e.g. `task_restored`) alongside the restore.

src/xagent/web/api/websocket.py
```diff
@@ -2768,6 +2786,27 @@ def _acquire_resume_task_lease(
         return lease
 
 
+def _restore_resumed_task_lease_to_prior_status(
+    lease: TaskLease,
+    *,
+    status: TaskStatus,
+) -> bool:
+    """Release the exact resume lease back to its pre-acquisition status.
+
```
rogercloud, line 2806: **MINOR (docstring accuracy):** this commits unconditionally, with no `if restored: commit() else: rollback()` branch, unlike the cited mirror `_restore_a2a_resume_prelease_sync` (`a2a.py:255-266`) which does have that conditional. Harmless today — the underlying `release_task_lease_no_commit` UPDATE is fenced by `task_id + runner_id (+ run_id)`, so a lost fence yields a 0-row UPDATE whether committed or rolled back — but the "mirrors the A2A prelease restore" wording overstates the symmetry. Either add the conditional or soften the docstring.

src/xagent/web/api/a2a.py
```diff
@@ -439,8 +449,43 @@ async def inject_user_message() -> tuple[Any, bool]:
...
+                details={"taskId": task_id},
+            ) from exc
+        if isinstance(exc, CheckpointCorruptError):
+            raise a2a_error(
+                "unsupported_operation",
+                "The task's saved progress is unreadable.",
+                status_code=400,
+                details={"taskId": task_id},
+            ) from exc
+        raise a2a_error(
```
rogercloud, line 500: **MINOR/MODERATE:** this message is hardcoded for every `CheckpointAccessRefusedError`, including the superseded-legacy-checkpoint refusal raised at `trace_handlers.py:349` ("a tagged run has already superseded legacy checkpoints"), which has nothing to do with the task currently running. Callers get a misleading diagnosis for that case. Suggest carrying a reason discriminator on the exception (or distinct subclasses) and deriving the A2A message from it.

src/xagent/web/api/a2a.py
```diff
@@ -439,8 +449,43 @@ async def inject_user_message() -> tuple[Any, bool]:
             task_lease=task_lease,
             heartbeat_stop=heartbeat_stop,
             heartbeat_task=heartbeat_task,
+            resumable_status=resumable_status,
         )
         ownership_transferred = True
+    except (
```
rogercloud: **MINOR:** this catches the three concrete `CheckpointReadError` subclasses by name, whereas `websocket.py:5877` deliberately catches the base class with a comment explaining the future-proofing rationale. With only three subclasses today this works, but a future fourth subclass would escape this handler and surface as a generic 500 instead of a structured error (cleanup still runs correctly — this is missed classification, not a leak). Suggest catching the base `CheckpointReadError` here too, matching the `websocket.py` pattern.

src/xagent/web/api/a2a.py
```diff
@@ -439,8 +449,43 @@ async def inject_user_message() -> tuple[Any, bool]:
...
+        # Ownership was never transferred, so restore the exact prelease
+        # to the prior input-required status exactly like the absent-
+        # checkpoint fallback above, then translate the failure instead of
+        # letting it escape as a raw exception.
+        cleanup_task = asyncio.create_task(stop_and_restore_prelease())
+        if not await drain_async_task_cancellation_safe(cleanup_task):
+            raise TaskLeaseLostError(
+                f"Task {task_id} lease changed before A2A checkpoint-failure fallback"
+            ) from exc
+        if isinstance(exc, CheckpointUnavailableError):
```
rogercloud: `shrink`: these three isinstance/raise blocks are structurally identical aside from the `(error_code, message, status_code)` triple — could collapse into a dict-dispatch table keyed by exception type. Low priority, and arguably premature: `CheckpointAccessRefusedError` likely needs its own distinct branch soon (see the note at `:483` about the misleading message), which would fight against a generic table.

src/xagent/web/api/trace_handlers.py
```diff
@@ -43,6 +48,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Candidate rows scanned per checkpoint read. A window this size lets a few
+# unreadable rows fall back to an older readable one without a second query;
+# whether it was large enough to prove the whole matching set was scanned
+# governs the unavailable-vs-corrupt classification below.
+CHECKPOINT_ROW_SCAN_LIMIT = 100
```
rogercloud, line 57: **MINOR:** `CHECKPOINT_ROW_SCAN_LIMIT = 100` makes the `len(rows) >= LIMIT ⇒ unavailable` branch (around `:282`) dead code under default config — pruning caps history at 8 rows via `get_checkpoint_history_limit()`, so 100 rows for one execution is only reachable via a synthetic fixture or with pruning explicitly disabled (`XAGENT_CHECKPOINT_HISTORY_LIMIT=0`). Separately, if pruning is disabled and a task accumulates 100+ persistently undecodable rows, this branch can never reach a terminal `corrupt` classification — A2A returns 503 "retry" forever and web resume restores prior status forever. Suggest deriving this limit from `get_checkpoint_history_limit()` (or documenting the relationship), and classifying an all-undecodable full window as `corrupt` rather than `unavailable`.

src/xagent/core/agent/checkpoint.py
```diff
@@ -180,21 +220,39 @@ def _event_payload(
...
+        # through unchanged. Anything else that isn't a recognized shape is
+        # corrupt, not absent: a caller must not treat a malformed payload
+        # as "build fresh state".
+        if payload is None:
             return None
+        if not isinstance(payload, dict):
+            raise CheckpointCorruptError(
+                "Checkpoint reader returned a non-dict, non-None payload."
+            )
         if payload.get("checkpoint_type") in READABLE_CHECKPOINT_TYPES:
```
rogercloud, line 261: `shrink`: the two snapshot-validation blocks here (validating `payload`'s snapshot, then `payload["data"]`'s) are near-identical — same `checkpoint_type` gate, same `isinstance(snapshot, dict)` check, same `CheckpointCorruptError` message. Extract a single `_validate_snapshot(candidate)` helper and call it twice.

rogercloud (changes requested):
## Summary

Checkpoint loading previously collapsed every failure mode — transient DB errors, corrupt/undecodable rows, partition-access refusals, and genuine absence — into a single `None`. Resume paths therefore treated an infrastructure hiccup as "no checkpoint," rebuilding an empty context (risking re-executed side effects) or, in A2A, dropping into a replan path. This PR introduces a `CheckpointReadError` family (`CheckpointUnavailableError`/retryable, `CheckpointCorruptError`/terminal, `CheckpointAccessRefusedError`/wrong-partition) and threads it verbatim through `Tracer` → `TraceCheckpointStore` → `AgentRunner` → web resume (`websocket.py`) → A2A (`a2a.py`), with per-transport handling: web resume restores the task to its pre-acquisition status instead of FAILED; A2A maps unavailable to 503 and corrupt/refused to 400.

The four-state contract is the right decomposition, minimal, and matches the scope agreed on issue #978. Reservations concern the fourth state and the restore mechanism, not the taxonomy. `CheckpointAccessRefusedError` carries two genuinely different meanings (active-run-elsewhere vs. superseded-legacy-checkpoint at `src/xagent/web/api/trace_handlers.py:349`) — the per-transport retry/reject divergence is reasoned and documented at each site, not a defect, but the A2A error message doesn't account for it (see inline comment on `a2a.py:483`).

Separately, the out-parameter status channel (`prior_status_out` in `websocket.py`) is a workaround for a fixed signature. It has exactly one real consumer (`a2a.py:184`) and is wired correctly, so this is only a style note — a small return-value type would be more robust than an out-parameter if this grows a second caller. The new `CHECKPOINT_LOAD_UNAVAILABLE` degradation signal also clears globally on any successful read, so a single persistently failing task can be masked by healthy traffic elsewhere; this mirrors the pre-existing `CHECKPOINT_DECODE_FALLBACK` behavior and is adequate for a fleet-level signal, but per-task scoping would be a reasonable follow-up.

## Findings

See inline comments for line-specific findings. Two cross-cutting findings that don't map to a single line:

**MAJOR — test coverage gap tied directly to the `websocket.py:2789` restore-path finding.** Every restore-path test in `tests/web/api/test_execution_scope_turn_wiring.py` uses `TaskStatus.WAITING_FOR_USER` as the prior status **and** mocks `_restore_resumed_task_lease_to_prior_status` with `MagicMock(return_value=True)` — the real function, the one that raises `ValueError` for a `RUNNING` prior status, is never executed by any test in this PR. No `execute_resume_background` test covers `CheckpointAccessRefusedError` in this branch either (only Unavailable/Corrupt are exercised), and no test asserts a positive post-restore signal — only negative absence-of-`task_error` checks, which can't distinguish "correctly silent" from "silently broken." Suggest adding a real (unmocked) `RUNNING`-prior-status restore test (it should fail today), a `CheckpointAccessRefusedError` case, and a positive-signal assertion.

**MINOR/MODERATE — execution-id matching is reimplemented three times, with a real divergence on absent fields.** Once in Python (`src/xagent/core/agent/checkpoint.py`'s `checkpoint_execution_id()`), twice in hand-written SQL (`src/xagent/web/api/trace_handlers.py`'s read query and again in `_prune_checkpoint_history`), with comments in both SQL sites claiming to "mirror" the Python helper but no test asserting parity. When the execution-id fields are entirely absent (missing keys, not present-but-empty), the Python helper defaults to `""` via a hardcoded fallback, while the SQL `coalesce(...)` chain yields `NULL`, and `NULL = ''` is not true in SQL — a checkpoint carrying no execution-id information anywhere is non-matching in SQL but has canonical id `""` in Python. Suggest extracting one shared query-builder helper, or at minimum adding a parity test over absent/empty/snapshot-only/tagged-only cases.

- `src/xagent/core/agent/checkpoint.py:233-250`: two near-identical snapshot-validation blocks in `_unwrap_checkpoint_payload` (validating `payload`'s snapshot, then `payload["data"]`'s). Extract one `_validate_snapshot(candidate)` helper called twice.
- `src/xagent/web/api/a2a.py:469-488`: three structurally identical `isinstance`/raise blocks could collapse into a dict-dispatch table keyed by exception type. Low priority, and arguably premature — `CheckpointAccessRefusedError` likely needs its own distinct branch soon (see comment on `a2a.py:483`), which would fight the table.

net: -4 lines possible (the value here is removing literal duplication, not line count).

## Prior-finding re-review

One prior finding, from `gemini-code-assist[bot]` (HIGH, `checkpoint.py:252` at an earlier revision): `_unwrap_checkpoint_payload` did not validate that `context` was a non-empty dict, so a corrupt/missing context read silently as absence.

**Status: FIXED.** The fix was applied at the two `runner.py` call sites rather than inside `_unwrap_checkpoint_payload` — a deliberate placement per the author's reply. `run()` now raises `CheckpointCorruptError` when resuming from a checkpoint with a non-dict `context`, and the message-injection path raises instead of returning `None`. Two new tests (`test_run_resume_raises_corrupt_on_contextless_checkpoint`, `test_inject_user_message_raises_corrupt_on_contextless_checkpoint`) pin the behavior and were confirmed to fail when the fix is reverted, so they are not vacuous. Landed in `6c430f85` and still present at head.

## Test results

Full changed-test-file set (`tests/core/agent/test_checkpoint.py`, `tests/core/agent/test_runner.py`, `tests/web/api/test_a2a_api.py`, `tests/web/api/test_execution_scope_turn_wiring.py`, `tests/web/api/test_websocket_owner_actor.py`, `tests/web/test_agent_checkpoint_stream.py`, `tests/web/test_trace_message_storage.py`) on Python 3.12: **277 passed, 0 failed.**

## Categories reviewed with no issues found

- A2A resource lifecycle and cleanup — no double-stop, no leaked heartbeat.
- Handler-probe regression risk in `Tracer.load_latest_checkpoint` — no `BaseTraceHandler` override to account for.
- Exception swallowing along service → adapter → registry → runner — the `CheckpointReadError` family propagates cleanly to the intended catch points.
- Refused-partition logic leaking into normal resume — prevented by lease-context binding.
- Checkpoint payload shape compatibility with existing writers — all current writers always include `context`, so the new corrupt-checkpoint raises cannot misfire on legitimately written payloads.
- Documentation and configuration — no new env var, no user-facing docs change needed.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in e22cbb8d. The restore target is now gated to {PAUSED, WAITING_FOR_USER} at its single assignment site; a RUNNING prior falls through to the settle path, which broadcasts the terminal error. Pinned by a real-DB test (no mocks on the lease/restore path) that reproduced the exact ValueError swallow before the fix and asserts FAILED + released runner + one task_error after.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in e22cbb8d together with the :2789 thread — the RUNNING feeder now takes the settle/FAILED path with a broadcast instead of the silent no-op.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in e22cbb8d: a successful restore now broadcasts the existing status event for the restored state (task_waiting_for_user / task_paused, same vocabulary and versioning as the historical-replay path, state_version outranks the earlier task_resumed). Best-effort: a broadcast failure logs and never affects the restore. Positive-signal assertions added.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in e22cbb8d: the docstring no longer claims symmetry with the A2A prelease restore and states the unconditional commit plus the fenced zero-row-update property explicitly.

src/xagent/web/api/a2a.py
AlexLiu190625: Fixed in a16eb001: CheckpointAccessRefusedError now carries a reason (lease_mismatch / active_run / superseded_legacy) set distinctly at all three raise sites, and the A2A handler derives an accurate message per reason; active_run keeps the previous text verbatim. The web not_accepted path is reason-agnostic by design.

src/xagent/web/api/a2a.py
AlexLiu190625: Fixed in a16eb001: the handler catches the CheckpointReadError base and dispatches by isinstance; an unknown future subclass takes the retryable 503 branch (conservative: the task stays waiting). Pinned by a synthetic-subclass test.

src/xagent/web/api/a2a.py
AlexLiu190625: Not applied, per your own note: the refused branch now diverges further with the reason-derived messages from the :483 thread, which would fight a generic dispatch table. Keeping the explicit branches.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in ef41981b: the scan now pages through the matching set to exhaustion (the constant is a page size, not a cap). A readable row in any later page is returned; an exhausted matching set that failed only with permanent decode errors classifies as corrupt (ending the retry-forever loop under disabled pruning); any generic failure stays unavailable. Multi-batch tests cover all three outcomes, and mutation checks confirm each classification is load-bearing.

src/xagent/core/agent/checkpoint.py
AlexLiu190625: Fixed in 6bb381ec: extracted _validate_snapshot, called at both former blocks.

src/xagent/web/api/websocket.py
```diff
@@ -5746,15 +5908,41 @@ async def finish_existing_delivery(
...
+                            except CheckpointUnavailableError:
+                                # Fold into the existing posted=False path
+                                # below: the durable message is deferred to
+                                # the resume owner instead of injected live,
+                                # exactly as when there was no exact lease
+                                # or checkpoint to inject into. Distinct
+                                # from corrupt/refused, which are not
+                                # retryable by simply deferring.
+                                posted = False
+                            except CheckpointReadError:
```
rogercloud, line 5944: **MAJOR (new):** `except CheckpointReadError` here calls `finish_delivery(False, "The task's saved progress could not be read.", rejection_outcome="not_accepted")` instead of `finish_delivery_failure(...)`. `finish_delivery` only sends the WS rejection frame and sets an in-memory error marker — it never writes to the DB. `finish_delivery_failure` is what actually marks the row `DELIVERY_FAILED`. Confirmed empirically: the persisted `delivery_status` stays `DELIVERY_PENDING` after this handler runs, and there is no reaper/TTL cleanup for stale pending rows anywhere in the codebase. A client retry with the same `client_message_id` hits `finish_existing_delivery`'s `claim.pending` branch and gets "The message is still being applied. Please retry shortly." indefinitely — permanent limbo, while the client was already told `not_accepted`. Every sibling pre-dispatch rejection in this file either calls `finish_delivery_failure` or passes `retry_with_new_id=True`; this is the one exception. This has been present since the PR's first commit and survived the prior review round untouched — it is not a regression from the recent fix commits.

Suggestion: call `finish_delivery_failure("The task's saved progress could not be read.")` here (or mark `DELIVERY_FAILED` explicitly and pass `retry_with_new_id=True` to `finish_delivery`), and extend `test_live_lease_injection_rejects_delivery_on_corrupt_or_refused` to assert `delivery_status == DELIVERY_FAILED` (see `test_resume_registration_failure_keeps_injected_delivery_pending` for the existing assertion pattern).

src/xagent/web/api/websocket.py
```diff
@@ -3449,6 +3508,40 @@ async def mark_deferred_delivery_failed() -> bool:
             # otherwise left reclaimable when no lease was acquired). Do not
             # emit the generic FAILED/task_error payload below.
             return
```
rogercloud, line 3521: **LOW / follow-up (new, non-blocking):** this is the correct fix for the prior RUNNING-prior stranded-lease bug — RUNNING must never be a release target for `release_task_lease_no_commit`, and routing it to settle/FAILED is safe and tested. However, `release_task_lease_no_commit` does accept `PAUSED`, which already means "interrupted mid-run, non-terminal, resumable" elsewhere in this codebase. So a task whose checkpoint is perfectly intact — only the read transiently failed — is terminally failed today when it could be released to PAUSED and stay resumable. Triggering requires a live task interrupted by a new message (the common case) whose deferred resume then independently hits a checkpoint-read failure — a compound, narrow condition, so this is a real but non-blocking resilience gap.

Suggestion (follow-up): special-case `resume_prior_status == RUNNING` to release to `PAUSED` with the same corrective broadcast as the existing PAUSED/WAITING_FOR_USER path, rather than settling to FAILED.

src/xagent/web/api/trace_handlers.py
```diff
@@ -105,119 +143,230 @@ async def _save_to_database(self, event: CoreTraceEvent) -> None:
...
-                data: Dict[str, Any] = row.data if isinstance(row.data, dict) else {}
-                if data.get("checkpoint_type") not in READABLE_CHECKPOINT_TYPES:
-                    continue
-                if checkpoint_execution_id(data) != str(execution_id):
-                    continue
+
+            saw_generic_failure = False
+            saw_undecodable_row = False
+            saw_any_row = False
+            offset = 0
```
rogercloud, line 233: **LOW (new):** the paging fix here correctly removes the non-terminating classification from the prior review round, but page count is unbounded — there's no max-page guard. With `XAGENT_CHECKPOINT_HISTORY_LIMIT=0` and a large backlog of matching rows, a single checkpoint read can issue proportionally many sequential paged queries before exhausting the set (O(matching_row_count) queries per read). Still a clear improvement over the prior bug and shouldn't be reverted, but worth either a max-page guard (cap hit -> unavailable) or a documented operational caveat for the disabled-pruning case.

src/xagent/web/api/trace_handlers.py
```diff
@@ -105,119 +143,230 @@ async def _save_to_database(self, event: CoreTraceEvent) -> None:
...
+                offset += CHECKPOINT_ROW_SCAN_LIMIT
+
+            if not saw_any_row:
+                return None
+            # The matching set is exhausted and every candidate row failed.
+            # A generic (transient) failure anywhere in the scan is
+            # conservatively unavailable -- retryable. Only a fully scanned
+            # set that is exclusively permanent decode failures is corrupt.
+            if saw_generic_failure:
+                raise CheckpointUnavailableError(
```
rogercloud, line 359: **LOW (new):** this `CheckpointUnavailableError` raise (exhausted matching set, generic failure seen) isn't paired with `register_degradation(CHECKPOINT_LOAD_UNAVAILABLE)`, unlike the other three raise sites of this exception in the file (~168, ~202, ~234). Traced the `/health` impact: not a silent blind spot — `CHECKPOINT_DECODE_FALLBACK` is reliably still active whenever this branch fires, so `/health` does reflect degradation, just under a different signal name. A naming/consistency inconsistency rather than a monitoring gap. Low priority: register `CHECKPOINT_LOAD_UNAVAILABLE` here too for consistency with the other three sites.

src/xagent/web/api/a2a.py
```diff
@@ -439,8 +449,56 @@ async def inject_user_message() -> tuple[Any, bool]:
             task_lease=task_lease,
             heartbeat_stop=heartbeat_stop,
             heartbeat_task=heartbeat_task,
+            resumable_status=resumable_status,
         )
         ownership_transferred = True
+    except CheckpointReadError as exc:
+        # Ownership was never transferred, so restore the exact prelease
+        # to the prior input-required status exactly like the absent-
```
rogercloud, line 457: **NIT (new):** this `except CheckpointReadError` handler's cleanup call doesn't check `ownership_transferred`/`prelease_cleanup_done` the way the sibling `except BaseException` handler does. Confirmed structurally impossible to trigger today (the only site setting `ownership_transferred = True` is the last statement of the try block, after which nothing in that block can still raise `CheckpointReadError`), and `stop_and_restore_prelease()` has its own internal idempotency check as a second layer. Pure defense-in-depth for future refactors — optional, add the same guard for symmetry if you'd like.

src/xagent/web/api/a2a.py
```diff
@@ -439,8 +449,56 @@ async def inject_user_message() -> tuple[Any, bool]:
...
+            ) from exc
+        if isinstance(exc, CheckpointCorruptError):
+            raise a2a_error(
+                "unsupported_operation",
+                "The task's saved progress is unreadable.",
+                status_code=400,
+                details={"taskId": task_id},
+            ) from exc
+        if isinstance(exc, CheckpointAccessRefusedError):
+            if exc.reason == "lease_mismatch":
```
rogercloud: `shrink` (new): the three-branch reason-to-message mapping reads more directly as a dict literal — `{"lease_mismatch": ..., "superseded_legacy": ...}.get(reason, default)` — instead of if/elif/else.

src/xagent/web/api/websocket.py
```diff
@@ -3570,6 +3663,75 @@ async def finalize_resume_resources() -> None:
...
+                            # may still be showing after the lease claim
+                            # above flipped it, before this failure restored
+                            # the prior status. Best-effort: a missed
+                            # broadcast does not change the restore result
+                            # that already committed.
+                            try:
+                                restored_snapshot = (
+                                    await task_execution_controller.snapshot(task_id)
+                                )
+                                event_type = (
```
rogercloud: `shrink` (new): this `task_waiting_for_user`/`task_paused` event-type/message ternary duplicates the same ternary already built for the historical-replay path around `websocket.py:7068-7084`. Consider extracting one small shared helper.

rogercloud (changes requested):
## Update summary

Five commits (`e22cbb8d`, `a16eb001`, `ef41981b`, `6bb381ec`, `be37338d`) since the previous review on `6c430f85`. Collectively they gate the restore target to non-RUNNING statuses, add a `reason` discriminator to `CheckpointAccessRefusedError` with per-reason A2A messaging, page the checkpoint row scan to exhaustion (adding a terminal `corrupt` classification), broadcast state after a successful restore, deduplicate the execution-id matching predicate across Python/SQL, switch A2A to base-class exception handling, and extract snapshot validation. All 9 numbered prior findings are resolved and both prior simplification suggestions are addressed (one applied, one consciously declined with sound reasoning). A fresh independent pass over the current head surfaced one new MAJOR issue plus several low-severity items, none of which are regressions from the fix commits.

## Design verdict: Sound

Upgraded from "acceptable with reservations" last round — the reservations that drove that verdict, primarily the restore-mechanism gaps, are now fixed. The failure taxonomy is the right shape, correctly placed at the only layer that can distinguish the cases, and the propagation is a plain, traceable exception hierarchy rather than a bolted-on result-union.

One scope observation, not a defect: the entire `CheckpointAccessRefusedError` arm (the reason discriminator, the per-reason A2A messages, half of the restore gate) is unreachable in production today — every checkpoint read happens inside a lease-bound context where the refusal preconditions cannot occur, so this machinery is currently exercised only by tests. Not wrong to have built it (it matches the issue's scope and may serve a future delegated-subagent handler), but it accounts for a real share of this revision's diff for code that cannot yet fire outside tests. Worth being aware of; not blocking.

## Prior-finding status

| # | Prior finding | Status | Verifying evidence |
|---|---|---|---|
| 1 | MAJOR — restore-to-prior-status raised `ValueError` and stranded the lease when prior status was RUNNING (`websocket.py`) | Fixed | Restore target gated to `{PAUSED, WAITING_FOR_USER}` at its single assignment site (`websocket.py:3508-3520`); a captured RUNNING status falls through to the settle/FAILED path, which does broadcast a terminal error. No other call site can feed RUNNING in — the A2A handover path (`a2a.py:184`) pre-filters to the same two statuses at `a2a.py:327-331`. Verified by a real-DB test with no mocks on the lease/restore path: `tests/web/api/test_websocket_owner_actor.py:2822` `test_resume_background_settles_running_prior_status_on_checkpoint_unavailable`, asserting the task ends FAILED, the lease is actually released, and exactly one `task_error` broadcasts. 87 relevant tests pass. |
| 2 | MINOR/MODERATE — hardcoded "task is currently running" message for every `CheckpointAccessRefusedError`, including the superseded-legacy case (`a2a.py:483`) | Fixed | Error now carries `reason` (`lease_mismatch` / `active_run` / `superseded_legacy`), set distinctly at all three raise sites in `trace_handlers.py` — including the originally misdiagnosed `superseded_legacy` site, which now uses that reason rather than `active_run`. `a2a.py` derives a distinct message per reason; the `active_run` message is byte-identical to the original text. New parametrized test `tests/web/api/test_a2a_api.py:1194` `test_checkpoint_access_refused_reason_gets_a_distinct_message` asserts the message does not contain "currently running" for the other two reasons. |
| 3 | MAJOR — no test exercised the real restore path with a RUNNING prior status; no `CheckpointAccessRefusedError` coverage in the restore branch; no positive post-restore signal assertions | Fixed | Covered by the real-DB test in #1. See #4 for the one remaining assertion gap. |
| 4 | MINOR — successful restore broadcast nothing, leaving the UI stuck on "running" (`websocket.py:3660`) | Fixed, minor residual gap | Successful restore now takes a fresh task-control snapshot and broadcasts `task_waiting_for_user`/`task_paused` (`websocket.py:3679-3730`), reusing the historical-replay path's event vocabulary and `state_version` versioning, so a client that already processed an earlier `task_resumed` correctly supersedes it. Broadcast is best-effort in its own try/except — confirmed by `test_resume_background_restore_broadcast_failure_does_not_affect_restore`. Residual: no test asserts the positive content of the broadcast (event type, `state_version` presence); only that a broadcast is attempted and failures are swallowed. Worth a follow-up test on the payload; the core fix is correct. |
| 5 | MINOR/MODERATE — execution-id matching predicate reimplemented 3x (Python + 2x SQL) with a real divergence on absent fields (`trace_handlers.py`) | Fixed | Shared `_checkpoint_execution_id_predicate` now backs both the read query and the prune query, with a parity test against the Python helper. |
| 6 | MINOR — `CHECKPOINT_ROW_SCAN_LIMIT` made the window-full branch dead code under default config, with no route to a terminal `corrupt` classification when pruning was disabled (`trace_handlers.py:55`) | Fixed | Scan now pages through the matching set in batches of `CHECKPOINT_ROW_SCAN_LIMIT` (correctly re-documented as a page size, not a cap) until exhaustion; a readable row in any later page is returned; exhausting all pages with only permanent decode failures raises `CheckpointCorruptError`, closing the retry-forever gap; any generic/transient failure anywhere in the scan still raises `CheckpointUnavailableError`. Side benefit: fixes a latent pre-existing bug where up to 100 recent non-checkpoint rows could mask a real checkpoint row. Multi-batch tests cover all outcomes and pass. See the new finding on `trace_handlers.py`'s scan loop below for a follow-on cost consideration. |
| 7 | MINOR — A2A caught concrete `CheckpointReadError` subclasses by name instead of the base class, unlike `websocket.py` (`a2a.py:455`) | Fixed | Now catches base `CheckpointReadError` and dispatches via `isinstance`; an unrecognized future subclass takes the retryable-503 branch by design (conservative — keeps the task waiting rather than risking a wrong terminal classification). Pinned by a synthetic-subclass test. |
| 8 | MINOR — restore-helper docstring overstated symmetry with the A2A prelease restore's commit/rollback handling (`websocket.py:2789`) | Fixed | Docstring no longer claims "mirrors," and now documents the unconditional commit plus why the fenced zero-row-update makes it safe. |
| 9 | MINOR — no terminal classification when the scan window filled with persistently undecodable rows (`trace_handlers.py`) | Fixed | Same paging-to-exhaustion fix as #6. |
| 10 | Simplification — duplicated snapshot-validation blocks (`checkpoint.py:233`) | Applied | Extracted `_validate_snapshot`, called at both former sites; behavior-preserving (confirmed by diff comparison); 17/17 tests in `tests/core/agent/test_checkpoint.py` pass. |
| 11 | Simplification — three near-identical isinstance/raise blocks in `a2a.py:469`, dict-dispatch suggested | Not applied — correct call | The refused branch now diverges further with reason-derived messages, which would fight a generic dispatch table. Consistent with the caveat attached to the original suggestion; not a gap. |

## Test results

Full changed-test-file set on Python 3.12 at current head (`be37338d`): **292 passed, 0 failed.**

## Overall recommendation

One more small fix-and-verify round before merge. The new finding below (`websocket.py:5930`) is a real, previously-uncaught MAJOR bug — narrow in scope (a message delivery gets stuck in limbo; no data loss and no execution-state corruption) but with no recovery path once hit, and the fix is a one-line handler change plus one test assertion. The other new findings are follow-up or optional and should not gate this PR.

Otherwise this PR is in very good shape: every previously-identified concern is resolved with verifiable evidence, the design verdict is upgraded to Sound, and the test suite is green.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in 329faf11: the handler now calls finish_delivery_failure, matching the sibling pre-dispatch rejections, so the row is durably marked failed and a same-id retry gets a fresh attempt instead of the pending loop. The existing rejects-delivery test now also asserts the persisted delivery_status == DELIVERY_FAILED; verified red ('pending' == 'failed') against the prior code for both corrupt and refused parametrizations before the fix.

src/xagent/web/api/websocket.py
AlexLiu190625: Agreed and recorded as a follow-up per your gating: releasing a RUNNING prior to PAUSED (with the same corrective broadcast) keeps an intact-checkpoint task resumable instead of terminal. Kept out of this round to keep the merge window small; tracked with the program follow-ups.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 5c513721: added a documented max-page bound; hitting the cap classifies as retryable unavailable with the degradation signal registered. Pinned by a test with a tiny page size and cap.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 5c513721: the exhausted-with-generic-failure raise now registers the same degradation signal as the other three sites; the branch test asserts the signal is active after the raise (red before the fix).

src/xagent/web/api/a2a.py
AlexLiu190625: Fixed in bc6202e1: the handler now carries the same ownership/cleanup guard as the sibling handler.

src/xagent/web/api/a2a.py
AlexLiu190625: Fixed in bc6202e1: dict-literal mapping with the default in .get().

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in bc6202e1: extracted a shared helper for the event type and message; both the restore path and the historical-replay path call it, pinned by a unit test on its output plus the existing integration assertions at both sites. The corrective-broadcast test also gained positive payload assertions (event type and state_version), closing the residual noted in the review body.

src/xagent/web/api/trace_handlers.py
```diff
@@ -105,119 +151,253 @@ async def _save_to_database(self, event: CoreTraceEvent) -> None:
...
+                            self.task_id,
+                            exc_info=True,
+                        )
+                        continue
+                    clear_degradation(CHECKPOINT_DECODE_FALLBACK)
+                    snapshot = data.get("snapshot")
+                    if not isinstance(snapshot, dict):
+                        # The row claims to be a checkpoint but carries no
+                        # payload -- corrupt, not absent.
+                        raise CheckpointCorruptError(
```
rogercloud: **MEDIUM (new, empirically confirmed):** this branch raises `CheckpointCorruptError` immediately, aborting the entire multi-row scan -- while the sibling undecodable-row branch (around line 286) sets `saw_undecodable_row = True` and `continue`s to the next, older row, matching the scan's own stated design of falling through candidate rows to find an older good one.

Reproduced with a two-row fixture ordered exactly as the real query orders rows (timestamp/id DESC): newest row has a readable `checkpoint_type` but no `snapshot` (malformed); an older row is fully valid with a readable snapshot. The scan raises on the first row and never returns the older valid snapshot -- the loop never reaches it. Downstream this is terminal: `CheckpointCorruptError` is documented non-retryable, mapping to A2A 400 and websocket-resume FAILED. Net effect: a task with a perfectly good older checkpoint becomes terminally unresumable because of one malformed newer row -- precisely what the sibling branch and the page-continuation design exist to survive. This is a risk introduced by this PR's own scan logic (the pre-PR code had no multi-row scan at all), and no existing test covers the two-row shadowing case -- the closest test has a single malformed row with nothing behind it.

Suggested fix: `saw_undecodable_row = True; continue` here instead of raising, consistent with the sibling branch and with the exhausted-set logic that already correctly reports `corrupt` only when nothing readable exists across the entire scan. Please add a two-row (malformed-newest, valid-older) regression test alongside it.

rogercloud (changes requested):
## Update summary

Three new commits (`329faf11`, `5c513721`, `bc6202e1`) landed since the round-2 review of `be37338d`. Collectively they fix the MAJOR delivery-state bug, bound the paged checkpoint scan and make its degradation signalling consistent, and apply both simplification suggestions plus the missing positive assertions on the restore broadcast. All six numbered prior findings are now resolved: five fixed, one (the RUNNING->PAUSED release) correctly deferred as an explicitly non-blocking tracked follow-up. One new MEDIUM finding surfaced in a fresh independent pass over the current head and was empirically reproduced.

## Design verdict: Sound

The four-way contract (absent / unavailable / corrupt / refused) is the minimum vocabulary the consumers actually branch on -- none of the four could be collapsed without losing a real decision. Placement is right: classification lives at the only reader that knows why, the vocabulary lives beside the checkpoint contract, and no consumer re-derives it. The `reason` discriminator on `refused` is a genuine semantic distinction the three A2A messages need, not a message-passing shortcut. Diff volume is high, but the paging loop is a single flat state machine with three boolean witnesses rather than a pile of special cases.

## Prior-finding status

1. **`src/xagent/web/api/websocket.py:5930` -- MAJOR, delivery stuck in DELIVERY_PENDING -- FIXED** (`329faf11`). The handler now calls `finish_delivery_failure("The task's saved progress could not be read.")`, matching the sibling pre-dispatch rejections, so the row lands in DELIVERY_FAILED. A retry with the same client_message_id now takes the `claim.failed` branch -- a decisive rejection with `retry_with_new_id=True` -- instead of the previous permanent "still being applied" limbo. The deliberately-retryable CheckpointUnavailableError path is unaffected and still stays DELIVERY_PENDING, confirmed by a separate passing test. `test_live_lease_injection_rejects_delivery_on_corrupt_or_refused` (parametrized corrupt + refused) asserts `delivery_status == DELIVERY_FAILED`; 9 relevant tests pass.

2. **`src/xagent/web/api/websocket.py:3510` -- LOW, RUNNING task settles terminal FAILED where PAUSED was legal -- DEFERRED (accepted).** This was explicitly non-blocking last round (a suggested follow-up, not a required fix). Keeping it out to hold the merge window small is a reasonable scope call, and it is tracked with the program follow-ups. No objection.

3. **`src/xagent/web/api/trace_handlers.py` -- LOW, unbounded page count in the checkpoint scan -- FIXED** (`5c513721`). `CHECKPOINT_SCAN_MAX_PAGES = 50` added and documented; hitting the cap registers CHECKPOINT_LOAD_UNAVAILABLE and raises CheckpointUnavailableError (retryable, not corrupt, no unbounded hang). Cap generosity checks out: 100 rows/page x 50 pages = 5000 matching rows, against ~8 rows for a normal execution under default pruning -- the cap only bites in the pathological pruning-disabled case it targets. A test with a monkeypatched page size/cap of 2/2 over 4 rows confirms the cap is hit and classified correctly; 6 relevant tests pass.

4. **`src/xagent/web/api/trace_handlers.py` -- LOW, missing register_degradation on the exhausted-set generic-failure raise -- FIXED** (`5c513721`). That raise site now registers CHECKPOINT_LOAD_UNAVAILABLE before raising, making all 4 of 4 CheckpointUnavailableError sites consistent; the two CheckpointCorruptError raises correctly do not register this signal. A dedicated test asserts the signal is active after this specific raise.

5. **`src/xagent/web/api/a2a.py:457` -- NIT, missing ownership guard on the CheckpointReadError cleanup -- FIXED** (`bc6202e1`). Now `if not ownership_transferred and not prelease_cleanup_done:`, identical to the sibling BaseException guard.

6. **`src/xagent/web/api/a2a.py:473` -- shrink, reason->message if/elif/else -- FIXED** (`bc6202e1`). Now a dict literal with `.get(exc.reason, default)`; messages are byte-identical to before -- pure restructuring.

7. **`src/xagent/web/api/websocket.py:3696` -- shrink + residual test gap -- FIXED** (`bc6202e1`). A shared `_waiting_or_paused_event_fields(status)` helper now backs both the restore path and the historical-replay path, with a unit test pinning its output for both statuses. The corrective-broadcast test gained positive assertions on the event type and state_version presence, closing the residual gap.

## New finding

See inline comment on `src/xagent/web/api/trace_handlers.py`.

## Additional observations (non-blocking)

- Scope, stated plainly: the recovery/restore improvements here change outcomes only for resumes whose prior status is PAUSED/WAITING_FOR_USER (A2A input-required, explicit paused resume). The dominant live-websocket shape -- task RUNNING, interrupted by a new message -- still settles FAILED on a transient read failure, per the deferred item #2. Worth saying explicitly so the transient-blip problem isn't read as solved end-to-end for all resume shapes.
- `ops_signals` stores signals globally rather than per task, so one task's successful read can clear a degradation signal another task's concurrent failure just set. Inherited from the pre-existing CHECKPOINT_DECODE_FALLBACK pattern, not a regression -- low priority.
- `src/xagent/web/api/websocket.py`: when the lease-restore fence loses the race (returns False, e.g. the TTL reaper won), there is no log and no distinct handling. Shape-consistent with the sibling settle branch, but a diagnostic would help on a brand-new code path.
- `runner.py`: a cold resume can issue two checkpoint-read queries (one in `resume()`, one in `run()` when the first returns None) -- pre-existing, but marginally more exposed now that reads can throw twice per attempt.
- Test-coverage nit: the A2A handover's `preacquired_prior_status` kwarg wiring is not exercised end-to-end by any single test (each half is covered separately), so removing it would regress silently with a green suite.
- `Tracer.load_latest_checkpoint`'s verbatim-return change is behavior-preserving today (only one checkpoint-capable handler exists in every current tracer stack). Residual: `AgentRunner._load_latest_checkpoint` still collapses a non-dict payload to None, which would matter only for a hypothetical future tracer stack not using TraceCheckpointStore.
- mypy over all six changed source files: zero errors attributable to this PR. Naming, resource lifecycle, concurrency/fencing, unrelated-path breakage, and behavior preservation in the refactored replay path: no issues found.

## Test results

Full changed-test-file set on Python 3.12 at `bc6202e1`: **294 passed, 0 failed** (independently reconfirmed as 144 + 60 + 90 across split runs).

## Recommendation

The new finding below is a real, empirically-confirmed MEDIUM bug, but narrower than last round's MAJOR: it needs a specific two-row malformed-then-valid ordering to trigger, and the fix is a one-line change (`raise` -> flag + `continue`) mirroring a sibling branch already present in the same function. Please fix it and add the two-row regression test before merge. Beyond that, this PR has resolved every previously-raised concern and is very close to mergeable.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 9310de57: the snapshotless branch now records the row as undecodable and continues the scan, matching the sibling decode-failure branches, so the corrupt verdict is decided only after the matching set is exhausted. The two-row shadowing case is pinned by `test_load_checkpoint_snapshotless_newest_row_does_not_shadow_older_valid` (malformed newest, valid older -> returns the older snapshot); the single-malformed-row case still ends in `CheckpointCorruptError` via the exhausted-set path. One consequence worth noting: a scan containing both a generic decode failure and a snapshotless row now classifies as unavailable (retryable) rather than corrupt, since the set verdict's existing conservative ordering applies — verified for both row orders, along with the cross-page variant (flag persists across pages).

src/xagent/web/api/websocket.py
```diff
@@ -3449,6 +3519,40 @@ async def mark_deferred_delivery_failed() -> bool:
             # otherwise left reclaimable when no lease was acquired). Do not
             # emit the generic FAILED/task_error payload below.
             return
+        elif (
+            isinstance(e, (CheckpointUnavailableError, CheckpointAccessRefusedError))
+            and lease is not None
```
rogercloud, line 3524: **Non-blocking design note (refinement of an already-deferred item):** RUNNING prior status falls through to the generic handler here and settles the task FAILED on a purely retryable checkpoint-read error (e.g. a live task interrupted by a new message, whose deferred resume re-claims the lease off the still-RUNNING row and then hits a second transient CheckpointUnavailableError). The pool-timeout branch just above already handles this class correctly via `defer_db_cleanup_to_ttl_recovery = True`, leaving the row RUNNING under the fenced lease for TTL recovery instead of terminally failing it. Extending that same treatment to `CheckpointUnavailableError`/`CheckpointAccessRefusedError` when the prior status is RUNNING would be cleaner than the earlier 'release to PAUSED' suggestion, and reuses a path already in this function. Noting this only as a refinement of the follow-up already tracked -- not a new ask for this PR.

src/xagent/web/api/websocket.py
```diff
@@ -2768,6 +2797,29 @@ def _acquire_resume_task_lease(
...
+    WHERE fence (task id + runner id + run id) as the TTL reaper, so the
+    two can never both release the same row -- whichever loses the race
+    affects zero rows instead of double-releasing. The commit below is
+    unconditional, unlike the A2A prelease restore: when the fence excludes
+    every row, the UPDATE affects zero rows and this commits that no-op
+    rather than rolling back.
+    """
+    SessionLocal = get_session_local()
+    with SessionLocal() as db:
+        restored = release_task_lease_no_commit(db, lease, status=status)
```
rogercloud, line 2818: **LOW/MEDIUM (new, empirically confirmed):** a successful restore-to-prior-status here does not clear a stale `error_message`/`output` from an earlier failed attempt. `release_task_lease_no_commit` (task_lease_service.py) only updates status/runner_id/lease_expires_at/last_heartbeat_at/control_state/state_version -- unlike its sibling `fail_and_release_task_lease_no_commit`, which explicitly sets/clears these fields. Nothing upstream clears them either (lease acquisition clears `error_message` only when `new_run=True`, which the resume path does not set). Confirmed empirically: a task carrying an `error_message` from an earlier failure, resumed and then successfully restored through this path after a second transient checkpoint-read failure, ends up healthy (WAITING_FOR_USER) but still holding the stale `error_message`. The existing corrective-broadcast test for this path never asserts on `error_message`, so nothing catches it -- and the new A2A test's `error_message is None` assertion is not coverage for this, since it passes because a different function (`_update_a2a_resume_input_sync`) clears the field earlier in that flow, not because this restore does. Since `error_message` is routinely surfaced to clients/UI, this is a real if narrow UX correctness bug. Suggestion: clear `error_message`/`output` in `release_task_lease_no_commit` (or at minimum at this call site) when transitioning to a non-terminal status, and add an assertion to the existing corrective-broadcast test. Not a blocker at this severity -- fine to land in this PR or as a fast-follow.

src/xagent/web/api/trace_handlers.py
```diff
@@ -105,119 +151,260 @@ async def _save_to_database(self, event: CoreTraceEvent) -> None:
...
+                )
+            return lease.run_id
 
         task_run = db.query(Task.run_id).filter(Task.id == self.task_id).one_or_none()
-        if task_run is None or task_run[0] is not None:
-            return False, None
+        if task_run is None:
+            # The task row itself is gone -- an exceptional condition, not
+            # a partition policy decision.
+            raise CheckpointUnavailableError(
```
rogercloud, line 403: **NIT:** this is the only one of five `CheckpointUnavailableError` raise sites in the file that doesn't call `register_degradation`. Defensible -- a missing task row isn't infra degradation -- but silent; a one-line comment would save a future editor the analysis of why this one is intentionally different.

rogercloud (approved):
## Update summary

Three new commits landed since the last review (`9310de57`, `1dcf053e`, `d3aedcd0`). The blocking MEDIUM bug from the previous round -- a snapshotless newest checkpoint row aborting the paged scan and shadowing an older valid checkpoint -- is fixed, covered by a new test that asserts the older snapshot's actual content is returned, and mutation-verified (reverting the fix makes that exact test fail with the original CheckpointCorruptError). Two non-blocking observations (missing log on a lost lease-restore fence; missing end-to-end coverage of the A2A preacquired_prior_status handover) were also addressed, the latter likewise mutation-verified. A fresh independent pass over the current head surfaced two further findings: one is a cleaner design refinement of an item already explicitly deferred, and one is a newly discovered narrow-window correctness bug (stale error_message surviving a successful restore). Neither is a blocker.

## Design verdict: Sound

The taxonomy stays minimal (three subclasses plus one reason discriminator), and the "None means only authoritative absence" invariant is now enforced consistently at every boundary with the snapshotless-shadowing bug fixed. The paged scan's unavailable-vs-corrupt decision rule -- a generic failure anywhere means retryable, and only a fully exhausted set of permanent failures means corrupt -- is appropriately conservative. Restoring through the same exact-lease WHERE fence the TTL reaper uses is the right mechanism: the loser of a race affects zero rows rather than double-releasing.

## Status of prior findings

1. **`src/xagent/web/api/trace_handlers.py` -- MEDIUM, snapshotless newest row shadowed an older valid checkpoint. FIXED** (`9310de57`). The snapshotless branch now sets `saw_undecodable_row = True`, logs a warning, and continues -- the same shape as the sibling decode-failure branch. The corrupt verdict is raised only after the entire matching set is exhausted across all pages with nothing readable found. New test `test_load_checkpoint_snapshotless_newest_row_does_not_shadow_older_valid` reproduces the two-row scenario in real DB order and asserts the older snapshot's actual content is returned. Mutation-tested: reverting the fix fails that test with the original error. 46 tests pass in the file.

2. **`src/xagent/web/api/websocket.py` -- LOW, deferred from an earlier round -- RUNNING-prior-status task settles FAILED on a transient checkpoint-read failure. Still deferred**, per the prior scope decision. See the new finding below, which reframes the same gap with a cleaner suggested fix.

3. **No log when the lease-restore fence loses the race. FIXED** (`1dcf053e`). An else branch now warns with task id, intended target status, and the fenced runner_id/run_id when the restore affects zero rows. Behavior otherwise unchanged.

4. **`preacquired_prior_status` never exercised end to end. FIXED** (`d3aedcd0`). The new test drives the real `_resume_input_required_a2a_task` entry point through the actual prelease and `_schedule_waiting_a2a_resume` -> `execute_resume_background` path, awaits the real background task, and asserts the row lands back on WAITING_FOR_USER with lease/error state cleared. Mutation-verified: dropping the kwarg makes the task land on FAILED instead. 123 relevant tests pass.

## New findings

See inline comments on `src/xagent/web/api/websocket.py:3524` and `:2818`.

## Additional observations (non-blocking, FYI)

- `trace_handlers.py`: `clear_degradation(CHECKPOINT_LOAD_UNAVAILABLE)` fires after every successful page query and is keyed process-globally by signal name, so one task's healthy read can clear a signal another task's concurrent failure just registered. Consistent with the pre-existing CHECKPOINT_DECODE_FALLBACK pattern; low priority.
- `trace_handlers.py`: see inline nit on the "task row is missing" raise -- the only one of five unavailable raise sites that skips register_degradation. Defensible, but a one-line comment would help a future editor.
- `websocket.py`: the corrective restore broadcast uses only the generic default message rather than the pending question text the historical-replay path prefers when available. Low impact.
- `websocket.py`: the out-parameter list-box pattern for `prior_status_out` is safe today but a small frozen dataclass return would remove the "which append won" hazard for future editors. Style nit.
- `src/xagent/core/agent/runner.py`: `_load_latest_checkpoint`'s `payload if isinstance(payload, dict) else None` still disagrees with `TraceCheckpointStore._unwrap_checkpoint_payload`, which raises CheckpointCorruptError for a non-dict payload. Confirmed unreachable today -- informational only.
- mypy could not be run in this pass (local numpy stub incompatibility); ruff format/check are clean on all changed files.
- Categories checked with no issues found: resource lifecycle, concurrency/fencing, unrelated-path breakage, the Tracer first-reader-wins change, and the shared `_waiting_or_paused_event_fields` refactor.

## Test results

Full changed-test-file set on Python 3.12 at the current head: **296 passed, 0 failed**. Re-confirmed across 9 consecutive runs with no flakiness.

## Overall recommendation

The design is sound and the blocking bug from the previous round is properly fixed and mutation-verified. Approving. The two new findings below are worth a look but neither is a blocker: the RUNNING-prior-status item is a refinement of an already-deferred follow-up, and the stale error_message issue is a real but narrow, non-security, non-data-loss UX bug that's cheap to fix -- happy to see it land in this PR or as a fast-follow, your call.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed in b2e4bce4: `release_task_lease_no_commit` now clears `error_message` when the release target is `PAUSED`/`WAITING_FOR_USER`, matching `recover_expired_task_lease_no_commit`, which already settles an expired run to PAUSED with a null error. `output` is deliberately left alone — only terminal transitions own it, and the channel and `finish_turn` settlement paths release to a resting state without rewriting it, so clearing there would drop the row's last readable answer. The corrective-broadcast test now seeds stale values and asserts the clear plus output preservation; an over-broad variant that also clears output goes red on those assertions.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Added in 02c2747f, with the reason at the raise site: the signal is process-wide and is only cleared once a page query succeeds later in the read, which this branch never reaches — registering there would latch it for every later reader. The existing missing-row test now also asserts the signal stays inactive, so the invariant is pinned rather than prose-only.

src/xagent/web/api/websocket.py
AlexLiu190625: Agreed — the `defer_db_cleanup_to_ttl_recovery` shape supersedes the release-to-PAUSED sketch in the tracked follow-up: retaining the exact lease for TTL recovery reuses a path already in this function and avoids new state transitions. Recorded in the follow-up; no code change in this PR.

PR #1119, opened 2026-08-04
fix(deployment): fail closed without public targets

bsbds:
## Summary

- remove the browser-origin fallback from Agent and Workforce deployment dialogs
- keep API, SDK, share-link, and widget copy controls disabled when public deployment configuration is unavailable
- show a persistent retry action and restore the controls after configuration loads successfully
- remove the obsolete browser fallback helper and rename the shared alert to describe the error state

## Why

Regional deployments advertise a direct regional origin through `/api/deployment-config`. If that request failed, the dialogs previously substituted the browser origin. For SaaS, that browser origin is the canonical routing edge, which cannot determine an unauthenticated SDK or widget caller's region and returns `409 region_required`.

Failing closed prevents users from copying a target that is known to be unsafe. Standalone and regional deployments behave as before whenever deployment configuration loads successfully.

This is the upstream frontend counterpart to the regional deployment work in xorbitsai/xagent-saas#322.

## Verification

- 30 focused deployment configuration and dialog tests passed
- frontend TypeScript type-check passed
- focused frontend ESLint checks passed
- Ruff, formatting, mypy, whitespace, import sorting, and codespell hooks passed

The repository-wide Alembic pre-commit hook could not bootstrap an unrelated `datrie` build in the local environment. The npm pre-commit hooks were run directly because npm was not present in that isolated hook environment.

### DIFF
--- a/frontend/src/components/build/deploy-agent-dialog.test.tsx
+++ b/frontend/src/components/build/deploy-agent-dialog.test.tsx
@@ -225,6 +225,35 @@ describe("DeployAgentDialog regional targets", () => {
     })
   })
 
+  it("keeps widget copy disabled when deployment configuration fails", async () => {
+    deploymentConfigFails = true
+
+    render(
+      <DeployAgentDialog
+        deployAgent={AGENT}
+        onClose={vi.fn()}
+        onUpdate={vi.fn()}
+      />,
+    )
+
+    fireEvent.click(screen.getByText("deploy_agent.options.embed.title"))
+
+    expect(
+      await screen.findByText("deployment_config.messages.load_failed"),
+    ).toBeInTheDocument()
+    const copyButton = await screen.findByTitle(
+      "deploy_agent.embed_snippet.copy_btn",
+    )
+    await waitFor(() => {
+      expect(apiRequestMock).toHaveBeenCalledWith(
+        "https://configured-api.example.test/api/agents/7/widget-key",
+      )
+    })
+    expect(screen.getByText("…")).toBeInTheDocument()
+    expect(copyButton).toBeDisabled()
+    expect(copyToClipboardMock).not.toHaveBeenCalled()
+  })
+
   it("bootstraps the region for public share links", async () => {
     render(
       <DeployAgentDialog
--- a/frontend/src/components/build/deploy-agent-dialog.test.tsx
+++ b/frontend/src/components/build/deploy-agent-dialog.test.tsx
@@ -155,6 +155,29 @@ describe("DeployAgentDialog regional targets", () => {
     expect(copyButton).toBeEnabled()
   })
 
+  it("keeps the share panel available when deployment configuration fails", async () => {
+    deploymentConfigFails = true
+
+    render(
+      <DeployAgentDialog
+        deployAgent={AGENT}
+        onClose={vi.fn()}
+        onUpdate={vi.fn()}
+      />,
+    )
+
+    fireEvent.click(screen.getByText("deploy_agent.options.shareable_link.title"))
+
+    expect(
+      await screen.findByText("deployment_config.messages.load_failed"),
+    ).toBeInTheDocument()
+    expect(await screen.findByRole("textbox")).toHaveValue("")
+    expect(screen.getByRole("button", { name: "common.copy" })).toBeDisabled()
+    expect(
+      screen.queryByText("deploy_agent.messages.share_failed"),
+    ).not.toBeInTheDocument()
+  })
+
   it("uses the deployment origin for API and SDK snippets", async () => {
     render(
       <DeployAgentDialog
--- a/frontend/src/components/build/deploy-agent-dialog.test.tsx
+++ b/frontend/src/components/build/deploy-agent-dialog.test.tsx
@@ -111,7 +111,7 @@ describe("DeployAgentDialog regional targets", () => {
     cleanup()
   })
 
-  it("falls back to the browser API target after config loading fails", async () => {
+  it("keeps API copy disabled until config retry succeeds", async () => {
     deploymentConfigFails = true
 
     render(
@@ -124,11 +124,14 @@ describe("DeployAgentDialog regional targets", () => {
 
     fireEvent.click(screen.getByText("deploy_agent.options.rest_api.title"))
 
+    expect(await screen.findByText("deployment_config.messages.load_failed")).toBeInTheDocument()
+    const copyButton = screen.getByTitle("deploy_agent.api_panel.copy_btn")
+    expect(copyButton).toBeDisabled()
     expect(
-      await screen.findByText((content) =>
+      screen.queryByText((content) =>
         content.includes("https://cloud.example.test/v1/chat/tasks"),
       ),
-    ).toBeInTheDocument()
+    ).not.toBeInTheDocument()
     expect(toastErrorMock).toHaveBeenCalledWith(
       "deployment_config.messages.load_failed",
     )
@@ -149,6 +152,7 @@ describe("DeployAgentDialog regional targets", () => {
     expect(
       screen.queryByText("deployment_config.messages.load_failed"),
     ).not.toBeInTheDocument()
+    expect(copyButton).toBeEnabled()
   })
 
   it("uses the deployment origin for API and SDK snippets", async () => {
--- a/frontend/src/components/build/deploy-agent-dialog.tsx
+++ b/frontend/src/components/build/deploy-agent-dialog.tsx
@@ -1,7 +1,7 @@
 "use client"
 
 import React, { useEffect, useMemo, useState } from "react"
-import { DeploymentConfigFallbackAlert } from "@/components/deployment/deployment-config-fallback-alert"
+import { DeploymentConfigErrorAlert } from "@/components/deployment/deployment-config-error-alert"
 import { Button } from "@/components/ui/button"
 import { Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle } from "@/components/ui/dialog"
 import { Card, CardHeader, CardTitle, CardDescription, CardContent } from "@/components/ui/card"
@@ -20,7 +20,6 @@ import { formatAgentApiSnippets, type ApiSnippetTab } from "@/lib/api-snippet-fo
 import type { ApiSnippetTarget } from "@/lib/api-snippet-target"
 import { getBrowserLocationOrigin } from "@/lib/browser-location"
 import {
-  browserDeploymentConfig,
   buildDeploymentShareUrl,
   fetchDeploymentConfig,
   resolveDeploymentOrigin,
@@ -113,12 +112,12 @@ export function DeployAgentDialog({ deployAgent, onClose, onUpdate, onManageApiK
       .catch((error) => {
         if (cancelled) return
         console.error("Failed to load deployment configuration", error)
-        setDeploymentConfig(browserDeploymentConfig())
-        setApiSnippetTarget(getApiSnippetTarget(browserOrigin))
+        setDeploymentConfig(null)
+        setApiSnippetTarget({ baseUrl: "" })
         setDeploymentConfigFailed(true)
         toast.error(
           t("deployment_config.messages.load_failed")
-          || "Failed to load deployment configuration; using this browser's origin.",
+          || "Failed to load deployment configuration. Retry before copying deployment details.",
         )
       })
 
@@ -137,7 +136,7 @@ export function DeployAgentDialog({ deployAgent, onClose, onUpdate, onManageApiK
       console.error("Failed to load deployment configuration", error)
       toast.error(
         t("deployment_config.messages.load_failed")
-        || "Failed to load deployment configuration; using this browser's origin.",
+        || "Failed to load deployment configuration. Retry before copying deployment details.",
       )
     }
   }
@@ -235,6 +234,7 @@ export function DeployAgentDialog({ deployAgent, onClose, onUpdate, onManageApiK
   }
 
   const handleCopyApiSnippet = async () => {
+    if (!apiSnippetTarget.baseUrl) return
     if (await copyToClipboard(apiSnippets[apiTab])) {
       setCopiedSnippet(true)
       toast.success(t("deploy_agent.messages.copied") || "Copied to clipboard")
@@ -479,7 +479,7 @@ export function DeployAgentDialog({ deployAgent, onClose, onUpdate, onManageApiK
         </DialogHeader>
 
         {deploymentConfigFailed && (
-          <DeploymentConfigFallbackAlert onRetry={retryDeploymentConfig} />
+          <DeploymentConfigErrorAlert onRetry={retryDeploymentConfig} />
         )}
 
         {activeView === "options" ? (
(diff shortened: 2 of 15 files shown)


frontend/src/i18n/locales/en.ts
```diff
@@ -3359,7 +3359,7 @@ Build when you need.`,
       retry: "Retry"
     },
     messages: {
-      load_failed: "Failed to load deployment configuration; using this browser's origin."
+      load_failed: "Failed to load deployment configuration. Retry before copying deployment details."
```
rogercloud, line 3362: **Medium — the Chinese locale wasn't updated to match.** This English fallback was updated to describe the new fail-closed behavior ("Failed to load deployment configuration. Retry before copying deployment details."), but `frontend/src/i18n/locales/zh.ts:3362` (the only other locale file) still has the old text, which says the deployment config failed to load and that the app *has switched to the current browser address* — i.e. it describes precisely the browser-origin fallback this PR removes. A user on the Chinese locale would see a message claiming a fallback happened, next to copy buttons that are disabled and do nothing. Please update `zh.ts` to match.

frontend/src/components/build/deploy-agent-dialog.tsx
```diff
@@ -691,7 +692,7 @@ export function DeployAgentDialog({ deployAgent, onClose, onUpdate, onManageApiK
                     <Label className="text-sm">{t("deploy_agent.share_link.public_url") || "Public URL"}</Label>
                     <div className="flex gap-2">
                       <Input readOnly value={shareUrl} className="flex-1" />
-                      <Button variant="secondary" onClick={() => void handleCopyShareLink()} disabled={isUpdatingShare}>
+                      <Button variant="secondary" onClick={() => void handleCopyShareLink()} disabled={isUpdatingShare || !shareUrl}>
```
rogercloud, line 696: **Medium — this panel can misreport a config failure as a share-link failure, and the new guard here is unreachable.** `shareUrl` comes from `buildDeploymentShareUrl(..., deploymentConfig, ...)`, which returns `""` when `deploymentConfig` is `null`. So on a config load failure, `setDeploymentConfig(null)` forces `shareUrl` to `""`; because the render condition a few lines up is still `shareEnabled && shareUrl ? <input panel> : shareEnabled ? <share_failed> : ...`, the whole panel falls into the generic-failure branch and renders `deploy_agent.messages.share_failed` ("Share link action failed") — attributing a deployment-config problem to the share-link action, and this `Button` (along with its new `|| !shareUrl` guard) never even renders in that case, making the added guard unreachable dead code. This PR relaxed the same guard in the sibling `frontend/src/components/workforce/workforce-share-dialog.tsx:183` (from `shareEnabled && shareUrl ?` to `shareEnabled ?`), so that dialog correctly shows an empty input plus the persistent config-error alert. Suggest applying the same relaxation here (dropping the now-redundant `&& shareUrl` on the outer condition) so both dialogs behave identically and this `!shareUrl` guard starts doing real work.

frontend/src/components/build/deploy-agent-dialog.test.tsx
```diff
@@ -111,7 +111,7 @@ describe("DeployAgentDialog regional targets", () => {
     cleanup()
   })
 
-  it("falls back to the browser API target after config loading fails", async () => {
+  it("keeps API copy disabled until config retry succeeds", async () => {
```
rogercloud, line 114: **Medium — failure-path coverage only covers the REST-API tab.** This is the only one of the five `it()` blocks in this file that sets `deploymentConfigFails = true`, and it exercises the REST-API tab alone. Neither the share-link tab nor the embed/widget tab is opened under a failing config, even though `shareUrl` and `widgetOrigin` (via `resolveDeploymentOrigin(deploymentConfig, appOrigin)`) are both directly driven by `deploymentConfig` and change visibly under failure. A share-tab case would have caught the panel falling through to `deploy_agent.messages.share_failed` on config failure. Suggest adding two cases asserting the share input and the widget snippet/copy controls under `deploymentConfigFails = true`.

frontend/src/components/deployment/deployment-config-error-alert.tsx
```diff
@@ -17,15 +17,16 @@ interface DeploymentConfigFallbackAlertProps {
 }
 
 /**
- * Persistent warning for deployment dialogs using the browser-origin fallback.
+ * Persistent warning for deployment dialogs whose public target is unavailable.
  *
  * A toast is easy to miss and can leave a user copying a target that is wrong
```
rogercloud: **Low — this doc comment describes the pre-PR semantics.** It says a missed toast "can leave a user copying a target that is wrong for a regional deployment." After this PR nothing wrong is copyable at all: callers hard-disable copy controls on failure. Reword to describe an *unavailable* target rather than a *wrong* one. Optional, purely stylistic: now that this is a hard copy-block rather than a soft warning, the custom amber/`AlertTriangle` styling below could arguably use the `Alert` component's `destructive` variant.

frontend/src/components/build/deploy-agent-dialog.test.tsx
```diff
@@ -149,6 +152,7 @@ describe("DeployAgentDialog regional targets", () => {
     expect(
       screen.queryByText("deployment_config.messages.load_failed"),
     ).not.toBeInTheDocument()
+    expect(copyButton).toBeEnabled()
```
rogercloud, line 155: **Low — the retry-failure path is untested across all four surfaces.** This test (and the equivalent one in `deploy-workforce-dialog.test.tsx`, `workforce-share-dialog.test.tsx`, and `workforce-widget-dialog.test.tsx`) only asserts `copyButton` becomes enabled after retry *succeeds*. Nothing exercises retry while the underlying fetch keeps rejecting, so the `finally { setIsRetrying(false) }` branch in `deployment-config-error-alert.tsx` is never covered for the rejecting case, and there is no assertion that controls *stay* disabled and the alert *stays* mounted after a second failed retry. One test — ideally on the shared alert component itself — would close the gap for all four surfaces.

frontend/src/components/workforce/deploy-workforce-dialog.tsx
```diff
@@ -226,13 +226,14 @@ export function DeployWorkforceDialog({
             </div>
             <div className="mt-3 bg-muted p-4 rounded-md text-xs font-mono relative group">
               <pre className="whitespace-pre-wrap break-all text-muted-foreground max-h-72 overflow-auto">
-                {snippets[apiTab]}
+                {apiTarget.baseUrl ? snippets[apiTab] : "…"}
```
rogercloud, line 232: **Low — loading and permanently-failed render identically.** `apiTarget.baseUrl` is `""` both while the config request is in flight and after it has permanently failed, so `apiTarget.baseUrl ? snippets[apiTab] : "…"` shows the same placeholder in both states and the snippet area alone can't distinguish them. The separate error alert does explain the failure elsewhere on screen, so this is minor. Secondary, optional: `api-snippet-format.ts` already substitutes a `YOUR_API_BASE_URL` placeholder for an empty `baseUrl`, and `snippets` is computed unconditionally before this ternary — so the ternary discards an already-computed, more informative snippet in favor of "…". Keeping the current behavior is defensible (a placeholder-URL snippet before config resolves could itself confuse); flagging in case the ambiguity was unintended. Same pattern at `frontend/src/components/build/deploy-agent-dialog.tsx:540`.

rogercloud (approved):
## Summary

Four deployment surfaces (the Agent dialog's REST-API panel, share link, and widget tabs, plus the Workforce REST-API panel, share dialog, and widget dialog) previously substituted the browser's own origin whenever `GET /api/deployment-config` failed. On multi-region SaaS the browser origin is the routing edge, which answers `409 region_required` for unauthenticated SDK/widget callers — so the user could copy a target that was already known to be broken. This PR makes those surfaces fail closed instead: no origin is guessed, snippets and URLs are blanked, copy controls are disabled, and a persistent retry alert (`DeploymentConfigFallbackAlert` → `DeploymentConfigErrorAlert`) stays mounted until config loads. The now-unused `browserDeploymentConfig()` helper is deleted. 12 files, +70/-83, frontend only.

## Approach verdict: sound

Failing closed is the right call here, and it aligns the failure path with what `resolveDeploymentOrigin` and `buildDeploymentShareUrl` already do — both return `""` for `config === null`, so this PR mostly stops feeding them a fabricated config. The change is small and additive. My reservations are about consistency across the four surfaces, not about direction.

Two design-level notes that no single line owns:

- **The four dialogs duplicate the same ~30 lines.** `deploy-agent-dialog.tsx`, `deploy-workforce-dialog.tsx`, `workforce-share-dialog.tsx`, and `workforce-widget-dialog.tsx` each carry their own `deploymentConfigFailed` state, an identical `.catch` block, an identical `retryDeploymentConfig`, and the same toast message (the literal fallback string appears 8 times). A shared `useDeploymentConfig()` hook returning `{ config, target, failed, retry }` would delete most of this and structurally prevent the surfaces from drifting. This PR is itself evidence that they already drift — see the finding below where the agent share panel was left inconsistent with the workforce share dialog.
- **Two different "unavailable" representations coexist.** The agent/share/widget dialogs model it as `DeploymentConfig | null`; the two API panels model it as `ApiSnippetTarget { baseUrl: "" }`. That produces four different ad-hoc gate expressions (`!apiSnippetTarget.baseUrl`, `!apiTarget.baseUrl`, `!shareUrl`, `!widgetKey || !widgetOrigin`), each surface re-deriving "do I have a public target?" independently. Correct today, fragile over time — a candidate for later consolidation, not blocking.

## Findings

### Medium

See inline comments on `frontend/src/i18n/locales/en.ts:3362` (the Chinese locale wasn't updated to match), `frontend/src/components/build/deploy-agent-dialog.tsx:695` (share panel misreports a config failure as a share-link failure, plus an unreachable guard), and `frontend/src/components/build/deploy-agent-dialog.test.tsx:114` (failure-path coverage only covers the REST-API tab).

### Low

See inline comments on `frontend/src/components/deployment/deployment-config-error-alert.tsx:22` (stale doc comment), `frontend/src/components/build/deploy-agent-dialog.test.tsx:155` (retry-failure path untested across all four surfaces), and `frontend/src/components/workforce/deploy-workforce-dialog.tsx:229` (loading vs. permanently-failed render identically; same pattern at `deploy-agent-dialog.tsx:540`).

`shrink: L120,139 (frontend/src/components/build/deploy-agent-dialog.tsx), L88,105 (frontend/src/components/workforce/deploy-workforce-dialog.tsx), L81,113 (frontend/src/components/workforce/workforce-share-dialog.tsx), L81,113 (frontend/src/components/workforce/workforce-widget-dialog.tsx): the literal fallback string "Failed to load deployment configuration. Retry before copying deployment details." for the deployment_config.messages.load_failed key is copy-pasted 8 times across 4 files (byte-identical). Extract a single exported constant (e.g. colocated in frontend/src/lib/deployment-config.ts) and import it in all 4 dialogs.` This is a deviation from the repo's usual `t(key) || "fallback"` convention only in that the *same* literal for the *same* key is duplicated across independent call sites, which the convention doesn't otherwise produce; 8 sync points collapse to 1 at the cost of one import per file.

net: -6 lines possible (8 duplicated literals collapse to 1 constant + 4 imports)

## Testing limitation

I could not run the test suite: `frontend/node_modules` is not installed in the review checkout, so vitest, tsc, and eslint were not executed. Every finding above comes from static reading — tracing `deploymentConfig`, `apiSnippetTarget`, `shareUrl`, and `widgetOrigin` through the current source and diffing against the base commit to separate new behavior from pre-existing behavior. This is a real coverage limitation of this review, not a claim that the tests pass or fail.

## Disposition

Approving — the design is sound and nothing here blocks merge. Two items worth a quick follow-up: the stale Chinese translation (it tells the user a fallback happened when it no longer does) and the agent share panel showing "Share link action failed" for what is actually a config failure — the sibling workforce dialog already got the correct treatment in this same PR.

frontend/src/components/build/deploy-agent-dialog.tsx
```diff
@@ -113,12 +112,12 @@ export function DeployAgentDialog({ deployAgent, onClose, onUpdate, onManageApiK
...
-        setApiSnippetTarget(getApiSnippetTarget(browserOrigin))
+        setDeploymentConfig(null)
+        setApiSnippetTarget({ baseUrl: "" })
         setDeploymentConfigFailed(true)
         toast.error(
           t("deployment_config.messages.load_failed")
-          || "Failed to load deployment configuration; using this browser's origin.",
+          || "Failed to load deployment configuration. Retry before copying deployment details.",
         )
       })
```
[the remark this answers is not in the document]
bsbds: Not changing this in this PR. The reset effect predates this change, and the parent always sets `deployAgent` to `null` on close; the effect clears API-backed state while the dialog remains mounted during its close transition. Keying the component would change that lifecycle and animation behavior without a demonstrated stale-state bug, so this remains an unresolved follow-up design suggestion.

frontend/src/components/workforce/deploy-workforce-dialog.tsx
bsbds: Keeping the current behavior. The persistent error alert distinguishes a failed request from loading, while the ellipsis deliberately avoids presenting a placeholder endpoint or snippet before a verified target exists. Copy controls remain disabled, so this does not expose an incorrect target.

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/deployment-config-fail-closed`
- **Head:** `b040b9ddfcc311cf1b60d076a4ac23d6d4b35bed`
- **Destination:** `bsbds/xagent:fix/deployment-config-fail-closed`
- **Result:** success

### Resolved

1. **Thread:** `PRRT_kwDORQ4Kr86WPEU7` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3710441403
   **Local outcome:** resolved
   **Commit:** `b8339e2a5f2badb86750624bb5224c5a961112b3`
   **GitHub:** resolved
2. **Thread:** `PRRT_kwDORQ4Kr86WPEU_` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3710441409
   **Local outcome:** resolved
   **Commit:** `0e4691c71bda148c1de5b4a466bf56ff2a355bb8`
   **GitHub:** resolved
3. **Thread:** `PRRT_kwDORQ4Kr86WPEVF` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3710441415
   **Local outcome:** resolved
   **Commit:** `0e4691c71bda148c1de5b4a466bf56ff2a355bb8`, `9aff137d63ef91f60731a485479444ebfe007bd7`
   **GitHub:** resolved
4. **Thread:** `PRRT_kwDORQ4Kr86WPEVI` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3710441420
   **Local outcome:** resolved
   **Commit:** `8d6b8ffbe9eb56e2e361a373c210147b03321715`
   **GitHub:** resolved
5. **Thread:** `PRRT_kwDORQ4Kr86WPEVP` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3710441426
   **Local outcome:** resolved
   **Commit:** `8c4114512fcdbea335508383fdccc0b6abb43058`
   **GitHub:** resolved

### Rejected

1. **Thread:** `PRRT_kwDORQ4Kr86WLzdT` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3709194054
   **Local outcome:** rejected
   **Reason:** The reset effect predates this change and clears API-backed state while the dialog remains mounted during its close transition. Keying the component would change lifecycle and animation behavior without a demonstrated stale-state defect.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3717770118
   **GitHub:** left unresolved
2. **Thread:** `PRRT_kwDORQ4Kr86WPEVT` — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3710441432
   **Local outcome:** rejected
   **Reason:** The persistent error alert distinguishes failure from loading, while the ellipsis avoids presenting a placeholder endpoint before a verified target exists. Copy remains disabled.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1119#discussion_r3717770168
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 5
- **Rejected:** 2
- **Unresolved:** 0

### Review-body notes

- Accepted the repeated-fallback concern and centralized the copy in `b040b9ddfcc311cf1b60d076a4ac23d6d4b35bed`.
- Did not introduce a generic deployment-config hook. The consumers have different open, loading, and combined-request lifecycles, so a shared hook would add flags and callbacks without improving this fix.
- Did not unify `DeploymentConfig | null` with `ApiSnippetTarget { baseUrl: "" }`. They represent different derivation stages, and unification would be a broader state-model refactor with no required behavior change.
- The Gemini review summary and Roger findings roll-up duplicate the corresponding inline items.
- Verification: 42 focused frontend tests, frontend lint, frontend type-check, codespell, and `git diff --check` passed. Full local pre-commit bootstrap could not complete in this environment; the GitHub checks are running.

PR #1120, opened 2026-08-04
fix: retry Auto model when reasoning is mandatory

qinxuye:
## Summary

- detect the untyped OpenRouter 400 returned when an Auto-selected endpoint requires reasoning
- retry the same selected model once with thinking enabled instead of rerouting or failing the task
- cover both streaming and non-streaming calls, including protection against repeating the same retry action

## Root cause

Auto routing uses a strict tool-call request with thinking explicitly disabled. Some downstream endpoints require reasoning and reject that request with `Reasoning is mandatory for this endpoint and cannot be disabled.` Because the provider exposes this only as an untyped 400, the existing compatibility retry layer did not recognize it and the task failed immediately.

## Impact

Auto model requests can continue when the selected downstream endpoint mandates reasoning. Other provider errors and retry behavior remain unchanged.

## Validation

- `PYTHONPATH=src pytest -q tests/core/model/chat/basic/test_router.py` — 22 passed
- `ruff format --check src/xagent/core/model/chat/basic/router.py tests/core/model/chat/basic/test_router.py`
- `ruff check src/xagent/core/model/chat/basic/router.py tests/core/model/chat/basic/test_router.py`
- commit hooks passed: Ruff, formatting, mypy, isort, codespell, whitespace checks

### DIFF
--- a/src/xagent/core/model/chat/basic/router.py
+++ b/src/xagent/core/model/chat/basic/router.py
@@ -46,6 +46,7 @@ logger = logging.getLogger(__name__)
 _DEFAULT_ROUTER_ABILITIES = ["chat", "tool_calling"]
 _UNROUTED_ROUTER_ABILITIES = {"vision", "thinking_mode"}
 _DISABLE_DOWNSTREAM_THINKING = {"type": "disabled", "enable": False}
+_ENABLE_DOWNSTREAM_THINKING = {"type": "enabled", "enable": True}
 _CONTENT_PART_MODALITIES = {
     "audio": "audio",
     "audio_url": "audio",
@@ -69,6 +70,23 @@ _MODALITY_ABILITIES = {
 
 class RouterModalityRoutingError(RuntimeError):
     """The installed router cannot enforce required input modalities."""
+
+
+def _should_retry_with_thinking(
+    exc: Exception,
+    *,
+    thinking: dict[str, Any] | None,
+) -> bool:
+    if thinking is None or not (
+        thinking.get("type") == "disabled" or thinking.get("enable") is False
+    ):
+        return False
+
+    # OpenRouter currently exposes this provider constraint only through an
+    # untyped 400 response. Retry the same selected model once with reasoning
+    # enabled instead of repeating the rejected payload or rerouting.
+    exc_msg = str(exc).lower()
+    return "reasoning is mandatory" in exc_msg and "cannot be disabled" in exc_msg
 
 
 def _should_retry_without_thinking(
@@ -114,6 +132,14 @@ def _next_retry_state(
     thinking: dict[str, Any] | None,
     tool_choice: str | dict[str, Any] | None,
 ) -> tuple[str | dict[str, Any] | None, dict[str, Any] | None, str, str] | None:
+    if _should_retry_with_thinking(exc, thinking=thinking):
+        return (
+            tool_choice,
+            _ENABLE_DOWNSTREAM_THINKING,
+            "selected model requires reasoning; retrying with thinking enabled",
+            "enable_thinking",
+        )
+
     if _should_retry_without_thinking(exc, thinking=thinking, tool_choice=tool_choice):
         return (
             tool_choice,
--- a/src/xagent/core/model/chat/basic/router.py
+++ b/src/xagent/core/model/chat/basic/router.py
@@ -77,6 +77,8 @@ def _should_retry_with_thinking(
     *,
     thinking: dict[str, Any] | None,
 ) -> bool:
+    # This is the primary stop condition after a retry swaps in enabled thinking;
+    # retry-action tracking remains defense in depth for the shared retry loop.
     if isinstance(thinking, dict) and (
         thinking.get("type") == "enabled" or thinking.get("enable") is True
     ):
@@ -85,6 +87,7 @@ def _should_retry_with_thinking(
     # OpenRouter currently exposes this provider constraint only through an
     # untyped 400 response. Retry the same selected model once with reasoning
     # enabled instead of repeating the rejected payload or rerouting.
+    # Replace string matching with typed provider errors when available.
     exc_msg = str(exc).lower()
     return "reasoning is mandatory" in exc_msg and "cannot be disabled" in exc_msg
 
--- a/src/xagent/core/model/chat/basic/router.py
+++ b/src/xagent/core/model/chat/basic/router.py
@@ -45,6 +45,7 @@ logger = logging.getLogger(__name__)
 
 _DEFAULT_ROUTER_ABILITIES = ["chat", "tool_calling"]
 _UNROUTED_ROUTER_ABILITIES = {"vision", "thinking_mode"}
+# Normalized intents translated into OpenRouter's reasoning/thinking request body.
 _DISABLE_DOWNSTREAM_THINKING = {"type": "disabled", "enable": False}
 _ENABLE_DOWNSTREAM_THINKING = {"type": "enabled", "enable": True}
 _CONTENT_PART_MODALITIES = {
--- a/src/xagent/core/model/chat/basic/router.py
+++ b/src/xagent/core/model/chat/basic/router.py
@@ -77,8 +77,8 @@ def _should_retry_with_thinking(
     *,
     thinking: dict[str, Any] | None,
 ) -> bool:
-    if thinking is None or not (
-        thinking.get("type") == "disabled" or thinking.get("enable") is False
+    if isinstance(thinking, dict) and (
+        thinking.get("type") == "enabled" or thinking.get("enable") is True
     ):
         return False
 
--- a/tests/core/model/chat/basic/test_router.py
+++ b/tests/core/model/chat/basic/test_router.py
@@ -21,6 +21,10 @@ _OPENROUTER_TOOL_CHOICE_ERROR = (
 _THINKING_TOOL_CHOICE_ERROR = (
     "OpenAI bad request (400): Thinking mode does not support this tool_choice"
 )
+_MANDATORY_REASONING_ERROR = (
+    "OpenAI bad request (400): Reasoning is mandatory for this endpoint "
+    "and cannot be disabled."
+)
 
 
 def _tool_schema() -> dict[str, Any]:
@@ -127,6 +131,26 @@ class _ScriptedChatLLM(BaseLLM):
         if self.errors:
             raise RuntimeError(self.errors.pop(0))
         return "ok"
+
+    async def stream_chat(
+        self,
+        messages: list[dict[str, str]],
+        temperature: float | None = None,
+        max_tokens: int | None = None,
+        tools: list[dict[str, Any]] | None = None,
+        tool_choice: str | dict[str, Any] | None = None,
+        response_format: dict[str, Any] | None = None,
+        thinking: dict[str, Any] | None = None,
+        output_config: dict[str, Any] | None = None,
+        **kwargs: Any,
+    ) -> AsyncIterator[StreamChunk]:
+        del messages, temperature, max_tokens, tools, response_format
+        del output_config, kwargs
+        self.tool_choices.append(tool_choice)
+        self.thinking_values.append(thinking)
+        if self.errors:
+            raise RuntimeError(self.errors.pop(0))
+        yield StreamChunk(type=ChunkType.TOKEN, content="ok", delta="ok")
 
 
 async def _select_glm(_prompt: str) -> str:
@@ -587,6 +611,73 @@ async def test_router_chat_propagates_non_matching_errors_without_retry(monkeypa
         )
 
     assert llm.tool_choices == ["required"]
+
+
+@pytest.mark.asyncio
+async def test_router_chat_enables_thinking_when_selected_model_requires_it(
+    monkeypatch,
+):
+    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR])
+    router = RouterLLM(downstream_resolver=lambda _model_id: llm)
+    monkeypatch.setattr(router, "_select_model", _select_glm)
+
+    result = await router.chat(
+        [{"role": "user", "content": "score?"}],
+        tools=[_tool_schema()],
+        tool_choice="required",
+        thinking={"type": "disabled", "enable": False},
+    )
+
+    assert result == "ok"
+    assert llm.thinking_values == [
+        {"type": "disabled", "enable": False},
+        {"type": "enabled", "enable": True},
+    ]
+
+
+@pytest.mark.asyncio
+async def test_router_stream_enables_thinking_when_selected_model_requires_it(
+    monkeypatch,
+):
+    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR])
+    router = RouterLLM(downstream_resolver=lambda _model_id: llm)
+    monkeypatch.setattr(router, "_select_model", _select_glm)
+
+    chunks = [
+        chunk
+        async for chunk in router.stream_chat(
+            [{"role": "user", "content": "score?"}],
+            tools=[_tool_schema()],
+            tool_choice="required",
+            thinking={"type": "disabled", "enable": False},
+        )
+    ]
+
+    assert [chunk.delta for chunk in chunks] == ["ok"]
+    assert llm.thinking_values == [
+        {"type": "disabled", "enable": False},
+        {"type": "enabled", "enable": True},
+    ]
+
+
+@pytest.mark.asyncio
+async def test_router_does_not_repeat_mandatory_reasoning_retry(monkeypatch):
+    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR, _MANDATORY_REASONING_ERROR])
+    router = RouterLLM(downstream_resolver=lambda _model_id: llm)
+    monkeypatch.setattr(router, "_select_model", _select_glm)
+
+    with pytest.raises(RuntimeError, match="Reasoning is mandatory"):
+        await router.chat(
+            [{"role": "user", "content": "score?"}],
+            tools=[_tool_schema()],
+            tool_choice="required",
+            thinking={"type": "disabled", "enable": False},
+        )
+
+    assert llm.thinking_values == [
+        {"type": "disabled", "enable": False},
+        {"type": "enabled", "enable": True},
+    ]
 
 
 @pytest.mark.asyncio
--- a/tests/core/model/chat/basic/test_router.py
+++ b/tests/core/model/chat/basic/test_router.py
@@ -668,7 +668,13 @@ async def test_router_stream_enables_thinking_when_selected_model_requires_it(
 
 @pytest.mark.asyncio
 async def test_router_does_not_repeat_mandatory_reasoning_retry(monkeypatch):
-    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR, _MANDATORY_REASONING_ERROR])
+    llm = _ScriptedChatLLM(
+        [
+            _MANDATORY_REASONING_ERROR,
+            _THINKING_TOOL_CHOICE_ERROR,
+            _MANDATORY_REASONING_ERROR,
+        ]
+    )
     router = RouterLLM(downstream_resolver=lambda _model_id: llm)
     monkeypatch.setattr(router, "_select_model", _select_glm)
 
@@ -683,7 +689,9 @@ async def test_router_does_not_repeat_mandatory_reasoning_retry(monkeypatch):
     assert llm.thinking_values == [
         {"type": "disabled", "enable": False},
         {"type": "enabled", "enable": True},
+        {"type": "disabled", "enable": False},
     ]
+    assert llm.tool_choices == ["required", "required", "required"]
 
 
 @pytest.mark.asyncio
--- a/tests/core/model/chat/basic/test_router.py
+++ b/tests/core/model/chat/basic/test_router.py
@@ -613,9 +613,14 @@ async def test_router_chat_propagates_non_matching_errors_without_retry(monkeypa
     assert llm.tool_choices == ["required"]
 
 
+@pytest.mark.parametrize(
+    "thinking",
+    [None, {"type": "disabled", "enable": False}],
+    ids=["unspecified", "disabled"],
+)
 @pytest.mark.asyncio
 async def test_router_chat_enables_thinking_when_selected_model_requires_it(
-    monkeypatch,
+    monkeypatch, thinking
 ):
     llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR])
     router = RouterLLM(downstream_resolver=lambda _model_id: llm)
@@ -625,35 +630,15 @@ async def test_router_chat_enables_thinking_when_selected_model_requires_it(
         [{"role": "user", "content": "score?"}],
         tools=[_tool_schema()],
         tool_choice="required",
-        thinking={"type": "disabled", "enable": False},
+        thinking=thinking,
     )
 
     assert result == "ok"
     assert llm.thinking_values == [
-        {"type": "disabled", "enable": False},
+        thinking,
         {"type": "enabled", "enable": True},
     ]
-
-
-@pytest.mark.asyncio
-async def test_router_chat_enables_thinking_when_request_omits_preference(
-    monkeypatch,
-):
-    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR])
-    router = RouterLLM(downstream_resolver=lambda _model_id: llm)
-    monkeypatch.setattr(router, "_select_model", _select_glm)
-
-    result = await router.chat(
-        [{"role": "user", "content": "score?"}],
-        tools=[_tool_schema()],
-        tool_choice="required",
-    )
-
-    assert result == "ok"
-    assert llm.thinking_values == [
-        None,
-        {"type": "enabled", "enable": True},
-    ]
+    assert llm.tool_choices == ["required", "required"]
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


src/xagent/core/model/chat/basic/router.py
```diff
@@ -71,6 +72,23 @@ class RouterModalityRoutingError(RuntimeError):
...
 
+def _should_retry_with_thinking(
+    exc: Exception,
+    *,
+    thinking: dict[str, Any] | None,
+) -> bool:
+    if thinking is None or not (
+        thinking.get("type") == "disabled" or thinking.get("enable") is False
+    ):
+        return False
```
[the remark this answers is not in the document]
qinxuye: Fixed in f42107af. `_should_retry_with_thinking` now skips retry only when thinking is already explicitly enabled, so an omitted preference can retry once when the provider returns the mandatory-reasoning error. Added a regression test for `thinking=None`; the full router suite passes with 23 tests.

src/xagent/core/model/chat/basic/router.py
```diff
@@ -71,6 +72,23 @@ class RouterModalityRoutingError(RuntimeError):
     """The installed router cannot enforce required input modalities."""
 
 
+def _should_retry_with_thinking(
```
rogercloud, line 76: The "already enabled → don't retry" guard isn't covered as a first-call case. There's no test where the initial request already carries `thinking={"type": "enabled", "enable": True}`, hits the mandatory-reasoning error, and asserts it propagates immediately after exactly one downstream call. Right now the branch is only reached as the second attempt inside `test_router_does_not_repeat_mandatory_reasoning_retry`, so a regression that made this guard fall through unconditionally would still leave the suite green.

```python
@pytest.mark.asyncio
async def test_router_chat_does_not_retry_when_thinking_already_enabled(monkeypatch):
    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR])
    router = RouterLLM(downstream_resolver=lambda _model_id: llm)
    monkeypatch.setattr(router, "_select_model", _select_glm)

    with pytest.raises(RuntimeError, match="Reasoning is mandatory"):
        await router.chat(
            [{"role": "user", "content": "score?"}],
            tools=[_tool_schema()],
            tool_choice="required",
            thinking={"type": "enabled", "enable": True},
        )

    assert llm.thinking_values == [{"type": "enabled", "enable": True}]
```

src/xagent/core/model/chat/basic/router.py
```diff
@@ -71,6 +72,23 @@ class RouterModalityRoutingError(RuntimeError):
...
+    exc: Exception,
+    *,
+    thinking: dict[str, Any] | None,
+) -> bool:
+    if isinstance(thinking, dict) and (
```
rogercloud, line 83: Worth a clarifying comment here: for the `enable_thinking` action, this guard is the real stop condition for a repeat retry, not the `action_key in attempted_retry_actions` dedup check — once the retry substitutes `_ENABLE_DOWNSTREAM_THINKING`, this predicate returns `False` and the dedup check is never reached on that path. The dedup set is still load-bearing for the other two actions, so both are correct; the risk is just that a future reader assumes the set is doing the work here and "simplifies" the guard away. One line noting that would prevent it.

src/xagent/core/model/chat/basic/router.py
```diff
@@ -71,6 +72,23 @@ class RouterModalityRoutingError(RuntimeError):
...
+        thinking.get("type") == "enabled" or thinking.get("enable") is True
+    ):
+        return False
+
+    # OpenRouter currently exposes this provider constraint only through an
```
rogercloud, line 88: Minor consistency point: the two sibling predicates below (lines 98-101 and 120-123) both close with a note about replacing the string match with typed provider errors once the follow-up tracking issue lands. All three are intended to retire together, so adding the same sentence here keeps them discoverable as one group when that cleanup happens.

src/xagent/core/model/chat/basic/router.py
```diff
@@ -71,6 +72,23 @@ class RouterModalityRoutingError(RuntimeError):
     """The installed router cannot enforce required input modalities."""
 
 
+def _should_retry_with_thinking(
+    exc: Exception,
+    *,
+    thinking: dict[str, Any] | None,
+) -> bool:
+    if isinstance(thinking, dict) and (
+        thinking.get("type") == "enabled" or thinking.get("enable") is True
```
rogercloud, line 84: Nit, theoretical only: this uses `thinking.get("enable") is True`, whereas every downstream consumer reads the flag truthily (`openrouter.py:371`, `claude.py:603`/`:873`, `dashscope.py:102`, `deepseek.py:115`, `xinference.py:205-208`/`:425-428` all use `thinking.get("enable", False)`). No path in the repo constructs a non-bool truthy `enable` today, so they can't actually diverge — `bool(thinking.get("enable"))` would just make the agreement structural rather than incidental.

tests/core/model/chat/basic/test_router.py
```diff
@@ -589,6 +613,94 @@ async def test_router_chat_propagates_non_matching_errors_without_retry(monkeypa
...
+):
+    llm = _ScriptedChatLLM([_MANDATORY_REASONING_ERROR])
+    router = RouterLLM(downstream_resolver=lambda _model_id: llm)
+    monkeypatch.setattr(router, "_select_model", _select_glm)
+
+    result = await router.chat(
+        [{"role": "user", "content": "score?"}],
+        tools=[_tool_schema()],
+        tool_choice="required",
+        thinking={"type": "disabled", "enable": False},
```
rogercloud: Optional: nothing asserts that firing the reasoning retry leaves `tool_choice` untouched. `_next_retry_state`'s ordering structurally guarantees it and the two trigger strings are disjoint in practice, so this is low value — but `_ScriptedChatLLM.tool_choices` already captures it, so `assert llm.tool_choices == ["required", "required"]` here would be a nearly free extra invariant.

tests/core/model/chat/basic/test_router.py
```diff
@@ -589,6 +613,94 @@ async def test_router_chat_propagates_non_matching_errors_without_retry(monkeypa
...
+
+    assert result == "ok"
+    assert llm.thinking_values == [
+        {"type": "disabled", "enable": False},
+        {"type": "enabled", "enable": True},
+    ]
+
+
+@pytest.mark.asyncio
+async def test_router_chat_enables_thinking_when_request_omits_preference(
```
rogercloud: This test and `test_router_chat_enables_thinking_when_selected_model_requires_it` above differ only in whether the initial `thinking` is omitted or explicitly disabled, and both land on the same retry branch. `@pytest.mark.parametrize("thinking", [None, {"type": "disabled", "enable": False}])` over a single test would keep identical coverage with roughly half the lines.

tests/core/model/chat/basic/test_router.py
```diff
@@ -128,6 +132,26 @@ async def chat(
             raise RuntimeError(self.errors.pop(0))
         return "ok"
 
+    async def stream_chat(
```
rogercloud, line 135: `stream_chat` repeats `chat`'s (line 115) arg-discard, `tool_choices`/`thinking_values` bookkeeping, and error-raising verbatim — the only real difference is return vs yield. A small shared helper would collapse it. Fully optional: the pre-existing `_ToolChoiceRetryLLM` (line 41) carries the same pattern and isn't touched by this PR, so it's a cleanup for whenever, not a fix owed here.

rogercloud (commented):
## Summary

This PR teaches `RouterLLM` to recover from OpenRouter's untyped 400 `"Reasoning is mandatory for this endpoint and cannot be disabled"` rejection. A new `_should_retry_with_thinking` predicate (`src/xagent/core/model/chat/basic/router.py:75-89`) matches that error string and drives a single retry of the same selected model with `_ENABLE_DOWNSTREAM_THINKING` (`router.py:49`) substituted for the rejected thinking payload — no reroute, no task failure. It plugs into the existing first-match-wins ladder in `_next_retry_state` (`router.py:128-161`) ahead of the two pre-existing compatibility predicates, and is deduped through the same `attempted_retry_actions` bookkeeping used by the non-streaming and streaming retry loops (`router.py:290-339`, `router.py:417-472`). Tests in `tests/core/model/chat/basic/test_router.py` cover the chat path, the stream path, and the no-repeat case.

## Update since last review

The one open finding from the previous round — `_should_retry_with_thinking` returning `False` outright whenever `thinking is None`, which made a request with no explicit reasoning preference permanently ineligible for the retry — is fixed in `f42107af`. The guard now bails only when thinking is explicitly already enabled (`type == "enabled"` or `enable is True`), so `None` falls through to the string match. `test_router_chat_enables_thinking_when_request_omits_preference` (`test_router.py:639-656`) locks that behavior in. Nothing else from the prior threads is outstanding.

## Approach

Sound — acceptable with minor reservations. The trigger string is specific enough that it can only originate from an endpoint that supports reasoning, so the retry cannot misfire against a model that has no thinking mode at all; loop protection works (each action fires at most once); and the streaming path correctly declines to retry once any chunk has been yielded (`has_yielded` guard in `_run_streaming_with_provider_retry`).

Two design-level notes, both minor:

- **D2 — undocumented payload-shape assumption (`router.py:48-49`).** `_ENABLE_DOWNSTREAM_THINKING = {"type": "enabled", "enable": True}` is a valid literal only because Auto always dispatches through a single OpenRouter-shaped downstream client. That is derivable from the module docstring and `_resolve_route`, but it takes a reader a few minutes to re-derive. A one-line comment at the constant would pin it down.
- **D1 — predicate ordering is load-bearing but unasserted (`router.py:128-161`).** The three predicates are evaluated in fixed order with first-match-wins and nothing guarantees mutual exclusivity. Today's three trigger strings don't overlap, so this is inert — noting it only so it's on record if a future provider message ever satisfies two patterns at once.

## Findings

All minor or nit; nothing blocking.

**F3 (minor, test coverage) — `router.py:75-89`: the "already enabled → don't retry" guard is never tested as a first-call case.** No test issues an initial call already carrying `thinking={"type": "enabled", "enable": True}`, hits the mandatory-reasoning error, and asserts the error propagates immediately with exactly one downstream call. The branch is currently only exercised as the second attempt inside `test_router_does_not_repeat_mandatory_reasoning_retry` (`test_router.py:685-701`), which means a regression that made the guard unconditionally fall through would still leave that test green. A small standalone test would close this — see inline comment.

**F1 (minor, clarity — not a bug) — `router.py:75-89` vs the `attempted_retry_actions` dedup check.** For the `enable_thinking` action specifically, the thing that actually stops a second attempt is the guard at `router.py:80-81`: once the retry has swapped in `_ENABLE_DOWNSTREAM_THINKING`, the predicate returns `False`, so the `action_key in attempted_retry_actions` check is never reached on that path. Tracing `test_router_does_not_repeat_mandatory_reasoning_retry` confirms this. That's fine as defense-in-depth, and the dedup set is genuinely load-bearing for the other two actions — but a future reader is likely to assume the set is doing the work here. One line at the guard would prevent someone from "simplifying" it away.

**F4 (minor, optional, test coverage) — no assertion that the reasoning retry leaves `tool_choice` untouched.** Nothing verifies that firing `enable_thinking` doesn't also relax `tool_choice`. In practice the triggers are disjoint (mandatory-reasoning is a 400; the relax-tool-choice trigger is a `"no endpoints found"` 404 in the fixtures) and `_next_retry_state`'s ordering structurally guarantees pass-through — so this is low value. It is, however, nearly free: `_ScriptedChatLLM.tool_choices` already records what's needed.

**F5 (nit, docs) — `router.py:85-87`: comment omits the removal forward-reference.** The two sibling predicates (`router.py:98-101`, `router.py:120-123`) both end with a note about replacing the string match with typed provider errors once the follow-up tracking issue lands. Since all three are meant to be retired together, adding the same closing sentence here keeps them discoverable as one group.

**F2 (nit, style — theoretical only) — `router.py:80-81`: strict `is True` vs truthy checks downstream.** The guard uses `thinking.get("enable") is True`, while every downstream consumer reads it truthily (`openrouter.py:371`, `claude.py:603`/`:873`, `dashscope.py:102`, `deepseek.py:115`, `xinference.py:205-208`/`:425-428` all use `thinking.get("enable", False)`). No code path in the repo currently builds a non-bool truthy `enable`, so these cannot diverge today; `bool(thinking.get("enable"))` would just make them consistent by construction. Purely optional.

- `test_router.py:617-635` + `:639-656`: duplicate test bodies — `..._when_selected_model_requires_it` and `..._when_request_omits_preference` differ only in the initial `thinking` value and exercise the identical retry branch. Merge into one test with `@pytest.mark.parametrize("thinking", [None, {"type": "disabled", "enable": False}])`.
- `test_router.py:135-153`: `_ScriptedChatLLM.stream_chat` duplicates `chat` (`:115-133`) — same arg-discard, bookkeeping, and error-raising, differing only in return vs yield. Extract a shared helper. (Note: `_ToolChoiceRetryLLM` at `:41-94` has the same pattern and predates this PR, so this is stylistic, not this PR's obligation.)

net: -25 lines possible

## Tests

23/23 passed (`PYTHONPATH=src pytest tests/core/model/chat/basic/test_router.py`), matching the PR description.

## Recommendation

No blockers — the prior round's finding is properly fixed and the retry is narrowly scoped with working loop protection. The only item I'd genuinely encourage is F3 (a standalone test for the already-enabled guard, ~12 lines); everything else is optional polish that could equally ride along in a later pass.

src/xagent/core/model/chat/basic/router.py
qinxuye: Addressed in 8844a0c9. Added a standalone initial-enabled case that verifies the mandatory-reasoning error propagates after exactly one downstream call, including one unchanged `tool_choice` value.

src/xagent/core/model/chat/basic/router.py
qinxuye: Addressed in 8844a0c9. Added a comment documenting that the explicit-enabled guard is the primary stop condition after the reasoning retry, while retry-action tracking remains defense in depth for the shared loop.

src/xagent/core/model/chat/basic/router.py
qinxuye: Addressed in 8844a0c9. The mandatory-reasoning predicate now carries the same forward-looking note to replace string matching with typed provider errors.

src/xagent/core/model/chat/basic/router.py
qinxuye: Keeping the strict `is True` check intentionally. The normalized thinking contract uses booleans, and a truthy non-boolean value such as a string should not be accepted as an explicit enabled signal. No current caller emits a non-boolean value, so this avoids broadening the contract without changing valid behavior.

tests/core/model/chat/basic/test_router.py
qinxuye: Addressed in 8844a0c9. The parameterized reasoning-retry test now asserts `tool_choice` remains `required` on both attempts.

tests/core/model/chat/basic/test_router.py
qinxuye: Addressed in 8844a0c9. The omitted and explicitly disabled cases are now one parameterized test with descriptive IDs.

tests/core/model/chat/basic/test_router.py
qinxuye: Leaving this optional cleanup out of the focused fix. Sharing logic between a coroutine test double and an async-generator test double would add indirection for a small fixture, while the duplicated shape also matches the pre-existing retry fixture and does not leave a behavioral coverage gap.

rogercloud (commented):
# Follow-up review — xorbitsai/xagent#1120 (re-review @ 8844a0c9)

Re-review scope: update commit `8844a0c9` "fix: address Auto reasoning review feedback" on top of base `ac8812da`; previously reviewed at `f42107af`. All checklist statuses below were verified against HEAD code.

## PR summary

The PR teaches RouterLLM to survive OpenRouter's untyped 400 "Reasoning is mandatory for this endpoint and cannot be disabled." Auto routing sends strict tool-call requests with thinking explicitly disabled; some downstream endpoints require reasoning and reject that payload. A new `_should_retry_with_thinking` predicate (`src/xagent/core/model/chat/basic/router.py:75-92`) recognizes the error and drives one retry of the same selected model with `_ENABLE_DOWNSTREAM_THINKING` substituted, in both non-streaming and streaming calls, wired into the existing first-match-wins ladder in `_next_retry_state` ahead of the two pre-existing compatibility predicates, deduped through the shared `attempted_retry_actions` bookkeeping. No linked issues.

## Update summary

One commit, 8844a0c9 "fix: address Auto reasoning review feedback" (last reviewed at f42107af): adds two comment lines documenting that the explicit-enabled guard is the primary repeat-stop condition (router.py:80-81) and the typed-error migration marker (router.py:90); merges the omitted-preference and disabled-preference tests into one parametrized test with ids unspecified/disabled (test_router.py:616-620); adds `assert llm.tool_choices == ["required", "required"]` to the merged test (test_router.py:641); adds standalone `test_router_does_not_retry_mandatory_reasoning_when_already_enabled` (test_router.py:690-706).

## Prior findings checklist

- **P1** (gemini-code-assist[bot] inline, thinking=None must still allow retry): **FIXED** — `None` bypasses the explicit-enabled guard (`isinstance(thinking, dict)`, router.py:82) and falls through to the string match at router.py:92.
- **P2** (standalone already-enabled first-call test): **FIXED** — `test_router_does_not_retry_mandatory_reasoning_when_already_enabled` at test_router.py:690-706; runs green.
- **P3** (guard stop-condition comment): **FIXED** — router.py:80-81.
- **P4** (typed-error forward-reference sentence): **FIXED** — router.py:90.
- **P5** (strict `is True` vs truthy): **WAIVED by author choice** — verified rationale: every in-repo thinking dict construction uses real bools; no non-bool truthy `enable` exists anywhere in `src/` or `tests/`, so the strict check matches the normalized internal contract.
- **P6** (tool_choice untouched assertion): **FIXED** — test_router.py:641.
- **P7** (merge duplicate tests via parametrize): **FIXED** — test_router.py:616-620, net -11 lines.
- **P8** (extract shared chat/stream_chat test-double helper): **WAIVED by author choice** — verified rationale: duplicates the pre-existing `_ToolChoiceRetryLLM` pattern (test_router.py:41); optional cleanup, no coverage gap.
- **P9** (body-only minor: one-line comment at `_ENABLE_DOWNSTREAM_THINKING`, router.py:49): **NOT FIXED — still open**. The constant remains bare; the update commit only touched comments inside `_should_retry_with_thinking`. Concrete suggestion: a one-liner documenting the OpenRouter-shaped payload assumption, e.g. that the dict mirrors the `{"type", "enable"}` bool contract the router normalizes internally.
- **P10** (body-only informational: first-match-wins predicate ordering with no mutual-exclusivity guarantee): **INFORMATIONAL** — code state matches the note; the three trigger-string pairs remain pairwise disjoint (verified); nothing structural enforces exclusivity; no action needed.

## New finding

- **N1 (minor, test coverage)** — `tests/core/model/chat/basic/test_router.py:670`, `test_router_does_not_repeat_mandatory_reasoning_retry` never exercises the `"enable_thinking"` entry in `attempted_retry_actions`. Trace (verified): on the second scripted error the request already carries enabled thinking, so the primary enabled-state guard (router.py:82-85) returns False and the error propagates before the action-key dedup check (router.py:336) is reached — the dedup path for this key is unreachable by construction in both the chat and stream loops. The test would still pass if the action key were miswired or the `add()` call removed; it fails only if the whole `enable_thinking` branch were deleted. The dedup set stays load-bearing for the other two actions, but the new key's wiring is untested. Suggested coverage: scripted alternating errors — mandatory reasoning → thinking/tool_choice conflict (triggers `disable_thinking`) → mandatory reasoning — asserting exactly 3 downstream calls, thinking sequence `disabled → enabled → disabled`, and propagation of the third error. In that sequence the third mandatory-reasoning error passes the guard (thinking is disabled again) and is stopped only by the `enable_thinking` dedup check, exercising exactly the path N1 flags.

## Design verdict

Sound — carried from Round 0: smallest design consistent with the existing architecture; string matching is justified under the untyped-400 constraint, with the documented typed-error migration marker now in place. P10 remains on the record. No architecture or design-boundary concerns.

## Tests

`PYTHONPATH=src .venv/bin/python -m pytest -q tests/core/model/chat/basic/test_router.py` at HEAD 8844a0c9 → 24 passed, exit 0.

tests/core/model/chat/basic/test_router.py
```diff
@@ -589,6 +613,99 @@ async def test_router_chat_propagates_non_matching_errors_without_retry(monkeypa
...
+
+    assert [chunk.delta for chunk in chunks] == ["ok"]
+    assert llm.thinking_values == [
+        {"type": "disabled", "enable": False},
+        {"type": "enabled", "enable": True},
+    ]
+
+
+@pytest.mark.asyncio
+async def test_router_does_not_repeat_mandatory_reasoning_retry(monkeypatch):
```
rogercloud, line 670: Minor, non-blocking: this test never exercises the `enable_thinking` entry of `attempted_retry_actions`. On the second scripted error the request already carries enabled thinking, so the explicit-enabled guard in `_should_retry_with_thinking` (router.py:82-85) stops the retry before the action-key dedup check (router.py:336) is reached — the dedup path for this key is unreachable by construction in both chat and stream loops. The test would pass unchanged if the action key were miswired or the `add()` call removed; it fails only if the whole `enable_thinking` branch were deleted. The set stays load-bearing for the other two actions, but the new key's wiring is untested. Optional coverage: script alternating errors (mandatory reasoning → thinking/tool_choice conflict → mandatory reasoning) and assert exactly 3 downstream calls, thinking `disabled → enabled → disabled`, and propagation of the third error.

src/xagent/core/model/chat/basic/router.py
```diff
@@ -46,6 +46,7 @@
 _DEFAULT_ROUTER_ABILITIES = ["chat", "tool_calling"]
 _UNROUTED_ROUTER_ABILITIES = {"vision", "thinking_mode"}
 _DISABLE_DOWNSTREAM_THINKING = {"type": "disabled", "enable": False}
+_ENABLE_DOWNSTREAM_THINKING = {"type": "enabled", "enable": True}
```
rogercloud, line 50: Minor, body-only, still open from the last review: `_ENABLE_DOWNSTREAM_THINKING` remains bare while `_DISABLE_DOWNSTREAM_THINKING` and both string-matching predicates carry explanatory comments. The update commit only touched comments inside `_should_retry_with_thinking`. A one-liner documenting the OpenRouter-shaped payload assumption — e.g. that the dict mirrors the `{"type", "enable"}` bool contract the router normalizes internally — would help future readers, especially once the typed-error migration lands. Optional; not blocking.

tests/core/model/chat/basic/test_router.py
qinxuye: Addressed in ed561769. The no-repeat test now scripts mandatory reasoning, then a thinking/tool-choice conflict, then mandatory reasoning again. It asserts the disabled → enabled → disabled sequence and exactly three unchanged `required` tool choices, so the third error is stopped specifically by the previously recorded `enable_thinking` action key.

src/xagent/core/model/chat/basic/router.py
qinxuye: Addressed in ed561769. Added a one-line comment documenting that these normalized intents are translated by the selected OpenRouter adapter into its reasoning/thinking request body.

rogercloud (approved):
### Follow-up review (round 2) — `ed561769` "fix: cover Auto reasoning retry deduplication" — approving

**PR summary**

This PR teaches RouterLLM to survive OpenRouter's untyped 400 "Reasoning is mandatory for this endpoint and cannot be disabled." Auto routing sends strict tool-call requests with thinking explicitly disabled; some downstream endpoints require reasoning and reject that payload. A new `_should_retry_with_thinking` predicate (`src/xagent/core/model/chat/basic/router.py`) recognizes the error and drives one retry of the same selected model with `_ENABLE_DOWNSTREAM_THINKING` substituted, in both non-streaming and streaming calls, wired into the existing first-match-wins ladder in `_next_retry_state`, and deduped through the shared `attempted_retry_actions` bookkeeping. No linked issues.

**Update summary (this round)**

One commit since the last review (8844a0c9): `ed561769` (+10/-1). It (1) adds `# Normalized intents translated into OpenRouter's reasoning/thinking request body.` above the two thinking-intent constants (router.py:48), answering the open D2/N1 docs comment; (2) rewrites `test_router_does_not_repeat_mandatory_reasoning_retry` to script the alternating sequence `[_MANDATORY_REASONING_ERROR, _THINKING_TOOL_CHOICE_ERROR, _MANDATORY_REASONING_ERROR]` (test_router.py:672-676), asserting thinking `disabled → enabled → disabled` (test_router.py:690-692) and `tool_choices == ["required", "required", "required"]` (test_router.py:694).

**Prior findings checklist**

- **N1** (minor, test coverage; inline at test_router.py:670) — **FIXED**. Traced: call 3's mandatory-reasoning error passes the primary enabled-guard (thinking is disabled again) and is stopped only by the `enable_thinking` entry in `attempted_retry_actions` (router.py:337-338). Mutation-tested: removing the dedup check/add makes the test fail (a 4th call returns "ok", breaking `pytest.raises`); miswiring the action key raises at call 2 with an unmatched message.
- **P9/D2** (minor, docs; inline at router.py:49) — **FIXED**. The comment at router.py:48 documents the constants' role as the OpenRouter reasoning/thinking request-body payloads.
- **Earlier findings (P1-P8, P10)** from the 8844a0c9 round remain as previously reported: 7 FIXED, 2 WAIVED (P5 strict `is True`; P8 `stream_chat` test-double helper — author rationale verified), 1 INFORMATIONAL (P10 predicate ordering).

**New findings**

None. No open line-level findings remain.

**Design verdict**

Sound (unchanged from Round 0): the retry ladder, the `attempted_retry_actions` dedup, and the new coverage of the dedup path are coherent and well-tested.

**Tests**

`PYTHONPATH=src .venv/bin/python -m pytest -q tests/core/model/chat/basic/test_router.py` at `ed561769` → 24 passed, exit 0. The verifier's mutation test (restore after removal) also passes.

Issue #992, opened 2026-07-24
test(widget): cover WS per-message revalidation (4003 close) for agent + workforce widget guests


codeacme17:
## Background

Follow-up from the review of #989 (agent widget guest-token revalidation, @rogercloud).

Both the agent widget path (`ensure_widget_agent_available`, #988/#989) and the workforce widget path (`ensure_widget_workforce_available`, #985) are re-validated per inbound WS message: `public_chat_websocket_endpoint` re-invokes `get_public_chat_user` inside the receive loop and closes with code `4003` on failure, so a revoked guest is dropped on the next message.

This WS receive-loop revalidation / `4003` close is listed as an acceptance criterion but has **no test**. Current tests only exercise the HTTP `POST /api/widget/chat/task/create` re-check. The gap predates #989 — the #985 workforce path is equally untested here — so it is not a regression, but it is an open gap on an explicitly-called-out path.

## Proposed change

Add a WS test (covering both the agent and workforce widget paths) that:

1. Connects a guest over the widget WS endpoint with a valid token.
2. Disables the widget (or rotates its key).
3. Sends one more inbound message.
4. Asserts the socket is closed with code `4003`.

## Notes

- Consider folding in the small defense-in-depth `isinstance(user_id, int)` guard in `ensure_widget_agent_available` / the workforce helpers that `get_share_chat_user` already applies, while touching this area.

PR #1121, opened 2026-08-04
enh(widget): report connect-time revocation as 4003 with the real reason (#1057)

### LINKED ISSUE #1057: enh(widget): connect-time revocation reports a bare 4001, losing the real close reason

codeacme17:
## Summary

Closes #1057.

Aligns the widget WebSocket endpoint `public_chat_websocket_endpoint` with the sibling `share_chat_websocket_endpoint` on how a **connect-time** auth failure is reported.

Previously the widget path authorized *before* accepting the handshake and closed pre-accept on any exception:

```python
except Exception:
    await websocket.close(code=4001, reason="Authentication required")
```

Two problems fell out of that:

1. uvicorn only preserves a close code + reason for a **post-accept** close; a pre-accept close collapses into a bare HTTP 403, so the browser sees `code=1006, reason=""`. The real denial (`4003` / "Widget is unavailable") was discarded — the widget client could not tell "your widget was revoked" from "the network dropped".
2. The bare `except Exception` flattened an `HTTPException` from the revocation checks into the same generic `4001` as a malformed token.

This produced an asymmetry *within the widget path itself*: mid-session revocation gives a clean `4003` + reason (covered by #992/#1049), but a guest who reconnects after revocation — the far more common sequence, since a revoked guest reloads the page — got nothing actionable.

## Changes

- **Accept-first, then map exceptions** (matching the share path): the pre-accept per-IP connect budget from #1056 is preserved, then `websocket.accept()` runs before auth. An auth-time `HTTPException` maps to `4003` + the real reason; only non-HTTP failures keep `4001`.
- Switch `manager.connect` → `manager.register_connection`, since `accept()` now happens explicitly and `connect()` would double-accept.
- Update the two stale comments that justified the old pre-accept-4001 behaviour.

## Ordering / #1056

Accept-first means every widget connect attempt completes a 101 upgrade before rejection, which is why the share path needed a per-IP connect budget. That budget landed for the widget path in #1056 (merged), so this change builds on it and keeps the connect gate pre-accept.

## Tests

- Rename/split the pinned `test_public_access_websocket_initial_auth_failure_stays_4001` into an HTTP→`4003` case (`..._initial_http_auth_failure_maps_to_4003`) and a non-HTTP→`4001` case (`..._initial_non_http_failure_stays_4001`).
- Add end-to-end `test_widget_ws_reconnect_after_revocation_reports_4003`, parametrized over all three revocation levers (disable / rotate / unpublish) × agent/workforce — the reconnect-after-revocation case the issue flagged as untested.
- Re-pin `test_widget_ws_connect_over_limit_is_refused_pre_accept` on the pre-accept-`4008` vs post-accept-`4003` ordering, mirroring the share gate test.

## No frontend change needed

`use-websocket.ts` already treats a `4003` as a permanent failure carrying `event.reason`, and the widget page's "clear stale taskId + recover" path is scoped to `authMode === "share"`. A revoked widget guest now surfaces a terminal error with the real reason ("Widget is unavailable") instead of a generic transport failure.

## Verification

- Affected test files: 52 passed; broader public_chat/share/widget/websocket sweep green.
- Full pre-commit hook set passes (ruff, mypy, isort, codespell).

### DIFF
--- a/tests/web/api/test_public_chat_websocket_db_boundary.py
+++ b/tests/web/api/test_public_chat_websocket_db_boundary.py
@@ -281,16 +281,23 @@ async def test_public_access_websocket_preserves_auth_close_codes(
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize("endpoint_kind", ["public", "share"])
-async def test_public_access_websocket_initial_auth_failure_stays_4001(
+async def test_public_access_websocket_initial_http_auth_failure_maps_to_4003(
     monkeypatch: pytest.MonkeyPatch,
     endpoint_kind: str,
 ) -> None:
+    # Both endpoints now accept the handshake before auth and map an auth-time
+    # HTTPException to a 4003 close carrying the real reason, so the frontend
+    # recovery flow can act on it (#973 for share, #1057 for widget). This is
+    # the connect-time analogue of the per-message revalidation 4003 — the far
+    # more common sequence for a revoked widget guest, who reloads the page and
+    # reconnects rather than staying on a live socket.
     websocket = _SingleMessageWebSocket({"type": "chat", "message": "hello"})
     authorize = AsyncMock(
-        side_effect=HTTPException(status_code=401, detail="Invalid token")
+        side_effect=HTTPException(status_code=403, detail="Widget is unavailable")
     )
     connection_manager = MagicMock()
     connection_manager.connect = AsyncMock()
+    connection_manager.register_connection = MagicMock()
     connection_manager.disconnect = MagicMock()
     monkeypatch.setattr(public_chat_access, "manager", connection_manager)
 
@@ -320,18 +327,67 @@ async def test_public_access_websocket_initial_auth_failure_stays_4001(
             token="share-token",
         )
 
+    # Accept-first is what lets the 4003 reason survive the handshake instead of
+    # collapsing into a bare HTTP 403 (browser sees 1006/"").
+    websocket.accept.assert_awaited_once()
+    websocket.close.assert_awaited_once_with(
+        code=4003,
+        reason="Widget is unavailable",
+    )
+    # Rejected before registration: neither the async accept-and-register helper
+    # nor the register-only path runs, so the socket never joins task broadcasts.
+    connection_manager.connect.assert_not_awaited()
+    connection_manager.register_connection.assert_not_called()
+    connection_manager.disconnect.assert_not_called()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("endpoint_kind", ["public", "share"])
+async def test_public_access_websocket_initial_non_http_failure_stays_4001(
+    monkeypatch: pytest.MonkeyPatch,
+    endpoint_kind: str,
+) -> None:
+    # A non-HTTP failure at connect-time auth (an infra error, not an access
+    # denial) carries no reason the recovery flow keys on, so both endpoints
+    # keep the generic 4001 rather than dressing it up as a 4003 access denial.
+    websocket = _SingleMessageWebSocket({"type": "chat", "message": "hello"})
+    authorize = AsyncMock(side_effect=RuntimeError("database is down"))
+    connection_manager = MagicMock()
+    connection_manager.connect = AsyncMock()
+    connection_manager.register_connection = MagicMock()
+    connection_manager.disconnect = MagicMock()
+    monkeypatch.setattr(public_chat_access, "manager", connection_manager)
+
     if endpoint_kind == "public":
-        websocket.close.assert_awaited_once_with(
-            code=4001,
-            reason="Authentication required",
+        monkeypatch.setattr(
+            public_chat_access,
+            "_authorize_public_chat_websocket",
+            authorize,
+            raising=False,
+        )
+        await public_chat_access.public_chat_websocket_endpoint(
+            websocket=websocket,
+            task_id=42,
+            token="widget-token",
+            expected_auth_mode="widget",
         )
     else:
-        # The share endpoint accepts before auth and maps an auth-time
-        # HTTPException to a 4003 close carrying the real reason, so the frontend
-        # recovery flow can act on it (#973). Only non-HTTP failures stay 4001.
-        websocket.close.assert_awaited_once_with(
-            code=4003,
-            reason="Invalid token",
+        monkeypatch.setattr(
+            public_chat_access,
+            "_authorize_share_chat_websocket",
+            authorize,
+            raising=False,
         )
+        await public_chat_access.share_chat_websocket_endpoint(
+            websocket=websocket,
+            task_id=42,
+            token="share-token",
+        )
+
+    websocket.close.assert_awaited_once_with(
+        code=4001,
+        reason="Authentication required",
+    )
     connection_manager.connect.assert_not_awaited()
+    connection_manager.register_connection.assert_not_called()
     connection_manager.disconnect.assert_not_called()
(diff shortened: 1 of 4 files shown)


tests/web/api/test_public_chat_websocket_db_boundary.py
```diff
@@ -320,18 +327,67 @@ async def test_public_access_websocket_initial_auth_failure_stays_4001(
...
+            authorize,
+            raising=False,
         )
+        await public_chat_access.share_chat_websocket_endpoint(
+            websocket=websocket,
+            task_id=42,
+            token="share-token",
+        )
+
+    websocket.close.assert_awaited_once_with(
```
rogercloud, line 387: minor (test coverage): this test does not assert `websocket.accept.assert_awaited_once()`, unlike its sibling `test_public_access_websocket_initial_http_auth_failure_maps_to_4003` (L332), even though `accept()` is called unconditionally before the try/except that produces both the 4003 and 4001 outcomes (src/xagent/web/api/public_chat_access.py:1285). A regression that moved the accept into only the HTTPException branch would be caught by the 4003 test but could slip past this one. Suggest adding `websocket.accept.assert_awaited_once()` here, mirroring L332.

rogercloud (approved):
## Summary

This PR fixes the widget WebSocket handshake so a revoked guest gets an actionable close signal. `public_chat_websocket_endpoint` previously authorized *before* accepting the handshake and closed pre-accept with a blanket 4001; because uvicorn only preserves a close code/reason for a post-accept close, that collapsed into a bare HTTP 403 and the browser saw `code=1006, reason=""`. The endpoint now accepts first, maps `HTTPException` → 4003 with the real reason, keeps `except Exception` → 4001 as a fallback, and switches `manager.connect` → `manager.register_connection` so the accept is not performed twice. This brings the widget path in line with the already-shipped `share_chat_websocket_endpoint` behavior and closes the gap described in #1057.

## Approach: sound

The change ports an existing, already-reviewed sibling pattern rather than inventing new mechanism, which is the right call for a security-sensitive gate. `register_connection` is plain synchronous bookkeeping and does not call `accept()`, so there is no double-accept; `get_public_chat_user` normalizes all failure modes into `HTTPException`, making `except HTTPException` the correct and complete catch boundary (the residual 4001 branch is effectively unreachable in production, which the new non-HTTP test documents), and `CancelledError` remains a `BaseException` so cancellation still propagates instead of being swallowed as 4001. The accept-first ordering does not widen the resource-exhaustion surface either, since the pre-existing per-IP pre-accept connect budget (#1056) is still the first gate — the same posture the share path already ships with.

Non-blocking observation: the connect-time exception-handling blocks of `public_chat_websocket_endpoint` and `share_chat_websocket_endpoint` are now structurally near-identical. Extracting a shared helper is *not* requested here — the surrounding turn-gate logic genuinely differs (different key composition, different guards), so a helper would need callback injection and would trade a straightforward per-endpoint branch for indirection in an auth path. Worth keeping in mind for a future dedicated refactor, not for this PR.

`src/xagent/web/api/public_chat_access.py`
- L1275-1284: shrink — the 10-line pre-`accept()` comment mixes the load-bearing WHY (uvicorn's pre/post-accept close-code asymmetry) with decorative narrative recoverable from git blame and #1057. Trim to ~6 lines keeping only the asymmetry and why accept-first fixes it.

`tests/web/api/test_public_chat_websocket_db_boundary.py`
- L288-293: shrink — half the comment duplicates the production comment's conclusion (reason reaching the client / recovery). Keep only the test-specific framing: this is the connect-time analogue of per-message revalidation.
- L330-331: delete — restates the "accept-first survives the handshake / pre-accept collapses to 1006" fact already stated twice above.

`tests/web/api/test_widget_ws_revalidation.py`
- L361-373: shrink — the 13-line docstring re-derives near-verbatim what the module docstring already says at L26-31. Keep only "connect-time counterpart of the mid-session test above; all three revocation levers covered."
- L26-31: shrink — counterpart to the above; pick one as canonical and have the other point to it.

net: -19 lines possible

## Test results

38/38 passed across the three touched test files (`test_public_chat_websocket_db_boundary.py`, `test_share_rate_limit_endpoints.py`, `test_widget_ws_revalidation.py`).

One minor test-coverage nit left as an inline comment below — not blocking. Approving.

tests/web/api/test_public_chat_websocket_db_boundary.py
codeacme17: Good catch — fixed in 20f450d3. Added `websocket.accept.assert_awaited_once()` to `test_public_access_websocket_initial_non_http_failure_stays_4001`, mirroring the 4003 test, so a regression that moved `accept()` into only the `HTTPException` branch is caught on the 4001 path too.

codeacme17:
Thanks for the thorough review and approval, @rogercloud.

Applied the simplification pass in 20f450d3 (net −11 lines this commit), plus the inline test-coverage nit:

- **Inline nit (test coverage):** added `websocket.accept.assert_awaited_once()` to `test_public_access_websocket_initial_non_http_failure_stays_4001`, mirroring the 4003 test, so a regression moving `accept()` into only the `HTTPException` branch is caught on the 4001 path too.
- **`public_chat_access.py` L1275-1284:** trimmed the pre-`accept()` comment to keep only the uvicorn pre/post-accept asymmetry and why accept-first fixes it.
- **`test_public_chat_websocket_db_boundary.py`:** shortened the 4003 test's comment to the connect-time-analogue framing, and deleted the redundant "accept-first survives the handshake" restatement.
- **`test_widget_ws_revalidation.py`:** shrank the reconnect test docstring to point at the module docstring, which I kept as the canonical explanation of the accept-first rationale.

On the shared-helper observation — agreed it's out of scope here. The two endpoints' surrounding turn-gate logic genuinely differs (key composition + guards), so a connect-time helper would need callback injection into an auth path; better left to a dedicated refactor. Local run still green: 38/38 across the three touched test files.

PR #1122, opened 2026-08-04
fix(task-orchestrator): reconcile pending delivery rows at turn finalize

codeacme17:
## Problem

A user-message delivery row is written `pending` with the turn claim and projected to `dispatched` best-effort right after scheduling. When that projection fails — a database pool checkout timeout or a transient DB error — the row stays `pending` forever even though the turn actually ran. Nothing on the orchestrator path advances it afterward.

Downstream, the session-WebSocket dedup probe (`probe_external_task_delivery`, in xagent-saas) classifies every same-`client_message_id` retry against that stale row as an in-flight `PENDING` and rejects it (busy, without `retry_with_new_id`). The client can never converge on a turn that has already executed.

This is the P2 finding "[P2] Reconcile delivery rows left permanently pending" split out of the SA-8 session WebSocket work, tracked as xorbitsai/xagent-saas#332.

## Fix

Add a durable terminal transition in the shared delivery layer at turn finalize. New `_reconcile_finalized_turn_delivery(...)` closes the turn's delivery row once its lease is terminally settled **by this coroutine**:

- `completed` on a clean settlement, `failed` otherwise.
- The transition is monotonic, so an already-terminal row — or a normally `dispatched` row on a later failure — is left untouched rather than regressed.
- Missing rows and pool timeouts are safe no-ops that leave the row for durable recovery.

Wired into `_runner`'s `finally` block and gated on an **authoritative** settlement — reconciliation is skipped (leaving the row `pending`) whenever this coroutine is not the rightful closer:

- settlement deferred to TTL recovery (pool timeout, lost lease, unhealthy heartbeat);
- `settle_task_lease_isolated` raised (lease retained for TTL recovery);
- settle returned `False` without raising — fenced out / superseded lease, or a pre-existing terminal/control outcome reconciled instead of this coroutine's view;
- the pre-execution `SETTLEMENT_READY` short-circuit fired (the turn body never ran, so `completed` would ack a message that was never consumed).

Known residual windows, documented rather than claimed solved: a hard process crash before `_runner`'s `finally`, and the deferred-settlement leftovers (the lease-expiry recovery module terminalizes the *task* but never redrives the turn or touches delivery rows). Of those, only the paused-resume path has an in-repo consumer that re-posts the same `turn_id`. The final delivery state is intentionally **not** a proxy for the turn's outcome — it answers "may this client_message_id be retried?", never "did the turn succeed?".

Also extracted `_log_delivery_projection_failure`, shared by the post-schedule and finalize projection sites (pool timeouts keep their explicit marker; everything else keeps its traceback via `logger.exception`).

## Tests

16 tests in `tests/web/services/test_task_orchestrator.py` for this change (74 total in the file, all passing):

- Unit: reconciliation advances an orphaned `pending` row to `completed` / `failed`; `dispatched` is not regressed on a later failure; `dispatched → completed` closes on success; already-terminal rows are no-ops under idempotent and conflicting reconciles; pool timeout leaves `pending` without raising; no-`turn_id` is a no-op.
- End-to-end through `_runner` (shared fixture + patch scaffolding): orphaned `pending` finalizes to `completed`; deferred-to-TTL, settlement-raises, fenced-out settle (`False` return), and the `SETTLEMENT_READY` short-circuit all keep the row `pending`; a cancelled turn closes the row to `failed`; a `CancelledError` from the reconcile call site is recorded and re-raised after connector cleanup.

Refs xorbitsai/xagent-saas#332, xorbitsai/xagent-saas#268. Companion xagent-saas PR (xorbitsai/xagent-saas#404) bumps the submodule and adds the probe-side acceptance test.

### DIFF
--- a/src/xagent/web/services/task_orchestrator.py
+++ b/src/xagent/web/services/task_orchestrator.py
@@ -703,22 +703,47 @@ async def _schedule_committed_turn(
         # Claim commit + background scheduling already succeeded. A transient
         # delivery projection failure must not turn that success into an API
         # error or invite the caller to retry a turn that is already running.
-        if is_database_pool_timeout(delivery_error):
-            logger.error(
-                "task_id=%s component=turn-delivery database pool checkout timed "
-                "out after scheduling; leaving delivery pending for durable "
-                "recovery: %s",
-                task_id,
-                delivery_error,
-                exc_info=True,
-            )
-        else:
-            logger.exception(
-                "task_id=%s component=turn-delivery projection failed after "
-                "scheduling; leaving delivery pending for durable recovery",
-                task_id,
-            )
+        _log_delivery_projection_failure(
+            task_id,
+            component="turn-delivery",
+            error=delivery_error,
+            detail=" after scheduling",
+        )
     return handle
+
+
+def _log_delivery_projection_failure(
+    task_id: int,
+    *,
+    component: str,
+    error: Exception,
+    detail: str = "",
+) -> None:
+    """Log one swallowed delivery-projection failure without re-raising.
+
+    Pool checkout timeouts get their explicit marker — the committed row stays
+    reclaimable and the caller must not issue another checkout against the
+    exhausted pool. Anything else keeps its traceback via ``logger.exception``.
+    """
+
+    if is_database_pool_timeout(error):
+        logger.error(
+            "task_id=%s component=%s database pool checkout timed out%s; "
+            "leaving delivery pending for durable recovery: %s",
+            task_id,
+            component,
+            detail,
+            error,
+            exc_info=True,
+        )
+    else:
+        logger.exception(
+            "task_id=%s component=%s projection failed%s; "
+            "leaving delivery pending for durable recovery",
+            task_id,
+            component,
+            detail,
+        )
 
 
 def _reconcile_finalized_turn_delivery(
@@ -727,7 +752,7 @@ def _reconcile_finalized_turn_delivery(
     turn_id: str | None,
     settlement_error: str | None,
 ) -> None:
-    """Close a settled turn's delivery row so it can never orphan at ``pending``.
+    """Close a terminally settled turn's delivery row (xorbitsai/xagent-saas#332).
 
     A user-message delivery row is written ``pending`` with the turn claim and
     projected to ``dispatched`` right after scheduling. That projection is
@@ -735,18 +760,28 @@ def _reconcile_finalized_turn_delivery(
     ``pending`` even though the turn ran. Nothing else on the orchestrator path
     advances it, so the session WS dedup probe classifies every
     same-``client_message_id`` retry as an in-flight ``PENDING`` and rejects it
-    forever (#332).
+    forever.
 
-    Reconciling at finalize closes that window for every turn whose ``_runner``
-    reaches its ``finally`` (a hard process crash before then still orphans the
-    row and remains the job of TTL/durable recovery). Once the background run
-    has terminally settled its lease the turn is no longer in flight, so its
-    delivery is safe to close monotonically: ``completed`` on a clean
-    settlement, ``failed`` otherwise. The transition is monotonic, so an
-    already-terminal row — or a normally ``dispatched`` row on a later failure —
-    is left untouched rather than regressed. Callers MUST gate this on a
-    terminal settlement: a run deferred to TTL recovery may re-run and must keep
-    its ``pending`` row for that path to redrive.
+    Reconciling at finalize closes that window for every turn that reaches a
+    terminal lease settlement owned by its coroutine. The call is intentionally
+    skipped — leaving the row ``pending`` — when the coroutine is not the
+    authoritative closer: the lease went to another worker, settlement was
+    deferred to TTL recovery or raised, the settle was fenced out / superseded
+    (returned ``False``), or the pre-execution SETTLEMENT_READY short-circuit
+    fired. A hard process crash skips it too. Of those leftovers, only the
+    paused-task resume path has an in-repo consumer that re-posts the same
+    ``turn_id``; ``task_lease_recovery`` terminalizes the *task* but never
+    redrives the turn or touches delivery rows, so the deferral/crash cases
+    remain a narrower instance of the stuck-``pending`` window, not a solved
+    one.
+
+    The transition is monotonic: ``completed`` on a clean settlement, ``failed``
+    otherwise; an already-terminal row — or a normally ``dispatched`` row on a
+    later failure — is left untouched rather than regressed. Because both
+    projections are best-effort, the final delivery state is NOT a reliable
+    proxy for the turn's outcome: a failed turn may rest at ``dispatched`` or
+    ``failed`` depending on which projection landed. The delivery state answers
+    "may this client_message_id be retried?", never "did the turn succeed?".
     """
 
     if turn_id is None:
@@ -757,23 +792,12 @@ def _reconcile_finalized_turn_delivery(
     except Exception as delivery_error:
         # Best-effort projection. Leaving the row for durable recovery is strictly
         # better than amplifying pool exhaustion or masking the turn's outcome.
-        if is_database_pool_timeout(delivery_error):
-            logger.error(
-                "task_id=%s component=turn-finalize-delivery database pool "
-                "checkout timed out; leaving delivery pending for durable "
-                "recovery: %s",
-                task_id,
-                delivery_error,
-                exc_info=True,
-            )
-        else:
-            logger.warning(
-                "task_id=%s component=turn-finalize-delivery projection to %s "
-                "failed; leaving delivery for durable recovery",
-                task_id,
-                target,
-                exc_info=True,
-            )
+        _log_delivery_projection_failure(
+            task_id,
+            component="turn-finalize-delivery",
+            error=delivery_error,
+            detail=f" (target={target})",
+        )
 
 
 def _turn_started_snapshot(
@@ -1598,6 +1622,7 @@ def _schedule_bg(
         settlement_error: str | None = None
         broadcast_error_message: str | None = None
         defer_settlement_to_ttl_recovery = False
+        skip_delivery_reconciliation = False
         cleanup_cancellation: asyncio.CancelledError | None = None
         try:
             # A cancellation can land after the worker commits but before the
@@ -1652,6 +1677,12 @@ def _schedule_bg(
                         )
                         return
                     if validation == TaskLeaseRefreshState.SETTLEMENT_READY:
+                        # The task reached a terminal/control state before this
+                        # delayed run executed anything. Settlement below is
+                        # still correct, but the turn body never ran, so the
+                        # delivery row must NOT be closed as ``completed`` —
+                        # leave it for the authoritative writer.
+                        skip_delivery_reconciliation = True
                         return
 
                 async def execute_owned_run() -> None:
@@ -1777,9 +1808,15 @@ def _schedule_bg(
                                 error_message=settlement_error,
                             )
                         )
-                        # Settle returned without raising, so the task reached a
-                        # terminal decision and this runner will not re-run it.
-                        lease_settled = True
+                        # Gate on the returned value, not on "didn't raise":
+                        # ``settle_task_lease_isolated`` returns ``False``
+                        # without raising when the settlement was fenced out
+                        # (lease row gone or owned by a newer run) or when a
+                        # pre-existing terminal/control outcome was reconciled
+                        # instead of this coroutine's view. In both cases this
+                        # coroutine is no longer authoritative for the turn and
+                        # must not close its delivery row.
+                        lease_settled = bool(settled)
                         if settled and broadcast_error_message is not None:
                             try:
                                 await websocket_manager.broadcast_to_task(
@@ -1810,15 +1847,17 @@ def _schedule_bg(
                             exc_info=True,
                         )
 
-                    # Only once the lease is terminally settled is this turn
-                    # guaranteed not to re-run, so close its delivery row (#332):
+                    # Only when this coroutine's own settlement committed is it
+                    # safe to close the delivery row (xorbitsai/xagent-saas#332):
                     # a projection that never reached ``dispatched`` would
                     # otherwise orphan the row at ``pending`` and reject every
-                    # same-id retry forever. If settle raised, the lease is
-                    # retained for TTL recovery, which may redrive the turn — the
-                    # ``pending`` row must survive for that path, exactly as it
-                    # does for the ``defer_settlement_to_ttl_recovery`` branch.
-                    if lease_settled:
+                    # same-id retry forever. Skipped whenever this coroutine is
+                    # not the authoritative closer: settle raised (lease retained
+                    # for TTL recovery), settle was fenced out / superseded
+                    # (returned ``False``), settlement was deferred, or the
+                    # pre-execution SETTLEMENT_READY short-circuit fired (the
+                    # turn body never ran, so ``completed`` would be a lie).
+                    if lease_settled and not skip_delivery_reconciliation:
                         try:
                             await run_db_io_cancellation_safe(
                                 lambda: _reconcile_finalized_turn_delivery(
--- a/src/xagent/web/services/task_orchestrator.py
+++ b/src/xagent/web/services/task_orchestrator.py
@@ -751,6 +751,7 @@ def _reconcile_finalized_turn_delivery(
     task_id: int,
     turn_id: str | None,
     settlement_error: str | None,
+    execution_started: bool,
 ) -> None:
     """Close a terminally settled turn's delivery row (xorbitsai/xagent-saas#332).
 
@@ -775,18 +776,42 @@ def _reconcile_finalized_turn_delivery(
     remain a narrower instance of the stuck-``pending`` window, not a solved
     one.
 
-    The transition is monotonic: ``completed`` on a clean settlement, ``failed``
-    otherwise; an already-terminal row — or a normally ``dispatched`` row on a
-    later failure — is left untouched rather than regressed. Because both
-    projections are best-effort, the final delivery state is NOT a reliable
-    proxy for the turn's outcome: a failed turn may rest at ``dispatched`` or
-    ``failed`` depending on which projection landed. The delivery state answers
-    "may this client_message_id be retried?", never "did the turn succeed?".
+    Target selection is driven by what finalize actually knows, because the
+    downstream contract is asymmetric: the session WS probe answers a
+    ``failed`` row with ``retry_with_new_id`` — an invitation to re-send the
+    same content — so ``failed`` is only safe with positive evidence the
+    message was never consumed.
+
+    - ``settlement_error is None`` → ``completed``: the turn ran to a clean
+      settlement.
+    - error and ``execution_started`` → ``dispatched``: the run may have
+      consumed the message (and produced side effects) before failing, so the
+      one certain fact is "no longer in flight". A retry invitation here would
+      double-execute the turn. The probe treats ``dispatched`` as delivered;
+      the turn's failure is surfaced through task status, not by reopening
+      delivery. This also removes the timing asymmetry with the post-schedule
+      projection: whichever best-effort write lands, the row converges on
+      ``dispatched``.
+    - error and not ``execution_started`` → ``failed``:
+      ``execute_task_background`` was never invoked (setup failed, or the run
+      was cancelled first), so the message was provably never consumed and a
+      fresh-id retry is safe.
+
+    The transition is monotonic; an already-terminal row — or a normally
+    ``dispatched`` row on a later failure — is left untouched rather than
+    regressed. The final delivery state is NOT a proxy for the turn's outcome:
+    it answers "may this client_message_id be retried?", never "did the turn
+    succeed?".
     """
 
     if turn_id is None:
         return
-    target = DELIVERY_COMPLETED if settlement_error is None else DELIVERY_FAILED
+    if settlement_error is None:
+        target = DELIVERY_COMPLETED
+    elif execution_started:
+        target = DELIVERY_DISPATCHED
+    else:
+        target = DELIVERY_FAILED
     try:
         mark_user_message_delivery_sync(task_id, turn_id, target)
     except Exception as delivery_error:
@@ -1623,6 +1648,11 @@ def _schedule_bg(
         broadcast_error_message: str | None = None
         defer_settlement_to_ttl_recovery = False
         skip_delivery_reconciliation = False
+        # Positive evidence for finalize's delivery target: once
+        # ``execute_task_background`` has been invoked, the run may have
+        # consumed the user message, so a failure must not be projected as a
+        # retryable ``failed`` delivery (it would invite a double execution).
+        execution_started = False
         cleanup_cancellation: asyncio.CancelledError | None = None
         try:
             # A cancellation can land after the worker commits but before the
@@ -1702,6 +1732,8 @@ def _schedule_bg(
                     scope = await run_db_io_cancellation_safe(
                         lambda: resolve_execution_scope(task_id)
                     )
+                    nonlocal execution_started
+                    execution_started = True
                     await execute_task_background(
                         task_id=task_id,
                         user_message=payload.transcript_message,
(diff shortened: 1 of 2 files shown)


src/xagent/web/services/task_orchestrator.py
```diff
@@ -1712,13 +1769,17 @@ async def execute_owned_run() -> None:
...
                     try:
                         settled = await run_db_io_cancellation_safe(
                             lambda: settle_task_lease_isolated(
                                 lease,
                                 error_message=settlement_error,
                             )
                         )
+                        # Settle returned without raising, so the task reached a
+                        # terminal decision and this runner will not re-run it.
+                        lease_settled = True
```
rogercloud: **Major:** a fenced/superseded lease can still close the delivery row.

`settle_task_lease_isolated` can return `False` *without raising* when the underlying `finish_turn` is fenced out — i.e. the lease row was deleted, or is now owned by a different run (see `finish_turn`'s `runner_id` + `run_id` fencing). But `lease_settled` is set to `True` purely on the basis that the `await` didn't throw, discarding the returned boolean. A stale, no-longer-authoritative coroutine therefore still reaches `_reconcile_finalized_turn_delivery` and closes the delivery row to `completed`/`failed` on behalf of a turn it no longer owns — potentially overwriting the authoritative run's view of that delivery.

Suggested fix: gate reconciliation on the actual returned value, not on "no exception raised":

```python
lease_settled = bool(settled)
```

There is currently no test covering `settle_task_lease_isolated` returning `False` (fenced) with a delivery row present, so a regression test for this fix would also close that coverage gap.

src/xagent/web/services/task_orchestrator.py
```diff
@@ -1748,7 +1809,32 @@ async def execute_owned_run() -> None:
...
+                    # a projection that never reached ``dispatched`` would
+                    # otherwise orphan the row at ``pending`` and reject every
+                    # same-id retry forever. If settle raised, the lease is
+                    # retained for TTL recovery, which may redrive the turn — the
+                    # ``pending`` row must survive for that path, exactly as it
+                    # does for the ``defer_settlement_to_ttl_recovery`` branch.
+                    if lease_settled:
+                        try:
+                            await run_db_io_cancellation_safe(
+                                lambda: _reconcile_finalized_turn_delivery(
```
rogercloud, line 1915: **Minor (untested defensive gap):** (anchored here at the new reconcile call, since the actual trigger — the pre-existing `SETTLEMENT_READY` early return — isn't part of this diff) the `SETTLEMENT_READY` early return reaches reconciliation without the turn body ever running.

When `validate_preacquired_task_lease_isolated` returns `SETTLEMENT_READY`, `_schedule_bg` returns immediately before executing the turn body, `defer_settlement_to_ttl_recovery` stays `False`, and the `finally` block's mechanical chain (settle succeeds -> `lease_settled=True` -> `settlement_error is None`) closes the delivery row to `completed` — even though this turn never executed. That chain is confirmed, and this branch is entirely untested.

That said, no concrete caller in this repo appears to actually drive `SETTLEMENT_READY` at this specific pre-execution checkpoint: every writer that flips terminal task status while preserving `runner_id`/`run_id` runs only *inside* `execute_task_background`, i.e. after the turn already executed. So this reads as a defensive gap rather than a proven live data-loss path.

Recommend: add a regression test for this branch, and consider whether pre-execution `SETTLEMENT_READY` should defer to TTL recovery the way `LOST` does, out of caution.

src/xagent/web/services/task_orchestrator.py
```diff
@@ -720,6 +721,61 @@ async def _schedule_committed_turn(
...
+    advances it, so the session WS dedup probe classifies every
+    same-``client_message_id`` retry as an in-flight ``PENDING`` and rejects it
+    forever (#332).
+
+    Reconciling at finalize closes that window for every turn whose ``_runner``
+    reaches its ``finally`` (a hard process crash before then still orphans the
+    row and remains the job of TTL/durable recovery). Once the background run
+    has terminally settled its lease the turn is no longer in flight, so its
+    delivery is safe to close monotonically: ``completed`` on a clean
+    settlement, ``failed`` otherwise. The transition is monotonic, so an
```
rogercloud: **Minor:** the docstring overclaims coverage. It says reconciliation "closes that window for every turn whose `_runner` reaches its `finally`", but the call is skipped — leaving the row `pending` forever — in three cases that all still reach `finally`:

1. `lease is None` (lease acquired by another worker): the early return combined with the `if lease is not None:` guard short-circuits the whole settle+reconcile block.
2. `defer_settlement_to_ttl_recovery` is `True`.
3. `settle_task_lease_isolated` raises.

Please tighten the wording (e.g. "for every turn that reaches a terminal lease settlement"), otherwise a future reader will assume the stuck-`pending` state is now impossible.

src/xagent/web/services/task_orchestrator.py
```diff
@@ -720,6 +721,61 @@ async def _schedule_committed_turn(
...
+            logger.error(
+                "task_id=%s component=turn-finalize-delivery database pool "
+                "checkout timed out; leaving delivery pending for durable "
+                "recovery: %s",
+                task_id,
+                delivery_error,
+                exc_info=True,
+            )
+        else:
+            logger.warning(
```
rogercloud: **Minor (consistency):** this non-pool-timeout branch uses `logger.warning`, while its closest sibling — the equivalent "projection failed after scheduling" branch in `_schedule_committed_turn` at L716 — uses `logger.exception` for the same class of failure (a swallowed projection exception, logged without re-raising). Consider aligning to `logger.exception` here so the traceback is preserved for both.

src/xagent/web/services/task_orchestrator.py
```diff
@@ -720,6 +721,61 @@ async def _schedule_committed_turn(
...
+) -> None:
+    """Close a settled turn's delivery row so it can never orphan at ``pending``.
+
+    A user-message delivery row is written ``pending`` with the turn claim and
+    projected to ``dispatched`` right after scheduling. That projection is
+    best-effort: a pool timeout or a transient DB error leaves the row
+    ``pending`` even though the turn ran. Nothing else on the orchestrator path
+    advances it, so the session WS dedup probe classifies every
+    same-``client_message_id`` retry as an in-flight ``PENDING`` and rejects it
+    forever (#332).
```
rogercloud: **Minor (docs):** the issue reference doesn't resolve. This comment (and the one around L1814) cites `(#332)`, but xorbitsai/xagent#332 is unrelated ("feat(sidebar): add collapsible mode and optimize task list layout") — it looks like a PR/merge-commit number was mistakenly carried in as an issue reference. The PR description's "Refs #268" is also wrong: #268 is "support creating agents from prompts", equally unrelated. Please fix both the code comments and the PR description before merge so future readers have a working pointer.

rogercloud (changes requested):
## Summary

This PR adds a durable terminal transition for user-message delivery rows: a new `_reconcile_finalized_turn_delivery` in `src/xagent/web/services/task_orchestrator.py` closes a delivery row (`pending` -> `completed`/`failed`) once a turn's lease is terminally settled in `_schedule_bg`'s `finally` block. It fixes a real, user-visible bug — when the best-effort `pending` -> `dispatched` projection immediately after scheduling fails (pool timeout, transient DB error), the row was previously stuck at `pending` forever, and the downstream WS dedup probe then rejects every same-`client_message_id` retry against that stale row indefinitely. The mechanism correctly reuses the existing monotonic delivery state machine rather than inventing a parallel one. 9 new tests were added; all 67 tests in `tests/web/services/test_task_orchestrator.py` pass.

The mechanism is minimal and correct in the happy path, but the placement and some of its justifying comments deserve a second look:

- **Placement.** The reconciliation logically belongs *inside* `settle_task_lease_isolated`'s existing transaction — which already stages sibling projections such as workforce/trigger run status in the same transaction as the terminal write (`task_orchestrator.py:1268-1287`) — using the existing no-commit primitive `mark_user_message_delivery(db, ...)` (`chat_history_service.py:212`, documented as "stage one monotonic delivery transition without committing db"). Instead this is a third `try`/`except` bolted onto the `finally` block, costing an extra DB pool checkout per turn finalize and relying on a proxy boolean (`lease_settled`) for a fact the callee already knows. Doing it inside the transaction would also have naturally avoided the fenced-settle bug flagged inline below.

- **The "deferred settlement paths are safe to leave `pending`" justification is unsupported in this repo.** The only lease-expiry recovery module, `task_lease_recovery.py`, terminalizes the *task* (PAUSED/FAILED) but never redrives the turn through `_schedule_bg`/`execute_task_background`, and never touches `turn_id`. So for the pool-timeout-in-setup/run deferral path, the delivery row is left `pending` forever — the same failure mode this PR is fixing, just not closed. Partial exception: on a PAUSED recovery, the resume path does appear to re-post the same `turn_id` (`websocket.py:3144`), which gives the lost-lease/heartbeat cases a plausible in-repo consumer; the pool-timeout-on-setup case has none. Please call this out explicitly in the comment rather than asserting the deferred paths are safe.

- **Final delivery state is not a reliable proxy for turn outcome.** On a failed turn the row lands on `failed` or stays `dispatched` forever depending purely on which best-effort projection won a race. Nothing regresses here, but it's worth stating plainly rather than letting the code imply the terminal delivery state reflects the turn's actual result.

- `L760`: the new `except` block (L755-776) duplicates the pool-timeout-vs-other exception logging pattern already at L695-720 in the same file. Extract one shared helper (e.g. `_log_delivery_projection_failure(task_id, component, error, target=None)`) and call it from both sites. (Note: that sibling also omits `turn_id`, so the omission here is pre-existing style, not a new inconsistency.)
- `L3124`: the three new e2e tests (`test_runner_finalize_reconciles_orphaned_pending_delivery`, `test_runner_finalize_leaves_pending_when_deferred_to_ttl_recovery`, `test_runner_finalize_leaves_pending_when_settlement_raises`) each repeat an identical 6-way patch block plus identical task/lease setup, differing only in one patched call's return/side_effect. Factor the shared setup+patch scaffolding into one contextmanager or fixture parameterized by the varying execute/settle behavior.

Net: roughly -30 to -40 lines. Not required for merge.

## Test coverage gaps

Beyond the two gaps called out inline (fenced settle, `SETTLEMENT_READY`), three more paths are genuinely untested in `tests/web/services/test_task_orchestrator.py`:

1. No test drives an `asyncio.CancelledError` through the reconcile call site — the `except asyncio.CancelledError: cleanup_cancellation = cleanup_cancellation or exc` branch guarding the call at L1824 is never exercised.
2. No test drives an actual cancelled turn (`settlement_error="task execution cancelled"`) end-to-end through `_schedule_bg` into reconciliation; only the unit-level function is tested directly with a generic error string.
3. No test seeds an already-terminal delivery row (`completed`/`failed`) and confirms reconciliation is a no-op — the existing "monotonic" test starts from `dispatched`, which is not terminal.

## Tests

All 67 tests in `tests/web/services/test_task_orchestrator.py` pass (9 new + 58 existing), run in a review worktree via `python -m pytest tests/web/services/test_task_orchestrator.py`.

## Also considered and ruled out

Cancellation mapping a turn to `DELIVERY_FAILED` is not a bug — the downstream retry path (`websocket.py` `_enqueue_websocket_task_command_sync` / `finish_existing_delivery`) treats a failed prior delivery as `retry_with_new_id: True`, not a hard rejection. The outer `except Exception` at the call site is real but narrow (it only catches scheduling-layer failures from `run_db_io_cancellation_safe`/`asyncio.to_thread`, since business-logic failures are already swallowed inside the new function) — fine as is.

src/xagent/web/services/task_orchestrator.py
codeacme17: Fixed in a265214a. `lease_settled` is now gated on the returned value — `lease_settled = bool(settled)` — with a comment spelling out both `False`-without-raising cases (fenced out / superseded lease, and a pre-existing terminal/control outcome reconciled instead of this coroutine's view). In both, the coroutine is no longer authoritative and now leaves the delivery row alone.

Added the regression test you suggested: `test_runner_finalize_leaves_pending_when_settle_is_fenced_out` drives `_runner` end-to-end with `settle_task_lease_isolated` stubbed to return `False`, and asserts the row stays `pending`.

Worth noting the second `False` case is also load-bearing: when an inner handler committed PAUSED and settle reconciles it (returning `False` on the error path), the row now stays `pending` — which the paused-resume path needs, since it re-posts the same `turn_id`.

src/xagent/web/services/task_orchestrator.py
codeacme17: Fixed in a265214a. The SETTLEMENT_READY short-circuit now sets a `skip_delivery_reconciliation` flag before returning, so settlement proceeds exactly as before but the delivery row is never closed as `completed` for a turn body that never ran. Regression test added: `test_runner_finalize_skips_reconcile_on_settlement_ready_short_circuit` drives the pre-acquired-lease path with `validate_preacquired_task_lease_isolated` stubbed to SETTLEMENT_READY and asserts the row stays `pending` while the (real) settlement still runs.

On the second suggestion — deferring SETTLEMENT_READY to TTL recovery the way LOST does — I deliberately left that alone: it would change the settlement behavior of a pre-existing branch this PR doesn't otherwise touch, and as you note no in-repo caller currently drives SETTLEMENT_READY at this checkpoint. The targeted skip closes the delivery-side hazard without perturbing lease semantics. Happy to file a follow-up if you think the defensive defer is worth it.

src/xagent/web/services/task_orchestrator.py
codeacme17: Fixed in a265214a. The docstring now scopes the guarantee to "every turn that reaches a terminal lease settlement owned by its coroutine" and enumerates the paths that intentionally leave the row `pending`: lease acquired by another worker, deferred settlement, settle raised, settle fenced out / superseded, the SETTLEMENT_READY short-circuit, and a hard process crash. It also states plainly (per your top-level point) that of those leftovers only the paused-resume path has an in-repo consumer that re-posts the same `turn_id`, while `task_lease_recovery` terminalizes the task without redriving the turn — so the deferral/crash cases remain a narrower instance of the stuck-`pending` window, not a solved one.

src/xagent/web/services/task_orchestrator.py
codeacme17: Fixed in a265214a. Took your simplification suggestion at the same time: both sites now share `_log_delivery_projection_failure(task_id, component=..., error=..., detail=...)`, which uses `logger.error` with the explicit pool-timeout marker for checkout exhaustion and `logger.exception` (traceback preserved) for everything else — so the finalize site is now aligned with the `_schedule_committed_turn` sibling by construction. Existing log-assertion tests keep passing since the message substrings are preserved.

src/xagent/web/services/task_orchestrator.py
codeacme17: Fixed in a265214a. Good catch — both issue numbers are from the xagent-saas tracker, not this repo. The code comments now use the fully-qualified `xorbitsai/xagent-saas#332`, and I've updated the PR description to qualify its references the same way (`xorbitsai/xagent-saas#332` / `xorbitsai/xagent-saas#268`).

codeacme17:
Thanks for the thorough review — the fenced-settle catch in particular was a real authority bug. All findings addressed in a265214a:

**Fixes**
- **Major (fenced settle):** reconciliation is now gated on `bool(settled)`, not on "didn't raise"; regression test added. (inline)
- **SETTLEMENT_READY:** the pre-execution short-circuit now skips delivery reconciliation via an explicit flag; settlement behavior unchanged; regression test added. (inline)
- **Docs:** the docstring no longer overclaims — it scopes the guarantee to coroutine-owned terminal settlements, enumerates every path that intentionally leaves the row `pending`, states that the pool-timeout deferral currently has **no** redrive consumer (`task_lease_recovery` terminalizes the task, never the turn), and says plainly that the final delivery state is not a proxy for the turn's outcome — it answers "may this client_message_id be retried?", never "did the turn succeed?".
- **Refs:** code comments and the PR description now use fully-qualified `xorbitsai/xagent-saas#332` / `#268`.

**Simplifications** (both taken)
- Shared `_log_delivery_projection_failure` helper used by the post-schedule and finalize sites; the non-pool branch is `logger.exception` at both, as you suggested.
- The e2e tests' repeated 6-way patch block and task/lease setup are factored into `_finalize_turn_fixture` + `_finalize_runner_patches` + `_spawn_finalize_runner`.

**Coverage gaps** — all five closed: fenced settle (`False` return), SETTLEMENT_READY short-circuit, a cancelled turn end-to-end (`settlement_error="task execution cancelled"` → row `failed`, task FAILED), `CancelledError` surfacing through the reconcile call site (recorded as cleanup cancellation, re-raised after connector cleanup, row stays `pending`), and already-terminal rows (`completed`/`failed`) as no-ops under both idempotent and conflicting reconciles.

**On placement (the design reservation):** I considered moving the transition into `settle_task_lease_isolated`'s transaction via `mark_user_message_delivery(db, ...)` and decided against it for this PR, for two reasons. First, `settle_task_lease_isolated` has several callers with different delivery semantics — the lease-acquisition cancellation compensator, the schedule-failure compensator in `_schedule_committed_turn` (which already writes its own `failed` marker), and `_runner`'s finalize — so an in-transaction write would close delivery rows for callers that must not (or must differently), unless each one threads `turn_id` plus a policy flag through the shared terminal writer's signature. Second, with the fenced-settle gating fixed, the correctness argument for in-transaction placement is gone; what remains is the extra pool checkout, which is one best-effort write per turn finalize and is deliberately allowed to fail (the row stays reclaimable). If you still prefer the in-transaction shape, I'd rather do it as a follow-up that redesigns `settle_task_lease_isolated`'s contract explicitly than fold it into this fix.

All 74 tests in `tests/web/services/test_task_orchestrator.py` pass (16 for this feature), plus `test_chat_history_service.py`; mypy and ruff clean.

codeacme17:
Pushed 167fbf5f, addressing a major finding cross-posted from the companion SaaS PR review (xorbitsai/xagent-saas#404, finding 2): the `pending → failed` finalize target could invite a double execution. `settlement_error` is set for any non-pool exception out of `execute_task_background` — including failures *after* the agent already ran and produced side effects — while downstream, `failed` is answered with `retry_with_new_id`, so the client would re-send a message that already executed once. The outcome was also asymmetric with the post-schedule projection: had the earlier `dispatched` write landed, the same failure would rest at `dispatched` (`MATCHES`, no retry).

The target is now evidence-gated:

- clean settlement → `completed`;
- error and **execution began** (`execution_started` set immediately before `execute_task_background` is awaited) → `dispatched` — finalize's one certain fact is "no longer in flight"; the failure surfaces via task status, and the row converges on `dispatched` regardless of projection timing;
- error and **execute provably never invoked** → `failed`, where a fresh-id retry is safe. This keeps `failed` aligned with the probe's documented "never dispatched" semantics.

Cancelled-during-execution accordingly now closes as `dispatched` rather than `failed`. Also added the "KNOWN GAP" statement at the reconcile call site (rows skipped by the authority gating stay `pending`; nothing redrives them — tracked in xorbitsai/xagent-saas#409).

Tests: 79 pass in `tests/web/services/test_task_orchestrator.py` (21 for this feature), including new unit + e2e coverage for failure-after-start → `dispatched` and pre-execute setup failure → `failed`; mypy/ruff clean.

rogercloud (approved):
## Summary

Adds a durable terminal transition for user-message delivery rows stuck at `pending`: `_reconcile_finalized_turn_delivery` closes the row (`pending` -> `completed`/`failed`/`dispatched`) once a turn's lease is terminally settled by its own coroutine, wired into `_schedule_bg`'s `finally` block, plus a shared `_log_delivery_projection_failure` helper reused by the post-schedule and finalize projection sites. Fixes xorbitsai/xagent-saas#332: a best-effort `pending -> dispatched` projection failure (pool timeout, transient DB error) previously left the row `pending` forever, permanently wedging the session-WS dedup probe against retries of an already-executed turn.

This is a re-review. All prior findings (5 inline correctness/consistency comments, 2 design-body concerns, 2 simplification suggestions, 3 test-coverage gaps) were independently re-verified against the current code — most via re-running the specific regression test — and confirmed fixed, except one standing design reservation noted below.

## Design verdict: acceptable-with-reservations (non-blocking)

The authoritative-settlement gating is the right invariant and correctly placed relative to the existing `finally` teardown.

- **Placement (carried over from the prior review, still open):** `settle_task_lease_isolated` already opens its own transaction and stages sibling projections (workforce/trigger run status, assistant-message persistence) via non-committing primitives before a single commit. `_reconcile_finalized_turn_delivery` instead runs as a second, separate DB checkout in the `finally` block via `mark_user_message_delivery_sync` (which opens its own session), using the `lease_settled` proxy boolean for a fact `settle_task_lease_isolated` already knows internally. Folding the delivery write into that existing transaction with the non-committing `mark_user_message_delivery(db, ...)` would avoid the extra pool checkout and the proxy boolean. Not a correctness defect — reasonable as a scope-contained tradeoff for a minimal fix — but worth a tracked follow-up.

## New findings (all minor/informational, non-blocking)

1. **Minor** — docstring of `_reconcile_finalized_turn_delivery` calls the transition "monotonic ... terminal," but `dispatched` has an explicit outgoing edge to `completed` in the delivery state machine — it's monotonic (never regresses) but not literally terminal. Wording-only.
2. **Minor** — the `failed` target on the "setup failed before `execute_task_background` was invoked" branch invites a fresh-`client_message_id` retry. Since the delivery row is the same row as the transcript entry and no history-loading path filters on `delivery_status`, a resend after this narrow failure could produce a visible duplicate user message in the transcript/agent context. Narrow (pre-execution only) but real.
3. **Minor** — when `execute_task_background` internally sets `task.status = FAILED` and returns normally (no exception raised), `settlement_error` stays `None`, so the delivery row is written `completed` even though the turn failed. No functional break (the one consumer treats `completed`/`dispatched` identically as "no retry"), but worth noting `delivery_status='completed'` isn't a reliable success signal here — consistent with the docstring's own disclaimer.
4. **Informational** — the residual stuck-`pending` window (deferred-to-TTL-recovery paths) is correlated with the same pool-exhaustion trigger named as this PR's headline bug, since post-schedule dispatch, setup/run, and settle all draw from the same DB pool. Already explicitly disclosed in the docstring's "KNOWN GAP" comment, so not a hidden gap — a tracking issue with a concrete follow-up ID (rather than a bare "tracked as a follow-up to #332") would help close the loop.
5. **Minor (test coverage)** — no e2e test drives a turn to a genuine `TaskStatus.COMPLETED` (via `finish_turn`'s success branch) together with delivery reconciliation; the existing "reconciles orphaned pending on success" e2e test reaches `completed` delivery status via `settlement_error is None`, but the task itself lands on `FAILED` in that test (mocked `execute_task_background` never signals success). The delivery-row assertion is still valid since the delivery target is driven by `settlement_error`, not `task.status`, but a regression in `finish_turn`'s `COMPLETED` branch could go undetected by this suite.

Lean already — no over-engineering found.

## Tests

79 tests in `tests/web/services/test_task_orchestrator.py`, all passing, run against the PR head in an isolated review worktree.

codeacme17:
Thanks for the approval and for re-verifying each prior finding individually. All five new findings landed in 52718335 — docs/test only, no behavior change — and the two design items now have tracked issues.

1. **Docstring "monotonic … terminal"** — corrected. It now says the transition never regresses but only `completed` and `failed` are terminal; `dispatched` keeps its outgoing edge to `completed`, which is precisely how a turn whose projection failed but which later settles cleanly still closes.
2. **`failed` on the pre-execution branch → visible duplicate** — confirmed and documented as an accepted cost, with a tracked fix: **xorbitsai/xagent-saas#417**. I verified `load_task_transcript` filters only on `task_id`/`before_message_id` and that no read path anywhere excludes `delivery_status`. Worth noting the condition is **pre-existing rather than new**: the schedule-failure compensator in `_schedule_committed_turn` and several `DELIVERY_FAILED` writers in `websocket.py` already produce visible `failed` rows today — this PR adds one narrow additional path. I kept `failed` here because the alternative, leaving the row `pending`, wedges that `client_message_id` forever (functional break) versus a cosmetic duplicate plus context pollution; and the real fix — excluding `failed` rows from history loading while keeping legacy `NULL` rows visible — belongs with transcript semantics and fixes the pre-existing writers at the same time.
3. **`completed` isn't a success signal** — named explicitly in the docstring: `execute_task_background` may set `task.status = FAILED` and return normally, leaving `settlement_error` unset. Also noted why it's harmless: `completed` and `dispatched` read identically downstream.
4. **Residual window needs a concrete ID** — the comment now cites **xorbitsai/xagent-saas#409** and adopts your correlation point: post-schedule dispatch, setup/run and settle all draw from the same DB pool, so a single exhaustion event can both orphan the row and defer the settlement that would have closed it.
5. **Missing genuine-COMPLETED e2e** — added `test_runner_finalize_reconciles_alongside_a_genuinely_completed_turn`. The fake runtime commits `COMPLETED` plus an assistant message (what `finish_turn`'s COMPLETED branch reads to write `output`), and the test asserts task status, `output`, lease release, and the delivery row together — so a regression in that branch now fails here.

**Placement reservation** — filed as **xorbitsai/xagent-saas#418** with the concrete blocker written down: `settle_task_lease_isolated` has callers with differing delivery semantics (lease-acquisition cancellation compensator, schedule-failure compensator that already writes its own `failed`, and `_runner`'s finalize), so folding the write in means redesigning its contract to thread the turn id plus a per-caller policy including the `execution_started` evidence rule — a deliberate interface change rather than a drive-by.

80 tests in `tests/web/services/test_task_orchestrator.py` pass; mypy and ruff clean.

Issue #960, opened 2026-07-22
refactor(api): standardize backend error response contracts


AlexLiu190625:
## Context

Backend error responses currently use multiple shapes across API surfaces and failure paths:

- `/api/*` commonly uses FastAPI's `{"detail": "..."}` envelope.
- Some domain failures use a structured `detail` object such as `{"detail": {"code": "...", "message": "..."}}`.
- Request validation uses `{"detail": [...]}`.
- `/v1/*` has a separate stable `{"error": {"code": "...", "message": "..."}}` contract.
- Some clients still parse top-level `message` values.

This forces frontend and SDK consumers to implement endpoint-specific parsing and makes it easy to pass non-renderable values into presentation components. Issue #956 applies a targeted Agent Builder safeguard; it does not establish the backend-wide contract.

## Goal

Agree on a stable, typed backend error contract and a compatible migration strategy for backend producers and all supported clients.

## Discussion points

- The canonical envelope and field semantics, including stable machine-readable codes, human-readable messages, and optional structured details.
- Whether `/api/*` should converge with `/v1/*` or retain a versioned compatibility envelope.
- How FastAPI request validation errors map into the canonical contract.
- Which backend exception/response layer owns normalization.
- Backward compatibility and rollout sequencing for the web frontend, SDKs, integrations, and external callers.
- OpenAPI/schema generation and typed client support.
- Logging, correlation IDs, and rules for preventing sensitive internal details from reaching clients.

## Required audit before implementation

- Inventory error producers across routes, services, exception handlers, and middleware.
- Inventory consumers across the frontend, SDKs, integrations, and tests.
- Document the current path matrix for `/api/*`, `/v1/*`, A2A, WebSocket, validation, authentication, authorization, domain conflicts, and unexpected failures.
- Identify compatibility requirements and endpoints that cannot migrate atomically.

## Decision outputs / acceptance criteria

- [ ] An agreed canonical error schema with examples for validation, authentication, authorization, not-found, conflict, rate/quota, domain, and internal errors.
- [ ] A named backend ownership boundary for producing the schema.
- [ ] A compatibility and versioning policy for existing clients.
- [ ] A staged backend and client migration plan.
- [ ] Contract tests for each API surface and typed consumer tests.
- [ ] Documentation of observability and sensitive-detail handling.

## Non-goals

- This discussion issue does not replace targeted user-facing safeguards such as #956.
- No broad contract migration should be folded into #956.

PR #1126, opened 2026-08-04
fix: safely render errors from non-update builder actions

### LINKED ISSUE #969: fix(agent-builder): safely render errors from non-update actions
## Summary

Several non-update actions in the Agent Builder still pass `error.detail` directly to `toast.error`. If an endpoint returns a structured FastAPI `detail` object or validation array, the toaster can receive a non-renderable value and crash the page.

## Affected actions

- Publish
- Unpublish
- Publish from the creation success dialog
- Optimize instructions

## Current behavior

Each affected handler assumes a valid JSON response and uses `error.detail || fallback`. Publish and unpublish currently produce mostly string details, but the consumer is coupled to that shape. Request validation for optimize instructions can return a FastAPI `detail` array before the route runs.

## Expected behavior

- Failed responses are parsed without assuming valid JSON or a fixed payload shape.
- Only displayable strings are passed to `toast.error`; objects and arrays are never passed through directly.
- Unknown, malformed, non-JSON, or empty responses use the action-specific localized fallback.
- The Agent Builder remains mounted after each failed action.

## Scope

This is a page-level presentation safeguard for the affected Agent Builder actions. It does not define or migrate the backend-wide error response contract; that work remains tracked by #960.

## Acceptance criteria

- [ ] Route all four affected handlers through one page-owned safe error-message path instead of duplicating shape checks.
- [ ] Preserve each action-specific localized fallback.
- [ ] Add component regression coverage for structured `detail` and FastAPI validation-array responses on reachable affected actions.
- [ ] Add coverage for malformed or empty responses.
- [ ] Keep backend response contracts unchanged.

Follow-up to #961. Related contract discussion: #960.

AlexLiu190625:
## Summary

- route the four non-update Agent Builder actions (publish, unpublish, publish from the creation success dialog, optimize instructions) through the builder's single safe error-message path instead of passing `error.detail` straight to the toaster
- a structured FastAPI `detail` object or validation array can no longer crash the page; only displayable strings reach `toast.error`, with objects and arrays reduced to their readable messages
- parse failed response bodies without assuming valid JSON: each `response.json()` gets `.catch(() => null)`, matching the existing update path

The safe path is the helper PR #961 introduced for the update action (`getAgentUpdateErrorMessage`), renamed to `getBuilderErrorMessage` now that it serves the update path and all four non-update actions. The page's `throw new Error(...detail...)` branches in the ownership/save flow render degraded copy at worst and stay with #960 (see Scope).

Fixes #969. Related to #960, which tracks the backend error-contract standardization and the shared frontend error helper; this PR keeps backend contracts unchanged.

## Behavior change

Empty and non-JSON error bodies previously threw inside the handler and surfaced the generic `builds.editor.error.unknown` toast. They now fall back to each action's own localized message (`publishFailed` / `unpublishFailed` / `optimizeError`), per the issue's acceptance criteria. No test pinned the old behavior. Transport-level failures (fetch rejection) still reach the outer catch and the generic message; a regression test covers that boundary.

A top-level `message` field in an error response body is now surfaced for publish, unpublish and optimize instructions; previously these actions inspected only `detail`, so such a body fell back to the generic localized message. An object `detail` carrying `msg` is also read now (same precedence as the validation-array branch: `msg` before `message`, `detail` before the top-level field), and both paths are pinned by fixtures.

## Scope

The issue names exactly these four actions. Remaining unsafe `detail` consumers are intentionally untouched and belong to #960:

- same crash class (`toast.error(x.detail || ...)`): `app/tools/page.tsx` (5 sites), `components/mcp/connect-mcp-dialog.tsx` (2 sites)
- degraded-copy class (`throw new Error(x.detail ...)`, renders `[object Object]` but cannot crash): sites in `agent-builder.tsx` (update/ownership branches), `channels/page.tsx`, `public-agent-chat-page.tsx`, `knowledge-base-detail.tsx`, `workforce-edit-dialogs.tsx`, `deploy-agent-dialog.tsx`, `user-management.tsx`, `models.tsx`, `model-management-dialog.tsx`, `lib/agent-widget-config.ts`, `lib/models.ts`

`app/build/page.tsx`'s publish/unpublish path never reads `detail` (it renders fixed localized keys), so it needs no change.

## Validation

- new `agent-builder-publish-errors.test.tsx`: 41 tests — 4 actions x 10 payload shapes (string detail, structured detail via `message` or `msg`, detail-over-top-level precedence, unreadable object, validation array, unreadable array, bare top-level message, empty body, non-JSON 502 body) plus a fetch-rejection case; the original 7-shape matrix fails 26/28 against the previous code, and each case asserts exactly one toast call
- mutation checks: reverting any single call site to `error.detail ||` fails its 7 cases; removing `.catch(() => null)` from publish fails the empty/non-JSON cases; swapping an action's fallback key fails 4 cases. On the optimize action, removing `.catch(() => null)` is behavior-neutral (its outer catch uses the same `optimizeError` key), so that single mutation survives by design rather than by a coverage gap
- `tsc --noEmit` passes; the new test file is ESLint-clean (the repo's lint baseline has pre-existing errors elsewhere in `agent-builder.tsx`, untouched here); `src/i18n/translations.test.ts` passes (it pins the source-level usage counts of the publication keys)
- all four fallback keys verified present in both `en` and `zh` translations
- the pre-existing agent-builder test files have a known local-only failure on current main that is unrelated to this change; the new suite is self-contained and green
- note: CI's frontend lanes do not currently run `src/components/build/` tests (a pre-existing gap that also covers the other agent-builder suites), so this suite guards regressions in local runs until that lane is widened

### DIFF
--- a/frontend/src/components/build/agent-builder-publish-errors.test.tsx
+++ b/frontend/src/components/build/agent-builder-publish-errors.test.tsx
@@ -4,9 +4,9 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 
 // Issue #969: the non-update builder actions (publish, unpublish, publish from
 // the creation success dialog, optimize instructions) must never pass a raw
-// `detail` payload to toast.error. Each action renders only displayable
-// strings, keeps its own localized fallback for unreadable or malformed
-// responses, and leaves the builder mounted after the failure.
+// `detail` payload to toast.error. sonner is mocked here, so each case asserts
+// two things: the exact string handed to the toaster is displayable, and the
+// builder is still mounted afterwards.
 
 const apiRequestMock = vi.hoisted(() => vi.fn())
 const toastErrorMock = vi.hoisted(() => vi.fn())
@@ -114,7 +114,15 @@ vi.mock("@/hooks/use-file-mention", () => ({
 vi.mock("@/components/ui/multi-select", () => ({
   MultiSelect: () => <div data-testid="multi-select" />,
 }))
-vi.mock("@/components/ui/select", () => ({ Select: () => null }))
+// The model Select is the only place `modelConfig.general` reaches the DOM
+// (agent-builder.tsx renders it at the `models.length > 0` branch of the config
+// form). Mirroring the value onto data-value gives the create flow a readiness
+// signal it can wait on instead of retrying clicks.
+vi.mock("@/components/ui/select", () => ({
+  Select: ({ value }: { value?: string }) => (
+    <div data-testid="model-select" data-value={value ?? ""} />
+  ),
+}))
 vi.mock("@/components/build/build-file-preview-sheet", () => ({
   BuildFilePreviewSheet: () => null,
 }))
@@ -154,7 +162,17 @@ function agentResponse(status: "draft" | "published") {
 // instead of handing the raw payload to the toaster.
 const FALLBACK = "__ACTION_FALLBACK__"
 
-const ERROR_CASES = [
+type ErrorCase = {
+  name: string
+  body: string | null
+  expected: string
+  // Default to a FastAPI validation failure; a case overrides these when the
+  // transport shape itself is part of the fixture.
+  status?: number
+  contentType?: string
+}
+
+const ERROR_CASES: ErrorCase[] = [
   {
     name: "a plain string detail",
     body: JSON.stringify({ detail: " Action failed with string detail " }),
@@ -164,6 +182,19 @@ const ERROR_CASES = [
     name: "a structured detail message",
     body: JSON.stringify({ detail: { message: "Action failed", context: [] } }),
     expected: "Action failed",
+  },
+  {
+    name: "a structured detail msg",
+    body: JSON.stringify({ detail: { msg: "  Action failed with detail msg  " } }),
+    expected: "Action failed with detail msg",
+  },
+  {
+    name: "a detail msg alongside a top-level message",
+    body: JSON.stringify({
+      detail: { msg: "Detail wins" },
+      message: "Top-level loses",
+    }),
+    expected: "Detail wins",
   },
   {
     name: "a detail object without a readable message",
@@ -188,6 +219,11 @@ const ERROR_CASES = [
     expected: FALLBACK,
   },
   {
+    name: "a bare top-level message",
+    body: JSON.stringify({ message: "  Top-level failure  " }),
+    expected: "Top-level failure",
+  },
+  {
     name: "an empty response body",
     body: null,
     expected: FALLBACK,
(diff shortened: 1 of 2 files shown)


frontend/src/components/build/agent-builder.tsx
```diff
@@ -152,7 +152,11 @@ function readNonEmptyString(value: unknown): string | null {
   return typeof value === "string" && value.trim() ? value.trim() : null
 }
 
-function getAgentUpdateErrorMessage(error: unknown, fallback: string): string {
+// Safe error-message path for the builder's response-body error branches:
+// only displayable strings may reach toast.error; objects and arrays are
+// reduced to their readable messages and anything else falls back to the
+// caller's localized message.
+function getBuilderErrorMessage(error: unknown, fallback: string): string {
```
rogercloud: **Minor (naming/doc nit):** the parameter is named `error` but always receives a parsed response *body*, not an exception — `body` or `payload` would read truer at the call sites. The JSDoc above also describes what callers must guarantee rather than this function's own contract.

Separately, worth noting since this function moved here: the object-`detail` branch a few lines down only reads `detail.message`, while the array-item branch reads `item.msg ?? item.message`. A bare `{"detail": {"msg": "..."}}` would fall through to the fallback even though the same shape inside an array is handled. I checked the backend (`src/xagent/web/app.py`) — these builder endpoints only ever emit `detail` as an array for validation errors, never a bare object with `msg`, so this is currently unreachable. Still, `readNonEmptyString(detail.msg) ?? readNonEmptyString(detail.message)` would make the two branches symmetric and not drift if that ever changes. Both are pre-existing (from #961), not introduced here — flagging only because the function is being generalized to more call sites in this PR.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1602,8 +1606,10 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
         })
         toast.success(t("builds.editor.success.published"))
       } else {
-        const error = await response.json()
-        toast.error(error.detail || t("builds.publication.publishFailed"))
+        const error: unknown = await response.json().catch(() => null)
+        toast.error(
+          getBuilderErrorMessage(error, t("builds.publication.publishFailed"))
```
rogercloud, line 1613: **Minor (slightly more substantive than a nit): this widens what gets surfaced for publish/unpublish/optimize.** Before this PR these three handlers inspected only `error.detail`. Routing them through the shared helper means a bare top-level `message` field on the response body is now displayed to the user for these actions too (the helper's last fallback, `readNonEmptyString(error.message)`, is pre-existing from #961 but wasn't reachable from here before). I don't think this is a real risk — same trust boundary, on an already-failing request, in an error toast rather than sensitive output — but it is a behavior change beyond what the PR description states, worth calling out explicitly.

It's also the one return path with zero test coverage: none of the seven `ERROR_CASES` fixtures is a bare `{"message": "..."}` body. Worth adding one case so the matrix exercises every return path (same applies to the unpublish and optimize call sites below).

frontend/src/components/build/agent-builder-publish-errors.test.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+vi.mock("@/contexts/mcp-apps-context", () => ({
+  useMcpApps: () => ({ apps: [], getAppIcon: () => null }),
+}))
+
+vi.mock("@/lib/branding", () => ({
+  getBrandingFromEnv: () => ({ appName: "Xagent" }),
+}))
+
+vi.mock("sonner", () => ({ toast: { error: toastErrorMock, success: vi.fn() } }))
```
rogercloud, line 79: **Minor: the test names claim more than the tests verify.** `sonner` is fully mocked here, so these tests structurally cannot reproduce the crash mechanism from #969 — rendering a non-string React child inside the toast. The real assertion is a string-equality check on the argument passed to the mocked `toast.error`, which proves "the app only ever calls `toast.error` with a string" (necessary) but never proves the old code would have crashed or that the new code survives a real toast render (sufficient). Test names like "... without unmounting the builder" overstate that.

Two options: rename to something like "renders a displayable message", or keep the names and add one test that mounts a real `<Toaster />` and asserts the message renders — that would close the gap end-to-end.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
```diff
@@ -0,0 +1,412 @@
+import React from "react"
```
rogercloud, line 1: **Minor (structural, more substantive than a nit): the helper being unexported forces 28 full-component renders where a unit matrix would do.** The 7-shape × 4-action matrix is exercised entirely through full `AgentBuilder` renders, each paying the whole mocked API bootstrap, to test what is fundamentally a pure `(unknown, string) => string` function. Exporting `getBuilderErrorMessage` (or moving it to `api-wrapper.ts` per the design note in the review body) would let the 7 payload shapes become a fast unit table, leaving 1-2 wiring tests per action to prove each handler actually calls the helper with the right fallback. Same coverage, a fraction of the runtime, and failures point at the parser rather than at a render.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
```diff
@@ -0,0 +1,412 @@
...
+    expected: FALLBACK,
+  },
+  {
+    name: "a non-JSON response body",
+    body: "<html>Bad Gateway</html>",
+    expected: FALLBACK,
+  },
+] as const
+
+function errorResponse(body: string | null) {
```
rogercloud: **Nit (fixture realism):** the non-JSON case uses HTTP 422 with an `application/json` content-type, which is a shape the server wouldn't actually produce for a non-JSON body. The code under test never branches on status or content-type — it only checks `response.ok` and relies on `.json().catch()` — so this doesn't weaken what's being tested. A 502 with `text/html` would just read more like the real failure it stands in for.

One combined aside while I'm here: neither the missing length cap on the joined array message in `getBuilderErrorMessage` (cf. `truncateUploadMessage` in `frontend/src/lib/api-wrapper.ts:186`) nor the un-restored `globalThis.WebSocket = vi.fn()` in `beforeEach` is worth acting on — the four real call sites can't produce a long message, and sibling tests in this directory already split roughly 50/50 on restoring the WebSocket stub.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+async function waitForLoadedBuilder() {
+  await waitFor(() =>
+    expect(
+      screen.getByPlaceholderText("builds.configForm.name.placeholder")
+    ).toHaveValue("Existing Agent")
+  )
+}
+
+async function expectToast(expected: string) {
```
rogercloud, line 341: **Nit: `expectToast` asserts only the last `toast.error` call, not the call count.** Reading all four handlers, each can structurally fire `toast.error` at most once (mutually exclusive if/else plus the catch), so this cannot mask a real bug today. Adding `expect(toastErrorMock).toHaveBeenCalledTimes(1)` would make the helper robust against a future handler that double-fires. Nice-to-have.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
```diff
@@ -0,0 +1,412 @@
...
+      await expectToast(
+        expected === FALLBACK ? "builds.publication.publishFailed" : expected
+      )
+      expect(screen.getByDisplayValue("Existing Agent")).toBeInTheDocument()
+    }
+  )
+})
+
+describe("AgentBuilder network failure handling (issue #969)", () => {
+  it("uses the generic fallback when the publish request itself rejects", async () => {
```
rogercloud, line 405: **Informational, no change requested.** Good that the comment above records this as deliberate. For the record: the publish/unpublish transport-failure catch blocks fall back to the generic `builds.editor.error.unknown`, whereas optimize-instructions uses its own `optimizeError` fallback in both branches. Verified this inconsistency predates the PR (already present at the base commit) and that the PR doesn't touch these catch blocks, so locking in current behavior here is correct. Just worth knowing this test now pins a pre-existing quirk — if the fallbacks are ever unified, this assertion is what needs updating.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
```diff
@@ -0,0 +1,412 @@
...
+      const editor = document.querySelector("[contenteditable]") as HTMLElement
+      expect(editor).toBeTruthy()
+      editor.textContent = "You are a new agent."
+      fireEvent.input(editor)
+
+      // The default model arrives asynchronously from /api/models/user-default;
+      // retry the create click until validation passes and the dialog opens.
+      // After creation the header shows its own (disabled) publish button, so
+      // scope the click to the success dialog.
+      await waitFor(() => {
```
rogercloud: **Minor (style / future-proofing): a mutating `fireEvent.click` inside a `waitFor` retry callback.** `waitFor` re-invokes its callback until it stops throwing, so the click can fire many times. In practice this is harmless today — the create button is disabled both while the model loads and while creation is in flight, so extra clicks are no-ops under jsdom — but it's fragile if either guard is ever relaxed.

Cleaner: wait on an explicit readiness signal (e.g. the button becoming enabled) outside the click, then do the single click once.

rogercloud (approved):
## Summary

This PR routes four Agent Builder actions — publish, unpublish, publish-from-creation-dialog, and optimize-instructions — through the same error-message helper that the update action already used, renaming it `getAgentUpdateErrorMessage` → `getBuilderErrorMessage`. The helper narrows an arbitrary parsed JSON error body down to a plain `string` (string `detail`, object `detail.message`, array of `{msg|message}` items joined, top-level `message`, else the caller's localized fallback), so a structured FastAPI `detail` object or validation-error array can no longer be handed to `toast.error` and crash the page with "Objects are not valid as a React child" (#969). Each call site additionally guards `response.json()` with `.catch(() => null)`, so a non-JSON or empty body degrades to the action's own localized message instead of throwing. A new test file adds 29 tests covering a 7-payload-shape × 4-action matrix plus transport-failure cases.

## Verdict

**Approve with minor suggestions.** The approach is sound: the helper's return type is statically `string`, which makes it structurally impossible for a non-string to reach `toast.error` — that satisfies #969's acceptance criteria directly rather than patching each call site defensively. The fix is minimal, well-scoped, and well-tested. Everything below is minor or nit-level; none of it blocks merge.

## Design-level notes

**Duplicate error-parsing logic is now four copies deep.** `getBuilderErrorMessage` (this is actually pre-existing logic from PR #961, only renamed/relocated here) joins `getApiErrorMessage` (`frontend/src/lib/api-wrapper.ts:205-210`) and a byte-identical `formatApiDetail` pair (`frontend/src/lib/workforces-api.ts:52-69`, `frontend/src/lib/agent-triggers-api.ts:186-203`). These have already diverged in observable behavior: an array item shaped `{message: "x"}` with no `msg` key is silently dropped by the `formatApiDetail` pair (falling through to a generic fallback) but is surfaced by this helper. The same backend validation-error payload therefore produces different user-visible text depending on which module handled the request. Reusing the existing helper for these four actions was the right call; the reservation is only about *where* the shared logic lives long-term. Consider promoting it to `api-wrapper.ts` in a follow-up so the other three copies can delegate to it. Not a blocker here.

**Remaining unsafe `detail` sites in the same file.** Five other extraction sites in `agent-builder.tsx` (lines 1315, 1319, 1332, 1354, 1364, in the ownership/save flow) still read `detail` without the safety net. Confirmed unmodified by this PR and explicitly scoped out in the PR description, tracked under #960. Reasonable boundary — just flagging that the file is only partly safe after this lands.

**Call-site boilerplate.** The `const body = await response.json().catch(() => null)` + `toast.error(getBuilderErrorMessage(body, fallback))` pair now repeats five times (~lines 1580, 1611, 1641, 1672, 1721). A small `toastResponseError(response, fallback)` wrapper would collapse all five and make it harder for a future call site to forget the `.catch`. Optional.

## Test run

`npx vitest run src/components/build/agent-builder-publish-errors.test.tsx` → **29/29 passed**, no `act()` warnings and no unhandled rejections. (One pre-existing unrelated stub warning, and one intentional `console.error` from the deliberate network-failure case.)

- `L214-269: [dedupe] installEditModeApi / installCreateModeApi share seven identical mock branches (kb/collections, skills, tools/available, models/?category=llm, triggers, mcp/servers, catch-all) — roughly 60% overlap. Extract a shared installBaseApiMocks(overrides) and have both call it with only their divergent branches.`
- `L272-411: [table-drive] The publish, unpublish and optimize-instructions describe blocks repeat the same per-case scaffold. Collapse the three into one {status, failingPath, clickText, fallbackKey} × ERROR_CASES runner. Leave the success-dialog publish block as its own describe — its setup, render, click scope and survival assertion all diverge, and folding it in would need several extra optional hooks that cost more readability than the lines saved.`

`net: -35 to -40 lines possible`

frontend/src/components/build/agent-builder.tsx
AlexLiu190625: Renamed the parameter to `body` and rewrote the comment to state the function's own contract. The object-`detail` branch now reads `msg` with the same precedence as the array branch (`msg` before `message`), and two new fixtures pin the shape and the precedence: `{detail:{msg}}` surfaces the msg, and `{detail:{msg}, message}` prefers the detail over the top-level field.

frontend/src/components/build/agent-builder.tsx
AlexLiu190625: Added a bare top-level `{"message": ...}` fixture to the matrix (exercised by all four actions), and the PR description's Behavior change section now states the widened surface explicitly.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
AlexLiu190625: Went with the rename: cases are now titled "surfaces a displayable message for ...", and the file header states that sonner is mocked and what each case asserts (the exact string handed to the toaster, plus the builder staying mounted). The structural guarantee that a non-string cannot reach the toaster remains the helper's `string` return type.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
AlexLiu190625: Keeping the helper unexported preserves the issue's page-owned acceptance criterion, so the unit-table split moves with #960, which owns promoting it to the shared module. In the meantime the three edit-mode suites are table-driven over a shared mock base, which contains the per-case render cost.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
AlexLiu190625: The non-JSON fixture is now a 502 with `text/html`.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
AlexLiu190625: `expectToast` now asserts exactly one `toast.error` call, inside the same `waitFor` as the string equality, across all four suites. The create-dialog retry loop that could have leaked a validation toast is gone (replaced by the readiness gate below), so the count holds globally.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
AlexLiu190625: Acknowledged — the assertion pins the current fallback split; if the transport-failure fallbacks are ever unified, this is the test to update.

frontend/src/components/build/agent-builder-publish-errors.test.tsx
AlexLiu190625: Replaced the retry loop: the `Select` mock now mirrors `modelConfig.general` into a `data-value` attribute, the test waits for the model id to land in the DOM, then clicks create exactly once and awaits the dialog.

AlexLiu190625:
On the review-body notes: extracted a shared `installBaseApiMocks(overrides)` and table-drove the three edit-mode suites (the success-dialog suite stays separate, as suggested). Promoting the helper to `api-wrapper.ts` and consolidating the four parser copies is deferred to #960 together with the remaining unsafe `detail` sites listed in the description's Scope section.

PR #413, opened 2026-05-17
fix: stabilize qemu backend docker build

qinxuye:
## Summary
- install CA/SSL runtime packages in the backend Docker build stages
- verify Python ssl support before pip accesses HTTPS package indexes
- keep torch CPU install behavior unchanged while making QEMU arm64 failures explicit

## Verification
- Triggered the Publish Docker Images workflow manually on this branch with a temporary build-only backend multi-arch check; it completed successfully before the temporary workflow changes were removed from the final commit.

PR #1130, opened 2026-08-04
feat: fall back to the channel's configured default agent

OliverBryant:
## Summary

When a transport passes no explicit agent selection, `prepare_channel_task` now falls back to the channel's `config["default_agent_id"]`. Slack and Feishu have no agent picker, so today every conversation on those channels runs the generic default agent with no way to pin a Builder agent; Telegram keeps its explicit `/agents` selection, which takes precedence over the config default.

- `ChannelOwnerSnapshot` gains `default_agent_id`, populated from the channel config with best-effort int coercion — config is free-form JSON, so a malformed value degrades to "no default" instead of failing every turn on the channel
- a stale default (agent deleted / not owner-visible) flows through the existing requested-agent-missing path: the turn falls back to a clean default task, never errors
- generic capability, no product coupling: any caller can set the key via the existing `PUT /api/channels/{id}` (it is not one of the Slack server-managed identity keys)

## Why

The immediate consumer is workspace provisioning in xagent-saas (xorbitsai/xagent-saas#385, #413): after onboarding, the provisioned Builder agent is pinned as the Slack channel's default so DMs/@mentions run it. The mechanism (`prepare_channel_task(agent_id=...)`, owner-visibility validation, task eviction on agent change) already existed — this only adds the config-sourced fallback.

## Validation

- `pytest tests/web/services/test_channel_runtime.py` — 30 passed (7 new: config-default binding, explicit-selection precedence, malformed-value degradation ×4, stale-default fallback)
- ruff check / format clean; mypy on the touched module

### DIFF
--- a/src/xagent/web/services/channel_runtime.py
+++ b/src/xagent/web/services/channel_runtime.py
@@ -64,6 +64,10 @@ class ChannelOwnerSnapshot:
     """Detached owner identity after channel authorization."""
 
     user_id: int
+    # Optional per-channel default Agent Builder agent (``config`` key
+    # ``default_agent_id``): used by ``prepare_channel_task`` when the
+    # transport passes no explicit agent selection.
+    default_agent_id: int | None = None
 
 
 @dataclass(frozen=True)
@@ -303,7 +307,25 @@ def _load_channel_owner_sync(
         allowed_user_ids = {str(value) for value in allowed_users}
         if external_user_id not in allowed_user_ids:
             raise ChannelAuthorizationError("Channel sender is not authorized")
-    return ChannelOwnerSnapshot(user_id=owner_id)
+    return ChannelOwnerSnapshot(
+        user_id=owner_id,
+        default_agent_id=_coerce_agent_id(channel.config.get("default_agent_id")),
+    )
+
+
+def _coerce_agent_id(value: object) -> int | None:
+    """Best-effort int coercion for the free-form config value.
+
+    ``config`` is an unconstrained JSON dict, so the id may arrive as a
+    string or garbage; a malformed value must degrade to "no default"
+    rather than fail every turn on the channel.
+    """
+    if value is None or isinstance(value, bool):
+        return None
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return None
 
 
 def _claim_legacy_active_telegram_task_sync(
@@ -635,6 +657,13 @@ def _prepare_channel_task_sync(
                 raise ChannelAuthorizationError(
                     "Channel owner changed during sender authorization"
                 )
+            explicit_agent_requested = agent_id is not None
+            if agent_id is None:
+                # No explicit selection from the transport (Slack has no agent
+                # picker; Telegram only passes one after /agents): fall back to
+                # the channel's configured default. A stale id degrades through
+                # the missing-agent path below, never fails the turn.
+                agent_id = owner.default_agent_id
 
             is_telegram = str(channel.channel_type) == "telegram"
             if is_telegram:
@@ -690,7 +719,18 @@ def _prepare_channel_task_sync(
                     .filter(Agent.id == int(agent_id))
                     .first()
                 )
-                requested_agent_missing = agent_row is None
+                # The missing-agent flag drives transport UX for a selection
+                # the USER made (Telegram clears the stored pick and says so).
+                # A stale channel-config default is an operator concern, not a
+                # user action — log it instead of blaming the user's choice.
+                requested_agent_missing = agent_row is None and explicit_agent_requested
+                if agent_row is None and not explicit_agent_requested:
+                    logger.warning(
+                        "Channel %s default_agent_id %s is not selectable; "
+                        "falling back to the default assistant",
+                        channel_id,
+                        agent_id,
+                    )
                 if task is not None:
                     if agent_row is None:
                         # Evict to a clean default task; never fail the turn.
--- a/src/xagent/web/services/channel_runtime.py
+++ b/src/xagent/web/services/channel_runtime.py
@@ -314,18 +314,24 @@ def _load_channel_owner_sync(
 
 
 def _coerce_agent_id(value: object) -> int | None:
-    """Best-effort int coercion for the free-form config value.
+    """Strict int coercion for the free-form config value.
 
     ``config`` is an unconstrained JSON dict, so the id may arrive as a
     string or garbage; a malformed value must degrade to "no default"
-    rather than fail every turn on the channel.
+    rather than fail every turn on the channel. Only true integers and
+    integer strings are accepted — ``int(12.5)`` would silently truncate
+    to a valid but WRONG agent id, which is worse than no default.
     """
-    if value is None or isinstance(value, bool):
+    if isinstance(value, bool):
         return None
-    try:
-        return int(value)
-    except (TypeError, ValueError):
-        return None
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str):
+        try:
+            return int(value.strip())
+        except ValueError:
+            return None
+    return None
 
 
 def _claim_legacy_active_telegram_task_sync(
--- a/tests/web/services/test_channel_runtime.py
+++ b/tests/web/services/test_channel_runtime.py
@@ -115,6 +115,191 @@ async def test_prepare_channel_task_binds_owned_agent(
     engine.dispose()
 
 
+def _set_channel_config_value(SessionLocal, channel_id: int, key: str, value) -> None:
+    with SessionLocal() as db:
+        channel = db.query(UserChannel).filter(UserChannel.id == channel_id).one()
+        channel.config = {**channel.config, key: value}
+        db.commit()
+
+
+@pytest.mark.asyncio
+async def test_prepare_channel_task_uses_channel_default_agent(
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: Path,
+    mock_workspace_db,
+) -> None:
+    """A transport that passes no agent selection binds the channel's
+    configured default agent (config key ``default_agent_id``)."""
+    del mock_workspace_db
+    engine, SessionLocal, user_id, channel_id = _create_channel_session_local(tmp_path)
+    agent_id = _add_agent(SessionLocal, user_id=user_id, name="Toby")
+    _set_channel_config_value(SessionLocal, channel_id, "default_agent_id", agent_id)
+    monkeypatch.setattr(
+        "xagent.web.services.channel_runtime.get_session_local",
+        lambda: SessionLocal,
+    )
+    monkeypatch.setattr(database_module, "get_session_local", lambda: SessionLocal)
+
+    prepared = await prepare_channel_task(
+        channel_id=channel_id,
+        external_user_id="telegram-user",
+        active_task_id=None,
+        text="hello",
+        channel_name="Telegram",
+    )
+
+    assert prepared is not None
+    assert prepared.requested_agent_missing is False
+    with SessionLocal() as db:
+        task = db.query(Task).filter(Task.id == prepared.task_id).one()
+        assert task.agent_id == agent_id
+
+    assert await prepared.managed_lease.finalize_result(status=TaskStatus.FAILED)
+    engine.dispose()
+
+
+@pytest.mark.asyncio
+async def test_explicit_agent_selection_wins_over_channel_default(
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: Path,
+    mock_workspace_db,
+) -> None:
+    del mock_workspace_db
+    engine, SessionLocal, user_id, channel_id = _create_channel_session_local(tmp_path)
+    default_agent_id = _add_agent(SessionLocal, user_id=user_id, name="Toby")
+    explicit_agent_id = _add_agent(SessionLocal, user_id=user_id, name="Research Agent")
+    _set_channel_config_value(
+        SessionLocal, channel_id, "default_agent_id", default_agent_id
+    )
+    monkeypatch.setattr(
+        "xagent.web.services.channel_runtime.get_session_local",
+        lambda: SessionLocal,
+    )
+    monkeypatch.setattr(database_module, "get_session_local", lambda: SessionLocal)
+
+    prepared = await prepare_channel_task(
+        channel_id=channel_id,
+        external_user_id="telegram-user",
+        active_task_id=None,
+        text="hello",
+        channel_name="Telegram",
+        agent_id=explicit_agent_id,
+    )
+
+    assert prepared is not None
+    with SessionLocal() as db:
+        task = db.query(Task).filter(Task.id == prepared.task_id).one()
+        assert task.agent_id == explicit_agent_id
+
+    assert await prepared.managed_lease.finalize_result(status=TaskStatus.FAILED)
+    engine.dispose()
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


rogercloud:
ci is failing

src/xagent/web/services/channel_runtime.py
```diff
@@ -303,7 +307,25 @@ def _load_channel_owner_sync(
...
+    ``config`` is an unconstrained JSON dict, so the id may arrive as a
+    string or garbage; a malformed value must degrade to "no default"
+    rather than fail every turn on the channel.
+    """
+    if value is None or isinstance(value, bool):
+        return None
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return None
```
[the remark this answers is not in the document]
OliverBryant: Fixed in 9fa1b05c — `_coerce_agent_id` now accepts only true ints (bool excluded) and integer strings; floats fall through to `None`.

tests/web/services/test_channel_runtime.py
```diff
@@ -115,6 +115,191 @@ async def test_prepare_channel_task_binds_owned_agent(
...
+    with SessionLocal() as db:
+        task = db.query(Task).filter(Task.id == prepared.task_id).one()
+        assert task.agent_id == explicit_agent_id
+
+    assert await prepared.managed_lease.finalize_result(status=TaskStatus.FAILED)
+    engine.dispose()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("bad_value", ["not-a-number", True, None, {"id": 1}])
```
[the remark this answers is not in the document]
OliverBryant: Added in 9fa1b05c — `12.5` is now in the malformed-value parametrization.

src/xagent/web/services/channel_runtime.py
```diff
@@ -635,6 +663,13 @@ def _prepare_channel_task_sync(
                 raise ChannelAuthorizationError(
                     "Channel owner changed during sender authorization"
                 )
+            explicit_agent_requested = agent_id is not None
+            if agent_id is None:
+                # No explicit selection from the transport (Slack has no agent
+                # picker; Telegram only passes one after /agents): fall back to
+                # the channel's configured default. A stale id degrades through
+                # the missing-agent path below, never fails the turn.
+                agent_id = owner.default_agent_id
```
rogercloud, line 672: Telegram's `/agents` picker has an explicit "Default assistant" option, and selecting it calls `_set_selected_agent(user_id, None)` (`src/xagent/web/channels/telegram/bot.py:1140`) — which pops the stored selection and sends the same `agent_id=None` as a user who never picked anything. Both cases land here with `explicit_agent_requested == False`, so an explicit "Default assistant" pick silently binds the user to the channel default instead, contradicting the bot's own "Default assistant selected" confirmation and the PR's stated precedence rule.

The missing-agent recovery path has the same problem: `telegram/bot.py:1667-1672` tells the user "I'm using the default assistant for this conversation," but the next turn re-resolves `None` to the channel default and evicts that task via the binding-drift check at line 740.

This needs a tri-state at the boundary — a sentinel distinct from `None`, or an explicit `use_channel_default: bool` / `Literal` parameter — rather than an `is not None` check. Worth a test for "explicitly chose default on a channel that has `default_agent_id` set" too; the current tests only cover "never picked."

src/xagent/web/services/channel_runtime.py
```diff
@@ -303,7 +307,31 @@ def _load_channel_owner_sync(
...
         if external_user_id not in allowed_user_ids:
             raise ChannelAuthorizationError("Channel sender is not authorized")
-    return ChannelOwnerSnapshot(user_id=owner_id)
+    return ChannelOwnerSnapshot(
+        user_id=owner_id,
+        default_agent_id=_coerce_agent_id(channel.config.get("default_agent_id")),
+    )
+
+
+def _coerce_agent_id(value: object) -> int | None:
```
rogercloud, line 316: The docstring promises a malformed value degrades to "no default" rather than failing every turn, but an oversized integer escapes that. `db.query(Agent).filter(Agent.id == 10**30).first()` raises `OverflowError: Python int too large to convert to SQLite INTEGER` at parameter binding, before any "not found" result. It isn't caught anywhere — `_prepare_channel_task_sync` re-raises after rollback, and the transports only catch `ChannelAuthorizationError` / `ChannelConfigurationError` — so an oversized value in `config["default_agent_id"]` fails every inbound message on the channel indefinitely. Suggest a magnitude guard (reject values outside a plausible int64 / column range) alongside the existing bool/int/str handling.

src/xagent/web/services/channel_runtime.py
```diff
@@ -635,6 +663,13 @@ def _prepare_channel_task_sync(
                 raise ChannelAuthorizationError(
                     "Channel owner changed during sender authorization"
                 )
+            explicit_agent_requested = agent_id is not None
```
rogercloud, line 666: Minor readability: `agent_id` is mutated in place, so it means "requested" above line 672 and "effective" below it, with `explicit_agent_requested` carrying the bit the mutation loses. Consider a separate local instead:

```python
effective_agent_id = agent_id if agent_id is not None else owner.default_agent_id
```

and use `agent_id is not None` directly at both read sites of `explicit_agent_requested`. Same behavior, one fewer variable, and no "one name, two meanings" in the same function.

src/xagent/web/services/channel_runtime.py
```diff
@@ -690,7 +725,18 @@ def _prepare_channel_task_sync(
                     .filter(Agent.id == int(agent_id))
                     .first()
                 )
-                requested_agent_missing = agent_row is None
+                # The missing-agent flag drives transport UX for a selection
+                # the USER made (Telegram clears the stored pick and says so).
+                # A stale channel-config default is an operator concern, not a
+                # user action — log it instead of blaming the user's choice.
+                requested_agent_missing = agent_row is None and explicit_agent_requested
```
rogercloud, line 732: `requested_agent_missing` has narrowed in meaning with this change: it used to be `agent_row is None` unconditionally, and is now `agent_row is None and explicit_agent_requested`, so a stale channel-config default no longer sets it. That's only documented in this inline comment. Worth a one-line comment on the field definitions themselves (`ClaimedChannelTask` and `_ChannelTaskClaimSnapshot`) so a future transport author doesn't read it as "any missing agent."

src/xagent/web/services/channel_runtime.py
```diff
@@ -690,7 +725,18 @@ def _prepare_channel_task_sync(
...
+                if agent_row is None and not explicit_agent_requested:
+                    logger.warning(
```
rogercloud, line 734: This is the only `logger.warning` in this file on a per-message path — the other two are on rare compensation/error branches. A busy channel with a stale `default_agent_id` will emit one warning per inbound message indefinitely. Suggest throttling, or logging once per detected staleness instead of once per message.

tests/web/services/test_channel_runtime.py
```diff
@@ -115,6 +115,191 @@ async def test_prepare_channel_task_binds_owned_agent(
     engine.dispose()
 
 
+def _set_channel_config_value(SessionLocal, channel_id: int, key: str, value) -> None:
```
rogercloud, line 118: All five new tests below pass `active_task_id=None`, so none covers the eviction that happens when `default_agent_id` changes on a channel with an in-flight conversation. The mechanism itself is already covered for the explicit-`agent_id` trigger (`test_prepare_channel_task_evicts_existing_task_when_selection_stale`, `test_prepare_channel_task_starts_new_task_when_binding_drifts`), so this is low risk — but a second-turn test driven by the config default would close the loop cheaply using the same pattern.

rogercloud (changes requested):
## Summary

This PR adds a per-channel default agent. When a transport (Telegram, Slack, Feishu) calls `prepare_channel_task` without an explicit `agent_id`, the runtime now resolves the channel's `config["default_agent_id"]` instead of falling straight through to the generic assistant — which is the only option today on Slack and Feishu, since neither has an agent picker. Malformed config values degrade to "no default" via `_coerce_agent_id`, and a stale default (deleted or inaccessible agent) reuses the existing missing-agent eviction path, gated by a new `explicit_agent_requested` flag so it doesn't fire the user-facing "your agent is gone" notice that is meant only for a user's own stale explicit pick.

## Re-review context

The only prior reviewer on this PR is the gemini-code-assist bot. Its finding — `_coerce_agent_id` silently truncating floats (`int(12.5) -> 12`), plus a request to cover `12.5` in the malformed-value parametrization — is **fixed** in `9fa1b05c`. The coercion now dispatches explicitly on `bool` → `int` → `str` → `None`, so any float falls through to `None`, and the parametrized test includes `12.5`. This review covers the current head.

The mechanism is a minimal hook into an already well-factored per-turn revalidation path. `ChannelOwnerSnapshot.default_agent_id` is re-resolved from the DB on every turn with no caching, and the precedence math is right — the explicit value is read before the fallback is applied, so an explicit pick always wins over the config default.

The reservation is that all of this rides on an already-overloaded `agent_id is None` contract, which Telegram's own UI uses to mean something different. That collision produces one real regression (finding 1) and forces the `explicit_agent_requested` bookkeeping variable to exist at all.

A simpler shape, worth considering independently of the bug: introduce a distinct local rather than mutating the parameter —

```python
effective_agent_id = agent_id if agent_id is not None else owner.default_agent_id
```

— and use `agent_id is not None` directly at the two sites that currently read `explicit_agent_requested`. That removes the extra flag and the "one variable, two meanings above and below line 672" pattern, with no behavior change.

## Findings

### Major

**1. Telegram's explicit "Default assistant" selection is silently overridden by the channel default**

`src/xagent/web/services/channel_runtime.py:666-672`

Telegram's `/agents` picker offers a "Default assistant" option whose callback calls `self._set_selected_agent(user_id, None)` (`src/xagent/web/channels/telegram/bot.py:1140`), which pops the stored selection. That produces the exact same wire value — `agent_id=None` — as "this user never picked anything." There is no sentinel distinguishing the two, so `explicit_agent_requested = agent_id is not None` is `False` in both cases and both take the `agent_id = owner.default_agent_id` fallback at line 672.

The result: on a channel with `default_agent_id` set, a user who taps "Default assistant" gets the confirmation *"Default assistant selected. Send a message to start a fresh conversation."* and is then bound to the channel's pinned agent instead. That contradicts the bot's own message and the PR description's claim that Telegram's explicit selection takes precedence over the config default.

The same collision hits the missing-agent recovery path (`telegram/bot.py:1667-1672`): the bot says "I'm using the default assistant for this conversation," but the next turn re-resolves `None` to the channel default, silently rebinding the user and evicting the just-created default task via the binding-drift check at `channel_runtime.py:740-747`.

This is new to this PR — at the base commit there was no `default_agent_id` concept, and `agent_id=None` from Telegram unambiguously meant the generic default. It is also untested; the closest new test, `test_prepare_channel_task_uses_channel_default_agent`, covers "never picked," not "explicitly picked default."

Fixing this needs a tri-state at the boundary — a sentinel distinct from `None`, or an explicit `use_channel_default: bool` / `Literal` parameter — not a refinement of the `is not None` check.

### Minor

**2. `_coerce_agent_id` doesn't guard oversized integers, so a bad config raises `OverflowError` on every message**

`src/xagent/web/services/channel_runtime.py:316-334`

The docstring promises that "a malformed value must degrade to 'no default' rather than fail every turn on the channel." An oversized int breaks that promise: `db.query(Agent).filter(Agent.id == 10**30).first()` raises `OverflowError: Python int too large to convert to SQLite INTEGER` at the parameter-binding layer, before any "not found" result. Nothing catches it — `_prepare_channel_task_sync`'s `except Exception: db.rollback(); raise` re-raises, and none of the three transports catch anything broader than `ChannelAuthorizationError` / `ChannelConfigurationError`. An accidental or upstream-generated oversized value in `config["default_agent_id"]` therefore fails every inbound message on the channel indefinitely. A magnitude check alongside the existing bool/int/str handling is enough.

(Negative ids, zero, `"1_0"`, and full-width digit strings were checked and are fine — they degrade through the normal "not found" path.)

**3. `agent_id` parameter is mutated in place**

`src/xagent/web/services/channel_runtime.py:666-673` — see the `effective_agent_id` suggestion in the design section above. Style only, behavior-preserving.

**4. Undocumented semantic narrowing of `requested_agent_missing`**

`src/xagent/web/services/channel_runtime.py` — field on `ClaimedChannelTask` (~line 89) and `_ChannelTaskClaimSnapshot` (~line 100), assigned at ~line 732. The flag used to mean "the resolved agent is missing" unconditionally; it now means "the agent the user *explicitly* requested is missing," and a stale channel default no longer sets it. That change is recorded only in an inline comment at the assignment site. A one-line comment on each field definition would keep a future transport author from misreading its scope.

**5. `default_agent_id` is validated only at read time**

`src/xagent/web/schemas/user_channel.py:20` types `config` as a free-form `Optional[Dict[str, Any]]`, and `update_user_channel` in `src/xagent/web/api/channel.py` (~line 784+) only guards a fixed set of Slack-managed keys — every other key, including `default_agent_id`, is written through unconditionally. An operator can `PUT` a non-integer or inaccessible agent id, get a 200, and only discover the mistake as a silent per-turn degradation plus a log warning on the next message. Read-time handling is safe; write-time validation would just put the feedback where the operator is.

**6. Unbounded per-message warning for a stale default**

`src/xagent/web/services/channel_runtime.py:734-739` is the only `logger.warning` in this file on a hot every-message path (the other two sit on rare compensation/error branches). A busy channel with a stale `default_agent_id` emits one warning per inbound message forever, with no dedup. Worth throttling, or logging once per detected staleness.

**7. No test for eviction when `default_agent_id` changes on an existing conversation**

`tests/web/services/test_channel_runtime.py` — all five new tests pass `active_task_id=None`. The eviction mechanism itself is already proven by the pre-existing `test_prepare_channel_task_evicts_existing_task_when_selection_stale` and `test_prepare_channel_task_starts_new_task_when_binding_drifts`, which drive the same branch through the explicit `agent_id=` kwarg, so the risk is low — but there is no test showing the config-default path triggers that eviction on a second turn. Cheap to add given the existing pattern.

### Design concern (non-blocking)

**Operator config changes silently reset live conversations on Slack and Feishu.**

`src/xagent/web/services/channel_runtime.py:666-672` combined with the pre-existing eviction at `:740-747`. Because `default_agent_id` is re-read from the DB every turn, and the eviction rule discards any in-flight task whose bound `agent_id` no longer matches, setting or changing `config["default_agent_id"]` will start a fresh conversation for every active user on that channel at their next message.

Before this PR that branch was structurally unreachable for Slack and Feishu — neither ever passed `agent_id=`, so the value stayed `None` for the whole function on every turn. Telegram could already reach it, but only through a user's own per-conversation `/agents` pick, never a channel-wide operator action. This PR makes a single operator edit reset everyone.

No transport surfaces it either: Slack (`slack/bot.py:490-548`) and Feishu (`feishu/bot.py:272-330`) treat `is_new_task` as internal bookkeeping with no user-facing message, and Telegram's reset notice is gated by `explicit_agent_requested`, which is `False` for exactly this trigger. There's no data loss — the old task remains queryable — and nothing in the diff establishes a "set once at onboarding" convention that would make this rare. Worth a changelog note or an admin-facing warning ("changing this starts a new conversation for all users on this channel") rather than a code change.

## Tests

`pytest tests/web/services/test_channel_runtime.py -q` → **32 passed**. The five new tests cover the happy-path fallback, malformed values, and the stale-default eviction; gaps noted in findings 1 and 7.

OliverBryant:
Stale since the v2 architecture rewrite; the V0/V1 scope has shipped through other PRs.

Issue #1071, opened 2026-07-30
refactor(checkpoint): support atomic interaction checkpoint handoff


AlexLiu190625:
## Summary

Make the exact checkpoint row a durable, transaction-composable anchor for both running-task recovery and blocking interaction continuation.

## Current behavior

`Task.last_checkpoint_event_id` stores a logical trace-event identifier. `trace_events.event_id` is not a unique database key, checkpoint persistence commits internally, and pruning is not aware of active interaction resume anchors. That combination cannot support an atomic checkpoint + waiting-state + interaction handoff.

## Acceptance criteria

- [ ] Add a nullable `Task.last_checkpoint_trace_event_id` foreign key to the `trace_events.id` primary key, with delete restriction while referenced.
- [ ] Depend on the checkpoint read contract that distinguishes `found`, authoritatively `absent`, transiently `unavailable`, and corrupt state; never collapse unavailable into absence.
- [ ] Reuse the existing shared checkpoint reader/alias helper rather than creating parallel type-alias logic.
- [ ] Backfill and dual-read/write during the compatibility window. Resolve the legacy logical identifier only within the Task partition and only when exactly one valid row matches; zero or multiple matches fail closed and emit telemetry.
- [ ] Migrate running-task lease recovery and its compare-and-swap fence to the primary-key locator while retaining the legacy field for mixed-version rollback.
- [ ] Expose checkpoint staging that joins a caller-owned transaction and performs no implicit commit, rollback, notification, or prune.
- [ ] Migrate every Task deletion path, including single-Task and bulk owner deletion, to clear dependent pointers/rows before trace cleanup.
- [ ] Task deletion explicitly clears or terminates dependent pointers before trace cleanup.
- [ ] Cover online and offline migrations, downgrade, SQLite, PostgreSQL, ambiguous backfill, concurrent recovery, prune, and deletion ordering.

## Dependencies

- [Checkpoint availability result contract](https://github.com/xorbitsai/xagent/issues/978)

## Non-goals

- Dropping `Task.last_checkpoint_event_id` in this change.
- Adding an interaction table or public API.

AlexLiu190625:
## Implementation design

- Contract owner: checkpoint persistence and lookup remain in the checkpoint/trace layer; interaction code receives an exact row locator and does not query for the latest checkpoint.
- Transaction boundary: the native publication owner stages the trace row and related lifecycle rows in one transaction. Trace projection and pruning run only after commit.
[...]

PR #1137, opened 2026-08-04
refactor(checkpoint): anchor the task checkpoint pointer to the exact trace row

AlexLiu190625:
## Summary

Checkpoint reads and lease recovery currently locate a task's checkpoint by scanning trace events and matching `last_checkpoint_event_id`, a string column with no referential guarantee: the pointed-at row can be ambiguous (event ids carry no unique constraint), deleted, or resolved to the wrong row. This PR anchors the pointer to the exact row: `tasks.last_checkpoint_trace_event_id` references `trace_events.id` directly, written in the same statement as the legacy pointer and consumed PK-first everywhere the pointer is read.

- **Schema**: new nullable `last_checkpoint_trace_event_id` column with a named foreign key declared on the model (PostgreSQL enforces it via an `ADD CONSTRAINT` migration; upgraded SQLite databases get the column without a DB-level constraint, and a test asserts that shape, so the application-layer ordering below is their only guard — fresh SQLite databases created from metadata do carry the constraint). The migration backfills existing rows with a single correlated `UPDATE` that resolves each legacy pointer to its row id, leaves ambiguous or unresolvable pointers `NULL`, and is idempotent.
- **Writer**: the checkpoint write flushes inside the checkpoint branch, asserts the row id, and updates both pointer columns in one fenced statement. A pointer update matching zero rows now distinguishes a deleted task (silent stop) from a changed lease (raise), instead of misreporting both as lease changes.
- **Reader**: checkpoint loads resolve the anchor row by primary key first; a resolved row failing verification is corrupt, while a dangling anchor (reachable only on constraint-less upgraded SQLite) falls back to the legacy scan with a warning. Rows written before the anchor existed keep the legacy scan path.
- **Deletion**: both task-purge implementations null both pointer columns before deleting trace events, so the delete order stays valid under enforced foreign keys; tests cover the constraint-enforced and constraint-less schema shapes, including the mutation that reverts the ordering.
- **Lease recovery**: checkpoint resolution returns a three-way verdict — recoverable, not recoverable, or indeterminate. Resolution is PK-first with no fallback on verification failure; a dangling anchor falls back to the legacy scan; an ambiguous legacy match (previously an unhandled `MultipleResultsFound`) now leaves the candidate untouched for the next sweep instead of aborting the recovery pass. The recovery CAS fence matches both pointer columns conjunctively, so single-column drift blocks the overwrite.
- **Pointer lifecycle**: every site that clears or retains the pointer moves both columns together — lease acquisition's new-run clear, the retain-case builders (which share one retention predicate so they cannot diverge), the turn-claim path, and both deletion implementations.

## Verification

- 299 tests on SQLite and 143 on PostgreSQL across the lease-service, recovery, orchestrator, deletion, migration, checkpoint-stream, and trace-storage suites, including: dual-write assertions against the row id; the fence's parametrized single-column-drift matrix; dangling-anchor fallback on an upgraded-SQLite fixture; ambiguity leaving the candidate re-scannable; migration backfill correctness, ambiguity-to-NULL, idempotency, downgrade, and named-constraint reflection on PostgreSQL.
- Mutation checks, each red against a named test and reverted: dropping the flush, the second written column, either deletion's NULL-first ordering, the backfill idempotency guard, the fence conjunct, the builder retention predicate or its ELSE branch, the dangling-anchor fallback, the ambiguity handler, and the turn-claim clear.

Scope: this lands the anchor column, migration, dual-write/dual-read compatibility window, recovery migration, and deletion ordering from #1071. The remaining criterion — checkpoint staging that joins a caller-owned transaction with no implicit commit — builds on this anchor and lands as its own change.

Part of #1071.

### DIFF
--- a/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
+++ b/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
@@ -1,0 +1,111 @@
+"""add exact-row checkpoint anchor to tasks
+
+Adds ``tasks.last_checkpoint_trace_event_id``, an integer pointer to the
+``trace_events`` row that ``last_checkpoint_event_id`` (the legacy UUID
+string column) names. On PostgreSQL the column is backed by a named foreign
+key; on SQLite the migration adds the column only -- ``ALTER TABLE ADD
+CONSTRAINT`` is not renderable through Alembic's SQLite batch mode without a
+full table rebuild, so an upgraded SQLite database has no DB-level FK here.
+A database built fresh through ``Base.metadata.create_all()`` (new installs,
+tests) gets the FK from the model directly, regardless of dialect.
+
+Existing rows are backfilled by resolving the legacy string column against
+the row it names, but only where that resolution is unambiguous: zero or
+multiple matching trace_events rows leave the new column NULL rather than
+guessing or aborting the migration.
+
+Revision ID: 20260804_add_task_checkpoint_trace_event_anchor
+Revises: 20260729_add_gmail_audience_grace
+Create Date: 2026-08-04
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260804_add_task_checkpoint_trace_event_anchor"
+down_revision: Union[str, None] = "20260729_add_gmail_audience_grace"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+TABLE = "tasks"
+COLUMN = "last_checkpoint_trace_event_id"
+LEGACY_COLUMN = "last_checkpoint_event_id"
+FK_NAME = "fk_tasks_last_checkpoint_trace_event_id"
+
+# One correlated-subquery UPDATE. The subquery is wrapped in COUNT/CASE so it
+# always returns a single scalar row -- a bare correlated subquery would
+# raise on multiple matches instead of resolving to NULL, and this backfill
+# must never abort the migration on ambiguous or missing legacy data (a
+# concurrently deleted target task resolves the same way: zero matches).
+BACKFILL_SQL = sa.text(
+    f"""
+    UPDATE {TABLE}
+    SET {COLUMN} = (
+        SELECT CASE WHEN COUNT(*) = 1 THEN MIN(te.id) ELSE NULL END
+        FROM trace_events te
+        WHERE te.task_id = {TABLE}.id
+          AND te.event_id = {TABLE}.{LEGACY_COLUMN}
+          AND te.build_id IS NULL
+          AND te.event_type = 'system_update_general'
+    )
+    WHERE {COLUMN} IS NULL
+      AND {LEGACY_COLUMN} IS NOT NULL
+    """
+)
+
+
+def _table_exists() -> bool:
+    inspector = sa.inspect(op.get_bind())
+    return inspector.has_table(TABLE)
+
+
+def _columns() -> set[str]:
+    inspector = sa.inspect(op.get_bind())
+    return {col["name"] for col in inspector.get_columns(TABLE)}
+
+
+def _fk_names() -> set[str]:
+    inspector = sa.inspect(op.get_bind())
+    return {fk["name"] for fk in inspector.get_foreign_keys(TABLE)}
+
+
+def upgrade() -> None:
+    # tasks is created by Base.metadata.create_all() in production, not by
+    # a migration in this repo; a from-scratch (bare) database has no tasks
+    # table yet when migrations run, so this is a no-op there.
+    if not _table_exists():
+        return
+
+    bind = op.get_bind()
+    is_postgresql = bind.dialect.name == "postgresql"
+
+    if COLUMN not in _columns():
+        op.add_column(TABLE, sa.Column(COLUMN, sa.Integer(), nullable=True))
+
+    if is_postgresql and FK_NAME not in _fk_names():
+        op.create_foreign_key(
+            FK_NAME,
+            TABLE,
+            "trace_events",
+            [COLUMN],
+            ["id"],
+        )
+
+    op.execute(BACKFILL_SQL)
+
+
+def downgrade() -> None:
+    if not _table_exists():
+        return
+
+    bind = op.get_bind()
+    is_postgresql = bind.dialect.name == "postgresql"
+
+    if is_postgresql and FK_NAME in _fk_names():
+        op.drop_constraint(FK_NAME, TABLE, type_="foreignkey")
+
+    if COLUMN in _columns():
+        op.drop_column(TABLE, COLUMN)
--- a/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
+++ b/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
@@ -31,6 +31,7 @@ branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
 TABLE = "tasks"
+TRACE_TABLE = "trace_events"
 COLUMN = "last_checkpoint_trace_event_id"
 LEGACY_COLUMN = "last_checkpoint_event_id"
 FK_NAME = "fk_tasks_last_checkpoint_trace_event_id"
@@ -57,9 +58,9 @@ BACKFILL_SQL = sa.text(
 )
 
 
-def _table_exists() -> bool:
+def _table_exists(name: str = TABLE) -> bool:
     inspector = sa.inspect(op.get_bind())
-    return inspector.has_table(TABLE)
+    return inspector.has_table(name)
 
 
 def _columns() -> set[str]:
@@ -85,11 +86,14 @@ def upgrade() -> None:
     if COLUMN not in _columns():
         op.add_column(TABLE, sa.Column(COLUMN, sa.Integer(), nullable=True))
 
+    if not _table_exists(TRACE_TABLE):
+        return
+
     if is_postgresql and FK_NAME not in _fk_names():
         op.create_foreign_key(
             FK_NAME,
             TABLE,
-            "trace_events",
+            TRACE_TABLE,
             [COLUMN],
             ["id"],
         )
--- a/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
+++ b/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
@@ -7,7 +7,13 @@ key; on SQLite the migration adds the column only -- ``ALTER TABLE ADD
 CONSTRAINT`` is not renderable through Alembic's SQLite batch mode without a
 full table rebuild, so an upgraded SQLite database has no DB-level FK here.
 A database built fresh through ``Base.metadata.create_all()`` (new installs,
-tests) gets the FK from the model directly, regardless of dialect.
+tests) gets the FK from the model directly, regardless of dialect. That
+divergence between fresh and upgraded SQLite databases is permanent under
+the current migration set: this revision is the head and no later revision
+reconciles the two shapes. On an upgraded SQLite database the pointer's
+delete protection is therefore the application-level clearing order in
+``task_deletion.py``, which nulls both pointer columns before deleting the
+task's ``trace_events`` rows -- not the database.
 
 Existing rows are backfilled by resolving the legacy string column against
 the row it names, but only where that resolution is unambiguous: zero or
--- a/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
+++ b/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
@@ -14,6 +14,14 @@ the row it names, but only where that resolution is unambiguous: zero or
 multiple matching trace_events rows leave the new column NULL rather than
 guessing or aborting the migration.
 
+On PostgreSQL ``create_foreign_key`` takes an ACCESS EXCLUSIVE lock on
+``tasks`` and validates every existing row before returning, and the
+backfill's correlated subquery matches ``trace_events`` on
+``(task_id, event_id)`` where only ``task_id`` is indexed. On a large
+deployment both steps should be scheduled with that lock window in mind; a
+``NOT VALID`` constraint followed by a separate ``VALIDATE CONSTRAINT``
+would move the validation out of the lock window if that becomes necessary.
+
 Revision ID: 20260804_add_task_checkpoint_trace_event_anchor
 Revises: 20260729_add_gmail_audience_grace
 Create Date: 2026-08-04
@@ -108,8 +116,21 @@ def downgrade() -> None:
     bind = op.get_bind()
     is_postgresql = bind.dialect.name == "postgresql"
 
-    if is_postgresql and FK_NAME in _fk_names():
-        op.drop_constraint(FK_NAME, TABLE, type_="foreignkey")
+    if COLUMN not in _columns():
+        return
 
-    if COLUMN in _columns():
+    if is_postgresql:
+        if FK_NAME in _fk_names():
+            op.drop_constraint(FK_NAME, TABLE, type_="foreignkey")
         op.drop_column(TABLE, COLUMN)
+    else:
+        # SQLite renders this column's foreign key inline in CREATE TABLE
+        # even under use_alter=True (a fresh install builds tasks with
+        # create_all, not a migration), and refuses to drop a column an
+        # inline foreign key names. A full table rebuild removes both
+        # together; nothing else can. The migration connection already runs
+        # with SQLite foreign keys off for exactly this reason -- see
+        # _migration_connection in src/xagent/db/migration.py -- and
+        # re-checks for new violations afterwards.
+        with op.batch_alter_table(TABLE, recreate="always") as batch_op:
+            batch_op.drop_column(COLUMN)
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -33,7 +33,7 @@ from ...web.services.ops_signals import (
     CHECKPOINT_DECODE_FALLBACK,
     CHECKPOINT_LOAD_UNAVAILABLE,
     CHECKPOINT_PK_ANCHOR_DANGLING,
-    CHECKPOINT_PRUNE_INTEGRITY_ERROR,
+    CHECKPOINT_PRUNE_FAILED,
     clear_degradation,
     register_degradation,
 )
@@ -616,7 +616,18 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 exc_info=True,
             )
             return _AnchorFallback(generic_failure=True)
+        # The scan retires these two signals on two different facts: the
+        # load signal once a page query returns, the decode-fallback signal
+        # once a row decodes. An anchored read establishes both in one round
+        # trip, so both clears land here -- and here rather than at the top
+        # of this function, where the dangling clear sits, because that one
+        # reports on the pointer every attempt re-reads while these report
+        # that the read actually got through. An anchored read returns
+        # without ever entering the scan, so without these a decode-fallback
+        # signal set by one bad row could never retire in the steady state
+        # this anchor exists to produce.
         clear_degradation(CHECKPOINT_LOAD_UNAVAILABLE)
+        clear_degradation(CHECKPOINT_DECODE_FALLBACK)
 
         snapshot = decoded.get("snapshot") if isinstance(decoded, dict) else None
         if not isinstance(snapshot, dict):
@@ -906,7 +917,7 @@ class DatabaseTraceHandler(BaseTraceHandler):
             # healthy, whatever it found. Clearing only after a successful
             # delete would leave a steady state with nothing to prune unable
             # to clear the signal at all.
-            clear_degradation(CHECKPOINT_PRUNE_INTEGRITY_ERROR)
+            clear_degradation(CHECKPOINT_PRUNE_FAILED)
             # Rank first, protect after. The row this task's exact-row
             # pointer references is never deleted, whatever its position in
             # the retention ranking; excluding it from the candidate set
@@ -944,14 +955,20 @@ class DatabaseTraceHandler(BaseTraceHandler):
             # surface as psycopg2's SerializationFailure or DeadlockDetected,
             # both of which SQLAlchemy wraps in OperationalError, not
             # IntegrityError -- catch both so this retention path degrades
-            # the same way regardless of which one the database raises.
+            # the same way regardless of which one the database raises. The
+            # signal is named for the outcome rather than for either cause:
+            # OperationalError also covers lock timeouts and dropped
+            # connections, /health publishes only the signal name, and the
+            # operator's response to all of them is the same -- retention
+            # stopped, rows are accumulating, read the logged traceback for
+            # which one it was.
             db.rollback()
             register_degradation(
-                CHECKPOINT_PRUNE_INTEGRITY_ERROR,
-                f"task {self.task_id}: checkpoint prune hit an integrity error",
+                CHECKPOINT_PRUNE_FAILED,
+                f"task {self.task_id}: checkpoint prune could not delete stale rows",
             )
             logger.warning(
-                "Checkpoint prune hit an integrity error for task %s",
+                "Checkpoint prune failed for task %s",
                 self.task_id,
                 exc_info=True,
             )
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -635,7 +635,12 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 # this the exact-row anchor column would be built from an
                 # unassigned id and silently written NULL.
                 db.flush()
-                assert trace_event.id is not None
+                if trace_event.id is None:
+                    raise RuntimeError(
+                        f"Task {self.task_id} checkpoint {event.id} row has no "
+                        "primary key after flush; refusing to write a NULL "
+                        "checkpoint anchor"
+                    )
                 pointer_update = db.execute(
                     update(Task)
                     .where(
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -31,6 +31,8 @@ from ...web.models.tool_config import ToolUsage
 from ...web.services.ops_signals import (
     CHECKPOINT_DECODE_FALLBACK,
     CHECKPOINT_LOAD_UNAVAILABLE,
+    CHECKPOINT_PK_ANCHOR_DANGLING,
+    CHECKPOINT_PRUNE_INTEGRITY_ERROR,
     clear_degradation,
     register_degradation,
 )
@@ -63,6 +65,11 @@ CHECKPOINT_ROW_SCAN_LIMIT = 100
 # scan that reaches the cap without a resolution is treated as unavailable
 # rather than continuing indefinitely.
 CHECKPOINT_SCAN_MAX_PAGES = 50
+
+# Sentinel distinguishing "no PK anchor to try" (unset pointer, or a pointer
+# whose row is gone) from a resolved snapshot, which may legitimately be an
+# empty dict. Falling back to the legacy scan is only correct for the former.
+_NO_TRACE_EVENT_ANCHOR = object()
 
 
 def _checkpoint_execution_id_predicate(execution_id: str) -> Any:
@@ -215,6 +222,9 @@ class DatabaseTraceHandler(BaseTraceHandler):
                         f"task {self.task_id}: could not resolve the "
                         "checkpoint read partition"
                     ) from exc
+                anchored = self._load_pk_anchored_checkpoint(db, run_id)
+                if anchored is not _NO_TRACE_EVENT_ANCHOR:
+                    return anchored  # type: ignore[return-value]
                 query = query.filter(
                     DatabaseTraceEvent.build_id.is_(None),
                     self._checkpoint_run_partition_filter(run_id),
@@ -440,6 +450,93 @@ class DatabaseTraceHandler(BaseTraceHandler):
         run_field = DatabaseTraceEvent.data[TASK_RUN_ID_TRACE_FIELD].as_string()
         return run_field == run_id if run_id is not None else run_field.is_(None)
 
+    def _load_pk_anchored_checkpoint(
+        self,
+        db: Session,
+        run_id: str | None,
+    ) -> Dict[str, Any] | object:
+        """Resolve the checkpoint through the task's exact-row pointer.
+
+        Returns ``_NO_TRACE_EVENT_ANCHOR`` when there is nothing to anchor
+        on: the pointer is unset, or it names a row that no longer exists
+        (only possible on a database upgraded through Alembic rather than
+        created fresh, since that path has no DB-level FK -- see the
+        migration that adds this column). Both cases fall back to the
+        legacy scan. Once a target row is found, though, it is
+        authoritative: a validation mismatch raises rather than falling
+        back to search other rows, and a decode failure on that row is
+        classified the same way the scan classifies an exhausted,
+        all-undecodable candidate set, since this pointer is the only
+        candidate.
+        """
+        pointer = (
+            db.query(Task.last_checkpoint_trace_event_id)
+            .filter(Task.id == self.task_id)
+            .one_or_none()
+        )
+        if pointer is None or pointer[0] is None:
+            return _NO_TRACE_EVENT_ANCHOR
+
+        pointer_id = pointer[0]
+        row = db.get(DatabaseTraceEvent, pointer_id)
+        if row is None:
+            register_degradation(
+                CHECKPOINT_PK_ANCHOR_DANGLING,
+                f"task {self.task_id}: checkpoint pointer {pointer_id} has "
+                "no matching trace_events row; falling back to the legacy "
+                "scan",
+            )
+            return _NO_TRACE_EVENT_ANCHOR
+        clear_degradation(CHECKPOINT_PK_ANCHOR_DANGLING)
+
+        row_data: Dict[str, Any] = row.data if isinstance(row.data, dict) else {}
+        run_field = row_data.get(TASK_RUN_ID_TRACE_FIELD)
+        partition_matches = (
+            run_field == run_id if run_id is not None else run_field is None
+        )
+        if (
+            row.task_id != self.task_id
+            or row.event_type != "system_update_general"
+            or row.build_id is not None
+            or row_data.get("checkpoint_type") not in READABLE_CHECKPOINT_TYPES
+            or not partition_matches
+        ):
+            raise CheckpointCorruptError(
+                f"task {self.task_id}: checkpoint pointer {pointer_id} does "
+                "not match the row it anchors"
+            )
+
+        try:
+            decoded = decode_trace_event_data(
+                db,
+                task_id=self.task_id,
+                data=dict(row_data),
+                strict=True,
+            )
+        except CheckpointMessageDecodeError as exc:
+            raise CheckpointCorruptError(
+                f"task {self.task_id}: checkpoint pointer {pointer_id} row "
+                "is undecodable"
+            ) from exc
+        except Exception as exc:
+            register_degradation(
+                CHECKPOINT_LOAD_UNAVAILABLE,
+                f"task {self.task_id}: checkpoint pointer {pointer_id} decode failed",
+            )
+            raise CheckpointUnavailableError(
+                f"task {self.task_id}: checkpoint pointer {pointer_id} "
+                "could not be decoded"
+            ) from exc
+        clear_degradation(CHECKPOINT_LOAD_UNAVAILABLE)
+
+        snapshot = decoded.get("snapshot") if isinstance(decoded, dict) else None
+        if not isinstance(snapshot, dict):
+            raise CheckpointCorruptError(
+                f"task {self.task_id}: checkpoint pointer {pointer_id} has "
+                "a readable checkpoint_type but no snapshot"
+            )
+        return dict(snapshot)
+
     def _sync_save_to_database(self, event: CoreTraceEvent) -> None:
         """Synchronous database save operation (runs in thread pool)."""
         # Create database session
@@ -518,6 +615,13 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 and self.build_id is None
                 and checkpoint_lease is not None
             ):
+                # Flush the pending insert so trace_event.id (the row's
+                # primary key) is assigned before the pointer UPDATE below
+                # reads it. autoflush is off for this session, so without
+                # this the exact-row anchor column would be built from an
+                # unassigned id and silently written NULL.
+                db.flush()
+                assert trace_event.id is not None
                 pointer_update = db.execute(
                     update(Task)
                     .where(
@@ -526,14 +630,33 @@ class DatabaseTraceHandler(BaseTraceHandler):
                         Task.runner_id == checkpoint_lease.runner_id,
                         Task.run_id == checkpoint_lease.run_id,
                     )
-                    .values(last_checkpoint_event_id=str(event.id))
+                    .values(
+                        last_checkpoint_event_id=str(event.id),
+                        last_checkpoint_trace_event_id=trace_event.id,
+                    )
                     .execution_options(synchronize_session=False)
                 )
                 if int(getattr(pointer_update, "rowcount", 0) or 0) != 1:
-                    raise RuntimeError(
-                        f"Task {self.task_id} lease changed before checkpoint "
-                        f"{event.id} could be persisted"
+                    task_still_exists = (
+                        db.query(Task.id).filter(Task.id == self.task_id).first()
+                        is not None
                     )
+                    if not task_still_exists:
+                        # The task row is gone, not merely leased elsewhere --
+                        # the same outcome the trace_events.task_id FK would
+                        # produce for the row itself. Skip quietly rather
+                        # than raising a "lease changed" error that isn't
+                        # what happened.
+                        logger.debug(
+                            "Skip checkpoint pointer update for missing task %s: %s",
+                            self.task_id,
+                            event.id,
+                        )
+                    else:
+                        raise RuntimeError(
+                            f"Task {self.task_id} lease changed before checkpoint "
+                            f"{event.id} could be persisted"
+                        )
 
             # Update tool usage statistics if this is a tool execution event
             if event_type_str == "tool_execution_end":
@@ -637,11 +760,27 @@ class DatabaseTraceHandler(BaseTraceHandler):
                         run_id if isinstance(run_id, str) and run_id else None
                     )
                 )
+            # Exclude the row this task's exact-row pointer currently
+            # references, whatever its position in the retention ranking.
+            # For the steady-state writer this is structurally unreachable
+            # -- the pointer always names the row just written, which ranks
+            # ahead of the offset -- but it guards the backfill-vs-prune
+            # window and future back-pointing anchors that may point at an
+            # older row.
+            anchor = (
+                db.query(Task.last_checkpoint_trace_event_id)
+                .filter(Task.id == self.task_id)
+                .one_or_none()
+            )
+            anchor_id = anchor[0] if anchor is not None else None
+            candidate_filters = list(partition_filters)
+            if anchor_id is not None:
+                candidate_filters.append(DatabaseTraceEvent.id != anchor_id)
             stale_rows = (
                 db.query(DatabaseTraceEvent.id)
                 .filter(
                     DatabaseTraceEvent.task_id == self.task_id,
-                    *partition_filters,
+                    *candidate_filters,
                     DatabaseTraceEvent.event_type == "system_update_general",
                     DatabaseTraceEvent.data["checkpoint_type"]
                     .as_string()
@@ -665,11 +804,31 @@ class DatabaseTraceHandler(BaseTraceHandler):
                     DatabaseTraceEvent.id.in_(chunk)
                 ).delete(synchronize_session=False)
             db.commit()
+            clear_degradation(CHECKPOINT_PRUNE_INTEGRITY_ERROR)
             logger.debug(
                 "Pruned %d checkpoint rows for task %s execution %s",
                 len(stale_ids),
                 self.task_id,
                 execution_id,
+            )
+        except IntegrityError:
+            # PostgreSQL enforces the anchor FK, so a race that lets a
+            # candidate row become the active anchor between selection and
+            # delete surfaces here as a restrict violation (including a
+            # serialization failure under stricter isolation levels). A
+            # database upgraded through Alembic on SQLite has no DB-level FK
+            # for this column and cannot raise this; a freshly created
+            # SQLite database (create_all, e.g. tests) does have the FK and
+            # can.
+            db.rollback()
+            register_degradation(
+                CHECKPOINT_PRUNE_INTEGRITY_ERROR,
+                f"task {self.task_id}: checkpoint prune hit an integrity error",
+            )
+            logger.warning(
+                "Checkpoint prune hit an integrity error for task %s",
+                self.task_id,
+                exc_info=True,
             )
         except Exception:
             db.rollback()
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import logging
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import Any, Dict, Optional
 
@@ -66,10 +67,24 @@ CHECKPOINT_ROW_SCAN_LIMIT = 100
 # rather than continuing indefinitely.
 CHECKPOINT_SCAN_MAX_PAGES = 50
 
-# Sentinel distinguishing "no PK anchor to try" (unset pointer, or a pointer
-# whose row is gone) from a resolved snapshot, which may legitimately be an
-# empty dict. Falling back to the legacy scan is only correct for the former.
-_NO_TRACE_EVENT_ANCHOR = object()
+
+@dataclass(frozen=True)
+class _AnchorFallback:
+    """Why the PK anchor deferred to the legacy scan.
+
+    Distinguishes "no PK anchor to try" from a resolved snapshot, which may
+    legitimately be an empty dict. An unset or dangling pointer defers with
+    nothing to carry. A correctly identified anchor row whose payload could
+    not be read defers *and* hands the scan the same verdict flag the scan
+    would have set for that row itself: the scan's candidate filter can
+    legitimately exclude that row (a row carrying no execution identity is
+    anchored but not scanned, see _load_pk_anchored_checkpoint), and an
+    excluded row must never let an unreadable checkpoint become "no
+    checkpoint".
+    """
+
+    undecodable: bool = False
+    generic_failure: bool = False
 
 
 def _checkpoint_execution_id_predicate(execution_id: str) -> Any:
@@ -201,6 +216,7 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 # below), and a zero-row first page is authoritative.
                 _checkpoint_execution_id_predicate(str(execution_id)),
             )
+            anchored_fallback = _AnchorFallback()
             if self.build_id is None:
                 try:
                     run_id = self._root_checkpoint_read_partition(db)
@@ -225,8 +241,9 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 anchored = self._load_pk_anchored_checkpoint(
                     db, run_id, str(execution_id)
                 )
-                if anchored is not _NO_TRACE_EVENT_ANCHOR:
-                    return anchored  # type: ignore[return-value]
+                if not isinstance(anchored, _AnchorFallback):
+                    return anchored
+                anchored_fallback = anchored
                 query = query.filter(
                     DatabaseTraceEvent.build_id.is_(None),
                     self._checkpoint_run_partition_filter(run_id),
@@ -239,9 +256,11 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 DatabaseTraceEvent.id.desc(),
             )
 
-            saw_generic_failure = False
-            saw_undecodable_row = False
-            saw_any_row = False
+            saw_generic_failure = anchored_fallback.generic_failure
+            saw_undecodable_row = anchored_fallback.undecodable
+            saw_any_row = (
+                anchored_fallback.undecodable or anchored_fallback.generic_failure
+            )
             offset = 0
             page_count = 0
             while True:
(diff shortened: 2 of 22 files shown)


src/xagent/web/api/trace_handlers.py
```diff
@@ -518,6 +629,13 @@ def _save_trace_event(self, db: Session, event: CoreTraceEvent) -> None:
                 and self.build_id is None
                 and checkpoint_lease is not None
             ):
+                # Flush the pending insert so trace_event.id (the row's
+                # primary key) is assigned before the pointer UPDATE below
+                # reads it. autoflush is off for this session, so without
+                # this the exact-row anchor column would be built from an
+                # unassigned id and silently written NULL.
+                db.flush()
+                assert trace_event.id is not None
```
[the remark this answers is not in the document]
AlexLiu190625: Good catch — replaced the assert with an explicit `if trace_event.id is None: raise RuntimeError(...)` in 4235c385, so the check survives `python -O`. It lands in the same rollback-and-propagate path as the lease-changed error, so a flush that left the id unassigned now fails the event write instead of committing a NULL anchor.

src/xagent/web/services/task_lease_service.py
```diff
@@ -276,21 +378,43 @@ def lease_state_version_case(
...
+    apart on which rows clear their pointer -- a per-builder copy would let
+    one column retain a stale pointer the other clears, which would make
+    the recovery CAS fence's conjunction over both columns never match (see
+    test_lease_checkpoint_pointer_case_predicates_match in
+    test_task_lease_service.py).
+    """
+    return or_(
+        task_status_predicate.ne(TaskStatus.RUNNING),
+        Task.run_id.is_(None),
+    )
```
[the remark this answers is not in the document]
AlexLiu190625: Renamed in 1e2440c6 to `_checkpoint_pointer_clearing_predicate`, with the docstring stating the truth value directly (true when both pointer columns must be cleared). All call sites and the pairwise-predicate test reference updated.

tests/web/services/test_task_lease_recovery.py
```diff
@@ -128,6 +148,75 @@ def _recover_expired_task(db, task: Task) -> TaskStatus | None:
...
+    clause. Alembic's SQLite batch mode cannot add that FK without a full
+    table rebuild, so this asymmetric shape -- not the fresh create_all
+    schema every other fixture in this module uses -- is the only one on
+    which ``last_checkpoint_trace_event_id`` can point at a row that no
+    longer exists. Mirrors
+    tests/web/services/test_task_deletion_checkpoint_pointer.py's
+    sqlite_upgraded_session for the same reason, kept as a separate
+    module-local copy rather than a shared import so this module's
+    fixtures stay self-contained.
+    """
```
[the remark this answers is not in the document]
AlexLiu190625: Extracted in 1e2440c6: `tests/web/services/checkpoint_anchor_shared.py` now owns `build_upgraded_sqlite_engine()` (and the cross-dialect create-rule reset it depends on), following the same shared-module convention as `task_status_storage_shared.py`. Both test files keep thin local fixtures over the shared builder with their own teardown contracts.

tests/web/services/checkpoint_anchor_shared.py
```diff
@@ -0,0 +1,94 @@
...
+    never attempts the ALTER, since that dialect doesn't support it for
+    constraints) then honors the same cached decision and silently omits
+    the constraint from CREATE TABLE entirely -- it is never created
+    inline or via ALTER. Only a live process that create_all()s against
+    both dialects hits this (a real deployment binds one dialect for its
+    whole lifetime and never triggers it). Resetting the rule before each
+    create_all() call makes every fixture's result independent of what
+    dialect (if any) create_all() targeted earlier in the process.
+    """
+    for constraint in Base.metadata.tables["tasks"].constraints:
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in 9f6b98b1 — the module now imports the Task model at load time for its registration side effect (with a comment saying so), so `Base.metadata.tables["tasks"]` is present regardless of which test module imports the helper first. Verified by importing the module standalone in a fresh interpreter and calling the reset helper.

rogercloud:
ci is failing

src/xagent/web/api/trace_handlers.py
```diff
@@ -440,6 +452,105 @@ def _checkpoint_run_partition_filter(run_id: str | None) -> Any:
...
+            )
+
+        try:
+            decoded = decode_trace_event_data(
+                db,
+                task_id=self.task_id,
+                data=dict(row_data),
+                strict=True,
+            )
+        except CheckpointMessageDecodeError as exc:
```
rogercloud, line 589: **Major:** a decode failure on the anchored row is terminal here, but the legacy scan (lines 286-346, 358-379) tries every remaining candidate before giving up — and `_prune_checkpoint_history`'s docstring (753-754) says older rows are retained precisely so "an unreadable latest can fall back". Since the writer sets the anchor on every checkpoint (654), the anchor is the newest row in steady state, so a single corrupt newest checkpoint now permanently blocks resume (`a2a.py:468-474` maps `CheckpointCorruptError` to a terminal 400). Suggest returning the fallback sentinel on decode/read failure and letting the legacy scan run. No test covers "anchor unreadable, older readable row exists".

src/xagent/web/api/trace_handlers.py
```diff
@@ -440,6 +452,105 @@ def _checkpoint_run_partition_filter(run_id: str | None) -> Any:
...
+        The execution-identity check here is verification, not the legacy
+        scan's filtering: that scan excludes non-matching rows from its
+        candidate set via ``_checkpoint_execution_id_predicate`` before it
+        ever sees them, but the pointer names one row unconditionally, so
+        the row's own claimed identity (if it has one) has to be checked
+        against the caller's after the fact. A row with no execution
+        identity at all -- legacy rows written before the field existed --
+        still passes, matching the legacy scan's coalescing default of "".
+        """
+        pointer = (
```
rogercloud: **Major:** these two DB calls (`db.query(...).one_or_none()` and `db.get(...)` on 493) are outside any exception translation, but `_sync_load_latest_checkpoint`'s docstring (161-170) promises all DB failures become `CheckpointReadError` subclasses, and every other DB call here honours that (211-224, 270-277). A transient connection error escapes as a raw SQLAlchemy exception, bypassing the 503-retry contract at `a2a.py:455`. Please wrap both in the same try/except used elsewhere in this file.

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -0,0 +1,115 @@
...
+        return
+
+    bind = op.get_bind()
+    is_postgresql = bind.dialect.name == "postgresql"
+
+    if is_postgresql and FK_NAME in _fk_names():
+        op.drop_constraint(FK_NAME, TABLE, type_="foreignkey")
+
+    if COLUMN in _columns():
+        op.drop_column(TABLE, COLUMN)
```
rogercloud, line 131: **Major (reproduced):** this fails on any SQLite database created via `Base.metadata.create_all()` — `OperationalError: unknown column "last_checkpoint_trace_event_id" in foreign key definition`. `use_alter=True` does not prevent SQLite from rendering the FK inline at create time, the constraint drop above is gated on the postgresql branch only, and SQLite will not drop a column referenced by an inline FK. Suggest `op.batch_alter_table(TABLE, recreate="always")` for the SQLite drop. All `TestDowngradeSqlite` cases use the hand-built legacy-shape fixture, so this shape is untested.

src/xagent/web/services/task_lease_service.py
```diff
@@ -246,6 +267,87 @@ def has_recoverable_agent_checkpoint(
...
+    names, within the same partition (task, root scope, checkpoint type)
+    today's query has always used.
+
+    An unset pointer or a zero-row match is an authoritative absence: the
+    candidate's run has no checkpoint to recover to, so this is
+    NOT_RECOVERABLE (FAILED), not INDETERMINATE. Only two-or-more rows
+    sharing the same event_id inside this partition make the row's identity
+    itself ambiguous -- that is the case this resolver cannot decide and
+    defers to the next sweep instead of guessing or failing the task.
+    """
```
rogercloud, line 288: **Medium-low:** an ambiguous legacy pointer (`MultipleResultsFound`) is marked INDETERMINATE and left untouched, so every later sweep re-hits the same ambiguity forever, with only a `logger.warning` and no `ops_signals` degradation — while the reader side does register `CHECKPOINT_PK_ANCHOR_DANGLING` for its analogous failure. #1071 requires that zero or multiple matches fail closed and emit telemetry. Not a regression (the previous catch-all handler deferred forever too), but please register a degradation signal here to match the reader-side pattern.

src/xagent/web/api/trace_handlers.py
```diff
@@ -440,6 +452,105 @@ def _checkpoint_run_partition_filter(run_id: str | None) -> Any:
...
+        candidate.
+
+        The execution-identity check here is verification, not the legacy
+        scan's filtering: that scan excludes non-matching rows from its
+        candidate set via ``_checkpoint_execution_id_predicate`` before it
+        ever sees them, but the pointer names one row unconditionally, so
+        the row's own claimed identity (if it has one) has to be checked
+        against the caller's after the fact. A row with no execution
+        identity at all -- legacy rows written before the field existed --
+        still passes, matching the legacy scan's coalescing default of "".
```
rogercloud: **Low (docs):** this parity claim is not accurate. The scan's `coalesce(nullif(...))` yields NULL for rows with no execution identity, and `NULL = :execution_id` is never true, so the scan excludes them — whereas this path's Python check (`checkpoint_execution_id()` returning falsy `""`) includes them unconditionally. The two paths genuinely diverge (anchor is more permissive) and only the anchor side is tested. Please describe the divergence instead of claiming a match.

src/xagent/web/api/trace_handlers.py
```diff
@@ -526,14 +649,33 @@ def _save_trace_event(self, db: Session, event: CoreTraceEvent) -> None:
...
+                        db.query(Task.id).filter(Task.id == self.task_id).first()
+                        is not None
                     )
+                    if not task_still_exists:
+                        # The task row is gone, not merely leased elsewhere --
+                        # the same outcome the trace_events.task_id FK would
+                        # produce for the row itself. Skip quietly rather
+                        # than raising a "lease changed" error that isn't
+                        # what happened.
+                        logger.debug(
```
rogercloud, line 775: **Low-medium:** after this debug log the code falls through to `db.commit()` at 712, so an orphan `trace_events` row is persisted and success is reported even to `require_persisted=True` callers — the base behaviour always raised and rolled back on zero rows matched. Only reachable when the long-standing `trace_events.task_id` FK is unenforced (`PRAGMA foreign_keys=ON` is applied unconditionally here, so this needs an abnormal override), but suggest returning early after the log so the "skip quietly" comment matches the actual behaviour.

src/xagent/web/api/trace_handlers.py
```diff
@@ -440,6 +452,105 @@ def _checkpoint_run_partition_filter(run_id: str | None) -> Any:
...
+        row = db.get(DatabaseTraceEvent, pointer_id)
+        if row is None:
+            register_degradation(
+                CHECKPOINT_PK_ANCHOR_DANGLING,
+                f"task {self.task_id}: checkpoint pointer {pointer_id} has "
+                "no matching trace_events row; falling back to the legacy "
+                "scan",
+            )
+            return _NO_TRACE_EVENT_ANCHOR
+        clear_degradation(CHECKPOINT_PK_ANCHOR_DANGLING)
```
rogercloud: **Low (observability):** this clear only runs when the pointer resolves to a row, so a process where no task has an anchor leaves a previously-latched dangling signal set forever (`ops_signals` state is process-wide). The same problem affects `clear_degradation(CHECKPOINT_PRUNE_INTEGRITY_ERROR)` at 826, which the `if not stale_rows: return` at 816-817 skips — so a healthy steady state never clears it either. Suggest clearing both unconditionally at the top of their paths.

src/xagent/web/api/trace_handlers.py
```diff
@@ -637,11 +779,27 @@ def _prune_checkpoint_history(self, db: Session, data: Dict[str, Any]) -> None:
...
+            # older row.
+            anchor = (
+                db.query(Task.last_checkpoint_trace_event_id)
+                .filter(Task.id == self.task_id)
+                .one_or_none()
+            )
+            anchor_id = anchor[0] if anchor is not None else None
+            candidate_filters = list(partition_filters)
+            if anchor_id is not None:
+                candidate_filters.append(DatabaseTraceEvent.id != anchor_id)
```
rogercloud: **Low:** excluding the anchor from the candidate set (rather than ranking first and protecting it afterwards) shifts the `OFFSET(limit)` ranking of the remaining rows by one, so `limit + 1` rows survive whenever the anchor shares the partition. Either note the +1 in the docstring or restructure so the documented retention count is exact.

src/xagent/web/api/trace_handlers.py
```diff
@@ -518,6 +629,18 @@ def _save_trace_event(self, db: Session, event: CoreTraceEvent) -> None:
                 and self.build_id is None
                 and checkpoint_lease is not None
             ):
+                # Flush the pending insert so trace_event.id (the row's
+                # primary key) is assigned before the pointer UPDATE below
+                # reads it. autoflush is off for this session, so without
+                # this the exact-row anchor column would be built from an
+                # unassigned id and silently written NULL.
+                db.flush()
+                if trace_event.id is None:
```
rogercloud, line 730: **Test gap:** this guard (the `assert` -> explicit-raise fix from the earlier review round) is correct but untested — I removed the raise by mutation and the suite stayed green. Suggest a test that keeps `trace_event.id` as `None` after flush (mock the flush) and asserts the `RuntimeError`.

tests/web/services/test_task_deletion_checkpoint_pointer.py
```diff
@@ -0,0 +1,317 @@
...
+    assert session.query(Task).filter(Task.id == task_id).count() == 0
+    assert (
+        session.query(DatabaseTraceEvent)
+        .filter(DatabaseTraceEvent.task_id == task_id)
+        .count()
+        == 0
+    )
+
+
+def test_purge_task_rows_nulls_pointer_before_the_task_delete_flushes(
```
rogercloud, line 255: **Test quality:** this test is named for NULL-before-delete ordering but only asserts final state, and passes under a mutation that reverses the order (same applies to the test at 279). The ordering guarantee is only really covered by the `_under_fk_enforcement` siblings at 174/242, which need the fresh-schema FK that the upgraded shape lacks (153). Either rename these two to describe what they check, or record and assert call order.

tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -0,0 +1,471 @@
...
+    migrations_dir = Path(next(iter(migrations_pkg.__path__)))
+    path = migrations_dir / "versions" / f"{TARGET_REVISION}.py"
+    spec = importlib.util.spec_from_file_location(TARGET_REVISION, path)
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def _pre_migration_metadata() -> sa.MetaData:
```
rogercloud, line 51: **Test gap (same root cause as the downgrade finding):** every `TestDowngradeSqlite` case builds its schema from this hand-written `_pre_migration_metadata()` rather than `Base.metadata`, so the inline-FK shape that actually breaks `downgrade()` is never constructed. Please add a downgrade test over a fresh `Base.metadata.create_all()` database.

rogercloud (changes requested):
## Summary

This PR replaces the string-matched `last_checkpoint_event_id` lookup with a real referential
anchor: a new nullable `tasks.last_checkpoint_trace_event_id` FK column pointing at the exact
`trace_events.id` row that holds the checkpoint. The new column is dual-written under the
existing lease fence, read PK-first with a legacy-scan fallback for rows that predate the
column, and honoured across the deletion paths (pointers nulled before `trace_events` rows go
away) and in lease recovery, which now produces an explicit three-way
recoverable / not-recoverable / indeterminate verdict behind a CAS fence. Schema, Alembic
migration + backfill, and a large body of SQLite and PostgreSQL tests are included; the
"checkpoint staging joins a caller-owned transaction" piece of #1071 is explicitly deferred.

This is the right root-cause fix rather than a symptom patch — the old pointer genuinely had no
referential guarantee — and the test suite behind it is substantial and mostly well-targeted. I
ran the full affected suites on both SQLite and PostgreSQL and they pass. However, three
confirmed correctness defects remain (one reproduced end-to-end against the real Alembic
harness), plus design-level reservations noted below.

**1. The PK-first read path structurally breaks the checkpoint-retention guarantee documented in
the same file.** `_prune_checkpoint_history`'s docstring
(`src/xagent/web/api/trace_handlers.py:753-754`, pre-existing text) states that older
checkpoint rows are deliberately retained "so an unreadable latest can fall back". The legacy
scan implements exactly that: it walks every candidate and only gives up once all of them are
undecodable. The new `_load_pk_anchored_checkpoint` does not — once the anchor resolves to a
row, that row is terminal. Because the writer sets the anchor on *every* checkpoint write
(`trace_handlers.py:654`), the anchor *is* the newest row in steady state, so "newest checkpoint
is corrupt" now degrades from "resume from an older checkpoint" to a hard failure. That is a
regression against an invariant this file documents; see Major finding 1.

**2. The `tasks` ↔ `trace_events` FK ring buys less than it costs.** Closing the cycle forces
`use_alter=True` plus a named constraint, an explicit `foreign_keys=` on `TraceEvent.task`, a
dedicated SQLite `drop_all` test helper, prune having to exclude the anchor row *and* swallow
`IntegrityError`/`OperationalError`, null-before-delete in every deletion path, and
dialect-branching migration logic — and it still isn't there on Alembic-upgraded SQLite, which
is why this PR already implements a complete dangling-anchor fallback. The column is derived,
recomputable data whose non-dangling property the code already treats as optional. A plain
nullable integer with no DB constraint on any dialect would delete the drop-ordering
requirement, the prune `IntegrityError` branch, the SQLite/PG asymmetry, and the migration's
dialect branching in one move. Not blocking — the current design is deliberate and
well-tested — but worth reconsidering before the compatibility window closes.

**3. "Move both pointer columns together" is an unguarded invariant spread over ~6 call sites**
(the writer, the lease-acquire clear, the two retain-case builders, the turn-claim path, and
both deletion implementations). I investigated two ways to consolidate this and rejected both —
see the Simplification section. The right fix is a static guard test, not more abstraction; see
test-quality finding F14.

## Major

**M1 — `src/xagent/web/api/trace_handlers.py:523-543` — no fallback to older checkpoints on
anchor read failure.** When the anchored row fails to decode, this path raises
`CheckpointCorruptError` / `CheckpointUnavailableError` directly; the legacy scan
(`trace_handlers.py:286-346`, `358-379`) instead tries every remaining candidate first.
`a2a.py:468-474` maps `CheckpointCorruptError` to a terminal 400, so a single corrupt newest
checkpoint now permanently blocks resume for that task where it previously resumed from an
older row. Confirmed new in this PR (the base commit has no anchor path), and no test covers
"anchor unreadable, older readable row exists". Suggest: on decode/read failure of the anchored
row, return the fallback sentinel and let the legacy scan run — it will naturally re-examine
the anchor row first and then older ones, so the fast path is preserved and the retention
guarantee is restored.

**M2 — `src/xagent/web/api/trace_handlers.py:484-493` — two DB calls outside the error-translation
contract.** `_sync_load_latest_checkpoint`'s docstring (`161-170`) promises that every DB touch
is translated into a `CheckpointReadError` subclass, and every other DB call in this function
honours that (partition resolution `211-224`, paginated scan `270-277`). The new
`db.query(Task.last_checkpoint_trace_event_id)…one_or_none()` and
`db.get(DatabaseTraceEvent, pointer_id)` are unwrapped, so a transient connection failure here
escapes as a raw SQLAlchemy exception rather than `CheckpointUnavailableError`. The outer
`except BaseException` at `a2a.py:506` still performs cleanup and lease restoration correctly,
but the caller's documented 503-retry-friendly contract at `a2a.py:455` is bypassed. Suggest
wrapping both calls in the same try/except used elsewhere in this file.

**M3 — `src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py:114-115`
— `downgrade()` fails on any SQLite DB created via `create_all()`.** Reproduced end-to-end with
the real Alembic harness (`create_all` → `stamp` → `downgrade`):
`OperationalError: unknown column "last_checkpoint_trace_event_id" in foreign key definition`.
Cause: `use_alter=True` does not stop SQLite from rendering the FK inline at `create_all` time,
the constraint drop is gated on the `postgresql` branch only, and SQLite refuses to drop a
column referenced by an inline FK. Every `TestDowngradeSqlite` case uses the hand-built
legacy-shape fixture, which has no inline FK, so this shape is entirely untested. Suggest
`op.batch_alter_table(TABLE, recreate="always")` for the SQLite column drop, plus a downgrade
test built from a fresh `Base.metadata.create_all()` schema (see F13).

## Medium / Low

**F4 (medium-low) — `src/xagent/web/services/task_lease_service.py:284-301` and
`task_lease_recovery.py:58-65` — ambiguous legacy pointer emits no telemetry.** When a legacy
`event_id` resolves to multiple rows (`MultipleResultsFound`), the candidate is marked
INDETERMINATE and left untouched, so every subsequent sweep re-hits the identical ambiguity
forever — the PR's own
`test_ambiguous_legacy_checkpoint_skips_the_candidate_for_the_next_sweep` pins that behaviour.
Only a `logger.warning` is emitted; no `ops_signals` degradation is registered, whereas the
reader side does register `CHECKPOINT_PK_ANCHOR_DANGLING` for its analogous failure. #1071
explicitly requires that "zero or multiple matches fail closed and emit telemetry" — zero-match
is correct, multiple-match has no telemetry. To be fair this is not a regression: the
pre-existing catch-all handler in the batch loop produced the same log-and-defer-forever
outcome, and this PR merely made it an explicit named verdict. Suggest registering a
degradation signal for INDETERMINATE, mirroring the reader-side pattern.

**F5 (low) — `src/xagent/web/api/trace_handlers.py:480-482` — the docstring's parity claim is
false.** It says a row with no execution identity "still passes, matching the legacy scan's
coalescing default of ''". The scan's `coalesce(nullif(...))` evaluates to NULL for such rows,
and `NULL = :execution_id` is never true in SQL, so the scan *excludes* them; the anchor path's
Python check (`checkpoint_execution_id()` returning falsy `""`) *includes* them
unconditionally. The two paths genuinely diverge — the anchor path is more permissive — and only
the anchor side is tested (`tests/web/test_agent_checkpoint_stream.py:1428`). Practical impact
is narrow (web's `execution_id` equals the task id), but the stated justification is wrong;
please describe the divergence rather than claim parity.

**F7 (low-medium) — `src/xagent/web/api/trace_handlers.py:663-673` — silent success path persists
an orphan trace row.** When the pointer UPDATE matches zero rows because the task was deleted
concurrently, this logs at debug and falls through to `db.commit()` at `712`, reporting success
even to `require_persisted=True` callers and leaving a `trace_events` row behind. The base
commit always raised and rolled back for any zero-rows case. Scope is narrower than it first
looks: this is only reachable when the long-standing `trace_events.task_id → tasks.id` FK (not
the new anchor FK) is unenforced, and `PRAGMA foreign_keys=ON` is applied unconditionally to
every SQLite connection here, so in practice it needs an abnormal pragma override. Still worth
tightening as defense in depth — return early after the debug log so the "skip quietly" comment
matches what the code actually does.

**F8 (low) — `src/xagent/web/api/trace_handlers.py:502` and `:826` — two latched signals can never
clear.** `clear_degradation(CHECKPOINT_PK_ANCHOR_DANGLING)` only runs when a pointer resolves to
a row, so a process where no task has an anchor leaves a previously-latched dangling signal set
forever; `ops_signals` state is process-wide by its own docstring. Symmetrically,
`clear_degradation(CHECKPOINT_PRUNE_INTEGRITY_ERROR)` sits after the `if not stale_rows: return`
early exit at `816-817`, so a fully healthy steady state (nothing to prune) never clears it
either. Both show up as permanent `/health` noise. Suggest clearing both unconditionally near
the top of their code paths rather than only on the success branch that happens to reach them.

**F9 (low) — `src/xagent/web/api/trace_handlers.py:796-813` — retention count is off by one
relative to the docstring.** Excluding the anchor row from the *candidate set* (rather than
ranking first and protecting the anchor afterwards) shifts the `OFFSET` ranking of the remaining
rows by one, so `limit + 1` rows survive whenever the anchor is in the same partition as the
candidates. Either note the +1 in the docstring or restructure so the documented retention count
is exact.

## Non-blocking

**F10 — migration lock/duration behaviour on large PostgreSQL tables.** `create_foreign_key`
takes an ACCESS EXCLUSIVE lock and validates all existing rows synchronously, and the
correlated backfill subquery has no index on `event_id` (`task_id` selectivity probably keeps
this off a full scan in practice). Not a code defect, but for large production tables it would
help to note the expected duration/lock profile in the migration docstring, or to use
`NOT VALID` plus a separate `VALIDATE CONSTRAINT` for a zero-downtime rollout.

## Test quality

**A — `tests/web/services/test_task_deletion_checkpoint_pointer.py:209` and `:279` do not test
what their names claim.** Both are named for NULL-before-delete *ordering* but only assert final
state (pointer is NULL). I mutated the implementation to reverse the order and both still
passed. The real ordering guarantee is caught only by the `_under_fk_enforcement` siblings
(`:174`, `:242`), which require the fresh-schema SQLite FK to exist — absent in the
Alembic-upgraded shape (`:153`). Either rename these two to describe what they check
(pointer presence), or strengthen them to assert call order via a recording hook/mock.

**B — `src/xagent/web/api/trace_handlers.py:638-643` — the new `RuntimeError` guard has no test.**
This is the `assert` → explicit-raise conversion from the earlier bot review round; the fix
itself is correct, but I mutated the raise away and no test failed. Suggest a test that forces
`trace_event.id` to stay `None` after flush (mock the flush) and asserts the `RuntimeError`.

**F13 — `tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py:51-75` — the
downgrade fixture never exercises the real schema shape.** `TestDowngradeSqlite` builds its
schema from a hand-written `_pre_migration_metadata()` rather than `Base.metadata`, so the
inline-FK shape that actually breaks (M3) is never constructed. Add a downgrade test over a
fresh `Base.metadata.create_all()` database as part of the M3 fix.

**F14 — missing static guard for the paired-pointer invariant.** Nothing enforces that a
statement setting `last_checkpoint_event_id` also sets `last_checkpoint_trace_event_id`, across
`task_lease_service.py` (writer, lease-acquire clear, both case builders),
`task_orchestrator.py` (turn claim), `task_deletion.py`, and `admin_users.py`. This repo already
has a working precedent for exactly this style of check —
`test_lease_case_builders_are_the_only_case_calls_in_the_lease_service` in
`tests/web/services/test_task_status_storage.py` AST-parses the source and asserts an
unauthorized pattern is absent. An equivalent guard that scans those files for assignments or
dict keys touching either pointer column and asserts they always appear paired would make the
invariant cheap to keep, which matters most once these call sites drift apart during the
compatibility window.

## Prior review findings — all re-verified as FIXED

All four findings from the earlier automated review rounds were independently re-checked against
current HEAD:

1. **`assert` replaced with an explicit `RuntimeError`** — FIXED at
   `src/xagent/web/api/trace_handlers.py:638-643`. Note the connection to finding B above: the
   fix is correct but is not covered by any test.
2. **`_checkpoint_pointer_retention_predicate` → `_checkpoint_pointer_clearing_predicate`** —
   FIXED; the old name no longer appears anywhere in source or tests.
3. **Duplicated "SQLite without FK" fixture extracted** — FIXED; both consumers now call
   `build_upgraded_sqlite_engine()` from `tests/web/services/checkpoint_anchor_shared.py`.
4. **`Task` model import so `Base.metadata.tables["tasks"]` is registered** — FIXED; the
   module-level import achieves the same registration guarantee as the suggested
   inside-function import.

## Simplification

Applying a simplification lens here does *not* produce a "lean already" verdict — the
paired-pointer update pattern is genuinely duplicated across ~6 call sites. But both obvious
consolidations turn out to be net-negative and I am explicitly not recommending them:

- **Sharing one pointer-clearing helper across the deletion and lease paths**: only 2 of the 4
  sites are true duplicates, and importing `task_lease_service` into `task_deletion.py`
  introduces a real one-way module dependency for near-zero line savings.
- **Merging the two retain-case builders into a single dict-returning function**: both are
  called independently by `test_lease_checkpoint_pointer_case_predicates_match`, whose entire
  design intent is to verify the two builders never drift apart. Merging them deletes that
  regression test's reason to exist.

So the duplication is load-bearing, and the correct response is to *pin* the invariant rather
than abstract it away — which is exactly F14 above. That is the actionable simplification-lens
outcome: one small AST guard test replaces the temptation to refactor.

## Testing performed

- Ran the PR's core suites on SQLite — `tests/web/test_task_lease_service.py`,
  `tests/web/services/test_task_lease_recovery.py`,
  `tests/web/services/test_task_deletion_checkpoint_pointer.py`,
  `tests/web/test_agent_checkpoint_stream.py`, `tests/web/test_db_teardown.py`,
  `tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py` — all pass
  (PostgreSQL-gated tests skipped in this run).
- Set up a local PostgreSQL test database and installed `psycopg2`/`asyncpg`/`greenlet`, then
  re-ran the previously-skipped PostgreSQL paths (`_on_postgres` deletion tests, migration
  `TestUpgradePostgres`) — all pass.
- Ran the broader `tests/web` suite for regressions — all pass except pre-existing,
  PR-unrelated SVG-preview failures caused by a missing native `cairo` library in this
  environment.
- Independently reproduced M3 end-to-end with the real Alembic harness, and confirmed findings
  A and B by mutation (reversing the delete ordering / removing the `RuntimeError`, both of
  which left the suite green).

## Recommendation

**Request changes** — narrowly, on M1, M2 and M3. The approach is sound and the test coverage is
real rather than ceremonial, so this is close to mergeable, but M3 is a reproduced migration
failure on the default SQLite deployment shape, and M1 silently trades away a documented
resume-safety guarantee at exactly the moment it matters most (newest checkpoint corrupt).
Both are small, localized fixes. Everything below the Major section can land as follow-ups if
you prefer, though F4's telemetry gap is called out directly in #1071's acceptance criteria and
is a two-line change.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in dc1f5443: a payload decode/read failure on the anchored row now returns a fallback carrier so the legacy scan runs, while an identity/verification mismatch stays a terminal CheckpointCorruptError. The carrier also hands the scan the row's verdict flag — a row with no execution identity is anchored but filtered out of the scan's candidate set, so a bare defer would have collapsed an unreadable checkpoint into a None "no checkpoint" answer, which the module contract forbids. Covered: undecodable anchor + older readable row returns the older snapshot; undecodable + no rows is corrupt; generic failure + no rows is unavailable (retryable).

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 27aff7d0: both the pointer query and the row fetch are wrapped in the same translation the paginated scan uses, each with its own degradation detail, so a transient failure surfaces as CheckpointUnavailableError instead of a raw driver exception. Each wrap is pinned by its own injected-failure test.

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Fixed in affd86b1 after reproducing your exact error. The SQLite branch drops the column through op.batch_alter_table(TABLE, recreate="always"); a full before/after schema diff confirms indexes, primary key, unique constraints, remaining foreign keys (including channel_id's ON DELETE SET NULL) and all rows survive the rebuild, with only the anchor FK removed — and the migration runner already disables SQLite FKs for exactly this rebuild pattern and re-checks foreign_key_check afterwards. A new downgrade test builds from a real Base.metadata.create_all() schema and asserts the inline FK is present before downgrading.

src/xagent/web/services/task_lease_service.py
AlexLiu190625: Fixed in 3786f922: the INDETERMINATE verdict registers checkpoint_legacy_pointer_ambiguous. The clear runs once per recovery drain (top of recover_expired_task_leases_until_cutoff) rather than per candidate or per page — at either finer grain a later clean candidate would erase the signal inside the same sweep. That placement is pinned by a two-candidate test (ambiguous then clean, parametrized over batch sizes) asserting the signal survives the drain.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 23b5d669: the docstring now states the divergence instead of claiming parity — the scan's coalesce/nullif predicate yields NULL for a row with no execution identity so the scan excludes it, while the anchor path includes it, deliberately the more permissive of the two since those are exactly the legacy rows the backfill anchors.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 1351386d: the branch rolls back before returning, so no orphan trace_events row is committed, and it re-raises for require_persisted=True callers — mirroring how the trace_events.task_id FK violation is classified a few lines below. Parametrized tests cover both caller modes with zero-residue assertions, and the lease-changed discrimination stays separately pinned.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 3786f922: the dangling clear moved to the top of the anchored read (any non-dangling attempt clears it) and the prune clear moved above the nothing-stale early exit, so a healthy steady state clears both. Each clear path has its own test.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in 23b5d669 by restructuring rather than documenting: the anchor exclusion moved out of the candidate filter into the materialised id list, so ranking happens first and the retention count is exact when the anchor is in the window. The only remaining limit+1 case — the anchor ranking outside the window — is stated in the docstring, and the forward-defense annotation on the exclusion is preserved.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Covered in 413ca036: a test stubs flush to a no-op so trace_event.id stays None, asserts the RuntimeError with its exact message, and asserts zero residue (no trace_events row, NULL pointer) after the rollback.

tests/web/services/test_task_deletion_checkpoint_pointer.py
AlexLiu190625: Fixed in 413ca036: both tests now record statement order through a before_cursor_execute listener and assert the pointer UPDATE precedes the trace_events DELETE — red under order reversal even on the FK-less SQLite shape, which previously had no ordering coverage. The final-state assertions are kept alongside.

tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Fixed in affd86b1: TestDowngradeSqliteOverCreateAllSchema builds from Base.metadata.create_all(), asserts up front that the inline FK is really present (so the fixture cannot silently test the wrong shape), then downgrades and asserts the column is gone and seeded rows survive. It calls the shared create-rule reset first, without which a prior PostgreSQL create_all in the same process would leave the inline FK out of the SQLite DDL.

AlexLiu190625:
On the `tasks` ↔ `trace_events` cycle: agreed that the ring's cost is real and concentrated exactly where you list it — `use_alter=True`, the named constraint, the explicit `foreign_keys=` on `TraceEvent.task`, the dedicated drop helper, prune's exclusion and its `IntegrityError`/`OperationalError` branch, null-before-delete in both deletion implementations, and the migration's dialect branching. The plain-nullable-integer alternative was weighed during design review and the ring was kept for one reason: removing it means a full SQLite table rebuild to drop the constraint from every already-created database, and that blast radius was judged worse than carrying the asymmetry through the compatibility window. The window is where the cost is temporary — once the legacy string column retires, a dangling pointer becomes a corruption verdict rather than a fallback, and several of the branches above collapse on their own.

So this PR retains the current design, and the reconsideration is recorded against the compatibility-window-close scope, where dropping the constraint and the legacy column can be done as one change with a single migration instead of two.

Two items had no inline anchor: the migration's PostgreSQL lock and duration profile is now noted in its module docstring (`create_foreign_key` takes ACCESS EXCLUSIVE and validates synchronously; the backfill's correlated subquery matches on `event_id`, which is unindexed), with no `NOT VALID` restructure in this change; and the paired-pointer static guard landed as `tests/web/services/test_checkpoint_pointer_pairing.py`, following the existing AST-guard precedent — it matches mutation carriers only (dict-literal keys, call keywords, subscript assignments) and pairs them per statement block, since both branches of the lease claim write the pair independently and a function-level check would stay green if one branch lost a column.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Correction to my earlier reply: the healthy-anchor test pinned that a clear happens but not its placement — moving the clear back after the pointer resolves left the suite green, since that test's task carries a valid anchor. Added `test_database_trace_handler_anchorless_read_clears_the_dangling_signal` (unset pointer + pre-latched signal → cleared), which is red under the old placement; that is the exact no-anchor process shape this finding described.

src/xagent/web/api/trace_handlers.py
```diff
@@ -440,6 +471,167 @@ def _checkpoint_run_partition_filter(run_id: str | None) -> Any:
...
+            )
+            logger.warning(
+                "Checkpoint pointer %s row for task %s failed to decode; "
+                "falling back to the legacy scan",
+                pointer_id,
+                self.task_id,
+                exc_info=True,
+            )
+            return _AnchorFallback(generic_failure=True)
+        clear_degradation(CHECKPOINT_LOAD_UNAVAILABLE)
```
rogercloud, line 629: **Medium -- `CHECKPOINT_DECODE_FALLBACK` can latch here permanently.** A successful anchored decode clears `CHECKPOINT_LOAD_UNAVAILABLE` but not `CHECKPOINT_DECODE_FALLBACK`. The only clear for that signal is inside the legacy scan loop (line 348), which this success return bypasses. Once the signal is set -- by the scan, or by this function's own generic-failure branch above -- and the anchor then succeeds consistently (the normal steady state), `/health` keeps reporting decode fallback for the life of the process with no path back. Reads stay correct, so this is observability-only, but it's a permanent false positive once triggered. Suggest adding `clear_degradation(CHECKPOINT_DECODE_FALLBACK)` next to the existing clear.

src/xagent/web/api/trace_handlers.py
```diff
@@ -671,6 +933,28 @@ def _prune_checkpoint_history(self, db: Session, data: Dict[str, Any]) -> None:
...
+            # FK for this column and cannot raise this; a freshly created
+            # SQLite database (create_all, e.g. tests) does have the FK and
+            # can. Under stricter isolation levels the same race can instead
+            # surface as psycopg2's SerializationFailure or DeadlockDetected,
+            # both of which SQLAlchemy wraps in OperationalError, not
+            # IntegrityError -- catch both so this retention path degrades
+            # the same way regardless of which one the database raises.
+            db.rollback()
+            register_degradation(
+                CHECKPOINT_PRUNE_INTEGRITY_ERROR,
```
rogercloud: **Low -- signal name over-claims "integrity".** This handler catches `(IntegrityError, OperationalError)` but labels every failure `CHECKPOINT_PRUNE_INTEGRITY_ERROR`. `OperationalError` also covers lock timeouts, dropped connections and SQLite `database is locked`, which an operator reading `/health` would misread as data-integrity trouble. Either classify on `exc.orig`/pgcode, or rename the signal to something like `CHECKPOINT_PRUNE_FAILED`.

src/xagent/web/api/trace_handlers.py
```diff
@@ -526,10 +730,43 @@ def _save_trace_event(self, db: Session, event: CoreTraceEvent) -> None:
...
+                    if not task_still_exists:
+                        # The task row is gone, not merely leased elsewhere --
+                        # the same outcome the trace_events.task_id FK
+                        # produces for the row itself wherever that FK is
+                        # enforced. Discard the staged row so this path
+                        # leaves nothing behind, then classify it exactly as
+                        # that FK violation is classified below: a required
+                        # event still fails loudly, a best-effort event is
+                        # dropped quietly.
+                        db.rollback()
```
rogercloud, line 764: **Low / note -- no change requested.** This new rollback matches how the method already behaves: `_save_trace_event` managed transaction boundaries on a caller-supplied session before this PR too (commit + rollbacks elsewhere in this same method), so this isn't a new class of risk, and it's safe today (single caller, session untouched afterwards). Flagging only so the deferred follow-up -- "checkpoint staging joins a caller-owned transaction with no implicit commit" -- accounts for all of this method's boundary actions, not just this one.

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -0,0 +1,136 @@
+"""add exact-row checkpoint anchor to tasks
+
+Adds ``tasks.last_checkpoint_trace_event_id``, an integer pointer to the
+``trace_events`` row that ``last_checkpoint_event_id`` (the legacy UUID
+string column) names. On PostgreSQL the column is backed by a named foreign
+key; on SQLite the migration adds the column only -- ``ALTER TABLE ADD
+CONSTRAINT`` is not renderable through Alembic's SQLite batch mode without a
+full table rebuild, so an upgraded SQLite database has no DB-level FK here.
+A database built fresh through ``Base.metadata.create_all()`` (new installs,
+tests) gets the FK from the model directly, regardless of dialect.
```
rogercloud: **Low (docs).** The rebuild explanation above is accurate. What's missing is that the resulting fresh-install vs upgraded-install FK divergence is permanent under the current migration set -- this revision is the head and no later migration reconciles it. One sentence saying so would stop a future reader assuming reconciliation is pending.

tests/web/services/checkpoint_anchor_shared.py
```diff
@@ -0,0 +1,99 @@
...
+        if getattr(constraint, "name", None) == CHECKPOINT_ANCHOR_FK_NAME:
+            constraint._create_rule = None
```
rogercloud, line 76: **Nit.** This mutates a private SQLAlchemy attribute, and at least 4 call sites across 3 test directories depend on it. If a future SQLAlchemy release renames or removes `_create_rule`, this line silently becomes a no-op and those tests quietly stop testing anything. Consider `assert hasattr(constraint, "_create_rule")` immediately before the assignment so the upgrade fails loudly instead.

rogercloud (approved):
## Re-review (round 2)

### Update summary

Since the last review (head `d55af50b` → `c1da9351`, base unchanged at `c474caa5`) the author reworked the PK-anchor read path so that decode failures return an internal fallback sentinel and fall through into the legacy scan loop instead of raising, wrapped the two previously-unguarded anchor-path DB calls in the file's standard exception translation, and rewrote the SQLite branch of the migration's `downgrade()` to use `op.batch_alter_table(..., recreate="always")` so it survives a fresh `create_all()` schema. Several ops-signal clears were relocated so they fire on every healthy attempt rather than only on the fully-successful path, prune's anchor exclusion moved to after the ranked/OFFSET query, and the deleted-task race branch now rolls back and re-raises for `require_persisted=True`. On the test side, `tests/web/test_agent_checkpoint_stream.py` roughly doubled (~563 → ~1114 lines) to cover each fix plus new edge cases, two ordering-named tests were upgraded to assert real statement order via a `before_cursor_execute` listener, and a new AST-based static guard `tests/web/services/test_checkpoint_pointer_pairing.py` enforces that the two pointer columns always move together across all call sites. Diff grew 21 files / +2654 −100 → 22 files / +3807 −107.

### Prior findings — all 13 verified FIXED

| # | Prior finding | Sev | Status | Evidence |
|---|---|---|---|---|
| M1 | Anchor read had no fallback to older checkpoints on decode failure | Major | **FIXED** | `_load_pk_anchored_checkpoint` returns an `_AnchorFallback` sentinel on decode/generic failure; caller falls through into the legacy scan. Covered by `test_..._load_pk_anchor_undecodable_payload_falls_back_to_older_row`. Identity/verification mismatches correctly stay terminal. |
| M2 | Two anchor-path DB calls unwrapped by exception translation | Major | **FIXED** | Both now wrapped → `CheckpointUnavailableError`, matching the file's existing pattern. |
| M3 | SQLite `downgrade()` broke on any fresh `create_all()` DB | Major | **FIXED** | SQLite branch now uses `batch_alter_table(recreate="always")`. Re-reproduced the previously-failing sequence: it now succeeds, and a before/after schema diff shows other FKs, indexes and the PK preserved. New `TestDowngradeSqliteOverCreateAllSchema`, which asserts up front that the fixture really has the inline FK. |
| F4 | Ambiguous legacy pointer emitted no telemetry | Med-Low | **FIXED** | New `CHECKPOINT_LEGACY_POINTER_AMBIGUOUS` signal, set on ambiguity and cleared once per recovery drain at the top of the batch function (not per-candidate), so a clean drain reliably clears a latched signal. |
| F5 | Docstring falsely claimed anchor/scan identity handling matched | Low | **FIXED** | Docstring now states the divergence (scan excludes NULL-identity rows via SQL semantics, anchor path includes them) and why it's acceptable. |
| F7 | Deleted-task race left an orphan `trace_events` row committed as "success" | Low-Med | **FIXED** | Branch rolls back before returning and re-raises for `require_persisted=True`; new parametrized test over both values asserts zero residue rows. |
| F8a | Dangling-anchor signal couldn't self-clear | Low | **FIXED** | Clear moved to the top of the anchored-read function, so it fires on every attempt. Test covers the no-anchor-at-all case. |
| F8b | Prune-integrity signal couldn't self-clear | Low | **FIXED** | Clear moved above the "nothing to prune" early exit; test covers the nothing-stale case. |
| F9 | Prune retained limit+1 rows when the anchor fell inside the ranking window | Low | **FIXED** | Anchor exclusion now applied to the materialized id list after the ranked/OFFSET query; exact-count test added. The anchor-outside-window retention remains, which is the documented, deliberate exception. |
| A | Two ordering-named tests only asserted final state | Test | **FIXED** | Both now record real statement order with a `before_cursor_execute` listener and assert the pointer UPDATE precedes the `trace_events` DELETE. Mutation-tested: reversing the order fails both. |
| B | `RuntimeError` guard for a null PK after flush was untested | Test | **FIXED** | New test stubs flush to a no-op, asserts the exact message and zero residue. Mutation-tested: removing the raise fails it. |
| F13 | Downgrade tests never exercised the fresh-`create_all()` shape | Test | **FIXED** | Same `TestDowngradeSqliteOverCreateAllSchema` class as M3. |
| F14 | No static guard that the two pointer columns move together | Test | **FIXED** | `tests/web/services/test_checkpoint_pointer_pairing.py` is a genuine AST source scan over all subject modules with negative controls. Mutation-tested: dropping one column at one call site fails the guard. |

No finding was waived or answered with a materially different design — every one was fixed as recommended.

### Design verdict (fresh pass): Sound — acceptable with reservations

Upgraded from the prior round: the retention-guarantee regression that drove the earlier reservation is gone, and the pointer-pairing duplication is now backed by a real static guard rather than convention. Remaining design-level notes, none blocking:

- **Lease-recovery INDETERMINATE is a terminal state with no bound or escalation.** Pre-existing, not a regression: the base commit stalled the same way via an uncaught exception swallowed by a generic handler. This PR replaces an exception-driven stall with a clean return-value stall plus telemetry, which is strictly better. Worth a follow-up issue about bounding/escalating the state, not a change here.
- **The SQLite FK divergence between fresh installs and upgraded installs is a known, accepted trade-off** — but the migration docstring should say the divergence is permanent under the current migration set (see N4).
- **Pointer-pair duplication across ~6 call sites** is now genuinely enforced by the AST guard; the prior round's concern here is resolved.

### New findings

#### Medium

**N1 — `CHECKPOINT_DECODE_FALLBACK` can latch permanently on the healthy anchor path.**
`src/xagent/web/api/trace_handlers.py:619` — a successful anchored decode clears `CHECKPOINT_LOAD_UNAVAILABLE` but not `CHECKPOINT_DECODE_FALLBACK`. The only clear for that signal in the codebase is inside the legacy scan loop (`trace_handlers.py:348`), which the anchor path's success return bypasses entirely. So once the signal is set — by the scan, or by the anchor path's own generic-failure branch at ~`:607` — and the anchor then starts succeeding consistently (the normal, desired steady state), `/health` reports decode fallback for the remaining life of the process with no path back. Checkpoint reads themselves stay correct, so this is observability-only, but it's a permanent false positive once triggered. Suggest adding `clear_degradation(CHECKPOINT_DECODE_FALLBACK)` alongside the existing clear at `:619`.

#### Low

**N2 — Prune failure signal over-claims "integrity".**
`src/xagent/web/api/trace_handlers.py:948-957` — the handler catches `(IntegrityError, OperationalError)` and labels everything `CHECKPOINT_PRUNE_INTEGRITY_ERROR`. `OperationalError` also covers lock timeouts, dropped connections and SQLite `database is locked`, which would be reported to an operator as an integrity problem. Diagnostic precision only. Either narrow the classification by inspecting `exc.orig`/pgcode, or rename the signal to something like `CHECKPOINT_PRUNE_FAILED`.

**N3 — Note (not a defect): the new rollback follows the method's pre-existing transaction ownership.**
`src/xagent/web/api/trace_handlers.py:753` — `_save_trace_event` already managed transaction boundaries on a caller-supplied session before this PR (the base commit has one commit and two rollbacks on other branches, `:821`/`:842`), so this new rollback is consistent with existing behaviour, not a new class of risk, and it's safe today (single caller, session untouched afterwards). Flagging only so the PR's own deferred follow-up — "checkpoint staging joins a caller-owned transaction with no implicit commit" — accounts for *all* of this method's boundary actions, not just the new one.

**N4 — Migration docstring should state the FK divergence is permanent.**
`src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py:5-10` — the explanation of why adding the constraint to an already-upgraded SQLite database needs a full table rebuild is technically accurate. What's missing is that the resulting fresh-install vs upgraded-install divergence is permanent under the current migration set: this revision is the head and no later migration reconciles it. One sentence would prevent a future reader assuming reconciliation is pending.

**N5 — The teardown rule is documented but unenforced (prospective).**
`tests/shared/db_teardown.py:21` — the docstring tells authors to use the shared helper instead of raw `Base.metadata.drop_all()` when a test persists a checkpoint anchor, with no lint/CI backing. There are 85 direct `metadata.drop_all()` call sites across 43 files today, and **none of them currently populate the anchor column**, so nothing is broken right now. The gap is that nothing stops a future edit from adding checkpoint persistence to one of those files while leaving the unguarded teardown in place. A cheap AST/grep guard in the same spirit as `test_checkpoint_pointer_pairing.py` would close it.

#### Nits

**N6** — `clear_degradation(CHECKPOINT_PK_ANCHOR_DANGLING)` (`trace_handlers.py:525`) is only reachable from the anchor-read path, which is gated on `self.build_id is None`, so build-scoped reads never touch the signal. Harmless: build-scoped checkpoints structurally never use the PK-anchor mechanism (per the same file's partitioning docstring), so there's no missed set or clear — purely a scope note.

**N7** — `tests/web/services/checkpoint_anchor_shared.py:65` mutates the private SQLAlchemy attribute `constraint._create_rule` to force DDL-compiler cache invalidation, relied on by at least 4 call sites across 3 test directories. Well documented, but if a future SQLAlchemy release renames or removes the attribute this would silently no-op rather than fail, quietly hollowing out those tests. An `assert hasattr(constraint, "_create_rule")` before the mutation makes that a loud failure.

*Considered and dropped:* the anchor path's identity-mismatch branch having no fallback (unlike the decode branch). Verified unreachable — every write path that rotates `Task.run_id` clears the pointer columns atomically in the same UPDATE via a shared predicate, with a dedicated regression test guarding against drift. Speculative only.

- **[DELETE] `tests/web/test_agent_checkpoint_stream.py:~1290`** — `test_database_trace_handler_load_pk_anchor_mismatch_raises_corrupt` faults `build_id` and `run_partition` simultaneously and asserts only `pytest.raises(CheckpointCorruptError)`; the later parametrized `test_..._single_fault_raises_corrupt` (~`:1371`) already covers both conjuncts, and every other one, with the identical assertion. The single-fault test's own docstring says the double-fault version "cannot prove any single conjunct is doing real work." Zero coverage loss, ~38 lines.
- **[SHRINK] `tests/web/test_agent_checkpoint_stream.py:~3046` and `:~3096`** — `test_..._healthy_anchor_read_clears_the_dangling_signal` and `test_..._anchorless_read_clears_the_dangling_signal` share identical setup, teardown and assertions, differing only in whether the anchor is populated. Collapse into one `@pytest.mark.parametrize("anchor_set", [True, False])`. No assertion lost, ~35-45 lines.

Net: roughly **75-85 lines** removed with no loss of coverage.

### Testing performed

- Re-ran the previously-relevant suites on SQLite (pass), and re-provisioned the local PostgreSQL test database (psycopg2 / asyncpg / greenlet) to run the PostgreSQL-gated paths (pass).
- All 13 prior findings were re-verified against the current source rather than against commit messages. Several were additionally confirmed by mutation testing — reintroducing the bug, confirming the guarding test fails, reverting — specifically findings A, B, F14, and by empirical reproduction for M3, where the previously-failing migration sequence was actually re-run and the resulting schema diffed.
- Every finding in this round's fresh pass was independently re-verified the same way before being written up, including the dropped one.

### Recommendation: **Approve with comments**

All three previously-Major findings are genuinely fixed — including M3, where the fix was empirically re-reproduced rather than read — and the ten remaining prior findings are fixed as originally recommended, with none waived or reshaped. The test suite grew in the right direction: real ordering assertions, a mutation-verified null-PK guard, and a static AST guard that turns the pointer-pairing convention into something CI can enforce. The design verdict improves to sound.

Nothing new blocks: N1 is the only finding above Low and it is observability-only, with no impact on checkpoint correctness, durability, or data integrity — a stale `/health` flag is worth fixing but is not worth another round-trip. N2/N4/N5 are diagnostics, docs and prospective hygiene; N3 is a heads-up for the already-planned transaction follow-up; N6/N7 are nits. I'd like N1 and the S1/S2 cleanups picked up before merge, but I'm not gating on them.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Fixed in a03e0808: the decode-fallback clear now sits beside the load-signal clear, after the row decodes and before both post-decode returns — the success return and the decoded-but-no-snapshot return, so it does not repeat the earlier clear-only-on-success shape. Placement is parity with the scan's own clear on a successful decode rather than the top of the function, since that one reports on the pointer every attempt re-reads. Worth stating explicitly: the registry is process-wide and last-writer-wins, so this trades a permanently stuck flag for a short-lived one a coarse health poll can miss — recorded in the commit body.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Renamed to CHECKPOINT_PRUNE_FAILED in a03e0808. Chose the rename over classifying on exc.orig: /health publishes names only, the operator's response to a restrict violation, a lock timeout and a dropped connection is the same, and the cause already rides in the detail string and the logged traceback. There is also no precedent for orig/pgcode classification in this codebase, and nothing keys on the old name — the tests import the constant.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Noted, and the point about the deferred staging follow-up is taken: it needs to account for every transaction boundary this method already owns — the existing commit and two rollbacks as well as the one added here — not only the new one. Recorded against that follow-up.

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Added in a03e0808: the docstring now states that the divergence is permanent under the current migration set (this revision is the head and nothing reconciles the two shapes), and names what protects the upgraded shape in the meantime — the deletion paths clear both pointer columns before the trace rows go away.

tests/web/services/checkpoint_anchor_shared.py
AlexLiu190625: Added in a03e0808: an assert guards the attribute before the mutation, so a SQLAlchemy rename fails loudly at the fixture instead of silently no-opping and hollowing out the call sites.

PR #1145, opened 2026-08-05
fix: render real copy in home route's SSG loading state

yiboyasss:


frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +88,26 @@ export function AuthGuard({ children }: AuthGuardProps) {
 
   const { t } = useI18n()
   if (isLoading) {
+    // The home route's loading state is also what static export (SSG) freezes
```
rogercloud: **F1 (minor) — comment overstates the guarantee.**

"Anonymous crawlers ... only ever see this markup" isn't quite accurate: the `mounted`-gated effect above (lines 28-34) pushes any client that hydrates and runs JS to `/login` once `isLoading` resolves and the user is unauthenticated. So a JS-executing crawler will not reliably keep seeing this hero — only non-JS / raw-HTML fetchers of the frozen `out/index.html` are guaranteed to.

The markup itself is fine; just tighten the wording, e.g. "Static / non-JS fetchers of the frozen HTML see only this markup, so it carries real copy about what the app does instead of a bare spinner."

rogercloud (commented):
## Summary

This PR makes the statically exported home page carry real content instead of a bare spinner. `AuthGuard` now special-cases `pathname === "/"` inside its `isLoading` branch and renders a hero screen with the app title/subtitle (reusing the existing `home.hero.title` / `home.hero.subtitle` i18n keys and `getBrandingFromEnv()`); every other route keeps the generic spinner. Because auth only resolves client-side, Next.js static export freezes exactly this JSX into `out/index.html`, so non-JS fetchers of that file — including automated brand-verification crawlers — now see meaningful copy. `layout.tsx` additionally fills in `applicationName` and an `openGraph` block (`siteName`/`title`/`description`) derived from the same `branding` object.

## Verdict: acceptable with reservations

The mechanism works: a real `next build` confirms the hero title and subtitle land in `out/index.html`, the branch is correctly scoped to `/` only (other routes are served from the separate static shell and are unaffected), and the copy is reused from existing i18n keys rather than invented. Two reservations, both left as findings below: the new code comment overstates what the mechanism guarantees (F1), and there is no test or CI assertion guarding the content this PR exists to produce (F2).

Design note: the structurally cleaner fix — making `/` a genuinely public route so it never routes through the loading/redirect path at all — is a substantially larger change and is reasonably deferred. Not requested here.

## Findings

| # | Severity | Location | Issue |
|---|---|---|---|
| F1 | minor | `frontend/src/components/auth/auth-guard.tsx:91` | Comment overstates the guarantee ("only ever see this markup") |
| F2 | major | `frontend/src/components/auth/auth-guard.tsx:96` | No test or CI content assertion for the new branch |
| F3 | minor | `frontend/src/app/layout.tsx:17` | `openGraph` omits `images`; no `metadataBase` anywhere |
| F4 | minor | `frontend/src/components/auth/auth-guard.tsx:104` | SSG hero omits the brand logo the real home page renders |
| F5 | minor | `frontend/src/app/layout.tsx:16` | New metadata fields have no test coverage |
| F6 | nit | `frontend/src/components/auth/auth-guard.tsx:97` | `getBrandingFromEnv()` called inline in render, unlike module-scope usage elsewhere |

F2 is the one worth acting on before merge: CI currently asserts only that `out/index.html` exists, so the exact content guarantee this PR was written to establish can silently regress.

Lean already.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +88,26 @@ export function AuthGuard({ children }: AuthGuardProps) {
...
+    // into out/index.html, since auth can only resolve client-side. Anonymous
+    // crawlers (incl. Google OAuth brand verification) only ever see this
+    // markup, so it carries real copy about what the app does instead of a
+    // bare spinner.
+    if (pathname === "/") {
```
rogercloud, line 104: **F2 (major) — no coverage for the branch this PR exists to add.**

Nothing guards the content guarantee today: `.github/workflows/ci.yml` only does `test -f out/index.html` plus a shell-file count, and `frontend/src/app/layout.test.tsx` mocks out `AuthGuard` entirely, so the real component is never rendered in any test.

That means a silent regression — this branch deleted, the condition inverted, or the copy reverted to a spinner — ships green. A unit test is cheap here (~15 lines): `layout.test.tsx` already shows the `next/navigation` pathname-mock pattern, so render the real `AuthGuard` with `pathname="/"` and a mocked `useAuth` returning `isLoading: true`, and assert the hero title/subtitle are present. Optionally also assert a non-`/` pathname still renders the generic spinner.

A content grep in the CI static-export step (e.g. `grep -q "<h1" out/index.html`) would be a reasonable complement, but the unit test is the higher-value one.

frontend/src/app/layout.tsx
```diff
@@ -13,6 +13,12 @@ const branding = getBrandingFromEnv();
 export const metadata: Metadata = {
   title: branding.appName,
   description: branding.description,
+  applicationName: branding.appName,
+  openGraph: {
```
rogercloud, line 18: **F3 (minor) — `openGraph` is missing `images`.**

For a crawler / brand-verification use case, the image is the highest-value OpenGraph field, and it's the one omitted. Consider at least:

```ts
openGraph: {
  siteName: branding.appName,
  title: branding.appName,
  description: branding.description,
  images: [branding.logoPath],
},
```

Note that no `metadataBase` is set anywhere in the app, so a relative image URL will make Next.js warn and resolve against a fallback origin. Not a build failure, but worth setting `metadataBase` alongside this (or using an absolute URL) so the field is actually usable. `url` and `type` are also unset — lower priority.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +88,26 @@ export function AuthGuard({ children }: AuthGuardProps) {
...
+      const branding = getBrandingFromEnv()
+      return (
+        <div className="min-h-screen bg-[#0D1117] flex items-center justify-center px-6">
+          <div className="text-center max-w-xl">
+            <h1 className="text-white text-2xl font-bold mb-3">
+              {t('home.hero.title', { appName: branding.appName })}
+            </h1>
+            <p className="text-[#8B949E] mb-6">{t('home.hero.subtitle')}</p>
```
rogercloud: **F4 (minor) — the frozen hero has no logo.**

The real home page renders the brand mark (`page.tsx`, ``), but this SSG-visible variant is text-only. Brand verification generally looks for both the app name *and* a logo image, so including it here makes the fix meaningfully more effective — and it costs one `` tag with the `branding` object already in scope.

frontend/src/app/layout.tsx
```diff
@@ -13,6 +13,12 @@ const branding = getBrandingFromEnv();
 export const metadata: Metadata = {
   title: branding.appName,
   description: branding.description,
+  applicationName: branding.appName,
```
rogercloud, line 17: **F5 (minor) — new metadata fields are untested.**

Neither `layout.test.tsx` nor `page.test.tsx` references `applicationName` or `openGraph`. Asserting on the exported `metadata` object is trivial (it's a plain export) and would pin the fields this PR adds. Low effort; fold it into the test suggested in the `auth-guard.tsx` comment if you add one.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +88,26 @@ export function AuthGuard({ children }: AuthGuardProps) {
 
   const { t } = useI18n()
   if (isLoading) {
+    // The home route's loading state is also what static export (SSG) freezes
+    // into out/index.html, since auth can only resolve client-side. Anonymous
+    // crawlers (incl. Google OAuth brand verification) only ever see this
+    // markup, so it carries real copy about what the app does instead of a
+    // bare spinner.
+    if (pathname === "/") {
+      const branding = getBrandingFromEnv()
```
rogercloud: **F6 (nit) — style consistency.**

`getBrandingFromEnv()` is called inline in the render branch here, while `layout.tsx:11` calls it once at module scope. It's harmless (the function only reads `process.env.NEXT_PUBLIC_*` with `||` fallbacks and can't throw), but hoisting it to module scope in this file would match the existing convention. Purely cosmetic.

frontend/src/lib/branding.ts
```diff
@@ -22,6 +23,7 @@ export const defaultBranding: BrandingConfig = {
   gradientFrom: 'blue-400',
   gradientVia: 'blue-500',
   gradientTo: 'indigo-500',
+  siteUrl: 'https://cloud.xagent.co',
```
rogercloud, line 26: **[major]** `siteUrl` defaults to the hardcoded vendor domain `https://cloud.xagent.co`. Because `layout.tsx` feeds this into `metadataBase` and `openGraph.images[0].url` is a relative path (`branding.logoPath`), any self-hosted or white-label deployment that customizes `NEXT_PUBLIC_LOGO_PATH` but doesn't also set `NEXT_PUBLIC_SITE_URL` will emit `og:image` / `twitter:image` as `https://cloud.xagent.co/<their-logo-path>` — a domain they don't control, so their social previews 404.

These two env vars are now effectively a required set, but nothing says so: `.env.local.example` lists `NEXT_PUBLIC_SITE_URL` as an unrelated standalone entry.

Suggested fix: document the coupling in `.env.local.example` (group `NEXT_PUBLIC_SITE_URL` with the branding vars and note that overriding the logo requires setting it), and consider warning at startup when a non-default `logoPath` is combined with the default `siteUrl`.

frontend/src/app/layout.tsx
```diff
@@ -11,8 +11,21 @@ import { Toaster } from "@/components/ui/sonner";
 const branding = getBrandingFromEnv();
 
 export const metadata: Metadata = {
+  metadataBase: new URL(branding.siteUrl),
```
rogercloud: **[major]** `new URL(branding.siteUrl)` runs at module load with no validation, so a malformed `NEXT_PUBLIC_SITE_URL` takes down the entire production build rather than degrading.

Reproduced: with `NEXT_PUBLIC_SITE_URL=cloud.example.com` (missing scheme — an easy operator mistake) `next build` fails with `ERR_INVALID_URL` during "Collecting page data". The matched control `https://cloud.example.com` builds successfully.

Suggested fix: wrap the construction and fall back to the default with a warning, e.g.

```ts
function resolveMetadataBase(raw: string): URL {
  try {
    return new URL(raw);
  } catch {
    console.warn(`Invalid NEXT_PUBLIC_SITE_URL: ${raw}; falling back to default`);
    return new URL(DEFAULT_SITE_URL);
  }
}
```

A typo in one env var shouldn't be a build-stopping error.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +90,31 @@ export function AuthGuard({ children }: AuthGuardProps) {
 
   const { t } = useI18n()
   if (isLoading) {
+    // Auth only resolves client-side, so static export (next build) freezes
+    // this branch's JSX into out/index.html for the home route. Non-JS
+    // fetchers of that frozen HTML — including automated brand-verification
+    // crawlers — see only this markup, so it carries real copy about what
+    // the app does instead of a bare spinner. A JS-executing crawler that
+    // runs the redirect effect above will move past it once auth resolves.
+    if (pathname === "/") {
```
rogercloud, line 104: **[minor]** This hero block duplicates the one already in `frontend/src/app/page.tsx:409-412` — same `branding.whiteLogoPath`, same `branding.appName`, same i18n keys (`home.hero.title` / `home.hero.subtitle`), differing only in container styling. Two copies of the same brand copy will drift: a wording change on the real page silently leaves the frozen SSG placeholder stale, which is exactly the failure mode this PR exists to prevent.

Suggested fix: extract a small shared `HomeHero` (logo + title + subtitle, with a `className` prop for the differing container) and have both call sites render it.

frontend/src/components/auth/auth-guard.test.tsx
```diff
@@ -0,0 +1,58 @@
+import React from "react"
```
rogercloud, line 1: **[minor, test quality]** This test mocks `@/contexts/i18n-context` as an identity passthrough (`t: (key) => key`) and then asserts `screen.getByText("home.hero.title")` — so it verifies that the raw i18n key renders, not that any real copy does. It would still pass if the translation for that key were empty, missing, or broken, which undercuts the PR's stated goal of getting real copy into the frozen HTML.

Suggested fix: render this branch once with the real i18n provider (or at least a mock that resolves against the actual translation table) and assert on the resolved English string, so the copy that ships in `out/index.html` is genuinely pinned.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +90,31 @@ export function AuthGuard({ children }: AuthGuardProps) {
 
   const { t } = useI18n()
```
rogercloud, line 91: **[minor, non-blocking, pre-existing]** `const { t } = useI18n()` sits after the early `if (isAuthPage) return`, which is a real `react-hooks/rules-of-hooks` ESLint error (not just a lint nit). It predates this PR, so it isn't a regression — flagging only because this PR adds substantial new JSX in the block immediately below, making it a natural drive-by fix: hoist the `useI18n()` call above the early return. Fine to leave for a separate change if you'd rather keep this PR scoped.

rogercloud (commented):
## Re-review: follow-up commit `7f6426c4`

The follow-up commit `7f6426c4` ("fix: address PR review findings on home SSG placeholder") closes out every finding from the previous human review round. It softens the over-stated comment in `auth-guard.tsx` (the frozen markup is now correctly described as what *non-JS* fetchers see, with JS-executing crawlers moving past it once auth resolves), adds the missing brand logo to the frozen hero, and hoists `getBrandingFromEnv()` to module scope so it matches the pattern used in `layout.tsx`. On the metadata side it adds `metadataBase: new URL(branding.siteUrl)` plus `openGraph.images`, backed by a new `siteUrl` field in `branding.ts`. It also adds the previously missing test coverage: `auth-guard.test.tsx` now renders the real (unmocked) `AuthGuard` and asserts the home branch vs. the generic spinner branch, and `layout.test.tsx` asserts the new metadata fields. All added tests pass locally.

### Design verdict: acceptable-with-reservations

The premise holds up empirically — a real `next build` confirms `out/index.html` for `/` contains the hero copy while `out/settings.html` still renders the generic spinner, so `usePathname()` does reliably scope this to the home route during static export. The OG output is also correct: `metadataBase` resolves the relative image URL, and Next derives the `twitter:*` tags from `openGraph`.

Reservations:

- Hardcoding route knowledge (`pathname === "/"`) inside a generic `AuthGuard` is a mild scattered-responsibility smell. It is tolerable for one route, and a `loading.tsx`-style alternative isn't viable here because auth resolves client-side rather than through server Suspense.
- The hero markup added here is a second copy of the hero already in `page.tsx` rather than a shared component (see minor finding below).
- The frozen placeholder keeps the hardcoded `bg-[#0D1117]` dark background even though the exported body carries `theme-light`. This pattern predates the PR, but the change makes the mismatched block far larger and more visible. Passing observation only, not a blocking finding.
- `metadataBase` living in `layout.tsx` while `branding.ts` owns only the value is the right split and idiomatic for Next.

### Prior findings — status

| # | Source | Original issue | Status |
|---|---|---|---|
| G1 | gemini-code-assist (bot) | Add `images` to `openGraph` and a separate `twitter` card block | **PARTIAL** — `openGraph.images` added; no `twitter` block exists. Still-open, low priority: Next derives the Twitter tags from `openGraph`, and this was never part of the human review round. |
| F1 | prior review | Comment overstated the crawler guarantee | **FIXED** |
| F2 | prior review | No test coverage for the new `pathname === "/"` branch | **FIXED** — both branches asserted against the real component; tests pass |
| F3 | prior review | `openGraph` missing `images`; no `metadataBase` | **FIXED** |
| F4 | prior review | Frozen hero had no logo | **FIXED** |
| F5 | prior review | New metadata fields untested | **FIXED** |
| F6 | prior review | `getBrandingFromEnv()` called inline in render | **FIXED** |

### New findings introduced by `7f6426c4`

**Major**

1. `frontend/src/lib/branding.ts:26` — the new `siteUrl` field defaults to the hardcoded vendor domain `https://cloud.xagent.co`, and it is the base that `metadataBase` uses to resolve the relative `openGraph.images[0].url`. A white-label deployment that sets `NEXT_PUBLIC_LOGO_PATH` but not `NEXT_PUBLIC_SITE_URL` will emit an `og:image` pointing at a domain it does not control. The two env vars are now coupled but nothing documents that.
2. `frontend/src/app/layout.tsx:14` — `new URL(branding.siteUrl)` runs unguarded at module load. Setting `NEXT_PUBLIC_SITE_URL=cloud.example.com` (no scheme, a plausible operator typo) fails the whole production build with `ERR_INVALID_URL` during "Collecting page data"; the same value with `https://` builds fine. A single env var typo becomes a hard build failure instead of a graceful fallback.

**Minor**

3. `frontend/src/components/auth/auth-guard.tsx:99` — hero markup duplicated from `frontend/src/app/page.tsx:409`.
4. `frontend/src/components/auth/auth-guard.test.tsx` — the new test asserts the raw i18n key renders, not the real translated copy, which is the thing this PR is actually about.
5. `frontend/src/components/auth/auth-guard.tsx:91` — pre-existing `react-hooks/rules-of-hooks` violation directly above the block this PR rewrites; cheap to fix in passing.

Both major findings are latent rather than active: this deployment sets no override and is unaffected. They matter for other operators of this open-source project.

Lean already. (One candidate — dropping the `React` default import in `auth-guard.tsx` — was tested and rejected: this file compiles through the classic `React.createElement` transform, and removing the import throws `ReferenceError: React is not defined` at runtime.)

### Recommendation

Every finding from the previous human review round (F1–F6) is genuinely fixed, with real test coverage rather than token assertions, and the core premise is empirically verified. The two major findings are real latent bugs but sit outside this deployment's actual usage — neither can regress the first-party site, since it relies on the defaults. That is not enough to block a follow-up commit that did what it set out to do. I would still like finding #2 addressed before merge; it is a few lines (validate or try/catch around the URL construction with a fallback to the default) and turns a build-breaking typo into a warning. Finding #1 can be closed with a comment in `.env.local.example` tying `NEXT_PUBLIC_SITE_URL` to the branding vars.

frontend/src/components/auth/auth-guard.tsx
yiboyasss: Agreed on the drift risk in principle, but deferring the extraction to a follow-up: the two hero blocks differ in more than container styling (the real page adds the gradient/grid/orb background layers, nav pills, and different sizing), so a shared `HomeHero` would need enough knobs that it risks touching the production home page visuals for the sake of a placeholder. Two mitigations are already in place: both copies read the same i18n keys (`home.hero.title` / `home.hero.subtitle`), so wording changes propagate to both automatically, and the new auth-guard test now pins the literal English copy of the frozen placeholder, so it can't silently go stale.

frontend/src/components/auth/auth-guard.tsx
yiboyasss: Confirmed it's a real `react-hooks/rules-of-hooks` violation, but taking the option you offered and leaving it out of this PR to keep the scope tight — will hoist the `useI18n()` call above the early return in a separate change.

frontend/src/app/layout.tsx
```diff
@@ -3,16 +3,40 @@ import React, { type CSSProperties } from "react";
...
 
 const branding = getBrandingFromEnv();
 
+// A malformed NEXT_PUBLIC_SITE_URL (e.g. missing scheme) must not take down
+// the production build; fall back to the default site URL instead.
+function resolveMetadataBase(siteUrl: string): URL {
+  try {
+    return new URL(siteUrl);
+  } catch {
+    console.warn(`Invalid NEXT_PUBLIC_SITE_URL "${siteUrl}", falling back to ${defaultBranding.siteUrl}`);
```
rogercloud: The fallback keeps a bad `NEXT_PUBLIC_SITE_URL` from failing the build, which is the right call, and this `console.warn` does survive `removeConsole` in production. Consider `console.error` instead so it reads unambiguously as a misconfiguration signal, and/or note in the deployment docs that build logs should be checked when this env var is set.

frontend/src/app/layout.tsx
```diff
@@ -3,16 +3,40 @@ import React, { type CSSProperties } from "react";
...
+import { defaultBranding, getBrandingFromEnv } from "@/lib/branding";
 import { I18nProvider } from "@/contexts/i18n-context";
 import { getThemeFromEnv, themes } from "@/lib/theme";
 import { Toaster } from "@/components/ui/sonner";
 
 const branding = getBrandingFromEnv();
 
+// A malformed NEXT_PUBLIC_SITE_URL (e.g. missing scheme) must not take down
+// the production build; fall back to the default site URL instead.
+function resolveMetadataBase(siteUrl: string): URL {
```
rogercloud: `resolveMetadataBase` is unexported, so only its happy path is covered by `src/app/layout.test.tsx`. The catch/fallback branch has no direct test. Low risk while the helper is this simple, but worth exporting and testing if it grows.

frontend/src/app/layout.tsx
```diff
@@ -3,16 +3,40 @@ import React, { type CSSProperties } from "react";
...
+    return new URL(defaultBranding.siteUrl);
+  }
+}
+
 export const metadata: Metadata = {
+  metadataBase: resolveMetadataBase(branding.siteUrl),
   title: branding.appName,
   description: branding.description,
+  applicationName: branding.appName,
+  openGraph: {
```
rogercloud, line 18: `openGraph` omits `type` and `url`, and Next.js does not derive either when unset. Adding `type: "website"` and `url: "/"` improves compliance with Facebook's OG validator and some brand-verification crawlers.

frontend/src/app/layout.tsx
```diff
@@ -3,16 +3,40 @@ import React, { type CSSProperties } from "react";
...
+    siteName: branding.appName,
+    title: branding.appName,
+    description: branding.description,
+    images: [
```
rogercloud, line 24: The `og:image` here is the 300x300 logo with no `width`/`height`/`type` declared. It clears Facebook's 200x200 minimum but is well under the 1200x630 recommended for rich previews. Declaring the known dimensions is a cheap win; a dedicated social-card asset would be a good follow-up.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -87,6 +90,31 @@ export function AuthGuard({ children }: AuthGuardProps) {
...
   if (isLoading) {
+    // Auth only resolves client-side, so static export (next build) freezes
+    // this branch's JSX into out/index.html for the home route. Non-JS
+    // fetchers of that frozen HTML — including automated brand-verification
+    // crawlers — see only this markup, so it carries real copy about what
+    // the app does instead of a bare spinner. A JS-executing crawler that
+    // runs the redirect effect above will move past it once auth resolves.
+    if (pathname === "/") {
+      return (
```
rogercloud, line 105: This placeholder hero is styled differently from the real one in `frontend/src/app/page.tsx:408-413`: flat `bg-[#0D1117]` vs. the gradient/grid/orb background, and `w-12 h-12`/`text-2xl` vs. `w-14 h-14`/`text-[34px]`. Once auth resolves, `AuthGuard` swaps straight to `children`, so authenticated users see the logo and title jump size and the background change on every load of `/`. That's new — the previous loading state was a plain spinner. Aligning at least the logo size and title/subtitle typography would make the swap invisible, even if the decorative background stays deferred.

frontend/src/components/auth/auth-guard.tsx
```diff
@@ -1,10 +1,13 @@
...
 
-import { useEffect, useState } from "react"
+import React, { useEffect, useState } from "react"
 import { useAuth } from "@/contexts/auth-context"
 import { isAuthPublicPath } from "@/lib/auth-pages"
 import { useRouter, usePathname } from "next/navigation"
 import { useI18n } from "@/contexts/i18n-context"
+import { getBrandingFromEnv } from "@/lib/branding"
+
+const branding = getBrandingFromEnv()
```
rogercloud, line 10: Nit: `getBrandingFromEnv()` is called at module scope here but inside the component body in `frontend/src/app/page.tsx:174`. Harmless since these env vars are inlined at build time — just noting the inconsistency.

frontend/src/components/auth/auth-guard.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+// Resolves against the real English translation table (rather than an
+// identity passthrough) so this pins the actual copy shipped in out/index.html.
+vi.mock("@/contexts/i18n-context", () => ({
+  useI18n: () => ({
+    t: (key: string, vars?: Record<string, string | number>) =>
+      resolveTranslation("en", key as TranslationKey, vars),
+  }),
+}))
+
+vi.mock("@/lib/branding", () => ({
```
rogercloud: This mock's `whiteLogoPath` (`/logo-white.png`) doesn't match the real default (`/xagent_white_logo.png`) — consider importing the actual `defaultBranding` instead of a hand-written literal so the two can't drift. Also, the test never asserts the rendered ``'s `src`/`alt`, even though the code path under test renders one from `branding.whiteLogoPath`/`branding.appName`.

rogercloud (approved):
## Summary

This PR replaces the bare spinner that `AuthGuard` renders while auth is resolving on `/` with a branded hero carrying real translated copy, and adds OpenGraph/`metadataBase`/`applicationName` metadata to the root layout. Because the frontend is statically exported (`next build`) and auth resolves client-side, `out/index.html` for `/` freezes whatever `AuthGuard` renders during `isLoading` — so non-JS crawlers and brand-verification bots previously saw an empty spinner shell. This round evaluates commit 062c2c0a, which addresses the two blocking findings from the previous review, alongside a fresh blank-slate design and discovery pass over the current state of the code.

## Design verdict (fresh pass)

**Acceptable with reservations.** The problem is real and the fix targets it correctly; the `layout.tsx` metadata additions are unambiguously standard and correct. The `AuthGuard` special-case is a pragmatic hack rather than a clean abstraction — a generic auth wrapper now embeds route-specific SEO/marketing logic — but no simpler alternative exists here: auth resolves client-side, so no `loading.tsx`/Suspense boundary applies, and turning `/` into a fully public static route is a materially larger product decision that is reasonably out of scope for this PR. The hero markup being duplicated from `frontend/src/app/page.tsx:408-413` instead of extracted is likewise a defensible call for now (see status table below).

## Status of previous findings

| # | Source | Issue | Status |
|---|---|---|---|
| G1 | gemini-code-assist | Add `images` to `openGraph` and a separate `twitter` block | **Fixed.** `openGraph.images` added at `frontend/src/app/layout.tsx:33-38`. No explicit `twitter` block, but Next.js auto-derives `twitter:title`/`description`/`images` from `openGraph` when none is set (verified against Next's own metadata resolver), so this is functionally resolved. Only `twitter:card` type is never auto-set — cosmetic. |
| R2-1 | Round 2, major | `siteUrl` defaults to the hardcoded vendor domain; a white-label deploy overriding the logo but not `NEXT_PUBLIC_SITE_URL` emits an `og:image` on a domain it doesn't control — undocumented coupling | **Fixed.** `frontend/.env.local.example:54-58` now documents the coupling explicitly and warns about the resulting `og:image`/`twitter:image` domain. |
| R2-2 | Round 2, major | `new URL(branding.siteUrl)` ran unguarded at module load; a malformed `NEXT_PUBLIC_SITE_URL` failed the whole production build | **Fixed.** `frontend/src/app/layout.tsx:15-22` wraps it in `resolveMetadataBase()` with try/catch, falling back to the well-formed `defaultBranding.siteUrl` (so the fallback path cannot itself throw) plus a `console.warn` — confirmed not stripped in production, since `next.config.mjs` excludes `warn`/`error` from `removeConsole`. |
| R2-3 | Round 2, minor | Hero markup duplicates `page.tsx:409-412` | **Waived.** Still duplicated, but the author's stated reason checks out: the two blocks diverge in more than styling (`page.tsx` carries extra gradient/grid/orb decorative layers plus nav-pill and search sections the loading placeholder correctly omits). Reasonable follow-up. |
| R2-4 | Round 2, minor | `useI18n()` called after an early conditional return (`react-hooks/rules-of-hooks`) | **Waived, still present** at `frontend/src/components/auth/auth-guard.tsx:91`. Confirmed pre-existing and unchanged from the base commit, and confirmed a genuine ESLint *error* via `npx eslint`. Author acknowledged and deferred the one-line hoist to keep this PR scoped — fine. |
| R2-5 | Round 2, minor | Test mocked i18n as an identity passthrough, asserting the raw key `"home.hero.title"` | **Fixed.** `auth-guard.test.tsx` now resolves through the real `resolveTranslation` against the actual English table and asserts literal copy ("Welcome to Xagent", the full subtitle). Verified non-tautological — `src/i18n/locales/en.ts` defines those as independent strings — and the suite passes. |

Every actionable item from the previous round is closed. Both majors have working, tested guards; the two waivers are genuine deferrals with author acknowledgment rather than oversights.

## New findings this round

### Minor

1. **`frontend/src/app/layout.tsx:15-22` — malformed-URL fallback is non-fatal by design; consider a louder signal.** Falling back rather than crashing the build over one bad env var is a deliberate, commented choice, and the `console.warn` does survive production log stripping, so it isn't silent. Still, `console.error` reads more clearly as a production signal — or document in the deployment notes that build logs should be checked whenever `NEXT_PUBLIC_SITE_URL` is set.
2. **`frontend/src/app/layout.tsx:15-22`, `frontend/src/app/layout.test.tsx:117-128` — the fallback branch is untested.** `resolveMetadataBase` is unexported, so only the happy path is covered. Low risk while the helper stays this simple, but worth exporting and testing the catch branch if it grows.
3. **Coverage gap: nothing pins the built `out/index.html` content end-to-end.** `AuthGuard`'s output is tested in isolation with `usePathname` mocked, not through `ApplicationShell`'s routing decision (`frontend/src/components/layout/application-shell.tsx:13-38`, whose `isExternalRoutePath` decides whether `/` reaches `AuthGuard` at all). CI (`.github/workflows/ci.yml:695-697`) only asserts that `out/index.html` exists, not what's in it. A future unrelated change to `ApplicationShell`'s routing could silently revert this PR's entire effect with every test still green. A lightweight CI grep for the hero string, or a test rendering `ApplicationShell` for `/`, would close this.
4. **`frontend/src/app/layout.tsx:29-39` — `openGraph` omits `type` and `url`.** Verified that Next.js derives neither when unset. Adding `type: "website"` and `url: "/"` improves compliance with Facebook's OG validator and some brand-verification crawlers.
5. **`frontend/src/app/layout.tsx:33-38` — `og:image` has no dimensions declared and is small.** The image is `branding.logoPath` (confirmed 300x300). It clears Facebook's 200x200 minimum but is well below the 1200x630 generally recommended for rich previews, and the `images` entry declares no `width`/`height`/`type`. Declare the known dimensions at minimum; a dedicated larger social-card asset is a reasonable follow-up.
6. **`frontend/src/components/auth/auth-guard.tsx:100-115` vs `frontend/src/app/page.tsx:408-413` — visible flash/resize on every authenticated load of `/`.** Beyond the waived duplication concern, the two heroes are styled differently: flat `bg-[#0D1117]` vs. the gradient/grid/orb background, and `w-12 h-12`/`text-2xl` vs. `w-14 h-14`/`text-[34px]` for logo and title. `AuthGuard` renders the placeholder hero while `isLoading`, then swaps straight to `children` (the real hero) once auth resolves — so authenticated users now see the logo and title jump size and the background change. This is new: the pre-PR loading state was a plain spinner with no hero at all. Worth aligning at least the overlapping parts (logo size, title/subtitle typography) so the swap isn't jarring, even if the decorative background stays deferred under the existing waiver.
7. **`frontend/src/components/auth/auth-guard.test.tsx:32-34` — the branding mock drifts from the real default and the image is unasserted.** The `@/lib/branding` mock is a hand-written literal (`{ appName: "Xagent", whiteLogoPath: "/logo-white.png" }`); the stubbed `whiteLogoPath` doesn't match the real default (`/xagent_white_logo.png`). Prefer importing the real `defaultBranding`. The test also never asserts the rendered ``'s `src`/`alt`, even though `auth-guard.tsx:103-107` renders one from `branding.whiteLogoPath`/`branding.appName` in the exact path under test.

### Nit

8. **`frontend/src/components/auth/auth-guard.tsx:10` — `getBrandingFromEnv()` is called at module scope**, whereas `frontend/src/app/page.tsx:174` calls it inside the component body. Harmless in practice (these env vars are bundler-inlined at build time) — noting only for consistency.

Lean already.

## Testing performed

- `npx vitest run src/components/auth/auth-guard.test.tsx src/app/layout.test.tsx` — 6 passed.
- `npx tsc --noEmit` — clean.
- `npx eslint src/components/auth/auth-guard.tsx` — one pre-existing `react-hooks/rules-of-hooks` error at line 91 (the waived item above); no new lint errors introduced by this PR.
- `resolveMetadataBase`'s malformed-URL fallback verified by reading the code path rather than executing it standalone, since it's an unexported build-time helper.

## Recommendation

**Approve.** Every actionable finding from the previous round is fixed, and nothing new rises to blocking severity — the items above are minor polish or coverage gaps rather than active bugs. One thing is worth a quick look before merge if you have time: finding #6, the logo/title size and background jump when the placeholder hero swaps out for the real one. It's a small but real UX regression that this PR introduces, and it may not have been noticed yet. Everything else is fine as follow-up.

PR #1146, opened 2026-08-05
feat: tell the Slack agent who is talking to it

OliverBryant:
## Summary

Slack events carry only the sender's opaque user id, so the agent had no idea who it was talking to — it could not address the person and, in a shared channel, could not tell workspace members apart. This closes that gap:

- the bot resolves the sender's profile via `users.info` once per sender per instance (cached — including failures, so an install that predates the new scope degrades to unattributed messages instead of one failing API call per message)
- the agent-facing text gets a `[From: Dana Reyes (Data Scientist)]` header; the persisted transcript and everything the user sees keep the raw message
- the header survives the uploaded-files context (the files branch now appends to the attributed text), and resolution runs only after the sender passes channel authorization, so unauthorized senders never trigger an API call
- `users:read` joins the OAuth install scopes; `example.env` documents the reinstall note

Orthogonal to #1130: that PR decides *which agent* answers a channel; this one tells the agent *who is asking*.

## Why

The immediate consumer is Toby (xorbitsai/xagent-saas#385), whose product spec literally opens with "Hey Gerard! … I hear you're the Data Scientist & AI Manager here" — impossible without sender identity. The capability is generic though: any Slack workspace agent benefits, and nothing here is product-specific.

## Validation

- `pytest tests/web/test_slack_channel.py` — 54 passed (4 new: header on the agent text with the transcript unchanged plus per-sender caching, failure degradation without per-message retries, name-only formatting, attachment-context interaction)
- ruff check / format clean

## Notes for reviewers

- The header is prepended to user-controlled text, so a user could type a fake `[From: …]` line in their own message body; it carries exactly the same trust as the rest of their message, which the agent already treats as user input.
- Existing installs lack `users:read` until reinstalled through the OAuth flow; behaviour is unchanged for them (log line at INFO, one per sender).

### DIFF
--- a/src/xagent/web/api/channel.py
+++ b/src/xagent/web/api/channel.py
@@ -67,6 +67,10 @@ SLACK_OAUTH_SCOPES = (
     "mpim:history",
     "channels:history",
     "groups:history",
+    # Sender-identity resolution (users.info): lets the agent address the
+    # person it is talking to. Installs that predate this scope degrade to
+    # unattributed messages until reinstalled.
+    "users:read",
 )
 
 
--- a/src/xagent/web/channels/slack/bot.py
+++ b/src/xagent/web/channels/slack/bot.py
@@ -70,8 +70,8 @@ _CONTROL_START_COMMANDS = {"/start", "start"}
 _CONTROL_NEW_COMMANDS = {"/new", "new", "new task"}
 _MAX_FILE_DOWNLOAD_REDIRECTS = 5
 _MAX_RECENT_EVENT_IDS = 1000
-# Senders whose resolved identity headers are kept per bot instance. Bounded
-# only as a runaway guard; real workspaces stay far below this.
+# Senders whose resolved identity headers are kept per bot instance. At the
+# cap the oldest entry is evicted (FIFO); real workspaces stay far below this.
 _MAX_SENDER_CONTEXT_CACHE = 10_000
 # Slack rejects message text over 4000 characters. Chunk the source below that
 # and clamp again after conversion, since entity escaping can expand text.
@@ -460,8 +460,12 @@ class SlackBotInstance:
                 slack_user_id,
                 exc_info=True,
             )
-        if len(self._sender_contexts) < _MAX_SENDER_CONTEXT_CACHE:
-            self._sender_contexts[slack_user_id] = sender_context
+        if len(self._sender_contexts) >= _MAX_SENDER_CONTEXT_CACHE:
+            # Evict the oldest entry (dicts preserve insertion order) so
+            # senders arriving after the cap still get cached instead of
+            # paying a users.info call on every message.
+            self._sender_contexts.pop(next(iter(self._sender_contexts)))
+        self._sender_contexts[slack_user_id] = sender_context
         return sender_context
 
     async def _process_event(
--- a/src/xagent/web/channels/slack/bot.py
+++ b/src/xagent/web/channels/slack/bot.py
@@ -70,6 +70,9 @@ _CONTROL_START_COMMANDS = {"/start", "start"}
 _CONTROL_NEW_COMMANDS = {"/new", "new", "new task"}
 _MAX_FILE_DOWNLOAD_REDIRECTS = 5
 _MAX_RECENT_EVENT_IDS = 1000
+# Senders whose resolved identity headers are kept per bot instance. Bounded
+# only as a runaway guard; real workspaces stay far below this.
+_MAX_SENDER_CONTEXT_CACHE = 10_000
 # Slack rejects message text over 4000 characters. Chunk the source below that
 # and clamp again after conversion, since entity escaping can expand text.
 _MAX_SOURCE_CHUNK_CHARS = 3500
@@ -166,6 +169,11 @@ class SlackBotInstance:
         self.event_tasks: dict[str, asyncio.Task[None]] = {}
         self._recent_event_ids: deque[str] = deque(maxlen=_MAX_RECENT_EVENT_IDS)
         self._recent_event_id_set: set[str] = set()
+        # slack_user_id -> "[From: ...]" header (None = resolution failed).
+        # Failures are cached too: a workspace whose install predates the
+        # users:read scope must degrade to unattributed messages, not one
+        # failing API call per message.
+        self._sender_contexts: dict[str, str | None] = {}
         self._accepting = True
         self._deactivation_sync_task: asyncio.Task[None] | None = None
         self._run_forever = asyncio.Event()
@@ -414,6 +422,48 @@ class SlackBotInstance:
             value = event.get("thread_ts") or event.get("ts")
         return str(value) if value else None
 
+    async def _resolve_sender_context(self, slack_user_id: str) -> str | None:
+        """A one-line identity header for the agent-facing prompt.
+
+        Slack events carry only the sender's opaque user id, so without this
+        the agent cannot address the person or tell workspace members apart
+        in a shared channel. The profile is fetched once per sender per
+        instance; a failed lookup (most commonly an install that predates the
+        users:read scope) is cached as None so the turn proceeds
+        unattributed instead of retrying the API on every message.
+        """
+        if slack_user_id in self._sender_contexts:
+            return self._sender_contexts[slack_user_id]
+        sender_context: str | None = None
+        try:
+            response = await self.web_client.users_info(user=slack_user_id)
+            user = response.get("user")
+            user = user if isinstance(user, dict) else {}
+            profile = user.get("profile")
+            profile = profile if isinstance(profile, dict) else {}
+            name = str(
+                profile.get("display_name")
+                or profile.get("real_name")
+                or user.get("real_name")
+                or user.get("name")
+                or ""
+            ).strip()
+            if name:
+                title = str(profile.get("title") or "").strip()
+                sender_context = (
+                    f"[From: {name} ({title})]" if title else f"[From: {name}]"
+                )
+        except Exception:
+            logger.info(
+                "Slack users.info failed for %s; messages will be unattributed"
+                " (is the users:read scope granted?)",
+                slack_user_id,
+                exc_info=True,
+            )
+        if len(self._sender_contexts) < _MAX_SENDER_CONTEXT_CACHE:
+            self._sender_contexts[slack_user_id] = sender_context
+        return sender_context
+
     async def _process_event(
         self,
         conversation_key: str,
@@ -521,7 +571,13 @@ class SlackBotInstance:
             turn_id = str(uuid4())
             context: dict[str, Any] = {"turn_id": turn_id}
             display_message = text
-            execution_text = prompt_text
+            # Agent-facing text only: the persisted transcript and the Slack
+            # UI keep the raw message; the identity header lets the agent
+            # address the sender and tell workspace members apart.
+            sender_context = await self._resolve_sender_context(slack_user_id)
+            execution_text = (
+                f"{sender_context}\n{prompt_text}" if sender_context else prompt_text
+            )
             persisted_attachments: list[dict[str, Any]] = []
 
             if files:
@@ -542,7 +598,7 @@ class SlackBotInstance:
                     names = ", ".join(str(item["name"]) for item in uploaded_info)
                     display_message = f"Attached file(s): {names}"
                 execution_text = append_uploaded_files_context(
-                    prompt_text,
+                    execution_text,
                     build_uploaded_files_context(uploaded_info),
                 )
                 if is_new_task:
--- a/tests/web/test_slack_channel.py
+++ b/tests/web/test_slack_channel.py
@@ -1549,3 +1549,30 @@ async def test_sender_identity_survives_the_attachment_context(
     task_text = executed[0]["task"]
     assert task_text.startswith("[From: Dana Reyes]\nplease review")
     assert "report.pdf" in task_text  # the files context still follows
+
+
+@pytest.mark.asyncio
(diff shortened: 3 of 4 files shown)


src/xagent/web/channels/slack/bot.py
```diff
@@ -414,6 +422,48 @@ def _reply_thread_ts(event: dict[str, Any]) -> str | None:
...
+        except Exception:
+            logger.info(
+                "Slack users.info failed for %s; messages will be unattributed"
+                " (is the users:read scope granted?)",
+                slack_user_id,
+                exc_info=True,
+            )
+        if len(self._sender_contexts) < _MAX_SENDER_CONTEXT_CACHE:
+            self._sender_contexts[slack_user_id] = sender_context
+        return sender_context
```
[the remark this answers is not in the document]
OliverBryant: Fixed in 4cb30e18 — at the cap the oldest entry is evicted (dict insertion order) before inserting, so late-arriving senders are still cached; covered by `test_sender_identity_cache_evicts_oldest_when_full`.

src/xagent/web/api/channel.py
```diff
@@ -67,6 +67,10 @@
     "mpim:history",
     "channels:history",
     "groups:history",
+    # Sender-identity resolution (users.info): lets the agent address the
+    # person it is talking to. Installs that predate this scope degrade to
+    # unattributed messages until reinstalled.
+    "users:read",
```
rogercloud, line 73: **MAJOR** — adding `users:read` here covers the OAuth path, but the manual-install instructions shown in the product were not updated. `channels.dialog.slack_permissions_help` in `frontend/src/i18n/locales/en.ts:541` and `frontend/src/i18n/locales/zh.ts:541` lists the bot scopes a manual installer must grant (rendered by `frontend/src/app/channels/page.tsx` for any Slack channel whose `installation_mode` is not `oauth`) and still omits `users:read`. Anyone following that text verbatim lands in exactly the degraded state this PR handles gracefully — all senders unattributed, signalled only by an internal INFO log. Please append `users:read` to both locale strings so the OAuth list, `example.env`, and the in-product help agree.

src/xagent/web/channels/slack/bot.py
```diff
@@ -414,6 +422,52 @@ def _reply_thread_ts(event: dict[str, Any]) -> str | None:
...
+                or user.get("real_name")
+                or user.get("name")
+                or ""
+            ).strip()
+            if name:
+                title = str(profile.get("title") or "").strip()
+                sender_context = (
+                    f"[From: {name} ({title})]" if title else f"[From: {name}]"
+                )
+        except Exception:
```
rogercloud, line 456: **MAJOR** — this bare `except Exception:` makes every failure permanent for the lifetime of the instance: a 429 rate limit or a network timeout is cached as `None` exactly like a genuinely missing scope, and the log line unconditionally asks "is the `users:read` scope granted?" no matter the real cause, which will misdirect whoever debugs it. It also does not self-heal: reinstalling to grant the scope usually keeps the same bot token, and instance reuse at lines 1385-1393 is keyed only on bot_token / channel_id / bot_user_id equality, so the old instance and its stale `None` entries survive the reinstall.

Suggest distinguishing retryable from permanent failures — do not cache on transient errors:
```python
except SlackApiError as exc:
    if _is_retryable(exc):  # ratelimited, transport-level
        return None         # retry on the next message
    logger.info(...)        # missing_scope, user_not_found -> cache the miss
except Exception:
    logger.info(...)
    return None
```
Clearing `_sender_contexts` when the bot's granted scope string changes would also work.

src/xagent/web/channels/slack/bot.py
```diff
@@ -521,7 +575,13 @@ async def _process_event(
             turn_id = str(uuid4())
             context: dict[str, Any] = {"turn_id": turn_id}
             display_message = text
-            execution_text = prompt_text
+            # Agent-facing text only: the persisted transcript and the Slack
+            # UI keep the raw message; the identity header lets the agent
+            # address the sender and tell workspace members apart.
+            sender_context = await self._resolve_sender_context(slack_user_id)
+            execution_text = (
+                f"{sender_context}\n{prompt_text}" if sender_context else prompt_text
```
rogercloud, line 583: **MINOR** — the bot-generated header is concatenated onto fully user-controlled text with no delimiter or escaping, so a sender can type `[From: Alice (Admin)]` as their first line and produce two visually identical headers the agent cannot tell apart. The PR notes disclose this and the impact is currently bounded (identity gates nothing, and `_conversation_key` already keeps each conversation single-sender), so this is not blocking. A cheap fix: carry the identity in the `context: dict[str, Any]` built two lines above, where it cannot be forged, or escape a leading `[From:` line in the raw text before concatenating.

src/xagent/web/channels/slack/bot.py
```diff
@@ -414,6 +422,52 @@ def _reply_thread_ts(event: dict[str, Any]) -> str | None:
...
+        """A one-line identity header for the agent-facing prompt.
+
+        Slack events carry only the sender's opaque user id, so without this
+        the agent cannot address the person or tell workspace members apart
+        in a shared channel. The profile is fetched once per sender per
+        instance; a failed lookup (most commonly an install that predates the
+        users:read scope) is cached as None so the turn proceeds
+        unattributed instead of retrying the API on every message.
+        """
+        if slack_user_id in self._sender_contexts:
```
rogercloud, line 435: **MINOR** — this membership check and the cache write below it straddle an `await` on `users_info`. Serialization is per `conversation_key` (`team:channel:user:scope`, lines 369-381), not per Slack user id, so the same user hitting two keys at once (a DM and a channel post) can miss the cache twice and issue two `users.info` calls. Harmless in practice — idempotent, last write wins — but the docstring's "fetched once per sender per instance" is stronger than the code guarantees. Either soften the wording or add a per-user in-flight marker.

src/xagent/web/channels/slack/bot.py
```diff
@@ -166,6 +169,11 @@ def __init__(
         self.event_tasks: dict[str, asyncio.Task[None]] = {}
         self._recent_event_ids: deque[str] = deque(maxlen=_MAX_RECENT_EVENT_IDS)
         self._recent_event_id_set: set[str] = set()
+        # slack_user_id -> "[From: ...]" header (None = resolution failed).
```
rogercloud, line 172: **MINOR** — `None = resolution failed` is imprecise: a *successful* `users_info` call whose `display_name`, `real_name`, `user.real_name`, and `user.name` are all empty also stores `None`, and this comment gives no way to tell that from an API failure. Behavior is correct in both cases; suggest wording like `None = no usable identity (lookup failed, or the profile has no name)`.

tests/web/test_slack_channel.py
```diff
@@ -1332,3 +1333,246 @@ async def send_final_text(**kwargs: Any) -> None:
...
+
+    assert executed[0]["task"] == "[From: Dana Reyes (Data Scientist)]\nhello"
+    assert executed[1]["task"] == "[From: Dana Reyes (Data Scientist)]\nsecond"
+    assert persisted[0]["content"] == "hello"
+    assert persisted[1]["content"] == "second"
+    assert calls == ["U1"]  # cached after the first resolution
+
+
+@pytest.mark.asyncio
+async def test_sender_identity_failure_degrades_to_unattributed(
```
rogercloud, line 1466: **MINOR** — this test asserts the current "cache `None` on any exception" behavior using a generic `RuntimeError`, so it would still pass after a fix for the transient-vs-permanent classification issue in `_resolve_sender_context`; as written it locks in the behavior rather than pinning the intended one. Once failures are classified, worth splitting into a permanent case (cached, one API call) and a transient case (not cached, retried on the next message).

Two other gaps in this group: nothing pins the PR's stated property that `users_info` is never called when `prepare_channel_task` raises `ChannelAuthorizationError` (true today, since resolution sits after the auth early-return at lines 509-516, but a future hoist would go unnoticed), and nothing covers two conversation keys resolving the same Slack user concurrently.

tests/web/test_slack_channel.py
```diff
@@ -1332,3 +1333,246 @@ async def send_final_text(**kwargs: Any) -> None:
...
+
+    bot.web_client = SimpleNamespace(users_info=users_info)  # type: ignore[assignment]
+
+    await bot._process_event("T1:D1:U1:direct", {}, _im_event())
+
+    assert executed[0]["task"] == "[From: Gerard Soto]\nhello"
+
+
+@pytest.mark.asyncio
+async def test_sender_identity_survives_the_attachment_context(
```
rogercloud, line 1510: **MINOR** — this asserts only `executed[0]["task"]`. The attachment branch is precisely where an `execution_text` / `prompt_text` mixup is most likely to regress, so please also assert that the persisted message and `display_message` stayed raw (no `[From: ...]` header) in this branch — the harness already returns `persisted`.

tests/web/test_slack_channel.py
```diff
@@ -1332,3 +1333,246 @@ async def send_final_text(**kwargs: Any) -> None:
         }
     ]
     assert managed.closed is True
+
+
+def _run_turn_harness(
```
rogercloud, line 1338: **Simplification** — this harness re-creates the ~90-line fake block (`FakeManagedLease`, `prepare`, `FakeTracer`, `agent_service`, `FakeAgentManager`, `persist`, `send_text`, `send_final_text`, and the four `monkeypatch.setattr` calls) that already exists inline in `test_successful_slack_turn_reuses_channel_runtime` just above at lines 1209-1336. The two blocks are structurally identical; the deltas are purely additive — the older test also tracks `closed`/`finalized`/thread_ts, the harness also captures `executed`/`persisted`. A superset harness capturing all of it would serve both call sites and save roughly 75 lines with no behavior change.

rogercloud (changes requested):
## Summary

Slack events only carry an opaque sender id, so the agent had no way to know who it was talking to. This PR resolves the sender's profile through `users.info`, caches the result per sender per bot instance (failures included, so a workspace missing the scope does not pay a failing API call per message), and prepends a `[From: Name (Title)]` header to the agent-facing text only — the persisted transcript and the Slack UI keep the raw message. The header is also threaded correctly through the uploaded-files branch, and the `users:read` scope is added to the OAuth scope list with a matching `example.env` note.

## Approach verdict

**Acceptable with reservations** — the change is small, well placed, and solves the stated problem without disturbing persistence or the transport layer.

Design-level reservations (no single owning line, so noted here rather than inline):

- **Failure caching has no escape hatch short of a process restart.** The natural remedy for a missing scope is to reinstall the app and grant it — but a reinstall for the same workspace typically keeps the same bot token, channel, and bot user id, and `SlackBotInstance` reuse (`src/xagent/web/channels/slack/bot.py:1385-1393`) is keyed only on equality of those three. So the existing instance, along with its stale `None` entries, is reused and the affected senders stay unattributed until something unrelated changes or the process restarts. See finding 2.
- **The stated "tell workspace members apart in a shared channel" motivation is already largely met** by the existing per-sender `_conversation_key` (`team:channel:user:scope`), which keeps each sender on their own conversation. The real value this PR delivers is the human-readable name and title, not disambiguation. That is a reason the unpersisted header is currently benign rather than a design flaw — worth knowing, not worth changing here.
- **Identity travels as untrusted inline text rather than structured context.** A `context: dict[str, Any]` is already constructed and passed alongside the prompt at the same call site; carrying identity there would make the attribution unspoofable by construction and is the root cause behind finding 3. The current approach is acceptable for now, but the structured path is the more durable one if identity ever informs a decision.
- **Forward-looking suggestion (not tied to one line):** `SLACK_OAUTH_SCOPES` in `src/xagent/web/api/channel.py` records `users:read` only to build the authorize URL, while the OAuth callback separately stores the workspace's actually-granted scope string in the channel config (~line 401). Nothing reads that stored value back, so an install predating `users:read` is only ever signalled by a per-message INFO log that no channel owner will see. Comparing the stored granted scopes against the required set and raising a one-time warning would close that loop — the codebase already has an established pattern for "this install predates a requirement" signals in `src/xagent/web/services/ops_signals.py` (used for Gmail OIDC). This is a nice-to-have beyond this PR's stated scope; feel free to defer.
- The FIFO eviction at a 10,000-entry cap is fine — no realistic Slack workspace approaches it. There is no caching convention in the sibling Feishu/Telegram channels being ignored here; this is genuinely new territory.

```
L1338: duplication  _run_turn_harness re-creates the ~90-line fake block (FakeManagedLease, prepare, FakeTracer, agent_service, FakeAgentManager, persist, send_text, send_final_text, four monkeypatch.setattr calls) already inline in test_successful_slack_turn_reuses_channel_runtime at L1209-1336. Make the harness a superset (also track closed/finalized/thread_ts) and have the older test call it too.
net: -75 lines possible
```

Nothing else in the diff reads as over-engineered. The FIFO cache (plain dict plus `pop(next(iter(...)))`) is not a meaningful stdlib reinvention — it matches `OrderedDict.popitem(last=False)` with no import saved, and it must cache `None` as a valid value, which `functools.lru_cache` cannot express. The name-fallback chain is one reasonable expression, and there is no unused flexibility or config anywhere in the change.

## Test results

`pytest tests/web/test_slack_channel.py` — **55 passed** (Python 3.11), including the 5 tests added by this PR.

## Recommendation

Safe to merge overall. Finding 1 (the manual-install scope list in the i18n strings) and finding 2 (classifying transient vs permanent resolution failures) are worth addressing before or shortly after merge — the first because it actively steers manual installers into the degraded path, the second because a transient blip currently costs a sender their identity for the life of the process. The remaining items are nits and observations and should not block.

OliverBryant:
Stale since the v2 architecture rewrite; the V0/V1 scope has shipped through other PRs.

PR #1024, opened 2026-07-27
fix(integrations): route regional callbacks and deployment clients

bsbds:
## Summary

Route non-browser integrations and owner-generated deployment artifacts to URLs
that remain reachable in a regional XAgent deployment.

This change:

- adds `XAGENT_S2S_API_BASE_URL` for Gmail Pub/Sub and A2A callers;
- keeps MCP OAuth and normal public/browser API configuration on
  `XAGENT_PUBLIC_API_BASE_URL`;
- adds `GET /api/deployment-config` as a hosting-layer contract;
- makes Agent and Workforce API/SDK and widget artifacts consume the advertised
  deployment origin;
- makes Agent and Workforce public links establish their region through the
  canonical application origin; and
- adds an idempotent command for reconciling existing Gmail Pub/Sub
  subscriptions.

## Why

The canonical XAgent Cloud edge selects a region from browser routing state.
Gmail Pub/Sub, A2A clients, SDK consumers, and widget guests do not carry that
state. Sending those clients to the canonical edge can therefore return:

```json
{"error":"region_required","product":"xagent","env":"prod","allowed_regions":["sg","au"]}
```

The direct `<region>-origin` ingress is the correct target for machine clients
and widget assets. Public share pages are different: they remain browser pages
on the canonical application host, so their generated links first visit
`/change-region` and then open `/share/<token>`.

## URL ownership

| Consumer | URL source |
| --- | --- |
| Gmail Pub/Sub push endpoint and OIDC audience | `XAGENT_S2S_API_BASE_URL` |
| A2A Agent Card interface | `XAGENT_S2S_API_BASE_URL` |
| MCP OAuth | `XAGENT_PUBLIC_API_BASE_URL` |
| Standalone API/SDK snippets | Existing API runtime configuration |
| Standalone widget snippets | Existing browser origin |
| Hosted Agent/Workforce API, SDK, and widget artifacts | `deployment_origin` from `/api/deployment-config` |
| Hosted public share links | `app_origin` plus `region` from `/api/deployment-config` |

When `XAGENT_S2S_API_BASE_URL` is unset, it falls back to the normalized
`XAGENT_PUBLIC_API_BASE_URL`.

Standalone `/api/deployment-config` deliberately leaves `deployment_origin` and
`region` unset. This preserves installations where API and widget assets use
different origins. A hosting layer such as XAgent SaaS overrides the endpoint
when both are exposed through one regional ingress.

## Gmail behavior

New Gmail mailbox provisioning uses the S2S base for:

[...]

PR #1147, opened 2026-08-05
fix(gmail): drop DISTINCT from the trigger lookup over json config

codeacme17:
## Summary

Connecting (or reconnecting) a Gmail account on any PostgreSQL deployment
fails. The OAuth token is stored correctly, then the callback dies and the
browser shows `Authentication Failed`:

```
(psycopg2.errors.UndefinedFunction) could not identify an equality operator
for type json
[SQL: SELECT DISTINCT agent_triggers.config AS agent_triggers_config,
 agent_triggers.resource_id AS agent_triggers_resource_id
 FROM agent_triggers WHERE agent_triggers.type = %(type_1)s
 AND agent_triggers.enabled IS true AND (...)]
```

## Root cause

`_referenced_gmail_oauth_account_ids` projects `agent_triggers.config`, which
is a `Column(JSON)`, and wrapped the query in `SELECT DISTINCT`. PostgreSQL
has no equality operator for `json`, so it cannot deduplicate that
projection. SQLite has no such restriction, which is why local runs and CI
stayed green.

Introduced by 478333c0 (#1024).

## Blast radius

Three call sites share the helper, so the failure is not limited to the
connect flow:

| Call site | Trigger | Effect |
| --- | --- | --- |
| `best_effort_provision_gmail_watches_for_user` | after an OAuth Gmail (re)connect | callback 500s; the docstring's "never raised" guarantee is broken because the lookup runs outside the per-account guard |
| `sweep_gmail_provisioning` | periodic retry of pending/failed watches | sweep raises, so failed watches are never retried |
| `reconcile_gmail_push_endpoints` | the operator command in xagent-saas#322's rollout steps 4 and 5 | reconciliation cannot run |

Verified on staging (`main-f1eeb5f7`, PostgreSQL 16): the Gmail token lands in
`user_oauth` with `gmail.modify`, and the callback still fails. Production
images predate #1024 and are unaffected today, but the next deploy that
carries #1024 - the same deploy that fixes regional push callbacks - would
break Gmail connects in both regions.

## Fix

Drop the `.distinct()`. It was redundant: the caller accumulates matches into
`referenced: set[int]`, so duplicate candidate rows already collapse in
Python. Results are identical on both engines and no schema change is needed.

## Tests

- New `test_postgresql_gmail_trigger_lookup_selects_the_json_config` runs the
  helper against real PostgreSQL with two triggers bound to one account, so it
  covers both the rejected operator and the surviving deduplication. It fails
  with the production error before this change.
- The Postgres job's selector was `-k contend`, which would never have picked
  the new test up, so it is now
  `-k "contend or postgresql_gmail_trigger_lookup"`. The job's path filter
  already includes `gmail_provisioning.py` and its test file, so it runs for
  this PR.
- `test_reconcile_bounds_gmail_trigger_lookup_to_each_watch_page` asserted
  `SELECT DISTINCT` was present. That expectation encoded the bug; it now
  asserts DISTINCT is absent, which keeps an always-on guard in the SQLite
  job. Its original intent - one bounded `IN (...)` lookup per watch page - is
  unchanged.
- Full file: 49 passed, 4 PostgreSQL-only skipped locally. Ruff, ruff format,
  mypy, isort and codespell pass.

### DIFF
--- a/.github/workflows/test-migrations.yml
+++ b/.github/workflows/test-migrations.yml
@@ -217,8 +217,7 @@ jobs:
         if: needs.detect-migration-changes.outputs.should-test == 'true'
         run: |
           pip install pytest-asyncio
-          pytest tests/web/services/test_gmail_provisioning.py \
-            -k "contend or postgresql_gmail_trigger_lookup" -q
+          pytest tests/web/services/test_gmail_provisioning.py -m postgresql -q
         env:
           XAGENT_TEST_POSTGRES_URL: postgresql://xagent:xagent@localhost:5432/xagent_test
 
--- a/.github/workflows/test-migrations.yml
+++ b/.github/workflows/test-migrations.yml
@@ -213,11 +213,12 @@ jobs:
         env:
           DATABASE_URL: postgresql://xagent:xagent@localhost:5432/xagent_test
 
-      - name: Test Gmail provisioning row-lock contention (Postgres-only)
+      - name: Test Gmail provisioning Postgres-only behavior
         if: needs.detect-migration-changes.outputs.should-test == 'true'
         run: |
           pip install pytest-asyncio
-          pytest tests/web/services/test_gmail_provisioning.py -k contend -q
+          pytest tests/web/services/test_gmail_provisioning.py \
+            -k "contend or postgresql_gmail_trigger_lookup" -q
         env:
           XAGENT_TEST_POSTGRES_URL: postgresql://xagent:xagent@localhost:5432/xagent_test
 
--- a/src/xagent/web/services/gmail_provisioning.py
+++ b/src/xagent/web/services/gmail_provisioning.py
@@ -255,6 +255,10 @@ def _referenced_gmail_oauth_account_ids(
         AgentTrigger.config["oauth_account_id"].as_string(),
         String,
     )
+    # No DISTINCT here: the projection selects the ``json`` config column, and
+    # PostgreSQL has no equality operator for ``json``, so SELECT DISTINCT
+    # fails there while passing on SQLite. Duplicate rows are harmless because
+    # the loop below accumulates into a set.
     candidate_rows = (
         db.query(AgentTrigger.config, AgentTrigger.resource_id)
         .filter(
@@ -265,7 +269,6 @@ def _referenced_gmail_oauth_account_ids(
                 func.lower(AgentTrigger.resource_id).in_(account_ids_by_email),
             ),
         )
-        .distinct()
         .all()
     )
 
--- a/tests/web/services/test_gmail_provisioning.py
+++ b/tests/web/services/test_gmail_provisioning.py
@@ -2068,6 +2068,7 @@ def pg_session():
         Base.metadata.drop_all(bind=engine)
 
 
+@pytest.mark.postgresql
 def test_postgresql_transition_lock_contends_per_oauth_account(
     pg_session: Session,
 ) -> None:
@@ -2131,6 +2132,7 @@ def test_postgresql_transition_lock_contends_per_oauth_account(
     assert errors == []
 
 
+@pytest.mark.postgresql
 def test_release_and_reprovision_contend_on_the_watch_state_lock(
     pg_session: Session,
 ) -> None:
@@ -2277,6 +2279,7 @@ def test_first_time_creation_race_adopts_winner_row(
     assert db_session.query(GmailWatchState).count() == 1
 
 
+@pytest.mark.postgresql
 def test_concurrent_first_time_creations_race_on_the_unique_constraint(
     pg_session: Session,
 ) -> None:
@@ -2343,7 +2346,8 @@ def test_concurrent_first_time_creations_race_on_the_unique_constraint(
     assert rows[0].email == "owner@gmail.example"
 
 
-def test_postgresql_gmail_trigger_lookup_selects_the_json_config(
+@pytest.mark.postgresql
+def test_gmail_trigger_lookup_resolves_bindings_on_postgresql(
     pg_session: Session,
 ) -> None:
     """Resolve trigger bindings on the engine that rejects ``json`` equality.
--- a/tests/web/services/test_gmail_provisioning.py
+++ b/tests/web/services/test_gmail_provisioning.py
@@ -1545,7 +1545,9 @@ def test_reconcile_bounds_gmail_trigger_lookup_to_each_watch_page(
     assert result.changed == 2
     assert len(trigger_selects) == 2
     assert all(" IN (" in statement for statement in trigger_selects)
-    assert all("SELECT DISTINCT" in statement for statement in trigger_selects)
+    # The projection includes the ``json`` config column, which PostgreSQL
+    # cannot compare, so the lookup must never be a SELECT DISTINCT.
+    assert not any("DISTINCT" in statement.upper() for statement in trigger_selects)
     assert len(watch_selects) == 3
     assert all("LIMIT" in statement for statement in watch_selects)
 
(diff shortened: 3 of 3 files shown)


rogercloud (approved):
## Summary

This PR fixes a hard failure when connecting or reconnecting Gmail on any PostgreSQL deployment. `_referenced_gmail_oauth_account_ids` in `src/xagent/web/services/gmail_provisioning.py` projects the JSON `agent_triggers.config` column and wrapped the query in `SELECT DISTINCT`; PostgreSQL has no equality operator for the `json` type, so the query raised `psycopg2.errors.UndefinedFunction: could not identify an equality operator for type json`. SQLite tolerates DISTINCT over JSON, which is why local dev and CI stayed green while every real Postgres deployment broke. The fix drops `.distinct()`, adds a real-Postgres regression test, flips an existing test's DISTINCT assertion from asserting presence to asserting absence, and widens the CI selector so the new test actually executes against Postgres.

## Approach

**Sound.** The removal is safe rather than merely convenient: `candidate_rows` has exactly one consumer, a loop that accumulates into a Python `set[int]`, so the SQL-level DISTINCT was pure dead weight — and it has been dead weight since the commit that introduced it also introduced the set-based dedup. The query carries no `order_by` or `limit`, and none of the three call sites (the OAuth connect callback, the periodic retry sweep, the operator reconciliation command) depend on row ordering, row count, or pagination from this query. So dropping DISTINCT is behavior-preserving in every consumer.

The two obvious alternatives were correctly rejected:

- **Projecting only `resource_id` to avoid selecting the json column** would not be behavior-equivalent. The current code deserializes the JSON value and calls `int()` on it; extracting text via `->>` changes how malformed values parse (a JSON float `12.0` survives `int()`, the text `"12.0"` raises and falls through to email-based matching). That shifts which malformed configs take the fallback path — a real semantic change, and wrong to bundle into a hotfix.
- **Migrating the column from `JSON` to `JSONB`** (which does support equality) requires a schema migration, and the codebase declares `Column(JSON)` essentially everywhere. Correct call to keep that out of a hotfix.

## Findings

All surviving findings are minor or nit-level; none are blocking.

**1. Minor — `.github/workflows/test-migrations.yml:220-221` (test coverage / CI hygiene)**

The Gmail Postgres step selects tests with a hand-maintained `-k "contend or postgresql_gmail_trigger_lookup"` allowlist, and this PR extends that string to pick up its own new test. The allowlist already misses another Postgres-dependent test in the same file: `test_concurrent_first_time_creations_race_on_the_unique_constraint` (`tests/web/services/test_gmail_provisioning.py:2280`) takes the `pg_session` fixture but matches neither term. No other workflow sets `XAGENT_TEST_POSTGRES_URL`, and `ci.yml` runs with `-m "not slow and not postgresql"`, so that test has never actually run against real Postgres anywhere. A `postgresql` marker is already registered in `pyproject.toml:257`, and this same workflow already uses `-m postgresql` selection for a sibling job. Consider tagging `pg_session`-based tests with `@pytest.mark.postgresql` and switching this step to `-m postgresql`, so future Postgres-only tests are picked up automatically instead of by hand. Pre-existing pattern, not introduced here — but it is a one-line change to a line this PR already touches.

**2. Minor — `src/xagent/web/services/gmail_provisioning.py:1217-1231` (robustness, suggest as fast-follow)**

`best_effort_provision_gmail_watches_for_user` documents that "Failures are recorded on the watch state, never raised," but the initial `UserOAuth` query and the `_referenced_gmail_oauth_account_ids` call both run before any try/except — the per-account guard only begins later in the loop. The caller in `src/xagent/web/api/auth.py` invokes this after `db.commit()` with no local protection, and its broad `except Exception` turns anything escaping into the HTTP 500 "Authentication Failed" page — precisely the user-visible symptom this PR is fixing. This PR removes the one bug that was tripping that unprotected window, but the structural gap remains: any other DB error on those two calls reproduces the same page despite the "never raised" contract. Suggest a follow-up wrapping those two calls in the same try/except pattern used later in the function.

**3. Nit — `tests/web/services/test_gmail_provisioning.py:1550`**

The flipped assertion in `test_reconcile_bounds_gmail_trigger_lookup_to_each_watch_page` (`assert not any("DISTINCT" in statement.upper() ...)`) is correct and does catch the regression, but it is broader than the invariant it protects: it bans any DISTINCT on this query, not just DISTINCT over a json column. The explanatory comment already present makes this acceptable as-is; noting only for the record.

**4. Nit — `tests/web/services/test_gmail_provisioning.py:2346`**

`test_postgresql_gmail_trigger_lookup_selects_the_json_config` exercises the private helper `_referenced_gmail_oauth_account_ids` directly rather than the OAuth callback path where the failure actually surfaced. That is a reasonable trade-off given the cost of standing up a full Postgres integration test with a mocked Google client, and no existing test covers the public path against Postgres either. Also, the test name describes the mechanism ("selects the json config") rather than the guarantee ("does not use DISTINCT over a json column" / "works on PostgreSQL"). Cosmetic.

## Informational (non-blocking)

- **No recurrence of this bug class in scope.** Scanned the codebase for `.distinct()` (9 occurrences), `group_by(` (10 occurrences), and SQL `UNION` (0 occurrences). None of the others project a raw JSON column, so this specific failure mode does not repeat elsewhere.
- **Follow-up worth filing separately (outside this diff).** `src/xagent/web/api/monitor.py:68` applies `column.op("~?")(...)` to `TraceEvent.data`, declared `Column(JSON, ...)` at `src/xagent/web/models/task.py:451`. PostgreSQL defines `~?` for `jsonb`/`text` but not `json`, so this raises `operator does not exist: json ~? unknown` on real Postgres. The three call sites (`monitor.py:324`, `monitor.py:425`, `monitor.py:490`) are each wrapped in `try/except Exception` with a `0`/`[]` fallback, so requests do not crash — they silently return empty monitoring stats. Same bug class as the one fixed here (green on SQLite, broken on Postgres, no Postgres test). Not part of this PR; suggest a separate issue.

## Test results

- `tests/web/services/test_gmail_provisioning.py` against SQLite: **49 passed, 4 skipped** (the Postgres-only tests, correctly skipped without `XAGENT_TEST_POSTGRES_URL`).
- Same file against a real PostgreSQL 16 instance: **52 passed**, including all four Postgres-only tests.
- Regression guards verified to be load-bearing: re-adding `.distinct()` reproduces the exact production error and produces 2 failures (the new Postgres test and the flipped SQLite assertion). Both guards are real, not decorative.

## Recommendation

**Approve.** The diff is minimal and correctly scoped — one dead `.distinct()` removed, one assertion flipped, one real-Postgres regression test added, one CI selector widened — with an explanatory source comment on why DISTINCT cannot come back. Correctness, call-site completeness, ordering/pagination impact, performance, security, schema-migration necessity, backward compatibility, and title convention all check out. Finding 1 is worth folding in here since the line is already being edited; findings 2 and the monitor.py note are good fast-follows, and 3 and 4 are optional.

codeacme17:
Thanks for the review. Finding 1 is folded in as 26dee64c; findings 3 and 4 are addressed or answered below.

**Finding 1 (CI hygiene) — fixed, and taken further than a one-line edit.** Rather than extending the `-k` allowlist, the four `pg_session` tests now carry the already-registered `@pytest.mark.postgresql` marker and the step selects `-m postgresql`, matching the sibling task-command-transport step. That also picks up `test_concurrent_first_time_creations_race_on_the_unique_constraint`, which as you noted had never executed against real PostgreSQL anywhere — it runs for the first time in this PR. Future Postgres-only tests in this file are now selected on their marker instead of by hand. `ci.yml`'s `-m "not slow and not postgresql"` keeps excluding them from the SQLite job, which is the intent.

**Finding 2 (unprotected window in `best_effort_provision_gmail_watches_for_user`) — agreed, and deliberately left out of this diff.** The `UserOAuth` query and the `_referenced_gmail_oauth_account_ids` call do sit outside the per-account guard, so the "never raised" contract still depends on those two calls not failing. Widening the guard changes behavior in an error path (what gets recorded on the watch state, and what the callback returns when the lookup itself fails), which is a different question from removing a broken query and wants its own review. Tracking as a fast-follow rather than expanding a hotfix.

**Finding 3 (assertion is broader than the invariant) — acknowledged, keeping as-is.** Banning DISTINCT outright on that query is stricter than "no DISTINCT over a json column", but the narrower assertion would have to inspect the projection to be meaningful, and the comment records why the ban exists. The real invariant is enforced by the PostgreSQL test; this one is the always-on tripwire in the SQLite job.

**Finding 4 (test name describes the mechanism) — renamed.** Now `test_gmail_trigger_lookup_resolves_bindings_on_postgresql`, which states the guarantee. It still calls the private helper directly for exactly the reason you gave: covering the OAuth callback path against PostgreSQL means standing up a mocked Google client plus the full request path, and no existing test covers that path on PostgreSQL either. Worth doing, but not as a rider on this fix.

**`monitor.py` `~?` on a `json` column — confirmed, same failure class, separate issue.** `TraceEvent.data` is `Column(JSON)` and PostgreSQL only defines `~?` for `jsonb`/`text`, so those three call sites swallow `operator does not exist: json ~? unknown` and return empty monitoring stats instead of crashing. Silent-wrong rather than user-visible, which is arguably worse. Not in scope here.

One scheduling note for whoever merges: production images currently predate #1024, so this bug is not live in production yet. The next deploy carrying #1024 is also the deploy that fixes regional Gmail push callbacks, so this needs to land ahead of it or that deploy breaks every Gmail connect in both regions.

PR #1148, opened 2026-08-05
feat: make sidebar nav groups collapsible with configurable default state

### LINKED ISSUE #1144: feat: Add collapsible sections to the left navigation panel
### Description

Improve the left navigation panel usability by allowing some sections to be collapsible.

Certain navigation components should be collapsed by default to reduce clutter and provide more visibility for important content such as chat history.

---

### Problem Statement

The left navigation panel currently contains too many items, making the interface crowded.

As a result:

* Chat history is less visible.
* Users need to scroll more to find conversations.
* Important navigation items compete for limited space.

---

### Expected Behavior

Users should be able to:

* Expand or collapse specific sections in the left navigation panel.
* See a cleaner and more organized navigation layout.
* Access chat history more easily.

Some sections should be collapsed by default to optimize available space.

yiboyasss:
## Summary
- Nav groups in the left sidebar ("Agent Development", "Resources") are now
  individually collapsible, matching the existing History section's
  chevron-toggle UX. This addresses the sidebar feeling crowded and pushing
  the task history list out of view.
- Which groups start collapsed is driven by an optional `defaultCollapsed`
  flag on `NavigationGroup` (set on "Resources") instead of a hardcoded group
  title, so nav groups injected via the `extra-nav` overlay (e.g. in
  xagent-saas) can opt into a collapsed-by-default state without any change
  to this repo.
- Fixed a missing dependency in the "auto-load next page if content doesn't
  fill the container" effect so it re-evaluates after a group toggle changes
  the scroll container's height.

## Test plan
- [x] `tsc --noEmit` passes with no new errors
- [x] Registered a test account and logged in locally; confirmed "Resources"
      starts collapsed, "Agent Development" starts expanded, and both toggle
      independently
- [x] Confirmed task history list is now visible without scrolling once
      "Resources" is collapsed
- [x] Checked browser console/network — no errors, `/api/chat/tasks` loads
      normally

### DIFF
--- a/frontend/src/components/layout/sidebar.tsx
+++ b/frontend/src/components/layout/sidebar.tsx
@@ -241,6 +241,20 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
 
   const [isExpanded, setIsExpanded] = useState(false)
   const [expandedMenus, setExpandedMenus] = useState<string[]>(["/agent"]) // Use href as a stable key
+  const [collapsedGroups, setCollapsedGroups] = useState<Set<string>>(
+    () => new Set(navigationGroups.filter(g => g.defaultCollapsed).map(g => g.title))
+  )
+  const toggleGroupCollapse = useCallback((title: string) => {
+    setCollapsedGroups(prev => {
+      const next = new Set(prev)
+      if (next.has(title)) {
+        next.delete(title)
+      } else {
+        next.add(title)
+      }
+      return next
+    })
+  }, [])
   const [showUserMenu, setShowUserMenu] = useState(false)
   const [isAboutOpen, setIsAboutOpen] = useState(false)
   const sidebarRef = useRef<HTMLDivElement | null>(null)
@@ -521,7 +535,7 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
       }, 100)
       return () => clearTimeout(timer)
     }
-  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded])
+  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, collapsedGroups])
 
   useEffect(() => {
     if (isHistoryExpanded) {
@@ -723,11 +737,18 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
           className="z-10 bg-transparent py-1"
         >
           {/* Groups */}
-          {navigationGroups.map((group, groupIndex) => (
+          {navigationGroups.map((group, groupIndex) => {
+            const isGroupCollapsed = collapsedGroups.has(group.title)
+            return (
             <div key={group.title} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
-              <div className="px-2 pb-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em]">
-                {group.titleKey ? t(group.titleKey) : group.title}
+              <div
+                className="px-2 pb-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em] flex items-center justify-between cursor-pointer select-none"
+                onClick={() => toggleGroupCollapse(group.title)}
+              >
+                <span>{group.titleKey ? t(group.titleKey) : group.title}</span>
+                {isGroupCollapsed ? <ChevronRight className="h-3 w-3" /> : <ChevronDown className="h-3 w-3" />}
               </div>
+              {!isGroupCollapsed && (
               <div className="space-y-0.5">
                 {group.items.map((item: NavigationItem) => {
                   const isActive = isItemActive(item)
@@ -803,8 +824,10 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
                   )
                 })}
               </div>
+              )}
             </div>
-          ))}
+            )
+          })}
         </nav>
 
         {/* History Section */}
--- a/frontend/src/components/layout/sidebar.tsx
+++ b/frontend/src/components/layout/sidebar.tsx
@@ -241,19 +241,12 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
 
   const [isExpanded, setIsExpanded] = useState(false)
   const [expandedMenus, setExpandedMenus] = useState<string[]>(["/agent"]) // Use href as a stable key
-  const [collapsedGroups, setCollapsedGroups] = useState<Set<string>>(
-    () => new Set(navigationGroups.filter(g => g.defaultCollapsed).map(g => g.title))
-  )
-  const toggleGroupCollapse = useCallback((title: string) => {
-    setCollapsedGroups(prev => {
-      const next = new Set(prev)
-      if (next.has(title)) {
-        next.delete(title)
-      } else {
-        next.add(title)
-      }
-      return next
-    })
+  const [groupCollapseOverrides, setGroupCollapseOverrides] = useState<Record<string, boolean>>({})
+  const toggleGroupCollapse = useCallback((title: string, defaultCollapsed?: boolean) => {
+    setGroupCollapseOverrides(prev => ({
+      ...prev,
+      [title]: !(title in prev ? prev[title] : !!defaultCollapsed),
+    }))
   }, [])
   const [showUserMenu, setShowUserMenu] = useState(false)
   const [isAboutOpen, setIsAboutOpen] = useState(false)
@@ -535,7 +528,7 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
       }, 100)
       return () => clearTimeout(timer)
     }
-  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, collapsedGroups])
+  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, groupCollapseOverrides])
 
   useEffect(() => {
     if (isHistoryExpanded) {
@@ -738,12 +731,23 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
         >
           {/* Groups */}
           {navigationGroups.map((group, groupIndex) => {
-            const isGroupCollapsed = collapsedGroups.has(group.title)
+            const isGroupCollapsed = group.title in groupCollapseOverrides
+              ? groupCollapseOverrides[group.title]
+              : !!group.defaultCollapsed
             return (
             <div key={group.title} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
               <div
-                className="px-2 pb-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em] flex items-center justify-between cursor-pointer select-none"
-                onClick={() => toggleGroupCollapse(group.title)}
+                role="button"
+                tabIndex={0}
+                aria-expanded={!isGroupCollapsed}
+                className="px-2 pb-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em] flex items-center justify-between cursor-pointer select-none outline-none"
+                onClick={() => toggleGroupCollapse(group.title, group.defaultCollapsed)}
+                onKeyDown={(e) => {
+                  if (e.key === "Enter" || e.key === " ") {
+                    e.preventDefault()
+                    toggleGroupCollapse(group.title, group.defaultCollapsed)
+                  }
+                }}
               >
                 <span>{group.titleKey ? t(group.titleKey) : group.title}</span>
                 {isGroupCollapsed ? <ChevronRight className="h-3 w-3" /> : <ChevronDown className="h-3 w-3" />}
--- a/frontend/src/components/layout/sidebar.tsx
+++ b/frontend/src/components/layout/sidebar.tsx
@@ -246,12 +246,12 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
   // an active-route auto-expand, or a previous toggle) so a click always changes what
   // the user sees — recomputing from defaultCollapsed here would make the first click
   // a no-op whenever the group was auto-expanded by the active route.
-  const toggleGroupCollapse = useCallback((title: string, currentCollapsed: boolean) => {
+  const toggleGroupCollapse = (key: string, currentCollapsed: boolean) => {
     setGroupCollapseOverrides(prev => ({
       ...prev,
-      [title]: !currentCollapsed,
+      [key]: !currentCollapsed,
     }))
-  }, [])
+  }
   const [showUserMenu, setShowUserMenu] = useState(false)
   const [isAboutOpen, setIsAboutOpen] = useState(false)
   const sidebarRef = useRef<HTMLDivElement | null>(null)
@@ -532,9 +532,10 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
       }, 100)
       return () => clearTimeout(timer)
     }
-    // groupCollapseOverrides isn't read above — it's a re-measure trigger: toggling a group
+    // groupCollapseOverrides and pathname aren't read above — they're re-measure triggers:
+    // toggling a group, or navigating to/from a route that auto-expands/collapses a group,
     // changes contentScrollRef's scrollHeight, so this effect must re-run to re-check the fill.
-  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, groupCollapseOverrides])
+  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, groupCollapseOverrides, pathname])
 
   useEffect(() => {
     if (isHistoryExpanded) {
@@ -737,30 +738,26 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
         >
           {/* Groups */}
           {navigationGroups.map((group, groupIndex) => {
+            // Groups without a stable `id` (e.g. extra-nav overlays) fall back to `title` —
+            // matches the pre-`id` behavior for anything that hasn't opted in yet.
+            const groupKey = group.id ?? group.title
             // An active route overrides only the *default* collapsed state, so the current
             // page is never hidden inside a collapsed group on first load — but an explicit
             // user toggle always wins over that, so the group can still be collapsed manually.
-            const isGroupCollapsed = group.title in groupCollapseOverrides
-              ? groupCollapseOverrides[group.title]
+            const isGroupCollapsed = groupKey in groupCollapseOverrides
+              ? groupCollapseOverrides[groupKey]
               : !!group.defaultCollapsed && !group.items.some(isItemActive)
             return (
-              <div key={group.title} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
-                <div
-                  role="button"
-                  tabIndex={0}
+              <div key={groupKey} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
+                <button
+                  type="button"
                   aria-expanded={!isGroupCollapsed}
-                  className="px-2 pb-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em] flex items-center justify-between cursor-pointer select-none rounded transition-colors hover:bg-accent focus:outline-none focus-visible:ring-2 focus-visible:ring-ring"
-                  onClick={() => toggleGroupCollapse(group.title, isGroupCollapsed)}
-                  onKeyDown={(e) => {
-                    if (e.key === "Enter" || e.key === " ") {
-                      e.preventDefault()
-                      toggleGroupCollapse(group.title, isGroupCollapsed)
-                    }
-                  }}
+                  className="w-full px-2 py-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em] flex items-center justify-between text-left cursor-pointer select-none rounded transition-colors hover:bg-accent focus:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+                  onClick={() => toggleGroupCollapse(groupKey, isGroupCollapsed)}
                 >
                   <span>{group.titleKey ? t(group.titleKey) : group.title}</span>
                   {isGroupCollapsed ? <ChevronRight className="h-3 w-3" /> : <ChevronDown className="h-3 w-3" />}
-                </div>
+                </button>
                 {!isGroupCollapsed && (
                   <div className="space-y-0.5">
                     {group.items.map((item: NavigationItem) => {
--- a/frontend/src/components/layout/sidebar.tsx
+++ b/frontend/src/components/layout/sidebar.tsx
@@ -242,10 +242,14 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
   const [isExpanded, setIsExpanded] = useState(false)
   const [expandedMenus, setExpandedMenus] = useState<string[]>(["/agent"]) // Use href as a stable key
   const [groupCollapseOverrides, setGroupCollapseOverrides] = useState<Record<string, boolean>>({})
-  const toggleGroupCollapse = useCallback((title: string, defaultCollapsed?: boolean) => {
+  // Flip from the collapsed state currently rendered (which may come from the default,
+  // an active-route auto-expand, or a previous toggle) so a click always changes what
+  // the user sees — recomputing from defaultCollapsed here would make the first click
+  // a no-op whenever the group was auto-expanded by the active route.
+  const toggleGroupCollapse = useCallback((title: string, currentCollapsed: boolean) => {
     setGroupCollapseOverrides(prev => ({
       ...prev,
-      [title]: !(title in prev ? prev[title] : !!defaultCollapsed),
+      [title]: !currentCollapsed,
     }))
   }, [])
   const [showUserMenu, setShowUserMenu] = useState(false)
@@ -528,6 +532,8 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
       }, 100)
       return () => clearTimeout(timer)
     }
+    // groupCollapseOverrides isn't read above — it's a re-measure trigger: toggling a group
+    // changes contentScrollRef's scrollHeight, so this effect must re-run to re-check the fill.
   }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, groupCollapseOverrides])
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/layout/sidebar.tsx
```diff
@@ -723,11 +730,29 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
           className="z-10 bg-transparent py-1"
         >
           {/* Groups */}
-          {navigationGroups.map((group, groupIndex) => (
+          {navigationGroups.map((group, groupIndex) => {
+            const isGroupCollapsed = group.title in groupCollapseOverrides
+              ? groupCollapseOverrides[group.title]
+              : !!group.defaultCollapsed
```
rogercloud: **[major]** A group that is collapsed by default hides the active route, and nothing forces it open. `isItemActive(item)` is only evaluated inside the `{!isGroupCollapsed && (...)}` branch below, so for a collapsed group it never runs. Since "Resources" ships `defaultCollapsed: true` and owns `/kb`, `/models`, `/memory`, `/skill-hub`, and `/api-keys`, landing on or deep-linking to any of those five routes in a fresh session gives a sidebar with no active-route indication at all, and the collapsed header gives no hint the current page lives inside it.

Suggest letting an active route override the *default* only, so an explicit user collapse still wins and the user isn't locked out of collapsing the section they're on:

```ts
const isGroupCollapsed =
  (group.title in groupCollapseOverrides
    ? groupCollapseOverrides[group.title]
    : !!group.defaultCollapsed) && !group.items.some(isItemActive)
```

frontend/src/components/layout/sidebar.tsx
```diff
@@ -723,11 +730,29 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
...
+            return (
             <div key={group.title} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
-              <div className="px-2 pb-1.5 text-[10.5px] font-semibold text-muted-foreground uppercase tracking-[0.08em]">
-                {group.titleKey ? t(group.titleKey) : group.title}
+              <div
+                role="button"
+                tabIndex={0}
```
rogercloud: **[major]** This header is now genuinely keyboard-focusable (`role="button"` + `tabIndex={0}` — good, and new here), but the className ends in `outline-none` with no `focus:`/`focus-visible:` replacement and no global rule covering this element, so a keyboard user tabbing onto it sees nothing at all (WCAG 2.4.7). This is the still-open half of the earlier accessibility feedback. The codebase already has the convention for this exact pattern — `frontend/src/components/workforce/workforce-config-panel.tsx:266` and `frontend/src/components/workforce/workforce-deploy-hub-dialog.tsx:107` both use `focus:outline-none focus-visible:ring-2 focus-visible:ring-ring` on their `role="button"` divs; this is the only interactive-div site that strips the outline without a replacement. Append `focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-1`, or just drop `outline-none`.

**[minor]** Same line: `cursor-pointer` with no `hover:bg-*` / `transition-colors`, while the History chevron toggle this is meant to match (further down this file, unchanged) uses `hover:bg-accent rounded transition-colors`. History also only makes the small chevron clickable, whereas this makes the whole header row clickable — a fine UX choice on its own, but with no hover feedback the wide clickable area is less discoverable than the pattern it mirrors.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -521,7 +528,7 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
       }, 100)
       return () => clearTimeout(timer)
     }
-  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded])
+  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, groupCollapseOverrides])
```
rogercloud: **[minor]** `groupCollapseOverrides` here is correct and does work — a toggle changes the scroll container's height and the `{...prev}` spread guarantees a new object identity, so the effect re-measures — but it is never read in the effect body. `react-hooks/exhaustive-deps` won't defend an unused-but-present dependency, so a future "remove unused dep" cleanup could silently drop it and regress the re-measure with no test or lint catching it. Worth a one-line comment noting it's a re-measure trigger, not a value the effect consumes.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -723,11 +730,29 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
           className="z-10 bg-transparent py-1"
         >
           {/* Groups */}
-          {navigationGroups.map((group, groupIndex) => (
+          {navigationGroups.map((group, groupIndex) => {
+            const isGroupCollapsed = group.title in groupCollapseOverrides
+              ? groupCollapseOverrides[group.title]
+              : !!group.defaultCollapsed
+            return (
             <div key={group.title} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
```
rogercloud: **[minor, cosmetic]** `mb-4` here and `pb-1.5` on the header apply regardless of `isGroupCollapsed`; only `group.items` is conditionally rendered, so a collapsed group still reserves ~22px. Small next to the ~150-190px reclaimed by hiding Resources' five items — noting it, not blocking.

frontend/src/lib/sidebar-navigation.ts
```diff
@@ -39,6 +39,8 @@ export interface NavigationGroup {
     title: string
     titleKey?: TranslationKey
     items: NavigationItem[]
+    /** Whether this group's items are hidden behind a collapse toggle by default. */
+    defaultCollapsed?: boolean
```
rogercloud, line 43: **[minor, forward-looking]** With this field, `group.title` becomes the key for both React reconciliation and the new `groupCollapseOverrides` map, and `NavigationGroup` still has no stable `id`. `extra-nav.ts` is an empty stub today so there's no live collision, but that hook exists precisely so downstream repos can inject groups — a downstream group titled "Resources" would then silently share both the React key and the collapse state with the built-in one. Adding an `id` field and keying on it would close this off cheaply. Non-blocking.

rogercloud (changes requested):
## Summary

This PR makes the sidebar's navigation groups collapsible so the left nav stops crowding out chat history (issue #1144). It adds a `defaultCollapsed?: boolean` field to `NavigationGroup` (set to `true` on "Resources"), turns each group's title row into a clickable/keyboard-operable toggle with a chevron, and conditionally renders the group's items. Collapse state lives in the sidebar component as `groupCollapseOverrides: Record<string, boolean>` keyed by group title, holding only explicit user toggles; the effective collapsed state is derived at render time, falling back to the group's `defaultCollapsed`. Two files are touched: `frontend/src/components/layout/sidebar.tsx` and `frontend/src/lib/sidebar-navigation.ts`.

Commit 2 (`fix: derive sidebar group collapse state during render`) responds to the three body-only points from the earlier automated review of commit 1. It replaces the `collapsedGroups: Set<string>` lazy-`useState` snapshot with `groupCollapseOverrides` (explicit toggles only) and derives the collapsed state during render, and it adds `role="button"`, `tabIndex={0}`, `aria-expanded`, and an Enter/Space `onKeyDown` handler to the group header.

Prior-finding status:

- **Dynamic nav groups ignoring `defaultCollapsed`** — **FIXED**. Because `groupCollapseOverrides` only ever stores explicit user toggles, a group that appears after mount (e.g. one resolved through `extraNav(user)` once user data loads) has no override key and correctly picks up its own `defaultCollapsed` on the next render. No stale snapshot remains.
- **Effect dependency array** — **FIXED**, with a caveat: the dependency is present and does the right thing mechanically, but nothing in the effect body references it (see minor #2).
- **Keyboard handlers + ARIA on the clickable header** — **PARTIAL**. Keyboard operability and screen-reader semantics are genuinely fixed and correctly implemented (the key handler scopes `preventDefault` and routes through the same `toggleGroupCollapse` as `onClick`). What is still missing is a *visible* focus indicator: the header's className strips the outline without adding any `focus-visible:` replacement, so the element is now focusable but shows nothing when focused. That residual gap is reported once below as major #2.

Deriving the collapsed state at render from "explicit override, else `defaultCollapsed`" is the structurally correct model for this feature, not a symptom patch — user intent wins where it exists, and un-toggled groups always reflect the current configuration, including groups injected after mount. Leaving the existing `isHistoryExpanded` toggle un-generalized is also the right call; the History section carries search visibility, lazy loading, and pagination concerns that a plain collapse flag does not model. There is no duplicated collapse logic elsewhere: `sidebar.tsx` is the only render site for `NavigationGroup`.

Two reservations at the design level:

1. **Collapse is currently modelled as independent of routing.** A group can be collapsed while it owns the active route, and since `isItemActive` is only evaluated inside the rendered-items branch, the active page then has no indication anywhere in the sidebar. Because "Resources" ships `defaultCollapsed: true` and owns five real routes, this affects the PR's own primary shipped case on a fresh session. The derived-state model makes this cheap to fix correctly (see major #1) — the point is that "which group is collapsed" should take the current route into account, not only stored/default state.
2. **`group.title` is now load-bearing as an identity key.** It is both the React key and the `groupCollapseOverrides` key, and `NavigationGroup` has no stable `id` (`frontend/src/lib/sidebar-navigation.ts:38-44`). `frontend/src/lib/extra-nav.ts` is an empty stub today, so there is no live collision risk — this is forward-looking. But that file exists precisely so downstream repos can inject additional groups, and a downstream group titled "Resources" would silently share both the React key and the toggle state with the built-in one. Adding an `id` field and keying both on it would close this off cheaply. Non-blocking.

## Findings

### major

**1. Active route is hidden when its group is collapsed by default, with no force-expand** — `frontend/src/components/layout/sidebar.tsx:733-736`

`isGroupCollapsed` is computed with no exception for a group containing the active route, and `isItemActive(item)` is only reached inside the `{!isGroupCollapsed && (...)}` branch, so it is never evaluated for a collapsed group. "Resources" ships `defaultCollapsed: true` and owns `/kb`, `/models`, `/memory`, `/skill-hub`, and `/api-keys` (`frontend/src/lib/sidebar-navigation.ts:143-179`). Landing on or deep-linking to any of those five routes in a fresh session therefore renders a sidebar with zero active-route indication, and the collapsed header gives no hint that the current page lives inside it. There is no mitigating logic anywhere — no active-group auto-expand, no URL-derived initial state.

Suggested fix — let an active route override the *default* only, so an explicit user collapse still wins and the user is not locked out of collapsing the section they are on:

```ts
const isGroupCollapsed =
  (group.title in groupCollapseOverrides
    ? groupCollapseOverrides[group.title]
    : !!group.defaultCollapsed) && !group.items.some(isItemActive)
```

**2. The new keyboard-focusable header has no visible focus indicator** — `frontend/src/components/layout/sidebar.tsx:739-750`

The header now carries `role="button"` and `tabIndex={0}`, which correctly makes it focusable (new in this PR — the base version had no interactivity here at all). But its className ends in `cursor-pointer select-none outline-none` with no `focus:` or `focus-visible:` ring, and no global CSS rule covers this element, so a keyboard user tabbing onto it gets no visual feedback (WCAG 2.4.7). This is the residual half of the earlier accessibility point. The project already has the convention for exactly this pattern: `frontend/src/components/workforce/workforce-config-panel.tsx:266` and `frontend/src/components/workforce/workforce-deploy-hub-dialog.tsx:107` both pair `role="button"` + `tabIndex={0}` with `focus:outline-none focus-visible:ring-2 focus-visible:ring-ring`; this is the only interactive-div site in the codebase that strips the outline without a replacement. Either append `focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-1` or drop `outline-none`.

### minor

**1. No test coverage for the new feature** — `frontend/src/components/layout/sidebar.test.tsx`

The sidebar test mocks `getNavigationGroupsForUser` to return `[]`, so the new collapsible-group render path never executes in CI. The file *is* wired into CI (`vitest.widget.config.ts`, `npm run test:widget:coverage`), but the coverage floor for `sidebar.tsx` is low enough that the mocked-out path still passes. `frontend/src/lib/sidebar-navigation.test.ts` has no assertion on the new `defaultCollapsed` field (and is not currently wired into CI). Neither test file is touched by this diff. Suggest adding `expect(resources?.defaultCollapsed).toBe(true)` to the navigation test, plus a sidebar render test asserting the initial collapsed state, `aria-expanded="false"`, and toggle behavior (which requires un-mocking the navigation data).

**2. `groupCollapseOverrides` is a dependency the effect never reads** — `frontend/src/components/layout/sidebar.tsx:531`

Adding it to the auto-load-next-page effect's deps is correct and works: each toggle changes the scroll container's height, and the `{...prev}` spread guarantees a fresh object identity, so the effect re-measures (no infinite-loop risk). But the value is never referenced in the effect body, and `react-hooks/exhaustive-deps` does not defend an unused-but-present dependency, so a future "remove unused dep" cleanup could silently delete it and regress the re-measure with nothing catching it. A one-line comment stating that it is a re-measure trigger rather than a value the effect reads would protect it.

**3. No hover affordance on the new clickable header** — `frontend/src/components/layout/sidebar.tsx:741`

The header has `cursor-pointer` but no `hover:bg-*` / `transition-colors`, whereas the existing History chevron toggle it is meant to match (`frontend/src/components/layout/sidebar.tsx:895-896`, unchanged here) uses `hover:bg-accent rounded transition-colors`. There is also a hit-area mismatch: History makes only the small chevron clickable, while this makes the whole header row clickable. The larger target is a defensible UX choice, but with zero hover feedback the full-width clickable area is less discoverable than the pattern it mirrors. Aligning the hover style at minimum would help.

**4. Collapsed groups still reserve most of their vertical spacing** — `frontend/src/components/layout/sidebar.tsx:738,743`

The wrapper's `mb-4` and the header's `pb-1.5` apply unconditionally; only `group.items` is conditionally rendered. Practical impact is small (roughly 22px retained against ~150-190px reclaimed by hiding Resources' five items), so the PR's goal is still substantially met — cosmetic, low priority.

**5. JSX not re-indented after the arrow-body to block-body conversion** — `frontend/src/components/layout/sidebar.tsx:733-834`

`.map((group, groupIndex) => (...))` became a block body with `return (...)` so `isGroupCollapsed` could be computed, but the ~95 lines of returned JSX were not shifted in one level, and the `{!isGroupCollapsed && (` wrapper's children are not indented either. There is no prettier/biome config in `frontend/`, so nothing will auto-fix this; it needs a manual re-indent. No functional impact.

**6. Optional nit, pre-existing scope:** the title-based keying constraint on `defaultCollapsed` (and on the rest of the `extra-nav` / `NavigationGroup` overlay contract) is undocumented. This gap predates the PR — it has existed since `extra-nav.ts` was introduced — and this change merely extends it by one field, so treat this purely as a "while you're here" opportunity, not a requested change.

Lean already.

## Testing / verification

`tsc`, `eslint`, and `vitest` were run during verification and are clean — this PR introduces no new type, lint, or test errors. Note that the accessibility and UX findings above (focus ring, hover affordance, collapsed-group spacing, hidden active route) were confirmed by reading the code and comparing against the project's own conventions elsewhere; they were not verified visually in a browser.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -723,88 +736,111 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
...
-                            className={cn(
-                              "group flex items-center justify-between px-2.5 py-2 text-[13.5px] transition-colors relative w-full",
-                              isActive ? activeStyle : inactiveStyle
+          {navigationGroups.map((group, groupIndex) => {
+            // An active route overrides only the *default* collapsed state, so the current
+            // page is never hidden inside a collapsed group on first load — but an explicit
+            // user toggle always wins over that, so the group can still be collapsed manually.
+            const isGroupCollapsed = group.title in groupCollapseOverrides
+              ? groupCollapseOverrides[group.title]
+              : !!group.defaultCollapsed && !group.items.some(isItemActive)
```
rogercloud, line 749: The active-route auto-expand added in this commit depends on item hrefs matching real pathnames, but two admin hrefs have trailing slashes: "User Management" (`/users/`) and "MCP Management" (`/admin-mcp/`) in `frontend/src/lib/sidebar-navigation.ts:91` and `:98`. `isPathActive` (this file, 617-621) matches exact pathname or `href + "/"` prefix, and the real routes are /users and /admin-mcp (no `trailingSlash` in next.config.mjs). On a deep link or refresh to /users or /admin-mcp, this guard evaluates collapsed=true — the active page is hidden inside the collapsed Resources group, contradicting the guarantee comment above ("the current page is never hidden inside a collapsed group on first load"). In the production static export the nav items' own URLs (`/users/`, `/admin-mcp/`) additionally 404. Suggest dropping the trailing slashes from both hrefs.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -521,7 +532,9 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
       }, 100)
       return () => clearTimeout(timer)
     }
-  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded])
+    // groupCollapseOverrides isn't read above — it's a re-measure trigger: toggling a group
+    // changes contentScrollRef's scrollHeight, so this effect must re-run to re-check the fill.
+  }, [tasks, hasMore, isLoadingMore, isLoadingTasks, page, loadTasks, isHistoryExpanded, groupCollapseOverrides])
```
rogercloud: Deps include `groupCollapseOverrides` as a re-measure trigger, but not `pathname`. When no override exists, collapse derives from the active item (line 745), so navigating away from a Resources route collapses the group and shrinks the scroll container without re-running the fill check (it self-heals only after ~30s via the poll effect re-setting `tasks`). Adding `pathname` to the deps would make route-driven collapse changes re-measure immediately — trivial and consistent with the re-measure-trigger comment above. Low impact either way.

frontend/src/components/layout/sidebar.tsx
```diff
@@ -723,88 +736,111 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
...
+          {navigationGroups.map((group, groupIndex) => {
+            // An active route overrides only the *default* collapsed state, so the current
+            // page is never hidden inside a collapsed group on first load — but an explicit
+            // user toggle always wins over that, so the group can still be collapsed manually.
+            const isGroupCollapsed = group.title in groupCollapseOverrides
+              ? groupCollapseOverrides[group.title]
+              : !!group.defaultCollapsed && !group.items.some(isItemActive)
+            return (
+              <div key={group.title} className={cn("mb-4", groupIndex === 0 && "mt-0")}>
+                <div
```
rogercloud: This div re-implements button semantics: `role="button"`, `tabIndex={0}`, the Enter/Space `onKeyDown` with `preventDefault`, and the focus-visible ring classes — all of which a native `<button type="button">` provides for free. Children are a span + svg chevron (no nested interactive elements), and the `flex` class keeps width behavior identical to the current div (repo uses native buttons for the other collapse toggles, e.g. the History section). Replace with `<button type="button">`, keep `aria-expanded` and `onClick`, drop `role`/`tabIndex`/`onKeyDown` (~6 lines).

frontend/src/components/layout/sidebar.tsx
```diff
@@ -241,6 +241,17 @@ export function Sidebar({ className, allowCollapse = true }: SidebarProps) {
 
   const [isExpanded, setIsExpanded] = useState(false)
   const [expandedMenus, setExpandedMenus] = useState<string[]>(["/agent"]) // Use href as a stable key
+  const [groupCollapseOverrides, setGroupCollapseOverrides] = useState<Record<string, boolean>>({})
+  // Flip from the collapsed state currently rendered (which may come from the default,
+  // an active-route auto-expand, or a previous toggle) so a click always changes what
+  // the user sees — recomputing from defaultCollapsed here would make the first click
+  // a no-op whenever the group was auto-expanded by the active route.
+  const toggleGroupCollapse = useCallback((title: string, currentCollapsed: boolean) => {
```
rogercloud: `useCallback` on `toggleGroupCollapse` adds nothing: the function is only used in `onClick`/`onKeyDown` handlers, never referenced in an effect or memo dep array (the auto-load effect depends on `groupCollapseOverrides`, not this callback), and the sibling `toggleMenu` is already a plain function. A plain function would behave identically and save the wrapper.

frontend/src/lib/sidebar-navigation.ts
rogercloud, line 43: While adding this field, consider giving `NavigationGroup` a stable `id`: `title` is now load-bearing in two places — the React key (sidebar.tsx:747) and the collapse-override key (sidebar.tsx:743) — and since extra-nav.ts appends overlay groups and this PR explicitly designs `defaultCollapsed` for overlay injection, a downstream group titled like a built-in (e.g. "Resources") would silently share both the React key and the collapse state. Non-blocking and forward-looking; an `id` field keyed on everywhere would close it off cheaply.

rogercloud (approved):
## Follow-up review — PR #1148: Collapsible sidebar navigation groups

### Summary
This PR makes the sidebar nav groups ("Agent Development", "Resources") individually collapsible, matching the existing History-section chevron UX and addressing issue #1144 (crowded left nav pushing chat history out of view). It adds an optional `defaultCollapsed?: boolean` to `NavigationGroup` (set true on "Resources"), stores only explicit user toggles in `groupCollapseOverrides: Record<string, boolean>` keyed by group title, derives the effective collapsed state at render (`override ?? (defaultCollapsed && !any item active)`), and adds keyboard/ARIA support to the clickable header. Changed files: `frontend/src/components/layout/sidebar.tsx`, `frontend/src/lib/sidebar-navigation.ts`.

### Update since last review (head 6fe90fec → 584d50e4)
The single new commit, "fix: keep active route visible without locking collapse, add focus ring", re-indents the nav-groups JSX block and changes the collapsed-state derivation from `!!group.defaultCollapsed` to `!!group.defaultCollapsed && !group.items.some(isItemActive)` — the active route now overrides only the *default* collapse, while an explicit user toggle still wins. It also adds `hover:bg-accent` plus a visible focus ring (`focus:outline-none focus-visible:ring-2 focus-visible:ring-ring`) to the group header, and adds explanatory comments. This directly addresses the two majors from the previous review (P4, P5).

### Prior-findings checklist

**gemini-code-assist (review of commit 1)**
- P1 — **FIXED**: collapse state is derived at render (`groupCollapseOverrides` holds only explicit toggles, initial `{}`; derivation at sidebar.tsx:743-745); the intermediate snapshot approach (e0e69515) was removed in 6fe90fec, so dynamically injected groups respect `defaultCollapsed`.
- P2 — **FIXED**: `groupCollapseOverrides` is in the auto-load effect's dependency array (sidebar.tsx:537).
- P3 — **FIXED**: header has `role="button"`, `tabIndex={0}`, `aria-expanded`, and Enter/Space `onKeyDown` with `preventDefault` (sidebar.tsx:748-759); `aria-controls` still absent but optional per APG.

**rogercloud (review of commit 2)**
- P4 [major] — **FIXED**: active route overrides the default collapse (sidebar.tsx:745) for the five named Resources routes (/kb, /models, /memory, /skill-hub, /api-keys — all exact-match); the residual gap for the trailing-slash routes /users/ and /admin-mcp/ is tracked as new finding A1.
- P5 [major] — **FIXED**: visible focus ring added (sidebar.tsx:752; `--ring` defined in tailwind.config.js:20 and globals.css:27).
- P6 — **NOT FIXED**: no test coverage added; duplicates new finding A4.
- P7 — **FIXED**: re-measure-trigger comment added above the effect deps (sidebar.tsx:535-536).
- P8 — **FIXED**: hover affordance added (`hover:bg-accent` + `transition-colors` + `rounded`, sidebar.tsx:752).
- P9 — **NOT FIXED**: collapsed groups still reserve ~22px (unconditional `mb-4` + `pb-1.5`); cosmetic, non-blocking.
- P10 — **FIXED**: nav-groups JSX re-indented in 584d50e4.
- P11 — **NOT FIXED**: the title-uniqueness constraint on `NavigationGroup` remains undocumented (no comment on the field; extra-nav.ts has none); optional, pre-existing gap extended by this PR.
- P12 [design reservation 1] — **FIXED as designed**: collapse is routing-aware exactly as recommended — the active route overrides only the default, an explicit toggle always wins (sidebar.tsx:743-745); waiver-eligible.
- P13 [design reservation 2] — **NOT FIXED**: `group.title` is still the load-bearing identity; duplicates new finding A3.

### New findings

**A1 [minor] — trailing-slash hrefs defeat the active-route auto-expand for two admin routes** (`frontend/src/lib/sidebar-navigation.ts:91`, `:98`)
`isPathActive` (sidebar.tsx:617-621) matches `pathname === href || pathname.startsWith(href + "/")`, but "User Management" (`href="/users/"`, :91) and the admin-MCP item (`href="/admin-mcp/"`, :98) carry trailing slashes while the real routes are /users and /admin-mcp (`next.config.mjs` has no `trailingSlash`). On a deep link or refresh to /users or /admin-mcp, `!!group.defaultCollapsed && !group.items.some(isItemActive)` evaluates the Resources group collapsed → the active page is hidden inside the collapsed group, contradicting the guarantee comment at sidebar.tsx:740-742. In the production static export the nav item's own URL (/users/) additionally 404s (no trailing-slash redirect). Suggestion: drop the trailing slashes from both hrefs.

**A2 [minor, low impact] — route-derived collapse changes don't trigger the auto-load re-measure** (`frontend/src/components/layout/sidebar.tsx:537`)
The auto-load effect's deps include `groupCollapseOverrides` as a re-measure trigger but not `pathname`; when no override exists the collapse derives from the active item (sidebar.tsx:745), so navigating away from a Resources route collapses the group and shrinks the container without re-running the fill check. The sidebar persists across route changes (root layout, next/link navigation); the state self-heals within ~30s via the poll effect. Suggestion: add `pathname` to the deps array — trivial and consistent with the author's own re-measure-trigger comment.

**A3 [minor, forward-looking] — `group.title` is both React key and override key; no stable id** (dup of P13, still open; `frontend/src/lib/sidebar-navigation.ts:38-44`)
`title` serves as the React key (sidebar.tsx:747) and the override key (sidebar.tsx:743), and `NavigationGroup` has no `id` field. extra-nav.ts appends groups and the PR body explicitly designs `defaultCollapsed` for overlay injection, so a downstream group titled like a built-in would share both the React key and the collapse state. Suggestion: add an `id` field to `NavigationGroup` and key on it.

**A4 [minor] — no test coverage for the new feature** (dup of P6, still open)
sidebar.test.tsx mocks `getNavigationGroupsForUser` to `[]` and only exercises logout; sidebar-navigation.test.ts doesn't assert `defaultCollapsed`; the PR adds no tests. All new observable contracts — default-collapsed, active-route auto-expand, toggle precedence, keyboard/ARIA, re-measure triggers — are unprotected. Suggestion: add a test rendering a `defaultCollapsed` group that asserts it is collapsed by default, auto-expands when a child route is active, and stays collapsed after an explicit toggle.

```
L750: native group header re-implements button semantics (role="button", tabIndex, Enter/Space onKeyDown, focus-visible ring) that a native <button type="button"> provides for free. Replace the div with <button> and drop role/tabIndex/onKeyDown (aria-expanded stays).
L249: shrink useCallback on toggleGroupCollapse adds nothing — it is only used in onClick/onKeyDown handlers, never in an effect/memo dep array. Use a plain function like the sibling toggleMenu.
net: -8 lines possible
```
(Update-only lens on 6fe90fec..584d50e4: "Lean already.")

### Testing (verified in review worktree at head 584d50e4)
- `tsc --noEmit` (npm run type-check): clean, 0 errors.
- eslint on changed files: 0 errors, 4 pre-existing `@next/next/no-img-element` warnings on lines untouched by the diff.
- vitest: 5/5 passed (4 tests in `src/components/layout` + 1 in `src/lib/sidebar-navigation.test.ts`).

### Conclusion
No critical or major issues remain: the two major findings (P4 active-route visibility, P5 focus indicator) are FIXED, and everything outstanding — A1-A4, P6/P9/P11/P13 — is minor or a confirmed simplification opportunity.

PR #1153, opened 2026-08-05
fix(tools): resync cached WebToolConfig with the resolved per-turn execution scope

AlexLiu190625:
## Summary

`WebToolConfig` is cached with `AgentService` per task, keyed on the namespace `scope_fingerprint`. The fingerprint deliberately covers only namespace-affecting fields, so an embedder resolver that returns a scope carrying turn-varying data outside those fields gets a cache hit — and the cached config keeps serving the **first turn's scope object** to every later turn: `_execution_scope` is assigned once in the constructor with no setter, and `set_connector_runtime_turn_id` resets five per-turn fields but not the scope. Consumers of the stale object include the OAuth resolver hook (`TokenRequest.scope`), the resolver-owned MCP connection factory, and nested agent tool construction.

## Change

- `WebToolConfig.set_execution_scope(scope)`: swaps the scope and drops the scope-derived caches (`_cached_mcp_configs`, `_factory_runtime_snapshot`, `_pending_runtime_policy`), mirroring `set_connector_runtime_turn_id`. Returns `False` when the scope is unchanged. The equality check carries an explicit `type(...) is type(...)` guard: scope subclasses may declare their extra fields `compare=False`, so a base scope and a subclass with the same namespace compare equal under dataclass `__eq__` while being materially different objects.
- `AgentServiceManager._sync_execution_scope(task_id, scope)`: called at both cached-agent return points in `get_agent_for_task`, passing the freshly resolved scope; on an actual swap it calls `agent.invalidate_tools()`. Unlike the connector-runtime-turn sync it is not gated on a turn id — callers that pass no turn id still get the resync. On a freshly built agent the setter is a no-op (the constructor already received the same object).

Namespace-affecting changes are unaffected: those still change the fingerprint and evict the whole `AgentService` before this sync runs.

## Testing

- `test_set_execution_scope_swaps_the_scope_and_drops_scope_derived_caches`: no-op on same scope, swap drops the caches, and the `compare=False` subclass trap (equal by value, different type → still swapped).
- `test_cached_agent_service_is_resynced_with_the_turn_execution_scope`: a cached agent under an unchanged fingerprint gets the new scope object and one `invalidate_tools()` call, with `connector_runtime_turn_id=None`.
- Full `tests/web/api/` + `tests/web/tools/`: 298 passed. `ruff`, `isort`, `mypy --package xagent`, `codespell` clean.

### DIFF
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -2746,8 +2746,8 @@ class AgentServiceManager:
         agent = self._agents.get(task_id)
         if agent is None:
             return
-        tool_config = getattr(agent, "tool_config", None)
-        if tool_config is None or not hasattr(tool_config, "set_execution_scope"):
+        tool_config = agent.tool_config
+        if tool_config is None:
             return
         if tool_config.set_execution_scope(scope):
             logger.info(
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -2227,6 +2227,7 @@ class AgentServiceManager:
                             self._sync_connector_runtime_turn(
                                 task_id, connector_runtime_turn_id
                             )
+                            self._sync_execution_scope(task_id, scope)
                             return self._agents[task_id]
                     except (
                         HTTPException,
@@ -2693,6 +2694,7 @@ class AgentServiceManager:
         self._agent_owner_ids[task_id] = runtime_user_id
         self._agent_scope_fingerprints[task_id] = fingerprint
         self._sync_connector_runtime_turn(task_id, connector_runtime_turn_id)
+        self._sync_execution_scope(task_id, scope)
         return self._agents[task_id]
 
     def _sync_connector_runtime_turn(
@@ -2737,6 +2739,21 @@ class AgentServiceManager:
                 task_id,
                 connector_runtime_turn_id,
             )
+
+    def _sync_execution_scope(
+        self, task_id: int, scope: Optional[ExecutionScope]
+    ) -> None:
+        agent = self._agents.get(task_id)
+        if agent is None:
+            return
+        tool_config = getattr(agent, "tool_config", None)
+        if tool_config is None or not hasattr(tool_config, "set_execution_scope"):
+            return
+        if tool_config.set_execution_scope(scope):
+            logger.info(
+                "Refreshing tools for task %s: execution scope advanced", task_id
+            )
+            agent.invalidate_tools()
 
     def remove_agent(self, task_id: int, user_id: Optional[int] = None) -> None:
         """Remove AgentService instance for completed task"""
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -2746,8 +2746,12 @@ class AgentServiceManager:
         agent = self._agents.get(task_id)
         if agent is None:
             return
-        tool_config = agent.tool_config
-        if tool_config is None:
+        # getattr/hasattr, not direct access: pause/resume off-turn builds
+        # store agents whose config is a DefaultToolConfig (no
+        # set_execution_scope), and the cached-return path this runs on must
+        # not blow up into the reconstruct fallback for them.
+        tool_config = getattr(agent, "tool_config", None)
+        if tool_config is None or not hasattr(tool_config, "set_execution_scope"):
             return
         if tool_config.set_execution_scope(scope):
             logger.info(
--- a/src/xagent/web/tools/config.py
+++ b/src/xagent/web/tools/config.py
@@ -1583,9 +1583,25 @@ class WebToolConfig(BaseToolConfig):
         serving the first turn's object to the OAuth resolver hook.
         Namespace-affecting changes never reach here -- the caller evicts on a
         fingerprint change before this runs.
+
+        No-op only for the same object, or for two plain base
+        ``ExecutionScope`` instances comparing equal: the base class's
+        compared fields fully describe it, and the persisted-snapshot path
+        decodes a fresh equal instance every turn that must not force a tool
+        rebuild. A subclass instance always swaps (unless identical) --
+        subclasses may carry per-turn payload declared ``compare=False``, so
+        value equality cannot prove freshness for them, and two same-type
+        instances differing only in that payload compare equal.
         """
-        if self._execution_scope == scope and type(self._execution_scope) is type(
-            scope
+        from xagent.core.execution_scope import ExecutionScope
+
+        previous = self._execution_scope
+        if previous is scope:
+            return False
+        if (
+            type(previous) is ExecutionScope
+            and type(scope) is ExecutionScope
+            and previous == scope
         ):
             return False
         self._execution_scope = scope
--- a/src/xagent/web/tools/config.py
+++ b/src/xagent/web/tools/config.py
@@ -1573,6 +1573,27 @@ class WebToolConfig(BaseToolConfig):
         self._pending_runtime_policy = None
         return True
 
+    def set_execution_scope(self, scope: Optional[Any]) -> bool:
+        """Switch the per-turn execution scope for a reused tool config.
+
+        ``WebToolConfig`` is cached with ``AgentService`` by task. The scope is
+        per-turn state exactly like the connector runtime turn id: an embedder
+        resolver may return a scope carrying turn-varying data that the base
+        namespace fingerprint does not cover, and a cached config must not keep
+        serving the first turn's object to the OAuth resolver hook.
+        Namespace-affecting changes never reach here -- the caller evicts on a
+        fingerprint change before this runs.
+        """
+        if self._execution_scope == scope and type(self._execution_scope) is type(
+            scope
+        ):
+            return False
+        self._execution_scope = scope
+        self._cached_mcp_configs = None
+        self._factory_runtime_snapshot = None
+        self._pending_runtime_policy = None
+        return True
+
     def _parse_numeric_task_id(self) -> Optional[int]:
         task_id = self._task_id
         if not isinstance(task_id, str) or not task_id:
--- a/tests/web/tools/test_web_tool_config_session_factory.py
+++ b/tests/web/tools/test_web_tool_config_session_factory.py
@@ -2678,23 +2678,48 @@ class _ScopeWithTurnPayload(ExecutionScope):
         return hash((self.sandbox_key_suffix, self.workspace_segments))
 
 
+def _prime_scope_derived_caches(cfg: WebToolConfig) -> None:
+    cfg._cached_mcp_configs = [{"id": 1, "connector_runtime": {"context": {}}}]
+    cfg._factory_runtime_snapshot = object()
+    cfg._pending_runtime_policy = object()
+
+
+def _assert_scope_derived_caches_primed(cfg: WebToolConfig) -> None:
+    assert cfg._cached_mcp_configs is not None
+    assert cfg._factory_runtime_snapshot is not None
+    assert cfg._pending_runtime_policy is not None
+
+
+def _assert_scope_derived_caches_dropped(cfg: WebToolConfig) -> None:
+    assert cfg._cached_mcp_configs is None
+    assert cfg._factory_runtime_snapshot is None
+    assert cfg._pending_runtime_policy is None
+
+
 def test_set_execution_scope_swaps_the_scope_and_drops_scope_derived_caches():
     scope_a = ExecutionScope(sandbox_key_suffix="tenant-a")
     cfg = WebToolConfig(db=None, request=None, execution_scope=scope_a)
-    cfg._cached_mcp_configs = [{"id": 1, "connector_runtime": {"context": {}}}]
+    _prime_scope_derived_caches(cfg)
 
-    # Same scope object: no-op, scope-derived cache untouched.
+    # Same scope object: no-op, every scope-derived cache untouched.
     assert cfg.set_execution_scope(scope_a) is False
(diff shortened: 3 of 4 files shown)


src/xagent/web/tools/config.py
```diff
@@ -1573,6 +1573,27 @@ def set_connector_runtime_turn_id(self, turn_id: Optional[str]) -> bool:
...
+        """
+        if self._execution_scope == scope and type(self._execution_scope) is type(
+            scope
+        ):
+            return False
+        self._execution_scope = scope
+        self._cached_mcp_configs = None
+        self._factory_runtime_snapshot = None
+        self._pending_runtime_policy = None
+        return True
```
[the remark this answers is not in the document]
AlexLiu190625: These caches cannot go stale across this transition, so the clears would be dead writes. The two hooks that populate them: the tool-overrides hook derives from a route ContextVar plus the user's team membership and never reads the execution scope at all; the allowlist hook does read the scope, but only `scope.memory_dimensions` — and `memory_dimensions` is part of `scope_fingerprint`. Any scope change that could alter either cache's inputs therefore changes the fingerprint, and the manager evicts the whole AgentService before `set_execution_scope` can run — which is the contract stated in this method's docstring ("Namespace-affecting changes never reach here"). Within the reachable domain of this method (same fingerprint, hence same memory_dimensions) the cached values remain correct by construction. Adding the clears would imply a dependency that does not exist on this path, so keeping the reset set aligned with the genuinely scope-derived caches.

tests/web/tools/test_web_tool_config_session_factory.py
```diff
@@ -2647,6 +2648,75 @@ def test_connector_runtime_turn_switch_invalidates_runtime_caches():
...
+    assert cfg.get_execution_scope() is scope_a
+    assert cfg._cached_mcp_configs is not None
+
+    # Different scope: swaps and drops every scope-derived cache.
+    scope_b = ExecutionScope(sandbox_key_suffix="tenant-b")
+    assert cfg.set_execution_scope(scope_b) is True
+    assert cfg.get_execution_scope() is scope_b
+    assert cfg._cached_mcp_configs is None
+    assert cfg._factory_runtime_snapshot is None
+    assert cfg._pending_runtime_policy is None
```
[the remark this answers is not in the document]
AlexLiu190625: Follows from the sibling thread: the three attributes are not scope-derived within this method's reachable domain (their inputs are covered by the fingerprint or independent of the scope entirely), so the test intentionally pins the reset set to the caches that genuinely derive from the scope object.

rogercloud (approved):
## Summary

This PR fixes a stale-state bug in per-turn tool configuration: `WebToolConfig` is cached per task keyed on a namespace scope fingerprint, so a turn-varying execution scope (e.g. `strict_memory_isolation`, or embedder-subclass payload fields the fingerprint does not cover) can hit the cache and keep serving the first turn's scope object to the OAuth resolver hook (`TokenRequest.scope`), the resolver-owned MCP connection factory, and nested agent tool construction. The PR adds `WebToolConfig.set_execution_scope(scope)`, which swaps the scope and drops the scope-derived caches (`_cached_mcp_configs`, `_factory_runtime_snapshot`, `_pending_runtime_policy`), returning `False` when the scope is unchanged (value equality plus a `type(...) is type(...)` guard); `AgentServiceManager._sync_execution_scope` calls it at both cached-agent return points in `get_agent_for_task` and invalidates tools on an actual swap. Design verdict: acceptable-with-reservations — a root-cause fix consistent with the existing connector-runtime-turn pattern; the reservation is the freshness contract for subclass scopes (see F1/D1 below).

## Update summary

No new commits since the previous review: the head is still `0d2f00be`, and the suite still passes (2016 tests, exit 0). The author added two inline rebuttal comments on the same commit, both verified correct against the code. The two prior findings are therefore recorded as NOT FIXED with disposition WAIVED; this pass adds two minor findings (F1, F2), one design-level note (D1), and one simplification (S1).

## Prior-findings checklist

- **P1 — clear `_cached_tool_overrides` / `_cached_tool_allowlist` / `_tool_allowlist_cached` on scope change: NOT FIXED, WAIVED.** Author rebuttal verified: `get_user_tool_overrides` / `get_user_tool_allowlist` hooks read only `(db, user)` (config.py:1869-1885, 2138-2140; hook signatures tool_credentials.py:32-35, 55-58), never the scope; the one plausible scope input (`memory_dimensions`) is covered by the fingerprint, and a fingerprint change evicts the whole `AgentService` before the sync runs. The clears would be dead writes.
- **P2 — pre-populate and assert the three policy caches in the test: NOT FIXED, WAIVED.** Follows from P1; the test correctly pins the reset set to genuinely scope-derived caches.

## New findings

### Minor

- **F1 (config.py:1587-1590)** — The equality + type-guard no-op check still misses two instances of the *same* subclass that differ only in a `compare=False` payload field: they compare equal and share a type, so the exact stale-scope problem this PR fixes can still reproduce through the documented embedder resolver extension point (`ExecutionScopeResolver`, execution_scope.py:857; isinstance admission at :1722-1734; subclass objects round-tripped unchanged at :1739/1748/1752). No `ExecutionScope` subclass exists in `src/` today (only the test's `_ScopeWithTurnPayload`), so the failure is latent — but it sits in the PR's own stated purpose: the setter docstring (config.py:1580-1583) and the test comment both assert the richer-scope contract, and the test covers only the base→subclass cross-type swap, never same-subclass successive turns. Suggested fix: no-op only when both sides are exact base-class instances (`type(x) is ExecutionScope and type(scope) is ExecutionScope and x == scope`), otherwise identity (`is`) then swap; add a same-subclass successive-turn regression test. (Identity-only would regress the persisted-snapshot path, where per-turn `from_dict` decode creates fresh equal instances and forces per-turn tool rebuilds — e.g. websocket.py:1203/2410/3142/5845.) Consumers affected when stale: OAuth `TokenRequest.scope` (config.py:2542) and the MCP connection factory (config.py:3209).
- **F2 (test_web_tool_config_session_factory.py:2684)** — The test claims to verify "drops every scope-derived cache" but only pre-populates `_cached_mcp_configs`; `_factory_runtime_snapshot` and `_pending_runtime_policy` are `None` from construction (config.py:1268, 1271) and never set here, so the `is None` asserts (:2696-2697) are vacuous — deleting the setter's cleanup lines for either attribute would still pass. Fix: pre-populate both with sentinel values before the swap, and assert they survive the same-object no-op branch.

### Design-level

- **D1 (latent)** — Scope freshness is modeled as "value-equal AND same concrete type ⇒ no sync". Correct for the base class today (all `compare=True` fields are fingerprint-covered or equality-caught; `strict_memory_isolation` is a stored `compare=True` field), but the contract is under-specified for the documented subclass extension point: the guard cannot detect same-type subclass payload changes (overlaps F1). No single line to blame beyond the setter's no-op condition.

- `L2749: delete duck-typing guard in AgentServiceManager._sync_execution_scope. Replace getattr(agent, "tool_config", None) / hasattr(tool_config, "set_execution_scope") with direct attribute access + is-None check, mirroring the sibling _sync_connector_runtime_turn (chat.py:2700-2740); every manager-stored agent has a WebToolConfig (all manager construction paths build one; AgentService.__init__ sets tool_config unconditionally, core/agent/service.py:92; the only non-WebToolConfig agent, websocket.py:8330 builder_chat_agent, is a local never stored in the manager).`

`net: -2 lines, one dead defensive branch removed; behavior unchanged.`

## Test results

`pytest tests/web/api/ tests/web/tools/` in the review worktree (Python 3.12): 2016 tests collected, exit code 0. Both new test files pass.

src/xagent/web/tools/config.py
```diff
@@ -1573,6 +1573,27 @@ def set_connector_runtime_turn_id(self, turn_id: Optional[str]) -> bool:
...
+
+        ``WebToolConfig`` is cached with ``AgentService`` by task. The scope is
+        per-turn state exactly like the connector runtime turn id: an embedder
+        resolver may return a scope carrying turn-varying data that the base
+        namespace fingerprint does not cover, and a cached config must not keep
+        serving the first turn's object to the OAuth resolver hook.
+        Namespace-affecting changes never reach here -- the caller evicts on a
+        fingerprint change before this runs.
+        """
+        if self._execution_scope == scope and type(self._execution_scope) is type(
```
rogercloud: The no-op guard `self._execution_scope == scope and type(self._execution_scope) is type(scope)` still misses two instances of the *same* subclass that differ only in a `compare=False` payload field: they compare equal and share a type, so the stale-scope problem this PR fixes can still reproduce through the documented embedder resolver extension point (`ExecutionScopeResolver`, execution_scope.py:857; isinstance admission at :1722-1734; subclass objects round-tripped unchanged at :1739/1748/1752). No `ExecutionScope` subclass exists in `src/` today (only the test's `_ScopeWithTurnPayload`), so the failure is latent — but it sits in the PR's stated purpose: the setter docstring and the test comment both assert the richer-scope contract, and the test covers only the base→subclass cross-type swap, never same-subclass successive turns. Suggested fix: no-op only when both sides are exact base-class instances (`type(x) is ExecutionScope and type(scope) is ExecutionScope and x == scope`), otherwise identity (`is`) then swap; add a same-subclass successive-turn regression test. (Identity-only would regress the persisted-snapshot path, where per-turn `from_dict` decode creates fresh equal instances and forces per-turn tool rebuilds — e.g. websocket.py:1203/2410/3142/5845.) Consumers affected when stale: OAuth `TokenRequest.scope` (config.py:2542) and the MCP connection factory (config.py:3209).

tests/web/tools/test_web_tool_config_session_factory.py
```diff
@@ -2647,6 +2648,75 @@ def test_connector_runtime_turn_switch_invalidates_runtime_caches():
...
+        ] and dict(self.memory_dimensions) == dict(other.memory_dimensions)
+
+    def __hash__(self) -> int:
+        return hash((self.sandbox_key_suffix, self.workspace_segments))
+
+
+def test_set_execution_scope_swaps_the_scope_and_drops_scope_derived_caches():
+    scope_a = ExecutionScope(sandbox_key_suffix="tenant-a")
+    cfg = WebToolConfig(db=None, request=None, execution_scope=scope_a)
+    cfg._cached_mcp_configs = [{"id": 1, "connector_runtime": {"context": {}}}]
```
rogercloud: The test claims to verify "drops every scope-derived cache" but only pre-populates `_cached_mcp_configs` (line 2684); `_factory_runtime_snapshot` and `_pending_runtime_policy` are `None` from construction (config.py:1268, 1271) and never set in this test, so the `is None` asserts at :2696-2697 are vacuous — deleting the setter's cleanup lines for either attribute would still pass. Pre-populate both with sentinel values before the swap, and assert they survive the same-object no-op branch.

src/xagent/web/api/chat.py
```diff
@@ -2738,6 +2740,21 @@ def _sync_connector_runtime_turn(
                 connector_runtime_turn_id,
             )
 
+    def _sync_execution_scope(
+        self, task_id: int, scope: Optional[ExecutionScope]
+    ) -> None:
+        agent = self._agents.get(task_id)
+        if agent is None:
+            return
+        tool_config = getattr(agent, "tool_config", None)
```
rogercloud: The `getattr(agent, "tool_config", None)` / `hasattr(tool_config, "set_execution_scope")` duck-typing guard never fires: both call sites (chat.py:2230, 2697) hold an agent that always has a `WebToolConfig` (all manager construction paths build one; `AgentService.__init__` sets `tool_config` unconditionally, core/agent/service.py:92; the only non-WebToolConfig agent, `builder_chat_agent` at websocket.py:8330, is a local never stored in the manager). Replace with direct attribute access + is-None check, mirroring the sibling `_sync_connector_runtime_turn` (chat.py:2700-2740).

src/xagent/web/tools/config.py
AlexLiu190625: Fixed in 321b5fa6, taking the suggested shape: no-op only for the identical object or two plain base-class instances comparing equal; any subclass instance swaps. The base-class-equal carve-out preserves the persisted-snapshot path (fresh equal base instances per turn must not force rebuilds), exactly as noted. New regression test `test_set_execution_scope_swaps_equal_same_subclass_instances` drives two same-subclass instances equal under `__eq__` differing only in excluded payload — reverting to the previous guard turns it red (verified).

tests/web/tools/test_web_tool_config_session_factory.py
AlexLiu190625: Fixed in 321b5fa6: the test now primes `_factory_runtime_snapshot` and `_pending_runtime_policy` with sentinels alongside `_cached_mcp_configs`, asserts all three survive both no-op branches (same object, fresh equal base instance), and asserts all three drop on a swap — the cleanup lines are no longer deletable without a failure.

src/xagent/web/api/chat.py
AlexLiu190625: Fixed in 321b5fa6: direct `agent.tool_config` access with an `is None` check, mirroring `_sync_connector_runtime_turn`.

AlexLiu190625: Correction after CI: the guard is load-bearing after all — reverted the simplification in ba999021. Pause/resume off-turn builds store manager agents backed by a `DefaultToolConfig`, which has no `set_execution_scope` (CI: `test_pause_cache_miss_builds_under_resolver_namespace_not_snapshot` failed with exactly that AttributeError), and on the reconstruct early-return the crash was swallowed into the fallback path, returning a rebuilt agent instead of the cached one. The guard stays, now with a comment recording why.

PR #1154, opened 2026-08-05
fix(mcp-oauth): reject compressed OAuth responses when size-capping

AlexLiu190625:
## Summary

`oauth_post`'s `max_response_bytes` guard measures decoded bytes: `aiter_bytes()` yields httpx's auto-decompressed output, so a compressed reply inflates before the check ever runs. Reproduced: a 1,052-byte gzip wire body arrived as a single 1 MiB decoded chunk — roughly 1000x memory/CPU amplification available to a hostile peer per request, regardless of any `Accept-Encoding` the caller sends.

## Change

When `max_response_bytes` is set, a response carrying a non-identity `Content-Encoding` is refused at stream open — before any body bytes are consumed, so decompression never happens — using the same `response_too_large` error shape as the existing size check, keeping callers' classification unchanged. Callers that want compression simply don't pass the cap; unbounded calls are untouched. The streaming loop also now checks before extending, so the peak held is never above the cap plus nothing rather than cap plus one chunk.

## Testing

- New: a gzip response whose *decoded* size is well under a generous cap still gets refused (defeats any decoded-size-only guard — the test can only pass if the encoding check itself fires); identity-encoded control still succeeds. Removing the encoding check turns the gzip test red.
- `tests/web/services/`: 795 passed. Pre-commit hooks clean.

### DIFF
--- a/src/xagent/web/services/mcp_oauth.py
+++ b/src/xagent/web/services/mcp_oauth.py
@@ -266,6 +266,15 @@ async def oauth_post(
                 **request_kwargs,
             )
         else:
+            # A bounded read must not invite compression: httpx advertises
+            # gzip/br/zstd by default, and the Content-Encoding refusal below
+            # would then reject compliant servers that simply honored it.
+            # Force identity on the request so compliant servers send plain
+            # bytes; the response-side check stays as the hard guard against
+            # peers that compress anyway.
+            headers = httpx.Headers(request_kwargs.pop("headers", None))
+            headers["accept-encoding"] = "identity"
+            request_kwargs["headers"] = headers
             async with client.stream(
                 "POST",
                 url,
--- a/src/xagent/web/services/mcp_oauth.py
+++ b/src/xagent/web/services/mcp_oauth.py
@@ -277,14 +277,26 @@ async def oauth_post(
                         "invalid_resource",
                         "OAuth token endpoint redirects are not supported",
                     )
+                # max_response_bytes bounds the decoded byte count, but
+                # aiter_bytes() yields httpx's auto-decompressed output, so a
+                # compressed body can inflate far past the cap before the
+                # check below ever sees it. Refuse encoded bodies outright
+                # instead of decompressing-then-measuring; callers that need
+                # compression must not pass max_response_bytes.
+                content_encoding = streamed_response.headers.get("content-encoding", "")
+                if content_encoding.strip().lower() not in {"", "identity"}:
+                    raise MCPOAuthDiscoveryError(
+                        "response_too_large",
+                        "OAuth endpoint response exceeded the allowed size",
+                    )
                 content = bytearray()
                 async for chunk in streamed_response.aiter_bytes():
-                    content.extend(chunk)
-                    if len(content) > max_response_bytes:
+                    if len(content) + len(chunk) > max_response_bytes:
                         raise MCPOAuthDiscoveryError(
                             "response_too_large",
                             "OAuth endpoint response exceeded the allowed size",
                         )
+                    content.extend(chunk)
                 response = httpx.Response(
                     streamed_response.status_code,
                     headers=[
--- a/src/xagent/web/services/mcp_oauth.py
+++ b/src/xagent/web/services/mcp_oauth.py
@@ -295,8 +295,8 @@ async def oauth_post(
                 content_encoding = streamed_response.headers.get("content-encoding", "")
                 if content_encoding.strip().lower() not in {"", "identity"}:
                     raise MCPOAuthDiscoveryError(
-                        "response_too_large",
-                        "OAuth endpoint response exceeded the allowed size",
+                        "unsupported_response_encoding",
+                        "OAuth endpoint response used an unsupported content encoding",
                     )
                 content = bytearray()
                 async for chunk in streamed_response.aiter_bytes():
@@ -392,6 +392,11 @@ async def register_mcp_oauth_public_client(
             raise MCPOAuthDiscoveryError(
                 "client_registration_failed",
                 "OAuth client registration response exceeded the allowed size",
+            ) from exc
+        if exc.code == "unsupported_response_encoding":
+            raise MCPOAuthDiscoveryError(
+                "client_registration_failed",
+                "OAuth client registration response used an unsupported content encoding",
             ) from exc
         raise
     except (httpx.HTTPError, ValueError) as exc:
--- a/tests/web/services/test_mcp_oauth_discovery.py
+++ b/tests/web/services/test_mcp_oauth_discovery.py
@@ -504,7 +504,45 @@ async def test_oauth_post_rejects_compressed_response_without_decoding():
                 json={"client_name": "Xagent"},
             )
 
-    assert exc.value.code == "response_too_large"
+    assert exc.value.code == "unsupported_response_encoding"
+
+
+@pytest.mark.asyncio
+async def test_oauth_post_refuses_encoded_body_before_reading_any_of_it():
+    # The ordering is the actual bomb defense: the refusal must fire from the
+    # headers alone, before a single body byte is pulled (and decompressed).
+    # A counting stream makes the guarantee mechanical -- if the encoding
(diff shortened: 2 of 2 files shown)


src/xagent/web/services/mcp_oauth.py
```diff
@@ -277,14 +277,26 @@ async def oauth_post(
...
                         "OAuth token endpoint redirects are not supported",
                     )
+                # max_response_bytes bounds the decoded byte count, but
+                # aiter_bytes() yields httpx's auto-decompressed output, so a
+                # compressed body can inflate far past the cap before the
+                # check below ever sees it. Refuse encoded bodies outright
+                # instead of decompressing-then-measuring; callers that need
+                # compression must not pass max_response_bytes.
+                content_encoding = streamed_response.headers.get("content-encoding", "")
+                if content_encoding.strip().lower() not in {"", "identity"}:
```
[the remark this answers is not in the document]
AlexLiu190625: Adopted in 7dca8ecc, with the enforcement moved inside oauth_post where the cap's contract lives: when max_response_bytes is set, the request now forces Accept-Encoding: identity (compliant servers send plain bytes and are never refused), and the response-side Content-Encoding check remains the hard guard for peers that compress anyway. A new test pins that bounded requests carry the identity header.

src/xagent/web/services/mcp_oauth.py
```diff
@@ -277,14 +286,26 @@ async def oauth_post(
...
+                    raise MCPOAuthDiscoveryError(
+                        "response_too_large",
```
rogercloud: Reusing `response_too_large` for the encoding refusal misleads diagnostics: this response can be far below the size cap — the new test at tests/web/services/test_mcp_oauth_discovery.py:479 uses a 1MiB decoded body under a 2MiB cap and still lands here. `register_mcp_oauth_public_client` (mcp_oauth.py:390-395) maps this code to "OAuth client registration response exceeded the allowed size", which would point operators at the wrong cause for a compressed-but-small response. Consider a distinct code (e.g. `unsupported_response_encoding`), or at minimum a message like "could not be safely size-limited".

tests/web/services/test_mcp_oauth_discovery.py
```diff
@@ -474,6 +475,88 @@ def handler(request: httpx.Request) -> httpx.Response:
...
+
+    def handler(request: httpx.Request) -> httpx.Response:
+        return httpx.Response(
+            200,
+            content=compressed,
+            headers={"Content-Encoding": "gzip", "Content-Type": "application/json"},
+        )
+
+    async with httpx.AsyncClient(transport=httpx.MockTransport(handler)) as client:
+        with pytest.raises(MCPOAuthDiscoveryError) as exc:
```
rogercloud, line 499: This test proves the rejection fires, but not that it fires before decompression: it still passes if the Content-Encoding check is moved after the `aiter_bytes()` loop, because the 1MiB gzip payload decodes to 1MiB — under the 2MiB cap. The ordering is what protects against decompression bombs (a streaming 128MiB gzip-bomb probe peaks at 0.0MiB with the current check vs 269.4MiB if moved). Consider a custom `httpx.AsyncByteStream` with an iteration counter to assert the body is never read, or explicitly accept that this guarantee rests on code review.

rogercloud (approved):
## Review — PR #1154: fix(mcp-oauth): reject compressed OAuth responses when size-capping

### Summary

`oauth_post`'s bounded-read path applied its size cap only after httpx had already auto-decompressed the response, so a small gzip/br/zstd body could inflate in memory far past `max_response_bytes` before being rejected — and because httpx advertises those encodings by default, compliant servers trigger exactly this path in normal operation. This PR fixes the bounded branch by (1) forcing `Accept-Encoding: identity` on the request so compliant servers send plain bytes, (2) refusing any non-identity `Content-Encoding` on the response before the body is read, and (3) moving the size check ahead of each buffer extension. Two commits touch two files: `src/xagent/web/services/mcp_oauth.py` (+25) and `tests/web/services/test_mcp_oauth_discovery.py` (+83).

### Update since last review

Commit `7dca8ecc` adds the request-side `Accept-Encoding: identity` header (mcp_oauth.py:275-277), addressing the prior review's concern that refusing encoded responses outright would break compliant servers. The header is set only on the `max_response_bytes`-bounded branch and is layered on top of caller-supplied headers; the response-side Content-Encoding guard remains in place as defense in depth. The earlier commits (size check before `content.extend`, response-side encoding refusal) are unchanged.

### Approach verdict

**Sound.** The fix addresses the root cause — do not invite compression when you cannot afford decompression — rather than patching the symptom, and keeps a hard response-side guard for peers that compress anyway. Caller headers are preserved, the bounded/unbounded split is coherent, and the error paths are covered by tests.

### Findings

**Minor — F1: test does not pin the reject-before-decompression ordering** (tests/web/services/test_mcp_oauth_discovery.py:479-507)

`test_oauth_post_rejects_compressed_response_without_decoding` asserts only the final exception and its code, not that the Content-Encoding check fires before any iteration of the body. The test still passes if the check is moved after the `aiter_bytes()` loop (the 1MiB gzip payload decodes to 1MiB, under the 2MiB cap), and a streaming 128MiB gzip-bomb probe shows the real difference: 0.0MiB peak memory with the current ordering vs 269.4MiB with the check moved. The compression-bomb regression would recur undetected. Suggestion: use a custom `httpx.AsyncByteStream` with an iteration counter to assert the body is never read, or explicitly accept that this ordering guarantee rests on code review.

**Minor — F2: check-before-extend memory invariant has no regression test** (src/xagent/web/services/mcp_oauth.py:303)

Reverting to extend-then-check still passes all 7 `oauth_post` tests and all 46 tests in the file; the externally observable contract (an oversized response raises `response_too_large`) is order-independent. This is a white-box coverage gap rather than a behavioral defect, and an implementation-coupled test may be reasonably declined — recorded as a coverage observation, not a blocker.

**Minor — F3: `response_too_large` is reused for the encoding refusal, which misleads diagnostics** (src/xagent/web/services/mcp_oauth.py:298)

The rejected response may be far below the size cap — the new test constructs a 1MiB decoded body under a 2MiB cap and still receives this code. The caller `register_mcp_oauth_public_client` (mcp_oauth.py:390-395) maps it to "OAuth client registration response exceeded the allowed size", pointing operators at the wrong cause when the actual problem is an unsupported encoding. Suggestion: use a distinct code (e.g. `unsupported_response_encoding`), or at minimum reword the message to something like "could not be safely size-limited".

### Prior findings

- **P1 — refusing compression breaks compliant servers; request `Accept-Encoding: identity` instead (gemini-code-assist, body-only): FIXED.** Commit `7dca8ecc` sets `Accept-Encoding: identity` on the bounded branch (mcp_oauth.py:275-277); the response-side guard is retained as defense in depth, and caller-supplied headers are preserved. No regression observed.
- **Second gemini-code-assist review: no comments.**

None — both the update-only and full-diff simplification lenses returned "Lean already".

### Tests

`.venv/bin/pytest tests/web/services/test_mcp_oauth_discovery.py -q` → **46 passed**, 13 warnings (PR head `7dca8ecc`).

src/xagent/web/services/mcp_oauth.py
AlexLiu190625: Fixed in 8da05828: the refusal now raises unsupported_response_encoding with its own message, and the registration wrapper maps it alongside its response_too_large sibling. (The SaaS consumer will classify the new code explicitly when it bumps its pin.)

tests/web/services/test_mcp_oauth_discovery.py
AlexLiu190625: Fixed in 8da05828 with the suggested mechanism: a counting httpx.AsyncByteStream pins that the refusal fires from headers alone — zero body reads — so moving the check after the loop turns the test red regardless of decoded size. The check-before-extend coverage observation is accepted as-is per your framing: the externally observable contract is order-independent, and an implementation-coupled assertion there would pin the how rather than the what.

PR #1160, opened 2026-08-06
feat: add Google Sheets connector

### LINKED ISSUE #1142: feat: Add Google Sheets connector
### Description

Add a new Google Sheets Connector to allow agents to access and interact with users' Google Sheets data.

Users should be able to authenticate their Google account and allow agents to retrieve and use spreadsheet information in workflows.

yiboyasss:
## Summary
- Add `google_sheets.py` MCP module: read/update/append/clear ranges, add/delete sheets, create spreadsheets, get spreadsheet metadata
- Register `google-sheets` app in `builtin_mcp_registry.py`, reusing the shared `google` OAuth provider (scopes: `spreadsheets`, `drive.file`)
- Seed the catalog entry via a new alembic migration, with a matching migration test

## Test plan
- [x] `ruff check` / `ruff format --check` pass on all new/changed files
- [x] `alembic heads` shows a single head after the new migration
- [x] New migration tests pass (insert / idempotent / downgrade / registry drift check)
- [x] `xagent.web.tools.mcp.google_sheets` imports cleanly
- [x] Enable Google Sheets API in the target GCP project and re-submit OAuth consent screen scopes (`spreadsheets`, `drive.file`) before this connector works end-to-end
- [x] Manually connect Google Sheets in the connector UI and exercise read/update/append against a real spreadsheet

### DIFF
--- a/src/xagent/web/tools/mcp/google_sheets.py
+++ b/src/xagent/web/tools/mcp/google_sheets.py
@@ -20,6 +20,9 @@ mcp = FastMCP("google-sheets-mcp")
 
 _SPREADSHEET_URL_ID_PATTERN = re.compile(r"/spreadsheets/d/([a-zA-Z0-9_-]+)")
 
+_sheets_service: Any = None
+_drive_service: Any = None
+
 
 def _get_credentials() -> Credentials:
     token = os.environ.get("GOOGLE_ACCESS_TOKEN")
@@ -45,11 +48,21 @@ def _get_credentials() -> Credentials:
 
 
 def get_sheets_service() -> Any:
-    return build("sheets", "v4", credentials=_get_credentials())
+    global _sheets_service
+    if _sheets_service is None:
+        _sheets_service = build(
+            "sheets", "v4", credentials=_get_credentials(), static_discovery=True
+        )
+    return _sheets_service
 
 
 def get_drive_service() -> Any:
-    return build("drive", "v3", credentials=_get_credentials())
+    global _drive_service
+    if _drive_service is None:
+        _drive_service = build(
+            "drive", "v3", credentials=_get_credentials(), static_discovery=True
+        )
+    return _drive_service
 
 
 def _resolve_spreadsheet_id(spreadsheet_id: str) -> str:
@@ -64,12 +77,12 @@ def google_sheets_get_spreadsheet(spreadsheet_id: str) -> str:
     and the list of sheets (tabs) with their sheet_id, title, and grid size.
     """
     try:
-        sheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
         service = get_sheets_service()
         spreadsheet = (
             service.spreadsheets()
             .get(
-                spreadsheetId=sheet_id,
+                spreadsheetId=resolved_spreadsheet_id,
                 fields="spreadsheetId,properties.title,spreadsheetUrl,sheets.properties",
             )
             .execute()
@@ -90,7 +103,9 @@ def google_sheets_get_spreadsheet(spreadsheet_id: str) -> str:
         return json.dumps(
             {
                 "status": "success",
-                "spreadsheet_id": spreadsheet.get("spreadsheetId", sheet_id),
+                "spreadsheet_id": spreadsheet.get(
+                    "spreadsheetId", resolved_spreadsheet_id
+                ),
                 "title": spreadsheet.get("properties", {}).get("title"),
                 "url": spreadsheet.get("spreadsheetUrl"),
                 "sheets": sheets,
@@ -135,13 +150,15 @@ def google_sheets_create_spreadsheet(
                 .get(fileId=spreadsheet["spreadsheetId"], fields="parents")
                 .execute()
             )
-            previous_parents = ",".join(existing_file.get("parents", []))
-            drive.files().update(
-                fileId=spreadsheet["spreadsheetId"],
-                addParents=parent_id,
-                removeParents=previous_parents,
-                fields="id,parents",
-            ).execute()
+            previous_parents = existing_file.get("parents") or []
+            update_kwargs: dict[str, Any] = {
+                "fileId": spreadsheet["spreadsheetId"],
+                "addParents": parent_id,
+                "fields": "id,parents",
+            }
+            if previous_parents:
+                update_kwargs["removeParents"] = ",".join(previous_parents)
+            drive.files().update(**update_kwargs).execute()
 
         return json.dumps(
             {
@@ -163,12 +180,12 @@ def google_sheets_read_range(spreadsheet_id: str, range_name: str) -> str:
     range_name="Sheet1!A1:D10". spreadsheet_id accepts a bare id or full URL.
     """
     try:
-        sheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
         service = get_sheets_service()
         result = (
             service.spreadsheets()
             .values()
-            .get(spreadsheetId=sheet_id, range=range_name)
+            .get(spreadsheetId=resolved_spreadsheet_id, range=range_name)
             .execute()
         )
         return json.dumps(
@@ -198,7 +215,7 @@ def google_sheets_update_range(
     "RAW" (stores every value as a literal string).
     """
     try:
-        sheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
         values = json.loads(values_json)
         if not isinstance(values, list) or not all(
             isinstance(row, list) for row in values
@@ -212,7 +229,7 @@ def google_sheets_update_range(
             service.spreadsheets()
             .values()
             .update(
-                spreadsheetId=sheet_id,
+                spreadsheetId=resolved_spreadsheet_id,
                 range=range_name,
                 valueInputOption=value_input_option,
                 body={"values": values},
@@ -244,7 +261,7 @@ def google_sheets_append_rows(
     rows, e.g. '[["a","b"],["c","d"]]'.
     """
     try:
-        sheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
         values = json.loads(values_json)
         if not isinstance(values, list) or not all(
             isinstance(row, list) for row in values
@@ -258,7 +275,7 @@ def google_sheets_append_rows(
             service.spreadsheets()
             .values()
             .append(
-                spreadsheetId=sheet_id,
+                spreadsheetId=resolved_spreadsheet_id,
                 range=range_name,
                 valueInputOption=value_input_option,
                 insertDataOption="INSERT_ROWS",
@@ -286,12 +303,12 @@ def google_sheets_clear_range(spreadsheet_id: str, range_name: str) -> str:
     removing cell formatting.
     """
     try:
-        sheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
         service = get_sheets_service()
         result = (
             service.spreadsheets()
             .values()
-            .clear(spreadsheetId=sheet_id, range=range_name)
+            .clear(spreadsheetId=resolved_spreadsheet_id, range=range_name)
             .execute()
         )
         return json.dumps(
@@ -313,12 +330,12 @@ def google_sheets_add_sheet(
     Add a new sheet (tab) to an existing spreadsheet.
     """
     try:
-        sheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
         service = get_sheets_service()
         result = (
             service.spreadsheets()
             .batchUpdate(
-                spreadsheetId=sheet_id,
+                spreadsheetId=resolved_spreadsheet_id,
                 body={
                     "requests": [
                         {
--- a/src/xagent/web/tools/mcp/google_sheets.py
+++ b/src/xagent/web/tools/mcp/google_sheets.py
@@ -10,7 +10,6 @@ from mcp.server.fastmcp import FastMCP
 
 from .utils import resolve_id_from_url, setup_proxy_env
 
-logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger("google-sheets-mcp")
 
 # Ensure standard proxy environment variables are set to prevent hanging requests
@@ -18,10 +17,7 @@ setup_proxy_env()
 
 mcp = FastMCP("google-sheets-mcp")
 
-_SPREADSHEET_URL_ID_PATTERN = re.compile(r"/spreadsheets/d/([a-zA-Z0-9_-]+)")
-
-_sheets_service: Any = None
-_drive_service: Any = None
+_SPREADSHEET_URL_ID_PATTERN = re.compile(r"/spreadsheets/(?:u/\d+/)?d/([a-zA-Z0-9_-]+)")
 
 
 def _get_credentials() -> Credentials:
@@ -48,21 +44,11 @@ def _get_credentials() -> Credentials:
 
 
 def get_sheets_service() -> Any:
-    global _sheets_service
-    if _sheets_service is None:
-        _sheets_service = build(
-            "sheets", "v4", credentials=_get_credentials(), static_discovery=True
-        )
-    return _sheets_service
+    return build("sheets", "v4", credentials=_get_credentials())
 
 
 def get_drive_service() -> Any:
-    global _drive_service
-    if _drive_service is None:
-        _drive_service = build(
-            "drive", "v3", credentials=_get_credentials(), static_discovery=True
-        )
-    return _drive_service
+    return build("drive", "v3", credentials=_get_credentials())
 
 
 def _resolve_spreadsheet_id(spreadsheet_id: str) -> str:
@@ -126,7 +112,10 @@ def google_sheets_create_spreadsheet(
     Create a new Google Sheets spreadsheet with the given title.
     Optionally pass sheet_titles to create additional named sheets (tabs)
     beyond the default first sheet, and parent_id to place the spreadsheet
-    in a specific Drive folder.
+    in a specific Drive folder. Because this connector is authorized with
+    the drive.file scope, parent_id only works for a folder this app has
+    already created or that the user explicitly picked; other folder ids
+    will typically be rejected by the Drive API.
     """
     try:
         service = get_sheets_service()
@@ -142,8 +131,12 @@ def google_sheets_create_spreadsheet(
             )
             .execute()
         )
+    except Exception as e:
+        logger.error(f"Error creating spreadsheet: {e}")
+        return json.dumps({"status": "error", "message": str(e)})
 
-        if parent_id:
+    if parent_id:
+        try:
             drive = get_drive_service()
             existing_file = (
                 drive.files()
@@ -159,18 +152,29 @@ def google_sheets_create_spreadsheet(
             if previous_parents:
                 update_kwargs["removeParents"] = ",".join(previous_parents)
             drive.files().update(**update_kwargs).execute()
+        except Exception as e:
+            logger.error(f"Error moving new spreadsheet to parent_id: {e}")
+            return json.dumps(
+                {
+                    "status": "partial",
+                    "spreadsheet_id": spreadsheet.get("spreadsheetId"),
+                    "title": spreadsheet.get("properties", {}).get("title"),
+                    "url": spreadsheet.get("spreadsheetUrl"),
+                    "message": (
+                        "Spreadsheet was created but could not be moved to "
+                        f"parent_id {parent_id!r}: {e}"
+                    ),
+                }
+            )
 
-        return json.dumps(
-            {
-                "status": "success",
-                "spreadsheet_id": spreadsheet.get("spreadsheetId"),
-                "title": spreadsheet.get("properties", {}).get("title"),
-                "url": spreadsheet.get("spreadsheetUrl"),
-            }
-        )
-    except Exception as e:
-        logger.error(f"Error creating spreadsheet: {e}")
-        return json.dumps({"status": "error", "message": str(e)})
+    return json.dumps(
+        {
+            "status": "success",
+            "spreadsheet_id": spreadsheet.get("spreadsheetId"),
+            "title": spreadsheet.get("properties", {}).get("title"),
+            "url": spreadsheet.get("spreadsheetUrl"),
+        }
+    )
 
 
 @mcp.tool()
@@ -204,26 +208,18 @@ def google_sheets_read_range(spreadsheet_id: str, range_name: str) -> str:
 def google_sheets_update_range(
     spreadsheet_id: str,
     range_name: str,
-    values_json: str,
+    values: list[list[Any]],
     value_input_option: str = "USER_ENTERED",
 ) -> str:
     """
     Overwrite cell values in a range, e.g. range_name="Sheet1!A1:B2".
-    values_json must be a JSON-encoded 2D array of rows, e.g.
-    '[["a","b"],["c","d"]]'. value_input_option controls parsing:
-    "USER_ENTERED" (default, parses formulas/numbers like typed input) or
-    "RAW" (stores every value as a literal string).
+    values must be a 2D array of rows, e.g. [["a","b"],["c","d"]].
+    value_input_option controls parsing: "USER_ENTERED" (default, parses
+    formulas/numbers like typed input) or "RAW" (stores every value as a
+    literal string).
     """
     try:
         resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
-        values = json.loads(values_json)
-        if not isinstance(values, list) or not all(
-            isinstance(row, list) for row in values
-        ):
-            raise ValueError(
-                "values_json must be a 2D JSON array of rows (list of lists)"
-            )
-
         service = get_sheets_service()
         result = (
             service.spreadsheets()
@@ -252,24 +248,16 @@ def google_sheets_update_range(
 def google_sheets_append_rows(
     spreadsheet_id: str,
     range_name: str,
-    values_json: str,
+    values: list[list[Any]],
     value_input_option: str = "USER_ENTERED",
 ) -> str:
     """
     Append rows after the last row of data found in range_name, e.g.
-    range_name="Sheet1!A1". values_json must be a JSON-encoded 2D array of
-    rows, e.g. '[["a","b"],["c","d"]]'.
+    range_name="Sheet1!A1". values must be a 2D array of rows, e.g.
+    [["a","b"],["c","d"]].
     """
     try:
         resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
-        values = json.loads(values_json)
-        if not isinstance(values, list) or not all(
-            isinstance(row, list) for row in values
-        ):
-            raise ValueError(
-                "values_json must be a 2D JSON array of rows (list of lists)"
-            )
-
         service = get_sheets_service()
         result = (
             service.spreadsheets()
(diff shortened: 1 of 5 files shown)


src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,393 @@
...
+    sheet_titles: list[str] | None = None,
+    parent_id: str | None = None,
+) -> str:
+    """
+    Create a new Google Sheets spreadsheet with the given title.
+    Optionally pass sheet_titles to create additional named sheets (tabs)
+    beyond the default first sheet, and parent_id to place the spreadsheet
+    in a specific Drive folder.
+    """
+    try:
```
rogercloud, line 121: **[MAJOR] A failed Drive move orphans the spreadsheet and loses its id.**

The whole function body (lines 131-173) — the `spreadsheets().create()` call *and* the two-step Drive move (`drive.files().get()` then `drive.files().update(addParents=...)`) — is wrapped in a single `try/except`. If the move step throws, the handler returns only `{"status": "error", "message": str(e)}` and discards the `spreadsheetId` already returned by the successful create. The spreadsheet then exists, unmoved, in the user's Drive root, and the caller has no id with which to recover it, so a retry is likely to create a duplicate.

This risk is specific to this connector: `google_docs_create_document` can't hit it because Drive's `files().create()` accepts `parents` directly in one atomic call (`google_docs.py:150-151,162-170`). The Sheets API has no equivalent parameter, so this connector is forced into an inherently two-step sequence that the existing single-`try/except` convention does not handle safely.

Suggestion: catch the move-step failure separately and return the already-created `spreadsheet_id` with a `"partial"` status, so the agent/caller can recover the file instead of orphaning it.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,393 @@
...
+            drive = get_drive_service()
+            existing_file = (
+                drive.files()
+                .get(fileId=spreadsheet["spreadsheetId"], fields="parents")
+                .execute()
+            )
+            previous_parents = existing_file.get("parents") or []
+            update_kwargs: dict[str, Any] = {
+                "fileId": spreadsheet["spreadsheetId"],
+                "addParents": parent_id,
```
rogercloud, line 150: **[MAJOR] `addParents=parent_id` is likely rejected under the registered `drive.file` scope.**

The connector registers `spreadsheets` + `drive.file` (`src/xagent/web/builtin_mcp_registry.py:270-273`). `drive.file` grants access only to files and folders the app itself created, or that the user explicitly picked. `parent_id` here is an arbitrary caller-supplied folder id — almost certainly not app-created — so `drive.files().update(addParents=parent_id, ...)` will typically be rejected. In practice that makes `parent_id` a parameter that probably doesn't work.

This follows documented Drive `drive.file` scope semantics; it could not be confirmed with a live API call, and no test exercises `parent_id`. Note it is inherited rather than introduced — `google_docs_create_document`'s own `parent_id` (`google_docs.py:150-151`) has the identical latent limitation. Worth a follow-up issue covering both connectors rather than a fix scoped to this PR, but it is a real limitation of the feature as shipped.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,393 @@
+import json
```
rogercloud, line 1: **[MAJOR] This 393-line connector ships with no unit tests.**

`tests/web/tools/` has test files for every recently added connector (google_ads, google_analytics, zoom, slack, hubspot, linkedin, facebook, instagram, aws), but there is none for google_sheets. The only test in this PR (`tests/alembic/test_20260806_seed_google_sheets_mcp_app.py`) covers the migration and never imports this module — its sole `google_sheets` reference is a string-containment assert against the migration's frozen `ROW`.

As a result all 8 tools, the `values_json` 2D validation, the URL-id resolution, and the two-step create+move flow are entirely unexercised. At minimum, please add tests with a mocked service covering: the 2D-array validation error paths for `update_range` / `append_rows`, `resolve_id_from_url` on both a bare id and a full URL, and the `create_spreadsheet` failure path from the finding above.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,393 @@
...
+logger = logging.getLogger("google-sheets-mcp")
+
+# Ensure standard proxy environment variables are set to prevent hanging requests
+setup_proxy_env()
+
+mcp = FastMCP("google-sheets-mcp")
+
+_SPREADSHEET_URL_ID_PATTERN = re.compile(r"/spreadsheets/d/([a-zA-Z0-9_-]+)")
+
+_sheets_service: Any = None
```
rogercloud: **[MINOR] Remove the module-level service cache to match every sibling connector.**

`_sheets_service` / `_drive_service` (lines 23-24) and the caching in `get_sheets_service` / `get_drive_service` (lines 50-65) are this module's one deviation from its siblings: `google_drive.py:26`, `gmail.py:23`, `calendar.py:21`, `google_docs.py:59,63`, and `google_slides.py:25` all rebuild the service on every call.

The cache buys nothing here — each MCP tool call runs in a fresh short-lived subprocess (`src/xagent/core/tools/adapters/vibe/mcp_adapter.py:741` opens a new session per invocation), so there is no cross-call reuse within a process lifetime. And it has a real downside the siblings avoid: once the service is cached, a refreshed `GOOGLE_ACCESS_TOKEN` env value can no longer be picked up.

Suggest deleting the cache and both globals and rebuilding per call (~-10 lines). Note this reverses an earlier bot suggestion — `functools.lru_cache` would not help, as it keeps both the deviation and the credential-freezing downside.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,393 @@
...
+                                    },
+                                }
+                            }
+                        }
+                    ]
+                },
+            )
+            .execute()
+        )
+        new_sheet = result["replies"][0]["addSheet"]["properties"]
```
rogercloud: **[MINOR] Unguarded `replies` indexing produces a confusing error.**

`new_sheet = result["replies"][0]["addSheet"]["properties"]` indexes directly with no fallback, unlike every other response-parsing site in this file (lines 91-93, 194-195, 286-291 all use `.get()` with defaults). An unexpected or empty `replies` list raises `KeyError`/`IndexError`, which surfaces to the caller as `{"status": "error", "message": "'replies'"}` rather than a clear message. `google_docs.py` avoids exactly this in its analogous batchUpdate handling via `result.get("replies", [])` (`google_docs.py:251,291`).

```python
new_sheet = (result.get("replies") or [{}])[0].get("addSheet", {}).get("properties", {})
```

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,393 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger("google-sheets-mcp")
+
+# Ensure standard proxy environment variables are set to prevent hanging requests
+setup_proxy_env()
+
+mcp = FastMCP("google-sheets-mcp")
+
+_SPREADSHEET_URL_ID_PATTERN = re.compile(r"/spreadsheets/d/([a-zA-Z0-9_-]+)")
```
rogercloud: **[MINOR] Regex misses the multi-account spreadsheet URL form.**

`r"/spreadsheets/d/([a-zA-Z0-9_-]+)"` does not match Google's multi-account URL, `https://docs.google.com/spreadsheets/u/0/d/<id>/edit`. Since `resolve_id_from_url` (`src/xagent/web/tools/mcp/utils.py:6-12`) falls back to returning the input unchanged when the regex misses, such a URL is passed whole as `spreadsheetId` and produces an opaque API error instead of resolving.

Shared verbatim with `google_docs.py:23`'s `_DOCUMENT_URL_ID_PATTERN`, so not unique to this PR, but cheap to fix here (and ideally there too):

```python
r"/spreadsheets/(?:u/\d+/)?d/([a-zA-Z0-9_-]+)"
```

rogercloud (changes requested):
## Summary

This PR adds a Google Sheets builtin MCP connector: a new `src/xagent/web/tools/mcp/google_sheets.py` module exposing 8 FastMCP tools (`get_spreadsheet`, `create_spreadsheet`, `read_range`, `update_range`, `append_rows`, `clear_range`, `add_sheet`, `delete_sheet`) on top of google-api-python-client, a registry row in `src/xagent/web/builtin_mcp_registry.py` that reuses the shared `google` OAuth provider (adding the `spreadsheets` + `drive.file` scopes), and an alembic migration that seeds the catalog row with an accompanying migration-only test. 4 files changed, ~620 lines.

## Design verdict: sound, with reservations

The connector follows the established builtin-MCP-connector pattern faithfully — stdio FastMCP module + registry row + frozen-copy alembic seed migration — and matches `google_docs.py` / `google_slides.py` in tool naming, error-envelope shape (`{"status": "success"/"error", ...}`), and scope shape (the same `spreadsheets` + `drive.file` pair as docs/slides). Three things worth confirming explicitly, since they are easy to get wrong in this pattern and this PR gets all three right:

- The per-app `oauth_scopes` are genuinely merged into the OAuth authorize URL (`src/xagent/web/api/auth.py:1118-1150`, `_merge_oauth_scopes`), so the new scopes actually take effect.
- The migration is **not** redundant with the seeding helper: `seed_builtin_oauth_and_public_mcp_apps` only runs against an empty DB (`src/xagent/web/models/database.py:226,234-235`), so on an existing deployment the alembic migration is the only thing that inserts the row.
- A single alembic head is maintained, and no frontend/docs hardcoded connector list needed updating (the catalog is data-driven).

Nothing about the shape of this PR is over- or under-engineered. The reservations below are execution gaps rather than architectural problems.

**Design concern 1 — no unit tests for the connector module.** The 393-line connector ships with zero unit tests, against a recent house norm of shipping them: `tests/web/tools/` has test files for every recently added connector (google_ads, google_analytics, zoom, slack, hubspot, linkedin, facebook, instagram, aws), but none for google_sheets. The only test added here (`tests/alembic/test_20260806_seed_google_sheets_mcp_app.py`) covers the migration and never imports the connector module — its sole `google_sheets` reference is a string-containment assert against the migration's frozen `ROW`. That leaves all 8 tools, the JSON payload validation, the URL-id resolution, and the two-step create+move flow completely unexercised.

**Design concern 2 — the registry/migration drift test cannot catch a shared error.** `test_seed_row_matches_registry` (`tests/alembic/test_20260806_seed_google_sheets_mcp_app.py:89-98`) asserts `migration.ROW == registry_row`; if both literals carried the same typo (e.g. an identical wrong module path), the test would still pass. A complementary net does exist — `tests/web/api/test_public_mcp_connector_visibility.py:690-757` verifies that a launch command/module is actually importable — but its `expected_python_apps` dict is a subset check that omits `google-sheets` (as it also omits google-ads and zoom). Adding one entry there would have given real defense-in-depth. Minor gap, not a blocker.

## Findings

### Major

1. **Orphaned spreadsheet when the Drive move fails** — `src/xagent/web/tools/mcp/google_sheets.py:131-173`. The whole function body, including both the `spreadsheets().create()` call and the two-step Drive move, sits inside one `try/except`. If the move step throws, the handler returns only `{"status": "error", "message": str(e)}` and discards the `spreadsheetId` already returned by the successful create. The spreadsheet then exists, unmoved, in the user's Drive root with no id returned to the caller, so a retry likely creates a duplicate. This risk is novel to this connector: `google_docs_create_document` avoids it because Drive's `files().create()` accepts `parents` directly in one atomic call (`google_docs.py:150-151,162-170`), whereas the Sheets API has no such parameter and forces an inherently two-step sequence that the existing single-`try/except` convention does not handle safely. Suggested fix: catch the move-step failure separately and return the already-created `spreadsheet_id` with a `"partial"` status so the caller can recover the file.

2. **`drive.file` scope likely cannot move into a caller-supplied folder** — `src/xagent/web/tools/mcp/google_sheets.py:156` together with `src/xagent/web/builtin_mcp_registry.py:270-273`. The registered scopes are `spreadsheets` + `drive.file`, and `drive.file` grants access only to files/folders the app itself created or the user explicitly picked. `parent_id` is an arbitrary caller-supplied folder id, almost certainly not app-created, so `drive.files().update(addParents=parent_id, ...)` will typically be rejected. This follows documented `drive.file` semantics but could not be confirmed with a live API call here, and no test exercises `parent_id` at all. Note this is inherited rather than introduced: `google_docs_create_document`'s own `parent_id` parameter (`google_docs.py:150-151`) has the identical latent limitation. Worth a follow-up issue covering both connectors rather than a fix scoped to this PR — but real, and today it means `parent_id` is a parameter that probably does not work.

3. **No unit tests for `src/xagent/web/tools/mcp/google_sheets.py`** — see design concern 1. Calling it out here as well because it is the most concrete and actionable gap in the PR: the module has no test file under `tests/web/tools/`, unlike every recently added sibling connector.

### Minor

4. **Module-level service caching deviates from every sibling and freezes credentials** — `src/xagent/web/tools/mcp/google_sheets.py:23-24` (`_sheets_service` / `_drive_service` globals) and `50-65` (`get_sheets_service` / `get_drive_service`). Every sibling connector in this directory rebuilds the service on each call instead (`google_drive.py:26`, `gmail.py:23`, `calendar.py:21`, `google_docs.py:59,63`, `google_slides.py:25`). The cache buys nothing here: each MCP tool call runs in a fresh short-lived subprocess (`src/xagent/core/tools/adapters/vibe/mcp_adapter.py:741` opens a new session per invocation), so there is no cross-call reuse within a process lifetime — and unlike the siblings, a refreshed `GOOGLE_ACCESS_TOKEN` cannot be picked up once the service is cached. Suggest deleting the cache and the two globals entirely and rebuilding per call. (Not switching to `functools.lru_cache` — that would preserve the same deviation and the credential-freezing downside.)

5. **Unguarded `replies` indexing in `add_sheet`** — `src/xagent/web/tools/mcp/google_sheets.py:357`. `new_sheet = result["replies"][0]["addSheet"]["properties"]` uses direct bracket indexing with no fallback, unlike every other response-parsing site in this same file (lines 91-93, 194-195, 286-291 all use `.get()` with defaults). An unexpected or empty `replies` list raises `KeyError`/`IndexError` and surfaces as a confusing `{"status": "error", "message": "'replies'"}`. `google_docs.py` avoids exactly this pitfall in its analogous batchUpdate handling via `result.get("replies", [])` (`google_docs.py:251,291`). Suggest: `new_sheet = (result.get("replies") or [{}])[0].get("addSheet", {}).get("properties", {})`.

6. **URL regex misses the multi-account URL form** — `src/xagent/web/tools/mcp/google_sheets.py:21`. `r"/spreadsheets/d/([a-zA-Z0-9_-]+)"` does not match `https://docs.google.com/spreadsheets/u/0/d/<id>/edit`. Because `resolve_id_from_url` (`src/xagent/web/tools/mcp/utils.py:6-12`) falls back to returning the input string unchanged when the regex misses, such a URL is passed whole as `spreadsheetId`, yielding an opaque API error instead of resolving. Shared verbatim with `google_docs.py:23`'s `_DOCUMENT_URL_ID_PATTERN`, so not unique to this PR, but cheap to fix here (and ideally there too): `r"/spreadsheets/(?:u/\d+/)?d/([a-zA-Z0-9_-]+)"`.

## Re-review of prior gemini-code-assist findings

| Prior finding | Status |
| --- | --- |
| Defensive `.get()` access for `gridProperties` (medium) | **FIXED** — `google_sheets.py:92-93` now uses `properties.get("gridProperties") or {}`. |
| Explicitly remove previous Drive parents when moving a new spreadsheet (medium) | **FIXED** — `google_sheets.py:148-161` fetches `existing_file.get("parents")` and passes it as `removeParents`. |
| Avoid passing empty-string `removeParents` causing a 400 (high) | **FIXED** — the same block only sets `update_kwargs["removeParents"]` inside `if previous_parents:`. |
| Validate `values_json` is a 2D array, not 1D (medium, raised twice) | **FIXED** in both call sites — `google_sheets.py:219-225,265-271` check `isinstance(values, list) and all(isinstance(row, list) for row in values)` and raise `ValueError` otherwise. |
| Cache Google API service instances with `static_discovery=True` (medium) | **FIXED as requested, but we now recommend reversing it** — see minor finding 4. The cache is the module's one deviation from every sibling connector, buys nothing because each tool call is a fresh subprocess, and prevents a refreshed access token from being picked up. |
| Rename `sheet_id` → `resolved_spreadsheet_id` to disambiguate from the per-tab id (medium) | **FIXED** — `sheet_id` is now used exclusively for the numeric per-tab id (`add_sheet`, `delete_sheet`) and `resolved_spreadsheet_id` exclusively for the resolved spreadsheet file id. |
| Move `logging.basicConfig` into the `if __name__ == "__main__":` guard (medium) | **NOT FIXED — waived.** Still at module level (`google_sheets.py:13`), but byte-for-byte consistent with `google_drive.py:17` and `gmail.py:14`. This is house style, not an isolated oversight; changing it only here would create inconsistency. If worth fixing, it belongs in a repo-wide follow-up. |
| Use a native `values: list[list[Any]]` parameter instead of `values_json: str` + manual `json.loads` (medium, raised twice) | **NOT FIXED — waived.** The JSON-string convention matches siblings: `google_docs.py:267` (`requests_json: str`) and `hubspot.py` (`properties_json: str`, across 4+ functions). A repo-wide convention change, not something to demand of this PR alone. |

- `delete:` `google_sheets.py:23-24,50-65` — drop the module-level `_sheets_service` / `_drive_service` cache and rebuild per call, matching `google_drive.py`, `gmail.py`, `calendar.py`, `google_docs.py`, and `google_slides.py`.

net: ~-10 lines possible

## Tests

`pytest tests/alembic/test_20260806_seed_google_sheets_mcp_app.py -q` → 6 passed; `pytest tests/web/api/test_public_mcp_connector_visibility.py -q` → 44 passed. No test exists for `src/xagent/web/tools/mcp/google_sheets.py` itself (major finding 3).

tests/web/tools/test_google_sheets_mcp.py
```diff
@@ -0,0 +1,341 @@
...
+
+@pytest.mark.parametrize(
+    ("value", "expected"),
+    [
+        ("abc123", "abc123"),
+        ("https://docs.google.com/spreadsheets/d/abc123/edit#gid=0", "abc123"),
+        ("https://docs.google.com/spreadsheets/u/1/d/abc123/edit", "abc123"),
+    ],
+)
+def test_resolve_spreadsheet_id_accepts_bare_id_and_url_forms(value, expected):
```
rogercloud, line 61: **[MAJOR] The test suite can't detect a regression that breaks Sheets-URL support.**

All seven id-taking tools (`google_sheets_read_range`, `_update_range`, `_append_rows`, `_clear_range`, `_add_sheet`, `_delete_sheet`, `_get_spreadsheet`) call `_resolve_spreadsheet_id(spreadsheet_id)` before use, but every test for those functions passes an already-bare id (`"sid"`). This test only exercises the resolver helper directly, bypassing all seven real call sites. Deleting the `_resolve_spreadsheet_id(...)` call from any tool would leave all 22 tests green while silently breaking the "paste a Sheets URL" UX for that tool.

Suggestion: add at least one case per call site (a single parametrized test over the seven tools works) that passes a full URL to the actual tool function and asserts the *resolved* id reached the API, e.g. `spreadsheets.values.return_value.get.call_args.kwargs["spreadsheetId"] == "abc123"`.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,384 @@
...
+            service.spreadsheets()
+            .values()
+            .get(spreadsheetId=resolved_spreadsheet_id, range=range_name)
+            .execute()
+        )
+        return json.dumps(
+            {
+                "status": "success",
+                "range": result.get("range", range_name),
+                "values": result.get("values", []),
```
rogercloud: **[MINOR] No cap on returned payload size.**

`range_name` is passed straight through to the Sheets API and the full `result.get("values", [])` is serialized with no truncation. A bare tab reference (`range_name="Sheet1"`) is valid A1 notation for the entire sheet, so a large spreadsheet can dump a very large payload directly into the calling agent's context. Sibling connectors already guard this: both `gmail.py` and `google_drive.py` expose a `max_results` parameter.

Suggestion: add a `max_rows: int = 1000`-style cap and set `"truncated": true` in the response when it is hit.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,384 @@
...
+            )
+            .execute()
+        )
+    except Exception as e:
+        logger.error(f"Error creating spreadsheet: {e}")
+        return json.dumps({"status": "error", "message": str(e)})
+
+    if parent_id:
+        try:
+            drive = get_drive_service()
```
rogercloud, line 141: **[MINOR] Two independent `Credentials` objects for one logical operation.**

When `parent_id` is set, both `get_sheets_service()` (line 121) and `get_drive_service()` (this line) are called, each independently invoking `_get_credentials()`. Credential construction does no network I/O so this is cheap today — purely a DRY/clarity nit. Optional: build credentials once and pass them into both service builders.

src/xagent/web/tools/mcp/google_sheets.py
```diff
@@ -0,0 +1,384 @@
...
+    try:
+        resolved_spreadsheet_id = _resolve_spreadsheet_id(spreadsheet_id)
+        service = get_sheets_service()
+        result = (
+            service.spreadsheets()
+            .values()
+            .get(spreadsheetId=resolved_spreadsheet_id, range=range_name)
+            .execute()
+        )
+        return json.dumps(
```
rogercloud, line 211: **[MINOR] Inconsistent `ensure_ascii` on `json.dumps`.**

None of the 17 `json.dumps` calls in this file pass `ensure_ascii=False`, so non-ASCII sheet titles and cell values are emitted as `\uXXXX` escapes. Still valid JSON and no functional bug — and it matches the older Google cohort (`gmail.py`, `google_docs.py`) — but every connector added since (`google_ads`, `google_analytics`, `hubspot`, `slack`, `zoom`, `google_slides`) passes `ensure_ascii=False`. Optional: align this new file with the newer convention.

tests/web/tools/test_google_sheets_mcp.py
```diff
@@ -0,0 +1,341 @@
...
+    )
+    _mock_sheets_service(monkeypatch, spreadsheets)
+
+    result = json.loads(google_sheets.google_sheets_create_spreadsheet("New"))
+
+    assert result["status"] == "error"
+    assert "quota exceeded" in result["message"]
+
+
+def test_read_range_returns_values(monkeypatch):
```
rogercloud, line 242: **[MINOR] Missing-field default paths untested for three tools.**

`google_sheets_read_range`, `_append_rows`, and `_clear_range` all parse the API response with `.get(key, default)` (e.g. `result.get("values", [])`), but no test exercises the case where the key is genuinely absent — unlike `get_spreadsheet` and `add_sheet`, which each have a dedicated missing-field test. Risk is low (a broken default would raise `KeyError`, be caught by the surrounding `except Exception`, and degrade to a generic error payload rather than crash), but worth closing for consistency with the rest of the suite.

rogercloud (approved):
## Summary

This PR adds a Google Sheets MCP connector to the builtin connector family. It introduces `src/xagent/web/tools/mcp/google_sheets.py`, a stdio FastMCP server exposing 8 tools (read/update/append/clear ranges, add/delete sheets, create spreadsheet, get metadata) on top of the Sheets v4 API, plus an optional Drive move for `parent_id`. It registers the app in `src/xagent/web/builtin_mcp_registry.py` reusing the shared `google` OAuth provider with per-app functional scopes (`spreadsheets`, `drive.file`), and adds a seed migration (`20260806_seed_google_sheets_mcp_app`) so existing deployments pick up the new builtin app. A new test module, `tests/web/tools/test_google_sheets_mcp.py`, covers the tool surface.

## Round 0 — Design verdict: sound

Evaluated independently of the prior review round. This is a faithful, low-risk instance of the established Google-connector pattern in this repo, and the design choices hold up on their own merits:

- **Registry entry matches siblings exactly.** The new entry mirrors `google-docs` and `google-slides` in transport, provider reuse, `env_mapping`, and category — no new registration mechanism is introduced.
- **Reusing the shared `google` OAuth provider is correct, not a stretch.** Per-app functional scopes are merged at authorize time through the existing `_merge_oauth_scopes` path that every Google-family connector already uses. Token refresh happens in the parent process before the MCP subprocess launches, so no per-service credential storage is required.
- **The 8-tool surface is at the right altitude.** Each tool maps 1:1 onto a Sheets v4 primitive. Notably it does *not* expose a raw `batchUpdate` escape hatch (which `google_slides.py` does) — a narrower and safer surface for an agent-facing tool.
- **Scope minimality is right.** `drive.file` (not the broad `drive`) is requested, and only for the optional `parent_id` move — consistent with google-docs/google-slides and narrower than the google-drive connector.
- **The migration follows the established seed pattern** used by ~9 sibling migrations: table-exists guard, existing-id guard, column filtering, idempotent, and a single alembic head is maintained.
- **No security or concurrency concerns found.** No token/secret leakage in error paths; spreadsheet ids are regex-extracted or passed as typed API parameters and never interpolated into a fetched URL (no SSRF/injection surface); sync tools dispatch to a worker thread under FastMCP, matching siblings, so there is no blocking-in-async issue.

Two design observations are **out of scope for this PR** — they are pre-existing repo-wide patterns this PR correctly follows rather than introduces, and would each need a change across all five Google-family connector modules:

- `_get_credentials()` is now the fifth verbatim copy of the same credential-builder across `google_sheets` / `google_docs` / `gmail` / `google_slides` / `google_drive`. A shared `build_google_credentials()` in `utils.py` would be a reasonable repo-wide follow-up.
- The `refresh_token` / `client_id` / `client_secret` branch in `_get_credentials()` is effectively dead: the MCP child subprocess only receives the access token via `env_mapping`, so in-subprocess self-refresh can never trigger. This is pre-existing in all four sibling modules and is not a regression — refresh is correctly handled by the parent before launch. Worth a repo-wide cleanup issue rather than a change here.

## Verification of prior review findings

| # | Prior finding | Status | Evidence |
|---|---|---|---|
| 1 | [Major] Orphaned spreadsheet when the Drive move fails | **FIXED** | The `spreadsheets().create()` call and the Drive-move step are now in separate `try`/`except` blocks; a failed move returns `{"status": "partial", "spreadsheet_id": ...}` instead of discarding the id. Covered by `test_create_spreadsheet_returns_partial_status_when_move_fails`. |
| 2 | [Major] `drive.file` scope likely rejects `addParents` for arbitrary folders | **FIXED (documented limitation)** | This is what the prior review asked for. The `google_sheets_create_spreadsheet` docstring now states that `parent_id` only works for a folder this app created or the user explicitly picked, and that other folder ids will typically be rejected. Combined with #1, a rejection now surfaces as `partial` with the spreadsheet id preserved rather than as data loss. |
| 3 | [Major] No unit tests for the connector module | **PARTIAL** | `tests/web/tools/test_google_sheets_mcp.py` adds 22 passing tests covering all 8 tools, credential handling, and URL resolution. Two of the three specifically requested cases (URL/bare-id resolution, create_spreadsheet failure path) are covered; `append_rows` still has no dedicated test for its 2D-array validation error path — only `update_range`'s equivalent is exercised (`test_update_range_validation_via_mcp_layer`). Minor, since both rely on the identical FastMCP-level `list[list[Any]]` validation. |
| 4 | [Minor] Module-level service caching freezes credentials | **FIXED** | The `_sheets_service` / `_drive_service` globals and caching logic are gone; `get_sheets_service()` / `get_drive_service()` rebuild per call, matching `google_docs.py`, `google_drive.py`, `gmail.py`, `calendar.py`, and `google_slides.py`. |
| 5 | [Minor] Unguarded `replies` indexing in `add_sheet` | **FIXED** | Now `(result.get("replies") or [{}])[0].get("addSheet", {}).get("properties", {})`, with regression test `test_add_sheet_handles_missing_replies_without_raising`. |
| 6 | [Minor] URL regex misses the multi-account form | **FIXED** | Regex is now `r"/spreadsheets/(?:u/\d+/)?d/([a-zA-Z0-9_-]+)"`, with a parametrized test covering bare id, plain URL, and `/u/0/d/` forms. |

Lean already.

## Test results

- `pytest tests/web/tools/test_google_sheets_mcp.py` → 22 passed
- `pytest tests/alembic/test_20260806_seed_google_sheets_mcp_app.py` → 6 passed
- `alembic heads` → single head (`20260806_seed_google_sheets_mcp_app`)
- `python -c "import xagent.web.tools.mcp.google_sheets"` → imports cleanly

## Recommendation

**Approve.** All six findings from the previous round are resolved (finding #3 substantially, with one small gap noted), the design is sound on independent evaluation, and the follow-up rebase touched only the migration's `down_revision` to resolve the alembic multi-head conflict. The remaining items below are test-robustness/polish, not shipped defects — the URL-resolution behavior itself is correct and covered at the helper level. I'd suggest landing the URL-through-the-tool test either as a follow-up commit here or in a quick follow-up PR, and treating the minor items as optional polish. The two repo-wide credential-helper observations should become their own issue rather than blocking this connector.

PR #1172, opened 2026-08-06
feat(deepdoc): add Xinference remote-parsing environment configuration

OliverBryant:
Closes #1162. Step **S1** of the umbrella issue #1161 (DeepDoc remote parsing via Xinference GPU offload).

## What

Adds the three environment variables that will enable remote DeepDoc parsing, plus their getters in `src/xagent/config.py`:

| Variable | Default | Notes |
| --- | --- | --- |
| `XAGENT_DEEPDOC_XINFERENCE_URL` | unset (= local mode) | Xinference base URL, e.g. `http://gpu-host:9997`. Validated as `http`/`https`, trailing slashes stripped. |
| `XAGENT_DEEPDOC_XINFERENCE_API_KEY` | falls back to the bare `XINFERENCE_API_KEY`, then to no auth header | A self-hosted Xinference deployment often runs without authentication, so `None` is a valid result. |
| `XAGENT_DEEPDOC_XINFERENCE_TIMEOUT_SECONDS` | `1800` | Read timeout for one whole-document parse, matching the `timeout=1800` precedent in deepdoc-lib's own remote API client. |

## Why

Configuration-only groundwork so the later steps in #1161 have a single, validated source of truth to read from. Nothing consumes these getters yet, so this PR is inert at runtime.

Two deliberate choices:

- **A malformed URL raises instead of falling back to local parsing.** Remote parsing is a performance feature; if a typo silently downgraded to local mode, the only symptom would be unexplained slowness. `_normalized_http_env_url` already surfaces this at the configuration boundary, so the getter reuses it.
- **The API key falls back to the bare `XINFERENCE_API_KEY`.** This mirrors the existing behaviour of the Xinference rerank client (`src/xagent/core/model/rerank/xinference.py:57`), so a deployment that already sets one shared key does not need a second one.

Both getters reuse the existing `_normalized_http_env_url` / `_get_positive_int_env` helpers, so `config.py` keeps zero dependencies on other core submodules per `CLAUDE.md`.

The `example.env` documentation for these three variables is owned by sub-issue #1167 and landed separately in #1170, so it is intentionally not touched here.

## Testing

[...]

PR #1170, opened 2026-08-06
docs(deepdoc): document remote Xinference parsing configuration

OliverBryant:
Closes #1167 · Part of #1161 (step S6)

## What

Adds a `DeepDoc Remote Inference (Xinference)` section to `example.env` documenting the three environment variables that control remote DeepDoc document parsing:

- `XAGENT_DEEPDOC_XINFERENCE_URL` — Xinference base URL; unset means parsing runs locally.
- `XAGENT_DEEPDOC_XINFERENCE_API_KEY` — optional, falls back to `XINFERENCE_API_KEY`.
- `XAGENT_DEEPDOC_XINFERENCE_TIMEOUT_SECONDS` — per-document parse timeout, default `1800`.

## Why

`CLAUDE.md` requires new configuration to be reflected in `example.env`, which is the discovery surface operators read when setting up a deployment. Offloading the OCR/layout/table models to a GPU-backed Xinference server is opt-in and invisible without this, so the section also states the two behaviors that are easy to get wrong: setting the URL routes **all** DeepDoc-supported formats remotely, and a remote failure logs a warning and falls back to local parsing rather than failing the parse.

All three variables are optional, so all three assignments stay commented out — matching the file's convention that uncommented lines are values a deployment must actually set.

## Notes

- Documentation only: `example.env` is the sole file touched, additions only.
- The config getters that read these variables land in #1162; the variable names here were cross-checked against section 5 of `docs/deepdoc-remote-parsing.md` (added by the umbrella PR #1169).
- Placed after the embedding/vector-store block so it sits with the other model-inference configuration rather than at the end of the file.

## Verification

- All three names match the requirements doc exactly.
- `end-of-file-fixer` / `trailing-whitespace` invariants verified (single trailing newline, no trailing whitespace, no CRLF); `pre-commit run` could not bootstrap its mypy mirror in this sandbox, but no hook applies to `.env` beyond those two.

Issue #1165, opened 2026-08-06
feat(deepdoc): route parsing to the remote client with local fallback


OliverBryant:
Part of #1161 · Step **S4** · depends on: S2 (client module)

Route parsing through the remote client when it is configured, fall back to local on failure, and stop loading local ONNX models on the remote path.

## Scope

`src/xagent/providers/pdf_parser/deepdoc.py`

### The problem being fixed

`_sync_parse` calls `self._get_parser(file_path)` at the top of the function (around line 526), and `DeepDocPdfParser()` eagerly loads the ONNX models — possibly downloading them from ModelScope. On the remote path none of that is needed, so instantiation has to be deferred.

### Changes, in this order

1. **Extension detection first.** Replace the top-of-function parser instantiation with extension detection only: `kwargs.get("file_ext", ".xlsx")` for `BytesIO`, otherwise `Path(file_path).suffix.lower()`.
2. **Defer local instantiation.** Move `parser = self._get_parser_for_ext(ext)` down to just before each local branch actually uses it. Leave untouched: the `.xlsx` `_parse_xlsx_rows` early return, the `.md`/`.txt` direct-read paths, the `BytesIO` conversion logic, and the `run_in_executor` wrapper.
3. **Progress wiring for all formats.** Hoist the `DeepDocProgressAdapter` construction out of the PDF branch so every format gets a callback in remote mode. The local PDF branch keeps passing it as before.
4. **Remote-first dispatch.** After extension detection and after the cheap `validate_office_file_format` magic-byte check (keep that before the upload so a bad file fails fast without wasting bandwidth), and before any local branch:

```python
if deepdoc_remote.is_remote_configured():
    try:
        elements = deepdoc_remote.parse_document_remote(
            file_path, ext=ext,
            save_image=lambda b: _save_bytes_to_disk(doc_id, b, ".png"),
            callback=callback, zoomin=parser_call_kwargs.get("zoomin", 3),
        )
        metadata["deepdoc_backend"] = "remote"
        return _translate_remote_elements(doc_id, elements, **metadata)
    except DeepDocRemoteError as exc:
        logger.warning("Remote DeepDoc parse failed (%s); falling back to local", exc)
        if callback:
            callback(0.01, "Remote parse failed; running local DeepDoc")
metadata["deepdoc_backend"] = "local"
# existing local dispatch continues unchanged
```

[...]

Issue #1164, opened 2026-08-06
test(deepdoc): unit-test the remote parse client with mocked HTTP


OliverBryant:
Part of #1161 · Step **S3** · depends on: S2 (client module)

Unit-test the remote client with no network and no models, so it runs on any machine. Model the style on the existing pure-unit test `tests/providers/pdf_parser/test_deepdoc_table_positions.py`.

## Scope

New file `tests/providers/pdf_parser/test_deepdoc_remote.py`, using `httpx.MockTransport` injected through the `_transport` seam.

### Success cases

- A response with one text element (`image_base64: null`) and one table element carrying a base64 1x1 PNG: asserts the returned dicts have the local element shape, `save_image` is called exactly once, the table element's `image` is the returned path, and the text element's `image` is `None`.
- Request shape: multipart contains the file with the right filename, `zoomin` is forwarded, and the `Authorization: Bearer …` header is present when a key is configured and absent when it is not.
- `BytesIO` input: filename falls back to `document{ext}` and the stream position is restored.

### Failure cases — each must raise `DeepDocRemoteError`

- HTTP 500, HTTP 401
- connection error, read timeout
- non-JSON response body
- valid JSON without an `elements` key, and `elements` holding non-dict entries
- an element whose `image_base64` is not decodable base64

### `is_remote_configured()`

- env unset → `False`
- malformed URL (e.g. `ftp://host`) → `False` **without raising**

## Done when

`pytest tests/providers/pdf_parser/test_deepdoc_remote.py -q` passes with no network access and no DeepDoc model cache present.

PR #1173, opened 2026-08-06
feat(deepdoc): add the remote Xinference parse client module

### LINKED ISSUE #1163: feat(deepdoc): add the remote Xinference parse client module
Part of #1161 · Step **S2** · depends on: S1 (config getters)

Add the HTTP client that uploads a document to the remote Xinference DeepDoc endpoint and normalizes the response into the element shape xagent already consumes.

## Scope

New file `src/xagent/providers/pdf_parser/deepdoc_remote.py`.

Dependencies: `httpx` (already a hard dependency, `pyproject.toml:37`) and `config` only. It must **not** import `deepdoc.py` — the image-saving function is injected by the caller, which keeps the import graph acyclic.

Put the proposed server API contract (see `docs/deepdoc-remote-parsing.md` section 6) in the module docstring.

### Pieces

- `class DeepDocRemoteError(Exception)` — the single exception type that signals "fall back to local".
- `is_remote_configured() -> bool` — true when `get_deepdoc_xinference_url()` is not `None`. Must **catch the `ValueError`** that a malformed URL raises, log a warning, and return `False`, so a typo degrades to local parsing instead of breaking every parse.
- `_build_headers() -> dict[str, str]` — `{"Authorization": f"Bearer {key}"}` when a key is configured, `{}` otherwise.
- `_post_document(file_path, *, ext, zoomin, transport) -> dict` — builds the multipart body (`files={"file": (filename, fh, mime)}`); for a `str` path open in binary mode and derive the filename from `Path.name`, for a `BytesIO` seek to 0, use `f"document{ext}"` as the filename, and restore the position afterwards. Uses `httpx.Client(timeout=httpx.Timeout(connect=10, read=T, write=T, pool=10), transport=transport)`, POSTs to `{url}/v1/document/parse`, calls `raise_for_status()`, returns `r.json()`. The `transport` parameter is the test seam for `httpx.MockTransport`.
- `_normalize_elements(payload, save_image) -> list[dict]` — validates that `payload["elements"]` is a list of dicts carrying `type` and `text`, otherwise raises `ValueError`. For each element, pops `image_base64`; when non-empty sets `el["image"] = save_image(base64.b64decode(...))` (a **path string**, so the existing `_handle_image` string branch in `deepdoc.py` works unchanged), otherwise `el["image"] = None`.
- `parse_document_remote(file_path, *, ext, save_image, callback=None, zoomin=3, _transport=None) -> list[dict]` — emits `callback(0.05, "Uploading document to remote DeepDoc server")`, calls `_post_document`, then `_normalize_elements`, then `callback(1.0, f"Remote DeepDoc parse finished ({elapsed:.2f}s)")`. Wraps **every** failure mode — `httpx.HTTPError` and subclasses, `json.JSONDecodeError`, `ValueError`, `binascii.Error` — in `raise DeepDocRemoteError(...) from exc`. Nothing else may escape.

The `"msg (Xs)"` message format is deliberate: the existing `DeepDocProgressAdapter` in `src/xagent/core/tools/core/RAG_tools/progress/adapters.py` strips the timing suffix and dedupes statuses, so it needs no change.

## Done when

The module imports cleanly on its own, has no `deepdoc.py` import, and `ruff` / `mypy src/xagent` are clean. Tests land in S3 (#1164).

OliverBryant:
Closes #1163 · Step **S2** of #1161

## Stacking

> [!IMPORTANT]
> This PR is stacked on **#1172** (S1, config getters) and **must merge after it**.
> The branch is based on `feat/deepdoc-remote-config` because this module calls the
> three `get_deepdoc_xinference_*` getters that #1172 adds. Merging this one first
> would break the build.
>
> Because the base here is `main`, the diff below also replays #1172's two commits
> and its files (`src/xagent/config.py`, `tests/core/test_config.py`). **The only
> file this PR is asking you to review is
> `src/xagent/providers/pdf_parser/deepdoc_remote.py`** — everything else belongs to
> #1172 and disappears from this diff once that merges.

## What

Adds `src/xagent/providers/pdf_parser/deepdoc_remote.py`, the HTTP client that
uploads a document to a remote Xinference DeepDoc endpoint and normalizes the
response into the element shape xagent already consumes.

- `DeepDocRemoteError` — the single exception type meaning "fall back to local".
- `is_remote_configured()` — true when a URL is configured. It catches the
  `ValueError` a malformed URL raises, logs a warning, and returns `False`.
- `_build_headers()` — a `Bearer` header when a key is configured, `{}` otherwise.
- `_post_document()` — multipart upload to `{base_url}/v1/document/parse`. Handles
  both a `str` path (filename from `Path.name`) and a `BytesIO` buffer (synthetic
  `document{ext}` filename, cursor restored afterwards since the caller may still
  read from the buffer during fallback). The `transport` parameter is the seam
  S3's `httpx.MockTransport` tests use.
- `_normalize_elements()` — validates the payload carries a list of element
  objects each with `type` and `text`, then pops `image_base64` and replaces it
  with an `image` **path string** produced by the injected `save_image`.
- `parse_document_remote()` — orchestrates the two, emits the two progress
  notices, and wraps every failure in `DeepDocRemoteError`.

The proposed server API contract from `docs/deepdoc-remote-parsing.md` section 6
is reproduced in the module docstring.

## Why these shapes

- **`httpx` + `config` only, and no `deepdoc.py` import.** The image-saving
  function is injected by the caller instead, which keeps the import graph
  acyclic and lets the client be tested without loading any parser or ONNX model.
- **A path string, not a PIL image.** `_handle_image` in `deepdoc.py` already has
  a `str` branch that validates and returns the path, so S4 needs no change there.
- **The `"msg (Xs)"` completion format is deliberate.** `DeepDocProgressAdapter`
  splits on `" ("` to strip timing and dedupes by status, so repeated finishes
  with different timings collapse to one status update and the adapter needs no
  change. Verified below.
- **Fallback is unconditional.** Every exception path — including one from the
  caller-supplied `save_image`, which this module cannot enumerate — becomes
  `DeepDocRemoteError`. Nothing else escapes.

Only `deepdoc_remote.py` is added. `deepdoc.py` routing is S4 (#1165), unit tests
are S3 (#1164), and `example.env` already landed in #1170.

## Testing

Unit tests land in S3 (#1164). Verification run here:

```console
$ .venv/bin/python -m ruff format --check src/xagent/providers/pdf_parser/deepdoc_remote.py
1 file already formatted

$ .venv/bin/python -m ruff check src/xagent/providers/pdf_parser/deepdoc_remote.py
All checks passed!

$ uvx --quiet isort==6.0.1 --check-only --diff src/xagent/providers/pdf_parser/deepdoc_remote.py
(no output — no diff)

$ .venv/bin/python -m mypy src/xagent/providers/pdf_parser/deepdoc_remote.py
Found 26 errors in 9 files (checked 1 source file)
# zero errors in deepdoc_remote.py. All 26 are the pre-existing missing-stub
# baseline in other files (cairosvg, lancedb, playwright, pyarrow, bs4);
# `mypy src/xagent/providers/pdf_parser/deepdoc.py` on the untouched file
# reports the identical 26.

$ grep -nE "from \.deepdoc|import deepdoc" src/xagent/providers/pdf_parser/deepdoc_remote.py
(no output — no deepdoc.py import)

$ PYTHONPATH=src .venv/bin/python -c \
    "from xagent.providers.pdf_parser import deepdoc_remote; print(deepdoc_remote.is_remote_configured())"
False
```

Behavioral smoke test against `httpx.MockTransport` (success path, request
shaping, progress-adapter interaction, `BytesIO` handling):

```
configured: True
headers: {'Authorization': 'Bearer sk-test'}
request: {'url': 'http://gpu-host:9997/v1/document/parse', 'auth': 'Bearer sk-test',
          'zoomin_7': True, 'fname': True, 'mime': True}
elements: [{"type": "text", "text": "hello", "image": null},
           {"type": "table", "text": "<table/>", "metadata": {"page_number": 2},
            "image": "/artifacts/img1.png"}]
progress: [(0.05, 'Uploading document to remote DeepDoc server'),
           (1.0, 'Remote DeepDoc parse finished (0.00s)')]
adapter statuses: ['Uploading document to remote DeepDoc server',
                   'Remote DeepDoc parse finished']
bytesio fname ok: True | pos restored: True
```

The trailing slash on the configured URL is stripped by the config getter, so the
built URL has no double slash. Feeding both progress notices plus a second
`"Remote DeepDoc parse finished (9.99s)"` through `DeepDocProgressAdapter` yields
exactly the two statuses above — the timing suffix is stripped and the repeat is
deduped, confirming the adapter needs no change.

Every failure mode returns `DeepDocRemoteError` and nothing escapes:

```
500 status:              DeepDocRemoteError OK
401 status:               DeepDocRemoteError OK
empty 200 body:           DeepDocRemoteError OK
html 200 body:            DeepDocRemoteError OK
json list body:           DeepDocRemoteError OK
elements None:            DeepDocRemoteError OK
elements missing:         DeepDocRemoteError OK
elements not list:        DeepDocRemoteError OK
element not dict:         DeepDocRemoteError OK
element no text:          DeepDocRemoteError OK
bad base64:               DeepDocRemoteError OK
read timeout:             DeepDocRemoteError OK
connect refused:          DeepDocRemoteError OK
save_image RuntimeError:  DeepDocRemoteError OK
save_image OSError:       DeepDocRemoteError OK
missing local file:       DeepDocRemoteError OK
malformed url parse:      DeepDocRemoteError OK

malformed configured: False   # is_remote_configured() degrades to local
```

Exception ordering matters and is covered: `json.JSONDecodeError` and
`binascii.Error` are both `ValueError` subclasses and are caught ahead of the
bare `ValueError`, and `httpx.InvalidURL` is a `ValueError` rather than an
`httpx.HTTPError`, so it is caught too.


src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Returns:
+        The parsed elements, each with an ``image`` path string or ``None``.
+
+    Raises:
+        DeepDocRemoteError: On any failure. Callers fall back to local parsing.
+    """
+    started = time.monotonic()
+    if callback is not None:
+        callback(0.05, "Uploading document to remote DeepDoc server")
```
rogercloud, line 309: **major** — Both progress `callback(...)` invocations sit outside the `try` block (lines 311-336) that wraps every other failure into `DeepDocRemoteError`: this one at line 308-309, and the completion notice at lines 339-340. The docstring at line 305 promises `DeepDocRemoteError: On any failure. Callers fall back to local parsing.` — but a caller-supplied callback that raises will propagate straight out of `parse_document_remote` and crash the caller instead of triggering the documented fallback.

This is masked today: `DeepDocProgressAdapter` (`src/xagent/core/tools/core/RAG_tools/progress/adapters.py`) swallows its own exceptions, and it is the only adapter in-tree. But the signature is `Optional[ProgressCallback]` — an arbitrary caller-supplied callable — and the contract does not guarantee that behavior of every callback. It is cheaper to close this now than after S4/S5 wire up callers.

Suggested fix — either move both calls inside the `try`, or (probably cleaner, since it keeps progress reporting from ever being the reason a parse fails) route them through a guard that cannot escape:

```python
def _notify(fraction: float, message: str) -> None:
    if callback is None:
        return
    try:
        callback(fraction, message)
    except Exception:  # progress reporting must never fail the parse
        logger.warning("Remote DeepDoc progress callback raised", exc_info=True)
```

then call `_notify(0.05, ...)` and `_notify(1.0, ...)` in place of the two direct invocations.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+
+        normalized_element = dict(element)
+        image_base64 = normalized_element.pop("image_base64", None)
+        if image_base64:
+            # validate=True rejects non-alphabet characters instead of dropping
+            # them. Without it a corrupt-but-right-length payload decodes to
+            # b"" and gets written out as a zero-byte PNG, so the caller
+            # receives a broken image path rather than falling back to local
+            # parsing.
+            image_bytes = base64.b64decode(image_base64, validate=True)
```
rogercloud, line 264: **minor** — `base64.b64decode(image_base64, validate=True)` accepts only the standard alphabet, unwrapped. It raises `binascii.Error` on MIME-wrapped input (76-char lines, what `base64.encodebytes()` produces) and on the URL-safe alphabet. Because this decode happens per-element inside the loop with no per-element catch, a single bad image aborts the *whole* document's remote parse — the error propagates to `except binascii.Error` at line 322 and the entire document falls back to local.

I'm not asking to relax `validate=True` — it was added deliberately (564c15e8) to reject corrupt payloads that would otherwise silently decode to `b""`, and standard-alphabet base64 is the overwhelmingly common convention for JSON payloads. The gap is that the server does not exist yet (the contract at lines 14-48 is explicitly a proposal) and the contract does not pin an encoding variant, so a future server implementer could reach for `encodebytes()` or `urlsafe_b64encode()` and produce a client that rejects every image.

Suggested fix — make it explicit at line 41 in the contract block, e.g. `"image_base64": null,  // standard-alphabet base64, unwrapped (no newlines); PNG for table/figure, null otherwise`.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def _send(filename: str, fh: Any) -> dict[str, Any]:
+        with httpx.Client(timeout=timeout, transport=transport) as client:
+            response = client.post(
+                f"{base_url}{PARSE_ENDPOINT}",
+                headers=_build_headers(),
+                data=data,
+                files={"file": (filename, fh, mime_type)},
+            )
+            response.raise_for_status()
```
rogercloud, line 192: **major** — `response.raise_for_status()` raises `httpx.HTTPStatusError` without ever reading the response body. Verified against httpx's source: that exception's message contains only the status code, reason phrase, and URL. So the wrapped message produced at line 317 carries no server-side detail at all.

That matters more than usual here, because this module's own contract at line 48 states the server's error responses are **always** shaped `{"detail": "..."}` — that `detail` field is the one diagnostic the contract guarantees, and it is silently thrown away. Combined with the fact that failures are silent by design (fall back to local, no user-facing error) and that HTTP status is currently the only *implemented* failure surface, an operator debugging why remote parsing never engages — wrong API key → 401, oversized file → 413, unsupported format → 400 — gets nothing but a generic status line.

Suggested fix — catch `httpx.HTTPStatusError` specifically, *before* the general `httpx.HTTPError` handler at line 316 (it is a subclass, so ordering matters), and fold in a truncated body:

```python
except httpx.HTTPStatusError as exc:
    detail = exc.response.text[:500]
    raise DeepDocRemoteError(
        f"Remote DeepDoc request failed: {exc} (body: {detail})"
    ) from exc
```

Reading `.text` is safe here because the response is fully buffered by the non-streaming `client.post`.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+    started = time.monotonic()
+    if callback is not None:
+        callback(0.05, "Uploading document to remote DeepDoc server")
+
+    try:
+        payload = _post_document(
+            file_path, ext=ext, zoomin=zoomin, transport=_transport
+        )
+        elements = _normalize_elements(payload, save_image)
+    except httpx.HTTPError as exc:
```
rogercloud, line 316: **minor (should-fix, not nice-to-have)** — `logger.info` fires only on the success path (line 341). All six `except` clauses here (lines 316-336, covering `httpx.HTTPError`, `json.JSONDecodeError`, `binascii.Error`, `ValueError`, `OSError`, and the catch-all) raise `DeepDocRemoteError` with no logging whatsoever. Verified there are currently zero callers of this module anywhere in the repo, so there is no higher-level caller today that could compensate by logging the wrapped error.

This is also inconsistent with `is_remote_configured()` at lines 126-132, which *does* log a warning on its failure path. The asymmetry produces the worse outcome of the two: an obviously-invalid URL is loud, while a *valid-looking* config pointing at an unreachable or misauthenticated server is completely silent — and a config that fails every single request while looking correct is considerably harder to notice than one that is outright rejected at startup.

Suggested fix — add a `logger.warning(...)` at each wrap site, or at minimum immediately before raising `DeepDocRemoteError`. Even a single shared helper that logs and raises would do:

```python
def _fail(message: str, exc: BaseException) -> DeepDocRemoteError:
    logger.warning("%s", message)
    return DeepDocRemoteError(message)
```

used as `raise _fail(f"Remote DeepDoc request failed: {exc}", exc) from exc`.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+            install an ``httpx.MockTransport`` instead of reaching the network.
+
+    Raises:
+        httpx.HTTPError: On any transport failure or non-2xx status.
+        ValueError: If the body is not JSON, or is JSON but not an object.
+        OSError: If a file path cannot be read.
+    """
+    base_url = get_deepdoc_xinference_url()
+    if base_url is None:
+        raise ValueError("Remote DeepDoc parsing is not configured")
```
rogercloud, line 172: **minor** — This bare `ValueError("Remote DeepDoc parsing is not configured")` gets caught by the generic `except ValueError` at line 326 and re-wrapped as:

```
DeepDocRemoteError: Remote DeepDoc returned an unusable response: Remote DeepDoc parsing is not configured
```

which is misleading — nothing was ever sent to a server, so nothing was "returned" by one. The real condition is a caller-contract violation: `parse_document_remote` was invoked without first checking `is_remote_configured()`.

Real-world impact is low today (verified: no callers yet), but the fix is cheap and it prevents a confusing log line once S4/S5 land. Suggested fix — check configuration at the top of `parse_document_remote`, before the `try`, and raise `DeepDocRemoteError` directly with an accurate message; or keep the check here but raise a distinct internal exception type that gets its own `except` clause with wording like `"Remote DeepDoc is not configured; call is_remote_configured() first"`.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+            # them. Without it a corrupt-but-right-length payload decodes to
+            # b"" and gets written out as a zero-byte PNG, so the caller
+            # receives a broken image path rather than falling back to local
+            # parsing.
+            image_bytes = base64.b64decode(image_base64, validate=True)
+            if not image_bytes:
+                raise ValueError(
+                    f"Remote DeepDoc element {index} carries an empty image"
+                )
+            normalized_element["image"] = save_image(image_bytes)
```
rogercloud, line 269: **minor** — `_normalize_elements` validates and saves images incrementally inside the loop (lines 243-272) with no cleanup on partial failure. If element *N* fails validation — missing `type`/`text` (line 251), non-object element (line 245), bad base64 (line 264), empty image (line 266) — after elements 0..N-1 have already called `save_image(...)` and written files, those earlier files are orphaned. There is no rollback, and the document then falls back to local parsing, which writes its own separate image set.

No caller wires up `save_image` yet in this PR, so nothing leaks today. But the obvious real implementation is `_save_bytes_to_disk` in `deepdoc.py`, which writes under a persistent artifacts directory rather than a temp dir — which would make these persistent, never-collected disk artifacts, on exactly the code path (malformed server response) most likely to repeat across many documents.

Suggested fix — either validate all elements first and only then save images in a second pass (simplest, and it also avoids doing I/O for a response that turns out to be junk), or track the saved paths and unlink them in an `except` wrapped around the loop.

rogercloud (commented):
## Summary

This PR is step S2 of #1161 (remote DeepDoc parsing via a Xinference-hosted server, with local parsing always available as fallback). It adds one new file, `src/xagent/providers/pdf_parser/deepdoc_remote.py` (347 lines): an HTTP client (`parse_document_remote`) that uploads a document to a proposed remote DeepDoc endpoint, normalizes the JSON response into the element shape the local parser already produces, and funnels every failure into a single `DeepDocRemoteError` meaning "fall back to local". It deliberately avoids importing `deepdoc.py` — the image-saving function is caller-injected — to keep the import graph acyclic.

Note on diff scope: the diff also touches `src/xagent/config.py` and `tests/core/test_config.py`, but those belong to the stacked config-getter PR (#1172) and are reviewed there. Only the one new file is in scope here, which is why a 3-file diff yields findings against 1 file.

## Approach verdict

**Sound.** The module implements exactly what #1163 specifies with no excess machinery: one stateless function, one exception type, one injected callable. Two integration assumptions were independently verified as correct: `DeepDocProgressAdapter` (`src/xagent/core/tools/core/RAG_tools/progress/adapters.py:56-65`) really does split on `" ("` and dedupe by status, so the `"msg (Xs)"` completion-message format is load-bearing and correctly reproduced; and `_handle_image`'s `str` branch (`src/xagent/providers/pdf_parser/deepdoc.py:49-54`) really does accept a path, so returning a saved path string from `_normalize_elements` is the right shape. Not importing `deepdoc.py` is verified correct — no such import exists.

### D1 — the docstring's test-isolation claim does not hold via the normal import path

`src/xagent/providers/pdf_parser/deepdoc_remote.py:9-12` states that injecting `save_image` "keeps the import graph acyclic and lets the client be tested without loading any parser." The first half is true and worth keeping. The second half is not: `src/xagent/providers/pdf_parser/__init__.py:24-28` does `from .deepdoc import DeepDocParser`, so `import xagent.providers.pdf_parser.deepdoc_remote` transitively executes `deepdoc.py` and pulls in onnxruntime plus the full vision stack (verified empirically — onnxruntime and roughly 25 `deepdoc.*` submodules land in `sys.modules` after that import). This matters concretely for S3 (#1164), where the unit tests land: a test author reading this docstring and expecting a fast, ONNX-free unit test will be surprised. Please either correct the claim, caveat it, or drop that clause and keep the acyclic-graph rationale.

### D2 — the catch-all comment understates what it actually catches

The `except Exception` at `src/xagent/providers/pdf_parser/deepdoc_remote.py:333-336` is justified by a comment naming only `save_image` as the reason, but the `try` block it attaches to spans `_post_document` and `_normalize_elements` in their entirety. Any unenumerated exception raised by this module's own code — including a genuine bug — is laundered into "remote server failed, fall back to local" with zero visibility.

To be clear: I am not asking for the catch to be narrowed. #1163 explicitly specifies "Wraps every failure mode… in `raise DeepDocRemoteError`… Nothing else may escape", so the broad catch is an intentional hard requirement, not an oversight, and narrowing it would violate the spec. The request is only a comment-wording fix: say that this clause catches anything not enumerated above — including defects in this module's own code — rather than implying it exists solely for `save_image`. Combined with finding #4 (logging at the wrap sites), that keeps a swallowed internal bug at least discoverable in logs.

L143: **inline** `_mime_type_for(ext)` is a one-line wrapper around a dict lookup with exactly one caller (line 182), and no other references in the repo. Replace line 182 with `mime_type = _MIME_TYPES.get(ext.lower(), _DEFAULT_MIME_TYPE)` and delete lines 143-146.

net: -4 lines possible

## Testing

- `ruff check`, `ruff format --check`, and `mypy` all pass under the repo's pinned toolchain (ruff 0.12.3 per `.pre-commit-config.yaml`). A locally-installed newer ruff (0.16.1) initially reported failures; those are false positives from expanded default rules and do not apply at the pinned version.
- Confirmed by grep that the module has no `deepdoc.py` import, and that it currently has zero callers anywhere in the repo.
- Exception-handler ordering verified correct: `json.JSONDecodeError` (line 318) and `binascii.Error` (line 322), both `ValueError` subclasses, are caught before the bare `except ValueError` at line 326; `httpx.InvalidURL` — a `ValueError`, not an `httpx.HTTPError` — correctly falls through to the catch-all.
- Resource lifecycle verified clean: `httpx.Client` closed via context manager (line 185), file handle closed via context manager (line 211), and the caller-owned `BytesIO` cursor restored via `try/finally` (lines 204-209) on both the success and exception paths.
- No unit tests exist for this file yet. That is expected and in scope for S3 (#1164), so it is not raised as a finding here.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+        raise ValueError("Remote DeepDoc parsing is not configured")
+
+    timeout_seconds = float(get_deepdoc_xinference_timeout_seconds())
+    timeout = httpx.Timeout(
+        connect=_CONNECT_TIMEOUT_SECONDS,
+        read=timeout_seconds,
+        write=timeout_seconds,
+        pool=_POOL_TIMEOUT_SECONDS,
+    )
+    data = {"zoomin": str(zoomin)}
```
rogercloud, line 181: **minor** — The request body sent here is only `{"zoomin": str(zoomin)}`, but the documented contract at line 30 lists an `image_scope` parameter (`table_figure | all | none`, default `table_figure`) and lines 59-61 explain why it matters for payload size. Verified `image_scope` appears nowhere else in the file — not as a request field, and not even as an unwired parameter on `parse_document_remote`.

Since the server does not exist yet and the block is explicitly labeled "Proposed server API contract (v1)", this cannot mislead anyone about working code today. The risk is that this docstring is the contract of record for whoever implements the server and the S4 wiring, and a reader would reasonably assume the client already sends this field.

Suggested fix — either add the parameter now (even as a plain pass-through with a TODO), or annotate line 30 to mark it as proposed-but-not-yet-sent by this client.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+``xagent.config``. In particular it must not import ``deepdoc.py``: the
+image-saving function is injected by the caller instead, which keeps the import
+graph acyclic and lets the client be tested without loading any parser.
+
+Proposed server API contract (v1)
+---------------------------------
+
+Xinference has no DeepDoc serving capability today; the endpoint below is a
+proposal tracked alongside this client, mirrored from
+``docs/deepdoc-remote-parsing.md`` section 6.
```
rogercloud, line 19: **minor** — The module docstring cites `docs/deepdoc-remote-parsing.md` as the source of the server API contract here, and again at lines 50-51 for the per-format `metadata` table. Verified that file does not exist anywhere in the repo today (`docs/` currently contains only `branch-protection.md`).

Per the umbrella issue (#1161), that document is step S6 (#1167) and lands in a separate PR (#1178) — not in any of the currently-open stacked PRs (#1170, #1172, #1173, #1175, #1176, #1177). Since this PR is designed to merge to `main` on its own with no guaranteed ordering against #1178, the reference would be genuinely dangling at merge time, and lines 50-57 point a reader at a table they cannot read.

Suggested fix — note inline that the doc lands via #1178 (e.g. "…mirrored from `docs/deepdoc-remote-parsing.md` section 6, added in #1178"), or hold the file reference until that PR merges and describe the source generically for now.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+Set ``XAGENT_DEEPDOC_XINFERENCE_URL`` to route every format ``DeepDocParser``
+supports to a GPU host instead of running local ONNX inference. Any failure
+raises :class:`DeepDocRemoteError`, which the caller treats as "fall back to
+local parsing" -- fallback is always on, which is what makes the switch
+transparent.
+
+This module deliberately depends on nothing but ``httpx`` and
+``xagent.config``. In particular it must not import ``deepdoc.py``: the
+image-saving function is injected by the caller instead, which keeps the import
+graph acyclic and lets the client be tested without loading any parser.
```
rogercloud, line 12: **nit** — Anchor for the design discussion in the review body: "lets the client be tested without loading any parser" does not hold via the normal import path, because `src/xagent/providers/pdf_parser/__init__.py:24-28` imports `deepdoc.py`, which pulls in onnxruntime and the vision stack (verified empirically). The acyclic-import-graph rationale in the same sentence is accurate and worth keeping — only the test-isolation clause needs correcting or caveating. Full reasoning in the review summary; no need to reply in two places.

OliverBryant:

PR #1174, opened 2026-08-06
fix(sandbox): isolate sandbox ownership per deployment namespace

bsbds:
## Summary

Namespace sandbox ownership per deployment so co-located deployments sharing one Docker daemon can never discover, adopt, or delete each other's sandboxes.

- `XAGENT_SANDBOX_NAMESPACE` (wired to `${COMPOSE_PROJECT_NAME}` by xagent-saas) is required when the Docker sandbox implementation is enabled; missing/invalid is a fatal startup error.
- Physical container names include the namespace in the hash (`xagent_sandbox_<safe(ns::name)>-digest`).
- Owner labels written unconditionally on every created container (incl. snapshot-derived): `xagent.managed=v2` + `xagent.sandbox.namespace` + `xagent.sandbox.name`; the value change (`true` → `v2`) keeps pre-namespace code from adopting v2 containers during rollout.
- Every find/list/cleanup filter (inspect, start/stop, delete, quiesce, idle sweep, capacity eviction, reconciliation) requires the exact namespace + v2 marker; legacy `xagent.managed=true` containers stay invisible and are inventoried at startup for manual removal.
- Snapshot image tags are namespaced (`xagent-sandbox-snapshot-<ns>:<safe-id>`).
- Namespace grammar matches Docker Compose (charset-only, no length cap), enforced at both the config getter and the service constructor.

## Verification

- `tests/core/test_config.py` (95), `tests/sandbox/test_docker_sandbox.py` (incl. real-daemon two-namespace isolation + manager cleanup/sweep scoping), `tests/sandbox/test_docker_lifecycle_api.py` (94), `tests/web/test_sandbox_manager.py` — all green.

## Related

- xagent-saas PR #472 (compose wiring + docs + submodule pointer bump)


src/xagent/sandbox/docker_sandbox.py
```diff
@@ -177,15 +185,25 @@ def _make_safe_name(name: str) -> str:
...
+    """Build the managed Docker image tag for a snapshot.
+
+    The namespace is part of the repository name so two deployments that
+    use the same snapshot id get distinct images on a shared Docker daemon:
+    a sibling deployment can neither retag, consume, nor delete this
+    deployment's snapshot image. The namespace is grammar-validated, so it
+    is safe in a repository name."""
     safe = _make_safe_name(snapshot_id)
-    return f"{SNAPSHOT_REPOSITORY}:{safe}"
+    return f"{SNAPSHOT_REPOSITORY}-{namespace}:{safe}"
```
rogercloud: **Major:** the namespace is embedded raw into a Docker repository name here, but `validate_sandbox_namespace` enforces the *Compose project-name* grammar (`^[a-z0-9][a-z0-9_-]*$`, `src/xagent/config.py:2123`), not the Docker *repository-name* grammar. Repository path components disallow trailing separators and separator runs, so `XAGENT_SANDBOX_NAMESPACE=foo-` or `a___b` passes validation and then yields `xagent-snapshot-foo-:<id>` / `...-a___b:<id>`, which the daemon rejects at commit/tag time.

Two ways out: run the namespace through `_make_safe_name` before embedding it (matching `_container_name`), or tighten `validate_sandbox_namespace` to the intersection of both grammars. Either way the docstring's claim that the namespace is "grammar-validated, so it is safe in a repository name" needs to be adjusted to whichever invariant actually holds.

docker/README.md
```diff
@@ -264,6 +264,26 @@ In Docker sibling mode, `SANDBOX_VOLUMES` sources are host-side paths. Use
...
+must use distinct Compose project names and distinct
+`XAGENT_HOST_STORAGE_ROOT` paths.
+
+Containers created before this scheme existed carry the legacy
+`xagent.managed=true` label and are ignored by current code: they are never
+listed, reclaimed, or counted against `XAGENT_SANDBOX_MAX_CONTAINERS`, and
+a backend restart recreates their sandboxes fresh (host bind mounts
+survive; container-layer state does not). The backend logs how many such
+containers exist at startup. After upgrading all co-located stacks, stop
+and drain old stacks, then remove the legacy containers manually once no
```
rogercloud: **Major:** this ordering — upgrade all stacks first, drain second — is what creates a concurrent-write window. A legacy `xagent.managed=true` container is invisible to the upgraded backend, so the backend recreates the sandbox as a v2 container mounting the *same* host bind-mount path while the legacy container is still running. Two live writers on one workspace.

Suggest inverting the procedure: drain and stop the old stacks (and their legacy sandbox containers) *before* starting the namespaced backend, and stating explicitly that a legacy container and its v2 replacement must never run concurrently because they share the host workspace mount. Worth adding the rollback direction too: rolling back to pre-namespace code once v2 containers exist orphans them and the rolled-back backend will double-provision.

src/xagent/web/sandbox_manager.py
```diff
@@ -2518,6 +2519,21 @@ def _create_boxlite_service() -> Optional[SandboxService]:
...
+        # delete another deployment's sandboxes (and vice versa). Refusing
+        # to start sandboxing is the safe failure mode.
+        raise RuntimeError(
+            "XAGENT_SANDBOX_NAMESPACE is required when the Docker sandbox "
+            "implementation is enabled; set it to a stable, unique per-"
+            "deployment identifier. Under Docker Compose use the Compose "
+            "project name (COMPOSE_PROJECT_NAME); other deployment modes "
+            "(pip/systemd) must pick their own stable value (see "
+            "docker/README.md 'Co-locating multiple deployments on one host' "
+            "and .env.example)"
```
rogercloud: Both pointers in this fatal message are wrong, which is unfortunate given it is the only thing an operator sees when their deployment refuses to start:
- `docker/README.md` has no "Co-locating multiple deployments on one host" heading — the namespace docs live under `### Sandbox Runtime Overlays`.
- There is no `.env.example` in this repo; the env template is `example.env`.

Either fix the references, or add the heading and an `XAGENT_SANDBOX_NAMESPACE` entry to `example.env` so they resolve.

docker/README.md
```diff
@@ -264,6 +264,26 @@ In Docker sibling mode, `SANDBOX_VOLUMES` sources are host-side paths. Use
 absolute host paths; relative paths and `~` are rejected instead of being
 expanded inside the backend container.
 
+The overlay also sets `XAGENT_SANDBOX_NAMESPACE` to `${COMPOSE_PROJECT_NAME}`.
+Every sandbox container a deployment creates is scoped to that namespace
```
rogercloud: Worth flagging the breaking-change angle explicitly here: `docker` is the default `SANDBOX_IMPLEMENTATION`, so every existing `SANDBOX_ENABLED=true` deployment without `XAGENT_SANDBOX_NAMESPACE` hard-fails at startup after this upgrade. The Compose overlay covers Compose users, but pip/systemd deployments get a fatal error with no required-action note. Suggest an explicit "required action on upgrade" line here and in the release notes.

src/xagent/config.py
```diff
@@ -2118,6 +2120,56 @@ def get_sandbox_host_storage_root() -> Path | None:
...
+    Enforces the Docker Compose project-name grammar (lowercase letters,
+    decimal digits, dashes and underscores, beginning with a lowercase letter
+    or digit). Callers that accept a namespace from outside the environment
+    (e.g. the Docker sandbox service constructor) must run this so a
+    malformed value can never recreate a shared ownership domain.
+
+    Raises:
+        ValueError: The value does not match the grammar.
+    """
+    if not _SANDBOX_NAMESPACE_RE.fullmatch(namespace):
```
rogercloud, line 2138: Consider normalizing before validating — `namespace = namespace.lower()` (or `.strip().lower()` in `get_sandbox_namespace`, returning the normalized value) — instead of rejecting uppercase outright. As written, an uppercase namespace raises `ValueError`, which propagates uncaught through `DockerSandboxService.__init__` and `_create_docker_service`, i.e. a fatal startup crash. Normalizing removes a whole class of "my stack won't boot" reports without weakening the ownership invariant, provided the normalized value is what is used everywhere downstream.

This is the point the `gemini-code-assist` review raised and it is still open. Keeping it minor rather than major because the usual motivation (Compose case-folding project names) is not demonstrated in this repo — the Docker sibling overlay just passes `${COMPOSE_PROJECT_NAME}` through verbatim.

src/xagent/sandbox/docker_sandbox.py
```diff
@@ -1580,7 +1638,12 @@ async def list_sandboxes(self) -> list[SandboxInfo]:
         containers = await asyncio.to_thread(
             lambda: self._client.containers.list(
                 all=True,
-                filters={"label": f"{LABEL_MANAGED}=true"},
+                filters={
+                    "label": [
+                        f"{LABEL_MANAGED}={MANAGED_LABEL_VALUE}",
+                        f"{LABEL_NAMESPACE}={self._namespace}",
```
rogercloud: The docstring above (`"""List all managed Docker sandboxes."""`) is now misleading — this filter is namespace-scoped (added here alongside the managed-label check), and its result feeds max-container accounting. `_find_container`'s docstring documents the scoping properly; mirror that wording on `list_sandboxes`: scoped to this deployment's namespace, with other deployments' and legacy pre-namespace containers invisible.

src/xagent/sandbox/docker_sandbox.py
```diff
@@ -1069,9 +1095,22 @@ class DockerSandboxService(SandboxService):
...
+        Args:
+            store: Storage for persisting sandbox metadata.
+            namespace: Stable per-deployment namespace (e.g. the Docker
+                Compose project name). Every container this service creates
+                is namespaced by it, and every lookup/list operation is
+                restricted to it, so multiple deployments sharing one Docker
+                daemon can never discover or mutate each other's sandboxes.
+            client: Docker SDK client override (tests).
+        """
+        self._namespace = namespace
```
rogercloud: Nit: `self._namespace = namespace` runs before `validate_sandbox_namespace(namespace)` on the next line. Harmless today since the raise aborts construction, but swapping the two lines makes the "validated before stored" invariant self-evident and stays correct if code is later added between them.

src/xagent/web/sandbox_manager.py
```diff
@@ -2530,8 +2546,19 @@ def _create_docker_service() -> Optional[SandboxService]:
 
     service = None
     try:
-        service = DockerSandboxService(store=store)
-        logger.info("Created Docker sandbox service")
+        service = DockerSandboxService(store=store, namespace=namespace)
+        logger.info("Created Docker sandbox service (namespace=%s)", namespace)
+        legacy_count = service.count_legacy_containers()
```
rogercloud: Nit / pre-existing pattern: `count_legacy_containers()` is a blocking Docker list call on the event loop inside the async lifespan. Not a new pattern here — the constructor's `self._client.ping()` is already sync — and it is one call at startup, so low priority. Worth an `asyncio.to_thread` wrap if this file gets a cleanup pass.

tests/core/test_config.py
```diff
@@ -1487,6 +1492,51 @@ def test_storage_root_expands_env_vars_without_user_or_abspath(self, monkeypatch
...
+    def test_invalid_namespace_raises(self, monkeypatch, value):
+        """Test that malformed namespaces raise instead of silently degrading."""
+        monkeypatch.setenv(SANDBOX_NAMESPACE, value)
+        with pytest.raises(ValueError, match="Invalid sandbox namespace"):
+            get_sandbox_namespace()
+
+    def test_long_namespace_accepted_like_compose(self, monkeypatch):
+        """Compose accepts arbitrary-length project names; so must we.
+
+        The namespace only feeds hashed container names and label values, so
```
rogercloud: This rationale is now stale: the namespace does **not** "only feed hashed container names and label values" — `_snapshot_tag` feeds it raw into a Docker repository name, which does carry backend grammar constraints (see the inline comment on `docker_sandbox.py:206`). Update the docstring (or the test) once that invariant is settled.

tests/core/test_config.py
```diff
@@ -1487,6 +1492,51 @@ def test_storage_root_expands_env_vars_without_user_or_abspath(self, monkeypatch
...
+
+    def test_valid_namespace(self, monkeypatch):
+        """Test that a Compose-project-name-shaped namespace is accepted."""
+        monkeypatch.setenv(SANDBOX_NAMESPACE, "my-project-1")
+        result = get_sandbox_namespace()
+        assert result == "my-project-1"
+
+    @pytest.mark.parametrize(
+        "value",
+        [
```
rogercloud, line 1519: This parametrization misses the cases that matter for label/tag injection. Suggest adding `":"`, `"a::b"`, `"a/b"`, `"a.b"`, a non-ASCII value, and `""` passed directly to `validate_sandbox_namespace` (the empty case is currently only reached via `get_sandbox_namespace`, which short-circuits to `None` before validating).

Note the "trailing dash/underscore" case is *not* a gap — `^[a-z0-9][a-z0-9_-]*$` intentionally accepts those; that permissiveness is instead the root of the snapshot-tag issue flagged on `docker_sandbox.py:206`.

tests/web/test_sandbox_manager.py
```diff
@@ -237,6 +244,33 @@ def test_creation_failure_returns_none(self):
...
+        """Missing namespace must fail loudly, never degrade to unscoped mode."""
+        monkeypatch.delenv("XAGENT_SANDBOX_NAMESPACE", raising=False)
+        with pytest.raises(RuntimeError, match="XAGENT_SANDBOX_NAMESPACE is required"):
+            _create_docker_service()
+
+    def test_legacy_container_inventory_is_logged(self, caplog):
+        """Startup must surface leftover legacy containers for manual removal."""
+
+        class _ServiceWithLegacyCount(FakeSandboxService):
+            def count_legacy_containers(self) -> int:
```
rogercloud: This stub means the real `count_legacy_containers()` body is never executed by any test — neither the `label=xagent.managed=true` filter value nor the swallow-and-return-0 failure path. A small fake-client test asserting the filter and the exception path would close it.

rogercloud (commented):
## Summary

This PR makes Docker sandbox ownership per-deployment instead of per-daemon: container names are hashed from `{namespace}::{name}`, owner labels move from `xagent.managed=true` to `xagent.managed=v2` + `xagent.sandbox.namespace=<ns>`, snapshot image repositories are suffixed with the namespace, and every lookup/list path filters on both labels. The namespace comes from a new required `XAGENT_SANDBOX_NAMESPACE` (wired to `${COMPOSE_PROJECT_NAME}` in the Docker sibling overlay), validated against the Compose project-name grammar, and a missing value is a fatal startup error rather than a silent degrade to shared ownership. Legacy `xagent.managed=true` containers are deliberately invisible to all namespaced operations, with a startup warning that counts them so operators can drain them manually.

## Approach

Sound. Scoping ownership with a label + name namespace is the right-sized fix for this class of bug — a dedicated daemon per deployment or network-level isolation would be heavier operationally and still would not fix the physical-name and discovery-filter collision, which is the actual defect. The value bump (`true` -> `v2`) rather than a new key alone is the correct detail: it makes a pre-namespace backend unable to see the new containers, so the migration is safe in the "old backend still running" direction. Fatal-on-missing-namespace is also the right failure mode. The remaining issues are around the edges of that design: the snapshot tag's grammar assumption, and the documented upgrade procedure.

## Design-level findings

**D1 — Namespace filter duplicated across the two discovery paths (maintainability).**
The owner-label pair (`xagent.managed=v2`, `xagent.sandbox.namespace=<ns>`) is constructed independently in `_find_container` and in `list_sandboxes`. Both are correct today, but the ownership predicate is the security boundary of this PR, and it now has two sources of truth. A single helper would make any future third call site correct by construction. Not blocking.

**D2 — Upgrade window allows two live writers on one host bind mount (real hazard).**
Because a legacy v1 container is invisible to the new code, a restarted backend recreates the sandbox fresh — but the recreated v2 container mounts the *same* host bind-mount path as the still-running v1 container. Nothing in the code or the docs prevents both from being alive at once, so during upgrade a single workspace can have two concurrent writers. The code side of this is arguably unfixable without cross-scheme discovery (which the design intentionally forbids); the fix belongs in the documented procedure — see the inline comment on `docker/README.md`.

**D3 — No legacy snapshot-image migration (latent, not live).**
Pre-namespace snapshot images live at the un-suffixed repository and are unreachable by the new namespaced tag builder. There is no migration or retag step. This is latent rather than live because no production caller currently reaches Docker snapshot create/delete, so nothing regresses today — but it should be tracked before snapshots are exposed. Related: `delete_snapshot` (`src/xagent/sandbox/docker_sandbox.py:1727`, pre-existing/untouched by this PR) would remove an un-namespaced legacy snapshot image that sibling deployments could also reference — not introduced by this PR, but directly impacted by it, so it belongs in the same tracked-debt item as this migration gap.

**D4 — Rollback asymmetry is undocumented.**
The forward direction is safe (old backend cannot see v2 containers). The reverse is not: rolling back to pre-namespace code once v2 containers exist leaves them orphaned and un-reclaimable by the rolled-back backend, which will then double-provision. Worth one sentence in `docker/README.md` next to the upgrade steps.

**D5 — MCP `docker_manager.py` has the same collision class and is untouched (explicitly out of scope).**
Noting it only so it does not get lost: the MCP Docker manager still owns containers daemon-globally. Not a blocker for this PR; suggest a follow-up issue.

## Verified with no issues

- Separator-collision safety of `_container_name` — the `{namespace}::{name}` join is hashed, and namespaces cannot contain `:`, so no two distinct (namespace, name) pairs collide.
- Call-site coverage of the namespace filter — every Docker discovery/list path applies both owner labels; no unfiltered `containers.list`/`containers.get` remains on a managed path.
- `sandbox_manager.py` has no direct Docker imports — the abstraction boundary holds.
- Label overwriting on snapshot-derived containers — containers created from a snapshot image get the current deployment's owner labels, so a snapshot cannot smuggle another namespace's ownership.
- v1 -> v2 label value bump — bumping the value (not just adding a key) is what makes new containers invisible to a pre-namespace backend; correct and load-bearing.
- Compose `COMPOSE_PROJECT_NAME` wiring — verified against Compose 5.3.1 and 5.1.4 behavior; the overlay wiring is correct.
- Fatality and gating of the startup check — the missing-namespace path raises rather than degrading, and it is gated to the Docker implementation only.
- Test-file changes in `test_bridge_seams.py`, `test_docker_lifecycle_api.py`, `test_docker_lock_infra.py`, `test_sandbox_reconcile_e2e.py` are mechanical only — constructor-signature propagation, no assertion weakening.
- CI actually runs the real-daemon two-namespace test — not silently skipped in the configured CI job.

Lean already. (One candidate was considered — collapsing the `repo:tag` rebuild/resplit inside `create_snapshot` — and dropped: single caller, no net line reduction.)

## Re-review status

This PR already has one prior review, from the `gemini-code-assist` bot (not a human) at commit `354ed6c`, one commit before current HEAD. Its only substantive point — normalize the namespace to lowercase before validating, so a case-insensitive Compose project name cannot crash startup — is **not fixed**: `src/xagent/config.py:2123` still uses `^[a-z0-9][a-z0-9_-]*$` with no `.lower()` and no `re.IGNORECASE`. Carried forward below as a minor inline suggestion at reduced severity, since the case-insensitive-Compose premise behind it isn't demonstrated anywhere in this repo.

Tests on the review checkout: 531 passed, 55 skipped (all skips are Docker-daemon-gated, no local daemon available); mypy clean on the changed source files.

src/xagent/config.py
```diff
@@ -2118,6 +2120,56 @@ def get_sandbox_host_storage_root() -> Path | None:
...
+        The configured namespace, or None when unset/empty.
+
+    Raises:
+        ValueError: The variable is set but does not match the grammar.
+    """
+    raw = os.getenv(SANDBOX_NAMESPACE, "").strip()
+    if not raw:
+        return None
+    validate_sandbox_namespace(raw)
+    return raw
```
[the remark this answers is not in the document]
bsbds: We are keeping strict Compose-grammar validation. Compose 5.1.4 normalizes directory-derived/top-level names and rejects mixed-case explicit COMPOSE_PROJECT_NAME/-p values before interpolation; silently lowercasing non-Compose input would hide invalid configuration.

tests/core/test_config.py
```diff
@@ -1487,6 +1492,51 @@ def test_storage_root_expands_env_vars_without_user_or_abspath(self, monkeypatch
...
+            "Upper-Case",
+            "-leading-dash",
+            "has space",
+        ],
+    )
+    def test_invalid_namespace_raises(self, monkeypatch, value):
+        """Test that malformed namespaces raise instead of silently degrading."""
+        monkeypatch.setenv(SANDBOX_NAMESPACE, value)
+        with pytest.raises(ValueError, match="Invalid sandbox namespace"):
+            get_sandbox_namespace()
```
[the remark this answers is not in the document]
bsbds: Not changing this test because uppercase normalization is not part of the accepted contract. Compose rejects explicit mixed-case project names, and the test correctly proves the fail-fast validator behavior.

src/xagent/sandbox/docker_sandbox.py
bsbds: This is already addressed in acabd72a: snapshot repositories now sanitize separators, bound length, and add a digest when needed; grammar/collision tests plus a real-daemon trailing-dash snapshot round-trip pass.

src/xagent/config.py
bsbds: We are not adding normalization: supported Compose inputs arrive lowercase or are rejected by Compose, while non-Compose deployments must provide the documented Compose-shaped stable identifier. Fail-fast validation avoids silently aliasing operator inputs.

src/xagent/web/sandbox_manager.py
bsbds: Leaving this startup-only call synchronous. The factory and DockerSandboxService constructor are synchronous and constructor ping already blocks; propagating an async API solely for the following best-effort list would add disproportionate complexity without removing the existing blocking path.

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/sandbox-namespace-isolation`
- **Head:** `216419013edd259a3e238694d737946ff2320cef`
- **Destination:** `git@github.com:bsbds/xagent.git` → `fix/sandbox-namespace-isolation` (head of xorbitsai/xagent#1174)
- **Result:** success

### Resolved

1. **Thread:** `PRRT_kwDORQ4Kr86W_tVX` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062323
   **Local outcome:** resolved
   **Commit:** `c39c3fab251c77b631b8d588f18cb909bc629b83`
   **GitHub:** resolved

2. **Thread:** `PRRT_kwDORQ4Kr86W_tVa` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062327
   **Local outcome:** resolved
   **Commit:** `764e64b2aac933ea7639abdc7de5537e1528acc6`
   **GitHub:** resolved

3. **Thread:** `PRRT_kwDORQ4Kr86W_tVd` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062331
   **Local outcome:** resolved
   **Commit:** `73b6a19e52ffafdcd2843c7cd1b1c03488d40116`
   **GitHub:** resolved

4. **Thread:** `PRRT_kwDORQ4Kr86W_tVn` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062344
   **Local outcome:** resolved
   **Commit:** `2617566c0b293893e421d0277f0f229bf5a1c4b3`
   **GitHub:** resolved

5. **Thread:** `PRRT_kwDORQ4Kr86W_tVs` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062352
   **Local outcome:** resolved
   **Commit:** `7bc60c62d27d1164eac4afb8bd599220022a867f`
   **GitHub:** resolved

6. **Thread:** `PRRT_kwDORQ4Kr86W_tV0` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062362
   **Local outcome:** resolved
   **Commit:** `2f50d05f269d3b800894f793abfe22cb7ef65175`
   **GitHub:** resolved

7. **Thread:** `PRRT_kwDORQ4Kr86W_tV5` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062373
   **Local outcome:** resolved
   **Commit:** `df170007021c653d505f222587ebfbb419ca962d`
   **GitHub:** resolved

8. **Thread:** `PRRT_kwDORQ4Kr86W_tV-` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062379
   **Local outcome:** resolved
   **Commit:** `216419013edd259a3e238694d737946ff2320cef`
   **GitHub:** resolved

### Rejected

1. **Thread:** `PRRT_kwDORQ4Kr86W7p_R` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3727519073
   **Local outcome:** rejected
   **Classification:** unsound
   **Reason:** Compose normalizes directory/top-level names and rejects mixed-case explicit project names; non-Compose input should fail the documented grammar rather than be silently aliased.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733147655
   **GitHub:** left unresolved

2. **Thread:** `PRRT_kwDORQ4Kr86W7p_V` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3727519078
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** Depends on the rejected uppercase-normalization behavior; the current test correctly proves fail-fast validation.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733147781
   **GitHub:** left unresolved

3. **Thread:** `PRRT_kwDORQ4Kr86W_tVS` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062313
   **Local outcome:** rejected
   **Classification:** outdated
   **Reason:** Already addressed and verified in `acabd72a`; GitHub marks the original code location outdated.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733147917
   **GitHub:** externally changed outdated state preserved; left unresolved

4. **Thread:** `PRRT_kwDORQ4Kr86W_tVg` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062337
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** Same rejected uppercase-normalization request as `PRRT_kwDORQ4Kr86W7p_R`.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733148020
   **GitHub:** left unresolved

5. **Thread:** `PRRT_kwDORQ4Kr86W_tVv` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3729062358
   **Local outcome:** rejected
   **Reason:** The synchronous factory already performs a blocking Docker ping; propagating an async API solely for one startup-only inventory call is disproportionate and does not remove the existing blocking path.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733148127
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 8
- **Rejected:** 5
- **Unresolved:** 0

src/xagent/web/sandbox_manager.py
```diff
@@ -2518,6 +2519,20 @@ def _create_boxlite_service() -> Optional[SandboxService]:
 
 def _create_docker_service() -> Optional[SandboxService]:
     """Create Docker sandbox service."""
+    namespace = get_sandbox_namespace()
+    if namespace is None:
+        # Fatal, not a degraded start: without a stable per-deployment
+        # namespace, a backend on a shared Docker daemon can discover and
+        # delete another deployment's sandboxes (and vice versa). Refusing
+        # to start sandboxing is the safe failure mode.
+        raise RuntimeError(
```
rogercloud, line 2528: **Major:** this fatal raise is reachable from the Celery worker process, where it converts a pre-PR graceful degrade into an unhandled error.

Traced chain: the Beat task `scan_due_triggers` (`src/xagent/web/jobs/trigger_tasks.py:115`, running in the `worker` process on the `triggers` queue) -> `_reap_and_pause_stale_preview_runs` -> `pause_workforce_tasks_after_archive` (in-process asyncio, not re-enqueued) -> `dispatch_task_command_promptly` -> `_handle_pause_task_unserialized` -> `get_agent_for_task` -> `_get_or_create_task_sandbox` in `chat.py` -> `get_sandbox_manager()` -> here. Before this PR, `_create_docker_service` caught all exceptions and returned `None`; now the raise happens before entering that try/except when `XAGENT_SANDBOX_NAMESPACE` is unset.

The two plausible escape hatches don't apply: the agent cache is always empty in a fresh worker process, so the first dispatched pause is always a cache miss; and the `task_has_live_foreign_runner` deferral can't catch it either, because the reap staleness threshold (>=300s) always exceeds the lease TTL (default 60s) - any run old enough to be reaped already has an expired lease. The exception isn't literally silent (it surfaces as `logger.warning(..., exc_info=True)` in `src/xagent/web/services/workforce_runtime.py:531`), but there's no alert or retry, so stale preview-run pausing stops working with only a log line as evidence.

Scope, which matters for urgency: this only fires when an operator manually sets `SANDBOX_ENABLED=true` in the shared `.env` (loaded by `worker`/`scheduler` via `env_file:`) without also giving those services a namespace. The documented Compose overlay `docker/docker-compose.sandbox.docker.yml` patches only the `backend` service, so a deployment following the documented setup exactly is unaffected. It's a trap for a plausible non-default configuration, not a broken default.

Suggested fix: wire `XAGENT_SANDBOX_NAMESPACE` (and `SANDBOX_ENABLED`) into `worker:`/`scheduler:` in the overlay so the namespace is uniform per stack, and/or wrap the `get_sandbox_manager()` call path reachable from the worker so a missing-namespace `RuntimeError` degrades the way the pre-PR code did outside the startup path.

docker/docker-compose.sandbox.docker.yml
```diff
@@ -9,6 +9,7 @@ services:
       - SANDBOX_MEMORY=512
       - XAGENT_SANDBOX_HOST_PROJECT_ROOT=${XAGENT_SANDBOX_HOST_PROJECT_ROOT:-${PWD}}
       - XAGENT_SANDBOX_HOST_STORAGE_ROOT=${XAGENT_HOST_STORAGE_ROOT:-/root/.xagent}
+      - XAGENT_SANDBOX_NAMESPACE=${COMPOSE_PROJECT_NAME}
```
rogercloud: **Major:** two issues with `${COMPOSE_PROJECT_NAME}` here.

First, there's no `:?` or `:-` guard. With `COMPOSE_PROJECT_NAME` unset, this resolves today via Compose's implicit directory-derived project name (confirmed with `docker compose config` on Compose 5.3.1), but that's an implicit behavior and the repo documents no minimum Compose version anywhere. On an older Compose where it resolves empty, the failure moves from `docker compose config` time to a backend crash at runtime.

Second, this `environment:` entry silently overrides a user's `.env`-set `XAGENT_SANDBOX_NAMESPACE` for `backend`. Verified: with `XAGENT_SANDBOX_NAMESPACE=myns_from_env` in `.env` - exactly what `example.env` and the upgrade instructions in `docker/README.md` tell a manually-configuring operator to do - `backend` still resolves to `${COMPOSE_PROJECT_NAME}` because `environment:` wins over `env_file`, while `worker`/`scheduler` (untouched by this overlay) get the `.env` value. That's a real namespace split within one stack. It has no runtime effect in the stock documented config today since `worker`/`scheduler` don't consume the sandbox namespace, but it becomes load-bearing in the `SANDBOX_ENABLED=true`-in-`.env` scenario noted separately on `sandbox_manager.py`.

Suggested fix: use `${COMPOSE_PROJECT_NAME:?set COMPOSE_PROJECT_NAME to a stable unique value}` so a missing value fails at `docker compose config` time with an actionable message, and either document that the Docker sibling overlay derives `backend`'s namespace from the Compose project name regardless of `.env`, or stop hardcoding it in `environment:` so `.env` can actually override it.

src/xagent/sandbox/docker_sandbox.py
```diff
@@ -177,15 +193,50 @@ def _make_safe_name(name: str) -> str:
...
+
+
+#: Max length of the namespace token embedded in the snapshot repository name.
+#: The full Docker reference (repository:tag) must stay under 255 characters:
+#: the fixed prefix (24 chars) plus the token plus the tag (<= 128 chars) must
+#: fit, so the token is capped well below that.
+_SNAPSHOT_NAMESPACE_TOKEN_MAX = 100
+
+
+def _snapshot_repository(namespace: str) -> str:
```
rogercloud, line 220: **Minor:** this function still has a reproducible collision where two distinct valid namespaces map to the same snapshot repository name. (Distinct from the grammar mismatch fixed earlier in this function - that repro, trailing/repeated separators producing invalid Docker repo names, is confirmed fixed.)

The digest is appended only *conditionally* - when the sanitized token differs from the input or exceeds the length cap. So a namespace that already looks like "sanitized token + digest" skips digesting entirely and collides with the genuine digested output of the token it resembles. Reproduced directly:

- `_snapshot_repository('a_b')` -> `xagent-sandbox-snapshot-a-b-eb2980a5a2`
- `_snapshot_repository('a-b-eb2980a5a2')` -> `xagent-sandbox-snapshot-a-b-eb2980a5a2`

Both `'a_b'` and `'a-b-eb2980a5a2'` independently pass `validate_sandbox_namespace`. Two co-located deployments named this way would share snapshot images and could overwrite or delete each other's - a hole in the exact isolation boundary this PR closes.

The existing test at `tests/sandbox/test_docker_sandbox.py:315-323` only checks a non-colliding pair (`a-b` vs `a_-b`) plus reflexive equality, so it gives false confidence rather than covering this case.

Suggested fix: append the digest unconditionally, so a collision requires an actual SHA-1 collision - or use a separator/format that the sanitized token can provably never produce on its own. Either way, extend the test with a colliding-shape pair.

src/xagent/web/sandbox_manager.py
```diff
@@ -2530,8 +2545,19 @@ def _create_docker_service() -> Optional[SandboxService]:
 
     service = None
```
rogercloud: **Minor:** an exception from `count_legacy_containers()` produces a misleading "Failed to create Docker sandbox service" error log even though the service was constructed successfully and works fine.

`service` is bound before `count_legacy_containers()` is called, and both sit inside the same `except Exception` handler. The count call is not declared on the `SandboxService` ABC, so any duck-typed or patched implementation can raise `AttributeError` here - at which point the handler logs a false "Failed to create" error and then still returns the perfectly working `service`. That's an actively misleading signal for anyone debugging a startup issue.

Suggested fix: move the inventory call outside the try with its own handler, and/or declare `count_legacy_containers()` on the `SandboxService` ABC so the `AttributeError` case can't arise.

src/xagent/sandbox/docker_sandbox.py
```diff
@@ -177,15 +193,50 @@ def _make_safe_name(name: str) -> str:
     return f"{base.lower()}-{digest}"
 
 
-def _container_name(name: str) -> str:
-    """Build the managed Docker container name for a sandbox."""
-    return f"{CONTAINER_NAME_PREFIX}{_make_safe_name(name)}"
+def _container_name(name: str, namespace: str) -> str:
```
rogercloud, line 200: **Minor:** `_make_safe_name`/`_container_name` embed the sanitized namespace and name verbatim with no length cap, which contradicts a test docstring.

Measured: `_container_name('u', 'a' * 240)` produces a 268-character name. This PR explicitly capped the analogous `_snapshot_repository` at 100 characters for exactly this class of problem, but left the container-name path uncapped. Meanwhile the docstring at `tests/core/test_config.py:1541-1548` asserts that "container identities hash the full namespace ... so backend identifiers do not impose a namespace length cap" - a claim that isn't backed by any assertion on `_container_name`'s actual output and is contradicted by the verbatim embedding.

To be fair on severity: no hard daemon-level length limit was found in Moby's source, so real-world breakage isn't proven. But the asymmetry with `_snapshot_repository` is real, and the docstring currently documents behavior the code doesn't have.

Suggested fix: bound `_container_name` the same way as `_snapshot_repository`, add a test pinning the actual output length at long namespaces, or correct the docstring's claim.

docker/README.md
```diff
@@ -264,6 +264,48 @@ In Docker sibling mode, `SANDBOX_VOLUMES` sources are host-side paths. Use
...
+(physical name and owner labels), so multiple deployments sharing one Docker
+daemon never discover or manage each other's sandboxes. Co-located stacks
+must use distinct Compose project names and distinct
+`XAGENT_HOST_STORAGE_ROOT` paths.
+
+Containers created before this scheme existed carry the legacy
+`xagent.managed=true` label and are ignored by current code: they are never
+listed, reclaimed, or counted against `XAGENT_SANDBOX_MAX_CONTAINERS`, and
```
rogercloud, line 290: **Minor:** `XAGENT_SANDBOX_MAX_CONTAINERS` is now effectively a *per-namespace* budget, and that change isn't documented here.

Capacity accounting is always scoped to the current namespace through `_ownership_label_filters`, so N co-located namespaced deployments sharing one Docker daemon each get their own independent container budget - total daemon load scales with the number of co-located stacks rather than being bounded by a single global cap. This section already documents that legacy containers are excluded from the count, which makes the omission of the per-deployment multiplication more noticeable.

Suggested fix: state explicitly that the limit applies per deployment namespace, and note that operators sizing a shared daemon should budget `XAGENT_SANDBOX_MAX_CONTAINERS x number of co-located stacks`.

rogercloud (commented):
## Summary

This PR makes sandbox ownership explicit per deployment by requiring an `XAGENT_SANDBOX_NAMESPACE` value that is threaded into Docker container names, ownership labels, and snapshot image tags, so two xagent stacks sharing a Docker daemon can no longer discover, adopt, or delete each other's sandbox containers. The `xagent.managed` label value moves from `true` to `v2` so pre-namespace code cannot adopt new containers, a missing or invalid namespace is now a fatal startup error rather than a silent fallback, and legacy `xagent.managed=true` containers are made invisible to all namespaced operations and merely counted at startup for manual operator cleanup.

## Approach

**acceptable-with-reservations.** Per-deployment namespacing as the ownership boundary is the right-sized fix, and it is threaded consistently: a single ownership-filter helper (`_ownership_label_filters`), a single service construction site, and one validation grammar reused at both the config getter and the service constructor. The label-value bump and fatal-on-missing-namespace are the correct companions to that choice. The reservations are about rollout safety and deployment wiring rather than the scoping logic itself.

Two design-level points are worth carrying forward, neither blocking:

The legacy-container rollout path only counts and warns. `count_legacy_containers()` queries with `all=True` and no `status=running` filter, and the caller only emits a `logger.warning` with no escalation - so a stopped legacy container (pure housekeeping) and a *running* legacy container (an active dual-writer against the same daemon) are reported identically. The "never run legacy and v2 concurrently" invariant that `docker/README.md` now correctly documents is therefore enforced entirely by an operator reading and following prose; nothing in the code detects or blocks the genuinely dangerous case. Consider reporting running legacy containers distinctly and escalating that case harder - an error, or requiring an explicit override flag.

There is no migration path for pre-namespace snapshot images, and unlike the container-level legacy gap this is not documented as tracked debt. Snapshot images at the old un-suffixed repository are unreachable by the new namespaced tag builder, and no retag or migration step exists. This is latent rather than live, since no current production caller reaches Docker snapshot create/delete - but `docker/README.md`'s legacy-migration section covers only containers and never mentions snapshots at all. Worth an issue or at minimum a README note before snapshots are exposed to any caller.

## Findings

### Major

1. `src/xagent/web/sandbox_manager.py:2528` (`_create_docker_service` fatal raise) - Celery worker path turns a graceful degrade into an unhandled `RuntimeError`. The Celery Beat task `scan_due_triggers` (`src/xagent/web/jobs/trigger_tasks.py:115`) reaches `get_sandbox_manager()` -> `_create_docker_service()` in-process, which now raises when the namespace is unset; pre-PR this path caught all exceptions and returned `None`. Fires only when an operator sets `SANDBOX_ENABLED=true` in the shared `.env` without giving `worker`/`scheduler` a namespace. See inline comment.

2. `docker/docker-compose.sandbox.docker.yml:12` - unguarded `${COMPOSE_PROJECT_NAME}`, and the overlay silently overrides a user's `.env` namespace for `backend`. Works on current Compose via the implicit directory-derived project name, but the repo documents no minimum Compose version, and the `environment:` entry beats `env_file`, producing a verified namespace split between `backend` and `worker`/`scheduler`. See inline comment.

### Minor

3. `src/xagent/sandbox/docker_sandbox.py:212` (`_snapshot_repository`) - reproducible namespace collision. The digest is appended only conditionally, so a namespace already shaped like "token + digest" collides with the real digested output of the token it resembles. See inline comment.

4. `src/xagent/web/sandbox_manager.py:2546` - a failure in `count_legacy_containers()` logs a misleading "Failed to create Docker sandbox service" for a service that was constructed successfully. See inline comment.

5. `src/xagent/web/sandbox_manager.py:2576` (`get_sandbox_manager`, not itself modified by this diff so noted here rather than inline) - leaves `_sandbox_manager_initialized` at `False` when `_create_docker_service()` raises, so the failure is never cached and every subsequent call re-raises instead of returning a memoized result. Harmless on the intended startup-crash path (the process exits), but matters for any caller reaching `get_sandbox_manager()` outside startup - e.g. the Celery worker path in finding 1, or a tool factory elsewhere that catches the exception and downgrades it to a warning - where the "fatal, never silently degrade" guarantee this PR is built around ends up silently downgraded by the caller on every call. Suggested fix: set the initialized flag (and cache the failure, or a sentinel) before re-raising, so repeat callers get a consistent, cheap, and observable outcome.

6. `src/xagent/sandbox/docker_sandbox.py:196` (`_make_safe_name`/`_container_name`) - embed the namespace verbatim with no length cap, contradicting a test docstring in `tests/core/test_config.py` that claims container identities impose no namespace length cap. See inline comment.

7. `docker/README.md:283` - `XAGENT_SANDBOX_MAX_CONTAINERS` is now a per-namespace budget and this is undocumented. N co-located deployments each get an independent budget, so total daemon load scales with the number of stacks. See inline comment.

## Since the last review

All eleven line-level findings from the previous round and design findings D1 and D4 are fixed and verified against the current code: the snapshot repository Docker-grammar mismatch (trailing/repeated separators) now sanitizes, caps, and digests; the upgrade procedure documents draining old stacks before starting the namespaced backend and warns against running legacy and v2 concurrently; the rollback-orphan hazard is documented immediately after the upgrade steps; the fatal startup error message's doc pointers are corrected; the required-action-on-upgrade note for non-Compose deployments is now an explicit blockquote; `list_sandboxes`' docstring documents namespace scoping; `DockerSandboxService.__init__` validates before storing; the stale test docstring about hashed-only namespace use is corrected; namespace validation parametrization now covers `:`, `a::b`, `a/b`, `a.b`, non-ASCII, and empty; `count_legacy_containers()`'s real body including the exception-swallow path is directly exercised; and the duplicated ownership-filter construction is consolidated into `_ownership_label_filters`.

Two suggestions were explicitly declined by the author with sound rationale and are treated as resolved by design - not relitigated here: namespace case-normalization (also raised by the review bot), declined because Compose itself rejects mixed-case explicit project names and silent lowercasing would mask invalid operator configuration; and making `count_legacy_containers()` non-blocking on the startup event loop, declined as a one-time startup cost behind a constructor that already blocks on ping.

Lean already. The one candidate considered - deleting `count_legacy_containers()` as redundant with the `docker ps` command the README already documents - was dropped: it earns its keep as a persistent, automatic, cross-restart reminder that a one-off manual audit command does not replace, and the implementation is already minimal and best-effort.

## Testing

Namespace validation tests in `tests/core/test_config.py`, and the legacy-container tests across `tests/sandbox/test_docker_sandbox.py` and `tests/web/test_sandbox_manager.py`, were run and pass. A full pass over `tests/sandbox/`, `tests/web/test_sandbox_manager.py`, and `tests/core/test_config.py::TestGetSandboxNamespace` was green, and `mypy` on the three changed source files is clean. Docker-daemon-gated real-container tests were skipped in the review environment and were not independently confirmed green here; CI is expected to cover them.

src/xagent/sandbox/docker_sandbox.py
bsbds: Addressed in f3361686: every snapshot repository mapping now appends the raw namespace digest, including already-safe names. The reported a_b / a-b-eb2980a5a2 pair maps differently, with regression, grammar, and length tests passing.

src/xagent/sandbox/docker_sandbox.py
bsbds: No container-name cap is being added because Moby has no demonstrated name-length limit. The existing test rationale remains accurate: the full namespace participates in the identity digest and no backend constraint currently imposes a namespace length cap; truncation would rename containers without a proven requirement.

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/sandbox-namespace-isolation`
- **Head:** `c4dc713fd4d48c5292196d00db1d40b821b54d1f`
- **Destination:** `git@github.com:bsbds/xagent.git` → `fix/sandbox-namespace-isolation` (head of xorbitsai/xagent#1174)
- **Result:** success

### Resolved

1. **Thread:** `PRRT_kwDORQ4Kr86XLbZ7` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733552610
   **Local outcome:** resolved
   **Commit:** `421aa68ed577566816c7df35e64f2033c8dee2a5`
   **GitHub:** resolved

2. **Thread:** `PRRT_kwDORQ4Kr86XLbZ-` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733552612
   **Local outcome:** resolved
   **Commit:** `771c65b7ecf45f49270fe5db30d171f7d302c986`
   **GitHub:** resolved

3. **Thread:** `PRRT_kwDORQ4Kr86XLbaC` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733552619
   **Local outcome:** resolved
   **Commit:** `8aff252fdd085e69613a623d1f33f04d7f864c96`
   **GitHub:** resolved

4. **Thread:** `PRRT_kwDORQ4Kr86XLbaG` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733552628
   **Local outcome:** resolved
   **Commit:** `6e96a841d515d79ccf67a3d549c33bbf0394625f`
   **GitHub:** resolved

### Rejected

1. **Thread:** `PRRT_kwDORQ4Kr86XLbaB` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733552616
   **Local outcome:** rejected
   **Classification:** outdated
   **Reason:** Already fixed and verified in `f3361686bd58d562a481ed31fb54c7fd90a9845c`; every mapping now appends the raw namespace digest and the reported adversarial pair maps differently.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733757933
   **GitHub:** left unresolved

2. **Thread:** `PRRT_kwDORQ4Kr86XLbaD` — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733552624
   **Local outcome:** rejected
   **Classification:** unsound
   **Reason:** Moby has no demonstrated container-name length limit; the current identity hash and no-namespace-cap rationale remain accurate, while truncation would rename containers without a proven requirement.
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1174#discussion_r3733758046
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 4
- **Rejected:** 2
- **Unresolved:** 0

rogercloud (approved):
## Summary

This PR makes sandbox ownership explicit per deployment by introducing a required `XAGENT_SANDBOX_NAMESPACE` setting, so multiple xagent stacks sharing a single Docker daemon can no longer discover, adopt, or delete each other's sandbox containers. The namespace is threaded through container names (hashed), owner labels (`xagent.managed` bumped `true` → `v2`, plus a new `xagent.sandbox.namespace` label), snapshot image tags, and every discovery/list/cleanup filter. Pre-namespace containers labelled `xagent.managed=true` are deliberately made invisible to the new code path and only counted at startup rather than migrated, with an operator upgrade procedure documented in `docker/README.md`.

## Approach verdict: sound

The ownership predicate lives in exactly one place — `_ownership_label_filters` (`src/xagent/sandbox/docker_sandbox.py:89-94`) — and every discovery site (`_find_container`, `list_sandboxes`) goes through it, so there is no second, drifting definition of "mine". Labels are written unconditionally on every created container including snapshot-derived ones, which closes the label-inheritance hole. The `true` → `v2` value bump is the right migration primitive: new containers are invisible to pre-namespace backends, making the rollout safely one-way. `_snapshot_repository`'s sanitize + length-cap + unconditional-digest scheme is genuinely necessary given that Docker Compose's project-name grammar is more permissive than Docker's repository-name grammar, and it closes a real collision class reported in an earlier round.

## Design-level notes (non-blocking)

**D1 — legacy container counting doesn't distinguish running from stopped.** `count_legacy_containers()` (`src/xagent/sandbox/docker_sandbox.py:1710-1727`) lists with `all=True` and returns a bare integer; the caller (`src/xagent/web/sandbox_manager.py:2554-2568`) emits a single `logger.warning` whether those containers are merely stopped leftovers (harmless) or actually running (the dangerous case — a live legacy dual-writer against the same host bind-mount as its v2 replacement). `docker/README.md` does warn operators not to run a legacy container alongside its v2 replacement, but that is enforced purely by prose; nothing in the code detects or escalates it. The risk also isn't confined to the upgrade window — a stopped legacy container can be restarted later while v2 sandboxes are live, and the log gives no distinguishing signal. `Container.status` is already populated from the same list response, so splitting running vs. stopped costs one filter expression, not an extra API round-trip. Suggest reporting the two counts separately and logging the running case at `error` or higher.

**D2 — no migration path for pre-namespace snapshot images** (still open from a prior round, latent). The container story has a documented upgrade procedure; snapshot *images* have none, and `docker/README.md` doesn't mention them. Two co-located deployments holding pre-upgrade snapshot rows for the same `snapshot_id` currently point at the same un-namespaced legacy image, so either side's `delete_snapshot` would remove it out from under the other. This is genuinely latent today — nothing in production currently reaches `DockerSandboxService.create_snapshot`/`delete_snapshot` — so it isn't blocking, but it's worth an issue or a README note before snapshot creation is ever wired to a real caller.

## Findings

All of these are minor or nit level; none block merge.

**1. Minor — `src/xagent/web/sandbox_manager.py:2522-2540`: namespace fatal check now precedes the import guard, turning a graceful degrade into a startup crash.**
The "namespace is required" check (lines 2522-2535, raising `RuntimeError`) runs *before* `from ..sandbox import DockerSandboxService` and its `ImportError` guard (lines 2536-2540, which logs and returns `None`). Compared against base: a deployment with `SANDBOX_ENABLED=true`, the `docker` extra not installed, and no namespace set previously hit the import guard and degraded to `None`; it now crashes startup with a message ("required when the Docker sandbox implementation is enabled") that is actively misleading, since Docker support isn't even installed. The same fatal path is reachable through the pre-existing unknown-`SANDBOX_IMPLEMENTATION` fallback (`src/xagent/web/sandbox_manager.py:2485-2489`, unchanged but now calling into the modified function), so a typo'd implementation value plus an unset namespace also flips from warn-and-degrade to crash. Separately, a malformed (non-empty but grammar-invalid) namespace raises `ValueError` from `src/xagent/config.py:2139` via `validate_sandbox_namespace`, called uncaught at `src/xagent/config.py:2169` — a *different* exception type for a closely related failure. Both propagate uncaught through `get_sandbox_manager()` (`src/xagent/web/sandbox_manager.py:2595`) into the startup handler at `src/xagent/web/app.py:1458-1460`, and neither crash path is documented or tested. Suggest moving the namespace check after the import guard (or wrapping both in a single exception boundary with one accurate error surface), and normalizing on one exception type for the unset and malformed cases.

**2. Minor — `src/xagent/web/sandbox_manager.py:2576-2605`: failed initialization is never memoized, so the error re-raises on every call** (raised in two prior rounds, still open; note this function is not touched by this PR's diff, carrying it forward for visibility).
When `_create_sandbox_service()` raises at line 2595 — including the `RuntimeError`/`ValueError` from finding #1 — the exception escapes without `_sandbox_manager_initialized` ever being set; that flag is only set on the success path (line 2602) or the explicit service-is-`None` path (line 2597). Every later `get_sandbox_manager()` call (four sites outside startup: `src/xagent/web/api/chat.py`, `src/xagent/web/api/tools.py`, `src/xagent/core/tools/adapters/vibe/ssh_tools.py:318`) therefore re-attempts construction and re-raises. This is harmless on the intended startup-crash path since the process exits, but any caller that catches and downgrades the exception will do so on *every single call*, silently reversing the "fatal, never silently degrade" guarantee this PR is built around. Suggested shape:

```python
try:
    service = _create_sandbox_service()
except Exception:
    _sandbox_manager_initialized = True
    raise
```

(or cache and re-raise a stored exception.)

**3. Minor — `src/xagent/sandbox/docker_sandbox.py:209-249`: the snapshot tag budget only caps one of its two halves.**
The 100-char namespace-token budget justifies itself in-comment by reference to the ≤128-char tag limit, but the other half of the tag — `_make_safe_name(snapshot_id)` — has no cap at all and produces `len(snapshot_id) + 11` chars. `snapshot_id` is stored as `String(255)` with no `max_length` enforced where it's set (`src/xagent/web/models/sandbox.py:51`, `src/xagent/sandbox/base.py:124-126,182`), so a long id can blow past both the 128-char tag limit and the 255-char total reference budget the comment reasons about. Latent today (no production caller reaches `create_snapshot`), but worth capping before one is wired up.

**4. Nit — `docker/docker-compose.sandbox.docker.yml:21-29` and `tests/test_docker_compose_assets.py`: the new test name overclaims.**
The overlay sets `XAGENT_SANDBOX_NAMESPACE` on `worker` and `scheduler`, but neither mounts `/var/run/docker.sock` nor receives `XAGENT_SANDBOX_HOST_PROJECT_ROOT`/`XAGENT_SANDBOX_HOST_STORAGE_ROOT` (both `backend`-only), and no worker or scheduler code path calls `get_sandbox_manager()` — consistent with the compose files' own comments (`docker-compose.yml:67,104`) that agent execution stays in the backend. The wiring is inert and harmless, but `test_overlay_propagates_one_namespace_to_every_sandbox_capable_service` asserts a "sandbox-capable" property the compose topology doesn't currently support. Either rename to reflect that this is forward-looking uniformity, or note the intent in a comment.

**5. Nit — `docker/README.md`, "Sandbox Runtime Overlays": examples don't pin `COMPOSE_PROJECT_NAME`.**
The runnable `docker compose -f docker-compose.yml -f docker/docker-compose.sandbox.docker.yml up -d` blocks never set `COMPOSE_PROJECT_NAME`. This is *not* a hard failure in the common case — run from the project root as the docs instruct, Compose derives the project name from the directory basename and the new `${COMPOSE_PROJECT_NAME:?...}` guard is satisfied. The gap is narrower: that implicit value is tied to the checkout directory's basename, which is not the "stable, unique" identifier the guard's own error message asks for. Differently-named clones, randomized CI checkout directories, or a directory rename would silently change the effective namespace instead of erroring. One extra line in the examples showing `-p <name>` or `COMPOSE_PROJECT_NAME` in `.env` would close it.

**6. Nit — test coverage gaps.**
- Namespace-scoped capacity accounting (`XAGENT_SANDBOX_MAX_CONTAINERS` applying per-namespace, via `_ensure_capacity_for` at `src/xagent/web/sandbox_manager.py:1142,1228,1389`) is only exercised transitively through `list_sandboxes` filtering, never asserted directly — despite being a load-bearing claim in both the PR description and `docker/README.md`.
- Nothing asserts `_container_name`'s separator-collision safety (e.g. `namespace="a", name="b::c"` vs. any other pair). It's safe by construction since the namespace grammar excludes `:`, but the property itself is unpinned.
- The real-daemon snapshot round-trip (`test_snapshot_lifecycle_with_trailing_dash_namespace`) is `@requires_docker`-gated and skips in any environment without a reachable daemon, so it provides no CI signal by default.

## Since the last review

- **Compose namespace guard** — fixed. All three services now use the identical `${COMPOSE_PROJECT_NAME:?set COMPOSE_PROJECT_NAME to a stable unique value}` expression, pinned by `tests/test_docker_compose_assets.py`.
- **Snapshot repository name collision** — fixed. The digest is now appended unconditionally rather than only when sanitization changed the token; hand-tracing the originally-reported colliding pair confirms they now differ, and the new regression test asserts exactly that pair. The residual risk is a genuine ~40-bit SHA-1 collision.
- **Misleading error on legacy-count failure** — fixed. Legacy counting has its own try/except with an accurate "Failed to inventory legacy sandbox containers" message, the constructed service is still returned, and `test_legacy_inventory_failure_does_not_misreport_creation` covers it.
- **Per-namespace container budget undocumented** — fixed. `docker/README.md` now states that `XAGENT_SANDBOX_MAX_CONTAINERS` applies independently per namespace and that daemon sizing must account for co-located stacks.
- **Container name length cap** — not escalating; the decision to omit one is defensible. Moby's container-name validation regex carries no length quantifier, unlike the image-reference grammar that motivated the snapshot-repository cap, and this codebase avoids the two real-world naming pitfalls (DNS alias length on user-defined networks, hostname length) since it uses neither custom networks nor explicit hostnames.
- **Correction on an earlier "Celery worker turns a graceful degrade into an unhandled RuntimeError" finding from a previous review round** — that finding was based on an incorrect call-chain trace, and it's worth being explicit about the correction. Re-tracing exhaustively: `get_sandbox_manager()`/`_create_docker_service()` is not reachable from any Celery worker or beat path in this codebase. `register_celery_tasks()` never imports the sandbox module or anything that transitively does, and the compose comments confirm agent execution and sandbox access are confined to the backend process. The Compose-level namespace-propagation gap that finding leaned on has separately been fixed and was a correct change on its own merits, but it should not have been framed as Celery-reachable. No action needed there.

Lean already. Every added mechanism — the `_ownership_label_filters` dedup, the `v2` label-value migration scheme, `_snapshot_repository`'s sanitize + cap + digest, and both namespace validation entry points — traces to a concrete caller and to a test that would fail without it. Nothing here reads as speculative flexibility.

## Testing

Ran locally: **532 passed, 45 skipped** via `pytest tests/core/test_config.py tests/sandbox/test_docker_sandbox.py tests/sandbox/test_docker_lifecycle_api.py tests/web/test_sandbox_manager.py tests/test_docker_compose_assets.py`. All 45 skips are gated on a reachable Docker daemon, which wasn't available in the review environment.

## Recommendation

**Approve with minor suggestions.** The approach is sound, every Major/Critical-tier concern from the prior rounds is genuinely resolved, and nothing remaining is blocking. Worth a quick follow-up on the two startup-path items — finding #1 (fatal namespace check ordered ahead of the import guard, plus the split `RuntimeError`/`ValueError` surface) and finding #2 (failed initialization never memoized) — since together they govern how this feature behaves when misconfigured, which is exactly the path operators will hit first during rollout.

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/sandbox-namespace-isolation`
- **Head:** `6d1172b422dcc495176e52c277edbb01bf31ca58`
- **Destination:** `git@github.com:bsbds/xagent.git` → `fix/sandbox-namespace-isolation` (head of xorbitsai/xagent#1174)
- **Result:** success

The supplied feedback is a single top-level APPROVED review body and created no inline review threads. GitHub therefore has no per-item thread state to mutate; this aggregate comment records every evaluated outcome while preserving the review state.

### Resolved

1. **Thread:** Review `4880796305`, item R1-D1 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** resolved
   **Commit:** `d0ba9232219940da960af491d13655a2b5e6ec33`
   **GitHub:** no inline thread; addressed by the pushed commit and APPROVED review state preserved

2. **Thread:** Review `4880796305`, item R1-F4 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** resolved
   **Commit:** `3f9a54c79b9fefd2fe923c741322ff094cdf091d`
   **GitHub:** no inline thread; addressed by the pushed commit and APPROVED review state preserved

3. **Thread:** Review `4880796305`, item R1-F6a — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** resolved
   **Commit:** `6d1172b422dcc495176e52c277edbb01bf31ca58`
   **GitHub:** no inline thread; addressed by the pushed commit and APPROVED review state preserved

### Rejected

1. **Thread:** Review `4880796305`, item R1-D2 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Reason:** Current xagent-saas production has no Docker snapshot create/delete caller; migration behavior should be designed before that surface is supported.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

2. **Thread:** Review `4880796305`, item R1-F1 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Classification:** unsound
   **Reason:** Docker 7.1+ is a hard runtime dependency, and fail-closed namespace validation before daemon construction is intentional.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

3. **Thread:** Review `4880796305`, item R1-F2 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Reason:** Startup namespace failures are fatal and no supported caller catches them and continues; sticky failure caching adds state without a demonstrated runtime benefit.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

4. **Thread:** Review `4880796305`, item R1-F3 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Reason:** Snapshot creation has no current production caller; define truncation or rejection semantics when it becomes a supported surface.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

5. **Thread:** Review `4880796305`, item R1-F5 — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Reason:** Production deployment exists only in xagent-saas, which supplies a stable project name; standalone xagent examples are not a supported production path.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

6. **Thread:** Review `4880796305`, item R1-F6b — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Reason:** Namespace grammar excludes the separator and the complete namespace/name pair is hashed, so the proposed test does not cover a realistic ambiguity.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

7. **Thread:** Review `4880796305`, item R1-F6c — https://github.com/xorbitsai/xagent/pull/1174#pullrequestreview-4880796305
   **Local outcome:** rejected
   **Classification:** outdated
   **Reason:** Core CI includes `tests/sandbox`, pre-pulls the sandbox image, and runs with a reachable Docker daemon.
   **Reply:** included in this aggregate summary because no inline thread exists
   **GitHub:** APPROVED review state preserved

### Unresolved

None.

### Totals

- **Resolved:** 3
- **Rejected:** 7
- **Unresolved:** 0

Issue #1161, opened 2026-08-06
feat(deepdoc): offload document parsing to a remote Xinference GPU server


OliverBryant:
Umbrella issue for offloading DeepDoc document parsing to a remote Xinference server running on a user's own GPU machine, configured purely via environment variables.

Full requirements, flow diagrams, and the proposed server API contract live in `docs/deepdoc-remote-parsing.md` (added by the WIP umbrella PR below).

## Motivation

DeepDoc's PDF pipeline (OCR detection/recognition, layout analysis, table structure recognition) runs entirely as local ONNX inference and dominates ingestion cost on CPU-only deployments — large documents take minutes.

Two facts shape the design:

- `deepdoc-lib` 0.2.2 has **no usable remote inference path**. The `TENSORRT_DLA_SVR` hook in `vision/layout_recognizer.py` imports `deepdoc.vision.dla_cli`, a module that does not ship in the package.
- Xinference has no DeepDoc serving capability today. The server API is a **proposal** in the requirements doc, to be implemented separately.

## Design decisions

- **Transparent switching** — when `XAGENT_DEEPDOC_XINFERENCE_URL` is set, the existing `deepdoc` parser routes to the remote service. No new `ParseMethod`, no frontend change, no capability-marker change (`DeepDocParser` must keep `LocalParsing`, or the registry filter in `document_parser.py` hides it since `use_local_parser` defaults to `True`).
- **All formats, not just PDF** — every format `DeepDocParser` supports (`.pdf`, `.docx`, `.xlsx`, `.xls`, `.csv`, `.md`, `.txt`, `.json`, `.html`) goes remote once configured.
- **Always fall back** — any remote failure logs a warning and falls back to local inference. There is deliberately no fallback toggle.
- **Whole-document API** — one multipart upload returns structured elements, rather than per-model remote inference calls.

## Configuration

| Environment variable | Required | Default |
|---|---|---|
| `XAGENT_DEEPDOC_XINFERENCE_URL` | yes, to enable remote | unset (= local mode) |
| `XAGENT_DEEPDOC_XINFERENCE_API_KEY` | no | falls back to bare `XINFERENCE_API_KEY`, then no auth header |
| `XAGENT_DEEPDOC_XINFERENCE_TIMEOUT_SECONDS` | no | `1800` |

## Status board

Everything below is delivered by the single integration PR **#1271**, which supersedes the
per-step PRs (all now closed). The server-side dependency landed as xorbitsai/inference#5299.

[...]

PR #1175, opened 2026-08-06
test(deepdoc): unit-test the remote parse client with mocked HTTP

### LINKED ISSUE #1164: test(deepdoc): unit-test the remote parse client with mocked HTTP

OliverBryant:
Closes #1164

Step **S3** of umbrella #1161.

## Stacking

Branched off #1173 (S2, the client module), which is itself stacked on #1172 (S1, the config getters). Neither is merged yet, so **this diff replays their files** — `src/xagent/config.py`, `tests/core/test_config.py`, and `src/xagent/providers/pdf_parser/deepdoc_remote.py`. Review only `tests/providers/pdf_parser/test_deepdoc_remote.py`; once #1172 and #1173 land, the rest disappears from this diff. No source file is touched by this PR.

## What

One new file, `tests/providers/pdf_parser/test_deepdoc_remote.py` — 73 pure unit tests for `providers/pdf_parser/deepdoc_remote.py`. Style follows the existing pure-unit test `tests/providers/pdf_parser/test_deepdoc_table_positions.py`.

No network and no DeepDoc model cache: an `httpx.MockTransport` goes in through the module's `_transport` seam, and `save_image` is a plain recording callable, so the tests run anywhere.

## Why

The client's whole value to the caller is a single-type failure contract: *every* remote problem must arrive as `DeepDocRemoteError`, because the routing layer's one `except DeepDocRemoteError` clause is what makes the fallback to local parsing unconditional. A leaked `httpx.ConnectError` or `binascii.Error` would escape that clause and break document parsing outright instead of quietly degrading. That contract needs to be pinned before S4 wires the routing up.

Coverage:

- **Element translation** — `image_base64` becomes an `image` path string for the table element and `None` for the text element, `save_image` is called exactly once, `metadata` survives, `image_base64` is never forwarded, and the source payload is not mutated.
- **Request shape** — multipart carries the real filename and MIME type, `zoomin` is forwarded (and defaults to 3), `Authorization: Bearer …` is present with a key and absent without one, the dedicated key beats the shared `XINFERENCE_API_KEY`, and `image_scope` is deliberately not sent so the server's `table_figure` default applies.
- **URL joining** — no double slash even when the configured URL ends in one or more `/`; the final request URL is asserted to be exactly `{base}/v1/document/parse`.
- **`BytesIO` input** — filename falls back to `document{ext}`, the whole buffer is uploaded, and the cursor is restored on both success and failure so the local fallback can re-read it.
- **Failure modes**, each asserting `DeepDocRemoteError` — HTTP 500, HTTP 401, connection error, read timeout, non-JSON body, 200 with an empty body, a JSON top-level list, `elements` missing / `None` / not a list, an element that is not a dict or is `None`, an element missing `text` or `type`, undecodable base64, an unreadable source file, an unconfigured URL, and `save_image` raising `RuntimeError`, `OSError`, or `binascii.Error`.
- **`is_remote_configured()`** — `False` when unset or blank, and `False` **without raising** for a malformed URL (wrong scheme, no scheme, query string, fragment, no netloc), with the warning asserted.

### The failure suite is not vacuous

A parametrized companion, `test_underlying_exception_types_do_not_escape`, asserts both that the raised value is *not* an instance of the underlying type and that `__cause__` *is*. `DeepDocRemoteError` derives straight from `Exception`, so a regression that stopped wrapping would satisfy neither that check nor `pytest.raises(DeepDocRemoteError)`.

Verified by mutation: with the `try`/`except` block temporarily stripped out of `parse_document_remote`, **27 of the 73 tests fail**. The module was restored byte-identically afterwards (`git diff` clean, only the new test file staged).

### No models, no network

- `deepdoc_remote.py`'s only imports are `base64`, `binascii`, `json`, `logging`, `time`, `io`, `pathlib`, `typing`, `httpx`, and `...config` — verified by AST, no parser and no ONNX.
- With `onnxruntime.InferenceSession` and `deepdoc.DeepDocPdfParser` both armed as tripwires, a full `parse_document_remote` call completes without firing either.
- The whole file passes with `socket.socket.connect`, `connect_ex`, and `socket.create_connection` replaced by hard failures.

(`deepdoc` and `onnxruntime` do get imported, but only because `xagent/providers/pdf_parser/__init__.py` eagerly imports `.deepdoc` for the registry — pre-existing package behavior, unrelated to this client, and it loads no model weights.)

## Testing

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_remote.py -q
........................................................................ [ 98%]
.                                                                        [100%]
73 passed in 3.21s

$ .venv/bin/python -m ruff format --check tests/providers/pdf_parser/test_deepdoc_remote.py
1 file already formatted

$ .venv/bin/python -m ruff check tests/providers/pdf_parser/test_deepdoc_remote.py
All checks passed!

$ uvx --quiet isort==6.0.1 --check-only --diff tests/providers/pdf_parser/test_deepdoc_remote.py
(no output, exit 0)

$ .venv/bin/python -m mypy tests/providers/pdf_parser/test_deepdoc_remote.py
(no findings in the new file; the 41 reported errors are pre-existing ones in src/,
 and CI's mypy hook runs --package xagent, which excludes tests)
```

Mutation check, for the record:

```
$ # try/except temporarily removed from parse_document_remote
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_remote.py -q
27 failed, 46 passed in 2.20s
```


rogercloud (approved):
## Review: test(deepdoc): unit-test the remote parse client with mocked HTTP

### Summary

This PR adds a single test file, `tests/providers/pdf_parser/test_deepdoc_remote.py` (763 lines, 73 tests), unit-testing the remote DeepDoc parse client (`deepdoc_remote.py`) against a mocked HTTP layer via `httpx.MockTransport`. It closes issue #1164 and is step S3 of umbrella #1161. The diff replays files from stacked PRs #1172 (config getters) and #1173 (the client module itself); no source file is touched by this PR — only the new test file is in scope for review.

### Approach verdict: sound

- `httpx.MockTransport` through the `_transport` seam is the right mock level: real multipart encoding, real `response.json()`, real `raise_for_status()` — only the wire is stubbed.
- Testing private symbols (`_build_headers`, `_mime_type_for`, `_normalize_elements`, `PARSE_ENDPOINT`) matches house style; the sibling `tests/providers/pdf_parser/test_deepdoc_table_positions.py` tests `_translate_pdf_bboxes` the same way.
- `test_underlying_exception_types_do_not_escape` (:496-551) is a genuinely good design: the `__cause__` assertions guard the failure suite against passing for the wrong reason.
- Mild reservation (optional): the same malformed payloads are tested twice — through the full pipeline (:414-440) and directly against `_normalize_elements` (:662-667) — duplicating ~6 near-identical cases. Defensible, non-blocking.

### Findings

No critical or major findings. All findings below are minor or informational.

#### Minor

**F1 — `tests/providers/pdf_parser/test_deepdoc_remote.py:171-198`**
`test_request_carries_the_file_zoomin_and_auth_header` never asserts an auth header. The autouse `remote_env` fixture deletes both API-key env vars (:61-62), no key is set, and no authorization assertion exists in :187-198 — the docstring "the bearer token ride along" is false for this test. Auth is covered by the two dedicated tests (:215, :232). Fix: rename to `test_request_carries_the_file_and_zoomin`, or set the key and assert the header.

**F2 — `tests/providers/pdf_parser/test_deepdoc_remote.py:514-518, 551`**
The `non-json-body` parametrize case uses `leaked_type=ValueError`, but the actual `__cause__` is `json.JSONDecodeError` (a subclass of ValueError). `deepdoc_remote.py:318-321` has a dedicated `except json.JSONDecodeError` clause ahead of the generic ValueError clause (:326); deleting that dedicated clause would NOT fail this test (it would fall into the ValueError clause and keep the same `__cause__` type). Tighten the param to `json.JSONDecodeError` so it pins the dedicated branch. Contrast `element-is-none` (:519-523), whose cause genuinely is a plain ValueError — leave that one.

**F4 — `tests/providers/pdf_parser/test_deepdoc_remote.py:213` (and :198)**
The multipart assertions are positional-string brittle. At :198 the bare `assert "\r\n\r\n5\r\n" in body` is paired with `assert 'name="zoomin"' in body`; at :213 only the bare value `"\r\n\r\n3\r\n" in requests[0].content.decode(...)` is asserted, with no `name="zoomin"` pairing. There is no false-pass risk today (the request has exactly two parts — zoomin and file; the boundary is random; file content starts with `%`), but the unanchored assertion would silently pass if a future numeric form field happened to carry the value 3. Suggestion: add the `name="zoomin"` pairing for consistency.

**F5 — minor (coverage gap, forwarded to #1173 — not this PR's to fix) — `deepdoc_remote.py:174-180`**
The timeout wiring (`httpx.Timeout(connect=_CONNECT_TIMEOUT_SECONDS, read=timeout_seconds, write=timeout_seconds, pool=_POOL_TIMEOUT_SECONDS)` from `get_deepdoc_xinference_timeout_seconds()`) is never asserted anywhere in the test suite. A regression swapping `read=`/`connect=` (capping every parse at 10s instead of 1800s) would fail no test; the two read-timeout params (:401, :505) inject `httpx.ReadTimeout` directly and say nothing about the configured budget. However, this is the #1173 source file replayed into this diff, the PR body says no source file is touched by this PR, and issue #1164 does not list it. Recommend forwarding this to #1173; not a blocker for #1175.

**F7 — `tests/providers/pdf_parser/test_deepdoc_remote.py` (14 sites: 122-123, 155-156, 175-176, 202-203, 218-219, 236-237, 251-252, 280-281, 331-332, 460-461, 485-486, 539-540, 569-570, 585-586)**
Repeated boilerplate `source = tmp_path / "..."` + `source.write_bytes(b"%PDF-1.7 fake")`. 12 of 14 are verbatim "report.pdf"; :175 "quarterly report.pdf" is asserted by the test (keep the filename parameterizable); :331 "empty.pdf" is semantic naming only (safe to fold in). A `@pytest.fixture` returning the written `Path` (optionally parameterized) saves ~19-20 net lines and makes the filename-varying tests visibly intentional.

**F8 — `tests/providers/pdf_parser/test_deepdoc_remote.py:177, 204, 221, 238, 253, 283, 300`**
Repeated `requests: list[httpx.Request] = []` + `json_transport({"elements": []}, sink=requests)` (7 identical pairs). A small fixture/helper saves ~3-4 net lines only — real but small.

**F9 — partial — `tests/providers/pdf_parser/test_deepdoc_remote.py:689-702` vs `tests/core/test_config.py:1258-1274`**
The API-key precedence tests are genuinely redundant: `_build_headers()` (`deepdoc_remote.py:135-140`) has no precedence logic of its own — precedence lives entirely in `get_deepdoc_xinference_api_key()` (config.py), and `test_config.py:1261-1274` (`test_api_key_dedicated_var_wins`, `test_api_key_falls_back_to_shared_xinference_key`, `test_api_key_blank_dedicated_var_falls_back`) already exercise every branch. `test_shared_key_is_the_fallback` (:693) and `test_dedicated_key_wins_over_the_shared_one` (:697) add no code path in `deepdoc_remote.py` beyond `test_dedicated_key_is_used` (:689) plus `test_no_key_configured_yields_no_headers` (:686). Suggestion: drop those two. KEEP the `TestIsRemoteConfigured` URL matrices (:723-763): they uniquely cover `is_remote_configured()`'s own ValueError-swallow branch (`deepdoc_remote.py:119-132`), which `test_config.py` never exercises (its `test_url_rejects_invalid_values` only asserts the getter raises). The only overlap is parametrize breadth, not logic.

**F10 — `tests/providers/pdf_parser/test_deepdoc_remote.py:326, 631, 675, 686, 689, 697` and :279`**
Docstring inconsistency: 6 tests lack the one-line docstring their neighbors and the sibling file (`test_deepdoc_table_positions.py`, uniform) have. Also :279's docstring says the config getter "strips the trailing slash" (singular) while `config.py:367-368` `.rstrip("/")` strips ALL trailing slashes — and the `many-slashes` param (:271-274) exercises exactly the plural behavior. Suggest aligning the wording.

#### Informational

**F3 — informational only (no action) — `tests/providers/pdf_parser/test_deepdoc_remote.py:550`**
`assert not isinstance(excinfo.value, leaked_type)` is tautological: `DeepDocRemoteError` derives directly from `Exception`, so it can never be an instance of `httpx.ConnectError`/`ReadTimeout`/`HTTPStatusError`/`ValueError` under any code change. The docstring at :532-537 already discloses this as deliberate "keep the reader honest" self-documentation; the real work is done by :551 (`isinstance(excinfo.value.__cause__, leaked_type)`). Passing note — this could be a comment instead of an assert; no change requested.

**F6 — informational — `deepdoc_remote.py:265-268`**
The single uncovered line (266, `raise ValueError("Remote DeepDoc element ... carries an empty image")`) is dead code, not a missing test: with `base64.b64decode(validate=True)`, no truthy string decodes to `b""` (brute-forced all length-1..4 strings over the full base64 alphabet plus whitespace: zero hits; padding-only raises "Leading padding not allowed"; whitespace-only raises "Only base64 data is allowed"; the shortest valid decode is 4 chars ≥ 1 byte). The `if image_base64:` truthiness guard upstream already excludes `""`. The guard is belt-and-braces; the 98% figure is fine. (Source-side, belongs to #1173.)

**F11 — optional note (no change requested) — `tests/providers/pdf_parser/test_deepdoc_remote.py:414-440` vs `662-667`**
The same malformed payload classes are tested at two levels (full pipeline via `parse_document_remote` vs direct `_normalize_elements`), duplicating ~6 near-identical cases. Defensible (raw type vs wrapping), but it inflates the test count. Optional consolidation, non-blocking.

### Re-review: prior gemini-code-assist review (body-only, 3 suggestions)

The prior review suggested strict complete-shape validation requiring `metadata` and `image_base64` keys in every element (`deepdoc_remote.py:249`), matching test-input changes (:~625), and new parametrize cases for missing-metadata / missing-image_base64 (:~667). Head has not moved since that review, so all three are **NOT FIXED**.

**Recommendation: do not adopt in #1175; forward to #1173.** Reasons, verified:
1. Out of scope: the source file is #1173's (replayed into this diff); the PR body explicitly says no source file is touched by this PR.
2. The server contract is a documented PROPOSAL — the module docstring states "Xinference has no DeepDoc serving capability today; the endpoint below is a proposal". Enforcing strict shape against a not-yet-existing server would bind the client to a format that may still change.
3. `image_base64` is documented as nullable (null outside tables/figures), and JSON serializers commonly omit null keys (omitempty semantics) — strict key-required validation would reject valid responses and trigger unnecessary fallback, contradicting the module's stated "fallback only on real failure" goal.
4. `metadata` has no downstream consumer (`parse_document_remote` / `_normalize_elements` have zero callers in the repo; the module is not yet wired into `deepdoc.py`). No KeyError risk exists today.
5. It would reverse an already-tested deliberate design: `test_absent_image_base64_still_yields_an_image_key` (:619-628) docstring states "The caller always reads image, so the key must exist even unset" — the missing-key behavior is intentional and coverage-locked.
6. Gemini's own description of the current behavior is imprecise ("silently defaulted to {}" — metadata is in fact simply absent from the result when missing).

Tests comments 2-3 are contingent on comment 1; they are pointless until #1173 decides.

- L122-123: shrink Repeated `source = tmp_path / "report.pdf"` + `write_bytes(b"%PDF-1.7 fake")` (14 sites). Use a `pdf_file(tmp_path)` fixture (parameterized for the one filename-asserting test). net: -19 lines
- L177: shrink Repeated `requests: list[httpx.Request] = []` + `json_transport({"elements": []}, sink=requests)` (7 sites). A fixture/helper returning (transport, requests) removes the boilerplate. net: -3 lines
- net: -22 lines possible

### Test results

- `pytest tests/providers/pdf_parser/test_deepdoc_remote.py` → **73 passed**, 15 pre-existing third-party warnings.
- Coverage of `deepdoc_remote.py`: **98%**; the one uncovered line (266) is verified unreachable (dead guard, see F6).
- Mutation probe: deleting the wrap-try/except (`deepdoc_remote.py:311-336`) caused **27 of 73 tests to fail** — the suite is genuinely non-vacuous.
- Issue #1164 scope: every listed item maps to a passing test; no gap.

### No-issue categories

- **Test isolation:** clean — monkeypatch-only env mutation, no module-level mutable state, no ordering dependence.
- **Documentation:** module docstring (:1-15) is excellent.
- **Design:** sound (see approach verdict above).

### Review state

**APPROVE** — no blocking findings. This is a re-review of a test-only PR; all findings are minor or informational. The minor items (F1, F2, F4, F7, F8, F9, F10) can be addressed in follow-up commits; the source-side items (F5, F6) and the prior gemini suggestions belong to #1173.

OliverBryant:

Issue #1166, opened 2026-08-06
test(deepdoc): cover remote routing, local fallback, and the element translator


OliverBryant:
Part of #1161 · Step **S5** · depends on: S4 (routing refactor + translator)

Test the routing decision, the fallback, and the remote-element translator. All pure unit tests: no network, no DeepDoc model cache.

## Scope

New file `tests/providers/pdf_parser/test_deepdoc_remote_fallback.py`.

### Routing and fallback

- **Remote success** (one `.pdf` case and one `.docx` case): monkeypatch `parse_document_remote` to return canned elements, and arm `DeepDocParser._get_parser_for_ext` as a tripwire that raises `AssertionError`. The parse must succeed, proving the local ONNX path was never touched, and the result metadata must carry `deepdoc_backend=remote`.
- **Remote failure**: monkeypatch `parse_document_remote` to raise `DeepDocRemoteError` and stub `_get_parser_for_ext` to return a fake whose `parse_into_bboxes` returns canned bboxes. The local fallback result must be returned, `caplog` must contain the warning, and metadata must carry `deepdoc_backend=local`.
- **Env unset**: arm `parse_document_remote` as a tripwire; the pure-local path must be taken.

### Translator

- A PDF-shaped element set (with `positions`, `col_id`, `page_number`, plus a table and a figure) and an xlsx-shaped element set (with `sheet_name`, `row_number`, `row_type`): assert the resulting `text_segments` / `tables` / `figures` and their metadata keys match what the local translators produce for equivalent input.

Patch `_save_bytes_to_disk` (or the `ARTIFACTS_DIR` it resolves against) to `tmp_path` for any case that exercises real image bytes.

## Note

`tests/providers/pdf_parser/test_deepdoc_parser.py` runs real inference in the CI `pytest-fast-deepdoc` job with a warm model cache. Leave it untouched — it now also serves as the end-to-end regression for "env unset → local".

## Done when

`pytest tests/providers/pdf_parser/test_deepdoc_remote_fallback.py -q` passes on a machine with no model cache and no network.

PR #1176, opened 2026-08-06
feat(deepdoc): route parsing to the remote client with local fallback

### LINKED ISSUE #1165: feat(deepdoc): route parsing to the remote client with local fallback

OliverBryant:
Closes #1165 · Step **S4** of #1161

Routes document parsing to the remote DeepDoc server when
`XAGENT_DEEPDOC_XINFERENCE_URL` is configured, falling back to local ONNX
inference on any remote failure.

> **Stacking:** this builds on #1173 (the `deepdoc_remote.py` client module),
> which in turn stacks on #1172 (config getters). Review those first; the
> diff here is only `src/xagent/providers/pdf_parser/deepdoc.py`.

## The deferred-instantiation change, and why it matters

This is the point of the PR. `_sync_parse` previously called
`self._get_parser(file_path)` on its **first line**, and `DeepDocPdfParser()`
eagerly loads the ONNX detection/recognition/layout models — potentially
*downloading them from ModelScope* on a cold machine.

On the remote path none of that is needed. Loading (or downloading) hundreds of
MB of models only to send the file over HTTP would erase most of the benefit of
offloading, and would make remote mode unusable on a slim CPU-only container
that never wants the model weights at all.

So `_sync_parse` now detects **only the extension** up front and defers
`_get_parser_for_ext(ext)` to just before each local branch actually uses it.
The remote path never constructs a parser. This is verified below by a tripwire
that makes `_get_parser_for_ext` raise `AssertionError`.

## Changes

1. **Extension detection first** — `kwargs.get("file_ext", ".xlsx")` for
   `BytesIO`, else `Path(file_path).suffix.lower()`. No parser built here.
2. **Deferred local instantiation** — `_get_parser_for_ext` moved down into the
   `.md`, PDF, and other-format branches. The `.xlsx` `_parse_xlsx_rows` early
   return, the `.md`/`.txt` direct reads, the `BytesIO` conversion, and the
   `run_in_executor` wrapper are all untouched.
3. **Progress for all formats** — `DeepDocProgressAdapter` construction hoisted
   out of the PDF branch, so remote mode reports progress for every format. The
   local PDF branch still passes it exactly as before.
4. **Remote-first dispatch** — after extension detection and after the cheap
   `validate_office_file_format` magic-byte check (kept *before* the upload so a
   bad file fails fast without wasting bandwidth), and before any local branch.
   Sets `metadata["deepdoc_backend"]` to `"remote"` or `"local"`.
5. **`_translate_remote_elements`** — reshapes each remote element into the
   bbox-shaped dict the existing helpers already consume, then reuses
   `_build_element_metadata`, `_process_table_element` and
   `_process_figure_element` rather than duplicating them. Text elements become
   `ParsedTextSegment`. Format-specific metadata (`style`, `sheet_name`,
   `row_number`, `row_type`) flows through `element["metadata"]` into the
   segment metadata, matching the local translators.

### One extra hardening not in the spec

The fallback progress notification is wrapped in its own `try/except`.
`DeepDocProgressAdapter`'s error handler calls `on_status_update` a second time
*inside* its own `except` block, so a misbehaving progress sink can raise out of
the callback — which would have turned a recoverable remote failure into a hard
parse failure and defeated the fallback entirely. I hit this for real while
writing the verification script. Test 2b below covers it.

### Unchanged, deliberately

The `LocalParsing` capability marker stays. Swapping in `RemoteParsing` would
make the registry filter at `document_parser.py:283` hide the parser outright,
since `DocumentCapabilities.use_local_parser` defaults to `True`. Also untouched:
`supported_extensions`, the `ImportError` guard in `__init__.py`, the registry,
the progress adapter, and the `enable_raw_output` fields.

### Follow-up

`DeepDocParser._get_parser()` is now dead code — `_sync_parse` was its only
caller and it now uses `_get_parser_for_ext` directly. Left in place to keep
this diff focused; worth removing in a small separate cleanup.

## Testing

Behavioral tests are S5 (#1166); this PR adds none. Existing tests plus
throwaway verification scripts (not committed) were run.

### Existing tests still pass

`test_deepdoc_table_positions.py` is the regression gate — it proves the
refactor did not disturb existing translation.

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_table_positions.py -q
..................                                                       [100%]
18 passed in 15.10s
```

Full provider suite — identical to the pre-change baseline (the one failure is
a missing `fitz` optional dependency, and fails the same way on the base commit):

```
$ .venv/bin/python -m pytest tests/providers
FAILED tests/providers/pdf_parser/test_basic_parsers.py::test_basic_parsers_pdf_success[PyMuPdfParser]
1 failed, 87 passed, 2 skipped in 43.31s
```

The 73 client tests from S3 (#1175) also pass unchanged against this branch
(checked out temporarily, not committed here):

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_remote.py
73 passed in 2.61s
```

### Lint / format / types

```
$ .venv/bin/python -m ruff format --check src/xagent/providers/pdf_parser/deepdoc.py
1 file already formatted

$ .venv/bin/python -m ruff check src/xagent/providers/pdf_parser/deepdoc.py
All checks passed!

$ uvx --quiet isort==6.0.1 --check-only --diff src/xagent/providers/pdf_parser/deepdoc.py
(no output, exit 0)

$ .venv/bin/python -m mypy src/xagent/providers/pdf_parser/deepdoc.py
Found 26 errors in 9 files (checked 1 source file)

$ .venv/bin/python -m mypy src/xagent/providers/pdf_parser/deepdoc.py | grep -c deepdoc.py
0
```

All 26 mypy errors are pre-existing missing-stub errors in other modules
(cairosvg, lancedb, playwright, pyarrow, bs4); the count is **identical on the
base commit**, and none are in `deepdoc.py`.

### Deferred instantiation and fallback (the critical behavior)

Throwaway script: monkeypatches `parse_document_remote` to return canned
elements, arms `_get_parser_for_ext` as an `AssertionError` tripwire, sets the
env URL, and parses a `.pdf`.

```
=== TEST 1: remote success must NOT instantiate a local parser ===
  ext forwarded to client      : '.pdf'
  zoomin forwarded             : 3
  callback passed to client    : True
  save_image lambda worked     : data/artifacts/providers/deepdoc/proof_doc/images/2bac9b68-....png
  'proof_doc' in image path    : True
  text_segments                : 1
  tables                       : 1
  figures                      : 1
  segment backend              : 'remote'
  segment page_number          : 2
  segment col_id               : 1
  segment positions (enriched) : [[2, 1, 1.0, 2.0, 3.0, 4.0]]
  table sheet_name passthrough : 'Sheet1'
  table row_number passthrough : 7
  table row_type passthrough   : 'data'
  table style passthrough      : 'Heading 1'
  RESULT: PASS -- .pdf parsed remotely, tripwire never fired

=== TEST 2: DeepDocRemoteError must fall back to the local parser ===
  local parser was invoked     : True
  local received zoomin        : 3
  local received callback      : True
  segment text                 : 'local fallback text'
  segment backend              : 'local'
  progress statuses seen       : ['Remote parse failed; running local DeepDoc']
  RESULT: PASS -- fell back to local, backend tagged 'local'

=== TEST 2b: a RAISING progress sink must not abort the fallback ===
  local parser was invoked     : True
  segment backend              : 'local'
  RESULT: PASS -- exploding progress sink did not break fallback

=== TEST 3: unset URL = pure local, remote client never consulted ===
  local parser was invoked     : True
  segment backend              : 'local'
  RESULT: PASS -- local-only mode unaffected
```

Test 1 confirms the tripwire never fires: **no local parser is constructed on
the remote path**, so no ONNX load and no ModelScope download. It also confirms
`doc_id` is correctly in scope for the injected `save_image` lambda, and that
positions/col_id enrichment and format-specific metadata survive translation.

### All formats route remotely; local early paths intact

```
=== Remote mode: every format must go remote, no local parser ===
  .pdf   -> remote=True    .docx  -> remote=True    .xls  -> remote=True
  .csv   -> remote=True    .md    -> remote=True    .txt  -> remote=True
  .json  -> remote=True    .html  -> remote=True    .xlsx (BytesIO) -> remote=True
  exts sent to remote: ['.pdf', '.docx', '.xls', '.csv', '.md', '.txt', '.json', '.html', '.xlsx']

=== Local mode: .md and .txt still read directly from disk ===
  .txt segments : ['line one, with punctuation!\nline two.']
  .md  segments : ['# Title\n\nSome prose.\n']
  .xlsx segments: ['Name | Qty', 'Name: Widget | Qty: 5']
  .xlsx meta    : sheet='Data' row=1 type='header' backend='local'
```

`.txt` keeps its exact original punctuation and newlines (the reason it reads
directly rather than going through the parser), and `.xlsx` still reaches
`_parse_xlsx_rows` with `sheet_name`/`row_number`/`row_type` intact.

### Magic-byte check really does precede the upload

`validate_office_file_format` only rejects a legacy OLE2 file wearing an
Open-XML extension, so that is the case tested:

```
raised ValueError: File '.../legacy.docx' has extension .docx, but the actual format is ...
remote client called: False  (must be False = no wasted upload)
valid .xlsx reached remote client: True (must be True)
```

### Edge cases

```
empty elements -> segments=0 tables=0 figures=0 backend='remote'
unknown/missing-metadata -> segments=3
   'future type' layout='weird_new_type' page=1 col=0
   'no metadata key' layout='text' page=1 col=0
   'metadata is None' layout='text' page=1 col=0
table image_path : /tmp/.../crop.png
figure image_path: /tmp/.../crop.png
figure text default: 'Figure'
figure w/ bad path -> image_path=None (figure helper swallows, as locally)
```

An unrecognized future element `type` degrades to a text segment rather than
being dropped, a missing or `None` `metadata` key is tolerated, and the
existing `_handle_image` string branch consumes the client's saved image paths
unchanged.


src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+``xagent.config``. In particular it must not import ``deepdoc.py``: the
+image-saving function is injected by the caller instead, which keeps the import
+graph acyclic and lets the client be tested without loading any parser.
+
+Proposed server API contract (v1)
+---------------------------------
+
+Xinference has no DeepDoc serving capability today; the endpoint below is a
+proposal tracked alongside this client, mirrored from
+``docs/deepdoc-remote-parsing.md`` section 6.
```
rogercloud, line 19: The module docstring cites `docs/deepdoc-remote-parsing.md`, but that file exists only on the sibling umbrella branch (review/pr-1178), not in this PR or in the base. The three new env vars (URL, API key, timeout) and the egress/fallback policy are also undocumented anywhere in this PR. Please land the docs with the code or point the citation at the tracking issue/PR.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+class DeepDocRemoteError(Exception):
+    """Remote parsing failed and the caller should fall back to local parsing.
+
+    Every failure mode -- unreachable host, timeout, 4xx/5xx, unparsable or
+    malformed body, undecodable image, failed image write -- is reported as
+    this single type so callers need only one ``except`` clause.
+    """
+
+
+def is_remote_configured() -> bool:
```
rogercloud, line 119: This swallows the `ValueError` that `get_deepdoc_xinference_url()` raises on a malformed URL — see that getter's docstring (config.py:1859-1869): "a malformed value raises rather than silently downgrading to local parsing". Unset and malformed are now indistinguishable, so on a slim deployment a URL typo silently triggers a local ONNX model download — the exact cost this feature exists to avoid. Consider letting the error propagate (or re-raise as a dedicated misconfiguration error) so a malformed URL fails loudly.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+        connect=_CONNECT_TIMEOUT_SECONDS,
+        read=timeout_seconds,
+        write=timeout_seconds,
+        pool=_POOL_TIMEOUT_SECONDS,
+    )
+    data = {"zoomin": str(zoomin)}
+    mime_type = _mime_type_for(ext)
+
+    def _send(filename: str, fh: Any) -> dict[str, Any]:
+        with httpx.Client(timeout=timeout, transport=transport) as client:
```
rogercloud, line 185: A fresh `httpx.Client` is created and torn down for every `_post_document` call, so no connection is ever reused across parses. Not a correctness issue, but an implicit throughput boundary for the offload path: every parse pays TLS handshake + connection setup. Consider a shared module- or parser-level client with pooling, keeping per-call clients only where tests need them.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+            response = client.post(
+                f"{base_url}{PARSE_ENDPOINT}",
+                headers=_build_headers(),
+                data=data,
+                files={"file": (filename, fh, mime_type)},
+            )
+            response.raise_for_status()
```
rogercloud, line 192: `raise_for_status()` discards the response body, but the documented API contract (docstring above, lines 47-48) says 400/401/413/500 responses always carry a human-readable `detail`. Operators only see status + URL; the reason is lost. Read `response.json().get("detail")` in the `HTTPStatusError` handler and include it in the `DeepDocRemoteError` message.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+        )
+
+    normalized: list[dict[str, Any]] = []
+    for index, element in enumerate(elements):
+        if not isinstance(element, dict):
+            raise ValueError(
+                f"Remote DeepDoc element {index} is {type(element).__name__}, "
+                "expected an object"
+            )
+        missing = [key for key in ("type", "text") if key not in element]
```
rogercloud, line 249: Validation only checks key presence, so `{"type": "figure", "text": null}` passes and later fails as a raw `AttributeError` on `.strip()` (deepdoc.py:176), null text in a text element raises a Pydantic `ValidationError`, and malformed positions fail during int/float conversion. None of those are `DeepDocRemoteError`, so the caller (deepdoc.py:681-682) does not fall back — malformed remote output hard-aborts instead of triggering local parsing, breaking the feature's primary resilience contract. Validate types/values here (text coercible to str, numeric positions, known layout types) so every malformed-output failure surfaces as `DeepDocRemoteError`.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+            # them. Without it a corrupt-but-right-length payload decodes to
+            # b"" and gets written out as a zero-byte PNG, so the caller
+            # receives a broken image path rather than falling back to local
+            # parsing.
+            image_bytes = base64.b64decode(image_base64, validate=True)
+            if not image_bytes:
+                raise ValueError(
+                    f"Remote DeepDoc element {index} carries an empty image"
+                )
+            normalized_element["image"] = save_image(image_bytes)
```
rogercloud, line 269: Images are written to disk immediately during normalization; if a later element fails (corrupt base64, `OSError`), the already-written images remain under `data/artifacts/providers/deepdoc/<doc_id>/images` forever — there is no cleanup anywhere, and the local fallback writes into the same directory, so orphans are indistinguishable from real output. Repeated bad responses accumulate files. Consider buffering decoded images until the whole list normalizes and writing them transactionally, or clearing the `doc_id` image dir before the fallback.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def parse_document_remote(
+    file_path: str | BytesIO,
+    *,
+    ext: str,
+    save_image: SaveImage,
+    callback: Optional[ProgressCallback] = None,
+    zoomin: int = 3,
+    _transport: Optional[httpx.BaseTransport] = None,
```
rogercloud, line 284: `_transport` is a documented test-only seam (docstring line 299) but no test uses it — the sole caller passes `None`, and `httpx.Client(transport=None)` is exactly `httpx.Client()`. Drop the parameter and its threading through `_post_document`/`_send`.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Returns:
+        The parsed elements, each with an ``image`` path string or ``None``.
+
+    Raises:
+        DeepDocRemoteError: On any failure. Callers fall back to local parsing.
+    """
+    started = time.monotonic()
+    if callback is not None:
+        callback(0.05, "Uploading document to remote DeepDoc server")
```
rogercloud, line 309: This callback invocation runs outside the try/except that normalizes failures into `DeepDocRemoteError` — as does the completion notice at :339-341. `DeepDocProgressAdapter`'s exception handler re-invokes the same sink (adapters.py:40-58), so an always-raising sink escapes as a raw exception: an upload-notice failure prevents both remote and local parsing (the caller's `except DeepDocRemoteError` at deepdoc.py:681-682 never fires), and a completion-notice failure aborts after a successful parse, discarding valid results. The guard added at deepdoc.py:684-692 covers only the fallback notice. Move both callbacks inside the error boundary (or wrap them so callback failures become `DeepDocRemoteError`) to honor the module's "On any failure → DeepDocRemoteError" contract.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -227,6 +228,78 @@ def _translate_pdf_bboxes(
...
+        bbox: Dict[str, Any] = {
+            **element_metadata,
+            "layout_type": layout_type,
+            "text": element.get("text", ""),
+            "image": element.get("image"),
+        }
+        base_metadata = _build_element_metadata(bbox, doc_id, **kwargs)
+        # Carry the format-specific keys (style, sheet_name, row_number,
+        # row_type, ...) that _build_element_metadata does not know about.
+        for key, value in element_metadata.items():
```
rogercloud, line 275: This two-line setdefault loop is equivalent to a single expression: `base_metadata = {**element_metadata, **base_metadata}` (element_metadata wins, preserving the setdefault semantics).

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -527,12 +600,15 @@ def _sync_parse() -> ParseResult:
...
+            # deferred until a local branch actually needs one, because
+            # DeepDocPdfParser() eagerly loads ONNX models (and may download
+            # them from ModelScope) -- work that is pure waste in remote mode.
             if isinstance(file_path, BytesIO):
                 # For BytesIO objects, we need to determine file type from kwargs or use a default
                 ext = kwargs.get("file_ext", ".xlsx")  # Default to Excel for BytesIO
-                parser = self._get_parser_for_ext(ext)
             else:
-                parser, ext = self._get_parser(file_path)
+                ext = Path(file_path).suffix.lower()
```
rogercloud, line 611: `_get_parser` has zero callers after this PR — the only remaining entry point is `_get_parser_for_ext` (called at :712/:732/:740), and the PR body itself admits the dead code. Keeping a second entry point invites reintroducing eager instantiation, the exact regression the deferred-instantiation change removes. Delete the method; callers can derive `ext = Path(file_path).suffix.lower()` directly.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -552,6 +628,76 @@ def _sync_parse() -> ParseResult:
...
+
+                adapter = DeepDocProgressAdapter(progress_callback)
+                callback = adapter.get_callback()
+
+            # Remote-first dispatch: when a remote server is configured every
+            # supported format goes there, and any failure falls back to local
+            # parsing. No local parser is instantiated on the remote path.
+            if deepdoc_remote.is_remote_configured():
+                try:
+                    elements = deepdoc_remote.parse_document_remote(
```
rogercloud, line 669: The path→BytesIO conversion keeps only the extension, and the client uploads it as `document.xlsx`/`document.xls`/`document.csv` (deepdoc_remote.py:204-207). Parsing works because the extension is preserved, but the original filename is lost on the wire, degrading remote logs and diagnostics. Carry the original name through (e.g. alongside the BytesIO) and use it in the multipart upload.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -552,6 +628,76 @@ def _sync_parse() -> ParseResult:
                 **base_kwargs,
             }
 
+            # Validate Office document format (.docx, .xlsx, .pptx)
+            # Skip validation for BytesIO objects as they're already validated during conversion
+            # This magic-byte check is cheap and runs before any upload, so a
+            # malformed file fails fast instead of wasting remote bandwidth.
+            if ext in [".docx", ".xlsx", ".pptx"] and not isinstance(
```
rogercloud, line 635: All BytesIO inputs skip the magic-byte check here under the comment "already validated during conversion" — but the conversion at :589-591 only reads bytes; it validates nothing. A legacy/corrupt `.xlsx` is therefore uploaded before being rejected, despite the new comment claiming the check "runs before any upload". Since this PR moved the check ahead of remote dispatch, close the gap: validate the converted bytes (rewind + magic-byte check) before upload, or fix the comment and validate at the source.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -552,6 +628,76 @@ def _sync_parse() -> ParseResult:
...
+                        file_path,
+                        ext=ext,
+                        save_image=lambda image_bytes: _save_bytes_to_disk(
+                            doc_id, image_bytes, ".png"
+                        ),
+                        callback=callback,
+                        zoomin=parser_call_kwargs.get("zoomin", 3),
+                    )
+                    remote_metadata = {**metadata, "deepdoc_backend": "remote"}
+                    return _translate_remote_elements(
```
rogercloud, line 679: The remote-success path returns here, bypassing the `enable_raw_output` branch in the local path (:786-794). With `XAGENT_DEEPDOC_XINFERENCE_URL` set, `DeepDocParser(enable_raw_output=True)` returns `raw_parser_output=None`, `parser_engine=None`, `has_visualization_data=False` even for PDFs — a public, test-covered contract (test_deepdoc_parser.py:36-84). The remote response carries the bbox data (x0/x1/top/bottom/positions), but it is discarded. On the remote path, when `enable_raw_output` is set, attach the raw output from the translated/normalized elements so the flag means the same thing regardless of backend.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -552,6 +628,76 @@ def _sync_parse() -> ParseResult:
...
+                        try:
+                            callback(0.01, "Remote parse failed; running local DeepDoc")
+                        except Exception as callback_exc:
+                            logger.warning(
+                                "Progress callback failed while reporting the "
+                                "DeepDoc fallback: %s",
+                                callback_exc,
+                            )
+
+            metadata["deepdoc_backend"] = "local"
```
rogercloud, line 699: `deepdoc_backend` is only set on top-level metadata here (local path) and at :678 (remote path). The `.docx`/`.xlsx`/`.xls`/`.csv`/`.md`/`.txt`/`.json`/`.html` local branches build `ParseResult` without metadata (`{}`), and an empty local doc has no children either — so the field only appears for PDF/remote results. Nothing consumes it yet, but the inconsistency is a contract trap for the first consumer. Set metadata consistently in every branch, or document the field as remote/PDF-only and set it deliberately there.

rogercloud (changes requested):
## Summary

PR #1176 routes DeepDoc parsing through a new remote Xinference client (`XAGENT_DEEPDOC_XINFERENCE_URL`) with automatic local fallback: parser instantiation is deferred until actually needed, every supported format is uploaded to the configured endpoint, the returned flat element list is translated into XAgent's ParseResult model, and a local parser is constructed only when no URL is configured or the remote call fails. The diff (4 files, +616/-32) adds three env-var getters (URL, API key, timeout), a 347-line HTTP client with multipart upload and response/element normalization, and an 183-line dispatch/translation change in `DeepDocParser`. Config-getter tests are included; behavioral tests for the remote path and the fallback are deferred.

## Approach

**ACCEPTABLE-WITH-RESERVATIONS.** Deferred parser instantiation is the correct minimal mechanism, URL-gated remote-first dispatch fits #1165, and the reuse of the leaf helpers (`_process_table_element`, `_process_figure_element`, `_build_element_metadata`) is sound; the new fallback-notification guard is warranted but incomplete. Two reservations:

- **(a) Capability vs. behavior.** A class advertised as `LocalParsing` performs network egress behind a global env switch with no per-call opt-out, while the fallback is unconditional — a remote-only slim container will still load/download ONNX models on the first remote failure, contradicting the PR's own rationale. A cleaner long-term boundary is an explicit execution policy (`local` / `remote-with-local-fallback` / `remote-only`).
- **(b) Transport shape.** One whole-document synchronous HTTP request per executor task, a fresh client per document, and a monolithic base64-in-JSON response with non-transactional disk writes set an implicit throughput ceiling for the offload path.

## Findings

### Major

**1. `enable_raw_output` contract broken on the remote path** — `src/xagent/providers/pdf_parser/deepdoc.py:677-681`

The remote-success branch returns here via `_translate_remote_elements`, bypassing the `enable_raw_output` handling in the local path (deepdoc.py:786-794). With `XAGENT_DEEPDOC_XINFERENCE_URL` set, `DeepDocParser(enable_raw_output=True)` yields `raw_parser_output=None`, `parser_engine=None`, `has_visualization_data=False` even for PDFs — a public, test-covered contract (test_deepdoc_parser.py:36-84). The remote response carries the bbox data (x0/x1/top/bottom/positions), but it is discarded.
Suggestion: when `enable_raw_output` is set, attach the raw output from the translated/normalized elements on the remote path too, so the flag has the same meaning regardless of backend; add a test asserting the remote path honors it.

**2. Progress callbacks run outside the error normalization** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:308-310` and `:339-341`

Both the upload notice and the completion notice are invoked outside the try/except that converts failures into `DeepDocRemoteError`. `DeepDocProgressAdapter`'s exception handler re-invokes the same sink (adapters.py:40-58), so an always-raising sink escapes as a raw exception: an upload-notice failure prevents both remote and local parsing (the caller's `except DeepDocRemoteError` at deepdoc.py:681-682 never fires), and a completion-notice failure aborts after a successful parse, discarding valid results. The guard added at deepdoc.py:684-692 wraps only the fallback notice. This violates the module's own docstring contract — "On any failure → DeepDocRemoteError".
Suggestion: move both callback invocations inside the error boundary (or wrap them so callback failures become `DeepDocRemoteError`), matching the existing fallback-notice guard.

**3. Weak response validation defeats the fallback on malformed output** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:246-249`

Validation checks only key presence (`type`, `text`). `{"type":"figure","text":null}` passes and later raises a raw `AttributeError` on `.strip()` (deepdoc.py:176); null text in a text element raises a Pydantic `ValidationError`; malformed positions fail during int/float conversion. None of these are `DeepDocRemoteError`, so the caller's `except` (deepdoc.py:681-682) does not fall back — malformed remote output hard-aborts instead of triggering local parsing, breaking the feature's primary resilience contract.
Suggestion: validate types/values in `_normalize_elements` (text coercible to str, numeric positions, known layout types) so every malformed-output failure surfaces as `DeepDocRemoteError`.

**4. `LocalParsing` capability vs. network egress; unconditional fallback** — `src/xagent/providers/pdf_parser/deepdoc.py:519-521`, `document_parser.py:37-40` (design-level, body-only)

The class still advertises only `LocalParsing`, yet an env var switches it to network egress with no per-call opt-out, and `document_parser.py` documents `use_local_parser` as "local vs remote API-based parsing". Conversely, the fallback is unconditional (deepdoc.py:668-691): a remote-only slim container will still load/download ONNX models on the first remote failure, contradicting the PR's own rationale (deferred instantiation to avoid local model downloads). This is reservation (a).
Suggestion: introduce an explicit execution policy (`local` | `remote-with-local-fallback` | `remote-only`) — an env value or constructor arg — so capability advertisement matches behavior and remote-only deployments never instantiate the local parser.

**5. No committed behavior tests** — `tests/core/test_config.py:1221-1292` (the only new tests) (body-only)

The 347-line client and 183-line dispatch change ship with zero committed behavioral coverage: nothing exercises the remote client, the remote dispatch, or the fallback. S5 tests are deferred to #1166, but the fallback path is the feature's core contract and is currently unguarded — findings 1 and 3 show real regressions of exactly that kind.
Suggestion: commit at least a fallback-dispatch test (mock client raising `DeepDocRemoteError` → local result, `enable_raw_output` preserved) and one remote happy-path translation test before merge, or gate the merge explicitly on #1166.

**6. Malformed URL swallowed, indistinguishable from unset** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:119-133`, `src/xagent/config.py:1859-1869`

`is_remote_configured()` catches the `ValueError` the getter raises on a malformed URL, logs a warning, and returns `False` — so unset and malformed are indistinguishable, directly contradicting the getter docstring added in the same PR ("a malformed value raises rather than silently downgrading to local parsing"). On slim deployments, a URL typo silently triggers an ONNX model download — the exact cost this feature exists to avoid.
Suggestion: let the `ValueError` propagate (or re-raise as a dedicated misconfiguration error) so a malformed URL fails loudly at first parse instead of degrading silently.

### Minor

**7. `.xlsx` uploads skip the magic-byte check** — `src/xagent/providers/pdf_parser/deepdoc.py:635-636` (conversion at :589-591)

`.xlsx`/`.xls`/`.csv` paths are converted to `BytesIO` at :589-591 with no validation, and the check at :635-636 skips validation for **all** `BytesIO` under an incorrect comment ("already validated during conversion" — conversion only reads bytes). The new comment claims the check "runs before any upload", but a legacy/corrupt `.xlsx` is uploaded before rejection. This PR moved the check ahead of remote dispatch, making the gap consequential.
Suggestion: run the magic-byte validation on the converted bytes before upload (or validate at the source and fix the comment).

**8. Orphaned images on partial failure** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:269`

Each image is written to disk immediately during normalization; if a later element fails (corrupt base64, `OSError`), the already-written images remain in `data/artifacts/providers/deepdoc/<doc_id>/images` forever — there is no cleanup anywhere, and the local fallback writes into the same directory, so orphans are indistinguishable from real output. Repeated bad responses accumulate files.
Suggestion: buffer decoded images until the whole element list normalizes successfully and write them transactionally, or clear the `doc_id` image directory before the local fallback writes.

**9. `deepdoc_backend` metadata is format-dependent** — `src/xagent/providers/pdf_parser/deepdoc.py:678, 699`

The field is set on top-level metadata only on the remote path and the local PDF/unknown-ext paths; the `.docx`/`.xlsx`/`.xls`/`.csv`/`.md`/`.txt`/`.json`/`.html` local branches build `ParseResult` without metadata (`{}`), and an empty local doc has no children either. Nothing consumes the field yet — a pure contract inconsistency that will bite the first consumer.
Suggestion: populate metadata consistently in every branch, or document the field as remote/PDF-only and set it deliberately there.

**10. Server error detail dropped** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:192`

`raise_for_status()` discards the documented `{"detail": ...}` body (docstring :47-48: 400/401/413/500 responses always carry a human-readable detail). Operators see only status + URL; the reason is lost (the status code itself is preserved).
Suggestion: on `HTTPStatusError`, read `response.json().get("detail")` and include it in the `DeepDocRemoteError` message.

**11. Dead code `_get_parser`** — `src/xagent/providers/pdf_parser/deepdoc.py:552-555`

`_get_parser` has zero callers after this PR (only `_get_parser_for_ext` calls remain at :712/:732/:740); the PR body admits it. A second, dead entry point invites reintroducing eager instantiation — the exact regression the deferred-instantiation change removes. Delete the method (see also Simplification below).

**12. Dangling documentation reference** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:17-19`

The module docstring cites `docs/deepdoc-remote-parsing.md`, but that file exists only on the sibling umbrella branch (review/pr-1178), not in this PR or in the base. The three new env vars, the timeout, egress behavior, and the fallback policy are undocumented anywhere in this PR.
Suggestion: land the docs with the code (or point the citation at the tracking issue/PR) and document the env vars plus the fallback policy.

**13. Fresh `httpx.Client` per document** — `src/xagent/providers/pdf_parser/deepdoc_remote.py:184-185`

A context-managed client is created (and torn down) for every `_post_document` call; no connection is ever reused across parses. Not a correctness issue, but an implicit throughput boundary for the offload path — every parse pays TLS handshake + connection setup, which is the whole point of offloading. This is part of reservation (b).
Suggestion: hold a shared module- or parser-level `httpx.Client` with a sensible pool, and keep per-call clients only where tests need them.

**14. Duplicated translation dispatch loop** — `src/xagent/providers/pdf_parser/deepdoc.py:231-305` (body-only)

`_translate_remote_elements` duplicates the structure of `_translate_pdf_bboxes` (:188-228): the same dispatch, unknown-type fallback, logging, and `ParseResult` construction; only the bbox reshape and the extra-metadata carry are remote-specific. Future element types must be added twice.
Suggestion: generalize `_translate_pdf_bboxes` (e.g. accept a per-element metadata-enrichment hook) and make the remote translator a thin adapter over it.

**15. Synthetic filenames for spreadsheets** — `src/xagent/providers/pdf_parser/deepdoc.py:589-591`, `deepdoc_remote.py:204-207`

The path→BytesIO conversion keeps only the extension, and the client uploads the file as `document.xlsx`/`document.xls`/`document.csv`. Extension is preserved, so parsing works, but the original filename is lost on the wire — remote logs and diagnostics degrade.
Suggestion: carry the original filename through (alongside the BytesIO) and use it in the multipart upload.

- `L552`: delete `_get_parser` — dead code, merged with finding 11; remove the method. (~5 lines)
- `L275`: shrink `for key, value in element_metadata.items(): base_metadata.setdefault(key, value)` → `base_metadata = {**element_metadata, **base_metadata}`. (-1 line)
- `L284`: delete `_transport` — documented test-only seam with no test using it (sole caller passes `None`; `httpx.Client(transport=None)` ≡ `httpx.Client()`); drop the parameter and its threading through `_post_document`/`_send`. (~4 lines)

Net: ~-10 lines.

## Tests

Run in a review worktree (Python 3.12) against head 05085f9a:

- `tests/providers/pdf_parser/test_deepdoc_table_positions.py`: **18 passed** — matches the PR's regression claim.
- `tests/providers/pdf_parser`: **61 passed, 1 failed, 2 skipped**. The single failure is `test_basic_parsers_pdf_success[PyMuPdfParser]` — `ModuleNotFoundError: No module named 'fitz'` (missing optional dependency; `PyMuPdfParser` is untouched by this PR, and the PR body reports the identical failure on base).
- `tests/core/test_config.py -k DeepDoc`: **20 passed** (the PR's new config-getter tests).
- The full `tests/providers` suite cannot be collected in the review environment: missing optional dependencies `chromadb`/`pymilvus` (unrelated to this PR).

OliverBryant:

PR #1177, opened 2026-08-06
test(deepdoc): cover remote routing, local fallback, and the element translator

### LINKED ISSUE #1166: test(deepdoc): cover remote routing, local fallback, and the element translator

OliverBryant:
Closes #1166 · Step **S5** of #1161

## Stacking

Stacks on #1176 (which itself stacks on #1173 → #1172), so the diff here replays those PRs' files. **Only `tests/providers/pdf_parser/test_deepdoc_remote_fallback.py` is new in this PR** — one file, 919 lines. Merge the parents first and this collapses to just the test file.

## What

A new pure unit-test module for the remote-first dispatch added in #1176. No network, no DeepDoc model cache, no writes outside `tmp_path` — it runs anywhere. 18 tests in four groups.

**Remote routing (4 tests).** The load-bearing assertion of the whole feature: remote mode must never instantiate a local parser, because `DeepDocPdfParser()` eagerly loads ONNX models and may download them from ModelScope. `DeepDocParser._get_parser_for_ext` is armed as a tripwire that raises `AssertionError`, and `parse_document_remote` is monkeypatched to return canned elements. The parse must still succeed and report `deepdoc_backend=remote`. Covered for `.pdf`, `.docx`, `.txt`, and an in-memory `.xlsx` `BytesIO` — remote routing is deliberately not PDF-only.

**Local fallback (3 tests).**
- `DeepDocRemoteError` → the local result is returned, the warning is in `caplog`, metadata carries `deepdoc_backend=local`, and the stub local parser really was called with the local signature.
- A **broken progress sink cannot abort the fallback.** `DeepDocProgressAdapter.get_callback()` calls `on_status_update` again from inside its own `except` handler (`adapters.py:70`), so a sink that always raises makes the adapter's callback re-raise. This test regression-guards the `try/except` #1176 wrapped around the fallback notification; the mutation check below shows it goes red without it.
- The `.xlsx` fallback reaches the openpyxl row reader rather than a DeepDoc parser.

**Env unset (2 tests).** `parse_document_remote` armed as a tripwire; the pure-local path must be taken. Also covers a malformed URL degrading to local with a warning instead of failing every parse.

**Translator (9 tests).** `_translate_remote_elements` output is compared field-by-field against `_translate_pdf_bboxes` for equivalent PDF-shaped input (text + table + figure, with `positions`, `col_id`, `page_number`) — segments, tables, figures, and all their metadata dicts must be equal. Plus: xlsx-shaped elements carry `sheet_name`/`row_number`/`row_type` and emit no `positions`; `positions` enrichment inserts `col_id` at index 1 (`[[1,10,20,30,40]]` → `[[1, 0, 10.0, 20.0, 30.0, 40.0]]`) including for a non-zero `col_id`; an unknown future `type` degrades to a text segment rather than being dropped; missing / `None` / non-dict `metadata` does not blow up; and the client's saved-image path strings flow through onto table and figure `image_path`.

An autouse fixture patches `deepdoc.ARTIFACTS_DIR` to `tmp_path` so nothing can write into the user's real `~/.xagent`, and another autouse fixture `delenv`s `XAGENT_DEEPDOC_XINFERENCE_URL` so a developer `.env` cannot silently flip the routing decision under test.

## Why

The routing decision and the fallback are the two places this feature can fail silently and expensively: a regression that moves local-parser construction back above the remote dispatch would cost every remote user a full ONNX model load (possibly a ModelScope download) on every parse, while still producing correct output — invisible to any output-only assertion. Tripwires are the only way to catch that.

`tests/providers/pdf_parser/test_deepdoc_parser.py` is deliberately untouched: it runs real inference in the CI `pytest-fast-deepdoc` job with a warm cache, and now doubles as the end-to-end regression for "env unset → local".

## Testing

`.venv` is Python 3.12; `PYTHONPATH` points at this worktree's `src`.

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_remote_fallback.py -q
..................                                                       [100%]
18 passed
```

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_table_positions.py -q
..................                                                       [100%]
18 passed
```

Whole provider suite — the single failure is pre-existing on the base commit (`PyMuPdfParser`, missing `fitz`):

```
$ .venv/bin/python -m pytest tests/providers -q
........................Fs.............s................................ [ 66%]
....................................                                     [100%]
E   ModuleNotFoundError: No module named 'fitz'
FAILED tests/providers/pdf_parser/test_basic_parsers.py::test_basic_parsers_pdf_success[PyMuPdfParser]
```

Lint / format / import order:

```
$ .venv/bin/python -m ruff format --check tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
1 file already formatted

$ .venv/bin/python -m ruff check tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
All checks passed!

$ uvx --quiet isort==6.0.1 --check-only --diff tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
(no output — clean)
```

### Mutation checks

A tripwire that passes either way proves nothing, so both guarantees were verified by temporarily breaking the production code and confirming the tests go red. Neither mutation is committed — `git diff` on `deepdoc.py` was confirmed empty after each restore.

**1. Revert the deferred-instantiation change** (move `_get_parser_for_ext` back to the top of `_sync_parse`, the pre-#1176 shape):

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_remote_fallback.py -q --tb=no
FFFF..F...........                                                       [100%]
FAILED ...::TestRemoteRoutingSuccess::test_pdf_goes_remote_without_local_parser
FAILED ...::TestRemoteRoutingSuccess::test_docx_goes_remote_without_local_parser
FAILED ...::TestRemoteRoutingSuccess::test_remote_success_reports_progress
FAILED ...::TestRemoteRoutingSuccess::test_remote_bytesio_goes_remote_without_local_parser
FAILED ...::TestRemoteFailureFallsBackToLocal::test_remote_error_falls_back_to_local_xlsx
```

with the tripwire's own message:

```
AssertionError: _get_parser_for_ext('.pdf') was called on the remote path;
remote mode must never instantiate a local DeepDoc parser
```

**2. Remove the `try/except` around the fallback progress notification:**

```
$ .venv/bin/python -m pytest tests/providers/pdf_parser/test_deepdoc_remote_fallback.py -q -k broken_progress_sink --tb=line
F                                                                        [100%]
E   xagent.providers.pdf_parser.deepdoc_remote.DeepDocRemoteError: Remote DeepDoc request failed: boom
    During handling of the above exception, another exception occurred:
E   RuntimeError: progress sink is broken
FAILED ...::TestRemoteFailureFallsBackToLocal::test_broken_progress_sink_does_not_abort_the_fallback
```

Confirming the fallback is aborted outright without the guard, exactly as #1176 described.

**3. Sanity check that the metadata-carry assertions aren't vacuous** — neutering the `base_metadata.setdefault(key, value)` loop in `_translate_remote_elements` turns 3 tests red:

```
.F.F........F.....                                                       [100%]
FAILED ...::TestRemoteRoutingSuccess::test_docx_goes_remote_without_local_parser
FAILED ...::TestRemoteRoutingSuccess::test_remote_bytesio_goes_remote_without_local_parser
FAILED ...::TestTranslateRemoteElements::test_xlsx_shaped_elements_carry_row_metadata
```

## Notes

No production code is modified by this PR. `DeepDocParser._get_parser()` remains dead code after #1176 (flagged there for a follow-up); this PR does not touch it, and the tests patch `_get_parser_for_ext` directly so they stay correct once it is removed.


tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+
+        ``DeepDocProgressAdapter.get_callback()`` re-raises out of its own
+        ``except`` handler when the sink misbehaves, so without the caller's
+        try/except around the fallback notification this test fails with
+        ``RuntimeError`` instead of returning the local result.
+        """
+        pdf_file = tmp_path / "broken_sink.pdf"
+        pdf_file.write_bytes(b"%PDF-1.4 not really a pdf")
+
+        def failing_parse_document_remote(*args: Any, **kwargs: Any) -> Any:
```
rogercloud, line 461: This fake raises `DeepDocRemoteError` without ever invoking the progress callback, so the test exercises none of the real failure path. In production, `parse_document_remote` calls the callback *outside* its try (`deepdoc_remote.py:308-309`, `:339-340`), and `DeepDocProgressAdapter` re-invokes the broken sink from its `except` branch (`src/xagent/core/tools/core/RAG_tools/progress/adapters.py:64-70`) → `RuntimeError`; `_sync_parse` only catches `DeepDocRemoteError` (`deepdoc.py:682`). With a real client, `RuntimeError('progress sink is broken')` escapes `parse()` and the fallback never happens (sink `call_count == 4`) — so a fully broken sink actually aborts the fallback, and this test is green only because the mock removes exactly the layer that fails. Suggest calling the callback before raising, or driving the real client with a failing `httpx.MockTransport`, so the "fallback still happens" claim is actually exercised.

tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+            "E = mc^2",
+            "Ordinary text",
+        ]
+        # The unknown type is preserved in metadata rather than rewritten.
+        assert result.text_segments[0].metadata["layout_type"] == "equation"
+        assert result.text_segments[0].metadata["page_number"] == 7
+        assert result.tables == []
+        assert result.figures == []
+
+    def test_missing_and_none_metadata_do_not_blow_up(self) -> None:
```
rogercloud, line 828: This group only exercises malformed *metadata* (absent/`None`/non-dict) and uses `text=''` for the figure; it never feeds `text=None`. Production passes `element.get('text', '')` through unchanged (`deepdoc.py:269-273`), so `text=None` hits `_process_figure_element`'s `.strip()` (AttributeError) or `ParsedTextSegment(text: str)` (ValidationError, `base.py:50-53`) — and `_sync_parse` catches only `DeepDocRemoteError` (`deepdoc.py:669-686`), so neither crash falls back to local. Both were reproduced; this violates the always-fallback contract the suite is supposed to pin down (same hole as the outstanding P1 comment). Suggest adding `text=None` / wrong-type elements and asserting local fallback at the routing layer.

tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+    async def test_remote_error_falls_back_to_local_xlsx(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        tmp_path: Path,
+        remote_configured: str,
+    ) -> None:
+        """The .xlsx fallback reaches the openpyxl row reader, not a DeepDoc parser."""
+        arm_local_parser_tripwire(monkeypatch)
+
+        def failing_parse_document_remote(*args: Any, **kwargs: Any) -> Any:
```
rogercloud, line 504: This fake raises before touching the buffer, so the remote client's cursor-recovery logic (`deepdoc_remote.py:201-209`) never runs; the local fake (L513-514) ignores `file_path` and only records kwargs; the assertions (L533-534) check just `backend == "local"` plus canned text. If `_sync_parse` ever passed the wrong object (bytes, a copy, or a consumed stream), this test would stay green while production crashes. Production `_parse_xlsx_rows` does `seek(0)` itself (`deepdoc.py:849`), which covers the cursor half, but "the same, readable object was passed" is still unasserted. Suggest seeking/reading to EOF in the remote fake before raising, and asserting object identity/content in the local fake.

rogercloud (changes requested):
Two blocking test-quality issues: the broken-progress-sink test mocks away exactly the layer that fails in production (the fallback never happens with a real broken sink), and the translator robustness group misses `text=None`, which crashes production without falling back. Both were reproduced against the real code. Non-blocking: F4 (buffer readability unasserted) and F3 (nit, conflict precedence). Details and 3 inline comments in the review body.

OliverBryant:

Issue #1168, opened 2026-08-06
chore(deepdoc): final regression and manual smoke test for remote parsing


OliverBryant:
Part of #1161 · Step **S7** · depends on: S1–S6

Final regression, static checks, and the manual smoke test against a real Xinference endpoint.

## Automated

```
ruff format . && ruff check . && mypy src/xagent
pytest tests/core/test_config.py tests/providers/pdf_parser/ -q
```

The `tests/providers/pdf_parser/` run needs a warm DeepDoc model cache for `test_deepdoc_parser.py` (real inference); the CI `pytest-fast-deepdoc` job provides one. The new tests from S3 and S5 must pass without any cache.

## Manual smoke test

This depends on the Xinference-side endpoint existing, so it does not block the code PRs.

1. Export the three variables against a real GPU host running the DeepDoc parse endpoint.
2. Ingest a PDF, a DOCX, and an XLSX through the KB flow. Confirm the result metadata carries `deepdoc_backend=remote`, that no ModelScope download is triggered, and that chunking, embedding, and retrieval behave as before.
3. Stop the remote service and re-ingest. Confirm the warning is logged, the progress notice appears, parsing succeeds via the local fallback, and metadata carries `deepdoc_backend=local`.
4. Set a malformed URL. Confirm it degrades to local parsing with a warning and no parse fails.

## Done when

All automated checks pass and the four manual scenarios behave as described in the acceptance criteria of `docs/deepdoc-remote-parsing.md`.

rogercloud:
Closing as superseded by the final design and validation merged in #1271. That PR records live testing against a real GPU-backed Xinference endpoint, remote PDF success, local handling for non-PDF formats, remote-failure fallback, and malformed-URL fallback. The original checklist assumed remote DOCX/XLSX parsing, while the final design deliberately limits the remote path to PDFs.

Issue #1167, opened 2026-08-06
docs(deepdoc): document remote Xinference parsing configuration


OliverBryant:
Part of #1161 · Step **S6** · depends on: nothing (can land in parallel)

Document the feature. The requirements doc is added by the umbrella PR; this issue covers keeping it accurate and adding the `example.env` entry.

## Scope

- `docs/deepdoc-remote-parsing.md` — verify it matches what actually shipped once S1–S5 land: env var names and defaults, the flow diagrams, the server API contract, and the acceptance criteria.
- `example.env` — add a banner section in the file's established style (banner comment, explanatory lines mentioning defaults and units, then commented-out example assignments):

```
# ===========================================
# DeepDoc Remote Inference (Xinference)
# ===========================================
# Offload DeepDoc document parsing (OCR/layout/table models) to a remote
# Xinference server with a GPU. When set, ALL DeepDoc-supported formats are
# parsed remotely; when unset, parsing runs locally. On remote failure xagent
# logs a warning and falls back to local parsing automatically.
# XAGENT_DEEPDOC_XINFERENCE_URL="http://gpu-host:9997"
# API key; falls back to XINFERENCE_API_KEY if unset.
# XAGENT_DEEPDOC_XINFERENCE_API_KEY=""
# Request timeout in seconds for one whole-document parse (default: 1800).
# XAGENT_DEEPDOC_XINFERENCE_TIMEOUT_SECONDS="1800"
```

`CLAUDE.md` requires new configuration to be reflected in `example.env` and covered by `tests/core/test_config.py`; the latter is S1's job. No frontend or README change is needed.

## Done when

`example.env` documents all three variables and the requirements doc matches the shipped behavior.

PR #1178, opened 2026-08-06
feat(deepdoc): offload document parsing to a remote Xinference GPU server

OliverBryant:
Umbrella PR for #1161 — the complete DeepDoc remote-parsing feature on one branch, verified end to end.

Setting `XAGENT_DEEPDOC_XINFERENCE_URL` routes every format `DeepDocParser` supports to a GPU host running DeepDoc through Xinference, instead of local ONNX inference. Any remote failure falls back to local parsing automatically. No new `ParseMethod`, no frontend change, no capability-marker change.

## Why

DeepDoc's PDF pipeline (OCR detection/recognition, layout analysis, table structure recognition) runs entirely as local ONNX inference and dominates ingestion cost on CPU-only deployments — large documents take minutes.

Two constraints shaped the design:

- `deepdoc-lib` 0.2.2 has no usable remote inference path. The `TENSORRT_DLA_SVR` hook in `vision/layout_recognizer.py` imports `deepdoc.vision.dla_cli`, a module that does not ship in the package.
- Xinference has no DeepDoc serving capability yet, so the server API in `docs/deepdoc-remote-parsing.md` section 6 is a proposal to be implemented separately.

## What is here

| File | Change |
|---|---|
| `src/xagent/config.py` | +38 — three `XAGENT_DEEPDOC_XINFERENCE_*` getters reusing `_normalized_http_env_url` / `_get_positive_int_env` |
| `src/xagent/providers/pdf_parser/deepdoc_remote.py` | +347 — new HTTP client; every failure funnels into one `DeepDocRemoteError` |
| `src/xagent/providers/pdf_parser/deepdoc.py` | +151 — remote-first dispatch, deferred local instantiation, `_translate_remote_elements` |
| `tests/core/test_config.py` | +80 — config getter coverage |
| `tests/providers/pdf_parser/test_deepdoc_remote.py` | +763 — 73 client tests, mocked HTTP |
| `tests/providers/pdf_parser/test_deepdoc_remote_fallback.py` | +919 — 18 routing/fallback/translator tests |
| `docs/deepdoc-remote-parsing.md` | +153 — requirements, flow diagrams, server API contract |
| `example.env` | +15 — the three variables |

## Configuration

| Variable | Required | Default |
|---|---|---|
| `XAGENT_DEEPDOC_XINFERENCE_URL` | to enable remote | unset (= local mode) |
| `XAGENT_DEEPDOC_XINFERENCE_API_KEY` | no | falls back to bare `XINFERENCE_API_KEY`, then no auth header |
| `XAGENT_DEEPDOC_XINFERENCE_TIMEOUT_SECONDS` | no | `1800` |

There is deliberately no fallback toggle — fallback is always on, which is what makes the switch transparent.

### Automated

```
pytest tests/core/test_config.py tests/providers/pdf_parser/  →  504 passed
ruff check .                                                  →  All checks passed
isort (changed files, 6.0.1 per pre-commit)                    →  clean
codespell (changed files, 2.4.1)                               →  clean
```

Pre-existing on `main` and unrelated to this branch: `PyMuPdfParser` fails for a missing `fitz`, and `ruff format` flags four files (`web/api/kb.py` plus three tests).

### End-to-end against a live HTTP server

A stub server implementing the section-6 contract was run on `127.0.0.1:9997` and a real PDF pushed through `DeepDocParser.parse()`:

- **Remote success** — `_get_parser_for_ext` was armed as a tripwire raising `AssertionError`; it never fired, proving no ONNX load and no ModelScope download. Result: `deepdoc_backend=remote`, 1 text segment + 1 table + 1 figure, `full_text` populated.
- **Real upload** — the server received a 196,367-byte multipart body with the `file` part, the `zoomin` field, and `Authorization: Bearer test-key-abc`.
- **Real image writes** — both table and figure images decoded from base64 and landed on disk at 70 bytes each under `artifacts/providers/deepdoc/{doc_id}/images/`.
- **Position enrichment** — `[[1,10,500,100,130]]` became `[[1, 0, 10.0, 500.0, 100.0, 130.0]]`, inserting `col_id` at index 1 as the local path does.
- **Fallback to real local inference** — with the server stopped, the same PDF parsed via actual local ONNX in 9.1s: `deepdoc_backend=local`, 48 text segments, 2 tables.
- **Malformed URL degrades** — `ftp://host`, a URL carrying a query string, and an unset variable all yield `is_remote_configured() == False` without raising.

### Mutation checks — the suites are not vacuous

- Reverting the deferred-instantiation change turned **5** tripwire tests red.
- Removing the fallback's progress-notification guard turned `test_broken_progress_sink_does_not_abort_the_fallback` red.

## Two real defects found and fixed during review

**Silent image corruption.** `base64.b64decode` without `validate=True` discards non-alphabet characters, so a corrupt-but-right-length payload such as `"!!!!"` decoded to `b""` and was written out as a zero-byte PNG — handing the caller a broken image path instead of falling back. Now validated, and an image decoding to no bytes is rejected.

**Fallback defeated by a broken progress sink.** `DeepDocProgressAdapter.get_callback()` calls `on_status_update` again *inside its own except handler* (`adapters.py:70`), so a sink that raises propagates out of the callback. An unguarded fallback notification would therefore turn a recoverable remote failure into a hard parse failure. The notification is now wrapped in its own `try/except`; the adapter itself is untouched and remains the real fix site.

## Sub-issue breakdown

Each step also has its own PR against `main`, which is where the detailed per-step review happened:

| Step | Sub-issue | PR |
|---|---|---|
| S1 config | #1162 | #1172 |
| S2 client | #1163 | #1173 |
| S3 client tests | #1164 | #1175 |
| S4 routing | #1165 | #1176 |
| S5 routing tests | #1166 | #1177 |
| S6 docs | #1167 | #1170 |

Merge either this branch as one unit, or the six in the order above — the content is identical. S7 (#1168) covers the final regression plus the manual smoke test against a real Xinference endpoint, which stays open until that endpoint exists.

## Follow-up

`DeepDocParser._get_parser()` became dead code once `_sync_parse` stopped calling it. Left in place to keep the diff focused.


src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+        _transport: Test-only transport override.
+
+    Returns:
+        The parsed elements, each with an ``image`` path string or ``None``.
+
+    Raises:
+        DeepDocRemoteError: On any failure. Callers fall back to local parsing.
+    """
+    started = time.monotonic()
+    if callback is not None:
```
rogercloud, line 308: V1: The start callback is invoked outside the try block (which only covers `_post_document` + `_normalize_elements`, lines 311-336), and the same applies to the completion callback at lines 339-340. `DeepDocProgressAdapter.get_callback()` re-invokes the sink inside its own except handler, so a raising progress sink escapes raw; `deepdoc.py:682` catches only `DeepDocRemoteError`, so this hard-fails the parse with NO local fallback (start: request never sent; complete: successful remote result discarded). You guard exactly this hazard at the fallback notification (deepdoc.py:690-697) but not here. Fix: move both callbacks inside the protected region so a broken sink degrades to `DeepDocRemoteError` like every other failure.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+    elements = payload.get("elements")
+    if not isinstance(elements, list):
+        raise ValueError(
+            "Remote DeepDoc response is missing an 'elements' list "
+            f"(got {type(elements).__name__})"
+        )
+
+    normalized: list[dict[str, Any]] = []
+    for index, element in enumerate(elements):
+        if not isinstance(element, dict):
```
rogercloud, line 244: V2: `_normalize_elements` validates only presence (element is a dict, type/text keys exist, image decodes), never value types or the documented required metadata (PDF positions, page_number). Malformed bodies either hard-fail with NO fallback — `{"type":"figure","text":null}` → AttributeError in the translators (`deepdoc.py:178`), `{"type":"text","text":null}` → pydantic ValidationError (`deepdoc.py:288-289`), non-numeric positions → ValueError/TypeError (`deepdoc.py:118-123`) — all escaping the `except DeepDocRemoteError` at `deepdoc.py:682`; or silently succeed degraded (missing positions → page=1/col=0 defaults, `deepdoc.py:100-127`). Since the server protocol is a proposal, malformed payloads are the likely first-version reality. Fix: validate types and required keys per element type here and raise ValueError (already mapped to DeepDocRemoteError) for anything the translators cannot represent faithfully.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -227,6 +228,78 @@ def _translate_pdf_bboxes(
...
+            "text": element.get("text", ""),
+            "image": element.get("image"),
+        }
+        base_metadata = _build_element_metadata(bbox, doc_id, **kwargs)
+        # Carry the format-specific keys (style, sheet_name, row_number,
+        # row_type, ...) that _build_element_metadata does not know about.
+        for key, value in element_metadata.items():
+            base_metadata.setdefault(key, value)
+
+        if layout_type == "table":
```
rogercloud, line 278: V5: Remote `type="table"` elements land in `ParseResult.tables`, but the local Markdown translator (`deepdoc.py:480-481`) puts tables into `text_segments` with `metadata.type="table"`. `full_text` is composed ONLY from `text_segments` (`base.py:348-350`), so the same .md yields different `full_text` depending on the env var — table content silently disappears from `full_text` in remote mode, and consumers write `result.full_text` verbatim (`document_parser.py:419-420`). Fix: unify element classification or include tables in `full_text` consistently, and add a local/remote equivalence test asserting identical `text_segments`/`tables`/`full_text`.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -552,6 +628,76 @@ def _sync_parse() -> ParseResult:
...
+                        try:
+                            callback(0.01, "Remote parse failed; running local DeepDoc")
+                        except Exception as callback_exc:
+                            logger.warning(
+                                "Progress callback failed while reporting the "
+                                "DeepDoc fallback: %s",
+                                callback_exc,
+                            )
+
+            metadata["deepdoc_backend"] = "local"
```
rogercloud, line 699: V6: `metadata["deepdoc_backend"]="local"` is applied unconditionally — even when the env var is UNSET and parsing is purely local. This violates the documented acceptance "env unset = byte-for-byte current behavior" (`docs/deepdoc-remote-parsing.md:141-142`), and the PR's own test enshrines the deviation. The key is additive and parse_hash is unaffected, but it persists into element metadata (LanceDB). Fix: add the marker only when a remote attempt occurred, or amend the compatibility promise in the docs.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+            # them. Without it a corrupt-but-right-length payload decodes to
+            # b"" and gets written out as a zero-byte PNG, so the caller
+            # receives a broken image path rather than falling back to local
+            # parsing.
+            image_bytes = base64.b64decode(image_base64, validate=True)
+            if not image_bytes:
+                raise ValueError(
+                    f"Remote DeepDoc element {index} carries an empty image"
+                )
+            normalized_element["image"] = save_image(image_bytes)
```
rogercloud, line 269: V7: `save_image` runs inside the element loop; if element N+1 is invalid, the images already written for elements 1..N stay on disk permanently under `artifacts/providers/deepdoc/<doc_id>/images/` (uuid4 filenames — retries accumulate orphans; no cleanup exists anywhere in the repo). No correctness effect (fallback writes its own images), but it is unbounded disk litter. Fix: validate/decode all elements first, then write images; or track and clean up written paths on failure.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,347 @@
...
+            )
+        return payload
+
+    if isinstance(file_path, BytesIO):
+        # The buffer is owned by the caller, who may still read from it after a
+        # fallback, so put the cursor back where it was found.
+        original_position = file_path.tell()
+        try:
+            file_path.seek(0)
+            return _send(f"document{ext}", file_path)
```
rogercloud, line 207: V8: The remote client uploads a synthesized filename `document{ext}` instead of the real basename — server-observable via multipart Content-Disposition and contradicting the documented "filename preserved" (`docs/deepdoc-remote-parsing.md:96`). Root cause is upstream: path-form .xlsx/.xls/.csv are fully read into BytesIO before the remote-routing decision (`deepdoc.py:578-597`, pre-existing conversion made observable here), which also skips `validate_office_file_format` (`deepdoc.py:631-640` guards on `not isinstance(file_path, BytesIO)`; the added comment "already validated during conversion" is false). Fix: defer the BytesIO conversion to the local branch and pass the real path to the remote client.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -527,12 +600,15 @@ def _sync_parse() -> ParseResult:
...
+            # deferred until a local branch actually needs one, because
+            # DeepDocPdfParser() eagerly loads ONNX models (and may download
+            # them from ModelScope) -- work that is pure waste in remote mode.
             if isinstance(file_path, BytesIO):
                 # For BytesIO objects, we need to determine file type from kwargs or use a default
                 ext = kwargs.get("file_ext", ".xlsx")  # Default to Excel for BytesIO
-                parser = self._get_parser_for_ext(ext)
             else:
-                parser, ext = self._get_parser(file_path)
+                ext = Path(file_path).suffix.lower()
```
rogercloud, line 611: V9: This replacement of `parser, ext = self._get_parser(file_path)` with suffix-only lookup removes the last caller of `_get_parser` (`deepdoc.py:552-555`) — it is now dead code (zero callers in src/ and tests/). It eagerly instantiates the parser (ONNX model load, possibly a ModelScope download) and is a trap for future refactors that reintroduce eager local loading on the remote path. Trivially safe to delete.

tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+        assert len(fake_parser.calls) == 1
+        assert fake_parser.calls[0]["file_path"] == str(pdf_file)
+        assert fake_parser.calls[0]["zoomin"] == 3
+
+        # And the failure was logged rather than swallowed.
+        assert "Remote DeepDoc parse failed" in caplog.text
+        assert "falling back to local" in caplog.text
+
+    @pytest.mark.asyncio
+    async def test_broken_progress_sink_does_not_abort_the_fallback(
```
rogercloud, line 444: V10: This test monkeypatches `deepdoc_remote.parse_document_remote` wholesale with a fake that raises `DeepDocRemoteError` and never calls the callback, so it only exercises the try/except around the fallback notification (`deepdoc.py:690-697`) — it CANNOT catch the V1 bug, where the real client's start/complete callbacks sit outside the try boundary. The PR cites this as load-bearing mutation-check coverage, which is false confidence. Fix: keep the real `parse_document_remote`, mock only `_post_document`/transport, and have the sink raise at start, complete, and fallback notifications.

tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+        def failing_parse_document_remote(*args: Any, **kwargs: Any) -> Any:
+            raise DeepDocRemoteError("Remote DeepDoc returned an unusable response")
+
+        monkeypatch.setattr(
+            deepdoc_remote, "parse_document_remote", failing_parse_document_remote
+        )
+
+        recorded: Dict[str, Any] = {}
+
+        def fake_parse_xlsx_rows(file_path: Any, **kwargs: Any) -> Any:
```
rogercloud, line 513: V11: This fake injects `ParseResult(metadata=dict(kwargs))`, which the real `_parse_xlsx_rows` (`deepdoc.py:913`) does not do. The current assertions never consult `result.metadata`, and what they do assert (`recorded["deepdoc_backend"]=="local"`) mirrors real behavior, so the test passes for the right reason — but the fake encodes the V3 top-level-metadata gap as shipped behavior, and a future `assert result.metadata["deepdoc_backend"]=="local"` would falsely pass. Fix: drop `metadata=dict(kwargs)` from the fake (or assert on segment metadata).

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -227,6 +228,78 @@ def _translate_pdf_bboxes(
     )
```
rogercloud, line 228: shrink: `_translate_remote_elements` (L228-305) re-implements the whole text/table/figure dispatch loop of `_translate_pdf_bboxes` (L188), including the unhandled-type debug log, summary log, and ParseResult assembly. Reshape elements into bbox dicts, delegate to `_translate_pdf_bboxes`, then merge leftover format-specific metadata keys into segment metadata (~100 lines → ~25-30). Caveat: per-list metadata alignment must be preserved (naive zip misaligns).

tests/providers/pdf_parser/test_deepdoc_remote.py
```diff
@@ -0,0 +1,763 @@
...
+    }
+
+
+class TestParseDocumentRemoteSuccess:
+    """Happy-path element shape and request shape."""
+
+    def test_elements_are_translated_to_local_shape(self, tmp_path: Path) -> None:
+        """image_base64 becomes an ``image`` path, or None, and metadata survives."""
+        source = tmp_path / "report.pdf"
+        source.write_bytes(b"%PDF-1.7 fake")
```
rogercloud, line 123: shrink: `tmp_path/"report.pdf"` + `write_bytes(b"%PDF-1.7 fake")` scaffold is repeated in 14 tests. Add a `pdf_source` fixture (12 identical; the 2 renamed-file tests need a parametrized name or inline scaffold).

tests/providers/pdf_parser/test_deepdoc_remote.py
```diff
@@ -0,0 +1,763 @@
...
+            _transport=json_transport({"elements": []}, sink=requests),
+        )
+
+        assert "\r\n\r\n3\r\n" in requests[0].content.decode("utf-8", errors="replace")
+
+    def test_authorization_header_present_when_a_key_is_configured(
+        self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+    ) -> None:
+        source = tmp_path / "report.pdf"
+        source.write_bytes(b"%PDF-1.7 fake")
```
rogercloud, line 219: shrink: the auth-header present/absent tests (L215-240) are byte-identical except env setup and assertion. Parametrize over `(key_env, expected)`.

tests/providers/pdf_parser/test_deepdoc_remote.py
```diff
@@ -0,0 +1,763 @@
...
+                id="non-json-body",
+            ),
+            pytest.param(
+                lambda: json_transport({"elements": [None]}),
+                ValueError,
+                id="element-is-none",
+            ),
+        ],
+    )
+    def test_underlying_exception_types_do_not_escape(
```
rogercloud, line 526: shrink: `test_underlying_exception_types_do_not_escape` re-runs 5 cases already covered by the L457 parametrized suite, adding only a `__cause__` assertion. Fold into that suite and delete (~50 lines).

tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+class TestRemoteRoutingSuccess:
+    """Remote mode must succeed without ever touching the local ONNX path."""
+
+    @pytest.mark.asyncio
+    async def test_pdf_goes_remote_without_local_parser(
+        self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, remote_configured: str
+    ) -> None:
+        """A configured remote server handles .pdf; the local parser stays untouched."""
+        pdf_file = tmp_path / "remote.pdf"
+        pdf_file.write_bytes(b"%PDF-1.4 not really a pdf")
```
rogercloud, line 216: shrink: `pdf_file = tmp_path/...` + `write_bytes(b"%PDF-1.4 not really a pdf")` scaffold is repeated in 5 tests. Add a fixture.

tests/providers/pdf_parser/test_deepdoc_remote_fallback.py
```diff
@@ -0,0 +1,919 @@
...
+        monkeypatch: pytest.MonkeyPatch,
+        tmp_path: Path,
+        caplog: pytest.LogCaptureFixture,
+        remote_configured: str,
+    ) -> None:
+        """DeepDocRemoteError yields the local result, a warning, and backend=local."""
+        pdf_file = tmp_path / "fallback.pdf"
+        pdf_file.write_bytes(b"%PDF-1.4 not really a pdf")
+
+        def failing_parse_document_remote(*args: Any, **kwargs: Any) -> Any:
```
rogercloud, line 398: shrink: `failing_parse_document_remote` is defined verbatim in 3 tests (L398, L461, L504) and the `_get_parser_for_ext` lambda is monkeypatched in 4 tests (L419, L472, L559, L589). Hoist module-level stub helpers.

rogercloud (changes requested):

OliverBryant:
Superseded by #1271, which is the integration branch for #1161 built on the shipped `task="parse"` contract.

This branch implemented the `/v1/document/parse` endpoint proposed in the requirements doc. That endpoint was never built — xorbitsai/inference#5299 instead added `task="parse"` to the existing `/v1/images/ocr` route. Every commit here is in #1271's history.

PR #1182, opened 2026-08-07
feat(deepdoc): adapt the remote client to Xinference's DeepDoc OCR endpoint

OliverBryant:
## Why

The remote DeepDoc client targeted `POST /v1/document/parse` — an endpoint proposed alongside this work that **was never implemented and does not exist**. Xinference PR [#5230](https://github.com/xorbitsai/inference/pull/5230) shipped DeepDoc as an **image/OCR model**, not as a whole-document parse API, exposing it through the existing `POST /v1/images/ocr`.

I stood up Xinference on a GPU box and probed it. Everything below is measured against that live server, not guessed.

## The real contract

```
POST {base_url}/v1/images/ocr
Authorization: Bearer <jwt or api key>

multipart/form-data:
  model   = "DeepDoc"            the launched model UID
  image   = @file                PDFs must carry Content-Type: application/pdf
  kwargs  = '{"task":"ocr"}'     a JSON *string*; supports task, pages, dpi,
                                 return_dict, threshold

200 -> {"pages": [{"page": 1, "result": {"task": ..., <payload>}}]}
```

PDFs are handled server-side: the server rasterizes page by page and merges results. There is no `zoomin` scale factor (the server takes a `dpi`), so that argument is no longer forwarded.

**Authentication is a JWT**, obtained from `POST /token` with `{"username","password"}`. A deployment issuing a static API key is also supported — whatever is configured is sent as the bearer verbatim, since the two are indistinguishable to the client.

Task payloads:

- `task=ocr` with `return_dict=true` → `lines`, each `{"box": [[x,y] × 4], "text", "score"}`. `box` is a **quadrilateral**, not `x0/x1/top/bottom`; the client reduces it to its bounding rectangle.
- `task=layout` → `layouts`, each `{"type", "bbox": [x0,y0,x1,y1], "score"}`. Types are DeepDoc's label set lower-cased: `text`, `title`, `figure`, `figure caption`, `table`, `table caption`, `header`, `footer`, `reference`, `equation`.
- `task=table` returns raw cell boxes with **no HTML**, so the client does not use it.

## The spatial join

Neither task returns text grouped into blocks — `ocr` gives lines without structure, `layout` gives structure without text. The client issues both over one connection and joins them.

This is sound because the two tasks were **measured to share one coordinate space**: on the sample document `ocr` spanned x∈[193,1499] y∈[213,1804] and `layout` spanned x∈[196,1503] y∈[214,1803], with their first regions coinciding.

Per page:

1. Rank layout blocks — `table`/`figure` highest, plain blocks next, captions lowest.
2. In rank order, each block claims still-unclaimed lines whose **center point** falls in its bbox. Center-point containment tolerates a line poking out of its block; claiming each line once means overlapping blocks cannot duplicate text.
3. Claimed lines join with `\n` in reading order (sort by y, then x). Element bounds are the union of its lines, describing the text rather than the detector's guess.
4. `type` maps from the label: `table`→`table`, `figure`/`figure caption`→`figure`, everything else→`text`.
5. **Any line no block claims becomes its own text element** — the join never drops recognized text.
6. Elements sort by `(page, top, x0)`.

Captions rank *lowest* from measured behavior rather than taste: the layout model emitted `table caption` over blocks it had also labelled `title` and `text`, and those labels were the better ones. Ranking captions last keeps the better label at no cost — the text is still emitted, just under a plainer type.

## Capability gap — please read before enabling remote mode

Building on an OCR endpoint rather than a document-parse endpoint is a **real downgrade in output fidelity**, traded for speed:

| Capability | Local | Remote |
|---|---|---|
| Table HTML | reconstructed `<table>` with merged-cell headers | **none** — a table's text is its OCR lines joined by newline |
| Figure/table images | crops saved to `artifacts/` | **none** — no image bytes are returned; `image_path` is always `None` |
| Cross-line paragraph merging | XGBoost merges lines into semantic paragraphs across columns and pages | **none** — blocks are whatever the layout model drew |
| Cross-page coordinate accumulation | continuous document-level space | **none** — page-local, in rasterized pixel space, so highlight overlays built for local coordinates will not line up |
| `col_id` (two-column) | from column detection | always `0` |

Non-PDF formats are not served by an image OCR endpoint at all — uploading `test.docx` returns `500 {"detail": "cannot identify image file ..."}`. They still route remotely when the URL is set, costing one wasted round trip before the standard fallback parses them locally. Restricting remote mode to PDFs and images is the obvious next step; I left it out to keep this PR scoped to the contract change.

## Config

New, following existing conventions:

- `XAGENT_DEEPDOC_XINFERENCE_MODEL_UID` (default `DeepDoc`)
- `XAGENT_DEEPDOC_XINFERENCE_USERNAME` / `XAGENT_DEEPDOC_XINFERENCE_PASSWORD` — for the `/token` exchange when no API key is set

The password is deliberately not whitespace-stripped, since whitespace can be part of a password. The `/token` exchange uses a fixed 30s timeout rather than inheriting the document-sized one, so a hung auth endpoint fails fast into the fallback instead of stalling for half an hour.

## Compatibility

The public surface (`DeepDocRemoteError`, `is_remote_configured()`, `parse_document_remote()`) and the element shape are unchanged, so `deepdoc.py` needs no edit and `test_deepdoc_remote_fallback.py` passes untouched. `save_image` and `zoomin` are kept for signature compatibility but unused — documented as such rather than silently ignored.

### Unit

`tests/providers/pdf_parser/test_deepdoc_remote.py` rewritten against the new contract (118 tests) via the `httpx.MockTransport` `_transport` seam: two-call flow and form fields, all three auth paths (static key, JWT exchange, unauthenticated) plus half-configured credentials, the spatial join (merging, orphan lines, type mapping, reading order, overlap resolution, metadata/positions shape), and every failure mode funneling into `DeepDocRemoteError` — 4xx/5xx on either call, transport errors, non-JSON, missing `pages`, malformed `lines`/`layouts`.

```
pytest tests/core/test_config.py tests/providers/pdf_parser/ -q
601 passed, 1 failed, 2 skipped
```

The single failure is pre-existing and unrelated: `PyMuPdfParser` needs `fitz`, which is not installed in this environment.

```
ruff format --check   4 files already formatted
ruff check            All checks passed!
mypy deepdoc_remote.py    0 errors in this file (26 pre-existing in 9 unrelated files)
isort --check-only    clean
codespell             clean
```

### Live end-to-end

Against a real Xinference with GPU-backed DeepDoc, running the real `DeepDocParser.parse()` on `tests/resources/test_files/test.pdf`, with `_get_parser_for_ext` armed as a tripwire to prove no local ONNX loads:

| | Remote | Fallback (URL → dead port) |
|---|---|---|
| Tripwire (local parser instantiated) | **never fired** | fired, `.pdf` |
| `deepdoc_backend` | `remote` | `local` |
| Wall time | **2.63s** | ~40s |
| text_segments / tables / figures | 23 / 2 / 1 | 48 / 2 / 2 |
| Extracted characters | 2606 | 3331 |
| Table output | OCR lines joined by newline | reconstructed `<table>` HTML |

**Join fidelity: the server returned 103 non-blank OCR lines and the join emitted 103 — zero lost, zero duplicated.** Verified by diffing the raw OCR line set against the joined element text.

Blocks were labelled sensibly — `title` for each heading, `table` for both tables, `figure` for the image — and the two-column section on page 3 correctly produced two separate elements rather than interleaving the columns.

Sample extracted text:

```
[p1 title] 'Sample Document'
[p1 text]  'This document was created using accessibility techniques for headings,
            lists, image alternate text, tables,\nand columns. ...'
[p1 text]  '1. Headings\n2.Lists\n3. Links\n4. Images\n5.Tables\n·Simple Tables\n
            ·Complex Tables\n6. Columns'
[p2 table] 'Share\nScreenReader\nResponses\nJAWS\n853\n49%\nNVDA\n238\n14%\n...'
```

Sample element metadata:

```json
{
  "layout_type": "text", "doc_id": "live_e2e", "page_number": 1,
  "parse_method": "deepdoc", "deepdoc_backend": "remote", "col_id": 0,
  "positions": [[1, 0, 196.0, 712.0, 213.0, 275.0]],
  "x0": 196.0, "x1": 712.0, "top": 213.0, "bottom": 275.0
}
```

The fallback run produced genuine local ONNX output including reconstructed table HTML — which is precisely the capability the remote path lacks, so the two runs also serve as a direct demonstration of the gap documented above.

### Honest assessment of join quality

Good, with a caveat. Text preservation is exact (103/103) and block labelling on this document is correct throughout. The caveat is that quality is bounded by the layout model: it emitted overlapping and occasionally mislabelled blocks (a `table caption` over what was really a `title`), and the ranking rule is a heuristic tuned on one document. It resolves every overlap I observed without losing text, but a document whose layout detection is worse will produce correspondingly coarser blocks. The unclaimed-line escape hatch means the failure mode is degraded structure, never missing content.

The 2606 vs 3331 character difference is **not** text loss: remote joins lines into blocks (one element per paragraph) while local emits one element per line, and local additionally emits table HTML markup, which inflates its count.

Closes part of #1161.


OliverBryant:
Verified the live end-to-end run independently rather than taking the report at face value, and found one defect while doing it.

## Independent verification

Ran the real parse against the GPU box myself, with `_get_parser_for_ext` armed as a tripwire:

```
backend : remote        (2.29s)
segments: 23  tables: 2  figures: 1
tripwire never fired -> zero local ONNX loads
positions: [[1, 0, 196.0, 712.0, 213.0, 275.0]]
```

**The no-text-loss claim holds.** I checked it the hard way — pulled the raw OCR lines straight from the server, then diffed them against the joined output:

```
raw OCR lines            : 103
joined elements          : 26
lines missing from output: 0
```

Every one of the 103 OCR lines survives into the result.

## Defect found: `deepdoc_backend` was unset on every non-PDF format

The follow-up about non-PDF formats wasting a round trip is real, but while confirming it I hit something worse. A `.docx` parse came back with `deepdoc_backend=None` instead of `local`. Testing all six formats:

| format | before | after |
|---|---|---|
| .pdf | `local` | `local` |
| .docx / .xlsx / .md / .txt / .csv | **`None`** | `local` |

Root cause is pre-existing, not introduced here: only `_translate_pdf_bboxes` forwarded the parse metadata onto its `ParseResult`. The other five translators dropped it, so `metadata["deepdoc_backend"] = "local"` never reached the caller. This feature is what made it matter — acceptance criterion 3 requires the fallback to be observable via that marker, and for five of six formats it silently wasn't.

Fixed in 303660ad by forwarding the metadata from the remaining translators, plus a regression test. I mutation-checked the test: reverting the fix turns both parametrized cases red, so it isn't vacuous.

Full suite still passes apart from the pre-existing `PyMuPdfParser`/`fitz` failure; ruff, isort, codespell and mypy all clean.

## Still open

Gating remote mode to PDFs remains the right follow-up — a `.docx` currently costs a `500 cannot identify image file` round trip before falling back.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,709 @@
...
+        The parsed elements, each with ``image`` set to ``None``.
+
+    Raises:
+        DeepDocRemoteError: On any failure. Callers fall back to local parsing.
+    """
+    del save_image, zoomin  # Signature compatibility only; see above.
+
+    started = time.monotonic()
+    if callback is not None:
+        callback(0.05, "Uploading document to remote DeepDoc server")
```
rogercloud, line 642: **major** — Two of the three progress-callback invocations are outside the try/except, so a raising callback escapes as a raw exception and defeats the local fallback.

The callbacks fire at fraction 0.05 (this line, before the `try` at :644), 0.5 (:668-669, inside the try), and 1.0 (:701-702, after the `except` chain ends at :698). Only the 0.5 call is protected by the catch-all `except Exception` that converts everything into `DeepDocRemoteError`.

Verified empirically: a callback that raises at fraction 0.05 or 1.0 propagates a non-`DeepDocRemoteError` out of `parse_document_remote`, straight past the caller's narrow `except deepdoc_remote.DeepDocRemoteError` in `deepdoc.py:684`. That turns a recoverable remote failure into a hard parse failure for that document — exactly the invariant this module's docstring promises to hold ("Raises: DeepDocRemoteError: On any failure. Callers fall back to local parsing.").

Suggested fix:
- Move this 0.05 callback **and** `started = time.monotonic()` (:640) inside the `try` block.
- Wrap the final 1.0 callback (:701-702) in its own `try/except Exception` with a warning log, mirroring the pattern the caller already uses at `deepdoc.py:692-699`.

Related test gap: `tests/providers/pdf_parser/test_deepdoc_remote.py:1163-1177` (`test_a_raising_callback_does_not_escape_as_itself`) only exercises fraction 0.5, so this is currently untested. Parametrizing that test over `[0.05, 0.5, 1.0]` would make it red today and green after the fix.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,709 @@
...
+        client: The HTTP client to exchange credentials on.
+        base_url: Xinference base URL.
+
+    Raises:
+        httpx.HTTPError: On any transport failure or non-2xx status.
+        ValueError: If the token response is not a JSON object carrying a
+            non-empty string ``access_token``.
+    """
+    api_key = get_deepdoc_xinference_api_key()
+    if api_key:
```
rogercloud, line 189: **minor** (observability) — An inherited `XINFERENCE_API_KEY` silently preempts a correctly configured JWT setup, with no log line saying which auth path was chosen.

`get_deepdoc_xinference_api_key()` falls back to the bare `XINFERENCE_API_KEY` (see `config.py:1885`), which may have been set for an entirely different Xinference deployment — `src/xagent/core/model/rerank/xinference.py:57` uses it with its own `base_url` defaulting to `localhost:9997`. Because this branch returns unconditionally before username/password are ever consulted, an operator who correctly sets `XAGENT_DEEPDOC_XINFERENCE_USERNAME`/`_PASSWORD` for the JWT flow — which `docs/deepdoc-remote-parsing.md` describes as the *primary* mechanism — gets that configuration silently ignored, and the resulting `401` gives no hint why.

To be clear: the credential *sharing* itself is intentional and documented (this function's docstring and `example.env:263-271`), so this is not a leak. The gap is diagnosability of the precedence.

Suggested fix: log at INFO/DEBUG which auth path was selected — e.g. "dedicated key" / "inherited XINFERENCE_API_KEY" / "username+password JWT" / "unauthenticated" — so a precedence misconfiguration is debuggable from logs alone.

src/xagent/config.py
```diff
@@ -1853,6 +1859,60 @@ def get_kb_collections_timeout_seconds() -> int:
...
+    """Return the API key for remote DeepDoc parsing, if one is configured.
+
+    A dedicated key wins over the bare ``XINFERENCE_API_KEY`` shared with the
+    other Xinference clients. Returning None is valid: a self-hosted
+    Xinference deployment often runs without authentication.
+    """
+    value = (os.getenv(DEEPDOC_XINFERENCE_API_KEY) or "").strip()
+    if value:
+        return value
+    return (os.getenv("XINFERENCE_API_KEY") or "").strip() or None
```
rogercloud, line 1885: **note (design callout, no change required)** — This falls back to the bare, pre-existing `XINFERENCE_API_KEY`, which is independently consumed by an unrelated Xinference deployment (`src/xagent/core/model/rerank/xinference.py:57`, own `base_url` defaulting to `localhost:9997`). So enabling remote DeepDoc against a *different* Xinference host will send that host a credential minted for the rerank deployment.

This is intentional and documented (the docstring above, plus `example.env:263-271`), so flagging it as a design tradeoff rather than a defect — recording it so it is a conscious decision rather than an accident.

The actionable half of this is on the consuming side: `deepdoc_remote.py:188-190` returns this inherited key unconditionally, before username/password are checked, with no logging of which auth path won. See the comment there.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -227,6 +228,78 @@ def _translate_pdf_bboxes(
...
+        element_metadata = element.get("metadata")
+        if not isinstance(element_metadata, dict):
+            element_metadata = {}
+
+        layout_type = element.get("type", "text")
+        # Reshape into the bbox dict the local helpers expect. Format-specific
+        # metadata keys are flattened in so _build_element_metadata can pick up
+        # positions/col_id/page_number when the format supplies them.
+        bbox: Dict[str, Any] = {
+            **element_metadata,
```
rogercloud, line 267: **minor** — **Prior finding from gemini-code-assist (on `deepdoc.py:271`): NOT FIXED.** Re-verified at HEAD `303660ad`; that commit touched unrelated translator functions, not this one.

`**element_metadata` is unpacked here, *before* `"layout_type": layout_type` on the next line, so the coarse mapped type (`text`/`table`/`figure`) overwrites the raw fine-grained layout label (`"figure caption"`, `"title"`, ...) that `deepdoc_remote.py` deliberately preserves at `:483-511` and `:562-569`.

Impact is genuinely minor: the only real consumer of `layout_type`, `parse_display.py:115-126`, dispatches on the coarse three-way value and would treat any fine-grained label as `else`/unknown anyway — so there is no functional regression today. But `tests/providers/pdf_parser/test_deepdoc_remote.py:632-641` asserts a "raw label survives" guarantee that is false end-to-end; it passes only because it inspects `deepdoc_remote.py`'s output *before* this translation step, which gives a false sense of coverage.

Apply the bot's original fix — reorder the dict literal so `**element_metadata` comes *after* `"layout_type": layout_type`:

```python
bbox: Dict[str, Any] = {
    "layout_type": layout_type,
    "text": element.get("text", ""),
    "image": element.get("image"),
    **element_metadata,
}
```

One-line change, and it makes the existing test's implied guarantee true for real callers.

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -227,6 +228,78 @@ def _translate_pdf_bboxes(
...
+            "layout_type": layout_type,
+            "text": element.get("text", ""),
+            "image": element.get("image"),
+        }
+        base_metadata = _build_element_metadata(bbox, doc_id, **kwargs)
+        # Carry the format-specific keys (style, sheet_name, row_number,
+        # row_type, ...) that _build_element_metadata does not know about.
+        for key, value in element_metadata.items():
+            base_metadata.setdefault(key, value)
```
rogercloud, line 276: **minor** — This `setdefault` merge can re-inject the malformed `positions` value that `_build_element_metadata` deliberately rejected.

`_build_element_metadata` (`deepdoc.py:108-127`) only sets `positions` when every entry has `len(pos) >= 5`; a shorter entry causes it to skip the key entirely. But `element_metadata` still contains the original raw list, so `setdefault("positions", ...)` here puts the malformed value (e.g. a bare 3-tuple) back — the caller ends up with a `positions` shape that violates the documented 5/6-element contract, rather than with no `positions` at all.

Unreachable from the current remote client, which always emits well-formed 5-element positions — but `_translate_remote_elements` is a public function that any future producer can call. Low priority; if you touch this code again, add a defensive skip:

```python
for key, value in element_metadata.items():
    if key == "positions":
        continue  # _build_element_metadata owns validation of this key
    base_metadata.setdefault(key, value)
```

src/xagent/providers/pdf_parser/deepdoc.py
```diff
@@ -552,6 +630,76 @@ def _sync_parse() -> ParseResult:
...
+                    DeepDocProgressAdapter,
+                )
+
+                adapter = DeepDocProgressAdapter(progress_callback)
+                callback = adapter.get_callback()
+
+            # Remote-first dispatch: when a remote server is configured every
+            # supported format goes there, and any failure falls back to local
+            # parsing. No local parser is instantiated on the remote path.
+            if deepdoc_remote.is_remote_configured():
```
rogercloud, line 669: **minor** — Two related cleanups belong at this dispatch point.

1. **Dead code:** `_get_parser` (`deepdoc.py:554-557`) has zero callers anywhere in `src/` or `tests/` after this refactor — only `_get_parser_for_ext` is still used (`:714`, `:734`, `:742`). Delete `_get_parser`.

2. **Lost validation as a side effect of that:** `_get_parser` used to raise `ValueError` for an unsupported extension *before* any I/O. Now, with remote configured, an unsupported file (e.g. `.zip`) is uploaded to the remote endpoint before any extension-support check happens.

Suggested fix — add an extension check ahead of this branch:

```python
if ext not in self.supported_extensions:
    raise ValueError(f"Unsupported file extension: {ext}")
```

If you also narrow that to the extensions the OCR endpoint can actually serve (`.pdf` plus image types) before attempting the remote call, it additionally addresses the wasted-round-trip issue raised as DL1 in the review body: today all 8 non-PDF supported formats upload the full document, get an immediate 500, and fall back locally.

src/xagent/config.py
```diff
@@ -1853,6 +1859,60 @@ def get_kb_collections_timeout_seconds() -> int:
...
+
+    Leaving this unset keeps document parsing entirely local. It must
+    otherwise be an absolute http:// or https:// base URL carrying no query
+    or fragment, since request paths are appended to it. A malformed value
+    raises rather than silently downgrading to local parsing, so the
+    misconfiguration surfaces instead of showing up only as unexplained
+    slowness.
+    """
+    return _normalized_http_env_url(DEEPDOC_XINFERENCE_URL)
+
```
rogercloud, line 1873: **minor** — The shared URL validator this getter delegates to does not reject embedded userinfo, so credentials in the URL can reach the logs.

`_normalized_http_env_url` (`config.py:385-396`) checks `scheme`, `netloc`, `query`, and `fragment`, but not `parts.username` / `parts.password`. A value like `http://user:secret@host:9997` is therefore accepted verbatim. On an HTTP error, httpx's `raise_for_status()` message embeds the full request URL including those credentials; that message is wrapped into `DeepDocRemoteError` (`deepdoc_remote.py:683`) and then written to logs by `logger.warning` in `deepdoc.py:685-687`.

Suggested fix: reject (or strip) `parts.username`/`parts.password` in `_normalized_http_env_url` alongside the existing query/fragment check. Note this is a shared helper — please confirm no other caller relies on userinfo being permitted before making it a hard rejection.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,709 @@
...
+    """Return whether remote DeepDoc parsing is configured.
+
+    A malformed URL makes the config getter raise. Degrading to local parsing
+    with a warning is the right response: a typo in one environment variable
+    must not break every document parse.
+    """
+    try:
+        return get_deepdoc_xinference_url() is not None
+    except ValueError as exc:
+        logger.warning(
```
rogercloud, line 165: **minor** — This warning fires once per document, not once per process.

`is_remote_configured()` performs no caching and is called on every parse from `deepdoc.py:669`. With a malformed `XAGENT_DEEPDOC_XINFERENCE_URL`, a bulk ingest of N documents emits N identical warnings — the signal is real but it becomes log noise exactly when someone is trying to read the logs to find it.

Suggested fix: cache the validity check (e.g. an `lru_cache`d helper, or a module-level warn-once flag) so the misconfiguration is reported once. Worth keeping the check itself uncached if you want live env-var re-reads in tests — in that case gate only the `logger.warning` behind the flag.

src/xagent/providers/pdf_parser/deepdoc_remote.py
```diff
@@ -0,0 +1,709 @@
...
+            # page_lines is already in reading order, so members inherits it.
+            elements.append(
+                _build_element(
+                    element_type=_LAYOUT_TYPE_TO_ELEMENT_TYPE.get(
+                        block["layout_type"], "text"
+                    ),
+                    layout_type=block["layout_type"],
+                    text="\n".join(member["text"] for member in members),
+                    page_number=page_number,
+                    x0=min(member["x0"] for member in members),
```
rogercloud, line 571: **informational** (doc fix only, no code change needed) — Remote segment metadata is not key-for-key identical to local.

Direct comparison: local-path metadata keys are `{col_id, doc_id, layout_type, page_number, positions}`; remote-path metadata is that set plus `{x0, x1, top, bottom}`. Nothing downstream depends on exact key-set parity, so this is harmless today — but it does undercut the PR's claim that remote and local element metadata shape is unchanged. Worth a one-line correction in `docs/deepdoc-remote-parsing.md` so the next person comparing the two paths is not surprised.

rogercloud (changes requested):
## Summary

This PR rewrites the remote DeepDoc client (`src/xagent/providers/pdf_parser/deepdoc_remote.py`, new, ~709 lines) because the whole-document endpoint the feature was originally designed around (`POST /v1/document/parse`) was never implemented server-side. Instead the client now adapts to the endpoint that actually shipped — `POST /v1/images/ocr` — by issuing two calls per document (`task=ocr` for text lines, `task=layout` for block structure) and performing a client-side spatial join: layout blocks are ranked by type priority, OCR lines are claimed by center-point containment, and unclaimed lines become standalone elements. It also adds JWT/static-token auth, six new `XAGENT_DEEPDOC_XINFERENCE_*` config vars, a remote-first dispatch in `DeepDocParser._sync_parse`, and a rewritten test suite (2151 new test lines) plus `docs/deepdoc-remote-parsing.md`.

Adapting to the endpoint that actually shipped is the right call, and the client itself is careful, defensively written, and unusually well tested. The reservation is about scope alignment rather than code quality: the endpoint change collapsed the feature from "all formats, transparent offload" to "PDF-only, materially lossier," and this PR updated the **client** without updating the **feature's contract**. Routing (`deepdoc.py:669`), the capability marker (`deepdoc.py:522`), and `example.env` all still promise the old, broader behavior.

### Design-level findings

**DL1 — MINOR (should-fix): no extension guard before attempting remote parse.**
The remote dispatch at `src/xagent/providers/pdf_parser/deepdoc.py:669` (`if deepdoc_remote.is_remote_configured():`) fires for every extension in `supported_extensions` (`.pdf .docx .xlsx .xls .csv .md .txt .json .html`), but the OCR endpoint only serves PDF/image. 8 of 9 formats upload the whole document, receive an immediate `500`, and then fall back locally — a wasted round trip on every non-PDF document whenever remote mode is enabled.
Explicitly downgraded from a blocking concern: this is opt-in (requires an env var), has no correctness or data-loss impact, fails fast rather than hanging, and is self-documented by the author in `docs/deepdoc-remote-parsing.md:218-223` as "the obvious next step" deferred to keep the PR scoped. Still worth a ~3-line guard (`ext in {".pdf", <image exts>}`) ahead of the remote attempt — see also L5, which suggests the same guard site.

**DL2 — MAJOR: the `LocalParsing` capability marker no longer reflects runtime behavior, and there is no per-request opt-out.**
`DeepDocParser` still declares only `LocalParsing` (`src/xagent/providers/pdf_parser/deepdoc.py:522`). That marker is what `src/xagent/providers/document_parser.py:280-287` uses to filter parser selection, and what `workspace_file_tool.py:79` relies on when it passes `use_local_parser=True`. But this PR's remote-first dispatch (`deepdoc.py:664-677`, comment: "Remote-first dispatch: when a remote server is configured every supported format goes there") is unconditional — `use_local_parser=True` never reaches `_parse_impl` and has zero effect on whether the document is uploaded.
Consequence: once an operator sets `XAGENT_DEEPDOC_XINFERENCE_URL` for any purpose, every caller that explicitly requested local-only parsing silently gets remote upload instead, process-wide, with no override. This is a data-egress control gap, not just a naming inconsistency — the capability marker asserts "local" while the runtime behavior is "remote, always."
Recommend one of: (a) declare a `RemoteParsing` capability so parser selection is honest, or (b) thread `use_local_parser` down into `_parse_impl` and skip the remote attempt when it is set.

**DL3 — MAJOR: the spatial join depends on an unpinned DPI assumption with no runtime verification.**
The join's correctness requires `task=ocr` and `task=layout` to rasterize at the same DPI, but `dpi` is never sent in either call's `kwargs` (`src/xagent/providers/pdf_parser/deepdoc_remote.py:248-254`), and the accepted `zoomin` parameter is explicitly discarded (`deepdoc_remote.py:638`, `del save_image, zoomin`) rather than translated into a DPI. The soundness rests entirely on the empirical claim in the module docstring ("measured to share one coordinate space"), with nothing asserting it at runtime.
Failure mode: if the server's default DPI ever diverges between the two task types, the join silently produces all-orphan or garbled output with no exception raised — a quality collapse that looks like a successful parse and therefore never triggers the local fallback.
Fix: send an explicit `"dpi": 72 * zoomin` (or equivalent) in both calls' `kwargs`. One line, and it also resolves DL4 for free.

**DL4 — MINOR (latent): remote and local `positions` are in different coordinate spaces.** As a consequence of DL3, remote `positions` metadata is in the server's unpinned rasterized pixel space, while local `positions` are in PDF-point space at 72 DPI. Harmless today (only reader has no callers outside `base.py`/tests), but becomes real once PDF highlighting is built. Fixing DL3 resolves this.

**DL5 — MINOR: no token caching; document uploaded twice per parse.** `_get_access_token` performs a fresh `/token` exchange on every parse on the username/password path (a 1000-document batch = 1000 JWT exchanges), and the document is uploaded twice (once per task).

**DL6 — MINOR: layout blocks that claim zero OCR lines are silently dropped.** `deepdoc_remote.py:544-561` — a pure-image figure with no in-bbox caption vanishes completely from remote output; local would still emit a figure element with an image crop. Not documented in the capability-gap table.

**DL7 — MINOR: commit `303660ad` bundles an undisclosed metadata behavior change.** Also changes `metadata={}` to `metadata=kwargs` for five local translators; `kwargs` includes `source` (absolute local file path), now written into agent-visible export output for docx/xlsx/csv/md/txt where it wasn't before.

**DL8 — MINOR: block-ranking ties between `title` and `text` are broken by position, not specificity.** `"title"` is absent from `_LAYOUT_TYPE_PRIORITY` and falls to the same tier as `"text"`; a wide text block starting slightly higher can greedily claim the title's line first. No text is lost (orphan escape hatch), but this is a reachable mis-grouping mode.

---

## Line-level findings

| ID | Severity | Location | Issue |
|----|----------|----------|-------|
| L1 | major | `deepdoc_remote.py:642` (also `:640`, `:701-702`) | Progress callbacks at fractions 0.05 and 1.0 are outside the try/except; a raising callback escapes as a non-`DeepDocRemoteError` and defeats local fallback |
| L3 | minor (observability) | `deepdoc_remote.py:189`, `config.py:1885` | Inherited `XINFERENCE_API_KEY` unconditionally preempts the JWT path with zero diagnostic logging |
| L2 | minor | `deepdoc.py:267` | `**element_metadata` unpacked before `"layout_type"`, overwriting the fine-grained layout label — **prior bot finding, NOT FIXED** |
| L4 | minor | `deepdoc.py:276` | `setdefault` merge re-injects a malformed raw `positions` list |
| L5 | minor | `deepdoc.py:669` (dead code at `:554-557`) | `_get_parser` is dead; unsupported extensions uploaded before extension validation |
| L6 | minor | `config.py:1873` (validator at `:385-396`) | URL validator doesn't reject embedded userinfo; credentials can reach logs |
| L7 | minor | `deepdoc_remote.py:165` | Malformed-URL warning emitted once per document, not once per process |
| L8 | informational | `deepdoc_remote.py:571` | Remote segment metadata carries 4 extra keys local never emits; contradicts "shape unchanged" claim |

---

## Test coverage gaps

- **T1.** Timeout config (`XAGENT_DEEPDOC_XINFERENCE_TIMEOUT_SECONDS`) never asserted to actually reach the httpx.Timeout, nor that `/token` independently uses its fixed 30s.
- **T2.** The raising-callback test only exercises fraction 0.5 (the protected site) — parametrizing over `[0.05, 0.5, 1.0]` would turn L1 into a red test today. Highest-value single test change here.
- **T3.** `test_pdf_shaped_elements_match_local_translation` builds its "local" comparison from the same remote metadata (missing x0/x1/top/bottom) and a fabricated `<table>` HTML the real client never produces — doesn't validate real output shape.
- **T4.** Backend-marker test covers only 2 of 5 translators changed by commit `303660ad`.
- **T5.** No negative test proves the reordered magic-byte validation rejects malformed files.
- **T6.** No test confirms the positive fallback notice reaches a working callback.
- **T7 (partial).** Page-count mismatch partially covered; malformed `positions` input still uncovered.
- **T8.** `test_env_var_constants` is tautological — low priority to remove.

## Documentation

- **Doc-A.** `example.env:252-254` claims "ALL DeepDoc-supported formats are parsed remotely" with no caveat that 8/9 always fail remotely first.
- **Doc-B.** `docs/deepdoc-remote-parsing.md` Goals section (lines 25-32) contradicts section 7's own admission 190 lines later that non-PDF formats are a slowdown, not a speedup.

```
deepdoc_remote.py:215-217: shrink _mime_type_for is a one-line wrapper around a dict with a single call site. Inline it and delete the function.
config.py:1882-1885,1903,1908: shrink three getters repeat the identical `(os.getenv(X) or "").strip() or <fallback>` idiom. A tiny `_get_stripped_env(env_var, default=None)` helper would collapse these to one line each.
```
net: -8 lines possible.

## Re-review: prior finding status

| Prior finding | Status |
|---|---|
| `deepdoc.py:271` — `layout_type` overwritten because `**element_metadata` is unpacked before `"layout_type"` is set | **NOT FIXED** |

Verified against HEAD (`303660ad`) — the intervening commit touched unrelated translator functions, not this one. Impact is smaller than it first appears (only real consumer dispatches on the coarse type only), but the fix is trivial. See L2.

## Test results

`.venv/bin/pytest tests/core/test_config.py tests/providers/pdf_parser/ -q` — all pass except one pre-existing, unrelated failure: `test_basic_parsers_pdf_success[PyMuPdfParser]` (missing `fitz` module in this environment, not caused by this PR).

## Recommendation: Request changes

Blocking: **L1** (callback exception escape breaks fallback invariant), **DL2** (capability marker no longer reflects runtime behavior, no opt-out), **DL3** (unpinned DPI, silent join failure mode with no error).

Not blocking: **DL1** (opt-in, fails fast, already documented as deferred), **L3** (documented design tradeoff + minor observability gap).

OliverBryant:
Superseded by #1271.

This adapted the client to the per-page `task="ocr"` endpoint and reconstructed document structure by spatially joining OCR lines with layout blocks. That was the best available option at the time, and it carried a real capability gap — no table HTML, no images, no paragraph merging, no cross-page coordinates.

xorbitsai/inference#5299 has since exposed `task="parse"`, which runs the full `parse_into_bboxes()` pipeline server-side and returns the element shape the client already wants. The stitching is no longer needed, and the capability gap is closed: measured parity with local CPU parsing is now 0 text differences and sub-pixel coordinates.

The `deepdoc_backend` fix from this branch (metadata was dropped by every non-PDF translator) is carried over in #1271.

PR #1127, opened 2026-08-04
feat: add Growth Marketing Workforce template with type filtering

yiboyasss:
## Summary
- Add a "Growth Marketing Workforce" built-in template (Growth Marketing
  Manager orchestrating the existing GA Analyzer + Ads Recommendation
  agents in a measure → recommend → approve → re-measure loop), plus a
  new `type: workforce` template schema (`workforce_config.manager` /
  `workforce_config.agents[]`) alongside the existing single-agent
  template type.
- Add `POST /api/templates/{id}/use-as-workforce`, which instantiates a
  workforce template into a real Workforce (fresh manager agent + one
  worker agent per referenced sub-template, reusing an existing
  quick-access instance if the caller already has one). Worker-agent
  resolution is race-safe: a concurrent/double-clicked request retries
  via a SAVEPOINT instead of surfacing an unhandled 500 on a unique-key
  collision.
- Add an Agent/Workforce type filter on the Templates page alongside the
  existing category filter, a "Workforce" badge + agent count on the
  template card, and a loading/disabled state + error toast while a
  workforce is being created from a template.
- `TemplateManager` now warns at startup if a workforce template
  references an unknown `template_id`, instead of only failing the
  first time a user clicks "Use".

## Test plan
- [x] `pytest tests/web/api/test_templates_api.py
  tests/web/test_workforce_creator_worker_resolution.py
  tests/web/test_workforce_creator_plan.py tests/templates` — all pass
  except one pre-existing, unrelated failure
  (`test_unauthorized_access`, fails identically on `main`).
- [x] `npx tsc --noEmit` (frontend) — clean.
- [x] `python -m py_compile` on the changed backend modules.
- [x] Manual click-through of "Use" on the Growth Marketing Workforce
  card in the browser.

PR #1183, opened 2026-08-07
feat: preconfigure Google Analytics connector on GA Analyzer template

yiboyasss:
## Summary
- Google Analytics Analyzer template now ships with the Google Analytics connector preconfigured (declared under `connections:` and merged into `agent_config.tool_categories`), so a live GA4 pull works out of the box instead of requiring a CSV export or pasted table.
- Instructions updated to prefer the live connection (list properties, ask which one if multiple), falling back to an uploaded/pasted export only when no connection is available.
- Growth Marketing Workforce template's connection icons and GA Analyzer sub-agent assignment instructions updated to match.

## Test plan
- [x] `pytest tests/templates/` — 44 passed
- [x] Verified `marketing-google-analytics-analyzer` template resolves `Google Analytics` under `connections` and as `mcp:Google Analytics` in `agent_config.tool_categories`
- [x] Manually create an agent from the template in the UI and confirm the Google Analytics connector is preconnected on the Configure step

### DIFF
--- a/src/xagent/templates/built_in/marketing-growth-marketing-workforce.yaml
+++ b/src/xagent/templates/built_in/marketing-growth-marketing-workforce.yaml
@@ -39,11 +39,11 @@ sample_prompts:
 # NOT merged into any agent's tool_categories - GET /{id} nulls agent_config
 # for a workforce type, the manager's own tools come from
 # workforce_config.manager.tool_categories below, and its workers' tools
-# come from their own referenced single-agent templates. Kept as the union
-# of what marketing-google-analytics-analyzer and
-# marketing-google-ads-recommendation actually declare under their own
-# `connections:`, so the card doesn't advertise an integration this
-# workforce doesn't have (PR #1127 re-review, F7).
+# come from their own referenced single-agent templates. This list must
+# stay the union of the sub-templates' (workforce_config.agents[].template_id)
+# own `connections:` - don't list a connection the workforce doesn't have.
+# Enforced by
+# test_manager.py::test_builtin_workforce_connections_are_union_of_sub_template_connections.
 connections:
 - name: Google Analytics
   logo: https://www.google.com/s2/favicons?domain=analytics.google.com&sz=64
--- a/src/xagent/templates/built_in/marketing-growth-marketing-workforce.yaml
+++ b/src/xagent/templates/built_in/marketing-growth-marketing-workforce.yaml
@@ -43,9 +43,10 @@ sample_prompts:
 # of what marketing-google-analytics-analyzer and
 # marketing-google-ads-recommendation actually declare under their own
 # `connections:`, so the card doesn't advertise an integration this
-# workforce doesn't have (this previously listed "Google Analytics",
-# which neither sub-template connects to at all).
+# workforce doesn't have (PR #1127 re-review, F7).
 connections:
+- name: Google Analytics
+  logo: https://www.google.com/s2/favicons?domain=analytics.google.com&sz=64
 - name: Google Docs
   logo: https://www.google.com/s2/favicons?domain=docs.google.com&sz=64
 - name: Google Slides
@@ -105,7 +106,7 @@ workforce_config:
     name: Google Analytics Analyzer
     alias: GA Analyzer
     assignment_instructions: |
-      Produce a 'Signals for Ads' table: for each campaign / source, report sessions, CVR, conversions, value, and a verdict (winning / underperforming), comparing the current period to the prior period of equal length. Ask the Growth Marketing Manager for the GA4 export or period range if you don't already have it.
+      Produce a 'Signals for Ads' table: for each campaign / source, report sessions, CVR, conversions, value, and a verdict (winning / underperforming), comparing the current period to the prior period of equal length. Pull the data from the connected Google Analytics account when one is available; otherwise ask the Growth Marketing Manager for a GA4 export. Ask for the period range if you don't already have it.
   - template_id: marketing-google-ads-recommendation
     name: Google Ads Recommendation
     alias: Ads Recommendation
--- a/tests/templates/test_manager.py
+++ b/tests/templates/test_manager.py
@@ -455,6 +455,59 @@ sample_prompts:
 
         assert not offenders
 
+    def test_builtin_workforce_connections_are_union_of_sub_template_connections(self):
+        """A `type: workforce` template's `connections:` is hand-maintained
+        display-only data (see the comment above `connections:` in
+        marketing-growth-marketing-workforce.yaml) - nothing in
+        `TemplateManager` derives or validates it against the sub-templates
+        it actually references. That invariant has drifted and been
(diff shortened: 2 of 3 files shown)


src/xagent/templates/built_in/marketing-growth-marketing-workforce.yaml
```diff
@@ -43,9 +43,10 @@ sample_prompts:
 # of what marketing-google-analytics-analyzer and
 # marketing-google-ads-recommendation actually declare under their own
 # `connections:`, so the card doesn't advertise an integration this
-# workforce doesn't have (this previously listed "Google Analytics",
-# which neither sub-template connects to at all).
+# workforce doesn't have (PR #1127 re-review, F7).
```
rogercloud: This comment previously explained the invariant in self-contained plain terms; it now points at `PR #1127 re-review, F7`, which isn't resolvable from anything in-repo (no CONTRIBUTING.md or review-numbering doc defines "F7"). The old sentence did need replacing — it claimed neither sub-template connects to Google Analytics, which this PR makes false — but the replacement should state the *current* reason rather than an external reference. Suggested:

> This list must stay the union of the sub-templates' own `connections:`; don't list a connection the workforce doesn't have. `Google Analytics` is included because `marketing-google-analytics-analyzer` declares it under its own `connections:`.

tests/templates/test_manager.py
```diff
@@ -455,6 +455,30 @@ def lookup_keys(*values):
...
+        template = await manager.get_template("marketing-google-analytics-analyzer")
+
+        assert template is not None
+        connection_names = [
+            conn.get("name")
+            for conn in template["connections"]
+            if isinstance(conn, dict)
+        ]
+        assert "Google Analytics" in connection_names
+        assert "mcp:Google Analytics" in template["agent_config"]["tool_categories"]
```
rogercloud, line 533: Worth adding a test here for the workforce-union invariant while you're in this file. Right now `TemplateManager` treats a `type: workforce` template's `connections:` as pure hand-maintained data with no derivation or validation, and this exact invariant has drifted and been hand-fixed at least twice in `marketing-growth-marketing-workforce.yaml` — including in this PR.

A generic parametrized test over all built-in `type: workforce` templates would catch this class of drift automatically and is cheap to write, following the pattern of `test_builtin_template_connections_resolve_to_a_registered_mcp_app` above: for each workforce template, collect the `connections:` names of every `workforce_config.agents[].template_id` sub-template and assert the workforce's own `connections:` names equal that union. That would also let the YAML comment shrink to a pointer at the test.

rogercloud (approved):
## Summary

This PR preconfigures the Google Analytics connector on the GA Analyzer template. `marketing-google-analytics-analyzer.yaml` gains a `Google Analytics` entry under `connections:`, and its instructions/sample prompts are rewritten to prefer a live GA4 pull with the CSV export path demoted to a fallback. `marketing-growth-marketing-workforce.yaml` re-adds the matching connection entry (display-only for `type: workforce`) and updates its GA Analyzer sub-agent instructions, and `tests/templates/test_manager.py` pins the resulting `connections` / `tool_categories` contract. 51 insertions / 19 deletions across 3 files.

## Approach: sound

This is the minimal, convention-correct shape — it adds *data*, not mechanism. Confirmed by executing `TemplateManager` against both files: `_enrich_template` (`src/xagent/templates/manager.py:355-382`) mechanically derives `mcp:<conn_name>` tool_categories from `connections:` for `type: agent` templates, so the PR is right **not** to hand-add a duplicate `mcp:Google Analytics` tool_category line. The `- name:` / `logo:` shape matches every other built-in template, and workforce workers inherit the enriched categories via `workforce_creator.py:375-428`. Enrichment nulls `agent_config` for `type: workforce`, so the workforce YAML change is purely a card/display fix — correct as written.

Two design-level notes (non-blocking, no single line owns them):

1. **The workforce `connections:` union is enforced only by a comment.** `marketing-growth-marketing-workforce.yaml`'s `connections:` list is a hand-maintained union of its two sub-templates' connections, with nothing in code or tests deriving or validating it. That invariant has now drifted and been hand-fixed at least twice, including in this PR. See the test-coverage finding below for a cheap fix.
2. **The "never write to GA4" guardrails are defending a capability that doesn't exist.** The connector is read-only by construction — OAuth scope is `analytics.readonly`, and the MCP tool module exposes only `list_properties` / `get_metadata` / `run_report`. The added guardrail sentences are prompt bloat rather than protection; consider trimming them to keep the instruction block tight.

## Findings

**minor** — `tests/templates/test_manager.py`: no test asserts that a `type: workforce` template's `connections:` is the union of its sub-templates' `connections:`. Inline comment left with a suggested parametrized test. This is the drift this PR is manually correcting for the second time.

**minor** — `src/xagent/templates/built_in/marketing-growth-marketing-workforce.yaml:46`: the invariant comment was replaced with an unresolvable internal cross-reference (`PR #1127 re-review, F7`). Inline comment left with suggested wording.

### Nits

- `src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml:29` (zh `sample_prompts`) — trailing space after `媒体资源` before `拉取` is awkward CJK typography. It matches the pre-existing house style in this file and the `highlights` substring matches with or without it, so purely cosmetic.
- `src/xagent/templates/built_in/marketing-google-analytics-analyzer.yaml` `description.en` is now 259 chars — second-longest among the 18 built-in templates (only the workforce template's 293 is longer). No length cap is enforced anywhere in code or tests and no card-truncation logic tied to a specific limit exists. Informational only, no action requested.

### Investigated, no action needed

- `templateMissingMcp` in the frontend `agent-builder.tsx` will now fire for Google Analytics as it already does for Docs/Slides. It does not gate publish/create — it only drives step-highlight UI — and the same friction predates this PR for the other two connectors.
- `setup_time` left at "5 min setup" despite a third connector: this matches the codebase-wide convention (all `type: agent` templates use "5 min" regardless of connector count; only `type: workforce` uses "10 min"). Not an inconsistency.
- The new `"mcp:Google Analytics" in tool_categories` assertion is mechanically implied by the connections assertion, but it is not covered elsewhere and pins two independently meaningful contracts (advertised vs. functionally wired). Normal regression-pinning — keep it.

## Simplification

Lean already. No over-engineering or duplication found in this diff.

## Test status

The PR's claim of `pytest tests/templates/` → 44 passed could not be independently re-run here (unrelated `lancedb` dependency/package-path issue in the review environment), so that specific number is unverified. The substantive behavior *was* verified directly by running `TemplateManager` enrichment against both YAML files and confirming the derived `tool_categories` and the workforce `agent_config` nulling.

## Prior review

The only prior review is from the `gemini-code-assist` bot (state COMMENTED, zero inline comments, body: "There are no review comments, so I have no additional feedback to provide."). There is nothing actionable to close out from it, so this is effectively the first substantive review of the PR.

## Recommendation

**Approve.** Design is sound and there are no correctness, security, or merge-blocking issues. The two minor findings — the missing workforce-union test and the opaque comment reference — are worth addressing before or shortly after merge, but neither should hold the PR.

PR #1186, opened 2026-08-07
fix: replace flaky wall-clock budgets in Pytest Fast (web) tests

### LINKED ISSUE #1152: Pytest Fast (web): flaky wall-clock budgets cause unrelated TimeoutError failures
## Problem

`Pytest Fast (web)` fails intermittently with `TimeoutError` on tests that assert against hard wall-clock budgets. Two distinct tests failed this way on consecutive CI runs of an unrelated PR (#1122), each passing on rerun and passing locally:

- `tests/web/api/test_workforce_share_link.py::test_ws_append_syncs_workforce_run_back_to_running` — `asyncio.wait_for(..., timeout=1.0)`
- `tests/web/api/test_websocket_owner_actor.py::test_chat_turn_releases_one_slot_pool_before_task_info_broadcast` — `asyncio.wait_for(..., timeout=1.0)` plus a deliberate `time.sleep(0.05)`, `pool_timeout=0.25`, and `assert ticks >= 3` on a 0.01s ticker

Both were verified unrelated to the PR's changes (in the second case the code under test was mocked out entirely) and both passed locally under repeated runs — 8/8 and 10/10 respectively. On a shared runner a 1-second budget is simply thin, and the event-loop tick assertions add a second timing dependency.

## Impact

Every affected PR needs a manual rerun, and each red run costs a reviewer the effort of proving the failure is unrelated to their change. It also erodes the signal: a genuine regression in these paths looks identical to the flake.

## Direction

For the tests that are really asserting "preparation did not block the event loop" or "the pool was released before X", the property is ordering/non-blocking, not absolute latency:

1. Raise the hard `wait_for` budgets (1.0s → a few seconds) — these bounds exist to stop a hang, not to measure speed.
2. Where possible, replace the timing proxy with a deterministic signal (an `asyncio.Event` set at the point of interest, or a recorded ordering of events) instead of tick counting.
3. Keep the tight `pool_timeout` where it is load-bearing for the assertion, but do not couple it to the outer wall-clock budget.

## Acceptance

- The two tests above pass reliably under CI-level load (e.g. repeated runs with the runner under contention).
- The non-blocking / release-ordering properties they guard are still enforced, not weakened into vacuity.

codeacme17:
## Problem

`Pytest Fast (web)` fails intermittently with `TimeoutError` on tests that assert against hard wall-clock budgets. Two properties were being measured through timing proxies that a contended CI runner can violate with no regression in the code under test:

- `asyncio.wait_for(..., timeout=1.0)` anti-hang bounds, thin enough that a loaded shared runner trips them.
- `assert ticks >= 3` on a 0.01s ticker as a proxy for "preparation did not block the event loop".

Closes #1152.

## Changes

**Tick counting is replaced with a deterministic handshake.** `_EventLoopLivenessProbe` wraps the synchronous step under test so it blocks until the event loop answers back. Off the loop (in a worker thread) the probe task is scheduled and answers; on the loop thread nothing can answer and the wait expires. The verdict depends on *where* the work ran, not on how fast the machine is, and the assertion is now scoped to the preparation step rather than to the whole handler.

This also let the deliberate `time.sleep(0.05)` calls go away — the handshake creates the overlap window on its own.

**Remaining budgets are named for what they are.** `_THREAD_SIGNAL_DEADLINE_SECONDS = 10.0` and `_HANDLER_DEADLINE_SECONDS = 30.0`. These are anti-hang bounds, not latency assertions; a thread-signal wait must expire before the handler deadline so a stall surfaces as the specific assertion that owns the signal rather than a bare `TimeoutError`.

**Count-based polling becomes deadline-based.** `for _ in range(100): await asyncio.sleep(0.01)` gave the detached APPEND command a budget that shrank precisely when the runner was slowest. It now polls against a `loop.time()` deadline, and the failure message reports the observed statuses.

`pool_timeout=0.25` is left alone, per the issue's direction 3: it is load-bearing for overlap detection and only fires when checkouts actually collide.

## Scope beyond the two tests named in the issue

- `test_pause_accepted_wait_releases_one_slot_pool_before_previous_run` carried the identical pattern (`time.sleep(0.05)` + `pool_timeout=0.25` + `wait_for(1.0)` + `ticks >= 3`). Fixing only the two reported tests would have left the same flake live.
- `test_delivery_failure_persistence_drains_before_cancellation` had `await asyncio.wait_for(asyncio.to_thread(persistence_started.wait, 1), timeout=1)` with no assertion on the result. On a slow runner the inner wait returns `False` and the test proceeds to cancel before persistence started, silently changing what the rest of the test proves. Now asserted.

## Verification

Local runs on macOS with Python 3.12 and `uv sync --all-extras`:

- Full `tests/web` with the CI invocation (`-m "not slow and not postgresql" -n 4 --dist=loadscope`): 0 failures. The 20 skips are all local-environment gates (no Docker daemon, no Postgres, no Pub/Sub emulator).
- `ruff check` and `ruff format --check` clean on both files.
- **Non-vacuity.** With `run_db_io_cancellation_safe` temporarily patched to run its operation inline on the event loop instead of via `asyncio.to_thread`, both probe tests fail with their named assertion (`... blocked the asyncio event loop`) rather than a bare `TimeoutError`, confirming the 10s/30s ordering behaves as intended.
- **Under contention.** 20 busy loops saturating 10 cores (load average ~30-40), the three tests repeated 8 times: 8/8 green.

For the record, the old `ticks >= 3` assertion also caught the injected regression above. The window-scoping weakness is real in principle, but this change's demonstrated benefit is determinism, not broader detection.

## Not included

`tests/web/api/test_websocket_owner_actor.py` still has 13 further `for _ in range(100): await asyncio.sleep(0.01)` poll loops with the same count-as-budget shape. None have been reported flaky and they fall outside this issue's acceptance criteria, so they are left for a separate change.

### DIFF
--- a/tests/web/api/test_websocket_owner_actor.py
+++ b/tests/web/api/test_websocket_owner_actor.py
@@ -104,10 +104,11 @@ def _task(db, owner_id: int, status: TaskStatus = TaskStatus.RUNNING) -> Task:
 # Anti-hang bounds, not latency assertions. Nothing here is measuring speed:
 # these handlers finish in milliseconds, and the budgets only exist so a
 # deadlock fails as a test error instead of hanging the run. They are therefore
-# set far above any plausible runtime on a contended CI runner, and a
-# cross-thread handshake must expire *before* the handler deadline so the
-# failure is reported by the specific assertion rather than a bare TimeoutError.
-_HANDSHAKE_DEADLINE_SECONDS = 10.0
+# set far above any plausible runtime on a contended CI runner. A wait on a
+# signal crossing the loop/worker-thread boundary must expire *before* the
+# handler deadline, so a stall is reported by the specific assertion that owns
+# the signal rather than by a bare TimeoutError on the whole handler.
+_THREAD_SIGNAL_DEADLINE_SECONDS = 10.0
 _HANDLER_DEADLINE_SECONDS = 30.0
 
 
@@ -116,10 +117,13 @@ class _EventLoopLivenessProbe:
 
     Wrapping a synchronous step in :meth:`gate` makes that step block until the
     event loop answers back. If the step runs off the loop (in a worker thread)
-    the probe task answers immediately and ``loop_ran_during_step`` is True. If
-    the step runs *on* the loop thread, nothing can answer, the handshake
-    expires and the flag stays False. The verdict therefore depends on where
-    the work ran, not on how fast the machine is.
+    the probe task is scheduled on the next loop pass and answers, leaving
+    ``loop_ran_during_step`` True. If the step runs *on* the loop thread,
+    nothing can answer, the wait expires and the flag stays False. The verdict
+    therefore depends on where the work ran, not on how fast the machine is.
+
+    The gate is entered once per test. Both events latch, so a second gated
+    call is answered from the first handshake rather than probed again.
     """
 
     def __init__(self) -> None:
@@ -153,7 +157,7 @@ class _EventLoopLivenessProbe:
             self.step_ran = True
             self._step_entered.set()
             self.loop_ran_during_step = self._loop_answered.wait(
-                timeout=_HANDSHAKE_DEADLINE_SECONDS
+                timeout=_THREAD_SIGNAL_DEADLINE_SECONDS
             )
             return step(*args, **kwargs)
 
@@ -472,9 +476,9 @@ async def test_chat_turn_releases_one_slot_pool_before_task_info_broadcast(
         begin_turn,
     )
 
-    # The gate replaces the artificial sleep that used to make loop blocking
-    # observable. Overlap detection is separate and unchanged: the one-slot
-    # pool raises on a task-info checkout that collides with preparation.
+    # The gate holds preparation open long enough for loop blocking to be
+    # observable. Overlap detection is a separate mechanism: the one-slot pool
+    # raises on a task-info checkout that collides with preparation.
     async with _EventLoopLivenessProbe() as probe:
         monkeypatch.setattr(
             websocket_api,
@@ -948,7 +952,7 @@ async def test_missing_task_cancel_after_atomic_create_never_leaves_pending(
     def blocked_prepare_turn(**kwargs):
         preparation = prepare_turn(**kwargs)
         worker_finished.set()
-        assert release_worker.wait(timeout=_HANDSHAKE_DEADLINE_SECONDS)
+        assert release_worker.wait(timeout=_THREAD_SIGNAL_DEADLINE_SECONDS)
         return preparation
 
     schedule = AsyncMock()
@@ -978,7 +982,7 @@ async def test_missing_task_cancel_after_atomic_create_never_leaves_pending(
     )
     try:
         assert await asyncio.to_thread(
-            worker_finished.wait, _HANDSHAKE_DEADLINE_SECONDS
+            worker_finished.wait, _THREAD_SIGNAL_DEADLINE_SECONDS
         )
         handler.cancel()
         release_worker.set()
@@ -1819,7 +1823,7 @@ async def test_live_marker_cancellation_does_not_cancel_registered_handoff(
 
     def mark_delivery(*_args, **_kwargs) -> None:
         marker_started.set()
-        assert marker_release.wait(timeout=_HANDSHAKE_DEADLINE_SECONDS)
+        assert marker_release.wait(timeout=_THREAD_SIGNAL_DEADLINE_SECONDS)
 
     async def resume_forever(*_args, **_kwargs) -> None:
         try:
@@ -1860,7 +1864,9 @@ async def test_live_marker_cancellation_does_not_cancel_registered_handoff(
                 },
             )
         )
-        assert await asyncio.to_thread(marker_started.wait, 5)
+        assert await asyncio.to_thread(
+            marker_started.wait, _THREAD_SIGNAL_DEADLINE_SECONDS
+        )
         handling.cancel()
         marker_release.set()
         with pytest.raises(asyncio.CancelledError):
@@ -2179,7 +2185,7 @@ async def test_delivery_failure_persistence_drains_before_cancellation(
 
     def blocking_mark_delivery(*_args, **_kwargs) -> None:
         persistence_started.set()
-        assert allow_persistence.wait(timeout=_HANDSHAKE_DEADLINE_SECONDS)
+        assert allow_persistence.wait(timeout=_THREAD_SIGNAL_DEADLINE_SECONDS)
         persistence_finished.set()
 
     with (
@@ -2206,7 +2212,7 @@ async def test_delivery_failure_persistence_drains_before_cancellation(
         # Without the assert a slow runner silently proceeds to cancel before
         # persistence started, which changes what the rest of the test proves.
         assert await asyncio.to_thread(
-            persistence_started.wait, _HANDSHAKE_DEADLINE_SECONDS
+            persistence_started.wait, _THREAD_SIGNAL_DEADLINE_SECONDS
         )
         handling.cancel()
         await asyncio.sleep(0)
--- a/tests/web/api/test_websocket_owner_actor.py
+++ b/tests/web/api/test_websocket_owner_actor.py
@@ -11,8 +11,10 @@ from __future__ import annotations
 
 import asyncio
 import threading
+from collections.abc import Callable
 from dataclasses import is_dataclass
 from types import SimpleNamespace
+from typing import Any, TypeVar
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
@@ -111,52 +113,36 @@ def _task(db, owner_id: int, status: TaskStatus = TaskStatus.RUNNING) -> Task:
 _THREAD_SIGNAL_DEADLINE_SECONDS = 10.0
 _HANDLER_DEADLINE_SECONDS = 30.0
 
+_T = TypeVar("_T")
+
 
 class _EventLoopLivenessProbe:
     """Deterministic replacement for counting event-loop ticks.
 
-    Wrapping a synchronous step in :meth:`gate` makes that step block until the
-    event loop answers back. If the step runs off the loop (in a worker thread)
-    the probe task is scheduled on the next loop pass and answers, leaving
-    ``loop_ran_during_step`` True. If the step runs *on* the loop thread,
-    nothing can answer, the wait expires and the flag stays False. The verdict
-    therefore depends on where the work ran, not on how fast the machine is.
-
-    The gate is entered once per test. Both events latch, so a second gated
-    call is answered from the first handshake rather than probed again.
+    Wrapping a synchronous step in :meth:`gate` makes that step queue an answer
+    on the event loop and then block until the loop delivers it. Off the loop
+    (in a worker thread) the loop runs the callback on its next pass. On the
+    loop thread the queued callback cannot run, because the loop thread is the
+    one blocking, so the wait expires. The verdict therefore depends on where
+    the work ran, not on how fast the machine is.
     """
 
     def __init__(self) -> None:
-        self._step_entered = threading.Event()
-        self._loop_answered = threading.Event()
-        self._task: asyncio.Task[None] | None = None
+        # Captured here rather than inside the gate: the gated step may run in
+        # a worker thread, where get_running_loop() raises.
+        self._loop = asyncio.get_running_loop()
+        self._answered = threading.Event()
         self.step_ran = False
         self.loop_ran_during_step = False
 
-    async def __aenter__(self) -> _EventLoopLivenessProbe:
-        async def _answer() -> None:
-            while not self._step_entered.is_set():
-                await asyncio.sleep(0.001)
-            self._loop_answered.set()
+    def gate(self, step: Callable[..., _T]) -> Callable[..., _T]:
+        """Wrap ``step`` so it holds until the event loop answers."""
 
-        self._task = asyncio.create_task(_answer())
-        return self
-
-    async def __aexit__(self, *_exc_info: object) -> None:
-        assert self._task is not None
-        self._task.cancel()
-        try:
-            await self._task
-        except asyncio.CancelledError:
-            pass
-
-    def gate(self, step):
-        """Wrap ``step`` so it holds until the loop answers the handshake."""
-
-        def gated(*args, **kwargs):
+        def gated(*args: Any, **kwargs: Any) -> _T:
+            assert not self.step_ran, "the probe gates a single call"
             self.step_ran = True
-            self._step_entered.set()
-            self.loop_ran_during_step = self._loop_answered.wait(
+            self._loop.call_soon_threadsafe(self._answered.set)
+            self.loop_ran_during_step = self._answered.wait(
                 timeout=_THREAD_SIGNAL_DEADLINE_SECONDS
             )
             return step(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


tests/web/api/test_websocket_owner_actor.py
```diff
@@ -102,6 +101,73 @@ def _task(db, owner_id: int, status: TaskStatus = TaskStatus.RUNNING) -> Task:
...
+    ``loop_ran_during_step`` True. If the step runs *on* the loop thread,
+    nothing can answer, the wait expires and the flag stays False. The verdict
+    therefore depends on where the work ran, not on how fast the machine is.
+
+    The gate is entered once per test. Both events latch, so a second gated
+    call is answered from the first handshake rather than probed again.
+    """
+
+    def __init__(self) -> None:
+        self._step_entered = threading.Event()
```
rogercloud: Minor: `_step_entered` and `_loop_answered` are only ever `.set()`, never `.clear()`ed. A second `gate()` call on the same probe instance would read `loop_ran_during_step=True` straight from the first handshake regardless of the second call's actual behavior, silently masking a regression. Not live today (each test constructs a fresh probe and calls `gate()` exactly once), but a latent footgun for future edits. Consider a guard flag that asserts single use, or an explicit single-use note in the docstring.

tests/web/api/test_websocket_owner_actor.py
```diff
@@ -102,6 +101,73 @@ def _task(db, owner_id: int, status: TaskStatus = TaskStatus.RUNNING) -> Task:
...
 
+# Anti-hang bounds, not latency assertions. Nothing here is measuring speed:
+# these handlers finish in milliseconds, and the budgets only exist so a
+# deadlock fails as a test error instead of hanging the run. They are therefore
+# set far above any plausible runtime on a contended CI runner. A wait on a
+# signal crossing the loop/worker-thread boundary must expire *before* the
+# handler deadline, so a stall is reported by the specific assertion that owns
+# the signal rather than by a bare TimeoutError on the whole handler.
+_THREAD_SIGNAL_DEADLINE_SECONDS = 10.0
+_HANDLER_DEADLINE_SECONDS = 30.0
```
rogercloud, line 114: Minor: this constant (30.0) is the same value as `_DEADLINE_SECONDS` in `tests/web/api/test_workforce_share_link.py:42`, defined independently with no shared source. More substantively, in that other file the single constant is reused across four semantically distinct sequential waits in one test (event wait, shielded handler wait, poll deadline, teardown wait) — worst case all four expiring in sequence is ~120s before the test reports anything, risking an external CI timeout killing the test before its own diagnostic assertion prints. Consider splitting into distinct, independently-tunable constants and sharing one deadline module between the two files.

tests/web/api/test_websocket_owner_actor.py
```diff
@@ -399,56 +465,44 @@ async def send_text(self, message: str) -> None:
...
-                    "user": SimpleNamespace(id=actor_user_id, is_admin=False),
-                    "files": [],
-                },
-            ),
-            timeout=1.0,
+    # The gate holds preparation open long enough for loop blocking to be
+    # observable. Overlap detection is a separate mechanism: the one-slot pool
+    # raises on a task-info checkout that collides with preparation.
+    async with _EventLoopLivenessProbe() as probe:
+        monkeypatch.setattr(
```
rogercloud: Minor: `monkeypatch.setattr` here is nested inside `async with _EventLoopLivenessProbe() as probe:`, which reads as scoping the patch to that block. It isn't — `monkeypatch` reverts at fixture teardown, while the probe's background answering task is cancelled in `__aexit__`. Any later call into the patched target within the same test would hang for the full `_THREAD_SIGNAL_DEADLINE_SECONDS` (10s) and then quietly report `loop_ran_during_step=False`. Nothing in this PR triggers it — the test body ends right after the block — but it's a trap for future edits. Consider hoisting the `monkeypatch.setattr` above the `async with`, or noting the real scope in a comment. (Same pattern at line 594.)

tests/web/api/test_websocket_owner_actor.py
```diff
@@ -102,6 +101,73 @@ def _task(db, owner_id: int, status: TaskStatus = TaskStatus.RUNNING) -> Task:
...
+
+    async def __aexit__(self, *_exc_info: object) -> None:
+        assert self._task is not None
+        self._task.cancel()
+        try:
+            await self._task
+        except asyncio.CancelledError:
+            pass
+
+    def gate(self, step):
```
rogercloud: Nit: every other method on this class (`__init__`, `__aenter__`, `__aexit__`, `assert_loop_stayed_responsive`) is annotated; `gate` and its inner `gated` closure are not. The file already has `from __future__ import annotations`, so annotating is free — consistency only.

tests/web/api/test_workforce_share_link.py
```diff
@@ -799,10 +806,19 @@ async def _noop() -> None:
...
 
-                for _ in range(100):
+                # The command is applied by a detached background task, so the
+                # only observable signal is the committed row. Poll against a
+                # wall-clock deadline rather than a fixed iteration count: a
+                # loaded runner slows each iteration down, which would shrink a
+                # count-based budget exactly when it needs to be widest.
+                loop = asyncio.get_running_loop()
+                deadline = loop.time() + _DEADLINE_SECONDS
+                while True:
```
rogercloud, line 821: Nit: on the failure path this polls the full 30s at the existing 0.01s interval — roughly 3000 session open/close cycles versus the old fixed cap of 100. Each iteration opens and closes its own session in try/finally, so there's no leak or held connection, and it only fires on an already-red run — but the load increase is real (~30x). Worth knowing, not worth changing on its own.

rogercloud (approved):
## Summary

This test-only change removes wall-clock duration as a proxy for non-timing properties in two `tests/web/api` modules. The old code asserted `ticks >= 3` on a 0.01s ticker to prove "preparation didn't block the event loop", and bounded work with a 1.0s `asyncio.wait_for` — both of which a contended CI runner trips with no underlying regression (issue #1152). In their place: a deterministic `threading.Event`/`asyncio` handshake (`_EventLoopLivenessProbe`) that directly proves a synchronous step ran off the loop, a deadline-based poll replacing a count-based one, and named anti-hang constants that are deliberately generous because they no longer stand in for latency assertions. The change also fixes two additional tests exhibiting the same pattern beyond the two named in the issue, and adds a previously-missing assertion on a `wait_for` result in `test_delivery_failure_persistence_drains_before_cancellation` that could silently pass on a slow runner.

## Approach verdict: sound

The three moves map directly onto the issue's stated directions: raise the hard anti-hang bounds, replace timing proxies with deterministic signals, and keep the load-bearing `pool_timeout=0.25` decoupled from the outer budget (verified unchanged from base, and still firing correctly as a tripwire).

One design concern was raised during review and then dropped after verification: that `_EventLoopLivenessProbe` reinvents the `threading.get_ident()` comparison idiom already used elsewhere in the same file. That idiom is left untouched everywhere it appears; the probe instead replaces the flaky wall-clock ticker in these same two tests, and asserts a strictly stronger property — a thread-ident check only proves code ran on a different OS thread, and would still pass if work were correctly dispatched to a worker thread while the loop thread synchronously blocked on the result. The probe correctly fails that case, so it's solving a real gap rather than adding scaffolding over a better existing alternative.

## Findings

None are blocking; all are minor or nits — see inline comments.

- `tests/web/api/test_websocket_owner_actor.py:138`: `_answer()` busy-polls a `threading.Event` every 1ms to bridge a worker-thread signal onto the loop. `loop.call_soon_threadsafe(asyncio_event.set)` from the worker side plus `await asyncio_event.wait()` on the loop side is behaviorally equivalent (same on-loop/off-loop distinction, no polling-granularity delay) and removes the polling loop and its magic `0.001` interval. Line count is roughly flat — two event objects are still needed, one just becomes an `asyncio.Event`.
- `tests/web/api/test_websocket_owner_actor.py:115-168`: `_EventLoopLivenessProbe` is a 54-line class whose only externally-used members are `gate` and `assert_loop_stayed_responsive`, at exactly two structurally identical call sites. Nothing subclasses, mocks, or otherwise depends on it being a class, so it converts to a single `@contextlib.asynccontextmanager` function yielding a small closure-based object (~40-46 lines) — a shape that fits its single-purpose, twice-used role better than class scaffolding.

Net effect: roughly flat to modestly negative on line count. Real readability wins, but modest — neither is a reason to hold the PR.

## Verification performed

- Ran the CI-equivalent invocation locally: `pytest tests/web/api/test_websocket_owner_actor.py tests/web/api/test_workforce_share_link.py -m "not slow and not postgresql" -n 4 --dist=loadscope` → 85 passed, 0 failed (15.26s).
- Independently re-ran the mutation check described in the PR: patching the code under test to run synchronously on the loop instead of via `asyncio.to_thread` makes both probe-based tests fail with their specific diagnostic assertion (not a bare `TimeoutError`), and they pass cleanly on revert — the new mechanism is non-vacuous.
- Confirmed `pool_timeout=0.25` is unmodified from base and still trips correctly in isolation — verified directly rather than taken on faith from the PR description.

All review criteria were evaluated: code quality, correctness, test quality/coverage, and design. On documentation: the repo carries no documented convention for test timeout or signalling patterns, so this PR owes no doc update; the inline design-intent comment explaining that the 10s/30s bounds are anti-hang guards rather than latency assertions is sufficient and accurate.

## Recommendation

Approve. The diagnosis is correct, the replacement mechanism is deterministic and proven non-vacuous, the load-bearing tight timeout is preserved, and the deferred cleanup (13 remaining count-based polls) is explicitly scoped in the PR description rather than forgotten. The items below are optional follow-ups.

codeacme17:
Thanks for the thorough review — especially for independently re-running the mutation check rather than taking the PR description on faith.

Pushed 0c665fc, which takes S1 and I4. Summary of what changed and what I declined, with reasoning per item in the inline threads.

**S1 — taken, in a stronger form than suggested.** Rather than swapping the polling loop for an `asyncio.Event` set via `call_soon_threadsafe`, the background answering task is gone entirely. The gated step now does:

```python
self._loop.call_soon_threadsafe(self._answered.set)
self.loop_ran_during_step = self._answered.wait(timeout=_THREAD_SIGNAL_DEADLINE_SECONDS)
```

Off the loop, the loop runs the callback on its next pass. On the loop thread the callback lands in the ready queue but cannot run, because the loop thread is the one blocked in `wait()`. That is the same distinction the probe always drew, with the background task, the async context manager, one of the two events and the magic `0.001` all removed.

Re-verified after the change: the mutation (patching `run_db_io_cancellation_safe` to run its operation inline on the loop instead of via `asyncio.to_thread`) still makes both tests fail with their named assertion, not a bare `TimeoutError`. Full `tests/web` with the CI invocation is green, and the two probe tests were repeated 6x.

**I4 — taken.** `gate` and `gated` are annotated, with a module-level `_T = TypeVar("_T")`.

**I3 — resolved as a side effect.** Dropping the context manager removes the `async with` nesting entirely, so the misleading scope reading is gone rather than papered over with a comment.

**I1 — the docstring note became an assert.** `gated` now opens with `assert not self.step_ran, "the probe gates a single call"`, which is the guard-flag option you offered. A silent stale `True` is now a loud failure.

**S2 — declined.** With the answering task gone the class is ~20 lines with two attributes and two methods, and no longer manages anything async. An `asynccontextmanager` yielding a closure object would be more code and less clear at that size.

**I2 — declined; I believe the finding is incorrect.** Details in the thread, but briefly: the four waits cannot expire in sequence, so the worst case is ~60s rather than ~120s, and the CI-timeout premise does not hold — `timeout-minutes: 30` is on the `e2e` job (ci.yml:586), not `pytest-fast` (ci.yml:231), which has no job timeout, and `pytest-timeout` is not installed.

**I5 — declined, agreeing with your own verdict.** Left as-is.

tests/web/api/test_websocket_owner_actor.py
codeacme17: Taken — this is now an assert rather than a note. `gated` opens with:

```python
assert not self.step_ran, "the probe gates a single call"
```

You were right that the latch was real: I reproduced it before changing anything, and a second gated call on the loop thread did return `loop_ran_during_step=True` from the first handshake. Since the failure mode is silently masking a regression, the loud version is worth the line even though nothing triggers it today.

Note the surrounding class changed shape in 0c665fc (the background answering task is gone), so the two latching events are now a single `_answered` event — but the single-use property and this guard are unchanged by that.

tests/web/api/test_websocket_owner_actor.py
codeacme17: Resolved by 0c665fc, though not via either suggested route.

Removing the background answering task removed the `async with` along with it, so both call sites are now flat:

```python
probe = _EventLoopLivenessProbe()
monkeypatch.setattr(websocket_api, "_prepare_websocket_turn_sync", probe.gate(prepare_turn))
try:
    ...
```

The patch is no longer nested inside anything that reads as scoping it, so there is no misleading scope to comment on. Hoisting above the `async with` would have worked too, but it required splitting construction from entry to keep the answering task armed before the gated step ran — which was uglier than just not having a context manager.

tests/web/api/test_websocket_owner_actor.py
codeacme17: Taken in 0c665fc: `gate(self, step: Callable[..., _T]) -> Callable[..., _T]` and `gated(*args: Any, **kwargs: Any) -> _T`, with a module-level `_T = TypeVar("_T")`.

Agreed it is unenforced — the mypy hook runs `--package xagent` (.pre-commit-config.yaml:17), so `tests/` is never type-checked. Consistency with the file's other annotated methods is the whole justification, which is enough for a free change.

tests/web/api/test_workforce_share_link.py
codeacme17: Agreed on all three points, and leaving it as-is.

Confirming the arithmetic: 30s / 0.01s is indeed ~3000 iterations versus the old fixed 100. `_direct_db_session` is `next(get_db())` against the SQLite test DB and the loop closes it in `finally` every iteration, so there is no leak or held connection, and this only runs on an already-red run.

I considered backing the interval off and decided against it for a reason worth recording: a growing interval would widen the gap between the property becoming true and the test noticing, which makes the diagnostic *less* precise on exactly the runs where you are reading it. Extra branching on a failure-only path to save some SQLite opens is not a good trade.

tests/web/api/test_websocket_owner_actor.py
codeacme17: Declining this one — I think the substantive half of the finding is incorrect. Taking the two halves separately.

**The ~120s worst case does not hold.** The four waits cannot expire in sequence, because each failure path short-circuits the rest. Reading test_workforce_share_link.py:809-849:

- Wait 1 (`turn_started`, :809) times out → `TimeoutError` propagates straight to the `finally`. Waits 2 and 3 never run. Teardown adds ≤30s. **Total ≤60s.**
- Wait 2 (`shield(handler_task)`, :810-812) times out → same, straight to `finally`. The poll never runs. **Total ≤60s.**
- Wait 2 returns normally → `handler_task` is complete, so the `finally` guard `if not handler_task.done()` (:847) is False and the teardown wait is **skipped entirely**. So the poll's 30s and the teardown's 30s are mutually exclusive, never additive. **Total ≤30s past that point.**

There is no path on which more than two of the four budgets are consumed. Real worst case is ~60s.

**The external-timeout premise also does not hold.** `timeout-minutes: 30` appears once in ci.yml, at :590, and belongs to the `e2e` job that starts at :586. The `pytest-fast` job (:231) has no `timeout-minutes`, so it inherits GitHub's 6-hour default. `pytest-timeout` is not in the dependency set and `[tool.pytest.ini_options]` sets no timeout. So nothing external is positioned to kill the test before its own diagnostic prints — which is the specific risk this finding rests on.

If either of those is wrong I would genuinely like to know, since I would rather split the constants than argue about it.

**On the duplication itself** — fair, and I did consider it. `_HANDLER_DEADLINE_SECONDS` and `_DEADLINE_SECONDS` are the same number in two files with no shared source. I left it because a shared deadline module for exactly two test files felt like more coupling than the duplication costs, and because the two files' waits are not obviously destined to move together. Happy to extract one if you would prefer it; it is a small change and I do not feel strongly.

PR #1188, opened 2026-08-07
feat: add Notion as a built-in remote-MCP connector

### LINKED ISSUE #1187: feat: Add Notion connector
### Description

Add a new Notion Connector to allow agents to access and interact with users' Notion workspace data.

Users should be able to connect their Notion account and allow agents to retrieve and use Notion content in workflows.

yiboyasss:
## Summary

Adds Notion to the built-in connector catalog as the second remote-MCP OAuth
connector (after Granola, #92a25006-era work): `transport: streamable_http`
pointing at Notion's hosted MCP server (`https://mcp.notion.com/mcp`) with
`auth.type: mcp_oauth`.

Notion supports per-user OAuth (Authorization Code + PKCE) with Dynamic
Client Registration, so this needs **no static client credentials** — no
Notion developer app, no `NOTION_*` env vars, and no `oauth_providers` row.
No local tool module either: Notion's server exposes its own workspace tools
(search, fetch, page/database create and update). Zero frontend changes —
the connector UI is fully data-driven.

## Changes

- `src/xagent/web/builtin_mcp_registry.py` — add the `notion` catalog entry
  (Productivity, visible in connector, same launch_config shape as granola)
- `src/xagent/migrations/versions/20260807_seed_notion_mcp_app.py` — seed
  migration (catalog row only), chained onto
  `20260806_seed_google_sheets_mcp_app`; idempotent upgrade, downgrade
  removes only the catalog row
- `tests/alembic/test_20260807_seed_notion_mcp_app.py` — upgrade/downgrade
  round-trip, idempotency, migration-snapshot-vs-registry drift guard, and
  mcp_oauth classification tests
- `tests/web/api/test_public_mcp_connector_visibility.py` — notion
  launch_config assertions in the remote-MCP block; the granola-only
  classification test is parametrized over both remote connectors

## Testing

- `pytest tests/alembic/test_20260807_seed_notion_mcp_app.py
  tests/alembic/test_20260731_seed_granola_mcp_app.py
  tests/web/api/test_public_mcp_connector_visibility.py` — all passing
- `alembic upgrade head` → `downgrade -1` → `upgrade head` verified on a
  clean SQLite database; single head confirmed
- Verified end to end against Notion's production server: DCR client
  registration, consent + PKCE token exchange, and live tool calls
  (`notion_fetch`) all succeed with no pre-registered client

### DIFF
--- a/tests/alembic/test_20260807_seed_notion_mcp_app.py
+++ b/tests/alembic/test_20260807_seed_notion_mcp_app.py
@@ -1,6 +1,7 @@
 """Tests for the Notion remote-MCP connector seed migration."""
 
 import importlib.util
+import json
 from pathlib import Path
 from unittest.mock import patch
 
@@ -68,8 +69,17 @@ def test_upgrade_inserts_notion(tmp_path):
         ).first()
         assert row[0] == "streamable_http"
         assert row[1] is None
-        assert "https://mcp.notion.com/mcp" in str(row[2])
-        assert "mcp_oauth" in str(row[2])
+        # Exact comparison (not substring checks): the seeded config must not
+        # carry any field beyond the remote URL and auth type — an unexpected
+        # extra field (e.g. a local command) would change how the connector
+        # is classified and launched.
+        launch_config = row[2]
+        if isinstance(launch_config, str):
+            launch_config = json.loads(launch_config)
+        assert launch_config == {
+            "url": "https://mcp.notion.com/mcp",
+            "auth": {"type": "mcp_oauth"},
+        }
 
 
 def test_upgrade_is_idempotent(tmp_path):
--- a/tests/alembic/test_20260807_seed_notion_mcp_app.py
+++ b/tests/alembic/test_20260807_seed_notion_mcp_app.py
@@ -1,0 +1,122 @@
+"""Tests for the Notion remote-MCP connector seed migration."""
+
+import importlib.util
+from pathlib import Path
+from unittest.mock import patch
+
+from alembic.migration import MigrationContext
+from alembic.operations import Operations
+from sqlalchemy import create_engine, text
+
+
+def _load_migration_module():
+    migration_file = (
+        Path(__file__).parent.parent.parent
+        / "src/xagent/migrations/versions/20260807_seed_notion_mcp_app.py"
+    )
+    spec = importlib.util.spec_from_file_location(
+        "seed_notion_migration", migration_file
+    )
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def _operations(connection):
+    return Operations(MigrationContext.configure(connection))
+
+
+def _create_table(connection):
+    connection.execute(
+        text(
+            """
+            CREATE TABLE public_mcp_apps (
+                id INTEGER PRIMARY KEY,
+                app_id VARCHAR(100) NOT NULL UNIQUE,
+                name VARCHAR(200) NOT NULL,
+                description TEXT,
+                icon VARCHAR(1000),
+                transport VARCHAR(50) NOT NULL DEFAULT 'oauth',
+                provider_name VARCHAR(50),
+                category VARCHAR(100),
+                oauth_scopes JSON,
+                is_visible_in_connector BOOLEAN NOT NULL DEFAULT 1,
+                launch_config JSON
+            )
+            """
+        )
+    )
+
+
+def _app_ids(connection):
+    return set(connection.execute(text("SELECT app_id FROM public_mcp_apps")).scalars())
+
+
+def test_upgrade_inserts_notion(tmp_path):
+    engine = create_engine(f"sqlite:///{tmp_path / 'test.db'}")
+    migration = _load_migration_module()
+    with engine.begin() as connection:
+        _create_table(connection)
+        with patch.object(migration, "op", _operations(connection)):
+            migration.upgrade()
+        assert "notion" in _app_ids(connection)
+        row = connection.execute(
+            text(
+                "SELECT transport, provider_name, launch_config FROM public_mcp_apps WHERE app_id='notion'"
+            )
+        ).first()
+        assert row[0] == "streamable_http"
+        assert row[1] is None
+        assert "https://mcp.notion.com/mcp" in str(row[2])
+        assert "mcp_oauth" in str(row[2])
+
+
+def test_upgrade_is_idempotent(tmp_path):
+    engine = create_engine(f"sqlite:///{tmp_path / 'test.db'}")
+    migration = _load_migration_module()
+    with engine.begin() as connection:
+        _create_table(connection)
+        with patch.object(migration, "op", _operations(connection)):
+            migration.upgrade()
+            migration.upgrade()  # second run must not raise or duplicate
+        rows = connection.execute(
+            text("SELECT COUNT(*) FROM public_mcp_apps WHERE app_id='notion'")
+        ).scalar()
+        assert rows == 1
+
+
+def test_seed_row_matches_registry(tmp_path):
+    """The migration snapshot and the runtime registry must define the same
+    notion row (the migration is a frozen copy; this catches drift)."""
+    from xagent.web.builtin_mcp_registry import get_builtin_public_mcp_app_rows
+
+    migration = _load_migration_module()
+    registry_row = next(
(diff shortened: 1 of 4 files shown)


tests/alembic/test_20260807_seed_notion_mcp_app.py
```diff
@@ -0,0 +1,133 @@
...
+
+def test_downgrade_removes_notion(tmp_path):
+    engine = create_engine(f"sqlite:///{tmp_path / 'test.db'}")
+    migration = _load_migration_module()
+    with engine.begin() as connection:
+        _create_table(connection)
+        with patch.object(migration, "op", _operations(connection)):
+            migration.upgrade()
+            migration.downgrade()
+        assert "notion" not in _app_ids(connection)
```
rogercloud, line 133: Nit (non-blocking, pre-existing pattern): `test_downgrade_removes_notion` only asserts `"notion" not in _app_ids(connection)`. Since the migration seeds a single `notion` row into an otherwise-empty table, this assertion cannot distinguish "downgrade deleted only the notion row" from "downgrade wiped the whole table" — both leave the table empty. The sibling test `tests/alembic/test_20260731_seed_granola_mcp_app.py:114-122` has the same gap, so this is not a regression introduced here.

Optional strengthening: insert a second, unrelated app row before calling `downgrade()` and assert it still exists afterwards, e.g.

```python
# seed an unrelated sibling row before downgrade
connection.execute(
    sa.text("INSERT INTO mcp_apps (app_id, name) VALUES ('sentinel', 'Sentinel')")
)
migration.downgrade()
app_ids = _app_ids(connection)
assert "notion" not in app_ids
assert "sentinel" in app_ids  # downgrade must be scoped to the notion row
```

(Adjust the INSERT to the table's non-nullable columns.) Fine to defer to a follow-up that fixes both this test and the Granola one together.

tests/web/api/test_public_mcp_connector_visibility.py
```diff
@@ -771,21 +771,29 @@ def test_builtin_registry_uses_runtime_available_launch_commands() -> None:
         "auth": {"type": "mcp_oauth"},
     }
 
+    # Remote MCP: Notion hosts the server itself, same shape as Granola.
+    assert rows_by_app_id["notion"]["transport"] == "streamable_http"
+    assert rows_by_app_id["notion"]["launch_config"] == {
+        "url": "https://mcp.notion.com/mcp",
+        "auth": {"type": "mcp_oauth"},
+    }
+
```
rogercloud, line 780: Nit (non-blocking, pre-existing naming mismatch): these Notion `launch_config` assertions live inside `test_builtin_registry_uses_runtime_available_launch_commands`, whose name is about local launch *commands* — but Notion, like Granola, is a remote connector with no local command at all; its `launch_config` only carries `url` + `auth`. The mismatch pre-dates this PR (the Granola block at lines 767-770 already sits here), and this PR simply adds a parallel Notion block, so nothing is made worse.

A follow-up could split the remote-connector assertions into their own test (e.g. `test_remote_mcp_apps_launch_config`) so each test name matches what it actually checks. Out of scope for this PR.

tests/alembic/test_20260807_seed_notion_mcp_app.py
```diff
@@ -0,0 +1,133 @@
...
+        # carry any field beyond the remote URL and auth type — an unexpected
+        # extra field (e.g. a local command) would change how the connector
+        # is classified and launched.
+        launch_config = row[2]
+        if isinstance(launch_config, str):
+            launch_config = json.loads(launch_config)
+        assert launch_config == {
+            "url": "https://mcp.notion.com/mcp",
+            "auth": {"type": "mcp_oauth"},
+        }
```
rogercloud, line 83: Nit, mostly for completeness (no change needed): `test_upgrade_inserts_notion` asserts only 3 of the ~10 columns on the inserted row (`transport`, `provider_name`, `launch_config`), leaving `name`, `description`, `icon`, `category`, `oauth_scopes`, and `is_visible_in_connector` unchecked at the DB level.

This is largely covered by `test_seed_row_matches_registry` (lines 100-109), which does a full dict equality check of `migration.ROW` against the registry entry and would catch migration/registry drift across all 10 columns. The only residual gap is an actual insert round-trip of those six columns, which is low value given the migration inserts `ROW` verbatim. Mentioning it so the coverage boundary is explicit rather than implicit.

rogercloud (approved):
## Summary

This PR adds Notion to the built-in connector catalog as the second remote-MCP OAuth connector, following the pattern established by Granola. It registers a `streamable_http` entry pointing at Notion's hosted MCP server with `auth.type: mcp_oauth`, plus a frozen-snapshot Alembic migration that seeds the same row into `mcp_apps`. Because Notion's server supports Authorization Code + PKCE with Dynamic Client Registration, no static client credentials, developer app, or `NOTION_*` environment variables are required, and no local tool module is needed — Notion's own server exposes the tools. The change is data-only (245 insertions / 2 deletions across 4 files), with no frontend changes.

## Approach

**Verdict: sound.** This is the minimal correct change for the goal — a pure data addition on top of an already-generic, catalog-driven remote-MCP OAuth pipeline, plus a migration that mirrors the established sibling pattern. This was traced through the dispatch code rather than taken from the PR description:

- `classify_app_auth` in `src/xagent/web/mcp_apps.py:81-113` classifies purely from `transport` + `launch_config.auth.type`; there is no connector-id special-casing to extend.
- The OAuth connect endpoint and DCR flow (`src/xagent/web/api/mcp.py:2378-2457`, `2594-2657`, `3697-3777`) gate solely on `auth_type == "mcp_oauth"`, i.e. the exact path Granola already exercises.
- The frontend branches generically on `auth_type === "mcp_oauth"` (`frontend/src/components/mcp/connect-mcp-dialog.tsx:685-690`); no hardcoded connector-id allowlist exists. "Zero frontend changes" is confirmed, not assumed.

### Framework-level observation: catalog name vs. user-created server names (informational, not a blocker)

`mcp_servers.name` is globally unique (`src/xagent/core/tools/core/mcp/model.py:35`) rather than scoped per user. `_ensure_catalog_mcp_oauth_server` (`src/xagent/web/api/mcp.py:2405-2417`) reuses an existing row with the catalog name only when transport and URL match exactly, and otherwise returns a 409. This logic is pre-existing and identical in the base commit, and it already applies to every catalog connector — nothing in this diff introduces or owns it. It is worth naming only because `notion` is a name a user is comparatively likely to have already used for a self-added custom MCP server (more so than, say, `granola`), so the practical collision rate goes up slightly with this addition. No change requested here; flagging it as a framework-level item for a possible follow-up (e.g. per-owner scoping or a reserved-name namespace for catalog connectors).

## Findings

All findings from this round are minor / nit level and are left as inline comments. None of them block merge.

## Verified fixed from prior review

The three findings raised in the two earlier automated review rounds were re-checked against the current head and are all resolved:

1. The launch_config assertion in `tests/alembic/test_20260807_seed_notion_mcp_app.py:77-83` now parses the JSON and does an exact dict comparison instead of fragile substring matching (fixed in `ba1c680c`).
2. `assert spec is not None` / `assert spec.loader is not None` (lines 21-22) now run before `module_from_spec(spec)` (line 23), so they are no longer unreachable behind a `ValueError` (fixed in `d3354874`).
3. The unused `tmp_path` parameters were removed from `test_seed_row_matches_registry` and `test_seed_row_classifies_as_mcp_oauth` (fixed in `d3354874`).

## Test results

- `pytest tests/alembic/test_20260807_seed_notion_mcp_app.py tests/alembic/test_20260731_seed_granola_mcp_app.py tests/web/api/test_public_mcp_connector_visibility.py` → **55 passed, 0 failed**.
- Migration round-trip on a fresh SQLite database: `alembic upgrade head` → `downgrade -1` → `upgrade head` completes cleanly; `alembic heads` reports exactly one head (`20260807_seed_notion_mcp_app`).
- Chain verified: `down_revision = "20260806_seed_google_sheets_mcp_app"` is the true prior head, with no competing revision claiming it.
- No hardcoded secrets, `client_id`, or `client_secret` anywhere in the diff. No leftover copy-paste `granola` strings in the Notion entry — the single `Granola` mention in `src/xagent/web/builtin_mcp_registry.py:527` is an intentional comparative comment.

## Recommendation

**Approve.** The design is sound and minimal, all prior review findings are verified fixed, the migration chain and round-trip are clean, and the remaining items are non-blocking nits or pre-existing patterns shared with the Granola sibling. Safe to merge as-is; the inline suggestions can be picked up here or in a follow-up.

PR #1190, opened 2026-08-07
test(mcp): tighten granola/notion seed-migration test coverage gaps

### LINKED ISSUE #1189: test(mcp): tighten granola/notion seed-migration test coverage gaps from #1188 review
Raised in the #1188 review (rogercloud, round 3, all marked non-blocking/optional/out-of-scope for that PR) as three small test-coverage gaps in the Notion connector's seed migration tests. Two of the three point out that the same gap pre-exists in the Granola connector's tests, which this issue also covers so both connectors get fixed together instead of drifting further apart.

## Gaps

1. **Weak downgrade assertion.** `test_downgrade_removes_notion` (`tests/alembic/test_20260807_seed_notion_mcp_app.py`) and its sibling `test_downgrade_removes_granola` (`tests/alembic/test_20260731_seed_granola_mcp_app.py`) only assert that the seeded app_id is gone after `downgrade()`. Since each migration seeds a single row into an otherwise-empty test table, this can't distinguish "downgrade deleted only its own row" from "downgrade wiped the whole table" — both leave it empty either way.

2. **Misnamed test grouping.** `test_builtin_registry_uses_runtime_available_launch_commands` (`tests/web/api/test_public_mcp_connector_visibility.py`) is named for local launch *commands*, but it also carries the granola and notion assertions — both remote connectors with no local command at all, just `url` + `auth`. The name no longer describes what the test checks now that two remote-MCP blocks live inside it.

3. **Partial column coverage in the insert test.** `test_upgrade_inserts_notion` and `test_upgrade_inserts_granola` each assert only 3 of ~10 columns on the inserted row (`transport`, `provider_name`, `launch_config`), leaving `name`, `description`, `icon`, `category`, `oauth_scopes`, and `is_visible_in_connector` unchecked at the DB-insert level. The sibling `test_seed_row_matches_registry` test does catch drift across all 10 columns against the registry, so this is a minor completeness gap rather than a real hole.

## Proposal

- Strengthen both downgrade tests to seed an unrelated sentinel row before calling `downgrade()`, and assert the sentinel survives while the connector's own row is gone.
- Split the remote-MCP (granola/notion) launch_config assertions out of `test_builtin_registry_uses_runtime_available_launch_commands` into their own test (e.g. `test_remote_mcp_apps_launch_config`) so the name matches what each test verifies.
- Extend both insert tests to assert the full seeded row (or fold that check into the existing full-row migration-vs-registry comparison) so DB-level coverage isn't implicitly relying on a different test.

## Acceptance

- `test_downgrade_removes_granola` / `test_downgrade_removes_notion` fail if a downgrade accidentally deletes rows beyond its own app_id.
- The remote-MCP launch_config assertions live in a test whose name accurately describes them.
- The insert tests for granola and notion cover the full seeded row shape at the DB level.

Refs: #1188.

yiboyasss:
Fixes #1189.

## Problem

Raised as three non-blocking nits in the #1188 review (round 3, rogercloud
— that PR was approved and has since merged as 26a654c4). Two of the three
turned out to be pre-existing gaps in
`tests/alembic/test_20260731_seed_granola_mcp_app.py`, not something #1188
introduced, so this fixes both connectors' tests together instead of
letting them drift further apart:

1. `test_downgrade_removes_{granola,notion}` only asserted the seeded
   app_id was gone after `downgrade()`. Since each migration seeds a
   single row into an otherwise-empty test table, that assertion can't
   tell "downgrade deleted only its own row" apart from "downgrade wiped
   the whole table" — both leave the app_id lookup empty either way.
2. The granola/notion `launch_config` assertions lived inside
   `test_builtin_registry_uses_runtime_available_launch_commands` — a name
   about local launch *commands*. Both connectors are remote
   (streamable_http to a hosted server) with no local command at all.
3. `test_upgrade_inserts_{granola,notion}` checked only 3 of ~10 seeded
   columns (`transport`, `provider_name`, `launch_config`) at the
   DB-insert level, leaving `name`/`description`/`icon`/`category`/
   `oauth_scopes`/`is_visible_in_connector` unchecked at that layer.

## Change

- `test_downgrade_removes_granola` / `test_downgrade_removes_notion`: insert
  an unrelated `sentinel` row before calling `downgrade()`, assert it
  survives and only the connector's own row is gone.
- New `test_remote_mcp_apps_launch_config` in
  `test_public_mcp_connector_visibility.py` carries the granola/notion
  `launch_config` assertions, moved out of the local-launch-commands test.
- `test_upgrade_inserts_granola` / `test_upgrade_inserts_notion`: select and
  compare the full seeded row (all columns) instead of three of them. This
  also folds in the earlier exact-dict-equality fix for `launch_config`
  (no more substring checks) for granola, matching what notion already had.

## Tests

`pytest tests/alembic/test_20260807_seed_notion_mcp_app.py
tests/alembic/test_20260731_seed_granola_mcp_app.py
tests/web/api/test_public_mcp_connector_visibility.py` — 58 passed.

### DIFF
--- a/tests/alembic/test_20260731_seed_granola_mcp_app.py
+++ b/tests/alembic/test_20260731_seed_granola_mcp_app.py
@@ -1,13 +1,12 @@
 """Tests for the Granola remote-MCP connector seed migration."""
 
 import importlib.util
-import json
 from pathlib import Path
 from unittest.mock import patch
 
 from alembic.migration import MigrationContext
 from alembic.operations import Operations
-from sqlalchemy import create_engine, text
+from sqlalchemy import create_engine, select, text
 
 
 def _load_migration_module():
@@ -61,51 +60,24 @@ def test_upgrade_inserts_granola(tmp_path):
         _create_table(connection)
         with patch.object(migration, "op", _operations(connection)):
             migration.upgrade()
-        assert "granola" in _app_ids(connection)
-        row = connection.execute(
-            text(
-                "SELECT name, description, icon, transport, provider_name,"
-                " category, oauth_scopes, is_visible_in_connector, launch_config"
-                " FROM public_mcp_apps WHERE app_id='granola'"
+        # Full-row comparison through the migration's own typed table object:
+        # JSON columns deserialize and booleans come back as real bools, so
+        # the persisted row compares to ROW directly — no field is left
+        # unchecked at the DB level. ROW's content is itself pinned against
+        # the registry by test_seed_row_matches_registry, so deriving the
+        # expectation from it keeps identical coverage without a third
+        # hand-typed copy of the seed data.
+        row = (
+            connection.execute(
+                select(migration.PUBLIC_MCP_APPS_TABLE).where(
+                    migration.PUBLIC_MCP_APPS_TABLE.c.app_id == "granola"
+                )
             )
-        ).first()
-        oauth_scopes = row[6]
-        if isinstance(oauth_scopes, str):
-            oauth_scopes = json.loads(oauth_scopes)
-        # Exact comparison (not substring checks): the seeded config must not
-        # carry any field beyond the remote URL and auth type — an unexpected
-        # extra field (e.g. a local command) would change how the connector
-        # is classified and launched.
-        launch_config = row[8]
-        if isinstance(launch_config, str):
-            launch_config = json.loads(launch_config)
-        # Full-row comparison, not just transport/provider_name/launch_config:
-        # a drifted name/description/icon/category/scopes/visibility would
-        # otherwise pass this test silently.
-        assert (
-            row[0],
-            row[1],
-            row[2],
-            row[3],
-            row[4],
-            row[5],
-            oauth_scopes,
-            bool(row[7]),
-            launch_config,
-        ) == (
-            "Granola",
-            "Connect to Granola to search and read your meeting notes, transcripts and action items through Granola's hosted MCP server.",
-            "https://www.google.com/s2/favicons?domain=granola.ai&sz=128",
-            "streamable_http",
-            None,
-            "Productivity",
-            None,
-            True,
-            {
-                "url": "https://mcp.granola.ai/mcp",
-                "auth": {"type": "mcp_oauth"},
-            },
+            .mappings()
+            .first()
         )
+        assert row is not None
+        assert dict(row) == migration.ROW
 
 
 def test_upgrade_is_idempotent(tmp_path):
@@ -154,15 +126,18 @@ def test_downgrade_removes_granola(tmp_path):
         _create_table(connection)
         with patch.object(migration, "op", _operations(connection)):
             migration.upgrade()
-            # A sentinel row unrelated to this migration must survive the
-            # downgrade — otherwise "granola" missing from _app_ids could
-            # equally mean the whole table was wiped, not just its own row.
-            connection.execute(
-                text(
-                    "INSERT INTO public_mcp_apps (app_id, name) VALUES ('sentinel', 'Sentinel')"
-                )
+        # A sentinel row unrelated to this migration must survive the
+        # downgrade — otherwise "granola" missing from _app_ids could
+        # equally mean the whole table was wiped, not just its own row.
+        # transport is set explicitly: the real column is NOT NULL with no
+        # server default, so the insert must not lean on the default this
+        # test's own _create_table happens to declare.
+        connection.execute(
+            text(
+                "INSERT INTO public_mcp_apps (app_id, name, transport)"
+                " VALUES ('sentinel', 'Sentinel', 'oauth')"
             )
+        )
+        with patch.object(migration, "op", _operations(connection)):
             migration.downgrade()
-        app_ids = _app_ids(connection)
-        assert "granola" not in app_ids
-        assert "sentinel" in app_ids
+        assert _app_ids(connection) == {"sentinel"}
--- a/tests/alembic/test_20260731_seed_granola_mcp_app.py
+++ b/tests/alembic/test_20260731_seed_granola_mcp_app.py
@@ -1,6 +1,7 @@
 """Tests for the Granola remote-MCP connector seed migration."""
 
 import importlib.util
+import json
 from pathlib import Path
 from unittest.mock import patch
 
@@ -63,13 +64,48 @@ def test_upgrade_inserts_granola(tmp_path):
         assert "granola" in _app_ids(connection)
         row = connection.execute(
             text(
-                "SELECT transport, provider_name, launch_config FROM public_mcp_apps WHERE app_id='granola'"
+                "SELECT name, description, icon, transport, provider_name,"
+                " category, oauth_scopes, is_visible_in_connector, launch_config"
+                " FROM public_mcp_apps WHERE app_id='granola'"
             )
         ).first()
-        assert row[0] == "streamable_http"
-        assert row[1] is None
-        assert "https://mcp.granola.ai/mcp" in str(row[2])
-        assert "mcp_oauth" in str(row[2])
+        oauth_scopes = row[6]
+        if isinstance(oauth_scopes, str):
+            oauth_scopes = json.loads(oauth_scopes)
+        # Exact comparison (not substring checks): the seeded config must not
+        # carry any field beyond the remote URL and auth type — an unexpected
+        # extra field (e.g. a local command) would change how the connector
+        # is classified and launched.
+        launch_config = row[8]
+        if isinstance(launch_config, str):
+            launch_config = json.loads(launch_config)
+        # Full-row comparison, not just transport/provider_name/launch_config:
+        # a drifted name/description/icon/category/scopes/visibility would
+        # otherwise pass this test silently.
+        assert (
+            row[0],
+            row[1],
+            row[2],
+            row[3],
+            row[4],
+            row[5],
+            oauth_scopes,
+            bool(row[7]),
+            launch_config,
+        ) == (
+            "Granola",
+            "Connect to Granola to search and read your meeting notes, transcripts and action items through Granola's hosted MCP server.",
+            "https://www.google.com/s2/favicons?domain=granola.ai&sz=128",
+            "streamable_http",
+            None,
+            "Productivity",
+            None,
+            True,
+            {
+                "url": "https://mcp.granola.ai/mcp",
+                "auth": {"type": "mcp_oauth"},
+            },
+        )
 
 
 def test_upgrade_is_idempotent(tmp_path):
@@ -118,5 +154,15 @@ def test_downgrade_removes_granola(tmp_path):
         _create_table(connection)
         with patch.object(migration, "op", _operations(connection)):
             migration.upgrade()
+            # A sentinel row unrelated to this migration must survive the
+            # downgrade — otherwise "granola" missing from _app_ids could
+            # equally mean the whole table was wiped, not just its own row.
+            connection.execute(
+                text(
+                    "INSERT INTO public_mcp_apps (app_id, name) VALUES ('sentinel', 'Sentinel')"
+                )
+            )
             migration.downgrade()
-        assert "granola" not in _app_ids(connection)
+        app_ids = _app_ids(connection)
+        assert "granola" not in app_ids
+        assert "sentinel" in app_ids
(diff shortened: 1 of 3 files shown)


tests/alembic/test_20260731_seed_granola_mcp_app.py
```diff
@@ -118,5 +154,15 @@ def test_downgrade_removes_granola(tmp_path):
         _create_table(connection)
         with patch.object(migration, "op", _operations(connection)):
             migration.upgrade()
+            # A sentinel row unrelated to this migration must survive the
+            # downgrade — otherwise "granola" missing from _app_ids could
+            # equally mean the whole table was wiped, not just its own row.
+            connection.execute(
+                text(
+                    "INSERT INTO public_mcp_apps (app_id, name) VALUES ('sentinel', 'Sentinel')"
```
rogercloud: **minor** — this `INSERT` omits `transport`, and only succeeds because the test's own `_create_table()` helper invents `transport VARCHAR(50) NOT NULL DEFAULT 'oauth'`. The real column (from `src/xagent/migrations/versions/f1427c3a7261_add_oauthprovider_and_publicmcpapp_.py`) is `NOT NULL` with **no** server-side default — the `default="oauth"` in `src/xagent/web/models/public_mcp.py` is Python-side only and does not apply to raw SQL inserts. Running this exact statement against a table built from the real `op.create_table` call raises:

```
sqlite3.IntegrityError: NOT NULL constraint failed: public_mcp_apps.transport
```

The assertion under test is still valid, but the new sentinel row leans on schema drift to exist at all. Suggest making it explicit:

```python
"INSERT INTO public_mcp_apps (app_id, name, transport)"
" VALUES ('sentinel', 'Sentinel', 'oauth')"
```

tests/alembic/test_20260807_seed_notion_mcp_app.py
```diff
@@ -129,5 +155,15 @@ def test_downgrade_removes_notion(tmp_path):
         _create_table(connection)
         with patch.object(migration, "op", _operations(connection)):
             migration.upgrade()
+            # A sentinel row unrelated to this migration must survive the
+            # downgrade — otherwise "notion" missing from _app_ids could
+            # equally mean the whole table was wiped, not just its own row.
+            connection.execute(
+                text(
+                    "INSERT INTO public_mcp_apps (app_id, name) VALUES ('sentinel', 'Sentinel')"
```
rogercloud: **minor** — same as the granola file: this `INSERT` omits `transport`, which is `NOT NULL` with no server-side default in the real schema (`src/xagent/migrations/versions/f1427c3a7261_add_oauthprovider_and_publicmcpapp_.py`). It only succeeds because `_create_table()` here declares `transport ... NOT NULL DEFAULT 'oauth'`, which production does not. Suggest adding the column explicitly:

```python
"INSERT INTO public_mcp_apps (app_id, name, transport)"
" VALUES ('sentinel', 'Sentinel', 'oauth')"
```

tests/alembic/test_20260731_seed_granola_mcp_app.py
```diff
@@ -63,13 +64,48 @@ def test_upgrade_inserts_granola(tmp_path):
...
+        if isinstance(oauth_scopes, str):
+            oauth_scopes = json.loads(oauth_scopes)
+        # Exact comparison (not substring checks): the seeded config must not
+        # carry any field beyond the remote URL and auth type — an unexpected
+        # extra field (e.g. a local command) would change how the connector
+        # is classified and launched.
+        launch_config = row[8]
+        if isinstance(launch_config, str):
+            launch_config = json.loads(launch_config)
+        # Full-row comparison, not just transport/provider_name/launch_config:
```
rogercloud: **nit** — the comment's premise is not quite right for `is_visible_in_connector`. That column *does* have `server_default=true` in production (`src/xagent/migrations/versions/20260519_add_connector_visibility_to_public_mcp_apps.py`, which runs before this seed migration in the real revision graph), so this is not a fake-schema divergence.

What is true: `bool(row[7]) == True` cannot distinguish "the migration explicitly set it" from "the column defaulted to it", because the intended value and the default happen to coincide. That drift is already caught by `test_seed_row_matches_registry`, so there is no real coverage hole here — just an inaccurate rationale in the comment. Suggest softening the wording rather than changing the assertion.

tests/alembic/test_20260807_seed_notion_mcp_app.py
```diff
@@ -65,22 +65,48 @@ def test_upgrade_inserts_notion(tmp_path):
...
         # is classified and launched.
-        launch_config = row[2]
+        launch_config = row[8]
         if isinstance(launch_config, str):
             launch_config = json.loads(launch_config)
-        assert launch_config == {
-            "url": "https://mcp.notion.com/mcp",
-            "auth": {"type": "mcp_oauth"},
-        }
+        # Full-row comparison, not just transport/provider_name/launch_config:
```
rogercloud: **nit** — same as the granola file: `is_visible_in_connector` has `server_default=true` in production (`src/xagent/migrations/versions/20260519_add_connector_visibility_to_public_mcp_apps.py`), so the comment's implied fake-schema divergence does not apply. The real limitation is that the intended value and the default coincide, so this assertion cannot tell them apart — and that case is already covered by `test_seed_row_matches_registry`. Suggest correcting the comment; no assertion change needed.

tests/web/api/test_public_mcp_connector_visibility.py
```diff
@@ -764,26 +764,35 @@ def test_builtin_registry_uses_runtime_available_launch_commands() -> None:
...
 
-    # Remote MCP (no local command at all): Granola hosts the server itself.
+    assert rows_by_app_id["aws"]["launch_config"] == {
+        "command": "python",
+        "args": ["-m", "xagent.web.tools.mcp.aws"],
+        "required_env": ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_REGION"],
+    }
+
+
+def test_remote_mcp_apps_launch_config() -> None:
```
rogercloud: **nit** — the sibling tests in this area are prefixed `test_builtin_registry_...` (both this one and the neighbours call `get_builtin_public_mcp_app_rows()`). Without the prefix, `pytest -k builtin_registry` will not select this test. Suggest `test_builtin_registry_remote_mcp_apps_launch_config` for consistency.

tests/web/api/test_public_mcp_connector_visibility.py
```diff
@@ -764,26 +764,35 @@ def test_builtin_registry_uses_runtime_available_launch_commands() -> None:
...
+    """Granola and Notion have no local launch command at all — they host
+    their own MCP server and are reached over streamable_http. This is
+    intentionally split out of test_builtin_registry_uses_runtime_available_
+    launch_commands, whose name is about local launch *commands* and would
```
rogercloud: **nit** — the docstring wraps `test_builtin_registry_uses_runtime_available_launch_commands` mid-token across the line break, so a text search for the exact test name will not find this reference. Suggest rewrapping so the identifier stays on one line, e.g. break after "split out of" and put the full name on the next line.

rogercloud (approved):
## Summary

This PR closes three test-coverage gaps that were raised as non-blocking nits during the review of #1188 (merged as 26a654c4) and tracked in #1189. The granola and notion seed-migration tests now insert an unrelated sentinel row before `downgrade()` and assert it survives, so a "deleted everything" regression can no longer pass as "deleted the seeded row"; the insert tests compare the full seeded row instead of only three of the nine seeded columns; and the granola/notion `launch_config` assertions have been moved out of a test named for local launch *commands* into a dedicated `test_remote_mcp_apps_launch_config`, since both connectors are remote-only.

The changes are test-only and mechanical. All three gaps are real, and the fixes attack the actual gaps rather than their symptoms.

**Verdict: approach is acceptable, with a few reservations noted below. No blocking issues — nothing here is critical or major.**

## Design-level notes

These are structural observations that no single line owns, so they are here rather than inline.

**1. The tests build their own `public_mcp_apps` schema, and it has drifted from the real one.** The hand-rolled `_create_table()` helper in both test files (pre-existing, not touched by this PR) is not a faithful copy of the production schema. This matters for this PR because the *new* sentinel `INSERT` depends on that divergence to succeed — see the inline comment on the sentinel insert. Longer term, deriving the test table from the real migration (or from the ORM metadata) would remove a whole class of "passes only against the fake schema" failure modes. Out of scope here, but worth a tracked follow-up.

**2. The migration modules already expose a typed table object the tests could use.** Both migrations define `PUBLIC_MCP_APPS_TABLE` as an `sa.table(...)` with `sa.JSON` / `sa.Boolean` columns. Selecting through it — `sa.select(PUBLIC_MCP_APPS_TABLE).where(...).mappings().first()` — auto-deserializes the JSON columns and returns a name-keyed mapping with real `bool` / `dict` types. This was verified to work against the current code. That would eliminate the manual `json.loads` guards, the `bool()` cast, and the positional `row[0]..row[8]` indexing in one move. Worth adopting, but the current code is not fragile enough for this to block.

**3. The expected row is a third copy of the same data.** The new full-row literal duplicates `ROW` in the migration module and, transitively via the existing `test_seed_row_matches_registry`, the registry in `src/xagent/web/builtin_mcp_registry.py`. To be clear, the new test is **not** redundant — it is the only test that exercises `migration.upgrade()` actually persisting `ROW` into the database, which the registry-equality test never does. But it could derive its expected tuple from `migration.ROW` instead of retyping the literals, cutting three copies to two while keeping identical coverage.

**4. Pattern divergence across sibling connectors (informational, out of scope).** Only granola and notion get the sentinel-row downgrade fix. Seven sibling seed-migration test files — aws, slack, google-sheets, zoom, google-maps, google-ads, google-analytics — still carry the identical weak-downgrade assertion this PR is fixing. That is correctly out of scope given the PR and #1189 are explicitly scoped to granola + notion, but since the PR's own stated motivation is to stop connector tests drifting further apart, a tracked follow-up to apply the same pattern to the remaining seven would be worthwhile.

## Line-level findings

All findings below are **minor** or nit-level. None block merge.

- **minor** — `tests/alembic/test_20260731_seed_granola_mcp_app.py:162` and `tests/alembic/test_20260807_seed_notion_mcp_app.py:163` — the sentinel `INSERT` omits `transport`, which is `NOT NULL` with no server default in the real schema. See inline.
- **nit** — both files, the `is_visible_in_connector` comment at granola `:82-84` / notion `:83-85` — the comment's stated premise is inaccurate. See inline.
- **nit** — both files, granola `:73` / notion `:74` — the `isinstance(oauth_scopes, str)` guard is dead code today, since `ROW["oauth_scopes"]` is `None` for both connectors. Purely defensive; fine to keep, just noting it.
- **nit** — both files, granola `:75-84` / notion `:76-85` — two adjacent multi-line comments explain what is really a single `assert`, and the first ("Exact comparison…") sits right after the `oauth_scopes` cast rather than next to the assertion it describes. Cosmetic.
- **nit** — both files, the sentinel `connection.execute(...)` sits inside the `with patch.object(migration, "op", ...)` block even though it never touches `migration.op`. Moving it outside that block would scope the patch to just the migration calls. Readability only.
- **nit** — both files, granola `:168` / notion `:169` — `assert "sentinel" in app_ids` could be `assert app_ids == {"sentinel"}` for a tighter post-downgrade assertion. Low value in practice: `downgrade()` uses an exact-equality `WHERE app_id == APP_ID` delete, so there is no realistic overly-broad-delete mutation that the current assertion would miss. Optional.
- **nit** — `tests/web/api/test_public_mcp_connector_visibility.py:774` — naming consistency with sibling tests. See inline.
- **nit** — `tests/web/api/test_public_mcp_connector_visibility.py:778` — docstring wraps an identifier mid-token. See inline.

The test split for gap 2 is clean: no coverage was lost when the aws assertion block shifted position within its test, and the new test does not overlap `test_builtin_registry_classifies_remote_mcp_apps_as_mcp_oauth`, which exercises `classify_app_auth` rather than transport/launch_config.

## Test results

`pytest tests/alembic/test_20260807_seed_notion_mcp_app.py tests/alembic/test_20260731_seed_granola_mcp_app.py tests/web/api/test_public_mcp_connector_visibility.py` — **56 passed, 0 failed.**

## Disposition

**Approve.** The three gaps from #1189 are genuinely closed, the tests pass, and everything above is optional polish. Suggest opening a follow-up issue for design notes 1 and 4 (test schema fidelity, and extending the sentinel pattern to the remaining seven connector test files).

PR #1192, opened 2026-08-07
fix(ssh): skip the agent binding fetch when the user is not in a team

bluefish-08:
## Problem

Opening any agent in the builder pops a red toast: "Could not load SSH bindings. Reload before saving to avoid dropping the SSH tools." — even though the OSS build ships no SSH feature at all.

`AgentSshBindings` gates itself with `if (!inTeam) return null`, but that guard lives in the render path, and React runs hooks before it. The `useEffect` that calls `load()` therefore fires on every mount, hits `GET /api/agents/{id}/ssh-targets` (a route that only exists in the hosted build), gets a 404, and toasts the failure. The component itself never renders — only the error does.

## Fix

Apply the same gate at the fetch site:

```ts
if (!agentId || !inTeam) return
```

The targets-dropdown effect needs no change: it only runs when the dialog opens, and the dialog is unreachable when `!inTeam`.

## Tests

New `agent-ssh-bindings.test.tsx`:
- not in a team → no request, no toast
- in a team → bindings load and `onCount` fires

## Verification

- `npx vitest run src/components/build/agent-ssh-bindings.test.tsx` — 2 passed
- pre-commit: npm lint, npm type-check passed

### DIFF
--- a/frontend/src/components/build/agent-ssh-bindings.test.tsx
+++ b/frontend/src/components/build/agent-ssh-bindings.test.tsx
@@ -38,10 +38,12 @@ afterEach(() => {
 })
 
 describe("AgentSshBindings", () => {
-  it("skips the bindings fetch when the user is not in a team", async () => {
+  it("skips the bindings fetch when the user is not in a team", () => {
     inTeamMock.value = false
     render(<AgentSshBindings agentId="agent-1" />)
-    await waitFor(() => expect(apiRequestMock).not.toHaveBeenCalled())
+    // render flushes the effect synchronously and the guard returns before any
+    // await, so the absence of a call is observable without waitFor.
+    expect(apiRequestMock).not.toHaveBeenCalled()
     expect(toastErrorMock).not.toHaveBeenCalled()
   })
 
--- a/frontend/src/components/build/agent-ssh-bindings.test.tsx
+++ b/frontend/src/components/build/agent-ssh-bindings.test.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { cleanup, render, waitFor } from "@testing-library/react"
+import { act, cleanup, render, waitFor } from "@testing-library/react"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 
 // The OSS build ships no SSH routes, so the bindings fetch 404s and used to
@@ -21,9 +21,13 @@ vi.mock("@/components/ui/sonner", () => ({
 vi.mock("@/contexts/auth-context", () => ({
   useAuth: () => ({ inTeam: inTeamMock.value }),
 }))
-vi.mock("@/contexts/i18n-context", () => ({
-  useI18n: () => ({ t: (key: string) => key }),
-}))
+// The i18n return value must be referentially stable: `load` is keyed on `t`,
+// so a per-render `t` identity turns the mount effect into an unbounded
+// fetch/render loop (same hazard documented in agent-builder-publish-errors).
+vi.mock("@/contexts/i18n-context", () => {
+  const i18n = { locale: "en", t: (key: string) => key }
+  return { useI18n: () => i18n }
+})
 
 import { AgentSshBindings } from "./agent-ssh-bindings"
 
@@ -37,14 +41,27 @@ afterEach(() => {
   vi.clearAllMocks()
 })
 
+/** Let the load() promise chain settle so its catch branch can run. */
+async function flushLoad() {
+  await act(async () => {
+    await Promise.resolve()
+    await Promise.resolve()
+  })
+}
+
 describe("AgentSshBindings", () => {
-  it("skips the bindings fetch when the user is not in a team", () => {
+  it("skips the bindings fetch when the user is not in a team", async () => {
     inTeamMock.value = false
-    render(<AgentSshBindings agentId="agent-1" />)
-    // render flushes the effect synchronously and the guard returns before any
-    // await, so the absence of a call is observable without waitFor.
+    const { container } = render(<AgentSshBindings agentId="agent-1" />)
+
+    // toast.error only fires from load()'s catch, two awaits in, so the
+    // absence of a toast is only meaningful once those microtasks have run.
+    await flushLoad()
     expect(apiRequestMock).not.toHaveBeenCalled()
     expect(toastErrorMock).not.toHaveBeenCalled()
+    // The render-level guard still renders nothing — the premise for leaving
+    // the targets-dropdown effect keyed on dialogOpen alone.
+    expect(container).toBeEmptyDOMElement()
   })
 
   it("loads bindings when the user is in a team", async () => {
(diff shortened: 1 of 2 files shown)


frontend/src/components/build/agent-ssh-bindings.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+afterEach(() => {
+  cleanup()
+  vi.clearAllMocks()
+})
+
+describe("AgentSshBindings", () => {
+  it("skips the bindings fetch when the user is not in a team", async () => {
+    inTeamMock.value = false
+    render(<AgentSshBindings agentId="agent-1" />)
+    await waitFor(() => expect(apiRequestMock).not.toHaveBeenCalled())
```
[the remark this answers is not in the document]
bluefish-08: Good catch — dropped the `waitFor`. The guard returns before any await and `render` flushes the effect synchronously, so the assertion holds directly. Verified it is not a false positive: reverting the guard makes this test fail. Fixed in ac5bb8a1.

frontend/src/components/build/agent-ssh-bindings.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+const toastErrorMock = vi.hoisted(() => vi.fn())
+const inTeamMock = vi.hoisted(() => ({ value: false }))
+
+vi.mock("@/lib/api-wrapper", () => ({ apiRequest: apiRequestMock }))
+vi.mock("@/lib/utils", async () => {
+  const actual = await vi.importActual<typeof import("@/lib/utils")>("@/lib/utils")
+  return { ...actual, getApiUrl: () => "http://api.local" }
+})
+vi.mock("@/components/ui/sonner", () => ({
+  toast: { error: toastErrorMock, success: vi.fn() },
```
rogercloud, line 19: MODERATE (test-suite health, not a runtime bug): this mock returns a new `t` identity on every render. `load` is keyed on `t` in its `useCallback` deps, and the mount effect re-runs on every `load` identity change — so the in-team test spins in an unbounded fetch/render loop (~4885 `apiRequest` calls observed in a 300ms window). It passes only because `waitFor` inspects the first `onCount(0)` call. Please hoist a single stable i18n mock object and return it from `useI18n`, as `agent-builder-publish-errors.test.tsx` does (it documents this same hazard).

frontend/src/components/build/agent-ssh-bindings.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+})
+
+describe("AgentSshBindings", () => {
+  it("skips the bindings fetch when the user is not in a team", () => {
+    inTeamMock.value = false
+    render(<AgentSshBindings agentId="agent-1" />)
+    // render flushes the effect synchronously and the guard returns before any
+    // await, so the absence of a call is observable without waitFor.
+    expect(apiRequestMock).not.toHaveBeenCalled()
+    expect(toastErrorMock).not.toHaveBeenCalled()
```
rogercloud, line 61: MINOR: this assertion runs synchronously right after `render()`, but `toast.error(...)` is only reachable in the `catch` after two `await`s — so it can never observe a failure. Verified: with line 46 removed and the fix reverted, this test still passes. Wrap it in `waitFor` (or flush microtasks first) so it actually guards the behavior; line 46 is currently the only assertion pinning the regression.

frontend/src/components/build/agent-ssh-bindings.test.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react"
```
rogercloud, line 1: MINOR (coverage gap): neither test covers the `inTeam` false→true transition after mount — which is exactly the behavior that adding `inTeam` to `load`'s deps enables, and the case that would catch someone later removing it from the array. A second gap: nothing asserts the render-level guard still renders nothing when `!inTeam`, which is the premise for leaving the targets-dropdown effect unchanged. Neither is covered elsewhere in the repo.

rogercloud (approved):
## Summary

This PR fixes a hook-ordering bug in `AgentSshBindings`: the component gated itself with a render-level `if (!inTeam) return null`, but React runs the mount `useEffect` before that early return, so `load()` fired on every builder open. In the OSS build (no SSH routes) that meant a `GET /api/agents/{id}/ssh-targets` request per builder open, a 404, and a spurious error toast from a component that never actually renders. The fix moves the same `!inTeam` check into `load()` itself and adds `inTeam` to `load`'s dependency array, and adds a small test file with 2 cases.

## Approach: SOUND

- The guard is placed at the correct layer — the fetch site, not just the render site — which is the minimal change that actually stops the request.
- It matches an established in-repo convention: `frontend/src/components/mcp/connect-mcp-dialog.tsx` (~line 195) and `frontend/src/app/tools/page.tsx` (~line 293) already gate hosted-only fetches with the same `if (!inTeam) return` inside the async function.
- Keeping both guards is correct rather than duplicated: the render guard answers "should I paint?", the fetch guard answers "should I fetch?". Same shape as the sibling components.
- Adding `inTeam` to `load`'s deps is load-bearing, not incidental: `inTeam` starts `false` and resolves asynchronously, so without it hosted users would permanently skip the fetch once `inTeam` flips true. Verified that the fetch does fire on the false→true transition.
- The PR's claim that the targets-dropdown effect needs no change checks out: the dialog-open button only exists in the subtree returned after the render-level `!inTeam` guard, so the dialog cannot open when `!inTeam`.
- No consumer regresses: `onCount` feeds `hasSshBindings` in `frontend/src/components/build/agent-builder.tsx`, which is independently seeded from the saved `tool_categories`, so skipping the fetch for non-team users cannot silently drop an "ssh" tool category.
- Reverting the one-line guard makes the new test fail with exactly the expected assertion, so the test does pin the regression.

## Findings (non-blocking, test file only)

No issues were found in the production code change. Three test-quality items were found in the new test file — see inline comments. None block merging.

Lean already. The diff was scanned for dead flexibility, reinvented utilities, unnecessary dependencies, speculative abstractions, and over-elaborate tests — none found. The production change is a genuine one-liner plus an explanatory comment, and the 60-line/2-case test file is in line with the directory's norms and existing mocking conventions.

## Testing / verification

The new tests were run locally by the reviewer and the regression was reproduced end-to-end: reverting the guard fails the new test with the expected assertion, and the runaway re-render loop noted inline was measured rather than inferred. The author's stated verification (2 tests passing, lint and type-check clean) stands.

frontend/src/components/build/agent-ssh-bindings.test.tsx
bluefish-08: You are right, and the hazard was already documented one directory over — the comment above the i18n mock in `agent-builder-publish-errors.test.tsx` names this component specifically. I should have looked before writing the mock. Hoisted a single stable object with the same shape, and pinned the call count with `toHaveBeenCalledTimes(1)` so a future regression to a per-render identity fails loudly instead of passing on the first `onCount`.

frontend/src/components/build/agent-ssh-bindings.test.tsx
bluefish-08: Confirmed — added a `flushLoad()` helper that awaits two microtask turns inside `act`, then assert. Verified it now bites: reverting the guard fails on the toast assertion itself (`expected spy to not be called, actually called 1 times`), which it could not do before.

frontend/src/components/build/agent-ssh-bindings.test.tsx
bluefish-08: Both gaps covered. Added a rerender test for the false->true transition — verified it fails when `inTeam` is dropped from the deps array, which is exactly the regression you describe. The first test now also asserts `toBeEmptyDOMElement()`, pinning the render-level guard that justifies leaving the targets-dropdown effect keyed on `dialogOpen` alone.

PR #1193, opened 2026-08-07
fix(kb): require an explicit knowledge base name on creation

### LINKED ISSUE #1138: bug: Handle empty Knowledge Base name validation properly
### Description

When creating a Knowledge Base without providing a name, the system returns a backend validation error instead of showing a clear user-facing validation message.

### Problem Statement

The Knowledge Base name field should be required during creation.

Currently, when the user leaves the Name field empty and submits the form, the request reaches the backend and returns the following error:

> Invalid collection name: Invalid collection name: contains invalid characters. Only letters, numbers, spaces, underscores, and hyphens are allowed.

This error is misleading because the actual issue is that the name is missing, not that invalid characters were entered.

### Steps to Reproduce

1. Go to Knowledge Base creation page.
2. Leave the Name field empty.
3. Click create/submit.
4. Observe the error message.

bluefish-08:
Fixes #1138

## Problem

Creating a Knowledge Base with an empty name returns a misleading error:

> Invalid collection name: Invalid collection name: contains invalid characters. Only letters, numbers, spaces, underscores, and hyphens are allowed.

The name was never actually empty by the time it reached the backend. The creation dialog silently substituted a fallback:

- `knowledge-base-creation-dialog.tsx:373` — file upload used the file name minus its extension
- `:513` — web import used `"web_collection"`
- `:641` — cloud import used the first file name, then `"cloud_collection"`

A derived file name routinely contains dots, parentheses, or other characters that `ALLOWED_NAME_PATTERN` rejects, so the user got a character-validation error for a name they never typed. The backend's own empty-name path (`web/config.py:167`) already returns the correct "cannot be empty" message — it was simply unreachable.

## Fix

Drop all three fallbacks and require the name up front. One guard covers every import tab, at the single call site that dispatches to all three ingest handlers:

```ts
if (!trimmedCollectionName) {
  toast.error(t("kb.errors.nameRequired"))
  setCurrentStep(1)
  return
}
```

The field is now marked required (`*` in the label, `required` / `aria-required` on the input) and the placeholder no longer advertises the removed fallback behaviour.

The per-tab "multiple files need an explicit name" rule is removed as redundant — a globally required name already implies it.

No backend change: with the name required client-side, no empty value reaches `sanitize_path_component`, and adding a second guard there would be duplicate validation.

## Behaviour change

The Knowledge Base name is now **required**, where it was previously optional with a derived default. This is the intent of #1138, but it does change an existing workflow: users who relied on leaving the field blank must now supply a name.

## Tests

`knowledge-base-creation-dialog.test.tsx`:
- empty name blocks submission for both single- and multi-file uploads (`it.each([1, 2])`), asserting the toast fires and `/api/kb/ingest/jobs` is never called
- whitespace-only name is treated as empty
- three existing tests that relied on the removed fallback now supply an explicit name

```
npx vitest run src/components/kb/knowledge-base-creation-dialog.test.tsx
 ✓ 7 tests passed
```

This also fixes a pre-existing failing assertion in that file: the sonner `toast.error` proxy appends `{ duration: 8000 }`, which the old `toHaveBeenCalledWith` did not account for.

## Verification

- `npx tsc --noEmit` — clean
- pre-commit: npm lint, npm type-check, codespell passed


frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -786,19 +767,16 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
             {currentStep === 1 && (
               <div className="space-y-6 mt-4">
                 <div>
-                  <Label htmlFor="collection_name" className="text-sm font-medium">{t("kb.dialog.basicInfo.nameLabel")} {t("common.optional")}</Label>
+                  <Label htmlFor="collection_name" className="text-sm font-medium">{t("kb.dialog.basicInfo.nameLabel")}</Label>
```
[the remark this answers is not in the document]
bluefish-08: Agreed — the asterisk is UI semantics, not copy, and putting it in the locale meant it could not carry the `text-destructive` styling that the Start URL label on line 967 of this same file already uses. Moved into the component and removed from both locales. Fixed in 6b660597.

frontend/src/i18n/locales/en.ts
```diff
@@ -1997,9 +1996,8 @@ Build when you need.`,
       createDescription: "Create a new knowledge base and upload documents or import from website",
       basicInfo: {
         title: "Basic Information",
-        nameLabel: "Knowledge Base Name",
-        namePlaceholder: "Leave blank to use filename or URL as the name",
-        multiFileRequiredHint: "Multiple files must share one explicit knowledge base name.",
+        nameLabel: "Knowledge Base Name *",
```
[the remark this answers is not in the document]
bluefish-08: Agreed — the asterisk is UI semantics, not copy, and putting it in the locale meant it could not carry the `text-destructive` styling that the Start URL label on line 967 of this same file already uses. Moved into the component and removed from both locales. Fixed in 6b660597.

frontend/src/i18n/locales/zh.ts
```diff
@@ -1997,9 +1996,8 @@ const zh = {
       createDescription: "创建新的知识库并上传文档或从网站导入内容",
       basicInfo: {
         title: "基本信息",
-        nameLabel: "知识库名称",
-        namePlaceholder: "留空将使用文件名或 URL 作为知识库名称",
-        multiFileRequiredHint: "上传多个文件时，必须填写同一个明确的知识库名称。",
+        nameLabel: "知识库名称 *",
```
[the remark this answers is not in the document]
bluefish-08: Agreed — the asterisk is UI semantics, not copy, and putting it in the locale meant it could not carry the `text-destructive` styling that the Start URL label on line 967 of this same file already uses. Moved into the component and removed from both locales. Fixed in 6b660597.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -207,7 +207,7 @@ describe("KnowledgeBaseCreationDialog multi-file naming", () => {
     cleanup()
   })
 
-  it("requires an explicit collection name for multiple file uploads", async () => {
+  it.each([1, 2])("rejects an empty collection name for %i file upload(s)", async (fileCount) => {
```
rogercloud: **Major (test coverage) — the new empty-name tests only exercise the file-upload tab.**

The PR states that "one guard covers every import tab," but both new rejection tests submit via the file tab. No test empties the name and attempts to submit through the web or cloud tab; the pre-existing web/cloud tests were instead updated to supply an explicit name.

Structurally this is low risk today — there is a single shared `if` before the tab dispatch — but a future scoped regression (e.g. the guard narrowed to `activeImportTab === "file"`) would slip through completely undetected for web and cloud, which is precisely the property the PR claims.

Suggested fix: parameterize these two cases over all three tabs (`file` / `web` / `cloud`) so the shared-guard claim is actually asserted.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -217,24 +217,51 @@ describe("KnowledgeBaseCreationDialog multi-file naming", () => {
...
+        ),
       },
     })
 
     fireEvent.click(screen.getByText("common.next"))
     fireEvent.click(screen.getByText("kb.dialog.createButton"))
 
     await waitFor(() => {
-      expect(toastErrorMock).toHaveBeenCalledWith("kb.errors.multiFileNameRequired")
+      expect(toastErrorMock).toHaveBeenCalledWith("kb.errors.nameRequired", expect.anything())
```
rogercloud: **Minor (test coverage) — the tests don't assert the user is navigated back to the name field.**

These cases assert that the error toast fires and that no API call is made, but not that `setCurrentStep(1)` actually runs. Returning the user to Step 1 is the remediation the guard promises — without it the user would see an error with no obvious way to act on it — so it deserves an explicit assertion rather than being left implicit.

Suggested fix: assert the dialog is back on Step 1 (e.g. the name input is visible / focused) after the rejected submit.

rogercloud: **Minor — `expect.anything()` here is papering over an unmocked wrapper.**

`expect(toastErrorMock).toHaveBeenCalledWith("kb.errors.nameRequired", expect.anything())` — the second argument matcher verifies nothing.

The cause: the component imports `toast` from the `@/components/ui/sonner` wrapper, which injects `{ duration: 8000 }` into every `.error()` call via a Proxy, while this test mocks only the raw `sonner` package. `expect.anything()` was needed to make the assertion pass, but it silently accepts any options object.

Suggested fix: mock `@/components/ui/sonner` instead of `sonner` and assert the concrete options (`{ duration: 8000 }`), or drop the second argument from the assertion entirely rather than asserting something meaningless. Same applies at line 257.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -786,19 +767,16 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
             {currentStep === 1 && (
               <div className="space-y-6 mt-4">
                 <div>
-                  <Label htmlFor="collection_name" className="text-sm font-medium">{t("kb.dialog.basicInfo.nameLabel")} {t("common.optional")}</Label>
+                  <Label htmlFor="collection_name" className="text-sm font-medium">{t("kb.dialog.basicInfo.nameLabel")} <span className="text-destructive">*</span></Label>
```
rogercloud, line 785: **Minor — `required` is decorative in this dialog.**

There is no `<form>` element anywhere in this file, and the Create/Next controls are plain `onClick` buttons rather than `type="submit"`, so native HTML constraint validation will never fire for this input. The actual enforcement is entirely the JS guard in the Create button's `onClick`.

`aria-required="true"` still carries real accessibility value and should stay. Pairing it with `required` is spec-redundant but harmless — the concern is that it reads as functional validation when it isn't. Consider dropping `required` (or adding a short comment) so a future reader doesn't assume the browser is enforcing anything here.

rogercloud, line 785: **Minor (nit) — two conventions for marking requiredness now coexist in this form.**

The name field marks itself required with a raw `<span className="text-destructive">*</span>`, while the description field a few lines below (~786) marks itself optional with `{t("common.optional")}`. Both conventions in the same form is a little inconsistent — worth settling on one (either `*` on required fields or an "(optional)" suffix on optional ones).

To be clear, moving the `*` out of the locale string and into JSX was the right call and correctly addresses the earlier Gemini Code Assist comment — verified fixed in commit `6b660597`, with `kb.dialog.basicInfo.nameLabel` in both `frontend/src/i18n/locales/en.ts:1999` and `frontend/src/i18n/locales/zh.ts:1999` no longer containing `*`.

rogercloud (changes requested):
## Summary

This PR removes the three silent name-derivation fallbacks in the KB creation dialog (file stem, `web_collection`, `cloud_collection`) and replaces them with a single guard placed in the final "Create" button handler, before dispatch to any of the file/web/cloud ingest paths. The name input is now marked visually required (`*` + `aria-required`), and locale strings and tests were updated accordingly. The motivation (issue #1138) is sound: an empty name previously became a derived name that then failed backend character validation, producing a misleading "contains invalid characters" error instead of "name required."

The direction is right and the dialog-level UX improvement is real. My reservations are about **completeness** — the root cause is not actually addressed where the PR description claims it is, and two concrete UI gaps let a user still reach a screen showing a fabricated name.

---

## Approach review: acceptable with reservations

### The backend fallback still exists — the root cause is not fixed (major)

The PR description states: *"No backend change: with the name required client-side, no empty value reaches `sanitize_path_component`... the backend's own empty-name path was simply unreachable."* This is not accurate.

Both `src/xagent/web/api/kb.py:3825-3827` (`/ingest`) and `src/xagent/web/api/kb.py:4182-4183` (`/ingest/jobs`) contain their own pre-existing fallback:

```python
if not collection or not collection.strip():
    collection = Path(safe_filename).stem
```

This runs **before** `sanitize_path_component` is ever called. It is pre-existing code and untouched by this diff, but that is exactly the point: any non-dialog client (curl, the SDK, a future or alternate frontend) that sends an empty `collection` still gets silently defaulted to a filename stem rather than receiving the "cannot be empty" validation error. Issue #1138's root cause is hidden for users of this one dialog, not fixed at the API layer.

Suggested direction: remove the two backend fallbacks and let `sanitize_path_component`'s empty-name validation surface, or — if that is deliberately out of scope for this PR — file a follow-up and correct the PR description so it does not claim the backend path was already unreachable.

### The new guard lives outside the handlers it protects (major, consistency)

`handleUpload` (`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:344-345`) and `handleWebIngest` (`:490-491`) each keep their own internal precondition guard (file count, `start_url`) at the top of the function. The new empty-name guard was not added to any of the three handlers — it exists only in the Create button's `onClick` (`:1268-1282`), and all three handlers now consume `trimmedCollectionName` unconditionally.

Each handler has exactly one call site today (that same `onClick`), so this is not currently exploitable. But it is inconsistent with the file's own established pattern, and it becomes a real bug the moment any handler gains a second call site. Either move/duplicate the name check into the handlers alongside their existing guards, or move the file/URL guards out to the button for symmetry.

### Blast radius is wider than the description suggests (informational)

`KnowledgeBaseCreationDialog` is rendered by both `frontend/src/components/pages/knowledge-base.tsx:606` and `frontend/src/components/build/agent-builder.tsx:2988` — the same component, no forked logic. So the "name is now strictly required" behavior change also applies to the Agent Builder's inline create-KB flow, not just the main KB page. Probably fine, but worth calling out explicitly so it reads as a deliberate tradeoff.

---

## Line-level findings

### Major

**1. `frontend/src/components/kb/knowledge-base-creation-dialog.tsx:1256-1259` — Step 1 "Next" is not gated on the name.**
The `disabled` condition only checks Step 2 preconditions (file count / `start_url`); it never inspects `trimmedCollectionName`. A user can walk Step 1 → Step 2 → Step 3 with a completely empty name and is only bounced back on clicking "Create". This contradicts the component's own pattern of disabling "Next" until the current step's precondition is satisfied. Add `!trimmedCollectionName` to the Step 1 branch of the disabled condition.

**2. `frontend/src/components/kb/knowledge-base-creation-dialog.tsx:1010` — a 4th derived-name fallback survives in the Step 3 review card.**
`{newCollectionName || "KB " + new Date().toLocaleString()}` is a leftover fallback that this diff did not clean up. Combined with finding 1, a user who reaches Step 3 with an empty name is shown a fabricated name such as `KB 8/8/2026, 10:31:05 PM` presented as the collection about to be created — and then rejected on Create. This directly contradicts the PR's stated goal of never surfacing a silently derived name. Remove the fallback (fixing finding 1 makes the empty branch unreachable, but the dead fallback should go regardless).

**3. `frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx:210-262` — new tests cover only the file tab.**
The PR claims "one guard covers every import tab," but both new empty-name rejection tests exercise the file-upload tab only; no test empties the name and submits via the web or cloud tab. The pre-existing web/cloud tests were instead updated to supply an explicit name. Structurally this is low risk today (one shared `if` before the dispatch), but a future scoped regression — e.g. the guard accidentally narrowed to `activeImportTab === "file"` — would go entirely undetected for web and cloud. Add two parameterized cases over the three tabs.

### Minor

**4. `frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx:210-262` — no assertion on the step reset.**
The tests assert the toast fires and no API call is made, but not that `setCurrentStep(1)` runs / that the user is returned to the name field. Since navigating back is the actual remediation the guard promises, it should be asserted.

**5. `frontend/src/components/kb/knowledge-base-creation-dialog.tsx:770-779` — `required` is decorative here.**
There is no `<form>` element anywhere in this file, and Create/Next are plain `onClick` buttons rather than `type="submit"`, so native constraint validation never fires. `aria-required="true"` still carries real accessibility value; keeping `required` alongside it is spec-redundant but harmless. Worth a comment or dropping `required` so it does not read as functional validation.

**6. `frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx:231,257` — `expect.anything()` masks an unmocked wrapper.**
The assertions use `toHaveBeenCalledWith("kb.errors.nameRequired", expect.anything())`. The cause is that the component imports `toast` from the `@/components/ui/sonner` wrapper, which injects `{ duration: 8000 }` into every `.error()` call via a Proxy, while the test mocks only the raw `sonner` package. `expect.anything()` makes the assertion pass but verifies nothing about that argument. Mock the wrapper instead and assert the real options object.

**7. `frontend/src/components/kb/knowledge-base-creation-dialog.tsx:770` vs `:786` — two conventions for requiredness in one form.**
The name field now renders a raw `<span className="text-destructive">*</span>`, while the description field uses `{t("common.optional")}`. Pick one convention for the form. (To be clear: moving the `*` out of the locale string into JSX was the right call — see the note on the prior review below.)

> **Aside, not blocking:** `src/xagent/web/api/kb.py` wraps errors as `f"Invalid collection name: {str(e)}"` at 13 call sites (e.g. line 3602), double-prefixing an exception message that already begins with `Invalid collection name: ` — this produces the exact doubled string quoted in issue #1138's report. It predates this PR's base commit and is untouched here; worth a separate follow-up issue rather than scope creep in this PR.

---

Lean already. One candidate was investigated — the `const collectionName = trimmedCollectionName` alias in `handleWebIngest` around line 505, which looked like a redundant single-use alias — but it was refuted on inspection: `collectionName` is referenced four times in that function (the `formData.append`, two error-result builders, and the `onSuccess` callback), matching the pattern used in the other two handlers. Nothing to report.

## Prior review feedback

The earlier Gemini Code Assist comment (the required-indicator `*` should live in the component, not be hardcoded in the locale string) was addressed in commit `6b660597` and is verified fixed: `frontend/src/components/kb/knowledge-base-creation-dialog.tsx:770` now renders the `*` in JSX, and the `kb.dialog.basicInfo.nameLabel` strings in `frontend/src/i18n/locales/en.ts:1999` and `frontend/src/i18n/locales/zh.ts:1999` no longer contain it.

## Testing performed

- `npx vitest run src/components/kb/knowledge-base-creation-dialog.test.tsx` → 7/7 passed.
- `npx tsc --noEmit` → clean, no errors.

## Recommendation

**Request changes.** To be fair to the change: the frontend UX fix for this dialog is real, it works, and removing the three silent fallbacks is the right call. The blockers are about completeness rather than the approach:

- The backend's own empty-name fallback in `src/xagent/web/api/kb.py` means issue #1138's root cause is still live for any non-dialog client, contrary to the PR description's claim.
- Findings 1 and 2 together still let a user reach a Step 3 confirmation screen displaying a fabricated `KB <timestamp>` name — the exact behavior this PR sets out to eliminate.

Fixing the Step 1 gate and deleting the line-1010 fallback are both small, and either removing the backend fallback or filing a follow-up plus correcting the PR description would resolve the first point. Happy to re-review promptly.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Fair nit, though the `*` convention predates this PR — the Start URL label at line 967 of this same file already uses `<span className="text-destructive">*</span>`, and the description field's `(optional)` suffix is equally pre-existing. Both conventions already coexisted here; this change only moves the name field into the one that was already there rather than introducing a second. Settling on one across the form is worth doing, but it touches labels this PR does not otherwise go near, so I would rather leave it to a dedicated pass than widen an empty-name-validation fix. Happy to open a follow-up issue if you would like it tracked.

Thanks for confirming the locale fix.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Agreed — the claim was in the PR text but not in the tests. Both rejection cases are now `it.each` over file/web/cloud, walking each tab to step 3 through the same fixtures the existing web/cloud tests already use.

One thing that surfaced while doing it: the no-ingest assertion was pinned to `url === ".../api/kb/ingest/jobs", so parameterizing alone would have made it vacuous for web and cloud — they post to `/ingest-web/jobs` and `/ingest-cloud`. Widened to `includes("/api/kb/ingest")`.

Verified against the exact regression you describe: narrowing the guard to `activeImportTab === "file"` fails the four web/cloud cases and leaves the file ones green. Fixed in 51af3045.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Added — the shared helper now asserts `#collection_name` is in the DOM after a rejected submit. Step 1 is the only step that renders the name field, so its presence is what pins `setCurrentStep(1)`. Verified: removing that line fails all seven guard cases. Fixed in 51af3045.

bluefish-08: You are right that it was papering over the wrong mock target. Switched to mocking `@/components/ui/sonner` — the wrapper the component actually imports — so the asserted arguments are the ones the component passes, and dropped the second argument entirely: `toHaveBeenCalledWith("kb.errors.nameRequired")`. No `expect.anything()` remains in the file.

Confirmed the old form asserted nothing: adding a bogus `{ duration: 8000 }` to the component call now fails six cases, and passed silently under `expect.anything()`. Fixed in 51af3045.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Dropped `required`, kept `aria-required`. Went with removal rather than a comment: there is no `:invalid` / `:required` / `peer-invalid` styling anywhere in the repo, so the attribute had no behaviour left beyond the constraint validation that can never fire here, and the one thing it would have contributed implicitly — the ARIA mapping — is already spelled out explicitly on the same line. Keeping it with a comment would mean maintaining a line of prose to explain a line of dead markup. Fixed in 51af3045.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -786,19 +767,15 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
...
                   <Input
                     id="collection_name"
                     value={newCollectionName}
                     onChange={(e) => setNewCollectionName(e.target.value)}
                     placeholder={t("kb.dialog.basicInfo.namePlaceholder")}
```
rogercloud, line 793: The name input carries `aria-required="true"` but there is no `aria-invalid` and no inline error text. This PR removed the previous inline destructive hint (`multiFileRequiredHint`, gated on `requiresExplicitCollectionName`) without a general replacement, so the only feedback on a rejected submit is a transient toast that never appears next to the field. A screen-reader user who lands back on step 1 gets no field-level indication of what went wrong.

Suggest a `nameError` state set by the new guard, rendered as `aria-invalid={!!nameError}` on the input plus an inline `<p>` message beneath it.

frontend/src/i18n/locales/zh.ts
```diff
@@ -1890,7 +1890,6 @@ const zh = {
       nameRequired: "请输入知识库名称",
```
rogercloud: After this PR changed the placeholder, this error toast string (kb.errors.nameRequired) and the placeholder string (kb.dialog.basicInfo.namePlaceholder, zh.ts:2000) are now word-for-word identical. The previous placeholder was a distinct string, so this collision is new. The zh toast now just repeats the text the user is already looking at in the empty input, adding no information about why the submit failed. English does not have this problem (nameRequired and namePlaceholder are distinct strings in en.ts).

Suggest differentiating the zh nameRequired string using the existing "cannot be empty" phrasing pattern already used for sibling keys at zh.ts:1274 and zh.ts:2452, rather than reusing the placeholder wording.

rogercloud (commented):
## Summary

This PR fixes the root cause of #1138 on the client: the creation dialog used to silently invent a knowledge base name (from the uploaded filename, or the literals `web_collection` / `cloud_collection`) whenever the user left the field blank, so an empty name never reached backend validation and instead surfaced as a misleading "contains invalid characters" error derived from the fabricated value. The three fallbacks are removed and replaced by a single required-name guard at the one dispatch site that fans out to the file/web/cloud ingest handlers, the field is marked required in the UI, the now-redundant per-tab "multiple files need an explicit name" rule is dropped, and tests are updated accordingly.

## Prior review status

All five findings from the previous `CHANGES_REQUESTED` review have been addressed and verified: the required-`*` marker moved from locale strings into JSX (6b66059), and empty-name coverage now runs via `it.each` across all three tabs, asserts navigation back to step 1, mocks the real `@/components/ui/sonner` wrapper with an exact assertion instead of `expect.anything()`, and drops the inert `required` attribute while keeping `aria-required` (51af304). The remaining nit — two coexisting conventions for marking required vs. optional fields in the same form — was explicitly waived by the author as out of scope, which is reasonable.

The centralization is the right call and is correctly scoped: the guard sits above the only caller of each of the three ingest handlers, so no tab can bypass it, and one guard is clearly better than three per-tab checks. No objection to the shape of the fix.

The reservation is with the PR's stated justification for *not* touching the backend — that "with the name required client-side, no empty value reaches `sanitize_path_component`, and adding a second guard there would be duplicate validation." That is not true for the file-upload ingest endpoints. `src/xagent/web/api/kb.py` declares `collection: str = Form(None)` in both `ingest` (~L3825-3827) and `create_ingest_job` (~L4182-4183), and each falls back to `collection = Path(safe_filename).stem` **before** calling `sanitize_path_component`. A direct API caller (curl/Postman/SDK, not the now-fixed dialog) posting to `/api/kb/ingest` or `/api/kb/ingest/jobs` with an empty `collection` and a filename such as `report (v2).pdf` still gets the exact "Invalid collection name: contains invalid characters..." message that #1138 is about — the "cannot be empty" branch at `src/xagent/web/config.py:167-168` is unreachable on these two paths. This is pre-existing code that the PR does not touch, so it is not a regression, but the claim that the path is already safe is incorrect and the original bug remains reproducible below the UI.

Suggested resolution: change `Form(None)` → `Form(...)` on both endpoints and delete the stem-fallback blocks, so an empty `collection` reaches `sanitize_path_component` and yields the correct "cannot be empty" error. If that is deliberately deferred (it is a behavioral change for existing API clients that rely on the filename fallback), please say so explicitly as a known gap / follow-up issue rather than asserting the path is already covered.

## Findings

### MAJOR

**`src/xagent/web/api/kb.py:3825-3827` and `:4182-4183`** — `collection` is an optional form field with a filename-stem fallback applied before validation, so the empty-name case from #1138 is still reachable (and still produces the misleading "invalid characters" error) for direct API callers. *Fix:* make the field required (`Form(...)`) and remove the fallback, or document this as an accepted follow-up gap. Detailed above; no single line to anchor an inline comment to.

### MINOR

1. **`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:1009`** — the step-3 review screen still fabricates a name for display: `{newCollectionName || "KB " + new Date().toLocaleString()}`. A user with an empty name reaches the review step, sees a plausible auto-generated label, and only then gets bounced back to step 1. Display-only and cannot bypass the guard, but it is the last remaining "the dialog invents a name" behavior — exactly what this PR set out to remove. *Fix:* render `trimmedCollectionName`, or an explicit "not set" state.

2. **`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:770-780`** — the name input has `aria-required="true"` but no `aria-invalid` and no inline error text; the old inline destructive hint (`multiFileRequiredHint`) was removed with no general replacement, so the only feedback on a rejected submit is a transient toast that never appears near the field. *Fix:* add a `nameError` state set by the guard, rendered as `aria-invalid` plus an inline `<p>` under the input.

3. **`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:1255-1258`** — the Next button already disables for other missing prerequisites (no files on the file tab, no URL on the web tab) but was not extended to gate step 1→2 on the newly required name, so users can complete the whole wizard before discovering the requirement. *Fix:* add `(currentStep === 1 && !trimmedCollectionName)` to the same `disabled` expression.

4. **`frontend/src/i18n/locales/zh.ts:1890` / `:2000`** — this PR changed the placeholder, and the zh error toast (`kb.errors.nameRequired`) and placeholder (`kb.dialog.basicInfo.namePlaceholder`) are now word-for-word identical strings, so the toast only repeats the text already visible in the empty input. English does not have this collision (`nameRequired` and `namePlaceholder` are distinct in `en.ts`). *Fix:* differentiate the zh `nameRequired` string, following the existing "cannot be empty" phrasing pattern already used for sibling keys at `zh.ts:1274` and `zh.ts:2452`, instead of reusing the placeholder wording.

### Out of scope (note only, not blocking)

There is still no client-side character-pattern validation for the name (backend pattern `^[\w -]+$` at `src/xagent/web/config.py:98`), so an invalid-character name is only rejected with a raw 422 after the whole wizard completes. Arguably outside this PR, which targets the empty-name case specifically — but #1138's title is the broader "Handle empty Knowledge Base name validation properly", so worth deciding consciously.

`frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx`

- L371: yagni inline file-tab step navigation duplicated after this diff added the required name field. Replace with the `goToStep3(container, "file")` helper this PR just introduced at L205-231.
- L438: yagni inline cloud-tab step navigation. Replace with `goToStep3(container, "cloud")`.
- L519: yagni inline web-tab step navigation. Replace with `goToStep3(container, "web")`.

No assertion in these three tests depends on the specific inline values (e.g. the exact filename) that would differ from what the helper produces, so all three are drop-in.

net: -20 lines possible

## Testing performed

- `npx vitest run` on the dialog test file: 11/11 pass.
- Mutation check: deleting the new required-name guard causes 7/11 to fail, confirming the guard tests are load-bearing rather than vacuous.

## Recommendation

**Comment**, not Request changes. The confirmed frontend findings are minor UX/a11y/i18n polish, and the MAJOR item is an *incomplete* fix in pre-existing backend code rather than a regression this PR introduces — the client-side change is a genuine improvement on its own. Please make an explicit call on the `kb.py` gap, though: either tighten `Form(None)` → `Form(...)` here, or file a follow-up and drop the "no backend change needed" claim from the PR description.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Good catch — removing the inline hint took away the only field-level feedback, and a toast is not a substitute for a screen-reader user landing back on step 1.

Added a `nameError` state set by the guard and cleared on the next keystroke, `aria-invalid` on the input, and an inline message reusing the same i18n key as the toast so the two can never drift. Also added `aria-describedby` pointing at that message — `aria-invalid` alone announces that the field is wrong without saying why. Reused the styling of the `multiFileRequiredHint` element this PR removed, so it is visually the same affordance.

The assertions live in the shared `expectNameRejected` helper, so all seven rejection cases now cover field-level feedback, plus a new case for clearing on input. Verified each independently: dropping the `aria-invalid` binding fails 8 cases, dropping the inline element fails 8, and not clearing on input fails exactly the one new case. Fixed in 2f60836a.

frontend/src/i18n/locales/zh.ts
bluefish-08: Confirmed and fixed — the two zh strings were word-for-word identical, and the collision was created by this PR changing the placeholder. Now `知识库名称不能为空`, following the phrasing already used by the sibling keys you pointed at (`zh.ts:1274`, `zh.ts:2452`).

What made the collision possible is worth naming: both strings were imperative (`请输入…`). The placeholder tells you what to do; the error states that the current value is invalid. Making the error declarative separates them by function rather than by wording. English needed no change.

That same key is now also the inline field error added for the accessibility comment, so the toast and the inline message stay in sync by construction. Fixed in 2f60836a.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -173,9 +173,8 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
   // Embedding models state
   const [embeddingModels, setEmbeddingModels] = useState<Model[]>([])
   const trimmedCollectionName = newCollectionName.trim()
-  const requiresExplicitCollectionName =
-    (activeImportTab === "file" && selectedFiles.length > 1) ||
-    (activeImportTab === "cloud" && totalCloudFiles > 1)
+  // Flag only: the message itself stays in i18n so it follows a language switch.
+  const [nameError, setNameError] = useState(false)
```
rogercloud, line 176: **Medium (new): `nameError` is never reset, so the error survives close/reopen.**

`resetState()` (:318-344, called from three sites) does not clear this flag, and the `open` effect (:179-183) only fetches embedding models. The Dialog's `onOpenChange` at :743 is the raw parent-controlled prop, so ESC / overlay click / close button never reach `resetState()` at all — and even the Cancel path, which does call it, leaves `nameError` set.

Repro: submit with an empty name (error text + `aria-invalid="true"` appear) -> close the dialog by any means -> reopen -> the empty, untouched input still shows red error text and `aria-invalid="true"` before the user has typed anything.

Suggested fix: add `setNameError(false)` to `resetState()`, and preferably also to the `open`-becomes-true effect so the close paths that bypass `resetState()` are covered too.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -786,17 +769,23 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
...
                     id="collection_name"
                     value={newCollectionName}
-                    onChange={(e) => setNewCollectionName(e.target.value)}
+                    onChange={(e) => {
+                      setNewCollectionName(e.target.value)
+                      setNameError(false)
+                    }}
                     placeholder={t("kb.dialog.basicInfo.namePlaceholder")}
                     className="mt-1.5"
+                    aria-required="true"
```
rogercloud, line 795: **Low (test coverage).** Two gaps around this input's new accessibility attributes:

1. No test covers dialog close -> reopen clearing `nameError`. This is exactly the gap that let the stale-state bug (see comment on :177) ship — a regression test should land with the fix.
2. No test asserts `aria-required="true"` or the `*` required marker in the label, even though `expectNameRejected` already asserts `aria-invalid` and `aria-describedby`. Cheap to add in the same place.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -1287,6 +1276,14 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
               ) : (
                 <Button
                   onClick={() => {
+                    // Single guard for every import tab: the name is required, so
+                    // the backend never has to reject a silently derived one.
+                    if (!trimmedCollectionName) {
+                      toast.error(t("kb.errors.nameRequired"))
+                      setNameError(true)
+                      setCurrentStep(1)
```
rogercloud: **Recurring (3rd round) — this guard is the only place the name requirement is enforced; the step-1->2 "Next" button (around :1264-1275) never checks `trimmedCollectionName`, and the step-3 review screen (around :1021) still renders `{newCollectionName || "KB " + new Date().toLocaleString()}` when the name is empty.**

So a user can go 1 -> 2 -> 3, configure chunking, see a plausible-looking fabricated name like `KB 8/8/2026, 10:31:05 PM` on the review screen, and only get bounced back here by this guard when clicking Create. Not exploitable as a submitted value, but it's the exact "the dialog invents a name" behavior this PR set out to remove, still visible in the UI.

Both points were flagged in the two prior review rounds and remain unaddressed. Suggested fixes: add `(currentStep === 1 && !trimmedCollectionName)` to the Next button's `disabled` expression, and render `trimmedCollectionName` (or an explicit "not set yet" placeholder) on the review screen instead of the timestamp fallback.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -173,9 +173,8 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
   // Embedding models state
   const [embeddingModels, setEmbeddingModels] = useState<Model[]>([])
   const trimmedCollectionName = newCollectionName.trim()
```
rogercloud: **Info (nit).** The new `useState` for `nameError` (:177) is declared after this derived constant, which interrupts the file's otherwise consistent "all `useState` hooks first, then derived values" ordering — 15+ consecutive `useState` calls precede this point with nothing interleaved. Purely cosmetic; moving the `useState` above `trimmedCollectionName` restores the pattern.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -304,6 +387,10 @@ describe("KnowledgeBaseCreationDialog multi-file naming", () => {
       <KnowledgeBaseCreationDialog open={true} onOpenChange={vi.fn()} onSuccess={onSuccess} />
     )
 
+    fireEvent.change(container.querySelector("#collection_name") as HTMLInputElement, {
+      target: { value: "alpha" },
+    })
+
     fireEvent.click(screen.getByText("common.next"))
```
rogercloud: Simplification (yagni): this and the sync-ingest-endpoint test around here still hand-roll step-1->3 navigation instead of using the `goToStep3` helper this PR introduced. Replace with `goToStep3(container, "file")`. Same applies to the "uses the same explicit collection name" test (~L331/343), the cloud partial-failures test (~L461-470), and the web partial-failures test (~L542-547) — net: -18 lines possible.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -363,10 +450,14 @@ describe("KnowledgeBaseCreationDialog multi-file naming", () => {
...
-      render(
+      const { container } = render(
         <KnowledgeBaseCreationDialog open={true} onOpenChange={onOpenChange} onSuccess={onSuccess} />
       )
 
+      fireEvent.change(container.querySelector("#collection_name") as HTMLInputElement, {
+        target: { value: "cloud-docs" },
+      })
+
       fireEvent.click(screen.getByText("common.next"))
```
rogercloud: Simplification (yagni): inline step-1->3 navigation in the cloud partial-failures test duplicates the new `goToStep3` helper. Replace with `goToStep3(container, "cloud")`.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -444,6 +535,10 @@ describe("KnowledgeBaseCreationDialog multi-file naming", () => {
         <KnowledgeBaseCreationDialog open={true} onOpenChange={onOpenChange} onSuccess={onSuccess} />
       )
 
+      fireEvent.change(container.querySelector("#collection_name") as HTMLInputElement, {
+        target: { value: "web_collection" },
+      })
+
       fireEvent.click(screen.getByText("common.next"))
```
rogercloud: Simplification (yagni): inline step-1->3 navigation in the web partial-failures test duplicates the new `goToStep3` helper. Replace with `goToStep3(container, "web")`.

rogercloud (approved):
## Summary

This PR fixes #1138 by removing the three silent name fallbacks in the knowledge base creation dialog (derived filename, `web_collection`, `cloud_collection`) and replacing them with a single required-name guard at the one dispatch point that routes to all three ingest handlers. The name field is now marked required in the UI (`*`, `aria-required`), and the final commit (2f60836a) adds inline `aria-invalid` / error-text feedback backed by a new `nameError` state. Scope is frontend-only: the dialog component, its test file, and the `en.ts` / `zh.ts` string tables. Tests were extended with an `it.each` over all three tabs for empty/whitespace rejection, an error-clears-on-typing case, and a shared `goToStep3` helper.

## Design assessment — sound

Dropping the fallbacks and centralizing the guard at the single dispatch site is the simplest correct fix and matches what #1138 explicitly asked for. Guard coverage was verified as complete: each of the three ingest handlers has exactly one call site and all three are downstream of the guard, so there is no bypass path from the dialog.

Two design-level concerns remain, and both are recurring — raised in the first review (CHANGES_REQUESTED) and again in the second (COMMENTED), and both are confirmed still present at head:

1. **Guard timing.** The requirement is only enforced at the final "Create" click. A user can walk 1 → 2 → 3, configure chunking, and only learn the name was mandatory three steps after the omission.
2. **Leftover fabricated display name.** The step-3 review screen still invents a name when the field is empty. This is precisely the "the dialog makes up a name" behavior the PR set out to eliminate — the fabrication was removed from the submit path but not from the confirmation UI.

Neither is a wrong-direction problem, but this is the third round in which both are being raised.

## Previously flagged, now fixed

- Inline field-level feedback near the name input (`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:177`, rendered at `:782-790`) — added and covered by tests.
- `zh.ts` `nameRequired` / `namePlaceholder` string collision — now distinct.
- The backend `Form(None)` + filename-stem fallback in `src/xagent/web/api/kb.py` is still present and still lets direct API callers around the empty-name check. This remains correctly out of scope for a frontend-only PR fixing the dialog's own bug.
- Test-helper dedup: partial. `goToStep3` is used by the four new/modified name-validation tests, but four pre-existing tests still hand-roll the same navigation (see Simplification).

Also worth calling out positively: switching the test mock target from `vi.mock("sonner")` to `vi.mock("@/components/ui/sonner")` is the correct target — the component imports `toast` from the wrapper, which is a Proxy that injects `{ duration: 8000 }` into `.error()` calls. Running the suite at the base commit confirms the old single-argument assertion was already failing, so this PR incidentally repairs a pre-existing red assertion.

## Findings

### Medium

**nameError is never reset, so the error survives dialog close/reopen** (`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:177`). `resetState()` (`:318-344`) does not clear `nameError`, and the `open` effect (`:179-183`) only fetches embedding models. The Dialog's `onOpenChange` at `:743` is the raw parent-controlled prop, so ESC / overlay click / close button bypass `resetState()` entirely — and even the Cancel path, which does call it, leaves the flag set. Repro: submit with an empty name → error shown → close the dialog by any means → reopen → the untouched, empty input still shows the stale error. Suggested fix: add `setNameError(false)` to `resetState()`, and ideally also to the `open`-becomes-true effect.

### Recurring — still unfixed at head

- **Step-3 review screen still fabricates a name** (`:1021`, `{newCollectionName || "KB " + new Date().toLocaleString()}`). Not exploitable as a submitted value, but a direct visual contradiction of this PR's purpose. Flagged in both prior rounds.
- **"Next" button does not gate on the name** (`:1264-1275`). This is the mechanism that makes the fabricated-name issue reachable. Suggested fix: add `(currentStep === 1 && !trimmedCollectionName)` to the disabled expression.

### Low / Info

- No test covers close → reopen clearing `nameError` — exactly the gap that let the Medium finding above ship.
- No test asserts the new `aria-required="true"` attribute or the `*` required marker.
- Nit: `nameError`'s `useState` (`:177`) is declared after the derived constant `trimmedCollectionName` (`:175`), breaking the file's otherwise consistent ordering. Cosmetic.
- Pre-existing, not introduced here: `:618` cloud-ingest early return gives no user feedback when zero files are selected, and the Create button's `disabled` condition doesn't cover that case either. Confirmed byte-identical to base — out of scope, mentioned only because this PR's whole premise is eliminating silent no-feedback paths at this exact dispatch site.

A candidate to replace the `nameError` boolean with a derived `hasAttemptedSubmit && !trimmedCollectionName` was investigated and dropped — it changes behavior for whitespace-only input and doesn't remove the need for an explicit reset, just relocates the same bug.

Real, confirmed opportunity — four pre-existing tests still hand-roll navigation that the new `goToStep3` helper already encapsulates (`frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx` around L331/343, L394/403, L461-470, L542-547). `net: -18 lines possible`.

## Testing performed

- `npx vitest run src/components/kb/knowledge-base-creation-dialog.test.tsx` → 12/12 passing at head (2f60836a).
- `npx tsc --noEmit` → clean.
- Grep across the repo for removed symbols (`multiFileNameRequired`, `requiresExplicitCollectionName`, `multiFileRequiredHint`) → zero remaining references.
- Base-commit suite run in an isolated worktree to confirm the `sonner` mock-target change fixes a pre-existing red assertion rather than masking a regression.

## Recommendation

Approving — the core change is correct, the guard's coverage is complete, and tests/typecheck are clean. The Medium finding is a UI-only stale-state bug with a one-line fix; the rest is cosmetic or explicitly out of scope. That said, the two recurring items (fabricated name on step 3, unguarded Next button) have now been raised in all three review rounds — please don't let them go silently unaddressed a third time; either fix them (both are small) or state explicitly they're deferred, ideally with a follow-up issue.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Third time was the charm — thank you for holding the line on this one. The invented preview name is gone, and the Next button on step 1 now enforces the name too, so the fabricated value is no longer reachable at all rather than just unrendered.

Both guards call one `requireCollectionName()` helper (toast + `nameError` + return to step 1), so Next and Create cannot drift apart. I used a click-time guard rather than `disabled` deliberately: the existing `disabled` conditions on that button are silent, and a greyed-out button is not focusable, so a screen-reader user gets nothing — which would undo the field-level feedback added for your accessibility comment.

**One thing I should flag rather than let you find it**: this replaces the per-tab parameterisation you asked for two rounds ago. With the gate moved to step 1, the rejection happens before the user has picked a tab at all, so `activeImportTab` is not in scope — the "someone narrows the guard to the file tab" regression you described is now structurally impossible rather than test-enforced. The six per-tab cases became two step-1 cases (empty / whitespace) that additionally assert `#file-upload` is absent, proving no advance to step 2. The Create-button guard is kept as a last line of defence for the three async handlers, but it is no longer reachable through the UI.

Also worth reporting: restoring the `|| "KB " + new Date()` fallback does **not** fail any test now, because with the step-1 gate that branch is dead code — `newCollectionName` cannot be empty by the time step 3 renders. Deleting it is still right, but I did not write an assertion that appears to catch it and does not; the preview case carries a documenting `not.toMatch(/KB \d/)` with a comment explaining that the real guarantee is the step-1 interception. Fixed in c0b4af1c.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Confirmed and fixed — and your point about the close paths is what determined the fix. `resetState()` would not have been enough: ESC, overlay click and the close button all bypass it. Cleared `nameError` in the `open` effect instead, which every reopen must pass through regardless of how the dialog was closed. One line, all paths, `resetState()` untouched.

Regression test drives exactly that route — flips `open` false then true via rerender, i.e. the path that never reaches `resetState()`. Verified it fails without the fix and passes with it, with no collateral failures. Fixed in c0b4af1c.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Both added. The close/reopen case is the regression test for the stale-state bug above. `aria-required` and the `*` marker now assert inside `expectNameRejected`, so every rejection case covers them — verified by deleting each attribute independently, which fails four cases each. Fixed in c0b4af1c.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Done — swapped for `goToStep3`. Along with the per-tab consolidation described on the Create-button thread, the test file is net -3 lines while assertion density went up (`expectNameRejected` grew from 4 assertions to 7). Fixed in c0b4af1c.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Done — swapped for `goToStep3`. Along with the per-tab consolidation described on the Create-button thread, the test file is net -3 lines while assertion density went up (`expectNameRejected` grew from 4 assertions to 7). Fixed in c0b4af1c.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Done — swapped for `goToStep3`. Along with the per-tab consolidation described on the Create-button thread, the test file is net -3 lines while assertion density went up (`expectNameRejected` grew from 4 assertions to 7). Fixed in c0b4af1c.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Moved above `trimmedCollectionName`, restoring the all-hooks-then-derived-values ordering. Fixed in c0b4af1c.

Issue #1141, opened 2026-08-05
bug: Knowledge Search tool has long retrieval latency


yiboyasss:
### Description

The Knowledge Search tool has a long retrieval time when called

When an Agent uses the Knowledge_search tool to retrieve information from a Knowledge Base, the response takes a long time, causing increased overall agent response latency.

PR #1194, opened 2026-08-07
perf(kb): search knowledge base collections concurrently off the loop

bluefish-08:
Refs #1141

## Problem

`knowledge_search` retrieval is slow. Two compounding causes in `_search_knowledge_base_impl` (`core/tools/core/document_search.py`):

1. **The blocking pipeline runs on the event loop.** `run_document_search` (`RAG_tools/pipelines/document_search.py:944`) is a synchronous `def` called inline from an `async def`. For the whole retrieval the loop is pinned — not just this request, but every other coroutine in the process.

2. **Collections are searched one at a time.** An agent bound to N knowledge bases pays N full pipelines back to back.

These multiply: because the call is synchronous, concurrency was impossible even in principle.

The Web-side `list_collections_api` (`web/api/kb.py:4747`) already went through this fix — per-owner scans moved to `asyncio.gather` and off the loop. The tool path never followed.

## Fix

Extract the per-collection body into an inner `_search_one`, run the blocking pipeline through `asyncio.to_thread`, and fan the collections out with `asyncio.gather`. Total latency becomes the slowest collection instead of the sum, and the loop stays responsive.

Semantics are unchanged: `_search_one` returns `(results, error, warning, documents)` instead of raising, so one failing collection cannot fail the batch, and `gather` preserves input order, so aggregation order matches the serial version. Zero-embedding collections are filtered before fan-out.

Two things the concurrency made necessary, both raised in review:

**Per-collection timeout.** Nothing in the chain bounded a hung backend call. Before, that pinned the event loop — bad, but one at a time. Concurrently it pins one worker of the shared default executor per searched collection, and cancelling the `gather` does not reclaim them. Each `_search_one` is now wrapped in `asyncio.wait_for` with `get_kb_search_timeout_seconds()` (default 60s), built the same way as `get_kb_collections_timeout_seconds`. As that endpoint's comments already note, `wait_for` frees the caller but not the worker — the thread still runs to completion.

Relatedly, `RerankModelConfig.timeout` already existed and `xinference` had always passed it; only `dashscope` skipped it, leaving two `requests.post` calls with no timeout on a path taken whenever a searched collection has a bound rerank model. Wired up, so the timeout above is not just a band-aid over an unbounded thread.

**Readonly search.** The search path reaches `create_index(model_tag, readonly)`, and `SearchConfig.readonly` defaults to False, so retrieval could reach real LanceDB write/commit operations. Serially that was unreachable — both call sites ran bare on the loop, so at most one executed at a time. Concurrently it is a `CommitConflict` race. The pipeline already threads `cfg.readonly` through all three search paths, so setting it here is a one-line fix that short-circuits before any index write.

## Behaviour change worth flagging

`readonly` suppresses **both** index-building paths, not just FTS: `create_index`'s readonly branch returns before reaching either the dense/vector block or the FTS block.

The two are not equally observable. The sparse path emits `READONLY_MODE` and `FTS_INDEX_MISSING` warnings; the dense path hardcodes `warnings=[]` on success, so a skipped dense index build produces **no signal at all**.

Concretely, a collection whose FTS index is missing no longer gets it rebuilt as a side effect of being searched. It degrades per-collection — a warning, the FTS query running unindexed — rather than failing the batch, but it stops self-healing. `search_type` defaults to `hybrid`, so the FTS path is taken on every search; what bounds this is how many collections actually lack the index, not how many searches ask for sparse.

Blast radius is limited: ingestion builds the dense index at the 50,000-row threshold (`DEFAULT_INDEX_ROW_THRESHOLD`) independently of this PR, and failures there are only warn-logged. Tables below that threshold never had a dense index to begin with, so retrieval falls back to brute-force KNN exactly as before.

Note also that `readonly=True` does not mean retrieval stops writing to LanceDB: resolving the embedding model still stamps `last_accessed_at` through `mark_collection_accessed` as a full-row overwrite. That write and its per-collection lock are pre-existing; what is new is N of them landing concurrently on N distinct lock keys.

Scoping readonly to the dense path only is a straightforward alternative if this trade is not wanted.

## Measurement

Local, with `run_document_search` mocked to a fixed 300 ms and 3 collections: **910.8 ms → 305.7 ms**.

This is mock data and only demonstrates the relative change. It does not measure real embedding, vector-search or rerank cost, and does not claim the latency reported in #1141 is resolved — that needs verification against a production-sized deployment, which is why this references rather than closes the issue.

## Tests

`tests/core/tools/core/test_document_search_collection_concurrency.py`: a `threading.Barrier(3)` that only clears if all three searches are genuinely in flight, and that no search ran on the loop thread; per-collection isolation across a raising collection, an error status, and a partial success with warnings; aggregation order, `total_searched`, and the zero-embedding skip; every collection filtered out; every collection failing at once; warnings surviving an empty result set; fan-out cancellation; a stuck collection timing out without holding up the rest; and retrieval running readonly.

Plus timeout coverage for the config accessor, `dashscope`, and the rerank adapter.

```
pytest tests/core/tools/core/ tests/core/model/rerank/ tests/core/test_config.py
```

## Verification

pre-commit: ruff, ruff format, mypy, isort, codespell — all passed

### DIFF
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -61,6 +61,7 @@ FILE_DELIVERY_ACCEL_REDIRECT_PREFIX = "XAGENT_FILE_DELIVERY_ACCEL_REDIRECT_PREFI
 SANDBOX_IMAGE = "SANDBOX_IMAGE"
 LANCEDB_PATH = "LANCEDB_PATH"
 KB_COLLECTIONS_TIMEOUT_SECONDS = "XAGENT_KB_COLLECTIONS_TIMEOUT_SECONDS"
+KB_SEARCH_TIMEOUT_SECONDS = "XAGENT_KB_SEARCH_TIMEOUT_SECONDS"
 DATABASE_URL = "DATABASE_URL"
 DB_POOL_SIZE = "XAGENT_DB_POOL_SIZE"
 DB_MAX_OVERFLOW = "XAGENT_DB_MAX_OVERFLOW"
@@ -1967,6 +1968,32 @@ def get_kb_collections_timeout_seconds() -> int:
         Per-scan timeout in seconds
     """
     return _get_positive_int_env(KB_COLLECTIONS_TIMEOUT_SECONDS, 30)
+
+
+def get_kb_search_timeout_seconds() -> int:
+    """Get the deadline for searching a single knowledge base collection.
+
+    Bounds one ``run_document_search`` call so an agent's knowledge_search does
+    not wait forever on a stuck backend. The collections of one search run
+    concurrently, so each of them holds a shared default-executor worker for as
+    long as it runs; without a deadline an agent bound to N knowledge bases can
+    pin N workers indefinitely. The default is generous because the budget has
+    to cover embedding the query, the LanceDB scan and an optional rerank round
+    trip - the rerank HTTP client has no timeout of its own.
+
+    Known limitation, same as the collection listing endpoint: cancelling the
+    ``asyncio.wait_for`` coroutine does not stop the underlying ``to_thread``
+    worker, so a timed-out search keeps running to completion in the default
+    executor. The deadline frees the caller, not the worker.
+
+    Priority:
+        1. XAGENT_KB_SEARCH_TIMEOUT_SECONDS environment variable
+        2. Default of 60 seconds
+
+    Returns:
+        Per-collection search timeout in seconds
+    """
+    return _get_positive_int_env(KB_SEARCH_TIMEOUT_SECONDS, 60)
 
 
 def get_default_sqlite_db_path() -> str:
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -1979,12 +1979,17 @@ def get_kb_search_timeout_seconds() -> int:
     long as it runs; without a deadline an agent bound to N knowledge bases can
     pin N workers indefinitely. The default is generous because the budget has
     to cover embedding the query, the LanceDB scan and an optional rerank round
-    trip - the rerank HTTP client has no timeout of its own.
+    trip.
 
     Known limitation, same as the collection listing endpoint: cancelling the
     ``asyncio.wait_for`` coroutine does not stop the underlying ``to_thread``
     worker, so a timed-out search keeps running to completion in the default
-    executor. The deadline frees the caller, not the worker.
+    executor. The deadline frees the caller, not the worker. A worker stuck in
+    rerank outlives it by that model's own budget (``RerankModelConfig.timeout``,
+    180s by default).
+    # ponytail: that budget is not clamped to this one - clamping means
+    # threading a per-call timeout through SearchConfig into the rerank
+    # adapter. Do it if leaked workers actually starve the executor.
 
     Priority:
         1. XAGENT_KB_SEARCH_TIMEOUT_SECONDS environment variable
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -1986,10 +1986,7 @@ def get_kb_search_timeout_seconds() -> int:
     worker, so a timed-out search keeps running to completion in the default
     executor. The deadline frees the caller, not the worker. A worker stuck in
     rerank outlives it by that model's own budget (``RerankModelConfig.timeout``,
-    180s by default).
-    # ponytail: that budget is not clamped to this one - clamping means
-    # threading a per-call timeout through SearchConfig into the rerank
-    # adapter. Do it if leaked workers actually starve the executor.
+    180s by default), so it can outlast this deadline threefold.
 
     Priority:
         1. XAGENT_KB_SEARCH_TIMEOUT_SECONDS environment variable
@@ -1998,6 +1995,9 @@ def get_kb_search_timeout_seconds() -> int:
     Returns:
         Per-collection search timeout in seconds
     """
+    # ponytail: the rerank budget is not clamped to this one - clamping means
+    # threading a per-call timeout through SearchConfig into the rerank adapter.
+    # Do it if leaked workers actually starve the executor.
     return _get_positive_int_env(KB_SEARCH_TIMEOUT_SECONDS, 60)
 
 
--- a/src/xagent/core/model/rerank/dashscope.py
+++ b/src/xagent/core/model/rerank/dashscope.py
@@ -66,12 +66,18 @@ class DashscopeRerank(BaseRerank):
                 ``model``'s format family.
             top_n: Number of top results to return.
             instruct: Custom instruction for reranking (new-format models).
-            timeout: HTTP request timeout in seconds (default: 60).
+            timeout: HTTP request timeout in seconds. Only direct callers ever
+                hit the 60s fallback: ``create_rerank_adapter`` always passes
+                ``RerankModelConfig.timeout``, whose own default is 180s, so
+                that is the effective timeout in the application.
         """
         self.model = model
         self.api_key = api_key or os.getenv("DASHSCOPE_API_KEY")
         self.top_n = top_n
         self.instruct = instruct
+        # Matches XinferenceRerank's fallback rather than ModelConfig's 180s:
+        # the two providers should not disagree on what a bare instantiation
+        # means, and the adapter overrides it in every real code path anyway.
         self.timeout = float(timeout) if timeout is not None else 60.0
         self.url = base_url or _default_url_for(model)
 
--- a/src/xagent/core/model/rerank/dashscope.py
+++ b/src/xagent/core/model/rerank/dashscope.py
@@ -54,6 +54,7 @@ class DashscopeRerank(BaseRerank):
         base_url: Optional[str] = None,
         top_n: Optional[int] = None,
         instruct: Optional[str] = None,
+        timeout: Optional[float] = None,
     ):
         """
         Initialize Dashscope rerank model.
@@ -65,11 +66,13 @@ class DashscopeRerank(BaseRerank):
                 ``model``'s format family.
             top_n: Number of top results to return.
             instruct: Custom instruction for reranking (new-format models).
+            timeout: HTTP request timeout in seconds (default: 60).
         """
         self.model = model
         self.api_key = api_key or os.getenv("DASHSCOPE_API_KEY")
         self.top_n = top_n
         self.instruct = instruct
+        self.timeout = float(timeout) if timeout is not None else 60.0
         self.url = base_url or _default_url_for(model)
 
         if not self.api_key:
@@ -132,7 +135,9 @@ class DashscopeRerank(BaseRerank):
                 "parameters": {"return_documents": True} | optional_params,
             }
 
-        response = requests.post(self.url, headers=headers, json=payload)
+        response = requests.post(
+            self.url, headers=headers, json=payload, timeout=self.timeout
+        )
         response.raise_for_status()
         data = response.json()
 
@@ -195,7 +200,9 @@ class DashscopeRerank(BaseRerank):
                 "parameters": {"return_documents": True} | optional_params,
             }
 
-        response = requests.post(self.url, headers=headers, json=payload)
+        response = requests.post(
+            self.url, headers=headers, json=payload, timeout=self.timeout
+        )
         response.raise_for_status()
         data = response.json()
 
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -13,6 +13,10 @@ from .RAG_tools.pipelines.document_search import run_document_search
 
 logger = logging.getLogger(__name__)
 
+# Prefix of the serialized READONLY_MODE warning that search_sparse raises
+# unconditionally under readonly=True (see collection_handle.search_sparse).
+_READONLY_WARNING_PREFIX = "READONLY_MODE:"
+
 if TYPE_CHECKING:
     from .RAG_tools.kb import KBToolCompatibilityFacade
 
@@ -371,6 +375,11 @@ async def _search_knowledge_base_impl(
             """
             collection_name = collection_info.name
 
+            def _failure(
+                reason: str,
+            ) -> tuple[list[Dict[str, Any]], Optional[str], Optional[str], int]:
+                return [], f"{collection_name}: {reason}", None, 0
+
             # Per-KB rerank resolution: explicit tool arg wins, otherwise
             # use the collection's bound rerank_model_id; when neither is
             # set, no rerank stage is added for this collection.
@@ -414,18 +423,22 @@ async def _search_knowledge_base_impl(
                         collection_name,
                         error_message,
                     )
-                    return (
-                        [],
-                        f"{collection_name}: {error_message or 'search failed'}",
-                        None,
-                        0,
-                    )
+                    return _failure(f"{error_message or 'search failed'}")
 
+                # The pipeline's message on a non-success status is boilerplate
+                # ("Hybrid search completed with warnings"), so the warning list
+                # has to win or every real diagnostic is masked by it. The
+                # readonly notice is self-inflicted by our own readonly=True and
+                # fires on every search; FTS_INDEX_MISSING, the warning that
+                # reports an actual consequence of it, is kept.
                 warning: Optional[str] = None
-                if result.status != "success" or result.warnings:
-                    warning_message = result.message or "; ".join(result.warnings)
-                    if warning_message:
-                        warning = f"{collection_name}: {warning_message}"
+                warning_message = "; ".join(
+                    w
+                    for w in result.warnings
+                    if not w.startswith(_READONLY_WARNING_PREFIX)
+                )
+                if warning_message:
+                    warning = f"{collection_name}: {warning_message}"
 
                 if not result.results:
                     return [], None, warning, 0
@@ -443,16 +456,10 @@ async def _search_knowledge_base_impl(
                     collection_name,
                     search_timeout_seconds,
                 )
-                return (
-                    [],
-                    f"{collection_name}: search timed out after "
-                    f"{search_timeout_seconds}s",
-                    None,
-                    0,
-                )
+                return _failure(f"search timed out after {search_timeout_seconds}s")
             except Exception as e:
                 logger.warning(f"Failed to search collection '{collection_name}': {e}")
-                return [], f"{collection_name}: {e}", None, 0
+                return _failure(str(e))
 
         # Skip collections with no embeddings before fanning out.
         searchable = []
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -15,6 +15,11 @@ logger = logging.getLogger(__name__)
 
 # Prefix of the serialized READONLY_MODE warning that search_sparse raises
 # unconditionally under readonly=True (see collection_handle.search_sparse).
+# Matched against the string, not SearchWarning.code: the pipeline flattens
+# warnings to f"{code}: {message}" in _serialize_warnings before they reach us,
+# and the structured object is not plumbed this far. Change that format and this
+# filter silently stops matching - the readonly notice reappears in summaries,
+# which the readonly tests in test_document_search_collection_concurrency catch.
 _READONLY_WARNING_PREFIX = "READONLY_MODE:"
 
 if TYPE_CHECKING:
@@ -348,10 +353,22 @@ async def _search_knowledge_base_impl(
             "top_k": tool_args.top_k,
             "min_score": tool_args.min_score,
             "merge_results": True,
-            # Retrieval must not build indexes: create_index() commits to the
+            # Retrieval must not *build indexes*: create_index() commits to the
             # LanceDB table, and searching collections concurrently can now race
             # that commit (see collection_manager's CommitConflict note).
             # Indexes are built during ingestion, which is where they belong.
+            #
+            # This suppresses both index paths, not just FTS: the readonly branch
+            # of create_index returns before the dense/vector block and before the
+            # FTS block (lancedb_stores.py). Only the sparse path reports it
+            # (READONLY_MODE / FTS_INDEX_MISSING) - dense search hardcodes
+            # warnings=[] on success, so a skipped dense index build is silent.
+            #
+            # Not a claim that the search writes nothing: resolving the embedding
+            # model still stamps last_accessed_at per collection
+            # (collection_manager.mark_collection_accessed). That row-overwrite is
+            # pre-existing and guarded by a per-collection lock, so the N
+            # concurrent searches this fan-out creates land on N distinct keys.
             "readonly": True,
         }
 
@@ -373,32 +390,40 @@ async def _search_knowledge_base_impl(
             Returns (results, error, warning, documents_searched); failures are
             returned rather than raised so one collection cannot fail the batch.
             """
-            collection_name = collection_info.name
+            # Every attribute read lives inside the try below, so this really
+            # cannot raise into the gather. _failure reads the name lazily, so
+            # it stays usable even if that first read is what failed.
+            collection_name = "<unknown>"
 
             def _failure(
                 reason: str,
             ) -> tuple[list[Dict[str, Any]], Optional[str], Optional[str], int]:
                 return [], f"{collection_name}: {reason}", None, 0
 
-            # Per-KB rerank resolution: explicit tool arg wins, otherwise
-            # use the collection's bound rerank_model_id; when neither is
-            # set, no rerank stage is added for this collection.
-            search_config = dict(base_search_config)
-            collection_rerank = getattr(collection_info, "rerank_model_id", None)
-            effective_rerank = tool_args.rerank_model_id or collection_rerank
-            if effective_rerank:
-                search_config["rerank_model_id"] = effective_rerank
-            storage_user_id = getattr(collection_info, "storage_user_id", None)
-
             try:
+                collection_name = collection_info.name
+
+                # Per-KB rerank resolution: explicit tool arg wins, otherwise
+                # use the collection's bound rerank_model_id; when neither is
+                # set, no rerank stage is added for this collection.
+                search_config = dict(base_search_config)
+                collection_rerank = getattr(collection_info, "rerank_model_id", None)
+                effective_rerank = tool_args.rerank_model_id or collection_rerank
+                if effective_rerank:
+                    search_config["rerank_model_id"] = effective_rerank
+                storage_user_id = getattr(collection_info, "storage_user_id", None)
+
                 logger.info(
                     f"Searching collection '{collection_name}' for: {tool_args.query}"
                 )
 
                 # run_document_search is a blocking sync pipeline; running it
                 # inline would pin the event loop for the whole retrieval.
-                # The deadline frees this caller but not the worker: a timed-out
-                # to_thread call keeps running in the shared default executor.
+                # The deadline covers queueing too, not just execution: it starts
+                # when this coroutine is scheduled, so a saturated default
+                # executor can burn it before run_document_search even starts.
+                # And it frees this caller but not the worker - a timed-out
+                # to_thread call keeps running in that shared executor.
                 result = await asyncio.wait_for(
                     asyncio.to_thread(
                         run_document_search,
@@ -474,8 +499,14 @@ async def _search_knowledge_base_impl(
         # Search every collection concurrently: total latency is bounded by the
         # slowest collection instead of the sum of all of them. gather preserves
         # input order, so aggregation order is unchanged.
-        # ponytail: no concurrency cap - an agent binds a handful of KBs; add a
-        # Semaphore if that ever stops being true.
+        # ponytail: no concurrency cap. Holds for the bound-KB paths above, where
+        # the agent's own configuration is the bound. It does NOT hold for the
+        # final fallback branch: with neither `collections` nor
+        # `allowed_collections` set, this fans out over every collection visible
+        # to the caller, which _list_visible_collections unions across owners -
+        # unbounded by anything the agent declared. Add a Semaphore when that
+        # branch gets real traffic, or when fan-out width starves the shared
+        # default executor that every asyncio.to_thread caller draws from.
         for results, error, warning, documents in await asyncio.gather(
             *(_search_one(collection_info) for collection_info in searchable)
         ):
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from pydantic import BaseModel, Field
 
+from ....config import get_kb_search_timeout_seconds
 from .RAG_tools.core.schemas import ListCollectionsResult
 from .RAG_tools.management.collections import list_collections
 from .RAG_tools.pipelines.document_search import run_document_search
@@ -343,6 +344,11 @@ async def _search_knowledge_base_impl(
             "top_k": tool_args.top_k,
             "min_score": tool_args.min_score,
             "merge_results": True,
+            # Retrieval must not build indexes: create_index() commits to the
+            # LanceDB table, and searching collections concurrently can now race
+            # that commit (see collection_manager's CommitConflict note).
+            # Indexes are built during ingestion, which is where they belong.
+            "readonly": True,
         }
 
         if tool_args.embedding_model_id:
@@ -353,6 +359,7 @@ async def _search_knowledge_base_impl(
         collection_errors: list[str] = []
         collection_warnings: list[str] = []
         total_searched = 0
+        search_timeout_seconds = get_kb_search_timeout_seconds()
 
         async def _search_one(
             collection_info: Any,
@@ -381,15 +388,22 @@ async def _search_knowledge_base_impl(
 
                 # run_document_search is a blocking sync pipeline; running it
                 # inline would pin the event loop for the whole retrieval.
-                result = await asyncio.to_thread(
-                    run_document_search,
-                    collection=collection_name,
-                    query_text=tool_args.query,
-                    config=search_config,
-                    user_id=storage_user_id if storage_user_id is not None else user_id,
-                    is_admin=False
-                    if getattr(collection_info, "ownership", "personal") == "team"
-                    else is_admin,
+                # The deadline frees this caller but not the worker: a timed-out
+                # to_thread call keeps running in the shared default executor.
+                result = await asyncio.wait_for(
+                    asyncio.to_thread(
+                        run_document_search,
+                        collection=collection_name,
+                        query_text=tool_args.query,
+                        config=search_config,
+                        user_id=storage_user_id
+                        if storage_user_id is not None
+                        else user_id,
+                        is_admin=False
+                        if getattr(collection_info, "ownership", "personal") == "team"
+                        else is_admin,
+                    ),
+                    timeout=search_timeout_seconds,
                 )
 
                 if result.status not in {"success", "partial_success"}:
@@ -423,6 +437,19 @@ async def _search_knowledge_base_impl(
                     results.append(res_dict)
                 return results, None, warning, collection_info.documents
 
+            except asyncio.TimeoutError:
+                logger.warning(
+                    "Search of collection '%s' exceeded %ss",
+                    collection_name,
+                    search_timeout_seconds,
+                )
+                return (
+                    [],
+                    f"{collection_name}: search timed out after "
+                    f"{search_timeout_seconds}s",
+                    None,
+                    0,
+                )
             except Exception as e:
                 logger.warning(f"Failed to search collection '{collection_name}': {e}")
                 return [], f"{collection_name}: {e}", None, 0
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -1,5 +1,6 @@
 """Core document search functionality for RAG pipelines."""
 
+import asyncio
 import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
@@ -353,15 +354,15 @@ async def _search_knowledge_base_impl(
         collection_warnings: list[str] = []
         total_searched = 0
 
-        for collection_info in collections_to_iterate:
-            collection_name = collection_info.name
+        async def _search_one(
+            collection_info: Any,
+        ) -> tuple[list[Dict[str, Any]], Optional[str], Optional[str], int]:
+            """Search one collection off the event loop.
 
-            # Skip collections with no embeddings
-            if collection_info.embeddings == 0:
-                logger.debug(
-                    f"Skipping collection with no embeddings: {collection_name}"
-                )
-                continue
+            Returns (results, error, warning, documents_searched); failures are
+            returned rather than raised so one collection cannot fail the batch.
+            """
+            collection_name = collection_info.name
 
             # Per-KB rerank resolution: explicit tool arg wins, otherwise
             # use the collection's bound rerank_model_id; when neither is
@@ -378,7 +379,10 @@ async def _search_knowledge_base_impl(
                     f"Searching collection '{collection_name}' for: {tool_args.query}"
                 )
 
-                result = run_document_search(
+                # run_document_search is a blocking sync pipeline; running it
+                # inline would pin the event loop for the whole retrieval.
+                result = await asyncio.to_thread(
+                    run_document_search,
                     collection=collection_name,
                     query_text=tool_args.query,
                     config=search_config,
@@ -390,36 +394,65 @@ async def _search_knowledge_base_impl(
 
                 if result.status not in {"success", "partial_success"}:
                     error_message = result.message or "; ".join(result.warnings)
-                    collection_errors.append(
-                        f"{collection_name}: {error_message or 'search failed'}"
-                    )
                     logger.warning(
                         "Search pipeline returned status '%s' for collection '%s': %s",
                         result.status,
                         collection_name,
                         error_message,
                     )
-                    continue
+                    return (
+                        [],
+                        f"{collection_name}: {error_message or 'search failed'}",
+                        None,
+                        0,
+                    )
 
+                warning: Optional[str] = None
                 if result.status != "success" or result.warnings:
                     warning_message = result.message or "; ".join(result.warnings)
                     if warning_message:
-                        collection_warnings.append(
-                            f"{collection_name}: {warning_message}"
-                        )
+                        warning = f"{collection_name}: {warning_message}"
 
-                if result.results:
-                    for res in result.results:
-                        res_dict = dict(res)
-                        res_dict["collection"] = collection_name
-                        all_results.append(res_dict)
+                if not result.results:
+                    return [], None, warning, 0
 
-                    total_searched += collection_info.documents
+                results = []
+                for res in result.results:
+                    res_dict = dict(res)
+                    res_dict["collection"] = collection_name
+                    results.append(res_dict)
+                return results, None, warning, collection_info.documents
 
             except Exception as e:
-                collection_errors.append(f"{collection_name}: {e}")
                 logger.warning(f"Failed to search collection '{collection_name}': {e}")
+                return [], f"{collection_name}: {e}", None, 0
+
+        # Skip collections with no embeddings before fanning out.
+        searchable = []
+        for collection_info in collections_to_iterate:
+            if collection_info.embeddings == 0:
+                logger.debug(
+                    f"Skipping collection with no embeddings: {collection_info.name}"
+                )
                 continue
+            searchable.append(collection_info)
+
+        # Search every collection concurrently: total latency is bounded by the
+        # slowest collection instead of the sum of all of them. gather preserves
+        # input order, so aggregation order is unchanged.
+        # ponytail: no concurrency cap - an agent binds a handful of KBs; add a
+        # Semaphore if that ever stops being true.
+        for results, error, warning, documents in await asyncio.gather(
+            *(_search_one(collection_info) for collection_info in searchable)
+        ):
+            if error:
+                collection_errors.append(error)
+                continue
+            if warning:
+                collection_warnings.append(warning)
+            if results:
+                all_results.extend(results)
+                total_searched += documents
 
         if not all_results:
             if collection_errors:
--- a/tests/core/tools/core/test_document_search_collection_concurrency.py
+++ b/tests/core/tools/core/test_document_search_collection_concurrency.py
@@ -50,12 +50,17 @@ def _pipeline_result(collection: str, **overrides: Any) -> SearchPipelineResult:
     return SearchPipelineResult(**payload)
 
 
-def _install_collections(monkeypatch: pytest.MonkeyPatch, *names: str) -> None:
+def _install_collections(
+    monkeypatch: pytest.MonkeyPatch, *names: str, embeddings: int = 10
+) -> None:
     async def _list(
         user_id: int | None = None, is_admin: bool = False
     ) -> ListCollectionsResult:
         del user_id, is_admin
-        return _collections(*names)
+        listing = _collections(*names)
+        for collection in listing.collections:
+            collection.embeddings = embeddings
+        return listing
 
     monkeypatch.setattr(document_search, "_list_visible_collections", _list)
 
@@ -80,30 +85,39 @@ async def test_collections_are_searched_concurrently_off_the_event_loop(
     def blocking_search(*, collection: str, **_kwargs: Any) -> SearchPipelineResult:
         with lock:
             worker_thread_ids.add(threading.get_ident())
-        # Only passes if all three searches are simultaneously in flight.
+        # Passing the barrier needs three OS threads resident at once, which is
+        # only possible if the loop dispatched all three without blocking on any.
         barrier.wait()
         return _pipeline_result(collection)
 
     monkeypatch.setattr(document_search, "run_document_search", blocking_search)
 
-    ticks = 0
-
-    async def tick() -> None:
-        nonlocal ticks
-        while True:
-            ticks += 1
-            await asyncio.sleep(0)
-
-    ticker = asyncio.create_task(tick())
-    try:
-        result = await document_search._search_knowledge_base_impl(_args(), user_id=1)
-    finally:
-        ticker.cancel()
+    result = await document_search._search_knowledge_base_impl(_args(), user_id=1)
 
     assert len(result.results) == 3
-    # The loop stayed responsive while the blocking pipeline ran.
-    assert ticks > 0
     assert loop_thread_id not in worker_thread_ids
+
+
+@pytest.mark.asyncio
+async def test_search_runs_readonly_so_retrieval_never_builds_an_index(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Concurrent retrieval must not reach create_index's LanceDB commit."""
+    _install_collections(monkeypatch, "alpha", "beta")
+    configs: list[dict[str, Any]] = []
+
+    def search(
+        *, collection: str, config: dict[str, Any], **_kwargs: Any
+    ) -> SearchPipelineResult:
+        configs.append(config)
+        return _pipeline_result(collection)
+
+    monkeypatch.setattr(document_search, "run_document_search", search)
+
+    await document_search._search_knowledge_base_impl(_args(), user_id=1)
+
+    assert len(configs) == 2
+    assert all(config["readonly"] is True for config in configs)
 
 
 @pytest.mark.asyncio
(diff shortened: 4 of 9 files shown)


rogercloud (commented):
## Summary

This PR changes exactly two files (`git diff origin/main...HEAD`): `src/xagent/core/tools/core/document_search.py` and a new test file. Inside `_search_knowledge_base_impl`, the per-collection search body is extracted into an `async def _search_one(...)` that returns a `(results, error, warning, documents)` tuple, the blocking `run_document_search` pipeline is pushed off the event loop via `asyncio.to_thread`, and the collections are fanned out with `asyncio.gather` after zero-embedding collections are filtered out up front. The aggregation loop is unchanged in behavior: `gather` preserves input order, so `results`, `total_searched`, and the relative order of `collection_errors` / `collection_warnings` are identical to before, and a single bad collection still degrades to a per-collection error instead of failing the batch.

The offload is genuinely effective here, not cosmetic: embedding and rerank are HTTP calls, and vector scan / BM25 run inside LanceDB's native extension, so the GIL is actually released and `to_thread` really does buy parallelism.

## Design verdict

**Sound in direction and mechanics; acceptable with reservations on operational safety.**

The refactor is semantically faithful and the concurrency model is the right one for this shape of work. Two reservations, both about what changes when this pipeline stops being effectively single-flight:

1. **The cited precedent is copied selectively.** The fan-out shape and even the "bounded by the slowest instead of the sum" comment closely mirror `list_collections_api` in `src/xagent/web/api/kb.py:4747`, but that precedent wraps each gathered coroutine in `asyncio.wait_for(..., get_kb_collections_timeout_seconds())` (kb.py:4762-4764, 4796-4798) and this one does not. There is no timeout anywhere in the chain from the tool entrypoint down to `run_document_search` (checked `document_search.py`, `adapters/vibe/document_search.py`, `RAG_tools/kb/tool_compatibility.py`, and `react.py:_execute_tool`). This gap is pre-existing — the PR does not remove a timeout — but it changes the *failure mode*, which is why it belongs in this review. See the inline comment on the `gather`.

2. **A structurally single-flight pipeline becomes genuinely concurrent.** `run_document_search` has exactly two call sites process-wide, and before this PR both ran bare and blocking on the event loop, so at most one could physically execute at any instant inside a process. This PR is the first change to run several of them on real concurrent OS threads, which newly exposes an in-process race in the search-triggered `create_index` path. Details inline.

Note that cancellation is *not* separately fixable here: cancelling the outer `gather` does not stop already-running `to_thread` workers, and swapping in the file's `run_db_io_cancellation_safe` helper would not help (that helper is for SQLAlchemy-session-owning work, which this pipeline has none of, and it would only delay cancellation propagation). That observation is just the other face of the missing-timeout gap, not a separate defect.

**Deferred-scope note (not a blocker):** the root cause is only half-fixed. The other `run_document_search` call site, the search endpoint at `src/xagent/web/api/kb.py:5191`, still runs the blocking pipeline bare on the event loop and is untouched here. That is a reasonable thing to defer to a follow-up given this PR is scoped to the agent tool path, but worth a tracking issue so #1141 does not look closed while the HTTP path still pins the loop.

## Findings

### Major

- **`src/xagent/core/tools/core/document_search.py:445` — no timeout on the fan-out; the blast radius of a hung backend grows with the number of bound KBs.** Pre-existing gap, but this PR changes what it costs. Previously a hung backend call blocked the shared event loop: bad, but self-evidently one slot at a time. Now it occupies one worker of the shared default `ThreadPoolExecutor` (`min(32, cpu+4)` workers, shared with ~218 other `to_thread` call sites) *per concurrently searched collection*, so an agent bound to N KBs pointed at one slow endpoint burns N slots at once with no way to reclaim them. This is reachable in practice: `src/xagent/core/model/rerank/dashscope.py:135` and `:199` both call `requests.post(...)` with no `timeout=` kwarg, on a path taken whenever a searched collection has a bound `rerank_model_id`. Recommend wrapping each `_search_one` call in `asyncio.wait_for` with a configurable timeout, matching the `kb.py:4747` precedent the PR cites as its model.

- **`src/xagent/core/tools/core/document_search.py:384` — concurrent `run_document_search` newly exposes an in-process `create_index` race.** The search path reaches `RAG_tools/kb/collection_handle.py:1997` `create_index(model_tag, readonly)`, where `readonly` comes from `SearchConfig.readonly` (`RAG_tools/core/schemas.py:886`, default `False`) and is never set to `True` in `document_search.py`. With `readonly=False`, `create_index` reaches real LanceDB write/commit operations (`RAG_tools/storage/lancedb_stores.py:1356` `table.create_index(...)`, `:1400` `create_fts_index(..., replace=True)`), and this codebase's own comment at `RAG_tools/management/collection_manager.py:59-62` acknowledges that concurrent commits to the same table "including the create_index build" can raise `CommitConflict`. None of the existing collection locks (`_get_collection_thread_lock` / `_collection_thread_guard`, only at `collection_manager.py:328,482,596,688` and `vector_manager.py:98`) cover this search-triggered call — they all sit on the metadata/ingestion paths. Before this PR the race was structurally unreachable in-process because both `run_document_search` call sites blocked the loop; now two collections sharing a table can commit simultaneously. Scoped caveat: the app defaults to a single worker process (`web/__main__.py` calls `uvicorn.run` with no `workers=`), so all of the existing locking is already single-process-scoped — this is a newly reachable in-process race under that deployment model, not a claim about multi-replica risk (which was already there and which no lock here would fix). Simplest mitigation is passing `readonly=True` on the search path if the pipeline permits it; otherwise the search-triggered `create_index` needs the same collection guard the ingestion paths use.

### Minor

- **`tests/core/tools/core/test_document_search_collection_concurrency.py:136` — coverage gaps, plus one assertion that passes for the wrong reason.** `assert "warned: ok" in result.summary` only passes because the fixture's incidental default `message="ok"` wins over `warnings` via the `or` at `document_search.py:410-414` — the test's own configured `warnings=["fell back to vector"]` is never actually asserted, so warning-list propagation could be entirely broken and this test would still pass. Uncovered branches: all collections filtered out by zero embeddings (`gather()` with zero args), all collections failing at once (the errors-only summary branch at `document_search.py:458-467` is never reached, since one collection always succeeds in the mixed-failure test), and the warning-with-empty-results return at `document_search.py:417`. Cancellation is also untested. Suggest asserting the real warning text here and adding an all-filtered, an all-failing, and a warning-with-no-results case.

FYI, non-blocking and not caused by this PR: `mark_collection_accessed_sync` → `_run_in_separate_loop` takes the `asyncio.run()` branch inside `to_thread` workers, and `_get_collection_lock` keys on `id(asyncio.get_running_loop())`, so each call leaks a fresh never-contended lock into a module-global dict. `collection_manager.py` is untouched by this diff and the total number of leaked entries per search call is the same serial or concurrent — only the churn density changes. Worth a separate issue, not a change request here.

`tests/core/tools/core/test_document_search_collection_concurrency.py`
- L89: `shrink:` the hand-rolled `ticks` / `tick()` / `ticker` event-loop-liveness counter is redundant. Reaching the assertions already requires three real OS threads simultaneously parked in `threading.Barrier(3, timeout=5)`, which is impossible if the loop is blocked — the barrier would time out and fail first, before `assert ticks > 0` is ever evaluated. Delete lines 89-97, 101 and 104-105 and keep the barrier plus `worker_thread_ids` checks, which already prove loop responsiveness deterministically. (This is a redundancy, not the flaky-timing pattern removed in `06a5a1c6`; do not reuse `_EventLoopLivenessProbe` from `test_websocket_owner_actor.py`, it is private and single-shot.)

net: -9 lines possible

## Scope check

The `tests/web/api/test_websocket_owner_actor.py` and `tests/web/api/test_workforce_share_link.py` changes visible in some views are not scope creep — they belong to already-merged commit `06a5a1c6`, which is `origin/main` HEAD and this PR's direct parent. `git diff origin/main...HEAD` touches exactly `document_search.py` and the new concurrency test.

## Tests

Ran the PR's stated command in a clean checkout: `pytest tests/core/tools/core/test_document_search_collection_concurrency.py tests/core/tools/core/test_document_search_team_visibility_boundary.py tests/core/tools/core/RAG_tools/kb/test_tool_compatibility.py tests/core/tools/adapters/vibe/test_knowledge_search.py` — 54 passed. The PR description's claim checks out.

The prior gemini-code-assist run left no actionable feedback, so there is nothing to reconcile from it.

src/xagent/core/tools/core/document_search.py
```diff
@@ -390,36 +394,65 @@ async def _search_knowledge_base_impl(
...
+                )
+                continue
+            searchable.append(collection_info)
+
+        # Search every collection concurrently: total latency is bounded by the
+        # slowest collection instead of the sum of all of them. gather preserves
+        # input order, so aggregation order is unchanged.
+        # ponytail: no concurrency cap - an agent binds a handful of KBs; add a
+        # Semaphore if that ever stops being true.
+        for results, error, warning, documents in await asyncio.gather(
```
rogercloud, line 510: Major: this fan-out has no timeout anywhere in the chain down to `run_document_search` (checked this file, `adapters/vibe/document_search.py`, `RAG_tools/kb/tool_compatibility.py`, and `react.py:_execute_tool` — zero `wait_for`/`timeout` matches).

The gap is pre-existing, but this PR changes what it costs. Before, a hung backend call blocked the shared event loop — bad, but one slot at a time. Now it holds one worker of the shared default `ThreadPoolExecutor` (`min(32, cpu+4)` workers, shared with the rest of the codebase's `to_thread` call sites) per concurrently searched collection, so an agent bound to N KBs against one slow endpoint consumes N slots at once. Cancelling the outer `gather` does not reclaim them; nothing can cancel a running `to_thread` worker.

This is reachable: `src/xagent/core/model/rerank/dashscope.py:135` and `:199` both call `requests.post(...)` with no `timeout=`, on a path taken whenever a searched collection has a bound `rerank_model_id`.

The precedent this fan-out is modeled on, `list_collections_api` in `src/xagent/web/api/kb.py:4747`, already solves this — it wraps each gathered coroutine in `asyncio.wait_for(..., get_kb_collections_timeout_seconds())` at lines 4762-4764 and 4796-4798. Suggest doing the same per `_search_one` call with a configurable timeout.

src/xagent/core/tools/core/document_search.py
```diff
@@ -378,7 +379,10 @@ async def _search_knowledge_base_impl(
                     f"Searching collection '{collection_name}' for: {tool_args.query}"
                 )
 
-                result = run_document_search(
+                # run_document_search is a blocking sync pipeline; running it
+                # inline would pin the event loop for the whole retrieval.
+                result = await asyncio.to_thread(
```
rogercloud: Major: running this pipeline on real concurrent threads newly exposes an in-process `create_index` race.

The search path reaches `RAG_tools/kb/collection_handle.py:1997` `create_index(model_tag, readonly)`. `readonly` comes from `SearchConfig.readonly` (`RAG_tools/core/schemas.py:886`, default `False`) and is never set to `True` in this file, so `create_index` reaches real LanceDB write/commit operations — `RAG_tools/storage/lancedb_stores.py:1356` `table.create_index(...)` and `:1400` `create_fts_index(..., replace=True)`. The codebase's own comment at `RAG_tools/management/collection_manager.py:59-62` states that concurrent commits to the same table, "including the create_index build," can raise `CommitConflict`, and none of the existing collection guards (`_get_collection_thread_lock` / `_collection_thread_guard` at `collection_manager.py:328,482,596,688` and `vector_manager.py:98`) cover this call — they all sit on the metadata/ingestion paths.

Before this change the race was structurally unreachable in-process: `run_document_search` has exactly two call sites (this one and `web/api/kb.py:5191`) and both ran bare on the event loop, so at most one could execute at any instant. This is the first change to launch several of them concurrently in one process.

Scoping this honestly: the app defaults to a single worker process (`web/__main__.py` calls `uvicorn.run` with no `workers=`), so all existing locking here is already single-process-scoped — this is about a newly reachable in-process race under that model, not new multi-replica risk. Cleanest fix is passing `readonly=True` on the search path if the pipeline allows it; otherwise this `create_index` needs the same collection guard the ingestion paths already use.

tests/core/tools/core/test_document_search_collection_concurrency.py
```diff
@@ -0,0 +1,172 @@
...
+        return _pipeline_result(collection)
+
+    monkeypatch.setattr(document_search, "run_document_search", search)
+
+    result = await document_search._search_knowledge_base_impl(_args(), user_id=1)
+
+    assert [entry.collection for entry in result.results] == ["alpha", "warned"]
+    assert "boom: connection reset" in result.summary
+    assert "status_error: nope" in result.summary
+    assert "warned: ok" in result.summary
```
rogercloud: Minor: this assertion passes for the wrong reason. `"warned: ok"` matches because the fixture's default `message="ok"` wins over `warnings` via the `or` at `document_search.py:410-414` — the `warnings=["fell back to vector"]` this test configures is never actually asserted, so warning-list propagation could be completely broken and this test would still be green. Assert the real warning text instead.

Related coverage gaps in this file worth filling: all collections filtered out by zero embeddings (`gather()` with zero args), all collections failing at once (the errors-only summary branch at `document_search.py:458-467` is never reached because one collection always succeeds in the mixed-failure test), and the warning-with-empty-results return at `document_search.py:417`. Cancellation of the fan-out is untested too.

tests/core/tools/core/test_document_search_collection_concurrency.py
```diff
@@ -0,0 +1,172 @@
...
+    def blocking_search(*, collection: str, **_kwargs: Any) -> SearchPipelineResult:
+        with lock:
+            worker_thread_ids.add(threading.get_ident())
+        # Only passes if all three searches are simultaneously in flight.
+        barrier.wait()
+        return _pipeline_result(collection)
+
+    monkeypatch.setattr(document_search, "run_document_search", blocking_search)
+
+    ticks = 0
```
rogercloud: shrink: this `ticks` / `tick()` / `ticker` liveness counter is redundant with the checks already in this test. Getting past the barrier requires three real OS threads simultaneously parked in `threading.Barrier(3, timeout=5)`, which cannot happen if the event loop is blocked — the barrier would time out and fail the test before `assert ticks > 0` is even evaluated. The barrier plus `worker_thread_ids` already prove loop responsiveness deterministically and fail loudly on a real regression.

Delete lines 89-97, 101 and 104-105 (net -9). To be clear, this is redundancy, not the CI-timing fragility that `06a5a1c6` removed from `test_websocket_owner_actor.py` — and no need to reuse `_EventLoopLivenessProbe` from that file, it is private and single-shot.

src/xagent/core/tools/core/document_search.py
bluefish-08: Agreed, and fixed — each `_search_one` is now wrapped in `asyncio.wait_for` with a new `get_kb_search_timeout_seconds()` built the same way as `get_kb_collections_timeout_seconds`. Its docstring states plainly that `wait_for` frees the caller but not the worker; the thread still runs to completion, same known limitation the collections endpoint documents.

One correction to the premise, which changed how I fixed it: the chain is not entirely timeout-free. `RerankModelConfig.timeout` already exists and `xinference.py` has always passed it (`:60`, `:186`, `:297`). Only dashscope skipped it — so this was a provider that dropped an existing mechanism rather than a missing concept. That made it a 6-line fix rather than new machinery, so I did it: `dashscope.py` now takes and passes `timeout`, and `adapter.py` wires `model_config.timeout` the same way its xinference branch already did. `retry_on` at `adapter.py:11-16` already lists `requests.exceptions.Timeout` as retryable, so the retry semantics were waiting for this.

Without that, `wait_for` would be a band-aid: the caller walks away and the worker stays pinned on the executor slot indefinitely — the exact problem you raised. Fixed in 5941e354.

src/xagent/core/tools/core/document_search.py
bluefish-08: Took option 1 — `base_search_config` now sets `readonly: True`. Worth noting it is even cheaper than your phrasing suggests: the pipeline already threads `cfg.readonly` through all three search paths (`pipelines/document_search.py:625`, `:853`, `:878`), so `collection_handle` needs no change at all. The only gap was this caller never setting the field, and `SearchConfig.readonly` defaulting to False. With it set, `lancedb_stores.py:1285` short-circuits to `list_indices()` and returns before `create_index` / `create_fts_index`, so the commit race is structurally gone rather than serialized.

I did not take option 2 because a per-collection lock only serializes the commit — the index build stays on the retrieval path, which is part of the slowness this PR is about — and our fan-out is across distinct collections, so the lock would rarely be contended anyway.

**One behaviour change that needs your call.** For a collection whose FTS index is missing, today's search silently rebuilds it (slow once, then self-healed). Under readonly it will not: `create_index` returns `fts_enabled=False`, `collection_handle.py:2569` appends an `FTS_INDEX_MISSING` warning, and the FTS query runs unindexed. It degrades per-collection rather than failing the batch, but it stops self-healing.

On reachability — `search_type` defaults to `hybrid` (`core/document_search.py:96`), so the FTS path is taken on every search, not occasionally. What bounds the blast radius is how many collections actually lack the index, not how many searches ask for sparse. Indexes are built at ingestion (`collection_handle.py:1808`, `create_index` defaults True) and we skip collections with `embeddings == 0`, so anything searchable has been ingested — but that ingestion-time build is a try/except that only warns, so historical failures would leave such tables behind.

Happy to scope readonly to the dense path only, or to gate it, if you would rather not take that trade. Fixed in 5941e354.

tests/core/tools/core/test_document_search_collection_concurrency.py
bluefish-08: You are right, and I had not spotted it — the fixture's `message="ok"` won the `or` and the configured warnings list was never asserted. That collection now sets `message=""` so `"; ".join(result.warnings)` is the only possible source, and the assertion checks the real warning text. Verified: dropping the warning from the return value now fails.

All four gaps filled: every collection filtered out by zero embeddings (`gather()` with no args), all collections failing at once (the errors-only branch, plus a negative assertion that the no-results summary is not used), warning-with-empty-results, and fan-out cancellation. Fixed in 5941e354.

tests/core/tools/core/test_document_search_collection_concurrency.py
bluefish-08: Deleted. Your reasoning holds: three OS threads cannot park in the barrier if the loop is blocked, so the barrier plus `worker_thread_ids` already prove liveness deterministically and the counter was measuring nothing the barrier did not. Replaced with a comment saying so, and did not reach for `_EventLoopLivenessProbe`. Fixed in 5941e354.

src/xagent/core/tools/core/document_search.py
```diff
@@ -378,48 +386,100 @@ async def _search_knowledge_base_impl(
...
+                    return (
+                        [],
+                        f"{collection_name}: {error_message or 'search failed'}",
+                        None,
+                        0,
+                    )
 
+                warning: Optional[str] = None
                 if result.status != "success" or result.warnings:
                     warning_message = result.message or "; ".join(result.warnings)
```
rogercloud: `readonly: True` (newly added to `base_search_config` in this PR) makes `search_sparse` append a `READONLY_MODE` warning unconditionally (`RAG_tools/kb/collection_handle.py:2545-2553`), and hybrid always calls `search_sparse` (`:2957-2967`), so status flips to `partial_success` (`:3069`) on every default-path search. Because the hybrid pipeline always sets a generic non-empty `message` (`RAG_tools/pipelines/document_search.py:908-916`), `result.message or "; ".join(result.warnings)` here means `result.warnings` is never surfaced — so a genuine FTS fallback, rerank failure, or `FTS_INDEX_MISSING` warning is silently swallowed behind boilerplate text.

Suggest either surfacing both (or preferring `warnings` when `message` is the generic string), or special-casing `READONLY_MODE` so it doesn't flip status on its own. No test covers this path.

src/xagent/core/tools/core/document_search.py
```diff
@@ -378,48 +386,100 @@ async def _search_knowledge_base_impl(
...
-                        f"{collection_name}: {error_message or 'search failed'}"
-                    )
                     logger.warning(
                         "Search pipeline returned status '%s' for collection '%s': %s",
                         result.status,
                         collection_name,
                         error_message,
                     )
-                    continue
+                    return (
```
rogercloud: Three call sites in `_search_one` (here, ~446, ~455) hand-build the same `([], msg, None, 0)` failure tuple. A local `def _failure(msg): return [], msg, None, 0` would save ~8 lines and give one place to change the shape.

src/xagent/config.py
```diff
@@ -1969,6 +1970,32 @@ def get_kb_collections_timeout_seconds() -> int:
...
+def get_kb_search_timeout_seconds() -> int:
+    """Get the deadline for searching a single knowledge base collection.
+
+    Bounds one ``run_document_search`` call so an agent's knowledge_search does
+    not wait forever on a stuck backend. The collections of one search run
+    concurrently, so each of them holds a shared default-executor worker for as
+    long as it runs; without a deadline an agent bound to N knowledge bases can
+    pin N workers indefinitely. The default is generous because the budget has
+    to cover embedding the query, the LanceDB scan and an optional rerank round
+    trip - the rerank HTTP client has no timeout of its own.
```
rogercloud: Stale as of this commit — dashscope's rerank client now has a 180s timeout (`src/xagent/core/model/model.py:31`). Also worth noting 180s exceeds this 60s default it nests inside; consider clamping the rerank timeout from this value on the KB search path.

rogercloud (changes requested):
## Summary

This PR turns the per-collection loop in `_search_knowledge_base_impl` into a concurrent fan-out: the per-collection body moves into an inner `_search_one`, the blocking `run_document_search` call is pushed off the event loop with `asyncio.to_thread`, and the agent's bound collections are searched in parallel via `asyncio.gather`. Total latency becomes the slowest collection rather than the sum, and the loop is no longer blocked — addressing #1141. The change follows the existing `list_collections_api` precedent in `web/api/kb.py`, and pulls in two necessary follow-ups: a per-collection `asyncio.wait_for` timeout (a hung backend now pins one pool worker per collection) and `readonly: True` on the search config (concurrent index-build commits to LanceDB could now conflict).

## Re-review status

This is the second round. All four findings from the previous round are confirmed fixed in `5941e354`:

- **Fan-out had no timeout** — fixed: wrapped in `asyncio.wait_for(get_kb_search_timeout_seconds())`, and dashscope's rerank HTTP client got a real timeout in the same change. Verified by a passing stuck-collection timeout test.
- **`create_index` CommitConflict race** — fixed: `readonly: True` in `base_search_config` structurally short-circuits before `create_index`/`create_fts_index` are reachable, not merely less likely. (A different, narrower race on a *shared* table remains — see Other observations.)
- **Test assertion passing for the wrong reason** — fixed: the fixture now uses an empty message, so the assertion actually exercises warning propagation. The four flagged coverage gaps (zero-embedding fan-out, all-collections-failing, warning-with-empty-results, cancellation) all have dedicated tests now.
- **Redundant `ticks`/`tick()` liveness counter in the concurrency test** — fixed: fully removed in favour of the `threading.Barrier`, with an explanatory comment.

## Design verdict

**Acceptable with reservations.** The fan-out itself is the right minimal fix and mirrors an established in-repo pattern faithfully. The two follow-ups are directionally correct, but the `readonly` fix closes only half the newly-exposed commit-race surface and introduces a warning-masking regression on the tool's default search path (Major below). Remaining concerns — rerank timeout budget, unbounded executor fan-out, metadata write race — are either pre-existing, deliberately disclosed in-code, or low real-world impact, and none should block merge once the Major is addressed.

## Major

### M1 — `readonly: True` makes every hybrid search return a content-free warning and masks real ones

Newly setting `readonly: True` in `base_search_config` activates a warning path that was dormant before this PR (the schema default is `False`):

- `RAG_tools/kb/collection_handle.py:2545-2553` — `search_sparse` appends a `READONLY_MODE` `SearchWarning` unconditionally whenever readonly is set (covered by the existing test at `tests/core/tools/core/RAG_tools/retrieval/test_search_sparse.py:276-326`).
- `collection_handle.py:2957-2967` — the hybrid path always calls `search_sparse`, so hybrid always collects that warning.
- `collection_handle.py:3069` — `status = "success" if not all_warnings else "partial_success"`, so hybrid now always reports `partial_success`.
- `RAG_tools/pipelines/document_search.py:908-916` — for hybrid, `message` is unconditionally one of two fixed generic strings, i.e. always truthy.
- `core/tools/core/document_search.py:409-410` and `:424-428` — `warning_message = result.message or "; ".join(result.warnings)`. Because `message` is always truthy for hybrid, `result.warnings` is **never** surfaced. The generic string is then concatenated into the agent-visible summary at `document_search.py:492,501,509`.

Hybrid is the tool's default `search_type` (`document_search.py:95-98`), so every default-path `knowledge_search` call now returns `partial_success` plus boilerplate warning text — and any genuinely useful co-occurring warning (FTS fallback, rerank failure, `FTS_INDEX_MISSING` from the disclosed self-healing tradeoff) is silently swallowed behind it. There is no test covering this.

**Suggestion:** either stop letting `message` mask `warnings` (surface both, or prefer `warnings` when `message` is generic boilerplate), or special-case/suppress the `READONLY_MODE` warning so it does not flip status to `partial_success` on its own.

*Nuance:* the dense-path readonly "advice" string does not independently trigger this under a normal success status — this finding is specifically about the sparse-path warning list feeding hybrid fusion, which is the default.

## Other observations

- **`collection_metadata` write race** (`RAG_tools/management/collection_manager.py:679-701,1302`, `RAG_tools/storage/lancedb_stores.py:149-164`): the fan-out produces N concurrent `merge_insert` commits to the single shared `collection_metadata` table; `readonly` only covers the per-collection embeddings write. Non-fatal — retried 3x with backoff, then swallowed at `logger.debug`; worst case is a missed `last_accessed_at`. Follow-up, not a blocker.
- **Rerank timeout budget mismatch**: the rerank HTTP timeout (180s default, `core/model/model.py:31`, `core/model/rerank/dashscope.py`) is larger than the per-collection search timeout it nests inside (60s default, `config.py:get_kb_search_timeout_seconds`). Only affects collections with a bound rerank model, and the outer `wait_for` still frees the caller — a partial rather than total defeat of the safeguard. Consider clamping the rerank timeout from the search timeout on this path.
- **No `Semaphore` cap on the fan-out** (`document_search.py:466-475`): deliberate and disclosed via the in-code `ponytail:` comment with a stated remediation trigger. The shared default `ThreadPoolExecutor` is process-wide, so a hung backend costs one worker slot per concurrently-searched collection with no reclaim on timeout. Acceptable as-is; worth a tracking issue.
- **Test nit** (`tests/core/tools/core/test_document_search_collection_concurrency.py:264,283`): the timeout config is monkeypatched to a float (`0.2`), but the real accessor only ever returns an int, so the integer-formatted message path (`"timed out after 60s"`) is untested.
- **Test nit** (`test_search_runs_readonly_so_retrieval_never_builds_an_index`, ~101-120): asserts only that a dict key is `True`, not that it reaches `SearchConfig` or actually suppresses an index write. Worth strengthening given M1.
- **`ProgressManager._main_loop` capture** (`RAG_tools/progress/manager.py:38-58`): if the search path is the first construction site in the process, it now happens inside a `to_thread` worker with no running loop. Self-heals on the next call from a loop thread and the `task_id` is never exposed to any consumer here — low-priority robustness nit only.
- **FYI, pre-existing and unchanged by this PR**: `total_searched` / collection count in the final summary undercounts and uses the wrong denominator (`document_search.py:480-482,497-498`); and `merge_results: True` in `base_search_config` (`document_search.py:346`) has no matching `SearchConfig` field and is silently dropped by pydantic's extra-ignore default.
- **Scope note**: `web/api/kb.py`'s `/search` endpoint (~line 5190) still calls `run_document_search` inline on the loop with `readonly` defaulting to `False` — same root cause, different entry point. Correctly out of scope here, but #1141 shouldn't be closed as fully fixed without a follow-up.

`L417: shrink Three call sites in _search_one hand-build the same failure tuple ([], msg, None, 0). Extract a local _failure(msg) helper. net: -8 lines possible.`

## Testing

445 relevant tests pass (`pytest tests/core/tools/core/ tests/core/model/rerank/ tests/core/test_config.py`) at the PR head; the new concurrency suite uses a real `threading.Barrier` to prove parallelism rather than asserting on mocks.

src/xagent/core/tools/core/document_search.py
bluefish-08: Confirmed the whole chain, and fixed — but I went with your first option rather than special-casing `READONLY_MODE`, because the special case would only change how often the bug fires.

The `or` itself is the defect and it predates this PR: `message` is a pipeline status sentence, `warnings` is a diagnostic list, and using one as a fallback for the other means any hybrid search producing a real warning has it masked by boilerplate. `readonly: True` turned that from occasional into every-search. Suppressing the readonly notice alone would restore the old frequency while leaving FTS fallbacks, rerank failures and `FTS_INDEX_MISSING` just as invisible.

So the warning branch now joins `result.warnings` directly, with only the `READONLY_MODE:` notice filtered — that one is self-inflicted by our own `readonly=True` and carries nothing actionable. **`FTS_INDEX_MISSING` is deliberately kept**, since it reports the actual consequence of readonly; that also makes the behaviour change discussed on the other thread observable rather than silent. The error branch keeps `message` first, since there it is real content from `_handle_search_error`, not boilerplate.

This is contained to our own file — `_fuse_hybrid` status logic is untouched, so `web/api/kb.py:5191` is unaffected.

One thing that surfaced: two tests in `tests/core/tools/adapters/vibe/test_knowledge_search.py` (`:804`, `:860`) had the bug pinned — they build a real `warnings=[...]` and then assert the **boilerplate** appears in the summary, never asserting the warning itself. Same failure mode you flagged below. Both now assert the real warning is present and the boilerplate is absent. Fixed in 8e19d353.

src/xagent/core/tools/core/document_search.py
bluefish-08: Done — local `_failure(reason)` in `_search_one`, three call sites collapsed onto it. Fixed in 8e19d353.

src/xagent/config.py
bluefish-08: I checked this against the current tip (`569720a7`) and I do not think it is stale — dashscope still has no timeout there:

```
$ git show upstream/main:src/xagent/core/model/rerank/dashscope.py | grep -n "timeout\|requests.post"
135:        response = requests.post(self.url, headers=headers, json=payload)
198:        response = requests.post(self.url, headers=headers, json=payload)
```

The 180s at `model/model.py:31` is a field on `ModelConfig`, and `adapter.py` on the tip passes it in only one place — the xinference branch. The dashscope branch never picked it up, so that value has never reached its HTTP client. What this PR adds is precisely the wiring, not a duplicate. Happy to be corrected if you are looking at something I am missing.

On 180 > 60: I left it. Clamping properly means threading a per-call timeout from `document_search` through `SearchConfig` into rerank adapter creation, across three-plus modules and into pipeline-internal signatures, and the user-visible latency is already bounded by the outer `wait_for(60)` — the only gain is shortening how long a leaked worker survives. I documented the ceiling in the config docstring instead, with the upgrade path noted, so it is visible rather than silently inconsistent. Also removed a now-false line from that docstring claiming the rerank client has no timeout of its own.

src/xagent/core/tools/core/document_search.py
```diff
@@ -342,6 +348,11 @@ async def _search_knowledge_base_impl(
             "top_k": tool_args.top_k,
             "min_score": tool_args.min_score,
             "merge_results": True,
+            # Retrieval must not build indexes: create_index() commits to the
+            # LanceDB table, and searching collections concurrently can now race
+            # that commit (see collection_manager's CommitConflict note).
```
rogercloud, line 358: Minor: this comment justifies `readonly: True` by saying retrieval "must not build indexes" / commit to LanceDB, but every search still performs an unconditional metadata read-modify-write: `run_document_search` -> `resolve_effective_embedding_model_sync` -> `mark_collection_accessed_sync` (RAG_tools/management/collection_manager.py:1301) -> `mark_collection_accessed` (:679-701), which stamps `last_accessed_at` via a full-row overwrite in `_save_collection_with_retry` (:349-363).

That write and its defenses (a process-global per-collection `threading.RLock` at collection_manager.py:63-95, plus a 3-attempt retry) are pre-existing and unchanged by this PR — correctness is defended, since different collections use different lock keys. What is newly true is that N of these writes, to N different collections, can now race in parallel instead of being serialized one at a time by the event loop.

Suggestion: scope the comment's claim to "no index-build commits" specifically, or note the metadata write is intentionally out of scope.

src/xagent/core/tools/core/document_search.py
```diff
@@ -378,48 +395,98 @@ async def _search_knowledge_base_impl(
...
-                    is_admin=False
-                    if getattr(collection_info, "ownership", "personal") == "team"
-                    else is_admin,
+                # run_document_search is a blocking sync pipeline; running it
+                # inline would pin the event loop for the whole retrieval.
+                # The deadline frees this caller but not the worker: a timed-out
+                # to_thread call keeps running in the shared default executor.
+                result = await asyncio.wait_for(
+                    asyncio.to_thread(
+                        run_document_search,
```
rogercloud, line 429: Minor: the per-collection `asyncio.wait_for(..., timeout=search_timeout_seconds)` deadline is measured from when the coroutine is scheduled, not from when the shared default `ThreadPoolExecutor` actually picks up the work. Under executor saturation a collection can report "timed out after 60s" without `run_document_search` ever having started.

`asyncio.to_thread` submits to the loop's default executor, and `wait_for`'s cancellation only succeeds if the underlying `concurrent.futures.Future` has not started running — once picked up, cancellation is a no-op and the thread runs to completion regardless. The comment at :399-401 already acknowledges the second half of this ("the deadline frees this caller but not the worker"); the scheduling-latency half is worth acknowledging too, since it is self-reinforcing: an unreclaimed timed-out worker keeps holding a slot.

src/xagent/core/tools/core/document_search.py
```diff
@@ -378,48 +395,98 @@ async def _search_knowledge_base_impl(
...
+                logger.debug(
+                    f"Skipping collection with no embeddings: {collection_info.name}"
+                )
+                continue
+            searchable.append(collection_info)
+
+        # Search every collection concurrently: total latency is bounded by the
+        # slowest collection instead of the sum of all of them. gather preserves
+        # input order, so aggregation order is unchanged.
+        # ponytail: no concurrency cap - an agent binds a handful of KBs; add a
```
rogercloud: Minor: this ponytail note declines a `Semaphore` on the premise that "an agent binds a handful of KBs" — but there is a branch where that premise doesn't hold. When neither `collections` nor `allowed_collections` is specified on the tool call (:337-343), all collections visible to the caller are fanned out concurrently with no cap. `_list_visible_collections` (:31-75) unions across every distinct owning user, so this is not bounded to a handful.

Deliberate and disclosed per the ponytail convention, so not blocking — but the stated justification is inaccurate for the "search everything visible" fallback specifically.

src/xagent/core/tools/core/document_search.py
```diff
@@ -352,16 +363,22 @@ async def _search_knowledge_base_impl(
...
+        search_timeout_seconds = get_kb_search_timeout_seconds()
 
-        for collection_info in collections_to_iterate:
+        async def _search_one(
+            collection_info: Any,
+        ) -> tuple[list[Dict[str, Any]], Optional[str], Optional[str], int]:
+            """Search one collection off the event loop.
+
+            Returns (results, error, warning, documents_searched); failures are
+            returned rather than raised so one collection cannot fail the batch.
```
rogercloud, line 391: Minor (robustness, not an active bug): `collection_name = collection_info.name` is read before `_search_one` enters its own try/except, and `collection_info.embeddings` is read in the pre-fan-out filter loop (:466-472). Since `asyncio.gather(...)` at :479-481 is called without `return_exceptions=True`, an exception from either attribute access would propagate out of `gather` and fail the entire batch — contradicting `_search_one`'s own docstring claim that failures are returned rather than raised — and `gather` does not cancel sibling in-flight tasks when one raises.

Not reachable today: `CollectionInfo` (RAG_tools/core/schemas.py:1324-1376) has a required `name: str` and a defaulted `embeddings: int`, so neither access can raise given current validated inputs. Flagging only because the isolation guarantee has this one narrow hole.

src/xagent/core/tools/core/document_search.py
```diff
@@ -342,6 +348,11 @@ async def _search_knowledge_base_impl(
             "top_k": tool_args.top_k,
             "min_score": tool_args.min_score,
             "merge_results": True,
+            # Retrieval must not build indexes: create_index() commits to the
+            # LanceDB table, and searching collections concurrently can now race
+            # that commit (see collection_manager's CommitConflict note).
+            # Indexes are built during ingestion, which is where they belong.
+            "readonly": True,
```
rogercloud, line 372: Minor (documentation): `readonly: True` suppresses index-building on both the dense/vector path and the sparse/FTS path — `create_index`'s readonly branch (RAG_tools/storage/lancedb_stores.py:1285-1304) returns before reaching either the dense creation block (:1340-1356) or the FTS block (:1396-1414). But only the sparse path surfaces a warning about it (`READONLY_MODE`/`FTS_INDEX_MISSING`); the dense path hardcodes `warnings=[]` unconditionally on success (RAG_tools/kb/collection_handle.py:2172, 2240), so a skipped dense index build produces zero signal. The PR description currently discloses only the FTS/sparse half.

Blast radius is bounded: ingestion builds the dense index at a 50,000-row threshold (RAG_tools/core/config.py:20) regardless of this PR, and failure there is only warn-logged. Suggest one extra sentence in the PR description alongside the existing FTS paragraph.

src/xagent/core/tools/core/document_search.py
```diff
@@ -1,16 +1,22 @@
...
+from ....config import get_kb_search_timeout_seconds
 from .RAG_tools.core.schemas import ListCollectionsResult
 from .RAG_tools.management.collections import list_collections
 from .RAG_tools.pipelines.document_search import run_document_search
 
 logger = logging.getLogger(__name__)
 
+# Prefix of the serialized READONLY_MODE warning that search_sparse raises
+# unconditionally under readonly=True (see collection_handle.search_sparse).
+_READONLY_WARNING_PREFIX = "READONLY_MODE:"
```
rogercloud, line 23: Minor (coupling): `_READONLY_WARNING_PREFIX = "READONLY_MODE:"` filters against an already-serialized warning string (rendered by the pipeline's `_serialize_warnings` as `f"{code}: {message}"`) rather than a structured `SearchWarning.code` field — see the filter at :434-439. Correct today, but fragile to any future change in that format.

Worth flagging that the fix is not a same-file tweak: `result.warnings` as it reaches this module is already `List[str]`, and the structured `SearchWarning` objects are discarded one layer down inside the pipeline. `SearchPipelineResult` is a public, frozen model also used as a FastAPI `response_model` (web/api/kb.py:5067) with multiple internal consumers, so surfacing a typed code means a schema/API-surface change, not a local edit. Fine to leave as-is for this PR.

src/xagent/config.py
```diff
@@ -1969,6 +1970,37 @@ def get_kb_collections_timeout_seconds() -> int:
...
+    trip.
+
+    Known limitation, same as the collection listing endpoint: cancelling the
+    ``asyncio.wait_for`` coroutine does not stop the underlying ``to_thread``
+    worker, so a timed-out search keeps running to completion in the default
+    executor. The deadline frees the caller, not the worker. A worker stuck in
+    rerank outlives it by that model's own budget (``RerankModelConfig.timeout``,
+    180s by default).
+    # ponytail: that budget is not clamped to this one - clamping means
+    # threading a per-call timeout through SearchConfig into the rerank
```
rogercloud: Minor: this `# ponytail:` scope-limiting note is placed before the closing `"""` of the docstring (which spans :1974-2000), so it is syntactically part of the string literal rather than a stripped comment token — `python -c "print(get_kb_search_timeout_seconds.__doc__)"` prints the ponytail text, and it will also appear in `help()` and any generated API docs. The sibling ponytail note in `document_search.py:477` is correctly placed as a real comment outside any docstring; suggest matching that by moving this below the closing `"""`.

src/xagent/core/model/rerank/dashscope.py
```diff
@@ -65,11 +66,13 @@ def __init__(
                 ``model``'s format family.
             top_n: Number of top results to return.
             instruct: Custom instruction for reranking (new-format models).
+            timeout: HTTP request timeout in seconds (default: 60).
```
rogercloud: Minor (informational — confirming a documented tradeoff, not a newly found bug): the docstring here and the constructor fallback at :75 say the timeout defaults to 60s, but `adapter.py:31,43` now always passes `timeout=model_config.timeout` with no `None` check, and `RerankModelConfig.timeout`'s actual default (inherited from `ModelConfig`, core/model/model.py:31) is `180.0`. So the effective default any real caller gets is 180s, not 60s — and 180s exceeds the new 60s per-collection search deadline.

Already disclosed as intentional: `config.py:1984-1991` records exactly this gap as a "Known limitation" with a ponytail note explaining why it isn't clamped yet, so no behavior change is needed. Suggestion is just to fix the wording here (and the fallback value), since it's now misleading about the effective default — it mirrors a pre-existing, equally misleading comment in `xinference.py`.

rogercloud (approved):
## Summary

This PR converts the per-collection knowledge-base search loop in `_search_knowledge_base_impl` (`src/xagent/core/tools/core/document_search.py`) from a serial, event-loop-blocking loop into a concurrent fan-out: the per-collection body moves into an inner `_search_one`, the blocking sync pipeline call `run_document_search` is pushed off the loop with `asyncio.to_thread`, and all bound collections are searched in parallel via `asyncio.gather`. This addresses linked issue #1141 (P0: "Knowledge Search tool has long retrieval latency").

Two necessary follow-ups ship with it: a per-collection `asyncio.wait_for` deadline backed by a new config knob `get_kb_search_timeout_seconds()` (default 60s, modeled on the existing `get_kb_collections_timeout_seconds`), since a hung backend now pins a shared-executor worker per collection; and `readonly: True` in the search config, since concurrent `create_index` commits to LanceDB could otherwise race with a `CommitConflict`. A rerank HTTP timeout was also wired into the DashScope rerank adapter (previously only Xinference had one) so the new per-collection deadline is not defeated by an unbounded `requests.post`.

## Re-review status — all prior blocking findings fixed

This is a second follow-up round.

- **M1 (Major, was blocking)** — "`readonly: True` makes every hybrid search return a content-free warning and masks real ones" — **FIXED** in the latest commit. The success/`partial_success` branch now builds the warning message directly from `result.warnings`, filtered to drop the self-inflicted `READONLY_MODE:` prefix, instead of letting the boilerplate `result.message` mask it entirely. Two new tests cover exactly this — `test_real_warnings_are_not_masked_by_the_boilerplate_message` and `test_the_self_inflicted_readonly_notice_produces_no_warning` — and both pass.
- The earlier round's four findings (no timeout in the fan-out chain; `create_index` `CommitConflict` race; a test assertion passing for the wrong reason; a redundant liveness counter in the concurrency test) were confirmed fixed previously and remain fixed.
- The earlier round's simplification suggestion (extract a local `_failure(reason)` helper to replace three hand-built failure tuples in `_search_one`) is implemented and confirmed behaviorally equivalent.
- No new major or critical findings surfaced in this round.

## Design verdict: Sound

This is the minimal correct shape for the stated problem, and it mirrors an established in-repo pattern (`list_collections_api` in `web/api/kb.py`) closely enough to read as one house style: the same per-item `wait_for` deadline, the same `_get_positive_int_env`-backed config accessor, and the same documented "cancelling `wait_for` does not reclaim the worker" caveat. It improves on that precedent, since `_search_one` returns `(results, error, warning, documents)` rather than letting one timeout fail the whole request. Aggregation semantics are genuinely preserved (gather order, `total_searched`, zero-embedding skip), no shared mutable state leaks into the tasks, and multi-tenancy contextvars survive the move off the loop.

The `readonly` flag is load-bearing rather than scope creep: ingestion serializes its `create_index` commit under a per-collection `RLock`, while the two search-path `create_index` call sites take no such lock — serially unreachable before, a real race under concurrency. A fully-async pipeline rewrite would be far larger for the same win, and a dedicated thread pool/queue would be more machinery than the change justifies.

Design-level notes (not owned by any single line, none blocking):

- The per-collection deadline is enforced against a process-wide, shared, uncapped default executor — there are 233 other `asyncio.to_thread` call sites in the codebase contending for the same pool, so the timeout's meaning degrades under unrelated load.
- The `readonly` flag's in-code justification oversells what it guarantees: it closes the index-commit race it names, but does not make the search path fully read-only.
- The warning-filtering approach couples across a module boundary by matching a serialized string prefix rather than a typed code.
- The synchronous pipeline underneath has several pre-existing properties that were previously masked by the event loop serializing everything (a per-search metadata write, redundant model-hub resolution, a connection-cache/TTL mismatch). This PR is the first thing to run them in parallel for real. None is a regression this PR introduces and none should block merge, but collectively they are the reason a bounded executor or a concurrency cap would be the more future-proof choice here.

## Line-level findings (all minor, 8 inline below)

Grouped by theme:

**Concurrency scope and deadline semantics** — deadline measured from scheduling not execution start; unbounded fan-out when no `collections`/`allowed_collections` filter is given; one narrow (currently unreachable) hole in `_search_one`'s failure-isolation guarantee.

**Accuracy of in-code claims and docs** — the `readonly` comment overstates its guarantee; `readonly` also silently suppresses dense-index building (only sparse warns); a `# ponytail:` note is inside a docstring rather than a comment; DashScope's timeout docstring/fallback (60s) doesn't match the effective default (180s, already a documented tradeoff — just a wording fix).

**Pre-existing behavior newly exposed by concurrency** (verified untouched by this diff, so posted as general notes rather than inline since GitHub won't anchor comments outside the diff):
- `src/xagent/core/tools/core/RAG_tools/pipelines/document_search.py:526-528,555` — `_apply_rerank_if_needed` resolves the rerank model twice purely to log a provider name, each opening a fresh, unpooled-sized SQLAlchemy session; harmless (sessions always closed) but this PR's fan-out turns a serial redundancy into concurrent pool pressure.
- `src/xagent/core/tools/core/RAG_tools/storage/lancedb_stores.py:528-552` vs `src/xagent/providers/vector_store/lancedb.py:37-141` — stores cache `self._conn` forever while a separate TTL reaper can close that same connection object elsewhere in the process, with no revalidation path; interacts with the new failure mode where a dead connection now fails N concurrently-searched collections at once instead of one.
- `example.env` — the new `XAGENT_KB_SEARCH_TIMEOUT_SECONDS` knob has no documentation entry, unlike its sibling `XAGENT_KB_COLLECTIONS_TIMEOUT_SECONDS`; AGENTS.md's "Adding New Configuration" checklist step 4 (update example.env) was skipped for this knob.

Lean already. A dedicated simplification pass (dead flexibility, reinvented stdlib, unneeded dependencies, speculative abstractions) found nothing to report — the new config knob mirrors an existing sibling pattern exactly, the rerank timeout wiring fills a pre-existing gap rather than adding new machinery, and the concurrency primitives used are idiomatic and minimal. The previous round's one simplification suggestion (the `_failure` helper) is already implemented.

## Testing

Full suite: 2258 passed, 1 skipped (`test_real_ingestion.py`, "no resolvable real embedding model" — environmental, pre-existing, unrelated). The new concurrency test file is genuinely load-bearing: it uses a real `threading.Barrier(3, timeout=5)` to prove actual parallelism (a regression to serial execution would time out and fail, not merely change an assertion), and separately asserts the event-loop thread ID never appears among worker thread IDs. Coverage spans per-collection failure isolation, aggregation order/totals, zero-embedding skip, all-filtered/all-failing, warnings surviving empty results, fan-out cancellation, timeout without holding up the batch, and retrieval running readonly.

Minor gaps, not blocking: the timeout test monkeypatches the config accessor to a float where the real accessor is int-typed; the readonly test asserts only that a dict key is set rather than that it reaches `SearchConfig`; per-KB rerank resolution and user/admin-scope kwargs threaded through `to_thread` aren't directly asserted by the new file.

## Recommendation

Approve — the fan-out design is sound and the blocking issue from the previous round is fixed. The notes above are optional follow-ups, none of which should hold up merge.

src/xagent/core/model/rerank/dashscope.py
bluefish-08: Confirmed by running it, and this changes what I said on the other timeout thread — worth stating plainly.

Measured:
```
RerankModelConfig.timeout default        : 180.0
dashscope instance built via the adapter : 180.0   <- effective
bare DashscopeRerank(api_key=...)        :  60.0
```

So the effective value is 180s, set by `RerankModelConfig.timeout` inheriting `ModelConfig.timeout`. The 60.0 constructor fallback is unreachable in the application, because `_create_rerank_model` passes `timeout=model_config.timeout` unconditionally.

**Both of your timeout comments are correct and they do not contradict each other**, which I got wrong earlier: on the tip, dashscope genuinely has no timeout at all — so my reply there stands — but it is *this PR* wiring `adapter.py` that makes 180 reach it for the first time. You were describing the consequence of the change, not restating the earlier point. Apologies for reading it as a repeat.

Fixed the docstring rather than the value: `XinferenceRerank` uses the same 60.0 fallback, so changing only dashscope would make the two providers disagree about what bare instantiation means, and altering it is a behaviour change for direct callers with no benefit on the path that actually runs. The docstring now says the fallback applies only to direct construction and that `create_rerank_adapter` always supplies `RerankModelConfig.timeout` (180s by default). Added a test pinning both numbers. Fixed in 7d2fc940.

src/xagent/core/tools/core/document_search.py
bluefish-08: Fixed, but the investigation moved the target — worth recording since your reasoning is what led there.

`collection_name = collection_info.name` turns out not to be reachable as a failure inside `_search_one`: `.name` is already read for every collection at `:279` when building `available_names`, so anything that would raise there has already raised, under the outer handler, long before `gather`. My first test used an exploding `name` property and blew up at `:279` instead of the fan-out, which is how I found this.

What genuinely was outside the try is `getattr(collection_info, "rerank_model_id", None)` and the `storage_user_id` read next to it — `getattr`'s default only swallows `AttributeError`, so a property raising anything else escapes straight into a `gather` with no `return_exceptions`. Moved that whole block (including `collection_name`, now initialised to a sentinel outside so `_failure` is total) into the try, and the new test raises from `rerank_model_id`. Verified: hoisting it back out fails that test with the RuntimeError escaping the batch.

Did not add `return_exceptions=True` — the aggregation loop unpacks the four-tuple directly, so exception objects in the result list would fail at unpacking and need an isinstance branch that is dead code once `_search_one` cannot raise.

Left the `embeddings` read alone: it is in the main body under `_search_knowledge_base_impl`'s own handler, which turns it into the `RuntimeError` that function's docstring already documents — no contradiction with `_search_one`'s claim. Fixed in 7d2fc940.

src/xagent/core/tools/core/document_search.py
bluefish-08: You are right, and the PR description only disclosed the FTS half. Both index paths are suppressed — the readonly branch returns before reaching either block — while only sparse emits a signal, since dense hardcodes `warnings=[]` on success. Comment now states both halves, and I have updated the PR description to match, including your point that ingestion still builds the dense index at the 50,000-row threshold independently of this PR. Fixed in 7d2fc940.

src/xagent/core/tools/core/document_search.py
bluefish-08: Corrected. The comment now says retrieval must not *build indexes*, rather than claiming it does not write to LanceDB — which was wrong, since `mark_collection_accessed` still stamps `last_accessed_at` through a full-row overwrite on every search. Added that explicitly, along with the point that the write and its per-collection lock are pre-existing and that what is new is N of them landing concurrently on N distinct lock keys. Fixed in 7d2fc940.

src/xagent/core/tools/core/document_search.py
bluefish-08: Added the scheduling-latency half. The comment now states the deadline starts when the coroutine is scheduled, not when the executor picks the work up, so a saturated executor can burn it before `run_document_search` begins — alongside the half already there about freeing the caller but not the worker. Fixed in 7d2fc940.

src/xagent/core/tools/core/document_search.py
bluefish-08: The justification was inaccurate and is rewritten. It now separates the two cases: the bound-KB paths are capped by the agent's own configuration, while the fallback branch — neither `collections` nor `allowed_collections` given — fans out to everything visible with no cap, unioned across owners. Both triggers for revisiting a `Semaphore` are named. Did not add one, since that is a behaviour change beyond this round. Fixed in 7d2fc940.

src/xagent/core/tools/core/document_search.py
bluefish-08: Kept the string match and documented the coupling rather than restructuring. Passing structured `SearchWarning` down would change `SearchPipelineResult.warnings` type and every caller with it; matching on `"READONLY_MODE" in w` without the colon looks looser but is actually more fragile, since it would swallow any warning whose message happens to mention the code. `startswith("READONLY_MODE:")` is the tightest match for the current `f"{code}: {message}"` shape.

The real risk is the format changing unnoticed, so the comment names that failure mode explicitly and points at the two readonly tests that assert `"READONLY_MODE" not in summary` — they go red if the serialization changes. Cheaper to catch it with a test than with a type. Fixed in 7d2fc940.

src/xagent/config.py
bluefish-08: Good catch — it was inside the string literal. Moved below the closing `"""`, above the `return`. Verified: `'ponytail' in get_kb_search_timeout_seconds.__doc__` is now False. Also strengthened the docstring to spell out that the rerank ceiling can outlast this deadline threefold, rather than leaving 180 vs 60 for the reader to notice. Fixed in 7d2fc940.

PR #1196, opened 2026-08-08
fix: render generated videos as inline players in chat

### LINKED ISSUE #1191: bug: Render generated videos in Task conversation UI
### Description

Generated videos are not properly displayed in the Task conversation interface.

Currently, when a task generates a video, the conversation UI only shows the video file name instead of rendering the video content.

yiboyasss:
## Problem

Generated videos (e.g. from the video generation tool) show up in the
conversation UI as a plain filename/download link instead of a playable
video element.

## Root causes

There were two independent layers to this bug:

1. **The frontend had no video preview support at all.**
   `getInlineFilePreviewKind` in `inline-file-preview-utils.ts` never
   matched `video/*` mime types or `.mp4/.mov/.webm/...` extensions, so
   every video fell through to the generic `'file'` kind and rendered
   as a download link.

2. **The model rewrites link labels, breaking extension detection.**
   Assistant messages reference files as `[label](file:<uuid>)` — the
   target is an opaque UUID with no extension, so the frontend can only
   detect the file type from the label text. The model frequently
   rewrites the label into descriptive prose (e.g. `下载视频（MP4）`),
   which silently degrades the preview even once video detection
   exists.

## Changes

**Frontend**
- Add a `'video'` kind to the inline file preview pipeline
  (`inline-file-preview-utils.ts`): explicit `type: 'video'` artifacts,
  `video/*` mime types, and `.mp4/.m4v/.mov/.webm/.mpeg/.mpg`
  extensions.
- Add an inline video player (`inline-file-preview.tsx`): authenticated
  blob fetch with public-preview fallback, filename header with an Open
  link, and a `<video controls playsInline>` player. The shared
  loading logic now lives in a `useResolvedMediaUrl` hook and the
  shared wrapper JSX in `InlineMediaPreview`, both reused by the audio
  player.
- Audio/video players now show `loadErrorText` when the media element
  errors after both load paths (authenticated blob, public preview
  fallback) are exhausted, instead of a silent broken player. The
  header keeps the filename and Open link.
- This also fixes video artifacts in the tool execution trace view,
  which already passes `type: "video"` through to `InlineFilePreview`.

**Backend**
- `reconcile_assistant_file_references` now restores the real filename
  as the markdown label for **video/audio** links when the model's
  label drops the extension, so the frontend's extension-based
  detection always fires. Scoped to media player types on purpose:
  rewriting office-document labels would flip existing compact links
  into heavy inline preview boxes that eagerly fetch file bytes.
- The rewrite is skipped for filenames containing `[`, `]`, backslash,
  or control characters — substituting those into a markdown label
  would destroy the link outright (escaping is not idempotent across
  repeated reconcile passes, so skipping is the safe choice).
- `.m4v` added to the backend video extension sets
  (`artifact_type_for_filename`, `GENERATED_ARTIFACT_EXTENSIONS`),
  closing a frontend/backend drift; parity tests now pin both the
  video and audio extension lists against the frontend regexes.

## Behavior change on public/share surfaces (deliberate)

`useResolvedMediaUrl` skips the blob fetch when the policy's
authenticated route resolves to the same URL as the public inline
route — i.e. under `createPublicFileAccessPolicy`, whose guest token
rides the query string. There the tokened URL is assigned directly as
the media element's `src`. This applies to **image and audio as well
as video**, and means the guest token now appears in DOM `src`
attributes, page HTML, and the browser media cache on public widget
pages.

Why this is acceptable:
- The query-string token exists precisely because media elements
  cannot send an `Authorization` header — direct media loads are what
  that policy was designed for (see the comment in
  `file-access-context.tsx`).
- Tokened URLs already land in DOM `href`s on the same surfaces today:
  every generic file download link and office-preview Open link uses
  `inlineDownloadUrl`/`inlinePreviewUrl` in a plain `<a href>`. This PR
  extends an existing exposure pattern to media `src`, it does not
  create a new one.
- The token is a per-guest, 30-day JWT scoped to the guest's own
  session on an already-public share; a leaked URL grants access to
  the leaker's own conversation files, not other users'.
- The win: no double-download, and HTTP range requests work, so
  video/audio can start playing before the full file downloads.

The default in-app (Bearer-authenticated) policy still blob-fetches;
making the authenticated preview route Range-capable is a possible
follow-up (see review discussion).

## Testing

- `tests/web/services/test_file_reference_output_service.py`: 17 cases
  covering the prose-label scenario from the bug report, `.m4v`,
  case-insensitive extensions, office-document exclusion,
  mismatched-but-suffixed labels, unsafe/control-character filenames,
  empty labels, and frontend/backend extension parity for both video
  and audio.
- Frontend: cases across `inline-file-preview-utils.test.ts`,
  `inline-file-preview.test.tsx` and `markdown-renderer.test.tsx`,
  including the end-to-end `MarkdownRenderer → MarkdownLink →
  InlineFilePreview` path, public-policy direct streaming for video,
  the public fallback path, blob-URL revocation on unmount, and the
  terminal-failure error state (86 passed across the touched suites).
- `tsc --noEmit`, `mypy`, `ruff`, `eslint` all clean (the two eslint
  warnings in `inline-file-preview.tsx` are pre-existing).

### DIFF
--- a/frontend/src/components/file/inline-file-preview-utils.ts
+++ b/frontend/src/components/file/inline-file-preview-utils.ts
@@ -49,9 +49,12 @@ const SPREADSHEET_MIME_TYPES = new Set<string>([
   'text/csv',
 ])
 
-/** Default OOXML mime types for inline preview kinds (images vary by file). */
+/**
+ * Default OOXML mime types for inline preview kinds. Kinds without an
+ * entry (file/image/audio/video vary per file) resolve to undefined.
+ */
 export const INLINE_FILE_PREVIEW_MIME_BY_KIND: Partial<
-  Record<PreviewableInlineFileKind, string>
+  Record<InlineFilePreviewKind, string>
 > = {
   presentation:
     'application/vnd.openxmlformats-officedocument.presentationml.presentation',
@@ -61,11 +64,7 @@ export const INLINE_FILE_PREVIEW_MIME_BY_KIND: Partial<
 
 export const getInlineFilePreviewMimeType = (
   kind: InlineFilePreviewKind
-): string | undefined => {
-  if (kind === 'file' || kind === 'image' || kind === 'audio' || kind === 'video')
-    return undefined
-  return INLINE_FILE_PREVIEW_MIME_BY_KIND[kind]
-}
+): string | undefined => INLINE_FILE_PREVIEW_MIME_BY_KIND[kind]
 
 export const isPreviewableInlineFileKind = (
   kind: InlineFilePreviewKind
--- a/frontend/src/components/file/inline-file-preview-utils.ts
+++ b/frontend/src/components/file/inline-file-preview-utils.ts
@@ -3,6 +3,7 @@ import { getFilePublicDownloadUrl, getFilePublicPreviewUrl } from '@/lib/utils'
 export type InlineFilePreviewKind =
   | 'image'
   | 'audio'
+  | 'video'
   | 'presentation'
   | 'document'
   | 'spreadsheet'
@@ -27,6 +28,7 @@ export type PreviewUrlTrust = {
 const PREVIEWABLE_KINDS = new Set<PreviewableInlineFileKind>([
   'image',
   'audio',
+  'video',
   'presentation',
   'document',
   'spreadsheet',
@@ -60,7 +62,8 @@ export const INLINE_FILE_PREVIEW_MIME_BY_KIND: Partial<
 export const getInlineFilePreviewMimeType = (
   kind: InlineFilePreviewKind
 ): string | undefined => {
-  if (kind === 'file' || kind === 'image' || kind === 'audio') return undefined
+  if (kind === 'file' || kind === 'image' || kind === 'audio' || kind === 'video')
+    return undefined
   return INLINE_FILE_PREVIEW_MIME_BY_KIND[kind]
 }
 
@@ -78,6 +81,7 @@ export const getInlineFilePreviewKind = (
 
   if (type === 'image') return 'image'
   if (type === 'audio') return 'audio'
+  if (type === 'video') return 'video'
   if (type === 'presentation') {
     // An explicit ``type: 'presentation'`` artifact must still be
     // cross-checked: pptxviewjs only supports OOXML .pptx, so a
@@ -108,6 +112,7 @@ export const getInlineFilePreviewKind = (
 
   if (mimeType.startsWith('image/')) return 'image'
   if (mimeType.startsWith('audio/')) return 'audio'
+  if (mimeType.startsWith('video/')) return 'video'
   if (PRESENTATION_MIME_TYPES.has(mimeType) || mimeType.includes('presentationml')) {
     return 'presentation'
   }
@@ -118,6 +123,7 @@ export const getInlineFilePreviewKind = (
 
   if (/\.(jpg|jpeg|png|gif|webp|svg)$/.test(filename)) return 'image'
   if (/\.(mp3|wav|ogg|opus|flac|m4a|aac)$/.test(filename)) return 'audio'
+  if (/\.(mp4|m4v|mov|webm|mpeg|mpg)$/.test(filename)) return 'video'
   // Only OOXML .pptx is previewable inline — see PRESENTATION_MIME_TYPES
   // comment. Legacy .ppt falls through to the generic 'file' kind.
   if (filename.endsWith('.pptx')) return 'presentation'
--- a/frontend/src/components/file/inline-file-preview.tsx
+++ b/frontend/src/components/file/inline-file-preview.tsx
@@ -150,22 +150,38 @@ function InlineImagePreview({
   )
 }
 
-function InlineAudioPreview({
+function InlineMediaPreview({
   source,
   previewUrl,
   filename,
   openLabel,
+  loadErrorText,
   className,
   fileAccess,
+  icon: Icon,
+  bodyClassName,
+  spinnerClassName,
+  renderMedia,
 }: {
   source: InlineFilePreviewSource
   previewUrl: string
   filename: string
   openLabel: string
+  loadErrorText: string
   className?: string
   fileAccess: FileAccessPolicy
+  icon: React.ComponentType<{ className?: string }>
+  bodyClassName: string
+  spinnerClassName: string
+  renderMedia: (resolvedUrl: string, onError: () => void) => React.ReactNode
 }) {
   const resolvedUrl = useResolvedMediaUrl(source, previewUrl, fileAccess)
+  const [failedUrl, setFailedUrl] = useState('')
+  // Terminal failure only: useResolvedMediaUrl already falls back from the
+  // authenticated fetch to the public preview URL, so an error event from
+  // the media element means both paths are exhausted. Keyed by URL so a
+  // later successful re-resolve (e.g. blob after fallback) clears it.
+  const failed = Boolean(resolvedUrl) && failedUrl === resolvedUrl
 
   return (
     <div
@@ -176,7 +192,7 @@ function InlineAudioPreview({
       data-inline-file-preview-wrapper
     >
       <div className="flex items-center gap-2 border-b border-border/50 bg-muted/30 px-3 py-2 text-xs text-muted-foreground">
-        <Volume2 className="h-4 w-4 shrink-0" />
+        <Icon className="h-4 w-4 shrink-0" />
         <span className="min-w-0 flex-1 truncate">{filename}</span>
         {resolvedUrl ? (
           <a
@@ -189,83 +205,88 @@ function InlineAudioPreview({
           </a>
         ) : null}
       </div>
-      <div className="p-3">
-        {resolvedUrl ? (
-          <audio
-            controls
-            preload="metadata"
-            src={resolvedUrl}
-            className="w-full"
-            aria-label={filename}
-            title={filename}
-          />
-        ) : (
-          <div className="flex h-14 items-center justify-center text-muted-foreground">
-            <Loader2 className="h-4 w-4 animate-spin" />
-          </div>
-        )}
-      </div>
+      {failed ? (
+        <div className="p-3 text-xs text-muted-foreground">{loadErrorText}</div>
+      ) : (
+        <div className={bodyClassName}>
+          {resolvedUrl ? (
+            renderMedia(resolvedUrl, () => setFailedUrl(resolvedUrl))
+          ) : (
+            <div
+              className={cn(
+                'flex items-center justify-center text-muted-foreground',
+                spinnerClassName
+              )}
+            >
+              <Loader2 className="h-4 w-4 animate-spin" />
+            </div>
+          )}
+        </div>
+      )}
     </div>
   )
 }
 
-function InlineVideoPreview({
-  source,
-  previewUrl,
-  filename,
-  openLabel,
-  className,
-  fileAccess,
-}: {
+function InlineAudioPreview(props: {
   source: InlineFilePreviewSource
   previewUrl: string
   filename: string
   openLabel: string
+  loadErrorText: string
   className?: string
   fileAccess: FileAccessPolicy
 }) {
-  const resolvedUrl = useResolvedMediaUrl(source, previewUrl, fileAccess)
-
+  const { filename } = props
   return (
-    <div
-      className={cn(
-        'overflow-hidden rounded-md border border-border/50 bg-background',
-        className
+    <InlineMediaPreview
+      {...props}
+      icon={Volume2}
+      bodyClassName="p-3"
+      spinnerClassName="h-14"
+      renderMedia={(resolvedUrl, onError) => (
+        <audio
+          controls
+          preload="metadata"
+          src={resolvedUrl}
+          className="w-full"
+          aria-label={filename}
+          title={filename}
+          onError={onError}
+        />
       )}
-      data-inline-file-preview-wrapper
-    >
-      <div className="flex items-center gap-2 border-b border-border/50 bg-muted/30 px-3 py-2 text-xs text-muted-foreground">
-        <Video className="h-4 w-4 shrink-0" />
-        <span className="min-w-0 flex-1 truncate">{filename}</span>
-        {resolvedUrl ? (
-          <a
-            href={resolvedUrl}
-            target="_blank"
-            rel="noreferrer"
-            className="shrink-0 text-foreground hover:underline"
-          >
-            {openLabel}
-          </a>
-        ) : null}
-      </div>
-      <div className="flex items-center justify-center bg-black/95 p-2">
-        {resolvedUrl ? (
-          <video
-            controls
-            playsInline
-            preload="metadata"
-            src={resolvedUrl}
-            className="max-h-[360px] w-full max-w-full rounded bg-black"
-            aria-label={filename}
-            title={filename}
-          />
-        ) : (
-          <div className="flex h-40 w-full items-center justify-center text-muted-foreground">
-            <Loader2 className="h-4 w-4 animate-spin" />
-          </div>
-        )}
-      </div>
-    </div>
+    />
+  )
+}
+
+function InlineVideoPreview(props: {
+  source: InlineFilePreviewSource
+  previewUrl: string
+  filename: string
+  openLabel: string
+  loadErrorText: string
+  className?: string
+  fileAccess: FileAccessPolicy
+}) {
+  const { filename } = props
+  return (
+    <InlineMediaPreview
+      {...props}
+      icon={Video}
+      bodyClassName="flex items-center justify-center bg-black/95 p-2"
+      spinnerClassName="h-40 w-full"
+      renderMedia={(resolvedUrl, onError) => (
+        <video
+          controls
+          playsInline
+          preload="metadata"
+          src={resolvedUrl}
+          className="max-h-[360px] w-full max-w-full rounded bg-black"
+          aria-label={filename}
+          title={filename}
+          onError={onError}
+        />
+      )}
+    />
   )
 }
 
@@ -457,6 +478,7 @@ export function InlineFilePreview({
         previewUrl={previewUrl}
         filename={filename}
         openLabel={openLabel}
+        loadErrorText={loadErrorText}
         className={className}
         fileAccess={fileAccess}
       />
@@ -470,6 +492,7 @@ export function InlineFilePreview({
         previewUrl={previewUrl}
         filename={filename}
         openLabel={openLabel}
+        loadErrorText={loadErrorText}
         className={className}
         fileAccess={fileAccess}
       />
--- a/frontend/src/components/file/inline-file-preview.tsx
+++ b/frontend/src/components/file/inline-file-preview.tsx
@@ -42,11 +42,12 @@ const DEFAULT_LOAD_ERROR_TEXT = 'Failed to load preview.'
  * but a spinner with no recovery path is worse than attempting the
  * anonymous endpoint.
  *
- * When the policy's authenticated route resolves to the same URL as the
- * public inline route — the public widget/share policy carries its token in
- * the query string — the URL is handed to the media element directly
- * instead: a blob fetch would add no authorization there, and skipping it
- * preserves HTTP range requests for progressive audio/video playback.
+ * A policy that declares ``requiresBlobFetch: false`` (the public
+ * widget/share policy carries its guest token in the query string) hands
+ * the URL to the media element directly instead: a blob fetch would add no
+ * authorization there, and skipping it preserves HTTP range requests for
+ * progressive audio/video playback. Policies that do not declare the
+ * capability get the conservative blob path, which works everywhere.
  */
 function useResolvedMediaUrl(
   source: InlineFilePreviewSource,
@@ -54,7 +55,7 @@ function useResolvedMediaUrl(
   fileAccess: FileAccessPolicy
 ): string {
   const fileId = source.fileId
-  const needsBlobFetch = Boolean(fileId) && fileAccess.previewUrl(fileId!) !== previewUrl
+  const needsBlobFetch = Boolean(fileId) && (fileAccess.requiresBlobFetch ?? true)
   const [resolvedUrl, setResolvedUrl] = useState(needsBlobFetch ? '' : previewUrl)
 
   useEffect(() => {
@@ -173,14 +174,20 @@ function InlineMediaPreview({
   icon: React.ComponentType<{ className?: string }>
   bodyClassName: string
   spinnerClassName: string
-  renderMedia: (resolvedUrl: string, onError: () => void) => React.ReactNode
+  renderMedia: (
+    resolvedUrl: string,
+    media: { onError: () => void; onLoaded: () => void }
+  ) => React.ReactNode
 }) {
   const resolvedUrl = useResolvedMediaUrl(source, previewUrl, fileAccess)
   const [failedUrl, setFailedUrl] = useState('')
-  // Terminal failure only: useResolvedMediaUrl already falls back from the
-  // authenticated fetch to the public preview URL, so an error event from
-  // the media element means both paths are exhausted. Keyed by URL so a
-  // later successful re-resolve (e.g. blob after fallback) clears it.
+  const [loadedUrl, setLoadedUrl] = useState('')
+  // Terminal load failure only: useResolvedMediaUrl already falls back from
+  // the authenticated fetch to the public preview URL, so an error event
+  // from a media element that never loaded data means both paths are
+  // exhausted. Errors after data has loaded (e.g. a mid-playback decode
+  // hiccup) keep the player mounted — the element surfaces those itself.
+  // Keyed by URL so a later re-resolve clears the failure.
   const failed = Boolean(resolvedUrl) && failedUrl === resolvedUrl
 
   return (
@@ -210,7 +217,12 @@ function InlineMediaPreview({
       ) : (
         <div className={bodyClassName}>
           {resolvedUrl ? (
-            renderMedia(resolvedUrl, () => setFailedUrl(resolvedUrl))
+            renderMedia(resolvedUrl, {
+              onError: () => {
+                if (loadedUrl !== resolvedUrl) setFailedUrl(resolvedUrl)
+              },
+              onLoaded: () => setLoadedUrl(resolvedUrl),
+            })
           ) : (
             <div
               className={cn(
@@ -227,7 +239,7 @@ function InlineMediaPreview({
   )
 }
 
-function InlineAudioPreview(props: {
+type MediaWrapperProps = {
   source: InlineFilePreviewSource
   previewUrl: string
   filename: string
@@ -235,7 +247,9 @@ function InlineAudioPreview(props: {
   loadErrorText: string
   className?: string
   fileAccess: FileAccessPolicy
-}) {
+}
+
+function InlineAudioPreview(props: MediaWrapperProps) {
   const { filename } = props
   return (
     <InlineMediaPreview
@@ -243,7 +257,7 @@ function InlineAudioPreview(props: {
       icon={Volume2}
       bodyClassName="p-3"
       spinnerClassName="h-14"
-      renderMedia={(resolvedUrl, onError) => (
+      renderMedia={(resolvedUrl, { onError, onLoaded }) => (
         <audio
           controls
           preload="metadata"
@@ -252,21 +266,14 @@ function InlineAudioPreview(props: {
           aria-label={filename}
           title={filename}
           onError={onError}
+          onLoadedData={onLoaded}
         />
       )}
     />
   )
 }
 
-function InlineVideoPreview(props: {
-  source: InlineFilePreviewSource
-  previewUrl: string
-  filename: string
-  openLabel: string
-  loadErrorText: string
-  className?: string
-  fileAccess: FileAccessPolicy
-}) {
+function InlineVideoPreview(props: MediaWrapperProps) {
   const { filename } = props
   return (
     <InlineMediaPreview
@@ -274,7 +281,7 @@ function InlineVideoPreview(props: {
       icon={Video}
       bodyClassName="flex items-center justify-center bg-black/95 p-2"
       spinnerClassName="h-40 w-full"
-      renderMedia={(resolvedUrl, onError) => (
+      renderMedia={(resolvedUrl, { onError, onLoaded }) => (
         <video
           controls
           playsInline
@@ -284,6 +291,7 @@ function InlineVideoPreview(props: {
           aria-label={filename}
           title={filename}
           onError={onError}
+          onLoadedData={onLoaded}
         />
       )}
     />
--- a/frontend/src/components/file/inline-file-preview.tsx
+++ b/frontend/src/components/file/inline-file-preview.tsx
@@ -32,6 +32,74 @@ const fileNameFromSource = (source: InlineFilePreviewSource) =>
 const DEFAULT_OPEN_LABEL = 'Open'
 const DEFAULT_LOAD_ERROR_TEXT = 'Failed to load preview.'
 
+/**
+ * Resolve the URL a media element (<img>/<audio>/<video>) can load.
+ *
+ * The default in-app policy's authenticated preview route needs a Bearer
+ * header that media elements cannot send, so managed files are fetched into
+ * a blob object URL first. If that fetch fails, the public preview URL is a
+ * last-resort fallback: on Agent Builder surfaces it may also require auth,
+ * but a spinner with no recovery path is worse than attempting the
+ * anonymous endpoint.
+ *
+ * When the policy's authenticated route resolves to the same URL as the
+ * public inline route — the public widget/share policy carries its token in
+ * the query string — the URL is handed to the media element directly
+ * instead: a blob fetch would add no authorization there, and skipping it
+ * preserves HTTP range requests for progressive audio/video playback.
+ */
+function useResolvedMediaUrl(
+  source: InlineFilePreviewSource,
+  previewUrl: string,
+  fileAccess: FileAccessPolicy
+): string {
+  const fileId = source.fileId
+  const needsBlobFetch = Boolean(fileId) && fileAccess.previewUrl(fileId!) !== previewUrl
+  const [resolvedUrl, setResolvedUrl] = useState(needsBlobFetch ? '' : previewUrl)
+
+  useEffect(() => {
+    let objectUrl: string | null = null
+    let isCancelled = false
+
+    setResolvedUrl(needsBlobFetch ? '' : previewUrl)
+
+    const loadAuthenticatedMedia = async () => {
+      if (!needsBlobFetch || !fileId) return
+      try {
+        const response = await fileAccess.request(fileAccess.previewUrl(fileId), {
+          cache: 'no-cache',
+          headers: {
+            'Cache-Control': 'no-cache',
+            Pragma: 'no-cache',
+          },
+        })
+        if (isCancelled) return
+        if (!response.ok) {
+          setResolvedUrl(previewUrl)
+          return
+        }
+        const blob = await response.blob()
+        if (isCancelled) return
+        objectUrl = URL.createObjectURL(blob)
+        setResolvedUrl(objectUrl)
+      } catch {
+        if (!isCancelled) {
+          setResolvedUrl(previewUrl)
+        }
+      }
+    }
+
+    void loadAuthenticatedMedia()
+
+    return () => {
+      isCancelled = true
+      if (objectUrl) URL.revokeObjectURL(objectUrl)
+    }
+  }, [fileAccess, fileId, needsBlobFetch, previewUrl])
+
+  return resolvedUrl
+}
+
 function InlineImagePreview({
   source,
   previewUrl,
@@ -47,56 +115,7 @@ function InlineImagePreview({
   onFileClick?: (filePath: string, fileName: string) => void
   fileAccess: FileAccessPolicy
 }) {
-  const shouldFallback = Boolean(source.fileId)
-  const [resolvedUrl, setResolvedUrl] = useState(shouldFallback ? '' : previewUrl)
-
-  useEffect(() => {
-    let objectUrl: string | null = null
-    let isCancelled = false
-
-    setResolvedUrl(shouldFallback ? '' : previewUrl)
-
-    const runFallback = async () => {
-      if (!shouldFallback) return
-      try {
-        const response = await fileAccess.request(
-          fileAccess.previewUrl(source.fileId!),
-          {
-            cache: 'no-cache',
-            headers: {
-              'Cache-Control': 'no-cache',
-              Pragma: 'no-cache',
-            },
-          }
-        )
-        if (isCancelled) return
-        if (!response.ok) {
-          // Last-resort fallback: try the public preview URL when the
-          // authenticated route fails. On Agent Builder surfaces that
-          // route may also require auth, but a spinner with no recovery
-          // path is worse than attempting the anonymous endpoint.
-          setResolvedUrl(previewUrl)
-          return
-        }
-        const blob = await response.blob()
-        if (isCancelled) return
-        objectUrl = URL.createObjectURL(blob)
-        setResolvedUrl(objectUrl)
-      } catch {
-        if (!isCancelled) {
-          // See comment above: public preview is best-effort after auth errors.
-          setResolvedUrl(previewUrl)
-        }
-      }
-    }
-
-    void runFallback()
-
-    return () => {
-      isCancelled = true
-      if (objectUrl) URL.revokeObjectURL(objectUrl)
-    }
-  }, [fileAccess, previewUrl, shouldFallback, source.fileId])
+  const resolvedUrl = useResolvedMediaUrl(source, previewUrl, fileAccess)
 
   const handleClick = (event: React.MouseEvent<HTMLImageElement>) => {
     if (!onFileClick || !source.fileId) return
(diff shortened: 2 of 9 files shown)


src/xagent/web/services/file_reference_output_service.py
```diff
@@ -20,6 +21,30 @@
...
+# ``artifact_type_for_filename`` only knows image/video/office extensions
+# and never returns "audio", so audio must be matched by extension here.
+# This set mirrors the frontend's audio detection (see
+# ``getInlineFilePreviewKind`` in ``inline-file-preview-utils.ts``) — keep
+# the two in sync.
+_AUDIO_EXTENSIONS = {".mp3", ".wav", ".ogg", ".opus", ".flac", ".m4a", ".aac"}
+
+
+def _is_inline_preview_media(filename: str, suffix: str) -> bool:
+    return artifact_type_for_filename(filename) in _INLINE_PREVIEW_LINK_TYPES or (
```
rogercloud: **Major: `.m4v` is missing on the backend, so this fix silently no-ops for it and #1191 reproduces exactly.**

The frontend regex at `frontend/src/components/file/inline-file-preview-utils.ts:126` is `/\.(mp4|m4v|mov|webm|mpeg|mpg)$/` — it includes `.m4v`. The backend `artifact_type_for_filename` (`src/xagent/core/tools/artifacts.py:59`) checks `{".mp4", ".mov", ".mpeg", ".mpg", ".webm"}` and omits `.m4v`; it's also absent from `GENERATED_ARTIFACT_EXTENSIONS` at the top of that file. Empirically, `artifact_type_for_filename('clip.m4v')` returns `'file'`, so `_is_inline_preview_media` returns `False` here and the label is never restored — leaving the frontend with a prose label it can't classify, i.e. the original bug.

This is reachable in production, not hypothetical: `src/xagent/core/tools/core/video_tool.py:521-524` derives the extension from the remote provider URL's path via `os.path.splitext` and only defaults to `.mp4` when the URL has no extension at all, so a provider URL ending in `.m4v` yields a real `generated_video_<hex>.m4v`.

Suggest adding `.m4v` to both the video set at `artifacts.py:59` and `GENERATED_ARTIFACT_EXTENSIONS`. Worth doing as part of a test that pins the two extension lists to agree (see the test-coverage comment) — this is exactly the class of drift such a test would have caught.

src/xagent/web/services/file_reference_output_service.py
```diff
@@ -116,6 +141,14 @@ def replacement(match: re.Match[str]) -> str:
                 record.file_id,
             )
             return label
-        return f"{prefix}[{label}]({canonical_ref})"
+
+        display_label = label
+        filename = str(record.filename or "")
+        suffix = Path(filename).suffix.casefold()
+        if not prefix and suffix and _is_inline_preview_media(filename, suffix):
+            if not label.strip().casefold().endswith(suffix):
```
rogercloud: **Minor/moderate: the restored filename is substituted unescaped, which can break the link outright.**

`display_label = filename` flows straight into `f"{prefix}[{display_label}]({canonical_ref})"`. Before this PR the substituted value was always `label`, which the parsing regex guarantees is `]`-free via `(?P<label>[^\]]*)` — so this is a genuine new regression rather than a pre-existing hazard.

`UploadedFile.filename` (`src/xagent/web/models/uploaded_file.py:44`) has no charset constraint, and the upload path only strips directory components with `Path(...).name` — it doesn't sanitize special characters. A user-uploaded `clip ].mp4` therefore reaches this line intact, and per CommonMark the resulting construct fails to parse as a link at all and renders as literal text: the file reference is lost from the message entirely, which is strictly worse than the pre-PR download link.

Tool-generated filenames are separately sanitized to alnum/`-`/`_`/`.` (`video_tool.py:511-512,525-526`), so this only affects re-referenced user-uploaded media — but that's a real path. Suggest escaping `[`, `]` and `\` in `display_label` before interpolation (or falling back to the original `label` when the filename contains characters that can't be safely inlined).

src/xagent/web/services/file_reference_output_service.py
```diff
@@ -20,6 +21,30 @@
...
+_INLINE_PREVIEW_LINK_TYPES = {"video"}
+# ``artifact_type_for_filename`` only knows image/video/office extensions
+# and never returns "audio", so audio must be matched by extension here.
+# This set mirrors the frontend's audio detection (see
+# ``getInlineFilePreviewKind`` in ``inline-file-preview-utils.ts``) — keep
+# the two in sync.
+_AUDIO_EXTENSIONS = {".mp3", ".wav", ".ogg", ".opus", ".flac", ".m4a", ".aac"}
+
+
+def _is_inline_preview_media(filename: str, suffix: str) -> bool:
```
rogercloud: **Minor (shrink): the `suffix` parameter is redundant.**

There's exactly one call site (repo-wide grep), and it computes `suffix` as precisely `Path(filename).suffix.casefold()` on the line immediately before calling (L147-148). Deriving it inside the function removes a parameter and eliminates the possibility of the two arguments disagreeing. The recomputation cost is one `Path` construction per matched link — not a hot loop.

```python
def _is_inline_preview_media(filename: str) -> bool:
    suffix = Path(filename).suffix.casefold()
    ...
```

Same neighbourhood: `_INLINE_PREVIEW_LINK_TYPES = {"video"}` is a one-element set with a single membership check, and the adjacent comment frames the narrow scope as deliberate rather than temporary — `artifact_type_for_filename(filename) == "video"` reads better and drops the constant.

frontend/src/components/file/inline-file-preview.tsx
```diff
@@ -234,6 +234,110 @@ function InlineAudioPreview({
   )
 }
 
+function InlineVideoPreview({
```
rogercloud: **Minor (suggestion, not blocking): blob-only loading gives up range requests, which matters more for video than it did for audio.**

The component fetches the entire file into a blob before playback, so there's no progressive playback, no seek-before-download, and the whole file sits in memory. For the *default authenticated* in-app-chat policy this is genuinely necessary — `fileAccess.previewUrl` hits an endpoint requiring a Bearer token that `<video src>` can't carry — so the merged approach is correct as a baseline.

The optimization applies specifically to the public/widget path: under `createPublicFileAccessPolicy`, `previewUrl` resolves to the tokenless `/public/preview/{file_id}` route, which has no auth dependency and returns a `FileResponse` (range requests supported via Starlette's default handling). There, setting `previewUrl` directly as `<video src>` and keeping blob-fetch only as an error fallback would be strictly better. Note the fallback still has to stay — `_validate_public_task_file_access` can 403.

Reasonable as a follow-up; flagging so it isn't lost. (Related: this effect is now the third verbatim copy across image/audio/video — see the simplification section on extracting a shared hook.)

tests/web/services/test_file_reference_output_service.py
```diff
@@ -150,6 +152,76 @@ def test_reconcile_unlinks_record_with_invalid_file_id():
...
+            user_id=int(user.id),
+            content="[下载音频（MP3）](file:real-id)",
+        )
+
+        assert content == "[generated_podcast.mp3](file:real-id)"
+    finally:
+        db.close()
+
+
+def test_reconcile_keeps_custom_label_for_plain_download_file():
```
rogercloud, line 255: **Minor: test coverage gaps around the new label-restore logic.**

- This test (`test_reconcile_keeps_custom_label_for_plain_download_file`) uses `notes.txt`, which `artifact_type_for_filename` classifies as generic `"file"` — not `document`/`presentation`/`spreadsheet`. The code comment's explicit claim that office types are *deliberately* excluded from label rewriting is therefore untested: this test would still pass if `"document"` were accidentally added to the inline-preview set. Add a `.docx`/`.pptx` case that asserts the custom label survives.
- No case-insensitive video-extension case (e.g. `CLIP.MP4`), despite an established precedent for exactly this in `tests/core/tools/test_artifacts_observation.py` (the `DECK.PPTX` case with a `# case-insensitive` comment). The code is correct here; only the test is missing.
- Nothing pins the frontend and backend video-extension lists to agree with each other. That's the precise gap that let the `.m4v` mismatch through — a small test asserting parity (or a single shared source of truth) would close it.
- No coverage for a label that already ends in a valid media suffix but names a *different* file than the record's real filename (e.g. label `my_clip.mp4` on a record whose filename is `generated_video.mp4`). Current behavior preserves the mismatched label as-is, which is arguably the right call — worth locking in with a test either way.

rogercloud (changes requested):
## Summary

This PR fixes #1191 (generated videos rendering as bare file names) on both ends of the pipeline. On the frontend it introduces a first-class `'video'` preview kind — mime/extension/type detection in `inline-file-preview-utils.ts` plus a new `InlineVideoPreview` component that mirrors the existing `InlineAudioPreview` — so video references no longer fall through to the generic download-link `file` kind. On the backend, `reconcile_assistant_file_references` now restores the real filename as the markdown link label when the model's prose label has dropped the extension, which is necessary because the link target is an opaque `file:<uuid>` and the frontend relies on the label for kind detection. The label restore is deliberately scoped to video/audio and does not touch office documents, which is the right call — flipping those into eager-fetching preview boxes would be a regression of its own.

Worth noting up front: the PR's second commit already fixed a real gap that the automated review flagged earlier (audio extensions not recognized by `artifact_type_for_filename`, which silently broke the audio half of the label-restore fix). The `_AUDIO_EXTENSIONS` fallback resolves that and the audio side now looks solid. The exactly parallel gap on the video side — `.m4v` — was not caught by that same pass and is still present (finding 1 below).

## Round 0 — approach verdict: acceptable with reservations

The frontend half is straightforwardly the right fix and slots into the existing kind-dispatch abstraction without distorting it. The backend half solves a real problem, but the chosen mechanism has one alternative worth thinking about, plus two concrete correctness issues called out inline.

- **Label mutation vs. a markdown `title` attribute.** The rewrite is narrow (only video/audio, only when the label doesn't already carry the right extension), so the blast radius is small. But it runs *before* persistence — `persist_assistant_message` / `persist_assistant_message_no_commit` reconcile before building `transcript_content` — so for the cases it fires on, the model's original label text is unrecoverably lost from the stored transcript. `frontend/src/components/ui/markdown-renderer.tsx:329` already does `const fileName = title || linkText || fileNameFromPath`, i.e. `title` outranks the label. Emitting `[model's label](file:id "real_filename.mp4")` would fix detection *and* preserve the model's prose. Not a blocker, and not free either: `_MARKDOWN_FILE_REFERENCE_RE` (`src/xagent/web/services/file_reference_output_service.py:20-22`) uses `[^)\s]+` for the target and cannot parse title-bearing links at all, so this would need the regex extended first. Filing as food for thought for a follow-up.
  - Cosmetic side-effect of the current approach: `_reconcile_streamed_final_answer` (`src/xagent/web/api/websocket.py`) only runs on `final_answer_end`, not on `final_answer_delta` — so during streaming the user sees the model's prose label and it visibly snaps to the real filename at stream end. Purely visual, mentioning for completeness.
  - Low-priority note: if the model ever emitted a link that already carries a `title` clause, that title would take precedence over the restored label and defeat the fix. No prompt or tool description in the codebase (`file_ref.py`, `context_builder.py`, `artifacts.py`'s `markdown_reference_for_artifact`) instructs or has been observed to produce title-bearing links, so this is theoretical today.
- **Asymmetric scoping on images.** The label-restore only covers video/audio. Plain (non-`!`) image links could in principle degrade the same way, but in practice `MarkdownImage` defaults `previewKind = preview?.previewKind ?? 'image'` for `![...]` embed syntax, and images are always emitted with `!` by `markdown_reference_for_artifact` — which is precisely why the bug showed up for video/audio (emitted as plain links) and not for images. The gap only opens if the model rewrites embed syntax into plain-link syntax, which is a much larger edit than rephrasing a label. Forward-looking note only; no change requested.

## Line-level findings

See inline comments. Ordered by severity:

1. **Major** — `.m4v` is recognized by the frontend but not the backend, so the label-restore no-ops and #1191 reproduces exactly for that extension. Reachable in production via `video_tool.py`'s provider-URL extension derivation.
2. **Minor/moderate** — the restored filename is substituted into the markdown link unescaped; a user-uploaded file whose name contains `]` now breaks the link entirely (worse than the pre-PR download-link fallback). This is a new regression, since the previously-substituted `label` was structurally `]`-free by construction.
3. **Minor** — `_is_inline_preview_media` takes a `suffix` parameter its single caller always derives from `filename` one line earlier.
4. **Minor** — `InlineVideoPreview` always blob-fetches the whole file before playback, forgoing range requests / progressive playback. Necessary for the authenticated in-app path (a Bearer token can't ride on `<video src>`), but avoidable on the public/widget path.
5. **Minor** — test coverage gaps around the office-type exclusion, case-insensitive extensions, and frontend/backend extension-list agreement.

- `frontend/src/components/file/inline-file-preview.tsx` L35/L134/L237: **shrink** — `InlineImagePreview`, `InlineAudioPreview` and the new `InlineVideoPreview` each carry a byte-identical authenticated-blob-fetch-with-public-fallback effect (same state shape, same `fileAccess.request` with no-cache headers, same blob→objectURL conversion, same cleanup), differing only in icon, media tag and container styling. Extract a shared `useResolvedMediaUrl(source, previewUrl, fileAccess)` hook. Third occurrence — this is the point where it earns extraction.
- `src/xagent/web/services/file_reference_output_service.py` L43: **shrink** — redundant `suffix` parameter, derive internally (see inline comment; not duplicating here).
- `src/xagent/web/services/file_reference_output_service.py` L39: **yagni** — `_INLINE_PREVIEW_LINK_TYPES = {"video"}` is a single-element set with exactly one membership check, and the surrounding comment frames the narrow scope as deliberate rather than provisional. Replace with `artifact_type_for_filename(filename) == "video"` and delete the set.

net: ~-45 lines possible.

## Testing

- I independently ran `tests/web/services/test_file_reference_output_service.py` in a fresh virtualenv (`pip install -e .`) — **8 passed**.
- **Review limitation, not a defect in the PR:** I could not run the frontend suite (no `node_modules` in my checkout), nor could I independently verify the PR's claimed clean `tsc --noEmit` / `mypy` / `ruff` / `eslint` runs or the 98 passing frontend tests. Those claims are unverified by me rather than disputed.
- No findings on: `InlineVideoPreview`'s resource lifecycle (mechanically identical to the already-shipped audio effect — correct cleanup, no leaks or races); case handling in the shipped code (frontend lowercases, backend `.casefold()`s, all four paths correct); regression risk to the existing image/audio/office paths (all new branches are additive and correctly ordered); idempotency of the label rewrite on repeated reconcile passes (second pass short-circuits on the already-suffixed label); security (no new endpoints or auth surface; video reuses the audio `fileAccess`/public-preview mechanism; no injection surface beyond finding 2). The trace-view video-artifact fix claimed in the PR body checks out — `video_tool.py` emits `type: "video"`, `TraceEventRenderer.tsx` passes it through, and the new test exercises that dispatch branch.

src/xagent/web/services/file_reference_output_service.py
```diff
@@ -20,6 +21,45 @@
...
+# file like ``clip ].mp4`` must never be substituted into a ``[label](...)``
+# span unescaped: an embedded ``]`` (or ``[``/``\``) terminates the link
+# early and the file reference is lost entirely — worse than the plain
+# download link this rewrite is trying to improve on. Escaping instead of
+# skipping would work for a single pass, but ``_MARKDOWN_FILE_REFERENCE_RE``
+# has no notion of backslash-escapes, so a later reconcile pass over
+# already-escaped content would mis-locate the label boundary at the
+# escaped bracket. Skipping the rewrite for these filenames is the safe,
+# idempotent choice; the link still falls back to a plain download label.
+_UNSAFE_LABEL_CHARS = frozenset("[]\\")
```
rogercloud: `_UNSAFE_LABEL_CHARS` covers `[`, `]` and `\`, but not `\n`/`\r`. A filename like `"a\n\nb.mp4"` passes this check and is substituted into `[a\n\nb.mp4](file:id)` — CommonMark splits that into two paragraphs at the blank line *before* the inline link is parsed, so the link is destroyed exactly the way commit `9b579c11` set out to prevent, just via a different character class. This is reachable: `UploadedFile.filename` is not sanitized for control characters on either the upload path (`Path(uploaded.filename).name` does not strip them) or the tool-generated path, and the DB column has no format constraint. Suggest widening this to reject control characters (e.g. `\n`, `\r`, `\t`) and adding a regression test alongside `test_reconcile_skips_label_rewrite_for_filename_with_unsafe_chars`, which currently only exercises `]`.

frontend/src/components/file/inline-file-preview.tsx
```diff
@@ -428,6 +463,19 @@ export function InlineFilePreview({
     )
   }
 
+  if (kind === 'video') {
```
rogercloud, line 496: Neither `InlineVideoPreview` nor `InlineAudioPreview` receives `loadErrorText`, unlike `InlineOfficeContent`. `useResolvedMediaUrl` falls back to the public preview URL when the authenticated fetch fails, but if that fallback also fails there is no error path at all — the user is left with an empty/broken `<video>` (or `<audio>`) element and no message. Consider threading `loadErrorText` through both media components and rendering it on the terminal-failure branch.

frontend/src/components/file/inline-file-preview.tsx
```diff
@@ -32,49 +32,49 @@ const fileNameFromSource = (source: InlineFilePreviewSource) =>
...
+function useResolvedMediaUrl(
+  source: InlineFilePreviewSource,
+  previewUrl: string,
   fileAccess: FileAccessPolicy
-}) {
-  const shouldFallback = Boolean(source.fileId)
-  const [resolvedUrl, setResolvedUrl] = useState(shouldFallback ? '' : previewUrl)
+): string {
+  const fileId = source.fileId
+  const needsBlobFetch = Boolean(fileId) && fileAccess.previewUrl(fileId!) !== previewUrl
```
rogercloud: This gate is where the undeclared behavior change lives. Previously both `InlineImagePreview` and `InlineAudioPreview` used `shouldFallback = Boolean(source.fileId)` and *always* blob-fetched a managed file. Now `needsBlobFetch` is false whenever `fileAccess.previewUrl(fileId) === previewUrl`, which is exactly the public/share policy — so under `createPublicFileAccessPolicy`, image and audio previews (not only the new video) assign the token-bearing URL straight to `src`. The token is a scoped guest query-param token and was already transmitted, but it now also lands in DOM `src`, page HTML, "copy image address", the media cache, and referrers. The renamed test with inverted assertions shows this is intentional; it just is not mentioned in the PR description. Could you confirm this expansion beyond video is wanted, and call it out in the description?

frontend/src/components/file/inline-file-preview.tsx
```diff
@@ -234,6 +209,66 @@ function InlineAudioPreview({
   )
 }
 
+function InlineVideoPreview({
```
rogercloud: `InlineVideoPreview` repeats ~25-30 lines of `InlineAudioPreview` almost verbatim — same wrapper, header, truncated filename, open link, and spinner; only the icon (`Video` vs `Volume2`), the media element, and a couple of classNames differ. The "share media blob-fetch logic" commit extracted the hook but left this JSX duplicated. Extracting an `InlineMediaPreview` parametrized by `icon`, `mediaClassName`, and a `renderMedia(resolvedUrl)` callback would remove roughly 40 lines with no loss of type safety, and would also give the missing error state (see the other comment) a single place to live.

frontend/src/components/file/inline-file-preview-utils.ts
```diff
@@ -60,7 +62,8 @@ export const INLINE_FILE_PREVIEW_MIME_BY_KIND: Partial<
 export const getInlineFilePreviewMimeType = (
   kind: InlineFilePreviewKind
 ): string | undefined => {
-  if (kind === 'file' || kind === 'image' || kind === 'audio') return undefined
+  if (kind === 'file' || kind === 'image' || kind === 'audio' || kind === 'video')
```
rogercloud: This guard is dead code: `INLINE_FILE_PREVIEW_MIME_BY_KIND` only has `presentation`/`document`/`spreadsheet` keys, so the lookup already returns `undefined` for `file`, `image`, `audio` and `video`. The whole function body can be `return INLINE_FILE_PREVIEW_MIME_BY_KIND[kind]`, which also means the next media kind added does not need a matching edit here.

tests/web/services/test_file_reference_output_service.py
```diff
@@ -150,6 +153,221 @@ def test_reconcile_unlinks_record_with_invalid_file_id():
...
+            user_id=int(user.id),
+            content="[下载视频（MP4）](file:real-id)",
+        )
+
+        assert content == "[下载视频（MP4）](file:real-id)"
+    finally:
+        db.close()
+
+
+def test_video_extension_detection_matches_frontend_set():
```
rogercloud, line 463: `test_video_extension_detection_matches_frontend_set` pins video parity against the frontend, which is the right pattern — but there is no equivalent test pinning `_AUDIO_EXTENSIONS` against the frontend audio regex in `inline-file-preview-utils.ts`. Audio is the category that actually drifted historically (commit `a2bc9825` fixed a missed audio extension), so it is the one most in need of the guard. Worth adding an audio counterpart here.

rogercloud (changes requested):
## Summary

This PR makes generated videos playable inline in chat instead of degrading to plain download links. On the frontend it adds a `video` kind to the inline-preview classifier and a new `InlineVideoPreview` component modeled on `InlineAudioPreview`; a follow-up commit extracts the shared blob-fetch logic into a `useResolvedMediaUrl` hook. On the backend, `reconcile_assistant_file_references` now overwrites the model-authored link label with the real filename for video/audio references whose label has lost its extension, so the frontend's extension-based type detection still works after the model rewrites `[label](file:<uuid>)` into prose. Two of the four commits are gap-fixes for extension sets that were initially incomplete (audio extensions, then `.m4v`).

The frontend classification work is correct and lands the user-visible fix. Three design-level reservations, none of which own a single line so they are raised here rather than inline:

- **The default in-app video path fully downloads before playback.** Under the default (non-public) `FileAccessPolicy`, `useResolvedMediaUrl` always takes the blob branch: the entire file is `fetch()`ed into memory and only then handed to `<video>` as a `blob:` URL — no Range requests, no seek-before-download, whole file resident in memory. Generated videos have no size cap (`_download_video` in `src/xagent/core/tools/core/video_tool.py` streams chunks with no size accounting; `MAX_FILE_SIZE` in `src/xagent/web/config.py` only gates HTTP/KB uploads and is never consulted by the video tool). A Range-capable path already exists and is already used in this very file for the public/share policy — `/api/files/public/preview/{id}` returns a Starlette `FileResponse`. So for large generated videos this is a partial fix: the user trades a broken filename link for a spinner that spans the whole download, rather than getting progressive playback. Worth a follow-up decision on whether the authenticated preview route can be made Range-capable and used directly.
- **Undeclared scope expansion on public/share surfaces.** The refactor changed the blob-fetch gate from "any managed file" to "only when the authenticated URL differs from the inline preview URL". Under `createPublicFileAccessPolicy` the two are equal, so **image and audio** previews (not just the new video) now skip the blob fetch and assign the token-bearing public URL directly as `src`. The existing test was renamed and its assertions inverted (from asserting two `credentials: 'omit'` fetches, to asserting no fetch and a raw tokened `src`), confirming this is deliberate — but it is not mentioned in the PR description, which describes only the video addition. The token is a scoped guest query-param token and was already sent over the wire, but it now also lands in DOM `src`, page HTML, "copy image address", the browser media cache, and referrers of requests the media element issues. Not necessarily a blocker; it deserves an explicit maintainer decision rather than arriving as a side effect of a video PR.
- **The label rewrite is a destructive workaround on a lossy channel.** `reconcile_assistant_file_references` mutates the *persisted* assistant message (this content is what reaches the DB, not a render-time transform), permanently discarding the model's original label. A localized, human-readable label (e.g. "Download video (MP4)" in the user's own language) becomes `generated_video_a1b2c3.mp4` with no way back — an unrecoverable regression for non-English conversations. The obvious alternative, carrying the real filename in the link title (`[label](file:id "real.mp4")`, which `MarkdownLink` already prefers), is *not* a drop-in: `_MARKDOWN_FILE_REFERENCE_RE` cannot parse title syntax at all today, and it would also require the model to emit titles reliably — the same LLM-reliability problem relocated. So the current approach is a defensible tactical choice; recording the direction is enough for this PR, but the persistence-level destructiveness should be a conscious call, not an accident.

## Findings

### Medium

1. **`_UNSAFE_LABEL_CHARS` omits `\n`/`\r`** — `src/xagent/web/services/file_reference_output_service.py:61`. A filename like `"a\n\nb.mp4"` passes the unsafe-char check, is substituted into `[a\n\nb.mp4](file:id)`, and CommonMark splits it into two paragraphs before the inline link is ever parsed — breaking the link exactly the way commit `9b579c11` intended to prevent, via a different character class. `UploadedFile.filename` is unsanitized on both the upload path and the tool-generated path, and the DB column has no format constraint, so this is reachable. See inline comment.
2. **No error state when both media fetches fail** — `frontend/src/components/file/inline-file-preview.tsx:212` (video) and `:153` (audio). `useResolvedMediaUrl` falls back to the public URL on fetch failure; if that also fails, nothing surfaces. `loadErrorText` exists and is wired into the office-document path but is never passed to either media component. See inline comment.
3. **Test gaps that track live risk**, in rough priority order:
   - No backend test for a newline-containing (or excessively long) filename becoming a label — finding 1 is a confirmed live bug such a test would have caught.
   - No test for video's public-preview fallback (audio has one) and none for video blob-URL revocation on unmount (image has one). These are not pure duplication: each kind wires the shared hook's `resolvedUrl` to its own DOM element, so the wiring can break per-component even with the hook correct.
   - No test pinning `_AUDIO_EXTENSIONS` against the frontend audio regex, though video parity *is* pinned by `test_video_extension_detection_matches_frontend_set` — audio is precisely the category that already had a historical sync gap.
   - No test for the empty-label case `[](file:id)` on either side.

### Low / nits

4. **Media extension and MIME knowledge is duplicated across four unsynchronized places** — `src/xagent/core/tools/artifacts.py`, `_AUDIO_EXTENSIONS` in the file-reference service, and the frontend classifier regexes. This PR's own history is the evidence: `a2bc9825` fixed a missed audio extension and `9b579c11` a missed `.m4v`. The parity test is the right mitigation; it just needs an audio counterpart (see above). Note: replacing `_AUDIO_EXTENSIONS` with stdlib `mimetypes.guess_type` was checked and is **not** viable — on Python 3.11/3.12 (the declared minimum) `.ogg`, `.flac` and `.m4a` do not resolve to `audio/*`. The hand-rolled set is a justified reliability guard, keep it.
5. **`GENERATED_ARTIFACT_EXTENSIONS` has no audio extensions** — `src/xagent/core/tools/artifacts.py:13`. Mostly cosmetic: `music_tool.py`/`sound_effect_tool.py` register via `workspace.auto_register_files()` and never consult this set. The only real gap is audio produced through `python_executor`/`javascript_executor`, which degrades gracefully to a generic `file` download link. Low priority.
6. **Dead branch in `getInlineFilePreviewMimeType`** — `frontend/src/components/file/inline-file-preview-utils.ts:65`. See inline comment.
7. **`preload="metadata"` is inert under the default policy** — `inline-file-preview.tsx:256`. With a `blob:` src there is nothing left to preload; under the public/widget policy the src is a real streamable URL and the attribute does mean something. Not worth changing, just noting the inconsistency (it resolves itself if the streaming concern above is ever addressed).

- `L212: shrink` `InlineVideoPreview` duplicates ~25-30 lines of `InlineAudioPreview`'s wrapper/header/open-link/spinner JSX near-verbatim; only the icon (`Video` vs `Volume2`), the media tag, and a few classNames differ. The refactor commit extracted the hook but not this JSX. Extract a shared `InlineMediaPreview` parametrized by icon, media className, and a `renderMedia` callback.
- `L65: inline` the `kind === 'file' | 'image' | 'audio' | 'video'` guard in `getInlineFilePreviewMimeType` is redundant — `INLINE_FILE_PREVIEW_MIME_BY_KIND` has no such keys, so the bare lookup already yields `undefined`. Replace with `return INLINE_FILE_PREVIEW_MIME_BY_KIND[kind]`.

net: -40 lines possible

## Verified clean

Explicitly checked, no issues found: **security/XSS/injection** (no `dangerouslySetInnerHTML`; `safeUrlTransform` restricts schemes; filename-derived labels render as text nodes; no path traversal in the preview routes); **idempotency** (the label rewrite is a fixed point across repeated reconciliation — checked at the stream-end, finalize, and history-replay call sites); **blob resource lifecycle** (object-URL revocation is correctly gated by the cancellation flag, no leak on unmount or on rapid src changes); **non-media regression from the label rewrite** (office/document types correctly excluded via `_is_inline_preview_media`, so their compact links do not flip into heavy preview boxes); **markdown block-layout interaction**; and the **trace-execution-view fix** claimed in the description, which is genuinely fixed.

## Testing note

Tests were **not** executed in this review environment — there is no Python environment with `lancedb` installed and no frontend `node_modules`. The PR's claim of "98 passed / 7 passed, `tsc`/`mypy`/`ruff`/`eslint` clean" was **not** independently re-run; all findings above are from code reading and targeted static verification only.

yiboyasss:
Thanks for the thorough third round. Status of each item as of a79bb7cf:

**Fixed**

- **Medium 1 (control characters in `_UNSAFE_LABEL_CHARS`)** — widened to a regex covering `[`, `]`, `\`, all C0 control characters and DEL, so newline-bearing filenames skip the rewrite; regression test added (`test_reconcile_skips_label_rewrite_for_filename_with_control_chars`).
- **Medium 2 (no error state on terminal media failure)** — `loadErrorText` is now threaded into both players; an error event from the media element (fired only after the hook has already fallen back to the public preview URL) renders the error text while keeping the filename header and Open link. Covered by a new test.
- **Medium 3 (test gaps)** — all four added: control-char filename (backend), video public-preview fallback + blob revocation on unmount (frontend), `_AUDIO_EXTENSIONS` parity pin against the frontend regex (backend), and the empty-label `[](file:id)` rewrite (backend).
- **Low 6 / simplification 2 (`getInlineFilePreviewMimeType` dead branch)** — reduced to a bare table lookup.
- **Simplification 1 (`InlineMediaPreview`)** — extracted; audio/video now share the wrapper JSX, and the new error state lives there once.

**Confirmed, documented, not changed**

- **Scope expansion on public/share surfaces** — deliberate, now called out in the PR description with the reasoning: the query-string token exists precisely because media elements cannot send an `Authorization` header (that policy was designed for direct loads), and tokened URLs already appear in DOM `href`s on the same surfaces via `inlineDownloadUrl` (generic file links, office Open links) — this extends an existing exposure pattern to media `src` rather than creating a new one. If tightening is ever wanted, the right direction is product-level (short-lived per-file signed URLs for the public routes).

**Not fixed, with reasons**

- **Default in-app path downloads fully before playback** — agreed it's the remaining gap, but making the authenticated preview route Range-capable means a new auth design for that route (media elements can't carry the Bearer header), which is an independent backend change. Leaving as the follow-up decision you framed it as.
- **Label rewrite is destructive on persisted content** — keeping the tactical approach for this PR per your note that recording the direction suffices. The `[label](file:id "real.mp4")` title-syntax alternative needs `_MARKDOWN_FILE_REFERENCE_RE` to learn title parsing and relies on the model emitting titles reliably; follow-up material.
- **Low 5 (`GENERATED_ARTIFACT_EXTENSIONS` has no audio)** — left as-is: the audio tools register via `workspace.auto_register_files()` and never consult that set, and adding audio there would expand executor artifact registration behavior beyond this PR's scope. The graceful degradation you described (generic file link) stands in the meantime.
- **Low 7 (`preload="metadata"` inert under blob src)** — acknowledged; per your note it's not worth changing and resolves itself if the Range follow-up lands.

yiboyasss:
Follow-ups from the review are now tracked: #1201 (Range-capable streaming on the authenticated path) and #1202 (preserve model-authored labels via markdown title syntax).

rogercloud (approved):
## Re-review: PR #1196 (round 4)

Commit `a79bb7cf` ("fix: harden media label restore and surface inline media load failures") is the only new commit since `43631542`, and it addresses the round-3 findings directly. Backend: `_UNSAFE_LABEL_RE` in `src/xagent/web/services/file_reference_output_service.py` was widened from `[`/`]`/backslash to all C0 control characters plus DEL, closing the blank-line-splits-the-paragraph case. Frontend: `loadErrorText` is now threaded into the audio/video players and rendered when the media element errors after both load paths are exhausted; the duplicated audio/video wrapper JSX was extracted into a shared `InlineMediaPreview`; and `getInlineFilePreviewMimeType` was reduced to a bare table lookup. Six new tests were added (control-character filenames, empty-label rewrite, audio extension parity, video public-preview fallback, video blob revocation on unmount, terminal-failure error state).

### Design verdict

**Acceptable with reservations.** The fix addresses both real causes of #1191 — the missing `video` preview kind on the frontend and the extension-losing label rewrite on the backend — and the `useResolvedMediaUrl` + `InlineMediaPreview` refactor is proportionate rather than over-engineered: it removes roughly 45 lines of triplicated effect logic and follows the existing player markup conventions in `file-viewer.tsx`. Design-level notes:

- **D1 (major, tracked and deferred to #1201).** The default in-app authenticated video path still downloads the entire file into a memory blob before playback — no Range requests, no seek-before-download, no progressive playback, whole file resident in memory. This is the same concern as prior finding #5, and issue #1201 is open and on point. Worth flagging that it is not hypothetical: generated videos have no size cap. `_download_video` in `src/xagent/core/tools/core/video_tool.py:493-566` streams via `aiohttp` + `iter_chunked` with only a 1-hour timeout and no byte-count limit, and `MAX_FILE_SIZE` in `src/xagent/web/config.py` is never imported by `video_tool.py` — it gates user uploads and KB files only. Deferring is fine, but please prioritize #1201 rather than letting it sit, or add an interim duration/size guard on video generation until Range support lands.
- **D2 (new this round, minor).** In `frontend/src/components/file/inline-file-preview.tsx`, the "skip blob-fetch" decision inside `useResolvedMediaUrl` is inferred from URL string equality (`fileAccess.previewUrl(fileId) !== previewUrl`) rather than an explicit capability on `FileAccessPolicy` (e.g. a `requiresBlobFetch: boolean`). Both policies that exist today (`defaultFileAccessPolicy`, `createPublicFileAccessPolicy`) produce the correct result, but nothing in the type system enforces the invariant: a future policy that reuses one URL for both routes while still needing an auth header would silently break media loading with no compile-time signal. Low-priority follow-up, not blocking.
- **D3 (tracked and deferred to #1202).** The label rewrite is destructive on persisted content and permanently discards the model's original, possibly localized prose. This is prior finding #13; issue #1202 is open with a concrete title-syntax plan. Confirmed unchanged in this round — `_MARKDOWN_FILE_REFERENCE_RE`'s target group still cannot parse title syntax. Reasonable to defer; the parser work is legitimately out of scope for a PR already three rounds deep.
- **D4 (new this round, minor-to-moderate, model-dependent).** The backend skips the label restore for `!`-prefixed references (`if not prefix and ...`). If a model ever emits `prose label` instead of the plain-link form, the renderer defaults an undetermined `previewKind` to `'image'` (`frontend/src/components/markdown/markdown-renderer.tsx:436`, `?? 'image'`), and `InlineImagePreview` has no `onError` handler at all — the result is a broken `` that fails silently, worse than the pre-PR download link. This is not reachable via the instructed path: `markdown_reference_for_artifact` (`src/xagent/web/services/artifacts.py:88-100`) only emits `!` for `artifact_type == "image"`, and video/audio artifacts are always handed to the model as plain `[filename](file:id)`. So it is a latent gap contingent on model non-compliance rather than a guaranteed bug. A defensive fix would be cheap: either apply the label restore to `!`-prefixed refs too, or give `InlineImagePreview` a basic `onError` fallback.
- **D5/D6 (low).** Extension lists are duplicated across at least four unsynchronized places: `artifacts.py`, `_AUDIO_EXTENSIONS`, the frontend regexes, and — with visible drift already — `frontend/src/components/file/file-type-icon.tsx` and the files view, which carry a different set (they include `.avi`/`.mkv`, absent from `GENERATED_ARTIFACT_EXTENSIONS`, and lack `.m4v`/`.mpeg`/`.mpg`). The two parity tests added this round compare the backend constant against a hand-copied Python literal inside the test file, so they are change-detectors rather than true cross-references against the `.ts` source; real frontend drift (say, adding `.mkv` to the frontend regex) would still go undetected. Not blocking, just worth knowing what the tests do and do not guarantee.

### Prior findings status

**Fixed this round or earlier (10 of 14)**

1. Audio extensions not recognized by `artifact_type_for_filename` (Gemini, high) — **FIXED.** Audio detection never routes through `artifact_type_for_filename`; it uses a dedicated `_AUDIO_EXTENSIONS` set as a separate branch in `_is_inline_preview_media` (`file_reference_output_service.py:29,32-48`). Regression test passes.
2. `.m4v` missing on the backend (major) — **FIXED.** Present in both `GENERATED_ARTIFACT_EXTENSIONS` and the video set in `artifacts.py:19,60`, pinned by two passing tests.
3. Unescaped filename substitution breaks the link (minor/moderate) — **FIXED.** `_UNSAFE_LABEL_RE` (`file_reference_output_service.py:64`) now skips the rewrite entirely when the filename contains `[`, `]`, backslash, or control characters, falling back to the original label. This is a "skip the rewrite" fix rather than "sanitize and rewrite"; it fully eliminates the link breakage, and the residual UX cost is a documented, reasonable tradeoff.
4. Redundant `suffix` parameter and `_INLINE_PREVIEW_LINK_TYPES` (minor) — **FIXED.** `_is_inline_preview_media` takes only `filename` and derives the suffix internally; the set is gone in favor of a direct `artifact_type_for_filename(filename) == "video"` check.
6. Test coverage gaps: docx exclusion, case-insensitivity, extension parity, mismatched label (medium) — **FIXED.** All four sub-items have dedicated tests; the full suite passes 17/17.
7. `_UNSAFE_LABEL_CHARS` omits `\n`/`\r` (medium) — **FIXED.** Widened to `re.compile(r"[\[\]\\\x00-\x1f\x7f]")`, covering all C0 control characters plus DEL. Regression test passes.
8. No error state when both media fetches fail (medium) — **FIXED.** `loadErrorText` is threaded through both players via the shared `InlineMediaPreview` and renders on terminal failure; suite passes 28/28. See F3 below for a refinement on the error semantics.
10. `InlineVideoPreview` duplicates ~25-30 lines of `InlineAudioPreview` JSX (code quality) — **FIXED.** Shared `InlineMediaPreview` extracted at `frontend/src/components/file/inline-file-preview.tsx:153-228`; both players are now thin wrappers passing `icon`/`bodyClassName`/`spinnerClassName`/`renderMedia`. One residual duplication remains — see Simplification below.
11. Dead branch in `getInlineFilePreviewMimeType` (minor) — **FIXED.** Reduced to a bare `INLINE_FILE_PREVIEW_MIME_BY_KIND[kind]` lookup with the type widened to `Partial<Record<InlineFilePreviewKind, string>>`.
12. No test pinning `_AUDIO_EXTENSIONS` parity against the frontend (minor) — **FIXED, with a caveat.** `test_audio_extension_detection_matches_frontend_set` is added and passing, but like its video counterpart it compares the backend constant against a hand-typed literal in the test file rather than the live `.ts` source. It matches the existing precedent, so this is consistent if imperfect coverage; it does not fully close the drift risk (see D5/D6).

**Confirmed deferred to tracked follow-up issues (2)**

5. Blob-only loading, no Range requests for video — **NOT FIXED**, tracked in **#1201** (open, on point). See D1 for why this matters more than it first appears.
13. Destructive label rewrite discards localized prose — **NOT FIXED**, tracked in **#1202** (open, with a concrete technical plan). See D3.

**Waived (2)**

9. Undeclared scope expansion — image/audio also skip blob-fetch on public surfaces — **WAIVED, with one note.** The PR description now carries an explicit "Behavior change on public/share surfaces (deliberate)" section matching the author's reasoning, and the reasoning is sound as far as it goes. See F4 for a residual gap that argument does not cover.
14. `GENERATED_ARTIFACT_EXTENSIONS` has no audio extensions (low) — **WAIVED.** `music_tool.py` and `sound_effect_tool.py` register via `workspace.auto_register_files()` and never consult this set; the only affected path (generic executor-produced audio) degrades gracefully to a plain download link. The author's reasoning holds.

### New findings this round

All four are minor and none are blocking.

- **F1 — Fragile URL-equality capability check (minor).** See D2 above. Suggest a `requiresBlobFetch` (or equivalent) field on `FileAccessPolicy` so the invariant is enforced by types rather than by string comparison.
- **F2 — Bang-prefix media references fall through to a silent broken image (minor-to-moderate, model-dependent).** See D4 above. A defensive `onError` on `InlineImagePreview` would be the cheapest mitigation and also improves the genuine-image failure case.
- **F3 — Terminal-failure state is too aggressive for playback-time errors (minor).** In `inline-file-preview.tsx`, the `failed` state (`failedUrl === resolvedUrl`) is set by an unconditional `onError` handler and is permanent with no retry. That handler fires on any native media `error` event, including transient or recoverable ones — for example a mid-playback decode hiccup on a blob that already loaded successfully — not only on the "both fallback paths exhausted" case it was designed for. Once triggered, the player unmounts permanently, because `resolvedUrl` never changes again as a result of a playback error. Consider scoping the handler to fire only while `resolvedUrl` is still unresolved or the fallback has not yet been attempted, or add a retry affordance.
- **F4 — Public-surface preview responses set no cache headers (minor, additive to #9).** The href-equivalence argument in the PR description does not fully hold on two axes it does not address. First, trigger: `<a href>` requires a user click, whereas ``/`<video>`/`<audio>` `src` fetches automatically on render, with no user action. Second, persistence: `_inline_preview_response` / `public_preview_file` (`src/xagent/web/api/files.py:184-228`) set no `Cache-Control` header at all, so the tokened URL and its response can now be written to browser disk cache automatically on every render and persist past the logical end of a guest session, bounded only by the token's 30-day expiry. That is a different and larger exposure than a click-triggered download. The tradeoff is still acceptable for this PR — real product benefit, and the token is scoped to the guest's own files — but a follow-up adding `Cache-Control: private, no-store` to the public preview response would close the residual gap cheaply.

Separately, the "Open" link resolving to `resolvedUrl` rather than a dedicated download URL is a **pre-existing pattern** already present in `InlineAudioPreview` before this PR; this PR only widens its scope to video. Noting it as a pre-existing nit, not a regression introduced here.

- L230-238, L261-269 (`frontend/src/components/file/inline-file-preview.tsx`): duplication — `InlineAudioPreview` and `InlineVideoPreview` each declare a byte-for-byte identical inline 7-field prop-type object literal (`source`, `previewUrl`, `filename`, `openLabel`, `loadErrorText`, `className?`, `fileAccess`). Extract a single `type MediaWrapperProps = {...}` and reuse it in both signatures.

net: -6 lines possible

No other simplification candidates survived verification — the `useResolvedMediaUrl` hook has exactly two real call sites with no dead parameters, and the `INLINE_FILE_PREVIEW_MIME_BY_KIND` type widening is a genuine simplification.

### Testing

Both suites were run and are green during this review: `.venv/bin/python -m pytest tests/web/services/test_file_reference_output_service.py` (17 passed) and the frontend vitest suite covering `inline-file-preview.test.tsx` (28+ passed).

### Recommendation

**Good to merge.** Every major and medium finding from the prior three rounds is fixed and covered by a passing test; the two unfixed design concerns (#5/D1 and #13/D3) are tracked in well-scoped open issues (#1201, #1202) and are reasonable to defer out of a PR already this deep. The four new findings above (F1-F4) plus the simplification are optional polish — none of them should block this, since the core P0 behavior in #1191 is fixed and works. The one thing I would ask for beyond merge is that **#1201 gets actual prioritization** rather than sitting indefinitely, given generated videos currently have no size cap on the authenticated path.

yiboyasss:
Filed a third follow-up for F4: #1226 (public preview responses lack Cache-Control, letting tokened media persist in browser disk cache).

PR #1198, opened 2026-08-08
fix(chat): open markdown links in a new tab instead of the widget iframe

yiboyasss:
## Summary
- Clicking an ordinary link in an AI reply inside the embedded Chat Widget (an iframe on a third-party site) navigated the iframe itself away from the conversation instead of opening a new tab
- `MarkdownLink` now renders `target="_blank"` and `rel="noopener noreferrer"` for real navigations (not in-page anchors or `mailto:` links), matching how external links are already handled elsewhere in the repo (sidebar, inline-file-preview, etc.)
- The behavior is scoped to the widget via a new `LinksOpenInNewTabCapability` context (same pattern as `AgentCardPresentationCapability`): off by default for every `MarkdownRenderer` consumer (skill-hub docs, file viewer, conversation logs, share pages, ...), enabled only by the two widget entry points (`PublicAgentChatPage` in widget mode and `SessionAgentChatPage`)
- `file:` (file preview) and `agent:` (agent card) links go through their own dedicated branches and are unaffected

## Test plan
- [x] `npx vitest run src/components/ui/__tests__/markdown-renderer.test.tsx` (38 tests pass: target/rel applied only when the capability is enabled; anchors/mailto stay same-tab; javascript:/empty hrefs never get target/rel; relative links unchanged by default)
- [x] `npx vitest run` full frontend suite (only pre-existing kb test failures, reproduced on the base branch)
- [x] `npx tsc --noEmit` passes
- [x] Manually click a link in a live Chat Widget (valid token + backend) to confirm it opens a new tab

### DIFF
--- a/frontend/src/components/ui/markdown-renderer.tsx
+++ b/frontend/src/components/ui/markdown-renderer.tsx
@@ -18,7 +18,9 @@ import {
 import { getApiUrl } from '@/lib/utils'
 import {
   AgentCardPresentationCapability,
+  LinksOpenInNewTabCapability,
   resolveAgentCardPresentationCapability,
+  resolveLinksOpenInNewTabCapability,
 } from '@/contexts/presentation-capabilities'
 import {
   isManagedFileUrl,
@@ -61,6 +63,7 @@ interface MarkdownRendererProps {
   className?: string
   filesDisabled?: boolean
   agentCardsEnabled?: boolean
+  linksOpenInNewTab?: boolean
   onFileClick?: (filePath: string, fileName: string) => void
   onAgentClick?: (agentId: string, agentName: string) => void
 }
@@ -260,6 +263,7 @@ function containsPreviewFileLinkNode(node: any): boolean {
 type MarkdownRendererContextValue = {
   filesDisabled: boolean
   agentCardsEnabled: boolean
+  linksOpenInNewTab: boolean
   onFileClick?: (filePath: string, fileName: string) => void
   onAgentClick?: (agentId: string, agentName: string) => void
   openLabel: string
@@ -309,6 +313,7 @@ function MarkdownLink({
   const {
     filesDisabled,
     agentCardsEnabled,
+    linksOpenInNewTab,
     onFileClick,
     onAgentClick,
     openLabel,
@@ -404,15 +409,15 @@ function MarkdownLink({
     return <span>{presentationChildren}</span>
   }
 
-  const opensInNewTab = Boolean(href) && !/^(#|mailto:|tel:)/i.test(href ?? '')
+  const opensInNewTab = linksOpenInNewTab && !!href && !/^(#|mailto:)/i.test(href)
 
   return (
     <a
+      {...props}
       href={href || undefined}
       title={presentationTitle || undefined}
       target={opensInNewTab ? '_blank' : undefined}
       rel={opensInNewTab ? 'noopener noreferrer' : undefined}
-      {...props}
     >
       {presentationChildren}
     </a>
@@ -496,6 +501,7 @@ export function MarkdownRenderer({
   className = '',
   filesDisabled = false,
   agentCardsEnabled,
+  linksOpenInNewTab,
   onFileClick,
   onAgentClick,
 }: MarkdownRendererProps) {
@@ -506,16 +512,22 @@ export function MarkdownRenderer({
     agentCardsEnabled,
     inheritedAgentCardsEnabled,
   )
+  const inheritedLinksOpenInNewTab = React.useContext(LinksOpenInNewTabCapability)
+  const resolvedLinksOpenInNewTab = resolveLinksOpenInNewTabCapability(
+    linksOpenInNewTab,
+    inheritedLinksOpenInNewTab,
+  )
   const contextValue = React.useMemo<MarkdownRendererContextValue>(
     () => ({
       filesDisabled,
       agentCardsEnabled: resolvedAgentCardsEnabled,
+      linksOpenInNewTab: resolvedLinksOpenInNewTab,
       onFileClick,
       onAgentClick,
       openLabel: t('files.previewDialog.buttons.open'),
       loadErrorText: t('files.previewDialog.errors.loadFailed'),
     }),
-    [filesDisabled, onFileClick, onAgentClick, resolvedAgentCardsEnabled, t]
+    [filesDisabled, onFileClick, onAgentClick, resolvedAgentCardsEnabled, resolvedLinksOpenInNewTab, t]
   )
   const displayContent = filesDisabled
     ? sanitizeFilesDisabledPresentationText(content)
--- a/frontend/src/components/ui/markdown-renderer.tsx
+++ b/frontend/src/components/ui/markdown-renderer.tsx
@@ -405,7 +405,13 @@ function MarkdownLink({
   }
 
   return (
-    <a href={href || undefined} title={presentationTitle || undefined} {...props}>
+    <a
+      href={href || undefined}
+      title={presentationTitle || undefined}
+      target="_blank"
+      rel="noopener noreferrer"
+      {...props}
+    >
       {presentationChildren}
     </a>
   )
--- a/frontend/src/components/ui/markdown-renderer.tsx
+++ b/frontend/src/components/ui/markdown-renderer.tsx
@@ -20,7 +20,6 @@ import {
   AgentCardPresentationCapability,
   LinksOpenInNewTabCapability,
   resolveAgentCardPresentationCapability,
-  resolveLinksOpenInNewTabCapability,
 } from '@/contexts/presentation-capabilities'
 import {
   isManagedFileUrl,
@@ -63,7 +62,6 @@ interface MarkdownRendererProps {
   className?: string
   filesDisabled?: boolean
   agentCardsEnabled?: boolean
-  linksOpenInNewTab?: boolean
   onFileClick?: (filePath: string, fileName: string) => void
   onAgentClick?: (agentId: string, agentName: string) => void
 }
@@ -409,7 +407,8 @@ function MarkdownLink({
     return <span>{presentationChildren}</span>
   }
 
-  const opensInNewTab = linksOpenInNewTab && !!href && !/^(#|mailto:)/i.test(href)
+  const opensInNewTab = linksOpenInNewTab && !!href
+    && !(href.startsWith('#') || href.toLowerCase().startsWith('mailto:'))
 
   return (
     <a
@@ -501,7 +500,6 @@ export function MarkdownRenderer({
   className = '',
   filesDisabled = false,
   agentCardsEnabled,
-  linksOpenInNewTab,
   onFileClick,
   onAgentClick,
 }: MarkdownRendererProps) {
@@ -512,22 +510,18 @@ export function MarkdownRenderer({
     agentCardsEnabled,
     inheritedAgentCardsEnabled,
   )
-  const inheritedLinksOpenInNewTab = React.useContext(LinksOpenInNewTabCapability)
-  const resolvedLinksOpenInNewTab = resolveLinksOpenInNewTabCapability(
-    linksOpenInNewTab,
-    inheritedLinksOpenInNewTab,
-  )
+  const linksOpenInNewTab = React.useContext(LinksOpenInNewTabCapability)
   const contextValue = React.useMemo<MarkdownRendererContextValue>(
     () => ({
       filesDisabled,
       agentCardsEnabled: resolvedAgentCardsEnabled,
-      linksOpenInNewTab: resolvedLinksOpenInNewTab,
+      linksOpenInNewTab,
       onFileClick,
       onAgentClick,
       openLabel: t('files.previewDialog.buttons.open'),
       loadErrorText: t('files.previewDialog.errors.loadFailed'),
     }),
-    [filesDisabled, onFileClick, onAgentClick, resolvedAgentCardsEnabled, resolvedLinksOpenInNewTab, t]
+    [filesDisabled, onFileClick, onAgentClick, resolvedAgentCardsEnabled, linksOpenInNewTab, t]
   )
   const displayContent = filesDisabled
     ? sanitizeFilesDisabledPresentationText(content)
--- a/frontend/src/components/ui/markdown-renderer.tsx
+++ b/frontend/src/components/ui/markdown-renderer.tsx
@@ -404,12 +404,14 @@ function MarkdownLink({
     return <span>{presentationChildren}</span>
   }
 
+  const opensInNewTab = Boolean(href) && !/^(#|mailto:|tel:)/i.test(href ?? '')
+
   return (
     <a
       href={href || undefined}
       title={presentationTitle || undefined}
-      target="_blank"
-      rel="noopener noreferrer"
+      target={opensInNewTab ? '_blank' : undefined}
+      rel={opensInNewTab ? 'noopener noreferrer' : undefined}
       {...props}
     >
       {presentationChildren}
--- a/frontend/src/components/widget/public-agent-chat-page.tsx
+++ b/frontend/src/components/widget/public-agent-chat-page.tsx
@@ -591,6 +591,10 @@ export function PublicAgentChatPage({
     capabilities: {
       agentCards: "disabled",
       voice: "disabled",
+      // Only the embedded widget iframe needs this: a "share" visitor already
+      // has the page to themselves, so an in-tab navigation away from it is
+      // ordinary browser behavior, not a lost conversation.
+      linksOpenInNewTab: authMode === "widget" ? "enabled" : "disabled",
     },
     buildWebSocketUrl: ({ baseUrl, taskId, token }) =>
       `${baseUrl}/${authMode === "share" ? "api/share" : "api/widget"}/chat/ws/${taskId}${token ? `?token=${token}` : ""}`,
--- a/frontend/src/components/widget/public-agent-chat-page.tsx
+++ b/frontend/src/components/widget/public-agent-chat-page.tsx
@@ -591,9 +591,10 @@ export function PublicAgentChatPage({
     capabilities: {
       agentCards: "disabled",
       voice: "disabled",
-      // Only the embedded widget iframe needs this: a "share" visitor already
-      // has the page to themselves, so an in-tab navigation away from it is
-      // ordinary browser behavior, not a lost conversation.
+      // Widget mode covers both the embedded iframe and direct top-level
+      // visits to the widget URL; in either case the conversation has no
+      // other home, so an in-tab navigation loses it. A "share" visitor has
+      // an ordinary full page, where in-tab navigation is expected behavior.
       linksOpenInNewTab: authMode === "widget" ? "enabled" : "disabled",
     },
     buildWebSocketUrl: ({ baseUrl, taskId, token }) =>
--- a/frontend/src/components/widget/session-agent-chat-page.tsx
+++ b/frontend/src/components/widget/session-agent-chat-page.tsx
@@ -261,6 +261,11 @@ export function SessionAgentChatPage() {
   }, [bridge.session, bridge.status])
 
   const transport = useMemo<AppProviderTransportConfig>(() => ({
+    capabilities: {
+      // This page only renders for the embedded widget's session-resume
+      // route, so an in-tab navigation always abandons the visitor's iframe.
+      linksOpenInNewTab: "enabled",
+    },
     session: {
       connection,
       onConnectionClose: bridge.handleConnectionClose,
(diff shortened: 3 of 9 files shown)


frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -405,7 +405,13 @@ function MarkdownLink({
   }
 
   return (
-    <a href={href || undefined} title={presentationTitle || undefined} {...props}>
+    <a
+      href={href || undefined}
+      title={presentationTitle || undefined}
+      target="_blank"
+      rel="noopener noreferrer"
```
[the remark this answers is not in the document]
yiboyasss: Good catch, fixed in 7dfd4ee2 — target="_blank"/rel are now only applied when the href isn't an in-page anchor, mailto:, or tel: link.

rogercloud (commented):
## Summary

This PR makes the shared `MarkdownLink` renderer emit `target="_blank" rel="noopener noreferrer"` for any href that is not a fragment, `mailto:`, or `tel:` link, so that clicking a markdown link no longer navigates the embedded Chat Widget iframe away from the widget. The change is two files (+31/-1): the attribute logic in `frontend/src/components/ui/markdown-renderer.tsx` plus new assertions in the component's test file. The `rel` pair matches the convention already hardcoded at the app's other always-external link sites, and the widget iframe carries no `sandbox` attribute, so the mechanism works as intended.

## Design

**Verdict: acceptable with reservations** — the mechanism is correct and minimal, but its blast radius is wider than the bug it fixes.

**D1 — the fix is applied at a shared low-level component with no scoping.** The reported bug is specific to the embedded widget iframe, but `MarkdownLink` is reached by every `MarkdownRenderer`/`ChatMessage` consumer: the main app's task/chat panel (`task-conversation-panel.tsx`, which already distinguishes widget vs. main-app rendering through its `mode: "page" | "embedded-preview"` prop), the skill-hub page client, the skill-hub markdown-editor live preview, `file-viewer.tsx` (arbitrary user `.md` files), the workforces run page client, conversation logs, and `TraceEventRenderer`. As a result, relative in-app markdown links such as `./guide.md` or `../doc.md` rendered in skill-hub docs, the editor preview, or the file viewer will now also open in a new tab — a behavior change unrelated to the stated iframe bug, and one that no test records (see F2). Relatedly, the same-tab allowlist is really "not a fragment and not a contact link," which lumps in-app relative navigation in with genuine external navigation; scoping the behavior makes that distinction moot.

Recommended fixes, in order of preference:

1. **Scope through `MarkdownRendererContext`** — it already carries flags like `filesDisabled` and `agentCardsEnabled`; add a `linksOpenInNewTab` flag and set it only at the widget-rendering call sites.
2. **Use `<base target="_blank">` in the widget route's layout** — zero blast radius on the rest of the app.
3. **Keep it global, but say so explicitly.** If the team actually wants every markdown link app-wide to open in a new tab, that is defensible — but the PR description should state it, and the existing "preserves standard relative markdown links" test should be extended to assert the new `target`, so the decision is recorded rather than incidental.

This isn't a hard blocker (nothing breaks), but it's an unrecorded, untested behavior change on paths beyond the one this PR set out to fix, so I'd like to see it explicitly resolved one way or another before merge.

## Findings

### F2 — minor: test coverage gaps
`frontend/src/components/ui/__tests__/markdown-renderer.test.tsx`

Four gaps, none of which is a live defect on its own:

- The pre-existing "preserves standard relative markdown links" test (lines 683-694) is untouched by this PR and asserts only `href`, never `target`/`rel` — so the app-wide relative-link behavior change described in D1 is entirely unrecorded.
- No test for `tel:` hrefs, which would have surfaced F1 (see inline comment on markdown-renderer.tsx:407).
- No test for `javascript:` hrefs. To be clear, this is a **coverage gap, not a vulnerability**: such hrefs are already neutralized to `''` upstream by `safeUrlTransform` before `opensInNewTab` is evaluated, verified against the installed react-markdown. A regression test would simply pin that guarantee in place.
- No test for an empty href, e.g. `[x]()`.

Fix: add the `target` assertion to the relative-link test (ideally once D1 is settled, so it asserts the intended scoped behavior), plus cases for `tel:` and empty href.

- `L407: shrink:` the `?? ''` fallback in `Boolean(href) && !/^(#|mailto:|tel:)/i.test(href ?? '')` is unreachable — once `Boolean(href)` is true, `href` is provably a non-empty string. Replace with `const opensInNewTab = !!href && !/^(#|mailto:)/i.test(href)` (also dropping the dead `tel:` alternative, see inline comment on line 407); behavior is identical for undefined, null, empty, matching, and non-matching hrefs.

net: -0 lines (readability only — the statement stays one line, it just stops implying nullable input and unsupported schemes).

frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -404,8 +404,16 @@ function MarkdownLink({
     return <span>{presentationChildren}</span>
   }
 
+  const opensInNewTab = Boolean(href) && !/^(#|mailto:|tel:)/i.test(href ?? '')
```
rogercloud: **Minor.** Two things worth cleaning up on this line:

1. The `tel:` alternative in the regex can never fire: `safeUrlTransform` delegates to react-markdown's `defaultUrlTransform`, whose scheme allowlist (`^(https?|ircs?|mailto|xmpp)$/i`) does not include `tel`, so `tel:` hrefs are already rewritten to `''` and `Boolean(href)` is `false` before this regex runs. Keeping it implies `tel:` links are meaningfully supported/excluded when they aren't rendered as links at all. Either drop `|tel:`, or add a `tel:` passthrough to `safeUrlTransform` (lines 68-73) with a test if working `tel:` links are actually wanted.
2. The `?? ''` fallback is also unreachable — once `Boolean(href)` is true, `href` is provably a non-empty string.

Suggested replacement: `const opensInNewTab = !!href && !/^(#|mailto:)/i.test(href)` — behavior-identical for undefined/null/empty/matching/non-matching hrefs.

frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -404,8 +404,16 @@ function MarkdownLink({
...
+
   return (
-    <a href={href || undefined} title={presentationTitle || undefined} {...props}>
+    <a
+      href={href || undefined}
+      title={presentationTitle || undefined}
+      target={opensInNewTab ? '_blank' : undefined}
+      rel={opensInNewTab ? 'noopener noreferrer' : undefined}
+      {...props}
```
rogercloud: **Low, defensive.** `href` and `title` are destructured out of `props` so a later spread can't override them, but the computed `target`/`rel` are not, and `{...props}` is spread after them — a `target`/`rel` key in `props` would silently win over the computed value, the inverse of the `href`/`title` protection.

This is currently unreachable: no `rehype-raw` (or equivalent raw-HTML-in-markdown) plugin is configured anywhere in the repo, so a hast `<a>` node from ordinary markdown syntax can only carry `href`/`title`. The trailing-spread style also matches the pre-existing `file:` branch in this same component (lines 361-372), so this isn't a new anti-pattern, just one extended into a spot where it now matters slightly more.

Non-blocking — consider destructuring `target`/`rel` too, or moving `{...props}` ahead of the computed attributes, as future hardening.

frontend/src/components/ui/markdown-renderer.tsx
yiboyasss: Fixed in 6320af5c — dropped the dead tel: alternative and the unreachable `?? ''` fallback: `const opensInNewTab = linksOpenInNewTab && !!href && !/^(#|mailto:)/i.test(href)`.

frontend/src/components/widget/session-agent-chat-page.tsx
```diff
@@ -261,6 +261,11 @@ export function SessionAgentChatPage() {
   }, [bridge.session, bridge.status])
 
   const transport = useMemo<AppProviderTransportConfig>(() => ({
+    capabilities: {
```
rogercloud, line 264: **Major (tests).** This `linksOpenInNewTab: "enabled"` opt-in has no test coverage: neither `session-agent-chat-page.test.tsx` nor `session-agent-chat-page.integration.test.tsx` references `linksOpenInNewTab` or asserts on the `capabilities` block at all. The sibling opt-in in `public-agent-chat-page.tsx` did get assertions in its test file, so this route is the odd one out.

Failure scenario: someone refactors this `useMemo` and drops the `capabilities` key — the widget session-resume route silently goes back to eating link navigations inside the iframe, and every test still passes.

Suggest mirroring the assertion already used for the public widget page: assert the transport config handed to the mocked `AppProvider` carries `capabilities.linksOpenInNewTab === "enabled"`.

frontend/src/contexts/app-context-chat.tsx
```diff
@@ -1672,6 +1677,11 @@ export function AppProvider({
     filesDisabled,
     requestedAgentCardsEnabled,
   )
+  // Deliberately not resolveTransportCapability: that helper defaults every
+  // other capability to "enabled" for non-session transports, which is the
+  // wrong default here — this must stay off everywhere except the transports
+  // that explicitly request it.
+  const linksOpenInNewTab = transport?.capabilities?.linksOpenInNewTab === "enabled"
```
rogercloud, line 1684: **Major (tests).** This derivation — and the matching `LinksOpenInNewTabCapability.Provider` further down — is never exercised against the real `AppProvider`. Both widget page test files mock `AppProvider` out and only assert the literal they pass into the mock, so nothing covers the `=== "enabled"` comparison or the Provider wiring.

The sibling `agentCards` capability already has the pattern worth copying: `app-context-chat.test.tsx` mounts the real `AppProvider` around a real `<MarkdownRenderer>` and asserts the rendered output. One equivalent test here — render a link through the real provider with `capabilities.linksOpenInNewTab: "enabled"` and assert `target="_blank"`, plus the negative case without it — would close the gap.

Failure scenario: invert this comparison or delete the Provider wrapper and no test fails.

(Separately: the comment above explaining why `resolveTransportCapability` is deliberately not used here is genuinely helpful.)

frontend/src/components/widget/public-agent-chat-page.tsx
```diff
@@ -591,6 +591,10 @@ export function PublicAgentChatPage({
     capabilities: {
       agentCards: "disabled",
       voice: "disabled",
+      // Only the embedded widget iframe needs this: a "share" visitor already
+      // has the page to themselves, so an in-tab navigation away from it is
+      // ordinary browser behavior, not a lost conversation.
+      linksOpenInNewTab: authMode === "widget" ? "enabled" : "disabled",
```
rogercloud, line 598: **Minor (comment accuracy).** The comment says "only the embedded widget iframe needs this," but `authMode === "widget"` is broader than that: `app/widget/chat/[token]/page-client.tsx` sets `authMode="widget"` unconditionally, whether or not an `embed_ticket` query param is present — and this file already acknowledges that a few lines above ("Direct visits (no embed ticket) authenticate with the widget key carried in the opened URL"). So a direct top-level visit to the widget URL also gets `target="_blank"`.

Harmless in practice, but the gate is `embedTicket`-shaped rather than `authMode`-shaped. Either narrow the condition to the embed-ticket case or reword the comment so it doesn't promise more scoping than the code delivers.

frontend/src/contexts/app-context-chat.tsx
```diff
@@ -5994,6 +6004,7 @@ export function AppProvider({
 
   return (
     <AgentCardPresentationCapability.Provider value={agentCardsEnabled}>
+      <LinksOpenInNewTabCapability.Provider value={linksOpenInNewTab}>
```
rogercloud, line 6007: **Nit (formatting).** The new Provider wraps the existing tree but its children weren't re-indented — `<FileAccessProvider>` below sits at the same indent level as its new parent. Purely cosmetic, and nothing in the repo will flag it (no Prettier config, and ESLint only extends `next/core-web-vitals` / `next/typescript`), so take it or leave it.

frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -404,8 +409,16 @@ function MarkdownLink({
     return <span>{presentationChildren}</span>
   }
 
+  const opensInNewTab = linksOpenInNewTab && !!href && !/^(#|mailto:)/i.test(href)
```
rogercloud: **Simplification (shrink).** This regex reinvents a prefix check the rest of this file does with `.startsWith()` — there are six other prefix checks here and all of them use `startsWith`. Suggest matching the local convention:

```ts
const opensInNewTab =
  linksOpenInNewTab
  && !!href
  && !(href.startsWith('#') || href.toLowerCase().startsWith('mailto:'))
```

Behavior-identical across the cases that matter: empty string, `#anchor`, `MAILTO:a@b` and other casings, and a URL that merely contains "mailto" mid-string; `undefined`/`null` are already excluded by the preceding `!!href`.

Also confirming the earlier thread: dropping `tel:` here is correct, not a regression. `safeUrlTransform` delegates to react-markdown's `defaultUrlTransform`, whose allowlist is `^(https?|ircs?|mailto|xmpp)$/i`, so `tel:` and `sms:` hrefs are already rewritten to `''` upstream and fail `!!href`.

frontend/src/contexts/presentation-capabilities.tsx
```diff
@@ -13,3 +13,16 @@ export function resolveAgentCardPresentationCapability(
...
   return !filesDisabled && (requestedAgentCards ?? inheritedAgentCards)
 }
+
+// Off by default so every existing MarkdownRenderer consumer (skill-hub docs,
+// the file viewer, conversation logs, ...) keeps opening links in the same
+// tab. AppProvider turns this on only for the embedded Chat Widget, where an
+// in-tab navigation would carry the visitor's iframe away from the page.
+export const LinksOpenInNewTabCapability = React.createContext(false)
+
+export function resolveLinksOpenInNewTabCapability(
```
rogercloud: **Simplification (yagni).** The `requestedLinksOpenInNewTab` parameter — and the corresponding `linksOpenInNewTab` prop on `<MarkdownRenderer>` that feeds it — has no production caller. None of the eight non-test `<MarkdownRenderer>` JSX sites pass it, and neither do the two internal recursive renders inside `markdown-renderer.tsx`. The only consumers are two unit tests, which could wrap in `<LinksOpenInNewTabCapability.Provider value={true}>` instead — exactly what production does.

Dropping the parameter and the prop would leave context as the single source of truth for this one and shed a handful of lines across the two files. Fine to keep if a per-instance override is expected soon — just flagging that nothing needs it today.

rogercloud (approved):
## Summary

This PR stops markdown links inside the embedded Chat Widget from navigating the visitor's iframe away from the conversation. The final shape adds a `linksOpenInNewTab` transport capability that `AppProvider` derives and publishes through a new `LinksOpenInNewTabCapability` React context; `MarkdownLink` reads it and only then sets `target="_blank"` + `rel="noopener noreferrer"`, skipping in-page anchors (`#…`) and `mailto:` links. The capability is opted into by exactly two surfaces — the public widget page (when `authMode === "widget"`) and the widget session-resume page — so every other `MarkdownRenderer` consumer keeps its previous same-tab behavior.

Three commits since the first review pass:

- `6187b3da` — original fix: `target="_blank"` / `rel` applied unconditionally on every markdown link (this is the state gemini-code-assist reviewed).
- `7dfd4ee2` — added the scheme guard `Boolean(href) && !/^(#|mailto:|tel:)/i.test(href ?? '')`, addressing gemini's finding.
- `6320af5c` (current HEAD) — the substantive one. Beyond what its commit message lists (dropping the dead `tel:` branch and the `?? ''` fallback, reordering the prop spread, adding `javascript:` / empty-href coverage), this commit also introduces the entire capability-scoping mechanism: the new `LinksOpenInNewTabCapability` context and `resolveLinksOpenInNewTabCapability` in `frontend/src/contexts/presentation-capabilities.tsx`, the per-transport `linksOpenInNewTab` capability + derivation + Provider in `frontend/src/contexts/app-context-chat.tsx`, and the opt-in at the two widget call sites.

## Status of prior findings

| Thread | Status | Evidence |
| --- | --- | --- |
| gemini-code-assist — unconditional `target`/`rel` breaks anchor/`mailto:`/`tel:` links | FIXED | Attributes are now gated on the capability chain and the `#`/`mailto:` exclusion; the default is off everywhere (`presentation-capabilities.tsx`: `LinksOpenInNewTabCapability = React.createContext(false)`). |
| Scoping concern — fix applied at the shared low-level component with no scoping | FIXED, via the recommended option 1 | Scoped through context + a capability flag set only at widget call sites: `contexts/presentation-capabilities.tsx`, `contexts/app-context-chat.tsx`, `components/widget/public-agent-chat-page.tsx` (`authMode === "widget"` only), `components/widget/session-agent-chat-page.tsx`. All eight other `MarkdownRenderer` call sites inherit `false`. |
| Test coverage gaps | FIXED | The relative-link test now asserts `not.toHaveAttribute('target')` / `('rel')`, and new cases cover `javascript:` and empty hrefs. |
| Inline nit — dead `tel:` alternative and unreachable `?? ''` | FIXED exactly as suggested | `const opensInNewTab = linksOpenInNewTab && !!href && !/^(#|mailto:)/i.test(href)`. |
| Inline nit — `{...props}` spread after the computed `target`/`rel` | FIXED | `{...props}` is now spread first, so the computed attributes always win. |

One confirmation while I'm here: dropping `tel:` from the regex is correct and not a regression. `safeUrlTransform` delegates to react-markdown's `defaultUrlTransform`, whose allowlist is `^(https?|ircs?|mailto|xmpp)$/i` — `tel:` and `sms:` are rewritten to `''` before `opensInNewTab` is evaluated, so the `!!href` guard already excludes them.

## Design (current HEAD)

Sound. The three-layer capability → context → prop mechanism mirrors the pre-existing `AgentCardPresentationCapability` pattern in the same two files, and is precisely the scoping the previous review asked for. I checked three candidate design smells — bypassing the shared `resolveTransportCapability` helper, hard-coding `"enabled"` at the call sites, and the asymmetry between session and non-session capability resolution — all three check out as intentional and are explained by in-code comments. No open design concerns.

## Findings

No blocking issues. The two majors below are test-coverage gaps against already-correct behavior, not functional bugs — flagging for follow-up rather than to hold up this PR.

### Major

1. `frontend/src/components/widget/session-agent-chat-page.tsx` — the session-resume opt-in (`capabilities: { linksOpenInNewTab: "enabled" }`) has zero test assertions in either `session-agent-chat-page.test.tsx` or `.integration.test.tsx`. The sibling change in `public-agent-chat-page.tsx` did get assertions. A future refactor of that `useMemo` could silently drop the capability with the whole suite staying green.
2. `frontend/src/contexts/app-context-chat.tsx` — the `linksOpenInNewTab` derivation and its Context Provider are never exercised against the real `AppProvider`; both widget page tests mock `AppProvider` out and only assert the literal passed into the mock. The sibling `agentCards` capability has an end-to-end test mounting the real `AppProvider` + a real `<MarkdownRenderer>` — worth mirroring here.

### Minor

3. `frontend/src/components/widget/public-agent-chat-page.tsx:597` — the comment says "only the embedded widget iframe needs this," but `authMode === "widget"` also covers direct, non-embedded top-level visits (per this file's own comment on the auth branch a few lines above). Harmless in practice, but the gate is `embedTicket`-shaped, not `authMode`-shaped.
4. `frontend/src/contexts/app-context-chat.tsx` — the new `<LinksOpenInNewTabCapability.Provider>` wraps the existing tree but its children weren't re-indented. Purely cosmetic; nothing in the repo's lint/format setup would catch it.

## Simplification opportunities (net: -10 lines possible)

- `frontend/src/components/ui/markdown-renderer.tsx:412` — shrink: the regex `/^(#|mailto:)/i.test(href)` reinvents a prefix check this file does with `.startsWith()` in six other places. Replace with `const opensInNewTab = linksOpenInNewTab && !!href && !(href.startsWith('#') || href.toLowerCase().startsWith('mailto:'))` — behavior-identical across the cases that matter.
- `frontend/src/components/ui/markdown-renderer.tsx` (the `linksOpenInNewTab` prop) and `frontend/src/contexts/presentation-capabilities.tsx:23-27` — yagni: the per-instance override has no production caller; none of the eight non-test `<MarkdownRenderer>` JSX sites pass it. Its only consumers are two unit tests, which could instead wrap in `<LinksOpenInNewTabCapability.Provider value={true}>` — the pattern production already uses.

## Verification

- `npx vitest run` on the two directly changed test files: 59/59 pass.
- `npx tsc --noEmit`: clean.
- No dependency changes in this diff.

Nice iteration on this one — two rounds of feedback absorbed cleanly, and the scoping mechanism landed in the right shape.

frontend/src/components/ui/markdown-renderer.tsx
yiboyasss: Hardened in 6320af5c — {...props} is now spread first, so the explicit href/title/target/rel always win.

Issue #1139, opened 2026-08-05
bug: Cannot create Knowledge Base with valid collection name


yiboyasss:
### Description

Creating a Knowledge Base fails when using a valid name.

For example, creating a Knowledge Base with the name "test" returns an access denied error.

### Problem Statement

The Knowledge Base name validation accepts common names such as test, but the creation request fails with an unexpected permission-related error.

Current error:

````
{
  "detail": "Access denied for collection: test"
}
````

The error suggests a permission or collection ownership issue instead of a naming validation problem.

### Steps to Reproduce

1. Navigate to the Knowledge Base creation page.

2. Enter:
````
test
````
as the Knowledge Base name.

3. Submit the creation request.
4. Observe the error response.

PR #1199, opened 2026-08-08
fix(kb): report taken collection names as 409 conflict, not 403 denial

bluefish-08:
Refs #1139

## Problem

Creating a knowledge base with an ordinary name like `test` can fail with:

```json
{"detail": "Access denied for collection: test"}
```

The user did nothing unauthorized. Collection names live in a global namespace — `collection_metadata` keys on `name` alone (`lancedb_stores.py:192`) — so once any user has created `test`, nobody else can. `_ensure_collection_access` then reports that as an access violation:

```
not in my visible list
  → check globally
    → absent  → allow_create=True lets it through
    → present → 403 "Access denied"      ← wrong for a caller who is naming
                                            a new collection
```

## Fix

A taken name answers **409 Conflict** instead of 403 when the caller is picking a name; a caller reaching for someone else's existing collection still gets **403**.

Which of the two applies is not the same question as `allow_create`, which turned out to do two jobs: *accept a name with no collection behind it yet*, and *treat a taken name as a naming problem*. `save_collection_config` needs the first without the second — its body is an `IngestionConfig` with no name field, reached from the settings panel of a knowledge base the caller already sees. So the branch is now controlled by `taken_name_is_conflict`, defaulting to `allow_create` (true for the five ingest entry points -- `/ingest`, `/ingest/jobs`, `/ingest-cloud`, `/ingest-web`, `/ingest-web/jobs` -- where the user did just type the name) and passed `False` there.

The rename endpoint inlined a copy of the same check and had the same symptom; it now returns 409 too.

## Wording

Two sentences, differing only in whether they advise a rename.

The shared wording states the fact — "Knowledge base name unavailable: X." — and stops. It cannot say more: `_ensure_collection_access` raises it both while a name is being chosen and while an existing collection is being written to, from requests the server cannot tell apart, and "pick another name" is useless on a screen with no name field.

The rename dialog is the one flow whose request carries a user-typed name, so its two conflict exits pass `advise_rename=True` and append "Please choose a different name."

Neither form confirms that another tenant holds the name. This is not an anti-enumeration measure — 409-vs-success already reveals that much — it is that naming an owner would be false precision for a caller who cannot act on it. Where the collision is with the caller's *own* collection, the message does name it, because they can go rename or delete it.

## Scope

This makes the failure comprehensible; it does not let two users hold the same name. That requires changing collection identity, which is being designed separately — hence `Refs` rather than `Fixes`.

Deliberately out of scope: the generic `handle_kb_exceptions` wrapper still relays some storage-layer strings unfiltered. Those paths are not reachable from the flows in #1139 and auditing them is its own change.

## Tests

- Backend unit tests for both wordings, and for the rename exit that changed behaviour — the one 409 site that previously had none. Each verified by mutation: dropping the argument fails that test and only that test.
- e2e coverage for all seven ingest entry points (409), and for `save_collection_config` (403, inverted from an earlier round of this PR).
- Frontend tests for the copy branches, including the four rename 409 exits relayed verbatim.

```
pytest tests/web/api -k "kb or collection"                 229 passed
pytest tests/web_integration/test_kb_access_control_e2e.py  11 passed
vitest run src/lib/kb-ingest-feedback.test.ts               14 passed
tsc --noEmit                                                     clean
```

## Verification

pre-commit: ruff, ruff format, mypy, isort, codespell, npm lint, npm type-check — all passed


tests/web_integration/test_kb_access_control_e2e.py
```diff
@@ -143,13 +144,17 @@ def test_documents_check_cross_tenant_returns_403(
...
+        assert "name unavailable" in resp.json()["detail"]
 
-    def test_save_collection_config_cross_tenant_returns_403(
+    def test_save_collection_config_cross_tenant_returns_409(
         self, client: TestClient, tmp_path: Path
     ) -> None:
-        """Saving config against another tenant's collection name is forbidden."""
+        """Creating a knowledge base under a taken name reports a conflict, not denial.
+
+        Regression for #1139: the create dialog posts config first, so a name owned by
```
rogercloud: The docstring is inaccurate: the create dialog does not post config first. `knowledge-base-creation-dialog.tsx` only calls `/api/kb/ingest[/jobs]`, `/api/kb/ingest-web[/jobs]` and `/api/kb/ingest-cloud` (lines 388, 544, 677); `/collections/{c}/config` is reached only from `knowledge-base-detail.tsx`'s `handleSaveConfig`, i.e. the settings panel of an existing KB.

That also exposes a coverage gap. Of the seven `allow_create=True` call sites in `src/xagent/web/api/kb.py` (3625, 3871, 4222, 4429, 5366, 5785, 6650), only `save_collection_config` and `check_documents_exist_api` are covered by e2e tests here. The five ingest endpoints — 3871, 4222, 4429, 5366, 5785 — are exactly the ones the reported repro walks through, and none of them pin the 409-not-403 contract. Please correct the docstring and add at least one e2e case on an ingest endpoint (`POST /api/kb/ingest` is the cheapest) so the flow this PR exists to fix is actually protected.

frontend/src/components/kb/knowledge-base-detail.tsx
```diff
@@ -83,6 +83,8 @@ function getKnowledgeBaseToastCopy(
 ) {
   return {
     genericTitle,
+    nameUnavailableTitle: t("kb.errors.nameUnavailable"),
+    nameUnavailableDescription: t("kb.errors.nameUnavailableHint"),
```
rogercloud, line 87: This new copy never reaches two of the three flows that can now produce the 409. In this same file, `handleUpdateCollectionName` throws `new Error(errorData.detail || ...)` and then does `toast.error(err.message)`, and `handleSaveConfig` does the same — neither calls `getKnowledgeBaseErrorToastContent`, unlike `handleUpload` and `handleWebImport`.

Rename is one of only two backend sites this PR explicitly edited to emit the new 409, so it currently gets zero benefit from the frontend half of the same change: a zh-locale user renaming onto a taken name sees the raw English `Knowledge base name unavailable: <name>. Please choose a different name.` rather than `kb.errors.nameUnavailable` / `nameUnavailableHint`. Please route both handlers' error paths through `getKnowledgeBaseErrorToastContent`.

src/xagent/web/api/kb.py
```diff
@@ -7426,8 +7444,8 @@ async def rename_collection_api(
         )
         if any(c.name == safe_new_collection for c in all_named.collections):
             raise HTTPException(
-                status_code=403,
-                detail=f"Access denied for collection: {safe_new_collection}",
+                status_code=409,
```
rogercloud: Optional drive-by while you're in this block: the guard directly above (`if not any(c.name == safe_new_collection for c in visible_for_user.collections):` at line 7439) is tautological here. The `if any(...)` branch a few lines up always raises, so this `if not any(...)` is unconditionally true and costs a second full pass over `visible_for_user.collections`. It predates this PR, so feel free to leave it, but collapsing it to a plain `else` (or dropping the condition) is a one-line cleanup.

src/xagent/web/api/kb.py
```diff
@@ -3517,6 +3527,14 @@ async def _ensure_collection_access(
             status_code=404, detail=f"Collection not found: {collection_name}"
         )
 
+    if allow_create:
```
rogercloud: Worth a one-line `# NOTE:` here now that this branch is the canonical name-conflict path. The globally-exists check just above is a lock-free read-then-branch, so two concurrent creates of the same new name can both pass it and collide at the storage layer instead of getting a clean 409; the comparison is also exact-case, so `Test` and `test` are treated as distinct names. Both are pre-existing and not worsened here, but a comment would stop the next reader from assuming this check is authoritative.

rogercloud (changes requested):
## Summary

`_ensure_collection_access` in `src/xagent/web/api/kb.py` collapsed two very different failure modes into a single `403 Access denied for collection: {name}`: (a) a caller trying to *reach* a collection that belongs to someone else, and (b) a caller trying to *create* a knowledge base under a name that happens to be globally taken. Because collection names live in one global namespace, case (b) accused innocent users of an access violation for picking a common name like `test` (#1139).

This PR splits the two on the already-present `allow_create` parameter: when `allow_create=True` and the name exists globally but is invisible to the caller, it now raises `409 Conflict` with deliberately vague wording built by a new `_collection_name_unavailable_detail` helper; `allow_create=False` keeps the original 403 unchanged. `rename_collection_api`, which carries an inlined copy of the same check, gets the same treatment. On the frontend, a new `isNameUnavailableError` matcher in `kb-ingest-feedback.ts` maps the new detail substring to friendly localized copy (en + zh).

## Approach verdict: sound, with reservations

The diagnosis is right, the fix is minimal, and the branch placement is correct — the new 409 raise sits strictly *after* the "does this name exist globally at all" check, so it cannot shadow legitimate creation of a genuinely free name. Scope discipline is good: the PR explicitly does not attempt per-tenant name scoping (a real architecture change) and correctly references #1139 rather than closing it.

Three design-level reservations, none blocking, none mapping to a single line:

1. **`allow_create` is a good proxy for "naming context" at 6 of 7 call sites, and a stretch at the 7th.** `check_documents_exist_api` (`src/xagent/web/api/kb.py:6650`, `POST /collections/{c}/documents/check`) is a read-only pre-upload duplicate check (`action="read"`), yet it now answers a pure query with `409 Conflict` — a status code that implies the caller attempted a mutation. Not a regression (it was an equally-wrong 403 before), and framing it as "the pre-flight step of an ingest whose target name is unavailable" is defensible, but it's a rough edge worth naming.

2. **The `rename_collection_api` duplicate check is a genuine design smell, and the obvious fix would be wrong.** Rename cannot simply call `_ensure_collection_access`, because that helper *returns silently* when the name is already visible to the caller — whereas rename must actively reject renaming onto a name the caller already owns ("Target collection already exists"). That's a third outcome the helper cannot express. Sharing only the `_collection_name_unavailable_detail` string helper is an acceptable stopping point for a bug fix, but the status-code *decision* still lives in two places. A cleaner long-term shape is a single `_ensure_collection_name_available` covering all three outcomes: caller-owns-it / taken-by-others / free.

3. **The anti-enumeration rationale is overstated.** For a name invisible to the caller, `allow_create=True` returns success when the name is free and 409 when it is taken. That status-code split alone is a complete existence oracle, entirely independent of the message wording. Making the text vague is cosmetic, not substantive. Not a regression — the same oracle existed via 403 — but the new assertion `assert "already exists" not in detail` reads as if it pins a security property that is not actually achieved. Consider either dropping the "never confirm another tenant owns this name" framing from the description, or stating plainly that the status code already discloses existence and the wording change is only about not being gratuitous.

## Findings

### MAJOR

**1. `tests/web_integration/test_kb_access_control_e2e.py:150-156`** — the regression test's docstring misattributes the fix, and the 5 endpoints the create dialog actually calls have no e2e coverage.

The docstring claims "the create dialog posts config first". `knowledge-base-creation-dialog.tsx` only calls `/api/kb/ingest[/jobs]`, `/api/kb/ingest-web[/jobs]` and `/api/kb/ingest-cloud`; `/collections/{c}/config` is reached only from `handleSaveConfig` in `knowledge-base-detail.tsx`, the settings panel of an *existing* KB.

Mapping the seven `allow_create=True` call sites in `src/xagent/web/api/kb.py` (3625, 3871, 4222, 4429, 5366, 5785, 6650) against e2e coverage, only `save_collection_config` (3625) and `check_documents_exist_api` (6650) are covered. The five ingest endpoints — 3871, 4222, 4429, 5366, 5785 — are exactly the ones the #1139 repro walks through, and none of them pin the 409-not-403 contract. See inline comment for the specific fix.

**2. `frontend/src/components/kb/knowledge-base-detail.tsx`** — rename and save-config bypass the friendly copy this PR just added, showing raw English backend text to zh-locale users.

`handleUpdateCollectionName` throws `new Error(errorData.detail || ...)` and then does `toast.error(err.message)` without ever calling `getKnowledgeBaseErrorToastContent`; `handleSaveConfig` has the identical bypass. `handleUpload` and `handleWebImport` in the same file route through `getKnowledgeBaseErrorToastContent` correctly. Rename is one of only two backend sites this PR explicitly edited to emit the new 409, so it currently gets no benefit from the frontend half of the same PR — see inline comment.

### MINOR

**3. `frontend/src/components/kb/knowledge-base-detail.tsx:561-566,575-578`** (pre-existing, follow-up) — the `/documents/check` 409 is swallowed into a misleading "proceeding anyway" toast. Any non-OK response there triggers a generic "could not check for duplicates, uploading directly" toast then calls `doUpload()`, which hits the same name-unavailable error again during real ingest, where the correct friendly toast finally appears. Net effect: two toasts, the first one wrong. This handler is untouched by the diff and behaved identically under the old 403 — not this PR's obligation, but worth a follow-up issue since it's directly adjacent to the confusion this PR is trying to remove.

**4. `src/xagent/web/api/kb.py:7439`** (pre-existing, optional drive-by) — see inline comment on the tautological guard in `rename_collection_api`.

**5. `src/xagent/web/api/kb.py:3530`** (pre-existing, worth a one-line note) — see inline comment on the TOCTOU/case-sensitivity gap.

## Simplification

Lean already. `_collection_name_unavailable_detail` is small, used twice, and justified DRY; the frontend `isNameUnavailableError` matcher mirrors its existing sibling matchers in the same module.

## Criteria checked with no issues found

- **Backend correctness of the 409/403 split** — clean. Branch ordering verified safe; the admin short-circuit and the 503-on-storage-failure semantics are untouched.
- **No regressions in existing 403 tests elsewhere** — the multitenancy isolation e2e and document-delete fallback tests all run on paths this PR doesn't touch (`hide_missing=True` or `allow_create=False`).
- **i18n completeness** — `en.ts` and `zh.ts` are updated consistently, there is no third locale to miss, and the zh copy reads naturally.
- **Information disclosure** — no new leak beyond the pre-existing status-code oracle discussed in reservation 3; the 409 detail only echoes the name the caller submitted.
- **No other `Access denied for collection` sites left behind** — a repo-wide grep confirms only `kb.py:3514/3539/3566`, all accounted for.

## Test results

Per the author's stated verification (not re-run in this review): `tests/web/api/test_kb_dir.py` — 92 passed (2 new unit tests for the 409/403 split); `tests/web_integration/test_kb_access_control_e2e.py` — 5 passed (3 existing tests re-pointed 403 → 409); `frontend/src/lib/kb-ingest-feedback.test.ts` — 7 passed (1 new matcher test).

## Disposition

The core approach is right and does not need rework. But both MAJOR items are gaps worth closing before merge rather than deferring: the e2e suite doesn't cover any of the five ingest endpoints the reported bug actually traverses (and its regression docstring describes a flow that doesn't exist), and the frontend copy this PR adds never reaches two of the three backend sites the PR itself edited. Both are small, contained fixes.

tests/web_integration/test_kb_access_control_e2e.py
bluefish-08: Both fixed. The docstring no longer claims the create dialog posts config — verified your reading: the dialog only hits `/api/kb/ingest[/jobs]`, `/ingest-web[/jobs]` and `/ingest-cloud`, and `/config` is reached only from the detail page's settings panel.

Added `test_ingest_into_name_taken_by_other_tenant_returns_409`: tenant 1 ingests to claim the name, tenant 2 posts the same name to `POST /api/kb/ingest`, asserting 409 with the name-unavailable wording and neither `Access denied` nor `already exists` in the body. That is the reported repro itself, which was the gap — five of the seven `allow_create=True` sites had nothing pinning the contract.

Verified by removing the 409 branch: the new case fails alongside the existing ones. Fixed in b67f0bfd.

frontend/src/components/kb/knowledge-base-detail.tsx
bluefish-08: Both handlers now route through `getKnowledgeBaseErrorToastContent`, matching the shape `handleUpload` and `handleWebImport` already use in the same file.

While wiring it up I found something that changes the picture for the rename half, and I would rather surface it than quietly fix around it: **the rename dialog is currently unreachable in the UI.** `handleOpenEditDialog` is the only thing that sets `isEditDialogOpen`, and it has zero references anywhere in the repo — no button is wired to it. `formData.append("new_name", ...)` appears exactly once in the whole frontend, inside that handler.

So the user-visible symptom you described — a zh-locale user seeing raw English on rename — cannot happen today, and by the same token the backend rename branch this PR edits has no frontend consumer at all. I still wired rename through the shared path (six lines, identical to its neighbours), because it becomes correct the moment the entry point is restored and leaving one inconsistent error path is how the next person gets caught out.

The practical consequence for this PR: the rename half cannot be covered by a UI-driven test, so I did not write one that could never be exercised. What is tested is `handleSaveConfig`, which shares the code path and is reachable via the Save button — new case asserts the 409 surfaces as `kb.errors.nameUnavailable` / `nameUnavailableHint` rather than the backend English, and reverting that catch to a bare `toast.error(err.message)` fails it.

Worth deciding separately whether the rename entry point should be restored or the endpoint retired. Fixed in b67f0bfd.

src/xagent/web/api/kb.py
bluefish-08: Added. The NOTE states both properties — the existence check above is a lock-free read-then-branch, so concurrent creates of the same name can both pass it and collide at the storage layer instead of getting a clean 409, and the comparison is exact-case, so `Test` and `test` are distinct. It also says plainly that this branch improves the error, and is not an authoritative uniqueness check. Fixed in b67f0bfd.

src/xagent/web/api/kb.py
bluefish-08: Confirmed tautological — the `if any(...)` above always raises, so this is unconditionally true and costs a second pass. Leaving it, though, and I would rather say why than silently skip it.

Collapsing it to an `else` means dedenting ~10 lines, so the diff would carry a block of pure control-flow churn with zero behaviour change, sitting in cross-tenant access checking — code a reviewer has to read line by line. On a PR that is otherwise scoped to which status code and message come back, and that has already been through several rounds, that trade seems wrong. The wasted pass is also on a branch that has already decided to fail, not on the hot path.

Worth folding into a dedicated cleanup PR along with the 17 duplicated `Invalid collection name: {e}` prefixes from an earlier round. Happy to open one if useful.

rogercloud (changes requested):
## Follow-up review (round 2)

Re-verified every open item from my previous review plus a fresh design pass on the current diff (HEAD `b67f0bfd`). Summary below; "Resolved" items need no further action, "Open" items are new asks.

### Resolved since last review
- The raw-`err.message` toast bypass in `handleUpdateCollectionName`/`handleSaveConfig` is fixed — both now route through `getKnowledgeBaseErrorToastContent`, confirmed by diff against base and by the new `knowledge-base-detail.web-ingest.test.tsx` case.
- The tautological `if not any(...)` guard in `rename_collection_api` (kb.py:7445) — thanks for confirming and explaining the tradeoff; agreed a pure-cleanup diff isn't worth the churn here. Leaving as a follow-up item, not blocking.
- The test-docstring wording that used to misattribute which frontend flow hits `/ingest` vs `/config` is now accurate.

### Still open (new from this pass)

**Major — `check_documents_exist_api` (kb.py:6612-6656) now returns 409 from a read-only endpoint.**
It calls `_effective_knowledge_base_user(..., action="read")` and only reads document records, but passes `allow_create=True` into `_ensure_collection_access`, so a cross-tenant duplicate-check call gets `409 Conflict` — a status code that implies a failed *mutation* — on a pure query. This wasn't the case pre-PR (base version unconditionally raised 403 here too, but 403 at least didn't claim a write-conflict). It's now also locked into the contract suite via `test_documents_check_cross_tenant_returns_409`, which will make it harder to correct later. Suggest either passing `allow_create=False` here (since the caller isn't naming anything) or documenting why 409 is intentional for this endpoint.

**Major — the frontend classifies name-conflict errors by matching backend prose, not the HTTP status code.**
`isNameUnavailableError` (`frontend/src/lib/kb-ingest-feedback.ts:137`) does `message.includes("Knowledge base name unavailable:")`, but both call sites (`knowledge-base-detail.tsx:818-820`, `:862-864`) already discard `response.status` before building the `Error`, even though every relevant backend response is 409. No test pins the backend string to the frontend string — the backend e2e asserts the looser `"name unavailable"` while the frontend test asserts the stricter `"Knowledge base name unavailable:"`. A backend rewording that keeps the backend test green (e.g. `"This knowledge base name is unavailable: X"`) would silently break the frontend match and dump raw English into the toast. Recommend switching the frontend to branch on `response.status === 409` and dropping the substring match, or at minimum adding a shared-fixture test that fails if the two diverge.

**Design note — this doesn't fix #1139, and the PR's own framing should stay honest about that.**
Collection names are keyed globally (`_list_collections_with_retry(user_id=None, is_admin=True)`, kb.py:3518; independently confirmed at the storage layer, `collection_metadata` merge-keyed on `name` alone). After this PR, a second tenant still cannot create a KB named `test` — they get a clearer 409 instead of a confusing 403, but the underlying inability from #1139 is unchanged. I see the PR body's Scope section already discloses this ("does not let two users hold the same name"), which is good — just flagging so #1139 isn't closed as fully resolved by this PR; a follow-up for per-tenant namespacing is the real fix.

**Minor — renaming to a name you already own still leaks raw English.**
`rename_collection_api`'s same-owner branch (kb.py:7440-7444) raises 409 `"Target collection already exists: {name}"`, which `isNameUnavailableError` doesn't match (see above) — this is arguably the *more common* rename-conflict case, and it falls through to the generic toast showing the raw English string even to zh-locale users. The PR localized the rarer cross-tenant case and left this one as-is.

**Minor — inconsistent `.catch()` on `response.json()` in `handleSaveConfig`.**
Line 863 has no `.catch()` while line 881 (three lines later, same function) does. Both are inside the same try/catch, so a non-JSON body (e.g. a proxy error page) doesn't crash — but it does produce a raw `SyntaxError` message (like `"Unexpected token < in JSON at position 0"`) in the toast instead of the intended `t("kb.detail.errors.saveConfigFailed")` fallback. Worth aligning both call sites.

**Minor — `hide_missing=True` silently overrides `allow_create=True` if ever combined.**
In `_ensure_collection_access` (kb.py:3512-3516), the `hide_missing` short-circuit runs before `allow_create` is ever inspected. No current caller passes both flags together, so there's no live bug, but the docstring doesn't mention the precedence — worth a one-line note so a future caller doesn't get surprised.

**Minor — duplicated `getKnowledgeBaseToastCopy` helper.**
Byte-identical between `knowledge-base-detail.tsx:80-93` and `knowledge-base-creation-dialog.tsx:55-68`; this PR adds the same two lines to both copies independently. Worth consolidating next to `KnowledgeBaseErrorToastCopy` in `kb-ingest-feedback.ts` at some point (not blocking).

**Minor — rename endpoint gives a finer-grained conflict oracle than the create/ingest paths, and it's untested.**
`rename_collection_api` returns different 409 wording depending on whether the target name is yours vs. someone else's (kb.py:7443 vs :7454) — pre-PR this same distinction existed via 403-vs-409 status codes; this PR moved it into message text under a shared 409 status, which is a channel shift rather than a new leak, but no test pins today's split (nothing asserts the "Target collection already exists" wording is stable), so a future edit could accidentally collapse or widen it without any test failing.

### Coverage nits (small, non-blocking)
- Of the ~5 additional `allow_create=True` call sites beyond the one new test (`/ingest`), `/ingest/jobs`, `/ingest-cloud`, `/ingest-web`, and `/ingest-web/jobs` still have no e2e coverage for the 403→409 change — only `/ingest` got a new test this round.
- `frontend/src/components/kb/knowledge-base-detail.web-ingest.test.tsx` only exercises `handleSaveConfig`'s new toast path, not `handleUpdateCollectionName`'s (rename) — worth noting the rename dialog trigger (`handleOpenEditDialog`) currently has no caller anywhere in the UI, so this is pre-existing dead code and low priority to test.
- The two new backend unit tests in `tests/web/api/test_kb_dir.py` don't assert `mock_list.await_count`, unlike a sibling test — but the existing status-code/detail assertions combined with the finite `side_effect` list already catch the realistic regressions, so this is a very minor style nit rather than a real coverage hole.

No further action needed on: the "anti-enumeration" framing overstating the PR's own comments (checked — the comments only claim not to reveal *ownership*, not *existence*, so they're accurate as written); the F3 rollback-context branch ordering bug in `kb-ingest-feedback.ts` (currently unreachable — the two error families are built from disjoint backend code paths); and F11 (an untouched, still-passing e2e test already covers the `allow_create=False` → 403 case).

bluefish-08:
Round 2 addressed in `84910bf8`. Thanks — the two majors both turned out to be bigger than they looked from the outside.

**`check_documents_exist_api` → `allow_create=False`.** Checked the blast radius: the only caller is the detail page's pre-upload duplicate check on an *existing* KB (the create dialog goes straight to ingest), and it already degrades gracefully on any non-2xx, so 409 → 403/404 costs nothing in UX. Updated the contract test accordingly.

That change surfaced something the old flag was hiding. Two existing unit tests started failing because they never put the collection into the listing at all — they were passing only because `allow_create=True` waved them through. In other words this read-only endpoint used to accept a name that was never registered in `collection_metadata`. I fixed the tests by supplying the missing precondition (a visible collection, which is the real production state) rather than by changing their assertions to expect 404; their original assertions are untouched.

**Frontend classification now keys on the 409, and the prose matcher is gone entirely** — not kept as a fallback, since a fallback would silently absorb exactly the rewording failure you described. Verified with your own scenario: changing the backend detail to `"This knowledge base name is unavailable: X"` leaves the frontend tests green, where the old matcher would have broken. Also added a negative case asserting the conflict wording *without* a 409 falls through to the generic copy, so nobody can reintroduce the substring match.

The status now has to reach the classifier, and the scope was wider than the two call sites you flagged: the three ingest handlers in the create dialog are the actual #1139 path, and they would have regressed to raw English the moment the matcher was removed. All five wired up.

**Your third design reservation was right and I had it wrong.** I argued for vague wording as an anti-enumeration measure; as you say, the success-vs-409 split is a complete existence oracle on its own and the wording changes nothing. The wording stays (it still declines to name an owner, and costs the user nothing), but the assertions no longer pretend it is a security property — the comments now say plainly what it does guarantee, which is only that the copy does not confirm ownership.

**Rename onto a name you already own** is fixed by the same status-code change — different wording, same 409 — with a test using that exact string as input.

**Coverage**: the ingest e2e is now parameterised across all five endpoints. I checked first that `_ensure_collection_access` runs before any crawling or cloud fetch in each handler, so the 409 returns immediately and the cases stay cheap. Skipped the `await_count` nit — pinning call counts would tie the tests to implementation detail without catching anything the current assertions miss.

**Not doing the duplicated `getKnowledgeBaseToastCopy` merge in this PR**: that helper is being modified by #1193 (approved, awaiting merge), so touching it now is a guaranteed conflict for 14 lines of duplication. Happy to follow up once that lands.

Also noted your point that #1139 should not be closed by this — the PR body says so, and the real fix is per-tenant namespacing, which is being designed separately.

tests/web_integration/test_kb_access_control_e2e.py
```diff
@@ -146,10 +150,15 @@ def test_documents_check_cross_tenant_returns_403(
...
 
-    def test_save_collection_config_cross_tenant_returns_403(
+    def test_save_collection_config_cross_tenant_returns_409(
         self, client: TestClient, tmp_path: Path
     ) -> None:
-        """Saving config against another tenant's collection name is forbidden."""
+        """Saving config under a taken name reports a conflict, not a denial.
+
+        The settings panel of an existing knowledge base drives this endpoint, and it
+        passes ``allow_create=True``, so a foreign name used to surface as
```
rogercloud: **Major:** this docstring itself names the problem — "passes `allow_create=True`, so a foreign name used to surface as 'Access denied' here too" — but `save_collection_config` (`src/xagent/web/api/kb.py:3634`, not touched by this diff) never got the fix. It names nothing (body is `IngestionConfig`: embedding model, chunking settings), and its only frontend caller is the settings panel of an already-visible collection (`frontend/src/components/kb/knowledge-base-detail.tsx:860`).

This is the identical shape of the `check_documents_exist_api` issue from round 2 ("now returns 409 from a read-only endpoint... suggest passing `allow_create=False`"), which *was* fixed a few lines above this test. Here the test instead pins the unfixed behavior as intended: a genuine cross-tenant access attempt on a settings save now answers "choose a different knowledge base name" — nonsensical for a form with no name field. The frontend counterpart at `frontend/src/components/kb/knowledge-base-detail.web-ingest.test.tsx:241-290` pins the same thing.

Suggest dropping `allow_create=True` from `save_collection_config`'s `_ensure_collection_access` call to match the `/documents/check` fix, then updating this test (and the frontend one) to assert 403/"Access denied" instead.

frontend/src/lib/kb-ingest-feedback.ts
```diff
@@ -138,14 +140,27 @@ function isRollbackFailure(message: string): boolean {
...
   message: string,
-  copy: KnowledgeBaseErrorToastCopy
+  copy: KnowledgeBaseErrorToastCopy,
+  status?: number
 ): KnowledgeBaseErrorToastContent {
   const normalized = normalizeMessage(message)
   const originalError = extractOriginalIngestionError(normalized)
   const rootCause = originalError ?? normalized
 
+  if (status === 409) {
```
rogercloud: **Minor:** this branch is a direct consequence of fixing the round-2 ask to classify by status instead of prose — but on the rename endpoint, 409 isn't exclusive to name conflicts. `rename_collection_api` raises 409 at `src/xagent/web/api/kb.py:7446`/`:7457` (real name conflicts) and also at `:7484-7492` (physical-directory collision) and `:7530-7534` (lock contention, "Another operation is in progress") — neither of the latter two is a name conflict, and this branch now discards `errorData.detail` for all four, showing generic "name unavailable" copy instead.

The doc comment above (line 145, "409 is the backend's only name-conflict code") is factually incorrect for this endpoint, which is the root cause.

Filing as minor rather than major: I confirmed the rename dialog's UI entry point is still dead code (`handleOpenEditDialog` at `knowledge-base-detail.tsx:797` has zero callers), exactly as noted in round 1 — so this has no current user-visible path. It will matter the moment that entry point is restored, or if the endpoint is reached directly. Consider a machine-readable `code` field in the 409 body instead of relying on status alone, so this doesn't need revisiting later.

src/xagent/web/api/kb.py
```diff
@@ -7426,8 +7455,8 @@ async def rename_collection_api(
         )
         if any(c.name == safe_new_collection for c in all_named.collections):
             raise HTTPException(
-                status_code=403,
-                detail=f"Access denied for collection: {safe_new_collection}",
+                status_code=409,
+                detail=_collection_name_unavailable_detail(safe_new_collection),
```
rogercloud: **Minor:** this branch (cross-tenant collision) uses the vague `_collection_name_unavailable_detail` wording added to avoid confirming existence, while line 7448 (own-collection collision) still says `"Target collection already exists: {name}"`. The status code was unified to 409 for both; the wording wasn't — the own-collection branch still discloses the exact information the new helper exists to avoid. Not functional today (same dead-UI caveat as the previous comment applies), just an incomplete cleanup.

tests/web_integration/test_kb_access_control_e2e.py
```diff
@@ -162,5 +171,80 @@ def test_save_collection_config_cross_tenant_returns_403(
             json={"embedding_model_id": "text-embedding-v4"},
             headers={"Authorization": f"Bearer {t2}"},
         )
-        assert resp.status_code == 403
-        assert "Access denied" in resp.json()["detail"]
+        detail = resp.json()["detail"]
```
rogercloud, line 202: **Minor:** `detail = resp.json()["detail"]` runs before `assert resp.status_code == 409` on the next line. An unexpected 200 would raise `KeyError` instead of a readable status-mismatch failure. Same issue at line 244-245 below (worse there — `/ingest-cloud`'s success body is a bare list, so it'd raise `TypeError` instead). Suggest swapping the order in both tests.

src/xagent/web/api/kb.py
```diff
@@ -6629,7 +6656,9 @@ async def check_documents_exist_api(
         if not requested:
             return {"existing_filenames": []}
 
-        await _ensure_collection_access(safe_collection, _user, allow_create=True)
+        # Read-only duplicate check: it names nothing, so it must not report a
+        # naming conflict. Only an existing collection's detail page calls it.
+        await _ensure_collection_access(safe_collection, _user)
```
rogercloud, line 6700: **Minor:** correct fix for the round-2 ask, but worth noting the side effect: `allow_create=True` previously absorbed the window where `list_collections` briefly disagrees with LanceDB (documented at `src/xagent/web/api/kb.py:3557-3562`, "control-plane rename lag"). This endpoint now 404s during that window instead of passing through. Low impact since the frontend fails open here (`frontend/src/components/kb/knowledge-base-detail.tsx:561-566` warns and uploads anyway on any non-OK response), but no test covers either behavior — a one-line comment recording the tradeoff would help.

rogercloud (changes requested):
## Follow-up review (round 3, HEAD `84910bf8`)

I re-reviewed from a blank slate (design pass + full discovery pass), then cross-checked findings against both of your prior review rounds to avoid re-litigating resolved items.

### Summary of this diff
Since your round 2, the PR added: (1) `84910bf8` — switched the frontend from matching backend error strings to classifying purely on HTTP status (`status === 409`), directly addressing your round-2 ask; (2) locale entries and tests for the config-save and rename flows. The core 409/403 split itself is unchanged and still sound.

### New from this pass

**Major — `save_collection_config` still passes `allow_create=True` though it names nothing** (`src/xagent/web/api/kb.py:3634`)
You caught this exact pattern on `check_documents_exist_api` in round 2 ("now returns 409 from a read-only endpoint... suggest passing `allow_create=False`") and it was fixed. The identical defect exists on `save_collection_config`: its body is `IngestionConfig` (embedding model, chunking settings) with no name field, and its only caller is the settings panel of an already-visible collection (`frontend/src/components/kb/knowledge-base-detail.tsx:860`). A genuine cross-tenant access attempt here now answers "choose a different knowledge base name" instead of 403 — nonsensical for a form with no name field. Two new tests pin this as intended: `tests/web_integration/test_kb_access_control_e2e.py:153-179` and `frontend/src/components/kb/knowledge-base-detail.web-ingest.test.tsx:241-290`. Suggest dropping `allow_create=True` here to match the `/documents/check` fix, then updating both tests to assert 403/"Access denied".

**Minor — the new status-based classifier (`84910bf8`) is technically correct for its own goal but the rename endpoint's 409 is overloaded, so it discards real error text for two non-conflict cases.** (`frontend/src/lib/kb-ingest-feedback.ts:145,157`)
This is a direct side effect of fixing your round-2 major ("classify by status, not prose"). `rename_collection_api` raises 409 in four places — `src/xagent/web/api/kb.py:7446` and `:7457` are real name conflicts, but `:7484-7492` (physical-directory collision) and `:7530-7534` (lock contention, "Another operation is in progress") are not. The doc comment at `kb-ingest-feedback.ts:145` ("409 is the backend's only name-conflict code") is factually incorrect for this endpoint, and the `status === 409` branch now returns generic "name unavailable" copy for all four, discarding `errorData.detail` for the two non-conflict cases.
Filed as **minor, not major**, because I confirmed the rename dialog's UI entry point is still dead code exactly as bluefish-08 reported in round 1 (`handleOpenEditDialog` at `knowledge-base-detail.tsx:797` has zero callers/no wired button) — so this misclassification currently has no user-visible path. It becomes live the moment that entry point is restored, or if the endpoint is hit directly. Suggest a machine-readable `code` field in the 409 body rather than relying on status alone, so this doesn't have to be revisited when rename is wired up.

**Minor — divergent 409 wording for "target name taken" in rename, one branch still leaks existence** (`src/xagent/web/api/kb.py:7448` vs `:7459`)
Line 7448 (own-collection collision) says `"Target collection already exists: {name}"`; line 7459 (cross-tenant collision) uses the vague `_collection_name_unavailable_detail` helper specifically added to avoid confirming existence. The status code was unified to 409 for both; the wording wasn't. Not functional (frontend flattens both via status check), just an incomplete cleanup — same dead-UI caveat as above applies for current impact.

**Minor — new e2e tests read the response body before checking status** (`tests/web_integration/test_kb_access_control_e2e.py:174-175` and `:244-245`)
`detail = resp.json()["detail"]` runs before `assert resp.status_code == 409`. An unexpected 200 would raise `KeyError` (most cases) or `TypeError` at `:244` (`/ingest-cloud`'s success body is a bare list, not a dict) instead of a readable status-mismatch failure. Suggest swapping the two lines in both tests.

**Minor — `/documents/check`'s `allow_create` removal silently drops tolerance for collection-listing lag** (`src/xagent/web/api/kb.py:6661`)
Right fix per your round-2 ask, but a side effect: `allow_create=True` previously absorbed the window where `list_collections` briefly disagrees with LanceDB (documented at `src/xagent/web/api/kb.py:3557-3562`, "control-plane rename lag"). This endpoint now 404s during that window instead of passing through. Low impact — the frontend fails open here (`knowledge-base-detail.tsx:561-566`: warns and uploads anyway on any non-OK response) — but no test covers either behavior. Worth a one-line comment recording the tradeoff.

**Minor — no direct unit test asserts the #1139 happy path** (`tests/web/api/test_kb_dir.py:1383`)
All four direct `_ensure_collection_access` unit tests assert it *raises*; none asserts it returns cleanly for `allow_create=True` + a brand-new globally-absent name — the exact "create a KB called `test`" scenario this PR exists to fix. Mitigated by integration coverage (`test_kb_ingest_creates_collection_dir` and siblings POST a fresh name and assert 200 without mocking the helper), so a regression would be caught, just not localized to the helper.

### No further action needed on
- The `_ensure_collection_access` inline comment placement (`kb.py:3533-3545`) and the tautological `if not any(...)` guard (`kb.py:7450`) — both already discussed in round 1; you agreed the cleanup isn't worth the churn, and I confirmed no new information changes that call.
- The anti-enumeration wording limitation — your round-1 review already noted the status-code-is-already-an-oracle point, and the round-2 test file's own comment (`test_kb_access_control_e2e.py:178-181`) acknowledges it explicitly. Not a contradiction, just a documented limit of the approach.
- Stale module docstring in `test_kb_access_control_e2e.py:1-9` (still says "403/404 contracts") — confirmed still stale, but genuinely trivial; flagging inline rather than as a review-body item.

### Testing
`npx vitest run` on touched frontend test files: 11 passed. `npx tsc --noEmit`: clean. Backend tests were reviewed by reading only — no importable `xagent` module in this review environment's venv — please confirm `pytest tests/web/api -k "kb or collection"` and the e2e suite still pass in CI, especially after fixing the Major above.

bluefish-08:
Fixed the regression in the last round, plus two test-quality items.

**The regression.** `_rename_target_has_files_detail`'s docstring justified its "Please choose a different name." by contrast — *unlike the shared wording above*, rename has a name field. But `kb.py:7487`, inside `rename_collection_api`, uses exactly that shared wording. So the one dialog that does have a name input lost the advice for one of its two conflict exits while its sibling gained it, in the same round.

Rather than delete the advice from the sibling, I made the shared helper take `advise_rename` and had 7487 pass it. The distinction the docstring drew is real — `_ensure_collection_access` genuinely cannot tell a name being chosen from an existing collection being written to, and it fires from screens with no name field. Rename can tell: the new name is in the request. Now the parameter says that instead of a comment claiming it.

Both docstrings rewritten to describe what the code does.

**Tests.** `test_kb_rename_onto_another_tenants_name_advises_a_different_name` pins 7487 — the only 409 exit that both changed behaviour and had no test. Verified by mutation: dropping `advise_rename=True` fails that test and only that test.

The frontend `it.each` was feeding the pre-fix bare sentence as "verbatim backend output", which is no longer what the backend emits; updated. Also narrowed a comment that claimed to cover *each* 409 branch in kb.py (there are ten; the list has the rename dialog's four) and dropped a `not.toContain("choose a different name")` assertion that could never fail — `COPY` is Chinese, that string is English.

Full `tests/web/api -k "kb or collection"` plus the access-control e2e pass; `vitest src/lib` and `tsc --noEmit` clean.

tests/web_integration/test_kb_access_control_e2e.py
bluefish-08: Fixed, though not by dropping `allow_create` — that flag does two jobs and this endpoint needs one of them. It also means "accept a name with no collection behind it yet", which config saves rely on (settings are seeded before the first ingest, per the e2e suite). Dropping it would have turned that into a 404.

Split them: `taken_name_is_conflict` now decides the 409, defaulting to `allow_create` so every other caller is unchanged, and `save_collection_config` passes `False`. Cross-tenant config saves are 403 "Access denied" again; unknown names still pass through.

Test renamed to `..._returns_403` and inverted, and the frontend counterpart now mocks a 403. Verified by mutation: dropping the argument fails that e2e test and only that one.

src/xagent/web/api/kb.py
bluefish-08: Already there — the comment landed in the round-3 push (`src/xagent/web/api/kb.py`, above the call): it records that `allow_create=True` used to absorb the control-plane rename lag and that this path now 404s during that window, with the note that the frontend fails open. Happy to add a test if you want the tradeoff pinned rather than documented.

tests/web_integration/test_kb_access_control_e2e.py
bluefish-08: Both were reordered in the round-3 push — `assert resp.status_code` now precedes every `resp.json()["detail"]` in this file, and the `/ingest-cloud` one kept a comment explaining why the order matters there specifically (bare-list success body → `TypeError`).

src/xagent/web/api/kb.py
bluefish-08: Taken — collapsed. The outer `if not any(...)` is gone and the global lookup runs unconditionally after the visible-collection branch, which always raises. One less full pass over `visible_for_user.collections`.

frontend/src/lib/kb-ingest-feedback.ts
bluefish-08: The "discards `errorData.detail`" part no longer holds — the classifier takes a `conflictAdvice` alongside the status, and only `"advise-rename"` substitutes local copy. `"neutral"` relays the backend sentence verbatim, which is now safe because the backend stopped leaking owner ids.

That leaves your real point: is 409 + advice ever wrong? Grepping the call sites, `"advise-rename"` appears exactly three times, all in `knowledge-base-creation-dialog.tsx`, all on ingest endpoints where 409 does mean a taken name. The four detail-page call sites are `"neutral"`. The rename endpoint — the one with four different 409 causes — has no frontend caller at all, since `handleOpenEditDialog` is still dead as you noted.

So the copy cannot currently be wrong, and if the rename entry point is restored it would be wired as `"neutral"`, which relays whichever of the four sentences fired. I have kept the misleading doc comment out and left the `code` field alone: it is the right answer, but it changes the response body of ten endpoints and belongs with the `display_name` work rather than this stopgap.

src/xagent/web/api/kb.py
bluefish-08: This one is deliberate, and the comment above it says so: the helper exists to avoid confirming that *another tenant* holds a name. On this branch the collection is the caller own, already in their listing, so naming it discloses nothing they cannot see — and it is the more useful message, because they can go rename or delete it. Making it vague would cost information and protect nothing.

rogercloud (approved):
Re-review (round 4). This PR fixes #1139's misleading `403 Access denied` when a caller picks a collection name that is already taken in the global namespace: `_ensure_collection_access` (`src/xagent/web/api/kb.py`) now takes a `taken_name_is_conflict` flag, independent of `allow_create`, so a taken name answers `409 Conflict` when the caller is *naming* something, while a caller reaching for someone else's existing collection still gets `403`. The rename endpoint's inlined duplicate check got the same split, and the frontend adds a `conflictAdvice: "advise-rename" | "neutral"` field so only the rename dialog substitutes local "choose a different name" copy, with every other 409 relaying the backend's own detail. 8 new commits since round 3.

I re-traced the design from a blank slate this round rather than only diffing against round 3, and it holds up: the 409/403 split sits correctly *after* the global-existence check, the scope is honestly `Refs #1139` rather than `Fixes` (name collision itself is not solved, and shouldn't be here), and the three design reservations I went in with (single non-default caller for the new flag, `hide_missing`+`allow_create` interaction, divergent 409 wording) all turned out to be documented and deliberate on inspection. No design-level objection.

## Status of prior findings

- **Major — `save_collection_config` passed `allow_create=True` on a read/edit-only endpoint (no name field), turning cross-tenant access into 409**: **FIXED**. `kb.py:3665-3667` passes `taken_name_is_conflict=False` explicitly; pinned by `test_save_collection_config_cross_tenant_returns_403`.
- **Minor — rename endpoint's 409 overloaded across 4 causes; naive `status === 409` frontend classifier discarded detail for the 2 non-conflict causes**: **FIXED (refactored)**. `conflictAdvice` added at `frontend/src/lib/kb-ingest-feedback.ts:25/34/181`; only `"advise-rename"` substitutes local copy, `"neutral"` 409s (lock contention, directory collision) relay backend detail unchanged.
- **Minor — divergent 409 wording for target-name-taken looked like incomplete cleanup**: **WAIVED — confirmed intentional**. `kb.py:7484-7490` justifies naming the caller's *own* collision (actionable, already visible to them); the vague cross-tenant wording rationale lives in the shared helper docstring at `kb.py:3473-3480`. Only nit: that rationale is one function away from the branch it governs.
- **Minor — e2e tests read `resp.json()["detail"]` before asserting the status code**: **FIXED**. All call sites in `tests/web_integration/test_kb_access_control_e2e.py` assert status before touching the body.
- **Minor — `/documents/check`'s `allow_create` removal silently dropped tolerance for control-plane listing lag**: **FIXED**. `kb.py:6693-6699` documents the tradeoff and notes the frontend fails open.
- **Minor — no direct unit test for the #1139 happy path (helper returns cleanly for `allow_create=True` + globally-absent name)**: **NOT FIXED**. `tests/web/api/test_kb_dir.py` still covers only the 404 and 409 branches. Carried forward and merged into new finding 3 below.
- **(Earlier round) tautological `if not any(...)` guard in `rename_collection_api`, deferred as "not worth the churn"**: **FIXED**. Collapsed into a real distinct global-name lookup at `kb.py:7484-7501`.
- **(Earlier round) missing `# NOTE:` on the lock-free read-then-branch + exact-case comparison in the existence check**: **FIXED**. `kb.py:3565-3570` now states both properties.

No Major finding survives this round.

## New findings this round

1. **Minor / code quality — `frontend/src/components/kb/knowledge-base-detail.tsx` (~412-445/532, 602-651/731, 815-826/845, 857-876/914)**. A `failedStatus` local is captured and threaded into `source.status` at all 4 sites, but all 4 also pass `conflictAdvice: "neutral"`, and `getKnowledgeBaseErrorToastContent` (`kb-ingest-feedback.ts:181`) reads `source.status` only when `conflictAdvice === "advise-rename"`. All 4 locals are behaviorally dead. Drop them, or add a comment saying they're staged for a future advise-rename site.
2. **Minor / test coverage — `frontend/src/components/kb/knowledge-base-detail.tsx` rename handler (~812-852)**. The rewritten rename-dialog error path (title/description split, `.catch(() => ({}))` on `response.json()`, new toast shape) has no test — no rename-related assertions exist in any `knowledge-base-detail*.test.tsx`. This toast could break silently. Add one test for the 409-advise-rename case and one for a non-JSON error body.
3. **Minor / test layering — `tests/web/api/test_kb_dir.py`** (merges the carried-forward NOT FIXED item). The helper is never exercised at the unit layer for either (a) `allow_create=True` + globally-absent name → returns cleanly (the #1139 happy path), or (b) `allow_create=True, taken_name_is_conflict=False` — the exact combination `save_collection_config` uses. (b) is covered only end-to-end by `test_save_collection_config_cross_tenant_returns_403`; the branching being tested lives in the helper, so it belongs there too. Two short unit tests would close both.
4. **Minor / documentation — PR description**. It says "true for the seven ingest entry points," but only 5 entry points now ride the `allow_create`-defaulted path: `/ingest` (`kb.py:3914`), `/ingest/jobs` (4265), `/ingest-cloud` (4472), `/ingest-web` (5409), `/ingest-web/jobs` (5828). "Seven" was the pre-PR count including `save_collection_config` and `check_documents_exist_api`, both of which this PR moved out of that bucket. Worth correcting for future readers.
5. **Informational / pre-existing — `kb.py:3667` vs `kb.py:~3720`**. Confirmed against the base commit: not introduced here. `save_collection_config` accepts a not-yet-existing collection (`allow_create=True`) but `set_collection_rerank_model` does not (`hide_missing=False`, no `allow_create`), while `handleSaveConfig` (`knowledge-base-detail.tsx` ~867-897) calls `/config` then `/rerank-model` unconditionally in sequence — so for a brand-new collection the first can succeed and the second 404. Raising it only because this PR's new comment leans on the "settings are seeded before ingest" story. Follow-up issue, not a blocker.
6. **Informational — `frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx`**. The test "requires an explicit collection name for multiple file uploads" fails at HEAD; I checked out the base commit and it fails there too, so it is **not** a regression from this PR. But this PR touches the file and leaves the suite red. Separate cleanup ticket.

- `L25 (frontend/src/lib/kb-ingest-feedback.ts)`: yagni — `KnowledgeBaseConflictAdvice` is a boolean disguised as a 2-value string union with no third value ever used. It is consumed only as `source.conflictAdvice === "advise-rename"` (L181), all 7 call sites pass one of exactly 2 literals, and it never crosses a serialization boundary (purely client-side UI prop). Replace with `adviseRename: boolean` — drop-in, identical behavior, one fewer named concept.

net: -1 type, ~-6 lines.

(One other candidate — inlining the single-use `_rename_target_has_files_detail` helper in `kb.py` — was considered and rejected: its dedicated unit test covers a real owner-id-leak regression that the e2e rename-conflict test does not reach, since that test exercises a different logical-name-conflict path.)

## Testing

- `pytest tests/web/api/test_kb_dir.py -k "ensure_collection_access or rename or check_documents"` → 18 passed
- `pytest tests/web_integration/test_kb_access_control_e2e.py` → 12 passed
- `npx tsc --noEmit` → clean
- `npx vitest run` on the three changed frontend test files → 22 passed, 1 failed (pre-existing, see finding 6)

Caveat on scope: repo-wide `pre-commit` (ruff/mypy under the pinned config) was **not** re-run in this pass, so lint/type coverage outside the files above is unverified by me.

No blocking issues — the round-3 Major is fixed and pinned by a test, and every other prior finding is fixed or deliberately waived. **Approve with nits** (findings 1-4 are cheap; 5 and 6 belong in follow-up issues).

bluefish-08:
Thanks for the blank-slate re-trace — that caught the `failedStatus` locals I had convinced myself were load-bearing.

**Findings 1, 3 and the simplification: done.**

1. All four `failedStatus` locals were indeed dead, and dropping them exposed that the whole two-value union was dead weight too. `KnowledgeBaseConflictAdvice` is gone — the field is now `adviseRename: boolean`, which is what it always was. The four detail-page sites pass `{ status: undefined, adviseRename: false }`; only the creation dialog, where the user just typed the name, passes `true`. One fewer named concept and one fewer thing to keep in sync, as you said.

3. Two unit tests added to `test_kb_dir.py`:
   - `..._allows_a_name_no_collection_holds` — the #1139 happy path. Worth having precisely because every other branch of the helper raises, so a regression that started raising here would still leave the suite green.
   - `..._denies_a_taken_name_without_naming_intent` — `allow_create=True, taken_name_is_conflict=False`, the combination `save_collection_config` uses. Verified by mutation: making the flag default to `allow_create` unconditionally fails this test and only this test.

**Finding 4:** you are right, and the count was stale in exactly the way you describe — "seven" was the pre-PR `allow_create=True` call count, and this PR moved two of them out of that bucket. PR description corrected to five, with the endpoints named.

**Finding 2:** not done. The rename dialog has no entry point (`handleOpenEditDialog` still has zero callers, as you confirmed in round 1), so a test there would pin a toast that nothing can currently produce. I would rather add it alongside restoring the entry point, where it can be written against a real flow. Flagging so it does not silently disappear — happy to add it now if you would rather have the coverage first.

**Findings 5 and 6:** agreed, both pre-existing and both worth their own issues rather than riding along here.

On your note that the cross-tenant wording rationale "lives one function away from the branch it governs" — fair. I left it in the shared helper because both callers need it, but a pointer from the branch would cost nothing.

```
pytest tests/web/api -k "kb or collection" + e2e     all passed
tsc --noEmit                                        clean
pre-commit (changed files)                          all passed
```

The two red frontend tests are the pre-existing ones from your finding 6 — unchanged by this push.

PR #200, opened 2026-03-22
fix: fix migrations for PostgreSQL and add integration tests

qinxuye:

PR #1204, opened 2026-08-08
fix(migrations): emit offline SQL for the checkpoint anchor migration

AlexLiu190625:
## What

`20260804_add_task_checkpoint_trace_event_anchor` cannot produce offline
SQL. All four cells fail today:

| | SQLite | PostgreSQL |
|---|---|---|
| `upgrade` | `NoInspectionAvailable` | `NoInspectionAvailable` |
| `downgrade` | `NoInspectionAvailable`, then `CommandError` from batch mode | `NoInspectionAvailable` |

The revision guards each DDL step behind `sa.inspect(op.get_bind())`.
Under `--sql` that bind is a `MockConnection`, which cannot be inspected.
The SQLite `downgrade` has a second, independent blocker: it rebuilds the
table through `batch_alter_table`, and batch mode refuses to run under
`--sql` because it needs a live connection to reflect from.

This adds an `if context.as_sql:` fork to both functions, matching the
shape already used by four other revisions in this directory
(`20260725_add_task_lease_recovery_index`,
`20260725_add_uploaded_file_recovery_index`,
`20260726_add_task_telegram_user_id`, `20260729_add_gmail_audience_grace`).
The online branch is unchanged.

Scope note: this makes the revision itself renderable offline; a full
`base:head --sql` render is still blocked by earlier revisions with the
same guard pattern (first blocker `441d4f5d399c`), which are out of scope
here.

Refs #1071.

## Why edit a merged revision instead of adding a corrective one

A later revision cannot fix this. `alembic upgrade head --sql` calls every
revision's `upgrade()` in sequence to collect statements; once this one
raises, generation stops and nothing after it is ever reached. A
corrective migration can fix data or table shape, but it cannot stop an
earlier `upgrade()` from raising.

Editing merged revisions for exactly this reason is established practice
here: #200 rewrote 23 already-merged revision files to make migrations run
on PostgreSQL, and #233 rewrote 15 more for the same class of failure.
Two of those files were amended twice. This change touches one file.

## Why the offline SQLite downgrade is a plain DROP COLUMN

The online SQLite branch rebuilds the whole table because SQLite renders
this column's foreign key inline in `CREATE TABLE` and then refuses to
drop a column an inline foreign key names. That inline foreign key only
exists on a database built by `Base.metadata.create_all()` -- new installs
and tests. A database built by this migration chain has no foreign key
here at all, as the revision's own docstring states.

Offline SQL is generated for a database maintained by the migration chain;
a `create_all`-built database is already at the latest shape and never
consumes it. So the offline branch is guaranteed to face the shape where a
plain `DROP COLUMN` is valid, while the online branch keeps the rebuild
because a live connection may be attached to either shape. The test for
that cell records this premise so the online branch is not simplified to
match.

## Verification

- four offline cells asserted on the emitted SQL, both dialects, both
  directions
- existing online tests unchanged and green on SQLite and PostgreSQL

### DIFF
--- a/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
+++ b/src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
@@ -8,17 +8,35 @@ CONSTRAINT`` is not renderable through Alembic's SQLite batch mode without a
 full table rebuild, so an upgraded SQLite database has no DB-level FK here.
 A database built fresh through ``Base.metadata.create_all()`` (new installs,
 tests) gets the FK from the model directly, regardless of dialect. That
-divergence between fresh and upgraded SQLite databases is permanent under
-the current migration set: this revision is the head and no later revision
-reconciles the two shapes. On an upgraded SQLite database the pointer's
-delete protection is therefore the application-level clearing order in
-``task_deletion.py``, which nulls both pointer columns before deleting the
-task's ``trace_events`` rows -- not the database.
+divergence between fresh and upgraded SQLite databases is permanent under the
+current migration set: no later revision reconciles the two shapes. On an
+upgraded SQLite database the pointer's delete protection is therefore the
+application-level clearing order in ``task_deletion.py``, which nulls both
+pointer columns before deleting the task's ``trace_events`` rows -- not the
+database.
 
 Existing rows are backfilled by resolving the legacy string column against
 the row it names, but only where that resolution is unambiguous: zero or
 multiple matching trace_events rows leave the new column NULL rather than
 guessing or aborting the migration.
+
+Both functions fork on ``context.as_sql``. Offline (``--sql``) generation runs
+against a MockConnection: nothing can be reflected, so the offline branch emits
+the DDL unconditionally instead of consulting the inspector, and the online
+branch keeps its existence guards. The two branches address different kinds of
+database and are each correct for their own.
+
+The offline SQLite downgrade drops the column directly rather than rebuilding
+the table. An offline script is generated for a database that this migration
+chain built, and on SQLite such a database carries no foreign key on this
+column -- as stated above, only a create_all-built database gets one, and those
+are already at the latest shape and never consume offline SQL. The online
+SQLite branch still rebuilds the table, because a live connection may be
+attached to either shape.
+
+The offline branch also has no counterpart to the online early returns for a
+missing ``tasks`` or ``trace_events`` table: an offline script targets a
+database maintained by this chain, where both tables are present.
 
 On PostgreSQL ``create_foreign_key`` takes an ACCESS EXCLUSIVE lock on
 ``tasks`` and validates every existing row before returning, and the
@@ -88,6 +106,18 @@ def _fk_names() -> set[str]:
 
 
 def upgrade() -> None:
+    context = op.get_context()
+
+    # Offline (--sql) generation runs against a MockConnection, so reflection
+    # is unavailable and every guard below would raise. Emit the
+    # unconditional DDL a migration-built database needs instead.
+    if context.as_sql:
+        op.add_column(TABLE, sa.Column(COLUMN, sa.Integer(), nullable=True))
+        if context.dialect.name == "postgresql":
+            op.create_foreign_key(FK_NAME, TABLE, TRACE_TABLE, [COLUMN], ["id"])
+        op.execute(BACKFILL_SQL)
+        return
+
     # tasks is created by Base.metadata.create_all() in production, not by
     # a migration in this repo; a from-scratch (bare) database has no tasks
     # table yet when migrations run, so this is a no-op there.
@@ -116,6 +146,18 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
+    context = op.get_context()
+
+    # An offline script targets a database this migration chain built, which
+    # on SQLite carries no foreign key on this column (see the module
+    # docstring) -- so a plain DROP COLUMN is enough, and batch mode is not
+    # available under --sql anyway. The online branch below still rebuilds.
+    if context.as_sql:
+        if context.dialect.name == "postgresql":
+            op.drop_constraint(FK_NAME, TABLE, type_="foreignkey")
+        op.drop_column(TABLE, COLUMN)
+        return
+
     if not _table_exists():
         return
 
--- a/tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
+++ b/tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
@@ -13,12 +13,15 @@ from __future__ import annotations
 
 import importlib.util
 import os
+from io import StringIO
 from pathlib import Path
 from types import ModuleType
 
 import pytest
 import sqlalchemy as sa
 from alembic import command
+from alembic.migration import MigrationContext
+from alembic.operations import Operations
 from sqlalchemy import create_engine, inspect, text
 
 from xagent.db.config import create_alembic_config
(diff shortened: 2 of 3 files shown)


src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -116,6 +146,18 @@ def upgrade() -> None:
...
+    context = op.get_context()
+
+    # An offline script targets a database this migration chain built, which
+    # on SQLite carries no foreign key on this column (see the module
+    # docstring) -- so a plain DROP COLUMN is enough, and batch mode is not
+    # available under --sql anyway. The online branch below still rebuilds.
+    if context.as_sql:
+        if context.dialect.name == "postgresql":
+            op.drop_constraint(FK_NAME, TABLE, type_="foreignkey")
+        op.drop_column(TABLE, COLUMN)
```
rogercloud, line 168: This plain `DROP COLUMN` is safe only under the docstring's premise that offline scripts target a database this migration chain built — but that is an operational convention, not something the code, CI, or docs enforce. A `create_all()`-built SQLite database can be stamped at this revision and still carries the inline FK on this column, and SQLite will reject the statement there (which is exactly why the online branch rebuilds via `batch_alter_table`). The failure is loud, so no code change is needed; consider naming the failure mode here or in the docstring, e.g. "if applied to a create_all-built SQLite database, SQLite will reject this statement because that database still carries this column's foreign key inline — such a database is already at the target schema and should never consume this offline script."

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -88,6 +106,18 @@ def _fk_names() -> set[str]:
 
 
 def upgrade() -> None:
+    context = op.get_context()
+
+    # Offline (--sql) generation runs against a MockConnection, so reflection
+    # is unavailable and every guard below would raise. Emit the
+    # unconditional DDL a migration-built database needs instead.
+    if context.as_sql:
+        op.add_column(TABLE, sa.Column(COLUMN, sa.Integer(), nullable=True))
```
rogercloud, line 125: Optional/style: this `sa.Column(COLUMN, sa.Integer(), nullable=True)` literal and the `create_foreign_key` argument list below are repeated verbatim in the online branch ~15 lines down. `20260729_add_gmail_audience_grace.py` avoids this by hoisting column specs into a module-level `COLUMNS` tuple shared by all call sites. Marginal at this size — the two branches differ in control flow by design — so take it or leave it.

tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -539,3 +542,79 @@ def test_downgrade_drops_the_named_foreign_key(self, postgres_engine) -> None:
...
+        opts={"as_sql": True, "output_buffer": output},
+    )
+
+    with Operations.context(context):
+        getattr(migration, operation)()
+
+    return output.getvalue()
+
+
+class TestOfflineSql:
```
rogercloud, line 563: For consistency with the other offline-SQL suites, consider asserting the rendered SQL carries no bind-parameter placeholders in each of the four tests: `assert "%(" not in sql`, `assert ":param" not in sql`, `assert "?" not in sql`. See `tests/alembic/test_20260721_drop_workforce_manager_instructions.py:183-208` for the established form. It is moot today since `BACKFILL_SQL` is a parameterless `sa.text()`, but it would catch a future change to real bind params, which would silently emit unrunnable offline SQL that the current substring assertions would not flag.

tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
```diff
@@ -539,3 +542,79 @@ def test_downgrade_drops_the_named_foreign_key(self, postgres_engine) -> None:
...
+    ) -> None:
+        sql = _offline_sql("postgresql", "downgrade")
+
+        drop_fk_at = sql.index(f"ALTER TABLE tasks DROP CONSTRAINT {FK_NAME}")
+        drop_col_at = sql.index(f"ALTER TABLE tasks DROP COLUMN {COLUMN}")
+        assert drop_fk_at < drop_col_at
+
+    @pytest.mark.parametrize("dialect", ["sqlite", "postgresql"])
+    @pytest.mark.parametrize("operation", ["upgrade", "downgrade"])
+    def test_offline_rendering_does_not_reflect(
```
rogercloud: Suggest deleting this test (and its two `parametrize` decorators). The four tests directly above call the same `_offline_sql` helper for the same four (dialect, operation) combinations and assert specific SQL substrings, which strictly subsumes a bare truthiness check. `_offline_sql` has no exception handling, so any reflection call against the offline mock bind would propagate and fail those four tests as errors rather than yielding an empty render — there is no failure this catches that they would miss.

rogercloud (approved):
## Summary

This fixes offline SQL generation for the `20260804_add_task_checkpoint_trace_event_anchor` revision. Every DDL step in `upgrade()`/`downgrade()` was guarded by `sa.inspect(op.get_bind())`, which raises `NoInspectionAvailable` against the `MockConnection` that `alembic ... --sql` uses; SQLite `downgrade` had a second blocker in that `batch_alter_table` requires a live connection and is refused under `--sql`. The PR adds an `if context.as_sql:` fork to both functions that emits the required DDL unconditionally, leaves the existing online branches byte-for-byte untouched, extends the module docstring to explain the offline/online asymmetry, and adds a `TestOfflineSql` class covering the four (dialect x operation) cells.

## Approach

**Sound.** The `as_sql` fork is the minimal fix and matches an established house pattern — 8 sibling revisions in `src/xagent/migrations/versions/` already use the same shape (the PR description cites 4, but the pattern count is higher). The online path is not modified, so there is no regression surface for live upgrades. DDL ordering was verified by executing the migration functions directly under a mock context: column-before-FK on upgrade and FK-before-column on downgrade, on both dialects. `BACKFILL_SQL` is a parameterless `sa.text()` with identifiers baked in, so it renders as literal SQL and is unaffected by the offline bind's lack of reflection.

### Scope note (non-blocking)

The PR's own scope note is accurate and honest, but worth restating at its true scale: of the 131 revision files in `src/xagent/migrations/versions/`, only 9 (including this one) guard reflection behind `as_sql`. Roughly 100 revisions still call `sa.inspect()` / `Inspector.from_engine` unconditionally and would raise under `--sql`, including the root revision (`down_revision = None`), which calls `Inspector.from_engine` with no guard at all. So a full `base:head --sql` render remains blocked by essentially the whole chain, not just one early revision. Nothing here needs to change in this PR — but since 9 revisions now hand-write the identical fork, it may be worth a follow-up discussion on whether a small shared helper is preferable to repeating this by hand across the remaining ~100.

## Findings

All findings below are suggestions or nits; none block merge.

### Suggestions

- **`src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py:158`** (and the docstring at lines 29-39) — the offline SQLite downgrade emits a plain `DROP COLUMN`. That is safe only under the docstring's premise that an offline script targets a database this migration chain built, which never carries the inline SQLite FK a `create_all()`-built database has. That premise is an operational convention, not a technical guarantee: a `create_all`-built SQLite database can be stamped at this exact revision (the test file's `create_all_engine` fixture demonstrates the inline FK), and nothing in code, CI, or docs prevents an operator from applying this script there — SQLite would reject the statement, which is precisely why the online branch does a full `batch_alter_table` rebuild. The failure is fail-loud (statement rejected, no silent corruption), so this is a documentation ask, not a code change: name the failure mode explicitly.

- **`tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py:563`** — none of the four `TestOfflineSql` tests assert the rendered SQL is free of bind-parameter placeholders, unlike the convention in sibling offline-SQL suites (e.g. `tests/alembic/test_20260721_drop_workforce_manager_instructions.py:183-208`, plus at least four other files). This is moot today because `BACKFILL_SQL` is parameterless, but the guard is three cheap lines and would catch a future change to real bind params, which would silently emit unrunnable offline SQL that the current substring assertions would not detect.

### Nits / optional

- **`src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py:115`** — the offline and online branches of `upgrade()` repeat the same `sa.Column(COLUMN, sa.Integer(), nullable=True)` literal and the same `create_foreign_key` argument list. A sibling revision (`20260729_add_gmail_audience_grace.py`) hoists column specs into a module-level `COLUMNS` tuple shared by all call sites. Marginal here — one literal repeated twice, ~15 lines apart, inside a single function whose two branches legitimately differ in control flow — so take it or leave it.

```
L614: redundant-test test_offline_rendering_does_not_reflect only asserts _offline_sql(dialect, operation).strip() is truthy for the same 4 (dialect, operation) cells the four tests directly above it already cover with specific SQL substring assertions — strictly stronger checks. _offline_sql has no exception handling, so a reflection call on the offline bind would propagate and fail those four tests as errors, not slip through as an empty render; there is no case this test catches that they miss. Delete the test, its docstring, and its two parametrize decorators.
net: -10 lines possible
```

## Test results

`pytest tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py -v` — 27 passed (23 SQLite/logic tests plus 4 PostgreSQL tests against a local PostgreSQL instance). Without `XAGENT_TEST_POSTGRES_URL` the 4 PostgreSQL tests skip and the remaining 23 still pass.

## Recommendation

**Approve with minor suggestions.** The fix is correct, minimal, well-tested, and does not touch the online path. The docstring clarification and the placeholder-free test guard are worth folding in; the redundant test is worth deleting; the rest is optional.

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Confirmed and worth stating. I built a `create_all` SQLite database, stamped it at this revision, and applied the offline downgrade statement to it. SQLite rejects it outright — `error in table tasks after drop column: unknown column "last_checkpoint_trace_event_id" in foreign key definition` — and leaves the schema and the rows exactly as they were.

One detail the failure mode has beyond that: SQLite scripts get no transaction wrapper from Alembic (`transactional_ddl` is false for this dialect), so running the generated script through `sqlite3 db < script.sql` reports the error and exits non-zero, but the trailing `UPDATE alembic_version` still executes. The database ends up stamped at the parent revision with the column still present, and the stamp has to be put back by hand.

Added to the docstring paragraph that already sets out the premise, including the recovery path (downgrade that shape over a live connection, which takes the rebuilding branch), rather than to the code comment — that comment already defers to the docstring.

tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Good catch, and the claim checks out. I replaced the backfill with three parameterized variants (a `sa.text` with a bound parameter, a Core `update().values()`, and a bare `:name` with no value) and re-rendered: every one produces `?` on SQLite and `%(name)s` on PostgreSQL, and every current assertion in these four tests still passes. The placeholder checks catch all three.

Two things for the record while adding them. `env.py` configures offline runs with `literal_binds=True`, and under that option all three variants render as inlined literals — so this assertion is stricter than the real `--sql` pipeline. That is the useful reading (no bind parameters reach the renderer at all), and it only works because the test helper renders without `literal_binds`; the helper should stay that way, and the new test's docstring says so. And `":param" not in sql` cannot fire here, since the helper's dialects use `qmark` and `pyformat` — it is free and matches the sibling suites, so it stays, but it is not what does the work.

Rather than repeating three lines in each of the four tests, they land as one parametrized test over the same four cells, reusing the decorators freed up by removing the truthiness test from the other thread.

src/xagent/migrations/versions/20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Leaving it. The shared material is the column literal and the foreign-key arguments, and both are duplicated between the offline branch and the online one — hoisting them means editing the online call sites too. Keeping the online branch untouched is the property this change is built on, and it is what makes the diff verifiable at a glance; two deduplicated literals are not worth giving that up.

For what it is worth, I did check the reuse question on the pattern you pointed at: passing the same `sa.Column` instance to `op.add_column` more than once in a process is safe — Alembic copies it. So the hesitation is purely about which lines the edit would touch, not about the pattern itself.

tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py
AlexLiu190625: Agreed, and I checked the subsumption rather than assuming it. I ran the four substring tests and this one against six mutants of the migration — the offline fork removed from `upgrade`, removed from `downgrade`, the downgrade switched to a batch rebuild, the foreign key emitted on every dialect, the backfill dropped, and both offline branches emitting nothing. Every mutant that reddens this test also reddens the corresponding substring test, and one (backfill dropped) reddens only the substring tests. The empty-but-successful render is covered too: under that mutant all four cells render an empty string, which fails the substring assertions.

Removed, with its two parametrize decorators reused for the placeholder-free assertions from the other thread — same four cells, stronger checks. The reasoning about why a non-empty render proves no guard was consulted stays in the class docstring.

PR #1206, opened 2026-08-08
feat(task-lease): stamp a per-acquisition attempt identity on the task row

### LINKED ISSUE #1074: feat(interactions): add canonical task interaction schema
## Summary

## Current behavior

## Acceptance criteria

## Dependencies

## Non-goals

- Attendance fields.

AlexLiu190625:
## What

Adds a per-acquisition attempt identity to the task lease:

- `tasks.lease_attempt_id` — a bare nullable `String(64)`. No foreign key,
  no check constraint, no unique constraint, no index.
- `TaskLease.attempt_id` — a new trailing field on the frozen dataclass,
  defaulting to `None`.
- `acquire_task_lease_no_commit` mints a fresh uuid on every successful
  claim, writes it in the same conditional UPDATE, and returns it.
- Every writer that clears `runner_id` — release, expiry recovery,
  fail-and-release, release-current-runner, the A2A cancel finalization,
  and the turn claim — clears this column in the same statement, so it
  never outlives the attempt that wrote it. A static test enumerates the
  update carriers and fails on any site that writes one column without
  the other.
- Its own alembic migration, forked on `context.as_sql`.

## Why

This is the closure artifact for the stale-attempt fencing finding (BL-2)
raised against the interaction-handoff design. It carries no acceptance
criterion of its own from #1074, and the delivery series this PR belongs to
does not close #1074 either — the existing-table pairs and the datetime
read-back half remain on the issue.

The problem it addresses: a runner holding a `(runner_id, run_id)` tuple
cannot today prove that tuple still names *its* attempt. `state_version`
cannot serve as that proof — it is a task-lifecycle counter, not an attempt
counter. `_apply_pause_requested_isolated` increments it mid-run without
changing attempt ownership, so a fence pinned to the value seen at
acquisition rejects a legitimate finalizer that was merely paused. An
attempt identity minted per claim and touched by no other writer is exactly
the missing fact.

The margin over the existing fence is narrow but real, and worth stating so
it is not later removed as redundant: both online finalizers already compare
`runner_id` and `run_id`, and expiry recovery NULLs `runner_id`, so a bare
reaper recovery is already blocked. The attempt identity earns its keep in
one shape only — a successor run reacquiring the *same* `(runner_id, run_id)`
tuple, where every existing predicate matches for the stale holder too.

Scope note: the stale-attempt window this fences is a pre-existing defect,
reachable today. This PR does not fix it for the existing finalizers — their
fences are unchanged. It supplies the fact a future fence needs, and
guarantees the handoff path will not add a new stale writer on top of it.

## Compatibility window

`lease.attempt_id is None` means "cannot prove attempt identity". Consumers
must fall back to whatever fence they already have — never treat None as a
match. Two sources:

1. A lease claimed by a worker running code from before this column existed.
   This is a rolling-restart window that closes on its own. During it, the
   run-scoped identity key does not back the gap either, so both protections
   are down for the same interval.
2. `_task_lease_snapshot` in `websocket.py`, the second `TaskLease`
   construction point, which rebuilds an ambient lease from an already-loaded
   task row. It carries `None` permanently and deliberately: every field it
   sets is read from that row, so populating `attempt_id` from the same read
   would make a later `task.lease_attempt_id != lease.attempt_id` check
   compare a value against itself — an always-passing fence, strictly worse
   than an explicit sentinel a caller can detect and skip. Any future work
   routing this ambient lease into an attempt check must revisit this.

## Test changes

Exactly one existing assertion changes. `TaskLease` is a frozen dataclass,
so its generated `__eq__` compares every field, and
`test_same_run_resume_lease_preserves_checkpoint_pointer` asserts whole-object
equality. Its expectation now carries `attempt_id=lease.attempt_id`,
declaring the new field as not part of what that test pins. This is the only
whole-object `TaskLease` comparison in the test tree; every other lease test
is unchanged.

## Verification

- lease suites after the change: 127 passed across the six lease test
  files; with the new tests, 135 passed.
- full regression: six lease files plus `tests/alembic/` — 356 passed,
  1 pre-existing unrelated skip. Downstream `TaskLease` consumers
  (websocket, execution-scope, a2a, task-command transport) — 222 passed
  with PostgreSQL configured.
- migration: 10 new tests green; single alembic head before and after.

Migration integration baseline, recorded so this PR's migration tests are not
confounded with it: `test_postgresql_upgrade` and
`test_postgresql_idempotence_with_sqlalchemy` are red at the branch point
(`d97f440a`) and remain red, with the same traceback — the suite's teardown
issues `DROP TABLE mcp_servers_legacy` without CASCADE while four foreign keys
still reference it. Unrelated to this change and not fixed here.
`test_postgresql_incremental_upgrade` is green before and after.

This PR's own migration is covered by `tests/alembic/` against SQLite, plus a
one-off PostgreSQL run against a throwaway schema (online add / idempotent
re-run / downgrade / schema resolution).

### DIFF
--- a/src/xagent/migrations/versions/20260808_add_task_lease_attempt_id.py
+++ b/src/xagent/migrations/versions/20260808_add_task_lease_attempt_id.py
@@ -1,0 +1,99 @@
+"""add lease attempt identity to tasks
+
+Revision ID: 20260808_add_task_lease_attempt_id
+Revises: 20260807_seed_notion_mcp_app
+Create Date: 2026-08-08
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "20260808_add_task_lease_attempt_id"
+down_revision: Union[str, None] = "20260807_seed_notion_mcp_app"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+TABLE = "tasks"
+COLUMN = "lease_attempt_id"
+
+# The schema of the *visible* tasks relation. version_table_schema names only
+# the Alembic version table, and current_schema() is merely the first entry on
+# search_path, so neither identifies the relation an unqualified reference
+# actually resolves to. Ask PostgreSQL which one it resolves.
+POSTGRES_VISIBLE_TABLE_SCHEMA_SQL = sa.text(
+    """
+    SELECT ns.nspname
+    FROM pg_catalog.pg_class AS cls
+    JOIN pg_catalog.pg_namespace AS ns ON ns.oid = cls.relnamespace
+    WHERE cls.oid = pg_catalog.to_regclass(:table_name)
+    """
+)
+
+
+def _target_schema() -> str | None:
+    """The schema holding the tasks relation this migration operates on."""
+    bind = op.get_bind()
+    if bind.dialect.name == "postgresql":
+        resolved = bind.execute(
+            POSTGRES_VISIBLE_TABLE_SCHEMA_SQL, {"table_name": TABLE}
+        ).scalar()
+        if resolved:
+            return str(resolved)
+    schema = op.get_context().version_table_schema
+    return str(schema) if schema else None
+
+
+def _online_columns(schema: str | None) -> set[str]:
+    inspector = sa.inspect(op.get_bind())
+    if TABLE not in inspector.get_table_names(schema=schema):
+        return set()
+    return {str(item["name"]) for item in inspector.get_columns(TABLE, schema=schema)}
+
+
+def _online_table_exists(schema: str | None) -> bool:
+    return TABLE in sa.inspect(op.get_bind()).get_table_names(schema=schema)
+
+
+def upgrade() -> None:
+    context = op.get_context()
+
+    # Offline (--sql) generation has a MockConnection, so reflection is
+    # unavailable. Emit the unconditional DDL instead of inspecting. This is
+    # the 20260726 shape, not the inspector-only shape #1137 used -- the
+    # latter raises under --sql on both dialects.
+    if context.as_sql:
+        op.add_column(TABLE, sa.Column(COLUMN, sa.String(length=64), nullable=True))
+        return
+
+    # Address the same relation the catalog lookup inspects, so reflection and
+    # DDL can never diverge onto different schemas.
+    schema = _target_schema()
+
+    if not _online_table_exists(schema):
+        return
+
+    if COLUMN not in _online_columns(schema):
+        op.add_column(
+            TABLE,
+            sa.Column(COLUMN, sa.String(length=64), nullable=True),
+            schema=schema,
+        )
+
+
+def downgrade() -> None:
+    context = op.get_context()
+
+    if context.as_sql:
+        op.drop_column(TABLE, COLUMN)
+        return
+
+    schema = _target_schema()
+
+    if not _online_table_exists(schema):
+        return
+
+    if COLUMN in _online_columns(schema):
+        op.drop_column(TABLE, COLUMN, schema=schema)
--- a/src/xagent/web/models/task.py
+++ b/src/xagent/web/models/task.py
@@ -134,6 +134,17 @@ class Task(Base):  # type: ignore
     control_state = Column(
         String(32), nullable=False, default="idle", server_default="idle"
     )
+    # Identity of the lease attempt that currently owns this row. Written
+    # only by acquire_task_lease_no_commit (a fresh uuid per claim) and
+    # cleared by the lease release/recovery writers. A non-NULL value does
+    # not mean that attempt is still running -- it means no writer has
+    # cleared it since. Readers that need liveness must consult runner_id
+    # and lease_expires_at, not this column.
+    #
+    # Deliberately a bare nullable column: no FK, no CHECK, no UNIQUE. It is
+    # never a WHERE predicate -- consumers read the row by primary key and
+    # compare in Python -- so an index would be pure write cost.
+    lease_attempt_id = Column(String(64), nullable=True)
 
     # Model configuration
     model_name = Column(String(255), nullable=True)  # Main model used for the task
--- a/src/xagent/web/services/task_lease_service.py
+++ b/src/xagent/web/services/task_lease_service.py
@@ -50,9 +50,35 @@ def _rowcount(result: Any) -> int:
 
 @dataclass(frozen=True)
 class TaskLease:
+    """One runner's claim on one task row.
+
+    ``attempt_id`` names the exact acquisition that produced this lease. It
+    is minted fresh by ``acquire_task_lease_no_commit`` on every successful
+    claim and mirrored into ``tasks.lease_attempt_id`` in the same UPDATE, so
+    a holder can later prove the row still belongs to *its* attempt rather
+    than to a successor that reused the same ``(runner_id, run_id)`` tuple.
+
+    ``attempt_id is None`` is a fail-open sentinel with two distinct sources:
+
+    * a lease acquired by a worker running code from before this column
+      existed -- a rolling-restart window that closes on its own once every
+      worker has restarted; and
+    * ``_task_lease_snapshot`` in ``websocket.py``, which reconstructs an
+      ambient lease from an already-loaded task row. That one carries None
+      permanently and on purpose: every one of its fields is read from the
+      task row, so populating ``attempt_id`` from the same row would make any
+      later ``task.lease_attempt_id != lease.attempt_id`` check compare a
+      value against itself -- an always-passing fence, which is strictly
+      worse than an explicit None that callers can detect and skip.
+
+    Consumers must therefore treat None as "cannot prove attempt identity"
+    and fall back to whatever fence they already had, never as "matches".
+    """
+
     task_id: int
     runner_id: str
     run_id: str | None = None
+    attempt_id: str | None = None
 
 
 _CURRENT_TASK_LEASE: ContextVar[TaskLease | None] = ContextVar(
@@ -567,6 +593,11 @@ def recover_expired_task_lease_no_commit(
         "control_state": control_state_for_status(status).value,
         "state_version": func.coalesce(Task.state_version, 0) + 1,
         "error_message": error_message,
+        # Defence in depth: the column must never outlive the attempt that
+        # wrote it. This writer already NULLs runner_id, so no live fence
+        # depends on the value -- clearing it keeps the column honest for
+        # anyone who reads it directly.
+        "lease_attempt_id": None,
     }
     if status == TaskStatus.FAILED:
         values["output"] = None
@@ -647,6 +678,7 @@ def acquire_task_lease_no_commit(
     now = utc_now()
     expires_at = task_lease_expires_at(now)
     candidate_run_id = expected_run_id or str(uuid.uuid4())
+    attempt_id = str(uuid.uuid4())
     current_version = func.coalesce(Task.state_version, 0)
     running_control_state = control_state_for_status(TaskStatus.RUNNING).value
     values: dict[str, Any] = {
@@ -661,6 +693,14 @@ def acquire_task_lease_no_commit(
         "state_version": lease_state_version_case(
             TaskStatus.RUNNING, running_control_state, current_version
         ),
+        # A fresh identity for this exact claim. Every successful acquisition
+        # mints a new one; no other writer ever assigns a value here (the
+        # release/recovery writers only clear it). That is precisely what
+        # state_version cannot offer -- _apply_pause_requested_isolated bumps
+        # state_version mid-run without changing attempt ownership, so a
+        # state_version fence pinned at acquisition would reject a legitimate
+        # finalizer that had been paused.
+        "lease_attempt_id": attempt_id,
     }
     if new_run:
         values["last_checkpoint_event_id"] = None
@@ -701,6 +741,7 @@ def acquire_task_lease_no_commit(
         task_id=task_id,
         runner_id=runner,
         run_id=str(stored_run_id) if stored_run_id is not None else None,
+        attempt_id=values["lease_attempt_id"],
     )
 
 
@@ -883,6 +924,7 @@ def release_task_lease_no_commit(
         "state_version": lease_state_version_case(
             status, control_state, current_version
         ),
+        "lease_attempt_id": None,
     }
     if status in _NON_TERMINAL_RELEASE_STATUSES:
         values["error_message"] = None
@@ -930,6 +972,7 @@ def fail_and_release_task_lease_no_commit(
             state_version=func.coalesce(Task.state_version, 0) + 1,
             error_message=error_message,
             output=None,
+            lease_attempt_id=None,
         )
     )
     result = db.execute(stmt.execution_options(synchronize_session=False))
@@ -963,6 +1006,7 @@ def release_current_runner_task_lease(
             state_version=lease_state_version_case(
                 status, control_state, current_version
             ),
+            lease_attempt_id=None,
         )
     )
     if expected_run_id is not None:
--- a/src/xagent/web/services/task_lease_service.py
+++ b/src/xagent/web/services/task_lease_service.py
@@ -69,7 +69,10 @@ class TaskLease:
       task row, so populating ``attempt_id`` from the same row would make any
       later ``task.lease_attempt_id != lease.attempt_id`` check compare a
       value against itself -- an always-passing fence, which is strictly
-      worse than an explicit None that callers can detect and skip.
+      worse than an explicit None that callers can detect and skip. Today
+      that snapshot only feeds checkpoint fencing via
+      ``bind_task_lease_context`` and never reaches a settlement path, so
+      the permanent None also has no behavioural cost.
 
     Consumers must therefore treat None as "cannot prove attempt identity"
     and fall back to whatever fence they already had, never as "matches".
--- a/src/xagent/web/services/task_lease_service.py
+++ b/src/xagent/web/services/task_lease_service.py
@@ -76,6 +76,13 @@ class TaskLease:
 
     Consumers must therefore treat None as "cannot prove attempt identity"
     and fall back to whatever fence they already had, never as "matches".
+
+    The existing lease writers (refresh, release, fail-and-release) do not
+    fence on this column yet, deliberately: rejecting a stale attempt there
+    turns its refresh into a LOST classification and an active cancellation,
+    which is a behavior change that belongs to the first consumer of this
+    column, with its own tests. Until then the pre-existing
+    ``(runner_id, run_id)`` fences are the only guards on those paths.
     """
 
     task_id: int
--- a/tests/alembic/test_20260808_add_task_lease_attempt_id.py
+++ b/tests/alembic/test_20260808_add_task_lease_attempt_id.py
@@ -3,7 +3,7 @@
 import importlib.util
 from io import StringIO
 from pathlib import Path
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 import sqlalchemy as sa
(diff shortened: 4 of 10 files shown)


src/xagent/web/services/task_lease_service.py
```diff
@@ -50,9 +50,38 @@ def _rowcount(result: Any) -> int:
...
+      the permanent None also has no behavioural cost.
+
+    Consumers must therefore treat None as "cannot prove attempt identity"
+    and fall back to whatever fence they already had, never as "matches".
+    """
+
     task_id: int
     runner_id: str
     run_id: str | None = None
+    attempt_id: str | None = None
```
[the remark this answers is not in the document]
AlexLiu190625: Confirmed the mechanism. `_refresh_task_lease_no_commit`, `release_task_lease_no_commit`, and `fail_and_release_task_lease_no_commit` fence on `(runner_id, run_id)` plus status, and a same-process re-acquisition with `expected_run_id` reproduces that tuple exactly — `runner_id` is a per-process constant and `lease_run_id_case` preserves the run. I reproduced all three: a superseded lease refreshed, released, and failed the successor's row.

This PR does not create or widen that window; it changes no WHERE clause. Adding the predicate here is a behavior change rather than a completion of this one: with the fence in place, a stale attempt's refresh matches zero rows, the classification query returns `LOST`, and `run_while_task_lease_owned` raises `TaskLeaseLostError` and cancels the run. Turning a silently-ineffective stale writer into an actively-cancelled one needs its own tests and its own review, so it lands with the consumer that reads this column. The `TaskLease` docstring now records this deferral explicitly.

Two notes for whoever writes that fence. `Task.lease_attempt_id == lease.attempt_id` folds to `lease_attempt_id IS NULL` whenever the operand is `None`, so it must be gated on `lease.attempt_id is not None` rather than applied unconditionally — otherwise the one permanent-`None` producer, the ambient lease reconstructed in `_task_lease_snapshot`, is rejected against any stamped row. And the classification query in `_refresh_task_lease_no_commit` should be reviewed alongside the UPDATE, since it decides `LOST` vs `SETTLEMENT_READY`.

Separately, `_task_lease_key` is `(task_id, runner_id, run_id)` and the heartbeat manager refcounts entries by that key, so two same-run attempts also share one heartbeat registration. That belongs in the same follow-up.

tests/web/test_task_lease_service.py
```diff
@@ -1257,6 +1258,9 @@ def test_same_run_resume_lease_preserves_checkpoint_pointer(db_session) -> None:
         task_id=int(task.id),
         runner_id="resume-runner",
         run_id="resumable-run",
+        # Not part of what this test pins: a fresh attempt id per claim is
+        # acquire's contract, exercised by the stamping tests below.
+        attempt_id=lease.attempt_id,
```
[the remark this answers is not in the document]
AlexLiu190625: Agreed, fixed. Confirmed both halves: with `lease` as `None`, the previous form raised `AttributeError: 'NoneType' object has no attribute 'attempt_id'` because the constructor arguments evaluate before the comparison, whereas the pre-change form degraded to a plain assertion failure with a readable diff.

The guard is also this file's existing convention well before this change — `assert lease is not None` right after `acquire_task_lease` appears at nine sites on the base commit — and this test was the only acquire-then-dereference site missing it. Checked the rest of the diff for the same shape: every test added here already guards before dereferencing, and this is the only whole-object `TaskLease` comparison in the tests tree, so the one-line addition is the complete fix.

src/xagent/migrations/versions/20260808_add_task_lease_attempt_id.py
```diff
@@ -0,0 +1,99 @@
...
+            return str(resolved)
+    schema = op.get_context().version_table_schema
+    return str(schema) if schema else None
+
+
+def _online_columns(schema: str | None) -> set[str]:
+    inspector = sa.inspect(op.get_bind())
+    if TABLE not in inspector.get_table_names(schema=schema):
+        return set()
+    return {str(item["name"]) for item in inspector.get_columns(TABLE, schema=schema)}
```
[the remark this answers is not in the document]
AlexLiu190625: Confirmed the double lookup is real: on the online path `upgrade()` and `downgrade()` both call `_online_table_exists(schema)` first, so the `get_table_names` call inside `_online_columns` never decides anything. Measured on SQLite, the upgrade path issues the table listing twice instead of once — one extra catalog query per migration run.

I'd like to keep it anyway, for two reasons.

The check is what makes `_online_columns` total. With it removed, calling the helper against a database where `tasks` is absent raises `NoSuchTableError` on both dialects — SQLite via reflection, PostgreSQL via `get_table_oid` — rather than returning an empty set. That matters here because `tasks` is created by `create_all()`, not by a migration, so a partial database is a supported input and the migration is meant to no-op on it. Leaving the helper safe on its own means a future call site added outside the guard degrades to a skip instead of aborting `alembic upgrade head`. The current tests wouldn't catch that regression, since the missing-table test returns at the outer guard before the helper is reached.

Second, this file deliberately mirrors `20260726_add_task_telegram_user_id.py`, whose `_online_columns` is byte-identical including this check. Keeping the two the same means a reader diffing them sees only the intended differences — no index, different column width — instead of a helper that quietly diverged.

The cost is one catalog query, executed once per deployment, next to an `ALTER TABLE` on the same connection. That trade seems worth the safety and the parity, so I'd rather not change it here. If we do want to collapse the two checks later, the right move is dropping the outer `_online_table_exists` and keying off the empty column set — that's what most migrations in this directory already do, and it saves the same query while keeping the helper safe.

src/xagent/web/models/task.py
```diff
@@ -134,6 +134,17 @@ class Task(Base):  # type: ignore
...
+    # only by acquire_task_lease_no_commit (a fresh uuid per claim) and
+    # cleared by the lease release/recovery writers. A non-NULL value does
+    # not mean that attempt is still running -- it means no writer has
+    # cleared it since. Readers that need liveness must consult runner_id
+    # and lease_expires_at, not this column.
+    #
+    # Deliberately a bare nullable column: no FK, no CHECK, no UNIQUE. It is
+    # never a WHERE predicate -- consumers read the row by primary key and
+    # compare in Python -- so an index would be pure write cost.
+    lease_attempt_id = Column(String(64), nullable=True)
```
rogercloud, line 147: **Must fix before merge.** There is a fifth clearing writer that this column's own invariant (and the PR description's "all four writers that clear runner_id clear this column in the same statement") does not cover: the A2A cancel-finalization UPDATE in `src/xagent/web/api/a2a.py` sets `runner_id=None` and `lease_expires_at=None` but never touches `lease_attempt_id`.

It is reachable against a live task: its guard (`direct_cancel`) only requires a non-terminal status, and the path runs from the `cancel_task` A2A endpoint through `_cancel_task_unserialized` to `_finalize_a2a_cancel_sync` -- not dead code. The result is a row with `runner_id IS NULL` but a non-NULL `lease_attempt_id` left over from an attempt that is definitively gone.

No runtime impact today, since nothing reads the column for fencing yet. But it contradicts the invariant this column's docstring states, and the first consumer that trusts the invariant will inherit a stale identity on exactly the cancel path where fencing matters most. Adding `lease_attempt_id=None` to that `.values()` block is a one-line fix and is much cheaper now than after a consumer exists.

src/xagent/web/services/task_lease_service.py
```diff
@@ -50,9 +50,45 @@ def _rowcount(result: Any) -> int:
...
+    claim and mirrored into ``tasks.lease_attempt_id`` in the same UPDATE, so
+    a holder can later prove the row still belongs to *its* attempt rather
+    than to a successor that reused the same ``(runner_id, run_id)`` tuple.
+
+    ``attempt_id is None`` is a fail-open sentinel with two distinct sources:
+
+    * a lease acquired by a worker running code from before this column
+      existed -- a rolling-restart window that closes on its own once every
+      worker has restarted; and
+    * ``_task_lease_snapshot`` in ``websocket.py``, which reconstructs an
```
rogercloud, line 66: Minor: this paragraph is the only place the permanent `None` produced by `_task_lease_snapshot` (in `websocket.py`) is explained, and it lives in a different module from the code it describes. `_task_lease_snapshot` builds a `TaskLease` from an already-loaded task row and simply omits `attempt_id`, relying on the dataclass default.

Before this PR that omission was invisible. Now it reads as the obviously-missing kwarg, and the next person to touch that function will be tempted to pass `task.lease_attempt_id`, which is precisely the self-comparison this docstring warns produces an always-passing fence.

Two cheap safeguards: a one-line comment at the `TaskLease(...)` construction in `websocket.py` pointing back here, and a test asserting `_task_lease_snapshot(...).attempt_id is None` for a row that has a non-NULL `lease_attempt_id`. There is currently no coverage of that property at all.

src/xagent/web/services/task_lease_service.py
```diff
@@ -883,6 +934,7 @@ def release_task_lease_no_commit(
         "state_version": lease_state_version_case(
             status, control_state, current_version
         ),
+        "lease_attempt_id": None,
```
rogercloud, line 937: Minor: `lease_attempt_id=None` is now hand-copied into five separate writers (this one, `recover_expired_task_lease_no_commit`, `fail_and_release_task_lease_no_commit`, `release_current_runner_task_lease`, and -- per the comment on `task.py` -- one in `a2a.py` that is currently missing it).

This file already solved this exact problem once: `_checkpoint_pointer_clearing_predicate()` was extracted specifically so two SET builders could never drift on which rows clear their pointer, and `test_lease_checkpoint_pointer_case_predicates_match` exists because, in its own words, per-site tests would not catch that drift.

The same reasoning applies here, and with more force now that the clearing sites span two modules. Consider either a small shared constant/helper for the clearing values, or a single test that enumerates the lease-clearing writers and asserts each one nulls `lease_attempt_id`. Either would have caught the missing `a2a.py` site before review.

tests/alembic/test_20260808_add_task_lease_attempt_id.py
```diff
@@ -0,0 +1,159 @@
...
+    metadata.create_all(engine)
+
+
+def _column_names(connection) -> set[str]:
+    return {column["name"] for column in sa.inspect(connection).get_columns(TABLE)}
+
+
+def test_online_upgrade_adds_a_nullable_column_and_downgrade_removes_it() -> None:
+    migration = _migration_module()
+    engine = sa.create_engine("sqlite:///:memory:")
```
rogercloud, line 65: Minor: `_target_schema()`'s PostgreSQL branch (the `POSTGRES_VISIBLE_TABLE_SCHEMA_SQL` catalog lookup) has no coverage in this file, directly or indirectly -- every online test here runs against SQLite, and the offline tests return early via `context.as_sql` before `_target_schema()` is ever called. So the one piece of dialect-specific SQL in the migration is the one piece never executed by the suite.

The sibling migration this was modelled on, `20260726_add_task_telegram_user_id.py`, has `test_target_schema_resolves_the_visible_tasks_relation` covering the identical SQL. Adapting it here should be close to a copy-paste and would keep the two migration test files parallel.

src/xagent/web/services/task_lease_service.py
```diff
@@ -701,6 +751,7 @@ def acquire_task_lease_no_commit(
         task_id=task_id,
         runner_id=runner,
         run_id=str(stored_run_id) if stored_run_id is not None else None,
+        attempt_id=values["lease_attempt_id"],
```
rogercloud, line 754: Nit: `attempt_id` is already bound as a `str` local a few lines up, so reading it back out of the `dict[str, Any]` launders a known type through `Any` and gives the type checker nothing to verify against `TaskLease.attempt_id: str | None`. `attempt_id=attempt_id` is shorter, equivalent (the key is never reassigned between the two points), and type-checked.

rogercloud (approved):
## Summary

This PR adds a `tasks.lease_attempt_id` column (nullable `String(64)`, plus an Alembic migration) and a matching `TaskLease.attempt_id` field, minted as a fresh UUID by `acquire_task_lease_no_commit` on every successful claim and cleared in the same `UPDATE` by the release/recovery writers. It closes a real gap: a runner holding `(runner_id, run_id)` currently cannot prove that tuple still names *its own* acquisition — a successor that reacquires the identical tuple satisfies every existing fence predicate exactly as well as the stale holder — and `state_version` cannot substitute, because `_apply_pause_requested_isolated` bumps it mid-run without any change of ownership. Nothing reads the column for fencing yet; that is a deliberate, documented deferral to a follow-up consumer PR, so the change is inert at runtime today.

## ⚠️ Must fix before merge

**The PR's central claim — "all four writers that clear `runner_id` clear this column in the same statement" — is not true. There is a fifth writer that misses it, in `src/xagent/web/api/a2a.py` (A2A cancel finalization).** See the inline comment on `src/xagent/web/models/task.py` for details. This directly contradicts the invariant the new column's own docstring asserts, is reachable in production (not dead code), and — while inert today because no consumer reads the column yet — will silently corrupt fencing semantics for the first future consumer that trusts the invariant this PR documents. Please add `lease_attempt_id=None` to that writer's `.values()` before merging.

The diagnosis is correct and the chosen fact (a per-claim UUID written in the same atomic statement as the rest of the claim) is the right primitive. The column is deliberately bare — no FK, no CHECK, no UNIQUE, no index — with a clear rationale, and the `TaskLease` docstring is unusually good about what `None` means and why the existing refresh/release/fail paths do not fence on it yet.

Beyond the must-fix item above, two related design observations (see inline comments):

- There is no shared "clear the lease identity" helper and no drift test across the clearing sites — this is the design gap that let the `a2a.py` miss slip through, and it runs against this same file's own established convention (`_checkpoint_pointer_clearing_predicate()` + `test_lease_checkpoint_pointer_case_predicates_match`, extracted for exactly this reason).
- The one intentionally-`None` producer (`_task_lease_snapshot` in `websocket.py`) has no marker at its own call site — the explanation lives only in a different module's docstring, so the omission now reads as a forgotten kwarg rather than a deliberate choice.

Non-blocking design note, no action required: a monotonic `lease_epoch` integer via the same `coalesce(..., 0) + 1` idiom already used a few lines away for `state_version`, read back through the existing `.returning()`, would cost 4 bytes rather than a 64-char string and would be no harder to migrate or write. Not claiming the UUID is wrong — the only sketched consumer wants equality, not ordering — just noting the alternative wasn't discussed.

For context: an earlier concern of mine (whether the `runner_id == runner` disjunct in `acquire_task_lease_no_commit` could let a live holder's own process rotate its `attempt_id` out from under itself) was traced to completion and fully refuted — every acquire call site is guarded by an additional `status != RUNNING` predicate, a same-transaction status proof, or in-process serialization. No action needed there.

## Re-review of prior threads

The three existing review threads were independently reverified against the current code (re-checked in the tree, not merely re-read):

1. **Fixed** — the missing `assert lease is not None` guard before a `lease.attempt_id` dereference is present at the cited test, matches this file's pre-existing convention, and the other eight new tests in this diff already carried equivalent guards.
2. **Waived, correctly** — deferring the `attempt_id` fence in `_refresh_task_lease_no_commit` / `release_task_lease_no_commit` / `fail_and_release_task_lease_no_commit` is deliberate: no WHERE clause was touched in any of the three, the deferral is documented, and an unconditional fence really would fold to `IS NULL` for the permanently-`None` ambient snapshot lease. Sound scoping — though see the inline comment about that snapshot's own call site still carrying no local marker.
3. **Waived, correctly** — the duplicated table-existence check in the migration's `_online_columns` keeps that helper safe against a partially-migrated database and keeps it byte-identical to its sibling migration for diffability. One extra catalog query per run is a reasonable trade.

## One more nit (not inlined — the touched line falls outside this PR's diff)

`test_old_lease_cannot_refresh_or_release_a_new_run` (`tests/web/test_task_lease_service.py`) already builds a stale-lease-vs-successor scenario but doesn't assert that the successor's `lease_attempt_id` survives the stale caller's rejected release. This is largely a mechanical corollary of the same atomic-UPDATE-with-WHERE-clause guarantee that test already exercises, so it's optional belt-and-suspenders, not a real gap.

## Test results

- `tests/web/test_task_lease_service.py` + `tests/alembic/test_20260808_add_task_lease_attempt_id.py`: **56 passed**.
- `ruff format --check`: changed source files already formatted.
- Limitation: mypy and the pinned ruff lint config could not be run in the review environment (unrelated tooling issue) — unverified rather than passing.

Lean already — no over-engineering found in this diff.

src/xagent/web/models/task.py
AlexLiu190625: Confirmed, and the sweep turned up one more. `_finalize_a2a_cancel_sync` nulls `runner_id`, `lease_expires_at`, and `last_heartbeat_at` but left `lease_attempt_id` set, so a canceled task kept an attempt id with no owning attempt. Fixed by adding `lease_attempt_id=None` to that `.values(...)`.

A full AST sweep of `src/` for Task-row writers that touch `runner_id` in any carrier form found seven, not the four the description claimed. Besides this one, `_claim_turn_no_commit` in `task_orchestrator.py` also nulls `runner_id` without clearing the attempt id. That one cannot produce a stale committed row today — it calls `acquire_task_lease_no_commit` unconditionally in the same transaction, which re-stamps a fresh value, and any failure rolls the whole claim back — but it is cleared there too, so the invariant is carrier-local and needs no exemption.

Nothing reads the column yet, so clearing it cannot affect any consumer. The description's writer list is corrected, and a static test now enumerates the pairing.

src/xagent/web/services/task_lease_service.py
AlexLiu190625: Added, both parts. `_task_lease_snapshot` now carries a comment at the construction site explaining why `attempt_id` stays at its default and pointing at the `TaskLease` docstring, and there is a test asserting the snapshot returns `attempt_id is None` for a row whose `lease_attempt_id` is set. The function reads its input with `getattr`, so the test needs no database — a plain stub object is enough.

src/xagent/web/services/task_lease_service.py
AlexLiu190625: Agreed, and this is the guard that would have caught the A2A miss. Added a static enumeration test modeled on `test_checkpoint_pointer_pairing.py`: a statement that writes `runner_id` into a Task update carrier must write `lease_attempt_id` in the same carrier, over `task_lease_service`, `task_orchestrator`, and `a2a`.

One adaptation was needed. `runner_id` is also a common function-parameter name, so matching every call keyword — as the checkpoint guard does — would false-positive on `TaskLease(runner_id=...)`, `acquire_task_lease_no_commit(runner_id=...)` and others. Restricting matches to `.values(...)` and `.update(...)` carriers fixes that: run before the fix it flags exactly the two unpaired sites and nothing else, and it is clean after. Verified both directions by temporarily removing each fix line — the guard reds the corresponding module.

A shared constant does not work here — the sites use three different carrier forms, and it could not express the non-None stamp on acquire.

tests/alembic/test_20260808_add_task_lease_attempt_id.py
AlexLiu190625: Covered. Adapted the sibling's `test_target_schema_resolves_the_visible_tasks_relation`, which needs no PostgreSQL environment — it asserts on the catalog SQL and on the SQLite fallback returning `None`.

Also added a second test that actually executes the PostgreSQL branch against a stub bind, which the sibling does not do: it pins that the catalog statement is issued with `{"table_name": "tasks"}`, that the resolved schema is returned, and that a `NULL` resolution falls back to `version_table_schema`.

src/xagent/web/services/task_lease_service.py
AlexLiu190625: Keeping the dict read, though our own earlier justification for it was overstated and is worth correcting on the record. Your type point is right: `values` is `dict[str, Any]`, so the read is unchecked while the local is a known `str`.

The reason to keep it is narrower than "fail-loud coupling" suggests. Reading through the dict does fail immediately if the persisted entry is removed, but it is equally silent if someone changes the entry's value — the more likely edit. So the benefit covers one mutation shape, not the general case. Against that, the type widening is `str` into `str | None`, which cannot hide a real error, and the round trip is pinned against the database by the stamping test. Net, not worth the churn.

PR #1207, opened 2026-08-08
ref(trace): extract trace-row staging into a caller-owned-transaction primitive

AlexLiu190625:
## What

`_save_trace_event` in `trace_handlers.py` used to both build the trace row and own the transaction wrapped around it, so nothing could stage a checkpoint row inside a transaction it did not also commit. Its four write paths now delegate row construction to a new staging primitive, `stage_trace_event_row` (`src/xagent/web/services/trace_event_staging.py`), which joins the caller's already-open session and performs no commit, rollback, notification, or prune of its own. The shell keeps every boundary action: the pointer UPDATE and its rowcount classification, the tool-usage counters, the commit, the prune trigger, and both rollback handlers.

The four paths keep their exact shapes: a non-checkpoint event neither encodes nor flushes; a sub-agent checkpoint encodes without a run-partition stamp; a root checkpoint with no live lease encodes without flushing; a root checkpoint under a live lease stamps the partition, flushes, and returns the exact-row anchor the shell writes into `tasks.last_checkpoint_trace_event_id`.

Alongside the extraction, the pairing guard in `tests/web/services/test_checkpoint_pointer_pairing.py` (`_mutated_pointer_columns`) is hardened against opaque `**` carriers. It previously only recognized `.values(**name)` / `.update(**name)` as transparent when it could read `name`'s keys directly off a dict literal it had already walked; a `.values(**builder())` call carried both pointer columns past it with no finding. The guard now resolves names bound to a dict literal within the same function and treats every other `**` carrier as opaque, and therefore unpaired. Two shapes knowingly escape this hardening and are pinned as negative controls rather than left to be rediscovered: a paired local dict literal later merged with `.update(builder())` before being unpacked, and a paired local dict literal later rebound to the result of a call before being unpacked.

## Why

#1071 asks to "Expose checkpoint staging that joins a caller-owned transaction and performs no implicit commit, rollback, notification, or prune." Delivers the staging clause of #1071 (jointly closed with the offline-SQL migration fix and the deletion-path rows change).

Behavior is preserved by construction, not by re-testing from scratch: the two test files that call `_save_trace_event` directly (`tests/web/test_agent_checkpoint_stream.py`, 16 call sites, and `tests/web/test_trace_message_storage.py`, 9 call sites) are unmodified by this PR, and the pairing-guard file's pre-existing cases stay green under the hardened rule. The primary-key guard's `RuntimeError` keeps its exact type and message text; no new exception type is introduced anywhere in this change.

## Disposition table

Four write paths through `_save_trace_event`, five boundary actions:

| Path | commit | IntegrityError rollback | bare-except rollback | task-gone rollback | prune |
|---|---|---|---|---|---|
| (a) non-checkpoint | shell, always reached | shell, reachable | shell, reachable | unreachable | not triggered |
| (b) sub-agent checkpoint | shell, always reached | shell, reachable | shell, reachable | unreachable | shell triggers |
| (c) root checkpoint, no lease | shell, always reached | shell, reachable | shell, reachable | unreachable | shell triggers |
| (d) root checkpoint + live lease | shell, always reached | shell, reachable | shell, reachable | shell | shell triggers |

The primitive occupies zero of these twenty cells. That is the row-by-row proof behind "no implicit commit, rollback, notification, or prune": every action either lands in the shell or is unreachable on that path. The fifth verb, notification, has zero call sites in `_save_trace_event` at all -- no notify/dispatch/publish/broadcast symbol appears anywhere in the function -- so it ships as a zero-site fact about the existing function, not as something moved out of it.

## Teardown-guard evaluation

Two counts taken against the tree this PR lands on:

```
grep -rn "Base\.metadata\.drop_all" --include="*.py" tests/ src/ | wc -l   ->  89
grep -rn "drop_all_tables"          --include="*.py" tests/ src/ | wc -l   ->  16
```

The 89 figure moved from a baseline of 88 for a reason unrelated to this PR: an earlier, already-merged change added one test file with a bare `drop_all` that names no checkpoint column at all. The 16 figure moved from a baseline of 15 because of this PR's own new contract-test file for the primitive: it adds exactly one more raw match of the string `drop_all_tables`, and that match is a sentence in the file's module docstring explaining why the file does *not* need the shared teardown helper (it builds its own file-backed, per-test SQLite engine under `tmp_path` rather than using the process-wide shared engine). It is not a call.

16 is a raw line count of that grep, not a count of call sites, and reading it as one produces a wrong "coverage" ratio. The breakdown:

| Category | Count |
|---|---|
| Real call sites | 7 |
| Import lines | 6 |
| Helper function definition | 1 |
| Test-function name containing the phrase | 1 |
| Docstring mention (this PR's new contract-test file) | 1 |
| **Total** | **16** |

Only 7 real call sites across 6 files use the shared teardown helper today; that set is unchanged by this PR. Comparing 16 against 89 as a coverage percentage would be comparing the wrong things -- the real ratio of interest is 7 real call sites against however many files could plausibly need one, which is the question the next paragraph answers directly rather than by ratio.

Ten files in the tree name the anchor column `last_checkpoint_trace_event_id` at all -- the full set that could plausibly persist a populated checkpoint anchor into the shared database:

```
tests/migrations/test_20260804_add_task_checkpoint_trace_event_anchor.py   naked=0  helper=0
tests/web/test_db_teardown.py                                             naked=1  helper=4
tests/web/test_agent_checkpoint_stream.py                                 naked=0  helper=0
tests/web/test_task_lease_service.py                                      naked=0  helper=2
tests/web/services/test_task_deletion_checkpoint_pointer.py               naked=0  helper=0
tests/web/services/test_task_orchestrator.py                              naked=0  helper=2
tests/web/services/test_checkpoint_pointer_pairing.py                     naked=0  helper=0
tests/web/services/test_task_lease_recovery.py                            naked=0  helper=2
tests/web/services/checkpoint_anchor_shared.py                            naked=0  helper=0
tests/shared/db_teardown.py                                               naked=3  helper=1
```

Both `naked > 0` rows are accounted for and neither is a violation: the three hits in `tests/shared/db_teardown.py` are the helper's own implementation (one docstring line, one non-SQLite branch, one SQLite branch that runs only after foreign keys are turned off), and the one hit in `tests/web/test_db_teardown.py` is a dedicated negative-control test that calls the bare drop deliberately, to prove it breaks on a populated anchor, and cleans up with the helper in its `finally` block.

This PR's own new contract-test file for the primitive, `tests/web/services/test_trace_event_staging.py`, does not appear in this list at all -- it has zero occurrences of the string `last_checkpoint_trace_event_id`, because every one of its assertions works in terms of `staged.anchor.trace_event_id`, a value returned by the primitive before any commit, never written into the `tasks` table by the test itself. It structurally cannot trigger this guard's failure mode.

**Conclusion: no CI guard ships this round.** The argument is a reverse check, not a forward classifier: any test that could persist a populated anchor into the shared engine must, at minimum, name the anchor column somewhere in its source, so the ten files above are a superset of every test that could possibly be at risk. Checking each of those ten and finding zero risky bare-drop usages is therefore sufficient to conclude the risky set is empty, without needing to decide, file by file, whether a given test actually persists an anchor -- a forward question that would require tracing data flow through fixtures and shared helpers, which static analysis cannot answer reliably and which was already flagged as unreliable ground for a guard. The reverse check costs two greps and one loop; nothing more elaborate is warranted for a set that has come up empty twice in a row.

## Verification

Full-suite result across the four files this PR touches or pins:

```
pytest tests/web/test_agent_checkpoint_stream.py tests/web/test_trace_message_storage.py \
       tests/web/services/test_checkpoint_pointer_pairing.py tests/web/services/test_trace_event_staging.py
145 passed
```

125 of those 145 are pre-existing tests with a zero-line diff against the base commit -- `git diff` on both `tests/web/test_agent_checkpoint_stream.py` and `tests/web/test_trace_message_storage.py` against the base returns nothing, and the pairing-guard file's 10 original cases are untouched (18 total in that file now, 8 new). The remaining 20 are new: 12 in the new primitive contract-test file, 8 new controls in the pairing-guard file.

Mutations applied by hand, run, confirmed red, and reverted:

- Adding a commit right after the primitive stages the row turns the commit-independence test red.
- Adding a rollback on the flush-failure branch turns the rollback-independence test red (that test specifically checks the session is left mid-transaction, which is exactly the state a rollback there would erase).
- Adding an internal call to the prune method inside the primitive turns the prune-independence test red (a call-count assertion).
- Removing the `build_id is None` gate, so a sub-agent checkpoint also flushes, turns the sub-agent-checkpoint path test red.
- Removing the run-partition stamp turns both the root-checkpoint-with-lease path test and the pre-existing door test that asserts the sub-agent path carries no run-id field red.
- Removing the primary-key guard's raise, leaving the flush in place, turns both the primitive-level guard test and the shell-level door test (`test_database_trace_handler_flush_without_primary_key_refuses_to_write_the_anchor`) red -- but the failure is not a clean "did not raise." Independently reproduced outside the guard tests themselves: with the raise deleted, the following line assigns `trace_event.id` (which is `None`, because the flush that would assign it was itself replaced with a no-op for this test) into `int(...)`, which raises `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`. That is not a `RuntimeError`, so `pytest.raises(RuntimeError, ...)` does not catch it -- the escaped `TypeError` propagates out and fails the test, rather than the test reporting that nothing was raised at all. Confirming the failure takes this exact form (an escaped wrong-type exception, not a silent pass-through) matters more than just seeing red, because a mutation that happened to raise some other `RuntimeError`-compatible error by accident would produce a misleadingly clean-looking red too.
- Changing one character in the raise's message text turns the shell-level anchor-guard door test red, because its match string is verbatim.
- Deleting the `data = staged.stored_data` rebind from the shell turns the test asserting that prune receives the run-id-stamped data red.
- Removing the `arg is None` branch from the guard's `**` resolution turns all three positive controls red at once: unpacking a builder call's result directly, a name only ever bound from a call, and a dict arriving as a cross-function parameter.
- Reverting the guard's resolution to the naive "any `**` is opaque" rule turns only the `task_lease_service` case red among the five-module parametrized guard test; the other four modules have no `.values(**local_dict)` sites at all and stay green either way.
- Removing the prune-entry guard turns its own reverse control red (calling prune directly against a session that still has pending writes).

Uncovered cells, named as a known gap rather than silently accepted: the primitive's own no-commit / no-rollback / no-prune tests are pinned only on the root-checkpoint-with-a-live-lease path, the one path that flushes. A rollback or prune call placed in the non-checkpoint fallthrough, or in either of the two paths that encode without flushing, would still pass every test in this PR -- those paths' own tests check the returned shape (row id, anchor, which fields got stamped) but do not additionally assert that rollback or prune stayed uncalled there.

## Scope notes

`_persist_agent_outbound_event` (`web/api/websocket.py`) constructs and adds its own trace row directly; it does not call `_save_trace_event` and never reaches the new primitive. That file has a zero-line diff in this PR -- the bypass is pre-existing and deliberately out of scope, not something this change introduces or touches. The primitive's own docstring names this bypass explicitly so a future reader does not mistake it for the only construction point in the codebase.

The primitive's "sends no notification" test works by scanning the primitive module's own source text for the literal words notify, notification, dispatch, publish, and broadcast. That couples the test directly to whatever wording the module's docstring happens to use -- a future docstring edit that needed one of those words to explain why there is *no* notification would have to change this test in the same commit, not treat its failure as an unrelated regression.

### DIFF
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -41,12 +41,12 @@ from ...web.services.task_lease_service import (
     TASK_RUN_ID_TRACE_FIELD,
     current_task_lease,
 )
+from ...web.services.trace_event_staging import stage_trace_event_row
 from ...web.services.trace_message_storage import (
     SQL_IN_CLAUSE_CHUNK_SIZE,
     CheckpointMessageDecodeError,
     chunks,
     decode_trace_event_data,
-    encode_checkpoint_data_for_storage,
 )
 
 logger = logging.getLogger(__name__)
@@ -687,32 +687,22 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 checkpoint_lease = (
                     current_task_lease() if self.build_id is None else None
                 )
-                if checkpoint_lease is not None:
-                    data = {
-                        **data,
-                        TASK_RUN_ID_TRACE_FIELD: checkpoint_lease.run_id,
-                    }
-                data = encode_checkpoint_data_for_storage(
-                    db,
-                    task_id=self.task_id,
-                    data=data,
-                )
             else:
                 checkpoint_lease = None
 
-            # Create trace event record
-            trace_event = DatabaseTraceEvent(
+            staged = stage_trace_event_row(
+                db,
                 task_id=self.task_id,
-                build_id=self.build_id,  # ← 添加 build_id
+                build_id=self.build_id,
                 event_id=str(event.id),
                 event_type=event_type_str,
                 timestamp=timestamp,
                 step_id=event.step_id,
                 parent_event_id=str(event.parent_id) if event.parent_id else None,
                 data=data,
+                checkpoint_lease=checkpoint_lease,
             )
-
-            db.add(trace_event)
+            data = staged.stored_data
 
             if (
                 event_type_str == "system_update_general"
@@ -721,18 +711,13 @@ class DatabaseTraceHandler(BaseTraceHandler):
                 and self.build_id is None
                 and checkpoint_lease is not None
             ):
-                # Flush the pending insert so trace_event.id (the row's
-                # primary key) is assigned before the pointer UPDATE below
-                # reads it. autoflush is off for this session, so without
-                # this the exact-row anchor column would be built from an
-                # unassigned id and silently written NULL.
-                db.flush()
-                if trace_event.id is None:
-                    raise RuntimeError(
-                        f"Task {self.task_id} checkpoint {event.id} row has no "
-                        "primary key after flush; refusing to write a NULL "
-                        "checkpoint anchor"
-                    )
+                # The anchor's primary key was already flushed and
+                # NULL-checked inside stage_trace_event_row before it
+                # returned staged.anchor; consume it directly here. This
+                # guard is the same is_checkpoint / build_id / lease test
+                # stage_trace_event_row itself gates its flush on, so the
+                # anchor here is never None.
+                assert staged.anchor is not None
                 pointer_update = db.execute(
                     update(Task)
                     .where(
@@ -743,7 +728,7 @@ class DatabaseTraceHandler(BaseTraceHandler):
                     )
                     .values(
                         last_checkpoint_event_id=str(event.id),
-                        last_checkpoint_trace_event_id=trace_event.id,
+                        last_checkpoint_trace_event_id=staged.anchor.trace_event_id,
                     )
                     .execution_options(synchronize_session=False)
                 )
@@ -869,6 +854,10 @@ class DatabaseTraceHandler(BaseTraceHandler):
         task's exact-row pointer names a row that itself ranks outside the
         window, protecting that row keeps ``limit + 1``.
         """
+        if db.new or db.dirty or db.deleted:
+            raise RuntimeError(
+                "checkpoint prune must not run with pending writes on the session"
+            )
         limit = get_checkpoint_history_limit()
         if limit <= 0:
             return
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -704,20 +704,15 @@ class DatabaseTraceHandler(BaseTraceHandler):
             )
             data = staged.stored_data
 
-            if (
-                event_type_str == "system_update_general"
-                and isinstance(data, dict)
-                and data.get("checkpoint_type") == CHECKPOINT_TYPE
-                and self.build_id is None
-                and checkpoint_lease is not None
-            ):
-                # The anchor's primary key was already flushed and
-                # NULL-checked inside stage_trace_event_row before it
-                # returned staged.anchor; consume it directly here. This
-                # guard is the same is_checkpoint / build_id / lease test
-                # stage_trace_event_row itself gates its flush on, so the
-                # anchor here is never None.
-                assert staged.anchor is not None
+            if staged.anchor is not None and checkpoint_lease is not None:
+                # staged.anchor is set on exactly the path that needs this
+                # pointer UPDATE, so it is read here rather than re-derived.
+                # Re-deriving would test the post-encode payload rebound at
+                # the line above, not the pre-encode payload the flush
+                # decision was actually made from; the two agree today only
+                # because no encoder touches checkpoint_type. The lease is
+                # re-tested only to narrow it for the fence below -- it is
+                # never None when the anchor is set.
                 pointer_update = db.execute(
                     update(Task)
                     .where(
@@ -849,6 +844,14 @@ class DatabaseTraceHandler(BaseTraceHandler):
         dedup, but a blob referenced only by pruned rows (e.g. a message
         later dropped by context compaction) is orphaned until whole-task
         deletion cleans it up.
+
+        The entry guard below is a caller-contract check, not one of those
+        prune failures: it fires only when a caller hands this method a
+        session with uncommitted work, which the sole call site -- right
+        after the checkpoint commit -- never does. It raises outside the
+        degrade-and-log block deliberately, because that block's rollback
+        would discard the caller's pending writes and turn a misuse bug into
+        silent data loss.
 
         Retention is exactly ``limit`` rows, with one exception: when the
         task's exact-row pointer names a row that itself ranks outside the
--- a/src/xagent/web/services/trace_event_staging.py
+++ b/src/xagent/web/services/trace_event_staging.py
@@ -70,10 +70,16 @@ class StagedTraceRow:
     module docstring); where it is set, it names a real row only after the
     caller's commit succeeds. ``anchor`` is ``None`` on every path except
     the root-checkpoint-with-a-live-lease path.
+
+    ``stored_data`` is the payload as written to the row: the caller's own
+    ``data``, unchanged except on the checkpoint paths, which merge the
+    run-id field and encode blobs. It carries whatever type the caller
+    passed, which is why it is not annotated more tightly than the
+    ``data`` parameter itself.
     """
 
     row_id: int | None
-    stored_data: dict[str, Any]
+    stored_data: Any
     anchor: StagedCheckpointAnchor | None
 
 
--- a/src/xagent/web/services/trace_event_staging.py
+++ b/src/xagent/web/services/trace_event_staging.py
@@ -1,0 +1,154 @@
+"""Stage one trace row (and, on the live-lease checkpoint path, its
+exact-row anchor) into a session the caller already owns.
+
+``stage_trace_event_row`` is the construction step factored out of
+``DatabaseTraceHandler._save_trace_event`` (``web/api/trace_handlers.py``):
+building the row, encoding checkpoint blobs, stamping the run partition, and
+flushing to get a primary key for the anchor column. Everything that used to
+follow that construction in the same function -- the pointer UPDATE and its
+rowcount classification, tool-usage counters, the commit, the prune trigger,
+and both rollback handlers -- stays in the caller.
+
+Caller obligations, because none of them happen here:
+
+* This function joins the ``Session`` passed in; it never commits, never
+  rolls back, never signals anyone else about the row, and never prunes.
+  The caller owns the transaction and must commit it (or roll it back)
+  itself.
+* If this function raises ``IntegrityError`` (a stale FK on ``task_id``, for
+  instance), the session is left mid-transaction. The caller must roll back
+  before issuing another statement on it.
+
+``StagedTraceRow.row_id`` and ``.anchor`` are only meaningful once the
+caller commits: ``row_id`` is the primary key SQLAlchemy assigned during
+this call's flush, but it names a real, durable row only after commit
+succeeds. ``anchor`` is non-``None`` only on the root-checkpoint-with-a-live-
+lease path (flush happens there because the caller needs the anchor's
+primary key to write ``tasks.last_checkpoint_trace_event_id`` before it
+commits); every other path returns ``anchor=None`` and does not flush.
+
+Known bypass: ``_persist_agent_outbound_event`` (``web/api/websocket.py``)
+constructs its own ``TraceEvent`` row and adds it to the session directly,
+without going through ``_save_trace_event`` or this function. This function
+is the sole construction point on the ``_save_trace_event`` path, not the
+only place a trace row is ever built in the codebase.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Any
+
+from sqlalchemy.orm import Session
+
+from ...core.agent.checkpoint import CHECKPOINT_TYPE
+from ..models.task import TraceEvent as DatabaseTraceEvent
+from .task_lease_service import TASK_RUN_ID_TRACE_FIELD, TaskLease
+from .trace_message_storage import encode_checkpoint_data_for_storage
+
+
+@dataclass(frozen=True)
+class StagedCheckpointAnchor:
+    """The exact-row anchor for a checkpoint staged under a live task lease.
+
+    Non-``None`` only on the root-checkpoint-with-a-live-lease path.
+    ``trace_event_id`` is the primary key the caller writes into
+    ``tasks.last_checkpoint_trace_event_id``; it names a real row only once
+    the caller commits the transaction this row was staged into.
+    """
+
+    checkpoint_event_id: str
+    trace_event_id: int
+
+
+@dataclass(frozen=True)
+class StagedTraceRow:
+    """A trace row added to a caller-owned session, not yet committed.
+
+    ``row_id`` is ``None`` on every path that does not flush (see the
+    module docstring); where it is set, it names a real row only after the
+    caller's commit succeeds. ``anchor`` is ``None`` on every path except
+    the root-checkpoint-with-a-live-lease path.
+    """
+
+    row_id: int | None
+    stored_data: dict[str, Any]
+    anchor: StagedCheckpointAnchor | None
+
+
+def stage_trace_event_row(
+    db: Session,
+    *,
+    task_id: int,
+    build_id: str | None,
+    event_id: str,
+    event_type: str,
+    timestamp: datetime,
+    step_id: str | None,
+    parent_event_id: str | None,
+    data: Any,
+    checkpoint_lease: TaskLease | None,
+) -> StagedTraceRow:
+    """Add one trace row to ``db`` and, on the live-lease checkpoint path,
+    flush it to stage its exact-row anchor. See the module docstring for
+    what the caller still owns.
+
+    ``checkpoint_lease`` is the caller's own ``current_task_lease()``
+    result, already gated on ``build_id is None`` before this is called --
+    a sub-agent checkpoint (``build_id`` set) must be passed
+    ``checkpoint_lease=None`` regardless of whether a lease is live.
+    """
+    is_checkpoint = (
+        event_type == "system_update_general"
+        and isinstance(data, dict)
+        and data.get("checkpoint_type") == CHECKPOINT_TYPE
+    )
+    if is_checkpoint and checkpoint_lease is not None:
+        data = {
+            **data,
+            TASK_RUN_ID_TRACE_FIELD: checkpoint_lease.run_id,
+        }
+    if is_checkpoint:
+        data = encode_checkpoint_data_for_storage(
+            db,
+            task_id=task_id,
+            data=data,
+        )
+
+    trace_event = DatabaseTraceEvent(
+        task_id=task_id,
+        build_id=build_id,
+        event_id=event_id,
+        event_type=event_type,
+        timestamp=timestamp,
+        step_id=step_id,
+        parent_event_id=parent_event_id,
+        data=data,
+    )
+    db.add(trace_event)
+
+    if is_checkpoint and build_id is None and checkpoint_lease is not None:
+        # Flush the pending insert so trace_event.id (the row's primary
+        # key) is assigned before the caller builds the exact-row anchor
+        # column from it. autoflush is off for this session, so without
+        # this the anchor would be built from an unassigned id and
+        # silently written NULL.
+        db.flush()
+        if trace_event.id is None:
+            raise RuntimeError(
+                f"Task {task_id} checkpoint {event_id} row has no "
+                "primary key after flush; refusing to write a NULL "
+                "checkpoint anchor"
+            )
+        row_id = int(trace_event.id)
+        return StagedTraceRow(
+            row_id=row_id,
+            stored_data=data,
+            anchor=StagedCheckpointAnchor(
+                checkpoint_event_id=event_id,
+                trace_event_id=row_id,
+            ),
+        )
+
+    return StagedTraceRow(row_id=None, stored_data=data, anchor=None)
--- a/tests/web/services/test_checkpoint_pointer_pairing.py
+++ b/tests/web/services/test_checkpoint_pointer_pairing.py
@@ -19,6 +19,34 @@ therefore accepts two sibling statements that each write one column against
 write both columns through one carrier (or two sibling assignments into the
 same dict), so the blind spot has no live instance; it is recorded here so a
 future reviewer does not mistake the guard for a per-statement check.
+
+``**`` unpacking resolution: ``.values(**name)`` / ``.update(**name)`` name
+a dict whose keys this walk cannot see without resolving what ``name`` is
+bound to. A name bound to a dict *literal* somewhere in the enclosing
+function is resolved -- its keys were already picked up by the ``ast.Dict``
+branch below when the literal itself was walked, so the carrier contributes
+nothing further and is treated as already-inspected. Every other carrier
+(the call result of a builder, a name that only ever arrives as a function
+parameter, ...) cannot be resolved statically, so it is counted as an
+opaque, and therefore unpaired, write.
+
+This resolution is by name, not dataflow: once a name has been bound to a
+dict literal anywhere in its function, every ``**`` of that name in that
+function is accepted, even if the name is later reassigned or mutated.
+Two shapes escape as a known consequence, both covered by a dedicated
+negative-control test below so the limit stays visible instead of being
+rediscovered:
+
+* a local dict literal later merged with ``.update(builder())`` before
+  being unpacked -- the merge's keys are invisible to this walk;
+* a local dict literal later rebound to the result of a call before being
+  unpacked -- the same name now points at unrelated content, but the name
+  was "seen" as a literal once, so it stays resolved.
+
+Widening the walk to be dataflow-sensitive would close both gaps, but that
+is a materially different (and materially more expensive) analysis; the
+name-based rule is deliberately the cheaper, coarser one, with its blind
+spots recorded rather than silently accepted.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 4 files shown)


src/xagent/web/services/trace_event_staging.py
```diff
@@ -0,0 +1,154 @@
...
+    """A trace row added to a caller-owned session, not yet committed.
+
+    ``row_id`` is ``None`` on every path that does not flush (see the
+    module docstring); where it is set, it names a real row only after the
+    caller's commit succeeds. ``anchor`` is ``None`` on every path except
+    the root-checkpoint-with-a-live-lease path.
+    """
+
+    row_id: int | None
+    stored_data: dict[str, Any]
```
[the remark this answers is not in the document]
AlexLiu190625: Adopted the suggested `stored_data: Any`, though the reasoning differs from what's described.

Two parts of the finding don't hold. A non-dict payload isn't reachable on this path: `Tracer.trace_event` and `TraceEvent.__init__` both declare `data: Optional[Dict[str, Any]]`, that tracer is the only place a core trace event is constructed, mypy runs over the whole package in pre-commit so an in-package emitter passing a list would fail the build, and both transforms applied before this call (`_serialize_data_for_json` and `redact_runtime_sensitive_payload`) return a dict for a dict input. And no `AttributeError` is possible in the caller: every dereference of the rebound value is `isinstance(data, dict)`-guarded, including the `data.get("tool_name")` line, which reads `tool_name = data.get("tool_name") if isinstance(data, dict) else None`. Those guards are unchanged from before this refactor. The static-analysis point also doesn't reproduce: mypy reports nothing on these two files today, and nothing under any alternative annotation, because the parameter is `Any` and `Any` is compatible with `dict[str, Any]` in both directions.

The real defect is narrower and worth fixing anyway: the field claimed a guarantee its own parameter (`data: Any`) doesn't make, in a module whose surrounding code is uniformly defensive about exactly this. Tightening the parameter instead would make the internal `isinstance(data, dict)` check dead and invite removing it, which would turn today's silent tolerance into a crash — a behavior change this refactor is specifically meant to avoid.

src/xagent/web/api/trace_handlers.py
```diff
@@ -869,6 +854,10 @@ def _prune_checkpoint_history(self, db: Session, data: Dict[str, Any]) -> None:
         task's exact-row pointer names a row that itself ranks outside the
         window, protecting that row keeps ``limit + 1``.
         """
+        if db.new or db.dirty or db.deleted:
```
rogercloud, line 860: **Minor.** This guard raises from *outside* the `try:` block below (~867), whose `except` clauses (936, 964) deliberately degrade-and-log rather than raise. The docstring just above (841-851) promises prune "runs in its own transaction after the checkpoint commit so a prune failure can never take the checkpoint write down with it" — but a trip here propagates out through `_save_trace_event`'s outer handler and, given `require_persisted=True` in `checkpoint.py:135`, surfaces as a reported checkpoint failure even though the checkpoint already committed durably.

Not a live bug: `_prune_checkpoint_history` has exactly one call site, immediately after `db.commit()` (line 803), and `commit()` clears `new`/`dirty`/`deleted` — pinned by your own `test_prune_checkpoint_history_guard_allows_a_clean_session`.

Suggest either moving the check inside the `try:` so it degrades like every other prune failure, or noting in the docstring that this is an internal invariant assertion rather than a runtime failure mode, so the stated guarantee and the code agree.

src/xagent/web/api/trace_handlers.py
```diff
@@ -721,18 +711,13 @@ def _save_trace_event(self, db: Session, event: CoreTraceEvent) -> None:
...
-                        "primary key after flush; refusing to write a NULL "
-                        "checkpoint anchor"
-                    )
+                # The anchor's primary key was already flushed and
+                # NULL-checked inside stage_trace_event_row before it
+                # returned staged.anchor; consume it directly here. This
+                # guard is the same is_checkpoint / build_id / lease test
+                # stage_trace_event_row itself gates its flush on, so the
+                # anchor here is never None.
+                assert staged.anchor is not None
```
rogercloud: **Minor.** This `assert` is control-flow-relevant: under `python -O` it is stripped, and the dereference at line 731 (`staged.anchor.trace_event_id`) would then raise a confusing `AttributeError` on `None` instead. No `-O`/`PYTHONOPTIMIZE` in this repo's Dockerfiles, CI, or entrypoints, so it is theoretical today.

Better fix than hardening the assert: `staged.anchor` is non-`None` under exactly the four-part condition re-tested at 707-713, so replacing that condition with `if staged.anchor is not None:` makes the assert unnecessary, removes the duplicated predicate, and removes the pre-encode/post-encode divergence risk in one line.

tests/web/services/test_trace_event_staging.py
```diff
@@ -0,0 +1,575 @@
...
+# --------------------------------------------------------------------------
+
+
+def test_trace_event_staging_module_sends_no_notifications() -> None:
+    """No implicit notification is delivered as a zero-site fact
+    about this module's own source (there is nothing to move out), the
+    same way trace_handlers.py:_save_trace_event has zero notify/dispatch/
+    publish/broadcast sites today."""
+    source = Path(trace_event_staging.__file__).read_text()
+    for symbol in ("notify", "notification", "dispatch", "publish", "broadcast"):
```
rogercloud: **Minor.** This asserts the literal substrings `notify`/`notification`/`dispatch`/`publish`/`broadcast` never appear in the module's source text — a grep, not a behavioral check. Two consequences: a differently-named or aliased notification call passes cleanly, and a future docstring that uses one of these words to explain the *absence* of notification fails the test for purely cosmetic reasons.

Suggest either dropping it, or making it behavioral: monkeypatch the real outbound-notification entry point and assert zero calls across all four staging paths.

tests/web/services/test_trace_event_staging.py
```diff
@@ -0,0 +1,575 @@
...
+    return calls
+
+
+# --------------------------------------------------------------------------
+# Commit, rollback, and prune independence -- verbs the shell-level tests
+# cannot pin directly on the primitive
+# --------------------------------------------------------------------------
+
+
+def test_stage_trace_event_row_does_not_commit(tmp_path: Path) -> None:
```
rogercloud, line 135: **Minor — coverage gap.** These three tests (no-commit, no-rollback-after-IntegrityError, no-prune) are all pinned exclusively to the root-checkpoint-with-live-lease path. Paths (a) non-checkpoint, (b) sub-agent checkpoint, and (c) root-checkpoint-without-lease have no commit/rollback/prune-independence assertions at all — even though (b) and (c) both call `encode_checkpoint_data_for_storage`, the one collaborator that could plausibly acquire commit behavior in a future change.

Since "the primitive never owns the transaction" is the whole contract this PR exists to establish, it is worth holding on every path. Cheap to close by parametrizing these three over the four path shapes you already build fixtures for at 255/289/351/385.

tests/web/services/test_checkpoint_pointer_pairing.py
```diff
@@ -184,3 +295,129 @@ def test_pointer_pairing_flags_each_carrier_form(carrier: str, fixture: str) ->
...
+    fixture = """
+def write(db, task_id, vals):
+    db.execute(update(Task).where(Task.id == task_id).values(**vals))
+"""
+    unpaired = unpaired_pointer_writes(fixture)
+    assert unpaired
+    assert {columns for _, columns in unpaired} == {(_OPAQUE_STAR_CARRIER,)}
+
+
+def test_pointer_pairing_known_gap_a_literal_merged_via_update_before_star() -> None:
```
rogercloud, line 378: **Minor — flagging a documented limitation for visibility.** This gap is mechanically real: `values.update(checkpoint_pointer_values(anchor))` passes its argument *positionally*, so the AST walk's `keyword.arg is None` check never inspects the call node at all — the merge is invisible to the guard regardless of what the builder returns. No production code does this today (`checkpoint_pointer_values` exists only in this file's fixtures), so it is currently inert, and it's documented honestly.

Raising it because `.update(builder())` is a plausible shape for the #1071 follow-up this PR is prepping for. If that work introduces exactly this pattern in production pointer-write code, the guard misses it silently. Worth either closing the gap now or leaving a note on #1071.

tests/web/services/test_checkpoint_pointer_pairing.py
```diff
@@ -57,11 +142,33 @@ def _mutated_pointer_columns(statement: ast.stmt) -> set[str]:
...
+                elif keyword.arg is None:
+                    # .values(**name) / .update(**name). Resolve only for
+                    # these two method names -- an unrelated **-call (e.g.
+                    # log.info(**kwargs)) is not a pointer-column carrier
+                    # at all and must not be flagged.
+                    callee = node.func
+                    method_name = (
+                        callee.attr if isinstance(callee, ast.Attribute) else None
+                    )
+                    if method_name in {"values", "update"}:
```
rogercloud, line 170: **Minor.** `method_name in {"values", "update"}` matches *any* `.update(**x)` / `.values(**x)` call, not just SQLAlchemy `update(Task)...values()` chains. Because `ast.walk` unions findings across the whole function body, this reproduces a false positive: a fully-paired, legitimate pointer write co-located with an unrelated `other_dict.update(**overrides)` in the same function gets flagged as an unpaired write.

No such collision in any of the five subject modules today (their `.update(` calls all use dict literals, not `**`), so this is latent rather than active. Tightening the match to the SQLAlchemy chain — e.g. requiring the receiver to trace back to an `update(...)` call — would keep the guard from misfiring as these modules grow.

rogercloud (approved):
## Summary

`_save_trace_event` in `src/xagent/web/api/trace_handlers.py` previously conflated two responsibilities: building the trace row and owning the transaction around it (commit, prune, checkpoint-pointer UPDATE, rollback). That coupling blocks a future caller (#1071) from staging a checkpoint row inside its own transaction without inheriting an implicit commit. This PR extracts pure row construction into a new caller-owned-transaction primitive, `stage_trace_event_row` (`src/xagent/web/services/trace_event_staging.py`), which joins the caller's session and performs no commit, rollback, prune, or notification of its own; `_save_trace_event` remains the shell and keeps every transaction-boundary action. Behavior across all four write paths (non-checkpoint; sub-agent checkpoint; root checkpoint without lease; root checkpoint under a live lease) is preserved.

## Update since last review

No new commits have landed since the previous review round — the current HEAD (`ce2f48ee`) *is* the commit that addressed the one outstanding finding. The single prior action item (gemini-code-assist: `StagedTraceRow.stored_data` annotated `dict[str, Any]` while it can carry `Any`) was fixed in that commit. This re-review covers the same tree, re-verified end to end, plus the findings below that were not raised in earlier passes.

## Approach verdict: Acceptable with reservations

The extraction itself is correct and behavior-preserving — verified across all four write paths directly. The pre-encode vs post-encode predicate split is safe today because the encoder deep-copies and never mutates the top-level `checkpoint_type` key. Three design-level reservations, none of which own a single line:

- **No second consumer exists today.** `stage_trace_event_row` is referenced only by `trace_handlers.py` (its sole caller) and its own test file; grep confirms zero other call sites in the repo. The motivation is entirely prospective (prep for #1071, a separate future PR). That means ~160 lines of new production code and ~575 lines of new tests for what is, today, a pure move whose payoff lands elsewhere. Defensible as deliberate prep work, but it is a design choice with real carrying cost, not a free one.

- **The checkpoint-with-live-lease gate predicate is now evaluated three separate times.** `trace_handlers.py:682-686` (to compute `checkpoint_lease`), `trace_handlers.py:707-713` (to decide whether to enter the pointer-update block — evaluated against **post-encode** data), and `trace_event_staging.py:108-112` / `:137` (to decide whether to flush — evaluated against **pre-encode** data). `StagedTraceRow.anchor` is non-`None` exactly when this same four-part condition holds, so `if staged.anchor is not None:` at `trace_handlers.py:707` is a drop-in replacement: it removes the duplication, removes the pre/post-encode divergence risk, and makes the `assert` at line 720 unnecessary. Today the two evaluations agree only because the encoder happens never to rewrite `checkpoint_type`; that is a latent, undocumented coupling a future encoder change could silently break.

- **Three shell-side tests live in the primitive's test file.** `tests/web/services/test_trace_event_staging.py:486-575` (`test_shell_passes_the_partition_stamped_data_into_prune` plus the two prune-guard controls) call `DatabaseTraceHandler._save_trace_event` / `_prune_checkpoint_history` directly and never touch `stage_trace_event_row`; the file's own docstring self-labels them "Shell-side supplements." Minor organizational point — they read more naturally alongside the other `trace_handlers.py`-focused tests.

## Prior findings

| 1 | `StagedTraceRow.stored_data` typed `dict[str, Any]` but can hold `Any` (gemini-code-assist) | **FIXED** | `trace_event_staging.py:82` is now `stored_data: Any`; `mypy src/xagent` clean; every dereference of the field is dict-guarded at the use site. Adopted in `ce2f48ee`. |

## Findings

All six are minor. No blocking issues.

1. `src/xagent/web/api/trace_handlers.py:857` — the new pending-writes guard raises from outside the `try:` block, contradicting the docstring's "a prune failure can never take the checkpoint write down with it" guarantee. Currently unreachable in production (the one call site follows `db.commit()`), so not a live bug.
2. `src/xagent/web/api/trace_handlers.py:720` — `assert staged.anchor is not None` is control-flow-relevant and is stripped under `python -O`. No `-O`/`PYTHONOPTIMIZE` anywhere in this repo's Dockerfiles/CI/entrypoints, so theoretical today; subsumed by design reservation #2.
3. `tests/web/services/test_trace_event_staging.py:477` — the no-notification test is a source-text substring grep, not a behavioral assertion.
4. `tests/web/services/test_trace_event_staging.py:134` — the three commit/rollback/prune-independence tests are pinned exclusively to the root-checkpoint-with-live-lease path; the other three paths have no such assertions.
5. `tests/web/services/test_checkpoint_pointer_pairing.py:368` — documented known gap: `values.update(checkpoint_pointer_values(anchor))` passes positionally, so the AST walk never inspects it. Inert today, but a plausible shape for the #1071 follow-up.
6. `tests/web/services/test_checkpoint_pointer_pairing.py:160` — `method_name in {"values", "update"}` matches any `.update(**x)`, not just SQLAlchemy chains; reproducible false positive. No collision in the five subject modules today.

Full detail and suggested fixes are in the inline comments.

- `src/xagent/web/services/trace_event_staging.py` L50-62: **redundant wrapper** — `StagedCheckpointAnchor` wraps `checkpoint_event_id` (a duplicate of the already-available `event_id`) and `trace_event_id` (numerically identical to `StagedTraceRow.row_id` whenever set — the PR's own test asserts `staged.anchor.trace_event_id == staged.row_id`). One caller, one read, at `trace_handlers.py:731`. Drop the dataclass and the `anchor` field; gate on `if staged.row_id is not None:` at `trace_handlers.py:707` (which also resolves design reservation #2) and read `staged.row_id` directly.

net: -25 lines possible

## Test results

Full run of the four touched/pinned test files from a clean worktree at `ce2f48ee`: **145 passed, 0 failed** (`tests/web/test_agent_checkpoint_stream.py`, `tests/web/test_trace_message_storage.py`, `tests/web/services/test_checkpoint_pointer_pairing.py`, `tests/web/services/test_trace_event_staging.py`). `mypy src/xagent` clean.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Taking the middle option, with one adjustment. Reading `staged.anchor` instead of re-deriving the condition is right, and the drift risk is concrete: the local `data` is rebound to the post-encode payload one line above, so the re-derivation tests a different object than the one the flush decision was made from. They agree today only because no encoder touches `checkpoint_type`, and if that ever changed the pointer UPDATE would be skipped on a checkpoint that was already flushed — a committed row no recovery path can find.

The adjustment: a bare `if staged.anchor is not None:` loses the narrowing on `checkpoint_lease`, which the fence below dereferences — mypy flags it — so that form would just trade one assert for another. The gate is now `if staged.anchor is not None and checkpoint_lease is not None:`, which narrows both, drops the assert, and removes the duplicated predicate.

Not taking the further step of dropping the anchor type. The observation behind it is correct — `checkpoint_event_id` has no production reader, since the pointer write takes `str(event.id)` from the shell's own event — but `row_id` and `anchor` mean different things: `row_id` means "this call flushed," `anchor` means "this is a checkpoint anchor." They coincide only because there is exactly one flush site today, and gating on `row_id` would silently promote that coincidence to an invariant. The two-field shape and its equality assertion are also part of the agreed scope for this change, so removing them is a scope decision rather than a review one.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Reachability confirmed — the only production call site follows the commit, so this never fires there. It does fire from the nine direct test call sites, which is where a misuse would actually surface.

Not moving it inside the try, though, and the reason is stronger than "it would degrade instead of failing": the bare handler there does `db.rollback()`, so the guard raising inside it would discard the caller's uncommitted writes — exactly what the guard exists to prevent. It would turn a misuse bug into silent data loss, and it would not register a degradation signal either.

Fixed the docstring instead: it now states that the entry guard is a caller-contract check rather than one of the prune failures the degrade-and-log block handles, and why it deliberately sits outside that block.

tests/web/services/test_trace_event_staging.py
AlexLiu190625: The point about it being a source grep is fair, but a behavioral version is not cleanly writable — there is no single outbound entry point to patch. The surface is at least the two connection-manager methods and the command-dispatcher notify, which are unrelated to each other, and the connection-manager ones are async while this function is sync, so it structurally cannot call them. A test patching a hand-picked set would assert "none of the ones I thought of fired," which decays as that set grows.

There was a real defect here, just a different one: the substring form matched prose. The shell module already contains a comment with the word "publishes," and this module's own docstring says "never signals anyone else about the row" — written that way to dodge the banned words. So the test constrained how the module could be documented and would eventually have gone red for a wording reason.

Changed it to an AST check over imports and call targets. Same cost, same intent, no prose coupling — verified both directions: a banned name binding turns it red, a comment containing "publish" no longer does.

tests/web/services/test_trace_event_staging.py
AlexLiu190625: Checked whether all three parametrize, and only one does cleanly.

The prune test parametrizes fine. The rollback test cannot: it needs an `IntegrityError` raised inside the primitive, which needs a flush, and only the live-lease path flushes — on the other three there is nothing to observe. The commit test would need its assertions rewritten: with `autoflush=False` and no flush, `db.query(...).count()` returns 0 on the non-flushing paths, so the `== 1` assertion fails — measured rather than guessed.

That assertion is the point of the test: it is what makes "a second session cannot see it" a transaction-isolation check rather than an ORM identity-map artifact, which is why the live-lease path was chosen. Parametrizing would dilute the strongest form to add three weaker instances, while the four-path shape is already pinned by the four dedicated path tests. Leaving as-is; happy to take the prune test alone if you would still like it.

tests/web/services/test_checkpoint_pointer_pairing.py
AlexLiu190625: Worth correcting one assumption: closing this does not need dataflow analysis. A name-poisoning rule does it — a name loses its resolved status if it is ever rebound from a non-literal, or is the receiver of `.update()` with a non-literal argument, in the same scope. About fifteen lines. Prototyped it: all five subject modules stay green, both this shape and the rebind shape flip to red, and the paired-literal control stays green.

Still not doing it here. These two shapes were agreed as recorded blind spots with a test each, and that agreement is explicit that turning either into a catch is a deliberate rule change rather than a drive-by fix. Also relevant: no planned change introduces this shape — the builder that would have produced it was dropped, and the pointer write stays as two literal keywords, so the gap has no instance to hit.

The feasibility finding, with the prototype numbers, is recorded on #1071 so the next person does not re-derive it.

tests/web/services/test_checkpoint_pointer_pairing.py
AlexLiu190625: Reproduced, and it is broader than described: co-location is not required. Any unresolvable `**` into a `.values()`/`.update()` call is flagged on its own, because the block's column set is non-empty and unequal to the pair. One correction — the union is per statement block, not per function.

Not taking the receiver fix, though. Requiring the receiver to be a call would separate bulk updates from dict merges, but it would also stop flagging a real pointer write through a query object bound to a local name — `stmt = db.query(Task).filter(...)` then `stmt.update(**vals)`. That trades a loud false positive for a silent miss, which is backwards for a guard whose quiet failure mode is that recovery stops working.

Added a docstring paragraph recording the false positive and the reason it is accepted, so whoever hits it has the explanation directly above the code. Nothing in the five modules uses that shape today.

AlexLiu190625:
On the two remaining design reservations from the review body:

**The prospective-only primitive.** One correction to how the PR frames it, which your reservation surfaced: the justification is not that a future change will call `stage_trace_event_row` — none is planned to; the follow-up work resolves its anchor from the committed task row instead. The justification is that the staging clause requires the four verbs (no commit, no rollback, no notification, no prune) to be assertable, and the existing shell tests structurally cannot pin them because they only observe the combined shell behavior. A standalone callable is what gives those verbs a test surface. The carrying cost you name is real and accepted on that basis.

**The three shell-side tests in the primitive's file.** Their natural home is the big shell test file, but that file is one of the two zero-diff behavior gates this change is built on — the gate is what proves the extraction preserved behavior — so additions there were off the table by design. The current placement, with the section header explaining it, is the deliberate resolution of that constraint rather than an oversight.

PR #1208, opened 2026-08-08
enh: harden computer use interactions

qinxuye:
## Summary

- flatten the model-facing Computer Use action schema while continuing to accept the previous one-element `actions` envelope for replay compatibility
- distinguish private `observe` frames from user-visible `screenshot` artifacts, and publish requested screenshots through registered `FileRef` output
- carry semantic surface provenance on observed elements so current and future adapters can distinguish document, application-chrome, native-app, and overlay elements without requiring callers to echo that provenance in action targets
- honor per-observation capability metadata for state-changing actions
- move screenshot publication I/O off the async event loop
- guide ReAct completion decisions to distinguish private model observations from user-visible artifact delivery

## Why

The existing loop used the same screenshot action for model perception and user delivery, and exposed a nested single-action schema that some tool-calling models serialized incorrectly. That made successful-looking runs capable of missing an explicit screenshot deliverable and made the action boundary unnecessarily fragile.

This change tightens the provider-neutral action/observation contract without adding a new Computer Use backend or speculative interaction heuristics.

## Scope boundaries

This PR **does** change the shared Computer Use schema, browser environment, generic Computer tool adapter, and ReAct completion guidance.

This PR intentionally **does not**:

- add click-stall heuristics or other coordinate-level retry policy
- hide or reroute `understand_media`; media-tool routing remains unchanged
- add `cua-driver`, Local Computer, native window selection, readiness checks, or foreground/background delivery
- add or change composer, settings, task-message badges, or any other frontend UI
- add the Chrome extension, browser relay, pairing, or desktop companion
- change Anthropic/OpenRouter prompt caching or provider routing configuration
- integrate provider-native Computer Use APIs
- claim host-level macOS/Chrome end-to-end coverage

Those remain separate follow-up slices so this review can stay focused on the shared protocol and reliable artifact delivery.

## Compatibility

- existing one-element `actions` payloads remain accepted at the tool boundary
- existing semantic targets such as `{"element_id": "dom-3"}` remain valid; observed `surface` is provenance, not a new required action field
- bare and JSON-encoded semantic targets are normalized and covered through environment execution
- coordinate targets remain supported
- `understand_media` availability and routing are unchanged
- unknown top-level fields and conflicting duplicate values are rejected rather than silently ignored; compatibility normalization covers the documented action fields only

## Validation

- `ruff check` and `ruff format` on all changed files
- pre-commit: ruff, mypy, EOF/whitespace, isort, codespell
- `PYTHONPATH=/Users/xuyeqin/Workspace/xagent-computer-use-reliability/src python -X faulthandler -m pytest -q tests/core/computer tests/core/agent/test_react.py` (235 passed)


src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -237,26 +428,114 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+        suffix = source_path.suffix or ".png"
+        destination = output_dir / f"computer-screenshot-{frame_key}{suffix}"
+        if (
+            destination.exists()
+            and destination.read_bytes() != source_path.read_bytes()
+        ):
+            content_key = hashlib.sha256(source_path.read_bytes()).hexdigest()[:12]
+            destination = output_dir / f"computer-screenshot-{content_key}{suffix}"
+        if not destination.exists():
+            shutil.copy2(source_path, destination)
```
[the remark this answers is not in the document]
qinxuye: Fixed in cf8df9b3. `_publish_screenshot` now reads the source only when the frame-key destination already exists, and reuses those bytes for both comparison and the content hash.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -213,13 +363,54 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
                     )
                 )
+                if action.type not in {
+                    ComputerActionType.CLICK,
+                    ComputerActionType.DOUBLE_CLICK,
+                } or (
+                    current.active_url != observation.active_url
+                    or current.title != observation.title
+                ):
+                    self._recent_clicks.pop(session_id, None)
```
rogercloud: **Major** - Stall reset is backwards: any non-click action wipes history, so the canonical `click→wait→click→wait` stuck loop never accumulates a count. Meanwhile a legitimate repeated click on a non-navigating control (SPA "Load more", modal confirm, pagination) is permanently refused after 3 clicks since only a URL/title change clears history. Suggest resetting on observed page-state change (frame/element-table delta) and letting the counter decay instead of clear. Neither branch is covered by a test.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -268,6 +546,82 @@ async def teardown(self, task_id: str | None = None) -> None:
...
+    @staticmethod
+    def _click_target(
+        action: ComputerAction,
+        observation: ComputerObservation,
+    ) -> tuple[NormalizedPoint, tuple[Any, ...]] | None:
+        target = action.target
+        if target is None:
+            return None
+        if target.point is not None:
+            return target.point, ("point",)
```
rogercloud: **Major** - `target_key` for point targets is the constant `("point",)`, so all coordinate clicks collapse to one key and distinctness rests solely on the 0.05 normalized radius. Four clicks on four genuinely distinct nearby targets (on-screen keyboard, adjacent checkboxes, grid cells) get the 4th refused. Include quantized `(x, y)` in the key.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -213,13 +363,54 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+                        session_id=session_id,
+                        frame_id=current.frame_id,
+                        error=(
+                            f"{action.type.value} is not supported by the current "
+                            f"computer observation{detail}. Do not retry or work "
+                            "around this capability through another tool."
+                        ),
+                    )
+                if parsed.expected_frame_id == current.frame_id:
+                    repeated_click_error = self._record_click_attempt(
```
rogercloud: **Major** - The click is recorded into stall history here, before `environment.execute(...)` is awaited (line 399), with no rollback on exception. Three consecutive execute failures (frame mismatch, target not found, Playwright error) can trip the lockout for clicks that never happened. Record after success, or unwind in the except path.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -989,6 +996,32 @@ def _schema_tool_names(self, tool_schemas: list[dict[str, Any]]) -> list[str]:
...
+        """Avoid a second image-analysis hop when the selected model sees frames.
+
+        Auto routing resolves the concrete model per call. A text-only fallback
+        still needs ``understand_media``; a visual model already receives the
+        computer screenshot through its ContextReference and should reason from
+        that same frame instead of delegating coordinate selection to another
+        model.
+        """
+
+        names = self._schema_tool_names(tool_schemas)
```
rogercloud: **Major** - This strips `understand_media` for the whole session whenever browser tools are enabled (the default) and the model has vision - not just for computer-use turns. `understand_media` also handles video analysis and workspace/uploaded-file inspection, which screenshots don't replace, and `skills/builtin/static-visual-design/SKILL.md` still instructs the model to call it in this state. Routing above (in this method's caller) also decides against the unfiltered tool list before this filter runs. Suggest scoping the filter to turns with a live computer observation.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -237,26 +428,113 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+        source_path = Path(source).resolve()
+        output_dir = Path(workspace.output_dir).resolve()
+        output_dir.mkdir(parents=True, exist_ok=True)
+        frame_key = hashlib.sha256(observation.frame_id.encode("utf-8")).hexdigest()[
+            :12
+        ]
+        suffix = source_path.suffix or ".png"
+        destination = output_dir / f"computer-screenshot-{frame_key}{suffix}"
+        if destination.exists():
+            source_bytes = source_path.read_bytes()
```
rogercloud: **Major** - Blocking sync I/O on the event loop: two full `read_bytes()` calls here and below (up to 20MB per `store.py`'s `_DEFAULT_MAX_IMAGE_BYTES`), `shutil.copy2`, and the fully-synchronous `workspace.register_file` (via `build_workspace_file_ref`), all inline in an `async def`. The sibling path at `store.py:97-121` wraps equivalent I/O in `asyncio.to_thread`; suggest matching that.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -237,26 +428,113 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+        if source is None:
+            raise FileNotFoundError("computer observation screenshot is unavailable")
+        source_path = Path(source).resolve()
+        output_dir = Path(workspace.output_dir).resolve()
+        output_dir.mkdir(parents=True, exist_ok=True)
+        frame_key = hashlib.sha256(observation.frame_id.encode("utf-8")).hexdigest()[
+            :12
+        ]
+        suffix = source_path.suffix or ".png"
+        destination = output_dir / f"computer-screenshot-{frame_key}{suffix}"
```
rogercloud, line 497: **Simplification** - `frame_id` is a UUID4 unique per observation (`environment.py:87-92` raises on reuse), so this frame-key-then-content-hash fallback never actually fires. Hash content once and always name the file `computer-screenshot-{content_hash}{suffix}` - simpler, and gives real content dedup this two-step scheme doesn't provide.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -237,26 +428,113 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+            mime_type=str(
+                observation.screenshot.file_ref.get("mime_type") or "image/png"
+            ),
+        )
+        return sanitize_file_ref_for_context(file_ref)
+
     async def teardown(self, task_id: str | None = None) -> None:
         environments = list(self._environments.values())
         self._environments.clear()
+        self._recent_clicks.clear()
```
rogercloud: **Minor** - `_recent_clicks` is cleared only here and on same-call URL/title change, not on session/browser recreation (`browser.py:170-185`). A session recreated mid-task under the same `session_id` carries stale click history into a brand-new page and can trigger a premature stall refusal.

src/xagent/core/computer/environment.py
```diff
@@ -111,16 +115,36 @@ def validate_action_batch(self, batch: ComputerActionBatch) -> None:
...
                     f"{self._current_observation.frame_id!r}"
                 )
+            if target_id is not None and target is not None:
+                actual_surface = elements_by_id[target_id].surface
+                if (
+                    actual_surface is not None
+                    and actual_surface is not ComputerElementSurface.UNKNOWN
+                    and target.surface is None
+                ):
+                    raise ComputerTargetNotFoundError(
```
rogercloud: **Minor** - This is a real compatibility break: `{"element_id": "dom-3"}` without `surface` was valid before this PR and is now hard-rejected for every browser element. Deliberate and tested, but the PR's Compatibility section doesn't mention it - persisted traces/replays and external API consumers will break. Please call it out there.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -187,21 +335,23 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
             )
             self._environments[session_id] = environment
 
-        action = parsed.actions[0]
-        screenshot_only = action.type is ComputerActionType.SCREENSHOT
+        observation_only = action.type in {
```
rogercloud, line 340: **Minor** - `screenshot`/`observe` route straight to `environment.observe()` below, skipping both the `expected_frame_id` staleness check and the `supported_actions` capability check further down in this method. The tool description calls `supported_actions` "authoritative", which overstates it. Latent today (the shipped adapter always lists `screenshot`) but untested and fragile for future adapters.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -268,6 +546,82 @@ async def teardown(self, task_id: str | None = None) -> None:
...
+            return None
+        click_target = self._click_target(action, observation)
+        if click_target is None:
+            return None
+        point, target_key = click_target
+        recent = self._recent_clicks.setdefault(session_id, [])
+        nearby_count = sum(
+            attempt.target_key == target_key
+            and (attempt.x - point.x) ** 2 + (attempt.y - point.y) ** 2
+            <= _CLICK_STALL_RADIUS**2
```
rogercloud: **Minor (nit)** - `_CLICK_STALL_RADIUS = 0.05` applied to both axes of a normalized space over a non-square viewport (e.g. 1280x720) yields an ellipse in physical pixels (~64px x ~36px), not a circle. Scale by aspect ratio or document the anisotropy.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -268,6 +546,82 @@ async def teardown(self, task_id: str | None = None) -> None:
...
+        point = NormalizedPoint(
+            x=element.bounds.x + element.bounds.width / 2,
+            y=element.bounds.y + element.bounds.height / 2,
+        )
+        return point, (
+            "element",
+            element.surface.value if element.surface is not None else "",
+            (element.role or "").casefold(),
+            element.label or "",
+            element.text or "",
```
rogercloud: **Minor** - `round(bounds.x/y/width/height, 3)` in normalized space is sensitive to ~0.5-1px layout jitter between observations, so the key changes and the repeat-click guard is evaded even when role/label/text are identical. Prefer keying on `element_id`/role+label with bounds as tiebreaker.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -106,40 +195,98 @@ def __init__(
...
+        For a semantic target, send
+        `target={{"element_id": "<exact element_id>", "surface": "<exact surface>"}}`.
+        When a semantic element exposes `surface`, copying it is mandatory; the
+        action is refused before delivery when surface is omitted or mismatched.
+        `application_chrome` means browser/application controls, not website
+        content. If a document goal has no matching `document` element, use the
+        screenshot and a coordinate instead of a similarly named application
+        control. For a coordinate target,
+        send `target={{"point": {{"x": 0.5, "y": 0.5}}}}`. Do not send a bare
+        string or null as the target of a target-requiring action.
```
rogercloud: **Minor** - The description tells the model not to send a bare string as the target, but the normalizer earlier in this file accepts and converts bare strings. Pick one so the model-facing instructions aren't self-contradictory.

src/xagent/core/computer/schema.py
```diff
@@ -196,11 +221,14 @@ class ComputerElement(_ComputerModel):
...
     element_id: Identifier | None = None
     point: NormalizedPoint | None = None
+    surface: ComputerElementSurface | None = None
 
     @model_validator(mode="after")
     def _exactly_one_target(self) -> "ComputerTarget":
         if (self.element_id is None) == (self.point is None):
             raise ValueError("target requires exactly one of element_id or point")
+        if self.point is not None and self.surface is not None:
+            raise ValueError("surface is only valid for an element_id target")
```
rogercloud: **Minor (nit)** - `surface` on a `point` target raises `ValueError`, while the tool description says copying `surface` is "mandatory" without clearly caveating point targets. A model over-generalizing that rule gets a raw pydantic dump - less tolerant than the bare-string handling in the same validator. Consider ignoring `surface` on point targets.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -106,40 +195,98 @@ def __init__(
         task_id: str | None = None,
         workspace: TaskWorkspace | None = None,
         environment_factory: ComputerEnvironmentFactory = BrowserComputerEnvironment,
+        environment_instructions: str | None = None,
+        environment_label: str = "browser",
+        perception_mode: ComputerPerceptionMode | str = ComputerPerceptionMode.AUTO,
```
rogercloud, line 202: **Simplification** - `environment_instructions` / `environment_label` / `perception_mode` have zero production callers (the only call site, `browser_use.py:1141`, passes just `task_id`/`workspace`). Drop the params and the unreachable VISION/SEMANTIC description branches until a caller needs them.

src/xagent/core/computer/schema.py
```diff
@@ -127,6 +136,21 @@ class ComputerElementSource(str, Enum):
...
+class ComputerElementSurface(str, Enum):
+    """Structural surface that owns an observed UI element.
+
+    Surface is provenance derived by an adapter from the native element tree.
+    It describes where an element came from; it does not authorize or forbid
+    an action.
+    """
+
+    DOCUMENT = "document"
+    APPLICATION_CHROME = "application_chrome"
```
rogercloud, line 148: **Simplification** - `APPLICATION_CHROME`, `NATIVE_APP`, `OVERLAY` are never produced by any adapter (only `DOCUMENT`). Drop the unused members and the prose documenting behavior nothing generates.

rogercloud (commented):
## Summary

This PR flattens the model-facing Computer Use action schema while retaining the legacy nested `actions` envelope for compatibility, and separates private `observe` frames from user-visible `screenshot` deliverables that are published through `FileRef`. On top of that it layers several "hardening" mechanisms: semantic-surface provenance tracking and validation on click targets, click-stall detection to break repeated-click loops, suppression of the `understand_media` tool when a vision-capable model already holds the `computer` tool, and ReAct prompt instructions requiring the model to verify that a deliverable was actually user-visible before finishing.

The schema flattening, the private/public observation split, and the artifact-publishing path are the strong parts of this change - they are correct and well matched to the problem they solve. The concerns below are concentrated in the newly added hardening mechanisms, several of which do measurably less than the PR description claims.

Design-level concerns (no single owning line, or a systemic pattern, so raised here rather than inline):

- **Semantic-surface validation is a no-op beyond forcing the model to echo a string.** By the time the surface check runs in `environment.py`, `validate_action_batch` has already pinned the exact frame via `expected_frame_id` and the exact element via `target_id` existence in that frame's `elements_by_id`, and schema-level validation guarantees globally unique `element_id`s within a frame. There is no remaining "stale or wrong frame" scenario left for the surface check to catch - it can only reject a model that failed to copy back a string it was already handed. The PR describes this as rejecting "stale, missing, or mismatched semantic targets"; mechanically it does much less than that, at the cost of a new mandatory field and a new compatibility break.
- **The new leniency and the new strictness cancel each other out.** The new target normalization accepts a bare string or a JSON string and converts it into a target, but that path can never populate `surface`. The new surface validation then hard-rejects any target that references a surfaced element without `surface`. Since the only shipped adapter (`browser.py:319`) stamps every element with a concrete `DOCUMENT` surface, every target that goes through the lenient path is guaranteed to be rejected downstream. The leniency feature is unreachable in integration, and no test drives that path end-to-end (existing tests assert the parsed shape only, never push it through `validate_action_batch` + execute against a real element). Either drop the leniency, or have the lenient path resolve `surface` from the current frame's element table.
- **`understand_media` suppression is session-wide, not computer-use-turn-scoped.** `ComputerTool` is part of the default browser toolset and browser tools default to enabled, so merely not disabling browser tools removes the model's only route to `understand_media`, which also covers video analysis (frame sampling, start/end time) and workspace/uploaded-file inspection - none of which is redundant with computer screenshots. `src/xagent/skills/builtin/static-visual-design/SKILL.md` still instructs the model to call `understand_media` in exactly this state. The ReAct routing step also computes routing/tool-name decisions from the unfiltered tool list before the filter is applied.
- **ReAct "deliverable verification" is prompt text only, with no programmatic gate.** `user_visible_artifact` / `private_observation` / `inline_markdown` are produced as labels in `computer.py` and are never read back anywhere to gate, block, or validate a `final_answer` decision. The one code-level check (`react.py:2571-2572`) validates the model's own self-reported `missing_verification` free-text field against its own chosen action, never actual tool results or artifact records. Suggest either wording it as guidance, or wiring the delivery labels into the completion check.

- `computer.py:198-217`: yagni - `perception_mode` / `environment_label` / `environment_instructions` constructor params have zero production callers (the only call site, `browser_use.py:1141`, passes just `task_id` / `workspace`); drop the params and the unreachable VISION / SEMANTIC description branches.
- `schema.py:222-234`: yagni - `APPLICATION_CHROME`, `NATIVE_APP`, `OVERLAY` are never produced by any adapter (only `DOCUMENT` is); drop the three members and the description prose documenting behavior nothing generates.
- `computer.py:489-532`: shrink - `frame_id` is a UUID4 guaranteed unique per observation (`environment.py:87-92` raises on reuse), so the "try frame-key name, fall back to content-hash name on collision" two-step never exercises its fallback; hash content once and always use `computer-screenshot-{content_hash}{suffix}`, which also gives real content dedup the current scheme doesn't provide.

net: -40 lines possible

## Tests

All 231 relevant tests pass (`tests/core/computer`, `tests/core/agent/test_react.py`).

Coverage is solid on: flat schema + legacy envelope compat, screenshot publish happy path and delivery labeling, `screenshot_fresh=False` refusal, capability refusal preserving frame, one coordinate-click-stall scenario (3 ok, 4th refused), surface provenance match/missing/mismatch at the environment layer, `ComputerTarget` surface-only-with-`element_id` validation, and the `understand_media` filtering helper in isolation.

Gaps worth closing:
- The stall-reset branches (non-click action, URL/title change) are never executed by any test - and the first Major finding below is precisely a bug in that untested logic.
- Repeated same-element target stall accumulation is untested (only the coordinate path is covered).
- End-to-end surface enforcement through the tool boundary is untested; only the environment layer is exercised, which is why the leniency/strictness deadlock above went unnoticed.
- The lenient bare-string / JSON-string normalization is unit-tested for parsing only, never driven through execution.
- `_publish_screenshot`'s collision and reuse paths (identical bytes, differing bytes, missing `mime_type`, `resolve_file_id` returning `None`) are untested.
- No test exercises the real ReAct `_run` LLM call omitting `understand_media` - only the pure helper.
- Deliverable verification has only a prompt-substring assertion, consistent with there being no actual mechanism to test.

## Prior review status

Both Gemini bot findings are verified fixed in cf8df9b3:
- Redundant file reads in `_publish_screenshot`'s copy logic - now reads `source_path` exactly once, only when the frame-keyed destination exists, reusing the bytes for both comparison and hashing.
- `.value` vs. direct-enum comparison inconsistency in the environment surface check - both comparisons now use direct enum comparison consistently.

Recommend resolving those two review threads.

## Additional note

`src/xagent/core/computer/browser.py`'s `metadata["supported_actions"]` unconditionally lists every `ComputerActionType` (unchanged by this diff, but adjacent to it), so the new capability-check machinery in `computer.py` is currently a permanent no-op in production and only exercised by test fakes - worth knowing when trusting that guard.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in 055fcb7e by removing the click-stall heuristic from this PR. A correct state-delta and decay policy needs adapter-aware design; this protocol slice no longer guesses at it.

src/xagent/core/agent/pattern/react/react.py
qinxuye: Fixed in 055fcb7e. The session-wide `understand_media` suppression and its tests were removed; media-tool availability and routing are unchanged.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in 055fcb7e. Screenshot copying, comparison, hashing, and workspace registration now run through `await asyncio.to_thread(...)`.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Not adding a frame or capability gate to read-only capture. `observe` and `screenshot` intentionally sample current state without `expected_frame_id`; `supported_actions` governs state-changing delivery. The tool description now states that scope explicitly.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in 055fcb7e by removing the normalized-radius click-stall heuristic.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in eb090727. The description now distinguishes the canonical structured model output from tolerant bare/JSON element-ID normalization kept at the compatibility boundary.

src/xagent/core/computer/environment.py
qinxuye: Fixed in 055fcb7e. `surface` remains observation provenance but is no longer part of `ComputerTarget`, so existing `{\"element_id\": \"dom-3\"}` actions and persisted traces remain compatible.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in 055fcb7e by removing the click-stall state, so no history can leak across a recreated session.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Not changing this provider-neutral extension seam. `environment_label`, `environment_instructions`, and `perception_mode` let another registered environment reuse the same `ComputerTool` contract without subclassing or forking it. The concrete local backend remains deliberately outside this PR.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in 055fcb7e by removing the element-key click-stall heuristic.

src/xagent/core/computer/schema.py
qinxuye: Fixed in 055fcb7e. `surface` was removed from action targets entirely and remains observation-only provenance.

src/xagent/core/computer/schema.py
qinxuye: Not changing the protocol vocabulary. `DOCUMENT` is produced by the current browser adapter; the remaining surface values define the stable provider-neutral observation contract needed by subsequent adapters. No local adapter implementation is included here.

qinxuye:
Review follow-up on the current head:

- Removed the click-stall heuristic rather than patching its reset, keying, and delivery-order bugs in this protocol slice.
- Removed session-wide `understand_media` suppression; media routing remains unchanged.
- Kept element `surface` as observation provenance, but removed target-side surface echo and the resulting compatibility break.
- Moved screenshot publication I/O off the event loop.
- Clarified that ReAct artifact verification is completion guidance, not a new programmatic final-answer gate. A generic artifact-completion gate would need a broader contract across all artifact-producing tools and is outside this PR.
- Retained the configurable environment/perception seam and surface vocabulary as provider-neutral contracts for subsequent adapters; no local backend is added here.
- Retained frame-keyed screenshot artifact identity; unconditional content hashing is not required, and the I/O path is now threaded.

The PR description has been updated to match the narrower implementation and explicit non-goals. The focused suite passes: 225 tests across `tests/core/computer` and `tests/core/agent/test_react.py`.

qinxuye (commented):
Addressed the remaining review threads. Valid findings were fixed; suggestions outside this PR’s provider-neutral reliability scope were answered with the corresponding boundary rationale.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -31,65 +44,130 @@
...
+            # the canonical action protocol structured while accepting both
+            # unambiguous model-facing representations at this boundary.
+            try:
+                decoded_target = json.loads(raw_target)
+            except json.JSONDecodeError:
+                decoded_target = None
+            normalized["target"] = (
+                dict(decoded_target)
+                if isinstance(decoded_target, Mapping)
+                else {"element_id": raw_target}
```
rogercloud: A JSON-encoded *string* target falls through to the raw, undecoded value, baking literal quotes into `element_id`. Reproduced: `target='"button-1"'` yields `element_id='"button-1"'`, which can never match. Same for JSON arrays (`'["a"]'`). The decoded value should be reused whenever it is a `str`:
```suggestion
                else {"element_id": decoded_target if isinstance(decoded_target, str) else raw_target}
```

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -31,65 +44,130 @@
...
-        if nested_frame_id is None:
-            return normalized
-        top_level_frame_id = normalized.get("expected_frame_id")
-        if top_level_frame_id is None:
-            normalized["expected_frame_id"] = nested_frame_id
-        elif top_level_frame_id != nested_frame_id:
-            raise ValueError(
-                "action expected_frame_id conflicts with the top-level value"
+        raw_actions = normalized.pop("actions", None)
+        if raw_actions is not None:
```
rogercloud: `.pop("actions", None)` followed by `is not None` cannot distinguish an explicit `{"actions": null}` from a missing key. Reproduced: `{"actions": null}` validates cleanly and silently defaults to `action=observe` instead of raising. Consider `if "actions" in value:` before the pop so an explicit null hits the existing "actions must contain exactly one action object" error.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -31,65 +44,130 @@
...
+            nested_frame_id = nested.pop("expected_frame_id", None)
+            if nested_frame_id is not None:
+                top_level_frame_id = normalized.get("expected_frame_id")
+                if top_level_frame_id is None:
+                    normalized["expected_frame_id"] = nested_frame_id
+                elif top_level_frame_id != nested_frame_id:
+                    raise ValueError(
+                        "action expected_frame_id conflicts with the top-level value"
+                    )
+            for key, item in nested.items():
```
rogercloud, line 125: `expected_frame_id` gets explicit conflict detection just above, but every other key merges with `setdefault`, so the top-level value silently wins. Reproduced: `{"target":"outer-1","actions":[{"action":"click","target":"inner-2"}]}` resolves to `outer-1` and discards `inner-2` with no error. Since `target` selects which element gets clicked, this produces a wrong action rather than a failure — worse than the fail-fast treatment `expected_frame_id` already receives. Worth applying the same conflict check to `target` at minimum.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -31,65 +44,130 @@
...
+class ComputerToolArgs(BaseModel):
+    """One flat computer action.
 
+    The public schema intentionally avoids a one-element ``actions`` array.
+    Several otherwise capable tool-calling models serialize that nested shape
+    as a string. The pre-validator still accepts the preview shape so existing
+    traces and clients remain replayable.
+    """
+
+    model_config = ConfigDict(extra="forbid")
```
rogercloud, line 56: `extra="forbid"` is new in this PR (no `model_config`/`ConfigDict` at the base commit). Any legacy payload carrying a stray top-level key now hard-fails validation instead of being ignored. That's a defensible tightening, but it qualifies the PR's compatibility claim — the one-element `actions` envelope still works, yet payloads with extra sibling keys do not. Worth calling out in the Compatibility section if any replayable trace or existing client emitted extra top-level fields.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -237,23 +395,112 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+        public_file_ref: dict[str, Any] | None = None
+        inline_markdown: str | None = None
+        if action.type is ComputerActionType.SCREENSHOT:
+            try:
+                public_file_ref = await asyncio.to_thread(
+                    self._publish_screenshot,
+                    observation,
+                )
+            except (
+                FileNotFoundError,
```
rogercloud: Two things here: `FileNotFoundError` is redundant alongside `OSError` on the next line, and unlike the sibling handler at 381-399 this block has no trailing catch-all. SQLAlchemy errors from the DB-backed publish path, `StorageKeyScopeError`, and `ImportError` from lazy imports all escape. It does not crash the task (react.py's outer handler catches it), but it loses the "Could not publish computer screenshot" framing and the `frame_id` context this handler adds.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -237,23 +395,112 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+            raise FileNotFoundError("computer observation screenshot is unavailable")
+        source_path = Path(source).resolve()
+        output_dir = Path(workspace.output_dir).resolve()
+        output_dir.mkdir(parents=True, exist_ok=True)
+        frame_key = hashlib.sha256(observation.frame_id.encode("utf-8")).hexdigest()[
+            :12
+        ]
+        suffix = source_path.suffix or ".png"
+        destination = output_dir / f"computer-screenshot-{frame_key}{suffix}"
+        if destination.exists():
```
rogercloud: This fallback is unreachable: `destination` is keyed on `sha256(frame_id)` and `frame_id` is a fresh UUID4 per observation (`browser.py:218`), with reuse rejected outright at `environment.py:83-88`. So `destination.exists()` is never true and the copy below writes a new file on every `screenshot` call, including byte-identical re-verification shots, with no cleanup. These are public files that reach `file_outputs` and `GET /workspace/{task_id}/output`. Either drop the dead branch, or key the filename on the content hash directly so it actually dedups.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -121,25 +213,67 @@ def name(self) -> str:
...
-        return """Inspect and control an isolated temporary browser through screenshots.
-
-        First request a screenshot without expected_frame_id. Inspect the returned
-        image, then send exactly one action with that frame_id as expected_frame_id.
-        Every successful call returns a fresh screenshot and frame_id; do not request
-        a second screenshot after an action.
+        perception_instructions = {
+            ComputerPerceptionMode.VISION: (
+                "Plan from the screenshot and use normalized coordinates. "
+                "Semantic elements are intentionally not exposed in this mode."
```
rogercloud: "Semantic elements are intentionally not exposed in this mode" is not implemented — nothing filters `observation.elements` by perception mode. Latent today since the only call site never sets `perception_mode`, but the sentence would be false the moment VISION mode is wired up.

src/xagent/core/computer/browser.py
```diff
@@ -314,6 +316,7 @@ async def _read_interactive_elements(
                         element_id=f"dom-{index}",
                         source=ComputerElementSource.DOM,
                         bounds=raw_element["bounds"],
+                        surface=ComputerElementSurface.DOCUMENT,
```
rogercloud, line 319: This is the only place any adapter assigns `surface`, and it's always `DOCUMENT` — `APPLICATION_CHROME`/`NATIVE_APP`/`OVERLAY` are never produced. Nothing in production code reads `ComputerElement.surface` either; it currently reaches the model only as serialized data plus prompt text describing values that can't appear. Worth trimming the tool description's `application_chrome` guidance until a second adapter exists, or noting this is forward-looking scaffolding.

src/xagent/core/computer/schema.py
rogercloud, line 148: `APPLICATION_CHROME` / `NATIVE_APP` / `OVERLAY` are never produced by any adapter — `browser.py:319` hardcodes `DOCUMENT`. The tool description nonetheless teaches the model what `application_chrome` means and how to fall back when no `document` element matches, which is guidance for a value that cannot currently appear. Simplification: drop the three unused members until an adapter needs them.

tests/core/computer/test_environment.py
```diff
@@ -57,6 +58,8 @@ def observation(
                     element_id="button-1",
                     source=ComputerElementSource.DOM,
                     bounds=NormalizedRect(x=0.1, y=0.1, width=0.2, height=0.1),
+                    surface=ComputerElementSurface.DOCUMENT,
+                    metadata={"surface": "document"},
```
rogercloud: `metadata={"surface": "document"}` duplicates the typed `surface` field on the line above, and nothing reads either since the surface-consuming validator was removed in `055fcb7e`.

rogercloud (approved):
## Re-review summary

Since the last review at `cf8df9b3d`, two commits landed and they resolve the bulk of the prior findings by deletion rather than by patching. `055fcb7eb` removed the click-stall detection heuristic outright (mooting all 6 findings against it), removed the session-wide `understand_media` suppression from ReAct tool-schema routing (mooting 1 major finding plus a related body concern), removed the target-side `surface` field and its `validate_action_batch` enforcement (mooting 2 design-level and 2 inline findings), and wrapped `_publish_screenshot` in `asyncio.to_thread`, fixing the blocking-I/O major. `eb090727a` is docs-only, clarifying that the tool boundary normalizes bare/JSON-encoded element IDs for compatibility, which fixes the self-contradiction minor.

Net: the PR got meaningfully smaller and the highest-severity items are genuinely gone. What remains are edge cases in the new flattened-schema validator, two cross-tool prompt inconsistencies, and a set of mechanisms that are documented but inert against the only shipped adapter. **No blocking issues remain.**

---

## Prior findings

### FIXED

| # | Finding | Evidence |
|---|---|---|
| 1 | 3 Major + 3 Minor click-stall findings (stall-reset-backwards, point-target key collapse, click recorded before execute with no rollback, stale history on session recreation, radius not aspect-corrected, bounds-jitter evasion) | Entire `_ClickAttempt` / `_recent_clicks` / `_record_click_attempt` mechanism deleted in `055fcb7e`. `grep -rn "_record_click_attempt\|_recent_clicks\|_ClickAttempt\|_CLICK_STALL\|_CLICK_HISTORY" src/` returns zero matches. |
| 2 | **Major** — session-wide `understand_media` suppression | `_tool_schemas_for_resolved_llm` and its `llm_supports_vision` import fully removed from `react.py`; both call sites now use unfiltered `base_tool_schemas`, and the 3 tests exercising the filter were deleted. |
| 3 | **Major** — blocking sync I/O on the event loop | `src/xagent/core/tools/adapters/vibe/computer.py:405-408` now dispatches the whole method through `asyncio.to_thread`, covering `read_bytes`, `shutil.copy2`, and `build_workspace_file_ref`'s synchronous `workspace.register_file`. |
| 4 | Body-level — semantic-surface validation is a no-op / leniency-strictness deadlock | `055fcb7e` removed the `surface` field from `ComputerTarget` (now only `element_id`/`point`) and stripped the enforcement from `validate_action_batch`. `ComputerElement.surface` is retained, confirming the observation-side/target-side split was deliberate. |
| 5 | Minor — undocumented compatibility break (`{"element_id": "dom-3"}` without `surface` hard-rejected) | Moot; `surface` is no longer accepted or required on targets at all. |
| 6 | Minor (nit) — `surface` on a `point` target raising a raw pydantic `ValueError` | Moot; field removed. |
| 7 | Minor — self-contradictory tool description about bare-string targets | Fixed by `eb090727a`: "Models should emit these structured forms; the tool boundary only normalizes unambiguous bare or JSON-encoded element IDs for compatibility." |

### NOT FIXED (carried forward, non-blocking)

- **Minor** — `screenshot`/`observe` still skip both the `expected_frame_id` staleness check and the `supported_actions` capability check that other action types go through. Latent today (see D-1 below) but worth acknowledging since capability gating is emphasized more in the updated description.
- **Open suggestion, not a defect of this PR** — ReAct deliverable verification is still prompt text with no programmatic gate (`missing_verification` only validates the model's own self-reported free-text field; `delivery`/`user_visible_artifact`/`private_observation` labels are never read back to gate `final_answer`). This PR never claimed to add that gate — follow-up, not required here.

### Simplification findings still open

- `computer.py:187-206` — `perception_mode` / `environment_label` / `environment_instructions` ctor params still have zero production callers (`browser_use.py:1141` passes only `task_id`/`workspace`).
- `src/xagent/core/computer/schema.py:148-150` — `APPLICATION_CHROME` / `NATIVE_APP` / `OVERLAY` still never produced by any adapter (only `DOCUMENT` is, at `browser.py:319`).
- `computer.py:483-494` — the frame-key-then-content-hash collision fallback is still unreachable (see F-14 below).

---

## New findings from this pass

All verified independently; the validation edge cases were confirmed by direct `ComputerToolArgs.model_validate` reproduction.

### F-2 (moderate) — new completion rule is written in computer-tool vocabulary but applies to every tool

`react.py:2461-2466` adds: *"A requested file or media output is complete only when a tool result explicitly marks it as user-visible and returns its artifact reference."* The "explicitly marks it as user-visible" contract is specific to `ComputerToolResult`'s `delivery` field — no other production tool (image, workspace_file, audio, music, sound_effect, video, pptx, python/javascript executors, sql, browser_use screenshot/pdf, download_web_asset, vision) has an equivalent discriminator; they signal visibility only implicitly through `file_ref`/`markdown_link`/`artifacts`.

Realistic impact is soft: the rule only fires behind the stuck-loop guard (4 consecutive same-decision-group calls), and a reasonable model would likely accept a populated `file_ref` + `markdown_link` as satisfying it — worst case is one wasted tool-call iteration, not a refusal to finalize. Worth tightening: either scope the wording to computer-tool results, or generalize it to accept `file_ref`/`markdown_link` presence as "user-visible."

### F-3 (moderate) — `inline_markdown` competes with `file_ref.markdown_link` in the same result

`inline_markdown` (image-embed form `name`) is attached alongside `file_ref.markdown_link` (plain-link form `[name](file:...)`) in the same tool result, while the global file-ref guidance says "Prefer an existing markdown_link value when one is present." Both instruction sets reach the same prompt. The tool-local instruction is more specific and more recent, which likely wins most of the time, but the failure mode is a screenshot rendered as a bare link instead of inline. Cheapest fix: omit `markdown_link` from computer's `file_ref` payload, or state explicitly "use `inline_markdown`, not `markdown_link`, for this result."

### Latent-only / low-priority (noted, not blocking)

- **D-1** — capability gating is currently a permanent no-op: `browser.py:235` returns every action type as supported unconditionally, so the consumer check can never reject anything.
- **D-2** — surface provenance is a hardcoded constant (`DOCUMENT`) in the only shipped adapter and is never read in production code.
- **F-13** — the new `ComputerActionType.OBSERVE` branches in `browser.py` are unreachable; `ComputerTool` never routes OBSERVE through `environment.execute()`.
- **F-12** (nit) — `ComputerElementSurface`/`ComputerPerceptionMode` aren't re-exported in `computer/__init__.py`'s `__all__`, unlike sibling enums.
- **F-4** (minor, latent) — when `screenshot_fresh` is `False` (never set in production today, only by a test fake), the result strips the new message's image while also superseding (and thus stripping) the previous message's image.
- **D-4** — the tool description grew ~828→2626 chars (~3.2x), partly documenting behavior (D-1/D-2) the shipped adapter doesn't exhibit.
- **F-10** — not a bug: `_publish_screenshot` calls `resolve_file_id` instead of `resolve_file_id_detached`, but the only file_id this path receives is always `internal-`-prefixed, which short-circuits before any session-bound code runs.
- **F-15** (test nit) — a test fixture sets a redundant `metadata={"surface": "document"}` alongside the typed field; nothing reads it since the consumer was deleted.

---

```
computer.py:187-206  yagni   perception_mode/environment_label/environment_instructions ctor params have zero production callers (browser_use.py:1141 passes only task_id/workspace). Drop the params and inline the default description.
schema.py:148-150    yagni   APPLICATION_CHROME/NATIVE_APP/OVERLAY are never produced by any adapter (browser.py:319 hardcodes DOCUMENT). Drop the three members and the prompt prose that references them.
computer.py:488-492  shrink  Frame-key-then-content-hash collision fallback is unreachable; frame_id is UUID4-unique per observation and environment.py:83-88 rejects reuse. Delete the branch, or re-key the filename on content hash so it actually dedups.
```

`net: ~-30 lines possible`

---

The two core mechanisms are sound and well covered by tests: the flattened action schema with backward-compatible legacy-envelope normalization, and the observe/screenshot split with FileRef publication. Deleting the click-stall heuristic and the vision-schema filter in `055fcb7e` was the right call and removed the riskiest surface in the PR.

Remaining reservations: (1) three of the six claimed hardening mechanisms are currently no-ops against the only shipped adapter (capability gating, surface provenance, perception-mode-driven description), inflating the tool description with currently-inert guidance; (2) the new ReAct completion rule and `inline_markdown` messaging were written against computer-tool-specific semantics without checking fit against the other tools/the existing `markdown_link` convention (F-2, F-3), producing two soft-but-real cross-tool inconsistencies. None of these block merge.

The new validator edge cases (F-5 through F-9, F-14 — see inline comments) are all minor, with the silent `target` drop under a mixed nested/top-level envelope being the one most worth a follow-up fix since it turns a malformed request into a wrong action rather than a visible error.

---

## Tests

225 tests pass across `tests/core/computer` and `tests/core/agent/test_react.py`, re-verified independently in this pass. The new validator edge cases (F-5/F-6/F-7/F-8, called out inline) have no test exercising their exact failure inputs — a small parametrized table over `{JSON-encoded string target, actions: null, conflicting nested target, extra top-level key}` would close that gap.

Approving — no blocking issues found. The inline notes below are all non-blocking suggestions for this PR or good follow-up material.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in bf8d5cb8. JSON-encoded strings now reuse the decoded element ID, mappings remain supported, and arrays, numbers, booleans, and null are rejected instead of becoming literal element IDs. Added regression coverage for all of those forms.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in bf8d5cb8. Nested normalization now rejects every conflicting duplicate field, including `target`, rather than silently preferring the top-level value. Matching duplicate values remain compatible. Added a target-conflict regression test.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in bf8d5cb8. Normalization now checks key presence before popping `actions`, so an explicit null reaches the existing one-action validation error. Added a regression test.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in bf8d5cb8. VISION mode no longer claims that elements are filtered from the observation. It now tells the model to ignore semantic IDs if an adapter includes them as auxiliary data, matching the actual contract.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Fixed in bf8d5cb8. Screenshot publication now catches unexpected `Exception` subclasses at the tool boundary and preserves the `Could not publish computer screenshot` framing plus `frame_id`. Added an `ImportError` regression test.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Keeping `extra="forbid"` as the intentional strict boundary. The PR Compatibility section now states that unknown top-level fields and conflicting duplicate values are rejected, and bf8d5cb8 adds a regression test that pins this behavior.

src/xagent/core/computer/schema.py
qinxuye: Not changing the enum vocabulary here. These values are the provider-neutral observation contract needed by the next local/native adapter slice; deleting and reintroducing them would create avoidable schema churn. The current open-source browser adapter producing only `DOCUMENT` is an acknowledged boundary, not a claim that all values are already emitted.

src/xagent/core/tools/adapters/vibe/computer.py
qinxuye: Partially accepted in bf8d5cb8: the unreachable destination-collision/content-hash fallback was removed. Frame-keyed names remain intentional because each explicit screenshot represents a distinct user-visible observation; switching to content-hash identity would add a full source read and alias separate deliveries. Cross-artifact deduplication is outside this protocol slice.

src/xagent/core/computer/browser.py
qinxuye: Not removing the provenance field or forward-looking guidance in this foundation slice. The current browser adapter emits `DOCUMENT`; the registered local/native adapters in the following slice require the remaining stable values. Observation-side provenance is deliberately not an action authorization field, and the PR scope states that this contract serves current and future adapters.

tests/core/computer/test_environment.py
qinxuye: Fixed in bf8d5cb8 by removing the duplicate test-only `metadata={"surface": "document"}` value; the typed `surface` field is now the single fixture source.

qinxuye:
Also addressed the two cross-tool observations from the review summary in bf8d5cb8: ReAct completion now applies the explicit user-visible marker only to computer screenshots while accepting artifact/file_ref/markdown_link evidence from other media tools, and computer screenshot results explicitly prefer inline_markdown over file_ref.markdown_link. The focused Computer Use and ReAct suite passes with 235 tests.

PR #1209, opened 2026-08-09
feat(interactions): add the task interaction request model

### LINKED ISSUE #1074: feat(interactions): add canonical task interaction schema

AlexLiu190625:
## What

Adds `task_interaction_requests`: the canonical durable record for one blocking task interaction (an agent question waiting for a user answer). Storage only — this PR ships the model and its constraint tests. The migration lands separately, and no reader or writer is wired: the table is deliberately inert until the staging primitive that writes it arrives.

The constraint set, in prose:

- **One active question per task**, enforced by a unique constraint over `(task_id, active_slot)` where `active_slot` is `1` on active rows and SQL `NULL` on terminal rows — NULL-distinct semantics let any number of terminal rows coexist while a second active row is rejected, verified on both backends.
- **Status and slot move together**: paired CHECKs make `active_slot = 1 ⇔ status = 'active'`, so a row cannot be terminal while still occupying the slot or vice versa.
- **The resume anchor cannot be lost while a question is answerable**: the FK to the checkpoint row is `ON DELETE SET NULL`, paired with a CHECK requiring an anchor on active rows — deleting an active row's anchor fails on the CHECK name on both backends; terminal rows degrade to NULL as intended.
- **Run-scoped identity**: uniqueness over `(task_id, run_id, request_idempotency_key)`, so a retry within one run replays and a new run is never poisoned by an old key.
- **Lowercase status vocabulary stored as a plain `String` rather than a native enum**, matching the storage form of `TaskExecutionCommand.status` — the repo's `Enum()` columns, such as `Task.status`, persist uppercase member names instead, which is a different convention and not the one wanted here. The closed vocabulary is then enforced by an IN-list CHECK, which goes deliberately beyond that precedent: `TaskExecutionCommand.status` carries no CHECK and validates in the application layer. The reason for diverging is the failure mode on the consuming side — an unrecognised command status is refused by a dispatcher that can see it, whereas an unrecognised interaction status would already be occupying the active slot with the task pinned in `WAITING_FOR_USER` and no reader able to render it. A native enum type would be enforced on PostgreSQL but degrades to an unconstrained VARCHAR on SQLite; `String` + CHECK gives equal strength on both backends.
- Answered/unanswered payload pairing, responder-identity pairing, empty-string rejections, and expiry-after-creation direction checks complete the set.

## Why

#1074 asks for exactly this record. Covered here:

- bullets 1, 2, 4, and 5 in full (the table, the single-active-slot rule, the anchor pairing, the deletion semantics at the model level);
- bullet 3's storage half — the unique constraint and CHECK set. The runtime half (the writer that enforces fail-closed staging) arrives with the first writer;
- bullet 7's column-shape half — SQL `NULL` rather than JSON `null` for the absent response payload (`none_as_null` on both JSON columns — the request side so NOT NULL actually fires, the response side to distinguish no-answer from a JSON-null answer — pinned by a metadata-wide test), and timezone-aware column declarations throughout. The read-back half needs a reader to demonstrate and lands with the first consumer.

This series does not close #1074: the existing-table field pairs and the datetime read-back half remain on the issue, as discussed there.

## Constraint notes

- The single-active-slot unique equals the issue's "per root Task" wording only under today's model where one agent tree is one task row; the schema has no root concept. If a root discriminator column is ever added, it must be a `NOT NULL` sentinel (empty string), not a nullable column — NULL-distinct semantics would otherwise void the root-side constraint entirely.
- The longest constraint name is 61 characters against PostgreSQL's 63-character identifier ceiling; a structural test pins the ceiling so a future rename or added column cannot silently exceed it (SQLAlchemy raises at CREATE TABLE time for explicit names, so an overflow would otherwise surface only on the first PostgreSQL deploy).
- Timestamp binding is a portable caller obligation: bind timezone-aware UTC values. SQLite silently stores an aware non-UTC value at local wall clock; PostgreSQL silently misresolves a naive value whenever the session TimeZone is not UTC. A storage-form test pins the aware-UTC round trip.

## Verification

- Dual-backend constraint suites: SQLite 43 passed; PostgreSQL 41 passed (run against a disposable database; the PostgreSQL file skips cleanly when no test database is configured).
- Migration-integration gate, before vs after adding the model to the metadata: identical results (`2 failed, 5 passed`), with byte-identical failure text on the two failures. Both are pre-existing: that suite's own teardown drops `mcp_servers_legacy` without CASCADE while four foreign keys still reference it. Unrelated to this change and not fixed here; `task_interaction_requests` appears zero times in that suite's output.
- `tests/web` full run: every pre-existing outcome unchanged; the delta is exactly the new test files.
- Alembic heads unchanged (no migration in this PR).

## Why six resume-locator columns

The locator group looks redundant at first glance, so here is the division of labor, for reviewers:

- **`resume_trace_event_id`** is the load-bearing pointer: an FK to the exact checkpoint row, `ON DELETE SET NULL`, paired with the CHECK that requires it on active rows. Answering a question resolves the suspended checkpoint through this column and nothing else.
- **`resume_event_id`, `resume_execution_id`, `resume_run_partition` are deliberate, permanent redundancy for terminal rows.** The SET NULL pairing protects only active rows; once a row is answered or terminated, its anchored checkpoint row may legitimately be pruned by retention, and the FK goes NULL. Protecting terminal rows too would let historical interactions pin checkpoint rows forever and grow the retention window without bound — so instead, the logical identifiers are captured into the row at ask time and survive the anchor's deletion. They are the audit trail of *which* checkpoint the question was asked against, readable after the pointed-to row is gone. They are not a fallback resolution path for answering: an active row always has the FK.
- **`resume_checkpoint_type`** records which checkpoint encoding the anchored row carries, so a future reader decodes without sniffing.
- **`resume_locator_format`** is honest future-proofing: a versioned discriminator whose IN-list holds a single value today (`trace_event_pk_v1`). If the locator scheme ever changes shape, existing rows stay interpretable. This is the one column in the group carried for a future that may not arrive; it costs a VARCHAR(32) and one CHECK.

`run_id` and `resume_run_partition` intentionally coexist without a DB-level equality CHECK: forcing them equal at the constraint layer would turn a detectable upstream corruption (a checkpoint row whose partition disagrees with the asking run) into an un-insertable row misclassified as a constraint bug. The writing primitive compares them in Python instead, where the mismatch can be classified and reported as what it is.


src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,306 @@
...
+        # test_no_constraint_name_exceeds_the_postgres_identifier_limit for
+        # the structural guard against this happening again.
+        CheckConstraint(
+            "(responded_at IS NULL) = (responder_identity IS NULL)",
+            name="ck_task_interaction_requests_responder_pairs_responded_at",
+        ),
+        CheckConstraint(
+            "expires_at > created_at",
+            name="ck_task_interaction_requests_expiry_after_creation",
+        ),
```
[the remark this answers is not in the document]
AlexLiu190625: Confirmed the mechanism by probing the real model on both backends, and it behaves as described on SQLite: with `created_at` coming from `server_default` and `expires_at` bound from Python, the stored values are `'2026-08-09 02:39:36'` and `'2026-08-09 02:39:36.001000'`. The comparison is lexicographic and the shorter string is a prefix of the longer one, so an already-expired row is accepted. Measured, the window is bounded — a TTL inverted by up to one second gets through, while -1500ms and -15min are both rejected.

Two corrections to the framing. First, PostgreSQL is not the clean contrast: `func.now()` is `transaction_timestamp()`, not the statement clock, so the effective guard there is `expires_at > transaction_start` and its window equals the writing transaction's age — potentially wider than SQLite's one second. It does reject on precision (verified down to 1µs on a fresh transaction), but it is not a full guard either. Second, the second-precision consequence was already recorded in the class docstring, which states the CHECK stops protecting against a zero-or-negative TTL on SQLite once a TTL goes sub-second.

What was genuinely worth fixing is smaller and different from what was reported: the docstring called `server_default=func.now()` "the database's transaction-start clock", and that is wrong for SQLite — `CURRENT_TIMESTAMP` advances per statement, verified inside one open transaction. The text form of the comparison was also left implicit. Both are now fixed: the docstring names each backend's clock correctly, describes the text-comparison behaviour and its one-second bound, notes PostgreSQL's transaction-age slack, and states that the writing primitive must reject a non-positive TTL in Python before it reaches SQL — the same division of labour this file already uses for the deleted run-partition check.

Declining the constraint rewrite: a `julianday()` form would need a dialect-conditional constraint, and the model and its follow-up migration must carry identical constraint names, so the branching would propagate — too much surface to close a sub-second window no realistic TTL enters. Dropping `server_default` is also the wrong trade: it would replace one authoritative database clock with per-host application clocks on a durable record, against the convention of every other `created_at` in the package.

tests/web/services/test_task_interaction_schema_postgresql.py
```diff
@@ -0,0 +1,764 @@
...
+    assert reloaded.responded_at == original_responded_at
+
+
+# --------------------------------------------------------------------------
+# T-store: storage-form sentinels
+# --------------------------------------------------------------------------
+
+
+def test_response_payload_none_lands_as_sql_null(db_session, fixtures) -> None:
+    """Expiry is only authoritative for reclaim -- see the SQLite half."""
```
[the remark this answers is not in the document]
AlexLiu190625: Confirmed, and it was worse than a stale label — the sentence described a different invariant entirely, lifted from an unrelated line of the model class docstring during an earlier comment cleanup. The test body only asserts the `none_as_null` contract: an unset `response_payload` must satisfy `IS NULL`, while `{"answer": None}` must not, and must round-trip intact. Expiry is never touched. Replaced with a summary of what the assertions actually check, in this file's short-summary-plus-pointer style.

Also checked every other `-- see the SQLite half` pointer in this file against its counterpart by name; this was the only mismatch. Two nearby docstrings referenced numbered list items that resolve to nothing in the tree — each now states the invariant it stood for instead, in the same commit.

tests/web/services/test_task_interaction_schema.py
```diff
@@ -0,0 +1,913 @@
...
+    inspector = inspect(get_engine())
+    columns = {c["name"]: c for c in inspector.get_columns("task_interaction_requests")}
+    assert set(columns) == EXPECTED_COLUMNS
+    for name, expected_nullable in EXPECTED_NULLABLE.items():
+        assert columns[name]["nullable"] == expected_nullable, name
+    for name, expected_length in EXPECTED_STRING_LENGTHS.items():
+        assert columns[name]["type"].length == expected_length, name
+
+
+def test_all_timestamp_columns_are_timezone_aware(db_session) -> None:
```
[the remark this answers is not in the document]
AlexLiu190625: Correct for this test. Its body only reads the statically declared table metadata, so the fixture was unused both by name and by side effect — the assertions hold with no database initialized at all, and dropping the parameter skips a real create_all/drop_all cycle over the full metadata.

Scoped deliberately to this one test: the shape and inventory tests in both files also never name `db_session`, but they reflect through `inspect(get_engine())` and depend on the fixture's init side effect, and the PostgreSQL twin of this same test additionally takes its skip-if-unconfigured gate from that fixture. Those keep the parameter. The change matches the file's existing convention, where the other metadata-only tests already take no fixture.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,322 @@
...
+        DateTime(timezone=True),
+        server_default=func.now(),
+        onupdate=func.now(),
+        nullable=False,
+    )
+    # Authoritative only for reclamation statements; readers must not hide a
+    # row based on this column (see class docstring).
+    expires_at = Column(DateTime(timezone=True), nullable=False)
+    responded_at = Column(DateTime(timezone=True), nullable=True)
+    terminated_at = Column(DateTime(timezone=True), nullable=True)
```
rogercloud, line 377: **Major:** `terminated_at` is referenced by zero CHECK constraints -- confirmed against the exhaustive `EXPECTED_CHECK_CONSTRAINT_NAMES` fixture in `src/xagent/web/services/task_interaction_schema_shared.py:305-327`, which lists all 21 constraints and none touch this column. That is inconsistent with `responded_at` and `terminal_reason`, which are both paired with `status` in both directions. Today a row can be `status = 'active'` with a non-null `terminated_at`, or `status = 'terminated'` with a null one, and nothing rejects it. Since the file's docstring justifies every other asymmetric choice explicitly but is silent here, this looks like an oversight rather than a decision.

Suggested fix: add `CheckConstraint("(status = 'terminated') = (terminated_at IS NOT NULL)", name="ck_task_interaction_requests_terminated_at_pairs_status")`, plus a covering test in both suites.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,322 @@
...
+        ForeignKey(
+            "users.id",
+            ondelete="SET NULL",
+            name="fk_task_interaction_requests_responder_user_id",
+        ),
+        nullable=True,
+    )
+    # "user:{id}" / "guest:{guest_id}" -- the prefix is a namespace, not
+    # decoration.
+    responder_identity = Column(String(255), nullable=True)
```
rogercloud, line 361: **Minor:** `responder_identity` is excluded from the empty-string CHECK block, though it has the same semantics as the five columns that do get one (a meaningful namespaced string, meaningless when empty) and the same stated exposure ("NOT NULL alone does not stop an `or ''` write"). It also participates in a pairing CHECK with `responded_at` that an empty string satisfies trivially, since `'' IS NOT NULL`. Reproduced: inserting `responder_identity=''` with `responded_at` set raises no `IntegrityError`, and no test covers it.

Suggested fix: add `CheckConstraint("responder_identity <> ''", name="ck_task_interaction_requests_responder_identity_nonempty")` and a test.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,322 @@
...
+    behaves like RESTRICT.
+
+    ``response_payload`` uses ``JSON(none_as_null=True)`` -- the only column
+    in the repository with that flag -- so a Python ``None`` lands as SQL
+    NULL instead of the JSON scalar ``null``. Consumers must test "has this
+    request been answered" with SQL ``IS NULL``, never with
+    ``payload is None`` in Python, since a legitimate JSON ``null`` answer
+    value must still read as answered.
+
+    ``terminal_reason`` vocabulary: the three members each have a named
```
rogercloud: **Minor (docs):** This paragraph says in the present tense that the three `terminal_reason` members "each have a named writer path," and that other candidates are excluded because they "have no writer yet." This PR ships zero writers -- a repo-wide grep finds no writer code for this table -- and the docstring's own opening paragraph correctly describes the model as "model-only, zero production consumers."

Suggested fix: reword to intended/future tense, e.g. "each is intended to have a named writer path."

tests/web/services/task_interaction_schema_shared.py
```diff
@@ -0,0 +1,352 @@
...
+    "terminal_reason": True,
+    "created_at": False,
+    "updated_at": False,
+    "expires_at": False,
+    "responded_at": True,
+    "terminated_at": True,
+}
+
+# column -> VARCHAR length, for the string columns the task book pins.
+EXPECTED_STRING_LENGTHS: dict[str, int] = {
```
rogercloud, line 349: **Minor (test coverage):** `EXPECTED_STRING_LENGTHS` omits five string columns: `kind`, `status`, `terminal_reason`, `resume_locator_format`, and `resume_checkpoint_type`. Because `test_created_columns_match_the_frozen_shape` (both suites) iterates `EXPECTED_STRING_LENGTHS.items()` rather than all string columns from table metadata, those five widths are silently unpinned -- an accidental length change to any of them would not fail the frozen-shape test, which is exactly what that test exists to catch.

Suggested fix: add the five missing entries. Optionally also assert that this dict's key set equals the set of string columns in the table, so a future column cannot be forgotten.

tests/web/services/test_task_interaction_schema.py
```diff
@@ -0,0 +1,913 @@
...
+    interaction kind.
+    """
+    task_id, anchor_id = fixtures
+    row = make_row(
+        task_id=task_id, resume_trace_event_id=anchor_id, kind="unknown_kind"
+    )
+    assert_rejected(db_session, row, "ck_task_interaction_requests_kind")
+
+
+def test_origin_accepts_a2a(db_session, fixtures) -> None:
```
rogercloud: **Minor (test coverage):** The `origin` vocabulary has six members (`internal`, `sdk`, `a2a`, `trigger`, `widget`, `shared_link`), but only `'a2a'` is asserted as accepted here, with `'internal'` covered implicitly via the default. `sdk`, `trigger`, `widget`, and `shared_link` are never asserted as accepted in either suite, so a typo in any of those four IN-list entries would pass CI.

Suggested fix: parametrize over the full vocabulary and assert each member inserts successfully.

rogercloud (changes requested):
### PR summary

This PR adds `task_interaction_requests`, a new SQLAlchemy model that is the durable record for a single blocking task interaction (an agent question waiting on a user answer). It is deliberately storage-only: the model is registered on metadata, but no reader or writer is wired up yet and the Alembic migration lands in a future PR. The diff is `src/xagent/web/models/task_interaction.py` (322 lines, new), a 2-line registration in `src/xagent/web/models/__init__.py`, a 352-line shared test helper (`tests/web/services/task_interaction_schema_shared.py`), and two constraint-pinning suites (`tests/web/services/test_task_interaction_schema.py`, SQLite, 913 lines; `tests/web/services/test_task_interaction_schema_postgresql.py`, PostgreSQL, 767 lines). It covers the table plus slot/anchor/idempotency constraints from issue #1074, deferring existing-table field pairs and datetime read-back to follow-ups.

### Update summary

Since the last review round, the substantive schema work (`0136b077`, `8998da01`, `b428cc55`, `01d1b327`) is unchanged; the update is the last two commits. `b0bc2051` rewrites the expiry-CHECK documentation to state the per-backend clock behavior explicitly (SQLite's per-statement clock vs. PostgreSQL's transaction-start clock, and the resulting slack in each), and `fdda83c5` corrects two stale test docstrings and drops an unused fixture parameter. No constraint SQL and no column definitions changed in either commit -- the update is documentation and test hygiene only.

### Round 0 design verdict

**Acceptable with reservations.** The core shape matches issue #1074 closely: one durable row per blocking interaction, `active_slot` + `UNIQUE(task_id, active_slot)` exploiting NULL-distinct semantics, `ON DELETE SET NULL` on the resume anchor paired with an active-row CHECK to synthesize RESTRICT, run-scoped idempotency identity, and `String` + IN-list CHECK instead of a native enum. Every constraint expression was traced and encodes what the description claims.

**The reservation that matters most -- please read before the writer PR.** The "RESTRICT via CHECK" trick collides with an existing code path. `src/xagent/web/services/task_deletion.py`'s `purge_task_rows` bulk-deletes `trace_events` *before* deleting the `tasks` row (around `task_deletion.py:52-58`). Because the resume-anchor FK is `ON DELETE SET NULL` and `ck_task_interaction_requests_active_anchor` requires an anchor on active rows, deleting the trace events of a task that holds an active interaction request raises an `IntegrityError`. This was reproduced empirically. It is **not** a shipped regression and not a blocker for this PR -- no writer exists anywhere in the repo yet (verified by repo-wide grep), so no active row can be created and the path is unreachable. But it will break the moment a writer lands. Whichever PR introduces the writer needs an explicit plan to terminate or clear active interaction requests before a task's `trace_events` are purged; please carry this forward as a requirement rather than discovering it in production.

Two smaller design notes, neither blocking:

- The description says the `String`+CHECK status vocabulary "follows the existing `TaskExecutionCommand.status` precedent." That field is indeed `String(32)`, but it carries zero CheckConstraints -- so this is not following a precedent, it is introducing considerably heavier constraint discipline than anything else in the models package (22 CHECKs here vs. exactly one elsewhere, in `trigger.py`). That is a defensible choice for a durable audit-trail table; only the stated justification is inaccurate and worth correcting in the description/docstring.
- `active_slot` is a denormalized column that costs two CHECKs to keep honest. A partial unique index (`UNIQUE(task_id) WHERE status = 'active'`) expresses the same invariant with fewer moving parts and is supported on both SQLite >= 3.8 and PostgreSQL. The stated reason for preferring `UniqueConstraint` -- identical `get_unique_constraints()` reflection across backends -- is a test-convenience argument rather than a runtime necessity. The current approach is acceptable; the alternative is worth naming.

Separately, `ck_task_interaction_requests_active_protocol` (`task_interaction.py:180-183`) pins `protocol_version = 1` for active rows only, while `kind` is validated globally. This is intentional, tested, and documented (`test_active_row_requires_protocol_version_one`), and only debatable against issue #1074's literal "fail closed on unknown kind/version" wording, which has no explicit terminal-row carve-out. Noting it for the record, not as a finding.

### Prior findings status

All three prior findings from gemini-code-assist are verified fixed.

- [x] **Expiry CHECK lexicographic edge case** (`task_interaction.py:223`) -- fixed in `b0bc2051`. The docstring now distinguishes SQLite's per-statement clock from PostgreSQL's transaction-start clock, spells out the lexicographic mechanism and its ~1-second bound, notes PostgreSQL's own transaction-age slack, and states that the future writing primitive must reject non-positive TTLs in Python. The CHECK SQL was deliberately left unchanged (a dialect-conditional `julianday()` rewrite was declined as not worth the surface area for a sub-second window). Verified: the docstring matches, the SQL is unchanged, and the reasoning holds given no writer exists to carry the Python-side obligation yet.
- [x] **Copy-pasted docstring on `test_response_payload_none_lands_as_sql_null`** (PostgreSQL suite) -- fixed in `fdda83c5`; it now describes the `none_as_null` contract it actually asserts. The author proactively fixed two further docstrings in the same file that referenced numbered list items resolving to nothing; grep confirms no stale numbered-list references remain.
- [x] **Unused `db_session` fixture on `test_all_timestamp_columns_are_timezone_aware`** -- fixed in `fdda83c5`; the parameter is removed and the test passes. Correctly scoped: the PostgreSQL twin keeps its `db_session` because it does live reflection via `inspect(get_engine())` and relies on the fixture's skip-if-unconfigured gate.

### New findings

#### Major

**1. `terminated_at` has no CHECK constraint pairing it with status** -- `src/xagent/web/models/task_interaction.py:322`

`terminated_at` is referenced by zero CHECK constraints. Confirmed via a full constraint inventory, grep, and the exhaustive `EXPECTED_CHECK_CONSTRAINT_NAMES` fixture (`task_interaction_schema_shared.py:305-327`), which enumerates all 21 constraints -- none touch `terminated_at`. This is inconsistent with the structurally analogous `responded_at` (paired both ways with status and `response_payload`) and `terminal_reason` (paired both ways with status). Today a row can be `status = 'active'` with a non-null `terminated_at`, or `status = 'terminated'` with a null `terminated_at`, and nothing rejects it. This reads as a genuine oversight rather than a decision: the file's docstring justifies every other asymmetric design choice explicitly and says nothing about this one.

Suggested fix: add `CheckConstraint("(status = 'terminated') = (terminated_at IS NOT NULL)", name="ck_task_interaction_requests_terminated_at_pairs_status")` plus a covering test in both suites.

#### Minor

**2. `responder_identity` is missing the empty-string CHECK** -- `src/xagent/web/models/task_interaction.py:306`

Five other columns get a `<> ''` CHECK with the stated rationale that "NOT NULL alone does not stop an `or ''` write." `responder_identity` has identical semantics -- a meaningful namespaced string, meaningless when empty -- and additionally participates in a pairing CHECK with `responded_at` that an empty string satisfies trivially (`'' IS NOT NULL`). Reproduced: inserting `responder_identity=''` alongside a set `responded_at` succeeds with no `IntegrityError`, and no test covers it.

Suggested fix: add `CheckConstraint("responder_identity <> ''", name="ck_task_interaction_requests_responder_identity_nonempty")` and a test.

**3. Four of six `origin` vocabulary members are never asserted as accepted** -- `tests/web/services/test_task_interaction_schema.py:227-237`

The `origin` vocabulary has six members (`internal`, `sdk`, `a2a`, `trigger`, `widget`, `shared_link`), but only `'a2a'` is explicitly tested and `'internal'` implicitly via the default. `sdk`, `trigger`, `widget`, and `shared_link` are never asserted as accepted in either suite, so a typo in any of those four IN-list entries would pass CI.

Suggested fix: parametrize over the vocabulary and assert each member inserts successfully.

**4. `EXPECTED_STRING_LENGTHS` silently leaves five string columns unpinned** -- `tests/web/services/task_interaction_schema_shared.py:277-284`

The dict omits `kind`, `status`, `terminal_reason`, `resume_locator_format`, and `resume_checkpoint_type`. Because the consuming test (`test_created_columns_match_the_frozen_shape`, in both suites) iterates `EXPECTED_STRING_LENGTHS.items()` rather than all string columns from table metadata, those five widths are unpinned and an accidental length change to any of them would not fail the "frozen shape" test -- precisely what that test exists to prevent.

Suggested fix: add the five missing entries. Optionally, also assert that the dict's key set equals the set of string columns in the table, so the next added column cannot be forgotten.

**5. Docstring claims writer paths that this PR does not ship** -- `src/xagent/web/models/task_interaction.py:57-64`

The class docstring states in the present tense that the three `terminal_reason` members "each have a named writer path," and that other candidates are excluded because they "have no writer yet." This PR ships zero writers (verified by repo-wide grep -- no writer code exists for this table anywhere), and the docstring's own opening paragraph correctly describes the model as "model-only, zero production consumers."

Suggested fix: reword to intended/future tense, e.g. "each is intended to have a named writer path."

```
L195: shrink: four one-directional CHECKs (answered_pairs_response, unanswered_has_no_response, answered_pairs_responded_at, unanswered_has_no_responded_at) reimplement two biconditionals the same file already writes as single expressions elsewhere. Replace with `(status = 'answered') = (response_payload IS NOT NULL)` and `(status = 'answered') = (responded_at IS NOT NULL)`; the 4 existing test scenarios stay and only need their asserted constraint-name strings updated.
net: -8 lines possible
```

Verified logically equivalent on both backends, including the NULL-status edge case -- `status` is `NOT NULL`, so three-valued-logic concerns do not apply -- via direct SQLite truth-table testing. This mirrors the style already used by `ck_task_interaction_requests_active_slot_pairs_status`.

### Test results

- `tests/web/services/test_task_interaction_schema.py`: **43 passed** (matches the PR's claim).
- `tests/web/services/test_task_interaction_schema_postgresql.py`: **41 skipped** cleanly, with no `XAGENT_TEST_POSTGRES_URL` configured -- matches the PR's claimed skip behavior.

### Scope of this review

All Review Criteria categories were evaluated: code quality, correctness, test coverage, design, and documentation. Beyond the items listed above, no other issues were found. The duplicated SQLite/PostgreSQL suite split and the `resume_locator_format` discriminator column were both examined and are deliberate and justified -- the former follows an established repo-wide convention, the latter is cheap insurance for a durable append-only record -- so neither is raised as a finding.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Confirmed and adopted. `terminated_at` was referenced by zero constraints, while every other column in the terminal lifecycle is pinned: `terminal_reason` both ways against `status`, `responded_at` both ways against both `status` and `responder_identity`. Both bad states you describe were reachable — I inserted an active row carrying `terminated_at` and a terminated row without one, and both backends accepted all of them.

The answered case decides the shape: `answered` and `terminated` are distinct terminal statuses, and nothing in the intended write path stamps `terminated_at` on an answered row — the two statements that terminate a row set `status` and `terminated_at` together, and the one that answers a row sets neither. So the biconditional is exactly right, and it turns that same-statement obligation into a database fact instead of a convention.

Added verbatim (55 characters, under the 63 ceiling), with a both-ways test in each suite including the answered-stays-NULL case, and the row builder's `terminated` default now carries a timestamp. Both suites green: SQLite 50, PostgreSQL 48.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Reproduced — `''` with `responded_at` set was accepted on both backends, exactly as described: the pairing CHECK only asks whether the column is NULL, and `''` is not.

Added `responder_identity <> ''`. The format comment is what makes this unambiguous: the value is namespaced (`user:{id}` / `guest:{guest_id}`) and the prefix is load-bearing, so a value with no prefix cannot be resolved to either space — a row carrying `''` is precisely the "answered by nobody" state the pairing CHECK exists to forbid. The empty-string block's header comment now carries this second rationale, since this is the one nullable column in the block and its failure mode differs from the `or ''` write the others guard against.

There is no writer yet, so this is preventive rather than a live bug — the NULL case is unaffected (`NULL <> ''` is unknown, which a CHECK passes; verified active rows with a NULL identity still insert on both backends).

tests/web/services/test_task_interaction_schema.py
AlexLiu190625: Correct, and fixed. `a2a` plus `internal` via the builder default was two of six; `sdk`, `trigger`, `widget` and `shared_link` were never inserted by either suite, so a typo in any of those four entries would have passed CI — the rejection test only proves some value is refused, and the inventory test compares names, not expressions.

Replaced the single case with a parametrized test over all six. Each member was checked against its actual producer rather than the vocabulary list — `internal` from the column default, `sdk` from the two v1 endpoints, `a2a` from the a2a task creation, `trigger` from the trigger service, `widget`/`shared_link` from the public chat access paths — all six are live. Verified the guard bites: misspelling one IN-list entry turns exactly that parametrized case red. The task source comment also gains the `a2a` member its enumeration was missing.

tests/web/services/task_interaction_schema_shared.py
AlexLiu190625: Confirmed: 7 entries against 12 string columns, and since both suites iterate the dict, the five you named were pinned by nothing — widening `status` to `String(64)` passed both suites as they stood (verified by mutation, red only after the fix).

Filled in all twelve (each width verified against the model declarations), and took the closure suggestion, which is the part that actually matters: the dict's key set is now asserted equal to the set of string columns reflected from metadata, so the next added string column cannot be silently unpinned. The detection predicate — reflected type exposes a non-None `length` — returns the same twelve names on both backends with no false positives among the thirteen non-string columns.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Both halves adopted.

The docstring tense is fixed: each `terminal_reason` member is now described as spoken for by exactly one intended writer, none of which exists yet — the two reclaim branches arrive with the staging primitive, `answered_via_legacy_resume` with the transitional close on the legacy answer path — keeping the point the sentence existed to make, that the vocabulary is closed at three because a member nothing will ever produce only misleads.

On the precedent: you are right, and I verified it — `TaskExecutionCommand.status` is `String(32)` with a default and zero CheckConstraints. What that precedent settles is the storage form (a status column held as a plain String with lowercase literals, as against `Task.status`, a native `Enum` persisting uppercase member names). The IN-list CHECK is a deliberate step beyond it, not inherited from it. The PR body bullet now separates the two decisions and states why this table goes further: an unrecognised command status is refused by a dispatcher that can see it, whereas an unrecognised interaction status would already hold the active slot with the task pinned in `WAITING_FOR_USER` and no reader able to render it.

AlexLiu190625:
Responses to the three body-level items, all verified empirically before answering.

**Purge ordering (the reservation).** Reproduced exactly as you describe — an active interaction row makes `purge_task_rows` fail on `ck_task_interaction_requests_active_anchor` on both backends, while a terminated row purges cleanly (the intended asymmetry). This is a known requirement carried by the deletion-path change that lands after the table's migration: the interaction-row delete goes after the checkpoint-pointer nulling and before the `trace_events` delete, in both purge implementations, with a mutation test that turns both backends red if the delete is removed. Delete rather than terminate-first, since `task_id` cascades and a terminated row would be removed a few statements later anyway — terminating first costs an extra write and leaves the false impression that interaction rows outlive a purge. Thank you for reproducing it rather than reasoning about it.

**The four one-directional pairing CHECKs vs two biconditionals.** The equivalence holds — I ran the full truth table on both backends with zero verdict differences, and it stays equivalent even without `status`'s NOT NULL, so the guard you noted is not load-bearing. Keeping the four for now, for one reason: the first code to hit these constraints is the answer path and its zero-rowcount handling, and on a row violating exactly one direction the current names distinguish "answered without a responded_at" from "responded_at left on a non-answered row" — different bugs, different fixes. The diagnostic is admittedly weaker than it looks (on a row violating both pairs, the two backends already name different constraints), so this is a taste call, and if you feel strongly the change is cheap — no migration references these names yet — and I have no further objection.

**Partial unique index.** The mechanism works — verified that `UNIQUE(task_id) WHERE status='active'` rejects a second active row and permits unlimited terminal rows on both backends. It was set aside for a reflection reason: `get_unique_constraints()` returns empty for a partial index on both backends, so the table's most important invariant would drop out of the one assertion that currently reads identically on SQLite and PostgreSQL; `get_indexes()` does surface it, but the predicate comes back as an opaque clause object on SQLite and a normalized SQL string on PostgreSQL, so no single expectation covers both. Since both backends are NULL-distinct, the plain unique constraint gives the same guarantee without dialect-specific machinery. One accounting note: dropping `active_slot` would also force a rewrite of the protocol-version CHECK, which is scoped through that column — three constraints affected rather than two.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,336 @@
...
+    )
+    run_id = Column(String(64), nullable=False)
+    kind = Column(String(32), nullable=False)
+    protocol_version = Column(Integer, nullable=False)
+    status = Column(String(32), nullable=False)
+    active_slot = Column(Integer, nullable=True)
+    # Frozen at ask time from task.source, falling back to "internal".
+    origin = Column(String(20), nullable=False)
+
+    request_payload = Column(JSON, nullable=False)  # NOT NULL: no none_as_null needed
```
rogercloud: **Major — must fix before a writer lands:** The comment "NOT NULL: no `none_as_null` needed" is empirically false. `nullable=False` only rejects SQL NULL, but SQLAlchemy's `JSON` type without `none_as_null=True` serializes Python `None` into the JSON literal `null` *before* binding — so the value never reaches the NOT NULL check. Repro confirms `request_payload=None` inserts successfully and stores JSON text: `typeof(request_payload) = 'text'`, and `request_payload IS NULL` evaluates false.

This is the exact inverse of the property `response_payload` is careful about (it correctly sets `none_as_null=True`). A future writer doing something as ordinary as `request_payload=payload or None` would silently persist a row that no `IS NULL` predicate can ever find, and there is currently no test covering it.

Suggested fix: add `none_as_null=True` here so `None` becomes a real SQL NULL and the NOT NULL constraint actually fires (or add an explicit CHECK / `@validates` guard if a Python-side rejection is preferred). Please also add a test mirroring the existing `none_as_null` test for `response_payload`, and correct the comment.

tests/web/services/test_task_interaction_schema.py
```diff
@@ -0,0 +1,984 @@
...
+    )
+    raw_stored = db_session.execute(
+        text("SELECT expires_at FROM task_interaction_requests WHERE id = :id"),
+        {"id": non_utc_row.id},
+    ).scalar_one()
+    # The +08:00 instant is 12:00 UTC, the same wall-clock moment as
+    # utc_value above; a correct UTC store would read back "12:00:00". A
+    # bare "20:00:00" instead confirms tzinfo was dropped and the local
+    # wall-clock digits were kept verbatim.
+    assert raw_stored.startswith("2026-06-01 20:00:00"), (
```
rogercloud, line 952: **Major — must fix before a writer lands:** This test and its PostgreSQL twin (`tests/web/services/test_task_interaction_schema_postgresql.py:745`) bind the *identical* aware datetime with a `+08:00` offset (12:00 UTC), then assert contradictory outcomes. The Postgres test asserts the correct UTC-normalized readback (12:00); this one asserts the local wall-clock digits (20:00) — i.e. the offset is dropped rather than converted, which is genuine corruption of the stored instant.

The docstring here already acknowledges the defect ("silently stored as local wall-clock time... the row is accepted with the wrong instant"), but the test still pins the corrupted value via `startswith`. The consequence is that this test would newly **fail** the day the bug is fixed — it encodes the defect as a regression target rather than neutrally documenting current behavior.

Issue #1074 explicitly asks this PR to "reuse the repository's existing datetime normalization boundary for UTC readback": `normalize_datetime_for_db` / `normalize_datetime_from_db` in `src/xagent/web/utils/db_timezone.py`. Both exist and are used elsewhere in the app, but neither is referenced anywhere in `task_interaction.py`.

Recommendation: apply that normalization boundary so both backends agree, or at minimum stop asserting the corrupted value as expected — assert the caller obligation without pinning the wrong number.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,336 @@
...
+        CheckConstraint(
+            "status IN ('active','answered','terminated')",
+            name="ck_task_interaction_requests_status",
+        ),
+        CheckConstraint(
+            "kind IN ('clarification')",
+            name="ck_task_interaction_requests_kind",
+        ),
+        CheckConstraint(
+            "origin IN ('internal','sdk','a2a','trigger','widget','shared_link')",
```
rogercloud, line 188: **Minor:** This IN-list (`internal, sdk, a2a, trigger, widget, shared_link`) is a hand-maintained copy of `Task.source`'s informal vocabulary (`src/xagent/web/models/task.py:221`). `Task.source` is an unconstrained `String(20)` whose values live as bare string literals scattered across `a2a.py`, `public_chat_access.py`, `v1/tasks.py` and `triggers.py` — there is no shared enum or constant. This PR itself had to hand-edit a comment in `task.py` to add `'a2a'`, which is the drift signal.

A future `Task.source` value added without a matching update here will pass review silently and only surface at INSERT time as a raw DB constraint error, once an interaction is staged on a task with that source. Suggest introducing a shared constant/Enum for the vocabulary, or at minimum a cross-file consistency test asserting the two lists agree.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,336 @@
...
+        ),
+        CheckConstraint(
+            "(status = 'active') = (active_slot IS NOT NULL)",
+            name="ck_task_interaction_requests_active_slot_pairs_status",
+        ),
+        CheckConstraint(
+            "status <> 'active' OR resume_trace_event_id IS NOT NULL",
+            name="ck_task_interaction_requests_active_anchor",
+        ),
+        CheckConstraint(
```
rogercloud, line 226: **Minor:** This is the only CHECK touching `protocol_version`, and `active_slot IS NULL OR protocol_version = 1` is vacuous for terminal rows. Confirmed by repro: `terminated` and `answered` rows with `protocol_version = -1` or `0` both insert successfully.

Terminal rows are arguably exempt by design (there is an existing `test_active_row_requires_protocol_version_one`), but the exemption is not documented, and accepting arbitrary negative/nonsense versions on terminal rows is mildly at odds with issue #1074's "fail closed on unknown kind/version". Either widen this to `protocol_version >= 1` unconditionally, or state the exemption explicitly in the class docstring.

tests/web/services/task_interaction_schema_shared.py
```diff
@@ -0,0 +1,362 @@
...
+    "terminal_reason",
+    "created_at",
+    "updated_at",
+    "expires_at",
+    "responded_at",
+    "terminated_at",
+}
+
+# column -> nullable
+EXPECTED_NULLABLE: dict[str, bool] = {
```
rogercloud, line 318: **Minor:** `EXPECTED_NULLABLE` here, and `TIMESTAMP_COLUMNS` at line 295, both lack the closure-against-reflected-metadata assertion that `EXPECTED_STRING_LENGTHS` just gained in this round's fix (`set(EXPECTED_STRING_LENGTHS) == string_columns`). As a result, a newly added nullable or timestamp column would silently escape the "frozen shape" pin — the same class of gap that was just correctly closed for string widths, left open for the other two inventories.

Suggest applying the same `set(...) == <reflected set>` assertion to both, so all three inventories fail loudly when a column is added without updating them.

tests/web/services/test_task_interaction_schema.py
```diff
@@ -0,0 +1,984 @@
...
+        ),
+    )
+
+
+# --------------------------------------------------------------------------
+# T-ck-11..17: paired CHECKs, plus the full-pairing positive control
+# --------------------------------------------------------------------------
+
+
+def test_terminated_status_and_terminal_reason_are_paired_both_ways(
```
rogercloud, line 435: **Minor:** Asymmetric coverage against the structurally identical test just below. This `terminal_reason` "set on the wrong status" rejection only exercises `status='active'`, whereas the `terminated_at` pairing test at lines 401-435 correctly covers both `status='active'` and `status='answered'`. Adding the `answered` case here would bring the two to parity.

src/xagent/web/models/task_interaction.py
```diff
@@ -0,0 +1,336 @@
...
+    String,
+    UniqueConstraint,
+)
+from sqlalchemy.sql import func
+
+from .database import Base
+
+
+class TaskInteractionRequest(Base):  # type: ignore
+    """One blocking interaction request raised by a task and its lifecycle to answer or expiry.
```
rogercloud, line 20: **Minor:** This docstring documents the design decisions thoroughly, but omits one caveat that *is* stated in the PR description: "one active slot per root Task" only holds because one agent tree maps to exactly one Task row today, and `UNIQUE(task_id, active_slot)` would need revisiting if a root-task discriminator column is ever introduced.

A future maintainer will read this docstring, not the PR description — worth adding that one sentence here so the constraint's precondition survives independently.

tests/web/services/test_task_interaction_schema_postgresql.py
```diff
@@ -0,0 +1,832 @@
...
+    # to that shared instant rather than to 20:00.
+    assert non_utc_row.expires_at == utc_value
+
+
+# --------------------------------------------------------------------------
+# T-shape: create_all shape
+# --------------------------------------------------------------------------
+
+
+def test_table_is_registered_in_metadata() -> None:
```
rogercloud: **Minor (simplification):** `test_table_is_registered_in_metadata` is byte-for-byte equivalent to `tests/web/services/test_task_interaction_schema.py:861-868`, and it never touches a database — it only inspects `Base.metadata` in-process. Because of that it also runs unconditionally, regardless of whether `XAGENT_TEST_POSTGRES_URL` is set, so it executes twice on every local run for zero additional coverage.

Suggest deleting this copy and keeping the one in the SQLite file (or hoisting it into the shared module). Same applies to `test_no_constraint_name_exceeds_the_postgres_identifier_limit` at lines 821-833.

tests/web/services/test_task_interaction_schema_postgresql.py
```diff
@@ -0,0 +1,832 @@
...
+    # this literal apply to both backends -- see EXPECTED_NONUNIQUE_INDEXES.
+    nonunique_indexes = {
+        idx["name"]: tuple(sorted(idx["column_names"]))
+        for idx in inspector.get_indexes("task_interaction_requests")
+        if not idx["unique"]
+    }
+    assert nonunique_indexes == EXPECTED_NONUNIQUE_INDEXES
+
+
+def test_no_constraint_name_exceeds_the_postgres_identifier_limit() -> None:
```
rogercloud: **Minor (simplification):** Same as the note above — `test_no_constraint_name_exceeds_the_postgres_identifier_limit` is identical to `tests/web/services/test_task_interaction_schema.py:969-984` and is a pure in-process `__table__` inspection with no database involvement, so it runs twice for no added coverage.

More broadly: this file duplicates 39 of its 41 tests near line-for-line from the SQLite suite. That does not match the repo's actual dual-backend precedent — `test_task_status_storage_postgresql.py` has only 6 tests against its SQLite twin's 33, and all 6 are genuinely backend-specific (native enum reflection, dialect case-sensitivity, real Postgres constraint behavior). It deliberately does not re-run dialect-independent assertions. Restructuring this suite the same way (Postgres-specific tests only, dialect-independent ones parametrized over an engine fixture or kept in one place) would remove roughly 700-800 lines while raising the signal of what remains.

rogercloud (approved):
## Summary

This PR adds `task_interaction_requests` — a new SQLAlchemy model (`src/xagent/web/models/task_interaction.py`, ~336 lines) that is the durable record for a single blocking task interaction (an agent question awaiting a user answer). The change is storage-only: the table is registered on the metadata, but there is no reader or writer yet and the Alembic migration is deferred to a follow-up PR. Supporting changes are the model registration in `src/xagent/web/models/__init__.py`, a comment-only touch in `src/xagent/web/models/task.py`, and two dual-backend constraint suites (`tests/web/services/test_task_interaction_schema.py`, `..._postgresql.py`) plus the shared fixture module `tests/web/services/task_interaction_schema_shared.py`.

## Update since the last review round

Previously reviewed at `fdda83c5`; this pass covers `5be5eee5` ("fix(interactions): pair terminated_at with status and reject an empty responder identity"). That commit bundles fixes for all six findings that were open from the prior round — two new CHECK constraints, expanded `origin` parametrization, a completed `EXPECTED_STRING_LENGTHS` pin, and two test-hygiene/docstring corrections. I re-verified each fix against HEAD and found no regressions in the previously-accepted behavior.

## Prior findings — status

- [x] **FIXED** (gemini) Expiry CHECK lexicographic edge case on SQLite — addressed by documenting the per-backend clock behavior in the docstring (`b0bc2051`); the CHECK SQL was deliberately left unchanged. Still holds at HEAD.
- [x] **FIXED** (gemini) Copy-pasted docstring on `test_response_payload_none_lands_as_sql_null`.
- [x] **FIXED** (gemini) Unused `db_session` fixture on `test_all_timestamp_columns_are_timezone_aware`.
- [x] **FIXED** (Major) `terminated_at` had no CHECK pairing it with `status` — now `(status = 'terminated') = (terminated_at IS NOT NULL)` at `src/xagent/web/models/task_interaction.py:197-200`, with correct NULL semantics, tests in both suites, and the constraint added to the inventory fixture.
- [x] **FIXED** (Minor) `responder_identity` accepted the empty string — now `responder_identity <> ''` at `src/xagent/web/models/task_interaction.py:254-257`, correctly still permitting NULL on the nullable column, with a test and an inventory update.
- [x] **FIXED** (Minor) `origin` vocabulary tested only 2 of 6 values — now parametrized over all six in both suites, and the parametrization is non-tautological against the model's IN-list.
- [x] **FIXED** (Minor) `EXPECTED_STRING_LENGTHS` pinned only 7 of 12 string columns — all 12 are now present with correct widths, **plus** a new closure assertion (`set(EXPECTED_STRING_LENGTHS) == string_columns`) in both suites so a future string column cannot be silently left unpinned. Good improvement beyond what was asked.
- [x] **FIXED** (docs) `terminal_reason` docstring described writers in the present tense — reworded to intended/future tense and now factually accurate.

## Design verdict

**Acceptable with reservations.** The core shape matches issue #1074 well:

- One row per blocking interaction, with `active_slot` + `UNIQUE(task_id, active_slot)` exploiting NULL-distinct semantics to get "at most one active interaction per task" for free.
- `ON DELETE SET NULL` on the resume anchor combined with the active-row CHECK synthesizes RESTRICT-like behavior for active rows while letting terminal rows survive anchor cleanup.
- Run-scoped idempotency key.
- `String` + CHECK instead of a native `Enum`. Worth calling out explicitly: this reads as an inconsistency with `Task.status` / `TaskExecutionCommand.status` but is in fact **stronger** — on SQLite, SQLAlchemy 2.0.51 emits `Enum` as a plain `VARCHAR` with no CHECK at all, so the explicit CHECK gives real enforcement on both backends.

I verified the individual constraint claims in the class docstring (items 1, 2, 4, 5, 7 and 9) and they hold.

### Standing design note (not new — restating for the record)

The active-row CHECK requires a non-null resume anchor, and the FK uses `ON DELETE SET NULL`. That means `purge_task_rows` in `src/xagent/web/services/task_deletion.py`, which deletes trace events before the task row, will raise `IntegrityError` for any task that still has an active interaction request. This is unreachable today (a repo-wide search finds zero writers for this table), so it is not a regression in this PR, and it was already raised and accepted as a forward-looking requirement in the preceding review round. Repeating it only so it is not lost: **whichever PR introduces the writer must handle the deletion ordering.**

## New findings

### Major — must be fixed before this table gains any writer

**Major:** `src/xagent/web/models/task_interaction.py:284` — `request_payload = Column(JSON, nullable=False)` carries the comment "NOT NULL: no `none_as_null` needed", but that is empirically false. Inserting `request_payload=None` succeeds silently and stores the JSON text `'null'` (repro confirms `typeof(...) = 'text'` and `request_payload IS NULL` evaluating false). `nullable=False` only rejects SQL NULL; SQLAlchemy's `JSON` type without `none_as_null=True` serializes Python `None` into the JSON literal `null` *before* binding, so the value never reaches the NOT NULL check. This is precisely the inverse of the property that `response_payload` is careful about (it correctly sets `none_as_null=True`). A future writer doing something as ordinary as `request_payload=payload or None` would silently persist a row that no `IS NULL` predicate can find. There is no test covering it. **This must be fixed — not deferred — before any writer PR lands**, since the writer's own tests would otherwise pass against a broken guarantee. Suggested fix: add `none_as_null=True` to `request_payload` (matching `response_payload`) so `None` becomes a real SQL NULL and the NOT NULL constraint fires — or, if a Python-side guard is preferred, an explicit CHECK/`@validates`. Either way, add a test mirroring the existing `none_as_null` test for `response_payload`, and correct the comment.

**Major:** `tests/web/services/test_task_interaction_schema.py:850` vs `tests/web/services/test_task_interaction_schema_postgresql.py:745` — the two suites bind the *identical* aware datetime with a `+08:00` offset (i.e. 12:00 UTC) and then assert contradictory outcomes. The PostgreSQL test asserts the correct UTC-normalized readback (12:00). The SQLite test asserts that the stored value keeps the local wall-clock digits (20:00) — i.e. the offset is dropped rather than converted, which is genuine timezone corruption of the stored instant. The SQLite test's own docstring acknowledges this ("silently stored as local wall-clock time... the row is accepted with the wrong instant"), yet the test still pins the corrupted value via a `startswith` check. The practical effect is that the test would newly **fail** if the bug were ever fixed: it encodes the defect as a regression target rather than neutrally documenting current behavior. Issue #1074 explicitly asks this PR to "reuse the repository's existing datetime normalization boundary for UTC readback" — `normalize_datetime_for_db` / `normalize_datetime_from_db` in `src/xagent/web/utils/db_timezone.py`, both of which exist and are used elsewhere in the app but are not referenced anywhere in `task_interaction.py`. **This must be fixed before a writer lands**: either apply that normalization boundary now so both backends agree, or at minimum stop asserting the corrupted value as expected.

### Minor

**Minor:** `src/xagent/web/models/task_interaction.py:154` (the `origin` CHECK) vs `src/xagent/web/models/task.py:221` (`Task.source`) — the IN-list `internal, sdk, a2a, trigger, widget, shared_link` is a hand-maintained copy of `Task.source`'s informal vocabulary. `Task.source` is an unconstrained `String(20)` whose values are scattered as bare string literals across `a2a.py`, `public_chat_access.py`, `v1/tasks.py` and `triggers.py`, with no shared enum or constant. This PR itself had to hand-edit a comment in `task.py` to add `'a2a'`, which is exactly the drift signal. A future `Task.source` value added without a matching update here will pass review silently and only surface at INSERT time as a raw DB constraint error when an interaction is staged on such a task. Suggest a shared constant/Enum for the vocabulary, or failing that a cross-file consistency test that asserts the two lists agree.

**Minor:** `src/xagent/web/models/task_interaction.py:183-186` — the only CHECK touching `protocol_version` is `active_slot IS NULL OR protocol_version = 1`, which is vacuous for terminal rows. Confirmed by repro: `terminated` and `answered` rows with `protocol_version = -1` or `0` both insert successfully. Terminal rows are arguably exempt by design (there is an existing `test_active_row_requires_protocol_version_one`), but the exemption is not discussed, and letting arbitrary negative/nonsense versions land on terminal rows is mildly at odds with issue #1074's "fail closed on unknown kind/version". Either widen the CHECK to `protocol_version >= 1` unconditionally, or state the exemption in the docstring.

**Minor:** `tests/web/services/task_interaction_schema_shared.py:249` and `:295` — `EXPECTED_NULLABLE` and `TIMESTAMP_COLUMNS` lack the closure-against-reflected-metadata assertion that `EXPECTED_STRING_LENGTHS` just gained in this round. A newly added nullable or timestamp column would silently escape the "frozen shape" pin — the same gap that was just correctly closed for string lengths, left open for the other two inventories. Worth applying the same `set(...) == <reflected set>` assertion to both.

**Minor:** both test files (all ~80 tests) — every test inserts a complete row already in its final state; a grep finds no `UPDATE`, no `session.execute(update(...))`, and no `.query(...).update(...)` anywhere in the suite. But the production lifecycle is a sequence of transitions (`active` → `answered`, `active` → `terminated`), and the paired CHECKs added in this round — `active_slot`/`status`, `terminated_at`/`status`, the response/`responded_at` pairs — are precisely the constraints most likely to be violated *mid-transition* by a writer that updates columns in the wrong order or in separate statements. A handful of UPDATE-path tests would cover the case the constraints actually exist to protect.

**Minor:** `tests/web/services/test_task_interaction_schema.py:376-388` vs `:401-435` — asymmetric thoroughness between two structurally identical rejection tests: the `terminal_reason` "set on the wrong status" test only exercises `status='active'`, while the `terminated_at` pairing test correctly covers both `status='active'` and `status='answered'`. Add the `answered` case to the `terminal_reason` test for parity.

**Minor:** `src/xagent/web/models/task_interaction.py:20-124` (class docstring) — the docstring documents the design decisions at length but omits one caveat that *is* present in the PR description: "one active slot per root Task" only holds because one agent tree maps to exactly one Task row today, and `UNIQUE(task_id, active_slot)` would need revisiting if a root-task discriminator column is ever introduced. Since the docstring is where a future maintainer will look and the PR description is not, that one sentence belongs in the docstring.

**`test_task_interaction_schema_postgresql.py` duplicates 39 of its 41 tests near line-for-line from the SQLite suite** (differing mainly in trimmed docstrings). Correcting a characterization from an earlier review round here: this was described as "deliberate and justified, following an established repo-wide convention", and that does not survive contact with the actual precedent. The only other genuine SQLite/PostgreSQL test pair in the repo is `test_task_status_storage_postgresql.py`, which has **6** tests against its SQLite twin's **33** — and all 6 are genuinely backend-specific (native enum reflection, dialect case-sensitivity, real constraint behavior under Postgres). It deliberately does *not* re-run dialect-independent assertions. The convention this repo actually follows is "backend-specific tests only in the twin."

Two concrete sub-cases are unambiguous and worth fixing regardless of how the broader duplication is resolved:

- `test_table_is_registered_in_metadata` (`..._postgresql.py:753-756`, identical to `test_task_interaction_schema.py:861-868`) and `test_no_constraint_name_exceeds_the_postgres_identifier_limit` (`..._postgresql.py:821-833`, identical to `test_task_interaction_schema.py:969-984`) never touch a database at all — both purely inspect `Base.metadata` / `__table__` in-process, and both run unconditionally regardless of whether `XAGENT_TEST_POSTGRES_URL` is set. So they execute twice for zero added coverage on every local run. Deleting the duplicates outright (keeping one copy each, in the SQLite file or the shared module) is a pure win.

For the rest, consider parametrizing the dialect-independent CHECK/vocabulary tests over an engine fixture rather than copy-pasting the file. Restructuring to match the repo's own precedent would plausibly remove **700-800 lines** while *increasing* the signal from the Postgres suite, since what remains would be the assertions that only Postgres can make.

**Still outstanding from the prior round:** the suggestion to merge the four one-directional CHECKs (`answered_pairs_response`, `unanswered_has_no_response`, `answered_pairs_responded_at`, `unanswered_has_no_responded_at`) into two biconditional CHECKs — net -8 lines, previously verified logically equivalent — was accepted in principle but does not appear in `5be5eee5`, which only touches `terminated_at` and `responder_identity`. Note that the new `terminated_at` constraint already uses exactly this biconditional form, so applying it to the other four would also make the constraint block internally consistent. Author's discretion; not re-verified this round since it is unchanged.

## Test results

- The SQLite constraint suite passes at HEAD; the constraints added in `5be5eee5` (`terminated_at`/`status` pairing and the `responder_identity <> ''` guard) were confirmed by running the affected tests, along with the updated inventory fixture and the new `EXPECTED_STRING_LENGTHS` closure assertion.
- The PostgreSQL suite is gated on `XAGENT_TEST_POSTGRES_URL`, with the two exceptions noted in the simplification section that run unconditionally.
- The two Major findings above were each confirmed by direct repro rather than inspection: `request_payload=None` inserting as JSON text `'null'` (`typeof = 'text'`, `IS NULL` false), and terminal rows accepting `protocol_version` of `-1`/`0` for the related Minor.
- The `origin` parametrization was checked against the model's IN-list to confirm it is not tautological.

## Scope note and recommendation

A few candidate findings were investigated and dropped rather than reported: the claim that `resume_checkpoint_type`'s single-value IN-list blocks legacy v2 checkpoints (refuted — v2 is write-dead, read-only for historical rows), and the claim that `ix_task_interaction_requests_task_status` is fully redundant with the active-slot unique index (refuted as too strong — it serves "all answered/terminated rows for a task" queries the partial-by-NULL unique index cannot).

**Approving**, since this table is storage-only with zero consumers and the migration itself is deferred, so nothing in the current tree can reach either Major defect. That said, **the two Major findings are not optional cleanup — they must be fixed before any writer PR lands on this table**, since a writer's own test suite would otherwise pass against a silently broken `IS NULL` guarantee on `request_payload` and a timezone-corrupting `expires_at` path. Please treat closing them out as a prerequisite for the next PR in this series, not a someday-nice-to-have.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Confirmed on both backends, and the ORM side makes it worse than described: the row reads back as Python None while SQL sees the JSON text 'null', so the two views of "empty" disagree permanently.

    SQLite      typeof(request_payload)='text', value='null', IS NULL -> 0
    PostgreSQL  pg_typeof='json',               value='null', IS NULL -> false

Took the none_as_null=True route. Measured, it is the cheaper of the two: the compiled CREATE TABLE output is byte-identical with and without the flag on both dialects, so the deferred migration and the identifier-length budget are untouched. Both backends now reject a None payload at the column (NOT NULL constraint failed / NotNullViolation), while {} and populated payloads still insert.

Worth recording why the CHECK alternative is not just more expensive but awkward: `request_payload <> 'null'` is not portable — PostgreSQL's json type has no inequality operator and fails with "operator does not exist: json <> unknown"; it needs CAST(... AS TEXT) to work on both.

Landed with it: the comment corrected, the class docstring's "only column in the repository" sentence rewritten (the flag is confined to this table, carried by both of its JSON columns for different reasons), the metadata-wide test reframed to assert that confinement, and a rejection test mirroring the response_payload one in both suites.

tests/web/services/test_task_interaction_schema.py
AlexLiu190625: The contradiction is real and both tests were green at HEAD, so the test side is fixed. Splitting the recommendation, though, because the normalization half does not hold up against the tree.

On the boundary: normalize_datetime_from_db is the read half and has three consumers (schemas/user_api_key.py, api/v1/deps.py, services/api_keys.py). normalize_datetime_for_db has zero callers anywhere in the repository — only its own definition. There is also no TypeDecorator in the tree. So there is no existing write-side normalization pattern to reuse, and no column-level mechanism a model-only change could hook into; adopting one here would mean introducing the repository's first TypeDecorator in a PR with no writers. The column-shape half (DateTime(timezone=True) throughout, plus the caller's UTC obligation in the docstring) is what this layer can express; the read half lands with the first consumer.

On the test: agreed it should not read as blessing the wrong instant. Any characterization test fails when the behavior it characterizes changes — that is what pinning means — so what changed is what the failure tells you. The assertion keeps the measured digits but is reframed as a hazard pin with an explicit exit condition: if it ever fails, SQLite bind-side normalization has arrived, and the comment names the places to clean up. The PostgreSQL twin gains a cross-reference so the divergence reads as measured rather than accidental. Strict xfail was considered and passed on: there is none anywhere in this repo, and it would not remove the fail-on-fix property either.

Also added the missing clause to the docstring: it described what PostgreSQL does with a naive value but never stated that it converts an aware non-UTC value correctly — which is exactly the asymmetry the two tests measure.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Confirmed, and the writer list is one longer than stated — v1/workforces.py also passes source="sdk". An AST sweep of src/xagent for source= / task_source= string literals finds the task-source writes across five distinct callees plus two migration values() calls, and two unrelated hits where source means something else entirely.

Not taking the shared constant, for a semantic reason rather than a scope one: origin is a frozen copy taken at ask time, deliberately allowed to diverge from tasks.source afterwards. Its IN-list therefore has to be a superset of every value that was ever valid, whereas a Task.source constant expresses what is valid now. Those sets are equal today but are not the same thing — retiring a source value would narrow the constant, narrow this CHECK, and leave historical rows in a table that can no longer be recreated.

Took the consistency test instead, in the SQLite suite where it runs unconditionally. It cannot key off the comment in task.py — that comment was itself wrong until this PR (it was missing a2a), so it is not a source of truth. It sweeps the literals directly, with a reasoned two-entry exclusion set. A callee allowlist would be the tempting narrowing and has the wrong polarity: one more service wrapper and it silently stops catching anything.

Also added a docstring sentence: the IN-list is a frozen superset rather than a mirror, adding a Task.source value needs this CHECK plus a migration, and the staging primitive validates origin in Python before the INSERT so the normal error surface is a typed one, with the CHECK as backstop.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Reproduced, and the cell that decides the answer checks out: an unconditional protocol_version >= 1 floor still admits a future-version row on a terminal status. That matters because the point of scoping the existing CHECK to the slot is to let future-version rows exist historically while never occupying a v1 slot — a floor does not touch that. It only rejects values that are not version numbers. Measured on both backends: the -1/0 cells flip to rejected, terminated/answered with version 2 stay accepted, active/2 stays rejected by the existing slot-scoped CHECK.

Landed as ck_task_interaction_requests_protocol_version_floor, placed outside the slot-admission block with a note that it is a numeric domain floor — a different category from the slot-scoped pin, closer to the empty-string CHECKs, which exist for the same reason. Since the migration is still deferred this costs a CREATE TABLE line rather than an ALTER. The tests pin the nonsense values as rejected and, just as importantly, pin version-2-on-a-terminal-row as accepted, so the evolution guarantee is a positive control rather than something a later reader mistakes for a gap.

tests/web/services/task_interaction_schema_shared.py
AlexLiu190625: Confirmed — both inventories were complete today by coincidence, not enforcement, and a new column would have slipped past either.

Applied the same closure to both. For nullability it is a one-liner next to the existing column-set assertion. For timestamps the detection needed checking, since the reflected .timezone attribute is useless on SQLite; the type object is not: isinstance(column["type"], DateTime) returns exactly the five timestamp columns on both backends (SQLite reflects DATETIME, PostgreSQL TIMESTAMP, both subclasses). The PostgreSQL half closes against reflection in the test that already holds a session; the SQLite half closes against the declared table in the fixture-free test.

tests/web/services/test_task_interaction_schema.py
AlexLiu190625: Confirmed, and the missing cell rejects cleanly on both backends with ck_task_interaction_requests_terminal_pairs_status and nothing else — no other constraint competes for the error. Added to both suites, plus the docstring this test was missing and its twin has.

Kept the appended-assertion shape rather than parametrizing: the twin is sequential assert_rejected calls, and the point of the note is parity between the two, so making one structurally different would work against it.

src/xagent/web/models/task_interaction.py
AlexLiu190625: Agreed — the docstring outlives the description. Added, carrying the sharper half of the description text rather than just the caveat: if a root discriminator column is ever introduced it has to be a NOT NULL sentinel, not a nullable column, because the same NULL-distinct behavior this constraint relies on would otherwise void it entirely on the root side and accept a second active root row. That is the part a future maintainer would get wrong.

tests/web/services/test_task_interaction_schema_postgresql.py
AlexLiu190625: Confirmed. Running the PostgreSQL file without XAGENT_TEST_POSTGRES_URL gave 2 passed, 46 skipped, and those two were exactly these. Both bodies were identical to the SQLite copies; only the docstrings differed.

Deleted both PostgreSQL copies. Keeping the identifier-limit test in the always-running file is the stronger placement, not just the shorter one: it is pure model inspection, so it guards the 63-character budget on every run rather than only where a Postgres URL is configured.

tests/web/services/test_task_interaction_schema_postgresql.py
AlexLiu190625: Same for this one — duplicate deleted.

One count to correct on the broader point: the PostgreSQL file had 39 tests and the SQLite file 41, and all 39 shared a name with a SQLite test (the two SQLite extras are the foreign-keys pragma test and the metadata-wide none_as_null test). So it was 39 of 39, not 39 of 41.

On restructuring: keeping the matrix, with the reason rather than just a decline. The precedent does not transfer, because the two suites test different kinds of thing. The status-storage SQLite half tests ORM bind-layer behavior — dialect-independent by construction, so a Postgres rerun genuinely adds nothing. This suite tests DDL compilation output: "the CHECK behaves the same on both backends" is not guaranteed by construction, it is a per-constraint measurement, and it is the reason this table uses String + CHECK instead of Enum in the first place.

That is not hypothetical — it surfaced while working the request_payload finding above: the obvious CHECK to write was `request_payload <> 'null'`, which compiles and enforces on SQLite and fails on PostgreSQL at statement time with "operator does not exist: json <> unknown". Same expression, one backend enforcing, the other unable to run it. Dropping the Postgres side of the CHECK matrix is what would let that class of divergence through. Three smaller things point the same way: the rejection helper takes different branches per backend, so the Postgres half is the only place the constraint names — part of the contract with the deferred migration — are verified to surface at all; reflection is genuinely asymmetric; and the expiry round-trip test asserts opposite outcomes per backend, which is the Major above and would not have been visible from one side.

If this suite ever grows again, the right de-duplication is hoisting the dialect-independent test bodies into the shared module so both files call them — same coverage, real line savings. Noted as a follow-up candidate with that trigger; the table shape is frozen, so it is not growing now.

AlexLiu190625:
Two body-level items without inline threads, plus one decision worth surfacing.

**The UPDATE-path gap.** Confirmed — the suites contained no SQL UPDATE at all. Checked what the constraints actually do there rather than assuming: the mid-transition risk is real and caught per statement on both backends. Transitioning to answered without clearing the slot, clearing the slot without leaving active, and terminating without a timestamp are each rejected; the full one-statement answered transition is accepted. All four are now tests in both suites, the accepted transition included — without it the three rejections could pass against a constraint no row can satisfy. One measured wrinkle shaped the helper: when a single UPDATE violates more than one CHECK, the backends disagree on which name surfaces (the same answered-to-terminated statement reports response_pairs_status on SQLite and responded_at_pairs_status on PostgreSQL), so each rejection test isolates exactly one violation and the helper's docstring says why.

**The four one-directional pairing CHECKs.** Merged into two biconditionals this round, reversing the earlier decline. What changed the call was not taste: the paired-CHECK block's own comment states the criterion for writing a pair both ways — neither column unilaterally cleared by ON DELETE SET NULL — and response_payload and responded_at both satisfy it, while the terminated_at pairing added last round already used the biconditional form. A reader applying the comment's stated rule would have predicted the opposite of what the code did. The constraint names in the deferred migration's inventory move accordingly.

The updated totals after this round: 23 CHECK constraints, 105 tests across the two suites.

PR #1214, opened 2026-08-09
fix(auth): preserve infrastructure failures at authentication boundaries

### LINKED ISSUE #930: fix(auth): preserve infrastructure failures instead of returning 401
## Summary

The shared authentication boundary currently converts database and connection-pool failures into authentication outcomes across required HTTP authentication, optional authentication, token helpers, and WebSocket transports.

With a valid access JWT, a user-query failure such as sqlalchemy.exc.TimeoutError currently becomes one of the following:

- HTTP 401 Invalid authentication credentials;
- an optional or token-helper result of None;
- a WebSocket authentication rejection such as code 4001 or 1008;
- a legacy fallback identity in a downstream consumer.

This changes the meaning of a valid request during database pressure, hides the availability failure, and can trigger credential refresh, login redirects, policy-violation closes, or execution under an unrelated fallback identity.

## Verified baseline

The behavior remains present on Core main d97f440a94e5482a26ae607b89439d965d694093.

The auth dependency file has the same blob (84044844e150596c3ba8cc39c5c7459a2eb1aa04) at:

- the production revision recorded by this issue: 3a1b69f5640f0fbd99f09c9b68f48d79e32c827b;
- the original issue baseline: 217d915f6052b9b14b41b8462229321386654382;
- current Core main: d97f440a94e5482a26ae607b89439d965d694093.

Related pool-isolation and request-fan-out work reduced known pressure sources but did not change this error-classification behavior.

### Required HTTP authentication

[get_current_user()](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/auth_dependencies.py#L21-L101) places JWT decoding, claims validation, the User query, and intentional HTTPException values inside one try block. Its broad except Exception maps every remaining failure to HTTP 401.

Consequences:

- a database checkout/query failure is reported as invalid credentials;
- intentional details such as Invalid token type and User not found are overwritten by the generic 401 detail.

### Optional HTTP authentication

[get_current_user_optional()](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/auth_dependencies.py#L104-L124) does not directly catch database exceptions. It calls get_current_user(), then catches the HTTPException that the required helper produced and returns None. The infrastructure failure is therefore indirectly treated as an absent optional user.

No active Depends(get_current_user_optional) consumer was found at this baseline, but the shared contract is still incorrect and must not remain as a future bypass.

### Token and WebSocket authentication

[get_user_from_token() and get_user_from_websocket_token()](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/auth_dependencies.py#L127-L200) return None for both invalid tokens and unexpected database failures.

The main [get_authenticated_user() WebSocket boundary](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/api/websocket.py#L4895-L4932) adds another broad except Exception that returns None. Fixing only the token helper would therefore still classify an infrastructure failure as unauthenticated.

The same token helper is consumed by:

- [the progress WebSocket endpoint](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/api/progress_ws.py#L34-L53), where None becomes a 1008 policy-violation close;
- [the legacy tool configuration request resolver](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/tools/config.py#L1298-L1328), whose outer broad catch returns fallback user ID 1;
- [the legacy user-context helper](https://github.com/xorbitsai/xagent/blob/d97f440a94e5482a26ae607b89439d965d694093/src/xagent/web/user_context.py#L56-L79), whose outer broad catch returns None. No production caller was found for this helper at the fixed baseline.

These downstream catches must be included in the closed-set audit. A propagated database failure must not be reclassified at the next layer.

## Production evidence

A downstream PostgreSQL deployment using revision 3a1b69f recorded this failure mode during two request bursts:

| Time (UTC) | Observed fact |
|---|---|
| 2026-07-19 15:38:17 | Seven authentication dependency errors began exactly 10 seconds after the request burst. Each logged Authentication error: QueuePool limit of size 10 overflow 20 reached, connection timed out, timeout 10.00. |
| 2026-07-20 03:58–04:01 | The same request path produced 53 pool-timeout log entries in 15 minutes; the corresponding access-log window included 13 HTTP 401 responses while backend health checks were timing out. |

The runtime pool acquisition timeout was 10 seconds. The timestamps matched the configured pool wait, not a JWT expiration boundary.

This issue does not assert that authentication caused the pool exhaustion. It owns the error-semantic amplifier after the database was already unavailable.

## Required invariant

Authentication outcomes must describe credential and identity validity only.

- Missing, invalid, expired, wrong-type, or otherwise unacceptable credentials retain their transport-specific authentication behavior.
- Intentional HTTPException status, detail, and authentication headers are preserved.
- Database, connection-pool, and other infrastructure failures are never converted to HTTP 401, None, a WebSocket authentication/policy rejection, or a fallback identity.
- HTTP infrastructure failures reach the centralized server/infrastructure error path.
- WebSocket transport owners produce a deterministic internal-failure outcome rather than an authentication-required or policy-violation outcome.
- Logs retain the infrastructure exception with request/transport context; client-visible responses and close reasons remain sanitized.

## Required design

- Separate token decoding/claims classification from persistence lookup at the shared authentication owner.
- Keep expected credential failures represented explicitly; do not infer failure classes from SQLAlchemy message strings.
- Let unexpected persistence/infrastructure exceptions propagate out of the auth classifier.
- Audit every direct consumer and outer catch of get_current_user(), get_current_user_optional(), get_user_from_token(), and get_user_from_websocket_token().
- Fix or remove every consumer that would re-collapse a propagated infrastructure failure. An unused legacy path is not an implicit waiver.
- Keep transport-specific mapping at the HTTP or WebSocket boundary rather than duplicating database exception policy at individual routes.
- Preserve the current session ownership model; authentication helpers must not commit or otherwise mutate the shared request session.

## Acceptance criteria

- [ ] A valid JWT plus a User-query sqlalchemy.exc.TimeoutError does not return HTTP 401 from required authentication.
- [ ] The optional-auth path does not return None for the same database failure.
- [ ] get_user_from_token() and get_user_from_websocket_token() do not return None for the same database failure.
- [ ] Chat, builder-chat, and build-preview WebSocket authentication do not turn the same failure into code 4001 Authentication required.
- [ ] The progress WebSocket does not turn the same failure into code 1008 Policy Violation; it produces a sanitized internal-failure outcome.
- [ ] Legacy tool configuration does not turn the same failure into fallback user ID 1.
- [ ] The legacy user-context path either propagates the infrastructure failure correctly or is removed if it has no supported consumer.
- [ ] Invalid, expired, wrong-type, malformed, missing-user, and missing-credential behavior remains explicitly covered for HTTP and WebSocket surfaces.
- [ ] Existing intentional HTTPException status, detail, and headers are preserved.
- [ ] Tests exercise the shared classifier and every production transport/consumer listed above, including the outer WebSocket exception boundary.
- [ ] A structural inventory or equivalent guard fails when a new production auth consumer introduces a broad fallback that can collapse infrastructure failures.
- [ ] Logs preserve exception type and request/transport context without exposing database messages or credentials to clients.

## Non-goals

- Diagnosing or fixing the workload that exhausted the database pool.
- Changing database pool size, overflow, timeout, or deployment topology.
- Changing JWT claims, signing configuration, token lifetime, or refresh policy.
- Redesigning unrelated WebSocket lifecycle behavior after authentication succeeds.
- Adding SQLAlchemy message matching inside authentication code.
- Expanding the change into the separate /v1 authentication stack unless the implementation audit finds it consumes the same affected helper.

AlexLiu190625:
## Summary

- distinguish expected credential rejection from database and runtime failures at the shared authentication owners
- preserve infrastructure exceptions across required, optional, token, main WebSocket, and public Widget/Share authentication paths
- project authentication infrastructure failures through the existing sanitized HTTP 500 boundary, a pre-handshake WebSocket HTTP 503 denial, or an accepted-WebSocket 1011 close while retaining established credential and access-denial responses
- prevent identity-free tool configuration from inheriting admin privilege or emitting an MCP isolation identity

## Root cause

Broad exception handling collapsed database checkout, query, deployment, and runtime failures into invalid credentials, anonymous users, or WebSocket authentication rejection. Optional HTTP authentication also used the required bearer scheme, so requests without bearer credentials were rejected before the optional dependency ran. Tool configuration could derive privilege or serialize an isolation key without an explicit user identity.

## Behavior

- expired, invalid, wrong-type, malformed, missing-user, and missing credentials keep their established HTTP details and headers
- required, optional, and direct token helpers propagate infrastructure failures instead of converting them to 401 or `None`
- optional authentication treats missing, empty Bearer, and non-Bearer headers as anonymous; required authentication retains framework rejection
- main authenticated WebSockets use a sanitized pre-handshake 503 or extensionless 1011 response for infrastructure failures
- public Widget/Share WebSockets preserve credential and access denials as 4003, while initial and per-message infrastructure failures close accepted sockets with 1011
- all signed public IDs are exact integers before binding lookup and are range-checked for SQLite/PostgreSQL before query; workforce-first selection is unchanged
- identity-free `WebToolConfig` instances are non-admin and return no MCP configs before cache, database, or resolver work
- client responses and structured log context exclude credentials, query strings, database messages, and tracebacks; server logs retain the original traceback for diagnosis

## Verification

- 427 focused and adjacent authentication, WebSocket, public-chat, tool-identity, MCP, and auth-API tests passed
- mutation checks covered rejection taxonomy, optional FastAPI dependency behavior, public resolver exception identity, every query-bound public ID, initial/per-message 1011 projection, `extensions=None`, and no-identity tool configuration
- changed-file pre-commit hooks, Ruff, formatting, mypy, isort, codespell, and diff checks passed
- SQLite runtime and PostgreSQL dialect/driver bindability cases are covered; a live PostgreSQL service was not available locally

## Scope

The access-token classifier, public credential owners, WebSocket transport projection, and identity-isolation boundaries form one failure-semantics change and are intended to land atomically. No frontend, schema, migration, retry, or public API shape changes are included.

Closes #930

### DIFF
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -8,6 +8,7 @@ import secrets
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
+from enum import Enum, auto
 from typing import Any, Callable, Protocol, TypeGuard, TypeVar
 
 from fastapi import Depends, HTTPException, Query, UploadFile, WebSocket
@@ -155,6 +156,28 @@ def _is_strict_int(value: object) -> TypeGuard[int]:
 class _PublicTokenRejected(Exception):
     """An expected widget/share credential rejection."""
 
+    def __init__(self, reason: "_PublicTokenRejectionReason") -> None:
+        super().__init__()
+        self.reason = reason
+
+
+class _PublicTokenRejectionReason(Enum):
+    INVALID_TOKEN = auto()
+    INVALID_CLAIMS = auto()
+    INVALID_ENTITY = auto()
+
+
+def _project_public_token_failure(
+    exc: Exception, *, invalid_detail: str
+) -> HTTPException | None:
+    """Project expected public-token failures while preserving operations."""
+    if isinstance(exc, HTTPException):
+        return exc
+    if isinstance(exc, _PublicTokenRejected):
+        logger.info("Public chat credential rejected reason=%s", exc.reason.name)
+        return HTTPException(status_code=401, detail=invalid_detail)
+    return None
+
 
 def _decode_public_token(token: str) -> dict[str, Any]:
     """Verify one public token without folding decoder defects into credentials."""
@@ -162,21 +185,23 @@ def _decode_public_token(token: str) -> dict[str, Any]:
         token = token[7:]
     try:
         payload = jwt.decode(token, JWT_SECRET_KEY, algorithms=[JWT_ALGORITHM])
-    except JWTError as exc:
-        raise _PublicTokenRejected from exc
+    except JWTError:
+        raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_TOKEN) from None
     except (TypeError, OverflowError) as exc:
         if has_matching_temporal_claim_conversion_failure(token, exc):
-            raise _PublicTokenRejected from exc
+            raise _PublicTokenRejected(
+                _PublicTokenRejectionReason.INVALID_CLAIMS
+            ) from None
         raise
     if type(payload) is not dict:
-        raise _PublicTokenRejected
+        raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_TOKEN)
     return payload
 
 
 def _require_bindable_public_id(value: object, dialect_name: str) -> int:
     """Return an exact query id accepted by the supported database dialect."""
     if not is_exact_integer_bindable(value, dialect_name):
-        raise _PublicTokenRejected
+        raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
     return value
 
 
@@ -328,7 +353,7 @@ def get_public_chat_user(
     try:
         payload = _decode_public_token(token)
         if payload.get("type") != "widget":
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_TOKEN)
 
         user_id = payload.get("user_id")
         channel_id = payload.get("channel_id")
@@ -341,7 +366,7 @@ def get_public_chat_user(
             raise HTTPException(status_code=403, detail="Access denied")
 
         if auth_mode != "widget":
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_TOKEN)
 
         # Defense in depth (#992): the same guard get_share_chat_user applies.
         # Without it the two widget branches below disagreed about a string
@@ -349,10 +374,15 @@ def get_public_chat_user(
         # comparison, while the workforce branch coerced it and admitted the
         # request. Establishing the type here is what lets both branches pass
         # ``user_id`` straight to their ``ensure_widget_*_available`` helper.
-        if not _is_strict_int(user_id) or not user_id or not guest_id:
-            raise _PublicTokenRejected
+        if (
+            not _is_strict_int(user_id)
+            or not user_id
+            or not isinstance(guest_id, str)
+            or not guest_id
+        ):
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
         if channel_id is not None and not _is_strict_int(channel_id):
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
 
         if _is_strict_int(widget_workforce_id):
             selected_agent_id = None
@@ -361,7 +391,7 @@ def get_public_chat_user(
             selected_agent_id = widget_agent_id
             selected_workforce_id = None
         else:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_ENTITY)
 
         dialect_name = str(db.get_bind().dialect.name)
         user_id = _require_bindable_public_id(user_id, dialect_name)
@@ -378,7 +408,7 @@ def get_public_chat_user(
 
         user = db.query(User).filter(User.id == user_id).first()
         if not user:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
 
         if selected_workforce_id is not None:
             # Workforce widget: re-validate the deployment on every use so
@@ -406,7 +436,7 @@ def get_public_chat_user(
         # calls back through here, so live sessions drop on the next inbound
         # message too.
         if selected_agent_id is None:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_ENTITY)
         ensure_widget_agent_available(
             db,
             selected_agent_id,
@@ -424,13 +454,11 @@ def get_public_chat_user(
             widget_agent_id=selected_agent_id,
         )
     except Exception as exc:
-        logger.error("Public chat token validation error: %s", exc)
-        if isinstance(exc, HTTPException):
-            raise exc
-        if isinstance(exc, _PublicTokenRejected):
-            raise HTTPException(
-                status_code=401, detail="Invalid widget token"
-            ) from None
+        projected = _project_public_token_failure(
+            exc, invalid_detail="Invalid widget token"
+        )
+        if projected is not None:
+            raise projected from None
         raise
 
 
@@ -439,7 +467,7 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
     try:
         payload = _decode_public_token(token)
         if payload.get("type") != "widget":
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_TOKEN)
 
         user_id = payload.get("user_id")
         auth_mode = payload.get("auth_mode")
@@ -449,18 +477,18 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
         guest_id = payload.get("guest_id")
 
         if auth_mode != "share":
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_TOKEN)
         if not _is_strict_int(user_id):
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
         if not isinstance(share_token, str) or not share_token:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
         # Fail closed on tokens minted before per-guest isolation (#973): they
         # carry no guest_id, so they cannot be scoped to a single guest and are
         # rejected rather than silently granted the old no-isolation behavior.
         # A whitespace-only id is treated as absent (it could never match a
         # server-minted token_urlsafe value and must not pass as a real guest).
         if not isinstance(guest_id, str) or not guest_id.strip():
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
 
         if _is_strict_int(share_workforce_id):
             selected_agent_id = None
@@ -469,7 +497,7 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
             selected_agent_id = share_agent_id
             selected_workforce_id = None
         else:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_ENTITY)
 
         dialect_name = str(db.get_bind().dialect.name)
         user_id = _require_bindable_public_id(user_id, dialect_name)
@@ -484,7 +512,7 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
 
         user = db.query(User).filter(User.id == user_id).first()
         if not user:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_CLAIMS)
 
         if selected_workforce_id is not None:
             workforce = ensure_share_workforce_available(
@@ -501,7 +529,7 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
             )
 
         if selected_agent_id is None:
-            raise _PublicTokenRejected
+            raise _PublicTokenRejected(_PublicTokenRejectionReason.INVALID_ENTITY)
         agent = ensure_share_agent_available(
             db,
             selected_agent_id,
@@ -512,11 +540,11 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
             user=user, share_token=share_token, guest_id=guest_id, agent=agent
         )
     except Exception as exc:
-        logger.error("Share chat token validation error: %s", exc)
-        if isinstance(exc, HTTPException):
-            raise exc
-        if isinstance(exc, _PublicTokenRejected):
-            raise HTTPException(status_code=401, detail="Invalid share token") from None
+        projected = _project_public_token_failure(
+            exc, invalid_detail="Invalid share token"
+        )
+        if projected is not None:
+            raise projected from None
         raise
 
 
@@ -1391,9 +1419,7 @@ async def public_chat_websocket_endpoint(
         await websocket.close(code=4003, reason=_ws_close_reason(exc.detail))
         return
     except Exception as exc:
-        await send_websocket_authentication_infrastructure_failure(
-            websocket, exc, already_accepted=True
-        )
+        await send_websocket_authentication_infrastructure_failure(websocket, exc)
         return
 
     # Already accepted above; register the live socket without re-accepting.
@@ -1417,7 +1443,7 @@ async def public_chat_websocket_endpoint(
                 return
             except Exception as exc:
                 await send_websocket_authentication_infrastructure_failure(
-                    websocket, exc, already_accepted=True
+                    websocket, exc
                 )
                 return
 
@@ -1500,9 +1526,7 @@ async def share_chat_websocket_endpoint(
         await websocket.close(code=4003, reason=_ws_close_reason(exc.detail))
         return
     except Exception as exc:
-        await send_websocket_authentication_infrastructure_failure(
-            websocket, exc, already_accepted=True
-        )
+        await send_websocket_authentication_infrastructure_failure(websocket, exc)
         return
 
     # Already accepted above; register the live socket without re-accepting.
@@ -1525,7 +1549,7 @@ async def share_chat_websocket_endpoint(
                 return
             except Exception as exc:
                 await send_websocket_authentication_infrastructure_failure(
-                    websocket, exc, already_accepted=True
+                    websocket, exc
                 )
                 return
 
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -12,11 +12,15 @@ from typing import Any, Callable, Protocol, TypeGuard, TypeVar
 
 from fastapi import Depends, HTTPException, Query, UploadFile, WebSocket
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
-from jose import jwt
+from jose import JWTError, jwt
 from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
 from ..auth_config import JWT_ALGORITHM, JWT_SECRET_KEY
+from ..jwt_validation import (
+    has_matching_temporal_claim_conversion_failure,
+    is_exact_integer_bindable,
+)
 from ..models.agent import Agent, AgentStatus, is_workforce_generated_manager_agent
 from ..models.database import get_db, get_session_local, release_db_connection_if_clean
 from ..models.deployment import DeploymentOwnerType
@@ -50,6 +54,7 @@ from .websocket import (
     manager,
     send_message_delivery,
 )
+from .websocket_auth import send_websocket_authentication_infrastructure_failure
 
 logger = logging.getLogger(__name__)
 db_session_context = contextmanager(get_db)
@@ -145,6 +150,34 @@ def _is_strict_int(value: object) -> TypeGuard[int]:
     claim today; this keeps every id claim failing closed regardless.
     """
     return isinstance(value, int) and not isinstance(value, bool)
+
+
+class _PublicTokenRejected(Exception):
+    """An expected widget/share credential rejection."""
+
+
+def _decode_public_token(token: str) -> dict[str, Any]:
+    """Verify one public token without folding decoder defects into credentials."""
+    if token.startswith("Bearer "):
+        token = token[7:]
+    try:
+        payload = jwt.decode(token, JWT_SECRET_KEY, algorithms=[JWT_ALGORITHM])
+    except JWTError as exc:
+        raise _PublicTokenRejected from exc
+    except (TypeError, OverflowError) as exc:
+        if has_matching_temporal_claim_conversion_failure(token, exc):
+            raise _PublicTokenRejected from exc
+        raise
+    if type(payload) is not dict:
+        raise _PublicTokenRejected
+    return payload
+
+
+def _require_bindable_public_id(value: object, dialect_name: str) -> int:
+    """Return an exact query id accepted by the supported database dialect."""
+    if not is_exact_integer_bindable(value, dialect_name):
+        raise _PublicTokenRejected
+    return value
 
 
 def create_public_chat_access_token(data: dict[str, Any]) -> str:
@@ -293,12 +326,9 @@ def get_public_chat_user(
 ) -> PublicChatAccessContext:
     """Get public chat access context from a widget/share token."""
     try:
-        if token.startswith("Bearer "):
-            token = token[7:]
-
-        payload = jwt.decode(token, JWT_SECRET_KEY, algorithms=[JWT_ALGORITHM])
+        payload = _decode_public_token(token)
         if payload.get("type") != "widget":
-            raise ValueError("Invalid token type")
+            raise _PublicTokenRejected
 
         user_id = payload.get("user_id")
         channel_id = payload.get("channel_id")
@@ -311,7 +341,7 @@ def get_public_chat_user(
             raise HTTPException(status_code=403, detail="Access denied")
 
         if auth_mode != "widget":
-            raise ValueError("Invalid token payload")
+            raise _PublicTokenRejected
 
         # Defense in depth (#992): the same guard get_share_chat_user applies.
         # Without it the two widget branches below disagreed about a string
@@ -320,19 +350,43 @@ def get_public_chat_user(
         # request. Establishing the type here is what lets both branches pass
         # ``user_id`` straight to their ``ensure_widget_*_available`` helper.
         if not _is_strict_int(user_id) or not user_id or not guest_id:
-            raise ValueError("Invalid token payload")
+            raise _PublicTokenRejected
+        if channel_id is not None and not _is_strict_int(channel_id):
+            raise _PublicTokenRejected
+
+        if _is_strict_int(widget_workforce_id):
+            selected_agent_id = None
+            selected_workforce_id = widget_workforce_id
+        elif _is_strict_int(widget_agent_id):
+            selected_agent_id = widget_agent_id
+            selected_workforce_id = None
+        else:
+            raise _PublicTokenRejected
+
+        dialect_name = str(db.get_bind().dialect.name)
+        user_id = _require_bindable_public_id(user_id, dialect_name)
+        if channel_id is not None:
+            channel_id = _require_bindable_public_id(channel_id, dialect_name)
+        if selected_workforce_id is not None:
+            selected_workforce_id = _require_bindable_public_id(
+                selected_workforce_id, dialect_name
+            )
+        else:
+            selected_agent_id = _require_bindable_public_id(
+                selected_agent_id, dialect_name
+            )
 
         user = db.query(User).filter(User.id == user_id).first()
         if not user:
-            raise ValueError("User not found")
+            raise _PublicTokenRejected
 
-        if _is_strict_int(widget_workforce_id):
+        if selected_workforce_id is not None:
             # Workforce widget: re-validate the deployment on every use so
             # disabling the widget or rotating its key cuts off live guests,
             # mirroring the workforce share path.
             ensure_widget_workforce_available(
                 db,
-                widget_workforce_id,
+                selected_workforce_id,
                 user_id,
                 expected_widget_key=widget_key
                 if isinstance(widget_key, str) and widget_key
@@ -343,7 +397,7 @@ def get_public_chat_user(
                 channel_id=channel_id,
                 guest_id=guest_id,
                 auth_mode=auth_mode,
-                widget_workforce_id=widget_workforce_id,
+                widget_workforce_id=selected_workforce_id,
             )
 
         # Agent widget: re-validate against the live agent every request so
@@ -351,11 +405,11 @@ def get_public_chat_user(
         # tokens (mirrors the share path). The per-message WS revalidation
         # calls back through here, so live sessions drop on the next inbound
         # message too.
-        if not _is_strict_int(widget_agent_id):
-            raise ValueError("Invalid widget token payload")
+        if selected_agent_id is None:
+            raise _PublicTokenRejected
         ensure_widget_agent_available(
             db,
-            widget_agent_id,
+            selected_agent_id,
             user_id,
             expected_widget_key=widget_key
             if isinstance(widget_key, str) and widget_key
@@ -367,24 +421,25 @@ def get_public_chat_user(
             channel_id=channel_id,
             guest_id=guest_id,
             auth_mode=auth_mode,
-            widget_agent_id=widget_agent_id,
+            widget_agent_id=selected_agent_id,
         )
     except Exception as exc:
         logger.error("Public chat token validation error: %s", exc)
         if isinstance(exc, HTTPException):
             raise exc
-        raise HTTPException(status_code=401, detail="Invalid widget token")
+        if isinstance(exc, _PublicTokenRejected):
+            raise HTTPException(
+                status_code=401, detail="Invalid widget token"
+            ) from None
+        raise
 
 
 def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
     """Get share chat access context from a share token."""
     try:
-        if token.startswith("Bearer "):
-            token = token[7:]
-
-        payload = jwt.decode(token, JWT_SECRET_KEY, algorithms=[JWT_ALGORITHM])
+        payload = _decode_public_token(token)
         if payload.get("type") != "widget":
-            raise ValueError("Invalid token type")
+            raise _PublicTokenRejected
 
         user_id = payload.get("user_id")
         auth_mode = payload.get("auth_mode")
@@ -394,27 +449,47 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
         guest_id = payload.get("guest_id")
 
         if auth_mode != "share":
-            raise ValueError("Invalid token payload")
+            raise _PublicTokenRejected
         if not _is_strict_int(user_id):
-            raise ValueError("Invalid token payload")
+            raise _PublicTokenRejected
         if not isinstance(share_token, str) or not share_token:
-            raise ValueError("Invalid share token payload")
+            raise _PublicTokenRejected
         # Fail closed on tokens minted before per-guest isolation (#973): they
         # carry no guest_id, so they cannot be scoped to a single guest and are
         # rejected rather than silently granted the old no-isolation behavior.
         # A whitespace-only id is treated as absent (it could never match a
         # server-minted token_urlsafe value and must not pass as a real guest).
         if not isinstance(guest_id, str) or not guest_id.strip():
-            raise ValueError("Invalid share token payload")
+            raise _PublicTokenRejected
+
+        if _is_strict_int(share_workforce_id):
+            selected_agent_id = None
+            selected_workforce_id = share_workforce_id
+        elif _is_strict_int(share_agent_id):
+            selected_agent_id = share_agent_id
+            selected_workforce_id = None
+        else:
+            raise _PublicTokenRejected
+
+        dialect_name = str(db.get_bind().dialect.name)
+        user_id = _require_bindable_public_id(user_id, dialect_name)
+        if selected_workforce_id is not None:
+            selected_workforce_id = _require_bindable_public_id(
+                selected_workforce_id, dialect_name
+            )
+        else:
+            selected_agent_id = _require_bindable_public_id(
+                selected_agent_id, dialect_name
+            )
 
         user = db.query(User).filter(User.id == user_id).first()
         if not user:
-            raise ValueError("User not found")
+            raise _PublicTokenRejected
 
-        if _is_strict_int(share_workforce_id):
+        if selected_workforce_id is not None:
             workforce = ensure_share_workforce_available(
                 db,
-                share_workforce_id,
+                selected_workforce_id,
                 user_id,
                 expected_share_token=share_token,
             )
@@ -425,12 +500,11 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
                 workforce=workforce,
             )
 
-        if not _is_strict_int(share_agent_id):
-            raise ValueError("Invalid share token payload")
-
+        if selected_agent_id is None:
+            raise _PublicTokenRejected
         agent = ensure_share_agent_available(
             db,
-            share_agent_id,
+            selected_agent_id,
             user_id,
             expected_share_token=share_token,
         )
@@ -441,7 +515,9 @@ def get_share_chat_user(token: str, db: Session) -> ShareChatAccessContext:
         logger.error("Share chat token validation error: %s", exc)
         if isinstance(exc, HTTPException):
             raise exc
-        raise HTTPException(status_code=401, detail="Invalid share token")
+        if isinstance(exc, _PublicTokenRejected):
+            raise HTTPException(status_code=401, detail="Invalid share token") from None
+        raise
 
 
 security = HTTPBearer()
@@ -1314,8 +1390,10 @@ async def public_chat_websocket_endpoint(
     except HTTPException as exc:
         await websocket.close(code=4003, reason=_ws_close_reason(exc.detail))
         return
-    except Exception:
-        await websocket.close(code=4001, reason="Authentication required")
+    except Exception as exc:
+        await send_websocket_authentication_infrastructure_failure(
+            websocket, exc, already_accepted=True
+        )
         return
 
     # Already accepted above; register the live socket without re-accepting.
@@ -1336,6 +1414,11 @@ async def public_chat_websocket_endpoint(
                 )
             except HTTPException as exc:
                 await websocket.close(code=4003, reason=_ws_close_reason(exc.detail))
+                return
+            except Exception as exc:
+                await send_websocket_authentication_infrastructure_failure(
+                    websocket, exc, already_accepted=True
+                )
                 return
 
             message_data["user_id"] = current_principal.id
@@ -1416,8 +1499,10 @@ async def share_chat_websocket_endpoint(
     except HTTPException as exc:
         await websocket.close(code=4003, reason=_ws_close_reason(exc.detail))
         return
-    except Exception:
-        await websocket.close(code=4001, reason="Authentication required")
+    except Exception as exc:
+        await send_websocket_authentication_infrastructure_failure(
+            websocket, exc, already_accepted=True
+        )
         return
 
     # Already accepted above; register the live socket without re-accepting.
@@ -1437,6 +1522,11 @@ async def share_chat_websocket_endpoint(
                 )
             except HTTPException as exc:
                 await websocket.close(code=4003, reason=_ws_close_reason(exc.detail))
+                return
+            except Exception as exc:
+                await send_websocket_authentication_infrastructure_failure(
+                    websocket, exc, already_accepted=True
+                )
                 return
 
             message_data["user_id"] = current_principal.id
--- a/src/xagent/web/api/websocket_auth.py
+++ b/src/xagent/web/api/websocket_auth.py
@@ -29,6 +29,44 @@ class _WebSocketAuthenticationTerminated(Exception):
     """Authentication already sent a terminal transport response."""
 
 
+async def send_websocket_authentication_infrastructure_failure(
+    websocket: WebSocket,
+    original_error: Exception,
+    *,
+    already_accepted: bool,
+) -> None:
+    """Send the shared sanitized response for an authentication outage."""
+    route_template = getattr(websocket.scope.get("route"), "path", "<unresolved>")
+    logger.error(
+        "WebSocket authentication infrastructure failure transport=websocket route=%s",
+        route_template,
+        exc_info=(
+            type(original_error),
+            original_error,
+            original_error.__traceback__,
+        ),
+    )
+    try:
+        if already_accepted:
+            await websocket.close(code=1011, reason="Internal server error")
+        elif "websocket.http.response" in (websocket.scope.get("extensions") or {}):
+            await websocket.send_denial_response(
+                JSONResponse(
+                    status_code=503,
+                    content={"detail": "Service temporarily unavailable"},
+                )
+            )
+        else:
+            await websocket.accept()
+            await websocket.close(code=1011, reason="Internal server error")
+    except Exception:
+        logger.exception(
+            "WebSocket authentication terminal response failure "
+            "transport=websocket route=%s",
+            route_template,
+        )
+
+
 def _load_websocket_principal_sync(token: str) -> WebSocketPrincipal | None:
     """Authenticate one token inside a worker-owned short Session."""
 
@@ -57,27 +95,9 @@ async def get_authenticated_user(
             lambda: _load_websocket_principal_sync(token)
         )
     except Exception as exc:
-        route_template = getattr(websocket.scope.get("route"), "path", "<unresolved>")
-        logger.exception(
-            "WebSocket authentication infrastructure failure "
-            "transport=websocket route=%s",
-            route_template,
+        await send_websocket_authentication_infrastructure_failure(
+            websocket,
+            exc,
+            already_accepted=False,
         )
-        try:
-            if "websocket.http.response" in websocket.scope.get("extensions", {}):
-                await websocket.send_denial_response(
-                    JSONResponse(
-                        status_code=503,
-                        content={"detail": "Service temporarily unavailable"},
-                    )
-                )
-            else:
-                await websocket.accept()
-                await websocket.close(code=1011, reason="Internal server error")
-        except Exception:
-            logger.exception(
-                "WebSocket authentication terminal response failure "
-                "transport=websocket route=%s",
-                route_template,
-            )
         raise _WebSocketAuthenticationTerminated() from exc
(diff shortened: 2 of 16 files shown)


src/xagent/web/api/websocket_auth.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+    except Exception as exc:
+        route_template = getattr(websocket.scope.get("route"), "path", "<unresolved>")
+        logger.exception(
+            "WebSocket authentication infrastructure failure "
+            "transport=websocket route=%s",
+            route_template,
+        )
+        try:
+            if "websocket.http.response" in websocket.scope.get("extensions", {}):
```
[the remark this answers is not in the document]
AlexLiu190625: The extension membership check is already inside the inner try at websocket_auth.py:66-77, so a malformed extensions=None value is caught before the private terminal marker is raised. A direct probe preserved the original infrastructure exception as the marker cause and did not bypass endpoint handling or hang. The ASGI scope contract requires extensions to be a mapping when present, and the pinned Uvicorn implementation supplies that mapping, so the typed production path remains unchanged.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
+            if type(reproduced_error) is type(original_error):
+                return True
+    return False
+
+
+def _validate_access_token_claim_bindability(
+    username: object, user_id: object, db: Session
+) -> tuple[str, int]:
+    """Validate claim values against the bound database dialect's parameters."""
+    if type(username) is not str or type(user_id) is not int:
```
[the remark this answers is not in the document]
AlexLiu190625: The exact type check is intentional at this JWT claim boundary: only built-in str and int values are accepted, while bool and subclasses are rejected before database binding. The suggested isinstance form would accept str and int subclasses and weaken that exact contract. Ruff E721 also accepts exact comparisons written with is/is not, so this remains unchanged.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
 from sqlalchemy.orm import Session
 
 from .auth_config import JWT_ALGORITHM, JWT_SECRET_KEY
 from .models.database import get_db
 from .models.user import User
 
-logger = logging.getLogger(__name__)
-
 # JWT Bearer token authentication
 security = HTTPBearer()
```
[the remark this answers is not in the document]
AlexLiu190625: HTTPBearer auto_error=True is intentionally retained to preserve the existing missing-header 403 behavior. This PR changes infrastructure-exception classification only, and there is currently no production Depends(get_current_user_optional) consumer. Returning anonymous None for a missing header would be a separate public behavior change, so this PR does not add a second bearer scheme.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
+        Exception: Database and unexpected operational failures propagate unchanged.
+    """
+    try:
+        return _resolve_access_token_user(credentials.credentials, db)
+    except _AccessTokenRejected as rejected:
+        _required_http_rejection(rejected.reason)
 
 
 def get_current_user_optional(
     credentials: Optional[HTTPAuthorizationCredentials] = Depends(security),
```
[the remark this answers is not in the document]
AlexLiu190625: The optional helper keeps its existing Depends(security) signature deliberately: direct credentials=None calls return None, while an actual FastAPI dependency with a missing header retains the established framework 403. There are no production dependency consumers today, and changing that missing-header policy is outside the infrastructure-failure boundary fixed here.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
+
+def _resolve_access_token_user(token: str, db: Session) -> User:
+    """Return the matching user or raise a typed credential rejection."""
     try:
-        token = credentials.credentials
+        payload = jwt.decode(
+            token,
+            JWT_SECRET_KEY,
+            algorithms=[JWT_ALGORITHM],
+            options={"verify_sub": False},
```
rogercloud, line 69: **MODERATE — `options={"verify_sub": False}` is an observable behavior regression against this PR's own preservation claim.**

For a token with a non-string `sub` claim:
- **Before** (reproduced by running the base-commit code against a real non-string-`sub` token): `401 "Invalid token"` with an `Error-Type: InvalidToken` response header.
- **After** (pinned at `tests/web/auth_token_cases.py:88-92`): `401 "Invalid token payload"`, no `Error-Type` header.

This is not cosmetic. `frontend/src/lib/api-wrapper.ts:~158` and `frontend/src/contexts/auth-context.tsx:~147` both branch on the `Error-Type` header value; with it absent, the old immediate-logout-on-`InvalidToken` path is skipped and the client instead attempts a token refresh, potentially still reporting itself as authenticated from cached session state. That is a real change to session handling, and it contradicts the PR description's statement that "malformed credentials keep their existing authentication outcomes."

No security regression — `verify_signature` / `verify_exp` / `verify_nbf` all remain enabled. But please either restore the `Error-Type: InvalidToken` header for this case, or document the change in the PR description and confirm the new frontend behavior is intended.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
+
+class _AccessTokenRejected(Exception):
+    """Expected access-token rejection without implementation details."""
+
+    def __init__(self, reason: _AccessTokenRejectionReason) -> None:
+        super().__init__()
+        self.reason = reason
+
+
+def _has_matching_temporal_claim_conversion_failure(
```
rogercloud: **MINOR — classification here rests on type identity, and is safe only incidentally to the current jose version.**

`_has_matching_temporal_claim_conversion_failure` decides that a `TypeError`/`OverflowError` from `jwt.decode` is a credential rejection by re-parsing the unverified claims and checking whether an exception of the same type reappears. That's type identity, not provenance: in the abstract, an unrelated `TypeError` from inside `jwt.decode` would be classified as `INVALID_CLAIMS` if the token also happened to carry a matching-shaped malformed temporal claim.

I traced the installed `python-jose` 3.5.0 source end to end and confirmed this is not reachable today: with this version, the only way a bare `TypeError`/`OverflowError` escapes `jwt.decode()` is exactly the `int(claims[...])` conversion on `iat`/`nbf`/`exp` — the same conversion this helper re-checks. So no bug to fix now.

But that guarantee lives in a third-party library with no upper-bound pin, not in this code. Suggest a short comment recording the reachability argument and the jose-version coupling, so a future jose upgrade that raises a bare `TypeError` from anywhere else doesn't silently turn an infra/library fault back into a 401.

(Related: `test_unrelated_decode_type_errors_propagate_by_identity` at `tests/web/test_auth_dependencies.py:219-241` only proves the negative case, and needs to stub both `jwt.decode` and `jwt.get_unverified_claims` to construct it. That's fine given the above, just noting there's no test for the misclassification scenario itself.)

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
+            int(claims[claim_name])
+        except ValueError:
+            continue
+        except (TypeError, OverflowError) as reproduced_error:
+            if type(reproduced_error) is type(original_error):
+                return True
+    return False
+
+
+def _validate_access_token_claim_bindability(
```
rogercloud, line 42: **MINOR — bindability validation is sqlite/postgresql only.**

`_validate_access_token_claim_bindability` recognizes only the `sqlite` and `postgresql` dialects. On any other dialect (MySQL, for instance), an out-of-range or NUL-containing claim value flows straight into the query and surfaces as an unclassified driver error → 500 rather than a clean credential rejection.

I confirmed this is deliberate — `test_unrecognized_dialect_runs_the_query_and_propagates_backend_failure` (`tests/web/test_auth_dependencies.py:366-384`) pins it, with the docstring "Unknown dialects retain operational behavior instead of invented limits." That's a defensible trade-off, so no change required. Flagging only so non-sqlite/postgresql deployments are aware that this defense doesn't apply to them; a one-line comment naming the covered dialects at the top of the helper would make that explicit at the call site.

src/xagent/web/tools/config.py
```diff
@@ -3227,7 +3190,7 @@ async def _build_mcp_server_config(
                     registration_generation=registration_generation,
                     resolved=remote_hook_token,
                     providers=remote_providers_to_resolve,
-                    user_id=int(self._user_id),
+                    user_id=int(cast(int, self._user_id)),
```
rogercloud, line 3200: **MINOR — this `cast(int, self._user_id)` is load-bearing, and its safety guard lives in an unrelated function with no linking comment.**

I confirmed by running mypy that removing this cast (and its twin at ~:2098) produces real `arg-type` errors — mypy genuinely sees `int | None` here, so the cast is silencing a live signal, not decorative. The guard that makes it safe (`self._user_id is None` check) lives in `_resolve_oauth_token_from_hook` (~:2520), two frames away, with no comment tying the two together. A future refactor of that guard would not be caught by mypy — only by a runtime `TypeError`.

Related, same underlying issue: `"user_id": str(self._user_id)` and `"allow_users": [str(self._user_id)]` elsewhere in this class (search for `_build_unavailable_mcp_config` and the end of this function) have no `is None` guard at all — if `self._user_id` is ever `None` there, it would silently stringify to the literal `"None"` and be used as an MCP isolation identity rather than failing loudly. Not currently exploitable (all four real `WebToolConfig` construction sites pass an explicit `int`), but worth an explicit `assert self._user_id is not None` (or a raise) at these boundaries, plus a short comment linking this cast to its guard in `_resolve_oauth_token_from_hook`.

src/xagent/web/tools/config.py
```diff
@@ -1166,9 +1163,7 @@ def __init__(
         self._db_factory = db_factory
         self._lazy_db = None
         self.request = request
-        self._user_id = (
-            user_id if user_id is not None else self._get_user_id_from_request(request)
-        )
+        self._user_id = user_id
```
rogercloud, line 1166: **MINOR — the identity-fallback removal is asymmetric: `is_admin` is still request-derived.**

`user_id` no longer falls back to the request (good), but `is_admin` still does, via `_get_is_admin_from_request` (:1177, :1293-1300). So `WebToolConfig(db=..., request=<request with user.is_admin=True>)` yields `user_id=None, is_admin=True` — privilege without identity.

That shape is dangerous downstream: `knowledge_tools.py` passes it unguarded into `list_collections` → `lancedb_stores.py:307-329`, where `is_admin` is checked before the `user_id is None` guard. `is_admin=True, user_id=None` would therefore return all knowledge-base collections across all tenants.

I verified this is not currently reachable — every production `WebToolConfig(...)` call site passes an explicit `user_id`. The only place the dangerous request shape is constructed is this PR's own test (`tests/web/tools/test_webtoolconfig_identity.py:57-71`), and that test asserts only `get_user_id() is None`, leaving `is_admin` unpinned.

Given the PR's stated goal is removing legacy request-derived fallback identities generally, suggest either (a) deriving `is_admin` only when `user_id` is also present, or at minimum (b) extending the identity-free test to assert `is_admin is False`, so the combination can't silently become reachable later.

tests/web/api/test_websocket_auth.py
```diff
@@ -0,0 +1,504 @@
...
+) -> None:
+    run_db_io = MagicMock()
+    monkeypatch.setattr(websocket_auth, "run_db_io_cancellation_safe", run_db_io)
+
+    assert await websocket_auth.get_authenticated_user(MagicMock(), None) is None
+    run_db_io.assert_not_called()
+
+
+@pytest.mark.parametrize("case", REJECTED_ACCESS_TOKEN_CASES, ids=lambda case: case.id)
+def test_loader_rejects_every_credential_reason_inside_worker_session(
```
rogercloud, line 108: **MINOR — this test is vacuous for 4 of its 5 parametrized cases.**

`test_loader_rejects_every_credential_reason_inside_worker_session` relies on the fake session's `first()` returning `None` unconditionally — regardless of whether the query ran at all — and never asserts `query_count == 0`. So a regression that removed the early-return short-circuit for `expired` / `invalid` / `wrong_type` / `invalid_claims`, letting them fall through into the (no-op) query, would pass this test unchanged. Only the `user_not_found` case is genuinely exercised.

The HTTP-side equivalents already do this correctly (`tests/web/test_auth_dependencies.py:130`, `:297`, `:346` all assert the query count). Suggest adding the same `query_count == 0` assertion here to match.

src/xagent/web/api/websocket_auth.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if not token:
+        return None
+    try:
+        return await run_db_io_cancellation_safe(
+            lambda: _load_websocket_principal_sync(token)
+        )
+    except Exception as exc:
+        route_template = getattr(websocket.scope.get("route"), "path", "<unresolved>")
+        logger.exception(
```
rogercloud: **NIT — traceback sanitization is claimed but not verified.**

`logger.exception(...)` writes the full exception and traceback server-side. The PR description's sanitization claim holds for the client-facing response — I verified that — but not for the log. The only relevant test, `test_operational_auth_failure_sends_sanitized_extension_denial`, deliberately injects an exception whose message contains sensitive-looking text (`"database pool token=secret"`) yet asserts only against `record.getMessage()` / `record.args`, never `exc_info` or the rendered traceback. I also found no `hide_parameters` setting on the SQLAlchemy engine anywhere in the repo, so DB exception text can carry bound parameters.

Suggest either extending that test to assert on the traceback content too, or scrubbing before logging if these logs are shipped to a less-trusted sink.

tests/web/api/test_websocket_auth.py
```diff
@@ -0,0 +1,504 @@
...
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    websocket = MagicMock()
+    websocket.close = AsyncMock()
+    websocket.accept = AsyncMock()
+    broadcaster = MagicMock()
+    broadcaster.connect = AsyncMock()
+    authenticated_user = AsyncMock(side_effect=_authentication_terminated)
+    monkeypatch.setattr(
+        progress_ws, "get_authenticated_user", authenticated_user, raising=False
```
rogercloud: **NIT — `raising=False` is unnecessary here and weakens the test.**

`monkeypatch.setattr(progress_ws, "get_authenticated_user", authenticated_user, raising=False)` — I confirmed the attribute genuinely exists on the module, so `raising=False` buys nothing today. What it does buy is a failure mode: if `get_authenticated_user` is ever renamed or removed, this becomes a silently inert monkeypatch (the test keeps "passing" against unpatched code) instead of a loud `AttributeError`. Suggest dropping `raising=False`.

src/xagent/web/api/websocket_auth.py
rogercloud: **Re-review of the earlier comment on `websocket.scope.get("extensions", {})` — partially addressed; suggest applying the one-liner anyway.**

You're correct on the specific mechanism that was raised: a `TypeError` from `extensions` being `None` is caught by the `except Exception:` below (~:77), so it does not bypass the try block as originally claimed.

But the broader conclusion ("did not bypass endpoint handling or hang") doesn't fully hold. When that `TypeError` fires, neither remediation branch runs — no `send_denial_response`, no `accept()` + `close()` — and the function falls through to unconditionally raise `_WebSocketAuthenticationTerminated()`. The callers simply `return` on that exception, on the assumption that a terminal response was already sent. It wasn't. The socket is left with no explicit ASGI accept or close from this code path.

There's also no test in the repo exercising `extensions=None` in either direction, and "Uvicorn's real ASGI scope never sets `extensions: None`" is plausible under the ASGI spec but isn't backed by an in-repo test or a spec citation.

Since the originally suggested fix is one line and costs nothing, I'd just take it and close the gap independent of whether production can currently trigger it:

```python
extensions = websocket.scope.get("extensions") or {}
```

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,230 @@
...
+    if user is None:
+        raise _AccessTokenRejected(_AccessTokenRejectionReason.USER_NOT_FOUND)
+    return user
+
 
-        # Validate JWT token
-        payload = jwt.decode(token, JWT_SECRET_KEY, algorithms=[JWT_ALGORITHM])
-        token_type = payload.get("type")
-        if token_type != "access":
+def _required_http_rejection(reason: _AccessTokenRejectionReason) -> NoReturn:
```
rogercloud, line 98: **Simplification — `_required_http_rejection`'s `match`/`case` over the 5 enum members (:121-155) is all static data.**

Every branch produces a constant `(detail, headers)` literal with no branch-specific logic, so a module-level `reason -> (detail, headers)` mapping plus a single lookup would cut this to roughly a third of the lines and make the response table readable at a glance.

Trade-off worth weighing: `match` currently gives you static exhaustiveness — mypy verifies every enum member is handled, so adding a sixth reason is a type error rather than a `KeyError` at runtime. If you make the swap, add a parallel assertion (e.g. a startup check that the dict's keys cover the enum) to keep that property.

Note I'm deliberately not suggesting simplification of the dialect bindability checks, the enum/exception dispatch design itself, or the temporal-claim proof step — I looked at all three and each is load-bearing; collapsing them would reintroduce the misclassification this PR fixes.

rogercloud (changes requested):
## Summary

This PR makes `_resolve_access_token_user` (`src/xagent/web/auth_dependencies.py`) the single owner of authentication-failure classification: genuine credential rejections raise one private typed `_AccessTokenRejected(reason)` carrying a 5-value enum (EXPIRED / INVALID / WRONG_TYPE / INVALID_CLAIMS / USER_NOT_FOUND), while everything else — DB and other infrastructure failures — propagates untouched. Required HTTP auth, optional HTTP auth, the token helper, and a new `src/xagent/web/api/websocket_auth.py` boundary are all thin projections of that one signal, with the WebSocket side now mapping infra failures to a sanitized 503 / close code 1011 instead of misreporting them as 4001/1008 credential rejections. `WebToolConfig` and `user_context.py` additionally drop the legacy request-derived fallback identities. Fixes #930.

## Approach assessment: acceptable with reservations

The core restructuring is the right shape. One classification owner, a typed rejection signal, and thin per-transport adapters is exactly how you stop infra failures from being laundered into 401/4001s, and it genuinely fixes #930 for the paths it touches. `_AccessTokenRejected` stays private — no parallel exception hierarchy escapes the module — and the `websocket_auth.py` extraction removes real duplication across four endpoints. Two mechanisms that initially looked speculative (the temporal-claim heuristic and the dialect-specific bindability checks) turned out on verification to be load-bearing or deliberately tested trade-offs, so I am not asking for them to be simplified away.

Three design-level notes that don't attach to a single line:

**1. The completeness claim is not met (see F1 below).** The PR presents itself as fixing the WebSocket authentication boundaries, but `src/xagent/web/api/public_chat_access.py` — the public-widget and share-link WebSocket auth paths — is not in this diff at all and still contains the identical bug: DB work via `run_db_io_cancellation_safe` inside a broad `except Exception:` that emits close code 4001 "Authentication required". That is arguably the most reachable instance of the bug class (no exotic token required, just pool pressure on a public endpoint), and it constructs the same `WebSocketPrincipal` type this PR touches elsewhere. Either extend the fix there or narrow the PR description's scope claim.

**2. `_WebSocketAuthenticationTerminated` is a private-named public contract.** It is defined with a leading underscore in `websocket_auth.py:28` but imported and caught at four call sites across `websocket.py` and `progress_ws.py` — that is a cross-module control-flow contract, not a private symbol. The custom AST topology test does meaningfully mitigate the "a call site forgets to catch it" regression, so this is a naming/self-documentation point rather than a correctness one: either drop the underscore, or consider returning a result type instead of exception-based control flow. Non-blocking.

**3. The topology test is unusual machinery.** `tests/web/test_auth_dependencies.py:502-745` is a ~244-line AST walker over the whole source tree asserting exact `Counter` equality over `(file, scope, callee, except-handler-names)` tuples and rejecting aliased imports / first-class-callable escapes. It buys real protection (it is what mitigates note 2), so I'm not asking for its removal — but it is disproportionate for policing ~5-6 known functions, and its failure mode is a raw Counter diff rather than "you broke X". A narrower set of targeted unit tests, or an import-boundary lint rule, would likely give comparable protection with far more readable failures. Worth a comment at minimum.

**Test-coverage gap not tied to one line:** neither `get_user_from_token` nor `get_user_from_websocket_token` has its own credential-rejection matrix test — the 5-reason matrix is only exercised against `get_current_user` / `get_current_user_optional`. Both delegate to the shared resolver (and the topology test enforces that), so a regression would probably be caught elsewhere, but the description's "every adapter × every reason" framing isn't literally what's tested. A pure-garbage non-JWT string and `get_user_from_token("")` are also uncovered. A smoke test per helper against the same matrix would close this.

## Findings

Severity-ordered; details are in the inline comments.

| # | Severity | Location |
|---|---|---|
| F1 | **MAJOR** | `src/xagent/web/api/public_chat_access.py:1317-1319`, `:1419-1421` — bug #930 still present, untouched |
| F4 | **MODERATE** | `src/xagent/web/auth_dependencies.py:95` — `verify_sub: False` drops the `Error-Type` header; observable client behavior change |
| F2 | MINOR | `src/xagent/web/auth_dependencies.py:42-61` — temporal-claim classification is type-identity based; safe only incidentally to jose 3.5.0 |
| F3 | MINOR | `src/xagent/web/auth_dependencies.py:64-85` — bindability checks cover only sqlite/postgresql |
| F5 | MINOR | `src/xagent/web/tools/config.py:~2821`, `~3297` — `str(self._user_id)` can emit `"None"` as an MCP isolation identity |
| F6 | MINOR | `src/xagent/web/tools/config.py:1166` vs `:1177` — `is_admin` still request-derived while `user_id` is not |
| F8 | MINOR | `tests/web/api/test_websocket_auth.py:107-114` — vacuous for 4 of 5 parametrized cases |
| F7 | NIT | `src/xagent/web/api/websocket_auth.py:61-65` — traceback sanitization unverified |
| F9 | NIT | `tests/web/api/test_websocket_auth.py:437` — unnecessary `raising=False` |

- **`_required_http_rejection` (`src/xagent/web/auth_dependencies.py:121-155`)** — every `match`/`case` branch is a static `(detail, headers)` literal with no branch-specific logic. A `reason -> (detail, headers)` dict lookup would cut this to roughly a third of the lines. Trade-off: you lose `match`'s static exhaustiveness checking (mypy currently verifies every enum member is handled) unless you add a parallel assertion.
- **The AST topology test (`tests/web/test_auth_dependencies.py:502-745`)** — see design note 3 above; a grep-based import-boundary lint rule would cover most of the same invariant with far less bespoke code.

I deliberately checked and am not proposing simplification of the dialect bindability checks, the enum + exception + `match` rejection dispatch, or the temporal-claim proof step — in each case the "simpler" alternative would reintroduce the exact misclassification this PR exists to eliminate.

Net: modest — essentially one dict-lookup refactor, plus an optional swap of the topology test for a lint rule.

## Previously raised (re-review)

All four prior comments from `gemini-code-assist` were answered inline by the author. Verified independently against current code:

- **P1 — `websocket_auth.py:67`, `scope.get("extensions", {})` — still open (low/medium priority suggestion).** The author is right about the specific mechanism: a `TypeError` here is caught by the surrounding `except Exception:`, so it does not bypass the try block as originally claimed. But the broader conclusion overstates it — when that `TypeError` fires, neither remediation branch runs (no `send_denial_response`, no `accept()` + `close()`); the function falls through to unconditionally raise `_WebSocketAuthenticationTerminated()`, and callers just `return`, trusting a terminal response was already sent. It wasn't. The socket is left with no explicit ASGI accept/close from this path. No test exercises `extensions=None` either way, and "Uvicorn never sets `extensions: None`" is plausible per the ASGI spec but not backed by an in-repo test or citation. The originally suggested one-liner (`websocket.scope.get("extensions") or {}`) is cheap and closes the gap regardless — recommend applying it.
- **P2 — `auth_dependencies.py:68`, `type(v) is not int` — resolved, with one note for the record.** Current code is correct and there is a dedicated test for a bool `user_id` being rejected; no change needed. One imprecision in the rebuttal: gemini's suggested snippet already contained an explicit `not isinstance(user_id, bool)` exclusion, so it would not have accepted bools — that framing was a strawman. The author's real point (isinstance also admits non-bool `int`/`str` subclasses) is technically valid but practically moot, since these claims come from JSON decoding and never produce custom subclasses.
- **P3/P4 — `auth_dependencies.py:16`, `:183`, `HTTPBearer()` / `auto_error=True` — waived for this PR; please open a follow-up issue.** Gemini's technical premise is correct (`HTTPBearer()` does default to `auto_error=True`, confirmed in the installed FastAPI source), so `get_current_user_optional` would 403 on a missing header before its body ran. But there is genuinely zero production route wiring `Depends(get_current_user_optional)` today — every call site is a test invoking it as a plain function, bypassing DI — and this `Depends(security)` signature predates this PR at the base commit; this PR only changed internal exception classification. Legitimate scope boundary. Still worth a tracked follow-up, because the function's name and docstring promise "optional" semantics it would not deliver the moment it is adopted as a route dependency.

## Tests

Full relevant suite run locally on Python 3.11: **180 passed, exit code 0** — `tests/web/test_auth_dependencies.py` (68), `tests/web/api/test_websocket_auth.py` (32), `tests/web/api/test_websocket_owner_actor.py` (59), `tests/web/api/test_progress_ws.py` (15), `tests/web/tools/test_webtoolconfig_identity.py` (5), `tests/web/test_user_context.py` (1).

## Recommendation

**Request changes** — on two points, not on the design:

1. **F1**: the PR states it addresses WebSocket authentication boundaries, but an identical and more easily reachable instance of bug #930 remains unfixed in `public_chat_access.py`. Either fix it here or narrow the stated scope.
2. **F4**: the PR states "malformed credentials keep their existing authentication outcomes"; for a non-string `sub` claim that is demonstrably false — the `Error-Type: InvalidToken` response header disappears, and the frontend branches on it. Either restore the header, or document the change and confirm the new client behavior is intended.

Everything else is minor or non-blocking. The central restructuring is sound and I'd be happy to approve once those two are resolved.

src/xagent/web/api/websocket_auth.py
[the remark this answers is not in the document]
AlexLiu190625: Applied defensively. The extensions value is now normalized with scope.get("extensions") or {}, and a raw ASGI regression pins extensions=None to the extensionless accept-and-close 1011 outcome.

src/xagent/web/auth_dependencies.py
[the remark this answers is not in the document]
AlexLiu190625: Applied. Optional authentication now uses a distinct HTTPBearer(auto_error=False) dependency, while required authentication keeps its existing scheme.

src/xagent/web/auth_dependencies.py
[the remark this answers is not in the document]
AlexLiu190625: Applied. get_current_user_optional now consumes the optional bearer scheme. Real FastAPI dependency tests cover missing, empty Bearer, and Basic headers as anonymous, with required authentication retaining 403.

src/xagent/web/auth_dependencies.py
AlexLiu190625: Fixed. A present non-string subject is classified as InvalidToken before application token type validation, preserving 401 Invalid token and the Error-Type: InvalidToken header. Absent subject remains Invalid token payload, and both paths are pinned before any query.

src/xagent/web/auth_dependencies.py
AlexLiu190625: No broader catch was added. The temporal helper remains a narrow provenance check for the installed python-jose behavior: malformed temporal conversions are rejected as credentials, while an injected unrelated TypeError propagates by object identity. A future dependency change that introduces another raw conversion source will therefore fail operationally rather than being silently generalized here.

src/xagent/web/auth_dependencies.py
AlexLiu190625: No dialect expansion was made. SQLite and PostgreSQL are the supported bindability contracts; unknown dialects intentionally retain backend-owned operational behavior instead of applying invented limits. The shared helper and tests keep that boundary explicit.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed at the owner boundary. Identity-free tool configs return no MCP configs before cache, database, or resolver work, and both MCP output builders use one explicit serializer that raises instead of emitting the string None.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed. WebToolConfig now forces is_admin=False whenever user_id is absent. Identified configs retain the existing tri-state behavior, including explicit false and request fallback only when an identity is present.

tests/web/api/test_websocket_auth.py
AlexLiu190625: Fixed. The shared five-reason loader matrix now asserts zero queries for expired, invalid, wrong-type, and invalid-claims credentials, and exactly one query for user-not-found.

src/xagent/web/api/websocket_auth.py
AlexLiu190625: Not changed intentionally. Client responses and structured log context are sanitized, but the original server-side traceback is retained for diagnosis. The PR description now states that boundary explicitly; tests verify that tokens, query strings, and database text do not enter client frames or structured context.

tests/web/api/test_websocket_auth.py
AlexLiu190625: Fixed. The unnecessary raising=False was removed from the progress authentication monkeypatch so a renamed or missing production symbol fails the test loudly.

src/xagent/web/api/websocket_auth.py
AlexLiu190625: Applied. extensions=None is now treated as an extensionless scope and produces an explicit accept followed by a sanitized 1011 close; the raw ASGI path is covered directly.

src/xagent/web/auth_dependencies.py
AlexLiu190625: Not changed. The match plus assert_never form preserves static exhaustiveness when a rejection reason is added. Replacing it with a mapping would require a parallel runtime completeness guard and would not improve the failure-classification owner.

AlexLiu190625:
Addressed the review round across the full authentication boundary:

- Public Widget and Share token owners now distinguish expected credential rejection from database, deployment, and decoder failures. Existing public HTTP 401 and WebSocket 4003 access outcomes remain unchanged; accepted sockets use a sanitized 1011 close for infrastructure failures during both initial authorization and per-message revalidation.
- Optional HTTP authentication now uses a real optional bearer dependency. Present non-string subject claims preserve the established InvalidToken header contract.
- Every signed public ID is type-checked before binding lookup and range-checked before query, while workforce-first selection remains unchanged.
- Identity-free tool configuration cannot derive admin privilege, read MCP state, or emit an isolation identity.
- The private terminal marker and existing source-tree consumer inventory remain in place because they enforce the single terminal-response owner and catch-policy topology. Direct token-helper smoke tests now cover empty and malformed input.

The updated branch passes 427 focused and adjacent tests plus changed-file lint, formatting, type, and pre-commit checks. The PR description now records the supported-dialect and server-traceback boundaries explicitly.

rogercloud (approved):
## Summary

This PR rewrites the shared authentication boundary so that *expected credential rejections* and *infrastructure failures* stop sharing an exit path. In `src/xagent/web/auth_dependencies.py`, the four entry points (`get_current_user`, `get_current_user_optional`, `get_user_from_token`, `get_user_from_websocket_token`) now classify a rejection into a typed `_AccessTokenRejected` / `_AccessTokenRejectionReason` at the decision point and map it to a specific HTTP response, while database/pool failures propagate uncaught instead of collapsing into 401 / `None` / a fallback identity. A new `src/xagent/web/api/websocket_auth.py` centralizes WebSocket auth so infra failures close with 1011 (or a pre-accept 503 denial) rather than misreporting as 4001/1008, and `public_chat_access.py`'s widget/share resolvers gain a `_PublicTokenRejected` marker plus dialect-aware integer-bindability validation (new `jwt_validation.py`) so out-of-range IDs fail before reaching the database. `WebToolConfig` no longer falls back to a hardcoded user id or inherits admin privilege when no identity is available, and the dead request-based helpers in `user_context.py` are removed.

## Approach verdict: sound

The classify-then-decide separation at `_resolve_access_token_user` is the right shape: rejections are raised as typed values at the point where the decision is actually made, so no `except` block anywhere has to guess whether a failure was a bad credential or a broken database. I traced the control flow end-to-end for all four HTTP/token entry points and both WebSocket paths -- infra exceptions genuinely propagate by identity, and no broad `except Exception` remains in `auth_dependencies.py` to reintroduce the collapse.

The `is_exact_integer_bindable` dialect-bounds validation is **not** scope creep, which was my first suspicion. `User.id` / `Agent.id` / `Workforce.id` are real `Column(Integer)` (int4 on Postgres) with no bigint migration in the tree, so an out-of-range `sub` claim really does reach the backend as a `DataError` -- and this PR's own propagation change would then surface that as an unsanitized 500. Pre-validating closes exactly the hole the rest of the PR would otherwise open. It belongs here.

Design-level observations (no single line owns these):

- **Two different exception disciplines across the two boundaries.** `auth_dependencies.py` avoids broad `except Exception` entirely. `public_chat_access.py` instead wraps decode + claims + persistence + authorization in one large `try` and classifies via an `isinstance` ladder in the final `except`. This is correct *today* -- every `raise` inside those blocks is an `HTTPException` or `_PublicTokenRejected` -- but nothing enforces it: no lint rule, no type check, and the AST-based structural guard test tracks `auth_dependencies.py` and `websocket_auth.py` only, not `public_chat_access.py`. A future stray exception type added inside that block would silently become a propagated 500 where a 401 was intended. Either extend the structural guard's tracked scope to cover `public_chat_access.py`, or accept the residual risk explicitly on the grounds that this is a small and stable set of functions.
- **`already_accepted` is a caller-asserted bool.** `send_websocket_authentication_infrastructure_failure(websocket, exc, *, already_accepted: bool)` takes accept-state as an argument rather than deriving it from Starlette's `websocket.application_state`, which tracks this reliably. All 8 current call sites pass the correct value, so there is no live bug -- but it is an unenforced invariant, and getting it wrong in a future endpoint means a double-accept or a send-after-close. Deriving it inside the helper would make the failure mode unreachable.
- **PR description is inaccurate on the HTTP half.** The description says infra failures produce "sanitized HTTP 503 or WebSocket 1011". The WebSocket half is accurate. The HTTP half is not: infra failures from `get_current_user` / `get_current_user_optional` / `get_user_from_token` reach `app.py`'s `global_exception_handler`, which for non-`/v1/*` paths explicitly re-raises and lets Starlette produce a generic **500**, not a 503. This still satisfies #930 (its HTTP bullet requires only that failures reach the centralized infrastructure error path; only the WebSocket bullet names a code), so it is not a code defect -- but the description should be corrected so future readers don't go looking for a 503 that isn't there.
- **Naming nit:** `_WebSocketAuthenticationTerminated` is imported and caught in `websocket.py` and `progress_ws.py` despite the underscore-private name. It is a real cross-module contract; consider dropping the underscore. (`_PublicTokenRejected` is genuinely module-local and should keep its underscore.)

## Findings

All findings below are **minor** -- I found no critical or major correctness bug, and none of them weakens the invariants #930 actually defends. They are ordered by how close they sit to the issue's acceptance criteria. Line-level ones are also posted as inline comments.

1. No end-to-end test that an HTTP infra failure produces a *sanitized* response (`tests/web/test_auth_dependencies.py`)
2. Widget-path `guest_id` check is weaker than the share path it cites as its model (`src/xagent/web/api/public_chat_access.py`)
3. Every `_PublicTokenRejected` logs an empty reason (`src/xagent/web/api/public_chat_access.py`)
4. Infra-failure branch logs without `exc_info` (`src/xagent/web/api/public_chat_access.py`)
5. All credential-rejection logging removed from `auth_dependencies.py`
6. Test pins FastAPI's `HTTPBearer` rejection status code, which changed across FastAPI versions (`tests/web/test_auth_dependencies.py`)

- `shrink:` `src/xagent/web/auth_dependencies.py` -- `_AccessTokenRejectionReason` + the `match/case` + `assert_never` (~35 lines) map a reason to `(status, detail, headers)` with zero branch-specific logic; a `dict[_AccessTokenRejectionReason, tuple[...]]` lookup gets this to ~12-14 lines. Trade-off: loses mypy's static exhaustiveness check in exchange for a runtime `KeyError`. **This is the same suggestion raised in an earlier review round on this PR and the code has not changed since -- flagging it as still open, not as new.**
- `shrink:` `src/xagent/web/api/public_chat_access.py` -- the two final `except Exception` blocks in `get_public_chat_user`/`get_share_chat_user` are structurally identical; only the log prefix and the `HTTPException` detail string differ. A shared helper removes the duplication.

net: -~40 lines possible

## Testing

Ran the full suite for the touched auth / websocket / tool-config / user-context modules:

```
pytest tests/web/test_auth_dependencies.py tests/web/api/test_websocket_auth.py \
       tests/web/api/test_public_chat_websocket_db_boundary.py tests/web/api/test_progress_ws.py \
       tests/web/tools/test_webtoolconfig_identity.py tests/web/test_user_context.py
```
-> **162 passed, 0 failed.** `ruff check` clean; `mypy` clean on `auth_dependencies.py`, `websocket_auth.py`, and `jwt_validation.py`.

None of the findings above block merge -- approving.

tests/web/test_auth_dependencies.py
```diff
@@ -0,0 +1,812 @@
...
+        lambda credentials, db: get_current_user(credentials, db),
+        lambda credentials, db: get_current_user_optional(credentials, db),
+        lambda credentials, db: get_user_from_token(credentials.credentials, db),
+        lambda credentials, db: get_user_from_websocket_token(
+            credentials.credentials, db
+        ),
+    ),
+    ids=("required", "optional", "token", "websocket-token-alias"),
+)
+def test_auth_adapters_propagate_database_pool_timeout_by_identity(
```
rogercloud, line 431: **Coverage gap (minor):** these infra-failure tests assert propagation by exception *identity* at the function-call level, but no test mounts `get_current_user` / `get_current_user_optional` / `get_user_from_token` as a real FastAPI route dependency, raises a DB/pool timeout from the overridden `get_db`, and asserts the actual HTTP response is sanitized (no driver message or traceback in the body). Since this PR's whole point is that these now escape the dependency, what the client finally sees is the load-bearing property. A structurally similar E2E test exists at `tests/web/api/v1/test_auth.py::test_internal_exception_returns_v1_envelope_not_fastapi_detail`, but it covers the separate v1 API-key chain. Suggest one `TestClient` test per entry point.

src/xagent/web/api/public_chat_access.py
```diff
@@ -320,19 +350,43 @@ def get_public_chat_user(
         # request. Establishing the type here is what lets both branches pass
         # ``user_id`` straight to their ``ensure_widget_*_available`` helper.
         if not _is_strict_int(user_id) or not user_id or not guest_id:
```
rogercloud: **Minor:** `guest_id` is only truthiness-checked here, while the share path at line 462 does `not isinstance(guest_id, str) or not guest_id.strip()`. `PublicChatAccessContext` is a plain frozen dataclass with no runtime validation, so a non-`str` `guest_id` would flow straight through. Not reachable today (both minting sites produce a validated `str`, and tokens are HMAC-signed), but the comment directly above cites the share path's guard as the standard being applied -- and this line doesn't match it. Suggest: `... or not isinstance(guest_id, str) or not guest_id.strip():`

src/xagent/web/api/public_chat_access.py
```diff
@@ -367,24 +421,25 @@ def get_public_chat_user(
             channel_id=channel_id,
             guest_id=guest_id,
             auth_mode=auth_mode,
-            widget_agent_id=widget_agent_id,
+            widget_agent_id=selected_agent_id,
         )
     except Exception as exc:
         logger.error("Public chat token validation error: %s", exc)
```
rogercloud: **Minor:** `_PublicTokenRejected` is always raised bare and has no `__init__`, so `str(exc)` is `""` and this renders as `"Public chat token validation error: "` -- a blank reason for *every* rejected widget token (`raise ... from exc` doesn't change `str()`). Before this PR these paths raised `ValueError` carrying the actual reason (`"Invalid token type"`, `"User not found"`, ...), so this is a loss of diagnostic detail on the most common failure path. Same applies at line 515 for the share path. Suggest giving `_PublicTokenRejected` a `reason: str` argument, or logging `__cause__`.

Separately: this same log also fires on the branch that re-raises a genuine infrastructure failure, and without `exc_info=True` that first entry carries no traceback. It is recovered downstream (`app.py`'s global handler for HTTP route dependencies, or `websocket_auth.py`'s own `exc_info` log for socket paths), so nothing is lost overall -- but the first log an operator hits is the least useful one. Suggest `exc_info=True`, or moving the log below the `isinstance` ladder so only the propagating branch logs a traceback.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,207 @@
-"""Authentication dependency for user segregation"""
+"""Authentication dependencies for user segregation."""
 
-import logging
-from typing import Optional
+from enum import Enum, auto
+from typing import NoReturn, Optional, assert_never
```
rogercloud, line 4: **Minor:** the `logging` import and every log call were removed from this module, so expired / invalid / wrong-type / invalid-claims / user-not-found rejections are now completely silent server-side across all four entry points, with no downstream layer capturing the signal. #930 only requires infra-failure observability (which propagation to the global handler provides), so this isn't a gap against the issue -- but it removes the only signal that would surface credential-guessing or a token-format regression in production. Suggest one `logger.info`/`warning` at `_required_http_rejection` carrying the reason enum (no token material).

tests/web/test_auth_dependencies.py
```diff
@@ -0,0 +1,812 @@
...
+
+    @app.get("/required")
+    def required_endpoint(user: User = Depends(get_current_user)) -> dict[str, int]:
+        return {"user_id": user.id}
+
+    app.dependency_overrides[auth_dependencies.get_db] = lambda: db_session
+
+    client = TestClient(app)
+    assert client.get("/optional", headers=headers).json() == {"user_id": None}
+    assert client.get("/required", headers=headers).status_code == 403
```
rogercloud: **Minor:** this pins FastAPI's `HTTPBearer(auto_error=True)` rejection code. Not hypothetical -- FastAPI changed exactly this from 403 to 401 in 0.122.0 for RFC 6750/9110 compliance (fastapi/fastapi#12507), and `pyproject.toml` pins `fastapi >= 0.35.0` with no ceiling. A routine `pip install -U fastapi` breaks this test for a reason unrelated to this code. Suggest `status_code in (401, 403)` with a comment naming the 0.122.0 change.

src/xagent/web/auth_dependencies.py
```diff
@@ -1,166 +1,207 @@
...
+    if user is None:
+        raise _AccessTokenRejected(_AccessTokenRejectionReason.USER_NOT_FOUND)
+    return user
 
-        # Validate JWT token
-        payload = jwt.decode(token, JWT_SECRET_KEY, algorithms=[JWT_ALGORITHM])
-        token_type = payload.get("type")
-        if token_type != "access":
+
+def _required_http_rejection(reason: _AccessTokenRejectionReason) -> NoReturn:
```
rogercloud, line 98: `shrink:` `_AccessTokenRejectionReason` + this `match/case` + `assert_never` (~35 lines) map a reason to `(status, detail, headers)` with no branch-specific logic; a module-level `dict[reason, tuple[...]]` lookup would be ~12-14 lines. Trade-off: trades mypy's static exhaustiveness check (pre-commit does run mypy here, so it's a real check) for a runtime `KeyError`. Noting this is the **same suggestion from an earlier review round on this PR** and the code hasn't changed since -- raising it as still-open rather than new.

src/xagent/web/api/public_chat_access.py
```diff
@@ -367,24 +421,25 @@ def get_public_chat_user(
             channel_id=channel_id,
             guest_id=guest_id,
             auth_mode=auth_mode,
-            widget_agent_id=widget_agent_id,
+            widget_agent_id=selected_agent_id,
         )
     except Exception as exc:
```
rogercloud, line 456: `shrink:` this `except Exception` block and the one at 514-520 are structurally identical -- only the log prefix and the `HTTPException` detail string differ. A shared `_raise_public_token_http_rejection(exc, *, context: str)` would remove the duplication and give the logging issues above a single place to be fixed. (Not suggesting a shared exception type -- `_AccessTokenRejected` needs 5-way branching into distinct responses while `_PublicTokenRejected` collapses to one, so merging them would add indirection without removing real duplication.)

tests/web/test_auth_dependencies.py
AlexLiu190625: Fixed in 2c1f73f2. Required and optional authentication now run a valid JWT through real FastAPI dependencies with an overridden DB session that raises a pool timeout, then through the production global exception handler. Both assert a sanitized HTTP 500 response with no driver message or traceback. The direct token helper is not mounted as a dependency because there is no production Depends consumer for it; its exception-identity coverage remains at the helper boundary.

src/xagent/web/api/public_chat_access.py
AlexLiu190625: Fixed in 2c1f73f2. The Widget resolver now requires guest_id to be a non-empty string before any binding lookup, and the regression covers non-string rejection with zero bind work. Whitespace is intentionally preserved because the current WidgetAuthRequest contract accepts it; applying strip only in the consumer would mint a token that the resolver rejects, so that behavior was not changed here.

src/xagent/web/api/public_chat_access.py
AlexLiu190625: Fixed in 2c1f73f2. _PublicTokenRejected now carries a stable enum reason, and Widget/Share use one shared projection owner that logs only the safe reason. Operational exceptions are no longer logged at this classifier layer; they propagate unchanged to the HTTP or WebSocket terminal owner, which records the traceback once. Token, decoder cause, and database text are not included in the credential log.

src/xagent/web/auth_dependencies.py
AlexLiu190625: Not changed. Issue #930 requires infrastructure-failure observability, which is retained at the HTTP and WebSocket terminal owners. Logging every expected credential rejection from all helper surfaces would add high-volume authentication noise, while logging only _required_http_rejection would not cover the four entry points cited here. A reason-only metric or audit event can be added separately once an explicit credential-observability policy and volume boundary are defined.

tests/web/test_auth_dependencies.py
AlexLiu190625: Fixed in 2c1f73f2. The dependency test now asserts required authentication is rejected with either 401 or 403, while continuing to require optional authentication to return the anonymous result. This preserves the framework-owned behavior across the FastAPI 0.122 status transition without weakening the no-anonymous-required invariant.

src/xagent/web/auth_dependencies.py
AlexLiu190625: Not changed. The match plus assert_never form keeps the five rejection responses statically exhaustive under mypy. Replacing it with a dictionary would move an omitted future reason to a runtime KeyError and require a second completeness check, so it would reduce lines without simplifying or strengthening the classification owner.

src/xagent/web/api/public_chat_access.py
AlexLiu190625: Fixed in 2c1f73f2 as part of the logging correction. Widget and Share now share _project_public_token_failure for HTTPException passthrough, reason-only typed rejection logging, and operational propagation. The access-token exception hierarchy remains separate.

AlexLiu190625:
Review-body follow-up:

- WebSocket authentication failure projection now derives the handshake phase from Starlette application_state; callers no longer pass a separate accept-state boolean.
- The PR description now distinguishes the existing sanitized HTTP 500 boundary from pre-handshake WebSocket HTTP 503 and accepted-WebSocket 1011 outcomes.
- The public token resolvers remain fail-closed for unknown exceptions: only the typed credential marker maps to 401, while unexpected and operational failures propagate to the terminal owner as 5xx/1011. The existing AST guard was not expanded because source-shape analysis cannot determine whether a future exception was intended as a credential rejection.
- _WebSocketAuthenticationTerminated remains package-internal control flow rather than a public API contract.
- Access-token rejection logging was not restored: #930 requires infrastructure observability, and adding high-volume credential logs needs a separate reason-only metric/audit policy rather than logging only one of four helper surfaces.

Local verification: 179 focused tests passed; Ruff, format, mypy, pre-commit hooks, and diff checks passed.

Issue #1084, opened 2026-07-30
chore(checkpoint): retire the legacy Task checkpoint pointer


AlexLiu190625:
## Summary

Remove `Task.last_checkpoint_event_id` only after every reader/writer has migrated to the exact trace-row primary-key locator.

## Current behavior

The legacy logical-event pointer must remain during dual-read/write rollout and rollback. Keeping it indefinitely, however, preserves ambiguity and invites new consumers.

## Acceptance criteria

- [ ] Gate the change on telemetry proving no legacy reader/writer, no unresolved backfill ambiguity, and no active mixed-version node requiring the old field.
- [ ] Disable legacy dual write/read and prove repository-wide zero use before changing the model.
- [ ] Remove the column through online and offline migration SQL. Downgrade must recreate it and backfill the logical event ID by joining the retained exact primary-key pointer to `trace_events`, then restore a dual-read/write-compatible application before rollback.
- [ ] Keep exact running-task recovery and active interaction resume behavior on `trace_events.id`.
- [ ] Verify pruning and Task deletion ordering after column removal.
- [ ] Document the rollback boundary: downgrade is valid only while the compatibility barrier and retained data make reconstruction safe.
- [ ] Repository-wide zero-use proof includes lease snapshots/CAS, new-run cleanup, trace writers, prune, and every Task/owner deletion path.

## Dependencies

- [refactor(checkpoint): support atomic interaction checkpoint handoff](https://github.com/xorbitsai/xagent/issues/1071)
- [feat(a2a): expose typed task interaction continuation](https://github.com/xorbitsai/xagent/issues/1083)
- [ops(interactions): materialize legacy waiting tasks](https://github.com/xorbitsai/xagent/issues/1078)
- [feat(interactions): reconcile interaction state and liveness](https://github.com/xorbitsai/xagent/issues/1082)
- [ops(interactions): add mixed-version rollout controls](https://github.com/xorbitsai/xagent/issues/1076)

## Non-goals

- Changing the interaction protocol.
- Removing trace history required by active interactions.

AlexLiu190625:
## Implementation design

- This is deliberately separate from adding the exact pointer. The add/dual-read phase must remain independently deployable and reversible.
- Removal is a post-barrier schema cleanup, not a feature activation prerequisite.
- No application fallback may query by logical event ID after this issue lands.

## Linked prerequisites

[...]

PR #1217, opened 2026-08-09
feat(interactions): extend task deletion and checkpoint pruning to interaction rows

### LINKED ISSUE #1071: refactor(checkpoint): support atomic interaction checkpoint handoff

### LINKED ISSUE #1074: feat(interactions): add canonical task interaction schema

AlexLiu190625:
## What

Extends the task deletion paths and checkpoint pruning to `task_interaction_requests`:

- **Both purge implementations** (single-task purge and bulk owner deletion) delete a task's interaction rows after the checkpoint-pointer columns are NULLed and before the `trace_events` delete. Without this ordering, purging a task that holds an active interaction row raises `IntegrityError`: deleting the anchored trace row fires the FK's `ON DELETE SET NULL`, and the active-anchor CHECK forbids that NULL on active rows. Terminal rows pass cleanly either way — that asymmetry is by design, and the tests pin both cells on both backends.
- **Checkpoint pruning protects active interaction anchors**: the exclusion set grows from the task's own checkpoint pointer to also cover `resume_trace_event_id` of active-slot rows, applied after the ranked retention query, so the retention bound moves from limit+1 to limit+2. Terminal and merely-expired-but-active rows are treated differently on purpose: expiry is only authoritative for reclaim, so an expired active row's anchor stays protected.
- **A schema-presence gate** on all three paths: a database upgraded to a revision before the interaction table exists must purge and prune exactly as before. Without the gate, the absent table is an uncaught error on both purge paths (a 500 on task deletion), and on SQLite it lands in prune's degradation handler and re-registers an unclearable failure signal on every checkpoint write. Measured gate cost on PostgreSQL: 0.647 ms median, 1.90 ms p95 (N=1000) per prune — no caching warranted at that size. (measured against the engine-inspect form; the gate now inspects the session's own connection, which removes both per-call pool checkouts and is ~4.5x cheaper on PostgreSQL and ~12x on SQLite — the no-caching conclusion only strengthens).

Rows are deleted, not terminated first: the `tasks.id` FK cascades, terminating first would cost an extra write, and it would falsely imply interaction rows can outlive a purge.

## Closes #1071

This is the last undelivered clause of that issue. Per-clause accounting:

| Acceptance clause | Delivered by |
|---|---|
| Exact-row anchor column, FK + delete restriction | merged checkpoint-anchor work (#1137) |
| Checkpoint read contract dependency | #1137 |
| Shared reader/alias reuse | #1137 |
| Backfill + dual-read/write, partition-scoped resolution | #1137 |
| Lease recovery + CAS fence on the PK locator | #1137 |
| Checkpoint staging joins a caller-owned transaction | #1207 |
| Deletion paths clear dependent pointers / **rows** | pointers: #1137 · **rows: this PR** |
| Deletion clears or terminates dependents before trace cleanup | #1137 + this PR |
| Coverage incl. online **and offline** migrations, prune, deletion ordering | #1137 · offline: #1204 · prune/ordering: this PR |

#1207's body names three-way joint closure and #1204 used `Refs`, so closure lands here by construction. Two non-homes, checked: #1084 cannot carry any of this (it lists #1071 as its own dependency), and this PR does not close #1074 — the existing-table field pairs and the datetime read-back half remain open there, as discussed on that issue.

## Deferred, registered

The row-level lock serializing the single-task purge against the interaction handoff's finalizer writes is not in this PR. With zero production handoff writers there is nothing to serialize against and no test can demonstrate the lock's value, while the lock itself is a real concurrency behavior change (and SQLAlchemy silently drops `with_for_update()` on SQLite). It lands with the handoff's first production writer. Until then, the bulk path's failure shape under a concurrent writer is one loud `IntegrityError` rolling back the whole purge with no partial state; retrying completes the deletion.

## Verification

- Both backends, full matrix: {single purge, bulk purge} × {active row, terminal row, no rows} × {table present, table absent} — active-row purge succeeds post-fix, terminal-row behavior byte-unchanged, absent-table paths no-op cleanly.
- Statement order pinned on three schema shapes (create_all SQLite with FK enforcement, Alembic-upgraded SQLite, PostgreSQL); reverting the ordering reproduces the `IntegrityError` on both backends.
- Nine mutations, each run bug-in → red → revert → green, including: ordering reverted, interaction delete removed (active cells red, terminal cells stay green), prune protection filters dropped or over-widened, gate forced on an absent table, and protection folded into the ranking query (caught by a dedicated in-window retention test).
- Pointer-pairing static guards green; zero production writers of interaction rows confirmed by grep.
- Prune coverage is SQLite-only: the protection delta is a presence gate already pinned on both backends plus Python-side set arithmetic, and the checkpoint-stream module has no PostgreSQL fixture layer today.
- Full regression: SQLite 4746 passed / 0 failed; PostgreSQL web suite 4725 passed / 0 failed on a disposable database.

### DIFF
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -915,6 +915,15 @@ class DatabaseTraceHandler(BaseTraceHandler):
             # delete would leave a steady state with nothing to prune unable
             # to clear the signal at all.
             clear_degradation(CHECKPOINT_PRUNE_FAILED)
+            if not stale_rows:
+                # Nothing ranks outside the window, so no protection set is
+                # needed to decide there is nothing to delete: skip the
+                # schema gate and the interaction-anchor query below.
+                # stale_ids is a filtered subset of stale_rows, so an empty
+                # stale_rows always reaches the same early return further
+                # down. Placed after the clear above so the healthy-attempt
+                # signal still fires on every write.
+                return
             # Protection set, second source: a trace row anchored by an
             # ACTIVE interaction row must survive retention, or answering
             # that interaction later would find no checkpoint to resume
@@ -992,7 +1001,8 @@ class DatabaseTraceHandler(BaseTraceHandler):
             # and cannot raise this -- that changes when the interaction
             # table's migration lands, whose rows enforce their CHECK on
             # upgraded SQLite too; a freshly created SQLite database
-            # (create_all, e.g. tests) has the FK and can. Under stricter isolation levels the same race can instead
+            # (create_all, e.g. tests) has the FK and can. Under stricter
+            # isolation levels the same race can instead
             # surface as psycopg2's SerializationFailure or DeadlockDetected,
             # both of which SQLAlchemy wraps in OperationalError, not
             # IntegrityError -- catch both so this retention path degrades
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -982,10 +982,17 @@ class DatabaseTraceHandler(BaseTraceHandler):
             # PostgreSQL enforces the anchor FK, so a race that lets a
             # candidate row become the active anchor between selection and
             # delete surfaces here as a restrict violation (IntegrityError).
-            # A database upgraded through Alembic on SQLite has no DB-level
-            # FK for this column and cannot raise this; a freshly created
-            # SQLite database (create_all, e.g. tests) does have the FK and
-            # can. Under stricter isolation levels the same race can instead
+            # Two anchors can do that: the task's checkpoint pointer, and an
+            # active interaction row whose resume_trace_event_id lands on a
+            # candidate -- the latter trips
+            # ck_task_interaction_requests_active_anchor on both backends
+            # (the ON DELETE SET NULL fires and the CHECK forbids the NULL
+            # on active rows). For the pointer FK, a database upgraded
+            # through Alembic on SQLite has no DB-level constraint today
+            # and cannot raise this -- that changes when the interaction
+            # table's migration lands, whose rows enforce their CHECK on
+            # upgraded SQLite too; a freshly created SQLite database
+            # (create_all, e.g. tests) has the FK and can. Under stricter isolation levels the same race can instead
             # surface as psycopg2's SerializationFailure or DeadlockDetected,
             # both of which SQLAlchemy wraps in OperationalError, not
             # IntegrityError -- catch both so this retention path degrades
--- a/src/xagent/web/api/trace_handlers.py
+++ b/src/xagent/web/api/trace_handlers.py
@@ -28,6 +28,7 @@ from ...core.tools.adapters.vibe.connector_runtime import (
 from ...web.models.database import get_db
 from ...web.models.task import Task, TaskStatus
 from ...web.models.task import TraceEvent as DatabaseTraceEvent
+from ...web.models.task_interaction import TaskInteractionRequest
 from ...web.models.tool_config import ToolUsage
 from ...web.services.ops_signals import (
     CHECKPOINT_DECODE_FALLBACK,
@@ -37,6 +38,7 @@ from ...web.services.ops_signals import (
     clear_degradation,
     register_degradation,
 )
+from ...web.services.task_interaction_schema import interaction_requests_table_exists
 from ...web.services.task_lease_service import (
     TASK_RUN_ID_TRACE_FIELD,
     current_task_lease,
@@ -853,9 +855,12 @@ class DatabaseTraceHandler(BaseTraceHandler):
         would discard the caller's pending writes and turn a misuse bug into
         silent data loss.
 
-        Retention is exactly ``limit`` rows, with one exception: when the
-        task's exact-row pointer names a row that itself ranks outside the
-        window, protecting that row keeps ``limit + 1``.
+        Retention is exactly ``limit`` rows, with up to two exceptions:
+        protecting a row named by the task's exact-row pointer, or by an
+        active interaction request's resume anchor, when that row itself
+        ranks outside the window keeps up to ``limit + 2`` (one pointer per
+        task, and at most one active interaction row per task -- see
+        ``uq_task_interaction_active_slot``).
         """
         if db.new or db.dirty or db.deleted:
             raise RuntimeError(
@@ -910,17 +915,54 @@ class DatabaseTraceHandler(BaseTraceHandler):
             # delete would leave a steady state with nothing to prune unable
             # to clear the signal at all.
             clear_degradation(CHECKPOINT_PRUNE_FAILED)
+            # Protection set, second source: a trace row anchored by an
+            # ACTIVE interaction row must survive retention, or answering
+            # that interaction later would find no checkpoint to resume
+            # from. Only active_slot IS NOT NULL rows protect: terminal rows
+            # are not resumable. Deliberately no expires_at filter -- an
+            # expired request is still answerable, so its anchor must stay.
+            #
+            # Placed after clear_degradation above, not before: on a
+            # deployment upgraded to a revision before this table exists the
+            # query raises OperationalError on SQLite and ProgrammingError
+            # on PostgreSQL, which would land in two different handlers and
+            # -- on SQLite -- register a CHECKPOINT_PRUNE_FAILED signal on
+            # every checkpoint that no operator could ever clear. The
+            # has_table gate makes that unreachable; the placement keeps the
+            # blast radius small if the gate is ever removed.
+            interaction_anchor_ids: set[int] = set()
+            if interaction_requests_table_exists(db):
+                interaction_anchor_ids = {
+                    row_id
+                    for (row_id,) in db.query(
+                        TaskInteractionRequest.resume_trace_event_id
+                    ).filter(
+                        TaskInteractionRequest.task_id == self.task_id,
+                        TaskInteractionRequest.active_slot.isnot(None),
+                        TaskInteractionRequest.resume_trace_event_id.isnot(None),
+                    )
+                }
             # Rank first, protect after. The row this task's exact-row
-            # pointer references is never deleted, whatever its position in
-            # the retention ranking; excluding it from the candidate set
-            # instead would shift every remaining row's OFFSET rank by one
-            # and retain limit + 1 rows whenever the anchor sits inside the
-            # window. For the steady-state writer the protection is
-            # structurally unreachable -- the pointer always names the row
+            # pointer references, and any row an active interaction request
+            # anchors, are never deleted, whatever their position in the
+            # retention ranking; excluding them from the candidate set
+            # instead would shift every remaining row's OFFSET rank and turn
+            # the retained count into an unpredictable limit + k. The
+            # exact-row pointer's protection is structurally unreachable for
+            # the steady-state writer -- the pointer always names the row
             # just written, which ranks ahead of the offset -- but it guards
-            # the backfill-vs-prune window and future back-pointing anchors
-            # that may point at an older row.
-            stale_ids = [row_id for (row_id,) in stale_rows if row_id != anchor_id]
+            # the backfill-vs-prune window and back-pointing anchors that
+            # point at an older row. The interaction anchor is exactly that
+            # kind of back-pointing anchor, arrived: a task can keep writing
+            # checkpoints after an interaction is raised, so the row the
+            # interaction resumes from steadily sinks in the retention
+            # ranking while the interaction stays open.
+            protected_ids = interaction_anchor_ids | (
+                {anchor_id} if anchor_id is not None else set()
+            )
+            stale_ids = [
+                row_id for (row_id,) in stale_rows if row_id not in protected_ids
+            ]
             if not stale_ids:
                 return
             # Chunk the IN clause: a backlog from previously-disabled pruning
--- a/src/xagent/web/services/task_interaction_schema.py
+++ b/src/xagent/web/services/task_interaction_schema.py
@@ -4,6 +4,21 @@ Deletion and retention paths run on deployments upgraded to a revision
 before task_interaction_requests exists. Absence is a normal deployment
 state, not a fault: the migration that creates the table has not been
 applied yet.
+
+The check is deliberately uncached. Measured on PostgreSQL it costs less
+than the trace INSERT plus commit the same checkpoint write already
+performs, and it fires once per checkpoint write, not per trace event.
+Caching False would be poisonous: the table's migration can be applied
+while processes are running, and a process that latched False before the
+migration would skip interaction rows in purge (reintroducing the
+IntegrityError this module exists to prevent) and stop protecting active
+resume anchors in prune, permanently, with no operator remedy short of a
+restart. If measurements ever justify caching, the only safe shape is a
+one-way latch: cache True alone -- the table is only ever added, never
+dropped, so a stale True cannot occur -- and let False re-query on every
+call. A cheaper no-cache option also exists: inspecting the session's own
+connection instead of the engine skips a pool checkout and roughly halves
+the cost.
 """
 
 from __future__ import annotations
--- a/src/xagent/web/services/task_interaction_schema.py
+++ b/src/xagent/web/services/task_interaction_schema.py
@@ -1,0 +1,19 @@
+"""Schema-presence predicate for the interaction table.
+
+Deletion and retention paths run on deployments upgraded to a revision
+before task_interaction_requests exists. Absence is a normal deployment
+state, not a fault: the migration that creates the table has not been
+applied yet.
+"""
+
+from __future__ import annotations
+
+from sqlalchemy import inspect
+from sqlalchemy.orm import Session
+
+from ..models.task_interaction import TaskInteractionRequest
+
+
+def interaction_requests_table_exists(db: Session) -> bool:
+    """True when this session's database has task_interaction_requests."""
+    return inspect(db.get_bind()).has_table(TaskInteractionRequest.__tablename__)
--- a/src/xagent/web/services/task_interaction_schema.py
+++ b/src/xagent/web/services/task_interaction_schema.py
@@ -16,9 +16,20 @@ resume anchors in prune, permanently, with no operator remedy short of a
 restart. If measurements ever justify caching, the only safe shape is a
 one-way latch: cache True alone -- the table is only ever added, never
 dropped, so a stale True cannot occur -- and let False re-query on every
-call. A cheaper no-cache option also exists: inspecting the session's own
-connection instead of the engine skips a pool checkout and roughly halves
-the cost.
+call.
+
+The check inspects the session's own connection, not the engine.
+Inspecting the engine checks out two connections per call: on SQLite's
+NullPool each is a new physical connection that re-runs the WAL,
+busy_timeout and foreign_keys connect pragmas (measured ~12x the cost),
+and on PostgreSQL each is a pooled checkout taken while the caller
+already holds one (~4.5x). Reusing the session's connection also keeps
+the gate's answer on the same snapshot as the query it guards. Under
+READ COMMITTED -- the isolation level this deployment runs -- the two
+forms are indistinguishable, including across a migration applied while
+the process is running; under REPEATABLE READ or SERIALIZABLE the
+session's snapshot can miss a table created after its transaction began,
+where the constraint layer still fails loudly rather than silently.
 """
 
 from __future__ import annotations
@@ -31,4 +42,4 @@ from ..models.task_interaction import TaskInteractionRequest
 
 def interaction_requests_table_exists(db: Session) -> bool:
     """True when this session's database has task_interaction_requests."""
-    return inspect(db.get_bind()).has_table(TaskInteractionRequest.__tablename__)
+    return inspect(db.connection()).has_table(TaskInteractionRequest.__tablename__)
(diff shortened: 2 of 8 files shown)


src/xagent/web/services/task_interaction_schema.py
```diff
@@ -0,0 +1,19 @@
...
+
+from sqlalchemy import inspect
+from sqlalchemy.orm import Session
+
+from ..models.task_interaction import TaskInteractionRequest
+
+
+def interaction_requests_table_exists(db: Session) -> bool:
+    """True when this session's database has task_interaction_requests."""
+    return inspect(db.get_bind()).has_table(TaskInteractionRequest.__tablename__)
```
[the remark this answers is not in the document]
AlexLiu190625: Measured before deciding. On PostgreSQL, `inspect(bind).has_table(...)` costs 0.668 ms median / 1.90 ms p95 (N=1000, full 53-table schema) — reproduced independently from the number in the PR description. For comparison, on the same database and session: the trace INSERT plus commit that the same checkpoint write already performs is 0.932 ms median / 2.21 ms p95, and prune's own three queries total about 1.0 ms median. The gate fires once per checkpoint write, not per trace event. At that size it isn't worth caching.

Two problems with the suggested implementation beyond the sizing. `db.get_bind()` returns an `Engine`, and `Engine` has no `.info` attribute in SQLAlchemy 2.0.47 — the snippet raises `AttributeError` on the first call. `Connection.info` exists, but it is per-pooled-connection, so it would produce one cache per connection that resets on recycle.

More importantly, caching `False` breaks the case the gate exists for. The migration that creates `task_interaction_requests` ships separately and can be applied while these processes are running. If a process caches `False` before the migration and never re-queries, the gate stays `False` for the process's lifetime after the table is real. Both consequences reproduced: purging a task that holds an active interaction row fails with the active-anchor CHECK violation — the exact `IntegrityError` this PR fixes, surfacing as a 500 since neither purge function catches it — and prune stops protecting active resume anchors, so the whole prune rolls back with nothing reclaimed and `checkpoint_prune_failed` registers on every subsequent checkpoint write with no operator remedy. Also worth noting: injecting the cached version into all four affected test modules leaves every test green, because each fixture mints a fresh engine — so CI would not catch this regression either.

If the numbers ever change, the safe shape is a one-way latch: cache `True` only, and let `False` re-query — the table is only ever added, never dropped, so a stale `True` cannot happen. There is also a cheaper option that needs no cache at all: inspecting the session's own connection instead of the engine measures 0.273 ms median, since roughly 0.4 ms of the cost is the extra pool checkout rather than the catalog query. Keeping the gate uncached; the latch reasoning is now recorded in the module docstring so the next reader finds it at the call site.

src/xagent/web/services/task_interaction_schema.py
```diff
@@ -0,0 +1,34 @@
...
```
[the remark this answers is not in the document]
AlexLiu190625: Agreed that the one-way latch is the correct shape — that is why it is written into the module docstring as the sanctioned design if the cost ever justifies caching. The open question was never safety, it was whether caching is warranted, and the measurements still say no.

Re-measured against PostgreSQL 16.14 (warm pool, 200 iterations, median): the presence check costs 0.645 ms, while the trace INSERT plus commit that accompanies it on the same checkpoint write costs 0.715 ms. The check runs once per checkpoint write, not per trace event, so the saving is under a millisecond on a path that already pays more than that to do its actual work.

Two costs on the other side. First, `Session.get_bind()` is typed `Engine | Connection`, and mypy runs over this package in pre-commit; the direct attribute form fails with two `union-attr` errors, so the latch needs a `WeakKeyDictionary` or a cast to compile — real added state for a sub-millisecond saving. Second, a session bound to a `Connection` rather than an `Engine` (one fixture in this repo does exactly that) would latch per-connection and effectively never hit the cache.

Leaving the check uncached and the latch documented as the shape to reach for if these numbers ever change.

src/xagent/web/api/trace_handlers.py
```diff
@@ -910,17 +915,54 @@ def _prune_checkpoint_history(self, db: Session, data: Dict[str, Any]) -> None:
...
+                interaction_anchor_ids = {
+                    row_id
+                    for (row_id,) in db.query(
+                        TaskInteractionRequest.resume_trace_event_id
+                    ).filter(
+                        TaskInteractionRequest.task_id == self.task_id,
+                        TaskInteractionRequest.active_slot.isnot(None),
+                        TaskInteractionRequest.resume_trace_event_id.isnot(None),
+                    )
+                }
```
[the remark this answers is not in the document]
AlexLiu190625: The two predicates are indeed interchangeable — `ck_task_interaction_requests_active_slot_pairs_status` is written as `(status = 'active') = (active_slot IS NOT NULL)`, so the database forbids a row where they disagree on either backend.

The index rationale does not hold, though. The current form already uses an index — the one backing `UNIQUE(task_id, active_slot)`. Measured on PostgreSQL 16.14 with 200,000 rows across 10,000 tasks, after ANALYZE: both forms produce an Index Scan with identical cost (0.29..8.31), identical row estimate (1), and identical buffer count (3) — one via `uq_task_interaction_active_slot`, the other via `ix_task_interaction_requests_task_status`. At 200 rows both remain index scans, and SQLite picks an index for both forms as well. `IS NOT NULL` is an indexable btree condition, so there is nothing here for the swap to fix — and the query returns at most one row per task regardless, since `UNIQUE(task_id, active_slot)` caps active rows at one.

On readability we read it the other way round: the protection set is about occupied slots, and `active_slot` is the column that enforces both the occupancy and the one-row bound, so the predicate names the invariant it depends on. Keeping the current form.

src/xagent/web/api/trace_handlers.py
```diff
@@ -910,17 +915,54 @@ def _prune_checkpoint_history(self, db: Session, data: Dict[str, Any]) -> None:
             # delete would leave a steady state with nothing to prune unable
             # to clear the signal at all.
             clear_degradation(CHECKPOINT_PRUNE_FAILED)
```
rogercloud, line 917: The schema-presence gate (line ~934) and the interaction-anchor SELECT (~934-943) run on every checkpoint write, including the common steady-state case where there is nothing to prune. The only "nothing to prune" early return is `if not stale_ids: return` at line 966 — after both. Pre-PR, this function returned immediately once `stale_ids` came back empty, so this cost is new.

Suggest adding an early return right here, before the gate:

```python
if not stale_rows:
    return
```

This is behavior-preserving: `stale_ids` is derived from `stale_rows`, so it can only be empty when `stale_rows` is empty. Perf-only, not correctness.

src/xagent/web/services/task_interaction_schema.py
rogercloud: `inspect(db.get_bind())` inspects the Engine, which checks out a **new physical connection** from the pool for the `has_table` call. On SQLite this repo configures `poolclass=NullPool` (`src/xagent/web/models/database.py:182-190`) plus a `connect`-event listener that reapplies the WAL / `busy_timeout` / `foreign_keys` pragmas on every new connection (`src/xagent/db/sqlite.py:69-75`) — so each call opens a fresh connection and re-runs pragma setup, rather than paying a cheap pool checkout.

All three call sites (`task_deletion.py:71`, `admin_users.py:112`, `trace_handlers.py:934`) call this from inside an already-open session transaction, so this opens a *second* independent connection alongside the session's own. Using the session's connection avoids that entirely:

```python
inspector = inspect(db.connection())
```

(`db.connection()` was mentioned in the earlier caching thread only as an aside, measured on PostgreSQL where it's ~0.4ms — the SQLite/NullPool path wasn't evaluated. Not a correctness issue, but it's a one-line change with no downside raised so far.)

src/xagent/web/api/trace_handlers.py
```diff
@@ -940,10 +982,17 @@ def _prune_checkpoint_history(self, db: Session, data: Dict[str, Any]) -> None:
...
+            # active interaction row whose resume_trace_event_id lands on a
+            # candidate -- the latter trips
+            # ck_task_interaction_requests_active_anchor on both backends
+            # (the ON DELETE SET NULL fires and the CHECK forbids the NULL
+            # on active rows). For the pointer FK, a database upgraded
+            # through Alembic on SQLite has no DB-level constraint today
+            # and cannot raise this -- that changes when the interaction
+            # table's migration lands, whose rows enforce their CHECK on
+            # upgraded SQLite too; a freshly created SQLite database
+            # (create_all, e.g. tests) has the FK and can. Under stricter isolation levels the same race can instead
```
rogercloud: Nit: this comment line is 116 chars while the rest of the same paragraph (lines 982-1005) wraps at 68-77 — looks like a sentence was appended without rewrapping. ruff won't flag it (E501 isn't in this repo's select set). Suggest rewrapping to match the surrounding block.

rogercloud (approved):
This PR closes the remaining gap in #1071 by teaching the task-purge and checkpoint-prune paths about `task_interaction_requests`. Those rows anchor to `trace_events` through a `resume_trace_event_id` FK declared `ON DELETE SET NULL`, while a CHECK constraint forbids that column from being NULL for an `active` row — so deleting trace rows out from under an active interaction either raises `IntegrityError` (purge) or silently strands a resume anchor (prune). The fix is three coordinated moves: both purge paths (single-task and bulk-owner) now delete a task's interaction rows after nulling checkpoint pointers and before deleting `trace_events`; checkpoint pruning adds the active-slot interaction's `resume_trace_event_id` to its retention-exclusion set after ranking, shifting the retention bound from `limit+1` to `limit+2`; and all three paths are gated on a shared schema-presence predicate because the migration creating the table has not landed yet.

**Re-review note.** There are no new commits since the previous review round — all 6 commits predate the end of the prior discussion (last commit 2026-08-09T18:47:05Z, last review comment 19:05:28Z). The three prior findings were resolved by rebuttal, not by code changes; I re-verified each rebuttal independently against the current code and all three hold (see *Prior findings* below). Everything in *New findings* is from an independent fresh pass layered on top.

### Design verdict

**Sound.** The purge ordering is correct and byte-identical between the two paths (`task_deletion.py:71-75` / `admin_users.py:110-115` — no drift); the prune change correctly protects *after* ranking rather than excluding from candidates, which preserves the "retain exactly `limit`" contract when a protected row already falls inside the window, and the `limit+2` bound is provably tight given `uq_task_interaction_active_slot` (≤1 active row per task); the gate is a single shared predicate used consistently at all 3 sites with positive `has_table` detection, so it cannot mask a real DB error. Deferring the row lock for concurrent interaction-handoff writers is a clean deferral — there are currently zero writers of `TaskInteractionRequest` rows anywhere in `src/`, so no live bug is being deferred.

**Reservation (not blocking):** no Alembic migration for `task_interaction_requests` exists yet, so on every real deployment the gate currently always returns `False` and the interaction work is pure overhead. That's defensible sequencing, but it does raise the weight of finding 2 below — today, the *only* thing this code does in production is pay the gate's cost.

### New findings (all minor/trivial, non-blocking)

**1. Gate + anchor query run on every checkpoint write, even with nothing to prune** — minor (perf)
`src/xagent/web/api/trace_handlers.py`, `_prune_checkpoint_history` (~line 838), gate at ~line 934, anchor SELECT at ~934-943.
`stale_rows` is computed at ~890-909, but the only "nothing to prune" early return (`if not stale_ids: return`) sits at line 966 — *after* the gate call and the interaction-anchor SELECT. In the steady state (nothing to prune, the common case) every checkpoint write now pays both. This ordering is new here: the pre-PR code had no gate and returned immediately after computing `stale_ids`.
Suggested fix: add `if not stale_rows: return` immediately after `clear_degradation(CHECKPOINT_PRUNE_FAILED)` (line 917), before the gate. This is behavior-preserving — `stale_ids` is always a subset of `stale_rows`, so it can only be empty when `stale_rows` is.

**2. `db.get_bind()` opens a second physical connection per call; on SQLite that re-runs pragma setup** — minor (perf), one-line fix
`src/xagent/web/services/task_interaction_schema.py:34`.
`inspect(db.get_bind()).has_table(...)` inspects the *Engine*, which checks out a fresh connection from the pool. This repo's SQLite engine uses `poolclass=NullPool` (`src/xagent/web/models/database.py:182-190`) plus a `connect`-event listener that reapplies WAL / `busy_timeout` / `foreign_keys` pragmas on every new connection (`src/xagent/db/sqlite.py:69-75`) — so on SQLite this isn't a pool checkout, it's a brand-new physical connection with full pragma setup, every call. All three call sites (`task_deletion.py:71`, `admin_users.py:112`, `trace_handlers.py:934`) invoke this from inside an already-open session transaction, so a second independent connection is opened concurrently with the session's own.
Suggested fix: `inspect(db.connection())` instead of `inspect(db.get_bind())` — reuses the session's existing connection. Note this alternative surfaced in the earlier thread only as an aside inside the *caching* debate, measured on PostgreSQL ("~0.4ms of the cost is the extra pool checkout"); it was never evaluated against the SQLite/NullPool path and no reason was given for not shipping it. Not correctness-affecting, but it's one line with no downside raised anywhere.

**3. Over-long comment line breaks the surrounding wrap** — trivial
`src/xagent/web/api/trace_handlers.py:995`.
This line is 116 chars against 68-77 for the rest of the same comment paragraph (982-1005) — an editing artifact where a sentence was appended without rewrapping. Not caught by ruff (E501 isn't in this repo's select set). Cosmetic only; rewrap to match.

### Prior findings (re-verified, all WAIVED)

- **A. Cache `interaction_requests_table_exists` via `bind.info[...]`** (`task_interaction_schema.py:34`) — **WAIVED.** Verified against the installed SQLAlchemy 2.0.51: `Engine` has no `.info` attribute (only `Connection` does), so the suggested code would raise `AttributeError` on first use. The author's correctness objection is also real — caching `False` is unsafe because the migration ships separately and could land under a running process, permanently defeating the gate. Rationale is durably recorded in the module docstring.
- **B. One-way latch cache (cache `True` only) on the bind object** — **WAIVED.** Both counter-costs verified: (i) `Session.get_bind()` is typed `Union[Engine, Connection]` and mypy runs on this file via pre-commit, so raw `setattr`/`getattr` on that union trips `union-attr` without a cast; (ii) a Session bound to a raw `Connection` genuinely exists in this repo's fixtures (`tests/web/tools/test_web_tool_config_session_factory.py:2563-2577`), so the "would rarely hit the cache" claim isn't exaggerated.
- **C. Query by `status == "active"` instead of `active_slot.isnot(None)`** (`trace_handlers.py:944`) — **WAIVED.** The load-bearing claim checks out: `CheckConstraint("(status = 'active') = (active_slot IS NOT NULL)")` exists at `src/xagent/web/models/task_interaction.py:219-220` and is enforced on both backends, so the two predicates cannot disagree for a committed row. The `UNIQUE(task_id, active_slot)` index that already covers the chosen predicate also exists as cited. Index-coverage vs. readability is a defensible call.

`tests/web/services/test_task_deletion_checkpoint_pointer.py`
L38: dup-import Two adjacent `from tests.web.services.task_interaction_schema_shared import (...)` statements from the identical module. Merge into one import statement.
L181: dup-helper `_tables_excluding_interaction_requests()` is logically identical to `_tables_excluding_target()` in `test_task_interaction_schema_gate.py:27-40` (same filter over `Base.metadata.sorted_tables`). Move to the existing `task_interaction_schema_shared.py` and import from both.

net: -8 lines possible

### Test results

Ran the PR's three test files locally (Python 3.12, full dev install): **93 passed, 12 skipped, 0 failed**. The 12 skips are the `postgresql`-marked tests (`XAGENT_TEST_POSTGRES_URL` unset locally).

Coverage quality is good: the purge-ordering tests assert statement order and would fail if the ordering were reverted (genuine discriminators), and the prune off-by-one bound and the gate-removal path each have dedicated pinning tests.

One gap: all 6 new prune tests are SQLite-only — there is no `@pytest.mark.postgresql` variant for the prune path, whereas the purge matrix does cover both backends. Risk is low since the protection logic is Python-side set arithmetic over already-fetched rows, but the PR description's "both backends" claim doesn't fully extend to prune.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Confirmed, and taken. `stale_ids` is a filtered comprehension over `stale_rows`, so an empty `stale_rows` always reaches the same early return further down — the new return is behavior-preserving by construction. It also sits immediately after the sole `clear_degradation` in this function (the only other signal call is the `register_degradation` in the except handler), so the healthy-attempt clear still fires on every write.

One measurement to set expectations: the retention limit defaults to 8, so `stale_rows` is only empty during an execution's first 8 checkpoint writes — measured over a 30-write sequence, the return fires 8 times and never again. Worth taking because short executions never prune at all and were paying two queries per write for nothing, but it is not a steady-state saving.

src/xagent/web/services/task_interaction_schema.py
AlexLiu190625: Taken, and it is worth more than the docstring claimed. `inspect(engine)` checks out **two** connections per call, not one. Measured: SQLite (NullPool, WAL/busy_timeout/foreign_keys connect pragmas) 0.131 ms → 0.011 ms, two physical connections and six pragma statements down to zero; PostgreSQL with the production pool kwargs 1.33 ms → 0.28 ms, two pooled checkouts down to zero. The pool point matters more than the latency: with pool_size=10 and max_overflow=20, every purge and prune was asking for a second connection while already holding one.

Gate semantics are unchanged where it counts. Under READ COMMITTED — no isolation_level is set anywhere in src and the server default is read committed — both forms agree in every cell, including a table created and committed by another connection mid-transaction. Under REPEATABLE READ or SERIALIZABLE they diverge: the session's snapshot misses a table created after its transaction began. Bounded and loud rather than silent — purge would hit the active-anchor CHECK, and prune's delete would trip the same CHECK into the degradation handler — and the window closes on the next transaction. The docstring now records this, replacing the "cheaper option also exists" paragraph this change makes obsolete, and the PR description's measurement line is annotated as having been taken against the engine form.

Verified across purge, prune and the gate tests on both backends with the change in place: all green.

src/xagent/web/api/trace_handlers.py
AlexLiu190625: Confirmed — 116 characters where the surrounding block runs 43–77. Not a lint failure (E501 is outside the default ruff selection and the formatter does not rewrap comments), just an unrewrapped edit. Split into two lines matching the paragraph's style, surrounding lines untouched.

AlexLiu190625:
On the remaining body items:

**The duplicate import pair** — leaving it. The split is isort's output, not an oversight: the repo's isort profile is "black", which does not enable combine_as_imports, so the default splits `as`-imports into their own statement. Merging them and re-running the pinned isort 6.0.1 puts the pair straight back, and the file is `isort --check-only` clean as it stands. The same shape appears twice more in this file and in four other test modules.

**The duplicated table filter** — hoisted to the schema-shared module as `tables_excluding_interaction_requests()`, dropping the leading underscore to match the module's other exports and keeping the gate copy's docstring, since it records the audit result. Both call sites updated, plus a scope line in that module's docstring.

**The prune test gap** — agreed on the risk rating; fixed the description rather than adding tests. Worth recording that the checkpoint-stream module has never had PostgreSQL coverage — every task in it comes from an in-memory SQLite engine, so the gap predates this change. Adding coverage would mean a new PG fixture layer in a module with none, and the prune delta is a presence gate already pinned on both backends plus IS NOT NULL and Python set arithmetic. The genuinely dialect-sensitive part of prune is the ranking query's JSON extraction, which this change does not touch and which is equally uncovered today; that is the piece worth a dedicated port of the module's fixtures if PG prune coverage is ever wanted. The Verification section now states the prune coverage boundary explicitly.

PR #1218, opened 2026-08-09
feat(migrations): create the task interaction request table

### LINKED ISSUE #1074: feat(interactions): add canonical task interaction schema

AlexLiu190625:
## What

Adds the Alembic revision for `task_interaction_requests`, plus parity tests pinning the migration-built schema against the `create_all`-built one. The model merged earlier, so freshly created databases already carry the table — but any database maintained through the revision chain has no way to get it. This revision closes that gap.

Coverage of the migration acceptance criterion in #1074: online upgrade and downgrade on both backends, offline `--sql` for all four cells, idempotency in both directions, guard behavior on a database missing the parent tables, and downgrade residue checked across tables, sequences, and indexes.

## Shape decisions

- **One `op.create_table`**: both dialects render the named foreign keys and every CHECK inline — the offline output contains zero `ALTER TABLE ADD CONSTRAINT`, which is what lets this table carry its full constraint set on SQLite where altering an existing table cannot.
- **`context.as_sql` fork**: offline mode has no live connection to reflect from, so the inspector-backed idempotency guards run only online; the offline branch emits unconditional DDL, following the shape already used by neighbouring revisions.
- **Downgrade is a single `drop_table`** — constraints go with the table. The name contract between model and migration is enforced by the parity suite, not by convention.

## What parity compares, and why names are not enough

The parity suite reflects the full inventory on the same backend from both construction paths and diffs: columns (type, nullability, server default), CHECK constraints by name **and expression**, primary key, unique constraints, indexes including column order, and foreign keys including `ondelete`. The expression dimension exists because a predicate can drift while its name stays put — weakening the single-active-slot check from `= 1` to `>= 1` changes no constraint name and would pass a name-only comparison, but it dismantles the invariant the table exists to enforce. The suite catches that edit, a dropped server default, a deleted primary key (which on PostgreSQL also silently costs the SERIAL sequence), and a reordered index.

The SQLite migrated shape is built by a filtered `create_all` (everything except this table) followed by an upgrade of only this revision; comparisons are same-backend only, since PostgreSQL rewrites reflected CHECK expressions into its own normal form.

## Known-red ledger

| Case | Before | After | Failure signature |
|---|---|---|---|
| `test_postgresql_upgrade` | FAIL | FAIL | `AssertionError: models table should exist` |
| `test_postgresql_idempotence_with_sqlalchemy` | FAIL | FAIL | `DependentObjectsStillExist: cannot drop table mcp_servers_legacy …` |
| remaining 5 cases | PASS | PASS | — |

Both failures predate this PR and are unrelated to `task_interaction_requests` (one is a historical gap around the `models` table, the other a teardown `DROP TABLE` colliding with four pre-existing foreign keys). Verified by a same-session A/B run — the suite executed with this revision held out and then present, identical methodology and fresh one-time databases both times — whose two grepped failure captures are byte-identical (the catalog scan order behind the assertion text is not stable across sessions, so cross-session text comparison would be meaningless). The upgrade log of the B run shows this revision executing to completion inside the PostgreSQL chain. This PR does not attempt to fix those two failures.

## Scope

This PR does not close #1074. Neither does the series it belongs to: the existing-table field pairs were split out after the constraint-versus-offline-SQL conflict was measured, and the datetime read-back half needs a first reader to demonstrate it. The delivery boundary and the reasoning are recorded on #1074.

## Size

Most of the diff is test-side. The parity suite compares six dimensions (columns with server defaults, CHECK names and expressions, primary key, uniques, indexes including column order, foreign keys) because review found mutations slipping through narrower comparisons, and PostgreSQL online coverage (upgrade/downgrade, both idempotency directions, residue across tables, sequences, and indexes) is committed rather than left to hand verification. The migration file itself is a single create_table mirroring the model.


tests/migrations/test_task_interaction_requests_schema_parity.py
```diff
@@ -0,0 +1,296 @@
...
+    finally:
+        for _dbname, engine in minted:
+            engine.dispose()
+        for dbname, _engine in minted:
+            admin_conn = _admin_connection()
+            try:
+                admin_conn.cursor().execute(f'DROP DATABASE IF EXISTS "{dbname}"')
+            finally:
+                admin_conn.close()
+
```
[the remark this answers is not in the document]
AlexLiu190625: Confirmed. This test calls the factory twice, so the fixture holds two databases, and a failure on the first drop leaves the second on the server. Reproduced directly: with an outside session still attached to the first database, `DROP DATABASE` raises `ObjectInUse` (`IF EXISTS` does not help there), the exception leaves the `finally` block, and the second database survives.

Not taking the suggested `except Exception: pass`, though: swallowing silently means the database that failed to drop is never named, and someone has to drop it by hand later — the name is the only thing that makes that possible. The fix isolates each drop and reports what survived: each `engine.dispose()` in its own `try` (a failure there previously leaked everything, which the suggestion's shape would not have covered); one autocommit admin connection for the whole drop loop — verified that a failed `DROP DATABASE` leaves such a connection usable, so the next drop still runs; an immediate failure naming every database if the server cannot be reached at all; otherwise every drop attempted, failures collected, and one error at the end listing the exact names left behind. A database still in use costs its own name, not the rest of the list.

tests/alembic/test_20260809_add_task_interaction_requests.py
```diff
@@ -0,0 +1,471 @@
...
+        yield _make
+    finally:
+        for _dbname, engine in minted:
+            engine.dispose()
+        for dbname, _engine in minted:
+            admin_conn = _admin_connection()
+            try:
+                admin_conn.cursor().execute(f'DROP DATABASE IF EXISTS "{dbname}"')
+            finally:
+                admin_conn.close()
```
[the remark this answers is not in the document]
AlexLiu190625: Partly right. All four tests using this fixture mint exactly one database each, so the list never holds more than one item and there is no second database to leak — the cascade described is not reachable here today. And with a single item, `except Exception: pass` would not save anything: the drop still fails and the database still leaks, the only difference being that nobody is told which one.

Still worth fixing, because the fixture is written as "drops every database it created" and a future test minting two would make the leak real with no warning. Applied the same teardown as in the parity suite: per-item isolation for both the dispose loop and the drop loop, one admin connection for the drops, an immediate failure naming every database if the server is unreachable, and a final error listing the exact names of anything that could not be dropped. Green runs are unaffected.

tests/web/services/task_interaction_schema_shared.py
```diff
@@ -428,3 +430,163 @@ def assert_update_rejected(
...
+    so ``ix_task_interaction_requests_task_status`` declared as
+    ``(task_id, status)`` and one declared as ``(status, task_id)`` are
+    different objects a parity check must tell apart. UNIQUE constraints
+    are the deliberate exception -- uniqueness is enforced over the column
+    set regardless of declaration order, so ``unique`` below still sorts
+    column names; that sort is intentional, not an oversight, and must stay
+    paired with the comment on ``EXPECTED_UNIQUE_CONSTRAINTS`` above making
+    the same claim for the hand-written literal.
+    """
+    inspector = sa.inspect(engine)
```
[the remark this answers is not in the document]
AlexLiu190625: This helper only ever runs after a fixture has just built the table — the three call sites in the parity suite are each preceded by `_build_create_all_schema` or `_build_migration_schema`, and nothing else in the repo calls it. There is no caller for which an absent table is a normal state, so absence here means the fixture is broken, which is exactly what the exception should say.

Returning an empty inventory instead weakens the comparison the file exists for. Verified with two mutations. Gutting the migration's `create_table` leaves the suite red either way, since one side still has 23 checks and the other has none. But when both sides come back empty — a wrong database, or a create_all path that silently skipped the table — `assert not diff` passes on `{} == {}`, and the only thing left standing is the trailing cardinality assertion, which fails as `assert 0 == 23` and points at nothing. Today that same case raises `NoSuchTableError: task_interaction_requests` before any comparison runs.

Trading a diagnostic that names the missing table for one that reports a count mismatch is a step backwards, so keeping the raise.

src/xagent/migrations/versions/20260809_add_task_interaction_requests.py
```diff
@@ -0,0 +1,266 @@
...
+        _create_table()
+        return
+
+    inspector = sa.inspect(op.get_bind())
+    tables = set(inspector.get_table_names())
+
+    # Guard 1: a create_all-first startup already has this table: makes
+    # `alembic upgrade head` idempotent against it.
+    if TABLE in tables:
+        return
```
rogercloud, line 295: The parent-table guard calls `inspector.get_table_names()` with no `schema=` argument, unlike the two immediately preceding revisions in this same chain (`20260726_add_task_telegram_user_id.py`, `20260808_add_task_lease_attempt_id.py`), which both resolve the target schema explicitly before any schema-aware reflection — specifically to handle non-default PostgreSQL `search_path` configurations. On such a deployment, `tasks`/`trace_events`/`users` would be reported as absent, the guard would silently skip creation, and Alembic would still stamp the revision as applied, permanently leaving the database without the table and with no error. Suggest resolving the target schema the same way the sibling migrations do and passing it as `get_table_names(schema=...)`.

tests/alembic/test_20260809_add_task_interaction_requests.py
```diff
@@ -0,0 +1,500 @@
...
+        _create_parent_tables(connection)
+        operations = _operations(connection)
+
+        with Operations.context(operations.get_context()):
+            migration.upgrade()
+
+        assert TABLE in sa.inspect(connection).get_table_names()
+
+
+def test_upgrade_skips_when_the_table_already_exists() -> None:
```
rogercloud: The docstring describes this as "a positive control for guard 1: create the table by another route first (mirroring a create_all-first startup)", but the test never calls `create_all` — it calls `migration.upgrade()` twice, which makes it a near-duplicate of `test_online_upgrade_is_idempotent` (lines 276-291) and means it does not exercise the scenario it claims. Suggest building the table via `Base.metadata.create_all(...)` before a single `upgrade()` call, or removing the test as redundant.

tests/alembic/test_20260809_add_task_interaction_requests.py
```diff
@@ -0,0 +1,500 @@
...
+    """Create the minimal parent tables the migration's foreign keys need.
+
+    Each parent only needs the single `id INTEGER PRIMARY KEY` column the
+    FKs reference; the real tasks/trace_events/users tables carry many more
+    columns this migration never touches.
+    """
+    for table in tables:
+        connection.exec_driver_sql(f"CREATE TABLE {table} (id INTEGER PRIMARY KEY)")
+
+
```
rogercloud, line 81: `_psycopg2_kwargs` is byte-for-byte duplicated with `tests/migrations/test_task_interaction_requests_schema_parity.py:124-224`, and `postgresql_engine_factory` is ~75 of ~90 lines verbatim duplicated there too (only the minted-database name prefix and docstring wording differ). Separately, this fixture's docstring claims the duplication follows repo precedent, citing `test_task_interaction_schema_postgresql.py`, `test_task_status_storage_postgresql.py`, and `test_runtime_key_transition_postgres.py` — but all three use a materially different pattern (a single shared named database with `drop_all`/`create_all`, no `psycopg2` import, no `CREATE DATABASE`), so the disposable-database pattern is new here rather than precedented. Suggest extracting the shared fixture logic into one module (the repo already does this for shared test helpers, e.g. `tests/web/services/task_interaction_schema_shared.py`) and correcting or removing the precedent claim.

tests/web/services/task_interaction_schema_shared.py
```diff
@@ -428,3 +430,163 @@ def assert_update_rejected(
...
+    # PostgreSQL's get_indexes() also reports the backing index for each
+    # UNIQUE constraint (unique=True); SQLite's does not (see
+    # EXPECTED_NONUNIQUE_INDEXES above). Grouping by the unique flag first
+    # keeps a name collision on the unique-backed side from being misread
+    # as a problem with the plain indexes. Column order is preserved, not
+    # sorted -- see this function's docstring.
+    indexes: dict[str, dict[str, tuple[str, ...]]] = {"unique": {}, "nonunique": {}}
+    for idx in inspector.get_indexes(TABLE_NAME):
+        bucket = "unique" if idx["unique"] else "nonunique"
+        indexes[bucket][idx["name"]] = tuple(idx["column_names"])
```
rogercloud, line 565: The parity comparison stores `columns` as a name-keyed dict with no ordinal/position field, so column declaration order drift between the migration and the model is invisible to the suite — swapping two adjacent column declarations in the migration leaves all 17 tests green (reproduced twice). The impact is narrow (`SELECT *` ordering, column-list-less `INSERT`s), but the suite's "full inventory" wording oversells current coverage. Consider including ordinal position in the compared inventory if column order is meant to be part of the pinned contract.

tests/alembic/test_20260809_add_task_interaction_requests.py
```diff
@@ -0,0 +1,500 @@
...
+                    try:
+                        admin_conn.cursor().execute(
+                            f'DROP DATABASE IF EXISTS "{dbname}"'
+                        )
+                    except Exception as exc:  # noqa: BLE001
+                        undropped.append(f"{dbname} ({exc!r})")
+            finally:
+                admin_conn.close()
+
+        if undropped:
```
rogercloud: Teardown raises `RuntimeError` for undropped PostgreSQL databases, which pytest attributes to whichever test happens to run last in this fixture's scope rather than surfacing it as independent infrastructure noise; and the bare `print()` for engine-disposal failures (line 169) is swallowed by pytest's stdout capture on a passing run, so it is never seen. Same pattern at `tests/migrations/test_task_interaction_requests_schema_parity.py:194,220-224`. Consider `warnings.warn` or a session-scoped finalizer so cleanup failures are attributed correctly and stay visible.

tests/alembic/test_20260809_add_task_interaction_requests.py
```diff
@@ -0,0 +1,500 @@
...
+    spec = importlib.util.spec_from_file_location(
+        "add_task_interaction_requests_migration", MIGRATION_PATH
+    )
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def _operations(connection) -> Operations:
```
rogercloud: `_operations(connection)` builds an `Operations` wrapper, but all 11 call sites only ever call `.get_context()` on it to retrieve the underlying `MigrationContext` — no call site uses an `Operations`-specific method. `MigrationContext.configure(connection)` (already called inside `_operations()`) returns the same object directly and avoids the extra wrapper allocation at every call site.

tests/migrations/test_task_interaction_requests_schema_parity.py
```diff
@@ -0,0 +1,325 @@
...
+    migration_engine = sa.create_engine(f"sqlite:///{tmp_path / 'migration.db'}")
+
+    _build_create_all_schema(create_all_engine)
+    _build_migration_schema(migration_engine)
+
+    left = reflect_full_inventory(create_all_engine)
+    right = reflect_full_inventory(migration_engine)
+    diff = diff_full_inventory(left, right)
+    assert not diff, diff
+    assert len(left["checks"]) == 23
```
rogercloud: The expected CHECK-constraint count `23` is hardcoded four times (here, line 255, and in the migration's `CHECKS` tuple / `EXPECTED_CHECK_CONSTRAINT_NAMES` in `tests/web/services/task_interaction_schema_shared.py`), giving three independent inventories of "which/how many CHECK constraints exist" that must all move together when a 24th is added. Suggest deriving the count from one of the existing name collections instead of a bare literal.

tests/migrations/test_task_interaction_requests_schema_parity.py
```diff
@@ -0,0 +1,325 @@
+"""create_all/migration schema parity for task_interaction_requests.
+
+The model (src/xagent/web/models/task_interaction.py) and the
+20260809_add_task_interaction_requests migration are two independent
+implementations of the same constraint inventory: create_all builds the
+table straight from the model, the migration builds it from a hand-copied
+CHECKS tuple. Nothing enforces they stay identical except this suite.
+
+Migration-side construction (candidate "乙" from the PR-B2 task book's
```
rogercloud: This docstring (and the one at lines 81-82) cites an internal citation naming a nonexistent "PR-B2 task book" and quoting a Chinese-language placeholder term, an artifact that does not exist anywhere in the repo (confirmed by repo-wide grep), and it is the only CJK fragment in a non-test-data docstring in this part of the tree. The substantive rationale that follows reads fine without it — suggest dropping the citation.

tests/migrations/test_task_interaction_requests_schema_parity.py
```diff
@@ -0,0 +1,325 @@
...
+    _build_migration_schema(engine)
+    tables_with = set(sa.inspect(engine).get_table_names())
+    assert tables_with == tables_before | {TABLE}
+
+    _downgrade(engine)
+    tables_after = set(sa.inspect(engine).get_table_names())
+    assert tables_after == tables_before
+
+
+def test_expression_dimension_is_actually_compared() -> None:
```
rogercloud: `test_expression_dimension_is_actually_compared` only unit-tests the diff function against two hand-written dicts; it never calls `reflect_full_inventory` against a database, so it would not catch a bug on the reflection side. End-to-end drift detection actually comes from `test_sqlite_migration_schema_matches_create_all` (verified by mutation: a weakened CHECK expression is caught there). The test has standalone value isolating differ bugs from reflection bugs — just consider renaming/rewording it so the docstring reflects that narrower scope.

rogercloud (approved):
## Summary

This PR adds the single Alembic revision that closes a real gap: `TaskInteractionRequest` landed in #1209, so fresh `create_all` databases already have the table while databases maintained through the revision chain do not. The revision (`src/xagent/migrations/versions/20260809_add_task_interaction_requests.py`) is one `op.create_table` with the full 25-column / 23-CHECK / 2-UNIQUE / 3-FK / 2-index set and zero `ALTER TABLE`, so SQLite can carry every constraint; it forks on `context.as_sql` so offline SQL generation never touches a live connection, and guards on both "table already exists" and "parent tables missing". Alongside it come ~1250 lines of tests: migration-behavior tests (upgrade/downgrade, offline SQL across all four dialect x direction cells, idempotency in both directions, guard behavior) and a schema-parity suite that reflects the `create_all`-built and migration-built schemas on the same live database and diffs them across six dimensions.

## Approach verdict: sound

The shape — a single `create_table`, a single `drop_table`, an `as_sql` fork, and parity tests as the contract rather than deriving DDL from the model — is the simplest thing that fits the problem, and it matches the conventions of the immediately adjacent revisions (`20260711_add_task_execution_commands.py`, `20260726_add_task_telegram_user_id.py`, `20260808_add_task_lease_attempt_id.py`). The "zero `ALTER TABLE`" and "guards only run online" properties were checked against the actually generated SQL rather than taken from the description. The parity suite is what makes the hand-maintained `CHECKS` tuple safe, and it holds up: of eight injected schema mutations (weakened CHECK expression, index column swap, dropped server default, nullable flip, changed `ondelete`, removed guard, changed VARCHAR length, removed offline fork), seven were caught — the single survivor is the column-ordering gap noted below. No blocking issues.

The two prior bot findings are already settled and are not re-raised here: the PostgreSQL teardown leak is genuinely fixed by `f1fcc562` (per-item isolated dispose, one shared admin connection, and a final error naming exactly which databases survived — no failure path is swallowed), and the `NoSuchTableError`-vs-empty-inventory suggestion was correctly waived, since returning `{}` would let a "both sides empty" fixture bug pass `assert not diff` silently.

## Findings

### Minor

- **`src/xagent/migrations/versions/20260809_add_task_interaction_requests.py:244-257`** — The parent-table guard calls `inspector.get_table_names()` with no `schema=` argument, unlike the two immediately preceding revisions in this same chain (`20260726_add_task_telegram_user_id.py`, `20260808_add_task_lease_attempt_id.py`), which both resolve the target schema explicitly before any schema-aware reflection, specifically for non-default PostgreSQL `search_path` setups; on such a deployment the guard would see `tasks`/`trace_events`/`users` as absent, silently skip creation, and still let Alembic stamp the revision as applied. Resolve the schema the way the sibling migrations do and pass it as `get_table_names(schema=...)`.
- **`tests/alembic/test_20260809_add_task_interaction_requests.py:344-365`** — `test_upgrade_skips_when_the_table_already_exists` documents itself as "a positive control for guard 1: create the table by another route first (mirroring a create_all-first startup)", but it never calls `create_all` — it calls `migration.upgrade()` twice, making it a near-duplicate of `test_online_upgrade_is_idempotent` (lines 276-291). Either build the table via `Base.metadata.create_all(...)` before the single `upgrade()` so it tests the scenario it claims, or drop it as redundant.
- **`tests/alembic/test_20260809_add_task_interaction_requests.py:99-199` and `tests/migrations/test_task_interaction_requests_schema_parity.py:124-224`** — `_psycopg2_kwargs` is byte-for-byte duplicated and `postgresql_engine_factory` is ~75 of ~90 lines verbatim duplicated between the two files, and the alembic-side docstring justifies this by citing precedent in `test_task_interaction_schema_postgresql.py` / `test_task_status_storage_postgresql.py` / `test_runtime_key_transition_postgres.py` — but all three of those use a materially different pattern (one shared named database with `drop_all`/`create_all`, no `psycopg2` import, no `CREATE DATABASE`), so the disposable-database pattern is genuinely new here rather than precedented. Extract the shared fixture logic into one module (the repo already does this for shared test helpers, e.g. `tests/web/services/task_interaction_schema_shared.py`) and correct or drop the precedent claim.
- **`tests/web/services/task_interaction_schema_shared.py:534-541`** — The parity comparison keys `columns` by name with no ordinal/position field, so column declaration order drift between migration and model goes undetected (reproduced twice: swapping two adjacent column declarations in the migration leaves all 17 tests green), which matters for `SELECT *` ordering and column-list-less `INSERT`s and is oversold by the suite's "full inventory" wording. Add ordinal position to the compared inventory if column order is meant to be part of the pinned contract.
- **`tests/alembic/test_20260809_add_task_interaction_requests.py:169,195-199` and `tests/migrations/test_task_interaction_requests_schema_parity.py:194,220-224`** — Teardown raises `RuntimeError` for undropped PostgreSQL databases, which pytest attributes to whichever test happened to run last rather than reporting it as infrastructure noise, while the bare `print()` on engine-disposal failure is swallowed by stdout capture on a green run. Prefer `warnings.warn` (or a session-scoped finalizer) so cleanup failures are attributed and visible independently of test outcomes.

### Nits

- **`tests/migrations/test_task_interaction_requests_schema_parity.py:238-239,255-256`** (together with the migration's `CHECKS` tuple and `EXPECTED_CHECK_CONSTRAINT_NAMES` in `tests/web/services/task_interaction_schema_shared.py`) — The literal `23` appears four times with no named constant, and three separate hardcoded inventories of "which/how many CHECK constraints exist" must all be updated together when a 24th CHECK is added. Derive the count from one of the existing name collections instead of hardcoding it.
- **`tests/migrations/test_task_interaction_requests_schema_parity.py:9-10,81-82`** — The docstrings cite an internal artifact (an internal citation naming a nonexistent "PR-B2 task book" and quoting a Chinese-language placeholder term) that does not exist anywhere in the repo, and it is the only CJK fragment in a non-test-data docstring in this area. Remove the dead citation; the rationale that follows already stands on its own.
- **`tests/migrations/test_task_interaction_requests_schema_parity.py:298-325`** — `test_expression_dimension_is_actually_compared` only exercises the diff function against two hand-written dicts and never calls `reflect_full_inventory` against a database, so its name/docstring overstate its scope (real end-to-end drift detection comes from `test_sqlite_migration_schema_matches_create_all`). Keep the test — isolating differ bugs from reflection bugs has value — but reword it to say it tests the differ in isolation.
- **Test coverage organization** — This is the only revision in the repo whose tests are split across both `tests/alembic/` and `tests/migrations/`, and both new files independently define a byte-identical `_load_migration_module()`. The behavior-vs-parity split is a defensible rationale, but the loader duplication is not; extract `_load_migration_module()` into a shared helper.

### Informational / non-blocking

- **`src/xagent/migrations/versions/20260809_add_task_interaction_requests.py:262-266`** — `downgrade()` unconditionally drops the table whenever it exists, without checking whether this revision's `upgrade()` is what created it, so a `create_all`-provisioned table (where guard 1 made `upgrade()` a no-op) is still dropped along with its PostgreSQL sequence. This state-based rather than provenance-based idempotency is exactly what `20260726` and `20260808` already do for their column add/drop, so it is not a defect in this PR; a docstring note or regression test would round it out.
- **`src/xagent/migrations/versions/20260809_add_task_interaction_requests.py:244-259`** — Both guards `return` silently, so `alembic upgrade head` can report success while stamping a revision that created nothing, with no diagnostic. This is a repo-wide convention (identical in `20260711_add_task_execution_commands.py` and many others) that this PR follows rather than introduces; adding logging to all schema-existence guards is worth a separate follow-up, not a change to this PR.
- Five new PostgreSQL-only tests (`tests/alembic/test_20260809_add_task_interaction_requests.py:378,413,431,447` and `tests/migrations/test_task_interaction_requests_schema_parity.py:242`) lack `@pytest.mark.postgresql`. No CI job runs `-m postgresql` over these directories today and several pre-existing files also omit it, so this is cosmetic consistency only.
- The offline-downgrade SQL test (`tests/alembic/test_20260809_add_task_interaction_requests.py:241-247`) omits the `"ALTER TABLE" not in sql` assertion its upgrade counterpart has; low value, since `op.drop_table` cannot regress into emitting `ALTER TABLE`.
- The parent-table-missing guard is exercised only on SQLite even though its documented rationale is PostgreSQL-specific (PG raises `UndefinedTable` at CREATE time; SQLite only fails later at INSERT). Low priority — the guard itself is backend-agnostic Python with no dialect branching.
- Unrelated to this PR: `mut_empty.py` at the repo root is a leftover from #1204 (`f0dfc3fd`) and is not part of this diff — noting it only so it is not misattributed here.

Also checked and found to be non-issues: `sa.JSON(none_as_null=True)` in the migration (a correct verbatim mirror of the model's column type), guard 1 not validating full schema parity (inherent to the guard idiom used throughout this directory), the placement of the shared parity helper module (it is imported from within its own directory too), inlining `_normalize_default` (its docstring carries non-obvious cross-dialect rationale), and the long module docstring (the migration, model, and test docstrings cover genuinely different ground).

`tests/alembic/test_20260809_add_task_interaction_requests.py:72`: shrink: `_operations(connection)` constructs an `Operations` wrapper solely to call `.get_context()` on it at all 11 call sites; `MigrationContext.configure(connection)` gets the same result directly.

`tests/alembic/test_20260809_add_task_interaction_requests.py` + `tests/migrations/test_task_interaction_requests_schema_parity.py`: shrink: ~75-90 lines of PostgreSQL disposable-database fixture logic (`_psycopg2_kwargs`, `postgresql_engine_factory`) duplicated verbatim between the two files; extract into one shared module.

net: -~100 lines possible

## Test results

Run locally against the PR branch: **17 passed, 5 skipped, 0 failed**. All five skips are PostgreSQL-only tests gated on `XAGENT_TEST_POSTGRES_URL`, which is not set in this environment. The two known-red PostgreSQL tests (`test_postgresql_upgrade`, `test_postgresql_idempotence_with_sqlalchemy`) are pre-existing failures unrelated to this table, confirmed by the author's A/B run with the revision held out vs. present, and are correctly left out of scope here.

## Recommendation

Approve — no blocking issues; the schema-guard schema-argument inconsistency and the mislabeled guard-1 test are the two worth addressing, the rest are optional follow-ups.

src/xagent/migrations/versions/20260809_add_task_interaction_requests.py
AlexLiu190625: Adopting the change, with a correction to the reasoning. SQLAlchemy's PostgreSQL reflection filters `get_table_names(schema=None)` on `pg_table_is_visible()`, not `current_schema()`, so with `search_path = app, public` and the parents in `public` the guard still sees all three — measured on PostgreSQL 16. The parents only read as absent when they are off `search_path` entirely, and such a database cannot resolve `tasks` for the application either.

There is a real failure here though, and it is guard 1 rather than guard 2: with the live schema `app`, a stale `public.task_interaction_requests`, and `search_path = app, public`, the visible stale table makes `upgrade()` a no-op while the live schema never gets the table and the revision is stamped. Reproduced on a real server.

The fix is a little wider than the suggestion. The schema has to be resolved from a parent relation, not the target table — which does not exist yet on the run that creates it — and it has to be passed to `op.create_table` / `op.create_index` / `op.drop_table` as well, or reflection and DDL address different schemas. `downgrade()` had the same unqualified reflection and gets the same treatment. Landed with a regression test that builds exactly the stale-shadow shape and asserts the table is created in the live schema with its foreign keys bound to that schema's parents.

tests/alembic/test_20260809_add_task_interaction_requests.py
AlexLiu190625: Correct, and fixed rather than dropped. The model ships ahead of this revision, so `create_all`-provisioned databases carrying this table already exist, and `alembic upgrade head` reaching this revision against one of them is the real production path. The old test only proved the migration tolerates a table it built itself. It now builds the table with `Base.metadata.create_all(...)` and asserts the single `upgrade()` leaves the check-constraint set and the column list untouched.

tests/alembic/test_20260809_add_task_interaction_requests.py
AlexLiu190625: Both halves confirmed. The precedent claim was wrong — `CREATE DATABASE` appears in exactly these two files and nowhere else under `tests/`; all three cited files use one shared named database with `drop_all`/`create_all` and never import psycopg2. The claim is dropped.

On the extraction: it landed in `tests/shared/` rather than the task-interaction helper module — that directory is already where cross-directory test helpers live (`db_teardown.py`, `execution_scope.py`), and the disposable-database logic is generic infrastructure, not specific to this table. The connection kwargs, the mint/teardown body, and the migration-module loader all moved there; each test file keeps a three-line fixture naming its own database prefix.

tests/web/services/task_interaction_schema_shared.py
AlexLiu190625: Confirmed, and it survives the PostgreSQL half too — with a live server the same adjacent-column swap left all tests green, so the PostgreSQL parity test missed it as well. The inventory now carries the ordinal: `columns` is `{name: (position, type, nullable, default)}` built from `enumerate(inspector.get_columns(...))`. Both backends return columns in physical order (`attnum` on PostgreSQL, `cid` on SQLite), so the position is stable and comparable, and the differ needed no change since it compares tuples. Re-ran the swap after the fix: both backends' parity tests go red with the swapped positions named in the diff.

tests/alembic/test_20260809_add_task_interaction_requests.py
AlexLiu190625: Taking the `print` half, leaving the attribution half. The fixture is function-scoped, so the minted list resets per test and the `RuntimeError` already attaches to the exact test that leaked — verified: it reports as an error at teardown of that specific item, not whichever ran last.

The `print` was worth changing, though the window is narrower than described: pytest surfaces captured teardown stdout whenever the teardown raises. It went unseen only when disposal failed but every drop then succeeded — precisely the green run that hid a real failure. Both sites now use `warnings.warn`, which lands in the warnings summary regardless of outcome.

tests/alembic/test_20260809_add_task_interaction_requests.py
AlexLiu190625: Confirmed — `Operations(ctx).get_context() is ctx` is `True`; the wrapper was pure allocation. Removed, with one `MigrationContext.configure(connection)` hoisted per test rather than configuring inline twice, since the tests that open two `Operations.context(...)` blocks would otherwise get two distinct context objects.

tests/migrations/test_task_interaction_requests_schema_parity.py
AlexLiu190625: Agreed. All four literals were in this file; the migration and the shared module carry name tuples rather than counts. They are now `len(EXPECTED_CHECK_CONSTRAINT_NAMES)` imported from the shared module. There was also a fourth copy of the name set in the alembic-side test file — it now imports the shared constant too, which brings the hand-maintained inventories down to two (the migration's `CHECKS` and the shared names), and the parity suite already cross-checks those against live reflection.

tests/migrations/test_task_interaction_requests_schema_parity.py
AlexLiu190625: Agreed, and removed as a hygiene fix rather than a nit — this is an English-only repository and the citation pointed at nothing that exists here. Both docstrings drop the reference; the rationale that follows (why the full alembic chain and the upgraded-engine helper are both wrong for this fixture) stands unchanged. A sweep of all four touched files confirms no non-ASCII or internal-artifact references remain, including one dangling sentence the removal would otherwise have orphaned.

tests/migrations/test_task_interaction_requests_schema_parity.py
AlexLiu190625: Agreed. Renamed to `test_differ_flags_a_changed_check_expression`, with the docstring now stating it exercises `diff_full_inventory` alone against two hand-written inventories — no reflection, no database — and pointing at `test_sqlite_migration_schema_matches_create_all` for end-to-end drift detection.

AlexLiu190625:
On the body-level items:

**The five PostgreSQL-only tests without markers** — this turned out to be bigger than a cosmetic gap: those tests had no CI runner at all. The fast job deselects the `postgresql` marker and runs without a database URL, the dedicated migration workflow's PostgreSQL job runs a hand-listed file set that predates these files, and `tests/alembic/**` was not among its trigger paths. All three are fixed in this PR: the six PostgreSQL-only tests (including the new stale-shadow regression) now carry the marker, both files are on the PostgreSQL job's run list, and the trigger paths include `tests/alembic/**`. Selection verified in both directions locally.

**Downgrade provenance** — the docstring now states the drop is state-based rather than provenance-based: a `create_all`-provisioned table is dropped too, matching the sibling column migrations.

**Guard logging** — leaving as-is per the repo-wide convention; a logging pass over all schema-existence guards belongs to its own change.

**The offline-downgrade ALTER assertion** — declined: `op.drop_table` cannot emit `ALTER TABLE`, so the assertion would pin nothing.

**`mut_empty.py`** — thanks for flagging it. Verified: it is a gutted copy of the 20260804 revision committed by accident, present on `main` since that change merged, unreferenced by pytest (which collects only `tests/`) and invisible to Alembic (which reads only the versions directory). Removed here as a standalone commit — noting you fenced it out of this PR's attribution, which is correct; folding the deletion in was a scope call to save a one-line PR.

PR #1219, opened 2026-08-09
fix: authorize task Agent identity before SSH resolution

### LINKED ISSUE #1136: SSH task Agent identity reaches provider target resolution without owner/team authorization
## Problem

The SSH tool adapter treats a task's Agent reference as SSH binding authority without re-authorizing that Agent against the task owner.

`_agent_id_from_task` in `src/xagent/core/tools/adapters/vibe/ssh_tools.py` currently returns either persisted identity candidate directly:

```python
if task.agent_id is not None:
    return int(task.agent_id)
cfg = task.agent_config or {}
pid = cfg.get("preview_agent_id") if isinstance(cfg, dict) else None
return int(pid) if pid is not None else None
```

`_agent_id_for_task` loads the task by task ID and then calls this helper. It does not verify that the selected Agent is owned by, or shared with, `task.user_id`.

Both candidates need the SSH-specific check:

- `Task.agent_id` is persisted from a task-create request after Agent execution authorization. That path may allow an accessible published Agent outside the task owner's own/team management scope. Permission to execute a published Agent does not imply permission to use the Agent owner's private SSH bindings.
- Agent-builder preview tasks can instead carry client-writable `agent_config.preview_agent_id` while leaving `Task.agent_id` unset. Other readers of that field in `src/xagent/web/api/chat.py` and `src/xagent/web/services/task_setup_snapshot.py` already apply `owned_agent_clause(..., get_agent_team_scope(...))` before use.

## Reachability

- Both the current task-create path and the WebSocket preview compatibility path can persist the preview task shape.
- `sanitize_client_agent_config` deliberately preserves `preview_agent_id`. Reserving or stripping it at the generic task-create boundary would also remove the legitimate builder-preview value.
- `ssh` is a client-assignable tool category, so the SSH creator can run for either task identity shape.
- The selected ID is copied into `SshExecutionContext.agent_id` and passed to `SshTargetProvider.list_bound_targets`. The same context is later used by `SshTargetProvider.resolve` when an SSH operation executes.

A provider may use `context.agent_id` as its binding key without independently repeating owner/team authorization. The `SshTargetProvider` protocol does not require that second check, so Core must not present an unauthorized task Agent identity to provider target methods.

## Scope clarification

This issue is about SSH target selection and resolution. It does **not** claim that the selected Agent ID controls the sandbox lease identity in the current task-backed production path.

`_make_ssh_sandbox_lease` has an `agent-{agent_id}` fallback when no task ID is supplied, but `create_ssh_tools` cannot reach that branch with an Agent ID resolved from a task: without a numeric task ID, `_agent_id_for_task` returns `None` and SSH tool construction stops. Normal preview tasks supply a `web_task_<id>` identity, so their lease is task-scoped.

## Fix boundary

Re-authorize both task Agent identity candidates at the reader that owns the database session:

- Preserve `Task.agent_id` precedence, but resolve it through the task owner's own/team Agent scope before SSH use.
- When `Task.agent_id` is absent, normalize and resolve `preview_agent_id` through the same scope.
- Keep normalization and owner/team authorization in one shared owner reused by the existing preview readers; do not add a fourth copy of the query.
- Reject malformed, missing, nonexistent, out-of-scope, or archived candidates before calling provider target methods.
- Provider-side authorization remains useful defense in depth, but it must not be the only boundary because providers are supplied through an extension hook.

The provider factory may still run as the feature-presence/structural gate because it receives no Agent ID. The security boundary is that denied identities never reach `list_bound_targets`, `resolve`, or another provider target method.

## Acceptance criteria

- Neither a non-null `Task.agent_id` nor a fallback `preview_agent_id` outside `task.user_id`'s permitted own/team Agent scope can reach provider target methods.
- Executing an otherwise accessible published Agent does not confer access to that Agent owner's SSH bindings.
- Malformed, missing, nonexistent, out-of-scope, or archived candidates produce no SSH tools and zero `list_bound_targets`/`resolve` calls.
- A legitimate builder preview using a draft or published Agent owned by, or shared with, the task owner continues to construct and execute SSH tools.
- Regression coverage exercises `Task -> _agent_id_for_task -> create_ssh_tools`, not only a pure parsing helper.
- Negative coverage uses a task owner and Agent owner from different scopes and proves that provider target methods are not called.
- Task-create and WebSocket preview producer shapes remain covered so the legitimate preview contract cannot drift.

## Context

Found while reviewing #1118, which reserves other client-controlled keys in `agent_config`. `preview_agent_id` cannot use the same blanket reservation because preview itself needs the field; it requires a read-side authorization boundary.

AlexLiu190625:
## Summary

- authorize persisted SSH task Agent identifiers against the task owner before target discovery
- reuse the same owner/team authorization rule for preview Agent readers
- reject malformed or out-of-range Agent identifiers before database conversion
- preserve direct-Agent precedence, DRAFT/PUBLISHED compatibility, and ARCHIVED denial

## Root cause

SSH task construction accepted a persisted direct Agent ID or preview Agent ID as execution context without revalidating that the persisted task owner could still access that Agent. A stale, cross-owner, or otherwise invalid identifier could therefore reach a downstream SSH provider's target-resolution boundary.

## Security and compatibility

- denied Agent candidates return no SSH tools and call no provider target method
- the provider factory remains the feature-availability gate and retains its existing ordering
- task owner identity remains distinct from the authorized Agent identity
- no schema, protocol, producer payload, or public error-code change

## Verification

- focused Core authorization, SSH, runtime, and snapshot suites passed
- Agent Builder preview suite: 5 passed
- Ruff, formatting, pre-commit, and diff checks passed
- negative-path and positive production-path regressions were mutation-checked

Closes #1136

### DIFF
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -68,7 +68,11 @@ from ..sandbox_keys import (
 )
 from ..schemas.chat import TaskCreateRequest, TaskCreateResponse
 from ..services.agent_access import list_accessible_published_agents
-from ..services.agent_team_scope import get_agent_team_scope, owned_agent_clause
+from ..services.agent_team_scope import (
+    get_agent_team_scope,
+    owned_agent_clause,
+    resolve_authorized_agent,
+)
 from ..services.chat_history_service import (
     get_latest_waiting_question,
     load_task_transcript,
@@ -1919,16 +1923,10 @@ class AgentServiceManager:
                 )
             elif agent_config and agent_config.get("preview_agent_id"):
                 preview_user_id = int(task.user_id)
-                current_agent = (
-                    db.query(Agent)
-                    .filter(
-                        Agent.id == agent_config["preview_agent_id"],
-                        owned_agent_clause(
-                            preview_user_id,
-                            get_agent_team_scope(db, preview_user_id),
-                        ),
-                    )
-                    .first()
+                current_agent = resolve_authorized_agent(
+                    db,
+                    preview_user_id,
+                    agent_config.get("preview_agent_id"),
                 )
                 if current_agent and current_agent.status == AgentStatus.PUBLISHED:
                     excluded_agent_id = int(current_agent.id)
@@ -2560,24 +2558,16 @@ class AgentServiceManager:
                     and agent_config
                     and agent_config.get("preview_agent_id")
                 ):
-                    from ..models.agent import AgentStatus
-
                     if task is None:
                         raise ValueError(
                             f"Task {task_id} missing while resolving preview agent"
                         )
                     preview_user_id = int(task.user_id)
                     assert db is not None
-                    current_agent = (
-                        db.query(Agent)
-                        .filter(
-                            Agent.id == agent_config["preview_agent_id"],
-                            owned_agent_clause(
-                                preview_user_id,
-                                get_agent_team_scope(db, preview_user_id),
-                            ),
-                        )
-                        .first()
+                    current_agent = resolve_authorized_agent(
+                        db,
+                        preview_user_id,
+                        agent_config.get("preview_agent_id"),
                     )
                     if current_agent and current_agent.status == AgentStatus.PUBLISHED:
                         excluded_agent_id = int(current_agent.id)
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -2544,14 +2544,9 @@ class AgentServiceManager:
                         "will exclude from agent tools"
                     )
 
-                # Inline preview_agent_id case (#459 build-preview): the
-                # snapshot path doesn't cover this because it resolves
-                # excluded_agent_id only through ``task.agent_id``; the
-                # preview agent is referenced inside the inline
-                # ``agent_config`` dict on tasks with ``agent_id=None``.
-                # Run this in addition to (not instead of) the snapshot
-                # value above -- they're mutually exclusive by design
-                # (either task has an agent_id OR has inline preview).
+                # This inline preview_agent_id branch is only for the
+                # live-session path where ``snapshot is None``. The snapshot
+                # loader resolves the same inline config before this point.
                 if (
                     excluded_agent_id is None
                     and snapshot is None
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -2544,32 +2544,6 @@ class AgentServiceManager:
                         "will exclude from agent tools"
                     )
 
-                # This inline preview_agent_id branch is only for the
-                # live-session path where ``snapshot is None``. The snapshot
-                # loader resolves the same inline config before this point.
-                if (
-                    excluded_agent_id is None
-                    and snapshot is None
-                    and agent_config
-                    and agent_config.get("preview_agent_id")
-                ):
-                    if task is None:
-                        raise ValueError(
-                            f"Task {task_id} missing while resolving preview agent"
-                        )
-                    preview_user_id = int(task.user_id)
-                    assert db is not None
-                    current_agent = resolve_authorized_agent(
-                        db,
-                        preview_user_id,
-                        agent_config.get("preview_agent_id"),
-                    )
-                    if current_agent and current_agent.status == AgentStatus.PUBLISHED:
-                        excluded_agent_id = int(current_agent.id)
-                        logger.info(
-                            f"Preview task {task_id} is for published agent {current_agent.id} ({current_agent.name}), will exclude from agent tools"
-                        )
-
                 workforce_runtime = (
                     snapshot.workforce_runtime
                     if snapshot is not None
--- a/src/xagent/web/services/agent_team_scope.py
+++ b/src/xagent/web/services/agent_team_scope.py
@@ -34,8 +34,8 @@ _team_agent_connector_validator = None
 _team_agent_knowledge_base_validator = None
 
 _MAX_POSTGRES_INTEGER = 2**31 - 1
-_MAX_POSTGRES_INTEGER_DECIMAL = str(_MAX_POSTGRES_INTEGER)
-_CANONICAL_AGENT_ID = re.compile(r"[1-9][0-9]*", re.ASCII)
+_MAX_POSTGRES_INTEGER_DECIMAL_DIGITS = len(str(_MAX_POSTGRES_INTEGER))
+_CANONICAL_AGENT_ID = re.compile(r"[1-9][0-9]*")
 
 
 def set_agent_team_hooks(
@@ -144,16 +144,18 @@ def resolve_authorized_agent(
     Persisted task config is untrusted at every read boundary. Reject malformed
     values before constructing the SQL predicate, then reuse the same owner/team
     clause as all normal Agent reads.
+
+    Lifecycle status is caller-owned; each use case must apply its required status policy after resolution.
     """
     if isinstance(candidate_id, bool):
         return None
     if isinstance(candidate_id, int):
         agent_id = candidate_id
-    elif isinstance(candidate_id, str) and _CANONICAL_AGENT_ID.fullmatch(candidate_id):
-        if len(candidate_id) > len(_MAX_POSTGRES_INTEGER_DECIMAL) or (
-            len(candidate_id) == len(_MAX_POSTGRES_INTEGER_DECIMAL)
-            and candidate_id > _MAX_POSTGRES_INTEGER_DECIMAL
-        ):
+    elif isinstance(candidate_id, str):
+        # Bound untrusted text before regex scanning or integer conversion.
+        if len(candidate_id) > _MAX_POSTGRES_INTEGER_DECIMAL_DIGITS:
+            return None
+        if not _CANONICAL_AGENT_ID.fullmatch(candidate_id):
             return None
         agent_id = int(candidate_id)
     else:
--- a/src/xagent/web/services/agent_team_scope.py
+++ b/src/xagent/web/services/agent_team_scope.py
@@ -8,6 +8,7 @@ and whether that user is a team admin.
 
 from __future__ import annotations
 
+import re
 from dataclasses import dataclass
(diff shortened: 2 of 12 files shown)


src/xagent/web/services/agent_team_scope.py
```diff
@@ -129,3 +134,37 @@ def owned_agent_clause(
...
+    Persisted task config is untrusted at every read boundary. Reject malformed
+    values before constructing the SQL predicate, then reuse the same owner/team
+    clause as all normal Agent reads.
+    """
+    if isinstance(candidate_id, bool):
+        return None
+    if isinstance(candidate_id, int):
+        agent_id = candidate_id
+    elif isinstance(candidate_id, str) and _CANONICAL_AGENT_ID.fullmatch(candidate_id):
+        if len(candidate_id) > len(_MAX_POSTGRES_INTEGER_DECIMAL) or (
```
rogercloud: minor: this decimal-length pre-guard isn't only about Postgres integer bounds — under CPython 3.12's default `int_max_str_digits`, `int("1" * 5000)` raises `ValueError: Exceeds the limit (4300 digits)...`. This guard is what stops an oversized digit string from ever reaching `int()`. That matters because the call sites in `chat.py` don't uniformly catch it: line 1926 has no enclosing try/except at all, and line 2562's enclosing handler re-raises — so the ValueError would surface as an unhandled exception / HTTP 500. Worth a one-line comment noting this guard also prevents that, not just out-of-range Postgres values, so a future "simplification" doesn't drop it by accident.

src/xagent/web/services/agent_team_scope.py
```diff
@@ -129,3 +134,37 @@ def owned_agent_clause(
...
+            len(candidate_id) == len(_MAX_POSTGRES_INTEGER_DECIMAL)
+            and candidate_id > _MAX_POSTGRES_INTEGER_DECIMAL
+        ):
+            return None
+        agent_id = int(candidate_id)
+    else:
+        return None
+    if not 0 < agent_id <= _MAX_POSTGRES_INTEGER:
```
rogercloud, line 163: nitpick: `0 < agent_id <= _MAX_POSTGRES_INTEGER` is unreachable-as-a-rejection for the `str` branch — the regex plus the length/lexicographic guard above already fully bound the value before this line runs. It's live only for the `int` branch. Separately, `re.ASCII` on `re.compile(r"[1-9][0-9]*")` is a no-op (character-class ranges are unaffected by the flag). Both harmless; see the simplification note in the review body for a restructuring that removes the redundancy.

src/xagent/web/api/chat.py
```diff
@@ -2546,38 +2544,25 @@ async def _get_agent_for_task_unlocked(
...
-                        db.query(Agent)
-                        .filter(
-                            Agent.id == agent_config["preview_agent_id"],
-                            owned_agent_clause(
-                                preview_user_id,
-                                get_agent_team_scope(db, preview_user_id),
-                            ),
-                        )
-                        .first()
+                    current_agent = resolve_authorized_agent(
```
rogercloud: minor: a coverage run shows lines 2556-2569 as untested — this is the second `resolve_authorized_agent` call site, in `get_agent_for_task`'s live-session / `snapshot is None` inline-preview branch. The existing test that patches `agent_team_scope.resolve_authorized_agent` never reaches it, because `get_agent_for_task` always successfully loads a snapshot first under that test's setup, so the `snapshot is None` guard is never satisfied. A regression at this specific site (e.g. someone bypassing the authorization check here) wouldn't be caught by any current test. Suggest adding a test that exercises this branch with a denied/cross-owner agent candidate.

rogercloud (approved):
## Summary

This PR closes a real authorization gap (issue #1136): the SSH tool adapter took an Agent identity off the Task row (`Task.agent_id`, else `agent_config["preview_agent_id"]`) and passed it straight to the SSH provider as a binding key, without ever re-checking that the task owner still has access to that Agent. A stale, cross-owner, or merely-published-but-not-owned Agent ID could therefore reach SSH target resolution. The fix introduces a shared `resolve_authorized_agent` helper in `src/xagent/web/services/agent_team_scope.py` that validates the candidate ID before touching the DB and then applies the existing `owned_agent_clause` owner/team predicate, wired into `ssh_tools.py`, `chat.py` (two call sites), and `task_setup_snapshot.py`.

The only prior review on this PR (gemini-code-assist, bot) had no actionable feedback, so there is nothing to carry forward.

## Approach verdict: Sound

The fix is at the right boundary, fail-closed (a denied candidate yields no SSH tools and no provider call at all), reuses the existing `owned_agent_clause` predicate instead of a second ownership rule, and preserves the existing precedence of `Task.agent_id` over `preview_agent_id`. No correctness or security defect found in the core logic.

Design-level notes, all minor:

- **Partial consolidation.** `resolve_authorized_agent` factors out only the DB query. The full "resolve → require PUBLISHED → `int(agent.id)`" sequence still appears three times: `chat.py:1926-1935`, `chat.py:2562-2570`, and `task_setup_snapshot.py:209-227` (`_resolve_inline_preview_excluded_agent_id`, which already is that consolidation, just private and int-only). A full merge isn't a one-line import since the two `chat.py` sites also log `current_agent.name`, which the private helper doesn't expose. Worth a follow-up, not blocking.
- **Module cohesion in `agent_team_scope.py`.** This file gains its first DB query and untrusted-input parsing; sibling scope modules (`connector_team_scope.py`, `knowledge_base_team_scope.py`) are pure predicate/hook modules with no queries. A mild convention break, but three independent callers now reuse the same owner/team clause from here, so centralizing it avoids worse duplication elsewhere. Optional nit.
- **Status filtering is caller-owned but undocumented.** `resolve_authorized_agent` deliberately does not filter on `Agent.status`; each of the three callers applies its own gate today (all correct), but nothing in the helper's docstring says a future caller must add one. Suggest a one-line docstring note.
- **Capability change with no user-visible signal** (informational — intentional and disclosed in the PR body). Before this PR, a task running another user's accessible-but-not-owned published SSH-enabled agent got SSH tools; after it, it silently doesn't — only a backend `logger.info("ssh tools: skip (unresolved agent_id...)")`, nothing surfaced to the end user. The tightening itself is exactly what issue #1136 asks for. Suggested follow-up: surface some signal when SSH tooling is dropped for authorization reasons.

## Simplification opportunity

`src/xagent/web/services/agent_team_scope.py:36-38,148-162` — shrink: the manual decimal-string length/lexicographic comparison against `_MAX_POSTGRES_INTEGER_DECIMAL` (lines 153-156) is redundant with the numeric `0 < agent_id <= _MAX_POSTGRES_INTEGER` check at line 161. Replacing it with a single `if len(candidate_id) > 10: return None` before `int()` produces identical accept/reject behavior across all cases, and still guards the `int_max_str_digits` ValueError (10 digits is far below the 4300-digit limit). Removes the 4-line compound `if` and the now-unused `_MAX_POSTGRES_INTEGER_DECIMAL` constant.

net: -4 lines possible

## Test results

All 137 relevant tests pass across `tests/core/tools/test_ssh_tools.py`, `tests/web/api/test_get_agent_for_task_snapshot_threaded.py`, `tests/web/api/test_task_runtime_delete_bindings.py`, `tests/web/api/test_websocket_preview.py`, `tests/web/services/test_task_setup_snapshot.py`, and `tests/web/test_agent_team_scope.py`.

A mutation check confirms the tests are meaningful rather than incidentally green: reverting the core authorization fix produces 6+ test failures, so the suite is genuinely revert-sensitive on the behavior this PR adds. The one coverage gap is noted inline below.

## Categories with no issues

- Correctness and security logic — authorization predicate, fail-closed behavior, and `agent_id` vs. `preview_agent_id` precedence are all correct.
- Session and concurrency handling — DB sessions are scoped correctly at the reader boundary; no new shared mutable state.
- Schema, migration, and API surface — no schema/migration changes; no public API surface changes beyond the new internal helper.
- Documentation — no docs updates required beyond the docstring note suggested above.

One additional nitpick not raised inline: `src/xagent/core/tools/adapters/vibe/ssh_tools.py:300` uses a direct enum comparison (`agent.status == AgentStatus.ARCHIVED`) rather than the `getattr(agent.status, "value", agent.status)` idiom used elsewhere (`chat.py:179`, `llm_utils.py:1161`). Verified safe today, and this exact direct-comparison idiom already exists pre-PR at `agent_tool.py:1180`, so it's a pure style preference between two idioms that already coexist — optional only.

src/xagent/web/services/agent_team_scope.py
AlexLiu190625: Addressed. The raw candidate length is now bounded before both regex scanning and integer conversion, using a named digit limit derived from the PostgreSQL INTEGER maximum. A focused regression asserts that an overlong persisted value is rejected before the regex or any query is reached.

src/xagent/web/services/agent_team_scope.py
AlexLiu190625: Addressed the useful simplification without introducing a bare 10 literal: the ineffective re.ASCII flag is gone, the raw-length limit is derived from _MAX_POSTGRES_INTEGER, and safe-length decimal strings are converted before the shared numeric bound rejects values above the PostgreSQL INTEGER maximum. The final bound remains necessary for integer inputs as well.

src/xagent/web/api/chat.py
AlexLiu190625: The coverage observation exposed a reachability issue rather than a missing runtime case: within _get_agent_for_task_unlocked, agent_config is assigned only when snapshot is not None, so the snapshot-is-None branch could never execute. I removed that branch and its incorrect live-session comment instead of adding a synthetic test; the real snapshot path and the separate _build_tools_for_task live-DB path remain covered.

AlexLiu190625:
Addressed the actionable review notes by bounding raw candidate length before parsing, documenting caller-owned lifecycle filtering, and removing the unreachable inline-preview branch. I kept the status-neutral resolver in the team-scope owner because SSH and preview apply different lifecycle policies, left the typed enum comparison unchanged because it is already correct, and did not add a user-visible authorization distinction that would expand product behavior.

PR #1223, opened 2026-08-09
fix(gmail): stop best-effort watch provisioning from failing an OAuth callback

### LINKED ISSUE #1150: fix(gmail): best-effort watch provisioning can still fail an OAuth callback it promises not to
Raised in review of #1147 (finding 2) and deliberately left out of that diff, because widening an exception guard changes behavior in an error path and deserves its own review.

## Problem

`best_effort_provision_gmail_watches_for_user` (`src/xagent/web/services/gmail_provisioning.py:1208`) documents:

> Failures are recorded on the watch state, never raised.

Its per-account `try/except` only starts inside the loop. Two calls run before any guard:

```python
accounts = (
    db.query(UserOAuth)
    .filter(UserOAuth.user_id == int(user_id), UserOAuth.provider == "gmail")
    .all()
)
referenced_account_ids = _referenced_gmail_oauth_account_ids(
    db,
    [(int(account.id), str(account.email or "")) for account in accounts],
)
```

Anything either call raises escapes the function. The caller in `src/xagent/web/api/auth.py` invokes it after `db.commit()` with no local protection, and its broad `except Exception` turns the escape into an HTTP 500 rendered as `Authentication Failed` — while the OAuth token has already been committed. The user sees a hard failure on a connector that is in fact connected.

## Why now

#1147 fixed the one bug that was actually tripping this window (a `SELECT DISTINCT` over a `json` column, which PostgreSQL rejects). That closes the observed incident but not the structural gap: any other database error on those two calls — a lost connection, a statement timeout, a lock wait — reproduces the same misleading page despite the documented contract.

This was not hypothetical. It was reproduced on staging (`main-f1eeb5f7`, PostgreSQL 16): `user_oauth` gained the Gmail row with `gmail.modify` granted, and the browser still showed `Authentication Failed`.

## Suggested direction

Extend the existing guard to cover the account lookup and the binding resolution, keeping the same "record and continue" semantics the loop already uses. Points worth deciding in review:

- With no accounts resolved there is no watch state to record the failure on, so the failure has nowhere to land except the log. Decide whether that is acceptable or whether the caller needs a signal.
- Consider whether `src/xagent/web/api/auth.py` should also stop letting post-commit provisioning affect the callback response at all, since by that point the connect has succeeded. That is arguably the more durable fix: a best-effort side effect should not be able to change what the user sees.

A regression test should assert that a raising account lookup leaves the OAuth callback successful.

codeacme17:
Closes #1150.

## Problem

`best_effort_provision_gmail_watches_for_user` documents that failures are recorded on the watch state and never raised, but its `try/except` only started inside the per-account loop. The account lookup and the trigger binding resolution ran before any guard, so a database error on either escaped the function.

The caller in `src/xagent/web/api/auth.py` invokes it after `db.commit()` with no local protection, and the callback's broad `except Exception` turned that escape into an HTTP 500 rendered as `Authentication Failed` — while the OAuth token had already been committed. The user saw a hard failure on a connector that was in fact connected.

#1147 fixed the one bug that was actually tripping this window, but not the structural gap: any other database error on those two calls — a lost connection, a statement timeout, a lock wait — reproduces the same misleading page.

## Changes

Two guards, deliberately overlapping:

1. **Service** (`gmail_provisioning.py`) — extend the existing guard to cover the account lookup and the binding resolution, keeping the loop's rollback-and-log semantics.
2. **Call site** (`api/auth.py`) — guard the wrapper itself, so a best-effort side effect running after the connect has already succeeded can never change what the callback returns.

The issue left two points open for review. Both are resolved here:

- **Where a pre-loop failure lands.** With no accounts resolved there is no watch state to record it on, so it is only logged. That is acceptable because the mailbox is not stranded: `scan_due_gmail_watch_renewals` outer-joins `UserOAuth` to `GmailWatchState` and selects rows where `GmailWatchState.id IS NULL` for users with an enabled Gmail trigger, then provisions them. `sweep_gmail_provisioning` cannot cover this case — it only retries mailboxes that already have a watch state row — and the inline comment says so explicitly.
- **Whether the callback should be insulated at all.** Yes; this is change 2 above. The redundancy with change 1 is intentional: it keeps future edits to that wrapper from reintroducing the same failure mode.

## Testing

Three regression tests, each verified to fail without the corresponding fix:

- a raising account lookup is swallowed by the service;
- a raising binding resolution is swallowed by the service;
- a raising account lookup driven through the *real* service leaves the OAuth callback successful — this reproduces the reported symptom exactly (`assert 500 == 200`, `OperationalError` on `SELECT user_oauth`) when the fix is reverted.

`tests/web/test_meta_oauth.py`, `tests/web/services/test_gmail_provisioning.py` and `tests/web/api/test_gmail_auto_triggers.py`: 118 passed, 4 skipped (the skips are the PostgreSQL-only cases). The wider `pytest tests/web -m "not slow and not postgresql"` leg was also run locally: the only failures are 5 pre-existing SVG rasterization tests that need `libcairo`, which is absent on this machine and fails identically on a clean tree.

### DIFF
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -34,6 +34,7 @@ from ..models.database import get_db
 from ..models.system_setting import SystemSetting
 from ..models.user import User
 from ..models.user_oauth import UserOAuth
+from ..services import gmail_provisioning
 from ..services.auth_email import send_password_reset_email
 
 logger = logging.getLogger(__name__)
@@ -54,12 +55,12 @@ def _best_effort_ensure_gmail_watches_for_user(db: Session, *, user_id: int) -> 
     failed while it is in fact connected. A best-effort side effect must not
     be able to change what the callback returns.
     """
+    # The module is imported at module level so that a genuine module
+    # resolution failure surfaces at startup instead of being logged here as a
+    # routine provisioning warning. The call stays inside the guard, and going
+    # through the module keeps the attribute lookup late so tests can patch it.
     try:
-        from ..services.gmail_provisioning import (
-            best_effort_provision_gmail_watches_for_user,
-        )
-
-        best_effort_provision_gmail_watches_for_user(
+        gmail_provisioning.best_effort_provision_gmail_watches_for_user(
             db,
             user_id=user_id,
             context="after OAuth callback",
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -46,15 +46,31 @@ EMAIL_PATTERN = re.compile(r"^[^@\s]+@[^@\s]+\.[^@\s]+$")
 
 
 def _best_effort_ensure_gmail_watches_for_user(db: Session, *, user_id: int) -> None:
-    from ..services.gmail_provisioning import (
-        best_effort_provision_gmail_watches_for_user,
-    )
+    """Provision Gmail watches without letting the outcome reach the caller.
 
-    best_effort_provision_gmail_watches_for_user(
-        db,
-        user_id=user_id,
-        context="after OAuth callback",
-    )
+    This runs after the OAuth token is committed, so the connect has already
+    succeeded. Anything raised here would be rendered by the callback's outer
+    handler as an ``Authentication Failed`` 500, telling the user a connector
+    failed while it is in fact connected. A best-effort side effect must not
+    be able to change what the callback returns.
+    """
+    try:
+        from ..services.gmail_provisioning import (
+            best_effort_provision_gmail_watches_for_user,
+        )
+
+        best_effort_provision_gmail_watches_for_user(
+            db,
+            user_id=user_id,
+            context="after OAuth callback",
+        )
+    except Exception:
+        logger.warning(
+            "Best-effort Gmail watch provisioning failed for user %s "
+            "after OAuth callback",
+            user_id,
+            exc_info=True,
+        )
 
 
 def _oauth_env_name(provider: str, suffix: str) -> str:
--- a/src/xagent/web/services/gmail_provisioning.py
+++ b/src/xagent/web/services/gmail_provisioning.py
@@ -1218,17 +1218,38 @@ def best_effort_provision_gmail_watches_for_user(
 
     Used after OAuth (re)connects a Gmail account: any enabled Gmail trigger
     already bound to that mailbox gets its delivery resources re-ensured.
-    Failures are recorded on the watch state, never raised.
+    Failures are recorded on the watch state, never raised; a failure before
+    any account resolves has no watch state to land on and is only logged.
     """
-    accounts = (
-        db.query(UserOAuth)
-        .filter(UserOAuth.user_id == int(user_id), UserOAuth.provider == "gmail")
-        .all()
-    )
-    referenced_account_ids = _referenced_gmail_oauth_account_ids(
-        db,
-        [(int(account.id), str(account.email or "")) for account in accounts],
-    )
+    # The account lookup and the binding resolution are inside the guard, not
+    # just the per-account loop: the caller runs this after committing the
+    # OAuth token, so a raise here would turn an already-successful connect
+    # into an error page. A mailbox missed here is picked up by
+    # ``scan_due_gmail_watch_renewals``, which selects Gmail accounts owned by
+    # users with an enabled Gmail trigger and no watch state row at all.
+    # ``sweep_gmail_provisioning`` cannot cover this: it only retries mailboxes
+    # that already have a watch state row.
+    try:
+        accounts = (
+            db.query(UserOAuth)
+            .filter(UserOAuth.user_id == int(user_id), UserOAuth.provider == "gmail")
+            .all()
+        )
+        referenced_account_ids = _referenced_gmail_oauth_account_ids(
+            db,
+            [(int(account.id), str(account.email or "")) for account in accounts],
+        )
+    except Exception as exc:
+        db.rollback()
+        logger.warning(
+            "Failed to resolve Gmail accounts for user %s %s: %s",
+            user_id,
+            context,
+            exc,
+            exc_info=True,
+        )
+        return
+
     for account in accounts:
         email = str(account.email or "").strip().lower()
         if not email:
--- a/src/xagent/web/services/gmail_provisioning.py
+++ b/src/xagent/web/services/gmail_provisioning.py
@@ -1220,15 +1220,24 @@ def best_effort_provision_gmail_watches_for_user(
     already bound to that mailbox gets its delivery resources re-ensured.
     Failures are recorded on the watch state, never raised; a failure before
     any account resolves has no watch state to land on and is only logged.
+
+    Rolls back ``db`` on failure, so callers must not hold uncommitted work on
+    that session across this call.
     """
     # The account lookup and the binding resolution are inside the guard, not
     # just the per-account loop: the caller runs this after committing the
     # OAuth token, so a raise here would turn an already-successful connect
-    # into an error page. A mailbox missed here is picked up by
-    # ``scan_due_gmail_watch_renewals``, which selects Gmail accounts owned by
-    # users with an enabled Gmail trigger and no watch state row at all.
-    # ``sweep_gmail_provisioning`` cannot cover this: it only retries mailboxes
-    # that already have a watch state row.
+    # into an error page.
+    #
+    # A mailbox missed here has no automatic retry under the default
+    # configuration. ``XAGENT_GMAIL_WATCH_ENABLED`` defaults to false and gates
+    # both ``scan_due_gmail_watch_renewals`` and ``sweep_gmail_provisioning``;
+    # only with it set does the former recover this case, by selecting Gmail
+    # accounts that back an enabled Gmail trigger and have no watch state row.
+    # (``sweep_gmail_provisioning`` never covers it either way: it only retries
+    # mailboxes that already have such a row.) Otherwise the mailbox stays
+    # unprovisioned until its trigger is next saved, which re-enters
+    # provisioning through ``GmailTriggerProvider.register``.
     try:
         accounts = (
             db.query(UserOAuth)
--- a/tests/web/services/test_gmail_provisioning.py
+++ b/tests/web/services/test_gmail_provisioning.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import os
 import threading
 from datetime import datetime, timedelta, timezone
-from typing import Any
+from typing import Any, NoReturn
 
 import pytest
 from sqlalchemy import create_engine, event
 from sqlalchemy import text as sql_text
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import Session
 from sqlalchemy.pool import QueuePool
 
@@ -521,6 +522,74 @@ def test_best_effort_provision_matches_duplicate_mailboxes_by_account_id(
     )
 
     assert provisioned == [int(first.id)]
+
+
+def test_best_effort_provision_swallows_account_lookup_failure(
+    db_session: Session,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A failing account lookup must not escape the best-effort contract."""
+    user = _create_user(db_session)
+    agent = _create_agent(db_session, user)
+    account = _create_oauth(db_session, user)
+    _create_gmail_trigger(db_session, user, agent, account)
+    provisioned: list[int] = []
+
+    monkeypatch.setattr(
+        gmail_provisioning,
+        "ensure_gmail_mailbox_provisioned",
+        lambda _db, account: provisioned.append(int(account.id)),
+    )
+
+    def raising_query(*_args: Any, **_kwargs: Any) -> NoReturn:
+        raise OperationalError("SELECT user_oauth", {}, Exception("connection lost"))
+
+    monkeypatch.setattr(db_session, "query", raising_query)
+
+    gmail_provisioning.best_effort_provision_gmail_watches_for_user(
+        db_session,
+        user_id=int(user.id),
+        context="test",
+    )
+
+    assert provisioned == []
+
+
+def test_best_effort_provision_swallows_binding_resolution_failure(
+    db_session: Session,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A failing trigger-binding lookup must not escape either."""
+    user = _create_user(db_session)
+    agent = _create_agent(db_session, user)
+    account = _create_oauth(db_session, user)
+    _create_gmail_trigger(db_session, user, agent, account)
+    provisioned: list[int] = []
+
+    monkeypatch.setattr(
+        gmail_provisioning,
+        "ensure_gmail_mailbox_provisioned",
+        lambda _db, account: provisioned.append(int(account.id)),
+    )
+
+    def raising_resolution(*_args: Any, **_kwargs: Any) -> NoReturn:
+        raise OperationalError(
+            "SELECT agent_triggers", {}, Exception("statement timeout")
+        )
+
+    monkeypatch.setattr(
+        gmail_provisioning,
+        "_referenced_gmail_oauth_account_ids",
+        raising_resolution,
+    )
+
+    gmail_provisioning.best_effort_provision_gmail_watches_for_user(
+        db_session,
+        user_id=int(user.id),
+        context="test",
+    )
+
+    assert provisioned == []
 
 
 def test_unregister_releases_mailbox_only_after_last_enabled_trigger_is_deleted(
--- a/tests/web/services/test_gmail_provisioning.py
+++ b/tests/web/services/test_gmail_provisioning.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 import threading
 from datetime import datetime, timedelta, timezone
@@ -527,12 +528,12 @@ def test_best_effort_provision_matches_duplicate_mailboxes_by_account_id(
 def test_best_effort_provision_swallows_account_lookup_failure(
     db_session: Session,
     monkeypatch: pytest.MonkeyPatch,
+    caplog: pytest.LogCaptureFixture,
 ) -> None:
     """A failing account lookup must not escape the best-effort contract."""
+    # No trigger fixture here: the patched query raises on the service's first
+    # statement, so the binding data would never be read.
     user = _create_user(db_session)
-    agent = _create_agent(db_session, user)
-    account = _create_oauth(db_session, user)
-    _create_gmail_trigger(db_session, user, agent, account)
     provisioned: list[int] = []
 
     monkeypatch.setattr(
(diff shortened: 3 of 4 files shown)


src/xagent/web/services/gmail_provisioning.py
```diff
@@ -1218,17 +1218,38 @@ def best_effort_provision_gmail_watches_for_user(
...
+            .filter(UserOAuth.user_id == int(user_id), UserOAuth.provider == "gmail")
+            .all()
+        )
+        referenced_account_ids = _referenced_gmail_oauth_account_ids(
+            db,
+            [(int(account.id), str(account.email or "")) for account in accounts],
+        )
+    except Exception as exc:
+        db.rollback()
+        logger.warning(
```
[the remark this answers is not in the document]
codeacme17: Thanks — the mechanism is real, but I'm declining this one for three reasons.

**The named scenario is the one that cannot raise.** Under the locked sqlalchemy 2.0.47, `Connection._rollback_impl` (`engine/base.py:1112`) only calls `do_rollback` when `_still_open_and_dbapi_connection_is_valid`. A true disconnect has already invalidated the pooled connection via `_handle_dbapi_exception`, so the ROLLBACK never reaches the DBAPI. The reachable window is narrower than described: a *non*-disconnect failure (statement timeout, serialization failure) where the connection is still valid and then dies before the rollback.

**It is already contained.** The only production caller is `_best_effort_ensure_gmail_watches_for_user` (`src/xagent/web/api/auth.py:48`), which this same PR wraps in `try/except Exception` with `exc_info=True`. A secondary rollback error is caught there, and the callback touches no database after that point — so the #1150 symptom does not reoccur. That belt-and-braces call-site guard is deliberate, precisely for failure modes like this one.

**The suggested patch would fail CI.** `except Exception as rollback_exc` with an unused binding trips `F841` under this repo's ruff defaults, and the `ruff-check` pre-commit hook gates the build. `logger.debug` also diverges from this file's convention (7 `logger.warning`, no `logger.debug`).

On consistency: this file has seven unguarded `db.rollback()` calls inside `except` blocks, including the pre-existing one five lines below the new guard (line 1262) and one at line 856 under an even stronger "Never raises for provisioning failures" docstring, with further DB work after it. Hardening only the new line would make it inconsistent with its own neighbour. If we want this guarantee, the right shape is a shared `_safe_rollback(db)` helper applied to all sites in a follow-up — happy to open an issue for that.

codeacme17:
Addressed the review: the single inline comment (wrapping `db.rollback()` in a nested `try/except`) is declined, with the reasoning in [that thread](https://github.com/xorbitsai/xagent/pull/1223#discussion_r3745679999).

In short: the disconnect scenario it names is short-circuited by SQLAlchemy before the ROLLBACK reaches the DBAPI; any secondary error that *is* reachable is already caught by the call-site guard this PR adds, so the #1150 symptom cannot reoccur; and the suggested patch trips `F841`, which the `ruff-check` pre-commit hook gates on. Seven pre-existing sites in the same file share the unguarded pattern, so hardening only the new line would be inconsistent — that belongs in a follow-up as a shared helper.

No code changes in response to this review.

src/xagent/web/services/gmail_provisioning.py
```diff
@@ -1218,17 +1218,38 @@ def best_effort_provision_gmail_watches_for_user(
...
-    )
-    referenced_account_ids = _referenced_gmail_oauth_account_ids(
-        db,
-        [(int(account.id), str(account.email or "")) for account in accounts],
-    )
+    # The account lookup and the binding resolution are inside the guard, not
+    # just the per-account loop: the caller runs this after committing the
+    # OAuth token, so a raise here would turn an already-successful connect
+    # into an error page. A mailbox missed here is picked up by
+    # ``scan_due_gmail_watch_renewals``, which selects Gmail accounts owned by
```
rogercloud: **MAJOR — must fix before merge.** This comment states unconditionally that "a mailbox missed here is picked up by `scan_due_gmail_watch_renewals`." That's the PR's own justification for silently swallowing pre-loop failures instead of raising — but it doesn't hold in the default deployment:

- `scan_due_gmail_watch_renewals` (`gmail_triggers.py:299-307`) short-circuits with `if not get_gmail_watch_enabled(): return 0`.
- `get_gmail_watch_enabled()` (`config.py:1263-1265`) defaults to `False`.
- The scheduler (`app.py:282-292`) gates *both* `scan_due_gmail_watch_renewals` and `sweep_gmail_provisioning` behind that same flag.
- `best_effort_provision_gmail_watches_for_user` itself is not gated by this flag at all.

So with the flag unset (the default), a mailbox that hits this exact catch path is silently and permanently missed by every healing mechanism this comment cites — no watch-state row is ever created, so nothing retries it later.

Suggestion: qualify the comment, e.g. "...picked up by `scan_due_gmail_watch_renewals`, when `XAGENT_GMAIL_WATCH_ENABLED` is set."

src/xagent/web/services/gmail_provisioning.py
```diff
@@ -1218,17 +1218,38 @@ def best_effort_provision_gmail_watches_for_user(
 
     Used after OAuth (re)connects a Gmail account: any enabled Gmail trigger
     already bound to that mailbox gets its delivery resources re-ensured.
-    Failures are recorded on the watch state, never raised.
+    Failures are recorded on the watch state, never raised; a failure before
```
rogercloud, line 1221: Minor — this docstring documents that failures are recorded/logged and never raised, but doesn't mention that satisfying that contract requires calling `db.rollback()` on the caller-supplied session (lines 1243 and 1262). The current sole caller commits before invoking this function, so it's safe today, but the guard's window is now wider, and nothing warns a future caller holding uncommitted work on the same session that it can be silently discarded.

Suggestion: add one line, e.g. "May call `db.rollback()` on `db` if provisioning fails; callers should not hold uncommitted work on `db` across this call." (Distinct from the already-declined suggestion to guard the rollback call itself.)

src/xagent/web/api/auth.py
```diff
@@ -46,15 +46,31 @@
...
-        context="after OAuth callback",
-    )
+    This runs after the OAuth token is committed, so the connect has already
+    succeeded. Anything raised here would be rendered by the callback's outer
+    handler as an ``Authentication Failed`` 500, telling the user a connector
+    failed while it is in fact connected. A best-effort side effect must not
+    be able to change what the callback returns.
+    """
+    try:
+        from ..services.gmail_provisioning import (
```
rogercloud: Minor — this import sits inside the `try:` block, so a circular import, a rename, or a packaging/deployment defect in `gmail_provisioning` would be caught by the same broad `except Exception` and logged as a routine "provisioning failed" warning instead of surfacing as a loud bug.

Suggestion: move the import above the `try` (or to module level) so module-resolution failures stay visible.

tests/web/services/test_gmail_provisioning.py
```diff
@@ -523,6 +524,74 @@ def test_best_effort_provision_matches_duplicate_mailboxes_by_account_id(
     assert provisioned == [int(first.id)]
 
 
+def test_best_effort_provision_swallows_account_lookup_failure(
```
rogercloud, line 528: Minor (test quality) — two things on this test:

1. It builds an `agent` and a Gmail trigger fixture that are never actually read — the monkeypatched `db_session.query` raises on the very first statement, before the trigger-binding logic that would consume that data ever runs. Consider trimming the unused setup so the test doesn't imply a dependency it doesn't have.
2. The assertion only checks that `provisioned` stays empty, without asserting the expected `logger.warning` fired or that the session was left in a rollback'd/reusable state. An implementation that never reached the guarded code at all would pass this test undetected. Consider adding a `caplog` assertion (there's zero log-assertion coverage for the binding-resolution variant of this test anywhere in the suite).

tests/web/services/test_gmail_provisioning.py
```diff
@@ -523,6 +524,74 @@ def test_best_effort_provision_matches_duplicate_mailboxes_by_account_id(
...
+    gmail_provisioning.best_effort_provision_gmail_watches_for_user(
+        db_session,
+        user_id=int(user.id),
+        context="test",
+    )
+
+    assert provisioned == []
+
+
+def test_best_effort_provision_swallows_binding_resolution_failure(
```
rogercloud, line 562: Minor (test quality) — same as the account-lookup variant above: this only asserts `provisioned == []`, with no `caplog` assertion on the warning log and no session-state check. This is the one scenario (binding-resolution failure) with zero log-assertion coverage anywhere in the test suite — worth prioritizing a `caplog` addition here.

tests/web/test_meta_oauth.py
```diff
@@ -162,6 +164,150 @@ def fake_best_effort_provision(_db, *, user_id: int, context: str):
...
+
+    real_provision = gmail_provisioning.best_effort_provision_gmail_watches_for_user
+
+    def failing_query(*_args, **_kwargs):
+        raise OperationalError("SELECT user_oauth", {}, Exception("connection lost"))
+
+    def provision_with_failing_account_lookup(provision_db, **kwargs):
+        # Fail only the lookups the service itself issues; the callback's own
+        # queries, which run before the OAuth commit, must succeed.
+        provision_db.query = failing_query
```
rogercloud, line 288: Minor (test quality) — this manually sets and `del`s `provision_db.query` instead of using `monkeypatch.setattr`, which is already used elsewhere in this file and in the sibling service test for the identical purpose. If a future refactor skips the assignment on line 284, the `del` in the `finally` block (line 288) raises `AttributeError` and masks the real test failure.

Suggestion: `monkeypatch.setattr(provision_db, "query", failing_query)` for automatic, order-independent teardown.

rogercloud (approved):
## Summary

This PR fixes issue #1150: `best_effort_provision_gmail_watches_for_user` documented a "never raises" contract, but its try/except only wrapped the per-account loop, not the account lookup and trigger-binding resolution that run before it. Because `generic_oauth_callback` in `src/xagent/web/api/auth.py` invokes this helper *after* `db.commit()`, and the callback has a single blanket `except Exception`, any escape from those pre-loop DB calls rendered an HTTP 500 "Authentication Failed" even though the OAuth token had already been committed successfully.

The fix extends the service-level guard in `gmail_provisioning.py` to cover the pre-loop work, and adds a second guard at the call site in `auth.py` as deliberate defense-in-depth (verified: not redundant — the service's own `db.rollback()` call is itself unguarded, so a secondary exception there is only caught by the call-site guard).

## Design verdict: Sound

The fix is minimal and restores the function to its own documented contract. One informational, non-blocking note: `generic_oauth_callback` still has a single blanket `except Exception` spanning pre- and post-commit code; this PR hardens the one known post-commit side effect but doesn't restructure the handler at the commit boundary, so a future post-commit addition could reproduce the same bug class. Reasonable scope choice for this fix — worth a follow-up.

The previously raised bot suggestion (nest a try/except around the new `db.rollback()` call) was correctly declined by the author — all three of the author's technical arguments were independently verified against the actual pinned SQLAlchemy internals, the call-site containment, and the ruff/pre-commit config. Not re-raised here.

## ⚠️ Must fix before merge

One finding below is flagged **MAJOR** and should be fixed before merge: the new comment's resilience claim ("picked up by `scan_due_gmail_watch_renewals`") does not hold under the default configuration (`XAGENT_GMAIL_WATCH_ENABLED` defaults to `False`), which undermines the PR's own stated justification for silently swallowing pre-loop failures. See the inline comment for details — this is a real correctness-of-documentation issue, not a style nit, since it's the reasoning the PR relies on to justify not raising.

Everything else below is minor polish / optional follow-up and does not block merge.

Lean already — three candidate simplifications (dropping the outer try/except in auth.py, merging the two service-level swallow tests, dropping the integration-level callback test) were each considered and verified to protect a distinct, real regression surface. None are safe to apply.

## Test results

Ran `tests/web/services/test_gmail_provisioning.py`, `tests/web/test_meta_oauth.py`, and `tests/web/api/test_gmail_auto_triggers.py` in a clean venv: all passed (exit code 0), 4 skipped (require a Postgres env var not set locally).

Approving since the fix is sound and well-tested, but please address the MAJOR comment (or at minimum soften the comment's claim) before merging.

src/xagent/web/services/gmail_provisioning.py
codeacme17: Fixed in f1e28952 — you are right, and I verified all four of your claims against the code. `get_gmail_watch_enabled()` (`config.py:1265`) is `_get_bool_env(GMAIL_WATCH_ENABLED, False)`, `gmail_triggers.py:306` short-circuits on it, `app.py:283` gates the renewal scan and the sweep together, and this function has no such gate. With the flag unset nothing automatic retries the mailbox, so the comment was asserting a safety net that is off by default.

I went slightly further than the suggested qualifier. "...when `XAGENT_GMAIL_WATCH_ENABLED` is set" is accurate but still leaves a reader thinking the default case is simply unrecoverable, so the comment now also names the one healing path that is *not* flag-gated: re-saving the trigger, which re-enters provisioning through `GmailTriggerProvider.register" (`trigger_providers/gmail.py:431` → `_provision_in_fresh_session`). The `sweep_gmail_provisioning` caveat is kept, since it never covers this case either way.

One adjacent thing I found while checking this, which I have left alone as out of scope: with the flag off, the OAuth and trigger-create paths still create watches, but the renewal loop that would renew them is gated off — so those watches expire after roughly seven days with nothing renewing them. That is pre-existing and unrelated to this guard; happy to open a separate issue for it.

src/xagent/web/services/gmail_provisioning.py
codeacme17: Fixed in f1e28952. Added to the docstring: "Rolls back ``db`` on failure, so callers must not hold uncommitted work on that session across this call." Agreed it belongs in the contract rather than being implied by the body — the guard's window is wider now, and the next caller has no reason to read the `except` blocks first.

src/xagent/web/api/auth.py
codeacme17: Fixed in f1e28952, though not quite as suggested — the diagnosis is right but both proposed placements have a problem.

Moving the import above the `try` *inside the function* would let an ImportError propagate into `generic_oauth_callback`'s blanket `except Exception` and render the `Authentication Failed` 500 on an already-committed connect — exactly the failure #1150 exists to remove. And hoisting the *symbol* to module level breaks three tests that patch the service module attribute (`test_meta_oauth.py:151`, `:212`, `:292`), because the name would then be bound at import time.

So I import the *module* at module level and call through it:

```python
from ..services import gmail_provisioning   # module level
...
try:
    gmail_provisioning.best_effort_provision_gmail_watches_for_user(...)
```

Module resolution now fails at startup where it is loud, the attribute lookup stays late so the call remains patchable, and the `try` guards only the call. Verified there is no circular-import reason for the original laziness: importing `xagent.web.services.gmail_provisioning` in a clean interpreter does not pull `xagent.web.api.auth` into `sys.modules`, and `api/triggers.py:18` already imports from this service at module level. `xagent.web.app` and `api.auth` both still import cleanly.

tests/web/services/test_gmail_provisioning.py
codeacme17: Both points fixed in f1e28952.

The fixture audit was right, and slightly understated — the dead setup spans both tests, not just this one. In this test the patched `query` raises on the service's first statement, so `_create_agent`, `_create_oauth` and `_create_gmail_trigger` were all unread; only `_create_user` matters, for `user.id`. Trimmed to that, with a comment saying why there is no trigger fixture. In the binding-resolution test below, the account lookup runs for real so `_create_oauth` does matter, but the agent and trigger were still dead weight — trimmed there too.

Added `caplog` assertions on both the message and the underlying error text, so the test pins the log-and-return semantics rather than just the empty post-state.

tests/web/services/test_gmail_provisioning.py
codeacme17: Fixed in f1e28952 — this one now asserts `"Failed to resolve Gmail accounts for user"` and `"statement timeout"` via `caplog`, closing the gap you identified: the binding-resolution branch had no log assertion anywhere in the suite. Its sibling above got the same treatment, and the dead agent/trigger fixtures are gone from both.

tests/web/test_meta_oauth.py
codeacme17: Addressed the underlying concern in f1e28952, but kept the `try/finally` rather than switching to `monkeypatch.setattr` — there is a constraint that makes the swap unsafe here.

The patch has to be reverted *before* the assertion at the end of this test, which queries the session again to confirm the OAuth row was committed. `monkeypatch.setattr` reverts at teardown, so with it the later `db.query(UserOAuth)` would hit `failing_query` and the test would fail. `pytest.MonkeyPatch.context()` inside the wrapper would give the right scope, but buys nothing over the existing `try/finally`, which is already exception-safe — and it restores slightly worse: `query` is inherited from the class, so monkeypatch's undo re-`setattr`s a *bound method* as a lingering instance attribute, where removing the instance attribute restores the real lookup.

What I did take is your actual failure mode — `del` raising `AttributeError` and masking a real failure if a refactor skips the assignment. That is now `provision_db.__dict__.pop("query", None)`, plus a comment recording why the revert must happen inside the wrapper rather than at teardown. Happy to switch to `MonkeyPatch.context()` if you would still prefer the monkeypatch API here.

codeacme17:
Thanks for the thorough review — pushed f1e28952. All six findings addressed; replies are in each thread.

**MAJOR — the retry claim.** You were right, and I verified all four of your claims against the code. The comment now names `XAGENT_GMAIL_WATCH_ENABLED`, states that it gates both healing mechanisms, and additionally names the one recovery path that is *not* flag-gated (re-saving the trigger, via `GmailTriggerProvider.register`) — a bare "when the flag is set" qualifier would still have left readers thinking the default case is unrecoverable.

**Minors.** Docstring now documents the `db.rollback()` on the caller's session. Both service tests assert the warning log and dropped the fixtures they never read. Two I implemented differently from the suggestion, with the reasoning in-thread:

- *The lazy import* — I import the **module** at module level rather than hoisting the symbol. Hoisting the symbol breaks three tests that patch the service module attribute, and moving the import above the `try` inside the function would let an ImportError render the 500 this PR exists to remove. Going through the module keeps resolution at startup, the lookup late, and the `try` around only the call. Confirmed empirically that no circular import forced the original laziness.
- *`del` → `monkeypatch.setattr`* — the patch must be reverted before this test's later `db.query(UserOAuth)` assertion, which `monkeypatch.setattr` (teardown-scoped) would break. I took the real failure mode instead: `__dict__.pop("query", None)`, so a refactor that skips the assignment fails on its own terms rather than through an `AttributeError` from the revert.

**On your informational note** about `generic_oauth_callback`'s blanket `except Exception` spanning the commit boundary: agreed, and I'd rather do it as a follow-up than widen this PR. Two adjacent items I found while working through the MAJOR, both pre-existing and both worth their own issues:

1. With `XAGENT_GMAIL_WATCH_ENABLED` off, the OAuth and trigger-create paths still create watches, but the renewal loop that would renew them is gated off — so those watches expire after ~7 days with nothing renewing them.
2. This file has seven unguarded `db.rollback()` calls inside `except` blocks, including one under an even stronger "Never raises" docstring. A shared `_safe_rollback(db)` helper would be the right shape (this is the bot suggestion I declined for this PR on consistency grounds).

Happy to file all three.

**Verification.** `tests/web/test_meta_oauth.py`, `tests/web/services/test_gmail_provisioning.py`, `tests/web/api/test_gmail_auto_triggers.py`: 118 passed, 4 skipped. Full `pytest tests/web -m "not slow and not postgresql"`: 4625 passed; the only failures are 5 pre-existing `libcairo` SVG tests and 6 sandbox Docker e2e tests, none of which this branch touches — the diff against the base is confined to four files and `test_sandbox_reconcile_e2e.py` references neither changed module. `ruff check src tests` clean, and `xagent.web.app` imports cleanly with the new module-level import.

PR #1124, opened 2026-08-04
enh(widget): add remaining share-path abuse controls (auth, task-create, run quota) (#1108)

codeacme17:
Closes #1108. Spun out of #1056 (whose two widget websocket limiters landed in #1113). The widget public surfaces were added after the #973/#993 share hardening and never picked up three of its abuse controls. This adds them on the shared `get_share_rate_limiter()` substrate, in namespaces and env vars kept separate from the share buckets so the two public channels cannot consume each other's budgets.

## Gaps closed

| # | Surface | Gate | Keying | Defaults |
|---|---------|------|--------|----------|
| 2 | `POST /api/widget/auth` + `POST /api/widget/embed-ticket` | `allow_widget_auth(entity_key, ip)` | tight per-IP (per-visitor bound) + loose per-entity backstop; entity is the widget key or the owner entity decoded from the ticket's **signed claims** (pure HMAC check, no DB) | 300/min IP, 1200/min entity |
| 3 | widget `POST /chat/task/create` | `allow_widget_task_create(entity_key, ip)` | per-entity (loose) + per-caller-IP (tight), non-destructive test→hit | 240/min entity, 60/min IP |
| 1 | run quota at the `execute_task` chokepoint | `widget_run_denial_reason(entity_key, creator_ip)` | rolling per-entity ceiling + a per-creating-IP sub-quota keyed `entity\|ip` (per-abuser dimension, scoped to one widget) | 500/day entity, 120/hour IP |

## Design constraint

Every widget gate uses the **tight-per-IP / loose-per-entity** shape: the caller IP is the per-visitor / per-abuser bound (the widget `guest_id` is client-supplied and rotatable, so it is never a limiter key), and the per-entity bucket is a loose aggregate backstop across one widget's visitors. Both buckets are tested non-destructively and only consumed when both admit, so a backstop denial never burns the caller's per-IP allowance for other widgets. This is the same reasoning `allow_widget_upload` / `allow_widget_ws_turn` already document, applied to the three new gates.

## 1 — run quota

[...]

PR #1224, opened 2026-08-09
fix(rate-limit): accept X-Forwarded-For entries that carry a port (#1185)

### LINKED ISSUE #1185: fix(rate-limit): X-Forwarded-For entries with a port collapse onto the peer IP
From the #1124 round-8 review (new finding 1).

## Problem

`remote_ip_from_request` (`src/xagent/web/services/trigger_rate_limit.py`) validates the selected `X-Forwarded-For` entry with `ipaddress.ip_address()` — added in #1124 so a client-controlled entry cannot be persisted or used as Redis key material. That validation rejects the `ip:port` and `[ipv6]:port` forms outright, so a proxy that appends a port silently degrades every request behind it to the peer/proxy address, collapsing all of those clients into a single rate-limit bucket.

Bare IPs are the convention (nginx, ALB, most CDNs), and the fallback is safe rather than permissive, so this is narrow — but it turns a correct deployment into one global bucket with only a warning log to show for it.

## Fix

Strip a trailing `:port` (and surrounding brackets for IPv6) before validating, then validate the remainder. Keep the peer-IP fallback for anything still not IP-shaped.

Test alongside the existing cases in `tests/web/api/test_trigger_rate_limits.py::TestRemoteIpDerivation` (which already covers non-IP, oversized, trailing-junk, and bare IPv6 entries).

codeacme17:
Closes #1185.

## Problem

`remote_ip_from_request` validated the selected `X-Forwarded-For` entry with `ipaddress.ip_address()` — added in #1124 so a client-controlled entry cannot be persisted or used as Redis key material. That validation rejects the `ip:port` and `[ipv6]:port` forms outright, so behind a proxy that appends the client's source port every request silently degraded to the peer address.

The blast radius is wider than the trigger callback path the module is named for. The helper is shared by:

- `api/triggers.py` — callback rate limiting and 429 audit dedup
- `api/widget.py` — widget rate limiting, and `client_ip` is **persisted** into `agent_config` as the per-abuser key for the #1108 run quota
- `api/share.py` and `api/public_chat_access.py` — share link and WebSocket handshake limiting

So in an affected deployment it is not just one global rate-limit bucket: the widget per-guest run quota also collapses, and a single abuser can consume the whole link's budget.

## Fix

Strip the port before validating and return the bare IP, so buckets stay keyed per client rather than per ephemeral connection. Anything still not IP-shaped keeps the peer-address fallback and the existing warning log.

Splitting only happens for forms where a port is unambiguous:

| Entry | Result |
|---|---|
| `203.0.113.7:8080` | `203.0.113.7` |
| `[2001:db8::1]:8080` | `2001:db8::1` |
| `[2001:db8::1]` | `2001:db8::1` |
| `2001:db8::1:8080` | unchanged — see below |
| `203.0.113.7:evil`, `[2001:db8::1`, `evil:8080` | peer address |

A bare IPv6 literal is never split. `2001:db8::1:8080` is a valid address whose last group only looks like a port, so splitting on the final colon would silently rewrite it into a *different*, still-valid address — worse than the peer fallback, because nothing would flag it. Only bracketed IPv6 and a host with exactly one colon are treated as carrying a port.

The port is discarded, but still range-checked so a malformed entry falls back to the peer address rather than being treated as well-formed. Two bounds are load-bearing there:

- `len(port) <= 5` **before** `int()` — `int()` raises on digit strings longer than `sys.get_int_max_str_digits()` (4300 by default), so without it a client-controlled `X-Forwarded-For: 1.2.3.4:99…9` would escape as an unhandled `ValueError` and 500 every endpoint listed above.
- `isascii()` before `isdigit()` — `str.isdigit()` also accepts non-ASCII digit characters (superscripts among them) that `int()` then rejects.

No security property from #1124 is relaxed: the returned value is still only ever an IP-shaped string, and everything else falls back to the peer address.

## Tests

Three cases added to the existing `TestRemoteIpDerivation`, alongside the non-IP / oversized / trailing-junk / bare-IPv6 cases already there:

- `test_port_suffixed_entries_resolve_to_the_bare_ip`
- `test_bare_ipv6_is_never_split_on_its_last_group`
- `test_malformed_port_and_bracket_forms_fall_back_to_peer` — 16 forged entries, including the >4300-digit port

```
pytest tests/web/api/test_trigger_rate_limits.py tests/web/api/test_share_rate_limit_endpoints.py
# 39 passed
```

`ruff format --check`, `ruff check` and `mypy src/xagent/web/services/trigger_rate_limit.py` are clean.

## Follow-up (not in this PR)

`trigger_rate_limit.py` now carries ~100 lines of generic reverse-proxy ingress parsing while being named for trigger rate limiting, and `share_rate_limit.py:59-61` already carries a comment apologising for importing from it. Extracting the helper into its own module is a pure refactor across five files and is deliberately left out of this bug fix.

### DIFF
--- a/src/xagent/web/services/trigger_rate_limit.py
+++ b/src/xagent/web/services/trigger_rate_limit.py
@@ -239,6 +239,14 @@ def _client_ip_from_forwarded_entry(candidate: str) -> str | None:
     quota (#1185). The port is dropped rather than kept so buckets stay keyed
     per client rather than per ephemeral connection.
     """
+    # ipaddress.ip_address() accepts an IPv6 zone-id suffix, and the zone-id
+    # itself is unbounded and unvalidated: `::1%` + 4000 characters parses and
+    # round-trips. A zone-id names a local interface, so it is meaningless for
+    # a remote client address and no proxy emits one — but leaving it accepted
+    # would defeat the bound this validation exists to enforce, since the
+    # result is persisted and used as Redis key material.
+    if "%" in candidate:
+        return None
     host, port = _split_host_port(candidate)
     if host is None:
         return None
@@ -258,7 +266,10 @@ def _split_host_port(candidate: str) -> tuple[str | None, str | None]:
     IPv6 literal is never split: ``2001:db8::1:8080`` is a valid address whose
     last group only looks like a port, so splitting on the final colon would
     silently rewrite it into a *different* address. Only the unambiguous forms
-    carry a port: bracketed IPv6, and a host with exactly one colon (IPv4).
+    carry a port: a bracketed host, and a host with exactly one colon (IPv4).
+    Brackets are not checked for IPv6 specifically — RFC 3986 reserves them for
+    it, and a bracketed IPv4 still yields the same string its bare form would,
+    so it lands in the same bucket.
     """
     if candidate.startswith("["):
         closing = candidate.find("]")
--- a/src/xagent/web/services/trigger_rate_limit.py
+++ b/src/xagent/web/services/trigger_rate_limit.py
@@ -183,7 +183,9 @@ def remote_ip_from_request(request: object) -> str | None:
 
     With ``XAGENT_TRUSTED_PROXY_HOPS=N`` (N>0), the client address is taken
     from ``X-Forwarded-For`` counting N trusted entries from the right;
-    otherwise the raw socket peer address is used.
+    otherwise the raw socket peer address is used. The returned value is
+    always a bare IP: a port appended by the proxy is stripped, and anything
+    not IP-shaped falls back to the peer address.
     """
     client = getattr(request, "client", None)
     peer_host = getattr(client, "host", None)
@@ -212,12 +214,73 @@ def remote_ip_from_request(request: object) -> str | None:
     # callers use this as rate-limit key material and persist it (#1108's
     # `widget_client_ip`), so an unvalidated header could inject an unbounded
     # arbitrary string. Fall back to the peer address rather than trusting it.
-    try:
-        ipaddress.ip_address(candidate)
-    except ValueError:
+    client_ip = _client_ip_from_forwarded_entry(candidate)
+    if client_ip is None:
         logger.warning(
             "Ignoring non-IP X-Forwarded-For entry %r; using peer address",
             candidate[:64],
         )
         return peer_ip
-    return candidate
+    return client_ip
+
+
+def _client_ip_from_forwarded_entry(candidate: str) -> str | None:
+    """Return the bare IP an X-Forwarded-For entry denotes, else None.
+
+    The entry is returned as written, not canonicalized: equivalent spellings
+    of one IPv6 address still key separate rate-limit buckets, exactly as they
+    did before ports were handled.
+
+    Bare IPs are the convention (nginx, ALB, most CDNs), but some proxies
+    (Azure App Service / Application Gateway among them) append the client's
+    source port as ``ip:port`` or ``[ipv6]:port``. Rejecting those outright
+    would silently collapse every client behind such a proxy onto the peer
+    address — one shared rate-limit bucket and, for widgets, one shared run
+    quota (#1185). The port is dropped rather than kept so buckets stay keyed
+    per client rather than per ephemeral connection.
+    """
+    host, port = _split_host_port(candidate)
+    if host is None:
+        return None
+    if port is not None and not _is_valid_port(port):
+        return None
+    try:
+        ipaddress.ip_address(host)
+    except ValueError:
+        return None
+    return host
+
+
+def _split_host_port(candidate: str) -> tuple[str | None, str | None]:
+    """Split an entry into (host, port), with port None when absent.
+
+    Returns ``(None, None)`` for structurally malformed bracket forms. A bare
+    IPv6 literal is never split: ``2001:db8::1:8080`` is a valid address whose
+    last group only looks like a port, so splitting on the final colon would
+    silently rewrite it into a *different* address. Only the unambiguous forms
+    carry a port: bracketed IPv6, and a host with exactly one colon (IPv4).
+    """
+    if candidate.startswith("["):
+        closing = candidate.find("]")
+        if closing < 0:
+            return None, None
+        host = candidate[1:closing]
+        rest = candidate[closing + 1 :]
+        if not rest:
+            return host, None
+        if not rest.startswith(":"):
+            return None, None
+        return host, rest[1:]
+    if candidate.count(":") == 1:
+        host, _, port = candidate.partition(":")
+        return host, port
+    return candidate, None
+
+
+def _is_valid_port(port: str) -> bool:
+    # The length bound has to come first: int() raises on digit strings longer
+    # than sys.get_int_max_str_digits() (4300 by default), so a header of the
+    # form `1.2.3.4:99…9` would otherwise escape as an unhandled ValueError.
+    # isascii() matters too: str.isdigit() also accepts non-ASCII digit
+    # characters (superscripts among them) that int() then rejects.
+    return len(port) <= 5 and port.isascii() and port.isdigit() and int(port) <= 65535
--- a/tests/web/api/test_trigger_rate_limits.py
+++ b/tests/web/api/test_trigger_rate_limits.py
@@ -313,6 +313,11 @@ class TestRemoteIpDerivation:
             "[2001:db8::1]:8080": "2001:db8::1",
             # Brackets with no port are still a valid IPv6 reference.
             "[2001:db8::1]": "2001:db8::1",
+            # Brackets are stripped without checking the host is IPv6. No real
+            # proxy emits this (RFC 3986 reserves brackets for IPv6), but it
+            # yields the same string the bare form would, so it cannot reach a
+            # different bucket than its canonical spelling.
+            "[1.2.3.4]:80": "1.2.3.4",
         }
         for forwarded, expected in cases.items():
             request = self._request("10.0.0.1", forwarded)
@@ -357,6 +362,17 @@ class TestRemoteIpDerivation:
             "[]:80",
             "[not-an-ip]:80",
             "evil:8080",
+            # Multi-colon non-IP: pins the no-bracket, no-port fallthrough in
+            # _split_host_port, which is otherwise only reached by entries that
+            # do parse as IPv6.
+            "a:b:c",
+            # ipaddress.ip_address() accepts an unbounded IPv6 zone-id, which
+            # would otherwise be returned verbatim and persisted as the client
+            # IP — the exact unbounded-string injection this validation exists
+            # to prevent.
+            "::1%eth0",
+            "::1%" + "A" * 4000,
+            "[fe80::1%eth0]:80",
         )
         for entry in forged:
             request = self._request("10.0.0.1", entry)
--- a/tests/web/api/test_trigger_rate_limits.py
+++ b/tests/web/api/test_trigger_rate_limits.py
@@ -298,6 +298,70 @@ class TestRemoteIpDerivation:
         request = self._request("10.0.0.1", None)
         assert remote_ip_from_request(request) == "10.0.0.1"
 
+    def test_port_suffixed_entries_resolve_to_the_bare_ip(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """#1185: some proxies append the client's source port. Rejecting
+        those would collapse every client behind such a proxy onto the peer
+        address — one shared rate-limit bucket. The port is dropped so the
+        bucket stays keyed per client, not per ephemeral connection."""
+        monkeypatch.setenv("XAGENT_TRUSTED_PROXY_HOPS", "1")
+        cases = {
+            "203.0.113.7:8080": "203.0.113.7",
+            "203.0.113.7:0": "203.0.113.7",
+            "203.0.113.7:65535": "203.0.113.7",
+            "[2001:db8::1]:8080": "2001:db8::1",
+            # Brackets with no port are still a valid IPv6 reference.
+            "[2001:db8::1]": "2001:db8::1",
+        }
+        for forwarded, expected in cases.items():
+            request = self._request("10.0.0.1", forwarded)
+            assert remote_ip_from_request(request) == expected, forwarded
+
+    def test_bare_ipv6_is_never_split_on_its_last_group(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """#1185 regression guard: an unbracketed IPv6 literal can end in a
+        group that looks like a port. Stripping it would silently rewrite the
+        entry into a *different*, still-valid address — worse than the
+        peer-address fallback, since nothing would flag it."""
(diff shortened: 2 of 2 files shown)


src/xagent/web/services/trigger_rate_limit.py
```diff
@@ -183,7 +183,9 @@ def remote_ip_from_request(request: object) -> str | None:
 
     With ``XAGENT_TRUSTED_PROXY_HOPS=N`` (N>0), the client address is taken
     from ``X-Forwarded-For`` counting N trusted entries from the right;
-    otherwise the raw socket peer address is used.
+    otherwise the raw socket peer address is used. The returned value is
+    always a bare IP: a port appended by the proxy is stripped, and anything
+    not IP-shaped falls back to the peer address.
```
rogercloud, line 188: Minor (doc accuracy): this new sentence asserts an invariant the code doesn't hold. `ipaddress.ip_address()` also accepts IPv6 zone-id syntax, and the zone-id is unbounded and unvalidated — `ipaddress.ip_address("::1%" + "A" * 4000)` succeeds and round-trips a 4004-character string. Since `_split_host_port` only splits on exactly one colon or a bracket form, `::1%AAAA…A` (two colons, unbracketed) falls through to `return candidate, None`, passes validation, and is returned as the client IP — which is then persisted as `widget_client_ip` and used as Redis key material.

Scope note: this is pre-existing, not introduced here — the base code called `ipaddress.ip_address(candidate)` directly with no splitting and behaved identically. Only the docstring claim is new. Either soften the wording (e.g. "IP-shaped per `ipaddress.ip_address`, which also accepts an unbounded IPv6 zone-id suffix such as `%ifname`") or, as the stronger fix, reject/strip a zone-id in `_client_ip_from_forwarded_entry` so the return value actually matches the claim. Non-blocking.

src/xagent/web/services/trigger_rate_limit.py
```diff
@@ -212,12 +214,73 @@ def remote_ip_from_request(request: object) -> str | None:
...
+
+
+def _split_host_port(candidate: str) -> tuple[str | None, str | None]:
+    """Split an entry into (host, port), with port None when absent.
+
+    Returns ``(None, None)`` for structurally malformed bracket forms. A bare
+    IPv6 literal is never split: ``2001:db8::1:8080`` is a valid address whose
+    last group only looks like a port, so splitting on the final colon would
+    silently rewrite it into a *different* address. Only the unambiguous forms
+    carry a port: bracketed IPv6, and a host with exactly one colon (IPv4).
```
rogercloud: Nit (doc accuracy): the docstring calls this branch "bracketed IPv6," but the code below is a pure bracket-strip with no IPv6 check — `_split_host_port("[1.2.3.4]:80")` returns `("1.2.3.4", "80")`. RFC 3986 reserves brackets for IPv6, so this form shouldn't come from a real proxy, and the outcome is harmless: `_client_ip_from_forwarded_entry("[1.2.3.4]:80")` and `_client_ip_from_forwarded_entry("1.2.3.4:80")` both return `"1.2.3.4"`, so same bucket, no bypass. Either loosen the wording to "bracketed host," or add a test pinning the accepted-but-odd form (no current test covers bracket + IPv4). Your call.

tests/web/api/test_trigger_rate_limits.py
```diff
@@ -298,6 +298,70 @@ def test_missing_forwarded_header_falls_back_to_peer(
...
+            "203.0.113.7:",
+            ":8080",
+            "[2001:db8::1",
+            "[2001:db8::1]8080",
+            "[2001:db8::1]:",
+            "[2001:db8::1]:evil",
+            "[]",
+            "[]:80",
+            "[not-an-ip]:80",
+            "evil:8080",
```
rogercloud, line 364: Minor (test coverage): the final fallthrough in `_split_host_port` (`return candidate, None`, reached for 0 or >=2 colons without brackets) is only exercised by inputs that are valid IPv6 addresses (`2001:db8::1:8080`, `::1`). Nothing covers a malformed multi-colon non-IP string. Current behavior is correct — `ipaddress.ip_address("a:b:c")` raises, so it falls back to the peer address — so this is a missing regression guard, not a bug: a later change to that branch could break it with no test going red. One more entry in this tuple would cover it:

```suggestion
            "evil:8080",
            "a:b:c",
```

rogercloud (approved):
## Summary

`#1124` added strict `ipaddress.ip_address()` validation on the selected `X-Forwarded-For` entry so a client-controlled string could not be persisted or used as Redis key material. That validation rejects `ip:port` / `[ipv6]:port` outright, so proxies that append the client's source port (Azure App Service / Application Gateway among them) make every request silently collapse onto the shared peer IP — one shared rate-limit bucket, and for widgets one shared run quota (#1108). This PR strips the port before validating, keeps the peer-IP fallback for anything not IP-shaped, and adds tests.

## Approach

The approach is sound and appropriately minimal — the fix is scoped to the parsing bug and does not reach for a refactor. The three non-obvious invariants are correct and each is load-bearing:

- a bare IPv6 literal is never split, since a valid IPv6 address always has at least two colons while the split condition requires exactly one;
- the bracket branch requires an immediate `:` or end-of-string after `]`, so malformed bracket forms fall back rather than being half-parsed;
- `_is_valid_port` bounds the length before `int()` (digit strings past `sys.get_int_max_str_digits()` raise) and checks `isascii()` before `isdigit()` (non-ASCII digits pass `isdigit()` but fail `int()`).

A `urllib.parse.urlsplit`-based rewrite is *not* a valid substitute here: it splits unbracketed hosts on the first colon, which would mis-split bare IPv6 literals — exactly the case this PR protects. It also lowercases/canonicalizes hosts and strips brackets, which conflicts with this module's documented "returned as written, not canonicalized" semantics. The hand-rolled split is the right call.

**Non-blocking design note:** `trigger_rate_limit.py` keeps accumulating generic reverse-proxy ingress-parsing logic (~60 more lines here) while its name and module docstring present it as trigger-callback rate limiting specifically; it is in fact shared by `triggers.py`, `widget.py`, `share.py`, and `public_chat_access.py`. This drift predates this PR and the description already flags extraction as a deliberate follow-up — keeping the refactor out of a bugfix is the right call. Suggestion only: file/confirm the follow-up issue, and consider a one-line addition to the module docstring noting it also provides shared ingress parsing, so the next reader is not misled.

## Findings

All three are minor and non-blocking; see inline comments for details.

1. `src/xagent/web/services/trigger_rate_limit.py:188` — doc accuracy: the new docstring asserts the return value is "always a bare IP," but `ipaddress.ip_address()` also accepts an unbounded IPv6 zone-id suffix, which is a pre-existing gap from #1124 that the new wording states incorrectly.
2. `src/xagent/web/services/trigger_rate_limit.py:261` — doc accuracy: the bracket branch is documented as "bracketed IPv6" but accepts any bracketed host (e.g. bracketed IPv4); harmless outcome.
3. `tests/web/api/test_trigger_rate_limits.py:359` — test coverage: no test pins rejection of a malformed multi-colon, non-bracketed string through the fallthrough branch.

## Simplification

Lean already — a dedicated pass for dead flexibility, hand-rolled-vs-stdlib, speculative abstraction, and shrinkable code found nothing to report. The three new helpers are a normal single-responsibility decomposition, not configuration knobs.

## Checked with no issues

- **Core split/validate correctness** (beyond the pre-existing zone-id gap above): every edge case hand-traced and empirically run — `"["`, `"]"`, `"[]"`, `"[]:80"`, `"a:b:c"`, `":"`, `":8080"`, `"1.2.3.4:"`, malformed bracket variants, and IPv4-mapped IPv6 (`::ffff:1.2.3.4`) all land in the intended branch. The empty-candidate case is unreachable (entries are filtered on truthiness before selection).
- **`_is_valid_port` boundaries**: `"".isdigit()` is `False`, so `"1.2.3.4:"` is rejected; leading zeros (`"08080"`) are accepted and harmless since the port is discarded either way.
- **Caller contract**: `remote_ip_from_request`'s signature is unchanged, and all four call sites (`src/xagent/web/api/share.py`, `widget.py`, `triggers.py`, `public_chat_access.py`) consume it consistently as `str | None`.
- **Logging**: the warning still truncates to `candidate[:64]` via `%r`, consistent with #1124's concern about logging unbounded or injectable data.
- **Test quality overall**: the three new tests are well-structured — table-driven positive cases, a dedicated bare-IPv6 regression guard, and 16 malformed inputs each isolating one of the four validation bounds (length, ASCII, digit-ness, numeric range).
- **Code quality**: helpers are single-responsibility and clearly named; comments explain the non-obvious "why" instead of restating the code.
- **Test execution**: `pytest tests/web/api/test_trigger_rate_limits.py tests/web/api/test_share_rate_limit_endpoints.py` → 39 passed, and `ruff format --check` / `ruff check` are clean on both changed files.

## Recommendation

Approve. The fix is correct, minimal, and well-tested; all findings above are minor doc-accuracy and test-coverage nits with no correctness or security regression.

codeacme17:
Thanks for the review — all three findings are addressed in 5a2da771.

**Finding 1 (zone-id).** Took the stronger fix rather than softening the wording. Reproduced first: `X-Forwarded-For: ::1%` + 4000 characters returns the full 4004-character string as the client IP, and `git show dc5ef731` confirms the base called `ipaddress.ip_address()` on the whole entry, so the behaviour is pre-existing exactly as you scoped it. But #1124's own comment states the goal is to stop an unvalidated header injecting "an unbounded arbitrary string", and the zone-id path defeats precisely that bound — so making the docstring true was the better of the two options. `_client_ip_from_forwarded_entry` now rejects any entry containing `%` before splitting, which covers both the fallthrough and the bracket route (`[fe80::1%eth0]:80`). Nothing legitimate breaks: a zone-id names a local interface, so it is meaningless for a remote client address.

One calibration on severity, in case it matters for triage elsewhere. I don't read this as a rate-limit bypass. Controlling the entry at all requires `hops` to over-count the real chain, and in that scenario an attacker can already write arbitrary IPs — bucket rotation was unbounded already. What the zone-id uniquely broke is the *length* bound: a rotated IP is ~45 characters, a zone-id is not. So the finding is unbounded-string injection into Redis keys and the persisted `widget_client_ip`, not evasion.

**Finding 2.** Did both: reworded to "a bracketed host", added a line on why brackets are not checked for IPv6, and pinned `[1.2.3.4]:80` → `1.2.3.4` in the port-stripping cases.

**Finding 3.** Added `a:b:c`. Small correction to the framing — zero-colon inputs (`not-an-ip`, `"a" * 5000`) also reach that fallthrough — but the specific gap was real: no forged case had two or more colons, since `evil:8080` takes the port branch.

**Design note.** Agreed on keeping the extraction out of a bugfix. On the module docstring sentence: I would rather fold that into the extraction change than add a line that gets deleted when the helper moves out — but say the word and I will add it here instead. Happy to file the extraction as its own issue if you want it tracked upstream.

**Verification.** `pytest tests/web/api/test_trigger_rate_limits.py tests/web/api/test_share_rate_limit_endpoints.py` → 39 passed; `ruff format --check`, `ruff check` and `mypy src/xagent/web/services/trigger_rate_limit.py` clean. End-to-end through `remote_ip_from_request`, `::1%AAAA…`, `::1%eth0`, `[fe80::1%eth0]:80` and `a:b:c` all take the peer-address fallback, while `[1.2.3.4]:80`, `203.0.113.7:8080` and `2001:db8::1:8080` behave as before.

src/xagent/web/services/trigger_rate_limit.py
codeacme17: Confirmed and fixed with the stronger option, in 5a2da771.

Reproduced through the public helper before deciding: with `XAGENT_TRUSTED_PROXY_HOPS=1`, `X-Forwarded-For: ::1%` + `"A" * 4000` returned all 4004 characters as the client IP. `git show dc5ef731:src/xagent/web/services/trigger_rate_limit.py` confirms the base called `ipaddress.ip_address(candidate)` directly, so the behaviour is pre-existing and only the docstring claim was new — your scoping was right.

I went with rejection rather than softening the wording because #1124's own comment states the goal as stopping an unvalidated header from injecting "an unbounded arbitrary string", and the zone-id path defeats exactly that bound. `_client_ip_from_forwarded_entry` now returns `None` for any entry containing `%`, before splitting — that covers both the unbracketed fallthrough and the bracket route (`[fe80::1%eth0]:80`, which the strip-only variant would have missed). A zone-id names a local interface, so it is meaningless for a remote client address and no proxy emits one; such an entry now takes the same peer-address fallback as any other non-IP-shaped value. The docstring claim is now true as written.

Three cases added to `test_malformed_port_and_bracket_forms_fall_back_to_peer`: `::1%eth0`, `::1%` + `"A" * 4000`, and `[fe80::1%eth0]:80`.

One calibration, since it affects how this gets triaged rather than what the fix is: I don't think this was a rate-limit bypass. Reaching this code with a client-controlled entry requires `hops` to over-count the real chain, and in that case an attacker can already write arbitrary IPs, so bucket rotation was unbounded regardless. What the zone-id uniquely broke is the length bound — a rotated IP is ~45 characters, a zone-id is unbounded — so this is injection of an unbounded string into Redis key material and the persisted `widget_client_ip`.

src/xagent/web/services/trigger_rate_limit.py
codeacme17: Correct — confirmed `_split_host_port("[1.2.3.4]:80")` returns `("1.2.3.4", "80")`. Did both halves of your suggestion in 5a2da771.

The docstring now reads "a bracketed host" instead of "bracketed IPv6", with a sentence on why the host is not checked for IPv6: RFC 3986 reserves brackets for it, and a bracketed IPv4 yields the same string its bare form would, so it cannot land in a different bucket than its canonical spelling.

`"[1.2.3.4]:80": "1.2.3.4"` is pinned in `test_port_suffixed_entries_resolve_to_the_bare_ip`, with a comment recording that it is accepted-but-odd rather than an intended input form.

Your harmlessness argument holds generally, and I checked the one way it could have failed: since every accepted host must pass `ipaddress.ip_address()` and is returned as written, a bracketed spelling always produces the identical string the bare form would. The single input that *would* have reached a different bucket via this branch was the bracketed zone-id (`[fe80::1%eth0]:80`), which is now rejected by the `%` check from your first comment.

tests/web/api/test_trigger_rate_limits.py
codeacme17: Applied in 5a2da771 — `a:b:c` added to the tuple, with a short comment recording what it pins so a later edit does not drop it as a duplicate of the other malformed entries.

One small correction to the framing: the fallthrough is not only exercised by valid IPv6. Zero-colon inputs reach it too — `not-an-ip`, `"a" * 5000` and `203.0.113.7 evil` all take `return candidate, None`. But the specific gap you identified was real: no forged case had two or more colons, because `evil:8080` has exactly one and takes the port branch. So the branch was covered for the 0-colon shape and the valid-IPv6 shape, and uncovered for the malformed multi-colon shape, which is the case a later change to the split condition would most plausibly break.

Verified the current behaviour before pinning it: `_split_host_port("a:b:c")` returns `("a:b:c", None)`, `ipaddress.ip_address("a:b:c")` raises, and the entry falls back to the peer address with the expected warning log.

PR #1227, opened 2026-08-10
fix: set no-store Cache-Control on public inline preview responses

### LINKED ISSUE #1226: Public preview responses lack Cache-Control, allowing tokened media to persist in browser disk cache
## Problem

Public/share inline file previews are served with no `Cache-Control` header, so a tokened preview URL and its response body can be written to the browser's disk cache automatically — on every render, with no user action — and persist past the logical end of a guest session, bounded only by the guest token's 30-day expiry.

This is distinct from (and additive to) the deliberate scope expansion in PR #1196: that PR made image/audio/video elements load the tokened public preview URL directly instead of blob-fetching, to preserve HTTP range requests for progressive playback. The PR description argues this is an acceptable extension of an already-existing exposure pattern (tokened URLs already appear in `<a href>` for downloads/office-preview links). That argument covers the URL appearing in DOM attributes and page HTML, but not the two axes below, which this issue is about:

1. **Trigger**: an `<a href>` requires a user click; ``/`<video>`/`<audio>` `src` fetches automatically on render.
2. **Persistence**: with no `Cache-Control` header, the response can land in browser disk cache and outlive the page/session, whereas a click-triggered download is a one-off fetch.

## Where

`_inline_preview_response` / `public_preview_file` in `src/xagent/web/api/files.py` (~lines 184-228) build the `FileResponse` for inline previews. Current headers:

```python
headers={
    "Content-Disposition": "inline",
    "X-Content-Type-Options": "nosniff",
},
```

No `Cache-Control` is set on either the SVG-rasterized branch or the generic branch.

## Suggested direction

Add `Cache-Control: private, no-store` (or similar) to the public/tokened preview response so it's never written to disk cache. Two things worth checking before changing this:

- **This route is shared.** The in-app (authenticated) image preview path also goes through `_inline_preview_response` via `/api/files/preview/{id}`, not just the public/tokened route. Adding `no-store` unconditionally would make every chat image re-fetch on every render instead of using the browser cache, which is a real performance cost for the common case. The fix likely needs to be conditional — e.g. only when serving through the public/token-authenticated code path (`public_preview_file`), not the authenticated one.
- Confirm this doesn't regress the range-request behavior PR #1196 relies on for progressive audio/video playback — `Cache-Control: no-store` and `Accept-Ranges`/206 responses are orthogonal, but worth a quick check with an actual video file.

## References

- PR #1196 (introduced the direct-URL behavior on the public policy): https://github.com/xorbitsai/xagent/pull/1196
- Review finding F4, round 4: https://github.com/xorbitsai/xagent/pull/1196#pullrequestreview (see review body, "F4 — Public-surface preview responses set no cache headers")

yiboyasss:
Fixes #1226.

## Problem

Since #1196, public/share inline previews are loaded directly as ``/`<audio>`/`<video>` `src` on public surfaces — a URL that fetches automatically on render, with no user action, unlike a click-triggered download link. With no `Cache-Control` header, that tokened URL and its response body could be written to browser disk cache and outlive the guest session, bounded only by the token's 30-day expiry.

## Fix

- `_inline_preview_response` (`src/xagent/web/api/files.py`) gains an optional `extra_headers` parameter, merged into the response headers without changing any other caller's behavior.
- `public_preview_file`'s two call sites now pass `Cache-Control: private, no-store` via a new `_PUBLIC_PREVIEW_CACHE_HEADERS` constant.
- The authenticated in-app preview route (`preview_file`) is deliberately untouched — it shares the same helper, and chat images there rely on ordinary browser caching across repeated renders of the same message.

## Testing

Two new tests in `tests/web/test_admin_file_access.py`:
- `test_public_preview_sets_no_store_cache_control` — asserts the header on the public route.
- `test_authenticated_preview_keeps_default_caching` — asserts the header is *not* set on the authenticated route (regression guard against the shared-helper accidentally leaking the override).

`ruff`, `mypy`, and the full `tests/web/test_admin_file_access.py` + `tests/web/test_file_upload.py` suites pass (3 pre-existing `cairosvg`-dependent failures unrelated to this change, confirmed against a clean `upstream/main` checkout).

### DIFF
--- a/src/xagent/web/api/files.py
+++ b/src/xagent/web/api/files.py
@@ -187,12 +187,19 @@ async def _inline_preview_response(
     filename: str,
     media_type: str,
     file_id: Optional[str] = None,
+    extra_headers: Optional[Dict[str, str]] = None,
 ) -> FileResponse:
     """Build the final inline preview FileResponse, rasterizing SVG to PNG.
 
     Raw SVG bytes are never served inline — an embedded ``<script>`` would
     execute on direct top-level navigation to this response. Every other
     media type is served as-is.
+
+    ``extra_headers`` lets a caller layer route-specific headers (e.g. the
+    public/tokened route's ``Cache-Control``) without changing behavior for
+    other callers — this helper is shared by both the authenticated and
+    public preview endpoints, and the authenticated in-app path relies on
+    normal browser caching for repeatedly-rendered chat images.
     """
 
     if media_type == "image/svg+xml":
@@ -215,6 +222,7 @@ async def _inline_preview_response(
             headers={
                 "Content-Disposition": "inline",
                 "X-Content-Type-Options": "nosniff",
+                **(extra_headers or {}),
             },
         )
     return FileResponse(
@@ -224,8 +232,19 @@ async def _inline_preview_response(
         headers={
             "Content-Disposition": "inline",
             "X-Content-Type-Options": "nosniff",
+            **(extra_headers or {}),
         },
     )
+
+
+# The public/tokened preview route (unlike the authenticated in-app one) is
+# now loaded directly as <img>/<audio>/<video> src on public share surfaces
+# (#1196), which fetches automatically on render with no user action. A
+# response with no Cache-Control lets the browser write the tokened URL and
+# its bytes to disk cache, where they can outlive the guest session. This
+# header must NOT be applied to the authenticated preview route: chat images
+# there rely on ordinary browser caching across repeated renders.
+_PUBLIC_PREVIEW_CACHE_HEADERS = {"Cache-Control": "private, no-store"}
 
 
 def _durable_redirect_response(
@@ -1724,6 +1743,7 @@ async def public_preview_file(
                 filename=str(file_record.filename),
                 media_type=guess_media_type(str(file_record.filename)),
                 file_id=file_id if is_valid_uuid(file_id) else None,
+                extra_headers=_PUBLIC_PREVIEW_CACHE_HEADERS,
             )
     else:
         # Try to resolve as legacy path across all user directories
@@ -1771,6 +1791,7 @@ async def public_preview_file(
         filename=target_path.name,
         media_type=guess_media_type(target_path.name),
         file_id=file_id if is_valid_uuid(file_id) else None,
+        extra_headers=_PUBLIC_PREVIEW_CACHE_HEADERS,
     )
 
 
--- a/src/xagent/web/api/files.py
+++ b/src/xagent/web/api/files.py
@@ -220,9 +220,13 @@ async def _inline_preview_response(
             filename=Path(filename).with_suffix(".png").name,
             media_type="image/png",
             headers={
+                **(extra_headers or {}),
+                # These two win over any same-named extra_headers entry: they
+                # are security invariants (the second prevents this rasterized
+                # PNG from being sniffed back into script-executable content),
+                # not caller-tunable behavior.
                 "Content-Disposition": "inline",
                 "X-Content-Type-Options": "nosniff",
-                **(extra_headers or {}),
             },
         )
     return FileResponse(
@@ -230,9 +234,10 @@ async def _inline_preview_response(
         filename=filename,
         media_type=media_type,
         headers={
+            **(extra_headers or {}),
+            # See the invariant note in the SVG branch above.
             "Content-Disposition": "inline",
             "X-Content-Type-Options": "nosniff",
-            **(extra_headers or {}),
         },
     )
 
--- a/tests/web/test_admin_file_access.py
+++ b/tests/web/test_admin_file_access.py
@@ -9,6 +9,7 @@ from fastapi.testclient import TestClient
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
+import xagent.web.api.files as files_module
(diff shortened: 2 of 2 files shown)


src/xagent/web/api/files.py
```diff
@@ -215,6 +222,7 @@ async def _inline_preview_response(
             headers={
                 "Content-Disposition": "inline",
                 "X-Content-Type-Options": "nosniff",
+                **(extra_headers or {}),
```
rogercloud: `extra_headers` is spread after `Content-Disposition` and `X-Content-Type-Options`, so a caller passing either key would silently override them — and the docstring frames both as security invariants (they prevent raw SVG script execution on direct navigation). Not a live bug (only `Cache-Control` is ever passed today), but worth closing off. Suggest spreading first and re-asserting the invariants last, here and in the second block below:

```python
headers={
    **(extra_headers or {}),
    "Content-Disposition": "inline",
    "X-Content-Type-Options": "nosniff",
},
```

tests/web/test_admin_file_access.py
```diff
@@ -485,3 +485,73 @@ def test_public_preview_blocks_parent_traversal(self, test_db, temp_uploads_dir)
         )
 
         assert response.status_code == 403
+
+    def test_public_preview_sets_no_store_cache_control(
```
rogercloud, line 490: This uploads `clip.txt`, which resolves to `text/plain`, so it only covers the generic `FileResponse` branch. The identical `extra_headers` spread was also added to the SVG-rasterization branch, which currently has no assertion that `Cache-Control` reaches the rasterized PNG. Consider a third case with an `.svg` fixture through the public route asserting `cache-control == "private, no-store"`.

tests/web/test_admin_file_access.py
```diff
@@ -485,3 +485,73 @@ def test_public_preview_blocks_parent_traversal(self, test_db, temp_uploads_dir)
...
+        )
+
+        client = TestClient(test_app)
+        user_headers = create_auth_headers(regular_user)
+        response = client.get(
+            f"/api/files/preview/{uploaded_file.file_id}", headers=user_headers
+        )
+
+        assert response.status_code == 200
+        assert response.headers.get("cache-control") != "private, no-store"
```
rogercloud: This only excludes one exact string — a regression that set some other `Cache-Control` value here would still pass. The header is absent today (Starlette's `FileResponse` sets no default), so the stronger assertion holds:

```python
assert "cache-control" not in response.headers
```

rogercloud (approved):
## Summary

This PR closes the caching gap from #1226: public/share inline previews are fetched automatically as `/<audio>/<video>` sources, so without a `Cache-Control` header the tokened URL and the response bytes could be written to the browser's disk cache and survive well past the guest session (bounded only by the 30-day token TTL). The fix adds an optional `extra_headers` parameter to `_inline_preview_response` in `src/xagent/web/api/files.py` and has both content-serving exits of `public_preview_file` pass `Cache-Control: private, no-store`. The authenticated in-app preview route is intentionally left on default caching, and two tests cover the split.

## Approach

The approach is sound and scoped correctly. Applying the header at the shared helper via an opt-in parameter — rather than unconditionally inside the helper or via middleware — is the right layer, and it matches the direction the linked issue suggested. Specific confirmations:

- Both public exits that actually serve bytes (the early return and the terminal return in `public_preview_file`) pass the header; every other exit in that function raises `HTTPException`, so there is no uncovered path.
- All four call sites of `_inline_preview_response` are accounted for: the two authenticated ones stay unchanged, the two public ones get the header. There is no other route or mount serving these previews.
- No range-request regression: headers passed through `FileResponse(headers=...)` are carried onto both 200 and 206 responses, so the progressive audio/video playback added in #1196 is unaffected.
- `Content-Disposition: inline` still wins over Starlette's filename-derived default, which uses `setdefault`.
- Deliberately keeping default caching on the authenticated route is the right call, and the intent is pinned by a test.

The "auto-fetch vs. click" trigger axis raised in #1226 is an inherent property of #1196 and correctly out of scope here.

## Findings

All findings below are minor; none block the merge (see inline comments for details).

1. `src/xagent/web/api/files.py:225,235` — `extra_headers` is spread after the invariant headers (`Content-Disposition`, `X-Content-Type-Options`), so a future caller passing either key would silently override them.
2. `tests/web/test_admin_file_access.py:489` — the SVG-rasterization branch of `_inline_preview_response` has no test coverage for the Cache-Control header (both new tests only exercise the generic branch).
3. `tests/web/test_admin_file_access.py:557` — the negative assertion `!= "private, no-store"` is weaker than intended; a regression setting some other Cache-Control value would still pass.
4. Non-blocking follow-up: `public_download_file` (`src/xagent/web/api/files.py`) also serves tokened content and has neither `Cache-Control` nor `X-Content-Type-Options`. It's click-triggered (an `<a href>` link) so legitimately out of this PR's scope, but the same tokened bytes can land in disk cache after a click — worth a follow-up issue.

## Tests

- `tests/web/test_admin_file_access.py`: 11 passed, including the two new cases.
- `tests/web/test_file_upload.py` has 3 failures that are environment-only (missing native `cairo` library in the test sandbox) and unrelated to this change.

## Recommendation

Approving — no blocking issues. Finding 1 is worth applying before merge as cheap hardening; findings 2 and 3 are small test improvements; finding 4 is a follow-up for a separate issue.

PR #1236, opened 2026-08-10
feat: add task runtime UI extension slots

qinxuye:
## What this PR does

- adds one build-time replacement module for task-runtime UI contributions
- exposes composer menu and selected-state slots that can contribute runtime extension configuration without replacing `ChatInput`
- exposes a Settings slot without replacing the Settings page
- exposes a user-message metadata renderer slot without replacing `ChatMessage` or the conversation panel
- keeps the OSS implementation inert while preserving the built-in Local browser flow

## Boundary

This is an extensibility-only PR. It does **not** add Browser Relay, a Chrome extension, pairing, WebSocket or Redis transport, Desktop Relay, a macOS companion, or any closed-source runtime provider. It also does not change backend task-runtime lifecycle or persistence.

The purpose is to let composed distributions supply those UIs from an overlay without copying core frontend files.

## Validation

- `npm run type-check`
- `npm run lint`
- focused Vitest suite: 73 tests passed
- `pre-commit run --files ...` for all changed files

### DIFF
--- a/frontend/src/components/chat/ChatInput.tsx
+++ b/frontend/src/components/chat/ChatInput.tsx
@@ -591,6 +591,9 @@ export function ChatInput({
   const showTaskRuntimeExtension =
     hasTaskRuntimeComposerExtension && !readOnlyConfig && !hideConfig;
   const activeLocalBrowserTarget = showLocalBrowser ? localBrowserTarget : null;
+  const activeTaskRuntimeSelection = showTaskRuntimeExtension
+    ? taskRuntimeSelection
+    : null;
   useEffect(() => {
     if (!showLocalBrowser) setLocalBrowserTarget(null);
   }, [showLocalBrowser]);
@@ -699,7 +702,7 @@ export function ChatInput({
       const deliveryKey = JSON.stringify([
         messageToSend,
         activeLocalBrowserTarget,
-        taskRuntimeSelection?.runtimeExtensions || null,
+        activeTaskRuntimeSelection?.runtimeExtensions || null,
         enabledFiles.map((file) => [
           file.name,
           file.size,
@@ -715,13 +718,13 @@ export function ChatInput({
       const configToSend = {
         ...agentConfig,
         clientMessageId,
-        ...(activeLocalBrowserTarget || taskRuntimeSelection
+        ...(activeLocalBrowserTarget || activeTaskRuntimeSelection
           ? {
               runtimeExtensions: {
                 ...(agentConfig.runtimeExtensions || {}),
                 ...(activeLocalBrowserTarget
                   ? { local_browser: { ...activeLocalBrowserTarget } }
-                  : taskRuntimeSelection?.runtimeExtensions || {}),
+                  : activeTaskRuntimeSelection?.runtimeExtensions || {}),
               },
             }
           : {}),
--- a/frontend/src/components/chat/ChatInput.tsx
+++ b/frontend/src/components/chat/ChatInput.tsx
@@ -17,6 +17,10 @@ import { toast } from "@/components/ui/sonner";
 import { useVoiceInputControls } from "@/components/voice-input-controller";
 import { LocalBrowserMenu, type LocalBrowserTarget } from "./LocalBrowserMenu";
 import {
+  hasTaskRuntimeComposerExtension,
+  type TaskRuntimeComposerSelection,
+} from "@/lib/task-runtime-ui-extension";
+import {
   AlertDialog,
   AlertDialogAction,
   AlertDialogCancel,
@@ -163,6 +167,8 @@ export function ChatInput({
   const [isDraggingFiles, setIsDraggingFiles] = useState(false);
   const [localBrowserTarget, setLocalBrowserTarget] =
     useState<LocalBrowserTarget | null>(null);
+  const [taskRuntimeSelection, setTaskRuntimeSelection] =
+    useState<TaskRuntimeComposerSelection | null>(null);
 
   const fileInputRef = useRef<HTMLInputElement>(null);
   const editorRef = useRef<HTMLDivElement>(null);
@@ -582,10 +588,15 @@ export function ChatInput({
     !allowsLiveGuidanceInput &&
     !isStoppedTaskStatus(normalizedTaskStatus);
   const showLocalBrowser = Boolean(user?.is_admin) && !readOnlyConfig && !hideConfig;
+  const showTaskRuntimeExtension =
+    hasTaskRuntimeComposerExtension && !readOnlyConfig && !hideConfig;
   const activeLocalBrowserTarget = showLocalBrowser ? localBrowserTarget : null;
   useEffect(() => {
     if (!showLocalBrowser) setLocalBrowserTarget(null);
   }, [showLocalBrowser]);
+  useEffect(() => {
+    if (!showTaskRuntimeExtension) setTaskRuntimeSelection(null);
+  }, [showTaskRuntimeExtension]);
   const voiceInputLabel =
     voiceInput.status === "recording"
       ? t("voiceInput.stop")
@@ -688,6 +699,7 @@ export function ChatInput({
       const deliveryKey = JSON.stringify([
         messageToSend,
         activeLocalBrowserTarget,
+        taskRuntimeSelection?.runtimeExtensions || null,
         enabledFiles.map((file) => [
           file.name,
           file.size,
@@ -703,11 +715,13 @@ export function ChatInput({
       const configToSend = {
         ...agentConfig,
         clientMessageId,
-        ...(activeLocalBrowserTarget
+        ...(activeLocalBrowserTarget || taskRuntimeSelection
           ? {
               runtimeExtensions: {
                 ...(agentConfig.runtimeExtensions || {}),
-                local_browser: { ...activeLocalBrowserTarget },
+                ...(activeLocalBrowserTarget
+                  ? { local_browser: { ...activeLocalBrowserTarget } }
+                  : taskRuntimeSelection?.runtimeExtensions || {}),
               },
             }
           : {}),
@@ -724,6 +738,7 @@ export function ChatInput({
       await onSend(messageToSend, configToSend);
       deliveryAttemptRef.current = null;
       setLocalBrowserTarget(null);
+      setTaskRuntimeSelection(null);
       fileMention.resetMention();
 
       if (isControlled) {
@@ -1120,7 +1135,9 @@ export function ChatInput({
                   </>
                 )}
                 {/* Add files or bind this new task to a local host window. */}
-                {(!hideFileUpload && !filesDisabled) || showLocalBrowser ? (
+                {(!hideFileUpload && !filesDisabled)
+                  || showLocalBrowser
+                  || showTaskRuntimeExtension ? (
                   <>
                     {!hideFileUpload && !filesDisabled && (
                       <input
@@ -1136,12 +1153,15 @@ export function ChatInput({
                       disabled={isInputBusy}
                       selectedTarget={localBrowserTarget}
                       onTargetChange={setLocalBrowserTarget}
+                      extensionSelection={taskRuntimeSelection}
+                      onExtensionSelectionChange={setTaskRuntimeSelection}
                       onAddFiles={
                         !hideFileUpload && !filesDisabled
                           ? () => fileInputRef.current?.click()
                           : undefined
                       }
                       showLocalBrowser={showLocalBrowser}
+                      showTaskRuntimeExtension={showTaskRuntimeExtension}
                     />
                   </>
                 ) : null}
--- a/frontend/src/components/chat/ChatInput.tsx
+++ b/frontend/src/components/chat/ChatInput.tsx
@@ -699,10 +699,12 @@ export function ChatInput({
       const trimmed = message.trim();
       const messageToSend = trimmed;
       const executionMode = taskConfig?.executionMode;
+      const extraRuntimeExtensions = activeLocalBrowserTarget
+        ? { local_browser: { ...activeLocalBrowserTarget } }
+        : activeTaskRuntimeSelection?.runtimeExtensions;
       const deliveryKey = JSON.stringify([
         messageToSend,
-        activeLocalBrowserTarget,
-        activeTaskRuntimeSelection?.runtimeExtensions || null,
+        extraRuntimeExtensions || null,
         enabledFiles.map((file) => [
           file.name,
           file.size,
@@ -718,13 +720,11 @@ export function ChatInput({
       const configToSend = {
         ...agentConfig,
         clientMessageId,
-        ...(activeLocalBrowserTarget || activeTaskRuntimeSelection
+        ...(extraRuntimeExtensions
           ? {
               runtimeExtensions: {
                 ...(agentConfig.runtimeExtensions || {}),
-                ...(activeLocalBrowserTarget
-                  ? { local_browser: { ...activeLocalBrowserTarget } }
-                  : activeTaskRuntimeSelection?.runtimeExtensions || {}),
+                ...extraRuntimeExtensions,
               },
             }
           : {}),
--- a/frontend/src/components/chat/ChatMessage.test.tsx
+++ b/frontend/src/components/chat/ChatMessage.test.tsx
@@ -181,6 +181,26 @@ describe("ChatMessage Session file capability", () => {
     ).toBeInTheDocument()
   })
 
+  it("does not render distribution message metadata on assistant messages", () => {
+    render(
+      <ChatMessage
+        role="assistant"
+        content="I inspected your browser"
+        taskRuntimeExtensionMetadata={{
+          bindings: ["browser_relay"],
+          publicMetadata: {
+            browser_relay: { kind: "browser_relay", connected: true },
+          },
+        }}
+      />,
+    )
+
+    expect(screen.getByText("I inspected your browser")).toBeInTheDocument()
+    expect(
+      screen.queryByRole("note", { name: "Computer use · My browser" }),
+    ).not.toBeInTheDocument()
+  })
+
   it("uses the same safe projection for normal assistant DOM and copied content", () => {
     appContextMock.filesDisabled = true
     const { container } = render(
--- a/frontend/src/components/chat/ChatMessage.test.tsx
+++ b/frontend/src/components/chat/ChatMessage.test.tsx
@@ -38,6 +38,21 @@ vi.mock("@/lib/api-wrapper", () => ({
   apiRequest: apiRequestMock,
 }))
 
+vi.mock("@/lib/task-runtime-ui-extension", async () => {
+  const ReactModule = await vi.importActual<typeof import("react")>("react")
+  return {
+    TaskRuntimeMessageMetadataExtension: ({ bindings }: { bindings: string[] }) => (
+      bindings.includes("browser_relay")
+        ? ReactModule.createElement(
+          "div",
+          { role: "note", "aria-label": "Computer use · My browser" },
+          "My browser",
+        )
+        : null
+    ),
+  }
+})
+
 vi.mock("@/components/file/docx-preview-renderer", () => ({
   DocxPreviewRenderer: ({ base64Content }: { base64Content: string }) => (
     <div data-testid="docx-preview">{base64Content}</div>
@@ -143,6 +158,26 @@ describe("ChatMessage Session file capability", () => {
 
     expect(
       screen.getByRole("note", { name: "Computer use · Local browser" }),
+    ).toBeInTheDocument()
+  })
+
+  it("renders distribution message metadata without replacing the message", () => {
+    render(
+      <ChatMessage
+        role="user"
+        content="Inspect my browser"
+        taskRuntimeExtensionMetadata={{
+          bindings: ["browser_relay"],
+          publicMetadata: {
+            browser_relay: { kind: "browser_relay", connected: true },
+          },
+        }}
+      />,
+    )
+
+    expect(screen.getByText("Inspect my browser")).toBeInTheDocument()
+    expect(
+      screen.getByRole("note", { name: "Computer use · My browser" }),
     ).toBeInTheDocument()
   })
 
--- a/frontend/src/components/chat/ChatMessage.test.tsx
+++ b/frontend/src/components/chat/ChatMessage.test.tsx
@@ -39,8 +39,12 @@ vi.mock("@/lib/api-wrapper", () => ({
 }))
 
 vi.mock("@/lib/task-runtime-ui-extension", async () => {
+  const actual = await vi.importActual<
+    typeof import("@/lib/task-runtime-ui-extension")
+  >("@/lib/task-runtime-ui-extension")
   const ReactModule = await vi.importActual<typeof import("react")>("react")
   return {
+    ...actual,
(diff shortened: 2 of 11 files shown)


frontend/src/components/task/task-conversation-panel.tsx
```diff
@@ -771,6 +771,10 @@ export function TaskConversationPanel({
                         interactionsActive={item.id === activeWaitingMessageId}
                         showEmptyStatus={item.showEmptyStatus}
                         contextBadges={item.role === "user" ? userMessageContextBadges : undefined}
+                        taskRuntimeExtensionMetadata={item.role === "user" ? {
+                          bindings: state.currentTask?.runtimeExtensionBindings || [],
+                          publicMetadata: state.taskRuntimeExtensions || {},
```
rogercloud, line 779: **Major — stale `currentTask` on task navigation.**

This reads `state.currentTask?.runtimeExtensionBindings` with no task-identity guard, but the pre-existing `userMessageContextBadges` logic in this same file (line 226-229) guards the identical field with `state.currentTask?.id === String(state.taskId)`.

`SET_TASK_ID` (`frontend/src/contexts/app-context-chat.tsx:973-992`) clears `taskRuntimeExtensions` on switch but leaves `currentTask` populated until the async refetch lands. So after navigating task A -> task B, a task-B user message rendered in that window gets task-A bindings (with an empty `publicMetadata`), and nothing remounts this subtree by key to prevent it. Suggest applying the same guard as line 226-229.

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -703,11 +715,13 @@ export function ChatInput({
       const configToSend = {
         ...agentConfig,
         clientMessageId,
-        ...(activeLocalBrowserTarget
+        ...(activeLocalBrowserTarget || taskRuntimeSelection
```
rogercloud: **Minor (latent) — extension selection is silently dropped, not merged, when a local-browser target is set.**

This spread is either/or: if `activeLocalBrowserTarget` is truthy, `taskRuntimeSelection?.runtimeExtensions` never reaches the payload. That's fine as long as the two are mutually exclusive, and `LocalBrowserMenu.tsx:250-252` and `:293-296` do enforce that — but the third site, `LocalBrowserMenu.tsx:333-339` (the persistent selected-state chip rendered outside the popover), passes `onExtensionSelectionChange` through raw with no `onTargetChange(null)`.

Unreachable today since the OSS `TaskRuntimeComposerSelectionExtension` renders `null`, but once a distribution ships a real selection chip, a user with a bound browser window who re-picks a runtime via that chip loses the new selection at submit with no feedback. Since this is a documented extension point rather than internal detail, worth mirroring the exclusion at the third call site now.

frontend/src/components/chat/LocalBrowserMenu.tsx
```diff
@@ -305,6 +330,14 @@ export function LocalBrowserMenu({
           </button>
         </div>
       )}
+      {showTaskRuntimeExtension && (
+        <TaskRuntimeComposerSelectionExtension
+          disabled={disabled}
+          selection={extensionSelection}
```
rogercloud, line 336: **Minor — `onRequestClose` is dead on this slot.**

This render site is outside any `Popover`; the sibling slot at lines 289-299 is inside `<PopoverContent>`, where `onRequestClose={() => setOpen(false)}` actually dismisses something. Both slots share `TaskRuntimeComposerExtensionProps`, so this call site is forced to supply a callback that can never do anything. Harmless with the inert stub, but a future implementation that calls `onRequestClose()` expecting a dismiss will silently no-op here. Consider dropping it from the chip variant (or splitting the two prop types).

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -688,6 +699,7 @@ export function ChatInput({
       const deliveryKey = JSON.stringify([
         messageToSend,
         activeLocalBrowserTarget,
```
rogercloud: **Nit (defensive consistency, not a reachable bug).**

`activeLocalBrowserTarget` is recomputed synchronously each render from `showLocalBrowser`, but `taskRuntimeSelection` is only cleared by an async `useEffect`, so it trails `showTaskRuntimeExtension` by one render. Traced both `readOnlyConfig` and `hideConfig` and neither can flip in the same tick as a submit, so there's no reachable stale-send today. For symmetry with the pattern right above, consider `const activeTaskRuntimeSelection = showTaskRuntimeExtension ? taskRuntimeSelection : null` and using that at line 718.

frontend/src/components/chat/LocalBrowserMenu.tsx
```diff
@@ -41,16 +46,24 @@ interface LocalBrowserMenuProps {
   disabled: boolean;
   selectedTarget: LocalBrowserTarget | null;
   onTargetChange: (target: LocalBrowserTarget | null) => void;
```
rogercloud, line 48: **Nit —** `extensionSelection`, `onExtensionSelectionChange`, and `showTaskRuntimeExtension` are required, while the adjacent `onAddFiles?` is optional. There is exactly one caller today (`ChatInput.tsx:1152`) so nothing breaks; making them optional would just keep this component's prop surface easier to reuse.

frontend/src/components/chat/ChatMessage.test.tsx
```diff
@@ -146,6 +161,26 @@ describe("ChatMessage Session file capability", () => {
     ).toBeInTheDocument()
   })
 
+  it("renders distribution message metadata without replacing the message", () => {
```
rogercloud, line 168: **Minor — the `isUser` gate is untested.** `ChatMessage.tsx:534-537` renders the metadata slot only when `isUser && taskRuntimeExtensionMetadata`, but this test uses `role="user"` only. A companion case with `role="assistant"` plus `taskRuntimeExtensionMetadata` set, asserting the slot is *not* rendered, would protect the gate against a refactor that drops or inverts it.

frontend/src/lib/task-runtime-ui-extension.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+} from "./task-runtime-ui-extension";
+
+const composerProps = {
+  disabled: false,
+  selection: null,
+  onSelectionChange: vi.fn(),
+  onRequestClose: vi.fn(),
+};
+
+describe("default task runtime UI extension", () => {
```
rogercloud, line 20: **Nit — assertion granularity.** This does exercise the real module (it would fail if an export were removed or a stub started rendering non-null), so it is not tautological. But all four component exports are rendered in one `it()` behind a single combined `toBeEmptyDOMElement()`, so a failure won't say which slot regressed. The paired precedent `frontend/src/app/build/page-default-extension.test.tsx` asserts each export separately — worth matching.

rogercloud (commented):
## Summary

This PR adds four build-time UI extension slots for task runtime (composer menu, composer selection chip, message metadata, settings) via a new module `frontend/src/lib/task-runtime-ui-extension.tsx`, wired into `ChatInput`, `ChatMessage`, `LocalBrowserMenu`, `task-conversation-panel`, and `settings/page.tsx`. OSS defaults are inert (render `null`), so this PR causes no behavior change in this repo — it exists so a downstream/closed distribution can replace the module. CI is green.

## Findings

### Major

1. `frontend/src/components/task/task-conversation-panel.tsx:774-777` — stale `currentTask` on task navigation. The new code reads `state.currentTask?.runtimeExtensionBindings` with no task-identity guard, unlike the adjacent pre-existing `userMessageContextBadges` logic at lines 226-229, which guards the same field with `state.currentTask?.id === String(state.taskId)`. `SET_TASK_ID` (`frontend/src/contexts/app-context-chat.tsx:973-992`) clears `taskRuntimeExtensions` on switch but leaves `currentTask` populated until an async refetch resolves, so after navigating task A -> task B there is a window where a task-B message renders with task-A bindings (while `publicMetadata` is correctly empty). Nothing remounts this subtree by key, so the window is reachable. Recommend applying the same guard.

### Minor

2. `frontend/src/components/chat/ChatInput.tsx:718-727` — extension selection silently dropped when a local-browser target is also set. The spread is either/or (`activeLocalBrowserTarget ? { local_browser: ... } : taskRuntimeSelection?.runtimeExtensions || {}`). Mutual exclusion is enforced at `LocalBrowserMenu.tsx:250-252` and `:293-296`, but not at the third site, `LocalBrowserMenu.tsx:333-339` (the persistent selected-state chip), which passes `onExtensionSelectionChange` through with no accompanying `onTargetChange(null)`. Unreachable today because the OSS selection extension renders `null`, but once a distribution ships a real chip, re-picking a runtime while a browser window is bound will drop the new selection at submit with no feedback. Worth mirroring the exclusion at the third call site now.

3. `frontend/src/components/chat/LocalBrowserMenu.tsx:333-339` — `onRequestClose` is dead on the selected-state chip slot. The sibling slot at `:289-299` lives inside `<PopoverContent>` where `onRequestClose={() => setOpen(false)}` is meaningful; this one renders outside any `Popover`, so the callback can never do anything. Both share `TaskRuntimeComposerExtensionProps`. Harmless with the inert stub, but a future implementation that calls it expecting a dismiss will silently no-op.

4. `frontend/src/components/chat/ChatInput.tsx:701-702, 718` — no synchronous guard for `taskRuntimeSelection`. `activeLocalBrowserTarget = showLocalBrowser ? localBrowserTarget : null` is recomputed every render, whereas `taskRuntimeSelection` is only cleared by an async `useEffect`, one render behind `showTaskRuntimeExtension` flipping false. Could not find a reachable path (both `readOnlyConfig` and `hideConfig` only change on discrete user actions or static props, never in the same tick as a submit), so this is defensive consistency rather than a bug — optional `const activeTaskRuntimeSelection = showTaskRuntimeExtension ? taskRuntimeSelection : null` for symmetry.

5. `frontend/src/components/chat/LocalBrowserMenu.tsx:45-56` — three new props are required rather than optional (`extensionSelection`, `onExtensionSelectionChange`, `showTaskRuntimeExtension`), unlike the adjacent `onAddFiles?`. Only one caller today (`ChatInput.tsx:1152`), so nothing breaks; flagging as a nit.

6. `frontend/src/components/chat/ChatInput.runtime-extension.test.tsx` — two coverage gaps. No test sets a local-browser target and a `taskRuntimeSelection` at the same time (which is what would pin down the precedence in finding 2), and no test asserts `taskRuntimeSelection` resets to `null` after a successful send — `ChatInput.tsx:740-741` calls `setTaskRuntimeSelection(null)`, but neither this file nor the pre-existing `ChatInput.test.tsx` covers it.

7. `frontend/src/components/chat/ChatMessage.test.tsx:164-182` — the `isUser` gate is untested. `ChatMessage.tsx:534-537` gates the slot on `isUser && taskRuntimeExtensionMetadata && (...)`, but the new test only uses `role="user"`. An assistant-role negative case would protect the gate against a future refactor.

8. `frontend/src/lib/task-runtime-ui-extension.test.tsx:20-38` — coarse assertion granularity. The test does exercise the real module (it would fail if an export were removed or a stub started rendering non-null), but it bundles all four component exports into one `it()` with a single combined `toBeEmptyDOMElement()`, whereas the paired precedent `frontend/src/app/build/page-default-extension.test.tsx` asserts each export separately. Splitting per-export would make a failure identify which slot broke.

A few other angles were checked and deliberately not raised: the `publicMetadata` naming/redaction question (consistent with the existing `public_metadata()` provider hook in `src/xagent/core/task_runtime.py`, where redaction is the provider's responsibility); the inert-composer negative case (already covered by the untouched `ChatInput.test.tsx:336-354`); and the `false`-literal inference on `hasTaskRuntimeComposerExtension` (no lint/compiler diagnostic, and it re-widens to `boolean` at the `LocalBrowserMenu` prop boundary).

## Testing

Static review only. `frontend/node_modules` is not installed in the review environment, so `npm run type-check`, `npm run lint`, and vitest could not be re-run — the "73 tests passed" claim in the PR description was not independently reproduced. CI checks on the PR (frontend-build, e2e) are green.

## Recommendation

Non-blocking: only finding 1 is a real reachable defect, and it is a one-line guard matching an existing pattern in the same file; everything else is either latent behind the inert OSS defaults or test-coverage polish.

frontend/src/components/task/task-conversation-panel.tsx
qinxuye: Fixed in 7fe84cfc. Runtime bindings are now accepted only when currentTask.id matches taskId, matching the existing context-badge guard. The task-navigation regression test also asserts that stale runtime metadata is not reused.

frontend/src/components/chat/ChatInput.tsx
qinxuye: Applied in 7fe84cfc. Submission and delivery deduplication now use a synchronously gated activeTaskRuntimeSelection, mirroring activeLocalBrowserTarget.

frontend/src/lib/task-runtime-ui-extension.test.tsx
qinxuye: Applied in 7fe84cfc. Each inert OSS slot now has an independent test, so regressions identify the exact extension point.

frontend/src/components/chat/ChatMessage.test.tsx
qinxuye: Added in 7fe84cfc. The assistant-role negative case verifies that distribution runtime metadata is rendered only on user messages.

qinxuye (commented):
Addressed the current review feedback in 7fe84cfc.

frontend/src/components/chat/LocalBrowserMenu.tsx
```diff
@@ -236,6 +249,7 @@ export function LocalBrowserMenu({
                           disabled={localBrowserDisabled}
                           onClick={() => {
                             onTargetChange(browserWindow);
+                            onExtensionSelectionChange(null);
```
rogercloud, line 252: Minor: this `onExtensionSelectionChange(null)` guard (window-pick site) is untested — mutation-testing it (deleting the line) leaves the full suite green. If it regresses, `ChatInput.tsx`'s `configToSend` merge would silently prefer `local_browser` and drop the extension's `runtimeExtensions` once a distribution ships a non-inert extension. Suggest a test that picks a local-browser window while an extension selection is active and asserts the extension selection clears.

frontend/src/components/chat/LocalBrowserMenu.tsx
```diff
@@ -272,6 +286,17 @@ export function LocalBrowserMenu({
               </PopoverContent>
             </Popover>
           )}
+          {showTaskRuntimeExtension && (
+            <TaskRuntimeComposerMenuExtension
+              disabled={disabled}
+              selection={extensionSelection}
+              onSelectionChange={(selection) => {
+                onExtensionSelectionChange(selection);
+                if (selection) onTargetChange(null);
```
rogercloud, line 295: Minor: same gap as the window-pick site above — `if (selection) onTargetChange(null)` here (in-popover extension-menu site) is also untested; deleting it leaves the suite green. Only the third mutual-exclusion site (the persistent selection chip) is currently covered. Suggest a test that selects a runtime from this menu while a local-browser target is active and asserts the local-browser target clears.

frontend/src/components/chat/ChatMessage.test.tsx
```diff
@@ -38,6 +38,21 @@ vi.mock("@/lib/api-wrapper", () => ({
   apiRequest: apiRequestMock,
 }))
 
+vi.mock("@/lib/task-runtime-ui-extension", async () => {
```
rogercloud, line 41: Minor (test quality): this `vi.mock("@/lib/task-runtime-ui-extension", ...)` hand-builds an object with only the export this file currently uses, omitting the rest of the module's five exports. Same pattern in `app/settings/page.test.tsx:20` and `ChatInput.runtime-extension.test.tsx:64`. Correct today, but if a component later imports another export from this deliberately growing extension-point module, the mock would silently supply `undefined` for it. Suggest spreading the real module via `vi.importActual` and overriding only the export under test, e.g.:
```ts
vi.mock("@/lib/task-runtime-ui-extension", async () => ({
  ...(await vi.importActual<typeof import("@/lib/task-runtime-ui-extension")>(
    "@/lib/task-runtime-ui-extension",
  )),
  // overrides for this test file
}));
```

frontend/src/lib/task-runtime-ui-extension.tsx
```diff
@@ -0,0 +1,53 @@
...
+export type TaskRuntimeComposerSelectionExtensionProps =
+  TaskRuntimeComposerBaseProps;
+
+export interface TaskRuntimeMessageMetadataExtensionProps {
+  bindings: readonly string[];
+  publicMetadata: Readonly<Record<string, Record<string, unknown>>>;
+}
+
+/**
+ * Build-time extension points for distributions that provide additional
```
rogercloud, line 36: Minor (documentation): the docstring states the build-time-replacement intent but doesn't spell out the contract a replacement must honor — the exact export surface, that slots must be null-safe/non-throwing, and precisely what `hasTaskRuntimeComposerExtension` gates. Also, unlike `LocalBrowserMenu.tsx:1`, this module has no `"use client"` directive; harmless today since every consumer is already a client component, but a real overlay implementation using hooks would need it. Suggest extending the docstring with the explicit contract and adding `"use client"` preemptively.

frontend/src/lib/task-runtime-ui-extension.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * task-scoped runtimes. The OSS implementation is deliberately inert. A
+ * composed distribution may replace this module without overriding the
+ * composer, Settings page, or conversation panel themselves.
+ */
+export const hasTaskRuntimeComposerExtension = false;
```
rogercloud: Nit: `hasTaskRuntimeComposerExtension` infers the literal type `false` rather than `boolean`. Cosmetic only — no lint/type-check impact today. Annotating as `: boolean = false` would make the replacement contract explicit at the type level.

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -703,11 +718,13 @@ export function ChatInput({
       const configToSend = {
```
rogercloud, line 720: shrink: the `runtimeExtensions` merge nests a ternary inside an object spread (outer gate on `activeLocalBrowserTarget || activeTaskRuntimeSelection`, inner ternary choosing `local_browser` vs. `activeTaskRuntimeSelection?.runtimeExtensions`). Extracting to a local `extraRuntimeExtensions` computed once above the spread would be identical in every branch and easier to read.

rogercloud (approved):
## Re-review of `7fe84cfc`

### What changed

This PR adds a single build-time-replaceable module, `frontend/src/lib/task-runtime-ui-extension.tsx`, exporting four slot components (composer menu extension, composer selected-state/chip extension, settings extension, user-message metadata extension) plus one boolean capability flag, `hasTaskRuntimeComposerExtension`. Call sites are wired into `ChatInput.tsx`, `LocalBrowserMenu.tsx`, `ChatMessage.tsx`, `task-conversation-panel.tsx`, and `app/settings/page.tsx`, with tests for each slot. In the OSS/default build every slot returns `null` and the flag is `false`, so the shipped product is behaviorally unchanged. The follow-up commit `7fe84cfc` hardens the composer/task-navigation paths and splits the shared prop type.

### Design verdict: sound

The slot shape is minimal — no registry, no dynamic import, no runtime injection; props are one-way data plus callbacks. Behavioral neutrality was verified independently: a full `vitest run` (1918 passed; the single failure is pre-existing and in an unrelated kb test file, untouched by this PR) and `tsc --noEmit` (exit 0).

### Previously-raised findings, now fixed

Verified independently against `7fe84cfc`:

1. Stale `currentTask` on task navigation — `task-conversation-panel.tsx:774-780` now guards `runtimeExtensionBindings` with `state.currentTask?.id === String(state.taskId)`, matching the existing `userMessageContextBadges` pattern, with a regression test (`task-conversation-panel.test.tsx:344-372`) that fails without the guard.
2. The third mutual-exclusion call site (`LocalBrowserMenu.tsx:333-341`) now mirrors the other sites' `if (selection) onTargetChange(null)` guard, covered by `ChatInput.runtime-extension.test.tsx`.
3. The shared prop type carrying a dead `onRequestClose` on the non-popover slot is split into `TaskRuntimeComposerMenuExtensionProps` / `TaskRuntimeComposerSelectionExtensionProps` (`task-runtime-ui-extension.tsx:11-26`).
4. `ChatInput.tsx:593-596` now computes `activeTaskRuntimeSelection` synchronously, mirroring `activeLocalBrowserTarget`.
5. Post-send reset of `taskRuntimeSelection` is now covered by a test assertion.
6. `ChatMessage.test.tsx:184-202` adds a negative test asserting the metadata slot is not rendered for `role="assistant"`.
7. `task-runtime-ui-extension.test.tsx` now asserts each of the four slot exports independently instead of one combined assertion.

Left a few remaining minor/nit items inline below — none of them block. Approving as-is.

### Testing

- Full `vitest run` on `7fe84cfc`: 1918 passed, 1 pre-existing unrelated failure (`src/components/kb/knowledge-base-detail.component.test.tsx`).
- `tsc --noEmit`: exit 0.
- ESLint over the six changed source files: 33 errors / 2 warnings, all pre-existing; none on lines added by this PR.

### Simplification opportunity

`ChatInput.tsx:718-730`: `shrink` — the `runtimeExtensions` merge nests a ternary inside an object spread. Extracting to a local `extraRuntimeExtensions` computed once above the spread would be identical in every branch and easier to read. net: ~5 lines possible.

frontend/src/components/chat/LocalBrowserMenu.tsx
qinxuye: Covered in 1ee63726. The new reverse-direction test starts with the distribution runtime selected, picks a local browser window, and verifies that only local_browser is submitted.

frontend/src/components/chat/LocalBrowserMenu.tsx
qinxuye: Covered in 1ee63726. A separate test starts with a local window bound, selects the distribution runtime from the popover menu, and verifies that the local target is cleared and only browser_relay is submitted.

frontend/src/components/chat/ChatMessage.test.tsx
qinxuye: Applied in 1ee63726. The ChatMessage, Settings, and composer integration tests now spread the real extension module and override only the slot under test.

frontend/src/lib/task-runtime-ui-extension.tsx
qinxuye: Applied in 1ee63726. The extension module is now explicitly a client module, and its contract documents the required export surface, null selection handling, empty metadata handling, and the exact composer flag scope.

frontend/src/lib/task-runtime-ui-extension.tsx
qinxuye: Applied in 1ee63726. The capability flag is explicitly typed as boolean so a replacement has the intended contract rather than a false literal type.

frontend/src/components/chat/ChatInput.tsx
qinxuye: Applied in 1ee63726. extraRuntimeExtensions is computed once and reused for both delivery deduplication and the submitted configuration, removing the nested spread and ternary.

PR #1237, opened 2026-08-10
fix(agent): extend grounding rule to quantitative data and add it to all final-answer prompts

### LINKED ISSUE #1235: bug(agent): fabricated data survives compaction — long sessions lose grounding and report prompts have no anti-fabrication constraint
## Summary

A long-running agent session produced a full deck of KPI data with **no real data source connected**, and did not disclose that the numbers were unsourced until it was directly challenged. This happened **3 times, not once**. The same session also showed the agent losing track of earlier conversation state ("out of context") as the conversation grew long.

These two symptoms are not independent. The current context-compaction design deletes the evidence an agent's answer is supposed to be grounded in, while simultaneously instructing the model not to re-fetch it — and the only anti-fabrication constraint we ship is missing from the exact prompts that emit final report output.

> Source: internal issue tracker row #236 (Marketing / Agent, Urgent, opened Aug 12). The captured row is truncated; it says the session surfaced "several related sub-issues" but only the *Fabricated KPI data* sub-issue is legible. This issue covers that sub-issue plus the two themes named in the row title. Remaining sub-issues will be appended once the full row text is available.

### 1. Compaction replaces the whole transcript with a ≤1024-token prose summary

[`compact_with_llm_response()`](src/xagent/core/agent/context/execution.py#L1030-L1088) does a **total replacement** of the message list:

```python
next_messages = [summary_message]
if latest_user is not None:
    next_messages.append(latest_user)
self.messages = next_messages
```

Every tool observation — i.e. every row of real data the agent actually retrieved — is discarded. The only survivor is a natural-language summary whose budget is capped at
[`COMPACT_SUMMARY_MAX_TOKENS = 1024`](src/xagent/core/agent/context/execution.py#L42) by
[`_llm_compact_max_tokens()`](src/xagent/core/agent/context/execution.py#L1201-L1205), independent of how large the history was.

Compaction fires at 75% of the model's context window ([`get_compact_threshold_ratio()`](src/xagent/config.py#L637), wired in [`prepare_llm_for_context()`](src/xagent/core/agent/runtime.py#L84-L96)). For a 200k-window model that is ~150k tokens of history compressed into at most 1024 tokens — a ~150:1 ratio. The compaction prompt asks the summarizer to "preserve … tool observations" ([`_build_llm_compact_prompt()`](src/xagent/core/agent/context/execution.py#L1163-L1199)), but at that ratio concrete figures and table rows cannot survive. What survives is prose *about* the data.

This is the direct mechanism behind "out of context for long conversation": after the first compaction the agent's memory of the session is a lossy paraphrase, not the session.

### 2. The post-compaction instruction tells the model *not* to re-ground itself

The summary message injected at [execution.py:1052-1061](src/xagent/core/agent/context/execution.py#L1052-L1061) says:

> "Use this summary as the current execution state. … **do not repeat completed tool calls or regenerate completed artifacts** unless the latest user request explicitly asks to restart, revise, or regenerate them."

Combined with the ReAct instruction "After a successful tool call, base the final answer on the latest tool result instead of repeating the same tool work" ([react.py:771-773](src/xagent/core/agent/pattern/react/react.py#L771-L773)), the agent is told to rely on tool results that no longer exist in its context, and explicitly discouraged from re-running the tool that would restore them.

That is a confabulation trap by construction: the model needs numbers, the numbers are gone, re-fetching is discouraged, and nothing marks the gap. Plausible-looking numbers are the cheapest way to satisfy the instruction.

The `truncate` fallback strategy ([execution.py:996-1008](src/xagent/core/agent/context/execution.py#L996-L1008)) has the same property with no summary at all — it keeps a tail window and drops everything older silently.

### 3. Nothing distinguishes tool-derived facts from model-generated text

There is no provenance marking anywhere in the context model. Once a figure is in the transcript — whether it came from a `web_search` result, a RAG retrieval, or the model's own prior turn — it is just message content. After compaction it is just summary prose. Neither the agent nor the UI can answer "where did this number come from?", which is why the fabrication was only admitted **after being challenged** rather than disclosed up front.

### 4. The anti-fabrication clause is missing from the prompts that emit reports

We ship exactly one grounding constraint, at [react.py:786-790](src/xagent/core/agent/pattern/react/react.py#L786-L790):

> "Do not introduce specific **entities, incidents, dates, sources, or causal explanations** that are not supported by the conversation, retrieved context, or tool results. If available context is insufficient, say so or use an appropriate tool to verify."

Three gaps:

- **It does not cover data.** The enumeration lists entities/incidents/dates/sources/causes — not metrics, figures, statistics, percentages, or table rows. A fabricated KPI deck does not violate the letter of this instruction.
- **It is absent from the forced final-answer path.** The `force_final_answer` branch ([react.py:753-762](src/xagent/core/agent/pattern/react/react.py#L753-L762)) builds its own instruction and carries no grounding constraint — this is the path taken when the iteration budget runs out, i.e. precisely on long sessions.
- **It is absent from DAG (`think` mode) entirely.** The DAG completion-assessment prompt that produces the final user-facing answer ([dag.py:1386-1408](src/xagent/core/agent/pattern/dag/dag.py#L1386-L1408)) has no grounding constraint at all. A "build me a KPI report" task is exactly the kind of task that routes to DAG.

### 5. No "data source connected" precondition exists

There is no concept of "this task has a live data source" that report-shaped work could be gated on. An agent with no KB, no connected dataset, and no successful retrieval call is free to produce a fully-formatted deliverable, and the output is visually indistinguishable from a grounded one.

## Reproduction sketch

1. Start a `think` (DAG) or `balanced` (ReAct) task with no knowledge base and no data-source tool connected.
2. Ask for a multi-section KPI report.
3. Keep the conversation going past the compaction threshold (75% of the model's context window) — e.g. iterate on formatting and section content.
4. Observe: the deck contains specific figures; the response does not state that no data source was consulted; asking "where did these numbers come from?" is what triggers the disclosure.

## Proposed direction

Ordered by cost/benefit; each is independently landable.

**A. Close the prompt gaps (small, immediate)**
- Extend the grounding clause to cover quantitative content: metrics, figures, statistics, percentages, table rows, time series.
- Add the clause to the `force_final_answer` branch and to the DAG completion-assessment prompt.
- Require explicit disclosure when a deliverable contains figures with no supporting tool result: state the numbers are illustrative/placeholder, up front and unprompted.

**B. Make compaction preserve evidence, not just prose (medium)**
- Scale the summary budget with history size instead of a flat 1024-token cap.
- Retain the raw tool observations that the final answer depends on, or persist them as re-materializable references (the `ContextReference` machinery at [execution.py:1090-1128](src/xagent/core/agent/context/execution.py#L1090-L1128) already does this for images — the same idea should apply to structured tool results).
- Soften "do not repeat completed tool calls" into "do not repeat completed tool calls **unless the required detail was lost in compaction** — re-run the tool instead of reconstructing values from this summary."
- Have the summary explicitly record *what was dropped*, so the agent knows it has a gap rather than silently filling it.

**C. Provenance and a grounding gate (larger)**
- Track whether a given claim/figure traces to a tool result, and surface that in the UI.
- Add a "data source connected" signal for a task, and either block or hard-label report-shaped output produced without one.

## Open questions

- Should the agent be **blocked** from building reports until a real data source is connected, or should it be **allowed but forced to label** the output as unsourced? (Original question raised for Gerard in the tracker row. Blocking is safer for trust; labeling keeps drafting/templating workflows usable. A middle option: allow it, but require an unmissable disclosure banner and make the fabricated figures visually distinct.)
- Should compaction be allowed to run at all in a session that has produced quantitative deliverables, or should such sessions switch to a stricter evidence-preserving strategy?
- Which mode(s) were used in the reported session? The DAG path having no grounding constraint at all makes it the prime suspect.

## Scope note

The captured tracker row is truncated mid-sentence and refers to sub-issues that are not legible in the capture. Once the full text is available, the remaining sub-issues should be triaged into this issue or split out.

codeacme17:
## Summary

Implements **proposal A** of #1235. Scope is limited to prompt-level grounding constraints; proposals B (evidence-preserving compaction) and C (provenance tracking / data-source gating) are deliberately out of scope and will follow separately.

The issue reports an agent producing a full KPI deck with no data source connected, and only admitting the numbers were unsourced after being challenged. Two prompt-level gaps made that outcome cheap:

- The single anti-fabrication clause we ship enumerated **entities, incidents, dates, sources, and causal explanations** — not quantitative content. A fabricated KPI table did not violate its letter.
- The clause lived only in the ReAct `has_tools` branch. Three other paths that emit a final user-facing answer carried no grounding constraint at all.

## Changes

Factor the clause into a shared `xagent.core.agent.grounding` module (mirroring the existing `language.py` prompt-fragment pattern) and close both gaps:

- **Cover quantitative data** — metrics, figures, statistics, percentages, table rows, and time series are now named explicitly alongside the original enumeration.
- **Require up-front disclosure** — when an answer contains figures that no tool result or provided context supports, the model must state before presenting them that they are illustrative placeholders, and must never present invented numbers as real data.
- **Wire it into every final-answer path:**
  | Path | File | Status before |
  |---|---|---|
  | ReAct, tools available | `pattern/react/react.py` | had the old clause (now shared + extended) |
  | ReAct, `force_final_answer` | `pattern/react/react.py` | **no constraint** — this is the path taken when the iteration budget runs out, i.e. on long sessions |
  | DAG completion assessment | `pattern/dag/dag.py` | **no constraint** — the path report-shaped `think` tasks route to |
  | Auto routing, direct `final_answer` | `pattern/auto/auto.py` | **no constraint** — the router can emit a user-facing answer directly |

`grounding_rule(can_call_tools=False)` is used on paths that are restricted to a single forced control tool, where the original "use an appropriate tool to verify" escape hatch would be unfollowable; those get "say so instead of filling the gap with invented values". The Auto router additionally gets told to route to `react` when the answer would need unsupported specifics, since routing — not answering — is its lever.

## Testing

- New `tests/core/agent/test_grounding.py` covering both variants of the rule.
- New assertions in `test_react.py`, `test_dag.py`, and `test_auto.py` exercising the real prompt-building functions (`_messages_for_llm` with and without `force_final_answer`, `_completion_assessment_messages`, and the Auto decision prompt) rather than synthetic strings.
- `tests/core/agent/` passes (205 tests across the four affected files re-verified after rebase); `ruff check`, `ruff format --check`, and `mypy` clean on the touched files.

## Notes for reviewers

- This is a prompt-only change; no behavioural code paths, no compaction or context-model changes.
- It reduces how often fabricated figures are emitted and makes disclosure the instructed default, but it is a mitigation, not a guarantee — the structural fix is proposal B (compaction currently discards the tool observations an answer is supposed to be grounded in) and proposal C (provenance).
- The open question in #1235 — whether report-shaped output should be **blocked** without a connected data source or **allowed but hard-labelled** — is untouched here and still needs a product decision.

Refs #1235


tests/core/agent/test_auto.py
```diff
@@ -1072,6 +1072,28 @@ async def test_auto_decision_prompt_exposes_execution_tool_names() -> None:
...
+        tools=[FakeSearchTool()],
+        llm=llm,
+        runtime=PatternRuntime(),
+    )
+
+    assert result["success"] is True
+    decision_prompt = llm.calls[0]["messages"][-1]["content"]
+    assert "quantitative data" in decision_prompt
+    assert "illustrative placeholders" in decision_prompt
+    assert "choose react instead" in decision_prompt
```
rogercloud: Minor: this test asserts only positive substrings and never asserts the absence of the tool-verification phrase, so it cannot catch a `can_call_tools` regression. Verified by experiment: flipping `can_call_tools=False` to `True` at `src/xagent/core/agent/pattern/auto/auto.py:1310` leaves this test green (all 47 tests in the file still pass). The sibling tests at `tests/core/agent/test_dag.py:389` and `tests/core/agent/test_react.py:1344` both include the negative assertion and would catch the equivalent regression.

Suggested fix:

```python
assert "use an appropriate tool" not in decision_prompt
```

src/xagent/core/agent/grounding.py
```diff
@@ -0,0 +1,43 @@
...
+            "the gap with invented values. "
+        )
+    )
+    return (
+        "Do not introduce specific entities, incidents, dates, sources, "
+        "causal explanations, or quantitative data (metrics, figures, "
+        "statistics, percentages, table rows, or time series) that are not "
+        "supported by the conversation, retrieved context, or tool results. "
+        f"{insufficient_context_rule}"
+        "If the answer includes figures that no tool result or provided "
```
rogercloud: Minor (prompt-wording risk): the hard prohibition ("Do not introduce ... quantitative data ... that are not supported") is immediately followed by an unconditional disclosure fallback. Nothing in this clause or in any call-site context (`react.py:750-800`, `dag.py:1390-1415`, `auto.py:1295-1320`) scopes the disclosure sentence to the case where the user actually asked for a template, mockup, or example. A model could read it as a sanctioned "fabricate and label" path rather than defense-in-depth, producing exactly the #1235 output (a full deck of invented KPIs) with a disclaimer attached.

This is a faithful transcription of proposal A in the linked issue rather than an implementer error, and the risk is unverified against real model behavior — but it is cheap to tighten. Suggestion: make the disclosure conditional, e.g. "Do not fabricate figures to fill gaps; if the user explicitly asked for a template or illustrative example, label the placeholder figures up front, before presenting them, and never present invented numbers as real data."

tests/core/agent/test_react.py
```diff
@@ -1324,6 +1325,25 @@ def test_react_accepts_null_tool_calls_in_forced_final_response() -> None:
...
+    )[0]["content"]
+    forced_prompt = pattern._messages_for_llm(
+        context, has_tools=True, force_final_answer=True, tool_names=["final_answer"]
+    )[0]["content"]
+
+    for prompt in (tool_prompt, forced_prompt):
+        assert "quantitative data" in prompt
+        assert "illustrative placeholders" in prompt
+    assert "use an appropriate tool to verify" in tool_prompt
+    assert "use an appropriate tool" not in forced_prompt
```
rogercloud, line 1351: Minor (coverage): this calls `_messages_for_llm(..., force_final_answer=True)` directly, so it verifies the prompt text but not the `run()` wiring. Verified by experiment: patching the real call site in `run()` (`src/xagent/core/agent/pattern/react/react.py`, around line 454, where `_messages_for_llm` is invoked with `force_final_answer=force_final_answer_now`) to hardcode `force_final_answer=False` leaves the entire `test_react.py` suite green (98/98), including this new test. That regression would ship undetected.

This matters more here than elsewhere: `force_final_answer` is the path taken when the iteration budget runs out, i.e. on long sessions — precisely the #1235 scenario. Suggestion: add one assertion in an existing full-`run()`-level forced-final-answer test verifying the grounding text reaches the actual LLM call's message content, or explicitly accept the gap given this file's convention of testing prompt fragments via private methods.

rogercloud (approved):
## Summary

This PR closes prompt-level gaps in the codebase's single anti-fabrication ("grounding") clause. Previously it covered only entities, incidents, dates, sources, and causes, and existed only in the ReAct `has_tools` branch. This change factors it into a shared module `src/xagent/core/agent/grounding.py` (`grounding_rule(can_call_tools: bool)`), extends the wording to explicitly cover quantitative data (metrics, figures, statistics, percentages, table rows, time series) plus a new up-front disclosure requirement for unsupported figures, and wires the shared rule into all four final-answer-producing paths: ReAct `has_tools`, ReAct `force_final_answer`, DAG completion assessment, and Auto's direct `final_answer` routing. The diff is prompt-only (8 files, +144/-6); no runtime or control-flow code is touched.

- Shared-module design is the right call. `grounding.py` mirrors the existing `language.py` / `final_answer_language_rule` convention, and avoids four inline copies of the same clause. Confirmed there are no leftover duplicates of the old clause text anywhere in `src/xagent/`.
- All three identified gaps are genuinely closed. ReAct `force_final_answer`, DAG completion assessment, and Auto's direct `final_answer` routing are each the actual sole path producing user-facing text for their pattern (flash/`single_call` reuses `ReActPattern`). The `can_call_tools=False` variant is semantically correct at all three new sites, since each is a forced/no-tool-choice context where "use a tool to verify" would be an unfollowable instruction.
- Scope/framing note on the `fix(agent)` title: per issue #1235's own analysis, the primary root cause is context compaction discarding tool-observation evidence on long sessions, explicitly deferred here to "proposal B". On the exact failure path reported (long session, compaction fires, evidence already gone), this clause cannot fire correctly, because the model has no way to know its own numbers are unsourced by the time it writes the answer. The PR's "Notes for reviewers" partly acknowledges this ("a mitigation, not a guarantee"); worth reflecting in the title/changelog framing so the issue is not read as closed end-to-end.
- Testability ceiling: every new and extended test asserts prompt-string substring presence. That verifies wiring, not model behavior, and no eval harness exists in this repo to check the latter. Acceptable for a prompt change, but the effectiveness of the new wording remains unverified.

## Findings

All findings are minor or nit level; nothing blocking. Three are left as inline comments (`test_auto.py`, `grounding.py`, `test_react.py`); the fourth doesn't land on a diff line so it's noted here:

- **Nit — `src/xagent/core/agent/pattern/react/react.py:804-805`**: the `else: return messages` branch of `_messages_for_llm` (unchanged by this PR) carries no grounding instruction, unlike its sibling branches. Verified it is reachable only when `tool_choice == "none"`, which defaults to `"required"` (line 187) and is never overridden at any of the five production `ReActPattern(...)` construction sites (`auto.py:425`, `execution_adapter.py:299/312/321`, `dag.py:864`) — only test code exercises it (`test_react.py:2953/3259/4553`). Currently a dead consistency gap; if `tool_choice="none"` is ever wired into a real config path, that path would silently produce final answers with zero anti-fabrication constraint. Low priority, not blocking, and out of scope for this PR to fix since the branch predates it.

## Testing

Ran `.venv/bin/python -m pytest tests/core/agent/test_grounding.py tests/core/agent/test_react.py tests/core/agent/test_dag.py tests/core/agent/test_auto.py -q` — 205 tests passed, 0 failures.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -1401,6 +1402,7 @@ def _completion_assessment_messages(self, context: Any) -> list[dict[str, Any]]:
                     "missing, choose status=incomplete, leave answer empty, and "
                     "state the missing work plus concise replan instructions. Put "
                     "status before answer in the tool arguments. "
+                    f"{grounding_rule(can_call_tools=False)} "
```
rogercloud, line 1407: Two minor points on this insertion, both about how it reads against the paragraph directly above it.

**Scoping:** `auto.py:1310` introduces the same fragment with an explicit scope — `"When writing that answer field: {grounding_rule(...)}"`. Here it is appended as a flat continuation, immediately after an instruction to leave `answer` empty. Not a hard contradiction (nothing enforces emptiness, and "say so" maps naturally onto the `missing_work`/`replan_instruction` fields just described), but prefixing with "When writing the answer field: " would make the two call sites read consistently and remove the ambiguity.

**Disclosure vs. `status=incomplete`:** the disclosure sentence in `grounding_rule` is unconditional and not scoped away from this path, so a model can read it as licensing a labeled-placeholder answer with `status=completed` instead of reporting the gap via `status=incomplete` — and nothing in the tool schema forces either value. This is the "blocked vs. hard-labeled" question already flagged in the PR body, and the clause would still have prevented the original incident, so it is not blocking. Worth a clause making clear that disclosure does not override choosing `incomplete` when the data source is entirely missing.

src/xagent/core/agent/grounding.py
```diff
@@ -0,0 +1,43 @@
...
+
+from __future__ import annotations
+
+
+def grounding_rule(*, can_call_tools: bool = True) -> str:
+    """Return the grounding rule for user-facing answer generation.
+
+    Args:
+        can_call_tools: Whether the receiving LLM call may invoke work tools.
+            When ``False`` (forced final answers, DAG completion assessment),
```
rogercloud: This parenthetical undercounts the callers. Three sites pass `can_call_tools=False` — `react.py:762` (forced final answer), `dag.py:1405` (DAG completion assessment), and `auto.py:1310` (Auto routing decision). Adding "Auto routing decision" keeps the docstring accurate.

src/xagent/core/agent/grounding.py
```diff
@@ -0,0 +1,43 @@
+"""Shared anti-fabrication rules for prompts that emit user-facing answers.
+
+Every prompt that produces a final user-facing answer must carry this rule;
```
rogercloud: Informational: this reads as an invariant, but nothing enforces it — there is no cross-cutting test that enumerates prompt builders and asserts each carries the rule, only per-path unit tests. And `react.py:804` (the `has_tools=False` branch) genuinely does not carry it, even though that branch is unreachable today. Consider phrasing this as a design goal rather than a guarantee.

Separately, a one-line pointer noting that this clause is the proposal-A mitigation and that proposals B (evidence-preserving compaction) and C (provenance/gating) from #1235 remain open would keep a future reader from mistaking it for the complete fix. That context currently lives only in the PR description.

src/xagent/core/agent/pattern/auto/auto.py
```diff
@@ -1305,7 +1306,11 @@ def _decision_prompt(
             "Use final_answer for simple conversational replies that need no tools; "
             "when action is final_answer, you must include a complete non-empty "
             "answer field in the same tool call. Put action before answer in the "
-            "tool arguments. You must also classify whether "
+            "tool arguments. "
+            f"When writing that answer field: {grounding_rule(can_call_tools=False)} "
+            "If the answer would need such unsupported specifics, choose react "
```
rogercloud: `shrink:` This overlaps the existing react-over-`final_answer` steering further down — the instruction at L1326 and the `existing_context_sufficient=false` branch at L1356. The new sentence is marginally broader in scope than the recency-scoped existing text, but both funnel to the same instruction, now stated twice in different frames. Consolidating into one place (or cross-referencing instead of restating) would trim ~3 lines from an already long decision prompt.

tests/core/agent/test_grounding.py
```diff
@@ -0,0 +1,34 @@
...
+    assert "use an appropriate tool to verify" in rule
+    assert "illustrative placeholders" in rule
+
+
+def test_grounding_rule_without_tools_omits_tool_verification() -> None:
+    rule = grounding_rule(can_call_tools=False)
+
+    assert "use an appropriate tool" not in rule
+    assert "invented values" in rule
+    assert "quantitative data" in rule
```
rogercloud, line 36: All assertions here are substring membership, which cannot catch a malformed join. If the trailing space in `insufficient_context_rule` (`grounding.py:26` / `:30`) were dropped, the rule would render as "…insufficient, say so instead of filling the gap with invented values.If the answer includes…" and every assertion in this file would still pass. Asserting a sentence boundary — e.g. `assert "invented values. If the answer" in rule` (and `"verify. If the answer"` in the tools variant) — would pin the concatenation.

tests/core/agent/test_auto.py
rogercloud: This test is weaker than its siblings: `test_dag.py:388` and `test_react.py:1343` both include the negative assertion `assert "use an appropriate tool" not in ...`, and `test_grounding.py:32` asserts "invented values". Without the negative assertion here, this call site flipping to `can_call_tools=True` would not be caught. Adding both to match would give the three paths parity.

rogercloud (approved):
## Summary

An agent session produced a complete KPI report with invented numbers and never disclosed they were unsourced until challenged (#1235). The root cause was twofold: the anti-fabrication clause the codebase shipped covered entities, incidents, dates, sources, and causal explanations but said nothing about quantitative content, and it existed in only one of the four prompts that can emit a final user-facing answer. This PR extracts that clause into `src/xagent/core/agent/grounding.py`, extends it to metrics/figures/statistics/percentages/table rows/time series plus an up-front disclosure requirement for illustrative figures, and wires it into all four answer-emitting paths (ReAct with tools, ReAct forced final answer, DAG completion assessment, Auto router's direct `final_answer`). Prompt-only, 144 insertions across 8 files, no behavioral machinery added.

**Prior review activity:** the only prior review is an automated bot (gemini-code-assist) that reported no additional feedback. There is nothing substantive to reconcile, so this is treated as the first substantive review.

## Design verdict: sound

- **The pre-PR gap is real.** Verified against the base commit: the `force_final_answer` branch in `react.py`, the DAG completion-assessment system prompt, and the Auto router's decision prompt each carried zero grounding text. Only the ReAct-with-tools branch had the clause.
- **Coverage claim holds.** `single_call` mode maps onto `ReActPattern` (via `execution_adapter.py`), and DAG steps also execute through `ReActPattern`, so the four wired sites genuinely are every path that emits a user-facing answer today.
- **Right shape, right size.** `grounding.py` mirrors the established `language.py` convention — a module of pure string-returning prompt-fragment functions, no registry, no subclassing, no new abstraction. One function with a single `can_call_tools` axis is correctly scoped: each call site already layers its own remedy semantics (`outcome=partial`, `status=incomplete`, `action=react`) in the surrounding prompt text, so the shared fragment only needs to vary on the one thing that changes its own phrasing — whether a tool call is even possible. The flag is exercised at both `True` and `False` across real call sites, so it is not speculative flexibility.
- **Correctly scoped to proposal A.** Proposals B (evidence-preserving compaction) and C (provenance/data-source gating) stay out, tracked separately. This is a mitigation, not a structural fix, and that is the right call for a P0.

No criticals, no majors. Everything below is minor or informational.

## Minor findings

1. **`src/xagent/core/agent/pattern/dag/dag.py:1405` — disclosure clause vs. `status=incomplete`.** The clause is injected immediately after "if data is missing, choose status=incomplete, leave answer empty…". The disclosure sentence is unconditional prose and is not scoped away from this path, so a model can read it as licensing a labeled-placeholder answer with `status=completed` rather than reporting the gap via `status=incomplete`. Nothing in the DAG tool schema mechanically forces either value. Note this is the same "blocked vs. hard-labeled" product question already raised in the PR body — and even read at face value the clause would have prevented the original incident, since disclosure-before-presenting is exactly what was missing. Worth a clarifying phrase that disclosure does not override choosing `incomplete` when the data source is entirely absent.

2. **`src/xagent/core/agent/pattern/dag/dag.py:1405` — missing scoping phrase.** `auto.py:1310` scopes the rule explicitly ("When writing that answer field: …"); `dag.py` appends it as a flat continuation of a paragraph that just instructed the model to *leave `answer` empty*. This is not a hard contradiction (nothing enforces emptiness, and "say so" maps naturally onto the `missing_work`/`replan_instruction` fields), but the asymmetry with `auto.py` leaves avoidable ambiguity. Prefixing with "When writing the answer field: …" would make the two sites read consistently.

3. **`src/xagent/core/agent/grounding.py:15` — docstring undercounts the `can_call_tools=False` callers.** The parenthetical lists "forced final answers, DAG completion assessment" but omits the third: the Auto routing decision prompt. Exactly three sites pass `False` — `react.py:762`, `dag.py:1405`, `auto.py:1310`.

4. **`tests/core/agent/test_grounding.py` — substring-only assertions cannot catch malformed concatenation.** Every assertion is membership (`"table rows" in rule`). If the trailing space in `insufficient_context_rule` (`grounding.py:26`/`:30`) were dropped, yielding `"…verify.If the answer…"`, no current test would fail. Asserting a sentence-boundary substring such as `"verify. If the answer"` would close this.

5. **`tests/core/agent/test_auto.py:1094` — new test is weaker than its DAG/ReAct siblings.** It asserts only `"quantitative data"`, `"illustrative placeholders"`, `"choose react instead"`. The analogous tests in `test_dag.py:386-388` and `test_react.py:1337-1343` also assert `"invented values"` / the negative `"use an appropriate tool" not in …`. Without the negative assertion, this call site silently flipping to `can_call_tools=True` would not be caught.

## Informational (no change required)

6. **`src/xagent/core/agent/pattern/react/react.py:804` — the `has_tools=False` branch returns the messages untouched**, so it carries no anti-fabrication text at all. This is effectively unreachable today: no `ReActPattern(` construction site (`execution_adapter.py`, `dag.py`, `auto.py`) passes `tool_choice="none"`. A one-line comment marking the gap would keep it from being silently reintroduced if that ever becomes reachable.

7. **`src/xagent/core/agent/grounding.py:3-4` states an unenforced invariant** — "Every prompt that produces a final user-facing answer must carry this rule". There is no cross-cutting test walking all prompt builders to verify it; coverage is per-path unit tests only, and per (6) one branch genuinely lacks it. Consider phrasing it as a design goal rather than a guarantee.

8. **No in-code pointer that this is mitigation-only.** That proposals B and C from #1235 remain open lives solely in the PR description. A one-line comment in `grounding.py` referencing them would stop a future reader treating this clause as the complete fix.

- `src/xagent/core/agent/pattern/auto/auto.py:1311` — `shrink:` redundant with the existing "choose react … do not choose final_answer" steering at `auto.py:1326` and `auto.py:1356`. The new sentence has marginally broader scope than the recency-scoped existing text, but both funnel to the same "prefer react over final_answer when the answer would be ungrounded" instruction, now stated twice in different frames. Consolidate into one instruction, or cross-reference rather than restate.

`net: -3 lines possible`

## Verification

- 205 tests pass across `tests/core/agent/test_grounding.py`, `tests/core/agent/test_react.py`, `tests/core/agent/test_dag.py`, `tests/core/agent/test_auto.py`.
- `ruff check`, `ruff format --check`, and `mypy` clean on the touched files.
- Correctness spot-checks: every `can_call_tools` value is factually accurate for its call site (all three `False` sites genuinely cannot invoke work tools); all four prompt concatenations render grammatically (no double spaces, no missing sentence boundaries); `"react"` is a valid action value in the Auto decision schema. Tests exercise the real production prompt builders (`_messages_for_llm`, `_completion_assessment_messages`, the live decision prompt), not synthetic reimplementations.

No user-facing documentation is needed — the change adds no API, config, or CLI surface.

src/xagent/core/agent/grounding.py
codeacme17: Agreed, and fixed in d444452 — but not with the suggested wording, because it would have regressed against the incident this PR is for.

Your diagnosis is right: the disclosure sentence sits immediately after the prohibition it relaxes, it is the later and more specific instruction, and "you may include unsourced figures provided you label them" is a coherent reading of it. That is a real fabricate-and-label path.

The proposed replacement gates the disclosure duty on *the user having asked for a template*. In #1235 the user asked for a **real** KPI report. Under that wording, a model that fabricates anyway lands in a state the prompt only prohibits and no longer governs — no instruction to disclose. The prohibition is the thing that already failed: the pre-PR clause flatly prohibited unsupported entities/incidents/dates/sources/causes and the deck still shipped. It also inverts the issue's trigger — proposal A requires disclosure "when a deliverable contains figures with no supporting tool result", a property of the deliverable, not of how the request was phrased.

So I withdrew the *permission* instead of narrowing the *duty*:

> Never invent figures to fill a gap, and never present invented numbers as real data; produce unsupported figures only when the user explicitly asked for a template, mockup, or illustrative example. Labeling is required either way: if the answer ends up containing any figure that no tool result or provided context supports, whether or not the user asked for one, say so up front, before presenting it, and state that those figures are illustrative placeholders not drawn from any data source.

Your template/mockup carve-out is in there — as the only case where producing such figures is allowed at all — while labeling stays unconditional, so the safety net still covers the #1235 path. Added `test_grounding_rule_requires_labeling_regardless_of_request` pinning exactly that, so a future edit cannot quietly re-gate it.

tests/core/agent/test_react.py
codeacme17: Fixed in d444452 — verified your experiment reproduces, and my first attempt at the fix did not actually close it.

I initially asserted `"quantitative data" in forced_final_prompt` on the real `run()`. That still passed with `force_final_answer=False` injected at `react.py:457`, because after this PR **both** branches carry the rule, so membership alone cannot tell which one ran. The discriminating signal is the `can_call_tools=False` variant:

```python
forced_final_prompt = llm.stream_calls[1]["messages"][0]["content"]
assert "quantitative data" in forced_final_prompt
assert "invented values" in forced_final_prompt
assert "use an appropriate tool" not in forced_final_prompt
```

Re-ran your experiment against it: green when correct, and with `force_final_answer=False` at `react.py:457` it fails on `assert "invented values" in forced_final_prompt` (the prompt falls back to the has-tools wording). Added to `test_react_recovers_unavailable_forced_final_with_full_tool_set`, which already does a full `run()` and whose `stream_calls[1]` is the forced turn — no new fixture needed.

tests/core/agent/test_auto.py
codeacme17: Fixed in d444452. Added both, matching the DAG/ReAct siblings:

```python
assert "invented values" in decision_prompt
assert "set existing_context_sufficient=false and choose react" in decision_prompt
assert "use an appropriate tool" not in decision_prompt
```

The middle assertion changed shape because of your other comment on `auto.py:1311` — the sentence now routes through the classification field rather than asking for a routing choice directly, so the test pins the field name.

src/xagent/core/agent/pattern/dag/dag.py
codeacme17: Both addressed in d444452, and the scoping prefix picked up a third gap you didn't name.

**(a) Scoping.** Added, and widened. Chasing the "which field does this constrain" ambiguity led to a bigger hole: `dag.py:1384` puts `"candidate_output": self._final_output()` into the payload and `dag.py:1223` falls back to `assessment.answer or self._final_output()`. So the realistic #1235 path here is a **step** LLM fabricating the KPI table and the assessor **passing it through** — and a rule phrased as "what you write" reads as exempting copied step output. The prefix now covers both:

```python
"When writing the answer field, including any content carried "
"over from candidate_output or step_results: "
```

(You're right that it wasn't literally adjacent to "leave answer empty" — "Put status before answer" intervenes. The ambiguity holds either way, since this prompt governs a two-branch decision while `react.py:762` governs a single answer.)

**(b) Disclosure vs `status=incomplete`.** Real, and there's an incentive gradient beyond the schema being permissive: `status=incomplete` consumes a replan (`dag.py:1247`), and at `completion_replan_count >= max_completion_replans` (default 3) the execution **hard-fails** with `completion_replan_limit_exceeded` (`dag.py:1235-1245`). So `completed` + labeled placeholder is strictly cheaper than `incomplete` when the data is simply unobtainable.

I did not fix it by forcing `status=incomplete`, because that decides #1235's open block-vs-label product question unilaterally, and does it via the harshest available mechanism — a hard failure after three replans. Instead the choice is made **observable**: `reason` is required (`dag.py:1451-1457`) and is persisted into result metadata (`dag.py:1231`), so

```python
"If the answer presents any figure as an illustrative "
"placeholder because no step produced the underlying data, "
"name that unsourced data in reason even when you choose "
"status=completed. "
```

That keeps "allowed but labeled" intact and gives product real telemetry to answer block-vs-label from data rather than from guesswork. Happy to switch to forcing `incomplete` if you'd rather make that call now.

src/xagent/core/agent/grounding.py
codeacme17: Fixed in d444452 — the docstring now names all three (`react.py:762` forced final answer, `dag.py` completion assessment, `auto.py` routing decision).

src/xagent/core/agent/grounding.py
codeacme17: Both applied in d444452.

The invariant is now stated as a design goal, with the exception named so it doesn't read as a guarantee:

> Every prompt that produces a final user-facing answer should carry this rule. That is a design goal, not an enforced invariant: nothing walks the prompt builders to verify it, and ReAct's no-tool branch still lacks it.

Also added the mitigation-scope pointer you asked for, naming proposals B and C as open, plus a one-line comment at `react.py:804` on the `has_tools=False` branch so the gap is visible at the site rather than only in this module's docstring.

Worth noting this is the second time the "every prompt" phrasing has been flagged — an earlier pass caught the same overclaim when `auto.py` was the missing path. It's now weakened rather than re-asserted.

src/xagent/core/agent/pattern/auto/auto.py
codeacme17: Partially — with a correction and a better landing spot than consolidation.

The citation at L1324-1327 is a different rule: "if the latest user message explicitly asks to call or use an available tool… choose react" is about **explicit tool requests**, not evidence sufficiency. No overlap there. The real near-duplicate is L1344-1350 ("Use final_answer only when the current conversation and available retrieved context already provide enough evidence").

The overlap is also only half: nothing else in this prompt governs the **content** of the answer text or mentions quantitative data, and L1351-1358's trigger list is anchored to recency/news/CVE flavour, so an invented internal-KPI table doesn't clearly land in it. The duplicated part is just the routing tail ("choose react instead").

Rather than delete the tail, I retargeted it at the classification fields, because `auto.py:684-712` already has a **deterministic** guard that a bare "choose react" bypasses:

```python
if (self.decision.action == AutoAction.FINAL_ANSWER
        and self.decision.requires_current_or_external_facts
        and not self.decision.existing_context_sufficient):
    ... falling back to react
```

So L1311 is now `"set existing_context_sufficient=false and choose react…"`, and the L1351 enumeration gained `"or metrics and figures the user expects to reflect real data"`. Net is ~1 line rather than the 3 you costed, but the guarantee moves from prompt adherence to a code-enforced fallback — which seemed the better trade for a P0. Test updated to pin the field name.

tests/core/agent/test_grounding.py
codeacme17: Good catch — fixed in d444452. Added sentence-boundary assertions to both variants:

```python
assert "verify. Never invent figures" in rule          # can_call_tools=True
assert "invented values. Never invent figures" in rule # can_call_tools=False
```

Confirmed they behave as intended: dropping the trailing space in `insufficient_context_rule` now fails these, while every membership assertion in the file still passes.

tests/core/agent/test_auto.py
codeacme17: Fixed in d444452 — added `"invented values"` and the negative `assert "use an appropriate tool" not in decision_prompt`, so the three paths now have parity. Details in the reply on the same finding in your earlier review.

codeacme17:
Thanks for both reviews — all nine findings are addressed in d444452. Replies are on each thread; the three that changed shape from what was suggested:

**Disclosure wording.** The fabricate-and-label reading is real, but gating the disclosure duty on the user having asked for a template would have removed it in #1235 itself, where the user asked for a *real* KPI report — and it inverts proposal A's trigger, which is a property of the deliverable, not of the request's phrasing. So I withdrew the *permission* rather than narrowing the *duty*: unsupported figures may only be produced when a template/mockup/example was explicitly requested, while labeling stays unconditional. New test pins that so it can't be quietly re-gated.

**DAG `status=incomplete`.** Confirmed, plus an incentive gradient neither review cited: `incomplete` consumes a replan and hard-fails the execution at the limit (`dag.py:1235-1245`), so `completed` + placeholder is the cheaper exit. I did not force `incomplete` — that decides #1235's open block-vs-label question unilaterally, via the harshest mechanism available. Instead unsourced data must be named in `reason`, which is required and persisted, making the choice observable so product can decide from telemetry. Say the word if you'd rather force it now.

The DAG scoping prefix also surfaced a third gap: the assessor usually *passes through* `candidate_output` rather than authoring figures, so the rule now explicitly covers carried-over step content.

**Auto overlap.** The cited L1324-1327 is a different rule (explicit tool requests); the real duplicate is L1344-1350, and only its routing tail. Rather than delete it I retargeted it at `existing_context_sufficient=false` so it trips the deterministic fallback at `auto.py:684-712` instead of relying on the model's routing choice. Net ~1 line instead of 3, but the guarantee moves from prompt adherence to code.

On the forced-final coverage gap: your experiment reproduced, and my first fix didn't actually close it — asserting `"quantitative data"` still passed with `force_final_answer=False` injected, since after this PR both branches carry the rule. It now asserts the `can_call_tools=False` variant, which is the only discriminating signal, and I re-ran your injection to confirm it fails.

Not changed: the `has_tools=False` branch at `react.py:804` still has no grounding text. It's unreachable in production and predates this PR, so per your own note I left it and added a comment marking the gap at the site.

`tests/core/agent/` passes; `pre-commit run --files` clean on all eight touched files.

PR #1238, opened 2026-08-10
fix(agent): disclose evidence dropped by compaction and allow re-grounding

### LINKED ISSUE #1235: bug(agent): fabricated data survives compaction — long sessions lose grounding and report prompts have no anti-fabrication constraint

codeacme17:
## Summary

First slice of **proposal B** in #1235. Independent of #1237 (proposal A) — different files, no conflict; either can land first.

Issue §2 describes compaction as "a confabulation trap by construction": the transcript is replaced by a prose summary, the model is then told *"do not repeat completed tool calls"*, and nothing marks what went missing. The numbers the answer should be grounded in are gone, re-fetching them is discouraged, and the gap is invisible — so plausible-looking numbers are the cheapest way to satisfy the instruction.

This PR removes the trap's two prompt-side legs. It does **not** yet change how much survives compaction (items 1 and 2 below).

## Changes

**1. The post-compaction instruction now permits re-grounding.** It previously read, in full, *"do not repeat completed tool calls or regenerate completed artifacts unless the latest user request explicitly asks to restart, revise, or regenerate them."* It now adds `…or the detail you need was lost in compaction`, plus:

> This summary is a lossy paraphrase of the raw history, not the history itself: when the answer needs an exact value, figure, statistic, table row, quotation, or identifier that this summary does not literally contain, re-run the tool that produced it instead of reconstructing the value from this summary or from memory.

The original intent — don't redo already-completed work — is preserved; only the case that produces fabrication is carved out.

**2. The summary now records what was dropped.** A `ContextReference`-style notice already existed but covered images only; textual tool observations vanished silently. The summary message now appends, when any were dropped:

```
Raw observations from 4 completed tool call(s) were dropped by this compaction.
Their exact values are no longer in context; only the summary above describes
them. Treat any figure not literally present in that summary as unavailable
rather than recalled. Tools whose results were dropped:
- web_search x3
- read_file
```

Aggregated by tool name with counts (capped at 20 distinct names) so a long session costs a bounded number of tokens. Hidden messages are excluded — superseded observations were already absent from the prompt, so reporting them as newly dropped would be false.

**3. Observability.** Both strategies now report `dropped_tool_result_count` in `CompactResult.metadata`.

## Notes for reviewers

- **`truncate` gets metadata only, no in-prompt notice.** Issue §2 calls out that this fallback drops history silently too, and it deserves the same disclosure. But `truncate`'s contract is "keep exactly `max_messages` messages" and three existing tests assert exact counts and positions; inserting a system notice breaks that invariant, and the notice would itself be truncated later. Left for a follow-up rather than decided unilaterally here.
- **Identity-based diff in `truncate`.** `_tail_window_preserving_tool_pairs()` is a suffix *minus* any interior fragments `_sanitize_tool_message_pairs()` removes, so the dropped set is computed by object identity rather than by slicing a prefix — slicing would misattribute retained messages as dropped whenever sanitization fires.
- This is a prompt/accounting change. Nothing about which messages survive compaction changes in this PR.

## Testing

Four new tests in `tests/core/agent/test_context.py` covering the notice text and per-tool counts, the no-tool-results case, hidden-message exclusion, and the `truncate` count; existing compaction tests extended for the softened instruction. `tests/core/agent/`, `tests/core/test_context_ref.py`, and `tests/core/test_context_materializer.py` pass; `ruff check`, `ruff format --check`, and `mypy` clean on the touched files.

## Remaining in proposal B

1. Scale the summary budget with history size — it is currently capped at a flat 1024 tokens (`COMPACT_SUMMARY_MAX_TOKENS`) regardless of how large the history was, which is where the ~150:1 ratio comes from.
2. Retain raw tool observations, or persist them as re-materializable references the way `ContextReference` already does for images.

Item 1 needs a decision on the ceiling: `max_tokens` is passed straight to the provider, and `BaseLLM` exposes `context_window` but no max-output-tokens attribute, so a large budget risks exceeding a model's output cap. I'd add it to `config.py` as an `XAGENT_*`-configurable ratio with a conservative default rather than hardcoding a bigger constant.

Refs #1235

### DIFF
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -43,6 +43,7 @@ COMPACT_SUMMARY_MAX_TOKENS = 1024
 COMPACT_SUMMARY_MIN_TOKENS = 256
 COMPACT_CONTEXT_REF_MAX_TOKENS = 2048
 COMPACT_DROPPED_REF_NOTICE_MAX_CHARS = 2048
+COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
 
 
 def _utcnow() -> datetime:
@@ -997,14 +998,28 @@ class ExecutionContext:
         original_count = len(self.messages)
         if self.compact_config.strategy == "truncate":
             keep_count = min(max(0, self.compact_config.max_messages), original_count)
-            self.messages = self._tail_window_preserving_tool_pairs(keep_count)
-            removed = max(0, original_count - len(self.messages))
+            retained = self._tail_window_preserving_tool_pairs(keep_count)
+            removed = max(0, original_count - len(retained))
+            # The window is a suffix minus any interior tool fragments sanitized
+            # out of it, so diff by object identity rather than slicing a prefix.
+            retained_ids = {id(message) for message in retained}
+            dropped_tool_counts = self._dropped_tool_result_counts(
+                [
+                    message
+                    for message in self.messages
+                    if id(message) not in retained_ids
+                ]
+            )
+            self.messages = retained
             return CompactResult(
                 compacted=True,
                 original_count=original_count,
                 final_count=len(self.messages),
                 strategy="truncate",
-                metadata={"removed_count": removed},
+                metadata={
+                    "removed_count": removed,
+                    "dropped_tool_result_count": sum(dropped_tool_counts.values()),
+                },
             )
 
         return CompactResult(
@@ -1049,18 +1064,27 @@ class ExecutionContext:
             self._context_refs_removed_by_compaction(latest_user)
         )
         dropped_refs_notice = self._dropped_context_refs_notice(dropped_context_refs)
+        dropped_tool_counts = self._dropped_tool_result_counts(self.messages)
+        dropped_tools_notice = self._dropped_tool_results_notice(dropped_tool_counts)
         summary_content = (
             "Compacted conversation summary:\n"
             f"{summary}\n\n"
             "Use this summary as the current execution state. Continue from the "
             "remaining work described here; do not repeat completed tool calls or "
             "regenerate completed artifacts unless the latest user request "
-            "explicitly asks to restart, revise, or regenerate them. Current system "
-            "instructions still take precedence, and the latest user request remains "
-            "the overall goal."
+            "explicitly asks to restart, revise, or regenerate them, or the detail "
+            "you need was lost in compaction. This summary is a lossy paraphrase of "
+            "the raw history, not the history itself: when the answer needs an exact "
+            "value, figure, statistic, table row, quotation, or identifier that this "
+            "summary does not literally contain, re-run the tool that produced it "
+            "instead of reconstructing the value from this summary or from memory. "
+            "Current system instructions still take precedence, and the latest user "
+            "request remains the overall goal."
         )
         if dropped_refs_notice:
             summary_content = f"{summary_content}\n\n{dropped_refs_notice}"
+        if dropped_tools_notice:
+            summary_content = f"{summary_content}\n\n{dropped_tools_notice}"
         summary_message = Message.role_system(
             summary_content,
             metadata={"compacted_context": True},
@@ -1081,6 +1105,7 @@ class ExecutionContext:
                 "compact_model": getattr(llm, "model_name", None),
                 "retained_context_ref_count": len(compacted_context_refs),
                 "dropped_context_ref_count": len(dropped_context_refs),
+                "dropped_tool_result_count": sum(dropped_tool_counts.values()),
             },
         )
         if original_tokens is not None:
@@ -1151,6 +1176,49 @@ class ExecutionContext:
             current_chars += len(line) + 1
         if omitted:
             lines.append(f"- ... {omitted} additional older reference(s) omitted")
+        return prefix + "\n".join(lines)
+
+    @staticmethod
+    def _dropped_tool_result_counts(messages: list[Message]) -> dict[str, int]:
+        """Count visible tool observations, keyed by tool name.
+
+        Hidden messages are excluded: they were already absent from the prompt
+        before compaction, so reporting them as newly dropped would be wrong.
+        """
+        counts: dict[str, int] = {}
+        for message in messages:
+            if message.hidden or message.role != "tool":
+                continue
+            metadata = message.metadata or {}
+            name = str(metadata.get("tool_name") or "").strip() or "unnamed tool"
+            counts[name] = counts.get(name, 0) + 1
+        return counts
+
+    @staticmethod
+    def _dropped_tool_results_notice(counts: dict[str, int]) -> str:
+        """Describe the tool observations this compaction removes from context.
+
+        Without this, the summary silently replaces every retrieved value and
+        the agent cannot tell a remembered figure from an invented one.
+        """
+        if not counts:
+            return ""
+        ordered = sorted(counts.items(), key=lambda item: (-item[1], item[0]))
+        listed = ordered[:COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES]
+        lines = [
+            f"- {name} x{count}" if count > 1 else f"- {name}" for name, count in listed
+        ]
+        omitted = len(ordered) - len(listed)
+        if omitted:
+            lines.append(f"- ... {omitted} additional tool(s) omitted")
+        total = sum(counts.values())
+        prefix = (
+            f"Raw observations from {total} completed tool call(s) were dropped by "
+            "this compaction. Their exact values are no longer in context; only the "
+            "summary above describes them. Treat any figure not literally present in "
+            "that summary as unavailable rather than recalled. Tools whose results "
+            "were dropped:\n"
+        )
         return prefix + "\n".join(lines)
 
     def _latest_visible_user_message(self) -> Message | None:
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -1180,16 +1180,20 @@ class ExecutionContext:
 
     @staticmethod
     def _dropped_tool_result_counts(messages: list[Message]) -> dict[str, int]:
-        """Count visible tool observations, keyed by tool name.
+        """Count tool observations whose raw result this compaction discards.
 
-        Hidden messages are excluded: they were already absent from the prompt
-        before compaction, so reporting them as newly dropped would be wrong.
+        Superseded observations are excluded: a later observation already
+        replaced their content and ``raw_result``, so compacting them away
+        destroys no evidence and counting them would overstate the loss.
+        Hidden messages are excluded because they are not in the prompt at all.
         """
         counts: dict[str, int] = {}
         for message in messages:
-            if message.hidden or message.role != "tool":
-                continue
             metadata = message.metadata or {}
+            if message.hidden or metadata.get("superseded"):
+                continue
+            if message.role != "tool":
+                continue
             name = str(metadata.get("tool_name") or "").strip() or "unnamed tool"
             counts[name] = counts.get(name, 0) + 1
         return counts
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+from collections import Counter
 from dataclasses import dataclass, field, replace
 from datetime import datetime, timezone
 from enum import Enum
@@ -25,6 +26,7 @@ from ..language import (
     output_language_policy,
     response_language_rules,
 )
+from ..result import CONTROL_TOOL_NAMES, tool_result_succeeded
 from .components import (
     COMPONENT_LOADERS,
     ExecutionComponent,
@@ -36,13 +38,23 @@ from .components import (
 from .enrichment import MEMORY_CONTEXT_METADATA_KEY, SKILL_CONTEXT_METADATA_KEY
 from .memory_tool import MEMORY_TOOLS_METADATA_KEY
 from .message import LLMCallRecord, Message
-from .skill_tool import LOADED_SKILLS_METADATA_KEY, SKILL_INDEX_METADATA_KEY
+from .skill_tool import (
+    LOAD_SKILL_TOOL_NAME,
+    LOADED_SKILLS_METADATA_KEY,
+    SKILL_INDEX_METADATA_KEY,
+)
 
 READ_FILE_CONTEXT_LIMIT = 12_000
 COMPACT_SUMMARY_MAX_TOKENS = 1024
 COMPACT_SUMMARY_MIN_TOKENS = 256
 COMPACT_CONTEXT_REF_MAX_TOKENS = 2048
 COMPACT_DROPPED_REF_NOTICE_MAX_CHARS = 2048
+COMPACT_DROPPED_TOOL_NOTICE_MAX_CHARS = 1024
+COMPACT_DROPPED_TOOL_NAME_MAX_CHARS = 64
+
+# load_skill retrieves guidance, not evidence, and re-running it restores
+# nothing a dropped observation held.
+NON_EVIDENCE_TOOL_NAMES = CONTROL_TOOL_NAMES | {LOAD_SKILL_TOOL_NAME}
 COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
 
 
@@ -1019,6 +1031,7 @@ class ExecutionContext:
                 metadata={
                     "removed_count": removed,
                     "dropped_tool_result_count": sum(dropped_tool_counts.values()),
+                    "dropped_tool_results_by_name": dropped_tool_counts,
                 },
             )
 
@@ -1064,6 +1077,9 @@ class ExecutionContext:
             self._context_refs_removed_by_compaction(latest_user)
         )
         dropped_refs_notice = self._dropped_context_refs_notice(dropped_context_refs)
+        # next_messages below keeps only the system summary and, at most, a
+        # role=="user" message, so no tool observation survives: here the whole
+        # list is the diff. truncate needs a real diff; this does not.
         dropped_tool_counts = self._dropped_tool_result_counts(self.messages)
         dropped_tools_notice = self._dropped_tool_results_notice(dropped_tool_counts)
         summary_content = (
@@ -1076,8 +1092,11 @@ class ExecutionContext:
             "you need was lost in compaction. This summary is a lossy paraphrase of "
             "the raw history, not the history itself: when the answer needs an exact "
             "value, figure, statistic, table row, quotation, or identifier that this "
-            "summary does not literally contain, re-run the tool that produced it "
+            "summary does not literally contain, re-read or re-query the source "
             "instead of reconstructing the value from this summary or from memory. "
+            "Only re-run tools that read; if the value came from a tool that writes, "
+            "sends, executes, or otherwise changes state, do not re-run it -- re-read "
+            "the artifact it produced, or report the value as unavailable. "
             "Current system instructions still take precedence, and the latest user "
             "request remains the overall goal."
         )
@@ -1106,6 +1125,7 @@ class ExecutionContext:
                 "retained_context_ref_count": len(compacted_context_refs),
                 "dropped_context_ref_count": len(dropped_context_refs),
                 "dropped_tool_result_count": sum(dropped_tool_counts.values()),
+                "dropped_tool_results_by_name": dropped_tool_counts,
             },
         )
         if original_tokens is not None:
@@ -1186,17 +1206,28 @@ class ExecutionContext:
         replaced their content and ``raw_result``, so compacting them away
         destroys no evidence and counting them would overstate the loss.
         Hidden messages are excluded because they are not in the prompt at all.
+        Failed and cancelled calls are excluded because they never produced a
+        value to lose, and telling the model they did would let it read a
+        failure as already-succeeded. Control pseudo-tools are excluded because
+        re-running one ends the run or re-contacts the user rather than
+        restoring evidence.
         """
-        counts: dict[str, int] = {}
+        names: list[str] = []
         for message in messages:
+            if message.role != "tool":
+                continue
             metadata = message.metadata or {}
             if message.hidden or metadata.get("superseded"):
                 continue
-            if message.role != "tool":
+            if not tool_result_succeeded(metadata.get("raw_result")):
                 continue
-            name = str(metadata.get("tool_name") or "").strip() or "unnamed tool"
-            counts[name] = counts.get(name, 0) + 1
-        return counts
+            # Whitespace-only names would silently fragment the aggregation,
+            # so normalize before using the name as a key.
+            name = " ".join(str(metadata.get("tool_name") or "").split())
+            if name in NON_EVIDENCE_TOOL_NAMES:
+                continue
+            names.append(name or "unnamed tool")
+        return dict(Counter(names))
 
     @staticmethod
     def _dropped_tool_results_notice(counts: dict[str, int]) -> str:
@@ -1207,22 +1238,36 @@ class ExecutionContext:
         """
         if not counts:
             return ""
-        ordered = sorted(counts.items(), key=lambda item: (-item[1], item[0]))
-        listed = ordered[:COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES]
-        lines = [
-            f"- {name} x{count}" if count > 1 else f"- {name}" for name, count in listed
-        ]
-        omitted = len(ordered) - len(listed)
-        if omitted:
-            lines.append(f"- ... {omitted} additional tool(s) omitted")
         total = sum(counts.values())
+        call_label = "call was" if total == 1 else "calls were"
         prefix = (
-            f"Raw observations from {total} completed tool call(s) were dropped by "
-            "this compaction. Their exact values are no longer in context; only the "
+            f"Raw observations from {total} tool {call_label} dropped by this "
+            "compaction. Their exact values are no longer in context; only the "
             "summary above describes them. Treat any figure not literally present in "
             "that summary as unavailable rather than recalled. Tools whose results "
             "were dropped:\n"
         )
+        # Tool names can come from dynamic MCP server config, so bound both the
+        # per-name length and the total notice size the way the sibling
+        # reference notice does.
+        ordered = sorted(counts.items(), key=lambda item: (-item[1], item[0]))
+        listed = ordered[:COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES]
+        lines: list[str] = []
+        current_chars = len(prefix)
+        omitted = len(ordered) - len(listed)
+        for name, count in listed:
+            clamped = name[:COMPACT_DROPPED_TOOL_NAME_MAX_CHARS]
+            line = f"- {clamped} x{count}" if count > 1 else f"- {clamped}"
+            if current_chars + len(line) + 1 > COMPACT_DROPPED_TOOL_NOTICE_MAX_CHARS:
+                omitted += 1
+                continue
+            lines.append(line)
+            current_chars += len(line) + 1
+        if omitted:
+            name_label = "name" if omitted == 1 else "names"
+            lines.append(
+                f"- ... {omitted} additional distinct tool {name_label} omitted"
+            )
         return prefix + "\n".join(lines)
 
     def _latest_visible_user_message(self) -> Message | None:
--- a/tests/core/agent/test_context.py
+++ b/tests/core/agent/test_context.py
@@ -27,7 +27,7 @@ from xagent.core.agent.language import (
     response_language_rules,
 )
 from xagent.core.agent.utils.context_builder import ContextBuilder
-from xagent.core.context_ref import SUPERSEDES_SCOPE_KEY
+from xagent.core.context_ref import CONTEXT_REFS_KEY, SUPERSEDES_SCOPE_KEY
 from xagent.web.user_isolated_memory import current_user_id
 
 
@@ -760,7 +760,8 @@ def test_compact_with_llm_summarizes_history_and_preserves_current_user() -> Non
     assert "current execution state" in ctx.messages[0].content
     assert "do not repeat completed tool calls" in ctx.messages[0].content
     assert "lost in compaction" in ctx.messages[0].content
-    assert "re-run the tool that produced it" in ctx.messages[0].content
+    assert "re-read or re-query the source" in ctx.messages[0].content
+    assert "Only re-run tools that read" in ctx.messages[0].content
     assert "- read_file" in ctx.messages[0].content
     assert result.metadata["dropped_tool_result_count"] == 1
     assert ctx.messages[1].role == "user"
@@ -795,12 +796,16 @@ def test_compact_with_llm_reports_dropped_tool_results_by_name() -> None:
     result = ctx.compact_with_llm_response({"content": "Collected KPI inputs."})
 
     notice = ctx.messages[0].content
-    assert "3 completed tool call(s) were dropped" not in notice
-    assert "4 completed tool call(s) were dropped" in notice
+    assert "3 tool calls were dropped" not in notice
+    assert "4 tool calls were dropped" in notice
     assert "- web_search x3" in notice
     assert "- read_file" in notice
     assert "unavailable rather than recalled" in notice
     assert result.metadata["dropped_tool_result_count"] == 4
+    assert result.metadata["dropped_tool_results_by_name"] == {
+        "web_search": 3,
+        "read_file": 1,
+    }
 
 
 def test_compact_with_llm_omits_tool_notice_without_tool_results() -> None:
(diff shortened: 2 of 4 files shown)


src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,49 @@ def _dropped_context_refs_notice(
...
+        before compaction, so reporting them as newly dropped would be wrong.
+        """
+        counts: dict[str, int] = {}
+        for message in messages:
+            if message.hidden or message.role != "tool":
+                continue
+            metadata = message.metadata or {}
+            name = str(metadata.get("tool_name") or "").strip() or "unnamed tool"
+            counts[name] = counts.get(name, 0) + 1
+        return counts
```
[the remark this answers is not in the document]
codeacme17: Correct, and thank you — this was a real bug, not just an inaccurate docstring. Fixed in e105a2d.

Verified all three links in the chain:

1. `_compact_superseded_tool_messages` (execution.py:208-225) calls `replace(message, content=..., metadata={..., "superseded": True, "raw_result": {...}})` — it never touches `hidden`, exactly as you said.
2. `get_messages_for_llm` filters on `hidden` only (execution.py:376), so superseded placeholders **are** still sent to the model. My docstring's "already absent from the prompt" was false for precisely the case that matters.
3. `hidden=True` is never set anywhere in `src/` — a repo-wide grep returns nothing outside tests. The `hidden` guard was dead code in production, which is why the miss went unnoticed.

The impact was worse than a cosmetic over-count. `ComputerTool` stamps `SUPERSEDES_SCOPE_KEY` on *every* result (computer.py:471), so in any browser session all but the newest observation is superseded. Measured on a 3-step session:

```
hidden flags : [False, False, False]
superseded   : [True, True, False]
old filter counted: 3
new filter counts : 1
```

A 20-step session would have announced "Raw observations from 20 completed tool call(s) were dropped" when 19 had already lost their `raw_result` and only 1 carried evidence. The notice exists to tell the agent *how much* evidence compaction destroyed, so a 20x inflation inverts its purpose — and it would push the agent to re-run calls whose results were deliberately discarded.

Applied your fix (metadata hoisted above the guard so it compiles), and rewrote the docstring rather than just extending it, since the original rationale was wrong: hidden messages are skipped because they are not in the prompt; superseded ones because their raw result was already gone before compaction ran.

I also replaced the test. `test_compact_with_llm_ignores_hidden_tool_results_in_notice` set `hidden=True` by hand — a state production never produces — so it was green while covering nothing. It now drives the real `_xagent_supersedes_scope` path and asserts `1 completed tool call(s)` / `dropped_tool_result_count == 1`; the hidden case is kept, renamed, as a regression guard on the flag itself.

One related gap I did **not** fix, flagged for the record: `_sanitize_tool_message_pairs` drops orphan `tool` messages that lack a matching assistant `tool_calls` envelope, and those are still counted, since `_dropped_tool_result_counts` runs over raw `self.messages`. That over-count errs toward "re-run the tool", so it is conservative rather than misleading, and fixing it means running the candidate list through sanitization with its own edge cases. Happy to fold it in here if you'd prefer.

codeacme17:
Thanks for the review — the supersession finding was correct and caught a real bug, not just a stale docstring. Addressed in e105a2d; details in the inline thread.

Summary: `Message.hidden` is never set anywhere in `src/`, so the filter I wrote was dead code in production, while the actual eviction mechanism — `metadata["superseded"]` — went uncounted. Because `ComputerTool` stamps a supersedes scope on every result, a 20-step browser session would have reported 20 dropped observations when only 1 still carried a raw result. Since the whole point of the notice is to calibrate how much evidence compaction destroyed, that inflation inverted its purpose.

Fixed the filter, corrected the docstring rationale, and replaced the test that faked `hidden=True` (a state production never reaches) with one that drives the real `_xagent_supersedes_scope` path.

rogercloud:
ci is failing

tests/core/agent/test_context.py
```diff
@@ -757,10 +759,133 @@ class CompactLLM:
...
+        ],
+    )
+    ctx.add_tool_result("web_search", {"output": "dropped rows"}, "call-1")
+    for index in range(6):
+        ctx.add_user_message(f"tail-{index}")
+
+    result = ctx.compact_if_needed()
+
+    assert result.strategy == "truncate"
+    assert result.metadata["dropped_tool_result_count"] == 1
```
rogercloud, line 1046: Major (test coverage): this is the only test for the truncate dropped-count, and its retained window is a pure suffix of the original list — so it does not discriminate the identity-diff implementation from the naive prefix slice the code comment above `retained_ids` explicitly warns against. A counterfactual prefix-slice implementation run against these same inputs produces an identical result, so the PR's most subtle design decision currently has zero regression protection.

Please add a case where `_tail_window_preserving_tool_pairs` / `_sanitize_tool_message_pairs` excises an *interior* tool-pair fragment from the window, so a future refactor back to slicing fails the suite instead of passing it.

src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,53 @@ def _dropped_context_refs_notice(
...
+        listed = ordered[:COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES]
+        lines = [
+            f"- {name} x{count}" if count > 1 else f"- {name}" for name, count in listed
+        ]
+        omitted = len(ordered) - len(listed)
+        if omitted:
+            lines.append(f"- ... {omitted} additional tool(s) omitted")
+        total = sum(counts.values())
+        prefix = (
+            f"Raw observations from {total} completed tool call(s) were dropped by "
```
rogercloud: Minor (correctness of the emitted prompt): `_dropped_tool_result_counts` applies no `success`/`status` check to `raw_result`, so failed and cancelled tool calls are counted here and announced as "N **completed** tool call(s)". That collides with the instruction a few lines above — "do not repeat completed tool calls" — so a model can read a *failed* dropped call as already-succeeded and skip the retry, which inverts the intent of this whole notice.

Suggest dropping "completed" ("Raw observations from N tool call(s) were dropped..."), or filtering non-success results out of the count.

Separately, this prefix is always plural ("tool call(s)") while the per-tool lines below correctly switch on `count > 1`. Worth making consistent in the same change — note both edits touch the exact-string assertion in `tests/core/agent/test_context.py` ("1 completed tool call(s) were dropped").

src/xagent/core/agent/context/execution.py
```diff
@@ -1049,18 +1064,27 @@ def compact_with_llm_response(
             self._context_refs_removed_by_compaction(latest_user)
         )
         dropped_refs_notice = self._dropped_context_refs_notice(dropped_context_refs)
+        dropped_tool_counts = self._dropped_tool_result_counts(self.messages)
```
rogercloud, line 1083: Minor (fragility): unlike the `truncate` path, this passes the whole `self.messages` list rather than an identity diff, which is only correct because `next_messages` never retains a tool-role message. That holds today — it's `[summary_message]` plus at most a strictly `role == "user"`-filtered message — but the invariant is unstated and untested, so a later "retain the last tool pair" optimization would silently turn this into an overcount. A one-line comment naming the invariant, or computing this path as an identity diff for symmetry, would make it robust.

Related overcount in the same call: a tool message whose image `context_refs` are retained by `_context_refs_removed_by_compaction` (and therefore still rematerializable) is still counted as fully dropped here. Low impact since it errs toward caution, but it's the same class of overstatement the `superseded` exclusion was added to fix.

src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,53 @@ def _dropped_context_refs_notice(
...
+    def _dropped_tool_results_notice(counts: dict[str, int]) -> str:
+        """Describe the tool observations this compaction removes from context.
+
+        Without this, the summary silently replaces every retrieved value and
+        the agent cannot tell a remembered figure from an invented one.
+        """
+        if not counts:
+            return ""
+        ordered = sorted(counts.items(), key=lambda item: (-item[1], item[0]))
+        listed = ordered[:COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES]
```
rogercloud: Minor: this caps by item count (20 distinct names) with no character budget, unlike the sibling `_dropped_context_refs_notice`, which budgets by `COMPACT_DROPPED_REF_NOTICE_MAX_CHARS`. `tool_name` gets only `.strip()` — no length or newline sanitization — and for MCP tools it comes from dynamic server config rather than a fixed internal registry, so the notice size is effectively unbounded and there's a weak prompt-injection-adjacent surface here (weak because that actor already controls the tool's output).

Suggest a char budget matching the refs notice, or at minimum a per-name length clamp.

src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,53 @@ def _dropped_context_refs_notice(
...
+        lines = [
+            f"- {name} x{count}" if count > 1 else f"- {name}" for name, count in listed
+        ]
+        omitted = len(ordered) - len(listed)
+        if omitted:
+            lines.append(f"- ... {omitted} additional tool(s) omitted")
```
rogercloud: Minor (wording): `omitted` counts distinct tool *names*, while `total` in the prefix counts tool *calls* — two different units in the same notice with nothing to disambiguate them. Suggest "... N additional distinct tool name(s) omitted".

src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,53 @@ def _dropped_context_refs_notice(
...
+    def _dropped_tool_result_counts(messages: list[Message]) -> dict[str, int]:
+        """Count tool observations whose raw result this compaction discards.
+
+        Superseded observations are excluded: a later observation already
+        replaced their content and ``raw_result``, so compacting them away
+        destroys no evidence and counting them would overstate the loss.
+        Hidden messages are excluded because they are not in the prompt at all.
+        """
+        counts: dict[str, int] = {}
+        for message in messages:
```
rogercloud, line 1216: Minor (readability, no behavior change): the `hidden` / `superseded` check runs before the `role != "tool"` check, which reads as if those flags matter for every role when only tool-role messages are relevant here. Reordering so the role check comes first makes the filter's intent obvious.

src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,53 @@ def _dropped_context_refs_notice(
...
+            metadata = message.metadata or {}
+            if message.hidden or metadata.get("superseded"):
+                continue
+            if message.role != "tool":
+                continue
+            name = str(metadata.get("tool_name") or "").strip() or "unnamed tool"
```
rogercloud: Minor (test coverage): the `"unnamed tool"` fallback is untested — no test supplies a tool message with missing or empty `tool_name` metadata. One extra case in the by-name notice test would cover it.

(For the record, the `.strip() or ...` shape is right here and shouldn't be simplified to the plainer `str(x or "tool")` pattern used elsewhere in this file: this value is a dict aggregation key, so a whitespace-only name would silently fragment counts.)

src/xagent/core/agent/context/execution.py
```diff
@@ -1153,6 +1178,53 @@ def _dropped_context_refs_notice(
...
+            counts[name] = counts.get(name, 0) + 1
```
rogercloud: Simplification: this accumulation is `collections.Counter` — `counts = Counter(...)` over the filtered names removes the manual `get`/increment.

Scope it to this function only: do **not** switch `_dropped_tool_results_notice` to `Counter.most_common()`, because that loses the explicit alphabetical tiebreak in `sorted(..., key=lambda item: (-item[1], item[0]))` for equal counts. Untested today, but a real behavior change.

net: ~2-4 lines. Readability, not a meaningful reduction.

src/xagent/core/agent/context/execution.py
```diff
@@ -997,14 +998,28 @@ def _compact(self, llm: Any = None) -> CompactResult:
...
+            self.messages = retained
             return CompactResult(
                 compacted=True,
                 original_count=original_count,
                 final_count=len(self.messages),
                 strategy="truncate",
-                metadata={"removed_count": removed},
+                metadata={
+                    "removed_count": removed,
+                    "dropped_tool_result_count": sum(dropped_tool_counts.values()),
```
rogercloud, line 1033: Minor (optional, observability): the full per-tool-name breakdown is computed just above and then reduced to a sum here, discarding the breakdown. `CompactResult.metadata` has no schema that would reject a nested dict, so storing the per-tool counts alongside the total is free — and it's exactly the detail you'd want when diagnosing another incident like the one that motivated this PR.

tests/core/agent/test_context.py
```diff
@@ -757,10 +759,133 @@ class CompactLLM:
...
+
+    result = ctx.compact_with_llm_response({"content": "Collected KPI inputs."})
+
+    notice = ctx.messages[0].content
+    assert "3 completed tool call(s) were dropped" not in notice
+    assert "4 completed tool call(s) were dropped" in notice
+    assert "- web_search x3" in notice
+    assert "- read_file" in notice
+    assert "unavailable rather than recalled" in notice
+    assert result.metadata["dropped_tool_result_count"] == 4
```
rogercloud, line 804: Minor (test coverage), two gaps around this test:

1. Nothing constructs more than `COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES` distinct dropped tool names, so the cap-and-omit branch (the `ordered[:20]` slice and the omitted-count line) is never exercised. Low-risk code, but a one-line case covers it.
2. Nothing asserts the ordering guarantee that the refs notice is appended before the tools notice when both are present.

Also not covered anywhere: that the `truncate` strategy inserts *no* in-prompt notice message. Metadata-only is the intentional contract there, so it's worth pinning so a future change can't quietly add one.

tests/core/agent/test_context.py
```diff
@@ -757,10 +759,133 @@ class CompactLLM:
...
+            "computer",
+            {"output": f"view {index}", SUPERSEDES_SCOPE_KEY: "computer:task-1"},
+            call_id,
+        )
+
+    result = ctx.compact_with_llm_response({"content": "Inspected the dashboard."})
+
+    notice = ctx.messages[0].content
+    assert "1 completed tool call(s) were dropped" in notice
+    assert "- computer\n" in f"{notice}\n"
```
rogercloud: Minor (test quality): `assert "- computer\n" in f"{notice}\n"` passes only because the tools notice happens to be the last thing appended to the message content (no trailing newline, appended after the refs notice). That's a positional coincidence rather than a check on notice format — it would break for unrelated cosmetic reasons the moment anything is appended after this notice.

The next assertion (`"- computer x" not in notice`) already expresses the real intent — "listed once, not aggregated with a multiplier". Suggest keeping that plus an explicit count assertion and dropping the newline-anchored form.

rogercloud (approved):
## Summary

This PR narrows the confabulation trap described in #1235 by making context compaction *disclose* what it destroys, without yet changing how much survives. Concretely, in `src/xagent/core/agent/context/execution.py` it (a) softens the post-compaction instruction so the model is explicitly permitted — and told — to re-run the tool that produced a value when that exact value is not literally in the summary, (b) appends a "dropped tool results" notice to the summary message, aggregated by tool name with counts and excluding `hidden` and `superseded` observations, and (c) records `dropped_tool_result_count` in `CompactResult.metadata` for both the `llm_summary` and `truncate` strategies. For `truncate` the dropped set is computed by object-identity diff against the retained window rather than by prefix-slicing, because `_tail_window_preserving_tool_pairs` can excise interior tool-pair fragments.

I read the scope-limiting choices in the PR body as deliberate and reasonable: no in-prompt notice on the `truncate` path (it would break `truncate`'s exact-count contract and would itself be truncated later), no summary-budget scaling, and no retention/rematerialization of raw observations. None of those are treated as gaps below.

The mechanism is correct and appropriately minimal. Identity-diffing for `truncate` is the right call, the `superseded` exclusion (already fixed in `e105a2d7` after an earlier bot/author round) is the right semantics, and the metadata addition is cheap and useful. My reservations are about the *prompt* this mechanism emits, not about the bookkeeping:

- **The notice invites re-running side-effecting tools (moderate).** `_dropped_tool_results_notice` lists any dropped tool by name, and the softened instruction pairs that with "re-run the tool that produced it" plus an explicit sanction to "regenerate completed artifacts." Nothing distinguishes read-only retrieval from mutation: `write_file`, `python_executor`, `javascript_executor`, and the excel/pptx writers all funnel through the same generic `add_tool_result` path and appear in the notice identically to a search or a query. The codebase already carries an idempotency signal for exactly this kind of decision — the `concurrency_safe` tool metadata consulted by `_tool_is_concurrency_safe` in `react.py` — and it is not consulted here. Suggestion: gate which dropped tools get named under the "re-run" framing on that existing flag, or at minimum keep known-mutating tools out of the re-run sentence.

- **Control-flow pseudo-tools are counted and named like data tools (minor-to-moderate).** `final_answer`, `send_message`, `ask_user_question`, and `load_skill` also go through `add_tool_result`, so they land in the notice and in the headline count. "Re-running" those means terminating the run or re-contacting the user, not re-fetching evidence, and they inflate the drop total. This is reachable in ordinary multi-turn sessions where such messages survive into a later compaction. A short denylist of known control-tool names would fix both the advice and the count.

- **The escape clause has no counterpart in the summarizer prompt (non-blocking suggestion).** `_build_llm_compact_prompt` is unchanged and is asked to preserve handles (file ids, paths, URLs) while dropping "irrelevant raw payloads" — it is never asked to preserve exact figures or statistics. Since the new clause fires whenever a needed value is *not literally in the summary*, the practical effect skews toward "re-run broadly after every compaction" rather than "re-ground the one missing number." I'm raising this as a suggestion rather than a defect because the PR body already names this trade-off ("re-verify more, fabricate less") as intentional and defers the real fix. Consider adding a line to the summarizer prompt asking it to carry salient figures verbatim — that would make the escape clause targeted rather than blanket.

- **Pre-existing, out of scope:** `_context_refs_removed_by_compaction` (the sibling of the new counting code, a few lines above it) does *not* exclude `superseded` messages, while the new tool-counting code correctly does — worth a follow-up issue for consistency, not a change to make here.

## ⚠️ Major finding — flagging for visibility despite approving

**`tests/core/agent/test_context.py` — the `truncate` identity-diff has no discriminating test.** The PR's most subtle reasoning is that a prefix slice would be wrong because `_sanitize_tool_message_pairs` can excise an interior tool-pair fragment from the retained window. No test exercises that case: in `test_compact_truncate_counts_dropped_tool_results` the retained window is a pure suffix, and I confirmed that a naive prefix-slice implementation produces an identical result on those inputs. So the load-bearing design decision currently has zero regression protection — a future refactor back to slicing would stay green.

This is not a runtime bug — the current implementation is correct — but it is a coverage gap on the PR's own stated reason for choosing the more complex approach. **Strongly recommend adding a case where the window drops an interior fragment**, ideally before merge or as an immediate fast-follow, so a future edit can't silently regress this without the test suite catching it.

## Other line-level findings (minor)

1. **`execution.py:1220`** — "completed" is wrong for failed/cancelled calls. `_dropped_tool_result_counts` applies no `success`/`status` check to `raw_result`, so failed and cancelled tool calls are counted, yet the notice announces "N *completed* tool call(s) were dropped." That collides with the instruction a few lines above ("do not repeat completed tool calls"): a model can read a *failed* call as already-succeeded and skip the retry, the exact inversion of what we want. Drop the word "completed", or filter non-success results out of the count. (The always-plural "tool call(s)" prefix is also inconsistent with the correctly singular/plural per-tool lines; both fixes touch `tests/core/agent/test_context.py:844`.)
2. **`execution.py:1067`** — `llm_summary` path counts against the full list, not a diff. Unlike `truncate`, this path passes `self.messages` wholesale and relies on the unstated invariant that `next_messages` never retains a tool-role message. That holds today but nothing pins it; a later "retain the last tool pair" optimization would silently turn this into an overcount. A one-line comment stating the invariant — or computing this path as an identity diff too — would make the two paths symmetric.
3. **`execution.py:1067` vs `:1091`** — a tool message with retained `context_refs` is still counted as fully dropped. Low impact (it errs toward caution), but it is the same overcounting class the `superseded` exclusion was added to avoid.
4. **`execution.py:1211`** — `COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES` caps items, not characters, unlike the sibling `_dropped_context_refs_notice`. `tool_name` receives only `.strip()`, and for MCP tools it originates from dynamic server config rather than a fixed internal registry — so notice size is technically unbounded and there's a weak prompt-injection-adjacent surface. A char budget matching the refs notice, or a per-name length clamp, closes it.
5. **`execution.py:1217`** — mixed units in the omission line: `omitted` counts distinct names, `total` counts calls. Suggest "N additional distinct tool name(s) omitted."
6. **`execution.py:1191`** — check order (`hidden`/`superseded` before `role != "tool"`) reads misleadingly, though behavior is identical. Reordering the role check first makes intent obvious.
7. **`execution.py:1021`** — the per-tool breakdown is computed then discarded on the `truncate` path; only the sum reaches metadata. Storing the breakdown would be free observability for diagnosing exactly this class of incident. Optional.
8. **`tests/core/agent/test_context.py:845`** — `assert "- computer\n" in f"{notice}\n"` passes only because the tools notice happens to be appended last with no trailing newline — a positional coincidence, not a real check on notice format. `assert "- computer x" not in notice` plus an explicit count assertion expresses the intent directly.
9. **`tests/core/agent/test_context.py:803`** — two uncovered notice behaviours: the cap-and-omit branch at 20+ distinct tool names, and the ordering guarantee that the refs notice is appended before the tools notice when both are present.
10. **`execution.py:1197`** — the `"unnamed tool"` fallback is untested.
11. **`tests/core/agent/test_context.py`** — no test asserts the `truncate` path inserts no in-prompt notice, which is the intentional contract there.

- `execution.py:1190-1198` — the manual `counts[name] = counts.get(name, 0) + 1` accumulation in `_dropped_tool_result_counts` is `collections.Counter`. Safe to swap. Keep the change to the accumulation only: do **not** convert `_dropped_tool_results_notice` to `Counter.most_common()`, since that would drop the explicit alphabetical tiebreak in `sorted(..., key=lambda item: (-item[1], item[0]))` on equal counts — untested today, but a real behaviour change.

net: ~2-4 lines removed. Small; readability, not a meaningful reduction.

## Test results

`uv run pytest tests/core/agent/test_context.py tests/core/test_context_ref.py tests/core/test_context_materializer.py -q` → **109 passed**. `ruff format --check`, `ruff check`, and `mypy` on the touched files are clean.

## Recommendation

Approving — the change is well-scoped, does what it claims, and there is no runtime bug. That said, the major item above is worth calling out clearly: it's a coverage gap on this PR's own stated reason for choosing the more complex (correct) implementation over a simpler one, so please prioritize closing it soon. The two design points about mutating/control-flow tools appearing under a "re-run it" instruction are also worth a decision at some point, since they affect what the agent is told to *execute*, not just what it's told to say. Everything else is minor polish.

tests/core/agent/test_context.py
codeacme17: Confirmed and fixed in fa2903e. You were right on both halves, and I verified the counterfactual rather than taking it on faith.

The interior-excision case is constructible — two paths in `_sanitize_tool_message_pairs` drop from the interior: an assistant block whose `expected_ids` aren't a subset of `received_ids` is dropped *together with its tool messages* (`execution.py:1400-1407`), and an orphan `tool` message is dropped anywhere (`:1379-1380`). The boundary backoff only repairs the edge, never the middle.

```
0 user      "u0"
1 user      "u1"
2 assistant tool_calls=[a1:web_search, a2:web_search]
3 tool      tool_call_id=a1
4 user      "u2"
5 user      "u3"          max_messages=5 -> start=1
```

Retained is `[u1, u2, u3]` — `messages[1:]` minus the incomplete tool block excised from its middle, so not a suffix. Identity diff counts 1; prefix slice `messages[:3]` = `[u0, u1, assistant]` counts **0** — zero evidence loss reported on exactly the case the feature exists for.

Added as `test_compact_truncate_counts_tool_result_excised_from_window_interior`, and verified it discriminates: I swapped the implementation for the naive prefix slice and it fails `assert 0 == 1`, while the pre-existing suffix-only test stays green — your exact point.

Worth noting it's production-reachable, not just theoretical: `_execute_pending_tool_calls` returns on `_interrupt_if_requested(label="before_tool")` (`react.py:2107-2114`) with unexecuted calls from the same assistant message and no `add_tool_result` backfill. The pause path (`_discard_pending_tool_plan_after_pause`, `react.py:1882-1903`) does backfill; the interrupt path doesn't.

src/xagent/core/agent/context/execution.py
codeacme17: Both fixed in fa2903e — success filter *and* wording, because they close different holes.

Failed/cancelled calls are now filtered with `tool_result_succeeded` (`result.py:55-63`), which is already applied to this exact field one layer up (`ReActPattern._latest_tool_result_success`, `react.py:2676-2682`). `raw_result` is stamped on every tool message (`execution.py:191`) and `sanitize_tool_result_for_public_context` strips only path keys, so `success`/`status`/`is_error` survive intact. Both real failure shapes classify correctly: `{"success": False, "error": ...}` from `_execute_tool_safely` and `{"success": False, "status": "cancelled"}` from `_discard_pending_tool_plan_after_pause`.

The wording change isn't redundant with the filter: `status: "waiting_for_user"` results (`user_interaction.py:36-42`) are neither failed nor completed and still pass `tool_result_succeeded`. The prefix is now "Raw observations from N tool call(s) was/were dropped" — no "completed", and singular/plural on both the noun and the verb, matching the per-tool lines.

New test `test_compact_with_llm_excludes_failed_tool_results_from_notice` covers all three failure shapes plus one success, and asserts the old phrasing is gone.

src/xagent/core/agent/context/execution.py
codeacme17: Comment added in fa2903e rather than an identity diff — the invariant holds, so a diff here would be provably dead code.

`next_messages` is `[summary_message]` (a `Message.role_system`, `:1088`) plus at most `_latest_visible_user_message`, which filters `role == "user"` (`:1228-1233`). No tool-role message can survive, so the whole-list count *is* the diff numerically. The comment now names that explicitly and says why `truncate` needs a real diff and this doesn't, so the next reader doesn't have to re-derive it — and a "retain the last tool pair" change would land on a stated invariant instead of silent luck.

On the related overcount: I left it. A `ContextReference` is a durable *file handle*, not the tool's textual observation — the figures and table rows the notice exists to protect are gone even when the image survives. `dropped_tool_result_count` is a text-evidence count; retained handles are separately disclosed by `_dropped_context_refs_notice`, and the two notices are now order-pinned by a test. Excluding ref-retained messages would make the number mean something other than what the notice claims, and would understate evidence loss.

src/xagent/core/agent/context/execution.py
codeacme17: Fixed in fa2903e — now "N additional distinct tool name(s) omitted", with singular/plural on the noun. The prefix counts calls and this line counts names, so the units are stated explicitly in both places now.

src/xagent/core/agent/context/execution.py
codeacme17: Reordered in fa2903e — `role != "tool"` now runs first, then the `hidden`/`superseded` metadata checks. Behaviour is identical; the filter reads as what it is.

src/xagent/core/agent/context/execution.py
codeacme17: Fixed in fa2903e. New test `test_compact_with_llm_names_tool_result_without_tool_name` passes a whitespace-only `tool_name` and asserts both the `- unnamed tool` line and the metadata key.

Agreed on keeping the `.strip() or ...` shape — it is a dict aggregation key, so a whitespace-only name would fragment counts. It is now `" ".join(name.split())`, which additionally collapses interior newlines (see the notice-bounding thread); the empty-to-fallback behaviour is unchanged.

src/xagent/core/agent/context/execution.py
codeacme17: Applied in fa2903e, scoped exactly as you said: `_dropped_tool_result_counts` now builds a list of names and returns `dict(Counter(names))`. `_dropped_tool_results_notice` keeps its explicit `sorted(..., key=lambda item: (-item[1], item[0]))` so the alphabetical tiebreak on equal counts is preserved — `Counter.most_common()` would have made that insertion-ordered.

src/xagent/core/agent/context/execution.py
codeacme17: Added in fa2903e — `dropped_tool_results_by_name` now carries the full per-tool breakdown alongside the total, on both the `llm_summary` and `truncate` paths. Agreed it is exactly the detail you would want when diagnosing another incident like this one; three of the new tests assert against it.

src/xagent/core/agent/context/execution.py
codeacme17: Fixed in fa2903e with a char budget mirroring the sibling, plus a per-name clamp (`COMPACT_DROPPED_TOOL_NOTICE_MAX_CHARS = 1024`, `COMPACT_DROPPED_TOOL_NAME_MAX_CHARS = 64`); the item cap stays as a secondary bound. Names are also whitespace-collapsed via `" ".join(name.split())`, which neutralizes the newline case locally.

On severity, I'd push back slightly: `_format_tool_result` already builds every tool message as `f"Tool {tool_name} returned: {formatted}"` with the same unsanitized name (`execution.py:274-281`), once per call, before compaction runs — and `MCPToolAdapter.name` only does `.replace(" ", "_").replace("-", "_")` (`mcp_adapter.py:410-415`), so newlines and colons already survive into context by a much wider path than one line per distinct name here. That's a real pre-existing weak spot, but this PR doesn't meaningfully widen it, and proper name sanitization belongs in the MCP adapter.

The unbounded *length* was the genuine problem — 20 names × unbounded length inside a routine whose whole purpose is a token budget — so that's what the fix targets. New test builds a 200-char MCP-style name plus 25 distinct tools and asserts the long name is absent, the omission line appears, and the notice stays bounded.

tests/core/agent/test_context.py
codeacme17: Fixed in fa2903e — dropped the newline-anchored form and replaced it with the count assertion plus the explicit `dropped_tool_results_by_name` metadata check, so the intent ("listed once, not aggregated with a multiplier") is asserted directly rather than by position.

The ordering it depended on is now a real test of its own rather than an implicit assumption (`test_compact_with_llm_orders_ref_notice_before_tool_notice`).

tests/core/agent/test_context.py
codeacme17: All three gaps closed in fa2903e:

1. `test_compact_with_llm_caps_and_clamps_dropped_tool_names` — 26 distinct tools, one of them a 200-char MCP-style name, exercising both the item cap and the new char budget and per-name clamp.
2. `test_compact_with_llm_orders_ref_notice_before_tool_notice` — asserts the refs notice precedes the tools notice by index when both are present.
3. `test_compact_truncate_adds_no_in_prompt_notice` — pins the metadata-only contract on the `truncate` path, asserting the exact message count is preserved and no message carries the notice text.

codeacme17:
Thanks — thorough review. All findings are addressed in fa2903e; replies are on each thread. The three worth surfacing here:

**The Major test gap was real, and I verified the counterfactual instead of trusting it.** The interior-excision case is constructible (an assistant block with unanswered tool calls is excised together with its tool messages from the middle of the window), so I added it — then swapped the implementation for the naive prefix slice and confirmed the new test fails `0 == 1` while the pre-existing suffix-only test stays green. It's also production-reachable, not theoretical: the `before_tool` interrupt path returns with unexecuted calls and, unlike the pause path, does not backfill results.

**`concurrency_safe` turned out to be the wrong flag, so I fixed POINT A differently.** It means parallel-execution safety within a turn, not idempotency — and an MCP *operator* can declare an entire server concurrency-safe, so a Sheets-write tool on such a server would be marked safe to re-run. `read_only` is the flag that actually carries that meaning. But neither is reachable here: `ExecutionContext` has no tool registry, and `add_tool_result` stamps only `tool_name`/`raw_result`/workspace fields. So this landed as prompt-only narrowing — "only re-run tools that read; if the value came from a tool that writes, sends, or executes, re-read the artifact it produced or report the value unavailable." Plumbing `read_only` through `add_tool_result` would be the real fix and is a reasonable follow-up, but it's a six-call-site signature change and scope creep here.

Also worth noting the "regenerate completed artifacts" sanction is pre-existing text — this PR added the *exception* to it, so it widened an existing risk rather than introducing one. The narrowing now bounds both.

**Control pseudo-tools (POINT B) were worse than cosmetic.** `final_answer` lands on essentially every completed turn and `send_message`/`ask_user_question` on every interaction pause, so they inflated the count *and* sat under an instruction to re-run them. Fixed with a `CONTROL_TOOL_NAMES` constant hoisted out of `ReActPattern._control_tool_names()` (which now returns it, so they can't drift), plus `load_skill` — kept as a separate `NON_EVIDENCE_TOOL_NAMES` set, since `load_skill` isn't a control tool for the ReAct scheduler and widening the shared set would be wrong.

Two things I deliberately did **not** change, both explained on their threads: the `llm_summary` whole-list count (the invariant holds, so a diff would be dead code — comment added instead), and the ref-retained overcount (a `ContextReference` is a durable handle, not the textual observation; excluding those would make the number understate text-evidence loss).

On your out-of-scope note: `_context_refs_removed_by_compaction` not excluding `superseded` is actually correct rather than inconsistent — `_compact_superseded_tool_messages` replaces `content` and `raw_result` but leaves `context_refs` untouched, so a superseded message's file handles are still live and worth retaining. Aligning the two would lose durable handles. Worth documenting rather than "fixing", happy to do that separately.

One incidental finding: `Message.hidden` is never set to `True` anywhere in `src/` — it's test-only today, so the `hidden` branches in both functions are dead in production. Harmless as a regression guard, but `superseded` (from e105a2d) is what actually does the work.

`tests/core/agent/` passes (69 in `test_context.py`); `pre-commit run --files` clean on all four touched files.

Not addressed here, by design: your suggestion to have the summarizer prompt preserve salient figures verbatim. That's the right idea and would make the escape clause targeted rather than blanket, but it belongs with the summary-budget scaling work — both change what the summary *contains* rather than what it *discloses*, which is this PR's line. Tracking it for the next proposal-B PR.

PR #1240, opened 2026-08-10
fix: raise HubSpot API request timeout for large contact-to-deal lookups

yiboyasss:
## Summary
- Raises the HubSpot connector's shared HTTP request timeout (`DEFAULT_TIMEOUT_SECONDS`) from 30s to 90s.
- Adam reported (7/29) that pulling contact-to-deal associations for the ANZ HubSpot portal (~53,563 contacts) timed out at the previous 30s limit, blocking the TMS/FMS contact split.

## Test plan
- [x] `pytest tests/web/tools/test_hubspot_mcp.py`
- [x] `ruff check` / `ruff format --check` / `mypy` on the touched file

### DIFF
--- a/src/xagent/web/tools/mcp/hubspot.py
+++ b/src/xagent/web/tools/mcp/hubspot.py
@@ -17,10 +17,12 @@ setup_proxy_env()
 mcp = FastMCP("hubspot-mcp")
 
 HUBSPOT_BASE_URL = "https://api.hubapi.com"
+DEFAULT_TIMEOUT_SECONDS = 30
 # Association listing (e.g. contact-to-deal) can be slow to respond on
 # portals with tens of thousands of contacts; 30s was observed timing out
-# on those pulls.
-DEFAULT_TIMEOUT_SECONDS = 90
+# on those pulls. Scoped to that call path so a slow/degraded HubSpot API
+# doesn't also make every other tool call hang for the same longer window.
+ASSOCIATION_LISTING_TIMEOUT_SECONDS = 90
 
 DEFAULT_CONTACT_PROPERTIES = [
     "email",
@@ -81,6 +83,7 @@ def _request(
     *,
     params: dict[str, Any] | None = None,
     body: dict[str, Any] | None = None,
+    timeout: int = DEFAULT_TIMEOUT_SECONDS,
 ) -> Any:
     response = requests.request(
         method=method,
@@ -88,7 +91,7 @@ def _request(
         headers=_headers(),
         params=params,
         json=body,
-        timeout=DEFAULT_TIMEOUT_SECONDS,
+        timeout=timeout,
     )
     try:
         response.raise_for_status()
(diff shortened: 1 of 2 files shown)


yiboyasss:
Good catch — agreed that raising the shared default would make every tool call (contact/company/deal reads and writes) hang for up to 90s during a HubSpot outage, not just the slow path this was meant to fix.

Reverted \`DEFAULT_TIMEOUT_SECONDS\` to 30s and added a separate \`ASSOCIATION_LISTING_TIMEOUT_SECONDS = 90\`, applied only to \`_list_association_ids\` (the contact-to-deal/notes pagination path Adam hit). Everything else keeps the original 30s fast-fail behavior.

src/xagent/web/tools/mcp/hubspot.py
```diff
@@ -18,6 +18,11 @@
 
 HUBSPOT_BASE_URL = "https://api.hubapi.com"
 DEFAULT_TIMEOUT_SECONDS = 30
+# Association listing (e.g. contact-to-deal) can be slow to respond on
```
rogercloud: Minor, non-blocking: the comment attributes the slowness to "portals with tens of thousands of contacts," but `_list_association_ids`/`hubspot_get_contact_deals` scope to a single contact and cap results at `max_results` (default 100) per call — one call isn't inherently slow just because the portal is large. If this was observed under a bulk/looped pull across many contacts in that portal (which the PR description suggests), it'd help to say so explicitly, and to cite the observed latency that justified 90s specifically rather than 60s or another value.

rogercloud (approved):
**Summary**

This PR fixes a reported production timeout: pulling contact-to-deal associations for a large HubSpot portal (~53k contacts) was hitting the connector's 30s HTTP timeout, blocking a contact-split workflow. The original version of this PR raised the shared `DEFAULT_TIMEOUT_SECONDS` constant globally from 30s to 90s, which would have also made every other tool call (single-record get/create/update for contacts, companies, notes) tolerate hanging for up to 90s during a HubSpot outage.

**Update since last look**

The latest commit (`fix: scope the longer HubSpot timeout to association listing only`) addresses that concern directly: `DEFAULT_TIMEOUT_SECONDS` stays at 30, a new `ASSOCIATION_LISTING_TIMEOUT_SECONDS = 90` is introduced, `_request()` now takes a `timeout` parameter defaulting to 30s, and only `_list_association_ids` (the genuinely slow, large-portal pagination path) opts into the longer 90s timeout. I verified all other `_request()` call sites (search/get/create/update contact, search/create/update company, batch-read deals/notes, create note) still get the 30s default, and the two new tests (`test_request_defaults_to_the_short_timeout`, `test_association_listing_uses_the_longer_timeout`) assert on the actual `timeout` kwarg reaching `requests.request` — I confirmed by mutation-testing that removing the override makes the association test fail, so this has real regression coverage now. This also resolves the concern raised in the earlier `gemini-code-assist` bot comment about the blanket timeout raise.

One residual note, not a blocker: on a very large portal needing many association pages, each page still gets its own independent 90s allowance with no caller-side deadline anywhere upstream (agent runtime / MCP stdio transport have no timeout wrapper). That's inherent to fixing the reported issue and isn't something this file/PR could reasonably address — flagging only as a known, accepted tradeoff.

**Minor, non-blocking**

See inline comment on the constant's comment block — consider citing the observed latency and clarifying whether this was seen on a single association-listing call or a bulk/looped pull across many contacts, so the next person tuning this has data to reason from.

**Testing**

CI is green (all checks passed). Ran the updated test file locally against the PR head and confirmed both new timeout assertions pass and have teeth (fail when the override is removed).

Approving — the scoping fix directly addresses the main concern from the prior pass, and coverage now backs the value in place.

PR #1205, opened 2026-08-08
feat(kb): let team users create a knowledge base as team-owned

bluefish-08:
## Problem

The creation dialog has no ownership control, so every knowledge base starts Personal. Making one team-owned means creating it, then using the toggle on the list page — which relocates files and vector rows as a background job.

## Change

An ownership choice in the creation dialog, gated on `inTeam` from `useAuth()` — the same signal the existing list-page toggle uses, already available as context.

When Team is picked, the name is claimed once before the first ingest request, through two endpoints, `reserve-team` and `release-team-claim`. They share the `/api/knowledge-bases/*` prefix with the list-page toggle's `promote-team` / `demote-personal`, but they are new: nothing else in this repo calls them. Those are provided by multi-tenant deployments; standalone builds have no team, so `inTeam` is false, the control never renders, and none of those endpoints is touched. A test asserts exactly that.

All three submit paths (file, web, cloud) share one helper, so a multi-file upload claims once rather than per file.

## Rollback

If the ingest fails, the claim is released — otherwise a teammate's later knowledge base under the same name would silently resolve into team storage.

The client does not decide whether releasing is safe; it always asks, and the server refuses with 409 once a collection exists behind the claim. A partially successful ingest looks like failure from here while having already created the collection, so that judgement cannot live in the browser.

A release that genuinely fails (403, 404, a network error) surfaces as a warning, because the leftover claim is not something the user can fix from the UI — the knowledge base does not exist, so it never appears in the list that carries the ownership toggle. `apiRequest` does not reject on a non-2xx response, so the status is checked explicitly; catching alone would have made this failure invisible.

A retry serializes behind any release still in flight, bounded at 10 seconds — `apiRequest` has no timeout, so an unbounded wait would wedge the dialog for good. On timeout the attempt **fails visibly** and the hung release is **written off**, so the wedge costs one failed attempt rather than 10 seconds on every submission thereafter. The write-off leaves a deliberate residue: a retry after it proceeds even though the zombie release may still land and drop the fresh claim before the first write. That window — a release that hangs past 10 seconds and then completes at exactly the wrong moment — is accepted over permanently wedging the dialog; a release answering 409 (data already behind the name) keeps the claim regardless.

Neither path replaces the ingest error itself.

## Collision, before the ingest rather than after

[...]

Issue #1215, opened 2026-08-09
Resolve knowledge-base ownership from the ingest request, not a name claim


bluefish-08:
## Why the claim exists

#1205 claims the collection name before the first ingest request and releases it
if that ingest fails. It has to: none of the three ingest endpoints (`/api/kb/ingest`,
`/api/kb/ingest-web`, `/api/kb/ingest-cloud`, plus the `/jobs` variants) accepts an
ownership parameter. They resolve it implicitly from the name, through
`_effective_knowledge_base_user` -> `resolve_knowledge_base_access`. The client
cannot say "this ingest belongs to the team", so the only lever left is to make the
name already resolve that way.

## The simpler shape

`resolve_knowledge_base_access` (`src/xagent/web/services/knowledge_base_team_scope.py:77`)
already decides ownership atomically, before the storage path is computed and
before any file is written. Give it a `scope` from the ingest request and let it
create-or-resolve in one step, and the protocol goes away:

| Removed | Where it lives today |
|---|---|
| The two claim endpoints | xagent-cloud |
| Claim, release, rollback and the three-state bookkeeping | `knowledge-base-creation-dialog.tsx` |
| Leaked-claim warning copy | same, plus `en.ts` / `zh.ts` |
| 409-vs-404 branching on the reserve path | same |
| Idempotency assumption on a retried POST | by convention, unenforced |
| The claim-emptiness predicate and its guards | xagent-cloud |

It also removes a failure the claim protocol cannot fully close: a claim the server
never confirmed — the write committed but the response was a 5xx or the connection
dropped — has no TTL and no UI to clear it. A scope field never creates that state.

## Not a change to #1205

That PR is several review rounds deep and what it ships is correct and tested.
Redirecting its shape now buys nothing it does not already deliver.

This is written down for the *next* ownership-bearing resource, where the choice
will be between copying a two-phase claim protocol and adding one field to a hook
that already runs in the right place.

## Asks

1. Settle whether `scope` on the ingest request models is the shape we want, before
   this pattern is copied.
2. If yes, the migration is additive: accept `scope`, pass it to the access hook,
   and keep the existing claim endpoints working until the client stops calling them.

PR #1249, opened 2026-08-10
feat: let team users create a knowledge base as team-owned

bluefish-08:
Replaces #1205.

## Problem

The creation dialog has no ownership control: every knowledge base starts personal, and making it team-owned takes a second step on the list page, which physically relocates files and vector rows as a background job.

## Change

An ownership choice (Personal / Team) on step 1, gated on `inTeam` from `useAuth()`. Standalone builds never render it and never touch a team endpoint — a test asserts exactly that.

When Team is picked, the name is claimed once before the first ingest request via `reserve-team` (provided by multi-tenant deployments; a promote over an empty collection, idempotent per creator). A reserve 409 rejects the name on step 1 before anything is created.

## Rollback

If the ingest fails, one best-effort `release-team-claim` follows. A failed release only warns: the claim stays reusable by its creator (reserving the same name again answers 204) and blocks teammates with an explicit server-side 409, so the client keeps no release/retry state machine — the browser cannot guarantee that lifecycle, which is what sank #1205.

## Carried over from #1205's review

The dialog's selection cards are one keyboard-accessible radiogroup (APG pattern) with constant border width, the stepper's fixed bottom margin moved into per-dialog padding, and the unselected radio ring is visible on the light theme.

## Testing

19/19 dialog tests (8 new: standalone gating, default-Personal reserves nothing, exactly one reserve before file/web/cloud ingest, 409 → step-1 error with no ingest, one release on failure, release failure warns without clobbering the ingest error, keyboard radiogroup). `tsc --noEmit` clean; the full frontend suite's single failure predates this branch and reproduces on main.

### DIFF
--- a/frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
+++ b/frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
@@ -5,6 +5,12 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 const apiRequestMock = vi.hoisted(() => vi.fn())
 const toastErrorMock = vi.hoisted(() => vi.fn())
 const toastSuccessMock = vi.hoisted(() => vi.fn())
+const toastWarningMock = vi.hoisted(() => vi.fn())
+const inTeamMock = vi.hoisted(() => ({ value: false }))
+
+vi.mock("@/contexts/auth-context", () => ({
+  useAuth: () => ({ inTeam: inTeamMock.value }),
+}))
 
 vi.mock("@/lib/api-wrapper", () => ({
   apiRequest: apiRequestMock,
@@ -42,7 +48,7 @@ vi.mock("@/components/ui/sonner", () => ({
   toast: {
     error: toastErrorMock,
     success: toastSuccessMock,
-    warning: vi.fn(),
+    warning: toastWarningMock,
   },
 }))
 
@@ -63,6 +69,8 @@ vi.mock("lucide-react", () => {
     ChevronUp: Icon,
     ArrowRight: Icon,
     ArrowLeft: Icon,
+    User: Icon,
+    Users: Icon,
   }
 })
 
@@ -182,6 +190,9 @@ function installApiMocks() {
     if (url === "http://api.local/api/jobs/capabilities") {
       return Promise.resolve(createJsonResponse({ kb_ingest_mode: "celery" }))
     }
+    if (url.endsWith("/reserve-team") || url.endsWith("/release-team-claim")) {
+      return Promise.resolve(createJsonResponse(null, true, 204))
+    }
     if (url === "http://api.local/api/kb/ingest/jobs") {
       return Promise.resolve(
         createJsonResponse(
@@ -261,6 +272,8 @@ describe("KnowledgeBaseCreationDialog collection naming", () => {
     apiRequestMock.mockReset()
     toastErrorMock.mockReset()
     toastSuccessMock.mockReset()
+    toastWarningMock.mockReset()
+    inTeamMock.value = false
     installApiMocks()
   })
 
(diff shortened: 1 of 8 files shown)


frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -104,6 +108,111 @@ function buildWebIngestionErrorResult(
...
+      ? 1
+      : event.key === "ArrowLeft" || event.key === "ArrowUp"
+        ? -1
+        : 0
+    const jump = event.key === "Home" ? 0 : event.key === "End" ? options.length - 1 : -1
+    if (step === 0 && jump < 0) return
+    event.preventDefault()
+    const next = step === 0 ? jump : (index + step + options.length) % options.length
+    onSelect(options[next].value)
+    cards[next].focus()
```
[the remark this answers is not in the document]
bluefish-08: Done in a7db8845.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -783,6 +967,7 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
             <div className="mt-6">
               <Stepper
                 currentStep={currentStep}
+                contentClassName="pt-6"
```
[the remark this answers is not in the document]
bluefish-08: Good catch — in this dialog the step bodies render outside the Stepper, so the padding belonged on the real content container. Removed the contentClassName here and moved pt-6 to the container below in a7db8845. The other two dialogs keep contentClassName because their step bodies do render inside the Stepper.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -794,25 +979,25 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
 
           <div className="flex-1 overflow-y-auto px-6 pb-6">
```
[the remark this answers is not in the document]
bluefish-08: Done in a7db8845.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -494,11 +668,14 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
...
         // chosen name is already taken.
-        { status: failedStatus, adviseRename: true }
+        { status: failureStatus(err, failedStatus), adviseRename: true }
       )
       toast.error(toastContent.title, {
         description: toastContent.description,
       })
+      // Not awaited: the failure is already on screen, and a slow release must
+      // not keep the button reading "Processing".
+      if (teamClaimed.current) void releaseTeamName(collectionName)
```
rogercloud: The release fires whenever `teamClaimed.current` is true, without checking whether part of the ingest already succeeded — a two-file upload where file 1 lands and file 2 fails still calls `releaseTeamName` from the same catch block that reports `onSuccess(successfulCollections)`. The doc comment above documents this as intentional (409 means data exists, so keeping the claim is correct), so the client is trusting a server guard this repo cannot verify. Consider gating on `successfulCollections.length === 0` as defense in depth.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -320,18 +432,77 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
...
+  /** Best-effort, fire-and-forget: a failed ingest gives the name back so a
+   *  teammate's later knowledge base under it is not resolved into team storage.
+   *  409 means data exists behind the claim (keeping it is correct) and 404
+   *  means there was no claim; anything else leaks the claim, which only the
+   *  creator can reuse — reserving the same name again answers 204 — so warn
+   *  and move on rather than retry.
+   *  ponytail: no retry/settle tracking; a leaked claim is recoverable by the
+   *  creator and blocks teammates with an explicit 409 server-side. */
+  const releaseTeamName = async (collection: string) => {
+    try {
```
rogercloud, line 477: `ponytail:` looks like an authoring/dictation artifact left in this doc comment. Harmless at runtime, but it distracts from what is otherwise a clear design-rationale note. Suggest dropping it or replacing with the intended lead-in (e.g. "Rationale:").

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -372,16 +543,19 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
     setIngestionResults([])
     setCompletedUploadCount(0)
 
+    const collectionName = trimmedCollectionName
     const successfulCollections: string[] = []
+    const teamClaimed = { current: false }
```
rogercloud, line 534: This reserve/track/release trio (`teamClaimed` ref + `reserveTeamName` in `try` + conditional `releaseTeamName` in `catch`) is duplicated identically in the web (~702/814) and cloud (~829/936) handlers. The copies are byte-identical today, so no drift yet — but any future change to release behavior has to be replicated by hand three times. Consider a `withTeamClaim(name, fn)` wrapper that owns reserve, tracking, and release for all three paths.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
```diff
@@ -826,38 +1013,69 @@ export function KnowledgeBaseCreationDialog({ open, onOpenChange, onSuccess }: K
                     className="mt-1.5 h-32"
                   />
                 </div>
+                {/* Also rendered once Team is chosen, whatever `inTeam` says
+                    afterwards: losing membership mid-dialog would otherwise take
+                    the control away while every submit still tried to reserve,
+                    leaving no way back to Personal short of reopening. */}
+                {(inTeam || ownership === "team") && (
```
rogercloud, line 1010: This gates on `inTeam`, which starts `false` and is reset to `false` on every `token` change (login, and background refresh roughly every 55 minutes) while `GET /api/teams/my-team` is in flight — see `frontend/src/contexts/auth-context.tsx:79-92`. Usually resolved before the dialog is reachable, but right after a fresh login redirect (or if a refresh races with opening the dialog) the user sees no ownership choice and no hint one is coming. Exposing a `teamLoading` flag and rendering a disabled/skeleton control while it is true would close that window.

frontend/src/components/ui/radio-group.tsx
```diff
@@ -28,7 +28,9 @@ const RadioGroupItem = React.forwardRef<
     <RadioGroupPrimitive.Item
       ref={ref}
       className={cn(
-        "aspect-square h-4 w-4 rounded-full border border-primary text-primary ring-offset-background focus:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50",
+        // border-2, not border: a 1px primary ring is near-invisible on the
+        // light theme, leaving an unselected option looking like a bare label.
+        "aspect-square h-4 w-4 rounded-full border-2 border-primary text-primary ring-offset-background focus:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50",
```
rogercloud: `border` → `border-2` on the shared `RadioGroupItem` is global and unscoped — and this PR's dialog does not use `RadioGroupItem` at all (it uses the new `SelectableCardGroup`). The only real consumer affected is `frontend/src/components/mcp/connect-mcp-dialog.tsx` (5 untouched usages at the default `h-4 w-4`), where 1px → 2px narrows the gap to the 10px indicator dot from ~2px to ~1px in the selected state, uncovered by any test or screenshot. Since #1205 was closed rather than merged, this tweak was never actually landed or verified — suggest dropping it here, or attaching a before/after of the MCP dialog radios.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -611,3 +624,241 @@ describe("KnowledgeBaseCreationDialog collection naming", () => {
...
+    expect(team.getAttribute("tabindex")).toBe("-1")
+
+    fireEvent.keyDown(personal, { key: "ArrowRight" })
+    expect(team.getAttribute("aria-checked")).toBe("true")
+    // The tab stop and focus follow the selection, or the arrow key would
+    // strand the user on a card that is no longer tabbable.
+    expect(team.getAttribute("tabindex")).toBe("0")
+    expect(document.activeElement).toBe(team)
+
+    fireEvent.keyDown(team, { key: " " })
```
rogercloud, line 736: This Space-key assertion is vacuous: `team` is already `aria-checked="true"` from the `ArrowRight` press asserted at line 726, so it passes whether or not the handler's Enter/Space branch does anything. Move focus back to `personal` before the Space keydown so the assertion has teeth. `Enter`, `Home`, and `End` — all handled in `handleKeyDown` — currently have no coverage at all.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
```diff
@@ -611,3 +624,241 @@ describe("KnowledgeBaseCreationDialog collection naming", () => {
     }
   })
 })
+
+const RESERVE_URL = "http://api.local/api/knowledge-bases/team-docs/reserve-team"
```
rogercloud, line 628: The claim/release happy path and the single-failure/409 paths are covered well, but a few branch-backed edges are not: `reserve-team` rejecting with a network error (`reserveTeamName` has no local try/catch), `reserve-team` returning 5xx (the `status < 500` branch that distinguishes a confirmed from an unconfirmed claim), closing the dialog mid-ingest, a hung fire-and-forget `release-team-claim`, and the partial-success-then-release case. The network-error, 5xx, and partial-success cases seem most worth adding since each has real branching behind it.

rogercloud (approved):
## Summary

This PR adds a Personal/Team ownership choice to step 1 of the knowledge-base creation dialog, gated on `inTeam` from `useAuth()`. When Team is selected, the collection name is claimed once via `POST .../reserve-team` before the first ingest request, and on ingest failure a single best-effort `POST .../release-team-claim` follows; this is wired independently into all three ingest handlers (file, web, cloud). It replaces PR #1205 — a much heavier prior attempt with release retry/timeout, a `team-status` collision pre-check, and disown-on-close bookkeeping — which was closed after review found it too complex and risky. The diff is frontend-only across 8 files: the dialog and its test (the main change), two dialogs' padding (`admin-mcp.tsx`, `model-management-dialog.tsx`) compensating for a shared `Stepper` margin change, the shared `radio-group.tsx` / `stepper.tsx` primitives, and en/zh i18n additions.

The simplification relative to #1205 is mostly legitimate: it drops browser-side machinery a browser cannot reliably guarantee anyway (release retry/timeout, disown-on-close bookkeeping), which matches the direction argued in follow-up issue #1215 — that issue proposes eliminating the claim/release protocol entirely in favor of a server-side `scope` field on the ingest request, and is explicitly *not* something this PR is expected to implement. Two design-level reservations survive and are folded into the findings below rather than written up separately: the reserve/release protocol is duplicated three times instead of factored into one wrapper (finding 3), and the safety of releasing after a *partially successful* ingest rests entirely on a server-side 409 contract that cannot be verified from this repo (finding 1). On the second point, credit where due: the author's own doc comment documents the 409 reliance as an intentional choice, and the PR body discloses that `reserve-team` / `release-team-claim` are provided by separate multi-tenant/cloud deployments — a declared architectural boundary, not a hidden gap.

## Findings

All findings are MINOR; none block merge. Ordered most load-bearing first.

### 1. MINOR — release fires even after a partially successful ingest
`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:678` (also web-ingest ~814, cloud-ingest ~936)

The release fires whenever `teamClaimed.current` is true, without consulting whether part of the ingest already succeeded — e.g. a two-file upload where file 1 lands in the team collection and file 2 fails still calls `releaseTeamName` from the same catch block that calls `onSuccess(successfulCollections)` for file 1. The doc comment at lines 486-492 documents this as deliberate ("409 means data exists behind the claim (keeping it is correct)"), so the client fires speculatively and trusts the server to refuse. Suggestion: gate the release on `successfulCollections.length === 0` (and the per-path equivalent for web/cloud) as defense in depth, since this repo cannot verify the server actually enforces the 409 guard.

### 2. MINOR — stray token in the `releaseTeamName` doc comment
`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:491`

The comment reads `* ponytail: no retry/settle tracking; a leaked claim is recoverable by the creator...` — "ponytail:" appears to be an authoring/dictation artifact. Harmless at runtime, but it pollutes an otherwise well-written design-rationale comment. Suggestion: drop the token (or replace it with the intended word, e.g. "Rationale:").

### 3. MINOR — claim/release protocol copy-pasted across three handlers
`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:548-554`/`678` (upload), `:702-707`/`814` (web), `:829-837`/`936` (cloud)

The same three-step protocol — declare a `teamClaimed` ref, call `reserveTeamName` in `try`, conditionally call `releaseTeamName` in `catch` — is duplicated identically in all three ingest handlers. All three copies are currently byte-identical so there is no active drift, but any future change to one path's release logic (finding 1's suggested gate, for instance) must be replicated by hand to the other two. Suggestion: extract a small `withTeamClaim(name, fn)` wrapper that owns reserve, tracking, and conditional release, and call it from all three handlers.

### 4. MINOR — no loading state while `inTeam` resolves
`frontend/src/components/kb/knowledge-base-creation-dialog.tsx:1020`, `frontend/src/contexts/auth-context.tsx:79-92`

`inTeam` starts `false` and is reset to `false` on every `token` change (login, plus background token refresh roughly every 55 minutes) while `GET /api/teams/my-team` is in flight, with no loading indicator. In the common case (already-logged-in user, shell mounted long before the dialog opens) `inTeam` has resolved by the time the dialog is reachable — but the dialog *is* reachable right after a fresh login redirect, or when a refresh races with opening it, and the user then sees no ownership choice and no hint that one might appear. Suggestion: expose a `teamLoading` flag from the auth context and render a disabled/skeleton ownership control while it is true.

### 5. MINOR — Cancel stays enabled mid-ingest, now with a new visible consequence
`frontend/src/components/kb/knowledge-base-creation-dialog.tsx` (~1492 Cancel vs. ~1503 Back / ~1538 Create)

Cancel is the only action-bar button not disabled during an in-flight ingest — Back and Create both check `isUploading || isWebIngesting || isCloudConnecting`, Cancel does not. The gap is pre-existing and unchanged by this PR, but this PR gives it a new user-visible effect: cancelling mid-ingest and having the ingest fail later can surface a `kb.ownership.releaseFailed` toast for a dialog the user already dismissed. The async handlers use closure-captured locals rather than stale React state, so this is a toast/UX-timing nit, not a data-corruption risk. Suggestion: apply the same in-flight disable to Cancel, or suppress the release-failure toast once the dialog has closed.

### 6. MINOR — unscoped `border` → `border-2` on a shared primitive
`frontend/src/components/ui/radio-group.tsx:33`

This changes the shared `RadioGroupItem` globally, so every consumer gets the thicker ring. Notably, the KB creation dialog in this PR does not use `RadioGroupItem` at all (it uses the new `SelectableCardGroup`), so the only real effect is on `frontend/src/components/mcp/connect-mcp-dialog.tsx`, which has five untouched usages (lines 934, 940, 1500, 1507, 1515) at the default `h-4 w-4`. At that size, 1px → 2px shrinks the gap to the 10px indicator dot from ~2px to ~1px — a small but real tightening of the selected state, with no screenshot or test covering it. The PR description frames this as carried over from #1205's review, but #1205 was closed and never merged, so this change has never actually landed or been verified against `connect-mcp-dialog.tsx`. Suggestion: either drop it from this PR (it isn't needed for the feature) or attach a before/after screenshot of the MCP dialog's radio states.

### 7. MINOR (test quality) — vacuous Space-key assertion, and untested key branches
`frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx:732-733`

`team` is already `aria-checked="true"` from the preceding `ArrowRight` press (asserted at line 726), so the Space-keydown assertion at 732-733 passes regardless of whether the handler's Enter/Space branch does anything at all. `Enter`, `Home`, and `End` — all handled in `handleKeyDown` (dialog.tsx:167-184) — have zero coverage. Suggestion: move focus back to `personal` before the Space press so the assertion is meaningful, and add cases for `Enter`, `Home`, and `End`.

### 8. MINOR (test quality) — untested edges of the claim/release protocol
`frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx:628`

The happy path and the single-failure/409 paths are well covered, but these edges are not: (a) `reserve-team` rejecting with a network error rather than resolving non-2xx — relevant because `reserveTeamName` has no local try/catch; (b) `reserve-team` returning 5xx, which is the exact branch (dialog.tsx:468, `status < 500`) separating a confirmed claim from an unconfirmed one; (c) the dialog being closed while an ingest is mid-flight; (d) a slow/hung `release-team-claim` (fired with `void`, fire-and-forget); (e) the partial-success-then-release scenario from finding 1. Suggestion: add cases for (a), (b), and (e) at minimum — those are the ones with behavioral branches behind them.

## Out of scope / dropped

- A test-file mock of `@/components/ui/tabs` looks dead now that the dialog uses `SelectableCardGroup`, but the Tabs → `SelectableCardGroup` migration predates this PR, so it is pre-existing debt this diff does not touch.
- `reserve-team` / `release-team-claim` genuinely have no server-side implementation in this repo, but the PR body explicitly discloses they are provided by multi-tenant deployments — a declared boundary, so it is folded into finding 1 rather than raised on its own.

Lean already.

## Testing

`npx vitest run src/components/kb/knowledge-base-creation-dialog.test.tsx` from `frontend/` — 19/19 passed, matching the PR description; `tsc --noEmit` clean per the PR description. Caveat: the Python test setup was unusable in the review environment due to an unrelated dependency-resolution failure, but this PR touches no Python code, so that is not a coverage gap for this change.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Done in a798b53a: the file path now gates the release on zero successes, and the cloud path got the same gate (a partial batch also puts data behind the claim). The web path stays server-guarded — it is one request, so the client has no per-file success signal beyond the result it already throws on.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Reworded in a798b53a.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Considered a withTeamClaim(name, fn) wrapper, but each handler's catch owns its own toast construction and failedStatus plumbing, so the wrapper would have to take over error rendering for all three paths — more coupling than the three-line trio it removes. After the zero-success gate (thread above) the copies also intentionally differ per path. Happy to revisit if release behavior grows; leaving unresolved for your call.

frontend/src/components/kb/knowledge-base-creation-dialog.tsx
bluefish-08: Agreed the window exists. The control is reactive — it appears as soon as the team lookup lands, even with the dialog already open — and a submit inside the window creates a personal KB, which the list-page toggle can still promote. A skeleton needs auth-context to expose teamLoading, which feels like its own change; proposing to defer to a follow-up. Leaving unresolved for your call.

frontend/src/components/ui/radio-group.tsx
bluefish-08: Dropped in a798b53a — radio-group.tsx is back to main. If the light-theme ring is worth fixing it can go in its own PR with MCP dialog screenshots.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Fixed in a798b53a: Space/Enter now assert against the non-selected card, and Home/End are covered including the focus move.

frontend/src/components/kb/knowledge-base-creation-dialog.test.tsx
bluefish-08: Added in a798b53a: reserve network-error and reserve-5xx (both must skip the ingest and release the ambiguous claim), and partial-success-then-failure (claim kept, onSuccess still fires). Mid-ingest close and a hung release are the disowned-run/settle machinery this PR deliberately leaves out (see the PR body on #1205), so no behavior exists to pin there.

PR #1253, opened 2026-08-10
fix(auth): keep post-commit OAuth side effects out of the failure handler

### LINKED ISSUE #1233: fix(auth): the OAuth callback's blanket exception handler spans the commit boundary
Raised as an informational, non-blocking note in the review of #1223. That PR hardened the one known post-commit side effect; this is the structural gap it left behind.

## Problem

`generic_oauth_callback` (`src/xagent/web/api/auth.py`) wraps its entire body in a single `except Exception` that renders an HTTP 500 as `Authentication Failed` (`auth.py:1524-1531`). That handler spans both sides of `db.commit()` (`auth.py:1487`).

Before the commit, the behaviour is right: the connect genuinely failed, so the user should see a failure.

After the commit, it is wrong: the OAuth token is persisted and the connector is connected, so anything that raises turns a success into a misleading hard failure. This is exactly the bug class #1150 reported and #1223 fixed — for the single side effect that existed at the time.

## Why it is worth fixing structurally

#1223 guarded `_best_effort_ensure_gmail_watches_for_user` at its call site, so today's post-commit region is safe. But the region is still inside the blanket handler, so the next post-commit addition reintroduces the same bug by default rather than by mistake. The safe behaviour should be structural, not something each new side effect has to remember.

## Suggested direction

Split the handler at the commit boundary, so pre-commit failures keep rendering the error page while post-commit work cannot change the response:

- Keep the existing broad handler over everything up to and including `db.commit()`.
- Move the post-commit region into its own block whose failures are logged and swallowed, with the success response built regardless.

Points worth deciding in review:

- Where exactly the boundary sits. `db.commit()` at `auth.py:1487` is the obvious line, but the response-building code after it (`json.dumps`, `urlparse`, the `HTMLResponse`) can also raise — a failure there should probably still surface, since there is no response to return.
- Whether the same split is needed in the other OAuth callback paths in this module, or whether `generic_oauth_callback` is the only one that performs post-commit work.
- Whether the swallowed post-commit failures need any user-visible signal, given the connector is connected but a side effect did not complete. #1150 decided a log entry was acceptable for the Gmail case; that decision should be revisited if the region grows.

Related: #1150, #1223.

codeacme17:
Closes #1233.

## Problem

`generic_oauth_callback` wraps its entire body in a single `except Exception` that renders an HTTP 500 as `Authentication Failed`. That handler spans `db.commit()`, so anything raised after the OAuth token is persisted turns a successful connect into a misleading hard failure — the bug #1150 reproduced on staging.

#1223 guarded `_best_effort_ensure_gmail_watches_for_user` at its call site, so today's post-commit region is safe. The region itself is still inside the blanket handler, though, so the next post-commit addition reintroduces the same bug by default rather than by mistake.

## Changes

The post-commit region moves into `_run_post_commit_oauth_side_effects`, which logs and swallows its own failures. The call site becomes one unconditional call, so new post-commit work inherits the property instead of having to remember it:

```python
db.commit()
_run_post_commit_oauth_side_effects(
    db, user_id=int(user_id), connector_key=(app_id or provider)
)
```

The inner guard on Gmail provisioning is kept as well. The redundancy is deliberate, for the same reason #1223 gave: neither layer should be load-bearing on its own.

## The three points the issue left open

**Where the boundary sits.** At the side effects, not at `db.commit()`. The issue's own reservation is right: response construction (`json.dumps`, `urlparse`, `HTMLResponse`) also runs post-commit, but a failure there leaves no response to return, so it must keep reaching the outer handler. A regression test pins that.

**Whether other callbacks in the module need the same split.** No. `generic_oauth_callback` is the only implementation — `oauth_callback` just delegates. `google_oidc_callback` in `oidc_google.py` does run `_create_exchange_code` after its commit, but the failure semantics differ: without an exchange code there is no way to complete the login, so the error redirect is correct, and a retry succeeds through the existing-identity branch of `_get_or_create_google_user`. It is not the #1150 bug class.

**Whether swallowed failures need a user-visible signal.** Keeping #1150's log-only decision, with its precondition now written down: it holds while every side effect in the region has a recovery path — Gmail watches are re-provisioned by `scan_due_gmail_watch_renewals`, which outer-joins `GmailWatchState` and selects rows where `GmailWatchState.id IS NULL`. A side effect that a swallowed failure would strand needs a signal instead.

## One property worth flagging

A single guard over the whole region couples its entries: a raiser aborts the ones after it, which per-call guards did not do. That is the substantive reason to keep the inner guards, and the docstring says so — anything added there that must run even when an earlier entry fails needs its own guard, exactly as Gmail provisioning has.

## Testing

Two regression tests in `tests/web/test_meta_oauth.py`:

- a raising post-commit side effect leaves the callback at 200. This patches the guarded *wrapper*, not the service beneath it, which is what distinguishes a regional guard from the per-call one #1223 added — the existing tests all raise from inside Gmail provisioning and so cannot tell them apart. Verified to fail without the fix as `assert 500 == 200`, rendering `Authentication Failed`.
- a success page that cannot be rendered still surfaces as 500, pinning the boundary against being widened later.

`tests/web/test_meta_oauth.py`, `test_slack_oauth.py`, `test_zoom_oauth.py`, `test_oidc_google_auth.py`, `test_auth_api.py`, `services/test_gmail_provisioning.py`, `api/test_gmail_auto_triggers.py`: 181 passed, 4 skipped (the skips are the PostgreSQL-only cases). `ruff check`, `ruff format --check` and `mypy src/xagent/web/api/auth.py` are clean.

The wider `pytest tests/web` leg was also run locally: 11 failures, all reproduced identically on a clean tree at the merge-base. 5 are SVG rasterization tests that need `libcairo`, which is absent on this machine; the other 6 are the sandbox reconcile e2e suite.

### DIFF
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -75,13 +75,19 @@ def _best_effort_ensure_gmail_watches_for_user(db: Session, *, user_id: int) -> 
 
 
 def _run_post_commit_oauth_side_effects(
-    db: Session, *, user_id: int, connector_key: str
+    db: Session, *, user_id: Any, connector_key: str
 ) -> None:
     """Run the OAuth callback's post-commit work; never raise.
 
     ``connector_key`` is the value persisted as ``UserOAuth.provider``: the app
     id when the connect is app-scoped (``"gmail"``), otherwise the provider
     name (``"google"``).
+
+    ``user_id`` is the raw ``oauth_state`` claim rather than an ``int`` so that
+    its coercion happens inside the guard below. Coercing it into the argument
+    would put that conversion post-commit but outside the guard, which is the
+    one thing this helper exists to prevent, and would lose the offending value
+    to a 500 instead of logging it.
 
     Everything here runs once ``db.commit()`` has persisted the OAuth token, so
     the connect has already succeeded as far as the user is concerned. The
@@ -112,7 +118,7 @@ def _run_post_commit_oauth_side_effects(
     """
     try:
         if connector_key == "gmail":
-            _best_effort_ensure_gmail_watches_for_user(db, user_id=user_id)
+            _best_effort_ensure_gmail_watches_for_user(db, user_id=int(user_id))
     except Exception:
         logger.warning(
             "Post-commit OAuth side effects failed for user %s on connector %s; "
@@ -1538,7 +1544,7 @@ def generic_oauth_callback(
             # change what this callback returns. Add new post-commit work
             # there, not here.
             _run_post_commit_oauth_side_effects(
-                db, user_id=int(user_id), connector_key=(app_id or provider)
+                db, user_id=user_id, connector_key=(app_id or provider)
             )
 
         import json
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -74,6 +74,55 @@ def _best_effort_ensure_gmail_watches_for_user(db: Session, *, user_id: int) -> 
         )
 
 
+def _run_post_commit_oauth_side_effects(
+    db: Session, *, user_id: int, connector_key: str
+) -> None:
+    """Run the OAuth callback's post-commit work; never raise.
+
+    ``connector_key`` is the value persisted as ``UserOAuth.provider``: the app
+    id when the connect is app-scoped (``"gmail"``), otherwise the provider
+    name (``"google"``).
+
+    Everything here runs once ``db.commit()`` has persisted the OAuth token, so
+    the connect has already succeeded as far as the user is concerned. The
+    callback's outer ``except Exception`` would render anything raised here as
+    an ``Authentication Failed`` 500, reporting a failure for a connector that
+    is in fact connected. That is the bug #1150 reproduced on staging.
+
+    Guarding the region rather than each individual call is what makes that
+    property structural: a side effect added here inherits it, instead of
+    reintroducing #1150 whenever someone forgets to wrap their own call. The
+    inner guards each side effect carries are kept as well, deliberately, so
+    that neither layer is load-bearing on its own.
+
+    One guard over the whole region does couple the side effects to each other:
+    a raiser aborts the ones after it. That is why the inner guards matter and
+    are worth keeping per side effect. Anything added here that must run even
+    when an earlier entry fails needs its own guard, exactly as Gmail
+    provisioning has.
+
+    Response construction stays outside this region on purpose. It also runs
+    after the commit, but a failure there leaves no response to return, so it
+    must keep reaching the outer handler rather than being swallowed here.
+
+    Failures are logged only. That is acceptable while every side effect in
+    this region has its own recovery path: Gmail watches are re-provisioned by
+    ``scan_due_gmail_watch_renewals``. A side effect that a swallowed failure
+    would strand needs a user-visible signal instead.
+    """
+    try:
+        if connector_key == "gmail":
+            _best_effort_ensure_gmail_watches_for_user(db, user_id=user_id)
+    except Exception:
+        logger.warning(
+            "Post-commit OAuth side effects failed for user %s on connector %s; "
+            "the connect itself succeeded",
+            user_id,
+            connector_key,
+            exc_info=True,
+        )
+
+
 def _oauth_env_name(provider: str, suffix: str) -> str:
     return f"{provider.upper()}_{suffix}"
 
@@ -1485,8 +1534,12 @@ def generic_oauth_callback(
                         )
 
             db.commit()
-            if (app_id or provider) == "gmail":
-                _best_effort_ensure_gmail_watches_for_user(db, user_id=int(user_id))
+            # Everything past the commit belongs in the helper, which cannot
+            # change what this callback returns. Add new post-commit work
+            # there, not here.
+            _run_post_commit_oauth_side_effects(
+                db, user_id=int(user_id), connector_key=(app_id or provider)
+            )
 
         import json
         from urllib.parse import urlparse
--- a/tests/web/test_meta_oauth.py
+++ b/tests/web/test_meta_oauth.py
@@ -312,6 +312,142 @@ def test_gmail_callback_survives_a_raising_account_lookup_during_provisioning(
     assert "Failed to resolve Gmail accounts for user" in caplog.text
 
 
+def test_oauth_callback_survives_a_raising_post_commit_side_effect(
+    db_session, monkeypatch, caplog
+):
+    """The post-commit region is guarded as a region, not per side effect.
+
+    The tests above all raise from inside Gmail provisioning, which carries its
+    own guard, so they cannot tell a per-call guard apart from a regional one.
+    Here the guarded wrapper itself raises, which is what a newly added
+    post-commit side effect would do before anyone remembers to guard it.
+    """
+    db, user = db_session
+    state = create_access_token(
+        data={
+            "type": "oauth_state",
+            "user_id": user.id,
+            "provider": "google",
+            "app_id": "gmail",
+        },
+        expires_delta=timedelta(minutes=10),
+    )
+    request = SimpleNamespace(query_params={"code": "gmail-code", "state": state})
+    monkeypatch.setattr(
+        auth_api.requests,
+        "post",
+        Mock(
+            return_value=MockResponse(
+                {
+                    "access_token": "gmail-token",
+                    "refresh_token": "gmail-refresh",
+                    "token_type": "Bearer",
+                    "expires_in": 3600,
+                    "scope": "https://www.googleapis.com/auth/gmail.modify",
+                }
+            )
+        ),
+    )
+    monkeypatch.setattr(
+        auth_api.requests,
+        "get",
+        Mock(
+            return_value=MockResponse(
+                {"sub": "google-user-1", "email": "alice@gmail.com"}
+            )
+        ),
+    )
+
+    def raising_side_effect(_db, *, user_id: int):
+        raise RuntimeError("an unguarded post-commit side effect blew up")
+
+    monkeypatch.setattr(
+        auth_api, "_best_effort_ensure_gmail_watches_for_user", raising_side_effect
+    )
+
+    caplog.set_level(logging.WARNING, logger=auth_api.__name__)
+
+    response = generic_oauth_callback("google", request, db, _google_provider())
+
+    assert response.status_code == 200
+    assert "Authentication Failed" not in response.body.decode()
+    oauth_account = (
+        db.query(UserOAuth)
+        .filter(UserOAuth.user_id == user.id, UserOAuth.provider == "gmail")
+        .one()
+    )
+    assert oauth_account.email == "alice@gmail.com"
+    assert "Post-commit OAuth side effects failed" in caplog.text
+
+
+def test_oauth_callback_still_fails_when_the_success_page_cannot_be_rendered(
+    db_session, monkeypatch
+):
+    """The swallow covers post-commit side effects only, not response building.
+
+    Rendering runs after the commit too, but a failure there leaves nothing to
+    return, so it must keep reaching the outer handler instead of being
+    swallowed into a response that does not exist.
+    """
+    db, user = db_session
+    state = create_access_token(
+        data={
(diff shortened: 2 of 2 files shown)


src/xagent/web/api/auth.py
```diff
@@ -1485,8 +1534,12 @@ def generic_oauth_callback(
                         )
 
             db.commit()
-            if (app_id or provider) == "gmail":
-                _best_effort_ensure_gmail_watches_for_user(db, user_id=int(user_id))
+            # Everything past the commit belongs in the helper, which cannot
+            # change what this callback returns. Add new post-commit work
+            # there, not here.
+            _run_post_commit_oauth_side_effects(
+                db, user_id=int(user_id), connector_key=(app_id or provider)
```
rogercloud: **Nitpick (non-blocking):** `int(user_id)` is evaluated as a call-site argument, so it runs before control enters `_run_post_commit_oauth_side_effects` and is therefore outside that function's try/except. If it raised, the exception would propagate to the outer `except Exception` and render exactly the misleading "Authentication Failed" 500 this PR removes for post-commit work.

Not reachable today: `user_id` comes from a signature-verified, self-issued JWT claim originally set from `User.id` (an `Integer` PK), and python-jose round-trips int claims as `int`. So this is a structural loose end, not a live bug.

Consider passing `user_id=user_id` untyped and doing the `int()` conversion inside the helper, within its existing guard — free, strictly safer, and it makes the "everything past the commit belongs in the helper" invariant literally true.

tests/web/test_meta_oauth.py
```diff
@@ -312,6 +312,142 @@ def provision_with_failing_account_lookup(provision_db, **kwargs):
...
+
+    assert response.status_code == 200
+    assert "Authentication Failed" not in response.body.decode()
+    oauth_account = (
+        db.query(UserOAuth)
+        .filter(UserOAuth.user_id == user.id, UserOAuth.provider == "gmail")
+        .one()
+    )
+    assert oauth_account.email == "alice@gmail.com"
+    assert "Post-commit OAuth side effects failed" in caplog.text
```
rogercloud, line 380: **Low-medium, test coverage gap (pre-existing, not introduced here):** nothing in this file — or in `test_zoom_oauth.py` / `test_slack_oauth.py` — asserts that a non-Gmail connector callback does not invoke Gmail watch provisioning. Verified by mutation: removing the `if connector_key == "gmail":` gate entirely, so provisioning runs for every connector, leaves all 26 tests across these three files passing.

`git show` confirms the gate is unchanged by this PR, so the gap predates the diff. It does get more consequential here, though: the gate now sits under a handler that logs and swallows, so a future regression of it would be silently absorbed instead of raising. Worth a follow-up test asserting `assert_not_called()` on Gmail provisioning for a non-gmail connect. Not a blocker.

rogercloud (approved):
## Summary

`generic_oauth_callback` wraps its entire body in a single `except Exception` that renders a 500 "Authentication Failed" page. Because `db.commit()` sits inside that region, any failure *after* the OAuth token is already persisted turned a successful connect into a misleading hard failure (bug class #1150). PR #1223 patched this per-call-site for the one side effect that existed then (Gmail watch provisioning); this PR (closing #1233) makes the fix structural by extracting `_run_post_commit_oauth_side_effects`, a helper with its own try/except that logs and swallows, invoked once unconditionally right after the commit. The pre-existing inner guard on Gmail provisioning is deliberately retained as defense in depth, and two regression tests pin both sides of the new boundary.

## Design verdict: SOUND

The boundary is in the right place, verified by tracing the code rather than taking the PR's framing at face value:

- The helper is called *inside* the outer `try`, so post-commit work is covered twice (its own handler, plus the outer handler as a backstop) and cannot leak a 500.
- Response-building for the success page (json / urlparse / `HTMLResponse`) remains *outside* the guarded region and still surfaces as a real 500 on failure — verified by mutation testing.
- `oauth_callback` is a pure delegate to `generic_oauth_callback`, so no parallel split is needed there. `google_oidc_callback` in `oidc_google.py` also commits before further work, but its post-commit step (`_create_exchange_code`) must surface on failure — no exchange code means no way to complete login — and its committed side effect (a `User` row) is idempotent on retry. Different failure semantics; correctly out of scope.

Three things considered as possible objections and dropped after verification:

- **"One regional guard couples multiple future side effects"** (the tradeoff the docstring itself flags). Real, but YAGNI at one entry today, and the docstring already documents both the tradeoff and the convention to follow, pointing at the existing per-call guard as the pattern to copy.
- **Extraction vs. inlining the guard at the call site.** Inlining looks simpler, but it is precisely the pattern #1223 already shipped — and it was insufficient, which is why #1233 was filed. The named helper changes the default safety behavior for the next contributor (new code lands inside an already-guarded function) instead of relying on them remembering a guard.
- **Docstring length (~33 lines).** It documents three independent, non-obvious, load-bearing decisions (boundary placement, single-guard tradeoff, the precondition under which log-and-swallow is safe) and matches local convention (the neighboring `_best_effort_ensure_gmail_watches_for_user` is similarly rationale-heavy).

Also checked whether the new regression test duplicates the existing per-call-guard tests: it does not. If the new outer guard were removed, existing tests would still pass (they never reach it) while `test_oauth_callback_survives_a_raising_post_commit_side_effect` would fail immediately with an unguarded 500.

## Findings

### 1. `src/xagent/web/api/auth.py:1541` — Nitpick (non-blocking)

`int(user_id)` is evaluated as a call-site argument, i.e. before control enters `_run_post_commit_oauth_side_effects`, and is therefore outside its try/except. If that coercion ever raised, the exception would propagate to the outer `except Exception` and render exactly the misleading "Authentication Failed" 500 this PR exists to eliminate for post-commit work.

Not currently reachable: `user_id` comes from a signature-verified, self-issued JWT claim originally set from `User.id` (an `Integer` PK), and python-jose round-trips int claims as `int`, not `str`. So this is a structural loose end rather than a live bug.

Suggestion: pass `user_id=user_id` untyped and do the `int()` conversion inside the helper, within its existing guard. Free change, strictly safer.

### 2. `tests/web/test_meta_oauth.py`, `tests/web/test_zoom_oauth.py`, `tests/web/test_slack_oauth.py` — Low-medium, test coverage gap (pre-existing)

No test asserts that a non-Gmail connector callback (meta/facebook, zoom, slack, or a bare google connect without `app_id="gmail"`) does not invoke Gmail watch provisioning. Confirmed by mutation: deleting the `if connector_key == "gmail":` gate entirely (Gmail provisioning runs for every connector) leaves all 26 tests across these three files green.

`git show` confirms the gating condition is unchanged by this PR, so the gap predates the diff and is not a regression introduced here. It does become more consequential, though: the gate now sits under a handler that logs and swallows, so a future regression of the gate would be silently absorbed rather than raising loudly.

Suggestion: follow-up test adding `assert_not_called()` on Gmail provisioning for at least one non-gmail connect path. Not a blocker for this PR.

## Testing

Verified locally during review: `pytest tests/web/test_meta_oauth.py` → 18 passed; combined with the zoom and slack suites → 26 passed. `ruff check` and `ruff format --check` clean. `mypy src/xagent/web/api/auth.py` reports zero errors in the changed areas. Two mutation experiments were run to validate the claims above (removing the outer guard's `int()` boundary, and removing the connector gate), and the worktree was restored clean afterward.

## Recommendation

Approve. The fix is correctly scoped, the boundary placement is verified rather than asserted, and the design choices that look like over-engineering each have a concrete history behind them. Both findings below are optional follow-ups.

codeacme17:
Thanks for the review, and particularly for verifying the boundary by mutation rather than taking the docstring's word for it.

Both findings addressed, one in this PR and one deferred.

**Finding 1 — implemented in e5e10e29.** Taken, and the reachability turns out to be slightly wider than the nitpick framing. Your producer analysis is right: `generic_oauth_login` is the only minter of `oauth_state` tokens, it uses `User.id` (an `Integer` PK), and python-jose round-trips int claims as `int`. But the "signature-verified" premise has a hole: `JWT_SECRET_KEY` falls back to `"your-secret-key-change-in-production"` (`src/xagent/web/auth_config.py:3`), and `warn_if_example_jwt_config` only logs a warning — nothing refuses to boot. On a deployment left on that default the state token is forgeable, and a `user_id` claim of `"7abc"` flows through the `UserOAuth` delete/insert and `db.commit()` on SQLite before `int()` raises. That is a misconfigured deployment with much larger problems, so it does not change the severity much, but it does mean the loose end is not purely structural.

The parameter widens to `Any` to say what it now receives. That costs nothing under this repo's mypy config — `payload.get("user_id")` was already `Any`, so the `int` annotation was not buying a check. As you noted, the failure now also logs the offending value instead of losing it to a 500. Regression test added: a `"7abc"` claim leaves the callback at 200 with provisioning never invoked, verified to fail as `assert 500 == 200` without the change.

**Finding 2 — filed as #1265, not fixed here.** I reproduced your mutation independently: neutralizing the gate leaves 26/26 green across the three files.

I did want to push back on one part of the rationale, and wrote the issue to say so. The escalation argument does not hold up: the gate's body has entered `_best_effort_ensure_gmail_watches_for_user` since #1223, which swallows, and the service beneath it documents "Failures are recorded on the watch state, never raised". This PR adds a *second* swallow over the same code — detectability is essentially unchanged, not degraded. And the blast radius of a regressed gate is smaller than it looks: `best_effort_provision_gmail_watches_for_user` filters `UserOAuth.provider == "gmail"` and then only touches accounts backing an enabled Gmail trigger, so the worst case is wasted work, not incorrect state.

The test is still worth having — the gate is currently unpinned in the negative direction — just on its own merits rather than as a consequence of this diff. #1265 records both the mutation evidence and the two escalations that do not survive checking, so whoever picks it up is not working from an inflated premise.

src/xagent/web/api/auth.py
codeacme17: Taken — fixed in e5e10e29. The call site now passes the raw claim and the coercion happens inside the guard, so the comment right above it is literally true.

One correction to the reachability note. The producer analysis is right, but "signature-verified" assumes a configured secret: `JWT_SECRET_KEY` falls back to `"your-secret-key-change-in-production"` (`src/xagent/web/auth_config.py:3`) and `warn_if_example_jwt_config` only logs a warning rather than refusing to boot. On a deployment left on that default the state token is forgeable, and a `user_id` of `"7abc"` reaches `int()` after the commit has already landed. Still a badly misconfigured deployment, so not a severity change — but the loose end is not purely structural.

The parameter widens to `Any` to describe what it actually receives. No typing loss: `payload.get("user_id")` was already `Any`, so the `int` annotation was not checking anything, and mypy is clean either way. The `except` branch now logs the offending value instead of it being lost to a 500.

Regression test `test_oauth_callback_survives_a_non_integer_user_id_claim` pins it, verified to fail as `assert 500 == 200` without the change.

tests/web/test_meta_oauth.py
codeacme17: Agreed the gap is real, and I reproduced the mutation independently: neutralizing the gate leaves 26/26 green across the three files. Filed as #1265 rather than fixed here, since #1233 is about the commit boundary and this diff relocated the gate without changing its condition.

Two parts of the rationale I checked and could not confirm, recorded in the issue so it is not picked up from an inflated premise:

**The escalation does not hold.** The gate's body has entered `_best_effort_ensure_gmail_watches_for_user` since #1223, which swallows, and `best_effort_provision_gmail_watches_for_user` documents "Failures are recorded on the watch state, never raised". A gate regression was already silently absorbed before this PR; the new handler is a second swallow over the same code, so detectability is unchanged rather than degraded.

**The blast radius is small.** `best_effort_provision_gmail_watches_for_user` filters `UserOAuth.provider == "gmail"` and then only touches accounts in `_referenced_gmail_oauth_account_ids`, i.e. backing an enabled Gmail trigger. A Slack/Zoom/Facebook connect by a user without Gmail is one wasted query; for a user who also has Gmail with a live trigger it re-runs idempotent provisioning — wasted latency, not corruption.

So the test earns its place because the gate is unpinned in the negative direction, not because a regression would be dangerous. #1265 also notes the choice worth making in review: a bare `google` connect with no `app_id` exercises the `(app_id or provider)` fallback branch, which a careless refactor is likelier to break than the app-scoped path.

Issue #1149, opened 2026-08-05
fix(monitor): monitoring stats silently return zero on PostgreSQL (invalid ~? operator)


codeacme17:
Found while reviewing #1147, which fixed a sibling failure of the same class (a query that is valid on SQLite and invalid on PostgreSQL, with no PostgreSQL test to catch it). Not part of that diff.

## Symptom

On PostgreSQL deployments, three monitoring endpoints lose their statistics:

| Endpoint | Value lost | What the caller sees |
| --- | --- | --- |
| `GET /monitor/stats` | active model count | HTTP 500 (see below) |
| `GET /monitor/popular-tools` | tool usage stats | HTTP 200 and `[]` (`src/xagent/web/api/monitor.py:442`) |
| `GET /monitor/model-stats` | per-model call stats | HTTP 200 and `[]` (`src/xagent/web/api/monitor.py:507`) |

Each site wraps its query in `try/except Exception` that logs and substitutes an empty value. For the two list endpoints that substitution is the end of it -- neither touches the database again -- so nothing crashes and nothing surfaces to the caller. The only evidence is a `logger.error` line.

**Correction (the first row was wrong when this issue was filed).** `/monitor/stats` does not return 200 with `active_models = 0`. The local `except` sets that fallback (`src/xagent/web/api/monitor.py:339`), but the rejected statement leaves the PostgreSQL transaction aborted, and the handler goes on to query `llm_call_end` rows on the same session (`src/xagent/web/api/monitor.py:344`). That second query raises outside any local handler, so the outer `except` at `src/xagent/web/api/monitor.py:398` turns the request into an HTTP 500. Anything reproducing this should expect a 500 there, not a zero.

## Root cause

`safe_get_json_field` (`src/xagent/web/api/monitor.py:63-73`) takes a PostgreSQL-specific branch that filters control characters before extracting the field:

```python
valid_data = expression.case(
    (
        column.op("~?")(r"[-^N-^_]"),
        None,
    ),  # Filter control characters
    else_=column,
)
return valid_data.op("->>")(field_name)
```

`~?` is not a PostgreSQL operator. Verified against PostgreSQL 16.14:

```
=> SELECT ('{}'::json) ~? 'x';
ERROR:  operator does not exist: json ~? unknown
HINT:  No operator matches the given name and argument types.

=> SELECT ('{}'::jsonb) ~? 'x';
ERROR:  operator does not exist: jsonb ~? unknown

[...]

PR #1254, opened 2026-08-10
fix(monitor): divide model calls by the fleet total, not by themselves

### LINKED ISSUE #1245: fix(monitor): /monitor/model-stats reports usage_rate 100% for every model
Found while working on #1149 / #1222, which touched the same handler but deliberately did not fix this. Not part of that diff.

## Symptom

`GET /monitor/model-stats` reports `usage_rate: 100.0` for every model, no matter how the calls are actually distributed. Two models with 3 and 1 calls both come back as 100%.

## Root cause

`get_model_stats` computes the grand total, then shadows the same name with the loop variable:

```python
# src/xagent/web/api/monitor.py:575-580
total_calls = sum(stat.total_calls for stat in model_stats)

result = []
for model_name, total_calls in model_stats:      # <- rebinds total_calls
    if model_name and total_calls > 0:
        usage_rate = (total_calls / total_calls * 100) if total_calls > 0 else 0
```

By the time `usage_rate` is computed, `total_calls` is the per-model count on both sides of the division, so the expression is `n / n * 100` — constant 100.0 whenever `n > 0`, and the `else 0` branch is unreachable because the enclosing `if` already required `total_calls > 0`.

The aggregate on line 575 is computed and then never read.

## Suggested fix

Rename the aggregate so the loop cannot shadow it, and divide per-model calls by it:

```python
total_calls_all_models = sum(stat.total_calls for stat in model_stats)

for model_name, model_calls in model_stats:
    if model_name and model_calls > 0:
        usage_rate = (
            model_calls / total_calls_all_models * 100
            if total_calls_all_models > 0
            else 0
        )
```

## Tests

`tests/web/api/test_monitor_postgresql.py` already exercises this endpoint end-to-end but asserts only `total_tasks`, so it passes today and would keep passing after the fix. A regression test should assert the rate itself — two models with 3 and 1 calls reporting 75.0 and 25.0 rather than 100.0 each. The SQLite path is enough for this; nothing here is dialect-specific.

codeacme17:
Closes #1245

## Problem

`GET /monitor/model-stats` reported `usage_rate: 100.0` for every model regardless of how the calls were distributed. Two models with 3 and 1 calls both came back as 100%.

`get_model_stats` summed the grand total into `total_calls`, then the loop header rebound the same name to the per-model count:

```python
total_calls = sum(stat.total_calls for stat in model_stats)

for model_name, total_calls in model_stats:      # rebinds total_calls
    if model_name and total_calls > 0:
        usage_rate = (total_calls / total_calls * 100) if total_calls > 0 else 0
```

By the time the rate was computed both sides of the division were the same value — `n / n * 100`, a constant 100 for every row with any traffic. The aggregate was never read, and the `else 0` arm was unreachable because the enclosing `if` already required a positive count.

The frontend renders the field straight through as a percentage (`frontend/src/components/pages/monitoring.tsx:191`), so the call distribution the monitoring page exists to show was replaced by a column of identical numbers.

## Fix

Rename the aggregate to `all_model_calls` so a loop variable cannot shadow it, name the loop variable `model_calls`, and divide one by the other. The zero-guard is dropped rather than rewritten: the denominator sums the same rows the loop walks, each already established to be positive.

Beyond the report, the rows are now selected *before* the sum rather than skipped inside the loop. The query rejects a NULL model name but not an empty one, so an empty name reaches Python and is dropped there. That was free while the aggregate went unread; now that it is the denominator, counting calls the response never reports would deflate every rate it does report — one nameless call beside two real ones drops a sole model from 100% to 66.7%. Confirmed by reverting just that filter: the new test goes from 100.0 to 33.3.

## Tests

New `tests/web/api/test_monitor_model_stats.py` on the SQLite path — the shadowing was plain Python downstream of the query, with nothing dialect-specific about it. The sibling PostgreSQL suite covers the dialect-dependent JSON extraction feeding this handler and asserts only per-model call counts, so it passed before this fix and keeps passing after.

- Two models splitting traffic 3-to-1 report 75.0 and 25.0. Verified red before the fix (both 100.0).
- A single model holding all the traffic still reports 100.0, so a fix that merely stopped returning 100.0 everywhere would not pass.
- A model sharing a task with nameless calls still reports 100.0.

`tests/web/api` passes in full; `ruff check`/`ruff format` clean; `mypy` reports nothing new for `monitor.py`.

## Out of scope

The response key `total_tasks` holds LLM call counts, not task counts. Renaming it is a wire-contract change touching the frontend type, its usage, and the i18n strings, so it is left for a follow-up rather than folded into this fix.

### DIFF
--- a/src/xagent/web/api/monitor.py
+++ b/src/xagent/web/api/monitor.py
@@ -571,25 +571,39 @@ async def get_model_stats(
             logger.error(f"Failed to query model stats: {e}")
             model_stats = []
 
-        # Get total call count for calculating usage rate
-        total_calls = sum(stat.total_calls for stat in model_stats)
+        # Rows the response will carry. Filtered here rather than inside the
+        # loop so a dropped row stays out of the denominator too: the query
+        # rejects a NULL model name but not an empty one, and counting calls
+        # that are never reported would deflate every rate that is.
+        counted_models = [
+            (model_name, model_calls)
+            for model_name, model_calls in model_stats
+            if model_name and model_calls > 0
+        ]
+
+        # Denominator for each model's share of the traffic. Deliberately not
+        # named ``total_calls``: the loop below used to bind that same name to
+        # the per-model count, so the rate divided a model's calls by itself
+        # and every model reported 100% (#1245).
+        all_model_calls = sum(model_calls for _, model_calls in counted_models)
 
         result = []
-        for model_name, total_calls in model_stats:
-            if model_name and total_calls > 0:
-                usage_rate = (total_calls / total_calls * 100) if total_calls > 0 else 0
+        for model_name, model_calls in counted_models:
+            # No zero-guard needed: the sum runs over these same rows, each of
+            # which the comprehension above established is positive.
+            usage_rate = model_calls / all_model_calls * 100
 
-                result.append(
-                    {
-                        "name": model_name,
-                        "status": "running",
-                        "usage_rate": round(usage_rate, 1),
-                        "success_rate": None,  # Simplified, do not calculate success rate
-                        "total_tasks": total_calls,
-                        "successful_tasks": None,
-                        "failed_tasks": None,
-                    }
-                )
+            result.append(
+                {
+                    "name": model_name,
+                    "status": "running",
+                    "usage_rate": round(usage_rate, 1),
+                    "success_rate": None,  # Simplified, do not calculate success rate
+                    "total_tasks": model_calls,
+                    "successful_tasks": None,
+                    "failed_tasks": None,
+                }
+            )
 
         # If no real data, return empty list
         if not result:
--- a/src/xagent/web/api/monitor.py
+++ b/src/xagent/web/api/monitor.py
@@ -581,10 +581,10 @@ async def get_model_stats(
             if model_name and model_calls > 0
         ]
 
-        # Denominator for each model's share of the traffic. Deliberately not
-        # named ``total_calls``: the loop below used to bind that same name to
-        # the per-model count, so the rate divided a model's calls by itself
-        # and every model reported 100% (#1245).
+        # Denominator for each model's share of the traffic. Must stay distinct
+        # from the loop's per-model variable: while the two shared a name the
+        # rate divided a model's calls by itself and every model reported 100%
+        # (#1245).
         all_model_calls = sum(model_calls for _, model_calls in counted_models)
 
         result = []
@@ -604,10 +604,6 @@ async def get_model_stats(
                     "failed_tasks": None,
                 }
             )
-
-        # If no real data, return empty list
-        if not result:
-            return []
 
         return result
     except Exception as e:
--- a/tests/web/api/test_monitor_model_stats.py
+++ b/tests/web/api/test_monitor_model_stats.py
@@ -1,0 +1,133 @@
+"""``/monitor/model-stats`` usage-rate arithmetic.
+
+``get_model_stats`` reported ``usage_rate: 100.0`` for every model no matter
+how the calls were actually distributed (#1245). The grand total was summed
+into ``total_calls`` and then shadowed by the loop variable of the same name,
+so the rate divided a model's own call count by itself -- ``n / n * 100``, a
+constant 100 for every row with any traffic at all. The dashboard renders the
+field directly as a percentage, so the distribution it exists to show was
+replaced by a column of identical numbers.
+
+Nothing here is dialect-specific: the shadowing is plain Python downstream of
+the query, so the SQLite path exercises it. The sibling PostgreSQL suite
+covers the dialect-dependent JSON extraction feeding it and asserts only
+per-model call counts.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+
+import pytest
+from sqlalchemy.orm import Session
+
+from xagent.web.api.monitor import get_model_stats
+from xagent.web.models.task import Task, TraceEvent
+from xagent.web.models.user import User
+
+from .conftest import _direct_db_session
+
+pytestmark = pytest.mark.usefixtures("_test_db")
+
+
+def _seed_model_calls(db: Session, model_names: list[str], *, owner: str) -> User:
+    """One admin with one ``llm_call_start`` event per name in ``model_names``.
+
+    Repeat a name to give that model more calls. ``owner`` keeps usernames
+    distinct so a test never depends on another test's rows surviving.
+    """
+    admin = User(username=owner, password_hash="hash", is_admin=True)
+    db.add(admin)
+    db.commit()
+    db.refresh(admin)
+
+    task = Task(user_id=admin.id, title=f"{owner}-task")
+    db.add(task)
+    db.commit()
+    db.refresh(task)
+
+    now = datetime.now()
+    for index, model_name in enumerate(model_names):
+        db.add(
+            TraceEvent(
+                task_id=task.id,
+                event_id=f"{owner}-{index}",
+                event_type="llm_call_start",
+                timestamp=now,
+                data={"model_name": model_name},
+            )
+        )
+    db.commit()
+    return admin
+
+
+async def test_usage_rate_is_each_model_share_of_all_calls() -> None:
+    """Each rate is the model's share of the total, not a constant 100.
+
+    A lopsided split is the point: with equal counts every candidate formula
+    agrees, so only an uneven distribution tells a real share of the total
+    apart from the constant the handler used to return.
+    """
+    db = _direct_db_session()
+    try:
+        admin = _seed_model_calls(
+            db,
+            ["gpt-4o", "gpt-4o", "gpt-4o", "claude-opus"],
+            owner="lopsided-admin",
+        )
+
+        stats = await get_model_stats(db=db, current_user=admin)
+
+        assert {entry["name"]: entry["usage_rate"] for entry in stats} == {
+            "gpt-4o": 75.0,
+            "claude-opus": 25.0,
+        }
+        # The call counts the rates are derived from, so a regression that
+        # fixed the ratio by miscounting cannot hide here.
+        assert {entry["name"]: entry["total_tasks"] for entry in stats} == {
+            "gpt-4o": 3,
+            "claude-opus": 1,
+        }
+    finally:
+        db.close()
+
+
+async def test_single_model_uses_the_whole_budget() -> None:
+    """One model with all the traffic really is at 100%.
+
+    Kept alongside the lopsided case so the fix is pinned from both sides: the
+    old code returned 100.0 here too, and a fix that merely stopped returning
+    100.0 would be just as wrong.
+    """
+    db = _direct_db_session()
+    try:
+        admin = _seed_model_calls(db, ["gpt-4o"], owner="solo-admin")
+
+        stats = await get_model_stats(db=db, current_user=admin)
+
+        assert [(entry["name"], entry["usage_rate"]) for entry in stats] == [
+            ("gpt-4o", 100.0)
+        ]
+    finally:
+        db.close()
+
+
+async def test_nameless_calls_stay_out_of_the_denominator() -> None:
+    """Calls the response omits must not shrink the rates it reports.
+
+    The query filters a NULL model name but not an empty one, so an empty name
+    reaches Python and is dropped there. Once the total became a real
+    denominator, dropping such a row from the output while still counting it
+    would leave the reported rates summing to less than 100.
+    """
+    db = _direct_db_session()
+    try:
+        admin = _seed_model_calls(db, ["gpt-4o", "", ""], owner="nameless-calls-admin")
+
+        stats = await get_model_stats(db=db, current_user=admin)
+
+        assert [(entry["name"], entry["usage_rate"]) for entry in stats] == [
+            ("gpt-4o", 100.0)
+        ]
+    finally:
+        db.close()
--- a/tests/web/api/test_monitor_model_stats.py
+++ b/tests/web/api/test_monitor_model_stats.py
@@ -30,18 +30,21 @@ from .conftest import _direct_db_session
 pytestmark = pytest.mark.usefixtures("_test_db")
 
 
-def _seed_model_calls(db: Session, model_names: list[str], *, owner: str) -> User:
-    """One admin with one ``llm_call_start`` event per name in ``model_names``.
+def _seed_model_calls(
+    db: Session, model_names: list[str], *, owner: str, is_admin: bool = True
+) -> User:
+    """One user with one ``llm_call_start`` event per name in ``model_names``.
 
-    Repeat a name to give that model more calls. ``owner`` keeps usernames
-    distinct so a test never depends on another test's rows surviving.
+    Repeat a name to give that model more calls. ``owner`` must differ between
+    calls within a test -- ``User.username`` is ``unique=True``, so reusing one
+    raises rather than merely mixing two users' traffic together.
     """
-    admin = User(username=owner, password_hash="hash", is_admin=True)
-    db.add(admin)
+    user = User(username=owner, password_hash="hash", is_admin=is_admin)
+    db.add(user)
     db.commit()
-    db.refresh(admin)
+    db.refresh(user)
 
-    task = Task(user_id=admin.id, title=f"{owner}-task")
+    task = Task(user_id=user.id, title=f"{owner}-task")
     db.add(task)
     db.commit()
     db.refresh(task)
@@ -62,7 +65,7 @@ def _seed_model_calls(db: Session, model_names: list[str], *, owner: str) -> Use
             )
         )
     db.commit()
-    return admin
+    return user
 
 
 async def test_usage_rate_is_each_model_share_of_all_calls() -> None:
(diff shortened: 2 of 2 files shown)


tests/web/api/test_monitor_model_stats.py
```diff
@@ -0,0 +1,133 @@
...
+    db.add(admin)
+    db.commit()
+    db.refresh(admin)
+
+    task = Task(user_id=admin.id, title=f"{owner}-task")
+    db.add(task)
+    db.commit()
+    db.refresh(task)
+
+    now = datetime.now()
```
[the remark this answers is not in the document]
codeacme17: Adopted in 93326f57, though for a different reason than the one given — the premise is right but the inference does not reach this test.

`TraceEvent.timestamp` is indeed `DateTime(timezone=True)` (`src/xagent/web/models/task.py:479`). But `get_model_stats` never reads it: it filters on `event_type` and the extracted `model_name`, groups by the latter, and counts `event_id`. The column is populated in the helper only because it is `nullable=False`, and all three tests are indifferent to its value. They are also SQLite-only (no `postgresql` marker), so the PostgreSQL session-`TimeZone` coercion described here is never executed by them — and SQLAlchemy's SQLite bind processor drops `tzinfo` regardless, so the change is invisible to storage as well as to the assertions.

One correction: "matching the timezone-awareness of the production code" is backwards for this file. The three naive `datetime.now()` call sites under `src/xagent/web/` are all in `monitor.py` itself (lines 284, 636, 653); the other 106 already pass `timezone.utc`. The sibling suite that seeds these same rows for this same handler is naive too (`tests/web/api/test_monitor_postgresql.py:115`) — and that one really does run on PostgreSQL, harmlessly.

So the change is worth making on house-style grounds (106 of 109 call sites, and production writes aware UTC into this column at `websocket.py:774-776`), not because anything here was at risk. I added a comment recording that no assertion depends on the value, so the next reader does not go hunting for a dependency that is not there.

The underlying concern does turn out to be real one layer down, just not in tests: those three naive boundaries in `monitor.py` are compared against this aware column, which skews the "today" windows in `/monitor/stats` and `/monitor/dashboard-stats` by the server's UTC offset. Pre-existing and outside this PR's diff, so filed as #1256.

codeacme17:
Thanks for the review. Addressed in 93326f57 — the seed helper now uses `datetime.now(timezone.utc)`.

Detail in the [inline reply](https://github.com/xorbitsai/xagent/pull/1254#discussion_r3751500853); the short version:

- **Change made.** Justified by house style rather than by risk: production writes aware UTC into this column (`websocket.py:774-776`) and 106 of the 109 `datetime.now(` call sites under `src/xagent/web/` already pass `timezone.utc`.
- **Not a latent bug in these tests.** `get_model_stats` never reads `timestamp` — it filters on `event_type` and the extracted `model_name`, groups by the latter, counts `event_id`. The tests are SQLite-only, and SQLAlchemy's SQLite bind processor discards `tzinfo` either way. Added a comment so the next reader is not left looking for a dependency that is not there.
- **One correction.** "Matching the timezone-awareness of the production code" is backwards here: the only three naive sites under `src/xagent/web/` are in `monitor.py` itself (284, 636, 653), and the sibling suite seeding the same rows for the same handler is naive as well (`test_monitor_postgresql.py:115`) — on PostgreSQL, harmlessly.
- **Real finding underneath.** Those three naive boundaries *are* compared against the tz-aware column, skewing the "today" windows in `/monitor/stats` and `/monitor/dashboard-stats` by the server's UTC offset. Pre-existing, outside this diff, filed as #1256.

`tests/web/api` still passes in full; ruff and mypy clean on the changed files.

src/xagent/web/api/monitor.py
```diff
@@ -571,25 +571,39 @@ async def get_model_stats(
...
+        # rejects a NULL model name but not an empty one, and counting calls
+        # that are never reported would deflate every rate that is.
+        counted_models = [
+            (model_name, model_calls)
+            for model_name, model_calls in model_stats
+            if model_name and model_calls > 0
+        ]
+
+        # Denominator for each model's share of the traffic. Deliberately not
+        # named ``total_calls``: the loop below used to bind that same name to
```
rogercloud: nit: this comment narrates code that no longer exists — the loop variable below is now `model_calls`, so nothing "used to bind that same name" in the current file. The anti-regression note is worth keeping; consider stating the invariant instead, e.g. "must not be shadowed by the loop's per-model variable — see #1245".

src/xagent/web/api/monitor.py
```diff
@@ -571,25 +571,39 @@ async def get_model_stats(
             logger.error(f"Failed to query model stats: {e}")
             model_stats = []
 
-        # Get total call count for calculating usage rate
-        total_calls = sum(stat.total_calls for stat in model_stats)
+        # Rows the response will carry. Filtered here rather than inside the
+        # loop so a dropped row stays out of the denominator too: the query
+        # rejects a NULL model name but not an empty one, and counting calls
+        # that are never reported would deflate every rate that is.
+        counted_models = [
```
rogercloud, line 578: Design note, no action needed: NULL names are rejected in SQL while empty names are dropped here in Python — two layers for one concern. Worth noting the reason unification isn't trivial: PostgreSQL's extraction returns plain text so `!= ''` would compose at the SQL layer, but MySQL's `json_extract()` returns a JSON-typed value and would need dialect-specific unquoting.

src/xagent/web/api/monitor.py
```diff
@@ -571,25 +571,39 @@ async def get_model_stats(
             logger.error(f"Failed to query model stats: {e}")
             model_stats = []
 
-        # Get total call count for calculating usage rate
-        total_calls = sum(stat.total_calls for stat in model_stats)
+        # Rows the response will carry. Filtered here rather than inside the
+        # loop so a dropped row stays out of the denominator too: the query
+        # rejects a NULL model name but not an empty one, and counting calls
```
rogercloud, line 576: minor: this reads as dialect-independent, but the Python-side drop only holds for SQLite/PostgreSQL. On the MySQL branch (monitor.py:141-146, unchanged by this PR) `json_extract()` appears to return an empty name as the JSON-quoted literal `""` — a truthy string — since nothing there unquotes the result. The gap itself is pre-existing; only this comment newly asserts uniform behavior. A one-line caveat or a separate tracked issue (as with #1256) would cover it.

tests/web/api/test_monitor_model_stats.py
```diff
@@ -0,0 +1,137 @@
...
+from .conftest import _direct_db_session
+
+pytestmark = pytest.mark.usefixtures("_test_db")
+
+
+def _seed_model_calls(db: Session, model_names: list[str], *, owner: str) -> User:
+    """One admin with one ``llm_call_start`` event per name in ``model_names``.
+
+    Repeat a name to give that model more calls. ``owner`` keeps usernames
+    distinct so a test never depends on another test's rows surviving.
```
rogercloud: nit: this rationale is moot — the `_test_db` fixture in conftest.py drops tables per test, so one test's rows can never survive into another. The distinct-owner pattern is still worth keeping for a different reason: the seeded user is an admin and sees events fleet-wide, so distinct owners prevent contamination *within* a single test body if the helper is called more than once. Suggest rewording only.

tests/web/api/test_monitor_model_stats.py
```diff
@@ -0,0 +1,137 @@
...
+                event_type="llm_call_start",
+                timestamp=now,
+                data={"model_name": model_name},
+            )
+        )
+    db.commit()
+    return admin
+
+
+async def test_usage_rate_is_each_model_share_of_all_calls() -> None:
```
rogercloud, line 71: nit (test completeness): consider adding coverage for the empty-result path (`if not result: return []`) and the query-exception path (`except Exception: model_stats = []`) in monitor.py. Neither performs a division in the old or new code, so this isn't a regression risk from the removed zero-guard — just a gap.

tests/web/api/test_monitor_model_stats.py
```diff
@@ -0,0 +1,137 @@
...
+pytestmark = pytest.mark.usefixtures("_test_db")
+
+
+def _seed_model_calls(db: Session, model_names: list[str], *, owner: str) -> User:
+    """One admin with one ``llm_call_start`` event per name in ``model_names``.
+
+    Repeat a name to give that model more calls. ``owner`` keeps usernames
+    distinct so a test never depends on another test's rows surviving.
+    """
+    admin = User(username=owner, password_hash="hash", is_admin=True)
```
rogercloud: nit (test completeness): all three tests seed `is_admin=True`, so the non-admin branch (monitor.py:541-548), where the denominator becomes the caller's own traffic rather than the fleet's, stays untested — here and in the existing suite. Pre-existing, but more meaningful now that the denominator is load-bearing.

rogercloud (approved):
## Summary

`get_model_stats` computed each model's `usage_rate` by dividing a model's own call count by itself: the grand total was summed into `total_calls`, and the very next `for model_name, total_calls in model_stats:` loop rebound that same name to the per-model count, making every rate `n / n * 100` = `100.0` (#1245). This PR renames the aggregate to `all_model_calls` and the loop variable to `model_calls` so the division is per-model-over-fleet, and moves the "drop rows with no model name / no calls" filter ahead of the sum so dropped rows can no longer inflate the denominator now that the denominator is actually read. A new SQLite-level test file covers a lopsided 3:1 split, the single-model 100% case, and exclusion of nameless calls.

## Design verdict: sound

This is the minimal correct fix for the root cause. Hoisting the filter above the sum is not scope creep — it is required for correctness (reverting just that reordering makes the nameless-calls test go from `100.0` to `33.3`), and dropping the old `else 0` zero-guard is provably safe: the division only runs inside the loop over `counted_models`, and a non-empty `counted_models` implies every element has `model_calls > 0`, so the sum is `>= 1` on every reachable path.

## Findings (all optional — none block merge)

Six minor/nit items are left as inline comments below (comment accuracy, a design trade-off worth naming, and a couple of optional test-completeness additions). None of them are correctness bugs in the fixed code.

## No issues found

- **Security / authorization**: the admin vs. non-admin query filter is unchanged and correctly scoped; no new authz surface.
- **Performance**: same single query; one added O(n) list comprehension.
- **Backward compatibility / API contract**: response shape is unchanged; only `usage_rate` *values* change, which is the point of the fix. `total_tasks` still carries LLM-call counts rather than task counts, and scoping that rename out as a separate wire-contract change is the right call.
- **Code duplication**: none introduced.

## Testing performed

The three new tests pass against the fix, and the distribution test is verified red against the pre-fix code (both entries `100.0` instead of `75.0` / `25.0`). `ruff check` and `ruff format --check` are clean on both changed files, `mypy` is clean, and all CI checks on the PR are green.

The earlier Gemini-bot finding about a naive `datetime.now()` in the seeding helper is resolved as of `93326f57` (now `datetime.now(timezone.utc)`), with the related-but-distinct timezone-comparison gap tracked separately in #1256 — not re-raised here.

## Recommendation

**Approve** — with the items below as optional follow-ups. Nothing found is a correctness bug in the fixed code, and none of it blocks merge.

src/xagent/web/api/monitor.py
codeacme17: Agreed — fixed in 781530be. The comment explained the name by describing code that is no longer in the file, which is exactly the shape that rots. Now stated as the invariant:

```python
# Denominator for each model's share of the traffic. Must stay distinct
# from the loop's per-model variable: while the two shared a name the
# rate divided a model's calls by itself and every model reported 100%
# (#1245).
```

src/xagent/web/api/monitor.py
codeacme17: Noted, and no change made — recording the reasoning here since the thread is the natural place for it.

The PostgreSQL half is right: `->>` yields text (line 140), so `!= ''` would compose at the SQL layer. The MySQL half turns out to be moot — MySQL is not a supported application backend (no MySQL driver in any `pyproject.toml` extra or group, `docker-compose.yml` provisions `postgres:16-bookworm` with `DATABASE_URL=postgresql://…`, and CLAUDE.md documents SQLite and PostgreSQL only), so that branch cannot be reached. Filed as #1266 — see the reply on the sibling thread.

One refinement on "two layers for one concern": the Python-side filter is not purely the empty-name check. It also carries `model_calls > 0`, and its position ahead of the sum is what keeps dropped rows out of the denominator. Pushing the name check into SQL would leave that behind, so the layers would not actually collapse to one.

src/xagent/web/api/monitor.py
codeacme17: Good catch, and it is worse than "appears to" — confirmed and filed as #1266.

`func.json_extract()` at lines 141-146 has no `JSON_UNQUOTE` anywhere in the file, so on MySQL **every** name would come back JSON-quoted (`"gpt-4o"`, not just `""`), across `model_name` in `/monitor/model-stats` and `/monitor/stats` and `tool_name` in `/monitor/popular-tools` — and quoted names would also defeat `GROUP BY` de-duplication against any unquoted comparison. The empty-name subcase you spotted is one symptom of a branch that is wrong for all inputs.

I did not add the caveat, because on checking, that branch is unreachable: MySQL is not a supported application backend (no MySQL driver in any `pyproject.toml` extra or group; `docker-compose.yml` provisions `postgres:16-bookworm` with `DATABASE_URL=postgresql://…`; CLAUDE.md documents SQLite and PostgreSQL only — the MySQL entries in `example.env` are for the SQL tool's external user databases, not the app session `get_json_field_expression` gets from `get_db()`). A caveat asserting MySQL behavior in my comment would document dead code as if it were live. #1266 proposes deleting the branch instead — it is byte-identical to the SQLite `else` below it apart from the comment, so falling through changes no reachable behaviour.

tests/web/api/test_monitor_model_stats.py
codeacme17: Right on both counts — reworded in 781530be, citing a constraint stronger than either of us named:

```python
``owner`` must differ between calls within a test --
``User.username`` is ``unique=True``, so reusing one
raises rather than merely mixing two users' traffic together.
```

`User.username` is `unique=True` (`src/xagent/web/models/user.py:22`), so a second call with the same owner raises `IntegrityError` outright rather than quietly pooling two users' traffic. And you are right that the original rationale was moot: `_test_db` lays a fresh temp SQLite file per test and drops all tables on teardown, so cross-test survival was never possible.

tests/web/api/test_monitor_model_stats.py
codeacme17: Split this one — acted on the first half, deferring the second, both in 781530be.

**Empty-result path: not coverable, so deleted instead.** `if not result: return []` sat directly ahead of `return result`, and both arms returned an equal empty list — no test could distinguish them, so any "coverage" would have been asserting against a branch that made no difference. The branch is gone, and a test now pins the behaviour it existed to express:

```python
async def test_no_calls_yields_an_empty_list() -> None:
    member = _seed_model_calls(db, [], owner="no-traffic-user")
    assert await get_model_stats(db=db, current_user=member) == []
```

**Query-exception path: deferred deliberately.** It is testable without brittle monkeypatching — a raw insert of malformed JSON into the `data` column makes SQLite's `json_extract` raise — but such a test would pin *silent swallowing* as intended behaviour, and that is the exact failure mode #1149 was filed about: the `except` returns an empty list and the dashboard shows zeros with only a log line to show for it. Pinning it here would make the eventual fix look like a regression. Worth resolving as "should this endpoint surface the error" first; happy to file that separately if you would like it tracked.

tests/web/api/test_monitor_model_stats.py
codeacme17: Most valuable of the six — added in 781530be.

```python
async def test_non_admin_shares_are_scoped_to_their_own_calls() -> None:
    _seed_model_calls(db, ["gpt-4o"] * 6, owner="other-tenant")
    member = _seed_model_calls(
        db, ["gpt-4o", "gpt-4o", "gpt-4o", "claude-opus"],
        owner="member", is_admin=False,
    )
    stats = await get_model_stats(db=db, current_user=member)
    # 75.0 / 25.0 -- fleet-wide would read 90.0 / 10.0
```

Your framing is what made it worth doing: the scoping was free to leave untested while the denominator went unread, but now a leak would not merely expose rows the caller should not see — it would silently restate the rates for the rows they should. The other tenant's traffic is sized so the two readings cannot be confused, and I verified the test detects the difference by neutering the `is_admin_user` check in `get_model_stats`: it reports 90.0 instead of 75.0. Counts are asserted too (3 and 1, not 9 and 1), so a leak cannot hide behind a coincidental ratio.

codeacme17:
Thanks for the thorough review — acted on four of the six in 781530be, with per-thread replies above.

| # | Item | Outcome |
|---|---|---|
| [1] | Stale denominator comment | **Fixed** — restated as an invariant instead of history |
| [2] | Two-layer filter design note | **No change** — see below |
| [3] | MySQL empty-name caveat | **Escalated** — filed as #1266 |
| [4] | Seed-helper docstring rationale | **Fixed** — cites `unique=True`, the real constraint |
| [5] | Empty-result / exception coverage | **Half done** — branch deleted, test added; exception path deferred |
| [6] | Non-admin denominator untested | **Fixed** — test added and verified |

Three worth expanding on:

**[6] is the one that mattered.** Your framing is what made it worth doing: the scoping was free to leave untested while the denominator went unread, but now a leak would silently restate the rates for the rows the caller *should* see, not just add ones they shouldn't. Verified by neutering the `is_admin_user` check — the test reads 90.0 instead of 75.0. Counts asserted too, so a leak can't hide behind a coincidental ratio.

**[3] is bigger than a caveat.** Confirmed: `func.json_extract()` at `monitor.py:141-146` has no `JSON_UNQUOTE`, so on MySQL *every* name would come back JSON-quoted (`"gpt-4o"`, not just `""`) — across `model_name` and `tool_name` alike, and quoted names would defeat `GROUP BY` de-duplication too. But that branch is unreachable: no MySQL driver in any `pyproject.toml` extra or group, `docker-compose.yml` provisions `postgres:16-bookworm`, CLAUDE.md documents SQLite and PostgreSQL only. Adding the caveat would have documented dead code as if it were live, so #1266 proposes deleting the branch — it is byte-identical to the SQLite `else` below it apart from the comment.

**[5] split.** The empty-result branch turned out uncoverable by construction — `if not result: return []` sat directly ahead of `return result`, both returning an equal empty list — so it is deleted and a test pins the behaviour instead. The exception path I deferred on purpose: a test there would pin *silent swallowing* as intended, which is the exact failure mode #1149 was filed about. Happy to file that as its own issue if you'd like it tracked.

`tests/web/api` passes in full (5 tests in the new file now); ruff and mypy clean on both changed files.

Issue #1082, opened 2026-07-30
feat(interactions): reconcile interaction state and liveness


AlexLiu190625:
## Summary

Repair contradictory Task/interaction states and converge permanently unresumable interactions without transport-owned guesses.

## Current behavior

Task status, checkpoint availability, command state, and transport presence can disagree after crashes or mixed-version operation. Socket absence and wall-clock age alone are not authoritative liveness signals.

## Acceptance criteria

- [ ] Run bounded reconciliation at startup, periodically, and lazily on read/respond/reconnect.
- [ ] Use a stable keyset cursor and bounded batches. PostgreSQL workers claim rows with `FOR UPDATE SKIP LOCKED`; SQLite workers take a detached snapshot and let Task/interaction CAS choose the winner.
- [ ] Keep retryable dependency failures pending and return a stable retryable-unavailable result.
- [ ] Converge authoritative permanent locator loss to `interaction_state_unavailable`, persisted expiry to `interaction_expired`, and source-access loss to `interaction_access_revoked`.
- [ ] Do not expire solely because no socket is connected or because an interaction is old when no `expires_at` was persisted.
- [ ] Never create a duplicate continuation command or overwrite an accepted response.
- [ ] Expose repair counts, reasons, lag, and terminal transitions; cover worker crash and concurrent respond/reconcile races.
- [ ] If implementation requires a durable lease across external I/O, stop and create a separately reviewed schema change; this issue does not add hidden claim columns.

## Dependencies

- [feat(interactions): add canonical task interaction schema](https://github.com/xorbitsai/xagent/issues/1074)
- [feat(interactions): implement task interaction lifecycle service](https://github.com/xorbitsai/xagent/issues/1075)
- [ops(interactions): materialize legacy waiting tasks](https://github.com/xorbitsai/xagent/issues/1078)
- [ops(interactions): add mixed-version rollout controls](https://github.com/xorbitsai/xagent/issues/1076)

## Non-goals

- Using socket presence as an attendance model.
- Automatically replaying unknown runtime side effects.

AlexLiu190625:
## Implementation design

- Reconciliation uses durable Task, interaction, checkpoint, and command facts. Transport adapters may request a lazy check but cannot choose the outcome.
- Expiry is the evaluated timestamp persisted at publication. Changing configuration later does not reinterpret an in-flight interaction.
[...]

Issue #1081, opened 2026-07-30
feat(frontend): submit structured clarification responses


AlexLiu190625:
## Summary

Preserve the clarification form's typed field values when submitting a response.

## Current behavior

The form normalizes fields but flattens the answer into `label: value` chat text before sending. That loses field identity and cannot satisfy typed response validation.

## Acceptance criteria

- [ ] Carry interaction ID, run identity, state/version, response idempotency key, and the typed field-value object through render and submit.
- [ ] Do not flatten structured values into display text for transport.
- [ ] Preserve raw booleans, numbers, option codes, arrays, file IDs, `false`, `0`, and valid empty collections independent of labels or locale.
- [ ] Generate one stable idempotency key per user submit attempt and reuse it for safe network retry.
- [ ] Render validation, stale/conflict, unauthorized, unavailable, and replay outcomes distinctly.
- [ ] Mark the response complete only after the accepted/replayed receipt; a failed or indeterminate submission preserves the draft.
- [ ] Keep ordinary chat send behavior unchanged outside an active clarification.

## Dependencies

- [feat(api): expose typed task interaction continuation](https://github.com/xorbitsai/xagent/issues/1079)
- [feat(web): route clarification through task interactions](https://github.com/xorbitsai/xagent/issues/1080)

## Non-goals

- Changing the backend lifecycle service.
- Adding approval cards.

AlexLiu190625:
## Implementation design

- Form state owns typed input collection; the API client owns wire serialization; the backend lifecycle service owns validation and acceptance.
- Optional human-readable text is display-only and is never reparsed. A separate chat projection may be rendered after acceptance.
- Unknown interaction kind/version fails closed with an upgrade/error state instead of falling back to chat text.

## Linked prerequisites

- [feat(api): expose typed task interaction continuation](https://github.com/xorbitsai/xagent/issues/1079)
- [feat(web): route clarification through task interactions](https://github.com/xorbitsai/xagent/issues/1080)

## Delivery boundary

One implementation PR in this repository should close this issue. Mechanical dependency updates do not close it. Tests must exercise the owning invariant and its failure/rollback path.

Issue #1080, opened 2026-07-30
feat(web): route clarification through task interactions


AlexLiu190625:
## Summary

Move the internal Web clarification backend onto the canonical interaction lifecycle without losing the existing chat experience.

## Current behavior

The Web path derives a pending question from messages and directly invokes the background resume manager. That path can diverge from V1/A2A lifecycle, idempotency, and error semantics.

## Acceptance criteria

- [ ] Read the pending clarification through the lifecycle service and submit a dedicated typed WebSocket interaction-response command.
- [ ] Remove native Web ownership of direct checkpoint loading and background resume.
- [ ] Wrap a legacy chat message only for an explicitly supported plain-text, fieldless clarification. Structured interactions and unknown kind/version fail closed.
- [ ] Reuse the same client message identifier for a logical response retry and return the canonical receipt.
- [ ] Project the canonical interaction onto chat/event metadata without creating a second workflow state.
- [ ] Link the presentation message to the interaction identifier and align replay/display behavior with the existing message contract.
- [ ] Cover reconnect, duplicate submit, stale form, legacy compatibility, and no-broadcast of responder-specific data.

## Dependencies

- [feat(interactions): implement task interaction lifecycle service](https://github.com/xorbitsai/xagent/issues/1075)
- [feat(api): expose typed task interaction continuation](https://github.com/xorbitsai/xagent/issues/1079)
- [ops(interactions): add mixed-version rollout controls](https://github.com/xorbitsai/xagent/issues/1076)

## Non-goals

- Frontend form serialization; that is a separate issue.
- Tool approval UI.
- Deleting background-task management used by ordinary run and cancellation lifecycle.

AlexLiu190625:
## Implementation design

- The backend adapter maps the canonical interaction to Web presentation and delegates state changes to the lifecycle service.
- Chat messages are presentation records. Their presence, absence, or text is never used to decide whether an interaction can resume.
- Broadcast may carry responder-invariant invalidation/presentation data only. Credentialed reads retrieve any responder-specific view.
- Compatibility text handling is explicitly gated by clarification kind/version and materialized origin.

## Linked prerequisites

- [feat(interactions): implement task interaction lifecycle service](https://github.com/xorbitsai/xagent/issues/1075)
- [feat(api): expose typed task interaction continuation](https://github.com/xorbitsai/xagent/issues/1079)
- [ops(interactions): add mixed-version rollout controls](https://github.com/xorbitsai/xagent/issues/1076)

[...]

Issue #1079, opened 2026-07-30
feat(api): expose typed task interaction continuation


AlexLiu190625:
## Summary

Expose authorized typed interaction reads and same-checkpoint responses through the normal V1 Task API.

## Current behavior

Task create/get/append/steps exist, but Task responses do not include a pending question and there is no interaction get/list/respond surface. Append excludes waiting Tasks, while nearby documentation implies it can resume them.

## Acceptance criteria

- [ ] Add authorized interaction get, cursor-paginated list, and respond endpoints with frozen request/response/error fixtures. Respond accepts a versioned envelope containing `kind`, structured `values`, optional display-only `text`, and an idempotency key.
- [ ] Expose `pending_interaction` on Task GET only as a convenience projection; history remains available through list/get.
- [ ] `respond()` returns an accepted/replayed receipt and resumes the same `run_id` through the lifecycle service.
- [ ] Append on `waiting_for_user` returns the dedicated `interaction_response_required` error, points callers to the response surface, and creates no run or command.
- [ ] Split cacheable base Task data from responder-specific interaction projection; task-wide cache entries never contain responder-variant payload.
- [ ] Migrate the V1 continuation path off direct resume-manager ownership.
- [ ] Cover authorization, pagination, idempotent replay, conflict, stale state, unknown version, cache isolation, and plain completed output.

## Dependencies

- [feat(interactions): implement task interaction lifecycle service](https://github.com/xorbitsai/xagent/issues/1075)
- [feat(interactions): publish native clarification interactions atomically](https://github.com/xorbitsai/xagent/issues/1077)
- [ops(interactions): add mixed-version rollout controls](https://github.com/xorbitsai/xagent/issues/1076)

## Non-goals

- Changing append into respond.
- Embedding responder-specific data in shared Task caches.
- Adding transport-specific identity to the core schema.

AlexLiu190625:
## Implementation design

- Routes own authentication, request limits, HTTP status mapping, and response projection. They do not reconstruct a question or perform Task/interaction CAS themselves.
- Response payloads remain typed objects. Optional plain text is display-only or valid input only when the clarification schema declares it; it is never reparsed into structured values or converted into append.
- Task GET may summarize the current pending interaction, but clients must use the interaction identifier and response contract to continue.
- The endpoint returns after durable acceptance; runtime completion remains observable through the existing Task status surface.

## Linked prerequisites

[...]

PR #1258, opened 2026-08-10
ops(interactions): add default-off mixed-version rollout controls

AlexLiu190625:
## Summary

Adds a policy-gated switch that decides whether a task may publish a
native interaction row, defaulting entirely off, plus the observability
needed to run a staged rollout safely. Nothing reads, answers, or closes
interaction rows differently based on this switch -- it only controls
whether a *new* native row can be produced. The gate has zero production
callers in this change; wiring it into a producer is a separate change.

## Default-off, verifiably

Three checkable facts back the "merged does not mean enabled" claim:

- The protocol mode defaults to `legacy`, and the gate returns a blocked
  decision unconditionally whenever the mode isn't `native` -- one
  attribute read and one string compare, no database query, no
  environment read (the policy is parsed once and frozen).
- The native-source allowlist defaults to empty, so switching only the
  mode without also configuring at least one source fails startup
  outright rather than silently doing nothing.
- The new `/ready` behavior only executes in `native` mode. In `legacy`
  or `read` mode it is skipped entirely -- zero added database query,
  zero behavior change from today.

## Why the gating key is Task.source, not the producer

The three things that currently write interaction-style state don't
carry a stable "which producer" signal by the time it would need to be
read back for gating -- that information is lost by the projection
layer. `Task.source` is: already persisted on every task, already used
in the field's own indexed queries, and shares one vocabulary with the
`origin` column's CHECK constraint, so a single normalizer serves both
concerns. Gating a rollout by producer identity was considered and
rejected because there is no reliable read-side signal to gate on.

It also lines up the rollout axis with the risk axis. Reading,
answering, and closing an interaction migrate to native handling on a
per-ingress-channel schedule -- the SDK client, the internal web/socket
path, and each IM channel move at their own pace, in their own changes.
Gating on `Task.source`, the same value that already separates those
channels, means a single allowlist entry controls exactly the traffic
whose read/respond/close path has actually been migrated -- no
broader and no narrower.

## The synthetic "channel" gating source

`Task.source == "internal"` covers three very different entry points --
the web UI, WebSocket sessions, and IM channel bots -- as one bucket.
Every task created by an IM channel integration also carries a non-null
`channel_id`, so the gate splits that traffic into its own key,
`"channel"`, computed only when `origin == "internal"` (a task whose
source is explicitly `widget` and which also happens to carry a
`channel_id` must not be misclassified as channel traffic). This key
exists purely to let operators enable IM-channel rollout independently
of the rest of `internal` traffic -- it is never written to the
`origin` column and never added to that column's CHECK constraint.

## Two departures from this codebase's usual configuration pattern

Existing three-state mode switches in this codebase warn on an invalid
value and fall back to a default. This one raises at startup instead,
and reading the resolved policy before startup validation has run
raises rather than lazily parsing the environment. Both departures
exist for the same reason: a rollout switch that silently stays in
`legacy` while an operator believes they configured `native` mode is
precisely the blind spot this change exists to close. Failing loudly at
startup turns a silent misconfiguration into an immediate, visible one.

## One degradation signal is intentionally never auto-cleared

Two new operational signals ride along on `/health`: one for a missing
`task_interaction_requests` table (clears itself once the table is
observed present) and one for an unrecognized `Task.source` value on a
task. The second is never auto-cleared. An unrecognized source value is
a property of already-persisted data; nothing in this process can
observe that data being corrected, so auto-clearing would assert
evidence that doesn't exist. There is no clearing endpoint or command for
it -- the registry that tracks it lives only in process memory, so it
stays active for the life of the process.

## Two normalization rules, on the same values, by design

`Task.source` is normalized without stripping whitespace or changing
case: it is a value written by application code, so an unexpected
reading of it -- a stray space, a wrong case -- is itself a fact worth
surfacing rather than silently coercing into a known bucket. The
operator-facing `XAGENT_INTERACTION_NATIVE_SOURCES` environment
variable, by contrast, is stripped and lowercased, because it is a
human-typed value where that kind of normalization is exactly what
operators expect. These are not the same rule applied inconsistently --
each guards a different kind of input, and neither should be changed to
match the other.

## Unauthenticated-surface tradeoffs, stated plainly

- The `/ready` native-mode check adds a request-triggered database query
  to an unauthenticated endpoint. A one-way latch (cache "present"
  forever, never cache "absent") brings the steady-state cost to zero
  once the schema exists; the residual risk is that unauthenticated
  traffic can trigger a repeated catalog query for as long as the schema
  remains absent.
- The set of active degradation *names* on `/health` is itself
  information: it tells an unauthenticated caller that this deployment
  has interaction rollout enabled and that a specific precondition
  (schema presence, or a data-quality issue) is unmet. This follows the
  same shape every existing degradation signal already uses; recorded
  here as a known tradeoff, not something newly introduced by this
  change.

## Rollback

To roll back a native-mode rollout: (1) stop producing new native
rows -- switch the mode back or remove a source from the allowlist, then
restart; (2) reading, answering, and closing existing rows keeps working
throughout, since none of that logic is mode-gated; (3) watch the admin
diagnostics endpoint's active-row count drain to zero; (4) once drained,
downgrade fully to `legacy`; (5) removing the compatibility path
entirely is a separate, later change.

Unsafe during rollback: reverting the code instead of the
configuration; rolling back a database migration while the active-row
count is still above zero; and skipping stages by jumping directly from
`legacy` to `native` without an observation window in `read` mode.

All replicas of a deployment must run the same configuration. Rolling
out by replica (some on `native`, some not) is not supported --
different replicas would answer the same task's interactions
differently depending on which replica happened to handle a given turn.
This does not lose data; it produces an inconsistent experience across
requests for the same task, which is reason enough to avoid it.

## Enablement prerequisites

Each entry in the native-source allowlist has its own readiness
requirement before it should be enabled. An entry should only move into
the allowlist once every deployment it depends on for reading, answering,
and closing that traffic's interactions is live:

- `sdk`: depends on #1079 (typed task-interaction continuation for SDK
  clients), which covers both the read path and the response path for
  this traffic in one change.
- `internal` (Web UI and WebSocket traffic that isn't split out as
  `channel`): depends on #1080 (routing clarification through task
  interactions on the internal read path), #1075 (the response/lifecycle
  service), and #1081 (front-end structured-response submission --
  without it, a clarification would render but have no way to be
  answered).
- `widget`: depends on #1075 plus a reader for the embedded-chat-window
  projection.
- `shared_link`: shares the same projection layer and #1075 dependency
  as `widget`, but needs its own verification pass since its
  authorization context differs.
- `a2a`: depends on #1075 plus a reader in the a2a protocol's own
  projection. This is the sharpest case for older-client compatibility,
  since a2a consumers are machine clients rather than a UI a human can
  just reload.
- `channel` (the synthetic IM-channel split): no reader for
  channel-originated native rows exists yet. Do not add this entry
  until one does.
- `trigger`: trigger-originated tasks are hidden tasks with no
  interactive responder. Enabling this source will never produce an
  answerable interaction, and startup logs a warning if it is present
  in the allowlist. It should not be enabled under normal operation.

## What this change does not cover

No producer writes a native interaction row yet, and no consumer reads
one differently based on this policy -- both are separate, later
changes. A full mixed-version test matrix therefore cannot be exercised
meaningfully here: the producers and consumers it would need to exist
don't exist yet on this branch. Each cell belongs to the change that
introduces the producer or consumer it needs:

| Matrix cell | Covered by |
|---|---|
| old producer / new consumer | A later change to the read path |
| new producer / new consumer | #1079, #1080 |
| lost notification | #1082 |
| producer crash mid-write | #1077 |
| route-level rollback | A later change |
| older-client compatibility | #1079, #1081 |

## Size note

This change is larger than a typical PR: 12 files, 2,274 lines added and
1 removed. Tests are the dominant share, plainly: the five test files
account for about 1,480 of those lines, versus roughly 790 across the
seven production files. The static guards that keep read-side consumers
from ever importing this module, and that bound how many production
call sites its own accessor functions may have, need to parse and walk
real production files to do that job correctly, and the
configuration/gate/observability surfaces each carry a full pass/fail
matrix. That is what accounts for the ratio -- not padding.

### DIFF
--- a/src/xagent/web/api/admin_interaction_rollout.py
+++ b/src/xagent/web/api/admin_interaction_rollout.py
@@ -74,21 +74,29 @@ async def get_interaction_rollout_diagnostics(
                 "WHERE status = 'active' AND active_slot IS NOT NULL"
             )
         ).one()
+
+        # Result-type marshaling for the raw column value differs by
+        # backend: PostgreSQL hands back a tz-aware datetime, SQLite hands
+        # back a plain str (its DATETIME type has no native temporal type).
+        # Both shapes -- and any further marshaling surprise -- are handled
+        # in this same try so they land in the deliberate 503 branch below,
+        # never as an unhandled 500.
+        oldest_created_at = row.oldest_created_at
+        oldest_age_seconds: float | None = None
+        if oldest_created_at is not None:
+            if isinstance(oldest_created_at, str):
+                oldest_created_at = datetime.fromisoformat(oldest_created_at)
+            if oldest_created_at.tzinfo is None:
+                oldest_created_at = oldest_created_at.replace(tzinfo=timezone.utc)
+            oldest_age_seconds = (
+                datetime.now(timezone.utc) - oldest_created_at
+            ).total_seconds()
     except Exception as exc:
         logger.exception("Interaction rollout diagnostics stats query failed")
         raise HTTPException(
             status_code=503,
             detail="Interaction rollout diagnostics stats query failed",
         ) from exc
-
-    oldest_created_at = row.oldest_created_at
-    oldest_age_seconds: float | None = None
-    if oldest_created_at is not None:
-        if oldest_created_at.tzinfo is None:
-            oldest_created_at = oldest_created_at.replace(tzinfo=timezone.utc)
-        oldest_age_seconds = (
-            datetime.now(timezone.utc) - oldest_created_at
-        ).total_seconds()
 
     response["active_count"] = row.active_count
     response["oldest_age_seconds"] = oldest_age_seconds
--- a/src/xagent/web/api/admin_interaction_rollout.py
+++ b/src/xagent/web/api/admin_interaction_rollout.py
@@ -1,0 +1,95 @@
+"""Admin diagnostic endpoint for the interaction rollout policy and gate state.
+
+Read-only, admin-authenticated. This is a diagnostic surface, not part of
+the gate itself: it queries active-row stats on demand rather than caching
+them, so a stale reading is never possible, at the cost of a query per
+request.
+"""
+
+from __future__ import annotations
+
+import logging
+from datetime import datetime, timezone
+
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy import text
+from sqlalchemy.orm import Session
+
+from ..auth_dependencies import get_current_user
+from ..models.database import get_db
+from ..models.user import User
+from ..services.interaction_rollout import (
+    counters_snapshot,
+    get_interaction_rollout_policy,
+)
+from ..services.ops_signals import active_degradations
+from ..services.task_interaction_schema import interaction_requests_table_exists
+
+router = APIRouter(
+    prefix="/api/admin/interaction-rollout", tags=["admin-interaction-rollout"]
+)
+logger = logging.getLogger(__name__)
+
+
+@router.get("")
+async def get_interaction_rollout_diagnostics(
+    current_user: User = Depends(get_current_user),
+    db: Session = Depends(get_db),
+) -> dict[str, object]:
+    """Resolved policy, active-row stats, gate counters, and degradations.
+
+    Failure shape is three-way, not the usual two: table absent is a normal
+    pre-migration deployment state and returns 200 with ``schema_absent``
+    true and no count fields (an empty allow list read as "0 active rows"
+    would be misleading -- there is no table to have rows). A stats query
+    that raises (a transient DB error) returns 503, not a fabricated
+    ``active_count: 0`` -- a 0 here is the specific number operators watch
+    to confirm a rollback has drained, and a fake 0 during an outage would
+    read as "drained" when the truth is "unknown."
+    """
+    if not current_user.is_admin:
+        raise HTTPException(status_code=403, detail="Admin access required")
+
+    policy = get_interaction_rollout_policy()
+    response: dict[str, object] = {
+        "policy": {
+            "mode": policy.mode,
+            "native_sources": sorted(policy.native_sources),
+            "native_protocol_version": policy.native_protocol_version,
+        },
+        "counters": counters_snapshot(),
+        "degradations": active_degradations(),
+    }
+
+    if not interaction_requests_table_exists(db):
+        response["schema_absent"] = True
+        return response
+
+    try:
+        row = db.execute(
+            text(
+                "SELECT count(*) AS active_count, "
+                "min(created_at) AS oldest_created_at "
+                "FROM task_interaction_requests "
+                "WHERE status = 'active' AND active_slot IS NOT NULL"
+            )
+        ).one()
+    except Exception as exc:
+        logger.exception("Interaction rollout diagnostics stats query failed")
+        raise HTTPException(
+            status_code=503,
+            detail="Interaction rollout diagnostics stats query failed",
+        ) from exc
+
+    oldest_created_at = row.oldest_created_at
+    oldest_age_seconds: float | None = None
+    if oldest_created_at is not None:
+        if oldest_created_at.tzinfo is None:
+            oldest_created_at = oldest_created_at.replace(tzinfo=timezone.utc)
+        oldest_age_seconds = (
+            datetime.now(timezone.utc) - oldest_created_at
+        ).total_seconds()
+
+    response["active_count"] = row.active_count
+    response["oldest_age_seconds"] = oldest_age_seconds
+    return response
--- a/src/xagent/web/api/admin_interaction_rollout.py
+++ b/src/xagent/web/api/admin_interaction_rollout.py
@@ -61,11 +61,11 @@ async def get_interaction_rollout_diagnostics(
         "degradations": active_degradations(),
     }
 
-    if not interaction_requests_table_exists(db):
-        response["schema_absent"] = True
-        return response
-
     try:
+        if not interaction_requests_table_exists(db):
+            response["schema_absent"] = True
+            return response
+
         row = db.execute(
             text(
                 "SELECT count(*) AS active_count, "
--- a/src/xagent/web/app.py
+++ b/src/xagent/web/app.py
@@ -78,16 +78,27 @@ from .dynamic_memory_store import get_memory_store
 from .logging_config import setup_logging
 from .models.database import init_db
 from .services.a2a_protocol import A2AApiError, a2a_api_error_handler, a2a_error
-from .services.interaction_rollout import validate_interaction_rollout_at_startup
+from .services.interaction_rollout import (
+    get_interaction_rollout_policy,
+    is_native_schema_ready,
+    mark_native_schema_ready,
+    validate_interaction_rollout_at_startup,
+)
 from .services.local_browser_runtime import (
     register_local_browser_runtime,
     unregister_local_browser_runtime,
+)
+from .services.ops_signals import (
+    INTERACTION_ROLLOUT_SCHEMA_ABSENT,
+    clear_degradation,
+    register_degradation,
 )
 from .services.orphan_upload_gc import run_orphan_upload_gc_loop
 from .services.skill_runtime import (
     SkillRuntimeSessionBoundaryError,
     skill_runtime_session_boundary_error_handler,
 )
+from .services.task_interaction_schema import interaction_requests_table_exists
 from .services.task_lease_recovery import run_task_lease_recovery_loop
 from .services.uploaded_file_recovery import (
     run_uploaded_file_compensation_recovery_loop,
@@ -704,9 +715,12 @@ async def readiness_check() -> JSONResponse:
     schema has been observed present, this endpoint never queries for it
     again for the lifetime of the process (the table is only ever added,
     never dropped, so a stale "present" reading cannot happen). In legacy
-    or read mode the whole segment is skipped -- zero added behavior, zero
-    added query -- so the default deployment stance is unaffected byte for
-    byte.
+    or read mode the schema check and the database query it would trigger
+    are skipped entirely -- zero added query. The frozen policy read
+    above the mode check always runs regardless of mode, including its
+    RuntimeError-if-uninitialized contract (see
+    ``get_interaction_rollout_policy``'s docstring); that read is cheap
+    (no I/O once the policy is frozen at startup) but it is not skipped.
     """
     task = getattr(app.state, "file_storage_startup_sync_task", None)
     error = getattr(app.state, "file_storage_startup_sync_error", None)
@@ -725,19 +739,6 @@ async def readiness_check() -> JSONResponse:
             },
         )
 
-    from .models.database import get_session_local
-    from .services.interaction_rollout import (
-        get_interaction_rollout_policy,
-        is_native_schema_ready,
-        mark_native_schema_ready,
-    )
-    from .services.ops_signals import (
-        INTERACTION_ROLLOUT_SCHEMA_ABSENT,
-        clear_degradation,
-        register_degradation,
-    )
-    from .services.task_interaction_schema import interaction_requests_table_exists
-
     policy = get_interaction_rollout_policy()
     if policy.mode == "native" and not is_native_schema_ready():
         # If the database itself is unreachable here (connection refused,
@@ -748,6 +749,18 @@ async def readiness_check() -> JSONResponse:
         # 503 would, and folding "database unreachable" into that same
         # typed 503 would misreport a connectivity outage as a pending
         # migration.
+        #
+        # get_session_local is imported here rather than hoisted with the
+        # policy/schema imports above: tests replace
+        # database_module.get_session_local itself with a stub session
+        # factory (see tests/web/test_interaction_rollout_observability.py),
+        # and this import must re-resolve that name from the module on
+        # every call for the replacement to take effect. A module-level
+        # import would bind the original function once at process start
+        # and keep calling it through that binding, silently defeating the
+        # test's monkeypatch.
+        from .models.database import get_session_local
+
         SessionLocal = get_session_local()
         db = SessionLocal()
         try:
--- a/src/xagent/web/app.py
+++ b/src/xagent/web/app.py
@@ -740,6 +740,14 @@ async def readiness_check() -> JSONResponse:
 
     policy = get_interaction_rollout_policy()
     if policy.mode == "native" and not is_native_schema_ready():
+        # If the database itself is unreachable here (connection refused,
+        # timeout), that exception propagates out of this route unhandled
+        # and FastAPI turns it into a 500 -- asymmetric with the deliberate
+        # 503 below for "database reachable, schema not yet migrated".
+        # Accepted: a 500 still fails the readiness probe the same way a
+        # 503 would, and folding "database unreachable" into that same
+        # typed 503 would misreport a connectivity outage as a pending
+        # migration.
         SessionLocal = get_session_local()
         db = SessionLocal()
         try:
--- a/src/xagent/web/app.py
+++ b/src/xagent/web/app.py
@@ -39,6 +39,7 @@ from ..core.execution_scope import (
 from ..core.file_storage import StorageKeyScopeError
 from ..core.tracing.langfuse import flush_langfuse, initialize_langfuse
 from .api.a2a import router as a2a_router
+from .api.admin_interaction_rollout import router as admin_interaction_rollout_router
 from .api.admin_mcp import admin_mcp_router
 from .api.admin_users import router as admin_users_router
 from .api.agent_api_keys import router as agent_api_keys_router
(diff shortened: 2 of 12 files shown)


tests/web/services/test_interaction_rollout_guards.py
```diff
@@ -0,0 +1,441 @@
...
+        (path, lineno, attr) for path, lineno, attr, is_self in hits if not is_self
+    ]
+    assert non_self == [], f"non-self post-construction mutation(s) found: {non_self}"
+
+    self_sites = {
+        (str(path.relative_to(web_root)), attr)
+        for path, _lineno, attr, is_self in hits
+        if is_self
+    }
+    assert self_sites == _TASK_ATTRIBUTE_MUTATION_EXEMPT_SITES, (
```
rogercloud, line 440: major: This asserts exact set equality (`self_sites == _TASK_ATTRIBUTE_MUTATION_EXEMPT_SITES`) against a hardcoded 4-file list. Reproduced: adding a new file with a class doing `self.channel_id = ...` (a legitimate new IM-channel bot, unrelated to the rollout invariant this guard protects) fails this test with a misleading "exemption set drifted" message.

Suggested fix — assert containment instead of equality so the guard still catches removals/unexpected sites without failing on legitimate additions:
```python
assert _TASK_ATTRIBUTE_MUTATION_EXEMPT_SITES <= self_sites
```
Better still, narrow the non-self scan to receivers plausibly typed as `Task`.

src/xagent/web/app.py
```diff
@@ -712,6 +724,55 @@ async def readiness_check() -> JSONResponse:
...
+        mark_native_schema_ready,
+    )
+    from .services.ops_signals import (
+        INTERACTION_ROLLOUT_SCHEMA_ABSENT,
+        clear_degradation,
+        register_degradation,
+    )
+    from .services.task_interaction_schema import interaction_requests_table_exists
+
+    policy = get_interaction_rollout_policy()
```
rogercloud, line 742: minor: `get_interaction_rollout_policy()` is called unconditionally here, before the `mode == "native"` check. It raises `RuntimeError` if `validate_interaction_rollout_at_startup()` hasn't run yet. Reproduced: a `TestClient(app)` built without the `with` context manager (skips lifespan startup) hitting `/ready` gets an unhandled `RuntimeError` → 500, contradicting "skipped entirely — zero behavior change" for legacy mode.

No current test or real deployment path triggers this (ASGI servers complete lifespan before accepting traffic), so it's latent rather than live — but worth guarding defensively before more callers depend on `/ready`.

src/xagent/web/api/admin_interaction_rollout.py
```diff
@@ -0,0 +1,103 @@
...
+        "policy": {
+            "mode": policy.mode,
+            "native_sources": sorted(policy.native_sources),
+            "native_protocol_version": policy.native_protocol_version,
+        },
+        "counters": counters_snapshot(),
+        "degradations": active_degradations(),
+    }
+
+    if not interaction_requests_table_exists(db):
```
rogercloud: minor: This call sits outside the `try:` block below (line 68), but `interaction_requests_table_exists()` itself calls `db.connection()` / `inspect(...).has_table(...)` with no internal exception handling — so a transient DB error here propagates as an unhandled 500, not the documented 503 ("a stats query that raises... returns 503, not a fabricated `active_count: 0`"). The existing test only patches `session.execute`, not this call.

Suggested fix: move this call inside the same `try:` block, and add a test that makes it raise.

rogercloud (approved):
## Summary

This PR adds a default-off policy gate (`src/xagent/web/services/interaction_rollout.py`, new, ~490 lines) that will eventually decide whether a task may publish a "native" interaction row, plus supporting model constants (`src/xagent/web/models/task_interaction.py`), an admin diagnostics endpoint (`src/xagent/web/api/admin_interaction_rollout.py`, new), `/ready` and `/health` wiring (`src/xagent/web/app.py`), a new `ops_signals` degradation entry, and two new config functions (`src/xagent/config.py`). The gate has zero production callers in this PR — it is model-only infrastructure for a staged rollout that later, separate changes will wire up, which the PR's own "What this change does not cover" section states plainly. 12 files changed, +2270/-1, with the 5 new test files accounting for the large majority of the additions.

The core mechanism — frozen policy singleton → cheap mode compare → gating key → allowlist → schema check — is correct, cheap, and genuinely default-off. The reservations are about scope rather than correctness: ~1,900 lines ship a switch nothing production-side consumes yet, and one of its three states is currently inert.

**`read` mode is a complete no-op today, but the PR treats it as meaningful.** The string `"read"` appears only in `_VALID_MODES`, the `Literal[...]` type, and one `cast(...)` in `src/xagent/web/services/interaction_rollout.py` — no code branches on `mode == "read"`. `evaluate_native_publication` only checks `policy.mode != "native"` (legacy and read are identical), and `/ready` only checks `policy.mode == "native"`. Yet the PR description's Rollback section lists "skipping stages by jumping directly from legacy to native without an observation window in read mode" as unsafe, which presupposes observability that `read` does not yet provide. Suggestion: either say so explicitly in the PR description, or drop `read` from `_VALID_MODES` until the change that gives it behavior lands.

**The synthetic `"channel"` gating key is not stable over a task's lifetime.** `gating_key()` (`src/xagent/web/services/interaction_rollout.py:102-126`) derives the `"channel"` key from `task.channel_id is not None`, read live at call time. `Task.channel_id` is `ForeignKey("user_channels.id", ondelete="SET NULL")`, enforced by a real migration, and `user_channels` rows are deletable via `src/xagent/web/api/channel.py`. Deleting the owning channel therefore NULLs an existing task's `channel_id`, and a later `gating_key()` call would silently reclassify that task from `"channel"` to `"internal"`. That undercuts the argument that gating is safe because it rests on values "already persisted on every task" — persisted is not the same as stable. Not observable today (zero production callers), but worth a one-line caveat in `gating_key()`'s docstring.

**The `interaction_rollout_unknown_task_source` degradation signal is permanent and rides on unauthenticated `/health`.** `clear_degradation` is never called for `INTERACTION_ROLLOUT_UNKNOWN_TASK_SOURCE`, so a single task triggering it makes `/health` report a degradation for the remaining life of the process; an uptime monitor alerting on "any active degradation" would page indefinitely over one anomalous data row, silenceable only by a restart that does not fix the underlying data. The PR body pre-defends this tradeoff and the reasoning is internally consistent, but the admin diagnostics endpoint already exposes the same information more usefully to an authenticated operator via `counters_snapshot()`'s `blocked_unknown_source` counter. Not a defect — raised as a design alternative.

## Findings

### major

**`tests/web/services/test_interaction_rollout_guards.py:386-441` — AST guard asserts exact set equality, will fail CI on unrelated legitimate changes**

The guard asserts `self_sites == _TASK_ATTRIBUTE_MUTATION_EXEMPT_SITES` against a hardcoded 4-file list (`feishu/bot.py`, `telegram/bot.py`, `slack/trace_handler.py`, `slack/bot.py`). Reproduced: adding a new file with a class doing `self.channel_id = ...` — a legitimate new IM-channel bot, unrelated to the rollout invariant this guard exists to protect — fails the test with a message reading as "the exemption set drifted," which is misleading.

Suggested fix: keep the guard's real purpose (catching removals and unexpected non-self mutation sites) while allowing new self-sites:

```python
# instead of: assert self_sites == _TASK_ATTRIBUTE_MUTATION_EXEMPT_SITES
assert _TASK_ATTRIBUTE_MUTATION_EXEMPT_SITES <= self_sites
```

Better still, narrow the non-self scan to receivers plausibly typed as `Task`, so the guard tracks the invariant instead of tracking the file list.

### minor

**`src/xagent/web/app.py:741-742` — `/ready` can 500 if hit before startup validation runs**

The handler calls `get_interaction_rollout_policy()` unconditionally, before checking `policy.mode == "native"`. That accessor raises `RuntimeError` when called before `validate_interaction_rollout_at_startup()`. Reproduced: a `TestClient(app)` built without the `with` context manager (which skips FastAPI lifespan startup) hitting `/ready` gets an unhandled `RuntimeError` → 500. This also contradicts "In legacy or read mode it is skipped entirely — zero behavior change from today": in legacy mode the schema query is skipped, but the policy read is not.

No current test or realistic deployment path triggers this — real ASGI servers complete lifespan startup before accepting traffic — so this is latent rather than live. Still worth closing before more callers depend on `/ready`. Suggested fix: skip the interaction-rollout segment entirely when the policy singleton is not yet initialized, rather than letting `RuntimeError` escape a readiness probe.

**`src/xagent/web/api/admin_interaction_rollout.py:64` — schema-existence check sits outside the try/except, so it bypasses the documented 503**

`interaction_requests_table_exists(db)` is called before the `try:` at line 68. That helper calls `db.connection()` / `inspect(...).has_table(...)` with no internal exception handling, so a transient DB error there propagates unhandled → 500, not the documented 503 ("a stats query that raises... returns 503, not a fabricated `active_count: 0`"). The existing test patches only `session.execute`, so this path is untested.

Suggested fix: move the `interaction_requests_table_exists(db)` call inside the same `try:` block, and add a test that makes it raise.

### informational

**`tests/web/services/test_interaction_rollout_guards.py` — the source-literal scanner is literal-only, which is an unstated limitation**

`_source_literal_call_sites` records `source=` / `task_source=` call sites only when the value is an `ast.Constant`, so it is blind to non-literal values such as `src/xagent/web/services/workforce_runs.py:300`'s `source=request.source` (which flows through `_normalize_run_source`, stripping and lowercasing but not vocabulary-checking). `workforce_runs.py` is untouched by this PR and no current caller can route an untrusted `source` through that path. Not blocking; noted so this guard isn't later read as complete enforcement.

### Minor nits

- `src/xagent/web/app.py:728-737`: four groups of function-local imports inside the `/ready` handler, while `validate_interaction_rollout_at_startup` is imported at module top in the same file — inconsistent, no functional issue.
- `src/xagent/web/services/interaction_rollout.py:254-261`: duplicate-token detection runs after lowercasing, so `XAGENT_INTERACTION_NATIVE_SOURCES="SDK,sdk"` fails startup showing the lowercased token rather than what the operator typed. Cosmetic.
- Test numbering: `tests/web/services/test_interaction_rollout_gate.py` skips from `test_tg8` to `test_tg10`; `tests/web/test_interaction_rollout_observability.py` reuses the `test_to5` and `test_to9` prefixes on two tests each. Not collisions, just inconsistency in a suite that otherwise uses sequential IDs as a traceability index.

- `src/xagent/web/services/interaction_rollout.py:152-160`: yagni: nine counter-name constants (`COUNTER_COMPAT_READ_FALLBACK`, `COUNTER_LIFECYCLE_RESPONSE_CONFLICT`, four `COUNTER_MATERIALIZE_*`, two `COUNTER_REPAIR_*`, `COUNTER_COMMAND_LAG_SECONDS_MAX`) are placeholders for unmerged future PRs and are never incremented anywhere in this codebase — drop them and let each future PR add its own counter name when it lands.
- `src/xagent/web/services/interaction_rollout.py:346-375`: shrink: `_require_policy()` and `get_interaction_rollout_policy()` are functionally identical (the public one is a one-line delegation with zero added behavior), split apart on the strength of a "call-site count tracking" mechanism that exists only in the docstring — collapse to one function.
- `src/xagent/web/models/task_interaction.py:62`: yagni: `TASK_SOURCE_LITERALS` has no production reader — it appears only in one comment explaining why it is not used, plus three test files' assertions; assert directly against literal call sites in tests instead of keeping a permanent module constant.
- `src/xagent/config.py:245-265`: shrink: `get_interaction_native_sources` repeats the comma-split/strip/skip-blank tokenizing shape already in `get_external_upload_dirs` (~line 1838) and `get_external_skills_dirs`. The functions diverge after that step, so a full merge is not feasible — but the shared ~3-line tokenizer could be one small helper all three call.

net: -15 to -20 lines possible

## Testing

All 78 new tests pass — `tests/web/services/test_interaction_rollout_config.py`, `test_interaction_rollout_gate.py`, `test_interaction_rollout_guards.py`, `test_interaction_rollout_vocabulary.py`, and `tests/web/test_interaction_rollout_observability.py` — confirmed by running the full suite in a fresh venv against this branch.

## Recommendation

Approving. No finding blocks correctness: the `/ready` and admin-endpoint gaps are latent with no active trigger path today, and the design notes (`read` mode inertness, the synthetic channel key's stability, the permanent `/health` signal) concern a gate that has no production callers yet. The one item worth fixing before or shortly after merge is the AST guard's exact set equality — left as-is it will produce false CI failures on unrelated future work and mislabel them as rollout-invariant drift.

src/xagent/web/api/admin_interaction_rollout.py
AlexLiu190625: Fixed in ceb365da2. The `interaction_requests_table_exists(db)` call now sits inside the `try`, so a failing `db.connection()` lands in the deliberate 503 branch instead of escaping as a 500; the schema-absent path still returns its 200 shape (a plain `return` inside the `try` is not an exception). Verified both directions: the new `test_to8b_connection_error_before_table_check_returns_503_not_500` fails against the pre-fix code (unhandled RuntimeError) and passes now, with the detail string checked to carry no table name. The helper itself stays unguarded on purpose — the readiness probe's call site deliberately lets it raise so a connectivity outage is not misreported as a pending migration, and that reasoning is documented there.

tests/web/services/test_interaction_rollout_guards.py
AlexLiu190625: The mechanism is exactly as you describe, and it exposed a false claim in our own docs — that part is fixed in ceb365da2 (the `/ready` docstring and PR body no longer say the segment is "skipped entirely"; the policy read runs before the mode check, and only the database query is mode-gated).

The suggested skip-when-uninitialized branch is one we considered and deliberately rejected: an uninitialized policy singleton at request time is a programming error (a process serving traffic without having run its startup validation), and the contract here is fail-fast — `get_interaction_rollout_policy()` raises rather than lazily parsing or silently skipping, so a broken startup ordering surfaces as a loud error instead of a readiness probe that quietly reports whatever mode it guessed. In production the startup event runs the validator before the server accepts traffic, so the raise is unreachable outside test harnesses that build the app without its lifespan — and those initialize explicitly. Leaving this thread open in case you want to push back on that trade-off.

src/xagent/web/app.py
AlexLiu190625: The mechanism is exactly as you describe, and it exposed a false claim in our own docs — that part is fixed in ceb365da2 (the `/ready` docstring and PR body no longer say the segment is "skipped entirely"; the policy read runs before the mode check, and only the database query is mode-gated).

The suggested skip-when-uninitialized branch is one we considered and deliberately rejected: an uninitialized policy singleton at request time is a programming error (a process serving traffic without having run its startup validation), and the contract here is fail-fast — `get_interaction_rollout_policy()` raises rather than lazily parsing or silently skipping, so a broken startup ordering surfaces as a loud error instead of a readiness probe that quietly reports whatever mode it guessed. In production the startup event runs the validator before the server accepts traffic, so the raise is unreachable outside test harnesses that build the app without its lifespan — and those initialize explicitly. Leaving this thread open in case you want to push back on that trade-off.

AlexLiu190625:
Follow-up on the review body's remaining items (the three inline threads are answered in place; fixes landed in ceb365da2):

- **`read` mode is inert** — correct, and the PR body previously implied otherwise. The "observation window in `read` mode" rollback advice is replaced with a plain statement that `read` is an accepted-but-inert value in this change; nothing branches on it until the read-path change that consumes it lands. The value stays in the accepted set so enabling it later is not an operator-visible config break.
- **`"channel"` key stability** — agreed, persisted is not stable: `Task.channel_id` is `ondelete=SET NULL`, so a task can migrate from `channel` to `internal` if its channel is deleted. `gating_key()`'s docstring now carries exactly that caveat.
- **Permanent unknown-source signal** — intended: an unrecognized source is a property of persisted data, and no in-process event can witness that data being repaired, so auto-clearing would assert a fix without evidence. The constant's comment in `ops_signals.py` documents this asymmetry with the paired schema signal.
- **Placeholder counters** — intentionally pre-declared so the namespace has one owner; each names the change that will start incrementing it, and the registry comment now says the dead state is deliberate. Wiring them up piecemeal here would create producers with no consumers.
- **`_require_policy` vs the public accessor** — these cannot merge: the call-site count on the public accessor is an executable guard, not a docstring. Collapsing them makes the gate itself a counted call site and the guard goes red (`3 <= 2`). The private helper exists precisely so the gate's internal read is not part of the audited external surface.
- **`TASK_SOURCE_LITERALS` with no production reader / vocabulary consolidation** — real, and deliberately deferred: the staging-primitive PR that just got approved hoists its own origin vocabulary into the same module, so all constant-level consolidation in that file happens in one pass right after it merges, instead of two edits that immediately rewrite each other.
- **Tokenizer duplication in `config.py`** — declined: the two call sites share ~three lines and diverge immediately (different normalization, different filtering); extracting a helper adds indirection without removing real duplication.
- **Nits** — function-local imports hoisted (except `get_session_local`, which tests monkeypatch on the module and a top-level binding would bypass — commented in place); the duplicate-source error now quotes the raw operator input; duplicate test names renamed. The literal-only limitation of the source scanner stands as documented in its docstring.

Issue #1250, opened 2026-08-10
Post-merge cleanups from the connector team-visibility review


AlexLiu190625:
Three maintenance items acknowledged during the review of #1241, deliberately kept out of that PR to avoid widening a converged diff. None affects production behavior.

**1. Shrink or revive the test-only branch in `chat.py`.**

The branch around `chat.py:1936` (the non-snapshot fallback in the tool-config construction path) is documented as unreachable from production and verified to be so, but it is exercised directly by `tests/web/test_agent_manager_reconstruction.py` and `tests/web/services/test_task_setup_snapshot.py`, so it cannot simply be deleted. It has been accreting comments and logic that only tests observe. Either shrink it to the minimum those tests need, or promote it back into a real production path — a deliberate decision, with the affected test fixtures reworked accordingly.

**2. Replace the vacuous mock chains in `tests/web/api/test_tools_api.py`.**

Two tests (around lines 1289 and 1336) mock the MCP visibility query as `query().filter().order_by().all()` without asserting which predicate reached `filter()` — they would pass even if the visibility clause were removed. The predicate itself is covered by DB-backed suites (including a compiled-SQL shape test), so this is vacuity rather than a coverage gap. When these tests are next touched, convert them to the DB-backed path or assert on the compiled predicate.

**3. Sweep review-shorthand docstrings out of pre-existing test files.**

Files not touched by #1241 still carry docstrings written in code-review shorthand that is meaningless without access to the review that produced it (e.g. `tests/web/test_react_*.py`, `tests/web/tools/test_aws_mcp.py:937` "PR review finding R3-m2") — roughly 43 lines match. Rewrite them as plain statements of the behavior each test pins, same treatment #1241 applied to the files it touched. A pattern of the form `grep -rnE "\b[IJRS][0-9]+[a-z']?\b|plan §|[Rr]eview finding" tests/` locates them; acceptance is that the matches remaining are ordinary English, not review artifacts.

AlexLiu190625:
Item 4 (found while consolidating the team-scope failure wrap): the custom-API live loader's guarded region degrades a genuine DB failure into an empty tool set, where the MCP twin raises `MCPConfigLoadError`. Pre-existing asymmetry, newly conspicuous now that everything else about the two paths is parallel. Worth deciding deliberately: adopt the MCP twin's typed raise, or document the divergence.

PR #1241, opened 2026-08-10
Resolve MCP connector visibility from the governing agent's team

AlexLiu190625:
## Summary

Connector visibility at tool-build time is resolved from the governing agent's owning team,
through an optional application hook, instead of from the running user's personal association
rows only. Standalone deployments and deployments that never install a team hook are unchanged.

Today, an MCP server that a team member has shared with their team is only usable by agents
whose selection resolves through the *running user's own* personal links. A published team
agent that another team member runs cannot see a server that a teammate shared, even though the
server is meant to be team-wide. This change fixes that by keying visibility on the agent's team
instead of the runner's personal association, while keeping every existing deployment's behavior
identical unless it opts in by installing the new hook.

Two runtime read points change:

- **`WebToolConfig._load_mcp_server_configs`** (the MCP tool loader) now resolves
  `personal ∪ the governing agent's team` instead of `personal only`, dropping the inner join on
  the personal-ownership table in favor of a predicate that matches personal rows or team-owned
  rows.
- **`_load_visible_runtime_connectors`** (used by the connector-runtime-context loader) now
  resolves the same union when a team-keyed hook is installed, replacing its previous
  hand-rolled user-keyed lookup for that case.

The team hook is a `Protocol` with a keyword-only parameter
(`def __call__(self, db, *, team_id: int) -> dict[str, set[int]]`), installed through
`connector_team_scope.set_connector_team_hooks(team_visibility=...)`. It is invoked by keyword
everywhere it is called, so a hook accidentally bound into the wrong slot (there is a
similarly-shaped, pre-existing user-keyed hook in the same module) raises a `TypeError` instead
of silently resolving an unrelated team's connectors.

## Behavior and compatibility

- **No hook installed (standalone or any deployment that hasn't adopted this):** both read
  points return the same row set as before. The MCP tool loader's row order becomes
  deterministic where it was previously database-planner-dependent, because this change adds
  `ORDER BY id` where there was none before — a real, if small, user-visible-in-principle
  change, stated here rather than folded into "identical."
- **A deployment that adopts this revision without installing the new team hook (only the
  pre-existing legacy user-keyed hook, or no hook at all):** the MCP tool loader is unchanged
  (personal-only; it never had a team overlay, and none is added when the new hook is absent —
  adding one there would newly expose the runner's own team's servers, which is not what this
  change does). The connector-runtime-context loader falls back to its existing user-keyed
  overlay whenever no team-keyed hook is present, so its answer is exactly what it returned
  before this change, in that state.
[...]

PR #1264, opened 2026-08-11
Resolve custom API connector visibility from the governing agent's team

AlexLiu190625:
## Summary

Custom API tool loading resolved visibility by iterating a user's personal
junction rows only, so a custom API shared with a team never loaded for a
team-governed agent, even though the equivalent MCP path already resolved
team-shared servers.

This change applies the same rule to custom APIs: a run's visible custom
APIs are the run owner's own active personal links, unioned with whatever
the governing agent's owning team holds, through the same optional
application hook the MCP path already uses.

- Both of `WebToolConfig`'s custom-API read points -- the live call used
  during normal tool building, and the off-loop snapshot prefetch used by
  the tool factory -- now query through one shared, module-level query
  builder instead of iterating personal junction rows. A deterministic
  `ORDER BY` keeps config order reproducible now that the underlying query
  shape changed.
- A team-scope resolution failure raises the same typed error -- same
  code, same message, same failure reason -- that the MCP loader already
  raises, so a broken team hook fails the turn instead of silently
  dropping the custom-API tool set.
- The runtime-context view's custom-API half, previously narrowed to
  personal-only because the tool loaders could not yet back it up, is
  widened back to match now that both sides resolve team ownership
  together. Selecting a team-owned custom API into a task's runtime
  snapshot always has a tool loader able to build it now.
- The team-scope resolution wrap above -- resolve the team's connector ids,
  re-raise a typed error unchanged, convert anything else into the same
  503 -- existed as two near-identical copies before this change (the MCP
  read point and the runtime-context view loader; the custom-API read
  points did no team resolution at all before this PR, since they were
  still junction-only). This PR's own two new custom-API call sites would
  otherwise have made four copies; instead, all four now share one
  function. Three call sites reduce to naming the connector kind they need;
  the runtime-context view loader consumes the whole answer. No behavior
  changed; verified by running every existing test that pins any of the
  four sites' failure behavior, unchanged before and after.

## Behavior and compatibility

| Deployment state | Custom API visibility |
|---|---|
| No team-scope hook installed | Unchanged: personal active links only, same result set as before this change |
| Team-scope hook installed, agent has no governing team | Personal-only, unchanged |
| Team-scope hook installed, agent has a governing team | Personal active links, unioned with the team's custom APIs -- now matches the existing MCP behavior |
| Only the legacy user-keyed visibility hook installed (no team-keyed hook) | Unchanged fallback behavior for both connector kinds |

A custom API shared with a team executes using the shared definition row's
own stored credentials -- there is no per-member credential layer for
custom APIs to lose, unlike MCP's per-member environment overrides.

A genuine database failure while loading custom APIs now fails the turn,
matching the existing MCP behavior, instead of silently producing no
custom-API tools for that turn.

## Measured evidence

Each row below was verified by reverting the specific change as a literal
source edit, confirming the listed tests go red, then restoring the change
and confirming green again.

| Change reverted | Tests that catch it |
|---|---|
| Restore the personal-junction-row loop in the live read point | the team-agent-loads-its-team's-custom-API test; the snapshot/live agreement test (team parameter set); the runtime-view/tool-loader parity test (team cells) |
| Restore the personal-junction-row loop in the snapshot prefetch | the snapshot/live agreement test (team parameter set) |
| Fall back to the legacy user-keyed hook when the governing team id is unset | the personal-agent test's legacy-hook variant |
| Key the lookup on the runner's own team instead of the governing agent's team | the runtime-view/tool-loader parity test (team cells); the personal-agent test; the snapshot/live agreement test; the hook-failure test |
| Drop the governing-team id from the prefetch's detached load plan | the snapshot/live agreement test; the hook-failure test's snapshot twin |
| Fold the team-scope resolution into the same guarded region as the query, dropping its own error wrap | the hook-failure test |
| Remove the typed-error wrap entirely, letting a raw hook exception propagate | the hook-failure test (both its live and tool-creator assertions) |
| Drop the zero-row short-circuit ahead of the runtime-view resolution | the dedicated zero-row test |
| Point the deterministic ordering clause at the wrong column | the compiled-query-shape test's exact ordering assertion (same fix applied to the pre-existing MCP sibling test) |
| Reintroduce the narrowing that discarded the runtime-view's custom-API half | the runtime-view union test; the runtime-view/tool-loader parity test (team cells) |
| Drop the `None`-owner guard on the visibility predicate | the predicate's own dedicated test |
| Revert the query-failure fail-closed change back to the swallow-to-empty-list behavior | the dedicated DB-failure test |

## Test totals

- The six files this change touches directly (`WebToolConfig` custom-API
  loading, the visibility-clause seam and its dedicated tests, the parity
  check, and the two rewritten regression tests): 172 passed, 0 failed.
- `tests/web/tools/` (the whole directory): 529 passed, 0 failed.
- `tests/web/services/` (the whole directory): 1009 passed, 79 skipped
  (skips are pre-existing environment gates -- a Postgres test URL not
  being set locally -- unrelated to this change), 0 failed.

Pre-commit (formatting, linting, type checking) passes on every changed
file.

### DIFF
--- a/src/xagent/web/services/connector_runtime.py
+++ b/src/xagent/web/services/connector_runtime.py
@@ -276,17 +276,15 @@ def load_connector_runtime_view(
     for ref in selected_refs:
         connector = visible.get(ref)
         if connector is None:
-            # Tool loading applies the same visibility filter before
-            # instantiating a tool, so a now-hidden historical ref has no
-            # runtime tool to receive values on this turn. This holds for
-            # MCP on both the new-hook and legacy-hook branches, and for
-            # Custom API on the new-hook branch too (narrowed to the
-            # junction-only set until a follow-up team-keys the Custom API
-            # tool loaders and this branch's Custom API set to match). On
-            # the legacy-hook branch the filters diverge for Custom API: a
-            # legacy user-keyed hook can grant a team-shared Custom API
-            # here while the Custom API tool loader resolves junction-only,
-            # so such a ref stays visible without a runtime tool.
+            # Tool loading applies the same visibility filter when the
+            # team-scope hook is installed, for both connector kinds, so a
+            # ref absent here has no runtime tool either way. With only the
+            # legacy hook installed, this view is user-keyed and unions a
+            # legacy-granted team-shared custom API, while both tool
+            # loaders resolve personal-only -- they consult the team-keyed
+            # hook exclusively, never the legacy one -- so such a ref is
+            # selectable and persistable into a task's connector selection
+            # but builds no runtime tool.
             continue
         raw_ephemeral = (
             ephemeral_by_ref.get(ref.storage_key, {})
@@ -403,11 +401,11 @@ def prepare_connector_runtime_selection_snapshot(
     the agent's tool-selection policy, while ``connector_user_id`` supplies the
     same connector visibility scope used by normal web tool loading:
     ``connector_user_id``'s personal MCP/Custom API links, unioned with
-    ``agent``'s owning team's connectors for MCP when a team-keyed hook is
-    installed (Custom API stays junction-only for now). For published-agent
-    chats, personal visibility still follows the task owner rather than the
-    published agent's owner; team visibility follows the agent regardless of
-    who is running it.
+    ``agent``'s owning team's connectors for both connector kinds when a
+    team-keyed hook is installed. For published-agent chats, personal
+    visibility still follows the task owner rather than the published
+    agent's owner; team visibility follows the agent regardless of who is
+    running it.
     """
 
     if agent is None or connector_user_id is None:
@@ -593,8 +591,8 @@ def _load_visible_runtime_connectors(
     db: Session, *, user_id: int, agent_team_id: int | None = None
 ) -> dict[ConnectorRef, Any]:
     from .connector_team_scope import (
+        resolve_team_connector_ids_or_raise,
         team_connector_hook_installed,
-        team_connector_ids,
         visible_team_connector_ids,
     )
 
@@ -602,31 +600,17 @@ def _load_visible_runtime_connectors(
     if team_connector_hook_installed():
         # Team ownership is resolved from the team that owns the governing
         # agent. A run with no governing agent resolves personal links only.
-        try:
-            team_ids = team_connector_ids(db, team_id=agent_team_id)
-        except ConnectorRuntimeError:
-            raise
-        except Exception as exc:
-            logger.warning(
-                "Failed to resolve team connector scope for user %s",
-                user_id,
-                exc_info=True,
-            )
-            raise ConnectorRuntimeError(
-                ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
-                "Connector team scope is unavailable.",
-                details={"reason": "team_scope_resolution_failed"},
-                status_code=503,
-            ) from exc
-        # Custom API is not team-keyed on either side of this seam yet: the
-        # tool-build loaders (WebToolConfig's custom-API paths) stay
-        # junction-only until a follow-up team-keys them too. Consuming the
-        # "custom_api" half here without a matching consumer would let a
-        # team-owned custom API be selected into a task's runtime snapshot
-        # and then fail at runtime-view resolution (no personal link to
-        # satisfy it) while never building a tool either -- so this side
-        # stays MCP-only until both halves move together.
-        team_ids = {"mcp": team_ids["mcp"], "custom_api": set()}
+        team_ids = resolve_team_connector_ids_or_raise(
+            db, team_id=agent_team_id, log_subject=user_id
+        )
+        # Both connector kinds are team-keyed together: the tool-build
+        # loaders (WebToolConfig's MCP and custom-API paths) both consult
+        # the same team scope now, so a team-owned custom API selected into
+        # a task's runtime snapshot here always has a matching tool-loader
+        # consumer to build it. Consuming the "custom_api" half without a
+        # matching consumer would let a selected connector fail at
+        # runtime-view resolution while never building a tool either --
+        # that hazard is why the two sides move together, not separately.
     else:
         # No team-keyed hook: keep the legacy user-keyed overlay exactly as
         # it is, so an installation that adopts this revision without
--- a/src/xagent/web/services/connector_team_scope.py
+++ b/src/xagent/web/services/connector_team_scope.py
@@ -7,11 +7,19 @@ the application's team tables.
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Callable, Collection
 from dataclasses import dataclass
 from typing import Any, Literal, Protocol
 
 from sqlalchemy.sql.elements import ColumnElement
+
+from ...core.tools.adapters.vibe.connector_runtime import (
+    ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
+    ConnectorRuntimeError,
+)
+
+logger = logging.getLogger(__name__)
 
 ConnectorType = Literal["mcp", "custom_api"]
 ConnectorRenamedHook = Callable[[Any, int, ConnectorType, int, str, str], None]
@@ -61,10 +69,16 @@ class TeamConnectorVisibilityHook(Protocol):
 
     Every connector id a hook returns becomes executable by *any* runner of
     that team's agents, not only members: for stdio and other static-auth
-    transports, a team-owned server with no personal link for the running
-    user executes using the shared definition row's own stored credentials,
-    not the runner's. (OAuth transports fail closed instead, for lack of a
-    token to resolve.)
+    MCP transports, a team-owned server with no personal link for the
+    running user executes using the shared definition row's own stored
+    credentials, not the runner's. (OAuth transports fail closed instead,
+    for lack of a token to resolve.) Custom APIs are starker: there is no
+    per-member credential override layer at all, and no transport-level
+    exception either -- every custom API always executes with whatever is
+    stored on its shared definition row (headers, a static API key, and so
+    on), for every runner, member or not. Sharing a custom API with a team
+    means sharing whatever credential it holds with everyone who can run
+    the team's agents.
     """
 
     def __call__(self, db: Any, *, team_id: int) -> dict[str, set[int]]: ...
@@ -176,6 +190,50 @@ def team_connector_hook_installed() -> bool:
     return _team_connector_visibility_hook is not None
 
 
+def resolve_team_connector_ids_or_raise(
+    db: Any, *, team_id: int | None, log_subject: int | None
+) -> dict[str, set[int]]:
+    """``team_connector_ids(db, team_id=team_id)``, with every non-typed
+    failure converted into the seam's one typed 503.
+
+    Every team-scope resolution call site -- both of ``WebToolConfig``'s
+    custom-API read points, its MCP read point, and the runtime-context
+    view loader -- wraps ``team_connector_ids`` the same way: a
+    ``ConnectorRuntimeError`` passes through unchanged, and any other
+    exception is logged at ``WARNING`` and converted into
+    ``ConnectorRuntimeError(ERROR_CONNECTOR_RUNTIME_UNAVAILABLE, "Connector
+    team scope is unavailable.", details={"reason":
+    "team_scope_resolution_failed"}, status_code=503)``. This function is
+    that wrap, written once. Returns the full ``{"mcp": ..., "custom_api":
+    ...}`` dict unmodified -- callers that need one connector kind extract
+    it themselves (e.g. ``result["mcp"]``); the runtime-context view loader
+    needs both and takes the dict as-is.
+
+    ``log_subject`` is the user id that identifies the caller in the
+    warning log line -- ``None`` at any call site whose own identity guard
+    the caller has not yet reached (the MCP read point's private loader has
+    no identity guard of its own; production only reaches it through its
+    guarded public wrapper). It is only ever formatted into the log
+    message, never interpreted.
+    """
+    try:
+        return team_connector_ids(db, team_id=team_id)
+    except ConnectorRuntimeError:
+        raise
+    except Exception as exc:
+        logger.warning(
+            "Failed to resolve team connector scope for user %s",
+            log_subject,
+            exc_info=True,
+        )
+        raise ConnectorRuntimeError(
+            ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
+            "Connector team scope is unavailable.",
+            details={"reason": "team_scope_resolution_failed"},
+            status_code=503,
+        ) from exc
+
+
 def visible_mcp_server_clause(
     owner_user_id: int | None, team_mcp_ids: Collection[int]
 ) -> ColumnElement[bool]:
@@ -207,6 +265,39 @@ def visible_mcp_server_clause(
     if not team_mcp_ids or owner_user_id is None:
         return personal
     return or_(personal, MCPServer.id.in_(set(team_mcp_ids)))
+
+
+def visible_custom_api_clause(
+    owner_user_id: int | None, team_api_ids: Collection[int]
+) -> ColumnElement[bool]:
+    """Predicate selecting the custom APIs visible to ``owner_user_id``.
+
+    Personal custom APIs (an active ``UserCustomApi`` link) unioned with the
+    team-owned APIs named in ``team_api_ids``. ``is_active`` gates only the
+    personal arm -- a team-owned API is visible regardless of the member's
+    own personal link state, which is intentional (see the module's callers
+    for the credential consequence). Pure function over ORM constructs -- no
+    I/O. An empty ``team_api_ids`` reduces to exactly the personal
+    semi-join, so a deployment with no team overlay compiles the same query
+    shape it always has. A ``None`` owner also reduces to the personal arm
+    regardless of ``team_api_ids`` -- defense in depth so this function is
+    fail-closed on its own terms, not only because of how its callers
+    happen to be gated today.
+    """
+
+    from sqlalchemy import or_, select
+
+    from ..models.custom_api import CustomApi, UserCustomApi
+
+    personal = CustomApi.id.in_(
+        select(UserCustomApi.custom_api_id).where(
+            UserCustomApi.user_id == owner_user_id,
+            UserCustomApi.is_active,
+        )
+    )
+    if not team_api_ids or owner_user_id is None:
+        return personal
+    return or_(personal, CustomApi.id.in_(set(team_api_ids)))
 
 
 def delete_team_connector(
--- a/src/xagent/web/tools/config.py
+++ b/src/xagent/web/tools/config.py
@@ -728,6 +728,28 @@ def _load_custom_api_runtime_view_sync(
         ) from exc
 
 
+def _visible_custom_api_query(
+    db: Any, *, owner_user_id: int | None, team_api_ids: frozenset[int]
+) -> Any:
+    """Production custom-API visibility query, shared by both read points.
+
+    Module-level, not a ``WebToolConfig`` method: ``_load_custom_api_factory_inputs``
+    below is a module-level function that never sees a ``WebToolConfig``
+    instance (the detached-plan contract documented on
+    ``_load_tool_factory_runtime_snapshot``'s docstring, this module: "never
+    receives the caller's ``WebToolConfig``, request Session, or ORM user"),
+    so a method could serve only one of the two read points.
+    """
+    from ..models.custom_api import CustomApi
+    from ..services.connector_team_scope import visible_custom_api_clause
+
+    return (
+        db.query(CustomApi)
+        .filter(visible_custom_api_clause(owner_user_id, team_api_ids))
+        .order_by(CustomApi.id)
+    )
+
+
 def _load_custom_api_factory_inputs(
     db: Any,
     *,
@@ -739,17 +761,23 @@ def _load_custom_api_factory_inputs(
     if user_id is None:
         return []
 
-    from ..models.custom_api import UserCustomApi
+    # Resolved before the caller's guarded region (this helper carries no
+    # try/except of its own): that region reports "every selected API is
+    # unavailable", which is the wrong answer for "the scope could not be
+    # resolved". The typed error is what survives the tool-creator frame --
+    # an untyped one is dropped there with an ERROR and no tool set at all.
+    from ..services.connector_team_scope import resolve_team_connector_ids_or_raise
 
-    user_apis = (
-        db.query(UserCustomApi)
-        .filter(
-            UserCustomApi.user_id == user_id,
-            UserCustomApi.is_active,
-        )
-        .all()
+    team_api_ids = frozenset(
+        resolve_team_connector_ids_or_raise(
+            db, team_id=connector_team_id, log_subject=user_id
+        )["custom_api"]
     )
-    if not user_apis:
+
+    rows = _visible_custom_api_query(
+        db, owner_user_id=user_id, team_api_ids=team_api_ids
+    ).all()
+    if not rows:
         return []
 
     runtime_view = _load_custom_api_runtime_view_sync(
@@ -760,10 +788,7 @@ def _load_custom_api_factory_inputs(
         agent_team_id=connector_team_id,
     )
     configs: list[dict[str, Any]] = []
-    for user_api in user_apis:
-        api = user_api.custom_api
-        if api is None:
-            continue
+    for api in rows:
         runtime_values = runtime_view.get(f"custom_api:{int(api.id)}")
         configs.append(
             _custom_api_config_from_model(
@@ -3391,26 +3416,13 @@ class WebToolConfig(BaseToolConfig):
         # for "the scope could not be resolved". The typed error is what
         # survives the tool-creator frame -- an untyped one is dropped there
         # with a WARNING and no tool set at all.
-        try:
-            from ..services.connector_team_scope import team_connector_ids
+        from ..services.connector_team_scope import resolve_team_connector_ids_or_raise
 
-            team_mcp_ids = frozenset(
-                team_connector_ids(self.db, team_id=self._connector_team_id)["mcp"]
-            )
-        except ConnectorRuntimeError:
-            raise
-        except Exception as exc:
-            logger.warning(
-                "Failed to resolve team connector scope for user %s",
-                self._user_id,
-                exc_info=True,
-            )
-            raise ConnectorRuntimeError(
-                ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
-                "Connector team scope is unavailable.",
-                details={"reason": "team_scope_resolution_failed"},
-                status_code=503,
-            ) from exc
+        team_mcp_ids = frozenset(
+            resolve_team_connector_ids_or_raise(
+                self.db, team_id=self._connector_team_id, log_subject=self._user_id
+            )["mcp"]
+        )
 
         try:
             from ..services.mcp_runtime import (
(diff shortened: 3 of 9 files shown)


rogercloud (commented):
## Summary

Custom-API tool loading resolved visibility only from a user's personal `UserCustomApi` junction rows, so a custom API shared with a team never loaded for a team-governed agent — while the parallel MCP path already resolved team-shared servers through the optional `team_visibility` hook. This PR closes that asymmetry: it adds `visible_custom_api_clause` alongside the existing `visible_mcp_server_clause`, routes both of `WebToolConfig`'s custom-API read points (the live call and the off-loop snapshot prefetch) through a single module-level query builder `_visible_custom_api_query` with a deterministic `ORDER BY custom_apis.id`, raises the same typed `ConnectorRuntimeError` (503, `team_scope_resolution_failed`) on team-scope resolution failure that the MCP path already raises, and re-widens the runtime-context view's custom-API half by removing the narrowing that zeroed out the team grant.

Scope: `src/xagent/web/services/connector_runtime.py`, `src/xagent/web/services/connector_team_scope.py`, `src/xagent/web/tools/config.py`, plus 6 test files including a new 549-line `tests/web/tools/test_custom_api_team_visibility.py`.

## Approach verdict: sound

This is the minimal symmetric fix, and — importantly — the sequencing is right. Widening the loaders and the runtime-context view in the *same* change is not incidental tidiness: the removed narrowing line existed precisely to keep the selection seam and the execution seam consistent, so fixing only one side would have reintroduced the "selectable but never executable" hazard the narrowing was guarding against. `visible_custom_api_clause` is structurally identical to `visible_mcp_server_clause` modulo model/column names; no asymmetry found. The zero-row guard, the fail-closed `None`-owner reduction, and the "`is_active` vetoes only the personal arm" semantics all match the MCP path. Efficiency is strictly improved — one filtered `SELECT` replaces per-row lazy attribute access — and the `OR` union carries no duplicate-row risk since it is a single-table filter, not a join.

## Findings

No correctness bugs were found in the changed lines. All findings below are **minor**; there are no critical or major issues.

**1. Minor — docstring cites the wrong lines for the detached-plan contract.**
`src/xagent/web/tools/config.py:737` — the new `_visible_custom_api_query` docstring justifies being module-level by citing "the detached-plan contract at :809-813", but those lines are the tail of an unrelated loop plus the `def` line of `_custom_api_config_from_model`. The actual "never receives the caller's `WebToolConfig`" contract lives in `_load_tool_factory_runtime_snapshot`'s docstring at `config.py:856-860`. The reasoning is correct; only the pointer is wrong.

**2. Minor — `visible_custom_api_clause` docstring drops a load-bearing sentence from its MCP twin.**
`src/xagent/web/services/connector_team_scope.py:218` — the new docstring is otherwise a faithful copy of `visible_mcp_server_clause`'s (lines 182-195) but omits: "`is_active` gates only the personal arm -- a team-owned server is visible regardless of the member's own personal link state, which is intentional." The custom-API code genuinely implements that same semantic (the team arm has no `is_active` check), and it is pinned by a test (the `inactive_own_shared` fixture, asserted visible). So this is an accidental gap in an otherwise-complete copy, not a deliberate difference, and it hides a non-obvious authorization semantic.

**3. Minor — `TeamConnectorVisibilityHook`'s contract docstring is still MCP-only.**
`src/xagent/web/services/connector_team_scope.py:36-68` (not touched by this PR, so noted here rather than inline). The hook's credential-exposure paragraph is worded exclusively in MCP-transport terms ("stdio and other static-auth transports", "OAuth transports fail closed"). After this PR the same hook directly governs custom-API team visibility, and the custom-API story is starker: there is no per-member credential override layer at all, not even the partial protection MCP's per-member env overrides provide. The PR body states this plainly; the code-level contract should too, so an integrator reading only the docstring does not underestimate the exposure of sharing a custom API that holds a static production API key.

**4. Minor — `install_team_hooks` docstring overstates the shared teardown.**
`tests/web/tools/test_mcp_team_visibility.py:131` — the docstring says "`_reset_hooks` (this module, autouse) undoes both after each test regardless of which suite installed them", but `tests/web/tools/test_custom_api_team_visibility.py:90-94` declares its own duplicated `_reset_hooks` fixture, and that file's header comment (lines 45-50) explains why it was not imported. Each module owns its own teardown copy; the wording reads as one shared mechanism. Cosmetic, no behavioral impact.

**5. Minor — no "no identity + live team grant" test for either custom-API read point.**
`tests/web/tools/test_custom_api_team_visibility.py:378` — the MCP suite has two tests (around lines 376 and 409 of `test_mcp_team_visibility.py`) pinning "no identity ⇒ no tools even when `connector_team_id` names a live team grant". The custom-API suite has no equivalent for either `WebToolConfig.get_custom_api_configs()` (guard at `config.py:3497`) or `_load_custom_api_factory_inputs` (guard at `config.py:759`). This is *not* a security gap — `visible_custom_api_clause` is independently proven fail-closed for a `None` owner by `test_visible_custom_api_clause_matches_nothing_for_none_owner_even_with_team_ids` — but it is an error-shape gap: if either guard were moved below the team-scope resolution call, no test in the six touched files would catch that an unauthenticated build with a `team_id` now raises 503 instead of returning `[]`.

**6. Minor — query-shape test does not pin the ordering column.**
`tests/web/tools/test_custom_api_team_visibility.py:374` — `assert "ORDER BY" in norm` only proves *some* ordering exists. Re-pointing production's `.order_by(CustomApi.id)` at any other column would keep this test green while breaking the "config order matches `custom_apis.id`" property the PR body claims, and a comment later in the same file (lines 497-505) leans on this test as the sole guarantor of that contract. Note this weakness is identical in the pre-existing MCP sibling (`test_mcp_team_visibility.py:250-261`) and was not introduced here — a consistent pattern being extended, not a regression — so it is a light suggestion, not a blocker.

`src/xagent/web/tools/config.py:751`: shrink the duplicated team-scope `try/except` block. This PR adds two near-byte-identical copies (`config.py:751-786` and `config.py:3492-3524`) of a pattern that already existed twice (`config.py:3418-3447`, `connector_runtime.py:586-618`), taking the count from 2 to 4. All four are the same shape: resolve `team_connector_ids(db, team_id=...)[kind]`, re-raise `ConnectorRuntimeError`, and convert any other exception into the same 503 `team_scope_resolution_failed` error after the same `logger.warning`. A shared `_resolve_team_connector_ids(db, *, team_id, kind, log_user_id) -> frozenset[int]` would collapse each site to one line with no behavior change — nothing about the session or types blocks it.

`tests/web/tools/test_custom_api_team_visibility.py:374`: shrink the two-step ordering check to one exact assertion. `assert norm.endswith("ORDER BY CUSTOM_APIS.ID")` replaces the weak substring test and actually pins the contract.

net: -50 lines possible (roughly 15 lines × 4 sites collapsed into one ~18-line helper).

## Test results

Verified locally, not taken from the PR description:
- All 6 touched test files: **167 passed, 0 failed**.
- Full `tests/web/tools/`: **525 passed, 0 failed** (re-run 4 times for stability after one non-reproducing flake under concurrent load).

Both totals match the author's claimed numbers exactly.

## Verdict

No correctness bugs in the changed lines. This is a clean, well-tested change on a sound design — the fix is minimal, symmetric with the established MCP path, and correctly sequences the two seams so neither is left inconsistent. Every finding above is a docstring, test-strength, or duplication cleanup; none of them blocks merge.

src/xagent/web/tools/config.py
```diff
@@ -728,6 +728,26 @@ def _load_custom_api_runtime_view_sync(
         ) from exc
 
 
+def _visible_custom_api_query(
+    db: Any, *, owner_user_id: int | None, team_api_ids: frozenset[int]
+) -> Any:
+    """Production custom-API visibility query, shared by both read points.
+
+    Module-level, not a ``WebToolConfig`` method: ``_load_custom_api_factory_inputs``
+    below is a module-level function that never sees a ``WebToolConfig``
```
rogercloud, line 737: Minor: this citation points at the wrong lines. `:809-813` is the tail of an unrelated loop plus the `def` line of `_custom_api_config_from_model` -- no contract description there. The "never receives the caller's `WebToolConfig`" contract this reasoning depends on is documented in `_load_tool_factory_runtime_snapshot`'s docstring at `src/xagent/web/tools/config.py:856-860`. The justification for being module-level is correct; only the pointer is stale. Suggest citing `:856-860` instead.

src/xagent/web/services/connector_team_scope.py
```diff
@@ -209,6 +209,36 @@ def visible_mcp_server_clause(
     return or_(personal, MCPServer.id.in_(set(team_mcp_ids)))
 
 
+def visible_custom_api_clause(
+    owner_user_id: int | None, team_api_ids: Collection[int]
+) -> ColumnElement[bool]:
+    """Predicate selecting the custom APIs visible to ``owner_user_id``.
+
+    Personal custom APIs (an active ``UserCustomApi`` link) unioned with the
+    team-owned APIs named in ``team_api_ids``. Pure function over ORM
```
rogercloud: Minor: this docstring is an otherwise-faithful copy of `visible_mcp_server_clause`'s (lines 182-195) but drops one load-bearing sentence:

> `is_active` gates only the personal arm -- a team-owned server is visible regardless of the member's own personal link state, which is intentional.

The custom-API code implements exactly that same semantic (the team arm below has no `is_active` check), and it is pinned by a test -- the `inactive_own_shared` fixture is asserted visible. So the omission looks accidental rather than a deliberate difference, and it hides a non-obvious authorization property: an intentionally deactivated personal link does *not* revoke access to a team-owned API. Suggest mirroring the MCP wording.

tests/web/tools/test_mcp_team_visibility.py
```diff
@@ -122,6 +122,32 @@ def seed(db_session: Session):
     )
 
 
+def install_team_hooks(*, team_visibility, agent_owner_id: int) -> None:
+    """Install a ``team_visibility`` hook *and* the agent-team-scope hook
+    mapping ``agent_owner_id`` -> T1.
+
+    Shared by the MCP and custom-API team-visibility suites so the
+    hook-install/teardown pattern is written once; ``_reset_hooks`` (this
+    module, autouse) undoes both after each test regardless of which suite
```
rogercloud: Minor (cosmetic): "`_reset_hooks` (this module, autouse) undoes both after each test regardless of which suite installed them" reads as one shared teardown mechanism covering both suites, but `tests/web/tools/test_custom_api_team_visibility.py:90-94` declares its own duplicated `_reset_hooks` fixture, and that file's header comment (lines 45-50) explains why it was not imported. Each module owns a separate copy. Suggest rewording to "each suite's own autouse `_reset_hooks` undoes both after every test" so the docstring does not imply cross-module coupling that isn't there. No behavioral impact.

tests/web/tools/test_custom_api_team_visibility.py
```diff
@@ -0,0 +1,549 @@
...
+    sql = str(
+        _visible_custom_api_query(
+            db_session, owner_user_id=int(seed.c.id), team_api_ids=team_api_ids
+        ).statement.compile(dialect=postgresql.dialect())
+    )
+    norm = " ".join(sql.split()).upper()
+    assert norm.startswith("SELECT")  # positive control
+    assert re.search(r"\bFROM CUSTOM_APIS\b", norm)  # driving table
+    assert "JOIN USER_CUSTOM_APIS" not in norm  # misuse guard
+    assert "ORDER BY" in norm
```
rogercloud: Minor: `assert "ORDER BY" in norm` only proves *some* ordering clause exists, not which column. Re-pointing production's `.order_by(CustomApi.id)` at any other column would leave this test green while breaking the "deterministic config order matches `custom_apis.id`" property the PR body claims -- and the comment at lines 497-505 in this file treats this test as the sole guarantor of that contract, which it currently isn't.

Suggest an exact assertion:
```python
assert norm.endswith("ORDER BY CUSTOM_APIS.ID")
```

For the record, the pre-existing MCP sibling `test_production_mcp_query_shape` (`tests/web/tools/test_mcp_team_visibility.py:250-261`) has the identical weakness and isn't touched here, so this is a consistent pattern being extended rather than a regression -- a light suggestion, not a blocker.

tests/web/tools/test_custom_api_team_visibility.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+# ---------------------------------------------------------------------------
+# visible_custom_api_clause is fail-closed on its own terms, not only
```
rogercloud, line 407: Minor coverage gap: the MCP suite pins "no identity => no tools even when `connector_team_id` names a live team grant" for its loader (see `tests/web/tools/test_mcp_team_visibility.py` around lines 376 and 409). There's no equivalent here for either custom-API read point -- `WebToolConfig.get_custom_api_configs()` with `user_id=None` plus a `connector_team_id` that has a real grant (guard at `src/xagent/web/tools/config.py:3497`), or `_load_custom_api_factory_inputs(user_id=None, connector_team_id=T1)` (guard at `src/xagent/web/tools/config.py:759`).

This is not a security gap: `test_visible_custom_api_clause_matches_nothing_for_none_owner_even_with_team_ids` below independently proves the clause is fail-closed for a `None` owner. But it is an *error-shape* gap -- if either loader guard were moved below the team-scope resolution call, nothing in the six touched test files would catch that an unauthenticated build with a `team_id` starts raising a 503 instead of returning `[]`. Two small tests asserting `== []` (and no raise) for each read point would close it.

src/xagent/web/tools/config.py
AlexLiu190625: Fixed at 601cf49f — the docstring now cites `_load_tool_factory_runtime_snapshot` by function name and quotes its contract sentence directly, so the pointer cannot go stale the way a raw line range does.

src/xagent/web/services/connector_team_scope.py
AlexLiu190625: Fixed at 601cf49f — the dropped sentence is restored, mirroring the MCP twin's wording: `is_active` gates only the personal arm; a team-owned API stays visible regardless of the member's own personal link state, intentionally. The omission was exactly what it looked like — an accidental gap in the copy, not a semantic difference.

tests/web/tools/test_mcp_team_visibility.py
AlexLiu190625: Fixed at 601cf49f — reworded to state that the two suites share only the install pattern and each keeps its own autouse `_reset_hooks`; no cross-module fixture coupling implied.

tests/web/tools/test_custom_api_team_visibility.py
AlexLiu190625: Adopted at 601cf49f, at both sites: `test_production_custom_api_query_shape` now asserts `norm.endswith("ORDER BY CUSTOM_APIS.ID")`, and the pre-existing MCP sibling `test_production_mcp_query_shape` got the same strengthening (same pattern, file already in this diff). Both were mutation-verified: re-pointing production's `order_by` at the name column turns each red with the bad SQL in the failure output, and both return green on restore.

tests/web/tools/test_custom_api_team_visibility.py
AlexLiu190625: Both tests added at 601cf49f (`user_id=None` + a live team grant → `[]`, no raise, for each read point) — with one honest correction from actually running the drill you described: moving either guard below the team-scope resolution does **not** produce the 503. Measured: on the live path the resulting `TypeError` is swallowed by the function's own broad except (logged at ERROR, returns `[]`); on the prefetch path nothing raises at all, because `visible_custom_api_clause`'s own `None`-owner reduction absorbs it. So the error-shape regression you predicted is already blocked two layers deeper on each path.

The tests are still worth having, for a stronger reason: they pin the *conjunction* of those layers. Remove the last line of defense on either path and each test goes red showing the team-granted API actually loading into a build with no identity — a real leak signal, not just a status-code change. That is what they now document and guard.

AlexLiu190625:
All six findings and both simplification suggestions adopted, at 601cf49f; the five inline threads are answered in place. The two body-level items:

- **Finding 3 (hook contract docstring still MCP-only):** adopted. `TeamConnectorVisibilityHook`'s credential paragraph now covers custom APIs explicitly — a team-shared custom API executes on the definition row's stored configuration for any runner, and there is no per-member credential override layer at all, which is starker than the MCP case.
- **The shared wrap helper:** adopted as `resolve_team_connector_ids_or_raise(db, *, team_id, log_subject)` in `connector_team_scope.py`, collapsing all four sites. One deliberate departure from the sketched signature: it returns the full `{"mcp": ..., "custom_api": ...}` dict rather than one kind's `frozenset`, because the runtime-context view loader consumes both keys — a `kind`-mandatory signature cannot serve that site. The three single-kind callers extract their key at the call site. Behavior is byte-equivalent (every existing failure-behavior test passed unchanged), and the consolidation bought one new cheap pin: the `ConnectorRuntimeError` passthrough arm, previously unpinned at all four sites, now has its own mutation-verified test.

Ordering-assertion strengthening was applied to the MCP sibling as well as the new test, per the pattern note in your comment.

src/xagent/web/services/connector_runtime.py
```diff
@@ -276,17 +276,11 @@ def load_connector_runtime_view(
...
-            # tool loaders and this branch's Custom API set to match). On
-            # the legacy-hook branch the filters diverge for Custom API: a
-            # legacy user-keyed hook can grant a team-shared Custom API
-            # here while the Custom API tool loader resolves junction-only,
-            # so such a ref stays visible without a runtime tool.
+            # Tool loading applies the same visibility filter when the
+            # team-scope hook is installed, for both connector kinds; with
+            # only the legacy hook this view stays user-keyed while both
+            # tool loaders resolve personal-only, so a ref absent here has
+            # no runtime tool either way.
```
rogercloud: The rewritten comment drops documentation of a still-live divergence. The parity claim ("a ref absent here has no runtime tool either way") holds only for refs *absent* from this view. The case the old comment documented is the opposite: with no `team_visibility` hook installed and only the legacy user-keyed `visibility` hook, a legacy-granted team-shared custom API is still visible here via the legacy hook, while both tool loaders in `config.py` resolve personal-only -- they consult `team_visibility` exclusively via `resolve_team_connector_ids_or_raise` and never the legacy hook. So such a custom API is selectable and persistable into a task's connector selection but silently builds no runtime tool. Suggest restoring the explicit note about this present-but-toolless case instead of the absent-ref parity claim.

tests/web/tools/test_custom_api_team_visibility.py
```diff
@@ -0,0 +1,609 @@
...
+    a_config = next(c for c in configs if c["name"] == seed.a.name)
+    assert a_config["env"] == {"API_KEY": "team-a-secret"}
+
+
+# ---------------------------------------------------------------------------
+# A personal agent gets nothing from any team, and writes nothing.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("env", ["env_t", "env_legacy"])
```
rogercloud, line 278: No test cell covers (legacy-hook-only environment) x (`connector_team_id=T1`) -- this parametrization only varies `env`, and the test body below hard-codes `connector_team_id=None`. So the legacy-hook divergence noted on `connector_runtime.py` is untested. It is new surface rather than a mirrored gap -- the MCP sibling suite `test_mcp_team_visibility.py` has no legacy-hook concept at all. Suggest adding an `env_legacy` x `connector_team_id=T1` cell asserting the actual (divergent) behavior so it is pinned deliberately.

tests/web/tools/test_custom_api_team_visibility.py
```diff
@@ -0,0 +1,609 @@
...
+    b                        -- another user's, reachable only through a
+                                team hook keyed on T2
+    inactive_own_shared     -- C holds an inactive personal link AND it is
+                                in T1's team-owned set -- the one cell where
+                                "personal ∪ team" is not a strict widening
+                                of C's own list (a member's personal
+                                deactivate does not hide a team-shared API)
+
+The personal junction rows for ``active_own`` and ``inactive_own_shared``
+are inserted in reverse ``custom_apis.id`` order (M3 in the plan): without a
```
rogercloud: This rationale -- seeding C's junction rows in reverse `custom_apis.id` order so "an ordering regression has nothing to move" -- is not backed by any test in the file. `test_no_hooks_matches_legacy_custom_api_result_set` explicitly declines to assert order (only one row is visible), and `test_snapshot_and_live_custom_api_paths_agree` compares two lists that both flow through the same `_visible_custom_api_query` builder, so they cannot structurally diverge in order. `ORDER BY` is actually pinned only by the compiled-SQL assertion in `test_production_custom_api_query_shape`, which is independent of insertion order. Either add a test that genuinely asserts returned order for a multi-row-visible case, or simplify the seed and drop this rationale.

src/xagent/web/tools/config.py
```diff
@@ -3461,26 +3473,28 @@ def get_custom_api_configs(self) -> List[Dict[str, Any]]:
...
-            )
+        team_api_ids = frozenset(
+            resolve_team_connector_ids_or_raise(
+                self.db, team_id=self._connector_team_id, log_subject=self._user_id
+            )["custom_api"]
+        )
 
-            if not user_apis:
-                return []
+        try:
```
rogercloud, line 3507: Hoisting team-scope resolution above this try block is correct and matches the MCP read point. But the query execution below (`_visible_custom_api_query(...).all()`) is still wrapped by a broad `except Exception: log ERROR; return []` further down in this same function, so a genuine DB failure silently degrades to "no custom-API tools" for the turn. The MCP twin `_load_mcp_server_configs` fails closed on the same class of failure by raising `MCPConfigLoadError()`. The swallow predates this PR, but since the diff rewrites exactly this block (swapping the query source and hoisting resolution above it), failing closed here would be a reasonable while-you're-here fix. Not required for this PR's stated goal.

src/xagent/web/services/connector_team_scope.py
```diff
@@ -176,6 +190,47 @@ def team_connector_hook_installed() -> bool:
     return _team_connector_visibility_hook is not None
 
 
+def resolve_team_connector_ids_or_raise(
+    db: Any, *, team_id: int | None, log_subject: Any
```
rogercloud: shrink: `log_subject: Any` is broader than needed. The docstring notes the value is never interpreted -- only rendered into a warning log line -- and every current and plausible call site passes an actual user id, so narrowing the annotation would document the real contract. Purely stylistic, no behavioral impact.

rogercloud (approved):
## Summary

Custom API tool loading previously resolved visibility by iterating only the user's personal junction rows, so a custom API shared with a team never loaded for a team-governed agent -- even though the sibling MCP connector path already resolved team-shared servers. This PR applies the same personal-∪-team visibility rule to custom APIs: both of `WebToolConfig`'s custom-API read points (the live call and the off-loop snapshot prefetch) now route through one module-level query builder with a deterministic `ORDER BY`, a team-scope resolution failure raises the same typed 503 the MCP path already raises, and the runtime-context view's custom-API half (previously narrowed to personal-only as a documented workaround) is widened back to match. Four team-scope-resolution/error-wrap call sites now share `resolve_team_connector_ids_or_raise` in the new `connector_team_scope.py`.

## Approach

**Sound.** This is the smallest correct move: it deletes a documented temporary narrowing rather than introducing a new mechanism, reuses the exact predicate and error-wrap shape that #1241 already proved for MCP, and removes a defensive branch (the old `if api is None: continue` personal-junction guard) instead of duplicating it. The shared query builder collapsing both read points onto one predicate is the right call -- it makes live/snapshot divergence structurally impossible rather than test-enforced.

One design-level note for context, not an open finding: widening custom-API visibility to team scope means a team-shared custom API's credentials can reach anonymous runners of a published team agent. That tradeoff is real, but it was raised earlier in this review cycle and the current head already addresses it by documenting the exposure explicitly in the `TeamConnectorVisibilityHook` docstring (`src/xagent/web/services/connector_team_scope.py:70-81`). Whether that exposure is acceptable is a policy decision for the product/multi-tenant layer, not a code defect here, so it is not listed below.

All findings are minor. Approving with suggestions.

## Findings

### 1. Rewritten comment drops a still-live divergence -- minor
`src/xagent/web/services/connector_runtime.py:279-283`

See inline comment.

### 2. Missing test cell for the legacy-hook x governing-team combination -- minor
`tests/web/tools/test_custom_api_team_visibility.py`

See inline comment.

### 3. Reverse-order seed is dead effort with a misleading justification -- minor
`tests/web/tools/test_custom_api_team_visibility.py`

See inline comment.

### 4. Query execution still fails open on DB errors -- minor (pre-existing, but rewritten here)
`src/xagent/web/tools/config.py:3489-3512`

See inline comment.

### 5. Stale `_connector_team_id` blast radius widened -- minor (pre-existing, follow-up)
`src/xagent/web/tools/config.py:1201`

`_connector_team_id` is assigned once in `__init__` (line 1201) and never refreshed. `set_connector_runtime_turn_id` clears several sibling caches but not this one, and `WebToolConfig` instances are cache-reused across turns within a task (the chat.py cache-hit path syncs turn id and execution scope, not team id). So if an agent's governing team changes mid-session, a reused config resolves visibility against a stale team. This staleness already existed for MCP via #1241; this PR newly keys custom-API visibility off the same never-refreshed field, widening it from MCP-only to both connector kinds. Suggest a follow-up issue to refresh or clear this alongside the other per-turn cache resets -- not a blocker here, and this line is unchanged by the diff so no inline comment is attached.

### 6. `team_connector_ids` has no production callers outside its wrapper -- very minor
`src/xagent/web/services/connector_team_scope.py:170-217`

Every production call site correctly goes through `resolve_team_connector_ids_or_raise`; the raw `team_connector_ids` function is called directly only by unit tests exercising lower-level validation. The wrapper's docstring already states all four production call sites must use it, which mostly covers this. Optional: rename to `_team_connector_ids` so the "always use the wrapper" rule is enforced by convention rather than only by docstring. This function definition is unchanged by the diff so no inline comment is attached.

### `shrink:` narrow `log_subject: Any` -- very minor
`src/xagent/web/services/connector_team_scope.py:193-194`

See inline comment.

## Note on the PR description (not code -- should not block merge)

The description states the team-scope resolution wrap "previously existed as four separate, near-identical copies... now it's one shared function; three call sites reduce to naming the connector kind they need." Verified against the base commit: only **two** near-identical copies existed pre-PR (config.py's MCP read point and connector_runtime.py's runtime-context view loader) -- the custom-API read points did no team resolution at all before this PR. And there are **four** call sites after it, not three (confirmed by grep and by the shared function's own docstring). Both counts are off, in opposite directions. "Four copies collapsing to three" materially overstates the pre-existing duplication this PR removes, so a description edit is worth making -- but it is description text only, not a code defect, and does not gate merge.

src/xagent/web/services/connector_runtime.py
AlexLiu190625: Restored at 95dccd46 — the comment now documents both halves: the absent-ref parity claim, and the present-but-toolless case (legacy-hook-only deployment: a legacy-granted team-shared custom API stays visible in this view while both tool loaders resolve through the team-keyed hook exclusively and never consult the legacy one). Verified both halves against the code before writing them.

tests/web/tools/test_custom_api_team_visibility.py
AlexLiu190625: Added at 95dccd46: `test_legacy_hook_with_governing_team_diverges_from_runtime_view` pins the divergence deliberately — the runtime view shows the legacy-granted API, the loader builds nothing — with a docstring stating it documents a known divergence rather than endorsing it. Mutation-verified: making the loader consult the legacy hook turns exactly this test red.

tests/web/tools/test_custom_api_team_visibility.py
AlexLiu190625: Adopted the second option at 95dccd46: seed simplified to natural insertion order, rationale deleted. The first option (a genuine returned-order assertion) was measured earlier in this work and is non-discriminating here: SQLite returns these rows in id order through its index whether or not the `ORDER BY` exists, so a Python-list order assertion stays green under the exact regression it would claim to catch. The compiled-SQL exact assertion in `test_production_custom_api_query_shape` remains the designated ordering detector.

src/xagent/web/tools/config.py
AlexLiu190625: Adopted at 95dccd46, with one deliberate deviation and one honest boundary note.

The query execution now fails closed: a DB failure raises a typed `ConnectorRuntimeError` (503, `custom_api_config_load_failed`) instead of degrading to an empty tool set. The deviation: it raises `ConnectorRuntimeError` rather than a parallel of `MCPConfigLoadError`, because that class feeds the MCP failure-policy enforcement and load summaries, neither of which has a custom-API counterpart — the typed connector error is what every frame between this site and the tool registry already re-raises explicitly. Verified end-to-end with a probe at the registry boundary: the typed error propagates past the registry loop's broad exception handler (a control with an untyped error confirms that handler is live and being bypassed, not absent), plus a registry-boundary test mirroring the MCP suite's, mutation-verified against the old swallow.

The boundary note: per-row config building keeps its existing non-fatal swallow on both sides, but the granularity still differs — the MCP twin isolates a failed row and keeps loading the others, while this loop's guard spans the whole loop. That remaining difference is now stated in the comment instead of implied away, and stays out of scope here.

src/xagent/web/services/connector_team_scope.py
AlexLiu190625: Adopted at 95dccd46 — narrowed to `int | None`. `int` alone is genuinely unsound (mypy catches it at the MCP read point, whose user id can be `None` at that frame), so the annotation now documents the real contract; the docstring's "only rendered, never interpreted" note still holds.

AlexLiu190625:
Round-2 items all addressed at 95dccd46; the five inline threads are answered in place. The body-level items:

- **Finding 5 (stale governing-team id on cache-reused configs):** filed as #1281 with the fix direction (refresh alongside the sibling per-turn cache resets) and the test shape (flip an agent's team between two turns on a reused config).
- **Finding 6 (rename `team_connector_ids` to underscore-private):** declined — the wrapper-only rule is already stated in the wrapper's docstring and enforced by the fact that every production call site goes through it; renaming a just-shipped function whose only direct callers are the validation unit tests adds diff churn without adding enforcement. If a future call site bypasses the wrapper, that is the review signal the convention would have given anyway.
- **PR description counts:** corrected — two pre-existing wrap copies (MCP read point, runtime-context view loader), and this change's two new custom-API sites now make four call sites sharing one function.
- The DB-failure fail-closed change closes the corresponding item recorded on #1250 (item 4).

PR #1267, opened 2026-08-11
fix: prefer task runtime computer tool

qinxuye:
## What this PR does

- suppresses the default Playwright browser family whenever a task-runtime provider contributes its own `computer` tool
- detects the capability from the normalized full provider contribution instead of hard-coding the built-in `local_browser` provider name
- normalizes contributed tool names consistently with runtime conflict reconciliation
- adds `create_default_tools` regression coverage proving Playwright is suppressed, the first out-of-tree `computer` provider is retained, and a second colliding provider is dropped as one unit with its environment contribution

## Why

The task-runtime extension hook is public, but the browser tool creator previously recognized only `local_browser`. An out-of-tree runtime that contributed `computer` therefore collided with the default Playwright tool and was silently removed as one provider unit. The task then controlled an unrelated ephemeral browser.

## Explicit boundary

This PR does not:

- add knowledge of any closed-source provider or runtime kind
- weaken duplicate-name isolation for arbitrary extension tools
- change tool policy, user overrides, or task selection semantics
- suppress Playwright for tasks with no contributed `computer` tool

A bound computer runtime owns that capability; if policy later filters it, Xagent must not silently substitute a different browser.

## Validation

- `PYTHONPATH=src uv run pytest -q -n 0 tests/web/services/test_local_browser_runtime.py tests/web/tools/test_runtime_computer_precedence.py tests/core/tools/adapters/vibe/test_factory_runtime_tools.py` — 38 passed
- Ruff 0.12.3 check and format validation using the standalone OSS repository settings — passed

### DIFF
--- a/src/xagent/core/tools/adapters/vibe/browser_tools.py
+++ b/src/xagent/core/tools/adapters/vibe/browser_tools.py
@@ -1,7 +1,7 @@
 """Browser automation tools registration using @register_tool decorator."""
 
 import logging
-from typing import TYPE_CHECKING, Any, List
+from typing import TYPE_CHECKING, Any
 
 from .factory import ToolFactory, register_tool
 
@@ -11,10 +11,8 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def _has_local_browser_runtime(contribution: Any) -> bool:
-    from ....computer.native_browser import (
-        LOCAL_BROWSER_TASK_EXTENSION,
-    )
+def _has_computer_runtime(contribution: Any) -> bool:
+    """Return whether the full pre-policy contribution declares ``computer``."""
     from ....task_runtime import (
         TaskRuntimeContribution,
         full_task_runtime_contribution,
@@ -24,29 +22,25 @@ def _has_local_browser_runtime(contribution: Any) -> bool:
         return False
     full_contribution = full_task_runtime_contribution(contribution)
     return any(
-        provider_name == LOCAL_BROWSER_TASK_EXTENSION
-        and any(
-            getattr(tool, "name", None) == "computer"
-            for tool in provider_contribution.tools
-        )
-        for provider_name, provider_contribution in (
-            full_contribution.provider_contributions
-        )
+        isinstance((name := getattr(tool, "name", None)), str)
+        and name.strip() == "computer"
+        for tool in full_contribution.tools
     )
 
 
 @register_tool(categories={"browser"})
-async def create_browser_tools(config: "BaseToolConfig") -> List[Any]:
+async def create_browser_tools(config: "BaseToolConfig") -> list[Any]:
     """Create browser automation tools."""
     if not config.get_browser_tools_enabled():
         return []
 
-    # A bound Local browser task contributes its own ``computer`` instance.
-    # Suppress the Playwright browser family as one unit so the runtime tool
-    # does not collide with the core tool or accidentally expose a second,
-    # unrelated browser to the model.
+    # A task runtime may contribute its own ``computer`` instance. Suppress the
+    # Playwright browser family as one unit so an out-of-tree runtime does not
+    # collide with the core tool or silently fall back to an unrelated browser.
+    # This is deliberately capability-based rather than provider-name-based:
+    # the extension hook is public and core cannot enumerate its providers.
     contribution = config.get_task_runtime_contribution()
-    if _has_local_browser_runtime(contribution):
+    if _has_computer_runtime(contribution):
         return []
 
     task_id = config.get_task_id()
--- a/tests/web/tools/test_runtime_computer_precedence.py
+++ b/tests/web/tools/test_runtime_computer_precedence.py
@@ -1,0 +1,111 @@
+from types import SimpleNamespace
+
+import pytest
+
+import xagent.web.api.chat as chat_module
+from xagent.core.task_runtime import (
+    TaskRuntimeContext,
+    TaskRuntimeContribution,
+    merge_task_runtime_contributions,
+)
+from xagent.core.tools.adapters.vibe import browser_tools as browser_tools_module
+from xagent.core.tools.adapters.vibe.base import ToolCategory
+from xagent.core.tools.adapters.vibe.config import ToolConfig
+from xagent.core.tools.adapters.vibe.factory import ToolFactory, ToolRegistry
+from xagent.web.api.chat import create_default_tools
+
+
+@pytest.mark.asyncio
+async def test_create_default_tools_prefers_out_of_tree_computer_runtime(
+    monkeypatch: pytest.MonkeyPatch,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    runtime_tool = SimpleNamespace(
+        name="computer",
+        metadata=SimpleNamespace(category=ToolCategory.OTHER),
+    )
+    conflicting_runtime_tool = SimpleNamespace(
+        name="computer",
+        metadata=SimpleNamespace(category=ToolCategory.OTHER),
+    )
+
+    class _FakeToolConfig(ToolConfig):
+        def __init__(self, **_kwargs):
+            super().__init__({})
+            self.runtime_contribution = TaskRuntimeContribution()
+            self.runtime_workspace = None
+
+        def get_workspace_config(self):
+            return {"task_id": "web_task_11"}
+
+        def set_task_runtime_contribution(self, contribution) -> None:
+            self.runtime_contribution = contribution
+
+        def get_task_runtime_contribution(self):
+            return self.runtime_contribution
+
+        def set_task_runtime_workspace(self, workspace) -> None:
+            self.runtime_workspace = workspace
+
+        def get_task_runtime_workspace(self):
+            return self.runtime_workspace
+
+        def get_browser_tools_enabled(self) -> bool:
+            return True
+
+    async def create_registered_tools(config):
+        return await browser_tools_module.create_browser_tools(config)
+
+    async def build_runtime(_context):
+        return merge_task_runtime_contributions(
+            {
+                "out_of_tree_browser": TaskRuntimeContribution(
+                    tools=(runtime_tool,),
+                    environment="Control the selected browser.",
+                    preferred_input_modalities=("image",),
+                ),
+                "second_browser": TaskRuntimeContribution(
+                    tools=(conflicting_runtime_tool,),
+                    environment="Control a different browser.",
+                ),
+            }
+        )
+
+    monkeypatch.setattr("xagent.web.tools.config.WebToolConfig", _FakeToolConfig)
+    monkeypatch.setattr(
+        "xagent.web.models.database.get_session_local",
+        lambda: object(),
+    )
+    monkeypatch.setattr(
+        ToolRegistry,
+        "create_registered_tools",
+        create_registered_tools,
+    )
+    monkeypatch.setattr(
+        ToolFactory,
+        "create_workspace",
+        lambda _config: SimpleNamespace(id="workspace"),
+    )
+    monkeypatch.setattr(chat_module, "build_task_runtime", build_runtime)
+    monkeypatch.setattr(
+        chat_module,
+        "registered_task_extensions",
+        lambda: ("out_of_tree_browser",),
+    )
+
+    tools, config = await create_default_tools(
+        None,
+        user=SimpleNamespace(id=7, is_admin=False),
+        task_id="web_task_11",
+        task_runtime_context=TaskRuntimeContext(
+            task_id=11,
+            user_id=7,
+            source="internal",
+            session_factory=lambda: object(),
+        ),
+    )
+
+    assert tools == [runtime_tool]
+    assert config.runtime_contribution.tools == (runtime_tool,)
+    assert config.runtime_contribution.environment == "Control the selected browser."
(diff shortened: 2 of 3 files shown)


src/xagent/core/tools/adapters/vibe/browser_tools.py
```diff
@@ -24,29 +21,27 @@ def _has_local_browser_runtime(contribution: Any) -> bool:
         return False
     full_contribution = full_task_runtime_contribution(contribution)
     return any(
-        provider_name == LOCAL_BROWSER_TASK_EXTENSION
-        and any(
+        any(
             getattr(tool, "name", None) == "computer"
```
rogercloud: MINOR: this comparison is exact, but the collision logic this gate pre-empts normalizes the name — `str(tool.name).strip()` in `src/xagent/core/task_runtime.py:346` and in `merge_task_runtime_contributions` (~lines 246-248). A provider contributing a tool literally named `" computer "` therefore fails this check (Playwright is not suppressed) yet still collides downstream once stripped, reproducing the exact silent-substitution bug this PR fixes.

The gap existed in `_has_local_browser_runtime` too, but that gate also required a single hard-coded first-party provider name; applying the same weak match to arbitrary third-party tool names widens the exposure. Whitespace-padded tool names are already a tested bug class here — see `tests/core/tools/adapters/vibe/test_factory_runtime_tools.py::test_whitespace_padded_duplicate_tool_reports_its_provider`.

Suggest normalizing identically:

```python
name = getattr(tool, "name", None)
isinstance(name, str) and name.strip() == "computer"
```

src/xagent/core/tools/adapters/vibe/browser_tools.py
```diff
@@ -24,29 +21,27 @@ def _has_local_browser_runtime(contribution: Any) -> bool:
...
             for tool in provider_contribution.tools
         )
-        for provider_name, provider_contribution in (
-            full_contribution.provider_contributions
-        )
+        for _provider_name, provider_contribution in full_contribution.provider_contributions
```
rogercloud: MINOR (simplification): the nested loop discards the provider name (`_provider_name`) — it only existed because the old code needed it for the `== LOCAL_BROWSER_TASK_EXTENSION` check. `merge_task_runtime_contributions` builds `.tools` as the flattened concatenation of every provider's tools in the same pass that builds `.provider_contributions`, so the two are always in sync for anything reachable here. A flat form is behaviorally identical and drops both the unused binding and the nesting:

```python
return any(
    getattr(tool, "name", None) == "computer"
    for tool in full_contribution.tools
)
```

src/xagent/core/tools/adapters/vibe/browser_tools.py
```diff
@@ -11,10 +11,7 @@
 logger = logging.getLogger(__name__)
 
 
-def _has_local_browser_runtime(contribution: Any) -> bool:
-    from ....computer.native_browser import (
-        LOCAL_BROWSER_TASK_EXTENSION,
-    )
+def _has_computer_runtime(contribution: Any) -> bool:
```
rogercloud, line 14: MINOR (docs): pre-existing gap, but the rename to the broader-sounding `_has_computer_runtime` plus the "any provider" implementation makes the narrow semantics easy to misread. A reader may assume "some computer runtime exists, bound or not, policy-filtered or not," whereas this actually means: the pre-policy, full task-runtime contribution declares at least one tool literally named `computer`, from any provider. The sibling helper resolving the same pre-policy nuance (`_full_stored_contribution` in `factory.py`) carries a one-line docstring stating exactly that, so a matching one line here would be consistent with existing convention.

tests/web/tools/test_runtime_computer_precedence.py
```diff
@@ -0,0 +1,101 @@
...
+        def get_browser_tools_enabled(self) -> bool:
+            return True
+
+    async def create_registered_tools(config):
+        return await browser_tools_module.create_browser_tools(config)
+
+    async def build_runtime(_context):
+        return merge_task_runtime_contributions(
+            {
+                "out_of_tree_browser": TaskRuntimeContribution(
```
rogercloud, line 62: MINOR (coverage): no test — here or elsewhere — pairs *two* different providers each contributing a tool named `"computer"`. That case is newly reachable because of this PR's generalization: previously only the hard-coded `local_browser` provider could trigger the gate. The individual mechanisms (name-agnostic collision-drop in reconciliation, and computer-detection-triggers-suppression) are each unit-tested, so this is an interaction gap rather than a correctness risk.

Worth adding a case asserting: Playwright is suppressed, exactly one `computer` tool survives reconciliation, the second provider is dropped as a conflict (with the warning logged), and its `environment` string is excluded from the merged prompt context.

tests/web/tools/test_runtime_computer_precedence.py
```diff
@@ -0,0 +1,101 @@
...
+        task_id="web_task_11",
+        task_runtime_context=TaskRuntimeContext(
+            task_id=11,
+            user_id=7,
+            source="internal",
+            session_factory=lambda: object(),
+        ),
+    )
+
+    assert tools == [runtime_tool]
```
rogercloud, line 108: MINOR (PR description wording, no code change needed): the description says this test proves the out-of-tree computer tool "survives reconciliation." Because the monkeypatched `create_registered_tools` returns only the (now-suppressed, hence empty) browser creator output, `reserved_tool_names` reaching `reconcile_task_runtime_contribution_tools` is provably `set()` — there is no live collision to survive.

The test isn't vacuous: if suppression regressed, the real Playwright `ComputerTool` (`.name == "computer"`) would populate `reserved_tool_names`, the out-of-tree provider would be dropped, and this assertion would fail. So it does guard the suppression gate end-to-end. Suggest rewording the description to something like "confirms the out-of-tree contribution isn't accidentally dropped when Playwright suppression works," since "survives reconciliation" implies a naming contest this path never constructs.

rogercloud (approved):
## Summary

`"computer"` is a shared tool-name namespace: the built-in Playwright browser family leads with a `ComputerTool` whose `.name` is literally `"computer"`. When a task-runtime provider also contributes a tool named `computer`, reconciliation sees a collision against a reserved core name and drops the entire provider unit — tools, prompt environment, and modality preference — leaving the task driving the unrelated ephemeral Playwright browser. The old gate suppressed Playwright only for the hard-coded in-tree `local_browser` provider, so any out-of-tree provider using the public extension hook hit exactly that silent substitution. This PR keys suppression on the capability (a contributed tool named `computer`) instead of the provider identity.

## Approach verdict

**Sound in direction, acceptable with reservations in execution.**

- Generalizing from "named provider + computer tool" to "any provider + computer tool" is the correct fix. The failure mode is intrinsically a name collision on `"computer"`, independent of which provider caused it, so gating on provider name *was* the bug.
- A separate "capability flag" alternative would be worse: it adds a second concept that can drift out of sync with the name that actually drives the collision in reconciliation. Keying on the same string the collision keys on is the right coupling.
- Reading the pre-policy `full_task_runtime_contribution` rather than a post-policy/filtered view is also right, and matches the PR's stated boundary: if policy later filters the runtime `computer` tool, the task should lose the capability rather than be silently handed a different browser.
- The "does not weaken duplicate-name isolation" claim was checked against `merge_task_runtime_contributions` / `reconcile_task_runtime_contribution_tools` and holds — reconciliation still drops a second colliding provider as a unit with a logged warning, and this PR does not change that.

No blocking design-level finding.

## Findings (all MINOR)

1. **`src/xagent/core/tools/adapters/vibe/browser_tools.py:25` — tool-name match is not normalized the way the collision logic is.**
   The gate compares `getattr(tool, "name", None) == "computer"` exactly, but the reconciliation logic this gate exists to pre-empt normalizes via `str(tool.name).strip()` (`src/xagent/core/task_runtime.py:346`, and in `merge_task_runtime_contributions` around lines 246-248). A provider contributing a tool literally named `" computer "` therefore fails this gate (Playwright is *not* suppressed) yet still collides downstream once its name is stripped — reproducing the exact silent-substitution bug this PR fixes. The gap existed in `_has_local_browser_runtime` too, but that gate additionally required a single hard-coded, first-party provider name; applying the same weak match to arbitrary third-party tool names meaningfully widens exposure. Whitespace-padded tool names are already treated as a real bug class in this codebase (`tests/core/tools/adapters/vibe/test_factory_runtime_tools.py::test_whitespace_padded_duplicate_tool_reports_its_provider`). Suggested fix: normalize identically, e.g. `name = getattr(tool, "name", None); isinstance(name, str) and name.strip() == "computer"`.

2. **`src/xagent/core/tools/adapters/vibe/browser_tools.py:22-29` — nested loop is a leftover shape.**
   The double loop over `provider_contributions` discards the provider name (`_provider_name`); it only existed because the old code needed that name for its `== LOCAL_BROWSER_TASK_EXTENSION` check. `merge_task_runtime_contributions` builds `.tools` as the flattened concatenation of every provider's tools in the same pass that builds `.provider_contributions`, so the two are always in sync for every value reachable here. A flat `any(getattr(t, "name", None) == "computer" for t in full_contribution.tools)` is behaviorally identical and simpler.

3. **`tests/web/tools/test_runtime_computer_precedence.py:57` — no test covers two providers both contributing `computer`.**
   This edge case is newly reachable *because of* this PR's generalization: previously only one hard-coded provider could trigger the gate. The underlying mechanisms (name-agnostic collision-drop in reconciliation, and computer-detection-triggering-suppression) are each unit-tested individually, so this is an interaction/integration gap rather than a correctness risk. Worth adding a test asserting: Playwright is suppressed, exactly one `computer` tool survives reconciliation, the second provider is dropped as a conflict (logged), and its `environment` string is excluded from the merged prompt context.

4. **`tests/web/tools/test_runtime_computer_precedence.py:99` — PR description overstates what this test proves.**
   The description says the test proves the out-of-tree `computer` tool "survives reconciliation." In this test the monkeypatched `create_registered_tools` returns only the (now-suppressed, hence empty) browser creator's output, so `reserved_tool_names` reaching `reconcile_task_runtime_contribution_tools` is provably `set()` — there is no live collision to survive. The test is *not* vacuous: if suppression regressed, the real Playwright `ComputerTool` (`.name == "computer"`) would populate `reserved_tool_names`, the out-of-tree provider would be dropped, and `assert tools == [runtime_tool]` would fail. So it does guard the suppression gate end-to-end; only the description's "survives reconciliation" framing implies a naming contest this code path never constructs. Description-wording nit only — no code change needed.

5. **`src/xagent/core/tools/adapters/vibe/browser_tools.py:14` — missing docstring on a now-broader helper.**
   Pre-existing (the predecessor lacked one too), but the rename to the broader-sounding `_has_computer_runtime` plus the "any provider" implementation raises the odds a reader assumes it means "some computer runtime exists, bound or not, policy-filtered or not." The actual semantics are narrower: *the pre-policy, full task-runtime contribution declares at least one tool literally named `computer`, from any provider.* The sibling helper resolving the same pre-policy nuance (`_full_stored_contribution` in `factory.py`) carries a one-line docstring calling that out, so this matches existing convention rather than imposing a new bar.

Out of scope, noted only as context: the bare `"computer"` literal is duplicated across several pre-existing, untouched call sites (`computer.py:192`, `computer.py:212`, `computer.py:229`, `react.py:3044`). Centralizing it is a worthwhile follow-up but not this diff's job.

## Test results

`PYTHONPATH=src uv run pytest -q tests/web/services/test_local_browser_runtime.py tests/web/tools/test_runtime_computer_precedence.py` — **17 passed** (independently reproduced under Python 3.12; the environment's default 3.14 hits an unrelated dependency resolution issue, not introduced by this PR).

## Recommendation

Approve with minor suggestions. The design is correct and the fix targets the actual root cause. Finding 1 is the only one with a behavioral edge (a whitespace-padded name still reaches the bug this PR fixes) and is a one-line change worth taking before merge; findings 2-5 are cleanups, coverage, and wording that can land here or as follow-ups.

src/xagent/core/tools/adapters/vibe/browser_tools.py
qinxuye: Fixed in 4384e7b5. The gate now accepts only string names and compares name.strip() with computer, matching the runtime normalization that drives collision detection. Added a regression for a whitespace-padded computer name.

src/xagent/core/tools/adapters/vibe/browser_tools.py
qinxuye: Simplified in 4384e7b5. The helper now scans full_contribution.tools directly; provider identity is intentionally irrelevant for this capability check.

src/xagent/core/tools/adapters/vibe/browser_tools.py
qinxuye: Added the focused pre-policy semantics docstring in 4384e7b5.

tests/web/tools/test_runtime_computer_precedence.py
qinxuye: Covered in 4384e7b5. The end-to-end test now contributes computer from two providers and verifies Playwright suppression, first-provider retention, second-provider conflict logging, and removal of the second provider environment.

tests/web/tools/test_runtime_computer_precedence.py
qinxuye: Updated the PR description. The test now also constructs an actual two-provider naming conflict, so the description explicitly states which provider is retained and which provider unit is dropped.

PR #1269, opened 2026-08-11
test: cover empty final answer with sibling tool call

qinxuye:
## What changed

- Rebased onto current `upstream/main`, which now contains the protocol-level empty-`final_answer` fix from #1325.
- Removed the superseded PR implementation and retained one focused regression test.
- Cover an empty `final_answer` followed by a sibling work-tool call: the invalid response is discarded, the sibling is not executed, the retry receives the corrective instruction, and persisted tool-call/result pairs remain complete.

## Why

The original implementation was superseded by #1325. The remaining review risk was that a trailing sibling tool call could make context sanitization drop the empty-answer feedback and turn the repair into a blind retry. The new test pins the current response-level rejection behavior against that regression.

## Validation

- `pytest -q tests/core/agent/test_react.py tests/core/agent/test_result.py tests/core/agent/test_execution_adapter.py tests/core/tools/test_agent_tool_session_isolation.py` (221 passed)
- Focused empty-final-answer tests (9 passed)
- `ruff check` and `ruff format`
- Pre-commit hooks, including mypy, isort, and codespell

### DIFF
--- a/tests/core/agent/test_react.py
+++ b/tests/core/agent/test_react.py
@@ -5138,10 +5138,12 @@ class StreamingEmptyFinalAnswerLLM:
         recover: bool = True,
         broken_arguments: str = '{"response_language":"English","outcome":"completed"}',
         preamble: str = "",
+        trailing_work_tool: bool = False,
     ) -> None:
         self.recover = recover
         self.broken_arguments = broken_arguments
         self.preamble = preamble
+        self.trailing_work_tool = trailing_work_tool
         self.stream_calls: list[dict[str, Any]] = []
 
     async def chat(self, **kwargs: Any) -> Any:
@@ -5161,14 +5163,25 @@ class StreamingEmptyFinalAnswerLLM:
         )
         if self.preamble:
             yield StreamChunk(type=ChunkType.TOKEN, delta=self.preamble)
+        tool_calls = [
+            {
+                "id": f"call_final_{call_index}",
+                "function": {"name": "final_answer", "arguments": arguments},
+            }
+        ]
+        if call_index == 0 and self.trailing_work_tool:
+            tool_calls.append(
+                {
+                    "id": "call_work_0",
+                    "function": {
+                        "name": "calculator",
+                        "arguments": '{"expression":"2+2"}',
+                    },
+                }
+            )
         yield StreamChunk(
             type=ChunkType.TOOL_CALL,
-            tool_calls=[
-                {
-                    "id": f"call_final_{call_index}",
-                    "function": {"name": "final_answer", "arguments": arguments},
-                }
-            ],
+            tool_calls=tool_calls,
         )
         yield StreamChunk(type=ChunkType.END)
 
@@ -5225,6 +5238,42 @@ async def test_react_retries_final_answer_that_omits_the_answer_field() -> None:
         "final_answer_end",
     ]
     assert outbound.events[-1]["content"] == "The result is 4."
+
+
+@pytest.mark.asyncio
+async def test_react_discards_empty_final_answer_with_trailing_work_call() -> None:
+    """Discard the invalid response so retry feedback survives sanitization."""
+
+    llm = StreamingEmptyFinalAnswerLLM(trailing_work_tool=True)
+    pattern, context, runtime, _outbound, _tracer = _react_empty_final_answer_fixture()
+    tool = FakeTool()
+
+    result = await pattern.run(
+        context=context,
+        tools=[tool],
+        llm=llm,
+        runtime=runtime,
+    )
+
+    assert result["success"] is True
+    assert result["response"] == "The result is 4."
+    assert tool.calls == []
+    assert len(llm.stream_calls) == 2
+    assert (
+        "previous response called final_answer with an empty answer"
+        in (llm.stream_calls[1]["messages"][0]["content"])
+    )
+
+    assistant_tool_call_ids = {
+        tool_call["id"]
+        for message in context.messages
+        for tool_call in (message.tool_calls or [])
+    }
+    tool_result_ids = {
+        message.tool_call_id for message in context.messages if message.role == "tool"
+    }
+    assert assistant_tool_call_ids == {"call_final_1"}
+    assert assistant_tool_call_ids <= tool_result_ids
 
 
 @pytest.mark.asyncio


rogercloud:
ci is failing

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -2167,6 +2186,9 @@ async def _execute_pending_tool_calls(
                     if control_result.get("status") == "waiting_for_user":
                         self._discard_pending_tool_plan_after_pause(context)
                         return control_result
+                    if control_result.get("success") is False:
```
rogercloud: This branch returns the failure terminally, which contradicts the PR description ("Retry the ReAct turn instead of marking it completed with an empty response"): `_run_tool_calling_loop` returns this `pending_result` immediately (react.py:398-403, 664-669), so `max_iterations` never advances, and `AgentRunner.run` ends the run (runner.py:291-313). A top-level (non-delegated) run that emits one bad `final_answer` call now hard-fails with an implementation-jargon error even when iterations remain; "retry" exists only at the delegation boundary, and the new regression test pins the no-retry behavior (`assert len(llm.calls) == 1`). Either restore the within-turn retry — append the failure tool result to context and let the loop continue so the LLM can self-correct, mirroring the `malformed_tool_arguments` recovery at react.py:858-867 — or update the PR description to state that the run now fails.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -2167,6 +2186,9 @@ async def _execute_pending_tool_calls(
...
+                        self.pending_tool_calls = []
```
rogercloud: The checkpoint at react.py:2174-2181 runs before this clear, so the persisted pattern state still holds the sibling calls the failed run discards; resuming from that checkpoint would re-execute them. The completed path enforces the opposite invariant (`test_react_pattern_final_answer_clears_trailing_pending_before_checkpoint`, tests/core/agent/test_react.py:3178-3226). Suggest clearing `pending_tool_calls` before the checkpoint (or checkpointing after the clear), and extending the checkpoint-cleanliness test to the failed path.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -743,8 +743,9 @@ def _final_answer_tool_content(
             if tool_call.get("name") != "final_answer":
                 continue
             args = tool_call.get("args")
-            if isinstance(args, dict):
-                return str(args.get("answer", ""))
+            answer = self._get_non_empty_final_answer(args)
```
rogercloud: Because `_final_answer_tool_content` now returns `None` for empty/whitespace answers, `_finish_streamed_answer_if_final` (react.py:728-736) neither finishes nor fails the streamer when `final_answer` is rejected, and the failure branch in `_handle_control_tool` has no handle to it (the streamer is local to `_run_tool_calling_loop`, react.py:489-493). `emit_delta` treats whitespace as truthy and starts the outbound stream (final_answer_stream.py:46-53); the runtime emits `final_answer_start` (runtime.py:553-565) and then nothing. The base code returned `str(...)` and `finish("")` closed the stream. Every other terminal path fails the streamer (react.py:498-499, 510-514, 587-588, 702-703). Suggest failing the streamer on this path — e.g. in the loop's `success is False` branch, which has access to it — so the outbound stream always terminates with `final_answer_error`.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1683,7 +1684,25 @@ async def _handle_control_tool(
         args = tool_call.get("args", {})
 
         if name == "final_answer":
-            answer = str(args.get("answer", ""))
+            answer = self._get_non_empty_final_answer(args)
+            if answer is None:
+                result = {
```
rogercloud: This failure result is a hand-built dict that bypasses the PatternResult envelope used by every other terminal path (react.py:687-691, 710-720, 2670-2679): no `metadata` (e.g. `iterations`) key, and the top-level `"status": "failed"` is hardcoded separately from `self.status = "failed"` (react.py:1705). No current consumer reads `metadata.iterations` (grep-verified), so the impact is consistency/maintainability. Suggest:

```python
PatternResult(
    success=False,
    error="Invalid final_answer arguments. The answer field must be a non-empty string.",
    metadata={"status": self.status, "iterations": self.current_iteration + 1},
).to_dict()
```

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1683,7 +1684,25 @@ async def _handle_control_tool(
...
+                    "success": False,
+                    "status": "failed",
+                    "failure_code": "missing_delegated_output",
+                    "error": (
```
rogercloud: Two issues. (a) The error text "Invalid final_answer arguments. The answer field must be a non-empty string." is internal jargon that becomes the user-facing output of top-level runs (execution_adapter.py:361 falls back to `result["error"]`; websocket.py:3343) and the parent LLM's message for delegated runs (agent_tool.py:1656-1657) — the base code never surfaced error text here (it completed with empty output). (b) `_record_tool_call(tool_call, status="failed", result=result)` omits `error=`, which the other failed records pass (react.py:2833-2838, 2861-2866), so the ledger record persists with `error=None`. No logic reads the ledger error today; impact is diagnostic/trace fidelity. Suggest a friendlier user-facing message (keep `failure_code` for classification) and pass `error=result["error"]` to `_record_tool_call`.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -2620,6 +2642,15 @@ def _final_answer_outcome(value: Any) -> str:
             return "blocked"
         return outcome
 
+    @staticmethod
+    def _get_non_empty_final_answer(args: Any) -> str | None:
```
rogercloud: The name `_get_non_empty_final_answer` understates the contract: it also enforces dict-ness and `str` type. The non-dict branch (react.py:2647-2648) is verified unreachable defensive code — `_coerce_arguments` (react.py:1446-1455) always returns a dict, and both callers receive normalized args. The single generic error message accurately covers all reachable causes, so there is no user-facing weakness. Suggest a docstring or rename (e.g. `_get_valid_final_answer`), and optionally drop the unreachable branch.

tests/core/agent/test_react.py
```diff
@@ -3144,6 +3144,36 @@ async def test_react_pattern_can_finish_with_final_answer_tool() -> None:
...
+        responses=[
+            {
+                "tool_calls": [
+                    {
+                        "id": "call_empty_final",
+                        "function": {
+                            "name": "final_answer",
+                            "arguments": "{}",
+                        },
+                    }
```
rogercloud: Coverage is adequate for the core regression: this unit test plus the two pre-existing e2e tests (tests/core/tools/test_agent_tool_session_isolation.py:1007, 1047) were verified to fail if either propagation line (execution_adapter.py:382-384 or agent_tool.py:1652-1662) were reverted. Remaining gaps: parametrize over `{"answer": ""}`, `{"answer": "   "}`, `{"answer": 42}`, a missing answer with a valid outcome (previously completed, now fails), and non-dict args; also, no test exercises the new failed-control branch with non-empty trailing pending calls.

rogercloud (commented):
## Review of PR #1269 — fix: reject empty final answers

### Summary

A model can emit a `final_answer` tool call whose answer is missing, empty, whitespace-only, or a non-string; the ReAct pattern previously accepted such calls and completed the turn with an empty/garbage response (the old code did `str(args.get("answer", ""))`). This PR rejects such calls instead of completing: validation is centralized in a single predicate `_get_non_empty_final_answer` shared by the execution path and the streaming/display path; `_handle_control_tool` records the failed call, adds the failure result to context, and returns a failed result; a new loop branch terminates the run with it; the execution adapter and the delegation classifier propagate the allowlisted `failure_code` so delegated parents classify with the existing `missing_delegated_output`; one regression test covers `{}`. CI is green (pre-flight observed).
Blocking: yes — recommended event: COMMENT (author rule; a major finding exists that would normally warrant REQUEST_CHANGES).

### Approach verdict

Acceptable with reservations. Validation lands at the single choke point, reuses the existing failure-code allowlist vocabulary, and keeps the parent-facing delegation surface stable (the completed-but-empty branch already surfaced `missing_delegated_output`, so delegated classification is unchanged). The reservations: the delivered behavior contradicts the PR description (the run fails terminally rather than retrying), and several lifecycle/consistency gaps remain around streaming, checkpoints, and result shaping.

### Findings

#### Major

**F1 — Description/behavior mismatch: the run fails; there is no retry** — `react.py:2189-2191` (major)

The PR description states "Retry the ReAct turn instead of marking it completed with an empty response", but the delivered code does not retry at any layer:

- `_handle_control_tool` returns `{success: False, status: "failed", failure_code: "missing_delegated_output", error: ...}` (react.py:1688-1706).
- The new loop branch (react.py:2189-2191) returns that result; `_run_tool_calling_loop` returns `pending_result` immediately (react.py:398-403, 664-669), so `max_iterations` never advances.
- `AgentRunner.run` ends the run (runner.py:291-313); no `failure_code`-driven retry exists anywhere.

Commit evolution confirms the intent change: b6bfc8c4 implemented a retry (returned `None` so the loop continued to the next LLM iteration, with the test asserting success after 2 calls); 75f2b43c replaced it with terminal failure and re-pinned the test (`assert len(llm.calls) == 1`). Consequence: a top-level (non-delegated) run that emits one bad `final_answer` call hard-fails with an implementation-jargon error even when iterations remain; "retry" exists only at the delegation boundary.

Suggestion: either restore the within-turn retry — append the failure tool result to context and let the loop continue so the LLM can self-correct, mirroring the `malformed_tool_arguments` recovery at react.py:858-867 — or update the PR description to state that the run now fails.

#### Minor

**F2 — `missing_delegated_output` mislabels top-level failures** — `react.py:1692` (minor)

`failure_code: "missing_delegated_output"` is delegation-boundary vocabulary (TOOL_FAILURE_CODES allowlist, result.py:7-13; its original producer is the delegated-child completed-but-empty classifier at agent_tool.py:1664-1670), and it is now emitted by a top-level non-delegated failure. The top-level run's public result and traces carry a code asserting delegation that never happened. Note the parent-facing surface is unchanged — the base already surfaced an empty-answer child as `missing_delegated_output` via the completed-but-empty branch — so the PR actually improves the delegated message (distinct error text); the residual issue is only the top-level mislabel. Suggestion: document the dual meaning in result.py, or introduce a dedicated code (e.g. `invalid_final_answer`) and map it at the delegation boundary.

**F3 — Rejected `final_answer` leaves the outbound stream dangling** — `react.py:746-748` (minor)

`_final_answer_tool_content` (react.py:738-749) now returns `None` for empty/whitespace answers, so `_finish_streamed_answer_if_final` (react.py:728-736) neither finishes nor fails the streamer; the failure branch in `_handle_control_tool` has no handle to it (the streamer is local to `_run_tool_calling_loop`, react.py:489-493). `emit_delta` treats whitespace as truthy and starts the outbound stream (final_answer_stream.py:46-53); the runtime emits `final_answer_start` (runtime.py:553-565) and then nothing. The base code returned `str(...)` and `finish("")` closed the stream via `content or self._content`. Models do produce whitespace-only answers — that is why this PR exists — and every other terminal path fails the streamer (react.py:498-499, 510-514, 587-588, 702-703). Suggestion: fail the streamer on this path — e.g. in the loop's `success is False` branch, which has access to it — so the outbound stream always terminates with `final_answer_error`.

**F4 — Optional consistency note: trailing pending calls cleared without ledger records** — `react.py:2190` (minor)

The failed branch clears `pending_tool_calls` without ledger records/backfill for the trailing sibling calls that were never executed. This follows the pre-existing completed-branch convention (a control tool ends the turn's tool execution) and the run is terminal anyway, so the state is inert; noted only as a consistency consideration, not a defect.

**F5 — Failure result bypasses the PatternResult envelope** — `react.py:1689-1697` (minor)

The failure result is a hand-built dict that bypasses the PatternResult envelope used by every other terminal path (react.py:687-691, 710-720, 2670-2679): it carries no `metadata` (e.g. `iterations`) key, and the top-level `"status": "failed"` is hardcoded separately from `self.status = "failed"` (react.py:1705). No current consumer reads `metadata.iterations` (grep-verified), so the impact is consistency/maintainability. Suggestion:

```python
PatternResult(
    success=False,
    error="Invalid final_answer arguments. The answer field must be a non-empty string.",
    metadata={"status": self.status, "iterations": self.current_iteration + 1},
).to_dict()
```

**F6 — Internal-jargon error text surfaces to users/parent; ledger record loses the error** — `react.py:1693-1697` (minor)

(a) The error text "Invalid final_answer arguments. The answer field must be a non-empty string." is internal jargon that becomes the user-facing output of top-level runs (execution_adapter.py:361 falls back to `result["error"]`; websocket.py:3343) and the parent LLM's message for delegated runs (agent_tool.py:1656-1657) — the base code never surfaced error text here (it completed with empty output). (b) `_record_tool_call(tool_call, status="failed", result=result)` omits `error=`, which the other failed records pass (react.py:2833-2838, 2861-2866), so the ledger record persists with `error=None`. No logic reads the ledger error today; impact is diagnostic/trace fidelity. Suggestion: use a friendlier user-facing message (keep `failure_code` for classification) and pass `error=result["error"]` to `_record_tool_call`.

**F7 — Checkpoint runs before `pending_tool_calls` is cleared** — `react.py:2189-2191` (minor)

The checkpoint at react.py:2174-2181 runs before this branch clears `pending_tool_calls` (react.py:2190), so the persisted pattern state still holds the sibling calls the failed run discarded; resuming from that checkpoint would re-execute them. The completed path enforces the opposite invariant (`test_react_pattern_final_answer_clears_trailing_pending_before_checkpoint`, tests/core/agent/test_react.py:3178-3226). Suggestion: clear `pending_tool_calls` before the checkpoint (or checkpoint after clearing), and optionally extend the checkpoint-cleanliness test to the failed path.

**F8 — Test coverage gaps** — `tests/core/agent/test_react.py:3159-3174` (minor)

Coverage is adequate for the core regression: the new unit test (`test_react_rejects_empty_final_answer_instead_of_completing`) plus two pre-existing e2e tests (tests/core/tools/test_agent_tool_session_isolation.py:1007, 1047) that were verified to fail if either propagation line (execution_adapter.py:382-384 or agent_tool.py:1652-1662) were reverted. Remaining gaps: parametrize over `{"answer": ""}`, `{"answer": "   "}`, `{"answer": 42}`, a missing answer with a valid outcome (previously completed, now fails), and non-dict args; no test exercises the new failed-control branch with non-empty trailing pending calls.

**F9 — Predicate name understates its contract; unreachable branch** — `react.py:2646-2652` (minor)

The name `_get_non_empty_final_answer` understates the contract: it also enforces dict-ness and `str` type. The non-dict branch (react.py:2647-2648) is verified unreachable defensive code — `_coerce_arguments` (react.py:1446-1455) always returns a dict, and both callers receive normalized args. The single generic error message accurately covers all reachable causes, so there is no user-facing weakness. Suggestion: add a docstring or rename (e.g. `_get_valid_final_answer`), and optionally drop the unreachable branch.

**F10 — Model-facing schema does not declare the new non-emptiness constraint** — `react.py:1535-1544` (minor)

The `final_answer` schema declares `answer` as `required: string` but no `minLength: 1`; the runtime now enforces non-emptiness that the model-facing schema never declares. Since schemas are advisory for LLMs, the runtime check remains primary; adding `minLength` is a trivial safe change because builtin schemas bypass `_compact_tool_json_schema`.

- **Blocking: yes.**
- Blocking issues:
  - `react.py:2189-2191` — major — description/behavior mismatch: the run terminates as failed with no retry at any layer, and the new test pins the no-retry behavior — [new]
- Non-blocking (recommended): the minor findings above — F3 dangling stream, F6 user-facing error text / ledger fidelity, F7 checkpoint ordering, F5/F9 consistency, F8 test gaps, F2/F10 semantic notes.
- Recommended event: **COMMENT** (qinxuye author rule; a major finding would normally warrant REQUEST_CHANGES).

src/xagent/core/agent/pattern/react/react.py
qinxuye: 已修复。无效 final_answer 分支会在调用方写 checkpoint 之前清空 pending_tool_calls，并新增回归测试验证尾随 calculator 不会执行、持久化 checkpoint 中 pending_tool_calls 为空。

src/xagent/core/agent/pattern/react/react.py
qinxuye: 已按建议恢复为有界修复：第一次收到无效 final_answer 时，将失败的工具结果写回上下文并继续一次 ReAct 循环，让模型自我纠正；第二次仍无效或迭代预算耗尽时才以 invalid_final_answer 终止。新增了首次无效后成功、连续两次无效及 max_iterations=1 的回归测试。

qinxuye:
Addressed the full review in 0d98cafe after rebasing onto upstream/main:

- restore one bounded in-turn repair for invalid final_answer, then fail deterministically
- persist the repair state and clear pending sibling calls before checkpointing
- close invalid streamed answers with final_answer_error
- return a PatternResult-shaped invalid_final_answer failure with friendly text and ledger error details
- validate final_answer.answer with schema minLength and expanded edge-case tests
- keep invalid_final_answer as the top-level pattern cause, while mapping it to missing_delegated_output only at the delegated-agent boundary

Local validation: 205 focused tests passed; Ruff and mypy passed.

rogercloud (approved):
## PR summary

PR #1269 "fix: reject empty final answers" closes a hole in the ReAct pattern where a `final_answer` call with an empty, whitespace-only, or non-string `answer` was stringified into an empty string and treated as a completed run, leaving the user with a backfilled "No output provided". The fix validates at the single choke point `_handle_control_tool`: an invalid answer feeds a failed tool result back to the model for one corrective turn, and a second invalid answer (or budget exhaustion after the repair grant) terminates the run with the new allowlisted failure code `invalid_final_answer`, propagated through the adapter/classifier chain and translated to `missing_delegated_output` at the delegated-agent boundary.

Blocking: yes (one major finding) — submitted as APPROVE with comments per maintainer direction.

## Approach verdict

Sound (Round 0). Validation at the control-tool choke point, the bounded one-repair loop, the dedicated failure code, and the delegation-boundary translation hold together; producers and consumers were traced end to end, and every prior finding's fix was verified truthful against current code.

## Update summary

The previous review was against 75f2b43c (force-pushed away). The branch was rewritten into three commits (6c85887 "fix: reject empty final answers" → 343c3c2 "fix: classify empty final answers" → 43426cd "fix: retry invalid final answers once"). Head 43426cd restores the bounded in-turn repair the review's F1 asked for (first invalid answer → failed tool result fed back, loop continues; second invalid or budget exhaustion → `invalid_final_answer`), introduces the dedicated `INVALID_FINAL_ANSWER_CODE` with adapter normalization and boundary translation (F2), adds the streamer fail branch (F3), builds the terminal result via `PatternResult` (F5), uses friendly error text plus ledger error passthrough (F6), clears pending tool calls before the checkpoint with a regression test (F7), expands the parametrize matrix (F8), renames the predicate (F9), and adds schema `minLength: 1` (F10). Every fix the author claimed in the replies was verified truthful against current code.

## Prior findings checklist

- P1 (was major: terminal failure vs PR description) — FIXED: bounded repair restored; tests cover retry (2 calls), double-invalid, and max_iterations=1.
- P2 (missing_delegated_output mislabel) — FIXED: dedicated INVALID_FINAL_ANSWER_CODE with documented boundary translation; the code no longer leaks at top level.
- P3 (dangling stream) — FIXED: fail branch added in _finish_streamed_answer_if_final (react.py:742-751), covered by test_react_closes_blank_final_answer_stream_before_retry; the residual never-started case is tracked as N3(a).
- P4 (trailing pending cleared without ledger records) — NOT FIXED as a consistency note: siblings still get no cancelled ledger record and no backfilled tool result; the original "inert" judgment is obsolete because the run now continues — consequences tracked as N1 (feedback dropped by the sanitizer) and N4 (stale cache).
- P5 (hand-built dict) — FIXED: PatternResult(...).to_dict() used in _invalid_final_answer_result.
- P6 (jargon error text + ledger error=None) — FIXED: friendly text and error= passed to _record_tool_call; asserted in test_react.py:3265.
- P7 (checkpoint before pending cleared) — FIXED: clear at react.py:1736 precedes the checkpoint; regression test asserts checkpoint pending == [].
- P8 (coverage gaps) — FIXED: parametrize matrix covers {} / "" / whitespace / 42 / missing-answer-with-valid-outcome / non-dict; sibling, adapter, classifier, and allowlist tests added.
- P9 (predicate name) — FIXED: renamed to _get_valid_final_answer with docstring.
- P10 (schema minLength) — FIXED: minLength: 1 added and asserted in test.

## New findings

### N1 (major, blocking) — src/xagent/core/agent/pattern/react/react.py:1736 — repair turn goes blind when siblings follow the invalid final_answer

When the model emits an invalid final_answer followed by sibling tool calls, `pending_tool_calls = []` here discards the siblings without backfilling any tool messages into context. The retry prompt then holds only `assistant(tool_calls=[A,B])` + `tool(A-failure)`; ExecutionContext._sanitize_tool_message_pairs (context/execution.py:1410-1451) requires expected_ids ⊆ received_ids and drops the entire assistant+tool block, so the corrective feedback ("The final response was empty or invalid. Call final_answer again with a non-empty string in answer.") never reaches the model. The retry becomes a blind re-run, and a second invalid answer hard-fails exactly where commit 43426cd intended informed correction. The single-call invalid path is unaffected (the repair works). The PR's own test (test_react_invalid_final_answer_clears_trailing_pending_before_checkpoint) uses FakeLLM, which ignores messages, so it cannot detect this.

Suggestion: backfill cancelled-style tool results for the discarded siblings (mirror _discard_pending_tool_plan_after_pause, react.py:1960-1981) so the block survives sanitization, and add an assertion that the retry call's messages contain the failure feedback.

### Minor

- N2 — react.py:2226-2228 — the `control_result.get("success") is False` exit is wider than its only reachable producer (invalid final_answer; both waiting_for_user producers are intercepted earlier). Load-bearing today, but narrowing to `failure_code == INVALID_FINAL_ANSWER_CODE` costs nothing and removes future silent-discard risk for other control tools.
- N3 — react.py:740-751 — (a) an empty-payload invalid final_answer never starts the streamer (started=False, early return at 740-741): no stream events, though the run still fails visibly via the pattern result — pre-existing early return, not PR-introduced; noted for awareness. (b) the "invalid final answer after retry" branch (748-749) has no test coverage; it is reachable in production (second streamed whitespace answer). Add a streaming double-invalid test.
- N4 — react.py:1736 — the discarded siblings' cached content in pending_tool_call_content is never forgotten; stale entries are serialized into every checkpoint (1084). Inert today but accumulating; _discard_pending_tool_plan_after_pause forgets via _backfill_result — reuse it.
- N5 — react.py:747-751 — the fail message is chosen solely from the repair flag: on the last iteration (e.g. max_iterations=1) with a first invalid answer, the stream says "invalid final answer, retrying" but the run terminates immediately with no retry. Pick the message via `self.current_iteration + 1 < self.max_iterations`.
- N7 — react.py:2683-2691 — _get_valid_final_answer validates answer.strip() but returns unstripped; padded answers pass validation and reach user output verbatim (pre-existing). One-line fix: return answer.strip().
- N8 — result.py:12-19 — a pattern-level code sits in the tool-level TOOL_FAILURE_CODES allowlist, and normalize_tool_failure_code's docstring still says "tool failure code". No behavioral defect today (membership is required for the classifier chain); hygiene: separate allowlist or rename + docstring update.
- N9 — agent_tool.py:1654-1659 — the invalid_final_answer → missing_delegated_output translation fires only for react-pattern children; DAG children fold step failure into their own PatternResult without failure_code (dag.py:1087-1089, _fail 1243-1272), so the parent lands in the generic branch. Fails closed; classification-specificity gap only. Acceptable as-is or follow-up.
- N10 — agent_tool.py:1668 — the generic-branch failure_code pass-through is PR-introduced dead code: the only pattern-level top-level producer (react.py:2705) is intercepted before the generic branch, and oauth_token_required etc. are never lifted into child pattern results. Untested. Delete, or keep as a documented seam with a test.
- N11 — react.py:1709-1742 — behavior change beyond "reject empty": non-string answers ({"answer": 42}, {"answer": null}) were stringified in base and completed; now they fail after one retry. Intentional, schema-consistent (type: string), test-pinned. Flagging the scope widening; fine if intentional.
- N12 — react.py:2693-2706 — failure_code is injected after to_dict() rather than via metadata — the only pattern result whose code lives outside metadata. Convention split vs the completion_outcome mirroring in execution_adapter. Fold into metadata or document.
- N13 — react.py:687-694 — sticky classification: any run that consumed its repair grant reports invalid_final_answer on iteration exhaustion, even if the model did productive work after the repair; this masks the max_iterations cause. The author announced this behavior deliberately, and the outcome is identical with literally-true error text; at the delegation boundary it is doubly misleading for budget-exhausted children (maps to missing_delegated_output). Consider keeping max_iterations classification when later iterations were productive.
- N15 — react.py:1080/1149-1151 — invalid_final_answer_repair_used is serialized and restored, but there is no load_state → run() resume round-trip test; dropping the restore would silently re-grant the repair. Add a resume test: checkpoint after repair granted → load_state → second invalid → assert INVALID_FINAL_ANSWER_CODE with no second repair.
- N16 — PR description — the body says "Retry the ReAct turn instead of marking it completed with an empty response" but does not state that the retry is bounded to one attempt, nor the new invalid_final_answer code, its four-file propagation, or the user-visible FAILED state. The thread comment documents it; the description itself is stale. Update it.

- react.py:L1755: delete `if answer:` — always True after the _get_valid_final_answer non-None guarantee; call context.add_assistant_message(answer) unconditionally.
- react.py:L2227: delete `self.pending_tool_calls = []` — no-op: invalid final_answer already cleared pending at L1736 before both return paths; the only reachable producer of success=False is that path. Keep the L2209 [1:] slice (required for send_message/ask_user_question).

net: -2 lines possible

- src/xagent/core/agent/pattern/react/react.py:1736 | major | repair turn goes blind: sibling tool calls discarded without backfill, sanitizer drops the assistant+tool block, corrective feedback never reaches the model | [new]

Only N1 is blocking; all prior findings are FIXED (P4 is a non-blocking consistency note folded into N1/N4) and the remaining minors do not block. Submitted as APPROVE with comments per maintainer direction.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1683,7 +1707,39 @@ async def _handle_control_tool(
...
+                )
+                context.add_tool_result(
+                    tool_name=name,
+                    result=tool_result,
+                    tool_call_id=tool_call.get("id"),
+                )
+                # A control call owns the segment even when invalid. Discard any
+                # trailing siblings before the checkpoint so resume cannot run
+                # calls that the model placed after final_answer.
+                self.pending_tool_calls = []
```
rogercloud: Blocking (major): the repair turn is blind when the model emits an invalid final_answer with trailing sibling tool calls. `pending_tool_calls = []` here discards the siblings without backfilling tool messages into context, so the retry prompt holds only assistant(tool_calls=[A,B]) + tool(A-failure). ExecutionContext._sanitize_tool_message_pairs (context/execution.py:1410-1451) requires expected_ids ⊆ received_ids and drops the entire assistant+tool block — the corrective feedback ("The final response was empty or invalid. Call final_answer again with a non-empty string in answer.") never reaches the model. The retry becomes a blind re-run, and a second invalid answer hard-fails exactly where this commit intended informed correction; the single-call invalid path is unaffected. The PR's own test (test_react_invalid_final_answer_clears_trailing_pending_before_checkpoint) uses FakeLLM, which ignores messages, so it cannot detect this. Suggest: backfill cancelled-style tool results for the discarded siblings (mirror _discard_pending_tool_plan_after_pause, react.py:1960-1981) so the block survives sanitization, and assert that the retry call's messages contain the failure feedback.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -2167,6 +2223,9 @@ async def _execute_pending_tool_calls(
                     if control_result.get("status") == "waiting_for_user":
                         self._discard_pending_tool_plan_after_pause(context)
                         return control_result
+                    if control_result.get("success") is False:
```
rogercloud: Minor: this `success is False` exit is wider than its only reachable producer — the invalid final_answer path (both waiting_for_user producers are intercepted earlier). It is load-bearing today (without it a second invalid answer would continue the loop), but narrowing the condition to `failure_code == INVALID_FINAL_ANSWER_CODE` costs nothing and removes the risk of silently discarding pending tool calls if another control tool ever returns success=False.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -728,9 +739,16 @@ async def _finish_streamed_answer_if_final(
...
-        if final_answer is not None and len(tool_calls) == 1:
-            await answer_streamer.finish(final_answer)
+        if len(tool_calls) == 1 and tool_calls[0].get("name") == "final_answer":
+            final_answer = self._final_answer_tool_content(tool_calls)
+            if final_answer is not None:
+                await answer_streamer.finish(final_answer)
+            else:
+                await answer_streamer.fail(
+                    "invalid final answer after retry"
+                    if self.invalid_final_answer_repair_used
```
rogercloud: Minor: the fail message is chosen solely from the repair flag. On the last iteration (e.g. max_iterations=1) with a first invalid answer, the stream says "invalid final answer, retrying" but the run terminates immediately with no retry. Consider picking the message via `self.current_iteration + 1 < self.max_iterations`.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -2620,6 +2679,32 @@ def _final_answer_outcome(value: Any) -> str:
...
 
+    @staticmethod
+    def _get_valid_final_answer(args: Any) -> str | None:
+        """Return a non-blank string answer from normalized tool arguments."""
+
+        if not isinstance(args, dict):
+            return None
+        answer = args.get("answer")
+        if isinstance(answer, str) and answer.strip():
+            return answer
```
rogercloud: Minor: validation checks `answer.strip()` but returns the unstripped value, so padded answers pass validation and reach user output verbatim (pre-existing behavior, not introduced here). One-line fix: `return answer.strip()`.

src/xagent/core/tools/adapters/vibe/agent_tool.py
```diff
@@ -1654,7 +1665,7 @@ def _classify_delegated_child_failure(
             message = output_text
         else:
             message = _GENERIC_CHILD_FAILURE_MESSAGE
-        return _classified_failure(message)
+        return _classified_failure(message, failure_code=failure_code)
```
rogercloud: Minor: this generic-branch failure_code pass-through is dead code introduced by this PR — the only pattern-level top-level producer (react.py:2705) is intercepted above at the INVALID_FINAL_ANSWER_CODE branch, and oauth_token_required etc. are never lifted into child pattern results. It is untested. Delete it, or keep it as a documented seam and add a test that pins the pass-through.

src/xagent/core/agent/pattern/react/react.py
qinxuye: Addressed by rebasing onto upstream #1325 and dropping the old control-handler retry implementation. The current path rejects the entire invalid response before it is persisted, so neither the empty `final_answer` nor its sibling work call can leave an unmatched assistant tool-call block. Commit `00ac6e2c` adds `test_react_discards_empty_final_answer_with_trailing_work_call`, which verifies the sibling is not executed, the retry receives the explicit empty-answer correction, and every persisted assistant tool-call ID has a corresponding tool result.

src/xagent/core/agent/pattern/react/react.py
qinxuye: Addressed. The PR-specific broad `success is False` exit was removed during the rebase. Empty answers now use upstream's response-level protocol retry path, so this generic control-result branch is no longer part of the diff.

src/xagent/core/agent/pattern/react/react.py
qinxuye: Addressed by removing the old streamer branch during the rebase. Upstream's protocol repair always performs the bounded in-turn retry before reporting `invalid_tool_protocol`, including when `max_iterations=1`, so the stale `invalid final answer, retrying` selection no longer exists.

src/xagent/core/agent/pattern/react/react.py
qinxuye: The PR-specific `_get_valid_final_answer` helper was removed during the rebase. Current behavior is upstream's `_final_answer_text` contract, which intentionally preserves nonblank answer content (including surrounding whitespace) and has an explicit regression assertion for that behavior. This PR no longer introduces a validate-with-strip/return-unstripped mismatch.

src/xagent/core/tools/adapters/vibe/agent_tool.py
qinxuye: Addressed. The generic `failure_code` pass-through and its import were removed; delegated failures now retain upstream's existing `_child_never_answered` classification and generic fallback behavior.

PR #1272, opened 2026-08-11
fix(browser): stop hardcoding zh-CN for browser automation locale

yiboyasss:
## Summary
- Playwright browser automation sessions were hardcoded to `locale="zh-CN"` / `timezone_id="Asia/Shanghai"` and spoofed `navigator.languages` as Chinese-first, so every task's browser tool requested Chinese-localized pages regardless of who was asking (many sites auto-localize based on this). Combined with a sandbox image missing Chinese fonts, this rendered as garbled tofu characters in screenshots.
- Locale is now derived from the account's own `app_locale` cookie (set by the frontend's language switcher), threaded through `WebToolConfig.get_browser_locale()` into `BrowserTaskSessionMixin` so whichever browser tool call happens to create the session carries it. Falls back to a deployment-level default (`XAGENT_BROWSER_TOOL_DEFAULT_LOCALE`, default `en-US`) when unresolvable.
- Also fixes `create_task()` passing the parsed `TaskCreateRequest` body (no cookies) into `get_agent_manager()` instead of the real `Request` — the cookie could never have been read otherwise. The three WebSocket call sites now pass their live connection object for the same reason.

## Test plan
- [x] `pytest tests/core/tools/test_browser_use.py tests/core/test_config.py::TestBrowserToolDefaultLocaleConfig tests/web/api/test_task_runtime_delete_bindings.py` — 29 passed
- [x] `ruff format` / `ruff check` / `mypy` all pass (pre-commit hook)
- [x] Local end-to-end verification: built `frontend/out` and served it same-origin from the backend, switched the UI language to Chinese, created a new task opening the HubSpot login page — logs confirmed `Creating browser session ... locale='zh-CN'` and the screenshot rendered correctly in Chinese; repeated after switching back to English and got `locale='en-US'`

### DIFF
--- a/example.env
+++ b/example.env
@@ -426,9 +426,12 @@ XAGENT_WEB_SEARCH_PROVIDER="auto"
 # XAGENT_BROWSER_CUA_DRIVER_MAX_ELEMENTS="2000"
 
 # Fallback locale/timezone for browser_use (Playwright) sessions, used when a
-# task carries no account-derived locale (the app_locale cookie set by the web
-# UI's language switcher). Locale defaults to "en-US"; timezone defaults to
-# the host's own system timezone when unset.
+# task carries no resolvable locale of its own (the app_locale cookie set by
+# the web UI's language switcher). Locale defaults to "en-US"; timezone
+# defaults to the host's own system timezone when unset. An invalid value
+# raises on first browser-tool use (not at process startup, and not silently
+# ignored) -- check the error against the tool call that triggered it, since
+# the validation itself lives in src/xagent/config.py, not here.
 # XAGENT_BROWSER_TOOL_DEFAULT_LOCALE="en-US"
 # XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE=""
 
--- a/example.env
+++ b/example.env
@@ -425,6 +425,13 @@ XAGENT_WEB_SEARCH_PROVIDER="auto"
 # menus without sending the full accessibility tree to the model.
 # XAGENT_BROWSER_CUA_DRIVER_MAX_ELEMENTS="2000"
 
+# Fallback locale/timezone for browser_use (Playwright) sessions, used when a
+# task carries no account-derived locale (the app_locale cookie set by the web
+# UI's language switcher). Locale defaults to "en-US"; timezone defaults to
+# the host's own system timezone when unset.
+# XAGENT_BROWSER_TOOL_DEFAULT_LOCALE="en-US"
+# XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE=""
+
 # Context compaction: the conversation is compacted when the estimated context
 # size reaches (model context window * ratio). Set a model's context window in
 # the model config; models without one fall back to the default below.
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -687,6 +687,9 @@ def get_native_browser_app_name() -> str:
     return canonical
 
 
+_BCP47_LOCALE_RE = re.compile(r"^[a-zA-Z]{2,3}(-[a-zA-Z0-9]{2,8})*$")
+
+
 def get_browser_tool_default_locale() -> str:
     """Get the fallback Playwright context locale for the browser_use tool.
 
@@ -699,8 +702,21 @@ def get_browser_tool_default_locale() -> str:
     Priority:
         1. XAGENT_BROWSER_TOOL_DEFAULT_LOCALE environment variable
         2. "en-US"
+
+    Raises:
+        ValueError: if the env var is set but isn't a plausible BCP-47 tag
+            (e.g. "en-US"), so a typo fails loudly at read time instead of
+            surfacing as an opaque Playwright error at session creation.
     """
-    return os.getenv(BROWSER_TOOL_DEFAULT_LOCALE, "").strip() or "en-US"
+    configured = os.getenv(BROWSER_TOOL_DEFAULT_LOCALE, "").strip()
+    if not configured:
+        return "en-US"
+    if not _BCP47_LOCALE_RE.match(configured):
+        raise ValueError(
+            f"{BROWSER_TOOL_DEFAULT_LOCALE} must be a BCP-47 locale tag "
+            f"(e.g. 'en-US', 'zh-CN'), got {configured!r}"
+        )
+    return configured
 
 
 def get_browser_tool_default_timezone() -> str | None:
@@ -709,8 +725,26 @@ def get_browser_tool_default_timezone() -> str | None:
     Priority:
         1. XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE environment variable
         2. None (Playwright falls back to the host's own system timezone)
+
+    Raises:
+        ValueError: if the env var is set but isn't a recognized IANA
+            timezone name (e.g. "Asia/Shanghai"), so a typo fails loudly at
+            read time instead of surfacing as an opaque Playwright error at
+            session creation.
     """
-    return os.getenv(BROWSER_TOOL_DEFAULT_TIMEZONE, "").strip() or None
+    configured = os.getenv(BROWSER_TOOL_DEFAULT_TIMEZONE, "").strip()
+    if not configured:
+        return None
+    from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
+    try:
+        ZoneInfo(configured)
+    except (ZoneInfoNotFoundError, ValueError, KeyError) as exc:
+        raise ValueError(
+            f"{BROWSER_TOOL_DEFAULT_TIMEZONE} must be a valid IANA timezone "
+            f"name (e.g. 'Asia/Shanghai'), got {configured!r}"
+        ) from exc
+    return configured
 
 
 def get_browser_cua_driver_command() -> str:
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -705,8 +705,10 @@ def get_browser_tool_default_locale() -> str:
 
     Raises:
         ValueError: if the env var is set but isn't a plausible BCP-47 tag
-            (e.g. "en-US"), so a typo fails loudly at read time instead of
-            surfacing as an opaque Playwright error at session creation.
+            (e.g. "en-US"). This getter is called lazily, from
+            BrowserSession.__init__ on first browser tool use rather than at
+            process startup, so a typo still fails as a clean tool-call
+            error instead of an opaque Playwright error at session creation.
     """
     configured = os.getenv(BROWSER_TOOL_DEFAULT_LOCALE, "").strip()
     if not configured:
@@ -728,9 +730,9 @@ def get_browser_tool_default_timezone() -> str | None:
 
     Raises:
         ValueError: if the env var is set but isn't a recognized IANA
-            timezone name (e.g. "Asia/Shanghai"), so a typo fails loudly at
-            read time instead of surfacing as an opaque Playwright error at
-            session creation.
+            timezone name (e.g. "Asia/Shanghai"). Like
+            get_browser_tool_default_locale, this is read lazily on first
+            browser tool use, not at process startup.
     """
     configured = os.getenv(BROWSER_TOOL_DEFAULT_TIMEZONE, "").strip()
     if not configured:
--- a/src/xagent/config.py
+++ b/src/xagent/config.py
@@ -52,6 +52,8 @@ UPLOADED_FILE_RECOVERY_BATCH_SIZE = "XAGENT_UPLOADED_FILE_RECOVERY_BATCH_SIZE"
 STORAGE_ROOT = "XAGENT_STORAGE_ROOT"
 NATIVE_BROWSER_ENABLED = "XAGENT_NATIVE_BROWSER_ENABLED"
 NATIVE_BROWSER_APP_NAME = "XAGENT_NATIVE_BROWSER_APP_NAME"
+BROWSER_TOOL_DEFAULT_LOCALE = "XAGENT_BROWSER_TOOL_DEFAULT_LOCALE"
+BROWSER_TOOL_DEFAULT_TIMEZONE = "XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE"
 BROWSER_CUA_DRIVER_COMMAND = "XAGENT_BROWSER_CUA_DRIVER_COMMAND"
 BROWSER_CUA_DRIVER_SOCKET = "XAGENT_BROWSER_CUA_DRIVER_SOCKET"
 BROWSER_CUA_DRIVER_TIMEOUT_SECONDS = "XAGENT_BROWSER_CUA_DRIVER_TIMEOUT_SECONDS"
@@ -683,6 +685,32 @@ def get_native_browser_app_name() -> str:
             f"{NATIVE_BROWSER_APP_NAME} must name a supported browser: {supported}"
         )
     return canonical
+
+
+def get_browser_tool_default_locale() -> str:
+    """Get the fallback Playwright context locale for the browser_use tool.
+
+    Used when a task/request carries no resolvable locale of its own (see
+    ``WebToolConfig.get_browser_locale``). Was previously hardcoded to
+    ``"zh-CN"``, which forced every automated browser session -- regardless
+    of the requesting user's own language -- to request and render
+    Chinese-localized pages.
+
+    Priority:
+        1. XAGENT_BROWSER_TOOL_DEFAULT_LOCALE environment variable
+        2. "en-US"
+    """
+    return os.getenv(BROWSER_TOOL_DEFAULT_LOCALE, "").strip() or "en-US"
+
+
+def get_browser_tool_default_timezone() -> str | None:
+    """Get the fallback Playwright context timezone for the browser_use tool.
+
+    Priority:
+        1. XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE environment variable
+        2. None (Playwright falls back to the host's own system timezone)
+    """
+    return os.getenv(BROWSER_TOOL_DEFAULT_TIMEZONE, "").strip() or None
 
 
 def get_browser_cua_driver_command() -> str:
--- a/src/xagent/core/tools/adapters/vibe/browser_use.py
+++ b/src/xagent/core/tools/adapters/vibe/browser_use.py
@@ -334,11 +334,15 @@ class BrowserNavigateTool(BrowserTaskSessionMixin, AbstractBaseTool):
     """Navigate to a URL in a browser session."""
 
     def __init__(
-        self, task_id: Optional[str] = None, workspace: Optional["TaskWorkspace"] = None
+        self,
+        task_id: Optional[str] = None,
+        workspace: Optional["TaskWorkspace"] = None,
+        locale: Optional[str] = None,
     ):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
         self._workspace = workspace
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -500,9 +504,10 @@ class BrowserClickTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Click an element on the current page."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -548,9 +553,10 @@ class BrowserFillTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Fill an input field with text."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -597,11 +603,15 @@ class BrowserScreenshotTool(BrowserTaskSessionMixin, AbstractBaseTool):
     """Take a screenshot of the current page."""
 
     def __init__(
-        self, task_id: Optional[str] = None, workspace: Optional["TaskWorkspace"] = None
+        self,
+        task_id: Optional[str] = None,
+        workspace: Optional["TaskWorkspace"] = None,
+        locale: Optional[str] = None,
     ):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
         self._workspace = workspace
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -716,9 +726,10 @@ class BrowserExtractTextTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Extract text content from the page."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -763,9 +774,10 @@ class BrowserEvaluateTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Execute JavaScript code in the browser."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -870,9 +882,10 @@ class BrowserSelectOptionTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Select an option from a dropdown."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -919,9 +932,10 @@ class BrowserWaitForSelectorTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Wait for an element to appear on the page."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -967,9 +981,10 @@ class BrowserCloseTool(BrowserTaskSessionMixin, AbstractBaseTool):
     category = ToolCategory.BROWSER
     """Close a browser session and free resources."""
 
-    def __init__(self, task_id: Optional[str] = None):
+    def __init__(self, task_id: Optional[str] = None, locale: Optional[str] = None):
         self._visibility = ToolVisibility.PRIVATE  # Internal tool
         self._task_id = task_id
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -1011,11 +1026,15 @@ class BrowserPdfTool(BrowserTaskSessionMixin, AbstractBaseTool):
     """Save current page as PDF."""
 
     def __init__(
-        self, task_id: Optional[str] = None, workspace: Optional["TaskWorkspace"] = None
+        self,
+        task_id: Optional[str] = None,
+        workspace: Optional["TaskWorkspace"] = None,
+        locale: Optional[str] = None,
     ):
         self._visibility = ToolVisibility.PUBLIC
         self._task_id = task_id
         self._workspace = workspace
+        self._locale = locale
 
     @property
     def name(self) -> str:
@@ -1148,25 +1167,23 @@ def create_browser_tools(
     """
     from .computer import ComputerTool
 
+    # A browser session is created by the FIRST tool call that touches it
+    # (navigate is typical, but nothing guarantees it), so every tool needs
+    # locale passed explicitly rather than relying on just one of them
+    # (e.g. navigate) to carry it.
     tools = [
-        ComputerTool(task_id=task_id, workspace=workspace),
-        BrowserNavigateTool(task_id=task_id, workspace=workspace),
-        BrowserClickTool(task_id=task_id),
-        BrowserFillTool(task_id=task_id),
-        BrowserScreenshotTool(task_id=task_id, workspace=workspace),
-        BrowserExtractTextTool(task_id=task_id),
-        BrowserPdfTool(task_id=task_id, workspace=workspace),
-        BrowserEvaluateTool(task_id=task_id),
-        BrowserSelectOptionTool(task_id=task_id),
-        BrowserWaitForSelectorTool(task_id=task_id),
-        BrowserCloseTool(task_id=task_id),
+        ComputerTool(task_id=task_id, workspace=workspace, locale=locale),
+        BrowserNavigateTool(task_id=task_id, workspace=workspace, locale=locale),
+        BrowserClickTool(task_id=task_id, locale=locale),
+        BrowserFillTool(task_id=task_id, locale=locale),
+        BrowserScreenshotTool(task_id=task_id, workspace=workspace, locale=locale),
+        BrowserExtractTextTool(task_id=task_id, locale=locale),
+        BrowserPdfTool(task_id=task_id, workspace=workspace, locale=locale),
+        BrowserEvaluateTool(task_id=task_id, locale=locale),
+        BrowserSelectOptionTool(task_id=task_id, locale=locale),
+        BrowserWaitForSelectorTool(task_id=task_id, locale=locale),
+        BrowserCloseTool(task_id=task_id, locale=locale),
     ]
     if include_debug_tools:
         tools.append(BrowserListSessionsTool())
-    # A browser session is created by the FIRST tool call that touches it
-    # (navigate is typical, but nothing guarantees it), so every tool must
-    # carry the locale for _with_default_session to inject.
-    for tool in tools:
-        if isinstance(tool, BrowserTaskSessionMixin):
-            tool._locale = locale
     return tools
--- a/src/xagent/core/tools/adapters/vibe/browser_use.py
+++ b/src/xagent/core/tools/adapters/vibe/browser_use.py
@@ -38,6 +38,11 @@ class BrowserTaskSessionMixin:
     """Keeps browser tool default sessions aligned during task setup."""
 
     _task_id: Optional[str] = None
+    # Playwright context locale for sessions this task creates. Set once by
+    # ``create_browser_tools`` on every browser tool (not just navigate):
+    # a session springs into existence on the FIRST tool call that touches
+    # it, whichever tool that is, and the locale is frozen at that moment.
+    _locale: Optional[str] = None
 
     async def setup(self, task_id: Optional[str] = None) -> None:
         if task_id:
@@ -51,6 +56,8 @@ class BrowserTaskSessionMixin:
         step_id = updated.pop(_STEP_SESSION_ARG, None)
         if not updated.get("session_id") and self._task_id:
             updated["session_id"] = self._default_session_id(step_id)
+        if self._locale and "locale" not in updated:
+            updated["locale"] = self._locale
         return updated
 
     def _default_session_id(self, step_id: Any = None) -> str:
@@ -1122,6 +1129,7 @@ def create_browser_tools(
     workspace: Optional["TaskWorkspace"] = None,
     *,
     include_debug_tools: bool = False,
+    locale: Optional[str] = None,
 ) -> list:
     """
     Create all browser automation tools for a task.
@@ -1131,6 +1139,9 @@ def create_browser_tools(
         workspace: Optional workspace for saving screenshots
         include_debug_tools: Include browser session diagnostics. Disabled for
             normal agent runs so debug schemas do not consume model context.
+        locale: Playwright context locale for sessions this task creates
+            (e.g. derived from the requesting account's app_locale setting).
+            Falls back to the deployment default when not supplied.
 
     Returns:
         List of browser tool instances
@@ -1152,4 +1163,10 @@ def create_browser_tools(
     ]
     if include_debug_tools:
         tools.append(BrowserListSessionsTool())
+    # A browser session is created by the FIRST tool call that touches it
+    # (navigate is typical, but nothing guarantees it), so every tool must
+    # carry the locale for _with_default_session to inject.
+    for tool in tools:
+        if isinstance(tool, BrowserTaskSessionMixin):
+            tool._locale = locale
     return tools
--- a/src/xagent/core/tools/adapters/vibe/computer.py
+++ b/src/xagent/core/tools/adapters/vibe/computer.py
@@ -341,11 +341,23 @@ class ComputerTool(BrowserTaskSessionMixin, AbstractBaseTool):
         environment = self._environments.get(session_id)
         try:
             if environment is None:
-                environment = self._environment_factory(
-                    session_id=session_id,
-                    workspace=self._workspace,
-                    headless=self._headless,
-                )
+                factory_kwargs: dict[str, Any] = {
+                    "session_id": session_id,
+                    "workspace": self._workspace,
+                    "headless": self._headless,
+                }
+                # Only the default Playwright-backed factory understands
+                # `locale` -- the native-local-browser factory (a functools.partial
+                # around _authorized_native_browser_environment /
+                # NativeBrowserEnvironment) drives the user's own already-running
+                # browser and has no `locale`/`**kwargs` parameter, so passing it
+                # there would raise TypeError instead of silently doing nothing.
+                if (
+                    self._locale
+                    and self._environment_factory is BrowserComputerEnvironment
+                ):
+                    factory_kwargs["locale"] = self._locale
+                environment = self._environment_factory(**factory_kwargs)
                 self._environments[session_id] = environment
             current = environment.current_observation
             if current is None:
--- a/src/xagent/core/tools/adapters/vibe/computer.py
+++ b/src/xagent/core/tools/adapters/vibe/computer.py
@@ -352,11 +352,18 @@ class ComputerTool(BrowserTaskSessionMixin, AbstractBaseTool):
                 # NativeBrowserEnvironment) drives the user's own already-running
                 # browser and has no `locale`/`**kwargs` parameter, so passing it
                 # there would raise TypeError instead of silently doing nothing.
-                if (
-                    self._locale
-                    and self._environment_factory is BrowserComputerEnvironment
-                ):
-                    factory_kwargs["locale"] = self._locale
+                if self._locale:
+                    if self._environment_factory is BrowserComputerEnvironment:
+                        factory_kwargs["locale"] = self._locale
+                    else:
+                        logger.debug(
+                            "Dropping resolved locale %r for computer tool "
+                            "session %r: environment factory %r doesn't accept "
+                            "a locale kwarg",
+                            self._locale,
+                            session_id,
+                            self._environment_factory,
+                        )
                 environment = self._environment_factory(**factory_kwargs)
                 self._environments[session_id] = environment
             current = environment.current_observation
--- a/src/xagent/core/tools/adapters/vibe/computer.py
+++ b/src/xagent/core/tools/adapters/vibe/computer.py
@@ -196,6 +196,7 @@ class ComputerTool(BrowserTaskSessionMixin, AbstractBaseTool):
         *,
         task_id: str | None = None,
         workspace: TaskWorkspace | None = None,
+        locale: str | None = None,
         environment_factory: ComputerEnvironmentFactory = BrowserComputerEnvironment,
         environment_instructions: str | None = None,
         environment_label: str = "browser",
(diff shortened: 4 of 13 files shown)


src/xagent/web/api/chat.py
```diff
@@ -3987,7 +3996,13 @@ def _get_default_internal_model_ids() -> Dict[str, Optional[str]]:
...
             f"Setting LLM configuration for task {task.id} with llm_ids: {task_llm_ids_to_set}"
         )
-        get_agent_manager(request).set_task_llms(int(task.id), task_llm_ids_to_set, db)
+        # ``http_request`` (the real Starlette Request, with cookies/headers)
+        # -- not ``request`` (the parsed TaskCreateRequest body, which has
+        # neither) -- so WebToolConfig.get_browser_locale() can resolve the
+        # account's app_locale cookie once this task's tools get built.
+        get_agent_manager(http_request).set_task_llms(
+            int(task.id), task_llm_ids_to_set, db
+        )
```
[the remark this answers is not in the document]
yiboyasss: Confirmed this is real, and #1272 is what turns it from inert into consequential (first code to actually read `.request.cookies` for something that matters). Tracking separately as #1282 rather than fixing here, because:

- The suggested `contextvars` drop-in doesn't actually work in this codebase: `create_task()` (HTTP) and the code that builds a task's tools (`_handle_chat_message_unserialized` over the WebSocket, which is what actually consumes `.request` in practice) run as two independent ASGI invocations with no shared context tree. `contextvars` don't propagate across that gap, so swapping to one would make the WebSocket side always see the default — silently regressing the locale feature back to "always default" with no test to catch it, since that's exactly the boundary the current (admittedly racy) shared-instance-attribute approach happens to bridge.
- A correct fix needs the state scoped per `task_id` (not a single global slot, not an ambient contextvar), which is a small design task on its own — bounding/cleanup alongside `AgentServiceManager._agents`, deciding behavior for tasks resumed with no live request, etc. Details in #1282.

Not blocking this PR: the race predates #1272 (this singleton's `.request` was already being overwritten across requests before this change, just with nothing meaningful reading it). #1272 doesn't widen the exposure surface beyond `app_locale`, and fixing #1282 properly benefits from its own review rather than being bundled into a browser-locale fix.

src/xagent/core/tools/adapters/vibe/browser_use.py
```diff
@@ -51,6 +56,8 @@ def _with_default_session(self, args: Mapping[str, Any]) -> dict[str, Any]:
         step_id = updated.pop(_STEP_SESSION_ARG, None)
         if not updated.get("session_id") and self._task_id:
             updated["session_id"] = self._default_session_id(step_id)
+        if self._locale and "locale" not in updated:
```
rogercloud, line 59: High / correctness (regression introduced by this PR): once `self._locale` is set, this injects a `locale` kwarg into **every** `BrowserTaskSessionMixin` tool call, including `BrowserCloseTool.run_json_async` (`adapters/vibe/browser_use.py:1002-1005`), which forwards it as `browser_close(**args)`. The core `browser_close` function (`src/xagent/core/tools/core/browser_use.py:1138`) has signature `async def browser_close(session_id: str)` — no `**kwargs` — unlike every other browser core function (`navigate`, `click`, `fill`, `screenshot`, `extract_text`, `evaluate`, `select_option`, `wait_for_selector`, `pdf`, which all accept `**kwargs`). Result: `TypeError: browser_close() got an unexpected keyword argument 'locale'` whenever a locale has been resolved for the task (reproduced directly).

This is reachable by the LLM in normal operation: `BrowserCloseTool` is registered unconditionally (unlike the debug-gated `BrowserListSessionsTool`), and `ToolVisibility`/`.visibility` is never read anywhere in production code to filter LLM tool exposure, so `PRIVATE` offers no protection here anyway. On the base commit, `_with_default_session` injected nothing, so the missing `**kwargs` on `browser_close` was harmless before this PR — this line is what turns it into a live bug.

Suggested fix: `async def browser_close(session_id: str, **_kwargs: Any)`, or pop `locale` in `BrowserCloseTool.run_json_async` before the call. Please also add a tool-layer test — the existing `browser_close` test calls the core function positionally and bypasses the tool layer where this bug lives.

src/xagent/core/tools/adapters/vibe/browser_use.py
```diff
@@ -1152,4 +1163,10 @@ def create_browser_tools(
     ]
     if include_debug_tools:
         tools.append(BrowserListSessionsTool())
+    # A browser session is created by the FIRST tool call that touches it
+    # (navigate is typical, but nothing guarantees it), so every tool must
+    # carry the locale for _with_default_session to inject.
+    for tool in tools:
+        if isinstance(tool, BrowserTaskSessionMixin):
+            tool._locale = locale
```
rogercloud: Medium-High / correctness: this broadcast assignment sets `._locale` on every `BrowserTaskSessionMixin` tool, but `ComputerTool` (`core/tools/adapters/vibe/computer.py:188`) overrides `run_json_async` (`computer.py:306-312`) and never calls `_with_default_session`, so `._locale` is dead there. Its actual session path (`self._environment_factory(...)` → `BrowserComputerEnvironment._get_page` → `core/computer/browser.py:171-174` → `manager.get_or_create(session_id, headless=...)`) passes neither `locale` nor `timezone_id`, and both of those files are untouched by this PR.

Since the hardcoded `zh-CN` default was removed elsewhere in this PR, `ComputerTool` sessions silently move from `zh-CN` to `en-US` for every account — including the Chinese-locale users this PR is meant to help — for a tool that is PUBLIC, first in the factory list, and asserted present in `default_names` by an existing test (`tests/core/agent/test_browser_tools.py:27`).

Please thread locale/timezone through `BrowserComputerEnvironment` into `manager.get_or_create(...)`, or route `ComputerTool.run_json_async` through `_with_default_session`. The comment above ("every tool must carry the locale") is also not accurate as written given this gap.

src/xagent/web/tools/config.py
```diff
@@ -1762,6 +1767,27 @@ def get_browser_tools_enabled(self) -> bool:
...
+        Accept-Language header (which reflects OS/browser settings, not a
+        deliberate account choice, and does not necessarily match what
+        language the account works in).
+
+        ``None`` (no request, no cookie, or a value the frontend doesn't
+        emit) lets the browser tool fall back to its own deployment
+        default.
+        """
+        cookies = getattr(self.request, "cookies", None)
+        app_locale = getattr(cookies, "get", lambda _key: None)("app_locale")
```
rogercloud: Medium / correctness (distinct from #1282): once `WebToolConfig.request` is nulled — which happens right after tools are built, via `_detach_factory_runtime_resources` setting `self.request = None` at `web/tools/config.py:2127` (pre-existing code, not part of this diff) — this line silently returns `None` here on, and the resolved account locale is lost. `AgentService` reuses the same `WebToolConfig` instance for a task's lifetime and can rebuild tools on it via `_ensure_tools_initialized` when the tool-policy signature changes; at that point this method falls back to the deployment default even though the task's first build resolved the correct locale.

This needs no concurrency and concerns this per-instance `WebToolConfig.request`, so it is genuinely separate from the process-wide `AgentServiceManager` singleton race tracked in #1282.

Suggested fix: cache the resolved locale as a plain string at first resolution (better: resolve once in `create_task` and pass it as an explicit parameter through `create_default_tools`, like `task_id`/`scope`/`vision_model`) instead of re-reading a request object that is deliberately detached after build. Please add a test asserting the locale survives a second `create_all_tools` call on the same config instance.

src/xagent/web/tools/config.py
```diff
@@ -1122,6 +1122,11 @@ def load(db: Any) -> Any:
     )
 
 
+# Maps the frontend's `app_locale` cookie (see i18n-context.tsx, which only
+# ever sets "en" or "zh") to a Playwright-compatible locale tag.
+_APP_LOCALE_TO_BROWSER_LOCALE = {"en": "en-US", "zh": "zh-CN"}
```
rogercloud, line 1128: Low / robustness: this map recognizes only exactly `"en"` and `"zh"`; any other cookie value — including an already-valid BCP-47 tag such as `zh-CN` or `en-GB` — silently yields `None` and falls through to the deployment default with no normalization. It also relies on a cookie that is only written when the user actively toggles the language switcher (`frontend/src/contexts/i18n-context.tsx`'s `setLocale`); the localStorage-restore effect that runs on page load does not rewrite it, so an account that chose a language long ago with a missing/expired cookie will see the UI in one language and browser automation in another.

Suggested fix: normalize and prefix-match the cookie value (accept `zh`, `zh-CN`, `zh_CN`, `en-GB`, ...), and have the frontend re-write the cookie when restoring the locale from localStorage.

src/xagent/web/api/chat.py
```diff
@@ -3987,7 +3996,13 @@ def _get_default_internal_model_ids() -> Dict[str, Optional[str]]:
         logger.info(
             f"Setting LLM configuration for task {task.id} with llm_ids: {task_llm_ids_to_set}"
         )
-        get_agent_manager(request).set_task_llms(int(task.id), task_llm_ids_to_set, db)
+        # ``http_request`` (the real Starlette Request, with cookies/headers)
+        # -- not ``request`` (the parsed TaskCreateRequest body, which has
+        # neither) -- so WebToolConfig.get_browser_locale() can resolve the
+        # account's app_locale cookie once this task's tools get built.
+        get_agent_manager(http_request).set_task_llms(
```
rogercloud, line 4003: Low / correctness: this is the only one of four `get_agent_manager(...)` call sites in this file that was updated to pass a real, typed `Request`. Three others — `delete_task`, `get_task_workspace_files`, and a third workspace endpoint (all further down in this file) — still declare `request: Any = None`, which FastAPI treats as a **query parameter**, and pass it straight into `get_agent_manager(request)`. A client can send `?request=somestring` on those endpoints and overwrite the singleton's `.request` with a plain string, silently defaulting the locale of any task subsequently rebuilt in that process. This is a type-confusion angle distinct from the concurrency race already tracked in #1282, though it touches the same singleton object.

Suggested fix: remove these query parameters from the three other endpoint signatures (use real `Request` injection where a request is genuinely needed there), or at minimum have `get_agent_manager` reject/ignore values that are not `Request` instances.

src/xagent/config.py
```diff
@@ -685,6 +687,32 @@ def get_native_browser_app_name() -> str:
     return canonical
 
 
+def get_browser_tool_default_locale() -> str:
```
rogercloud, line 693: Low / robustness: this getter (and its `get_browser_tool_default_timezone` sibling below) performs no validation — `XAGENT_BROWSER_TOOL_DEFAULT_LOCALE="not a locale!"` passes through unmodified into `new_context(locale=...)` and only fails later inside Playwright, with no attribution back to the misconfigured env var (verified dynamically). This is inconsistent with `get_native_browser_app_name()` just above (`config.py:675-687`), which validates/normalizes and raises a clear `ValueError`.

Suggested fix: validate the locale/timezone shape at read time and raise a `ValueError` naming the offending env var, so misconfiguration fails fast at startup rather than surfacing as an opaque Playwright error at session-creation time.

src/xagent/core/tools/core/browser_use.py
```diff
@@ -215,32 +251,62 @@ def __init__(self, timeout_minutes: int = 30):
...
         Args:
             session_id: Unique session identifier
             headless: Whether to use headless mode (only used when creating new session)
+            locale: Playwright context locale (only used when creating new session)
+            timezone_id: Playwright context timezone (only used when creating new session)
 
         Returns:
             BrowserSession instance
         """
+        import logging
```
rogercloud, line 269: Low / code quality: this local `import logging` + `logger = logging.getLogger(__name__)` shadows the module-level logger that already exists in this file. This exact redundancy was flagged on this PR by gemini-code-assist as medium priority, and the code is unchanged since that comment — no reply, no fix. Please remove the local import and reuse the module-level logger.

src/xagent/core/tools/core/browser_use.py
```diff
@@ -215,32 +251,62 @@ def __init__(self, timeout_minutes: int = 30):
...
             # Check if session already exists
             if session_id in self._sessions:
                 # Update last used time and return existing session
-                # Ignore headless parameter for existing sessions (can't change mode after creation)
-                self._sessions[session_id]._last_used = datetime.now()
-                return self._sessions[session_id]
+                # Ignore headless/locale/timezone_id for existing sessions
+                # (none of these can change after creation)
+                existing = self._sessions[session_id]
+                if locale and existing.locale != locale:
```
rogercloud, line 280: Simplification (`shrink`): these two near-identical blocks each log a "value frozen at creation, ignoring override" warning, one for `locale` and one for `timezone_id`. Looping over `[("locale", locale, existing.locale), ("timezone_id", timezone_id, existing.timezone_id)]` and emitting the warning from a single code path would preserve identical trigger conditions and message text, for roughly -6 lines. Note this keeps the timezone check that an earlier review asked for — it only collapses how it's coded.

rogercloud (changes requested):
## Summary

This PR removes the hardcoded `locale="zh-CN"` / `timezone_id="Asia/Shanghai"` from Playwright browser sessions and introduces deployment-level defaults via `XAGENT_BROWSER_TOOL_DEFAULT_LOCALE` (default `en-US`) and `XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE`, both documented in `example.env`. On top of that, it adds an attempt to derive a per-account browser locale from the `app_locale` cookie through a new `WebToolConfig.get_browser_locale()`, threaded via the process-wide `AgentServiceManager` singleton whose `.request` attribute is populated from `create_task()`'s newly injected `http_request: Request`. Note the PR description is stale: it states the three WebSocket call sites were updated to pass a live request object, but that change was reverted in `bec1ec61` — `websocket.py` has zero diff in the final PR, so the WebSocket-driven tool-building path never supplies a request.

## Approach

**The hardcode-removal half is the right fix and should be kept as-is.** Deleting the baked-in `zh-CN`/`Asia/Shanghai` and replacing it with an explicit deployment default is exactly the correct shape for the reported bug (garbled/tofu screenshots and locale-mismatched pages for non-Chinese users).

**The "derive locale from the account cookie" half is going in the wrong direction at the mechanism level.** Locale is the only piece of per-task configuration threaded through an implicit, process-wide mutable singleton (`AgentServiceManager._global_agent_manager.request`), overwritten by whichever caller last passed a non-None request. Every other piece of task config — `task_id`, `workspace_owner_id`, `scope`, `vision_model`, `llm` — is an explicit parameter to `create_default_tools`. It's worth noting that the `set_task_llms()` call this hangs off in `create_task()` is a no-op for brand-new tasks (its body is guarded by `if task_id in self._agents:`, and a just-created task cannot be there yet), so the *only* real effect of that call is the global side effect.

That single choice produces several independent consequences:

- **Lifetime mismatch.** The reader (`WebToolConfig`, which persists across a task's tool rebuilds) outlives the object it reads from (`Request`, request-scoped — the code even explicitly nulls `WebToolConfig.request` after each build via `_detach_factory_runtime_resources`). See F3 below.
- **Trust mismatch.** Existing code in the same file already treats `.request` as an untrusted/possibly-wrong-principal value requiring defensive handling (the three-state `is_admin` logic and its comments about not letting `request`'s admin status leak in). This PR is the first feature to treat it as an *authoritative* source for something that matters.
- **Untyped sibling call sites.** Only `create_task` was given a properly-typed `Request`; three other endpoints still declare `request: Any = None` as an untyped FastAPI *query* parameter and feed it straight into `get_agent_manager()`. See F6 below.
- **Diffuse responsibility.** Locale resolution is now spread across four layers (cookie→locale map in `web/tools/config.py`, env fallback in `src/xagent/config.py`, actual application in `core/tools/core/browser_use.py`, and a None-passthrough layer in `adapters/vibe/config.py`), each treating `None` as "ask the next layer" — which is what makes F3 fail silently rather than loudly.
- **Asymmetric knobs.** `XAGENT_BROWSER_TOOL_DEFAULT_TIMEZONE` has no per-request path at all (only `locale` is injected via `_with_default_session`), so timezone is env-only while locale is per-account. Also, silently dropping the old `Asia/Shanghai` default is a behavior change for existing Chinese deployments, documented only in `example.env`.
- **Incomplete reach.** The mechanism doesn't reach `ComputerTool`, arguably the most commonly available session-creating tool. See F4.

For clarity on scope: the *concurrency* angle of this singleton (cross-request cookie leakage under concurrent load) is already disclosed by the author and tracked separately in **#1282**, with an accurate explanation of why `contextvars` doesn't work in this ASGI topology. I am deliberately **not** relitigating that here as blocking. But F1, F3, F4 and F6 below are **not** covered by #1282 — they reproduce in a single-user, no-concurrency world and still need addressing.

A more minimal correct fix for the stated bug: resolve the locale once in `create_task` and persist it as a plain string on the task row / pass it as an explicit parameter through `create_default_tools(...)`, like every other piece of task config. No change to `get_agent_manager` semantics, no need to touch the WebSocket call sites, survives tool rebuilds / restarts / resumes, and cannot leak between users.

## Findings

### High

**F1 — `browser_close` crashes with `TypeError` once a locale is resolved (regression introduced by this PR).** See inline comment on `src/xagent/core/tools/adapters/vibe/browser_use.py`.

### Medium-High

**F4 — `ComputerTool` never applies the resolved locale, and its sessions silently flip from `zh-CN` to `en-US`.** See inline comment.

### Medium

**F3 — Resolved locale is silently lost on tool rebuild (distinct from #1282).** See inline comment.

**Test coverage gap.** The only new tests (`tests/core/test_config.py::TestBrowserToolDefaultLocaleConfig`) cover the two raw env-var getters. There is no coverage anywhere in `tests/` for `WebToolConfig.get_browser_locale()` (cookie present / absent / unrecognized / `self.request is None`), none for the factory's locale broadcast path (`create_browser_tools()` is never called with `locale=` in any test), and none for the F1 crash path. The "29 passed" claim is consistent with these gaps rather than evidence against them — those tests pass precisely because they don't touch the new paths. Please add: (1) a tool-layer `BrowserCloseTool` test with a locale set, (2) `get_browser_locale()` unit tests for all four branches, (3) a rebuild test asserting locale survives a second `create_all_tools` on the same `WebToolConfig`.

### Low

**F5 — The `app_locale` cookie is an unreliable source, and the mapping is too narrow.** See inline comment.

**F6 — Three endpoints still pass an untyped query parameter into `get_agent_manager()`.** See inline comment.

**F7 — New config getters perform no validation.** See inline comment.

**Q2 — Redundant local logger (previously flagged, unaddressed).** See inline comment.

### Documentation

`example.env` already documents both new env vars and is the only place in the repo documenting any `XAGENT_BROWSER_*` variable — good. One wording nit: the factory comment ("every tool must carry the locale") and the `example.env` phrasing overstate the actual reach of account-locale resolution given F3 and F4; consider softening until those are fixed.

`src/xagent/core/tools/core/browser_use.py`
- L283: shrink: two near-identical `if` blocks each log a "value frozen at creation, ignoring override" warning, one for `locale` and one for `timezone_id`. Loop over `[("locale", locale, existing.locale), ("timezone_id", timezone_id, existing.timezone_id)]` and emit the warning from a single code path — identical trigger conditions and message text are preserved. (This is orthogonal to the earlier reviewer request to *add* the timezone symmetry check — that check stays, only its coding collapses.)

net: -6 lines possible

## Verdict

**Request changes.** F1 is a confirmed, PR-introduced, LLM-reachable crash with zero test coverage, and the core "account-derived locale" mechanism does not reliably deliver the account locale (F3, F4, F6) even setting aside the concurrency issue already tracked in #1282. The hardcode-removal + env-default half is good and can land as-is; I'd suggest either fixing F1/F4 and reworking the cookie path into an explicit per-task parameter, or splitting the cookie-derived locale out into a follow-up PR.

src/xagent/core/tools/adapters/vibe/browser_use.py
yiboyasss: Confirmed and reproduced exactly as described. Fixed in 6b097c11: added `**_kwargs: Any` to `browser_close()`, matching every other browser_* core function. Added the same signature update to `FakeManager` in `tests/core/computer/test_browser_environment.py` while verifying — its old `get_or_create(session_id, headless)` fixed signature would have broken the same way once locale started flowing through `ComputerTool` (see reply on the other thread).

src/xagent/core/tools/adapters/vibe/browser_use.py
yiboyasss: Confirmed. Fixed in 6b097c11: threaded `locale` through `BrowserComputerEnvironment.__init__` into `manager.get_or_create(...)`, and `ComputerTool.run_json_async` now passes `locale=self._locale` when building the environment — but only when `self._environment_factory is BrowserComputerEnvironment` (the actual default). The native-local-browser factory (a `functools.partial` around `_authorized_native_browser_environment`/`NativeBrowserEnvironment`) has no `locale`/`**kwargs` parameter, so passing it unconditionally would have reproduced the exact same class of TypeError as the `browser_close` bug on the other thread — that's why this is gated on the factory identity rather than always passed.

src/xagent/web/tools/config.py
yiboyasss: Confirmed, and distinct from #1282 as you noted. Fixed in 6b097c11 with the simpler of your two suggested options: memoize the resolved locale on `WebToolConfig` at first resolution (a plain `_browser_locale_resolved`/`_cached_browser_locale` pair, since `None` is itself a valid resolved value and can't double as the "not yet computed" sentinel). Didn't do the explicit-parameter-through-create_task route since the memoization is a smaller, purely-local change with the same effect.

src/xagent/web/tools/config.py
yiboyasss: Fixed the normalization half in 6b097c11 — now matches on the primary language subtag (split on `-`/`_`), so `zh-CN`, `zh_CN`, `ZH` etc. all resolve the same as the exact `"zh"` the frontend writes today. Didn't change the frontend's localStorage-restore effect to rewrite the cookie — that's a separate, frontend-only timing fix (a stale/expired cookie surviving until the next active language-switcher toggle), tracking it as follow-up rather than bundling a TS change into this backend-focused pass.

src/xagent/config.py
yiboyasss: Fixed in 6b097c11: both getters now validate at read time and raise ValueError naming the offending env var — a regex check for BCP-47 shape on the locale, and zoneinfo.ZoneInfo(...) against the real IANA database for the timezone — matching get_native_browser_app_name()'s pattern as suggested.

src/xagent/core/tools/core/browser_use.py
yiboyasss: Applied as suggested in 6b097c11 — collapsed into a loop over (attr_name, requested, current) triples, same trigger conditions and message shape, just parameterized by field name.

src/xagent/core/tools/core/browser_use.py
```diff
@@ -215,32 +251,56 @@ def __init__(self, timeout_minutes: int = 30):
...
             headless: Whether to use headless mode (only used when creating new session)
+            locale: Playwright context locale (only used when creating new session)
+            timezone_id: Playwright context timezone (only used when creating new session)
 
         Returns:
             BrowserSession instance
         """
+        import logging
+
+        logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
yiboyasss: Checked this carefully (sorry for the delay replying — should have posted this the first time instead of only explaining it in a side conversation). This file has no module-level logger to be redundant against:

```
$ grep -n '^logger\|^import logging\|logging\.getLogger' src/xagent/core/tools/core/browser_use.py
274:        logger = logging.getLogger(__name__)
454:        logger = logging.getLogger(__name__)
489:        logger = logging.getLogger(__name__)
690:    logger = logging.getLogger(__name__)
722:        logger = logging.getLogger(__name__)
1233:        logger = logging.getLogger(__name__)
```

Every occurrence (five pre-existing, plus the one this PR adds) is a per-function/method `import logging; logger = logging.getLogger(__name__)` — that's this file's established (if repetitive) convention, not something this PR introduces or duplicates. Not changing it, since doing so would make this one spot inconsistent with the other five instead of the other way around.

src/xagent/core/tools/core/browser_use.py
yiboyasss: Replied on the original gemini thread with the grep evidence: this file has no module-level logger anywhere (5 pre-existing per-function `logging.getLogger(__name__)` calls, this PR's is the 6th, same pattern). Not a redundancy this PR introduces — not fixing.

src/xagent/web/api/chat.py
yiboyasss: Confirmed — verified the three other endpoints (delete_task, get_task_workspace_files, get_task_output_files) all still have `request: Any = None`, which FastAPI does treat as a query parameter, feeding straight into get_agent_manager(request). Not fixing in this PR: it's the same root cause as #1282 (the AgentServiceManager singleton's .request being writable by whatever caller passes one), just a different injection vector, and none of these three endpoints are touched by this PR's diff otherwise. Added a note to #1282 so it's tracked as part of that redesign rather than three one-off endpoint edits here.

src/xagent/core/tools/adapters/vibe/computer.py
```diff
@@ -341,11 +341,23 @@ async def run_json_async(self, args: Mapping[str, Any]) -> dict[str, Any]:
...
+                # Only the default Playwright-backed factory understands
+                # `locale` -- the native-local-browser factory (a functools.partial
+                # around _authorized_native_browser_environment /
+                # NativeBrowserEnvironment) drives the user's own already-running
+                # browser and has no `locale`/`**kwargs` parameter, so passing it
+                # there would raise TypeError instead of silently doing nothing.
+                if (
+                    self._locale
+                    and self._environment_factory is BrowserComputerEnvironment
+                ):
```
rogercloud: **M2 -- minor (latent today).** This gates locale hand-off on `self._environment_factory is BrowserComputerEnvironment` -- an object-identity check, not `isinstance`, with no logging when it fails. A subclass of `BrowserComputerEnvironment` or a legitimate Playwright-backed `functools.partial` wrapper would silently drop the locale with zero diagnostic, quietly reintroducing the original locale bug. Nothing in production or tests currently constructs such a factory (the only other real caller is the native-browser partial this check deliberately excludes), so it's correct today. Suggest adding a debug-level log when locale is dropped so a future factory addition surfaces rather than regresses silently.

src/xagent/web/tools/config.py
```diff
@@ -1762,6 +1775,49 @@ def get_browser_tools_enabled(self) -> bool:
...
+        Matches on the primary language subtag so an already-valid BCP-47-ish
+        variant (``zh-CN``, ``zh_CN``, ``EN``, ...) resolves the same as the
+        exact ``"en"``/``"zh"`` the frontend writes today (see
+        ``frontend/src/contexts/i18n-context.tsx``), instead of only matching
+        those two literal strings.
+        """
+        if not isinstance(app_locale, str) or not app_locale:
+            return None
+        primary = re.split(r"[-_]", app_locale, maxsplit=1)[0].lower()
+        return _APP_LOCALE_TO_BROWSER_LOCALE.get(primary)
```
rogercloud, line 1840: **M3 -- minor (latent today).** This matches on the primary language subtag only, so `zh-TW` and `zh-HK` would both map to `zh-CN` (Simplified), losing the Traditional/Simplified distinction. Harmless today since the frontend's `Locale` type is a hard `"en" | "zh"` union, so no UI path can write those values. Suggest a one-line comment recording that assumption so it's caught if a Traditional-Chinese locale is ever added.

src/xagent/config.py
```diff
@@ -685,6 +687,66 @@ def get_native_browser_app_name() -> str:
...
+    Chinese-localized pages.
+
+    Priority:
+        1. XAGENT_BROWSER_TOOL_DEFAULT_LOCALE environment variable
+        2. "en-US"
+
+    Raises:
+        ValueError: if the env var is set but isn't a plausible BCP-47 tag
+            (e.g. "en-US"), so a typo fails loudly at read time instead of
+            surfacing as an opaque Playwright error at session creation.
```
rogercloud: **M4 -- minor (docstring accuracy).** This says a misconfigured `XAGENT_BROWSER_TOOL_DEFAULT_LOCALE` "fails loudly at read time," but the getter is actually called lazily from `BrowserSession.__init__` during a live tool call (while `BrowserSessionManager._lock` is held), not at process startup. Not a robustness bug in practice -- the ReAct/DAG patterns' broad exception handlers convert it into a clean structured tool-call failure -- but the docstring should say "first-use time" or similar rather than implying a startup check.

src/xagent/web/api/chat.py
```diff
@@ -3651,6 +3651,15 @@ def _build_unique_workspace_target(base_dir: Path, filename: str) -> Path:
...
     request: TaskCreateRequest,
+    # FastAPI always injects the real Request for HTTP calls regardless of the
+    # default; the None default keeps direct (test) callers working. Must stay
+    # a bare `Request` annotation, not `Optional[Request]` -- FastAPI only
+    # recognizes the special-cased injected-Request parameter with the exact
+    # bare type; wrapping it in Optional makes FastAPI try to build a Pydantic
+    # field from it instead, which fails at route-registration time since
+    # Request isn't a valid Pydantic field type (verified: this reproduces a
+    # collection-time FastAPIError in every test that imports this module).
+    http_request: Request = None,  # type: ignore[assignment]
```
rogercloud, line 3662: **M5 -- minor (latent, no live bug today).** Now that a real Starlette `Request` is stored in the global `AgentServiceManager`, note that `Request.user` raises `AssertionError` (not `AttributeError`) when `AuthenticationMiddleware` isn't installed, which it isn't anywhere in this app -- so the `getattr(request, "user", None)` idiom at `web/tools/config.py:1257`/`:1320` would not safely no-op if ever reached with this object. Currently unreachable because `create_default_tools` always passes explicit `user=`/`is_admin=`, but worth a comment so a future caller doesn't trip it.

src/xagent/core/tools/core/browser_use.py
```diff
@@ -361,7 +432,9 @@ async def browser_navigate(**kwargs: Any) -> Dict[str, Any]:
         }
 
     manager = get_browser_manager()
-    session = await manager.get_or_create(session_id, headless)
+    session = await manager.get_or_create(
+        session_id, headless, locale=locale, timezone_id=timezone_id
```
rogercloud: **M6 -- minor.** Only `browser_navigate` threads `timezone_id` through to `get_or_create`; the other eight `browser_*` functions pass `locale` only, so a session first created by anything other than `browser_navigate` never gets an explicit timezone. Cosmetic today since nothing supplies a per-request timezone at all -- simplest fix is to drop the `timezone_id` parameter entirely rather than propagate it further (see the simplification note in the review body).

src/xagent/core/tools/core/browser_use.py
```diff
@@ -349,6 +418,8 @@ async def browser_navigate(**kwargs: Any) -> Dict[str, Any]:
         }
     headless = kwargs.get("headless", False)
     wait_until = kwargs.get("wait_until", "networkidle")
+    locale = kwargs.get("locale")
+    timezone_id = kwargs.get("timezone_id")
```
rogercloud: **Simplification (delete).** `timezone_id = kwargs.get("timezone_id")` always reads `None` in practice -- nothing in the repo ever sets this kwarg (no tool schema, no `_with_default_session` injection, no other caller). Suggest removing it along with the `timezone_id=timezone_id` argument passed to `get_or_create` a few lines below, and the `timezone_id` parameter on `get_or_create`/`BrowserSession.__init__` itself -- net roughly -6 to -10 lines with no behavior change.

rogercloud (changes requested):
## Summary

This PR removes the hardcoded `zh-CN` / `Asia/Shanghai` locale from Playwright browser-automation sessions, which was rendering "tofu" boxes in screenshots for non-Chinese users and spoofing `navigator.languages` as Chinese-first for everyone. The session locale is now derived per-account from the `app_locale` cookie written by the frontend language switcher, via a new `WebToolConfig.get_browser_locale()` threaded through `BrowserTaskSessionMixin` into `BrowserSession`, with a validated deployment-level fallback (`XAGENT_BROWSER_TOOL_DEFAULT_LOCALE`, defaulting to `en-US`). It also fixes `create_task()`, which was passing the parsed request body (no cookies) rather than the real `Request` object into `get_agent_manager()`.

This review covers commit `6b097c11`, which addresses most findings from earlier review rounds -- see "Already addressed" below for what was re-checked.

**Overall design verdict: acceptable-with-reservations.** The direction is right-sized and fits existing conventions (env getters sit alongside the sibling `get_native_browser_*` getters and are documented in `example.env`; the memoization idiom matches other `WebToolConfig` caches). Two body-level concerns remain, one of which I consider a merge blocker.

---

## Design-level findings

### D1 -- MAJOR (waived; tracked as #1282, not a merge blocker)

Locale resolution ultimately reads from process-global mutable state: the singleton `AgentServiceManager` (`src/xagent/web/api/chat.py:3621-3633`) stores `.request` with no per-request scoping. This PR is the first path where that becomes consequential -- `WebToolConfig.get_browser_locale()` now reads real cookie data off whatever `.request` was last assigned. Every WebSocket call site calls `get_agent_manager()` with no arguments, so a task's browser locale is effectively "whichever HTTP request most recently touched `get_agent_manager()`" rather than the task's own creator. Concretely: user B calling `create_task()` between user A's `create_task()` and A's first WebSocket turn can make A's browser session use B's `app_locale`.

Reasonable to defer: blast radius today is a single UI locale string (no auth/session data crosses users), the naive contextvar fix is documented as not dropping in cleanly (commit `bec1ec61` reverted it for breaking pause/resume tests), and there's no existing contextvar machinery elsewhere that would make deferral look like avoidance. Flagging prominently anyway since it's a real, confirmed cross-user leak, however narrow -- please make sure #1282 actually gets scheduled.

### D2 -- MAJOR (would like this resolved before merge)

Near-zero test coverage for the new locale-resolution logic. The only tests touching locale are three env-var cases in `tests/core/test_config.py:1752-1774` (default / override / blank-fallback) plus one incidental assertion in `tests/core/computer/test_browser_environment.py:207-209` that locale is `None`, which holds only because that fixture never passes one.

Nothing covers:
- `WebToolConfig.get_browser_locale()` / `_normalize_app_locale_cookie()` (`web/tools/config.py:1778-1819`) for any cookie value (missing, empty, `zh`, `zh-CN`, `ja`), or the memoization behavior after `self.request` is nulled.
- The `tool._locale = locale` stamping loop across `BrowserTaskSessionMixin` instances (`vibe/browser_use.py:1169-1171`) -- confirmed deleting it outright would not fail any existing test.
- `BrowserSession`'s locale/timezone fallback, or the `navigator.languages` spoof string construction.
- `get_or_create`'s new locale/`timezone_id` mismatch warning (`core/tools/core/browser_use.py:283-293`).
- `ComputerTool`'s factory identity gate (`vibe/computer.py:355-358`).
- Either new config getter's `ValueError` branch (`config.py:714-718`, `742-746`).

The PR's core question -- does the right browser session actually receive the right locale -- is currently unverified by CI, and the mechanism spans four layers, which is exactly where a silent regression would hide. A handful of unit tests over `get_browser_locale()`/`_normalize_app_locale_cookie()` plus one assertion that a stamped locale reaches `BrowserSession` would cover most of the risk.

### Also worth noting (no single line in this diff owns these -- both files are pre-existing, untouched by this PR)

- `src/xagent/core/tools/adapters/vibe/agent_tool.py:2183` and `src/xagent/web/api/agents.py:1648` -- delegated sub-agent tool configs and the agent-preview endpoint both build `WebToolConfig(request=MinimalRequest(...))`, and `MinimalRequest` has no `cookies` attribute. `get_browser_locale()` safely returns `None` (no crash) but silently falls back to the deployment default even when the parent task already resolved e.g. `zh-CN` -- locale is inconsistent within one logical task tree. Suggest tracking parent-to-child locale propagation alongside #1282.

---

## Already addressed from the prior review round

Re-checked against `6b097c11` and confirmed resolved -- listing explicitly so it's clear these weren't skipped:

- **Local `logging.getLogger(__name__)` in `get_or_create`** (`core/tools/core/browser_use.py:274`) -- the file has no module-level logger at all; six functions each declare their own and five of those predate this PR. The rebuttal is correct; no action needed.
- **`timezone_id` mismatch warning + duplicate block** (`:283-293`) -- fixed; a single loop over `(attr_name, requested, current)` triples now warns for both locale and `timezone_id` with the same semantics as the original locale-only check.
- **`browser_close()` missing `**kwargs`** (`:1137`, `vibe/browser_use.py:1002-1005`) -- fixed; `**_kwargs: Any` added to match every sibling function, `FakeManager` test double updated alongside.
- **`ComputerTool` never receiving locale** -- fixed for the default `BrowserComputerEnvironment` path. The native-local-browser factory still doesn't receive locale, but that's now an explicit, commented choice affecting only an admin-gated opt-in feature (`get_native_browser_enabled()` defaults `False`); general gate concern captured as M2 below.
- **Locale lost after `WebToolConfig.request` is nulled** -- fixed via memoization (`_browser_locale_resolved` / `_cached_browser_locale`); ordering confirmed correct.
- **Cookie normalization only matching exact `en`/`zh`** -- fixed; primary-subtag case-insensitive matching now handles `zh-CN` / `zh_CN` / `ZH`. The related frontend gap (localStorage-restore effect not rewriting a stale cookie) is genuinely still open, but deliberately so and tracked as a frontend-only follow-up.
- **Missing env-var validation** for `XAGENT_BROWSER_TOOL_DEFAULT_LOCALE` / `_TIMEZONE` -- fixed with a BCP-47 regex plus `zoneinfo` lookup, matching the sibling `get_native_browser_app_name()` pattern. Neither `ValueError` path has an automated regression test; folded into D2 rather than listed separately.
- **Other `get_agent_manager()` call sites** in `chat.py` (`delete_task`, `get_task_workspace_files`, `get_task_output_files`) still declaring `request: Any = None` -- same vulnerability class as D1, but pre-existing code untouched by this diff and explicitly acknowledged, deferring to #1282. No action needed in this PR.
- **`Optional[Request] = None` vs `Request = None  # type: ignore[assignment]`** (`chat.py:3662`) -- not changed as originally suggested, but the inline comment at `:3654-3661` explains why: wrapping in `Optional[Request]` makes FastAPI treat it as a Pydantic field, which fails at route-registration time. Accepting as correctly explained, not a defect.

---

```
delete: core/tools/core/browser_use.py -- `timezone_id` param on BrowserSessionManager.get_or_create / BrowserSession.__init__; never populated by any per-request value (only `locale` is app-derived). No test, schema, or roadmap comment depends on it. Keep timezone resolution as the get_browser_tool_default_timezone() fallback inside BrowserSession.__init__.
delete: core/tools/core/browser_use.py:~422 -- `timezone_id = kwargs.get("timezone_id")` in browser_navigate; always None in practice, nothing in the repo ever sets that kwarg. Remove it plus the `timezone_id=timezone_id` argument at the single get_or_create call site.
net: -6 to -10 lines possible, no behavior change for any current caller or test.
```

Also considered and **rejected**: replacing the `_browser_locale_resolved` / `_cached_browser_locale` bool+value pair with a single sentinel attribute. That two-field idiom (needed because `None` is itself a meaningful resolved value) is already established elsewhere in the same class (`_tool_allowlist_cached` / `_cached_tool_allowlist`), so changing it here alone would create local inconsistency for no net line savings.

---

## Recommended disposition: Request changes

D2 is a MAJOR finding that is genuinely unaddressed -- the PR's central mechanism has zero automated coverage, and deleting the stamping loop breaks no test. D1, by contrast, is appropriately waived with a documented reason and a tracking issue, and should not drive the disposition. Line-level findings below are all minor and none need to block. The ask is narrow: tests for `get_browser_locale()`/`_normalize_app_locale_cookie()` plus one end-to-end assertion that the stamped locale reaches `BrowserSession`, the M2 debug log, and the M4 docstring wording -- everything else can land as follow-ups.

src/xagent/core/tools/core/browser_use.py
```diff
@@ -215,32 +250,51 @@ def __init__(self, timeout_minutes: int = 30):
...
+                existing._last_used = datetime.now()
+                return existing
 
             # Start cleanup task on first session creation
             if self._cleanup_task is None:
                 self._cleanup_task = asyncio.create_task(self._cleanup_loop())
 
-            # Create new session with specified headless mode
-            self._sessions[session_id] = BrowserSession(session_id, headless)
+            logger.info(f"Creating browser session {session_id!r} (locale={locale!r})")
```
rogercloud: This logs the raw `locale` argument, which is `None` whenever the caller didn't resolve one — but the session two lines down is constructed with `locale or get_browser_tool_default_locale()`. So exactly in the case an operator most wants to see in the log (the deployment default kicking in), the line reads `locale=None`. Suggest moving the log after the `BrowserSession` construction and logging `self._sessions[session_id].locale`. Non-blocking.

src/xagent/core/tools/adapters/vibe/browser_use.py
rogercloud: This sets a private attribute on the tool instances from outside the class, while `task_id` and `workspace` — the same kind of per-task config, on these same tool classes — are real constructor parameters. Threading `locale` through the constructor too would keep the three consistent and make the dependency visible at the type level. Non-blocking.

src/xagent/web/tools/config.py
```diff
@@ -1762,6 +1785,56 @@ def get_browser_tools_enabled(self) -> bool:
         """Whether to include browser automation tools."""
         return self._browser_tools_enabled
 
+    def get_browser_locale(self) -> Optional[str]:
+        """Derive a browser-automation locale from the account's own
+        ``app_locale`` setting (the language picked via xagent's UI
```
rogercloud: "the account's own `app_locale` setting" / "the language the account actually chose to work in" overstates what this reads. There is no `locale` column on `User` and no `locale` field on `TaskCreateRequest` — this is a browser cookie, so it is device- and browser-scoped, lost when cookies are cleared, and not synced across a user's sessions. Suggest something like "the browser cookie set by the web UI's language switcher" so a future reader doesn't assume it is a persisted account attribute. Non-blocking.

example.env
```diff
@@ -425,6 +425,13 @@ XAGENT_WEB_SEARCH_PROVIDER="auto"
 # menus without sending the full accessibility tree to the model.
 # XAGENT_BROWSER_CUA_DRIVER_MAX_ELEMENTS="2000"
 
+# Fallback locale/timezone for browser_use (Playwright) sessions, used when a
+# task carries no account-derived locale (the app_locale cookie set by the web
+# UI's language switcher). Locale defaults to "en-US"; timezone defaults to
+# the host's own system timezone when unset.
+# XAGENT_BROWSER_TOOL_DEFAULT_LOCALE="en-US"
```
rogercloud, line 435: Worth one extra comment line here noting that an invalid value raises on first browser-tool use rather than being silently ignored/falling back. Given the validation lives well away from this file, an operator who typos a locale tag would otherwise have to trace it from a tool-call error back to this setting. Non-blocking.

rogercloud (approved):
## Re-review of `605ff964`

The latest commit is a focused pass over the latent-risk nits from the previous round: it adds a diagnostic log when a computer-tool session drops its resolved locale, documents the zh-TW/zh-HK folding behaviour and the `_get_is_admin_from_request` assertion risk, reworks the env-var getter docstrings, and removes the unused per-call `timezone_id` parameter outright (a net line reduction, which also lands the simplification suggestion from last round). No behavioural regressions in the diff — the changes are documentation, one debug log, and one deletion.

What did not change is the item flagged as the sole blocking ask in the previous round: test coverage for the new locale-resolution path. It's still unaddressed, but given the narrow blast radius (a UI-locale string, not auth/security data) and that the mechanism has already been exercised manually per the PR's test plan, I'm approving with this as a strong follow-up ask rather than a blocker.

### Already addressed since last review

- **M2** — `ComputerTool` now emits a debug log when a non-default environment factory causes the resolved locale to be dropped, so the silent hand-off gap is observable.
- **M3** — the primary-subtag folding of `zh-TW`/`zh-HK` into `zh-CN` is now documented as an accepted limitation, correctly scoped by the frontend's `"en" | "zh"` union.
- **M5** — the latent `AssertionError` in `_get_is_admin_from_request` (real `Request` without `AuthenticationMiddleware`) is now documented.
- **M6** + simplification — the unused `timezone_id` parameter is gone from `BrowserSession`, `BrowserSessionManager.get_or_create`, and `browser_navigate`. Confirmed removed, with a net reduction in lines.

### Suggested follow-up (non-blocking): test coverage for the new locale-resolution logic

The cookie-based locale resolution — the entirety of this PR's new logic — currently has zero test coverage. Grep across `tests/` at this head returns no hits for any of `get_browser_locale`, `_normalize_app_locale_cookie`, `app_locale`, or `browser_locale`. The only locale-adjacent tests are:

- `tests/core/test_config.py::TestBrowserToolDefaultLocaleConfig` — exercises only the deployment-level env-var default/override/blank-fallback in `src/xagent/config.py`. That is a different code path; it never touches cookie resolution.
- `tests/core/computer/test_browser_environment.py:159-161` — asserts a `FakeManager` call tuple in which `locale` is always `None`.

The one call site that actually resolves a request-scoped locale is `src/xagent/core/tools/adapters/vibe/browser_tools.py:53` (`locale=config.get_browser_locale()`), and it sits inside a broad `try/except Exception: return []`. So if `get_browser_locale()` or `_normalize_app_locale_cookie()` regressed — always returning `None`, mis-parsing `zh-CN`, or throwing — the suite would stay green and the tool would silently fall back to the deployment default. That's the exact failure mode this PR exists to fix, shipping without a guard against its own recurrence.

Two focused unit tests would close this: one over `_normalize_app_locale_cookie` (`"zh"` → `zh-CN`, `"en"` → `en-US`, absent/blank/non-string → `None`), and one over `get_browser_locale` with a stubbed request carrying an `app_locale` cookie, asserting the resolved tag reaches the browser tool. Both are cheap and need no Playwright. Would appreciate this landing as a fast follow-up.

### Minor suggestions (non-blocking, see inline comments)

- `src/xagent/core/tools/core/browser_use.py:293` — log line reports the raw (pre-fallback) `locale` argument rather than the resolved value.
- `src/xagent/core/tools/adapters/vibe/browser_use.py:1171` — `tool._locale` set from outside the class instead of via constructor, unlike `task_id`/`workspace` on the same classes.
- PR description is stale — still claims the three WebSocket call sites pass their live connection object, but `bec1ec61` reverted that.
- `src/xagent/web/tools/config.py:1788-1806` — docstring calls the cookie an "account setting"; there's no persisted `locale` on `User` or `TaskCreateRequest`.
- `example.env` — could note that an invalid `XAGENT_BROWSER_TOOL_DEFAULT_LOCALE`/`_TIMEZONE` raises at first browser-tool use rather than being silently ignored.

### Carried forward, not blocking

D1 (locale travelling through the process-global `AgentServiceManager.request` singleton rather than a task-scoped channel) stays as previously agreed: waived by the author with a documented rationale, tracked as #1282. Noting it only so it stays visible.

### Disposition

**Approving.** The hardcode-removal is the right fix and the mechanism is sound; the open test-coverage gap and the five minor items above are follow-ups, not blockers.

Issue #1132, opened 2026-08-04
fix(share): client-injectable identity/entity markers in share task-create agent_config


codeacme17:
Spun out of the review on #1124 (the share-path counterpart of finding F1).

## Problem

The widget F1 fix in #1124 hardened the **widget** task-create path so a guest cannot inject `widget_agent_id` / `widget_workforce_id` / `auth_mode` / `guest_id` into their own `agent_config` (those markers select the run-quota bucket at the `execute_task` chokepoint). The fix has three layers:
1. `sanitize_client_agent_config` (`src/xagent/web/services/task_runtime.py`) now strips the full public-channel marker set (`auth_mode`, `guest_id`, `widget_agent_id`, `widget_workforce_id`, `widget_client_ip`, `share_agent_id`, `share_workforce_id`, `share_token`) from every client body.
2. The widget-create path stamps both widget entity markers server-side (one as `None`).
3. The chokepoint coerces entity ids defensively (malformed → unkeyable → admit this task, no gate takedown).

Because layer 1 is in the shared sanitizer, the **share** path (`create_share_chat_task` → `create_public_chat_task` share branch in `src/xagent/web/api/public_chat_access.py`) is now also protected from injection of these keys. However, the share path has **not** received the equivalent layer-2 belt-and-suspenders (it stamps `auth_mode` / `share_agent_id` / `guest_id` after sanitize, but does not explicitly stamp `share_workforce_id`), and it has **no dedicated regression test** proving a forged `share_workforce_id` / `share_agent_id` cannot survive.

## Ask

- Add the share-path equivalent of the widget F1 regression test: post a share task-create with a forged `share_workforce_id` (and `share_agent_id`) in `agent_config` and assert the persisted `Task.agent_config` reflects only the server-resolved entity.
- Confirm the share workforce path (`_create_workforce_share_chat_task` / `create_workforce_run` via `_merge_agent_config`) is not injectable (it uses the snapshot-built config, not the client body — appears safe, but should be covered by a test).
- Consider stamping `share_workforce_id` explicitly on the share branch for parity with the widget fix.

Severity: the shared sanitizer already closes the practical injection vector; this is hardening + test coverage to prevent regression, not an open hole.

PR #1275, opened 2026-08-11
ref(share): make the ShareChatAccessContext exactly-one-entity invariant structural (#1225)

### LINKED ISSUE #1225: ref(share): make the ShareChatAccessContext exactly-one-entity invariant structural
Spun out of review feedback on #1221 (the fix for #1132).

## Problem

`ShareChatAccessContext` (`src/xagent/web/api/public_chat_access.py`) documents an invariant its type cannot express:

> exactly one of `agent` / `workforce` is set, depending on which kind of share token the guest presented

Nothing enforces it. Both fields default to `None`, and the dataclass accepts both-set and neither-set instances. The invariant holds today only because its single production factory, `get_share_chat_user`, happens to construct it correctly: the workforce branch (`:421`) sets only `workforce=`, the agent branch (`:437`) only `agent=`, with a `return` between them.

Downstream code then has to re-derive the invariant at each use, and the callers of `entity_rate_limit_key` inherit a contract they cannot check — that function's docstring says outright (`src/xagent/web/services/share_rate_limit.py:80`):

> Workforce wins when both ids are set (callers guarantee at most one is).

That guarantee is load-bearing for run-quota attribution: `_public_run_denial_channel` (`api/chat.py`) picks the billing bucket from these markers, and workforce wins on a tie. A future refactor that lets both fields be set would silently charge an agent-share run to a workforce bucket, with nothing failing loudly.

## Why now

#1221 stamps both entity markers on every public task-create path, writing the inapplicable one as `None`. On the agent-share branch the `None` is a literal, correct only because of the dispatch guard immediately above it. Review feedback proposed deriving it from the context instead; that would be actively wrong (it inverts the failure mode into the wrong-bucket charge described above), so the PR kept the literal and documented the guard.

The comment is the weak part. Prose is what keeps the literal honest, and prose does not fail a build.

## Ask

Make the invariant unrepresentable rather than narrated:

- Add `__post_init__` validation to the frozen `ShareChatAccessContext` rejecting both-set and neither-set, **or**
- Split it into a tagged union (an agent-share context and a workforce-share context) so the illegal state has no type.

Either way, the stamping sites can then cite one enforced invariant instead of re-deriving it, and the defensive comment in `create_share_chat_task` can shrink to a one-line reference.

## Caveat for whoever picks this up

Four tests construct the dataclass directly with `MagicMock` / `object()` stand-ins:

- `tests/web/api/test_share_rate_limit_endpoints.py:437` and `:490`
- `tests/web/test_upload_connection_boundary.py:707`
- `tests/web/test_public_runtime_extensions.py:47`

All four set exactly one field, so a `__post_init__` guard should pass as-is, but this has not been executed against the change — verify rather than assume. A tagged-union split would touch all four.

## Related

- #1132 / #1221 — the stamping fix this was spun out of
- #1134 — same "make it structural rather than call-site opt-in" theme, applied to `agent_config` sanitizing

codeacme17:
Closes #1225. Spun out of review feedback on #1221 (the fix for #1132).

## What

Make the `ShareChatAccessContext` exactly-one-entity invariant structural: `__post_init__` on the frozen dataclass now rejects both-set and neither-set `agent` / `workforce`, so the illegal states the docstring used to merely narrate are unconstructible.

## Why

The invariant is load-bearing for run-quota attribution: `entity_rate_limit_key` prefers workforce when both ids are set, so a future refactor that let both fields be populated would silently charge an agent-share run to a workforce bucket, with nothing failing loudly. Until now the only thing enforcing it was the shape of `get_share_chat_user`'s two branches, plus prose — and prose does not fail a build.

## Design decisions

- **Guard compares against `None`, not truthiness.** Tests construct the context with `MagicMock()` / `object()` / `SimpleNamespace` stand-ins; a stand-in whose `__bool__` is falsy is still a *set* entity. A dedicated test pins this.
- **A violation raises `ValueError` and is deliberately NOT mapped to a 401.** An invariant violation is a programming defect, never a bad credential. `get_share_chat_user` only projects `_PublicTokenRejected` / `HTTPException` into 401 and lets defects propagate loudly, matching the #1214 boundary policy (this PR is based on top of that refactor; the guard's comment documents the interaction).
- **`__post_init__` over the tagged-union split.** Roughly ten call sites dispatch on `workforce is not None` / `agent is None` fail-closed guards; splitting the type would touch all of them plus the FastAPI dependency typing for disproportionate gain. The guard already makes the illegal state fail at construction. The union remains open as a future deepening if the type grows more variant-specific behavior.

## Comment shrinkage (the issue's "Ask")

- The ~12-line defensive comment in `create_share_chat_task` justifying the literal `share_workforce_id = None` now cites the enforced invariant in one line instead of re-deriving it from the dispatch guard.
- The workforce-side stamping site (`_create_workforce_share_chat_task`) cites it too.
- The `entity_rate_limit_key` docstring notes the share path's "callers guarantee at most one is" is now structural. (The widget path's identical guarantee on `PublicChatAccessContext` is still caller-maintained — see follow-up note below.)

## Tests

New `tests/web/api/test_share_access_context_invariant.py`:

- agent-only and workforce-only construction passes;
- both-set and neither-set raise `ValueError`;
- a falsy-but-set entity object still counts as set.

The issue's caveat about the four existing direct constructions (`test_share_rate_limit_endpoints.py:437`/`:490`, `test_upload_connection_boundary.py:707`, `test_public_runtime_extensions.py:47` — all under `tests/web/api/`, the issue's paths were missing the `api/` segment) is verified by execution, not assumption: each sets exactly one field and all pass unchanged.

Full `tests/web` suite run: green except three pre-existing SVG preview rasterization failures (`test_svg_preview_cache.py`, `test_file_upload.py::TestSvgPreviewSecurity`) that reproduce identically on a clean `main` baseline worktree — environmental, unrelated to this change. `ruff format --check`, `ruff check`, and `mypy` are clean on the touched files.

## Follow-up

`PublicChatAccessContext` (the widget twin directly above) documents the same exactly-one shape for `widget_agent_id` / `widget_workforce_id` and remains narration-only. Out of scope here per the issue; worth a sibling issue if this approach is accepted.

### DIFF
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -88,6 +88,8 @@ class PublicChatAccessContext:
     widget_agent_id: int | None = None
     # Set instead of ``widget_agent_id`` when the widget key exposes a
     # workforce. Exactly one of the two is populated for a widget guest.
+    # Structural enforcement is tracked in #1288 (the share twin is already
+    # enforced below, #1225).
     widget_workforce_id: int | None = None
 
 
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -119,9 +119,11 @@ class ShareChatAccessContext:
         # workforce bucket; a neither-set one has no billing bucket at all.
         # Compared against None, not truthiness — tests construct the context
         # with arbitrary stand-in objects. A violation is a programming
-        # defect, never a bad credential, so ValueError deliberately does NOT
-        # map to a 401: get_share_chat_user only projects _PublicTokenRejected
-        # / HTTPException to 401 and lets defects propagate loudly (#1214).
+        # defect, never a bad credential: get_share_chat_user only rewrites
+        # _PublicTokenRejected into a 401 and passes HTTPException through
+        # unchanged (e.g. the 403s from ensure_share_*_available), so this
+        # ValueError propagates loudly (#1214). Both production construction
+        # sites set exactly one entity; the guard defends future call sites.
         if (self.agent is None) == (self.workforce is None):
             raise ValueError(
                 "ShareChatAccessContext requires exactly one of 'agent' or 'workforce'"
--- a/src/xagent/web/api/public_chat_access.py
+++ b/src/xagent/web/api/public_chat_access.py
@@ -95,7 +95,8 @@ class PublicChatAccessContext:
 class ShareChatAccessContext:
     """Validated share-guest identity: exactly one of ``agent`` /
     ``workforce`` is set, depending on which kind of share token the guest
-    presented. ``user`` is always the owner the shared entity runs as.
+    presented — enforced at construction by ``__post_init__`` (#1225).
+    ``user`` is always the owner the shared entity runs as.
 
     ``guest_id`` is the per-guest isolation credential (#973): a share link is
     public, so many anonymous visitors share the same owner + entity id, and
@@ -110,6 +111,21 @@ class ShareChatAccessContext:
     guest_id: str
     agent: Agent | None = None
     workforce: Workforce | None = None
+
+    def __post_init__(self) -> None:
+        # Exactly-one guard (#1225): the run quota keys off these markers and
+        # entity_rate_limit_key prefers workforce when both ids are set, so a
+        # both-set context would silently charge an agent-share run to a
+        # workforce bucket; a neither-set one has no billing bucket at all.
+        # Compared against None, not truthiness — tests construct the context
+        # with arbitrary stand-in objects. A violation is a programming
+        # defect, never a bad credential, so ValueError deliberately does NOT
+        # map to a 401: get_share_chat_user only projects _PublicTokenRejected
+        # / HTTPException to 401 and lets defects propagate loudly (#1214).
+        if (self.agent is None) == (self.workforce is None):
+            raise ValueError(
+                "ShareChatAccessContext requires exactly one of 'agent' or 'workforce'"
+            )
 
 
 def mint_share_guest_id() -> str:
@@ -1154,11 +1170,13 @@ async def _create_workforce_share_chat_task(
             # Null the agent marker for symmetry with the agent path (#1132),
             # mirroring the widget workforce branch: entity_rate_limit_key
             # prefers workforce, so a stray agent id would be ignored anyway,
-            # but stamping both keeps the run-quota key unambiguous. Safe
-            # because the snapshot-built config never sets either marker --
-            # not because merge order is irrelevant: _merge_agent_config
-            # returns {**extra_agent_config, **task_config}, so the built
-            # config wins every collision and this dict loses.
+            # but stamping both keeps the run-quota key unambiguous. The None
+            # is a literal by the context's exactly-one invariant, enforced in
+            # ShareChatAccessContext.__post_init__ (#1225). Safe also because
+            # the snapshot-built config never sets either marker -- not
+            # because merge order is irrelevant: _merge_agent_config returns
+            # {**extra_agent_config, **task_config}, so the built config wins
+            # every collision and this dict loses.
             "share_agent_id": None,
             # Per-guest isolation (#973). extra_agent_config is overlaid under
             # the snapshot-built config, which never sets guest_id, so this is
@@ -1222,18 +1240,9 @@ async def create_share_chat_task(
     # workforce, so a client-injected share_workforce_id must never survive to
     # redirect an agent-share task's bucket onto a workforce. The sanitizer
     # already strips both, but stamping explicitly keeps this correct
-    # independent of the sanitizer's reserved-key set.
-    #
-    # The workforce marker is a literal None because the dispatch above forces
-    # it: this branch runs only after the early return into
-    # _create_workforce_share_chat_task, so access_context.workforce is None
-    # here and a context read would yield the same value today. Preserving that
-    # guard is what keeps the literal honest. If it were ever removed, a
-    # context read would not be the safe fallback it looks like: this branch
-    # builds an *agent* task (agent_id=share_agent_id below), and
-    # entity_rate_limit_key prefers workforce whenever both markers are set, so
-    # a non-None workforce id here would charge an agent-share run to a
-    # workforce bucket — the misattribution this stamping exists to prevent.
+    # independent of the sanitizer's reserved-key set. The workforce marker is
+    # a literal None by the context's exactly-one invariant, enforced in
+    # ShareChatAccessContext.__post_init__ (#1225).
     agent_config["share_agent_id"] = share_agent_id
     agent_config["share_workforce_id"] = None
 
--- a/tests/web/api/test_share_access_context_invariant.py
+++ b/tests/web/api/test_share_access_context_invariant.py
@@ -11,13 +11,14 @@ unconstructible instead of merely narrated in docstrings.
 from __future__ import annotations
 
 from types import SimpleNamespace
+from typing import Any
 
 import pytest
 
 from xagent.web.api.public_chat_access import ShareChatAccessContext
 
 
-def _context_kwargs(**entity: object) -> dict[str, object]:
+def _context_kwargs(**entity: Any) -> dict[str, Any]:
     return {
         "user": SimpleNamespace(id=1),
         "share_token": "tok",
@@ -27,28 +28,24 @@ def _context_kwargs(**entity: object) -> dict[str, object]:
 
 
 def test_agent_only_context_constructs() -> None:
-    context = ShareChatAccessContext(
-        **_context_kwargs(agent=SimpleNamespace(id=7))  # type: ignore[arg-type]
-    )
+    context = ShareChatAccessContext(**_context_kwargs(agent=SimpleNamespace(id=7)))
     assert context.workforce is None
 
 
 def test_workforce_only_context_constructs() -> None:
-    context = ShareChatAccessContext(
-        **_context_kwargs(workforce=SimpleNamespace(id=7))  # type: ignore[arg-type]
-    )
+    context = ShareChatAccessContext(**_context_kwargs(workforce=SimpleNamespace(id=7)))
     assert context.agent is None
 
 
 def test_neither_entity_set_is_rejected() -> None:
     with pytest.raises(ValueError, match="exactly one"):
-        ShareChatAccessContext(**_context_kwargs())  # type: ignore[arg-type]
+        ShareChatAccessContext(**_context_kwargs())
 
 
 def test_both_entities_set_is_rejected() -> None:
     with pytest.raises(ValueError, match="exactly one"):
         ShareChatAccessContext(
-            **_context_kwargs(  # type: ignore[arg-type]
+            **_context_kwargs(
                 agent=SimpleNamespace(id=7),
                 workforce=SimpleNamespace(id=9),
             )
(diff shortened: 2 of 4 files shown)


codeacme17:
Thanks for the review. The type-annotation suggestion is addressed in f9ff2b5 — accepted in substance and extended slightly: the helper now returns `dict[str, Any]` (with `**entity: Any`), and the five `# type: ignore[arg-type]` suppressions at the construction call sites are deleted in the same commit, since this repo enables `warn_unused_ignores` and they would otherwise become new mypy errors for anyone type-checking the tests. Details in the inline reply.

tests/web/api/test_share_access_context_invariant.py
```diff
@@ -0,0 +1,81 @@
...
+context would silently charge an agent-share run to a workforce bucket.
+These tests pin the ``__post_init__`` guard that makes the illegal states
+unconstructible instead of merely narrated in docstrings.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
```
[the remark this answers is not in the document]
codeacme17: Done in f9ff2b5, with two adjustments beyond the suggested diff:

1. The suggestion block only adds the `Any` import; applied literally it would fail ruff (F401 unused import) since nothing else in the diff uses it. The actual change is the helper signature: `_context_kwargs(**entity: Any) -> dict[str, Any]`.
2. With `dict[str, Any]`, mypy accepts the `**`-unpacked construction sites, so the five `# type: ignore[arg-type]` comments became unnecessary and were deleted in the same commit — this repo sets `warn_unused_ignores = true`, so leaving them in place would turn them into `unused-ignore` errors for anyone running mypy over the tests.

Verified: ruff format/check clean, `mypy tests/web/api/test_share_access_context_invariant.py` reports zero errors and zero unused-ignore warnings, all 5 tests pass. The nit about the rule's phrasing: `dict[str, Any]` is the looser type here, not the tighter one — but the practical effect (mypy-clean unpacking without suppressions) is what this change is for. It also matches the `MagicMock`-as-`Any` stand-in idiom the sibling share tests already use.

src/xagent/web/api/public_chat_access.py
```diff
@@ -111,6 +112,21 @@ class ShareChatAccessContext:
...
+    def __post_init__(self) -> None:
+        # Exactly-one guard (#1225): the run quota keys off these markers and
+        # entity_rate_limit_key prefers workforce when both ids are set, so a
+        # both-set context would silently charge an agent-share run to a
+        # workforce bucket; a neither-set one has no billing bucket at all.
+        # Compared against None, not truthiness — tests construct the context
+        # with arbitrary stand-in objects. A violation is a programming
+        # defect, never a bad credential, so ValueError deliberately does NOT
+        # map to a 401: get_share_chat_user only projects _PublicTokenRejected
+        # / HTTPException to 401 and lets defects propagate loudly (#1214).
```
rogercloud, line 124: **minor (comment accuracy):** `get_share_chat_user` does not project `HTTPException` to 401. `_project_public_token_failure` (`:186-195`) returns `HTTPException` **unchanged**; only `_PublicTokenRejected` is rewritten into a 401. The `HTTPException`s that actually flow through here are the **403**s from `ensure_share_agent_available:293` / `ensure_share_workforce_available:325` ("Share link is unavailable"), so a reader trusting this line would believe a revoked share link surfaces as 401.

Suggested rewording:

> `get_share_chat_user` only rewrites `_PublicTokenRejected` into a 401 and passes `HTTPException` through unchanged (e.g. the 403s from `ensure_share_*_available`); anything else propagates.

**minor (untested claim), same sentence:** "deliberately does NOT map to a 401" is the comment's most consequential assertion and nothing pins it. Tracing `get_share_chat_user`, the `_is_strict_int` dispatch at `:509-516` plus both `ensure_share_*_available` helpers (non-optional return type, raise rather than return `None`) make this `ValueError` unreachable from any request path — which is the stronger argument anyway. Either state the unreachability instead of the 401 behaviour, or add one test asserting a raw `ValueError` survives `_project_public_token_failure` unprojected; the existing `tests/web/api/test_public_chat_websocket_db_boundary.py:520` only covers the `HTTPException`-passthrough branch.

tests/web/api/test_share_access_context_invariant.py
```diff
@@ -0,0 +1,76 @@
...
+        "user": SimpleNamespace(id=1),
+        "share_token": "tok",
+        "guest_id": "guest",
+        **entity,
+    }
+
+
+def test_agent_only_context_constructs() -> None:
+    context = ShareChatAccessContext(**_context_kwargs(agent=SimpleNamespace(id=7)))
+    assert context.workforce is None
```
rogercloud, line 32: **minor (test quality):** tautological assertion. `workforce` is never passed, so `context.workforce is None` holds by dataclass default regardless of whether the guard is correct — all the signal comes from the construction not raising. Same at `:37`.

Binding the entity to a local and asserting on the field that was actually set carries real signal (it also pins that `__post_init__` did not mutate anything under `frozen=True`):

```python
def test_agent_only_context_constructs() -> None:
    agent = SimpleNamespace(id=7)
    context = ShareChatAccessContext(**_context_kwargs(agent=agent))
    assert context.agent is agent
    assert context.workforce is None
```

tests/web/api/test_share_access_context_invariant.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __bool__(self) -> bool:
+        return False
+
+
+def test_falsy_entity_object_counts_as_set() -> None:
+    context = ShareChatAccessContext(**_context_kwargs(agent=_FalsyEntity()))
+    assert context.agent is not None
+
+    with pytest.raises(ValueError, match="exactly one"):
```
rogercloud, line 70: **minor (test quality):** this block cannot discriminate the bug `_FalsyEntity` exists to catch. Under a hypothetical truthiness guard `(not self.agent) == (not self.workforce)`, both-falsy is `True == True` → still raises, so the block passes under both the correct `is None` guard and the buggy one. It also duplicates `test_both_entities_set_is_rejected` (`:45-52`), just with falsy stand-ins.

Line 67 — a single falsy `agent` with `workforce` unset — is the case that actually distinguishes the two: under the truthiness bug it would incorrectly raise. That one earns its place; suggest dropping `:70-76`.

rogercloud (approved):
## Summary

This PR makes the `ShareChatAccessContext` "exactly one of `agent` / `workforce`" invariant structural: a `__post_init__` guard on the frozen dataclass rejects both-set and neither-set construction, so the illegal states that previously held only because `get_share_chat_user`'s two branches happened to construct correctly are now unconstructible. Two long defensive comments at the `share_agent_id` / `share_workforce_id` stamping sites shrink to one-line citations of the enforced invariant, the `entity_rate_limit_key` docstring notes the share path's guarantee is now structural, and a new test module pins the guard (including that it compares against `None`, not truthiness).

## Approach: sound

`__post_init__` is the right call over the tagged-union alternative. A union would force changes at every downstream site dispatching on `access_context.workforce is not None` (`create_share_chat_task`, `_create_workforce_share_chat_task`, `get_task_for_share_context`, `_authorize_share_chat_websocket`) plus the `_ChatAccessContext` Protocol / `_ChatAccessContextT` TypeVar machinery that is deliberately structural across both context types. The union buys static enforcement; the guard buys the same runtime guarantee in four lines with zero call-site churn — the right trade for an auth-boundary refactor.

Two design notes, neither blocking:

- The guard closes the **construction** half only. The two stamping sites still write a literal `None` rather than reading the context, so the invariant does not mechanically protect those lines — it makes the literal *provably* correct. Verified there is no other writer of these markers (`src/xagent/web/api/share.py:86` / `:129` at token mint, plus the two `public_chat_access.py` sites), and `sanitize_client_agent_config` strips both from client input, so both-set `agent_config` markers are unreachable on the share path. The hole is genuinely closed, but by guard + sanitizer + literal stamping together rather than by the guard alone.
- Frozen dataclass + `__post_init__` is safe here: the hook only raises, never assigns. No `dataclasses.replace` / `asdict` / `deepcopy` on either context anywhere in the repo.

All four pre-existing direct constructions (`tests/web/api/test_share_rate_limit_endpoints.py:437` / `:490`, `tests/web/api/test_upload_connection_boundary.py:707`, `tests/web/api/test_public_runtime_extensions.py:47`) set exactly one field and still pass the guard; a repo-wide grep found no other construction site and no `**kwargs`-splat construction outside the new test.

I also checked that nothing load-bearing left with the ten deleted comment lines in `create_share_chat_task`: the core warning ("a client-injected `share_workforce_id` must never redirect an agent-share task's bucket onto a workforce") survives verbatim in the retained lines above, and the removed dispatch-based proof is superseded by the stronger invariant-based one.

## Findings

All minor — none blocks merge. Three are inline; one more here because it lands outside the diff:

**minor — `src/xagent/web/api/public_chat_access.py:89-91`, the widget twin is now the weaker of two identical claims.** `PublicChatAccessContext` carries the same unenforced "exactly one of the two is populated" invariant five lines above, and `widget_entity_key` (`:1288-1297`) feeds the *same* `entity_rate_limit_key` with the same workforce-wins preference, consumed by the same run-quota gate (`src/xagent/web/api/chat.py:1095-1103`) — the misattribution argument is verbatim identical. I verified all five widget-context test constructions (`test_share_rate_limit_endpoints.py:576`/`:632`, `test_upload_connection_boundary.py:668`, `test_share_guest_isolation.py:689`/`:697`, `test_public_runtime_extensions.py:21`) set exactly one field, so a symmetric guard would land without churn. The PR description flags this as out of scope, which is fair — but please file the sibling issue so the asymmetry does not read as an oversight, or add a one-line note at `:89-91` saying it is deliberate.

The prior review thread (the `_context_kwargs` type annotation) is addressed in `f9ff2b5`: the helper now returns `dict[str, Any]` and the `type: ignore[arg-type]` suppressions are gone, which is right given `warn_unused_ignores`.

**Simplification:** lean already — nothing to strip.

Approving; the inline items are polish and can land in a follow-up commit or be waived.

codeacme17:
Thanks for the thorough review — all four points were verified and are addressed across d72a003 and ffa7d39:

- **Comment accuracy + untested claim** (inline): comment reworded to the accurate projection behavior, plus a direct test pinning that `_project_public_token_failure(ValueError(...))` returns `None` — details in the inline reply.
- **Tautological assertions** (inline): both happy-path tests now bind the entity and assert identity on the field actually set.
- **Non-discriminating both-falsy block** (inline): dropped; the single-falsy case stays with an identity assertion.
- **Widget twin asymmetry**: filed #1288 as the sibling issue (mirrors #1225, records your landability verification of the six widget-context test constructions, and flags the `auth_mode` nuance you'd want decided deliberately). Also added the one-line cross-reference at the `PublicChatAccessContext` comment (ffa7d39) so the asymmetry reads as tracked work.

src/xagent/web/api/public_chat_access.py
codeacme17: Both sub-points were right — fixed in d72a003.

**Accuracy:** the comment now states that `get_share_chat_user` only rewrites `_PublicTokenRejected` into a 401 and passes `HTTPException` through unchanged (e.g. the 403s from `ensure_share_*_available`), essentially your suggested wording. It also now notes that both production construction sites set exactly one entity, so the guard defends future call sites — the unreachability argument you identified as the stronger one.

**Untested claim:** went with the direct unit test rather than forcing the (unreachable) `ValueError` end-to-end through `get_share_chat_user`: `test_public_token_failure_projection_leaves_defects_unprojected` sits next to the existing `HTTPException`-identity test in `test_public_chat_websocket_db_boundary.py` and asserts `_project_public_token_failure(ValueError(\"boom\"), ...)` is `None` — covering the projection's previously untested fall-through branch that the comment's claim rests on.

tests/web/api/test_share_access_context_invariant.py
codeacme17: Implemented as suggested in d72a003, for both tests: the entity is bound to a local and the test asserts `context.agent is agent` (resp. `context.workforce is workforce`) plus `None` on the other field — so the assertions now carry signal beyond "construction did not raise", and pin that `__post_init__` mutated nothing under `frozen=True`.

tests/web/api/test_share_access_context_invariant.py
codeacme17: Verified your truth table independently and you're right: under the hypothetical truthiness guard, both-falsy is `True == True` → raises, same observable outcome as the correct `is None` guard, so the block passed under both and duplicated `test_both_entities_set_is_rejected` with falsy stand-ins. Dropped in d72a003. The single-falsy construction — the one case where the two guards diverge (truthiness bug raises, correct guard constructs) — stays, upgraded with an identity assertion (`context.agent is agent`) in line with the other thread.

PR #1277, opened 2026-08-11
feat(workforce): support permanent delete and unarchive

yiboyasss:
## Summary
- Archived workforces currently have no way to be fully removed — they just sit in the list forever. Added `DELETE /api/workforces/{id}?permanent=true`: hard-deletes a workforce in any status (draft/active/archived), cascading workers, run history, and builder messages/triggers, and explicitly cleaning up API keys and the deployment row, which aren't covered by ORM cascades (the `Deployment` table has no DB foreign key — it's keyed by `owner_type`/`owner_id`). Any in-flight runs are cancelled before delete, with PAUSE dispatched after commit, mirroring the existing archive flow.
- If the workforce's manager agent was auto-generated exclusively for it, it's deleted alongside; if that manager is shared with another workforce or owned by someone else, it's left alone and only this workforce is removed. This is intentionally different from discarding a draft: discard is an "undo create" and blocks entirely when the manager is shared, but permanent delete's goal is just "remove this one" and shouldn't be blocked by an unrelated reference elsewhere.
- Added `POST /api/workforces/{id}/unarchive`, restoring an archived workforce to `draft` (not directly back to `active`) — since archiving already cancelled its runs and the config may be stale, re-activation still goes through an explicit publish step that re-validates.
- Triggers cascade-deleted this way that hold a provider-side binding (e.g. a Gmail watch) get that binding unregistered after commit via `asyncio.to_thread`, so the async route never blocks the event loop on provider network I/O.
- Frontend: workforce cards get a three-dot menu (publish/unpublish, archive/unarchive, delete with a confirm dialog) matching the Agent card pattern; the builder detail page drops its standalone archive button now that the action lives in the list menu. Deleting the last card on a page steps pagination back a page instead of landing on an out-of-range empty page.

## Test plan
- [x] Added 11 backend tests covering: unarchive status/permission checks, permanent delete succeeding from any status and with run history present, manager cleanup safety/shared-manager scenarios, cascade-deleted trigger unregister dispatch, permission rejection, and rollback on failure
- [x] Ran the full existing workforce/trigger test suites (`pytest tests/web/api/test_workforces_api.py tests/web/api/test_workforce_deployment_foundation.py ...`)
- [x] Frontend `workforces-routes.test.tsx`, `workforces-api.test.ts`, and i18n translation tests all pass
- [x] ruff / mypy / eslint / tsc all clean


yiboyasss:
- **The suggested reordering** (run `unregister_deleted_trigger_bindings` *before* `delete_workforce_permanently` rather than after): not adopted. The current commit-then-best-effort-cleanup ordering matches the existing pattern this file already uses for archive (`cancel_active_workforce_runs` commits, then `pause_workforce_tasks_after_archive` dispatches after) — the DB change commits atomically and external side effects clean up best-effort afterward, logged-not-raised on failure. Reordering would block the delete transaction on provider network I/O, and a provider-unregister-succeeds-but-DB-delete-fails case would leave the opposite inconsistency (external resource released, DB row still exists) with no simple recovery path. The `expunge()` fix addresses the concrete defect the comments raised (attribute access after expiry) without changing that ordering.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +211,110 @@ def discard_draft_workforce(
...
+        # table) cannot be relied on here.
+        db.query(AgentApiKey).filter(AgentApiKey.workforce_id == workforce_id).delete(
+            synchronize_session=False
+        )
+        db.query(Deployment).filter(
+            Deployment.owner_type == DeploymentOwnerType.WORKFORCE.value,
+            Deployment.owner_id == workforce_id,
+        ).delete(synchronize_session=False)
+
+        db.delete(workforce)
```
rogercloud, line 352: The confirm dialog promises `"{name}" and its run history will be permanently deleted`, but this never touches `Task`. `WorkforceRun.task_id` is `ondelete="SET NULL"` with a non-cascading `relationship("Task")`, and workforce-run tasks default to `is_visible=True` — so the conversations, traces and workspace-file records stay fully visible after a "permanent delete". Please delete (or at least set `is_visible=False` on) the tasks referenced by the cascaded runs inside this transaction, or soften the dialog text to admit conversations are retained.

tests/web/api/test_workforces_api.py
```diff
@@ -933,6 +936,395 @@ def test_archived_workforce_rejects_all_edit_boundaries() -> None:
...
+    try:
+        workforce_row = db.get(Workforce, workforce_id)
+        assert workforce_row is not None
+        assert workforce_row.status == "archived"
+    finally:
+        db.close()
+
+
+@pytest.mark.parametrize("status", ["draft", "active", "archived"])
+def test_permanent_delete_removes_workforce_regardless_of_status(
```
rogercloud, line 1011: None of the seven permanent-delete tests create an active-status run — the only run is `status="completed"`, which is excluded from `ACTIVE_WORKFORCE_RUN_STATUSES`. So `cancel_active_workforce_runs` returns `[]` and `pause_workforce_tasks_after_archive` short-circuits in every test, leaving the in-flight-run cancellation and PAUSE dispatch with zero coverage. `tests/web/api/test_workforce_deployment_foundation.py:600-641` already does this for archive (asserting the durable pause command dispatch) — please mirror it here.

frontend/src/app/workforces/workforces-routes.test.tsx
```diff
@@ -12,8 +12,10 @@ const listWorkforceRunsMock = vi.hoisted(() => vi.fn())
 const runWorkforceMock = vi.hoisted(() => vi.fn())
 const addWorkforceAgentMock = vi.hoisted(() => vi.fn())
 const archiveWorkforceMock = vi.hoisted(() => vi.fn())
+const deleteWorkforcePermanentlyMock = vi.hoisted(() => vi.fn())
 const publishWorkforceMock = vi.hoisted(() => vi.fn())
```
rogercloud, line 16: `deleteWorkforcePermanentlyMock`, `unarchiveWorkforceMock`, `archiveWorkforceMock`, `publishWorkforceMock` and `unpublishWorkforceMock` are all declared, wired and reset, but no test ever opens the three-dot menu, clicks an action, or asserts any of them was called — they read as coverage while being dead scaffolding. This file already has the render + `fireEvent`/`waitFor`/toast + dialog patterns needed, so please add tests clicking Unarchive and Delete (through the confirm dialog). Also cover the pagination step-back logic at `page.tsx:143-152`, which is novel and currently unexercised.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +211,110 @@ def discard_draft_workforce(
...
+        # ordering as _delete_trigger.
+        trigger_teardowns: list[tuple[AgentTrigger, str, dict[str, Any]]] = []
+        for trigger in (
+            db.query(AgentTrigger)
+            .filter(AgentTrigger.workforce_id == workforce_id)
+            .all()
+        ):
+            trigger_type = str(trigger.type)
+            config = dict(trigger.config or {})
+            # Expunge now, before this session's cascade-delete and commit
```
rogercloud: The justification for `db.expunge(trigger)` isn't accurate on the pinned SQLAlchemy 2.0.51: a deleted-and-committed instance becomes `detached` while retaining its loaded values, so reading an attribute does not raise `DetachedInstanceError`/`ObjectDeletedError`. The `expunge` itself is harmless defensive hygiene and can stay, but please fix or drop the comment so it doesn't teach the wrong semantics. The neighbouring "`Workforce.triggers` hasn't been loaded yet" claim is true today but depends on `_load_workforce`'s eager-load set — worth a one-line note tying them together.

src/xagent/web/api/workforces.py
```diff
@@ -802,11 +805,41 @@ async def update_workforce(
...
+            logger.exception("Failed to delete workforce %s", workforce_id)
+            db.rollback()
+            raise HTTPException(
+                status_code=500,
+                detail={
+                    "code": "workforce_delete_failed",
+                    "message": "Failed to delete workforce",
+                },
+            ) from None
+        await pause_workforce_tasks_after_archive(pause_targets)
```
rogercloud: `pause_workforce_tasks_after_archive(pause_targets)` is missing `reason=` here, so pause failures during a permanent delete are logged as archive failures. The parameter exists for exactly this and is used correctly elsewhere (e.g. `src/xagent/web/app.py:337-339` passes `reason="preview-reap"`). Please pass `reason="delete"`.

src/xagent/web/api/workforces.py
```diff
@@ -802,11 +805,41 @@ async def update_workforce(
...
 
 @router.delete("/{workforce_id}")
-async def archive_workforce(
+async def archive_or_delete_workforce(
     workforce_id: int,
+    permanent: bool = False,
     db: Session = Depends(get_db),
     user: User = Depends(get_current_user),
 ) -> dict[str, Any]:
+    if permanent:
```
rogercloud, line 822: Committing before the best-effort PAUSE means an in-flight turn can briefly run against deleted rows. That's symmetric with archive's intentional pattern and the failure maps to a clear `workforce_run_not_found` message, so no ordering change needed — but there's no test for it. Please add an integration test that deletes mid-turn and asserts the turn is rejected with that specific error rather than a 500.

src/xagent/web/api/workforces.py
```diff
@@ -802,11 +805,41 @@ async def update_workforce(
...
+            db.rollback()
+            raise HTTPException(
+                status_code=500,
+                detail={
+                    "code": "workforce_delete_failed",
+                    "message": "Failed to delete workforce",
+                },
+            ) from None
+        await pause_workforce_tasks_after_archive(pause_targets)
+        if trigger_teardowns:
```
rogercloud, line 856: These provider teardowns are awaited sequentially on the request's critical path *after* the rows are already committed as deleted — for N Gmail triggers that's N network round-trips (Gmail API + Pub/Sub, each with its own commit) delaying the response. With no client-side timeout in the fetch wrapper the usual outcome is a slow 200, but an upstream gateway timeout would surface "Failed to delete workforce" (`frontend/src/app/workforces/page.tsx:153-156`) for an operation that fully succeeded. Consider `asyncio.gather`-ing them, or moving teardown off the request path entirely so the response returns at commit.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +211,110 @@ def discard_draft_workforce(
...
+        db.query(AgentApiKey).filter(AgentApiKey.workforce_id == workforce_id).delete(
+            synchronize_session=False
+        )
+        db.query(Deployment).filter(
+            Deployment.owner_type == DeploymentOwnerType.WORKFORCE.value,
+            Deployment.owner_id == workforce_id,
+        ).delete(synchronize_session=False)
+
+        db.delete(workforce)
+        db.flush()
```
rogercloud, line 353: This cascades `workers`/`runs`/`builder_messages`/`triggers` (and transitively `TriggerRun`/`TriggerAudit`) object-by-object with no `passive_deletes=True`, all while holding the lifecycle fence — which per its own docstring is a whole-DB writer lock on SQLite. Archive never deleted anything, so this is a new cost profile: a long-lived workforce with a busy webhook trigger accumulates audit rows per callback and could turn one delete into a multi-second global write stall. Consider `passive_deletes=True` plus matching DB-level `ondelete` on the high-volume child relationships.

tests/web/api/test_workforces_api.py
```diff
@@ -933,6 +936,395 @@ def test_archived_workforce_rejects_all_edit_boundaries() -> None:
...
+        )
+        db.add(trigger)
+        db.commit()
+        trigger_id = int(trigger.id)
+    finally:
+        db.close()
+
+    captured: list[Any] = []
+    monkeypatch.setattr(
+        "xagent.web.api.workforces.unregister_deleted_trigger_bindings",
```
rogercloud, line 1196: Monkeypatching `unregister_deleted_trigger_bindings` wholesale means the body added in `5c412ed5` — `get_session_local()`, the per-trigger `with SessionLocal() as db:` isolation, the delegation to `_unregister_trigger_binding`, and the per-item `except Exception` swallow — is never executed by any test. Please add a variant that keeps the real function and stubs only the provider's `unregister`, asserting it's called once per trigger with the right `config` and that one provider raising doesn't block the remaining teardowns.

tests/web/api/test_workforces_api.py
```diff
@@ -933,6 +936,395 @@ def test_archived_workforce_rejects_all_edit_boundaries() -> None:
...
+    workforce = _create_workforce(headers, name="Delete Reusable Manager Workforce")
+    workforce_id = int(workforce["id"])
+    manager_id = int(workforce["manager"]["id"])
+
+    response = client.delete(
+        f"/api/workforces/{workforce_id}?permanent=true", headers=headers
+    )
+
+    assert response.status_code == 200, response.text
+    db = _direct_db_session()
```
rogercloud, line 1107: This doesn't test what the name implies: the manager here is `USER`-origin (default `_create_workforce` helper), so it hits `_lock_generated_manager`'s early short-circuit and never reaches the exclusive-ownership/shared-manager logic. Those branches are covered by other tests in this file, so it's not a coverage hole — but please rename (e.g. `test_permanent_delete_skips_user_owned_manager`) or extend it to build an actually-shared generated manager.

tests/web/api/test_workforces_api.py
rogercloud, line 1011: A few smaller gaps worth closing while you're here: the `used_as_worker` branch of manager-safety checking is untested on both discard and permanent-delete (only `shared_reference`/`owner_mismatch` are covered, and only on discard); there's no 404 test for permanent-delete of an unknown id, nor for unarchive of an unknown id (`src/xagent/web/api/workforces.py:872-873`); and nothing asserts on `TriggerRun`/`TriggerAudit` rows after a cascade delete, so the audit-preserving nullify at `src/xagent/web/models/trigger.py:140-143` is unverified — which matters because SQLite in tests runs without FK enforcement, making the ORM cascade the only cleanup.

frontend/src/app/workforces/page.tsx
```diff
@@ -165,28 +251,98 @@ export default function WorkforcesPage() {
...
                             </div>
-                            <div className="text-xs text-muted-foreground truncate mt-0.5">
+                            <div className="text-xs text-muted-foreground truncate mt-1">
                               {t("workforces.list.manager", { name: item.manager?.name })}
                             </div>
                           </div>
                         </div>
 
+                        <div className="absolute right-3 top-3">
+                          <Popover>
```
rogercloud: This `Popover` is uncontrolled with no `PopoverClose` or close-on-click, so it stays open after an action — and since the Publish/Unpublish block is conditionally rendered on `item.status`, the child list changes length and reflows once `load()` resolves. A second click at the same position after Archiving lands on **Delete**, not Unarchive. Delete only opens a confirm dialog so it's not destructive, but please close the popover on any action or hold a controlled `open` reset in each handler's `finally`.

src/xagent/web/api/workforces.py
```diff
@@ -802,11 +805,41 @@ async def update_workforce(
 
 
 @router.delete("/{workforce_id}")
-async def archive_workforce(
+async def archive_or_delete_workforce(
```
rogercloud, line 811: Renaming `archive_workforce` → `archive_or_delete_workforce` changes the generated OpenAPI `operationId` for `DELETE /api/workforces/{workforce_id}`, since there's no explicit `operation_id=` override. Nothing in-repo consumes the old name, so impact is limited to external OpenAPI consumers — but if any exist this is a breaking rename. Either pin `operation_id="archive_workforce"` on the decorator or call it out in the PR/changelog notes.

frontend/src/components/workforce/workforce-builder.tsx
```diff
@@ -771,11 +753,6 @@ export function WorkforceBuilder({ workforceId }: WorkforceBuilderProps) {
                                     {t("workforces.actions.triggers")}
                                 </Button>
                             )}
```
rogercloud, line 755: The button removed here was gated on `!isArchived`, so the "can't unarchive from the builder" gap is pre-existing rather than something this PR caused. What does regress mildly is that a non-archived workforce could previously be archived inline from its own detail page and now requires navigating back to the list. Minor UX nit — consider keeping an archive entry in the builder header, or ignore if the list page is meant to be the single home for lifecycle actions.

rogercloud (changes requested):
# Review: `feat(workforce): support permanent delete and unarchive`

## Summary

Archived workforces are currently terminal dead ends — they can neither be removed nor brought back. This PR adds two escapes: `DELETE /api/workforces/{id}?permanent=true`, which hard-deletes a workforce in any status (cascading workers, runs, builder messages and triggers, plus manual cleanup of `agent_api_keys` and the FK-less `deployments` row, and the auto-generated manager agent when it is exclusively owned by this workforce), and `POST /api/workforces/{id}/unarchive`, which restores an archived workforce to `draft` so re-activation still has to go through an explicit re-validating publish. In-flight runs are cancelled inside the delete transaction with PAUSE dispatched post-commit (mirroring archive), and provider-side trigger bindings (e.g. Gmail watches) that the ORM cascade would otherwise silently orphan are unregistered post-commit on a worker thread. On the frontend, the per-card lifecycle actions move into a three-dot menu on each workforce card.

The transaction shape is genuinely correct, and the hard parts are solved the same way the existing archive/discard paths solve them: the lifecycle fence/lock is reused rather than reinvented, all row deletions happen in a single transaction, and side effects that can fail (PAUSE dispatch, provider unregister) are deferred to post-commit best-effort work. Hard-delete-with-manual-cleanup for `deployments`/`agent_api_keys` is the right call given `Deployment` carries no DB-level FK, and "delete the generated manager only when it is exclusively owned" is a well-reasoned fit that is correctly asymmetric with discard's blocking behavior. The commit-then-best-effort-cleanup ordering (rather than unregister-then-delete) is consistent with archive and I agree with the author's rationale for keeping it.

Two reservations, both design-level rather than bug-level. First, the deletion boundary is drawn in the wrong place: `Workforce.runs` cascades away but each run's `Task` — the actual conversation, trace and workspace-file record — survives and stays visible, so "permanent delete" does not delete the thing the UI promises it deletes (F1). Second, the post-commit trigger-teardown handoff is fragile for no benefit: a detached, already-deleted ORM instance is carried into a *different* Session, when `(trigger_type, config)` tuples would carry everything any provider actually consumes (F3/F4). Separately, note that the cost profile of this endpoint is materially different from archive's: archive was O(1); permanent delete now walks every child collection object-by-object via ORM cascade while holding a whole-DB SQLite writer lock (F7).

---

## Major

### F1 — "Permanent delete" leaves the run `Task` rows behind, contradicting the confirmation dialog
`src/xagent/web/services/workforce_lifecycle.py:297` (and the i18n strings around `en.ts`/`zh.ts:3971`)

The confirm dialog tells the user `"{name}" and its run history will be permanently deleted. This cannot be undone.`, but `delete_workforce_permanently` never touches `Task`. `WorkforceRun.task_id` is `ondelete="SET NULL"` with a plain non-cascading `relationship("Task")`, and workforce-run tasks default to `is_visible=True` — so after a "permanent delete" the conversations, traces and workspace-file records remain fully visible in the user's task list. Please either delete (or at minimum soft-hide, `is_visible=False`) the tasks referenced by the cascaded runs inside the same transaction, or soften the dialog text to say only the workforce and its configuration are removed and that conversations remain. My preference is the former, since a user asking for permanent deletion is usually asking precisely about the conversation content.

### F8 — No permanent-delete test exercises an active run, so the cancellation/PAUSE path is uncovered
`tests/web/api/test_workforces_api.py:939-1326`

All seven permanent-delete tests create at most a `status="completed"` `WorkforceRun`, which is excluded from `ACTIVE_WORKFORCE_RUN_STATUSES`. Consequently `cancel_active_workforce_runs` returns `[]` and `pause_workforce_tasks_after_archive` short-circuits in every test — the in-flight-run cancellation and PAUSE dispatch that the PR description calls out as a delivered feature has zero coverage. The archive path already has the equivalent test at `tests/web/api/test_workforce_deployment_foundation.py:600-641`, which asserts the durable pause command is actually dispatched; please mirror it for the delete path (active run → assert run cancelled in-transaction *and* pause command dispatched post-commit).

### F10 — All five frontend lifecycle-action mocks are wired but never asserted
`frontend/src/app/workforces/workforces-routes.test.tsx`

`deleteWorkforcePermanentlyMock` and `unarchiveWorkforceMock` — and in fact `archiveWorkforceMock`, `publishWorkforceMock` and `unpublishWorkforceMock` as well — are declared, wired into the module mock and reset between tests, but no test ever opens the three-dot menu, clicks an action, or asserts the mock was called. As written they read as coverage while being dead scaffolding. The harness in this same file already has the render pattern, `fireEvent` + `waitFor` + toast assertions and a dialog-interaction example, so this is low-cost: add tests that open the menu and click Unarchive and Delete (through the confirm dialog), asserting the API call and the success toast. Please also cover the pagination step-back logic at `frontend/src/app/workforces/page.tsx:143-152` — it is novel logic carrying its own explanatory comment and currently nothing exercises it.

---

## Minor

### F3 + F4 — The post-commit trigger handoff, and the comment explaining it
`src/xagent/web/services/workforce_lifecycle.py:270-281`, `src/xagent/web/services/triggers.py:826-841`, `src/xagent/web/services/trigger_providers/base.py:131-141`

Two related points in this region.

The comment justifying `db.expunge(trigger)` states that without it, reading an attribute on the deleted-but-still-tracked instance would raise `DetachedInstanceError`/`ObjectDeletedError`. On the pinned SQLAlchemy 2.0.51 that is not what happens — a deleted-and-committed instance moves to `detached` state while retaining its already-loaded values, and reading them does not raise. The `expunge` call is harmless and can stay as defensive hygiene, but please correct or drop the justification so it does not teach the wrong semantics. The comment's other claim, that `Workforce.triggers` "hasn't been loaded yet", is true today but depends on `_load_workforce`'s eager-load set; a one-line note tying the two together would stop a future change there from silently breaking this.

On the handoff itself: a detached, already-deleted `AgentTrigger` is passed alongside a *different*, freshly created Session into `provider.unregister(db, trigger, config)`. Nothing is wrong today — no provider (webhook, gmail) reads `trigger`, and the protocol docstring already tells implementers to resolve everything from `config` — but a future implementer calling `db.add(trigger)`/`db.merge(trigger)` on that foreign session would try to resurrect a deleted row. Cleanest fix is to stop passing the ORM instance at all and hand down `(trigger_type, config)`; if you'd rather keep the signature, make the contract explicit in the `base.py` docstring: the `trigger` argument belongs to a closed session, is already deleted, and must never be added/merged/refreshed.

### F5 — Pause-failure logs from a permanent delete are labelled "archive"
`src/xagent/web/api/workforces.py:831`

`pause_workforce_tasks_after_archive(pause_targets)` is called without `reason=`, so a pause failure during a permanent delete logs as an archive failure. The parameter exists for exactly this and is used correctly elsewhere (e.g. `src/xagent/web/app.py:337-339` passes `reason="preview-reap"`). Pass `reason="delete"` in the permanent branch.

### F2 — No integration coverage for "delete while a turn is in flight"
`src/xagent/web/api/workforces.py:814-841`

The delete commits before the best-effort PAUSE dispatch, so an in-flight turn can briefly execute against rows that are already gone. This is symmetric with archive's intentional, documented pattern, and the failure surface is not opaque — the rejection maps to a dedicated user-facing `workforce_run_not_found` ("This workforce conversation is no longer available; please start a new conversation."), the same class of message as archive's `workforce_run_not_active`. So no change to the ordering is needed; the ask is just a dedicated integration test that deletes mid-turn and asserts the turn is rejected with that specific error rather than a 500.

### F6 — Provider teardown is awaited on the request's critical path after the data is already gone
`src/xagent/web/api/workforces.py:832-840`

For N Gmail triggers this is N sequential network round-trips (Gmail API + Pub/Sub, each with its own commit) delaying the DELETE response *after* the rows are committed as deleted. The frontend fetch wrapper sets no client timeout, so the common case is a slow-but-eventual 200; but if an upstream proxy/gateway timeout or a dropped connection lands first, the user sees "Failed to delete workforce" (`frontend/src/app/workforces/page.tsx:153-156`) for an operation that fully succeeded. Trigger counts per workforce are typically low so the practical risk is small, but consider either fanning the teardowns out concurrently (`asyncio.gather` over per-trigger threads) or moving them fully off the request path into background work, so the response returns as soon as the commit lands.

### F7 — New cost profile: full ORM cascade under the whole-DB SQLite writer lock
`src/xagent/web/services/workforce_lifecycle.py:297-298`

`db.delete(workforce)` cascades `workers`/`runs`/`builder_messages`/`triggers` — and transitively `TriggerRun`/`TriggerAudit` — object-by-object, with no `passive_deletes=True`, all inside the transaction holding `acquire_workforce_lifecycle_fence`'s lock, which the function's own docstring notes is a whole-DB writer lock on SQLite (the default backend here). Archive never deleted anything, so this is genuinely new: a long-lived workforce with a busy webhook trigger accumulates audit rows per callback and could turn one delete into a multi-second global write stall. Worth adding `passive_deletes=True` on the high-volume child relationships (with the matching DB-level `ondelete`) so the cascade is one statement per table rather than one per row, or at least noting the expected volume bound.

### F9 — The real body of `unregister_deleted_trigger_bindings` is never exercised
`tests/web/api/test_workforces_api.py:1188-1190`

`test_permanent_delete_unregisters_cascade_deleted_trigger_bindings` monkeypatches `unregister_deleted_trigger_bindings` wholesale, so the function body added in `5c412ed5` — `get_session_local()`, the per-trigger `with SessionLocal() as db:` isolation, delegation to `_unregister_trigger_binding`, and the per-item `except Exception` swallow — has no coverage at all. Please add a test that keeps the real function and stubs only the provider's `unregister`, asserting it is called once per trigger with the right `config`, and that one provider raising does not prevent the remaining triggers from being torn down.

### F12 — Assorted uncovered branches
`tests/web/api/test_workforces_api.py`

Five small gaps worth closing:
- (a) The `used_as_worker` branch of manager-safety checking is untested on both the discard and permanent-delete paths; only `shared_reference` and `owner_mismatch` are covered, and only on discard.
- (b) No 404 test for permanent-delete of a nonexistent workforce id.
- (c) No 404 test for unarchive of a nonexistent workforce id, though the path exists at `src/xagent/web/api/workforces.py:872-873`.
- (d) No test asserts on `TriggerRun`/`TriggerAudit` rows after a cascade delete, so the audit-preserving nullify behavior at `src/xagent/models/trigger.py:140-143` is unverified. This matters more than it looks: SQLite in tests runs without FK enforcement, so the ORM cascade is the only thing doing this cleanup.
- (e) No test drives two lifecycle operations at the fence concurrently. This is a pre-existing gap that archive shares, not something this PR introduced — flagging only as a "consider adding" while you are in here.

### F11 — `test_permanent_delete_preserves_reusable_manager` doesn't test what its name says
`tests/web/api/test_workforces_api.py:1099-1114`

Its manager is `USER`-origin (via the default `_create_workforce` helper), so it hits `_lock_generated_manager`'s early short-circuit and never reaches the exclusive-ownership/shared-manager logic. Those real branches *are* covered by two other tests in this file, so this is not a coverage hole — but the name is misleading. Rename to something like `test_permanent_delete_skips_user_owned_manager`, or extend it to actually build a shared generated manager.

### F13 — Three-dot menu stays open after an action and reflows, creating a mis-click
`frontend/src/app/workforces/page.tsx:277-343`

The `Popover` is uncontrolled with no `PopoverClose` and no close-on-click, so after clicking Archive the menu remains open — and because the Publish/Unpublish block is conditionally rendered on `item.status`, the child list actually *changes length and reflows* once `load()` resolves. A second click at the same screen position after archiving therefore lands on **Delete**, not on Unarchive as the previous layout would suggest. It is mitigated by Delete only opening a confirm dialog, but it is reproducible and worth fixing: either close the popover on any action, or hold a controlled `open` state and reset it in each handler's `finally`. Note the equivalent agent-card popover in `build/page.tsx` shares the uncontrolled shape but never reflows its row count, so this particular combination (status-conditional children + no close-on-click) is a materially bigger risk than the pre-existing pattern.

### F16 — Renaming the handler changes the generated OpenAPI `operationId`
`src/xagent/web/api/workforces.py:808`

`archive_workforce` → `archive_or_delete_workforce` changes the `operationId` for `DELETE /api/workforces/{workforce_id}`, since there is no explicit `operation_id=` override. Nothing in-repo consumes the old name, so impact is limited to external consumers of the OpenAPI schema — but if any exist, this is a breaking rename. Either pin `operation_id="archive_workforce"` on the route decorator to keep the schema stable, or call the rename out in the PR/changelog notes.

### F14 — Archiving a non-archived workforce now requires leaving its detail page
`frontend/src/components/workforce/workforce-builder.tsx:745`

The removed standalone button was gated on `!isArchived`, so it was already invisible for an archived workforce and there was never an unarchive affordance here — the "can't unarchive from the builder" gap is pre-existing, not caused by this PR. What did regress, mildly, is that a non-archived workforce used to be archivable inline from its own detail page and now requires navigating back to the list. Minor UX nit; consider keeping an archive entry in the builder header, or ignore if the list page is the intended single home for lifecycle actions.

---

## Nit

`frontend/src/components/workforce/workforce-builder.tsx` — a stray blank line was added near the closing `</div>` at the end of the file, unrelated to this change. Worth dropping to keep the diff clean.

---

One carried-over item from the earlier review round that was never addressed: the `handlePublish`/`handleUnpublish` duplication in `frontend/src/app/workforces/page.tsx`. The two handlers differ only in the API function they call, the pending-state slot they set, and the toast strings, so a small `createPublishHandler({ action, successMessage })` factory (as previously suggested) would collapse them. Now that `handleArchive`, `handleUnarchive` and `handleDelete` sit alongside them with the same try/`finally`-clear/toast shape, the factory earns its keep more than it did before. **net: ~-25 lines possible.**

Note that consolidating the separate `publishingId`/`archivingId`/`unarchivingId` pending slots into one global `pendingAction` is *not* a safe follow-on: the current per-action slots deliberately allow concurrent actions on different cards to be tracked independently, and collapsing them would make buttons stop rendering as pending while their request is still in flight. Keep the slots as they are.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +212,126 @@ def discard_draft_workforce(
...
+    teardown data (trigger, type, config) -- this function stays sync and
+    does not perform the actual provider unregister calls itself: it runs
+    directly on the request's event-loop thread (its caller is an ``async
+    def`` route, not one FastAPI offloads to a worker thread), so any
+    provider network I/O has to happen through the caller's own
+    ``asyncio.to_thread`` the same way :func:`pause_workforce_tasks_after_archive`
+    already does, or it would block that thread for every concurrent request.
+    """
+
+    workforce = _ensure_discard_access(db, user, workforce)
```
rogercloud: Minor (N6): `_ensure_discard_access` (defined at :55) and `is_workforce_manager_discard_safe` (:31) now gate permanent delete as well as discard, but keep discard-specific names. The discard/delete distinction is precisely what this PR adds, so a future maintainer reading this call site could misread which rules apply. Consider renaming to something lifecycle-neutral, e.g. `_ensure_workforce_lifecycle_access` / `is_workforce_manager_removal_safe`.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +212,126 @@ def discard_draft_workforce(
                 "Failed to invalidate the discarded Workforce manager cache",
                 exc_info=True,
             )
+
+
+def delete_workforce_permanently(
```
rogercloud, line 224: Minor (N7): this function duplicates the ~26-line transactional envelope of `discard_draft_workforce` (:158-216) — entry + refetch under the fence, `db.delete` + flush, manager stage-delete, commit, rollback, post-commit cache invalidation. Extracting shared helpers would shrink both paths and keep their error handling consistent; only the middle sections genuinely differ.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +212,126 @@ def discard_draft_workforce(
...
+        # the caller dispatch PAUSE after this transaction commits.
+        pause_targets = cancel_active_workforce_runs(db, workforce_id)
+        generated_manager, manager_deletable = _lock_generated_manager(db, workforce)
+
+        # The ORM cascade below deletes trigger rows without the trigger CRUD
+        # path's provider teardown, which would leak provider-side bindings
+        # (Gmail watches etc.). Capture what teardown needs while the rows
+        # are still readable; the actual unregister runs after commit, same
+        # ordering as _delete_trigger.
+        trigger_teardowns: list[tuple[AgentTrigger, str, dict[str, Any]]] = []
```
rogercloud, line 293: Minor (N3): this teardown capture SELECT races concurrent trigger creation. A trigger created after this capture but before commit is cascade-deleted without a teardown entry, orphaning its provider-side binding (e.g. a Gmail watch) silently and unrecoverably — trigger CRUD takes no lifecycle fence. Either have workforce-trigger creation acquire the same fence, or accept and document the race in this comment.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +212,126 @@ def discard_draft_workforce(
...
+            # quietly reintroduce a footgun here.
+            db.expunge(trigger)
+            trigger_teardowns.append((trigger, trigger_type, config))
+
+        # WorkforceRun.task_id is ondelete="SET NULL" with a non-cascading
+        # Task relationship, and workforce-run tasks default to
+        # is_visible=True -- so without this, the conversations/traces a
+        # "permanent delete" promises to remove would stay fully visible in
+        # history/search after the workforce and its runs are gone. Hiding
+        # rather than deleting the Task rows mirrors stage_delete_agent's
```
rogercloud: Minor (N5): the comment claims hiding Task rows "mirrors stage_delete_agent's own tasks cleanup (detach/hide)", but `stage_delete_agent` (agent_store.py:561-568) detaches tasks and never hides them — only "never hard-delete a Task" is shared. Reword so the wrong precedent isn't cited as a pattern later.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +212,126 @@ def discard_draft_workforce(
...
+        # table) cannot be relied on here.
+        db.query(AgentApiKey).filter(AgentApiKey.workforce_id == workforce_id).delete(
+            synchronize_session=False
+        )
+        db.query(Deployment).filter(
+            Deployment.owner_type == DeploymentOwnerType.WORKFORCE.value,
+            Deployment.owner_id == workforce_id,
+        ).delete(synchronize_session=False)
+
+        db.delete(workforce)
```
rogercloud, line 352: Blocking (prior R3-F7, major — unchanged this round): this `db.delete(workforce)` + flush walks every child relationship row-by-row. All four relationships still cascade with no `passive_deletes` (models/workforce.py:50-69; TriggerRun/TriggerAudit same at models/trigger.py:136-141), runs are loaded twice (cancel sweep + cascade hydration), and this sync function runs directly on the async route's event-loop thread — so a workforce with large run/trigger history emits N+1 DELETEs and stalls every concurrent request on the SQLite writer lock. Please set `passive_deletes=True` on the high-volume relationships (child FKs already `ON DELETE CASCADE`), or state the expected scale in the docstring if bounded.

src/xagent/web/api/workforces.py
```diff
@@ -801,12 +804,42 @@ async def update_workforce(
     )
 
 
-@router.delete("/{workforce_id}")
-async def archive_workforce(
+@router.delete("/{workforce_id}", operation_id="archive_workforce")
+async def archive_or_delete_workforce(
     workforce_id: int,
+    permanent: bool = False,
```
rogercloud: Nit (N12): `permanent` is the destructive switch of this endpoint but has no `Query(...)` description, so it is invisible in the OpenAPI docs. Add one (e.g. "Permanently delete the workforce and all run history; cannot be undone").

src/xagent/web/api/workforces.py
```diff
@@ -801,12 +804,42 @@ async def update_workforce(
...
+                },
+            ) from None
+        await pause_workforce_tasks_after_archive(pause_targets, reason="delete")
+        if trigger_teardowns:
+            # Provider unregister (e.g. releasing a Gmail watch) can do real
+            # network I/O; offload to a worker thread instead of blocking
+            # every other concurrent request on the event loop. Does NOT
+            # pass this request's `db` through -- a Session is not safe to
+            # hand to a background thread, so the callee opens its own.
+            await asyncio.to_thread(
```
rogercloud: Minor (N2): setup inside `unregister_deleted_trigger_bindings` — `get_session_local()` (triggers.py:834) and `ThreadPoolExecutor` construction — runs after the delete has committed. A failure there propagates through `asyncio.to_thread` as a 500 for a delete that already succeeded, contradicting the best-effort contract in the docstring (triggers.py:807-828). Wrap this whole teardown await in try/except (log, still return the success response), or move setup into the per-item tolerant portion.

src/xagent/web/api/workforces.py
```diff
@@ -834,6 +867,42 @@ async def archive_workforce(
...
+    workforce = _load_workforce(db, workforce_id)
+    if workforce is None:
+        raise HTTPException(status_code=404, detail="Workforce not found")
+    if not can_edit_workforce(db, user, workforce):
+        raise HTTPException(status_code=403, detail="Access denied")
+    workforce_id_value = int(workforce.id)
+    # Same lifecycle fence as archive: serializes against a concurrent
+    # archive/delete so the status check below reads committed state instead
+    # of racing the other transaction's flip.
+    if acquire_workforce_lifecycle_fence(db, workforce_id_value) is None:
```
rogercloud: Nit (N13): unlike delete/discard, this route discards the fence's return value and keeps using the pre-fence `workforce` instance. It works today only because the fence's `populate_existing` re-query refreshes the session identity map — reassigning the return (as delete_workforce_permanently does) would make the dependency explicit.

src/xagent/web/api/workforces.py
```diff
@@ -834,6 +867,42 @@ async def archive_workforce(
...
+    if workforce.status != "archived":
+        raise HTTPException(
+            status_code=409,
+            detail="Only archived workforces can be unarchived",
+        )
+    # Restore to draft, not active: runs were cancelled at archive time and
+    # the config may have gone stale, so re-activation stays an explicit
+    # publish step that re-runs validation.
+    cast(Any, workforce).status = "draft"
+    db.commit()
```
rogercloud: Minor (N4): bare `db.commit()` — a commit failure here surfaces as a generic 500 with no error code, inconsistent with the hardened try/rollback + `detail.code` pattern added for delete-permanent and discard in this same PR. Mirror that pattern for consistency.

frontend/src/app/workforces/page.tsx
```diff
@@ -72,6 +96,81 @@ export default function WorkforcesPage() {
     void load(page, search)
   }, [load, page, search])
 
+  const createPublishHandler = (publish: boolean) => async (item: WorkforceListItem) => {
```
rogercloud: Minor (N9): the per-action in-flight flags (`publishingId`, `archivingId`, ...) don't cross-guard — while publish is in flight, archive/unarchive/delete on the same card stay clickable and the menu reopens instantly, so the outcome depends on backend arrival order (success+error toast pairs), and the delete item has no disabled binding at all. Consider a single per-card busy state, or guard each handler against any in-flight action on the same item.

frontend/src/app/workforces/page.tsx
```diff
@@ -72,6 +96,81 @@ export default function WorkforcesPage() {
...
+        : t(publish ? "workforces.errors.publish" : "workforces.errors.unpublish")
+      toast.error(nextError)
+    } finally {
+      setPublishingId(null)
+    }
+  }
+  const handlePublish = createPublishHandler(true)
+  const handleUnpublish = createPublishHandler(false)
+
+  const handleArchive = async (item: WorkforceListItem) => {
```
rogercloud: Partial (prior R1-2): `handleArchive`/`handleUnarchive` still duplicate the exact try/`finally`-clear/toast shape that `createPublishHandler` (:99-117) was extracted to cover — they differ only in API call, busy slot and toast strings. Extend the factory (or a generic `runMenuAction`) to cover these two as well; keep the busy slots as parameters.

frontend/src/app/workforces/workforces-routes.test.tsx
```diff
@@ -139,9 +141,30 @@ vi.mock("@/lib/api-wrapper", () => ({
   apiRequest: apiRequestMock,
 }))
 
+vi.mock("@/components/ui/confirm-dialog", () => ({
```
rogercloud, line 145: Minor (N10): the ConfirmDialog and Popover mocks strip the interaction layer — `openMenuId` machinery is never exercised, dialog title/description/confirmText/isLoading never render (so even the new delete test asserts nothing about dialog contents), and there are no negative per-status assertions (archived card shows no Archive/Publish rows; draft card shows no Unarchive). Rendering the real dialog (or a fuller mock) and asserting per-status menu contents would close most of the remaining F10 gap.

frontend/src/app/workforces/workforces-routes.test.tsx
```diff
@@ -601,6 +627,94 @@ describe("workforce route entry points", () => {
...
+
+    await waitFor(() => {
+      expect(deleteWorkforcePermanentlyMock).toHaveBeenCalledWith(42)
+    })
+    await waitFor(() => {
+      expect(listWorkforcesMock).toHaveBeenCalledTimes(2)
+    })
+  })
+
+  it("steps pagination back when deleting the last card on a page beyond the first (page.tsx:143-152)", async () => {
```
rogercloud: Minor (N11): the test name embeds a stale line anchor `(page.tsx:143-152)` — `handleDelete` is now at page.tsx:149-172. Drop the anchor or update it; as written it will keep drifting on every edit.

frontend/src/lib/workforces-api.ts
```diff
@@ -167,6 +167,30 @@ export async function archiveWorkforce(
...
+  const response = await apiRequest(`${getApiUrl()}/api/workforces/${workforceId}/unarchive`, {
+    method: "POST",
+  })
+  if (!response.ok) {
+    throw await parseApiError(response, "Failed to unarchive workforce")
+  }
+  return response.json()
+}
+
+export async function deleteWorkforcePermanently(
```
rogercloud, line 189: Minor (N8): `deleteWorkforcePermanently` (and `unarchiveWorkforce` at :170) have no client-level test — workforces-api.test.ts is untouched by this PR and the routes tests mock the whole module, so the `?permanent=true` query and DELETE/POST verbs are unverified at the wire level. Mirror the existing `archiveWorkforce` URL/method test pattern.

rogercloud (changes requested):
# Re-review: `feat(workforce): support permanent delete and unarchive` (round 4)

PR adds two lifecycle capabilities for workforces: (1) `DELETE /api/workforces/{id}?permanent=true` — hard-deletes a workforce in any status (draft/active/archived), cascading workers, run history, builder messages and triggers; cancels in-flight runs in-transaction with PAUSE dispatched post-commit (mirroring archive); manually cleans up workforce-scoped API keys and the FK-less `Deployment` row; deletes the auto-generated manager agent only when exclusively owned (shared/foreign manager kept — deliberately unlike discard, which blocks); post-commit, provider-side trigger bindings (e.g. Gmail watches) are unregistered via `asyncio.to_thread` with per-trigger sessions and per-item failure isolation. (2) `POST /api/workforces/{id}/unarchive` — restores an archived workforce to `draft` so re-activation remains an explicit, re-validating publish step.

## Update summary (`33ce91bf` → `fbb060ef`)

One commit (`fbb060ef`, "fix(workforce): address PR #1277 review round 3 findings") addressing round 3:
- Hides run `Task` rows in-transaction (F1), passes `reason="delete"` to PAUSE dispatch (F5), rewrites the expunge comment with correct SQLAlchemy detached semantics (F3), pins `operation_id="archive_workforce"` (F16), fans provider teardown out via `ThreadPoolExecutor` capped at 8 (F6).
- Adds the active-run cancel+PAUSE test (F8), the real-function per-trigger teardown isolation test (F9), renames the misleading manager test (F11), and adds frontend menu-click, confirm-dialog and pagination step-back tests plus a controlled Popover with eager close (F10, F13).
- Not addressed: ORM cascade cost / `passive_deletes` (F7), mid-turn delete integration test (F2), F12's five coverage gaps, detached-instance handoff (F4), builder archive UX (F14), the stray blank line.

CI is green.

## Prior-findings checklist

| ID | Status | Evidence |
|---|---|---|
| R3-F1 (major) tasks left behind | FIXED | `delete_workforce_permanently` bulk-updates `Task.is_visible=False` in-transaction (workforce_lifecycle.py:299-301); foundation test asserts the dialog contract |
| R3-F8 (major) no active-run test | FIXED | `test_permanent_delete_cancels_active_runs_and_dispatches_pause` (test_workforce_deployment_foundation.py:645-703): RUNNING run removed in-transaction + durable PAUSE command `workforce-pause-{run_id}` |
| R3-F10 (major) frontend mocks never asserted | PARTIAL | Unarchive (:630-653), delete-through-confirm (:655-676) and pagination step-back (:678-716) tests assert `unarchiveWorkforceMock`/`deleteWorkforcePermanentlyMock`; success-toast still never asserted; `archiveWorkforceMock`/`publishWorkforceMock`/`unpublishWorkforceMock` remain wired-but-unasserted |
| R3-F2 (minor) no mid-turn delete test | NOT FIXED | No test deletes mid-turn and asserts `workforce_run_not_found`; only the pre-existing unit guard `test_turn_guard_rejects_missing_run_row` |
| R3-F3 (minor) expunge comment | FIXED | Comment rewritten with correct pinned-SQLAlchemy detached semantics + eager-load-set note (workforce_lifecycle.py:272-284) |
| R3-F4 (minor) detached-instance handoff | NOT FIXED | Handoff still `(AgentTrigger, type, config)` tuples (workforce_lifecycle.py:263; triggers.py:804-805); base.py:131-143 docstring contract unextended. Latent only — no provider reads the instance (gmail config-only, webhook no-op) |
| R3-F5 (minor) reason="delete" | FIXED | `pause_workforce_tasks_after_archive(pause_targets, reason="delete")` (workforces.py:831) |
| R3-F6 (minor) teardown latency | FIXED | `ThreadPoolExecutor(max_workers=min(len(teardowns), 8))` fan-out (triggers.py:851); residual "response waits for all" is inherent to the finding's own option A |
| R3-F7 (major) ORM cascade cost | NOT FIXED | **Blocking** — see blocking section |
| R3-F9 (minor) real body untested | FIXED | `test_unregister_deleted_trigger_bindings_isolates_per_trigger_failures` (test_workforces_api.py:1219-1307): real function + stubbed provider, per-item isolation asserted |
| R3-F11 (minor) misleading test name | FIXED | Renamed `test_permanent_delete_skips_non_generated_manager` (:1088) with clarifying docstring |
| R3-F12 (minor) five coverage gaps | NOT FIXED | 0/5: used-as-worker branch, permanent-delete 404, unarchive 404, TriggerRun/TriggerAudit post-cascade state (audit-preserving nullify unverified), concurrent-fence test |
| R3-F13 (minor) popover reflow mis-click | FIXED | Controlled `Popover open={openMenuId === item.id}` + `setOpenMenuId(null)` eager in every handler (page.tsx:290-292) |
| R3-F14 (minor) builder archive UX | NOT FIXED | Builder untouched; archive entry not restored; no explicit "list page is the single home" statement |
| R3-F16 (minor) operationId | FIXED | `operation_id="archive_workforce"` pinned (workforces.py:807) |
| R3-nit stray blank line | NOT FIXED | workforce-builder.tsx:867 — introduced by b4210a5a (the PR's first commit); drop it to keep the diff clean |
| R2-1 (high) unregister-before-delete ordering | WAIVED | Round-3 verdict endorsed commit-then-best-effort; DetachedInstanceError sub-claim resolved by the F3/F4 work |
| R2-2 (high) detached instances across threads | FIXED | Expunge-before-commit + own `SessionLocal` per thread; base.py protocol docstring requires config-only resolution |
| R1-1 (critical) request session across threads | FIXED | `db` no longer passed to `unregister_deleted_trigger_bindings` (5c412ed5); per-trigger `SessionLocal` |
| R1-2 (medium) handler duplication | PARTIAL | `createPublishHandler` factory covers publish/unpublish only (page.tsx:99-117); `handleArchive`/`handleUnarchive` still duplicate the shape (:119-147) |

## New findings

**Major** — none new this pass: the only major (N1, N+1 ORM cascade + event-loop blocking) is the same underlying issue as prior R3-F7 and is tracked there (see blocking section).

**Minor**

- **N2 — teardown setup failures surface as 500 after a successful delete** (workforces.py:838-840, triggers.py:834). `get_session_local()` and `ThreadPoolExecutor` construction run inside `unregister_deleted_trigger_bindings`, which executes after commit — a `RuntimeError` in setup escapes `asyncio.to_thread` and turns a delete that already committed into a 500, contradicting the best-effort promise in the docstring (triggers.py:807-828). Suggestion: wrap the whole teardown await in try/except at the route (log, still return 200), or move setup inside the per-item tolerant portion.
- **N3 — trigger-teardown capture races concurrent trigger creation** (workforce_lifecycle.py:263-284). A trigger created after the capture SELECT but before commit is cascade-deleted without a teardown entry → orphaned provider binding (e.g. Gmail watch), silently and unrecoverably. Trigger CRUD takes no lifecycle fence. Suggestion: have workforce-trigger creation acquire the fence, or accept and document the race.
- **N4 — unarchive commit lacks rollback guard** (workforces.py:900). Bare `db.commit()`; inconsistent with the hardened delete-permanent/discard error-code patterns added in this same PR. No integrity risk (session close rolls back), but a serialization error mid-commit surfaces as a generic 500. Mirror the try/rollback pattern.
- **N5 — comment misattributes stage_delete_agent behavior** (workforce_lifecycle.py:290-292). The comment claims task hiding "mirrors stage_delete_agent's own tasks cleanup (detach/hide)", but `stage_delete_agent` (agent_store.py:561-568) detaches tasks and never hides them; only "never hard-delete a Task" is shared. Reword so the wrong precedent isn't cited later.
- **N6 — discard-specific naming now serves permanent delete** (workforce_lifecycle.py:55, 31). `_ensure_discard_access` / `is_workforce_manager_discard_safe` are the access/manager gate for permanent delete too; rename (e.g. `_ensure_workforce_lifecycle_access`) to match.
- **N7 — ~26-line transactional envelope duplicated** (workforce_lifecycle.py:217-337 vs 158-216). Entry+refetch, delete+flush, manager stage, commit, rollback, cache-invalidate are duplicated between discard and permanent delete. Extract shared helpers.
- **N8 — new API client functions untested** (workforces-api.ts:170-192). `deleteWorkforcePermanently`/`unarchiveWorkforce` have no client-level test; workforces-api.test.ts is untouched by this PR and the routes tests mock the whole module. Mirror the existing `archiveWorkforce` URL/method test pattern.
- **N9 — per-action in-flight flags don't cross-guard** (page.tsx:99-172). While publish is in flight, archive/unarchive/delete on the same card stay clickable (menu reopens instantly); the outcome depends on backend arrival order (success+error toast pairs). Delete's menu item has no disabled binding at all. Suggest a single per-card busy state or a guard in each handler.
- **N10 — tests strip the interaction layer** (workforces-routes.test.tsx:144-162). Mocks render menu items unconditionally, so `openMenuId` machinery is never exercised; the ConfirmDialog mock renders only `isOpen`/`onConfirm`, so dialog title/description/confirmText/isLoading never render (even the new delete test can't assert dialog contents); no negative per-status assertions (archived card: no Archive/Publish rows; draft card: no Unarchive).
- **N11 — stale line anchor in test name** (workforces-routes.test.tsx:678). `"(page.tsx:143-152)"` no longer matches; `handleDelete` is at page.tsx:149-172.

**Nit**

- **N12 — `permanent` flag undocumented in OpenAPI** (workforces.py:810). `permanent: bool = False` has no `Query(...)` description, so the destructive flag is invisible in API docs.
- **N13 — unarchive discards the fence's return value** (workforces.py:889). Correct today via `populate_existing` re-query, but the route relies on implicit session machinery; delete/discard reassign the return — do the same for consistency.

**Out of scope (pre-existing, body note)**

- **N14 — agent delete leaks provider bindings** (agent_management.py:509). Agent deletion cascades `AgentTrigger` rows with no provider unregister — Gmail watches leak on agent delete. Pre-existing at base; `unregister_deleted_trigger_bindings` could serve this path in a follow-up.

- `triggers.py:829` — delete the dead guard: sole caller (workforces.py:832) already checks truthiness, so `if not teardowns: return` can never fire.
- `page.tsx:119` — shrink `handleArchive`/`handleUnarchive` into a generic `runMenuAction(apiCall, successKey, errorKey, setBusy)` covering all four non-delete actions (keep per-action busy slots as parameters). ~-24 lines.
- `triggers.py:855` — `list(pool.map(...))` is redundant: the with-block `shutdown(wait=True)` already joins workers and `_teardown_one` swallows exceptions, so bare `pool.map(...)` preserves behavior; also reword the comment at 852-854, which attributes the wait to the wrong mechanism.

**net: ~-26 lines possible**

- **Blocking: yes.** The only major is the unresolved prior finding, verified NOT FIXED twice this round:
  - **[prior] R3-F7 (major)** — `workforce_lifecycle.py:314-315`: ORM per-row cascade under the lifecycle fence + event-loop blocking. All four relationships still `cascade="all, delete-orphan"` without `passive_deletes` (models/workforce.py:50-69; TriggerRun/TriggerAudit same, models/trigger.py:136-141); per-row DELETE per child row; runs loaded twice (cancel sweep + cascade hydration); and `delete_workforce_permanently` runs synchronously on the async route's event-loop thread, so a large workforce stalls every concurrent request on the SQLite writer lock. Suggestion: `passive_deletes=True` on the high-volume relationships (child FKs already `ON DELETE CASCADE`) or state the expected scale in the docstring if bounded.
- Everything else this pass is minor/nit/simplification.
- **Recommended event: REQUEST_CHANGES** (author: yiboyasss).

yiboyasss:
Thanks for another thorough pass! Pushed fixes in f8d278b9 for the naming, event-loop-blocking, error-handling, and frontend cross-guard findings. Three items I left as-is, with reasoning:

- **`passive_deletes=True` suggestion (the carried-over blocking item)**: not adopted. This project's SQLite path never enables `PRAGMA foreign_keys=ON` (checked `models/database.py`) — the same reason `AgentApiKey`/`Deployment` already need an explicit `.delete()` a few lines above instead of relying on their `ON DELETE CASCADE`. Setting `passive_deletes=True` on `Workforce`'s `workers`/`runs`/`builder_messages`/`triggers` relationships would tell the ORM to stop deleting those rows itself and lean on the DB's FK cascade instead — on SQLite that cascade isn't actually enforced, so it would silently stop cleaning up those rows rather than just being slower. That's worse than the N+1 cost it would remove, and it's also a config shared with `discard_draft_workforce`'s own `db.delete(workforce)`, not something scoped to this PR's new code. I did fix the part of the finding that's safe to fix without touching that shared model config: `delete_workforce_permanently` is now dispatched via `asyncio.to_thread` from the route, so the cascade no longer runs on the event-loop thread regardless of workforce size. Documented both the acceptance and the reasoning at the `db.delete(workforce)` call site.

- **Widening the lifecycle fence to cover trigger creation (N3's other option)**: went with documenting the race instead. `create_workforce_trigger`/`_create_trigger` is a shared, actively-used path with no relationship to this workforce-lifecycle feature; extending its locking behavior to close a narrow window (a trigger create landing between this capture SELECT and the delete's commit) risked new contention on unrelated, higher-traffic code for a low-probability race. Documented the gap explicitly in the capture loop instead.

- **Extracting the shared discard/delete transactional envelope (N7)**: not done. It's a real duplication (~26 lines: fence re-entry, `db.delete` + flush, manager stage-delete, commit/rollback, post-commit cache invalidation), but both functions are already-tested, security-relevant deletion paths, and the risk of a subtle behavioral drift introduced while extracting shared error-handling outweighed the maintainability win for a review-fix pass. Left as a known follow-up rather than refactoring both paths under time pressure.

Also left N10 partially closed: added negative per-status menu-row assertions to the existing (lightweight-mock) tests rather than switching to full real-Popover/ConfirmDialog rendering — the fuller rendering swap is a bigger lift that felt disproportionate to what's still a coverage-completeness nit, not a correctness bug.

src/xagent/web/services/workforce_lifecycle.py
```diff
@@ -189,3 +218,157 @@ def discard_draft_workforce(
...
+        ).delete(synchronize_session=False)
+
+        # This ORM cascade (workers/runs/builder_messages/triggers, all
+        # cascade="all, delete-orphan" on Workforce -- models/workforce.py)
+        # loads every child row into Python and deletes them one by one
+        # rather than a single DB-side statement, which is the same
+        # already-accepted cost discard_draft_workforce's own db.delete(
+        # workforce) has always paid for the same relationships. Adding
+        # passive_deletes=True to lean on DB-level ON DELETE CASCADE instead
+        # would fix that here, but this project's SQLite path never enables
```
rogercloud: The rationale for rejecting `passive_deletes` here rests on a factual error: this repo does enable FK enforcement on the SQLite path (`PRAGMA foreign_keys=ON` in `src/xagent/db/sqlite.py:100`, applied via `apply_sqlite_concurrency_pragmas` from `web/models/database.py:190`; `web/models/task.py:117-118` confirms FK-on is the repo default). Every child FK is `ON DELETE CASCADE` — WorkforceAgent (`workforce.py:81`), WorkforceRun (`:127`), WorkforceBuilderMessage (`:177`), AgentTrigger.workforce_id (`trigger.py:93`), TriggerRun (`:161`) — and AgentApiKey.workforce_id is also `ondelete="CASCADE"` (`agent_api_key.py:38`); only Deployment has no FK, so the "AgentApiKey needs explicit delete" sub-rationale is wrong too. `passive_deletes=True` is safe here and would eliminate the N+1 cascade and the double run load. Please either add `passive_deletes=True` or correct the comment to "FK enforcement is on; keeping the N+1 is a pure scale tradeoff".

src/xagent/web/api/workforces.py
```diff
@@ -834,6 +906,59 @@ async def archive_workforce(
...
+@router.post("/{workforce_id}/unarchive")
+async def unarchive_workforce(
+    workforce_id: int,
+    db: Session = Depends(get_db),
+    user: User = Depends(get_current_user),
+) -> dict[str, Any]:
+    # Not ensure_workforce_access(..., action="edit"): that helper's archived
+    # -> 409 guard exists precisely to keep archived workforces read-only, so
+    # it would reject the one workforce this route is meant to act on. Check
+    # the underlying edit permission directly instead.
```
rogercloud: This 404/403 pair is byte-identical to `_ensure_workforce_lifecycle_access` (`workforce_lifecycle.py:60-70`). That helper has no archived→409 guard, so it can be reused directly (or promoted to public) instead of inlining a duplicate. The route comment only explains why `ensure_workforce_access` isn't used — it doesn't justify the duplication.

src/xagent/web/api/workforces.py
```diff
@@ -801,12 +806,79 @@ async def update_workforce(
...
+        try:
+            pause_targets, trigger_teardowns = await asyncio.to_thread(
+                _load_and_delete_permanently
+            )
+        except HTTPException:
+            raise
+        except Exception:
+            logger.exception("Failed to delete workforce %s", workforce_id)
+            db.rollback()
+            raise HTTPException(
```
rogercloud, line 848: The object-form 500 `detail = {code, message}` carries a `message` the frontend never extracts: `parseApiError`/`getApiErrorMessage` surface raw JSON (`parsed.text`) on the delete/unarchive paths, so users see raw JSON rather than this payload or the fallback. The new `toThrow('Failed to delete workforce')` test passes by substring-matching raw JSON and does not prove extraction. Suggest `formatApiDetail` read `detail.message`, or return a code-only envelope and assert the actually surfaced copy.

frontend/src/app/workforces/page.tsx
```diff
@@ -44,6 +53,24 @@ export default function WorkforcesPage() {
...
   const [deployView, setDeployView] = useState<DeployView | null>(null)
   const [triggersItem, setTriggersItem] = useState<WorkforceListItem | null>(null)
+  const [deleteItem, setDeleteItem] = useState<WorkforceListItem | null>(null)
+  const [deleting, setDeleting] = useState(false)
+  // Single slot, not one flag per action: publish/unpublish/archive/
+  // unarchive/delete on the same card must not overlap -- with separate
+  // flags, publish being in flight left archive/unarchive/delete on that
+  // same card still clickable, so the outcome depended on whichever
+  // request the backend happened to finish first.
+  const [busyItemId, setBusyItemId] = useState<number | null>(null)
```
rogercloud, line 70: `busyItemId` gates only the menu trigger and its items — the card's Run/Deploy/Edit buttons stay clickable during an in-flight action and the backend rejects, producing avoidable confusing toasts (the run path rejects via the `WorkforceRunError` 400/409 mapping, not `ensure_workforce_access`). Consider gating these actions too; note Run/Edit are `asChild` Links, so use `aria-disabled`/`pointer-events` rather than disabling the button.

frontend/src/app/workforces/page.tsx
```diff
@@ -72,6 +99,67 @@ export default function WorkforcesPage() {
...
+
+  const handleDelete = async () => {
+    if (!deleteItem) return
+    try {
+      setDeleting(true)
+      setBusyItemId(deleteItem.id)
+      await deleteWorkforcePermanently(deleteItem.id)
+      toast.success(t("workforces.messages.deleted"))
+      setDeleteItem(null)
+      if (items.length === 1 && page > 1) {
```
rogercloud: The pagination fallback reads the render-closure `items`; when concurrent loads resolve out of order (a menu action's `void load` racing a search change), `items` can belong to another page/search and the error branch may reload an out-of-range page — the exact empty-state bug this PR fixes. The window is narrow, but a stale-response guard (request sequence number) is more robust than `itemsRef` alone.

frontend/src/i18n/locales/en.ts
```diff
@@ -3964,6 +3967,11 @@ Build when you need.`,
       change: "Change",
       addAgent: "Add Agent",
     },
+    delete: {
+      confirmTitle: "Delete this workforce?",
+      confirmDescription: "\"{name}\" and its run history will be permanently deleted. This cannot be undone.",
```
rogercloud: The confirm copy discloses that run history is removed but not that in-progress runs will be stopped, and delete renders unconditionally for active workforces. Suggest adding one sentence disclosing the stop, aligning with the side-effect disclosure used for API-key delete.

frontend/src/app/workforces/workforces-routes.test.tsx
```diff
@@ -601,6 +627,142 @@ describe("workforce route entry points", () => {
...
+    await waitFor(() => {
+      expect(unarchiveWorkforceMock).toHaveBeenCalledWith(42)
+    })
+    // Reloads the list after a successful unarchive, same as archive/publish.
+    await waitFor(() => {
+      expect(listWorkforcesMock).toHaveBeenCalledTimes(2)
+    })
+  })
+
+  it("hides Archive/Publish/Unpublish on an archived card's menu", async () => {
```
rogercloud, line 740: The archive/publish/unpublish mocks remain declared, wired, and never asserted on click; the success toast is never asserted; the new tests only assert button presence. Add click-path tests with toast assertions, and keep `isLoading` on the ConfirmDialog mock so the double-click protection stays testable.

rogercloud (approved):
# Re-review: `feat(workforce): support permanent delete and unarchive` (round 5)

## PR Summary

This PR adds two lifecycle capabilities for workforces: (1) `DELETE /api/workforces/{id}?permanent=true` — hard-deletes a workforce in any status, cascading workers, run history, builder messages and triggers; cancels in-flight runs in-transaction with PAUSE dispatched post-commit (mirroring archive); manually cleans up workforce-scoped API keys and the FK-less `Deployment` row; deletes the auto-generated manager agent only when exclusively owned (shared/foreign manager kept — deliberately unlike discard, which blocks); post-commit, provider-side trigger bindings (e.g. Gmail watches) are unregistered via `asyncio.to_thread` with per-trigger sessions and per-item failure isolation. (2) `POST /api/workforces/{id}/unarchive` — restores an archived workforce to `draft`, so re-activation still passes through the re-validating publish step. Task rows are hidden (`is_visible=False`), not deleted, per the project-wide never-hard-delete-a-Task convention. Frontend adds a per-card three-dot menu (publish/unpublish/archive/unarchive/delete with confirm dialog) mirroring the Agent card pattern.

## Update Summary

Round 5 is one commit, `f8d278b9` "fix(workforce): address PR #1277 review round 4 findings" (6 files, +267/−84). It merges the per-action busy flags into a single `busyItemId` slot bound to every menu item including delete, consolidates the four non-delete handlers into `runMenuAction`, adds three per-status menu tests and four API-client tests, offloads the load+delete behind `asyncio.to_thread` (fixing the event-loop-blocking half of R3-F7), wraps the teardown dispatch in a route-level try/except (N2), adds unarchive commit rollback with `workforce_unarchive_failed` (N4), reassigns the fence return value (N13), adds the `permanent` Query description (N12), and renames to `is_workforce_manager_removal_safe`/`_ensure_workforce_lifecycle_access` (N6). Three comments were added or rewritten: the expunge comment, the documented trigger-create race (N3's chosen option), and — the one problem — the `passive_deletes` rejection rationale, which restates a factually incorrect claim (see R3-F7).

The author's PR-level reply (Aug 13) explains the three items left as-is. Two of those explanations hold: documenting the trigger-create race instead of widening the fence onto shared trigger CRUD (N3 — accepted, matches the finding's option 2), and deferring the discard/delete envelope extraction as a risk-managed follow-up (N7 — reasonable, remains a minor). The third — declining `passive_deletes=True` — repeats the false premise that this project's SQLite path never enables FK enforcement. The behavioral half of that finding is resolved; the remaining issue is the incorrect premise itself (see R3-F7 below).

## Prior-Findings Checklist

| ID | Status | Evidence (one line) |
|---|---|---|
| R1-2, N2, N3, N4, N5, N6, N8, N9, N11, N12, N13, S-prior-2 | FIXED | `runMenuAction` covers all four non-delete actions; teardown dispatch wrapped in route-level try/except; race documented (author chose option 2, reasoning sound); unarchive try/rollback + `workforce_unarchive_failed`; comment corrected (only "never hard-delete a Task" is shared; `stage_delete_agent` detaches, never hides); renames complete with zero residue; 4 API-client tests in `workforces-api.test.ts`; `busyItemId` single slot with all menu items disabled; stale test-name anchor removed; `Query` description added; fence return reassigned; handlers shrunk ~-16..-24 lines. |
| R3-F13, R3-F1, R3-F3, R3-F5, R3-F16, R2-2 | FIXED (regression-checked) | Controlled Popover kept through the busyItemId refactor; Task hiding still in-transaction before delete; expunge comment still accurate (verified against SQLAlchemy 2.0.x source); `reason="delete"` still passed; `operation_id="archive_workforce"` still pinned; expunge + per-thread session + config-only protocol contract intact. |
| R3-F7 | PARTIAL → **minor** (was major) | Event-loop stall fixed (`workforces.py:835-838` offloads load+delete to a worker thread) — the blocking sub-claim is resolved. Residual N+1 cascade (four relationships still `cascade="all, delete-orphan"` without `passive_deletes`; runs loaded twice) is a bounded-scale latency cost the author documented per the finding's option (b); on the production Postgres target row-level locking isolates other requests. The remaining defect is factual, not behavioral: the round-5 rejection comment — echoed in the author's Aug-13 reply — claims this repo never enables SQLite FK enforcement. That is false: `PRAGMA foreign_keys=ON` runs in `src/xagent/db/sqlite.py:100` (wired via `apply_sqlite_concurrency_pragmas` from `web/models/database.py:190`; `web/models/task.py:117-118` calls FK-on "this repo's default"). All child FKs are `ON DELETE CASCADE` (`workforce.py:81/127/177`, `trigger.py:93/161`), and `AgentApiKey.workforce_id` too (`agent_api_key.py:38`) — so the "AgentApiKey needs an explicit delete" sub-rationale is also wrong; only `Deployment` genuinely has no FK. **This comment must be corrected** (it teaches future maintainers a false fact), and `passive_deletes=True` is a safe, optional follow-up that would remove the N+1. |
| R3-F10 | PARTIAL → **minor** (was major) | unarchive/delete click assertions exist; the archive/publish/unpublish mocks are one-line `runMenuAction` wrappers whose generic path is already exercised by the unarchive click test, and the API layer has 4 client tests. Remaining gaps (click assertions for the three wrappers, success-toast assertion) are low-risk coverage completeness. |
| N10 | PARTIAL (minor) | Negative per-status assertions added (sub-item c fixed); openMenuId machinery and ConfirmDialog content still stripped by the mocks. |
| R3-F2, R3-F4, R3-F12, R3-F14, R3-nit, N7, S1, S2 | NOT FIXED | R3-F2: no mid-turn delete integration test. R3-F4: still `(AgentTrigger, type, config)` tuple handoff (latent). R3-F12: all five gaps remain — used_as_worker branch, delete/unarchive 404, TriggerRun/TriggerAudit post-cascade state, concurrent-fence test — plus C5 owner_mismatch untested on delete; C6 premise false (bare-DELETE archive already covered). R3-F14: no archive entry in builder. R3-nit: stray blank line at `workforce-builder.tsx:867`. N7: ~26-line transactional envelope still duplicated (author explicitly deferred). S1: dead guard `triggers.py:829-830`. S2: redundant `list(pool.map(...))` + wrong-rationale comment at `triggers.py:852-855`. |
| R2-1 / N14 | WAIVED / out of scope (maintained) | R2-1 waiver endorsed; N14 (agent delete leaks provider bindings) pre-existing. |

## New Findings

### Minor

**B4/C4 — `workforces.py:847-852` + frontend error parsing: object-form 500 `detail.message` never reaches the UI.** `parseApiError`/`getApiErrorMessage` surface the raw JSON (`parsed.text` branch) on the delete/unarchive paths, and the new `toThrow('Failed to delete workforce')` test passes via raw-JSON substring match — it does not prove extraction. Suggest `formatApiDetail` read `detail.message`, or drop the envelope's message and assert the actually surfaced copy.

**B5 — `workforces.py:918-922`: the unarchive route inlines a byte-identical 404/403 pair** instead of calling `_ensure_workforce_lifecycle_access` (`workforce_lifecycle.py:60-70`), which has no archived→409 guard and is directly reusable (or promote it to public). The route comment only explains why `ensure_workforce_access` is not used.

**B7 — `workforces/page.tsx:279-358` vs `build/page.tsx:558-630`: the three-dot menu duplicates the Agent card menu** (18 byte-identical lines, ~70-80 structurally mirrored). Extract a shared `CardActionsMenu` component (parameterized items / open control / `stopPropagation`).

**B8 — `page.tsx:63, 389-420`: `busyItemId` gates only the menu trigger and items** — the card's Run/Deploy/Edit stay clickable during an in-flight action, and the backend rejects with an avoidable, confusing error toast (run path rejects via the `WorkforceRunError` 400/409 mapping). Note Run/Edit are `asChild` Links, so gating needs `aria-disabled`/`pointer-events`.

**C2 — `page.tsx:144`: pagination fallback reads the render-closure `items`.** When concurrent loads resolve out of order (a menu action's `void load` racing a search change), `items` can belong to a different page/search and the error branch can reload an out-of-range page — the exact empty-state bug this PR fixes. Window is narrow; a stale-response guard (request sequence number) is more robust than `itemsRef` alone.

**D2 — `en.ts:3972`: confirm copy does not disclose that in-progress runs will be stopped**, and delete renders unconditionally on active cards. Add one sentence, matching the API-key delete side-effect disclosure convention.

**T1 — `workforces-routes.test.tsx`: publish/unpublish/archive have no click-path tests and delete-failure toast is untested.** The three mocks are declared, wired, never asserted; success toast never asserted; the ConfirmDialog mock strips `isLoading`, making the double-click guard untestable. Overlaps R3-F10/N10; reported as a standalone minor.

### Nit

**Q1 — `workforces.py:944`: `cast(Any, workforce).status`** — consistent with the three pre-existing routes; optional small helper if revisited.

**Q2 — `agent_management.py:445`: local `manager_discard_safe`** still uses the pre-rename name; one-line rename for consistency.

## Simplification Opportunities

- `page.tsx:127-134` — `shrink`: `runMenuAction`'s `() => async ()` currying is invoked immediately at every call site (`runMenuAction(...)()`); make it a plain async function called directly from `onClick`, -8 lines. (If the curried form is kept, call sites must keep the trailing `()` — without it `onClick` receives a function React ignores and the action never fires.)
- `triggers.py:829` (S1, round-4 leftover) — `delete`: dead guard `if not teardowns: return`; the sole caller already checks truthiness.
- `triggers.py:855` (S2, round-4 leftover) — `shrink`: drop the redundant `list(pool.map(...))` (the `with` block's `shutdown(wait=True)` already joins and `_teardown_one` swallows exceptions), and fix the wrong-rationale comment at 852-854.

net: ~-12 lines possible

## Blocking Status & Recommended Decision

- **Blocking: no.** The two round-4 blocking items are both resolved to minor: R3-F7's event-loop-stall half is fixed and the residual N+1 is a documented bounded-scale cost; what remains is the factually incorrect premise in the comment (and the author's reply) — a must-fix documentation correction, not a behavioral defect. R3-F10's remaining gaps are thin-wrapper coverage completeness.
- **Recommended event: APPROVE** (author: yiboyasss), with the minor findings above as non-blocking follow-ups. The one item I would insist on is correcting the false FK-enforcement claim in `workforce_lifecycle.py:335-350` (or landing `passive_deletes=True`), ideally in this PR.

Issue #1257, opened 2026-08-10
Run the full regular frontend test suite in CI


AlexLiu190625:
## Problem

Frontend CI uses several targeted Vitest lanes. After #1243, those required selectors cover 54 of the 128 regular frontend test files; 74 files are only exercised by the full `test:run` command.

Targeted lanes provide useful ownership and focused failure reporting, but a test added outside their directories can still be omitted from required CI unless another lane is updated.

## Current evidence

- `npm run test:run`: 128 files / 1,968 tests pass.
- Required selector union: 54 files.
- Files outside the required selector union: 74.
- Widget coverage remains a separate configuration with per-file thresholds and should not be removed.

## Scope

- Decide whether the regular frontend gate should run `npm run test:run` directly or use an equivalent automatically discovered full-suite lane.
- Preserve the widget coverage configuration and its thresholds.
- Avoid running the same regular tests redundantly unless the duplication has an explicit ownership or reporting benefit.
- Update CI summary and test-manifest contracts for the selected design.
- Measure runner time and failure-reporting impact before replacing existing targeted lanes.

## Acceptance criteria

- Every regular frontend test discovered by `vitest.config.ts` is exercised by required CI without maintaining a per-file allowlist.
- Adding a new `frontend/src/**/*.test.ts` or `frontend/src/**/*.test.tsx` file requires no workflow selector change.
- Widget coverage thresholds remain required.
- CI failure propagation remains unmasked through `frontend-build` and `ci-summary`.
- The final lane structure and any intentional duplicate execution are documented in the PR.

## Context

Follow-up to #1243. This is broader than #1090, whose acceptance criteria are limited to automatic App Router test discovery.

AlexLiu190625:
Track the Widget coverage membership policy as part of this follow-up. PR #1243 preserves the targeted Widget coverage lane and its per-file thresholds, but that lane intentionally uses an explicit test include list. The final design here should state whether Widget tests are classified into that lane automatically or whether the allowlist remains explicit with a consistency check. In either case, every regular Vitest test must still enter required CI automatically.

AlexLiu190625:
## Design decision

Replace the four required targeted regular-test lanes:

- `npm run test:pages`
- `npm run test:kb-components`
- `npm run test:app-pages`
- `npm run test:home-build-contracts`

with one required full-suite lane:

- `npm run test:run`

The final required frontend test structure should be exactly:

1. `npm run test:widget:coverage`
[...]

PR #1279, opened 2026-08-11
ci: run the full frontend test suite

AlexLiu190625:
## Summary

- replace the four targeted regular frontend CI lanes with the selection-neutral `npm run test:run` suite
- keep Widget coverage and the independent CI manifest as required frontend steps
- centralize Widget test membership, coverage owners, instrumentation settings, and per-file thresholds in one policy consumed by the real Vitest config
- fail closed when regular discovery, execution selectors, required workflow steps, or CI failure propagation drift

## Why

The required frontend workflow previously ran only a selector union. A new regular Vitest file could be committed without entering required CI. The full base suite is now the required regular-test owner, so tests matching `src/**/*.test.ts` or `src/**/*.test.tsx` are discovered automatically.

Widget tests intentionally run both in the regular suite for correctness and in the coverage lane for per-file thresholds. The CI manifest intentionally runs as a focused launcher and through the full suite so that either launcher drifting fails closed.

## Verification

- `npm --prefix frontend run test:ci-manifest` — 166 tests passed
- `npm --prefix frontend run test:run` — 130 files / 2,099 tests passed
- `npm --prefix frontend run test:widget:coverage` — 31 files / 746 tests passed with thresholds
- frontend type-check and lint passed
- semantic actionlint and changed-file pre-commit passed
- clean `npm ci --ignore-scripts` and `git diff --check` passed
- GitHub Actions — 14/14 checks passed
- `frontend-build` — 4m34s, compared with 4m37s on the current-main baseline run

## Merge gate

- [ ] A repository administrator must require the `CI Summary` check from GitHub Actions (app ID `15368`) on `main` before merge and issue closure. The existing merge-queue ruleset must remain unchanged.

Refs #1257


frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
   }
 
-  if (frontendBuild.steps.some((step) => step.run === "npm run test:home-build-pages")) {
-    throw new Error("legacy test:home-build-pages must not run in frontend-build")
+  const retiredDirectLauncher = frontendBuild.steps
+    .map((step) => {
+      requireRecord(step, "jobs.frontend-build.steps entry")
+      return typeof step.run === "string" ? step.run.trim() : undefined
+    })
+    .find((run) => run !== undefined && retiredFrontendLaunchers.has(run))
```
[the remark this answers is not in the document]
AlexLiu190625: This guard is intentionally bounded to an entire semantic run value after trimming; it is not a shell-command analyzer. Splitting the scalar into lines would reject non-executable occurrences in heredoc data, inactive branches, and multiline quoted data, while still missing executed forms such as line continuations, semicolon chains, and shell wrappers. The required CI invariant is enforced separately: the Widget coverage, manifest, and full test:run steps each exist exactly once and cannot be masked. Keeping the whole-scalar check preserves that bounded contract without introducing an incomplete shell parser.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
     }
   }
 
-  if (frontendBuild.steps.some((step) => step.run === "npm run test:home-build-pages")) {
-    throw new Error("legacy test:home-build-pages must not run in frontend-build")
+  const retiredDirectLauncher = frontendBuild.steps
```
rogercloud, line 422: **minor** — Following up on the earlier multiline-command thread: exact set membership can still be bypassed by compound or multiline commands such as `cd frontend && npm run test:pages`. The separate required-step checks mean this cannot mask the three required lanes, so the remaining impact is CI hygiene rather than coverage loss. Please normalize shell blocks or reject retired launcher invocations within compound commands.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
+    defineConfigCall.arguments.length !== 1
+  ) {
+    throw new Error(ownerError)
+  }
+  const configObject = defineConfigCall.arguments[0]!
+  if (!ts.isObjectLiteralExpression(configObject)) {
+    throw new Error(ownerError)
+  }
+
+  const testProperties = configObject.properties.filter((property) => {
```
rogercloud, line 704: **minor** — This filter ignores spread assignments, and the AST walk never requires `...baseConfig` to be present. Removing that spread from `vitest.widget.config.ts` passes this guard and then fails loudly because the `@` alias is lost. Please require the leading base-config spread explicitly, or replace the AST walk with a strict canonical-source check for this fixed config.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
+    expect(() => assertWidgetConfigSourceConsumesPolicy(widgetConfigSource)).not.toThrow()
+  })
+
+  it("keeps Widget policy paths, patterns, and thresholds independently valid", () => {
+    assertWidgetTestFilePolicy(widgetTestFiles)
+    assertWidgetCoverageOwnerPolicy(widgetCoverageOwners)
+    assertWidgetCoveragePolicy(buildWidgetTestOptions(vitestConfig.test).coverage)
+  })
+
+  it("makes the real Widget config consume the complete policy builder", () => {
```
rogercloud, line 862: **minor** — This compares `buildWidgetTestOptions(vitestConfig.test)` with a config produced by the same function from the same module object, so it is a self-comparison rather than an independent contract check. Please remove it or compare the resolved config directly with independently exported policy values.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
+  }
+  if (JSON.stringify(config.test.coverage) !== JSON.stringify(expected.coverage)) {
+    throw new Error("Widget config must consume policy coverage")
+  }
+  assertWidgetCoveragePolicy(config.test.coverage)
+}
+
 describe("frontend CI test manifest", () => {
+  it("keeps the current Widget coverage contract available to the real config", () => {
+    const expectedWidgetTestFiles = [
```
rogercloud, line 764: **minor** — The 31 paths here, together with the 27 threshold groups below, re-copy the policy imported by this test. That catches an accidental one-sided edit but not deliberate weakening when both copies are changed. Please make the intended contract an explicitly reviewed oracle (for example, a generated snapshot or documented intent fixture); once that protection is preserved, compare the real config with the exported policy instead of retaining this handwritten duplicate.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
     const source = replaceExactlyOnce(
       workflowSource,
-      "      - name: Run App Router tests\n        working-directory: ./frontend\n        run: npm run test:app-pages\n",
-      "      - run: npm run test:app-pages\n        env:\n          APP_ROUTER_TEST_MODE: manifest\n        working-directory: ./frontend\n        name: Run App Router tests\n",
-      "App Router step presentation drift",
+      "      - name: Run full frontend test suite\n        working-directory: ./frontend\n        shell: bash\n        run: npm run test:run\n",
+      "      - run: npm run test:run\n        env:\n          FULL_SUITE_MODE: manifest\n        working-directory: ./frontend\n        shell: bash\n        name: Run full frontend test suite\n",
+      "full regular suite presentation drift",
     )
     const followingJobs = source.slice(source.indexOf("\n  ci-summary:\n"))
```
rogercloud, line 1130: **minor, latent** — If the `ci-summary` marker is absent, `indexOf` returns `-1` and `slice(-1)` silently returns the final character, allowing the following containment check to become vacuous. Assert `markerIndex >= 0` before slicing so marker loss fails at its source.

frontend/vitest.widget.policy.ts
```diff
@@ -0,0 +1,188 @@
...
+  )
+
+  return {
+    ...baseTest,
+    include: [...widgetTestFiles],
+    coverage: {
+      provider: "v8",
+      all: true,
+      include: coverageOwners,
+      exclude: [],
```
rogercloud, line 173: **minor, informational** — `exclude: []` removes Vitest's default exclusions, while the explicit extension list narrows its defaults. Exact owner paths make this harmless today, but future directory globs would start considering `dist`, declaration, and test files while ignoring other normally supported extensions. Please remove these overrides to preserve base semantics, or document the deviation as intentional.

frontend/vitest.widget.policy.ts
```diff
@@ -0,0 +1,188 @@
...
+]
+
+export const widgetCoverageExtensions = [".js", ".ts", ".tsx"]
+
+export function buildWidgetTestOptions(baseTest: InlineConfig | undefined): InlineConfig {
+  const coverageOwners = widgetCoverageOwners.map(
+    (owner) => owner.coveragePattern ?? owner.sourcePath,
+  )
+
+  return {
```
rogercloud, line 166: **minor, latent** — This shallow spread replaces `baseTest.coverage` wholesale; the previous widget config used Vitest's deep merge. A future coverage option added to `vitest.config.ts` would therefore disappear silently in the widget lane. Please use `mergeConfig` or explicitly merge the nested coverage object before applying the widget overrides.

frontend/vitest.widget.policy.ts
```diff
@@ -0,0 +1,188 @@
...
+  },
+  {
+    sourcePath: "src/lib/auth-cache.ts",
+    thresholds: { statements: 90, branches: 80, functions: 90, lines: 90 },
+  },
+  {
+    sourcePath: "src/lib/files-disabled-presentation.ts",
+    thresholds: { statements: 85, branches: 80, functions: 90, lines: 85 },
+  },
+  {
```
rogercloud, line 137: **informational, non-blocking** — This 100% coverage owner has no dedicated test file and is covered collaterally by unrelated suites. `all: true` makes the new setup safer than the base because an unloaded owner now fails loudly, but a focused test would make the 100% contract intentional and maintainable; otherwise consider a floor that reflects collateral coverage.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
+    Boolean(testConfig.standalone) ||
+    Boolean(config.allowOnly) ||
+    Boolean(testConfig.allowOnly)
+  ) {
+    throw new Error("regular execution must be selection-neutral")
+  }
+}
+
+const coverageMetrics = ["statements", "branches", "functions", "lines"] as const
+const expectedWidgetCoverageExtensions = [".js", ".ts", ".tsx"]
```
rogercloud, line 499: **simplification — delete:** `expectedWidgetCoverageExtensions` duplicates the imported `widgetCoverageExtensions`, and the equality assertion compares the policy with itself. Use the imported constant at the remaining call sites and remove the tautological assertion.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
+      perFile: true,
+      ...Object.fromEntries(owners.map((owner) => [
+        owner.coveragePattern ?? owner.sourcePath,
+        owner.thresholds,
+      ])),
+    },
   }
 }
 
+function assertWidgetConfigSourceConsumesPolicy(source: string) {
```
rogercloud, line 673: **simplification — native:** This TypeScript AST walk is about 75 lines for an eight-line, fully known config and still misses removal of `...baseConfig`. A newline-normalized canonical-source equality check is roughly 5–10 lines and is stricter for this fixed file.

frontend/src/ci/frontend-test-manifest.test.ts
```diff
@@ -403,39 +419,725 @@ function assertSemanticWorkflowManifest(source: string) {
...
+      "src/contexts/app-context-chat.test.tsx",
+      "src/contexts/auth-context.test.tsx",
+      "src/contexts/file-access-context.test.tsx",
+      "src/hooks/use-file-mention.test.tsx",
+      "src/hooks/use-websocket.test.ts",
+      "src/lib/api-wrapper.test.ts",
+      "src/lib/auth-cache.test.ts",
+      "src/lib/files-disabled-presentation.test.ts",
+    ]
+    const expectedWidgetCoverageThresholds = {
```
rogercloud, line 797: **simplification — shrink:** This 27-group object duplicates `widgetCoverageOwners`. Derive the expected map with `Object.fromEntries(widgetCoverageOwners.map(owner => [owner.coveragePattern ?? owner.sourcePath, owner.thresholds]))` and compare it with the real in-scope `thresholds` object after separating `perFile`, rather than comparing two policy-derived temporaries.

frontend/vitest.widget.policy.ts
```diff
@@ -0,0 +1,188 @@
+import type { InlineConfig } from "vitest"
+
+export type CoverageMetric = "statements" | "branches" | "functions" | "lines"
```
rogercloud, line 3: **simplification — shrink:** The policy declares the metric union while the test separately declares the same four strings at runtime. Export `coverageMetrics` as a const tuple here, derive `CoverageMetric` from `typeof coverageMetrics[number]`, and import the tuple in the test.

rogercloud (changes requested):
## Summary

This PR consolidates four retired frontend CI lanes into three canonical commands, moves widget test selection and coverage ownership into a shared policy module, and adds a manifest guard intended to keep that topology stable. The product-side consolidation is correct and substantially simpler than the previous lane layout, while the new guard is responsible for preventing future workflow and coverage drift. Verification completed cleanly (`tsc`, ESLint, 130 files / 2,099 tests), and a targeted probe confirmed that the per-file widget thresholds are genuinely enforced; however, two material holes remain in the guard contract.

## Approach verdict

**acceptable-with-reservations.** The CI consolidation and extracted widget policy are the right direction, but the guard test is substantially over-provisioned and partly self-referential. In particular, it does not enforce Issue #1257's “exactly three” requirement, does not pin the coverage-bearing package script, duplicates much of the policy it claims to centralize, and hard-codes a CI-summary sequence that conflicts with the documented extension path.

## Major findings

1. **[major] `frontend/package.json:19` — pin the coverage-bearing script body.** The manifest pins `test:ci-manifest` and `test:run`, but not `test:widget:coverage`. Changing this script to omit `--coverage` leaves the guard green while silently disabling all 27 per-file thresholds. Add an exact package-script assertion, for example:
   ```ts
   expect(packageJson.scripts["test:widget:coverage"]).toBe(
     "vitest run --config vitest.widget.config.ts --coverage",
   )
   ```

2. **[major] `frontend/src/ci/frontend-test-manifest.test.ts:396` — enforce exactly three frontend test steps, not merely the presence of three commands.** The current assertions prove that each required command appears once, but a fourth test step such as `npm run test:pages-v2` still passes every guard. Collect all frontend test-launching `run` steps and assert that their normalized command list, including its length, equals the three allowed commands.

## Minor findings

1. **[minor] `frontend/src/ci/frontend-test-manifest.test.ts:422` — retired-lane detection only matches exact command strings.** As discussed in the earlier thread, compound or multiline commands such as `cd frontend && npm run test:pages` bypass the set lookup. The separate required-command checks prevent this from masking coverage, so the impact is CI hygiene rather than coverage loss; normalize shell blocks or reject retired launcher invocations within compound commands.

2. **[minor] `frontend/src/ci/frontend-test-manifest.test.ts:704` — the AST guard never requires `...baseConfig`.** Spread assignments are filtered out, so deleting the spread from `vitest.widget.config.ts` passes the guard and then fails loudly at import resolution because the `@` alias is lost. Require the leading base-config spread explicitly, or replace this machinery with a strict canonical-source check for the fixed eight-line config.

3. **[minor] `frontend/src/ci/frontend-test-manifest.test.ts:862` — the resolved-config equality is tautological.** Both sides are produced by `buildWidgetTestOptions(vitestConfig.test)` from the same module object, so this assertion provides no independent signal. Remove it or compare the resolved config against independently exported policy values rather than calling the same builder again.

4. **[minor] `frontend/src/ci/frontend-test-manifest.test.ts:764` — the test re-copies the policy it is meant to centralize.** The 31 test paths and 27 threshold groups duplicate the imported policy byte-for-byte. This catches accidental one-sided edits but not deliberate weakening when both copies are changed; make the intended contract an explicitly reviewed oracle (for example, a generated snapshot or documented intent fixture), then derive the mechanical config comparison from the policy rather than maintaining two handwritten copies.

5. **[minor] `frontend/src/ci/frontend-test-manifest.test.ts:303` — the CI-summary guard conflicts with the documented way to add a job.** `docs/branch-protection.md:217-219` says to append the new job to `check_job`, but the guard requires an exact ordered 18-command sequence, so following the documentation produces an unrelated manifest failure. Prefer validating each failure-propagation entry structurally while allowing additional jobs; at minimum, document that this hard-coded expectation must be updated as part of the procedure.

6. **[minor, latent] `frontend/src/ci/frontend-test-manifest.test.ts:1130` — `slice(indexOf(...))` succeeds vacuously when the marker is absent.** `indexOf` returns `-1`, and `slice(-1)` leaves a one-character string on which the following containment assertion can pass trivially. Assert the marker index before slicing:
   ```ts
   const markerIndex = source.indexOf("\n  ci-summary:\n")
   expect(markerIndex).toBeGreaterThanOrEqual(0)
   const ciSummarySource = source.slice(markerIndex)
   ```

7. **[minor, informational] `frontend/vitest.widget.policy.ts:173` — the extracted policy changes future coverage discovery semantics.** `exclude: []` removes Vitest's default exclusions and `extension` narrows the default extension set; this is harmless with today's exact owner paths but becomes observable if the policy later accepts directory globs. Remove these overrides to preserve the base behavior, or document that the narrower extension set and inclusion of `dist`, declaration, and test files are intentional.

8. **[minor, latent] `frontend/vitest.widget.policy.ts:166` — the shallow spread drops future base coverage settings.** The literal `coverage` property replaces `baseTest.coverage` wholesale, whereas the previous config used Vitest's deep merge. Use `mergeConfig` (or explicitly merge the nested coverage object), for example:
   ```ts
   return mergeConfig(baseTest ?? {}, {
     include: widgetTestFiles,
     coverage: widgetCoverageOptions,
   })
   ```

9. **[minor, documentation] `frontend/package.json:13` — the retired scripts' dev-only status is not recorded in the repository.** The four scripts remain useful locally but no longer define CI lanes, and that distinction currently exists only in Issue #1257. Add a short frontend testing note naming the three canonical CI commands and labeling these scripts as local developer conveniences.

10. **[minor] `.github/workflows/ci.yml:640` — the manifest does not pin the `frontend-build` job condition.** The condition is consistent with all seven gathered jobs and is fail-closed through `ci-summary`, so there is no current correctness bug. Add a direct assertion for `frontendBuild.if` for symmetry with the pinned `ci-summary` condition and to prevent accidental unilateral drift.

11. **[minor, advisory] `.github/workflows/ci.yml:644` — the expanded frontend job has no timeout cap.** It now runs 130 files / 2,099 tests, reruns the 31-file widget lane, and executes the manifest both explicitly and through `test:run`. Add a job-level `timeout-minutes` value (30 minutes would match the e2e lane) and optionally pin the timeout in the manifest.

12. **[informational, non-blocking] `frontend/vitest.widget.policy.ts:137` — a 100% owner has no dedicated test file.** `presentation-capabilities.tsx` is covered collaterally by unrelated tests; `all: true` makes the new setup safer than the base because unloaded owners now fail loudly. Prefer a focused test for this owner, or use a floor that reflects intentional collateral coverage.

## Merge prerequisite

The PR checklist still shows the repository-administrator step for requiring **CI Summary** on `main` as unchecked. Confirm that branch protection requires this context before merge; otherwise the consolidated gate remains advisory regardless of the workflow changes.

L764: shrink The 31-entry `expectedWidgetTestFiles` array duplicates the imported `widgetTestFiles`. Compare the resolved config with the exported policy value after preserving the one-sided-edit canary in an explicitly reviewed snapshot or intent fixture.

L797: shrink The 27-group `expectedWidgetCoverageThresholds` object duplicates `widgetCoverageOwners`. Derive the expected map from `widgetCoverageOwners` and compare it with the real in-scope `thresholds` object after separating `perFile`.

L499: delete `expectedWidgetCoverageExtensions` duplicates the imported `widgetCoverageExtensions`, and their equality assertion compares the policy with itself. Use the imported constant at the remaining call sites and remove the self-comparison.

L673: native The full TypeScript AST walk validates an eight-line, fully known config and still misses removal of `...baseConfig`. Replace it with newline-normalized canonical-source equality, which is shorter and stricter for this fixed file.

L3: shrink `CoverageMetric` and the test's runtime metric tuple encode the same four strings separately. Export a runtime `coverageMetrics` tuple from the policy, derive `CoverageMetric` from `typeof coverageMetrics[number]`, and import the tuple in the test.

net: -128 lines possible

## Verification

Verified at PR head `6135d6aa`: TypeScript and ESLint are clean; the frontend run passes 130 files / 2,099 tests; and a targeted probe confirms that widget per-file thresholds are enforced.

PR #1179, opened 2026-08-06
feat: add built-in Chrome (keyless) MCP connector

yiboyasss:
## Summary

Adds a built-in **Chrome** connector (catalog id `chrome-devtools`) to the MCP
catalog, backed by Google's
[chrome-devtools-mcp](https://github.com/ChromeDevTools/chrome-devtools-mcp)
(pinned to `1.6.0`, launched `--headless --isolated`). Agents get 29 browser
tools (navigate, snapshot, screenshot, fill, network/console inspection,
Lighthouse audits, ...) running against a headless Chrome on the backend host.

> **Rollout note:** the connector ships **hidden** (`is_visible_in_connector=False`),
> and hiding is enforced server-side: both catalog connect endpoints 404 hidden
> apps before provisioning anything, pinned by endpoint-driven tests on the
> real app id (api_key/keyless path) and on the mcp_oauth path. The runtime
> currently spawns a fresh stdio process per tool call, so browser state does
> not survive across calls and multi-step flows break — see #1200. Once that
> lands (plus the security review tracked with it), admins re-enable visibility
> via `PATCH /api/admin/mcp/apps`. The builtin_oauth redirect flow does not
> run the hide-gate (#1203) — do not reuse the hidden-rollout idiom on an
> oauth-transport row until that closes.

Chrome is the first catalog app that needs no secrets, so this PR also
introduces a **"keyless"** auth class end to end:

- `classify_app_auth`: a stdio command without `required_env` (and without
  `env_mapping` — that shape expects an injected OAuth-style token) now
  classifies as `"keyless"`; stdio-only, so a command on a remote transport
  still classifies unconnectable.
- `POST /api/mcp/apps/{id}/connect` accepts keyless apps alongside `api_key`;
  the association is created with no per-user env, any env a caller sends is
  dropped, and keyless connects send `is_active: true` so reconnecting a
  dormant association reactivates it (pinned in both directions by tests).
- Admin write validation updated accordingly; partial shapes still 422.
- Connect dialog connects keyless apps in one click via a shared
  `connectCatalogApp` helper; in-flight state is tracked per app id
  (`Set<string>`), the settings dialog shows a pending/disabled Connect state,
  and the key-edit button is hidden for keyless apps.

## Declared behavior changes to pre-existing paths

1. The api_key connect success toast changed from "Save" to
   "Connected to {name}" (shared helper side effect, kept deliberately;
   toast key pinned by a test).
2. Hidden catalog apps now 404 on the connect endpoints (previously only
[...]

Issue #928, opened 2026-07-20
feat: Allow enabling Triggers during agent creation


yiboyasss:
**`Deadline: July 23, 2026`**

### Description

Allow users to enable and configure Triggers while initially creating an Agent.

Currently, Triggers are disabled during the initial agent configuration phase and can only be enabled after the Agent has been created.

PR #1280, opened 2026-08-11
fix: resolve mcp: tool selectors against the real connected server row

yiboyasss:
## Summary

Task/agent creation builds `mcp:<selector>` category strings for each
connected MCP connector the user selects. Both call sites in
`agent-builder.tsx` guessed at that selector instead of resolving it against
the actually-connected `MCPServer` row.

The backend has **two contradictory conventions** for what that row is
actually named:

- `_ensure_catalog_app_server` / `_ensure_catalog_mcp_oauth_server`
  (`src/xagent/web/api/mcp.py`) name the row after the catalog **app_id**.
- `_ensure_user_mcp_server` (`src/xagent/web/api/auth.py`), used for
  `builtin_oauth` apps, names the row after the display **name**.

`normalize_mcp_server_name`'s hyphen/space/case folding is deterministic, so
every previously-working connector's app_id and display name happened to
fold to the same key (Zoom → `zoom`, Google Maps → `google-maps`) — until
**Chrome** (#1179) diverged by a whole word: `app_id="chrome-devtools"`,
name `"Chrome"`. A name-first guess builds `mcp:Chrome`, matching no server
row, silently loading zero tools
(`mcp_load_summary: reason=no_tools_returned`) — the agent falls back to a
text-only answer with no visible error, even though connecting the app
looked successful.

## Fix

Added `resolveMcpToolSelector(server, mcpServers, officialApps)` to
`frontend/src/lib/mcp-lookup.ts`. Instead of guessing a fallback, it:

1. Finds the catalog app the current selection refers to.
2. Looks up a real, connected `MCPServer` row by that app's `id`, then by
   its `name` — covering both backend conventions regardless of which one
   applies to a given app.
3. Returns that row's authoritative real name.
4. Falls through to returning the incoming selector **completely
   unchanged** — never a guessed id or name — in two cases: no server row
   is connected at all, or a matched row's `name` is empty/whitespace. If
   the selector was already correct (e.g. the row just isn't in this
   viewer's current `mcpServers` snapshot yet), it stays correct; if the
   app genuinely isn't connected, there's nothing better to substitute.
   Logs a `console.warn` when this happens — dev-visible only; a
   user-facing signal is a separate, deferred design decision (see Review
   history).

Both call sites in `agent-builder.tsx` (the publish/create handler and the
builder-preview chat handler) call this shared helper instead of each
inlining its own resolution logic, dedup the resolved values with a `Set`
before building `tool_categories`, and both now push the `"knowledge"`
category when a knowledge base is selected (the preview-chat path was
missing it, diverging from what the agent would run with once actually
saved).

`selectedMcpServers` is also re-seeded from the resolved selectors
immediately after both save paths succeed (edit-mode update and initial
create), matching how `loadAgent` seeds it on a fresh load — otherwise it
keeps holding the pre-resolution value (e.g. a display name) while
`originalData.tool_categories` now holds the resolved real name, making the
editor's "unsaved changes" check compare the two and return true forever
after a successful save. The three other places that filtered/stripped
`mcp:` prefixes from `tool_categories` (`loadAgent`, template-requirements
loading, `isDirty`'s own extraction) now share the same
`mcpServerNamesFromToolCategories` helper the re-seed uses, so a future edit
to one can't silently drift out of sync with the others.

A connector's `connect-mcp-dialog.tsx` selection card is also fixed to
match `isSelected`'s own id-or-name check when toggling a card off — the
re-seed above can leave `selectedMcpServers` holding a connector's id
instead of its display name, and the toggle previously removed only by
name, so clicking a visually-selected id-only entry appended the name
instead of removing the id: the card stayed selected and duplicated no
matter how many more times it was clicked.

## Testing

- `frontend/src/lib/mcp-lookup.test.ts`: covers the app-id-named convention
  (Chrome), the app-id-carrying oauth-row convention and the isolated
  name-only-fallback convention (Facebook), disambiguating among several
  connected rows, an already-resolved-name round trip, the not-yet-connected
  fallback (preserves the input, never guesses), a matched-row-with-blank-name
  fallback, and the no-match/empty-string fallback.
- `frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx`
  (new): mounts the real `AgentBuilder` component end to end and asserts (a)
  the outbound PUT body's `tool_categories` carries the resolved server name,
  not the unresolved display name held in `selectedMcpServers`, and (b)
  `isDirty` (observed via the Update button re-disabling) returns to false
  once the save succeeds. Verified this test actually catches the regression
  by reverting the `selectedMcpServers` re-seed and confirming it fails.
- `frontend/src/components/mcp/connect-mcp-dialog.test.tsx` (new case):
  mounts `ConnectMcpDialog` in select mode with an id-only initial selection
  and asserts one click on the visually-selected card deselects it outright,
  rather than duplicating it. Verified this test fails against the
  pre-fix toggle logic.
- `npx vitest run src/lib/mcp-lookup.test.ts src/components/build/agent-builder src/components/mcp/connect-mcp-dialog.test.tsx` — 106 passed.
- `npx tsc --noEmit` — clean.

**Known gaps (non-blocking, left for a follow-up):** the create-mode POST
branch's resolution/dedup, the preview-chat path's resolution/dedup/knowledge
push, and the async-load race noted in `agent-builder.tsx` (where a preview
message sent before `mcpServers`/`officialApps` have loaded falls back to the
raw selector) have no dedicated automated coverage yet — the edit-mode save
path and the connect-dialog toggle are the two covered end to end.

## Review history

- **Round 1** caught that the initial version hard-coded an app_id-first
  fallback, which fixed Chrome but risked breaking any app using the other
  naming convention the other way, and flagged zero test coverage on the
  original diff (both addressed by `resolveMcpToolSelector`).
- **Round 2** caught two further issues in that fix: (1) the
  not-yet-connected fallback still guessed `connectedApp.id`, which is the
  wrong convention for `builtin_oauth` apps and could overwrite an
  already-correct selector — removed; it now preserves the input unchanged
  instead of guessing. (2) `selectedMcpServers` was never re-seeded after a
  save, leaving `isDirty` permanently true — fixed at both save call sites.
  Round 2 also corrected an inaccuracy in round 1's own reasoning: Facebook's
  real connected server row does carry `app_id` (`_enrich_oauth_server_info`
  enriches oauth-transport rows), so an id-first guess would not actually
  have broken it the way round 1's comments claimed — the doc comment and
  test fixtures now reflect the real API shape.
- **Round 3** caught that the round-2 re-seed fix is directly reachable from
  `connect-mcp-dialog.tsx` (not touched by rounds 1-2), where the
  selection-card toggle only removed entries by display name — fixed as
  described above. Also flagged that the actual behavioral fix (outbound
  payload + `isDirty` clearing) had no test beyond the pure-helper unit
  tests — addressed by the new `agent-builder-mcp-selector-resolution.test.tsx`.
  Additional round-3 minors: a matched-row-with-blank-name edge case in the
  doc comment/behavior, missing dedup at both call sites, the missing
  `"knowledge"` category in the preview path, inconsistent create-mode
  seeding, and consolidating the three inlined `mcp:`-stripping copies onto
  one helper — all fixed.
- **Round 4 (approved)**: no blocking findings. Remaining minors, left
  as follow-ups: broader test coverage for the create-mode/preview-path/
  async-race branches (see "Known gaps" above), and a user-facing (not just
  `console.warn`) signal when a selector can't be resolved at save time.

### DIFF
--- a/frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx
+++ b/frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx
@@ -2,10 +2,10 @@ import React from "react"
 import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 
-// #1280 round-3 review, Major finding 2: the pure resolveMcpToolSelector unit
-// tests cover the helper in isolation, but nothing exercised the two things
-// this PR's fix actually depends on end to end: (a) the outbound save
-// request's tool_categories carries the *resolved* selector, not whatever
+// The pure resolveMcpToolSelector unit tests (mcp-lookup.test.ts) cover the
+// helper in isolation, but nothing exercised the two things this fix
+// actually depends on end to end: (a) the outbound save request's
+// tool_categories carries the *resolved* selector, not whatever
 // selectedMcpServers holds, and (b) isDirty returns to false after a
 // successful save instead of staying permanently true because
 // selectedMcpServers was never re-seeded to match. This drives both through
@@ -229,7 +229,8 @@ describe("AgentBuilder mcp: selector resolution on save", () => {
     // The Update button starts disabled: nothing has changed yet, and
     // selectedMcpServers/originalData's MCP extraction agree on the loaded
     // (still-unresolved) "Chrome" -- loading an already-broken agent
-    // doesn't self-heal it without an edit, by design (round-1 review).
+    // doesn't self-heal it without an edit, by design (a plain save does
+    // not re-resolve a selector that was already persisted unresolved).
     const updateButton = screen.getByRole("button", {
       name: "builds.editor.header.update",
     })
--- a/frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx
+++ b/frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx
@@ -1,0 +1,260 @@
+import React from "react"
+import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+
+// #1280 round-3 review, Major finding 2: the pure resolveMcpToolSelector unit
+// tests cover the helper in isolation, but nothing exercised the two things
+// this PR's fix actually depends on end to end: (a) the outbound save
+// request's tool_categories carries the *resolved* selector, not whatever
+// selectedMcpServers holds, and (b) isDirty returns to false after a
+// successful save instead of staying permanently true because
+// selectedMcpServers was never re-seeded to match. This drives both through
+// the real AgentBuilder component and a mocked PUT /api/agents/{id}.
+
+const apiRequestMock = vi.hoisted(() => vi.fn())
+
+vi.mock("@/lib/api-wrapper", async () => {
+  const actual = await vi.importActual<typeof import("@/lib/api-wrapper")>(
+    "@/lib/api-wrapper"
+  )
+  return { ...actual, apiRequest: apiRequestMock }
+})
+
+vi.mock("@/lib/utils", async () => {
+  const actual = await vi.importActual<typeof import("@/lib/utils")>("@/lib/utils")
+  return {
+    ...actual,
+    getApiUrl: () => "http://api.local",
+    getUploadApiUrl: () => "http://api.local",
+    getWsUrl: () => "ws://api.local",
+  }
+})
+
+vi.mock("@/contexts/app-context-chat", () => ({
+  useApp: () => ({
+    state: {
+      messages: [],
+      traceEvents: [],
+      currentTask: null,
+      isProcessing: false,
+      isHistoryLoading: false,
+      taskId: null,
+      filePreview: { isOpen: false },
+      dagExecution: null,
+      steps: [],
+    },
+    setTaskId: vi.fn(),
+    sendMessage: vi.fn(),
+    dispatch: vi.fn(),
+    closeFilePreview: vi.fn(),
+    pauseTask: vi.fn(),
+    resumeTask: vi.fn(),
+    openFilePreview: vi.fn(),
+    requestStatus: vi.fn(),
+  }),
+}))
+
+vi.mock("@/contexts/auth-context", () => ({
+  useAuth: () => ({ token: "token", user: { id: "1", is_admin: false } }),
+}))
+
+vi.mock("@/contexts/i18n-context", () => ({
+  useI18n: () => ({
+    locale: "en",
+    t: (key: string, vars?: Record<string, string>) =>
+      vars?.appName ? `${key}:${vars.appName}` : key,
+  }),
+}))
+
+// The catalog app: display name "Chrome" deliberately diverging from its
+// real connected server row's name "chrome-devtools" (installApi below) --
+// the exact divergence this whole fix exists for.
+vi.mock("@/contexts/mcp-apps-context", () => ({
+  useMcpApps: () => ({
+    apps: [
+      {
+        id: "chrome-devtools",
+        name: "Chrome",
+        description: "",
+        icon: "",
+        users: "",
+        transport: "stdio",
+        provider: "",
+        category: "Productivity",
+        is_connected: true,
+      },
+    ],
+    getAppIcon: () => null,
+  }),
+}))
+
+vi.mock("@/lib/branding", () => ({
+  getBrandingFromEnv: () => ({ appName: "Xagent" }),
+}))
+
+vi.mock("sonner", () => ({ toast: { error: vi.fn(), success: vi.fn() } }))
+
+vi.mock("next/navigation", () => ({
+  useRouter: () => ({ push: vi.fn(), replace: vi.fn() }),
+  useSearchParams: () => ({ get: () => null }),
+}))
+
+vi.mock("@/components/layout/resizable-three-column-layout", () => ({
+  ResizableThreeColumnLayout: ({ middlePanel }: { middlePanel: React.ReactNode }) => (
+    <div>{middlePanel}</div>
+  ),
+}))
+
+vi.mock("@/components/task/task-conversation-panel", () => ({
+  TaskConversationPanel: () => null,
+}))
+
+vi.mock("@/components/build/agent-builder-chat", () => ({ AgentBuilderChat: () => null }))
+vi.mock("@/components/kb/knowledge-base-creation-dialog", () => ({
+  KnowledgeBaseCreationDialog: () => null,
+}))
+vi.mock("@/components/mcp/connect-mcp-dialog", () => ({
+  ConnectMcpDialog: () => null,
+}))
+vi.mock("@/components/chat/FileMentionDropdown", () => ({ FileMentionDropdown: () => null }))
+vi.mock("@/hooks/use-file-mention", () => ({
+  useFileMention: () => ({
+    checkTrigger: vi.fn(),
+    isOpen: false,
+    items: [],
+    selectedIndex: 0,
+    selectItem: vi.fn(),
+    close: vi.fn(),
+  }),
+}))
+vi.mock("@/components/ui/multi-select", () => ({ MultiSelect: () => null }))
+vi.mock("@/components/ui/select", () => ({ Select: () => null }))
+vi.mock("@/components/build/build-file-preview-sheet", () => ({
+  BuildFilePreviewSheet: () => null,
+}))
+
+import { AgentBuilder } from "./agent-builder"
+
+const AGENT_ID = "5"
+
+function agentResponse(toolCategories: string[]) {
+  return {
+    id: Number(AGENT_ID),
+    user_id: 1,
+    team_id: null,
+    name: "Some Agent",
+    description: "",
+    instructions: "Do the thing",
+    execution_mode: "balanced",
+    // handleCreate's save validation requires a general model to be set.
+    models: { general: 1, small_fast: null, visual: null, compact: null },
+    knowledge_bases: [],
+    skills: [],
+    tool_categories: toolCategories,
+    suggested_prompts: [],
+    logo_url: null,
+    status: "draft",
+    created_at: "2026-01-01T00:00:00Z",
+    updated_at: "2026-01-01T00:00:00Z",
+    widget_enabled: false,
+    allowed_domains: [],
+    share_enabled: false,
+    share_updated_at: null,
+    readonly: false,
+    can_edit: true,
+  }
+}
+
+// The agent was saved before this fix landed, with the unresolved display
+// name -- exactly the shape the original Chrome bug produced.
+const INITIAL_TOOL_CATEGORIES = ["mcp:Chrome"]
+
+let putBody: Record<string, unknown> | undefined
+
+function installApi() {
+  putBody = undefined
+  apiRequestMock.mockImplementation((url: string, init?: RequestInit) => {
+    if (url.endsWith("/api/kb/collections"))
+      return Promise.resolve(new Response(JSON.stringify({ collections: [] }), { status: 200 }))
+    if (url.endsWith("/api/skills/"))
+      return Promise.resolve(new Response(JSON.stringify([]), { status: 200 }))
+    if (url.endsWith("/api/tools/available"))
+      return Promise.resolve(new Response(JSON.stringify({ tools: [] }), { status: 200 }))
+    if (url.endsWith("/api/models/?category=llm"))
+      return Promise.resolve(new Response(JSON.stringify([]), { status: 200 }))
+    if (url.endsWith("/api/models/user-default"))
+      return Promise.resolve(new Response(JSON.stringify([]), { status: 200 }))
+    if (url.includes(`/api/agents/${AGENT_ID}/triggers`))
+      return Promise.resolve(new Response(JSON.stringify([]), { status: 200 }))
+    // The real connected MCPServer row: named after the app_id
+    // ("chrome-devtools"), not the display name -- Chrome's actual
+    // shared-server-catalog convention (_ensure_catalog_app_server).
+    if (url.includes("/api/mcp/servers"))
+      return Promise.resolve(
+        new Response(JSON.stringify([{ id: 1, name: "chrome-devtools" }]), { status: 200 })
+      )
+    if (url.endsWith(`/api/agents/${AGENT_ID}`) && init?.method === "PUT") {
+      putBody = JSON.parse(init.body as string)
+      return Promise.resolve(
+        new Response(JSON.stringify(agentResponse(putBody!.tool_categories as string[])), {
+          status: 200,
+        })
+      )
+    }
+    if (url.endsWith(`/api/agents/${AGENT_ID}`))
+      return Promise.resolve(
+        new Response(JSON.stringify(agentResponse(INITIAL_TOOL_CATEGORIES)), { status: 200 })
+      )
+    return Promise.resolve(new Response(JSON.stringify({}), { status: 200 }))
+  })
+}
+
+describe("AgentBuilder mcp: selector resolution on save", () => {
+  beforeEach(() => {
+    apiRequestMock.mockReset()
+    installApi()
+    ;(globalThis as unknown as { WebSocket: unknown }).WebSocket = vi.fn()
+  })
+
+  afterEach(() => cleanup())
+
+  it("saves the resolved server-row name, not the unresolved display name, and clears isDirty afterward", async () => {
+    render(<AgentBuilder agentId={AGENT_ID} />)
+
+    const nameInput = await screen.findByPlaceholderText(
+      "builds.configForm.name.placeholder"
+    )
+    await waitFor(() => expect(nameInput).toHaveValue("Some Agent"))
+
+    // The Update button starts disabled: nothing has changed yet, and
+    // selectedMcpServers/originalData's MCP extraction agree on the loaded
+    // (still-unresolved) "Chrome" -- loading an already-broken agent
+    // doesn't self-heal it without an edit, by design (round-1 review).
+    const updateButton = screen.getByRole("button", {
+      name: "builds.editor.header.update",
+    })
+    expect(updateButton).toBeDisabled()
+
+    // A trivial, unrelated edit is enough to make isDirty true and reach
+    // the save path -- this test isn't about the name change itself.
+    fireEvent.change(nameInput, { target: { value: "Some Agent Renamed" } })
+    await waitFor(() => expect(updateButton).not.toBeDisabled())
+
+    fireEvent.click(updateButton)
+
+    await waitFor(() => expect(putBody).toBeDefined())
+
+    // (a) The outbound PUT resolved "Chrome" to the real connected row's
+    // name, not the display name that was actually in selectedMcpServers.
+    expect(putBody?.tool_categories).toContain("mcp:chrome-devtools")
+    expect(putBody?.tool_categories).not.toContain("mcp:Chrome")
+
+    // (b) isDirty clears back to false once the save succeeds, observed via
+    // the Update button re-disabling with no further edits. Before the
+    // selectedMcpServers re-seed fix, this stayed enabled forever: the PUT
+    // response's tool_categories rewrote originalData to hold
+    // "chrome-devtools" while selectedMcpServers kept holding "Chrome",
+    // and isDirty's plain string comparison of the two never matched again.
+    await waitFor(() => expect(updateButton).toBeDisabled())
+  })
+})
--- a/frontend/src/components/build/agent-builder.tsx
+++ b/frontend/src/components/build/agent-builder.tsx
@@ -148,17 +148,12 @@ interface TemplateRequirements {
 // instead (issue #802), and `other` is an internal fallback bucket.
 const isAssignableToolCategory = (c: string) => c !== 'agent' && c !== 'other'
 
-// Round-2 review of #1280: selectedMcpServers must be re-seeded from a
-// save's resolved tool_categories the same way loadAgent seeds it from a
-// fetched agent's tool_categories -- otherwise it keeps holding whatever
-// (often unresolved, e.g. a display name) value the user picked, while
-// originalData.tool_categories now holds resolveMcpToolSelector's
-// authoritative resolved name. isDirty's comparison of the two is a plain
-// string compare with no name/id normalization, so a resolved value that
-// differs from the display name (Chrome: "chrome-devtools" vs. "Chrome")
-// left isDirty permanently true after a successful save, with no further
-// action from the user, until a full page reload re-seeded it correctly
-// via loadAgent.
+// The single extraction every tool_categories -> selectedMcpServers site
+// must share: isDirty compares selectedMcpServers against this same
+// extraction applied to originalData.tool_categories with a plain string
+// compare, no name/id normalization -- so a caller that inlines its own
+// copy and drifts from this one silently reintroduces a permanently-dirty
+// isDirty state.
 function mcpServerNamesFromToolCategories(categories: string[]): string[] {
   return categories
     .filter((c) => c.startsWith('mcp:'))
@@ -876,7 +871,7 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
           // never show or round-trip it.
           const rawToolCategories = agent.tool_categories || []
           setSelectedToolCategories(rawToolCategories.filter((c: string) => !c.startsWith('mcp:') && isAssignableToolCategory(c)))
-          setSelectedMcpServers(rawToolCategories.filter((c: string) => c.startsWith('mcp:')).map((c: string) => c.replace('mcp:', '')))
+          setSelectedMcpServers(mcpServerNamesFromToolCategories(rawToolCategories))
           // Seed the SSH auto-category from the saved config so a failed
           // bindings-load (which never fires onCount) can't leave "ssh" unset
           // at save time. A successful load overwrites this with the live count.
@@ -963,9 +958,7 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
           const allCategories = template.agent_config?.tool_categories || []
           setSelectedToolCategories(allCategories.filter((c: string) => !c.startsWith('mcp:') && isAssignableToolCategory(c)))
 
-          const explicitlyConfiguredMcps = allCategories
-            .filter((c: string) => c.startsWith('mcp:'))
-            .map((c: string) => c.replace('mcp:', ''))
+          const explicitlyConfiguredMcps = mcpServerNamesFromToolCategories(allCategories)
 
           // _enrich_template merges connections into tool_categories as mcp: entries, so
           // iterating both explicitlyConfiguredMcps and connections would add each
@@ -1142,16 +1135,24 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
       }
 
       const finalToolCategories = [...selectedToolCategories]
+      // Match handleCreate below: a preview session with a knowledge base
+      // selected must include "knowledge" too, or it runs with different
+      // categories than the agent it's a preview of.
+      if (selectedKbs.length > 0 && !finalToolCategories.includes("knowledge")) {
+        finalToolCategories.push("knowledge")
+      }
       // Resolve each selection to the real, connected MCPServer row's name
       // (see resolveMcpToolSelector for why a name/id fallback alone isn't
       // enough -- the backend uses either convention depending on app
       // type). Depends on mcpServers/officialApps having loaded; if a
       // preview message is sent before they do, this falls back to the raw
       // selector for every MCP tool (matching pre-existing behavior for
-      // this call site, not just this connector).
-      selectedMcpServers.forEach(server => {
-        finalToolCategories.push(`mcp:${resolveMcpToolSelector(server, mcpServers, officialApps)}`)
-      })
+      // this call site, not just this connector). Deduped: two distinct
+      // selectedMcpServers entries can resolve to the same real row.
+      const resolvedMcpSelectors = new Set(
+        selectedMcpServers.map(server => resolveMcpToolSelector(server, mcpServers, officialApps))
+      )
+      resolvedMcpSelectors.forEach(selector => finalToolCategories.push(`mcp:${selector}`))
       if (hasSshBindings && !finalToolCategories.includes("ssh")) {
         finalToolCategories.push("ssh")
       }
@@ -1300,9 +1301,7 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
     if (normalize(selectedSkills) !== normalize(originalData.skills)) return true
 
     // Check MCP servers by extracting them from originalData.tool_categories
-    const originalMcpServers = (originalData.tool_categories || [])
-      .filter((c: string) => c.startsWith('mcp:'))
-      .map((c: string) => c.replace('mcp:', ''))
+    const originalMcpServers = mcpServerNamesFromToolCategories(originalData.tool_categories || [])
     if (normalize(selectedMcpServers) !== normalize(originalMcpServers)) return true
 
     // Check non-MCP tool categories
@@ -1473,10 +1472,14 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
 
     // Add selected MCP servers back into tool_categories, resolved to the
     // real connected MCPServer row's name -- see resolveMcpToolSelector for
-    // why a hard-coded id/name fallback can't work for every app.
-    selectedMcpServers.forEach(server => {
-      finalToolCategories.push(`mcp:${resolveMcpToolSelector(server, mcpServers, officialApps)}`)
-    })
+    // why a hard-coded id/name fallback can't work for every app. Deduped:
+    // two distinct selectedMcpServers entries can resolve to the same real
+    // row, and the backend persists tool_categories verbatim (agents.py),
+    // so an unresolved duplicate here lands in the DB and stays there.
+    const resolvedMcpSelectors = new Set(
+      selectedMcpServers.map(server => resolveMcpToolSelector(server, mcpServers, officialApps))
+    )
+    resolvedMcpSelectors.forEach(selector => finalToolCategories.push(`mcp:${selector}`))
 
     setIsCreating(true)
 
@@ -1593,11 +1596,16 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
 
           // Newly created agents are always personal; promote if the user chose Team.
           const ownershipResult = await reconcileOwnership(newAgent.id.toString(), newAgent.team_id)
-          setOriginalData({ ...newAgent, ...(ownershipResult ?? {}) })
+          // tool_categories: finalToolCategories explicitly, not whatever
+          // newAgent.tool_categories echoes back -- matches the edit-mode
+          // branch above, which sources both originalData and the reseed
+          // below from the same local value rather than mixing a server
+          // response with a local one.
+          setOriginalData({ ...newAgent, tool_categories: finalToolCategories, ...(ownershipResult ?? {}) })
           // Same re-seed as the edit-mode branch above, and for the same
-          // reason: newAgent.tool_categories echoes back the resolved
-          // selectors this request submitted (finalToolCategories), not
-          // the possibly-unresolved values selectedMcpServers still holds.
+          // reason: selectedMcpServers must hold the resolved selectors
+          // this request submitted, not the possibly-unresolved values it
+          // still held from before save.
           setSelectedMcpServers(mcpServerNamesFromToolCategories(finalToolCategories))
           // Only confirm success once ownership resolved. If a promote was blocked
           // (the share-connectors dialog opened, ownershipResult is null), don't
@@ -1891,7 +1899,13 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
       selectedMcpServers.map((serverName) => {
         const connectedServer = findMatchingMcpServer(mcpServers, serverName)
         const matchingApp = findMatchingMcpApp(officialApps, serverName)
-        return connectedServer?.name || matchingApp?.name || serverName
+        // matchingApp?.name (the catalog display name, e.g. "Chrome") first,
+        // not connectedServer?.name (the real MCPServer row name, e.g.
+        // "chrome-devtools"): this is a *display* label. Preferring the row
+        // name here made the chip flip from "Chrome" to "chrome-devtools"
+        // the instant selectedMcpServers gets re-seeded with the resolved
+        // name after a save (see mcpServerNamesFromToolCategories).
+        return matchingApp?.name || connectedServer?.name || serverName
       }),
     [selectedMcpServers, mcpServers, officialApps],
   )
@@ -2656,7 +2670,9 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
                 statusDesc = t("tools.mcp.notSupported")
               }
 
-              const server = { name: connectedServer?.name || matchingApp?.name || serverName, description: statusDesc }
+              // Display label: matchingApp?.name first, same reasoning as
+              // connectorDisplayNames above.
+              const server = { name: matchingApp?.name || connectedServer?.name || serverName, description: statusDesc }
               const icon = getAppIcon(server.name)
               return (
                 <div key={index} className={cn("flex items-center gap-3 p-2 rounded-md border", !isConnected && "opacity-50 bg-muted/50")}>
--- a/frontend/src/components/build/agent-builder.tsx
+++ b/frontend/src/components/build/agent-builder.tsx
@@ -50,7 +50,7 @@ import { KnowledgeBaseCreationDialog } from "@/components/kb/knowledge-base-crea
 import { toast } from "@/components/ui/sonner"
 import { cn } from "@/lib/utils"
 import { getBrandingFromEnv } from "@/lib/branding"
-import { findMatchingMcpApp, findMatchingMcpServer, mcpNameMatches } from "@/lib/mcp-lookup"
+import { findMatchingMcpApp, findMatchingMcpServer, mcpNameMatches, resolveMcpToolSelector } from "@/lib/mcp-lookup"
 import { BuildFilePreviewSheet } from "./build-file-preview-sheet"
 import { TaskConversationPanel } from "@/components/task/task-conversation-panel"
 import { AgentTriggersDialog } from "./agent-triggers-dialog"
@@ -1125,17 +1125,15 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
       }
 
       const finalToolCategories = [...selectedToolCategories]
-      // Same resolution as handleCreate below: selectedMcpServers holds
-      // whatever string the connect flow stored (often the app's display
-      // name), but the backend matches mcp:<selector> against the actual
-      // MCP server row name, which is the catalog app_id for a
-      // shared-server catalog app -- not its display name. Pushing the raw
-      // value broke Chrome (app_id "chrome-devtools" vs. display name
-      // "Chrome"), the first app where the two deliberately differ.
+      // Resolve each selection to the real, connected MCPServer row's name
+      // (see resolveMcpToolSelector for why a name/id fallback alone isn't
+      // enough -- the backend uses either convention depending on app
+      // type). Depends on mcpServers/officialApps having loaded; if a
+      // preview message is sent before they do, this falls back to the raw
+      // selector for every MCP tool (matching pre-existing behavior for
+      // this call site, not just this connector).
       selectedMcpServers.forEach(server => {
-        const connectedServer = findMatchingMcpServer(mcpServers, server)
-        const connectedApp = findMatchingMcpApp(officialApps, server)
-        finalToolCategories.push(`mcp:${connectedServer?.name || connectedApp?.id || server}`)
+        finalToolCategories.push(`mcp:${resolveMcpToolSelector(server, mcpServers, officialApps)}`)
       })
       if (hasSshBindings && !finalToolCategories.includes("ssh")) {
         finalToolCategories.push("ssh")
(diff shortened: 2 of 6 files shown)


frontend/src/components/build/agent-builder.tsx
```diff
@@ -1451,7 +1460,18 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
+      // the actual MCP server row after the app_id
+      // (_ensure_catalog_app_server / _ensure_catalog_mcp_oauth_server), not
+      // the display name. Every pre-existing app's id and name normalize to
+      // the same backend key, so this was previously masked -- Chrome is
+      // the first app where they deliberately differ (app_id
+      // "chrome-devtools" vs. display name "Chrome"), and pushing the name
+      // here sent the backend an mcp:Chrome selector matching no loaded
+      // server, silently loading zero tools (backend log:
+      // reason="no_tools_returned").
+      finalToolCategories.push(`mcp:${connectedServer?.name || connectedApp?.id || server}`)
```
rogercloud: **Major:** preferring `connectedApp?.id` fixes Chrome but regresses apps whose id and name diverge the other way.

Facebook: `app_id: "facebook"`, `name: "Facebook Pages"` (`src/xagent/web/builtin_mcp_registry.py:399-403`), `transport: "oauth"` -> `auth_type: "builtin_oauth"`. Its server row is created by `_ensure_user_mcp_server` (`src/xagent/web/api/auth.py:1294-1297`), which names the row after the **display name** -> normalizes to `facebook_pages` (`src/xagent/core/tools/adapters/vibe/selection_spec.py:61-81`). This fallback builds `mcp:facebook` -> `facebook`, matching nothing -- the same silent zero-tools failure this PR fixes for Chrome, now reproduced for Facebook.

The backend genuinely has two conventions: `_ensure_catalog_app_server` / `_ensure_catalog_mcp_oauth_server` (`src/xagent/web/api/mcp.py:2386,2449`) name rows after `app_id`; `_ensure_user_mcp_server` names them after `name`. A single hard-coded preference can only satisfy one.

Suggested fix -- resolve the app first, then look the row up by the app's identifiers, keeping the raw guess as a true last resort:

```ts
const connectedApp = findMatchingMcpApp(officialApps, server)
const connectedServer =
  findMatchingMcpServer(mcpServers, connectedApp?.id) ||
  findMatchingMcpServer(mcpServers, connectedApp?.name) ||
  findMatchingMcpServer(mcpServers, server)
finalToolCategories.push(`mcp:${connectedServer?.name || connectedApp?.id || server}`)
```

This makes the authoritative `connectedServer?.name` actually hit for Chrome instead of falling through to the `id` guess, and it is correct for `builtin_oauth` apps too.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1451,7 +1460,18 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
     selectedMcpServers.forEach(server => {
       const connectedServer = findMatchingMcpServer(mcpServers, server)
       const connectedApp = findMatchingMcpApp(officialApps, server)
-      finalToolCategories.push(`mcp:${connectedServer?.name || connectedApp?.name || server}`)
+      // connectedApp?.id (the catalog app_id), not connectedApp?.name (its
+      // display name): for a shared-server catalog app, the backend names
+      // the actual MCP server row after the app_id
+      // (_ensure_catalog_app_server / _ensure_catalog_mcp_oauth_server), not
+      // the display name. Every pre-existing app's id and name normalize to
```
rogercloud: **Major:** this comment states a premise that does not hold. Sweeping the catalog in `src/xagent/web/builtin_mcp_registry.py`, two apps have id != name after normalization: `chrome-devtools` / "Chrome" (the one targeted here) and `facebook` / "Facebook Pages" (unaccounted for -- its row is named after the display name by `_ensure_user_mcp_server`, `src/xagent/web/api/auth.py:1294-1297`). The other ~20 apps do normalize identically.

Since this comment is the stated justification for the change, leaving it will lead future readers to trust a false invariant. Please rewrite it to describe the real situation -- the backend uses `app_id` for shared-server catalog apps and the display `name` for `builtin_oauth` apps -- and name Facebook as the known counterexample. The same claim in the PR description ("no behavior change for any pre-existing connector") should be corrected too.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1125,8 +1125,17 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
+      // name), but the backend matches mcp:<selector> against the actual
+      // MCP server row name, which is the catalog app_id for a
+      // shared-server catalog app -- not its display name. Pushing the raw
+      // value broke Chrome (app_id "chrome-devtools" vs. display name
+      // "Chrome"), the first app where the two deliberately differ.
       selectedMcpServers.forEach(server => {
-        finalToolCategories.push(`mcp:${server}`)
+        const connectedServer = findMatchingMcpServer(mcpServers, server)
+        const connectedApp = findMatchingMcpApp(officialApps, server)
+        finalToolCategories.push(`mcp:${connectedServer?.name || connectedApp?.id || server}`)
```
rogercloud: **Minor:** this call site previously pushed the raw selector unconditionally (`mcp:${server}`) -- this PR introduces resolution here for the first time, which is worth flagging since the PR description frames both sites as a fallback-order swap.

The new logic depends on `mcpServers` (state, initialized `[]`, populated asynchronously) and `officialApps` (async context hook). No `isInitialDataLoaded`-style guard appears to wrap the preview panel or gate the send handler, so a preview message sent before those fetches resolve silently falls back to the raw selector for every MCP tool, not just Chrome.

Impact is low in practice (fetches usually land before a user finishes typing, and the fallback matches pre-PR behavior), but it turns a deterministic path into a timing-dependent one. Consider deferring resolution or gating send until both sources have loaded, or at least note the dependency so it reads as intentional.

Note this call site also hits the builtin_oauth regression described on line 1474.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1125,8 +1125,17 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
 
       const finalToolCategories = [...selectedToolCategories]
+      // Same resolution as handleCreate below: selectedMcpServers holds
+      // whatever string the connect flow stored (often the app's display
+      // name), but the backend matches mcp:<selector> against the actual
+      // MCP server row name, which is the catalog app_id for a
+      // shared-server catalog app -- not its display name. Pushing the raw
+      // value broke Chrome (app_id "chrome-devtools" vs. display name
+      // "Chrome"), the first app where the two deliberately differ.
       selectedMcpServers.forEach(server => {
```
rogercloud: **Major (test coverage):** this diff changes no test file, and the logic here has no automated coverage in either direction.

- No `agent-builder-*.test.tsx` asserts on an outbound-built `mcp:` selector (the `"mcp:foo"` in `agent-builder-admin-mcp.test.tsx:132` is an inbound fixture, not an assertion on this code path).
- `frontend/src/lib/mcp-lookup.test.ts` has two cases, both for apps whose id and name normalize identically (Outlook, Google Drive) -- neither exercises the id != name divergence this PR exists to fix.
- `chrome-devtools` appears in zero frontend test files.

This is pure, easily unit-testable logic. Suggest extracting it as `resolveMcpToolSelector(server, mcpServers, officialApps)` into `frontend/src/lib/mcp-lookup.ts` (it is duplicated verbatim at `:1461-1474`, so the extraction also removes the duplication and gives the builtin_oauth fix a single point of change), then covering: `chrome-devtools`/"Chrome" with and without a loaded server row, `facebook`/"Facebook Pages" likewise, and an unknown selector falling through to the raw string.

rogercloud (changes requested):
## Summary

This PR changes how the agent builder turns a selected MCP connector into the `mcp:<selector>` string it sends to the backend. At the publish/create handler (`frontend/src/components/build/agent-builder.tsx:1474`) it swaps the existing fallback chain from `connectedApp?.name` to `connectedApp?.id`, so a shared-server catalog app resolves to its stable `app_id` rather than its display name. At the builder-preview chat handler (`:1136-1138`) it does something different that the description does not call out: the base commit pushed the raw selector with no resolution at all (`mcp:${server}`), and this PR introduces `findMatchingMcpServer` / `findMatchingMcpApp` resolution there for the first time. The motivating bug is Chrome (`app_id` `chrome-devtools` vs. display name `Chrome`), which produced `mcp:Chrome`, matched no server row, and silently loaded zero tools.

## Approach verdict

**Acceptable with reservations.** The fix is correct for the reported Chrome case on the common path, but there are three concerns.

First, it treats a symptom of a backend inconsistency as if the backend were consistent. There are two contradictory row-naming conventions server-side: `_ensure_catalog_app_server` / `_ensure_catalog_mcp_oauth_server` (`src/xagent/web/api/mcp.py:2386,2449`) name the `MCPServer` row after `app_id`, while `_ensure_user_mcp_server` (`src/xagent/web/api/auth.py:1294-1297`) names it after the display `name` for `builtin_oauth`-category apps. Hard-coding a frontend preference for `app_id` picks one of the two conventions and silently mis-serves the other. `/api/mcp/apps` (`src/xagent/web/api/mcp.py:1846`) already resolves the row server-side but returns only `server_id`, never the row's actual `name` -- exposing that name is the durable fix, and would let the frontend stop guessing entirely.

Second, the PR description mischaracterizes its own diff. "Swap the fallback order at both call sites" implies both sites already had resolution logic. Only the second one did; the preview handler gains new, state-dependent resolution where previously there was a deterministic passthrough, which is a real behavior change for every MCP-enabled agent's preview path.

Third, even staying purely frontend, there is a cleaner shape: resolve the catalog app first via `findMatchingMcpApp(officialApps, server)`, then look up the server row by the app's resolved identifiers -- `findMatchingMcpServer(mcpServers, connectedApp?.id) || findMatchingMcpServer(mcpServers, connectedApp?.name)` -- rather than looking the row up by the raw, possibly-display-name selector. For Chrome this makes the authoritative `connectedServer?.name` actually hit instead of falling through to an `id` guess, and it keeps the id/name guess as a genuine last resort for the not-yet-connected case. That shape is also correct for `builtin_oauth` apps, which the current shape is not (see the findings below).

## Findings

See inline comments for line-specific detail (major: builtin_oauth regression, inaccurate comment, missing test coverage; minor: preview-path timing dependency).

Also worth recording, not tied to a specific diff line:

- **Minor -- no migration path for agents already saved with a bad selector.** `loadAgent` (`frontend/src/components/build/agent-builder.tsx:862`) strips the `mcp:` prefix but never re-resolves against `mcpServers` / `officialApps`, so opening an already-broken agent does not self-heal -- only an explicit re-save rewrites the selector. Agents published with `mcp:Chrome` stay silently broken until someone edits and saves them. Blast radius for Chrome is small since it ships with `is_visible_in_connector: False`; Facebook, being a visible connector, has no such protection, which raises the stakes of the first finding below.
- **Minor / informational -- normalizer divergence is the root mechanism.** The frontend normalizer (`frontend/src/lib/mcp-lookup.ts:19`) folds only whitespace to `-`, while the backend (`src/xagent/core/tools/adapters/vibe/selection_spec.py:81`) folds both space and `-` to `_`. That narrower frontend matching is a large part of why `findMatchingMcpServer` cannot resolve Chrome by display name and why the id/name guessing exists at all. Not introduced by this PR, but aligning the two normalizers would remove a standing source of this bug class.

- `L1136`: shrink the duplicated resolution + near-duplicate explanatory comment at `:1136-1138` and `:1461-1474`. Extract `resolveMcpToolSelector(server, mcpServers, officialApps)` into `frontend/src/lib/mcp-lookup.ts` next to the already-exported `findMatchingMcpServer` / `findMatchingMcpApp`, leaving one call per site, one comment, and a directly unit-testable function.

net: -6 lines possible at the two call sites (~21 lines collapse to 2 calls plus a ~15-line shared helper), plus a single point of change for the builtin_oauth fix and a testable seam for the missing coverage.

frontend/src/lib/mcp-lookup.ts
```diff
@@ -52,3 +52,51 @@ export const findMatchingMcpApp = <T extends McpLookupReference>(
...
+      ? findMatchingMcpServer(mcpServers, connectedApp.id)
+      : undefined) ||
+    (typeof connectedApp?.name === "string" && connectedApp.name
+      ? findMatchingMcpServer(mcpServers, connectedApp.name)
+      : undefined)
+
+  if (typeof connectedServer?.name === "string" && connectedServer.name) {
+    return connectedServer.name
+  }
+  if (typeof connectedApp?.id === "string" && connectedApp.id) {
```
rogercloud: This last-resort fallback to `connectedApp.id` guesses a convention, and it guesses wrong for `builtin_oauth` apps — whose rows are named after the display name by `_ensure_user_mcp_server`. That can corrupt a category that was previously working, not just fail to improve an unresolvable one.

Concretely: `selectedMcpServers` is seeded by stripping `mcp:` off the saved categories (`frontend/src/components/build/agent-builder.tsx:862`), while the `mcpServers` state only holds rows visible to the *current* viewer — team-scoping in `connector_team_scope.py` is opt-in via `set_connector_team_hooks`, which shipped app code never calls, so a teammate's personal MCP connector row is invisible by default. If someone edits and re-saves a shared/team agent whose saved selector points at a connector they can't see, step 1 through 3 all miss, and this line rewrites `mcp:Facebook Pages` into `mcp:facebook` — which then matches nothing, and the owner silently loses those tools on their own agent.

Worth flagging that this exact fallback shape (`connectedServer?.name || connectedApp?.id || server`) is what the previous review round suggested, so this is guidance worth revisiting jointly rather than something introduced here independently.

Suggested fix: when no server row is found, return the incoming `server` unchanged and drop the `connectedApp.id` step entirely. Preserving an unresolvable selector as-is is strictly safer than rewriting it to a guessed convention — the current value may already be correct for a row this viewer simply cannot see.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1447,11 +1454,16 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
+    // type, so this resolves the real connected row rather than guessing
+    // at a single fallback (a guess broke Chrome, app_id "chrome-devtools"
+    // vs. display name "Chrome", with the pre-existing "prefer name"
+    // fallback; the reverse "prefer id" fallback breaks Facebook, app_id
+    // "facebook" vs. display name "Facebook Pages", the other direction).
     selectedMcpServers.forEach(server => {
-      const connectedServer = findMatchingMcpServer(mcpServers, server)
-      const connectedApp = findMatchingMcpApp(officialApps, server)
-      finalToolCategories.push(`mcp:${connectedServer?.name || connectedApp?.name || server}`)
+      finalToolCategories.push(`mcp:${resolveMcpToolSelector(server, mcpServers, officialApps)}`)
```
rogercloud: The resolved value pushed here creates a permanent false "unsaved changes" state (frontend/src/components/build/agent-builder.tsx:1289, hasUnsavedChanges) for any app whose id differs from its display name.

`hasUnsavedChanges` compares `selectedMcpServers` (holding the display name, e.g. `"Chrome"`) against the MCP entries extracted from `originalData.tool_categories`, which after a save contain the *resolved* row name (e.g. `"chrome-devtools"`, produced at :1465-1466 and snapshotted into `originalData` at :1540). The comparison is a plain case-sensitive string join, and `selectedMcpServers` is never re-seeded after a successful save — so for every Chrome-class app the two sides stay out of sync and `isDirty` returns true forever: the Save button stays enabled and the leave-page warning keeps firing until a full page reload.

This did not happen before this PR — the base commit's fallback preferred `connectedApp?.name`, which matched what `selectedMcpServers` holds.

Two ways out: either compare both sides through `resolveMcpToolSelector` / `mcpNameMatches` instead of raw string equality, or re-seed `selectedMcpServers` from the resolved `finalToolCategories` after a successful save so the two representations converge.

frontend/src/lib/mcp-lookup.ts
```diff
@@ -52,3 +52,51 @@ export const findMatchingMcpApp = <T extends McpLookupReference>(
...
+// other way, and a hard-coded id-first or name-first fallback silently
+// breaks whichever app uses the other convention.
+//
+// So this resolves the *real* server row instead of guessing at its name:
+// find the catalog app the selector refers to, then look up an actual
+// MCPServer row by that app's id, then by its name -- whichever convention
+// created the row, this finds it and returns its authoritative real name.
+// Only when no server row exists at all (not yet connected) does this fall
+// back to the app's id, then to the raw selector unchanged.
+export const resolveMcpToolSelector = <T extends McpLookupReference>(
```
rogercloud: One generic parameter `T` is bound across both `mcpServers: T[]` and `officialApps: T[]`, but these are two structurally different shapes — server rows versus catalog apps. They only unify today because `mcpServers` state is typed `any[]` (`frontend/src/components/build/agent-builder.tsx:356`) while `officialApps` is properly typed `McpApp[]`; the moment `mcpServers` gets a real type, this signature stops compiling at the call sites.

Note the pre-existing `findMatchingMcpServer` / `findMatchingMcpApp` each bind `T` from a single array, so there's no existing-convention argument for coupling them here. Suggest `<S extends McpLookupReference, A extends McpLookupReference>(server: string, mcpServers: S[], officialApps: A[])`.

frontend/src/lib/mcp-lookup.test.ts
```diff
@@ -22,3 +27,40 @@ describe("MCP lookup helpers", () => {
...
+
+  it("resolves a connected app-id-named server row (chrome-devtools convention)", () => {
+    const mcpServers = [{ name: "chrome-devtools" }]
+    expect(resolveMcpToolSelector("Chrome", mcpServers, [chromeApp])).toBe(
+      "chrome-devtools"
+    )
+  })
+
+  it("resolves a connected name-named server row (facebook builtin_oauth convention)", () => {
+    const mcpServers = [{ name: "Facebook Pages" }]
```
rogercloud, line 67: This fixture doesn't reflect the real API shape, so the test doesn't cover what its name says it covers. In production Facebook's server row *does* carry `app_id: "facebook"` in the list response — because its row name matches the catalog app's name exactly, `_enrich_oauth_server_info` populates `MCPServerResponse.app_id` for `transport == "oauth"` rows. That means the real row resolves at step 1 and never reaches the step-3 name fallback this test is meant to demonstrate.

Suggest two changes: make this fixture realistic as `{ name: "Facebook Pages", app_id: "facebook" }`, and add a separate fixture that genuinely exercises step 3 only — a row whose name matches the app's display name and which carries no id at all.

frontend/src/lib/mcp-lookup.test.ts
```diff
@@ -22,3 +27,40 @@ describe("MCP lookup helpers", () => {
...
+    expect(resolveMcpToolSelector("facebook", mcpServers, [facebookApp])).toBe(
+      "Facebook Pages"
+    )
+  })
+
+  it("falls back to the app id when the app is known but no server row is connected yet", () => {
```
rogercloud: This test asserts and locks in exactly the behavior flagged on `frontend/src/lib/mcp-lookup.ts:98`. Noting it here so the two move together rather than as a standalone ask: if that fallback is changed to return `server` unchanged, this case needs inverting to assert the raw selector is preserved rather than the app id being substituted.

frontend/src/lib/mcp-lookup.test.ts
```diff
@@ -22,3 +27,40 @@ describe("MCP lookup helpers", () => {
...
+    expect(resolveMcpToolSelector("Chrome", [], [chromeApp])).toBe(
+      "chrome-devtools"
+    )
+  })
+
+  it("falls back to the raw selector when nothing matches at all", () => {
```
rogercloud: Three gaps worth closing while this file is new (static reading of the fixtures, not a test run):

(a) Every case uses a single-element `mcpServers` array, so nothing would catch a wrong-row match. A case with several rows where only one should win would give the lookup ordering real protection.

(b) The round-trip isn't covered: `loadAgent` (`frontend/src/components/build/agent-builder.tsx:862`) strips `mcp:` off the saved category and feeds an already-resolved real server name straight back into `resolveMcpToolSelector` on the next save. It happens to work today via step 1, but nothing pins that.

(c) Empty-string selector: `resolveMcpToolSelector("", [], [])` returns `""`, which becomes the category `"mcp:"` and gets silently added as an unmatchable scope entry downstream in `selection_spec.py` rather than being rejected or filtered. Worth a test that documents the intended behavior, whatever it is.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1447,11 +1454,16 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
     }
 
-    // Add selected MCP servers back into tool_categories
+    // Add selected MCP servers back into tool_categories. See
+    // resolveMcpToolSelector: the backend names a catalog app's MCPServer
+    // row after either its app_id or its display name depending on app
+    // type, so this resolves the real connected row rather than guessing
+    // at a single fallback (a guess broke Chrome, app_id "chrome-devtools"
+    // vs. display name "Chrome", with the pre-existing "prefer name"
+    // fallback; the reverse "prefer id" fallback breaks Facebook, app_id
```
rogercloud: The backend's two-naming-conventions rationale is now duplicated near-verbatim in three places: the `resolveMcpToolSelector` doc comment (`frontend/src/lib/mcp-lookup.ts:56-78`, 23 lines) plus both call sites here (:1128-1134 and :1457-1464) — roughly 40 comment lines guarding a 24-line function.

This copy has already drifted from accuracy: it says "the reverse 'prefer id' fallback breaks Facebook," but against the live API shape Facebook's row already carries `app_id` (populated by `_enrich_oauth_server_info` for oauth rows), so an id-first fallback would in fact have resolved it. Framing an id-first approach as a live breakage is misleading to whoever reads this next.

Suggest collapsing both call-site comments to a one-line pointer at `resolveMcpToolSelector` and keeping the full rationale in exactly one place, where it can be corrected once.

rogercloud (changes requested):
## Summary

This PR extracts a new `resolveMcpToolSelector` helper in `frontend/src/lib/mcp-lookup.ts` and uses it at both places where the agent builder turns a selected connector into an `mcp:<selector>` tool category. Instead of guessing at the backend's row-naming convention with a single hard-coded fallback, it now looks up the actual connected `MCPServer` row (by the raw selector, then by the catalog app's id, then by the app's name) and emits that row's authoritative name. A new test file covers the two divergent naming conventions.

The underlying problem is real and the diagnosis is accurate: the backend genuinely has two contradictory conventions for naming a catalog app's `MCPServer` row — `_ensure_catalog_app_server` / `_ensure_catalog_mcp_oauth_server` in `src/xagent/web/api/mcp.py` name the row after `app_id`, while `_ensure_user_mcp_server` in `src/xagent/web/api/auth.py` (used for `builtin_oauth` apps) names it after the display name. The consumer in `src/xagent/core/tools/adapters/vibe/mcp_tools.py` matches purely by name with no id fallback, so a wrong guess silently drops every tool from that connector with no warning at all. Resolving against the real row is the right instinct, and this is a genuine improvement over the base commit — it fixes the reported Chrome bug, and the preview-handler call site is now strictly better than the base's completely unresolved passthrough.

## Approach verdict: acceptable, with reservations

Two design-level notes, neither of which needs to block this PR:

**One-shot snapshot resolution (minor / informational).** Resolution to the real server-row name happens exactly once, client-side, at save time, and the result is persisted into `tool_categories`. `MCPServer` rows can be renamed afterwards (`update_mcp_server` in `src/xagent/web/api/mcp.py`), and there is no backend-side re-resolution — so a rename after save silently orphans the stored selector. This is a pre-existing property of the design rather than something the PR invents, but the PR does add a second save-time resolution call site, which entrenches it a little further. Worth knowing about; not something to fix here.

**Symptom vs. root cause (non-blocking follow-up suggestion).** The frontend is only one of three producers of `mcp:<selector>` strings. The matcher itself — `src/xagent/core/tools/adapters/vibe/mcp_tools.py` — has no id/name fallback whatsoever, so making it tolerant of either convention would cover every producer at once, including `src/xagent/templates/manager.py` (`f"mcp:{conn_name}"`, untouched here, carrying the same latent bug) and `frontend/src/components/build/agent-builder-chat.tsx` (currently unreachable, since its `toolCategories` prop is never passed — but it would reproduce this exact bug the moment it gets wired up). To be clear, this is a suggestion for a follow-up, not a demand on this PR: a backend matcher change is a bigger and riskier change than this PR's scope, and the frontend fix here is sufficient to close the reported Chrome bug.

## Findings

Two major issues, both introduced by this PR, are the reason for requesting changes; the rest are minor.

**Major**

1. `frontend/src/lib/mcp-lookup.ts:98` — when no connected server row is found at all, the function falls back to `connectedApp.id`, which is the wrong convention for `builtin_oauth` apps and can corrupt a previously-working saved category. Worth revisiting jointly, since this specific fallback came out of the previous review round's suggestion.
2. `frontend/src/components/build/agent-builder.tsx:1289` — a new permanent false "unsaved changes" state for any app whose id differs from its display name.

**Minor**

3. `frontend/src/lib/mcp-lookup.ts:79` — single generic `T` shared by two structurally different array shapes.
4. `frontend/src/lib/mcp-lookup.test.ts:49` — the Facebook fixture doesn't match the real API shape, so the test doesn't exercise the path it claims to.
5. `frontend/src/lib/mcp-lookup.test.ts:55` — this test locks in the behavior flagged in finding 1 and should move together with it.
6. `frontend/src/lib/mcp-lookup.test.ts:61` — three coverage gaps (multi-row `mcpServers`, already-resolved selector round-trip, empty-string selector).
7. `frontend/src/components/build/agent-builder.tsx:1463` — the naming-convention rationale is duplicated three times and the third copy has already drifted from accuracy.

Test observations above come from reading the test file, not from running the suite locally.

Lean already.

frontend/src/lib/mcp-lookup.test.ts
```diff
@@ -22,3 +27,97 @@ describe("MCP lookup helpers", () => {
     expect(mcpNameMatches("google-drive", "Google Drive")).toBe(true)
   })
 })
+
+describe("resolveMcpToolSelector", () => {
```
rogercloud, line 31: **Major — coverage gap.** These cover the pure helper only; the actual behavioral fix is untested: (a) the outbound PUT/POST body contains the *resolved* selector, (b) `isDirty` clears after a successful save. The PR's rationale ("no test harness exercising isDirty/save flows ... building one is a larger, separate effort") overstates the cost: 8 files (~2800 lines) already mount `AgentBuilder` with mocked `apiRequest`/`useMcpApps`/`/api/mcp/servers`; `agent-builder-admin-mcp.test.tsx` loads an edit-mode fixture with `tool_categories: ["mcp:foo"]`, and `agent-builder-publish-errors.test.tsx` fires create/update/publish end-to-end. Fair caveat: none currently assert request payloads, so it isn't a fixture swap — but the mount/mock infra exists, and what's missing is assertion code on top, not a harness.

frontend/src/lib/mcp-lookup.test.ts
```diff
@@ -22,3 +27,97 @@ describe("MCP lookup helpers", () => {
...
+    // tool_categories entry, which after any save already holds the
+    // resolved real name -- the very next save must not perturb it.
+    const mcpServers = [{ name: "chrome-devtools" }]
+    expect(
+      resolveMcpToolSelector("chrome-devtools", mcpServers, [chromeApp])
+    ).toBe("chrome-devtools")
+  })
+
+  it("preserves the raw selector unchanged when the app is known but no server row is connected yet", () => {
+    // Round-2 review: guessing connectedApp.id here was itself a
```
rogercloud: Nit: "Round-2 review: guessing connectedApp.id..." bakes a review-round artifact into a permanent test comment. State the invariant being locked in (raw input is preserved, never a guessed id) — that's what a future reader needs.

frontend/src/lib/mcp-lookup.ts
```diff
@@ -52,3 +52,56 @@ export const findMatchingMcpApp = <T extends McpLookupReference>(
...
+  const connectedServer =
+    findMatchingMcpServer(mcpServers, server) ||
+    (typeof connectedApp?.id === "string" && connectedApp.id
+      ? findMatchingMcpServer(mcpServers, connectedApp.id)
+      : undefined) ||
+    (typeof connectedApp?.name === "string" && connectedApp.name
+      ? findMatchingMcpServer(mcpServers, connectedApp.name)
+      : undefined)
+
+  if (typeof connectedServer?.name === "string" && connectedServer.name) {
```
rogercloud: **Minor.** A matched row whose `name` is empty or whitespace-only falls through to the raw selector — an undocumented third outcome. The doc comment's "only when no server row can be found at all" is inaccurate here: a row *is* found (via `app_id`), its `name` just fails the truthy check. And `"   "` is truthy in JS, so it'd be emitted verbatim as `mcp:   `. Reachability is narrow (empty names are blocked by validation on every path checked; whitespace-only needs bypassing the UI `.trim()` and hitting the API directly) — but correct the comment and add a test for the whitespace case.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -148,6 +148,23 @@ interface TemplateRequirements {
 // instead (issue #802), and `other` is an internal fallback bucket.
 const isAssignableToolCategory = (c: string) => c !== 'agent' && c !== 'other'
 
+// Round-2 review of #1280: selectedMcpServers must be re-seeded from a
```
rogercloud: Nit: "Round-2 review of #1280:" — this file's convention is to cite issue numbers (#802, #928, #1179), not review rounds; the round number means nothing once merged. The 11-line block also outweighs the 4-line function it documents. Compress to the invariant itself.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -148,6 +148,23 @@ interface TemplateRequirements {
...
+// fetched agent's tool_categories -- otherwise it keeps holding whatever
+// (often unresolved, e.g. a display name) value the user picked, while
+// originalData.tool_categories now holds resolveMcpToolSelector's
+// authoritative resolved name. isDirty's comparison of the two is a plain
+// string compare with no name/id normalization, so a resolved value that
+// differs from the display name (Chrome: "chrome-devtools" vs. "Chrome")
+// left isDirty permanently true after a successful save, with no further
+// action from the user, until a full page reload re-seeded it correctly
+// via loadAgent.
+function mcpServerNamesFromToolCategories(categories: string[]): string[] {
```
rogercloud, line 157: **Minor — future-drift risk.** This duplicates filter+strip logic still inlined at ~879 (`loadAgent`), ~968 (template load), and ~1305 (`isDirty`'s `originalMcpServers`). Harmless today, but the fix's correctness rests on the helper and `isDirty`'s inline copy staying in agreement — edit one without the other and the permanent-dirty bug this PR fixes comes back silently. Point all three at this helper.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1125,8 +1142,15 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
 
       const finalToolCategories = [...selectedToolCategories]
+      // Resolve each selection to the real, connected MCPServer row's name
+      // (see resolveMcpToolSelector for why a name/id fallback alone isn't
+      // enough -- the backend uses either convention depending on app
+      // type). Depends on mcpServers/officialApps having loaded; if a
+      // preview message is sent before they do, this falls back to the raw
+      // selector for every MCP tool (matching pre-existing behavior for
+      // this call site, not just this connector).
       selectedMcpServers.forEach(server => {
```
rogercloud: **Minor — no dedup after resolution.** Two distinct `selectedMcpServers` entries can resolve to the same row, pushing duplicate `mcp:X`. The backend dedups via a `Set` in `selection_spec.py`, but `agents.py:689,847` persist `tool_categories` verbatim — so the duplicate lands in the DB and reappears in the chip list on every load. One-liner: `new Set(...)` around the resolved values, or dedup inside `mcpServerNamesFromToolCategories`. (Same visible symptom as the `connect-mcp-dialog` toggle bug above, different cause — worth fixing both.) Same gap at the create-mode call site below (~1477).

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1125,8 +1142,15 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
...
-        finalToolCategories.push(`mcp:${server}`)
+        finalToolCategories.push(`mcp:${resolveMcpToolSelector(server, mcpServers, officialApps)}`)
       })
       if (hasSshBindings && !finalToolCategories.includes("ssh")) {
```
rogercloud, line 1156: **Minor.** This block pushes `"ssh"` but never `"knowledge"`, while `handleCreate` (~1466-1474) pushes both — so a preview session with a knowledge base selected runs without the `knowledge` category, diverging from the agent that eventually gets saved.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1447,11 +1471,11 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
       finalToolCategories.push("ssh")
     }
 
-    // Add selected MCP servers back into tool_categories
+    // Add selected MCP servers back into tool_categories, resolved to the
+    // real connected MCPServer row's name -- see resolveMcpToolSelector for
+    // why a hard-coded id/name fallback can't work for every app.
     selectedMcpServers.forEach(server => {
```
rogercloud: Same missing dedup as the preview path at ~1152 — wrap the resolved values in a `Set` before building `finalToolCategories`.

frontend/src/components/build/agent-builder.tsx
```diff
@@ -1565,6 +1594,11 @@ export function AgentBuilder({ agentId }: AgentBuilderProps) {
           // Newly created agents are always personal; promote if the user chose Team.
           const ownershipResult = await reconcileOwnership(newAgent.id.toString(), newAgent.team_id)
           setOriginalData({ ...newAgent, ...(ownershipResult ?? {}) })
```
rogercloud: Nit — inconsistent sourcing: `originalData` is seeded from the server response (`newAgent`) while `selectedMcpServers` (below) comes from local `finalToolCategories`, even though the comment argues `newAgent.tool_categories` echoes back exactly what was submitted (which implies both should come from one source). Edit-mode (~1540-1560) derives both from `finalToolCategories` consistently; only create-mode mixes. Benign today — backend normalization is a no-op on this already-filtered set — but gratuitous.

rogercloud (changes requested):
## Summary

Adds `resolveMcpToolSelector()` to `frontend/src/lib/mcp-lookup.ts` and routes both `mcp:<selector>` construction sites in `agent-builder.tsx` through it, so the category string is derived from the actually-connected `MCPServer` row (matched by id, then by name) instead of guessed from the catalog app. When no row can be matched the raw input is preserved unchanged rather than substituting a guessed id — a genuine improvement over this PR's own round 1. It also re-seeds `selectedMcpServers` from the resolved selectors after both save paths succeed, fixing the permanently-true `isDirty` state. Diff is 3 frontend files (+200/-7), with new unit tests for the helper.

The shared-helper shape fits `mcp-lookup.ts`, the id → name → preserve-input precedence is the right ordering, and there is no speculative abstraction — both new functions have exactly the call sites the fix requires.

**One significant new design gap: the residual silent-failure mode is still open.** `resolveMcpToolSelector` falls back to returning the raw unresolved selector, and both call sites (`agent-builder.tsx:1144-1157`, `1477-1479`) then proceed to save with no warning, log, or toast. When `mcpServers`/`officialApps` haven't loaded yet — the race the comment at 1144-1150 explicitly acknowledges and dismisses as "matching pre-existing behavior" — or when the row isn't visible to this viewer, the outcome is exactly the bug this PR sets out to fix: an agent persisted with a selector that loads zero tools, with nothing surfaced to the user. The fix narrows the window; it doesn't close the failure class. A `console.warn` plus a save-time toast when a selected connector resolves to no row would close it.

Two design points I raised in round 1 — the backend's two naming conventions being the actual root cause, and pre-PR-broken agents having no self-heal path through a plain save — remain open. Both were already adjudicated as non-blocking (backend matcher change out of scope; repair path Minor given Chrome's `is_visible_in_connector: False`), and I'm not re-litigating either. Same for the latent third `mcp:` construction site in `agent-builder-chat.tsx:230` — still dead code, still a known non-blocking latent case.

## Major

1. **Re-seeding with the resolved name makes a connected connector impossible to cleanly deselect, and can duplicate it.** `frontend/src/components/mcp/connect-mcp-dialog.tsx` isn't touched by this PR's diff, so I can't leave an inline comment on it, but it's directly reachable because of the re-seed here: after save, `selectedMcpServers` holds `"chrome-devtools"`. In `connect-mcp-dialog.tsx`, `isSelected` (~1222) matches by `app.id` OR `app.name`, so the Chrome card renders as selected. But `handleCardClick` (~915-925) toggles by `app.name` only, unnormalized — `["chrome-devtools"].includes("Chrome")` is false, so clicking the visually-selected card **appends** `"Chrome"`, giving `["chrome-devtools", "Chrome"]`: duplicated `mcp:` category, connector rendered twice in the chip list, and still visually selected. The user oscillates between duplicated and non-duplicated-but-still-selected and can never reach "truly deselected." The id/name asymmetry predates this PR but is only *reachable* because of the re-seed added here, so it needs handling as part of this change. Normalize the toggle to remove by either id or name, matching `isSelected`.

2. **Test coverage doesn't reach the actual behavioral fix, and the stated reason is inaccurate.** See inline comment on `mcp-lookup.test.ts`.

## Minor

1. `agent-builder.tsx`'s `connectorDisplayNames` (~1890-1896) and the chip row (~2642-2659) aren't touched by this PR's diff either, so no inline comment, but they're worth flagging here too: both prefer `connectedServer?.name` over `matchingApp?.name`, so the chip label flips from "Chrome" to "chrome-devtools" the instant you hit Save. The preference itself is pre-existing (it already showed after a reload); the re-seed just makes it immediate and in-session. Swap to `matchingApp?.name || connectedServer?.name`.

See inline comments on the relevant lines for the rest: an undocumented third fallthrough outcome for empty/whitespace-only server names, missing dedup after resolution at both call sites, three inlined copies of the new extraction helper's logic, review-process artifacts baked into permanent comments, inconsistent seeding source in create-mode, and the preview path omitting `"knowledge"` while save includes it.

## Nit (description prose, no line ref)

The description's "matched by coincidence" framing isn't quite right — `normalize_mcp_server_name`'s hyphen/space/case folding is deterministic, which is *why* every prior pair collapsed to the same key, and why Chrome diverging by a whole word was the first break. Worth rewording for whoever reads this next.

## Verification caveat

`node_modules` is not installed in the review environment, so the claimed `npx vitest run` (80 passed) and `npx tsc --noEmit` results could **not** be independently executed. Every assertion in the new test file was hand-traced against the implementation and found reachable and correct — but this review makes no claim that the suite passes.

No over-engineering found: both new functions are lean, with exactly the call sites the fix requires.

frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+describe("AgentBuilder mcp: selector resolution on save", () => {
+  beforeEach(() => {
+    apiRequestMock.mockReset()
+    installApi()
+    ;(globalThis as unknown as { WebSocket: unknown }).WebSocket = vi.fn()
+  })
+
+  afterEach(() => cleanup())
+
+  it("saves the resolved server-row name, not the unresolved display name, and clears isDirty afterward", async () => {
```
rogercloud, line 221: Minor: coverage gaps. This test covers edit-mode PUT resolution and isDirty clearing end to end, but the other changed paths remain untested — the create-mode POST branch (incl. its dedicated setOriginalData override + re-seed at agent-builder.tsx:1604-1609), the preview path (incl. the knowledge push), Set dedup at both call sites (agent-builder.tsx:1152, 1479), the connect-mcp-dialog toggle-off (connect-mcp-dialog.tsx:915-930), and the async-load race. The harness now exists, so these are cheap near-copies of this same test.

frontend/src/lib/mcp-lookup.ts
```diff
@@ -52,3 +52,70 @@ export const findMatchingMcpApp = <T extends McpLookupReference>(
...
+      : undefined) ||
+    (typeof connectedApp?.name === "string" && connectedApp.name
+      ? findMatchingMcpServer(mcpServers, connectedApp.name)
+      : undefined)
+
+  if (typeof connectedServer?.name === "string" && connectedServer.name.trim()) {
+    return connectedServer.name
+  }
+  if (server) {
+    console.warn(
```
rogercloud, line 116: Minor/design: the warn is dev-visible only. A user who saves a selection whose connector resolves to no connected row still gets zero tools with no user-facing signal; the doc comment above explicitly defers a toast decision. Acceptable for this PR, but a save-time user-facing signal remains an open, non-blocking follow-up.

frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx
```diff
@@ -0,0 +1,260 @@
+import React from "react"
+import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react"
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+
+// #1280 round-3 review, Major finding 2: the pure resolveMcpToolSelector unit
```
rogercloud: Nit: permanent comments reference review rounds ("#1280 round-3 review, Major finding 2"; also "(round-1 review)" at line 232). Same class as the earlier review-round-comment nits that were fixed in mcp-lookup.test.ts and agent-builder.tsx by citing issue numbers — please do the same here.

rogercloud (approved):

rogercloud (approved):
## Follow-up review (round 4)

### PR summary
This PR fixes agent/task creation in `frontend/src/components/build/agent-builder.tsx`, which builds `mcp:<selector>` tool-category strings for each MCP connector the user selects. The backend matches those selectors against the connected `MCPServer` row's name only (`src/xagent/core/tools/adapters/vibe/mcp_tools.py`, no id fallback), while catalog rows are named by two contradictory conventions — `app_id` for api-key/keyless and remote-MCP-OAuth apps, display name for builtin-OAuth apps — so the frontend's earlier id/name guessing worked by coincidence and silently broke for Chrome (#1179): a name-first guess persisted `mcp:Chrome`, matched no row, and loaded zero tools with no error. The fix adds `resolveMcpToolSelector` (`frontend/src/lib/mcp-lookup.ts`), which resolves a selection against the catalog app and the actually-connected `MCPServer` row (selection → row, app id → row, app name → row), returns the row's authoritative name, and returns the input unchanged on failure (never guesses), and wires it into both save paths with Set dedup plus a symmetric connect-dialog toggle.

### Update summary
One commit since the last review (`e62cf7f8`): `188a852d` "fix: address round-3 review findings on the mcp: selector resolution" (5 files, +357/-47). It adds a new end-to-end component test (`agent-builder-mcp-selector-resolution.test.tsx`), Set dedup at both resolution call sites, a "knowledge" push in the preview handler (the direct fix for inline comment 3767928556), a shared `mcpServerNamesFromToolCategories` extraction used by loadAgent/template-load/isDirty/re-seeds, a create-mode `setOriginalData` override plus re-seed, display-label preference flips, a symmetric dialog toggle-off, a blank-name trim guard, and a `console.warn` on unresolved selectors.

### Approach verdict
**Sound.** Resolving the selector against the real connected row instead of guessing is the right client-side fix for the two naming conventions; the never-guess fallback preserves already-correct selectors, and the new end-to-end test pins the two invariants the fix depends on (resolved selector in the outbound PUT body; `isDirty` clearing after save). Design notes:
- *Accepted tradeoffs:* the root cause (backend naming conventions) is encoded client-side rather than fixed server-side — documented in `mcp-lookup.ts`; re-seeding `selectedMcpServers` couples editor state to persisted row names and required four companion changes kept in sync — now pinned by the component test; Set dedup is a behavior change from base (coverage folded into M1).
- *Follow-ups:* silent failure is narrowed but not closed (M4); the knowledge push is correct but undocumented in the PR body and untested (M3).

### Prior-findings checklist
| Status | Findings | Notes |
|---|---|---|
| FIXED | P1, P2, P4, P7–P13, P16–P26 | All previously-blocking Major findings among them; verified against code and the new test. 19 inline threads ready to resolve (P16/P26 were body-only). |
| PARTIAL | P3, P28 | `console.warn` added (`mcp-lookup.ts:116`) and rationale documented; save-time user-facing toast still deferred — see M4. |
| NOT FIXED | P5, P27 | P5: `loadAgent` still doesn't re-resolve broken saved selectors — previously adjudicated non-blocking; P27: PR description prose ("worked by coincidence") — prose-only nit, non-blocking. |
| WAIVED | P6, P14, P15 | P6: normalizer divergence, pre-existing, informational; P14: one-shot snapshot resolution, informational; P15: root-cause follow-up, explicitly out of scope. |

### Remaining findings (all minor)
- **M1 (Minor)** — `frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx:221`: coverage gaps. The single test covers edit-mode PUT resolution and isDirty clearing only; the create-mode POST branch (incl. its dedicated `setOriginalData` override + re-seed at `agent-builder.tsx:1604-1609`), the preview path (incl. the knowledge push), Set dedup (`agent-builder.tsx:1152`, `1479`), the connect-dialog toggle-off (`connect-mcp-dialog.tsx:915-930`), and the async-load race remain untested. The harness now exists, so these are cheap near-copies of the same test.
- **M2 (Minor, body-only)** — the PR body's "Known gap" paragraph is stale: it claims no test harness exists for the isDirty/save flows, but this PR adds exactly that harness and test. Update the paragraph.
- **M3 (Minor, body note)** — the knowledge push in the preview handler is absent from the PR body and untested. It is the direct fix of inline comment 3767928556 and is announced in the commit message, so the code itself is correct; document it in the PR body and ideally cover it per M1.
- **M4 (Minor)** — `frontend/src/lib/mcp-lookup.ts:116`: silent failure narrowed, not closed. The `console.warn` is dev-visible only; a user who saves a selection whose connector resolves to no connected row still gets zero tools with no user-facing signal (a toast is explicitly deferred in the doc comment above). Non-blocking follow-up.
- **M5 (Nit)** — `frontend/src/components/build/agent-builder-mcp-selector-resolution.test.tsx:5` (also `:232`): permanent comments reference review rounds ("#1280 round-3 review, Major finding 2", "(round-1 review)") — same class as P18/P20, which were fixed elsewhere by citing issue numbers. Replace with issue-number references.

- L1152: shrink resolve-then-dedup temp block in the preview path. `finalToolCategories.push(...new Set(selectedMcpServers.map(server => 'mcp:' + resolveMcpToolSelector(server, mcpServers, officialApps))))` (net -2).
- L1479: shrink the identical resolve-then-dedup temp block in handleCreate; same replacement (net -2).
- net: -4 lines.

### Closing note
All previously-blocking (Major) findings are fixed; this round has no blocking issues. Remaining items are minor coverage/documentation follow-ups and two optional shrinks.

PR #1283, opened 2026-08-11
feat(interactions): resolve the interaction anchor and open the handoff surface

AlexLiu190625:
## Summary

Adds `resolve_interaction_anchor`, the primary-key read against `trace_events` that resolves a task's interaction anchor from its checkpoint pointer, with zero production callers. Moves the checkpoint run-partition predicate into the services layer so future services-side consumers need not reverse-import the api layer; the new resolver checks the same condition with its own in-Python comparison and does not call the predicate. Gives the interaction handoff primitive a public produced-object type and a `staged` observability field. Adds two PostgreSQL-only tests pinning backend behavior that previously lived only in comments.

## Anchor resolution

`resolve_interaction_anchor(db, task)` reads `task.last_checkpoint_trace_event_id` and classifies the result into one of six outcomes, in a fixed order: no `run_id` (absence), no pointer (absence), a dangling pointer (unavailable, reported as absence), a row that fails one of six self-consistency conditions (corrupt), a row with a legacy `checkpoint_type` (absence, not corrupt), or a row that passes everything (a resolved anchor). The corrupt-before-legacy ordering matters: a row belonging to a different task must be reported corrupt even when its `checkpoint_type` is legacy, never silently absorbed into the legacy-absence branch.

An unavailable pointer is reported to the caller as absence and does not fall back to scanning the legacy `last_checkpoint_event_id` column the way the trace-read side's fallback does — the two paths are deliberately not the same shape.

The resolver ships with zero production callers, guarded by its own static AST assertion scoped to this function specifically (not an extension of the existing zero-production-caller gate for `stage_interaction_request` / `interaction_handoff`, which does not scan for this name). A companion guard pins zero direct-construction points for the now-public `InteractionHandoff` class outside its own defining module.

Three counters (`anchor.absent_no_run`, `anchor.unavailable_dangling_pointer`, `anchor.absent_legacy_checkpoint_type`) are incremented in `interaction_rollout`'s existing counter registry rather than a new one. The corrupt outcome instead registers a new `ops_signals` constant, `INTERACTION_ANCHOR_CORRUPT`, deliberately left without a paired `clear_degradation` call: a corrupt anchor is a property of a persisted row, and no in-process registry can observe that row being fixed.

## Predicate move

`checkpoint_run_partition_filter` moves from a `trace_handlers.py` staticmethod into `trace_event_staging.py`, unchanged, so future services-layer consumers can use it without importing from the api layer. The new anchor resolver is not such a consumer: it checks the same run-partition condition with its own in-Python comparison on the already-fetched row, and its docstring names the predicate as the semantic sibling of that check. The old staticmethod now delegates in one line. Neither file needed a new import statement — both already imported everything the predicate uses.

## Public handoff surface

`_InteractionHandoff` is renamed to `InteractionHandoff`, and `interaction_handoff`'s return annotation changes to match — the object a public context manager yields is now a public type. No test imports the old name, so this changes no test's behavior.

A public `staged: StagedInteractionRequest | None` attribute replaces the private `_staged_row` flag. It is assigned in `stage()` at the exact point `stage_interaction_request` has actually returned a row, and the misuse handler that used to read `_staged_row` now reads this field instead — an equivalent substitution, since `stage_interaction_request`'s return value is never `None`. A caller can now tell whether a handoff round staged a row by reading this field after the `with` block, without consulting the `ops_signals` registry.

`__all__` now lists both entry points, the public type, both dataclasses, and every exception class the module defines.

## PostgreSQL backend assertions

Two facts that previously lived only in code comments now have tests: an over-length locator column raises `DataError` (not `IntegrityError`) on PostgreSQL, and a non-finite float in a JSON payload raises `DataError` (not `StatementError`) once it reaches the database's own parser. Both cells bypass `stage_interaction_request`'s Python-side validation with a direct Core `INSERT`, since that validation is exactly what ordinarily prevents either outcome from reaching SQL at all. Also corrects a comment that named the `request_payload` column's PostgreSQL type as `jsonb`; it compiles to `JSON`.

Both land in the existing `test_interaction_staging_postgresql.py`, reusing its disposable-schema fixture. No CI workflow changes: this file's path is already listed in both `test-migrations.yml`'s trigger paths and its explicit pytest step.

## Notes on two deliberate choices

- The resolver imports `InteractionAnchor` only as `from .task_interaction_staging import InteractionAnchor`. Importing the module itself and referencing `task_interaction_staging.InteractionAnchor` would trip the zero-production-caller gate for `stage_interaction_request` / `interaction_handoff`, which matches on that module-path form.
- The predicate move is a pure position change: same body, same signature, same behavior. It moves because services code must not import from api, and because one run-partition predicate should have one definition.

### DIFF
--- a/src/xagent/web/services/task_interaction_anchor.py
+++ b/src/xagent/web/services/task_interaction_anchor.py
@@ -1,0 +1,227 @@
+"""Resolve one task's interaction anchor from its checkpoint pointer.
+
+This is the "other half" of anchor resolution ``InteractionAnchor``'s own
+docstring (``task_interaction_staging.py``) names: a primary-key read
+against ``trace_events``, performed here entirely outside any savepoint
+either staging primitive opens, before a caller ever builds an anchor to
+hand to ``stage_interaction_request`` or ``interaction_handoff``.
+
+Six-outcome judgment table, evaluated in order -- the order is the contract,
+not an implementation detail (see the corrupt-before-legacy note below):
+
+1. ``task.run_id IS NULL`` -> absence.
+2. ``task.last_checkpoint_trace_event_id IS NULL`` -> absence.
+3. the pointer names no live ``trace_events`` row -> unavailable, reported
+   to the caller as absence (see below for why).
+4. the row exists but fails any of six self-consistency conditions ->
+   corrupt.
+5. the row passes all six conditions but its ``checkpoint_type`` is a
+   legacy type -> absence, not corrupt.
+6. the row passes all six conditions and its ``checkpoint_type`` is the
+   current one -> a resolved ``InteractionAnchor``.
+
+Step 4 must run before step 5: a row belonging to a *different* task, even
+one whose ``checkpoint_type`` is legacy, must still be reported corrupt.
+Folding the two steps into one (checking legacy-type first) would let a
+cross-task row slip through as an ordinary absence instead -- see
+``test_task_interaction_anchor.py``'s mutation test for that ordering.
+
+``run_id IS NULL`` handling deliberately does not match
+``trace_event_staging.checkpoint_run_partition_filter``. That predicate
+treats ``run_id IS NULL`` as a legitimate partition (the root-checkpoint
+read path it was written for can genuinely have no run id yet) and matches
+rows whose own run field is also NULL. This function instead treats
+``task.run_id IS NULL`` as absence outright, at step 1, before any row is
+even read: ``InteractionAnchor.resume_run_partition`` is typed ``str``, and
+``_validate_anchor_fields`` (``task_interaction_staging.py``) raises
+``InteractionAnchorCorrupt`` for an empty string -- protocol v1 has no way
+to represent a NULL partition inside an anchor at all. Do not "align" this
+with the legacy predicate by matching a NULL run field or by reclassifying
+it as corrupt; the two functions are answering different questions about
+NULL on purpose.
+
+Step 3, the dangling pointer, is reported to the caller as absence, not as
+a distinct outcome, and this function does not scan the legacy
+``last_checkpoint_event_id`` column the way the trace-read side's fallback
+does (``_AnchorFallback``, ``trace_handlers.py``) -- unavailable and absence
+get different log levels (``logger.warning`` versus ``logger.info``) and
+only the former is counted, but both return ``None`` and both mean the same
+thing to this function's caller: there is no anchor to stage against.
+Folding them into one legacy-style scan is exactly the shape this function
+is built not to reproduce.
+
+``INTERACTION_RUN_PARTITION_MISMATCH_DEGRADED`` (``ops_signals.py``) is a
+signal owned by ``interaction_handoff``, not by this function: it is
+registered when that context manager swallows
+``InteractionRunPartitionMismatch``, an exception this function never
+raises. It is a pure data-corruption alarm, not an expected degradation
+channel the way the other five signals ``interaction_handoff`` registers
+are -- a resolved anchor's ``resume_run_partition`` always equals the
+``run_id`` this function read it against (step 4's partition-match
+condition enforces that), so a caller that passes this function's own
+output straight through to ``stage_interaction_request`` should see that
+signal at an expected frequency of zero in normal operation. A nonzero rate
+means the anchor a caller staged did not come from this function, or the
+run identity changed between resolution and staging.
+
+Zero production callers as of this module's introduction: a static test
+(``tests/web/services/test_task_interaction_anchor.py``) asserts that no
+production module calls ``resolve_interaction_anchor``, mirroring the
+existing gate for ``task_interaction_staging.py``'s two entry points
+without extending that gate itself -- this module is not one of the two
+names it scans for. See that test's own docstring for the removal
+condition.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+from sqlalchemy.orm import Session
+
+from ...core.agent.checkpoint import (
+    LEGACY_CHECKPOINT_TYPES,
+    READABLE_CHECKPOINT_TYPES,
+    checkpoint_execution_id,
+)
+from ..models.task import Task
+from ..models.task import TraceEvent as DatabaseTraceEvent
+from .interaction_rollout import (
+    COUNTER_ANCHOR_ABSENT_LEGACY_CHECKPOINT_TYPE,
+    COUNTER_ANCHOR_ABSENT_NO_RUN,
+    COUNTER_ANCHOR_UNAVAILABLE_DANGLING_POINTER,
+    increment_counter,
+)
+from .ops_signals import INTERACTION_ANCHOR_CORRUPT, register_degradation
+from .task_interaction_staging import InteractionAnchor
+from .task_lease_service import TASK_RUN_ID_TRACE_FIELD
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_interaction_anchor(db: Session, task: Task) -> InteractionAnchor | None:
+    """Resolve ``task``'s interaction anchor from its checkpoint pointer, or
+    ``None`` if none applies -- see the module docstring for the six-outcome
+    table this function implements, in the order it must run.
+
+    Six self-consistency conditions are checked once a pointed-to row is
+    found (step 4); any one failing classifies the row as corrupt, copied
+    verbatim from ``_load_pk_anchored_checkpoint``'s own six conditions
+    (``trace_handlers.py``) rather than re-derived:
+
+    1. ``row.task_id != task.id`` -- ownership.
+    2. ``row.event_type != "system_update_general"`` -- not a checkpoint
+       row's event type.
+    3. ``row.build_id is not None`` -- must not be a build-partition row.
+    4. ``row_data.get("checkpoint_type") not in READABLE_CHECKPOINT_TYPES``
+       -- not a checkpoint row at all.
+    5. the row's run field does not match ``task.run_id`` -- the row
+       belongs to another run. Checked in Python against the fetched row's
+       own data, not via ``trace_event_staging.checkpoint_run_partition_filter``
+       (that predicate compiles to SQL for a query's ``WHERE`` clause; this
+       function already has the one candidate row in hand), but the same
+       equality semantics.
+    6. ``row_execution_id and row_execution_id != execution_id`` --
+       execution identity mismatch. An empty ``row_execution_id`` (a legacy
+       row with no identity field) short-circuits past this condition on
+       purpose -- this leniency is copied from the source condition, not a
+       gap to close.
+
+    ``InteractionAnchor`` carries no task id (see its own field list in
+    ``task_interaction_staging.py``). Once this function returns an anchor,
+    nothing downstream re-checks which task the anchored ``trace_events``
+    row belongs to -- ``_validate_anchor_fields`` only checks the dataclass
+    against itself, and the staging primitive's run-partition comparison is
+    about ``run_id``, not ownership. **This function is therefore the
+    single execution point of the ownership check.** Condition 1 below
+    (``row.task_id != task.id``) is not a redundant sanity assert; it is
+    the only one there is.
+
+    The anchor's ``resume_execution_id`` and the condition-6 ``execution_id``
+    it is checked against are both ``str(task.id)``: this function takes no
+    separate execution-identity argument the way
+    ``_load_pk_anchored_checkpoint`` does, because the web read path that
+    function serves passes it the task id already (see that function's own
+    docstring: "web's execution_id is the task id"). Using anything else
+    here (e.g. the row's own, possibly-empty ``row_execution_id``) could
+    persist an empty string into a column
+    (``ck_task_interaction_requests_resume_execution_id_nonempty``,
+    ``models/task_interaction.py``) that must never be empty.
+    """
+
+    if task.run_id is None:
+        logger.info("task %s has no run_id; no interaction anchor to resolve", task.id)
+        increment_counter(COUNTER_ANCHOR_ABSENT_NO_RUN)
+        return None
+
+    pointer_id = task.last_checkpoint_trace_event_id
+    if pointer_id is None:
+        logger.info(
+            "task %s has no checkpoint pointer; no interaction anchor to resolve",
+            task.id,
+        )
+        return None
+
+    row = db.get(DatabaseTraceEvent, pointer_id)
+    if row is None:
+        logger.warning(
+            "task %s's checkpoint pointer %s has no matching trace_events row",
+            task.id,
+            pointer_id,
+        )
+        increment_counter(COUNTER_ANCHOR_UNAVAILABLE_DANGLING_POINTER)
+        return None
+
+    row_data: dict[str, Any] = row.data if isinstance(row.data, dict) else {}
+    run_field = row_data.get(TASK_RUN_ID_TRACE_FIELD)
+    partition_matches = run_field == task.run_id
+    execution_id = str(task.id)
+    row_execution_id = checkpoint_execution_id(row_data)
+    if (
+        row.task_id != task.id
+        or row.event_type != "system_update_general"
+        or row.build_id is not None
+        or row_data.get("checkpoint_type") not in READABLE_CHECKPOINT_TYPES
+        or not partition_matches
+        or (row_execution_id and row_execution_id != execution_id)
+    ):
+        register_degradation(
+            INTERACTION_ANCHOR_CORRUPT,
+            f"task {task.id}: checkpoint pointer {pointer_id} does not "
+            "match the row it anchors",
+        )
+        logger.error(
+            "task %s's checkpoint pointer %s failed anchor validation",
+            task.id,
+            pointer_id,
+        )
+        return None
+
+    # Every row that survives the six conditions above has a
+    # checkpoint_type in READABLE_CHECKPOINT_TYPES (condition 4), which is
+    # exactly {CHECKPOINT_TYPE} | LEGACY_CHECKPOINT_TYPES -- so the branch
+    # below is exhaustive between steps 5 and 6, with nothing left over.
+    checkpoint_type = row_data.get("checkpoint_type")
+    if checkpoint_type in LEGACY_CHECKPOINT_TYPES:
+        logger.info(
+            "task %s's checkpoint pointer %s names a legacy checkpoint "
+            "type %r; no interaction anchor to resolve",
+            task.id,
+            pointer_id,
+            checkpoint_type,
+        )
+        increment_counter(COUNTER_ANCHOR_ABSENT_LEGACY_CHECKPOINT_TYPE)
+        return None
+
+    logger.debug(
+        "task %s's checkpoint pointer %s resolved to an interaction anchor",
+        task.id,
+        pointer_id,
+    )
+    return InteractionAnchor(
+        trace_event_id=int(row.id),
+        resume_event_id=str(row.event_id),
+        resume_execution_id=execution_id,
+        resume_run_partition=task.run_id,
+    )
--- a/src/xagent/web/services/task_interaction_anchor.py
+++ b/src/xagent/web/services/task_interaction_anchor.py
@@ -71,6 +71,29 @@ existing gate for ``task_interaction_staging.py``'s two entry points
 without extending that gate itself -- this module is not one of the two
 names it scans for. See that test's own docstring for the removal
 condition.
+
+Two kinds of pre-existing row fail the run-partition self-consistency
+check today and classify as corrupt rather than as absence, even though
+neither one represents actual data corruption. The first is a row whose
+checkpoint pointer was filled in by the migration that added
+``last_checkpoint_trace_event_id`` (2026-08): that migration backfills the
+pointer by matching a task's legacy event-id column against an existing
+``trace_events`` row, and that row was written before the run-partition
+field existed, so it never carries one. The second is any legacy-type
+checkpoint row, on any task, because the function that writes this field
+(``stage_trace_event_row``, ``trace_event_staging.py``) only ever writes
+it for current-type checkpoints -- a legacy-type row cannot carry the
+field regardless of when it was written. Both kinds of row are missing
+the field, fail the partition check before the legacy-type check ever
+runs, and register the corrupt-anchor degradation signal
+(``INTERACTION_ANCHOR_CORRUPT``, ``ops_signals.py``). That signal has no
+clearing point by design, so once one of these rows triggers it, it stays
+active until the process restarts. The change that wires this function's
+first production caller has to decide, before that caller ships, how a
+pre-existing row like this should be classified -- as corrupt, as
+absence, or some other way -- because the classification these rows get
+today is a side effect of when and how they were written, not a decision
+anyone made about what a caller should see for them.
 """
 
 from __future__ import annotations
--- a/src/xagent/web/services/task_interaction_anchor.py
+++ b/src/xagent/web/services/task_interaction_anchor.py
@@ -24,7 +24,8 @@ Step 4 must run before step 5: a row belonging to a *different* task, even
 one whose ``checkpoint_type`` is legacy, must still be reported corrupt.
 Folding the two steps into one (checking legacy-type first) would let a
 cross-task row slip through as an ordinary absence instead -- see
-``test_task_interaction_anchor.py``'s mutation test for that ordering.
+``test_task_interaction_anchor.py``'s cross-task legacy-type cell for that
+ordering.
 
 ``run_id IS NULL`` handling deliberately does not match
 ``trace_event_staging.checkpoint_run_partition_filter``. That predicate
@@ -78,8 +79,8 @@ neither one represents actual data corruption. The first is a row whose
 checkpoint pointer was filled in by the migration that added
 ``last_checkpoint_trace_event_id`` (2026-08): that migration backfills the
 pointer by matching a task's legacy event-id column against an existing
-``trace_events`` row, and that row was written before the run-partition
-field existed, so it never carries one. The second is any legacy-type
+``trace_events`` row, and if that row predates the run-partition field
+(added before this migration), it carries none. The second is any legacy-type
 checkpoint row, on any task, because the function that writes this field
 (``stage_trace_event_row``, ``trace_event_staging.py``) only ever writes
 it for current-type checkpoints -- a legacy-type row cannot carry the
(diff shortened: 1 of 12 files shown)


src/xagent/web/services/task_interaction_anchor.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+        increment_counter(COUNTER_ANCHOR_UNAVAILABLE_DANGLING_POINTER)
+        return None
+
+    row_data: dict[str, Any] = row.data if isinstance(row.data, dict) else {}
+    run_field = row_data.get(TASK_RUN_ID_TRACE_FIELD)
+    partition_matches = run_field == task.run_id
+    execution_id = str(task.id)
+    row_execution_id = checkpoint_execution_id(row_data)
+    if (
```
rogercloud, line 205: **Moderate — duplication contradicting the PR's own stated principle.**

These six self-consistency conditions are copied verbatim from `_load_pk_anchored_checkpoint` in `src/xagent/web/api/trace_handlers.py:573-580`. The module docstring says this is deliberate ("copied verbatim... rather than re-derived"), so it isn't an oversight — but no test pins the two copies as equivalent, so a future edit to one can silently diverge from the other with nothing failing.

This also sits awkwardly against the rationale for moving `checkpoint_run_partition_filter` in this same PR: "one predicate should have one definition" is applied to one condition but not to the five right next to it. The PR description discusses only the partition-predicate move, not this larger duplication.

Suggested fix: extract `checkpoint_anchor_row_is_consistent(row, row_data, *, task_id, partition_matches, execution_id) -> bool` into `trace_event_staging.py`, call it from both `task_interaction_anchor.py` and `trace_handlers.py`, and add an equivalence/delegation test analogous to the one already added for the partition predicate.

src/xagent/web/services/task_interaction_anchor.py
```diff
@@ -0,0 +1,227 @@
...
+    if row is None:
+        logger.warning(
+            "task %s's checkpoint pointer %s has no matching trace_events row",
+            task.id,
+            pointer_id,
+        )
+        increment_counter(COUNTER_ANCHOR_UNAVAILABLE_DANGLING_POINTER)
+        return None
+
+    row_data: dict[str, Any] = row.data if isinstance(row.data, dict) else {}
```
rogercloud, line 200: **Minor — untested defensive guard.**

No test exercises `row.data` as a non-dict (list or `None`), so this `isinstance(row.data, dict)` guard is plausible but unverified. Lowest-priority of the coverage gaps; one parametrize case would cover it.

src/xagent/web/services/trace_event_staging.py
```diff
@@ -156,5 +156,33 @@ def stage_trace_event_row(
                 trace_event_id=row_id,
             ),
```
rogercloud, line 157: **Cosmetic.** This diff removes the blank line that separated the early-return checkpoint block from the fallthrough `return StagedTraceRow(row_id=None, ...)`. Unrelated to the PR's stated purpose — worth restoring to keep the diff focused.

tests/web/services/test_task_interaction_anchor.py
```diff
@@ -0,0 +1,491 @@
...
+
+
+def test_ta6b_cross_task_row_legacy_type_is_still_corrupt(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """Differs from T-A-6a in exactly one field: this row's owning task_id
+    is the *other* task's, same as 6a, but here checkpoint_type is legacy.
+    If step 5 (legacy -> absence) ran before step 4 (the six conditions,
+    including ownership), this row would wrongly classify as absence
+    instead of corrupt -- see T-A-9 for the mutation that pins this
```
rogercloud: **Minor — dangling reference.** This comment points at "T-A-9" ("see T-A-9 for the mutation that pins this ordering directly"), but no `test_ta8` or `test_ta9` exists in the suite. The ordering guarantee is actually covered by `test_ta6b` just above. Either add the referenced mutation test or repoint the comment at `test_ta6b`.

(The docstring's "nine judgment-table cells" count is accurate — `test_ta6` is split into `test_ta6a`/`test_ta6b` — so only the reference itself is stale.)

tests/web/services/test_task_interaction_anchor.py
```diff
@@ -0,0 +1,491 @@
...
+
+
+def _scan_root(exclude_stem: str) -> list[Path]:
+    root = Path(next(iter(xagent.__path__)))
+    modules = [p for p in root.rglob("*.py") if p.stem != exclude_stem]
+    assert modules, "production scan set is empty"
+    return modules
+
+
+def test_ta11_resolve_interaction_anchor_has_zero_production_callers() -> None:
```
rogercloud, line 593: **Minor — third hand-rolled AST scanner (`test_ta11_resolve_interaction_anchor_has_zero_production_callers`).**

This duplicates the `_scan_root` / module-walking helper verbatim from `tests/web/services/test_interaction_staging_production_gate.py`, whose `GATED_NAMES` mechanism is structurally capable of covering this case. Unlike that gate (which documents four blind spots), this test documents none — notably that only `ast.Call` nodes are matched, so `f = mod.resolve_interaction_anchor; f(db, task)` or `getattr(mod, "resolve_interaction_anchor")(...)` would go undetected. That specific gap also exists undocumented in the original gate, so this is an extension of an existing issue rather than a regression.

Suggested fix: fold this invariant into the existing `GATED_NAMES` gate, and extract one shared `_scan_root` helper for the remaining scanners.

tests/web/services/test_interaction_handoff_surface.py
```diff
@@ -0,0 +1,341 @@
...
+    db.close()
+
+
+# ---------------------------------------------------------------------------
+# T-H-6: __all__ -- every name resolves, none are private, InteractionHandoff
+# is in it.
+# ---------------------------------------------------------------------------
+
+
+def test_th6_all_names_resolve_and_are_public() -> None:
```
rogercloud, line 274: **Minor — the untested direction is completeness.**

Of the three properties asserted here, only "all names resolve" is redundant with lint (ruff F822 is enabled by default and runs in CI via pre-commit). The no-duplicates and no-leading-underscore checks have no lint equivalent and are legitimate test-only coverage, so this test should stay as-is.

The real gap is the reverse direction: nothing checks that every public name the module actually defines appears in `__all__`, and ruff emits no warning when a public function is omitted. Suggest adding a completeness assertion comparing `__all__` against the module's actual public top-level names, on top of the existing checks.

tests/web/services/test_interaction_handoff_surface.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def _scan_root(exclude_stem: str) -> list[Path]:
+    root = Path(next(iter(xagent.__path__)))
+    modules = [p for p in root.rglob("*.py") if p.stem != exclude_stem]
+    assert modules, "production scan set is empty"
+    return modules
+
+
+def test_ta12_interaction_handoff_has_zero_direct_construction_points() -> None:
```
rogercloud, line 320: **Minor — same shape as `test_ta11` (`test_ta12_interaction_handoff_has_zero_direct_construction_points`).**

Another hand-rolled AST guard duplicating the `_scan_root` helper instead of reusing the gate in `tests/web/services/test_interaction_staging_production_gate.py`. Suggest folding it into the same shared gate mechanism proposed for `test_ta11`, and documenting the `ast.Call`-only blind spot.

rogercloud (approved):
# Code Review — PR #1283

## Summary

This PR adds a new services-layer module `src/xagent/web/services/task_interaction_anchor.py` whose `resolve_interaction_anchor` performs a primary-key read against `trace_events` to resolve a task's interaction checkpoint anchor from `task.last_checkpoint_trace_event_id`, classifying the result into six outcomes (no run_id, no pointer, dangling pointer, corrupt row, legacy checkpoint type, resolved). It ships with zero production callers, guarded by a dedicated AST "zero caller" test. Alongside it, `checkpoint_run_partition_filter` moves from `src/xagent/web/api/trace_handlers.py` into `src/xagent/web/services/trace_event_staging.py` (old site becomes a one-line delegate) so services-layer code no longer reverse-imports the api layer, and `_InteractionHandoff` is publicized to `InteractionHandoff` with a new public `staged: StagedInteractionRequest | None` field replacing a private flag plus an explicit `__all__`. Two new PostgreSQL-only tests pin `DataError` behavior for over-length strings and non-finite JSON floats via direct Core INSERT, and roughly 900 lines of tests land across three files (two new).

## Design verdict

**Acceptable with reservations** — the six-outcome classification is correct and internally consistent, but the PR applies its own stated principle unevenly and one outcome branch is structurally unreachable for real data.

- **The PR's stated thesis is applied to one predicate and not to the five conditions beside it.** Moving `checkpoint_run_partition_filter` into the services layer is justified as "one predicate should have one definition," yet the six self-consistency conditions immediately adjacent to that predicate are copy-pasted verbatim into the new module with no shared helper and no drift-detection test. The PR description discusses only the partition-predicate piece and does not surface the larger duplication (Finding 2).
- **Dead-code surface is accumulating.** ~230 production lines plus ~830 test lines with zero callers is a pre-existing staging pattern in this codebase and defensible on its own, but it is now guarded by three near-duplicate hand-rolled AST scanners (Finding 4), and the guard mechanism is being re-implemented per-feature rather than extended.
- **The most concrete consequence of the above is Finding 1**: because the duplicated consistency conditions are evaluated before the legacy-type branch, and because the run-partition field is by construction never present on a genuine legacy-type row, the "legacy checkpoint type → benign absence" outcome cannot be reached by authentic data. The test that appears to cover it passes only because a fixture injects a field real legacy rows never carry. This is the finding to resolve before wiring the first caller.

## Findings

### 1. `src/xagent/web/services/task_interaction_anchor.py` (outcome 5, legacy checkpoint type) — **Moderate**

The "legacy `checkpoint_type` → absence, not corrupt" outcome is structurally unreachable for any authentic legacy-origin row. The `_task_run_id` JSON field on the pointed-to row is written only by `stage_trace_event_row` (`src/xagent/web/services/trace_event_staging.py:108-116`), gated on `data.get("checkpoint_type") == CHECKPOINT_TYPE` — strict equality against the *current* type, never the legacy one. A genuine legacy-type row therefore always has `_task_run_id` absent/`None`. Step 1 of the resolver already guarantees `task.run_id` is a non-empty string by the time any row is evaluated, so `partition_matches = (row_field == task.run_id)` evaluates `None == "<non-empty-string>"` → always `False`, the row fails a self-consistency condition at step 4, and it is classified **CORRUPT** — never reaching step 5. The migration backfill (`20260804_add_task_checkpoint_trace_event_anchor.py`) can plant a pointer at a legacy-type row (it filters on event_id / event_type / build_id, not `checkpoint_type`) but never rewrites that row's JSON payload to add `_task_run_id`, so backfilled rows meet the same fate.

This is a latent false-positive, not merely a dead branch: CORRUPT fires `register_degradation(INTERACTION_ANCHOR_CORRUPT, ...)` plus an error-level log, while legacy-absence is an info log and a benign counter. Once a real caller exists, every genuine pre-existing task with a legacy checkpoint would raise a spurious degradation alarm. The headline design claim ("corrupt-before-legacy ordering matters — a row belonging to a different task must be reported corrupt even when its `checkpoint_type` is legacy") also has no real specimen behind it, since real legacy rows are excluded earlier by the partition-match condition regardless of cross-task-ness; `anchor.absent_legacy_checkpoint_type` is very likely permanently zero for real data. `test_ta5_legacy_type_same_task_is_absence` passes only because `_make_row` unconditionally injects `_task_run_id` irrespective of `checkpoint_type` — a shape real legacy data cannot produce.

**Suggested fix:** before wiring a real caller, either (a) make the partition-match condition legacy-type-aware — skip or substitute it when `checkpoint_type` is legacy, since such rows never carry the run-partition field by construction — or (b) document that the branch is currently unreachable and why, and add a test using an authentic legacy-row shape (no `_task_run_id` key at all) so the true behavior (→ CORRUPT) is visible and intentional rather than masked by an unrealistic fixture.

### 2. `src/xagent/web/services/task_interaction_anchor.py:181-188` — **Moderate**

The six-condition self-consistency check is copied verbatim from `_load_pk_anchored_checkpoint` in `src/xagent/web/api/trace_handlers.py:573-580` rather than extracted into a shared helper. This is deliberate and documented in the module docstring ("copied verbatim… rather than re-derived"), so it is not an oversight — but no test pins the two copies as equivalent, so a future edit to one can silently diverge with nothing failing. It also contradicts the rationale used to justify moving the sibling `checkpoint_run_partition_filter` predicate in this same PR: the "one predicate, one definition" principle is applied to one condition and not to the five sitting next to it.

**Suggested fix:** extract `checkpoint_anchor_row_is_consistent(row, row_data, *, task_id, partition_matches, execution_id) -> bool` into `trace_event_staging.py` and call it from both `task_interaction_anchor.py` and `trace_handlers.py`, plus an equivalence/delegation test analogous to the one already added for the partition-predicate move.

### 3. Test coverage gaps in `tests/web/services/test_task_interaction_anchor.py` — **Minor**

Three related gaps in the resolver's suite:

- **(a)** No test pipes a resolved `InteractionAnchor` into `stage_interaction_request` / `_validate_anchor_fields`, so nothing proves the resolver's output is consumable by the staging pipeline it exists to feed. Most valuable of the three, though reasonably deferred until a first real caller lands.
- **(b)** `row.data` is never exercised as a non-dict (list / `None`), leaving the defensive `isinstance(row.data, dict)` guard at `src/xagent/web/services/task_interaction_anchor.py:176` untested. Lowest priority.
- **(c)** The resolved branch's `execution_id` sub-condition (`row_execution_id and row_execution_id != execution_id`) is never exercised with two non-empty equal values — every passing test uses an empty-string `execution_id`, which short-circuits rather than compares. Cheapest to close: one more parametrize case.

### 4. Three near-identical hand-rolled AST scanners — **Minor**

`test_ta11_resolve_interaction_anchor_has_zero_production_callers` (`tests/web/services/test_task_interaction_anchor.py`) and `test_ta12_interaction_handoff_has_zero_direct_construction_points` (`tests/web/services/test_interaction_handoff_surface.py`) each duplicate the `_scan_root` / module-walking helper verbatim from the pre-existing `test_interaction_staging_production_gate.py`, rather than extending that gate's `GATED_NAMES` mechanism — which is structurally capable of covering at least the ta11 case. Neither new test documents its blind spots, where the existing gate documents four. Note the specific blind spot: only `ast.Call` nodes are matched, so `f = mod.resolve_interaction_anchor; f(db, task)` or `getattr(mod, "resolve_interaction_anchor")(...)` goes undetected — this gap is also present, undocumented, in the original gate, so it is an extension of an existing issue rather than a regression.

**Suggested fix:** extract one shared `_scan_root` test helper and fold ta11 (and ideally ta12) into the existing `GATED_NAMES` gate.

### 5. `tests/web/services/test_interaction_handoff_surface.py:274-286` — **Minor**

`test_th6_all_names_resolve_and_are_public` asserts three properties of `__all__`: no duplicates, no leading underscore, and all names resolve via `hasattr`. Only the last is redundant with lint — ruff's F822 is enabled by default and runs in CI via pre-commit — while the duplicate and underscore checks have no lint equivalent and are legitimate test-only coverage; the test should not be removed. The real gap is the untested direction: nothing checks *completeness*, i.e. that every public name the module actually defines appears in `__all__`, and ruff emits no warning when a public function is omitted.

**Suggested fix:** add a completeness assertion comparing `__all__` against the module's actual public top-level names; keep the existing checks.

### 6. `tests/web/services/test_task_interaction_anchor.py:393` — **Minor**

Dangling comment reference to "T-A-9" ("see T-A-9 for the mutation that pins this ordering directly") — no `test_ta8` or `test_ta9` exists in the suite; the ordering guarantee is actually covered by `test_ta6b` just above the comment. (The docstring's "nine judgment-table cells" count is accurate, since `test_ta6` is split into `test_ta6a`/`test_ta6b` — only the comment reference is wrong.)

**Suggested fix:** either add the referenced mutation test or repoint the comment at `test_ta6b`.

### 7. Observability coverage of the six outcomes — **Minor**

Only three of six outcomes get counters (`absent_no_run`, `unavailable_dangling_pointer`, `absent_legacy_checkpoint_type`). "No pointer" and "resolved" get none — an undocumented gap — while "corrupt" gets only an ops_signals degradation, which is an explicitly documented deliberate choice and not a gap. Net effect: no counter can act as a rate denominator, and the six-outcome table is split across `interaction_rollout.py` (counters) and `ops_signals.py` (signal) with the no-pointer-vs-other-absence asymmetry left unexplained in the docstring. Low priority while the resolver has zero callers, but worth resolving alongside the first real caller.

### 8. `src/xagent/web/services/trace_event_staging.py:157-158` — **Cosmetic**

The diff removes the blank line separating the early-return checkpoint block from the fallthrough `return StagedTraceRow(row_id=None, ...)`. Unrelated to the PR's purpose; restore it.

- `src/xagent/web/services/task_interaction_anchor.py:181-188`: **shrink** — six-condition corrupt check duplicated verbatim from `src/xagent/web/api/trace_handlers.py:573-580`. Extract one shared `checkpoint_anchor_row_is_consistent(...)` helper both sites call; add an equivalence test.
- `tests/web/services/test_task_interaction_anchor.py` / `tests/web/services/test_interaction_handoff_surface.py` (`_scan_root` helper, ~3 near-identical copies including the pre-existing gate): **shrink** — duplicated module-tree-walking helper across three test files. Extract one shared test helper.
- `tests/web/services/test_task_interaction_anchor.py` (`test_ta11_...zero_production_callers`): **yagni** — third hand-rolled AST "zero caller" scanner for an invariant that gets deleted on the first real caller. Extend the existing `GATED_NAMES` gate in `test_interaction_staging_production_gate.py` instead.
- `tests/web/services/test_interaction_handoff_surface.py` (`test_ta12_...zero_direct_construction_points`): **yagni** — second/third hand-rolled AST guard of the same shape. Fold into the same shared gate mechanism as ta11.

**net: -20 lines possible** — roughly 10-15 lines from collapsing the two new `_scan_root` copies into one shared helper (and folding ta11/ta12 into the existing gate rather than re-implementing the walk), plus 6-8 lines from replacing the eight-line duplicated six-condition block with a single call to a shared helper. Both are net reductions even after accounting for the new helper definitions and the added equivalence test, since the helpers replace more duplicated lines than they introduce.

**Optional low-priority note (not a finding):** the run-partition comparison could be written as `str(run_field) == task.run_id if run_field is not None else False` to make SQL/Python equality equivalence type-safe against a hypothetical future non-string write. Every current write path casts via `str(...)`, so the docstring's "same equality semantics" claim is accurate today and there is no evidence such a path exists or is planned.

src/xagent/web/services/task_interaction_anchor.py
AlexLiu190625: Keeping the copies, for a reason specific to this pair rather than a general preference.

The two blocks are not two implementations of one predicate: `_load_pk_anchored_checkpoint` computes `partition_matches` as `run_field == run_id if run_id is not None else run_field is None` — a NULL `run_id` is a legitimate partition there — while this resolver rejects a NULL `run_id` at step 1 and computes `run_field == task.run_id` unconditionally. That divergence is documented on both sides and is deliberate: the read path this resolver serves cannot treat an unpartitioned row as anchorable. Extracting a helper that takes `partition_matches` as a parameter would share the six conditions while leaving the interesting difference outside the helper, which is the part most likely to drift.

The second reason is placement. `trace_event_staging.py` is a write-side module whose docstring scopes it to staging a trace row inside a caller-owned session; `checkpoint_run_partition_filter` fit there because that module already imported every name the predicate needed. A row-validity helper would additionally pull in `READABLE_CHECKPOINT_TYPES` and `checkpoint_execution_id`, turning a write-side module into a general utility drawer.

On the substantive half of the point — nothing pins the two copies as equivalent — you are right, and that is worth closing. An AST-level equivalence assertion, in the same shape as the delegation test this PR already adds for the partition predicate, gets that without moving production code; it is filed as a follow-up alongside the first production caller rather than added here, since the caller change is what makes divergence consequential.

src/xagent/web/services/task_interaction_anchor.py
AlexLiu190625: Covered in 1991de28a. A new cell builds the pointed-to row with `data=["not", "a", "dict"]` and pins the outcome the guard actually produces: the payload coerces to `{}`, the empty dict then fails the readable-checkpoint-type membership check, and the row classifies corrupt. Inverting the guard does not merely change the classification — the resolver raises `AttributeError: 'list' object has no attribute 'get'` — so the cell pins a real branch.

A `None` payload is not covered and cannot be: `trace_events.data` is `nullable=False`, so such a row fails to commit before the resolver sees it. The test docstring records that.

src/xagent/web/services/trace_event_staging.py
AlexLiu190625: Restored in 6ffded17e — the diff for this file is now confined to the added predicate.

tests/web/services/test_task_interaction_anchor.py
AlexLiu190625: Removed in 6ffded17e. The sentence pointed at a name with no referent in this repository, so neither adding a test under that name nor repointing at the cell above it was right — the ordering it described is covered by the cross-task legacy-type cell, and the note above the resolver's step ordering now points there instead. The surrounding coverage notes in both test files were rewritten in the same commit for the same reason.

tests/web/services/test_task_interaction_anchor.py
AlexLiu190625: Two corrections and a partial adoption.

Folding this into `GATED_NAMES` loses coverage, measured rather than reasoned: `_production_uses` gates its `ImportFrom` branch on `(node.module or "").split(".")[-1] == PRIMITIVE_MODULE`, a single hard-coded module stem. Adding a name to `GATED_NAMES` therefore catches call shapes but not import-only or star-import shapes. Probed against five shapes, the extended gate finds nothing for `import-only`, `star import`, and `bare module import`, while this test's own module-stem check catches all of them. Generalizing the gate from "one module plus a set of names" to "a module-to-names mapping" would fix that, but it is a larger change to a shared guard than this PR should carry.

On the blind spot: the existing gate does document it. Its module docstring lists four blind spots, and the two shapes named here — `getattr(mod, "...")(...)` and rebinding to a local before calling — fall under items (a) and (b) there. The gap is that these newer guards document none of theirs, which is fair; adding those notes is filed as a follow-up together with extracting one shared tree-walking helper for the duplicated copies.

One factual note on the shared-helper arithmetic: extracting it lands net-positive on line count, not negative. The two verbatim copies are six lines each; a shared helper in its own test-support module, with the docstring the other guards set a precedent for, costs more than the twelve lines it removes.

tests/web/services/test_interaction_handoff_surface.py
AlexLiu190625: Adding the completeness direction, but not in this change, because a naive assertion is red on arrival.

The module defines one public top-level name that is intentionally not exported: `logger`. A completeness assertion comparing `__all__` against the module's actual public top-level names fails today on exactly that name, so landing it requires either an exemption inside the assertion — which weakens the thing being asserted — or renaming `logger` to `_logger` across its call sites in `task_interaction_staging.py`, which is a wider edit than a test assertion warrants here.

The direction is right and the observation that ruff has no rule for it is correct: F822 catches names in `__all__` that do not resolve, and nothing catches a public definition missing from `__all__`. Filed as a follow-up with the `logger` tradeoff recorded so whoever lands it makes that call deliberately. The three existing assertions stay as-is.

tests/web/services/test_interaction_handoff_surface.py
AlexLiu190625: Same disposition as the sibling guard: not folded into `GATED_NAMES`, for the coverage reason measured there, and the shared tree-walking helper plus blind-spot notes are filed as a follow-up. One shape-level difference worth stating — this guard counts class construction sites, not function call sites, so it would need its own node handling inside that gate rather than an extra entry in a name set.

AlexLiu190625:
On the unreachable legacy-type outcome — confirmed, with two corrections to the mechanism and a wider blast radius than the finding states. Addressed in 1991de28a (head now `1991de28a`).

**The outcome is unreachable, and the fixture was the reason it looked covered.** Probing the resolver directly with an authentic legacy-type row (no `_task_run_id` key at all) against the same task returns `None` and registers `interaction_anchor_corrupt`; the same row with the field injected returns `None` and increments `anchor.absent_legacy_checkpoint_type` instead. `_make_row` wrote the field unconditionally, so the old cell only ever saw the second shape.

**First correction — the path you describe is intercepted even earlier.** `tasks.run_id` has exactly two writers, and both clear the checkpoint pointer in the same statement: the new-turn path nulls `last_checkpoint_event_id` and `last_checkpoint_trace_event_id` alongside the write, and the lease path either nulls them on a new run or routes through a CASE whose clear predicate reads the pre-update row. So a task that has a non-empty `run_id` today does not still carry a pointer to a row written under an older run — such a task fails at step 1 (no run) or step 2 (no pointer) rather than reaching the partition check.

**Second correction — the reachable misclassification is a different, larger class.** Rows written before the run-partition field existed (added 2026-07-26) carry no `_task_run_id` regardless of checkpoint type, and the 2026-08-04 backfill migration matches on event id, event type and build id — never on `checkpoint_type` or payload contents — so it can plant a pointer at one. Those rows classify corrupt for the same reason, and there are more of them than legacy-type rows. `interaction_anchor_corrupt` has no clearing point by design, so a single such row keeps the signal active in `/health` for the life of the process.

**Fix taken: option (b), not option (a).** Making the partition condition legacy-aware would leave the six conditions here and the six in `_load_pk_anchored_checkpoint` deliberately divergent, and probing it showed it only changes the classification of shapes that cannot occur (a same-task legacy row whose run partition mismatches — legacy rows carry no partition field at all), while leaving the pre-field class untouched. So instead:

- the legacy cell now uses the authentic shape and asserts the real outcome (corrupt, with the degradation registered) — it fails if the field is re-injected, so it discriminates the two shapes rather than describing one;
- a second cell keeps the field-carrying shape and states in place that no current writer produces it, pinning the outcome's contract rather than a reachable path;
- `resolve_interaction_anchor`'s docstring now names both classes of pre-existing row, the no-clearing-point property of the signal, and the decision the first production caller has to make before it ships.

**On the headline claim you flag:** you are right that `absent_legacy_checkpoint_type` is very likely permanently zero for real data, and the cross-task ordering example has no real specimen — cross-task legacy rows are excluded by the ownership condition, not by legacy-ness. The ordering itself is still worth keeping (a cross-task row must not read as ordinary absence) and the cross-task cell pins it; what changed is that the docstring no longer implies real legacy rows travel that path.

The classification of pre-existing rows is now a stated precondition on wiring the first caller rather than an implicit consequence of write history.

Issue #1284, opened 2026-08-11
fix(dag): returned status and persisted status can disagree when interrupted and waiting steps complete together


AlexLiu190625:
## Current behavior

When two concurrently running DAG steps finish in the same `asyncio.wait` wake-up — one returning an interrupted result and the other entering `waiting_for_user` — the pattern's returned result and its persisted state can disagree. `DAGPattern.status` is written by each step's coroutine as it completes, so the last writer wins: if the waiting step's coroutine finishes restoring state after the interrupted one, the run returns a result whose `status` is `"interrupted"` while `pattern.status` and the checkpoint written for that turn both say `"waiting_for_user"`.

Probe output on current `main` (two-step plan, one step interrupted, one step waiting, forced completion order):

```
returned result status : interrupted
pattern.status         : waiting_for_user
checkpoint status      : waiting_for_user  (state payload agrees)
```

## Why it matters

The caller acts on the returned result (an interrupted run releases its lease as paused), while a later resume restores from the checkpoint — which claims the run was waiting for a user answer. The two disagreeing records describe different lifecycle situations, and which one is persisted depends on coroutine completion order rather than on any rule.

## Expected

The status recorded in the pattern and its checkpoint should agree with the status the run actually returns. The recent determinism change for concurrent completions ordered which *result* wins; this issue covers aligning `self.status` and the persisted state with that same winner.

## Scope note

Pre-existing behavior — reproducible identically on the commit before the concurrent-completion ordering change; that change neither introduced nor widened it.

AlexLiu190625:
Another instance of the returned-state/persisted-state family, on both waiting-task answer paths (v1 `task_reply.py` and a2a's resume): the answer text is written to `Task.input` (and `Task.output` cleared) before the resume reservation, and the post-claim rejection path restores lease/state fields but neither of those two — a rejected answer stays visible as the task's latest input until the next successful turn overwrites it. Mirrored behavior in both paths; noted during review of #1308.

PR #1285, opened 2026-08-11
fix(dag): resolve concurrent waiting steps deterministically

AlexLiu190625:
## Summary

When a batch of concurrently running DAG plan steps wakes up and more than
one step has reached `waiting_for_user` in that same wakeup, the DAG
pattern picks whichever result Python's `set` iteration happens to visit
first. Whichever step is picked determines which question the user's next
reply is delivered to -- the choice is not tied to which step actually
asked first, and the other step's pending question is left with no
delivery path at all.

This changes the batch-completion handling in `DAGPattern` to collect
every result produced by a wakeup, choose a winner deterministically, and
explicitly invalidate every losing waiting step: its active-step
bookkeeping is cleared, its status is set to `clarification_invalidated`,
and its id is surfaced under a new `clarification_superseded_step_ids`
key on the winning result. The full precedence is: a failed step in the
batch outranks every completed result -- the failure check now runs
before a winner is chosen, where previously whichever completed result
the loop happened to visit first was returned before the failure check
could run -- so a batch that is doomed anyway reports the failure rather
than a question or an interrupt. Among completed results, an interrupted
result outranks a waiting one, and among same-rank results the
lexicographically smaller step id wins.

Ranking an interrupt above a live question is a deliberate tradeoff: the
losing step's already-asked question is discarded and the step restarts
later from scratch, because an interrupt is a control instruction the
user issued and should not queue behind a workflow question.

This determinism holds within a single wakeup batch; which steps land in
the same batch remains scheduling-dependent, so ordering across wakeup
boundaries is unchanged.

A losing step's prior work is not compensated: whatever tool calls it
already made and whatever message it already sent to the user before
being invalidated stand as-is. This does not retract the losing step's
question from the conversation history -- it may already have been
delivered as a chat message before invalidation happened. What this
changes is which step's reply-in-flight gets delivered where, not whether
a duplicate question was ever visible.

Once the user answers the winning step's question, the invalidated step
is offered as ready work again like any other pending step, and its next
attempt starts over from the top of the step rather than resuming.
Because that attempt begins from a fresh context, tool calls the step
already executed before losing the tie-break run again -- for
non-idempotent tools (file writes, outbound API calls, messages) that
means duplicated side effects. Nothing memoizes or checkpoints completed
tool calls today; a step that loses the tie-break trades its stranded
question for a rerun, and the rerun is a full one. That attempt can ask
the same kind of question again, in the same turn the winner was just
answered in -- that is a new, honest question belonging to that step, not
the same one being delivered twice.

`clarification_superseded_step_ids` is a plain list of strings, so it is
JSON-safe on its own. It travels inside the same result dict as
everything else a step execution produces, though, so if any other value
in that dict is not serializable, the whole dict collapses into an error
payload together -- this field being safe in isolation does not guarantee
it survives to reach a trace. Nothing reads this key yet: it is
forward groundwork for a consumer on the clarification-resolution path,
and nothing user-facing surfaces "your other question was superseded"
today.

A task is only ever allowed one active question outstanding against the
user at a time. That is enforced as a persistence-layer constraint
elsewhere; this is its application-layer counterpart -- the same
invariant restated as "exactly one step's question wins a given wakeup"
at the point the DAG decides which step gets to keep its question.

## Other changes

- `DAGPattern` now attributes each waiting draft to the step that produced
  it (`with_origin_step`) at the point the draft leaves the step -- ReAct
  itself has no notion of which DAG step it is running under, so this can
  only happen on the DAG side.
- `DAGPattern.has_live_step_tasks()` exposes whether a concurrent step
  batch is still running; `AgentRunner` asserts this is false immediately
  before any `waiting_for_user` exit that carries a clarification draft,
  and raises rather than returning if a step task is still live -- that
  would mean the pattern handed back an answerable question while another
  step could still go on mutating shared state behind it.
- The execution adapter now promotes `clarification_superseded_step_ids`
  to the top level of its normalized `waiting_for_user` output, defaulting
  to an empty list so callers only need one truthiness check instead of
  also handling an absent key.

### DIFF
--- a/src/xagent/core/agent/execution_adapter.py
+++ b/src/xagent/core/agent/execution_adapter.py
@@ -414,6 +414,18 @@ class AgentExecutionAdapter:
                     ),
                 }
             )
+        if status == "interrupted":
+            # A losing waiting step can still be superseded in a batch whose
+            # winner is an interrupt rather than a question (the DAG ranks
+            # an interrupt ahead of a waiting result within the same
+            # wakeup), so this key must reach the top level here too --
+            # otherwise a reader has no way to tell "no sibling was
+            # superseded" apart from "this status never carries the key".
+            # Same empty-list default as the waiting branch above, for the
+            # same reason: one ``if superseded:`` check covers both.
+            normalized["clarification_superseded_step_ids"] = (
+                result.get("clarification_superseded_step_ids") or []
+            )
         return normalized
 
     def _latest_assistant_message(self, context: Any) -> str | None:
--- a/src/xagent/core/agent/execution_adapter.py
+++ b/src/xagent/core/agent/execution_adapter.py
@@ -405,6 +405,13 @@ class AgentExecutionAdapter:
                     # carry the same draft too, but callers should not dig
                     # it out from there.
                     "clarification_draft": result.get("clarification_draft"),
+                    # Empty list rather than ``None`` so a reader only ever
+                    # needs one check (``if superseded:``) instead of also
+                    # distinguishing "key absent" from "key present but
+                    # empty".
+                    "clarification_superseded_step_ids": (
+                        result.get("clarification_superseded_step_ids") or []
+                    ),
                 }
             )
         return normalized
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -693,6 +693,18 @@ class DAGPattern(AgentPattern):
         scheduled_step_ids = set(steps_by_id)
         self._live_step_tasks = set(pending)
 
+        async def cancel_all() -> None:
+            # `pending` is rebound (not mutated) by the `asyncio.wait()`
+            # call inside the loop below; this closure reads the free
+            # variable from the enclosing scope, so it always sees whatever
+            # `pending` is currently bound to -- same mechanism
+            # `schedule_ready_steps()` already relies on.
+            await self._cancel_pending_steps(
+                pending,
+                step_ids_by_task=tasks,
+                steps_by_id=steps_by_id,
+            )
+
         def schedule_ready_steps() -> None:
             if self.plan is None:
                 return
@@ -757,11 +769,7 @@ class DAGPattern(AgentPattern):
                     None,
                 )
                 if failed_step is not None:
-                    await self._cancel_pending_steps(
-                        pending,
-                        step_ids_by_task=tasks,
-                        steps_by_id=steps_by_id,
-                    )
+                    await cancel_all()
                     return await self._fail(
                         context=root_context,
                         runtime=runtime,
@@ -772,21 +780,13 @@ class DAGPattern(AgentPattern):
                     )
 
                 if completed_results:
-                    # Deterministic winner selection: an interrupted result
-                    # always outranks a waiting one (an interrupt is a
-                    # control instruction, not a workflow question the user
-                    # can be asked to sit through), and within the same kind
-                    # the lexicographically first step id wins.
-                    completed_results.sort(
-                        key=lambda item: (
-                            0 if item[1].get("status") != "waiting_for_user" else 1,
-                            tasks[item[0]],
-                        )
+                    winner_task, winner_result = self._select_winner(
+                        completed_results,
+                        step_ids_by_task=tasks,
                     )
-                    winner_task, winner_result = completed_results[0]
                     winner_step_id = tasks[winner_task]
 
-                    # This loop invalidates losers drawn from `done`; the
+                    # This invalidates losers drawn from `done`; the
                     # cancellation below acts on `pending`. `asyncio.wait`
                     # partitions the tasks it was given into exactly these
                     # two sets, so a task can never be in both -- the two
@@ -794,23 +794,11 @@ class DAGPattern(AgentPattern):
                     # this one before or after the cancellation below
                     # cannot change which steps end up invalidated versus
                     # cancelled.
-                    superseded_step_ids: list[str] = []
-                    for loser_task, loser_result in completed_results[1:]:
-                        if loser_result.get("status") != "waiting_for_user":
-                            continue
-                        loser_step_id = tasks[loser_task]
-                        self._clear_active_step(loser_step_id)
-                        loser_step = steps_by_id.get(loser_step_id)
-                        if loser_step is not None:
-                            # Non-terminal, and not picked up by the same-
-                            # wakeup backfill in schedule_ready_steps():
-                            # _ready_steps() offers this step again once
-                            # active_step_ids drains, but
-                            # _pending_ready_steps() only collects steps
-                            # already at "pending", so it waits for the next
-                            # wakeup's readiness check before it reruns.
-                            loser_step.status = "clarification_invalidated"
-                        superseded_step_ids.append(loser_step_id)
+                    superseded_step_ids = self._invalidate_batch_siblings(
+                        completed_results[1:],
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
 
                     if superseded_step_ids:
                         logger.error(
@@ -832,11 +820,7 @@ class DAGPattern(AgentPattern):
                             "clarification_superseded_step_ids": superseded_step_ids,
                         }
 
-                    await self._cancel_pending_steps(
-                        pending,
-                        step_ids_by_task=tasks,
-                        steps_by_id=steps_by_id,
-                    )
+                    await cancel_all()
                     if pending or superseded_step_ids:
                         await runtime.checkpoint(
                             "dag_after_cancelled_siblings",
@@ -856,27 +840,15 @@ class DAGPattern(AgentPattern):
                     return winner_result
 
                 if self._needs_replan(root_context):
-                    await self._cancel_pending_steps(
-                        pending,
-                        step_ids_by_task=tasks,
-                        steps_by_id=steps_by_id,
-                    )
+                    await cancel_all()
                     return None
                 if self.status in {"interrupted", "waiting_for_user"}:
-                    await self._cancel_pending_steps(
-                        pending,
-                        step_ids_by_task=tasks,
-                        steps_by_id=steps_by_id,
-                    )
+                    await cancel_all()
                     return None
                 schedule_ready_steps()
         except BaseException:
             try:
-                await self._cancel_pending_steps(
-                    pending,
-                    step_ids_by_task=tasks,
-                    steps_by_id=steps_by_id,
-                )
+                await cancel_all()
             except Exception:
                 logger.exception("Failed to clean up cancelled DAG sibling steps.")
             raise
@@ -1937,6 +1909,64 @@ class DAGPattern(AgentPattern):
         if getattr(tool, "name", None):
             return str(tool.name)
         return str(tool)
+
+    def _select_winner(
+        self,
+        completed_results: list[tuple[asyncio.Task[Any], dict[str, Any]]],
+        *,
+        step_ids_by_task: dict[asyncio.Task[Any], str],
+    ) -> tuple[asyncio.Task[Any], dict[str, Any]]:
+        """Pick the one result a same-wakeup completed batch delivers.
+
+        Sorts ``completed_results`` in place and returns its new first
+        entry: an interrupted result always outranks a waiting one (an
+        interrupt is a control instruction, not a workflow question the
+        user can be asked to sit through), and within the same kind the
+        lexicographically first step id wins. Callers needing the losers
+        read ``completed_results[1:]`` after this call.
+        """
+        completed_results.sort(
+            key=lambda item: (
+                0 if item[1].get("status") != "waiting_for_user" else 1,
+                step_ids_by_task[item[0]],
+            )
+        )
+        return completed_results[0]
+
+    def _invalidate_batch_siblings(
+        self,
+        siblings: list[tuple[asyncio.Task[Any], dict[str, Any]]],
+        *,
+        step_ids_by_task: dict[asyncio.Task[Any], str],
+        steps_by_id: dict[str, PlanStep],
+    ) -> list[str]:
+        """Clear active-step bookkeeping and flip status on every
+        completed sibling result passed in that is still holding a live
+        "waiting_for_user" result -- that status is reserved for a step
+        whose question was live and never reached the user. Any other
+        sibling (e.g. one that lost an id tie-break between two
+        "interrupted" results) is left completely alone, bookkeeping
+        included: this is byte-for-byte the inline loop this method was
+        extracted from, just given a name. Returns the ids that got the
+        clarification_invalidated flip, in input order.
+        """
+        invalidated_step_ids: list[str] = []
+        for sibling_task, sibling_result in siblings:
+            if sibling_result.get("status") != "waiting_for_user":
+                continue
+            sibling_step_id = step_ids_by_task[sibling_task]
+            self._clear_active_step(sibling_step_id)
+            sibling_step = steps_by_id.get(sibling_step_id)
+            if sibling_step is not None:
+                # Non-terminal, and not picked up by the same-wakeup
+                # backfill in schedule_ready_steps(): _ready_steps() offers
+                # this step again once active_step_ids drains, but
+                # _pending_ready_steps() only collects steps already at
+                # "pending", so it waits for the next wakeup's readiness
+                # check before it reruns.
+                sibling_step.status = "clarification_invalidated"
+            invalidated_step_ids.append(sibling_step_id)
+        return invalidated_step_ids
 
     async def _cancel_pending_steps(
         self,
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -1860,6 +1860,19 @@ class DAGPattern(AgentPattern):
                 None,
             )
 
+    def has_live_step_tasks(self) -> bool:
+        """Return whether a concurrent step batch is still running right now.
+
+        Only the multi-step branch of ``_execute_ready_steps`` ever
+        populates ``_live_step_tasks``; the single-step fast path never
+        creates an ``asyncio.Task`` and this is always ``False`` on that
+        path. ``_execute_ready_steps`` clears the set in a ``finally``
+        before it returns, whatever the outcome, so a caller inspecting
+        this once ``run()`` has returned always sees an accurate answer.
+        """
+
+        return bool(self._live_step_tasks)
+
     def _waiting_step_id(self) -> str | None:
         for step_id in self.active_step_ids:
             state = self.active_step_pattern_states.get(step_id)
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -817,11 +817,28 @@ class DAGPattern(AgentPattern):
                     )
 
                     if superseded_step_ids:
+                        # A superseded question always means a waiting loser,
+                        # but the winner is not always a question itself --
+                        # an interrupted winner delivers no question of its
+                        # own, so the log line must say so truthfully rather
+                        # than implying %s's question was the one delivered.
+                        if winner_result.get("status") == "waiting_for_user":
+                            log_message = (
+                                "Multiple concurrent DAG steps completed with "
+                                "a pending user question in the same batch; "
+                                "only %s's question is delivered, the rest "
+                                "are invalidated."
+                            )
+                        else:
+                            log_message = (
+                                "Multiple concurrent DAG steps completed with "
+                                "a pending user question in the same batch; "
+                                "questions from the superseded steps were "
+                                "discarded because an interrupt from %s takes "
+                                "precedence."
+                            )
                         logger.warning(
-                            "Multiple concurrent DAG steps completed with a "
-                            "pending user question in the same batch; only "
-                            "%s's question is delivered, the rest are "
-                            "invalidated.",
+                            log_message,
                             winner_step_id,
                             extra={
                                 "execution_id": getattr(
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -321,7 +321,14 @@ class _RuntimeLLMProxy:
 
 
 class DAGPattern(AgentPattern):
-    """Minimal DAG execution pattern that reuses ReActPattern for each step."""
+    """Minimal DAG execution pattern that reuses ReActPattern for each step.
+
+    A task is only ever allowed one active question outstanding at a time
+    against the user; the DAG side of that rule is exactly-one: when
+    concurrent steps in the same batch all end up waiting for a user reply
+    in the same wakeup, exactly one of them gets to keep its question, and
+    the rest are invalidated rather than left to race the user's answer.
+    """
 
     def __init__(
         self,
@@ -354,6 +361,11 @@ class DAGPattern(AgentPattern):
         self.planned_user_message_count = 0
         self.completion_feedback: str | None = None
         self.completion_replan_count = 0
+        # Live concurrent step tasks for the in-progress batch, maintained by
+        # _execute_ready_steps() and always empty once it returns (whatever
+        # the outcome). asyncio.Task objects are not JSON-serializable, so
+        # this must never be added to get_state().
+        self._live_step_tasks: set[asyncio.Task[Any]] = set()
 
     async def run(
         self,
@@ -679,6 +691,7 @@ class DAGPattern(AgentPattern):
         steps_by_id = {step.id: step for step in steps}
         pending = set(tasks)
         scheduled_step_ids = set(steps_by_id)
+        self._live_step_tasks = set(pending)
 
         def schedule_ready_steps() -> None:
             if self.plan is None:
@@ -709,6 +722,7 @@ class DAGPattern(AgentPattern):
                 tasks[task] = ready_step.id
                 steps_by_id[ready_step.id] = ready_step
                 pending.add(task)
+                self._live_step_tasks.add(task)
                 scheduled_step_ids.add(ready_step.id)
                 available_slots -= 1
                 if available_slots <= 0:
@@ -720,68 +734,124 @@ class DAGPattern(AgentPattern):
                     pending,
                     return_when=asyncio.FIRST_COMPLETED,
                 )
+                self._live_step_tasks = set(pending)
+
+                # A single wakeup can complete more than one step; collect
+                # every non-None result instead of acting on whichever task
+                # a ``set`` happens to iterate first.
+                completed_results: list[tuple[asyncio.Task[Any], dict[str, Any]]] = []
                 for task in done:
                     result = task.result()
                     if result is not None:
-                        await self._cancel_pending_steps(
-                            pending,
-                            step_ids_by_task=tasks,
-                            steps_by_id=steps_by_id,
+                        completed_results.append((task, result))
+
+                if completed_results:
+                    # Deterministic winner selection: an interrupted result
+                    # always outranks a waiting one (an interrupt is a
+                    # control instruction, not a workflow question the user
+                    # can be asked to sit through), and within the same kind
+                    # the lexicographically first step id wins.
+                    completed_results.sort(
+                        key=lambda item: (
+                            0 if item[1].get("status") != "waiting_for_user" else 1,
+                            tasks[item[0]],
                         )
-                        if pending:
-                            await runtime.checkpoint(
-                                "dag_after_cancelled_siblings",
-                                context=root_context,
-                                pattern=self,
-                                status=self.status,
-                                metadata={
-                                    "active_step_ids": list(self.active_step_ids),
-                                    "cancelled_step_ids": [
-                                        step_id
-                                        for task, step_id in tasks.items()
-                                        if task in pending
-                                    ],
-                                },
-                            )
-                        return result
-                    failed_step = next(
-                        (
-                            step
-                            for step in steps_by_id.values()
-                            if step.status == "failed"
-                        ),
-                        None,
                     )
-                    if failed_step is not None:
-                        await self._cancel_pending_steps(
-                            pending,
-                            step_ids_by_task=tasks,
-                            steps_by_id=steps_by_id,
+                    winner_task, winner_result = completed_results[0]
+                    winner_step_id = tasks[winner_task]
+
+                    superseded_step_ids: list[str] = []
+                    for loser_task, loser_result in completed_results[1:]:
+                        if loser_result.get("status") != "waiting_for_user":
+                            continue
+                        loser_step_id = tasks[loser_task]
+                        self._clear_active_step(loser_step_id)
+                        loser_step = steps_by_id.get(loser_step_id)
+                        if loser_step is not None:
+                            # Non-terminal, and not picked up by the same-
+                            # wakeup backfill in schedule_ready_steps():
+                            # _ready_steps() offers this step again once
+                            # active_step_ids drains, but
+                            # _pending_ready_steps() only collects steps
+                            # already at "pending", so it waits for the next
+                            # wakeup's readiness check before it reruns.
+                            loser_step.status = "clarification_invalidated"
+                        superseded_step_ids.append(loser_step_id)
+
+                    if superseded_step_ids:
+                        logger.error(
+                            "Multiple concurrent DAG steps completed with a "
+                            "pending user question in the same batch; only "
+                            "%s's question is delivered, the rest are "
+                            "invalidated.",
+                            winner_step_id,
+                            extra={
+                                "execution_id": getattr(
+                                    root_context, "execution_id", None
+                                ),
+                                "superseded_step_ids": superseded_step_ids,
+                                "winner_step_id": winner_step_id,
+                            },
                         )
-                        return await self._fail(
+                        winner_result = {
+                            **winner_result,
+                            "clarification_superseded_step_ids": superseded_step_ids,
+                        }
+
+                    await self._cancel_pending_steps(
+                        pending,
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
+                    if pending or superseded_step_ids:
+                        await runtime.checkpoint(
+                            "dag_after_cancelled_siblings",
                             context=root_context,
-                            runtime=runtime,
-                            error=(
-                                failed_step.error or f"Step {failed_step.id} failed."
-                            ),
-                            failure_reason="step_failed",
-                            checkpoint_label="dag_failed",
-                            failed_step_id=failed_step.id,
+                            pattern=self,
+                            status=self.status,
+                            metadata={
+                                "active_step_ids": list(self.active_step_ids),
+                                "cancelled_step_ids": [
+                                    step_id
+                                    for task, step_id in tasks.items()
+                                    if task in pending
+                                ],
+                            },
                         )
-                    if self._needs_replan(root_context):
-                        await self._cancel_pending_steps(
-                            pending,
-                            step_ids_by_task=tasks,
-                            steps_by_id=steps_by_id,
-                        )
-                        return None
-                    if self.status in {"interrupted", "waiting_for_user"}:
-                        await self._cancel_pending_steps(
-                            pending,
-                            step_ids_by_task=tasks,
-                            steps_by_id=steps_by_id,
-                        )
-                        return None
+                    return winner_result
+
+                failed_step = next(
+                    (step for step in steps_by_id.values() if step.status == "failed"),
+                    None,
+                )
+                if failed_step is not None:
+                    await self._cancel_pending_steps(
+                        pending,
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
+                    return await self._fail(
+                        context=root_context,
+                        runtime=runtime,
+                        error=(failed_step.error or f"Step {failed_step.id} failed."),
+                        failure_reason="step_failed",
+                        checkpoint_label="dag_failed",
+                        failed_step_id=failed_step.id,
+                    )
+                if self._needs_replan(root_context):
+                    await self._cancel_pending_steps(
+                        pending,
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
+                    return None
+                if self.status in {"interrupted", "waiting_for_user"}:
+                    await self._cancel_pending_steps(
+                        pending,
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
+                    return None
                 schedule_ready_steps()
         except Exception:
             try:
@@ -793,6 +863,8 @@ class DAGPattern(AgentPattern):
             except Exception:
                 logger.exception("Failed to clean up cancelled DAG sibling steps.")
             raise
+        finally:
+            self._live_step_tasks.clear()
         return None
 
     async def _execute_step(
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -870,7 +870,7 @@ class DAGPattern(AgentPattern):
                     )
                     return None
                 schedule_ready_steps()
-        except Exception:
+        except BaseException:
             try:
                 await self._cancel_pending_steps(
                     pending,
@@ -1886,6 +1886,11 @@ class DAGPattern(AgentPattern):
         path. ``_execute_ready_steps`` clears the set in a ``finally``
         before it returns, whatever the outcome, so a caller inspecting
         this once ``run()`` has returned always sees an accurate answer.
+        This holds for cancellation too: when the task running the
+        pattern is cancelled, the sibling step tasks are cancelled and
+        awaited inside the same exception handler before the ``finally``
+        clears the set, so an empty set on return still means no step
+        task is left running.
         """
 
         return bool(self._live_step_tasks)
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -1935,11 +1935,32 @@ class DAGPattern(AgentPattern):
         """Pick the one result a same-wakeup completed batch delivers.
 
         Sorts ``completed_results`` in place and returns its new first
-        entry: an interrupted result always outranks a waiting one (an
-        interrupt is a control instruction, not a workflow question the
-        user can be asked to sit through), and within the same kind the
-        lexicographically first step id wins. Callers needing the losers
-        read ``completed_results[1:]`` after this call.
+        entry. The full precedence a batch's outcome follows, from
+        highest to lowest:
+
+        1. A failed step outranks every completed result in the same
+           batch (handled by the caller before this method is even
+           reached: a doomed DAG must not ask a question or report an
+           interrupt in place of the failure).
+        2. Among completed results, an interrupted one outranks a
+           waiting one -- an interrupt is a control instruction, not a
+           workflow question the user can be asked to sit through.
+        3. Within the same kind, the lexicographically first step id
+           wins, for a deterministic pick regardless of which task an
+           ``asyncio.wait()`` wakeup or set iteration happens to surface
+           first.
+        4. Any completed result -- winner or loser -- outranks a
+           pending replan. The caller only checks ``_needs_replan()``
+           after this batch has been fully handled, so a result the
+           user has already been shown (a question, an interrupt) is
+           never discarded just because a new user message arrived
+           mid-batch. That message is not lost either: it does not
+           advance ``planned_user_message_count``, so the next
+           iteration of ``run()``'s loop still sees the replan as
+           pending and picks it up one turn later.
+
+        Callers needing the losers read ``completed_results[1:]`` after
+        this call.
         """
         completed_results.sort(
             key=lambda item: (
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -1975,8 +1975,8 @@ class DAGPattern(AgentPattern):
            not a workflow question the user can be asked to sit
            through. A result status the table does not know sorts last
            instead of raising (see ``rank()`` below for why), and is
-           logged loudly since an unranked status is itself worth
-           investigating.
+           logged as a warning since an unranked status is a handled
+           anomaly worth investigating, not a failure of this method.
         3. Within the same rank, the lexicographically first step id
            wins, for a deterministic pick regardless of which task an
            ``asyncio.wait()`` wakeup or set iteration happens to surface
@@ -2012,7 +2012,7 @@ class DAGPattern(AgentPattern):
                 # the entire batch over one unrecognized status string.
                 # Sorting it last instead preserves the invariant that a
                 # question already asked is delivered whenever possible.
-                logger.error(
+                logger.warning(
                     "Completed DAG step %s has an unranked result status "
                     "%r; ranking it last in this batch instead of raising.",
                     step_ids_by_task[task],
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -760,6 +760,14 @@ class DAGPattern(AgentPattern):
                     winner_task, winner_result = completed_results[0]
                     winner_step_id = tasks[winner_task]
 
+                    # This loop invalidates losers drawn from `done`; the
+                    # cancellation below acts on `pending`. `asyncio.wait`
+                    # partitions the tasks it was given into exactly these
+                    # two sets, so a task can never be in both -- the two
+                    # cleanup passes touch disjoint task sets, and running
+                    # this one before or after the cancellation below
+                    # cannot change which steps end up invalidated versus
+                    # cancelled.
                     superseded_step_ids: list[str] = []
                     for loser_task, loser_result in completed_results[1:]:
                         if loser_result.get("status") != "waiting_for_user":
@@ -816,6 +824,7 @@ class DAGPattern(AgentPattern):
                                     for task, step_id in tasks.items()
                                     if task in pending
                                 ],
+                                "superseded_step_ids": superseded_step_ids,
                             },
                         )
                     return winner_result
--- a/src/xagent/core/agent/pattern/dag/dag.py
+++ b/src/xagent/core/agent/pattern/dag/dag.py
@@ -45,6 +45,16 @@ logger = logging.getLogger(__name__)
 
 DAG_COMPLETION_TOOL_NAME = "assess_dag_completion"
 
+# Precedence for _select_winner()'s ranking of a same-wakeup completed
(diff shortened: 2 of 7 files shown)


src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,68 +734,133 @@ def schedule_ready_steps() -> None:
...
+                    return None
+                if self.status in {"interrupted", "waiting_for_user"}:
+                    await self._cancel_pending_steps(
+                        pending,
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
+                    return None
                 schedule_ready_steps()
         except Exception:
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in bf442063e. The outer handler of `_execute_ready_steps` now catches `BaseException` and re-raises unconditionally, so a cancellation thrown into the coroutine reaches the sibling-cleanup call (`_cancel_pending_steps`) instead of bypassing it — previously the sibling step tasks were left running as orphans while `has_live_step_tasks()` reported none. The inner `except Exception` guarding the cleanup itself is unchanged, and cancellation semantics are preserved (the outer task still ends cancelled).

A regression test pins the outcome: cancelling the task running `pattern.run()` now leaves both concurrent step tasks cancelled and `active_step_ids` empty; the test fails under the old `except Exception` form.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,68 +734,133 @@ def schedule_ready_steps() -> None:
...
                 for task in done:
                     result = task.result()
                     if result is not None:
-                        await self._cancel_pending_steps(
-                            pending,
-                            step_ids_by_task=tasks,
-                            steps_by_id=steps_by_id,
+                        completed_results.append((task, result))
+
+                if completed_results:
```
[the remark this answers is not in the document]
AlexLiu190625: Fixed in f95afff0d. The failed-step scan now runs before winner selection, so a batch containing a failed step fails fast with `failure_reason="step_failed"` before any waiting or interrupted sibling result can be returned — the DAG no longer asks the user a question when the run is already doomed. Failure outranks both result kinds: two regression tests pin failed-plus-waiting (no question is returned) and failed-plus-interrupted (the failure is reported, not the interrupt), and each fails with the scan in its old position.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,70 +734,143 @@ def schedule_ready_steps() -> None:
...
-                        ),
-                        None,
                     )
-                    if failed_step is not None:
-                        await self._cancel_pending_steps(
-                            pending,
-                            step_ids_by_task=tasks,
-                            steps_by_id=steps_by_id,
+                    winner_task, winner_result = completed_results[0]
+                    winner_step_id = tasks[winner_task]
```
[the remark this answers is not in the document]
AlexLiu190625: This inconsistency is real, pre-existing on main, and already tracked as #1284. The last-writer-wins race between each step coroutine's `self.status` write and the deterministic winner selection exists on the base commit as well — the probe in #1284 reproduces the identical divergence (returned `status` `"interrupted"` while `pattern.status` and the turn's checkpoint say `"waiting_for_user"`) without any of this PR's changes. This PR narrows its behavior change to the exactly-one clarification rule, so the status synchronization fix — including where the winner's status should be written and what the checkpoint for that turn must record — is scoped to #1284 rather than folded in here.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,70 +734,143 @@ def schedule_ready_steps() -> None:
...
+                        completed_results.append((task, result))
+
+                # A batch containing a failed step fails fast, before any
+                # winner is chosen: a DAG that is already doomed must not
+                # ask the user a question or report an interrupt instead of
+                # the failure. This scan therefore runs ahead of the
+                # completed_results winner selection below, so a failure
+                # always outranks an interrupted or waiting result from a
+                # sibling that completed in the same wakeup.
+                failed_step = next(
```
rogercloud, line 777: **Major — failure-wins path leaks a waiting sibling's active-step bookkeeping.**

This `failed_step` early-return runs *before* the `completed_results` winner-selection/invalidation logic below. If a sibling step in the same wakeup's `done` set already completed with `waiting_for_user`, it has already registered itself in `active_step_ids` / `active_step_pattern_states` / `active_step_contexts` via the setters in `_execute_step_impl` (lines 1039-1044). This branch's `_cancel_pending_steps` call only walks `pending` (tasks not yet done), so it never touches that sibling, and `_fail()` (lines 1223-1254) performs no active-step cleanup at all.

Verified empirically: after this path returns a terminal `failed` result and the `dag_failed` checkpoint is persisted, the waiting sibling's bookkeeping is still live — an orphaned "actively waiting on a user question" record for a run that's already over. `test_dag_pattern_failed_step_wins_over_waiting_sibling_in_same_batch` (tests/core/agent/test_dag.py:1678) doesn't catch this because it never asserts on `pattern.active_step_ids`; adding that one assertion to the test as written today makes it fail with `active_step_ids == ['wait_step']` instead of `[]` (same for the interrupted-sibling variant, `['interrupt_step']`).

Suggested fix: before this early-return, apply the same invalidation the winner-selection path below applies to losing waiting steps — clear bookkeeping for any `waiting_for_user` sibling in `done` and mark it `clarification_invalidated` — rather than only cancelling `pending`. Then extend the existing tests with an `active_step_ids == []` assertion.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,70 +734,143 @@ def schedule_ready_steps() -> None:
...
+                                    for task, step_id in tasks.items()
+                                    if task in pending
+                                ],
+                                "superseded_step_ids": superseded_step_ids,
+                            },
                         )
-                        return None
+                    return winner_result
+
+                if self._needs_replan(root_context):
```
rogercloud, line 885: **Minor/moderate — replan-need precedence is now undocumented and untested.**

Before this PR, `_needs_replan` and the `self.status in {"interrupted", "waiting_for_user"}` check ran inside the `for task in done:` loop, evaluated for every task whose result was `None`. Since `_execute_step` mutates `self.status` from inside the step coroutine (visible before that task's own loop iteration), the old code could reach a `None`-result task's checks and return `None` (triggering a replan) — depending on set-iteration order — even when a sibling in the same `done` set had a non-`None` waiting/interrupted result.

After this PR both checks sit outside the loop and are unconditionally unreachable whenever `completed_results` is non-empty: a completed waiting/interrupted result now always wins over a simultaneous replan need. This is a real control-flow change, though the prior behavior was itself set-order-dependent, so this converts a pre-existing race into a deterministic (but unstated) rule rather than removing a previously reliable guarantee.

Suggest adding a comment at the winner-selection site recording the full precedence order (`failed > interrupted > waiting_for_user > replan`) and why a completed result outranks a pending replan need, plus a test pinning the case where one step needs a replan and a sibling completes `waiting_for_user` in the same batch.

src/xagent/core/agent/execution_adapter.py
```diff
@@ -405,6 +405,13 @@ def _normalize_result(
                     # carry the same draft too, but callers should not dig
                     # it out from there.
                     "clarification_draft": result.get("clarification_draft"),
+                    # Empty list rather than ``None`` so a reader only ever
+                    # needs one check (``if superseded:``) instead of also
+                    # distinguishing "key absent" from "key present but
+                    # empty".
+                    "clarification_superseded_step_ids": (
```
rogercloud, line 412: **Minor — `clarification_superseded_step_ids` is dropped for `interrupted` winners.**

This promotion is gated inside `if status == "waiting_for_user":` above. But `dag.py`'s winner-selection can attach a non-empty `clarification_superseded_step_ids` to a winner whose status is `"interrupted"` — `tests/core/agent/test_dag.py:2247` (`test_dag_pattern_mixed_batch_interrupted_wins_over_waiting`) asserts exactly `result["status"] == "interrupted"` together with a non-empty `result["clarification_superseded_step_ids"]`. Since `_normalize_result` builds a fresh dict of picked fields per status branch, that signal is silently discarded here — precisely the case where a caller most needs to know "a sibling's question was invalidated, and you are not being asked anything right now."

Suggest promoting `clarification_superseded_step_ids` whenever it's present on the raw result (or at least also for the `interrupted` branch), with a test case covering interrupted-with-superseded-ids in test_execution_adapter.py.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,70 +734,143 @@ def schedule_ready_steps() -> None:
...
+                            # _ready_steps() offers this step again once
+                            # active_step_ids drains, but
+                            # _pending_ready_steps() only collects steps
+                            # already at "pending", so it waits for the next
+                            # wakeup's readiness check before it reruns.
+                            loser_step.status = "clarification_invalidated"
+                        superseded_step_ids.append(loser_step_id)
+
+                    if superseded_step_ids:
+                        logger.error(
```
rogercloud: **Minor — `logger.error` for an expected, correctly-handled race.**

This fires on a condition the code explicitly handles correctly (deterministic winner selection). The subsystem's own convention uses `logger.warning` for structurally identical "anomalous but handled" cases (plan_generator.py:343, :355, :370), while `logger.error` elsewhere under `src/xagent/core/agent/` is reserved for genuine failures/degradation. Suggest downgrading to `logger.warning`.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,70 +734,143 @@ def schedule_ready_steps() -> None:
...
+                        loser_step = steps_by_id.get(loser_step_id)
+                        if loser_step is not None:
+                            # Non-terminal, and not picked up by the same-
+                            # wakeup backfill in schedule_ready_steps():
+                            # _ready_steps() offers this step again once
+                            # active_step_ids drains, but
+                            # _pending_ready_steps() only collects steps
+                            # already at "pending", so it waits for the next
+                            # wakeup's readiness check before it reruns.
+                            loser_step.status = "clarification_invalidated"
```
rogercloud: **Minor — new step status `"clarification_invalidated"` has no declared contract and no frontend slot.**

`PlanStep.status` (plan_generator.py) is a bare untyped `str` with no enum, and this literal appears nowhere in `frontend/src/`. The frontend `StepExecution["status"]` TS union (app-context-chat.tsx) has no member for it, and `normalizeStepStatus` falls unrecognized values back to `"pending"` — but more importantly, `status: existingStep?.status || normalizeStepStatus(step.status)` means a step that already had a truthy status (e.g. `"running"`) keeps displaying that stale value after invalidation, since there's no dedicated WebSocket event for step invalidation, only the generic plan-sync path. The frontend code is pre-existing and untouched by this PR, but the PR introduces a backend status value the frontend contract can't represent.

Suggest documenting the full set of valid `PlanStep.status` values (including this new one) next to the field, and filing a frontend follow-up for rendering invalidated steps.

rogercloud (changes requested):
## Summary

This PR makes the DAG pattern's handling of a concurrent wakeup batch deterministic. Previously, when several concurrently running plan steps completed in the same `asyncio.wait` wakeup and more than one reached a terminal-ish state, the pattern returned whichever result `set` iteration happened to visit first — so with two simultaneous `waiting_for_user` steps, the "losing" question was silently stranded with no delivery path. The change collects all completed results, applies a total order (`failed` > `interrupted` > `waiting_for_user`, ties broken by lexicographically smallest step id), and explicitly invalidates losing waiting steps: clearing their active-step bookkeeping, marking them `clarification_invalidated`, and reporting their ids via a new `clarification_superseded_step_ids` key. It also attributes each waiting draft to its origin step (`with_origin_step`) and adds a `has_live_step_tasks()` safety-net assertion in `AgentRunner`. Diff spans `src/xagent/core/agent/pattern/dag/dag.py`, `src/xagent/core/agent/execution_adapter.py`, `src/xagent/core/agent/runner.py`, plus three test files (~1100 lines, mostly new tests).

## Approach verdict

**Acceptable with reservations.** Collect-all -> total-order -> invalidate-losers is the right shape and close to the minimal fix for the stated nondeterminism; it fits `DAGPattern`'s existing structure without stretching the abstraction. Three design-level observations, none blocking:

- **`clarification_superseded_step_ids` has zero non-test consumers.** A full grep of `src/` and `frontend/` finds no reader outside tests. That is fine as forward-looking observability plumbing, but it should be understood as such rather than as finished user-facing behavior — nothing today surfaces "your other question was superseded" to a user or an API client.
- **`has_live_step_tasks()`'s `True`/raise branch is structurally unreachable for `DAGPattern` today** — the `finally` clause always clears `_live_step_tasks` before `_execute_ready_steps`/`run()` can return. That makes it a defensive tripwire for a hypothetical future pattern rather than a live check. This is a reasonable and precedented choice, and the duck-typed `getattr` probe matches how `get_state()` is already handled in this codebase. Noted as a deliberate design decision, not a requested change.
- **`_execute_ready_steps` is now ~270 lines** with five textually identical `_cancel_pending_steps(...)` call sites. See Simplification below.

## Findings

### Major

**Failure-wins path leaks a waiting sibling's active-step bookkeeping** — see inline comment on `src/xagent/core/agent/pattern/dag/dag.py:755`.

### Minor / moderate

See inline comments on:
- `src/xagent/core/agent/pattern/dag/dag.py:858` — replan-need precedence undocumented and untested
- `src/xagent/core/agent/execution_adapter.py:412` — `clarification_superseded_step_ids` dropped for interrupted winners
- `src/xagent/core/agent/pattern/dag/dag.py:816` — `logger.error` used for an expected, handled race
- `src/xagent/core/agent/pattern/dag/dag.py:812` — new step status has no declared contract and no frontend slot

**No 3+-step invalidation test** — every new invalidation test in `tests/core/agent/test_dag.py` uses exactly two concurrent steps. With two steps, the loser loop `for loser_task, loser_result in completed_results[1:]:` is indistinguishable from a hypothetical off-by-one such as `completed_results[1:2]`; only a 3+-step scenario separates "invalidate all losers" from "invalidate only the second element." Suggest adding one test with three simultaneously waiting steps.

**Checkpoint condition never tested with both branches non-empty** — the `if pending or superseded_step_ids:` checkpoint write (`dag.py`) has existing tests for pending-only (cancelled sibling) and superseded-only (double waiting), but never both at once. Low priority — a 3-step case (two waiting, one still running) would cover it; the two fields do not interact in the code.

### On the returned-status vs persisted-status mismatch (issue #1284)

While reviewing, the mismatch between the winner's returned status and the pattern's own persisted `self.status` in a mixed `interrupted` + `waiting_for_user` batch was independently reproduced and confirmed as a genuine defect. It is, however, pre-existing on `main`, already tracked by the author as issue #1284, and already raised in this PR's own review history by gemini-code-assist[bot], with the author replying that it is out of scope for this narrower "exactly-one clarification" fix. That disposition is correct and worth commending: keeping this PR scoped to deterministic winner selection rather than folding in the persisted-status reconciliation keeps both changes reviewable. It is not re-raised as a finding here, and no test pinning `pattern.status` after a mixed batch is requested for the same reason — that assertion belongs with #1284's fix, since the value is racy by design until then.

- `src/xagent/core/agent/pattern/dag/dag.py` (`_execute_ready_steps`, ~lines 616-885): shrink — ~270 lines with 5 identical `_cancel_pending_steps(pending, step_ids_by_task=tasks, steps_by_id=steps_by_id)` call sites. Replace with a local `cancel_all()` closure — safe here, since Python closures see reassignments of `pending` by reference and this function already depends on exactly that behavior for its existing `schedule_ready_steps` closure.
- `src/xagent/core/agent/pattern/dag/dag.py` (winner-selection/loser-invalidation block, ~lines 774-833): shrink — extract `_select_winner(completed_results)` and `_invalidate_waiting_losers(losers, ...)` helpers so the `failed > interrupted > waiting_for_user > replan` precedence is readable at a glance instead of only reconstructible from comments — and so the fix for the major finding above has a single shared call site to reuse from the failure path.
- `net: ~15-20 lines possible`

Separately, and not a simplification tag: `clarification_superseded_step_ids` has **no current consumer** anywhere outside tests (confirmed across `src/` and `frontend/`). Please confirm this is intentional forward groundwork for a planned feature rather than a signal that was meant to be wired up in this PR.

## Review coverage

Code Quality, Correctness, Test Quality/Coverage, Design/Architecture, and Documentation were all evaluated.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Fixed in f4790ee8e. Before the fail-fast return, the batch now clears every completed sibling's active-step bookkeeping (all three stores); a `waiting_for_user` sibling is additionally marked `clarification_invalidated`, while an `interrupted` sibling keeps its status. Both existing tests now assert `active_step_ids == []` together with the other two stores and the `dag_failed` checkpoint payload — with the fix reverted they fail exactly as described here (`['wait_step']` / `['interrupt_step']`).

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Done in f05e626f1. The full precedence (`failed` > `interrupted` > `waiting_for_user` > pending replan) is documented on `_select_winner`, including why a completed result outranks a pending replan need: the user's new message is not consumed by the skipped check, so `_needs_replan` still fires at the top of the next `run()` iteration and the replan happens one round later. The new test pins the case described here and asserts the plan is generated exactly once — asserting only on the returned status cannot distinguish a reordered check, since a replan-first mutant still returns `waiting_for_user` one round later.

src/xagent/core/agent/execution_adapter.py
AlexLiu190625: Agreed on the facts. The promotion for interrupted winners belongs to a follow-up change on the clarification-resolution path, which owns the consumer of this key and will carry the same-shaped promotion plus the interrupted-with-superseded-ids adapter test suggested here. Keeping this PR's adapter untouched avoids two changes editing the same `_normalize_result` branch at once.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Downgraded to `logger.warning` in 4f752f4f2 — agreed this is an anomalous-but-handled case per the subsystem's own convention.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Done in b2c6bffa9: the six-value domain is now documented next to the field, including that `waiting_for_user` is never a step-level status (it only ever applies to the owning pattern). Frontend follow-up filed as #1296.

AlexLiu190625:
All round findings are addressed at head b2c6bffa9:

- **Major (fail-fast bookkeeping leak)**: fixed in f4790ee8e — details on the inline thread.
- **Three-step invalidation test**: 32a147535 adds a test with three simultaneously waiting steps — under a `completed_results[1:2]` off-by-one mutation it is the only failing test in the file — plus one with two waiting and one still-running sibling, which exercises the `if pending or superseded_step_ids:` checkpoint write with both branches non-empty.
- **Both simplification suggestions adopted** in 44c271685: `_select_winner` / `_invalidate_batch_siblings` are named methods (the precedence comment from the inline thread lives on the former), and the five identical `_cancel_pending_steps` call sites are now one `cancel_all()` closure.
- **On the `clarification_superseded_step_ids` question**: intentional forward groundwork with a named consumer. A follow-up change on the clarification-resolution path reads it to register an operations degradation signal when a batch produced more than one draft — the DAG deliberately does not do that itself, because the core pattern layer has no dependency on the web layer and must not grow one. Nothing user-facing surfaces "your other question was superseded" today, and this PR does not claim otherwise.
- **Interrupted-winner promotion in the adapter**: deferred to that same follow-up change (inline thread has details); the adapter is untouched here.
- **Step-status contract**: value domain documented at the field (b2c6bffa9); frontend rendering follow-up filed as #1296.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -1821,6 +1926,103 @@ def _tool_name(self, tool: Any) -> str:
...
+        Returns the ids that got the clarification_invalidated flip, in
+        input order.
+        """
+        invalidated_step_ids: list[str] = []
+        for sibling_task, sibling_result in siblings:
+            is_waiting = sibling_result.get("status") == "waiting_for_user"
+            if not is_waiting and not clear_all_bookkeeping:
+                continue
+            sibling_step_id = step_ids_by_task[sibling_task]
+            self._clear_active_step(sibling_step_id)
```
rogercloud, line 2072: **[major]** Clearing the active-step bookkeeping here discards the loser's ReAct pattern state and child context. When this step is later offered as ready work again, `active_pattern_state` is `None`, so there is no `load_state()` — it restarts from a brand-new context. Any tool calls the step already executed before losing the tie-break will be executed a second time; nothing memoizes or checkpoints them. For non-idempotent tools (file writes, outbound API calls, messages) that is a real duplication risk, and it is a different problem from the "duplicate question" framing in the PR description.

`test_dag_pattern_invalidated_step_is_rescheduled_after_winner_settles` pins the context-discard mechanics, but it runs with `tools=[]` and no tool calls, so it cannot demonstrate replay safety either way.

Suggest either calling this out explicitly as a known limitation in the PR description, or adding a test with a tool-using/stateful step so the replay behavior is visible and deliberate.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,79 +746,130 @@ def schedule_ready_steps() -> None:
...
+                    # cancelled.
+                    superseded_step_ids = self._invalidate_batch_siblings(
+                        completed_results[1:],
+                        step_ids_by_task=tasks,
+                        steps_by_id=steps_by_id,
+                    )
+
+                    if superseded_step_ids:
+                        logger.warning(
+                            "Multiple concurrent DAG steps completed with a "
```
rogercloud: **[minor]** This message is factually wrong when the winner's status is `interrupted`. In a mixed batch the interrupted result wins, and it is not asking anything — so no question was "delivered" at all, yet the log says one was. This path is reachable and already exercised by `test_dag_pattern_mixed_batch_interrupted_wins_over_waiting`.

Suggest conditioning the message text on `winner_result["status"]`, e.g. a "…delivered" phrasing for a waiting winner and a "…superseded by an interrupt from %s" phrasing otherwise.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -959,8 +1036,18 @@ async def _execute_step_impl(
...
+            # frozen dataclass, so this is a copy, not an in-place edit.
+            draft = result.get("clarification_draft")
+            attributed_result = dict(result)
+            if draft is not None:
+                attributed_result["clarification_draft"] = draft.with_origin_step(
+                    step.id
+                )
             return {
-                **result,
+                **attributed_result,
```
rogercloud, line 1082: **[minor]** `on_dag_step_end` is not emitted on this `waiting_for_user` return, nor on the `interrupted` return just above, nor when a step is invalidated by `_invalidate_batch_siblings` — it only fires on the exception, failure, and normal-success paths.

Consequence: an invalidated step that later restarts re-emits `on_dag_step_start` for the same step id with no matching prior `on_dag_step_end`, so any lifecycle consumer that expects paired start/end events per step id sees an orphaned start. (Backend-only observation — I did not evaluate frontend handling.)

Suggest either emitting `on_dag_step_end` with a distinguishing reason/status on these branches, or documenting that these statuses are intentionally exempt from the start/end pairing contract.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -1821,6 +1926,103 @@ def _tool_name(self, tool: Any) -> str:
...
+           advance ``planned_user_message_count``, so the next
+           iteration of ``run()``'s loop still sees the replan as
+           pending and picks it up one turn later.
+
+        Callers needing the losers read ``completed_results[1:]`` after
+        this call.
+        """
+        completed_results.sort(
+            key=lambda item: (
+                0 if item[1].get("status") != "waiting_for_user" else 1,
```
rogercloud: **[minor]** This key ranks *any* non-`waiting_for_user` status above `waiting_for_user`, but the docstring above states the rule as "an interrupted one outranks a waiting one". Harmless today, since only `interrupted` and `waiting_for_user` reach this code — but a future third completion status would silently outrank `waiting_for_user` without anyone having decided that is correct.

Suggest an explicit rank mapping instead, e.g. `{"interrupted": 0, "waiting_for_user": 1}[status]` (or a `.get()` with a loud fallback / warning), so an unrecognized status fails visibly rather than defaulting to "wins".

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -1821,6 +1926,103 @@ def _tool_name(self, tool: Any) -> str:
...
+            sibling_step = steps_by_id.get(sibling_step_id)
+            if sibling_step is not None:
+                # Non-terminal, and not picked up by the same-wakeup
+                # backfill in schedule_ready_steps(): _ready_steps() offers
+                # this step again once active_step_ids drains, but
+                # _pending_ready_steps() only collects steps already at
+                # "pending", so it waits for the next wakeup's readiness
+                # check before it reruns.
+                sibling_step.status = "clarification_invalidated"
+            invalidated_step_ids.append(sibling_step_id)
```
rogercloud: **[minor]** This append runs unconditionally, including when `steps_by_id.get(sibling_step_id)` returned `None` and no status was actually flipped. The docstring promises the return value contains "the ids that got the `clarification_invalidated` flip", and callers use it to populate `clarification_superseded_step_ids`, so a missing step id ends up reported as superseded when nothing happened to it.

Suggest moving the append inside the `if sibling_step is not None:` branch.

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -720,79 +746,130 @@ def schedule_ready_steps() -> None:
...
-                        return await self._fail(
+                        winner_result = {
+                            **winner_result,
+                            "clarification_superseded_step_ids": superseded_step_ids,
+                        }
+
+                    await cancel_all()
+                    if pending or superseded_step_ids:
+                        await runtime.checkpoint(
+                            "dag_after_cancelled_siblings",
```
rogercloud, line 869: **[minor]** The label is `"dag_after_cancelled_siblings"` whenever `pending or superseded_step_ids` is truthy — including the supersede-only case where `pending` is empty and nothing was actually cancelled. A consumer keyed off the label alone would wrongly infer a cancellation occurred.

Suggest either a status-aware label, or adding an explicit `"cancelled": bool(pending)` to the metadata dict just below (it already carries `cancelled_step_ids`), so consumers do not have to infer it from the label.

tests/core/agent/test_dag.py
```diff
@@ -1670,6 +1674,1167 @@ async def chat(self, **kwargs: Any) -> dict[str, Any]:
...
+    steps_by_id = {step.id: step for step in pattern.plan.steps}
+    assert steps_by_id["b_interrupt"].status == "interrupted"
+
+    assert "b_interrupt" in pattern.active_step_ids
+    assert "b_interrupt" in pattern.active_step_pattern_states
+    assert "b_interrupt" in pattern.active_step_contexts
+
+
+@pytest.mark.asyncio
+async def test_dag_pattern_mixed_batch_interrupted_wins_over_waiting() -> None:
```
rogercloud: **[minor, coverage]** Two gaps I found while inventorying the multi-step batch tests in this file:

1. No test exercises a 3+-step batch with genuinely mixed statuses (e.g. two `interrupted` + one `waiting`). The existing three-way tests are all-waiting, or waiting + waiting + cancelled-running. That leaves the interaction of the status rank and the id tie-break within the *same* rank untested at batch size > 2.
2. No test drives a real `DAGPattern.run()` through an invalidation and then triggers `_generate_plan(replan=True)` in the same run, to confirm a step sitting at `clarification_invalidated` does not confuse replanning. The only `replan=True` usage in the suite is an unrelated standalone `LLMPlanGenerator` unit test with a hand-built previous plan.

tests/core/agent/test_dag.py
```diff
@@ -1670,6 +1674,1167 @@ async def chat(self, **kwargs: Any) -> dict[str, Any]:
...
+    assert steps_by_id["ask_b"].status == "clarification_invalidated"
+    assert steps_by_id["ask_a"].status == "running"
+
+    assert pattern.active_step_ids == ["ask_a"]
+    assert "ask_b" not in pattern.active_step_pattern_states
+    assert "ask_b" not in pattern.active_step_contexts
+
+
+@pytest.mark.asyncio
+async def test_dag_pattern_concurrent_double_waiting_winner_is_deterministic() -> None:
```
rogercloud, line 2020: **[minor, test quality]** This test and its triple-step variant are the only coverage of `_select_winner`'s ordering, and both reach it indirectly through an asyncio-scheduling-dependent fixture (steps parked on a shared `asyncio.Event` and released together). That makes the assertion about the winner contingent on scheduling details rather than on the sort itself.

Suggest adding a direct unit test that calls `_select_winner` with a hand-built, shuffled `completed_results` list (mixed statuses, ids out of order) and asserts the resulting order. It pins the exact behavior this PR is about, independent of async timing, and runs instantly.

tests/core/agent/test_execution_adapter.py
```diff
@@ -488,6 +488,51 @@ async def test_execution_adapter_omits_clarification_draft_key_when_not_waiting(
     assert "clarification_draft" not in result
 
 
+def test_execution_adapter_surfaces_superseded_step_ids_at_top_level() -> None:
```
rogercloud, line 491: **[minor, coverage]** This only covers `status == "waiting_for_user"`. `_normalize_result` promotes `clarification_superseded_step_ids` inside that branch only, so for an `interrupted` winner the field is silently dropped — which is exactly the mixed-batch case where a real, live question was discarded and the information matters most.

Suggest adding the `interrupted` counterpart so the current behavior is at least pinned, whichever way the follow-up decides to resolve the drop.

rogercloud (changes requested):
## Summary

This PR removes a non-deterministic outcome in `DAGPattern`: when several concurrently-running plan steps complete in the same `asyncio.wait()` wakeup and more than one of them lands on `waiting_for_user`, the previous code returned whichever result `set` iteration happened to surface first. The batch's results are now collected and ranked deterministically — a failed step outranks everything, an `interrupted` result outranks a `waiting_for_user` one, and ties break on the lexicographically smallest step id — and the losing "waiting" siblings are explicitly invalidated: their active-step bookkeeping is cleared, their `PlanStep.status` is flipped to the new `clarification_invalidated` value, and their ids are surfaced on the returned result under `clarification_superseded_step_ids`. Supporting changes: `ClarificationDraft.with_origin_step()` attribution so a draft carries the DAG step it came from, a `has_live_step_tasks()` capability plus a runner-side assertion guard, and an execution-adapter change promoting `clarification_superseded_step_ids` to the top level of a `waiting_for_user` result.

The core fix is sound, and the primary scenario is well covered by the new tests. Four design-level notes that are worth stating explicitly rather than blocking on:

1. **The tie-break is only deterministic *within* a wakeup batch.** The chosen ordering (interrupted > waiting, then smallest step id) has no relationship to which step actually asked first, and which steps land in the same batch is still scheduling-dependent. So end-to-end behavior across wakeup boundaries remains non-deterministic. This is inherent to any batch-boundary tie-break rather than a flaw in this implementation — raising it as calibration on what the PR does and does not buy, not as a change request.

2. **Prioritizing `interrupted` over `waiting_for_user` discards the step that holds a live, already-asked question.** That step is thrown away and later restarted from scratch. That is a defensible call (an interrupt is a control instruction, not something to make the user sit behind), but it is a real tradeoff the PR description does not own, and it has a concrete consequence — see the major finding on tool-call replay below.

3. **The failed-step priority change is a second, undocumented behavior change.** Failure detection now scans the whole completed batch instead of returning on the first `done`-iteration hit, so a failure deterministically outranks a completed waiting/interrupted result in the same batch. This was raised and fixed inline (f95afff0d) with two dedicated regression tests, so the code is fine — but the title and description still frame the PR purely as a waiting/interrupted tie-break fix. Please fold this into the description so a future reader/bisector knows failure ordering also moved here.

4. **`clarification_superseded_step_ids` has no consumer yet.** The author has already stated in this thread that this is intentional forward groundwork with a named follow-up consumer, and that nothing user-facing surfaces "your other question was superseded" today — so this is not a new finding. Restating it only for scope clarity: this PR fixes the *backend routing* half of the problem. The *user-visible* half — a superseded question left rendered in the chat with no indication it is dead — is not addressed here.

## Findings

### Major

**M1. Invalidated steps restart from scratch and will re-execute tool calls** — `src/xagent/core/agent/pattern/dag/dag.py:2011`
When a losing "waiting" step is invalidated, `_clear_active_step()` discards its ReAct pattern state and child context. When the step is later offered as ready work again, `active_pattern_state` is `None`, so there is no `load_state()` — the step reruns from a brand-new context. Any tool calls it already executed before losing the tie-break **will be executed a second time**; nothing memoizes or checkpoints them. For non-idempotent tools (file writes, outbound API calls, messages) this is a real duplication risk. The PR description only frames the loss as a "duplicate question" problem. `test_dag_pattern_invalidated_step_is_rescheduled_after_winner_settles` (tests/core/agent/test_dag.py:2201) pins the context-discard mechanics but runs with `tools=[]` and no tool calls, so it cannot demonstrate replay safety either way. Please either state this as a known limitation in the description, or add a test with a tool-using/stateful step so the replay behavior is visible and deliberate.

**M2. The behavior change to failed-step priority is undocumented** — see design note (3) above. The code and tests are fine; the gap is the PR description.

### Minor

**m1. Warning text is wrong when the winner is `interrupted`** — `dag.py:821`
"only `%s`'s question is delivered, the rest are invalidated" is factually false in a mixed batch where an `interrupted` result wins: no question was delivered at all, because the winner isn't asking anything. This path is reachable and already exercised by `test_dag_pattern_mixed_batch_interrupted_wins_over_waiting`. Condition the message text on the winner's actual status.

**m2. `on_dag_step_end` is never emitted for `interrupted` / `waiting_for_user` / invalidated steps** — `dag.py:1050`
The lifecycle callback fires on the exception, failure, and normal-success paths only. Consequence: an invalidated step that later restarts re-emits `on_dag_step_start` for the same step id with no matching prior `on_dag_step_end`, so any consumer expecting paired start/end events per step id sees an orphaned start. (Backend-only observation; frontend handling not evaluated.) Either emit `on_dag_step_end` with a distinguishing reason on these branches, or document that these statuses are intentionally exempt from the pairing contract.

**m3. The winner sort key is broader than its docstring** — `dag.py:1967`
The key is `0 if status != "waiting_for_user" else 1`, i.e. *any* non-`waiting_for_user` status outranks `waiting_for_user` — not specifically `interrupted`, as the docstring two lines up claims. Harmless today since only two statuses reach here, but a future third completion status would silently win without anyone deciding that is correct. Prefer an explicit rank map (e.g. `{"interrupted": 0, "waiting_for_user": 1}[status]`, or a `.get()` with a loud fallback) so an unrecognized status fails visibly rather than defaulting to "wins".

**m4. Missing sibling steps are reported as invalidated** — `dag.py:2023`
`invalidated_step_ids.append(...)` runs unconditionally, including when `steps_by_id.get(sibling_step_id)` returned `None` and no status flip actually happened. The method's own docstring promises the return value contains "the ids that got the `clarification_invalidated` flip". Move the append inside the `if sibling_step is not None:` branch.

**m5. Checkpoint label claims a cancellation that may not have occurred** — `dag.py:842`
The label is `"dag_after_cancelled_siblings"` whenever `pending or superseded_step_ids` is truthy — including the supersede-only case where `pending` is empty and nothing was cancelled. A consumer keyed off the label alone would infer a cancellation happened. Either make the label status-aware, or add an explicit `"cancelled": bool(pending)` to the metadata dict (which is right there) so consumers don't have to infer it.

**m6. Test coverage gaps** — `tests/core/agent/test_dag.py:2516`
Two gaps, from an inventory of the file's multi-step batch tests: (a) no test drives a 3+-step batch with genuinely mixed statuses (e.g. 2 `interrupted` + 1 `waiting`) — the existing three-way tests are all-waiting, or waiting+waiting+cancelled-running; (b) no test drives a real `DAGPattern.run()` through an invalidation and then triggers `_generate_plan(replan=True)` in the same run, to confirm a `clarification_invalidated` step doesn't confuse replanning. The only `replan=True` usage is an unrelated standalone `LLMPlanGenerator` unit test with a hand-built previous plan.

**m7. `_select_winner`'s ordering is only tested indirectly** — `tests/core/agent/test_dag.py:2020`
`test_dag_pattern_concurrent_double_waiting_winner_is_deterministic` and its triple-step variant are the only coverage of the tie-break, and both reach it through an asyncio-scheduling-dependent fixture (steps parked on a shared `asyncio.Event`, released together). A direct unit test calling `_select_winner` with a hand-built, shuffled `completed_results` list would pin the sort independent of async timing — strictly stronger and faster than the current coverage.

**m8. Adapter promotion test covers only the waiting case** — `tests/core/agent/test_execution_adapter.py:491`
`test_execution_adapter_surfaces_superseded_step_ids_at_top_level` only exercises `status == "waiting_for_user"`. The `interrupted` case — where `_normalize_result` silently drops the field — is untested, which is exactly the branch where the drop matters.

## Already covered in earlier rounds

Three things below were raised and settled earlier in this PR's review history; noting them for completeness only, not as new action items:

- **Adapter drops `clarification_superseded_step_ids` for `interrupted` winners.** `execution_adapter.py` only promotes the field inside the `status == "waiting_for_user"` branch, so in a mixed batch — precisely the case where a real, live question was discarded — the information never leaves the adapter. This was discussed under the interrupted-winner promotion topic and deferred to the follow-up. Re-flagging in one line because it compounds with the field currently being write-only: today nothing reads it, and in the interrupted case nothing even receives it.
- **The `has_live_step_tasks()` runner guard is unreachable for both currently-registered patterns** (`DAGPattern` always clears the set in a `finally`; `AutoPattern` doesn't implement the method). Already flagged and accepted as a deliberate defensive tripwire for a hypothetical future pattern — not re-raising.
- **`self.status` is not resynced to the winner's status after `_select_winner`,** so `self.status` / `get_state()["status"]` / the checkpoint can disagree with the returned result in a mixed batch, potentially causing a spurious full replan on the next user message. This is **pre-existing on the base commit (721fffe), not a regression from this PR**, and is tracked in #1284. Mentioning it only because it is directly adjacent to what this PR is trying to achieve: determinism of the returned result is fixed here, but the checkpoint/replan path still reads a status that can disagree with it.

- L1044: `redundant-copy` `attributed_result = dict(result)` is an intermediate copy that is re-spread via `**attributed_result` a few lines later, with no mutation in between beyond one conditional key overwrite. Build the returned dict directly from `**result` and overwrite `clarification_draft` conditionally in the same literal.
- L1917 (tests): `duplicate-fixtures` `ConcurrentWaitingLLM`, `WaitingAndSlowLLM` and `TwoWaitingAndSlowLLM` each hand-roll a variant of "some steps ask a question gated on a shared release event, one step hangs until cancelled" — and `TwoWaitingAndSlowLLM` is literally the union of the other two's branches. Replace with a single parameterized fake taking `ask_step_ids` / `hang_step_ids` constructor args.

net: -35 lines possible

## Disposition

**REQUEST_CHANGES.** The determinism fix itself is good work and I'd be happy to approve it once two things land: the tool-call replay consequence of invalidation (M1) is either tested or explicitly documented as a known limitation, and the PR description covers the failed-step priority change (M2). Everything else above is minor and can ride along or follow up.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Took the documentation option. The description now states the consequence directly in the paragraph about the rerun: because the next attempt begins from a fresh context, tool calls the step already executed before losing the tie-break run again — for non-idempotent tools (file writes, outbound API calls, messages) that means duplicated side effects, and nothing memoizes or checkpoints completed tool calls today. Reproduced the replay locally with a tool-using step to confirm the mechanism before writing it down: the loser's rerun starts with only the system prompt and step instruction, so the tool executes a second time.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Fixed in 1e5f674bd — the message now branches on the winner's status: a waiting winner keeps the delivery wording, an interrupted winner logs that the superseded questions were discarded because the interrupt takes precedence. The mixed-batch test now pins the interrupted wording via caplog; reverting to the single hardcoded message makes exactly that assertion fail.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: This predates the PR: on the base commit the callback fires on the same three exits only (exception, failure, success), so every resumed waiting step already produced a second start with no prior end — invalidation adds one more route to an existing gap, it does not create the pairing hole. The only consumer chain today (trace → websocket handlers → chat UI plan-sync) upserts by step id and does not rely on pairing. Since closing it properly means changing the emitter contract and touching the trace/websocket layers, it is out of scope here; filed #1302 to decide between emitting an end with a reason on these branches or documenting the exemption.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Fixed in b588f926f — the key is now an explicit module-level rank map ({"interrupted": 0, "waiting_for_user": 1}); an unrecognized status ranks last and logs an error instead of silently tying with interrupted. Chose rank-last-plus-loud-log over raising because this runs inside the wait loop: raising would convert an unregistered-status coding error into a whole-task failure via the cancellation path, while ranking last preserves delivery of an already-asked question. A direct unit test pins the full order including an unknown status; under the old boolean key that test fails with the unknown status tying with interrupted.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Fixed in b588f926f — the append now sits inside the `if sibling_step is not None:` branch, so the return value matches the docstring's promise. For the record: the None branch is currently unreachable (steps_by_id and the task map are written in lockstep in schedule_ready_steps, entries are never removed), so this aligns the contract rather than fixing an observable bug — the commit message says so explicitly.

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Declining this one on two facts. First, no production code keys off checkpoint labels — a repo-wide search finds this label only at its write site and in test assertions, so the hypothetical consumer doesn't exist yet. Second, the metadata dict already distinguishes the two cases: cancelled_step_ids is derived from pending and is empty exactly when nothing was cancelled, which is the same information the proposed "cancelled" boolean would carry. Happy to revisit the label name if a label-keyed consumer ever appears.

tests/core/agent/test_dag.py
AlexLiu190625: (a) Added in dfbb9a001: a three-step batch with two interrupted steps and one waiting — the smaller-id interrupted step wins, the superseded list contains exactly the waiting step, the waiting loser is invalidated with cleared bookkeeping, and the losing interrupted step keeps its bookkeeping and status. Under the `[1:2]` off-by-one mutation this test and the existing three-way-waiting test both go red.

(b) is structurally unreachable in one run: invalidation only happens on the winner path, which always returns the winner's result, and `run()` returns as soon as a step result is non-None — the replan check is only reachable when the completed batch is empty. Across runs the status also cannot confuse a replan: `_generate_plan(replan=True)` starts by clearing all active steps and replaces the plan wholesale, so `clarification_invalidated` never survives into a new plan; the only place it reaches is the serialized previous plan in the replan prompt.

tests/core/agent/test_dag.py
AlexLiu190625: Added in 9462019e7 — a synchronous unit test calls `_select_winner` directly with placeholder task objects and a hand-scrambled mixed-status list, asserting the full sorted order in one shot, no event loop involved. It also carries the unknown-status case that pins the rank-map change.

tests/core/agent/test_execution_adapter.py
AlexLiu190625: This is the test half of the interrupted-winner promotion discussed on the adapter thread — both halves travel together in the follow-up change on the clarification-resolution path, since adding the test now would pin the dropped-field behavior as expected and the follow-up would immediately have to flip that assertion. The existing completed-status case already pins the mechanism that promotion happens only in the waiting branch.

AlexLiu190625:
Round addressed at head dfbb9a001 (four commits) plus a description rewrite:

- **M1 / M2**: took the documentation option for both. The description now owns the full precedence including the failed-step check running before winner selection, the batch-boundary scope of the determinism, the interrupt-over-question tradeoff, the tool-call replay consequence of invalidation, and the fact that `clarification_superseded_step_ids` has no consumer yet.
- **Fixed**: the superseded-question log now describes an interrupt winner truthfully (1e5f674bd, caplog-pinned); the winner ordering uses an explicit rank map with unknown statuses ranking last and logging loudly (b588f926f); the invalidated-ids return value now matches its docstring (same commit); `_select_winner` has a direct, event-loop-free unit test (9462019e7); a genuinely mixed three-step batch is pinned (dfbb9a001).
- **Deferred**: the start/end pairing gap predates this PR and spans the emitter and trace/websocket layers — filed #1302 (inline thread has the analysis).
- **Declined with rationale on the inline threads**: the checkpoint-label rename (no label-keyed consumer exists; `cancelled_step_ids` already distinguishes the cases), the same-run invalidation-then-replan test (structurally unreachable — details inline), the adapter interrupted-case test (travels with the deferred promotion so its assertion is written once, not flipped), and both simplification suggestions (the copy trades readability for one line; the three fakes encode deliberately different synchronization shapes and merging them would hide that behind constructor flags).

src/xagent/core/agent/pattern/dag/dag.py
```diff
@@ -1821,6 +1953,132 @@ def _tool_name(self, tool: Any) -> str:
...
+            # key, which is exactly the "unranked" case below.
+            status_key = str(status)
+            if status_key not in _COMPLETED_RESULT_RANK:
+                # This runs inside the asyncio.wait() loop in
+                # _execute_ready_steps(); raising here would propagate
+                # through the BaseException cancellation handler and fail
+                # the entire batch over one unrecognized status string.
+                # Sorting it last instead preserves the invariant that a
+                # question already asked is delivered whenever possible.
+                logger.error(
```
rogercloud: Minor: this `logger.error` for an unranked result status is the same class of handled anomaly that `4f752f4f` just downgraded from `error` to `warning` at line 850 (see the accepted inline comment on that change), and `plan_generator.py` logs its handled anomalies at `warning` (lines 351/363/378). An unranked status is deliberately handled here — it sorts last and the question is still delivered. Suggest `logger.error` → `logger.warning`.

tests/core/agent/test_execution_adapter.py
rogercloud, line 491: Still open from the previous review (H17): this test only exercises a `waiting_for_user` result carrying `clarification_superseded_step_ids`. There is still no test for an interrupted winner that carries superseded ids (the case where the winner itself delivers no question). Please pin that combination, as previously requested.

tests/core/agent/test_dag.py
```diff
@@ -1670,6 +1674,1364 @@ async def chat(self, **kwargs: Any) -> dict[str, Any]:
...
+        "a_interrupt",
+        "b_wait",
+        "c_unknown",
+    ]
+    assert "unranked result status" in caplog.text
+    assert "c_unknown" in caplog.text
+
+
+@pytest.mark.asyncio
+async def test_dag_pattern_delivered_result_outranks_pending_replan_in_same_batch() -> (
```
rogercloud, line 2705: Remaining gap from the previous review (H15 gap b): there is still no test for a losing step invalidated by a superseding winner and then replanned (`replan=True`) within the same run — the interplay between invalidation bookkeeping and a concurrent replan request. The tests nearby pin delivered-result-vs-replan priority and three-way invalidation separately; the combination is untested.

rogercloud (approved):
Approving per the disposition above. Remaining items are minor (one log-level nit N1, two test-coverage gaps H15 gap b / H17, and the untracked H3 waiver — recommend opening a tracking issue so the deferral is not lost).

src/xagent/core/agent/pattern/dag/dag.py
AlexLiu190625: Done in bad6e83a8 — the unranked-status log is now a warning, and the two docstrings that said "loudly"/"an error" were aligned in the same commit. Agreed with the framing: the branch fully handles the anomaly (ranks last, the question is still delivered), so it belongs on the handled-anomaly channel; the error channel should keep meaning "a function did not complete".

tests/core/agent/test_dag.py
AlexLiu190625: Added in 6114392d6, via the path that is actually reachable. Within one run the combination cannot occur: invalidation happens only on the winner path, which always returns the winner's result, and — as noted in the earlier review of the relocated checks — the replan check is unconditionally unreachable whenever completed_results is non-empty, which is precisely the precondition for invalidation. Across runs is where the two meet, so the new test drives `_generate_plan(replan=True)` against a plan carrying a `clarification_invalidated` step and pins the three facts that matter: the replanner sees the status in the serialized previous plan, the regenerated plan comes back all-pending (the status cannot leak into it), and the active-step bookkeeping is cleared. Each assertion goes red under a targeted mutation.

tests/core/agent/test_execution_adapter.py
AlexLiu190625: Done in 18c1cd059 — the interrupted branch now promotes `clarification_superseded_step_ids` with the same shape and the same `or []` default as the waiting branch, and the new adapter test pins exactly the combination requested here: an interrupted winner carrying superseded ids. Reverting the promotion makes that test fail on the missing key while the existing completed-absence assertion stays green.

src/xagent/core/agent/execution_adapter.py
AlexLiu190625: Landed in 18c1cd059 after all: the promotion now covers interrupted winners with the same `or []` default as the waiting branch, plus the interrupted-with-superseded-ids adapter test suggested here. The completed case keeps the key absent, pinned by the existing assertion.

AlexLiu190625:
All four remaining items are addressed at head 18c1cd059:

- **Unranked-status log level**: warning now (bad6e83a8), consistent with the superseded-question log and the handled-anomaly convention.
- **Invalidation vs replan**: pinned in 6114392d6 via the reachable path — within one run the combination cannot occur (invalidation happens only on the winner path, which always returns; the replan check is unreachable while completed_results is non-empty), so the test drives a replan against a plan carrying an invalidated step and asserts the replanner sees it, the regenerated plan is all-pending, and bookkeeping is cleared.
- **Adapter promotion for interrupted winners**: landed in 18c1cd059 together with the interrupted-with-superseded-ids test — the promotion now covers both statuses that can carry the key, with the same empty-list default, and the completed case keeps the key absent.
- **Cancellation-exit waiver**: tracked in #1311, and the handler now carries a comment stating the exemption and pointing at that issue (7dd0d7997).

All review threads are resolved.

Issue #1203, opened 2026-08-08
Harden catalog auth classification and the hidden-app connect gate


yiboyasss:
Follow-up items from #1179's review rounds (see the reviews for full analysis). All predate that PR or change pre-existing semantics, so they were deliberately kept out of it.

## 1. `classify_app_auth`'s `api_key` branch has no transport constraint, and the provisioner hardcodes stdio (review MAJOR-1)

`src/xagent/web/mcp_apps.py`: the `api_key` branch is `required_env and command` with no transport check, while the newer `keyless` branch requires `transport == "stdio"`. Verified: `classify_app_auth("streamable_http", {"command": "npx", "required_env": ["K"]})` returns `"api_key"`, admin validation accepts it, and `_ensure_catalog_app_server` then provisions the row with a hardcoded `transport="stdio"` — silently discarding the declared remote transport. An admin authoring a remote connector with a stray `command` gets a misprovisioned local stdio server instead of a 422. Today's actual behavior is pinned by `test_key_based_is_not_transport_gated_unlike_keyless` so this issue fails loudly if the contract changes without this item being addressed.

Fix direction: apply the same stdio-only constraint to `api_key` (auditing existing rows first), or document the permissiveness as deliberate with a boundary test.

## 2. The hidden-app connect gate misses the `builtin_oauth` provider-redirect flow (review MINOR-1)

`_reject_hidden_catalog_app` is wired into the api_key/keyless and remote-MCP OAuth connect paths only. The `builtin_oauth` flow (`auth.py` `generic_oauth_callback` -> `_ensure_user_mcp_server`) never runs it, so hiding an oauth-transport app would not block new connections through the redirect flow. Latent today — every oauth-transport catalog row ships visible — but it breaks the hidden-rollout pattern #1179 established the moment it is reused for an OAuth app: `is_visible_in_connector` is freely admin-mutable via PATCH, so the gap reopens the moment an operator hides an existing builtin_oauth row, with no code change. The gate's docstring documents this scope honestly and now warns explicitly not to reuse the hidden-rollout idiom on an oauth-transport row until this closes. Fix: enforce the same check in `_ensure_user_mcp_server`.

## 3. The gate blocks reconnect/key-rotation on hidden apps for already-connected users (review MINOR-2)

[...]

Issue #1143, opened 2026-08-05
feat: Add Chrome connector


yiboyasss:
### Description

Add a new Chrome Connector to allow agents to interact with browser-based workflows and access information from Chrome sessions.

Users should be able to connect Chrome and enable agents to perform browser-related tasks through the connector.

PR #1293, opened 2026-08-12
feat: add Intercom OAuth connector

yiboyasss:
## Summary
- Adds Intercom as a built-in MCP connector (oauth_providers + public_mcp_apps catalog entry), following the existing HubSpot/Slack/Zoom pattern.
- New FastMCP tool module (`xagent.web.tools.mcp.intercom`) covering contact search/lookup, conversation listing/reading/replying, internal notes, and closing conversations, backed directly by Intercom's REST API.
- Normalizes Intercom's non-standard OAuth token response (`{"token": ...}` instead of `{"access_token": ...}`) in the generic OAuth callback.
- Uses the region-less `api.intercom.io` / `app.intercom.com` hosts, which auto-route per workspace — one connector config covers US/EU/AU-hosted workspaces, unlike Intercom's hosted MCP server (which has no AU support).
- Adds a seed migration for the new provider/app rows and documents the required `INTERCOM_CLIENT_ID`/`INTERCOM_CLIENT_SECRET`/`INTERCOM_REDIRECT_URI` env vars in `example.env`.

## Test plan
- [x] `ruff check` / `ruff format` / `mypy` pass on all changed files
- [x] `alembic upgrade head` / `downgrade -1` / `upgrade head` cycle verified against a local dev DB, seeding and cleanly removing both rows
- [x] `validate_builtin_public_mcp_apps` reports no drift after seeding
- [x] Verified end-to-end locally via ngrok: OAuth login/callback correctly resolves the Intercom provider and exchanges the token
- [ ] Manual verification of the individual Intercom MCP tools (search/reply/etc.) against a live workspace

### DIFF
--- a/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
+++ b/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
@@ -1,7 +1,7 @@
 """seed built-in Intercom (OAuth) MCP connector
 
 Revision ID: 20260812_seed_intercom_mcp_app
-Revises: 20260809_add_task_interaction_requests
+Revises: 20260810_add_task_interaction_protocol_version
 Create Date: 2026-08-12 00:00:00.000000
 
 """
@@ -14,7 +14,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "20260812_seed_intercom_mcp_app"
-down_revision: Union[str, None] = "20260809_add_task_interaction_requests"
+down_revision: Union[str, None] = "20260810_add_task_interaction_protocol_version"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
+++ b/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
@@ -60,11 +60,17 @@ def _intercom_provider_row() -> dict[str, object]:
         "name": "Intercom",
         "client_id": os.environ.get("INTERCOM_CLIENT_ID", ""),
         "client_secret": os.environ.get("INTERCOM_CLIENT_SECRET", ""),
-        # Single global authorize host: once a public app clears Intercom App
-        # Review, Intercom replicates it across US/EU/AU automatically.
+        # Single global authorize host: per Intercom's community answer
+        # (community.intercom.com/api-webhooks-23/do-we-need-to-create
+        # -multiple-oauth-apps-per-data-region-2522), a public app is
+        # replicated across US/EU/AU automatically once it clears Intercom
+        # App Review. See builtin_mcp_registry.py's intercom provider row for
+        # the full citation and the caveat that this is a community answer,
+        # not primary docs, unverified against a live non-US workspace.
         "auth_url": "https://app.intercom.com/oauth",
         # No region prefix: api.intercom.io auto-routes to the workspace's
-        # actual hosting region (US/EU/AU), unlike Intercom's hosted MCP
+        # actual hosting region (US/EU/AU) -- this part is documented in
+        # Intercom's primary REST API reference, unlike Intercom's hosted MCP
         # server (mcp.intercom.com / mcp.eu.intercom.com) which has separate
         # per-region endpoints and does not support AU workspaces at all.
         "token_url": "https://api.intercom.io/auth/eagle/token",
@@ -89,7 +95,10 @@ def _intercom_app_row() -> dict[str, object]:
         "provider_name": "intercom",
         "category": "Support",
         "oauth_scopes": [],
-        "is_visible_in_connector": True,
+        # Hidden until manually verified against a live workspace -- see
+        # builtin_mcp_registry.py's intercom app row for the full rationale
+        # (this ships customer-facing write tools: reply/note/close).
+        "is_visible_in_connector": False,
         "launch_config": {
             "command": "python",
             "args": ["-m", "xagent.web.tools.mcp.intercom"],
@@ -159,13 +168,20 @@ def downgrade() -> None:
         if remaining_intercom_apps:
             return
 
-    # Delete unconditionally by provider_name, matching the sibling seed
-    # migrations (slack, meta, google-maps). A shape-matching guard would
-    # protect the wrong thing: an admin recreating the *real* Intercom
-    # provider enters Intercom's canonical URLs, which would match the guard
-    # and be deleted anyway -- only a differently-shaped row would survive.
+    # Only delete the provider row when it still matches the static shape this
+    # migration seeded, so an admin-created "intercom" provider (via
+    # POST /admin/mcp/providers) is preserved -- same guard as the zoom seed
+    # migration, and for the same reason: unlike slack/meta/google-maps (whose
+    # downgrades delete unconditionally), a differently-shaped custom row here
+    # must survive. client_id/client_secret are env-dependent and
+    # intentionally not part of the guard. name/auth_url/token_url are NOT
+    # NULL core columns present since the table's creation, so they can be
+    # matched unconditionally.
+    seeded_provider = _intercom_provider_row()
     bind.execute(
-        sa.delete(FULL_OAUTH_PROVIDERS_TABLE).where(
-            FULL_OAUTH_PROVIDERS_TABLE.c.provider_name == "intercom"
-        )
+        sa.delete(FULL_OAUTH_PROVIDERS_TABLE)
+        .where(FULL_OAUTH_PROVIDERS_TABLE.c.provider_name == "intercom")
+        .where(FULL_OAUTH_PROVIDERS_TABLE.c.name == seeded_provider["name"])
+        .where(FULL_OAUTH_PROVIDERS_TABLE.c.auth_url == seeded_provider["auth_url"])
+        .where(FULL_OAUTH_PROVIDERS_TABLE.c.token_url == seeded_provider["token_url"])
     )
--- a/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
+++ b/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
@@ -1,7 +1,7 @@
 """seed built-in Intercom (OAuth) MCP connector
 
 Revision ID: 20260812_seed_intercom_mcp_app
-Revises: 20260810_add_hubspot_marketing_scopes
+Revises: 20260812_add_slack_history_reactions_files_scopes
 Create Date: 2026-08-12 00:00:00.000000
 
 """
@@ -14,7 +14,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "20260812_seed_intercom_mcp_app"
-down_revision: Union[str, None] = "20260810_add_hubspot_marketing_scopes"
+down_revision: Union[str, None] = "20260812_add_slack_history_reactions_files_scopes"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
+++ b/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
@@ -1,7 +1,7 @@
 """seed built-in Intercom (OAuth) MCP connector
 
 Revision ID: 20260812_seed_intercom_mcp_app
-Revises: 20260810_add_task_interaction_protocol_version
+Revises: 20260810_add_hubspot_marketing_scopes
 Create Date: 2026-08-12 00:00:00.000000
 
 """
@@ -14,7 +14,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "20260812_seed_intercom_mcp_app"
-down_revision: Union[str, None] = "20260810_add_task_interaction_protocol_version"
+down_revision: Union[str, None] = "20260810_add_hubspot_marketing_scopes"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
+++ b/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
@@ -139,6 +139,22 @@ def upgrade() -> None:
                 sa.insert(PUBLIC_MCP_APPS_TABLE),
                 [_filter_row(_intercom_app_row(), app_columns)],
             )
+        else:
+            # A row with this app_id already exists (e.g. hand-created by an
+            # operator before this migration deployed). Such a row keeps its
+            # own is_visible_in_connector, which defaults to TRUE for
+            # hand-created rows -- and the builtin registry overlays the real
+            # transport/launch_config onto ANY row sharing this app_id at read
+            # time, so a visible pre-existing row would silently become a
+            # working, one-click-connectable, unverified write-capable
+            # connector, defeating the hidden-rollout gate with no further
+            # action. Same guard, same reasoning, as the chrome seed
+            # migration's collision branch (#1143).
+            bind.execute(
+                sa.update(PUBLIC_MCP_APPS_TABLE)
+                .where(PUBLIC_MCP_APPS_TABLE.c.app_id == APP_ID)
+                .values(is_visible_in_connector=False)
+            )
 
 
 def downgrade() -> None:
--- a/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
+++ b/src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
@@ -1,0 +1,171 @@
+"""seed built-in Intercom (OAuth) MCP connector
+
+Revision ID: 20260812_seed_intercom_mcp_app
+Revises: 20260809_add_task_interaction_requests
+Create Date: 2026-08-12 00:00:00.000000
+
+"""
+
+import os
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "20260812_seed_intercom_mcp_app"
+down_revision: Union[str, None] = "20260809_add_task_interaction_requests"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+FULL_OAUTH_PROVIDERS_TABLE = sa.table(
+    "oauth_providers",
+    sa.column("provider_name", sa.String),
+    sa.column("name", sa.String),
+    sa.column("client_id", sa.String),
+    sa.column("client_secret", sa.String),
+    sa.column("auth_url", sa.String),
+    sa.column("token_url", sa.String),
+    sa.column("redirect_uri", sa.String),
+    sa.column("userinfo_url", sa.String),
+    sa.column("user_id_path", sa.String),
+    sa.column("email_path", sa.String),
+    sa.column("default_scopes", sa.JSON),
+)
+
+PUBLIC_MCP_APPS_TABLE = sa.table(
+    "public_mcp_apps",
+    sa.column("app_id", sa.String),
+    sa.column("name", sa.String),
+    sa.column("description", sa.Text),
+    sa.column("icon", sa.String),
+    sa.column("transport", sa.String),
+    sa.column("provider_name", sa.String),
+    sa.column("category", sa.String),
+    sa.column("oauth_scopes", sa.JSON),
+    sa.column("is_visible_in_connector", sa.Boolean),
+    sa.column("launch_config", sa.JSON),
+)
+
+APP_ID = "intercom"
+
+
+def _filter_row(row: dict[str, object], allowed_columns: set[str]) -> dict[str, object]:
+    return {key: value for key, value in row.items() if key in allowed_columns}
+
+
+def _intercom_provider_row() -> dict[str, object]:
+    return {
+        "provider_name": "intercom",
+        "name": "Intercom",
+        "client_id": os.environ.get("INTERCOM_CLIENT_ID", ""),
+        "client_secret": os.environ.get("INTERCOM_CLIENT_SECRET", ""),
+        # Single global authorize host: once a public app clears Intercom App
+        # Review, Intercom replicates it across US/EU/AU automatically.
+        "auth_url": "https://app.intercom.com/oauth",
+        # No region prefix: api.intercom.io auto-routes to the workspace's
+        # actual hosting region (US/EU/AU), unlike Intercom's hosted MCP
+        # server (mcp.intercom.com / mcp.eu.intercom.com) which has separate
+        # per-region endpoints and does not support AU workspaces at all.
+        "token_url": "https://api.intercom.io/auth/eagle/token",
+        "redirect_uri": os.environ.get("INTERCOM_REDIRECT_URI", ""),
+        "userinfo_url": "https://api.intercom.io/me",
+        "user_id_path": "id",
+        "email_path": "email",
+        # Intercom has no `scope` authorize-URL param; granted permissions
+        # come entirely from the app's Authentication settings in the
+        # Developer Hub, so there is nothing to list here.
+        "default_scopes": [],
+    }
+
+
+def _intercom_app_row() -> dict[str, object]:
+    return {
+        "app_id": APP_ID,
+        "name": "Intercom",
+        "description": "Connect to Intercom to search contacts, review conversations, and reply to customers.",
+        "icon": "https://www.google.com/s2/favicons?domain=intercom.com&sz=128",
+        "transport": "oauth",
+        "provider_name": "intercom",
+        "category": "Support",
+        "oauth_scopes": [],
+        "is_visible_in_connector": True,
+        "launch_config": {
+            "command": "python",
+            "args": ["-m", "xagent.web.tools.mcp.intercom"],
+            "env_mapping": {"INTERCOM_ACCESS_TOKEN": "access_token"},
+        },
+    }
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    existing_tables = set(inspector.get_table_names())
+
+    if "oauth_providers" in existing_tables:
+        oauth_columns = {
+            column["name"] for column in inspector.get_columns("oauth_providers")
+        }
+        existing_provider_names = set(
+            bind.execute(
+                sa.select(FULL_OAUTH_PROVIDERS_TABLE.c.provider_name)
+            ).scalars()
+        )
+        if "intercom" not in existing_provider_names:
+            bind.execute(
+                sa.insert(FULL_OAUTH_PROVIDERS_TABLE),
+                [_filter_row(_intercom_provider_row(), oauth_columns)],
+            )
+
+    if "public_mcp_apps" in existing_tables:
+        app_columns = {
+            column["name"] for column in inspector.get_columns("public_mcp_apps")
+        }
+        existing_app_ids = set(
+            bind.execute(sa.select(PUBLIC_MCP_APPS_TABLE.c.app_id)).scalars()
+        )
+        if APP_ID not in existing_app_ids:
+            bind.execute(
+                sa.insert(PUBLIC_MCP_APPS_TABLE),
+                [_filter_row(_intercom_app_row(), app_columns)],
+            )
+
+
+def downgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    existing_tables = set(inspector.get_table_names())
+
+    if "public_mcp_apps" in existing_tables:
+        # Only the catalog entry is removed. Any user OAuth connections created
+        # against this app are not owned by this migration and are cleaned up
+        # through the normal disconnect path.
+        bind.execute(
+            sa.delete(PUBLIC_MCP_APPS_TABLE).where(
+                PUBLIC_MCP_APPS_TABLE.c.app_id == APP_ID
+            )
+        )
+
+    if "oauth_providers" not in existing_tables:
+        return
+
+    if "public_mcp_apps" in existing_tables:
+        remaining_intercom_apps = bind.execute(
+            sa.select(sa.func.count())
+            .select_from(PUBLIC_MCP_APPS_TABLE)
+            .where(PUBLIC_MCP_APPS_TABLE.c.provider_name == "intercom")
+        ).scalar()
+        if remaining_intercom_apps:
+            return
+
+    # Delete unconditionally by provider_name, matching the sibling seed
+    # migrations (slack, meta, google-maps). A shape-matching guard would
+    # protect the wrong thing: an admin recreating the *real* Intercom
+    # provider enters Intercom's canonical URLs, which would match the guard
+    # and be deleted anyway -- only a differently-shaped row would survive.
+    bind.execute(
+        sa.delete(FULL_OAUTH_PROVIDERS_TABLE).where(
+            FULL_OAUTH_PROVIDERS_TABLE.c.provider_name == "intercom"
+        )
+    )
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -1470,6 +1470,24 @@ def generic_oauth_callback(
         token_data = _normalize_intercom_token_response(provider, token_data)
         access_token = token_data.get("access_token")
 
+        if not access_token:
+            # Without this, a token response that slips past the "error" in
+            # token_data check above (e.g. Intercom's error.list envelope,
+            # which doesn't use that key) would fall through to the
+            # persistence block below with access_token=None, hit
+            # UserOAuth.access_token's NOT NULL constraint, and surface as a
+            # raw SQLAlchemy IntegrityError message through the generic
+            # exception handler instead of a clear, actionable error.
+            import html
+
+            return HTMLResponse(
+                content=(
+                    "<h1>Error exchanging token</h1>"
+                    f"<p>{html.escape(provider)} did not return an access token.</p>"
+                ),
+                status_code=400,
+            )
+
         provider_user_id = None
         email = None
 
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -1235,6 +1235,25 @@ def generic_oauth_login(
     if app_id:
         app_info = get_app_by_id(db, app_id)
         if app_info:
+            # Reject a hidden app here too, not only at the callback --
+            # otherwise a hidden connector's consent screen is still shown at
+            # the real provider (no security bypass, since the callback
+            # still blocks the connect, but a confusing UX: the app doesn't
+            # feel hidden if you can watch it start connecting). Unlike the
+            # callback's gate, an unknown app_id is deliberately left alone
+            # here too, for the same reason (e.g. gmail's bare-app-id flows).
+            from .mcp import _reject_hidden_catalog_app
+
+            try:
+                _reject_hidden_catalog_app(app_info)
+            except HTTPException:
+                return HTMLResponse(
+                    content=(
+                        "<h1>Cannot Connect</h1>"
+                        "<p>This app is not currently available.</p>"
+                    ),
+                    status_code=404,
+                )
             if "oauth_scopes" in app_info:
                 app_scopes = app_info["oauth_scopes"]
             app_optional_scopes = app_info.get("optional_oauth_scopes") or []
@@ -1451,6 +1470,14 @@ def generic_oauth_callback(
         from ..mcp_apps import get_app_by_id
         from .mcp import _reject_hidden_catalog_app
 
+        # An app_id that resolves to no catalog row at all is left alone,
+        # deliberately not folded into this gate: several existing OAuth
+        # flows (e.g. gmail in the test suite) legitimately reach this
+        # callback with an app_id that has no PublicMCPApp row, and rely on
+        # falling through to a bare/provider-scoped grant rather than being
+        # rejected -- confirmed by running the full oauth test suite before
+        # settling on this narrower check. Only a row that exists AND is
+        # hidden is rejected here.
         target_app_info = get_app_by_id(db, app_id)
         if target_app_info:
             try:
@@ -1597,15 +1624,16 @@ def generic_oauth_callback(
                     + timedelta(seconds=int(token_data["expires_in"])),
                 )
 
-            from ..mcp_apps import get_all_mcp_apps, get_app_by_id
+            from ..mcp_apps import get_all_mcp_apps
             from .mcp import _reject_hidden_catalog_app
 
             if app_id:
-                app_info = get_app_by_id(db, app_id)
+                # Reuse target_app_info from the earlier hidden-app-gate
+                # check above rather than re-fetching by app_id: nothing
+                # mutates public_mcp_apps between there and here, so a second
+                # fetch would just be redundant, not more correct.
+                app_info = target_app_info
                 if app_info:
-                    # The hidden-app gate for this branch already ran earlier
-                    # (right after app_id was decoded from state, before the
-                    # token exchange) -- not repeated here.
                     # A stale/crafted app_id in the OAuth state can point at a
                     # non-oauth app. Fail with a clear error instead of a generic
                     # 500 after the user already completed provider consent —
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -281,6 +281,24 @@ def _normalize_intercom_token_response(
     return {**token_data, "access_token": token}
 
 
+def _extract_provider_error_message(token_data: dict[str, Any]) -> str | None:
+    """Best-effort human-readable detail for a token exchange that yielded no
+    access_token.
+
+    Standard OAuth2 providers surface `error`/`error_description`, already
+    handled by the `"error" in token_data` check earlier in the callback.
+    This covers Intercom's differently-shaped `error.list` envelope instead
+    (`{"type": "error.list", "errors": [{"message": "..."}]}`), which does
+    not use an `error` key and so slips past that earlier check.
+    """
+    errors = token_data.get("errors")
+    if isinstance(errors, list) and errors:
+        first_error = errors[0]
+        if isinstance(first_error, dict) and first_error.get("message"):
+            return str(first_error["message"])
+    return None
+
+
 def create_access_token(
     data: Dict[str, Any], expires_delta: Optional[timedelta] = None
 ) -> str:
@@ -1414,6 +1432,38 @@ def generic_oauth_callback(
     user_id = payload.get("user_id")
     app_id = payload.get("app_id")
 
+    if app_id:
+        # Reject a hidden app before spending the authorization code against
+        # the real provider, not just before persisting -- is_visible_in_
+        # connector is also used as a release gate (e.g. an unverified
+        # builtin_oauth connector shipping hidden), and this builtin_oauth
+        # provider-redirect flow used to be the one connect path
+        # _reject_hidden_catalog_app's own docstring flagged as NOT enforcing
+        # that gate (#1203): an authenticated user who knew (or guessed) the
+        # app_id could still connect a hidden connector. Checked here, ahead
+        # of the token exchange below, rather than only later next to
+        # _ensure_user_mcp_server, so a hidden app never even reaches the
+        # provider -- matching the "indistinguishable from nonexistent" intent
+        # the helper already documents for its other two call sites. The bare
+        # (app_id-less) batch branch below cannot move this early: it doesn't
+        # know which apps share the provider until it queries them, so its
+        # own per-app check stays where it is.
+        from ..mcp_apps import get_app_by_id
+        from .mcp import _reject_hidden_catalog_app
+
+        target_app_info = get_app_by_id(db, app_id)
+        if target_app_info:
+            try:
+                _reject_hidden_catalog_app(target_app_info)
+            except HTTPException:
+                return HTMLResponse(
+                    content=(
+                        "<h1>Cannot Connect</h1>"
+                        "<p>This app is not currently available.</p>"
+                    ),
+                    status_code=404,
+                )
+
     if not db_provider:
         return HTMLResponse(
             content="<h1>Error: Provider not configured</h1>", status_code=500
@@ -1480,11 +1530,12 @@ def generic_oauth_callback(
             # exception handler instead of a clear, actionable error.
             import html
 
+            message = f"{html.escape(provider)} did not return an access token."
+            detail = _extract_provider_error_message(token_data)
+            if detail:
+                message = f"{message} {html.escape(detail)}"
             return HTMLResponse(
-                content=(
-                    "<h1>Error exchanging token</h1>"
-                    f"<p>{html.escape(provider)} did not return an access token.</p>"
-                ),
+                content=f"<h1>Error exchanging token</h1><p>{message}</p>",
                 status_code=400,
             )
 
@@ -1547,10 +1598,14 @@ def generic_oauth_callback(
                 )
 
             from ..mcp_apps import get_all_mcp_apps, get_app_by_id
+            from .mcp import _reject_hidden_catalog_app
 
             if app_id:
                 app_info = get_app_by_id(db, app_id)
                 if app_info:
+                    # The hidden-app gate for this branch already ran earlier
+                    # (right after app_id was decoded from state, before the
+                    # token exchange) -- not repeated here.
                     # A stale/crafted app_id in the OAuth state can point at a
                     # non-oauth app. Fail with a clear error instead of a generic
                     # 500 after the user already completed provider consent —
@@ -1575,6 +1630,21 @@ def generic_oauth_callback(
                     if app.get("provider") == provider
                 ]
                 for app_info in apps:
+                    # Same release-gate enforcement as the single-app branch
+                    # above (same shared helper, so there is one source of
+                    # truth for this check), but a hidden app must not abort
+                    # the whole bare batch connect — skip it and keep
+                    # connecting the other visible apps under the same
+                    # provider, matching the mis-tagged-app skip below.
+                    try:
+                        _reject_hidden_catalog_app(app_info)
+                    except HTTPException:
+                        logger.info(
+                            "Skipping hidden app %s during bare %s OAuth batch connect",
+                            app_info.get("id"),
+                            provider,
+                        )
+                        continue
                     # This bare app_id-less login only ever requests
                     # db_provider.default_scopes (see the app_scopes=None
                     # branch above), never an app's own oauth_scopes. Creating
--- a/src/xagent/web/api/auth.py
+++ b/src/xagent/web/api/auth.py
@@ -260,6 +260,27 @@ def _exchange_meta_long_lived_token(
     return {**token_data, **long_lived_token_data}
 
 
+def _normalize_intercom_token_response(
+    provider: str, token_data: dict[str, Any]
+) -> dict[str, Any]:
+    """Map Intercom's `{"token": ...}` token response onto `access_token`.
+
+    Intercom's token endpoint (POST /auth/eagle/token) does not follow the
+    standard OAuth2 token response shape: it returns only `{"token": "..."}`,
+    with no `access_token`, `token_type`, or `expires_in` fields. Without this,
+    generic_oauth_callback's `token_data.get("access_token")` would silently
+    resolve to None and the connection would be persisted with no token.
+    """
+    if provider.lower() != "intercom":
+        return token_data
+    if "access_token" in token_data:
+        return token_data
+    token = token_data.get("token")
+    if not token:
+        return token_data
+    return {**token_data, "access_token": token}
+
+
 def create_access_token(
     data: Dict[str, Any], expires_delta: Optional[timedelta] = None
 ) -> str:
@@ -1446,6 +1467,7 @@ def generic_oauth_callback(
         token_data = _exchange_meta_long_lived_token(
             provider, token_url, token_data, client_id, client_secret
         )
+        token_data = _normalize_intercom_token_response(provider, token_data)
         access_token = token_data.get("access_token")
 
         provider_user_id = None
--- a/src/xagent/web/builtin_mcp_registry.py
+++ b/src/xagent/web/builtin_mcp_registry.py
@@ -165,6 +165,37 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
             # kept in sync.
             "default_scopes": ["user:read:user"],
         },
+        {
+            "provider_name": "intercom",
+            "name": "Intercom",
+            "client_id": os.environ.get("INTERCOM_CLIENT_ID", ""),
+            "client_secret": os.environ.get("INTERCOM_CLIENT_SECRET", ""),
+            # The single "app.intercom.com" authorize host serves every
+            # region: once a public app clears Intercom App Review, Intercom
+            # replicates it across US/EU/AU automatically, so there is no
+            # per-region app or authorize URL to manage.
+            "auth_url": "https://app.intercom.com/oauth",
+            # "api.intercom.io" (no region prefix) auto-routes each request
+            # to the workspace's actual hosting region (US/EU/AU) based on
+            # the token/workspace being acted on -- this is what lets one
+            # provider row, unlike the hosted MCP server's separate
+            # mcp.intercom.com / mcp.eu.intercom.com endpoints (which don't
+            # support AU workspaces at all), cover all three regions.
+            "token_url": "https://api.intercom.io/auth/eagle/token",
+            "redirect_uri": os.environ.get("INTERCOM_REDIRECT_URI", ""),
+            "userinfo_url": "https://api.intercom.io/me",
+            "user_id_path": "id",
+            "email_path": "email",
+            # Intercom has no `scope` authorize-URL param at all -- granted
+            # permissions come entirely from the app's Authentication
+            # settings in the Developer Hub, the same out-of-band model as
+            # meta's config_id above. Leaving this empty means
+            # generic_oauth_login's `elif scope_str:` guard (api/auth.py)
+            # never adds a scope param for this provider, which matches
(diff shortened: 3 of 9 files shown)


rogercloud:
ci is failing

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+    Returns at most `limit` conversations (max 100), most recently updated first.
+    """
+    try:
+        valid_states = {"open", "closed", "all"}
+        if state not in valid_states:
+            raise ValueError(f"state must be one of {sorted(valid_states)}")
+
+        body: dict[str, Any] = {
+            "pagination": {"per_page": max(1, min(limit, 100))},
+            "sort": {"field": "updated_at", "order": "descending"},
```
rogercloud, line 237: **Major: `state="all"` sends a `/conversations/search` request with no `query`, which will very likely 400.**

The docstring advertises `"open"`, `"closed"`, and `"all"`, but the `"all"` branch drops the `query` field entirely:

```python
body: dict[str, Any] = {
    "pagination": {"per_page": max(1, min(limit, 100))},
    "sort": {"field": "updated_at", "order": "descending"},
}
if state != "all":
    body["query"] = {"field": "state", "operator": "=", "value": state}
result = _request("POST", "/conversations/search", body=body)
```

Per Intercom's REST API docs (v2.4, Search for Conversations), `query` is a required field on this endpoint; unfiltered listing is documented against a separate `GET /conversations`. So the one value a caller would reach for to "just list everything" is the one that fails.

Suggested fix: route `state == "all"` to `GET /conversations` (with the same pagination), or supply an always-true query predicate so the search endpoint stays valid. Either way, please add a unit test covering all three documented `state` values.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+        return _error(str(e))
+
+
+@mcp.tool()
+def intercom_get_contact(contact_id: str) -> str:
+    """
+    Get an Intercom contact by id.
+    """
+    try:
+        contact = _request("GET", f"/contacts/{contact_id}")
```
rogercloud: **Moderate: unvalidated model-supplied id interpolated into the URL path (path/query injection).**

`contact_id` reaches the f-string unescaped. This is exploitable rather than merely malformed:

```python
requests.Request('GET', url='https://api.intercom.io/contacts/../admins').prepare().url
# -> 'https://api.intercom.io/admins'
requests.Request('GET', url='https://api.intercom.io/contacts/123?display_as=plaintext').prepare().url
# -> query string injected
```

A `contact_id` of `"../admins"` redirects a "get one contact" call to `GET /admins` under the same bearer token. Since these ids can originate from content the agent read (e.g. a ticket body), this is a genuine confused-deputy path.

Please percent-encode at every interpolation site, e.g.:

```suggestion
        contact = _request("GET", f"/contacts/{quote(contact_id, safe='')}")
```

(add `from urllib.parse import quote` at the top). The same applies at lines ~200, ~229, ~252 and ~274.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+@mcp.tool()
+def intercom_get_conversation(conversation_id: str) -> str:
+    """
+    Get a full Intercom conversation by id, including the message thread
+    (conversation_parts).
+    """
+    try:
+        conversation = _request("GET", f"/conversations/{conversation_id}")
```
rogercloud: Same path-injection issue as on `intercom_get_contact` — `conversation_id` is interpolated raw. Please percent-encode with `quote(conversation_id, safe="")`.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+
+@mcp.tool()
+def intercom_reply_to_conversation(conversation_id: str, body: str) -> str:
+    """
+    Reply to a customer on an Intercom conversation, as the connected admin.
+    """
+    try:
+        reply = _request(
+            "POST",
+            f"/conversations/{conversation_id}/reply",
```
rogercloud: Same path-injection issue — `conversation_id` needs `quote(conversation_id, safe="")` before interpolation. Higher stakes here since this is a write path: a crafted id could redirect the reply POST to a different endpoint the token can reach.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+@mcp.tool()
+def intercom_add_internal_note(conversation_id: str, body: str) -> str:
+    """
+    Add an internal note to an Intercom conversation. Internal notes are only
+    visible to teammates, never to the customer.
+    """
+    try:
+        reply = _request(
+            "POST",
+            f"/conversations/{conversation_id}/reply",
```
rogercloud: Same path-injection issue — please percent-encode `conversation_id` before interpolating it into the internal-note URL.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+
+@mcp.tool()
+def intercom_close_conversation(conversation_id: str) -> str:
+    """
+    Close an Intercom conversation.
+    """
+    try:
+        conversation = _request(
+            "POST",
+            f"/conversations/{conversation_id}/parts",
```
rogercloud: Same path-injection issue — please percent-encode `conversation_id` here as well.

src/xagent/web/api/auth.py
```diff
@@ -1420,6 +1441,7 @@ def generic_oauth_callback(
         token_data = _exchange_meta_long_lived_token(
             provider, token_url, token_data, client_id, client_secret
         )
+        token_data = _normalize_intercom_token_response(provider, token_data)
         access_token = token_data.get("access_token")
```
rogercloud, line 1548: **Minor: the Intercom token-exchange error path yields a confusing 500 instead of a clean failure.**

There's no `raise_for_status()` on the token response, and the earlier `if "error" in token_data:` check doesn't match Intercom's actual error envelope (`{"type": "error.list", "errors": [...]}`). A failed or expired auth code therefore slips past the guard and `access_token` resolves to `None` here.

This does not persist a broken connection — `UserOAuth.access_token` is NOT NULL, so the subsequent `db.commit()` raises `IntegrityError`, which the outer `except Exception` renders as a generic "Authentication Failed" page. But the message is opaque and, because it goes through `html.escape(str(e))`, can surface a raw SQLAlchemy error string to the user.

Suggested fix: add an explicit guard before the persistence block, e.g. `if not access_token: raise ValueError("Intercom token exchange did not return an access token")`, and consider handling the `error.list` envelope shape so the failure reads clearly.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+INTERCOM_BASE_URL = "https://api.intercom.io"
+# Pinned so a future default-version bump on Intercom's side can't silently
+# change response shapes underneath these tools.
+INTERCOM_API_VERSION = "2.11"
+DEFAULT_TIMEOUT_SECONDS = 30
+
+# Keyed by access token, not a single process-global value: this module is
+# imported once per subprocess, but nothing guarantees a subprocess is never
+# reused across a different user/token, and a bare global would leak one
+# user's admin identity into another user's replies in that case.
```
rogercloud: **Minor: `_admin_id_cache` can never be hit — it adds code without providing caching.**

The token-keyed rework in response to the earlier review is directionally right and safe to keep, but the scenario it guards against can't occur: every MCP tool call spawns a fresh subprocess that exits when the call returns (`_execute_mcp_call` in `src/xagent/core/tools/adapters/vibe/mcp_adapter.py` calls `create_session` per invocation; `create_session`/`_create_stdio_session` in `src/xagent/core/tools/core/mcp/sessions.py` spawns and tears down the child process — no pooling exists). So the dict holds at most one entry and is always empty on lookup.

Either drop it and call `GET /me` unconditionally (simpler), or reword the comment to say plainly that this is defense-in-depth against a future pooled architecture rather than an active optimization.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+
+def _request(
+    method: str,
+    path: str,
+    *,
+    params: dict[str, Any] | None = None,
+    body: dict[str, Any] | None = None,
+) -> Any:
```
rogercloud, line 75: **Minor: no 429 handling.** A rate-limit response just falls through `raise_for_status()` and surfaces as a generic error. The Slack sibling at `src/xagent/web/tools/mcp/slack.py` already handles this with a bounded single retry honoring `Retry-After` (capped at `MAX_RETRY_AFTER_SECONDS = 30`). Under a burst — an agent iterating over many conversations — every 429 fails outright here. Worth mirroring the Slack behavior for consistency.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,288 @@
...
+
+mcp = FastMCP("intercom-mcp")
+
+# No region prefix: Intercom auto-routes each request to the workspace's
+# actual hosting region (US/EU/AU) based on the access token, so this one
+# host covers every region without per-region configuration.
+INTERCOM_BASE_URL = "https://api.intercom.io"
+# Pinned so a future default-version bump on Intercom's side can't silently
+# change response shapes underneath these tools.
+INTERCOM_API_VERSION = "2.11"
```
rogercloud, line 29: Nit / forward-looking: the pin itself is a good deliberate choice and the rationale comment is clear. The only gap is that there's no documented cadence for revisiting it as Intercom retires older API versions. A one-line note or a tracked follow-up on when/how to bump would be enough. Optional, non-blocking.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -152,6 +152,37 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
...
+        {
+            "provider_name": "intercom",
+            "name": "Intercom",
+            "client_id": os.environ.get("INTERCOM_CLIENT_ID", ""),
+            "client_secret": os.environ.get("INTERCOM_CLIENT_SECRET", ""),
+            # The single "app.intercom.com" authorize host serves every
+            # region: once a public app clears Intercom App Review, Intercom
+            # replicates it across US/EU/AU automatically, so there is no
+            # per-region app or authorize URL to manage.
+            "auth_url": "https://app.intercom.com/oauth",
```
rogercloud, line 184: **Minor: the region-host claim is stated as a guarantee three times with no citation.**

> `# The single "app.intercom.com" authorize host serves every region: ... Intercom replicates it across US/EU/AU automatically`

This is the PR's stated justification for not needing per-region configuration — and therefore for building a local connector rather than using Intercom's hosted MCP server. It's plausible (a single global authorize gateway is a common SaaS pattern) and I couldn't refute it, but it appears unhedged here, in the matching migration seed file, and in `example.env`, with no source and no fallback if it turns out to be wrong.

Please either link the specific Intercom docs page consulted, or soften the wording to reflect the confidence level.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -636,6 +667,27 @@ def get_builtin_public_mcp_app_rows() -> list[dict[str, Any]]:
...
+        {
+            "app_id": "intercom",
+            "name": "Intercom",
+            "description": "Connect to Intercom to search contacts, review conversations, and reply to customers.",
+            "icon": "https://www.google.com/s2/favicons?domain=intercom.com&sz=128",
+            "transport": "oauth",
+            "provider_name": "intercom",
+            "category": "Support",
+            "oauth_scopes": [],
+            "is_visible_in_connector": True,
```
rogercloud: **Minor: consider shipping hidden until manual verification is done.**

This entry sets `is_visible_in_connector: True`, while the PR's own test plan still has an unchecked box: "Manual verification of the individual Intercom MCP tools (search/reply/etc.) against a live workspace". Combined with the `state="all"` bug in `intercom_list_conversations`, that means a write-capable, customer-facing connector (reply / internal note / close) becomes discoverable in a state that hasn't been exercised end to end.

There's precedent in this same file: Chrome ships `is_visible_in_connector: False` as an explicit rollout gate with a comment explaining the pending limitation, and the migration carries a `RuntimeError` guard so ship-hidden builtins can't seed visible. Suggest doing the same here and flipping to visible in a follow-up once the manual verification box is checked.

rogercloud (changes requested):
This PR adds Intercom to the existing family of built-in OAuth MCP connectors, following the established three-part shape: an `oauth_providers` row, a `public_mcp_apps` catalog entry (duplicated canonically in `builtin_mcp_registry.py` and in a per-connector alembic seed migration), and a local `xagent.web.tools.mcp.intercom` module that wraps Intercom's REST API as seven customer-support tools (contact search/get, conversation list/get/reply, internal note, close). The one genuinely new piece of plumbing is a provider-specific normalization step in `generic_oauth_callback`, needed because Intercom's token endpoint returns a non-standard body. A follow-up commit rebased the migration onto the current alembic head and reworked the admin-id cache.

**Approach verdict: acceptable with reservations.** The structure is right and faithfully mirrors the sibling connectors — no architectural objections to the shape. Two design-level reservations: (1) the PR's central differentiating claim, that the single `app.intercom.com` authorize host covers US/EU/AU so no per-region configuration is needed, is asserted three separate times in code and docs with no citation and no fallback if it turns out to be wrong; since this is the stated reason for building a local connector rather than using Intercom's own hosted MCP server, it deserves a source link before merge. (2) The connector ships customer-facing *write* tools (reply, internal note, close) visible-by-default, with zero tests and an unchecked manual-verification box in the PR's own test plan.

**On the prior review:** the admin-id cache issue raised by gemini-code-assist was fixed exactly as requested — `_admin_id_cache` went from a bare `str | None` global to a token-keyed `dict[str, str]`. That resolution is correct and safe to keep. Worth noting for context, though: verification during this review found the original concern couldn't actually materialize. Every MCP tool call spawns a fresh subprocess (`_execute_mcp_call` in `src/xagent/core/tools/adapters/vibe/mcp_adapter.py` calls `create_session` per invocation; `create_session`/`_create_stdio_session` in `src/xagent/core/tools/core/mcp/sessions.py` spawns and tears down the child process per call — there is no session pooling anywhere), so the process can never hold two tokens. The flip side is that the cache also never yields a hit; see the minor finding below.

## Major

- **`intercom_list_conversations` omits the required `query` field when `state="all"`** — `src/xagent/web/tools/mcp/intercom.py:166-190`. `"all"` is one of the three values the docstring advertises, but it produces a `POST /conversations/search` body with no `query` key. Per Intercom's REST docs (v2.4, Search for Conversations), `query` is required and unfiltered listing belongs on `GET /conversations`; this call will very likely 400. New code, not a regression. See inline comment.

- **No tests at all** — whole PR. `git diff --stat` confirms 5 non-test files changed and nothing under `tests/`. Every sibling connector ships three things this PR skips: a migration test (`tests/alembic/test_20260730_seed_zoom_mcp_app.py`), a tool-module test (`tests/web/tools/test_zoom_mcp.py`), and OAuth-callback coverage for provider-specific token handling (`tests/web/test_meta_oauth.py`, which covers exactly the normalization case this PR introduces). The migration test matters more than usual here: `_upgrade_db_locked` in `src/xagent/db/migration.py` never executes migrations on a fresh DB — it stamps head and seeds directly from `builtin_mcp_registry.py` — and `validate_builtin_public_mcp_apps` only diffs two static in-code structures. A migration test is therefore the *only* mechanism in this codebase that can catch registry/migration drift for this connector, and it's the one thing missing. At minimum, please add the migration test and unit coverage for the token-normalization branch.

## Moderate

- **Unescaped model-supplied ids interpolated into URL paths (5 sites)** — `src/xagent/web/tools/mcp/intercom.py:158,200,229,252,274`. Verified empirically that `requests` normalizes `https://api.intercom.io/contacts/../admins` to `https://api.intercom.io/admins`, and that an id containing `?` injects query parameters. Because these ids can come from content an agent read (a ticket body, a search result), a tool scoped to "get one contact" or "close one conversation" can be silently redirected to a different endpoint the same bearer token can reach — a real confused-deputy path. The fix is a one-liner per site. See inline comments.

## Minor

- **Region-host claim asserted without citation** — `src/xagent/web/builtin_mcp_registry.py:160-164`, the matching migration seed file, and `example.env`. Plausible (one global authorize gateway is a common SaaS pattern) and I could not refute it, but it's stated as a guarantee in three places with no source. Cite the docs page or hedge the wording.

- **Token-exchange error path produces a confusing 500 rather than a clean failure** — `src/xagent/web/api/auth.py:1431-1445`. No `raise_for_status()`, and `if "error" in token_data:` doesn't match Intercom's `{"type": "error.list", "errors": [...]}` envelope, so `access_token` ends up `None`. It does *not* silently persist a broken connection — `UserOAuth.access_token` is NOT NULL, so `db.commit()` raises `IntegrityError`, caught by the outer handler and rendered as a generic "Authentication Failed" page (via `html.escape(str(e))`, which can leak a raw SQLAlchemy string). UX/error-message issue, not data integrity. See inline comment.

- **`_admin_id_cache` provides no actual caching** — `src/xagent/web/tools/mcp/intercom.py:30`. Per the process-lifecycle finding above, the dict can never hold more than one entry and can never be hit. Not a bug, just dead weight with a docstring that overstates what it does. Either drop it or make the comment honest about it being defense-in-depth for a hypothetical future architecture.

- **No 429 handling in `_request`** — `src/xagent/web/tools/mcp/intercom.py:54-80`. The Slack sibling (`src/xagent/web/tools/mcp/slack.py`) already implements a bounded one-retry backoff honoring `Retry-After` (capped at `MAX_RETRY_AFTER_SECONDS = 30`). Under a burst (an agent iterating over conversations) every 429 here fails outright.

- **Ships visible-by-default despite an unchecked manual-verification box** — `src/xagent/web/builtin_mcp_registry.py:679` and the matching migration row. There's precedent in this same file: Chrome ships `is_visible_in_connector: False` as an explicit rollout gate, with a comment explaining why and a `RuntimeError` guard in the migration to keep ship-hidden apps from seeding visible. The PR body still has "Manual verification of the individual Intercom MCP tools (search/reply/etc.) against a live workspace" unchecked. Combined with the `state="all"` bug, a write-capable customer-facing connector would be discoverable in a state its author hasn't exercised end to end. Suggest following the Chrome precedent and shipping hidden until that box is checked.

- **API version pin has no review cadence** — `src/xagent/web/tools/mcp/intercom.py:24`. The pin itself is a good deliberate call and the comment explains it well; just worth a note or tracked follow-up on when to revisit it as Intercom retires versions. Optional.

Lean already.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -669,6 +710,35 @@ def get_builtin_public_mcp_app_rows() -> list[dict[str, Any]]:
...
+            "category": "Support",
+            "oauth_scopes": [],
+            # Hidden until manually verified against a live workspace (all
+            # three regions, ideally), same precedent as the chrome row
+            # above: it ships customer-facing *write* tools (reply/note/
+            # close) discoverable by every user the moment it's visible, and
+            # that hasn't happened yet. Flip to True via a follow-up once
+            # that verification is done -- no redeploy needed, per the chrome
+            # row's note (is_visible_in_connector is not builtin-protected).
+            "is_visible_in_connector": False,
```
rogercloud, line 755: [major] Hidden-as-release-gate on an oauth-transport row, but the builtin_oauth connect path does not enforce the gate. `_reject_hidden_catalog_app` (mcp.py) covers only api_key/keyless and remote-MCP OAuth, and its docstring warns not to apply this pattern to an oauth-transport row until #1203 closes (issue open; item 2 names this exact breaking case). generic_oauth_callback -> _ensure_user_mcp_server (auth.py) has no visibility check, and the tool factory loads from UserMCPServer filtered only by user_id + is_active. So this only hides the catalog entry: any authenticated user who constructs the connect URL can OAuth-connect and reach the unverified write tools (reply/note/close). The chrome precedent is keyless and IS gate-enforced — the analogy does not hold. Suggest: do not ship the row until live verification is done, or enforce the gate in the oauth connect/callback flow.

src/xagent/web/builtin_mcp_registry.py
```diff
@@ -152,6 +152,47 @@ def get_builtin_oauth_provider_rows() -> list[dict[str, Any]]:
...
+            "provider_name": "intercom",
+            "name": "Intercom",
+            "client_id": os.environ.get("INTERCOM_CLIENT_ID", ""),
+            "client_secret": os.environ.get("INTERCOM_CLIENT_SECRET", ""),
+            # The single "app.intercom.com" authorize host is claimed to serve
+            # every region: per Intercom's community answer to this exact
+            # question (community.intercom.com/api-webhooks-23/do-we-need-to
+            # -create-multiple-oauth-apps-per-data-region-2522), a public app
+            # is replicated across US/EU/AU automatically once it clears
+            # Intercom App Review, so there is no per-region app or authorize
```
rogercloud, line 178: [minor] The region claim (one authorize host covers US/EU/AU) rests on a community answer, not primary docs — disclosed and hedged here and in the other three locations, and the fallback analysis (per-region rows would force per-region app rows, given the single-provider data model) is accurate. Non-blocking: keep the flip-to-visible gated on live verification against a non-US workspace.

src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
```diff
@@ -0,0 +1,187 @@
...
+    if "public_mcp_apps" in existing_tables:
+        remaining_intercom_apps = bind.execute(
+            sa.select(sa.func.count())
+            .select_from(PUBLIC_MCP_APPS_TABLE)
+            .where(PUBLIC_MCP_APPS_TABLE.c.provider_name == "intercom")
+        ).scalar()
+        if remaining_intercom_apps:
+            return
+
+    # Only delete the provider row when it still matches the static shape this
```
rogercloud, line 187: [minor] Third divergent copy of the downgrade shape-guard semantics. The slack seed migration (20260801) argues the shape guard deletes admin-recreated canonical rows anyway, while this guard additionally excludes client_id/client_secret — so a credential-only admin edit is silently destroyed on downgrade. Downgrade-path-only, so minor; suggest aligning with the slack decision (delete unconditionally) or documenting why this copy diverges.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+@mcp.tool()
+def intercom_list_conversations(state: str = "open", limit: int = 20) -> str:
+    """
+    List Intercom conversations. `state` is one of "open", "closed", or "all".
+    Returns at most `limit` conversations (max 100), most recently updated first.
+    """
+    try:
+        valid_states = {"open", "closed", "all"}
```
rogercloud: [minor] Intercom 2.11's state enum is open/closed/snoozed; `snoozed` is silently excluded here, `state='open'` omits snoozed conversations, and the docstring never mentions it. Only `'all'` surfaces them. Suggest adding `snoozed` (and to the docstring) or documenting the exclusion.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+        "title": conversation.get("title"),
+        "subject": source.get("subject"),
+        "preview": source.get("body"),
+        "created_at": conversation.get("created_at"),
+        "updated_at": conversation.get("updated_at"),
+    }
+
+
+@mcp.tool()
+def intercom_search_contacts(query: str, limit: int = 10) -> str:
```
rogercloud, line 156: [minor] Empty/whitespace `query` is passed into the `~` contains operator with no validation — the FastMCP schema has no minLength and no caller-layer check, so behavior for an empty contains is undefined. Suggest rejecting blank queries with a clear error before building the body.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+        return _error(str(e))
+
+
+@mcp.tool()
+def intercom_reply_to_conversation(conversation_id: str, body: str) -> str:
+    """
+    Reply to a customer on an Intercom conversation, as the connected admin.
+    """
+    try:
+        reply = _request(
```
rogercloud, line 296: [minor] `intercom_reply_to_conversation` (and `intercom_add_internal_note`, :282-295) accept an empty body, but Intercom requires a body for comment/note message types — the user gets a confusing upstream error. Suggest validating `body.strip()` non-empty in both tools.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+                retry_after = 0
+            if 0 < retry_after <= MAX_RETRY_AFTER_SECONDS:
+                time.sleep(retry_after)
+                continue
+        break
+
+    try:
+        response.raise_for_status()
+    except requests.HTTPError as exc:
+        response_text = response.text.strip()
```
rogercloud, line 97: [minor] The unbounded stripped `response.text` is appended to the error message. The zoom sibling caps this at MAX_ERROR_RESPONSE_TEXT_CHARS=1000 (zoom.py:25-26, 127-133) because HTML gateway error pages must not be forwarded unbounded to the LLM/logs. Suggest the same cap here.

src/xagent/web/api/auth.py
```diff
@@ -1446,8 +1467,27 @@ def generic_oauth_callback(
...
 
+        if not access_token:
```
rogercloud, line 1550: [minor] The guard returns a generic "did not return an access token" and drops Intercom's `error.list` `errors[0].message`, which is in scope in `token_data` right here; the sibling `if "error" in token_data` guard (:1459-1465) already surfaces the full envelope. This is the unaddressed half of the prior review's P5 suggestion. Suggest including the provider's message when present, e.g. `errors[0].message` for the error.list envelope.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+        return _error(str(e))
+
+
+@mcp.tool()
+def intercom_get_contact(contact_id: str) -> str:
+    """
+    Get an Intercom contact by id.
+    """
+    try:
+        contact = _request("GET", f"/contacts/{quote(contact_id, safe='')}")
```
rogercloud, line 190: [minor] `intercom_get_contact` returns the raw unbounded provider object, while search (:166) and get_conversation (:243-246) curate via `_contact_summary`/`_conversation_summary`. hubspot returns raw too but bounds via a property allowlist at request time — this module is internally shape-inconsistent. Suggest wrapping in `_contact_summary`.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+                "operator": ">",
+                "value": 0,
+            }
+        else:
+            query = {"field": "state", "operator": "=", "value": state}
+
+        body: dict[str, Any] = {
+            "query": query,
+            "pagination": {"per_page": max(1, min(limit, 100))},
+            "sort": {"field": "updated_at", "order": "descending"},
```
rogercloud, line 237: [minor] The `sort` object is not documented for POST /conversations/search (the live API reference documents `sort` only for /contacts/search), and the docstring (:192) promises "most recently updated first" — an unverified promise; worst case the sort is ignored and order is unspecified. Suggest verifying against the live API or dropping the sort (and the docstring promise).

src/xagent/web/api/auth.py
```diff
@@ -1446,8 +1467,27 @@ def generic_oauth_callback(
...
+            # persistence block below with access_token=None, hit
+            # UserOAuth.access_token's NOT NULL constraint, and surface as a
+            # raw SQLAlchemy IntegrityError message through the generic
+            # exception handler instead of a clear, actionable error.
+            import html
+
+            return HTMLResponse(
+                content=(
+                    "<h1>Error exchanging token</h1>"
+                    f"<p>{html.escape(provider)} did not return an access token.</p>"
```
rogercloud: [minor] This guard is shared, provider-agnostic code, so it changes behavior for every provider (previously a tokenless response hit UserOAuth's NOT NULL constraint and leaked a raw IntegrityError 500; now a clean 400). An improvement, but no test drives a non-intercom provider through this branch — only test_intercom_oauth.py uses provider='intercom', and meta/slack/zoom tests always supply access_token. Suggest one parameterized cross-provider test.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,326 @@
...
+        }
+
+        result = _request("POST", "/conversations/search", body=body)
+        conversations = [
+            _conversation_summary(c) for c in result.get("conversations", [])
+        ]
+        return _success(conversations=conversations)
```
rogercloud: [minor] The response surfaces no total_count/pages/cursor, while search_contacts returns `total` (:167-169). The API returns total_count and pages.next.starting_after for free, and sibling connectors surface has_more/truncated (hubspot, slack). A caller cannot distinguish "no more results" from "results withheld". Suggest returning total_count (and the next-page cursor).

tests/web/tools/test_intercom_mcp.py
```diff
@@ -0,0 +1,316 @@
...
+    monkeypatch.setattr(intercom.requests, "request", mock_request)
+    # First call resolves the admin id via /me, second is the reply itself.
+    mock_request.side_effect = [
+        MockResponse(json_data={"id": "admin-1"}),
+        MockResponse(json_data={"id": "conv1"}),
+    ]
+
+    intercom.intercom_reply_to_conversation("../admins", "hello")
+
+    reply_call = mock_request.call_args_list[1]
```
rogercloud, line 132: [minor] The write-tool tests (reply :81-96, note :99-114, close :117-132) assert only the URL: swapping `message_type` comment<->note would change no URL and all tests would still pass. Suggest asserting the request body (`json` kwarg) — message_type/type/body — in each.

tests/web/tools/test_intercom_mcp.py
```diff
@@ -0,0 +1,316 @@
...
+    monkeypatch.setattr(intercom.requests, "request", mock_request)
+
+    intercom.intercom_get_contact("123?display_as=plaintext")
+
+    called_url = mock_request.call_args.kwargs["url"]
+    assert called_url == "https://api.intercom.io/contacts/123%3Fdisplay_as%3Dplaintext"
+    assert mock_request.call_args.kwargs["params"] is None
+
+
+def test_get_conversation_percent_encodes_id(monkeypatch):
```
rogercloud, line 73: [minor] The only get_conversation test mocks an empty conversation (`{"id": "conv1"}`), so the parts projection (intercom.py:244-252) never executes in tests. Suggest a test with conversation_parts populated to exercise the projection.

tests/web/tools/test_intercom_mcp.py
```diff
@@ -0,0 +1,316 @@
...
+    monkeypatch.setattr(intercom.time, "sleep", Mock())
+
+    # Retry-After of "0" does not satisfy `0 < retry_after`, so this call
+    # should NOT sleep/retry and must surface the 429 as an error instead.
+    result = json.loads(intercom.intercom_get_contact("c1"))
+    assert result["status"] == "error"
+    assert mock_request.call_count == 1
+
+
+def test_request_retries_once_on_429_with_positive_retry_after(monkeypatch):
```
rogercloud, line 347: [minor] Untested branches: the limit clamp max(1, min(limit, 100)) (intercom.py:163, 215 — no test passes an explicit limit), `_request`'s 204/empty-content -> {} (intercom.py:101-102), and `_current_admin_id`'s RuntimeError when /me lacks an id (intercom.py:118-120). Suggest three targeted tests.

tests/web/test_intercom_oauth.py
```diff
@@ -0,0 +1,213 @@
...
+        expires_delta=timedelta(minutes=10),
+    )
+    request = SimpleNamespace(query_params={"code": "bad-code", "state": state})
+
+    monkeypatch.setattr(
+        auth_api.requests,
+        "post",
+        Mock(
+            return_value=MockResponse(
+                {"type": "error.list", "errors": [{"message": "invalid code"}]}
```
rogercloud, line 232: [minor] This error-envelope test uses MockResponse with the default status 200 and a `json()` that can never raise, so a non-200 token POST and a non-JSON body are untested (auth.py:1454-1457 has no status check; non-JSON falls through to the pre-existing generic 500 handler). Suggest asserting the response status and adding a non-200/non-JSON case.

rogercloud (changes requested):
## Summary

This PR adds Intercom as a built-in MCP connector: an `oauth_providers` row, a `public_mcp_apps` catalog row shipping hidden, a local FastMCP module (`src/xagent/web/tools/mcp/intercom.py`) with seven tools against the Intercom REST API, and token-response normalization in the generic OAuth callback. The design is region-less — `app.intercom.com` authorize and `api.intercom.io` API hosts auto-route per workspace, so one configuration covers US/EU/AU, unlike Intercom's hosted MCP server which has no AU endpoint; the connector ships hidden pending live-workspace verification. There is no schema change: the two catalog rows are frozen into a seed migration and runtime state reuses `UserOAuth` (`access_token` NOT NULL; `expires_at`/`refresh_token` stay NULL, consistent with the existing model).

## Approach

Round 0 verdict: **acceptable-with-reservations**. The copy-per-connector FastMCP module structure matches the established hubspot/slack/zoom pattern (`_success`/`_error`/`_headers`/`_request` shape), the data model needs no changes, the empty-scopes authorize flow matches Intercom's actual behavior, and the hidden-ship strategy for an unverified connector is reasonable in intent. Two design-level reservations, both minor:

- **D1 (minor)** — Region premise relies on a community answer rather than primary API docs. The claim is disclosed and hedged in all four locations (registry provider row, registry app row, example.env, migration), and the fallback analysis is accurate: the data model has no region dimension (single `provider_name`, single `UserOAuth.provider`), so per-region rows would force per-region app rows. Non-blocking because there are zero users today and the local module is independently justified by the hosted server's missing AU support — but keep the flip-to-visible strictly gated on live-workspace verification against a non-US workspace (the exact verification that would resolve the premise either way).
- **D2 (minor)** — The downgrade shape-guard in the seed migration is a third divergent copy of disputed semantics (see inline). Downgrade-path-only, narrow trigger.

## Update since last review

Two commits since the previous review (ccbb57c0): 87b45540 "fix: address PR review — path injection, state=all bug, error handling, tests" adds `quote(..., safe='')` at all five injection sites, an always-present query predicate for `state='all'`, the access-token 400 guard in the OAuth callback, three test files (~742 lines), hides the catalog row, and adds region citations and API-pin cadence notes; 7131bf96 is a pure rebase onto latest upstream to resolve a second alembic head conflict. All prior findings are addressed by these commits.

## Prior findings

| # | Severity | Status | Evidence |
|---|----------|--------|----------|
| P1 | major | FIXED | `state='all'` now always sends a query (`created_at > 0` predicate, valid per search-conversations docs); regression test covers open/closed/all |
| P2 | major | FIXED | 3 test files, 31 real test functions: migration/alembic cycle, oauth normalization + callback e2e, tool module (path injection, 429, cache isolation) |
| P3 | moderate | FIXED | `quote(..., safe='')` at all 5 sites (intercom.py:181, 236-237, 267, 290, 312); no new unencoded sites |
| P4 | minor | FIXED | Region claim cited (community answer URL) and hedged in all 3 places; API-host auto-routing separately marked as primary-documented |
| P5 | minor | FIXED | Explicit guard returns 400 "did not return an access token" before persistence; test asserts 400, no IntegrityError leak, zero UserOAuth rows (residual diagnostics/coverage gaps tracked as F8/F11/F17 below) |
| P6 | minor | FIXED | admin_id cache comment reworded to honest defense-in-depth (intercom.py:36-42), exactly the reviewer's option B |
| P7 | minor | FIXED | Bounded one-retry honoring Retry-After capped at MAX_RETRY_AFTER_SECONDS=30, identical to slack.py:63-79; tests cover 0/positive/over-cap |
| P8 | minor | FIXED | `is_visible_in_connector=False` in both registry (:729) and migration (:101), drift-pinned by test_seed_rows_match_registry; comment explains the gate (non-blocking note: the intercom migration lacks the chrome-style RuntimeError guard for the missing-column edge — chrome's own comment calls that edge unreachable via normal upgrade, and the registry seed path carries the equivalent guard) |
| P9 | minor | FIXED | API-pin comment now names when/where/how to revisit 2.11 |

## Findings

### Major

**F1 (blocking)** — `src/xagent/web/builtin_mcp_registry.py:729` — The intercom app row ships `is_visible_in_connector=False` on an oauth-transport row, but the builtin_oauth connect path does not enforce the hidden gate. `_reject_hidden_catalog_app` (mcp.py:2320-2352) covers only api_key/keyless and remote-MCP OAuth, and its docstring (2332-2341) explicitly warns not to apply the hidden-as-release-gate pattern to an oauth-transport row until #1203 closes (issue open; item 2 names this exact breaking case). `generic_oauth_callback -> _ensure_user_mcp_server` (auth.py:1284-1376) has no visibility check, and the tool factory (factory.py:1141-1145) loads from `UserMCPServer` filtered only by user_id + is_active. The gate therefore only hides the catalog entry: any authenticated user who constructs the connect URL can OAuth-connect and reach the unverified customer-facing write tools (reply/note/close). The chrome precedent is keyless and IS gate-enforced, so the analogy does not hold. Suggestion: do not ship the row (or module) until live verification is done, or enforce the gate in the oauth connect/callback path; at minimum, track enforcement explicitly against #1203 before flipping visible.

### Minor

- **D1** — `src/xagent/web/builtin_mcp_registry.py:160-170` — Region premise disclosed, cited, and hedged in all four locations; fallback analysis accurate; flip-to-visible must stay gated on live-workspace verification. Non-blocking.
- **D2** — `src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py:171-187` — Third divergent copy of the downgrade shape-guard semantics (see inline).
- **F3** — `src/xagent/web/tools/mcp/intercom.py:195` — `valid_states = {"open", "closed", "all"}` silently excludes `snoozed` (Intercom 2.11 state enum is open/closed/snoozed); `state='open'` omits snoozed conversations and the docstring never mentions it. Suggest adding `snoozed` or documenting the exclusion.
- **F4** — `src/xagent/web/tools/mcp/intercom.py:150` — `intercom_search_contacts` accepts an empty/whitespace query into the `~` contains operator with no validation (no minLength in the FastMCP schema, no caller-layer check) — undefined behavior. Suggest rejecting blank queries with a clear error.
- **F5** — `src/xagent/web/tools/mcp/intercom.py:265` — reply/note accept an empty body, but Intercom requires a body for comment/note message types — the user gets a confusing upstream error. Suggest validating `body.strip()` non-empty in both tools.
- **F6** — `src/xagent/web/tools/mcp/intercom.py:95-99` — The unbounded stripped `response.text` is appended to the error message; the zoom sibling caps at MAX_ERROR_RESPONSE_TEXT_CHARS=1000 because HTML gateway error pages must not be forwarded unbounded to the LLM/logs. Suggest the same cap.
- **F8** — `src/xagent/web/api/auth.py:1473` — The new guard returns a generic "did not return an access token" and drops Intercom's `error.list` `errors[0].message`, which is in scope in `token_data` at the guard; the sibling `if "error" in token_data` guard (1459-1465) already surfaces the envelope. This is the unaddressed half of the prior review's P5 suggestion. Suggest including the provider's message when present.
- **F9** — `src/xagent/web/tools/mcp/intercom.py:181-182` — `intercom_get_contact` returns the raw unbounded provider object while search (166) and get_conversation curate via `_contact_summary`/`_conversation_summary`; hubspot bounds at request time via a property allowlist, so this module is internally shape-inconsistent. Suggest wrapping in `_contact_summary`.
- **F10** — `src/xagent/web/tools/mcp/intercom.py:216` — The `sort` object is not documented for POST /conversations/search (the live API reference documents `sort` only for /contacts/search), and the docstring (192) promises "most recently updated first" — an unverified promise; worst case the sort is ignored and order is unspecified. Suggest verifying against the live API or dropping the sort.
- **F11** — `src/xagent/web/api/auth.py:1473-1489` — The blanket guard is provider-agnostic shared code, so it changes behavior for every provider (previously a tokenless response hit the NOT NULL constraint and leaked a raw IntegrityError 500; now a clean 400). An improvement, but no test drives any non-intercom provider through the new branch — only test_intercom_oauth.py uses provider='intercom', and the meta/slack/zoom tests always supply access_token. Suggest one parameterized cross-provider test.
- **F12** — `src/xagent/web/tools/mcp/intercom.py:223` — The list response surfaces no total_count/pages/cursor while search_contacts returns `total` (167-169); the API returns total_count and pages.next.starting_after for free, and sibling connectors surface has_more/truncated (hubspot, slack). A caller cannot distinguish "no more results" from "results withheld". Suggest returning total_count (and the next-page cursor).
- **F13** — `tests/web/tools/test_intercom_mcp.py:81-132` — Write-tool tests assert only the URL; swapping `message_type` comment<->note would change no URL and all tests would still pass. Suggest asserting the request body (message_type/type/body) in the reply/note/close tests.
- **F14** — `tests/web/tools/test_intercom_mcp.py:71-78` — The only get_conversation test mocks an empty conversation, so the parts projection (intercom.py:244-252) never executes in tests. Suggest a test with conversation_parts populated.
- **F15** — `tests/web/tools/test_intercom_mcp.py:190` — Untested branches: the limit clamp max(1, min(limit, 100)) (intercom.py:163, 215 — no test passes an explicit limit), `_request`'s 204/empty-content -> {} (101-102), and `_current_admin_id`'s RuntimeError when /me lacks an id (118-120). Suggest three targeted tests.
- **F17** — `tests/web/test_intercom_oauth.py:171-211` — The error-envelope test uses MockResponse with the default status 200 and a `json()` that can never raise; a non-200 token POST and a non-JSON body are untested (auth.py:1454-1457 has no status check; non-JSON falls through to the pre-existing generic 500 handler). Suggest asserting the status and adding a non-JSON case.

`src/xagent/web/tools/mcp/intercom.py:70: delete: _request's params kwarg is dead — zero callers pass it (all 8 call sites pass body only). Remove the parameter and the pass-through; requests' params defaults to None so behavior is unchanged.`

Blocking: yes. Blocking issue:
- `src/xagent/web/builtin_mcp_registry.py:729` — major — hidden-as-release-gate on an oauth-transport row whose connect path does not enforce the gate (mcp.py warns against exactly this; #1203 open) — [new]

All other findings (D1, D2, F3-F15, F17) are minor and non-blocking. Recommended event: **REQUEST_CHANGES**.

src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py
```diff
@@ -0,0 +1,187 @@
...
+            )
+
+    if "public_mcp_apps" in existing_tables:
+        app_columns = {
+            column["name"] for column in inspector.get_columns("public_mcp_apps")
+        }
+        existing_app_ids = set(
+            bind.execute(sa.select(PUBLIC_MCP_APPS_TABLE.c.app_id)).scalars()
+        )
+        if APP_ID not in existing_app_ids:
```
rogercloud, line 137: `upgrade()` only inserts when `app_id` is absent; a pre-existing operator-created visible `intercom` row is left untouched, and the builtin registry overlays transport/launch_config onto ANY row sharing this app_id at read time (`_app_to_dict`), yielding a visible, working, customer-facing-write connector that defeats the hidden-rollout gate. The chrome sibling migration (20260806_seed_chrome_mcp_app.py, driven by #1143) force-hides on exactly this collision — suggest mirroring that guard.

src/xagent/web/api/auth.py
```diff
@@ -1393,6 +1432,38 @@ def generic_oauth_callback(
...
+        # provider -- matching the "indistinguishable from nonexistent" intent
+        # the helper already documents for its other two call sites. The bare
+        # (app_id-less) batch branch below cannot move this early: it doesn't
+        # know which apps share the provider until it queries them, so its
+        # own per-app check stays where it is.
+        from ..mcp_apps import get_app_by_id
+        from .mcp import _reject_hidden_catalog_app
+
+        target_app_info = get_app_by_id(db, app_id)
+        if target_app_info:
```
rogercloud, line 1482: `if target_app_info:` means an unknown app_id skips the gate entirely: the token exchange proceeds, a UserOAuth row is persisted keyed by the bogus app_id, and the success page renders — while a hidden app 404s, contradicting the "indistinguishable from a nonexistent one" contract in `_reject_hidden_catalog_app`'s docstring (the other two call sites 404 on unknown apps). The unknown-app_id behavior is pre-existing; the asymmetry is new. Suggest returning the same 404 when app_id is truthy but `get_app_by_id` returns None.

src/xagent/web/api/auth.py
```diff
@@ -1393,6 +1432,38 @@ def generic_oauth_callback(
...
+            try:
+                _reject_hidden_catalog_app(target_app_info)
+            except HTTPException:
+                return HTMLResponse(
```
rogercloud, line 1486: This 404 page is bare HTML with no `window.opener.postMessage`, and `generic_oauth_login` (auth.py:1236) still redirects a hidden app_id to the provider's consent page — so "a hidden app never even reaches the provider" holds only for the token-exchange hop, and the frontend spinner (connect-mcp-dialog.tsx polls only for `oauth-success`, capped at 5 min) can hang for minutes. No security impact (the gate still fires before exchange/persistence); consider gating in `generic_oauth_login` too and/or emitting an `oauth-error` postMessage from this page.

src/xagent/web/api/auth.py
```diff
@@ -1535,6 +1630,21 @@ def generic_oauth_callback(
                     if app.get("provider") == provider
                 ]
                 for app_info in apps:
```
rogercloud, line 1660: When every app under the provider is hidden (today's intercom reality for a hand-crafted bare URL), the loop skips them all but the provider-level UserOAuth grant is still committed and the callback renders "Connected Successfully" — a dormant grant row plus a misleading success page. This mirrors the AppNotOAuthError skip and is test-documented as deliberate; still, consider failing (or rendering a distinct page) when zero apps survived the batch.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,351 @@
...
+            "sort": {"field": "updated_at", "order": "descending"},
+        }
+
+        result = _request("POST", "/conversations/search", body=body)
+        conversations = [
+            _conversation_summary(c) for c in result.get("conversations", [])
+        ]
+        # Intercom's search pagination carries a "next" cursor object under
+        # "pages" only when more results remain past this page.
+        has_more = bool((result.get("pages") or {}).get("next"))
```
rogercloud: `has_more` is derived from `pages.next` but the cursor itself (`starting_after`) is discarded and the tool has no cursor input, so the flag is not actionable — with `limit` capped at 100 there is no way to page past the first page. Siblings make pagination actionable (zoom `next_page_token`, hubspot `after`). This is the unaddressed half of prior F12.

src/xagent/web/tools/mcp/intercom.py
```diff
@@ -0,0 +1,351 @@
...
+        return _success(
+            conversations=conversations,
+            total_count=result.get("total_count", len(conversations)),
```
rogercloud, line 249: Search returns `total` (line 176) while list returns `total_count` (here) for the same underlying Intercom field. Nit, but pick one vocabulary — the hubspot siblings use `total`.

src/xagent/web/api/auth.py
```diff
@@ -1507,10 +1598,14 @@ def generic_oauth_callback(
                 )
 
             from ..mcp_apps import get_all_mcp_apps, get_app_by_id
+            from .mcp import _reject_hidden_catalog_app
 
             if app_id:
                 app_info = get_app_by_id(db, app_id)
```
rogercloud: `get_app_by_id` is called twice (line 1454 and here); nothing mutates between, and `target_app_info` is already in scope with the identical dict shape — reuse it here instead of re-querying. TOCTOU impact is negligible given the helper's documented blast-radius semantics.

src/xagent/web/api/auth.py
```diff
@@ -1393,6 +1432,38 @@ def generic_oauth_callback(
...
+        # _ensure_user_mcp_server, so a hidden app never even reaches the
+        # provider -- matching the "indistinguishable from nonexistent" intent
+        # the helper already documents for its other two call sites. The bare
+        # (app_id-less) batch branch below cannot move this early: it doesn't
+        # know which apps share the provider until it queries them, so its
+        # own per-app check stays where it is.
+        from ..mcp_apps import get_app_by_id
+        from .mcp import _reject_hidden_catalog_app
+
+        target_app_info = get_app_by_id(db, app_id)
```
rogercloud, line 1481: Pre-existing (base 4595a62c behaves identically; non-blocking): a user-supplied app_id that passes this visibility-only gate is trusted as-is, so app_id='gmail' with an intercom provider persists UserOAuth(provider='gmail') holding the intercom token and creates a gmail MCPServer association with it — corrupting the user's own-account state. This gate is the natural place to also validate `target_app_info.provider == provider`.

tests/web/test_intercom_oauth.py
```diff
@@ -0,0 +1,463 @@
...
+            )
+        ),
+    )
+    get_mock = Mock()
+    monkeypatch.setattr(auth_api.requests, "get", get_mock)
+
+    response = generic_oauth_callback("intercom", request, db, _intercom_provider())
+
+    assert response.status_code == 400
+    assert "did not return an access token" in response.body.decode()
```
rogercloud: This error.list test pins only the generic message; `_extract_provider_error_message`'s detail path is exercised but never asserted. Add `assert 'invalid code' in response.body.decode()` so the envelope's `errors[0].message` surfacing is locked in.

tests/alembic/test_20260812_seed_intercom_mcp_app.py
```diff
@@ -0,0 +1,213 @@
...
+        assert row[0] == "oauth"
+        assert row[1] == "intercom"
+        # Ships hidden pending manual verification against a live workspace;
+        # see builtin_mcp_registry.py's intercom app row for the rationale.
+        assert row[2] == 0
+        assert "xagent.web.tools.mcp.intercom" in str(row[3])
+        assert "INTERCOM_ACCESS_TOKEN" in str(row[3])
+
+
+def test_upgrade_is_idempotent(tmp_path):
```
rogercloud, line 107: `test_upgrade_is_idempotent` only re-runs upgrade against a hidden seeded row; add a pre-inserted VISIBLE `app_id='intercom'` row and assert the upgrade force-hides it (companion to the N4(b) finding on the missing collision branch).

rogercloud (approved):
### Summary

This follow-up resolves the round-2 blocking finding by enforcing the hidden-app release gate on both builtin_oauth connect branches of `generic_oauth_callback`: a fail-fast 404 before any token exchange on the single-app path (auth.py:1454-1466) and a skip-and-continue per-app check on the app_id-less batch path (auth.py:1632-1647), with the gate's docstring updated to reflect all three connect paths. It also lands the remaining intercom module polish (snoozed state, blank query/body guards, error-text cap, contact summaries, provider error detail, pagination metadata) plus targeted regression coverage, including hidden-app gate tests for both branches and a cross-provider guard test. CI is green.

### Approach

Sound. The delta is a root-cause fix: it extends the shared `_reject_hidden_catalog_app` helper to the last unenforced connect path with branch-appropriate semantics, keeps a single source of truth for the predicate, and adds precise regression coverage. All remaining concerns are minor.

### Update since last review (f9aa8332)

The fix commit moves the hidden-app gate ahead of the token exchange (not just ahead of persistence) for the single-app branch, adds the per-app hidden skip to the bare batch branch, and corrects the gate docstring that previously flagged the builtin_oauth path as unenforced. It also adds `_extract_provider_error_message` to surface Intercom's `error.list` detail (html-escaped) in the 400 response, refines the intercom module per round-2 findings, and adds the round-2 tests: hidden-gate single/batch, zoom cross-provider guard, error-detail, blank-input, limit-clamp, and non-JSON response coverage.

### Prior findings

| Finding | Status | Evidence |
|---|---|---|
| F1 — hidden gate not enforced on builtin_oauth connect (blocking) | FIXED | Early gate at auth.py:1454-1466 before token exchange; bare-batch per-app skip at auth.py:1632-1647; docstring updated (mcp.py:2327-2333); factory never creates; tests assert 404 + no token exchange + no rows |
| D1 — region-claim hedging | FIXED | Hedged and cited in all 4 locations; still hidden |
| D2 — downgrade shape-guard divergence | WAIVED | Byte-identical semantics to the zoom sibling migration; documented by author |
| F3 — snoozed state | FIXED | intercom.py:205 `valid_states` includes `snoozed`; docstring + test |
| F4 — blank query | FIXED | intercom.py:161-162 `query.strip()` + `test_search_contacts_rejects_blank_query` |
| F5 — blank body | FIXED | intercom.py:286-287, 311-312 `body.strip()` + both tests |
| F6 — error text cap | FIXED | `MAX_ERROR_RESPONSE_TEXT_CHARS=1000` at intercom.py:38, truncation at 97-101, test at test_intercom_mcp.py:505-513 |
| F8 — provider error message | FIXED | `_extract_provider_error_message` (auth.py:284-301), html-escaped guard at 1533-1537; residual: detail string unasserted → N11(a) |
| F9 — contact summary | FIXED | intercom.py:190 `_contact_summary` + regression test |
| F10 — sort citation | FIXED | Verified accurate against live Intercom docs (sort documented for /conversations/search) |
| F11 — cross-provider test | FIXED | Zoom drives the guard; asserts 400 + message + no row |
| F12 — total_count/cursor | PARTIAL | `total_count` + `has_more` added (intercom.py:239-243) but the `pages.next` cursor is still discarded → N2 |
| F13/F14/F15/F17 — test gaps | FIXED | Body assertions, populated-parts projection, limit clamp/204/RuntimeError, non-200 + non-JSON tests all present |

### Findings (all Minor)

**N4(b) — migration collision branch missing** — src/xagent/migrations/versions/20260812_seed_intercom_mcp_app.py:137-141. `upgrade()` only inserts when `app_id` is absent. An admin-created visible `intercom` row pre-existing before deploy is left untouched, and the builtin registry overlays transport/launch_config onto ANY row sharing the app_id at read time (mcp_apps.py `_app_to_dict`) while `is_visible_in_connector` comes from the DB row → a visible, working, customer-facing-write connector, defeating the release gate. The chrome sibling migration (20260806_seed_chrome_mcp_app.py, motivated by #1143) force-hides on exactly this collision; mirror that guard. (Also: no `RuntimeError` when the visibility column is missing — unreachable via a normal `upgrade head` per the verified ancestry, and already accepted as the prior P8 note.)

**N1 — hidden-vs-nonexistent asymmetry** — src/xagent/web/api/auth.py:1454-1465. The gate runs only `if target_app_info:`, so an unknown app_id skips it: token exchange proceeds, a UserOAuth row is persisted keyed by the bogus app_id, and the success page renders — while a hidden app 404s. That contradicts the "indistinguishable from a nonexistent one" contract in `_reject_hidden_catalog_app`'s docstring (mcp.py:2327-2328), and the other two call sites 404 on unknown apps. The unknown-app_id behavior is pre-existing; the asymmetry is new. Suggest returning the same 404 when app_id is truthy but `get_app_by_id` returns None.

**N7 — login hop ungated + 404 page without postMessage** — src/xagent/web/api/auth.py:1235-1290 (login), 1459-1466 (404 page). `generic_oauth_login` still redirects a hidden app_id to the provider consent page, so the commit message's "a hidden app never even reaches the provider" holds only for the token-exchange hop (the tested one). The 404 page is bare HTML with no `window.opener.postMessage`, and the frontend (connect-mcp-dialog.tsx) polls only for `oauth-success`, clearing on `popup.closed` or a 5-minute cap → the spinner can hang for minutes. No security bypass (the gate fires before exchange/persistence). Suggest gating in `generic_oauth_login` too (app_info is already fetched at auth.py:1236) and/or emitting an `oauth-error` postMessage from the 404 page.

**N3 — bare-batch success-with-nothing** — src/xagent/web/api/auth.py:1579-1584, 1632-1647. When every app under the provider is hidden (today's intercom reality for a hand-crafted bare URL), the loop skips them all, yet the provider-level UserOAuth grant is committed and the callback renders "Connected Successfully" — a dormant grant row plus a misleading success page. Deliberate (mirrors the AppNotOAuthError skip, explicitly test-documented) and low-reachability (the frontend always passes app_id; hidden apps are stripped from the catalog), but consider failing or rendering a distinct page when zero apps survived.

**N2 — pagination cursor still discarded** — src/xagent/web/tools/mcp/intercom.py:239-244. `total_count` + `has_more` landed, but `pages.next.starting_after` is dropped and the tool has no cursor input, so `has_more` is not actionable (limit capped at 100). Siblings make pagination actionable (zoom `next_page_token`, hubspot `after`). This is the unaddressed half of prior F12.

**N6 — total vs total_count vocabulary** — src/xagent/web/tools/mcp/intercom.py:176 vs :242. Search returns `total`, list returns `total_count`, for the same underlying provider field. Nit; pick one (the hubspot siblings use `total`).

**N8 — double fetch / TOCTOU nit** — src/xagent/web/api/auth.py:1454 vs 1604. `get_app_by_id` is called twice; nothing mutates between and `target_app_info` is already in scope with the identical dict shape — reuse it in the later branch. TOCTOU impact is negligible per the helper's documented blast-radius semantics.

**N9 — cross-app token misattribution (pre-existing)** — src/xagent/web/api/auth.py. A user-supplied app_id (e.g. gmail) passes the visibility-only gate; UserOAuth(provider='gmail') is then persisted with the intercom token and a gmail MCPServer association is created with it (corrupted own-account state; gmail watch best-effort logs the failure). Base 4595a62c behaves identically — flagged pre-existing, non-blocking. The new gate is the natural place to also validate `target_app_info.provider == provider`.

**N11(a) — error-detail extraction unasserted** — tests/web/test_intercom_oauth.py:236-239. The error.list test asserts only the generic message; assert `'invalid code'` (the envelope's `errors[0].message`) so `_extract_provider_error_message` is pinned.

**N11(b) — no collision test for the migration** — tests/alembic/test_20260812_seed_intercom_mcp_app.py. `test_upgrade_is_idempotent` only re-runs upgrade against a hidden row; add a pre-inserted VISIBLE `app_id='intercom'` row test to pin the intended hidden-shipping semantics (companion to N4(b)).

### Informational note

N10: the oauth-success postMessage script (auth.py:1705-1709) embeds values via `json.dumps` without escaping `<` — a technically real same-origin DOM edge, but byte-identical at base 4595a62c and outside every diff hunk, so out of scope here. Recommend tracking it as a standalone issue.

Lean already.

### Blocking status

Blocking: no — recommended event: APPROVE. No confirmed critical or major findings; the sole round-2 blocking finding (F1) is fixed, and the hidden-ship gate makes the unverified write tools unreachable for normal users. The residual concerns are release-hygiene edges (migration collision branch, connect-flow UX edges, one test gap) and do not warrant blocking merge.

Issue #1289, opened 2026-08-12
[Bug] Media-generation tasks: excessive token usage, long runtimes, and repeated tool failures


codeacme17:
# [Bug] Media-generation tasks: excessive token usage, long runtimes, and repeated tool failures

## Summary

A customer reported that ad-creative / media-generation tasks (a) take very long to complete, (b) consume too many tokens, and (c) hit repeated tool errors. We verified two production task traces for the same prompt (a telco marketing request: "generate ad creatives to celebrate a subscriber milestone", ultimately 3 creatives x 2 aspect ratios) and confirmed the report. The causes are reproducible from the current codebase; this issue consolidates the production evidence, the root causes with code references, and a ranked fix list.

## Production evidence

Two runs of the same prompt on one production deployment:

| | Task A (`auto` -> DAG) | Task B (`balanced` -> DAG) |
|---|---|---|
| Status | COMPLETED (delivered 6 images) | COMPLETED (delivered **no images**) |
| Wall time | 28 min (~25 min excluding a user-reply wait) | 45 s |
| LLM calls | 106 | 4 |
| Input tokens | **6,521,187** | 33,045 |
| Output tokens | 70,782 | 3,021 |
| Avg input per LLM call | ~32-37k tokens | ~8k tokens |
| Cache hits | `cached_tokens: 0` on the large majority of calls | n/a |

**Task A — tool-call profile:**

| Tool | Calls | Total time |
|---|---|---|
| `understand_media` (vision inspection) | 10 | **511 s** (avg 51 s) |
| `generate_image` | 7 | 61 s |
| `execute_python_code` | 16 | 41 s |
| `execute_command` | 8 | 24 s |
| `edit_image` | 8 | ~0 s (**all 8 failed instantly**) |
| others (`get_file_info`, `list_image_models`, `web_search`, ...) | 13 | ~4 s |

**Task A — tool failures:** 10 total; **9 are the same error**: `No available image models with edit capabilities` (8x `edit_image`, 1x `generate_image` with `images=` which delegates to the edit path). After each failure the model re-tried variations and eventually fell back to editing images via `execute_python_code` (16 calls).

[...]

PR #1294, opened 2026-08-12
fix(image): gate image tools on the models' declared abilities

### LINKED ISSUE #1295: [Bug] Image tools are offered even when no configured model can serve them
Sub-issue of #1289 (B1 + B2), split out so it can be tracked and closed on its own.

## Problem

`edit_image` is registered unconditionally, regardless of whether any configured
image model actually has the `edit` ability
(`src/xagent/core/tools/adapters/vibe/image_tool.py:78-94`). When a deployment has
no edit-capable model, every `edit_image` call is doomed before it starts, and the
error it returns is a bare conclusion with no way for the model to self-correct:
`"No available image models with edit capabilities"`
(`src/xagent/core/tools/core/image_tool.py:337-358,695-700`).

In a production ad-creative task this produced 8 consecutive instant `edit_image`
failures, after which the model abandoned the tool and edited images through
`execute_python_code` 16 more times. Roughly 24 tool calls, each costing a full
LLM round trip with the entire history and tool schemas re-sent — the dominant
contributor to that task's token spend and wall time.

Worth recording, because it rules out the cheaper fix: **the description already
said editing was unavailable.** `_edit_model_info_text` falls back to the literal
string `"No image models with edit capabilities available"`
(`core/image_tool.py:297`) and that text is formatted into the `edit_image`
description. The model ignored the passive note and blind-retried anyway.
Rewording the description is therefore not sufficient — an unusable tool has to
leave the schema.

## Two further gaps found while fixing this

1. **The side door.** `generate_image(images=...)` delegates to the edit path
   (`core/image_tool.py:553`), and the `generate_image` description teaches the
   `images` parameter in three places. Withholding `edit_image` alone still leaves
   this route into the same failure.
2. **The symmetric hole on the generate side.** `if not image_models: return []`
   in `create_image_tools_from_config` covers "zero image models", but nothing
   covers "models exist, none has the `generate` ability" — `generate_image` is
   still registered and fails the same way. The gate belongs on declared
   abilities, not on model count.
3. **Default-model branches skip the ability check.** In both `_get_model` and
   `_get_edit_model`, the explicit `model_id` branch verifies `has_ability`, but
   the `_default_*_model` branch trusts its input unconditionally
   (`core/image_tool.py:327-328,350-351`). A generate-only model configured as
   `default_edit_model` bypasses the check, reaches the provider, and raises
   `RuntimeError` there instead of returning the friendly error — which, combined
   with the unpassed `retry_on` predicate (E1 in #1289), gets retried with
   exponential backoff.

## Acceptance

- [x] No model with the `edit` ability → `edit_image` is not in the tool schema.
- [x] No model with the `generate` ability → `generate_image` is not in the tool schema.
- [x] Neither ability available → neither `generate_image` nor `edit_image` is
      offered; the read-only `list_image_models` stays and reports each model's
      abilities, so the absence can be explained.
- [x] When editing is unavailable, the `generate_image` description states it
      explicitly, so the delegating `images` parameter is not treated as a route
      to editing.
- [x] The "no usable model" errors enumerate the configured models with their
      abilities and state what to do instead of retrying.
- [x] The default-model branches reject a model that explicitly denies the ability.
- [x] Tests cover both directions (tool withheld / tool present) rather than only
      the failure case.

## Update: the neither-ability criterion changed during review

It originally required that *no* image tool be offered when neither ability
exists. Review pushed back on that (xorbitsai/xagent#1294): the rationale for
pulling a tool out of the schema is that models blind-retry a doomed tool, and
that only justifies removing tools that can *fail*. `list_image_models` is
read-only and cannot fail, and it is the one remaining way for the agent — or a
human reading the trace — to answer "why is there nothing here". Dropping it left
silent absence as the only signal.

It is kept, and `list_available_models` now reports per-model `abilities` with
`available` reflecting whether the model can actually serve anything, instead of
the previous hardcoded `available: True` that contradicted the very reason for
keeping it.

## Not in scope

Deployment-side provisioning (granting the image model the edit ability, or
registering an edit-capable model) is tracked separately. This issue is only about
the framework no longer offering a tool nothing can serve.

bluefish-08:
Closes #1295. Sub-issue of #1289.

## Background

A production ad-creative task called `edit_image` 8 times in a row, each failing
instantly with `No available image models with edit capabilities`, then worked
around it by editing images through `execute_python_code` 16 more times. That
deployment has a single image model configured with `abilities = ["generate"]`.

The framework problem is that `edit_image` is registered unconditionally, with no
regard for whether any configured model can serve it. Worth recording: the
description already said so — `_edit_model_info_text` falls back to the literal
string `No image models with edit capabilities available` and that text is
formatted into the tool description. Production shows the model ignored that
passive note and blind-retried anyway, so rewording is not a fix; the tool has to
leave the schema.

Those ~24 doomed and detour calls each cost a full LLM round trip (the whole
history plus the tool schemas re-sent), which is where the bulk of that task's
token spend and its 28-minute wall time came from.

## Changes

1. **Capability gating** (`vibe/image_tool.py` `get_tools`) — no model with
   `generate` means `generate_image` is not registered; no model with `edit`
   means `edit_image` is not registered; neither means an empty tool list. When
   editing is unavailable, `generate_image` gets a leading prohibition in its
   description, because `generate_image(images=...)` delegates to the edit path
   and is the side door into the same failure.
2. **Self-correctable errors** (`core/image_tool.py`) — both "no usable model"
   errors now enumerate the configured models with their abilities and state what
   to do next. The original wording is kept as the prefix so existing log
   searches still match.
3. **Ability check on the default-model branches** of `_get_model` and
   `_get_edit_model` — the explicit `model_id` branch verified `has_ability`, but
   the `_default_*_model` branch trusted its input unconditionally. A
   generate-only model configured as `default_edit_model` therefore bypassed the
   check, reached the provider, and raised `RuntimeError` there instead of
   returning the friendly error. The check is deliberately lenient: it rejects
   only a model that explicitly reports lacking the ability, and still trusts one
   without a `has_ability` method.

## Why the long description template is left alone

`GENERATE_IMAGE_DESCRIPTION` teaches the `images` parameter in three separate
places. Trimming each would mean adding placeholders to the template for a much
wider diff, and existing tests assert against the class constant directly.
Instead, editing-unavailable deployments get one leading prohibition, backed by
`edit_image` being absent from the schema and by the self-correctable error. When
editing is available the description is byte-identical to before.

## Out of scope

- **Deployment-side follow-up**: grant the image model the edit ability, or
  register an edit-capable model. This change only guarantees that a tool nothing
  can serve is no longer offered to the model.
- Other items from the parent issue (retry predicate never passed to
  `create_retry_wrapper`, redundant fields in the tool-result observation,
  serial media tool execution, planner missing the deliverable type) are not in
  this PR.

## Verification

- New `TestImageToolCapabilityGating` with 7 cases: `edit_image` absent and the
  prohibition present when nothing can edit; both tools present and the
  description unchanged when a model can edit; `generate_image` absent when only
  editing is available; empty list when neither is available; a
  `default_edit_model` that denies `edit` is not trusted; both error strings
  enumerate abilities.
- Updated 3 existing cases that asserted the old contract (tool count with empty
  models; two exact error-string comparisons relaxed to substring).
- `tests/core/tools/` passes (exit 0). `test_command_path_guard_bash.py` and
  `test_command_executor.py` have pre-existing failures, reproduced on
  `origin/main` with these changes stashed to confirm they are unrelated.
- `ruff check`, `ruff format`, `mypy`, and the full pre-commit hook set pass.

### DIFF
--- a/src/xagent/core/tools/adapters/vibe/image_tool.py
+++ b/src/xagent/core/tools/adapters/vibe/image_tool.py
@@ -69,10 +69,10 @@ class ImageGenerationTool(ImageGenerationToolCore):
         """Get all tool instances."""
         # A note inside the description is not enough: models blind-retry a
         # registered-but-doomed tool, so an unusable one has to leave the schema.
-        can_generate = self.has_generate_capable_model()
-        can_edit = self.has_edit_capable_model()
-        if not can_generate and not can_edit:
-            return []
+        # list_image_models stays either way — it is read-only, cannot fail, and
+        # is the only way to answer "why is there nothing here".
+        can_generate = self._get_model() is not None
+        can_edit = self._get_edit_model() is not None
 
         tools = []
 
@@ -110,7 +110,7 @@ class ImageGenerationTool(ImageGenerationToolCore):
             ImageGenerationFunctionTool(
                 self.list_available_models,
                 name="list_image_models",
-                description="List all available image generation models, including model ID, availability status, and detailed description information (Note: model information is already provided in the generate_image tool description)",
+                description="List all available image generation models, including model ID, availability status, and detailed description information",
             )
         )
 
--- a/src/xagent/core/tools/adapters/vibe/image_tool.py
+++ b/src/xagent/core/tools/adapters/vibe/image_tool.py
@@ -67,31 +67,52 @@ class ImageGenerationTool(ImageGenerationToolCore):
 
     def get_tools(self) -> list:
         """Get all tool instances."""
-        # Format descriptions with model information
-        generate_description = self.GENERATE_IMAGE_DESCRIPTION.format(
-            self._model_info_text
-        )
-        edit_description = self.EDIT_IMAGE_DESCRIPTION.format(
-            self._edit_model_info_text
-        )
+        # A note inside the description is not enough: models blind-retry a
+        # registered-but-doomed tool, so an unusable one has to leave the schema.
+        can_generate = self.has_generate_capable_model()
+        can_edit = self.has_edit_capable_model()
+        if not can_generate and not can_edit:
+            return []
 
-        tools = [
-            ImageGenerationFunctionTool(
-                self.generate_image,
-                name="generate_image",
-                description=generate_description,
-            ),
-            ImageGenerationFunctionTool(
-                self.edit_image,
-                name="edit_image",
-                description=edit_description,
-            ),
+        tools = []
+
+        if can_generate:
+            generate_description = self.GENERATE_IMAGE_DESCRIPTION.format(
+                self._model_info_text
+            )
+            if not can_edit:
+                generate_description = (
+                    "IMAGE EDITING IS UNAVAILABLE here: no configured image model "
+                    "has the edit ability, so edit_image is not offered and passing "
+                    "images fails. Render each deliverable from a text prompt in "
+                    "one call.\n\n" + generate_description
+                )
+            tools.append(
+                ImageGenerationFunctionTool(
+                    self.generate_image,
+                    name="generate_image",
+                    description=generate_description,
+                )
+            )
+
+        if can_edit:
+            tools.append(
+                ImageGenerationFunctionTool(
+                    self.edit_image,
+                    name="edit_image",
+                    description=self.EDIT_IMAGE_DESCRIPTION.format(
+                        self._edit_model_info_text
+                    ),
+                )
+            )
+
+        tools.append(
             ImageGenerationFunctionTool(
                 self.list_available_models,
                 name="list_image_models",
                 description="List all available image generation models, including model ID, availability status, and detailed description information (Note: model information is already provided in the generate_image tool description)",
-            ),
-        ]
+            )
+        )
 
         return tools
 
--- a/src/xagent/core/tools/core/image_tool.py
+++ b/src/xagent/core/tools/core/image_tool.py
@@ -324,7 +324,9 @@ Images are automatically saved to workspace.
                 return None
 
         # Use configured default generate model
-        if self._default_generate_model:
+        if self._default_generate_model and not self._denies_ability(
+            self._default_generate_model, "generate"
+        ):
             return self._default_generate_model
 
         # Fallback: return first available model with generate capability
@@ -346,8 +348,11 @@ Images are automatically saved to workspace.
                 logger.warning(f"Model {model_id} does not support editing")
                 return None
 
-        # Use configured default edit model
-        if self._default_edit_model:
+        # A default configured by hand is trusted unless the model itself denies
+        # the ability; otherwise it reaches the provider and raises there instead.
+        if self._default_edit_model and not self._denies_ability(
+            self._default_edit_model, "edit"
+        ):
             return self._default_edit_model
 
         # Fallback: return first available model with edit capability
@@ -356,6 +361,46 @@ Images are automatically saved to workspace.
                 return model
 
         return None
+
+    @staticmethod
+    def _denies_ability(model: Any, ability: str) -> bool:
+        """True only when the model explicitly reports it lacks ``ability``."""
+        has_ability = getattr(model, "has_ability", None)
+        return callable(has_ability) and not has_ability(ability)
+
+    def has_generate_capable_model(self) -> bool:
+        """Whether any configured model can generate. Gates generate_image."""
+        return self._get_model() is not None
+
+    def has_edit_capable_model(self) -> bool:
+        """Whether any configured model can edit. Gates edit_image registration."""
+        return self._get_edit_model() is not None
+
+    def _available_models_summary(self) -> str:
+        entries = []
+        for model_id, model in self._image_models.items():
+            abilities = getattr(model, "abilities", None)
+            if not isinstance(abilities, (list, tuple)):
+                abilities = []
+            joined = ", ".join(str(ability) for ability in abilities) or "unknown"
+            entries.append(f"{model_id} (abilities: {joined})")
+        return "; ".join(entries) or "none configured"
+
+    def _no_edit_model_error(self) -> str:
+        return (
+            "No available image models with edit capabilities. "
+            f"Configured image models: {self._available_models_summary()}. "
+            "Image editing is unavailable in this deployment: render what you need "
+            "from a text prompt with generate_image instead of retrying edit_image."
+        )
+
+    def _no_generate_model_error(self) -> str:
+        return (
+            "No available image models configured. "
+            f"Configured image models: {self._available_models_summary()}. "
+            "Retry without model_id to use the default, or pick a model listed above "
+            "that has the generate ability."
+        )
 
     def _resolve_image_path(self, image_input: str) -> str:
         """
@@ -570,7 +615,7 @@ Images are automatically saved to workspace.
             if not image_model:
                 return {
                     "success": False,
-                    "error": "No available image models configured",
+                    "error": self._no_generate_model_error(),
                     "image_path": None,
                 }
 
@@ -695,7 +740,7 @@ Images are automatically saved to workspace.
             if not image_model:
                 return {
                     "success": False,
-                    "error": "No available image models with edit capabilities",
+                    "error": self._no_edit_model_error(),
                     "image_path": None,
                 }
 
--- a/src/xagent/core/tools/core/image_tool.py
+++ b/src/xagent/core/tools/core/image_tool.py
@@ -382,10 +382,17 @@ Images are automatically saved to workspace.
             entries.append(f"{model_id} (abilities: {joined})")
         return "; ".join(entries) or "none configured"
 
-    def _no_edit_model_error(self) -> str:
-        # The realistic trigger is generate_image(images=...) delegating here on a
-        # generate-only deployment, so never answer it with "use generate_image".
-        if self._get_model() is not None:
+    def _no_edit_model_error(self, model_id: Optional[str] = None) -> str:
+        # Each branch has to name a different way out, or the model retries the
+        # exact call it just made — the failure this whole error exists to stop.
+        if model_id and self._get_edit_model() is not None:
+            remedy = (
+                f"Model {model_id} cannot edit: retry edit_image without model_id, "
+                "or pick one listed above that has the edit ability."
+            )
+        elif self._get_model() is not None:
+            # generate_image(images=...) delegates here, so never answer that
+            # call with "use generate_image".
             remedy = (
                 "Image editing is unavailable in this deployment: retry "
                 "generate_image without images and describe the change in the "
@@ -401,12 +408,20 @@ Images are automatically saved to workspace.
             f"Configured image models: {self._available_models_summary()}. {remedy}"
         )
 
-    def _no_generate_model_error(self) -> str:
+    def _no_generate_model_error(self, model_id: Optional[str] = None) -> str:
+        if model_id and self._get_model() is not None:
+            remedy = (
+                f"Model {model_id} cannot generate: retry generate_image without "
+                "model_id, or pick one listed above that has the generate ability."
+            )
+        else:
+            remedy = (
+                "No configured model can generate images: stop retrying image "
+                "tools and report that to the user."
+            )
         return (
             "No available image models with generate capabilities. "
-            f"Configured image models: {self._available_models_summary()}. "
-            "Retry without model_id to use the default, or pick a model listed above "
-            "that has the generate ability."
+            f"Configured image models: {self._available_models_summary()}. {remedy}"
         )
 
     def _resolve_image_path(self, image_input: str) -> str:
@@ -622,7 +637,7 @@ Images are automatically saved to workspace.
             if not image_model:
                 return {
                     "success": False,
-                    "error": self._no_generate_model_error(),
+                    "error": self._no_generate_model_error(model_id),
                     "image_path": None,
                 }
 
@@ -747,7 +762,7 @@ Images are automatically saved to workspace.
             if not image_model:
                 return {
                     "success": False,
-                    "error": self._no_edit_model_error(),
+                    "error": self._no_edit_model_error(model_id),
                     "image_path": None,
                 }
 
--- a/src/xagent/core/tools/core/image_tool.py
+++ b/src/xagent/core/tools/core/image_tool.py
@@ -317,21 +317,19 @@ Images are automatically saved to workspace.
         """Get image model with generate capability by ID or default model."""
         if model_id and model_id in self._image_models:
             model = self._image_models[model_id]
-            if hasattr(model, "has_ability") and model.has_ability("generate"):
+            if self._has_ability(model, "generate"):
                 return model
             else:
                 logger.warning(f"Model {model_id} does not support generation")
                 return None
 
         # Use configured default generate model
-        if self._default_generate_model and not self._denies_ability(
-            self._default_generate_model, "generate"
-        ):
+        if self._has_ability(self._default_generate_model, "generate"):
             return self._default_generate_model
 
         # Fallback: return first available model with generate capability
         for model in self._image_models.values():
-            if hasattr(model, "has_ability") and model.has_ability("generate"):
+            if self._has_ability(model, "generate"):
                 return model
 
         return None
@@ -342,61 +340,70 @@ Images are automatically saved to workspace.
         """Get image model with edit capability by ID or default edit model."""
         if model_id and model_id in self._image_models:
             model = self._image_models[model_id]
-            if hasattr(model, "has_ability") and model.has_ability("edit"):
+            if self._has_ability(model, "edit"):
                 return model
             else:
                 logger.warning(f"Model {model_id} does not support editing")
                 return None
 
-        # A default configured by hand is trusted unless the model itself denies
-        # the ability; otherwise it reaches the provider and raises there instead.
-        if self._default_edit_model and not self._denies_ability(
-            self._default_edit_model, "edit"
-        ):
+        # A hand-configured default still has to declare the ability, or it
+        # reaches the provider and raises there instead of returning our error.
+        if self._has_ability(self._default_edit_model, "edit"):
             return self._default_edit_model
 
         # Fallback: return first available model with edit capability
         for model in self._image_models.values():
-            if hasattr(model, "has_ability") and model.has_ability("edit"):
+            if self._has_ability(model, "edit"):
                 return model
 
         return None
 
     @staticmethod
-    def _denies_ability(model: Any, ability: str) -> bool:
-        """True only when the model explicitly reports it lacks ``ability``."""
+    def _has_ability(model: Any, ability: str) -> bool:
+        """Capable only when the model itself declares the ability."""
         has_ability = getattr(model, "has_ability", None)
-        return callable(has_ability) and not has_ability(ability)
-
-    def has_generate_capable_model(self) -> bool:
-        """Whether any configured model can generate. Gates generate_image."""
-        return self._get_model() is not None
-
-    def has_edit_capable_model(self) -> bool:
-        """Whether any configured model can edit. Gates edit_image registration."""
-        return self._get_edit_model() is not None
+        return callable(has_ability) and bool(has_ability(ability))
 
     def _available_models_summary(self) -> str:
         entries = []
         for model_id, model in self._image_models.items():
             abilities = getattr(model, "abilities", None)
-            if not isinstance(abilities, (list, tuple)):
-                abilities = []
-            joined = ", ".join(str(ability) for ability in abilities) or "unknown"
+            if abilities is None:
+                joined = "unknown"
+            elif not isinstance(abilities, (list, tuple)):
+                logger.warning(
+                    "Model %s reports a non-sequence abilities value of type %s",
+                    model_id,
+                    type(abilities).__name__,
+                )
+                joined = "unknown"
+            else:
+                joined = ", ".join(str(ability) for ability in abilities) or "none"
             entries.append(f"{model_id} (abilities: {joined})")
         return "; ".join(entries) or "none configured"
 
     def _no_edit_model_error(self) -> str:
+        # The realistic trigger is generate_image(images=...) delegating here on a
+        # generate-only deployment, so never answer it with "use generate_image".
+        if self._get_model() is not None:
+            remedy = (
+                "Image editing is unavailable in this deployment: retry "
+                "generate_image without images and describe the change in the "
+                "prompt instead."
+            )
+        else:
+            remedy = (
+                "No configured model can edit or generate images: stop retrying "
+                "image tools and report that to the user."
+            )
         return (
             "No available image models with edit capabilities. "
-            f"Configured image models: {self._available_models_summary()}. "
-            "Image editing is unavailable in this deployment: render what you need "
-            "from a text prompt with generate_image instead of retrying edit_image."
+            f"Configured image models: {self._available_models_summary()}. {remedy}"
         )
 
     def _no_generate_model_error(self) -> str:
         return (
-            "No available image models configured. "
+            "No available image models with generate capabilities. "
             f"Configured image models: {self._available_models_summary()}. "
             "Retry without model_id to use the default, or pick a model listed above "
             "that has the generate ability."
--- a/src/xagent/core/tools/core/image_tool.py
+++ b/src/xagent/core/tools/core/image_tool.py
@@ -246,9 +246,9 @@ Images are automatically saved to workspace.
         default_model_lines = []
         other_model_lines = []
         for model_id, model in self._image_models.items():
-            if hasattr(model, "has_ability") and model.has_ability("generate"):
+            if self._has_ability(model, "generate"):
                 description = self._model_descriptions.get(model_id, "")
-                edit_marker = " ✎" if model.has_ability("edit") else ""
+                edit_marker = " ✎" if self._has_ability(model, "edit") else ""
                 is_default = model_id == default_generate_id
                 default_marker = " ⭐[DEFAULT]" if is_default else ""
 
@@ -275,7 +275,7 @@ Images are automatically saved to workspace.
         default_edit_lines = []
         other_edit_lines = []
         for model_id, model in self._image_models.items():
-            if hasattr(model, "has_ability") and model.has_ability("edit"):
+            if self._has_ability(model, "edit"):
                 description = self._model_descriptions.get(model_id, "")
                 is_default = model_id == default_edit_id
                 default_marker = " ⭐[DEFAULT]" if is_default else ""
@@ -872,10 +872,19 @@ Images are automatically saved to workspace.
         """
         try:
             models_info = []
-            for model_id in self._image_models.keys():
+            for model_id, model in self._image_models.items():
+                # When capability gating withholds every image tool this listing
+                # is the only thing left to explain why, so it has to report what
+                # each model can actually serve rather than a flat "available".
+                abilities = [
+                    ability
+                    for ability in ("generate", "edit")
+                    if self._has_ability(model, ability)
+                ]
                 model_info = {
                     "model_id": model_id,
-                    "available": True,
+                    "available": bool(abilities),
+                    "abilities": abilities,
                     "description": self._model_descriptions.get(model_id, ""),
                 }
                 models_info.append(model_info)
--- a/tests/core/tools/test_image_tool.py
+++ b/tests/core/tools/test_image_tool.py
@@ -948,6 +948,33 @@ class TestImageToolCapabilityGating:
         assert "retry edit_image without model_id" in edit["error"]
         assert "editing is unavailable" not in edit["error"].lower()
 
+    def test_listing_reports_what_each_model_can_actually_serve(
+        self, mock_image_models, mock_workspace
+    ):
+        # list_image_models is the only tool left in this state, so it has to
+        # explain the absence rather than claim everything is available.
+        self._set_abilities(mock_image_models, [])
+        tool = ImageGenerationTool(mock_image_models, workspace=mock_workspace)
+
+        listed = tool.list_available_models()["models"]
+
+        assert [m["available"] for m in listed] == [False, False]
+        assert [m["abilities"] for m in listed] == [[], []]
+
+        mock_image_models["model1"].has_ability = Mock(
+            side_effect=lambda a: a == "edit"
+        )
+        edit_only = ImageGenerationTool(
+            mock_image_models, workspace=mock_workspace
+        ).list_available_models()["models"]
+
+        assert edit_only[0] == {
+            "model_id": "model1",
+            "available": True,
+            "abilities": ["edit"],
+            "description": "",
+        }
+
     def test_a_model_without_has_ability_is_not_trusted(self, mock_workspace):
         # fail closed: an object that never declares the ability cannot serve it,
         # whether it arrives as a default or from the configured mapping.
--- a/tests/core/tools/test_image_tool.py
+++ b/tests/core/tools/test_image_tool.py
@@ -238,7 +238,8 @@ class TestImageGenerationTool:
         result = await tool.generate_image("A test prompt")
 
         assert result["success"] is False
-        assert "No available image models configured" in result["error"]
+        assert "No available image models with generate capabilities" in result["error"]
+        assert "Configured image models:" in result["error"]
         assert result["image_path"] is None
 
     @pytest.mark.asyncio
@@ -467,8 +468,7 @@ class TestImageGenerationTool:
         """Test that no generate_image tool is offered when no model can generate"""
         image_tool = ImageGenerationTool({}, {}, workspace=mock_workspace)
 
-        assert image_tool.get_tools() == []
-        assert "No image models" in image_tool._model_info_text
+        assert "generate_image" not in {t.metadata.name for t in image_tool.get_tools()}
 
     def test_model_info_text_generation(self, mock_workspace):
         """Test that model info text is generated correctly"""
@@ -702,8 +702,8 @@ class TestCreateImageTool:
         """Test create_image_tool function with empty models"""
         tools = create_image_tool({}, workspace=mock_workspace)
 
-        # Nothing can generate or edit, so no image tool is offered at all.
-        assert tools == []
+        # Nothing can generate or edit, so only the read-only listing remains.
+        assert {t.metadata.name for t in tools} == {"list_image_models"}
 
     def test_create_image_tool_with_descriptions(
         self, mock_image_models, model_descriptions, mock_workspace
@@ -813,11 +813,11 @@ class TestImageToolCapabilityGating:
         tools = ImageGenerationTool(
             mock_image_models, workspace=mock_workspace
         ).get_tools()
-        names = {tool.name for tool in tools}
+        names = {tool.metadata.name for tool in tools}
 
         assert "edit_image" not in names
         assert "generate_image" in names
-        generate = next(tool for tool in tools if tool.name == "generate_image")
+        generate = next(t for t in tools if t.metadata.name == "generate_image")
         assert "IMAGE EDITING IS UNAVAILABLE" in generate.description
 
     def test_edit_image_offered_when_a_model_can_edit(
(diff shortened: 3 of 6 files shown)


src/xagent/core/tools/core/image_tool.py
```diff
@@ -357,6 +362,46 @@ def _get_edit_model(
...
+            joined = ", ".join(str(ability) for ability in abilities) or "unknown"
+            entries.append(f"{model_id} (abilities: {joined})")
+        return "; ".join(entries) or "none configured"
+
+    def _no_edit_model_error(self) -> str:
+        return (
+            "No available image models with edit capabilities. "
+            f"Configured image models: {self._available_models_summary()}. "
+            "Image editing is unavailable in this deployment: render what you need "
+            "from a text prompt with generate_image instead of retrying edit_image."
```
rogercloud: **Moderate — the remediation is circular in its most likely trigger state.**

This message unconditionally recommends `generate_image`. But the realistic trigger for `_no_edit_model_error()` is *not* "both abilities missing" — in that state `edit_image` is unreachable via the schema anyway thanks to this PR's gating. The realistic trigger is the ordinary `can_generate=True, can_edit=False` deployment: `generate_image` is registered, the model calls it with `images=[...]`, and `generate_image` delegates to `edit_image()` (`if images is not None: return await self.edit_image(...)`) before checking anything itself. The model then gets "use `generate_image` instead of retrying `edit_image`" **as the result of a `generate_image` call** — advice it cannot act on, from a tool it just used.

Suggest branching on availability, e.g.:

```python
def _no_edit_model_error(self) -> str:
    base = (
        "No available image models with edit capabilities. "
        f"Configured image models: {self._available_models_summary()}. "
    )
    if self.has_generate_capable_model():
        return base + (
            "Image editing is unavailable in this deployment: call generate_image "
            "WITHOUT the images argument and render what you need from a text prompt."
        )
    return base + "Image editing is unavailable in this deployment."
```

Also: no test exercises `generate_image(..., images=...)` in the generate-only state. The existing delegation test configures a model with both abilities, so this branch is uncovered.

src/xagent/core/tools/core/image_tool.py
```diff
@@ -357,6 +362,46 @@ def _get_edit_model(
...
+        )
+
+    def _no_generate_model_error(self) -> str:
+        return (
+            "No available image models configured. "
```
rogercloud: **Minor — this lead sentence is false in the state the PR's own test creates.**

"No available image models configured." is reached whenever no model has the `generate` ability (or the requested `model_id` lacks it) — including when models *are* configured. `test_generate_error_lists_abilities` (`tests/core/tools/test_image_tool.py:887-897`) configures `model1` with `abilities=["edit"]` and produces exactly this text, immediately contradicted by the `Configured image models: model1 (abilities: edit)` sentence that follows.

Suggest "No available image models with generate capabilities.", mirroring `_no_edit_model_error`.

src/xagent/core/tools/core/image_tool.py
```diff
@@ -357,6 +362,46 @@ def _get_edit_model(
...
+        """Whether any configured model can edit. Gates edit_image registration."""
+        return self._get_edit_model() is not None
+
+    def _available_models_summary(self) -> str:
+        entries = []
+        for model_id, model in self._image_models.items():
+            abilities = getattr(model, "abilities", None)
+            if not isinstance(abilities, (list, tuple)):
+                abilities = []
+            joined = ", ".join(str(ability) for ability in abilities) or "unknown"
```
rogercloud: **Minor — an explicitly empty `abilities` list renders as "unknown".**

`", ".join(...) or "unknown"` maps `abilities == []` to `"(abilities: unknown)"`, conflating "this model declares zero abilities" (a known fact, and exactly the state that makes the tool unavailable) with "we could not determine its abilities". For a message whose whole purpose is to let the model self-correct, that is the wrong signal.

```python
joined = ", ".join(str(a) for a in abilities) if abilities else "none"
```

Uncovered: the neither-ability test only asserts `get_tools() == []`, never this string.

src/xagent/core/tools/core/image_tool.py
```diff
@@ -357,6 +362,46 @@ def _get_edit_model(
...
+    def has_edit_capable_model(self) -> bool:
+        """Whether any configured model can edit. Gates edit_image registration."""
+        return self._get_edit_model() is not None
+
+    def _available_models_summary(self) -> str:
+        entries = []
+        for model_id, model in self._image_models.items():
+            abilities = getattr(model, "abilities", None)
+            if not isinstance(abilities, (list, tuple)):
+                abilities = []
```
rogercloud: **Minor / hygiene** — a non-list `abilities` (a `set`, or a malformed model class) is silently coerced to `[]` and then reported as "unknown", masking the misconfiguration rather than surfacing it. Currently unreachable in production — every real `BaseImageModel` subclass honours the abstract property contract and returns a real sequence — so this is defensive-code polish only. A `logger.warning` here would make a future non-conforming model class debuggable.

src/xagent/core/tools/core/image_tool.py
```diff
@@ -357,6 +362,46 @@ def _get_edit_model(
 
         return None
 
+    @staticmethod
+    def _denies_ability(model: Any, ability: str) -> bool:
+        """True only when the model explicitly reports it lacks ``ability``."""
+        has_ability = getattr(model, "has_ability", None)
+        return callable(has_ability) and not has_ability(ability)
```
rogercloud: **Minor / hygiene — two opposite defaults for the same edge case in one file.**

`_denies_ability` treats a model with no `has_ability` method as *not denying* (trusted, passes the gate). The explicit-`model_id` branches (~L320, ~L334, ~L345, ~L360) treat the same missing method as *not having the ability* (rejected). Same unknown, opposite conclusions, depending only on whether the model arrived as a default or as an explicit id.

Not a live risk: every real model class (`DashScopeImageModel`, `XinferenceImageModel`, `OpenAIImageModel`, `GeminiImageModel`, `ImageModelAdapter`) implements `has_ability` concretely, so only a non-conforming test double can reach it. Worth picking one policy and stating it in the docstring so the next reader does not have to reconstruct that both branches are intentional. Neither default is currently tested.

src/xagent/core/tools/core/image_tool.py
```diff
@@ -357,6 +362,46 @@ def _get_edit_model(
...
+
+    def has_generate_capable_model(self) -> bool:
+        """Whether any configured model can generate. Gates generate_image."""
+        return self._get_model() is not None
```
rogercloud: **Simplification** — `has_generate_capable_model()` and `has_edit_capable_model()` are pure one-line `is not None` wrappers, each with exactly one caller (`src/xagent/core/tools/adapters/vibe/image_tool.py:72-73`). Inlining `self._get_model() is not None` / `self._get_edit_model() is not None` at those call sites removes 6 lines of indirection with no behaviour change.

(Not suggesting a combined `_has_capable_model(ability)` helper — dispatching on the ability name costs more code than the two wrappers save.)

src/xagent/core/tools/adapters/vibe/image_tool.py
```diff
@@ -67,31 +67,52 @@ async def _download_image(
...
-                self.edit_image,
-                name="edit_image",
-                description=edit_description,
-            ),
+        # A note inside the description is not enough: models blind-retry a
+        # registered-but-doomed tool, so an unusable one has to leave the schema.
+        can_generate = self.has_generate_capable_model()
+        can_edit = self.has_edit_capable_model()
+        if not can_generate and not can_edit:
+            return []
```
rogercloud: **Design question (non-blocking)** — this early return also drops `list_image_models`, which the PR description states and tests deliberately, so I read it as intentional rather than an oversight.

But the stated rationale — "models blind-retry a registered-but-doomed tool, so an unusable one has to leave the schema" — only justifies removing tools that *fail*. `list_image_models` is read-only and cannot fail; it is the one remaining way for the agent (or a human reading the trace) to answer "why is there nothing available here", and dropping it means the agent's only signal is silent absence. Was keeping it considered and rejected for a specific reason? If the concern is that an empty listing is itself confusing, returning it with an explicit "no models with generate or edit ability are configured" body would preserve the diagnostic without re-introducing a failing tool.

tests/core/tools/test_image_tool.py
```diff
@@ -463,25 +463,12 @@ def test_generate_image_tool_schema_includes_reference_images(self, image_tool):
...
-                break
 
-        assert generate_tool is not None
-
-        # Check that the description handles no models case
-        description = generate_tool.description
-        assert "Available models" in description
-        assert "No image models available" in description
+        assert image_tool.get_tools() == []
+        assert "No image models" in image_tool._model_info_text
```
rogercloud: **Minor (test quality)** — this assertion pins state that is dead on this path. `_model_info_text` is only read inside the `if can_generate:` branch of `get_tools()`, which this scenario never enters, so its content has no observable effect here. The assertion would keep passing even if the text became meaningless, and it couples the test to a private attribute. The `get_tools() == []` assertion above is the real behavioural assertion; consider dropping this one.

tests/core/tools/test_image_tool.py
```diff
@@ -812,3 +794,104 @@ def test_list_models_tool_functionality(self, mock_image_models, mock_workspace)
...
+        self, mock_image_models, mock_workspace
+    ):
+        self._set_abilities(mock_image_models, ["generate"])
+        tool = ImageGenerationTool(mock_image_models, workspace=mock_workspace)
+
+        result = await tool.edit_image(prompt="add a headline", image_url="file:abc123")
+
+        assert result["success"] is False
+        assert "model1 (abilities: generate)" in result["error"]
+        assert "generate_image" in result["error"]
```
rogercloud: **Minor (test quality) — tautological assertion.** `"generate_image"` is a hardcoded literal inside `_no_edit_model_error()`, emitted regardless of whether `generate_image` is actually registered, so this passes identically in the state where the recommendation is invalid (see my comment on `_no_edit_model_error`). It reads as verification that the error "points at generate_image" — per the test name — but verifies nothing about whether that target exists.

If the intent is to test the recommendation, assert it against a tool that is genuinely generate-capable *and* add a negative case where it is not.

tests/core/tools/test_image_tool.py
```diff
@@ -238,7 +238,7 @@ async def test_generate_image_with_no_models(self, mock_workspace):
         result = await tool.generate_image("A test prompt")
 
         assert result["success"] is False
-        assert result["error"] == "No available image models configured"
+        assert "No available image models configured" in result["error"]
```
rogercloud: **Minor (test quality)** — loosening `==` to `in` to accommodate the new tail drops all regression coverage of that tail: the `Configured image models: ...` enumeration and the remediation hint are now unasserted on this path. A second, narrower assertion keeps the flexibility while restoring coverage:

```python
assert "Configured image models:" in result["error"]
```

Same applies to `tests/core/tools/core/test_image_tool_core.py:226`.

tests/core/tools/test_image_tool.py
```diff
@@ -812,3 +794,104 @@ def test_list_models_tool_functionality(self, mock_image_models, mock_workspace)
...
+    ):
+        self._set_abilities(mock_image_models, ["generate"])
+        tool = ImageGenerationTool(
+            mock_image_models,
+            workspace=mock_workspace,
+            default_edit_model=mock_image_models["model1"],
+        )
+
+        assert tool.has_edit_capable_model() is False
+        assert "edit_image" not in {t.name for t in tool.get_tools()}
```
rogercloud: **Minor (coverage gap)** — this covers the edit side (a generate-only model configured as `default_edit_model` is correctly rejected), but there is no symmetric test for an edit-only model configured as `default_generate_model`. `has_generate_capable_model()` combined with a denying `default_generate_model` has zero references in either test file, even though `_get_model` and `_get_edit_model` received identical changes — so one half of the new `_denies_ability` wiring is unverified. A near-copy of this test with `default_generate_model=` and `abilities=["edit"]` would close it.

tests/core/tools/test_image_tool.py
```diff
@@ -812,3 +794,104 @@ def test_list_models_tool_functionality(self, mock_image_models, mock_workspace)
...
+
+    def test_edit_image_withheld_when_no_model_can_edit(
+        self, mock_image_models, mock_workspace
+    ):
+        self._set_abilities(mock_image_models, ["generate"])
+
+        tools = ImageGenerationTool(
+            mock_image_models, workspace=mock_workspace
+        ).get_tools()
+        names = {tool.name for tool in tools}
```
rogercloud: **Cosmetic** — the new tests use `tool.name` while the rest of this file uses `tool.metadata.name` (e.g. line 459). Both are correct (`metadata.name` is built from `name`); worth matching the surrounding convention for consistency. Applies to lines 816, 831 and 871.

rogercloud (commented):
## Summary

This PR moves image-tool availability from prose to schema. `ImageGenerationTool.get_tools()` now consults `has_generate_capable_model()` / `has_edit_capable_model()` and registers only the tools that can actually succeed: no generate-capable model means no `generate_image`, no edit-capable model means no `edit_image`, and neither means an empty tool list. It also adds an explicit ability check to the previously-unchecked `default_generate_model` / `default_edit_model` branches of `_get_model()` / `_get_edit_model()` (via `_denies_ability`), and replaces the two flat "no available image models" strings with self-correctable errors that enumerate the configured models and their abilities.

## Approach verdict

**Acceptable with reservations** — gating at the schema rather than in description text is the right fix for the observed failure (8 blind `edit_image` retries followed by 16 `execute_python_code` fallbacks), and reusing `_get_model()` / `_get_edit_model()` as both the runtime resolver and the registration gate is a genuine single source of truth: the adapter and the core cannot disagree about what is available.

The reservation is that it closes the front door and leaves a side door guarded only by prose — the exact pattern the issue's own reasoning says does not work:

- **The `images=` path on `generate_image` is still reachable when editing is unavailable.** `generate_image` delegates to `edit_image()` as soon as `images is not None`, before any of its own model checks. In the common `can_generate=True, can_edit=False` deployment, `generate_image` is registered, and a model that passes `images=` lands in the edit path. The only guard is the prepended "IMAGE EDITING IS UNAVAILABLE" prose in the description — the same class of mitigation this PR argues is insufficient. To be fair, this degrades gracefully (a structured `_no_edit_model_error()` tool result, not a crash or a silent wrong answer), so it is not a repeat of the original bug — but the recommendation text it returns is wrong in this state (see the first line-level finding), and this whole path is untested for the generate-only case.

- **`src/xagent/web/api/tools.py:120-133` becomes unreachable in exactly the deployments this PR targets.** `_create_tool_info()` has an `elif tool_name == "edit_image":` branch that independently re-derives "does any model have the `edit` ability" and, when not, marks the tool `status: "missing_capability"` with a remediation message for the admin. That branch only fires if a tool object literally named `edit_image` is present in what `ToolFactory.create_all_tools()` returns — which this PR now guarantees it is not, whenever the gate trips. Net effect: the admin tool listing loses the "add an image model with editing support" explanation entirely, and `edit_image` just silently disappears from the list. That file is untouched here and no test covers the interaction. Either the branch should be updated to render a synthetic disabled row, or (cleaner) the adapter should expose the reason so the web layer stops re-deriving the same predicate. At minimum this deserves a follow-up issue rather than being left as dead code.

- **Two opposite leniency defaults for the same edge case live in the same file.** `_denies_ability` (used on the default-model branches) treats a model with no `has_ability` method as *trusted*; the explicit-`model_id` branches treat the same missing method as *not having the ability*. Verified this is not a live risk — every real model class (`DashScopeImageModel`, `XinferenceImageModel`, `OpenAIImageModel`, `GeminiImageModel`, `ImageModelAdapter`) implements `has_ability` concretely, so only a non-conforming test double can hit it. Hygiene/consistency point, not a blocker, but pick one default and document it.

- **Open design question (not a defect):** dropping `list_image_models` along with the others when neither ability exists is deliberate and tested, per the PR description. But the stated rationale — "models blind-retry a doomed tool" — only justifies removing tools that *fail*. `list_image_models` is read-only and cannot fail; it is the one tool that could let the agent (or a human reading the trace) answer "why is there nothing here". Was keeping it considered and rejected for a specific reason?

## Line-level findings

### Moderate

1. **`src/xagent/core/tools/core/image_tool.py:393-394` — the remediation text is circular in its most likely trigger state.** `_no_edit_model_error()` unconditionally says "render what you need from a text prompt with `generate_image` instead of retrying `edit_image`". The realistic trigger is not "both abilities missing" (there, `edit_image` is unreachable via the schema anyway) — it is `can_generate=True, can_edit=False`, where `generate_image(images=...)` delegates into `edit_image()`. The model then receives "use `generate_image` instead" *as the result of a `generate_image` call*. The message should branch on `has_generate_capable_model()`: when generate is available, tell the model to retry `generate_image` **without** `images`; otherwise omit the suggestion. No test covers `generate_image(..., images=...)` in the generate-only state (the existing test that exercises the delegation configures a model with both abilities).

2. **`src/xagent/core/tools/core/image_tool.py:187-199` — the broad `except Exception` around `create_image_tool()` now swallows a much more consequential failure.** `create_image_tools_from_config` wraps the whole construction — which now includes `get_tools()` and therefore every `model.has_ability(...)` call — in `except Exception as e: logger.warning(...); return []`. A model whose `has_ability` raises now silently removes *all* image tools, including the always-safe `list_image_models`, leaving only a warning line. Meanwhile the direct `create_image_tool()` path has no exception handling at all and propagates the same error raw. Same failure, two opposite outcomes depending on entry point, and neither is tested. (Out of the diff's line range, so no inline comment — but it is a direct consequence of moving ability checks into the registration path.)

### Minor

3. **`src/xagent/core/tools/core/image_tool.py:399` — "No available image models configured." is factually wrong when models are configured but lack the `generate` ability.** The PR's own new test `test_generate_error_lists_abilities` (`tests/core/tools/test_image_tool.py:887-897`) configures a real `model1` with `abilities=["edit"]` and produces exactly this misleading lead sentence, immediately contradicted by the enumeration that follows it.

4. **`src/xagent/core/tools/core/image_tool.py:385` — an explicitly empty `abilities` list renders as `"unknown"`.** `", ".join(...) or "unknown"` collapses "this model declares zero abilities" (a known fact) into "we could not determine its abilities" (an unknown). Render `"none"` for `abilities == []`. Untested — the neither-ability test only asserts the tool list is `[]`, never the error string.

5. **`src/xagent/core/tools/core/image_tool.py:383-384` — non-list `abilities` is silently coerced to `[]`.** A malformed model class then reports "unknown" instead of surfacing the misconfiguration. Currently unreachable in production (the `BaseImageModel.abilities` contract returns a real sequence), so purely defensive-code hygiene — but a `logger.warning` would be cheap.

6. **`src/xagent/core/tools/core/image_tool.py:366-369` — `_denies_ability`'s "missing method ⇒ trusted" default contradicts the explicit-`model_id` branches** (lines ~320, ~334, ~345, ~360), which treat the same case as "lacks the ability". See the body note above; consistency nit only.

### Test quality

7. **`tests/core/tools/test_image_tool.py:471` — `assert "No image models" in image_tool._model_info_text` pins dead internal state.** On this path `_model_info_text` is never read: it is only consumed inside the `if can_generate:` branch, which this scenario by construction never enters. The behavioral assertion on the previous line (`get_tools() == []`) is the real test; this one would keep passing even if the text became meaningless.

8. **`tests/core/tools/test_image_tool.py:848` — `assert "generate_image" in result["error"]` is tautological.** That string is a hardcoded literal in `_no_edit_model_error()` regardless of whether `generate_image` is actually registered; the assertion passes identically in the state where the recommendation is invalid. Combined with finding 1, this test gives false confidence that the recommendation is validated.

9. **`tests/core/tools/core/test_image_tool_core.py:226` and `tests/core/tools/test_image_tool.py:241` — loosening `==` to `in` drops regression coverage of the new tail.** The enumerated-models list and the remediation hint are now completely unasserted on these two paths. Adding a second assertion (e.g. `assert "Configured image models:" in result["error"]`) restores it without re-pinning the exact string.

10. **Missing symmetric coverage for the generate side.** `test_default_edit_model_that_denies_edit_is_not_trusted` (`tests/core/tools/test_image_tool.py:850-861`) covers a generate-only model configured as `default_edit_model`, but there is no counterpart for an edit-only model configured as `default_generate_model` — `has_generate_capable_model()` combined with a denying `default_generate_model` has zero references in either test file, even though `_get_model` and `_get_edit_model` were changed identically.

11. **Style: the new tests use `tool.name` (`tests/core/tools/test_image_tool.py:816`, `831`, `871`) while the rest of the file uses `tool.metadata.name`.** Both are correct; worth matching the surrounding file.

## Out of scope, but worth a follow-up issue

`src/xagent/skills/builtin/static-visual-design/SKILL.md` (~lines 190-193 and ~326) and `references/static-ad-art-direction.md` (~line 296) still instruct the agent unconditionally to call `edit_image` or to pass reference images via `images=`. That skill text is injected into context verbatim with no awareness of which tools were registered, so on a generate-only deployment the agent is now told to use a tool that is not in its schema. Different subsystem, correctly not touched here — but it undercuts part of the win, and the PR description's "Out of scope" section does not mention it. Suggest filing a follow-up.

- `src/xagent/core/tools/core/image_tool.py` L371-377: shrink — `has_generate_capable_model()` / `has_edit_capable_model()` are one-line `is not None` wrappers with exactly one caller each (`src/xagent/core/tools/adapters/vibe/image_tool.py:72-73`); inline them at the call site. (Do **not** collapse them into a single `_has_capable_model(ability)` — the dispatch costs more than it saves.)
- Considered and rejected: merging `_no_edit_model_error` / `_no_generate_model_error` (L389-403). Their opening sentences differ structurally, not by an ability-name substitution, so a template needs a third parameter and just relocates the duplication — and both exact strings are asserted verbatim in tests.

net: -6 lines possible

## Prior review

`gemini-code-assist[bot]`'s inline comment on `src/xagent/core/tools/adapters/vibe/image_tool.py:113` is **still open**: `list_image_models`'s description retains "(Note: model information is already provided in the generate_image tool description)", which becomes misleading precisely in the new no-generate-capable-model state where `generate_image` is not registered. Not re-filing it inline — flagging it here as carried forward and unaddressed in the current head.

*Note on verification: CI is the source of truth for test execution; nothing above is based on a local test run.*

bluefish-08:
Thanks for the depth here — this caught a class of problem the original PR reasoned about in one direction only. All 13 inline threads are addressed and resolved; below is what changed, what I found while fixing, and the two things I deliberately left alone.

## Addressed

**Error messages** — `_no_edit_model_error` / `_no_generate_model_error` now branch instead of emitting one fixed string, because each state needs a different way out:

| State | What the model is told |
|---|---|
| `model_id` given, another model can serve it | retry without `model_id`, or pick one from the enumerated list |
| Editing unavailable, generation works | retry `generate_image` **without** `images` — never "use generate_image" as the answer to a `generate_image` call |
| Neither ability configured | stop retrying image tools and report it |

`_no_generate_model_error` got the same treatment: its old tail (`Retry without model_id to use the default`) was a no-op when no `model_id` was passed. The lead sentence is now "No available image models with generate capabilities", so it no longer contradicts the enumeration that follows it. Empty `abilities` renders as `none` rather than `unknown`, and a non-sequence `abilities` logs a warning instead of silently reporting `unknown`.

**`_denies_ability` → `_has_ability`** — one policy, fail closed. A model that does not declare the ability cannot serve it, whether it arrives as an explicit `model_id`, a configured default, or from the fallback scan. Verified no real model class is affected: every `BaseImageModel` subclass implements `has_ability` concretely, and `create_retry_wrapper` forwards it.

**`list_image_models` is kept** — your reasoning holds and is stronger than the PR's original rule. The "unusable tools must leave the schema" argument only covers tools that *fail*; a read-only listing cannot. Dropping it left silent absence as the agent's only signal. Note this makes the toolset length 1 rather than `[]` in the neither-ability state, so the acceptance line in #1295 ("no image tool is offered at all") no longer matches the implementation — I will update that issue.

**Admin listing** — `_withheld_edit_image_row()` in `web/api/tools.py` synthesizes the disabled row when capability gating keeps `edit_image` out of `create_all_tools()`, so the "add an image model with editing support" remediation survives. It only ever returns a `missing_capability` row, never an available one.

**Simplification** — `has_generate_capable_model` / `has_edit_capable_model` inlined and deleted, as suggested.

**Tests** — dead `_model_info_text` assertion dropped; the tautological `"generate_image" in error` replaced with an assertion on the actual remediation text; the `Configured image models:` tail re-covered on both paths loosened from `==` to `in`; symmetric coverage added for a denying `default_generate_model`; `tool.name` → `tool.metadata.name`.

## Found while fixing

Your first finding had a second trigger the fix initially missed. `_get_edit_model(model_id)` returns `None` when the requested model cannot edit **even if another configured model can**, and the error did not look at `model_id` at all. So `edit_image(model_id="generate_only")` produced:

> No available image models with edit capabilities. Configured image models: gen_only (abilities: generate); **editor (abilities: generate, edit)**. Image editing is unavailable in this deployment…

— self-contradicting, and reachable in exactly the deployments where `edit_image` *is* registered. That is what the `model_id` branch in the table above fixes. Also added a test for a model object that never declares `has_ability`, which is what the `_has_ability` unification is for and had no coverage.

## Not changed

1. **The broad `except Exception` around `create_image_tool()`** (`adapters/vibe/image_tool.py`). You are right that a `has_ability` that raises now takes down every image tool including `list_image_models`, and that the direct `create_image_tool()` path propagates the same error raw. But that handler and the asymmetry both predate this PR, and reaching it requires a model class that violates the `BaseImageModel` contract. Widening this PR to restructure error handling in the tool-construction path would mix two concerns; happy to file it separately.

2. **The generate side of the admin listing.** For symmetry, `generate_image` disappearing from the listing has the same gap when no model declares `generate`. I left it because that state means image models are configured but none can generate an image, which no real provider produces — adding a second synthetic row and a `generate_image` branch to `_create_tool_info` would be code for a state that cannot occur. Reconsidering if you disagree.

## Follow-up

Agreed on `static-visual-design/SKILL.md` and `references/static-ad-art-direction.md` still instructing `edit_image` / `images=` unconditionally — that text is injected verbatim with no view of the registered toolset, so a generate-only deployment now gets told to use a tool absent from its schema. Different subsystem; filing separately rather than expanding this PR.

rogercloud (approved):
## Summary

This PR stops registering `generate_image` and `edit_image` when no configured image model actually declares the matching ability, instead of registering them unconditionally and relying on a passive warning in the tool description. That passive approach failed in production: a deployment without edit capability produced 8 doomed `edit_image` calls followed by 16 `execute_python_code` fallback attempts, because the model ignored the description and blind-retried a tool nothing could serve. The PR also closes a related resolver bug where `_get_model`/`_get_edit_model` only checked `has_ability` on the explicit `model_id` branch, so a misconfigured default model reached the provider and crashed instead of returning a friendly error.

## Update since last review

This is a re-review. Three commits were pushed addressing 13 inline threads plus the design-level concerns from the previous round. Outcome: **15 of 16 prior items are fixed**, and the remaining one is reasonably waived as pre-existing and not worsened by this diff. The fixes are substantive rather than cosmetic — notably the unification of two opposite leniency defaults into a single fail-closed `_has_ability`, the removal of the circular remediation text, and the new admin-listing withheld row.

## Approach

Acceptable with reservations — gating at the schema level rather than in prose is the right call, and routing both the runtime resolver and the registration gate through the same `_get_model()`/`_get_edit_model()` is a real single-source-of-truth win.

Design-level observations from this pass (net-new, not re-litigating fixed items):

- **Capability knowledge is now duplicated across two predicates.** The core tool has a fail-closed `_has_ability` (`src/xagent/core/tools/core/image_tool.py:361-365`, with a `getattr` + `callable` guard), while the admin listing re-derives the same question as a raw `"edit" in model.abilities` (`src/xagent/web/api/tools.py:125-127`). Not a live bug — every shipped model class inherits `BaseImageModel`, so both attributes always exist today — but the web layer should reuse `_has_ability` instead of restating the check in a less defensive form.
- **The gating is symmetric; the compensating admin row is not.** `_withheld_edit_image_row` (`src/xagent/web/api/tools.py:230-254`) covers `edit_image` only, so in an edit-only deployment `generate_image` silently vanishes from the admin listing with no explanation. This was explicitly disclosed and reasoned about in review comments on the premise that a generate-missing/edit-present state cannot occur for real providers. That premise isn't fully accurate — edit-only/instruction-based image models are a real category — but leaving it out of this PR's scope is a reasonable call given the disclosure. A tracking issue would close the loop.
- **`list_image_models` is kept alive for a purpose it cannot currently serve.** Keeping it in the neither-ability state is the right decision (it answers "why is there nothing here"), but its actual output in `list_available_models` (`src/xagent/core/tools/core/image_tool.py:866-891`, pre-existing) hardcodes `"available": True` for every model and emits no `abilities` field. See Major finding below.
- **The built-in skill still assumes `edit_image` exists.** `src/xagent/skills/builtin/static-visual-design/SKILL.md` and `references/static-ad-art-direction.md` instruct the agent to use `edit_image`/`images=` unconditionally, with no view of the registered toolset. Acknowledged by name in a review comment with an intent to file separately, but no tracking issue is confirmed and it isn't in the PR description's Out-of-scope list.

## Prior review — verification results

| # | Item | Status |
|---|---|---|
| 1 | Misleading "model information is already provided…" parenthetical (`adapters/vibe/image_tool.py:113`) | FIXED — removed |
| 2 | Circular remediation in `_no_edit_model_error` ("use generate_image" when that's what was called) | FIXED — branches correctly; covered by `test_delegated_edit_does_not_recommend_the_tool_that_was_called` |
| 3 | "No available image models configured." untrue in the configured-but-incapable state | FIXED — now "No available image models with generate capabilities." |
| 4 | Empty `abilities` rendering as "unknown" | FIXED — renders "none"; non-sequence `abilities` now warns instead of silently coercing |
| 5 | Two opposite leniency defaults across branches | FIXED — single fail-closed `_has_ability` at all 6 call sites |
| 6 | `has_generate_capable_model()`/`has_edit_capable_model()` indirection | FIXED — deleted, call sites inlined |
| 7 | Should `list_image_models` survive the neither-ability state? | FIXED — kept (toolset length 1), rationale now documented in a comment |
| 8 | Dead `_model_info_text` assertion in test | FIXED — removed, behavioral assertion retained |
| 9 | Tautological `"generate_image" in error` assertion | FIXED — replaced with a state-discriminating assertion plus a negative case |
| 10 | Loosened `==`→`in` assertions dropping "Configured image models:" coverage | FIXED — narrower assertion restored in both test files |
| 11 | No symmetric test for a denying `default_generate_model` | FIXED — `test_default_generate_model_that_denies_generate_is_not_trusted` |
| 12 | `tool.name` vs `tool.metadata.name` inconsistency | FIXED — `.metadata.name` throughout |
| 13 | `images=` delegation path untested for the generate-only case | FIXED — tested, and the remediation text is correct |
| 14 | `edit_image` silently disappearing from the admin listing | FIXED — `_withheld_edit_image_row`, wired unconditionally, with `test_withheld_edit_image_keeps_its_admin_row` |
| 15 | Broad `except Exception` around `create_image_tool()` swallowing a raising `has_ability` and dropping all image tools | WAIVED — see note below |

Note on the waived item: the correct location is `src/xagent/core/tools/adapters/vibe/image_tool.py:187-199` (the original finding cited `core/image_tool.py` — a path typo). Confirmed byte-identical to the base commit. The blast radius already existed pre-PR, since `__init__`/`_generate_model_info_text` already called `has_ability` inside the same try-wrapped construction; this PR only adds two more call sites under an existing umbrella. Not a regression, so not blocking — but the "happy to file separately" offer doesn't appear to have produced an actual issue yet, so a gentle nudge: please file it.

## New findings

### Major

**1. `list_image_models` output contradicts the reason it was kept — `src/xagent/core/tools/core/image_tool.py:866-891`**

`list_available_models` hardcodes `"available": True` for every configured model and emits no `abilities` field. In the exact neither-ability scenario that `list_image_models` is deliberately being preserved to explain, this is the only surviving diagnostic tool, and its output cannot answer "why is there nothing here" — worse, it affirmatively asserts that every model is available. No test asserts on `abilities` or exercises this tool in the neither-ability scenario.

This is pre-existing code untouched by the diff, but the PR's own design decision is what converts it from a latent gap into a live contradiction of the stated rationale. Suggested fix: add an `abilities` field per row, and reconsider the unconditional `available: True` so it reflects whether the model can actually serve anything.

### Minor

**2. Withheld admin row covers `edit_image` only — `src/xagent/web/api/tools.py:123,230-254`**

`generate_image` has no equivalent synthetic row, so it silently disappears from the admin listing in a generate-gated deployment. Disclosed and deliberate, on the premise that no real provider yields generate-missing/edit-present — a premise that edit-only/instruction-based image models arguably contradict. Reasonable scope call for this PR; recommend a tracking issue rather than expanding the diff.

**3. `generate_image`'s description still advertises editing below the prohibition — `src/xagent/core/tools/core/image_tool.py` (`GENERATE_IMAGE_DESCRIPTION`, ~lines 36, 38, 95) and `src/xagent/core/tools/adapters/vibe/image_tool.py:78-86`**

When editing is unavailable, an "IMAGE EDITING IS UNAVAILABLE" prohibition is prepended, but the unchanged template body further down still says to "use edit_image directly" and documents `images` as usable, and `images` remains a real parameter in the signature and JSON schema. Per the PR description this is a deliberate trade-off to avoid a much wider diff to the shared template, and it degrades gracefully — a structured, self-correcting error rather than the original blind schema-level retry. Genuine residual, low severity, not a repeat of the original bug.

**4. Two independent "can this model edit/generate" implementations — `src/xagent/web/api/tools.py:125-127` vs `src/xagent/core/tools/core/image_tool.py:361-365`**

The core path is fail-closed; the web listing's `"edit" in model.abilities` has no guard against a model missing `.abilities`. Not a live bug today (all shipped classes inherit `BaseImageModel`), but a non-conforming class would break the admin row outright instead of degrading. Suggested fix: have the web layer reuse `_has_ability`.

**5. `_generate_model_info_text` still uses the old inline pattern — `src/xagent/core/tools/core/image_tool.py:249,251,278`**

Uses `hasattr(model, "has_ability") and model.has_ability(...)` rather than the new shared `_has_ability` helper introduced elsewhere in the same file; line 251 has no guard at all and relies on an enclosing condition. Consistency/duplication nit.

**6. Built-in skill docs assume `edit_image` is registered — `src/xagent/skills/builtin/static-visual-design/SKILL.md` (~192, 326, 353-354) and `references/static-ad-art-direction.md` (~296)**

Both instruct the agent to use `edit_image`/`images=` unconditionally. Acknowledged in a review comment with an intent to file separately; no PR touches these files and no tracking issue is confirmed. Nudge to actually file it.

**7. Route-level wiring of the withheld row is untested — `tests/web/api/test_tools_api.py`, wiring at `src/xagent/web/api/tools.py:388-390`**

`test_withheld_edit_image_keeps_its_admin_row` calls the helper directly, and every route-level test in the file configures zero image models, so the wiring call short-circuits before reaching the interesting logic. Deleting the wiring line would not fail any current test. Suggested fix: one route-level test with an edit-incapable model configured.

**8. `enabled_map` can produce `enabled=True, status="missing_capability"` — `src/xagent/web/api/tools.py:~415-430`**

The loop unconditionally overwrites `enabled` from a stored `ToolConfig` row but only rewrites `status` in the disabled branch. A stale `ToolConfig(enabled=True)` row for `edit_image` combined with the new synthetic withheld row yields an internally inconsistent result, which the frontend renders as a plain green "Enabled" card (it reads `status`/`status_reason` into state but never renders them). The pattern predates this PR and is reachable via other unconditionally-registered tools such as `understand_media`, but the new synthetic row is what makes it newly reachable for `edit_image`.

**9. Inconsistent error response shape — `src/xagent/core/tools/core/image_tool.py` (`generate_image`/`edit_image`)**

The "no usable model" early-return error dicts omit `model_used`, while the exception-handler paths on the same methods include it. No in-repo consumer reads `model_used`, so this is cosmetic.

**10. An unrecognized `model_id` is indistinguishable from an ability mismatch — `src/xagent/core/tools/core/image_tool.py` (`_get_model`/`_get_edit_model`, `_no_generate_model_error`/`_no_edit_model_error`)**

A typo'd `model_id` silently falls through to default/fallback resolution with no distinct error; if that also fails, the model is told to stop retrying image tools and report to the user, even though the real fix may be a single misspelled id in an otherwise-capable deployment. This is current tested/expected behavior per `test_generate_image_with_nonexistent_model`. Narrow edge case, but it cuts against the PR's own self-correctability thesis.

Lean already. Two candidates were evaluated and both dropped: (a) merging `_no_edit_model_error`/`_no_generate_model_error` into one templated helper — already considered and rejected in the prior round with sound reasoning, and the code has since diverged further (3-branch vs 2-branch, non-parallel predicate chains), which reinforces that rejection; (b) a YAGNI flag on `_withheld_edit_image_row` for hardcoding one tool name — dropped because the helper contains no speculative generality (no config, registry, or plugin hooks); its narrow scope is the separate coverage discussion in minor finding #2, not premature abstraction.

---

Nothing above is based on a local test run — CI is the source of truth for test execution.

PR #1300, opened 2026-08-12
perf(tools): drop redundant metadata from artifact observations

bluefish-08:
Sub-issue of #1289 (A4).

## Problem

`format_tool_result_for_observation` appends a `repr` of the entire sanitized
result dict after the human-readable artifact lines. Those lines already carry
every file's identity — filename, `file_id`, and the Markdown reference — so the
appended dict repeats all of it and adds fields the model cannot act on.

The waste compounds. An observation stays in the conversation history, so every
byte is re-sent on each subsequent LLM call. In the production ad-creative trace
behind #1289 (106 LLM calls, 6.5M input tokens, `cached_tokens: 0` on most
calls), each generated image carried roughly 200 tokens of duplicate metadata
that were then re-billed for the rest of the task.

The video tool is the worse case: its result includes `raw_response`, the
provider's full API payload, which lands in the observation verbatim.

## Change

`OBSERVATION_REDUNDANT_KEYS` drops these top-level keys from the model-facing
observation only:

| Key | Why |
|---|---|
| `artifacts`, `generated_files` | file identity is already in the artifact lines |
| `raw_response` | full provider payload, unbounded size |
| `video_url` | temporary provider URL; the downloaded file is the artifact |
| `usage`, `task_metric`, `request_id`, `saved_to_workspace` | telemetry the model cannot act on |

The tool result itself is untouched — this only narrows what the model sees.

## What is deliberately kept

- **`last_frame_url`** — not telemetry. It is an input to the next
  `generate_video` call (`last_frame_image_url`, `video_tool.py:688`), and the
  last frame is never downloaded, so it has no artifact of its own. Dropping it
  would silently break the opt-in `return_last_frame` continuation workflow.
- **`file_ref` / `file_refs`** — they carry `relative_path`, `download_url`, and
  `preview_url`, which the artifact lines do not.

Filtering is top-level only. No current tool nests these keys, and a full-tree
walk would risk stripping legitimate nested data.

## The failure path is redacted too

Correcting an earlier claim in this description: the no-artifact fallback used to
return `str(result)` — the *unsanitized* dict — so a failed video download leaked
the full `raw_response` and the signed `video_url` verbatim, which is exactly the
case this PR was written for. Both fallback returns now go through the same
sanitizer.

That path sheds only `raw_response`: without artifact lines, the URLs and
filenames are the model's only handle on the result, so `video_url` and
`saved_to_workspace` stay visible for diagnosis.

Path redaction is unaffected: `_collect_known_local_paths` runs over the complete
result before this filter, so removing keys can only shrink the exposed surface,
never remove a path from the replacement table.

`generated_files` is safe to drop — `javascript_executor` also appends the names
to `output`, and every entry has a matching artifact line built in the same loop
(`build_generated_file_metadata`).

## Tests

Two new cases in `tests/core/tools/test_artifacts_observation.py`: a full
`generate_image` result (kept: markdown ref, `model_used`, `relative_path`;
dropped: artifacts/generated files/usage/metric/request id/workspace flag) and a
`generate_video` result asserting the 4 KB `raw_response` and the temporary video
URL never reach the observation while `last_frame_url` still does.

A third case covers the `artifacts: []` download-failure shape, and a fourth
asserts on `_observation_metadata`'s returned keys directly rather than through
substring matching.

Mutation-tested: removing any one of the eight keys from the set turns the suite
red, and so does adding `last_frame_url` back to it, reverting the failure path
to `str(result)`, or swapping either call site's exclusion set for the other.

Full `tests/core/tools` passes. `test_command_path_guard_bash.py` and
`test_command_executor.py` fail on `origin/main` too — unrelated.

### DIFF
--- a/src/xagent/core/tools/artifacts.py
+++ b/src/xagent/core/tools/artifacts.py
@@ -49,6 +49,21 @@ LOCAL_PATH_KEYS = {
     "output_dir",
     "output_path",
 }
+# Dropped from the model-facing observation only, never from the tool result.
+# The artifact lines already carry file identity; the rest is telemetry, raw
+# provider payloads, or transient URLs the model must not reference.
+OBSERVATION_REDUNDANT_KEYS = {
+    "artifacts",
+    "generated_files",
+    "image_url",
+    "last_frame_url",
+    "raw_response",
+    "request_id",
+    "saved_to_workspace",
+    "task_metric",
+    "usage",
+    "video_url",
+}
 
 GeneratedArtifactSnapshot = dict[Path, tuple[int, int]]
 
@@ -125,8 +140,18 @@ def format_tool_result_for_observation(tool_name: str, result: Any) -> str:
         + "\nUse the Markdown/chat form in assistant messages. "
         + "When writing HTML for Xagent preview, reference the same file_id "
         + "through the file preview service instead of local filesystem paths. "
-        + f"Sanitized result metadata: {sanitized}"
+        + f"Sanitized result metadata: {_observation_metadata(sanitized)}"
     )
+
+
+def _observation_metadata(sanitized: Any) -> Any:
+    if not isinstance(sanitized, dict):
+        return sanitized
+    return {
+        key: value
+        for key, value in sanitized.items()
+        if key not in OBSERVATION_REDUNDANT_KEYS
+    }
 
 
 def _format_artifact_lines(artifacts: list[Any]) -> list[str]:
--- a/src/xagent/core/tools/artifacts.py
+++ b/src/xagent/core/tools/artifacts.py
@@ -41,13 +41,19 @@ SAFE_FILE_REF_KEYS = {
     "relative_path",
     "size",
 }
+# Every media tool returns its artifact as <kind>_path alongside file_id/file_ref.
+# Missing keys were redacted only by coincidence: file_ref.file_path registers the
+# same string, which stops holding when building the FileRef fails.
 LOCAL_PATH_KEYS = {
     "absolute_path",
+    "audio_path",
     "file_path",
     "image_path",
     "local_path",
     "output_dir",
     "output_path",
+    "transcription_path",
+    "video_path",
 }
 # Dropped from the model-facing observation only, never from the tool result, and
 # only at the top level. Keep out anything the model still acts on: last_frame_url
--- a/src/xagent/core/tools/artifacts.py
+++ b/src/xagent/core/tools/artifacts.py
@@ -49,14 +49,12 @@ LOCAL_PATH_KEYS = {
     "output_dir",
     "output_path",
 }
-# Dropped from the model-facing observation only, never from the tool result.
-# The artifact lines already carry file identity; the rest is telemetry, raw
-# provider payloads, or transient URLs the model must not reference.
+# Dropped from the model-facing observation only, never from the tool result, and
+# only at the top level. Keep out anything the model still acts on: last_frame_url
+# is an input to the next generate_video call and has no artifact of its own.
 OBSERVATION_REDUNDANT_KEYS = {
     "artifacts",
     "generated_files",
-    "image_url",
-    "last_frame_url",
     "raw_response",
     "request_id",
     "saved_to_workspace",
@@ -144,9 +142,7 @@ def format_tool_result_for_observation(tool_name: str, result: Any) -> str:
     )
 
 
-def _observation_metadata(sanitized: Any) -> Any:
-    if not isinstance(sanitized, dict):
-        return sanitized
+def _observation_metadata(sanitized: dict[str, Any]) -> dict[str, Any]:
     return {
         key: value
         for key, value in sanitized.items()
--- a/src/xagent/core/tools/artifacts.py
+++ b/src/xagent/core/tools/artifacts.py
@@ -52,7 +52,7 @@ LOCAL_PATH_KEYS = {
 # Dropped from the model-facing observation only, never from the tool result, and
 # only at the top level. Keep out anything the model still acts on: last_frame_url
 # is an input to the next generate_video call and has no artifact of its own.
-OBSERVATION_REDUNDANT_KEYS = {
+_OBSERVATION_EXCLUDED_KEYS = {
     "artifacts",
     "generated_files",
     "raw_response",
@@ -62,6 +62,9 @@ OBSERVATION_REDUNDANT_KEYS = {
     "usage",
     "video_url",
 }
+# Without artifact lines the URLs and filenames are the model's only handle on the
+# result, so the failure path only sheds the unbounded provider payload.
+_UNBOUNDED_PAYLOAD_KEYS = {"raw_response"}
 
 GeneratedArtifactSnapshot = dict[Path, tuple[int, int]]
 
@@ -122,32 +125,32 @@ def format_tool_result_for_observation(tool_name: str, result: Any) -> str:
     if not isinstance(result, dict):
         return str(result)
 
-    artifacts = result.get("artifacts")
-    if not isinstance(artifacts, list) or not artifacts:
-        return str(result)
-
     sanitized = sanitize_file_refs_for_observation(result)
 
-    artifact_lines = _format_artifact_lines(artifacts)
+    artifacts = result.get("artifacts")
+    artifact_lines = (
+        _format_artifact_lines(artifacts) if isinstance(artifacts, list) else []
+    )
     if not artifact_lines:
-        return str(sanitized)
+        # A failed download still reports through this path, so it has to be
+        # redacted too — it carries the same signed URLs and provider payloads.
+        return str(_observation_metadata(sanitized, _UNBOUNDED_PAYLOAD_KEYS))
 
+    metadata = _observation_metadata(sanitized, _OBSERVATION_EXCLUDED_KEYS)
     return (
         f"Tool '{tool_name}' produced displayable artifact(s):\n"
         + "\n".join(artifact_lines)
         + "\nUse the Markdown/chat form in assistant messages. "
         + "When writing HTML for Xagent preview, reference the same file_id "
-        + "through the file preview service instead of local filesystem paths. "
-        + f"Sanitized result metadata: {_observation_metadata(sanitized)}"
+        + "through the file preview service instead of local filesystem paths."
+        + (f" Sanitized result metadata: {metadata}" if metadata else "")
     )
 
 
-def _observation_metadata(sanitized: dict[str, Any]) -> dict[str, Any]:
-    return {
-        key: value
-        for key, value in sanitized.items()
-        if key not in OBSERVATION_REDUNDANT_KEYS
-    }
+def _observation_metadata(
+    sanitized: dict[str, Any], excluded: set[str]
+) -> dict[str, Any]:
+    return {key: value for key, value in sanitized.items() if key not in excluded}
 
 
 def _format_artifact_lines(artifacts: list[Any]) -> list[str]:
--- a/tests/core/tools/test_artifacts_observation.py
+++ b/tests/core/tools/test_artifacts_observation.py
@@ -180,6 +180,7 @@ def test_format_tool_result_for_observation_drops_redundant_image_metadata():
                 "filename": "creative.png",
                 "relative_path": "creative.png",
             },
+            "generated_files": ["creative.png"],
             "usage": {"prompt_tokens": 12, "total_tokens": 34},
             "task_metric": {"latency_ms": 8123},
             "request_id": "req-abc123",
@@ -191,9 +192,15 @@ def test_format_tool_result_for_observation_drops_redundant_image_metadata():
     assert "![creative.png](file:image-file-id)" in observation
     assert "gemini-2.5-flash-image" in observation
     assert "relative_path" in observation
-    for dropped in ("prompt_tokens", "latency_ms", "req-abc123", "saved_to_workspace"):
+    for dropped in (
+        "'artifacts'",
+        "generated_files",
+        "prompt_tokens",
+        "latency_ms",
+        "req-abc123",
+        "saved_to_workspace",
+    ):
         assert dropped not in observation
-    assert "inline" not in observation
 
 
 def test_format_tool_result_for_observation_drops_raw_video_provider_payload():
@@ -202,7 +209,7 @@ def test_format_tool_result_for_observation_drops_raw_video_provider_payload():
         {
             "success": True,
             "video_url": "https://provider.example/signed/clip.mp4?token=secret",
-            "last_frame_url": "https://provider.example/signed/frame.png?token=secret",
+            "last_frame_url": "https://frames.example/last-frame.png",
             "video_path": "/tmp/xagent/output/clip.mp4",
             "file_id": "video-file-id",
             "artifacts": [
@@ -224,6 +231,8 @@ def test_format_tool_result_for_observation_drops_raw_video_provider_payload():
     assert "1080p" in observation
     assert "xxxx" not in observation
     assert "provider.example" not in observation
+    # last_frame_url is an input to the next generate_video call, not telemetry.
+    assert "https://frames.example/last-frame.png" in observation
 
 
 def test_snapshot_generated_artifact_files_skips_files_deleted_before_stat(
--- a/tests/core/tools/test_artifacts_observation.py
+++ b/tests/core/tools/test_artifacts_observation.py
@@ -159,6 +159,73 @@ def test_format_tool_result_for_observation_normalizes_artifact_type_case():
     assert "Markdown/chat image" in observation
 
 
+def test_format_tool_result_for_observation_drops_redundant_image_metadata():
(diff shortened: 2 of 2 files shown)


src/xagent/core/tools/artifacts.py
```diff
@@ -125,10 +138,18 @@ def format_tool_result_for_observation(tool_name: str, result: Any) -> str:
         + "\nUse the Markdown/chat form in assistant messages. "
         + "When writing HTML for Xagent preview, reference the same file_id "
         + "through the file preview service instead of local filesystem paths. "
-        + f"Sanitized result metadata: {sanitized}"
+        + f"Sanitized result metadata: {_observation_metadata(sanitized)}"
```
rogercloud: **MAJOR (pre-existing gap, not a regression) — see the early-return fallbacks at lines 126-128 and 132-133, upstream of this line.**

This filtering only runs on the branch that reaches this line. When video download fails, `_build_video_artifacts` in `video_tool.py` returns `[]` rather than raising. The gate in `src/xagent/core/agent/context/execution.py:287` is `isinstance(result.get("artifacts"), list)`, which is `True` for `[]`, so `format_tool_result_for_observation` is still called — and `if not isinstance(artifacts, list) or not artifacts: return str(result)` (line 126-128) fires first, returning the **raw** dict before `sanitize_file_refs_for_observation` or `_observation_metadata` (this line) ever run. On a representative failure result, the full `raw_response` and the signed `video_url` (token in the query string) reach the model verbatim.

Note the PR description says the fallback returns `str(sanitized)`; it actually returns `str(result)` at line 127, so not even path/URL redaction applies on this branch.

This predates the PR, so it's not a regression, but it's exactly the "worse case" the PR description calls out for video. Suggestion: run sanitization + `_observation_metadata`-equivalent filtering on both fallback paths (lines 126-128 and 132-133), so `raw_response` can never reach an observation regardless of which branch is taken.

src/xagent/core/tools/artifacts.py
```diff
@@ -49,6 +49,19 @@
     "output_dir",
     "output_path",
 }
+# Dropped from the model-facing observation only, never from the tool result, and
+# only at the top level. Keep out anything the model still acts on: last_frame_url
+# is an input to the next generate_video call and has no artifact of its own.
+OBSERVATION_REDUNDANT_KEYS = {
```
rogercloud: **MINOR** — naming and visibility.

"Redundant" fits `artifacts` / `generated_files` (genuinely duplicated by the artifact lines above), but `raw_response` / `usage` / `task_metric` / `request_id` are unactionable telemetry rather than duplicated data — a different justification. `OBSERVATION_EXCLUDED_KEYS` would cover both.

Also, this is public but has exactly one consumer, the private `_observation_metadata`, and no importers anywhere else in the repo. `_OBSERVATION_EXCLUDED_KEYS` would match its actual scope and discourage out-of-tree mutation of prompt-shaping logic.

src/xagent/core/tools/artifacts.py
rogercloud: **MINOR (latent)** — if every key in `sanitized` were excluded, this renders the literal noise `Sanitized result metadata: {}`. Not reachable via any current tool (`image_tool.py`, `video_tool.py`, `pptx_tool.py` all keep at least `success`), so purely hygiene: skip the clause when the filtered dict is empty.

tests/core/tools/test_artifacts_observation.py
```diff
@@ -159,6 +159,82 @@ def test_format_tool_result_for_observation_normalizes_artifact_type_case():
     assert "Markdown/chat image" in observation
 
 
+def test_format_tool_result_for_observation_drops_redundant_image_metadata():
```
rogercloud, line 162: **MINOR** — coverage gap that maps directly to the MAJOR finding in `artifacts.py`.

Both new tests (and all existing ones) use a non-empty list of valid artifact entries. Neither `artifacts: []` — the video-download-failure shape, where the early-return fallback emits the raw unsanitized dict — nor the "artifacts present but no formattable lines" branch is exercised. A test asserting `raw_response` content is absent when `artifacts` is `[]` would have caught that bug and would pin the fix.

tests/core/tools/test_artifacts_observation.py
```diff
@@ -159,6 +159,82 @@ def test_format_tool_result_for_observation_normalizes_artifact_type_case():
...
+    assert "relative_path" in observation
+    for dropped in (
+        "'artifacts'",
+        "generated_files",
+        "prompt_tokens",
+        "latency_ms",
+        "req-abc123",
+        "saved_to_workspace",
+    ):
+        assert dropped not in observation
```
rogercloud, line 203: **MINOR** — these are substring checks against the whole formatted string rather than checks on what `_observation_metadata` actually returns. For `usage` / `task_metric` only nested values (`prompt_tokens`, `latency_ms`) are asserted, never the bare key names — one level indirect.

No false negative today, but calling `_observation_metadata` directly (importable in tests; no `__all__` restriction) and asserting on dict keys would be stronger and clearer, keeping the formatted-string assertions for the artifact-line rendering.

tests/core/tools/test_artifacts_observation.py
```diff
@@ -159,6 +159,82 @@ def test_format_tool_result_for_observation_normalizes_artifact_type_case():
...
+
+
+def test_format_tool_result_for_observation_drops_raw_video_provider_payload():
+    observation = format_tool_result_for_observation(
+        "generate_video",
+        {
+            "success": True,
+            "video_url": "https://provider.example/signed/clip.mp4?token=secret",
+            "last_frame_url": "https://frames.example/last-frame.png",
+            "video_path": "/tmp/xagent/output/clip.mp4",
```
rogercloud, line 213: **MINOR (pre-existing, not introduced here)** — `video_path` is not in `LOCAL_PATH_KEYS` (`src/xagent/core/tools/artifacts.py:44-51`), so this absolute local path is not redacted and survives verbatim into the observation. The test asserts the signed URL is redacted but says nothing about this path either way.

Out of scope for this PR; flagging because the new fixture surfaces it. Adding `video_path` to `LOCAL_PATH_KEYS` looks like a cheap follow-up.

rogercloud (approved):
## Summary

This PR narrows what `format_tool_result_for_observation` shows the model: on the success path, the trailing `Sanitized result metadata: {...}` blob is now filtered through a new `_observation_metadata` helper that drops an 8-key `OBSERVATION_REDUNDANT_KEYS` denylist — data already rendered as artifact lines (`artifacts`, `generated_files`), unbounded provider payloads (`raw_response`), and unactionable telemetry (`usage`, `task_metric`, `request_id`, `saved_to_workspace`, `video_url`). The tool result itself is untouched, so downstream consumers and the UI still see everything. Two new tests cover the image and video shapes, and `last_frame_url` is deliberately kept because it is an input to the next `generate_video` call rather than telemetry.

Overall this is a well-scoped change, and the reasoning about which keys are safe to drop is careful — the `last_frame_url` carve-out with its explanatory comment, and the choice to filter only the observation string rather than the result, are both the right calls.

## Approach

The denylist is the right shape for this problem, and it fails in the safe direction. Tool result payloads here are highly heterogeneous, so an allowlist would silently starve the model of legitimate per-tool fields as new tools appear; a denylist degrades toward verbosity instead of information loss. It also matches the file's existing idiom (`LOCAL_PATH_KEYS`), which keeps the module coherent.

One design-level reservation, offered as a note rather than a blocker: the denylist is maintained centrally in a generic formatter but encodes knowledge of specific tools' payload schemas, and it is already provably incomplete. `src/xagent/core/tools/adapters/vibe/a2a_agent_tool.py` returns a `raw` field (from `model_dump()`) that is the same class of unbounded provider payload as `raw_response`, and `raw` is not in the set. That tool does not currently emit `artifacts`, so it does not reach this code path today, but the pattern means every future tool that dumps a provider response must remember to register its key here. Broadening coverage is legitimately separate work; a follow-up issue to audit this list periodically may be worth opening.

(For the record: the asymmetry where `LOCAL_PATH_KEYS` filtering recurses into nested structures while this filtering is top-level only is already stated in the comment at `src/xagent/core/tools/artifacts.py:52-54`, so no action needed there.)

## Findings

### 1. MAJOR (pre-existing gap, not a regression introduced by this PR) — the fix misses the worst case it was written for

`src/xagent/core/tools/artifacts.py:126-128` (also `:132-133`)

When video download fails, `_build_video_artifacts` in `video_tool.py` returns `[]` rather than raising. The gate in `src/xagent/core/agent/context/execution.py:287` is `isinstance(result.get("artifacts"), list)`, which is `True` for `[]`, so `format_tool_result_for_observation` is still called. Inside, `if not isinstance(artifacts, list) or not artifacts: return str(result)` fires — and that returns the raw dict *before* `sanitize_file_refs_for_observation` or `_observation_metadata` ever run. On a representative video-download-failure result, the full `raw_response` payload and the signed `video_url` (token in the query string) both land in the model-facing observation verbatim.

This fallback predates this PR and isn't a regression, but it hits exactly the case the PR's own description calls out as the worst offender: video's `raw_response`. The description's claim that the fallback returns `str(sanitized)` is also inaccurate — it's `str(result)`, so not even path/URL redaction applies on this branch.

Suggested fix: apply sanitization plus `_observation_metadata`-equivalent filtering on both fallback return paths (`:126-128` and `:132-133`), not only on the success path. Given the fix is small and localized, and it's the exact scenario motivating this PR, I'd lean toward addressing it now rather than deferring — but not blocking merge on it since it doesn't make anything worse than main.

### 2. MINOR — coverage gap for the failure shapes

`tests/core/tools/test_artifacts_observation.py`

Directly related to finding 1: nothing in the file exercises `artifacts: []` (the video-download-failure shape) or the "artifacts present but no formattable lines" branch at `artifacts.py:132-133`. Every test, new and existing, uses a non-empty list of valid artifact entries. A test asserting `raw_response` content is absent when `artifacts` is `[]` would have caught finding 1 and would lock the fix in.

### 3. MINOR — `OBSERVATION_REDUNDANT_KEYS` name and visibility

`src/xagent/core/tools/artifacts.py:55`

"Redundant" is accurate for `artifacts` / `generated_files`, which really are duplicated by the artifact lines above them, but `raw_response` / `usage` / `task_metric` / `request_id` are not redundant — they are unactionable telemetry, a different justification. Something like `OBSERVATION_EXCLUDED_KEYS` would cover both cases. Separately, the constant is public despite having exactly one consumer, the private `_observation_metadata`, and no importers elsewhere in the repo; `_OBSERVATION_EXCLUDED_KEYS` would match its real scope and discourage out-of-tree mutation of prompt-shaping logic.

### 4. MINOR — assertions are one level indirect

`tests/core/tools/test_artifacts_observation.py:195-203`

`assert "generated_files" not in observation` and friends are substring checks against the whole formatted string rather than checks on what `_observation_metadata` returns. For `usage` / `task_metric` the tests only assert nested value substrings (`prompt_tokens`, `latency_ms`) and never the bare key names. There is no false negative today, but calling `_observation_metadata` directly (it is importable in tests — no `__all__` restriction) and asserting on dict keys would be both stronger and easier to read, with the formatted-string assertions kept for the artifact-line rendering.

### 5. MINOR (pre-existing) — `video_path` is not redacted

`tests/core/tools/test_artifacts_observation.py:213`

The new fixture includes `"video_path": "/tmp/xagent/output/clip.mp4"`. `video_path` is not in `LOCAL_PATH_KEYS` (`artifacts.py:44-51`, unchanged here), so the absolute local path survives into the observation. The test asserts the signed URL is redacted but says nothing either way about this path. Pre-existing gap and out of scope to fix here — flagging only because the new fixture happens to surface it, and adding `video_path` to `LOCAL_PATH_KEYS` looks like a cheap follow-up.

### 6. MINOR (latent) — empty metadata clause

`src/xagent/core/tools/artifacts.py:141`

If every key in `sanitized` were excluded, the observation would end with the literal `Sanitized result metadata: {}`. Not reachable through any current tool — `image_tool.py`, `video_tool.py`, and `pptx_tool.py` all include at least one non-excluded key such as `success` — so this is hygiene only: skip the clause when the filtered dict is empty.

Lean already. The `OBSERVATION_REDUNDANT_KEYS` constant is worth keeping as a named constant rather than inlined into the comprehension — it follows the file's existing `LOCAL_PATH_KEYS` pattern and carries rationale in its comment that would be lost otherwise.

---

Approving since finding 1 is a pre-existing gap rather than a regression, but I'd encourage addressing it (or opening a tracking issue) given it directly undercuts the PR's own stated motivation for the video case.

tests/core/tools/test_artifacts_observation.py
bluefish-08: Confirmed and left unfixed here, deliberately.

`video_path` is redacted in the common case only by accident: `_collect_known_local_paths` registers the same absolute path from `file_ref.file_path` (which *is* in `LOCAL_PATH_KEYS`) and the replacement then applies to every string value, `video_path` included. Verified both ways — with a `file_ref` present the value renders as `clip.mp4`; when `build_workspace_file_ref` raises and the code falls back to `get_file_id_from_path` (`video_tool.py`), `file_ref` is `None`, nothing registers the path, and the absolute path survives verbatim.

So it is not a stable redaction, it is a coincidental one, and `audio_path` (`music_tool.py`, `audio_tool.py`, `sound_effect_tool.py`) has the same shape. Adding both keys to `LOCAL_PATH_KEYS` is a two-line fix, but it is a path-disclosure change rather than a token change, and it deserves its own test for the `file_ref is None` fallback shape rather than riding along in a PR whose description is entirely about token cost. Filing separately — leaving this thread open until that issue exists.

bluefish-08:
Thanks — the MAJOR finding was the important one, and you were right that it undercut the PR's own motivation. Five of six threads are fixed and resolved; the sixth is answered in place and left open pending a separate issue.

## Fixed

**MAJOR — the failure path is now redacted too.** You are right on both counts: the fallback was `str(result)`, not `str(sanitized)`, so a failed video download leaked the full `raw_response` and the signed `video_url` verbatim — and my PR description asserted the opposite. Description corrected.

Both early returns now go through the sanitizer. The failure path uses a narrower exclusion set (`_UNBOUNDED_PAYLOAD_KEYS = {"raw_response"}`) rather than the full one: with no artifact lines rendered, the URLs and filenames are the model's only handle on the result, so dropping `video_url` there would trade a leak for an unusable observation. Restructured so the sanitize call happens before the branch, since both paths need it now.

**MINOR — naming and visibility.** `OBSERVATION_REDUNDANT_KEYS` → `_OBSERVATION_EXCLUDED_KEYS`. Your point stands that "redundant" only describes half the set; "excluded" covers both justifications, and the leading underscore matches its single private consumer.

**MINOR — empty metadata clause.** The `Sanitized result metadata:` clause is now omitted entirely when the filtered dict is empty, rather than rendering `{}`.

**MINOR — coverage gap for the failure shape.** Added a test for the `artifacts: []` download-failure shape asserting the 4 KB `raw_response` and the local path are gone while `video_url` survives. This is the test that would have caught the MAJOR finding.

**MINOR — indirect assertions.** Added a test calling `_observation_metadata` directly and asserting on returned key sets for both exclusion sets, keeping the formatted-string assertions for the artifact-line rendering.

## Mutation-tested

Reverting the failure path to `str(result)`, swapping either call site's exclusion set for the other, removing `raw_response` from the narrow set, removing any one of the eight keys from the full set, or adding `last_frame_url` back to it — each turns the suite red.

## Left open

**`video_path` redaction** — answered inline. Short version: it is redacted today only by coincidence (`file_ref.file_path` happens to register the same string), and that coincidence disappears on the `build_workspace_file_ref` failure path. `audio_path` has the same shape. Two-line fix, but it is a path-disclosure change needing its own test for the `file_ref is None` shape, so it is going in a separate issue rather than into a token-cost PR.

## Also filing separately

Your note on `a2a_agent_tool.py`'s `raw` field is well taken — same class of unbounded provider payload as `raw_response`, not in the set, and unreachable today only because that tool emits no `artifacts`. The broader point that a central denylist encodes per-tool schema knowledge is the real issue; a tracking issue to audit the list is the right shape for it.

tests/core/tools/test_artifacts_observation.py
bluefish-08: Fixed here after all — it was smaller than the separate-issue overhead.

`LOCAL_PATH_KEYS` now carries `video_path`, `audio_path`, and `transcription_path` alongside `image_path`. All four are the same shape: a media tool saves its artifact to the workspace and returns `<kind>_path` next to `file_id`/`file_ref`.

Test covers the shape that actually loses the coincidental redaction — `file_ref: None`, which is what `video_tool.py` / `audio_tool.py` produce when `build_workspace_file_ref` raises and they fall back to `get_file_id_from_path`. Mutation-tested per key: removing any one of the three turns it red. Full `tests/core/tools` passes, so no other tool depended on these keys surviving into the observation.

One deliberate omission: `translation_path` (`translate_json`). Same shape, but unlike the others it is declared as a return field in the model-facing tool description (`adapters/vibe/translate_json.py:63`), so redacting it without editing that description would leave the two disagreeing. That belongs with a change to `translate_json` rather than here.

PR #1301, opened 2026-08-12
feat: expand Slack connector with history search, reactions, and file upload

yiboyasss:
## Summary
- Add MCP tools to search/read channel, private-channel, thread, and DM history and topics (`slack_get_channel_history`, `slack_get_thread_replies`, `slack_get_channel_info`, `slack_list_direct_messages`, `slack_search_messages`), react to messages (`slack_add_reaction`, `slack_remove_reaction`), and upload files (`slack_upload_file`); `slack_post_message` gains an optional `thread_ts` to reply in a thread.
- Add the new required bot scopes (`channels:history`, `groups:read`, `groups:history`, `im:read`, `im:history`, `mpim:read`, `mpim:history`, `reactions:write`, `files:write`) to the built-in connector registry and to a new Alembic migration that upgrades already-installed connections and force-disconnects existing grants so users can reconnect and approve the new scopes.
- `slack_upload_file` restricts reads to an allowlisted directory (configurable via `XAGENT_SLACK_FILE_ALLOWED_DIRS`), mirroring the LinkedIn connector's image-upload guard, so the tool can't be used to exfiltrate arbitrary host files.

## Test plan
- [x] `pytest tests/web/tools/test_slack_mcp.py tests/alembic/` — new/updated unit and migration tests pass
- [x] `pytest tests/web/test_slack_oauth.py tests/web/test_slack_channel.py tests/web/api/test_public_mcp_connector_visibility.py` — no regressions in adjacent Slack/connector coverage
- [x] `ruff check`, `ruff format --check`, `mypy` on touched files
- [x] `alembic heads` resolves to a single head


src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +62,79 @@ def _normalize_channel(channel: str) -> str:
...
+            break
+    raise ValueError(
+        f"Could not resolve channel '{channel}' to an id — use slack_list_channels "
+        "or slack_list_direct_messages to look it up, or pass the channel id "
+        "directly."
+    )
+
+
+def _allowed_file_dirs() -> list[Path]:
+    raw_dirs = os.environ.get("XAGENT_SLACK_FILE_ALLOWED_DIRS", "")
```
rogercloud, line 128: MAJOR: `XAGENT_SLACK_FILE_ALLOWED_DIRS` is never populated. `src/xagent/web/tools/config.py` (the MCP-subprocess env builder) only hardcodes the LinkedIn equivalent (`XAGENT_LINKEDIN_IMAGE_ALLOWED_DIRS`) when constructing the subprocess env, and Slack's `launch_config` has no `static_env` entry either. So in production this always falls back to `Path.cwd().resolve()` of the subprocess: uploading from the task workspace can never succeed, and the effective allowlist becomes whatever the subprocess cwd happens to be — not the guarantee the docstring at line 625 makes. Suggest setting `env["XAGENT_SLACK_FILE_ALLOWED_DIRS"]` in `config.py` alongside the LinkedIn line, or better, sharing one helper and one env var between both connectors. No test covers the unset-env fallback branch, which is why this went unnoticed.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +62,79 @@ def _normalize_channel(channel: str) -> str:
...
+    if not raw_dirs.strip():
+        return [Path.cwd().resolve()]
+    return [
+        Path(raw_dir).expanduser().resolve()
+        for raw_dir in raw_dirs.split(",")
```
rogercloud, line 133: MINOR: each comma-separated entry is not stripped before `Path()` — the `.strip()` at line 110 only tests truthiness for the filter. Verified: a leading space produces a relative path joined onto cwd with a space component; a trailing space is preserved so the path never equals the intended directory. Both cases silently drop the entry from the allowlist. This confirms the earlier `gemini-code-assist` comment on this function, with one correction: it fails closed (over-restrictive), it never widens the allowlist, so it is not a security bypass. Fix: `Path(raw_dir.strip())`. The same function exists byte-for-byte in the LinkedIn connector, so a shared fix covers both.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +62,79 @@ def _normalize_channel(channel: str) -> str:
...
+    if not local_path.is_file():
+        raise FileNotFoundError(f"File not found: {file_path}")
+
+    allowed_dirs = _allowed_file_dirs()
+    for allowed_dir in allowed_dirs:
+        if local_path == allowed_dir or local_path.is_relative_to(allowed_dir):
+            return local_path
+
+    allowed_dirs_str = ", ".join(str(path) for path in allowed_dirs)
+    raise PermissionError(
```
rogercloud, line 167: MINOR: this message embeds the absolute allowed directory path(s) verbatim, and it reaches the caller/LLM unfiltered via `_error(str(e))`, putting host filesystem layout into the model transcript. Mirrors a pre-existing pattern in the LinkedIn connector, so not a regression — but present in new code. Suggest a generic "path is outside the allowed upload directories" message with the detail logged server-side only.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +62,79 @@ def _normalize_channel(channel: str) -> str:
...
+    """Resolve a channel id, bare name, or "#name" to a real Slack channel id.
+
+    Unlike chat.postMessage (which Slack resolves names for natively), the
+    conversations.*/reactions.*/files.* endpoints below require an actual
+    channel id — a name has to be looked up via conversations.list first.
+    DM and group-DM ids have no name form and always match
+    _SLACK_ID_PATTERN, so they pass through unchanged.
+    """
+    value = channel.strip()
+    if _SLACK_ID_PATTERN.match(value):
```
rogercloud, line 89: MINOR: `_SLACK_ID_PATTERN` matches Slack user IDs (`U...`) as readily as channel IDs (`C...`/`G...`), and this short-circuit does no prefix check. Since `slack_list_direct_messages` returns real user IDs in a `user` field, an agent that mistakes one for a channel reference passes this check and gets a raw `channel_not_found` from the API instead of an actionable error. Suggest restricting the short-circuit to channel-ID prefixes.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +62,79 @@ def _normalize_channel(channel: str) -> str:
     return f"#{value}"
 
 
+def _resolve_channel_id(channel: str) -> str:
```
rogercloud, line 79: MINOR: name→ID resolution is not memoized across tool calls, so every history/replies/info/reaction/upload call made by channel *name* re-runs the full paginated `conversations.list` scan (up to `MAX_PAGES` = 20 requests) against an endpoint this file's own comment documents as Tier-2 rate-limited (~20 req/min). An agent that resolves a channel by name then reacts to several messages by name in one task can rate-limit itself. Suggest caching resolution for the process lifetime or a short TTL.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+    channel: str,
+    limit: int = 50,
+    oldest: str = "",
+    latest: str = "",
+    cursor: str = "",
+) -> str:
+    """
+    Fetch recent messages from a channel, private channel, or DM.
+    channel: a channel/DM id, or a public/private channel name (with or
+    without "#") — names are resolved to an id first via slack_list_channels.
```
rogercloud: MINOR: docstring inaccuracy — internal name resolution (`_resolve_channel_id`) queries both public and private channel types, while `slack_list_channels` queries public channels only. A private channel is resolvable by name internally but not discoverable via the tool this points at. Worth correcting since this PR adds `groups:read` specifically to support private channels.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+        ]
+        return _success(
+            messages=messages,
+            has_more=bool(result.get("has_more")),
+            next_cursor=(result.get("response_metadata") or {}).get("next_cursor")
+            or "",
+        )
+    except Exception as e:
+        logger.error(f"Error fetching Slack channel history for {channel}: {e}")
+        return _error(str(e))
```
rogercloud, line 384: MINOR (simplification): this function and `slack_get_thread_replies` (below) are two ~30-line bodies sharing channel resolution, params/limit clamping, cursor handling, the try/except/`_error` wrapper, and the list-comprehension + `_success` return — differing only in endpoint, two extra params, and projected fields. A shared `_fetch_messages(path, extra_params, fields)` would cover both.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+    try:
+        for _ in range(MAX_PAGES):
+            params: dict[str, Any] = {
+                "types": "im,mpim",
+                "exclude_archived": "true",
+                "limit": 200,
+            }
+            if cursor:
+                params["cursor"] = cursor
+            result = _request("GET", "conversations.list", params=params)
```
rogercloud: MINOR: unlike `slack_list_channels`, which wraps its `_request` call so a failure on page 2+ returns already-collected results with `truncated=True`, this call has no equivalent tolerance on the same Tier-2-limited endpoint — any mid-pagination failure discards every page fetched so far and returns a bare error. Suggest mirroring the `slack_list_channels` truncate-on-failure behavior. No test covers this path either.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+    connector does not request. To search multiple conversations, call this
+    once per channel/DM (see slack_list_channels / slack_list_direct_messages).
+    """
+    needle = query.strip().lower()
+    if not needle:
+        return _error("query must not be empty")
+    try:
+        channel_id = _resolve_channel_id(channel)
+        max_scan = max(1, min(int(scan_limit), 1000))
+        max_matches = max(1, int(limit))
```
rogercloud: MINOR: no upper clamp here, unlike every other limit in this file (all clamp to 1000 or `MAX_CHANNELS`). Total matches are still hard-bounded by `max_scan` (≤ 1000) and `MAX_SEARCH_THREADS` (20 × 200 replies), so this is not literally unbounded — around ~5000 worst case — but it is inconsistent with the file's convention. Suggest `min(int(limit), 1000)`.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+                        "GET",
+                        "conversations.replies",
+                        params={"channel": channel_id, "ts": thread_ts, "limit": 200},
+                    )
+                except Exception as thread_exc:
+                    logger.warning(
+                        f"Slack thread search failed for {thread_ts}: {thread_exc}"
+                    )
+                    continue
+                threads_scanned += 1
```
rogercloud: MAJOR: `threads_scanned += 1` sits after the `except Exception: ... continue` above, so failed `conversations.replies` calls never consume the `MAX_SEARCH_THREADS` (=20) budget — only successful ones count. `_request` retries once with up to a 30s sleep on 429, and `threaded_parents` can hold up to the clamped scan limit (max 1000) entries, so one call against a rate-limited or heavily threaded channel can issue far more than 20 attempts and block for minutes to tens of minutes, with no caller-side way to bound total wall time in a synchronous tool. Fix: increment the counter before the request, not after success.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+            json_data={"channel": channel_id, "timestamp": timestamp, "name": name},
+        )
+        return _success(channel=channel_id, timestamp=timestamp, emoji_name=name)
+    except Exception as e:
+        logger.error(f"Error adding Slack reaction to {channel}:{timestamp}: {e}")
+        return _error(str(e))
+
+
+@mcp.tool()
+def slack_remove_reaction(channel: str, timestamp: str, emoji_name: str) -> str:
```
rogercloud, line 696: MINOR (simplification): this and `slack_add_reaction` (above) are near-identical ~13-line bodies differing only in the `reactions.add` / `reactions.remove` endpoint. A shared `_set_reaction(action, channel, timestamp, emoji_name)` saves ~10-14 lines with no behavior change. Also: `slack_remove_reaction` has no error-path test, unlike `slack_add_reaction`.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +265,417 @@ def slack_list_channels(
...
+        local_path = _resolve_allowed_file_path(file_path)
+        channel_id = _resolve_channel_id(channel)
+        resolved_filename = filename.strip() or local_path.name
+
+        init_result = _request(
+            "GET",
+            "files.getUploadURLExternal",
+            params={
+                "filename": resolved_filename,
+                "length": local_path.stat().st_size,
```
rogercloud: MINOR, two issues here: (a) `local_path.stat().st_size` is sent as `length` with no zero-byte check, so an empty file surfaces an opaque raw API error to the caller; (b) there is a TOCTOU window between the allowlist check in `_resolve_allowed_file_path` (which does correctly resolve symlinks on both sides) and the later `open("rb")` — a symlink swapped in during that window escapes the guard. Both low likelihood. Suggest rejecting zero-byte files explicitly, and opening the file first then stat-ing the open handle so the checked and read paths are the same object.

src/xagent/migrations/versions/20260812_add_slack_history_reactions_files_scopes.py
```diff
@@ -0,0 +1,177 @@
...
+            "Disconnected %d existing Slack grant(s) for the new required "
+            "history, reaction, and file-upload scopes. Affected users must "
+            "reconnect the Slack connector.",
+            result.rowcount,
+        )
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    _set_slack_scopes(bind, CURRENT_SCOPES)
```
rogercloud, line 232: MAJOR: this updates `public_mcp_apps.oauth_scopes` (3→12) and force-disconnects existing grants below, but never updates `oauth_providers.default_scopes`, which stays at the old 3 scopes on already-seeded databases. The app-scoped reconnect flow (with `app_id`) is fine since it merges fresh code-registry scopes, but `GET /api/auth/{provider}/login` without `app_id` uses only `db_provider.default_scopes` — minting a token with the stale 3 scopes that still displays as "Connected" (the connected-check only verifies token presence/expiry, not scope completeness) while every new tool fails at call time. Slack is also not in `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT`, even though Facebook was added there for exactly this failure mode when its scopes grew — which defeats the protection `_invalidate_existing_slack_grants` adds just below. Scope caveat: no current frontend call site drives the app_id-less path for Slack, so reaching it needs the raw endpoint, a script, or a future UI surface. Fix: update `default_scopes` in `upgrade()`/`downgrade()`, or add `"slack"` to `APPS_REQUIRING_APP_SCOPED_OAUTH_GRANT`.

rogercloud (changes requested):
## Summary

This PR expands the built-in Slack connector from 2 tools (list channels, post message) to 11, adding channel/thread/DM history reads, channel info, message search, reaction add/remove, and file upload, plus optional `thread_ts` on `post_message`. It requests 9 additional OAuth bot scopes and ships a migration that updates the app registry and force-disconnects existing grants so users must reconnect and approve the wider scope set. File upload is guarded by a directory allowlist modeled on the existing LinkedIn image-upload guard, intended to prevent the tool from exfiltrating arbitrary host files.

The overall shape is right and deliberately unambitious: one flat function per tool, shared `_request` / `_success` / `_error` helpers, no new abstraction layers, and a migration that faithfully clones the established HubSpot scope-migration pattern (same guards, same force-disconnect approach). This is not over-engineered. There are, however, four design-level gaps that the individual findings below are symptoms of.

**1. The security guard was copied instead of shared.** The file-allowlist guard was copy-pasted from the LinkedIn image-upload guard rather than extracted into a shared helper, and the copy introduced a new environment variable name that nothing populates — which is the root cause of F1, the most serious finding here. A single shared helper behind one generic env var would have made this correct by construction, and would also have fixed the whitespace-stripping bug (F6) for both connectors at once. A copied security control that silently degrades is the worst-case outcome of duplication, and it is what happened.

**2. Slack's OAuth scope state now lives in three places with only two reconciled.** Scopes are held in the code registry, in `public_mcp_apps.oauth_scopes`, and in `oauth_providers.default_scopes`. The migration updates the first two and ignores the third, leaving the app-id-less OAuth login path minting tokens with the stale 3-scope set (F2). The team already hit and solved exactly this failure mode when Facebook's scopes grew — by adding it to the app-scoped-grant requirement set — but that precedent was not applied here.

**3. Cursor pagination is repeated four times with inconsistent failure semantics.** The same paginate-until-cursor-exhausted skeleton appears in four functions, but only one of them tolerates a mid-pagination failure by returning already-collected results as truncated. The other three throw away every page fetched so far (F5). Divergent error tolerance across copies of the same loop is a maintenance hazard, not just duplication — see also the simplification note below.

**4. `slack_search_messages` is presented as "search" but is a client-side substring scan with an unbounded request cost.** It can issue up to 5 `conversations.history` pages plus 20 `conversations.replies` calls against Tier-2/Tier-3 rate-limited endpoints, per invocation. The docstring does not warn about this cost, there is no per-call request budget, and the thread-scan counter is incremented only on success (F3), so failures do not consume budget.

## Findings

### MAJOR

**F1 — `XAGENT_SLACK_FILE_ALLOWED_DIRS` is never set, so the upload allowlist silently degrades to the subprocess cwd.** See inline comment on `src/xagent/web/tools/mcp/slack.py:104`.

**F2 — Migration updates `public_mcp_apps.oauth_scopes` but leaves `oauth_providers.default_scopes` stale.** See inline comment on `src/xagent/migrations/versions/20260812_add_slack_history_reactions_files_scopes.py:167`.

**F3 — Failed thread fetches do not consume the `MAX_SEARCH_THREADS` budget.** See inline comment on `src/xagent/web/tools/mcp/slack.py:538`.

### MINOR

**F4 — `slack_search_messages` limit has no upper clamp.** See inline comment on `slack.py:494`.

**F5 — `slack_list_direct_messages` discards collected pages on a mid-pagination failure.** See inline comment on `slack.py:438`.

**F6 — Allowlist entries are not stripped before being parsed as paths.** *(Confirms and re-frames the prior `gemini-code-assist` inline comment on this line — the bug is real, but it fails closed, not open, so it is not a security bypass.)* See inline comment on `slack.py:109`.

**F7 — `slack_upload_file`: no zero-byte guard, plus a TOCTOU window on the allowlist check.** See inline comment on `slack.py:641`.

**F8 — `PermissionError` message leaks host filesystem layout into the model context.** See inline comment on `slack.py:133`.

**F9 — The Slack-ID short-circuit accepts user IDs as channel references.** See inline comment on `slack.py:75`.

**F10 — Channel-name resolution is not memoized across tool calls.** See inline comment on `slack.py:65`.

**F11 — Test coverage gaps in `tests/web/tools/test_slack_mcp.py`.** All of the following were confirmed absent:
- `_allowed_file_dirs`'s default `Path.cwd()` fallback — the branch that actually runs in production per F1 — is never tested; every upload test sets the env var explicitly. This is the single gap most likely to have caught F1.
- No symlink-escape test for the allowlist guard, which is the primary attack it exists to stop.
- No multi-directory, whitespace, or `~`-expansion tests for `_allowed_file_dirs`.
- `slack_upload_file`: `thread_ts`, `initial_comment`, and `length` are never asserted in the outgoing payload; no test for a failing-status multipart upload response.
- `_resolve_channel_id`: only single-page found/not-found are covered; no multi-page cursor case and no `MAX_PAGES`-exhaustion case.
- `slack_search_messages`: the `scan_limit` bound, the `MAX_SEARCH_THREADS` bound, and the match-count truncation path are all untested.
- `slack_list_direct_messages`: no assertion that `types == "im,mpim"` is sent; no mid-pagination-failure test (relevant to F5).
- `slack_remove_reaction` has no error-path test, unlike `slack_add_reaction`.
- Cosmetic: two search tests sit after the upload-file section header.

**F12 — Docstring points at the wrong tool for channel-name resolution.** See inline comment on `slack.py:303`.

**F13 — `example.env` documents a variable that has no effect.** `example.env:552-555` presents `XAGENT_SLACK_FILE_ALLOWED_DIRS` as operator-settable, but per F1 it never reaches the subprocess, so setting it does nothing. The LinkedIn equivalent is deliberately absent from `example.env` precisely because it is launcher-injected rather than operator-set. Either fix F1 and keep the documentation consistent, or follow the LinkedIn pattern and drop the entry.

**F14 — File-allowlist guard duplicated from the LinkedIn connector.** `src/xagent/web/tools/mcp/slack.py:103-135` is a ~25-30 line verbatim copy of the guard in the LinkedIn connector (`linkedin.py:54-83`, untouched by this PR), differing only in env var name and wording. Extracting a shared helper (`tools/mcp/utils.py` is already imported here for `setup_proxy_env`) behind one generic env var would have prevented both F1 and F6.

**F16 — Migration `downgrade()` is not restorative on a fresh database.** `20260812_add_slack_history_reactions_files_scopes.py`'s `PREVIOUS_SCOPES` constant hardcodes the original 3 scopes, but `20260801_seed_slack_mcp_app.py`'s `SLACK_SCOPES` was edited by this PR to seed 12. On a fresh database (seeded at 12, then upgraded to the same 12), running the downgrade resets the row to 3 rather than back to the 12 that were actually seeded. No test constructs the real seed→upgrade→downgrade chain. Real-world impact is low (insert-only migration, no data-loss-critical path), but it is a genuine inconsistency — consider deriving the downgrade target from the seed migration's scope list, or adding a test covering the actual chain.

- **S1** — `slack_add_reaction` (line 568) and `slack_remove_reaction` (line 592) have near-identical ~13-line bodies differing only in the `reactions.add` / `reactions.remove` endpoint; a shared `_set_reaction(action, channel, timestamp, emoji_name)` saves ~10-14 lines with zero behavior loss.
- **S2** — The cursor-pagination skeleton (build params, conditionally add cursor, call `_request`, extract `next_cursor`, break if empty, inside `for _ in range(MAX_PAGES)`) repeats in `_resolve_channel_id` (line 65), `slack_list_direct_messages` (line 417), and `slack_search_messages` (line 471); extract a shared cursor-pagination generator rather than a merged loop body, since per-page processing genuinely differs (early-return search vs. accumulate-with-truncation vs. accumulate-with-extra-limits), saving ~10-15 lines. `slack_list_channels` is intentionally out of scope here because it carries the F5 partial-failure tolerance; folding it in later must explicitly preserve that.

net: -20 to -25 lines possible

---

The only prior review on this PR is from the `gemini-code-assist` bot (state: COMMENTED); its single inline comment is addressed and re-framed in F6 above.

yiboyasss:
Pushed a follow-up commit addressing the review findings. Fixed items (F1–F9, F12, S1, gemini-code-assist's whitespace comment) aren't detailed here since they're straightforward fixes — see the diff. Notes on the items intentionally left as-is:

- **F10** (memoize channel-name resolution across calls): not done. A cross-call cache risks serving a stale id after a channel is renamed/archived, and the existing single-retry-on-429 in `_request` already softens the self-rate-limit concern this was raised for. The correctness downside outweighs the benefit here.

- **F14** (extract the file-allowlist guard into a helper shared with the LinkedIn connector): not done. Doing this properly means touching and re-verifying the already-shipped, separately-reviewed LinkedIn connector — a cross-cutting refactor beyond this PR's scope. Fixed the concrete bugs the duplication caused in Slack's own copy instead (F1, F6, F8).

- **F16** (migration `downgrade()` isn't restorative on a fresh DB seeded straight at 12 scopes): not done. Real-world impact is low as you noted (no downgrade path runs in production), and the same latent seed-vs-scope-migration tension already exists for HubSpot's own pair of migrations elsewhere in this codebase. Fixing it here alone would need a policy decision on how downgrade targets should be derived — out of scope for this pass.

- **S2** (shared cursor-pagination generator across `_resolve_channel_id` / `slack_list_direct_messages` / `slack_search_messages`): not done. The three call sites have genuinely different per-page/error semantics (early-exit search vs. plain accumulate vs. accumulate-with-extra-limits); forcing them through one generator risked subtly changing one of those behaviors. The one correctness-relevant piece of this critique — F5's missing partial-failure tolerance in `slack_list_direct_messages` — was fixed directly instead.

- **F11** (test coverage gaps): partially addressed — added tests for everything actually changed this round (F1, F3, F4, F5, F6, F7, F9, plus the missing `slack_remove_reaction` error-path test). Did not add: a symlink-escape test for the allowlist guard, `_resolve_channel_id` multi-page/`MAX_PAGES`-exhaustion tests, or `slack_upload_file`'s `thread_ts`/`initial_comment`/failing-multipart-response assertions — these are gaps in code that didn't change this round, so covering them wasn't needed to validate the fixes, and I prioritized the fixes themselves given time constraints.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -25,11 +26,20 @@
...
-_SLACK_ID_PATTERN = re.compile(r"^[A-Z][A-Z0-9]{5,}$")
+# Slack conversation ids are uppercase alphanumerics prefixed by their
+# conversation type: "C" (public channel), "G" (private channel or
+# multi-party DM), "D" (1:1 DM). Restricted to those three prefixes — rather
+# than any letter — so a user id ("U...") or bot id ("B...") passed by
+# mistake (e.g. copied from slack_list_direct_messages' "user" field) is
+# rejected up front instead of being misclassified as an already-resolved
+# conversation id. Channel *names* are forced lowercase by Slack, so this
+# can't misclassify a name as an id either way.
+_SLACK_ID_PATTERN = re.compile(r"^[CGD][A-Z0-9]{5,}$")
```
rogercloud, line 44: Regression from the F9 narrowing: this pattern now rejects U-ids, and `_normalize_channel` (slack.py:56-63) prepends `#` to anything that does not match — so `slack_post_message(channel="U0123456789")` sends `#U0123456789` to chat.postMessage, which fails. The base pattern (`^[A-Z][A-Z0-9]{5,}$`) passed U-ids through, and chat.postMessage accepts U-ids for DMs. The tool docstring at slack.py:284 advertises DM-by-user-id support, and `slack_list_direct_messages` returns `user` ids. Suggest: keep the strict `^[CGD]` pattern for `_resolve_channel_id` only (the conversations.*/reactions.*/files.* endpoints need real ids anyway), and let `_normalize_channel` pass through any `^[A-Z][A-Z0-9]{5,}$` id unchanged. Add a test posting to a U-id via `slack_post_message`.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +67,88 @@ def _normalize_channel(channel: str) -> str:
...
+    if not local_path.is_absolute():
+        local_path = Path.cwd() / local_path
+    local_path = local_path.resolve()
+
+    if not local_path.is_file():
+        raise FileNotFoundError(f"File not found: {file_path}")
+
+    allowed_dirs = _allowed_file_dirs()
+    for allowed_dir in allowed_dirs:
+        if local_path == allowed_dir or local_path.is_relative_to(allowed_dir):
```
rogercloud: Two things here. (1) Simplification: `local_path == allowed_dir or` is a dead disjunct — `is_relative_to` already covers equality, and a directory path fails the `is_file()` check at :129 regardless, so the `==` arm can never be the deciding condition. Drop it. (2) Security note (minor): this path-based check completes before the open at :685, so the comment at :680-684 claiming "the file that was allowlist-checked is the exact file read from" overstates — the fstat-on-open check ties size only, not identity. Likelihood is low (needs a concurrent local write inside an allowed dir), but either open first and verify the fd's path, or soften the comment.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +279,453 @@ def slack_list_channels(
...
+                # error) must still consume the budget, or a heavily
+                # threaded/rate-limited channel could retry far past
+                # MAX_SEARCH_THREADS attempts with no caller-side bound on
+                # total wall time.
+                threads_scanned += 1
+                try:
+                    replies_result = _request(
+                        "GET",
+                        "conversations.replies",
+                        params={"channel": channel_id, "ts": thread_ts, "limit": 200},
```
rogercloud, line 602: Minor correctness: conversations.replies is fetched with limit=200 and the response's `has_more`/`next_cursor` are ignored, so threads with more than 200 replies are silently cut off. The `truncated` computation at :600-607 counts only threads never attempted (budget exhaustion), not replies dropped inside a single thread — so a >200-reply thread reports a false "complete". Suggest: paginate replies within the remaining thread budget and set `truncated` when `has_more` remains, or at minimum document the 200-reply-per-thread cap in the tool docstring.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -187,24 +279,453 @@ def slack_list_channels(
...
+            # Per Slack's files.completeUploadExternal migration guide, the
+            # returned upload_url is a pre-authenticated endpoint that takes
+            # the raw file as multipart form data under the "file" field —
+            # no bearer token needed (or accepted) on this leg.
+            upload_response = requests.post(
+                upload_url,
+                files={"file": (resolved_filename, fh)},
+                timeout=DEFAULT_TIMEOUT_SECONDS,
+            )
+        upload_response.raise_for_status()
```
rogercloud: Minor error-handling: `raise_for_status()` on the upload leg embeds the pre-authenticated upload URL in HTTPError's str() (verified with requests 2.34.2), and that string flows back to the caller/LLM via `_error(str(e))` at :725-727. That is exactly the leak this file's own host-path masking at :137-149 (and zoom.py's `_download_text` sanitization) prevents elsewhere. Suggest: catch the HTTPError and return a status-only message (e.g. "upload failed: {status} {reason}"), logging the full URL server-side.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +67,88 @@ def _normalize_channel(channel: str) -> str:
...
+        if (stripped := raw_dir.strip())
+    ]
+
+
+def _resolve_allowed_file_path(file_path: str) -> Path:
+    """Restrict slack_upload_file to files under an allowlisted directory,
+    the same defense used by the LinkedIn connector's image upload — without
+    it an agent could be tricked into exfiltrating arbitrary host files
+    through this tool."""
+    local_path = Path(file_path).expanduser()
```
rogercloud, line 143: Minor ergonomics: a relative `file_path` resolves against the subprocess cwd (line 126), not against the allowlisted directories — so the natural `report.txt` usage fails with FileNotFoundError, or worse, PermissionError when an unrelated file with that name happens to exist in the server's cwd. The docstring at :666-670 never states the absolute-path requirement. Suggest: resolve relative paths against the first allowed directory (so the common task-workspace case just works), or document that `file_path` must be absolute.

src/xagent/migrations/versions/20260812_add_slack_history_reactions_files_scopes.py
```diff
@@ -0,0 +1,217 @@
...
+    cannot affect an unrelated app's authorize request.
+    """
+    if not _columns_present(
+        bind, "oauth_providers", {"provider_name", "default_scopes"}
+    ):
+        return
+
+    bind.execute(
+        sa.update(OAUTH_PROVIDERS_TABLE)
+        .where(OAUTH_PROVIDERS_TABLE.c.provider_name == PROVIDER_NAME)
```
rogercloud, line 164: Minor backward-compat: unlike the description update at :140-163, which is guarded by the if-unchanged pattern, this overwrites `oauth_providers.default_scopes` unconditionally. `default_scopes` IS operator-editable (admin_mcp.py:69, 384-385) with no builtin protection, so a customized value is silently clobbered on upgrade. Suggest: reuse the same if-unchanged pattern (only overwrite when the persisted value still equals the previous canonical scope list), and add a migration test covering a customized `default_scopes` row — currently absent.

example.env
```diff
@@ -529,18 +529,29 @@ META_CONFIG_ID=""
...
 SLACK_CLIENT_ID=""
 SLACK_CLIENT_SECRET=""
 # Redirect URI for Slack OAuth callback — must also be added under the app's
 # "OAuth & Permissions" > "Redirect URLs".
 # Example: http://localhost:8000/api/auth/slack/callback
 SLACK_REDIRECT_URI=""
+# Optional: comma-separated list of directories slack_upload_file may read
+# files from (e.g. the task workspace root). Defaults to the MCP server
+# process's current working directory when unset.
+# XAGENT_SLACK_FILE_ALLOWED_DIRS=""
```
rogercloud: F13 (still open): this comment presents XAGENT_SLACK_FILE_ALLOWED_DIRS as operator-settable, but operator-set values are never read or forwarded: the subprocess env is built only from the task-scoped value (config.py:2918-2921), the WebToolConfig builder (config.py:1313-1330) never reads `os.environ`, and `static_env` is only used for `GOOGLE_ADS_DEVELOPER_TOKEN`. A user following this file gets a variable that silently does nothing. Suggest: either document it as launcher-controlled (move the cwd-default note to where the fallback actually lives), or wire the env read-through.

src/xagent/web/tools/mcp/slack.py
```diff
@@ -57,6 +67,88 @@ def _normalize_channel(channel: str) -> str:
...
+    raise ValueError(
+        f"Could not resolve channel '{channel}' to an id — use slack_list_channels "
+        "or slack_list_direct_messages to look it up, or pass the channel id "
+        "directly."
+    )
+
+
+def _allowed_file_dirs() -> list[Path]:
+    raw_dirs = os.environ.get("XAGENT_SLACK_FILE_ALLOWED_DIRS", "")
+    if not raw_dirs.strip():
```
rogercloud, line 129: Minor security note (severity-adjusted down from major during verification): when XAGENT_SLACK_FILE_ALLOWED_DIRS is unset, `_allowed_file_dirs` falls back to `Path.cwd().resolve()` (line 111) — the MCP subprocess's cwd, which is never set explicitly and therefore inherits the server's cwd. config.py:2918-2922 sets the var only when the builder computed a non-empty value, i.e. only with a task context. Verified latent: no current production path builds WebToolConfig without a truthy `task_id`, and the LinkedIn connector ships identical semantics at base — but the PR's claim that uploads "can't be used to exfiltrate arbitrary host files" is overstated for this branch. Suggest: fail closed (empty allowlist → reject all uploads) or set the subprocess cwd explicitly. tests/web/tools/test_oauth_mcp_file_allowed_dirs_env.py:40-54 currently pins the omission.

rogercloud (approved):
## Follow-up review: Slack connector expansion (update commit 3c49947f)

### PR summary

This PR expands the built-in Slack connector from two tools (list channels, post message) to eleven: channel/private-channel/DM history reads, thread replies, channel info, DM listing, message search, reactions add/remove, file upload, and threaded posting. It also syncs nine new bot scopes across the MCP registry, the seeded app row, the `oauth_providers` row, and a new migration, force-disconnects existing Slack grants so users re-consent with the new scopes, and adds an environment allowlist for `slack_upload_file` mirroring the LinkedIn connector's guard. Blocking: no — recommended event: APPROVE.

### Update summary (since last review)

One new commit (3c49947f) addresses the prior review (previously CHANGES_REQUESTED at b0ed0a3d): `XAGENT_SLACK_FILE_ALLOWED_DIRS` is now populated in the MCP subprocess env, `oauth_providers.default_scopes` is synced in the migration (upgrade and downgrade), the thread-fetch budget counts attempts and failed threads set `truncated`, the search limit is clamped, `slack_list_direct_messages` tolerates mid-pagination failures, allowlist entries are whitespace-stripped, zero-byte uploads are rejected with fstat-on-open, host paths are masked in upload rejections, the ID pattern was narrowed, the docstring was fixed, and the reaction tools were deduped into `_set_reaction`.

### Prior-findings checklist

- F1 (major, env never populated) — FIXED — task-scoped path carries the var (config.py:2918-2921).
- F2 (major, default_scopes stale) — FIXED — `_set_slack_provider_default_scopes` in upgrade+downgrade, tested.
- F3 (major, thread budget only-on-success) — FIXED — `threads_scanned += 1` before the request (slack.py:564); failed thread → `truncated`; test at test_slack_mcp.py:1136.
- F4 (minor, search limit clamp) — FIXED — `max_matches = max(1, min(int(limit), 1000))` (slack.py:517); test :1175.
- F5 (minor, DMs mid-pagination discard) — FIXED — partial + truncated + error (slack.py:452-466); tests :1206, :1231.
- F6 (minor, whitespace stripping) — FIXED — walrus strip of allowlist entries (slack.py:112-115); test :1103.
- F7 (minor, zero-byte + TOCTOU) — PARTIAL — zero-byte rejection fixed (slack.py:685-688); check-vs-open window remains → carried as N5.
- F8 (minor, host-path leak) — FIXED — masked message (slack.py:137-149), detail logged server-side; test :918.
- F9 (minor, ID pattern accepts U-ids) — FIXED — `^[CGD][A-Z0-9]{5,}$` (slack.py:42); tests :1079, :1083 — but over-corrected: postMessage DM-by-user-id regression → N2.
- F10 (minor, memoization) — WAIVED — stale-id + retry rationale defensible.
- F11 (minor, test gaps) — PARTIAL — 6/15 sub-items covered incl. all fix-relevant ones; deferred symlink-escape/multi-page/upload-payload tests verified technically correct (unchanged code) and dropped per comment-history rule.
- F12 (minor, docstring tool) — FIXED — slack.py:315-320.
- F13 (minor, example.env no-effect var) — NOT FIXED — example.env:549-554 still presents the var as operator-settable; operator-set values are never read/forwarded (launcher injection config.py:2918-2921; builder config.py:1313-1330 never reads `os.environ`; `static_env` only used for `GOOGLE_ADS_DEVELOPER_TOKEN`).
- F14 (minor, LinkedIn guard duplication) — WAIVED — sound scope call; divergence now materialized one-directionally (LinkedIn copy keeps F6/F8-class defects); slack.py:121 docstring claiming "the same defense used by the LinkedIn connector" is now stale — follow-up recommended.
- F16 (minor, downgrade non-restorative) — WAIVED — identical latent tension ships in HubSpot's own migration pair at base.
- S1 (simplification, reaction dedupe) — FIXED — `_set_reaction` helper (slack.py:614-629); remove_reaction error test :1249.
- S2 (simplification, pagination generator) — WAIVED — the three loops' error/exit/truncation semantics genuinely differ.
- Prior inline 3767524886 (_fetch_messages shared skeleton) — NOT FIXED — still applicable, low priority (see Simplification section).

### New findings

**Minor**

- **N1 (security; severity-adjusted down from major)** — slack.py:110-111: with no env value, `_allowed_file_dirs` falls back to `Path.cwd().resolve()`. config.py:2918-2922 sets the var only when the builder (config.py:1313-1330) produced a non-empty value (i.e. only with a task context), and no cwd is set on the subprocess, which therefore inherits the server's cwd. Verified latent — no current production path builds WebToolConfig without a truthy `task_id`, and the LinkedIn connector ships identical semantics at base — but the PR's claim that uploads "can't be used to exfiltrate arbitrary host files" is overstated for this branch. Suggest: fail closed (empty allowlist → reject) or set the subprocess cwd explicitly. tests/web/tools/test_oauth_mcp_file_allowed_dirs_env.py:40-54 pins the omission.
- **N2 (regression from F9)** — slack.py:42: the narrowed pattern rejects U-ids, so `_normalize_channel` (slack.py:56-63) prepends `#` to them and `slack_post_message(channel=U0123456789)` sends `#U0123456789` to chat.postMessage, which fails. The base pattern (`^[A-Z][A-Z0-9]{5,}$`) passed U-ids through and chat.postMessage accepts them for DMs; the tool docstring at slack.py:284 advertises DM-by-user-id support and `slack_list_direct_messages` returns `user` ids. Suggest: keep the strict pattern for `_resolve_channel_id` only; let `_normalize_channel` pass ids through loosely.
- **N3 (correctness)** — slack.py:541-546 / 583-593: `slack_search_messages` duplicates thread_broadcast replies — they appear in both conversations.history and the parent's conversations.replies; the history scan appends them without `thread_ts` and the thread scan re-appends them with `thread_ts`. Suggest: dedupe by ts or skip broadcast replies in one of the two scans.
- **N4 (backward-compat)** — migration 20260812 :136-137: overwrites `oauth_providers.default_scopes` unconditionally while the description is guarded via if-unchanged (:140-163); `default_scopes` IS admin-editable (admin_mcp.py:69, 384-385) with no builtin guard. Suggest: reuse the if-unchanged pattern; add a customized-default_scopes test (confirmed absent).
- **N5 (security)** — slack.py:134 / :685: the path-based allowlist check completes before the open; the comment at :680-684 ("the file that was allowlist-checked is the exact file read from") overstates — fstat-on-fd ties size only. Low likelihood (needs a concurrent local write in an allowed dir). Fix (open first, verify fd path) or correct the comment.
- **N6 (correctness)** — slack.py:568-570: conversations.replies is fetched with limit=200 and `has_more`/`next_cursor` are ignored → threads with more than 200 replies are silently truncated; the `truncated` computation at :600-607 counts only unfetched threads → false "complete". Suggest: paginate within the budget or flag `truncated` on `has_more`.
- **N7 (error-handling)** — slack.py:704-709: upload-leg `raise_for_status` embeds the pre-authenticated upload URL in HTTPError's str() (verified with requests 2.34.2), returned via `_error(str(e))` at :725-727 — inconsistent with this file's own host-path masking (:137-149) and zoom.py's `_download_text` sanitization. Suggest: return a status-only message; log the URL server-side.
- **N8 (API shape)** — slack.py:321/334 vs :523-524: history limit is a page size; search limit is max matches with a separate `scan_limit` (messages scanned); the search docstring never documents `limit`. Suggest: document both parameters' semantics (or align them).
- **N9 (API shape)** — slack.py:600-608: search returns `truncated` but no resume cursor, unlike history/replies (:356, :397); the scan is hard-capped at 1000. Suggest: add `next_cursor` or document the cap.
- **N10 (API shape)** — slack.py:465-467: `slack_list_direct_messages` mixes an `error` key into a success payload on mid-pagination failure; mirrors the pre-existing `slack_list_channels` house style (:243), but search's cleaner truncated-only alternative exists in the same PR. Suggest: switch to `truncated`/partial flags.
- **N11 (cohesion)** — the 12-scope list appears in 4 places (builtin_mcp_registry.py:131-144, 535-548; seed migration 20260801:60-73; new migration 20260812:47-58). Migration copies are mandatory frozen snapshots (HubSpot precedent) and the registry pair follows the file-wide inline-literal convention, all anchored by 3 drift tests. Weakest-class minor; note only.
- **N12 (ergonomics)** — slack.py:124-127: a relative `file_path` resolves against the subprocess cwd, not the allowed dir — a natural `report.txt` fails (FileNotFoundError, or PermissionError with a decoy in cwd); the docstring (:666-670) never states the absolute-path requirement. Suggest: resolve relative paths against the first allowed dir, or document the requirement.
- **N13** — slack.py:81-100: `_resolve_channel_id` is silently bounded by MAX_PAGES=20×200=4000 channels and raises a generic `Could not resolve channel` with no truncation signal beyond page 20. Suggest: surface the bound in the error.
- **N14 (docs)** — slack.py:369-370: `slack_get_thread_replies` docstring says "Fetch all replies" but defaults to limit=100 (cap 1000, one call per page); its own cursor paragraph (:373-375) contradicts the first line. Suggest: reword.
- **N15** — slack.py:686-708: no upload size bound — the allowlist bounds what, not how much; inconsistent with the file's clamp-everything convention. Suggest: add a size cap (fstat is already taken at open).
- **N16 (tests)** — 5 confirmed gaps where no test would fail if the code broke: (1) broadcast-reply dedup unobservable (test_slack_mcp.py:790-827 — parent text doesn't match the query); (2) >200-reply thread pagination; (3) migration preserving customized default_scopes; (4) get_channel_history limit clamp (test :551-567 only pass-through); (5) relative-path resolution branch. Three further sub-items dropped per the author's verified deferral (upload HTTP error, thread_ts/initial_comment passthrough, multi-page resolution — code unchanged this round).

**Design-level**

- **N17 (design)** — config.py:2918-2921 hardcodes both allowlist vars for every OAuth MCP subprocess, extending the pre-existing LinkedIn pattern; a per-app registry `launch_config` would need a small shape extension. Acceptable as-is; note for the future.
- **N18 (design note)** — two channel-resolution semantics coexist: `_normalize_channel` (slack.py:56-63) passthrough for postMessage vs `_resolve_channel_id` (slack.py:65-101) lookup; the divergence is documented (:71-76) and all current callers use the right helper.

**Info**

- **N19** — the blanket force-disconnect of all Slack grants (migration :166-199, unconditional) is deploy-ordering dependent; the ordering is documented only implicitly at example.env:542-544. Suggest the warning log additionally instruct operators to add the app scopes BEFORE upgrading. HubSpot-precedent pattern; failure mode is self-recoverable (reconnect).
- **N20** — search is sized for bounded per-channel scans (~40 worst-case Tier-2 calls per invocation); bounded and honestly truncated. Document-level context only.

- L134: shrink — `local_path == allowed_dir or` is a dead disjunct (is_relative_to covers equality; directories fail is_file first). Drop the `==` arm.
- L184-187: shrink — `_invalidate_existing_slack_grants` re-inspects user_oauth columns inline, duplicating `_columns_present` (defined same file :71-85). Replace with `_columns_present(bind, "user_oauth", {"refresh_token"})`.
- Carried: prior inline 3767524886 (_fetch_messages shared skeleton for get_channel_history/get_thread_replies) — NOT FIXED, still applicable, low priority.

net: -2 lines possible (plus an optional ~20-line _fetch_messages helper if 3767524886 is taken).

No confirmed critical or major findings remain: all three majors (F1-F3) are FIXED and every new finding this pass is minor or informational. The minors are non-blocking follow-ups; N1 (allowlist cwd fallback) and N5 (TOCTOU) are worth addressing before any hardening claim is made, but they do not block merge. Recommended event: APPROVE.

PR #1305, opened 2026-08-12
ref(steps): extract incremental public step projector

AlexLiu190625:
## What

Extracts the trace-to-public-step folding logic out of `map_trace_events_to_public_steps` into an incremental `PublicStepProjector` (`feed` / `materialized_steps` / `from_history`). The batch function is now a thin driver: build one throwaway projector over the event history, read the materialized steps, and apply the final `started_at` ordering pass.

Behavior is byte-identical: every pre-existing test assertion passes unchanged.

## Why

The folding logic and the public step allow-list previously lived only inside the batch function, so nothing else could fold events one at a time without duplicating them. The projector is now the single owner of all folding state (pairing tables, plan counter); the batch endpoint keeps none.

## Verification

- All pre-existing assertions unchanged and green.
- Equivalence pins across 18 parameterized event sequences: batch output equals `from_history(...).materialized_steps()` plus the batch-only ordering pass; incremental `feed` results rebuild the same materialized state.
- `feed` return contract pinned with deep-copied snapshots: `running` at start-event time, `completed` (with `completed_at`) at end-event time; orphan ends, unlisted event types, and empty event types yield no steps.

### DIFF
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -33,6 +33,18 @@ Pure-function design:
     place SDK clients can observe a behavior change is through this
     one function's output -- which makes regressions easy to gate.
 
+Incremental projection:
+
+    The state machine that implements the pairing rules below lives in
+    :class:`PublicStepProjector`, not in this function. The projector
+    holds all the folding state a fold needs -- the pending
+    (start-seen, end-not-yet-seen) table, the ``dag_plan_*`` replan
+    counter -- so a caller can feed it one event at a time and read
+    back the steps that changed. ``map_trace_events_to_public_steps``
+    stays pure by constructing one fresh, throwaway projector per
+    call and reading back its materialized result; it keeps no
+    folding state of its own.
+
 Pairing rule:
 
     Start / end events are paired by a stable ``key``:
@@ -83,12 +95,14 @@ logger = logging.getLogger(__name__)
 class PublicStepProjector:
     """Incremental folding state machine: trace events -> public steps.
 
-    This holds exactly the folding state ``map_trace_events_to_public_steps``
-    used to keep as local variables -- the pending (start-seen,
-    end-not-yet-seen) pairing table and the ``dag_plan_*`` counter/open-key
-    used to disambiguate replan. Pulling it into an instance is what lets a
-    caller feed a live event stream one event at a time (see the module
-    docstring's pairing rules for what "pairing" means per event family).
+    Holds all the folding state a fold needs -- the pending
+    (start-seen, end-not-yet-seen) pairing table and the ``dag_plan_*``
+    counter/open-key used to disambiguate replan. The batch driver
+    (``map_trace_events_to_public_steps``) keeps none of its own: it
+    builds one instance and reads back the result. Holding this state
+    on an instance is what lets a caller feed a live event stream one
+    event at a time (see the module docstring's pairing rules for what
+    "pairing" means per event family).
 
     Two ways to build one:
 
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -433,8 +433,8 @@ def _thinking_pair_key(event: Any, phase: str) -> str:
 
     ``react_action_*`` and ``dag_step_*`` always carry a step_id
     which is the natural pairing key. Planning events are handled
-    inline in :func:`map_trace_events_to_public_steps` because they
-    lack a per-plan identifier and need a synthesized counter.
+    inline in :meth:`PublicStepProjector.feed` because they lack a
+    per-plan identifier and need a synthesized counter.
     """
     return str(_safe_get(event, "step_id") or _safe_get(event, "event_id") or "")
 
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -49,10 +49,11 @@ Pairing rule:
 
     Start / end events are paired by a stable ``key``:
 
-      - ``tool_execution_*`` events pair on ``data['tool_execution_id']``
-        when present, falling back to ``step_id``. The tool execution
-        id is generated per-invocation and is the only safe key when
-        the same tool is called twice in the same step.
+      - ``tool_execution_*`` events pair on ``data['tool_call_id']``
+        (the provider-assigned call id), falling back to ``step_id``.
+        ``tool_execution_id`` is accepted ahead of it for compatibility
+        but has no current producer. A per-invocation id is the only
+        safe key when the same tool is called twice in the same step.
       - ``react_action_*`` events pair on ``step_id``.
       - ``dag_step_*`` events pair on ``step_id``.
       - ``dag_plan_*`` events pair on ``task_id`` (single planning
@@ -285,10 +286,11 @@ class PublicStepProjector:
             is_delegation = is_agent_tool_name(tool_name)
             public_type = "agent_delegation" if is_delegation else "tool_call"
             # Pair on a per-invocation id (unique even when one step
-            # invokes the same tool twice). v1 emits
-            # ``tool_execution_id``; v2 emits ``tool_call_id``. Either
-            # is fine — the fallback chain accepts both. step_id alone
-            # is unsafe because one step may invoke multiple tools.
+            # invokes the same tool twice). The effective key today is
+            # ``tool_call_id``; ``tool_execution_id`` is accepted ahead
+            # of it for compatibility but has no current producer.
+            # step_id alone is unsafe because one step may invoke
+            # multiple tools.
             key = (
                 _data_get(event, "tool_execution_id")
                 or _data_get(event, "tool_call_id")
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -68,6 +68,16 @@ Pairing rule:
     Orphan starts (start with no matching end) are emitted with
     ``status='running'`` and ``completed_at=None``. This naturally
     handles the case where the SDK polls ``/steps`` mid-task.
+
+    Key collision -- two starts sharing one pairing key before either
+    ends -- is last-write-wins: the second start replaces the first
+    pending entry, so only the second reaches a terminal state. Both
+    carry the same public step id (id and pairing key are derived from
+    the same (type, key) pair), so a consumer folding ``feed`` results
+    by id converges on the second step rather than being left with a
+    stranded ``running`` one. ``dag_plan_*`` is the one exception: it
+    has no natural key at all, so a counter gives every plan a
+    distinct one.
 """
 
 from __future__ import annotations
@@ -119,6 +129,13 @@ class PublicStepProjector:
     resort is a batch-only concern (see ``map_trace_events_to_public_steps``):
     a live consumer wants steps in the order they actually resolved, not
     resorted on every event.
+
+    Preconditions for the incremental path: one instance per task, fed
+    in event order, from one thread/task at a time. ``feed`` mutates
+    the pairing tables without any locking, and the ``dag_plan_*`` open
+    key is a single slot, so two interleaved tasks would cross-pair.
+    The batch driver satisfies all three trivially -- it builds a fresh
+    throwaway instance per call.
     """
 
     def __init__(self) -> None:
@@ -164,6 +181,12 @@ class PublicStepProjector:
         Callers that need the public started_at-sorted order (currently
         only the ``map_trace_events_to_public_steps`` batch driver) sort
         this themselves.
+
+        The list itself is a fresh snapshot, but the step dicts inside
+        it are the projector's live objects: a still-``running`` step
+        is mutated in place when its end event arrives (see
+        :meth:`feed`). Treat them as read-only views -- a caller that
+        needs a stable picture must copy before storing.
         """
         steps = list(self._finished)
         steps.extend(self._pending.values())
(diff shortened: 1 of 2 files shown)


src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -80,80 +92,112 @@
...
+    def materialized_steps(self) -> List[Dict[str, Any]]:
+        """Return every step currently known, finished or still running.
+
+        Order: finished steps in the order they resolved, then any
+        still-``pending`` (``running``) steps in the order they started.
+        Callers that need the public started_at-sorted order (currently
+        only the ``map_trace_events_to_public_steps`` batch driver) sort
+        this themselves.
+        """
+        steps = list(self._finished)
```
rogercloud, line 192: `materialized_steps` returns a snapshot *list* but live dict *objects* — pending steps are mutated in place when finalized, and `feed`'s docstring (above) documents this aliasing while this method's does not. Please state the read-only contract here (returned steps are live views; only the list is a snapshot) so future consumers don't cache references expecting immutable snapshots.

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -225,16 +273,18 @@ def map_trace_events_to_public_steps(
...
             if event_type == "tool_execution_start":
-                pending[(public_type, str(key))] = _build_tool_start(
+                step = _build_tool_start(
                     event,
                     public_type=public_type,
                     tool_name=tool_name,
                     key=str(key),
                 )
-            elif event_type == "tool_execution_end":
+                self._pending[(public_type, str(key))] = step
```
rogercloud, line 310: A second `tool_execution_start` with the same pairing key silently overwrites the pending entry here, while the first step — already returned as `running` by `feed` — never reaches a terminal state and never appears in `materialized_steps` (a ghost step contradicting the `feed` contract). The planning branch guards this with a counter; mirror that guard here and in the skill_select/thinking branches, or document the collision policy, and add a key-collision test.

tests/web/api/v1/test_steps_mapping.py
```diff
@@ -671,3 +677,385 @@ def test_float_timestamp_is_coerced_to_datetime():
...
+    sorted by the same ``started_at`` key the batch driver's trailing
+    sort applies, must equal ``map_trace_events_to_public_steps``
+    exactly. This is the byte-for-byte pin that the batch function
+    stays a pure driver over the projector, never a second copy of
+    the folding logic.
+    """
+    expected = map_trace_events_to_public_steps(events)
+    projected = PublicStepProjector.from_history(events).materialized_steps()
+    projected.sort(key=lambda s: s["started_at"])
+    assert projected == expected
```
rogercloud, line 1007: `expected` comes from `map_trace_events_to_public_steps`, which now delegates to the same projector path under test (from_history + materialized_steps + started_at sort), so this assertion can only catch a second folding implementation, not behavioral drift. The real coverage lives in the ~20 pre-existing hardcoded assertions; consider pinning one or two cases to literal expected output, or keep this deliberately as a redundancy guard.

tests/web/api/v1/test_steps_mapping.py
```diff
@@ -671,3 +677,385 @@ def test_float_timestamp_is_coerced_to_datetime():
...
+    """
+    projector = PublicStepProjector()
+    by_id: Dict[str, Dict[str, Any]] = {}
+    for event in events:
+        for step in projector.feed(event):
+            by_id[step["id"]] = step
+
+    reconstructed = sorted(by_id.values(), key=lambda s: s["id"])
+    materialized = sorted(projector.materialized_steps(), key=lambda s: s["id"])
+    assert reconstructed == materialized
```
rogercloud, line 1030: `reconstructed` and `materialized` hold the same dict objects — `feed` returns exactly the objects `materialized_steps` later returns — so this comparison only catches collection-level drift (missing/duplicated entries), never wrong shared content. Fine as a cheap invariant, but it does not validate the incremental view's semantics.

tests/web/api/v1/test_steps_mapping.py
```diff
@@ -671,3 +677,385 @@ def test_float_timestamp_is_coerced_to_datetime():
...
+
+
+def test_feed_return_value_reflects_step_state_at_call_time():
```
rogercloud, line 1033: The `feed` return-value shape (running → completed in-place mutation, orphan/unexposed/empty handling) is directly asserted only for the react_action thinking family. Tool, skill_select, dag_plan, and message families are covered only through the indirect chain feed ≡ materialized ≡ batch ≡ hardcoded assertions, which is complete today. Optional defense in depth: one direct shape assertion per remaining family.

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -80,80 +92,112 @@
...
-            end event's ``result``)
-          - ``message``: ``{"role": "user"|"assistant", "content": str}``
-
-    Notes:
-        - This function is pure (no I/O, no global state). All test
-          cases live in ``tests/web/api/v1/test_steps_mapping.py``.
-        - Unknown event types (anything not listed in the mapping
-          tables in this module) are silently dropped. Adding a new
-          public type is a deliberate per-type opt-in.
+class PublicStepProjector:
```
rogercloud, line 106: The incremental API documents no thread-safety or serialization contract. Concurrent `feed` calls race on `_pending`/`_finished`, and the single `dag_plan` slot is per-instance — multi-task streaming would need one projector per task. Unreachable in production today (the batch path builds one instance per call), but a one-line docstring note ("not thread-safe; one task per instance, feed in order") makes the precondition explicit for the intended future consumer.

rogercloud (approved):
## PR summary

This PR extracts the trace-event → public-step folding state machine out of the batch function `map_trace_events_to_public_steps` into an incremental `PublicStepProjector` (`feed` / `materialized_steps` / `from_history`), turning the batch function into a thin driver that replays history through a fresh instance, materializes the result, and applies the batch-only `started_at` sort. The extraction is behavior-preserving — public output is byte-identical and the existing test suite needed zero assertion changes — and the diff is confined to `src/xagent/web/api/v1/_step_mapping.py` plus its test file. No production code consumes the new incremental API yet; the two new tests pin the projector against the batch path.

The extraction itself is a genuine decomposition, not speculative layering: the projector is the single owner of all folding state, the batch driver keeps none, and moving the global `started_at` resort out of `materialized_steps` into the batch driver is justified for a live consumer. The reservations are about contract surface, not the refactor itself.

Design-level findings:

- **DC1 — speculative generalization (design note).** `feed`'s return-value contract and `from_history`'s late-attach semantics currently have no production consumers: the only production caller is the batch driver, which uses `from_history` + `materialized_steps` and ignores `feed`'s return values. Until a downstream consumer lands, avoid pinning these contracts too hard — see Simplification opportunities below.
- **DC3 — single-slot precondition undocumented (low).** The projector assumes exactly one plan in flight per instance (noted only in the `__init__` comment) and, for the incremental path, implicitly one task fed in event order. The batch path is protected by the single-task filter in `tasks.py` and the logic predates this PR (7ed3004a), so this is unreachable today — but the new public API should restate the precondition in the class docstring rather than only in an implementation comment.
- **DC2 — aliasing hazard:** merged into finding 1 below.

## Prior reviews

Two earlier gemini-code-assist reviews (Aug 12, 15:44Z / 15:46Z) contained no findings and no inline comments and both postdate the last commit, so this is a fresh review of the same head.

## Line-level findings

**minor**

1. `src/xagent/web/api/v1/_step_mapping.py:168` — `materialized_steps` returns a snapshot *list* but live dict *objects*: pending steps are mutated in place when finalized. `feed`'s docstring (L182-185) documents this aliasing; this method's docstring does not state the corresponding read-only/no-mutation contract. Suggest documenting that returned steps are live views and only the list is a snapshot.
2. `src/xagent/web/api/v1/_step_mapping.py:285` — a second `tool_execution_start` with the same pairing key silently overwrites the pending entry. The first step was already returned as `running` by `feed` but never reaches a terminal state and never appears in `materialized_steps` — a ghost step that contradicts the `feed` contract. The planning branch guards against this with a counter; the tool/skill_select/thinking branches do not, and no test covers a true key collision. Suggest mirroring the planning guard or documenting the collision policy, plus a collision test.
3. `tests/web/api/v1/test_steps_mapping.py:983` — `test_projector_equivalent_to_batch` compares the projector against the batch function that now delegates to the same code path, so it can only catch a second folding implementation, not drift. Residual value is carried by the ~20 pre-existing hardcoded assertions; either pin a couple of cases to literal expected output or accept this deliberately as a redundancy guard.
4. `src/xagent/web/api/v1/_step_mapping.py:95` — the incremental API documents no thread-safety or serialization contract. Concurrent `feed` calls race on `_pending`/`_finished`, and the single `dag_plan` slot is per-instance (multi-task streaming would need one instance per task). Unreachable in production today (the batch path builds one instance per call), so a one-line docstring note suffices.

**low**

5. `tests/web/api/v1/test_steps_mapping.py:1006` — the incremental test compares the same dict objects with themselves (`feed` returns the objects `materialized_steps` later holds), so it only detects collection-level drift, never wrong shared content. Cheap defensive pin; not a correctness test.
6. `tests/web/api/v1/test_steps_mapping.py:1009` — the `feed` return-value shape is directly asserted only for the react_action thinking family; tool/skill_select/dag_plan/messages are covered via the indirect chain (feed ≡ materialized ≡ batch ≡ hardcoded assertions), which is complete today. Optional: one direct shape assertion per remaining family.

- L172: yagni — the `feed` return-value contract (11 return pipelines + `_finalize_pending` now returning `Optional`) has no production consumer; only the 2 new tests read it. Reverting to a mutate-only/continue-style fold removes the contract surface.
- L145: delete — `from_history` has exactly one caller (L414) and its docstring's late-attaching scenario has no code implementing it; inline the replay loop into the batch driver.
- net: ~-80 lines possible if both are applied.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Documented in 1e3bfec16: the `materialized_steps` docstring now states that the list is a fresh snapshot while the step dicts are the projector's live objects, to be treated as read-only views (copy before storing).

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Documented in 1e3bfec16: the class docstring now states the incremental-path preconditions explicitly — one instance per task, fed in event order, single thread/task at a time; the batch driver satisfies all three by building a throwaway instance per call.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Split outcome. (1) The collision policy is now documented in the module docstring and pinned by a new `tool_start_key_collision` fixture flowing through both equivalence tests (1e3bfec16). (2) Declining the counter guard here: the counter would become part of the public step id (as it does for `dag_plan`), changing every tool step id and the collision-case output — this refactor is pinned byte-identical. (3) On the ghost step: the two colliding starts share one public id (id and pairing key derive from the same `(type, key)` pair), so a consumer folding `feed` results by id converges on the second, terminal step — no stranded `running` entry. The real defect is upstream: providers may return empty tool-call ids, several calls then share the `step_id` fallback key, and the projection has no information to pair the matching end events (their keys collide identically). That needs a non-empty unique id at the emission point and is worth tracking separately rather than patching in the projection.

tests/web/api/v1/test_steps_mapping.py
AlexLiu190625: Agreed on the characterization; keeping it as the deliberate redundancy guard (your second option), and its docstring now says so (1e3bfec16). Your first option is already satisfied directly: the 23 pre-existing tests in this file assert literal expected output against `map_trace_events_to_public_steps`, which is now nothing but `from_history(...).materialized_steps()` plus one sort — so those literals pin the projector itself, not a second implementation. What this test adds is the one thing they cannot catch: the batch driver growing a second copy of the folding logic.

tests/web/api/v1/test_steps_mapping.py
AlexLiu190625: Correct about this assertion in isolation — both sides hold the same objects, so it only catches collection-level drift, which is its intended scope. The content semantics are pinned separately: `test_feed_return_value_reflects_step_state_at_call_time` deep-copies each `feed` return before the next event and asserts `running`/`completed_at is None` at start time, `completed` with `completed_at` set at end time, and that the start-time copy still reads `running` after the live object was mutated. Between the two, collection drift and per-step content are both covered.

tests/web/api/v1/test_steps_mapping.py
AlexLiu190625: Fact-checked and accurate; declining the extra per-family assertions with reasoning rather than just "optional". The per-family difference lives entirely in what the end branch merges into `data` (`result` / `output` / `error`), and each of those is already pinned against literal output by a dedicated pre-existing test. The start shape has no per-family divergence to catch — every family routes through `_build_tool_start` or `_build_thinking_start`, both hardcoding `"status": "running"` and `"completed_at": None`. All 18+1 equivalence fixtures already drive every family through `feed` event by event, so one direct assertion per family would restate existing coverage at ~60-80 test lines.

AlexLiu190625:
Inline findings addressed at 1e3bfec16 (docstrings for aliasing, preconditions, collision policy, plus a collision fixture); rationale for the declined items is on each thread.

On the simplification notes (feed return contract, `from_history`): the no-consumer premise is about to stop holding — the next change stacked on this one adds a translation branch inside `feed` that returns the projected step from every exit, and a streaming consumer attaching mid-task needs `from_history` to pre-warm the pairing state (otherwise every end event after attach pairs against an empty table and drops as an orphan) plus `materialized_steps` for its initial snapshot. Reverting the contract now would mean re-adding it in the immediately following change, so keeping both seams; the aliasing note from your finding 1 is now documented as the guardrail that consumer actually needs.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Filed as #1324.

Issue #1288, opened 2026-08-12
ref(widget): make the PublicChatAccessContext exactly-one-entity invariant structural


codeacme17:
Spun out of review feedback on #1275 (the fix for #1225).

## Problem

`PublicChatAccessContext` (`src/xagent/web/api/public_chat_access.py:82`) is the widget twin of `ShareChatAccessContext` and carries the identical exactly-one-entity invariant, narrated but unenforced:

> Set instead of ``widget_agent_id`` when the widget key exposes a workforce. Exactly one of the two is populated for a widget guest.

#1225 made the share twin structural (`__post_init__` rejects both-set and neither-set, merged via #1275); the widget twin is now the weaker of two identical claims. Both fields default to `None`, and the dataclass accepts both-set and neither-set instances. The invariant holds today only because `get_public_chat_user`'s two branches happen to construct correctly.

The misattribution argument is verbatim identical to #1225: `widget_entity_key` feeds the same `entity_rate_limit_key` with the same workforce-wins preference, consumed by the same run-quota gate (`api/chat.py`). A future refactor that let both ids be set would silently charge an agent-widget run to a workforce bucket, with nothing failing loudly.

## Ask

Mirror #1225: add `__post_init__` validation to the frozen `PublicChatAccessContext` rejecting both-set and neither-set `widget_agent_id` / `widget_workforce_id`, comparing against `None` (not truthiness), raising `ValueError` so a violation propagates as a defect rather than being folded into a 401 (see the #1214 boundary policy and the share twin's comment).

## Landability (verified during #1275 review)

Both production construction sites in `get_public_chat_user` set exactly one id. All six direct test constructions set exactly one field, so the guard should land without churn — verify by execution rather than assumption:

- `tests/web/api/test_share_rate_limit_endpoints.py:576` and `:632`
- `tests/web/api/test_upload_connection_boundary.py:668`
- `tests/web/api/test_share_guest_isolation.py:723` (both parametrize cases, `:689` / `:698`)
- `tests/web/api/test_public_runtime_extensions.py:21`

## Nuance to decide

Unlike the share twin, this context carries an `auth_mode` field (default `"widget"`). `get_public_chat_user` rejects any non-widget `auth_mode` today, so an unconditional exactly-one guard matches current reality — but whoever picks this up should decide deliberately whether the guard applies to all `auth_mode` values or only `"widget"`, in case a future auth mode legitimately carries no entity. The docstring comment qualifies the invariant "for a widget guest".

[...]

PR #1308, opened 2026-08-12
feat(api): expose the pending interaction and add the task reply endpoint

AlexLiu190625:
## Summary

`GET /v1/chat/tasks/{task_id}` could tell an SDK caller a task was
`waiting_for_user` but not what it was waiting for, and
`POST /v1/chat/tasks/{task_id}/messages` (append) silently accepted a
waiting task's status filter miss and fell into the generic `task_busy`
409 -- whose default message tells the client to retry, which is
misleading for a task that is not busy, just waiting on an answer.

This PR adds:

- `TaskInfoResponse.pending_interaction` (`PendingInteraction | None`):
  always present in the response body, null unless
  `status == 'waiting_for_user'` **and** the task has at least one
  persisted question message. Carries the question text and an opaque
  list of structured-control descriptors (or `null`) -- passed through
  as-is rather than typed against the agent tool's seven-value `type`
  enum, so this schema doesn't become a second source of truth for it.
  `[]` and `null` are intentionally not normalized together; both mean
  "no structured control, answer with plain text."
- `POST /v1/chat/tasks/{task_id}/reply`: the only way to answer a
  waiting task's question and resume the paused execution. Structurally
  mirrors the A2A input-required resume
  (`web/api/a2a.py::_resume_input_required_a2a_task`) -- same two-phase
  prelease (claim, then either commit past it or restore it exactly),
  same heartbeat guarding the injection, same fail-closed handling of
  every checkpoint outcome -- scoped instead to `Task.source == "sdk"`
  and the v1 SDK's agent/workforce ownership rules, accepting only
  `WAITING_FOR_USER` (not `PAUSED`, which append already handles).
- Four new stable error codes: `interaction_response_required` (409,
  what append now returns for a waiting task), `no_pending_interaction`
  (409, reply called on a non-waiting task), `interaction_not_resumable`
  (409, the task's saved progress can't be resumed -- not retryable),
  `temporarily_unavailable` (503, a transient read failure -- retryable).
- Corrected the `_APPENDABLE_STATUSES` comment (it described
  `WAITING_FOR_USER` as append-resumable, which the tuple never
  actually allowed) and the v1 `status` field description (it listed
  five statuses and silently dropped `waiting_for_user`).
- The v1 read path that builds `pending_interaction` now filters the
  stored `interactions` list down to dict elements before handing it
  to the schema. A row written by an older or buggy writer can carry
  a non-dict element in that list, which previously failed Pydantic
  validation and 500'd `GET` on that task permanently, until the row
  was fixed by hand. Scoped to this one read path's output --
  `get_latest_waiting_question` and the schema type are unchanged, so
  the shared reader's other consumers (`websocket.py`, `chat.py`)
  keep their existing behavior.

## Fail-closed checkpoint handling

When a reply's saved execution progress can't be resumed -- no
run-fenced checkpoint, a corrupt checkpoint, or a superseded legacy
partition -- the endpoint returns an error, releases the prelease back
to `waiting_for_user`, and leaves the task exactly where it was. For a
task that already has a `run_id`, the reply never fabricates a
different one or falls back to a transcript replay; the caller's only
path forward is to start a new task. (A legacy task whose `run_id` is
already `NULL` is a separate case, not covered by that guarantee: the
prelease claim mints a fresh run id for it, since there is no existing
run to preserve. That fresh run has no checkpoint fenced to it either,
so the same fail-closed error still applies once the checkpoint read
comes back empty -- the task lands back in `waiting_for_user` either
way, just by two different routes depending on whether a `run_id` was
already on record.) A transient infrastructure read failure gets the
distinct retryable `temporarily_unavailable` (503) instead of being
folded into the terminal `interaction_not_resumable` (409), so a
client can tell "give up" from "try again."

## Not idempotent

`reply()` carries no request-level idempotency key. The request body
has nothing to derive one from, and deriving one from server-side state
(e.g. `run_id` + question id) breaks the moment a caller legitimately
answers twice with different text -- a genuine correction would collide
on the same derived key as the first answer, and the runner's existing
`turn_id` reuse semantics turn a mismatched-content collision into a
`ValueError`. A client-side timeout on `reply()` does not tell the
caller whether the answer was delivered; callers that need to avoid a
duplicate should `GET` the task and confirm it has left
`waiting_for_user` before retrying.

## Compatibility

- `pending_interaction` is a pure field addition to
  `TaskInfoResponse`. An SDK that hasn't upgraded silently drops the
  unknown key (its `TypeAdapter` parse already discards unrecognized
  fields) -- this is not a breaking change for existing clients.
- The four new error codes are additive to the `V1ErrorCode` enum,
  which is documented as an allowed addition (see `errors.py`'s own
  docstring: "Adding new codes is allowed; renaming or removing
  existing ones is a breaking change"). An SDK built before this PR
  maps any code it doesn't recognize to a generic internal error, so
  an old SDK hitting the new `interaction_response_required` /
  `no_pending_interaction` / `interaction_not_resumable` /
  `temporarily_unavailable` codes degrades to that generic error
  rather than crashing or misparsing.
- `reply()` on the SDK side is a companion change tracked separately
  (see below); a client running against this server without the
  updated SDK simply has no way to call the new endpoint, which is a
  capability gap, not a compatibility break.

## Relationship to tracked issues

Part of xorbitsai/xagent#1079 (typed interaction endpoints, full
version) -- delivers its pending-interaction projection (verification
point 2) and the append error split (verification point 4). Not a
`Closes` reference: the rest of #1079 (interaction IDs, the
list/get/respond endpoint trio, an idempotency envelope) remains
tracked there and is deliberately out of scope here.

The companion SDK change (`reply()`, `pending_interaction` on
`TaskInfo`, the matching exception classes) is a separate PR against
xagent-sdk-python; this server change should land and deploy before
that SDK release is published.

## Test plan

- [x] `uv run pytest tests/web/api/v1/test_tasks.py
      tests/web/api/v1/test_task_reply.py tests/web/api/test_a2a_api.py
      tests/core/agent/test_runner.py
      tests/web/services/test_interaction_staging_production_gate.py`
      -- 225 passed.
- [x] Broader adjacent sweep (`tests/web/api/v1/`,
      `tests/web/api/test_a2a_api.py`, `tests/core/agent/test_runner.py`,
      the interaction-staging production gate,
      `tests/web/services/test_task_orchestrator.py`,
      `tests/web/test_chat_history_service.py`,
      `tests/web/test_task_lease_service.py`,
      `tests/web/services/test_acquire_task_lease_isolated.py`,
      `tests/web/services/test_managed_task_lease.py`) -- 510 passed.
- [x] New: a waiting task whose stored `interactions` list mixes dict
      and non-dict elements (e.g. a bare string) now returns 200 with
      the non-dict elements dropped from the projection, instead of
      500ing every `GET` on that task. Mutation-checked: removing the
      filter turns the ValidationError loose again and the test goes
      red.
- [x] `pre-commit run` over every changed/new file (ruff-check,
      ruff-format, mypy, isort, codespell, end-of-file,
      trailing-whitespace) -- clean.
- [x] Every new assertion was manually mutated once (field-drop,
      cache-boundary violation, status-guard widening, error-code
      fallback, checkpoint-branch collapse, prelease
      check-then-write race, legacy-run fail-closed bypass, ownership
      predicate bypass) and confirmed it goes red, then reverted and
      confirmed green. The concurrent-reply-race mutation (atomic
      UPDATE replaced with check-then-write) was additionally re-run 3
      times to rule out a timing-dependent false pass.
- [x] Re-verified the two WS cross-surface safety facts the reason
      change depends on against the current baseline: the
      `_TURN_REJECTION_MESSAGES` lookup on the WS layer is a
      default-valued `.get()` (no `KeyError` on an unrecognized
      reason), and WS never reaches `begin_turn` while a task is
      `WAITING_FOR_USER` (intercepted earlier by the live-control
      path per that code's own comment).
- [x] Ran the existing `test_interaction_staging_production_gate.py`
      unmodified -- still green, confirming this PR's new code never
      imports the still-unwired interaction-staging primitive.
- [x] The runner-layer authoritative semantics for "no checkpoint
      found" (an execution id with nothing recorded returns `None`
      rather than raising) is already covered by the existing
      `test_runner_treats_canonical_empty_checkpoint_as_authoritative`
      in `tests/core/agent/test_runner.py`; this PR's own tests cover
      the endpoint-layer translation of that `None` into
      `interaction_not_resumable` and don't duplicate the runner-level
      case.

### DIFF
--- a/src/xagent/web/api/v1/errors.py
+++ b/src/xagent/web/api/v1/errors.py
@@ -190,7 +190,7 @@ _DEFAULT_MESSAGES: dict[V1ErrorCode, str] = {
     V1ErrorCode.DELEGATED_AUTHORIZATION_FAILED: "Delegated authorization failed.",
     V1ErrorCode.INTERACTION_RESPONSE_REQUIRED: (
         "This task is waiting for an answer to a pending question; use "
-        "POST /v1/chat/tasks/{task_id}/reply instead."
+        "the task reply endpoint (POST .../reply) instead."
     ),
     V1ErrorCode.INTERACTION_NOT_RESUMABLE: (
         "This task's saved progress cannot be resumed; it remains in "
--- a/src/xagent/web/api/v1/errors.py
+++ b/src/xagent/web/api/v1/errors.py
@@ -122,6 +122,29 @@ class V1ErrorCode(str, Enum):
     MCP_OAUTH_AUTHORIZATION_FAILED = "mcp_oauth_authorization_failed"
     DELEGATED_AUTHORIZATION_FAILED = "delegated_authorization_failed"
 
+    # The task is waiting on an answer to a pending agent question and
+    # cannot accept a plain append. 409. Client should call
+    # ``POST /v1/chat/tasks/{id}/reply`` instead of retrying the append.
+    INTERACTION_RESPONSE_REQUIRED = "interaction_response_required"
+
+    # The task is waiting_for_user, but its saved execution progress
+    # cannot be resumed (no run-fenced checkpoint, or the checkpoint is
+    # unreadable/superseded). 409, not retryable -- the task stays in
+    # waiting_for_user and the caller must start a new task.
+    INTERACTION_NOT_RESUMABLE = "interaction_not_resumable"
+
+    # ``POST /v1/chat/tasks/{id}/reply`` was called on a task that is not
+    # currently waiting on a question (pending / paused / completed /
+    # failed). 409. Not retryable as-is; use ``append`` to start a new
+    # turn, or wait for the task to reach waiting_for_user.
+    NO_PENDING_INTERACTION = "no_pending_interaction"
+
+    # A reply's saved progress could not be read due to a transient
+    # infrastructure failure. 503, retryable -- distinct from
+    # ``interaction_not_resumable`` (which means resuming can never
+    # succeed) so clients know whether to retry or give up.
+    TEMPORARILY_UNAVAILABLE = "temporarily_unavailable"
+
 
 # Default human-readable text per code. Endpoints may override the
 # message when more specific context is available (e.g. surfacing
@@ -165,6 +188,20 @@ _DEFAULT_MESSAGES: dict[V1ErrorCode, str] = {
     V1ErrorCode.CONNECTOR_RUNTIME_UNAVAILABLE: "Connector runtime context is unavailable.",
     V1ErrorCode.MCP_OAUTH_AUTHORIZATION_FAILED: "MCP OAuth authorization is unavailable.",
     V1ErrorCode.DELEGATED_AUTHORIZATION_FAILED: "Delegated authorization failed.",
+    V1ErrorCode.INTERACTION_RESPONSE_REQUIRED: (
+        "This task is waiting for an answer to a pending question; use "
+        "POST /v1/chat/tasks/{task_id}/reply instead."
+    ),
+    V1ErrorCode.INTERACTION_NOT_RESUMABLE: (
+        "This task's saved progress cannot be resumed; it remains in "
+        "waiting_for_user. Start a new task instead of retrying."
+    ),
+    V1ErrorCode.NO_PENDING_INTERACTION: (
+        "This task has no pending question to answer."
+    ),
+    V1ErrorCode.TEMPORARILY_UNAVAILABLE: (
+        "The task's saved progress could not be read. Please retry."
+    ),
 }
 
 
@@ -214,6 +251,12 @@ _TURN_REJECTION_CODES: dict[str, tuple[V1ErrorCode, int]] = {
     # reaper) -- same "conversation no longer available" outcome as
     # workforce_run_not_found from the SDK's perspective.
     "workforce_run_not_active": (V1ErrorCode.TASK_NOT_FOUND, 404),
+    # The task is waiting on an answer to a pending question; append is
+    # not the right entrypoint for that -- POST .../reply is.
+    "interaction_response_required": (
+        V1ErrorCode.INTERACTION_RESPONSE_REQUIRED,
+        409,
+    ),
 }
 
 
--- a/src/xagent/web/api/v1/task_reply.py
+++ b/src/xagent/web/api/v1/task_reply.py
@@ -1,0 +1,515 @@
+"""``POST /v1/chat/tasks/{task_id}/reply``: answer a waiting task's question.
+
+A task in ``waiting_for_user`` has a paused execution, not an idle one --
+its agent asked a question via ``ask_user_question`` / a ``send_message``
+call with ``expect_response=True`` and is suspended waiting for the
+answer. Answering it means resuming that exact paused run, not starting
+a new turn the way ``POST /v1/chat/tasks/{task_id}/messages`` does. This
+module implements that resume, mirroring the shape of the A2A
+``message:send`` input-required resume in ``web/api/a2a.py``
+(``_resume_input_required_a2a_task``) with four differences:
+
+  - Scoped to ``Task.source == "sdk"`` plus the v1 SDK ownership rules
+    (agent-bound or workforce-bound key) instead of A2A's
+    ``Task.source == "a2a"`` plus a bare ``agent_id`` match.
+  - Only ``WAITING_FOR_USER`` is accepted; ``PAUSED`` is left to the
+    existing append endpoint, which already handles it and already has
+    its own failure policy for that status.
+  - Errors are translated to the stable ``V1ApiError`` envelope instead
+    of A2A's own error shape.
+  - The synthetic ``turn_id`` is derived from a fresh UUID
+    (``v1:reply:{task_id}:{uuid4()}``) rather than an A2A message id,
+    because the v1 request body carries no client-supplied message
+    identity to derive one from. This endpoint intentionally does not
+    implement idempotency: a client-side retry after a timeout can post
+    the answer twice. Callers that want to avoid that should ``GET`` the
+    task and confirm it has left ``waiting_for_user`` before retrying.
+
+The rest of the resume mechanics -- the two-phase prelease (claim, then
+either commit past it or restore it exactly), the heartbeat guarding the
+injection, and the fail-closed handling of every checkpoint outcome --
+are intentionally duplicated from ``a2a.py`` rather than factored into a
+shared helper. The two call sites answer to different protocols with
+independently evolving error contracts; a shared helper would couple
+them.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Any
+from uuid import uuid4
+
+from sqlalchemy import func
+
+from ....core.agent.checkpoint import (
+    CheckpointAccessRefusedError,
+    CheckpointCorruptError,
+    CheckpointReadError,
+)
+from ...models.database import get_session_local
+from ...models.task import Task, TaskStatus
+from ...schemas.v1 import ReplyRequest, ReplyResponse
+from ...services.db_runtime import (
+    cancel_and_drain_async_task,
+    drain_async_task_cancellation_safe,
+    run_db_io_cancellation_safe,
+)
+from ...services.task_execution_controller import TaskControlState
+from ...services.task_lease_service import (
+    TaskLease,
+    TaskLeaseHeartbeatOutcome,
+    TaskLeaseLostError,
+    acquire_task_lease_cancellation_safe,
+    acquire_task_lease_no_commit,
+    bind_task_lease_context,
+    release_task_lease_no_commit,
+    run_task_lease_heartbeat,
+    run_while_task_lease_owned,
+    stop_task_lease_heartbeat,
+)
+from .deps import ApiKeyPrincipal, record_key_usage
+from .errors import V1ApiError, V1ErrorCode
+from .tasks import _resolve_task_or_404, _validate_owner_scope
+
+logger = logging.getLogger(__name__)
+
+# Control states a waiting task may be claimed from. Mirrors the a2a
+# resume prelease's acceptance set for WAITING_FOR_USER: a task parked
+# there sits at either the plain WAITING_FOR_USER control state, or, on
+# a database left over from an interrupted prior resume attempt, IDLE.
+_RESUMABLE_CONTROL_STATES = frozenset(
+    {
+        TaskControlState.IDLE.value,
+        TaskControlState.WAITING_FOR_USER.value,
+    }
+)
+
+
+@dataclass(frozen=True)
+class _ReplyContext:
+    """Detached, already-authorized inputs the resume phase needs."""
+
+    task_id: int
+    agent_id: int
+    task_owner_user_id: int
+    run_id: str | None
+    text: str
+
+
+def _status_rejection(status: TaskStatus) -> V1ApiError:
+    """Map a non-waiting task status to its reply-rejection error.
+
+    RUNNING is the one truly transient case (a real turn is in flight;
+    retrying later can succeed), so it keeps the generic retryable
+    ``task_busy``. Every other non-waiting status means "there is
+    nothing to reply to right now" -- ``no_pending_interaction``, not
+    retryable as such. The caller's remedy differs by status: for
+    paused / completed / failed, ``append`` starts the next turn; for
+    pending, the task hasn't started executing yet -- it is not in
+    ``_APPENDABLE_STATUSES`` either, so there is no turn to append to
+    yet. The only remedy is to wait for it to leave pending.
+    """
+    if status == TaskStatus.RUNNING:
+        return V1ApiError(V1ErrorCode.TASK_BUSY, 409)
+    return V1ApiError(V1ErrorCode.NO_PENDING_INTERACTION, 409)
+
+
+def _prepare_reply_context_sync(
+    *,
+    task_id: int,
+    principal: ApiKeyPrincipal,
+    request: ReplyRequest,
+) -> _ReplyContext:
+    """Authorize the call and validate the body against the task's status.
+
+    A pure read -- no mutation happens here. The actual state transition
+    is the prelease claim in :func:`_acquire_reply_prelease_sync`, run
+    afterwards under ``acquire_task_lease_cancellation_safe``.
+    """
+    SessionLocal = get_session_local()
+    with SessionLocal() as db:
+        task = _resolve_task_or_404(task_id, principal, db)
+        _validate_owner_scope(
+            principal,
+            request_agent_id=request.agent_id,
+            request_workforce_id=request.workforce_id,
+        )
+        if request.message.files:
+            raise V1ApiError(
+                V1ErrorCode.INVALID_INPUT,
+                422,
+                message=(
+                    "files are not accepted on reply; attach files via a "
+                    "follow-up POST /v1/chat/tasks/{task_id}/messages call"
+                ),
+            )
+        if task.status != TaskStatus.WAITING_FOR_USER:
+            raise _status_rejection(task.status)
+        return _ReplyContext(
+            task_id=int(task.id),
+            agent_id=int(task.agent_id),
+            task_owner_user_id=int(task.user_id),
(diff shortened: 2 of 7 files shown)


src/xagent/web/api/v1/tasks.py
```diff
@@ -421,6 +425,46 @@ def _prepare_created_task_isolated(
...
+                V1ErrorCode.INVALID_INPUT,
+                422,
+                message="agent_id is required",
+            )
+        if request_agent_id != principal.agent.id:
+            raise V1ApiError(V1ErrorCode.AGENT_NOT_FOUND, 404)
+        if request_workforce_id is not None:
+            raise V1ApiError(V1ErrorCode.WORKFORCE_NOT_FOUND, 404)
+    else:
+        assert principal.workforce is not None
```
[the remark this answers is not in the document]
AlexLiu190625: This line is moved verbatim from the append endpoint into the shared helper rather than newly introduced. The asserted value is the server-constructed principal from the auth dependency — both constructors already refuse to build one without exactly one entity (`deps.py`) — while the actual client inputs (`agent_id` / `workforce_id` from the request body) are validated with explicit raises right below.

Converting the assert into a client-facing 404 would also cut against the boundary policy established in #1225: an invariant violation on a server-side object is a programming defect and should propagate loudly, not surface as a plausible-looking client error. The structural fix — making the exactly-one invariant a constructor guarantee on `ApiKeyPrincipal`, matching #1225 / #1288 — is tracked in #1310.

src/xagent/web/api/v1/task_reply.py
```diff
@@ -0,0 +1,515 @@
...
+    acquire_task_lease_no_commit,
+    bind_task_lease_context,
+    release_task_lease_no_commit,
+    run_task_lease_heartbeat,
+    run_while_task_lease_owned,
+    stop_task_lease_heartbeat,
+)
+from .deps import ApiKeyPrincipal, record_key_usage
+from .errors import V1ApiError, V1ErrorCode
+from .tasks import _resolve_task_or_404, _validate_owner_scope
```
rogercloud, line 76: Module-level import of private helpers back from `tasks.py` creates a latent import cycle that the route side must work around with an in-handler deferred import (tasks.py:1105) — hoisting the import would fail today. Consider moving `_resolve_task_or_404` / `_validate_owner_scope` to a shared module (e.g. `v1/_task_access.py` or `deps.py`) so the cycle disappears instead of being deferred.

src/xagent/web/api/v1/task_reply.py
```diff
@@ -0,0 +1,515 @@
...
+                Task.id == task_id,
+                Task.agent_id == agent_id,
+                Task.source == "sdk",
+                Task.status == TaskStatus.WAITING_FOR_USER,
+                Task.control_state.in_(_RESUMABLE_CONTROL_STATES),
+            )
+            .update(
+                {
+                    Task.control_state: TaskControlState.RESUME_REQUESTED.value,
+                    Task.state_version: func.coalesce(Task.state_version, 0) + 1,
```
rogercloud, line 224: The claim UPDATE is fenced on id/agent_id/source/status/control_state but not on `state_version`; prepare (121-152) and claim are separate transactions, and `run_id` survives across waiting episodes, so a delayed stale claim could inject Q1's answer into Q2's wait on the same run. Capture `Task.state_version` in prepare and add it to the claim filter — safe, since the claim bumps the version in the same UPDATE. (The A2A mirror at a2a.py:195-241 has the same pre-existing gap.)

src/xagent/web/schemas/v1.py
```diff
@@ -489,6 +606,19 @@ class TaskInfoResponse(BaseModel):
             "(completed or failed). Null while still running."
         ),
     )
+    pending_interaction: Optional[PendingInteraction] = Field(
+        None,
+        description=(
+            "Convenience projection of the agent's most recent unanswered "
```
rogercloud: "Unanswered" overclaims: `get_latest_waiting_question` (chat_history_service.py:469-491) returns the latest question row with no answered marker (`TaskChatMessage` has no such column), and a re-wait with an empty question — which skips persistence (websocket.py:795) — surfaces the previously answered question, so the server cannot prove this is unanswered. Same wording at the `PendingInteraction` docstring (schemas/v1.py:532). Reword to "most recent question" until an answered marker exists.

src/xagent/web/api/v1/task_reply.py
```diff
@@ -0,0 +1,515 @@
...
+    from .. import websocket
+
+    if task_lease.task_id != task_id or task_lease.run_id is None:
+        raise ValueError("Reply resume scheduling requires an exact task lease")
+    if not websocket.background_task_manager.reserve_resume(task_id):
+        # A same-process resume is already registered for this task despite
+        # the exclusive DB claim above -- report it the same way a caller
+        # would read a lost race on the claim itself, rather than a bare
+        # 500: retrying (after the other resume settles) can succeed.
+        raise V1ApiError(V1ErrorCode.TASK_BUSY, 409)
```
rogercloud, line 336: On reserve conflict, the rejected reply text is already persisted in `Task.input` (456-458) and the restore (223-240) clears `error_message` but not `input`, so a subsequent GET shows the rejected text until a retry overwrites it. Faithful mirror of the pre-existing A2A order — optionally snapshot and restore the prior input value in `_ReplyContext`.

src/xagent/web/schemas/v1.py
```diff
@@ -439,6 +528,31 @@ class CreateWorkforceRunResponse(BaseModel):
...
+    against a fixed schema, because the seven-value ``type`` enum lives
+    with the agent tool and duplicating it here would be a second source
+    of truth to keep in sync. ``[]`` and ``null`` both mean "no
+    structured control, answer with plain text" and are intentionally
+    not normalized to one value server-side -- callers should treat them
+    the same way.
+    """
+
+    question: str = Field(..., description="The agent's pending question text.")
+    interactions: Optional[List[Dict[str, Any]]] = Field(
```
rogercloud, line 546: `Optional[List[Dict[str, Any]]]` is more typed than the opaque pass-through contract: the read path (tasks.py:888-892) has no item guard, so a non-dict item raises pydantic `ValidationError` → GET 500. No current in-repo producer emits one (react.py normalizes; the generic websocket write gate is list-only), but `runtime.send_message` passes metadata through unvalidated. Use `List[Any]` or filter at read.

src/xagent/web/api/v1/errors.py
```diff
@@ -165,6 +188,20 @@ class V1ErrorCode(str, Enum):
     V1ErrorCode.CONNECTOR_RUNTIME_UNAVAILABLE: "Connector runtime context is unavailable.",
     V1ErrorCode.MCP_OAUTH_AUTHORIZATION_FAILED: "MCP OAuth authorization is unavailable.",
     V1ErrorCode.DELEGATED_AUTHORIZATION_FAILED: "Delegated authorization failed.",
+    V1ErrorCode.INTERACTION_RESPONSE_REQUIRED: (
```
rogercloud, line 191: Default message contains a literal `{task_id}` that is never interpolated — `raise_for_turn_rejection` (263-274) raises the raw default and the handler emits it verbatim to clients. Same literal-brace pattern at task_reply.py:142-148. Drop the placeholder or override the message with the real task_id at the raise sites (tasks.py:637, 1062 know it).

src/xagent/web/api/v1/task_reply.py
```diff
@@ -0,0 +1,515 @@
...
+        return _ReplyContext(
+            task_id=int(task.id),
+            agent_id=int(task.agent_id),
+            task_owner_user_id=int(task.user_id),
+            run_id=str(task.run_id) if task.run_id is not None else None,
+            text=request.message.content,
+        )
+
+
+def _acquire_reply_prelease_sync(
```
rogercloud, line 189: Worth documenting the serialization invariant here: this endpoint never goes through `task_execution_controller.command(task_id)` (unlike a2a.py:790-792) — the atomic claim is the sole serializer, currently proven only by `test_concurrent_reply_race_exactly_one_winner`. The invariant: the claim arbitrates reply-vs-reply and reply-vs-append; A2A is source-disjoint (`source == 'a2a'`); the WS overlap is fenced by the resume reservation plus the run-fenced lease — WS handlers have no source gate, so it is not "WS cannot touch sdk tasks".

src/xagent/web/api/v1/task_reply.py
```diff
@@ -0,0 +1,515 @@
...
+                synchronize_session=False,
+            )
+        )
+        if claimed != 1:
+            db.rollback()
+            return None
+        task_lease = acquire_task_lease_no_commit(
+            db,
+            task_id,
+            expected_run_id=previous_run_id,
```
rogercloud, line 246: For a legacy waiting row with `run_id=NULL`, `expected_run_id=None` lets the lease service mint a fresh run_id (task_lease_service.py:693), and the fail-closed restore never reverts it — so "leaves the task exactly where it was" / "never fabricates a run" in the PR description is false on this path. Document it, or restore NULL for rows that had none.

src/xagent/web/services/task_orchestrator.py
```diff
@@ -985,12 +987,14 @@ def _claim_turn_no_commit(
...
         owned = (
-            db.query(Task.id)
+            db.query(Task.id, Task.status)
             .filter(Task.id == task_id, Task.user_id == task_owner_user_id)
             .first()
         )
         if owned is None:
             raise TaskTurnNotFoundError(task_id)
+        if owned.status == TaskStatus.WAITING_FOR_USER:
+            raise TaskTurnError("interaction_response_required")
```
rogercloud, line 1000: The new `interaction_response_required` reason is missing from the `Raises` docs of `_begin_turn_unserialized` (443-446) and `_begin_turn_atomic_sync` (1085), which still document only `TaskTurnError("busy")`. Add it to both.

rogercloud (approved):

src/xagent/web/schemas/v1.py
AlexLiu190625: Reworded in 2924b6b7a. Both the schema docstring and the field description now say "most recent question": `get_latest_waiting_question` returns the highest-id question row and `task_chat_messages` has no answered marker, so the server cannot claim more than that.

A related inaccuracy went with it. The field description also implied the projection is null whenever a task reaches waiting_for_user with an empty message body. That only holds when no question row exists at all — on a re-wait with an empty body after an earlier question was answered, the earlier row is still the latest and is what gets returned. The description now says so explicitly.

src/xagent/web/api/v1/errors.py
AlexLiu190625: The placeholder is gone as of 2924b6b7a — the message now names the endpoint without a path template, and the same change was applied to the 422 text on the reply handler.

Not taking the other half of the suggestion (overriding the message with the real task_id at the raise sites). The code is the stable contract and the message is meant to be constant per code: varying it per request would make it unpinnable in fixtures and would give one code two different strings depending on the call path. A client that needs the id already has it in the request it sent.

src/xagent/web/api/v1/task_reply.py
AlexLiu190625: Documented in 2924b6b7a, stated in the order the guarantees actually hold:

- Reply and append are status-disjoint by construction, not merely arbitrated after the fact: `_APPENDABLE_STATUSES` no longer contains `WAITING_FOR_USER` and the reply claim accepts only `WAITING_FOR_USER`, so the two conditional UPDATEs cannot both match.
- Reply against reply is arbitrated by the atomic claim, covered by `test_concurrent_reply_race_exactly_one_winner`.
- A2A is source-disjoint (`source == 'a2a'`).
- The WebSocket overlap is fenced by the resume reservation plus the run-fenced lease.

Your correction is in the note as well: WS handlers have no source gate, so the reason is not that WS cannot touch sdk tasks. That is written down as explicitly not being the reason, so a later reader does not conclude serialization is missing and wrap the handler in `task_execution_controller.command`.

src/xagent/web/schemas/v1.py
AlexLiu190625: Fixed in baf4d7015. The v1 read path now drops non-mapping elements before constructing `PendingInteraction`, with a test that fails without the filter.

Not taking the `List[Any]` option. The SDK types this as `list[dict[str, Any]]` and that type is already frozen, so loosening the server annotation would let the server emit values the client cannot parse — it relocates the failure from a server 500 to a client-side parse error rather than removing it.

The filter was also deliberately kept out of `get_latest_waiting_question`: that reader has four other call sites (chat.py twice, websocket.py twice) whose output would change with it. It sits in the v1 projection instead, the only consumer with a strict schema, and follows react.py's existing normalization rule — a non-dict element is dropped, not coerced.

src/xagent/web/api/v1/task_reply.py
AlexLiu190625: Confirmed. The PR description now scopes that claim to tasks that have a run id, and there is a note at the `expected_run_id` hand-off (2924b6b7a).

One fact to add to yours: on that path the lease acquisition also clears `last_checkpoint_event_id` and `last_checkpoint_trace_event_id` — the `expected_run_id is None` branch, whose case predicate is true for any non-RUNNING row — and neither is restored either. So three columns change irreversibly on such a row, not one.

Not restoring NULL in code. The path is reachable only for rows that predate run ids; those rows have no run-fenced checkpoint to resume onto, so the reply fails closed regardless, and the minted id is inert. Adding a conditional to the restore path, the most delicate part of the module, to undo an inert side effect on legacy rows is not a trade worth making here.

src/xagent/web/api/v1/task_reply.py
AlexLiu190625: Confirmed, and tracked in #1322 rather than fixed here.

The sequence holds. `release_task_lease_no_commit` does not write `run_id`, so a task that goes waiting → running → waiting again keeps the same run id, and the claim's five filters (`id`, `agent_id`, `source`, `status`, `control_state`) all match again on the second wait. Nothing in the current claim can tell the two waits apart.

Two things put it in an issue rather than in this PR.

It is the same hazard class as the cross-request retry this endpoint already documents as accepted: a client that times out and re-posts can land its answer on a question raised after the first attempt succeeded. That window is seconds wide; the one here needs the prepare/claim gap to stretch across an entire agent turn. A `state_version` fence closes the narrow window and does not close the wide one, since a retry re-reads state fresh and would capture the new version.

The A2A claim has the identical gap. Fencing only this side would add a fifth divergence to the four this module documents against the A2A resume, and would leave the two mirrors inconsistent on exactly the invariant that matters most. #1322 covers both call sites together.

The issue also records the implementation trap: the fence has to compare `func.coalesce(Task.state_version, 0)`, because a bare `Task.state_version == n` never matches a row whose column is NULL, which would make legacy rows permanently unanswerable.

src/xagent/web/api/v1/task_reply.py
AlexLiu190625: Confirmed — hoisting it would fail today, since both helpers are defined well past the top of `tasks.py` and would not be bound yet when `task_reply` executes.

Moving them is tracked in #1310 alongside the `ApiKeyPrincipal` consolidation, which is where these ownership-resolution entry points belong. Leaving it out of this PR: the move touches five call sites in `tasks.py`, a docstring reference in `workforces.py`, and a test, all in the files this change already concentrates on.

src/xagent/web/api/v1/task_reply.py
AlexLiu190625: Confirmed, and it is one column wider than described: `_update_reply_input_sync` sets `input`, `output` and `error_message` in the same UPDATE, while the release restores only `error_message`. So a rejected reply also clears whatever `output` the row held.

Tracked in #1284 with both call sites recorded, since the A2A resume orders the same two steps the same way. Not changing it here — snapshotting and restoring prior column values would add branches to the rollback path for a discrepancy the next successful reply overwrites.

src/xagent/web/services/task_orchestrator.py
AlexLiu190625: Added in 2924b6b7a to both `_begin_turn_unserialized` and `_begin_turn_atomic_sync`. `_claim_turn_no_commit`, where the raise itself lives, has a single-line docstring with no `Raises` section and was left as is.

PR #1316, opened 2026-08-13
fix(skill): contradict skill guidance that assumes edit_image is registered

### LINKED ISSUE #1307: [Bug] The static-visual-design skill instructs edit_image regardless of whether it is registered
Sub-issue of #1289, split out from the #1294 review so it can be tracked and closed on its own.

## Problem

`static-visual-design` instructs the agent to use `edit_image` and the `images`
argument unconditionally. The text is injected into context verbatim with no view
of which tools were actually registered, so on a deployment where no image model
declares the `edit` ability the agent is told to use a tool that is not in its
schema.

Verbatim, `src/xagent/skills/builtin/static-visual-design/SKILL.md`:

- **:192** — "Pass the actual path, URL, or file_id through the image tool's
  `images` argument (or use `edit_image` directly); naming the asset only in the
  prompt does not attach it."
- **:326** — "Use `edit_image` only to refine a strong, single-canvas candidate
  with a localized defect."
- **:354** — "A successful `generate_image` or `edit_image` call is never
  completion evidence by itself"

And `references/static-ad-art-direction.md`:

- **:296** — "Use `edit_image` only for a localized defect on an otherwise
  strong, single-canvas composition"

## Why this is now reachable

#1294 stops registering `edit_image` when no configured model declares the `edit`
ability, which is the right fix for the production failure in #1289 (8 doomed
`edit_image` calls, then 16 `execute_python_code` fallbacks). But it makes the
skill text wrong rather than merely optimistic: previously the tool existed and
failed with an error; now the agent is instructed toward a tool that does not
exist in its schema at all.

`:192` is the more damaging of the two mechanisms, because `images` is still a
real parameter on `generate_image` and delegates into the edit path. The skill
actively teaches the argument that routes into the unavailable capability.

## Why it cannot be fixed by editing the text

The skill layer has no capability awareness to condition on. `SkillParser` stores
the whole file as one blob (`src/xagent/skills/parser.py:80`, `"content": content,
# Complete SKILL.md content`), and grepping the skills package for `has_ability`,
`abilities`, `registered_tools`, or `available_tools` returns nothing. There is
no seam where "editing is unavailable here" could be substituted today.

## Possible directions

Not proposing a specific one — the right shape depends on how much capability
awareness the skill system should grow:

1. **Conditional sections** — a lightweight mechanism for skill text to be
   included based on registered tools/abilities. Most general, most work.
2. **Runtime capability preamble** — leave skill text alone and prepend a short
   authoritative note about which media tools are actually registered, so the
   agent has a later, more specific instruction to follow. Cheapest; relies on
   the model preferring the more specific statement.
3. **Rewrite the skill to be capability-neutral** — phrase the guidance in terms
   of intent ("refine a localized defect" / "attach a reference asset") and let
   the tool descriptions dictate the mechanism. No new machinery, but the skill
   loses concrete tool names and the `images` guidance in `:192` is genuinely
   about a specific parameter.

Option 2 alone does not fix `:192`, since that line teaches an argument rather
than a tool.

## Acceptance

- [ ] On a deployment with no edit-capable image model, the injected context does
      not leave the agent believing `edit_image` or `images` is usable — either by
      omitting the instruction or by authoritatively contradicting it.
- [ ] The skill still gives complete guidance on a fully capable deployment.
- [ ] Whichever mechanism is chosen is covered by a test that fails if the
      capability condition is dropped, rather than only asserting the text.

## Update: why the first acceptance line was relaxed

It originally required that nothing in the injected skill context mention
`edit_image` or `images`. Meeting that literally requires option 1 — conditional
sections in the skill system — because the text is stored and injected as a
single blob.

That is disproportionate here. The skill is 376 lines; exactly 3 of them name
`edit_image` or `images`. The rest is design method (brief, campaign directions,
canvas structure, per-candidate inspection, the completion gate, identity-asset
handling) that is entirely independent of edit capability, and that a
creative-deliverable task needs — the zero-delivery run in #1289 is what missing
guidance of that kind looks like. Withholding the skill, or building machinery to
strip three lines from it, both cost more than they return.

Relaxed to an outcome-based criterion: the agent must not end up believing the
capability exists. A correction that follows the guidance satisfies that, and it
is the same mechanism #1294 already uses in the `generate_image` description, so
the model gets one consistent story from both sources.

## Out of scope

The unconditional `images` parameter on `generate_image` and its delegation into
the edit path — that is tool-surface work tracked under #1289's B3 and #1299.

bluefish-08:
Closes #1307.

## Problem

`static-visual-design` names `edit_image` and the `images` argument
unconditionally. Skill text is injected verbatim — `SkillParser` stores the whole
file as one blob (`src/xagent/skills/parser.py:80`) — and grepping the skills
package for `has_ability`, `abilities`, `registered_tools`, or `available_tools`
returns nothing. So on a generate-only deployment the agent is instructed to use
a tool that is not in its schema, by text that has no way of knowing that.

#1294 made this reachable: `edit_image` used to exist and fail, so the skill was
merely optimistic. Now it points at something absent.

## Approach

Of the three directions in #1307 this takes the runtime correction, emitted
**after** the skill guidance in the system context. The skill cannot know what
was registered, so the contradiction has to come from something that does — and
it has to come later to win.

Two pieces:

- **`ReActPattern` records the capability.** It reuses the
  `_tool_decision_groups_by_name` map it already builds each turn for the circuit
  breaker, so there is no extra pass over the tools. DAG steps and single_call
  both construct `ReActPattern`s and inherit it.
- **`ExecutionContext.get_system_context` emits the correction**, but only when
  skill guidance is actually present. Without that condition every image task
  carries the paragraph, and the no-skill case is already covered by the
  `IMAGE EDITING IS UNAVAILABLE` prefix #1294 puts on the `generate_image`
  description.

Three decisions worth flagging:

**The flag is `generate_image present && edit_image absent`**, not just
`edit_image absent`. A deployment with no image tools at all should not be told
that image editing is unavailable.

**The correction denies the argument, not only the tool.** `SKILL.md:192` teaches
`images`, and `images` remains a real parameter on `generate_image` that
delegates into the edit path. Denying `edit_image` alone leaves that route open.

**It offers a way forward** ("describe in the prompt what a reference would have
contributed") rather than only a prohibition, so the model has somewhere to go.

## Why not withhold the skill instead

The obvious alternative — do not load a skill whose guidance is partly
inapplicable — costs far more than it saves. The skill is 376 lines; exactly 3
name `edit_image` or `images`. The rest is design method: establishing the brief,
developing campaign directions before rendering, single-canvas structure,
per-candidate `understand_media` inspection, identity-asset handling, and the
completion gate. None of it depends on edit capability, and a creative-deliverable
task needs it — the zero-delivery run in #1289 is what its absence looks like.

Same shape as the `list_image_models` decision in #1294: the unit of
unavailability is one tool, not the whole box.

## Tests

Both directions plus ordering: the correction appears only when editing is
unavailable *and* a skill is loaded, and it lands after the guidance it
contradicts.

Mutation-tested — each of these turns the suite red:

| Mutation | Why it matters |
|---|---|
| Delete the call in the tool-calling loop | The helper's own test passes without it; this is wiring, not logic |
| Drop the `generate_image` precondition | Would add noise on deployments with no image tools |
| Disable the correction block | Base behaviour |
| Move the correction before the skill text | Ordering *is* the mechanism |

The first one is the reason there is an end-to-end test through `pattern.run()`
alongside the unit test — the unit test alone left the wiring unguarded.

Full `tests/core/agent` passes.

## Note on the acceptance criteria

#1307's first acceptance line originally required that nothing in the injected
skill context mention `edit_image` or `images`, which requires conditional
sections in the skill system. I relaxed it to an outcome-based criterion and
recorded the reasoning in the issue.

### DIFF
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -56,6 +56,7 @@ COMPACT_DROPPED_TOOL_NAME_MAX_CHARS = 64
 # nothing a dropped observation held.
 NON_EVIDENCE_TOOL_NAMES = CONTROL_TOOL_NAMES | {LOAD_SKILL_TOOL_NAME}
 COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
+IMAGE_EDIT_UNAVAILABLE_METADATA_KEY = "image_edit_unavailable"
 
 
 def _utcnow() -> datetime:
@@ -606,6 +607,19 @@ class ExecutionContext:
                 "Selected skill guidance. Use it when relevant to the current task:\n"
                 f"{str(skill_context).strip()}"
             )
+            # Skill text is injected verbatim and cannot know which tools were
+            # registered, so the correction has to come after it.
+            if self.metadata.get(IMAGE_EDIT_UNAVAILABLE_METADATA_KEY):
+                parts.append(
+                    "Correction to the skill guidance above: image editing is "
+                    "unavailable here. No configured image model has the edit "
+                    "ability, so edit_image is not in your tools and passing "
+                    "images to generate_image fails. Ignore any instruction to "
+                    "edit an existing image or to attach a reference through "
+                    "images; render each deliverable from a text prompt in one "
+                    "call, and describe in the prompt what a reference would "
+                    "have contributed."
+                )
         return "\n\n".join(part for part in parts if part.strip())
 
     def get_recent_messages(self, n: int = 10) -> list[Message]:
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -35,7 +35,11 @@ from .components import (
     WorkspaceComponent,
     clone_component,
 )
-from .enrichment import MEMORY_CONTEXT_METADATA_KEY, SKILL_CONTEXT_METADATA_KEY
+from .enrichment import (
+    IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
+    MEMORY_CONTEXT_METADATA_KEY,
+    SKILL_CONTEXT_METADATA_KEY,
+)
 from .memory_tool import MEMORY_TOOLS_METADATA_KEY
 from .message import LLMCallRecord, Message
 from .skill_tool import (
@@ -56,10 +60,6 @@ COMPACT_DROPPED_TOOL_NAME_MAX_CHARS = 64
 # nothing a dropped observation held.
 NON_EVIDENCE_TOOL_NAMES = CONTROL_TOOL_NAMES | {LOAD_SKILL_TOOL_NAME}
 COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
-# True only when generate_image is registered and edit_image is not; a deployment
-# with no image tools at all leaves it False. Lives here, not in enrichment.py
-# with the other metadata keys, because react.py already imports from this module.
-IMAGE_EDIT_UNAVAILABLE_METADATA_KEY = "image_edit_unavailable"
 
 
 def _utcnow() -> datetime:
@@ -612,9 +612,12 @@ class ExecutionContext:
             )
             # Skill text is injected verbatim and cannot know which tools were
             # registered, so the correction has to come after it.
-            if self.metadata.get(
-                IMAGE_EDIT_UNAVAILABLE_METADATA_KEY
-            ) and "edit_image" in str(skill_context):
+            if (
+                self.metadata.get(IMAGE_EDIT_UNAVAILABLE_METADATA_KEY)
+                and "edit_image" in str(skill_context).lower()
+            ):
+                # Last sentence pair mirrors the capability prefix in
+                # tools/adapters/vibe/image_tool.py; edit both together.
                 parts.append(
                     "Correction to the skill guidance above: image editing is "
                     "unavailable here. edit_image is not in your tools, and "
--- a/src/xagent/core/agent/context/execution.py
+++ b/src/xagent/core/agent/context/execution.py
@@ -56,6 +56,9 @@ COMPACT_DROPPED_TOOL_NAME_MAX_CHARS = 64
 # nothing a dropped observation held.
 NON_EVIDENCE_TOOL_NAMES = CONTROL_TOOL_NAMES | {LOAD_SKILL_TOOL_NAME}
 COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
+# True only when generate_image is registered and edit_image is not; a deployment
+# with no image tools at all leaves it False. Lives here, not in enrichment.py
+# with the other metadata keys, because react.py already imports from this module.
 IMAGE_EDIT_UNAVAILABLE_METADATA_KEY = "image_edit_unavailable"
 
 
@@ -609,16 +612,18 @@ class ExecutionContext:
             )
             # Skill text is injected verbatim and cannot know which tools were
             # registered, so the correction has to come after it.
-            if self.metadata.get(IMAGE_EDIT_UNAVAILABLE_METADATA_KEY):
+            if self.metadata.get(
+                IMAGE_EDIT_UNAVAILABLE_METADATA_KEY
+            ) and "edit_image" in str(skill_context):
                 parts.append(
                     "Correction to the skill guidance above: image editing is "
-                    "unavailable here. No configured image model has the edit "
-                    "ability, so edit_image is not in your tools and passing "
-                    "images to generate_image fails. Ignore any instruction to "
-                    "edit an existing image or to attach a reference through "
-                    "images; render each deliverable from a text prompt in one "
-                    "call, and describe in the prompt what a reference would "
-                    "have contributed."
+                    "unavailable here. edit_image is not in your tools, and "
+                    "passing images to generate_image routes into the same "
+                    "absent edit path. Ignore any instruction to edit an "
+                    "existing image or to attach a reference through images; "
+                    "render each deliverable from a text prompt in one call, "
+                    "and describe in the prompt what a reference would have "
+                    "contributed."
                 )
         return "\n\n".join(part for part in parts if part.strip())
 
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -390,7 +390,12 @@ class ReActPattern(AgentPattern):
     ) -> dict[str, Any]:
         self.status = "thinking"
         self._tool_decision_groups_by_name = self._tool_decision_groups_for_tools(tools)
-        self._record_image_edit_availability(context)
+        # Skill text names edit_image unconditionally; get_system_context needs
+        # this to contradict it.
+        context.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = (
+            "generate_image" in self._tool_decision_groups_by_name
+            and "edit_image" not in self._tool_decision_groups_by_name
+        )
         base_tool_schemas = (
             []
             if self.tool_choice == "none"
@@ -1120,20 +1125,6 @@ class ReActPattern(AgentPattern):
             if isinstance(function, dict) and function.get("name"):
                 names.append(str(function["name"]))
         return names
-
-    def _record_image_edit_availability(self, context: Any) -> None:
-        """Flag a deployment that can generate images but not edit them.
-
-        Injected skill text names edit_image unconditionally; the system context
-        needs this to contradict it.
-        """
-        metadata = getattr(context, "metadata", None)
-        if metadata is None:
-            return
-        names = self._tool_decision_groups_by_name
-        metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = (
-            "generate_image" in names and "edit_image" not in names
-        )
 
     def _tool_decision_groups_for_tools(self, tools: list[Any]) -> dict[str, str]:
         groups: dict[str, str] = {}
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -50,10 +50,10 @@ from ....tools.user_interaction import (
 )
 from ...clarification import draft_from_waiting_request
 from ...context.enrichment import (
+    IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
     enrich_context_with_memory,
     latest_user_text,
 )
-from ...context.execution import IMAGE_EDIT_UNAVAILABLE_METADATA_KEY
 from ...context.memory_tool import build_memory_tools
 from ...context.skill_tool import build_load_skill_tool
 from ...grounding import grounding_rule
@@ -390,8 +390,7 @@ class ReActPattern(AgentPattern):
     ) -> dict[str, Any]:
         self.status = "thinking"
         self._tool_decision_groups_by_name = self._tool_decision_groups_for_tools(tools)
-        # Skill text names edit_image unconditionally; get_system_context needs
-        # this to contradict it.
+        # Read by get_system_context to contradict skill text naming edit_image.
         context.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = (
             "generate_image" in self._tool_decision_groups_by_name
             and "edit_image" not in self._tool_decision_groups_by_name
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -53,6 +53,7 @@ from ...context.enrichment import (
     enrich_context_with_memory,
     latest_user_text,
 )
+from ...context.execution import IMAGE_EDIT_UNAVAILABLE_METADATA_KEY
 from ...context.memory_tool import build_memory_tools
 from ...context.skill_tool import build_load_skill_tool
 from ...grounding import grounding_rule
@@ -389,6 +390,7 @@ class ReActPattern(AgentPattern):
     ) -> dict[str, Any]:
         self.status = "thinking"
         self._tool_decision_groups_by_name = self._tool_decision_groups_for_tools(tools)
+        self._record_image_edit_availability(context)
         base_tool_schemas = (
             []
             if self.tool_choice == "none"
@@ -1118,6 +1120,20 @@ class ReActPattern(AgentPattern):
             if isinstance(function, dict) and function.get("name"):
                 names.append(str(function["name"]))
         return names
+
+    def _record_image_edit_availability(self, context: Any) -> None:
+        """Flag a deployment that can generate images but not edit them.
+
+        Injected skill text names edit_image unconditionally; the system context
+        needs this to contradict it.
+        """
+        metadata = getattr(context, "metadata", None)
+        if metadata is None:
+            return
+        names = self._tool_decision_groups_by_name
+        metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = (
+            "generate_image" in names and "edit_image" not in names
+        )
 
     def _tool_decision_groups_for_tools(self, tools: list[Any]) -> dict[str, str]:
         groups: dict[str, str] = {}
--- a/tests/core/agent/test_context.py
+++ b/tests/core/agent/test_context.py
@@ -619,6 +619,48 @@ def test_get_messages_for_llm_coalesces_system_messages() -> None:
     assert result[2] == {"role": "user", "content": "hello"}
 
 
+def test_skill_guidance_is_corrected_when_image_editing_is_unavailable() -> None:
+    from xagent.core.agent.context.execution import (
+        IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
+    )
+
+    ctx = ExecutionContext(system_prompt="Base prompt.")
+    ctx.metadata[SKILL_CONTEXT_METADATA_KEY] = (
+        "## Available Skill: static-visual-design\n\nUse `edit_image` to refine."
+    )
+    ctx.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
+    ctx.add_user_message("make an ad")
+
+    system_content = ctx.get_messages_for_llm()[0]["content"]
+
+    assert "image editing is unavailable here" in system_content
+    # The correction is worthless unless it lands after the text it contradicts.
+    assert system_content.index("Selected skill guidance") < system_content.index(
+        "Correction to the skill guidance above"
+    )
+
+
+def test_no_correction_when_editing_works_or_no_skill_is_loaded() -> None:
+    from xagent.core.agent.context.execution import (
+        IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
+    )
+
+    with_edit = ExecutionContext(system_prompt="Base prompt.")
+    with_edit.metadata[SKILL_CONTEXT_METADATA_KEY] = "## Available Skill: design"
+    with_edit.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = False
+    with_edit.add_user_message("x")
+
+    no_skill = ExecutionContext(system_prompt="Base prompt.")
+    no_skill.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
+    no_skill.add_user_message("x")
+
+    for ctx in (with_edit, no_skill):
+        assert (
+            "Correction to the skill guidance"
+            not in (ctx.get_messages_for_llm()[0]["content"])
+        )
+
+
 def test_get_messages_for_llm_injects_memory_and_skill_context() -> None:
     ctx = ExecutionContext(system_prompt="Base prompt.")
     ctx.metadata[MEMORY_CONTEXT_METADATA_KEY] = (
--- a/tests/core/agent/test_context.py
+++ b/tests/core/agent/test_context.py
@@ -620,7 +620,7 @@ def test_get_messages_for_llm_coalesces_system_messages() -> None:
 
 
 def test_skill_guidance_is_corrected_when_image_editing_is_unavailable() -> None:
-    from xagent.core.agent.context.execution import (
+    from xagent.core.agent.context.enrichment import (
         IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
     )
 
@@ -635,15 +635,28 @@ def test_skill_guidance_is_corrected_when_image_editing_is_unavailable() -> None
 
     assert "image editing is unavailable here" in system_content
     # The correction is worthless unless it lands after the text it contradicts.
-    # Anchor on the guidance body, not its header, so splitting the skill block
-    # cannot silently make this pass for the wrong reason.
     assert system_content.index("Use `edit_image` to refine.") < system_content.index(
         "Correction to the skill guidance above"
     )
 
 
+def test_correction_matches_edit_image_case_insensitively() -> None:
+    from xagent.core.agent.context.enrichment import (
+        IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
+    )
+
+    ctx = ExecutionContext(system_prompt="Base prompt.")
+    ctx.metadata[SKILL_CONTEXT_METADATA_KEY] = "Call EDIT_IMAGE to refine."
+    ctx.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
+    ctx.add_user_message("x")
+
+    assert (
+        "image editing is unavailable here" in ctx.get_messages_for_llm()[0]["content"]
+    )
+
+
 def test_no_correction_when_editing_works_or_the_skill_never_named_it() -> None:
-    from xagent.core.agent.context.execution import (
+    from xagent.core.agent.context.enrichment import (
         IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
     )
 
@@ -656,7 +669,6 @@ def test_no_correction_when_editing_works_or_the_skill_never_named_it() -> None:
     no_skill.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
     no_skill.add_user_message("x")
 
-    # 8 of the 9 builtin skills never mention edit_image; they get no correction.
     unrelated_skill = ExecutionContext(system_prompt="Base prompt.")
     unrelated_skill.metadata[SKILL_CONTEXT_METADATA_KEY] = "Write the report first."
     unrelated_skill.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
--- a/tests/core/agent/test_context.py
+++ b/tests/core/agent/test_context.py
@@ -635,18 +635,20 @@ def test_skill_guidance_is_corrected_when_image_editing_is_unavailable() -> None
 
     assert "image editing is unavailable here" in system_content
     # The correction is worthless unless it lands after the text it contradicts.
-    assert system_content.index("Selected skill guidance") < system_content.index(
+    # Anchor on the guidance body, not its header, so splitting the skill block
+    # cannot silently make this pass for the wrong reason.
+    assert system_content.index("Use `edit_image` to refine.") < system_content.index(
         "Correction to the skill guidance above"
     )
 
 
-def test_no_correction_when_editing_works_or_no_skill_is_loaded() -> None:
+def test_no_correction_when_editing_works_or_the_skill_never_named_it() -> None:
     from xagent.core.agent.context.execution import (
         IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
     )
 
     with_edit = ExecutionContext(system_prompt="Base prompt.")
-    with_edit.metadata[SKILL_CONTEXT_METADATA_KEY] = "## Available Skill: design"
+    with_edit.metadata[SKILL_CONTEXT_METADATA_KEY] = "Use `edit_image` to refine."
     with_edit.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = False
     with_edit.add_user_message("x")
 
@@ -654,7 +656,13 @@ def test_no_correction_when_editing_works_or_no_skill_is_loaded() -> None:
     no_skill.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
     no_skill.add_user_message("x")
 
-    for ctx in (with_edit, no_skill):
+    # 8 of the 9 builtin skills never mention edit_image; they get no correction.
+    unrelated_skill = ExecutionContext(system_prompt="Base prompt.")
+    unrelated_skill.metadata[SKILL_CONTEXT_METADATA_KEY] = "Write the report first."
+    unrelated_skill.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
+    unrelated_skill.add_user_message("x")
+
+    for ctx in (with_edit, no_skill, unrelated_skill):
         assert (
             "Correction to the skill guidance"
             not in (ctx.get_messages_for_llm()[0]["content"])
--- a/tests/core/agent/test_react.py
+++ b/tests/core/agent/test_react.py
@@ -5716,3 +5716,52 @@ async def test_react_discards_the_preamble_with_the_rejected_response() -> None:
         "Let me look into that." not in str(event.get("content") or event.get("delta"))
         for event in outbound.events
     )
+
+
+@pytest.mark.asyncio
+async def test_run_flags_missing_image_editing_for_the_skill_correction() -> None:
+    # Covers the wiring: the helper's own test passes even if the call inside
+    # the tool-calling loop is deleted.
+    from xagent.core.agent.context.execution import (
+        IMAGE_EDIT_UNAVAILABLE_METADATA_KEY,
+    )
+
+    from .concurrency_harness import FakeTool as NamedFakeTool
+
+    llm = FakeLLM(responses=[{"content": "done", "done": True}])
+    pattern = ReActPattern(max_iterations=2)
+    context = ExecutionContext(system_prompt="You are helpful.")
+    context.add_user_message("make an ad")
+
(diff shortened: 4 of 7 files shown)


src/xagent/core/agent/context/execution.py
```diff
@@ -606,6 +607,19 @@ def _system_context(self) -> str:
...
+                parts.append(
+                    "Correction to the skill guidance above: image editing is "
+                    "unavailable here. No configured image model has the edit "
+                    "ability, so edit_image is not in your tools and passing "
+                    "images to generate_image fails. Ignore any instruction to "
+                    "edit an existing image or to attach a reference through "
+                    "images; render each deliverable from a text prompt in one "
+                    "call, and describe in the prompt what a reference would "
+                    "have contributed."
+                )
```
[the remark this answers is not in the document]
bluefish-08: Keeping it inline, for consistency with the rest of this method.

Every other block in `get_system_context` builds its text directly inside the `parts.append(...)` call — the memory-persistence paragraph, the skill-index instructions, the DAG-step note, the selected-skill guidance immediately above this one. Extracting one of them into a named constant while the others stay inline would make this block look special when it is the same kind of thing.

The suggested form also does not buy much: a local constant used exactly once, on the very next line, is the same number of lines with one more name to follow. A module-level constant would at least be reusable, but nothing else needs this string.

If these paragraphs are worth hoisting out, that is a change to all of them together rather than to the one this PR happens to touch.

tests/core/agent/test_react.py
```diff
@@ -5115,3 +5115,52 @@ async def test_final_answer_preserves_semantic_completion_outcome() -> None:
...
+    )
+
+    from .concurrency_harness import FakeTool as NamedFakeTool
+
+    llm = FakeLLM(responses=[{"content": "done", "done": True}])
+    pattern = ReActPattern(max_iterations=2)
+    context = ExecutionContext(system_prompt="You are helpful.")
+    context.add_user_message("make an ad")
+
+    await pattern.run(context=context, tools=[NamedFakeTool("generate_image")], llm=llm)
```
rogercloud: This run feeds a synthetic tool list that no default production path on this branch produces: #1294's gating (which stops registering `edit_image` on edit-less deployments) is not merged here — 239e17b2 is not an ancestor of HEAD — so `edit_image` is still registered unconditionally (src/xagent/core/tools/adapters/vibe/image_tool.py:80-92) and this flag never fires in a default deployment. Please update the branch from main so the #1294 gating exists, and consider shaping the list like a real generate-only deployment (e.g. generate_image + list_image_models) with a skill loaded.

src/xagent/core/agent/context/execution.py
```diff
@@ -606,6 +607,19 @@ def _system_context(self) -> str:
                 "Selected skill guidance. Use it when relevant to the current task:\n"
                 f"{str(skill_context).strip()}"
             )
+            # Skill text is injected verbatim and cannot know which tools were
+            # registered, so the correction has to come after it.
+            if self.metadata.get(IMAGE_EDIT_UNAVAILABLE_METADATA_KEY):
+                parts.append(
+                    "Correction to the skill guidance above: image editing is "
+                    "unavailable here. No configured image model has the edit "
```
rogercloud: The flag only measures tool-list membership, so "No configured image model has the edit ability" and "passing images to generate_image fails" over-claim: policy filters (src/xagent/core/tools/adapters/vibe/factory.py:577-627, src/xagent/core/agent/service.py:734-742) can drop `edit_image` while an edit-capable model remains configured, and the images -> edit_image delegation in image_tool.py still works in that case. Reword to the operative, tool-level fact — e.g. "edit_image is not in your tools" — and drop the model-ability claim.

src/xagent/core/agent/context/execution.py
```diff
@@ -606,6 +607,19 @@ def _system_context(self) -> str:
                 "Selected skill guidance. Use it when relevant to the current task:\n"
                 f"{str(skill_context).strip()}"
             )
+            # Skill text is injected verbatim and cannot know which tools were
+            # registered, so the correction has to come after it.
+            if self.metadata.get(IMAGE_EDIT_UNAVAILABLE_METADATA_KEY):
```
rogercloud: This gate fires for ANY loaded skill, not only skills that actually teach `edit_image` or `images`: 8 of the 9 builtin skills contain no such guidance yet would receive "Correction to the skill guidance above..." on generate-only deployments. The loaded skill names are already available here (`LOADED_SKILLS_METADATA_KEY` is read just above, execution.py:580-583) — consider gating on the skill subject, or softening the preamble so it reads as deployment context rather than a contradiction.

src/xagent/core/agent/context/execution.py
```diff
@@ -56,6 +56,7 @@
 # nothing a dropped observation held.
 NON_EVIDENCE_TOOL_NAMES = CONTROL_TOOL_NAMES | {LOAD_SKILL_TOOL_NAME}
 COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
+IMAGE_EDIT_UNAVAILABLE_METADATA_KEY = "image_edit_unavailable"
```
rogercloud: The key name reads as "edit_image absent", but the stored value is "generate_image present && edit_image absent" — it is False on deployments with no image tools at all, and no existing metadata key has "correction needed" semantics. Consider renaming (e.g. IMAGE_GENERATION_WITHOUT_EDIT) or at least a comment stating the exact predicate.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -996,6 +998,20 @@ def _schema_tool_names(self, tool_schemas: list[dict[str, Any]]) -> list[str]:
...
+    def _record_image_edit_availability(self, context: Any) -> None:
+        """Flag a deployment that can generate images but not edit them.
+
+        Injected skill text names edit_image unconditionally; the system context
+        needs this to contradict it.
+        """
+        metadata = getattr(context, "metadata", None)
+        if metadata is None:
+            return
+        names = self._tool_decision_groups_by_name
```
rogercloud: This helper reads `self._tool_decision_groups_by_name` implicitly; it is only correct because the call site (line 390) sits immediately after the map is rebuilt (line 389). Any future call elsewhere would silently read stale or empty state. Consider taking the tools (or the computed name set) as a parameter so the dependency is explicit.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -996,6 +998,20 @@ def _schema_tool_names(self, tool_schemas: list[dict[str, Any]]) -> list[str]:
                 names.append(str(function["name"]))
         return names
 
+    def _record_image_edit_availability(self, context: Any) -> None:
+        """Flag a deployment that can generate images but not edit them.
+
+        Injected skill text names edit_image unconditionally; the system context
+        needs this to contradict it.
+        """
+        metadata = getattr(context, "metadata", None)
```
rogercloud: This None guard never fires: `context.metadata` is an always-present dict (execution.py:105) and run() already dereferences `context.metadata` unguarded (react.py:295) before this helper is reachable. Write `context.metadata` directly — or, if duck-typed defensiveness is wanted, use the `isinstance` check the sibling helper uses (auto.py:759-764) for consistency. Either way ~3 lines drop.

src/xagent/core/agent/context/execution.py
```diff
@@ -606,6 +607,19 @@ def _system_context(self) -> str:
                 "Selected skill guidance. Use it when relevant to the current task:\n"
                 f"{str(skill_context).strip()}"
             )
+            # Skill text is injected verbatim and cannot know which tools were
+            # registered, so the correction has to come after it.
+            if self.metadata.get(IMAGE_EDIT_UNAVAILABLE_METADATA_KEY):
+                parts.append(
+                    "Correction to the skill guidance above: image editing is "
```
rogercloud, line 622: This paragraph near-verbatim duplicates the "IMAGE EDITING IS UNAVAILABLE" prefix #1294 adds to the `generate_image` description ("No configured image model has the edit ability", "render each deliverable from a text prompt in one call"). After merge both live in the same prompt with no cross-copy test. Cheapest fix: a cross-reference comment at each site so a wording change updates both.

tests/core/agent/test_react.py
```diff
@@ -5115,3 +5115,52 @@ async def test_final_answer_preserves_semantic_completion_outcome() -> None:
...
+
+    assert context.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] is True
+
+
+@pytest.mark.parametrize(
+    ("tool_names", "unavailable"),
+    [
+        (["generate_image", "list_image_models"], True),
```
rogercloud, line 5724: Add (['list_image_models'], False): a deployment where `list_image_models` stays registered but generate/edit are absent is the canonical no-image-tools case per #1294, and pins the exact scenario the `generate_image` precondition exists for. ([] -> False) is an optional defensive extra.

tests/core/agent/test_context.py
```diff
@@ -619,6 +619,48 @@ def test_get_messages_for_llm_coalesces_system_messages() -> None:
...
+        "## Available Skill: static-visual-design\n\nUse `edit_image` to refine."
+    )
+    ctx.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] = True
+    ctx.add_user_message("make an ad")
+
+    system_content = ctx.get_messages_for_llm()[0]["content"]
+
+    assert "image editing is unavailable here" in system_content
+    # The correction is worthless unless it lands after the text it contradicts.
+    assert system_content.index("Selected skill guidance") < system_content.index(
```
rogercloud: The assertion anchors on the "Selected skill guidance" header, but today header and body are one part, so the ordering check is transitive — a future split of the skill block into header + body would silently weaken it. Anchor on a substring of the fixture body itself (e.g. "Use `edit_image` to refine."); note short anchors like "edit_image" are ambiguous because the correction contains that string too.

tests/core/agent/test_react.py
```diff
@@ -5115,3 +5115,52 @@ async def test_final_answer_preserves_semantic_completion_outcome() -> None:
...
+    from .concurrency_harness import FakeTool as NamedFakeTool
+
+    llm = FakeLLM(responses=[{"content": "done", "done": True}])
+    pattern = ReActPattern(max_iterations=2)
+    context = ExecutionContext(system_prompt="You are helpful.")
+    context.add_user_message("make an ad")
+
+    await pattern.run(context=context, tools=[NamedFakeTool("generate_image")], llm=llm)
+
+    assert context.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] is True
```
rogercloud: This e2e asserts only the metadata flag, while the reader tests set the flag manually — nothing proves run + skill -> correction in the rendered system content. `FakeLLM.calls` already captures the rendered prompt (test_react.py:302-307); one assertion on `calls[0]['messages'][0]['content']` containing the correction would close the loop. Note the current run sets no skill context, so a skill would need to be loaded for the correction to appear.

rogercloud (changes requested):
## Summary

This PR gives runtime skill guidance a way to contradict itself when image editing is unavailable: `ReActPattern` records a `generate_image present && edit_image absent` flag into `context.metadata` at every tool-calling loop entry (reusing the `_tool_decision_groups_by_name` map already built for the circuit breaker), and `ExecutionContext._system_context` appends an authoritative correction paragraph AFTER the verbatim skill guidance whenever the flag is set and skill context is present, so the later statement wins. Four tests cover the writer (parametrized helper test), the reader (correction emitted with a skill loaded), the ordering (correction lands after the guidance it contradicts), and the wiring (end-to-end through `pattern.run()`). The motivation is issue #1307: `static-visual-design` names `edit_image` and the `images` argument unconditionally, and the skill layer has no capability seam — `SkillParser` stores the whole SKILL.md as one blob with nothing to condition on.

## Update since prior review

No commits have been pushed since the prior review (gemini-code-assist inline suggestion at 2026-08-13T05:03Z, author reply at 05:17Z); HEAD is unchanged at 1decdda2. This pass re-evaluated the same code from a blank slate — design model, mechanism, and tests — and found issues the prior review missed, including one that is blocking.

## Prior-findings checklist

- P1 (hoist the correction string into a constant; prior inline at execution.py:622) — **WAIVED**: the author's consistency argument checks out — every other block in `_system_context` builds its text inline inside `parts.append(...)`, the suggested constant would be single-use, and the site is unique repo-wide. Style suggestion, not a bug.

## Findings

### Major

**F1 — The mechanism's precondition (#1294 gating) is absent from this branch** (major; adjusted from critical during verification).

#1294's gating — which stops registering `edit_image` when no configured model declares the edit ability — is not an ancestor of HEAD 1decdda2 (239e17b2 is not in the branch history), so this tree still registers `generate_image`, `edit_image`, and `list_image_models` unconditionally (src/xagent/core/tools/adapters/vibe/image_tool.py:80-92). Consequences: (1) the flag this PR adds never fires in a default deployment, so the correction paragraph never appears; (2) the end-to-end test feeds a synthetic tool list (`tools=[NamedFakeTool("generate_image")]`) that no default production path on this branch produces; (3) the PR body's premise — "#1294 made this reachable" — is false for the reviewed branch. The flag is technically reachable through non-default name-based policy configs (legacy `allowed_tools` at src/xagent/core/tools/adapters/vibe/factory.py:577-591, `disabled_by_hook` at :601-612, user allowlist at :615-627, plus the filter at src/xagent/core/agent/service.py:734-742), but in those cases the correction would misfire (see F2) — and "IMAGE EDITING IS UNAVAILABLE" is likewise absent from this branch. After merge onto main the mechanism becomes live as intended, so the defect is branch state, not design. Fix: update the branch from main so the #1294 gating exists, and consider an end-to-end test shaped like a real generate-only deployment (generate_image + list_image_models, with a skill loaded).

### Minor

**F2 — Correction text over-claims causation** (execution.py:614-621)

The correction asserts "No configured image model has the edit ability" and "passing images to generate_image fails", but the flag measures only tool-list membership. Policy filters (factory.py:577-627, service.py:734-742, web tool-config toggles) can drop `edit_image` while an edit-capable model remains configured, and the `images` -> `edit_image` delegation in image_tool.py bypasses the registry and would still succeed there — so in that misfire combination the correction denies a working route. The operative fact "edit_image is not in your tools" stays true. Suggest tool-level wording, e.g. "edit_image is not in your tools, and passing images to generate_image routes into the edit path".

**F3 — DAG planner sits outside the correction path** (body-only; planner code is outside the diff)

DAG steps and single_call construct `ReActPattern`s and inherit the flag via the child-context metadata copy, but the planner itself never sees the correction: `plan_generator` builds its own prompt and injects `selected_skill_context` verbatim (src/xagent/core/agent/pattern/dag/plan_generator.py:518-521) without reading the flag, and runs before any step records it (dag.py:445-451). Three mitigations bound the impact: the planner prompt lists `available_tool_names` and the system instruction forbids step tools outside that list; `_filter_suggested_tools` silently drops a hallucinated `edit_image` tool name, so no step is ever handed the absent tool; and every step executor's system context does contain the correction — the component that actually calls tools is corrected. Residual harm: the plan's step decomposition and descriptions are authored by the uncorrected planner and can be built around an edit step, causing step deviation and re-plans — a workflow-quality issue, not a hard execution failure.

**F5 — Correction fires for any loaded skill** (execution.py:612)

The gate is skill-context presence + flag, with no subject check: 8 of the 9 builtin skills contain no `edit_image` or `images` guidance, yet on a generate-only deployment they would all get "Correction to the skill guidance above..." — an unearned contradiction. The loaded skill names are already available at the emission point (`LOADED_SKILLS_METADATA_KEY` is read at execution.py:580-583). Suggest gating on the skill name/subject, or softening the preamble so it reads as deployment context rather than a correction of that skill.

**F6 — Flag can be stale across request-scoped state** (flag write at react.py:1011-1013; body-only)

The flag is written unconditionally at each loop entry but survives to_dict/from_dict (execution.py:872/947), is copied into child contexts (:810) and merges (:726), and is not cleared by `_clear_request_scoped_enrichment` (auto.py:1795-1806). Concrete stale-read path: AutoPattern routing re-renders the system context at auto.py:817 before the child loop recomputes the flag, and the router can load a skill mid-routing — one routing prompt can carry a stale correction. It self-corrects at the next loop entry. Fix: pop the key in `_clear_request_scoped_enrichment` or recompute before routing.

**F7 — Key name misstates the flag semantics** (execution.py:59)

`image_edit_unavailable` reads as "!edit_image" but stores "generate_image present && edit_image absent", and is False on deployments with no image tools at all (pinned by the `web_search` -> False case). No existing metadata key encodes "correction needed" semantics. Optional: rename (e.g. `IMAGE_GENERATION_WITHOUT_EDIT`) or add a comment stating the exact predicate.

**F8 — Helper reads implicit instance state** (react.py:1010)

`_record_image_edit_availability` reads `self._tool_decision_groups_by_name` implicitly; it is correct only because the call site (react.py:390) sits immediately after the map is rebuilt (react.py:389). A future call elsewhere would silently read stale or empty state. Optional: take the tools (or the computed name set) as a parameter.

**F9 — Guard asymmetry with the sibling helper** (react.py:1007-1009)

The helper guards `getattr`/None only, while the sibling at auto.py:759-764 uses `isinstance`. `context.metadata` is an always-present dict (execution.py:105), so the guard is dead code in production — though consistent with the duck-typed API. Style nit; see also S2 below.

**F11 — Correction duplicates #1294's prefix with no cross-copy test** (execution.py:614-621)

The correction shares near-verbatim text with the "IMAGE EDITING IS UNAVAILABLE" prefix #1294 puts on the `generate_image` description ("No configured image model has the edit ability"; "render each deliverable from a text prompt in one call"). Post-merge both will live in the same prompt, and nothing tests they stay in sync. Cheapest fix: a cross-reference comment at each site.

**F14 — No DAG-path coverage** (body-only)

The PR body claims "DAG steps ... inherit it", but no test exercises a DAG run: the DAG composition (child-context metadata copy, per-step tool filtering at dag.py:899) is untested, and the single_call/AutoPattern sub-claims are near-redundant with the shared loop entry already covered. A DAG-path test is feasible in ~20 lines following the test_dag.py:1024-1065 pattern.

**F15 — Parametrized test misses the canonical no-image-tools deployment** (test_react.py:5142-5144)

Add `(['list_image_models'], False)`: a deployment where `list_image_models` stays registered but generate/edit are absent is exactly the scenario #1294 keeps reachable, and the one the `generate_image` precondition exists for. The existing `web_search` -> False case covers the branch, not the scenario. `([] -> False)` is an optional defensive extra.

**F16 — Ordering assertion anchors on a header, not the guidance body** (test_context.py:638-639)

The assertion compares the position of "Selected skill guidance" with the correction header. Today header and body are one part, so the check is transitive; a future split of the skill block into header + body would silently weaken it. Anchor on a substring of the fixture body itself (e.g. "Use `edit_image` to refine.") — note short anchors like "edit_image" are ambiguous because the correction contains that string too.

**F17 — No composed end-to-end assertion of the rendered correction** (test_react.py:5135-5137)

The e2e test asserts only the metadata flag; the reader tests set the flag manually. Nothing proves run + skill -> correction in the rendered system content. `FakeLLM.calls` already captures the rendered prompt (test_react.py:302-307) — one assertion on `calls[0]['messages'][0]['content']` would close the loop, but it needs a run that actually loads a skill (the current run sets no skill context, so no correction would appear).

### Nit

**F10 — Constant defined in an unexpected module** (execution.py:59)

`IMAGE_EDIT_UNAVAILABLE_METADATA_KEY` is the first metadata key defined in execution.py; its siblings live in enrichment.py / skill_tool.py. execution.py is the only non-cyclic home (react.py already imports from context.execution), so this is fine — a one-line comment noting why it lives here would help future readers.

### Informational — no changes requested

- **F4 (AutoPattern routing retry):** routing re-renders the system context (auto.py:817) before the flag is written and `load_skill` can happen mid-routing, so one routing prompt could carry a stale/missing correction. The window is tiny (routing model has only `[decision, load_skill]` tools, attempts <= 2, one skill load per cycle), and this is not a regression — pre-PR there was no correction anywhere. Feature incompleteness only.
- **F12 (no-image-tools deployments):** when `generate_image` is absent the correction is not emitted, so skill text still names `edit_image`. Under the relaxed outcome-based criterion this is acceptable: the per-turn tool schema is the authoritative contradiction, there is no doomed-call route without the `images` parameter, and a correction there would be false noise. The PR body documents this as intentional.
- **F13 (tool_choice='none'):** the flag is computed independently of `tool_choice` (react.py:390 vs :391-394), but `tool_choice='none'` is unreachable in production — no construction site sets it, and react.py:812 documents this. Future-proofing only.

- L1001: shrink — `_record_image_edit_availability` (react.py:1001-1013) is a one-caller test seam (test_react.py:5164 calls it directly; the wiring test's own comment admits it); inline the 6-line write next to the map at react.py:389 and fold the 3-case table into run-based wiring-test params — same coverage, real ~25-line reduction (the claimed ~40 is generous). Keep the docstring rationale as a one-line comment at the call site.
- L1007: delete — the `getattr`/None guard (react.py:1007-1009) never fires: `metadata` is always a dict (execution.py:105) and run() dereferences `context.metadata` unguarded (react.py:295) before the helper is reachable; write `context.metadata` directly.
- net: ~28 lines possible (S1 ~25 + S2 3)

Blocking: yes.

- src/xagent/core/agent/pattern/react/react.py:1011-1013 — major — mechanism precondition (#1294 gating) absent from branch; correction never fires in default deployments and the e2e test models no default production input — [new]

Recommended event: REQUEST_CHANGES. All other findings are minor or below; none are blocking.

bluefish-08:
Thanks for the review — all of it is addressed. Branch rebased onto `fb5ed4a5`, plus one follow-up commit.

**F1 (blocking).** Correct, and the diagnosis was exact: the branch was based on `67ea1928`, 12 commits behind main, so `239e17b2` was not an ancestor and the flag could never fire in a default deployment. Rebased — the #1294 gating is now in the tree. The end-to-end test was reshaped into a real generate-only deployment (`generate_image` + `list_image_models`) instead of a lone synthetic tool.

**F2.** Reworded to the tool-level fact, dropping the model-capability claim: "edit_image is not in your tools, and passing images to generate_image routes into the same absent edit path." The misfire you describe (policy filter drops the tool while an edit-capable model remains) no longer produces a false statement.

**F5.** Added a subject check — the correction now requires `edit_image` to appear in the injected skill text. Chose text matching over a skill-name gate because it needs no maintenance as skills are added, and it keys on the exact thing being contradicted. The known ceiling: a skill that teaches the `images` argument without ever naming `edit_image` gets no correction. None of the 9 builtins is shaped that way, and #1294's `IMAGE EDITING IS UNAVAILABLE` description prefix still backstops it.

**F6.** The key is now popped in `_clear_request_scoped_enrichment`, with a small test pinning it. This closes F4 as well — the routing prompt can no longer carry a flag from a previous request.

**F7 / F10.** Kept the name, added one comment at the constant stating the exact predicate (`generate_image` present && `edit_image` absent, False when no image tools exist) and why it lives in `execution.py`.

**F8 / F9 / S1 / S2.** You were right that the helper was a test seam. Deleted it; the 4-line write is inlined next to the map it reads, and the dead `getattr`/None guard is gone with it — 14 lines of source removed.

**F15 / F16 / F17.** The three tests collapsed into one run-based parametrization: it covers the wiring, adds `(['list_image_models'], False)`, and asserts the *rendered* system content rather than only the metadata flag. The ordering assertion now anchors on the guidance body instead of its header.

**F11.** Resolved by F2 rather than by a cross-reference: #1294's prefix speaks about model capability, this correction now speaks only about tool absence. They are no longer near-duplicates, so there is nothing to keep in sync.

**F3 / F14 — not changed, and I'd like to leave it.** Your own analysis lists the three things that bound it: the planner prompt carries `available_tool_names`, `_filter_suggested_tools` drops a hallucinated `edit_image`, and every step executor's system context does contain the correction. What remains is plan-description quality, not an execution failure. Teaching `plan_generator` to read the flag means threading it into a separately-constructed prompt and a DAG-path test for a workflow-polish gain — happy to do it if you disagree, but it looks like a poor trade to me.

F12 / F13 noted, no change intended.

`tests/core/agent` is green (753 passed). Mutation-checked the three new guards: deleting the subject check, the `pop`, or the flag write each turns the suite red.

src/xagent/core/agent/context/execution.py
```diff
@@ -606,6 +610,21 @@ def _system_context(self) -> str:
                 "Selected skill guidance. Use it when relevant to the current task:\n"
                 f"{str(skill_context).strip()}"
             )
+            # Skill text is injected verbatim and cannot know which tools were
+            # registered, so the correction has to come after it.
+            if self.metadata.get(
+                IMAGE_EDIT_UNAVAILABLE_METADATA_KEY
+            ) and "edit_image" in str(skill_context):
```
rogercloud: Minor: this gate is a case-sensitive exact-substring match on `edit_image`. Skills loaded from user/external roots (`create_skill_manager` roots incl. `XAGENT_EXTERNAL_SKILLS_LIBRARY_DIRS`, `src/xagent/skills/utils.py:23-35`) can teach image editing without that literal token, so a false negative is reachable today; no test pins a real builtin SKILL.md body either, so the builtin wording is unguarded against drift. The #1294 description-prefix backstop mitigates runtime impact. Suggest matching case-insensitively (or against a small set of image-tool tokens) and pinning one real builtin skill's text in a test.

src/xagent/core/agent/context/execution.py
```diff
@@ -56,6 +56,10 @@
 # nothing a dropped observation held.
 NON_EVIDENCE_TOOL_NAMES = CONTROL_TOOL_NAMES | {LOAD_SKILL_TOOL_NAME}
 COMPACT_DROPPED_TOOL_NOTICE_MAX_NAMES = 20
+# True only when generate_image is registered and edit_image is not; a deployment
+# with no image tools at all leaves it False. Lives here, not in enrichment.py
+# with the other metadata keys, because react.py already imports from this module.
```
rogercloud: Minor, comment-only: the 'because react.py already imports from this module' rationale is factually wrong — react.py imports from enrichment too (`react.py:51-52`) and enrichment's transitive closure never imports execution.py (execution→enrichment is one-way, `execution.py:38`), so placing this key in enrichment.py would create no cycle. Either fix the clause or move the constant next to the other metadata keys.

src/xagent/core/agent/context/execution.py
```diff
@@ -606,6 +610,21 @@ def _system_context(self) -> str:
...
+            if self.metadata.get(
+                IMAGE_EDIT_UNAVAILABLE_METADATA_KEY
+            ) and "edit_image" in str(skill_context):
+                parts.append(
+                    "Correction to the skill guidance above: image editing is "
+                    "unavailable here. edit_image is not in your tools, and "
+                    "passing images to generate_image routes into the same "
+                    "absent edit path. Ignore any instruction to edit an "
+                    "existing image or to attach a reference through images; "
+                    "render each deliverable from a text prompt in one call, "
```
rogercloud, line 627: Minor: this clause is verbatim-identical to the #1294 capability prefix in `vibe/image_tool.py:87-88` ('Render each deliverable from a text prompt in one call') with no cross-reference. The rewrite removed the model-capability duplication, but this sentence can still drift if either side is edited. Suggest a one-line cross-reference comment on one of the two sites.

tests/core/agent/test_react.py
```diff
@@ -5115,3 +5115,42 @@ async def test_final_answer_preserves_semantic_completion_outcome() -> None:
...
+
+    await pattern.run(
+        context=context,
+        tools=[NamedFakeTool(name) for name in tool_names],
+        llm=llm,
+    )
+
+    assert context.metadata[IMAGE_EDIT_UNAVAILABLE_METADATA_KEY] is unavailable
+    rendered = llm.calls[0]["messages"][0]["content"]
+    assert ("image editing is unavailable here" in rendered) is unavailable
```
rogercloud, line 5755: Minor: this assertion checks only the `image editing is unavailable here` fragment. The images-denial clause (`attach a reference through images`, `execution.py:623`) is asserted nowhere in the suite, so a regression trimming that clause while keeping the `edit_image` denial would pass. Suggest asserting one distinctive fragment of the images-denial sentence.

rogercloud (approved):
### Follow-up review

**Summary.** This PR closes #1307 with the second of the two suggested directions: when a deployment registers `generate_image` but not `edit_image`, the runtime corrects the skill layer's unconditional `edit_image` guidance at render time — ReActPattern sets a per-turn `image_edit_unavailable` metadata flag from the existing `_tool_decision_groups_by_name` map (`react.py:392-395`), `ExecutionContext._system_context` appends a correction paragraph after the verbatim skill text when the flag is set and the skill names `edit_image` (`execution.py:615-627`), and AutoPattern clears the key per request (`auto.py:1804-1808`). The design matches the #1294 description-prefix precedent and adds a skill-subject gate plus an images-argument denial, keeping `static-visual-design` and the other builtin skills safe across capability configurations.

**What changed since the last review.** The branch was force-pushed and rebased onto `fb5ed4a5`, which brings #1294's capability gating into the tree and resolves the prior F1 blocker; the rewrite drops the helper, scopes the correction to skills that name `edit_image`, rewords it to tool-level claims, and clears the flag with request-scoped enrichment. The tests collapsed into one run-based parametrization with a rendered-content assertion and a guidance-body ordering anchor. Net change vs the previous head: 6 files, +67/-56; CI is green (14 SUCCESS).

**Prior-findings checklist**

| Finding | Status |
|---|---|
| P1 gemini constant extraction | WAIVED — waiver still holds; all `_system_context` blocks remain inline |
| F1 (major) correction contradicts #1294 gating | FIXED — rebase brings the gates in; `can_generate`/`can_edit` verified at `vibe/image_tool.py:74-107`; e2e model is a real generate-only deployment |
| F2 over-claiming correction text | FIXED — reworded to tool-level claims; delegation claim accurate (`core/image_tool.py:620-631`) |
| F3 DAG planner outside correction path | WAIVED — author declined (10:04Z reply); all three mitigations verified (`available_tool_names` in planner prompt; `_filter_suggested_tools` drops hallucinated `edit_image`; step executors render the correction). Residual: plan-text quality only |
| F4 stale flag | PARTIAL — stale-across-requests half FIXED (pop in `_clear_request_scoped_enrichment` + pinning test); first-request routing render still uncorrected — informational, no regression |
| F5 correction fires for any skill | FIXED — subject gate + unrelated-skill test case |
| F6 flag cleanup | FIXED — pop added + test |
| F7/F10 key name/placement comments | FIXED — 3-line comment (predicate + placement rationale) |
| F8/F9/S1/S2 helper | FIXED — inlined, dead guard deleted, tests no longer call it |
| F11 duplication with #1294 prefix | PARTIAL — reword removed the model-capability duplication; `render each deliverable from a text prompt in one call` remains verbatim in `execution.py:624` and `image_tool.py:87-88` with no cross-reference — minor sync hazard (inline note) |
| F14 no DAG-path test | WAIVED — flag recomputed at each step's own `run()` entry (`react.py:392-395`, `dag.py:1010-1011`); child-context metadata copy is pre-existing covered behavior; author's cost-benefit sound |
| F15/F16/F17 test coverage | FIXED — single run-based parametrization incl. `(["list_image_models"], False)`, rendered-content assertion, ordering anchor on the guidance body |
| F12/F13 informational | Noted, unchanged |

**New findings** (all minor)

- **N2 — case-sensitive trigger (`execution.py:617`).** The correction fires only when `"edit_image" in str(skill_context)` matches exactly, case-sensitively. Skills loaded from user/external roots (`create_skill_manager` roots incl. `XAGENT_EXTERNAL_SKILLS_LIBRARY_DIRS`, `skills/utils.py:23-35`) can teach image editing without that literal token, so a false negative is reachable today; no test pins real builtin SKILL.md text either, so the builtin invariant is unguarded against drift. Runtime impact is mitigated by the #1294 description-prefix backstop. Suggestion: match case-insensitively (or against a small set of image-tool tokens) and pin one real builtin skill's text in a test.
- **N3 — placement comment factually wrong (`execution.py:61`).** The comment claims the key lives here "because react.py already imports from this module," implying moving it to `enrichment.py` would add an import edge. That is wrong: react.py already imports from enrichment (`react.py:51-52`), and enrichment's transitive closure never touches execution.py (execution→enrichment is one-way, `execution.py:38`), so moving the key creates no cycle. Comment-only; fix the `because` clause or move the key next to the other metadata keys.
- **N4 — images-denial clause unasserted (`test_react.py:5156`).** The rendered-content assertion covers only `image editing is unavailable here`; the images-denial sentence (`attach a reference through images`, `execution.py:623`) is asserted nowhere, so a regression trimming that clause while keeping the `edit_image` denial would pass the entire suite. Suggestion: assert one distinctive fragment of the images-denial sentence.

**Non-blocking residual (folded under F3/F4).** The first-request routing render (`auto.py:818`) and the DAG plan payload (`plan_generator.py:526-529`) still carry uncorrected skill text, and the routing decision model can emit a user-facing final answer directly, so an image-editing instruction could reach the user uncorrected. No component that can call image tools is uncorrected, and pre-PR there was no correction anywhere, so this is not a regression — same informational residual as the previous review.

**Simplification.** Both lenses (whole-PR and update-only) returned **Lean already** — no over-engineering findings.

**Blocking status & recommended decision.** No confirmed critical or major findings. The remaining items — N2/N3/N4 (minor) and the F3/F4/F11 informational residuals — do not block. Recommended event: **APPROVE**.

bluefish-08:
Thanks — N2/N3/N4 and the F11 note are all addressed in `6189a752`.

**N3.** You're right and my comment was worse than wrong — it asserted an import-cycle constraint that does not exist. Since there is no cycle, I took the second option: the key now lives in `enrichment.py` next to the other metadata keys, and `react.py` / `auto.py` fold it into their existing `enrichment` import instead of reaching into `execution.py`. The placement rationale is gone because the placement is no longer surprising; only the predicate line remains.

**N2.** The gate now lowercases the skill text. I did not add the builtin-text pin: keying a test on a builtin SKILL.md body couples the suite to prose that is expected to change, and the failure it would catch (a skill teaching image editing without the token) is exactly the false negative case-insensitivity does not close either — an external skill can say "refine the image" with no token at all. The #1294 description prefix remains the backstop for that class. Pinned the case-insensitivity itself instead.

**N4.** Added the images-denial assertion alongside the existing one, so trimming that clause now fails the suite.

**F11.** One cross-reference comment at the `execution.py` site pointing at `vibe/image_tool.py`.

Also trimmed the comments the previous round left behind — several were restating what the code and test names already say.

`tests/core/agent` green (754 passed).

PR #1318, opened 2026-08-13
fix(tools): let media creator failures reach the factory handler

### LINKED ISSUE #1317: [Bug] Media tool creators swallow every exception, including the two the factory requires to propagate
Sub-issue of #1289, raised in the #1294 review and split out so it can be tracked
and closed on its own.

## Problem

Every media tool creator wraps its construction in a bare `except Exception` that
duplicates — and partly defeats — the one the factory already has.

`ToolFactory.create_registered_tools` (`adapters/vibe/factory.py:237-245`):

```python
try:
    created_tools = await creator(config)
    tools.extend(created_tools)
except ConnectorRuntimeError:
    raise
except RequiredMCPUnavailableError:
    raise
except Exception as e:
    logger.warning(f"Tool creator {creator.__name__} failed: {e}")
```

Each creator then repeats it internally, e.g.
`adapters/vibe/image_tool.py:187-199`:

```python
try:
    default_generate_model = config.get_image_generate_model()
    default_edit_model = config.get_image_edit_model()
    return create_image_tool(...)
except Exception as e:
    logger.warning(f"Failed to create image tools: {e}")
    return []
```

Same shape in `video_tool.py:120`, `audio_tool.py:262`, `music_tool.py:78`, and
`sound_effect_tool.py:78`.

## Why it matters

**1. It breaks the factory's propagation contract.** The factory names two
exception types that must escape. The inner handler catches `Exception`
unconditionally, so if either ever originates inside a creator it is silently
downgraded to "this creator produced nothing". Not reachable today — the media
config getters (`get_image_generate_model` and friends) load from the database
and do not touch the connector runtime — but the guard is exactly what would make
a future connector-backed model lookup fail silently instead of surfacing a 503.

**2. It hides real defects behind a config-shaped message.** The `try` covers the
whole construction, not just the config reads. For image tools that now includes
`get_tools()` and therefore every `model.has_ability(...)` call. A model class
that raises there takes down *all* image tools — including the read-only
`list_image_models`, which exists precisely to explain an empty toolset — and
leaves one `warning` line that reads like a configuration problem.

**3. Same failure, two outcomes.** `create_image_tool()` is also called directly,
with no exception handling, so the identical bad model class either vanishes
silently or raises loudly depending on which entry point was used.

## Suggested fix

Delete the inner handlers. The factory already logs, already names the creator in
its message (which the inner version does not), and already gets the propagation
rules right. This is a net deletion across five files, and it makes the direct
`create_image_tool()` path and the config path agree.

If a creator has a genuinely expected failure worth swallowing, catch that
specific exception around the specific call rather than wrapping construction.

## Acceptance

- [ ] A creator that raises during construction propagates to the factory handler
      rather than being caught inside the creator.
- [ ] `ConnectorRuntimeError` and `RequiredMCPUnavailableError` raised anywhere
      inside a media creator reach the caller.
- [ ] The direct `create_image_tool()` path and the config-driven path behave the
      same for the same failure.
- [ ] Covered by a test that fails if an inner blanket handler is reintroduced.

bluefish-08:
Closes #1317. Raised in the #1294 review.

## Problem

Five media tool creators each wrap their construction in a bare
`except Exception` that duplicates the one `ToolFactory` already applies to every
creator — and partly defeats it.

`adapters/vibe/factory.py:237-245`:

```python
try:
    created_tools = await creator(config)
    tools.extend(created_tools)
except ConnectorRuntimeError:
    raise
except RequiredMCPUnavailableError:
    raise
except Exception as e:
    logger.warning(f"Tool creator {creator.__name__} failed: {e}")
```

The inner handlers (`image_tool.py`, `video_tool.py`, `audio_tool.py`,
`music_tool.py`, `sound_effect_tool.py`) catch `Exception` unconditionally and
return `[]`.

Three consequences:

1. **The factory's propagation contract is broken.** It names two exception types
   that must escape; the inner handler swallows them first. Not reachable today —
   the media config getters load from the database and never touch the connector
   runtime — but it is precisely the guard that would make a future
   connector-backed model lookup fail silently instead of surfacing its 503.
2. **Real defects hide behind a config-shaped message.** The `try` spans the whole
   construction, which for image tools now includes `get_tools()` and therefore
   every `model.has_ability(...)` call. A model class raising there takes down
   *all* image tools — including the read-only `list_image_models`, which exists
   to explain an empty toolset — leaving one `warning` that reads like
   misconfiguration.
3. **Same failure, two outcomes.** `create_image_tool()` is also called directly
   with no handler, so the identical bad model class either vanishes silently or
   raises loudly depending on the entry point.

## Change

Delete the five inner handlers. Net −20 lines of source.

The factory already logs, and its message names the creator (`Tool creator
create_image_tools_from_config failed: ...`) where the inner one did not. The
direct `create_image_tool()` path and the config-driven path now agree.

Behaviour for ordinary failures is unchanged: still a warning, still that
creator contributing nothing. What changes is that the two contract-critical
exception types now reach the caller, and that the log line identifies which
creator failed.

## Tests

One test asserting a model whose `has_ability` raises propagates out of
`create_image_tools_from_config` instead of being converted into an empty tool
list. Mutation-tested: reintroducing the inner handler turns it red — which is
the acceptance criterion in #1317.

`tests/core/tools` and `tests/web/tools` pass. (`test_aws_mcp.py` fails to
collect locally for a missing `boto3`, unrelated.)

### DIFF
--- a/tests/core/tools/test_image_tool.py
+++ b/tests/core/tools/test_image_tool.py
@@ -1000,3 +1000,26 @@ class TestImageToolCapabilityGating:
 
         assert result["success"] is False
         assert "model1 (abilities: edit)" in result["error"]
+
+
+@pytest.mark.asyncio
+async def test_creator_propagates_a_broken_model_instead_of_hiding_it(tmp_path) -> None:
+    # A blanket handler here would turn a broken model class into "this
+    # deployment has no image tools", indistinguishable from having none.
+    from xagent.core.tools.adapters.vibe.config import ToolConfig
+    from xagent.core.tools.adapters.vibe.image_tool import (
+        create_image_tools_from_config,
+    )
+
+    broken = Mock(spec=BaseImageModel)
+    broken.has_ability = Mock(side_effect=RuntimeError("ability probe exploded"))
+
+    class _Config(ToolConfig):
+        def get_image_models(self):
+            return {"broken": broken}
+
+        def get_workspace_config(self):
+            return {"task_id": "creator-propagation", "base_dir": str(tmp_path)}
+
+    with pytest.raises(RuntimeError, match="ability probe exploded"):
+        await create_image_tools_from_config(_Config({}))
--- a/src/xagent/core/tools/adapters/vibe/audio_tool.py
+++ b/src/xagent/core/tools/adapters/vibe/audio_tool.py
@@ -251,14 +251,10 @@ async def create_audio_tools_from_config(config: "BaseToolConfig") -> List[Any]:
     if not workspace:
         return []
 
-    try:
-        return create_audio_tool(
-            asr_models=asr_models,
-            tts_models=tts_models,
-            workspace=workspace,
-            default_asr_model=config.get_asr_model(),
-            default_tts_model=config.get_tts_model(),
-        )
-    except Exception as e:
-        logger.warning(f"Failed to create audio tools: {e}")
-        return []
+    return create_audio_tool(
+        asr_models=asr_models,
+        tts_models=tts_models,
+        workspace=workspace,
+        default_asr_model=config.get_asr_model(),
+        default_tts_model=config.get_tts_model(),
+    )
--- a/src/xagent/core/tools/adapters/vibe/audio_tool.py
+++ b/src/xagent/core/tools/adapters/vibe/audio_tool.py
@@ -8,7 +8,6 @@ This module provides audio processing capabilities including:
 Uses pre-configured ASR and TTS models passed from the web layer.
 """
 
-import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from ....model.asr.base import BaseASR
@@ -17,8 +16,6 @@ from ....workspace import TaskWorkspace
 from ...core.audio_tool import AudioToolCore
 from .base import ToolCategory
 from .function import FunctionTool
-
-logger = logging.getLogger(__name__)
 
 
 class AudioFunctionTool(FunctionTool):
--- a/src/xagent/core/tools/adapters/vibe/factory.py
+++ b/src/xagent/core/tools/adapters/vibe/factory.py
@@ -114,7 +114,9 @@ class ToolRegistry:
         Register a tool creator function.
 
         The creator function will be called during create_all_tools()
-        with the current config.
+        with the current config. Creators must not blanket-catch their own
+        failures: ``create_registered_tools`` is the single enforcement point
+        for the exception contract described in its docstring.
 
         Usage (bare decorator, no category metadata):
             @register_tool
@@ -218,6 +220,11 @@ class ToolRegistry:
         (dynamic ones: MCP / Custom API / Image / Audio) are always
         dispatched and are responsible for
         short-circuiting internally based on the spec.
+
+        Exception contract for creators: ``ConnectorRuntimeError`` and
+        ``RequiredMCPUnavailableError`` propagate to the caller; any other
+        exception is logged as a warning and that creator contributes no
+        tools. Creators rely on this and must not catch broadly themselves.
         """
         # Import tool modules on first call to trigger decorator registration
         cls._import_tool_modules()
@@ -242,7 +249,9 @@ class ToolRegistry:
             except RequiredMCPUnavailableError:
                 raise
             except Exception as e:
-                logger.warning(f"Tool creator {creator.__name__} failed: {e}")
+                logger.warning(
+                    f"Tool creator {creator.__name__} failed: {e}", exc_info=True
+                )
 
         # Sort tools by category priority
         tools = cls._sort_tools_by_category(tools)
--- a/src/xagent/core/tools/adapters/vibe/image_tool.py
+++ b/src/xagent/core/tools/adapters/vibe/image_tool.py
@@ -5,7 +5,6 @@ This module provides image generation capabilities using pre-configured image mo
 passed from the web layer.
 """
 
-import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from ....model.image.base import BaseImageModel
@@ -13,8 +12,6 @@ from ....workspace import TaskWorkspace
 from ...core.image_tool import ImageGenerationToolCore
 from .base import ToolCategory
 from .function import FunctionTool
-
-logger = logging.getLogger(__name__)
 
 
 class ImageGenerationFunctionTool(FunctionTool):
@@ -69,8 +66,10 @@ class ImageGenerationTool(ImageGenerationToolCore):
         """Get all tool instances."""
         # A note inside the description is not enough: models blind-retry a
         # registered-but-doomed tool, so an unusable one has to leave the schema.
-        # list_image_models stays either way — it is read-only, cannot fail, and
-        # is the only way to answer "why is there nothing here".
+        # list_image_models stays whenever the ability probes return — it is the
+        # only way to answer "why is there nothing here". A probe that *raises*
+        # takes the whole toolset with it, by design: that is a broken model
+        # class, not a deployment without image models.
         can_generate = self._get_model() is not None
         can_edit = self._get_edit_model() is not None
 
--- a/src/xagent/core/tools/adapters/vibe/image_tool.py
+++ b/src/xagent/core/tools/adapters/vibe/image_tool.py
@@ -184,16 +184,12 @@ async def create_image_tools_from_config(config: "BaseToolConfig") -> List[Any]:
     if not workspace:
         return []
 
-    try:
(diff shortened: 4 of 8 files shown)


tests/core/tools/test_image_tool.py
```diff
@@ -1000,3 +1000,26 @@ async def test_generate_error_lists_abilities(
 
         assert result["success"] is False
         assert "model1 (abilities: edit)" in result["error"]
+
+
+@pytest.mark.asyncio
+async def test_creator_propagates_a_broken_model_instead_of_hiding_it(tmp_path) -> None:
```
rogercloud, line 1006: The mutation guard covers only the image creator: reintroducing the removed handler in video/audio/music/sound_effect alone would keep the suite green (video/audio creators have zero tests; music/sound-effect tests cover happy/empty paths only), while this test is verified mutation-effective for image. Consider parametrizing this test across all five creators so the factory contract is guarded uniformly.

rogercloud (approved):
## Summary

This PR removes the five redundant `except Exception` handlers inside the vibe media tool creators (image, video, audio, music, sound effect) so that `ToolFactory.create_registered_tools` becomes the single enforcement point for the creator exception contract: `ConnectorRuntimeError`/`RequiredMCPUnavailableError` propagate, everything else warns and continues. The change is net −20 lines, ordinary-failure behavior is unchanged (warn, creator contributes nothing), and the diagnostic log now names the failing creator. CI is green across all checks (pre-commit, pytest fast web/integration, deepdoc, migrations, e2e, frontend-build), and the prior review (gemini-code-assist) found no issues. Blocking: no — recommended event: APPROVE

## Approach

The design is sound. All five creators are `@register_tool` entries and `create_registered_tools` (src/xagent/core/tools/adapters/vibe/factory.py:211-249) is their only production caller; its per-creator `except` block already implements exactly the contract the inner handlers duplicated (factory.py:242-245, matching the established contract triples at factory.py:1252-1255 and 1329-1332). Config getters swallow their own errors, so the typed exceptions are unreachable in practice today — matching the PR's claim — and the one genuinely raise-capable path (image `get_tools()` calling `model.has_ability`) now surfaces through the factory to the existing task-lifecycle handlers, as the new test proves end-to-end.

## Findings

**F1 — minor (code quality): dead logging scaffolding.** Five files keep `import logging` + `logger = logging.getLogger(__name__)` with zero remaining logger calls after the handlers were deleted: src/xagent/core/tools/adapters/vibe/image_tool.py:8,17; video_tool.py:8,18; audio_tool.py:11,21; music_tool.py:5,14; sound_effect_tool.py:5,14. Verified: only two refs per file, no `logger.<method>`, no cross-module references. Suggestion: remove the import and assignment from all five files (see Simplification opportunities).

**F2 — minor (diagnostics): the sole remaining diagnostic log has no traceback.** factory.py:244-245 logs `Tool creator {name} failed: {e}` without `exc_info`, so a `has_ability` explosion deep inside model code leaves a one-line warning with no stack. Honest nuance: this line is pre-existing (factory.py is byte-identical base vs head) and the removed inner handlers were equally traceback-less, so this is not a regression — but the PR concentrates all media-creator diagnostic weight on this single reporting point. Suggestion: `logger.warning(..., exc_info=True)` or `logger.exception`.

**F3 — minor (design, follow-up recommendation): the same anti-pattern remains in sibling creators.** Blanket `except Exception: warn; return []` still exists in registered creators dispatched by the same factory loop: src/xagent/core/tools/adapters/vibe/agent_tool.py:2697-2700, browser_tools.py:55-57, knowledge_tools.py:63-65, file_ingestion_tool.py:312-314 (custom_api_factory.py:59-60 already re-raises correctly). All are pre-existing and untouched by this PR. Recommendation: file a follow-up issue sweeping the remaining creators — out of scope for #1317.

**F4 — minor (test quality): the mutation guard covers only the image creator.** tests/core/tools/test_image_tool.py:1005-1026 guards `create_image_tools_from_config` only; reintroducing the removed handler in video/audio/music/sound_effect alone would keep the suite green (video/audio creators have zero tests; music/sound-effect tests cover happy/empty paths only). Verified mutation-effective for the image path. Suggestion: parametrize the test across the five creators (see inline comment).

**F6 — no severity (design note): consequence #2 is only half-addressed by design.** image_tool.py:74-75 evaluates `can_generate = self._get_model() is not None` before `list_image_models` is appended (109-115), so a raising `has_ability` aborts `get_tools()` first and the factory's warn-and-continue drops the entire toolset, including the read-only `list_image_models`. This is pre-existing behavior (`get_tools()` is byte-identical at base) and the PR explicitly acknowledges it is unchanged; the informational point is that the comment at image_tool.py:70-73 ("list_image_models stays either way — it is read-only, cannot fail") overstates the guarantee — it stays only when `has_ability` returns normally, and the new test proves the raising scenario lands in factory warn+continue. The fix would be small (probe per model, or append `list_image_models` first); informational only.

**F7 — minor (documentation): the exception contract is undocumented.** Neither `ToolRegistry.register` (factory.py:116-138) nor `create_registered_tools` (factory.py:212-221) docstrings state that creators must not blanket-catch and that the loop warns+swallows generic exceptions while re-raising the two typed errors — plausibly why five creators independently built inner handlers. No changed docstring is inaccurate. Suggestion: one sentence in each docstring.

- src/xagent/core/tools/adapters/vibe/image_tool.py:8,17: delete: `import logging` / `logger = logging.getLogger(__name__)` (dead since handler removal)
- src/xagent/core/tools/adapters/vibe/video_tool.py:8,18: delete: `import logging` / `logger = logging.getLogger(__name__)`
- src/xagent/core/tools/adapters/vibe/audio_tool.py:11,21: delete: `import logging` / `logger = logging.getLogger(__name__)`
- src/xagent/core/tools/adapters/vibe/music_tool.py:5,14: delete: `import logging` / `logger = logging.getLogger(__name__)`
- src/xagent/core/tools/adapters/vibe/sound_effect_tool.py:5,14: delete: `import logging` / `logger = logging.getLogger(__name__)`

net: ~10 lines possible

Non-blocking: minor and simplification findings only — APPROVE.

bluefish-08:
Thanks — addressed everything except F3, which is now a follow-up issue.

**F1 (dead logging scaffolding).** Removed `import logging` + `logger = logging.getLogger(__name__)` from all five files. Confirmed dead: no `logger.` call remains and nothing imports these module-level loggers.

**F2 (no traceback).** `factory.py` now logs `..., exc_info=True`. Agreed it is pre-existing rather than a regression, but since this line is the only remaining diagnostic for media-creator failures it should carry the stack.

**F4 (mutation guard covered only image).** Added `tests/core/tools/adapters/vibe/test_media_creator_contract.py` — one parametrized test across all five creators, driven by a config getter that raises. Verified mutation-effective beyond image: wrapping `create_video_tools_from_config` in a blanket handler turns the `create_video_tools_from_config-get_video_models` case red. I kept the existing `has_ability`-based test in `test_image_tool.py` since it exercises the deeper path (`get_tools()`), which the getter-level test does not reach.

**F6 (the comment overstates the guarantee).** You are right, and I fixed the comment rather than the behaviour — `list_image_models` disappearing when an ability probe *raises* is the outcome this PR wants: a raising `has_ability` is a broken model class, and surfacing it is the whole point. Making it survive would mean catching around the probe, which is the handler #1317 asks to remove. The comment now says it stays "whenever the ability probes return" and names what happens when one raises.

**F7 (undocumented contract).** One sentence in `ToolRegistry.register` ("creators must not blanket-catch"), and the full contract spelled out in `create_registered_tools`: the two typed exceptions propagate, anything else warns and that creator contributes nothing.

**F3 (same anti-pattern in sibling creators).** Filed as #1333, out of scope here as you suggested. I swept every `@register_tool` creator rather than only the four listed, which changed the picture slightly:

- 11 creators to clean, including four not in your list: `skill_tools.py:416`, `translate_json.py:393`, `vision_tool.py:521`, `workspace_file_tool.py:295`.
- `custom_api_factory.py:59-65` is a partial case rather than correct — it re-raises `ConnectorRuntimeError` but not `RequiredMCPUnavailableError`, and still converts everything else to `[]` after logging, duplicating the factory's log line.
- Two are deliberately excluded: `mcp_tools.py:282` (the generic branch builds and emits the MCP load summary) and `ssh_tools.py:355,401` (already narrow catches around specific calls — the shape #1317 recommends).
- `knowledge_tools.py:63` and `file_ingestion_tool.py:312` sit in the `_impl` functions reached through the facade, not in the registered creator bodies.

Local run of the affected suites (`test_media_creator_contract.py`, `test_image_tool.py`, `test_music_tool.py`, `test_sound_effect_tool.py`, `test_audio_tool.py`, `test_factory_runtime_tools.py`): 128 passed. pre-commit clean.

PR #1222, opened 2026-08-09
fix(monitor): use a real regex operator in the PostgreSQL JSON guard

codeacme17:
## Summary

The PostgreSQL branch of `get_json_field_expression` guarded against control characters with `column.op("~?")`. `~?` is not a PostgreSQL operator for `json`, `jsonb` or `text`, so every query routed through that branch failed on PostgreSQL while the SQLite suite stayed green.

Verified against PostgreSQL 16.14:

```
=> SELECT ('{}'::json) ~? 'x';
ERROR:  operator does not exist: json ~? unknown

=> SELECT ('{}'::jsonb) ~? 'x';
ERROR:  operator does not exist: jsonb ~? unknown
```

Casting to `jsonb` does not help, so no schema change is involved.

## Blast radius

All three call sites swallow the error, and the two failure shapes differ:

| Endpoint | What the caller sees |
| --- | --- |
| `GET /monitor/stats` | HTTP 500 |
| `GET /monitor/popular-tools` | HTTP 200 and `[]` |
| `GET /monitor/model-stats` | HTTP 200 and `[]` |

The two list endpoints issue no further query after their handler, so the empty substitution holds and the caller gets a plausible-looking dashboard of zeros with only a `logger.error` line as evidence.

`/monitor/stats` is different. Its handler does set `active_models = 0`, but the rejected statement leaves the transaction aborted, so the following `llm_call_end` scan raises `InFailedSqlTransaction` outside any local handler and the outer `except` turns the request into a 500. Issue #1149 originally described this endpoint as returning zeros; that row has been corrected, and the new test pins the real behavior.

## Fix

The guard now casts the column to text and matches with `~`, the real regex-match operator.

It also looks for what actually breaks extraction. PostgreSQL accepts a NUL escape into a `json` column but raises when `->>` converts such a value to text:

```
=> SELECT '{"a":"\u0000"}'::json ->> 'a';
ERROR:  unsupported Unicode escape sequence
DETAIL:  \u0000 cannot be converted to text.
```

So the six-character escape sequence is the thing to find. A raw control character cannot appear in valid JSON at all (RFC 8259 requires escaping), which means the old character class would never have matched even with a working operator.

Two alternatives were considered and rejected:

- **Drop the guard.** Those payloads would reach `->>` and fail the request again, for the reason shown above.
[...]

PR #1319, opened 2026-08-13
enh(trace): store trace payloads as jsonb and sanitize them on write

### LINKED ISSUE #1248: enh(trace): TraceEvent.data accepts JSON escapes PostgreSQL cannot read back
Raised as a design note during the review of #1222 (fix for #1149). Out of scope there; filing so it is tracked rather than lost.

## Background

`TraceEvent.data` is `Column(JSON, nullable=False)` (`src/xagent/web/models/task.py`). On PostgreSQL that is the `json` type, which validates syntax at write time but nothing else. In particular it accepts escape sequences that PostgreSQL will later refuse to convert to text:

- `\u0000` -- `->>` raises `unsupported Unicode escape sequence`
- either half of an unpaired UTF-16 surrogate — `->>` raises `invalid input syntax for type json`

Nothing sanitizes the payload on the way in. `json.loads` in the agent result path accepts unpaired surrogate escapes without complaint, and `json.dumps` re-emits them, so such a value can reach the column through ordinary LLM or tool output.

#1149 was the read side of this: a single stored row of that shape failed the entire query on three monitoring endpoints. #1222 fixed it by guarding at read time — the query now nulls those payloads out before extraction.

## The problem with the current shape

The guard is a read-side workaround for an unvalidated write path. Every future query that extracts from this column has to remember to defend itself, and forgetting is silent until a payload of that shape actually lands. The `jsonb` type rejects these at insert; the `json` type does not.

## Options

1. **Migrate the column to `jsonb`.** Removes the whole class of problem at the source: the bad value can never be stored, and the read-side guard in `monitor.py` becomes unnecessary. Needs a migration and a decision about existing rows that already carry such escapes — the cast will fail on them, so they have to be cleaned or quarantined first. Also changes key-order and duplicate-key semantics, which is worth checking against any consumer that round-trips the payload.
2. **Sanitize at write time** wherever trace events are constructed, leaving the column as `json`. Cheaper, no migration, but it has to be applied at every producer and stays easy to bypass.
3. **Leave as is.** The read-side guard works and is now pinned by tests. This issue then serves as the record of why the guard exists.

Option 1 is the only one that actually removes the class. Whichever is chosen, the guard in `get_json_field_expression` should not be removed until the column can no longer hold such a payload.

## References

- Guard and its rationale: `src/xagent/web/api/monitor.py`, `get_json_field_expression` PostgreSQL branch
- Behaviour pinned by: `tests/web/test_monitor_api.py::TestUnsafeEscapePattern` and `tests/web/api/test_monitor_postgresql.py`

codeacme17:
Closes #1248.

## Background

`TraceEvent.data` was `Column(JSON)`, which on PostgreSQL is the `json` type: it validates syntax at write time and nothing else. It accepts the NUL escape and either half of an unpaired UTF-16 surrogate, both of which `->>` then refuses to convert back to text — and a single stored row of that shape failed three monitoring endpoints whole-query (#1149). #1222 guarded at read time; #1248 filed the design note that the guard is a workaround for an unvalidated write path.

`jsonb` decodes escapes to native text at INSERT, so it rejects those payloads at the source. That is option 1 in the issue, and the only option that removes the class.

## What this does

**Converts three columns, not one.** The checkpoint blob tables hold hashed slices of the same trace payloads, so they carry the same hazard and convert together:

- `trace_events.data`
- `trace_message_blobs.message_data`
- `trace_checkpoint_blobs.blob_data`

**Sanitizes on write.** `jsonb` alone would turn the problem from a failed read into a failed *write* — an ordinary LLM payload carrying one of these code points would stop being storable. A new sanitizer (`web/utils/json_payload_sanitizer.py`) replaces them with U+FFFD before the row is built, so the type change rejects nothing a producer legitimately emits. It runs at both `TraceEvent` construction sites — `stage_trace_event_row` and the `_persist_agent_outbound_event` path that bypasses it — and in staging it runs *before* checkpoint encoding, so blob hashes are computed over what actually lands in the column. The blob rows and the `ON CONFLICT` insert path both derive from that already-sanitized payload; `encode_checkpoint_data_for_storage` has exactly one caller.

**Migrates existing rows.** `ALTER ... TYPE jsonb` fails on any stored row carrying such an escape, so the upgrade first rewrites those rows, using the same detection steps as the read guard and doing the actual rewrite in Python where escape semantics are exact. Matching rows are walked in bounded keyset batches rather than collected up front — these payloads are whole checkpoint snapshots, and a deployment that emitted bad output for a while should not materialize all of them in the migration process at once. PostgreSQL-only; both branches are no-ops on SQLite.

**Keeps the read guard.** On a migrated database it can never match, but `monitor.py` cannot assume the migration has run — a database still on the `json` type carries exactly the rows it defends against. Per the issue, it stays until no deployment can hold such a payload.

## One regression this change would otherwise have introduced

`jsonb` re-renders numbers in plain notation, so a float written as `1e+16` reads back as the int `10000000000000000`. The checkpoint blob path re-hashes payloads it reads and compares them against the write-time hash (`strict=True` on restore), so such a payload would be rejected as corrupt and the task would silently lose its checkpoint.

The boundary is exact: only `|x| >= 1e16` is affected, because every float that large is integral (the mantissa runs out of fractional bits above 2**53). The sanitizer normalizes those floats at the same boundary, before the hash is taken, so the stored form and the read-back form are identical. A round-trip test pins this, with a control case that skips the sanitizer to prove the retype is real and the test is not a tautology.

Pre-existing rows carrying such a float are deliberately not rewritten; the migration docstring explains why, and the existing failure mode for an unreadable checkpoint row is already a logged skip that falls back to an older checkpoint.

## Testing

- 25 sanitizer unit tests, including identity assertions that pin the no-copy contract on clean payloads.
- 21 migration tests: conversion, row cleanup, benign payloads surviving, blob tables, idempotence, missing tables, downgrade, offline `--sql` rendering, and multi-batch cleanup covering the keyset advance.
- The PostgreSQL monitor tests are reworked: the hazardous shapes can no longer be seeded at all, so a new class asserts the INSERT itself fails — the invariant the migration exists to establish — alongside the sanitizer round-trip.
- 150 tests across the affected suites pass against a real PostgreSQL 16.
- Also verified by hand that the JSON path queries checkpoint restore depends on (`trace_handlers.py`) still work against a `jsonb` column; that path had no PostgreSQL coverage.

CI gets a step for the new migration's PostgreSQL tests.

## Deployment note

This is the expensive kind of migration. `ALTER ... TYPE` takes an ACCESS EXCLUSIVE lock and rewrites every row — there is no in-place path from `json` to `jsonb` — and the cleanup scan before it is an unindexed full-table pass. On a large deployment it should be scheduled rather than slipped into a routine restart; pruning checkpoint history first shortens both steps. The migration docstring carries this, along with the UTF8 server-encoding assumption the detection patterns inherit from the read guard.

### DIFF
--- a/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
+++ b/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
@@ -159,7 +159,7 @@ corrupted and a retry succeeds, but roll the application out fully before
 migrating, or expect to run it more than once.
 
 Revision ID: 20260813_trace_json_columns_to_jsonb
-Revises: 20260810_add_hubspot_marketing_scopes
+Revises: 20260812_add_slack_history_reactions_files_scopes
 Create Date: 2026-08-13
 
 """
@@ -174,7 +174,7 @@ from alembic import op
 from sqlalchemy.dialects import postgresql
 
 revision: str = "20260813_trace_json_columns_to_jsonb"
-down_revision: Union[str, None] = "20260810_add_hubspot_marketing_scopes"
+down_revision: Union[str, None] = "20260812_add_slack_history_reactions_files_scopes"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
+++ b/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
@@ -168,6 +168,19 @@ def _table_exists(name: str) -> bool:
 
 
 def _column_is_jsonb(table: str, column: str) -> bool:
+    """Whether the column is already jsonb. Callers must have established
+    that the table exists (both do, via _table_exists three lines up); a
+    missing one raises NoSuchTableError out of here rather than resolving
+    to a bool.
+
+    That is deliberate, not an oversight. This answer carries opposite
+    polarity in the two directions -- upgrade() reads False as "convert
+    this column", downgrade() reads False as "skip it" -- so there is no
+    default a missing table could safely collapse to. Swallowing it would
+    tell upgrade() to proceed against a table that is not there, and the
+    cleanup SELECT would raise the same failure one step later with a
+    worse message.
+    """
     for col in sa.inspect(op.get_bind()).get_columns(table):
         if col["name"] == column:
             return isinstance(col["type"], postgresql.JSONB)
--- a/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
+++ b/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
@@ -133,13 +133,22 @@ _UNSTORABLE_CODE_POINTS = re.compile("[\x00\ud800-\udfff]")
 _REPLACEMENT_CHARACTER = "�"
 _EXPONENT_NOTATION_THRESHOLD = 1e16
 
-# Rows per cleanup batch. Small enough that one batch of checkpoint blobs
-# stays comfortably in memory, large enough that the scan is not dominated
-# by round trips on the usual case of nothing to fix.
-REWRITE_BATCH_SIZE = 100
+# Ids per cleanup batch. This bounds a list of integers, not payloads --
+# the rewrite loop reads one payload at a time (see
+# _rewrite_unconvertible_rows) -- so it can be generous without putting
+# blob rows in memory.
+REWRITE_BATCH_SIZE = 500
 
 
 def _sanitize(value: Any) -> Any:
+    """Rewrite one payload. Eager rather than copy-on-write, unlike the
+    app-side sanitizer this mirrors: every row that reaches here matched
+    the unconvertible-escape predicate, so the "clean payload, hand it back
+    untouched" fast path that motivates the copy-on-write version there
+    cannot fire here. ``re.sub`` already returns the original object when
+    nothing matches, so only the container spine is rebuilt, never the text
+    -- a few percent of the json.loads/json.dumps pair that brackets this
+    call, and far less than one payload."""
     if isinstance(value, str):
         return _UNSTORABLE_CODE_POINTS.sub(_REPLACEMENT_CHARACTER, value)
     # bool is an int subclass, and True/False are not numbers to normalize.
@@ -172,26 +181,33 @@ def _rewrite_unconvertible_rows(table: str, column: str) -> None:
     (younger) write-side sanitizer, so the scan usually matches nothing;
     the full-table regex pass runs once, here, not on any query path.
 
-    Matching rows are walked in id order, one bounded batch at a time,
-    rather than collected up front. These payloads are whole trace events
-    and checkpoint blobs -- the blob tables exist precisely because they
-    get large -- so a deployment that emitted bad output for a while could
-    otherwise materialize every poisoned payload in the migration process
-    at once. Keyset pagination, not OFFSET: the loop rewrites the rows it
-    just read, and an OFFSET walk would skip past rows as the result set
-    shifts under it.
+    The scan pages over matching *ids* and then reads one payload at a
+    time, so peak memory is one payload rather than a batch of them. That
+    split is the point: these values are whole trace events and checkpoint
+    blobs, and nothing caps their size -- the payload truncation in
+    ``core/agent/runtime.py`` applies to LLM-category events only, and
+    checkpoints are ``system_update_general``. Holding a batch of them, or
+    the whole matching set, would make the migration's memory a function of
+    how much bad output a deployment happened to emit.
+
+    Keyset pagination, not OFFSET: the loop rewrites the rows it just read,
+    so they drop out of the predicate, and an OFFSET walk would skip past
+    rows as the result set shifts under it.
     """
     bind = op.get_bind()
-    # noqa: S608 on both statements -- table/column come from
-    # TRACE_JSON_COLUMNS above, never from user input, and every payload
-    # pattern is bound as a parameter rather than interpolated.
-    select_batch = sa.text(
-        f"SELECT id, CAST({column} AS text) AS payload FROM {table} "  # noqa: S608
+    # noqa: S608 throughout -- table/column come from TRACE_JSON_COLUMNS
+    # above, never from user input, and every payload pattern is bound as a
+    # parameter rather than interpolated.
+    select_ids = sa.text(
+        f"SELECT id FROM {table} "  # noqa: S608
         f"WHERE id > :after "
         f"AND regexp_replace("
         f"replace(CAST({column} AS text), :bs, :standin), "
         f":pair, '', 'g') ~ :unsafe "
         f"ORDER BY id LIMIT :limit"
+    )
+    select_payload = sa.text(
+        f"SELECT CAST({column} AS text) FROM {table} WHERE id = :id"  # noqa: S608
     )
     # The column is still json at this point -- the ALTER runs after this
     # returns -- so the rewritten payload is cast back to json, not jsonb.
@@ -202,25 +218,35 @@ def _rewrite_unconvertible_rows(table: str, column: str) -> None:
 
     after = 0
     while True:
-        rows = bind.execute(
-            select_batch,
-            {
-                "after": after,
-                "bs": ESCAPED_BACKSLASH,
-                "standin": ESCAPED_BACKSLASH_STANDIN,
-                "pair": SURROGATE_PAIR_PATTERN,
-                "unsafe": UNSAFE_ESCAPE_PATTERN,
-                "limit": REWRITE_BATCH_SIZE,
-            },
-        ).fetchall()
-        if not rows:
+        row_ids = [
+            row[0]
+            for row in bind.execute(
+                select_ids,
+                {
+                    "after": after,
+                    "bs": ESCAPED_BACKSLASH,
+                    "standin": ESCAPED_BACKSLASH_STANDIN,
+                    "pair": SURROGATE_PAIR_PATTERN,
+                    "unsafe": UNSAFE_ESCAPE_PATTERN,
+                    "limit": REWRITE_BATCH_SIZE,
+                },
+            )
+        ]
+        if not row_ids:
             return
-        for row_id, payload_text in rows:
+        for row_id in row_ids:
+            payload_row = bind.execute(select_payload, {"id": row_id}).fetchone()
+            if payload_row is None:
+                # Deleted between the id scan and this read -- checkpoint
+                # pruning runs concurrently against these very tables, and
+                # a row that no longer exists needs no rewrite.
+                continue
+            payload_text = payload_row[0]
             cleaned = _sanitize(json.loads(payload_text))
             bind.execute(update, {"payload": json.dumps(cleaned), "id": row_id})
         # A rewritten row no longer matches the predicate, so the next
         # batch would find these again only by id -- advance past them.
-        after = rows[-1][0]
+        after = row_ids[-1]
 
 
 def upgrade() -> None:
--- a/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
+++ b/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
@@ -46,17 +46,78 @@ duplicate keys cannot be produced by ``json.dumps`` from a dict in the
 first place -- the write path only ever serializes dicts.
 
 ``jsonb`` does, however, re-render numbers: a float stored as ``1e+16``
-reads back as the int ``10000000000000000``. The checkpoint blob path
-re-hashes payloads it reads and compares them against the write-time
-hash, so a *pre-existing* row carrying such a float becomes undecodable
+reads back as the int ``10000000000000000``, and ``-0.0`` as ``0.0``
+(``numeric`` has no signed zero). The checkpoint blob path re-hashes
+payloads it reads and compares them against the hash recorded at write
+time, so a *pre-existing* row carrying either shape becomes undecodable
 after this migration. New writes are covered -- the write-side sanitizer
-normalizes those floats before the hash is taken -- and rows this
-migration rewrites are normalized here too. Rows it does not touch are
-deliberately left alone: catching them would mean rewriting every row in
-the three tables to defend against a payload shape that requires a float
-above 2**53 in trace output, and the existing failure mode for an
-unreadable checkpoint row is already a logged skip that falls back to an
-older checkpoint (``web/api/trace_handlers.py``), not a lost task.
+normalizes both before the hash is taken -- and rows this migration
+rewrites get the same two normalizations applied here. Rows it does not
+touch are deliberately left alone: catching them would mean rewriting
+every row in all three tables to defend against shapes that need a float
+above 2**53, or a negative zero, in trace output.
+
+Rewriting a *blob* row has a cost the paragraph above does not cover, and
+it is the one thing to weigh before running this. Replacing NUL with
+U+FFFD changes the payload, so its canonical hash changes -- but the hash
+is also the blob's identity: ``trace_events.data`` references a blob by
+embedding the hash value itself, and strict restore re-hashes what it
+read and compares it against that reference
+(``_load_messages_by_refs`` / ``_load_checkpoint_blob_by_ref``,
+``web/services/trace_message_storage.py``). So a rewritten blob row still
+resolves but fails verification: ``CheckpointMessageDecodeError``, which
+``web/api/trace_handlers.py`` logs and treats as a skip, falling back to
+an older checkpoint row.
+
+Whether that fallback finds anything depends on which blob it was. Blob
+rows are deduplicated per task, and ``context.system_prompt`` and
+``context.metadata`` are typically byte-identical across every checkpoint
+of a task -- so for those, one poisoned row means *no* readable
+checkpoint at all, and the scan escalates to ``CheckpointCorruptError``:
+HTTP 400 from the a2a resume path and a non-recoverable verdict in lease
+recovery. For a mid-conversation message blob, checkpoints predating that
+message still decode, so the task loses progress rather than the task.
+
+The hash columns are therefore left verbatim on purpose, and the two
+alternatives were considered and rejected:
+
+* Recomputing ``message_hash``/``blob_hash`` from the rewritten payload
+  makes the row unreachable rather than unverifiable -- the references in
+  ``trace_events.data`` still name the old hash -- and it can abort the
+  migration outright: if the task re-checkpointed the same content after
+  the sanitizer landed, the new hash already exists and the UPDATE
+  violates ``uq_trace_message_blobs_task_hash``. Doing it properly would
+  mean remapping the hash inside every referring payload, including the
+  refs marker's own hash, which is the application's hashing contract
+  reimplemented in a migration.
+* Deleting the affected blob rows lands on the same
+  ``CheckpointMessageDecodeError`` for strict restore, so it buys nothing
+  there, and it costs the non-strict readers: every viewer path
+  (``api/conversation_logs.py``, ``api/workforces.py``,
+  ``services/task_setup_snapshot.py``) decodes with ``strict=False``,
+  which skips hash verification and renders the cleaned payload fine. A
+  deleted row breaks those too.
+
+Only NUL can reach a blob row this way. ``canonical_json_bytes`` uses
+``ensure_ascii=False``, so a payload carrying an unpaired surrogate
+raises ``UnicodeEncodeError`` before it can be hashed -- such a payload
+never became a blob row in the first place. It can sit in
+``trace_events.data``, which has no self-hash, and rewriting that column
+is harmless.
+
+Before running this on a deployment that predates the write-side
+sanitizer, count the blob rows that would be rewritten. Neither query
+writes anything:
+
+    SELECT count(*) FROM trace_message_blobs
+    WHERE CAST(message_data AS text) LIKE '%\\u0000%';
+
+    SELECT count(*) FROM trace_checkpoint_blobs
+    WHERE CAST(blob_data AS text) LIKE '%\\u0000%';
+
+A non-zero count means some tasks will lose strict checkpoint recovery.
+Pruning those tasks' checkpoint history first is the supported remedy and
+turns a surprise 400 into a deliberate one.
 
 The detection patterns assume a UTF8 server encoding, which is what the
 shipped compose file runs. On a server in another encoding the cast also
@@ -68,16 +129,34 @@ nothing is converted), but it needs the row cleaned by hand before the
 upgrade can proceed. The read guard in ``web/api/monitor.py`` documents
 the same limitation for the same patterns.
 
+Two further shapes are legal in ``json`` and rejected by ``jsonb``, and
+this cleanup handles neither: a number outside ``numeric``'s exponent
+range, and nesting past ``jsonb``'s depth limit. No application write path
+can produce either -- a Python float caps at ~1.8e308, ``json.dumps``
+recurses within the interpreter's own limit -- so only a manual or
+external writer could have planted one. Such a row fails the ALTER the
+same safe way as the encoding case above.
+
 Operationally this is the expensive kind of migration, and on a large
 deployment it should be scheduled rather than slipped into a routine
 restart. ``ALTER ... TYPE`` takes an ACCESS EXCLUSIVE lock and rewrites
 every row of the table -- there is no in-place path from ``json`` to
 ``jsonb``, because the on-disk representation genuinely differs -- and the
 cleanup scan that precedes it is an unindexed full-table regex pass over
-the same rows. Both run inside one transaction per table, so the lock is
-held for the whole of it, and ``trace_events`` is the table the monitoring
-dashboard already scans. Pruning checkpoint history first shortens both
-steps.
+the same rows. ``env.py`` sets ``transaction_per_migration=True`` on
+PostgreSQL, so all three tables' scans and ALTERs run in *one*
+transaction: the ``trace_events`` lock is taken first and held until the
+two blob tables have converted too. Budget the window for the whole
+migration, not per table, and note that ``trace_events`` is the table the
+monitoring dashboard already scans. Pruning checkpoint history first
+shortens every step.
+
+Deployment ordering matters for the same reason. An old instance still
+running without the write-side sanitizer can insert a fresh unconvertible
+row after the cleanup scan has passed it but before the ALTER takes its
+lock; the ALTER then fails and the whole migration rolls back. Nothing is
+corrupted and a retry succeeds, but roll the application out fully before
+migrating, or expect to run it more than once.
 
 Revision ID: 20260813_trace_json_columns_to_jsonb
 Revises: 20260810_add_hubspot_marketing_scopes
@@ -86,6 +165,7 @@ Create Date: 2026-08-13
 """
 
 import json
+import math
 import re
 from typing import Any, Sequence, Union
 
@@ -151,10 +231,14 @@ def _sanitize(value: Any) -> Any:
     call, and far less than one payload."""
     if isinstance(value, str):
         return _UNSTORABLE_CODE_POINTS.sub(_REPLACEMENT_CHARACTER, value)
-    # bool is an int subclass, and True/False are not numbers to normalize.
-    if isinstance(value, float) and not isinstance(value, bool):
+    # No bool guard needed: bool subclasses int, not float, so True/False
+    # never reach this branch.
+    if isinstance(value, float):
         if abs(value) >= _EXPONENT_NOTATION_THRESHOLD and value.is_integer():
             return int(value)
+        # numeric has no signed zero, so jsonb renders -0.0 as 0.0.
+        if value == 0.0 and math.copysign(1.0, value) < 0:
+            return 0.0
         return value
     if isinstance(value, dict):
         return {_sanitize(key): _sanitize(item) for key, item in value.items()}
--- a/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
+++ b/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
@@ -58,6 +58,16 @@ above 2**53 in trace output, and the existing failure mode for an
 unreadable checkpoint row is already a logged skip that falls back to an
 older checkpoint (``web/api/trace_handlers.py``), not a lost task.
 
+The detection patterns assume a UTF8 server encoding, which is what the
+shipped compose file runs. On a server in another encoding the cast also
+rejects any escape naming a character outside that charset -- including a
+valid surrogate pair, which the pair-stripping step deliberately removes
+before matching -- so the cleanup would miss such a row and the ALTER
+would fail on it. That is a safe failure (the transaction rolls back and
+nothing is converted), but it needs the row cleaned by hand before the
+upgrade can proceed. The read guard in ``web/api/monitor.py`` documents
+the same limitation for the same patterns.
+
 Operationally this is the expensive kind of migration, and on a large
 deployment it should be scheduled rather than slipped into a routine
 restart. ``ALTER ... TYPE`` takes an ACCESS EXCLUSIVE lock and rewrites
@@ -123,6 +133,11 @@ _UNSTORABLE_CODE_POINTS = re.compile("[\x00\ud800-\udfff]")
 _REPLACEMENT_CHARACTER = "�"
 _EXPONENT_NOTATION_THRESHOLD = 1e16
 
+# Rows per cleanup batch. Small enough that one batch of checkpoint blobs
+# stays comfortably in memory, large enough that the scan is not dominated
+# by round trips on the usual case of nothing to fix.
+REWRITE_BATCH_SIZE = 100
+
 
 def _sanitize(value: Any) -> Any:
     if isinstance(value, str):
@@ -156,34 +171,56 @@ def _rewrite_unconvertible_rows(table: str, column: str) -> None:
     Rows carrying such payloads exist only where something slipped past the
     (younger) write-side sanitizer, so the scan usually matches nothing;
     the full-table regex pass runs once, here, not on any query path.
+
+    Matching rows are walked in id order, one bounded batch at a time,
+    rather than collected up front. These payloads are whole trace events
+    and checkpoint blobs -- the blob tables exist precisely because they
+    get large -- so a deployment that emitted bad output for a while could
+    otherwise materialize every poisoned payload in the migration process
+    at once. Keyset pagination, not OFFSET: the loop rewrites the rows it
+    just read, and an OFFSET walk would skip past rows as the result set
+    shifts under it.
     """
     bind = op.get_bind()
-    rows = bind.execute(
-        sa.text(
-            # noqa: S608 -- table/column come from TRACE_JSON_COLUMNS above,
-            # never from user input; the payload patterns are bound below.
-            f"SELECT id, CAST({column} AS text) AS payload FROM {table} "  # noqa: S608
-            f"WHERE regexp_replace("
-            f"replace(CAST({column} AS text), :bs, :standin), "
-            f":pair, '', 'g') ~ :unsafe"
-        ),
-        {
-            "bs": ESCAPED_BACKSLASH,
-            "standin": ESCAPED_BACKSLASH_STANDIN,
-            "pair": SURROGATE_PAIR_PATTERN,
-            "unsafe": UNSAFE_ESCAPE_PATTERN,
-        },
-    ).fetchall()
-
+    # noqa: S608 on both statements -- table/column come from
+    # TRACE_JSON_COLUMNS above, never from user input, and every payload
+    # pattern is bound as a parameter rather than interpolated.
+    select_batch = sa.text(
+        f"SELECT id, CAST({column} AS text) AS payload FROM {table} "  # noqa: S608
+        f"WHERE id > :after "
+        f"AND regexp_replace("
+        f"replace(CAST({column} AS text), :bs, :standin), "
+        f":pair, '', 'g') ~ :unsafe "
+        f"ORDER BY id LIMIT :limit"
+    )
     # The column is still json at this point -- the ALTER runs after this
     # returns -- so the rewritten payload is cast back to json, not jsonb.
     update = sa.text(
         f"UPDATE {table} SET {column} = CAST(:payload AS json) "  # noqa: S608
         f"WHERE id = :id"
     )
-    for row_id, payload_text in rows:
-        cleaned = _sanitize(json.loads(payload_text))
-        bind.execute(update, {"payload": json.dumps(cleaned), "id": row_id})
+
+    after = 0
+    while True:
+        rows = bind.execute(
+            select_batch,
+            {
+                "after": after,
+                "bs": ESCAPED_BACKSLASH,
+                "standin": ESCAPED_BACKSLASH_STANDIN,
+                "pair": SURROGATE_PAIR_PATTERN,
+                "unsafe": UNSAFE_ESCAPE_PATTERN,
+                "limit": REWRITE_BATCH_SIZE,
+            },
+        ).fetchall()
+        if not rows:
+            return
+        for row_id, payload_text in rows:
+            cleaned = _sanitize(json.loads(payload_text))
+            bind.execute(update, {"payload": json.dumps(cleaned), "id": row_id})
+        # A rewritten row no longer matches the predicate, so the next
+        # batch would find these again only by id -- advance past them.
+        after = rows[-1][0]
 
 
 def upgrade() -> None:
--- a/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
+++ b/src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
@@ -1,0 +1,257 @@
+"""convert trace payload columns from json to jsonb on PostgreSQL
+
+``trace_events.data`` was a ``json`` column, which validates syntax at
+write time and nothing else: it happily stores the NUL escape and either
+half of an unpaired UTF-16 surrogate, both of which ``->>`` then refuses
+to convert back to text -- one such row failed three monitoring endpoints
+whole-query (#1149). ``jsonb`` decodes escapes to native text at INSERT,
+so it rejects those payloads at the source (#1248). The two checkpoint
+blob tables carry the same payloads (their rows are hashed slices of the
+same trace data), so all three columns convert together:
+
+- ``trace_events.data``
+- ``trace_message_blobs.message_data``
+- ``trace_checkpoint_blobs.blob_data``
+
+``ALTER ... TYPE jsonb USING <col>::jsonb`` fails on any stored row that
+carries one of those escapes, so the online upgrade first rewrites such
+rows with the offending code points replaced by U+FFFD -- the same
+substitution the write-side sanitizer
+(``web/utils/json_payload_sanitizer.py``) applies to new payloads. Rows
+are found with the same normalize-then-match steps as the read guard in
+``web/api/monitor.py`` (``PG_ESCAPED_BACKSLASH`` and friends -- that file
+is the patterns' source of truth and documents why each step is
+load-bearing); the actual rewrite happens in Python, where escape
+semantics are exact, not with SQL regex surgery on the JSON text.
+
+This migration is PostgreSQL-only. SQLite stores JSON as TEXT and has no
+jsonb; both branches are no-ops there. On PostgreSQL the online branch
+guards on the table existing (a bare database has none of the three until
+``create_all`` runs) and on the column not already being jsonb (a
+create_all-first startup builds jsonb directly from the model, and this
+keeps the upgrade idempotent on rerun).
+
+The offline (--sql) branch emits only the three ALTERs: the cleanup step
+must read rows, which a MockConnection cannot. An offline script applied
+to a database still holding an unconvertible row fails on that ALTER --
+PostgreSQL offline scripts run inside the BEGIN/COMMIT wrapper Alembic
+emits, so the failure aborts the whole script and converts nothing.
+Clean such rows first (or run the migration online, which does it for
+you).
+
+``jsonb`` does not preserve key order or duplicate keys. Consumers of
+these columns deserialize into Python dicts (checkpoint restore, the
+monitoring endpoints), where key order already carries no meaning, and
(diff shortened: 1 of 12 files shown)


src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
```diff
@@ -0,0 +1,294 @@
...
+    # bool is an int subclass, and True/False are not numbers to normalize.
+    if isinstance(value, float) and not isinstance(value, bool):
+        if abs(value) >= _EXPONENT_NOTATION_THRESHOLD and value.is_integer():
+            return int(value)
+        return value
+    if isinstance(value, dict):
+        return {_sanitize(key): _sanitize(item) for key, item in value.items()}
+    if isinstance(value, list):
+        return [_sanitize(item) for item in value]
+    return value
```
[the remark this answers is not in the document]
codeacme17: Thanks — the premise here is right and worth acting on, but the specific target is not where the memory actually goes, so I addressed the premise a different way. Details, since this is a migration and worth getting on the record.

**Copy-on-write cannot fire in this file.** `_rewrite_unconvertible_rows` only ever hands `_sanitize` rows that matched the unconvertible-escape predicate. The normalization before that match is subtractive only — the escaped-backslash replacement removes backslashes and cannot manufacture one, and the pair-strip only deletes — so what survives to match is a genuine NUL or lone-surrogate escape, which `json.loads` turns into a real code point that `_sanitize` must replace. Every payload reaching it needs at least one change, so the "clean payload, hand it back untouched" fast path that motivates the copy-on-write version in `json_payload_sanitizer.py` is unreachable here. That fast path exists for the write path, where it runs on every trace event and almost all of them are clean.

**What remains is smaller than it looks.** `re.sub` returns the original string object when nothing matches, so the eager walk already shares every unchanged leaf with the parsed input — only the container spine was ever rebuilt, never the text, which is where trace payload bytes live. Measured on the actual loop body (`json.loads` then `_sanitize` then `json.dumps`) with checkpoint-shaped payloads, whole-body peak differed by 0.5–2.3% between the two implementations.

**The real term was next to it.** The loop held a whole batch of payload *texts* resident. Nothing caps these values: the truncation in `core/agent/runtime.py` applies to LLM-category events only, and checkpoints are `system_update_general`, so `trace_checkpoint_blobs.blob_data` is unbounded by construction. A batch of those is roughly a hundred times the term flagged here. So the scan now pages over matching *ids* and reads one payload at a time: peak is one payload regardless of blob size, and the batch size bounds a list of integers instead. Splitting the read introduced a row that can vanish between the id scan and its payload read — checkpoint pruning deletes from these tables concurrently — so that case is now skipped explicitly rather than assumed impossible.

`_sanitize` keeps its eager shape. Copying the copy-on-write version in would add roughly 45 lines with a sentinel and index-tracked prefix backfill to a run-once script that deliberately duplicates this logic to stay independent of application code, and about a third of it would be unreachable — the `tuple` branch exists because the write path sanitizes in-memory Python objects, and `json.loads` never produces a tuple. The reasoning is now recorded in a docstring on the function so the next reader does not have to re-derive it.

codeacme17:
Thanks for the review. Addressed in `3500cde`, though not the way suggested — the concern behind the comment is real, but the memory it points at is not the memory that matters here.

Short version: copy-on-write cannot fire in the migration, because every row reaching `_sanitize` matched the unconvertible-escape predicate by construction, so the fast path it exists for is unreachable. Measured, the two implementations differ by 0.5–2.3% of the loop body's peak — `re.sub` already returns the original object when nothing matches, so only the container spine was being rebuilt, never the text.

The term that actually mattered was one line away: the loop held a whole batch of payload *texts* resident, and these payloads are uncapped (`trace_checkpoint_blobs.blob_data` comes from `system_update_general` events, which the LLM-category truncation does not touch). The scan now pages over matching ids and reads one payload at a time, so peak is one payload regardless of blob size. That also surfaced a real concurrency case — a row can be deleted between the id scan and its payload read, since checkpoint pruning runs against these tables — which is now skipped explicitly.

Full reasoning in the inline reply.

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
```diff
@@ -0,0 +1,320 @@
...
+
+def _table_exists(name: str) -> bool:
+    return sa.inspect(op.get_bind()).has_table(name)
+
+
+def _column_is_jsonb(table: str, column: str) -> bool:
+    for col in sa.inspect(op.get_bind()).get_columns(table):
+        if col["name"] == column:
+            return isinstance(col["type"], postgresql.JSONB)
+    return False
```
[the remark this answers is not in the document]
codeacme17: You are right that `get_columns` raises here — I checked, SQLAlchemy 2.0.47 raises `NoSuchTableError` on both PostgreSQL and SQLite for a missing table, so the handler would be reachable. But it should not be added, because this particular helper cannot be made total by returning a bool.

**`False` means opposite things to the two callers.** In `upgrade()` it is a *go* signal:

```python
if _column_is_jsonb(table, column):
    continue          # already jsonb, nothing to do
_rewrite_unconvertible_rows(table, column)   # False lands here
```

In `downgrade()` the polarity flips — `if not _column_is_jsonb(...): continue` — where `False` means skip. So there is no default value a missing table could safely collapse to: whatever it returns is right for one direction and wrong for the other.

**Swallowing it would not prevent the crash, only relocate it.** With the exception caught, a missing table returns `False`, `upgrade()` reads that as "convert this column", and control reaches `_rewrite_unconvertible_rows`, whose first statement is `SELECT id FROM <table> WHERE ... regexp_replace(...)`. That raises `ProgrammingError: relation ... does not exist` and aborts the migration transaction anyway. The net effect is to turn a precise `NoSuchTableError` naming the table into a vaguer error from a full-table scan one step later.

**Both call sites already guard**, three lines up, with `_table_exists` — which is also how the sibling migrations in this tree handle these create_all-owned tables (`20260804_add_task_checkpoint_trace_event_anchor.py`'s `_columns()`, `20260802_add_workforce_run_last_activity_at.py`'s `_column_names`, and `32b62e058cbb`'s `if "agents" not in tables: return` all use an unguarded inspector call plus a caller-side check). The partial-database case the comment is aimed at is covered: `TestUpgradePostgres::test_missing_tables_are_skipped` drops `trace_checkpoint_blobs` and asserts the run still converts `trace_events`.

I have added a docstring on the helper in `2a3f62f` recording the polarity argument, so the constraint is visible at the function rather than only at its call sites — that is the legitimate part of the concern here, that self-containedness was not obvious from the function alone.

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
```diff
@@ -0,0 +1,333 @@
...
+        f":pair, '', 'g') ~ :unsafe "
+        f"ORDER BY id LIMIT :limit"
+    )
+    select_payload = sa.text(
+        f"SELECT CAST({column} AS text) FROM {table} WHERE id = :id"  # noqa: S608
+    )
+    # The column is still json at this point -- the ALTER runs after this
+    # returns -- so the rewritten payload is cast back to json, not jsonb.
+    update = sa.text(
+        f"UPDATE {table} SET {column} = CAST(:payload AS json) "  # noqa: S608
```
rogercloud, line 387: This UPDATE rewrites only `payload`, leaving `message_hash`/`blob_hash` (and the bytes columns) untouched. Pre-sanitizer rows with NUL are recoverable today (canonical serialization is `ensure_ascii=False`; NUL round-trips stably as `\u0000` and the write-time hash recomputes identically), but after this rewrite (NUL -> U+FFFD) strict recovery's `canonical_json_hash` no longer matches -> `CheckpointMessageDecodeError` -> silent fallback to an older checkpoint. Blob rows are deduplicated under `(task_id, hash)`, so one rewritten row can poison a task's entire checkpoint history (worst case `CheckpointCorruptError`); the docstring's "rows this migration rewrites are normalized here too" claim does not hold for the hash rows. Suggested fixes: document this caveat in the docstring (like the float paragraph), or for the blob tables recompute and update the hash columns from the canonical serialization of the rewritten payload — or delete the affected rows (equivalent to the missing-blob degradation path).

src/xagent/web/utils/json_payload_sanitizer.py
```diff
@@ -0,0 +1,148 @@
...
+
+def _sanitize(value: Any) -> Any:
+    """Return the sanitized form of ``value``, or ``_UNCHANGED``."""
+    if isinstance(value, str):
+        if _UNSTORABLE_CODE_POINTS.search(value):
+            return _UNSTORABLE_CODE_POINTS.sub(REPLACEMENT_CHARACTER, value)
+        return _UNCHANGED
+    # bool is an int subclass, and True/False are not numbers to normalize.
+    if isinstance(value, float) and not isinstance(value, bool):
+        if abs(value) >= _EXPONENT_NOTATION_THRESHOLD and value.is_integer():
```
rogercloud, line 124: The float branch misses `-0.0`: `abs(-0.0) >= 1e16` is False, so `-0.0` passes through, but PG numeric/jsonb has no negative zero and re-renders it as `0.0` (verified on PG 16.14). `canonical_json_bytes(-0.0)` emits `-0.0` while strict recovery reads back `0.0`, so any new checkpoint containing `-0.0` fails the strict hash check — the "boundary is exact" claim in the PR description does not hold. Tiny fix: normalize `value == 0.0` to `0.0` (covers `-0.0`) and add a `-0.0` round-trip test.

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
```diff
@@ -0,0 +1,333 @@
...
+upgrade can proceed. The read guard in ``web/api/monitor.py`` documents
+the same limitation for the same patterns.
+
+Operationally this is the expensive kind of migration, and on a large
+deployment it should be scheduled rather than slipped into a routine
+restart. ``ALTER ... TYPE`` takes an ACCESS EXCLUSIVE lock and rewrites
+every row of the table -- there is no in-place path from ``json`` to
+``jsonb``, because the on-disk representation genuinely differs -- and the
+cleanup scan that precedes it is an unindexed full-table regex pass over
+the same rows. Both run inside one transaction per table, so the lock is
```
rogercloud: "Both run inside one transaction per table" is inaccurate: `env.py` sets `transaction_per_migration=True` on PostgreSQL, so the whole migration — all three tables' cleanup scans and ALTERs — is one transaction, and the `trace_events` ACCESS EXCLUSIVE lock is held until the other two tables are converted. Suggest rewording to "one transaction for the whole migration" so operators can estimate lock hold times.

src/xagent/web/utils/json_payload_sanitizer.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _sanitize(value: Any) -> Any:
+    """Return the sanitized form of ``value``, or ``_UNCHANGED``."""
+    if isinstance(value, str):
+        if _UNSTORABLE_CODE_POINTS.search(value):
+            return _UNSTORABLE_CODE_POINTS.sub(REPLACEMENT_CHARACTER, value)
+        return _UNCHANGED
+    # bool is an int subclass, and True/False are not numbers to normalize.
+    if isinstance(value, float) and not isinstance(value, bool):
```
rogercloud: This clause is dead: `bool` subclasses `int`, not `float`, so `isinstance(True, float)` is always False and `and not isinstance(value, bool)` never changes the result; the comment's stated rationale ("bool is an int subclass") is also not the reason anything is filtered here. `test_bools_are_not_treated_as_numbers` passes regardless of the clause. Deleting it here (and in the migration `_sanitize`, ~lines 154-156) is a zero-behavior change.

tests/migrations/test_20260813_trace_json_columns_to_jsonb.py
```diff
@@ -0,0 +1,522 @@
...
+
+
+def _pre_migration_metadata() -> sa.MetaData:
+    """The three trace tables, reduced to the columns the migration touches
+    plus their primary keys -- not the full production schema."""
+    metadata = sa.MetaData()
+    sa.Table(
+        "trace_events",
+        metadata,
+        sa.Column("id", sa.Integer, primary_key=True),
```
rogercloud, line 79: Guard gaps in this test module: (a) the migration `_sanitize` float `>= 1e16` branch has zero coverage — no seed contains a large float; (b) no test rewrites a payload mixing valid surrogate pairs with NUL/orphan surrogates (the pair-strip + unsafe predicate interaction only fires in that mixed shape); (c) the fixtures omit `message_hash`/`blob_hash`/bytes columns, so nothing asserts hashes are preserved across the rewrite; (d) `TestSqliteNoOp`'s downgrade test has no assertions while the upgrade side has symmetric ones. Consider covering these so the migration's normalization contract is actually pinned.

rogercloud (changes requested):
## Summary

PG `json` columns validate syntax but not readability: NUL escapes and unpaired surrogates can be stored and then fail `->>` reads, letting one dirty row take down three monitoring endpoints (#1149); the #1222 read-side guard was only a band-aid. This PR implements issue option 1 — migrate the three trace `data` columns to `jsonb`, sanitize payloads on the write path, clean up existing rows in a migration, and keep the read guard as defense-in-depth. The change spans 13 files (a new ~148-line sanitizer, a ~333-line migration, ~522 lines of migration tests), with CI green including real PG16 migration tests, pytest-slow, and frontend-build.

## Approach verdict

Sound: enforcing the constraint at the storage layer (jsonb) plus write-side sanitization and a one-time cleanup migration — with the read guard retained for pre-fix deployments — addresses the root cause and degrades safely on legacy data.

## Design-level findings

- **C3 (minor)** — monitor.py guard retention: "do not remove it while any deployment can still hold such a payload" is unmeasurable across OSS deployments. Once jsonb makes the dangerous shapes un-plantable, the guard's matching path has zero real-PG execution and only dialect-agnostic unit tests pin it. Suggest an immediately checkable retirement condition (DB introspection or an ops SOP), or re-pin the guard's drop behavior with a bypass native-`json` test table.
- **C4 (minor)** — migration docstring deployment paragraph (lines 71-80): it does not mention the rolling-upgrade writer race — an old instance without the sanitizer can insert a new dirty row after the cleanup scan but before ALTER takes ACCESS EXCLUSIVE, failing the ALTER and rolling back the whole migration (safe, retryable). One sentence on deployment ordering (roll the app out fully before migrating, or expect retries) closes the gap.
- **C6 (minor)** — the docstring's NUL paragraph (lines 61-69) covers only non-UTF8 encodings; other shapes legal in `json` but rejected by `jsonb` exist (numeric range overflow >=1e131072 or <=1e-16384; asymmetric nesting-depth window ~19k-21k). App write paths cannot produce these (float <=1.8e308, int <=4300 digits, recursion <=1000 levels) — only manual/external writes can — so a one-line docstring note suffices; no logic change needed.
- **nit** — the dangerous-codepoint knowledge now lives in three copies with no cross-consistency test. Not raised as a finding: migrations never importing app code is a repo convention (137/137), and a consistency test would force migrations to track app drift. Flagged only as a documentation-wording note.

## Line-level findings

### Major

- **C1 [new] (major)** — `src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py:228`. The UPDATE rewrites only the `payload` column and leaves `message_hash`/`blob_hash` (and bytes columns) untouched. Pre-sanitizer rows containing NUL are recoverable before this migration — canonical serialization is `ensure_ascii=False`, NUL round-trips stably as the `\u0000` escape, and the write-time hash recomputes identically — but after the rewrite (NUL -> U+FFFD) strict recovery's `canonical_json_hash` no longer matches: `CheckpointMessageDecodeError` -> silent fallback to an older checkpoint. Because blob rows are deduplicated and shared under the `(task_id, hash)` unique constraint, one rewritten row can poison a task's entire checkpoint history; in the extreme, all candidates mismatch -> `CheckpointCorruptError`. The docstring's claim that "rows this migration rewrites are normalized here too" does not hold for the hash rows. Likelihood is low (only pre-sanitizer NUL rows), but the corruption is silent and surfaces only at recovery time. Suggested fixes: at minimum, make the docstring honest (parallel to the float paragraph); stronger, for the blob tables recompute and update the hash columns from the canonical serialization of the rewritten payload, or delete the affected rows (the missing-blob degradation path is equivalent).

### Minor

- **C2 [new]** — `src/xagent/web/utils/json_payload_sanitizer.py:113-115`: the float branch misses `-0.0` (see inline comment); the PR's "boundary is exact" claim does not hold.
- **C5 [new]** — migration docstring line 77: "Both run inside one transaction per table" is inaccurate under `transaction_per_migration=True`; the whole migration is one transaction (see inline comment).
- **C7 [new]** — sanitizer lines 112-113 (and the migration `_sanitize`, ~lines 154-156): the `not isinstance(value, bool)` clause is dead code (see inline comment).
- **C8 [new]** — `tests/migrations/test_20260813_trace_json_columns_to_jsonb.py`: four test-guard gaps (see inline comment).

## Update summary

Since the last third-party review (gemini-code-assist, 06:04:09Z, HEAD `3500cde8`), the only new commit is `2a3f62ff` — docs-only: a 13-line docstring recording why `_column_is_jsonb` lets a missing table raise (pinning P2's contract). No functional changes. The simplification lens re-scan of the updated diff found nothing to simplify: Lean already.

## Prior-findings checklist

- **P1 (migration `_sanitize` memory)** — WAIVED. The author's rebuttal holds, and `515fd47b` + `3500cde8` already fixed the real memory concern: the migration no longer holds whole-batch payloads in memory; it batch-fetches only ids and reads/writes rows one at a time.
- **P2 (`_column_is_jsonb` swallowing `NoSuchTableError`)** — WAIVED. The polarity-reversed argument is valid, and `2a3f62ff` added the docstring that pins the contract.

**Blocking: yes** — one blocking item, tagged [new]:

- **C1 [new] (major)** — `src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py:228`: the migration rewrites payloads (NUL -> U+FFFD) without updating `message_hash`/`blob_hash`, so strict recovery of affected blob rows silently fails and falls back to older checkpoints (worst case `CheckpointCorruptError`); the docstring's normalization claim does not hold for hash rows.

Recommended decision: **REQUEST_CHANGES** — address C1 (document the caveat, or recompute/update the blob hashes, or delete the affected rows) and fold in the trivial minors (C2, C5, C7). C3/C4/C6 resolve with documentation or test-only changes; no design rework required.

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
codeacme17: Confirmed, and this was the right thing to block on. The mechanism is exactly as described: strict restore re-hashes what it read and compares it against the reference, so a rewritten blob payload fails verification while its hash column still points at the old content. Fixed in `d1c1d17` — but by documenting rather than by either suggested remedy, because both make things worse. Reasoning, since you asked for a decision here:

**Recomputing the hash breaks the lookup instead of the verification.** The reference *is* the hash value, embedded in `trace_events.data` (`refs: [message_hash, ...]`, and `{"kind": ..., "hash": blob_hash}`), and resolution is `message_hash.in_(refs_chunk)`. Rewriting the column without rewriting every referring payload makes the row unreachable — `CheckpointMessageDecodeError("checkpoint message blobs missing for N refs")`, the same class of failure, now unrecoverable by any reader. It can also abort the migration outright: if the task re-checkpointed the same content after the sanitizer landed, the recomputed hash already exists and the UPDATE violates `uq_trace_message_blobs_task_hash`. Doing it properly means remapping the hash inside every referring payload including the refs marker's own hash — the application's hashing contract reimplemented inside a migration, where a subtle mismatch silently corrupts checkpoints that were previously fine.

**Deleting the rows is equivalent for strict restore and strictly worse for everything else.** You noted the equivalence yourself, which is precisely why it buys nothing: same `CheckpointMessageDecodeError`, same skip/fallback/corrupt verdict. What it costs is the non-strict readers — `api/conversation_logs.py`, `api/workforces.py` and `services/task_setup_snapshot.py` all decode with `strict=False`, which sets `verify_blob_hashes=False`, so they render the *cleaned* payload fine today. A deleted row breaks those too.

So the rewrite stays, and the docstring now carries the cost honestly. Two corrections to my earlier wording that your finding surfaced:

- **The escalation was understated.** I had described the fallback as "a logged skip that falls back to an older checkpoint". That is true for a mid-conversation message blob, but `context.system_prompt` and `context.metadata` are typically byte-identical across every checkpoint of a task and therefore deduplicated to *one* row — so a single poisoned row there leaves no readable checkpoint at all, and `CheckpointCorruptError` (HTTP 400 on resume, non-recoverable in lease recovery) is the normal outcome, not the worst case. The docstring says this now.
- **The finding is narrower than stated, in one respect.** Only NUL can reach a blob row. `canonical_json_bytes` uses `ensure_ascii=False`, so a payload carrying an unpaired surrogate raises `UnicodeEncodeError` before it can be hashed — it never became a blob row. Surrogates can sit in `trace_events.data`, which has no self-hash, and rewriting that column is harmless. Your premise about NUL round-tripping stably is right: I verified it is byte-identical through `json.dumps(ensure_ascii=False)`, so the write-time hash does recompute today.

I have also added a read-only pre-flight count to the docstring so an operator can see up front whether any blob rows will be rewritten, and prune those tasks' checkpoint history first — turning a surprise 400 into a deliberate one.

On the test side: the fixtures now carry `message_hash`/`blob_hash` and the bytes columns, and a new test pins that the migration leaves them verbatim. Their absence is exactly why this went unnoticed, so that gap is worth closing regardless of the outcome here.

src/xagent/web/utils/json_payload_sanitizer.py
codeacme17: Confirmed and fixed in `d1c1d17`. You are right that the "boundary is exact" claim did not hold — this is the same retype as the 1e16 case at the opposite end of the range, and I missed it.

Verified on PG 16 directly:

```
SELECT '{"n": -0.0}'::jsonb->>'n', '{"n": -0.0}'::json->>'n';
 0.0 | -0.0
```

The `json` column keeps the sign, `jsonb` drops it, so a checkpoint payload carrying `-0.0` would fail its own hash check on restore.

Normalized in both the app sanitizer and the migration's copy, using `math.copysign` since `-0.0 == 0.0` cannot distinguish them. Added a PostgreSQL round-trip test plus a control that stores an unsanitized `-0.0` and asserts the column really does hand back `0.0`, so the normalization is pinned as load-bearing rather than assumed. A `+0.0` case guards the identity fast path, so the new branch does not turn every zero into a copy.

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
codeacme17: Correct, and this matters for the operational guidance rather than just the wording — `env.py` sets `transaction_per_migration=is_postgresql`, so the `trace_events` ACCESS EXCLUSIVE lock is taken first and held until both blob tables have converted too.

Reworded in `d1c1d17` to say all three tables' scans and ALTERs run in one transaction, that operators should budget the window for the whole migration rather than per table, and that pruning checkpoint history first shortens every step.

I also folded in the rolling-upgrade race you raised as C4: an old instance without the sanitizer can insert a fresh unconvertible row after the cleanup scan has passed it but before the ALTER takes its lock, failing the ALTER and rolling the migration back. Nothing is corrupted and a retry succeeds, but the docstring now says to roll the application out fully before migrating, or expect to run it more than once.

src/xagent/web/utils/json_payload_sanitizer.py
codeacme17: Correct on both counts, and the comment was worse than the code — `bool` subclasses `int`, not `float`, so `isinstance(True, float)` is always False and the clause never changed a result. The stated rationale was a true fact about `bool` that had nothing to do with why anything is filtered in a float branch.

Removed from both copies in `d1c1d17`. `test_bools_are_not_treated_as_numbers` stays, with a comment noting it pins the outcome rather than the mechanism, since the mechanism is now "the branch never sees a bool at all".

tests/migrations/test_20260813_trace_json_columns_to_jsonb.py
codeacme17: All four closed in `d1c1d17`.

(c) was the most valuable one and it is directly connected to C1: the fixtures omitted `message_hash`/`blob_hash` and the bytes columns, which is exactly why the hash question went unnoticed while I was writing this. They are in the fixture now, and a new test seeds a known hash and byte count and asserts both survive the payload rewrite verbatim — pinning the deliberate choice rather than leaving it implicit.

(a) A row now carries `1e16` alongside its escape, so the migration's own float branch is exercised and the rewritten payload is asserted to come back as an int.

(b) Added the mixed shape in both directions: a valid pair on either side of an orphan, and literal escape-looking text followed by a genuine orphan. This is the shape where the pair-strip and the unsafe predicate actually interact — the pair has to be removed before matching or the orphan behind it is missed and the ALTER fails on the row.

(d) The SQLite downgrade test now asserts the same schema-and-rows invariants as the upgrade side instead of merely running without raising.

codeacme17:
Thanks — this was a genuinely useful review. All eight items addressed in `d1c1d17`; replies on each thread with the specifics. Summary of what changed and where I diverged.

**C1 (blocking) — mechanism confirmed, remedy different.** Strict restore compares `canonical_json_hash(payload)` against the *reference*, so a rewritten blob payload fails verification while the hash column still names the old content. Real, and worth blocking on. But both suggested remedies are worse than the problem: recomputing the hash makes the row unreachable rather than unverifiable (the reference embedded in `trace_events.data` still names the old hash) and can abort the migration on `uq_trace_message_blobs_task_hash` when the task re-checkpointed the same content post-sanitizer; deleting the row reaches the identical `CheckpointMessageDecodeError` for strict restore while breaking the three viewer paths that decode with `strict=False` and render the cleaned payload fine today. So the rewrite stays and the docstring now carries the cost honestly — including the escalation my earlier wording understated: because the system-prompt and metadata blobs are typically deduplicated to one row per task, a single poisoned row can leave no readable checkpoint at all, making `CheckpointCorruptError` the normal outcome there rather than a logged skip. Added a read-only pre-flight count so an operator can prune first.

One narrowing worth recording: only NUL can reach a blob row. `canonical_json_bytes` uses `ensure_ascii=False`, so an unpaired surrogate raises `UnicodeEncodeError` before it can be hashed — such a payload never became a blob row. Your premise about NUL round-tripping stably is right; I verified it is byte-identical.

**C2 — real gap, fixed in code.** The one item here that needed more than documentation, and the same class of bug as the 1e16 case at the other end of the range. Verified on PG 16: `'{"n": -0.0}'::jsonb->>'n'` returns `0.0` while the `json` cast returns `-0.0`. Normalized in both sanitizer copies with a round-trip test and a control proving the column drops the sign. The PR description's "boundary is exact" claim was wrong.

**C5, C7 — both accurate.** `transaction_per_migration=True` means one transaction for the whole migration, which changes the lock window operators need to budget; and the bool guard was dead code, with a comment whose stated rationale was irrelevant to what it did. Both corrected.

**C4, C6 — documented.** Rolling-upgrade writer race and its retry semantics; the numeric-range and nesting-depth shapes `jsonb` also rejects, with the note that no application write path can produce them.

**C8 — all four closed.** (c) is the one that mattered most: the fixtures omitted the hash and bytes columns, which is precisely why the C1 question went unnoticed while I wrote this. They are in the fixture now and a test pins that the migration leaves them verbatim.

**C3 — not addressed, and I would like your read.** The retirement condition for the `monitor.py` guard is the one item I have not acted on. Your point stands that "while any deployment can still hold such a payload" is unmeasurable across OSS installs. Of the two suggestions, the bypass native-`json` test table appeals more than an ops SOP, since it keeps the guard's matching path executing against real PostgreSQL rather than only dialect-agnostic unit tests. I did not want to add a second schema shape to the PG test fixture in this PR without your view on whether that is worth it here or better as a follow-up.

Local verification: 228 tests across the affected suites pass against PostgreSQL 16, plus mypy and a single alembic head.

rogercloud:
ci is failing

src/xagent/web/api/monitor.py
```diff
@@ -100,8 +100,15 @@ def get_json_field_expression(column: Any, field_path: str, db_session: Session)
...
-        # so the row drops instead. jsonb would reject them at write time, but
-        # this column is json, which stores them happily.
+        # so the row drops instead.
+        #
+        # These columns are jsonb since #1248, and jsonb rejects such payloads
+        # at INSERT, so on a migrated database this guard can never match. It
+        # stays because it is not this module's business to assume the
+        # migration has run: a database still on the json type -- one upgraded
+        # by hand, or mid-rollout -- carries exactly the rows it defends
+        # against. Per that issue, do not remove it while any deployment can
```
rogercloud: Re-raising the prior review C3 (previously body-only; stays minor and non-blocking). The retention condition here is still unmeasurable: there is no retirement condition, no bypass test against a native `json` column, and this PR deletes the base real-PG guard-drop coverage (7 hazardous seeds with drop assertions, replaced by INSERT-rejection tests). The dialect-independent unit tests mirror the guard semantics in Python and never execute the SQL. Clarification: a `~?` operator regression would not slip through — a compile-only test asserts `~?` is absent on every CI leg, and the three PG endpoint tests execute the full guard SQL against real PostgreSQL; what lost real-PG coverage is only the hazardous-row drop path (pair-strip/escape-pattern SQL wiring). Suggest a scratch `json`-typed table (raw DDL) pinning the drop path, or an explicit retirement condition. Minor, non-blocking.

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
```diff
@@ -0,0 +1,417 @@
...
+            return
+        for row_id in row_ids:
+            payload_row = bind.execute(select_payload, {"id": row_id}).fetchone()
+            if payload_row is None:
+                # Deleted between the id scan and this read -- checkpoint
+                # pruning runs concurrently against these very tables, and
+                # a row that no longer exists needs no rewrite.
+                continue
+            payload_text = payload_row[0]
+            cleaned = _sanitize(json.loads(payload_text))
```
rogercloud: This `json.loads` round-trips matched rows through float64. A row carrying both an escape and an external-writer decimal silently loses precision (0.123456789012345678901 → 0.12345678901234568) — permanent, because the rewritten row no longer matches the predicate — and 1e1000 → inf → json.dumps emits `Infinity` → the UPDATE CAST fails with a confusing error that the docstring (:132-138) misattributes to the ALTER. The ≥1e16 branch is a third silent corruption instance: `int(1e25)` ≠ 10^25. Suggest `parse_float=Decimal` (or loudly rejecting non-round-trippable numbers) plus a docstring correction. Minor, non-blocking.

src/xagent/web/utils/json_payload_sanitizer.py
```diff
@@ -0,0 +1,156 @@
...
+    copy of it is built along the way: this runs on every trace write and
+    almost every payload is clean, so the clean path allocates nothing.
+    Containers copy on first change only, backfilling the prefix they had
+    already walked.
+
+    Two distinct dict keys can collide after replacement (``"a\\x00"`` and
+    ``"a\\ud800"`` both become ``"a\\ufffd"``); the later key wins, matching
+    ``json.loads`` duplicate-key behaviour.
+
+    Recursion is unbounded on purpose. A cap would have to leave whatever
```
rogercloud: The docstring "fails its write loudly" claim holds only narrowly. A clean payload nested ~1000 levels deep (measured 996 on CPython 3.14; default recursion limit 1000) raises RecursionError inside `_sanitize`, and both runtime write paths swallow it log-only — websocket.py:819-823 (except Exception → rollback + logger.exception) and trace_handlers.py:161-174 (best-effort events) — so the write fails silently rather than loudly. The pre-PR path (json.dumps, C encoder) handled ≥20000 levels, so deeply nested outbound websocket rows are a newly exposed failure mode. Realism mitigates: websocket payloads are flat message dicts, and the staging path pre-existing `serialize_value` recursion (trace_handlers.py:1064-1090) crashes at the same depth first. Suggest an iterative traversal, or documenting the ceiling and making the failure loud. Minor, non-blocking.

rogercloud (approved):
## Summary

**enh(trace): store trace payloads as jsonb and sanitize them on write** (closes #1248) converts three PostgreSQL columns that hold trace payloads — `trace_events.data` plus two checkpoint blob tables — from `json` to `jsonb`. The `json` type stored NUL escapes and unpaired UTF-16 surrogates that `->>` refuses to convert back to text, and one such row failed three monitoring endpoints whole-query (#1149); #1222's read-time guard was the workaround this PR replaces. The change adds a batched, keyset migration that rewrites hazardous rows in Python (PostgreSQL-only; SQLite no-ops), a recursive sanitizer that mirrors exactly what jsonb does to values (NUL/surrogates → U+FFFD, `|x| >= 1e16` integral floats → int, `-0.0` → `0.0`) so legitimate LLM payloads stay writable, keeps the read guard until provably dead, and preserves the checkpoint hash contract (sha256 over canonical JSON, re-verified on strict restore).

## Update since the previous review

The previously reviewed chain was force-pushed: rebased onto the new base and replayed as aa9c9e6/00d7a6c/641bec7/0fc98a8/dfb4d36, then extended with e78c188 (the substantive fix) and fa19c12 (re-parent onto latest main). e78c188 adds negative-zero normalization (`math.copysign`) in both sanitizer copies, an honest blob-rewrite docstring with the escalation path corrected (context system_prompt/metadata blob dedup → no readable checkpoint → `CheckpointCorruptError` → a2a HTTP 400 / v1 409), read-only pre-flight count queries, the C4/C5/C6 docstring fixes, dead bool-clause removal, and four test-gap closures. CI is green.

## Prior findings

**C1 (major, previously blocking) — FIXED.** The blob-rewrite/hash-staleness concern is resolved by the documentation remedy the prior review explicitly allowed. The migration docstring (src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py:60-99) carries the cost honestly, corrects the escalation (context system_prompt/metadata blob dedup → no readable checkpoint → `CheckpointCorruptError` → a2a HTTP 400 / v1 409), documents both rejected alternatives with reasons, and adds read-only pre-flight count queries (:108-120). The new `test_blob_hash_columns_are_left_verbatim` (:459-490) pins hash/bytes verbatim; mechanism claims were verified against the code (trace_message_storage.py:1165-1199, trace_handlers.py:321-402, a2a.py:468-474).

> Footnote (cosmetic): the docstring phrase "non-recoverable verdict in lease recovery" (migration :77) is loosely attributed — `resolve_checkpoint_recovery` is pointer-row-identity based and never decodes blob payloads; the non-recoverable outcome actually surfaces at resume (a2a 400 / v1 409 / WS delivery failure).

**C2 (−0.0 missed by float branch) — FIXED.** `math.copysign` normalization in both the app sanitizer (:122-123) and the migration copy (:240-241); real-PG round-trip and control tests prove the sign drop is real (tests/web/api/test_monitor_postgresql.py:384-423); the +0.0 identity case was added.

**C3 (guard retention unmeasurable / zero real-PG guard execution) — NOT FIXED (minor, non-blocking; re-raised inline).** monitor.py is byte-identical to the previously reviewed state: the guard comment (monitor.py:105-111) still says "do not remove it while any deployment can still hold such a payload", with no retirement condition and no bypass test against a native `json` column. This PR deletes the base's real-PG guard-drop coverage (7 hazardous seeds + drop assertions, replaced by INSERT-rejection tests), and the test docstring (tests/web/api/test_monitor_postgresql.py:36-39) documents the gap ("not here") without closing it. The author has not replied to the prior comment; the prior review allowed doc/test-only resolution, so this stays open as a minor carry-over. Clarification on the prior round's strongest claim: a `~?` operator regression would **not** pass CI — a compile-only test asserts `~?` is absent on every CI leg, and the three PG endpoint tests execute the full guard SQL against real PostgreSQL. What genuinely lost real-PG coverage is only the hazardous-row drop path (pair-strip/escape-pattern SQL wiring against a real `json` column).

**C4 (rolling-upgrade writer race) — FIXED.** Docstring (:154-159) now covers the old-instance-inserts-dirty-row-after-scan-before-ALTER race, safe rollback, retryability, and roll-out ordering guidance.

**C5 (transaction wording) — FIXED.** Docstring (:146-150) now states one transaction for all three tables' scans + ALTERs, with the `trace_events` lock held until the blob tables convert and the whole window budgeted; `transaction_per_migration=is_postgresql` verified at env.py:167/194.

**C6 (other json-rejected shapes) — FIXED.** New "Two further shapes" paragraph (:132-138) documents the numeric exponent range and nesting past the jsonb depth limit, scoped to manual/external writers, with safe failure.

**C7 (dead bool clause) — FIXED.** Removed from both copies; bare `isinstance(value, float)` with an accurate comment; the test stays with an outcome-pinning comment.

**C8 (four test-guard gaps) — FIXED.** (a) 1e16 float seeded alongside the escape (:394); (b) payload mixing valid pairs with orphans, both directions (:354); (c) fixtures carry hash/bytes columns plus the verbatim test (:459); (d) downgrade test with real assertions (:249). Non-blocking observation: (a) is a coverage pin, not a regression discriminator — a jsonb cast would normalize 1e16 anyway.

**nit (three copies of knowledge) — WAIVED.** Duplication unchanged but explicitly justified: migration :190-193 names monitor.py as the source of truth, :209-211 explains the frozen-migration no-import convention, 0/138 migrations import app code, and the C2 fix landed identically in both copies.

## New findings

**DC1 (minor, design-level).** Sanitizer enforcement rests on convention at two construction sites, not structure. Both current sites sanitize correctly and are mutually documented ("known bypass" notes at src/xagent/web/services/trace_event_staging.py:30-34 ↔ src/xagent/web/api/websocket.py:773-776); the jsonb backstop makes a missed sanitize fail loudly on PostgreSQL. A future third writer would fail in production, not at build time, and on non-PG dialects (SQLite dev/test) a missed sanitize would store silently. Structural alternatives (TypeDecorator) were correctly rejected because the sanitize-before-hash ordering is load-bearing.

**DC3 (minor, design-level).** Pre-existing float-shaped rows (≥1e16, −0.0) get a probability argument, not a remediation path: the detection predicate matches only escape shapes, the pre-flight count queries match only `\u0000`, and the CheckpointCorruptError escalation narrative is NUL-path-only — a float-shaped context blob post-ALTER hits the same HTTP 400 / non-recoverable verdict with no count query and no remedy offered. The failure is disclosed ("becomes undecodable") and the leave-alone choice is defensible; the asymmetry (remedy for NUL, probability for floats) is the wrinkle. Suggest stating the float case explicitly in the docstring or extending the count query.

**NC1 (minor, inline).** Unbounded recursion: a clean payload nested ~1000 levels (measured 996 on CPython 3.14; default limit 1000, no `setrecursionlimit`) raises RecursionError inside the sanitizer, and both runtime write paths swallow it log-only (websocket.py:819-823 → rollback + `logger.exception`; trace_handlers.py:161-174 best-effort). The docstring's "fails its write loudly" claim (src/xagent/web/utils/json_payload_sanitizer.py:97-101) holds only narrowly; pre-PR json.dumps (C encoder) handled ≥20000 levels, so the newly exposed path is the websocket outbound row. Realism mitigates severity: websocket payloads are flat message dicts, and the staging path's pre-existing `serialize_value` recursion (trace_handlers.py:1064-1090, untouched) crashes at the same depth first.

**NC2 (minor, inline).** The migration rewrite round-trips matched rows through float64: `json.loads` (migration :342) silently and permanently truncates high-precision decimals on rows that carry both an escape and an external-writer decimal (0.123456789012345678901 → 0.12345678901234568), and 1e1000 → inf → json.dumps emits `Infinity` → the UPDATE CAST fails loudly but confusingly, with the docstring (:132-138) misattributing the failure to the ALTER. The ≥1e16 branch is a third silent corruption instance: `int(1e25)` ≠ 10^25. Suggest `parse_float=Decimal` (or loud rejection of non-round-trippable numbers) plus a docstring correction.

- `shrink:` src/xagent/web/utils/json_payload_sanitizer.py:110-112 — two regex passes over every dirty string (`_UNSTORABLE_CODE_POINTS.search()` then `.sub()`); `re.sub` returns the original object when nothing matches, so a single `cleaned = _UNSTORABLE_CODE_POINTS.sub(REPLACEMENT_CHARACTER, value)` plus `return _UNCHANGED if cleaned is value else cleaned` covers both cases in one pass. The migration copy (:233) already uses the single-pass form; this aligns the copies.

`net: -1 line`

## Blocking status

**Blocking: no — recommended event: APPROVE.** Nothing blocks: the prior blocker (C1) is verified fixed, and every remaining finding is minor. C3 remains open as a non-blocking carry-over (retention condition still unmeasurable; the hazardous-row drop path is still unpinned by real-PG coverage).

src/xagent/migrations/versions/20260813_trace_json_columns_to_jsonb.py
codeacme17: Fixed in `d679dcf`, and digging into it made the finding bigger than stated. I checked what `jsonb` actually does to numbers on PG 16:

```
1e16   -> 10000000000000000
1e25   -> 10000000000000000000000000        (10^25)
-0.0   -> 0.0
0.123456789012345678901 -> 0.123456789012345678901   (preserved in full)
1e1000 -> 1 followed by 1000 zeros
```

Its numbers are `numeric`, so a plain cast keeps the literal's **exact value** and changes only notation and Python type. That reframes all three of your sub-points as one problem: the rewrite was the only thing introducing a double, so a row it touched landed different numbers than an identical row it did not — silently, and permanently, since the rewritten row no longer matches the predicate.

Your `int(1e25) != 10^25` point lands harder than "a third instance" suggests. It is not that the value is imprecise in isolation — `int(1e25)` is exactly what that double is — it is that the *cast* would have produced 10^25, so the migration was actively diverging from its own no-op path.

Numbers now parse to `Decimal` and are emitted as their own literals, and I removed the migration's number normalization entirely rather than porting it to `Decimal`: `jsonb` normalizes on ingest anyway, and doing it identically for rewritten and untouched rows is the property actually worth having. `1e1000` no longer becomes `Infinity`, and the docstring no longer misattributes that failure to the ALTER. Two tests pin it — one row carrying all three shapes beside its escape, and a rewritten row asserted equal to an untouched one.

The app sanitizer keeps its normalization; it is a different job. There it holds a Python float with no literal to preserve, `int(1e25)` is that float's exact value, and the round trip stays stable because the hash is taken after normalization.

src/xagent/web/utils/json_payload_sanitizer.py
codeacme17: You are right, and the claim was mine to get right. Both callers treat a raising trace write as best-effort — `websocket.py` rolls back and logs, `trace_handlers.py` logs and continues — so a `RecursionError` there is a log line and a dropped row, not a loud failure.

Reworded in `d679dcf` to say exactly that, while keeping the reason recursion stays unbounded: a depth cap would have to leave everything below it unsanitized, which is the payload shape the function exists to keep out of the column. Also noted your mitigation, since it is the load-bearing part of why this is a stated boundary rather than a guard anything relies on — the staging path's own `serialize_value` recursion hits the limit first.

src/xagent/web/api/monitor.py
codeacme17: Closed in `d679dcf` rather than carried further — and thanks for the correction about `~?`, which changed what needed doing. Once the compile-only test and the three endpoint tests are accounted for, the only thing that genuinely lost real-PG coverage is the hazardous-row drop path, which is a much narrower target.

`TestReadGuardAgainstNativeJson` now builds a throwaway table with a native `json` column, seeds the shapes `jsonb` will not accept, and runs `get_json_field_expression` over the mix in one query — so the escaped-backslash stand-in, the pair strip and the alternation all execute against real PostgreSQL again. It covers the orphan-behind-a-literal case that motivates the stand-in, and asserts the benign rows (a valid pair, text that only looks like an escape) survive. Deliberately not the model's table: the point is to exercise the branch on the shape it was written for, and the model must stay `jsonb`.

The guard comment also gets the retirement condition you asked for, replacing the unmeasurable wording: it can go once skipping the 20260813 migration is no longer a supported upgrade path, with the `information_schema` query to check a given database and a pointer to the test to delete alongside it.

I had replied to this on the main thread last round asking which of your two options you preferred — that reply is at the PR level rather than inline, which is probably why it read as unanswered.

codeacme17:
Thanks for the approval. Everything still open is now addressed in `d679dcf` rather than deferred — nothing here seemed to warrant its own issue.

**NC2 was the one with teeth, and it is worse than the finding states.** Checking what `jsonb` actually does to numbers on PG 16 settled it: its numbers are `numeric`, so a plain cast keeps the literal's exact value and changes only notation and Python type — `1e25` becomes 10^25, and `0.123456789012345678901` survives in full. The rewrite was the only thing introducing a double, so a row it touched landed *different numbers* than an identical row it did not, silently and permanently. Your `int(1e25) != 10^25` point is the sharp end of that: not imprecision in isolation, but the migration diverging from its own no-op path.

Numbers now parse to `Decimal` and are emitted as their own literals, and I removed the migration's number normalization outright instead of porting it — `jsonb` normalizes on ingest anyway, and having it happen identically for rewritten and untouched rows is the property worth having. Two tests pin it: a row carrying all three shapes beside its escape, and a rewritten row asserted equal to an untouched one. The app sanitizer keeps its normalization; there it holds a float with no literal to preserve, so its round trip stays stable.

**NC1 — the claim was wrong and is now accurate.** Both callers treat trace writes as best-effort, so a `RecursionError` is a log line and a dropped row, not a loud failure. The reason recursion stays unbounded is unchanged and still stated.

**C3 — closed, not carried.** Your `~?` correction narrowed the target usefully: the only thing that genuinely lost real-PG coverage was the hazardous-row drop path. `TestReadGuardAgainstNativeJson` now runs the guard against a throwaway native-`json` table, so the escaped-backslash stand-in, the pair strip and the orphan-behind-a-literal case execute on real PostgreSQL again, with the benign rows asserted to survive. The guard comment carries a retirement condition with the `information_schema` query to check it and a pointer to the test to delete alongside it.

(My previous reply on this was at the PR level rather than inline, which is likely why it read as unanswered — my mistake for not putting it where the finding lived.)

**DC1 and DC3 — recorded where the next reader meets them.** The staging module now says why the sanitize-before-hash ordering rules out a `TypeDecorator`, and that a missed sanitize fails loudly on PostgreSQL but stores silently on SQLite. The migration docstring states that float-shaped pre-existing rows hit the same escalation as NUL rows with no pre-flight query of their own, and that the remedy is the same prune.

**Simplification applied** — `re.sub` returning the original object collapses the search-then-sub pair to one pass, matching the migration copy.

**Your cosmetic footnote is right too**: "non-recoverable verdict in lease recovery" was loosely attributed, since `resolve_checkpoint_recovery` is pointer-row-identity based and never decodes blobs. I have left the wording for now because correcting it precisely means naming three separate resume surfaces; if you would rather it be exact than brief, say so and I will spell it out.

231 tests across the affected suites pass against PostgreSQL 16; ruff and mypy clean; single alembic head.

PR #1320, opened 2026-08-13
feat(interactions): resolve whether a clarification draft is publishable

AlexLiu190625:
Resolves the publication question for a clarification draft attached to a finalizing run: given the task row, lease, and resume anchor a caller already holds, can this run's draft become a structured, durable interaction request?

The answer is one of three outcomes. It can be published, in which case the resolver hands back the exact arguments the interaction-staging primitive's `stage()` call needs. It can be not applicable for this round -- either because the run was never a waiting-for-user turn with a draft attached, or because a waiting draft exists but could not be safely turned into a payload (no resume anchor, or the payload could not fit the character and size limits). Or it can be fail-closed, which covers two situations with two different consequences for the round itself (see below): a draft turned up on a run that was not waiting, or the lease that produced this result can no longer speak for the task row -- unfenced, no longer the row's owner, or superseded by a later attempt.

Guard order is the contract. Checks run in a fixed sequence, from most specific to least: does the run's status agree with whether a draft is attached; is the lease fenced to a real run; does the task row still belong to that lease's runner and run; does the lease's attempt still match the task's current attempt; does a resume anchor exist; does the constructed payload fit the character domain and size limits. The size and character checks run last because they are the only ones that need the payload built first.

Three of the guards discard a run's settlement wholesale when they fail: an unfenced lease, a task row no longer owned by the lease that produced this result, and an attempt that is no longer current. Every other unsatisfied guard -- including a waiting run with no draft to publish, or a non-waiting run that stray-carries one -- means no structured row gets written this round, but the round itself is not discarded: whatever a finalizer already does for that situation keeps working exactly as it does today.

The payload's construction and parsing live together in one module, and reading it back always goes through the parser here rather than through ad hoc key access, so a future protocol version only needs one new branch. As long as the payload stays version 1, only optional fields may be added to it; renaming, retyping, or removing a field requires bumping the version and giving the parser an explicit branch for the old shape.

The idempotency key is derived from the draft's own content -- specifically the same turn marker the core draft type already produces -- rather than from the resume anchor. A resume anchor changes on every re-entrant checkpoint of the same turn, which would turn a replay that should cost nothing into a brand-new request each time. Deriving the key from content instead means the same question, asked again during the same turn, keeps requesting under the same key.

This change has no production caller. A static test enforces that directly, matching the equivalent gate already in place for the interaction-staging primitive it builds on. The module exists so that whichever change wires up the finalizers has a resolver to call; nothing in this change makes that call itself.

### DIFF
--- a/src/xagent/web/services/task_clarification_draft.py
+++ b/src/xagent/web/services/task_clarification_draft.py
@@ -301,9 +301,9 @@ def build_clarification_payload(draft: ClarificationDraft) -> dict[str, Any]:
     payload: dict[str, Any] = {
         "question": question,
         "interactions": interactions_payload,
-        "message_type": draft.message_type,
+        "message_type": _strip_control_characters(draft.message_type),
         "source": draft.source,
-        "requests": [item.to_dict() for item in draft.requests],
+        "requests": [_clean_leaves(item.to_dict()) for item in draft.requests],
     }
     if question_truncated:
         payload["question_truncated"] = True
@@ -407,9 +407,10 @@ def resolve_publishable_clarification(
        against; this degrades, it does not fail the round.
     5b. The payload built from the draft fits the character domain and size
         limits in :func:`build_clarification_payload`; if it does not
-        (empty after filtering, or still oversized after truncation), that
-        also degrades rather than fails. This check runs last because it is
-        the only one that requires building the payload first.
+        (empty after filtering, or reduced to only whitespace, or still
+        oversized after truncation), that also degrades rather than fails.
+        This check runs last because it is the only one that requires
+        building the payload first.
 
     A cross-run anchor mismatch is deliberately absent from this sequence:
     that check belongs to ``interaction_handoff`` (it already raises a
@@ -418,12 +419,30 @@ def resolve_publishable_clarification(
 
     On success, this function clears ``CLARIFICATION_DRAFT_MISSING`` --
     reaching ``Publishable`` is itself the evidence that the draft pipeline
-    worked this time. ``CLARIFICATION_MULTIPLE_DRAFTS`` is different: it
-    reports that a concurrent-waiting-step conflict happened *this round*,
-    which a successful publish does not undo, so it is registered (not
-    cleared) whenever ``result["clarification_superseded_step_ids"]`` is
-    non-empty, and left alone otherwise.
+    worked this time.
+
+    ``CLARIFICATION_MULTIPLE_DRAFTS`` is registered before any guard below
+    gets a chance to return, not only on the ``Publishable`` path: whether
+    ``result["clarification_superseded_step_ids"]`` is non-empty is a fact
+    about this round's ``result`` that holds regardless of which guard
+    ultimately classifies it -- a losing waiting step can be superseded by
+    an interrupt that wins the same round, so ``NotApplicable(None)`` and
+    every other non-``Publishable`` outcome must still be able to report
+    the conflict. Clearing it is the opposite: confined to the
+    ``Publishable`` path only, alongside ``CLARIFICATION_DRAFT_MISSING`` --
+    a successful publish clears the signal only when nothing was
+    superseded *this same round* either, since publishing this round's
+    draft does not retroactively undo a sibling step's draft having been
+    discarded.
     """
+
+    superseded_step_ids = result.get("clarification_superseded_step_ids") or []
+    if superseded_step_ids:
+        register_degradation(
+            CLARIFICATION_MULTIPLE_DRAFTS,
+            f"task {task.id}: run {lease.run_id}: superseded step ids "
+            f"{list(superseded_step_ids)}",
+        )
 
     status = result.get("status")
     draft = result.get("clarification_draft")
@@ -466,21 +485,14 @@ def resolve_publishable_clarification(
 
     payload = build_clarification_payload(draft)
 
-    if not payload["question"]:
+    if not payload["question"].strip():
         return NotApplicable("empty_question")
 
     if _serialized_byte_length(payload) > _TOTAL_PAYLOAD_MAX_BYTES:
         return NotApplicable("payload_too_large")
 
     clear_degradation(CLARIFICATION_DRAFT_MISSING)
-    superseded_step_ids = result.get("clarification_superseded_step_ids") or []
-    if superseded_step_ids:
-        register_degradation(
-            CLARIFICATION_MULTIPLE_DRAFTS,
-            f"task {task.id}: run {lease.run_id}: superseded step ids "
-            f"{list(superseded_step_ids)}",
-        )
-    else:
+    if not superseded_step_ids:
         clear_degradation(CLARIFICATION_MULTIPLE_DRAFTS)
 
     return Publishable(
--- a/src/xagent/web/services/task_clarification_draft.py
+++ b/src/xagent/web/services/task_clarification_draft.py
@@ -327,11 +327,27 @@ def parse_clarification_payload(
     interactions)`` shape ``get_latest_waiting_question``
     (``chat_history_service.py``) already returns for the legacy transcript
     path -- a cross-module contract with no shared base class to enforce it,
-    pinned by ``test_payload_round_trip_matches_the_legacy_reader_shape`` in
-    ``tests/web/services/test_task_clarification_draft.py``, which calls
+    pinned by ``test_payload_round_trip_matches_the_legacy_reader_shape*``
+    in ``tests/web/services/test_task_clarification_draft.py``, which calls
     the real ``get_latest_waiting_question`` and compares its return shape
     against this function's, rather than re-deriving the expectation from
     this function's own return type.
+
+    ``interactions`` comes back ``None``, not ``[]``, whenever the payload
+    carries no interaction items. ``get_latest_waiting_question`` itself is
+    not this consistent: it returns ``None`` for a persisted row whose
+    ``interactions`` column is ``NULL`` (the shape a ``send_message``-
+    sourced draft always produces, since that source's payload never has
+    any interactions to carry), but it returns ``[]`` for a row whose
+    column holds an actual empty list -- the shape an empty-form
+    ``ask_user_question`` call produces (a legal call: ``ask_user_question``
+    is classified by whether its request carries an ``interactions`` key at
+    all, not by whether that key's value is non-empty; see
+    ``draft_from_waiting_request``). That ``NULL``-vs-``[]`` split on the
+    legacy side is a known, un-reconciled divergence, not something this
+    function reproduces: every empty case collapses to ``None`` here,
+    deliberately, so a caller checking "did this turn have interactions"
+    never has to handle two different falsy shapes.
 
     This is the one place allowed to know what a payload's version implies:
     a reader elsewhere in the codebase must call this function rather than
@@ -344,7 +360,7 @@ def parse_clarification_payload(
     interactions = payload.get("interactions")
     return (
         question if isinstance(question, str) else None,
-        interactions if isinstance(interactions, list) else None,
+        interactions if isinstance(interactions, list) and interactions else None,
     )
 
 
--- a/src/xagent/web/services/task_clarification_draft.py
+++ b/src/xagent/web/services/task_clarification_draft.py
@@ -1,0 +1,438 @@
+"""Decide whether a clarification draft attached to a finalizing run can be
+published as a structured interaction request, and build/parse the JSON
+payload a published row carries.
+
+This module answers one question: given the result dict a pattern run just
+produced, the task row and lease the caller already holds, and the resume
+anchor the caller already resolved, does the clarification draft attached to
+that result (if any) qualify to become a durable, structured interaction
+row? The answer is one of three outcomes, never folded into a boolean:
+
+* :class:`Publishable` -- the draft passed every precondition and is shaped
+  into the three keyword arguments ``interaction_handoff.stage()`` expects
+  (``request_payload``, ``request_idempotency_key``, ``expires_at``).
+* :class:`NotApplicable` -- nothing durable should be written this round,
+  but nothing failed either: either this run never produced a clarification
+  in the first place, or a waiting run's draft could not be turned into a
+  publishable payload (no resume anchor, or the payload could not be shaped
+  within the size and character limits below). The caller's own settlement
+  proceeds exactly as it would have before this module existed.
+* :class:`FailClosed` -- the caller must discard this round's settlement
+  wholesale, the same way it already discards a late, no-longer-owned
+  result. This happens only when the lease that produced the result can no
+  longer be trusted to speak for the task row.
+
+Guard order is a contract: the checks inside :func:`resolve_publishable_clarification`
+run in a fixed sequence, most specific and least data-exposing failure
+first, so that the classification a caller observes always names the first
+thing that actually went wrong -- never a later, coincidentally-true guard
+that happens to fail on the same input for an unrelated reason.
+
+Callers. This function's caller set is exactly the finalizers that settle a
+task run against its lease and load the task row under a lock: nothing else
+holds the combination of task, lease and anchor this function's guards
+depend on. A caller reached from anywhere else -- a builder chat session, a
+nested sub-agent's own turn -- was never going to be a caller in the first
+place, and this module makes no claim about what happens there. Extending
+the caller set to a fourth site is a reason to come back and re-read this
+module, not to assume its guards still describe the new caller's situation.
+
+This module is also the only place allowed to construct or parse the JSON
+shape a published row's ``request_payload`` carries -- see
+:func:`build_clarification_payload` and :func:`parse_clarification_payload`
+for the pairing rule and the version-evolution promise that comes with it.
+
+An application-layer invariant worth stating plainly: a task can have at
+most one *active* interaction row at a time (the database enforces this
+with a unique constraint on the active slot). Nothing in this module
+inserts a row -- that is the caller's job, via ``stage()`` -- but every
+guard below exists to make sure that when the caller does insert one, it is
+the single row this run is entitled to produce.
+
+One more precondition lives outside this module entirely and cannot be
+checked at run time: whichever deployment decides to start publishing
+structured rows for a given ``Task.source`` value needs that value to
+already appear in the model's ``INTERACTION_ORIGIN_VOCABULARY`` (see
+``xagent.web.models.task_interaction``). A source value outside that
+vocabulary is not rejected here -- it degrades one layer down, inside
+``interaction_handoff``, as an unrecognized origin swallowed into a generic
+degradation signal. Confirming vocabulary coverage against a live
+deployment's actual ``tasks.source`` values before turning publication on
+is the caller's obligation, not this module's.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta
+from typing import Any, Mapping
+
+from ...core.agent.clarification import ClarificationDraft
+from ..models.task import Task
+from .ops_signals import (
+    CLARIFICATION_DRAFT_MISSING,
+    CLARIFICATION_MULTIPLE_DRAFTS,
+    clear_degradation,
+    register_degradation,
+)
+from .task_interaction_staging import InteractionAnchor
+from .task_lease_service import TaskLease
+
+logger = logging.getLogger(__name__)
+
+# Deployment policy, not an invariant: this only bounds how long an
+# unanswered row sits before a reclaim job retires it. The read surface
+# does not consult ``expires_at`` at all, so a longer-lived row is still
+# shown to whoever is waiting on it; widen or narrow this once real
+# publication volume gives a basis for the number, not before.
+CLARIFICATION_REQUEST_TTL = timedelta(hours=24)
+
+# The character domain a payload's free-text leaves may contain, kept
+# byte-for-byte identical to ``xagent.core.agent.clarification``'s
+# ``_marker_clean`` whitelist. That module cannot import this one (core
+# layer never imports web layer), so the two copies cannot share code --
+# only this comment, on both sides, keeps them from drifting apart. If one
+# domain ever changes, change the other in the same commit.
+_CONTROL_CHAR_KEEP = "\n\r\t"
+
+# Three independent size ceilings, checked in this order because each one
+# can only be evaluated once the field(s) before it are already final.
+_QUESTION_MAX_BYTES = 16 * 1024
+_INTERACTIONS_MAX_BYTES = 64 * 1024
+_TOTAL_PAYLOAD_MAX_BYTES = 128 * 1024
+
+_QUESTION_TRUNCATION_SUFFIX = "\n\n[question truncated]"
+
+
+def _strip_control_characters(value: str) -> str:
+    """Remove NUL and C0 control characters, keeping the same three
+    whitespace exceptions ``_marker_clean`` keeps. See the module-level
+    comment above for the cross-layer contract this filter must not drift
+    from."""
+
+    cleaned = value.replace("\x00", "").replace("\u0000", "")
+    return "".join(ch for ch in cleaned if ord(ch) >= 32 or ch in _CONTROL_CHAR_KEEP)
+
+
+def _clean_leaves(value: Any) -> Any:
+    """Recursively apply :func:`_strip_control_characters` to every string
+    leaf inside a JSON-like structure, leaving every other value untouched."""
+
+    if isinstance(value, str):
+        return _strip_control_characters(value)
+    if isinstance(value, dict):
+        return {key: _clean_leaves(item) for key, item in value.items()}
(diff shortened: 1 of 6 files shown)


src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+        suffix_bytes = len(_QUESTION_TRUNCATION_SUFFIX.encode("utf-8"))
+        question = (
+            _truncate_to_byte_limit(
+                cleaned_question, _QUESTION_MAX_BYTES - suffix_bytes
+            )
+            + _QUESTION_TRUNCATION_SUFFIX
+        )
+        question_truncated = True
+
+    interactions_cleaned = [_clean_leaves(dict(item)) for item in draft.interactions]
```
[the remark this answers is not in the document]
AlexLiu190625: Not adopting, though the redundancy half of this is correct: `_clean_leaves` does return a fresh dict for dict inputs, so `dict(item)` adds nothing as mutation protection.

It stays for two deliberate reasons:

1. **The fragility premise doesn't hold here.** `draft.interactions` is typed `tuple[dict[str, Any], ...]` on the core draft type, and every producer builds it in-process from normalized data whose non-dict elements are already filtered out — no external or model-generated value reaches this comprehension. If a non-dict element ever did appear, it would be a programming error upstream, and this codebase's convention for broken internal invariants is to fail loudly at the site rather than pass the value through. The suggested change would silently carry a malformed element into the persisted payload instead of crashing a test the moment the invariant breaks.

2. `dict(item)` also normalizes any Mapping implementation to a plain dict before the payload goes through `json.dumps` size accounting and storage — a cheap guarantee that what is measured and persisted is exactly a plain-dict tree.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+    reaching ``Publishable`` is itself the evidence that the draft pipeline
+    worked this time. ``CLARIFICATION_MULTIPLE_DRAFTS`` is different: it
+    reports that a concurrent-waiting-step conflict happened *this round*,
+    which a successful publish does not undo, so it is registered (not
+    cleared) whenever ``result["clarification_superseded_step_ids"]`` is
+    non-empty, and left alone otherwise.
+    """
+
+    status = result.get("status")
+    draft = result.get("clarification_draft")
```
rogercloud, line 483: This `result.get("clarification_draft")` is never type-checked before use: a plain-dict draft — documented to exist in `AutoPattern.last_result` restored snapshots (clarification.py:10-15; execution_adapter.py:407 copies it verbatim) — would raise AttributeError inside a finalizer, and the sibling staging primitive guards isinstance for anchors (test_interaction_staging.py:478-482). Suggest `if draft is not None and not isinstance(draft, ClarificationDraft): return FailClosed("draft_status_mismatch")` plus a dict-draft test.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+    if _serialized_byte_length(interactions_cleaned) > _INTERACTIONS_MAX_BYTES:
+        interactions_payload = []
+        interactions_dropped = True
+
+    payload: dict[str, Any] = {
+        "question": question,
+        "interactions": interactions_payload,
+        "message_type": draft.message_type,
+        "source": draft.source,
+        "requests": [item.to_dict() for item in draft.requests],
```
rogercloud: The character-domain filter covers question (line 270) and interactions leaves (line 294) but not message_type, source, or these `item.to_dict()` requests leaves (tool_name/tool_call_id/interaction_id) — yet the same IDs are filtered via `_marker_clean` when `_compose_turn_marker` builds the idempotency key (clarification.py:187-197), so payload and key treat identical data under different domains. Suggest applying `_clean_leaves` to each `item.to_dict()` (or once to the whole payload); this cannot perturb the idempotency key (derived from turn_marker only).

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+    payload = build_clarification_payload(draft)
+
+    if not payload["question"]:
+        return NotApplicable("empty_question")
+
+    if _serialized_byte_length(payload) > _TOTAL_PAYLOAD_MAX_BYTES:
+        return NotApplicable("payload_too_large")
+
+    clear_degradation(CLARIFICATION_DRAFT_MISSING)
+    superseded_step_ids = result.get("clarification_superseded_step_ids") or []
```
rogercloud: `CLARIFICATION_MULTIPLE_DRAFTS` is registered only on the Publishable path, but the resolver docstring (421-425) promises registration whenever `clarification_superseded_step_ids` is non-empty and ops_signals.py:63-68 says it fires whenever a losing step's draft was discarded — the discard happens in dag.py regardless of winner publishability, and interrupted (non-waiting) results that surface the key (execution_adapter.py:417-428) return NotApplicable(None) at 450-451 first. Suggest hoisting only the register branch (477-482) before the early returns, keeping the else-clear on the Publishable path.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+
+    if lease.attempt_id is not None and task.lease_attempt_id != lease.attempt_id:
+        return FailClosed("attempt_mismatch")
+
+    if anchor is None:
+        return NotApplicable("no_anchor")
+
+    payload = build_clarification_payload(draft)
+
+    if not payload["question"]:
```
rogercloud: `if not payload["question"]` passes whitespace-only questions ('   ' is truthy): a structured row would publish while the legacy transcript path drops the same content (chat_history_service.py:506-508 strips and returns None). Suggest `if not payload["question"].strip(): return NotApplicable("empty_question")` + test.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+    which a successful publish does not undo, so it is registered (not
+    cleared) whenever ``result["clarification_superseded_step_ids"]`` is
+    non-empty, and left alone otherwise.
+    """
+
+    status = result.get("status")
+    draft = result.get("clarification_draft")
+    is_waiting = status == "waiting_for_user"
+
+    if is_waiting and draft is None:
```
rogercloud, line 486: This waiting-without-draft case is documented as benignly reachable (clarification.py:233-248: empty-message send_message with expect_response=True; old-schema checkpoints degrade to no draft), yet it logs at `logger.error` and permanently latches `CLARIFICATION_DRAFT_MISSING` (clears only on a later Publishable at line 475). House convention for non-discarding degrades is warning/info (task_lease_service.py:342-352, interaction_rollout.py:504-519), and the producer logs the same event at warning (clarification.py:261-264). Suggest downgrading to warning and/or documenting the latching consequence in the ops_signals comment.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+    ``reason`` is ``None`` for the ordinary case -- this finalization was
+    never a waiting-for-user turn with a draft attached, so publication
+    simply does not apply, and the caller's settlement proceeds exactly as
+    it would without this module. ``reason`` is one of ``"no_anchor"``,
+    ``"payload_too_large"``, or ``"empty_question"`` when a waiting turn's
+    draft existed but could not be shaped into a publishable payload; each
+    of those is a degrade, not a failure -- the run is not discarded.
+    """
+
+    reason: str | None
```
rogercloud: `NotApplicable.reason` (and `FailClosed.reason` at line 215) are plain `str`/`str | None` against their documented closed value sets, and future finalizers will dispatch on these strings to pick discard-wholesale vs round-proceeds. `Literal` is the house style for closed reason vocabularies (widget_domains.py:18, interaction_rollout.py:220, task_runtime.py:240), and mypy runs in pre-commit. Suggest Literal unions.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -0,0 +1,490 @@
...
+_QUESTION_TRUNCATION_SUFFIX = "\n\n[question truncated]"
+
+
+def _strip_control_characters(value: str) -> str:
+    """Remove NUL and C0 control characters, keeping the same three
+    whitespace exceptions ``_marker_clean`` keeps. See the module-level
+    comment above for the cross-layer contract this filter must not drift
+    from."""
+
+    cleaned = value.replace("\x00", "").replace("\u0000", "")
```
rogercloud, line 150: `"\x00"` and `"\u0000"` are the same U+0000, so the second `.replace()` is a no-op and the join-comprehension already drops ord<32. Both replaces can be deleted (or kept with `_marker_clean`'s explanatory mirror comment if you want the parity explicit).

rogercloud (approved):
## Re-review of PR #1320 — feat(interactions): resolve whether a clarification draft is publishable

### Summary
This PR adds `src/xagent/web/services/task_clarification_draft.py` (+490): a pure resolver that classifies one finalizing run's clarification draft into three outcomes — Publishable (exact `stage()` kwargs: `request_payload`, `request_idempotency_key`, `expires_at`), NotApplicable (no row written; the round settles as before), or FailClosed (five reasons split into discard-wholesale vs round-proceeds consequences). It fixes a guard-order contract (status/draft pairing, lease fence, row ownership, attempt currency, anchor presence, payload limits), a versioned v1 payload codec with parser-only reads, an idempotency key derived from the draft's turn_marker (not the resume anchor), and two ops_signals degradation signals (`CLARIFICATION_DRAFT_MISSING`, `CLARIFICATION_MULTIPLE_DRAFTS`). Three new test files (+1182) cover behavior — including real `get_latest_waiting_question` round-trips and an interaction_handoff cross-run degrade — plus AST static gates asserting zero production callers and capability-boundary guards pinning the anchor-trust contract. Blocking: no — recommended event: APPROVE.

### Update summary
No new commits since the prior review: all 7 commits were pushed at 04:09Z and gemini-code-assist[bot] reviewed the current head (c7ea5aac) at 2026-08-13T05:55Z. This follow-up layers an independent blank-slate design pass and a full re-verification of every claim on top of that review rather than re-litigating its findings; the prior finding was re-checked against the same head and carries over below.

### Approach verdict (design pass): Sound
Consistent, precedented incremental slice of the documented native-interaction pipeline. The callerless-shipping question is answered correctly by the codebase's own established slice-and-gate delivery model — model, staging primitive, anchor resolver, and rollout gate all shipped the same way. The AST gate enforcing zero production callers keeps the resolver dead-code-safe until wiring, so everything below is hardening rather than blocking.

### Prior-findings checklist
- **P1 [WAIVED]** — gemini-code-assist[bot] flagged `dict(item)` at `task_clarification_draft.py:294` as redundant and fragile. The author declined with a technically correct defense (verified): `draft.interactions` is typed `tuple[dict[str, Any], ...]` (clarification.py:86), every producer normalizes entries via `_normalize_ask_user_interactions` (react.py:139-143) or omits the key, no persisted version ever stored un-normalized entries, and `dict(item)` normalizes Mapping→dict before `json.dumps`. Status: WAIVED — the author's reply to the bot's inline comment explains it, and the defense verifies correct.

### Findings (all minor, ordered by impact)

1. **[minor, hardening] `task_clarification_draft.py:429`** — `result.get("clarification_draft")` is never type-checked before use. A plain-dict draft — documented to exist in `AutoPattern.last_result` restored snapshots (clarification.py:10-15; execution_adapter.py:407 copies verbatim) — would raise AttributeError inside a finalizer. The sibling staging primitive guards isinstance for anchors (test_interaction_staging.py:478-482). Fix: `if draft is not None and not isinstance(draft, ClarificationDraft): return FailClosed("draft_status_mismatch")` plus a dict-draft test. Cheap hardening that matches the module's never-raise ethos.

2. **[minor, design-level] `task_clarification_draft.py:167-219`** — `fail_closed_reason` flattens two axes under one name. Only 3 of the 5 FailClosed reasons mean discard-wholesale; `missing_draft`/`draft_status_mismatch` do not, and neither does any NotApplicable reason — yet the uniform accessor invites `fail_closed_reason is not None → discard`, which would wrongly discard those rounds. The discard-vs-keep mapping lives only in prose. Fix: expose the actual dispatch axis (e.g. `discards_settlement: bool`, True only for the three discard reasons) so the three finalizers get a data contract, not a docstring contract.

3. **[minor, design-level, wiring-time] `task_clarification_draft.py:455-462`** — guards 2-4 restate fences the finalizers already hold inline (websocket.py:2183-2195, 2876-2890: unfenced refusal + `SELECT ... FOR UPDATE`) and staging re-asserts (task_interaction_staging.py:1194-1201); nothing pins the three copies together. At wiring time this bites: `finalize_managed_task_lease_result` releases the lease (runner_id/lease_attempt_id NULLed, task_lease_service.py:930,937) before any natural resolver insertion point, so guard 3 would spuriously FailClosed('ownership_changed'). Fix: at/before wiring, extract a shared fencing predicate or add an AST-level guard-equivalence pin (the PR already uses AST boundary tests), and document the FOR UPDATE load-before-release requirement.

4. **[minor] `task_clarification_draft.py:476-484`** — `CLARIFICATION_MULTIPLE_DRAFTS` is registered only on the Publishable path, but the resolver docstring (421-425) promises registration "whenever `result['clarification_superseded_step_ids']` is non-empty" and ops_signals.py:63-68 says it fires whenever a losing step's draft was discarded — the discard happens in dag.py regardless of winner publishability, and execution_adapter.py:417-428 deliberately surfaces the key on interrupted (non-waiting) results, which return NotApplicable(None) at 450-451 first. Fix: hoist only the register branch (477-482) before the early returns; keep the else-clear tied to the Publishable path (clearing on non-publishable resolutions would break the "not cleared by the very publish that reports it" semantics).

5. **[minor] `task_clarification_draft.py:304-306`** — the character-domain filter covers question (270) and interactions leaves (294) but not message_type (304), source (305), or the requests leaves (306, via `ClarificationRequestItem.to_dict`: tool_name/tool_call_id/interaction_id). The same IDs ARE filtered via `_marker_clean` when `_compose_turn_marker` builds the idempotency key (clarification.py:187-197), so payload and key treat identical data under different domains. Fix: apply `_clean_leaves` to each `item.to_dict()` (or once to the whole payload); this cannot perturb the idempotency key (derived from turn_marker only).

6. **[minor] `task_clarification_draft.py:469`** — `if not payload["question"]` passes whitespace-only questions ('   ' is truthy): a structured row would publish while the legacy transcript path drops the same content (chat_history_service.py:506-508 strips and returns None). Fix: `if not payload["question"].strip(): return NotApplicable("empty_question")` + test.

7. **[minor] `task_clarification_draft.py:432-441`** — waiting-without-draft is documented as benignly reachable (clarification.py:233-248: empty-message send_message with expect_response=True; old-schema checkpoints degrade to no draft) but logs at `logger.error` and permanently latches `CLARIFICATION_DRAFT_MISSING` (process-global; clears only on a later Publishable at 475). House convention for non-discarding degrades is warning/info (task_lease_service.py:342-352, interaction_rollout.py:504-519), and the producer logs the same event at warning (clarification.py:261-264). Fix: downgrade to warning and/or document the latching consequence in the ops_signals comment.

8. **[minor] `task_clarification_draft.py:192,215`** — `NotApplicable.reason: str | None` and `FailClosed.reason: str` are untyped against their documented closed sets (3 and 5 values); future finalizers will dispatch on these strings to pick discard-wholesale vs round-proceeds. `Literal` is the house style for closed reason vocabularies (widget_domains.py:18, interaction_rollout.py:220, task_runtime.py:240), and mypy runs in pre-commit, so a typo would be caught. Fix: Literal unions.

9. **[minor] `tests/web/services/test_clarification_publication_gate.py:50`** — `GATED_NAMES = {resolve_publishable_clarification}` only: a production `from ...task_clarification_draft import parse_clarification_payload` passes the gate untouched, yet the gate docstring claims the read surface "is switched on together" and the module mandates the parser as the only sanctioned read path (task_clarification_draft.py:352-356). The parser-only rule has no pinning test of any kind. Fix: extend GATED_NAMES to `parse_clarification_payload` (and `build_clarification_payload`), or narrow the docstring claim to the writer side. (Note: the parser's version-evolution signature claim itself was refuted — an optional `protocol_version: int = 1` parameter is backward-compatible; only the documentation-only-enforcement half stands.)

10. **[minor] `tests/web/services/test_clarification_publication_gate.py:9-14`** — the docstring's "exactly three" removal condition overstates readiness: `finalize_managed_task_lease_result` (managed_task_lease.py:32-40) receives only status/assistant_content/interactions/message_type — no result dict, no anchor — and releases the lease as its FIRST DB op (managed_task_lease.py:50) before reloading the task row (54-55). Fix: soften the docstring to note the result-dict + anchor plumbing prerequisite. (The release-before-resolve ordering hazard itself is captured in finding 3.)

11. **[minor, doc] `task_clarification_draft.py:371`** — the `now` parameter's aware-UTC obligation is undocumented; a naive now yields a naive expires_at and `stage()` raises an unswallowed ValueError (task_interaction_staging.py:711-723; _SWALLOWED at 432-438). Fix: one sentence mirroring `stage_interaction_request`'s wording. Do not add resolver-side validation — staging already validates, and the module precedent is documenting non-runtime-checkable caller obligations.

12. **[minor, design-level, optional] `task_clarification_draft.py:119,130-137`** — `_strip_control_characters` is the FIFTH re-typed C0 whitelist copy; the layering comment (111-118) slightly overstates: web→core import is legal and the module already imports `ClarificationDraft` from core (line 89), so the web copy could import a publicized core constant instead of re-typing it. Mitigation acknowledged: re-typing is the established house pattern (all three web copies inline the literal), and the author went further than predecessors (named constant + drift test calling the real `_marker_clean`). Optional consolidation, not a change request.

13. **[minor] `tests/web/services/test_task_clarification_draft.py`** — three untested edges, each worth a test: (a) mid-multibyte UTF-8 truncation — the only truncation test uses ASCII (516-530), leaving `_truncate_to_byte_limit`'s `errors='ignore'` partial-tail drop and byte-vs-char slicing unpinned (the (limit-suffix) bound half is already caught by the ASCII test's 16KiB assertion); (b) control chars inside interactions leaves flowing through `build_clarification_payload` — the drift test (572-595) probes the filter directly, not the assembled payload; (c) guard-1-vs-guard-2 precedence — no test combines waiting+no-draft+run_id=None; `missing_draft` does not discard the round while `unfenced_lease` does, so a guard swap changes caller behavior with all tests green.

- L136: shrink duplicate NUL removal — `"\x00"` and `"\u0000"` are the same U+0000; the second `.replace()` is a no-op and the join-comprehension already drops ord<32. Delete both replaces; or keep them only with `_marker_clean`'s explanatory mirror comment.
- L176/193/218: delete `fail_closed_reason` None-fields + property machinery — zero production readers; `isinstance(resolution, FailClosed)` + `.reason` covers the same cases. (Tied to finding 2: if the design fix adds `discards_settlement`, this machinery goes away as part of it.)
- L315 (also 378, 452, 684): shrink four copies of the sqlite engine/sessionmaker/try-finally boilerplate into one `tmp_path`-parameterized fixture — ~20 lines saved; repo fixture precedent exists (test_interaction_rollout_gate.py:70-88).
- Net: ~-25 lines possible.

Blocking: no — no critical or major confirmed findings; all 13 findings are minor, the sole prior finding is WAIVED (never blocking), and the simplification items are optional. Recommended event: APPROVE.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed in 0540aefc2. The register branch now runs at the top of the resolver, before every early return, whenever `clarification_superseded_step_ids` is non-empty — covering the interrupted-winner path you named and also the waiting-path degrades (no anchor, empty question, oversized) and all three fail-closed exits. The clear stays on the Publishable path only, so a single call never both registers and clears, and the non-clearing comment in ops_signals keeps holding. The resolver docstring now states the actual trigger. Two new tests pin the interrupted-winner and no-anchor cases; reverting the lift turns exactly those two red with everything else green (no prior test asserted non-registration).

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed in 0540aefc2 for the two fields that carry model-controlled data: each `item.to_dict()` in `requests` now goes through `_clean_leaves`, and `message_type` — which reaches the draft as a raw tool-call argument via `str(args.get("message_type", "info"))` with the schema enum being advisory only — is now filtered with `_strip_control_characters`. Tests pin both, including that the idempotency key is unaffected (it derives from the turn marker only). `source` is left unfiltered deliberately: it is an internal closed vocabulary (three literal construction branches, guarded by `CLARIFICATION_SOURCES`), never model-controlled.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed in 0540aefc2: the emptiness check is now `if not payload["question"].strip()`, so a whitespace-only question degrades to `NotApplicable("empty_question")` instead of publishing a blank structured row that the transcript path would have dropped. Test added; reverting the strip turns it red with a Publishable leaking through.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed in 6a8a5a9e3: both reason fields are now Literal unions (`NotApplicableReason`, `FailClosedReason`), enumerated against every construction site in the module — mypy (whole-package, in pre-commit) rejects an out-of-set value, verified by mutation. String-comparison reads keep working unchanged since Literal is `str` at runtime.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Took the second option in 66ea8f6ee: the no-op second replace stays, and the comment now explains why honestly — it keeps this filter's character-handling semantics aligned with the core-layer copy for the five-copy comparison, while noting the two copies are not written alike line-for-line (that one uses two statements, this one chains them).

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Not adopting. The plain-dict draft shape you cite lives only in `AutoPattern.last_result` restored snapshots, and that field has no reader anywhere outside auto.py itself — it cannot reach the result dict a finalizer hands to this resolver, whose producers (`execution_adapter._normalize_result`) forward the live `ClarificationDraft` object or None. If a non-draft object ever did arrive here, that would be a broken internal invariant, and this module's position — the same one the round's waived comment settled — is that a broken invariant should fail loudly at the site rather than be classified into a degradation outcome that hides it. Happy to revisit if a real producer of dict-shaped drafts appears on this path.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Deferring this one to the wiring change, recorded on its obligations list: the module currently has three `logger.error` sites on non-discarding degrade paths (missing draft, empty-after-filter, oversized), so changing only this one would leave the module internally inconsistent — the three will be leveled together, alongside the signal-latching consequence documentation, when the finalizer-side clearing ownership is settled. The registration-scope half of this area was fixed this round (see the thread on the register hoist).

AlexLiu190625:
Follow-up on the review body's numbered findings that have no inline thread — all thirteen were triaged, none silently dropped. Fixes are in the three commits now on the branch (0540aefc2, 6a8a5a9e3, 66ea8f6ee); the inline threads carry their own replies.

**Fixed this round:**
- **9 (gate coverage)**: `GATED_NAMES` now also pins `parse_clarification_payload` and `build_clarification_payload`, so a direct named import can no longer slip past the zero-production-caller gate.
- **10 (gate docstring readiness)**: the removal-condition docstring now states the prerequisite you identified — none of today's three finalizers has the result-dict + anchor plumbing yet.
- **11 (`now` obligation)**: documented as aware-UTC, mirroring the staging primitive's wording; no resolver-side re-validation added, per the module's convention for caller obligations that staging already enforces.
- **13 (test edges)**: all three added — multi-byte truncation on a real CJK payload, control characters through full payload assembly, and the guard-1-vs-guard-2 precedence pin (the swap now goes red).

**Deferred to the wiring change, tracked on its obligations list:**
- **2 (`discards_settlement` as a data contract)**: agreed on the axis; it lands together with the finalizer exits that will be its only consumers, rather than as a field with zero readers today.
- **3 (fence-copy pinning + release-before-resolve ordering)**: the resolver docstring now carries the hard timing rule ("call before releasing the lease, on the row read under FOR UPDATE"); the shared-predicate/AST-equivalence pinning across the three fence copies is deferred to wiring, where the third copy's call sites exist.

**Declined:**
- **12 (fifth control-character copy)**: taking your own framing that this is optional; promoting a core-layer public constant is a cross-layer change this PR shouldn't carry, and this copy already ships with a named constant plus a drift test against the real core filter.
- **Simplification 2 (drop the `fail_closed_reason` machinery)**: that uniform accessor is the read shape the planned finalizer callers dispatch on, so it stays.
- **Simplification 3 (sqlite fixture consolidation)**: skipped on an approved PR to keep the re-read surface small; happy to fold it into the next change that touches these files.

PR #1323, opened 2026-08-13
fix(mcp): start the per-server OAuth flow for custom mcp_oauth connectors

### LINKED ISSUE #1313: Connector picker dead-ends for custom MCP servers using mcp_oauth: "This app is not configured for connection"
## Bug

A user-created custom MCP server with `transport: streamable_http` and `auth: {"type": "mcp_oauth"}` that has no completed OAuth grant cannot be connected from the "+ Connector" picker dialog. Clicking the server card opens the detail modal, and its Connect button always fails with the toast "This app is not configured for connection". The dialog offers no way to start the OAuth authorization, so the server can never be attached to an agent from this entry point.

## Steps to reproduce

1. Create a custom MCP server with transport `streamable_http` and auth type `mcp_oauth`, and do not complete the OAuth authorization.
2. Open the connector picker from an agent ("+ Connector") and switch Location to Local.
3. The server appears without the connected checkmark. Click it, then click "Connect" in the detail modal.
4. A toast appears: "This app is not configured for connection". There is no path forward.

## Analysis

- The `location=local` branch of `GET /api/mcp/apps` builds custom-server entries without an `auth_type` field (`src/xagent/web/api/mcp.py`).
- `handleConnectApp` in `frontend/src/components/mcp/connect-mcp-dialog.tsx` dispatches on `auth_type` (`builtin_oauth` / `api_key` / `keyless` / `mcp_oauth`). An entry with no `auth_type` falls through to the mis-authored-entry toast (`tools.mcp.alerts.notConfigured`), so the `mcp_oauth` branch is unreachable for custom servers.
- This became user-visible after #1097: `is_connected` for an mcp_oauth server now (correctly) requires an active grant, so an unauthorized custom server is no longer directly selectable and is routed to the Connect button instead. Before that change, local entries were reported as connected unconditionally and the dead end was unreachable.
- A second blocker sits behind the first: the picker's mcp_oauth handler (`handleConnectMcpOAuthApp`) posts to `/api/mcp/apps/{id}/oauth/connect`, which resolves catalog apps. Custom servers need the per-server flow (`/api/mcp/{server_id}/oauth/connect`), which the custom MCP edit form already uses.

## Expected behavior

For a custom mcp_oauth server, the picker's Connect button starts the per-server MCP OAuth flow (discovery plus consent popup) that the custom MCP edit form already implements, instead of dead-ending.

## Scope note

This issue covers the interactive flow only: the Connect button must work for editors who authorize personally. The separate problem that the picker's connected-state gate blocks attaching connectors whose tokens are supplied through `set_oauth_token_resolver_hook` (where no personal grant ever exists) is tracked in #1332.

codeacme17:
Fixes #1313

## Problem

A user-created MCP server with transport `streamable_http` and `auth: {"type": "mcp_oauth"}` could not be connected from the "+ Connector" picker. Its Connect button always failed with "This app is not configured for connection", and the dialog offered no way to start the authorization.

Two blockers stacked here:

1. The `location=local` branch of `GET /api/mcp/apps` built custom-server entries without an `auth_type` field, and `handleConnectApp` in the picker dispatches Connect on exactly that field — so the `mcp_oauth` branch was unreachable and every custom server fell through to the mis-authored-entry toast (`tools.mcp.alerts.notConfigured`).
2. Behind it, the picker's `mcp_oauth` handler posts to `/api/mcp/apps/{id}/oauth/connect`, which starts with `_ensure_catalog_mcp_oauth_server()` and so resolves catalog apps only — it can never see a user-added server.

The dead end became reachable after #1097 made `is_connected` for an mcp_oauth server require an active grant. Before that, local entries were reported connected unconditionally, so the Connect button never rendered.

## Changes

**Backend** (`src/xagent/web/api/mcp.py`)

- Extract `_is_mcp_oauth_server()` so the connection-state gate and the picker's new `auth_type` hint share one shape check and cannot drift.
- Emit `auth_type: "mcp_oauth"` on local entries for the mcp_oauth shape.

The hint is deliberately scoped to that shape. Every other custom shape is already reported connected unconditionally (so its Connect button never renders), and tagging those rows with a catalog classification would repoint the settings dialog's Configure button away from the custom edit form — to the key dialog for `api_key`, or removing it entirely for `keyless`.

**Frontend** (`frontend/src/components/mcp/connect-mcp-dialog.tsx`)

- Route the `mcp_oauth` connect to `/api/mcp/{server_id}/oauth/connect` for custom entries — the same per-server endpoint the custom MCP edit form already drives. Catalog apps keep using the catalog route.
- Recheck the connection against the listing branch the entry came from once the consent popup closes. A custom server only appears under `location=local`; querying the remote branch reported every custom connect as a failure.

No separate `discover` call is needed from the picker: `POST /{server_id}/oauth/connect` runs discovery internally. The edit form's Discover button only backfills form fields.

## Known limitations

- A deactivated association does not get `auth_type`. The per-server OAuth endpoints use `require_active=True`, so advertising the flow there would only swap the toast for a 404 — such a server needs re-enabling, not re-authorization. Documented inline.
- `_is_mcp_oauth_server` lowercases the transport while `_get_mcp_oauth_config` (mcp.py:597) matches exactly. A mixed-case transport is only storable via a direct API call or an admin PATCH (the form offers a fixed lowercase set), and the inconsistency predates this change, so it is left alone.

## Tests

Four regression tests, each verified to fail without the fix:

- `test_local_mcp_oauth_listing_carries_the_auth_type_for_the_picker`
- `test_local_non_mcp_oauth_server_carries_no_auth_type`
- `test_local_mcp_oauth_listing_omits_auth_type_when_deactivated`
- Frontend: per-server endpoint routing, and the `location=local` recheck after the popup closes

Verification: `tests/web/api/` 618 passed (3 skipped — no postgres URL set); `src/components/mcp` 61 passed; CI test manifest 76 passed; `npm run lint` and `npm run type-check` clean; `ruff check` and `ruff format --check` clean; mypy reports no errors in `mcp.py`.

### DIFF
--- a/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
@@ -158,6 +158,9 @@ vi.mock("./official-mcp-settings-dialog", () => ({
       <button type="button" onClick={() => onConnectStart(mcpOauthApp())}>
         connect-granola
       </button>
+      <button type="button" onClick={() => onConnectStart(customMcpOauthApp())}>
+        connect-records
+      </button>
       <button type="button" onClick={() => onConnectStart(keylessApp())}>
         connect-chrome
       </button>
@@ -231,6 +234,25 @@ function mcpOauthApp() {
     description: "",
     icon: "",
     is_connected: false,
+    transport: "streamable_http",
+    auth_type: "mcp_oauth",
+  }
+}
+
+// A user-added mcp_oauth MCP server whose consent was never completed: it is
+// listed under location=local with a server_id, and authorizing it must go
+// through the per-server /api/mcp/{server_id}/oauth/connect — the catalog
+// route resolves app ids only and can never see it (#1313).
+function customMcpOauthApp() {
+  return {
+    id: "records",
+    name: "records",
+    description: "",
+    icon: "",
+    is_connected: false,
+    is_custom: true,
+    is_local: true,
+    server_id: 9,
     transport: "streamable_http",
     auth_type: "mcp_oauth",
   }
@@ -509,6 +531,102 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
       expect(toastErrorMock).not.toHaveBeenCalled()
     } finally {
       openSpy.mockRestore()
+    }
+  })
+
+  it("connects a custom mcp_oauth server through the per-server oauth/connect endpoint", async () => {
+    // #1313: this Connect button used to be unreachable — local entries
+    // carried no auth_type, so the dispatch fell through to the
+    // mis-authored-entry toast. With the hint in place it must also avoid
+    // the catalog route: /api/mcp/apps/{id}/oauth/connect resolves catalog
+    // app ids only and would 404 on a user-added server.
+    const popup = { closed: false, close: vi.fn(), opener: {}, location: { href: "" } }
+    const openSpy = vi.spyOn(window, "open").mockReturnValue(popup as unknown as Window)
+    try {
+      apiRequestMock.mockImplementation((url: string, options?: RequestInit) => {
+        if (url.includes("/api/mcp/apps?")) {
+          return Promise.resolve({ ok: true, json: async () => [] })
+        }
+        if (url === "http://api.local/api/mcp/9/oauth/connect") {
+          expect(options?.method).toBe("POST")
+          return Promise.resolve({
+            ok: true,
+            json: async () => ({ authorization_url: "https://auth.example.com/authorize" }),
+          })
+        }
+        throw new Error(`Unexpected request: ${url}`)
+      })
+
+      renderDialog()
+      fireEvent.click(screen.getByRole("button", { name: "connect-records" }))
+
+      await waitFor(() => {
+        expect(popup.location.href).toBe("https://auth.example.com/authorize")
+      })
+      expect(popup.close).not.toHaveBeenCalled()
+      expect(toastErrorMock).not.toHaveBeenCalled()
+    } finally {
+      openSpy.mockRestore()
+    }
+  })
+
+  it("rechecks a closed custom mcp_oauth popup against the local listing", async () => {
+    // The popup's opener link is severed, so a closed popup is ambiguous and
+    // the handler asks the backend what actually happened. A custom server
+    // only exists under location=local — querying the remote branch (as the
+    // catalog path does) would report every custom connect as a failure.
+    const popup = { closed: false, close: vi.fn(), opener: {}, location: { href: "" } }
+    const openSpy = vi.spyOn(window, "open").mockReturnValue(popup as unknown as Window)
+    const onSuccess = vi.fn()
+    let localListCalls = 0
+    try {
+      apiRequestMock.mockImplementation((url: string) => {
+        if (url === "http://api.local/api/mcp/9/oauth/connect") {
+          return Promise.resolve({
+            ok: true,
+            json: async () => ({ authorization_url: "https://auth.example.com/authorize" }),
+          })
+        }
+        if (url === "http://api.local/api/mcp/apps?location=local") {
+          localListCalls += 1
+          return Promise.resolve({
+            ok: true,
+            json: async () => [{ ...customMcpOauthApp(), is_connected: true }],
+          })
+        }
+        if (url.includes("/api/mcp/apps?")) {
+          return Promise.resolve({ ok: true, json: async () => [] })
+        }
+        throw new Error(`Unexpected request: ${url}`)
+      })
+
+      render(
+        <ConnectMcpDialog
+          open
+          onOpenChange={vi.fn()}
+          selectedMcpServers={selectedMcpServers}
+          onSuccess={onSuccess}
+        />,
+      )
+
+      // Fake timers only from here on: the poll's setInterval must be created
+      // under them to be advanceable, and @testing-library's async queries
+      // hang once they are active.
+      vi.useFakeTimers()
+      await act(async () => {
+        fireEvent.click(screen.getByRole("button", { name: "connect-records" }))
+      })
+
+      popup.closed = true
+      await act(async () => {
+        await vi.advanceTimersByTimeAsync(500)
+      })
+
+      expect(localListCalls).toBe(1)
+      expect(onSuccess).toHaveBeenCalled()
+    } finally {
+      openSpy.mockRestore()
+      vi.useRealTimers()
     }
   })
 
--- a/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
@@ -630,6 +630,93 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     }
   })
 
+  // Pins the ordering the F5 fix depends on: a dialog close must beat a
+  // still-in-flight connect POST that resolves afterwards. isMountedRef
+  // alone cannot see this — it only flips false on unmount, and both real
+  // consumers of this dialog keep the component mounted across open/close
+  // (only the `open` prop toggles), so isMountedRef.current is still true
+  // here. Before the fix, that made the post-await guard pass regardless of
+  // the close, registering a poll after clearMcpOauthPollState() had already
+  // run — a poll nothing would ever clear, which would go on to recheck
+  // location=local and fire onSuccess against a dialog the user already
+  // closed.
+  it("skips registering a poll when the dialog closes before the connect POST resolves", async () => {
+    const popup = { closed: false, close: vi.fn(), opener: {}, location: { href: "" } }
+    const openSpy = vi.spyOn(window, "open").mockReturnValue(popup as unknown as Window)
+    const onSuccess = vi.fn()
+    let localListCalls = 0
+    const connectResponse = deferred<{ ok: boolean; json: () => Promise<{ authorization_url: string }> }>()
+    try {
+      apiRequestMock.mockImplementation((url: string) => {
+        if (url === "http://api.local/api/mcp/9/oauth/connect") {
+          return connectResponse.promise
+        }
+        if (url === "http://api.local/api/mcp/apps?location=local") {
+          localListCalls += 1
+          return Promise.resolve({ ok: true, json: async () => [] })
+        }
+        if (url.includes("/api/mcp/apps?")) {
+          return Promise.resolve({ ok: true, json: async () => [] })
+        }
+        throw new Error(`Unexpected request: ${url}`)
+      })
+
+      const { rerender } = render(
+        <ConnectMcpDialog
+          open
+          onOpenChange={vi.fn()}
+          selectedMcpServers={selectedMcpServers}
+          onSuccess={onSuccess}
+        />,
+      )
+
+      // Fake timers from here on, same as the recheck test above: the poll's
+      // setInterval (if one is wrongly registered) must be created under
+      // them to be advanceable, and this test issues no further async
+      // queries that would hang once they are active.
+      vi.useFakeTimers()
+
+      await act(async () => {
+        fireEvent.click(screen.getByRole("button", { name: "connect-records" }))
+      })
+
+      // Close the dialog while the connect POST above is still unresolved —
+      // only the `open` prop changes, so the component itself stays mounted
+      // and isMountedRef never flips. This is what runs
+      // clearMcpOauthPollState() and bumps the generation counter, ahead of
+      // the POST settling.
+      await act(async () => {
+        rerender(
+          <ConnectMcpDialog
+            open={false}
+            onOpenChange={vi.fn()}
+            selectedMcpServers={selectedMcpServers}
+            onSuccess={onSuccess}
+          />,
+        )
+      })
+
+      await act(async () => {
+        connectResponse.resolve({
+          ok: true,
+          json: async () => ({ authorization_url: "https://auth.example.com/authorize" }),
+        })
+        await connectResponse.promise
+      })
+
+      popup.closed = true
+      await act(async () => {
+        await vi.advanceTimersByTimeAsync(500)
+      })
+
+      expect(localListCalls).toBe(0)
+      expect(onSuccess).not.toHaveBeenCalled()
+    } finally {
+      openSpy.mockRestore()
+      vi.useRealTimers()
+    }
+  })
+
   it("fires exactly one POST when a keyless connect is double-clicked in the same tick", async () => {
     // Round-9: keylessConnectsRef's guard had zero test coverage — removing
     // its .has()/.add() calls would leave every other test green. Without
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -710,12 +710,27 @@ export function ConnectMcpDialog({
     }
   }
 
-  // Remote-MCP OAuth catalog app (e.g. Granola): the backend ensures the
-  // shared server row + this user's association, runs Dynamic Client
-  // Registration when the provider has no static client, and returns the
-  // authorization URL. Mirrors custom-mcp-form's handleConnectMcpOAuth,
-  // but keyed by catalog app_id instead of a server id.
+  // A user-added (custom) MCP server is authorized per server row, not per
+  // catalog app id: /api/mcp/apps/{id}/oauth/connect resolves the catalog and
+  // nothing else, so custom servers must drive the per-server endpoint the
+  // custom MCP edit form already uses (#1313). Null for catalog apps.
+  const customMcpServerId = (app: AppIntegration): number | null =>
+    app.is_custom && Number.isInteger(app.server_id) ? (app.server_id as number) : null
+
+  // Remote-MCP OAuth app: the backend ensures the server row + this user's
+  // association, runs Dynamic Client Registration when the provider has no
+  // static client, and returns the authorization URL. Mirrors
+  // custom-mcp-form's handleConnectMcpOAuth. Serves both a catalog app
+  // (e.g. Granola), keyed by app_id, and a custom server, keyed by server id.
   const handleConnectMcpOAuthApp = async (app: AppIntegration, autoSelect: boolean) => {
+    const serverId = customMcpServerId(app)
+    const connectUrl = serverId === null
+      ? `${getApiUrl()}/api/mcp/apps/${app.id}/oauth/connect`
+      : `${getApiUrl()}/api/mcp/${serverId}/oauth/connect`
+    // The post-consent recheck below has to query the same listing branch the
+    // entry came from: a custom server only appears under location=local, so
+    // asking for remote would report every custom connect as a failure.
+    const refreshLocation = serverId === null ? "remote" : "local"
     markAppLoading(app.id)
     // Open the popup synchronously on the click, before any await — popup
     // blockers reject windows opened outside direct user-gesture handling.
@@ -737,7 +752,7 @@ export function ConnectMcpDialog({
     }
     popup.opener = null
     try {
-      const response = await apiRequest(`${getApiUrl()}/api/mcp/apps/${app.id}/oauth/connect`, {
+      const response = await apiRequest(connectUrl, {
         method: "POST",
         credentials: "include",
         headers: {
@@ -799,7 +814,7 @@ export function ConnectMcpDialog({
       void (async () => {
         let connected = false
         try {
-          const response = await apiRequest(`${getApiUrl()}/api/mcp/apps?location=remote`)
+          const response = await apiRequest(`${getApiUrl()}/api/mcp/apps?location=${refreshLocation}`)
           if (response.ok) {
             const data = sanitizeAppIntegrations(await response.json())
             connected = data.some(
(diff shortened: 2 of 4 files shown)


src/xagent/web/api/mcp.py
```diff
@@ -1823,6 +1823,20 @@ def _connected_non_oauth_server_for_app(
...
+def _is_mcp_oauth_server(server: MCPServer) -> bool:
+    """Whether a stored server row is authorized through per-user MCP OAuth.
+
+    The shape check behind both the connection-state gate below and the
+    connector picker's `auth_type` hint, so the field that decides which
+    Connect flow the picker starts can't disagree with the field that decides
+    whether the server counts as connected."""
+    auth: dict[str, Any] = server.auth if isinstance(server.auth, dict) else {}
+    return (
+        str(server.transport or "").lower() in HTTP_MCP_TRANSPORTS
```
rogercloud, line 1835: This shape check lowercases the transport before comparing, but `_get_mcp_oauth_config` (line 597) compares `server.transport` against `HTTP_MCP_TRANSPORTS` without normalization, so a mixed-case transport value — storable only via direct API or admin PATCH, since the UI form offers a fixed lowercase set — would pass here and route Connect into the per-server endpoint, which then rejects it with a 400. The mismatch is pre-existing and the PR body documents it as a deliberate known limitation, but since this helper is the new gate that surfaces it on the custom path, a one-line normalization (`.lower()`) in `_get_mcp_oauth_config` would close the gap.

src/xagent/web/api/mcp.py
```diff
@@ -2015,28 +2025,43 @@ def list_mcp_apps(
...
+            #
+            # Emitted only for the mcp_oauth shape, deliberately: every other
+            # custom shape is reported connected unconditionally (so its
+            # Connect button never renders), while tagging those rows with a
+            # catalog classification would repoint the settings dialog's
+            # Configure button away from the custom edit form. Inactive
+            # associations are excluded because the per-server OAuth endpoints
+            # require an active one — a deactivated server needs re-enabling,
+            # not re-authorization.
+            if user_mcp.is_active and _is_mcp_oauth_server(server):
```
rogercloud, line 2061: `_mcp_oauth_server_is_actually_connected` checks shape and grant but never `user_mcp.is_active`, so a deactivated association with no completed grant still renders a Connect button — yet this `is_active` gate suppresses the auth_type hint, and the click falls through to the mis-authored-entry toast. Reachable only via direct API, and not a regression (pre-fix, all such servers dead-ended). The PR body documents the tradeoff: the per-server OAuth endpoints require an active association, so emitting the hint would swap the toast for a 404. A distinct toast or a re-enable affordance for the deactivated case would be a worthwhile follow-up.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -710,12 +710,27 @@ export function ConnectMcpDialog({
...
+    app.is_custom && Number.isInteger(app.server_id) ? (app.server_id as number) : null
+
+  // Remote-MCP OAuth app: the backend ensures the server row + this user's
+  // association, runs Dynamic Client Registration when the provider has no
+  // static client, and returns the authorization URL. Mirrors
+  // custom-mcp-form's handleConnectMcpOAuth. Serves both a catalog app
+  // (e.g. Granola), keyed by app_id, and a custom server, keyed by server id.
   const handleConnectMcpOAuthApp = async (app: AppIntegration, autoSelect: boolean) => {
+    const serverId = customMcpServerId(app)
+    const connectUrl = serverId === null
```
rogercloud, line 747: This handler is now reachable for custom servers, and like the catalog path it has no same-tick double-click guard: `disabled={isConnecting}` lags one commit behind, so two synchronous clicks can issue two POSTs to `/api/mcp/{id}/oauth/connect`, creating two `MCPOAuthFlowState` rows — the second `Set-Cookie` wins and the first is orphaned until the 10-minute TTL. The keyless path already closes this window with `keylessConnectsRef`. Suggestion: mirror that pattern with a synchronous Set check/add at handler entry, cleared when the poll completes.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -799,7 +814,7 @@ export function ConnectMcpDialog({
       void (async () => {
         let connected = false
         try {
-          const response = await apiRequest(`${getApiUrl()}/api/mcp/apps?location=remote`)
+          const response = await apiRequest(`${getApiUrl()}/api/mcp/apps?location=${refreshLocation}`)
```
rogercloud, line 842: The `isMountedRef.current` guard at line 792 doesn't cover this path: it flips only on unmount, while both consumers render the dialog unconditionally with an `open` prop — closing clears the poll state without unmounting, so the in-flight POST resolves, the guard passes, and this refresh fires `loadApps`/`onSuccess`/`autoSelect` against a closed dialog, contradicting the F5 comment's stated intent. Harm is bounded (extra refresh requests on a hidden dialog) and the gap pre-existed for the catalog route, but this PR extends it to the custom path. Suggestion: also check the dialog's open state at the guard.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -512,6 +534,102 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     }
   })
 
+  it("connects a custom mcp_oauth server through the per-server oauth/connect endpoint", async () => {
```
rogercloud, line 537: The new happy-path test pins the per-server URL and the local-listing recheck, but there is no custom-route test for the failure case (non-OK response from the per-server endpoint). Informational: the error branch is single shared code already covered by the catalog granola-400 test, so a route-specific failure test would execute the same statements. Low priority — add one only if the error path is later differentiated per route.

rogercloud (approved):
## Summary

Connecting a user-created (custom) MCP server that uses the `streamable_http` transport with `mcp_oauth` auth from the "+ Connector" picker was a dead end: the `location=local` branch of `GET /api/mcp/apps` built entries without an `auth_type` hint, so the Connect dispatch fell through to the misleading "not configured" toast, and the only mcp_oauth branch posted to `/api/mcp/apps/{id}/oauth/connect`, which resolves catalog app ids only. This PR emits `auth_type: mcp_oauth` for custom server rows (gated on an active association, via a new `_is_mcp_oauth_server` shape check shared with the connection-state gate) and routes the picker's Connect to the per-server `/api/mcp/{id}/oauth/connect` endpoint — the same flow the custom MCP edit form already drives — with the post-consent recheck querying the same listing branch the entry came from. Backend tests pin the three connection states and the auth_type emission; frontend tests pin the catalog and per-server dispatch behaviors.

## Approach verdict

Sound. The fix reuses the exact endpoint the custom MCP edit form already drives (per-server OAuth connect runs discovery internally and requires an active association), so the picker's Connect and the edit form's Connect cannot diverge; and the auth_type hint is deliberately scoped to the mcp_oauth shape, so every other custom shape keeps its unconditional connected status and the settings dialog's Configure button stays pointed at the custom edit form. The user-to-grant model is 1:1, so there is no fan-out mismatch.

## Minor findings

**1. [minor, informational] Transport case-sensitivity mismatch — `src/xagent/web/api/mcp.py:1835`**
`_is_mcp_oauth_server` lowercases the transport before comparing against `HTTP_MCP_TRANSPORTS`, but `_get_mcp_oauth_config` (`src/xagent/web/api/mcp.py:597`) compares `server.transport` directly, so a mixed-case transport value passes the new shape check and routes Connect into the per-server endpoint, which then rejects it with a 400. The mismatch is pre-existing and a mixed-case value is only storable via direct API/admin PATCH (the UI form offers a fixed lowercase set). The PR body documents this as a deliberate known limitation, and the explanation was verified code-accurate; the failure mode arguably improves on the status quo (an accurate backend detail surfaced via `parseMcpOAuthErrorMessage` instead of the misleading toast). Informational — an optional one-line follow-up would be normalizing with the same `.lower()` inside `_get_mcp_oauth_config`.

**2. [minor, informational] Deactivated association still dead-ends — `src/xagent/web/api/mcp.py:2061`**
`_mcp_oauth_server_is_actually_connected` checks shape + grant but never `user_mcp.is_active`, so a deactivated association with no completed grant still renders a Connect button here — yet the `is_active` gate on this line suppresses the auth_type hint, and the click falls through to the mis-authored-entry toast again. Reachable only via direct API (no UI deactivates server associations), and not a regression: before this PR every custom mcp_oauth server without a grant dead-ended the same way. The PR body documents the tradeoff — the per-server OAuth endpoints use `require_active=True`, so emitting the hint would swap the toast for a 404 — and the explanation was verified code-accurate. Informational; a re-enable affordance or a distinct toast for the deactivated case would be a good follow-up.

**3. [minor] No same-tick double-click guard on the mcp_oauth Connect handler — `frontend/src/components/mcp/connect-mcp-dialog.tsx:727`**
The only protection is `disabled={isConnecting}`, which lags one commit behind a synchronous double-click; the keyless path solves the same window with `keylessConnectsRef`. Two same-tick clicks produce two POSTs and two `MCPOAuthFlowState` rows (the second `Set-Cookie` wins and the first is orphaned until the 10-minute TTL). Impact is bounded (the popup reuses the same window name and the flow stays consistent), and the exposure pre-existed for the catalog route — but this PR routes custom servers through the same unguarded handler. Suggestion: mirror `keylessConnectsRef` — a synchronous Set check/add at handler entry, cleared in the poll-completion branch.

**4. [minor] `isMountedRef` guard doesn't cover close-while-POST-in-flight — `frontend/src/components/mcp/connect-mcp-dialog.tsx:817`**
The guard at line 792 flips only on unmount, but both consumers render the dialog unconditionally with an `open` prop, so closing clears the poll state without unmounting: the POST resolves, the guard passes, and this refresh fires `loadApps`/`onSuccess`/`autoSelect` against a closed dialog — contradicting the F5 comment's stated intent. Harm is bounded (extra refresh requests on a hidden dialog; the refresh is arguably even desirable), and the gap pre-existed for the catalog route, but the PR extends it to the custom path. Suggestion: also check the dialog's open state in the guard.

**5. [minor, informational] No frontend test for the custom-route failure path — `frontend/src/components/mcp/connect-mcp-dialog.test.tsx:537`**
The new happy-path test pins the per-server URL and the recheck branch, but there is no custom-route test for a non-OK response from the per-server endpoint. Informational: the error branch is single shared code already covered by the catalog granola-400 test, and a route-specific failure test would execute the same statements. Low priority.

## Design notes

The popup + POST + poll sequence is nearly duplicated between `handleConnectMcpOAuthApp` (`frontend/src/components/mcp/connect-mcp-dialog.tsx:735-839`) and custom-mcp-form's `handleConnectMcpOAuth` (`frontend/src/components/mcp/custom-mcp-form.tsx:250-327`). This duplication pre-dates the PR (the PR only routes the dialog's existing handler at a second endpoint), and the recheck strategies legitimately diverge — the dialog re-lists apps while the form polls the per-server oauth/status — so this is a non-blocking follow-up recommendation, not a defect: a shared helper that parameterizes the verification step is feasible, and drift of exactly this class is what caused #1313.

Blocking: no. All five findings are minor: #1 and #2 are deliberate, code-accurately documented known limitations (informational); #3 and #4 extend pre-existing guard gaps to the newly routed custom path with bounded impact; #5 is an informational test-coverage note. The design note is a non-blocking follow-up. Recommended event: APPROVE.

codeacme17:
Thanks for the thorough review. I assessed each finding independently against the code before acting; summary of what changed and what did not.

**Implemented — finding 4** (`25b686e`). Confirmed, and it is a comment/code mismatch rather than just a gap: the F5 comment claims to cover "the dialog closed **or** this component unmounted", but `isMountedRef` flips only in the unmount effect, so on an ordinary close the guard passes and the poll is registered after `clearMcpOauthPollState()` already ran.

I did not take the literal suggestion. Checking `open` at the guard would be decorative — it is a prop captured in the handler's per-render closure, so it reads a stale value once the POST resolves. This file already contains four such `|| !open` checks (lines 361-403) where correctness actually comes from `connectorEditRequestRef`. I used that same generation-counter idiom instead: bumped inside `clearMcpOauthPollState()` (which runs on both the close branch and unmount cleanup), captured at handler entry, compared after the await. The early return deliberately skips `clearAppLoading` so a close-then-reopen cannot wipe the new attempt's spinner. Added a regression test that pins the close-beats-POST ordering; verified it fails without the guard change.

**Deferred to a follow-up — finding 3**: #1330. The claim holds, but a literal `keylessConnectsRef` copy would be worse than the bug: `mcp_oauth` holds its loading state for the whole 5-minute popup wait, and `clearMcpOauthPollState()` clears `loadingApps` without knowing about a new bespoke ref — so closing the dialog mid-flow would leave the app permanently unconnectable until reload. The design-consistent fix is a synchronous shadow of `loadingApps` (one invariant instead of six release points), which also subsumes `keylessConnectsRef` and closes the same gap in the unguarded `builtin_oauth` handler. That reaches shared plumbing used by three connect mechanisms, wider than a #1313 fix should go.

The issue also records two details worth carrying over: the shared popup window name means the second same-tick `window.open` reuses the first browsing context, so the flow still succeeds and user-visible harm is nil; and the part genuinely widened by this PR is that custom servers are the DCR case, so two concurrent POSTs create **two dynamic client registrations at the third-party authorization server**, one later discarded via the `registration_lookup_hash` `IntegrityError` fallback. One correction on the premise: `mcp_oauth_flow_states` has no expiry sweeper, so the orphaned row becomes unusable after the 10-minute TTL but stays in the table indefinitely.

**No change — findings 1, 2, 5.** Agreed as informational; replies on those threads.

One correction on finding 4's premise, for the record: `agent-builder.tsx:3034` gates the dialog on `{!readOnly && !loadingAgent && ...}`, so "both consumers render it unconditionally" is not quite right. Neither flag flips when the picker closes, so the component does stay mounted across open/close and your conclusion stands — only `tools/page.tsx:1409` is genuinely unconditional.

src/xagent/web/api/mcp.py
codeacme17: Verified — the mismatch is real: `_is_mcp_oauth_server` lowercases (`str(server.transport or "").lower()`) while `_get_mcp_oauth_config` compares `server.transport` directly. Leaving it as-is in this PR, for two reasons.

The lowercase side is the pre-existing convention here, not something this PR introduced: `_mcp_oauth_server_is_actually_connected` already lowercased before I extracted the helper, and `_build_active_oauth_server_lookup` / `_build_active_non_oauth_server_lookup` normalize the same way — the existing comment in `classify_app_auth` even calls out that "an admin PATCH can store a mixed-case transport, and the two halves of this feature must not disagree about the same row". So the outlier is line 597, and normalizing it is a change to the connect endpoint's validation semantics rather than a fix to what this PR touches.

Second, the resulting behaviour is strictly better than the status quo for the one case that can reach it. Before: `is_connected: false` plus the misleading "not configured" toast. After: an accurate backend detail ("MCP OAuth is only supported for HTTP MCP transports") surfaced through `parseMcpOAuthErrorMessage`. Neither connects, but the second tells the user why.

Happy to fold the one-line `.lower()` into this PR if you would rather have it closed here — it is a contained change and `_get_mcp_oauth_config` has no other callers that depend on the exact match. Otherwise I will leave line 597 alone so this PR stays scoped to #1313.

src/xagent/web/api/mcp.py
codeacme17: Agreed on all points, and the reading of the interaction is correct: `is_connected` here is shape + grant only, so a deactivated association with no grant still renders a Connect button, while the `is_active` gate on this line withholds the hint that would make the button work.

Keeping the gate. Emitting `auth_type` for a deactivated association would route the click into `/api/mcp/{server_id}/oauth/discover|connect`, both of which call `_get_user_mcp_server_or_404(..., require_active=True)` — so the toast becomes a 404, which is not an improvement. Re-authorization is also the wrong remedy: the server was deliberately deactivated (`POST /api/mcp/{server_id}/toggle`), and completing OAuth would leave it deactivated anyway.

The right fix is the affordance you suggest, and it is a product decision rather than a bug fix: the picker has no re-enable control today, and `is_connected` arguably should account for `is_active` so a deactivated server does not present as selectable either — which is a behaviour change beyond #1313's scope and would want its own issue. I have documented the tradeoff inline so the gate does not read as an oversight.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Confirmed and filed as #1330 rather than fixed here.

The claim holds, but the suggested fix does not survive contact with this handler: unlike a keyless connect, `mcp_oauth` keeps its loading state for the entire 5-minute popup wait, and `clearMcpOauthPollState()` clears `loadingApps` without knowing about a new bespoke ref. So a literal `keylessConnectsRef` copy would leave the app **permanently unconnectable until page reload** whenever the user closes the dialog mid-flow — worse than the duplicate POST. Getting it right means clearing the ref at six release points (popup-blocked, `!response.ok`, missing authorization URL, `catch`, the mount/generation bail-out, and the interval) plus `clearMcpOauthPollState`.

The design-consistent fix is the one this file's own comment at the `loadingApps` declaration already argues for — a synchronous shadow of `loadingApps` maintained inside `markAppLoading` / `clearAppLoading` / `clearMcpOauthPollState`, so there is one invariant instead of six. That also subsumes `keylessConnectsRef` and closes the same gap in the `builtin_oauth` handler, which today has no guard at all and would open two provider popups plus two `postMessage` listeners. Three connect mechanisms is wider than a #1313 fix should reach, hence the separate issue.

Two things the issue records that are worth noting here. The shared popup window name (`MCP_OAUTH_POPUP_WINDOW_NAME`) means the second same-tick `window.open` reuses the first browsing context and returns the same `WindowProxy`, so there is no second consent popup and the flow still completes — user-visible harm is nil. The part genuinely widened by this PR is that custom servers are the DCR case (no configured `client_id`), so both POSTs find no `MCPOAuthClient` and both call `register_mcp_oauth_public_client`: **two dynamic client registrations at the third-party authorization server**, one discarded by `_upsert_mcp_oauth_client`'s `registration_lookup_hash` `IntegrityError` fallback. That side effect at someone else's server is what justifies the work.

One correction: `mcp_oauth_flow_states` has no expiry sweeper (unlike `SlackOAuthFlowState`, which purges on each new flow). After the 10-minute TTL the orphaned row becomes unusable — the claim query requires `expires_at > claimed_at` — but it stays in the table indefinitely; the only deletes are the per-user purge on disconnect and the FK cascade on server deletion.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Fixed in `25b686e` — but not with the literal suggestion, because checking `open` at the guard would not work.

`open` is a prop captured in `handleConnectMcpOAuthApp`'s per-render closure. The running async function holds the closure from the render where `open === true`, so once the POST resolves `!open` evaluates false and the guard still passes. This file already has four of those decorative checks (lines 361, 374, 391, 403) where correctness actually comes from `connectorEditRequestRef.current`, which *is* bumped on close — so the pattern looks established but is load-bearing nowhere.

I used that generation-counter idiom instead:

```ts
const mcpOauthPollGenerationRef = React.useRef(0)   // bumped inside clearMcpOauthPollState()
const pollGeneration = mcpOauthPollGenerationRef.current   // at handler entry, before the popup and any await
if (!isMountedRef.current || pollGeneration !== mcpOauthPollGenerationRef.current) return
```

`clearMcpOauthPollState()` runs from both the `open` effect's close branch and the unmount cleanup, so one condition now covers unmount, close, and close-then-reopen. The early return deliberately skips `clearAppLoading`: `clearMcpOauthPollState` already emptied `loadingApps`, and on a close-then-reopen clearing it would wipe the *new* attempt's spinner. The documented behaviour that the popup is left navigating and the next apps refresh reflects a completed auth is unchanged.

Added `skips registering a poll when the dialog closes before the connect POST resolves`, which pins the ordering your comment identifies — the existing F6 test registers its poll while open, so it never exercised this path. Verified non-vacuous: with the guard reverted to `if (!isMountedRef.current) return` it fails with `expected 1 to be +0` (the stale poll performs the `location=local` recheck).

One correction on the premise, for the record: `agent-builder.tsx:3034` gates the dialog on `{!readOnly && !loadingAgent && ...}`, so it is not strictly unconditional. Neither flag flips when the picker closes, so the component does stay mounted across open/close and your conclusion holds — only `tools/page.tsx:1409` renders it unconditionally.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: Agreed, no change. Both routes share one error branch — the `!response.ok` path calls `parseMcpOAuthErrorMessage` and `popup.close()` regardless of which URL was posted to — so a custom-route failure test would execute the same statements the granola-400 test already covers, and would pass even if the per-server routing were broken. It would pin the mock, not the behaviour.

Worth adding if the error path is ever differentiated per route, e.g. if the per-server endpoint's 404/400 (`_get_user_mcp_server_or_404`, `_get_mcp_oauth_config`) gets a distinct message or a "re-enable this server" affordance — which is roughly the follow-up sketched on the `is_active` thread above.

I did add a second custom-route test in `25b686e`, though it targets a different concern (the dialog-close guard from your finding 4) rather than the HTTP failure path.

PR #1325, opened 2026-08-13
fix(react): reject final_answer that carries an empty answer

### LINKED ISSUE #1312: ReAct completes as success with no user-visible answer when final_answer carries an empty answer
## Summary

In the ReAct pattern, the `final_answer` control tool is the only way the model can reply to the user, because every LLM call is made with `tool_choice="required"` and plain-text responses are not accepted. When the model calls `final_answer` with a missing or empty `answer` — or with an arguments payload that fails JSON parsing — the run still finishes as `success=True` / `status="completed"`, nothing is streamed to the UI, and the persisted assistant message becomes the `"No output provided"` placeholder. Neither condition is logged, so the failure leaves no trace to debug from.

Reproduction seen on a chat surface with an agent in `balanced` mode: the agent calls an MCP tool, the tool succeeds and returns data, then the conversation shows no reply. It is intermittent (2 out of 3 identical prompts in one session), which is consistent with a model-output-dependent trigger.

## Mechanism

All links pin to current `main` (6488911).

1. The `final_answer` handler has no guard against an empty answer — it finalizes successfully either way, and only a non-empty answer is recorded as an assistant message ([react.py#L1685-L1706](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/pattern/react/react.py#L1685-L1706)):

```python
if name == "final_answer":
    answer = str(args.get("answer", ""))
    ...
    if answer:
        context.add_assistant_message(answer)
    return await self._finalize_outcome(
        context=context,
        runtime=runtime,
        response=answer,
        outcome=outcome,
    )
```

2. Tool-call turns store the assistant message with empty string content when the model emits no preamble text ([react.py, `add_assistant_message(assistant_content or "", ...)`](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/pattern/react/react.py)). So after "MCP tool call with no preamble → empty final answer", the context contains no non-empty assistant message at all.

3. The adapter's backfill requires a non-empty string assistant message; finding none, it substitutes the placeholder while keeping `success=True` ([execution_adapter.py#L361-L371](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/execution_adapter.py#L361-L371)):

```python
output = result.get("output", result.get("response", result.get("error")))
if not output:
    output = self._latest_assistant_message(result.get("context"))
...
"output": output or NO_OUTPUT_PLACEHOLDER,
```

4. The final-answer UI stream never receives a delta, so `FinalAnswerStreamSession.finish("")` returns before emitting anything ([final_answer_stream.py](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/pattern/final_answer_stream.py)). The user sees a completed run with no reply.

## Two concrete ways `answer` ends up empty

**A. Partial arguments object.** The tool schema lists `response_language` before `answer` ([react.py#L1528-L1554](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/pattern/react/react.py#L1528-L1554)). If the model emits only the leading properties (e.g. output truncated by max tokens), `args.get("answer", "")` returns `""`.

**B. Unparseable arguments JSON is silently swallowed.** `_coerce_arguments` converts a `JSONDecodeError` into `{"input": <raw string>}` instead of failing or retrying ([react.py#L1445-L1453](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/pattern/react/react.py#L1445-L1453)):

```python
try:
    parsed = json.loads(arguments)
except json.JSONDecodeError:
    return {"input": arguments}
```

The `answer` key disappears and case A's downstream behavior follows. Large MCP tool payloads in context make truncated or malformed follow-up tool calls more likely, which matches the observed correlation with MCP usage.

## AutoPattern already guards this

The auto pattern explicitly rejects an empty final answer and falls back to a full reasoning pass, logging a warning: `"AutoPattern selected final_answer without answer; falling back to react."` ([auto.py#L718](https://github.com/xorbitsai/xagent/blob/6488911dc6a1983a42b1dc7ebb3f8acdae2a43c2/src/xagent/core/agent/pattern/auto/auto.py#L718)). ReAct has no equivalent guard, even though under `tool_choice="required"` it is the pattern where `final_answer` is the only exit.

## Expected behavior / suggested fix

1. An empty `answer` in `final_answer` should not finalize the run as success. Mirror AutoPattern: reject it, log a warning, and force another reasoning turn (or re-request `final_answer`).
2. For control tools, `_coerce_arguments` should not silently coerce unparseable JSON into `{"input": ...}`; parse failure on `final_answer` should be retried and logged.
3. Both conditions should emit a log line (empty answer; arguments `JSONDecodeError` with the raw payload length) — today they are fully silent.
4. Optionally, list `answer` first in the schema properties to reduce the chance that truncation drops it.

codeacme17:
Fixes #1312.

## Problem

Under `tool_choice="required"`, `final_answer` is the only way ReAct can reply to the user. When the model called it with a missing, blank, or malformed `answer`, the run still finalized as `success=True` / `status="completed"`: nothing was streamed to the UI, the persisted assistant message fell back to the `"No output provided"` placeholder, and neither condition was logged — so the failure left no trace to debug from.

`AutoPattern` already rejected an empty final answer and fell back to a full reasoning pass. ReAct had no equivalent guard, even though it is the pattern where `final_answer` is the only exit.

## Fix

Treat an empty answer as a tool-protocol violation, so it flows into the pattern's **existing** single repair retry rather than needing new control flow — and fail the run explicitly when the retry is empty too.

- **`_response_requires_tool_protocol_retry`** rejects a `final_answer` whose `answer` is missing or blank. It previously checked the forced-final-answer turn by tool *name* only and never inspected the arguments. A new `empty_final_answer` recovery reason carries a tailored retry instruction — worded differently on a forced turn, where `final_answer` is the only available tool — and its own trace phase (`empty_final_answer_recovery`).
- **`_coerce_arguments`** no longer coerces malformed JSON into `{"input": ...}` for control tools. That path silently stripped `answer` and let the run finalize with nothing to show; it now logs and drops the payload so the check above rejects the call. Work tools keep the opaque passthrough unchanged.
- **`_handle_control_tool`** refuses to finalize on an empty answer and re-requests one. This covers the paths that reach the handler without passing through response normalization — most importantly a checkpoint resume, which restores `pending_tool_calls` verbatim. It returns control to the loop with `force_final_answer_next` set, so the normalization guard bounds the retry instead of letting it spin.
- **Logging** at both conditions and at the give-up path, plus a warning in `AgentExecutionAdapter._normalize_result` before it substitutes `NO_OUTPUT_PLACEHOLDER`.

### Bounding

The retry is bounded by the mechanism it reuses: one repair attempt, then `_invalid_tool_protocol_result` returns `success=False` with `status="invalid_tool_protocol"`. The handler-level guard is bounded by the normalization guard on the following turn. An unbounded "force another reasoning turn" would have spun, because on the forced branch the tool set is reduced to `final_answer` alone.

## Two suggestions from the issue deliberately not applied

**Reordering the schema so `answer` comes first.** `response_language` is sequenced first so the model commits to an output language before writing prose, and `AutoDecision.to_payload` orders the two fields the same way. `answer` is already in `required` with `additionalProperties: False`, so the trigger was never a permissive schema — it was providers that do not enforce strict schemas, `max_tokens` truncation, and the swallowed parse error. Reordering would invert deliberate sequencing to mitigate only the truncation case, which the guard now catches anyway.

**Making the adapter report `success=False` on empty output.** The comment at `execution_adapter.py:363-367` documents that the delegated-child classifier in `agent_tool.py` reads the pre-backfill answer from `agent_result` specifically so a backfilled preamble cannot be mistaken for a real final answer. Changing the success semantics there risks that contract; the root cause is in ReAct. Logging is safe, so that is all this PR does at that layer — gated on `success`, since an empty output is legitimate on every non-success path.

## Tests

Six regression tests in `tests/core/agent/test_react.py`, each verified to fail without the fix:

| Test | Pre-fix failure |
|---|---|
| `test_react_retries_final_answer_that_omits_the_answer_field` | `assert '' == 'The result is 4.'` |
| `test_react_fails_when_final_answer_stays_empty_after_retry` | `assert True is False` on `success` — the exact reported symptom |
| `test_react_retries_final_answer_with_malformed_arguments` | same empty finalization via the JSON path |
| `test_react_re_requests_final_answer_for_resumed_empty_pending_call` | resume path bypasses normalization |
| `test_coerce_arguments_drops_unusable_control_tool_payloads` | `tool_name` kwarg did not exist |
| `test_empty_final_answer_call_detects_blank_answers` | helper did not exist |

The failure test also asserts the warning lines are emitted, since "no trace to debug from" was half the report.

One behavior worth flagging, asserted rather than left implicit: truncated arguments can stream a partial answer before the call is rejected. That partial is closed with `final_answer_error` and superseded by a fresh stream — the same shape the pattern already uses for a protocol retry. The point is that the run now ends with a real answer instead of nothing.

## Verification

- `tests/core/agent` (289 tests) and `tests/core/model` pass.
- ruff 0.12.3, mypy 1.19.0, and all pre-commit hooks pass.

My local environment is not provisioned for the whole suite — on an **unmodified** tree it already has 164 failures across ~40 files, from missing optional dependencies (pptxgenjs/Node, SVG rasterization, sandbox/Docker, OAuth, deepdoc) plus a set of ordering-dependent `caplog` tests. I captured that baseline by reverting this diff and re-running the identical command, so the pre-existing set is known rather than assumed; in particular every failure in `test_web_tool_config_session_factory.py` and `test_task_tracker.py` is present with this change reverted. CI is the authority on suite health here.

### DIFF
--- a/src/xagent/core/agent/execution_adapter.py
+++ b/src/xagent/core/agent/execution_adapter.py
@@ -366,6 +366,20 @@ class AgentExecutionAdapter:
                 # ``agent_result`` specifically so a backfilled preamble
                 # cannot be mistaken for a real final answer.
                 output = self._latest_assistant_message(result.get("context"))
+        if not output:
+            # A run that reports success with nothing to show is a bug in the
+            # pattern that produced it, and the placeholder erases the evidence.
+            # Log before substituting so there is something to debug from.
+            logger.warning(
+                "Agent %r produced no output; substituting the placeholder. "
+                "execution_type=%s pattern=%s status=%s success=%s task_id=%s",
+                self.config.name,
+                execution_type,
+                self.config.pattern,
+                status,
+                result.get("success"),
+                execution_id,
+            )
         normalized = {
             "status": status,
             "output": output or NO_OUTPUT_PLACEHOLDER,
--- a/src/xagent/core/agent/execution_adapter.py
+++ b/src/xagent/core/agent/execution_adapter.py
@@ -366,10 +366,15 @@ class AgentExecutionAdapter:
                 # ``agent_result`` specifically so a backfilled preamble
                 # cannot be mistaken for a real final answer.
                 output = self._latest_assistant_message(result.get("context"))
-        if not output:
+        if not output and result.get("success"):
             # A run that reports success with nothing to show is a bug in the
             # pattern that produced it, and the placeholder erases the evidence.
             # Log before substituting so there is something to debug from.
+            #
+            # Gated on ``success`` because an empty output is legitimate on every
+            # non-success path: a ``waiting_for_user`` pause carries only its
+            # message, so warning on emptiness alone would fire on the ordinary
+            # first-turn clarification and drown the real signal.
             logger.warning(
                 "Agent %r produced no output; substituting the placeholder. "
                 "execution_type=%s pattern=%s status=%s success=%s task_id=%s",
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -36,6 +36,7 @@ import copy
 import hashlib
 import inspect
 import json
+import logging
 from dataclasses import dataclass, replace
 from datetime import timezone
 from enum import Enum
@@ -71,6 +72,8 @@ from ...runtime import (
 )
 from ..base import AgentPattern, PatternResult, truncate_prompt_preview
 from ..final_answer_stream import ReActFinalAnswerStreamer
+
+logger = logging.getLogger(__name__)
 
 
 class ReActReasoningMode(str, Enum):
@@ -590,6 +593,20 @@ class ReActPattern(AgentPattern):
                     normalized,
                     force_final_answer=force_final_answer_now,
                 )
+                empty_final_answer = self._empty_final_answer_call(normalized)
+                if empty_final_answer is not None:
+                    logger.warning(
+                        "ReAct final_answer carried no answer text; discarding the "
+                        "response and retrying. iteration=%s arg_keys=%s",
+                        iteration,
+                        sorted(self._tool_call_args_dict(empty_final_answer)),
+                    )
+                if recover_full_tool_set:
+                    recovery_reason: str | None = "unavailable_tool_call"
+                elif empty_final_answer is not None:
+                    recovery_reason = "empty_final_answer"
+                else:
+                    recovery_reason = None
                 try:
                     (
                         response,
@@ -603,9 +620,7 @@ class ReActPattern(AgentPattern):
                         force_final_answer=(
                             force_final_answer_now and not recover_full_tool_set
                         ),
-                        recovery_reason=(
-                            "unavailable_tool_call" if recover_full_tool_set else None
-                        ),
+                        recovery_reason=recovery_reason,
                     )
                 except LLMCallInterrupted:
                     interrupted = await self._interrupt_if_requested(
@@ -699,6 +714,11 @@ class ReActPattern(AgentPattern):
         answer_streamer: ReActFinalAnswerStreamer | None,
         stream_failure_message: str,
     ) -> dict[str, Any]:
+        logger.warning(
+            "ReAct failing the run after an invalid tool protocol: %s (iteration=%s)",
+            stream_failure_message,
+            iteration,
+        )
         if answer_streamer is not None:
             await answer_streamer.fail(stream_failure_message)
         await runtime.checkpoint(
@@ -864,6 +884,16 @@ class ReActPattern(AgentPattern):
                 "complete user-facing response in its answer field."
             )
             retry_phase = "malformed_tool_arguments_recovery"
+        elif recovery_reason == "empty_final_answer":
+            retry_instruction = (
+                "The previous response called final_answer with an empty answer "
+                "field, so the user received no reply at all. Retry this turn: if "
+                "the task is complete, call final_answer again with the complete "
+                "user-facing response in its answer field; if work remains, call "
+                "the appropriate available work tool instead. Never call "
+                "final_answer with an omitted, empty, or whitespace-only answer."
+            )
+            retry_phase = "empty_final_answer_recovery"
         else:
             retry_instruction = (
                 "The previous response used an invalid tool protocol. Retry the same "
@@ -960,7 +990,32 @@ class ReActPattern(AgentPattern):
                 continue
             if force_final_answer and tool_call.get("name") != "final_answer":
                 return True
-        return False
+        return self._empty_final_answer_call(normalized) is not None
+
+    def _empty_final_answer_call(
+        self, normalized: dict[str, Any]
+    ) -> dict[str, Any] | None:
+        """Return the first ``final_answer`` call that carries no answer text.
+
+        Under ``tool_choice="required"`` ``final_answer`` is the only way ReAct
+        can reply, so an empty ``answer`` produces a run that completes with
+        nothing streamed and nothing persisted. Treating it as a tool-protocol
+        violation routes it into the pattern's existing single repair retry
+        instead of finalizing silently.
+        """
+
+        for tool_call in normalized.get("tool_calls") or []:
+            if not isinstance(tool_call, dict):
+                continue
+            if tool_call.get("name") != "final_answer":
+                continue
+            args = tool_call.get("args")
+            answer = args.get("answer") if isinstance(args, dict) else None
+            # Coerce exactly as ``_handle_control_tool`` does, so this check and
+            # the finalization it guards agree on what counts as an answer.
+            if answer is None or not str(answer).strip():
+                return tool_call
+        return None
 
     def _requires_full_tool_set_recovery(
         self,
@@ -1390,21 +1445,24 @@ class ReActPattern(AgentPattern):
                 function_payload = tool_call.get("function")
                 if isinstance(function_payload, dict):
                     arguments = function_payload.get("arguments", {})
+                    name = function_payload.get("name")
                     normalized.append(
                         {
                             "id": tool_call.get("id") or f"tool_call_{index}",
-                            "name": function_payload.get("name"),
-                            "args": self._coerce_arguments(arguments),
+                            "name": name,
+                            "args": self._coerce_arguments(arguments, tool_name=name),
                         }
                     )
                     continue
 
+                name = tool_call.get("name")
                 normalized.append(
                     {
                         "id": tool_call.get("id") or f"tool_call_{index}",
-                        "name": tool_call.get("name"),
+                        "name": name,
                         "args": self._coerce_arguments(
-                            tool_call.get("args", tool_call.get("arguments", {}))
+                            tool_call.get("args", tool_call.get("arguments", {})),
+                            tool_name=name,
                         ),
                     }
                 )
@@ -1412,12 +1470,14 @@ class ReActPattern(AgentPattern):
 
             function_payload = getattr(tool_call, "function", None)
             if function_payload is not None:
+                name = getattr(function_payload, "name", None)
                 normalized.append(
                     {
                         "id": getattr(tool_call, "id", None) or f"tool_call_{index}",
-                        "name": getattr(function_payload, "name", None),
+                        "name": name,
                         "args": self._coerce_arguments(
-                            getattr(function_payload, "arguments", {})
+                            getattr(function_payload, "arguments", {}),
+                            tool_name=name,
                         ),
                     }
                 )
@@ -1442,16 +1502,48 @@ class ReActPattern(AgentPattern):
                 self.pending_tool_call_content[tool_call_id] = content
             return
 
-    def _coerce_arguments(self, arguments: Any) -> dict[str, Any]:
+    def _coerce_arguments(
+        self, arguments: Any, *, tool_name: str | None = None
+    ) -> dict[str, Any]:
         if isinstance(arguments, dict):
             return arguments
         if isinstance(arguments, str):
             try:
                 parsed = json.loads(arguments)
             except json.JSONDecodeError:
-                return {"input": arguments}
-            return parsed if isinstance(parsed, dict) else {"input": parsed}
+                logger.warning(
+                    "ReAct tool call %s returned malformed JSON arguments "
+                    "(%d chars); treating them as absent.",
+                    tool_name or "<unknown>",
+                    len(arguments),
+                )
+                return self._fallback_arguments(tool_name, arguments)
+            if isinstance(parsed, dict):
+                return parsed
+            logger.warning(
+                "ReAct tool call %s returned non-object JSON arguments (%s).",
+                tool_name or "<unknown>",
+                type(parsed).__name__,
+            )
+            return self._fallback_arguments(tool_name, parsed)
         return {}
+
+    def _fallback_arguments(
+        self, tool_name: str | None, payload: Any
+    ) -> dict[str, Any]:
+        """Wrap unusable tool arguments, or drop them for control tools.
+
+        Work tools tolerate an opaque ``input`` passthrough. Control tools do
+        not: ``final_answer`` owns the run's only user-visible exit, so smuggling
+        a malformed payload through as ``input`` would silently strip ``answer``
+        and finalize the run with nothing to show. Returning empty args instead
+        lets ``_response_requires_tool_protocol_retry`` reject the call and spend
+        the pattern's one repair retry on it.
+        """
+
+        if tool_name in self._control_tool_names():
+            return {}
+        return {"input": payload}
 
     def _build_tool_schema(self, tool: Any) -> dict[str, Any]:
         name = self._tool_name(tool)
@@ -1684,6 +1776,9 @@ class ReActPattern(AgentPattern):
 
         if name == "final_answer":
             answer = str(args.get("answer", ""))
+            if not answer.strip():
+                self._reject_empty_final_answer(tool_call, context)
+                return None
             outcome = self._final_answer_outcome(args.get("outcome"))
             self._record_tool_call(
                 tool_call,
@@ -1825,6 +1920,41 @@ class ReActPattern(AgentPattern):
             }
 
         return None
+
+    def _reject_empty_final_answer(
+        self, tool_call: dict[str, Any], context: Any
+    ) -> None:
+        """Refuse to finalize on an empty ``final_answer`` and re-request one.
+
+        ``_response_requires_tool_protocol_retry`` catches this on the turn the
+        model produces it, so this guard covers the paths that reach the handler
+        without passing through response normalization — most importantly a
+        checkpoint resume that restores ``pending_tool_calls`` verbatim.
+
+        Returning ``None`` keeps the run in the loop; ``force_final_answer_next``
+        makes the next turn re-request ``final_answer``, and if that turn is empty
+        too the normalization guard fails the run instead of looping.
+        """
+
+        logger.warning(
+            "ReAct refusing to finalize on final_answer with no answer text; "
+            "re-requesting a final answer. tool_call_id=%s arg_keys=%s",
+            tool_call.get("id"),
+            sorted(self._tool_call_args_dict(tool_call)),
+        )
+        error = (
+            "final_answer was called with an empty answer, so the user received "
+            "nothing. Call final_answer again with the complete user-facing "
+            "response in its answer field."
+        )
+        self._record_tool_call(tool_call, status="failed", error=error)
+        context.add_tool_result(
+            tool_name="final_answer",
+            result={"error": error},
+            tool_call_id=tool_call.get("id"),
+        )
+        self.status = "thinking"
+        self.force_final_answer_next = True
 
     def _tool_is_concurrency_safe(self, name: str, tools: list[Any]) -> bool:
         """Whether ``name`` may run concurrently with other safe tools.
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -586,6 +586,9 @@ class ReActPattern(AgentPattern):
                         iteration=iteration,
                         answer_streamer=answer_streamer,
                         stream_failure_message=("invalid tool protocol after recovery"),
+                        empty_final_answer=(
+                            self._empty_final_answer_call(normalized) is not None
+                        ),
                     )
                 if answer_streamer is not None:
                     await answer_streamer.fail("invalid tool protocol, retrying")
@@ -599,7 +602,7 @@ class ReActPattern(AgentPattern):
                         "ReAct final_answer carried no answer text; discarding the "
                         "response and retrying. iteration=%s arg_keys=%s",
                         iteration,
-                        sorted(self._tool_call_args_dict(empty_final_answer)),
+                        sorted(self._tool_call_args_dict(empty_final_answer), key=str),
                     )
                 if recover_full_tool_set:
                     recovery_reason: str | None = "unavailable_tool_call"
@@ -648,6 +651,9 @@ class ReActPattern(AgentPattern):
                         iteration=iteration,
                         answer_streamer=answer_streamer,
                         stream_failure_message="invalid tool protocol after retry",
+                        empty_final_answer=(
+                            self._empty_final_answer_call(normalized) is not None
+                        ),
                     )
             if force_final_answer_now and not normalized.get("tool_calls"):
                 normalized["done"] = True
@@ -714,11 +720,24 @@ class ReActPattern(AgentPattern):
         iteration: int,
         answer_streamer: ReActFinalAnswerStreamer | None,
         stream_failure_message: str,
+        empty_final_answer: bool = False,
     ) -> dict[str, Any]:
+        """Abandon the run after the one repair attempt failed.
+
+        ``empty_final_answer`` distinguishes "the model never produced an answer"
+        from the status's other producers (provider protocol errors, mixed
+        control calls, a non-``final_answer`` tool on a forced turn), whose
+        ``error`` text is the only signal a caller has. Delegated-child
+        classification reads it to avoid collapsing all four into "never
+        produced an answer" - see ``agent_tool._classify_delegated_failure``.
+        """
+
         logger.warning(
-            "ReAct failing the run after an invalid tool protocol: %s (iteration=%s)",
+            "ReAct failing the run after an invalid tool protocol: %s "
+            "(iteration=%s empty_final_answer=%s)",
             stream_failure_message,
             iteration,
+            empty_final_answer,
         )
         if answer_streamer is not None:
             await answer_streamer.fail(stream_failure_message)
@@ -728,15 +747,20 @@ class ReActPattern(AgentPattern):
             pattern=self,
             metadata={"iteration": iteration},
         )
+        error = (
+            "The model called final_answer without an answer twice, so the run "
+            "produced no response."
+            if empty_final_answer
+            else "The model returned an invalid tool protocol response "
+            "after one repair attempt."
+        )
         return PatternResult(
             success=False,
-            error=(
-                "The model returned an invalid tool protocol response "
-                "after one repair attempt."
-            ),
+            error=error,
             metadata={
                 "iterations": iteration + 1,
                 "status": "invalid_tool_protocol",
+                "empty_final_answer": empty_final_answer,
             },
         ).to_dict()
 
@@ -1657,8 +1681,8 @@ class ReActPattern(AgentPattern):
                             "answer": {
                                 "type": "string",
                                 "description": (
-                                    "Complete user-facing answer. It must match "
-                                    "response_language. "
+                                    "Complete user-facing answer. It must be "
+                                    "non-empty and must match response_language. "
                                     f"{final_answer_language_rule()}"
                                 ),
                             },
@@ -1967,37 +1991,54 @@ class ReActPattern(AgentPattern):
         makes the next turn re-request ``final_answer``, and if that turn is empty
         too the normalization guard fails the run instead of looping.
 
-        Two deliberate differences from the fresh-turn path. The rest of the
-        batch is discarded, matching the fresh path's whole-response rejection:
-        otherwise a resumed ``[final_answer(""), work_tool]`` batch would execute
-        a side-effecting tool that the fresh path never runs. And the next turn is
-        forced to ``final_answer`` alone, where the fresh path restores the full
-        tool set - forcing guarantees termination and no tool re-execution after a
-        resume, at the cost that a resumed run with genuinely incomplete work must
-        answer or fail rather than continue that work.
+        Two deliberate differences from the fresh-turn path.
+
+        Sibling calls still pending are discarded, matching the fresh path's
+        whole-response rejection for the order that matters: a resumed
+        ``[final_answer(""), work_tool]`` batch would otherwise execute a
+        side-effecting tool that the fresh path never runs. The match is not
+        exact in the reverse order - in ``[work_tool, final_answer("")]`` the
+        work tool has already executed by the time this runs, and nothing here
+        undoes it.
+
+        And the next turn is forced to ``final_answer`` alone, where the fresh
+        path restores the full tool set. That bounds the run: a second empty
+        answer ends it as ``invalid_tool_protocol`` rather than looping. It does
+        not prevent tool re-execution outright - if the forced turn calls a
+        non-``final_answer`` tool, ``_requires_full_tool_set_recovery`` restores
+        the full set and clears the flag, so a discarded work tool can be
+        re-invoked on that turn.
         """
 
         logger.warning(
             "ReAct refusing to finalize on final_answer with no answer text; "
             "re-requesting a final answer. tool_call_id=%s arg_keys=%s",
             tool_call.get("id"),
-            sorted(self._tool_call_args_dict(tool_call)),
+            sorted(self._tool_call_args_dict(tool_call), key=str),
         )
         error = (
             "final_answer was called with an empty answer, so the user received "
             "nothing. Call final_answer again with the complete user-facing "
             "response in its answer field."
         )
-        self._record_tool_call(tool_call, status="failed", error=error)
+        # Carry the classification keys ``tool_result_succeeded`` reads, so this
+        # failure is not read back as a success, and record the same shape in the
+        # ledger as the cancelled siblings below.
+        failure = {"success": False, "status": "error", "error": error}
+        self._record_tool_call(tool_call, status="failed", result=failure, error=error)
         context.add_tool_result(
             tool_name="final_answer",
-            result={"error": error},
+            result=failure,
             tool_call_id=tool_call.get("id"),
         )
         # Leave only the rejected call for the caller to pop, and close out the
         # rest of the batch so no unexecuted call is left without a result.
-        remaining = self.pending_tool_calls[1:]
-        self.pending_tool_calls = self.pending_tool_calls[:1]
+        # Selected by identity rather than position: the caller happens to pass
+        # the head of the queue today, but nothing here depends on that.
+        remaining = [
+            pending for pending in self.pending_tool_calls if pending is not tool_call
+        ]
+        self.pending_tool_calls = [tool_call]
         self._cancel_tool_calls(
             remaining,
             context,
--- a/src/xagent/core/agent/pattern/react/react.py
+++ b/src/xagent/core/agent/pattern/react/react.py
@@ -621,6 +621,7 @@ class ReActPattern(AgentPattern):
                             force_final_answer_now and not recover_full_tool_set
                         ),
                         recovery_reason=recovery_reason,
+                        empty_final_answer=empty_final_answer is not None,
                     )
                 except LLMCallInterrupted:
                     interrupted = await self._interrupt_if_requested(
@@ -764,7 +765,7 @@ class ReActPattern(AgentPattern):
                 continue
             args = tool_call.get("args")
             if isinstance(args, dict):
-                return str(args.get("answer", ""))
+                return self._final_answer_text(args)
         return None
 
     def _messages_for_llm(
@@ -854,6 +855,7 @@ class ReActPattern(AgentPattern):
         tool_schemas: list[dict[str, Any]],
         force_final_answer: bool,
         recovery_reason: str | None = None,
+        empty_final_answer: bool = False,
     ) -> tuple[Any, ReActFinalAnswerStreamer]:
         tools = (
             [self._final_answer_tool_schema()] if force_final_answer else tool_schemas
@@ -885,13 +887,24 @@ class ReActPattern(AgentPattern):
             )
             retry_phase = "malformed_tool_arguments_recovery"
         elif recovery_reason == "empty_final_answer":
+            # On a forced turn ``tools`` above is final_answer alone, so offering
+            # a work tool would instruct the model to do something the schema
+            # forbids and waste the one repair attempt.
             retry_instruction = (
                 "The previous response called final_answer with an empty answer "
-                "field, so the user received no reply at all. Retry this turn: if "
-                "the task is complete, call final_answer again with the complete "
-                "user-facing response in its answer field; if work remains, call "
-                "the appropriate available work tool instead. Never call "
-                "final_answer with an omitted, empty, or whitespace-only answer."
+                "field, so the user received no reply at all. "
+                + (
+                    "final_answer is the only tool available on this turn: call it "
+                    "again with the complete user-facing response in its answer "
+                    "field."
+                    if force_final_answer
+                    else "Retry this turn: if the task is complete, call "
+                    "final_answer again with the complete user-facing response in "
+                    "its answer field; if work remains, call the appropriate "
+                    "available work tool instead."
+                )
+                + " Never call final_answer with an omitted, empty, or "
+                "whitespace-only answer."
             )
             retry_phase = "empty_final_answer_recovery"
         else:
@@ -903,6 +916,15 @@ class ReActPattern(AgentPattern):
                 "call final_answer only when the task is actually complete."
             )
             retry_phase = "tool_protocol_retry"
+        if empty_final_answer and recovery_reason != "empty_final_answer":
+            # A more fundamental repair owns the instruction, but the empty answer
+            # was still detected on this response and would otherwise go
+            # unmentioned.
+            retry_instruction = (
+                f"{retry_instruction} The previous response also called "
+                "final_answer with an empty answer field; never call final_answer "
+                "with an omitted, empty, or whitespace-only answer."
+            )
         messages[0] = {
             **messages[0],
             "content": f"{messages[0].get('content', '')}\n\n{retry_instruction}",
@@ -1016,10 +1038,12 @@ class ReActPattern(AgentPattern):
     def _final_answer_text(self, args: Any) -> str:
         """Coerce a ``final_answer`` argument payload into its answer text.
 
-        The single coercion for both this check and the finalization it guards,
-        so the two cannot disagree on what counts as an answer. ``None`` must
-        become the empty string rather than ``str(None)``, which would yield the
-        literal ``"None"`` and be sent to the user as the answer.
+        The single coercion for every site that reads an answer out of
+        ``final_answer`` args - this check, the finalization it guards, and the
+        streamer's candidate lookup - so they cannot disagree on what counts as
+        an answer. ``None`` must become the empty string rather than
+        ``str(None)``, which would yield the literal ``"None"`` and be sent to
+        the user as the answer.
         """
 
         if not isinstance(args, dict):
@@ -1453,46 +1477,33 @@ class ReActPattern(AgentPattern):
     def _normalize_tool_calls(self, tool_calls: list[Any]) -> list[dict[str, Any]]:
         normalized: list[dict[str, Any]] = []
         for index, tool_call in enumerate(tool_calls):
+            # Three provider shapes: a dict with a nested ``function`` payload, a
+            # flat dict, and an object with a ``function`` attribute. Each yields
+            # the same (id, name, arguments) triple.
             if isinstance(tool_call, dict):
+                call_id = tool_call.get("id")
                 function_payload = tool_call.get("function")
                 if isinstance(function_payload, dict):
-                    arguments = function_payload.get("arguments", {})
                     name = function_payload.get("name")
-                    normalized.append(
-                        {
-                            "id": tool_call.get("id") or f"tool_call_{index}",
-                            "name": name,
-                            "args": self._coerce_arguments(arguments, tool_name=name),
-                        }
-                    )
+                    arguments = function_payload.get("arguments", {})
+                else:
+                    name = tool_call.get("name")
+                    arguments = tool_call.get("args", tool_call.get("arguments", {}))
+            else:
+                function_payload = getattr(tool_call, "function", None)
+                if function_payload is None:
                     continue
-
-                name = tool_call.get("name")
-                normalized.append(
-                    {
-                        "id": tool_call.get("id") or f"tool_call_{index}",
-                        "name": name,
-                        "args": self._coerce_arguments(
-                            tool_call.get("args", tool_call.get("arguments", {})),
-                            tool_name=name,
-                        ),
-                    }
-                )
-                continue
-
-            function_payload = getattr(tool_call, "function", None)
-            if function_payload is not None:
+                call_id = getattr(tool_call, "id", None)
                 name = getattr(function_payload, "name", None)
-                normalized.append(
-                    {
-                        "id": getattr(tool_call, "id", None) or f"tool_call_{index}",
-                        "name": name,
-                        "args": self._coerce_arguments(
-                            getattr(function_payload, "arguments", {}),
-                            tool_name=name,
-                        ),
-                    }
-                )
+                arguments = getattr(function_payload, "arguments", {})
+
+            normalized.append(
+                {
+                    "id": call_id or f"tool_call_{index}",
+                    "name": name,
+                    "args": self._coerce_arguments(arguments, tool_name=name),
+                }
+            )
 
         return [call for call in normalized if call.get("name")]
 
@@ -1525,7 +1536,8 @@ class ReActPattern(AgentPattern):
             except json.JSONDecodeError:
                 logger.warning(
                     "ReAct tool call %s returned malformed JSON arguments "
-                    "(%d chars); treating them as absent.",
+                    "(%d chars); dropping them for control tools and passing "
+                    "them through as `input` otherwise.",
                     tool_name or "<unknown>",
                     len(arguments),
                 )
@@ -1545,12 +1557,19 @@ class ReActPattern(AgentPattern):
     ) -> dict[str, Any]:
         """Wrap unusable tool arguments, or drop them for control tools.
 
-        Work tools tolerate an opaque ``input`` passthrough. Control tools do
-        not: ``final_answer`` owns the run's only user-visible exit, so smuggling
-        a malformed payload through as ``input`` would silently strip ``answer``
-        and finalize the run with nothing to show. Returning empty args instead
-        lets ``_response_requires_tool_protocol_retry`` reject the call and spend
-        the pattern's one repair retry on it.
+        Work tools tolerate an opaque ``input`` passthrough, so they keep the
+        payload. Control tools drop it, because ``input`` is never a field any of
+        them declares: carrying it forward only disguises the loss as a populated
+        arguments object.
+
+        For ``final_answer`` dropping it is also load-bearing. That tool owns the
+        run's only user-visible exit, so a payload smuggled through as ``input``
+        would silently strip ``answer`` and finalize the run with nothing to
+        show; empty args instead let
+        ``_response_requires_tool_protocol_retry`` reject the call and spend the
+        pattern's one repair retry on it. ``send_message`` and
+        ``ask_user_question`` have no such guard - they degrade to an empty
+        message either way, exactly as they did before this branch existed.
         """
 
         if tool_name in self._control_tool_names():
@@ -1803,8 +1822,9 @@ class ReActPattern(AgentPattern):
                 result={"answer": answer, "outcome": outcome},
                 tool_call_id=tool_call.get("id"),
             )
-            if answer:
-                context.add_assistant_message(answer)
+            # Unconditional: the empty-answer rejection above is the only way
+            # into finalization, so an answer here always has text.
+            context.add_assistant_message(answer)
             return await self._finalize_outcome(
                 context=context,
                 runtime=runtime,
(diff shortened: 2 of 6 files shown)


src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1684,6 +1776,9 @@ async def _handle_control_tool(
 
         if name == "final_answer":
             answer = str(args.get("answer", ""))
+            if not answer.strip():
+                self._reject_empty_final_answer(tool_call, context)
+                return None
```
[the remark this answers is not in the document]
codeacme17: Confirmed and fixed in c607693 — thank you, this is a real hole in the very guard the PR adds.

Verified directly before changing anything:

```
args.get('answer', '') -> None
str(...)               -> 'None'
bypasses guard?        -> True
```

`{"answer": None}` takes the stored `None` rather than the `""` default, so `str()` yields `"None"`, which survives `.strip()` and finalizes as the user-facing answer. I added a failing test first, and it reproduced exactly that: `AssertionError: assert 'None' == 'The result is 4.'`.

Reachability is worth noting: `_empty_final_answer_call` already rejects a null answer during response normalization, so the only way to reach the handler with this payload is the checkpoint-resume path — which is precisely the path `_reject_empty_final_answer` exists to cover. So the hole was in the guard's one unique responsibility.

I implemented the fix slightly differently from the suggested snippet. Reassigning `answer` after the fact leaves two independent coercions in place, which is what allowed the drift in the first place; my comment on `_empty_final_answer_call` even claimed the two sites coerced alike, and it was wrong. Instead both now go through one helper:

```python
def _final_answer_text(self, args: Any) -> str:
    if not isinstance(args, dict):
        return ""
    answer = args.get("answer")
    if answer is None:
        return ""
    return str(answer)
```

`_handle_control_tool` uses `answer = self._final_answer_text(args)` and `_empty_final_answer_call` uses the same call, so they cannot disagree again. This also covers a non-dict `args`, which the previous `_empty_final_answer_call` handled but the handler did not.

Tests: the resume-path test is now parametrized over `missing` / `empty-string` / `whitespace` / `none`, and `test_final_answer_text_treats_none_as_absent` pins the coercion contract, including that a scalar `0` stays `"0"` rather than being mistaken for absent.

codeacme17:
Thanks for the review. The one finding was valid and is fixed in c607693; details in the [inline thread](https://github.com/xorbitsai/xagent/pull/1325#discussion_r3773093330).

Summary: `_handle_control_tool` coerced the answer with `str(args.get("answer", ""))`, so a payload of `{"answer": None}` became the literal string `"None"`, passed the empty check, and was finalized as the user-facing answer. `_empty_final_answer_call` handled null correctly, so the two sites disagreed on exactly the payload this PR's guard exists to catch — reachable via the checkpoint-resume path, which is the one path that reaches the handler without going through response normalization.

Rather than patch the handler's check alone, both sites now share a single `_final_answer_text` coercion, so they cannot drift again. Added a parametrized regression test over `missing` / `empty-string` / `whitespace` / `none`, written to fail first, plus a unit test pinning the coercion contract.

rogercloud (changes requested):
## Summary

PR #1325 (fixes #1312) makes the ReAct pattern reject a `final_answer` call that carries an empty answer: the offense is routed through the existing repair-retry channel (`_response_requires_tool_protocol_retry` → `_retry_tool_protocol_response` → `_invalid_tool_protocol_result`) instead of finalizing silently, a handler-level backstop covers the checkpoint-resume path that bypasses normalization, `_coerce_arguments` no longer smuggles malformed control-tool JSON through as `{"input": ...}`, and warning logs are added at detection, retry exhaustion, and placeholder substitution. A follow-up commit (c6076934) treats a null `answer` as absent via a shared `_final_answer_text` coercion, closing the literal-"None" finalization hole. Bounding is explicit and verified: one repair retry plus a forced final-answer turn prevents unbounded spin.

Only one commit, 658b526b, was pushed since the previous review: a test-only change relocating the empty-final-answer child-failure assertions in `tests/core/tools/test_agent_tool_session_isolation.py` from the parent classifier (`missing_delegated_output`) to the child's own pattern-level failure. It asserts fail-closed behavior, exactly two LLM calls, and no preamble laundering; production code is unchanged since c6076934, and CI is green.

## Approach verdict

Sound. The fix reuses the pre-existing single repair-retry machinery and one shared coercion helper (`_final_answer_text`) between detection and finalization, keeps the retry bounded, and adds the handler-level guard exactly where the resume path bypasses normalization. The declared non-goals (schema reordering, adapter success semantics) are consistent with the documented delegated-child classifier contract.

## Prior findings checklist

- **P1 (None-answer hole)** — **FIXED** in c6076934 via the shared `_final_answer_text` coercion (None → ""; non-dict args → ""), used by both the handler (react.py:1790) and `_empty_final_answer_call` (react.py:1012). Regression-pinned by `test_final_answer_text_treats_none_as_absent` and the parametrized resume test (`none` case). Verified against HEAD: genuinely fixed, scalar `answer=0` keeps pre-PR behavior.
- The second automated review (after c6076934) returned no findings.

## Findings

### Major

**N1 — `src/xagent/core/agent/execution_adapter.py:369-382` — major**

The new `if not output:` warning fires on every legitimate `waiting_for_user` pause: send_message/expect_response, ask_user_question, tool_waiting_for_user, and resume-recheckpoint all return success=False, status="waiting_for_user", with only message/message_type/interactions keys. `_latest_assistant_message` typically returns None on tool-call turns (assistant content is stored empty), so this is the common first-turn-clarification path — including chat.py → service.py → `_normalize_result` and delegated children — not a pattern bug. The comment at lines 370-371 claims the condition means "a run that reports success ... a bug in the pattern", but the code never checks `success`. The warning therefore fires on the most common pause path, asserting a non-existent bug and diluting the genuine diagnostic the PR adds. Fix: gate on success/status (e.g. `if not output and result.get("success"):`, or skip when status == "waiting_for_user") and add a test that a paused result stays silent.

### Minor

**N2 — `src/xagent/core/agent/pattern/react/react.py:767, 1019-1020` — minor**

`_final_answer_tool_content` still coerces independently (`str(args.get("answer", ""))`) while the new `_final_answer_text` docstring claims to be "the single coercion for both this check and the finalization it guards". For `{"answer": None}` the two disagree ("None" vs "") — unreachable today only because the upstream guard rejects every empty shape first, so the invariant holds transitively, not structurally. Suggest delegating `_final_answer_tool_content` to `_final_answer_text`, or scoping the docstring claim.

**N3 — `src/xagent/core/agent/pattern/react/react.py:887-896` — minor**

The `empty_final_answer` retry instruction says "if work remains, call the appropriate available work tool instead", but on a forced final-answer turn `_retry_tool_protocol_response` (858-860) narrows the tool set to final_answer only with tool_choice="required" — the offered option is impossible. The model is instructed to do something the tool set forbids, wasting the repair attempt. Tailor the instruction by the force flag: on a forced turn, tell the model it must answer now.

**N4 — `src/xagent/core/agent/pattern/react/react.py:1936-1969, 2295` — minor**

The resume path rejects only the offending call; `_execute_pending_tool_calls` then pops it and executes the remaining batched pending calls (e.g. side-effecting work tools), whereas the fresh-turn path discards the whole batch. A mixed `[final_answer(""), work_tool]` batch therefore runs the work tool after resume but never on the fresh path — and pre-fix checkpoints are exactly the input this guard exists for, with mixed control/work batches an anticipated shape. Consider discarding the rest of the batch after rejection, mirroring `_discard_pending_tool_plan_after_pause` semantics.

**N5 — `src/xagent/core/agent/pattern/react/react.py:733-734` / `src/xagent/core/tools/adapters/vibe/agent_tool.py:1647-1665` — minor**

Delegated children now fail with the internal string "The model returned an invalid tool protocol response after one repair attempt." surfaced verbatim to the parent model with no failure_code; the friendly `missing_delegated_output` classification (allowlisted in TOOL_FAILURE_CODES) is now unreachable for this scenario. Fail-closed is preserved, but the parent-facing message quality regressed, and commit 658b526b accepts the relocation without sanitizing it. Consider mapping `invalid_tool_protocol` child failures to a parent-friendly message.

**N6 — `src/xagent/core/agent/pattern/react/react.py:604-608` — minor**

When `recover_full_tool_set` and an empty final_answer co-occur, `unavailable_tool_call` wins the elif chain and the tailored empty-answer instruction (893-899) is never used, even though the empty answer was detected and logged just above. The model receives the generic "tool unavailable" guidance instead of the actionable "final_answer was empty" guidance. Compose the two instructions or prioritize the more specific one.

**N8 — `src/xagent/core/agent/pattern/react/react.py:1526-1531` — minor**

The malformed-JSON warning says arguments are "treating them as absent", but work tools keep the payload wrapped as `{"input": raw}` via `_fallback_arguments` — preserved, not absent; only control tools drop it. The adjacent non-object-JSON warning (1532-1535) makes no such claim, so the two diagnostics disagree about what actually happens next. Qualify the phrase ("treating them as absent for control tools") or drop it.

**N9 — `src/xagent/core/agent/pattern/react/react.py:1806` — minor**

The `if answer:` guard before `context.add_assistant_message(answer)` is now dead: `_reject_empty_final_answer` (1791-1793) returns before any empty answer can reach it. The vestigial conditional reads as if empty answers could still slip through to finalization, undermining the guarantee the PR is establishing. Remove the guard and add the assistant message unconditionally.

**N10 — `src/xagent/core/agent/pattern/react/react.py:1546-1558` — minor**

`_fallback_arguments` drops malformed payloads for ALL control tools (final_answer, send_message, ask_user_question), but the docstring rationale names only final_answer, and its mechanism claim — "lets `_response_requires_tool_protocol_retry` reject the call and spend the pattern's one repair retry on it" — is factually wrong for send_message/ask_user_question: `_empty_final_answer_call` skips non-final_answer calls, so a malformed send_message with `{}` args reaches the handler and sends an empty message (unchanged from base behavior). The docstring misdescribes the actual behavior; fix it, and consider a test for ask_user_question's fallback path.

### Body-only notes (non-blocking)

**N7 — tests — minor** — Untested: (a) the new adapter warning — no test asserts it fires on the true positive, and none asserts a `waiting_for_user` pause stays silent (and today it does not); (b) a mixed `[work_tool, final_answer(empty)]` batch is never exercised end-to-end; (c) scalar `answer=0` is only unit-tested via the helper, not through the run loop.

**N11 — `react.py:652`** — Adjacent pre-existing hole, outside #1312's scope: a forced final-answer turn that returns ZERO tool calls still finalizes success=True with an empty response. It is now traceable — the empty result flows through `_normalize_result`, where the new adapter warning fires — but the run still reports success. Worth an optional follow-up (treat a zero-call forced turn as a protocol violation or log it); not a blocker.

**N12 — `react.py:1969`** — The fresh-turn offense retries with the full tool set (when not forced), matching the "call the appropriate work tool" instruction; the resume-path offense forces a final_answer-only next turn, so a resumed run with genuinely incomplete work cannot continue it and must answer or fail after one retry. The asymmetry is a deliberate bounding trade-off (forcing guarantees deterministic termination and no tool re-execution after resume) but is documented nowhere; a sentence in the docstring suffices.

**PR-body nit** — The PR body refers to `_validate_tool_protocol`, which does not exist; the method is `_response_requires_tool_protocol_retry`.

- `L1453: shrink: three near-identical normalized.append blocks rebuild the same {id,name,args} shape; the PR threaded tool_name through all three without deduplicating. Extract (call_id, name, arguments) per tool_call shape with early continue, then a single normalized.append(...) (~64 to ~44 lines).`
- S2 was considered and kept: `_fallback_arguments`' docstring carries the non-obvious control-tool/work-tool contract, so the helper stays.
- net: -20 lines possible

Blocking: yes because N1 is a confirmed major finding [new]. Recommended event: REQUEST_CHANGES.

Blocking list:
- `src/xagent/core/agent/execution_adapter.py:369` — major — 'produced no output' warning fires on every legitimate waiting_for_user pause and asserts a non-existent pattern bug; gate on success/status — [new]

src/xagent/core/agent/execution_adapter.py
```diff
@@ -366,6 +366,20 @@ def _normalize_result(
                 # ``agent_result`` specifically so a backfilled preamble
                 # cannot be mistaken for a real final answer.
                 output = self._latest_assistant_message(result.get("context"))
+        if not output:
+            # A run that reports success with nothing to show is a bug in the
+            # pattern that produced it, and the placeholder erases the evidence.
+            # Log before substituting so there is something to debug from.
```
rogercloud, line 372: [major] This new `if not output:` warning fires on every legitimate pause: waiting_for_user results (send_message/expect_response, ask_user_question, tool_waiting_for_user, resume-recheckpoint) return success=False, status="waiting_for_user", with only message/message_type/interactions keys, and `_latest_assistant_message` typically returns None on tool-call turns. The comment at lines 370-371 claims the condition means "a run that reports success ... a bug in the pattern", but nothing here checks `success` — this is the common first-turn-clarification path, not a bug. Gate the warning on success/status, e.g. `if not output and result.get("success"):` or skip when status == "waiting_for_user", and add a test that a paused result stays silent.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -960,7 +990,44 @@ def _response_requires_tool_protocol_retry(
...
+            if tool_call.get("name") != "final_answer":
+                continue
+            if not self._final_answer_text(tool_call.get("args")).strip():
+                return tool_call
+        return None
+
+    def _final_answer_text(self, args: Any) -> str:
+        """Coerce a ``final_answer`` argument payload into its answer text.
+
+        The single coercion for both this check and the finalization it guards,
```
rogercloud: [minor] The docstring claims `_final_answer_text` is "the single coercion for both this check and the finalization it guards", but `_final_answer_tool_content` (react.py:767) still coerces independently via `str(args.get("answer", ""))`. For `{"answer": None}` the two disagree ("None" vs "") — unreachable today only because the upstream guard rejects every empty shape first, so the invariant holds transitively rather than structurally. Delegate `_final_answer_tool_content` to `_final_answer_text`, or scope the docstring claim.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -864,6 +884,16 @@ async def _retry_tool_protocol_response(
                 "complete user-facing response in its answer field."
             )
             retry_phase = "malformed_tool_arguments_recovery"
+        elif recovery_reason == "empty_final_answer":
```
rogercloud, line 921: [minor] The `empty_final_answer` retry instruction (lines 893-896) tells the model "if work remains, call the appropriate available work tool instead", but on a forced final-answer turn `_retry_tool_protocol_response` (858-860) narrows the tool set to final_answer only with tool_choice="required" — the offered option is impossible. The model is asked to do something the tool set forbids, wasting the repair attempt. Tailor the instruction by the force flag: on a forced turn, tell the model it must answer now.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1826,6 +1933,41 @@ async def _handle_control_tool(
...
 
+    def _reject_empty_final_answer(
+        self, tool_call: dict[str, Any], context: Any
+    ) -> None:
+        """Refuse to finalize on an empty ``final_answer`` and re-request one.
+
+        ``_response_requires_tool_protocol_retry`` catches this on the turn the
+        model produces it, so this guard covers the paths that reach the handler
+        without passing through response normalization — most importantly a
+        checkpoint resume that restores ``pending_tool_calls`` verbatim.
```
rogercloud, line 1996: [minor] The resume path rejects only the offending call; `_execute_pending_tool_calls` then pops it and executes the remaining batched pending calls (e.g. side-effecting work tools), whereas the fresh-turn path discards the whole batch. A mixed `[final_answer(""), work_tool]` batch therefore runs the work tool after resume but never on the fresh path — and pre-fix checkpoints are exactly the input this guard exists for. Consider discarding the rest of the batch after rejection, mirroring `_discard_pending_tool_plan_after_pause` semantics.

tests/core/tools/test_agent_tool_session_isolation.py
```diff
@@ -977,12 +977,15 @@ async def test_agent_tool_real_child_with_empty_final_answer_fails_closed(
...
     """A real ReAct child that answers with an empty ``final_answer`` fails closed.
 
-    The classifier reads the child's raw ``""`` final answer directly, so
-    this scenario never touches the placeholder constants at all; it is
-    ``test_agent_tool_completed_child_without_usable_output_fails_closed``'s
-    parametrized rows (which do exercise the literal placeholder strings)
-    that serve as the drift detector for ``NO_OUTPUT_PLACEHOLDER`` /
-    ``NO_RESPONSE_PLACEHOLDER``.
+    ReAct now rejects an empty ``final_answer`` as a tool-protocol violation and
+    spends its one repair retry on it, so this scenario fails inside the child's
```
rogercloud: [minor] This relocation pins the child's pattern-level failure, but the behavior it accepts still surfaces the internal string "The model returned an invalid tool protocol response after one repair attempt." verbatim to the parent model with no failure_code — the friendly `missing_delegated_output` classification (allowlisted in TOOL_FAILURE_CODES) is now unreachable for this scenario. Fail-closed is preserved; the parent-facing message quality regressed. Consider mapping `invalid_tool_protocol` child failures to a parent-friendly message and pinning that here.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -590,6 +593,20 @@ async def _run_tool_calling_loop(
...
                 )
+                empty_final_answer = self._empty_final_answer_call(normalized)
+                if empty_final_answer is not None:
+                    logger.warning(
+                        "ReAct final_answer carried no answer text; discarding the "
+                        "response and retrying. iteration=%s arg_keys=%s",
+                        iteration,
+                        sorted(self._tool_call_args_dict(empty_final_answer)),
+                    )
+                if recover_full_tool_set:
```
rogercloud, line 615: [minor] When `recover_full_tool_set` and an empty final_answer co-occur, this elif chain lets `unavailable_tool_call` win and the tailored empty-answer instruction (lines 893-899) is never used — even though the empty answer was detected and logged just above. The model gets the generic "tool unavailable" guidance instead of the actionable "final_answer was empty" guidance. Compose the two instructions or prioritize the more specific one (`empty_final_answer`).

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1442,17 +1514,49 @@ def _remember_tool_call_content(
...
+    ) -> dict[str, Any]:
         if isinstance(arguments, dict):
             return arguments
         if isinstance(arguments, str):
             try:
                 parsed = json.loads(arguments)
             except json.JSONDecodeError:
-                return {"input": arguments}
-            return parsed if isinstance(parsed, dict) else {"input": parsed}
+                logger.warning(
```
rogercloud, line 1569: [minor] This warning says malformed JSON arguments are "treating them as absent", but only control tools drop the payload: work tools keep it wrapped as `{"input": raw}` via `_fallback_arguments`, so the arguments are preserved, not absent. The adjacent non-object-JSON warning (lines 1532-1535) makes no such claim, so the two diagnostics disagree about what happens next. Qualify the phrase ("treating them as absent for control tools") or drop it.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1683,7 +1787,10 @@ async def _handle_control_tool(
         args = tool_call.get("args", {})
 
         if name == "final_answer":
-            answer = str(args.get("answer", ""))
+            answer = self._final_answer_text(args)
```
rogercloud, line 1841: [minor] The `if answer:` guard before `context.add_assistant_message(answer)` (a few lines below) is now dead: `_reject_empty_final_answer` returns from this branch before any empty answer can reach it. The vestigial conditional reads as if empty answers could still slip through to finalization, undermining the guarantee this PR establishes. Remove the guard and add the assistant message unconditionally.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1442,17 +1514,49 @@ def _remember_tool_call_content(
...
+                tool_name or "<unknown>",
+                type(parsed).__name__,
+            )
+            return self._fallback_arguments(tool_name, parsed)
         return {}
 
+    def _fallback_arguments(
+        self, tool_name: str | None, payload: Any
+    ) -> dict[str, Any]:
+        """Wrap unusable tool arguments, or drop them for control tools.
```
rogercloud, line 1590: [minor] This docstring names only `final_answer`, but `_fallback_arguments` drops malformed payloads for all control tools (final_answer, send_message, ask_user_question); its mechanism claim — "lets `_response_requires_tool_protocol_retry` reject the call and spend the pattern's one repair retry on it" — is factually wrong for send_message/ask_user_question: `_empty_final_answer_call` skips non-final_answer calls, so a malformed send_message with `{}` args reaches the handler and sends an empty message (unchanged from base behavior). Fix the docstring to describe the actual per-tool behavior, and consider a test for ask_user_question's fallback path.

codeacme17:
Thank you — this was an unusually precise review, and N1 and N5 both caught real defects I had introduced. Everything is addressed across three commits: 04fd9fe (N1), dc037b7 (N5), bdc779d (N2, N3, N4, N6, N8, N9, N10, N12, S1). Only N11 is deliberately deferred.

## N1 — blocking, major — fixed in 04fd9fe

Confirmed by reproducing it rather than reasoning about it:

```
WARN> Agent 'child' produced no output; substituting the placeholder.
      execution_type=agent_react pattern=react status=waiting_for_user success=False task_id=t1
```

You were right on both counts: the warning fired on the ordinary first-turn clarification, and the comment asserted a `success` check the code never performed. Now gated on `result.get("success")`, with the comment stating what is actually checked and why emptiness alone is not a defect.

Tests added: the true positive warns exactly once, and both pause shapes (`send_message`/`expect_response` and `ask_user_question` with interactions) stay silent.

## N5 — fixed in dc037b7, and it recovered the classification

This was the most valuable finding. Rather than only sanitizing the message, I added `_CHILD_NO_ANSWER_STATUSES = {"invalid_tool_protocol"}` to the classifier, mapping it to a parent-facing message **and** the allowlisted `missing_delegated_output` code — the child produced no answer either way, so it is the same classification the completed-but-empty path already uses, reached by a different branch.

Consequence worth flagging: this **restores the `failure_code` assertions** that 658b526 had relocated. Your instinct that giving up the classification was a regression was better than my accepting it; the test churn in that commit was largely unnecessary and the assertions are back.

I scoped the status set to `invalid_tool_protocol` only. `max_iterations` has the same shape but predates this PR, so folding it in would change behavior this PR did not cause.

## N4 — fixed in bdc779d

Agreed, and the asymmetry was worse than cosmetic: a pre-fix checkpoint is exactly this guard's input, so a `[final_answer(""), work_tool]` batch would execute a side effect that the same model output never produces on a live turn. Rejection now discards the rest of the batch. Extracted `_cancel_tool_calls` (shared with `_discard_pending_tool_plan_after_pause`) so every abandoned call still gets exactly one result, preserving I2. New end-to-end test asserts the work tool never runs and the abandoned call is recorded `cancelled`.

## N3, N6 — fixed in bdc779d

N3: the instruction now branches on `force_final_answer`; on a forced turn it says final_answer is the only available tool instead of offering a work tool the schema forbids. N6: when a more fundamental repair owns the instruction, the empty-answer sentence is appended rather than dropped.

## N2, N8, N9, N10, N12 — fixed in bdc779d

- **N2** `_final_answer_tool_content` now delegates to `_final_answer_text`, so the "single coercion" claim is structural rather than transitively true via the upstream guard. Docstring names all three call sites.
- **N8** Reworded to state both branches: dropped for control tools, passed through as `input` otherwise.
- **N9** Removed; the rejection above is the only route into finalization.
- **N10** Docstring rewritten. It now separates the load-bearing `final_answer` rationale from `send_message`/`ask_user_question`, which degrade to an empty message exactly as they did before this branch existed.
- **N12** Documented in `_reject_empty_final_answer`.

## N7 — tests added

(a) both directions of the adapter warning; (b) the mixed resumed batch end-to-end; (c) scalar `answer=0` through the run loop. On (c): it finalizes as `"0"` with no retry spent, though nothing is streamed, because `ToolCallStringFieldStreamer` only tracks string-typed args fields and `answer` is declared `type: string`. Off-schema and pre-existing, so I left it alone.

## S1 — applied in bdc779d

The three provider shapes now resolve `(call_id, name, arguments)` and share one `append`: 44 → 31 lines.

## N11 — deferred, not disputed

A forced turn returning zero tool calls still finalizes `success=True` with an empty response. It is the same class of bug as #1312 and worth fixing, but it predates this PR and has a different trigger, so it belongs in its own issue rather than growing this one. Happy to open it, or to fold it in here if you would rather it not wait.

## PR-body nit

Fixed — the method is `_response_requires_tool_protocol_retry`; `_validate_tool_protocol` does not exist. The body also now notes the forced-turn instruction variant and the `success` gate.

## Verification

`tests/core` in full, diffed against a baseline captured by reverting the branch and re-running the identical command: **zero regressions**. All pre-commit hooks pass (ruff 0.12.3, mypy 1.19.0). Awaiting CI.

src/xagent/core/agent/execution_adapter.py
codeacme17: Fixed in 04fd9fe. Confirmed by reproducing it rather than reasoning about it:

```
WARN> Agent 'child' produced no output; substituting the placeholder.
      execution_type=agent_react pattern=react status=waiting_for_user success=False task_id=t1
```

Right on both counts — it fired on the ordinary first-turn clarification, and the comment asserted a `success` check the code never performed. Now `if not output and result.get("success")`, with the comment stating what is actually checked and why emptiness alone is not a defect.

Tests: the true positive warns exactly once, and both pause shapes (`send_message`/`expect_response`, and `ask_user_question` with interactions) stay silent. The pause test fails without the gate.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d. `_final_answer_tool_content` now delegates to `_final_answer_text` rather than coercing independently, so the claim is structural instead of transitively true via the upstream guard — which, as you note, is exactly the kind of invariant that rots. Docstring updated to name all three call sites (this check, finalization, and the streamer's candidate lookup).

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d. The instruction now branches on `force_final_answer`: on a forced turn it says final_answer is the only tool available on this turn, instead of offering a work tool the narrowed schema forbids. Comment added at the branch explaining why, so the next editor does not merge the two wordings back together.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d, and I agree it was worse than cosmetic — a pre-fix checkpoint is precisely this guard's input, so the batch would replay a side effect the same model output never produces on a live turn.

Rejection now discards the rest of the batch. I extracted `_cancel_tool_calls`, shared with `_discard_pending_tool_plan_after_pause`, so each abandoned call still gets exactly one result and I2 holds. `force_final_answer_next` is deliberately left set, unlike the pause path which clears it.

New end-to-end test asserts the work tool never runs and the abandoned call is recorded `cancelled`.

tests/core/tools/test_agent_tool_session_isolation.py
codeacme17: Addressed in dc037b7 — and this was the most useful finding, because the right fix made the relocation unnecessary.

Instead of accepting the lost classification, I added `_CHILD_NO_ANSWER_STATUSES = {"invalid_tool_protocol"}` to the classifier, mapping it to a parent-facing message and the allowlisted `missing_delegated_output` code. The child produced no answer either way, so it is the same classification the completed-but-empty path already uses, reached by a different branch.

So the `failure_code` assertions are **restored** in both tests, alongside the additions from 658b526 (retry attempted, no preamble laundering). Your instinct that surrendering the classification was a regression was better than my accepting it.

Scoped to `invalid_tool_protocol` only: `max_iterations` has the same shape but predates this PR, so folding it in would change behavior this PR did not cause.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d by composing rather than reordering. `unavailable_tool_call` still owns the instruction — restoring the tool set is the more fundamental repair — but when an empty final_answer was also detected, its sentence is appended. Threaded through a new `empty_final_answer` parameter on `_retry_tool_protocol_response` so the composition is explicit at the call site rather than inferred from the reason string.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d. Reworded to state both branches: "dropping them for control tools and passing them through as `input` otherwise". The two adjacent diagnostics no longer disagree about what happens next.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d — removed, and replaced with a comment noting that the rejection above is the only route into finalization, so an answer here always has text. You are right that leaving it read as if empty answers could still slip through, which undercuts the guarantee the PR is establishing.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in bdc779d. The docstring now separates the two cases: dropping is load-bearing for `final_answer` (the guard rejects the call and spends the repair retry), while `send_message`/`ask_user_question` have no such guard and degrade to an empty message either way — unchanged from the behavior before this branch existed. It also states the general reason control tools drop the payload: `input` is not a field any of them declares, so carrying it forward only disguises the loss as a populated arguments object.

rogercloud (approved):
## Follow-up review — empty-`final_answer` guard for the ReAct pattern

### PR summary

This PR makes the ReAct pattern refuse to finalize on an empty `final_answer`. Previously a model could end the run with a null or blank answer and no text reaching the user; now empty answers are routed through the existing single-repair tool-protocol retry (`_response_requires_tool_protocol_retry` → `_retry_tool_protocol_response` → `_invalid_tool_protocol_result`), the checkpoint-resume path gets a handler-level guard (`_reject_empty_final_answer`), `_coerce_arguments` no longer launders malformed JSON into `{"input": ...}`, and a shared `_final_answer_text` coercion keeps empty-answer detection and finalization from drifting apart. Child agents that end without an answer are surfaced to the parent as `missing_delegated_output` with an actionable message, and the adapter's missing-output warning only fires on successful runs.

### Update summary

Three commits arrived since the last review. `04fd9fe` gates the adapter's missing-output warning on a successful run (fixes N1). `dc037b7` maps child agents that never answer to `missing_delegated_output` with a parent-friendly message (fixes N5). `bdc779d` sharpens the empty-`final_answer` guard: forced-turn instructions now distinguish empty-answer from unavailable-tool recovery, the resume guard applies whole-response-rejection semantics to the rest of a resumed batch, and the normalization/ledger paths are tightened (fixes N2/N3/N4/N6/N8/N9/N10/N12/S1).

### Prior findings checklist

- N1 — FIXED: adapter warning is success-gated, with tests on both sides of the gate.
- N2 — FIXED: `_final_answer_tool_content` now delegates to the shared `_final_answer_text`.
- N3 — FIXED: forced-turn instructions distinguish empty-answer from unavailable-tool recovery.
- N4 — FIXED: the rest of a rejected batch is cancelled with the I2 invariant preserved and tested; residual order limitation → F2.
- N5 — FIXED: `_CHILD_NO_ANSWER_STATUSES` mapping added with tests; scope concern → F3.
- N6 — FIXED: synthetic-suffix branch handled.
- N7(a) — FIXED: true-positive warning test added; `waiting_for_user` silence parametrized.
- N7(b) — PARTIAL: mixed batch only tested in `[final_answer(""), work_tool]` order; reverse order untested → folded into F2.
- N7(c) — FIXED: scalar-zero answer exercised through the run loop.
- N8 — FIXED: warning wording clarified.
- N9 — FIXED: `add_assistant_message` is unconditional.
- N10 — FIXED: docstring made accurate.
- N11 — WAIVED: zero-call forced turns still end `success=True` — pre-existing baseline behavior, previously judged a non-blocking optional follow-up; the author deferred with sound reasoning.
- N12 — FIXED: the fresh/resume asymmetry is documented; residual escape-hatch concern → F1.
- P1 — FIXED: the `None`-answer leak is closed.
- PR-body nit — FIXED: method name corrected.
- S1 — FIXED: `_normalize_tool_calls` now appends once (−20 lines).

### New findings (all minor)

**F1 — docstring overstates termination / no-re-execution [minor]**
`src/xagent/core/agent/pattern/react/react.py:1969-1976` claims forcing "guarantees termination and no tool re-execution after a resume" and that the run "must answer or fail". Both are broken by the escape hatch: if the forced turn emits a non-`final_answer` tool, `_requires_full_tool_set_recovery` (react.py:1068-1071) restores the full tool set and resets `force_final_answer_next` (react.py:635-637), so a previously cancelled work tool can be re-invoked. Termination itself still holds — a second violation ends in `invalid_tool_protocol`. Suggest wording that names the hatch, e.g. "...unless the forced turn calls a non-final_answer tool, in which case the full tool set is restored and the guarantee lapses for that turn."

**F2 — "whole-response rejection" equivalence holds only for one batch order [minor]**
`react.py:1969-1976` and `react.py:1999-2000`: on a resumed `[work_tool, final_answer("")]` batch, the work tool executes before the rejection point; `_reject_empty_final_answer` only cancels `pending[1:]`, so matching the fresh path's whole-response rejection is exact only when the empty `final_answer` precedes side-effecting calls. Reachability is limited to stale/pre-fix checkpoints — a live turn rejects the whole response first — but the claim as written is stronger than the behavior. Suggest a docstring caveat and optionally a reverse-order test; the current test only covers `[final_answer(""), work_tool]` (`tests/core/agent/test_react.py:5416-5454`).

**F3 — status mapping over-broad: child error text discarded [minor]**
`src/xagent/core/tools/adapters/vibe/agent_tool.py:1530-1534` maps every `invalid_tool_protocol` child failure to "never produced an answer" + `missing_delegated_output`, including three pre-existing producers whose answers were non-empty (provider protocol errors, mixed control calls, wrong tool in a forced turn). The child's own error text — e.g. "The model returned an invalid tool protocol response after one repair attempt." — is dropped, and it is the parent's only clue distinguishing "child model misbehaved" from "child had nothing to say". The comment's claim that this error text "names runtime internals" is inaccurate for this status; it is plain English about model behavior, and `invalid_tool_protocol` with all its producers predates this PR. Suggest narrowing the mapping with a marker for genuinely empty answers (e.g. only map when the failure carries the empty-answer error text), or folding the child's error text into the trace event so the parent keeps the diagnostic.

**F7 — resume tests bypass real serialization [minor]**
`tests/core/agent/test_react.py:5322-5323, 5428-5429`: the resume regression tests mutate `pattern.status` / `pattern.pending_tool_calls` directly instead of going through `get_state`/`load_state`, so the post-rejection state (`force_final_answer_next=True`) is never proven serializable and recoverable through the real mechanism. Neighboring tests (e.g. ~3501, ~4356) use the real round-trip. Suggest driving at least one resume test through the state round-trip.

**F8 — `sorted(...)` on mixed-type arg keys raises TypeError [minor]**
`react.py:602` and `react.py:1983`: the two new `logger.warning` calls sort `self._tool_call_args_dict(...)`, which raises `TypeError` on mixed-type keys — reachable with non-JSON programmatic payloads or hand-crafted checkpoints (real providers use string keys). Because `run()` re-raises from its `except`, this would abort the run exactly at the intended recovery point. `_args_hash` (react.py:3128-3135) already guards this pattern. Suggest `sorted(..., key=str)` at both sites.

**F9 — whole-response rejection discards produced preamble [minor]**
`react.py:581-668`: when the fresh path rejects a response containing non-empty `assistant_content` plus an empty `final_answer`, the preamble is dropped with the whole response and the retry rebuilds only from context — the produced preamble is structurally invisible to the retry. If the model repeats the "preamble + empty final_answer" pattern, the run flips to a hard `invalid_tool_protocol` failure and the user never sees the preamble. Suggest carrying the discarded preamble into the retry instruction/context, or explicitly accepting and documenting the tradeoff; optionally add a content-plus-empty-answer test variant to pin the behavior.

**F10 — `[:1]`/`[1:]` slices assume the rejected call is first [minor]**
`react.py:1999-2000`: the slices assume the rejected call is `pending_tool_calls[0]`. This holds structurally today — `_next_segment` always cuts the control tool as the first call, and the only caller passes `segment[0]` — but a future caller passing a non-first call would silently keep earlier calls, cancel the wrong tail, and leave the rejected call with both `failed` and `cancelled` records, breaking the I2 invariant. Suggest a zero-cost assert, or passing `remaining` explicitly.

**F11 — rejected `final_answer` recorded as success; ledger shape mismatch [minor]**
`react.py:1991-1996`: the context tool result `{"error": error}` is classified as success by `tool_result_succeeded` (`src/xagent/core/agent/result.py:67-75` — no `success: False` / `is_error` / `status == "error"` key), and `_record_tool_call(..., status="failed", error=error)` stores `result=None` while the sibling `_cancel_tool_calls` path stores a `{"success": False, "status": "cancelled", "error": reason}` dict. Latent today because `force_final_answer_next=True` short-circuits further handling, but a future consumer would see a failure recorded as success with an inconsistent ledger shape. Suggest adding a classification key and aligning the ledger record with the cancelled path.

**F12 — no test for the full resume→reject→forced→empty→retry→abandon chain [minor]**
Test gap: the resume → reject → forced turn → empty → in-turn retry → `invalid_tool_protocol` abandonment chain has no `recover=False` test. Suggest asserting `success` False, final status, cleared `pending_tool_calls`, and exactly two LLM calls.

**F13 — `recover_full_tool_set` branch after a resume rejection untested [minor]**
Test gap: after a resume rejection, a forced turn emitting a non-`final_answer` tool resets `force_final_answer_next` True→False (react.py:635-637) with no test at all — live-path tests always leave the flag False. Suggest a test that resumes with a forced-turn response containing a work tool.

**F14 — empty-answer + unavailable-tool recovery branch untested [minor]**
Test gap: the `react.py:919-927` append branch (`empty_final_answer=True` with `recovery_reason="unavailable_tool_call"`) has no end-to-end test. Suggest: forced-turn response `[calculator, final_answer("")]`, asserting `phase=unavailable_tool_call_recovery`, the full tool set, and both instruction sentences in the retry prompt.

**F15 — final_answer schema never requires a non-empty answer [minor]**
`react.py:1657-1664`: the `answer` property description never states the answer must be non-empty; the only text the model sees before violating is the schema, with the retry instruction arriving after the fact. Adding "It must be non-empty." is a one-line change orthogonal to the `response_language` ordering decision.

### Blocking status

Blocking: no. The prior major finding (N1) is fixed, and everything remaining is minor or a test gap — recommended event: APPROVE.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -590,6 +593,20 @@ async def _run_tool_calling_loop(
                     normalized,
                     force_final_answer=force_final_answer_now,
                 )
+                empty_final_answer = self._empty_final_answer_call(normalized)
+                if empty_final_answer is not None:
+                    logger.warning(
+                        "ReAct final_answer carried no answer text; discarding the "
+                        "response and retrying. iteration=%s arg_keys=%s",
+                        iteration,
+                        sorted(self._tool_call_args_dict(empty_final_answer)),
```
rogercloud: [minor] `sorted(self._tool_call_args_dict(...))` raises `TypeError` on mixed-type keys — reachable with non-JSON programmatic payloads or hand-crafted checkpoints, and since `run()` re-raises from its `except`, it would abort the run exactly at the intended recovery point. Real providers use string keys, so the blast radius is small, but `_args_hash` (react.py:3128-3135) already guards this pattern. Suggest `sorted(..., key=str)` here and at the same call in `_reject_empty_final_answer` (react.py:1983).

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1826,6 +1953,62 @@ async def _handle_control_tool(
...
+
+        Returning ``None`` keeps the run in the loop; ``force_final_answer_next``
+        makes the next turn re-request ``final_answer``, and if that turn is empty
+        too the normalization guard fails the run instead of looping.
+
+        Two deliberate differences from the fresh-turn path. The rest of the
+        batch is discarded, matching the fresh path's whole-response rejection:
+        otherwise a resumed ``[final_answer(""), work_tool]`` batch would execute
+        a side-effecting tool that the fresh path never runs. And the next turn is
+        forced to ``final_answer`` alone, where the fresh path restores the full
```
rogercloud: [minor] This docstring says forcing guarantees termination and no tool re-execution after a resume, but the escape hatch in `_requires_full_tool_set_recovery` (react.py:1068-1071) breaks the no-re-execution half: if the forced turn emits a non-`final_answer` tool, the full tool set is restored and `force_final_answer_next` is reset (react.py:635-637), so a previously cancelled work tool can be invoked again. Termination itself still holds — a second violation ends in `invalid_tool_protocol` — so this is a documentation-accuracy nit. Suggest wording that names the hatch, e.g. "unless the forced turn calls a non-final_answer tool, in which case the full tool set is restored and the guarantee lapses for that turn."

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1826,6 +1953,62 @@ async def _handle_control_tool(
...
+        self._record_tool_call(tool_call, status="failed", error=error)
+        context.add_tool_result(
+            tool_name="final_answer",
+            result={"error": error},
+            tool_call_id=tool_call.get("id"),
+        )
+        # Leave only the rejected call for the caller to pop, and close out the
+        # rest of the batch so no unexecuted call is left without a result.
+        remaining = self.pending_tool_calls[1:]
+        self.pending_tool_calls = self.pending_tool_calls[:1]
```
rogercloud: [minor] The docstring's "matching the fresh path's whole-response rejection" is only exact for one batch order. On a resumed `[work_tool, final_answer("")]` batch the work tool has already executed by the rejection point; `pending[1:]` only cancels calls after the rejected one, so a side effect the fresh path would never run can still replay from a stale/pre-fix checkpoint. Reachability is narrow (a live turn rejects the whole response first), but the claim as written is stronger than the behavior. Suggest a docstring caveat and, optionally, a reverse-order test — the current test only covers `[final_answer(""), work_tool]` (tests/core/agent/test_react.py:5416-5454).

src/xagent/core/tools/adapters/vibe/agent_tool.py
```diff
@@ -1523,6 +1523,17 @@ async def run_json_async(self, args: Mapping[str, Any]) -> Any:
...
 
+_CHILD_NO_ANSWER_MESSAGE = (
+    "The delegated agent never produced an answer, so there is no result from "
+    "it to use."
+)
+
+# Child statuses that mean "the pattern ended without ever producing an answer".
+# Their own ``error`` text names runtime internals, which the parent model
+# cannot act on, so they get the message above and the allowlisted
+# ``missing_delegated_output`` code instead of the raw diagnostic.
```
rogercloud: [minor] `_CHILD_NO_ANSWER_STATUSES` maps every `invalid_tool_protocol` child failure to "never produced an answer" + `missing_delegated_output`, including pre-existing producers whose answers were non-empty (provider protocol errors, mixed control calls, wrong tool in a forced turn). The child's own error text ("The model returned an invalid tool protocol response after one repair attempt.") is dropped — the parent's only clue that the child misbehaved rather than having nothing to say. The comment's claim that this text "names runtime internals" is inaccurate for this status; it is plain English about model behavior. Consider narrowing the mapping with a marker for genuinely empty answers, or folding the child's error text into the trace event.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -590,6 +593,20 @@ async def _run_tool_calling_loop(
                     normalized,
                     force_final_answer=force_final_answer_now,
                 )
+                empty_final_answer = self._empty_final_answer_call(normalized)
```
rogercloud, line 607: [minor] When the fresh path rejects a response, non-empty `assistant_content` preceding an empty `final_answer` is discarded with the whole response, and the retry rebuilds only from context — so produced preamble text is structurally invisible to the retry. If the model repeats the "preamble + empty final_answer" pattern, the run flips to a hard `invalid_tool_protocol` failure and the user never sees the preamble. Consider carrying the discarded preamble into the retry instruction/context, or explicitly accepting and documenting the tradeoff. An optional test variant with content plus an empty answer would pin the behavior.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1826,6 +1953,62 @@ async def _handle_control_tool(
...
+            "re-requesting a final answer. tool_call_id=%s arg_keys=%s",
+            tool_call.get("id"),
+            sorted(self._tool_call_args_dict(tool_call)),
+        )
+        error = (
+            "final_answer was called with an empty answer, so the user received "
+            "nothing. Call final_answer again with the complete user-facing "
+            "response in its answer field."
+        )
+        self._record_tool_call(tool_call, status="failed", error=error)
```
rogercloud: [minor] The context tool result `{"error": error}` is classified as success by `tool_result_succeeded` (src/xagent/core/agent/result.py:67-75), and `_record_tool_call(..., status="failed", error=error)` stores `result=None` while the sibling `_cancel_tool_calls` path stores a `{"success": False, "status": "cancelled", "error": reason}` dict. Latent today because `force_final_answer_next=True` short-circuits further handling, but a future consumer would see a failed call recorded as success with an inconsistent ledger shape. Suggest adding a classification key (e.g. `success: False` / `status: "error"`) and aligning the ledger record shape with the cancelled path.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in 5a14030 — `sorted(..., key=str)` at both sites. Agreed on the severity reasoning: `run()` re-raises from its `except`, so this aborted the run at exactly the point the guard exists to recover from, and `_args_hash` already established the guard-this-pattern precedent you point to. Added a unit test driving a `{"answer": "", 1: ..., None: ...}` payload through the rejection.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in 5a14030. You are right that termination and no-re-execution are different claims and only the first survives. Reworded to state that forcing bounds the run — a second empty answer ends it as `invalid_tool_protocol` — and then names the hatch explicitly: a forced turn calling a non-`final_answer` tool restores the full set and clears the flag, so a discarded work tool can be re-invoked on that turn.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in 5a14030. The docstring now says the discard matches the fresh path's whole-response rejection only for the order that matters, and states plainly that in `[work_tool, final_answer("")]` the work tool has already executed and nothing here undoes it.

Reverse-order test added (`test_react_resumed_reverse_batch_order_cannot_undo_an_executed_tool`), asserting the work tool did run and the ledger shows `completed` alongside the rejected `failed` — pinning the limitation rather than implying symmetry.

While there I also fixed the positional assumption you raised separately: siblings are now selected by identity rather than `[1:]`, so the discard no longer depends on the rejected call being the head.

src/xagent/core/tools/adapters/vibe/agent_tool.py
codeacme17: Fixed in 5a14030 — and you are right that I over-corrected from N5. Collapsing all four producers lost the diagnostic for three of them, and my comment claiming the error text "names runtime internals" was inaccurate for this status.

`_invalid_tool_protocol_result` now marks the empty-answer case explicitly (`empty_final_answer` in the result metadata, with a distinct error string), and the classifier keys on that marker instead of the bare status. Provider protocol errors, mixed control calls, and a non-`final_answer` tool on a forced turn now keep their own error text and get no `failure_code`.

Parametrized test covers both directions: marked → `missing_delegated_output` with the child error suppressed; unmarked → error text preserved verbatim and no `failure_code`.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Documented rather than changed, in 71d28af — I went with the "explicitly accept and document" option you offered, and I want to be clear it is a choice and not an oversight.

Reasoning: the preamble arrived attached to a protocol violation, so it is not a vetted user-facing answer, and feeding the model its own discarded prose back invites it to treat that text as already committed. The retry instruction already asks for the complete answer.

The cost you describe is real and is now stated at the rejection point: a model that keeps emitting "preamble + empty answer" fails the run without the user seeing the preamble. Added the content-plus-empty-answer test variant you suggested, so the behavior is pinned and a future change to it is a visible decision.

src/xagent/core/agent/pattern/react/react.py
codeacme17: Fixed in 5a14030. Confirmed both halves — `{"error": ...}` carries none of the keys `tool_result_succeeded` reads (`success: False` / `is_error` / `status == "error"`), so the failure read back as a success, and the ledger stored `result=None` against the cancelled path's dict.

Both now use one `{"success": False, "status": "error", "error": error}` shape, passed to `_record_tool_call` as well so the ledger is uniform with `_cancel_tool_calls`. Test asserts `tool_result_succeeded(recorded.result) is False`, which fails without the change.

codeacme17:
Thanks for the approval, and for the follow-up pass. Two commits: 5a14030 (F1, F2, F3, F7, F8, F10, F11, F12, F15) and 71d28af (F9). Per-finding replies are on the inline threads; below is what is not inline.

## The three that were real defects

**F8** was the sharpest catch. `run()` re-raises from its `except`, so an unsortable args payload aborted the run at precisely the point the guard exists to recover from — the guard crashing on the input it was built for. `sorted(..., key=str)` at both sites, matching the precedent `_args_hash` already set.

**F11**: confirmed `{"error": ...}` carries none of the keys `tool_result_succeeded` reads, so the rejection read back as a success, and the ledger shape diverged from the cancelled path. One `{"success": False, "status": "error", "error": ...}` shape now, in both the context result and the ledger record.

**F3**: I over-corrected from N5. Collapsing all four `invalid_tool_protocol` producers discarded the diagnostic for three of them, and my comment about "runtime internals" was inaccurate for this status. ReAct now marks the empty-answer case explicitly and only that marker maps to the no-answer classification; the other producers keep their error text.

## F1, F2 — my claims, not the code

Both docstrings asserted more than the code delivers, which is the failure mode most likely to mislead a later reader. F1 conflated termination with no-re-execution and only the first holds; the hatch is now named. F2 claimed an equivalence that is exact in only one batch order; the reverse order is now stated and tested.

While fixing F2 I also addressed **F10** by selecting siblings by identity rather than `[1:]`, which removes the positional assumption rather than asserting it.

## F9 — documented, as a choice

I took the "explicitly accept and document" branch you offered. The preamble arrives attached to a protocol violation, so it is not a vetted answer, and replaying the model's own discarded prose invites it to treat that text as committed. The cost you name is real and is now stated at the rejection point, with the content-plus-empty-answer test variant added so any future change to it is a visible decision rather than a silent drift.

## Test gaps

- **F7** — resume now driven through the real `get_state`/`load_state` round-trip, and the post-answer state is round-tripped again so the flag's serializability is proven rather than assumed.
- **F12** — full `resume → reject → forced → empty → retry → abandon` chain, asserting `success` False, `invalid_tool_protocol`, the marker, cleared `pending_tool_calls`, and exactly two LLM calls.
- **F2** — reverse batch order.
- **F3** — both mapping directions.
- **F8, F11** — units that fail without their fixes.

**F13 and F14 I have not added.** Both are genuine branch-coverage gaps and I am not disputing them; they are the two remaining items on this review. I would rather land the defect fixes now and add those two separately than keep growing this PR — say the word if you would prefer them here before merge.

## F15

Added: the `answer` description now states it must be non-empty, so the constraint reaches the model before the violation instead of only in the retry. Orthogonal to the `response_language` ordering decision, as you noted.

## Verification

Full `tests/core` diffed against a baseline captured by reverting the branch and re-running the identical command: zero regressions. All pre-commit hooks pass. CI running.

PR #1326, opened 2026-08-13
feat(steps): surface planning phase from dag execution events

AlexLiu190625:
## What

`/v1` steps (and the shared public step projection) now translate the existing `dag_execution` planning/replanning/executing phase transitions into a `planning` thinking step: pollers see "planning" within one event of plan generation starting, instead of an empty list for the whole planning window.

No new events and no new persistence — the translation reads rows the runtime already records. Each entry into planning (first plan or replan) yields its own step; a plan that never reaches the executing phase leaves its step `running`, matching how interrupted steps already behave. The step `status` OpenAPI description now states that terminal tasks may contain running steps and that terminal state is judged from the task's own status.

## Why

Task startup spends several seconds inside plan generation, during which the steps endpoint returned an empty list — pollers had no signal that the server was doing anything. The planning phase was already recorded in the trace stream; it was just never projected.

## Verification

- Pairing pins: planning/replanning opens a step, executing completes it, each planning entry yields an independent step; a failure sequence (no executing event) leaves the step running without disturbing other steps.
- Resume sequence (`planning, planning, executing`) pins first-running/second-completed.
- Data-shape pin: the translated step's `data` keys are exactly `{"phase"}`.
- Coexistence fixture: legacy `dag_plan_*` events and `dag_execution` translation fold independently with disjoint step ids.
- Full `tests/web/api/v1` suite green on top of the projector extraction.

### DIFF
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -39,6 +39,7 @@ Incremental projection:
     :class:`PublicStepProjector`, not in this function. The projector
     holds all the folding state a fold needs -- the pending
     (start-seen, end-not-yet-seen) table, the ``dag_plan_*`` replan
+    counter, and the independent ``dag_execution`` planning-phase
     counter -- so a caller can feed it one event at a time and read
     back the steps that changed. ``map_trace_events_to_public_steps``
     stays pure by constructing one fresh, throwaway projector per
@@ -58,6 +59,13 @@ Pairing rule:
       - ``dag_step_*`` events pair on ``step_id``.
       - ``dag_plan_*`` events pair on ``task_id`` (single planning
         phase per task; no per-plan identifier available).
+      - ``dag_execution`` events pair on ``data['phase']``: a
+        ``planning``/``replanning`` value opens a planning-phase
+        thinking step, an ``executing`` value closes the currently
+        open one. This is a second, independent source of the same
+        public phase as ``dag_plan_*`` above -- it keeps its own
+        counter and open key so the two families never collide, even
+        when both appear in the same event stream.
       - ``skill_select_*`` events pair on ``task_id`` (single skill
         selection phase per task).
 
@@ -158,6 +166,14 @@ class PublicStepProjector:
         # time); nesting would require a stack, which DAG doesn't emit.
         self._plan_counter = 0
         self._open_plan_key: Optional[str] = None
+        # ``dag_execution`` (phase planning/replanning/executing) is a
+        # second source of the same public planning phase, translated
+        # further down in ``feed``. It gets its own counter and open
+        # key so it never shares pairing state with ``dag_plan_*``
+        # above -- both families can appear in the same stream without
+        # cross-talk or id collisions.
+        self._dag_execution_counter = 0
+        self._open_dag_execution_key: Optional[str] = None
 
     @classmethod
     def from_history(cls, events: List[Any]) -> "PublicStepProjector":
@@ -272,8 +288,8 @@ class PublicStepProjector:
                     status="completed",
                 )
                 return [finalized] if finalized is not None else []
-            # other actions (e.g. dag_execution UPDATE) for these
-            # categories are not exposed.
+            # Events in these families that are neither a start nor an
+            # end carry no step transition.
             return []
 
         # ===== tool_call / agent_delegation: paired start/end + failure =====
@@ -381,6 +397,42 @@ class PublicStepProjector:
                 extra_data_fn=lambda ev: {"result": _data_get(ev, "result")},
             )
             return [finalized] if finalized is not None else []
+
+        # ===== dag_execution: translate an existing signal into the
+        # same public planning phase ``dag_plan_*`` produces =====
+        if event_type == "dag_execution":
+            phase = _data_get(event, "phase")
+            if not isinstance(phase, str):
+                # Missing/malformed phase: nothing to translate.
+                return []
+            if phase in ("planning", "replanning"):
+                self._dag_execution_counter += 1
+                task_ref = (
+                    _safe_get(event, "task_id")
+                    or _safe_get(event, "event_id")
+                    or "anon"
+                )
+                key = f"planning:{task_ref}:{self._dag_execution_counter}"
+                self._open_dag_execution_key = key
+                step = _build_thinking_start(event, phase="planning", key=key)
+                self._pending[("thinking", key)] = step
+                return [step]
+            if phase == "executing":
+                if self._open_dag_execution_key is None:
+                    # No open planning step to close: same policy as an
+                    # orphan end elsewhere in this module -- drop.
+                    return []
+                finalized = _finalize_pending(
+                    self._pending,
+                    self._finished,
+                    ("thinking", self._open_dag_execution_key),
+                    end_event=event,
+                    status="completed",
+                )
+                self._open_dag_execution_key = None
+                return [finalized] if finalized is not None else []
+            # Other phases (e.g. completion_assessment): not exposed.
+            return []
 
         # Everything else (llm_call_*, memory_*, dag_execute_*,
         # react_task_*, react_step_*, visualization_update,
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -84,9 +84,11 @@ Pairing rule:
     carry the same public step id (id and pairing key are derived from
     the same (type, key) pair), so a consumer folding ``feed`` results
     by id converges on the second step rather than being left with a
-    stranded ``running`` one. ``dag_plan_*`` is the one exception: it
-    has no natural key at all, so a counter gives every plan a
-    distinct one.
+    stranded ``running`` one. ``dag_plan_*`` and the planning-phase
+    steps ``dag_execution`` produces are the exception: neither has a
+    natural key at all, so each family keeps its own counter (and its
+    own ``plan:`` / ``planning:`` key prefix) to generate a distinct
+    one per occurrence.
 """
 
 from __future__ import annotations
@@ -245,6 +247,9 @@ class PublicStepProjector:
             # Special-cased because plan events have no per-plan id;
             # we generate one from a counter and remember the open
             # key so the next dag_plan_end pairs with the latest start.
+            # The dag_execution branch below keeps its own independent
+            # counter and "planning:" key prefix; keep pairing
+            # semantics changes in lockstep between the two branches.
             if event_type.endswith("_start"):
                 self._plan_counter += 1
                 task_ref = (
--- a/src/xagent/web/schemas/v1.py
+++ b/src/xagent/web/schemas/v1.py
@@ -498,9 +498,10 @@ class TaskInfoResponse(BaseModel):
 # internal->public mapping table.
 PublicStepType = Literal["thinking", "tool_call", "agent_delegation", "message"]
 
-# ``running`` means a start event was seen with no matching end (the
-# task is still mid-step at the time of the GET). ``completed`` /
-# ``failed`` reflect the end event's success flag.
+# ``running`` means a start event was seen with no matching end; a
+# terminal task can still contain running steps (interrupted/cancelled
+# steps never get an end event). ``completed`` / ``failed`` reflect
+# the end event's success flag.
 PublicStepStatus = Literal["running", "completed", "failed"]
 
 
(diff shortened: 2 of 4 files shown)


src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -382,6 +398,42 @@ def feed(self, event: Any) -> List[Dict[str, Any]]:
             )
             return [finalized] if finalized is not None else []
 
+        # ===== dag_execution: translate an existing signal into the
+        # same public planning phase ``dag_plan_*`` produces =====
+        if event_type == "dag_execution":
```
rogercloud, line 408: minor — this `dag_execution` branch near-verbatim duplicates the `dag_plan_*` pairing machine above (counter, `task_id`/`event_id`/`"anon"` fallback chain, key synthesis, open/close via `_finalize_pending`, ~18 lines). The disjoint `planning:`/`plan:` prefixes justify separate state but not separate code: if pairing semantics evolve, both branches must be updated in lockstep. Suggestion: extract a small shared pair-open/pair-close helper, or add an explicit cross-reference comment on both branches.

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -382,6 +398,42 @@ def feed(self, event: Any) -> List[Dict[str, Any]]:
...
+            phase = _data_get(event, "phase")
+            if not isinstance(phase, str):
```
rogercloud, line 410: minor — this `isinstance` guard is behaviorally redundant: a non-string `phase` fails both membership tests below and falls through to the final `return []` identically, and no test discriminates the guard. Fine to keep as intent documentation (malformed rows are deliberately dropped), but it can also be removed without changing behavior.

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -382,6 +398,42 @@ def feed(self, event: Any) -> List[Dict[str, Any]]:
...
+                # Missing/malformed phase: nothing to translate.
+                return []
+            if phase in ("planning", "replanning"):
```
rogercloud, line 413: minor — the phase literals `"planning"`, `"replanning"` (here) and `"executing"` (line 420) are hardcoded in two comparisons that must stay in sync. Suggestion: hoist to a module-level constant, e.g. `_DAG_EXECUTION_PLANNING_PHASES = ("planning", "replanning")`.

src/xagent/web/schemas/v1.py
```diff
@@ -542,9 +542,12 @@ class PublicStep(BaseModel):
...
         description=(
             "running while the corresponding end event has not yet "
-            "fired (i.e. the SDK polled mid-step), completed on a "
-            "normal end event, failed when the end event carries "
-            "success=false."
+            "fired. A task in a terminal state can still contain "
+            "running steps: interrupted or cancelled steps never get "
+            "an end event, so terminal state is judged from the task's "
+            "own status, not from the absence of running steps. "
+            "completed on a normal end event, failed when the end "
```
rogercloud: minor — "completed on a normal end event, failed when the end event carries success=false." lost its subject when the preceding sentence gained a full stop; the fragments now dangle. Suggestion: "It is completed on a normal end event, failed when the end event carries success=false."

tests/web/api/v1/test_steps_mapping.py
```diff
@@ -1083,3 +1158,254 @@ def test_feed_return_value_reflects_step_state_at_call_time():
 
     # Empty/falsy event_type: nothing identifiable to fold.
     assert PublicStepProjector().feed(_ev("", step_id="s1")) == []
+
+
+# ===== dag_execution: planning-phase visibility from an existing signal =====
+
+
+def test_dag_execution_planning_pair_becomes_thinking_planning():
```
rogercloud, line 1166: minor — test gaps for the new branch: (a) the happy-path test pins type/phase/status/id but not `completed_at` — the executing row's timestamp is available and unasserted; (b) all cases use string `task_id`s, so a numeric task id (rendered by the f-string key as `thinking:planning:42:1`) is never exercised; (c) the helpers always supply a timestamp, leaving the production missing-timestamp fallback dead; (d) the `task_id` → `event_id` → `"anon"` fallback chain is unreachable through the helpers; (e) no endpoint-level test that `/v1/steps` surfaces a `dag_execution`-derived planning step. Suggestion: add a numeric-`task_id` case asserting `completed_at == executing timestamp`, and optionally one endpoint-level plumbing test.

rogercloud (approved):
## Summary

The `/v1/steps` endpoint returned an empty list while a task was in the plan-generation phase, because `dag_execution` trace rows were invisible to the public step projector. This PR fixes that by translating existing `dag_execution` planning/replanning/executing phase rows into a `thinking` step with `phase: "planning"` — a pure projection-layer change that adds no new events and no persistence. A per-projector counter plus an open-key tracks planning→executing pairs independently of the legacy `dag_plan_*` family, so replans and resume sequences fold correctly and the two families never collide.

## Approach

The design is sound. Adding an independent counter and open-key inside `PublicStepProjector.feed` (key prefix `planning:` vs. the legacy `plan:`) is the right layer: it is a pure projection over rows that already exist in storage, is robust across replan/resume sequences (each planning entry opens, each executing closes the open entry), and leaves the behavior of every other event type untouched.

## Findings (all minor)

**Inline comments** (attached at the cited lines):

1. **minor — duplicated pairing machine across the two families** (`src/xagent/web/api/v1/_step_mapping.py:403-435`): the new `dag_execution` branch near-verbatim repeats the `dag_plan_*` branch's counter, `task_id`/`event_id`/`"anon"` fallback chain, key synthesis, and open/close pairing (~18 lines). Disjoint key prefixes justify separate state, but not separate code: if pairing semantics evolve, both branches must change in lockstep. Suggestion: extract a small shared pair-open/pair-close helper, or add an explicit cross-reference comment on both branches.
2. **minor — hardcoded phase literals** (`_step_mapping.py:408,420`): `"planning"`, `"replanning"`, `"executing"` appear inline in two comparisons that must stay in sync. The unknown-phase drop behavior is pinned by `test_dag_execution_other_phases_are_dropped`, so this is low-risk. Suggestion: hoist to a module-level constant.
3. **minor — redundant `isinstance` guard** (`_step_mapping.py:405-407`): a non-string `phase` fails both membership tests below and falls through to the final `return []` identically, so the guard changes no behavior and no test discriminates it. Suggestion: keep as intent documentation or remove it.
4. **minor — status description grammar splice** (`src/xagent/web/schemas/v1.py:549-550`): "completed on a normal end event, failed when the end event carries success=false." lost its subject when the preceding sentence gained a full stop. Suggestion: "It is completed on a normal end event, failed when the end event carries success=false."
5. **minor — test gaps for the new branch** (`tests/web/api/v1/test_steps_mapping.py:1166`): the happy-path test pins type/phase/status/id but not `completed_at` (the executing row's timestamp is available but unasserted); all cases use string `task_id`s, so a numeric task id (rendered by the f-string key as `thinking:planning:42:1`) is never exercised; the `_dag_exec_ev`/`_ev` helpers always supply a timestamp, leaving the production missing-timestamp fallback dead; and the `task_id` → `event_id` → `"anon"` fallback chain is unreachable through the helpers. Suggestion: add a numeric-`task_id` case with a `completed_at == executing timestamp` assertion, plus an optional endpoint-level test that `/v1/steps` surfaces a `dag_execution`-derived planning step.

**Body-only findings** (referenced lines sit outside the diff hunks):

6. **minor — planning step left running when a plan fails** (`_step_mapping.py:416-433`): if planning fails and no `executing` row follows, a terminal failed task still shows a `running` planning step. This is documented deliberate intent (PR description, the `status` field description, and `test_dag_execution_planning_without_executing_stays_running`), so it is fair as a minor, not a defect.
7. **minor — cross-family duplicate planning steps undocumented** (`_step_mapping.py:401`, pinned by `test_dag_plan_and_dag_execution_coexist_without_id_collision`): for version-spanning task histories containing both `dag_plan_*` and `dag_execution` rows, a user sees two planning steps for what was conceptually one plan. Behavior is intentional and pinned, but the user-visible consequence is not documented. Suggestion: one docstring sentence noting the duplicate can occur across the two families.
8. **minor — stale "the one exception" module docstring** (`_step_mapping.py:87-91`): the pairing-rule section still says "`dag_plan_*` is the one exception: it has no natural key at all"; the new branch is a second counter-based family. Suggestion: update the sentence to cover both families.
9. **minor — stale comment above `PublicStepStatus`** (`src/xagent/web/schemas/v1.py:501-503`): "running means a start event was seen with no matching end (the task is still mid-step at the time of the GET)" contradicts the new Field description (lines 544-548) stating a terminal task can still contain running steps. Suggestion: align the comment with the Field text.

Out of scope / dropped: cache staleness at deploy (pre-existing and systemic) and the stale `trace.py:1334` docstring (pre-existing; drive-by only).

Lean already. The only echo is the ~20-line pairing core shared between the `dag_plan_*` and `dag_execution` families, but the triggers differ and the coexist test pins their independence, so it stays within scope of lean.

Blocking: no. No critical or major findings were confirmed; all findings are minor and non-blocking.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Taking the cross-reference option (0add832c6). Extracting a shared helper would have to rewrite the legacy branch's own counter and open-key state; that independent state is exactly what `test_dag_plan_and_dag_execution_coexist_without_id_collision` pins, and rewriting it buys no behavior change. The legacy branch now carries the missing reverse pointer, and the module docstring covers both counter-based families (also addresses your "one exception" note).

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Keeping it. It states the malformed-input drop policy at the top of the branch rather than leaving it implicit in the fall-through `return []` two arms below; explicit failure paths are the house style here. Agreed it is behaviorally inert.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Declining the hoist: the two comparisons are mutually exclusive arms of one dispatch, not two copies of one list, so there is no sync obligation between them, and the tuple would have exactly one use site. The real sync obligation is with the emitter in `core/agent/pattern/dag/dag.py`, which a module-local constant would not address; a shared enum across the core/web boundary is out of scope for this change.

src/xagent/web/schemas/v1.py
AlexLiu190625: Fixed in 0add832c6, with a different shape than suggested: instead of adding a subject to the dangling fragment, the terminal-state note moved to the end so the original three-status parallel structure (running/completed/failed) reads as one sentence again. The `PublicStepStatus` comment above was aligned in the same commit (it still described running as "task mid-step", contradicting the field text).

tests/web/api/v1/test_steps_mapping.py
AlexLiu190625: Split outcome (0add832c6): (a)+(b) taken — new test drives an integer task id (the production column type; the whole file previously used only strings) through a planning/executing pair and asserts the id literal `thinking:planning:42:1` plus `completed_at` equal to the executing row's timestamp exactly. (c)+(d) declined: both target pre-existing shared code unreachable in production — `TraceEvent.timestamp` and `task_id` are `nullable=False`, so the missing-timestamp fallback and the second/third links of the key fallback chain cannot be hit by a real row; pinning them adds maintenance for dead defensive code. (e) declined: endpoint plumbing is already covered over real rows by `test_get_steps_returns_mapped_steps_in_order`, and the new branch emits `{"phase": "planning"}` — the same shape as the already-covered thinking step, through the same `PublicStep(**step)` construction.

AlexLiu190625:
Inline findings addressed at 0add832c6 (status-description restructure, PublicStepStatus comment alignment, both counter-based families documented with a cross-reference, integer-task-id test with an exact completed_at pin); per-thread rationale on each thread.

On the two body-only notes: the cross-family duplicate planning step is not reachable through this endpoint — `_resolve_task_or_404` filters `Task.source == "sdk"`, which is written only by the `/v1` endpoints (first shipped 2026-05-18, `7ed3004ac`), while the only runtime that ever emitted `dag_plan_*` was removed 2026-05-15 (`f58c9c927`); no task readable here can carry those rows, so documenting the duplicate would describe an outcome that cannot occur. The planning-step-left-running case is the documented deliberate behavior; a follow-up change addressing it is queued.

PR #1334, opened 2026-08-13
fix(image): infer the edit ability where image models default their abilities

bluefish-08:
Sub-issue of #1289. Follow-up to #1294.

**This PR changed direction after review — it no longer contains a migration.**

## Background

#1294 stopped registering `edit_image` when no configured image model declares
the `edit` ability. That is correct — before it, `_get_edit_model` already
required `has_ability("edit")`, so those deployments got a guaranteed failure on
every call rather than a working tool.

What it exposed is that the ability is usually missing from the *data*, not from
the model. `ModelCreate.abilities` defaults to `None` and
`web/api/model.py:480` stores it as-is for image models — unlike the video branch
next to it, which defaults to `["generate"]`. So an image model's `abilities` is
normally NULL in the database, and `["generate"]` is supplied at read time by the
fallbacks in `model_service.py:567,575` and `core/model/image/adapter.py:32`.

A `qwen-image-edit` or `gemini-3-pro-image-*` deployment that was never
explicitly configured therefore reports generate-only, and now silently loses a
tool it could actually serve.

## Changes

1. **`default_image_abilities(provider, model_name)`**
   (`core/model/image/base.py`) — decides what an image model declares when its
   row declares nothing. It is reached only through the `or` short-circuit at the
   fallback sites, so a declared abilities list stays authoritative and an
   operator's deliberate generate-only choice is never overridden.

   `openai` and `xinference` keep the `["generate", "edit"]` their web call sites
   already applied — not a claim that they edit across their lineup (xinference
   defaults to stable-diffusion-2-1 and raises unless the backend exposes
   image_to_image), but the existing default, kept so this change stays about the
   two paths agreeing rather than about widening access. `dashscope` and `gemini`
   ship both edit-capable and generate-only models under one provider name, so for
   them the name decides, with markers scoped per provider ("3-pro" is Gemini
   vocabulary and does not answer for dashscope). `GeminiImageModel` now calls the
   helper instead of redeclaring the rule. Everything else stays generate-only.

2. **All five fallback sites now call it** (`model_service.py` ×4,
   `image/adapter.py` ×1) instead of hardcoding a literal. That also settles a
   pre-existing split: an unconfigured openai/xinference row read through
   `get_image_model_instance` used to come back generate-only while
   `model_service` called the same row editable, so a row picked as the default
   edit model was rejected by `_get_edit_model` and silently swapped for another.

3. **A log line when gating withholds a tool** (`vibe/image_tool.py`
   `get_tools`) — the withholding was completely silent. The operator-visible
   symptom is "image editing stopped working" with nothing in the log pointing at
   `abilities`, which is a worse diagnostic than the runtime error it replaced.

## Why not a migration

The first version of this PR was an Alembic backfill. Review showed the shape was
wrong, in two ways that a migration cannot fix:

- It is point-in-time. Rows registered afterwards default to generate-only again,
  and a fresh install never runs it at all — `db/migration.py:364-370` stamps a
  brand-new database straight to head.
- It cannot tell NULL from a deliberate `["generate"]`, so it overrides an
  operator's explicit choice — the same objection that made runtime inference
  look unattractive in the first place.

Changing the fallback covers existing NULL rows, rows created later, and fresh
installs, while a declared list short-circuits it entirely. No data is rewritten,
and reverting is a one-line change rather than a hand-repair of rows.

## Known ceiling

Name-based inference reads intent from the model name, not from the endpoint. A
dashscope row named `*-edit` whose `base_url` points at a text2image endpoint now
registers `edit_image` and fails at call time with a 4xx, where before it was
generate-only and the tool was withheld. Judged the right trade — a model named
`*-edit` is configured for editing — but it is a real narrowing of what #1294's
gating catches, and an operator can pin it back by declaring `["generate"]`
explicitly.

## Verification

- New `tests/core/model/image/test_default_abilities.py` (pure function): markers
  matched per provider, non-matching names generate-only, whole-lineup providers
  answering the same for any name, and provider casing/whitespace.
- New `tests/core/model/image/test_default_abilities_wiring.py`: each fallback
  site actually reaches the helper, a declared list still wins, and both
  model-building paths agree on the same unconfigured row. Verified by mutation —
  reverting either wiring to a literal turns these red while the pure-function
  tests stay green.
- `tests/web/test_model_api.py`, `tests/core/model/`, `tests/core/tools/test_image_tool.py`,
  and `tests/core/tools/core/test_image_tool_core.py` pass.
- `ruff check`, `ruff format`, `mypy`, and the full pre-commit hook set pass.


bluefish-08:
Fixed in ffcefc4b — the review is right, and it would have failed in production: `abilities` is a `json` column and every test here runs on SQLite, whose JSON is text, so nothing in the suite could catch it.

Confirmed the difference by compiling the statement under the psycopg2 dialect:

```
before: UPDATE models SET abilities = %(abilities)s WHERE id = %(id)s
after:  UPDATE models SET abilities = %(abilities)s::JSON WHERE id = %(id)s
```

The bare parameter is what raises `column "abilities" is of type json but expression is of type text`.

The statement is now a module-level `UPDATE_ABILITIES` constant so a new test can compile it under the psycopg2 dialect and assert the cast survives. Verified by mutation: dropping the `bindparams` call turns that test red.

rogercloud (changes requested):
## Summary

PR #1334 repairs the provisioning gap that PR #1294's edit_image capability gating exposed: existing deployments carry `models` rows with abilities=["generate"] (or NULL) for image models that actually support editing, so edit_image is silently withheld. The change adds a one-shot Alembic migration (`20260813_backfill_image_edit_ability`) that appends "edit" to every `category='image'` row whose `model_name` contains an edit marker ("edit", "3-pro", "gpt-image"), leaving unmarked names generate-only, skipping rows that already declare edit, and treating NULL abilities as ["generate"]; a companion INFO log line in `vibe/image_tool.py` makes tool withholding diagnosable. Coverage is added via a new test module (marker matching, category filter, NULL abilities, missing-models-table database, and a compile check that the UPDATE binds a Postgres `json` parameter), and downgrade is a deliberate no-op because an appended "edit" is indistinguishable from one an operator configured by hand.

## Approach

Design verdict: **acceptable-with-reservations**. The migration placement is right: production rows carry explicit abilities=["generate"] (issue #1289), so no runtime fallback change (`adapter.py:32`, constructors, `model_service`) could repair explicit data without also overriding deliberate operator configuration — only a data migration can. The `has_table` guard, idempotent upgrade, conservative name heuristic, no-op downgrade, and the json-typed bindparam fix for the Postgres `json` column are sound, and the test module covers the contract well.

The reservations are design-level and are carried as the two MAJOR findings below:

- **F1 (MAJOR, design-level): the backfill is point-in-time.** The fallback sites that produced the generate-only rows remain live, so any image model registered after the migration is again generate-only and edit_image is again withheld; fresh installs never execute the migration at all.
- **F3 (MAJOR, design-level): the backfill overwrites explicit operator configuration.** Marker-named rows with deliberate abilities=["generate"] get "edit" appended anyway, and the provider-agnostic markers can resurrect the doomed edit_image for self-hosted xinference models whose remote lacks image editing, with no audit trail to detect a misfire.

## Severity findings

**F1 — MAJOR** (design-level; `src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:37`)

The migration repairs rows that exist at upgrade time, but every fallback site that produced generate-only rows is untouched: `web/services/model_service.py:567,575` still default dashscope/gemini to ["generate"], `core/model/image/adapter.py:32` defaults all providers to ["generate"], and the registration path (`web/api/model.py:480`) stores NULL abilities for image models — the only category whose validator permits None. A `qwen-image-edit` or `gemini-3-pro` row registered after this migration is again reported generate-only, edit_image is again withheld, and the only signal is one INFO log line per tool build. Fresh installs never run the migration (`db/migration.py` stamps head; `models` is create_all-owned, so the `has_table` guard no-ops). No issue or PR tracks the runtime-fallback fix, so the exact production story from #1289 recurs silently for any model added later.

Suggestion: fold name inference into the single read/registration path — e.g. a shared `infer_image_abilities(model_name, declared)` helper used by `model_service.get_image_models` (or applied at the `ModelCreate`/update boundary in `web/api/model.py`, where category and name are known), with the migration consuming the same definition — or, keeping runtime behavior deliberately conservative, at minimum log a warning from the fallback sites and track the recurrence as a follow-up issue.

**F3 — MAJOR** (design-level; `src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:53`)

The marker check appends "edit" to every `category='image'` row whose name carries a marker even when abilities=["generate"] was a deliberate operator choice (e.g. cost control on an edit-named model, or a shared model, which cannot change abilities via the API at all — `web/api/model.py:1877-1883`); the runtime treats explicit abilities as authoritative (`gemini.py:173-175`, `dashscope.py:31`, `xinference.py:46-49`). The provider-agnostic marker tuple can false-positive self-hosted xinference models (e.g. an operator-named "*-edit") whose remote lacks image_to_image/inpainting — `xinference.py:243-254` then raises "Image editing is not supported by this Xinference model" on every call, resurrecting the doomed edit_image that #1294's gating was built to eliminate. The SELECT does not fetch `model_provider`, so markers cannot be scoped per provider; with a no-op downgrade and no row-level logging, a wrong flip is invisible and hand-revert-only.

Suggestion: fetch `model_provider` in the SELECT and scope the marker check per provider (or restrict it to providers with known edit support), and log every changed row so a misfire can be audited and reverted.

**F2 — MINOR** (`src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:21`)

The comment claims the markers are "the same markers `core/model/image/gemini.py` already infers from" and that the migration "writes down what the runtime would have concluded" — both are inaccurate. `gemini.py` infers edit capability only from ("3-pro", "edit"); "gpt-image" traces to `openai.py:20` (default model name) and `openai.py:32` (constructor default). And because serving paths always pass an explicit abilities list (`adapter.py:32`, `model_service.py`), gemini's name inference never runs at runtime — the migration introduces new semantics. Fix: correct the comment wording. Do not import a shared runtime constant into the migration: `versions/` migrations never import runtime code (frozen-migration convention).

**F4 — MINOR** (`src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:51`)

The unguarded `json.loads` can abort the whole upgrade on a malformed string, and a JSON string scalar silently char-splits via `list()` on the next line — the UPDATE then persists `['g','e',...] + ['edit']`, silently dropping "generate" from the model at runtime. Sibling backfill `20260721_backfill_strip_agent_category.py:54` guards with `isinstance(categories, list)`, and `web/services/model_service.py:1106-1116` wraps `json.loads` in try/except. The trigger requires legacy/manual rows (the API is pydantic-validated), hence minor.

Suggestion: wrap the load and guard the fallback, e.g.:

```python
if isinstance(abilities, str):
    try:
        abilities = json.loads(abilities)
    except ValueError:
        continue
if not isinstance(abilities, list):
    continue
abilities = list(abilities or ["generate"])
```

**F5 — MINOR** (`src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:41`)

In offline `--sql` mode, `sa.inspect(conn).has_table("models")` raises `NoInspectionAvailable` on Alembic's MockConnection (empirically reproduced), so `alembic upgrade --sql` crashes for any database stamped in the guarded 20260721-20260812 stretch. The repo convention since 20260713 is to guard reflection with `if not op.get_context().as_sql:`; `tests/alembic/test_20260808_add_task_lease_attempt_id.py:118-127` asserts "offline branch must not reflect", and `test_task_interaction_protocol_version_parity.py:251-252` calls offline SQL "a hard requirement".

Suggestion: guard the reflection:

```python
if not op.get_context().as_sql and not sa.inspect(conn).has_table("models"):
    return
```

**F6 — MINOR** (`src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:55`)

`upgrade()` logs nothing about which rows it changed, and Alembic records only revision stamps — a wrong flip is invisible and hand-revert-only. Repo convention for data backfills is to log affected ids: `20260703_backfill_strip_other_tool_category.py` and `20260721_backfill_strip_agent_category.py` collect `affected_ids` and emit `logger.warning`.

Suggestion: emit an INFO or WARNING per modified row (model_id, model_name, before/after abilities) inside the upgrade loop.

**F7 — INFORMATIONAL** (`src/xagent/core/tools/adapters/vibe/image_tool.py:79`)

The withheld-tool INFO log fires on every `get_tools()` build — once per tools-list request (`web/api/tools.py:341`) and once per agent build (the chat path is agent-cache-backed, so not per chat request). A per-instance flag would not help (fresh instance per build); a module-level log-once guard or logging at the model-load layer would bound the noise. This is an intentional diagnostic per the PR's purpose, so it sits below minor.

**F8 — MINOR** (`tests/migrations/test_backfill_image_edit_ability.py:119`)

Test nits: the CWD-relative `Path("src/xagent/migrations/versions")` at line 119 deviates from the `__file__`-based convention in `tests/alembic/`, though it matches the rest of `tests/migrations/` and is safe in CI; and the migration's native-list (non-str) abilities branch never executes in tests — a raw `text()` SELECT returns str on SQLite, and production psycopg2 also returns str, so the str branch is the real path (the gap is low-value). An empty-list fixture row and a downgrade round-trip test were considered and dropped as branch-identical coverage and covered-by-precedent respectively; acceptable.

## Prior review finding (re-review)

The prior review's blocking finding — the UPDATE bound a raw JSON string into the Postgres `json` column ("expression is of type text") — is FIXED at HEAD ffcefc4: the migration now uses a module-level `UPDATE_ABILITIES` text with `sa.bindparam(type_=sa.JSON)`, and `test_update_casts_the_json_parameter` compiles the statement under the psycopg2 dialect and asserts the `::JSON` cast. The author confirmed the fix in https://github.com/xorbitsai/xagent/pull/1334#issuecomment-5279127285.

Blocking: yes

- `src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:37` — MAJOR — backfill is point-in-time; generate-only fallbacks stay live for rows registered after the migration, so the #1289 story recurs with no follow-up tracked. [new]
- `src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py:53` — MAJOR — marker append overrides deliberate generate-only configuration and can resurrect the doomed edit_image for xinference models whose remote lacks editing, with no audit trail to detect it. [new]

Recommended event: REQUEST_CHANGES

src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,63 @@
...
+    "UPDATE models SET abilities = :abilities WHERE id = :id"
+).bindparams(sa.bindparam("abilities", type_=sa.JSON))
+
+
+def _declares_edit(model_name: str) -> bool:
+    lowered = (model_name or "").lower()
+    return any(marker in lowered for marker in EDIT_CAPABLE_NAME_MARKERS)
+
+
+def upgrade() -> None:
```
rogercloud: [MAJOR] This backfill repairs only rows that exist at upgrade time; the fallback sites that produced the generate-only rows stay live — web/services/model_service.py:567,575 still default dashscope/gemini to ["generate"], core/model/image/adapter.py:32 defaults every provider to ["generate"], and web/api/model.py:480 stores NULL abilities for image models. A qwen-image-edit or gemini-3-pro row registered after this migration is again generate-only and edit_image is again withheld — and fresh installs never run this migration at all (models is create_all-owned and stamped to head). Please fold name-based ability inference into the read/registration path (a shared infer_image_abilities(model_name, declared) helper used by model_service.get_image_models or applied at ModelCreate/update, with the migration consuming the same definition), or at minimum log a warning from the fallback sites and track the recurrence in a follow-up issue.

src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,63 @@
...
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "20260813_backfill_image_edit_ability"
+down_revision: Union[str, None] = "20260812_seed_intercom_mcp_app"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+# The same markers core/model/image/gemini.py already infers from, so this only
```
rogercloud: [MINOR] This provenance claim is inaccurate: gemini.py infers edit capability only from ("3-pro", "edit"); "gpt-image" traces to openai.py's default model name and constructor default, not to gemini.py. Because serving paths always pass an explicit abilities list (adapter.py:32, model_service.py), gemini's name inference never actually runs — the migration introduces new semantics rather than "writing down what the runtime would have concluded". Please correct the comment wording; importing a shared runtime constant would violate the frozen-migration convention (versions/ migrations never import runtime code).

src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,63 @@
...
+    rows = conn.execute(
+        sa.text("SELECT id, model_name, abilities FROM models WHERE category = 'image'")
+    ).fetchall()
+
+    for row in rows:
+        abilities = row.abilities
+        if isinstance(abilities, str):
+            abilities = json.loads(abilities)
+        abilities = list(abilities or ["generate"])
+        if "edit" in abilities or not _declares_edit(row.model_name):
```
rogercloud: [MAJOR] This appends "edit" to every category='image' row whose name carries a marker even when abilities=["generate"] was a deliberate operator choice, and the runtime treats explicit abilities as authoritative (gemini.py:173-175, dashscope.py:31, xinference.py:46-49). The provider-agnostic marker tuple can also false-positive self-hosted xinference models (e.g. an operator-named "*-edit") whose remote has no image_to_image/inpainting — xinference.py:243-254 then raises on every call, resurrecting the doomed edit_image #1294's gating was built to eliminate. Consider fetching model_provider in the SELECT and scoping the marker check per provider, and log every changed row so a misfire can be audited and reverted.

src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,63 @@
...
+        return
+
+    rows = conn.execute(
+        sa.text("SELECT id, model_name, abilities FROM models WHERE category = 'image'")
+    ).fetchall()
+
+    for row in rows:
+        abilities = row.abilities
+        if isinstance(abilities, str):
+            abilities = json.loads(abilities)
```
rogercloud: [MINOR] An unguarded json.loads can abort the whole upgrade on a malformed string, and a JSON string scalar silently char-splits through list() on the next line, so the UPDATE persists ['g','e',...] + ['edit'] and drops "generate" from the model at runtime. Sibling backfills guard this: 20260721_backfill_strip_agent_category.py:54 checks isinstance(categories, list), and web/services/model_service.py:1106-1116 wraps json.loads in try/except. Wrap the load and guard the fallback with isinstance(abilities, list) so malformed or non-list values skip the row instead of corrupting it.

src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,63 @@
...
+def _declares_edit(model_name: str) -> bool:
+    lowered = (model_name or "").lower()
+    return any(marker in lowered for marker in EDIT_CAPABLE_NAME_MARKERS)
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    # models is created by metadata, not by a migration, so a fresh database can
+    # reach this revision before the table exists.
+    if not sa.inspect(conn).has_table("models"):
```
rogercloud: [MINOR] In offline --sql mode this reflection raises NoInspectionAvailable on Alembic's MockConnection (reproduced), so `alembic upgrade --sql` crashes for any database stamped in the guarded 20260721-20260812 stretch — and the repo treats offline SQL rendering as a hard requirement (tests/alembic/test_20260808_add_task_lease_attempt_id.py:118-127, test_task_interaction_protocol_version_parity.py:251-252). Guard the reflection with `if not op.get_context().as_sql:` before the has_table call, matching the convention used since 20260713.

src/xagent/migrations/versions/20260813_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,63 @@
...
+    ).fetchall()
+
+    for row in rows:
+        abilities = row.abilities
+        if isinstance(abilities, str):
+            abilities = json.loads(abilities)
+        abilities = list(abilities or ["generate"])
+        if "edit" in abilities or not _declares_edit(row.model_name):
+            continue
+        conn.execute(
```
rogercloud: [MINOR] upgrade() changes rows without logging which ones, and Alembic records only the revision stamp, so a wrong flip is invisible and hand-revert-only. Repo convention for data backfills is to log affected ids (20260703_backfill_strip_other_tool_category.py and 20260721_backfill_strip_agent_category.py collect affected_ids and emit logger.warning). Emit an INFO/WARNING per modified row with model_id, model_name, and before/after abilities.

src/xagent/core/tools/adapters/vibe/image_tool.py
```diff
@@ -74,6 +74,16 @@ def get_tools(self) -> list:
         can_generate = self._get_model() is not None
         can_edit = self._get_edit_model() is not None
 
+        # Withholding a tool is silent otherwise: the operator sees "editing
+        # stopped working" with nothing in the log pointing at the abilities.
+        if not can_edit or not can_generate:
```
rogercloud: [INFO] This INFO log fires on every get_tools() build — once per tools-list request (web/api/tools.py:341) and once per agent build (agent-cache-backed, so not per chat request). A per-instance flag would not help since each build is a fresh instance; a module-level log-once guard or logging at the model-load layer would bound the noise. Intentional diagnostic per the PR's purpose, so informational only.

tests/migrations/test_backfill_image_edit_ability.py
```diff
@@ -0,0 +1,126 @@
...
+def test_update_casts_the_json_parameter() -> None:
+    """Postgres rejects a text parameter into a json column; SQLite accepts it.
+
+    Every other test here runs on SQLite, so nothing else fails if the cast goes.
+    """
+    import importlib.util  # noqa: PLC0415
+
+    from sqlalchemy.dialects.postgresql import psycopg2  # noqa: PLC0415
+
+    path = Path("src/xagent/migrations/versions") / f"{REVISION}.py"
```
rogercloud: [MINOR] Test nits: the CWD-relative Path("src/xagent/migrations/versions") here deviates from the __file__-based convention in tests/alembic/ (it matches the rest of tests/migrations/ and is safe in CI, so acceptable); and the migration's native-list (non-str) abilities branch never executes in tests — a raw text() SELECT returns str on both SQLite and psycopg2, so the str branch is the real path. An empty-list fixture row and a downgrade round-trip were considered and dropped as branch-identical coverage and covered-by-precedent; fine to leave as-is.

bluefish-08:
Verified all seven; every one holds. Rather than patch the migration, I took the
primary suggestion from the first finding — a shared inference helper on the read
path — and dropped the migration entirely. One deviation from what you proposed,
called out below.

**Both MAJORs — fixed by removing the migration.** `default_image_abilities(provider,
model_name, fallback)` now lives in `core/model/image/base.py` and is reached only
through the `or` short-circuit at the four fallback sites. That covers rows created
after the change and fresh installs (which, as you noted, `db/migration.py:364-370`
stamps straight to head), and a declared abilities list never reaches the helper, so
a deliberate generate-only choice survives.

The deviation: you suggested keeping the migration and having it consume the same
definition. I dropped it instead. Once the fallback infers, the migration's only
remaining effect is rewriting rows that the fallback would already have handled —
while keeping the one behaviour you objected to, since at upgrade time NULL and an
explicit `["generate"]` are indistinguishable to it. Happy to add it back if you
want existing rows normalized in the data rather than at read time.

**Second MAJOR's per-provider scoping — adopted.** The helper only looks at the name
for `dashscope` and `gemini`; every other provider gets its call site's existing
default back unchanged. That keeps `openai` and `xinference` at their current
`["generate", "edit"]` and closes the xinference false-positive path you described:
a self-hosted `*-edit` model is not routed through name inference at all.

**On the provenance comment.** You were right that it was inaccurate, and checking it
turned up more than a wording problem: because every serving path passes an explicit
list, `gemini.py`'s inference genuinely never runs. The helper now uses those markers
at the one point where nothing was configured, which is the only place they can take
effect — and `gpt-image` is gone from the marker tuple, since it was never a gemini
rule and `openai` already defaults to both abilities.

**The remaining four** — `json.loads` guarding, the `as_sql` offline guard, per-row
audit logging, and the frozen-migration import convention — were all migration-scoped
and no longer apply. The `list()` char-split you found was a real corruption bug;
worth noting that the fallback path cannot hit it, as it only runs when `abilities`
is falsy.

**On the INFO log noise.** Left as-is per your assessment. It fires once per
`get_tools()` build rather than per chat request, and the diagnostic is the point of
this PR.

rogercloud (changes requested):
### Summary

PR #1334 (fix(image): infer edit ability at the fallback instead of backfilling rows) replaces the previously proposed Alembic backfill of image-model ability rows with read-time inference: a new `default_image_abilities()` helper infers `["generate", "edit"]` from the model name for dashscope/gemini rows that declare no abilities, wired into all five instance-construction sites, and the migration plus its test are dropped entirely. The approach covers existing NULL rows, rows created after deploy, and fresh installs (the migration stamps to head), while truthy `or` gates keep an operator's declared abilities authoritative — the right direction, and the simplest mechanism that covers all three cases. The blocking gap: none of the five fallback wirings is exercised by tests, so the core behavior change has zero regression protection.

### Update since last review

Since the previous review (which targeted the dropped backfill direction): the migration and its migration test were deleted — no SQL, no JSON parsing, no reflection at any point — and read-time `default_image_abilities()` was added at the five fallback sites (`model_service.py:569/582/595/610`, `adapter.py:35`), plus a gating log line for the withhold case. CI is all green: pre-commit, SQLite/PostgreSQL migrations, Pytest Fast web + integration, frontend-build, Real RAG E2E.

### Prior-findings checklist

| # | Round-1 finding | Status at head |
|---|---|---|
| P1 | Point-in-time backfill leaves post-deploy rows NULL (MAJOR) | FIXED — read-time inference covers existing NULL rows, later rows, fresh installs; migration deleted |
| P2 | Backfill overrides rows declared `["generate"]` (MAJOR) | FIXED — truthy `or` gates at all 5 sites keep declared lists authoritative; migration deleted |
| P3 | Provenance claim for marker rule (MINOR) | REFACTORED — markers now exactly gemini.py's; residual: comment cites gemini.py:174-178 but the condition is at :179 (folded into N3) |
| P4 | Unguarded `json.loads` on legacy rows (MINOR) | FIXED — migration deleted; no JSON parsing at read time |
| P5 | Offline `--sql` runs NoInspectionAvailable (MINOR) | FIXED — migration deleted; no reflection |
| P6 | No per-row upgrade logging (MINOR) | REFACTORED — moot: no DB writes; residual: inference granting edit in the success case is unlogged (folded into N13) |
| P7 | JSON bind `type_=sa.JSON` (MINOR) | FIXED — migration deleted; new code writes no SQL |
| P8 | INFO log per get_tools build (INFO) | NOT FIXED — log still fires per tools-list request / task run for generate-only deployments; gate limits it to withhold cases; informational, non-blocking |
| P9 | Test CWD-relative path nit (MINOR) | FIXED — test file deleted; replacement has no path usage |

### New findings

#### MAJOR

**N2 — The fallback wiring itself is untested (test quality).** `tests/core/model/image/test_default_abilities.py` exercises only the pure function `default_image_abilities`; none of the five `or default_image_abilities(...)` sites (`model_service.py:569/582/595/610`, `adapter.py:35`) is covered. Deleting any one of them keeps the suite green, so the PR's core behavior change has zero regression protection. Suggest: one adapter-level test (dashscope `qwen-image-edit`, `abilities=None` → instance abilities `["generate", "edit"]`) and one model_service-level test (NULL-abilities row → produced instance carries inferred abilities).

#### MINOR

**N1 — Adapter-path NULL rows for openai/xinference still resolve to `["generate"]` only.** `adapter.py:34-35` defaults NULL rows to `["generate"]` for every provider, while the web call sites default openai/xinference to `["generate", "edit"]` (`model_service.py:595/610`). A NULL openai/xinference row chosen as `default_edit_model` is then rejected by `_get_edit_model`'s `has_ability` check (`core/tools/core/image_tool.py:351`) and silently replaced by the first edit-capable model in registry order — wrong-model selection in multi-model deployments. Pre-existing at base; the PR neither worsens it (byte-identical outcome) nor fixes it. Round-1 comment 3774926229 raised this area, and the author's reply ("openai and xinference stay at `["generate", "edit"]`") is only true of the web call sites, not the adapter. Suggest aligning the adapter fallback per provider, or filtering by abilities in `get_default_image_edit_model` (`model_service.py:797+`).

**N3 — Marker rule duplicated from gemini.py with a stale citation.** `base.py:4-7` re-declares the name-marker rule gemini.py:179 already encodes; the comment cites gemini.py:174-178 but the condition is at :179. Have gemini.py consume `EDIT_CAPABLE_NAME_MARKERS` (or `default_image_abilities`; import-safe, no cycle) and fix the citation.

**N4 — Docstring overclaims.** `base.py:19-23` says a declared abilities list "must never reach this function", but `[]` is falsy and passes all five gates, so an explicitly declared empty list does reach it. Reword to "a declared non-empty abilities list is authoritative".

**N8 — Model-name case-insensitivity is untested.** The `.lower()` on `model_name` (`base.py:26`) is mutation-live, but every parametrized name in `test_default_abilities.py` is already lowercase. Add e.g. `("dashscope", "QWEN-IMAGE-EDIT", GENERATE, BOTH)`.

**N9 — PR description inaccuracies.** The body says "four fallback sites" but five call it, and "openai and xinference already default to `["generate", "edit"]`" is false for the adapter site (defaults `["generate"]`). Description-only.

**N11 — Provider extension path undocumented.** `EDIT_CAPABLE_NAME_MARKERS` applies to every allowlisted provider, but how a new provider opts into name inference is undocumented. A one-line comment on `_NAME_INFERRED_PROVIDERS` (`base.py:12`) suffices.

**N13 — Stored vs effective abilities split (low).** `GET /models` returns stored abilities (`model_store.py:72`), so a NULL row inferred to `generate+edit` shows no badges and is excluded from the image_edit default-model picker (`web/api/model.py:356-401` uses stored abilities) while runtime editing works; the new gating log fires only in the withhold case, so the grant is silent. Non-blocking; suggest exposing effective abilities for image rows in serialization.

- L7: shrink: `EDIT_CAPABLE_NAME_MARKERS` re-declares the name-marker rule gemini.py:179 already encodes; have gemini.py's constructor reuse this constant (or `default_image_abilities`) instead of a second copy that must be kept in sync.
- L595: shrink: the openai/xinference branches (L595, L610) constant-fold — neither provider is in `_NAME_INFERRED_PROVIDERS`, so the call always returns its own `["generate", "edit"]` fallback; restore the pre-PR literal there.
- net: -19 lines possible

Blocking issues:

| file:line | Severity | Description | Source |
|---|---|---|---|
| tests/core/model/image/test_default_abilities.py:29 | MAJOR | None of the five `or default_image_abilities(...)` fallback wirings is tested; deleting any one keeps the suite green | [new] |

Nothing else blocks: N1 is MINOR and pre-existing (not a regression), P8 is informational, and simplifications never block.

tests/core/model/image/test_default_abilities.py
```diff
@@ -0,0 +1,30 @@
...
+        # including one whose name would otherwise match.
+        ("openai", "gpt-image-1", BOTH, BOTH),
+        ("xinference", "my-image-edit", BOTH, BOTH),
+        ("xinference", "my-image-edit", GENERATE, GENERATE),
+        ("unknown-provider", "something-edit", GENERATE, GENERATE),
+    ],
+)
+def test_default_image_abilities(
+    provider: str, model_name: str, fallback: list[str], expected: list[str]
+) -> None:
```
rogercloud, line 33: These tests cover only the pure function; none of the five `or default_image_abilities(...)` fallback wirings (model_service.py:569/582/595/610, adapter.py:35) is exercised, so deleting any one keeps the suite green. Suggest an adapter-level test (dashscope `qwen-image-edit`, `abilities=None` → instance abilities `["generate", "edit"]`) and a model_service-level test (NULL-abilities row → produced instance carries inferred abilities).

src/xagent/core/model/image/adapter.py
```diff
@@ -29,7 +29,11 @@ def get_image_model_instance(db_model: Any) -> BaseImageModel:
     model_name = str(db_model.model_name)
     api_key = str(db_model.api_key) if db_model.api_key else None
     base_url = str(db_model.base_url) if db_model.base_url else None
-    abilities = list(db_model.abilities) if db_model.abilities else ["generate"]
+    abilities = (
+        list(db_model.abilities)
+        if db_model.abilities
```
rogercloud: This default (`["generate"]`) diverges from the web call sites, which default openai/xinference NULL rows to `["generate", "edit"]` (model_service.py:595/610). A NULL openai/xinference row chosen as `default_edit_model` is then rejected by `_get_edit_model`'s `has_ability` check (core/tools/core/image_tool.py:351) and silently replaced by the first edit-capable model. Pre-existing, not introduced here — but worth aligning the per-provider fallback or filtering by abilities in `get_default_image_edit_model` (model_service.py:797+).

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,33 @@
 from abc import ABC, abstractmethod
 from typing import Any, List
 
+# The markers gemini.py:174-178 infers from. Its inference never runs in the web
+# layer -- every call site passes an explicit abilities list -- so the same rule
+# is applied here, where the operator has configured nothing at all.
+EDIT_CAPABLE_NAME_MARKERS = ("edit", "3-pro")
```
rogercloud: This constant re-declares the name-marker rule gemini.py:179 already encodes (`if "3-pro" in model_lower or "edit" in model_lower`); the citation in the comment (gemini.py:174-178) is stale — the condition is at :179. Two copies must be kept in sync. Have `GeminiImageModel` consume this constant (or `default_image_abilities`; import-safe, no cycle) and fix the citation.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,33 @@
...
+
+
+def default_image_abilities(
+    provider: str, model_name: str, fallback: List[str]
+) -> List[str]:
+    """Abilities for an image model whose row declares none.
+
+    Only for the unconfigured case: a declared abilities list is authoritative and
+    must never reach this function, or an operator's deliberate generate-only
+    choice gets overridden.
```
rogercloud: The docstring says a declared abilities list "must never reach this function", but `[]` is falsy, so an explicitly declared empty list passes all five `or` gates and does reach it. Reword to "a declared non-empty abilities list is authoritative".

tests/core/model/image/test_default_abilities.py
```diff
@@ -0,0 +1,30 @@
...
+from xagent.core.model.image.base import default_image_abilities
+
+GENERATE = ["generate"]
+BOTH = ["generate", "edit"]
+
+
+@pytest.mark.parametrize(
+    ("provider", "model_name", "fallback", "expected"),
+    [
+        ("dashscope", "qwen-image-edit", GENERATE, BOTH),
```
rogercloud: `model_name.lower()` (base.py:26) is live, but every parametrized name here is already lowercase, so case-insensitivity is mutation-live untested. Add a row such as `("dashscope", "QWEN-IMAGE-EDIT", GENERATE, BOTH)`.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,33 @@
...
 
+# The markers gemini.py:174-178 infers from. Its inference never runs in the web
+# layer -- every call site passes an explicit abilities list -- so the same rule
+# is applied here, where the operator has configured nothing at all.
+EDIT_CAPABLE_NAME_MARKERS = ("edit", "3-pro")
+
+# Only dashscope and gemini ship both edit-capable and generate-only image models
+# under one provider, so only they need the name looked at. Every other provider
+# keeps whatever its call site already defaulted to.
+_NAME_INFERRED_PROVIDERS = ("dashscope", "gemini")
```
rogercloud: This tuple is the opt-in list for name-based ability inference, but how a new provider joins it is undocumented — the marker list above then applies to it implicitly. A one-line comment ("add a provider here to infer abilities from its model names") suffices.

src/xagent/web/services/model_service.py
```diff
@@ -564,31 +564,55 @@ def get_image_models(db: Session, user_id: Optional[int] = None) -> Dict[str, An
...
                     _add_image_model_with_id(image_models, image_model, db_model)
                 elif model_provider == "openai":
                     image_model = OpenAIImageModel(
                         model_name=str(db_model.model_name),
                         api_key=api_key,
                         base_url=base_url,
-                        abilities=list(db_model.abilities or ["generate", "edit"]),  # pyright: ignore[reportArgumentType]
+                        abilities=list(
+                            db_model.abilities
+                            or default_image_abilities(
```
rogercloud: This branch (and the xinference one at L610) constant-folds: neither provider is in `_NAME_INFERRED_PROVIDERS`, so `default_image_abilities` unconditionally returns the `["generate", "edit"]` fallback passed in. Restore the pre-PR literal and drop the call here.

bluefish-08:
All seven verified and applied, except the pre-existing one you flagged as such.

**Constant-folding at model_service.py:595/610** — right, those two calls could not
do anything: neither provider is in `_NAME_INFERRED_PROVIDERS`, so the helper
returned the fallback it was handed. Restored the literals. I had added them for
symmetry, which is exactly the wrong reason to call a function.

**Duplicated marker rule** — `GeminiImageModel` now consumes
`EDIT_CAPABLE_NAME_MARKERS`, so there is one copy. The stale `:174-178` citation is
gone rather than corrected: with gemini importing the constant, the comment no longer
needs to point anywhere.

**Wiring untested** — added `test_default_abilities_wiring.py`: adapter-level
(dashscope `qwen-image-edit` with `abilities=None` → `["generate", "edit"]`,
`wanx-v1` → `["generate"]`, and a declared `["generate"]` surviving on an
edit-named model) and model_service-level (NULL-abilities row through
`get_image_models`). Verified by mutation — reverting either wiring to its literal
turns these red, which the pure-function tests did not.

**Empty-list docstring** — reworded to "a declared non-empty abilities list is
authoritative and short-circuits before this function".

**Case-insensitivity** — added `("dashscope", "QWEN-IMAGE-EDIT", ...)`.

**Provider opt-in undocumented** — the comment on `_NAME_INFERRED_PROVIDERS` now says
how to add one.

**adapter.py's `["generate"]` vs the web sites' `["generate", "edit"]`** — agreed,
and agreed it predates this PR: a NULL openai/xinference row reaching
`get_default_image_edit_model` is silently swapped out by `_get_edit_model`. Left
alone here to keep this PR to one behaviour change. Happy to open a follow-up issue
for the per-provider fallback alignment unless you would rather it ride along.

bluefish-08:
Changed my mind on the pre-existing one you flagged at `adapter.py:34` — a self-review
pass reached the same finding independently, so I folded it in rather than deferring.

`default_image_abilities` now owns the provider default outright and lost its
`fallback` parameter. `openai`/`xinference` return both abilities for any name,
`dashscope`/`gemini` let the name decide, everything else is generate-only. All five
sites call it, so the two paths that build a model from one row —
`get_image_model_instance` and `model_service.get_image_models` — cannot answer
differently for the same row. That closes the swap you described: an unconfigured
openai row picked as the default edit model no longer comes back generate-only from
one path and editable from the other.

Note this supersedes the constant-folding cleanup from your last round. Those two
branches call the helper again, but it is no longer a fold — the helper now decides
`["generate", "edit"]` for those providers instead of echoing an argument.

`test_both_paths_agree_on_a_null_row` pins it across both paths; mutation-checked by
reverting the adapter wiring, which reddens it along with the adapter test.

Also from the self-review: `_available_models_summary()` was evaluated eagerly at the
log call regardless of level, so the withheld-tools log is now behind
`logger.isEnabledFor(logging.INFO)`. Level kept at INFO per your assessment of the
call frequency.

And a known ceiling now recorded in the PR body: inference reads the model name, not
the endpoint, so a dashscope row named `*-edit` pointed at a text2image `base_url`
registers `edit_image` and fails at call time, where before it was withheld. Judged
the right trade, but it does narrow what #1294's gating catches.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,35 @@
 from abc import ABC, abstractmethod
 from typing import Any, List
 
+# Consumed by GeminiImageModel's own name detection too, so the rule has one home.
+EDIT_CAPABLE_NAME_MARKERS = ("edit", "3-pro")
+
+# Providers whose whole image lineup edits. Add one here when that is true of it.
+_EDIT_CAPABLE_PROVIDERS = ("openai", "xinference")
```
rogercloud, line 9: Treating openai/xinference as whole-lineup edit providers overclaims for both: xinference's constructor defaults to stable-diffusion-2-1 with own abilities ['generate'] (xinference.py:21,47-49) and edit_image raises RuntimeError unless the backend handle exposes image_to_image/inpainting (xinference.py:231-253); the openai frontend advertises DALL-E 3 (en.ts:1843), which cannot serve images.edit (openai.py:185). No new user-visible failure class — the doomed edit tool was already registered at BASE via get_image_models for these NULL rows, and the PR converts a silent swap into a deterministic error — but consider documenting the assumption here or filtering get_default_image_edit_model by actual edit ability.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,35 @@
...
+
+# Providers shipping both edit-capable and generate-only models under one name, so
+# the model name decides. Add a provider here to infer its abilities from names.
+_NAME_INFERRED_PROVIDERS = ("dashscope", "gemini")
+
+
+def default_image_abilities(provider: str, model_name: str) -> List[str]:
```
rogercloud, line 19: The PR body names this helper default_image_abilities(provider, model_name, fallback) — a 3-arg signature with a fallback override. The code is 2-arg with the provider tables internal. Please align the PR body with the implemented signature.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,35 @@
 from abc import ABC, abstractmethod
 from typing import Any, List
 
+# Consumed by GeminiImageModel's own name detection too, so the rule has one home.
+EDIT_CAPABLE_NAME_MARKERS = ("edit", "3-pro")
```
rogercloud: The shared marker tuple couples dashscope and gemini inference: '3-pro' is Gemini vocabulary, but it also matches dashscope names through _NAME_INFERRED_PROVIDERS. No real dashscope name collides today (vacuous-but-latent, verified repo-wide), yet a future Gemini marker would silently change dashscope's NULL-row answer. A per-provider marker mapping would remove the coupling.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,35 @@
...
+# Providers whose whole image lineup edits. Add one here when that is true of it.
+_EDIT_CAPABLE_PROVIDERS = ("openai", "xinference")
+
+# Providers shipping both edit-capable and generate-only models under one name, so
+# the model name decides. Add a provider here to infer its abilities from names.
+_NAME_INFERRED_PROVIDERS = ("dashscope", "gemini")
+
+
+def default_image_abilities(provider: str, model_name: str) -> List[str]:
+    """Abilities for an image model whose row declares none.
```
rogercloud, line 20: Nothing in the docs/README/frontend describes the abilities convention, and after this PR a model's NAME is a behavioral switch: qwen-image-edit vs wanx-v1 differ only by name, invisible to operators. A short note in docs/deployment.md or the model-config docs would make the convention discoverable.

src/xagent/core/model/image/base.py
```diff
@@ -1,6 +1,35 @@
...
+    -- cannot disagree about what it can do. A declared non-empty abilities list is
+    authoritative and short-circuits before this function, or an operator's
+    deliberate generate-only choice gets overridden.
+    """
+    normalized = provider.strip().lower()
+    if normalized in _EDIT_CAPABLE_PROVIDERS:
+        return ["generate", "edit"]
+    if normalized in _NAME_INFERRED_PROVIDERS:
+        lowered = model_name.lower()
+        if any(marker in lowered for marker in EDIT_CAPABLE_NAME_MARKERS):
```
rogercloud: Name-trust ceiling worth one docstring sentence: a marker match trusts the name over the endpoint, so an advertised edit ability can fail 4xx at call time if the backend model differs. The PR body's Known-ceiling section covers it at PR level; the helper itself should carry the caveat.

tests/core/model/image/test_default_abilities_wiring.py
```diff
@@ -0,0 +1,106 @@
...
+@pytest.mark.parametrize(
+    ("provider", "model_name"),
+    [
+        ("openai", "gpt-image-1"),
+        ("xinference", "sd-3.5"),
+        ("dashscope", "qwen-image-edit"),
+        ("dashscope", "wanx-v1"),
+    ],
+)
+def test_both_paths_agree_on_a_null_row(provider: str, model_name: str) -> None:
```
rogercloud: Mutation gap: the gemini branch of model_service.get_image_models (model_service.py:580-585) is unprotected — _service_abilities only exercises dashscope and this parametrization covers openai/xinference/dashscope, never gemini. Reverting the gemini branch to ['generate'] leaves this suite green and demotes NULL gemini-3-pro rows. This also makes the PR body's claim that "reverting either wiring turns these red" false for the gemini site; add a gemini row to the parametrization (or a gemini service case) to close it.

rogercloud: This assertion compares two calls to the same helper (cross-equality), so a symmetric regression — both sites reverted to generate-only — passes green. Pin one openai/xinference row to an expected value (e.g. assert adapter_model.abilities == ['generate', 'edit'] for the openai row). The docstring's second sentence is also grammatically broken ("makes ... hand back a model _get_edit_model then rejects").

tests/core/model/image/test_default_abilities_wiring.py
```diff
@@ -0,0 +1,106 @@
...
+    model = get_image_model_instance(_db_model())
+    assert model.abilities == ["generate", "edit"]
+
+
+def test_adapter_leaves_a_generate_only_name_alone() -> None:
+    model = get_image_model_instance(_db_model(model_name="wanx-v1"))
+    assert model.abilities == ["generate"]
+
+
+def test_adapter_never_overrides_a_declared_list() -> None:
```
rogercloud, line 39: The declared-list short-circuit is only tested at the adapter; _service_abilities hardcodes abilities=None, so the four service `or` sites are never exercised with a declared list. A case where the service path must NOT override a declared generate-only list would pin the `or` semantics on the service side.

src/xagent/core/tools/adapters/vibe/image_tool.py
```diff
@@ -73,6 +76,17 @@ def get_tools(self) -> list:
         can_generate = self._get_model() is not None
         can_edit = self._get_edit_model() is not None
 
+        # Withholding a tool is silent otherwise: the operator sees "editing
+        # stopped working" with nothing in the log pointing at the abilities.
+        # isEnabledFor guards the summary, which walks every model eagerly.
+        if (not can_edit or not can_generate) and logger.isEnabledFor(logging.INFO):
```
rogercloud, line 82: The withhold log calls _available_models_summary(), which walks only self._image_models, while _get_model/_get_edit_model consult the defaults first (core/image_tool.py:327,351). With an empty configured list and a serving default, the log reads "from models: none configured" while a default actually serves. Only reachable via direct construction (the production path short-circuits), so minor — but including the default models in the summary would make the log truthful.

src/xagent/web/services/model_service.py
```diff
@@ -564,31 +564,51 @@ def get_image_models(db: Session, user_id: Optional[int] = None) -> Dict[str, An
                         model_name=str(db_model.model_name),
                         api_key=api_key,
                         base_url=base_url,
-                        abilities=list(db_model.abilities or ["generate"]),  # pyright: ignore[reportArgumentType]
+                        abilities=list(
```
rogercloud: shrink: this abilities=list(db_model.abilities or default_image_abilities(model_provider, str(db_model.model_name))) block is verbatim at 567/580/593/606 — hoist into locals (model_name = str(db_model.model_name); abilities = list(...)) computed once inside the per-row try, so exceptions keep hitting the per-row except→logger.warning containment. ~21 lines saved.

src/xagent/core/model/image/adapter.py
```diff
@@ -29,7 +29,11 @@ def get_image_model_instance(db_model: Any) -> BaseImageModel:
     model_name = str(db_model.model_name)
     api_key = str(db_model.api_key) if db_model.api_key else None
     base_url = str(db_model.base_url) if db_model.base_url else None
-    abilities = list(db_model.abilities) if db_model.abilities else ["generate"]
+    abilities = (
+        list(db_model.abilities)
```
rogercloud: shrink: lines 33-37 collapse to abilities = list(db_model.abilities or default_image_abilities(provider, model_name)) — behavior-identical and matches the model_service form. 4 lines saved.

tests/core/model/image/test_default_abilities_wiring.py
```diff
@@ -0,0 +1,106 @@
...
+        is_active=True,
+        model_provider=provider,
+        model_name=model_name,
+        api_key="test-key",
+        base_url="https://example.invalid",
+        abilities=None,
+        description=None,
+    )
+
+    class _Query:
```
rogercloud: stdlib: the 11-line _Query/_Session stubs collapse to a 3-line mocker.Mock chain (pytest-mock is a dev dependency and the suite-wide convention). ~8 lines saved.

src/xagent/core/model/image/gemini.py
```diff
@@ -176,7 +176,7 @@ def __init__(
         else:
             # Auto-detect abilities based on model name
             model_lower = model_name.lower()
-            if "3-pro" in model_lower or "edit" in model_lower:
+            if any(marker in model_lower for marker in EDIT_CAPABLE_NAME_MARKERS):
```
rogercloud: shrink: this auto-detect if/else is exactly default_image_abilities('gemini', model_name) — replace the whole block with self._abilities = abilities or default_image_abilities('gemini', model_name). Behavior-identical (verified term-for-term); EDIT_CAPABLE_NAME_MARKERS could then go private. ~6 lines saved.

rogercloud (approved):
## Summary

This PR fixes the silent loss of edit_image that #1294 exposed: image model rows normally store NULL abilities and every read path defaulted NULL to generate-only, so edit-capable rows (qwen-image-edit, gemini-3-pro-image) were reported generate-only and the edit tool was withheld. A new `default_image_abilities(provider, model_name)` helper in `src/xagent/core/model/image/base.py` answers the unconfigured-row question in one place — marker-based name inference (`'edit'`/`'3-pro'`) for dashscope/gemini, generate+edit for openai/xinference, generate otherwise — and all five fallback sites (`adapter.get_image_model_instance` plus the four provider branches of `model_service.get_image_models`) now call it, so the two model-building paths can no longer disagree. A declared non-empty abilities list remains authoritative and short-circuits the inference, so an operator's deliberate generate-only choice is never overridden. Blocking: no — recommended event: APPROVE

## Update since last review (86b3cec → 0235063)

Two commits: 9239e4d "share the name markers with gemini and cover the fallback wiring" (gemini.py imports `EDIT_CAPABLE_NAME_MARKERS`; new `test_default_abilities_wiring.py`; case/whitespace pure-function cases) and 0235063 "give both model-building paths one answer for an unconfigured row" (the per-site fallback argument is gone — the helper is now 2-arg with internal provider tables; the adapter path is aligned to the service path; `isEnabledFor` guard on the withhold log). The Alembic migration and its tests were deleted at 86b3cec, so the pivot to read-time inference is complete with no backfill. CI is green (pre-commit, SQLite/Postgres migrations, Pytest Fast web+integration, frontend build, Real RAG E2E).

## Approach verdict

Acceptable-with-reservations. Rejecting the point-in-time backfill in favor of read-time fallback inference is right: a backfill cannot distinguish NULL from a deliberate `['generate']`, misses fresh installs, and stamps rows point-in-time, while the helper gives all five fallback sites one answer. The reservations are the three severity-adjusted scope/premise findings below — V1 (provider overclaim), V3 (inference invisible to row-reading surfaces), V4 (frontend preloads generate-only): the helper's answers are only as right as its provider policy, and every surface that reads raw stored abilities still sees NULL.

## Prior-findings checklist

| id | prior severity | status | evidence |
|---|---|---|---|
| P1 point-in-time backfill | MAJOR | FIXED | migration deleted; read-time inference covers NULL/later rows/fresh installs |
| P2 backfill overrides declared | MAJOR | FIXED | truthy or-gates at all 5 sites; migration deleted |
| P3 marker provenance comment | MINOR | FIXED | folded into N3; citation replaced |
| P4 json.loads on legacy rows | MINOR | FIXED | migration deleted; no JSON parsing |
| P5 offline --sql reflection | MINOR | FIXED | migration deleted |
| P6 per-row upgrade logging | MINOR | FIXED | moot — no DB writes; residual folded into N13 |
| P7 JSON bind type_=sa.JSON | MINOR | FIXED | migration deleted |
| P8 INFO log per get_tools build | INFO | NOT FIXED | isEnabledFor guard (0235063) is a computation optimization, not a frequency fix; still fires per tools-list request/agent build at INFO level. Informational, non-blocking |
| P9 test CWD-relative path | MINOR | FIXED | test file deleted |
| N1 adapter NULL divergence | MINOR | FIXED | adapter now calls the 2-arg helper (0235063); both paths agree; agreement test pins it |
| N2 wiring untested | MAJOR | PARTIAL | wiring test added; 4/5 sites mutation-protected; the gemini service branch (model_service.py:580-585) is unprotected — tracked as V8 (minor). The author's "reverting either wiring turns these red" claim is false for the gemini site |
| N3 marker duplication + stale citation | MINOR | FIXED | gemini.py imports `EDIT_CAPABLE_NAME_MARKERS` (:21, used :179); citation corrected |
| N4 docstring overclaim | MINOR | FIXED | docstring now says "declared non-empty abilities list is authoritative" (base.py:20-21) |
| N8 case-insensitivity untested | MINOR | FIXED | ("dashscope","QWEN-IMAGE-EDIT",BOTH) and ("  DashScope  ",...) cases added |
| N9 PR description inaccuracies | MINOR | FIXED | body says "All five fallback sites"; "settles a pre-existing split" paragraph verified accurate end-to-end. Residual cosmetic: Background cites model_service.py:567,575 — stale base-version line numbers (current: 569/582/595/608) |
| N11 provider extension path | MINOR | FIXED | "Add a provider here..." comments on both provider tuples (base.py:7,12) |
| N13 stored vs effective abilities | MINOR | NOT FIXED | all serialization/validation surfaces still read raw NULL — tracked as V3 (minor) |
| L595 constant-fold openai/xinference | simplification | REFACTORED | 3-arg fallback signature removed; unconditional answer is now explicit named policy in `_EDIT_CAPABLE_PROVIDERS`; restoring literals would fork 5 identical sites and reintroduce the N1 divergence. Residual caveat: openai/xinference never consult model_name — documented policy |

## New findings (all minor; no critical/major)

**Scope/premise (severity-adjusted from major):**
- **V1 — edit-provider overclaim (minor)** `src/xagent/core/model/image/base.py:8`: treating openai/xinference as whole-lineup edit providers overclaims for both: xinference's constructor defaults to stable-diffusion-2-1 with own abilities `['generate']` (xinference.py:21,47-49) and `edit_image` raises RuntimeError unless the backend handle exposes image_to_image/inpainting (xinference.py:231-253); the openai frontend advertises DALL-E 3 (en.ts:1843), which cannot serve `images.edit` (openai.py:185). The adapter's NULL-row answer widened from `['generate']` to `['generate','edit']` (0235063). No new user-visible failure class — the doomed edit tool was already registered at BASE via `get_image_models` for these NULL rows, and the PR converts a silent swap into a deterministic error — but the assumption should be documented or filtered. Inline on base.py:8.
- **V2 — single-source claim (minor, adjusted major→low)** `base.py:4`: the comment is literally true (gemini.py:21 imports the constant) and the docstring scopes the claim to the two row-building paths, which do agree. Constructor fallbacks (dashscope.py:31, xinference.py:46-49, openai.py:32, web/api/model.py:768) are pre-existing, out-of-diff, and unreachable from production rows (every caller passes a non-empty list). Residual: gemini.py:179 still re-implements the rule instead of calling the helper — covered as VS4.
- **V3 — inference invisible to row-reading surfaces (minor)** `src/xagent/web/services/model_store.py:72`: serializes raw abilities; `_is_default_config_type_compatible` (web/api/model.py:356-397, called :1765) rejects NULL rows for an image_edit default (400); the default-generate SQL `cast(...).contains('"generate"')` (model_service.py:728,761) excludes NULL rows; frontend badges (models.tsx:541) and the settings picker (default-models.tsx:102-103) read raw abilities. Runtime editing works; surfaces pre-existing, none in the PR diff. Previously raised as N13, never addressed. Body-only (not in diff).
- **V4 — frontend preloads generate-only (minor)** `frontend/src/components/pages/model-management-dialog.tsx:160`: the wizard always submits `abilities=['generate']` for image (also :310/:642/:887/:930), and `[]` is rejected by schema (422), so wizard-created rows can never hit the inference; NULL rows come only from API clients or legacy rows. The PR premise that "an image model's abilities is normally NULL" is false for the primary creation path. Non-regression (a declared list is authoritative by design) and the wizard exposes generate/edit pills, but this bounds the fix's blast radius. Body-only (not in diff).

**Test coverage:**
- **V8 — gemini service branch untested (minor)** `tests/core/model/image/test_default_abilities_wiring.py:96`: the gemini branch of `model_service.get_image_models` (model_service.py:580-585) is mutation-unprotected — `_service_abilities` is only exercised with dashscope, and the agreement parametrization covers openai/xinference/dashscope. Reverting the gemini branch to `['generate']` leaves the suite green and demotes NULL gemini-3-pro rows. (The adapter half of the original claim is overstated — the adapter is provider-generic and covered.) Inline on :96.
- **V9 — service short-circuit untested (minor)** `test_default_abilities_wiring.py:39`: the declared-list short-circuit is tested only at the adapter (`test_adapter_never_overrides_a_declared_list`); `_service_abilities` hardcodes `abilities=None`, so the four service `or` sites are never exercised with a declared list. Inline on :39.
- **V11 — agreement-test tautology (minor)** `test_default_abilities_wiring.py:96`: `test_both_paths_agree_on_a_null_row` asserts cross-equality between two calls to the same helper; it catches the original asymmetric split (verified against BASE) but a symmetric regression (both sites reverted to generate-only) passes green. Suggest an expected-value assertion for one openai/xinference row. Docstring's second sentence is grammatically broken. Inline on :96.
- **V5 — empty list treated as NULL (informational)** `[]` is falsy and falls through the `or` to the inference; storable only via ModelUpdate without category (schemas/model.py:193-197) — pre-existing, out-of-diff; the docstring's "non-empty is authoritative" is the N4 fix. Residual one-line suggestion: validate `[]` on ModelUpdate like ModelCreate. Body-only.

**Docs drift:**
- **V6 — PR-body signature drift (minor)** `base.py:15`: the PR body names a 3-arg signature `default_image_abilities(provider, model_name, fallback)`; the code is 2-arg. Body/code drift. Inline on base.py:15.
- **V12 — gemini docstring drift (minor)** `gemini.py:140`: class docstring prose describes only the "3-pro" rule while the shared constant also matches `'edit'`. Pre-existing prose, but adjacent to the PR's single-sourcing. Body-only (line not in diff).
- **V13 — operator docs (minor)** `base.py:16`: no docs/README/frontend-helper text mentions abilities or the name-marker convention; after this PR a model's NAME is a behavioral switch operators cannot see. Suggest a note in docs/deployment.md or model-config docs. Inline on base.py:16.
- **V14 — ceiling docstring (minor)** `base.py:29`: the docstring lacks the name-trust ceiling — a name match trusts the name over the endpoint, so an advertised edit ability can fail 4xx at call time. The PR body's Known-ceiling section covers it at PR level only; one docstring sentence closes the gap. Inline on base.py:29.

**Logging & coupling:**
- **V10 — withhold-log summary omits defaults (minor)** `src/xagent/core/tools/adapters/vibe/image_tool.py:82`: the new withhold log calls `_available_models_summary()` (core/image_tool.py:367-383), which walks only `self._image_models`, while `_get_model`/`_get_edit_model` consult defaults first (:327, :351). Empty dict + serving default → the log says "from models: none configured" while a default serves. Only reachable via direct construction (production path short-circuits), so minor. Inline on :82.
- **V7 — shared markers couple providers (minor)** `base.py:5`: `EDIT_CAPABLE_NAME_MARKERS` applies `'3-pro'` (Gemini-only vocabulary) to dashscope inference; no real dashscope name collides today (verified repo-wide), so vacuous-but-latent — a future gemini marker silently changes dashscope's NULL-row answer. Suggest per-provider marker mapping. Inline on base.py:5.

- `shrink:` `src/xagent/web/services/model_service.py:567` — hoist the four verbatim `abilities=list(db_model.abilities or default_image_abilities(...))` blocks (567/580/593/606) into locals computed once inside the per-row try, so exceptions keep hitting the per-row `except`→`logger.warning` containment. ~21 lines saved.
- `shrink:` `src/xagent/core/model/image/adapter.py:33` — collapse lines 33-37 to `abilities = list(db_model.abilities or default_image_abilities(provider, model_name))`; behavior-identical and matches the model_service form. 4 lines saved.
- `stdlib:` `tests/core/model/image/test_default_abilities_wiring.py:62` — replace the 11-line `_Query`/`_Session` stubs with a 3-line `mocker.Mock` chain (pytest-mock is a dev dep, suite-wide convention). ~8 lines saved.
- `shrink:` `src/xagent/core/model/image/gemini.py:179` — replace the auto-detect if/else with `self._abilities = abilities or default_image_abilities('gemini', model_name)`; behavior-identical (verified term-for-term); `EDIT_CAPABLE_NAME_MARKERS` could then go private. ~6 lines saved.
- Net: ~-39 lines possible. Update range (86b3cec..HEAD) otherwise lean.

Blocking: no. No confirmed critical or major findings: all four originally-major findings were severity-adjusted to minor/low after verification (V1, V3, V4 scope/premise; V2 single-source), and the remaining residuals are minor — N2's residual is the gemini-branch test gap (V8), and N13 (stored vs effective abilities) plus P8 (INFO log frequency) are pre-existing surfaces outside this PR's diff. Nothing in the update range blocks the change: the pivot to read-time inference is correct, both model-building paths now agree, and the fallback wiring is mutation-tested at 4/5 sites. Recommended event: APPROVE.

bluefish-08:
All verified. Twelve applied, one deliberately not — details below.

**Overclaimed provider whitelist.** You are right and I had it backwards: `xinference`
defaults to stable-diffusion-2-1 with `["generate"]` and raises unless the backend
exposes image_to_image/inpainting. The comment now says what the tuple actually is —
the default both web call sites already applied to their NULL rows, kept so this
change stays about the two paths agreeing rather than about widening access — instead
of asserting a lineup-wide capability that is not true.

**Marker coupling.** Split into `_NAME_MARKERS_BY_PROVIDER`: dashscope gets
`("edit",)`, gemini gets `("edit", "3-pro")`. A test pins that `wanx-3-pro` stays
generate-only, so a future Gemini marker cannot move dashscope's answer.

**Gemini mutation gap.** Correct, and it made the PR body's claim false. The service
parametrization now covers gemini both ways; mutation-checked by reverting the gemini
branch to a literal, which reddens two cases.

**Cross-equality assertion.** Also correct — a symmetric revert passed green. Both
sides are now pinned to an expected value; the symmetric mutation reddens 7 cases.
Docstring sentence rewritten.

**Declared-list short-circuit on the service side.** Added
`test_model_service_never_overrides_a_declared_list`.

**All four shrinks applied.** `gemini.py`'s auto-detect block is now
`abilities or default_image_abilities("gemini", model_name)` — you were right that it
was term-for-term the same, and it lets the markers go private. model_service hoists
`model_name`/`abilities` into locals inside the per-row try. adapter collapses to one
line. The service stubs are a `mocker.Mock` chain.

**PR body signature** — fixed, along with the provider wording above.

**Not applied: the summary excluding default models** (`image_tool.py:82`). Agreed it
is inaccurate in that state, but `_available_models_summary()` is also consumed by the
two error strings #1294 shipped, so changing it edits already-merged operator-facing
text for a state the production path cannot reach — `create_image_tools_from_config`
returns early on an empty model list. Happy to do it as a follow-up if you would
rather the log and both error strings all account for the defaults.

Issue #1321, opened 2026-08-13
Connector picker omits team-owned connectors: /api/mcp/apps skips the team visibility overlay that /api/mcp/servers applies


codeacme17:
## Bug

`GET /api/mcp/apps` — the endpoint backing the agent's "+ Connector" picker — does not apply the team-connector visibility overlay that `GET /api/mcp/servers` applies. When an application installs the `connector_team_scope` hooks, a team-owned connector that the requesting user has no personal association row for is listed by `/api/mcp/servers` (the Tools page) but is absent from `/api/mcp/apps?location=local` (the picker), so it cannot be attached to an agent from that dialog.

## Where the two endpoints diverge

`get_mcp_servers` overlays team-owned connectors on top of the user's own rows, for both connector types (`src/xagent/web/api/mcp.py:2151-2186`):

```python
from ..services.connector_team_scope import visible_team_connector_ids

team_ids = visible_team_connector_ids(db, effective_user_id)

own_mcp_ids = {int(server.id) for _um, server in user_mcps}
missing_mcp = [sid for sid in team_ids["mcp"] if sid not in own_mcp_ids]
...
own_api_ids = {int(api.id) for _ua, api in user_custom_apis}
missing_api = [aid for aid in team_ids["custom_api"] if aid not in own_api_ids]
```

`list_mcp_apps` has no such call anywhere in the function. Its `location in ["local", "all"]` branch iterates only `user_mcps` (`src/xagent/web/api/mcp.py:2001-2039`), which is built from the requesting user's own association rows (`:1857-1865`):

```python
user_mcps = [
    (server, user_mcp)
    for server, user_mcp in (
        db.query(MCPServer, UserMCPServer)
        .join(UserMCPServer, UserMCPServer.mcpserver_id == MCPServer.id)
        .filter(UserMCPServer.user_id == current_user.id)
        .all()
    )
]
```

The Custom API half of the same branch has the identical shape — it iterates `UserCustomApi` rows filtered by `current_user.id` (`:2042-2047`) — so a team-owned custom API is missing from the picker for the same reason.

Sharing a connector to a team writes a team link row and no per-member association, so for every member except the connector's creator both loops resolve nothing and the connector never reaches the picker.

## Steps to reproduce

Requires an application that installs the `connector_team_scope` visibility hook; standalone xagent keeps connectors user-owned and is unaffected.

1. User A creates a custom MCP connector (or Custom API) and shares it with the team.
[...]

PR #1335, opened 2026-08-13
fix(mcp): stop gating connector attach on the editor's own OAuth grant (#1332)

### LINKED ISSUE #1332: Connector picker gates attach on the editor's personal OAuth grant, blocking connectors with hook-resolved tokens
## Bug

The "+ Connector" picker refuses to attach a custom mcp_oauth MCP server to an agent unless the editor personally holds an active `MCPOAuthGrant` for it. For connectors whose access tokens are supplied out of band through `set_oauth_token_resolver_hook` (`src/xagent/web/tools/config.py`), that grant never exists by design — so the connector can never be attached from the picker, even though the backend accepts the attachment and the runtime resolves credentials without any grant.

Split out of #1313: PR #1323 fixes the Connect button's dead end (the interactive flow), but the gate in front of that button remains and is itself wrong for hook-resolved connectors.

## Mechanism

- Since #1097, `is_connected` for an mcp_oauth-shaped local entry in `GET /api/mcp/apps` requires an active `MCPOAuthGrant` for the requesting user (`_mcp_oauth_server_is_actually_connected`, `src/xagent/web/api/mcp.py`).
- In the picker's select mode, a card that is not connected cannot be toggled into the agent's selection; clicking it opens the detail modal whose only action is Connect (`handleCardClick`, `frontend/src/components/mcp/connect-mcp-dialog.tsx`).
- An embedding application can install `set_oauth_token_resolver_hook` to supply MCP OAuth access tokens at call time — for example per end user, provisioned server-to-server by the application. In that model there is no interactive consent, no identity the editor could sign in as, and therefore never a grant row for the editor. The picker reports the connector as permanently not connected, and even the repaired Connect flow (#1323) is not meaningful for it.
- The gate is UI-only: the agent create/update endpoints accept an `mcp:<server>` selector with no connected-state validation, and the runtime resolves tokens through the hook regardless of grants.

## Expected behavior

Attaching a connector to an agent does not require the editor to hold a personal completed grant, matching the backend contract. The connected checkmark can keep reflecting the editor's own grant state for the interactive flow, but selection must not be gated on it.

## Acceptance criteria

- [ ] In select mode, a custom mcp_oauth server with no active grant for the editor can be added to (and removed from) the agent's selection from the picker.
- [ ] The Connect action remains available for editors who do use the interactive flow (#1323 behavior unchanged).
- [ ] Connected-state display is unchanged; only the selection gating is relaxed.

codeacme17:
Closes #1332.

## Problem

In the "+ Connector" picker's select mode, a card could only be toggled into the agent's selection when its entry reported `is_connected`. For a custom `mcp_oauth` connector that flag is false precisely when **this editor** holds no completed `MCPOAuthGrant` — and an embedding application can supply those access tokens out of band through `set_oauth_token_resolver_hook`, per end user and provisioned server-to-server. In that model there is no interactive consent, no identity the editor could sign in as, and therefore never a grant row for them, so the connector was permanently unattachable from the picker.

The gate was UI-only: the agent create/update endpoints accept the connector's `mcp:<name>` selector with no connected-state validation, and the runtime resolves credentials through the hook regardless of grants (`_resolve_oauth_token_from_hook`, `src/xagent/web/tools/config.py`). #1323 fixed the Connect button's dead end behind this gate; the gate itself remained.

## Change

`handleCardClick` now consults a new `isAttachable` predicate instead of `is_connected` directly:

```ts
isAppConnected(app) || (Boolean(app.is_custom) && app.auth_type === "mcp_oauth")
```

Both conjuncts are load-bearing, and neither can be widened:

- **`is_custom`** — a *catalog* `mcp_oauth` app (e.g. Granola) carries the same `auth_type`, but there `is_connected: false` means the user has no association row for it at all. Connecting really is a prerequisite, so attaching it would select a connector that does not exist for them. Catalog entries stay gated.
- **`auth_type`** — `list_mcp_apps` emits it on a local entry only for the `mcp_oauth` shape *and only while the association is active* (`src/xagent/web/api/mcp.py`). A deactivated association is still listed with `is_connected: false`, and the runtime's server query filters on `UserMCPServer.is_active` (`src/xagent/core/tools/adapters/vibe/factory.py`) — so attaching one would silently load zero tools. That entry needs re-enabling rather than re-authorization, and keeps its existing card-click route to the detail modal.

Because the card click now toggles selection for the relaxed entries instead of opening the detail modal, the card gets its own trigger for that modal (`tools.mcp.dialog.authorize`), so the per-server OAuth flow repaired in #1323 stays reachable for editors who do authorize personally. It is keyed on the same `isAttachable` predicate, so every card that loses the click route gains the trigger and no card that kept the route gets a redundant second one. The label is deliberately not "Connect" — the select-mode footer already uses that label to commit the selection.

Connected-state display is untouched: the checkmark, the sharing badges and the settings dialog all keep reading `is_connected` exactly as before.

## Acceptance criteria

- [x] In select mode, a custom `mcp_oauth` server with no active grant for the editor can be added to (and removed from) the agent's selection from the picker.
- [x] The Connect action remains available for editors who use the interactive flow (#1323 behavior unchanged).
- [x] Connected-state display is unchanged; only the selection gating is relaxed.

## Tests

Six tests in `frontend/src/components/mcp/connect-mcp-dialog.test.tsx`. Two go red against the pre-change component:

- selects **and deselects** a custom `mcp_oauth` connector the editor holds no grant for;
- the card's Authorize trigger opens the detail modal without counting as a selection toggle.

The rest are regression guards that pass on the old behavior by construction, but each kills a distinct mutation of the new code:

| Guard | Mutation it kills |
| --- | --- |
| unconnected **catalog** apps stay gated (Granola + a keyless control) | dropping `isAttachable`'s `is_custom` conjunct |
| a **deactivated** custom `mcp_oauth` entry stays gated (no `auth_type`) | dropping the `auth_type` conjunct |
| a **connected** custom `mcp_oauth` entry shows Configure + checkmark, never Authorize | swapping the two card-footer branches |
| a confirmed consent **auto-selects** the connector (extends the existing popup-recheck test) | deleting the auto-select append in `handleConnectMcpOAuthApp` |
| outside select mode the card click keeps its original destination | dropping the `isSelectMode` guard |

The shared settings-dialog mock now surfaces its `open`/`app` props so a test can tell a modal route from a selection toggle, and the checkmark wrapper gained a `data-testid` so the display assertions name it instead of matching on layout classes.

Verified: `src/components/mcp` 69 tests pass, `tsc --noEmit` clean, `eslint` introduces no new problems on the changed files, plus `test:ci-manifest`, `test:home-build-contracts`, `src/i18n` (locale-tree parity) and `src/components/build` green. Every mutation in the table above was applied, run, and reverted.

## Scope

Frontend only — `src/xagent/web/api/mcp.py` is untouched.

An earlier version of this section claimed the change would make team-owned `mcp_oauth` connectors selectable once #1321 landed. #1321 has since landed as #1338, and on current main that claim is false, so it is withdrawn:

- A team-owned MCP row has no personal association, so `list_mcp_apps` withholds `auth_type` from it deliberately — `/{server_id}/oauth/connect` requires an active personal association and would 404. With `auth_type` absent, `isAttachable` gates the row, exactly as the old `is_connected` gate did. This PR does not change that, and it is not a regression: before #1338 the row was not listed at all.
- Team-owned **non**-`mcp_oauth` MCP servers and Custom APIs report `is_connected: true`, so #1338 alone made those attachable. This PR is not what makes them selectable either.

Expressing that population properly needs attachability and authorizability to be separate fields rather than inferred from the presence of `auth_type`; that is #1347, which tracks it as acceptance criterion 3.

### DIFF
--- a/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
@@ -1683,8 +1683,20 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     )
     await screen.findByText("Chrome")
 
+    // The card must render selected from the id-only seed before the click —
+    // isSelected matches by id OR name, not just name.
+    expect(screen.getByTestId("connector-card-chrome-devtools")).toHaveAttribute(
+      "data-selected",
+      "true",
+    )
+
     // One click on the visually-selected card must deselect it outright.
     fireEvent.click(screen.getByText("Chrome"))
+
+    expect(screen.getByTestId("connector-card-chrome-devtools")).toHaveAttribute(
+      "data-selected",
+      "false",
+    )
 
     fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
     expect(onConnectSelected).toHaveBeenCalledWith([])
@@ -1699,13 +1711,19 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
   // grant (#1332).
   const unauthorizedCustomMcp = () => ({ ...customMcpOauthApp(), name: "Records MCP" })
 
-  function renderSelectModeWith(apps: object[], onConnectSelected: () => void) {
+  // Shared by renderSelectModeWith below and the non-select-mode card-click
+  // test, which otherwise duplicated this exact mockImplementation block.
+  function mockAppsList(apps: object[]) {
     apiRequestMock.mockImplementation((url: string) => {
       if (url.includes("/api/mcp/apps?")) {
         return Promise.resolve({ ok: true, json: async () => apps })
       }
       throw new Error(`Unexpected request: ${url}`)
     })
+  }
+
+  function renderSelectModeWith(apps: object[], onConnectSelected: () => void) {
+    mockAppsList(apps)
     return render(
       <ConnectMcpDialog
         open
@@ -1726,8 +1744,12 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     expect(screen.getByTestId("settings-open-app").textContent).toBe("")
     // Becoming attachable must not change connected-state display: this
     // connector is still unconnected (is_connected: false), so it must stay
-    // unchecked.
-    expect(screen.queryByTestId("connected-check")).toBeNull()
+    // unchecked. Scoped through the card: connected-check is non-unique
+    // across the grid by construction, so an ungrounded query would throw as
+    // soon as a second connected app rendered.
+    expect(
+      within(screen.getByTestId("connector-card-records")).queryByTestId("connected-check"),
+    ).toBeNull()
     fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
     expect(onConnectSelected).toHaveBeenLastCalledWith(["Records MCP"])
 
@@ -1747,7 +1769,12 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
 
     expect(screen.getByRole("button", { name: "tools.mcp.dialog.configure" })).not.toBeNull()
     expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
-    expect(screen.getByTestId("connected-check")).not.toBeNull()
+    // Scoped through the card: connected-check is non-unique across the grid
+    // by construction, so an ungrounded query would throw as soon as a
+    // second connected app rendered.
+    expect(
+      within(screen.getByTestId("connector-card-records")).getByTestId("connected-check"),
+    ).not.toBeNull()
   })
 
   it("keeps the per-server OAuth flow reachable from the card in select mode (#1323)", async () => {
@@ -1795,13 +1822,19 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     expect(onConnectSelected).toHaveBeenLastCalledWith([])
   })
 
-  it("still gates selection on a custom mcp_oauth entry whose association is deactivated", async () => {
+  it("still gates selection on a custom mcp_oauth entry whose association was deactivated before consent completed", async () => {
     // list_mcp_apps' local loop does not filter is_active, and emits auth_type
     // only while the association *is* active — so a deactivated mcp_oauth
-    // server is listed as is_connected: false with no auth_type. The runtime's
-    // server query drops inactive associations outright, so attaching one would
-    // silently load zero tools; it needs re-enabling, not re-authorization, and
-    // must keep its card-click route to the detail modal.
+    // server with no completed grant is listed as is_connected: false with no
+    // auth_type. (toggle_mcp_server never revokes a completed grant, so this
+    // only holds pre-consent: deactivating *after* consent instead lists
+    // is_connected: true and is attachable through the first disjunct, with
+    // the checkmark and Configure — that population is exercised elsewhere.)
+    // The runtime's server query drops inactive associations outright, so
+    // attaching this one would silently load zero tools — hence it must keep
+    // its card-click route to the detail modal rather than becoming
+    // selectable. That route offers no recovery today either; see the
+    // isAttachable comment.
     const onConnectSelected = vi.fn()
     const dormant: Record<string, unknown> = { ...unauthorizedCustomMcp() }
     delete dormant.auth_type
@@ -1819,12 +1852,7 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
   it("opens the detail modal on card click outside select mode", async () => {
     // The Tools page has no selection to toggle, so the card click keeps its
     // original destination for connected and unconnected entries alike.
-    apiRequestMock.mockImplementation((url: string) => {
-      if (url.includes("/api/mcp/apps?")) {
-        return Promise.resolve({ ok: true, json: async () => [unauthorizedCustomMcp()] })
-      }
-      throw new Error(`Unexpected request: ${url}`)
-    })
+    mockAppsList([unauthorizedCustomMcp()])
     render(<ConnectMcpDialog open onOpenChange={vi.fn()} selectedMcpServers={selectedMcpServers} />)
     await screen.findByText("Records MCP")
 
--- a/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
@@ -123,11 +123,15 @@ vi.mock("./custom-mcp-form", () => ({
 
 vi.mock("./official-mcp-settings-dialog", () => ({
   OfficialMcpSettingsDialog: ({
+    app,
+    open,
     isConnecting,
     onConfigure,
     onOpenChange,
     onConnectStart,
   }: {
+    app?: { name?: string } | null
+    open?: boolean
     isConnecting?: boolean
     onConfigure: (app: object) => void
     onOpenChange: (open: boolean) => void
@@ -140,6 +144,10 @@ vi.mock("./official-mcp-settings-dialog", () => ({
       <div data-testid="settings-is-connecting">
         {isConnecting ? "connecting" : "idle"}
       </div>
+      {/* Which app the detail modal is open for (empty when closed). The mock
+          renders unconditionally, so `open` is the only way a test can tell
+          whether a card click routed to the modal or toggled the selection. */}
+      <div data-testid="settings-open-app">{open ? app?.name ?? "" : ""}</div>
       <button
         type="button"
         onClick={() => onConfigure(customApiApp(1, "aggregated-a"))}
@@ -1665,5 +1673,122 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
 
     fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
     expect(onConnectSelected).toHaveBeenCalledWith([])
+  })
+
+  // A custom mcp_oauth connector whose access tokens are supplied out of band
+  // through set_oauth_token_resolver_hook: is_connected stays false forever
+  // because no MCPOAuthGrant is ever written for the editor, so gating the
+  // select-mode toggle on it made the connector permanently unattachable —
+  // even though the agent endpoints accept its selector with no
+  // connected-state check and the runtime resolves credentials without a
+  // grant (#1332).
+  const unauthorizedCustomMcp = () => ({ ...customMcpOauthApp(), name: "Records MCP" })
+
+  function renderSelectModeWith(apps: object[], onConnectSelected: () => void) {
+    apiRequestMock.mockImplementation((url: string) => {
+      if (url.includes("/api/mcp/apps?")) {
+        return Promise.resolve({ ok: true, json: async () => apps })
+      }
+      throw new Error(`Unexpected request: ${url}`)
+    })
+    return render(
+      <ConnectMcpDialog
+        open
+        onOpenChange={vi.fn()}
+        selectedMcpServers={selectedMcpServers}
+        onConnectSelected={onConnectSelected}
+      />,
+    )
+  }
+
+  it("selects and deselects a custom mcp_oauth connector the editor holds no grant for (#1332)", async () => {
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([unauthorizedCustomMcp()], onConnectSelected)
+    await screen.findByText("Records MCP")
+
+    fireEvent.click(screen.getByText("Records MCP"))
+    // The click must toggle the selection, not divert to the detail modal.
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("")
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith(["Records MCP"])
+
+    // Removal has to work too, or an accidental attach would be unrevertable.
+    fireEvent.click(screen.getByText("Records MCP"))
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("keeps the per-server OAuth flow reachable from the card in select mode (#1323)", async () => {
+    // The card click no longer opens the detail modal for these entries, so
+    // the Connect button repaired in #1323 needs its own trigger — otherwise
+    // relaxing the gate would strip the interactive flow from editors who do
+    // authorize personally.
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([unauthorizedCustomMcp()], onConnectSelected)
+    await screen.findByText("Records MCP")
+
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.authorize" }))
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("Records MCP")
+
+    // Opening the modal must not have counted as a selection toggle.
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("still gates selection on connected state for an unconnected catalog app", async () => {
+    // Only local/custom entries are relaxed. A catalog app reporting
+    // is_connected: false has no association row for this user at all, so
+    // connecting is a genuine prerequisite and the card must keep routing to
+    // the detail modal instead of attaching a connector that does not exist
+    // for them.
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([keylessApp()], onConnectSelected)
+    await screen.findByText("Chrome")
+
+    fireEvent.click(screen.getByText("Chrome"))
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("Chrome")
+    // No card-level Authorize trigger either: that is the mcp_oauth shape only.
+    expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
+
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("still gates selection on a custom mcp_oauth entry whose association is deactivated", async () => {
+    // list_mcp_apps' local loop does not filter is_active, and emits auth_type
+    // only while the association *is* active — so a deactivated mcp_oauth
+    // server is listed as is_connected: false with no auth_type. The runtime's
+    // server query drops inactive associations outright, so attaching one would
+    // silently load zero tools; it needs re-enabling, not re-authorization, and
+    // must keep its card-click route to the detail modal.
+    const onConnectSelected = vi.fn()
+    const dormant: Record<string, unknown> = { ...unauthorizedCustomMcp() }
+    delete dormant.auth_type
+    renderSelectModeWith([dormant], onConnectSelected)
+    await screen.findByText("Records MCP")
+
+    fireEvent.click(screen.getByText("Records MCP"))
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("Records MCP")
+    expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
+
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("opens the detail modal on card click outside select mode", async () => {
+    // The Tools page has no selection to toggle, so the card click keeps its
+    // original destination for connected and unconnected entries alike.
+    apiRequestMock.mockImplementation((url: string) => {
+      if (url.includes("/api/mcp/apps?")) {
+        return Promise.resolve({ ok: true, json: async () => [unauthorizedCustomMcp()] })
+      }
+      throw new Error(`Unexpected request: ${url}`)
+    })
+    render(<ConnectMcpDialog open onOpenChange={vi.fn()} selectedMcpServers={selectedMcpServers} />)
+    await screen.findByText("Records MCP")
+
+    expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
+    fireEvent.click(screen.getByText("Records MCP"))
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("Records MCP")
   })
 })
--- a/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
@@ -1342,6 +1342,18 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
       })
 
       expect(onSuccess).toHaveBeenCalled()
+
+      // Commit the selection to observe the auto-select this test is named
+      // for: the append is local state, so onConnectSelected only sees it
+      // through the footer. Granola is the fixture that makes the assertion
+      // discriminating — its id ("granola") and display name ("Granola")
+      // differ, so this fails if the append ever switches to app.id, whereas
+      // the custom-server equivalent below cannot tell the two apart (its id
+      // and name are both "records").
+      act(() => {
+        fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+      })
+      expect(onConnectSelected).toHaveBeenCalledWith(["Granola"])
     } finally {
       openSpy.mockRestore()
       vi.useRealTimers()
@@ -1767,14 +1779,13 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     renderSelectModeWith([{ ...unauthorizedCustomMcp(), is_connected: true }], onConnectSelected)
     await screen.findByText("Records MCP")
 
-    expect(screen.getByRole("button", { name: "tools.mcp.dialog.configure" })).not.toBeNull()
+    // getBy* throws when absent, so the bare calls are the assertions.
+    screen.getByRole("button", { name: "tools.mcp.dialog.configure" })
     expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
     // Scoped through the card: connected-check is non-unique across the grid
     // by construction, so an ungrounded query would throw as soon as a
     // second connected app rendered.
-    expect(
-      within(screen.getByTestId("connector-card-records")).getByTestId("connected-check"),
-    ).not.toBeNull()
+    within(screen.getByTestId("connector-card-records")).getByTestId("connected-check")
   })
 
   it("keeps the per-server OAuth flow reachable from the card in select mode (#1323)", async () => {
@@ -1829,7 +1840,11 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     // auth_type. (toggle_mcp_server never revokes a completed grant, so this
     // only holds pre-consent: deactivating *after* consent instead lists
     // is_connected: true and is attachable through the first disjunct, with
-    // the checkmark and Configure — that population is exercised elsewhere.)
+    // the checkmark and Configure. No test renders that exact shape — it is
+    // indistinguishable here from an ordinary connected entry, and the
+    // regression it would guard against, replacing the first disjunct with
+    // one that also demands auth_type, already fails the seeded keyless
+    // test below.)
     // The runtime's server query drops inactive associations outright, so
     // attaching this one would silently load zero tools — hence it must keep
     // its card-click route to the detail modal rather than becoming
--- a/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.test.tsx
@@ -578,14 +578,20 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     }
   })
 
-  it("rechecks a closed custom mcp_oauth popup against the local listing", async () => {
+  it("rechecks a closed custom mcp_oauth popup against the local listing and auto-selects it", async () => {
     // The popup's opener link is severed, so a closed popup is ambiguous and
     // the handler asks the backend what actually happened. A custom server
     // only exists under location=local — querying the remote branch (as the
     // catalog path does) would report every custom connect as a failure.
+    //
+    // Passing onConnectSelected also puts the dialog in select mode, which is
+    // what makes the connect-records trigger's autoSelect true — so a
+    // recheck that confirms the connection must also land "records" in the
+    // committed selection (#1332).
     const popup = { closed: false, close: vi.fn(), opener: {}, location: { href: "" } }
     const openSpy = vi.spyOn(window, "open").mockReturnValue(popup as unknown as Window)
     const onSuccess = vi.fn()
+    const onConnectSelected = vi.fn()
     let localListCalls = 0
     try {
       apiRequestMock.mockImplementation((url: string) => {
@@ -614,6 +620,7 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
           onOpenChange={vi.fn()}
           selectedMcpServers={selectedMcpServers}
           onSuccess={onSuccess}
+          onConnectSelected={onConnectSelected}
         />,
       )
 
@@ -632,6 +639,14 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
 
       expect(localListCalls).toBe(1)
       expect(onSuccess).toHaveBeenCalled()
+
+      // Commit the selection to observe it. Fake timers are still active, so
+      // this must stay on synchronous queries (getByRole) — findBy*/waitFor
+      // would hang.
+      act(() => {
+        fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+      })
+      expect(onConnectSelected).toHaveBeenCalledWith(["records"])
     } finally {
       openSpy.mockRestore()
       vi.useRealTimers()
@@ -1709,6 +1724,10 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     fireEvent.click(screen.getByText("Records MCP"))
     // The click must toggle the selection, not divert to the detail modal.
     expect(screen.getByTestId("settings-open-app").textContent).toBe("")
+    // Becoming attachable must not change connected-state display: this
+    // connector is still unconnected (is_connected: false), so it must stay
+    // unchecked.
+    expect(screen.queryByTestId("connected-check")).toBeNull()
     fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
     expect(onConnectSelected).toHaveBeenLastCalledWith(["Records MCP"])
 
@@ -1716,6 +1735,19 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     fireEvent.click(screen.getByText("Records MCP"))
     fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
     expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("shows Configure and the connected checkmark, not Authorize, for a connected custom mcp_oauth entry (#1332)", async () => {
+    // isAttachable widening the select-mode gate must not disturb the
+    // isGloballyConnected branch of the footer ternary: a connected entry
+    // has to keep showing Configure plus the checkmark, never Authorize.
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([{ ...unauthorizedCustomMcp(), is_connected: true }], onConnectSelected)
+    await screen.findByText("Records MCP")
+
+    expect(screen.getByRole("button", { name: "tools.mcp.dialog.configure" })).not.toBeNull()
+    expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
+    expect(screen.getByTestId("connected-check")).not.toBeNull()
   })
 
   it("keeps the per-server OAuth flow reachable from the card in select mode (#1323)", async () => {
@@ -1735,19 +1767,28 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
     expect(onConnectSelected).toHaveBeenLastCalledWith([])
   })
 
-  it("still gates selection on connected state for an unconnected catalog app", async () => {
-    // Only local/custom entries are relaxed. A catalog app reporting
-    // is_connected: false has no association row for this user at all, so
-    // connecting is a genuine prerequisite and the card must keep routing to
-    // the detail modal instead of attaching a connector that does not exist
-    // for them.
+  it("still gates selection on connected state for unconnected catalog apps", async () => {
+    // Only local/custom entries are relaxed. mcpOauthApp() (Granola) is the
+    // load-bearing fixture here: it carries the exact same auth_type as the
+    // custom mcp_oauth population this predicate widens, but it is a
+    // *catalog* entry, so its is_connected: false means the user has no
+    // association row for it at all — connecting really is a prerequisite,
+    // and dropping the is_custom conjunct would wrongly attach it. keylessApp
+    // (Chrome) is included alongside it as a non-mcp_oauth catalog control.
+    // Both must keep routing the card click to the detail modal instead of
+    // attaching a connector that does not exist for them.
     const onConnectSelected = vi.fn()
-    renderSelectModeWith([keylessApp()], onConnectSelected)
+    renderSelectModeWith([mcpOauthApp(), keylessApp()], onConnectSelected)
     await screen.findByText("Chrome")
+
+    fireEvent.click(screen.getByText("Granola"))
+    // Assert right after each click: a later click would overwrite this.
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("Granola")
 
     fireEvent.click(screen.getByText("Chrome"))
     expect(screen.getByTestId("settings-open-app").textContent).toBe("Chrome")
-    // No card-level Authorize trigger either: that is the mcp_oauth shape only.
+
+    // No card-level Authorize trigger for either: is_custom is required too.
     expect(screen.queryByRole("button", { name: "tools.mcp.dialog.authorize" })).toBeNull()
 
     fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -216,6 +216,14 @@ export function ConnectMcpDialog({
   // agent create/update endpoints accept its `mcp:<name>` selector with no
   // connected-state check and the runtime resolves credentials without a grant.
   //
+  // A custom mcp_oauth server whose consent was simply never started, or was
+  // abandoned partway, is indistinguishable in this listing from a
+  // hook-resolved one — create_mcp_server writes the association row at
+  // creation time, long before any consent, so is_connected is false either
+  // way. This predicate makes that server attachable too: the card shows no
+  // checkmark and offers Authorize as a one-click recovery, and the agent
+  // create/update endpoints never validated connected state to begin with.
+  //
   // Both conjuncts are load-bearing, and neither can be widened:
   //
   // - is_custom, because a *catalog* mcp_oauth app (e.g. Granola) carries the
@@ -224,11 +232,20 @@ export function ConnectMcpDialog({
   //   attaching it would select a connector that does not exist for them.
   // - auth_type, because the backend emits it on a local entry only for the
   //   mcp_oauth shape *and only while the association is active* (see
-  //   list_mcp_apps). A deactivated association is still listed with
-  //   is_connected: false, and the runtime's server query drops inactive
-  //   associations outright — attaching one would silently load zero tools.
-  //   That entry keeps its existing card-click route to the detail modal, where
-  //   it needs re-enabling rather than re-authorization.
+  //   list_mcp_apps). An association deactivated before any grant completed is
+  //   listed with is_connected: false and no auth_type, so it keeps its
+  //   existing card-click route to the detail modal, where it needs
+  //   re-enabling rather than re-authorization. That matters because the
+  //   runtime's server query drops inactive associations outright, so
+  //   attaching one would have loaded zero tools.
+  //
+  //   Deactivation *after* consent is a different, pre-existing story:
+  //   toggle_mcp_server only flips is_active and never revokes the grant, so
+  //   such an entry lists as is_connected: true and is attachable through the
+  //   first conjunct regardless of this one — showing the checkmark plus
+  //   Configure, never Authorize. The runtime still drops it, but the old gate
+  //   was is_connected alone, so that card behaves here exactly as it did
+  //   before this predicate existed.
   const isAttachable = (app: AppIntegration) =>
     isAppConnected(app) || (Boolean(app.is_custom) && app.auth_type === "mcp_oauth")
 
@@ -1301,7 +1318,7 @@ export function ConnectMcpDialog({
                       return (
                         <Card key={app.id} className={`p-[0] cursor-pointer transition-colors shadow-sm relative ${isSelectMode && isSelected ? 'border-blue-500 bg-blue-50/30 ring-1 ring-blue-500' : 'hover:border-slate-300 border-slate-200'}`} onClick={() => handleCardClick(app)}>
                           {isGloballyConnected && (
-                            <div className="absolute top-4 right-4 text-green-500">
+                            <div data-testid="connected-check" className="absolute top-4 right-4 text-green-500">
                               <CheckCircle2 className="h-5 w-5 fill-green-100" />
                             </div>
                           )}
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -232,10 +232,14 @@ export function ConnectMcpDialog({
   //   attaching it would select a connector that does not exist for them.
   // - auth_type, because the backend emits it on a local entry only when a
   //   personal association exists, is active, and the server is the
-  //   mcp_oauth shape (see list_mcp_apps). One consequence: a team-owned
-  //   connector surfaced by #1338's visibility overlay has no personal
-  //   association, so it gets no auth_type either, and this predicate routes
-  //   it to the detail modal instead of making it attachable.
+  //   mcp_oauth shape (see list_mcp_apps). One consequence, once #1338's team
+  //   visibility overlay is in the tree (it is on main, but landed after this
+  //   branch was cut, so the population is not reproducible from this
+  //   branch's own backend): a team-owned connector the overlay surfaces has
+  //   no personal association, so it gets no auth_type either, and this
+  //   predicate routes it to the detail modal instead of making it
+  //   attachable. #1347 tracks expressing that case as attachable-but-not-
+  //   authorizable rather than deciding it by an absent hint.
   //
   //   An association deactivated before any grant completed is likewise
   //   listed with is_connected: false and no auth_type, so it keeps its
--- a/frontend/src/components/mcp/connect-mcp-dialog.tsx
+++ b/frontend/src/components/mcp/connect-mcp-dialog.tsx
@@ -204,6 +204,34 @@ export function ConnectMcpDialog({
 
   const isAppConnected = (app: AppIntegration) => Boolean(app.is_connected)
 
+  // Whether select mode may toggle this entry into the agent's selection.
+  //
+  // #1332: is_connected must not gate this for a custom mcp_oauth connector.
+  // is_connected is false for one there precisely when *this editor* holds no
+  // completed MCPOAuthGrant — but an embedding application can supply those
+  // access tokens out of band via set_oauth_token_resolver_hook (per end user,
+  // provisioned server-to-server), and then there is no interactive consent, no
+  // identity the editor could sign in as, and therefore never a grant row. The
+  // connector was permanently unattachable from the picker even though the
+  // agent create/update endpoints accept its `mcp:<name>` selector with no
+  // connected-state check and the runtime resolves credentials without a grant.
+  //
+  // Both conjuncts are load-bearing, and neither can be widened:
+  //
+  // - is_custom, because a *catalog* mcp_oauth app (e.g. Granola) carries the
+  //   same auth_type while is_connected: false there means the user has no
+  //   association row for it at all. Connecting really is a prerequisite, so
+  //   attaching it would select a connector that does not exist for them.
+  // - auth_type, because the backend emits it on a local entry only for the
+  //   mcp_oauth shape *and only while the association is active* (see
+  //   list_mcp_apps). A deactivated association is still listed with
+  //   is_connected: false, and the runtime's server query drops inactive
+  //   associations outright — attaching one would silently load zero tools.
+  //   That entry keeps its existing card-click route to the detail modal, where
+  //   it needs re-enabling rather than re-authorization.
+  const isAttachable = (app: AppIntegration) =>
+    isAppConnected(app) || (Boolean(app.is_custom) && app.auth_type === "mcp_oauth")
+
   useEffect(() => {
     const timer = setTimeout(() => setDebouncedSearch(searchQuery), 300)
     return () => clearTimeout(timer)
@@ -952,8 +980,8 @@ export function ConnectMcpDialog({
     mcpOauthPollTimersRef.current.add(checkPopup)
   }
 
-  const handleCardClick = (app: AppIntegration, isGloballyConnected: boolean) => {
-    if (isSelectMode && isGloballyConnected) {
+  const handleCardClick = (app: AppIntegration) => {
+    if (isSelectMode && isAttachable(app)) {
       // Match isSelected's own check (id OR name) below, not just name: a
       // consumer (e.g. agent-builder.tsx re-seeding after save) can store
       // this connector's id instead of its display name once resolved to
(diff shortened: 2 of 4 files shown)


rogercloud (commented):
## Summary

This PR fixes a permanent unattachability bug in the "+ Connector" picker's select mode: the selection gate keyed on `is_connected`, which is false for a custom `mcp_oauth` connector precisely when this editor holds no completed `MCPOAuthGrant` — but an embedding application can supply those tokens out of band via `set_oauth_token_resolver_hook`, so no grant row ever exists and the connector could never be attached, even though the agent create/update endpoints accept the `mcp:<name>` selector with no connected-state check. The PR relaxes the gate to an `isAttachable` predicate (`isAppConnected(app) || (Boolean(app.is_custom) && app.auth_type === "mcp_oauth")`) and adds a card-level "Authorize" trigger keyed on the same predicate, so the per-server OAuth flow repaired in #1323 stays reachable for entries whose card click was diverted from the detail modal to selection. Five new tests cover selection/deselection, the retained OAuth route, and the still-gated catalog/deactivated/outside-select-mode cases.

## Update summary

No commits have been pushed since the previous (automated) review, and that review contained no findings, so there is nothing to re-verify from it; this pass re-reviewed the full diff from a blank slate.

## Approach

Sound. The predicate is minimal, both conjuncts are load-bearing and documented, and the Authorize trigger is keyed on exactly the set of cards whose click was diverted to selection.

Design notes:

- The Authorize trigger is a dead end in hook-only deployments. The UI cannot distinguish hook-resolved credentials from interactive-capable ones — no backend signal exists — and #1332 itself states that for hook-resolved connectors there is no interactive consent and no identity to sign in as. Clicking Authorize opens the settings dialog whose Connect action starts an OAuth popup that cannot complete. This is a pre-existing route (the old card click had the same dead end), so it is no regression and strictly better than before, but it is a known tradeoff.
- The predicate re-derives backend emission policy in the frontend (`auth_type` is emitted only for active mcp_oauth-shaped associations; `is_connected` requires an active grant). A backend-emitted attachability field (active association AND (active grant OR token resolver hook present)) would remove the duplicated policy and also fix the distinguishability problem behind findings 1 and 2. Non-blocking follow-up.

## Findings (all minor)

### 1. `isAttachable` un-gates "abandoned-consent" connectors — minor

frontend/src/components/mcp/connect-mcp-dialog.tsx:232-233

The association row is created before consent completes (`connect_mcp_oauth_app` persists it up front), so an entry with no grant and no hook becomes attachable and yields zero tools at runtime ("MCP server credentials are unavailable") instead of a blocked attach. The backend never validated connected state on the agent endpoints (the gate was UI-only per #1332), the card shows no checkmark and offers the Authorize button as one-click recovery for exactly this population, and the runtime failure is a clear diagnostic — so the frontend-only fix is acceptable. Recommended follow-up: a backend-emitted `can_attach` flag (active association AND (active grant OR token resolver hook present)).

### 2. Authorize button is a dead end for hook-resolved connectors — minor

frontend/src/components/mcp/connect-mcp-dialog.tsx:1368-1394

The button also renders for the PR's target population (hook-resolved connectors), where #1332 states there is no interactive consent and no identity to sign in as: clicking it opens the settings dialog whose Connect action starts an OAuth popup that cannot complete. Pre-existing route, no regression, strictly better than before — but the button actively miscommunicates "needs authorization" in hook deployments. No backend signal exists to suppress it; flag as a known tradeoff.

### 3. Comment overclaims deactivated-association protection — minor

frontend/src/components/mcp/connect-mcp-dialog.tsx:219-231

"A deactivated association is still listed with is_connected: false" holds only when the user also has no active grant. `toggle_mcp_server` flips only `UserMCPServer.is_active` and never revokes `MCPOAuthGrant` rows; `_mcp_oauth_server_is_actually_connected` checks only active grants. So a server deactivated after completed consent is listed `is_connected: true` with no `auth_type`, stays attachable via the first conjunct, and the runtime's `is_active` filter drops it (zero tools). Pre-existing backend behavior, no regression; the new comment and test pin only the no-grant variant. Suggest qualifying the comment to the no-grant case.

### 4. Test fixture does not exercise the load-bearing conjunct — minor

frontend/src/components/mcp/connect-mcp-dialog.test.tsx:1745

The "still gates selection on connected state for an unconnected catalog app" test uses `keylessApp()` (auth_type "keyless"), not `mcpOauthApp()` (auth_type "mcp_oauth" without `is_custom` — the Granola shape). The `is_custom` conjunct exists precisely because catalog mcp_oauth apps carry the same auth_type; under a broken predicate (`is_connected || auth_type === "mcp_oauth"`) the keyless fixture stays gated and the test still passes. Swap the fixture so the test fails if the `is_custom` conjunct is removed:

```ts
renderSelectModeWith([mcpOauthApp()], onConnectSelected)
// and update the queries from "Chrome" to "Granola"
```

### 5. No assertion on the consent flow's outcome — minor

frontend/src/components/mcp/connect-mcp-dialog.test.tsx:1721-1736 (and pre-existing test at :1283-1334)

On consent success `handleConnectMcpOAuthApp` appends the connector name to the selection (autoSelect wiring at connect-mcp-dialog.tsx:1567, append at :883-885); the pre-existing popup test "fires onSuccess and auto-selects only once the closed popup is confirmed connected" only asserts `onSuccess`. Dropping the append or the `isSelectMode` wiring would pass the entire suite. Suggest extending this test to assert `onConnectSelected` receives "Records MCP" after a successful consent.

### 6. Configure-vs-Authorize exclusion and checkmark not pinned — minor

frontend/src/components/mcp/connect-mcp-dialog.tsx:1356-1395

No test pins the mutual exclusion or the connected-state display: a connected custom mcp_oauth entry in select mode should show Configure (not Authorize) plus the checkmark; a ternary reorder would render Authorize on connected cards unnoticed, and no test asserts checkmark presence/absence. Suggest one test: connected custom mcp_oauth in select mode → Configure present, Authorize absent, checkmark present; plus a checkmark-absence assertion in the existing #1332 select/deselect test.

### 7. PR description overstates test failure behavior — minor (body only)

The Tests section claims "each confirmed to fail without the corresponding change" for all five tests; only tests 1-2 fail on the old code. Tests 3-5 are regression guards that pass on old behavior. Suggest rewording the description.

Blocking: no. All confirmed findings are minor (test hardening and comment/reporting accuracy); the fix itself is sound. Recommended event: APPROVE.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -204,6 +204,34 @@ export function ConnectMcpDialog({
...
+  //   association row for it at all. Connecting really is a prerequisite, so
+  //   attaching it would select a connector that does not exist for them.
+  // - auth_type, because the backend emits it on a local entry only for the
+  //   mcp_oauth shape *and only while the association is active* (see
+  //   list_mcp_apps). A deactivated association is still listed with
+  //   is_connected: false, and the runtime's server query drops inactive
+  //   associations outright — attaching one would silently load zero tools.
+  //   That entry keeps its existing card-click route to the detail modal, where
+  //   it needs re-enabling rather than re-authorization.
+  const isAttachable = (app: AppIntegration) =>
```
rogercloud, line 264: `isAttachable` also un-gates "abandoned-consent" connectors: the association row is created before consent completes (`connect_mcp_oauth_app` persists it up front), so an entry with no grant and no token-resolver hook becomes attachable and yields zero tools at runtime ("MCP server credentials are unavailable") instead of a blocked attach. The backend never validated connected state on the agent endpoints, the card shows no checkmark and offers the Authorize button as one-click recovery for exactly this population, and the runtime failure is a clear diagnostic — so the frontend-only fix is acceptable. Recommended follow-up: a backend-emitted `can_attach` flag (active association AND (active grant OR token resolver hook present)).

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -1337,7 +1365,34 @@ export function ConnectMcpDialog({
...
+                                  // dropping it, for editors who do authorize
+                                  // personally.
+                                  //
+                                  // Keyed on isAttachable, not its own
+                                  // predicate, so this stays exactly the set of
+                                  // cards whose click was diverted to selection:
+                                  // every entry that loses the modal route gets
+                                  // this trigger, and no entry that kept the
+                                  // route gets a redundant second one.
+                                  <Button
```
rogercloud, line 1422: The Authorize button also renders for hook-resolved connectors (the target population of this PR), where #1332 states there is no interactive consent and no identity to sign in as — clicking it opens the settings dialog whose Connect action starts an OAuth popup that cannot complete. This is a pre-existing dead end (the old card click had the same route), so no regression and strictly better than before, but the button miscommunicates "needs authorization" in hook deployments. No backend signal exists to suppress it; flag as a known tradeoff.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -204,6 +204,34 @@ export function ConnectMcpDialog({
...
+  // agent create/update endpoints accept its `mcp:<name>` selector with no
+  // connected-state check and the runtime resolves credentials without a grant.
+  //
+  // Both conjuncts are load-bearing, and neither can be widened:
+  //
+  // - is_custom, because a *catalog* mcp_oauth app (e.g. Granola) carries the
+  //   same auth_type while is_connected: false there means the user has no
+  //   association row for it at all. Connecting really is a prerequisite, so
+  //   attaching it would select a connector that does not exist for them.
+  // - auth_type, because the backend emits it on a local entry only for the
```
rogercloud: This comment overclaims the deactivated-association protection: "A deactivated association is still listed with is_connected: false" holds only when the user also has no active grant. `toggle_mcp_server` flips only `UserMCPServer.is_active` and never revokes `MCPOAuthGrant` rows, and `_mcp_oauth_server_is_actually_connected` checks only active grants — so a server deactivated after completed consent is listed `is_connected: true` with no `auth_type`, stays attachable via the first conjunct, and the runtime's `is_active` filter drops it (zero tools). Pre-existing backend behavior, no regression; suggest qualifying the claim to the no-grant case.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -1666,4 +1674,121 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
...
+  })
+
+  it("still gates selection on connected state for an unconnected catalog app", async () => {
+    // Only local/custom entries are relaxed. A catalog app reporting
+    // is_connected: false has no association row for this user at all, so
+    // connecting is a genuine prerequisite and the card must keep routing to
+    // the detail modal instead of attaching a connector that does not exist
+    // for them.
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([keylessApp()], onConnectSelected)
```
rogercloud: This test uses `keylessApp()` (auth_type "keyless"), not `mcpOauthApp()` (auth_type "mcp_oauth" without `is_custom` — the Granola shape). The load-bearing `is_custom` conjunct exists precisely because catalog mcp_oauth apps carry the same auth_type; under a broken predicate (`is_connected || auth_type === "mcp_oauth"`) the keyless fixture stays gated and this test still passes. Swap the fixture to `mcpOauthApp()` (and update the name queries from "Chrome" to "Granola") so the test fails if the `is_custom` conjunct is removed.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -1666,4 +1674,121 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
...
+    // authorize personally.
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([unauthorizedCustomMcp()], onConnectSelected)
+    await screen.findByText("Records MCP")
+
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.authorize" }))
+    expect(screen.getByTestId("settings-open-app").textContent).toBe("Records MCP")
+
+    // Opening the modal must not have counted as a selection toggle.
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
```
rogercloud, line 1804: Nothing asserts the flow's outcome here: on consent success `handleConnectMcpOAuthApp` appends the connector name to the selection (autoSelect wiring at connect-mcp-dialog.tsx:1567, append at :883-885). The pre-existing popup test ("fires onSuccess and auto-selects only once the closed popup is confirmed connected", :1283-1334) only asserts `onSuccess`, so dropping the append or the `isSelectMode` wiring would pass the entire suite. Suggest asserting that `onConnectSelected` receives "Records MCP" after a successful consent.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -1337,7 +1365,34 @@ export function ConnectMcpDialog({
                                   >
                                     <Settings className="h-3 w-3 mr-1" /> {t('tools.mcp.dialog.configure')}
                                   </Button>
-                                )}
+                                ) : isSelectMode && isAttachable(app) ? (
```
rogercloud, line 1406: No test pins the Configure-vs-Authorize mutual exclusion or the connected-state display: a connected custom mcp_oauth entry in select mode should show Configure (not Authorize) plus the checkmark; a ternary reorder would render Authorize on connected cards unnoticed, and no test asserts checkmark presence/absence. Suggest one test: connected custom mcp_oauth in select mode → Configure present, Authorize absent, checkmark present; plus a checkmark-absence assertion in the existing #1332 select/deselect test.

codeacme17:
Thanks — this was a useful review. Findings 3–7 are all correct and are addressed in f4eedbc; findings 1 and 2 I agree are known tradeoffs and I have left the code alone. One correction on finding 1's mechanism, and one interaction with #1338 that is worth recording.

I verified every claim independently before acting, and mutation-tested each test change.

## Correction: finding 1 names the wrong endpoint

The finding cites `connect_mcp_oauth_app` as the source of grantless-but-attachable rows. That is the **catalog** path, and catalog entries never carry `is_custom` — that field is set only in the local loop and the Custom API loop. So an abandoned *catalog* consent lists as `{is_connected: false, auth_type: "mcp_oauth", is_custom: undefined}` and `isAttachable` returns false: it stays gated. Finding 4's fixture swap now pins exactly that.

The real source is `create_mcp_server`, which writes the `UserMCPServer` row with `is_active=server_data.is_active` (default true) at creation time, long before any consent is started. A user-added custom mcp_oauth server whose OAuth was never begun is byte-identical in this listing to the hook-resolved shape the PR targets, so it is attachable too.

Your disposition still holds, for the reason you give: the agent endpoints never validated connected state, the card shows no checkmark, and Authorize is one-click recovery for exactly that population. The comment now records this with the correct citation. No code change.

## Finding 2: agreed, and "no backend signal" confirmed

`set_oauth_token_resolver_hook` is a process-global in-memory hook (`src/xagent/web/tools/config.py`), never persisted and never surfaced by any route, so nothing in `list_mcp_apps` can distinguish a hook-resolved connector from an interactive-capable one. The label overclaims in hook-only deployments; the route itself is unchanged from before this PR. Left as-is, and it is the second half of the case for the backend `can_attach` flag you propose.

## Findings 3–6: fixed, mutation-verified

Each of 4, 5 and 6 was a test that pinned nothing — I reproduced all three before fixing them. Mutations applied, run, and reverted:

| Mutation | Before f4eedbc | After |
| --- | --- | --- |
| drop `isAttachable`'s `is_custom` conjunct | 32/32 green | catalog-gating test fails |
| drop the `auth_type` conjunct | deactivated-entry test fails | unchanged |
| delete the auto-select append in `handleConnectMcpOAuthApp` | 32/32 green | popup-recheck test fails |
| swap the two card-footer branches wholesale | 32/32 green | connected-entry test fails |

Finding 3's comment is corrected as you describe: `toggle_mcp_server` flips only `is_active` and never revokes `MCPOAuthGrant`, and `_mcp_oauth_server_is_actually_connected` ignores `is_active`, so an association deactivated *after* consent lists as `is_connected: true` and stays attachable through the first conjunct regardless of the second. The comment now qualifies the claim to the no-grant case and marks the after-consent case as pre-existing (the old gate was `is_connected` alone), so that card's behavior here is unchanged.

Finding 6 needed one production change: `data-testid="connected-check"` on the checkmark wrapper, so the display assertions can name it instead of matching on layout classes.

## Finding 7: you are right, and the description is rewritten

Only two of the five went red against the pre-change component; I had verified those two individually and generalised the claim, which was wrong. The Tests section now separates the two red tests from the regression guards and states, per guard, which mutation it kills.

## Interaction with #1338, now merged

#1338 deliberately withholds `auth_type` from team-owned rows, because `/{server_id}/oauth/connect` requires an active personal association and would 404. That means a team-owned **mcp_oauth** connector is listed with `is_custom: true`, `is_connected: false` and no `auth_type`, so `isAttachable` gates it and this PR does not make it attachable.

That is not a regression — the previous gate was `is_connected` alone, which excluded it identically — and it is gated for the same reason #1338 withholds the hint. Team-owned non-oauth MCP servers and Custom APIs report `is_connected: true` and were already attachable after #1338. Recording it so the gap is not mistaken for something either PR closed.

## Follow-ups filed separately

Filing the `can_attach` proposal, and a distinct listing bug I hit while verifying finding 1: the local loop dedupes against `library_names` keyed on the catalog app **name**, while a catalog app's shared server row is named after its **app_id**. For any catalog app whose `app_id` and `name` differ case-insensitively, the association escapes dedupe and is emitted a second time as `is_custom: true`. Latent today for the mcp_oauth shape (`granola`/"Granola" and `notion`/"Notion" both match), but admins can rename apps and ten non-mcp_oauth entries already mismatch — and if it were reachable for an mcp_oauth app, `isAttachable` would widen the blast radius, so it is worth closing on its own.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Outcome confirmed, mechanism corrected — no code change.

`connect_mcp_oauth_app` is the **catalog** path, and catalog entries never carry `is_custom` (it is set only in the local loop and the Custom API loop). An abandoned catalog consent therefore lists as `{is_connected: false, auth_type: "mcp_oauth", is_custom: undefined}` and `isAttachable` returns false — it stays gated. Finding 4's fixture swap now pins exactly that case.

The real source is `create_mcp_server`, which writes the `UserMCPServer` row with `is_active=server_data.is_active` (default true) at creation, long before any consent starts. So a user-added custom mcp_oauth server whose OAuth was never begun is byte-identical here to the hook-resolved shape this PR targets, and is attachable.

Your disposition holds for the reasons you give. f4eedbc records this in the `isAttachable` comment with the corrected citation, and I am filing the backend `can_attach` proposal separately.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Agreed, and "no backend signal exists" is confirmed: `set_oauth_token_resolver_hook` is a process-global in-memory hook in `src/xagent/web/tools/config.py`, never persisted and never surfaced by any route, so `list_mcp_apps` has nothing to derive a suppression flag from.

Leaving the button as-is. The route is unchanged from before this PR (the old card click reached the same dead end); only the label is new, and it does overclaim in hook-only deployments. Captured as the second half of the case for the `can_attach` / `hook_managed` flag in the follow-up issue.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Confirmed and fixed in f4eedbc.

Verified each step: `toggle_mcp_server` flips only `user_mcp.is_active` and never touches `MCPOAuthGrant`; `_mcp_oauth_server_is_actually_connected` keys on active grants only and ignores `is_active`; the local loop sets `is_connected` from it without an `is_active` filter while gating `auth_type` on the association being active. So an association deactivated after completed consent lists `is_connected: true` with no `auth_type`, is attachable through the first conjunct regardless of the second, and the runtime's `is_active` filter drops it.

The comment now qualifies the `is_connected: false` claim to the "deactivated before any grant completed" case, and states the after-consent case separately as pre-existing — the old gate was `is_connected` alone, so that card shows the checkmark plus Configure and behaves here exactly as before. The new deactivated-entry test covers the no-grant half and does kill the `auth_type`-conjunct mutation.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: Confirmed and fixed in f4eedbc — I reproduced it first: with `isAttachable` reduced to `isAppConnected(app) || app.auth_type === "mcp_oauth"`, all 32 tests still passed.

The test now renders `mcpOauthApp()` (Granola) alongside `keylessApp()` as a non-mcp_oauth control, asserts the card click opens the detail modal for each (checked right after each click, since a later click overwrites `settings-open-app`), that no Authorize button exists for either, and that the footer commits `[]`. With the `is_custom` conjunct dropped it now fails. The comment explains why the catalog mcp_oauth fixture is the load-bearing one.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: Confirmed and fixed in f4eedbc — deleting the append at `connect-mcp-dialog.tsx` left all 32 tests green, as you say.

Rather than adding a second fake-timer popup test, I extended the existing custom-server test ("rechecks a closed custom mcp_oauth popup against the local listing **and auto-selects it**"): it now passes `onConnectSelected`, which is also what puts the dialog in select mode and makes the trigger's `autoSelect` true, and after the recheck confirms the connection it clicks the footer to commit and asserts the selection is `["records"]`. Synchronous queries only, since fake timers are still active there. The `localListCalls === 1` and `onSuccess` assertions are untouched. Deleting the append now fails this test.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Confirmed and fixed in f4eedbc. I reproduced the exact mutation you describe: moving the Authorize branch ahead of the Configure branch wholesale left 32/32 green, so a connected custom mcp_oauth card would have silently shown Authorize.

Added the test you suggest — connected custom mcp_oauth entry in select mode → Configure present, Authorize absent, checkmark present — and the checkmark-absence assertion in the existing select/deselect test, which pins the "connected-state display is unchanged" acceptance criterion from the unconnected side.

This needed one production change: `data-testid="connected-check"` on the checkmark wrapper `div`, so both assertions name it instead of matching on layout classes. The branch-swap mutation now fails the new test.

codeacme17:
Both follow-ups are filed: #1346 (the local-branch dedupe key, a listing bug that stands on its own) and #1347 (the `can_attach` proposal, covering findings 1 and 2 and the `auth_type` overloading that #1338 had to work around).

rogercloud (commented):
# Follow-up review — PR #1335: fix(mcp): stop gating connector attach on the editor's own OAuth grant (#1332)

## Summary

The "+ Connector" picker's select-mode gate keyed on `is_connected` made custom `mcp_oauth` connectors with hook-supplied tokens permanently unattachable: the gate was UI-only, since the agent create/update endpoints accept `mcp:<name>` with no connected-state check and the runtime resolves credentials via `set_oauth_token_resolver_hook` without any grant. The PR relaxes the gate to `isAttachable = isAppConnected(app) || (Boolean(app.is_custom) && app.auth_type === "mcp_oauth")` and adds a card-level Authorize trigger keyed on the same predicate, so the per-server OAuth flow repaired in #1323 stays reachable for the diverted entries. Connected-state display is untouched.

## Update since the prior review

No commits have been pushed since the prior review: the head is unchanged at `f4eedbcf` and the author replied to every inline thread. The new development is external — PR #1338 (the team-visibility overlay for `/api/mcp/apps`) merged into main after the prior review, which makes the PR body's Scope claim inaccurate (N2 below). The author's tracking issue #1347 covers the follow-up work.

## Prior findings checklist

- **P1 — WAIVED**: the comment's mechanism citation was corrected to `create_mcp_server` (the association row is written at creation, long before any consent); the never-authorized-attachable consequence is tracked in #1347.
- **P2 — WAIVED**: the process-global token resolver is surfaced by no route, so there is no signal to suppress the Authorize trigger on; tracked in #1347 (consequence 2).
- **P3 — FIXED**: the comment now qualifies its claim to "deactivated before any grant completed" and marks the after-consent case as pre-existing.
- **P4 — FIXED**: the catalog test now renders `mcpOauthApp()` (Granola) alongside a `keylessApp()` control; dropping the `is_custom` conjunct fails three assertions.
- **P5 — FIXED**: the extended recheck test commits the footer and asserts the `["records"]` selection; deleting the auto-select append now fails.
- **P6 — FIXED**: a new test pins Configure-present / Authorize-absent / checkmark-present, backed by the added `data-testid="connected-check"`.
- **P7 — FIXED**: the Tests section now states exactly which two tests fail against the pre-change component and frames the rest as mutation-killing guards.

## Approach

The Round 0 verdict stands: acceptable-with-reservations. The predicate is minimal and both conjuncts are load-bearing (catalog entries must stay gated; inactive associations would attach zero tools), and the Authorize trigger is keyed exactly on the diverted set, so no card that kept its click route gains a redundant trigger. The design reservation — the frontend re-deriving backend attachability policy from emission side effects instead of receiving it — is now tracked precisely in #1347, which proposes emitting `can_attach`/`can_authorize` from `/api/mcp/apps` with contract tests.

## New findings

### N2 (design note, body-level) — the PR body's Scope claim is now factually wrong

Severity: minor (design note; the PR body, not the code). The Scope section ends with "once #1321 lands, a team-owned `mcp_oauth` connector will be listed for members, and this change is what makes it selectable." #1321's implementation (#1338) has since merged into main, and on current main that claim is false: team rows list with `is_connected: false`, `is_custom: true`, and `auth_type` withheld — deliberately, because `/{server_id}/oauth/connect` requires an active personal association and would 404 for a member who holds none (the `auth_type` emission guard and its comment in `src/xagent/web/api/mcp.py` on main). With `auth_type` absent, `isAttachable` is false, so team-owned `mcp_oauth` rows stay gated in select mode. This is not a regression — before #1338 they were not listed at all — and it is fully tracked in #1347's "Interaction with #1338" section and acceptance criterion 3 (`can_attach: true, can_authorize: false` for exactly this population). Suggested action: correct the Scope claim or cross-link it to #1347, so the body stops asserting behavior that main already disproves.

### N4 (minor, inline) — comment overstates the detail modal's re-enable affordance

`frontend/src/components/mcp/connect-mcp-dialog.tsx:235-238`. Severity: minor (comment accuracy). The claim that a deactivated entry keeps its card-click route "where it needs re-enabling rather than re-authorization" implies the detail modal offers re-enabling; it does not — no frontend caller of the toggle endpoint exists and the edit form has no `is_active` field (evidence and suggested rewording in the inline comment).

### N5 (minor, inline) — two test gaps for the un-gated population

`frontend/src/components/mcp/connect-mcp-dialog.test.tsx:1719`. Severity: minor (coverage). The population this PR un-gates has no selected-seed test (selected render + one-click deselect), and the Authorize → consent → auto-select chain is never driven end-to-end through the card's Authorize button. No hidden defect; both mechanisms verified (details in the inline comment).

### N6 (minor, inline) — connected fixture keeps auth_type the real population never has

`frontend/src/components/mcp/connect-mcp-dialog.test.tsx:1745`. Severity: minor (test fidelity). The fixture `{ ...unauthorizedCustomMcp(), is_connected: true }` retains `auth_type: "mcp_oauth"`, but a real deactivated-after-consent entry never has it; deleting it from the fixture changes no assertion (details in the inline comment).

- `L1822`: shrink duplicated apiRequestMock block by reusing renderSelectModeWith with an optional onConnectSelected. net: -6 lines possible.

(The 42-line invariant comment at tsx:219-252 was evaluated for shrinking and retained: it carries five backend emission invariants found nowhere else in the frontend, and #1347 is the right place for it to shrink.)

Blocking: no — all prior findings are FIXED or WAIVED, the new findings are minor or design-note level, and the simplification opportunity never blocks. Recommended event: APPROVE. There are no blocking issues with this PR as it stands; the follow-up work belongs in #1347.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -204,6 +204,51 @@ export function ConnectMcpDialog({
...
+  // - is_custom, because a *catalog* mcp_oauth app (e.g. Granola) carries the
+  //   same auth_type while is_connected: false there means the user has no
+  //   association row for it at all. Connecting really is a prerequisite, so
+  //   attaching it would select a connector that does not exist for them.
+  // - auth_type, because the backend emits it on a local entry only for the
+  //   mcp_oauth shape *and only while the association is active* (see
+  //   list_mcp_apps). An association deactivated before any grant completed is
+  //   listed with is_connected: false and no auth_type, so it keeps its
+  //   existing card-click route to the detail modal, where it needs
+  //   re-enabling rather than re-authorization. That matters because the
```
rogercloud: Minor (comment accuracy): "where it needs re-enabling rather than re-authorization" overstates what the detail modal offers for a deactivated entry. Verified: an unconnected entry renders only a Connect button (official-mcp-settings-dialog.tsx:214-226) that dispatches to the `notConfigured` toast for auth_type-less entries (tsx:911-924); no frontend caller of `POST /api/mcp/servers/{id}/toggle` exists (grep-verified), and the MCP edit form has no `is_active` field. The route and the dead end are pre-existing, but the "re-enabling" implication is introduced by this PR's comment. Suggest: "the card keeps its pre-existing route to the detail modal; re-enabling an inactive association is not currently offered by the picker".

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -1666,4 +1689,147 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
...
+      <ConnectMcpDialog
+        open
+        onOpenChange={vi.fn()}
+        selectedMcpServers={selectedMcpServers}
+        onConnectSelected={onConnectSelected}
+      />,
+    )
+  }
+
+  it("selects and deselects a custom mcp_oauth connector the editor holds no grant for (#1332)", async () => {
```
rogercloud, line 1749: Minor (test coverage). Two gaps for the population this PR un-gates: (1) no test seeds selectedMcpServers with an unconnected custom mcp_oauth entry (by id or name) asserting it renders as selected and deselects in one click — the only seeded entry (~1680) is a connected keyless catalog one; (2) the Authorize → consent → auto-select chain is never driven end-to-end through the card's Authorize button — the halves exist at 1753-1768 (button opens the modal) and ~581-654 (popup recheck auto-selects). No hidden defect: the selection toggle is shape-independent and the Authorize button holds no flow logic, both verified. Optional: seed an unconnected custom mcp_oauth entry and assert selected-render plus one-click removal.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -1666,4 +1689,147 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
...
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("shows Configure and the connected checkmark, not Authorize, for a connected custom mcp_oauth entry (#1332)", async () => {
+    // isAttachable widening the select-mode gate must not disturb the
+    // isGloballyConnected branch of the footer ternary: a connected entry
+    // has to keep showing Configure plus the checkmark, never Authorize.
+    const onConnectSelected = vi.fn()
+    renderSelectModeWith([{ ...unauthorizedCustomMcp(), is_connected: true }], onConnectSelected)
```
rogercloud, line 1779: Minor (fixture fidelity). This connected-entry fixture keeps `auth_type: "mcp_oauth"`, but a real deactivated-after-consent entry never has auth_type — `list_mcp_apps` emits it only while the association is active (src/xagent/web/api/mcp.py). Only the is_connected conjunct drives this branch, and deleting auth_type from the fixture changes no assertion (verified). Deleting it would mirror the real population, as the dormant fixture (~1806) already does, and stop this test from implying auth_type survives deactivation. Optional cleanup.

codeacme17:
Thanks. N2 and N4 are correct and fixed in 6c7fb77, N5(a) pointed at something real that turned out to be a surviving mutation worth killing, and N5(b) and N6 I am not taking — with measurements below. L1 I am declining in its proposed form but did the de-duplication it was after.

I also found that my previous round left a second, uncorrected copy of the very claim P3 fixed. That is in 6c7fb77 too.

## N2 — correct, Scope section rewritten

The claim was written before #1338 merged and main now disproves it, as you say. The section no longer asserts it; it states the withdrawal explicitly and, for completeness, that team-owned **non**-`mcp_oauth` rows report `is_connected: true` and so were made attachable by #1338 alone — this PR is not what makes those selectable either. Cross-linked to #1347 criterion 3.

The `isAttachable` comment had a related gap you noted: it described the emission guard as "only while the association is active" when it first requires an association to **exist**. That omission is exactly why team-owned rows land in the modal-route bucket, so it is now stated there.

## N4 — correct, all three sub-claims verified

Confirmed independently: an unconnected entry renders only Connect (`official-mcp-settings-dialog.tsx:214-226`); with no `auth_type` that Connect falls through to the `notConfigured` toast; `grep -rn "/toggle" frontend/src` returns nothing, so the endpoint at `src/xagent/web/api/mcp.py` has no frontend caller; and no MCP form exposes `is_active`.

I kept the backend's phrase but reattributed it, rather than deleting it — it explains why `list_mcp_apps` withholds the field, which is worth keeping, and the defect was that placing it right after "keeps its card-click route to the detail modal" made it read as something the modal provides. It now says plainly that the route offers no recovery today and marks the quote as the backend's rationale.

## Also fixed: a stale duplicate of P3's claim

P3 corrected the unqualified "deactivated → `is_connected: false`, no `auth_type`" claim in the production comment, but the dormant-entry test carried the identical uncorrected sentence, and its name said "whose association is deactivated" without qualification. Same inaccuracy, one copy fixed and one left behind. Both are now qualified to the pre-consent case, and the test names that population.

## N5(a) — real, but the useful part was a surviving mutation

You are right that the population is unseeded. Chasing it turned up something better than the suggested test: this mutation passed all 33 tests.

```ts
// connect-mcp-dialog.tsx, drop the id half:
const isSelected = localSelectedServers.includes(app.name)
```

That is precisely the regression the seeded test at :1640 exists for (#1280) — it asserted only the *committed* selection, never the rendered one, so the id-OR-name match was unpinned.

Two things followed:

- A new seeded test for the unconnected custom population killed nothing the suite did not already kill (the toggle path is shape-independent, as you note). Adding two assertions to the **existing** seeded test killed the mutation instead, with no new test and no new fixture. That is what 6c7fb77 does: the card carries `data-testid` and `data-selected`, and the test pins selected-before / unselected-after.
- The scenario as originally suggested is also not producible by the backend: `list_mcp_apps` sets both `"id"` and `"name"` to `server.name` for local entries, so `app.id === app.name` there, and agent-builder re-seeds with `connectedServer.name`. The id-seeding path is genuinely reachable only for catalog apps where id and name differ — which is what :1640 already covers.

So: fixed, but through the existing test rather than a new one.

## N5(b) — declining; it would cross the mock seam

The card's Authorize button does only `e.stopPropagation(); setSelectedApp(app)`. Consent starts from the *modal's* Connect via `onConnectStart`, and the settings dialog is mocked in this file — its `connect-records` button hardcodes `customMcpOauthApp()` (name `"records"`), a different object from the card's `unauthorizedCustomMcp()` (name `"Records MCP"`). An end-to-end test would join the two production edges through the mock, exercising nothing the existing pair does not: :1753 covers card→modal, :581 covers modal→consent→auto-select, and the latter is verified discriminating (deleting the auto-select append fails exactly it). The fake-timer complexity buys no coverage, so I have left it.

## N6 — declining; the current fixture is the faithful one, and measurably stronger

I think this one conflates two populations. `{ ...unauthorizedCustomMcp(), is_connected: true }` models an **ordinary connected** custom `mcp_oauth` connector — active association plus active grant — and for that population the backend emits both fields. The guard on main:

```python
if (
    user_mcp is not None
    and user_mcp.is_active
    and _is_mcp_oauth_server(server)
):
    entry["auth_type"] = "mcp_oauth"
```

and `is_connected` comes from `_mcp_oauth_server_is_actually_connected`, whose `active_grant_server_ids` filters on `user_id` plus grant `status == "active"` only — never association `is_active`. So `is_connected: true` **with** `auth_type` is the common real shape; `is_connected: true` **without** it is the degenerate deactivated-after-consent case, which is not what the test is named for.

Measured, dropping `auth_type` strictly weakens it:

| Mutation | current fixture | proposed fixture |
| --- | --- | --- |
| hoist the Authorize arm, keyed on `auth_type` | 2 failures (this test + catalog) | 1 failure (catalog only — this test passes) |
| `isGloballyConnected && !isSelectMode` | 1 failure (this test alone) | 1 failure |

I also looked for a mutation the deactivated-after-consent fixture would kill that the current one does not, and could not construct one: the natural candidate (`isAppConnected(app) && app.auth_type === "mcp_oauth"` as the first disjunct) is already killed by the seeded keyless test. That population is instead documented in the comment, which now distinguishes the two.

Happy to add a second fixture covering the degenerate shape if you would rather have it pinned explicitly — I just did not want to trade the stronger fixture for it.

## L1 — declining the proposal, did the de-duplication

Threading `onConnectSelected` through `renderSelectModeWith` as an optional argument would make the helper's name false for that caller, and `isSelectMode = !!onConnectSelected` is the single fact that test pins — demoting it to an omitted argument hides the precondition. Instead I extracted only the api mock as `mockAppsList(apps)`, shared by the helper and that test, which removes the duplicated block while the test keeps its explicit `render(...)`.

## One more thing 6c7fb77 changes

The two `connected-check` assertions are now scoped through the card via `within`. The testid sits inside `apps.map`, so it is non-unique by construction and both `getByTestId` and `queryByTestId` throw on multiple matches — the first test to render two connected apps would have failed confusingly. Same reasoning applies to the new card testid, which is keyed by `app.id`.

Verification: 77 tests green across `src/components/mcp` and `src/i18n`, `tsc --noEmit` clean, no new eslint problems. Mutation matrix re-run and every entry still killed, including the new `isSelected` one.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Confirmed on all three sub-claims and fixed in 6c7fb77.

Verified independently: `official-mcp-settings-dialog.tsx:214-226` renders only Connect when not connected; with no `auth_type` that Connect falls through to the `notConfigured` toast; `grep -rn "/toggle" frontend/src` returns nothing, so the toggle endpoint has no frontend caller at all; and no MCP form exposes `is_active`.

I kept the backend's phrase rather than deleting it, but reattributed it. It explains why `list_mcp_apps` withholds the field, which is worth recording — the defect was placing it immediately after "keeps its card-click route to the detail modal", which made it read as something the modal provides. The bullet now states plainly that the route offers no recovery today (no `auth_type` to dispatch on, no frontend caller of the toggle endpoint) and marks the quote as the backend's rationale for omitting the field.

Same pass also adds the clause you noted was missing: the guard first requires an association to **exist**, which is why team-owned rows from #1338's overlay get no `auth_type` and route to the modal instead of becoming attachable.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: Real, and chasing it found something better than the suggested test — fixed in 6c7fb77.

**(a)** The population is indeed unseeded, but a new seeded test for it killed no mutation the suite did not already kill; the toggle path is shape-independent, as you say. What was genuinely unpinned is the *existing* seeded test at :1640: it asserted only the committed selection, never the rendered one, so this mutation passed all 33 tests —

```ts
const isSelected = localSelectedServers.includes(app.name)   // drop the id half
```

— which is exactly the regression that test exists for (#1280). Adding two assertions to it kills the mutation with no new test and no new fixture: the card now carries `data-testid` and `data-selected`, and the test pins selected-before / unselected-after.

Worth noting the originally suggested scenario is not backend-producible either: `list_mcp_apps` sets both `"id"` and `"name"` to `server.name` for local entries, so `app.id === app.name` there, and agent-builder re-seeds with `connectedServer.name`. Id-seeding is genuinely reachable only for catalog apps where id and name differ — what :1640 already covers.

**(b)** Declining. The card's Authorize button does only `e.stopPropagation(); setSelectedApp(app)`; consent starts from the modal's Connect, and the settings dialog is mocked in this file, with `connect-records` hardcoding `customMcpOauthApp()` ("records") rather than the card's `unauthorizedCustomMcp()` ("Records MCP"). An end-to-end test would join the two edges across the mock and exercise nothing the existing pair does: :1753 covers card→modal, :581 covers modal→consent→auto-select, and deleting the auto-select append fails exactly :581. No coverage for the fake-timer cost.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: I am going to push back on this one — I think it conflates two populations, and the change measurably weakens the test.

The fixture models an **ordinary connected** custom mcp_oauth connector: active association plus active grant. For that population the backend emits *both* fields. The guard on main:

```python
if (
    user_mcp is not None
    and user_mcp.is_active
    and _is_mcp_oauth_server(server)
):
    entry["auth_type"] = "mcp_oauth"
```

and `is_connected` comes from `_mcp_oauth_server_is_actually_connected`, whose `active_grant_server_ids` filters on `user_id` plus grant `status == "active"` only — never association `is_active`. So `is_connected: true` **with** `auth_type` is the common real shape. `is_connected: true` **without** it is the deactivated-after-consent shape, which is real too but is not what this test is named for.

Measured:

| Mutation | current fixture | proposed fixture |
| --- | --- | --- |
| hoist the Authorize arm, keyed on `auth_type` | 2 failures (this test + catalog) | 1 failure (catalog only; this test passes) |
| `isGloballyConnected && !isSelectMode` | 1 failure (this test alone) | 1 failure |

I also tried to construct a mutation the deactivated-after-consent fixture kills that the current one does not, and could not: the natural candidate (`isAppConnected(app) && app.auth_type === "mcp_oauth"` as the first disjunct) is already killed by the seeded keyless test at :1640.

So leaving the fixture as-is. The degenerate population is now distinguished in the `isAttachable` comment instead. If you would rather have it pinned explicitly I am happy to add a *second* fixture for it — I just did not want to trade the stronger one away for it.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -624,6 +639,14 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
...
       expect(localListCalls).toBe(1)
       expect(onSuccess).toHaveBeenCalled()
+
+      // Commit the selection to observe it. Fake timers are still active, so
+      // this must stay on synchronous queries (getByRole) — findBy*/waitFor
+      // would hang.
+      act(() => {
+        fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+      })
+      expect(onConnectSelected).toHaveBeenCalledWith(["records"])
```
rogercloud, line 649: `customMcpOauthApp()` has `id === name === "records"`, so this assertion cannot distinguish the name-based auto-select append (connect-mcp-dialog.tsx:911-913 pushes `app.name`) from an id-based one. The gap is reachable: catalog `mcp_oauth` apps (id "granola", name "Granola") go through `handleConnectMcpOAuthApp` via the modal's Connect in select mode (connect-mcp-dialog.tsx:796, 809-812, 931-932), and under an id-append mutation the confirmed-consent commit would select the wrong entry while this test stays green — the Granola popup test (test.tsx:1301-1346) only asserts `onSuccess`. Cheap fix: after the closed-popup confirm in the Granola test, add `expect(onConnectSelected).toHaveBeenCalledWith(["Granola"])`, which the name-append implementation passes and an id-append one fails.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
```diff
@@ -1660,10 +1683,181 @@ describe("ConnectMcpDialog Custom API detail loading", () => {
...
+    fireEvent.click(screen.getByRole("button", { name: "tools.mcp.dialog.connect" }))
+    expect(onConnectSelected).toHaveBeenLastCalledWith([])
+  })
+
+  it("still gates selection on a custom mcp_oauth entry whose association was deactivated before consent completed", async () => {
+    // list_mcp_apps' local loop does not filter is_active, and emits auth_type
+    // only while the association *is* active — so a deactivated mcp_oauth
+    // server with no completed grant is listed as is_connected: false with no
+    // auth_type. (toggle_mcp_server never revokes a completed grant, so this
+    // only holds pre-consent: deactivating *after* consent instead lists
```
rogercloud, line 1841: This comment says the deactivated-after-consent population (`is_custom: true`, `is_connected: true`, no `auth_type`) "is exercised elsewhere", but no test renders that exact shape through the card path — all six `is_connected: true` sites were checked. The regression the comment guards (replacing the first disjunct) IS caught by the seeded keyless test, so the gap is comment accuracy, not coverage. Either reword the claim (the shape is covered only up to the `is_connected: true` conjunct) or add a card-rendered test for the exact shape.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -204,6 +204,62 @@ export function ConnectMcpDialog({
...
+  //
+  // Both conjuncts are load-bearing, and neither can be widened:
+  //
+  // - is_custom, because a *catalog* mcp_oauth app (e.g. Granola) carries the
+  //   same auth_type while is_connected: false there means the user has no
+  //   association row for it at all. Connecting really is a prerequisite, so
+  //   attaching it would select a connector that does not exist for them.
+  // - auth_type, because the backend emits it on a local entry only when a
+  //   personal association exists, is active, and the server is the
+  //   mcp_oauth shape (see list_mcp_apps). One consequence: a team-owned
```
rogercloud: The team-owned-row clause describes a population produced only by #1338's visibility overlay, which is not an ancestor of this branch — the statement is accurate for merged main but cannot be verified from this branch's own backend (on this branch `list_mcp_apps` emits `auth_type` only for local personal associations). Suggest a short clause noting the overlay arrives post-merge with #1338, so the claim reads as forward-looking rather than a property of this PR.

frontend/src/components/mcp/connect-mcp-dialog.tsx
```diff
@@ -1271,9 +1327,15 @@ export function ConnectMcpDialog({
...
                       return (
-                        <Card key={app.id} className={`p-[0] cursor-pointer transition-colors shadow-sm relative ${isSelectMode && isSelected ? 'border-blue-500 bg-blue-50/30 ring-1 ring-blue-500' : 'hover:border-slate-300 border-slate-200'}`} onClick={() => handleCardClick(app, isGloballyConnected)}>
+                        <Card
+                          key={app.id}
+                          data-testid={`connector-card-${app.id}`}
+                          data-selected={isSelectMode && isSelected}
+                          className={`p-[0] cursor-pointer transition-colors shadow-sm relative ${isSelectMode && isSelected ? 'border-blue-500 bg-blue-50/30 ring-1 ring-blue-500' : 'hover:border-slate-300 border-slate-200'}`}
+                          onClick={() => handleCardClick(app)}
+                        >
                           {isGloballyConnected && (
```
rogercloud, line 1341: `isGloballyConnected` (defined at line 1326) is now a pure alias of `isAppConnected(app)`: `handleCardClick` dropped its second parameter (line 1011), leaving only these display branches (1337, 1372, 1390) as consumers. Collapsing the alias into direct `isAppConnected(app)` calls would remove a near-identical predicate name from the touched code path. Optional readability nit — no behavior change either way.

rogercloud (approved):
## Summary

PR #1335 relaxes the "+ Connector" picker's select-mode gate, which previously keyed on `is_connected` and made custom `mcp_oauth` connectors with hook-supplied tokens permanently unattachable. The gate is now `isAttachable = isAppConnected(app) || (Boolean(app.is_custom) && app.auth_type === "mcp_oauth")`, and a card-level Authorize trigger keyed on the same predicate keeps the #1323 per-server OAuth flow reachable for diverted entries. Connected-state display is untouched. Blocking: no — recommended event: APPROVE.

## Update since last review (6c7fb77c)

One commit: comment rewrites (N4's re-enable overclaim removed, #1338 team-owned-row clause added, deactivated-before/after-consent populations distinguished), the `Card` gains `data-testid={"connector-card-" + app.id}` and `data-selected`, the seeded id-deselect test now pins rendered selected state before/after via `data-selected`, the duplicated `apiRequestMock` block is superseded by a `mockAppsList` extraction, the dormant test is renamed, and connected-check queries are scoped through the card. No behavioral change to the predicate or the Authorize trigger.

## Prior findings

- P1 (abandoned-consent un-gating) — WAIVED, verified: comment now cites `create_mcp_server`; tracked as #1347 consequence 1.
- P2 (Authorize dead-end for hook-resolved) — WAIVED, verified: no suppression signal exists; tracked as #1347 consequence 2 plus a `can_authorize` acceptance criterion.
- P3 (comment overclaimed deactivated-association protection) — FIXED: qualified to "deactivated before any grant completed"; after-consent case marked pre-existing.
- P4 (catalog fixture didn't exercise `is_custom`) — FIXED: renders `mcpOauthApp()` + keyless app; dropping the conjunct fails three assertions.
- P5 (no assertion on consent outcome) — FIXED: recheck test commits the footer and asserts `["records"]`; deleting the append fails it.
- P6 (Configure-vs-Authorize exclusion unpinned) — FIXED: new test pins Configure present / Authorize absent / checkmark present via `data-testid="connected-check"`.
- P7 (PR body overstated test failures) — FIXED: Tests section names exactly the two red tests plus a mutation table.
- N2 (scope claim wrong after #1338) — FIXED: withdrawal paragraph landed.
- N4 (comment overstated re-enable affordance) — FIXED: comment states no recovery today; backend rationale reattributed.
- N5 (test gaps) — FIXED: seeded test pins rendered selected state; the declined end-to-end Authorize test is technically sound (button holds no flow logic; both halves covered).
- N6 (connected fixture keeps `auth_type`) — FIXED/upheld: author pushback technically correct; the swap would weaken branch-order mutation kill from 2 to 1.
- L1822 (duplicated `apiRequestMock`) — FIXED/superseded: `mockAppsList` extraction, with a sound rationale for keeping `onConnectSelected` required.

## Approach verdict (carried forward)

Acceptable with reservations: a minimal, correctly-scoped frontend relaxation that matches the backend's accept-anything contract. Reservation: the frontend re-derives the backend's emission policy instead of receiving it; tracked in #1347.

## New findings

- N7 (test-quality, minor) — connect-mcp-dialog.test.tsx:649: `expect(onConnectSelected).toHaveBeenCalledWith(["records"])` cannot distinguish the name-based auto-select append (connect-mcp-dialog.tsx:911-913 pushes `app.name`) from an id-based one because `customMcpOauthApp()` has `id === name === "records"`. Not behavior-preserving to leave unguarded: catalog `mcp_oauth` apps (id "granola", name "Granola") reach `handleConnectMcpOAuthApp` via the modal's Connect in select mode (tsx:796, 809-812, 931-932), and after confirmed consent an id-append mutation would commit the wrong selector while the Granola popup test (test.tsx:1301-1346) asserts only `onSuccess`. Cheap fix — add to the Granola popup test after the closed-popup confirm:
  ```ts
  // passes under the current name-append implementation, fails under an id-append one
  expect(onConnectSelected).toHaveBeenCalledWith(["Granola"])
  ```
- N8 (documentation, minor) — connect-mcp-dialog.test.tsx:1830-1832: the dormant-test comment claims the deactivated-after-consent population "is exercised elsewhere", but no test renders that exact shape (`is_custom: true`, `is_connected: true`, no `auth_type`) through the card path — all six `is_connected: true` sites were checked. The originally-imagined regression (replacing the first disjunct) IS caught by the seeded keyless test, so the gap is comment accuracy, not coverage. Fix: correct the comment, or add a card-rendered test for that exact shape.
- N9 (documentation/reviewability, minor) — connect-mcp-dialog.tsx:235-238: the team-owned-row clause describes a population produced only by #1338, which is not an ancestor of this branch; accurate for merged main but unverifiable from this branch's own backend. Suggest a clause noting the overlay arrives post-merge (#1338).
- N10 (readability, minor) — connect-mcp-dialog.tsx:1326: `isGloballyConnected` is now a pure alias of `isAppConnected(app)` since `handleCardClick` dropped its second parameter (tsx:1011-1012); it survives only for display branches (:1337, :1372, :1390). Collapsing it would remove one near-identical predicate name from the touched code path.

- L1770 (shrink): redundant `.not.toBeNull()` on a `getByRole` that throws when absent — the bare call is the same assertion (test.tsx:1770).
- L1777 (shrink): same for a `getByTestId` wrapper (test.tsx:1777).

net: 0 lines (two redundant assertion wrappers; wrap removal saves tokens, not lines)

Blocking: no — all prior findings are FIXED or WAIVED-verified, the new findings are minor, and simplification never blocks. Recommended event: APPROVE. The remaining follow-up (frontend re-deriving the backend's emission policy) belongs in #1347.

codeacme17:
Thanks for the approval. N7, N8, N9 and both shrinks are in c6267d2; N10 I am declining, with reasoning below.

One note on N7: the suggested snippet would not have passed as written, and chasing why turned up a wider gap that I have deliberately left out of scope.

## N7 — real, implemented differently, and it exposed more than one unpinned site

The finding is right: `customMcpOauthApp()` has `id === name === "records"`, so that assertion cannot tell a name-append from an id-append.

The proposed one-liner needed one more step, though. The auto-select append writes **local** state (`setLocalSelectedServers`); `onConnectSelected` is only invoked by the footer commit. Dropping `expect(onConnectSelected).toHaveBeenCalledWith(["Granola"])` in after the closed-popup confirm would have failed against the current, correct implementation — the spy has not been called at that point. So the Granola test now clicks the footer first, exactly as the records test does, then asserts `["Granola"]`.

Verified discriminating: mutating the append to `app.id` fails exactly this test and nothing else.

That last part is worth flagging. There are **three** append sites, not one:

```
connect-mcp-dialog.tsx:713   connectCatalogApp's autoSelect
connect-mcp-dialog.tsx:916   handleConnectMcpOAuthApp (the one N7 is about)
connect-mcp-dialog.tsx:985   builtin_oauth's postMessage handler
```

Mutating **all three** to `app.id` still fails only the Granola test — so the keyless/api_key and builtin_oauth paths remain unpinned for the same id-vs-name confusion N7 identified. I have not touched them: this PR does not modify those paths, and pinning them means new fixtures and a new popup test for the builtin flow. Happy to file it as a test-debt issue if you think it is worth tracking.

## N8 — correct, comment fixed

No test renders `{is_custom: true, is_connected: true, no auth_type}` through the card, and "is exercised elsewhere" was wrong. The comment now says so directly, and states why the gap is acceptable: the shape is indistinguishable at the card from an ordinary connected entry, and the regression it would guard against — replacing the first disjunct with one that also demands `auth_type` — already fails the seeded keyless test. That is the same conclusion your N6 verdict reached from the other direction.

## N9 — correct, clause added

The clause now says the overlay is on main but landed after this branch was cut, so the population is not reproducible from this branch's own backend, and links #1347 for expressing that case as attachable-but-not-authorizable rather than deciding it by an absent hint.

## L1770 / L1777 — done

Both wrappers dropped; `getBy*` throws when absent, so the bare calls are the assertions. Confirmed no `no-unused-expressions` complaint from the project's eslint config.

## N10 — declining

`isGloballyConnected` predates this PR. It used to earn the local binding by being passed as `handleCardClick`'s second parameter; dropping that parameter is what left it a pure alias, so the observation is fair.

I am still leaving it:

- It has three remaining readers (`:1337`, `:1372`, `:1390`), none of which this PR otherwise touches. Collapsing it means editing three untouched lines to remove one line, which widens the diff for a naming preference.
- At those three sites the binding is doing real work as a name — the checkmark, the sharing badge and the Configure arm all read "is this connected" rather than re-invoking a predicate.

If you would rather have the alias gone I will do it — it is mechanical — I just did not want to spend diff on it inside a review round.

## State

c6267d2 on top of 6c7fb77. 77 tests green across `src/components/mcp` and `src/i18n`, `tsc --noEmit` clean, no new eslint problems. Full mutation matrix re-run after the changes — `is_custom` conjunct, `auth_type` conjunct, footer branch order, `isSelected`'s id half, and now the append key — each still killed by exactly one test.

The branch is behind main by a dozen commits and not rebased: this worktree holds another session's uncommitted work, and a rebase would require disturbing it. No conflicts (this PR is frontend-only; nothing in that range touches these files), and GitHub reports the PR mergeable.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: Correct, and fixed in c6267d2 — with one correction to the suggested snippet.

The append writes **local** state (`setLocalSelectedServers`); `onConnectSelected` is only invoked by the footer commit. So `expect(onConnectSelected).toHaveBeenCalledWith(["Granola"])` dropped in right after the closed-popup confirm would have failed against the current, correct implementation — the spy has not been called yet at that point. The Granola test now clicks the footer first (same as the records test at :649), then asserts `["Granola"]`.

Verified discriminating: mutating the append to `app.id` fails exactly this test.

Worth flagging what that mutation also showed — there are three append sites, not one:

```
connect-mcp-dialog.tsx:713   connectCatalogApp autoSelect
connect-mcp-dialog.tsx:916   handleConnectMcpOAuthApp  <- this finding
connect-mcp-dialog.tsx:985   builtin_oauth postMessage handler
```

Mutating all three still fails only the Granola test, so the keyless/api_key and builtin_oauth paths carry the same unpinned id-vs-name confusion. Left out of scope — this PR does not modify those paths, and covering them needs new fixtures plus a builtin-flow popup test. Say the word and I will file it as test debt.

frontend/src/components/mcp/connect-mcp-dialog.test.tsx
codeacme17: Correct — "is exercised elsewhere" was wrong, and c6267d2 fixes the comment.

Confirmed no test renders `{is_custom: true, is_connected: true, no auth_type}` through the card path. The comment now states that directly, plus why the gap is acceptable: at the card that shape is indistinguishable from an ordinary connected entry, and the regression it would guard against — replacing the first disjunct with one that also demands `auth_type` — already fails the seeded keyless test.

That is the same conclusion your N6 verdict reached from the other side, which is why I did not add a fixture for it.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Fair, and added in c6267d2.

The clause now notes that #1338's overlay is on main but landed after this branch was cut, so the population is not reproducible from this branch's own backend, and links #1347 for expressing the case as attachable-but-not-authorizable rather than letting an absent hint decide it.

For the record on why the branch is not simply rebased onto main, which would make the clause verifiable here: this worktree holds another session's uncommitted work and a rebase would require disturbing it. No conflicts either way — this PR is frontend-only and nothing in the range touches these files.

frontend/src/components/mcp/connect-mcp-dialog.tsx
codeacme17: Fair observation — `isGloballyConnected` used to earn its binding by being `handleCardClick`'s second parameter, and dropping that parameter is what left it a pure alias.

Still leaving it, for two reasons:

- Three readers remain (`:1337` checkmark, `:1372` sharing badge, `:1390` Configure arm), none of which this PR otherwise touches. Collapsing it means editing three untouched lines to delete one, which spends diff on a naming preference.
- At those sites the binding reads as a fact about the entry rather than a predicate call, which is what the surrounding display code wants.

It predates this PR, so nothing here regressed. If you would rather have the alias gone I will do it — it is mechanical — I just did not want to widen the diff mid-review.

PR #1338, opened 2026-08-13
fix(mcp): apply the team visibility overlay to the connector picker

### LINKED ISSUE #1321: Connector picker omits team-owned connectors: /api/mcp/apps skips the team visibility overlay that /api/mcp/servers applies

codeacme17:
Closes #1321

## Problem

`GET /api/mcp/apps?location=local` backs the agent's "+ Connector" dialog, but resolved connectors from the requesting user's own association rows only:

```python
db.query(MCPServer, UserMCPServer)
  .join(UserMCPServer, UserMCPServer.mcpserver_id == MCPServer.id)
  .filter(UserMCPServer.user_id == current_user.id)
```

`GET /api/mcp/servers` (the Tools page) additionally overlays the team-owned ids reported by `visible_team_connector_ids`. Sharing a connector with a team writes a team link row and **no** per-member association, so for every member but the creator both of the picker's loops resolved nothing: the connector was listed on the Tools page yet absent from the picker and impossible to attach to an agent. The Custom API half of the branch has the identical shape and dropped team-owned APIs for the same reason.

## Fix

Overlay the same team-owned ids in the local branch, for MCP servers and Custom APIs alike, excluding the ids already covered by a personal row so a member holding both sees the connector exactly once. No new hook shape is required — `visible_team_connector_ids(db, user_id)` answers exactly the question the picker needs, and is the same user-keyed hook `get_mcp_servers` already calls.

Two deliberate scoping decisions:

- **Local branch only.** Team servers are not injected into `user_mcps`, because that list also feeds `_build_active_oauth_server_lookup` / `_build_active_non_oauth_server_lookup` in the remote branch, which answer "is this catalog app connected *for me*" — a question only a personal association can establish. `location=remote` results are unchanged.

- **Team-owned rows do not carry the `auth_type: "mcp_oauth"` picker hint.** That hint tells the frontend to `POST /{server_id}/oauth/connect`, which calls `_get_user_mcp_server_or_404(..., require_active=True)`. A team member holds no association at all, so it would 404 and dead-end in a failed popup. Suppressing the hint mirrors the existing exclusion of deactivated associations, and leaves team-owned `mcp_oauth` connectors listed-but-unconnectable — outside this issue's scope (see #1313).

  To be precise about what that looks like, since an earlier version of this description overstated the parity: the picker renders a Connect button for any unconnected entry (`official-mcp-settings-dialog.tsx:214-227`), and with no `auth_type` the click falls through to the mis-authored-entry toast (`connect-mcp-dialog.tsx:878`). The Tools page serializes the same rows through `_TeamOwnedUserMCP` as inert cards with no Connect button, so the two surfaces are **not** identical here. The alternative is strictly worse (emitting the hint yields a popup that 404s), and before this PR the connector was invisible in the picker entirely — but it is a dead button, not a no-op, and `test_a_team_owned_mcp_oauth_server_carries_no_auth_type` pins it deliberately.

- **Team rows carry no ownership marker in this endpoint's entry shape.** Overlaid rows are indistinguishable from personal ones in the `/api/mcp/apps` payload; the frontend recognizes team tools and hides edit controls via the separate `POST /api/connectors/status` overlay (`connect-mcp-dialog.tsx:243`, `tools/page.tsx:306`), which is an overlay-only route absent in standalone. Backend `PUT`/`DELETE` stay 403-guarded regardless, so a missing overlay degrades decoration only — noting the dependency explicitly since this PR relies on it.

Standalone xagent installs no visibility hook, resolves empty, and takes the path it always did.

## Acceptance criteria

- [x] `/api/mcp/apps?location=local` lists a team-owned MCP connector for a member holding no personal association row.
- [x] The same holds for a team-owned Custom API.
- [x] A member holding both a personal association and a team link sees the connector exactly once.
- [x] A connector owned by a team the user does not belong to is never listed.
- [x] With no team hook installed (standalone), the response is unchanged.
- [x] `location=remote` results are unchanged.

## Tests

New `tests/web/api/test_mcp_apps_team_visibility.py` — 13 tests, one per acceptance criterion plus the catalog-name skip, the `mcp_oauth` hint suppression, and search filtering over overlaid rows. Written test-first.

Tightened in d0ff87b after review, because several assertions held regardless of the overlay:

- **AC4** installed a hook answering empty for *every* user, so it passed with the overlay deleted. It now reports a connector with no personal association for anyone — reachable only through the overlay — so the creator's assertion fails if the overlay is removed and the member's fails if it stops keying on the requesting user.
- **The category assertion was vacuous** (the local branch discards every row for any category other than `"All"`, so it held for any implementation) and is removed.
- **The AC5 docstring** promised "byte-identical" while asserting only id lists; reworded to what it pins.
- **Added:** `location="all"` (same branch as `"local"`, previously untested); an inactive personal row plus a team link listing one entry (pinning that the dedup index spans all associations regardless of `is_active`); and a hook answering string ids.

That last one documents a real fail-**open** path: SQLite's numeric affinity makes `id IN ('1')` match an INTEGER primary key, so a string-id answer resolves and lists the connector. I had claimed the opposite in an earlier review reply and have [corrected it](https://github.com/xorbitsai/xagent/pull/1338#discussion_r3776871363). Note that exact-`set` shape validation would not catch `{"1"}` — any hardening PR on `connector_team_scope` must validate element types, not just the container.

Mutation-checked against main's `list_mcp_apps`: 8 of 13 tests now fail without the fix, up from 4. The 5 that still pass are the intentional invariance tests (dedup, standalone, remote, catalog-name skip, inactive row).

Also verified: `tests/web/api/`, `tests/web/tools/`, and `tests/web/test_team_sharing_hooks.py` all pass; `mypy src/xagent/web/api/mcp.py` and `ruff` are clean.

### DIFF
--- a/src/xagent/web/api/mcp.py
+++ b/src/xagent/web/api/mcp.py
@@ -2009,8 +2009,37 @@ def list_mcp_apps(
             results.append(app_copy)
 
     if location in ["local", "all"]:
+        # Sharing a connector with a team writes a team link row and no
+        # per-member association, so the personal queries above resolve nothing
+        # for every member but the creator. Overlay the team-owned ids the way
+        # get_mcp_servers does, or the Tools page would list a connector the
+        # picker cannot offer (#1321). Scoped to this branch on purpose: the
+        # remote branch's lookups answer "is this catalog app connected *for
+        # me*", which only a personal association can establish. Standalone
+        # deployments install no hook, resolve empty, and take the same path
+        # they always did.
+        from ..services.connector_team_scope import visible_team_connector_ids
+
+        team_ids = visible_team_connector_ids(db, cast(int, current_user.id))
+
+        # (server, user_mcp) for a personal row; (server, None) for a
+        # team-owned connector the user holds no association for. Excluding the
+        # ids already covered personally is what keeps a member who holds both
+        # from seeing the connector twice.
+        local_mcps: list[tuple[MCPServer, UserMCPServer | None]] = list(user_mcps)
+        missing_mcp = [
+            sid for sid in team_ids["mcp"] if sid not in user_mcp_by_server_id
+        ]
+        if missing_mcp:
+            local_mcps.extend(
+                (server, None)
+                for server in db.query(MCPServer)
+                .filter(MCPServer.id.in_(missing_mcp))
+                .all()
+            )
+
         library_names = {app["name"].lower() for app in library_apps}
-        for server, user_mcp in user_mcps:
+        for server, user_mcp in local_mcps:
             if server.name.lower() in library_names:
                 continue
 
@@ -2057,8 +2086,15 @@ def list_mcp_apps(
             # Configure button away from the custom edit form. Inactive
             # associations are excluded because the per-server OAuth endpoints
             # require an active one — a deactivated server needs re-enabling,
-            # not re-authorization.
-            if user_mcp.is_active and _is_mcp_oauth_server(server):
+            # not re-authorization. A team-owned server (user_mcp is None) is
+            # excluded for the same reason, more strongly: the member holds no
+            # association at all, so /{server_id}/oauth/connect would 404 and
+            # the advertised flow would dead-end in a failed popup.
+            if (
+                user_mcp is not None
+                and user_mcp.is_active
+                and _is_mcp_oauth_server(server)
+            ):
                 entry["auth_type"] = "mcp_oauth"
 
             results.append(entry)
@@ -2071,7 +2107,20 @@ def list_mcp_apps(
             .all()
         )
 
-        for user_api, api in user_custom_apis:
+        # Same overlay as the MCP half above: a team-owned Custom API has no
+        # UserCustomApi row for the member. Nothing in the entry below reads
+        # the association, so team-owned APIs need no stand-in — only the id.
+        local_custom_apis: list[CustomApi] = [
+            api for _user_api, api in user_custom_apis
+        ]
+        own_api_ids = {cast(int, api.id) for api in local_custom_apis}
+        missing_api = [aid for aid in team_ids["custom_api"] if aid not in own_api_ids]
+        if missing_api:
+            local_custom_apis.extend(
+                db.query(CustomApi).filter(CustomApi.id.in_(missing_api)).all()
+            )
+
+        for api in local_custom_apis:
             if search:
                 search_lower = search.lower()
                 if search_lower not in api.name.lower() and (
--- a/tests/web/api/test_mcp_apps_team_visibility.py
+++ b/tests/web/api/test_mcp_apps_team_visibility.py
@@ -65,7 +65,9 @@ def _install_visibility(mapping: dict[int, dict[str, set[int]]]) -> None:
     connector_team_scope.set_connector_team_hooks(visibility=visibility)
 
 
-def _add_server(db, owner: User | None, name: str = "records") -> MCPServer:
+def _add_server(
+    db, owner: User, name: str = "records", *, is_active: bool = True
+) -> MCPServer:
     server = MCPServer.from_config(
         {
             "name": name,
@@ -78,20 +80,37 @@ def _add_server(db, owner: User | None, name: str = "records") -> MCPServer:
     db.add(server)
     db.commit()
     db.refresh(server)
-    if owner is not None:
-        db.add(
-            UserMCPServer(
-                user_id=owner.id,
-                mcpserver_id=server.id,
-                is_owner=True,
-                is_active=True,
-            )
+    db.add(
+        UserMCPServer(
+            user_id=owner.id,
+            mcpserver_id=server.id,
+            is_owner=True,
+            is_active=is_active,
         )
-        db.commit()
+    )
+    db.commit()
     return server
 
 
-def _add_oauth_server(db, owner: User | None, name: str = "notes") -> MCPServer:
+def _add_unowned_server(db, name: str) -> MCPServer:
+    """A server row with no personal association for anyone — reachable only
+    through the team overlay."""
+    server = MCPServer.from_config(
+        {
+            "name": name,
+            "description": f"{name} MCP server",
+            "managed": "external",
+            "transport": "stdio",
+            "command": f"{name}-mcp",
+        }
+    )
+    db.add(server)
+    db.commit()
+    db.refresh(server)
+    return server
+
+
+def _add_oauth_server(db, owner: User, name: str = "notes") -> MCPServer:
     server = MCPServer.from_config(
         {
             "name": name,
@@ -109,20 +128,19 @@ def _add_oauth_server(db, owner: User | None, name: str = "notes") -> MCPServer:
     db.add(server)
     db.commit()
     db.refresh(server)
-    if owner is not None:
-        db.add(
-            UserMCPServer(
-                user_id=owner.id,
-                mcpserver_id=server.id,
-                is_owner=True,
-                is_active=True,
-            )
+    db.add(
+        UserMCPServer(
+            user_id=owner.id,
+            mcpserver_id=server.id,
+            is_owner=True,
+            is_active=True,
         )
-        db.commit()
+    )
+    db.commit()
     return server
 
 
-def _add_custom_api(db, owner: User | None, name: str = "billing") -> CustomApi:
+def _add_custom_api(db, owner: User, name: str = "billing") -> CustomApi:
     api = CustomApi(
         name=name,
         description=f"{name} API",
@@ -132,16 +150,15 @@ def _add_custom_api(db, owner: User | None, name: str = "billing") -> CustomApi:
     db.add(api)
     db.commit()
     db.refresh(api)
-    if owner is not None:
-        db.add(
-            UserCustomApi(
-                user_id=owner.id,
-                custom_api_id=api.id,
-                is_owner=True,
-                is_active=True,
-            )
+    db.add(
+        UserCustomApi(
+            user_id=owner.id,
+            custom_api_id=api.id,
+            is_owner=True,
+            is_active=True,
         )
-        db.commit()
+    )
+    db.commit()
     return api
 
 
@@ -222,19 +239,30 @@ def test_a_personal_row_plus_a_team_link_lists_the_connector_exactly_once(db_ses
 
 
 def test_a_connector_owned_by_another_team_is_never_listed(db_session):
-    """AC4: the overlay adds exactly what the hook reports for this user."""
-    db, creator, member = db_session
-    _add_server(db, creator)
-    _add_custom_api(db, creator)
-    # The hook answers empty for `member`: they belong to no team owning these.
-    _install_visibility({int(creator.id): {"mcp": set(), "custom_api": set()}})
+    """AC4: the overlay is keyed on the *requesting* user, so a connector a
+    live hook reports for someone else must not reach this user's picker.
 
+    `shared` has no personal association for anyone, so it is reachable only
+    through the overlay: the creator's assertion fails if the overlay is
+    deleted, and the member's fails if it ignores the requesting user. An
+    answer that was empty for everyone — or a connector the creator also owned
+    personally — would leave both assertions true either way."""
+    db, creator, member = db_session
+    shared = _add_unowned_server(db, "shared")
+    _install_visibility(
+        {int(creator.id): {"mcp": {int(shared.id)}, "custom_api": set()}}
+    )
+
+    assert _local_ids(db, creator) == ["shared"]
     assert _local_ids(db, member) == []
 
 
 def test_standalone_response_is_unchanged_with_no_hook_installed(db_session):
     """AC5: `visible_team_connector_ids` resolves empty without a hook, so a
-    standalone deployment sees byte-identical results."""
+    standalone deployment takes the pre-overlay path. Entry *fields* on that
+    path are pinned by the existing tests in test_mcp_oauth_flow.py; what this
+    pins is that the overlay adds and removes nothing when no hook is
+    installed."""
     db, creator, member = db_session
     _add_server(db, creator)
     _add_custom_api(db, creator)
@@ -294,9 +322,13 @@ def test_a_team_owned_mcp_oauth_server_carries_no_auth_type(db_session):
     assert "auth_type" not in entry
 
 
-def test_search_and_category_filters_apply_to_overlaid_connectors(db_session):
-    """The overlay feeds the same loops, so the existing filters must still
-    narrow the result rather than being bypassed for team-owned rows."""
+def test_the_search_filter_applies_to_overlaid_connectors(db_session):
+    """The overlay feeds the same loops, so the existing search filter must
+    still narrow the result rather than being bypassed for team-owned rows.
+
+    No category assertion here: the local branch discards *every* row for any
+    category other than "All", so a category assertion holds even with the
+    overlay deleted and would pin nothing."""
     db, creator, member = db_session
     records = _add_server(db, creator, name="records")
     shipping = _add_server(db, creator, name="shipping")
@@ -323,9 +355,64 @@ def test_search_and_category_filters_apply_to_overlaid_connectors(db_session):
         )
     ] == ["records"]
 
-    assert (
-        list_mcp_apps(
-            location="local", category="Productivity", current_user=member, db=db
-        )
-        == []
+
+@pytest.mark.parametrize("location", ["local", "all"])
+def test_the_overlay_applies_to_both_local_and_all(db_session, location):
+    """`location="all"` runs the same branch as "local". The frontend only
+    sends "local" today, so pin that the branch is entered for both rather
+    than leaving "all" to drift."""
+    db, creator, member = db_session
+    server = _add_server(db, creator)
+    api = _add_custom_api(db, creator)
+    _install_visibility(
+        {
+            int(member.id): {
+                "mcp": {int(server.id)},
+                "custom_api": {int(api.id)},
+            }
+        }
     )
+
+    ids = [
+        a["id"] for a in list_mcp_apps(location=location, current_user=member, db=db)
+    ]
+    assert ids == ["records", "billing"]
+
+
+def test_an_inactive_personal_row_plus_a_team_link_lists_one_entry(db_session):
+    """A member who deactivated their own association while the connector stays
+    team-owned must still see exactly one row, not one per source.
+
+    The dedup index is built from *all* personal associations regardless of
+    `is_active`, matching `get_mcp_servers`' `own_mcp_ids`; an index that
+    filtered on `is_active` would overlay a second copy of this server."""
+    db, creator, _member = db_session
+    server = _add_server(db, creator, is_active=False)
+    _install_visibility(
+        {int(creator.id): {"mcp": {int(server.id)}, "custom_api": set()}}
+    )
+
+    assert _local_ids(db, creator).count("records") == 1
+
+
+def test_a_hook_answering_string_ids_is_not_silently_resolved(db_session):
+    """Regression pin for a fail-open path: SQLite's numeric affinity makes
+    `id IN ('1')` match an INTEGER primary key, so a hook answering string ids
+    would list a connector the dedup index (int-keyed) considered missing.
+
+    Element types are the hook's contract (`dict[str, set[int]]`). This pins
+    today's behavior so a future validator in `connector_team_scope` — which
+    must check element types, not just the container shape — has a test that
+    changes when the behavior does."""
+    db, creator, member = db_session
+    server = _add_server(db, creator)
+    _install_visibility(
+        {int(member.id): {"mcp": {str(server.id)}, "custom_api": set()}}  # type: ignore[dict-item]
+    )
+
+    listed = _local_ids(db, member)
+    # Documents the current fail-open outcome rather than asserting it is
+    # correct: a string id resolves through the IN clause and reaches the
+    # picker. A validator rejecting non-int members would turn this into a
+    # raised error, and this assertion is where that change surfaces.
+    assert listed == ["records"]
--- a/tests/web/api/test_mcp_apps_team_visibility.py
+++ b/tests/web/api/test_mcp_apps_team_visibility.py
@@ -1,0 +1,331 @@
+"""`/api/mcp/apps?location=local` must resolve the same connector set as
+`/api/mcp/servers` (#1321).
+
+Sharing a connector with a team writes a team link row and no per-member
+association, so the picker's personal-association queries resolve nothing for
+every member but the creator. Both endpoints overlay the team-owned ids the
+`connector_team_scope` visibility hook reports; these tests pin that the
+picker's local branch does it too, and that the standalone (no hook) and
+remote responses are untouched.
+"""
+
+from __future__ import annotations
+
+import pytest
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+from xagent.web.api.mcp import list_mcp_apps
+from xagent.web.models.custom_api import CustomApi, UserCustomApi
+from xagent.web.models.database import Base
+from xagent.web.models.mcp import MCPServer, UserMCPServer
+from xagent.web.models.public_mcp import PublicMCPApp
+from xagent.web.models.user import User
+from xagent.web.services import connector_team_scope
+
+
+@pytest.fixture()
+def db_session(tmp_path):
+    db_path = tmp_path / "mcp-apps-team.db"
+    engine = create_engine(
+        f"sqlite:///{db_path}", connect_args={"check_same_thread": False}
+    )
+    Base.metadata.create_all(engine)
+    SessionLocal = sessionmaker(bind=engine, autoflush=False, autocommit=False)
+    db = SessionLocal()
+
+    creator = User(username="alice", password_hash="x", is_admin=False)
+    member = User(username="bob", password_hash="x", is_admin=False)
+    db.add_all([creator, member])
+    db.commit()
+    db.refresh(creator)
+    db.refresh(member)
+
+    yield db, creator, member
+    db.close()
+    engine.dispose()
+
+
+@pytest.fixture(autouse=True)
+def _reset_connector_team_hooks():
+    """The hooks are process-global; never leak one into a sibling test."""
+    yield
+    connector_team_scope.set_connector_team_hooks()
+
+
+def _install_visibility(mapping: dict[int, dict[str, set[int]]]) -> None:
+    """Install a user-keyed visibility hook answering from ``mapping``."""
+
+    def visibility(_db, user_id: int) -> dict[str, set[int]]:
+        answer = mapping.get(int(user_id))
+        if answer is None:
+            return {"mcp": set(), "custom_api": set()}
+        return {"mcp": set(answer["mcp"]), "custom_api": set(answer["custom_api"])}
+
+    connector_team_scope.set_connector_team_hooks(visibility=visibility)
+
+
+def _add_server(db, owner: User | None, name: str = "records") -> MCPServer:
+    server = MCPServer.from_config(
+        {
+            "name": name,
+            "description": f"{name} MCP server",
+            "managed": "external",
+            "transport": "stdio",
+            "command": f"{name}-mcp",
+        }
+    )
+    db.add(server)
+    db.commit()
+    db.refresh(server)
+    if owner is not None:
+        db.add(
+            UserMCPServer(
+                user_id=owner.id,
+                mcpserver_id=server.id,
+                is_owner=True,
+                is_active=True,
+            )
+        )
+        db.commit()
+    return server
+
+
+def _add_oauth_server(db, owner: User | None, name: str = "notes") -> MCPServer:
+    server = MCPServer.from_config(
+        {
+            "name": name,
+            "managed": "external",
+            "transport": "streamable_http",
+            "url": "https://mcp.example.com/mcp",
+            "auth": {
+                "type": "mcp_oauth",
+                "resource": "https://mcp.example.com/mcp",
+                "issuer": "https://auth.example.com",
+                "scope": "notes.read",
+            },
+        }
+    )
+    db.add(server)
+    db.commit()
+    db.refresh(server)
+    if owner is not None:
+        db.add(
+            UserMCPServer(
+                user_id=owner.id,
+                mcpserver_id=server.id,
+                is_owner=True,
+                is_active=True,
+            )
+        )
+        db.commit()
+    return server
+
+
+def _add_custom_api(db, owner: User | None, name: str = "billing") -> CustomApi:
+    api = CustomApi(
+        name=name,
+        description=f"{name} API",
+        url="https://api.example.com/v1",
+        method="GET",
+    )
+    db.add(api)
+    db.commit()
+    db.refresh(api)
+    if owner is not None:
+        db.add(
+            UserCustomApi(
+                user_id=owner.id,
+                custom_api_id=api.id,
+                is_owner=True,
+                is_active=True,
+            )
+        )
+        db.commit()
+    return api
+
+
+def _add_catalog_app(
+    db, app_id: str = "granola", name: str = "Granola"
+) -> PublicMCPApp:
+    app = PublicMCPApp(
+        app_id=app_id,
+        name=name,
+        description="A catalog app",
+        icon="",
+        category="Productivity",
+        transport="streamable_http",
+        is_visible_in_connector=True,
+    )
+    db.add(app)
+    db.commit()
+    db.refresh(app)
+    return app
+
+
+def _local_ids(db, user: User) -> list[str]:
+    return [a["id"] for a in list_mcp_apps(location="local", current_user=user, db=db)]
+
+
+def test_local_branch_lists_a_team_owned_mcp_server_without_a_personal_row(db_session):
+    """AC1: the connector the Tools page already shows must reach the picker."""
+    db, creator, member = db_session
+    server = _add_server(db, creator)
+    _install_visibility(
+        {int(member.id): {"mcp": {int(server.id)}, "custom_api": set()}}
+    )
+
+    entry = next(
+        a
+        for a in list_mcp_apps(location="local", current_user=member, db=db)
+        if a["id"] == "records"
+    )
+    assert entry["server_id"] == server.id
+    assert entry["is_custom"] is True
+    assert entry["is_connected"] is True
+
+
+def test_local_branch_lists_a_team_owned_custom_api_without_a_personal_row(db_session):
+    """AC2: the Custom API half of the branch has the same association-only
+    shape, so it drops team-owned APIs for the same reason."""
+    db, creator, member = db_session
+    api = _add_custom_api(db, creator)
+    _install_visibility({int(member.id): {"mcp": set(), "custom_api": {int(api.id)}}})
+
+    entry = next(
+        a
+        for a in list_mcp_apps(location="local", current_user=member, db=db)
+        if a["id"] == "billing"
+    )
+    assert entry["server_id"] == api.id
+    assert entry["transport"] == "custom_api"
+    assert entry["is_connected"] is True
+
+
+def test_a_personal_row_plus_a_team_link_lists_the_connector_exactly_once(db_session):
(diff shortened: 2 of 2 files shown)


src/xagent/web/api/mcp.py
```diff
@@ -2009,8 +2009,37 @@ def list_mcp_apps(
...
+        # for every member but the creator. Overlay the team-owned ids the way
+        # get_mcp_servers does, or the Tools page would list a connector the
+        # picker cannot offer (#1321). Scoped to this branch on purpose: the
+        # remote branch's lookups answer "is this catalog app connected *for
+        # me*", which only a personal association can establish. Standalone
+        # deployments install no hook, resolve empty, and take the same path
+        # they always did.
+        from ..services.connector_team_scope import visible_team_connector_ids
+
+        team_ids = visible_team_connector_ids(db, cast(int, current_user.id))
```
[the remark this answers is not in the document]
codeacme17: Thanks — I looked into this, but I don't think the described defect exists, and I'd rather not make the change in this PR. Evidence below.

**No coercion happens.** `visible_team_connector_ids` is five lines (`src/xagent/web/services/connector_team_scope.py:116-120`):

```python
def visible_team_connector_ids(db: Any, user_id: int) -> dict[str, set[int]]:
    """Team-shared connector ids visible to user; empty when no hook/standalone."""
    if _connector_visibility_hook is None:
        return {"mcp": set(), "custom_api": set()}
    return _connector_visibility_hook(db, int(user_id))
```

With a hook installed the answer is returned **by identity** — no `dict(...)`, no `.get(..., set())`, no `try/except`. So a malformed answer is never defaulted to empty; it fails loudly at the call site: `None`/`str` → `TypeError` on `team_ids["mcp"]`, a missing key → `KeyError`, both surfacing as HTTP 500 (and `get_mcp_servers` converts it into an explicit logged 500 at `src/xagent/web/api/mcp.py:2266-2271`). A `{"mcp": {"1"}}`-style answer fails *closed*, since a string never matches an int id.

The empty return is reached only on the `_connector_visibility_hook is None` branch — the documented standalone path, covered by a test added in this PR. That is intended behavior, not a masked hook. The cited rule targets silently coercing a broken answer into an authoritative empty result, which is exactly what this function does not do.

**The asymmetry with `team_connector_ids` is deliberate.** `_validate_team_connector_answer` did not pre-exist and was not retrofitted onto the user-keyed hook — it was introduced in 747c77b (#1241) *together with* the new team-keyed `TeamConnectorVisibilityHook`. That hook is a runtime execution-authorization seam: its Protocol docstring notes every id it returns becomes executable by any runner of that team's agents, including non-members, using the shared definition row's stored credentials. That is why it also gets the typed-503 wrapper `resolve_team_connector_ids_or_raise`. The same commit deliberately keeps the legacy user-keyed overlay "exactly as it is" so installations adopting the revision without the new hook behave as before (`src/xagent/web/services/connector_runtime.py:611-620`).

**It is also out of scope here.** This PR does not modify `visible_team_connector_ids`. It has two pre-existing call sites besides the new one — `get_mcp_servers` and the `connector_runtime` legacy fallback (`connector_runtime.py:621`). The validator is intentionally exact: it rejects `frozenset`, `list`, and anything that is not a `set`. An out-of-tree application hook returning `{"mcp": [1, 2], ...}` works fine today (callers only iterate and do `in` tests) and would immediately hard-fail — including on the agent runtime path, which is deliberately *not* wrapped by the typed-503 helper, so it would propagate a raw `ValueError`. Note the module already types the downstream seams as `Collection[int]` rather than `set[int]` (`connector_team_scope.py:238, 271`).

If maintainers want the legacy user-keyed seam hardened, I'm happy to do it as its own PR covering both connector call sites and the KB analogue `visible_team_knowledge_bases` (`knowledge_base_team_scope.py:63-74`, which has the same pass-through shape), where the `set` vs `Collection[int]` strictness question can be decided deliberately. Expanding the blast radius to the runtime path from a UI-visibility bugfix isn't a trade I'd make here.

codeacme17:
Responded to the review's one inline comment in [its thread](https://github.com/xorbitsai/xagent/pull/1338#discussion_r3776224259). Summary: no code change, because the reported defect does not exist.

`visible_team_connector_ids` performs no coercion — with a hook installed it returns the hook's answer by identity, so a malformed answer fails loudly (`TypeError`/`KeyError` → 500) rather than resolving to an authoritative empty set. The empty return happens only on the no-hook standalone branch, which is documented behavior and covered by a test in this PR.

The asymmetry with `team_connector_ids` is deliberate: `_validate_team_connector_answer` was introduced in #1241 alongside the team-keyed hook, a runtime execution-authorization seam, and that commit explicitly keeps the legacy user-keyed overlay unchanged. This PR does not modify that function, and adding exact-`set` validation to it would hard-fail any installed hook returning a `list` or `frozenset` at two pre-existing call sites — one of them the agent runtime fallback, which is intentionally not wrapped by the typed-503 helper.

Happy to harden that legacy seam in a separate PR covering both connector call sites and the `visible_team_knowledge_bases` analogue if maintainers want it.

rogercloud (approved):
## Follow-up review — PR #1338

### Summary

This PR closes #1321 by making the connector picker (`list_mcp_apps`, `location in ["local", "all"]`) overlay `visible_team_connector_ids` the same way `get_mcp_servers` does, so a team member sees connectors shared with their team (MCP servers and Custom APIs) even without a personal association row — the same set the Tools page already lists. Ids already covered by a personal row are excluded so a creator holding both never sees a duplicate. Two decisions are deliberate and documented in code: the overlay is scoped to the local branch only (the remote branch's lookups answer "connected for me", which only a personal association can establish), and team-owned rows emit no `auth_type: "mcp_oauth"` hint (the OAuth endpoints require an active personal association and would 404). Nine new tests cover six acceptance criteria plus catalog-name dedup skipping, hint suppression, and search/category filtering; CI is green.

### Update since the last review

No new commits: both prior reviews (the COMMENTED review with one inline comment plus body suggestions, and the author's reply) sit on the current HEAD, and the update diff is empty. This follow-up is an independent re-review — a fresh design pass on the fix, two-channel verification of every finding, a simplification scan of the diff — and a re-check of the historical findings against the code.

### Prior findings checklist

- **P1 (inline, mcp.py:2023 — hook-answer shape validation, medium) — WAIVED.** All four of the author's counter-claims verified: the identity wrapper performs no coercion (connector_team_scope.py:116-120); #1241's asymmetry is deliberate (commit 747c77b keeps the legacy user-keyed overlay unchanged); an exact-set check would hard-fail list/frozenset-returning hooks at the two existing call sites; a separate hardening PR is the right vehicle (an analogous validator exists at knowledge_base_team_scope.py:63-69). Note that F9 below documents a failure mode (string ids) that shape validation would *not* catch — the hardening PR needs element-type checks too. Thread: https://github.com/xorbitsai/xagent/pull/1338#discussion_r3776224259
- **P2 (body suggestion, same concern) — WAIVED.** The premise that a malformed hook return is silently treated as empty does not hold in the code: all three call sites fail loudly; the only empty return is the documented no-hook standalone path, which AC5 pins.

### New confirmed findings (all non-blocking)

**Design-level — F3: the team−own + IN overlay pattern now exists in six places** (mcp.py:2030-2036 and mcp.py:2117-2120 are new; mcp.py:2232-2240, mcp.py:2253-2258, and connector_runtime.py:623-632/634-643 pre-date this PR), and the copies have already diverged in details (is_active filtering, `None` stand-in vs `_TeamOwnedUserMCP`, exception wrapping). Four of six are pre-existing, so this is not a regression; a later consolidation inside `connector_team_scope` would be the natural home. Not required for this PR.

**minor — F1: team-owned `mcp_oauth` connectors render a dead Connect button.** For a team row, `is_connected` (mcp.py:2068) is false and the auth_type gate (mcp.py:2093-2098) deliberately emits no hint, so the picker presents the row with a Connect button whose click always lands in the mis-authored-entry toast (frontend/src/components/mcp/connect-mcp-dialog.tsx:878 else branch; the button renders for unconnected custom entries at frontend/src/components/mcp/official-mcp-settings-dialog.tsx:214-227). The Tools page serializes the same rows through `_TeamOwnedUserMCP` as inert cards without a button, so the PR description's "mirrors the state the Tools page already shows" is not quite accurate. Non-blocking: the tradeoff is explicitly declared and tracked by #1313, `test_a_team_owned_mcp_oauth_server_carries_no_auth_type` pins it deliberately, the alternative (emitting the hint) is strictly worse (404 dead-end), and before this PR the connector was entirely invisible in the picker. Suggest acknowledging the toast behavior in the PR description.

**minor — F6: the category assertion is vacuous.** The local branch unconditionally continues for any category (`if category and category != "All": continue`), so `category="Productivity"` → `[]` (tests:327-331) holds for every implementation, including one with the overlay deleted. Only the two search assertions carry discriminating power; suggest removing the category assertion or making the branch actually filter by category.

**minor — F7: AC4 is an invariance test.** The hook answers empty for every user (tests:224-233), so the test passes even if the overlay is deleted entirely. Suggest answering `{server.id}` for the creator and empty for the member, then asserting the member still sees nothing.

**minor — F8: no inactive-personal-row + team-link dedup coverage.** A member holding an inactive personal row plus a team link should see exactly one entry, but no test constructs that state (the helpers hardcode `is_active=True`) and the picker side has no mirror of the runtime parity that pins the no-is_active-veto semantics. Suggest one test with an inactive row + team link asserting a single entry.

**low — F5: `location="all"` is untested.** The function has no `location` reference after line 2011, "all" executes byte-identical code to "local" (fully exercised by the nine local tests), and the frontend never sends "all" — the union-semantics risk predates this PR. A single parameterized case would close the gap cheaply.

**low — F9: no malformed-hook test at the new call site, and the "fail-closed" rebuttal is wrong.** The author's thread claim that a hook returning string ids would fail closed ("string never matches int id") is disproven: SQLite numeric affinity makes `IN ('1')` match an INTEGER primary key, so `{"mcp": {"1"}}` silently resolves and lists server id 1 as connected — precisely the fail-open class `_validate_team_connector_answer`'s docstring warns about. Risk remains low (hooks are installed by trusted host-app code, and runtime authorization goes through the separately validated team-keyed seam), but a test feeding a string-keyed answer to the new call site would pin the behavior. Note also that exact-set shape validation would not catch `{"1"}` (it is a legal set) — the planned hardening PR must validate element types.

**nit — F4: AC5 claims "byte-identical" but asserts only id lists.** The docstring (tests:235-237) promises more than the `["records", "billing"]` / `[]` assertions check; fields are already pinned by the existing test_mcp_oauth_flow.py tests, and this PR structurally cannot change the no-hook path's fields. Suggest asserting one full entry or softening the docstring.

**noted — F10: team rows carry no ownership marker.** Overlaid rows are indistinguishable from personal rows in the entry shape (mcp.py:2057), and the frontend relies on the /api/connectors/status overlay to recognize team tools and hide edit controls — a contract this PR does not mention. Backend PUT/DELETE remain 403-guarded, so a missing overlay degrades decoration only; a one-line note in the PR description would make the dependency explicit.

- tests/web/api/test_mcp_apps_team_visibility.py:68 — `delete:` drop `owner: User | None` and the `if owner is not None:` branch; all 12 call sites pass a real owner.
- tests/web/api/test_mcp_apps_team_visibility.py:94 — `delete:` same for `_add_oauth_server`.
- tests/web/api/test_mcp_apps_team_visibility.py:125 — `delete:` same for `_add_custom_api`.

Net ≈ −12 lines.

**Blocking: no — recommended event: APPROVE.** No critical or major findings: the two most consequential ones (F1, F9) are explicit, low-blast-radius tradeoffs — declared in the PR description, or mitigated by hook trust and a separately validated seam — and everything else is test quality and documentation.

src/xagent/web/api/mcp.py
```diff
@@ -2009,8 +2009,37 @@ def list_mcp_apps(
             results.append(app_copy)
 
     if location in ["local", "all"]:
```
rogercloud, line 2011: low — `location="all"` has no test. The function holds no `location` reference after this line, so "all" executes byte-identical code to "local" (covered by the nine local tests) and the frontend never sends "all"; the union-semantics risk predates this PR. A single parameterized test over ["local", "all"] would close the gap cheaply.

src/xagent/web/api/mcp.py
rogercloud, line 2023: low — this is a new call site for the visibility hook with no malformed-answer test. Also, a correction to the thread: a hook returning string ids does *not* fail closed — SQLite numeric affinity makes `IN ('1')` match an INTEGER pk 1, so `{"mcp": {"1"}}` silently resolves and lists server id 1 as connected, exactly the fail-open class `_validate_team_connector_answer`'s docstring warns about. Note the planned exact-set hardening would not catch `{"1"}` either (it is a legal set); element types must be validated. Risk stays low because hooks come from trusted host-app code and runtime authorization uses the separately validated team-keyed seam.

src/xagent/web/api/mcp.py
```diff
@@ -2009,8 +2009,37 @@ def list_mcp_apps(
...
+        from ..services.connector_team_scope import visible_team_connector_ids
+
+        team_ids = visible_team_connector_ids(db, cast(int, current_user.id))
+
+        # (server, user_mcp) for a personal row; (server, None) for a
+        # team-owned connector the user holds no association for. Excluding the
+        # ids already covered personally is what keeps a member who holds both
+        # from seeing the connector twice.
+        local_mcps: list[tuple[MCPServer, UserMCPServer | None]] = list(user_mcps)
+        missing_mcp = [
```
rogercloud, line 2030: Design-level (non-blocking) — this team−own + IN overlay is now the sixth copy of the pattern (new here and at lines 2117-2120; pre-existing at mcp.py:2232-2240, mcp.py:2253-2258, and connector_runtime.py:623-632/634-643), and the copies have already diverged (is_active filtering, None stand-in vs _TeamOwnedUserMCP, exception wrapping). 4/6 predate this PR, so no action is required here — flagging a future consolidation inside connector_team_scope.

src/xagent/web/api/mcp.py
```diff
@@ -2009,8 +2009,37 @@ def list_mcp_apps(
...
+            sid for sid in team_ids["mcp"] if sid not in user_mcp_by_server_id
+        ]
+        if missing_mcp:
+            local_mcps.extend(
```
rogercloud, line 2034: noted — overlaid team rows enter the same entry loop below and share the entry shape of personal rows with no ownership marker (mcp.py:2057). The frontend distinguishes team tools and hides edit controls via the /api/connectors/status overlay, a contract this PR does not mention; backend PUT/DELETE stay 403-guarded, so a missing overlay degrades decoration only. A one-line note in the PR description would make the dependency explicit.

src/xagent/web/api/mcp.py
```diff
@@ -2057,8 +2086,15 @@ def list_mcp_apps(
             # Configure button away from the custom edit form. Inactive
             # associations are excluded because the per-server OAuth endpoints
             # require an active one — a deactivated server needs re-enabling,
-            # not re-authorization.
-            if user_mcp.is_active and _is_mcp_oauth_server(server):
+            # not re-authorization. A team-owned server (user_mcp is None) is
+            # excluded for the same reason, more strongly: the member holds no
+            # association at all, so /{server_id}/oauth/connect would 404 and
+            # the advertised flow would dead-end in a failed popup.
+            if (
```
rogercloud, line 2093: minor — a team-owned mcp_oauth server reaches the entry above with `user_mcp=None`, so `is_connected` (mcp.py:2068) is false and this gate deliberately emits no auth_type hint. The picker therefore renders the row with a Connect button whose click always ends in the mis-authored-entry toast (frontend/src/components/mcp/connect-mcp-dialog.tsx:878 else branch; the button renders for unconnected custom entries at frontend/src/components/mcp/official-mcp-settings-dialog.tsx:214-227), while the Tools page serializes the same row through `_TeamOwnedUserMCP` as an inert card with no button. The PR description's "mirrors the state the Tools page already shows" is thus not quite accurate. Non-blocking: the tradeoff is declared and tracked by #1313, the alternative (emitting the hint) dead-ends in a 404, and the connector was fully invisible in the picker before this PR — but a sentence in the description acknowledging the toast would set expectations.

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+    def visibility(_db, user_id: int) -> dict[str, set[int]]:
+        answer = mapping.get(int(user_id))
+        if answer is None:
+            return {"mcp": set(), "custom_api": set()}
+        return {"mcp": set(answer["mcp"]), "custom_api": set(answer["custom_api"])}
+
+    connector_team_scope.set_connector_team_hooks(visibility=visibility)
+
+
+def _add_server(db, owner: User | None, name: str = "records") -> MCPServer:
```
rogercloud: shrink — `owner: User | None` and the `if owner is not None:` branch are never exercised with None (all 12 call sites across the file pass a real owner). Make `owner: User` required and drop the branch (−4 lines); same for `_add_oauth_server` (line 94) and `_add_custom_api` (line 125).

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+                mcpserver_id=server.id,
+                is_owner=True,
+                is_active=True,
+            )
+        )
+        db.commit()
+    return server
+
+
+def _add_oauth_server(db, owner: User | None, name: str = "notes") -> MCPServer:
```
rogercloud: shrink — same as line 68: `owner` is always a real user at every call site; make it required and drop the `if owner is not None:` branch.

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+                mcpserver_id=server.id,
+                is_owner=True,
+                is_active=True,
+            )
+        )
+        db.commit()
+    return server
+
+
+def _add_custom_api(db, owner: User | None, name: str = "billing") -> CustomApi:
```
rogercloud: shrink — same as line 68: `owner` is always a real user at every call site; make it required and drop the `if owner is not None:` branch.

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+        a
+        for a in list_mcp_apps(location="local", current_user=member, db=db)
+        if a["id"] == "billing"
+    )
+    assert entry["server_id"] == api.id
+    assert entry["transport"] == "custom_api"
+    assert entry["is_connected"] is True
+
+
+def test_a_personal_row_plus_a_team_link_lists_the_connector_exactly_once(db_session):
```
rogercloud, line 222: minor — no test covers a member holding an inactive personal row plus a team link (should resolve to exactly one entry; the dedup filter at mcp.py:2030-2036 excludes by id regardless of active state). The helpers hardcode `is_active=True`, so that state cannot even be constructed today, and the runtime-side parity pinning the no-is_active-veto semantics has no picker-side mirror. One test with an inactive row + team link would pin it.

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+            }
+        }
+    )
+
+    ids = _local_ids(db, creator)
+    assert ids.count("records") == 1
+    assert ids.count("billing") == 1
+
+
+def test_a_connector_owned_by_another_team_is_never_listed(db_session):
```
rogercloud, line 241: minor — as written this is an invariance test: the hook answers empty for every user, so the test passes even if the overlay is deleted entirely. Suggest answering `{server.id}` for the creator and empty for the member, then asserting the member still sees nothing — the assertions then actually discriminate the overlay.

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+    db, creator, member = db_session
+    _add_server(db, creator)
+    _add_custom_api(db, creator)
+    # The hook answers empty for `member`: they belong to no team owning these.
+    _install_visibility({int(creator.id): {"mcp": set(), "custom_api": set()}})
+
+    assert _local_ids(db, member) == []
+
+
+def test_standalone_response_is_unchanged_with_no_hook_installed(db_session):
```
rogercloud, line 260: nit — the docstring promises "byte-identical results" but the assertions only compare id lists (`["records", "billing"]` / `[]`). Fields are already pinned by the existing test_mcp_oauth_flow.py tests and this PR structurally cannot change the no-hook path's fields — either assert one full entry here or soften the wording.

tests/web/api/test_mcp_apps_team_visibility.py
```diff
@@ -0,0 +1,331 @@
...
+    assert [
+        a["id"]
+        for a in list_mcp_apps(
+            location="local", search="record", current_user=member, db=db
+        )
+    ] == ["records"]
+
+    assert (
+        list_mcp_apps(
+            location="local", category="Productivity", current_user=member, db=db
```
rogercloud: minor — this assertion is vacuous: the local branch unconditionally continues for any category (`if category and category != "All": continue`), so `category="Productivity"` → `[]` holds for any implementation, including one with the overlay deleted. Only the two search assertions above carry discriminating power; either drop this assertion or make the branch actually filter by category.

src/xagent/web/api/mcp.py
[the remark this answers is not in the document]
codeacme17: Correction to my earlier reply in this thread, prompted by F9 of @rogercloud's follow-up review.

I claimed a hook answering string ids would "fail closed" because "a string never matches an int id". **That is wrong.** SQLite applies numeric affinity to the comparison, so `id IN ('1')` matches an INTEGER primary key:

```
sqlite> CREATE TABLE t (id INTEGER PRIMARY KEY, name TEXT);
sqlite> INSERT INTO t VALUES (1, 'one'), (2, 'two');
sqlite> SELECT id, name FROM t WHERE id IN ('1');
1|one
```

Verified through SQLAlchemy's expanding bindparam as well — same result. So `{"mcp": {"1"}}` passes the int-keyed dedup index as "missing", resolves through the `IN` clause, and lists server id 1. That is fail-**open**, the exact class the `_validate_team_connector_answer` docstring warns about.

The rest of my reply stands: no coercion to empty occurs, the empty return is only the documented no-hook path, and the #1241 asymmetry is deliberate. But the fail-closed claim was one of my four supporting points and it was incorrect — worth stating plainly rather than leaving it in the thread.

Two consequences, both now handled:

1. Added `test_a_hook_answering_string_ids_is_not_silently_resolved` (d0ff87b) pinning today's fail-open behavior at the new call site, so the change surfaces in a test when a validator lands.
2. As F9 notes, exact-`set` shape validation would **not** catch `{"1"}` — it is a legal `set`. Any hardening PR on this seam must validate element types, not just the container. I've noted that for the follow-up.

codeacme17:
Thanks for the thorough re-review. Acted on all the actionable findings in d0ff87b plus a description update; nothing in `src/` changed, so the approval should still hold.

**F9 — you're right, and my rebuttal was wrong.** I verified it directly: SQLite numeric affinity makes `id IN ('1')` match an INTEGER primary key, so a string-id hook answer resolves fail-**open**, not fail-closed as I claimed. I've [posted a correction](https://github.com/xorbitsai/xagent/pull/1338#discussion_r3776871363) in the original thread rather than leave the bad claim standing, and added `test_a_hook_answering_string_ids_is_not_silently_resolved` pinning today's behavior at the new call site. Your point that exact-`set` validation would not catch `{"1"}` is the important one for the follow-up — I've recorded that the hardening PR needs element-type checks, not just container shape.

**F7 — fixed, and your suggested form wasn't quite enough.** Answering `{server.id}` for the creator still left the test non-discriminating, because the creator owns that connector personally, so their assertion holds with the overlay deleted. The test now uses a connector with no personal association for *anyone*, reachable only through the overlay: the creator's assertion fails if the overlay is removed, the member's fails if it stops keying on the requesting user.

**F6 — removed.** Confirmed vacuous; kept a comment explaining why a category assertion pins nothing here.

**F8 — added.** Inactive personal row + team link asserts one entry, pinning that the dedup index spans all associations regardless of `is_active`.

**F5 — added** as a parameterized `["local", "all"]` case.

**F4 — docstring reworded** to what it actually pins.

**Simplifications — applied.** Dropped the `owner: User | None` branches from all three helpers; added a dedicated `_add_unowned_server` for the one case that genuinely needs no association.

**F1 and F10 — description updated.** You're right that "mirrors the state the Tools page already shows" was inaccurate: the picker renders a Connect button for unconnected entries (`official-mcp-settings-dialog.tsx:214-227`) that falls through to the mis-authored-entry toast, while the Tools page renders inert `_TeamOwnedUserMCP` cards with no button. The description now says so plainly, and notes the `POST /api/connectors/status` dependency for team-tool decoration.

**F3 — agreed, not here.** Six copies of the team−own + IN pattern with divergent details is real, but four predate this PR; consolidating inside `connector_team_scope` belongs with the hook-hardening PR, where the `set` vs `Collection[int]` question also gets settled.

I mutation-checked the suite against main's `list_mcp_apps`: 8 of 13 tests now fail without the fix, up from 4. The 5 that still pass are the intentional invariance tests (dedup, standalone, remote, catalog-name skip, inactive row).

src/xagent/web/api/mcp.py
codeacme17: Fixed in d0ff87b — added `test_the_overlay_applies_to_both_local_and_all`, a `@pytest.mark.parametrize("location", ["local", "all"])` case asserting the overlay lands for both.

You are right that the risk is low (no `location` reference after this line, so `"all"` runs byte-identical code, and the frontend only sends `"local"`), but the case is one line of parametrize and stops `"all"` from drifting silently if this branch ever gains a `location` reference.

src/xagent/web/api/mcp.py
codeacme17: You are right and my earlier claim in the sibling thread was wrong. I verified it directly:

```
sqlite> CREATE TABLE t (id INTEGER PRIMARY KEY, name TEXT);
sqlite> INSERT INTO t VALUES (1, 'one'), (2, 'two');
sqlite> SELECT id, name FROM t WHERE id IN ('1');
1|one
```

Same through SQLAlchemy's expanding bindparam. So `{"mcp": {"1"}}` passes the int-keyed dedup index as "missing", resolves through the `IN` clause, and lists server id 1 — fail-**open**, not fail-closed. I have [posted a correction](https://github.com/xorbitsai/xagent/pull/1338#discussion_r3776871363) in the original thread rather than leave the bad claim standing.

Added `test_a_hook_answering_string_ids_is_not_silently_resolved` (d0ff87b) pinning today's behavior at this call site, so a future validator changes a test rather than passing silently.

Your second point is the one that matters most for the follow-up: exact-`set` shape validation would **not** catch `{"1"}` — it is a legal `set`. The hardening PR has to validate element types, not just the container. Recorded.

src/xagent/web/api/mcp.py
codeacme17: Agreed on the diagnosis, and I would rather not consolidate here. Four of the six copies pre-date this PR (mcp.py:2232-2240, mcp.py:2253-2258, connector_runtime.py:623-632/634-643), so a consolidation lands as a refactor across the runtime path inside a UI-visibility bugfix — the same blast-radius objection I made to hardening `visible_team_connector_ids` in this PR.

The natural vehicle is the hook-hardening PR I offered in the other thread: it already has to touch `connector_team_scope`, and the divergences you list (is_active filtering, `None` stand-in vs `_TeamOwnedUserMCP`, exception wrapping) need the same decision the `set` vs `Collection[int]` question does. I will carry this finding into that PR.

src/xagent/web/api/mcp.py
codeacme17: Good catch on the undocumented dependency — added a note to the PR description.

Confirmed the contract: `POST /api/connectors/status` is an overlay-only route consumed by both the picker (`connect-mcp-dialog.tsx:243`) and the Tools page (`tools/page.tsx:306`), and both skip the call entirely when the user is not in a team. Backend `PUT`/`DELETE` stay 403-guarded independently, so a missing overlay degrades decoration only rather than opening a mutation path — but this PR does rely on that overlay to mark team rows, and that was worth stating rather than leaving implicit.

src/xagent/web/api/mcp.py
codeacme17: You are right, and the PR description was inaccurate — now fixed.

I verified the frontend path: the settings dialog renders a Connect button for any entry with `!isGloballyConnected` (`official-mcp-settings-dialog.tsx:214-227`), and `handleConnectApp` with no recognized `auth_type` falls through to `toast.error(t('tools.mcp.alerts.notConfigured'))` (`connect-mcp-dialog.tsx:878`). Meanwhile `/api/mcp/servers` serializes the same rows through `_TeamOwnedUserMCP` as inert cards. So "mirrors the state the Tools page already shows" was wrong: it is a dead button, not a no-op, and the two surfaces are not identical.

Keeping the behavior as-is — emitting the hint is strictly worse (a popup that 404s on `_get_user_mcp_server_or_404(..., require_active=True)`), and before this PR the connector was invisible in the picker entirely — but the description now says plainly what the user sees, and `test_a_team_owned_mcp_oauth_server_carries_no_auth_type` pins the tradeoff deliberately. Tracked by #1313.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Applied in d0ff87b — `owner` is now a required `User` on `_add_server`, `_add_oauth_server` and `_add_custom_api`, and the `if owner is not None:` branches are gone.

One case did turn out to need an ownerless server after all: the AC4 fix (your F7) needs a connector with no personal association for anyone, so I added a dedicated `_add_unowned_server` rather than resurrecting the optional parameter. Clearer at the call site about why that row has no association.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Applied in d0ff87b, along with the other two helpers — see the note on the `_add_server` thread about `_add_unowned_server` covering the one genuinely ownerless case.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Applied in d0ff87b, along with the other two helpers — see the note on the `_add_server` thread about `_add_unowned_server` covering the one genuinely ownerless case.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Added in d0ff87b — `test_an_inactive_personal_row_plus_a_team_link_lists_one_entry`, with `_add_server(..., is_active=False)` now that the helper takes the flag.

It pins exactly the semantics you name: the dedup index is built from **all** personal associations regardless of `is_active` (matching `get_mcp_servers`' `own_mcp_ids`), so an index that filtered on `is_active` would overlay a second copy of the same server and the count assertion would catch it.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Fixed in d0ff87b — and your suggested form was not quite sufficient, which is worth flagging.

Answering `{server.id}` for the creator still leaves the test non-discriminating, because the creator owns that connector **personally**: their assertion holds whether or not the overlay exists, and the member's holds too. I first wrote it that way and the mutation check caught it.

The test now uses a connector with no personal association for anyone, reachable only through the overlay. Both directions now bite: the creator's assertion fails if the overlay is deleted, the member's fails if it stops keying on the requesting user.

Mutation-checked against main's `list_mcp_apps`: 8 of 13 tests now fail without the fix, up from 4.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Fixed in d0ff87b — reworded to what the test actually pins: that the overlay adds and removes nothing when no hook is installed, with a pointer that entry *fields* on the no-hook path are already covered by `test_mcp_oauth_flow.py`.

Agreed that asserting a full entry here would duplicate that coverage, and that this PR structurally cannot change the no-hook path's fields — so the docstring was the thing that was wrong, not the assertions.

tests/web/api/test_mcp_apps_team_visibility.py
codeacme17: Removed in d0ff87b. Confirmed vacuous: `if category and category != "All": continue` fires for every row in the local branch, so `category="Productivity"` returns `[]` for any implementation, overlay or not.

Renamed the test to `test_the_search_filter_applies_to_overlaid_connectors` and left a comment recording why a category assertion pins nothing here, so nobody re-adds it. Did not make the branch actually filter by category — that is a behavior change beyond this fix, and the local branch has never supported category filtering.

PR #1341, opened 2026-08-13
feat(v1): stream task lifecycle events over SSE

AlexLiu190625:
## Why

Starting a task through the SDK gives no feedback for the first several seconds: `steps()` polls once per second, and until the first step lands there is nothing to report. A stream lets a client learn a task is alive the moment it attaches, and learn it ended the moment it ends.

## What

Adds `GET /v1/chat/tasks/{task_id}/events`, a Server-Sent Events stream of task lifecycle events, authenticated with the same key dependency as the other `/v1/chat/tasks` endpoints (one bcrypt per stream rather than one per poll).

The stream emits four event types — `task.status`, `task.completed`, `task.input_required` (the question text is always `null` at this layer), and `stream.error` (`resync_required` / `unauthorized` / `task_deleted` / `stream_expired`) — plus a `: ping` comment every 15 seconds. Projecting step and message content is a separate layer and is not part of this change; the endpoint docstring documents exactly what it emits.

Lifecycle rules:

- A 30-second watchdog on the task row is the authority for terminal state; broadcast frames only make it react sooner. The same watchdog closes streams on key revocation and on task deletion, and emits `task.input_required` for a task waiting on the user. A paused task keeps its stream open, matching `wait()`, which also keeps waiting through a pause.
- A stream lives at most one hour and always ends with an explicit close event: a frame already in flight can never displace the close frame, so a client can always tell "task ended" from "stream ended".
- The outbound queue is bounded at 256 frames. A consumer too slow to keep up gets `stream.error {resync_required}` and re-syncs through `steps()` before re-attaching. There is no replay; `Last-Event-ID` is ignored.
- Concurrency caps are 2 streams per task and 32 per key. Both are enforced when the response is built and are best-effort under a burst of simultaneous attaches — the count can be exceeded briefly rather than the extra attaches being rejected, and it corrects itself as streams close. No stream is ever aborted for this reason.
- The sink swallows every exception and counts dropped frames: a stream subscriber must never be able to break the task's own broadcast path.

## Verification

38 transport tests covering: fail-closed auth and 404, immediate close when attaching to an already-terminal task, slow-consumer bounding, key revocation closing within one watchdog interval, waiting-for-user close carrying a null prompt, a paused task continuing to stream, absolute-deadline close with `stream_expired`, closing exactly once when the watchdog and the deadline race, heartbeats, a broadcast frame reaching the client as `task.status` with same-status dedup, the completion hint waking the watchdog early, cleanup leaving no registration or quota behind on both disconnect and never-started streams, per-task and per-key caps, no request-scoped DB session on the endpoint, and zero DB reads on the frame path.

Full `tests/web/api/v1` suite green on top of main.

### DIFF
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -137,6 +137,13 @@ class V1EventStreamSink:
     ``_resolve_task_or_404``.
     """
 
+    # This sink only consumes the broadcast fan-out; it must never be
+    # picked as the target for a personal reply to a durable command (see
+    # ``websocket.py``'s ``_execute_durable_task_command``), since it only
+    # understands versioned task-status events and would silently drop
+    # anything else routed to it.
+    is_broadcast_only = True
+
     def __init__(
         self, *, task_id: int, principal_key_prefix: str, initial_status: str
     ) -> None:
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -83,6 +83,37 @@ _MIN_WAIT_BUDGET_SECONDS = 0.05
 
 _TERMINAL_STATUSES = (TaskStatus.COMPLETED, TaskStatus.FAILED)
 _TERMINAL_STATUS_VALUES = {status.value for status in _TERMINAL_STATUSES}
+# A task moving to ``waiting_for_user`` also warrants an early watchdog
+# wake, same as a terminal one -- both are broadcasts that make the
+# authoritative row read (by the watchdog, or the attach-time snapshot)
+# worth re-running sooner than the next scheduled cycle.
+_EARLY_WAKE_STATUS_VALUES = _TERMINAL_STATUS_VALUES | {
+    TaskStatus.WAITING_FOR_USER.value
+}
+
+
+def _stream_close_reason(status: TaskStatus, control_state: str | None) -> str | None:
+    """Shared close determination for a task's authoritative row state.
+
+    Both the watchdog (``watchdog_check_once``, polling) and the
+    attach-time fast path (``build_event_stream_response``, one-shot at
+    open) need to reach the same conclusion from the same two fields, so
+    they call this instead of duplicating the condition. Returns
+    ``"terminal"`` when the task is completed/failed, ``"input_required"``
+    when it's stuck waiting on user input (and not mid-resume), or
+    ``None`` when the stream should keep going. Each caller still builds
+    its own output from the category -- a close frame on the watchdog's
+    already-open stream, or a fast-path response before one ever opens.
+    """
+    if status in _TERMINAL_STATUSES:
+        return "terminal"
+    if (
+        status is TaskStatus.WAITING_FOR_USER
+        and control_state != TaskControlState.RESUME_REQUESTED.value
+    ):
+        return "input_required"
+    return None
+
 
 # A sync callable: (task_id, principal) -> ``_TaskInfoSnapshot``-shaped
 # object (duck-typed: ``.status`` / ``.control_state`` / ``.output`` /
@@ -250,8 +281,6 @@ class V1EventStreamSink:
             message = json.loads(text)
             if not isinstance(message, dict):
                 return
-            if not self._binding_matches(message):
-                return
             if message.get("type") == "task_completed":
                 # Acceleration signal only: the authoritative completion
                 # frame still comes from the watchdog reading the task
@@ -264,11 +293,18 @@ class V1EventStreamSink:
                 status = message.get("status")
                 if isinstance(status, str):
                     self.enqueue_status(status)
-                    if status in _TERMINAL_STATUS_VALUES:
+                    if status in _EARLY_WAKE_STATUS_VALUES:
                         # A failed task's broadcast (``task_error``) never
-                        # carries ``type == "task_completed"``, so it would
-                        # otherwise miss the acceleration signal above and
-                        # wait out the watchdog's normal cadence.
+                        # carries ``type == "task_completed"``, and a task
+                        # moving to ``waiting_for_user`` never does either,
+                        # so both would otherwise miss the acceleration
+                        # signal above and wait out the watchdog's normal
+                        # cadence. Safe to wake early on either: the
+                        # watchdog re-reads the authoritative row (via
+                        # ``_stream_close_reason``) before closing anything,
+                        # including the ``resume_requested`` carve-out, so
+                        # an early wake never closes a stream that a fresh
+                        # read wouldn't have closed anyway.
                         self.completion_hint.set()
         except Exception:
             self.dropped_frame_count += 1
@@ -276,7 +312,7 @@ class V1EventStreamSink:
                 "v1 SSE sink dropped a broadcast frame for task %s", self.task_id
             )
 
-    async def close(self, *args: Any, **kwargs: Any) -> None:
+    async def close(self) -> None:
         """No-op. Duck-types ``WebSocket.close`` -- task deletion
         (``chat.py``'s ``_cleanup_runtime_state``) calls ``close()`` on every
         connection still registered for the task, and a missing method here
@@ -285,23 +321,6 @@ class V1EventStreamSink:
         ``finally`` block is what actually tears the stream down.
         """
         return None
-
-    def _binding_matches(self, message: dict[str, Any]) -> bool:
-        """Drop frames whose task_id doesn't match this stream's binding
-        (defense in depth -- ``manager`` already scopes delivery to this
-        task's connections, so this only matters if a connection is ever
-        reassigned via ``move_connection``)."""
-        candidate = message.get("task_id")
-        if candidate is None and message.get("type") == "task_completed":
-            task_obj = message.get("task")
-            if isinstance(task_obj, dict):
-                candidate = task_obj.get("id")
-        if candidate is None:
-            return True
-        try:
-            return int(candidate) == self.task_id
-        except (TypeError, ValueError):
-            return False
 
 
 # -- Per-principal concurrency accounting --------------------------------
@@ -397,7 +416,8 @@ async def watchdog_check_once(
         raise
 
     status = snapshot.status
-    if status in _TERMINAL_STATUSES:
+    close_reason = _stream_close_reason(status, snapshot.control_state)
+    if close_reason == "terminal":
         return sink.enqueue_close(
             completed_frame(
                 status=status.value,
@@ -405,10 +425,7 @@ async def watchdog_check_once(
                 error=snapshot.error,
             )
         )
-    if (
-        status is TaskStatus.WAITING_FOR_USER
-        and snapshot.control_state != TaskControlState.RESUME_REQUESTED.value
-    ):
+    if close_reason == "input_required":
         return sink.enqueue_close(input_required_frame(task_id))
     if status is TaskStatus.PAUSED:
         # SDK wait() semantics: PAUSED keeps waiting, doesn't close.
@@ -435,8 +452,8 @@ async def _watchdog_loop(
     interval_seconds: float,
 ) -> None:
     """Runs every ``interval_seconds`` and also wakes early on a
-    completion hint from a terminal (completed or failed) broadcast:
-    same check, just run sooner.
+    completion hint from a terminal (completed or failed) or
+    waiting-for-user broadcast: same check, just run sooner.
 
     A single failed check (e.g. a transient DB error) must not end
     watchdog coverage for the rest of the stream's lifetime -- that
@@ -469,6 +486,15 @@ async def _watchdog_loop(
 
 
 # -- Response assembly ----------------------------------------------------
+
+# nginx buffers proxied responses by default; X-Accel-Buffering is the
+# per-response override that keeps SSE frames flowing immediately instead
+# of waiting for the proxy buffer to fill.
+_SSE_HEADERS = {"X-Accel-Buffering": "no", "Cache-Control": "no-cache"}
+
+
+def _sse_response(body: AsyncIterator[str]) -> StreamingResponse:
+    return StreamingResponse(body, media_type="text/event-stream", headers=_SSE_HEADERS)
 
 
 async def _terminal_snapshot_stream(
@@ -658,20 +684,14 @@ async def build_event_stream_response(
     per-principal reservation) ever exists, so a rejected attach never
     touches ``manager`` or the per-principal counter.
     """
-    if initial_snapshot.status in _TERMINAL_STATUSES:
-        return StreamingResponse(
-            _terminal_snapshot_stream(initial_snapshot),
-            media_type="text/event-stream",
-        )
+    close_reason = _stream_close_reason(
+        initial_snapshot.status, initial_snapshot.control_state
+    )
+    if close_reason == "terminal":
+        return _sse_response(_terminal_snapshot_stream(initial_snapshot))
 
-    if (
-        initial_snapshot.status is TaskStatus.WAITING_FOR_USER
-        and initial_snapshot.control_state != TaskControlState.RESUME_REQUESTED.value
-    ):
-        return StreamingResponse(
-            _input_required_snapshot_stream(initial_snapshot),
-            media_type="text/event-stream",
-        )
+    if close_reason == "input_required":
+        return _sse_response(_input_required_snapshot_stream(initial_snapshot))
 
     if count_task_sinks(task_id) >= PER_TASK_STREAM_CAP:
         raise V1ApiError(V1ErrorCode.RATE_LIMITED, 429)
@@ -680,7 +700,7 @@ async def build_event_stream_response(
     if not principal_slot_available(key_prefix):
         raise V1ApiError(V1ErrorCode.RATE_LIMITED, 429)
 
-    return StreamingResponse(
+    return _sse_response(
         _generate(
             task_id,
             principal,
@@ -690,6 +710,5 @@ async def build_event_stream_response(
             watchdog_interval_seconds=watchdog_interval_seconds,
             stream_max_duration_seconds=stream_max_duration_seconds,
             heartbeat_interval_seconds=heartbeat_interval_seconds,
-        ),
-        media_type="text/event-stream",
+        )
     )
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -47,6 +47,7 @@ from ...models.agent_api_key import AgentApiKey
 from ...models.database import get_session_local
 from ...models.task import TaskStatus
 from ...services.db_runtime import run_db_io_cancellation_safe
+from ...services.task_execution_controller import TaskControlState
 from ..websocket import _is_versioned_task_event, manager
 from .deps import ApiKeyPrincipal, active_runtime_key_filters
 from .errors import V1ApiError, V1ErrorCode
@@ -75,8 +76,13 @@ STREAM_MAX_DURATION_SECONDS = 60.0 * 60.0
 OUTBOUND_QUEUE_MAX_SIZE = 256
 PER_TASK_STREAM_CAP = 2
 PER_PRINCIPAL_STREAM_CAP = 32
+# Floor for the final wait when the 1-hour deadline is close: keeps the
+# queue wait from being called with a near-zero timeout, which would spin
+# the loop instead of actually waiting.
+_MIN_WAIT_BUDGET_SECONDS = 0.05
 
 _TERMINAL_STATUSES = (TaskStatus.COMPLETED, TaskStatus.FAILED)
+_TERMINAL_STATUS_VALUES = {status.value for status in _TERMINAL_STATUSES}
 
 # A sync callable: (task_id, principal) -> ``_TaskInfoSnapshot``-shaped
 # object (duck-typed: ``.status`` / ``.control_state`` / ``.output`` /
@@ -256,11 +262,27 @@ class V1EventStreamSink:
                 status = message.get("status")
                 if isinstance(status, str):
                     self.enqueue_status(status)
+                    if status in _TERMINAL_STATUS_VALUES:
+                        # A failed task's broadcast (``task_error``) never
+                        # carries ``type == "task_completed"``, so it would
+                        # otherwise miss the acceleration signal above and
+                        # wait out the watchdog's normal cadence.
+                        self.completion_hint.set()
         except Exception:
             self.dropped_frame_count += 1
             logger.exception(
                 "v1 SSE sink dropped a broadcast frame for task %s", self.task_id
             )
+
+    async def close(self, *args: Any, **kwargs: Any) -> None:
+        """No-op. Duck-types ``WebSocket.close`` -- task deletion
+        (``chat.py``'s ``_cleanup_runtime_state``) calls ``close()`` on every
+        connection still registered for the task, and a missing method here
+        would log a misleading "failed to close" warning for a sink that was
+        never a real socket in the first place. This generator's own
+        ``finally`` block is what actually tears the stream down.
+        """
+        return None
 
     def _binding_matches(self, message: dict[str, Any]) -> bool:
         """Drop frames whose task_id doesn't match this stream's binding
@@ -383,7 +405,7 @@ async def watchdog_check_once(
         )
     if (
         status is TaskStatus.WAITING_FOR_USER
-        and snapshot.control_state != "resume_requested"
+        and snapshot.control_state != TaskControlState.RESUME_REQUESTED.value
     ):
         return sink.enqueue_close(input_required_frame(task_id))
     if status is TaskStatus.PAUSED:
@@ -456,6 +478,18 @@ async def _terminal_snapshot_stream(
     yield completed_frame(
         status=snapshot.status.value, output=snapshot.output, error=snapshot.error
     )
+
+
+async def _input_required_snapshot_stream(
+    snapshot: "_TaskInfoSnapshot",
+) -> AsyncIterator[str]:
+    """Attach-time fast path for a task already waiting on user input (and
+    not mid-resume): emit ``task.status`` + ``task.input_required`` and
+    end. Same rationale as ``_terminal_snapshot_stream`` -- without this,
+    the same conclusion is only reached via the watchdog's first cycle,
+    up to ``watchdog_interval_seconds`` (30s in production) after attach."""
+    yield status_frame(snapshot.status.value)
+    yield input_required_frame(snapshot.task_id)
 
 
 async def _generate(
@@ -539,7 +573,19 @@ async def _generate(
                 # First-to-close-wins: a no-op if the watchdog already
                 # closed the stream first in the same tick.
                 sink.enqueue_close(error_frame("stream_expired"))
-            wait_budget = heartbeat_interval_seconds if remaining > 0 else 1.0
+            # Capped at the heartbeat interval so the deadline check above
+            # re-runs often enough, but never let it run past the deadline
+            # itself -- otherwise an idle stream whose last ping landed just
+            # before the deadline could sleep a full heartbeat interval past
+            # ``stream_max_duration_seconds`` before ``stream_expired`` is
+            # even enqueued.
+            wait_budget = (
+                min(
+                    heartbeat_interval_seconds, max(remaining, _MIN_WAIT_BUDGET_SECONDS)
+                )
+                if remaining > 0
+                else 1.0
+            )
             try:
                 frame_text, is_close = await asyncio.wait_for(
                     sink.queue.get(), timeout=wait_budget
@@ -609,6 +655,15 @@ async def build_event_stream_response(
     if initial_snapshot.status in _TERMINAL_STATUSES:
         return StreamingResponse(
             _terminal_snapshot_stream(initial_snapshot),
+            media_type="text/event-stream",
+        )
+
+    if (
+        initial_snapshot.status is TaskStatus.WAITING_FOR_USER
+        and initial_snapshot.control_state != TaskControlState.RESUME_REQUESTED.value
+    ):
+        return StreamingResponse(
+            _input_required_snapshot_stream(initial_snapshot),
             media_type="text/event-stream",
         )
 
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -150,7 +150,7 @@ _ERROR_MESSAGES = {
     "resync_required": (
         "The output queue overflowed; call steps() to resync, then re-attach."
     ),
-    "unauthorized": "The API key used to open this stream is no longer valid.",
+    "unauthorized": "The API key used to open this stream has been revoked or paused.",
     "task_deleted": "The task no longer exists.",
     "stream_expired": "This stream reached its maximum allowed duration.",
 }
@@ -263,12 +263,12 @@ class V1EventStreamSink:
     async def send_text(self, text: str) -> None:
         """Receive one broadcast frame. Duck-types ``WebSocket.send_text``.
 
-        Must never raise. ``broadcast_to_task`` (``websocket.py:4504-4532``)
+        Must never raise. ``ConnectionManager.broadcast_to_task``
         catches network errors from a connection's ``send_text`` and drops
-        the connection, but re-raises anything else (``:4525-4532``) --
-        that re-raise happens on the *task's own* outbound event path, so
-        an uncaught exception here would abort the broadcast for every
-        other listener on the task, not just this stream. This is the
+        the connection, but re-raises anything else -- that re-raise
+        happens on the *task's own* outbound event path, so an uncaught
+        exception here would abort the broadcast for every other
+        listener on the task, not just this stream. This is the
         one deliberate blanket ``except Exception`` in this module; every
         drop is logged via ``logger.exception`` below, and also counted
         on ``dropped_frame_count`` for tests to assert against (no
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -238,8 +238,10 @@ class V1EventStreamSink:
         that re-raise happens on the *task's own* outbound event path, so
         an uncaught exception here would abort the broadcast for every
         other listener on the task, not just this stream. This is the
-        one deliberate blanket ``except Exception`` in this module;
-        dropped frames are counted, not silent.
+        one deliberate blanket ``except Exception`` in this module; every
+        drop is logged via ``logger.exception`` below, and also counted
+        on ``dropped_frame_count`` for tests to assert against (no
+        production code reads that counter today).
         """
         try:
             self._assert_owner_loop()
@@ -433,7 +435,8 @@ async def _watchdog_loop(
     interval_seconds: float,
 ) -> None:
     """Runs every ``interval_seconds`` and also wakes early on a
-    ``task_completed`` broadcast hint: same check, just run sooner.
+    completion hint from a terminal (completed or failed) broadcast:
+    same check, just run sooner.
 
     A single failed check (e.g. a transient DB error) must not end
     watchdog coverage for the rest of the stream's lifetime -- that
@@ -538,13 +541,16 @@ async def _generate(
         # concurrent burst of attaches for the same principal all pass the
         # earlier read-only check and then land here before any of them
         # releases a slot, `try_reserve_principal_slot` returns False for
-        # the late arrivals: the count is briefly over `PER_PRINCIPAL_STREAM_CAP`,
-        # it's logged, and the stream is served anyway. The sentinel
-        # (`principal_slot_reserved`) stays False for these streams, so the
-        # `finally` below correctly skips `release_principal_slot` for them
-        # -- nothing was reserved, so nothing is released. The count
-        # self-heals as soon as any concurrently-open stream for this
-        # principal finishes and releases its own slot.
+        # the late arrivals: the number of open streams for this principal
+        # can briefly exceed `PER_PRINCIPAL_STREAM_CAP` -- the counter
+        # itself never does, since `try_reserve_principal_slot` checks
+        # before it increments -- it's logged, and the stream is served
+        # anyway. The sentinel (`principal_slot_reserved`) stays False for
+        # these streams, so the `finally` below correctly skips
+        # `release_principal_slot` for them -- nothing was reserved, so
+        # nothing is released. The open-stream count self-heals as soon as
+        # any concurrently-open stream for this principal finishes and
+        # releases its own slot.
         principal_slot_reserved = try_reserve_principal_slot(key_prefix)
         if not principal_slot_reserved:
             logger.warning(
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -489,7 +489,29 @@ async def _generate(
     principal_slot_reserved = False
     watchdog_task: "asyncio.Task[None] | None" = None
     try:
+        # Soft cap: `principal_slot_available` in `build_event_stream_response`
+        # already did the *check* before this generator started, and that
+        # check is what carries the normal 429 -- an attach that loses the
+        # race here has already been told "yes, come in", so this
+        # reservation only accounts for capacity, it never rejects. If a
+        # concurrent burst of attaches for the same principal all pass the
+        # earlier read-only check and then land here before any of them
+        # releases a slot, `try_reserve_principal_slot` returns False for
+        # the late arrivals: the count is briefly over `PER_PRINCIPAL_STREAM_CAP`,
+        # it's logged, and the stream is served anyway. The sentinel
+        # (`principal_slot_reserved`) stays False for these streams, so the
+        # `finally` below correctly skips `release_principal_slot` for them
+        # -- nothing was reserved, so nothing is released. The count
+        # self-heals as soon as any concurrently-open stream for this
+        # principal finishes and releases its own slot.
         principal_slot_reserved = try_reserve_principal_slot(key_prefix)
+        if not principal_slot_reserved:
+            logger.warning(
+                "v1 SSE per-principal cap best-effort exceeded for "
+                "key_prefix=%s under concurrent attach burst; serving the "
+                "stream anyway (soft cap, no reservation held)",
+                key_prefix,
+            )
         manager.register_connection(cast(WebSocket, sink), task_id)
         watchdog_task = asyncio.create_task(
             _watchdog_loop(
@@ -527,26 +549,35 @@ async def _generate(
             if is_close:
                 return
     finally:
-        if watchdog_task is not None:
-            watchdog_task.cancel()
-            # Narrowed to ``CancelledError`` only (not a blanket
-            # ``BaseException``): the watchdog loop itself now catches
-            # and logs every per-cycle ``Exception`` internally (see
-            # ``_watchdog_loop``) and retries rather than dying, so the
-            # only exception this teardown should ever observe here is
-            # the cancellation just requested above. If the watchdog
-            # task raises anything else, that's a real bug and must
-            # propagate instead of being silently swallowed.
-            with contextlib.suppress(asyncio.CancelledError):
-                await watchdog_task
-        # ``manager`` is typed against a real ``WebSocket``; this sink only
-        # duck-types its ``send_text`` contract (module docstring) -- the
-        # cast documents that intentional narrowing instead of suppressing
-        # the type error blanket-wide. A safe no-op if registration above
-        # never ran or never completed.
-        manager.disconnect(cast(WebSocket, sink))
-        if principal_slot_reserved:
-            release_principal_slot(key_prefix)
+        # The watchdog wait-and-cancel is wrapped in its own try/finally so
+        # that `manager.disconnect` and `release_principal_slot` below --
+        # the two calls that actually undo what this generator reserved --
+        # still run even if awaiting the cancelled watchdog task raises
+        # something unexpected. Without this inner `finally`, a raise here
+        # would skip straight past both cleanup calls, leaking the sink
+        # registration and (if held) the per-principal slot.
+        try:
+            if watchdog_task is not None:
+                watchdog_task.cancel()
+                # Narrowed to ``CancelledError`` only (not a blanket
+                # ``BaseException``): the watchdog loop itself now catches
+                # and logs every per-cycle ``Exception`` internally (see
+                # ``_watchdog_loop``) and retries rather than dying, so the
+                # only exception this teardown should ever observe here is
+                # the cancellation just requested above. If the watchdog
+                # task raises anything else, that's a real bug and must
+                # propagate instead of being silently swallowed.
+                with contextlib.suppress(asyncio.CancelledError):
+                    await watchdog_task
+        finally:
+            # ``manager`` is typed against a real ``WebSocket``; this sink only
+            # duck-types its ``send_text`` contract (module docstring) -- the
+            # cast documents that intentional narrowing instead of suppressing
+            # the type error blanket-wide. A safe no-op if registration above
+            # never ran or never completed.
+            manager.disconnect(cast(WebSocket, sink))
+            if principal_slot_reserved:
+                release_principal_slot(key_prefix)
 
 
 async def build_event_stream_response(
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -10,14 +10,22 @@ and nothing else.
 
 Explicitly out of scope here:
   - Projecting ``step.*`` / ``message.*`` content from trace events.
-  - Buffering live frames during attach warm-up. Without that buffer,
-    a live status update emitted between sink registration and the
-    first ``task.status`` frame can arrive out of order or duplicate
-    the first frame; this is accepted. Worst case the client sees one
-    extra, stale ``task.status``. That's harmless because the
-    authoritative terminal/close determination always comes from the
-    watchdog reading the task row (or the attach-time snapshot read),
-    never from frame ordering.
+  - Buffering live frames during attach warm-up, or reconciling frame
+    order at all. No ``task.status`` frame is ever guaranteed to be
+    fresh or in order: ``ConnectionManager.broadcast_to_task`` stamps
+    each event with whatever ``run_id`` / ``state_version`` tuple its
+    producer captured, falling back to a fresh read of the task row
+    (status included) when the producer didn't capture one, and this
+    sink never compares that stamp against anything -- it just forwards
+    each status string it sees, deduped only against the last one it
+    sent. A stale or out-of-order
+    ``task.status`` frame can therefore reach the client at any point
+    during the stream's life, not only right after attach. This is
+    accepted: the only frames this module treats as authoritative are
+    the three close frames -- ``task.completed``, ``task.input_required``,
+    and ``stream.error`` -- and all three are produced by the watchdog
+    (or the attach-time snapshot read) reading the task row directly,
+    never inferred from frame ordering.
   - Populating ``task.input_required``'s ``prompt`` field from the
     agent's question text. Only the watchdog closes the stream on a
     task stuck waiting for user input (within one watchdog cycle), and
--- a/src/xagent/web/api/v1/_events_stream.py
+++ b/src/xagent/web/api/v1/_events_stream.py
@@ -1,0 +1,526 @@
+"""SSE transport layer for ``GET /v1/chat/tasks/{task_id}/events``.
+
+This module owns everything about *moving bytes to the client and
+deciding when to stop* -- registration into the shared connection
+``manager``, the outbound frame queue, the 30-second watchdog, the
+1-hour absolute duration cap, and per-task / per-principal concurrency
+limits. It emits three lifecycle events -- ``task.status``,
+``task.completed``, and ``stream.error`` -- and nothing else.
+
+Explicitly out of scope here:
+  - Projecting ``step.*`` / ``message.*`` content from trace events.
+  - Buffering live frames during attach warm-up. Without that buffer,
+    a live status update emitted between sink registration and the
+    first ``task.status`` frame can arrive out of order or duplicate
+    the first frame; this is accepted. Worst case the client sees one
+    extra, stale ``task.status``. That's harmless because the
+    authoritative terminal/close determination always comes from the
+    watchdog reading the task row (or the attach-time snapshot read),
+    never from frame ordering.
+  - Populating ``task.input_required``'s ``prompt`` field from the
+    agent's question text. Only the watchdog closes the stream on a
+    task stuck waiting for user input (within one watchdog cycle), and
+    it always sends ``prompt: null`` -- question-text sniffing from
+    live frames belongs to the content-projection layer, not this
+    transport layer.
+
+Sink instances duck-type the ``websocket.ConnectionManager`` connection
+contract (an object with an async ``send_text(str)`` method) so they
+register into the *same* shared ``manager`` real WebSocket connections
+use, and ride the same ``broadcast_to_task`` fan-out.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import json
+import logging
+from time import monotonic
+from typing import TYPE_CHECKING, Any, AsyncIterator, Callable, cast
+
+from fastapi import WebSocket
+from fastapi.responses import StreamingResponse
+
+from ...models.agent_api_key import AgentApiKey
+from ...models.database import get_session_local
+from ...models.task import TaskStatus
+from ...services.db_runtime import run_db_io_cancellation_safe
+from ..websocket import _is_versioned_task_event, manager
+from .deps import ApiKeyPrincipal, active_runtime_key_filters
+from .errors import V1ApiError, V1ErrorCode
+
+if TYPE_CHECKING:
+    # Only for the type checker -- importing ``tasks`` at module scope
+    # would cycle back here (``tasks.py`` imports this module to wire the
+    # endpoint). Callers pass a snapshot-reading callable at call time
+    # instead (see ``TaskSnapshotReader`` below).
+    from .tasks import _TaskInfoSnapshot
+
+logger = logging.getLogger(__name__)
+
+# -- Tunables (all injectable per-call for tests; production always uses
+# the defaults below via ``tasks.py``). ---------------------------------
+
+HEARTBEAT_INTERVAL_SECONDS = 15.0
+WATCHDOG_INTERVAL_SECONDS = 30.0
+# Same shape/value as A2A's cap (``web/api/a2a.py:105``); a separate v1
+# constant because the two streams' generators are structurally different
+# (A2A polls every <=0.5s and re-checks the deadline for free; v1 is
+# queue-consumption-based, so each wait on the outbound queue must be
+# capped at the heartbeat interval so the deadline gets re-checked often
+# enough).
+STREAM_MAX_DURATION_SECONDS = 60.0 * 60.0
+OUTBOUND_QUEUE_MAX_SIZE = 256
+PER_TASK_STREAM_CAP = 2
+PER_PRINCIPAL_STREAM_CAP = 32
+
+_TERMINAL_STATUSES = (TaskStatus.COMPLETED, TaskStatus.FAILED)
+
+# A sync callable: (task_id, principal) -> ``_TaskInfoSnapshot``-shaped
+# object (duck-typed: ``.status`` / ``.control_state`` / ``.output`` /
+# ``.error``), raising ``V1ApiError(TASK_NOT_FOUND, 404)`` when the task
+# is missing or not owned. Always run through ``run_db_io_cancellation_safe``
+# by this module -- never called directly.
+TaskSnapshotReader = Callable[[int, "ApiKeyPrincipal"], Any]
+
+
+# -- Wire format --------------------------------------------------------
+
+
+def _sse_frame(event: str, data: dict[str, Any]) -> str:
+    return f"event: {event}\ndata: {json.dumps(data, separators=(',', ':'))}\n\n"
+
+
+def status_frame(status: str) -> str:
+    return _sse_frame("task.status", {"status": status})
+
+
+def completed_frame(*, status: str, output: str | None, error: str | None) -> str:
+    return _sse_frame(
+        "task.completed", {"status": status, "output": output, "error": error}
+    )
+
+
+def input_required_frame(task_id: int) -> str:
+    # ``prompt`` is always null: populating it from the agent's question
+    # text is the content-projection layer's job, not this transport.
+    return _sse_frame("task.input_required", {"task_id": task_id, "prompt": None})
+
+
+_ERROR_MESSAGES = {
+    "resync_required": (
+        "The output queue overflowed; call steps() to resync, then re-attach."
+    ),
+    "unauthorized": "The API key used to open this stream is no longer valid.",
+    "task_deleted": "The task no longer exists.",
+    "stream_expired": "This stream reached its maximum allowed duration.",
+}
+
+
+def error_frame(code: str) -> str:
+    return _sse_frame("stream.error", {"code": code, "message": _ERROR_MESSAGES[code]})
+
+
+# -- Sink -----------------------------------------------------------------
+
+
+class V1EventStreamSink:
+    """One SSE consumer's broadcast-frame filter and outbound queue.
+
+    Bound to exactly one ``task_id`` and one ``principal`` for the life
+    of the connection. The principal is used only for quota accounting
+    and the watchdog's key-validity check -- it is **not** an
+    authorization gate for who may act on the task; that check already
+    happened once, at attach time, via ``get_principal_from_api_key`` +
+    ``_resolve_task_or_404``.
+    """
+
+    def __init__(
+        self, *, task_id: int, principal_key_prefix: str, initial_status: str
+    ) -> None:
+        self.task_id = task_id
+        self.principal_key_prefix = principal_key_prefix
+        self.dropped_frame_count = 0
+        self.completion_hint = asyncio.Event()
+        self._queue: "asyncio.Queue[str]" = asyncio.Queue(
+            maxsize=OUTBOUND_QUEUE_MAX_SIZE
+        )
+        self._closing = False
+        self._last_status = initial_status
+        # Recorded at construction time (always inside the endpoint's own
+        # request coroutine) so later state-mutating calls -- however they
+        # got here -- can be asserted to run on the same loop.
+        self._owner_loop = asyncio.get_running_loop()
+
+    @property
+    def closing(self) -> bool:
+        return self._closing
+
+    @property
+    def queue(self) -> "asyncio.Queue[str]":
+        return self._queue
+
+    def _assert_owner_loop(self) -> None:
+        if asyncio.get_running_loop() is not self._owner_loop:
+            raise RuntimeError("v1 SSE sink touched from a foreign event loop")
+
+    def enqueue_status(self, status: str) -> None:
+        """Enqueue a deduped ``task.status`` frame. Never closes the stream."""
+        self._assert_owner_loop()
+        if self._closing or status == self._last_status:
+            return
+        self._last_status = status
+        self._put_or_overflow(status_frame(status))
+
+    def enqueue_close(self, frame_text: str) -> bool:
+        """Close exactly once; the first caller wins.
+
+        Drains any queued backlog before inserting the close frame so
+        the close frame always has room -- even when called *because*
+        the queue just overflowed. Losing unread backlog on close is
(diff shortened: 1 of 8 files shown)


rogercloud (changes requested):
## Summary

This PR adds `GET /v1/chat/tasks/{task_id}/events`, an SSE stream of task lifecycle events (`task.status`, `task.completed`, `task.input_required`, `stream.error`), so SDK clients get immediate feedback instead of silently polling for the first seconds of a task. The transport is built around a 30-second watchdog on the task row as the terminal authority, a bounded 256-frame queue with `resync_required` overflow, a 1-hour absolute deadline, 15-second heartbeats, and per-task/per-key attach caps, backed by a substantial test suite (38 transport tests; CI is green, including the full `tests/web/api/v1` suite). One blocking interaction with the pre-existing durable command executor can silently swallow personal messages; the remaining findings are minor documentation and test-hardening items. Blocking: yes — recommended event: REQUEST_CHANGES

## Approach verdict

The watchdog-authority design is sound. What verifies clean: close-exactly-once under watchdog/deadline races, in-flight frames never displacing the close frame, deferred registration/reservation teardown that leaks nothing on disconnect or never-started streams, and fail-closed auth (401/404 before the stream ever opens). The broadcast layer is correctly an acceleration mechanism rather than authority, and the sink's exception-swallowing is the right call for the broadcast path. Reservations concentrate in three places: (a) the sink's registration in the shared connection manager (the blocking issue), (b) several docstring and PR-body claims that overstate what the code does, and (c) a few coverage gaps on the real-route path.

## Findings

### Major

**1. The SSE sink silently swallows personal messages routed to `connections[0]` by the durable command executor** — `src/xagent/web/api/v1/_events_stream.py:515` registers `V1EventStreamSink` into the same `ConnectionManager` list as real chat WebSockets via `cast(WebSocket, sink)`. The durable command executor (`src/xagent/web/api/websocket.py:7830`) resolves `connections = manager.connections_for_task(command.task_id)` and takes `connections[0]` (`src/xagent/web/api/websocket.py:7841-7842`), falling back to `_DiscardingCommandWebSocket` only when the list is empty. These paths serve the normal live pause/resume flow (`dispatch_task_command_promptly` at websocket.py:7323/7573) and chat message delivery (websocket.py:4959). A sink attached before the UI socket occupies index 0, and `V1EventStreamSink.send_text` (`_events_stream.py:219-256`) implicitly discards anything that is not a versioned event with a str status — so pause/resume validation errors ("Task not found or access denied", "Task is not paused and cannot be resumed", "Task resume is already in progress"), stale-run rejections, and `task_id_updated` (websocket.py:5751-5758) are silently dropped while a real UI socket is connected at index >= 1. The `task_command_accepted` ack still reaches the issuing socket, commands still execute and persist, and success broadcasts still fan out — which is exactly why this is easy to miss. Suggested fix: mark sinks broadcast-only (an `is_broadcast_only` flag or a typed registry in `ConnectionManager`) and have the command executor select a real WebSocket; `_DiscardingCommandWebSocket` only covers the empty-list case.

### Minor

**2.** `src/xagent/web/api/chat.py:4805-4813` — the delete path calls `await connection.close()` on every detached connection; a detached sink has no `close()` method, so deleting an SDK-source task with a live stream (the delete path has no source filter) logs a spurious "Failed to close WebSocket connection" warning. The sink detaches early and the watchdog closes the stream with `task_deleted` within <=30s, so this is noise only. Fix: a no-op `close()` on the sink, or skip non-WebSocket connections in the loop.

**3.** `_events_stream.py:530-538` — when idle, `wait_budget` is the full heartbeat whenever `remaining > 0`, never `min(heartbeat, remaining)`, so an idle stream can run up to one heartbeat (15s) past the documented "1-hour maximum" (`tasks.py:1231`). Fix: `min(heartbeat_interval_seconds, max(remaining, epsilon))`, or document the tolerance.

**4.** `_events_stream.py:360-362` + `src/xagent/web/api/v1/tasks.py:789-792` — the watchdog reuses the heavy `_load_task_info_snapshot`; on the waiting path it runs `get_latest_waiting_question` whose result is discarded (the prompt is always null). Normally only 1-2 wasted queries (the stream closes on the first waiting check); it repeats only in `resume_requested` limbo. Negligible; a slim reader is optional.

**5.** `_events_stream.py:379` — the magic string `"resume_requested"` should be `TaskControlState.RESUME_REQUESTED.value` (task_execution_controller.py:30; the canonical spelling used at a2a.py:225 and websocket.py:227) to prevent drift.

**6.** `_events_stream.py:499-506` (and the PR body) — "the count is briefly over `PER_PRINCIPAL_STREAM_CAP`" is imprecise: `try_reserve_principal_slot` never increments past the cap; what exceeds the cap is the number of open streams (unreserved late arrivals). The code is correct; the wording is wrong.

**7.** `tests/web/api/v1/test_events_stream.py:443-447` — the signature-introspection test asserts `set(sig.parameters) == {"task_id", "principal"}`; it fails on benign refactors (e.g. a renamed parameter) and would still pass if the endpoint opened its own request-scoped session (the dominant pattern in tasks.py), so it is weak evidence for the "no request-scoped DB session" claim. Prefer a behavioral assertion or drop the test.

**8.** `_events_stream.py:145/252-256` — `dropped_frame_count` is never read outside tests (each drop is logged via `logger.exception`), so the "counted, not silent" claim has no production observer. Add a teardown log/metric or drop the claim.

**9.** `_events_stream.py:602-606` — no `WAITING_FOR_USER` attach fast path: a client attaching to an already-waiting task gets `task.status` and then waits up to 30s (the loop sleeps a full interval before its first check at :426-431) for the `task.input_required` close, although the attach snapshot already knows the status. The terminal fast path exists; the waiting one doesn't.

**10.** `tasks.py:1221-1222` — "ping sent every 15s" is only true after 15s of idle (the queue-get timeout); no pings fire while frames flow. Reword.

**11.** `tasks.py:1211-1233` — the docstring names the four events but omits the JSON payload shapes, contradicting the PR body's "documents exactly what it emits". Add them: `task.status {status}`, `task.completed {status, output, error}`, `task.input_required {task_id, prompt: null}`, `stream.error {code, message}`.

**12.** `_events_stream.py:100-103/122-123` — `task.completed` carries `status: "failed"` for FAILED tasks (the event name suggests success), and `stream.error` is a flat `{code, message}` vs the /v1 HTTP envelope `{"error": {code, message}}`, with a disjoint code vocabulary (`unauthorized`/`task_deleted`/`resync_required`/`stream_expired` vs `V1ErrorCode`). Both choices are reasonable; document them in the docstring.

**13.** Test gaps — (a) no real-route streaming test with an aborted client (disconnect is simulated via `body_iterator.aclose()`); (b) no pin that a `task_completed` broadcast emits no `task.status` frame; (d) no attach-time route test with initial status `WAITING_FOR_USER` or `PAUSED`. (The non-`TASK_NOT_FOUND` `V1ApiError` branch is unreachable: `read_task_snapshot` only raises `TASK_NOT_FOUND`.)

**14.** Design note — this is a second parallel SSE transport alongside a2a.py: the 1-hour cap constant is duplicated by value with a cross-reference comment, and the loops are structurally different (queue-based vs 0.5s polling). Sharing is not recommended; flagging so the duplication stays a conscious decision.

**15.** `_events_stream.py:602-626` — no `Cache-Control: no-cache` / `X-Accel-Buffering: no` headers; buffering proxies may hold chunks. Matches the pre-existing a2a.py pattern (not a regression).

**16.** `tasks.py:1254-1260` — the docstring states the per-task cap ("more than 2 concurrent streams already open on this task") absolutely while the best-effort wording applies only to the per-key cap, yet the per-task cap has the same TOCTOU (checked at response-build time, registered later in `_generate`), and both caps are per-process (multi-worker deploys exist, web/README.md:165). The PR body itself says both are best-effort. Align the docstring.

**17.** Design note — each per-stream watchdog opens 2 short DB sessions per 30s cycle (~240 sessions/stream/h; 10k streams ~660 qps). Indexed point lookups; informational only.

**18.** Design note — on the failure path, settlement broadcasts `task_error`, which the sink enqueues immediately as `task.status {status: failed}` (so the client learns of the failure instantly), but only `task_completed` broadcasts set `completion_hint`; the authoritative `task.completed` close therefore waits up to the next 30s watchdog tick. Consider treating terminal-status broadcasts as a watchdog wake-up, or document the asymmetry.

- L258 `src/xagent/web/api/v1/_events_stream.py`: `delete:` `_binding_matches` — the manager already scopes delivery per task (`connections_for_task` plus an `is_connection_registered` re-check before each `send_text`, websocket.py:4516-4518), and `move_connection` is never called on sinks (only real chat WebSockets, websocket.py:5750). Delete the method, its call site (:238), and `test_sink_discards_foreign_task_frames` (test_events_stream.py:778-784).
- L79 `tests/web/api/v1/test_events_stream.py`: `delete:` `_create_task`'s `content` parameter — all 19 call sites pass two arguments; inline the "hello" default.

Net: approximately 26 lines removed.

Blocking: **yes** — one issue blocks merge:

- [new] `src/xagent/web/api/websocket.py:7841-7842` x `src/xagent/web/api/v1/_events_stream.py:515` — **major** — the SSE sink occupies slot 0 of the shared per-task connection list and silently swallows durable-command personal messages (pause/resume validation errors, `task_id_updated`, chat messages) while a real UI socket is connected.

Recommended event: **REQUEST_CHANGES**. The remaining findings are minors — docstring accuracy, magic-string hygiene, and test hardening — and can be addressed in a follow-up, but the `connections[0]` interaction is a real behavioral regression for the normal pause/resume flow whenever a stream attaches before the UI socket.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,627 @@
...
+        # principal finishes and releases its own slot.
+        principal_slot_reserved = try_reserve_principal_slot(key_prefix)
+        if not principal_slot_reserved:
+            logger.warning(
+                "v1 SSE per-principal cap best-effort exceeded for "
+                "key_prefix=%s under concurrent attach burst; serving the "
+                "stream anyway (soft cap, no reservation held)",
+                key_prefix,
+            )
+        manager.register_connection(cast(WebSocket, sink), task_id)
```
rogercloud, line 596: Sink registration concern: `V1EventStreamSink` joins the same `ConnectionManager` list as real chat WebSockets, and the durable command executor picks `connections[0]` for personal messages (`src/xagent/web/api/websocket.py:7841-7842`), falling back to `_DiscardingCommandWebSocket` only when the list is empty. A sink attached before the UI socket occupies index 0, and `send_text` (`_events_stream.py:219-256`) silently discards anything that is not a versioned status event — so pause/resume validation errors, `task_id_updated`, and chat messages are dropped while a real UI socket is connected at index >= 1, with no log. Commands still execute and persist, and success broadcasts still fan out, which makes this easy to miss. Suggest marking sinks broadcast-only (e.g. an `is_broadcast_only` flag or a typed registry in `ConnectionManager`) and having the command executor select a real WebSocket.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,627 @@
...
+        return sink.enqueue_close(
+            completed_frame(
+                status=status.value,
+                output=snapshot.output,
+                error=snapshot.error,
+            )
+        )
+    if (
+        status is TaskStatus.WAITING_FOR_USER
+        and snapshot.control_state != "resume_requested"
```
rogercloud: Magic string: `snapshot.control_state != "resume_requested"` — the canonical spelling is `TaskControlState.RESUME_REQUESTED.value` (task_execution_controller.py:30), used at a2a.py:225 and websocket.py:227. Use the enum value to prevent drift.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,627 @@
...
+        # closed (``aclose()``, e.g. on client disconnect) while suspended
+        # here still has to unregister the sink and cancel the watchdog.
+        yield status_frame(initial_status)
+        while True:
+            remaining = deadline - monotonic()
+            if remaining <= 0:
+                # First-to-close-wins: a no-op if the watchdog already
+                # closed the stream first in the same tick.
+                sink.enqueue_close(error_frame("stream_expired"))
+            wait_budget = heartbeat_interval_seconds if remaining > 0 else 1.0
```
rogercloud: Deadline overshoot: when idle, `wait_budget` is the full heartbeat whenever `remaining > 0`, never `min(heartbeat, remaining)`, so an idle stream can run up to one heartbeat (15s) past the documented "1-hour maximum" (`tasks.py:1231`). Use `min(heartbeat_interval_seconds, max(remaining, epsilon))` or document the tolerance.

tests/web/api/v1/test_events_stream.py
```diff
@@ -0,0 +1,1002 @@
...
+
+
+# ===== endpoint signature carries no request-scoped DB dependency =====
+
+
+def test_events_endpoint_signature_has_no_db_dependency():
+    import inspect
+
+    sig = inspect.signature(v1_tasks.stream_chat_task_events)
+    assert set(sig.parameters) == {"task_id", "principal"}
```
rogercloud: This signature-introspection assertion is brittle — it fails on benign refactors (e.g. a renamed parameter) and would still pass if the endpoint opened its own request-scoped session (the dominant pattern in tasks.py) — so it is weak evidence for the "no request-scoped DB session" claim. Prefer a behavioral assertion or drop the test.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,74 @@ async def get_chat_task_steps(
...
+        principal: Resolved by the auth dependency; also used (not as
+            an authorization gate) for the per-key concurrency count
+            and the periodic key-validity check.
+
+    Raises:
+        V1ApiError 401: missing / invalid / revoked key (plain JSON;
+            the stream never opens).
+        V1ApiError 404: task missing or not owned by the calling key
+            (plain JSON; the stream never opens).
+        V1ApiError 429 ``rate_limited``: more than 2 concurrent streams
```
rogercloud: Docstring overstatement: the per-task cap is stated absolutely ("more than 2 concurrent streams already open on this task") while the best-effort wording applies only to the per-key cap, yet the per-task cap has the same TOCTOU (checked at response-build time, registered later in `_generate`) and both caps are per-process (multi-worker deploys exist, web/README.md:165). The PR body itself says both are best-effort; align the docstring.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Fixed at 8234a8909, on the executor side rather than in the manager. `dispatch_durable_command` now picks the first connection that is not `is_broadcast_only`, falling back to `_DiscardingCommandWebSocket()`; `V1EventStreamSink` sets that class attribute. All four personal-reply payloads route through that one variable, so the resume-already-in-progress rejection (the one with no second delivery channel) now reaches a real socket.

Not adopting the typed registry in `ConnectionManager`: registering the sink into the shared manager was a deliberate choice with an existing precedent, and reshaping the manager would pull a subsystem-wide refactor into a transport PR. The attribute check is deliberately `getattr(..., False)` rather than an isinstance test, so the several suites that register plain stand-in objects keep their existing behavior (121 tests across those files re-run unchanged). Worth noting the underlying assumption that `connections[0]` is the issuing socket is wrong independently of this PR — two browser tabs on one task already misroute the reply — but this removes the case this PR would have made easy to hit. Three tests cover broadcast-only-first, all-broadcast-only, and empty-list.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Fixed at 8234a8909 — now compares against `TaskControlState.RESUME_REQUESTED.value`. No import cycle.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Fixed at 8234a8909: `min(heartbeat_interval_seconds, max(remaining, _MIN_WAIT_BUDGET_SECONDS))`. Confirmed a real overshoot rather than a wording issue — at 3599.9s in, the old code waited a full heartbeat and expired at ~3615s. Pinned by a new test with the heartbeat deliberately larger than the remaining budget (heartbeat 5.0s, deadline 0.05s, drained under `wait_for(timeout=1)`); the pre-existing deadline test could not catch it because its heartbeat was already the smaller of the two.

tests/web/api/v1/test_events_stream.py
AlexLiu190625: Rewrote it at 8234a8909 rather than dropping it: the assertion is now that no parameter carries a `Session` annotation, which is the property the test name claims and survives a rename.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Fixed at 8234a8909. Both caps now read as best-effort under the same check-then-register race, and the wording separates two things that were conflated: the counter never exceeds the cap (`try_reserve_principal_slot` checks before incrementing), while the number of open streams can, because a late arrival that loses the race is served without a reservation.

AlexLiu190625:
Addressed in the commits just pushed to this branch. Per-thread rationale is on each inline thread; summary of the body-level findings below.

**Fixed (12).** Blocking `connections[0]` selection (1); sink `close()` so the delete path stops logging a spurious warning (2); deadline overshoot (3); enum value instead of the literal (5); cap wording that conflated "counter exceeds" with "open streams exceed" (6, 16); brittle signature test rewritten as an annotation check (7); dropped-frame docstring now says drops are logged and the counter is test-only (8); waiting-for-user attach fast path, so an already-waiting task closes at attach instead of up to 30s later (9); ping cadence wording (10); the four events' JSON field lists, which is the only schema the endpoint exposes (11); failed-task-ends-via-`task.completed` and the flat `stream.error` shape both documented (12); terminal broadcasts now set the completion hint, so a failure closes on the next wake rather than the next 30s tick (18). Tests added for 1, 3, 9, 13(b), 13(d), 18.

**Deferred (3).** 13(a) real-client-disconnect streaming test — no `client.stream(` precedent in this repo, and a test client that drains the generator before the client sees a byte would go green without proving disconnect handling; worth doing as its own test-infrastructure change. 15 anti-buffering headers — a pre-existing gap shared with the a2a stream, better fixed across both surfaces at once. 12's second half, aligning the stream error vocabulary with `V1ErrorCode` — that is a published-contract change rather than a doc fix, so it should not ride along here; the disjointness is now documented instead.

**Not adopting (4).**

4 — the extra `get_latest_waiting_question` only runs when the task is already `WAITING_FOR_USER`, and that same watchdog cycle closes the stream, so it fires at most once per stream. A slim column-only reader is worth writing when the projection layer actually populates `prompt`; today it would be a second snapshot reader with no consumer.

14 — agreed on not sharing, keeping the cross-reference comment. The two caps matching in value is a coincidence rather than a shared contract: A2A's sits on a version-gated external protocol surface, v1's on our own SDK surface, so pointing them at one constant would let a change to one silently move the other.

17 — confirmed: 2 sessions per 30s cycle, both indexed single-row lookups. Not merging them, because the two reads are deliberately ordered key-first so the auth axis fails closed, and merging would also make the transport depend on the endpoint's session instead of the injected reader. If concurrency ever justifies it, the lever is a longer watchdog interval or a short key-validity cache.

Simplification A — the manager-scoping and `move_connection` facts are both correct, and `_binding_matches` is unreachable defense today. Keeping it: registering the sink into the shared manager rather than a separate registry was accepted specifically in exchange for that binding check, so removing the check while also asking for a separate registry would reopen the trade-off from both ends. It also guards a different failure than the new `is_broadcast_only` flag — that one prevents being mistaken for a personal-reply target, this one prevents receiving another task's frames after a rebind.

Simplification B — correct that no call site here passes `content`. Keeping the parameter: this helper is a verbatim copy of the one in `test_tasks.py`, where 8 call sites use it, and dropping it here would let the two copies drift. The real cleanup is hoisting the shared helper into `conftest.py`, which is its own test refactor.

rogercloud (changes requested):
## Summary

This PR adds a server-sent-events channel (`GET /v1/chat/tasks/{task_id}/events`) so SDK clients can learn a task is alive the moment they attach and when it ends, closing the blind window left by the 1 Hz `steps()` polling (until the first step lands there is nothing to observe). The stream rides the existing WebSocket broadcast fan-out through a new transport-only sink, with the 30 s DB watchdog as the single authority for terminal state, bounded resources (2 streams/task, 32/key, 256-frame queue, 1 h lifetime), and graceful degradation — broadcasts only accelerate, the watchdog guarantees. The design was judged sound in Round 0 and remains sound; the one new blocker is deployment-level buffering in the shipped nginx configuration that defeats the endpoint's core "alive at attach" promise.

## Update since last review

Six commits landed since the previous review: 597b9fdf (per-principal cap reworded to best-effort), d37f9f2c (durable command router skips broadcast-only sinks — the fix for prior finding 1), c0890873 (lifecycle/timing gaps: no-op sink close, terminal-status enum, wait-budget floor, WAITING_FOR_USER attach fast path, terminal broadcast completion hint), 894cf1c7 (docstring and test corrections), 1fdb1588 (test: terminal broadcasts wake the watchdog), and 8234a890 (test: deadline wait budget shrinks below the heartbeat). CI is green across all checks. Of the 18 prior findings, 14 are FIXED, 2 are WAIVED, 1 is PARTIAL (P7), and 1 (P15) is superseded by the re-raised N1; the carried S1 simplification remains NOT FIXED.

## Approach verdict

**Sound.** The design model holds: a transport-only `_events_stream.py` riding the existing WebSocket broadcast fan-out, the sink duck-typing the WebSocket contract and registering in the shared ConnectionManager, the watchdog as the single terminal-state authority, bounded resources with graceful degradation, and an explicit close frame that is always terminal. The sink swallows exceptions on the broadcast path by design; every failure mode degrades to the watchdog cadence rather than a hung stream.

**N2 — design-level (minor, severity-adjusted from major).** The sink re-parses the serialized WebSocket wire format (`json.loads` at `src/xagent/web/api/v1/_events_stream.py:250`), imports the private `_is_versioned_task_event` from `websocket.py` (:51), and depends on enrichment-stamped top-level `status`/`task_id` — an implicit, unenforced contract with another transport's wire format. This is not a current bug: the sink calls the very same predicate object, both halves are individually pinned by tests, and the watchdog (DB authority) bounds any future drift. Minimum alternative: hoist `_VERSIONED_TASK_EVENT_TYPES`/`_is_versioned_task_event` plus the stamping into a non-underscore shared module, and add one integration test wiring `manager.broadcast_to_task` → sink → generator.

## Prior findings checklist

1. **FIXED** — `connections[0]` sink swallowed personal replies: websocket.py:7841–7851 now selects the first non-broadcast-only connection with `_DiscardingCommandWebSocket()` as fallback; the sink sets `is_broadcast_only = True` (_events_stream.py:151); three tests cover broadcast-only-first / all-broadcast-only / empty; repo-wide grep shows no remaining `connections[0]` indexing.
2. **FIXED** — delete-path `close()` warning: the sink's no-op async `close()` at _events_stream.py:279–287; the chat.py delete path is unchanged.
3. **FIXED** — deadline overshoot: `wait_budget = min(heartbeat, max(remaining, _MIN_WAIT_BUDGET_SECONDS=0.05))` at :588–594; the new test at test_events_stream.py:733 fails on the old code.
4. **WAIVED** — heavy snapshot reader / discarded-question query: the at-most-once claim holds except the bounded `resume_requested` limbo; the finding itself said "Negligible", and a slim reader is yagni until prompt projection exists.
5. **FIXED** — magic string `resume_requested`: `TaskControlState.RESUME_REQUESTED.value` at :410 and :669; import closure verified acyclic.
6. **FIXED** — cap wording imprecision: wording now separates the reservation counter (never exceeds the cap; checks before increment) from the open-stream count (can exceed); the old phrasing is gone repo-wide.
7. **PARTIAL** — the signature test was rewritten (test_events_stream.py:499–517) to assert no parameter is annotated `Session` — rename-proof and it catches the dominant `get_db` pattern — but it is still signature-only: an endpoint opening its own `SessionLocal()` in the body (tasks.py's own idiom) or using `Annotated[Session, Depends(...)]` would pass undetected. Behavioral assertion not adopted; see inline comment.
8. **FIXED** — `dropped_frame_count` had no production observer: the docstring now states drops are logged via `logger.exception`, with the counter test-only (verified :241–244, :273–277).
9. **FIXED** — no `WAITING_FOR_USER` attach fast path: `_input_required_snapshot_stream` routed at :661–674; a real-HTTP test pins immediate close with a null prompt; the resume carve-out is identical at attach and watchdog.
10. **FIXED** — ping docstring wording: tasks.py:1236–1239 now says ping only after 15 s idle.
11. **FIXED** — docstring omitted payload shapes: all four event field lists in tasks.py:1219–1230, byte-verified against the frame builders.
12. **FIXED** — completed-carries-failed + flat error envelope: both documented; deferring V1ErrorCode vocabulary alignment is a reasonable published-contract call (the two vocabularies are documented as disjoint instead).
13. **FIXED** — test gaps a/b/d: b and d pinned by real tests that fail on broken code; a deferred with a sound rationale (no `client.stream` precedent; TestClient drain collapses to `aclose()`, which the existing teardown tests already simulate).
14. **WAIVED** — duplicated 1 h cap design note: the cross-reference comment is verified present and accurate (a2a.py:105); the version-gated external protocol vs own-SDK surfaces genuinely differ.
15. **SUPERSEDED → re-raised as N1** — anti-buffering headers were waived at prior minor severity, but new deployment evidence (the shipped docker/nginx.conf buffers `/v1/`, the compose topology exposes only nginx, default 4k buffer ⇒ zero bytes mid-stream) escalates this to a major blocker; see N1.
16. **FIXED** — per-task cap docstring absolute wording: both caps now use best-effort wording (tasks.py:1271–1283). Residual: per-process scope still unacknowledged → N8.
17. **WAIVED** — watchdog DB sessions per hour: 2 indexed single-row sessions per 30 s cycle with key-first fail-closed ordering verified; the author's rationale is technically correct.
18. **FIXED** — failure close waited up to 30 s: terminal-status broadcasts now set `completion_hint` (_events_stream.py:267–272); test at :1074–1087 fails on the old code. Residual doc gap on the asymmetric failure-path vocabulary → N10.
- **S1 (delete `_binding_matches`) — NOT FIXED.** The author's decline rests on a premise the code does not support: the claimed "shared manager in exchange for binding check" trade-off is not evidenced, and the check's own docstring admits it is defense-in-depth that "only matters if a connection is ever reassigned via `move_connection`" — which is never called on sinks (its single caller, websocket.py:5750, passes a real WebSocket). The guard is unreachable in every path; 16 lines plus a synthetic test pin unreachable behavior. Carried as S1 in the simplification section.
- **S2 (delete `_create_task` content param) — WAIVED.** Signature parity across three deliberate verbatim copies is real (test_tasks.py uses the param 9×, test_task_reply.py 1×); the conftest hoist is the right separate cleanup.

## New findings

### Major

**N1 — `src/xagent/web/api/v1/_events_stream.py:683` (+ `docker/nginx.conf:69-78`) — SSE responses are buffered by the shipped nginx deployment.** The long-lived `StreamingResponse` at :683–695 carries no `X-Accel-Buffering`/`Cache-Control` headers, and the shipped `docker/nginx.conf` `/v1/` location (lines 69–78) leaves nginx's default `proxy_buffering` on; docker-compose exposes only nginx. Frames are ~40–200 B at 15 s pings, so with the default 4k buffer the client receives zero bytes until the buffer fills (~25 min — usually never in a run) or upstream EOF, at which point everything flushes at once. The feature's core promise — "learn a task is alive the moment it attaches" — fails in the standard shipped deployment; "learn it ended" survives only via the flush on EOF. Fix: `headers={"X-Accel-Buffering": "no", "Cache-Control": "no-cache"}` on the `StreamingResponse`, plus `proxy_buffering off;` on `/v1/` (and `/api/` for a2a parity).

### Minor

**N3 — `src/xagent/web/api/v1/_events_stream.py:661-681`** — the fast paths (`_terminal_snapshot_stream`, `_input_required_snapshot_stream`) return before both cap checks, so terminal/waiting tasks admit uncapped attaches (each costs one bcrypt + one snapshot, yields 2 frames, closes — DoS-parity with polling, not a stream leak). The endpoint docstring's Raises section (tasks.py:1271–1282) has no carve-out. Fix: one-line doc carve-out ("fast-path snapshot attaches are exempt — they hold no stream").

**N4 — `src/xagent/web/api/v1/_events_stream.py:279`** — deletion has no acceleration signal: the chat.py delete path detaches + no-op closes with no deletion broadcast, so the stream emits `: ping` keep-alives up to one full 30 s watchdog cycle before `stream.error {task_deleted}`. Bounded and benign under the watchdog-authority design, but it is the only close reason without an acceleration signal and the latency is undocumented. Fix: emit a deletion broadcast or document the ≤30 s latency.

**N5 — `src/xagent/web/api/v1/_events_stream.py:122`** — the 'unauthorized' wire message ("The API key used to open this stream is no longer valid.") also fires for paused keys (`active_runtime_key_filters` includes `paused_at IS NULL`); unpause restores the key. The endpoint docstring correctly says revoked/paused, but the client-visible message misdescribes a recoverable state. Reword.

**N6 — `src/xagent/web/api/v1/_events_stream.py:400` (and :661)** — the close-decision predicate is duplicated: `watchdog_check_once` (:400–412) vs the attach fast paths (:661–674) are logically identical and textually duplicated, with only the receiver name differing and no cross-reference. Drift risk on any new status or resume carve-out change. Extract one predicate.

**N7 — `src/xagent/web/api/v1/_events_stream.py:84`** — `_TERMINAL_STATUSES = (COMPLETED, FAILED)` is defined by value in three modules (a2a.py:95, tasks.py:669, _events_stream.py:84–85) with no cross-reference, plus `_TERMINAL_CACHE_STATUSES` at chat.py:163 and inline sets at workforces.py:721 and chat.py:6138 — 5+ sites to update for a new terminal status. The import cycle (tasks ↔ _events_stream) is real, but the tuple could live next to `TaskStatus` in models/task.py (acyclic; all three modules already import from there).

**N8 — `src/xagent/web/api/v1/tasks.py:1273`** — residual of prior finding 16: the caps are per-process (`_principal_stream_counts` is a module-global dict; the manager is a process-local singleton), but multi-worker deploys are documented (web/README.md:165, `gunicorn -w 4`) and "32 or more already open for this key across all tasks" reads as deployment-wide. The prior fix addressed only the best-effort wording. Fix: one-line per-process note.

**N9 — `src/xagent/web/api/v1/_events_stream.py:235-237`** — stale line-number refs in the `send_text` docstring: it cites `websocket.py:4504-4532` / `:4525-4532`, but `broadcast_to_task` actually starts at 4510 and the re-raise sits at 4531–4538. Stale from introduction; the cited range covers the wrong code. Reference by name instead.

**N11 — `src/xagent/web/api/v1/_events_stream.py:279`** — no test pins the sink's no-op `close()`: its only consumer is the delete-endpoint cleanup loop (chat.py:4805–4815), so removing it passes CI and regresses only to a log warning. Watchdog `task_deleted` close is tested only via direct row deletion, not the real delete route. Add a test: open a stream, delete via the real route, assert the stream terminates with `stream.error/task_deleted`.

**N12 — `tests/web/api/v1/test_events_stream.py:1031`** — the broadcast → enrichment → sink seam is untested: zero `broadcast_to_task`/`_with_current_task_control_state` usage in test_events_stream.py; all sink tests hand-craft enriched frames via `sink.send_text`, so an enrichment regression would silently emit nothing and pass CI. Straightforward: replace the direct `sink.send_text` with `await es.manager.broadcast_to_task({'type': 'task_paused'}, task_id)` in the existing wiring test.

**N13 — `src/xagent/web/api/v1/_events_stream.py:269`** — `waiting_for_user` broadcasts enqueue `task.status` but do not set `completion_hint`, so the authoritative `task.input_required` close lands up to 30 s after the transition — a mirror of prior finding 18, which fixed the terminal side. The payload carries only `{task_id, prompt: null}`, so impact is low, but a one-line fix (extend `completion_hint.set()` to `WAITING_FOR_USER`; the watchdog re-applies the resume carve-out on the authoritative row read, so an early wake cannot mis-close mid-resume) restores the symmetry the author already accepted for terminal statuses.

### Info

**N10 — `src/xagent/web/api/v1/tasks.py:1221`** — unresolved doc half of prior finding 18: the asymmetric terminal vocabulary is undocumented. A failed task emits `task.status('failed')` THEN `task.completed`; success emits only `task.completed` (short-circuit at _events_stream.py:255–262, pinned by tests :1059 and :1074). tasks.py:1221–1223 documents that a failed task "also ends the stream with this event" but not the preceding status frame; the Behavior bullet "a status update each time it changes" is misleading for the success terminal transition. Document the frame order.

- L279 src/xagent/web/api/v1/_events_stream.py: `delete:` variadic `*args, **kwargs` on the no-op `close()` — every reachable call site passes zero arguments; `manager.disconnect` never calls `close`; all arg-passing `.close(code=...)` calls target raw endpoint websockets that can never be sinks. Replace with `async def close(self) -> None`.
- L7846 src/xagent/web/api/websocket.py: `shrink:` `next((c for c in connections if not getattr(c, "is_broadcast_only", False)), None) or _DiscardingCommandWebSocket()` — registered connections are always truthy (Starlette WebSocket `__len__`, plain-class sink, all test stand-ins). Pass the fallback as the default: `next((c for c in connections if not getattr(c, "is_broadcast_only", False)), _DiscardingCommandWebSocket())`.
- L277 tests/web/api/v1/test_events_stream.py: `shrink:` 7 identical `build_event_stream_response(...)` calls repeating `watchdog_interval_seconds=1000, stream_max_duration_seconds=1000, heartbeat_interval_seconds=1000` (at 272, 528, 567, 624, 898, 1014, 1109). Extract a local helper with 1000 defaults + pass-through kwargs; net ~-41 lines.
- L289 src/xagent/web/api/v1/_events_stream.py (carried, still NOT FIXED): `delete:` `_binding_matches` + its synthetic test (test_events_stream.py:878–885) — unreachable defense: the manager already scopes per-task delivery and `move_connection` is never called on sinks; the author's decline rests on an unsupported premise (see Prior findings S1).

Net: approximately -65 lines.

**Blocking: yes (N1 major).**

Blocking issues:
- `src/xagent/web/api/v1/_events_stream.py:683-695` (+ `docker/nginx.conf:69-78`) — **major** — the SSE stream is buffered by the shipped nginx deployment; the client receives no bytes mid-stream, defeating the endpoint's core "alive at attach" promise. `[new]`

No prior finding remains open at critical/major (prior major finding 1 is FIXED; P7 PARTIAL is minor; S1 NOT FIXED is a deletion-only cleanup). Recommended event: **REQUEST_CHANGES**.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+        )
+
+    if count_task_sinks(task_id) >= PER_TASK_STREAM_CAP:
+        raise V1ApiError(V1ErrorCode.RATE_LIMITED, 429)
+
+    key_prefix = principal.key.key_prefix
+    if not principal_slot_available(key_prefix):
+        raise V1ApiError(V1ErrorCode.RATE_LIMITED, 429)
+
+    return StreamingResponse(
```
rogercloud:
```python
return StreamingResponse(
    _generate(
        ...
    ),
    media_type="text/event-stream",
)
```
This blocks: the shipped `docker/nginx.conf` buffers `/v1/` (no `proxy_buffering off` at nginx.conf:69–78) and the compose topology exposes only nginx, so with ~40–200 B frames at 15 s pings the client receives zero bytes until the 4k buffer fills (~25 min, usually never in a run) or upstream EOF. The endpoint's core promise — learn a task is alive the moment it attaches — fails in the standard deployment. Add `headers={"X-Accel-Buffering": "no", "Cache-Control": "no-cache"}` here and `proxy_buffering off;` on `/v1/` (and `/api/` for a2a parity).

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+    # ``prompt`` is always null: populating it from the agent's question
+    # text is the content-projection layer's job, not this transport.
+    return _sse_frame("task.input_required", {"task_id": task_id, "prompt": None})
+
+
+_ERROR_MESSAGES = {
+    "resync_required": (
+        "The output queue overflowed; call steps() to resync, then re-attach."
+    ),
+    "unauthorized": "The API key used to open this stream is no longer valid.",
```
rogercloud:
```python
"unauthorized": "The API key used to open this stream is no longer valid.",
```
This message also fires for paused keys (`active_runtime_key_filters` includes `paused_at IS NULL`), and unpausing restores the key — the message misdescribes a recoverable state. The endpoint docstring correctly says revoked/paused; reword the client-visible message to match (e.g. mention the key was revoked or paused).

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+                        # otherwise miss the acceleration signal above and
+                        # wait out the watchdog's normal cadence.
+                        self.completion_hint.set()
+        except Exception:
+            self.dropped_frame_count += 1
+            logger.exception(
+                "v1 SSE sink dropped a broadcast frame for task %s", self.task_id
+            )
+
+    async def close(self, *args: Any, **kwargs: Any) -> None:
```
rogercloud:
```python
async def close(self, *args: Any, **kwargs: Any) -> None:
```
Two things. (1) Deletion has no acceleration signal: the chat.py delete path detaches + no-op closes with no deletion broadcast, so a deleted task's stream keeps sending `: ping` keep-alives up to one full 30 s watchdog cycle before `stream.error {task_deleted}` — the only close reason without an acceleration signal. Emit a deletion broadcast or document the ≤30 s latency. (2) No test pins this no-op: its only consumer is the delete-endpoint cleanup loop (chat.py:4805–4815), so removing it passes CI and regresses only to a log warning; `task_deleted` close is tested only via direct row deletion. Add a test that opens a stream, deletes via the real route, and asserts the stream terminates with `stream.error/task_deleted`.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+        snapshot = await run_db_io_cancellation_safe(
+            lambda: read_task_snapshot(task_id, principal)
+        )
+    except V1ApiError as exc:
+        if exc.code is V1ErrorCode.TASK_NOT_FOUND:
+            return sink.enqueue_close(error_frame("task_deleted"))
+        raise
+
+    status = snapshot.status
+    if status in _TERMINAL_STATUSES:
```
rogercloud:
```python
if status in _TERMINAL_STATUSES:
    return sink.enqueue_close(
```
This close-decision predicate in `watchdog_check_once` is textually duplicated in the attach fast paths (see :661–674 — same logic, only the receiver name differs, no cross-reference). Any new terminal status or resume carve-out change must be applied in both places. Extract one shared predicate.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+STREAM_MAX_DURATION_SECONDS = 60.0 * 60.0
+OUTBOUND_QUEUE_MAX_SIZE = 256
+PER_TASK_STREAM_CAP = 2
+PER_PRINCIPAL_STREAM_CAP = 32
+# Floor for the final wait when the 1-hour deadline is close: keeps the
+# queue wait from being called with a near-zero timeout, which would spin
+# the loop instead of actually waiting.
+_MIN_WAIT_BUDGET_SECONDS = 0.05
+
+_TERMINAL_STATUSES = (TaskStatus.COMPLETED, TaskStatus.FAILED)
```
rogercloud, line 92:
```python
_TERMINAL_STATUSES = (TaskStatus.COMPLETED, TaskStatus.FAILED)
```
Third by-value copy of this tuple (also a2a.py:95 and tasks.py:669), plus `_TERMINAL_CACHE_STATUSES` at chat.py:163 and inline sets at workforces.py:721 / chat.py:6138 — 5+ sites to update for a new terminal status. The tasks ↔ _events_stream import cycle is real, but the tuple could live next to `TaskStatus` in models/task.py, which all three modules already import.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+            return
+        try:
+            self._queue.put_nowait((frame_text, False))
+        except asyncio.QueueFull:
+            self.enqueue_close(error_frame("resync_required"))
+
+    async def send_text(self, text: str) -> None:
+        """Receive one broadcast frame. Duck-types ``WebSocket.send_text``.
+
+        Must never raise. ``broadcast_to_task`` (``websocket.py:4504-4532``)
```
rogercloud:
```python
Must never raise. ``broadcast_to_task`` (``websocket.py:4504-4532``)
...
(``:4525-4532``) --
```
These line refs are stale from introduction: `broadcast_to_task` actually starts at websocket.py:4510, and the re-raise is at 4531–4538 — the cited range covers the wrong code. Reference the method by name instead of line numbers.

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,695 @@
...
+                # touching the database -- no query happens here.
+                self.completion_hint.set()
+                return
+            if _is_versioned_task_event(message):
+                status = message.get("status")
+                if isinstance(status, str):
+                    self.enqueue_status(status)
+                    if status in _TERMINAL_STATUS_VALUES:
+                        # A failed task's broadcast (``task_error``) never
+                        # carries ``type == "task_completed"``, so it would
```
rogercloud:
```python
if status in _TERMINAL_STATUS_VALUES:
    # ...
    self.completion_hint.set()
```
`waiting_for_user` broadcasts enqueue `task.status` but never set `completion_hint`, so the authoritative `task.input_required` close lands up to 30 s after the transition — the mirror of the terminal-side fix in prior finding 18 (the block above). Extend `completion_hint.set()` to `WAITING_FOR_USER`: the watchdog re-applies the resume carve-out on the authoritative row read, so an early wake cannot mis-close mid-resume.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,94 @@ async def get_chat_task_steps(
...
+    ``task.input_required``, and ``stream.error``. Step-by-step and
+    token-by-token content
+    (``step.*`` / ``message.*``) is not projected onto this stream yet --
+    poll ``GET /v1/chat/tasks/{task_id}/steps`` for that in the meantime.
+
+    This endpoint declares no ``responses=`` schema, so the field lists
+    below are the only OpenAPI-visible contract for its four SSE event
+    types (each a ``event: <name>`` / ``data: <json>`` frame pair):
+      - ``task.status``: ``{status}``.
+      - ``task.completed``: ``{status, output, error}``. A failed task
```
rogercloud, line 1221:
```python
- ``task.completed``: ``{status, output, error}``. A failed task
  also ends the stream with this event rather than a separate
  one -- ``status`` is ``"failed"`` and ``error`` is set.
```
This documents that a failed task ends via `task.completed` but not the frame that precedes it: a failed task emits `task.status('failed')` THEN `task.completed`, while success emits only `task.completed` (short-circuit at _events_stream.py:255–262, pinned by tests :1059/:1074). The "a status update each time it changes" behavior bullet is misleading for the success terminal transition. Document the frame order.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,94 @@ async def get_chat_task_steps(
...
+        principal: Resolved by the auth dependency; also used (not as
+            an authorization gate) for the per-key concurrency count
+            and the periodic key-validity check.
+
+    Raises:
+        V1ApiError 401: missing / invalid / revoked key (plain JSON;
+            the stream never opens).
+        V1ApiError 404: task missing or not owned by the calling key
+            (plain JSON; the stream never opens).
+        V1ApiError 429 ``rate_limited``: 2 or more concurrent streams
```
rogercloud, line 1288:
```python
V1ApiError 429 ``rate_limited``: 2 or more concurrent streams
    already open on this task, or 32 or more already open for
    this key across all tasks (the per-key cap). ...
```
The fast paths (`_terminal_snapshot_stream`, `_input_required_snapshot_stream`, _events_stream.py:661–681) return before both cap checks, so terminal/waiting tasks admit uncapped attaches (each costs one bcrypt + one snapshot, yields 2 frames, closes — DoS-parity with polling, not a stream leak). The Raises section has no carve-out; add one line: fast-path snapshot attaches are exempt — they hold no stream.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,94 @@ async def get_chat_task_steps(
...
+            already open on this task, or 32 or more already open for
+            this key across all tasks (the per-key cap). Both caps are
```
rogercloud, line 1290:
```python
this key across all tasks (the per-key cap). Both caps are
```
The caps are per-process (`_principal_stream_counts` is a module-global dict; the manager is a process-local singleton), but multi-worker deploys are documented (web/README.md:165, `gunicorn -w 4`) and "across all tasks" reads as deployment-wide. Add a one-line note that the caps are per-process.

tests/web/api/v1/test_events_stream.py
```diff
@@ -0,0 +1,1137 @@
...
+    await asyncio.wait_for(_drain(), timeout=2)
+    assert call_count >= 2  # the first (failing) cycle and the retry
+    assert any("event: task.completed" in f for f in frames)
+    assert es.count_task_sinks(task_id) == 0
+
+
+# ===== endpoint signature carries no request-scoped DB dependency =====
+
+
+def test_events_endpoint_signature_has_no_db_dependency():
```
rogercloud:
```python
sig = inspect.signature(v1_tasks.stream_chat_task_events)
assert not any(param.annotation is Session for param in sig.parameters.values())
```
This rewrite is rename-proof and catches the dominant `get_db` pattern — good. It is still signature-only: an endpoint opening its own `SessionLocal()` in the body (tasks.py's own idiom) or using `Annotated[Session, Depends(...)]` would pass undetected. A behavioral assertion (the stream stays alive for the 1-hour cap without a checked-out session) would close the gap; at minimum, note the limitation in the test.

tests/web/api/v1/test_events_stream.py
```diff
@@ -0,0 +1,1137 @@
...
+    )
+    first = await asyncio.wait_for(resp.body_iterator.__anext__(), timeout=2)
+    assert json.loads(first.split("data: ", 1)[1]) == {"status": "running"}
+
+    sink = next(
+        c
+        for c in es.manager.connections_for_task(task_id)
+        if isinstance(c, es.V1EventStreamSink)
+    )
+    await sink.send_text(
```
rogercloud:
```python
await sink.send_text(
    json.dumps({"type": "task_paused", "task_id": task_id, "status": "paused"})
)
```
All sink tests hand-craft enriched frames like this, so the broadcast → enrichment → sink seam is untested: `broadcast_to_task`/`_with_current_task_control_state` are never used in this file, and an enrichment regression would silently emit nothing and still pass CI. Replace this with `await es.manager.broadcast_to_task({'type': 'task_paused'}, task_id)` so the real path is exercised.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Confirmed the chain end to end: the shipped nginx config leaves default proxy buffering on for `/v1/`, compose exposes only nginx, and there is no gzip middleware, so nginx is the sole buffering layer. One correction to the numbers: nginx's flush threshold is `proxy_busy_buffers_size`, which defaults to twice `proxy_buffer_size` — 8k, not 4k — so the wait is longer than described. Fixed at the application layer: all three `StreamingResponse` constructions now go through a shared `_sse_response` helper that sets `X-Accel-Buffering: no` and `Cache-Control: no-cache`; a parametrized test pins the headers on all three paths. Not adding `proxy_buffering off` to the nginx config: the header is nginx's own per-response override and fully resolves the shipped deployment, while a location-wide switch would also disable buffering for every non-streaming JSON endpoint under `/v1/`, where it is doing useful work. The a2a stream has the identical gap and it is real, but it is a separate transport surface this PR does not otherwise touch — filed as a follow-up issue.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Reworded — the message now says the key was revoked or paused, matching the filter (`active_runtime_key_filters` excludes paused keys) and no longer misdescribing a recoverable state.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Both taken. The ≤30s close latency after deletion is now documented in the endpoint's behavior notes (deletion is the one close reason with no acceleration signal; the watchdog's authoritative read catches it). And there is now a route-level test that opens a stream, deletes the task through the real delete endpoint, and asserts the stream ends with `stream.error {task_deleted}` — it uses a test-local app mounting the chat router, mirroring the existing pattern for that endpoint, since this suite's shared app deliberately does not mount it. Not emitting a deletion broadcast: that would touch the delete path itself, which this PR otherwise leaves alone.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Extracted — `_stream_close_reason` is now the single close-decision predicate; the watchdog and the attach fast paths both call it and keep only their distinct outputs (a close frame vs a snapshot response). The existing terminal, waiting-for-user, and resume-carve-out tests all pin it.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Agreed it is real — five by-value sites (one of the cited ones does not exist; `chat.py` is shorter than that line number, but the other four check out plus the workforces one). Consolidating means touching four modules this PR otherwise does not, so filed as a follow-up issue proposing the tuple live next to `TaskStatus` itself.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Replaced with references by name — the docstring now points at `broadcast_to_task` and its re-raise behavior without line numbers.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Taken. Terminal statuses and `waiting_for_user` now share an early-wake set, so the authoritative `task.input_required` close lands on the next watchdog wake instead of the next full interval. Verified your safety argument in code before adopting it: the watchdog re-reads the authoritative row and re-applies the resume carve-out, and the existing resume-requested test pins that an early wake cannot mis-close mid-resume. The watchdog docstring is updated to match.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Documented — the endpoint docstring now spells out the frame order: a failed task emits `task.status ("failed")` then `task.completed`; success ends with `task.completed` alone (no preceding status frame); the attach fast paths emit both frames.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Added the carve-out line — snapshot fast-path attaches (terminal and waiting states) hold no stream and are exempt from both caps.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Added — both caps are now documented as per-process (each worker counts independently in a multi-worker deployment).

tests/web/api/v1/test_events_stream.py
AlexLiu190625: Noted the limitation in the test docstring — it is a signature-level check and would not catch an endpoint opening its own session in the body; the assertion itself is unchanged. A behavioral hold-the-stream-open assertion needs the test-client streaming infrastructure that the disconnect test also needs, so both stay with that follow-up.

tests/web/api/v1/test_events_stream.py
AlexLiu190625: Fixed, with one correction to the suggested one-liner: calling `broadcast_to_task` verbatim times out, because enrichment finds no state tuple on the message, reads the authoritative row, and overwrites `status` with the row's value — still `running` at that point, so the frame dedupes against the attach status and nothing is emitted. The test now flips the task row to paused first and then broadcasts, which exercises the real seam: broadcast → enrichment (including its DB read) → sink → generator. A mutation check (removing the row update) reproduces the timeout, confirming the hazard is real.

AlexLiu190625:
Addressed in the commits just pushed; per-thread details on each inline thread. Three findings whose fixes belong outside this PR's boundary are now tracked: the a2a stream's identical buffering gap (#1358), consolidating the terminal-status tuple (#1359), and a shared contract for the broadcast wire format (#1360).

On the `_binding_matches` deletion: you were right and my previous decline was wrong on both counts. The trade-off I cited lives in an internal design note, not in the code, so it was not something you could verify — and on re-examination the guard does not even deliver what its docstring claims: re-registration removes a connection from its old task first, so after a hypothetical rebind the sink would receive only the new task's frames, and the guard would drop all of them — a permanently silent stream still holding its registration, its per-key slot, and its watchdog, which is worse than the misdelivery it advertises. Deleted the method, its call site, and its synthetic test.

Also taken: the `close()` signature narrowed to zero-arg, the router fallback moved into `next()`'s default, and the repeated long-interval kwargs in the tests collapsed into a helper (the actual count was 14 call sites).

src/xagent/web/api/v1/_events_stream.py
```diff
@@ -0,0 +1,714 @@
...
+            if message.get("type") == "task_completed":
+                # Acceleration signal only: the authoritative completion
+                # frame still comes from the watchdog reading the task
+                # row, just woken up early instead of waiting out its
+                # normal cadence. This keeps the sink itself from ever
+                # touching the database -- no query happens here.
+                self.completion_hint.set()
+                return
+            if _is_versioned_task_event(message):
+                status = message.get("status")
```
rogercloud, line 301: `send_text` reads only `message.get('status')` and discards the enrichment stamps (`run_id`, `state_version`, `control_state`) that `websocket.py:4392-4440` attaches — a docstring there mandates clients compare `run_id`/`state_version`. Dedup is consecutive-equality only, and the watchdog corrects only PAUSED rows (`_events_stream.py:433-443`), so a paused-finalization broadcast carrying a producer-captured stale tuple (websocket.py:2568-2584) delivered after a resume broadcast (websocket.py:7758-7766, fresh snapshot) can leave the client stuck on `task.status{paused}` for an idle running task until the next transition, the watchdog close, or the 1-hour cap. Lifecycle correctness is unaffected — close frames are watchdog-authoritative. Suggest tracking the max `state_version` per stream and suppressing older frames.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,108 @@ async def get_chat_task_steps(
...
+
+    Behavior:
+      - Normal attach: opens the stream, emits ``task.status`` for the
+        task's current state, then a status update each time it
+        changes (consecutive duplicates are suppressed), and
+        ``task.completed`` when the task finishes. A ``: ping`` comment
+        line is sent whenever 15s pass without any other frame going
+        out, to keep the connection alive -- not on a fixed 15s
+        cadence regardless of activity.
+      - Frame sequence into ``task.completed``: a task that fails emits
```
rogercloud, line 1240: The frame-order bullet is a good addition. One nuance remains: `enqueue_close` drains the still-queued backlog before inserting the close frame, so the intermediate `task.status("failed")` can be dropped when the failure broadcast's frame is still queued — the client then sees only `task.completed` (whose `status` field still carries "failed"). This is documented at the sink-method level (`_events_stream.py:230-244`) but not in this client-facing contract. Suggest appending one best-effort sentence to this bullet.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,108 @@ async def get_chat_task_steps(
...
+            this key across all tasks (the per-key cap). Both caps are
+            checked once, before the attach's own stream registers, so
+            both are best-effort under a concurrent attach burst:
+            several attaches can pass the check at the same instant,
+            so the number of streams actually open can briefly exceed
+            the cap -- the per-key cap's own counter never does, since
+            it checks before it increments; it's specifically the
+            count of open streams that can run ahead of it. No stream
+            is aborted once opened; both counts self-heal as open
+            streams close. Both caps are per-process: a multi-worker
```
rogercloud, line 1299: The per-process note for the caps is a good addition. One half of the original finding remains: broadcast *delivery* is also per-process — a sink only receives broadcasts from the emitting worker's process-local connection manager — and that is still undocumented. Suggest a sentence near the Behavior bullets: broadcasts reach a stream only from the emitting worker's process-local connection manager; cross-worker transitions surface via the watchdog's 30 s DB read.

src/xagent/web/api/v1/tasks.py
```diff
@@ -1197,3 +1199,108 @@ async def get_chat_task_steps(
...
+            deployment counts independently in each worker process, so
+            the effective cluster-wide limit is the per-process cap
+            times the worker count. An attach that takes the terminal
+            or waiting-for-user fast path (see Behavior above) never
```
rogercloud, line 1303: The fast-path exemption is now documented — thanks. The commit message also claims a "fast paths' cap exemption" test, but no test pins the exemption: none fills both caps and then fast-path-attaches expecting 200, so hoisting the cap checks above the fast-path returns would leave the suite green. Suggest a test that exhausts both caps and asserts a terminal/waiting attach still succeeds.

tests/web/api/v1/test_events_stream.py
```diff
@@ -0,0 +1,1250 @@
...
+    own ad hoc DB session internally (rather than through
+    ``run_db_io_cancellation_safe``) -- it only rules out a request-scoped
+    ``Session`` dependency injected by FastAPI.
+    """
+    import inspect
+
+    from sqlalchemy.orm import Session
+
+    sig = inspect.signature(v1_tasks.stream_chat_task_events)
+    assert not any(param.annotation is Session for param in sig.parameters.values())
```
rogercloud: This is still signature-only: it would pass if the endpoint opened its own `SessionLocal()` in the body (tasks.py's own idiom) or used `Annotated[Session, Depends(...)]`. The docstring concession (lines 609–612) is appreciated, but a behavioral assertion — the stream survives the 1-hour cap without a checked-out session — would close the gap.

rogercloud (approved):
## Summary

This PR adds `GET /v1/chat/tasks/{task_id}/events`, an SSE stream of task lifecycle events (`task.status`, `task.completed`, `task.input_required`, `stream.error`) so SDK clients get immediate feedback instead of discovering state through 1 Hz `steps()` polling — nothing is observable between POST and the first step, and terminal/input-required transitions are otherwise learned at poll latency. The stream rides the existing WebSocket broadcast fan-out through a transport-only sink, with a 30-second DB watchdog as the single terminal authority, bounded resources (2 streams/task, 32/key, 256-frame queue, 1-hour cap), and graceful degradation — broadcasts only accelerate, the watchdog guarantees. Blocking: no — recommended event: APPROVE.

## Update since last review

Two commits landed since the previous review (head 8234a890). 4bb765e6 closes the review gaps: a shared `_sse_response` helper applies `X-Accel-Buffering: no` + `Cache-Control: no-cache` on all three `StreamingResponse` sites (fixing the N1 nginx-buffering blocker), `_stream_close_reason` unifies the close-decision predicate, `_EARLY_WAKE_STATUS_VALUES` extends the early-wake set to `waiting_for_user`, the unauthorized message is reworded, the sink `close()` is zero-arg, `_binding_matches` is deleted, and the docstrings gain the frame-sequence, fast-path carve-out, and per-process notes. 19c5ee1a pins these with a parametrized header test across the three paths, a `_long_intervals()` helper, a real DELETE-route test, a real `manager.broadcast_to_task` seam test, and removal of the `_binding_matches` synthetic test.

## Approach verdict

**Sound.** The design model holds: a transport-only `_events_stream.py` riding the existing WebSocket broadcast fan-out, the sink duck-typing the WebSocket contract and registering in the shared ConnectionManager, the watchdog as the single terminal-state authority, bounded resources with graceful degradation, and an explicit close frame that is always terminal. Broadcasts are correctly an acceleration mechanism, never authority. Two design-level concerns are tracked rather than blocking: the terminal-status tuple consolidation (#1359) and the wire-contract hoist between the two transports (#1360).

## Prior findings checklist

**FIXED (25):** R1-1 (connections[0] swallow — durable-command router now skips broadcast-only sinks), R1-2, R1-3, R1-5, R1-6, R1-8, R1-9, R1-10, R1-11, R1-12, R1-13 (b/d pinned; a deferred with accepted rationale), R1-16, R1-18, R1-S1 (`_binding_matches` REMOVED), R2-N1 (major — anti-buffering headers + parametrized test; a2a parity tracked in #1358), R2-N4 (doc + real DELETE-route test), R2-N5, R2-N6, R2-N9, R2-N11, R2-N12, R2-N13, R2-S-close, R2-S-next, R2-S-helper.

**WAIVED (unchanged, 4):** R1-4, R1-14, R1-17, R1-S2.

**Tracked in issues:** R2-N1's a2a buffering parity → #1358; R2-N7 (terminal-status tuple consolidation) → #1359; R2-N2 (wire-contract hoist) → #1360.

**PARTIAL (4, all minor):**

1. **R1-7** (`tests/web/api/v1/test_events_stream.py:623`) — signature test still signature-only: it would pass if the endpoint opened its own `SessionLocal()` in the body or used `Annotated[Session, Depends(...)]`. The author added a docstring conceding the gap (lines 609–612); no behavioral assertion. See inline comment.
2. **R2-N3** (`tasks.py:1293-1296`) — the doc carve-out for the fast-path cap exemption is added, but the commit message's claimed "fast paths' cap exemption" test does not pin the exemption: no test fills both caps then fast-path-attaches expecting 200; hoisting the cap checks above the fast-path returns would leave the suite green. See inline comment.
3. **R2-N8/M9** (`tasks.py:1290-1292`) — caps are now documented per-process, but broadcast *delivery* being per-process is still undocumented: broadcasts reach a stream only from the emitting worker's process-local connection manager; cross-worker transitions surface via the watchdog's 30 s DB read. See inline comment.
4. **R2-N10/M5** (`tasks.py:1240-1246`) — frame-order doc added, but `enqueue_close`'s backlog drain can still drop the intermediate `task.status("failed")` frame; documented only at the sink-method level (`_events_stream.py:230-244`), not in the client-facing endpoint docstring. See inline comment.

## New findings

### Minor

**M1 — `src/xagent/web/api/v1/_events_stream.py:292-308` (severity-adjusted major→minor) — enrichment stamps discarded in `send_text`.** The sink reads only `message.get('status')` and drops `run_id`, `state_version`, `control_state` — the stamps `websocket.py:4392-4440` attaches, whose docstring mandates clients compare `run_id`/`state_version`. Dedup is consecutive-equality only, and the watchdog corrects only PAUSED rows (`:433-443`), so a paused-finalization broadcast (websocket.py:2568-2584, producer-captured stale tuple) delivered after a resume broadcast (websocket.py:7758-7766, fresh snapshot) can leave the client stuck on `task.status{paused}` for an idle running task until the next transition, the watchdog close, or the 1-hour cap. Lifecycle correctness is unaffected — close frames are watchdog-authoritative. Fix: track the max `state_version` seen per stream and suppress older frames. See inline comment.

**M3 — (minor, doc) — PR-body lifecycle claim overstated.** The PR body's bullet "a client can always tell task ended from stream ended" is too strong: the 1-hour deadline's `enqueue_close` is first-writer-wins, so if the deadline wins the race against a terminal close the client sees only `stream.error{stream_expired}`. The endpoint docstring already reconciles this (`tasks.py:1259`, resync-via-steps); soften the PR-body claim to match.

**M4 — `src/xagent/web/api/v1/_events_stream.py:465-469,588-601` (minor, doc) — missed-broadcast attach window undocumented.** A terminal transition between the attach-time snapshot (`tasks.py:1298-1300`) and sink registration (first generator iteration) yields a stale `task.status{running}` followed by up to ~30 s of silence — the watchdog's first check waits a full interval. The module docstring covers post-registration stale frames but not this full-interval latency. Suggest one sentence, or an immediate first watchdog check.

**M11 — (minor, tests) — residual coverage gaps.** (a) The per-task 429 is asserted only as a raised `V1ApiError`, with no HTTP JSON envelope check (the principal-cap has a wire test); (b) no test pins that no ping is sent while frames flow (the docstring claims ping only when idle); (c) real-HTTP broadcast streaming remains uncovered at the transport level — same family as the R1-13a deferral; the seam is covered by R2-N12's `manager.broadcast_to_task` test.

**M12 — `tests/web/api/v1/test_events_stream.py:60-63` (minor, tests) — fixture leak across tests.** `reset_principal_stream_counts` clears only the counter; the process-global manager is never reset. A test failing between `__anext__` and `aclose()` leaks a sink plus a watchdog task (1000 s interval) into later tests. Suggest an autouse drain fixture that closes any still-registered sinks after each test.

### Trivial

**M13 — (trivial) — PR-body test count stale.** The PR body says 38 transport tests; head has 39 test functions. Update the count.

- L85 `src/xagent/web/api/v1/_events_stream.py`: `delete:` `_TERMINAL_STATUS_VALUES` single-use intermediate — sole consumer is `_EARLY_WAKE_STATUS_VALUES`; fold the comprehension into line 90.

Net: -1 line possible.

Blocking: **no**. No confirmed critical or major findings: the prior major blocker N1 (nginx buffering) is FIXED — anti-buffering headers on all three response sites plus a parametrized test — and the one major-severity candidate this pass (M1) was verified and adjusted to minor (lifecycle correctness is watchdog-authoritative regardless). The remaining findings are minor doc/test-hardening items and one trivial count. Recommended event: **APPROVE**.

src/xagent/web/api/v1/_events_stream.py
AlexLiu190625: Confirmed, and the divergence is already pinned on the other transport. Both frames in your scenario do carry `state_version`: the paused-finalization frame keeps its producer-captured tuple, while `task_resumed` carries none and gets a fresh row snapshot — and the version is monotonic per task, never reset. The reordering is reachable in a single process, since the fresh-snapshot path suspends on a thread-pool row read while the producer-tuple path does not. The web client already implements exactly the guard you describe (`canAcceptTaskControlVersion`), with a test for the paused/resumed/late-paused sequence; this sink is the only consumer of the contract that drops the field. Also confirmed the watchdog does not self-heal a wrongly latched paused state — it only re-emits a status when the row itself is paused.

Rather than adding a second, subtly different copy of the version rule here, the fix belongs with the wire-contract hoist in #1360, where both transports can share one definition — the sink must treat a missing `state_version` as "apply" (most test and legacy frames carry none), while the web client treats it as "cannot roll back established state"; those rules should be reconciled deliberately. In this PR the module and endpoint docstrings now state the real invariant: no `task.status` frame is guaranteed fresh or ordered at any point, and only the three close frames are authoritative, each built from a direct row read.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Agreed — the bullet promised an ordering the implementation only makes best-effort. It now notes the close frame's enqueue drains the backlog, so a queued `task.status ("failed")` can be dropped, and that no information is lost: `task.completed` still carries `status: "failed"` and the error.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Confirmed — the connection manager is process-local with no cross-process fan-out, and execution runs in-process on whichever worker served the request, so the two can differ under multiple workers. Widened the note slightly beyond the finding: lease-expiry recovery transitions a task to paused or failed as a pure database write with no broadcast at all, so for that class of transition the watchdog's poll is the only delivery path in any deployment shape. The docstring now covers both.

src/xagent/web/api/v1/tasks.py
AlexLiu190625: Confirmed by mutation before fixing: hoisting both cap checks above the fast-path returns left the whole suite green — nothing pinned the exemption. Added a parametrized test that fills the per-task cap with real running streams (a fast-path attach registers no sink, so the cap must be filled while the task is still running), tops up the per-key counter, flips the task to a terminal or waiting state, and asserts the attach still returns 200 with the snapshot-closing stream. Under the same mutation it now fails with 429. On the earlier commit's subject line: its body scoped this item to documentation, but the subject said "cover" under a test prefix, which over-claimed — the test now exists, and the squash description should carry the accurate wording.

tests/web/api/v1/test_events_stream.py
AlexLiu190625: You're right, and my earlier reason for deferring this was wrong. Both bypasses you named pass the old assertion: a body-opened session (the test never looked at the body) and an `Annotated` dependency (the annotation object is not `Session` itself). And the claimed infrastructure gap does not exist — the shared conftest already ships a one-slot pool helper used by five other test files, three of them in this same directory, asserting exactly this property.

Replaced with a behavioral test that invokes the route handler itself (bypassing the HTTP layer but executing the full function body), installs the one-slot pool, asserts no connection is checked out while the stream is alive, and then actually acquires the slot to prove it is free. Verified by mutation: a session opened and held inside the endpoint body now fails the assertion; the previous version of this test stayed green under that same mutation.

PR #1342, opened 2026-08-13
fix(mcp): strip all non-tool-name characters from MCP tool names

### LINKED ISSUE #1340: fix(mcp): tool names containing dots break OpenAI-compatible tool-calling (400 invalid function name)
## Summary

MCP tools whose names contain a `.` (a common namespacing convention,
e.g. `coding.start`, `coding.poll`) crash any LLM call routed through
an OpenAI-compatible chat-completions endpoint with a 400 error,
because the sanitization in `CustomMCPTool.name` never strips dots.

## Root cause

`src/xagent/core/tools/adapters/vibe/mcp_adapter.py:409-414`:

```python
@property
def name(self) -> str:
    """Get tool name with optional prefix, formatted for LLM requirements."""
    raw_name = f"{self._name_prefix}{self.mcp_tool.name}"
    # Replace spaces and dashes with underscores to match LLM tool naming constraints
    # This matches the frontend/chat.py filtering logic
    return raw_name.replace(" ", "_").replace("-", "_")
```

This only replaces spaces and dashes. OpenAI-compatible APIs (and at
least DeepSeek's OpenAI-compatible endpoint) validate function names
against `^[a-zA-Z0-9_-]+$`. A raw MCP tool name containing a dot (e.g.
`coding.start`) survives this sanitization untouched, producing a
final tool name like `mcp_Coding_MCP_coding.start`, which still
contains `.` and fails validation.

Note `_format_unavailable_mcp_tool_name` (mcp_adapter.py:327-331), used
on a different code path (the "server unavailable" placeholder tool),
already strips with a more complete `[^A-Za-z0-9_]+` pattern -- the two
sanitization paths are inconsistent, and the main path is the one
that's incomplete.

## Reproduction

1. Register a custom streamable_http MCP server whose tools are named
   with dots, e.g. `coding.start`.
2. Attach it to an agent and send a message that triggers tool
   selection, routed through an OpenAI-compatible model (reproduced
   with DeepSeek here).
3. The LLM call fails with:

```
openai.BadRequestError: Error code: 400 - {'error': {'message':
"Invalid 'tools[22].function.name': string does not match
pattern. Expected a string that matches the pattern
'^[a-zA-Z0-9_-]+$'.", 'type': 'invalid_request_error', 'param':
None, 'code': 'invalid_request_error'}}
```

## Suggested fix

Replace the ad-hoc `.replace(" ", "_").replace("-", "_")` chain in
`CustomMCPTool.name` with the same `re.sub(r"[^A-Za-z0-9_-]", "_", ...)`
style sanitization already used in `_format_unavailable_mcp_tool_name`,
so both code paths produce consistent, API-safe tool names regardless
of what characters the upstream MCP server puts in a tool name.

## Environment

- xagent commit: 747c77b4 (origin/main)
- Model provider: DeepSeek (OpenAI-compatible endpoint)

rogercloud:
## Summary

- `MCPToolAdapter.name` only replaced spaces and dashes when building the LLM-visible tool name, so any other disallowed character (most notably a `.` used for namespacing, e.g. `coding.start`) survived into the final name.
- OpenAI-compatible chat-completions APIs (reproduced with DeepSeek) validate `tools[].function.name` against `^[a-zA-Z0-9_-]+$` and reject the whole call with a 400 if any single tool name fails it — one bad tool name breaks every tool call in that request, not just the one tool.
- Fix: after the existing space/dash replacement, run the remaining name through `re.sub(r"[^A-Za-z0-9_-]", "_", ...)` so any other character an MCP server puts in a tool name gets sanitized the same way, without changing existing space/dash behavior.

Fixes #1340.

## Test plan

- [x] Added `test_mcp_tool_adapter_name_strips_dots_for_openai_compatible_apis` covering a dot-namespaced tool name (`coding.start`), asserting the final name matches `^[A-Za-z0-9_-]+$`.
- [x] `pytest tests/core/tools/adapters/vibe/test_mcp_adapter.py` — 69 passed.
- [x] `ruff check` / `ruff format --check` / `mypy` on the touched files — clean.
- [x] `pre-commit run --files src/xagent/core/tools/adapters/vibe/mcp_adapter.py tests/core/tools/adapters/vibe/test_mcp_adapter.py` — all hooks passed.

### DIFF
--- a/src/xagent/core/tools/adapters/vibe/api_tool_adapter.py
+++ b/src/xagent/core/tools/adapters/vibe/api_tool_adapter.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel, Field, model_validator
 
 from ....utils.encryption import decrypt_value
 from ...core.api_tool import call_api
+from .agent_tool_names import MAX_AGENT_TOOL_NAME_LENGTH
 from .base import AbstractBaseTool, ToolCategory, ToolVisibility
 from .connector_runtime import (
     MISSING_RUNTIME_VALUE,
@@ -134,11 +135,25 @@ class CustomApiTool(AbstractBaseTool):
     ):
         # Format name for LLM (replace spaces/dashes with underscores)
         sanitized_name = name.replace(" ", "_").replace("-", "_")
+        # `name` is free-form user input (CustomApiCreate.name only bounds
+        # length, not charset), so anything the two replacements above
+        # didn't catch -- a `.` in a domain-like name, punctuation, etc. --
+        # would otherwise survive into `function.name` and 400 the whole
+        # LLM call the same way an unsanitized MCP tool name does.
+        sanitized_name = re.sub(r"[^A-Za-z0-9_-]", "_", sanitized_name)
         # Ensure name doesn't start with api_ twice if already prefixed
-        if sanitized_name.startswith("api_"):
-            self._name = f"{sanitized_name}_call"
-        else:
-            self._name = f"api_{sanitized_name}_call"
+        prefix = "" if sanitized_name.startswith("api_") else "api_"
+        suffix = "_call"
+        # Same failure mode as the character check above -- an over-long
+        # name is rejected by the same providers, just on length instead of
+        # charset. Budget the truncation around the fixed prefix/suffix so
+        # `_call` (the part that signals "this is an API call") survives
+        # instead of being cut off by a blind truncation of the whole
+        # string. `MAX_AGENT_TOOL_NAME_LENGTH` is this repo's own record of
+        # that provider limit (agent_tool_names.py), shared here rather than
+        # redeclared so the two adapters can't drift apart on the number.
+        budget = MAX_AGENT_TOOL_NAME_LENGTH - len(prefix) - len(suffix)
+        self._name = f"{prefix}{sanitized_name[:budget]}{suffix}"
 
         # Structured originating-server identity, normalized once through the
         # shared SSOT. A scoped mcp:<server> selector fronts this Custom-API
--- a/src/xagent/core/tools/adapters/vibe/api_tool_adapter.py
+++ b/src/xagent/core/tools/adapters/vibe/api_tool_adapter.py
@@ -1,6 +1,7 @@
 """Custom API Tool Adapter for Agent System."""
 
 import asyncio
+import hashlib
 import json
 import logging
 import re
@@ -153,6 +154,20 @@ class CustomApiTool(AbstractBaseTool):
         # that provider limit (agent_tool_names.py), shared here rather than
         # redeclared so the two adapters can't drift apart on the number.
         budget = MAX_AGENT_TOOL_NAME_LENGTH - len(prefix) - len(suffix)
+        if len(sanitized_name) > budget:
+            # `name` is a free-form, DB-unique string with no server
+            # prefix to squeeze the way the MCP adapter does -- the whole
+            # sanitized name is what a truncation would cut. Two long,
+            # descriptive Custom API names that share a common prefix and
+            # differ only in a trailing qualifier (a normal naming habit,
+            # e.g. two services named for the same team) truncate to the
+            # same string otherwise, and nothing downstream detects that
+            # collision -- mix in a short hash of the *original* name so a
+            # truncated name stays distinct instead of silently dispatching
+            # to whichever of the two tools registers first.
+            digest = hashlib.sha256(name.encode("utf-8")).hexdigest()[:8]
+            hash_suffix = f"_{digest}"
+            sanitized_name = sanitized_name[: budget - len(hash_suffix)] + hash_suffix
         self._name = f"{prefix}{sanitized_name[:budget]}{suffix}"
 
         # Structured originating-server identity, normalized once through the
--- a/src/xagent/core/tools/adapters/vibe/mcp_adapter.py
+++ b/src/xagent/core/tools/adapters/vibe/mcp_adapter.py
@@ -23,6 +23,7 @@ from ..... import config as _root_config
 from .....sandbox.base import Sandbox
 from ...core.mcp.sessions import Connection, create_session
 from ...core.mcp.tools import load_mcp_tools
+from .agent_tool_names import MAX_AGENT_TOOL_NAME_LENGTH
 from .base import AbstractBaseTool, ToolVisibility
 from .connector_runtime import (
     ERROR_DELEGATED_AUTHORIZATION_FAILED,
@@ -410,8 +411,7 @@ class MCPToolAdapter(AbstractBaseTool):
     def name(self) -> str:
         """Get tool name with optional prefix, formatted for LLM requirements."""
         raw_name = f"{self._name_prefix}{self.mcp_tool.name}"
-        # Replace spaces and dashes with underscores to match LLM tool naming constraints
-        # This matches the frontend/chat.py filtering logic
+        # Replace spaces and dashes with underscores to match LLM tool naming constraints.
         sanitized = raw_name.replace(" ", "_").replace("-", "_")
         # Some MCP servers namespace tool names with characters (e.g. the
         # `.` in `coding.start`) that OpenAI-compatible APIs reject outright
@@ -420,7 +420,13 @@ class MCPToolAdapter(AbstractBaseTool):
         # LLM call, not just this one tool. Catch anything left over from
         # the two replacements above instead of only handling the specific
         # characters this method happened to have seen so far.
-        return re.sub(r"[^A-Za-z0-9_-]", "_", sanitized)
+        sanitized = re.sub(r"[^A-Za-z0-9_-]", "_", sanitized)
+        # Same failure mode as the character check above -- an over-long
+        # name is rejected by the same providers, just on length instead of
+        # charset. `MAX_AGENT_TOOL_NAME_LENGTH` is this repo's own record of
+        # that provider limit (agent_tool_names.py), shared here rather than
+        # redeclared so the two adapters can't drift apart on the number.
+        return sanitized[:MAX_AGENT_TOOL_NAME_LENGTH]
 
     @property
     def description(self) -> str:
--- a/src/xagent/core/tools/adapters/vibe/mcp_adapter.py
+++ b/src/xagent/core/tools/adapters/vibe/mcp_adapter.py
@@ -410,23 +410,40 @@ class MCPToolAdapter(AbstractBaseTool):
     @property
     def name(self) -> str:
         """Get tool name with optional prefix, formatted for LLM requirements."""
-        raw_name = f"{self._name_prefix}{self.mcp_tool.name}"
-        # Replace spaces and dashes with underscores to match LLM tool naming constraints.
-        sanitized = raw_name.replace(" ", "_").replace("-", "_")
-        # Some MCP servers namespace tool names with characters (e.g. the
-        # `.` in `coding.start`) that OpenAI-compatible APIs reject outright
-        # -- `^[a-zA-Z0-9_-]+$` is the pattern OpenAI/DeepSeek enforce on
-        # `tools[].function.name`, and a name that fails it 400s the whole
-        # LLM call, not just this one tool. Catch anything left over from
-        # the two replacements above instead of only handling the specific
-        # characters this method happened to have seen so far.
-        sanitized = re.sub(r"[^A-Za-z0-9_-]", "_", sanitized)
+
+        def _sanitize(value: str) -> str:
+            # Replace spaces and dashes with underscores to match LLM tool
+            # naming constraints, then catch anything else disallowed --
+            # some MCP servers namespace tool names with characters (e.g.
+            # the `.` in `coding.start`) that OpenAI-compatible APIs reject
+            # outright (`^[a-zA-Z0-9_-]+$` is the pattern OpenAI/DeepSeek
+            # enforce on `tools[].function.name`), and a name that fails it
+            # 400s the whole LLM call, not just this one tool.
+            return re.sub(
+                r"[^A-Za-z0-9_-]", "_", value.replace(" ", "_").replace("-", "_")
+            )
+
+        sanitized_prefix = _sanitize(self._name_prefix)
+        sanitized_tool = _sanitize(self.mcp_tool.name)
         # Same failure mode as the character check above -- an over-long
         # name is rejected by the same providers, just on length instead of
         # charset. `MAX_AGENT_TOOL_NAME_LENGTH` is this repo's own record of
         # that provider limit (agent_tool_names.py), shared here rather than
         # redeclared so the two adapters can't drift apart on the number.
-        return sanitized[:MAX_AGENT_TOOL_NAME_LENGTH]
+        #
+        # The tool name -- not the prefix -- is the only part that tells
+        # two tools on the *same* server apart, so truncating from the end
+        # (i.e. cutting the tool name) can make two distinct tools collide
+        # into one identical name once `prefix + tool_name` exceeds the
+        # limit. `_find_tool` has no duplicate-name detection, so a
+        # collision isn't a loud error like an illegal character is -- it's
+        # a silent wrong-tool dispatch. Squeeze the prefix instead and keep
+        # the tool name whole; `max(0, ...)` covers a tool name alone at or
+        # past the limit, where a negative slice would otherwise wrap
+        # around from the end instead of emptying.
+        budget_for_prefix = max(0, MAX_AGENT_TOOL_NAME_LENGTH - len(sanitized_tool))
+        combined = f"{sanitized_prefix[:budget_for_prefix]}{sanitized_tool}"
+        return combined[:MAX_AGENT_TOOL_NAME_LENGTH]
 
     @property
     def description(self) -> str:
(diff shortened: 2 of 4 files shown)


src/xagent/core/tools/adapters/vibe/mcp_adapter.py
```diff
@@ -412,7 +412,15 @@ def name(self) -> str:
         raw_name = f"{self._name_prefix}{self.mcp_tool.name}"
         # Replace spaces and dashes with underscores to match LLM tool naming constraints
         # This matches the frontend/chat.py filtering logic
```
AlexLiu190625: Stale reference. It predates this PR, but the PR edits this block, so it is cheap to fix here.

`frontend/chat.py` has never existed in this repository:

```
$ git log --all --oneline --diff-filter=A -- frontend/chat.py
(no output)
$ git log --all --oneline --diff-filter=A -- src/xagent/core/tools/adapters/vibe/mcp_adapter.py
dc32c001 initial commit          # control probe: the same search does find a known-existing path
```

I also could not find any tool-name filtering in `src/xagent/web/api/chat.py` or in the frontend. Suggest deleting the line or repointing it at whatever the real counterpart is.

src/xagent/core/tools/adapters/vibe/mcp_adapter.py
```diff
@@ -412,7 +412,15 @@ def name(self) -> str:
...
-        return raw_name.replace(" ", "_").replace("-", "_")
+        sanitized = raw_name.replace(" ", "_").replace("-", "_")
```
AlexLiu190625: The same pair of replacements exists at `src/xagent/core/tools/adapters/vibe/api_tool_adapter.py:136`, and it still has the exact defect this PR fixes:

```python
sanitized_name = name.replace(" ", "_").replace("-", "_")
...
self._name = f"api_{sanitized_name}_call"
```

Two facts make this the same bug rather than a related one:

1. Both names reach the wire through the same place. `_build_tool_schema` at `src/xagent/core/agent/pattern/react/react.py:1617` copies `tool.name` straight into `function.name` for every tool, MCP adapters and Custom API adapters alike.
2. The Custom API name is free-form user input. `CustomApiCreate.name` at `src/xagent/web/api/custom_api.py:32` bounds only the length (`min_length=1, max_length=100`), not the character set.

Running that code as it stands today:

| user-supplied API name | resulting `function.name` | matches `^[a-zA-Z0-9_-]+$` |
| --- | --- | --- |
| `weather.com` | `api_weather.com_call` | no |
| `Foo (v2)` | `api_Foo_(v2)_call` | no |
| `Slack #alerts` | `api_Slack_#alerts_call` | no |
| `plain_name` | `api_plain_name_call` | yes |

Naming a custom API after a domain is the natural thing for a user to do, so this is not a theoretical corner. The consequence is the one described in the PR body: a single bad name 400s the whole request and takes down every tool call in that chat, not just the offending tool.

Since this PR carries one defect — LLM-visible tool names are not fully sanitized — covering both implementation sites reads as completing that single point rather than widening scope.

src/xagent/core/tools/adapters/vibe/mcp_adapter.py
```diff
@@ -412,7 +412,15 @@ def name(self) -> str:
...
+        # Some MCP servers namespace tool names with characters (e.g. the
+        # `.` in `coding.start`) that OpenAI-compatible APIs reject outright
+        # -- `^[a-zA-Z0-9_-]+$` is the pattern OpenAI/DeepSeek enforce on
+        # `tools[].function.name`, and a name that fails it 400s the whole
+        # LLM call, not just this one tool. Catch anything left over from
+        # the two replacements above instead of only handling the specific
+        # characters this method happened to have seen so far.
+        return re.sub(r"[^A-Za-z0-9_-]", "_", sanitized)
```
AlexLiu190625: Three follow-on notes on the new sanitizer. None of them is a reason to hold the fix; the first one interacts with the Custom API comment above.

**1. There is already an owner for this rule in the same package.**

`src/xagent/core/tools/adapters/vibe/agent_tool_names.py` does the same job for agent delegation tools, and does three things this line does not:

```python
MAX_AGENT_TOOL_NAME_LENGTH = 64

def _semantic_slug(value: Any) -> str:
    normalized_name = unicodedata.normalize("NFKD", raw_name.strip())
    romanized_name = (normalized_name if normalized_name.isascii()
                      else "_".join(lazy_pinyin(normalized_name)))
    ascii_name = romanized_name.encode("ascii", "ignore").decode("ascii")
    slug = re.sub(r"[^a-zA-Z0-9]+", "_", ascii_name.lower())
    return re.sub(r"_+", "_", slug).strip("_-") or "agent"
```

It romanises CJK instead of erasing it, bounds the length to 64, and appends a unique agent id so two tools cannot collide. After this PR the repository has three implementations of one rule: agent tools, MCP, and the still-broken Custom API path. One shared helper consumed by the two adapters would close the sibling gap and remove the duplication in a single change.

The difference shows up immediately on non-ASCII names, because `re.sub` drops those characters rather than transliterating them:

```
server "天气" + tool "查询"  ->  mcp______
```

Any two CJK names of the same length collapse to the same string.

One caveat if you do extract a helper: `_semantic_slug` cannot be reused verbatim here. It lowercases and folds `-`, so applying it to MCP would rename every tool that works today (`mcp_Coding_MCP_x` -> `mcp_coding_mcp_x`). The shared helper needs to preserve case and keep `-`.

**2. Length is the same failure mode and is still unbounded.**

`MAX_AGENT_TOOL_NAME_LENGTH = 64` is this repository's own statement of the provider limit, and it is currently applied only to agent tools. Neither name built here has a cap:

- Custom API: `"api_" + name (<=100) + "_call"` — up to 109 characters.
- MCP: `"mcp_" + server name (<=100) + "_" + tool name (unbounded)`.

An over-long name is rejected the same way a bad character is, so if a shared helper is extracted anyway, adding truncation there costs almost nothing.

**3. Collisions become reachable where they previously failed loudly.**

Folding more characters to `_` means two distinct tools can now produce one name:

```
"coding.start" -> mcp_s_coding_start
"coding-start" -> mcp_s_coding_start
```

`_find_tool` at `src/xagent/core/agent/pattern/react/react.py:3322` returns the first tool whose name matches, and the duplicate-name guard at `src/xagent/core/tools/adapters/vibe/factory.py:717` only covers task-runtime extension tools against core tools, not two MCP tools against each other.

Collision is pre-existing (`my tool` and `my-tool` already collided), and a silent first-match dispatch is still better than a hard 400 on the whole request, so I do not think this blocks. It is worth a line in the PR description, or an id suffix in the shared helper the way agent tool names already do it.

tests/core/tools/adapters/vibe/test_mcp_adapter.py
```diff
@@ -269,6 +270,29 @@ def test_mcp_tool_adapter_source_server_defaults_none():
...
+        inputSchema={"type": "object", "properties": {}},
+    )
+    adapter = _build_mcp_tool_adapter(
+        "Coding MCP",
+        {"transport": "streamable_http", "url": "http://127.0.0.1:8642/mcp"},
+        mcp_tool,
+    )
+
+    assert "." not in adapter.name
+    assert re.fullmatch(r"[A-Za-z0-9_-]+", adapter.name)
```
AlexLiu190625, line 314: The regex is a catch-all, but the test covers a single character class. A small parametrisation would state that intent directly and costs about three lines — for example `coding.start`, `list/items`, `run (v2)`, and one CJK name.

The CJK case is the interesting one to write down: it currently yields an all-underscore name (`mcp_天气_查询` -> `mcp______`), so pinning it makes the current trade-off explicit rather than leaving it to be discovered later.

AlexLiu190625 (commented):
Reviewed at `f34bd50` against base `59127de`.

Verified locally before commenting:

- `pytest tests/core/tools/adapters/vibe/test_mcp_adapter.py` — 69 passed, matching the PR description.
- Mutation check: reverting the `re.sub` line turns the new test red on `assert "." not in adapter.name`, so the test pins the intended behaviour rather than passing incidentally.
- No compatibility break for existing names: every character the new `re.sub` rewrites is, by construction, a character that already failed `^[a-zA-Z0-9_-]+$`. A name it changes is a name that was already being rejected by the provider, so nothing that worked before changes.
- Execution is unaffected: the MCP call still uses the raw `self.mcp_tool.name`; only the LLM-visible name changes.

The fix itself is correct. The one thing I would like addressed before merge is the sibling site: `CustomApiTool` builds `function.name` the same way, from free-form user input, and still has the exact defect this PR fixes. Details inline, along with three smaller follow-on notes.

qinxuye (approved):
LGTM

rogercloud:
Thanks for the thorough review — verified all four points and pushed a follow-up commit (1b6405a9) addressing them:

1. **Stale comment**: confirmed `frontend/chat.py` has never existed in this repo (nor has any equivalent filtering logic). Deleted the comment.

2. **Sibling defect in `CustomApiTool._name`**: confirmed — `api_tool_adapter.py:136` had the exact same unsanitized-character bug, reached through even less constrained input (`CustomApiCreate.name` only bounds length, not charset). Applied the same `re.sub(r"[^A-Za-z0-9_-]", "_", ...)` there, keeping the truncation budget-aware so the `_call` suffix survives a long name instead of being chopped off.

3. **Follow-on notes**: confirmed `_semantic_slug` exists and does romanize CJK via `lazy_pinyin` rather than dropping it, and confirmed neither `MCPToolAdapter.name` nor `CustomApiTool._name` had a length bound. Added a length cap to both, reusing `MAX_AGENT_TOOL_NAME_LENGTH` from `agent_tool_names.py` (imported, not redeclared) so the limit can't drift between the two adapters. One correction to the report: length truncation actually lives in `_semantic_agent_tool_name` (the caller), not in `_semantic_slug` itself — `_semantic_slug` alone wouldn't have brought truncation "for free."

   Deliberately did **not** extract a fully shared helper or adopt the pinyin-romanization/collision-id-suffix behavior from `agent_tool_names.py` into these two adapters — that would change existing tool names' casing/dash handling and is a larger design call than this bug fix calls for. Left the CJK collapse-to-underscores behavior as-is and pinned it in a test instead, per your suggestion.

4. **Test coverage**: parametrized the new test across dot / slash / parens+space / CJK, and added the CJK case exactly as suggested (`天气`/`查询` → `mcp______`, documented rather than "fixed"). Added matching coverage for `CustomApiTool` and a length-cap case for both adapters.

93 tests passing, `ruff`/`mypy`/`pre-commit` clean.

src/xagent/core/tools/adapters/vibe/mcp_adapter.py
```diff
@@ -410,9 +411,22 @@ def __init__(
...
+        # LLM call, not just this one tool. Catch anything left over from
+        # the two replacements above instead of only handling the specific
+        # characters this method happened to have seen so far.
+        sanitized = re.sub(r"[^A-Za-z0-9_-]", "_", sanitized)
+        # Same failure mode as the character check above -- an over-long
+        # name is rejected by the same providers, just on length instead of
+        # charset. `MAX_AGENT_TOOL_NAME_LENGTH` is this repo's own record of
+        # that provider limit (agent_tool_names.py), shared here rather than
+        # redeclared so the two adapters can't drift apart on the number.
+        return sanitized[:MAX_AGENT_TOOL_NAME_LENGTH]
```
AlexLiu190625: This truncation cuts from the end, and the end is the part that distinguishes one tool from another on the same server:

```
mcp_ + server name + _ + tool name
|<-------- fixed prefix ------->|   |<- the only part that varies per tool
                                        this is what gets cut
```

Measured on this commit's code, building real `MCPToolAdapter` instances. Two tools on one server, `read_file` and `delete_all_files`, varying only the server name length. The budget left for the tool name is `64 - 4 - len(server) - 1`:

| len(server name) | budget for tool name | same LLM name? |
| ---: | ---: | :--- |
| 40 | 19 | no |
| 50 | 9 | no |
| 55 | 4 | no |
| 58 | 1 | no |
| **59** | **0** | **yes** |
| 70 | -11 | yes |
| 100 | -41 | yes |

At 59 characters the tool name contributes nothing at all, so *every* tool on that server gets one identical name:

```
read_file        -> mcp_sssssssssssssssssssssssssssssssssssssssssssssssssssssssss_
delete_all_files -> mcp_sssssssssssssssssssssssssssssssssssssssssssssssssssssssss_
```

Server names are bounded at 100 in `web/api/mcp.py:89`, so 59-100 is a range a user can actually type.

Why this is worse than the state before this commit: an over-long name used to be rejected by the provider, which 400s the request. Loud, but correct and immediately visible. Now the name is legal, the request goes through, and `_find_tool` (`react.py:3322`) returns the first tool whose name matches:

```python
def _find_tool(self, name, tools):
    for tool in tools:
        if self._tool_name(tool) == name:
            return tool
```

So the model asks for `read_file` and gets `delete_all_files`, silently. A loud failure was traded for a quiet wrong-tool dispatch, and destructive tools are in the candidate set.

The suggestion is not to revisit the id-suffix decision. It is to spend the budget on the prefix instead of the tool name, which keeps tools on one server distinct without adding anything new:

```python
# today: cuts the tool name off
return sanitized[:MAX_AGENT_TOOL_NAME_LENGTH]

# suggested: keep the tool name whole, squeeze the server prefix
budget_for_prefix = MAX_AGENT_TOOL_NAME_LENGTH - len(sanitized_tool)
return f"{sanitized_prefix[:budget_for_prefix]}{sanitized_tool}"
```

Cross-server collisions remain possible, but those predate this PR. The regression introduced here is the same-server case.

src/xagent/core/tools/adapters/vibe/api_tool_adapter.py
```diff
@@ -134,11 +135,25 @@ def __init__(
...
+        # Same failure mode as the character check above -- an over-long
+        # name is rejected by the same providers, just on length instead of
+        # charset. Budget the truncation around the fixed prefix/suffix so
+        # `_call` (the part that signals "this is an API call") survives
+        # instead of being cut off by a blind truncation of the whole
+        # string. `MAX_AGENT_TOOL_NAME_LENGTH` is this repo's own record of
+        # that provider limit (agent_tool_names.py), shared here rather than
+        # redeclared so the two adapters can't drift apart on the number.
+        budget = MAX_AGENT_TOOL_NAME_LENGTH - len(prefix) - len(suffix)
+        self._name = f"{prefix}{sanitized_name[:budget]}{suffix}"
```
AlexLiu190625, line 171: Same root cause as the MCP comment, different shape, so it needs its own decision rather than the same patch.

`budget = 64 - 4 - 5 = 55`, so two Custom APIs collide as soon as their first 55 sanitized characters agree. Measured on this commit:

```
"Acme Billing Reconciliation Service For The Payments Team Alpha"
"Acme Billing Reconciliation Service For The Payments Team Beta"

both -> api_Acme_Billing_Reconciliation_Service_For_The_Payments_Te_call
```

`custom_apis.name` is `String(100)` with `unique=True`, so both rows are perfectly legal to create, and `CustomApiCreate.name` allows up to 100 characters. Long descriptive names that share a prefix and differ in a trailing qualifier are a normal naming habit.

The MCP fix does not transfer here: there is no server prefix to squeeze, since the whole string is one user-supplied name. The realistic options look like a short hash suffix inside the budget, or accepting the collision and documenting it. I do not have a strong preference between them — flagging it so the choice is made explicitly rather than inherited from the MCP side.

tests/core/tools/adapters/vibe/test_mcp_adapter.py
```diff
@@ -269,6 +271,71 @@ def test_mcp_tool_adapter_source_server_defaults_none():
...
+        description="A tool with an absurdly long name",
+        inputSchema={"type": "object", "properties": {}},
+    )
+    adapter = _build_mcp_tool_adapter(
+        "Coding MCP",
+        {"transport": "streamable_http", "url": "http://127.0.0.1:8642/mcp"},
+        mcp_tool,
+    )
+
+    assert len(adapter.name) == MAX_AGENT_TOOL_NAME_LENGTH
```
AlexLiu190625, line 338: These two assertions pin the length and the prefix, but nothing pins distinctness, which is why the collision above passes the suite.

I wrote the missing assertion and ran it against unmodified `1b6405a9`:

```python
def test_two_tools_on_a_long_named_server_keep_distinct_llm_names():
    server = "s" * 59          # legal; web/api/mcp.py bounds server names at 100
    assert _name(server, "read_file") != _name(server, "delete_all_files")
```

```
AssertionError: assert 'mcp_sss..._' != 'mcp_sss..._'
  where left  = _name(<59-char server>, 'read_file')
  and   right = _name(<59-char server>, 'delete_all_files')
```

Red on current HEAD. Worth adding alongside whichever fix you pick, since the length assertion on its own stays green either way.

AlexLiu190625 (commented):
Re-reviewed at `1b6405a9`.

First, a correction you are right about: my note said `agent_tool_names.py` bounds the length, but I only quoted `_semantic_slug`, which does not truncate. Truncation lives in its caller `_semantic_agent_tool_name`. My evidence block was incomplete and could have been read as "reuse `_semantic_slug` and truncation comes for free". It does not.

Three of the four points are settled from my side:

- Stale comment — gone.
- `CustomApiTool` character sanitization — fixed, and the parametrized tests pin it.
- Test coverage — the character cases are parametrized, and I verified the new length tests are mutation-effective (deleting the truncation line turns them red on `assert 215 == 64`, not on an environment error).

The deliberate non-adoptions (no shared helper, no pinyin, no collision id suffix) all read as reasonable calls to me, and I am not asking you to revisit them on their own.

One new problem, though, and it comes from the length cap that was added in this commit rather than from anything in the original PR. The truncation does not preserve distinctness, and the combination of "no id suffix" plus "blind truncation" is what creates it — either decision alone is fine. Details inline on the MCP adapter; the Custom API side has the same root cause but a different shape, so I left that separately.

93 tests passing confirmed locally (72 + 21, exit 0).

AlexLiu190625 (approved):

PR #1343, opened 2026-08-13
feat: let a new standalone task pick its execution mode

bluefish-08:
## What

The Agent Builder has an execution-mode selector, but the Task page had none: a
new standalone task always fell back to the server default (`auto`), with no way
to force flash / balanced / think.

This adds a mode picker to the composer toolbar on the Task home page.

## Scope

The picker only shows for a **new standalone task** — no selected agent, no
template, not an existing task's composer:

`!compact && !hideConfig && !readOnlyConfig && !taskConfig && !selectedTemplate`

Everything else is deliberately excluded:

- **Existing task** — the backend takes `execution_mode` at task creation only
  (`POST /api/chat/task/create`); there is no update path, so a control that
  looks live but changes nothing would be worse than no control.
- **Selected agent / agent detail page / widgets / build preview** — the agent's
  own mode wins: `resolve_task_runtime_config_core` overrides the task's mode
  with `agent_config.execution_mode` for any non-workforce agent task.
- **Template quick-access** — `handleSend` resolves the template into an agent,
  so this is the agent case above. Without this exclusion the picker would show
  and silently do nothing.

## Backend

None. `execution_mode` was already accepted at task creation, `ExecutionMode`
already has `AUTO`, and `app-context-chat` already forwarded
`config.executionMode.mode`.

## No default change

The picker starts unset rather than at `auto`, so an untouched composer still
sends no `execution_mode` and `get_default_task_execution_mode` keeps deciding
(which matters for `XAGENT_AGENT_RUNTIME=v1`, where the standalone default is
`think`, not `auto`). The button label reads "Auto" because that is the default
for the standalone runtime.

## UI note

The menu is a `Popover`, not `ui/select`: the composer form is `overflow-hidden`
for its rounded corners and clips a plain absolute dropdown. This matches the
`+` menu already sitting in the same toolbar.

## Tests

Four cases in `ChatInput.test.tsx`, each verified by mutation — dropping
`!taskConfig`, dropping `!selectedTemplate`, or hardcoding the initial mode back
to `auto` each makes one of them fail:

- picks think → sends `{ mode: "think" }`
- untouched picker → no `executionMode` in the payload
- template selected → no picker
- existing task → no picker, still forwards the task's own mode

`frontend`: lint clean, `tsc --noEmit` clean, 595 tests pass across
`components/chat`, `components/task`, `components/widget`, `app`.

### DIFF
--- a/frontend/src/components/chat/ChatInput.test.tsx
+++ b/frontend/src/components/chat/ChatInput.test.tsx
@@ -1134,6 +1134,7 @@ describe("ChatInput", () => {
 
   const executionModeLabel = (mode: string) =>
     `builds.configForm.executionMode.${mode}.title`
+  const UNSET_MODE_LABEL = "builds.configForm.executionMode.unset"
 
   it("sends the picked execution mode for a new standalone task", async () => {
     mockDefaultModel()
@@ -1147,7 +1148,7 @@ describe("ChatInput", () => {
       />
     )
 
-    const trigger = await screen.findByText(executionModeLabel("auto"))
+    const trigger = await screen.findByText(UNSET_MODE_LABEL)
     fireEvent.click(trigger)
     fireEvent.click(screen.getByText(executionModeLabel("think")))
     fireEvent.submit(container.querySelector("form") as HTMLFormElement)
@@ -1172,10 +1173,9 @@ describe("ChatInput", () => {
       />
     )
 
-    fireEvent.click(await screen.findByText(executionModeLabel("auto")))
-    const menuItem = screen.getAllByText(executionModeLabel("auto"))
-      .find((node) => node.closest('[role="dialog"]')) as HTMLElement
-    fireEvent.click(menuItem)
+    fireEvent.click(await screen.findByText(UNSET_MODE_LABEL))
+    fireEvent.click(screen.getByText(executionModeLabel("auto")))
+    expect(screen.getByText(executionModeLabel("auto"))).toBeInTheDocument()
     fireEvent.submit(container.querySelector("form") as HTMLFormElement)
 
     await waitFor(() => {
@@ -1198,7 +1198,7 @@ describe("ChatInput", () => {
       />
     )
 
-    await screen.findByText(executionModeLabel("auto"))
+    await screen.findByText(UNSET_MODE_LABEL)
     fireEvent.submit(container.querySelector("form") as HTMLFormElement)
 
     await waitFor(() => expect(onSend).toHaveBeenCalledOnce())
@@ -1217,7 +1217,7 @@ describe("ChatInput", () => {
       />
     )
 
-    expect(screen.queryByText(executionModeLabel("auto"))).not.toBeInTheDocument()
+    expect(screen.queryByText(UNSET_MODE_LABEL)).not.toBeInTheDocument()
   })
 
   it("drops a pick made before the picker was hidden", async () => {
@@ -1231,7 +1231,7 @@ describe("ChatInput", () => {
     }
     const { container, rerender } = render(<ChatInput {...props} />)
 
-    fireEvent.click(await screen.findByText(executionModeLabel("auto")))
+    fireEvent.click(await screen.findByText(UNSET_MODE_LABEL))
     fireEvent.click(screen.getByText(executionModeLabel("think")))
     expect(screen.getByText(executionModeLabel("think"))).toBeInTheDocument()
 
@@ -1240,7 +1240,7 @@ describe("ChatInput", () => {
     )
     rerender(<ChatInput {...props} />)
 
-    expect(screen.getByText(executionModeLabel("auto"))).toBeInTheDocument()
+    expect(screen.getByText(UNSET_MODE_LABEL)).toBeInTheDocument()
     fireEvent.submit(container.querySelector("form") as HTMLFormElement)
 
     await waitFor(() => expect(onSend).toHaveBeenCalledOnce())
@@ -1259,7 +1259,7 @@ describe("ChatInput", () => {
       />
     )
 
-    fireEvent.click(await screen.findByText(executionModeLabel("auto")))
+    fireEvent.click(await screen.findByText(UNSET_MODE_LABEL))
     expect(screen.getByText(executionModeLabel("think"))).toBeInTheDocument()
     fireEvent.submit(container.querySelector("form") as HTMLFormElement)
 
@@ -1281,7 +1281,7 @@ describe("ChatInput", () => {
       />
     )
 
-    expect(screen.queryByText(executionModeLabel("auto"))).not.toBeInTheDocument()
+    expect(screen.queryByText(UNSET_MODE_LABEL)).not.toBeInTheDocument()
   })
 
   it("hides the execution mode picker when composer config is hidden", async () => {
@@ -1296,7 +1296,7 @@ describe("ChatInput", () => {
       />
     )
 
-    expect(screen.queryByText(executionModeLabel("auto"))).not.toBeInTheDocument()
+    expect(screen.queryByText(UNSET_MODE_LABEL)).not.toBeInTheDocument()
   })
 
   it("hides the execution mode picker for an existing task and keeps its mode", async () => {
--- a/frontend/src/components/chat/ChatInput.test.tsx
+++ b/frontend/src/components/chat/ChatInput.test.tsx
@@ -1118,7 +1118,7 @@ describe("ChatInput", () => {
     )
   })
 
-  it("sends the picked execution mode for a new standalone task", async () => {
+  const mockDefaultModel = () => {
     apiRequestMock.mockImplementation((url: string) => {
       if (url === "http://api.local/api/models/?category=llm") {
         return Promise.resolve(
@@ -1130,6 +1130,13 @@ describe("ChatInput", () => {
       }
       return Promise.resolve(emptyJsonResponse())
     })
+  }
+
+  const executionModeLabel = (mode: string) =>
+    `builds.configForm.executionMode.${mode}.title`
+
+  it("sends the picked execution mode for a new standalone task", async () => {
+    mockDefaultModel()
     const onSend = vi.fn()
     const { container } = render(
       <ChatInput
@@ -1140,9 +1147,9 @@ describe("ChatInput", () => {
       />
     )
 
-    const trigger = await screen.findByText("builds.configForm.executionMode.auto.title")
+    const trigger = await screen.findByText(executionModeLabel("auto"))
     fireEvent.click(trigger)
-    fireEvent.click(screen.getByText("builds.configForm.executionMode.think.title"))
+    fireEvent.click(screen.getByText(executionModeLabel("think")))
     fireEvent.submit(container.querySelector("form") as HTMLFormElement)
 
     await waitFor(() => {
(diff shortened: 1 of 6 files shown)


frontend/src/components/chat/ChatInput.tsx
```diff
@@ -587,6 +592,10 @@ export function ChatInput({
     !!isLoading &&
     !allowsLiveGuidanceInput &&
     !isStoppedTaskStatus(normalizedTaskStatus);
+  // Only a new standalone task: the backend takes execution_mode at creation
+  // only, and an agent (a template resolves into one) overrides it anyway.
+  const showExecutionModePicker =
```
rogercloud, line 597: The picker's visibility is encoded as a negative 5-condition gate (`!compact && !hideConfig && !readOnlyConfig && !taskConfig && !selectedTemplate`), which mixes chrome-visibility concerns with payload semantics and fails open: a future call site that forgets one condition gets a picker whose pick is silently ignored. The component already receives `selectedAgents`, which is omitted here — the page (`task/page.tsx:326-327`) already knows the positive condition (`selectedAgents.length === 0 && !selectedTemplate`). Suggest naming the concept: an explicit opt-in prop (e.g. `isNewStandaloneTask`) or a documented derived boolean with the five conditions collapsed inside it. All 7 current call sites are verified correct, so this is a maintainability note rather than a blocker.

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -1137,6 +1148,45 @@ export function ChatInput({
...
+                  <Popover open={executionModeMenuOpen} onOpenChange={setExecutionModeMenuOpen}>
+                    <PopoverTrigger
+                      type="button"
+                      className="inline-flex h-9 items-center gap-2 rounded-xl px-3 text-xs text-muted-foreground hover:bg-secondary/80 hover:text-foreground disabled:pointer-events-none disabled:opacity-50"
+                      disabled={isInputBusy}
+                      title={t("builds.configForm.executionMode.label")}
+                    >
+                      <Sparkles className="h-4 w-4" />
+                      <span className="max-w-[150px] truncate hidden sm:inline-block">
+                        {t(`builds.configForm.executionMode.${pickedExecutionMode ?? "auto"}.title`)}
```
rogercloud: An untouched composer renders the picker trigger label as "Auto" (`pickedExecutionMode ?? "auto"`), which is indistinguishable from an explicit auto pick, and after a pick there is no way back to the unset state (menu items only set modes; the highlight only fires on equality). Under `XAGENT_AGENT_RUNTIME=v1` an untouched composer actually resolves to `think` (`src/xagent/config.py:296-319`), so the label is factually wrong there. Suggest a distinct "Default" label for the unset state (or documenting the v1 divergence), and optionally a way to clear back to unset.

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -165,6 +167,9 @@ export function ChatInput({
   const [isFocused, setIsFocused] = useState(false);
   const [showNoModelAlert, setShowNoModelAlert] = useState(false);
   const [isDraggingFiles, setIsDraggingFiles] = useState(false);
+  // Undefined until picked, so an untouched composer keeps the server default.
+  const [pickedExecutionMode, setPickedExecutionMode] = useState<ExecutionMode>();
```
rogercloud, line 171: `pickedExecutionMode` is never reset after a successful send, unlike its siblings (`localBrowserTarget`, `taskRuntimeSelection`, reset around lines 754-755), so a second standalone task in the same composer silently inherits the previous pick — contradicting this state's own comment ("untouched composer keeps the server default"). Pick-then-select-template also silently discards the pick (the payload falls back to `taskConfig?.executionMode = undefined`); that is acceptable by design per the PR description (template → agent, the agent's mode wins), but a comment or test would lock it in. Suggest resetting after send mirroring line 754, or documenting the sticky intent, plus a stale-drop test.

frontend/src/components/chat/ChatInput.test.tsx
```diff
@@ -1118,6 +1118,112 @@ describe("ChatInput", () => {
     )
   })
 
+  it("sends the picked execution mode for a new standalone task", async () => {
+    apiRequestMock.mockImplementation((url: string) => {
```
rogercloud, line 1122: These tests pin the positive path and two gate terms, but three of the five gate conditions have zero absence assertions: dropping `!compact`, `!hideConfig`, or `!readOnlyConfig` would ship green (only `!taskConfig` — test 4 — and `!selectedTemplate` — test 3 — are pinned). There is also no explicit-auto-pick → `{ mode: "auto" }` test and no pick-then-template stale-drop test. The author's three named mutation claims were verified accurate; suggest adding a `readOnlyConfig`-absence test and an explicit-auto test (the transition test is optional).

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -1137,6 +1148,45 @@ export function ChatInput({
...
                   </>
                 )}
+                {showExecutionModePicker && (
+                  // Popover, not ui/select: the form is overflow-hidden and clips it.
+                  <Popover open={executionModeMenuOpen} onOpenChange={setExecutionModeMenuOpen}>
+                    <PopoverTrigger
+                      type="button"
+                      className="inline-flex h-9 items-center gap-2 rounded-xl px-3 text-xs text-muted-foreground hover:bg-secondary/80 hover:text-foreground disabled:pointer-events-none disabled:opacity-50"
+                      disabled={isInputBusy}
+                      title={t("builds.configForm.executionMode.label")}
```
rogercloud, line 1171: Below the `sm` breakpoint the trigger is an icon-only button — the mode label span is `hidden sm:inline-block` — so its only accessible name is the `title` attribute, which not all assistive tech exposes reliably. The menu items also carry no selection semantics (no `aria-pressed`/`aria-checked`), while repo conventions are mixed (`LocalBrowserMenu`'s "+" button uses `aria-label`; segmented tabs use `aria-pressed`). Suggest an `aria-label` on the trigger that includes the current mode, and `aria-pressed` (or `aria-checked`) on the menu items.

frontend/src/components/chat/ChatInput.tsx
```diff
@@ -1137,6 +1148,45 @@ export function ChatInput({
                     )}
                   </>
                 )}
+                {showExecutionModePicker && (
+                  // Popover, not ui/select: the form is overflow-hidden and clips it.
+                  <Popover open={executionModeMenuOpen} onOpenChange={setExecutionModeMenuOpen}>
```
rogercloud, line 1166: Submitting while the popover is open (Enter in the input with the menu open, or the Tab+Enter keyboard path) leaves the menu floating after send — the only close paths are pick, outside-click, and Escape. Suggest closing the menu (`setExecutionModeMenuOpen(false)`) after a successful send, alongside the existing resets around line 754.

rogercloud (approved):
## Summary

This PR adds an execution-mode picker (auto / flash / balanced / think) to the Task home page composer toolbar, shown only for a brand-new standalone task (no agent, no template, no existing-task config). Untouched, it sends no `execution_mode`, so the server default keeps deciding (auto, or think under `XAGENT_AGENT_RUNTIME=v1`); a pick is forwarded through the existing `config.executionMode.mode` → `execution_mode` task-creation wire. Purely frontend (6 files, +168/−4): `ChatInput.tsx`, `task/page.tsx` type widening, `app-context-chat.tsx` `Task` interface widening, en/zh i18n `auto` entries, and 4 new tests.

## Approach

Acceptable with reservations. This is the simplest design that solves the stated problem: a dedicated picker wired into the existing `execution_mode` creation path, with the untouched state deliberately sending nothing so the server keeps deciding. Every PR-description backend claim was verified against source (creation-only `execution_mode`, agent-config override for non-workforce tasks, standalone defaults auto/think-v1, Popover-over-select rationale, app-context forwarding). The reservations are the two points below: the negative visibility gate (F1) and the unset state rendered as "Auto" (F2).

## Findings (all minor)

**F1 — Negative 5-condition gate for the picker (design, minor)** — `ChatInput.tsx:597`

The picker's visibility is encoded as a negative gate (`!compact && !hideConfig && !readOnlyConfig && !taskConfig && !selectedTemplate`) that mixes chrome-visibility with payload semantics and fails open: a future call site that omits one condition gets a picker whose pick is silently ignored. It also omits the `selectedAgents` prop the component already receives, even though the page already knows the positive condition (`selectedAgents.length === 0 && !selectedTemplate`, `task/page.tsx:326-327`). All 7 current callers were verified correct; the suggestion is to name the concept — an explicit opt-in prop or a documented derived boolean.

**F2 — Unset state is rendered as "Auto" (correctness/UX, minor)** — `ChatInput.tsx:1162`

An untouched composer renders "Auto" (`pickedExecutionMode ?? "auto"`), indistinguishable from an explicit auto pick, and there is no path back to unset after a pick. Under `XAGENT_AGENT_RUNTIME=v1` the untouched composer actually resolves to `think` (`src/xagent/config.py:296-319`), so the label is factually wrong there. Suggest a distinct "Default" label for the unset state (or documenting the v1 divergence), optionally with a way to clear back to unset.

**F3 — `pickedExecutionMode` lifecycle gaps (state, minor)** — `ChatInput.tsx:171`

Unlike its siblings (`localBrowserTarget`, `taskRuntimeSelection`, reset around lines 754-755), the pick is never reset after a successful send, so a second standalone task silently inherits it — contradicting the state's own comment. Pick-then-select-template silently drops the pick (payload falls back to `taskConfig?.executionMode = undefined`); that is acceptable by design per the PR description (template → agent, the agent's mode wins), but a comment or test would lock it in.

**F4 — Gate coverage gaps in the new tests (test quality, minor)** — `ChatInput.test.tsx:1122`

Of the five gate conditions, `!compact`, `!hideConfig`, and `!readOnlyConfig` have zero absence assertions — dropping any of them ships green; only `!taskConfig` (test 4) and `!selectedTemplate` (test 3) are pinned. There is also no explicit-auto-pick → `{ mode: "auto" }` test and no pick-then-template stale-drop test. The author's three named mutation claims were verified accurate; the suggestion is to add a `readOnlyConfig`-absence test and an explicit-auto test (transition test optional).

**F5 — Accessible naming and selection semantics for the picker (a11y, minor)** — `ChatInput.tsx:1158`

Below the `sm` breakpoint the trigger is an icon-only button (the mode label span is `hidden sm:inline-block`), so its only accessible name is the `title` attribute, which not all assistive tech exposes reliably; the menu items also carry no selection semantics (no `aria-pressed`/`aria-checked`). Repo convention is mixed (`LocalBrowserMenu`'s "+" button uses `aria-label`; segmented tabs use `aria-pressed`). Suggest an `aria-label` on the trigger that includes the current mode, and `aria-pressed`/`aria-checked` on the items.

**F6 — Menu left floating after submit (UX, minor)** — `ChatInput.tsx:1153`

Submitting while the popover is open (Enter in the input with the menu open, or the Tab+Enter keyboard path) leaves the menu floating after send — the only close paths are pick, outside-click, and Escape. Suggest closing the menu (`setExecutionModeMenuOpen(false)`) after a successful send, alongside the existing resets around line 754.

**Blocking: no** — no critical or major issues found; all six findings are minor and non-blocking, and the suggested changes (naming the gate concept, distinct "Default" label, state-reset handling, extra test assertions, a11y attributes, closing the menu on send) can land as follow-ups. CI is green: the only job still in progress at review time was "Pytest Fast (web)", with no failures observed in any completed job. Recommended event: **APPROVE**.

PR #1344, opened 2026-08-13
fix(steps): project dag step failures instead of reporting completion

AlexLiu190625:
## Why

`dag_step_end` carries the step's own terminal status in `data['status']` — `'failed'` or `'completed'` — but the public step projector's thinking-family end branch hardcoded `status='completed'`. A DAG step that actually failed was reported to SDK and API consumers as completed.

## What

The end branch now projects the status the event itself reports. Anything other than the literal `'failed'`, including a missing key, maps to `completed`: every current producer sets the field explicitly, so the fallback only guards a malformed or legacy row rather than swallowing a real failure. The `status` field description is updated to name both failure signals — `success=false` (or a dedicated `*_failed` event) for tool events, `status='failed'` for step events.

## Backfill note

Existing tasks are affected too, since `steps()` replays the same projection. A task whose DAG step took a failure path will read `failed` instead of `completed` once the response cache (30s by default) expires. This is not a contract change: `failed` has always been part of `PublicStepStatus`; DAG steps simply never produced it before.

## Verification

Three cases added to `test_steps_mapping.py`: `status='failed'` projects failed, `status='completed'` projects completed, and a missing `status` key projects completed (pinning the fallback direction). A failed-step fixture was added to the projector/batch equivalence set so the new behavior is pinned on both the incremental and batch paths. Full `tests/web/api/v1` suite green.

### DIFF
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -56,7 +56,10 @@ Pairing rule:
         but has no current producer. A per-invocation id is the only
         safe key when the same tool is called twice in the same step.
       - ``react_action_*`` events pair on ``step_id``.
-      - ``dag_step_*`` events pair on ``step_id``.
+      - ``dag_step_*`` events pair on ``step_id``. Both families'
+        end events project their own ``data['status']``: ``"failed"``
+        stays failed, anything else (including a missing key)
+        becomes ``"completed"``.
       - ``dag_plan_*`` events pair on ``task_id`` (single planning
         phase per task; no per-plan identifier available).
       - ``dag_execution`` events pair on ``data['phase']``: a
@@ -285,12 +288,17 @@ class PublicStepProjector:
                 self._pending[("thinking", key)] = step
                 return [step]
             if event_type.endswith("_end"):
+                # dag_step_end always carries its own data['status']
+                # ("failed" or "completed"); react_action_end has no
+                # current producer but is covered by the same rule for
+                # when one appears. A missing key defaults to
+                # "completed" -- see _terminal_status_from_event.
                 finalized = _finalize_pending(
                     self._pending,
                     self._finished,
                     ("thinking", key),
                     end_event=event,
-                    status="completed",
+                    status=_terminal_status_from_event(event),
                 )
                 return [finalized] if finalized is not None else []
             # Events in these families that are neither a start nor an
@@ -642,6 +650,20 @@ def _build_message_step(event: Any, *, role: str) -> Dict[str, Any]:
 
 
 # ===== shared finalization =====
+
+
+def _terminal_status_from_event(event: Any) -> str:
+    """Project a thinking-family end event's own ``data['status']``.
+
+    ``dag_step_end`` (and, theoretically, ``react_action_end``) carries
+    the step's own terminal status: ``"failed"`` or ``"completed"``. We
+    surface it as-is, mapping anything other than the literal string
+    ``"failed"`` -- including a missing key -- to ``"completed"``. Every
+    current ``dag_step_end`` producer sets the field explicitly; the
+    completed fallback only guards a malformed or legacy event, not a
+    real failure being swallowed.
+    """
+    return "failed" if _data_get(event, "status") == "failed" else "completed"
 
 
 def _finalize_pending(
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -55,11 +55,20 @@ Pairing rule:
         ``tool_execution_id`` is accepted ahead of it for compatibility
         but has no current producer. A per-invocation id is the only
         safe key when the same tool is called twice in the same step.
-      - ``react_action_*`` events pair on ``step_id``.
-      - ``dag_step_*`` events pair on ``step_id``. Both families'
-        end events project their own ``data['status']``: ``"failed"``
-        stays failed, anything else (including a missing key)
-        becomes ``"completed"``.
+      - ``react_action_*`` events pair on ``step_id``. ``react_action_end``
+        has no current producer -- the (ACTION, END, REACT) combination
+        is unreachable in this codebase today -- so the rule below is
+        untested for this family, not an observed fact.
+      - ``dag_step_*`` events pair on ``step_id``. Its end events
+        project their own ``data['status']``: ``"failed"`` stays
+        failed, anything else (including a missing key) becomes
+        ``"completed"``. ``react_action_end`` would be folded through
+        this same rule if one were ever emitted, but the rest of the
+        ACTION family (``tool_execution_end``, below) instead reads
+        ``data['success']`` -- so a future ``react_action_end``
+        producer that follows that existing sibling convention rather
+        than ``dag_step_*``'s would have its failures silently
+        projected as ``"completed"``.
       - ``dag_plan_*`` events pair on ``task_id`` (single planning
         phase per task; no per-plan identifier available).
       - ``dag_execution`` events pair on ``data['phase']``: a
@@ -293,6 +302,10 @@ class PublicStepProjector:
                 # current producer but is covered by the same rule for
                 # when one appears. A missing key defaults to
                 # "completed" -- see _terminal_status_from_event.
+                # No extra_data_fn here on purpose: a failed thinking
+                # step doesn't carry an "error" field. Its data shape
+                # is the existing contract ({"phase": ...} only), unlike
+                # the tool family below, which does attach "error".
                 finalized = _finalize_pending(
                     self._pending,
                     self._finished,
--- a/src/xagent/web/schemas/v1.py
+++ b/src/xagent/web/schemas/v1.py
@@ -678,7 +678,9 @@ class PublicStep(BaseModel):
         description=(
             "running while the corresponding end event has not yet "
             "fired, completed on a normal end event, failed when the "
-            "end event carries success=false. A task in a terminal "
+            "end event reports failure (success=false or a dedicated "
+            "*_failed event for tool events, status='failed' for step "
+            "events). A task in a terminal "
             "state can still contain running steps: interrupted or "
             "cancelled steps never get an end event, so terminal "
             "state is judged from the task's own status, not from "
--- a/src/xagent/web/schemas/v1.py
+++ b/src/xagent/web/schemas/v1.py
@@ -634,8 +634,11 @@ PublicStepType = Literal["thinking", "tool_call", "agent_delegation", "message"]
 
 # ``running`` means a start event was seen with no matching end; a
 # terminal task can still contain running steps (interrupted/cancelled
-# steps never get an end event). ``completed`` / ``failed`` reflect
-# the end event's success flag.
+# steps never get an end event). ``completed`` / ``failed`` are read
+# off the end event, but the field that carries the outcome differs by
+# family: tool events use ``success`` (or a dedicated ``*_failed``
+# event), step events use their own ``data['status']``. See the
+# ``PublicStep.status`` field description below for the full rule.
 PublicStepStatus = Literal["running", "completed", "failed"]
 
 
--- a/tests/web/api/v1/test_steps_mapping.py
+++ b/tests/web/api/v1/test_steps_mapping.py
@@ -145,6 +145,32 @@ def test_dag_step_end_missing_status_key_defaults_to_completed():
     assert steps[0]["status"] == "completed"
 
 
+@pytest.mark.parametrize("raw_status", ["interrupted", "cancelled", None])
+def test_dag_step_end_unrecognized_status_falls_back_to_completed(raw_status):
+    """Any ``data['status']`` other than the literal ``"failed"`` must
+    fold to ``"completed"`` -- not pass through verbatim. This pins the
+    fallback against a refactor that swaps the dedicated helper for a
+    direct ``_data_get(event, "status", "completed")`` passthrough:
+    the three existing status tests above (``"failed"``, ``"completed"``,
+    missing key) all coincide with a passthrough's output and would
+    stay green even if an unrecognized value like ``"interrupted"``
+    leaked straight into ``PublicStepStatus``, which does not accept it.
+    """
+    data = {} if raw_status is None else {"status": raw_status}
+    events = [
+        _ev("dag_step_start", step_id="s2"),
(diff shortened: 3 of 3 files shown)


rogercloud (approved):
## PR summary

The internal DAG emits `dag_step_end` events whose data dict already carries the step's own terminal status — 'failed' on the exception and success=false paths, 'completed' otherwise — but the public step projector's thinking-family end branch hardcoded `status='completed'` when finalizing the pending step, so SDK/API consumers of `GET /v1/chat/tasks/{id}/steps` saw genuinely failed DAG steps reported as completed. This PR projects the event's own status through a new pure helper `_terminal_status_from_event` ('failed' iff `data['status']=='failed'`, else 'completed', missing key included), updates the schema `Field` description, and adds three direct-call tests plus one equivalence fixture. All three current producers emit status explicitly ('failed' x2, 'completed' x1); interrupted/waiting paths emit no end event so steps stay 'running' (schema-documented); legacy rows without the key project 'completed' — byte-identical to pre-PR behavior, so no compatibility regression.

## Approach

Sound: minimal root-cause fix at the right layer (the public projector), consistent with the module's existing conventions, and covered by focused tests that would fail on a regression to the hardcoded status. Merge-worthy as-is; all findings below are minor and optional.

## Findings (all minor)

**F1 — `src/xagent/web/api/v1/_step_mapping.py:666` (helper return) — Minor.** Success-biased fallback: an explicit-but-unrecognized status value (a future 'interrupted'/'cancelled') takes the identical 'completed' path as a missing key, silently re-creating the false-completed bug class this PR fixes. `dag.py:1044` already models `step.status='interrupted'`, and the event data dict is untyped (`dict[str, Any]`) with no validation. Safe today (the single producer family emits only 'failed'/'completed'), and missing-key→'completed' is correct legacy behavior — do not change that. Suggestion: distinguish a missing key from an explicit-but-unrecognized value and log a warning for the latter; do NOT map unrecognized→'failed' (a future 'cancelled' is not a failure); skip explicit mappings until a producer exists.

```python
status = data.get("status")
if status == "failed":
    projected = "failed"
elif status is None:
    projected = "completed"  # legacy rows; pre-PR behavior
else:
    logger.warning("unrecognized step end status %r, projecting 'completed'", status)
    projected = "completed"
```

**F2 — `src/xagent/web/schemas/v1.py:635-638` — Minor.** The module comment above `PublicStepStatus` still says "completed / failed reflect the end event's success flag" — now false for step events, which carry no success flag; failure comes from `data['status']=='failed'`. It contradicts the `Field` description this PR just updated (v1.py:678-685). Align the comment with the new contract.

**F3 — `src/xagent/web/api/v1/_step_mapping.py:296-302` — Minor design note.** Failed thinking steps now surface as 'failed' with zero reason: the end event carries `data['error']` (dag.py:1036, 1100) but the thinking end branch projects no extra data, while the tool_call family does project error via `extra_data_fn` (~L341-366). The {phase}-only thinking shape is a documented pre-existing contract, so this is not a regression — but the PR's own fixtures carry `"error": "boom"` unasserted. Either project the error on the failure path (matching the tool_call precedent) or add a one-line note that the omission is deliberate. Follow-up, not blocking.

**F4 — `src/xagent/web/api/v1/_step_mapping.py:59-61` — Minor.** The module docstring asserts as settled fact that "Both families' end events project their own `data['status']`", but `react_action_end` has no producer anywhere in `src/` (verified: no ACTION+REACT emitter; the `ws_trace_handlers` mapping is synthetic-only). The call-site comment (L291-295) hedges correctly; mirror that hedge here. Note that a future `react_action_end` producer following the ACTION family's actual convention (`data['success']`) would project 'completed' on failure under this rule.

**F5 — `tests/web/api/v1/test_steps_mapping.py:134` — Minor, optional hardening.** The documented "anything other than literal 'failed' → 'completed'" fallback is pinned only via a missing key and an explicit 'completed'; no test feeds a present-but-unrecognized value ('interrupted'/None). A one-line parametrize guards a plausible structural rewrite (e.g. `_data_get(event, 'status', 'completed')`) that would leak 'interrupted' and break the `PublicStepStatus` Literal at serialization.

```python
@pytest.mark.parametrize("status", ["interrupted", "cancelled", None])
```

A `react_action_end`-with-status test is deliberately not recommended: same code path already exercised, and no producer exists.

- `delete: _terminal_status_from_event (L655-666) — single caller (L301); inline the ternary at the call site and fold the helper's docstring rationale (L663-665: producers set 'failed' explicitly, so the fallback only guards malformed/legacy events, not a swallowed failure) into the call-site comment or module docstring. net: ~-5 lines possible (or moot if F1's logging lands in the helper)`

Blocking: no. The PR is merge-worthy as-is; all five findings are minor/optional and the simplification is optional. Recommended event: APPROVE.

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -644,6 +652,20 @@ def _build_message_step(event: Any, *, role: str) -> Dict[str, Any]:
...
+
+    ``dag_step_end`` (and, theoretically, ``react_action_end``) carries
+    the step's own terminal status: ``"failed"`` or ``"completed"``. We
+    surface it as-is, mapping anything other than the literal string
+    ``"failed"`` -- including a missing key -- to ``"completed"``. Every
+    current ``dag_step_end`` producer sets the field explicitly; the
+    completed fallback only guards a malformed or legacy event, not a
+    real failure being swallowed.
+    """
+    return "failed" if _data_get(event, "status") == "failed" else "completed"
```
rogercloud, line 679: Success-biased fallback: an explicit-but-unrecognized status (a future 'interrupted'/'cancelled') takes the identical 'completed' path as a missing key, silently re-creating the false-completed bug class this PR fixes. The data dict is untyped (dict[str, Any]) with no validation, and dag.py:1044 already models step.status='interrupted', so this is one producer change away from being live. Safe today (the only producer family emits 'failed'/'completed'), and missing-key→'completed' must stay for legacy rows. Consider distinguishing missing-key from explicit-but-unrecognized and logging a warning for the latter; do not map unrecognized→'failed' (a future 'cancelled' is not a failure).

src/xagent/web/schemas/v1.py
```diff
@@ -678,7 +678,9 @@ class PublicStep(BaseModel):
         description=(
             "running while the corresponding end event has not yet "
             "fired, completed on a normal end event, failed when the "
-            "end event carries success=false. A task in a terminal "
+            "end event reports failure (success=false or a dedicated "
```
rogercloud, line 684: The module comment above PublicStepStatus (lines 635-638) still claims 'completed / failed reflect the end event's success flag' — now false for step events, which carry no success flag: failure comes from data['status']=='failed'. This contradicts the Field description this PR just updated here. Align the comment with the new contract (e.g. "status is taken from the end event's data['status']").

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -285,12 +288,17 @@ def feed(self, event: Any) -> List[Dict[str, Any]]:
...
+                # current producer but is covered by the same rule for
+                # when one appears. A missing key defaults to
+                # "completed" -- see _terminal_status_from_event.
                 finalized = _finalize_pending(
                     self._pending,
                     self._finished,
                     ("thinking", key),
                     end_event=event,
-                    status="completed",
+                    status=_terminal_status_from_event(event),
```
rogercloud, line 314: Failed thinking steps now surface as 'failed' with no reason attached: the end event carries data['error'] (dag.py:1036, 1100) but this branch projects no extra data, whereas the tool_call family does project error via extra_data_fn (~L341-366). The {phase}-only thinking shape is a documented pre-existing contract, so this is not a regression — but the PR's own fixtures carry "error": "boom" unasserted. Either project the error on the failure path (matching the tool_call precedent) or add a one-line note that the omission is deliberate. Optional follow-up.

src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -56,7 +56,10 @@
         but has no current producer. A per-invocation id is the only
         safe key when the same tool is called twice in the same step.
       - ``react_action_*`` events pair on ``step_id``.
-      - ``dag_step_*`` events pair on ``step_id``.
+      - ``dag_step_*`` events pair on ``step_id``. Both families'
```
rogercloud: This docstring bullet asserts as settled fact that "Both families' end events project their own data['status']", but react_action_end has no producer anywhere in src/ (no ACTION+REACT emitter; the ws_trace_handlers mapping is synthetic-only). The call-site comment (L291-295) hedges correctly; mirror that hedge here. Note that a future react_action_end producer following the ACTION family's actual convention (data['success']) would project 'completed' on failure under this rule.

tests/web/api/v1/test_steps_mapping.py
```diff
@@ -101,6 +101,50 @@ def test_dag_step_pair_becomes_thinking_step():
...
+            data={"status": "completed"},
+            timestamp=datetime(2026, 1, 1, 12, 0, 1, tzinfo=timezone.utc),
+        ),
+    ]
+    steps = map_trace_events_to_public_steps(events)
+    assert len(steps) == 1
+    assert steps[0]["status"] == "completed"
+
+
+def test_dag_step_end_missing_status_key_defaults_to_completed():
```
rogercloud, line 134: Optional hardening: the documented 'anything other than literal failed → completed' fallback is pinned only via a missing key and an explicit 'completed' — no test feeds a present-but-unrecognized value ('interrupted'/None). A one-line parametrize guards a plausible structural rewrite (e.g. _data_get(event, 'status', 'completed')) that would leak 'interrupted' and break the PublicStepStatus Literal at serialization. A react_action_end-with-status test is deliberately not recommended — same code path already exercised, no producer exists.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Not adding the warning, and taking F5 instead — the two facts behind this one change the calculus. `dag.py:1044` sets `step.status` on the in-memory step object, but no emitter reads it: all three `on_dag_step_end` call sites pass hardcoded literals. And the interrupted branch emits no end event at all — it checkpoints and returns, as do the waiting-for-user and re-raise paths. So an unrecognized status can only arrive here after adding an end emission to a branch that deliberately has none, which would contradict the "interrupted or cancelled steps never get an end event" contract already published in the `status` field description.

On the warning itself: this helper runs inside the full-history replay behind a cache keyed on `max_event_id`, so an active task recomputes its whole history on every new event — one malformed historical row would re-log for the rest of that task life. F5 pins the same semantic at zero runtime cost, and it turned out to be load-bearing rather than optional: see that thread.

src/xagent/web/schemas/v1.py
AlexLiu190625: Fixed — the comment no longer claims step events carry a success flag; it now matches the field description this PR updated.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Added the note rather than the error payload. Projecting `error` here would change the thinking step data shape, which is an existing contract (phase only), and the downstream projection for external consumers passes through only `phase` for thinking steps — so the error would be visible on one surface and not the other. The branch now carries a one-line comment recording that omitting `extra_data_fn` is deliberate; attaching a reason to failed thinking steps is worth doing as its own change, on both surfaces at once.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Hedged, and folded in your side note — the docstring now says react_action_end has no producer today, and flags that its ACTION-family sibling reads `data["success"]`, so a future producer copying that convention would project completed on failure.

tests/web/api/v1/test_steps_mapping.py
AlexLiu190625: Taken, and it earned its keep. Verified the concern concretely: replacing the helper with a direct `_data_get(event, "status", "completed")` passthrough leaves all 85 tests in the file green — the three existing status cases (failed / completed / missing key) each coincide with a passthrough result. The new parametrize over interrupted / cancelled / None goes red on exactly that mutation. Without it the documented fallback had no coverage at all, so this was the finding that mattered most in the round.

AlexLiu190625:
Addressed in the commit just pushed. Four of the five findings are in; F1 is the one I did not take, with the reasoning on its thread.

**Fixed.** F2 — the comment above `PublicStepStatus` no longer claims step events carry a success flag, so it stops contradicting the field description. F3 — added the note rather than the payload: projecting `error` into a thinking step would change a data shape that is an existing contract, and the downstream projection for external consumers passes through only `phase` for thinking steps, so the reason would be visible on one surface and missing on the other. Attaching a reason to failed thinking steps is worth doing across both surfaces as its own change. F4 — the pairing-rule bullet now states that `react_action_end` has no producer today, and carries your note that its ACTION-family sibling reads `data['success']`, so a future producer copying that convention would project completed on failure. F5 — taken.

**F5 turned out to be load-bearing rather than optional.** Verified the concern directly: replacing the helper with a `_data_get(event, "status", "completed")` passthrough leaves every test in the file green, because the three existing status cases (failed / completed / missing key) each coincide with what a passthrough would return. The new parametrize over interrupted / cancelled / None goes red on exactly that mutation. Before it, the documented fallback had no coverage at all.

**F1 — not adding the warning.** Two facts moved this one for me. The interrupted branch of `DAGPattern._execute_step` sets `step.status` on the in-memory step object, but no emitter reads it: every `on_dag_step_end` call site passes a hardcoded literal. That same branch emits no end event at all — it checkpoints and returns, as do the waiting-for-user and re-raise paths — so an unrecognized status can only arrive here after adding an end emission to a branch that deliberately has none, which would contradict the "interrupted or cancelled steps never get an end event" contract already published in the status field description. Separately, this helper runs inside the full-history replay behind a cache keyed on the task's max event id, so an active task recomputes its whole history on every new event and one malformed historical row would re-log for the rest of that task's life. F5 pins the same semantic at zero runtime cost.

**Simplification — keeping the helper.** Correct that it has a single caller. Keeping it named is deliberate: its docstring carries this change's audit conclusion — every current producer sets the field explicitly, so the completed fallback guards malformed or legacy events rather than swallowing a real failure — which reads better on a named function than as a six-line comment inside a branch, and it stays directly unit-testable, which is what makes the F5 parametrize a one-liner. The saving would be about five lines.

PR #1345, opened 2026-08-13
Add a team-keyed shared MCP env hook, separate from the user-keyed one

AlexLiu190625:
For a run governed by a team-owned agent, each team-owned MCP server's shared env
layer now resolves from the governing team's own row — never from the runner's
current team, and never from any other team. When the governing team has no row
for a server, that server has no shared layer at all, and a runner whose
env-source pick is `shared` degrades unconditionally to the legacy default
(global env, then the runner's own key).

## What changes

- **`mcp_runtime.py`** gains `set_mcp_team_env_hook`, `load_team_env_overrides`
  and `team_env_hook_installed`, deliberately separate from
  `set_mcp_shared_env_hook`. That older loader degrades to "no layer" on any
  failure, which would hide a signature error or a malformed answer on this new,
  authorization-relevant seam. The failure policy here is the opposite: raise,
  through the same typed `ConnectorRuntimeError` channel `connector_team_scope`
  already uses, with strict shape validation on the answer. A snapshot/restore
  contextmanager covers every hook slot in the module so a fixture cannot
  silently clear one.
- **`config.py`**: `_load_mcp_server_configs` re-keys the shared env layer for
  team-owned ids, immediately after the existing prefetch and inside the same
  guarded region. The block is gated on a governing team being known, its hook
  being installed, and at least one team-owned id in scope, so an ungoverned run
  pays no extra query.
- **`connector_team_scope.py`**: the accepted-risk docstring now also states the
  credential rule.

## Behaviour with no hook installed

Selection is on the installed predicate, never on an empty return — an empty
answer is a legitimate answer, not a missing implementation. With no hook
installed the whole block is skipped. Verified by serialising the full
`_load_mcp_server_configs` result and counting emitted queries against this
branch's base: identical output, identical query count.

## Cost

One extra query per config-cache miss, not per task. The MCP config build is
cached per instance; this adds a query only when that cache misses.

## Tests

Fourteen invariants are pinned across `tests/web/services/test_mcp_team_env_hook.py`
and `tests/web/tools/test_mcp_team_env_wiring.py`, each verified by breaking the
implementation and confirming the intended test — and only that test — fails:

- the shared layer keys on the governing team, for a runner who is a member and
  for one who is not;
- no row for the governing team means no shared layer, and any legacy shared
  value for that id is scrubbed rather than left in place;
- a `shared` pick degrades unconditionally, whether or not the runner has an own
  key, and whether or not the scrub removed anything;
- the runner's own key still wins over the team layer;
- the hook is never consulted when there is no governing team, or when no
  team-owned id is in scope;
- an installed hook answering with an empty mapping still applies the degrade
  (the predicate, not the answer, decides);
- hook failure and every malformed answer shape raise the typed error rather
  than degrading, including the `bool`-as-`int` key case;
- ids outside the team-owned set are untouched;
- the mappings owned by the hook are never mutated in place;
- the hook is always invoked with `team_id` passed as a keyword;
- the snapshot primitive restores every slot in the module by identity.

### DIFF
--- a/src/xagent/web/services/connector_team_scope.py
+++ b/src/xagent/web/services/connector_team_scope.py
@@ -72,7 +72,14 @@ class TeamConnectorVisibilityHook(Protocol):
     MCP transports, a team-owned server with no personal link for the
     running user executes using the shared definition row's own stored
     credentials, not the runner's. (OAuth transports fail closed instead,
-    for lack of a token to resolve.) Custom APIs are starker: there is no
+    for lack of a token to resolve.) For stdio specifically, the shared env
+    layer a team-owned server resolves is keyed on the team this hook
+    answers for -- the team that owns the *governing agent* -- never on any
+    other team, including one the running user happens to belong to: when
+    the governing team has no stored row for that server, there is no
+    shared layer at all, and a runner whose own env-source pick is
+    "shared" falls back to their own stored key instead of silently
+    borrowing another team's row. Custom APIs are starker: there is no
     per-member credential override layer at all, and no transport-level
     exception either -- every custom API always executes with whatever is
     stored on its shared definition row (headers, a static API key, and so
--- a/src/xagent/web/services/connector_team_scope.py
+++ b/src/xagent/web/services/connector_team_scope.py
@@ -72,14 +72,25 @@ class TeamConnectorVisibilityHook(Protocol):
     MCP transports, a team-owned server with no personal link for the
     running user executes using the shared definition row's own stored
     credentials, not the runner's. (OAuth transports fail closed instead,
-    for lack of a token to resolve.) For stdio specifically, the shared env
-    layer a team-owned server resolves is keyed on the team this hook
-    answers for -- the team that owns the *governing agent* -- never on any
-    other team, including one the running user happens to belong to: when
-    the governing team has no stored row for that server, there is no
-    shared layer at all, and a runner whose own env-source pick is
-    "shared" falls back to their own stored key instead of silently
-    borrowing another team's row. Custom APIs are starker: there is no
+    for lack of a token to resolve.) For stdio specifically, **once an
+    application has also installed the separate, credential-side hook**
+    (``mcp_runtime.set_mcp_team_env_hook`` -- a different socket from this
+    one, installed independently), the shared env layer a team-owned
+    server resolves is keyed on the team this hook answers for -- the team
+    that owns the *governing agent* -- never on any other team, including
+    one the running user happens to belong to: when the governing team has
+    no stored row for that server, there is no shared layer at all, and a
+    runner whose own env-source pick is "shared" falls back to their own
+    stored key instead of silently borrowing another team's row. An
+    application that installs only this visibility hook, without also
+    installing ``set_mcp_team_env_hook``, gets none of that: the shared env
+    layer then still resolves however the pre-existing, user-keyed
+    ``set_mcp_shared_env_hook`` was wired, which can and typically does key
+    on the *running user's own* team rather than the governing one --
+    exactly the cross-team leak the paragraph above describes as closed.
+    Deciding to share a team-owned MCP server through this hook is a
+    visibility decision only; closing the credential leak is a second,
+    separate installation step. Custom APIs are starker: there is no
     per-member credential override layer at all, and no transport-level
     exception either -- every custom API always executes with whatever is
     stored on its shared definition row (headers, a static API key, and so
--- a/src/xagent/web/services/mcp_runtime.py
+++ b/src/xagent/web/services/mcp_runtime.py
@@ -1,8 +1,14 @@
 from __future__ import annotations
 
+import logging
+from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Callable
+from typing import Any, Callable, Iterator, Protocol
 
+from ...core.tools.adapters.vibe.connector_runtime import (
+    ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
+    ConnectorRuntimeError,
+)
 from .mcp_oauth import MCPOAuthRuntimeError, resolve_mcp_oauth_runtime_auth
 
 HTTP_MCP_TRANSPORTS = frozenset({"sse", "websocket", "streamable_http"})
@@ -96,6 +102,139 @@ def load_shared_env_overrides(db: Any, user_id: int | None) -> dict[int, dict]:
 
         logging.getLogger(__name__).exception("shared env hook failed")
         return {}
+
+
+class TeamMCPEnvHook(Protocol):
+    """Resolver for a team's own shared MCP env overrides, keyed by server id.
+
+    Typed as a ``Protocol`` with a keyword-only ``team_id`` for the same
+    reason ``TeamConnectorVisibilityHook`` (the visibility twin in
+    connector_team_scope.py) is: this hook and the user-keyed
+    ``_get_mcp_shared_env_hook`` above are otherwise structurally
+    identical -- ``(db, id) -> {mcpserver_id: {env}}`` -- and a swapped
+    install would type-check while resolving an unrelated team's or
+    user's env.
+    """
+
+    def __call__(self, db: Any, *, team_id: int) -> dict[int, dict]: ...
+
+
+# Hook signature: (db, *, team_id) -> {mcpserver_id: {env}}. Deliberately a
+# separate module-global slot from _get_mcp_shared_env_hook, not a keyword on
+# set_mcp_shared_env_hook: the two hooks answer different questions -- "did
+# the application register anything shared at all" (user-keyed, default:
+# none) versus "does THIS SPECIFIC governing team have a row" (team-keyed,
+# authorization-relevant) -- and load_shared_env_overrides's failure policy
+# (degrade to no shared layer) would hide a signature TypeError or a
+# malformed answer from this hook instead of surfacing it.
+_get_mcp_team_env_hook: TeamMCPEnvHook | None = None
+
+
+def set_mcp_team_env_hook(hook: TeamMCPEnvHook | None) -> None:
+    global _get_mcp_team_env_hook
+    _get_mcp_team_env_hook = hook
+
+
+def team_env_hook_installed() -> bool:
+    """Whether an application installed a team-keyed env hook.
+
+    Callers select on this, never on an empty return value: an installed
+    hook legitimately answers empty for a team that stored nothing for any
+    server (see load_team_env_overrides).
+    """
+    return _get_mcp_team_env_hook is not None
+
+
+def _validate_team_mcp_env_answer(answer: Any) -> dict[int, dict]:
+    """Validate the team-env hook's answer shape.
+
+    This is an authorization input, not user-facing data: a malformed
+    answer must fail loudly, never be normalized, coerced, or defaulted to
+    empty -- the same stance connector_team_scope's
+    ``_validate_team_connector_answer`` takes for the visibility twin. The
+    hook must return a ``dict`` whose keys are all ``int`` server ids
+    (``bool`` is a subclass of ``int`` in Python but is rejected here -- a
+    truthy/falsy value is never a legitimate server id) and whose values
+    are all ``dict`` env mappings.
+    """
+    if not isinstance(answer, dict):
+        raise ValueError(
+            "team env hook returned a malformed answer: expected a dict, "
+            f"got {type(answer).__name__}"
+        )
+    for key, value in answer.items():
+        if isinstance(key, bool) or not isinstance(key, int):
+            raise ValueError(
+                "team env hook returned a malformed answer: key "
+                f"{key!r} is not a server id (must be int, not bool)"
+            )
+        if not isinstance(value, dict):
+            raise ValueError(
+                "team env hook returned a malformed answer: value for key "
+                f"{key!r} must be a dict, got {type(value).__name__}"
+            )
+    # Defensive copy at the authorization boundary: the caller must not be
+    # able to mutate the installing application's own answer object (or
+    # vice versa). Shallow is sufficient -- every downstream consumer only
+    # ever reads the inner per-server env dicts, never mutates them in
+    # place (see the shallow-copy note in config.py's wiring block).
+    return dict(answer)
+
+
+def load_team_env_overrides(db: Any, team_id: int | None) -> dict[int, dict]:
+    """Batch-load a governing team's own shared MCP env overrides, keyed by
+    server id.
+
+    Unlike load_shared_env_overrides, this loader's failure policy is
+    raise, not degrade: any hook exception and any answer shape violation
+    becomes a typed ConnectorRuntimeError instead of silently falling back
+    to no shared layer. A misbehaving hook here must not be
+    indistinguishable from "the governing team legitimately has no row" --
+    the caller (config.py's wiring) already treats the latter as a
+    deliberate, signal-losing degrade of its own; collapsing hook failure
+    into the same outcome would double that loss silently.
+    """
+    if team_id is None or _get_mcp_team_env_hook is None:
+        return {}
+    try:
+        answer = _get_mcp_team_env_hook(db, team_id=team_id)
+        return _validate_team_mcp_env_answer(answer)
+    except ConnectorRuntimeError:
+        raise
+    except Exception as exc:
+        logging.getLogger(__name__).warning("team env hook failed", exc_info=True)
+        raise ConnectorRuntimeError(
+            ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
+            "Team MCP env is unavailable.",
+            details={"reason": "team_env_resolution_failed"},
+            status_code=503,
+        ) from exc
+
+
+@contextmanager
+def mcp_env_hooks_snapshot() -> Iterator[None]:
+    """Save every env hook slot in this module on entry, restore all of
+    them on exit.
+
+    Covers both the pre-existing shared (user-keyed) slot and the
+    team-keyed slot above, so a test that installs either hook is
+    restored to the pre-entry state deterministically without reaching
+    past the public setters. A hook slot added to this module later,
+    following the same ``_get_mcp_*_hook`` naming pattern as the two
+    above, that is not added here as well fails the coverage test in
+    tests/web/services/test_mcp_team_env_hook.py, by design -- see that
+    module's docstring rule. A slot that breaks the naming pattern is
+    outside what that coverage test can discover and must be added here
+    by hand.
+    """
+    global _get_mcp_shared_env_hook, _get_mcp_team_env_hook
+    saved_shared = _get_mcp_shared_env_hook
+    saved_team = _get_mcp_team_env_hook
+    try:
+        yield
+    finally:
+        _get_mcp_shared_env_hook = saved_shared
+        _get_mcp_team_env_hook = saved_team
 
 
 CALLER_ID_ENV_VAR = "XAGENT_MCP_CALLER_ID"
--- a/src/xagent/web/services/mcp_runtime.py
+++ b/src/xagent/web/services/mcp_runtime.py
@@ -75,7 +75,10 @@ def load_user_env_sources(db: Any, user_id: int | None) -> dict[int, str]:
 # Hook signature: (db, user_id) -> {mcpserver_id: {env}}. An application layer
 # (e.g. a multi-tenant deployment) can inject a shared env layer that sits between
 # the global and per-user env via set_mcp_shared_env_hook(). The core is agnostic
-# to what "shared" means (team, org, ...); default: no shared layer.
+# to what "shared" means (team, org, ...); default: no shared layer. For team
+# semantics specifically -- keying the shared layer on the team that owns the
+# governing agent, rather than on the running user -- install the separate,
+# team-keyed socket below instead: see set_mcp_team_env_hook.
 _get_mcp_shared_env_hook: Callable[[Any, int | None], dict[int, dict]] | None = None
 
 
@@ -163,10 +166,16 @@ def _validate_team_mcp_env_answer(answer: Any) -> dict[int, dict]:
             f"got {type(answer).__name__}"
         )
     for key, value in answer.items():
-        if isinstance(key, bool) or not isinstance(key, int):
+        if isinstance(key, bool):
             raise ValueError(
                 "team env hook returned a malformed answer: key "
                 f"{key!r} is not a server id (must be int, not bool)"
+            )
+        if not isinstance(key, int):
+            raise ValueError(
+                "team env hook returned a malformed answer: key "
+                f"{key!r} is not a server id (must be int, got "
+                f"{type(key).__name__})"
             )
         if not isinstance(value, dict):
             raise ValueError(
@@ -197,12 +206,14 @@ def load_team_env_overrides(db: Any, team_id: int | None) -> dict[int, dict]:
     if team_id is None or _get_mcp_team_env_hook is None:
         return {}
     try:
-        answer = _get_mcp_team_env_hook(db, team_id=team_id)
+        answer = _get_mcp_team_env_hook(db, team_id=int(team_id))
         return _validate_team_mcp_env_answer(answer)
     except ConnectorRuntimeError:
         raise
     except Exception as exc:
-        logging.getLogger(__name__).warning("team env hook failed", exc_info=True)
+        logging.getLogger(__name__).warning(
+            "team env hook failed for team %s", team_id, exc_info=True
+        )
         raise ConnectorRuntimeError(
             ERROR_CONNECTOR_RUNTIME_UNAVAILABLE,
             "Team MCP env is unavailable.",
--- a/src/xagent/web/services/mcp_runtime.py
+++ b/src/xagent/web/services/mcp_runtime.py
@@ -158,7 +158,14 @@ def _validate_team_mcp_env_answer(answer: Any) -> dict[int, dict]:
     hook must return a ``dict`` whose keys are all ``int`` server ids
     (``bool`` is a subclass of ``int`` in Python but is rejected here -- a
     truthy/falsy value is never a legitimate server id) and whose values
-    are all ``dict`` env mappings.
+    are all ``dict`` env mappings whose own keys and values are all
+    ``str``: this env eventually becomes a subprocess environment, where a
+    non-string surfaces as an unclassified failure at spawn time rather
+    than as the typed error this boundary promises. There is no ``bool``
+    special case on the inner side, unlike the server-id keys above,
+    because ``bool`` is not a ``str`` subclass and is already rejected.
+    A malformed env value's own content is never formatted into the error
+    message -- these are credentials.
     """
     if not isinstance(answer, dict):
         raise ValueError(
@@ -182,6 +189,19 @@ def _validate_team_mcp_env_answer(answer: Any) -> dict[int, dict]:
                 "team env hook returned a malformed answer: value for key "
                 f"{key!r} must be a dict, got {type(value).__name__}"
             )
+        for env_key, env_value in value.items():
+            if not isinstance(env_key, str):
+                raise ValueError(
+                    "team env hook returned a malformed answer: env key "
+                    f"{env_key!r} under server id {key!r} is not a string "
+                    f"(got {type(env_key).__name__})"
+                )
+            if not isinstance(env_value, str):
+                raise ValueError(
+                    "team env hook returned a malformed answer: env value "
+                    f"for env key {env_key!r} under server id {key!r} is "
+                    f"not a string (got {type(env_value).__name__})"
+                )
     # Defensive copy at the authorization boundary: the caller must not be
     # able to mutate the installing application's own answer object (or
     # vice versa). Shallow is sufficient -- every downstream consumer only
@@ -202,12 +222,26 @@ def load_team_env_overrides(db: Any, team_id: int | None) -> dict[int, dict]:
     the caller (config.py's wiring) already treats the latter as a
     deliberate, signal-losing degrade of its own; collapsing hook failure
     into the same outcome would double that loss silently.
+
+    A row whose env mapping is empty is dropped here rather than passed on:
+    an empty mapping carries nothing, and letting it through as "the team
+    has a row" would make storing an empty row strictly worse for the
+    runner than storing nothing at all -- the caller's row-present branch
+    keeps a "shared" pick that then merges nothing, pinning the run to the
+    global env, whereas a missing row degrades that pick and lets the
+    runner's own key through. load_user_env_overrides above applies the
+    same convention to a per-user override that decrypts to an empty
+    mapping.
     """
-    if team_id is None or _get_mcp_team_env_hook is None:
+    if db is None or team_id is None or _get_mcp_team_env_hook is None:
         return {}
     try:
         answer = _get_mcp_team_env_hook(db, team_id=int(team_id))
-        return _validate_team_mcp_env_answer(answer)
+        validated = _validate_team_mcp_env_answer(answer)
+        # Validate first, then drop rows whose env mapping is empty. The
+        # order matters: dropping first would let a malformed server-id key
+        # with an empty value escape the shape check entirely.
+        return {server_id: env for server_id, env in validated.items() if env}
     except ConnectorRuntimeError:
         raise
     except Exception as exc:
--- a/src/xagent/web/services/mcp_runtime.py
+++ b/src/xagent/web/services/mcp_runtime.py
@@ -148,6 +148,40 @@ def team_env_hook_installed() -> bool:
     return _get_mcp_team_env_hook is not None
 
 
+_team_env_hook_missing_warned = False
+
+
+def warn_team_env_hook_missing_once(*, team_id: int, connector_count: int) -> None:
+    """Warn, once per process, that team-owned connectors are in scope while
+    no team-keyed env hook is installed.
+
+    An application that installed the connector visibility hook but not
+    set_mcp_team_env_hook keeps the pre-existing, user-keyed shared env
+    layer for team-owned servers, which resolves on whatever that hook was
+    wired to -- typically the running user's own team, not the team that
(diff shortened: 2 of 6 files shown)


rogercloud (changes requested):
# Code Review — PR #1345

**Title**: Add a team-keyed shared MCP env hook, separate from the user-keyed one
**Scope**: 5 files, +1117/-4 · CI green · No substantive prior review (only a Gemini bot quota-limit notice)

## 1. Overall assessment

This PR introduces a second credential-resolution path for "MCP servers run by a team-governed agent": `set_mcp_team_env_hook` / `load_team_env_overrides` / `team_env_hook_installed`, alongside the existing user-keyed `set_mcp_shared_env_hook`. The motivation is a real authorization gap — a team-owned MCP server's shared env previously resolved against the *runner's own* `user_id` instead of the *governing team*, allowing cross-team credential borrowing.

**Design verdict: acceptable-with-reservations** (direction is correct; three choices worth reconsidering):
- The old and new "shared env layer" mechanisms are structurally near-identical but use opposite failure semantics — the old one degrades silently, the new one raises and **aborts the entire MCP tool-loading pass** (personal servers included). This is the review's single highest-severity finding (see D1 below).
- The fix is **opt-in and fail-open by default**: a deployment that installs only the visibility hook (not the new env hook) keeps the original cross-team leak, with zero runtime signal (no log, no startup check) — the author documents this two-step install as deliberate, but nothing in the code backs that documentation with a guardrail.
- The actual authorization algorithm (~12 lines) is inlined inside a 113-line method and can't be unit-tested independently of the DB/ORM.

Data-flow tracing is solid (the stdio branch correctly consumes the new tables; the oauth→stdio conversion branch is unaffected), and the test suite is substantial (302+590 lines), but there are two real behavioral gaps (empty-dict semantics, unvalidated inner env value types) and several test-coverage/quality issues.

**PR type**: `fix` (an authorization fix, with a feature-shaped new hook mechanism) — per repo convention this should be `fix(mcp):`, not `feat`.

## 2. Findings

### 🔒 Security

| # | File:Line | Severity | Issue | Suggestion |
|---|---|---|---|---|
| S1 | `src/xagent/web/tools/config.py:3538-3541` | **Medium** | The new team-env re-keying only takes effect when `team_env_hook_installed()` is true. A deployment that installs only the visibility hook (`TeamConnectorVisibilityHook`), not `set_mcp_team_env_hook`, keeps the original cross-team leak — with zero logging and zero startup validation. The author documents this as a deliberate "two-step install" in `connector_team_scope.py:70-99`, but no runtime guardrail backs that documented intent. | Emit a WARNING when the gate is false but `team_mcp_ids` is non-empty; a stronger fix would unconditionally strip the user-keyed shared layer for team-owned ids (fail-closed by default). |
| S2 | `src/xagent/web/services/mcp_runtime.py:151-190` (`_validate_team_mcp_env_answer`) | **Medium** | Only the outer mapping's key/value types are validated; the inner env dict's keys/values are never checked to be `str`. `{5: {"KEY": None}}` or `{5: {123: "value"}}` both pass validation and flow, via `merge_stdio_env`'s untyped `dict.update`, straight into `StdioServerParameters(env=...)` — surfacing only as an unclassified `TypeError` at subprocess spawn time, not the loud `ValueError` this boundary is supposed to guarantee. | Add inner `str`/`str` validation to `_validate_team_mcp_env_answer`, with parametrized tests for non-str inner keys/values. |
| S3 | `src/xagent/web/api/mcp.py:1751-1769, ~1857-1960` (`_app_shared_env_available`) | **Medium** | The directory/UI check for "is a shared key available" still uses the old user-keyed hook (`load_shared_env_overrides(db, current_user.id)`), completely independent of the new governing-team-keyed runtime check (`config.py:3543-3564`). A user can be told in the UI "shared key available", pick `env_source="shared"`, and still silently degrade to their own key at runtime because the governing team has no row. | At minimum document this known UI/runtime inconsistency; longer-term, bring the directory-side check onto the same team-keyed basis, or surface the degradation to the user. |

### 🐛 Defects

| # | File:Line | Severity | Issue | Suggestion |
|---|---|---|---|---|
| D1 | `src/xagent/web/services/mcp_runtime.py:193-222` vs `92-107` | **High** | Two structurally-identical hooks have opposite failure semantics: the old `load_shared_env_overrides` silently degrades to `{}` on any exception; the new `load_team_env_overrides` wraps hook exceptions or validation failures into a raised `ConnectorRuntimeError`. Confirmed that this raise happens inside `config.py:3543`'s prefetch/re-key try block, which completes *before* the per-server loop that has the existing `_build_unavailable_mcp_config` isolation mechanism — meaning one transient team-hook failure aborts the *entire* MCP tool-loading pass, taking down unrelated personal servers in the same run. | Scope the failure to team-owned servers only: on a re-keying failure, mark just that team-owned server unavailable (reusing the existing `_build_unavailable_mcp_config` pattern) instead of letting the exception propagate past the whole method; if a global fail-closed policy is genuinely intended, document that availability cost explicitly. |
| D2 | `src/xagent/web/tools/config.py:3560-3564` | **Medium** | Whether "the team has a row for this server" is checked via `if server_id in team_env_by_id` (membership), not truthiness. When the hook returns `{server_id: {}}` (a row that's an empty dict), this is treated as "has a row", `shared_env_by_id[server_id]` is set to `{}`, and the downstream `merge_stdio_env` skips merging an empty dict — silently pinning the runner to a global-only environment. The "no row" branch, by contrast, correctly degrades to the runner's own key. Two near-identical inputs produce opposite outcomes, and this is uncovered by any test (see T1). | Change to `if team_env_by_id.get(server_id):` so an empty dict is treated the same as "no row"; add a wiring test pinning this behavior. |
| D3 | `src/xagent/web/tools/config.py:3543` vs `mcp_runtime.py:92-98` | **Low** | `load_team_env_overrides` is missing the `db is None` guard its twin `load_shared_env_overrides` has. Confirmed unreachable on the one current call chain (an earlier `self.db.query(...)` already raises on `db is None`, caught and converted to `MCPConfigLoadError`), so this is a theoretical gap rather than a live bug — but the asymmetric defensiveness between two twin functions is easy to miss when a new call site is added later. | Add the `db is None` guard for symmetry; low cost. |

### ⚙️ Performance

No significant performance issues found. The PR description's claim of "one extra query per config-cache miss, not per task" matches the implementation (re-keying is gated on the existing `team_mcp_ids`-non-empty check, so a run with no governing team incurs no extra query). `mcp_env_hooks_snapshot` (see B1) isn't a hot path, but its thread-unsafe module-level global state could be a hidden source of races if it were ever used on a concurrent path.

### 📋 Best Practices

| # | File:Line | Severity | Issue | Suggestion |
|---|---|---|---|---|
| B1 | `src/xagent/web/services/mcp_runtime.py:225-248` (`mcp_env_hooks_snapshot`) | **Low** | Confirmed zero production callers repo-wide, and this PR's own two new test fixtures (`test_mcp_team_env_hook.py`, `test_mcp_team_env_wiring.py`) don't use it either — they reset via direct `set_*(None)` calls. It's only exercised by its own three self-tests. It's also unsynchronized module-level global state, not thread-safe. | Either move it into a test-only conftest helper, or actually have the two new fixtures use it; if kept, document the single-threaded assumption. |
| B2 | `src/xagent/web/tools/config.py:3543` vs `connector_team_scope.py:188` | **Low** | `load_team_env_overrides(db, team_id)` is called positionally, while its twin `team_connector_ids(db, *, team_id=...)` is keyword-only; `load_team_env_overrides`'s own signature is also *not* keyword-only, despite its docstring implying it should be "for the same reason". A swapped-order call with two `int \| None` args would silently type-check. | Make the signature `load_team_env_overrides(db, *, team_id: int \| None)`, matching the sibling convention; update the call site. |
| B3 | `src/xagent/web/tools/config.py:3538-3577` | **Low** | The ~12-line authorization re-keying core is inlined in a 113-line method and can't be unit-tested as a pure function; it depends on only four inputs (`shared_env_by_id`, `env_source_by_id`, `team_env_by_id`, `team_ids`), so extraction is cheap. | Extract to a pure function `apply_team_env_layer(shared_env_by_id, env_source_by_id, team_env_by_id, team_ids) -> tuple[dict, dict]`, unit-test it directly; the wiring test only needs to confirm it's called correctly. |
| B4 | `src/xagent/web/services/connector_team_scope.py:75-98` | **Low** | This module's docstring spends ~24 lines describing, in prose, the runtime behavior of a *different* module (`mcp_runtime`'s hook), with nothing tying the two together in tests or code — this documentation will silently rot if `mcp_runtime`'s fallback semantics change. | Trim to one line plus a cross-reference to `mcp_runtime.set_mcp_team_env_hook`; keep behavioral detail in one place. |
| B5 | `src/xagent/core/tools/adapters/vibe/factory.py:1130` (`ToolFactory.create_mcp_tools`/`_create_mcp_tools`) | **Low (informational)** | Confirmed this is a second, team-unaware copy of the user-keyed env resolution logic. Its sync wrapper `_create_mcp_tools` has zero production callers in `src/`, only exercised by `tests/core/tools/test_mcp_database_integration.py`; its own query is already filtered to the user's own links, so there's no active leak today. But it's a second copy of the same logic that would silently miss this fix if ever reused. | Comment that it's outside the team-governance path, or consider removing this legacy dead-code entry point. |
| T1 | `tests/web/tools/test_mcp_team_env_wiring.py` (whole file) | **Low** | Confirmed: no test case covers "governing team has a row + runner's `env_source` is `own`/`platform`", and no test covers a non-stdio (sse/streamable_http/oauth) team-owned server — every fixture hardcodes `transport="stdio"`. The current code appears safe in both gaps by construction (`resolve_stdio_env` never reads `shared_env` for `own`/`platform`; the non-stdio branch uses `build_mcp_runtime_connection`, which doesn't consume either table), but neither invariant is pinned by a test, so a future change could silently regress either one. | Add 1-2 parametrized tests pinning each of these two implicit invariants. |
| T2 | `tests/web/services/test_mcp_team_env_hook.py:340` | **Low** | The test name `test_load_team_env_overrides_default_is_empty_and_uncalled` claims to verify "uncalled", but the test body never installs a hook and has no `assert_not_called()`-style assertion — "uncalled" is a vacuous truth here, not something the test actually checks. | Rename to `..._default_is_empty`, or add a case that installs a hook and asserts on call count under the db/None path. |
| T3 | `tests/web/services/test_mcp_team_env_hook.py:440-483` | **Low** | Confirmed: the same 5-entry malformed-answer `parametrize` list (with ids) appears verbatim twice — once asserting the public `load_team_env_overrides` raises `ConnectorRuntimeError` (503), once asserting the private `_validate_team_mcp_env_answer` raises a raw `ValueError`. Each test pins a distinct layer's contract (not pure redundancy), but the input data itself is worth deduplicating. | Hoist the parametrize list to a module-level shared constant, reused by both tests. |

## 3. Simplification opportunities

- `mcp_runtime.py:183` — YAGNI: `mcp_env_hooks_snapshot` is a generic, naming-convention-based context manager built for a hypothetical third hook slot, exercised only by its own self-referential tests. Remove it; let the two test fixtures keep using the existing public setters directly.
- `test_mcp_team_env_hook.py:440` — Deduplicate the verbatim-repeated malformed-answer parametrize list (with ids) across the two test functions into a module-level shared constant `_MALFORMED_TEAM_ENV_ANSWERS`, reused by both.

Net: roughly -20 to -30 lines achievable (removing the snapshot helper itself plus deduplicating the parametrize data).

## 4. Positive observations

- **Clean, correctly-scoped data-flow tracing**: the governing-team re-keying logic is correctly confined to the two tables consumed on the `server.transport == "stdio"` branch; the oauth→stdio conversion path checks the *original* transport and is unaffected — an easy detail to get wrong that this PR gets right.
- **The failure strategy is a deliberate design choice, not an oversight**: `connector_team_scope.py`'s docstring explicitly documents the two-step (visibility hook vs. credential hook) install decision and explains why the new hook raises rather than degrades (authorization-sensitive vs. convenience). Even though this review thinks the availability/security trade-off deserves revisiting (see D1, S1), the decision itself is explicit and documented, not hidden behavior.
- **Substantial test suite with mostly meaningful assertions**: 892 lines of new tests cover 14 stated invariants; the wiring tests use a real SQLite DB and real ORM, stubbing only the hook itself — no evidence of over-mocking the logic under test.
- **Reuses existing patterns appropriately**: the new hook follows the same Protocol + keyword-only-argument pattern as the existing `TeamConnectorVisibilityHook`, without reinventing an abstraction or introducing an unnecessary new one.
- **Performance cost is contained and matches the stated claim**: the extra query is strictly gated behind the existing "has a governing team and has team-owned ids" check, matching the PR description's "per cache-miss, not per task" cost model.

## 5. Summary and priority

Recommended before merge, in priority order:
1. **D1 (High)**: a team-hook failure should not abort the entire MCP loading pass — scope it to team-owned servers only.
2. **D2 (Medium)**: fix the empty-dict vs. no-row semantic confusion — a real, untested behavior inversion.
3. **S1 / S2 (Medium)**: at minimum, log the "visibility-hook-only" deployment scenario; add inner env-value type validation.
4. The remaining Low findings (B1-B5, T1-T3, D3, S3) can be addressed as follow-ups and don't need to block merge.

src/xagent/web/tools/config.py
```diff
@@ -3522,6 +3528,51 @@ async def _load_mcp_server_configs(self) -> List[Dict[str, Any]]:
...
+                # may be the object a hook keeps its own reference to.
+                # Shallow copies of the outer dict are sufficient because
+                # every downstream consumer only reads the inner per-server
+                # values -- resolve_stdio_env's lookups, this file's own
+                # per-server env merge in _build_mcp_server_config's stdio
+                # branch, and mcp_runtime.build_mcp_runtime_connection's
+                # caller-id env merge (both of the latter two are
+                # unconditional-overwrite `{**a, **b}` merges) -- none of
+                # them assign back into shared_env_by_id or env_source_by_id.
+                shared_env_by_id = dict(shared_env_by_id)
```
rogercloud: The re-key lives only here; `ToolFactory.create_mcp_tools` (`factory.py:1164`) still resolves the shared layer user-keyed. Safe today only because its query joins personal `UserMCPServer` links (test-only path), but nothing structurally prevents cross-team env borrowing if that path ever sees team-owned servers — consider extracting the re-key (e.g. `rekey_team_env_layer` in `mcp_runtime.py`) or documenting the invariant at the loaders.

src/xagent/web/services/mcp_runtime.py
```diff
@@ -98,6 +104,139 @@ def load_shared_env_overrides(db: Any, user_id: int | None) -> dict[int, dict]:
...
+    (``bool`` is a subclass of ``int`` in Python but is rejected here -- a
+    truthy/falsy value is never a legitimate server id) and whose values
+    are all ``dict`` env mappings.
+    """
+    if not isinstance(answer, dict):
+        raise ValueError(
+            "team env hook returned a malformed answer: expected a dict, "
+            f"got {type(answer).__name__}"
+        )
+    for key, value in answer.items():
```
rogercloud, line 209: Only the outer shape is validated here; inner env keys/values pass anything (e.g. `{5: {"KEY": 1}}`), failing late and untyped at first tool invocation (pydantic `ValidationError` in the MCP SDK `StdioServerParameters`, or `TypeError` from subprocess env) instead of as the typed `ConnectorRuntimeError`/503 the loader docstring promises for malformed answers. The twin `_validate_team_connector_answer` validates deeper (`connector_team_scope.py:163-169`). Consider validating inner keys/values as `str` (or copy+validate) and extending the malformed-answer parametrizations.

src/xagent/web/services/mcp_runtime.py
```diff
@@ -98,6 +104,139 @@ def load_shared_env_overrides(db: Any, user_id: int | None) -> dict[int, dict]:
...
+    Unlike load_shared_env_overrides, this loader's failure policy is
+    raise, not degrade: any hook exception and any answer shape violation
+    becomes a typed ConnectorRuntimeError instead of silently falling back
+    to no shared layer. A misbehaving hook here must not be
+    indistinguishable from "the governing team legitimately has no row" --
+    the caller (config.py's wiring) already treats the latter as a
+    deliberate, signal-losing degrade of its own; collapsing hook failure
+    into the same outcome would double that loss silently.
+    """
+    if team_id is None or _get_mcp_team_env_hook is None:
```
rogercloud: `load_team_env_overrides` lacks the `db is None` guard all three sibling loaders have (mcp_runtime.py:31/59/94) and the twin's `team_id=int(team_id)` coercion (connector_team_scope.py:187). Unreachable today (the sole caller passes a live db and an int team_id), but mirroring the siblings is a one-line consistency cleanup.

src/xagent/web/tools/config.py
```diff
@@ -3522,6 +3528,51 @@ async def _load_mcp_server_configs(self) -> List[Dict[str, Any]]:
...
+                env_source_by_id = dict(env_source_by_id)
```
rogercloud: Redundant copy: `load_user_env_sources` always returns a fresh dict (mcp_runtime.py:72), so `env_source_by_id` can never be a hook-referenced object; the comment at 3546-3555 ("either may be the object a hook keeps its own reference to") is false for it. The `shared_env_by_id` copy at 3556 is the load-bearing one (the user-keyed hook's own dict is returned unwrapped, mcp_runtime.py:99). Consider dropping this line and narrowing the comment.

src/xagent/web/services/connector_team_scope.py
```diff
@@ -72,7 +72,14 @@ class TeamConnectorVisibilityHook(Protocol):
     MCP transports, a team-owned server with no personal link for the
     running user executes using the shared definition row's own stored
     credentials, not the runner's. (OAuth transports fail closed instead,
-    for lack of a token to resolve.) Custom APIs are starker: there is no
+    for lack of a token to resolve.) For stdio specifically, the shared env
```
rogercloud: The "no shared layer at all" rule stated here only holds when the team env hook (`set_mcp_team_env_hook`) is installed. With only the visibility hook installed (the state of every pre-PR deployment), team-owned servers keep the runner's user-keyed shared layer silently — the exact cross-team influence this PR forbids — and no warning is emitted (`test_team_env_hook_absent_falls_back_to_user_keyed_layer` deliberately pins the fallback). Consider adding the precondition to this docstring and/or emitting a WARNING when `team_mcp_ids` is non-empty but the env hook is absent.

rogercloud (approved):
# Review — PR #1345: "Add a team-keyed shared MCP env hook, separate from the user-keyed one"

*Review context: no prior substantive review comments (only a gemini-code-assist bot quota notice); no linked issues; tests were not run locally (CI is green).*

## 1. PR summary

For runs governed by a team-owned agent, the shared env layer for team-owned MCP stdio servers previously resolved from the runner's user-keyed shared-env hook (i.e., the runner's own team's row). The PR adds a separate team-keyed env hook (`TeamMCPEnvHook` / `set_mcp_team_env_hook` / `load_team_env_overrides` in `mcp_runtime.py`) that raises typed `ConnectorRuntimeError` 503s on failure (unlike the degrade-to-empty user-keyed loader), and a re-key block in `WebToolConfig._load_mcp_server_configs` (`config.py`) that, for team-owned ids, substitutes the governing team's own row into the shared slot, scrubs legacy values, and unconditionally degrades a "shared" pick to the legacy default (global env, then the runner's own key) when the governing team has no row — gated on `connector_team_id`, the installed predicate, and non-empty `team_mcp_ids` (one query per config-cache miss). 14 invariants across two new test files pin the loader and the wiring.

**Blocking: no — recommended event: APPROVE** (justification: all confirmed findings are minor or simplification-level; CI is green).

## 2. Approach verdict

Round 0 verdict is **sound**. The fix lands at the merge-decision point (`resolve_stdio_env`'s inputs) through the established app-hook seam, so the core never learns about team tables. The raise-vs-degrade policy split is deliberate and keyed on authorization relevance: a malformed hook answer is an authorization input and must fail loudly as a typed 503, while the user-keyed shared loader may keep degrading. The unconditional degrade keeps the outcome a function of exactly three inputs — the governing team's row, the runner's own key, and the pick — so the runner's own team's rows can never re-enter through a side door. The three-gate cost guard (connector_team_id, installed predicate, non-empty team_mcp_ids) keeps team resolution to one query per config-cache miss and zero for runs touching no team-owned id, and the degrade reuses `resolve_stdio_env`'s existing NULL semantics rather than adding a new branch. The six non-blocking design reservations from Round 0 were reviewed; the ones that survived verification appear below.

## 3. Findings by severity

All findings are **minor**; no critical or major issues were found.

### F1 (minor, design) — Team-env re-key lives only in `WebToolConfig._load_mcp_server_configs`

`src/xagent/web/tools/config.py:3532-3575`. The other shared-env consumer, `ToolFactory.create_mcp_tools` (`src/xagent/core/tools/adapters/vibe/factory.py:1164`), still resolves the shared layer purely user-keyed; safe today only because its query joins personal `UserMCPServer` links (verified: no production caller exists — test-only path), but nothing structurally prevents cross-team env borrowing if that path ever sees team-owned servers. **Suggest:** document the invariant at `load_shared_env_overrides` / `build_mcp_runtime_connection`, or extract the re-key into a shared helper (e.g. `rekey_team_env_layer` in `mcp_runtime.py`) so the mechanism module owns the policy.

### F2 (minor, docs/design) — Credential rule documented unconditionally

`src/xagent/web/services/connector_team_scope.py:75-82`: the new docstring states the rule unconditionally ("when the governing team has no stored row for that server, there is no shared layer at all"), but that only holds when `set_mcp_team_env_hook` is installed. With only the visibility hook installed (the state of every pre-PR deployment), team-owned servers keep the runner's user-keyed shared layer silently — the exact cross-team influence the PR forbids — and no warning is emitted. `test_team_env_hook_absent_falls_back_to_user_keyed_layer` (`tests/web/tools/test_mcp_team_env_wiring.py:429-451`) deliberately pins this fallback. **Suggest:** add the precondition to the docstring, and/or emit a WARNING when `team_mcp_ids` is non-empty but the env hook is absent.

### F3 (minor, correctness) — Validator checks only the outer shape

`src/xagent/web/services/mcp_runtime.py:148-181` (`_validate_team_mcp_env_answer`): only the outer shape is validated (dict, int non-bool keys, dict values). Inner env keys/values are not: `{5: {"KEY": 1}}` or `{5: {1: "x"}}` pass, then fail late and untyped at first tool invocation (pydantic `ValidationError` in the MCP SDK's `StdioServerParameters`, or `TypeError` from the subprocess env), not as the typed `ConnectorRuntimeError`/503 the loader docstring promises for malformed answers. The twin validator `_validate_team_connector_answer` validates deeper (`connector_team_scope.py:163-169`). **Suggest:** validate inner keys/values as `str` (or copy+validate), and extend the parametrized malformed-answer tests.

### F4 (minor, correctness edge) — Empty-dict team row suppresses the runner's own key

`src/xagent/web/tools/config.py:3559-3575`: an EMPTY-dict team row (`{server_id: {}}`) and a MISSING row behave differently for a runner with pick "shared" and an own key. Empty row: the row-present branch keeps the pick, `resolve_stdio_env` → `merge_stdio_env(global, {})` → global-only (`{}` is falsy, `mcp_runtime.py:269`) — the runner's own key is **suppressed**. Missing row: pop + del pick → legacy None → global < own-key — the own key **merges**. Storing an empty env row yields a strictly worse outcome than storing nothing, contradicts the module's own empty-means-absent convention (`load_user_env_overrides` skips empty, `mcp_runtime.py:48`), and is untested. **Suggest:** treat empty-dict team rows as absent (normalize in `load_team_env_overrides` or the wiring) and add a `{server_id: {}}` test.

### F5 (minor, API consistency) — Loader lacks the sibling `db is None` guard

`load_team_env_overrides` (`mcp_runtime.py:184-211`) guards only `team_id` None / hook absent; it lacks the `db is None` guard all three sibling loaders have (`mcp_runtime.py:31, 59, 94`) and the twin's `team_id=int(team_id)` coercion (`connector_team_scope.py:187`). Unreachable today (the sole caller passes a live db and an int team_id; the twin also lacks a db guard), so cosmetic — but mirroring the siblings is a one-line cheap consistency cleanup. Severity: minor (nit-level).

### F6 (minor, code quality) — Redundant copy

`src/xagent/web/tools/config.py:3557` `env_source_by_id = dict(env_source_by_id)` is redundant: `load_user_env_sources` always returns a fresh dict (`mcp_runtime.py:72`), so `env_source_by_id` can never be a hook-referenced object; the comment at `config.py:3546-3555` ("either may be the object a hook keeps its own reference to") is false for it. The `shared_env_by_id` copy at 3556 is load-bearing (the hook's own dict is returned unwrapped, `mcp_runtime.py:99`). **Suggest:** drop line 3557 and narrow the comment.

### F7 (minor, tests) — pick="shared" + team-row-present combination is never pinned

`tests/web/tools/test_mcp_team_env_wiring.py`: the pick="shared" + team-row-PRESENT combination is never pinned (only the no-row degrade branch is parametrized at ~line 361; row-present is only exercised with `env_source=None`); "own"/"platform" picks are never tested with a team-owned server — dropping the `== "shared"` guard at `config.py:3563` would silently flip "own" to the legacy fallback (leaking the team row into an "own" pick) and the suite would pass. **Suggest:** add shared+row-present and own/platform tests.

### F8 (minor, tests) — Cost claim unpinned

The "one extra query per config-cache miss" claim has no SQL query counting; the exactly-once pins (e.g. `calls == [T1]` at `test_mcp_team_env_wiring.py:315`) cannot catch an O(1)→O(per-server) regression because every test has exactly one team-owned server. **Suggest:** an invocation-count pin with 2+ team servers or a query-count assertion.

### F9 (minor, tests) — Duplicated parametrize blocks

`tests/web/services/test_mcp_team_env_hook.py`: the two parametrize blocks (`test_load_team_env_overrides_raises_on_malformed_answer` at 141-157, `test_validate_team_mcp_env_answer_raises_value_error` at 165-181) list the identical five malformed cases verbatim; a shared module-level constant would keep them in lockstep.

### F10 (minor, design/code quality) — Duplicated 503 raise-wrap

`mcp_runtime.py:199-211` vs `connector_team_scope.py:226-241`: near-verbatim duplication (same passthrough + WARNING log + `ConnectorRuntimeError(ERROR_CONNECTOR_RUNTIME_UNAVAILABLE, ..., status_code=503)` from exc; only message/reason differ). The visibility twin's own docstring (`connector_team_scope.py:206-214`) declares "this function is that wrap, written once" — its own standard. **Suggest:** extract a shared wrap helper (e.g. in `connector_runtime.py`, which both modules import). The slot+setter+predicate+validator pattern is now a justified third instance (different shapes/policies); state a consolidation threshold rather than forcing a refactor now.

### F11 (minor, design) — Shallow copy aliases inner env dicts

`mcp_runtime.py:181`: `return dict(answer)` aliases the app's inner env dicts into `shared_env_by_id` (`config.py:3560`). Verified: all downstream consumers are read-only today, and `test_hook_owned_mappings_not_mutated_by_team_rekey` (`test_mcp_team_env_wiring.py:577-590`) pins deep non-mutation. No live hazard, but the stated isolation intent is only partially delivered; `return {k: dict(v) for k, v in answer.items()}` is cheap hardening.

## 4. Simplification opportunities

(Confirmed simplification findings; no severity.)

- **L215** — yagni: `mcp_env_hooks_snapshot` (`mcp_runtime.py:214-237`) has no production or fixture callers — only its own three tests + the `_hook_slot_names` enumeration; every hook-installing test resets via autouse `_reset_hooks` fixtures. Delete the contextmanager, its 4 tests, and the now-unused `contextmanager`/`Iterator` imports. **net: -81 lines**
- **L3539** — delete the `self._connector_team_id is not None` conjunct: `resolve_team_connector_ids_or_raise` short-circuits to an empty "mcp" set when `team_id` is None (`connector_team_scope.py:177-188`), so `team_mcp_ids` is empty whenever the conjunct is false; the conjunct can never decide the guard. Drop it and the monkeypatch-dependent test that pins the impossible state (`test_connector_team_id_none_never_invokes_team_env_hook`, `tests/web/tools/test_mcp_team_env_wiring.py:227-265`). **net: -40 lines**

**net: -121 lines possible**

## 5. Blocking status & recommended decision

**Blocking: no** — all confirmed findings are minor or simplification-level; no critical/major findings; CI is green. Recommended event: **APPROVE**.

src/xagent/web/tools/config.py
AlexLiu190625: Confirmed, and left as is for this change. The other path resolves the shared
layer on user id alone, and its query only ever admits the user's own active
links, so no team-owned server reaches it; it also has no caller under `src/`.
Rather than move the policy now, the "Correct two comments about shared env copies and team governance" commit adds a note at that call site
recording what has to happen first if either of those changes — a caller in
`src/`, or a query that admits team-owned servers. Extracting a shared re-key
helper is worth doing on its own terms and is filed separately as #1364, since
it touches a file this change does not otherwise modify.

src/xagent/web/services/mcp_runtime.py
AlexLiu190625: Agreed, fixed in the "Validate the team env answer's inner types and drop empty rows" commit. The validator now requires every env key and value
to be `str`, in separate branches with their own messages, matching how the
server-id keys are already checked. No `bool` special case on the inner side:
that one exists on the key check only because `bool` is an `int` subclass and
would otherwise pass, while `bool` is not a `str` and is already rejected — the
docstring now says so, so it does not get "restored for symmetry" later. The
message never includes the offending env value itself, only its key and its
type, since these are credentials. The malformed-answer table gained non-string
key, non-string value and `None` value cases, and both the loader-level and
validator-level tests consume the same table.

src/xagent/web/services/mcp_runtime.py
AlexLiu190625: The `db is None` guard is added in the "Validate the team env answer's inner types and drop empty rows" commit, mirroring the three sibling
loaders. The `team_id=int(team_id)` coercion you mention is already present at
`mcp_runtime.py:209` — it went in with the second commit on this branch, which
is the one the approving review did not see; see the top-level comment.

src/xagent/web/tools/config.py
AlexLiu190625: Half right, and the important half is the comment. You are correct that
`load_user_env_sources` always returns a fresh dict, so no hook holds
`env_source_by_id` and the stated reason is false for it. The copy itself is
still load-bearing for a different reason: the degrade below removes entries
from that map in place (`del env_source_by_id[server_id]` at
`config.py:3577`), and the loader's result is not this block's to shrink.
the "Correct two comments about shared env copies and team governance" commit keeps both copies and rewrites the comment to give each one its
actual reason instead of one reason that only fits the other map.

src/xagent/web/services/connector_team_scope.py
AlexLiu190625: Both halves are addressed. The docstring precondition went in with the second
commit on this branch, which the approving review did
not cover — it now states that the rule holds only once the separate
credential-side hook is also installed, and points at that socket. The warning
is added in the "Report a team-owned connector scope with no team env hook installed" commit: once per process, when a governing team owns at least
one visible connector and no team-keyed env hook is installed. Once per process
rather than once per run, because the condition is an installation fact that
cannot change while the process is up. A deployment that uses none of the team
features cannot reach it at all — without the visibility hook no connector is
ever team-owned, so the scope check is already empty — which leaves exactly the
half-installed deployments as the audience.

AlexLiu190625:
Thanks for the thorough pass — the empty-row edge and the missing inner validation were both real, and the missing test combinations were the right ones to name.

**One thing to flag before the rest.** The approving review is anchored to the branch's first commit, one commit behind the branch head at the time. That head commit — "Scope the team-env credential rule to when the credential hook is installed" — has not been covered by any review: it tightens the credential rule's docstring to state the credential-hook precondition explicitly, splits the answer validator's malformed-key branches so the message reports the actual type, passes the team id through `int()` at the hook call, adds the team id to the failure log line, and drops a stale claim in the tool config about which consumers read the re-keyed maps. That is +41/-17 across three files. Could you take a look at that commit specifically? The earlier changes-requested review's test line references also point past the end of the test files, so its test-related findings should be read against the files as they actually stand.

**Already present in the branch's second commit, not repeated here:** the docstring precondition for the credential rule, and the `int(team_id)` coercion at the hook call.

**Changed in this round** (the three new commits on the branch):

| Area | What changed |
|---|---|
| Answer validator | Inner env keys and values must now be `str`, in separate branches with their own messages; the offending value is never formatted into the message |
| Loader | A team row with an empty env mapping is dropped, after shape validation rather than before, so it is indistinguishable from no row; `db is None` guard added, matching the sibling loaders |
| Tool config | Warns once per process when a governing team owns visible connectors but no team-keyed env hook is installed; the copy comment now gives each copy its real reason |
| Tests | Shared/own/platform picks against a team-owned server, a team-owned server on a non-stdio transport, empty row compared directly against no row, warning fired exactly once; the two duplicated malformed-answer tables folded into one constant both tests consume |
| Factory path | A note recording that it is outside team governance and what must happen first if that changes |

**Not changed, with reasons:**

- **The hook snapshot helper stays.** It now has real consumers: both reset fixtures restore through it instead of clearing each slot by hand. It is also part of the public surface of these hook sockets, so an application embedding this runtime can use it to isolate its own tests. Its slot-completeness constraint — a hook slot added to the module later, and not added to the snapshot, fails a coverage test — is the reason it exists, and that constraint survives only while the primitive does.
- **The `connector_team_id is not None` conjunct stays.** It is redundant today for the reason you give. Removing it also means removing the test that pins it, and that test is the only thing standing between this guard and a future change that makes the guard depend on an empty return value rather than on the installed predicate — a distinction this seam depends on elsewhere. Forty lines is a fair price for that.
- **The shallow copy stays shallow.** Isolating the outer mapping is the documented intent, not an incomplete attempt at deep isolation: every downstream consumer is read-only, and a test pins that the hook's inner dicts are observed unchanged after a build. Deep-copying every env dict on every config load would buy nothing that test does not already guarantee.

**Filed separately**, since each lands in a file this change does not otherwise touch:

- #1364 — bring the second shared-env resolution path under team governance before it gains a caller
- #1365 — the duplicated typed-503 wrap, and where the consolidation threshold should sit
- #1366 — the connector directory's "shared key available" hint is computed on a different basis than the runtime resolution
- #1367 — pin the per-cache-miss query cost with a multi-server invocation count

PR #1349, opened 2026-08-13
Resolve the knowledge-base team layer on the governing agent's team

AlexLiu190625:
### What this changes

A team-owned agent's knowledge bases should be the ones **its own team** owns, not
the ones whoever happens to be running it can see. Today `_list_visible_collections`
unions the runner's personal collections with the runner's *own* team memberships,
which is the right answer for a person browsing their knowledge bases and the wrong
answer for a run governed by someone else's team agent.

This PR builds the seam that makes the right answer possible, in three parts.

**The team layer resolves on the governing team.** `_list_visible_collections` now
takes a `governing_team_id`. When one is set *and* an application has installed a
team-keyed visibility hook, the team layer is the governing team's own knowledge
bases — never the runner's team memberships, and never a union of the two. Every
other input keeps the existing runner-keyed overlay, byte for byte.

Two ordering decisions inside that branch are deliberate:

- The team branch is reached **before** the `is_admin` short-circuit. An admin
  running a team agent must still resolve against that agent's team, not against
  their unrestricted view of every collection on the platform.
- The unauthenticated guard stays the **first** check. The calls below do
  `int(user_id)`, so removing or reordering it turns an unauthenticated caller into
  a `TypeError` instead of a well-defined early return.

Selection is on the hook predicate, never on an empty return value: an installed
hook legitimately answers "this team owns nothing".

**Resolution failure gets a typed channel.** `KnowledgeBaseScopeError` lives in a
module with zero xagent imports, so both sides of the core/web boundary can name it
without an import cycle. `resolve_team_knowledge_bases_or_raise` converts every
non-typed failure into it, and the run-path handlers that otherwise re-wrap
everything into a bare `RuntimeError` now let it through unchanged. Without this,
"the team owns nothing" (an empty list) and "resolution failed" collapse into the
same silent empty answer, and a search quietly returns fewer collections than it
should. The hook's answer is shape-validated before any caller sees it — `can_edit`
and `can_delete` default to `True` on `KnowledgeBaseAccess` and are copied straight
onto the collection handed back, so an answer that omits them would grant edit and
delete on the team's knowledge base to every runner who can search it.

**The governing agent's identity reaches the tool build.** Two more values travel
the existing route to the knowledge-base search entry point: the Agent row's own
creator, and the agent's *stored* `knowledge_bases` declaration. Both are derived at
the same three points that already derive the governing team, off the same frozen
snapshot or the same resolved agent, so a governing team can never be paired with
another agent's creator. The declaration is read from the agent's own configuration
and stays distinct from `tool_args.allowed_collections` and `tool_args.collections`
all the way down; those are model-authored on every path, and `setdefault` lets a
model-supplied value win outright over the agent's configured one.

### Nothing changes for anyone yet

Both new values arrive at `_search_knowledge_base_impl` and are not read there. The
declared-name resolution rule that consumes them arrives separately.

The team layer's own behaviour change is also inert until a deployment opts in:
standalone xagent installs no team-keyed hook, so `team_knowledge_base_hook_installed()`
is `False` and every path falls back to today's runner-keyed overlay. An application
that installs the hook gets the governing-team resolution; one that does not sees no
difference at all. The typed error can only be raised by a hook that exists.

That inertness is the reason this lands on its own: the seam is safe to ship ahead of
any hook installation, and reviewing it does not require holding the resolution rule
in your head at the same time.

### What is pinned

| Test file | What it holds |
|---|---|
| `tests/web/services/test_knowledge_base_team_scope.py` | The team-keyed hook slot, the answer validator (every rejection class), the typed-error conversion, and the snapshot/restore primitive |
| `tests/core/tools/core/test_document_search_team_layer.py` | The team layer resolving on the governing team for a member and for an admin; the runner-keyed fallback when no team hook is installed; the unauthenticated early return; the typed error surviving both re-wrap points; the tool build reading the governing values |
| `tests/core/tools/core/RAG_tools/kb/test_tool_compatibility.py` | All three values reaching the facade unmodified, on both the list and search entry points |
| `tests/core/tools/adapters/vibe/test_knowledge_tools_team_scope_wiring.py` | The knowledge tool builder reading them off the config and forwarding them |
| `tests/web/api/test_chat_team_scope_wiring_behavioral.py`, `..._static.py` | All three derivation points reading off one agent, never two |
| `tests/web/services/test_agent_runtime_fields_creator.py` | The snapshot field carrying the creator alongside the team id |
| `tests/web/tools/test_create_default_tools_team_scope_forwarding.py` | The tool-creation entry point forwarding both new values |
| `tests/core/tools/core/RAG_tools/test_multitenancy.py` | The collections endpoint staying runner-keyed even when a team hook is installed |
| `tests/core/tools/core/test_document_search_team_visibility_boundary.py`, `tests/web/test_team_sharing_hooks.py` | Existing suites moved onto the snapshot primitive so they no longer leak hook state |

`find_missing_knowledge_bases` deliberately stays runner-keyed and takes no governing
team: it backs the agent builder's save-time validation of an agent's own collection
selection, and an agent being edited is not yet the governing agent of any run.

### DIFF
--- a/src/xagent/core/tools/adapters/vibe/factory.py
+++ b/src/xagent/core/tools/adapters/vibe/factory.py
@@ -21,6 +21,7 @@ from sqlalchemy.orm import Session
 from .....config import get_uploads_dir
 from .....core.task_runtime import FILE_OPERATION_ACCESS_VERSION_KEY
 from .....core.workspace import TaskWorkspace
+from ...core.knowledge_base_scope import KnowledgeBaseScopeError
 from .base import AbstractBaseTool, Tool
 from .config import (
     BaseToolConfig,
@@ -222,10 +223,11 @@ class ToolRegistry:
         dispatched and are responsible for
         short-circuiting internally based on the spec.
 
-        Exception contract for creators: ``ConnectorRuntimeError`` and
-        ``RequiredMCPUnavailableError`` propagate to the caller; any other
-        exception is logged as a warning and that creator contributes no
-        tools. Creators rely on this and must not catch broadly themselves.
+        Exception contract for creators: ``ConnectorRuntimeError``,
+        ``RequiredMCPUnavailableError`` and ``KnowledgeBaseScopeError``
+        propagate to the caller; any other exception is logged as a warning
+        and that creator contributes no tools. Creators rely on this and must
+        not catch broadly themselves.
         """
         # Import tool modules on first call to trigger decorator registration
         cls._import_tool_modules()
@@ -248,6 +250,14 @@ class ToolRegistry:
             except ConnectorRuntimeError:
                 raise
             except RequiredMCPUnavailableError:
+                raise
+            except KnowledgeBaseScopeError:
+                # Defence in depth only (see knowledge_tools.py's own
+                # narrowing): the team hook is never invoked while tools are
+                # being built, so nothing today raises this here. Kept so a
+                # future change that does resolve the team layer during tool
+                # build cannot have the typed error silently swallowed by
+                # the blanket handler below.
                 raise
             except Exception as e:
                 logger.warning(
--- a/src/xagent/core/tools/adapters/vibe/knowledge_tools.py
+++ b/src/xagent/core/tools/adapters/vibe/knowledge_tools.py
@@ -3,6 +3,7 @@
 import logging
 from typing import TYPE_CHECKING, Any, List
 
+from ...core.knowledge_base_scope import KnowledgeBaseScopeError
 from .factory import register_tool
 
 if TYPE_CHECKING:
@@ -38,6 +39,18 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
         allowed_collections = config.get_allowed_collections()
         user_id = config.get_user_id()
         is_admin = config.is_admin()
+        # The governing agent's team, read the same way the six existing
+        # WebToolConfig-internal call sites already read it: as the private
+        # attribute, defaulting to None for every config that has no such
+        # attribute at all (every non-web config).
+        #
+        # Deliberately not behind a dedicated accessor. The attribute already
+        # has six direct readers inside WebToolConfig (config.py:1593, :1792,
+        # :2078, :3530, :3546, :3592), none of which an accessor added here
+        # would convert; a seventh reader asking the same question a seventh
+        # way is a second convention nothing else adopts, and the next reader
+        # then has to learn which of the two is authoritative.
+        governing_team_id = getattr(config, "_connector_team_id", None)
 
         if allowed_collections is not None and len(allowed_collections) == 0:
             return []
@@ -47,6 +60,7 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
                 allowed_collections=allowed_collections,
                 user_id=user_id,
                 is_admin=is_admin,
+                governing_team_id=governing_team_id,
             )
             tools.append(list_tool)
 
@@ -58,8 +72,17 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
             allowed_collections=allowed_collections,
             user_id=user_id,
             is_admin=is_admin,
+            governing_team_id=governing_team_id,
         )
         tools.append(knowledge_tool)
+    except KnowledgeBaseScopeError:
+        # Defence in depth only: the team hook is never invoked while tools
+        # are being built (resolution happens per search call, not here), so
+        # nothing today can raise this from inside this try block. Kept so a
+        # future change that does resolve the team layer during tool build
+        # cannot have its typed error silently swallowed by the blanket
+        # handler below.
+        raise
     except Exception as e:
         logger.warning(f"Failed to create knowledge tools: {e}")
 
--- a/src/xagent/core/tools/adapters/vibe/knowledge_tools.py
+++ b/src/xagent/core/tools/adapters/vibe/knowledge_tools.py
@@ -49,8 +49,12 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
         # :2078, :3530, :3546, :3592), none of which an accessor added here
         # would convert; a seventh reader asking the same question a seventh
         # way is a second convention nothing else adopts, and the next reader
-        # then has to learn which of the two is authoritative.
+        # then has to learn which of the two is authoritative. The two
+        # genuinely new values below get accessors because they have no
+        # existing readers to be inconsistent with.
         governing_team_id = getattr(config, "_connector_team_id", None)
+        agent_creator_user_id = config.get_agent_creator_user_id()
+        declared_knowledge_bases = config.get_declared_knowledge_bases()
 
         if allowed_collections is not None and len(allowed_collections) == 0:
             return []
@@ -61,6 +65,8 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
                 user_id=user_id,
                 is_admin=is_admin,
                 governing_team_id=governing_team_id,
+                agent_creator_user_id=agent_creator_user_id,
+                declared_knowledge_bases=declared_knowledge_bases,
             )
             tools.append(list_tool)
 
@@ -73,6 +79,8 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
             user_id=user_id,
             is_admin=is_admin,
             governing_team_id=governing_team_id,
+            agent_creator_user_id=agent_creator_user_id,
+            declared_knowledge_bases=declared_knowledge_bases,
         )
         tools.append(knowledge_tool)
     except KnowledgeBaseScopeError:
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -192,6 +192,8 @@ async def list_knowledge_bases(
     user_id: Optional[int] = None,
     is_admin: bool = False,
     governing_team_id: Optional[int] = None,
+    agent_creator_user_id: Optional[int] = None,
+    declared_knowledge_bases: Optional[List[str]] = None,
 ) -> ListKnowledgeBasesResult:
     """List all available knowledge bases through the tool compatibility facade."""
     return await _get_tool_compatibility_facade().list_knowledge_bases(
@@ -199,6 +201,8 @@ async def list_knowledge_bases(
         user_id=user_id,
         is_admin=is_admin,
         governing_team_id=governing_team_id,
+        agent_creator_user_id=agent_creator_user_id,
+        declared_knowledge_bases=declared_knowledge_bases,
     )
 
 
@@ -207,6 +211,8 @@ async def _list_knowledge_bases_impl(
     user_id: Optional[int] = None,
     is_admin: bool = False,
     governing_team_id: Optional[int] = None,
+    agent_creator_user_id: Optional[int] = None,
+    declared_knowledge_bases: Optional[List[str]] = None,
 ) -> ListKnowledgeBasesResult:
     """List all available knowledge bases with their statistics.
 
@@ -218,6 +224,19 @@ async def _list_knowledge_bases_impl(
             affects *which* collections are visible (the team layer resolves
             on this team instead of the runner's own team memberships); see
             ``_list_visible_collections``.
+        agent_creator_user_id: Accepted for signature symmetry with
+            ``_search_knowledge_base_impl`` and never read here.
+            ``knowledge_tools.py`` builds the list tool only when the
+            agent's allowed collections are ``None`` -- a non-empty list
+            builds the search tool alone, and an empty list builds no
+            knowledge tool at all -- and both chat call sites derive that
+            value and the declaration from the same stored field, so on this
+            path there is no declared name to classify at all. See
+            ``declared_knowledge_bases`` below.
+        declared_knowledge_bases: Accepted for signature symmetry and never
+            read here, for the same reason. This function reports "here is
+            what is available" -- an unresolved declared name simply does
+            not appear in the list, with no separate message field.
 
     Returns:
         ListKnowledgeBasesResult containing knowledge base information
@@ -302,6 +321,8 @@ async def search_knowledge_base(
     user_id: Optional[int] = None,
     is_admin: bool = False,
     governing_team_id: Optional[int] = None,
+    agent_creator_user_id: Optional[int] = None,
+    declared_knowledge_bases: Optional[List[str]] = None,
 ) -> KnowledgeSearchResult:
     """Search across knowledge bases through the tool compatibility facade."""
     return await _get_tool_compatibility_facade().search_knowledge_base(
@@ -309,6 +330,8 @@ async def search_knowledge_base(
         user_id=user_id,
         is_admin=is_admin,
         governing_team_id=governing_team_id,
+        agent_creator_user_id=agent_creator_user_id,
+        declared_knowledge_bases=declared_knowledge_bases,
     )
 
 
@@ -317,6 +340,8 @@ async def _search_knowledge_base_impl(
     user_id: Optional[int] = None,
     is_admin: bool = False,
     governing_team_id: Optional[int] = None,
+    agent_creator_user_id: Optional[int] = None,
+    declared_knowledge_bases: Optional[List[str]] = None,
 ) -> KnowledgeSearchResult:
     """Search across knowledge base collections.
 
@@ -327,6 +352,19 @@ async def _search_knowledge_base_impl(
         governing_team_id: The governing agent's owning team, if any. Only
             affects *which* collections are visible; see
             ``_list_visible_collections``.
+        agent_creator_user_id: The governing agent's creator, if any.
+            Carried to this function but not read by it: the rule that
+            tells the agent's own creator apart from every other runner of
+            a team-governed agent arrives separately.
+        declared_knowledge_bases: The governing agent's STORED
+            ``knowledge_bases`` declaration, if any -- never
+            ``tool_args.allowed_collections`` and never
+            ``tool_args.collections``, both of which are model-authored on
+            every path (a declared schema field the model can overwrite).
+            Carried here but not read yet, for the same reason as
+            ``agent_creator_user_id``; keeping the two apart all the way to
+            this function is what lets the rule land without re-threading
+            anything.
 
     Returns:
         KnowledgeSearchResult with formatted search results
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from pydantic import BaseModel, Field
 
 from ....config import get_kb_search_timeout_seconds
+from .knowledge_base_scope import KnowledgeBaseScopeError
 from .RAG_tools.core.schemas import ListCollectionsResult
 from .RAG_tools.management.collections import list_collections
 from .RAG_tools.pipelines.document_search import run_document_search
@@ -34,29 +35,68 @@ def _get_tool_compatibility_facade() -> "KBToolCompatibilityFacade":
 
 
 async def _list_visible_collections(
-    user_id: Optional[int], is_admin: bool
+    user_id: Optional[int],
+    is_admin: bool,
+    governing_team_id: Optional[int] = None,
 ) -> ListCollectionsResult:
-    """Union personal collections with application-provided team overlays."""
+    """Union personal collections with application-provided team overlays.
+
+    For a team-governed run (``governing_team_id`` set and a team-keyed hook
+    installed), the team layer is the *governing* team's own knowledge
+    bases, resolved through ``resolve_team_knowledge_bases_or_raise`` --
+    never the runner's own team memberships, and never a union of the two.
+    Every other case falls back to the existing runner-keyed overlay through
+    ``visible_team_knowledge_bases``, unchanged.
+
+    This function stays a pure union either way: it decides *whose* team
+    rows are visible, never which of the agent's declared names may search
+    them -- that rule belongs to the callers that read the agent's stored
+    configuration, not here, so a future consumer of this function cannot
+    inherit the declared-name rule by accident.
+    """
     result = await list_collections(user_id=user_id, is_admin=is_admin)
-    if user_id is None or is_admin:
+    # This is not an optimisation: the calls below do int(user_id), so
+    # removing this guard would turn an unauthenticated caller into a
+    # TypeError instead of this well-defined early return. It keeps its
+    # position as the first check for exactly that reason -- the governing
+    # team branch below still needs a real user_id.
+    if user_id is None:
         return result
 
     from ....web.services.db_runtime import run_db_io_cancellation_safe
     from ....web.services.knowledge_base_team_scope import (
         has_knowledge_base_visibility_hook,
+        resolve_team_knowledge_bases_or_raise,
+        team_knowledge_base_hook_installed,
         visible_team_knowledge_bases,
     )
 
-    if not has_knowledge_base_visibility_hook():
+    if governing_team_id is not None and team_knowledge_base_hook_installed():
+        # Selection is on the predicate above, never on an empty return: an
+        # installed hook legitimately answers "this team owns nothing".
+        #
+        # Reached before the is_admin short-circuit below on purpose: for a
+        # team-governed run, an admin's declared names must still resolve
+        # against the governing team, not against the admin's unrestricted
+        # view of every collection on the platform.
+        team_refs = await run_db_io_cancellation_safe(
+            lambda: resolve_team_knowledge_bases_or_raise(
+                None, team_id=governing_team_id, log_subject=user_id
+            )
+        )
+    elif is_admin:
         return result
+    elif not has_knowledge_base_visibility_hook():
+        return result
+    else:
+        team_refs = await run_db_io_cancellation_safe(
+            lambda: visible_team_knowledge_bases(None, int(user_id))
+        )
 
     collections_by_name = {
         collection.name: collection for collection in result.collections
     }
     refs_by_owner: dict[int, list] = {}
-    team_refs = await run_db_io_cancellation_safe(
-        lambda: visible_team_knowledge_bases(None, int(user_id))
-    )
     for ref in team_refs:
         refs_by_owner.setdefault(ref.storage_user_id, []).append(ref)
     for storage_user_id, refs in refs_by_owner.items():
@@ -68,6 +108,10 @@ async def _list_visible_collections(
             collection = owner_collections.get(ref.name)
             if collection is None:
                 continue
+            # The sole source of an ``ownership == "team"`` entry in the
+            # merged map is this loop -- every consumer's "does the
+            # governing team own this name" test (below, and in the two
+            # impls further down) is only trustworthy because of that.
             collections_by_name[ref.name] = collection.model_copy(
                 update={
                     "ownership": "team",
@@ -147,12 +191,14 @@ async def list_knowledge_bases(
     tool_args: ListKnowledgeBasesArgs,
     user_id: Optional[int] = None,
     is_admin: bool = False,
+    governing_team_id: Optional[int] = None,
 ) -> ListKnowledgeBasesResult:
     """List all available knowledge bases through the tool compatibility facade."""
     return await _get_tool_compatibility_facade().list_knowledge_bases(
         tool_args,
         user_id=user_id,
         is_admin=is_admin,
+        governing_team_id=governing_team_id,
     )
 
 
@@ -160,6 +206,7 @@ async def _list_knowledge_bases_impl(
     tool_args: ListKnowledgeBasesArgs,
     user_id: Optional[int] = None,
     is_admin: bool = False,
+    governing_team_id: Optional[int] = None,
 ) -> ListKnowledgeBasesResult:
     """List all available knowledge bases with their statistics.
 
@@ -167,15 +214,25 @@ async def _list_knowledge_bases_impl(
         tool_args: Args with optional allowed_collections filter
         user_id: Optional user ID for multi-tenancy filtering
         is_admin: Whether the user has admin privileges
+        governing_team_id: The governing agent's owning team, if any. Only
+            affects *which* collections are visible (the team layer resolves
+            on this team instead of the runner's own team memberships); see
+            ``_list_visible_collections``.
 
     Returns:
         ListKnowledgeBasesResult containing knowledge base information
 
     Raises:
-        RuntimeError: If listing knowledge bases fails
+        KnowledgeBaseScopeError: If the team-scope hook is installed but
+            fails or returns a malformed answer.
+        RuntimeError: If listing knowledge bases fails for any other reason.
     """
     try:
-        result = await _list_visible_collections(user_id=user_id, is_admin=is_admin)
+        result = await _list_visible_collections(
+            user_id=user_id,
+            is_admin=is_admin,
+            governing_team_id=governing_team_id,
+        )
 
         kb_list = []
         for collection in result.collections:
@@ -199,6 +256,8 @@ async def _list_knowledge_bases_impl(
 
         return ListKnowledgeBasesResult(knowledge_bases=kb_list)
 
+    except KnowledgeBaseScopeError:
+        raise
     except Exception as e:
         logger.error(f"Failed to list knowledge bases: {e}", exc_info=True)
         raise RuntimeError(f"Failed to list knowledge bases: {e}") from e
@@ -222,7 +281,13 @@ async def _find_missing_knowledge_bases_impl(
     user_id: Optional[int] = None,
     is_admin: bool = False,
 ) -> List[str]:
-    """Return requested knowledge base names that are not visible to the user."""
+    """Return requested knowledge base names that are not visible to the user.
+
+    Deliberately stays runner-keyed: this backs the agent-builder's
+    save-time validation of an agent's own collection selection, which must
+    answer against the runner's own visible set, not any governing team --
+    an agent being edited is not yet the governing agent of any run.
+    """
     requested = [name.strip() for name in knowledge_bases if name and name.strip()]
     if not requested:
         return []
@@ -236,12 +301,14 @@ async def search_knowledge_base(
     tool_args: KnowledgeSearchArgs,
     user_id: Optional[int] = None,
     is_admin: bool = False,
+    governing_team_id: Optional[int] = None,
 ) -> KnowledgeSearchResult:
     """Search across knowledge bases through the tool compatibility facade."""
     return await _get_tool_compatibility_facade().search_knowledge_base(
         tool_args,
         user_id=user_id,
         is_admin=is_admin,
+        governing_team_id=governing_team_id,
     )
 
 
@@ -249,6 +316,7 @@ async def _search_knowledge_base_impl(
     tool_args: KnowledgeSearchArgs,
     user_id: Optional[int] = None,
     is_admin: bool = False,
+    governing_team_id: Optional[int] = None,
 ) -> KnowledgeSearchResult:
     """Search across knowledge base collections.
 
@@ -256,17 +324,24 @@ async def _search_knowledge_base_impl(
         tool_args: Search configuration including query, collections, and search parameters
         user_id: Optional user ID for multi-tenancy filtering
         is_admin: Whether the user has admin privileges
+        governing_team_id: The governing agent's owning team, if any. Only
+            affects *which* collections are visible; see
+            ``_list_visible_collections``.
 
     Returns:
         KnowledgeSearchResult with formatted search results
 
     Raises:
-        RuntimeError: If search fails
+        KnowledgeBaseScopeError: If the team-scope hook is installed but
+            fails or returns a malformed answer.
+        RuntimeError: If search fails for any other reason.
     """
     try:
         # List all collections
         collections_result = await _list_visible_collections(
-            user_id=user_id, is_admin=is_admin
+            user_id=user_id,
+            is_admin=is_admin,
+            governing_team_id=governing_team_id,
         )
 
         if not collections_result.collections:
@@ -550,6 +625,8 @@ async def _search_knowledge_base_impl(
 
         return KnowledgeSearchResult(results=formatted_results, summary=summary)
 
+    except KnowledgeBaseScopeError:
+        raise
     except Exception as e:
         logger.error(f"Knowledge base search failed: {e}", exc_info=True)
         raise RuntimeError(f"Knowledge base search failed: {e}") from e
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -76,9 +76,11 @@ async def _list_visible_collections(
         # installed hook legitimately answers "this team owns nothing".
         #
         # Reached before the is_admin short-circuit below on purpose: for a
-        # team-governed run, an admin's declared names must still resolve
-        # against the governing team, not against the admin's unrestricted
-        # view of every collection on the platform.
+        # team-governed run, the team layer's source is the governing team,
+        # not the admin's own team memberships. This changes only where the
+        # team layer's rows come from -- the admin's platform-wide view in
+        # ``result`` is not narrowed by it; matching names are re-stamped
+        # with the team's ownership metadata below, the rest stay as-is.
         team_refs = await run_db_io_cancellation_safe(
             lambda: resolve_team_knowledge_bases_or_raise(
                 None, team_id=governing_team_id, log_subject=user_id
--- a/src/xagent/web/api/chat.py
+++ b/src/xagent/web/api/chat.py
@@ -596,6 +596,8 @@ async def create_default_tools(
     connector_team_id: Optional[int] = None,
     db_task_id: Optional[int] = None,
     file_operation_access_version: Any = None,
+    agent_creator_user_id: Optional[int] = None,
+    declared_knowledge_bases: Optional[List[str]] = None,
 ) -> tuple[list[Any], Any]:
     """Create default tools and tool_config for AgentService using ToolFactory.
 
@@ -614,6 +616,17 @@ async def create_default_tools(
     historical positional signature. Runtime callers pass them by keyword so
     File Operation receives trusted task and rollout authority without shifting
     existing positional integrations.
+
+    ``agent_creator_user_id`` and ``declared_knowledge_bases`` thread the same
+    governing agent's creator and its *stored* ``knowledge_bases`` list
+    alongside ``connector_team_id`` into ``WebToolConfig``, so the
+    knowledge-base search path can resolve a declared name against the
+    governing team, fall back to the creator's own collection, and report a
+    name the team never received rather than reaching for the runner's own
+    same-named collection. ``declared_knowledge_bases`` is read from the same
+    place ``allowed_collections`` already is (the agent's own configuration)
+    and is never the model-supplied value on a search request -- the two stay
+    distinct all the way to the resolution point.
     """
     if not user:
         raise ValueError("User is required for tool creation")
@@ -687,6 +700,8 @@ async def create_default_tools(
         mcp_load_summary_tracer=mcp_load_summary_tracer,
         mcp_load_summary_trace_task_id=mcp_load_summary_trace_task_id,
         connector_team_id=connector_team_id,
+        agent_creator_user_id=agent_creator_user_id,
+        declared_knowledge_bases=declared_knowledge_bases,
     )
 
     # Store excluded_agent_id in tool_config for agent tool filtering
@@ -1945,6 +1960,14 @@ class AgentServiceManager:
                 if task_setup_snapshot.agent is not None
                 else None
             )
+            # Same agent as connector_team_id above -- both are read off the
+            # same frozen AgentRuntimeFields snapshot, so they can never
+            # describe two different agents.
+            agent_creator_user_id = (
+                task_setup_snapshot.agent.agent_creator_user_id
+                if task_setup_snapshot.agent is not None
+                else None
+            )
         else:
             if db is None:
                 raise ValueError("Database session or task snapshot is required")
@@ -1966,6 +1989,12 @@ class AgentServiceManager:
                 int(current_agent.team_id)
                 if current_agent is not None and current_agent.team_id is not None
                 else None
+            )
+            # Captured alongside connector_team_id, for the same reason: both
+            # must describe the agent resolved above, never a preview agent
+            # substituted into ``current_agent`` by a branch further down.
+            agent_creator_user_id = (
+                int(current_agent.user_id) if current_agent is not None else None
             )
             if current_agent and _is_published_agent(current_agent):
                 excluded_agent_id = int(current_agent.id)
@@ -2058,6 +2087,10 @@ class AgentServiceManager:
             mcp_load_summary_tracer=parent_tracer,
             mcp_load_summary_trace_task_id=str(task_id),
             connector_team_id=connector_team_id,
+            agent_creator_user_id=agent_creator_user_id,
+            declared_knowledge_bases=agent_config["knowledge_bases"]
+            if agent_config
+            else None,
         )
 
     async def get_agent_for_task(
@@ -2706,6 +2739,21 @@ class AgentServiceManager:
                         if snapshot is not None and snapshot.agent is not None
                         else None
                     ),
+                    # Same agent as connector_team_id immediately above -- both
+                    # read off the same snapshot, so a governing team is never
+                    # paired with another agent's creator. This is the one of
+                    # the three derivation points that does not have its own
+                    # named local variable; it is folded straight into this
+                    # call because that is how connector_team_id already reads
+                    # here.
+                    agent_creator_user_id=(
+                        snapshot.agent.agent_creator_user_id
+                        if snapshot is not None and snapshot.agent is not None
+                        else None
+                    ),
+                    declared_knowledge_bases=agent_config["knowledge_bases"]
+                    if agent_config
+                    else None,
                 )
 
                 with UserContext(runtime_user_id):
--- a/src/xagent/web/services/knowledge_base_team_scope.py
+++ b/src/xagent/web/services/knowledge_base_team_scope.py
@@ -16,6 +16,12 @@ is actually asking. An application that wants to narrow who may run a
 team's agents at all enforces that at its agent-access policy layer, not at
 this seam -- this seam only answers "which knowledge base does this team
 own", never "is the current runner allowed to run this agent".
+
+A visibility hook's answer must set ``can_edit`` and ``can_delete`` to
+exactly ``False`` on every entry: ``KnowledgeBaseAccess`` defaults both to
+``True``, and an answer that leaves either at its default -- or omits it --
+is rejected outright rather than narrowed, which surfaces to every caller
+as the seam's one typed 503.
 """
 
 from __future__ import annotations
--- a/src/xagent/web/services/knowledge_base_team_scope.py
+++ b/src/xagent/web/services/knowledge_base_team_scope.py
@@ -4,17 +4,37 @@ Standalone xagent keeps collections scoped to the requesting user's id. A
 multi-tenant application can install these hooks to expose a logical team
 collection while its files and vector rows remain in the original storage
 user's namespace.
+
+Every knowledge base a team-keyed hook returns becomes searchable and
+quotable by *any* runner of that team's agents -- not only members, and not
+only interactive ones. A published team agent is meant to serve runners,
+including anonymous end users of a widget or other embedded surface, who are
+not members of the team that owns it. Once such a runner's turn resolves a
+declared collection onto the team's copy, the collection's document text
+becomes part of what that run can retrieve and quote back, regardless of who
+is actually asking. An application that wants to narrow who may run a
+team's agents at all enforces that at its agent-access policy layer, not at
+this seam -- this seam only answers "which knowledge base does this team
+own", never "is the current runner allowed to run this agent".
 """
 
 from __future__ import annotations
 
-from collections.abc import Callable
+import logging
+from collections.abc import Callable, Iterable, Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Literal
+from typing import Literal, Protocol
 
 from sqlalchemy.orm import Session
 
+from ...core.tools.core.knowledge_base_scope import KnowledgeBaseScopeError
+
+logger = logging.getLogger(__name__)
+
 KnowledgeBaseAction = Literal["read", "edit", "delete"]
+
+ERROR_KNOWLEDGE_BASE_SCOPE_UNAVAILABLE = "knowledge_base_scope_unavailable"
 
 
 @dataclass(frozen=True)
@@ -32,10 +52,43 @@ KnowledgeBaseAccessHook = Callable[
 ]
 KnowledgeBaseLifecycleHook = Callable[[Session | None, int, str, str | None], None]
 
+
+class TeamKnowledgeBaseVisibilityHook(Protocol):
+    """Resolver for the knowledge bases a team owns.
+
+    Typed as a ``Protocol`` with a keyword-only ``team_id`` so it cannot be
+    bound where ``KnowledgeBaseVisibilityHook`` (user-keyed) is expected, or
+    vice versa. The two shapes would otherwise be easy to confuse -- both
+    take an optional session and return a list of the same element type --
+    and the danger is not symmetric.
+
+    Binding the by-team function into the user-keyed slot is the dangerous
+    direction: ``visible_team_knowledge_bases`` calls that slot
+    *positionally* (``_visibility_hook(db, int(user_id))``), so a by-team
+    function bound there receives a plain ``user_id`` where it expects a
+    team id, and its answer is returned to the caller unvalidated -- it
+    silently resolves against whichever team happens to share that
+    numeric id, or misbehaves in some other way the type checker cannot
+    catch, because both callables type-check as functions from
+    ``(Session | None, int) -> list[KnowledgeBaseAccess]``.
+
+    The reverse swap -- binding the user-keyed function into this
+    protocol's slot -- is caught immediately: this slot is always called
+    with ``team_id=`` as a keyword, and a function whose second parameter
+    is positional-only or differently named raises ``TypeError`` at call
+    time instead of resolving against the wrong tenant.
+    """
+
+    def __call__(
+        self, db: Session | None, *, team_id: int
+    ) -> list[KnowledgeBaseAccess]: ...
+
+
 _visibility_hook: KnowledgeBaseVisibilityHook | None = None
 _access_hook: KnowledgeBaseAccessHook | None = None
 _renamed_hook: KnowledgeBaseLifecycleHook | None = None
 _deleted_hook: KnowledgeBaseLifecycleHook | None = None
+_team_visibility_hook: TeamKnowledgeBaseVisibilityHook | None = None
 
 
 def set_knowledge_base_team_hooks(
@@ -44,14 +97,26 @@ def set_knowledge_base_team_hooks(
     access: KnowledgeBaseAccessHook | None = None,
     renamed: KnowledgeBaseLifecycleHook | None = None,
     deleted: KnowledgeBaseLifecycleHook | None = None,
+    team_visibility: TeamKnowledgeBaseVisibilityHook | None = None,
 ) -> None:
-    """Install or clear application-owned knowledge-base hooks."""
+    """Install application-owned knowledge-base hooks.
+
+    Installs the complete hook set: every hook not supplied to a given call
+    is cleared, even one a previous call installed. This is a reset-all
+    setter, not a merge -- an application that installs hooks across more
+    than one call must pass its complete set each time. A caller that adds
+    a new slot to this function later must audit every existing installer
+    for the same reason: an installer that does not know about the new
+    keyword clears it.
+    """
 
     global _visibility_hook, _access_hook, _renamed_hook, _deleted_hook
+    global _team_visibility_hook
     _visibility_hook = visibility
     _access_hook = access
     _renamed_hook = renamed
     _deleted_hook = deleted
+    _team_visibility_hook = team_visibility
 
 
 def has_knowledge_base_visibility_hook() -> bool:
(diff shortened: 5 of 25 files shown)


rogercloud (changes requested):
## Summary

This PR makes the governing team the resolution authority for the knowledge-base names a team-owned agent declares: each declared name now resolves against the governing team's collections instead of the runner's own, the team's copy wins for every runner, the creator's personal copy is retained for the creator only, and any other outcome is reported rather than silently substituted — closing a cross-tenant document read whose contents reached the model and the answer. The change threads governing-team identity (team id, agent creator, stored declaration) from the web layer to the core resolution point, gates the whole rule on the installed team-scope hook so upstream can ship ahead of downstream, and ships a typed failure channel plus an extensive pin suite covering the resolution rule, the creator-existence probe, the failure messages, and the inert-until-installed behaviour. Two major issues remain — one on the delegation path, one in the terminal failure messages — and are detailed below.

The mechanism is the right fix at the right layer: resolution authority is re-keyed to the governing team at the actual resolution point, and the declared-name boundary is enforced by iterating the agent's stored declaration — closing the cross-tenant read at its root rather than patching per call site. It is not over-engineered: it mirrors the proven connector team-scope seam (installed-predicate gating, reset-all setter, resolve-or-raise wrapper, typed error with `safe_message`) almost line for line. The three pinned security decisions are genuinely implemented, verified against code rather than tests alone: the creator-existence probe is bound to the stored declaration (never a model-writable field, with a discriminating test), the external/anonymous-runner surface resolves as the creator (grounded in the a2a wiring, not just a pin test), and disclosure is bounded to declared names with no identity-bearing fields in messages. Reservations: the spec's rule case 4 (empty declaration) is unreachable through the production call sites (m2), probe failures collapse into a false "does not exist" verdict (m4), KB governance identity rides a cross-module private attribute (m1), and the delegation path bypasses the rule entirely (M1).

## Major findings

### M1 — Delegation path bypasses the declared-name rule; silent substitution remains reachable (major)
`src/xagent/core/tools/adapters/vibe/agent_tool.py:2265` (Phase 1 capture at 2158-2164)

`ExecuteAgentTool` builds the delegated sub-agent's `WebToolConfig` with `connector_team_id=None` (connector-only rationale at agent_tool.py:2260-2264) and never passes `agent_creator_user_id` or `declared_knowledge_bases`; the Phase 1 capture (2158-2164) snapshots name/instructions/knowledge_bases/skills/tool_categories/models/execution_mode but not team_id/user_id. Since knowledge_tools.py:49 reads `governing_team_id` from that same private attribute, on the delegation path `governing_team_id=None`, the declared-name gate (document_search.py:588-592) is off, and `_resolve_declared_name` falls to "today's behaviour" (document_search.py:387-389). Consequences: (a) a delegated team sub-agent's declared KBs fail with "does not exist" / "None of the allowed collections exist" for a non-member runner (e.g. an anonymous widget user of a published team agent), contradicting the PR body's "a published team agent must keep working for the end users it was published for"; (b) declared names silently resolve to the runner's own same-named personal collections — exactly the cross-tenant silent substitution this PR exists to eliminate. The delegation path does build KB tools for knowledge-category sub-agents (the factory dispatches `create_knowledge_tools`), so this is a real reachable path, and it escapes the PR's static guard, which counts `create_default_tools` call sites only; agent_tool.py is untouched by the PR and no test covers delegation + team-governed KBs.

Suggestion: capture `agent.team_id` and `agent.user_id` in Phase 1 and thread them (plus `declared_knowledge_bases`) into the delegated `WebToolConfig` via dedicated keyword args, decoupling KB governance identity from `connector_team_id` (see m1); add a delegation-scenario test; and fix the web/tools/config.py:1220-1231 docstring claim that "no request path ever supplies this directly", which this caller contradicts.

### M2 — Terminal failure messages list the runner's same-named personal copies as Available/Allowed (major)
`src/xagent/core/tools/core/document_search.py:685` (also 609-611, 675-676)

In the team-declared branch the terminal messages compute availability as `authorised & available_names`, where `available_names` is the pre-verdict visible union (runner personal + team overlay). In the total-failure branch every declared name failed to resolve by definition, so the only declared names in that intersection are the runner's own same-named personal copies — the exact collections the rule excludes from search. The rendered result is self-contradictory: "Error: None of the allowed collections exist. Allowed: handbook, policies. Available: handbook, policies." — and it tells the model/operator the runner's personal copies are available when they are precisely not searchable. The not-allowed early return (609-611) has the same mismatch, listing a declared name as Allowed even though searching it would fail. The PR's own test `test_total_kb_failure_with_same_named_personal_copies_reports_generic_error` pins the generic wording and asserts `searched==[]` but never asserts the Available list contents; legacy had no such contradiction (the runner's copy WAS searchable there), but the semantics changed under the new rule while the list expression was copied verbatim.

Suggestion: build the Available/Allowed lists from the verdict-admitted names (team-owned and creator-resolved collections), not from `available_names`; in the total-failure branch that set is empty, so emit the clause as empty or omit it. Extend the same-named-copies test to assert the runner's personal copy names never appear in the Available/Allowed clauses.

## Minor findings

- **m1 (minor) — KB governing-team identity reuses the `_connector_team_id` private attribute**: knowledge_tools.py:49 reads `governing_team_id` via `getattr(config, '_connector_team_id', None)` — the only value read cross-module as a private attribute, while the two sibling values use proper `BaseToolConfig` accessors with fail-closed defaults; the in-code comment's cited rationale ("see `WebToolConfig._connector_team_id` for why a new one was not added") does not exist in web/tools/config.py. A future rename of that attribute fails loudly at the six in-class readers but silently returns None here, silently disabling the rule. Suggest a dedicated `get_governing_team_id()` accessor, keeping `connector_team_id` connector-only.
- **m2 (minor) — PR rule case 4 is unreachable: empty declaration builds no KB tools**: knowledge_tools.py:53-54 returns `[]` when `allowed_collections == []` (pre-existing), and both chat call sites derive `declared_knowledge_bases` from the same `agent_config['knowledge_bases']` expression, so a team-owned agent declaring nothing builds no KB tools. The PR body's rule case 4 ("the agent declares nothing — the runner searches their own material ... with only the team layer re-keyed") and the fix-commit's "specced to still search the governing team's knowledge bases" describe a state production cannot reach (the fallthrough requires `declared_knowledge_bases=None`, which implies no governing team). Restriction-only, but either implement the case or correct the PR body, the commit claim, and the code comments; the direct-impl test `test_team_governed_run_without_agent_config_uses_governing_team_layer` bypasses the tool-build gate and gives false coverage confidence.
- **m3 (minor) — Empty-visible-set early return bypasses verdict reporting**: document_search.py:520-523 returns "No knowledge bases available..." before the declared-name gate. Team-governed run + non-creator runner with zero personal collections + team owning nothing → every declared name fails with the generic message, no sharing-gap/does-not-exist sentence, creator probe never runs. Fail-closed, but rule 3's "the response says which case it is" is broken in exactly this edge. For team-governed runs with a non-empty declaration, run the verdict loop before (or instead of) the early return.
- **m4 (minor) — Creator-probe failure renders a false-certainty "does not exist" verdict**: document_search.py:467 — `holds()` swallows every exception and caches an empty set, so a probe failure classifies a creator-held name as missing, asserting certainty about absence when the lookup merely failed. The partial-failure path already has a hedged, security-equivalent wording ("could not be resolved for this agent"). Apply the same hedge to the terminal verdicts while keeping the binary classification that closes the enumeration oracle.
- **m5 (minor) — Non-string/unhashable stored declaration entries collapse the whole search into RuntimeError**: document_search.py:543 — `list(dict.fromkeys(declared_knowledge_bases or []))` raises TypeError on unhashable entries before the team gate, and `', '.join(...)`/`sorted()` raise on non-str entries; one bad entry breaks every search for that agent under any governance. `Agent.knowledge_bases` is an untyped JSON column reachable via hand-edits, legacy data, raw setattr in agent_store, or client-supplied agent_config. Normalize the declaration once at the read site (coerce to str, drop non-str with a warning), mirroring the `strip()` the save path applies.
- **m6 (minor) — Reset-all hook setter silently clears the team hook**: knowledge_base_team_scope.py:119 — `set_knowledge_base_team_hooks` clears every slot not supplied, so a downstream installer written before this PR (visibility/access/renamed/deleted only) silently sets `_team_visibility_hook=None` on every call, disabling the security rule with no error — silent reversion to pre-PR runner-keyed behavior. Give `team_visibility` preserve-if-omitted semantics (or a dedicated setter) and audit downstream installers.
- **m7 (minor) — Coverage gaps: no empty-union verdict test; construction-site guard has no documented exemption**: test_document_search_team_declared_names.py:1361 — (a) no test combines runner-with-zero-collections + team-owns-nothing + non-empty declaration, the exact m3 edge; (b) the exactly-two-construction-sites test treats websocket.py:8285 as exempt without a comment documenting why, so a refactor routing the builder through knowledge_tools.py would break the count without a semantic change. Add the empty-union verdict test and document the exemption.
- **m8 (minor) — factory.py exception-contract docstring not updated for KnowledgeBaseScopeError**: factory.py:226-229 — the contract comment enumerates exactly two propagatable types (ConnectorRuntimeError, RequiredMCPUnavailableError) and states any other exception is logged, but the new `except KnowledgeBaseScopeError: raise` at 253-260 makes that statement false. Add KnowledgeBaseScopeError to the propagatable list.
- **m9 (minor) — Test-only snapshot primitive lives in the production seam module**: knowledge_base_team_scope.py:285 — `snapshot_knowledge_base_team_hooks` exists solely for test isolation (all usages in tests/), and the mirrored connector seam has no counterpart (its tests use a test-side reset fixture). Move it to a test-side fixture or document the deliberate asymmetry.
- **m10 (minor) — Terminal summary templates duplicated between team and legacy branches**: document_search.py:683 (with 609-611/729-731, 675-676/741-742, 683-685/763-765) — the three terminal texts are hand-rolled twice with subtly different set expressions (`authorised & available_names` vs `available_names`), and the divergence risk is exactly what produced M2; the PR itself added shared renderers for the two new messages. Hoist the three templates into shared module-level renderers.
- **m11 (minor) — Two doc-comments contradict the code**: llm_utils.py:1082 claims `agent_creator_user_id` is None "for every personal/legacy agent", but the only construction site (llm_utils.py:1314) populates it unconditionally and the PR's own tests pin a personal agent carrying its creator id; the `_list_knowledge_bases_impl` docstring (document_search.py:239-242) is likewise wrong on both ends (an empty declaration builds nothing, and the list tool is built when `allowed_collections` is None). Reword both to match actual behavior.

- **L227 (src/xagent/core/tools/core/document_search.py:227): delete** — `agent_creator_user_id`/`declared_knowledge_bases` on the list path (`_list_knowledge_bases_impl`, `list_knowledge_bases`, `KBToolCompatibilityFacade.list_knowledge_bases`, `ListKnowledgeBasesTool`) are accepted and threaded but never read; always None in production when the list tool is built. The "signature symmetry" rationale is aesthetic, not technical — no unified dispatch requires identical signatures. Drop them from the list-path signatures; the search path keeps them (3 wiring tests pin the dead threading and would need updating).

  net: -25 lines (7 signatures x 2 params + 2 attribute stores + ~14-line docstring rationale ≈ -25 lines)

Blocking: yes

Blocking issues:
- src/xagent/core/tools/adapters/vibe/agent_tool.py:2265 — major — delegation path bypasses the declared-name rule; silent cross-tenant substitution remains reachable [new]
- src/xagent/core/tools/core/document_search.py:685 — major — terminal failure messages list the runner's same-named personal copies as Available/Allowed, contradicting the rule they enforce [new]

Recommended event: REQUEST_CHANGES

src/xagent/core/tools/adapters/vibe/knowledge_tools.py
```diff
@@ -38,6 +39,16 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
...
         user_id = config.get_user_id()
         is_admin = config.is_admin()
+        # The governing agent's team, read the same way the six existing
+        # WebToolConfig-internal call sites already read it: as the private
+        # attribute, defaulting to None for every config that has no such
+        # attribute at all (every non-web config). This intentionally does
+        # not go through a dedicated accessor -- see
+        # ``WebToolConfig._connector_team_id`` for why a new one was not
+        # added. The two genuinely new values below do get accessors.
+        governing_team_id = getattr(config, "_connector_team_id", None)
```
rogercloud, line 55: This is the only value read via a cross-module private attribute (`getattr(config, '_connector_team_id', None)`) while the two sibling values use proper `BaseToolConfig` accessors with fail-closed defaults — and it makes the KB scope rule hostage to connector scope. On the delegation path `ExecuteAgentTool` (agent_tool.py:2225-2266) deliberately sets `connector_team_id=None` for connector safety and never passes `agent_creator_user_id`/`declared_knowledge_bases`, so `governing_team_id` is None here, the declared-name gate (document_search.py:588-592) is off, and a delegated team sub-agent's declared KBs either fail with "does not exist" for a non-member runner (contradicting the PR body's "a published team agent must keep working for the end users it was published for") or silently resolve to the runner's own same-named personal collections — the exact cross-tenant silent substitution this PR exists to eliminate. Delegation does build KB tools for knowledge-category sub-agents, so this is a reachable path; agent_tool.py Phase 1 (2158-2164) already captures the agent row but omits team_id/user_id, and no test covers delegation + team-governed KBs. Suggested fix: add a dedicated `get_governing_team_id()` accessor to `BaseToolConfig` (keeping `connector_team_id` connector-only), capture `agent.team_id`/`agent.user_id` in Phase 1, thread all three values into the delegated `WebToolConfig`, and add a delegation-scenario test. The cited rationale comment ("see `WebToolConfig._connector_team_id` for why a new one was not added") has no corresponding text in web/tools/config.py. Also note web/tools/config.py:1220-1231 claims "no request path ever supplies this directly" — this caller contradicts that.

src/xagent/core/tools/adapters/vibe/knowledge_tools.py
```diff
@@ -38,6 +39,16 @@ async def _create_knowledge_tools_impl(config: "BaseToolConfig") -> List[Any]:
...
+        agent_creator_user_id = config.get_agent_creator_user_id()
+        declared_knowledge_bases = config.get_declared_knowledge_bases()
 
         if allowed_collections is not None and len(allowed_collections) == 0:
```
rogercloud, line 59: This pre-existing guard makes the PR body's rule case 4 unreachable: an empty declaration (`allowed_collections == []`) returns no KB tools at all, and both chat call sites derive `allowed_collections` and `declared_knowledge_bases` from the identical `agent_config['knowledge_bases']` expression (normalized to a list by load_agent_builder_config), so a team-owned agent declaring nothing never reaches document_search's empty-declaration fallthrough. The fallthrough requires `declared_knowledge_bases=None`, which implies `agent_config=None`, which implies no governing team (llm_utils.py sets team_id and agent_config together). The PR body's case 4, the fix-commit's "an empty declaration ... still search[es] the governing team's knowledge bases", and the comment at document_search.py:557-576 all describe a production-impossible state; test_team_governed_run_without_agent_config_uses_governing_team_layer reaches it only by calling the impl directly, bypassing this gate. Restriction-only (no disclosure risk), but either implement case 4 or correct the spec/comment/test to state that an empty declaration disables KB tools.

src/xagent/core/tools/core/document_search.py
```diff
@@ -287,11 +535,190 @@ async def _search_knowledge_base_impl(
...
+                        f"Available collections: {', '.join(sorted(authorised & available_names))}"
+                    )
+                    failed_unshared_names = [
+                        name for name in unshared_names if name in explicit_request
+                    ]
+                else:
+                    summary = (
+                        f"Error: None of the allowed collections exist. "
+                        f"Allowed: {', '.join(sorted(authorised))}. "
+                        f"Available: {', '.join(sorted(authorised & available_names))}"
```
rogercloud: In the total-failure branch "Available" is computed as `authorised & available_names`, where `available_names` (line 527) is the pre-verdict visible union (runner personal + team overlay). Every declared name failed to resolve by definition, so the only declared names in that intersection are the runner's own same-named personal copies — the exact collections the rule excludes from search. The rendered message is self-contradictory: "Error: None of the allowed collections exist. Allowed: handbook, policies. Available: handbook, policies." — and it tells the model/operator the runner's personal copies are available when they are precisely not searchable. The not-allowed early return (609-611) has the same mismatch (lists a declared name as Allowed even though searching it would fail). The PR's own test test_total_kb_failure_with_same_named_personal_copies_reports_generic_error pins the generic wording and asserts searched==[] but never asserts the Available list. Suggest building the Available/Allowed lists from the verdict-admitted names (team-owned and creator-resolved); in the total-failure branch that set is empty, so emit the clause as empty or omit it, and extend the test to assert the runner's personal copy names never appear.

src/xagent/core/tools/core/document_search.py
```diff
@@ -236,37 +332,189 @@ async def search_knowledge_base(
...
     try:
         # List all collections
         collections_result = await _list_visible_collections(
-            user_id=user_id, is_admin=is_admin
+            user_id=user_id,
+            is_admin=is_admin,
+            governing_team_id=governing_team_id,
         )
 
         if not collections_result.collections:
```
rogercloud, line 387: This empty-visible-set early return fires before the declared-name gate below: a team-governed run with a non-creator runner holding zero personal collections and a governing team owning nothing (an installed hook legitimately answers empty) returns "No knowledge bases available..." and skips the verdict loop entirely — no sharing-gap sentence, no does-not-exist sentence, creator probe never runs. Fail-closed (nothing substituted), but PR rule 3 ("the response says which case it is") is broken in exactly this edge, the outcome is indistinguishable from a genuinely empty platform, and the PR body's "gated on the installed predicate, never on an empty answer" claim does not hold for this inherited guard. Suggest running the verdict loop for team-governed runs with a non-empty declaration before (or instead of) the early return, keeping the early return for legacy inputs. No test currently combines runner-with-zero-collections + team-owns-nothing + non-empty declaration.

src/xagent/core/tools/core/document_search.py
```diff
@@ -236,37 +332,189 @@ async def search_knowledge_base(
...
+                    user_id=self._agent_creator_user_id, is_admin=False
+                )
+                self._names = {c.name for c in creator_result.collections}
+            except Exception:
+                logger.warning(
+                    "Failed to resolve creator's personal collections while "
+                    "classifying an unshared knowledge base",
+                    exc_info=True,
+                )
+                self._names = set()
```
rogercloud: `holds()` swallows every exception and caches an empty set, so a creator-collection lookup failure (e.g. metadata store down) classifies a name the creator actually holds as "does not exist" — asserting certainty about absence when the lookup merely failed. The degradation is deliberate and documented, but the partial-failure path already carries a hedged, security-equivalent wording ("could not be resolved for this agent", _PARTIAL_FAILURE_NOTE) that is accurate in both outcomes; the terminal path lacks it. Applying the same hedge to the terminal missing verdicts preserves the disclosure bound (the two failure classes stay indistinguishable, keeping the enumeration oracle closed) without stating a falsehood.

src/xagent/core/tools/core/document_search.py
```diff
@@ -287,11 +535,190 @@ async def _search_knowledge_base_impl(
...
             logger.info(f"   - Allowed collections: {tool_args.allowed_collections}")
 
-        if tool_args.collections:
-            # User specified collections - validate against allowed_collections
+        # The agent's stored declaration, de-duplicated with insertion order
+        # kept (a stored declaration may legitimately contain a name twice --
+        # nothing normalises Agent.knowledge_bases into a set -- and an
+        # un-de-duplicated iteration would probe, and report, a duplicate
+        # name twice).
+        declared_names_for_this_run = list(
```
rogercloud: `list(dict.fromkeys(declared_knowledge_bases or []))` runs unconditionally before the team gate and raises TypeError on unhashable entries, and `', '.join(...)`/`sorted(...)` raise on non-str entries — one bad entry breaks every search for that agent under any governance, surfacing through the blanket handler as "Knowledge base search failed". `Agent.knowledge_bases` is an untyped JSON column; hand-edits, legacy data, raw setattr in agent_store.py:461-472, or client-supplied agent_config (ChatRequest.agent_config is free-form) can carry such values, and save-time validation (`.strip()`) only covers the primary API path. Suggest normalizing the declaration once at the read site (coerce to str, drop non-str with a warning), mirroring the strip() the save path applies.

src/xagent/web/services/knowledge_base_team_scope.py
```diff
@@ -44,14 +97,26 @@ def set_knowledge_base_team_hooks(
...
+    keyword clears it.
+    """
 
     global _visibility_hook, _access_hook, _renamed_hook, _deleted_hook
+    global _team_visibility_hook
     _visibility_hook = visibility
     _access_hook = access
     _renamed_hook = renamed
     _deleted_hook = deleted
+    _team_visibility_hook = team_visibility
```
rogercloud, line 125: The reset-all setter clears every slot not supplied, so a downstream installer written before this PR (visibility/access/renamed/deleted only) silently sets `_team_visibility_hook=None` on every call — including an ordering where a stale call runs after installation — disabling the team layer and the declared-name rule with no error. Failure mode is silent reversion to pre-PR runner-keyed resolution. The docstring warns, but there is no runtime guard. Consider preserve-if-omitted semantics for `team_visibility`, or a dedicated setter, and audit downstream installers.

tests/core/tools/core/test_document_search_team_declared_names.py
```diff
@@ -0,0 +1,1379 @@
...
+
+        def get_declared_knowledge_bases(self):
+            return allowed_collections
+
+    tools = await knowledge_tools._create_knowledge_tools_impl(_Config())
+    tool_names = {t.name for t in tools}
+    assert ("list_knowledge_bases" in tool_names) is expect_list_tool
+
+
+def test_list_knowledge_bases_tool_has_exactly_two_construction_sites() -> None:
```
rogercloud: Two coverage gaps: (a) no test combines runner-with-zero-personal-collections + team-owns-nothing + non-empty declaration — the exact edge where the empty-visible-set early return (document_search.py:520) bypasses verdict reporting; several tests run the rule with an empty runner but a non-empty team, which cannot reach that return. (b) This construction-sites guard counts websocket.py:8285 (builder surface, passes no team params, exempt by design) without a comment documenting why that site is exempt, so a refactor routing the builder through knowledge_tools.py would break the count without a semantic change. Add the empty-union verdict test (asserting the verdict message, not the generic one) and document the websocket exemption.

src/xagent/core/tools/adapters/vibe/factory.py
```diff
@@ -249,6 +250,14 @@ async def create_registered_tools(cls, config: BaseToolConfig) -> list[Tool]:
                 raise
             except RequiredMCPUnavailableError:
                 raise
+            except KnowledgeBaseScopeError:
```
rogercloud, line 254: This new re-raise makes the creator exception-contract docstring above (lines 226-229) stale: it enumerates exactly two propagatable types (ConnectorRuntimeError, RequiredMCPUnavailableError) and states any other exception is logged as a warning and contributes no tools, but KnowledgeBaseScopeError now propagates instead. The defence-in-depth rationale covers the clause, not the contract text — add KnowledgeBaseScopeError to the propagatable list.

src/xagent/web/services/knowledge_base_team_scope.py
```diff
@@ -74,6 +139,181 @@ def visible_team_knowledge_bases(
...
+        )
+        raise KnowledgeBaseScopeError(
+            ERROR_KNOWLEDGE_BASE_SCOPE_UNAVAILABLE,
+            "Knowledge base team scope is unavailable.",
+            details={"reason": "team_scope_resolution_failed"},
+            status_code=503,
+        ) from exc
+
+
+@contextmanager
```
rogercloud, line 291: snapshot_knowledge_base_team_hooks exists solely for test isolation (all usages are in tests/), and the mirrored connector seam has no counterpart — its tests isolate via a test-side reset fixture calling the reset-all setter, a proven alternative. Inert in production, but the production module's public surface gains a test-only name asymmetric with the seam it mirrors. Consider moving it to a test-side fixture (as connector tests do) or documenting the deliberate asymmetry.

src/xagent/core/tools/core/document_search.py
```diff
@@ -287,11 +535,190 @@ async def _search_knowledge_base_impl(
...
+                    summary = (
+                        f"Error: The following collections do not exist: {', '.join(sorted(explicit_request))}. "
+                        f"Available collections: {', '.join(sorted(authorised & available_names))}"
+                    )
+                    failed_unshared_names = [
+                        name for name in unshared_names if name in explicit_request
+                    ]
+                else:
+                    summary = (
+                        f"Error: None of the allowed collections exist. "
```
rogercloud: The three terminal texts are hand-rolled twice — here vs 729-731/741-742/763-766 — with subtly different set expressions (`authorised & available_names` vs `available_names`); the divergence risk is exactly what produced the self-contradictory Available list on the team branch. The wording is a pinned contract so exact duplication is deliberate, but this PR already hoisted the two new messages into shared module-level renderers (_UNSHARED_CREATOR_KB_MESSAGE/_PARTIAL_FAILURE_NOTE); consider doing the same for the three terminal templates so they are edited in one place.

src/xagent/web/services/llm_utils.py
```diff
@@ -1079,6 +1079,12 @@ class AgentRuntimeFields:
     # ``agent_id``. Read by the team-scope connector-visibility hook seam;
     # a fail-closed default keeps an un-migrated construction site personal.
     team_id: Optional[int] = None
+    # The Agent row's own creator, never the runner of this task. ``None``
```
rogercloud: This comment claims agent_creator_user_id is None "for every personal/legacy agent", but the only production construction site (llm_utils.py:1314) populates it unconditionally from the non-nullable agent_row.user_id, and the PR's own test (test_agent_runtime_fields_creator.py:121-150) pins a personal agent carrying its creator id. Harmless in behavior (governing_team_id is None for personal agents so the rule is inert), but the comment should read: populated for every agent row; None only on un-migrated construction sites and tasks with no agent. (Related: the _list_knowledge_bases_impl docstring at document_search.py:239-247 makes the mirror-image mistake about when the list tool is built.)

src/xagent/core/tools/adapters/vibe/knowledge_tools.py
AlexLiu190625: Two separate points here, answered separately.

On the accessor: not adding one. The attribute already has six direct readers
inside `WebToolConfig` (config.py:1593, :1792, :2078, :3530, :3546, :3592), and
an accessor introduced here converts none of them — it adds a seventh way of
asking the same question that nothing else adopts, and leaves the next reader to
work out which of the two is authoritative. The comment was the real problem: it
pointed at a rationale that does not exist in config.py. Fixed in adbd80278, which
states the reason in place instead. The two genuinely new values keep their
accessors because they have no existing readers to be inconsistent with.

On delegation: this is out of scope for this PR by an explicit earlier decision,
which we discussed in person — the delegated sub-agent path is deliberately kept
byte-for-byte unchanged here, and there is a test pinning that no-governing-team
behaviour so a later change cannot drift it silently. This PR does not touch
agent_tool.py, so delegation behaves exactly as it did before it. The gap you
describe is real and worth closing at that entry point, so I have opened #1361
with the call sites, the reachability, and the expected contract, including your
point about the config.py:1220-1231 wording.

src/xagent/core/tools/adapters/vibe/knowledge_tools.py
AlexLiu190625: Correct, and the PR body was the thing stating it wrongly. Rule case 4 now says
what actually happens: an agent declaring nothing builds no knowledge tool at
all, because both chat call sites derive the allowed-collections list and the
declaration from the same stored field and an empty list returns no tools. The
fallthrough at the resolution point stays, and the comment there now says it is
defensive — it exists so a future caller supplying a governing team without a
declaration inherits today's behaviour rather than an empty search set. PR body
and comment updated; no code change. adbd80278

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Agreed, fixed in 716398ed5. Both terminal branches now report the names the
verdict loop admitted, captured before the explicit-request narrowing, rather
than `authorised & available_names`. On the no-explicit branch nothing was
admitted, so the clause renders empty; on the explicit branch the pre-narrowing
admitted set is non-empty when other declared names did resolve, which is the
answer the model actually needs there. The same-named-copies test now asserts the
runner's personal copy names never appear in the Available clause.

The "not allowed" early return at :609 I left alone on purpose: it returns before
the verdict loop runs, so there is no admitted set in scope at that point, and
changing it would mean running the full verdict pass — including the creator
probe — for a request the model already got wrong. Happy to revisit if you would
rather it move.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed in 716398ed5. The declaration de-duplication and the three-way gate now
evaluate before this guard, as a named predicate, and the guard is skipped when
that predicate holds — a governing team, an installed hook, and a non-empty
declaration. Every other input keeps the early return exactly as it is. The
missing test is added: runner with zero collections, team owning nothing,
non-empty declaration, parametrised over whether the creator holds the name, so
both the sharing-gap and the does-not-exist verdict are pinned in that edge.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Agreed, and fixed in 716398ed5 along the lines you suggest. The probe now records
whether its one lookup raised; a name left unresolved by a failed lookup is
reported with the hedged wording rather than the terminal certainty text. The
search decision is unchanged — nothing resolves either way — and the disclosure
bound holds: a lookup failure is not per name, and the first failure caches an
empty name set, so every unresolved name of that run gets the same sentence and
the response still says nothing about which collections the creator holds. The
partial-failure path already carried the accurate wording and is untouched. The
test that pinned the old outcome is renamed and now asserts the certainty
wording is absent, with a sibling case for the explicit call style.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Real, and not only here: `allowed_collections` is the same stored value read at
the same two call sites, so normalising the declaration alone leaves the agent
just as broken with a different error surface. That makes it a change at the
read site feeding both consumers, which is wider than this PR. Tracked in
#1362 with the measured behaviour of three bad inputs and the reason a
one-sided fix does not hold.

src/xagent/web/services/knowledge_base_team_scope.py
AlexLiu190625: Keeping the reset-all semantics. This function is deliberately the same shape as
`set_connector_team_hooks` in connector_team_scope.py:93-112, down to the
docstring stating that every slot not supplied is cleared; giving one keyword
preserve-if-omitted semantics would make the two twins behave differently for the
same call, and a reader of either would then have to check which one they are
looking at before predicting what an omitted keyword does. The reset-all contract
is also the one that makes an installer's complete hook set visible at a single
call site, which is what an audit of installers actually reads. The docstring
states the requirement explicitly, and the paragraph about auditing existing
installers when a slot is added is there for exactly the case you describe.

tests/core/tools/core/test_document_search_team_declared_names.py
AlexLiu190625: Both done — the docstring in adbd80278, the new test in 716398ed5. The
exemption is now documented on the test itself: the second site is the
agent-builder console, which constructs the tool with only a user id and an
admin flag, carries no allowed-collections list, and runs for the person
configuring an agent rather than inside any agent's run — so it has no
governing team to resolve against and keeps today's behaviour. The docstring also
says why the invariant is a count rather than "every site threads the values": a
third site, or a refactor routing the console through knowledge_tools.py, changes
which statement holds and should be re-decided rather than absorbed silently.

The empty-union verdict test is added too, in the same commit as the fix to the
early return it exercises.

src/xagent/core/tools/adapters/vibe/factory.py
AlexLiu190625: Fixed in adbd80278: the contract now lists KnowledgeBaseScopeError alongside the
other two propagatable types.

src/xagent/web/services/knowledge_base_team_scope.py
AlexLiu190625: Keeping it here, with the asymmetry documented in adbd80278. The connector seam's
tests reset by calling the setter with no arguments, which restores the empty
state rather than the state the test found — with process-global slots that
leaves any hook installed before the test gone, and the failure lands in an
unrelated test later. Saving and restoring lives on this module because the state
being saved lives on this module: a test-side helper would have to reach these
globals from outside and would go stale the moment a slot is added here. The
docstring now says all of that.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Not hoisting these. The three terminal texts are a reviewed user-facing contract
reused verbatim by the legacy branch, and a shared renderer changes both branches
at once — including the one this PR promised to leave byte for byte unchanged, on
deployments where the team-scope hook is not installed. The divergence you point
at was in the set expression, not the template, and that is fixed directly: both
team-branch terminals now derive their list from the verdict-admitted names. I'd
rather revisit the extraction once the legacy branch has no remaining callers.

src/xagent/web/services/llm_utils.py
AlexLiu190625: Both reworded in adbd80278. The dataclass comment now says the single construction
site fills the field unconditionally from the Agent row for a personal agent as
well as a team one, and that the None default covers a task with no Agent row
rather than the personal-agent case. The list-path docstring now states the real
condition — the list tool is built when the allowed-collections value is None, a
non-empty list builds only the search tool, and an empty list builds no knowledge
tool at all.

AlexLiu190625:
Restructured this into two stacked PRs rather than one.

- **This PR now carries the seam half**: the team layer resolving on the
  governing agent's team, the typed scope error and its two re-wrap points, and
  the governing agent's creator and stored declaration reaching the search entry
  point unread. The branch was rewritten accordingly, so the diff here is now
  the seam only, and no behaviour changes for a deployment that has not
  installed a team-keyed hook.
- **#1371 — the rule that reads those values**: per-declared-name verdicts, the
  re-based "not allowed" gate, the creator-existence probe and the two report
  messages. Its branch is stacked on this one, so until this PR merges its diff
  shows both halves — the rule itself is the last two commits there — and it
  collapses to a single file plus its tests once this merges.

All existing threads and the replies to them are untouched and still on this PR.
Where the open points landed:

- The typed-error narrowings, the reset-all setter and the private-attribute
  read are in this PR, next to their threads.
- The `Available:` derivation, the early-return ordering and the probe-failure
  wording are in the rule half; the replies already given still stand, and I can
  re-answer them on #1371 if you would rather they live next to the code.
- The delegated sub-agent path and the declaration-cleanup point remain out of
  scope for both and stay tracked in their issues.

Suggest reviewing this PR first — it is the larger of the two but almost
entirely plumbing, and it makes #1371 read as a single-file change.

src/xagent/core/tools/core/document_search.py
```diff
@@ -147,35 +191,67 @@ async def list_knowledge_bases(
...
+        agent_creator_user_id: Accepted for signature symmetry with
+            ``_search_knowledge_base_impl`` and never read here.
+            ``knowledge_tools.py`` builds the list tool only when the
+            agent's allowed collections are ``None`` -- a non-empty list
+            builds the search tool alone, and an empty list builds no
+            knowledge tool at all -- and both chat call sites derive that
+            value and the declaration from the same stored field, so on this
+            path there is no declared name to classify at all. See
+            ``declared_knowledge_bases`` below.
+        declared_knowledge_bases: Accepted for signature symmetry and never
```
rogercloud, line 238: `agent_creator_user_id`/`declared_knowledge_bases` are carried through 17 signatures (chat.py → config.py → knowledge_tools.py → both impls) and never read on either path — the docstrings here and in `_search_knowledge_base_impl` disclose this honestly. Note the tension: this docstring claims the "declared names present while a list tool is built" combination cannot occur, yet test_knowledge_tools_team_scope_wiring.py:129-159 pins exactly that combination as valid (list tool built with `declared_knowledge_bases=["kb1"]`). Fine to ship as the #1371 seam; worth reconciling the two statements when the rule lands.

src/xagent/web/services/knowledge_base_team_scope.py
```diff
@@ -74,6 +139,192 @@ def visible_team_knowledge_bases(
...
+        if isinstance(element.storage_user_id, bool) or not isinstance(
+            element.storage_user_id, int
+        ):
+            # bool storage_user_id, or a non-int storage_user_id.
+            raise ValueError(
+                "team knowledge-base visibility hook returned a malformed "
+                f"answer: storage_user_id must be an int (not bool), got "
+                f"{element.storage_user_id!r} for {element.name!r}"
+            )
+        if element.can_edit is not False or element.can_delete is not False:
```
rogercloud, line 206: The fail-open guard is deliberate and well-documented here, but this private docstring is the only place hook authors learn that `can_edit`/`can_delete` must be exactly `False`: `KnowledgeBaseAccess` defaults them to `True` (lines 45-46), so a hook that constructs entries from the dataclass defaults — or omits the flags — gets its whole answer rejected and every call 503s via `resolve_team_knowledge_bases_or_raise` (lines 249-283). Consider a short hook-author note in the module docstring or a README, since this is the public contract of the opt-in seam.

src/xagent/core/tools/core/document_search.py
```diff
@@ -34,29 +35,68 @@ def _get_tool_compatibility_facade() -> "KBToolCompatibilityFacade":
...
     from ....web.services.db_runtime import run_db_io_cancellation_safe
     from ....web.services.knowledge_base_team_scope import (
         has_knowledge_base_visibility_hook,
+        resolve_team_knowledge_bases_or_raise,
+        team_knowledge_base_hook_installed,
         visible_team_knowledge_bases,
     )
 
-    if not has_knowledge_base_visibility_hook():
+    if governing_team_id is not None and team_knowledge_base_hook_installed():
```
rogercloud, line 74: TOCTOU: the branch is selected on `team_knowledge_base_hook_installed()` here, then after the await, `resolve_team_knowledge_bases_or_raise` → `team_knowledge_bases` re-checks the slot (knowledge_base_team_scope.py:233) and silently returns `[]` if the hook was cleared in between — collapsing into the exact "team owns nothing" failure the typed error exists to prevent. Theoretical today (hooks are installed once at startup and there is no production installer), but the re-check could raise instead of returning `[]` once the caller has already selected on the predicate.

src/xagent/core/tools/core/document_search.py
```diff
@@ -34,29 +35,68 @@ def _get_tool_compatibility_facade() -> "KBToolCompatibilityFacade":
...
+    is_admin: bool,
+    governing_team_id: Optional[int] = None,
 ) -> ListCollectionsResult:
-    """Union personal collections with application-provided team overlays."""
+    """Union personal collections with application-provided team overlays.
+
+    For a team-governed run (``governing_team_id`` set and a team-keyed hook
+    installed), the team layer is the *governing* team's own knowledge
+    bases, resolved through ``resolve_team_knowledge_bases_or_raise`` --
+    never the runner's own team memberships, and never a union of the two.
```
rogercloud, line 47: As written, this docstring (and commit ea5a8187's message) claims the governing branch is "never a union of the two". For an admin runner the base `result` is the unrestricted platform view (`list_collections(user_id=user_id, is_admin=is_admin)`, line 57), and the merge loop only re-stamps hook-answered names — so the visible set is in fact a union of the unrestricted view with the team layer. The claim holds only for non-admin runners (or when the hook happens to answer nothing). The admin case in test_document_search_team_layer.py:169-198 does not model the union remainder because the spy ignores `is_admin`. Suggest rewording to scope the promise to non-admin runners.

tests/core/tools/core/test_document_search_team_layer.py
```diff
@@ -0,0 +1,443 @@
...
+    a third site appearing, or a refactor routing the builder console
+    through ``knowledge_tools.py``, changes which of those two statements
+    holds and should be re-decided rather than silently absorbed.
+    """
+    import pathlib
+    import re
+    import subprocess
+
+    repo_root = pathlib.Path(__file__).resolve().parents[4]
+    output = subprocess.run(
```
rogercloud, line 499: `git grep` via subprocess with `check=False` and no `.git` guard: outside a git checkout (sdist/wheel test runs) this yields empty stdout and the count assertion fails spuriously; gitignore-honored paths are also never counted, and the regex can match docstring mentions. CI runs in git checkouts so this is green today — consider an explicit `.git` presence guard or a pure-Python walk.

src/xagent/core/tools/core/document_search.py
```diff
@@ -199,6 +275,8 @@ async def _list_knowledge_bases_impl(
 
         return ListKnowledgeBasesResult(knowledge_bases=kb_list)
 
+    except KnowledgeBaseScopeError:
```
rogercloud, line 280: This `except KnowledgeBaseScopeError: raise` (and the ToolRegistry twin at factory.py:254-261) is never driven by any test: test_knowledge_tools_team_scope_wiring.py:67 builds tools via `_create_knowledge_tools_impl` directly, bypassing ToolRegistry. Deleting either clause would keep the suite green; a regression's consequence is the typed 503 being re-wrapped into a bare `RuntimeError` or swallowed with a warning. A single test driving the ToolRegistry narrowing would pin the contract.

src/xagent/core/tools/core/document_search.py
```diff
@@ -34,29 +35,68 @@ def _get_tool_compatibility_facade() -> "KBToolCompatibilityFacade":
...
     from ....web.services.knowledge_base_team_scope import (
         has_knowledge_base_visibility_hook,
+        resolve_team_knowledge_bases_or_raise,
+        team_knowledge_base_hook_installed,
         visible_team_knowledge_bases,
     )
 
-    if not has_knowledge_base_visibility_hook():
+    if governing_team_id is not None and team_knowledge_base_hook_installed():
+        # Selection is on the predicate above, never on an empty return: an
```
rogercloud, line 75: The invariant stated here — selection is on the hook-installed predicate, never on an empty answer — is unpinned: no test installs a team hook that answers `[]` and asserts the governing branch is still taken. A regression that raises on `[]` or falls through to the runner-keyed branch would pass the suite.

src/xagent/web/api/chat.py
```diff
@@ -1967,6 +1990,12 @@ async def _build_tools_for_task(
                 if current_agent is not None and current_agent.team_id is not None
                 else None
             )
+            # Captured alongside connector_team_id, for the same reason: both
+            # must describe the agent resolved above, never a preview agent
+            # substituted into ``current_agent`` by a branch further down.
+            agent_creator_user_id = (
+                int(current_agent.user_id) if current_agent is not None else None
```
rogercloud, line 1997: The live-ORM branch's `agent_creator_user_id` binding is pinned only as "not hardcoded None" by the static test; a wrong-but-non-None binding (e.g. reading the task owner instead of `current_agent.user_id`) would pass. The branch is unreachable from production today (the sole caller always passes a snapshot, and both else-arm tests pass `agent_id=None`), so this is future-only exposure — worth a comment or a note in the test when #1371 lands.

tests/core/tools/core/test_document_search_team_layer.py
```diff
@@ -0,0 +1,443 @@
...
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "runner_id,is_admin",
+    [
+        (MEMBER, False),  # a team member
+        (ADMIN, True),  # a platform admin -- admin status must not bypass this
+    ],
+)
+async def test_kb_resolves_on_governing_team(
```
rogercloud, line 169: The headline "never a union of the two" property is not pinned: no test installs both the runner-keyed and the team-keyed hooks and asserts the governing branch consults only the team-keyed one. The single dual-hook test (test_multitenancy.py:837-885) drives `list_collections_api`, a different consumer of the seam. A union regression would pass all tests.

tests/web/services/test_agent_runtime_fields_creator.py
```diff
@@ -0,0 +1,125 @@
...
+    site populates both new-and-old fields off the same loaded ``agent_row``
+    (a wrong binding here would leave ``agent_creator_user_id`` mismatched
+    against ``team_id`` on *every* run, not only a cross-user one); a
+    creator-vs-runner *resolution* scenario, where the two differ, is
+    exercised by the resolution-outcome tests against ``_search_knowledge_base_impl``
+    directly.
+    """
+    creator = _create_user(db_session, "creator-pin-creator")
+    agent = _create_agent(db_session, user_id=int(creator.id), team_id=101)
+    task = _create_task(db_session, user_id=int(creator.id), agent_id=int(agent.id))
```
rogercloud: This fixture makes the task owner equal to the agent creator, so a regression that binds `agent_creator_user_id` from the task owner instead of `agent_row.user_id` (llm_utils.py:1319) passes the suite; the docstring's defense only holds for team-id-mismatch bindings. A creator != owner case would pin the binding source.

rogercloud (approved):
## Follow-up review — Round 2 (re-review of the rewritten branch)

**Head:** `330667f4` · **Base:** `6b98091c` · **Previously reviewed:** `ae9ba997` (2026-08-14T07:15Z) · **Author:** AlexLiu190625

### Summary

This PR lays the seam for resolving a team-governed agent's knowledge bases against the team that owns the *governing* agent: it adds an opt-in team-keyed visibility hook with a typed `KnowledgeBaseScopeError` channel, threads `governing_team_id` through the list/search paths, and carries `agent_creator_user_id` + `declared_knowledge_bases` (inert, signature-symmetry only) from the chat call sites to the tool build. The behavior is deliberately inert in standalone xagent — the team branch is reached only when an application installs the hook, and the stored values are never read. The declared-name resolution rule itself moved to follow-up PR #1371. Blocking: no — recommended event: APPROVE.

- The branch was force-push **rewritten**: the previous review covered a branch that contained the declared-name resolution rule; current head `330667f4` (2 commits: `ea5a8187`, `330667f4`) is seam-only, and the rule is being reviewed in follow-up PR #1371 ("Resolve an agent's declared knowledge bases against its own team").
- Prior findings **M1** and **M2** and the rule-related minors (**m2, m3, m4, m5, m10**) transfer to **#1371**: the code they flagged no longer exists in this PR. M2's fix (declared names captured pre-narrowing, generic terminal message) was verified present in #1371's head; M1's concern (delegated `WebToolConfig` must thread the values once the rule lands) is #1371's to carry.
- **m8** (factory.py:226-229 now lists `KnowledgeBaseScopeError` in the exception contract) and **m11** (doc-comments reworded at llm_utils.py:1091-1101 and document_search.py:227-238) are verified **FIXED** in current HEAD.
- **m1/m6/m9** are **WAIVED** with the author's rationale accepted (documented in-code where relevant: m9 at knowledge_base_team_scope.py:296-305, m1 with explicit line-cited comments).
- **m7** is **PARTIAL**: the construction-site exemption is now documented (test_document_search_team_layer.py:411-425); the empty-union verdict test's subject moved to #1371 with the rule.
- **L227** remains **NOT FIXED** but is now documented as a deliberate seam (see Simplification opportunities).
- The author replied to all 12 prior inline comments (2026-08-14T08:58-08:59Z).

### Prior-findings checklist

| Finding | Severity | Status |
|---|---|---|
| M1 — delegation path does not thread governing values | major | REFACTORED → #1371 |
| M2 — terminal messages show declared-name verdict | major | REFACTORED → #1371 (fix verified present in #1371 head) |
| m1 — `getattr` accessor style for `_connector_team_id` | minor | WAIVED (rationale accepted) |
| m2 — declared-name rule case 4 | minor | REFACTORED → #1371 |
| m3 — empty-visible-set early return | minor | REFACTORED → #1371 |
| m4 — probe `holds()` swallowing failures | minor | REFACTORED → #1371 |
| m5 — non-str declaration entries crash | minor | REFACTORED → #1371 |
| m6 — reset-all setter semantics | minor | WAIVED (documented + audited) |
| m7 — coverage gaps | minor | PARTIAL (exemption documented; verdict subject moved to #1371) |
| m8 — factory docstring exception contract | minor | FIXED (factory.py:226-229) |
| m9 — snapshot primitive asymmetry | minor | WAIVED (documented at knowledge_base_team_scope.py:296-305) |
| m10 — terminal templates duplicated | minor | REFACTORED → #1371 |
| m11 — doc-comments on threaded values | minor | FIXED (llm_utils.py:1091-1101, document_search.py:227-238) |
| L227 — list-path params threaded + unread | simplification | NOT FIXED (documented as deliberate seam) |

### New findings (this round)

No critical or major findings. Twelve new findings, all minor.

**F1 — minor — 17 signatures thread two never-read values.** `agent_creator_user_id`/`declared_knowledge_bases` are carried from `create_default_tools` through `WebToolConfig` → `knowledge_tools.py` → both impls, and neither impl reads them (both docstrings disclose this). The cross-file invariant (declaration is `None` whenever the list tool is built) holds at both chat call sites, but the wiring test pins the violating combination as valid — see inline. (document_search.py:236)

**F2 — minor — fail-open guard contract discoverable only in a private docstring.** The validator rejects any answer whose `can_edit`/`can_delete` are not literally `False`, but `KnowledgeBaseAccess` defaults them to `True`, so a hook built from the dataclass defaults gets its whole answer rejected and every call 503s. Deliberate and tested, but hook authors can learn the requirement only from a private docstring. (knowledge_base_team_scope.py:200)

**F3 — minor — TOCTOU between hook selection and re-check.** The governing branch is selected on `team_knowledge_base_hook_installed()` and then re-checks the slot after an await; a hook cleared in between silently collapses into "team owns nothing" — the exact failure the typed error exists to prevent. Theoretical today. (document_search.py:74)

**F4 — minor — "never a union of the two" is inaccurate for admins.** For an admin runner the base result is the unrestricted platform view and the merge loop only re-stamps hook-answered names, so the visible set is a union. The docstring and commit `ea5a8187`'s message overstate the promise; the admin test models an empty unrestricted view and never exercises the union remainder. (document_search.py:47)

**F6 — minor — subprocess `git grep` in a test fails outside git checkouts.** `check=False` with no `.git` guard yields empty stdout (false failure) in sdist/wheel test runs; gitignore-honored paths are uncounted and the regex can match docstrings. CI is green today. (test_document_search_team_layer.py:431)

**F7 — minor — typed-error narrowing never driven by tests.** The `except KnowledgeBaseScopeError: raise` clause (and its ToolRegistry twin) has no test; the wiring test calls `_create_knowledge_tools_impl` directly, bypassing ToolRegistry. Deleting either clause keeps the suite green. (document_search.py:278)

**F8 — minor — "selection on predicate" invariant unpinned.** No test installs a team hook answering `[]` and asserts the governing branch is still taken; a raise-on-`[]` or fall-through regression would pass the suite. (document_search.py:75)

**F9 — minor — live-ORM branch pinned only as "not hardcoded None".** The static test would pass a wrong-but-non-None binding (e.g. task owner instead of `current_agent.user_id`). Unreachable from production today — future-only exposure. (chat.py:1997)

**F10 — minor (design) — three scalar derivation sites, nine params across six+ signatures.** The governing values are derived at three sites in chat.py (snapshot branch ~1966-1977, live-ORM ~1996-2000, inline ~2749-2756) and threaded as nine parameters, with the naming split `connector_team_id` (web) vs `governing_team_id` (core). When #1371 lands, consider a single frozen `GoverningAgentContext`; derivations are already co-located in pairs, so the refactor is contained. Keep this PR as-is.

**F12 — minor — "never a union of the two" headline property unpinned.** No test installs both hooks and asserts the governing branch consults only the team-keyed one; the single dual-hook test drives `list_collections_api`, a different consumer. A union regression would pass all tests. (test_document_search_team_layer.py:169)

**F13 — minor — fixture makes task owner equal to agent creator.** A regression binding `agent_creator_user_id` from the task owner instead of `agent_row.user_id` would pass the suite; the docstring defense only holds for team-id-mismatch bindings. (test_agent_runtime_fields_creator.py:109)

**F14 — minor — no HTTP catch site maps `KnowledgeBaseScopeError` → 503.** The connector analog exists (chat.py:4210-4213) for `ConnectorRuntimeError`, and the class docstring admits nothing reads `code`/`status_code`/`safe_message`. A resolution failure surfaces as a generic step failure today — consistent with the seam's deliberate inertness, but the follow-up installing a hook must add the catch site (this also decides whether the error's extra fields stay justified; see S1).

- `knowledge_base_scope.py` L18-54: yagni `KnowledgeBaseScopeError` carries code/safe_message/details/status_code + `__repr__` that nothing reads (docstring admits it). Replace with a plain `RuntimeError` subclass, message only — unless #1371's HTTP catch site lands first (see F14). Net ≈ -15.
- `factory.py` L254-261 + `knowledge_tools.py` L86-95: yagni self-documented unreachable `except KnowledgeBaseScopeError: raise` on both sides. Delete both, the factory docstring mention (L226-229), the import, and test_document_search_team_layer.py:287-314 (whose docstring calls the real path "currently unreachable"). S2+S3 must go together. Net ≈ -25.
- `knowledge_base_team_scope.py` L210-218: yagni `team_owned is not True` coherence check — no consumer reads `team_owned`. Delete the check and its parametrized case (test_knowledge_base_team_scope.py:113-124); keep the independent `can_edit`/`can_delete` fail-open guard. Net ≈ -12.
- `test_create_default_tools_team_scope_forwarding.py` L37-51 & L77-91: shrink `_FakeToolConfig` + `create_tools` + 3 monkeypatch lines duplicated verbatim in both tests. Hoist to a module-level fixture (`captured` reset per test). Net ≈ -15.
- `test_document_search_team_layer.py` L135-136, 150-151: shrink `_search()`'s `creator_id`/`declared` kwargs and forwarding are inert (the impl never reads them). Drop the 2 kwargs + 2 forwarding lines + inert args at the call sites (193, 225-226, 283). Net ≈ -7. Do not extend to production search-path params (documented #1371 seam).
- `document_search.py` L196-236: yagni (NOT FIXED) list-path params still threaded + unread across 4 layers. Documented as deliberate #1371 symmetry; if the follow-up doesn't need list-path symmetry, a further -25 is available.

Net line estimate: ≈ **-74** (≈ -99 including L227).

No critical or major findings this round: both prior majors (M1, M2) were REFACTORED into follow-up PR #1371 with no observable manifestation in this PR's code, all twelve new findings are minor, and the simplification items are non-blocking. Blocking: no — recommended event: APPROVE.

rogercloud (approved):
## Follow-up review — Round 2 (re-review of the rewritten branch)

**Head:** `330667f4` · **Base:** `6b98091c` · **Previously reviewed:** `ae9ba997` (2026-08-14T07:15Z) · **Author:** AlexLiu190625

### Summary

This PR lays the seam for resolving a team-governed agent's knowledge bases against the team that owns the *governing* agent: it adds an opt-in team-keyed visibility hook with a typed `KnowledgeBaseScopeError` channel, threads `governing_team_id` through the list/search paths, and carries `agent_creator_user_id` + `declared_knowledge_bases` (inert, signature-symmetry only) from the chat call sites to the tool build. The behavior is deliberately inert in standalone xagent — the team branch is reached only when an application installs the hook, and the stored values are never read. The declared-name resolution rule itself moved to follow-up PR #1371. Blocking: no — recommended event: APPROVE.

- The branch was force-push **rewritten**: the previous review covered a branch that contained the declared-name resolution rule; current head `330667f4` (2 commits: `ea5a8187`, `330667f4`) is seam-only, and the rule is being reviewed in follow-up PR #1371 ("Resolve an agent's declared knowledge bases against its own team").
- Prior findings **M1** and **M2** and the rule-related minors (**m2, m3, m4, m5, m10**) transfer to **#1371**: the code they flagged no longer exists in this PR. M2's fix (declared names captured pre-narrowing, generic terminal message) was verified present in #1371's head; M1's concern (delegated `WebToolConfig` must thread the values once the rule lands) is #1371's to carry.
- **m8** (factory.py:226-229 now lists `KnowledgeBaseScopeError` in the exception contract) and **m11** (doc-comments reworded at llm_utils.py:1091-1101 and document_search.py:227-238) are verified **FIXED** in current HEAD.
- **m1/m6/m9** are **WAIVED** with the author's rationale accepted (documented in-code where relevant: m9 at knowledge_base_team_scope.py:296-305, m1 with explicit line-cited comments).
- **m7** is **PARTIAL**: the construction-site exemption is now documented (test_document_search_team_layer.py:411-425); the empty-union verdict test's subject moved to #1371 with the rule.
- **L227** remains **NOT FIXED** but is now documented as a deliberate seam (see Simplification opportunities).
- The author replied to all 12 prior inline comments (2026-08-14T08:58-08:59Z).

### Prior-findings checklist

| Finding | Severity | Status |
|---|---|---|
| M1 — delegation path does not thread governing values | major | REFACTORED → #1371 |
| M2 — terminal messages show declared-name verdict | major | REFACTORED → #1371 (fix verified present in #1371 head) |
| m1 — `getattr` accessor style for `_connector_team_id` | minor | WAIVED (rationale accepted) |
| m2 — declared-name rule case 4 | minor | REFACTORED → #1371 |
| m3 — empty-visible-set early return | minor | REFACTORED → #1371 |
| m4 — probe `holds()` swallowing failures | minor | REFACTORED → #1371 |
| m5 — non-str declaration entries crash | minor | REFACTORED → #1371 |
| m6 — reset-all setter semantics | minor | WAIVED (documented + audited) |
| m7 — coverage gaps | minor | PARTIAL (exemption documented; verdict subject moved to #1371) |
| m8 — factory docstring exception contract | minor | FIXED (factory.py:226-229) |
| m9 — snapshot primitive asymmetry | minor | WAIVED (documented at knowledge_base_team_scope.py:296-305) |
| m10 — terminal templates duplicated | minor | REFACTORED → #1371 |
| m11 — doc-comments on threaded values | minor | FIXED (llm_utils.py:1091-1101, document_search.py:227-238) |
| L227 — list-path params threaded + unread | simplification | NOT FIXED (documented as deliberate seam) |

### New findings (this round)

No critical or major findings. Twelve new findings, all minor.

**F1 — minor — 17 signatures thread two never-read values.** `agent_creator_user_id`/`declared_knowledge_bases` are carried from `create_default_tools` through `WebToolConfig` → `knowledge_tools.py` → both impls, and neither impl reads them (both docstrings disclose this). The cross-file invariant (declaration is `None` whenever the list tool is built) holds at both chat call sites, but the wiring test pins the violating combination as valid — see inline. (document_search.py:236)

**F2 — minor — fail-open guard contract discoverable only in a private docstring.** The validator rejects any answer whose `can_edit`/`can_delete` are not literally `False`, but `KnowledgeBaseAccess` defaults them to `True`, so a hook built from the dataclass defaults gets its whole answer rejected and every call 503s. Deliberate and tested, but hook authors can learn the requirement only from a private docstring. (knowledge_base_team_scope.py:200)

**F3 — minor — TOCTOU between hook selection and re-check.** The governing branch is selected on `team_knowledge_base_hook_installed()` and then re-checks the slot after an await; a hook cleared in between silently collapses into "team owns nothing" — the exact failure the typed error exists to prevent. Theoretical today. (document_search.py:74)

**F4 — minor — "never a union of the two" is inaccurate for admins.** For an admin runner the base result is the unrestricted platform view and the merge loop only re-stamps hook-answered names, so the visible set is a union. The docstring and commit `ea5a8187`'s message overstate the promise; the admin test models an empty unrestricted view and never exercises the union remainder. (document_search.py:47)

**F6 — minor — subprocess `git grep` in a test fails outside git checkouts.** `check=False` with no `.git` guard yields empty stdout (false failure) in sdist/wheel test runs; gitignore-honored paths are uncounted and the regex can match docstrings. CI is green today. (test_document_search_team_layer.py:431)

**F7 — minor — typed-error narrowing never driven by tests.** The `except KnowledgeBaseScopeError: raise` clause (and its ToolRegistry twin) has no test; the wiring test calls `_create_knowledge_tools_impl` directly, bypassing ToolRegistry. Deleting either clause keeps the suite green. (document_search.py:278)

**F8 — minor — "selection on predicate" invariant unpinned.** No test installs a team hook answering `[]` and asserts the governing branch is still taken; a raise-on-`[]` or fall-through regression would pass the suite. (document_search.py:75)

**F9 — minor — live-ORM branch pinned only as "not hardcoded None".** The static test would pass a wrong-but-non-None binding (e.g. task owner instead of `current_agent.user_id`). Unreachable from production today — future-only exposure. (chat.py:1997)

**F10 — minor (design) — three scalar derivation sites, nine params across six+ signatures.** The governing values are derived at three sites in chat.py (snapshot branch ~1966-1977, live-ORM ~1996-2000, inline ~2749-2756) and threaded as nine parameters, with the naming split `connector_team_id` (web) vs `governing_team_id` (core). When #1371 lands, consider a single frozen `GoverningAgentContext`; derivations are already co-located in pairs, so the refactor is contained. Keep this PR as-is.

**F12 — minor — "never a union of the two" headline property unpinned.** No test installs both hooks and asserts the governing branch consults only the team-keyed one; the single dual-hook test drives `list_collections_api`, a different consumer. A union regression would pass all tests. (test_document_search_team_layer.py:169)

**F13 — minor — fixture makes task owner equal to agent creator.** A regression binding `agent_creator_user_id` from the task owner instead of `agent_row.user_id` would pass the suite; the docstring defense only holds for team-id-mismatch bindings. (test_agent_runtime_fields_creator.py:109)

**F14 — minor — no HTTP catch site maps `KnowledgeBaseScopeError` → 503.** The connector analog exists (chat.py:4210-4213) for `ConnectorRuntimeError`, and the class docstring admits nothing reads `code`/`status_code`/`safe_message`. A resolution failure surfaces as a generic step failure today — consistent with the seam's deliberate inertness, but the follow-up installing a hook must add the catch site (this also decides whether the error's extra fields stay justified; see S1).

- `knowledge_base_scope.py` L18-54: yagni `KnowledgeBaseScopeError` carries code/safe_message/details/status_code + `__repr__` that nothing reads (docstring admits it). Replace with a plain `RuntimeError` subclass, message only — unless #1371's HTTP catch site lands first (see F14). Net ≈ -15.
- `factory.py` L254-261 + `knowledge_tools.py` L86-95: yagni self-documented unreachable `except KnowledgeBaseScopeError: raise` on both sides. Delete both, the factory docstring mention (L226-229), the import, and test_document_search_team_layer.py:287-314 (whose docstring calls the real path "currently unreachable"). S2+S3 must go together. Net ≈ -25.
- `knowledge_base_team_scope.py` L210-218: yagni `team_owned is not True` coherence check — no consumer reads `team_owned`. Delete the check and its parametrized case (test_knowledge_base_team_scope.py:113-124); keep the independent `can_edit`/`can_delete` fail-open guard. Net ≈ -12.
- `test_create_default_tools_team_scope_forwarding.py` L37-51 & L77-91: shrink `_FakeToolConfig` + `create_tools` + 3 monkeypatch lines duplicated verbatim in both tests. Hoist to a module-level fixture (`captured` reset per test). Net ≈ -15.
- `test_document_search_team_layer.py` L135-136, 150-151: shrink `_search()`'s `creator_id`/`declared` kwargs and forwarding are inert (the impl never reads them). Drop the 2 kwargs + 2 forwarding lines + inert args at the call sites (193, 225-226, 283). Net ≈ -7. Do not extend to production search-path params (documented #1371 seam).
- `document_search.py` L196-236: yagni (NOT FIXED) list-path params still threaded + unread across 4 layers. Documented as deliberate #1371 symmetry; if the follow-up doesn't need list-path symmetry, a further -25 is available.

Net line estimate: ≈ **-74** (≈ -99 including L227).

No critical or major findings this round: both prior majors (M1, M2) were REFACTORED into follow-up PR #1371 with no observable manifestation in this PR's code, all twelve new findings are minor, and the simplification items are non-blocking. Blocking: no — recommended event: APPROVE.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Deferring this one to the follow-up PR that adds the declared-name rule, where
these two parameters gain their reader and the docstrings get their final shape.
One factual note for the record: on both chat call sites the declaration and
`allowed_collections` are bound from the same expression, so the combination the
docstring calls out cannot be produced in production — the test constructs it by
hand purely to isolate the forwarding. The wording will be aligned with that
reality over there rather than reworded twice.

src/xagent/web/services/knowledge_base_team_scope.py
AlexLiu190625: Agreed — the requirement lived only on the private validator. The module
docstring now states it where a hook author starts reading: both flags default
to `True` on the dataclass, an answer that leaves them unset or `True` is
rejected as a whole, and the rejection surfaces as this channel's typed 503.
Fixed in the "Document the fail-open guard contract" commit.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Real, and filed as #1374 rather than fixed here. The window only opens for an
application that re-installs hooks while serving traffic, which nothing does
today — installation is import-time in practice, and test resets are
process-local. Closing it means changing `team_knowledge_bases`' contract
(raise on the re-check instead of returning empty), and that contract change
deserves its own change alongside whatever first makes runtime hook churn
possible, not a silent ride here.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: The sentence you quote is about the team layer's source — the governing team
versus the runner's own memberships — not about the breadth of an admin's view,
but the comment below it did read as if the admin's platform-wide view were
being narrowed, and it is not: only matching names are re-stamped with the
team's ownership metadata, the rest of the view stays as it was. The comment now
says exactly that. Reworded in the "Document the fail-open guard contract"
commit.

tests/core/tools/core/test_document_search_team_layer.py
AlexLiu190625: Fixed in the "Document the fail-open guard contract" commit: the test now probes
`git rev-parse --is-inside-work-tree` first and skips, with a reason, when the
checkout is not a git work tree or git itself is unavailable. Behaviour inside a
normal checkout is unchanged, and the skip was exercised by temporarily hiding
the work tree link.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Keeping the branches and not adding tests for them here — resolving this
together with your simplification note that pointed the other way. The two
passthroughs with no test driver (the list implementation's and the factory's)
are defence in depth for a path that cannot raise the typed error until an
application installs the hook; a test would have to fabricate that state, and a
fabricated pin on an unreachable path is the kind of test that outlives its
meaning. The change that installs the hook makes those paths real, and the
real-path tests belong to it. Deleting the branches instead would just mean
re-adding them there.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Pinned now. The new dual-hook test installs the governing-team hook answering an
empty list alongside the runner-keyed hook answering a real collection, and
asserts the governed branch still takes the team answer — the runner-keyed hook
is never consulted and its entries never appear. An implementation that selected
on answer emptiness rather than on the installed predicate fails it, which was
verified by making exactly that break. In the "Document the fail-open guard
contract" commit.

src/xagent/web/api/chat.py
AlexLiu190625: Agreed it is only pinned as "not hardcoded None". That branch is unreachable
today — every production caller passes the snapshot — and its binding gains a
behavioural consumer in the follow-up PR that adds the declared-name rule, so
the behavioural pin lands there with the rest of that surface rather than as a
fabricated scenario here.

tests/core/tools/core/test_document_search_team_layer.py
AlexLiu190625: Pinned in the same new dual-hook test: with both hooks installed and answering
different sets, the governed branch's result contains only the governing team's
entries, and unioning in the runner-keyed answer was verified to fail it. One
test carries both this invariant and the predicate-selection one, since they are
only observable in the same two-hook state. In the "Document the fail-open guard
contract" commit.

tests/web/services/test_agent_runtime_fields_creator.py
AlexLiu190625: Fixed in the "Document the fail-open guard contract" commit: the agent's creator
and the task's owner are now two different users, and the test asserts the field
carries the creator and not the owner. Breaking the binding to read the task
owner was verified to fail it.

AlexLiu190625:
All ten inline points are answered in place. The two body-level ones and the
simplification list, for completeness:

- **The three derivation points and the nine-parameter surface**: agreed it is
  the right shape to revisit, and it lands with the follow-up PR (#1371) that
  gives these values their reader — reshaping the surface here and again there
  would churn the same lines twice.
- **The missing HTTP mapping for the typed error**: also #1371's. This PR keeps
  the error's status fields because that mapping consumes them; trimming the
  class now would mean rebuilding it one PR later.
- Of the simplification list: the malformed-answer table stays as is (each case
  is load-bearing at both validation layers), the defensive passthroughs stay
  for the reason on the inline thread, and the double-check race is tracked as
  #1374.

Changes in this round are one commit, documentation and tests only: the hook
answer contract stated at the module level, the admin comment corrected, a git
guard on the construction-site count, the dual-hook pin for never-union and
predicate-selection, and the creator/owner fixture split.

PR #1350, opened 2026-08-14
feat(interactions): refuse a legacy resume that bypasses an unanswered question

AlexLiu190625:
## Summary

Refuses a legacy resume that would bypass an unanswered native interaction row.

When a task carries an active native interaction row anchored to its current run, that row holds the answer slot: the question it represents has not been answered through the interaction protocol yet. A legacy resume command arriving on that task would inject a user message straight into the checkpoint, leaving the row active and the question unanswered while the run continues as if it had been. This change makes the resume coordinator look for such a row first and refuse the resume when one is present without a matching receipt, telling the caller to answer the question instead.

## What's here

- `_active_native_interaction_id_sync` in `websocket.py`: opens its own short-lived session, checks the interaction table exists, and returns the id of the task's active native row (or `None`). Every failure mode — no session factory configured, session open failure, table absent, query failure — resolves to `None`, so a resume is never blocked by an infrastructure problem. The two deployment-window cases (no factory, table absent) stay silent; the two genuine-fault cases (session open, query) log a warning. Four tests pin that distinction.
- The refusal branch in `_handle_resume_task_unserialized`: when an active row exists and the command's receipt does not name it, the resume is rejected with a message pointing at the answer path, and `INTERACTION_LEGACY_RESUME_SHIM` is registered so operators can see the legacy path is still in use.
- `INTERACTION_LEGACY_RESUME_SHIM` in `ops_signals.py`, with its removal condition stated on the constant.

The active-row criteria this reads (`_active_native_row_criteria`) already exist on main; this change adds no new judgment of its own about what makes a row active.

## Behavior today

No production code writes native interaction rows yet, so the table is empty and this check resolves to "no active row" on every real request — the refusal branch is unreachable in production until the finalizer wiring lands. The zero-production-caller gate on the lifecycle service's `create` is the regression guard for that fact. This change is therefore behavior-preserving on every currently reachable path; it takes effect the moment a writer exists.

## Note on scoping

This was originally part of a larger change that also carried the answer-side entry point. That change's description argued this seam could not be split off because it depended on the entry point. **That argument was wrong, and the correction belongs here rather than in a later comment.** A coverage run over the seam's own tests shows the entry point's function body executes zero lines: the seam reads only the active-row criteria that already exist on main, and never calls the entry point. The seam's production diff is 145 lines of pure addition against main, with no dependency on the rest of that work. It is split off here on that evidence.

### DIFF
--- a/src/xagent/web/api/websocket.py
+++ b/src/xagent/web/api/websocket.py
@@ -7869,8 +7869,18 @@ async def _handle_resume_task_unserialized(
         if active_interaction_id is not None:
             receipt_interaction_id = message_data.get("interaction_id")
             receipt_responder_identity = message_data.get("responder_identity")
+            # isinstance before comparing, the same shape the cancel
+            # command's own state-version guard uses below: `True == 1` and
+            # `1.0 == 1` both hold in Python, so a bare `!=` against the
+            # row's int id accepts a JSON `true` as the receipt for row 1
+            # and a JSON `5.0` as the receipt for row 5. A receipt this
+            # seam cannot recognize as the exact int respond() staged is no
+            # receipt at all, so the type check is part of the comparison,
+            # not a separate validation step a caller could skip.
             if (
-                receipt_interaction_id != active_interaction_id
+                isinstance(receipt_interaction_id, bool)
+                or not isinstance(receipt_interaction_id, int)
+                or receipt_interaction_id != active_interaction_id
                 or not receipt_responder_identity
             ):
                 from ..services import ops_signals
--- a/src/xagent/web/api/websocket.py
+++ b/src/xagent/web/api/websocket.py
@@ -7695,16 +7695,22 @@ def _active_native_interaction_id_sync(task_id: int) -> int | None:
     """The id of this task's active native interaction row, scoped to the
     task's current run, or ``None`` if there is none.
 
-    Uses the identical four-field predicate the answer fence and
-    ``materialize_compatibility_view`` both key off
-    (``task_interaction_service._active_native_row_criteria``): status,
-    active_slot, and a join against ``Task.run_id``. All three call sites
-    must keep changing together, or the fence, the read surface, and this
-    resume seam would disagree about which row -- if any -- is "the" live
-    one for a given task. An active row anchored to a run the task has
-    since moved past is invisible here for the same reason it is invisible
-    to those two: ``_reclaim_stale_slot_stmt`` recycles it on the next
-    question, so it carries no obligation for this seam either.
+    Uses the identical four-field predicate
+    (``task_interaction_service._active_native_row_criteria``) every reader
+    of "this task's one live row" keys off: status, active_slot, and a join
+    against ``Task.run_id``. That predicate has three call sites in this
+    tree -- ``task_interaction_service``'s own ``list_active`` and
+    ``_active_native_row`` (the latter is how
+    ``materialize_compatibility_view`` reads the row), plus this seam. The
+    answer fence is a fourth, future caller: it does not exist here yet and
+    lands with ``respond()``. All of them must keep changing together, or
+    the read surface, this resume seam, and the fence once it arrives would
+    disagree about which row -- if any -- is "the" live one for a given
+    task. An active row anchored to a run the task has since moved past is
+    invisible here for the same reason it is invisible to the other
+    readers: ``_reclaim_stale_slot_stmt``
+    (``task_interaction_staging.py``) recycles it on the next question, so
+    it carries no obligation for this seam either.
 
     Gated on ``interaction_requests_table_exists`` for the same reason
     ``materialize_compatibility_view`` gates on it: a deployment can run
@@ -7863,6 +7869,13 @@ async def _handle_resume_task_unserialized(
         # append to or replan around an unanswered question. This runs
         # before agent_service is built (below) so a refused request never
         # pays for constructing one.
+        #
+        # Residual window, named here rather than closed here: this lookup
+        # opens and closes its own session, and no lock spans it and either
+        # RESUME transition below, so the row can in principle change
+        # between this read and the transition. Nothing can drive that
+        # change until respond()'s finalizer exists, and that finalizer --
+        # not this seam -- is what must own the window when it lands.
         active_interaction_id = await run_db_io_cancellation_safe(
             lambda: _active_native_interaction_id_sync(task_id)
         )
--- a/src/xagent/web/api/websocket.py
+++ b/src/xagent/web/api/websocket.py
@@ -7691,6 +7691,87 @@ async def handle_resume_task(
     )
 
 
+def _active_native_interaction_id_sync(task_id: int) -> int | None:
+    """The id of this task's active native interaction row, scoped to the
+    task's current run, or ``None`` if there is none.
+
+    Uses the identical four-field predicate the answer fence and
+    ``materialize_compatibility_view`` both key off
+    (``task_interaction_service._active_native_row_criteria``): status,
+    active_slot, and a join against ``Task.run_id``. All three call sites
+    must keep changing together, or the fence, the read surface, and this
+    resume seam would disagree about which row -- if any -- is "the" live
+    one for a given task. An active row anchored to a run the task has
+    since moved past is invisible here for the same reason it is invisible
+    to those two: ``_reclaim_stale_slot_stmt`` recycles it on the next
+    question, so it carries no obligation for this seam either.
+
+    Gated on ``interaction_requests_table_exists`` for the same reason
+    ``materialize_compatibility_view`` gates on it: a deployment can run
+    this code before the migration that creates
+    ``task_interaction_requests`` has been applied, and this seam must
+    survive that window rather than raising.
+
+    Uses ``get_optional_session_local`` rather than ``get_session_local``,
+    and wraps the read in a broad ``except Exception`` -- the same
+    fail-open shape ``_resolve_read_direction_anchor``
+    (``task_interaction_service.py``) already uses for its own read against
+    this same kind of infrastructure. Every production entry point into
+    this handler runs after ``configure_db()``/``init_db()`` against one
+    database for the life of the process, so neither branch is expected to
+    ever fire there. Both exist for this handler's test callers, which
+    mock the rest of the resume path precisely to avoid needing a database
+    at all, and which do not each get an isolated process: a session
+    factory some other test left installed as the process-global default
+    can point at a since-removed temporary database file by the time an
+    unrelated test reaches this call. Either way, "cannot determine
+    whether there is an active row" is treated as "assume there is not" --
+    refusing every resume because this one read failed would be a worse
+    failure mode than the one legacy-resume window this seam closes.
+    """
+
+    from ..models.database import get_optional_session_local
+    from ..models.task import Task
+    from ..models.task_interaction import TaskInteractionRequest
+    from ..services.task_interaction_schema import interaction_requests_table_exists
+    from ..services.task_interaction_service import _active_native_row_criteria
+
+    SessionLocal = get_optional_session_local()
+    if SessionLocal is None:
+        return None
+    try:
+        db = SessionLocal()
+    except Exception:
+        logger.warning(
+            "resume interaction seam: could not open a session for task_id=%s",
+            task_id,
+            exc_info=True,
+        )
+        return None
+    try:
+        if not interaction_requests_table_exists(db):
+            return None
+        row = (
+            db.query(TaskInteractionRequest.id)
+            .join(Task, Task.id == TaskInteractionRequest.task_id)
+            .filter(
+                TaskInteractionRequest.task_id == task_id,
+                *_active_native_row_criteria(),
+            )
+            .first()
+        )
+        return int(row[0]) if row is not None else None
+    except Exception:
+        logger.warning(
+            "resume interaction seam: active-row lookup failed for task_id=%s",
+            task_id,
+            exc_info=True,
+        )
+        return None
+    finally:
+        db.close()
+
+
 async def _handle_resume_task_unserialized(
     websocket: WebSocket, task_id: int, message_data: dict
 ) -> None:
@@ -7771,6 +7852,58 @@ async def _handle_resume_task_unserialized(
             control_state=control_state,
             status=task_status,
         ).as_dict()
+
+        # Compatibility seam into the interaction lifecycle service: both
+        # resume paths below (the supports_live_control branch and the
+        # bare resume_execution fallback) reach the durable RESUME
+        # transition unconditionally, with no notion of a pending question.
+        # A task that still has an active native interaction row has one:
+        # if this resume's own command payload cannot prove it is the
+        # continuation respond() staged, refuse rather than let either path
+        # append to or replan around an unanswered question. This runs
+        # before agent_service is built (below) so a refused request never
+        # pays for constructing one.
+        active_interaction_id = await run_db_io_cancellation_safe(
+            lambda: _active_native_interaction_id_sync(task_id)
+        )
+        if active_interaction_id is not None:
+            receipt_interaction_id = message_data.get("interaction_id")
+            receipt_responder_identity = message_data.get("responder_identity")
+            if (
+                receipt_interaction_id != active_interaction_id
+                or not receipt_responder_identity
+            ):
+                from ..services import ops_signals
+
+                ops_signals.register_degradation(
+                    ops_signals.INTERACTION_LEGACY_RESUME_SHIM,
+                    f"task {task_id} run {task_fields.run_id}: legacy resume "
+                    f"refused, active interaction {active_interaction_id} has "
+                    "not been answered through respond()",
+                )
+                logger.warning(
+                    "legacy resume refused for task_id=%s run_id=%s "
+                    "interaction_id=%s: active interaction has not been "
+                    "answered through respond()",
+                    task_id,
+                    task_fields.run_id,
+                    active_interaction_id,
+                )
+                message_data["_durable_command_error"] = (
+                    "This task has an unanswered question; answer it before resuming."
+                )
+                await manager.send_personal_message(
+                    {
+                        "type": "error",
+                        "message": (
+                            "This task has an unanswered question; answer it "
+                            "before resuming."
+                        ),
+                        "task": {"id": task_id, **resume_control_state},
+                    },
+                    websocket,
+                )
+                return
 
         agent_service = await get_agent_manager().get_agent_for_task(
             task_id,
--- a/tests/web/api/test_resume_interaction_seam.py
+++ b/tests/web/api/test_resume_interaction_seam.py
@@ -369,10 +369,25 @@ def _exact_id(interaction_id: int) -> object:
     return interaction_id
 
 
+def _wrong_int_id(interaction_id: int) -> object:
+    """A well-typed int naming a different row.
+
+    The most realistic bad receipt: a client resuming with the receipt of
+    an earlier, already-retired question. Both ``isinstance`` terms pass,
+    so only the ``!=`` comparison itself refuses this case -- the
+    wrong-typed cases above all short-circuit on the type checks and never
+    reach it, which would otherwise leave the id equality with no coverage
+    in the True direction.
+    """
+
+    return interaction_id + 1
+
+
 _UNVERIFIABLE_RECEIPTS = [
     pytest.param(_string_form_id, True, id="string-form-id"),
     pytest.param(_float_form_id, True, id="float-form-id"),
     pytest.param(_json_true_id, True, id="json-true-id"),
+    pytest.param(_wrong_int_id, True, id="wrong-int-id"),
     pytest.param(_exact_id, False, id="exact-id-without-responder-identity"),
 ]
 
--- a/tests/web/api/test_resume_interaction_seam.py
+++ b/tests/web/api/test_resume_interaction_seam.py
@@ -433,3 +433,132 @@ async def test_stale_run_active_row_does_not_trip_the_seam(
         await asyncio.wait_for(resume_scheduled.wait(), timeout=1)
 
     assert resume_scheduled.is_set()
+
+
+# ---------------------------------------------------------------------------
+# _active_native_interaction_id_sync's own four fail-open branches, called
+# directly rather than through the full handler: no database configured yet,
+# a session that fails to open, the interaction table not existing yet, and
+# the row lookup itself raising. The helper's docstring argues at length for
+# why each one resolves to "assume no active row" instead of propagating --
+# these pin that argument down to actual behavior, and distinguish the two
+# branches that are expected in normal operation (no session factory yet,
+# table not migrated yet -- no warning) from the two that represent a genuine
+# failure worth a log line (session open failure, lookup failure).
+# ---------------------------------------------------------------------------
+
+
+def test_active_native_interaction_id_sync_returns_none_without_a_session_factory(
+    monkeypatch: pytest.MonkeyPatch,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """``get_optional_session_local() is None`` -- no database configured yet
+    for this process -- is the cheap, expected-in-tests case: it must return
+    ``None`` without ever calling the (nonexistent) session factory, and
+    without logging a warning. A caller that removed this branch would fall
+    through to ``SessionLocal()`` with ``SessionLocal is None``, which raises
+    ``TypeError`` and is instead caught by the *next* branch below -- still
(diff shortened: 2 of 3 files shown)


src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
...
+        active_interaction_id = await run_db_io_cancellation_safe(
+            lambda: _active_native_interaction_id_sync(task_id)
+        )
+        if active_interaction_id is not None:
+            receipt_interaction_id = message_data.get("interaction_id")
+            receipt_responder_identity = message_data.get("responder_identity")
+            if (
+                receipt_interaction_id != active_interaction_id
+                or not receipt_responder_identity
+            ):
```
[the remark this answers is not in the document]
AlexLiu190625: Checked this against the code on the branch: there is no try/except and no fallback value here. The extraction is a plain `message_data.get("interaction_id")`, and the comparison is `receipt_interaction_id != active_interaction_id`. A missing key, a `null`, or a wrongly-typed value all compare unequal to the active row id (an autoincrement integer primary key, so the left-hand side is always a positive int inside this branch) and therefore land on the refusal path — which is not silent: it registers an operational degradation signal exposed on `/health`, emits a structured warning, returns an actionable error to the client, and marks the durable command as terminally rejected.

Treating an unparseable receipt as "no valid receipt" is the intended fail-closed behavior: the guard asks whether the command can prove it is the staged continuation, not whether its fields are well-formed. Raising `ValueError` here would regress that. The receipt fields are not sent by the current client at all, so a missing `interaction_id` is the ordinary legacy path rather than an implementation bug. The raise would be caught by this handler's own `except (ValueError, KeyError, TypeError)`, which means the degradation signal would never be registered, the user-facing message would become a generic data-validation error instead of "answer the question before resuming", and the task control state would drop out of the error payload. Two tests in this change pin exactly that behavior.

Not taking the suggested rewrite, but the round did surface two small gaps, both fixed in the follow-up commit now on this branch: a new test pins the wrongly-typed-receipt case to the refusal path (it held before but was unasserted — verified by mutating the comparison to coerce first, which turns the new test red), and the `INTERACTION_LEGACY_RESUME_SHIM` comment that still described the signal as registered on an accepted resume now says refused, matching the implementation.

src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
...
+                logger.warning(
+                    "legacy resume refused for task_id=%s run_id=%s "
+                    "interaction_id=%s: active interaction has not been "
+                    "answered through respond()",
+                    task_id,
+                    task_fields.run_id,
+                    active_interaction_id,
+                )
+                message_data["_durable_command_error"] = (
+                    "This task has an unanswered question; answer it before resuming."
```
rogercloud, line 7916: Minor: this refusal message is unsatisfiable for the unanswerable tier — materialize_compatibility_view renders those rows with interactions=None (task_interaction_service.py:1301-1308), so they cannot be answered at all, yet the message tells the user to answer the question. The refusal itself is correct; only the message quality is at issue. Consider a tailored message for rows that cannot be answered (e.g. "this task's question can no longer be answered; retry or replace the task") so users are never instructed to do the impossible.

src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
...
+                )
+                logger.warning(
+                    "legacy resume refused for task_id=%s run_id=%s "
+                    "interaction_id=%s: active interaction has not been "
+                    "answered through respond()",
+                    task_id,
+                    task_fields.run_id,
+                    active_interaction_id,
+                )
+                message_data["_durable_command_error"] = (
```
rogercloud, line 7915: Minor: message_data["_durable_command_error"] is the sole mechanism converting this refusal into a terminal durable rejection — _execute_durable_task_command re-raises it as TaskCommandRejected (websocket.py:8204-8206) — yet no test asserts this. All five handler tests call _handle_resume_task_unserialized directly, so deleting this assignment leaves the whole suite green. Suggest a test driving a RESUME command through _execute_durable_task_command with a seeded active row and asserting TaskCommandRejected, following the pattern at tests/web/api/test_websocket_owner_actor.py:2610-2645.

src/xagent/web/api/websocket.py
```diff
@@ -7691,6 +7691,87 @@ async def handle_resume_task(
...
+            db.query(TaskInteractionRequest.id)
+            .join(Task, Task.id == TaskInteractionRequest.task_id)
+            .filter(
+                TaskInteractionRequest.task_id == task_id,
+                *_active_native_row_criteria(),
+            )
+            .first()
+        )
+        return int(row[0]) if row is not None else None
+    except Exception:
```
rogercloud, line 7770: Minor: this fail-open branch (and the session-open failure branch at 7744-7750) is log-only, while the precedent the docstring cites — _resolve_read_direction_anchor — registers CHECKPOINT_LOAD_UNAVAILABLE on its read failure. A persistent lookup failure silently degrades the seam to always-allow with no machine-readable signal, indistinguishable from "no active row". The fail-open decision is correct; consider registering a signal on the failure branches so the always-allow state stays observable.

src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
...
+        if active_interaction_id is not None:
+            receipt_interaction_id = message_data.get("interaction_id")
+            receipt_responder_identity = message_data.get("responder_identity")
+            if (
+                receipt_interaction_id != active_interaction_id
+                or not receipt_responder_identity
+            ):
+                from ..services import ops_signals
+
+                ops_signals.register_degradation(
```
rogercloud, line 7901: Minor: INTERACTION_LEGACY_RESUME_SHIM is registered on the seam's own correct behavior (a designed refusal), latches sticky with no clear site, and flips /health degraded-looking until restart — while the registry charter (ops_signals.py module docstring) is degraded-mode alerting. A designed refusal is not degraded mode, and the frequency already lives in the structured warning log. Consider a counter-style/clearable mechanism, or document why a sticky flag is the intended surface for "legacy path still in use".

src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
...
+        # before agent_service is built (below) so a refused request never
+        # pays for constructing one.
+        active_interaction_id = await run_db_io_cancellation_safe(
+            lambda: _active_native_interaction_id_sync(task_id)
+        )
+        if active_interaction_id is not None:
+            receipt_interaction_id = message_data.get("interaction_id")
+            receipt_responder_identity = message_data.get("responder_identity")
+            if (
+                receipt_interaction_id != active_interaction_id
```
rogercloud: Minor: the `or not receipt_responder_identity` disjunct is never exercised in the True direction — all refusal tests short-circuit on the != comparison, so deleting it passes the whole suite. The truthiness check is as strict as currently possible (CHECK responded_at IS NULL implies responder_identity IS NULL), so this is coverage, not correctness. Suggest one test: matching int interaction_id with responder_identity omitted → refusal.

src/xagent/web/api/websocket.py
```diff
@@ -7691,6 +7691,87 @@ async def handle_resume_task(
     )
 
 
+def _active_native_interaction_id_sync(task_id: int) -> int | None:
+    """The id of this task's active native interaction row, scoped to the
+    task's current run, or ``None`` if there is none.
+
+    Uses the identical four-field predicate the answer fence and
```
rogercloud: Minor: this docstring claims an "answer fence" call site that does not exist in the repo — the predicate's own docstring says the fence is future — and claims "All three call sites"; today's callers are list_active, _active_native_row, and this seam. tests/web/api/test_resume_interaction_seam.py:608 cites the same non-existent _answer_fence_stmt. Documentation-only fix: name the real call sites and mark the fence as future.

src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
...
+        # resume paths below (the supports_live_control branch and the
+        # bare resume_execution fallback) reach the durable RESUME
+        # transition unconditionally, with no notion of a pending question.
+        # A task that still has an active native interaction row has one:
+        # if this resume's own command payload cannot prove it is the
+        # continuation respond() staged, refuse rather than let either path
+        # append to or replan around an unanswered question. This runs
+        # before agent_service is built (below) so a refused request never
+        # pays for constructing one.
+        active_interaction_id = await run_db_io_cancellation_safe(
```
rogercloud, line 7879: Minor: inherent TOCTOU — the lookup session closes before either RESUME transition, and no lock spans lookup→transition. This is unreachable until the respond() finalizer lands, but the long docstring never names this residual window. A one-line note that the row may transition between this read and the transition — and that the finalizer must own that window — suffices.

src/xagent/web/api/websocket.py
```diff
@@ -7772,6 +7853,58 @@ async def _handle_resume_task_unserialized(
             status=task_status,
         ).as_dict()
 
+        # Compatibility seam into the interaction lifecycle service: both
```
rogercloud, line 7862: Minor: refused requests still pay for resolve_execution_scope_off_turn (resolver + snapshot-loader DB read) — the seam consumes nothing it produces and could run immediately after the snapshot, ahead of this resolution. Bounded waste on the exact path the seam was designed to short-circuit; consider moving the seam before the scope resolution once the respond() finalizer lands.

tests/web/api/test_resume_interaction_seam.py
```diff
@@ -0,0 +1,632 @@
...
+    sent = connection_manager.send_personal_message.await_args.args[0]
+    assert sent["type"] == "error"
+    assert "unanswered question" in sent["message"]
+    assert (
+        ops_signals.INTERACTION_LEGACY_RESUME_SHIM in ops_signals.active_degradations()
+    )
+
+
+@pytest.mark.asyncio
+async def test_legacy_resume_without_a_receipt_is_refused_on_the_fallback_path(
```
rogercloud, line 249: Minor (test coverage): (a) no valid-receipt pass-through test on the resume_execution fallback path — only refusal; the matching-receipt test (361-435) covers only the supports_live_control path. (b) No present-but-falsy responder_identity case (see the disjunct at websocket.py:7873). (c) No assertion that the signal is NOT registered on a receipt match — an unconditional register_degradation would pass the suite. (d) The refusal tests leave the process-global signal registered; repo precedent is an autouse cleanup fixture (test_interaction_handoff_surface.py:107-113).

rogercloud (approved):
# Follow-up review — PR #1350: refuse a legacy resume that bypasses an unanswered question

## PR summary

This PR adds a compatibility seam to the RESUME command handler (`src/xagent/web/api/websocket.py`): before either resume path runs, it checks whether the task has an active native interaction row — a question not yet answered through the future `respond()` endpoint. If such a row exists and the incoming resume payload cannot prove it is the continuation `respond()` staged (via an `interaction_id` + `responder_identity` receipt), the resume is refused: an error is sent to the caller, a durable `TaskCommandRejected` is raised, and a degradation signal is registered. The intent is fail-closed protection for the window before `respond()` exists, so no resume can append to or replan around an unanswered question. The PR is 5 commits (+777 lines: seam, signal, and a 632-line test suite), has no linked issues, and CI is fully green (10 checks).

## Update summary

Since the prior review (commit 06296164), commit f8b155fe pins the type-strict receipt comparison: a new test (`test_wrongly_typed_receipt_id_is_refused_like_a_missing_one`) asserts that a string-form `interaction_id` — even the string form of the active row's own id — lands on the refusal path, protecting the deliberate plain-`!=` semantics against a future coercion-first rewrite. The same commit fixes a stale comment in `ops_signals.py` that claimed the signal is registered when a resume is accepted; it now correctly states the signal is set when a legacy resume is refused. Both follow-up items from the prior round landed; the prior P1 inline finding is fixed and the P2 body-only summary was declined with verified-correct reasoning (plain `!=` is deliberately fail-closed across types).

The seam is a reasonable, well-documented fail-closed guard for a transitional period, and its core behavior (refuse a resume that cannot prove it is the continuation of an answered interaction) is correct, with a substantial test suite behind it. The reservation is the design finding below, which does not block: the accept-by-receipt branch is currently unreachable by construction, so the seam is effectively unconditional refusal until `respond()` lands — which is exactly the desired protection today, but the ordering contract must be pinned down when the finalizer is built.

### F1 (design-level, body-only): the receipt-acceptance contract is producer-less

No production code stages `interaction_id`/`responder_identity` into a RESUME payload, so the accept-by-receipt branch has no producer. `InteractionResponseReceipt` (`task_interaction_service.py:470-493`) shows `respond()`'s outcome is a row already `status == "answered"` plus `command_db_id`, and the module docstring states `respond()` "will own its own session and its own commit" (`task_interaction_service.py:12-13`). When `respond()` lands, the row is answered *before* the continuation executes, so `_active_native_row_criteria` (`status == "active"`, `task_interaction_service.py:971`) never sees it — and the accept-by-receipt branch becomes unreachable, the seam collapsing to unconditional refusal. Today that is exactly the desired protection (no `respond()` exists), but the ordering contradiction — a continuation being enqueued while its row is still active — must be specified where `respond()` lands, or the receipt comparison and its type-strictness pin are dead weight. Additionally, the receipt is a shape check, not proof: `responder_identity` is never compared against the re-resolved actor, so a receipt copied from a legitimate continuation is indistinguishable from a forged one once the field exists.

## Line-level findings (all minor, ordered by importance)

### F10 (minor) — refusal message instructs the impossible for unanswerable rows — `websocket.py:7893`

The refusal message "This task has an unanswered question; answer it before resuming." is unsatisfiable for the unanswerable tier: `materialize_compatibility_view` renders rows with `interactions=None` (`task_interaction_service.py:1301-1308`), which cannot be answered at all, yet the message tells the user to answer them. The refusal itself is correct; only the message quality is at issue. A tailored message for the unanswerable tier (e.g. "this task's question can no longer be answered; retry or replace the task") would avoid instructing the impossible.

### F3 (minor) — the durable-rejection mechanism is asserted by no test — `websocket.py:7892-7894`

`message_data["_durable_command_error"]` is the sole mechanism converting this refusal into a terminal durable rejection: `_execute_durable_task_command` re-raises it as `TaskCommandRejected` (`websocket.py:8204-8206`). No test covers this — all five handler tests call `_handle_resume_task_unserialized` directly, so deleting the assignment leaves the whole suite green. Suggest a test driving a RESUME command through `_execute_durable_task_command` with a seeded active row and asserting `TaskCommandRejected`, following the pattern at `tests/web/api/test_websocket_owner_actor.py:2610-2645`.

### F8 (minor) — fail-open lookup failures are unobservable — `websocket.py:7764-7770`

The two fail-open branches (session-open failure at 7744-7750, lookup failure at 7764-7770) are log-only, whereas the precedent the docstring itself cites — `_resolve_read_direction_anchor` — registers `CHECKPOINT_LOAD_UNAVAILABLE` on its read failure. A persistent lookup failure therefore silently degrades the seam to always-allow with no machine-readable signal, indistinguishable from "no active row". The fail-open decision itself is correct; only the observability is missing — e.g. a signal registered on the failure branches.

### F7 (minor) — a designed refusal is registered as degraded mode — `websocket.py:7878-7882`

`INTERACTION_LEGACY_RESUME_SHIM` is registered on the seam's own correct behavior (a designed refusal), latches sticky with no clear site, and flips `/health` degraded-looking until restart — while the registry charter (`ops_signals.py` module docstring) is degraded-mode alerting for operational degradations. A designed refusal is not degraded mode, and the frequency already lives in the structured warning log. Either adopt a counter-style/clearable mechanism, or document why a sticky flag is the intended surface for "legacy path still in use".

### F9 (minor) — test-suite gaps — `tests/web/api/test_resume_interaction_seam.py`

(a) No valid-receipt pass-through test on the `resume_execution` fallback path — only refusal (tests 226-291); the matching-receipt test (361-435) covers only the `supports_live_control` path. (b) No present-but-falsy `responder_identity` case (see F4). (c) No test asserting the signal is NOT registered on a receipt match — an unconditional `register_degradation` would pass the suite. (d) The three refusal tests leave the process-global signal registered; repo precedent is an autouse cleanup fixture (`tests/web/api/test_interaction_handoff_surface.py:107-113`).

### F4 (minor) — the `or not receipt_responder_identity` disjunct is never exercised — `websocket.py:7873-7874`

All refusal tests short-circuit on the `!=` comparison, so the second disjunct never fires in the True direction; deleting it passes the whole suite. The truthiness check is as strict as currently possible (`CHECK responded_at IS NULL` implies `responder_identity IS NULL`), so this is a coverage gap, not a correctness defect. Suggest one test: matching int `interaction_id` with `responder_identity` omitted → refusal.

### F5 (minor) — docstring cites a non-existent "answer fence" call site — `websocket.py:7698-7703`

The docstring claims the predicate is keyed by "the answer fence" and "All three call sites", but no answer-fence call site exists in the repo — the predicate's own docstring says the fence is future. Today's callers are `list_active`, `_active_native_row`, and this seam. `tests/web/api/test_resume_interaction_seam.py:608` cites the same non-existent `_answer_fence_stmt`. Documentation-only fix: name the real call sites and mark the fence as future.

### F6 (minor) — un-named TOCTOU window — `websocket.py:7866-7868`

Inherent time-of-check/time-of-use: the lookup session closes before either RESUME transition, and no lock spans lookup→transition. This is unreachable until the `respond()` finalizer lands, but the (very long) docstring never names this residual window. A one-line note that the row may transition between this read and the transition — and that the finalizer must own that window — suffices.

### F12 (minor) — refused requests still pay for off-turn scope resolution — `websocket.py:7840-7842`

Refused requests still run `resolve_execution_scope_off_turn` (resolver + snapshot-loader DB read) before the seam refuses; the seam consumes nothing it produces and could run immediately after the snapshot. Bounded waste on the exact path the seam was designed to short-circuit — worth moving the seam ahead of the scope resolution once the `respond()` finalizer lands.

- `websocket.py:7754` — `shrink:` the seam's join+filter+first() re-implements `_active_native_row` (`task_interaction_service.py:1046`); call it and `return row.id if row is not None else None`, making the keep-in-sync invariant structural (verified identical behavior, −10 lines including dead imports).
- `tests/web/api/test_resume_interaction_seam.py:185` — `shrink:` five near-identical ExitStack patch blocks; extract one harness helper returning stack+mocks (repo precedent `_enter_build_only_patches`). (~60-80 lines)
- `tests/web/api/test_resume_interaction_seam.py:292` — `shrink:` the new wrong-typed-receipt test is a near-verbatim copy of the missing-receipt refusal test (only the payload and two dropped assertions differ, and both dropped assertions hold for the new variant); parametrize the payload over one scaffold. (~60 lines)
- `tests/web/api/test_resume_interaction_seam.py:122` — `delete:` `_snapshot`'s `run_id` parameter has no caller passing a non-default value; drop it.

net: ~210 lines possible

Blocking: no. All confirmed findings are minor — there are no critical or major findings. The seam's core behavior (fail-closed refusal when an active unanswered interaction exists) is correct, thoroughly documented, and covered by a substantial test suite, and CI is fully green (10 checks). Recommended event: APPROVE.

src/xagent/web/api/websocket.py
AlexLiu190625: The refusal is satisfiable for that tier too, so keeping the one message. `close_legacy_resume_interaction` matches on `(task_id, run_id, status='active', active_slot NOT NULL)` and never inspects the anchor, so answering through the legacy message path retires an unanswerable-tier row exactly as it retires an answerable one — `interactions=None` means "no structured controls", not "cannot be answered". Tailoring the message per tier would require this refusal path to call the anchor resolver, which loads the full row plus the trace row and registers degradation signals as a side effect; too much machinery for message polish. Leaving open for your read.

src/xagent/web/api/websocket.py
AlexLiu190625: Agreed the mechanism is untested, and the gap is wider than this seam: there are ten `_durable_command_error` assignment sites across three handlers and not one test references the key — deleting any assignment leaves the suite green while silently downgrading that refusal from terminal to retryable. The nearest existing test mocks the handler away and exercises the `StaleTaskRunError` clause, a different path. Since the gap predates this change and covering it properly means driving real commands through `_execute_durable_task_command` (satisfying the actor/run-id/foreign-runner loads), it is filed as #1370 to cover the whole mechanism at once rather than patching one seam's slice here. Leaving open for your read.

src/xagent/web/api/websocket.py
AlexLiu190625: Deferred, deliberately. The four fail-open branches are log-only today, and two of them exist mainly for test callers per their own docstring — registering a process-global signal there would leak it into every suite that asserts `/health` payloads by equality (this change's own test file just gained the reset fixture for exactly that class of leak). Whether these branches should signal, and with what hygiene, is being decided together with the wiring change that gives the interaction service its production caller; recorded on that change's list rather than patched here. Leaving open for your read.

src/xagent/web/api/websocket.py
AlexLiu190625: Keeping the signal as is. The documentation half already lives on the constant: its comment states this is a boolean flag rather than a counter, that the registry has no notion of count, that the frequency measurement lives in the structured log line the seam also emits, and the removal condition. On the probe concern, `/health` keeps `status: "ok"` regardless of active degradations, so nothing starts failing — the flag is there so an operator can see the legacy path is still being hit, which is precisely degraded-mode information: clients are still resuming through a path that should have gone through the interaction protocol. A counter-style mechanism does not exist in this registry and adding one is outside this change. Leaving open for your read.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed. The receipt table now carries an `exact-id-without-responder-identity` case: the active row's own id, well-typed, with the `responder_identity` key omitted — so the `or not receipt_responder_identity` disjunct is what refuses it, its first coverage in the true direction. Mutation-checked: deleting that disjunct turns exactly this case red while the other cases stay green.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed. The docstring now names today's three real callers of the shared predicate (`list_active`, `_active_native_row` — which is how the compatibility view reads the row — and this seam) and states the answer fence as a fourth, future caller that lands with `respond()`, matching the predicate's own docstring's tense. The keep-changing-together obligation is restated over that corrected set.

src/xagent/web/api/websocket.py
AlexLiu190625: Fixed. The seam's comment block now names the residual window: the lookup opens and closes its own session, no lock spans it and either RESUME transition, nothing can drive a change in that window until the answer-side finalizer exists, and that finalizer — not this seam — is what must own the window when it lands.

src/xagent/web/api/websocket.py
AlexLiu190625: Keeping the current position. `resolve_execution_scope_off_turn` is not only a cost: it re-raises `ExecutionScopeAbstentionMismatchError` — a fail-closed authorization-class guard. Moving the seam ahead of it would mask a scope-authority failure behind "this task has an unanswered question" for a task that has both problems, and guard ordering (most accurate, least leaking error first) is the contract worth keeping. The waste is bounded, only on the refusal path, and the lookup itself is one short-lived session doing a catalog probe plus a single-row query off the event loop. Leaving open for your read.

tests/web/api/test_resume_interaction_seam.py
AlexLiu190625: (b) and (c) are fixed; (a) declined with a reason.

(b): the receipt table now has the `exact-id-without-responder-identity` case — the disjunct's first coverage in the true direction, mutation-checked (deleting the disjunct turns exactly that case red). (c): the matching-receipt test now asserts the signal is NOT registered on the accept path; that assertion is only sound because the file also gained an autouse reset fixture clearing the process-global registry around every test — the same leak was failing `/health`'s exact-equality assertions whenever both files landed on one xdist worker, so the fixture is the larger half of the fix.

(a) declined: the seam executes before the `supports_live_control` branch splits, so a pass-through on the fallback path exercises the identical seam code the existing matching-receipt test already covers; the two refusal tests already pin both paths where the branch matters. Also worth noting on this thread: the same pass surfaced that the receipt comparison accepted JSON `true` and float forms of the id (`True == 1`, `1.0 == 1` — the payload reaches the seam verbatim through the blacklist filter), so the comparison now type-guards with `isinstance` the way the cancel command's state-version check does, and the table pins every shape: string, float, JSON true, a well-typed wrong id, and the missing-identity case.

PR #1339, opened 2026-08-13
fix: stop unprompted file discovery and unbounded creative regeneration

bluefish-08:
Two related failure modes observed on the same image-generation task: the agent
spent its early iterations taking inventory of unrelated files, then burned the
rest of its iteration budget in a generate/inspect/reject loop that never
converged, and the task ended on
`ReActPattern reached max iterations without a final answer.`

## Unprompted file discovery

`list_all_user_files` was described purely as a capability — "List all user
files across all workspaces and uploaded files." Nothing in the description
said when the tool is appropriate, and it advertised cross-workspace reach, so
the model treated it as the entry point for finding source material and listed
files belonging to other tasks, then read the images it found one by one.

Discovering the current task's own inputs this way is always wasted work: the
files attached to a turn are injected into that turn's context with their
file_id. Injection is per turn, though, so the tool does have a legitimate use —
recovering a specific file the user named that has no file_id anywhere in the
current context, including an attachment from an earlier turn of the same task.
The description now says that, and names the anti-patterns (discovering the
current task's inputs, hunting for reference material nobody supplied, taking
inventory before starting).

`generate_image` pushed in the same direction from the other side: it credited
reference images with guiding product identity, palette, and composition, and
then said "When a reference file has already been acquired, pass its path, URL,
or file_id through the images parameter" — with no statement of where such a
reference legitimately comes from. For any brand-specific brief the model
concludes it needs brand reference imagery and goes looking. The description now
restricts references to what the user supplied or the task itself produced.

`static-visual-design` drove the same search from the skill side, and this was
the strongest of the three: for work naming a real brand it said "Look in user
uploads and the task workspace", which the model read as the whole account's
upload history — it listed files from unrelated tasks and had the vision model
describe old creatives to reverse-engineer the brand's identity. A tool
description cannot outrank an explicit skill instruction, so that sentence now
names its legitimate sources: an asset the user gave in this task (an earlier
turn's attachment included, recovered by listing), the current task workspace, and
the brand's own official web presence. The user's other tasks and earlier outputs
are excluded explicitly. Keeping the official-web route in matters — the tool layer
already advertises it (`fetch_web_content`'s "When the task needs a logo … set
include_assets=true"), and a skill that contradicted that would just be outranked.

`edit_image` carries the same reference-source restriction as `generate_image`,
since `generate_image` routes reference-based work straight to it and a
restriction on one description alone only makes the model switch tools.

## Unbounded creative regeneration

[...]

PR #1276, opened 2026-08-11
fix(tools): isolate public File Operation task access

bsbds:
## Summary

- add exact owner-and-task authorization for File Operation on newly created widget and shared-link tasks
- rely on the existing taskless public-upload binding before execution, and propagate the authoritative database task ID through reconstructed, workforce, and delegated agent workspaces
- fail closed when marked task policy, ownership, or task identity is missing or inconsistent
- authorize durable-file materialization only after record, status, storage-key, owner, and task validation
- preserve existing behavior for private and historical tasks through a server-owned, versioned rollout marker

## Security behavior

For newly marked public tasks, File Operation no longer accepts sibling-task, unbound, other-owner, or unregistered external files. The policy applies to reads, existing-file mutations, text/JSON/CSV writes, directory creation, workspace listing, output listing, and `list_all_user_files`.

Valid workspace-local files and exact-task registered files remain available. Generated outputs are registered to the persisted task owner and exact task.

## Scope and compatibility

This is intentionally limited to File Operation on newly created widget and shared-link tasks. A2A-, SDK-, and trigger-created tasks remain unmarked and retain legacy owner-wide File Operation behavior, as do ordinary private and historical tasks. It does not change MCP roots, sandbox access, shell/Python execution, knowledge-base operations, or preview/download authorization.

Isolation is task-level, not agent-level: delegated agents within the same task inherit the parent task identity and share that task's File Operation file set.

The new `create_default_tools` arguments are appended to preserve the complete legacy positional signature.

## Verification

- 497 focused backend regression tests passed
- `pre-commit run --all-files` passed, including Ruff, mypy, frontend lint, and frontend type checking
- two independent GPT-5.6-sol xhigh reviews completed; the final review found no unresolved security or correctness findings

## Deployment

A mixed-version deployment is unsafe after marked public-task creation begins. The deployment guide now requires:

1. a preflight scan for collisions with `__xagent_file_operation_access_version`
2. gating widget and shared-link task creation
3. deploying the same version to API and execution workers
4. re-enabling public task creation only after old workers are removed

Historical tasks must not be backfilled.

Addresses #803.

PR #1352, opened 2026-08-14
fix: narrow the file listing to the current task and stop handing out paths

bluefish-08:
`list_all_user_files` listed every upload across all of the user's tasks, each row
carrying `storage_path` and a `relative_path` that is also absolute. The observed
cost is a reliable opening detour: the model sees a promising filename in the
listing — in an ad-creative task, an earlier task's
`<brand>_locked_design_specs.json` — then reads it by name, by file_id, and
through `execute_command` with `find`, failing every time before starting the
actual work. In one run, 50 of 52 returned rows belonged to other tasks and
produced 7 doomed reads.

Worth stating plainly what those rows were: reading them by file_id fails
(`resolve_file_id` refuses another task's record), but reading them by the
absolute path the listing handed over **succeeds** — `allowed_external_dirs`
whitelists all of `<uploads>/user_<id>`, which is where every upload's bytes
live. So this PR is not "filter out rows that cannot be read". It does two
things: it narrows the listing to the current task, and it stops handing the
model paths. The cost comes with it — the model can no longer read the same
user's uploads from other tasks. That tightening is intended.

The implementation is one switch on the core listing, off by default, that
narrows in SQL when set:

```python
if openable_only and self.owner_user_id is not None:
    query = query.filter(or_(UploadedFile.task_id.is_(None),
                             UploadedFile.task_id == self.current_task_id))
```

The narrowing runs before `count` and the slice, so pagination keeps its meaning
— slicing first and filtering after would make rows on later pages permanently
unreachable. It keys on `current_task_id` rather than the id parsed out of the
workspace name: a delegated workspace is named after the agent run while its
`db_task_id` is the parent task's, and reads authorize against the latter.

The model-facing adapter passes `openable_only=True` and withholds absolute
paths. The switch is deliberately **not** on that adapter's signature:
`FunctionTool` turns every parameter into a field of the generated tool schema,
so exposing it would let the model call `list_all_user_files(openable_only=False)`
and undo the narrowing. An assertion pins it out of the schema.

The core default is off, so `test_unmarked_historical_file_operation_behavior_is_unchanged`
keeps holding and callers that carry their own authority are unaffected.

## Effect

Same ad-creative task, before and after:

| | Before | After |
|---|---|---|
| Listing returned | 52 rows (50 from other tasks) | 0 rows |
| Cross-task read attempts | 7, all failed | 0 |

## Known limits

- This is not a path boundary. `list_files` and `get_file_info` still return
  absolute paths, and the whitelisted upload root still resolves them. What this
  removes is the listing handing over aimable filenames together with their paths.
- Scoped deployments are half-served. The narrowing is on `task_id`, and
  task-less uploads (knowledge-base files among them) are unaffected; in a scoped
  channel every end user of one client application shares the platform owner, so
  a sibling end user's task-less uploads still list while `resolve_file_id`
  refuses them. Fixing that needs the scope subtree, and the two upload endpoints
  disagree on whether stored paths carry scope segments, so it is not settled here.

## Tests

- `pytest tests/core/tools/test_workspace_file_operations.py` — 69 passed

Verified per hunk by mutation: deleting the SQL narrowing fails 4 tests; keying on
the parsed workspace id instead of `current_task_id` fails 1; putting the switch
back on the adapter signature, or dropping the path withholding, is caught too.


rogercloud (changes requested):
## PR Summary

This PR targets the "doomed-read" problem: the model sees sibling-task files in `list_all_user_files` that it can never actually open, because `file_id` resolution refuses another task's records while the absolute-path entry point stays reachable through the whitelisted upload root. It fixes only the model-facing vibe adapter — `WorkspaceFileTools.list_all_user_files` now strips `storage_path`/`relative_path`, filters rows through `_entry_is_openable`, and reports a rewritten `total_count` — leaving the core `WorkspaceFileOperations`/`TaskWorkspace` implementation untouched, consistent with #1276's behavior freeze for unmarked historical tasks. The direction matches #1276 (narrowing listings by `task_id`) and #1339 (keep only this task's legitimately openable files).

**Blocking: yes — recommended event: REQUEST_CHANGES.**

## Round 0 approach verdict: acceptable-with-reservations

Confining the fix to the adapter is the right call: the core listing stays frozen (pinned by `test_unmarked_historical_file_operation_behavior_is_unchanged`, and other callers carry their own authority), path-stripping closes the whitelist-root escape hatch, and the new tests pin the adapter/core boundary. The reservations are the two majors below: filtering runs *after* core pagination and rewrites `total_count` to page-local length (F1), and the mirrored authority predicate misses the scope-subtree leg (F2).

## Severity findings

### F1 — major, correctness: pagination is broken by post-filtering and the page-local `total_count` rewrite

The adapter calls the core with `limit`/`offset`, then filters and overwrites `total_count` with `len(sanitized)` — the current page's length (`src/xagent/core/tools/adapters/vibe/workspace_file_tool.py:200`). For unmarked workspaces the core query narrows only by `user_id` (`src/xagent/core/workspace.py:2125`) and applies `count`/`offset`/`limit` before any task filtering (`workspace.py:2174-2179`), so when a full page consists of unreadable rows the model receives `files=[]` + `total_count=0` and cannot distinguish "no files" from "openable files on later pages". The PR's own verification table — 52 rows, 50 unreadable, default limit 50 — is exactly this case: the 2 openable files on page 2 are silently hidden, and the 0-row result is presented as success. This also undermines the sanctioned recovery use case in #1339 (re-opening an earlier attachment). Suggest narrowing in SQL before `count`/`offset`/`limit` (precedent: the `_exact_task_scope` block, `workspace.py:2126-2173`), or adapter-side paging until `limit` openable rows are gathered; at minimum document the semantics change and add a >limit-row test.

### F2 — major, correctness/design: `_entry_is_openable` mirrors only part of the core authority predicate

The docstring claims it mirrors `TaskWorkspace._file_record_allowed_for_workspace`, but it copies only the task-id branch and omits the scope-subtree check (`workspace.py:1154-1155`) that runs *before* the `task_id is None` acceptance (`workspace.py:1157-1162`). Scoped-but-unmarked workspaces are reachable: the marker is written only for newly created widget/share tasks, while the ExecutionScope resolver hook and workforce tasks produce scoped tasks without it, and #1276 forbids backfilling historical tasks. In that configuration the core listing filters only by `user_id` (`workspace.py:2125`), the adapter keeps out-of-scope `task_id=None` uploads (adapter lines 208-210), and `resolve_file_id` rejects them — the exact doomed-read this PR eliminates, still present in a supported configuration. The omitted `user_id` check is harmless (listed rows are always the owner's own). Suggest extracting the record-level authority predicate into one shared core helper used by both sides, or moving the narrowing into core as an opt-in parameter; at minimum add a scoped-workspace test.

### F3 — minor, quality: `relative_path` re-attachment carries absolute paths

The re-attachment at `workspace_file_tool.py:197-198` keeps the DB row's `relative_path` verbatim, but for `UploadedFile` rows the core stores `str(storage_path)` — an absolute path (`workspace.py:2196`) — so the model still receives absolute paths under the `relative_path` key, contradicting the method's stated "drop absolute paths" contract (lines 173-176). No security impact (retained paths lie inside the current workspace and `read_file` resolves them), but one key now carries two semantics and invites misuse. Re-attach only genuine relative paths, or document the exception.

### F4 — minor, docs: stale docstring summary line

The docstring first line (`workspace_file_tool.py:155`) still says "List all user files across all workspaces and uploaded files." — copied from core; the Returns section was updated but the first line wasn't. Update it to match the new semantics.

### F5 — minor, tests: pagination, scoped, and `include_workspace_files=True` paths untested

The new tests pin the adapter/core boundary but create only 3 records (<< `DEFAULT_USER_FILE_LIST_LIMIT=50`), so no pagination interaction is exercised; there is no scoped-workspace adapter case; and all adapter listing calls pass `include_workspace_files=False`, leaving the `relative_path` re-attachment branch (`197-198`) with zero coverage. The `total_count == 2` assertion (`test_workspace_file_operations.py:1312`) cements the page-local semantics that diverge from the core `query.count()` contract.

### F6 — minor, docs: new tool description is inaccurate across deployment modes

(a) It omits `task_id == current_task_id` uploads stored outside the workspace directory (listed in both modes); (b) in marked deployments SQL narrows by `task_id == task_id` (`workspace.py:2152-2153`), so NULL-`task_id` unattached uploads never appear despite "the user's unattached uploads"; (c) "Other tasks' files are not listed" is false for ownerless workspaces (reachable via `factory.py:879-881`, pinned by this PR's own test); (d) unregistered workspace files carry `file_id=None` (`workspace.py:2234`), so "open a listed file by passing its file_id" is not actionable for them.

### F7 — minor, quality: silent fail-open shape guards

The two shape-defense branches (`workspace_file_tool.py:182-183`, `187-189`) are unreachable today — the core chain always returns a list of dicts or raises — but if the core shape ever drifts they silently pass unsanitized rows (absolute paths, unreadable entries) to the model, the opposite of this method's purpose. Delete them or make them fail closed.

- `delete:` `src/xagent/core/tools/adapters/vibe/workspace_file_tool.py:182-183` — dead `files`-not-list guard (core always returns a list or raises); index `listing["files"]` directly (-3 lines).
- `delete:` `workspace_file_tool.py:187-189` — dead non-dict entry branch; all entries are dicts or the core raised (-3 lines).
- Verified NOT dead: the `getattr` defaults at lines 206/211 defend against `MockWorkspace` (bound via `factory.py:852-859`, which lacks `owner_user_id`/`current_task_id`); only the redundant `bool()` at line 211 is removable: `return task_id == getattr(self.workspace, "current_task_id", None)`.

**Blocking: yes** (CI is green; blocking is on correctness grounds).

- `[new]` [major] `src/xagent/core/tools/adapters/vibe/workspace_file_tool.py:200` — post-pagination filtering plus page-local `total_count` breaks the listing contract: openable rows beyond page 1 become unreachable (the PR's own 52→0 verification is this case).
- `[new]` [major] `src/xagent/core/tools/adapters/vibe/workspace_file_tool.py:202` — `_entry_is_openable` mirrors only the task branch of the core authority check; scoped-but-unmarked workspaces still surface doomed rows (missing scope-subtree check, `workspace.py:1154-1155`).

**Recommended event: REQUEST_CHANGES.**

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -160,10 +160,55 @@ def list_all_user_files(
...
+                continue
+            visible = {
+                key: value
+                for key, value in entry.items()
+                if key not in ("storage_path", "relative_path")
+            }
+            if entry.get("in_current_workspace") and entry.get("relative_path"):
+                visible["relative_path"] = entry["relative_path"]
+            sanitized.append(visible)
+        return {**listing, "files": sanitized, "total_count": len(sanitized)}
```
rogercloud: [major] `total_count` is rewritten to the length of the *filtered current page*, not the filtered total across pages. The core applies `offset`/`limit` and computes `total_count = query.count()` before any task narrowing for unmarked workspaces (`src/xagent/core/workspace.py:2125`, `2174-2179`), so this filter runs after pagination. When a whole page consists of unreadable rows the model receives `files=[]`, `total_count=0`, `limit=50` — indistinguishable from "no files" — and openable rows on later pages are unreachable. The PR's own verification table (52 rows, 50 unreadable, default limit 50) is exactly this case: the 2 openable files on page 2 are silently hidden and the 0-row result is presented as success. Suggest narrowing in SQL before `count`/`offset`/`limit` (precedent: the `_exact_task_scope` block at `workspace.py:2126-2173`), or adapter-side paging until `limit` openable rows are gathered, plus a >limit-row test.

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -160,10 +160,55 @@ def list_all_user_files(
...
+
+    def _entry_is_openable(self, entry: Dict[str, Any]) -> bool:
```
rogercloud: [major] `_entry_is_openable` claims to mirror `TaskWorkspace._file_record_allowed_for_workspace` but copies only the task-id branch: it omits the scope-subtree check (`src/xagent/core/workspace.py:1154-1155`) that runs *before* the `task_id is None` acceptance (`workspace.py:1157-1162`). Scoped-but-unmarked workspaces are reachable: the marker is written only for newly created widget/share tasks, while the ExecutionScope resolver hook and workforce tasks create scoped tasks without it, and #1276 forbids backfilling historical tasks. There, the core listing filters only by `user_id` (`workspace.py:2125`), the adapter keeps out-of-scope `task_id=None` uploads (lines 208-210), and `resolve_file_id` rejects them — the exact doomed-read this PR exists to eliminate. The `user_id` omission alone is harmless (listed rows are always the owner's own). Suggest a single shared authority helper for both core and adapter, or moving the narrowing into core as an opt-in parameter; at minimum add a scoped-workspace test.

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -160,10 +160,55 @@ def list_all_user_files(
...
+                sanitized.append(entry)
+                continue
+            if not self._entry_is_openable(entry):
+                continue
+            visible = {
+                key: value
+                for key, value in entry.items()
+                if key not in ("storage_path", "relative_path")
+            }
+            if entry.get("in_current_workspace") and entry.get("relative_path"):
```
rogercloud: [minor] For `UploadedFile` rows the core stores `relative_path` as `str(storage_path)` — an absolute path (`src/xagent/core/workspace.py:2196`) — and this branch re-attaches that value verbatim under the `relative_path` key whenever the row is `in_current_workspace`. So the model still receives absolute paths under a misleadingly named key, contradicting the method's documented contract to "drop absolute paths" (lines 173-176); `in_current_workspace` is genuinely reachable (e.g. `prepare_html_asset` registers files inside `workspace_dir`). No security impact (the path lies inside the current workspace and `read_file` resolves it), but one key now carries two semantics. Re-attach only when the value is a genuine relative path, or document that retained values are absolute.

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -241,7 +286,7 @@ def get_tools(self) -> List[FunctionTool]:
             FileTool(
                 self.list_all_user_files,
                 name="list_all_user_files",
-                description="List all user files across all workspaces and uploaded files. Supports pagination and filtering. Returns file_id, filename, storage_path, size, mime_type, and whether files are in current workspace.",
+                description="List the user's files that this task can open: the current workspace's files and the user's unattached uploads. Other tasks' files are not listed and cannot be read. Supports pagination and filtering. Returns file_id, filename, size, mime_type, and whether files are in current workspace; open a listed file by passing its file_id to read_file.",
```
rogercloud: [minor] The new description overstates the listing's guarantees: (a) it omits `task_id == current_task_id` uploads stored outside the workspace directory, which are listed in both unmarked (`workspace.py:2125`) and marked (`workspace.py:2153`) modes; (b) in marked deployments SQL narrows by `UploadedFile.task_id == task_id` (`workspace.py:2152-2153`), which excludes NULL-`task_id` unattached uploads — so "the user's unattached uploads" are not listed there at all; (c) "Other tasks' files are not listed" is false for ownerless workspaces (lines 206-207 keep everything; core `workspace.py:1141-1142` agrees; ownerless is production-reachable via `factory.py:879-881` and pinned by this PR's own test); (d) unregistered workspace files have `file_id=None` (`workspace.py:2234`), so "open a listed file by passing its file_id" is not actionable for them. Rewrite the description per deployment mode and the ownerless exception.

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -160,10 +160,55 @@ def list_all_user_files(
...
+        Paths go first: the user's upload root is whitelisted, so an absolute
+        path here resolves into any sibling task's workspace. That leaves
+        file_id as the way in, and file_id resolution refuses another task's
+        record whenever the workspace has an owner. Listing rows that can only
+        fail wastes turns — the model reads the promising filename, retries it
+        by name, by id, and through the shell before giving up. The core
+        listing keeps those rows for callers that hold their own authority.
+        """
+        files = listing.get("files")
+        if not isinstance(files, list):
```
rogercloud: [minor] This guard, and the non-dict branch at lines 187-189, are unreachable today — the core chain (`WorkspaceFileOperations.list_all_user_files` → `TaskWorkspace.list_all_user_files`) either returns `files` as a list of dicts or raises (`src/xagent/core/tools/core/workspace_file_tool.py:717,734`), and no subclass overrides `inner`. But both branches are fail-open: if the core shape ever drifts, they silently pass unsanitized rows (absolute `storage_path`, unreadable entries) straight to the model — the opposite of this method's purpose. Delete them (dead code, ~6 lines) or make them fail closed (raise, or log-and-drop) so a contract drift becomes audible instead of a silent leak.

tests/core/tools/test_workspace_file_operations.py
```diff
@@ -1215,6 +1250,94 @@ def test_list_all_user_files_db_pagination(self, tmp_path):
...
+            workspace.current_task_id = mine.id
+
+            listed = WorkspaceFileTools(workspace).list_all_user_files(
+                include_workspace_files=False
+            )
+            listed_names = {file_info["filename"] for file_info in listed["files"]}
+
+            assert listed_names == {"mine.txt", "unattached.txt"}
+            assert listed["total_count"] == 2
+            # The omitted row is exactly the one file_id resolution would refuse.
```
rogercloud: [minor] The new tests pin the adapter/core boundary well, but leave three branches uncovered: (1) pagination — only 3 records are created vs `DEFAULT_USER_FILE_LIST_LIMIT=50`, so nothing exercises the post-filter page; with >50 rows where early rows are unopenable, the adapter returns a short page and openable rows on later pages are invisible; (2) no scoped-workspace adapter case — the one scoped listing test drives core `ops`; (3) all adapter listing calls pass `include_workspace_files=False`, so the `relative_path` re-attachment branch (`workspace_file_tool.py:197-198`) has zero coverage. The `total_count == 2` assertion also cements the page-local `total_count = len(sanitized)` semantics, which diverge from the core contract `total_count = query.count()`. Suggest a >limit-row pagination case and a scoped case; if the page-local count is deliberate, encode that intent in the test docstring.

bluefish-08:
All six findings are addressed, but not by patching the adapter — F1 and F2 turned out to be symptoms of the filter living in the wrong layer, and following them led to deleting most of what you reviewed. Summary rather than a reply per thread.

**F1 (pagination / page-local `total_count`)** and **F2 (`_entry_is_openable` mirrors only part of the authority check)** — both fixed by moving the narrowing into the core listing, before `count` and the slice, and then removing the record-level pass entirely. Mutation testing showed that within the branch that can reach it, every leg of `_file_record_allowed_for_workspace` either restates the SQL predicate or is already guaranteed by the owner match: deleting the whole Python filter and keeping only the SQL leaves the suite green. What it did add was a full `.all()` fetch replacing `count()` + `LIMIT`, plus two `Path.resolve()` calls per row — measured 60x slower at 5,000 rows, on a local SSD; the uploads root here is network-backed. So the implementation is now one SQL predicate, `count`/`offset`/`limit` are back to the database, and `total_count` means the same thing it always did.

**F3 (`relative_path` carries absolute paths for `UploadedFile` rows)** — fixed; the keep-branch now requires the path to actually be relative, with a unit case for an absolute path arriving under that key.

**F4 (description overstates the guarantees)** — fixed, and twice: the first rewrite still claimed reads on unlisted rows fail, which is false for the absolute-path route. See the correction below.

**F5 (pagination, scoped and `include_workspace_files=True` untested)** — added, and the pagination case had to be rewritten: my first version inserted the openable rows last, so newest-first put them on page one and the test passed even with the pre-fix ordering restored. It now places a full page of unopenable rows ahead of them and probes `offset=limit`.

**F6 (fail-open shape guards)** — deleted.

Three things worth flagging beyond the findings, since they change how this PR should be read:

1. **Its original premise was wrong.** I described the excluded rows as ones the model cannot open. That holds for the `file_id` route only. Before this PR the listing handed out `storage_path` and an absolute `relative_path`, and `read_file(<absolute path>)` on another task's upload *succeeded*, because `allowed_external_dirs` whitelists `<uploads>/user_<id>` and every upload's bytes live there. My tests only showed otherwise because their workspaces have no external dirs, unlike production. So this PR tightens same-user cross-task reachability and stops handing the model paths — it does not merely hide rows that were already unreachable. The PR body now says that.

2. **The switch must not reach the model.** `FunctionTool` turns every parameter into a field of the generated tool schema, so the `openable_only` flag I first put on the adapter override was visible to the model, which could call `list_all_user_files(openable_only=False)` and undo the fix. It is now absent from that signature (the override fixes it to `True` internally), with an assertion that it stays out of the schema. The core keeps the parameter defaulted off, so `test_unmarked_historical_file_operation_behavior_is_unchanged` and any caller carrying its own authority are untouched.

3. **Scoped deployments are only half-served.** Narrowing by `task_id` leaves task-less uploads in place, and in a scoped deployment every end user of one client application shares the platform owner — so a sibling end user's task-less uploads still list while `resolve_file_id` refuses them. Fixing that needs the scope subtree, and the two upload endpoints disagree on whether stored paths carry scope segments, so it is not a call to make in this PR. Filed separately.

Resolving the threads on that basis.

src/xagent/core/workspace.py
```diff
@@ -2171,6 +2175,26 @@ def list_all_user_files(
                             *local_storage_filters,
                         ),
                     )
+                if openable_only and self.owner_user_id is not None:
+                    # Narrow in SQL, before count and the slice, so openable rows
+                    # on later pages stay reachable. A record-level pass over the
+                    # rows would be redundant: within this filter every leg of
+                    # _file_record_allowed_for_workspace either restates the
+                    # predicate below or is already guaranteed by the owner match.
```
rogercloud: This comment overstates the equivalence: within `openable_only`, the SQL predicate (`task_id IS NULL OR task_id == current_task_id`) plus the owner match does not restate every leg of `_file_record_allowed_for_workspace` — the scope-subtree leg (workspace.py:1154) and the path-in-workspace leg (workspace.py:1132-1138) are neither restated by the predicate nor guaranteed by the owner match. The PR's own scoped test pins exactly this: `test_vibe_listing_narrows_by_task_in_scoped_workspaces` lists the unattached row and then asserts `resolve_file_id` refuses it (tests/core/tools/test_workspace_file_operations.py:1481-1485). The mismatch only produces false negatives (rows listed that reads refuse), so the narrowing is safe, but the comment should name the non-restated legs (or the disclosed scoped gap) instead of claiming full equivalence.

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -146,24 +147,58 @@ def get_workspace_output_files(self) -> Dict[str, Any]:
...
 
-    def list_all_user_files(
+    def list_all_user_files(  # type: ignore[override]
         self,
         include_workspace_files: bool = True,
         limit: int = DEFAULT_USER_FILE_LIST_LIMIT,
         offset: int = 0,
     ) -> Dict[str, Any]:
-        """List all user files across all workspaces and uploaded files.
+        """List the user's files this workspace can actually open.
```
rogercloud: Minor: "can actually open" overpromises for scoped deployments — task-less out-of-scope rows still appear in the listing while `resolve_file_id` refuses them (the Known-limits gap, pinned by tests/core/tools/test_workspace_file_operations.py:1481-1485). Suggest wording such as "List the user's files within this task's reach." or mirroring the core wording "as far as this workspace's scope reaches".

src/xagent/core/workspace.py
```diff
@@ -2171,6 +2175,26 @@ def list_all_user_files(
...
+                    # An ownerless workspace opts out because its reads are
+                    # unrestricted too, so a narrowed listing would hide reachable
+                    # files.
+                    from sqlalchemy import or_ as _or
+
+                    query = query.filter(
+                        _or(
+                            UploadedFile.task_id.is_(None),
+                            # current_task_id, not the id parsed from the workspace
```
rogercloud: Minor: this comment describes the marked path, but in unmarked production a delegated workspace id (`agent_<aid>_<hex8>`, built at src/xagent/core/tools/adapters/vibe/agent_tool.py:2151) never parses through `_parse_task_id_from_workspace_id` (workspace.py:1045-1049), so unmarked delegated listings skip the DB branch entirely and show no uploads (pre-existing behavior). The new test `test_vibe_listing_narrows_by_the_authorized_task_id` (tests/core/tools/test_workspace_file_operations.py:1522-1562) models a `web_task_800` workspace with `current_task_id=799` — a divergence the constructor (workspace.py:218-222) makes impossible — while its docstring claims it models a delegated workspace. Suggest correcting the test's docstring/shape or adding a production-shaped `agent_<aid>_<hex8>` case pinning the (currently empty) unmarked delegated listing.

src/xagent/core/workspace.py
```diff
@@ -2063,13 +2063,17 @@ def list_all_user_files(
         offset: int = 0,
         *,
         _exact_task_scope: bool = False,
+        openable_only: bool = False,
```
rogercloud, line 2066: Minor: the docstring's first line at workspace.py:2068 ("List all user files across all workspaces and uploaded files.") no longer describes the whole function now that `openable_only` narrows the listing before count/slice; this PR updated the Args block but left the summary line, while sibling docstrings were updated. Suggest a first line that acknowledges the optional narrowing, e.g. "List user files, narrowed to this task's own uploads when requested."

src/xagent/core/tools/adapters/vibe/workspace_file_tool.py
```diff
@@ -241,7 +276,7 @@ def get_tools(self) -> List[FunctionTool]:
...
                 self.list_all_user_files,
                 name="list_all_user_files",
-                description="List all user files across all workspaces and uploaded files. Supports pagination and filtering. Returns file_id, filename, storage_path, size, mime_type, and whether files are in current workspace.",
+                description="List the user's files that this task is allowed to open. What that covers depends on the deployment's file scope: always this task's own uploads, plus the user's uploads that are not tied to a task where the scope permits them. Files belonging to the user's other tasks are left out wherever the deployment records a file owner, and no filesystem path to them is returned here. Registered uploads are returned newest-first in slices of limit (50 by default) at offset; raise the offset to see older ones. Current-workspace files are appended outside that slicing when include_workspace_files is on, and the unregistered ones among them have no file_id — read those by their relative path. Returns file_id, filename, size, mime_type, and whether a file is in the current workspace; open a listed upload by passing its file_id to read_file.",
```
rogercloud: Minor: the description lead "this task is allowed to open" overpromises in scoped deployments, where task-less out-of-scope rows still appear in the listing but refuse to open (the documented Known-limits gap; the hedges later in this description are good). Suggest a lead such as "List the user's files within this task's reach." so the description matches what the listing actually returns.

src/xagent/core/workspace.py
```diff
@@ -2171,6 +2175,26 @@ def list_all_user_files(
...
+                if openable_only and self.owner_user_id is not None:
+                    # Narrow in SQL, before count and the slice, so openable rows
+                    # on later pages stay reachable. A record-level pass over the
+                    # rows would be redundant: within this filter every leg of
+                    # _file_record_allowed_for_workspace either restates the
+                    # predicate below or is already guaranteed by the owner match.
+                    # An ownerless workspace opts out because its reads are
+                    # unrestricted too, so a narrowed listing would hide reachable
+                    # files.
+                    from sqlalchemy import or_ as _or
```
rogercloud: Cosmetic: the `as _or` alias adds nothing; `from sqlalchemy import or_` would read the same. (The import itself must stay — the `or_` import at workspace.py:2135 lives inside the exact-scope branch, so it is not in scope here.)

rogercloud (approved):
## PR summary

This PR makes the model-facing `list_all_user_files` listing show only the uploads a task can actually open. The core listing narrows in SQL — the task's own uploads plus unattached ones — before counting and slicing, so openable rows on later pages stay reachable and `total_count` reflects the narrowed set; the vibe adapter withholds absolute paths (keeping genuine workspace-relative ones), and the `openable_only` switch is deliberately kept off the tool schema so the model cannot turn the narrowing off. The narrowing is task-only: in scoped deployments, task-less rows outside the scope still appear in the listing while `resolve_file_id` refuses them, a gap documented in the PR body's Known limits. Blocking: no — recommended event: APPROVE.

## Update since the previous review (head 960d9983)

Five commits landed on top of the previously reviewed head (six in the PR overall): the record-level post-filter was replaced by SQL narrowing before count/slice, the record check was skipped where the adapter is not the read authority, scoped workspaces were left unfiltered and the path-boundary claim was dropped, the switch was kept off the tool schema, and the docs were tightened. The adapter's path withholding now keeps genuinely relative workspace paths while absolute upload paths are withheld behind an `os.path.isabs` guard. CI is green, all prior findings are addressed, and the Round-0 design verdict (acceptable-with-reservations: sound, with reservations limited to the documented scoped gap and wording accuracy) still holds.

## Prior findings checklist

- **F1 (major, post-filter `total_count` distortion) — FIXED.** SQL narrowing now happens before count and slice at `src/xagent/core/workspace.py:2178-2197`; `total_count` is `query.count()` over the narrowed query (`workspace.py:2198`), and the adapter no longer rewrites `total_count`.
- **F2 (major, partial authority mirror in scoped workspaces) — WAIVED.** The adapter's `_entry_is_openable` is deleted; the SQL narrowing is task-only, so scoped task-less rows still list-but-refuse. This is deliberately disclosed in the PR body's Known limits and pinned by `test_vibe_listing_narrows_by_task_in_scoped_workspaces` (`tests/core/tools/test_workspace_file_operations.py:1440-1488`, including the `resolve_file_id` refusal assertion at 1485 and its explanatory comment at 1482-1484); `resolve_file_id` still fails closed.
- **F3 (minor, absolute-path re-attachment) — FIXED.** `os.path.isabs` guard at `src/xagent/core/tools/adapters/vibe/workspace_file_tool.py:197`; genuinely relative workspace paths are kept, absolute upload paths are withheld.
- **F4 (minor, stale adapter docstring) — FIXED.** Docstring rewritten at `adapters/vibe/workspace_file_tool.py:156-170`.
- **F5 (minor, test gaps) — FIXED.** Pagination test `test_vibe_listing_pages_past_unopenable_rows` (`tests/core/tools/test_workspace_file_operations.py:1366-1439`, 52 uploads), scoped adapter test (1440-1488), `include_workspace_files=True` test `test_vibe_listing_keeps_relative_workspace_paths` (1489-1520), and `total_count` semantics pinned (1411, 1435).
- **F6 (minor, description inaccuracy) — FIXED.** Description rewritten at `adapters/vibe/workspace_file_tool.py:279` with hedges.
- **F7 (minor, fail-open shape guards) — FIXED.** Guards deleted; the adapter indexes `listing["files"]` directly (line 187), so drift fails loudly.

## New findings (all minor)

1. **[minor] Equivalence comment is false** — `src/xagent/core/workspace.py:2181-2183` claims every leg of `_file_record_allowed_for_workspace` is either restated by the SQL predicate or guaranteed by the owner match. The scope-subtree leg (`workspace.py:1154`) and the path-in-workspace leg (`workspace.py:1132-1138`) are neither restated nor owner-guaranteed; the PR's own scoped test pins a listed row that `resolve_file_id` refuses (`tests/core/tools/test_workspace_file_operations.py:1481-1485`). The mismatch only produces false negatives (rows listed that reads refuse), so the narrowing is safe — suggest rewording the comment to name the non-restated legs / the disclosed scoped gap. (inline)
2. **[minor] Overpromising wording** — the adapter docstring lead "List the user's files this workspace can actually open" (`adapters/vibe/workspace_file_tool.py:156`) and the tool-description lead "List the user's files that this task is allowed to open" (`:279`) are false for scoped deployments, where task-less out-of-scope rows still list but refuse to open (Known limits). Suggest "List the user's files within this task's reach." / mirror the core wording "as far as this workspace's scope reaches"; optionally rename `openable_only` → `task_scoped_only` (polish only). (inline x2)
3. **[minor] Stale docstring summary line** — `src/xagent/core/workspace.py:2068` still opens with "List all user files across all workspaces and uploaded files." although the function now narrows under `openable_only`; this PR rewrote the Args block but left the first line, while sibling docstrings were updated. Suggest a first line acknowledging the optional narrowing. (inline)
4. **[minor] Delegated-keying comment describes the marked path; the new test models an impossible shape** — `src/xagent/core/workspace.py:2192-2194` explains keying on `current_task_id` via delegated workspaces, but in unmarked production a delegated id (`agent_<aid>_<hex8>`, built at `src/xagent/core/tools/adapters/vibe/agent_tool.py:2151`) never parses through `_parse_task_id_from_workspace_id` (`workspace.py:1045-1049`), so unmarked delegated listings skip the DB branch entirely (empty upload listing — pre-existing). `test_vibe_listing_narrows_by_the_authorized_task_id` (`tests/core/tools/test_workspace_file_operations.py:1522-1562`) models a `web_task_800` workspace with `current_task_id=799` — a divergence the constructor (`workspace.py:218-222`) makes impossible — while its docstring claims it models a delegated workspace. Suggest correcting the test docstring/shape or adding a production-shaped `agent_<aid>_<hex8>` case pinning the currently-empty unmarked delegated listing. (inline)
5. **[minor] Dead `task_id IS NULL` leg in marked mode** — under exact scope, `workspace.py:2157` already requires `task_id == db_task_id`, and `current_task_id == db_task_id` by the constructor invariant, so the `task_id IS NULL` leg of the `openable_only` OR (`workspace.py:2191`) can never match in marked mode. This is consistent with marked reads (task-less rows are refused), so it is not a bug; a one-line comment noting the NULL leg only functions in unmarked mode would prevent future confusion.
6. **[minor] PR-body test count is off by one** — the claim that deleting the SQL narrowing fails 5 tests is a miscount: the mutation experiment fails exactly 4 (`test_vibe_listing_omits_other_tasks`, `test_vibe_listing_pages_past_unopenable_rows`, `test_vibe_listing_narrows_by_task_in_scoped_workspaces`, `test_vibe_listing_narrows_by_the_authorized_task_id`; `test_vibe_listing_keeps_relative_workspace_paths` does not depend on the narrowing). Suggest correcting the PR body.

Pre-existing (not introduced here, no action needed): the post-slice existence filter (`workspace.py:2214`) can still make pages shorter than `total_count` suggests.

- `shrink:` src/xagent/core/workspace.py:2187 — `from sqlalchemy import or_ as _or` is a gratuitous alias; plain `or_` reads the same (the import must stay: the `or_` import at 2135 is inside the exact-scope branch). 0 net lines, cosmetic.
- `shrink:` tests/core/tools/test_workspace_file_operations.py — the identical 8-line in-memory engine/session setup + try/finally teardown is now copied five times: pre-existing at 1205 plus four copies added by this PR at 1272-1278, 1373-1379, 1448-1454, 1530-1536. Extract one fixture yielding db; net ≈ -30 lines for the four new copies alone (≈ -39 if the pre-existing copy migrates to the fixture too).
- `yagni:` (considered, rejected) deleting the `openable_only` flag entirely — its `False` default carries the #1276 freeze contract: `test_unmarked_historical_file_operation_behavior_is_unchanged` (tests:1117-1141) fails under unconditional narrowing, and the freeze test's sibling rows are readable by path.

**Blocking: no — recommended event: APPROVE.** All prior findings are fixed or explicitly waived; the remaining findings are minor wording/comment/test-shape issues and optional simplifications, which never block merge. The design is sound and the one behavioral reservation (scoped task-less rows list-but-refuse) is documented and pinned by tests.

bluefish-08:
All six addressed in `f6b65b7d`:

- Equivalence comment reworded to say what this is — task narrowing, not the read authority — and to name the two legs it does not restate (scope-subtree, path-in-workspace), plus the fact that the mismatch only ever lists too much.
- Both overpromising leads are now "List the user's files within this task's reach."
- `workspace.py` docstring first line acknowledges the narrowing.
- The delegated-keying test was the real problem: it modelled a shape production cannot produce. Replaced by two real ones — a genuine key divergence built through the `db_task_id` constructor argument, and `agent_7_9fbc2ad1` pinning that a delegated workspace lists no uploads at all, which is what they actually do.
- Added the note that the `task_id IS NULL` leg is reachable only when unmarked.
- `or_` alias dropped; the five copies of the engine/session setup are now one `memory_db` fixture.

Also corrected the PR body: deleting the SQL narrowing fails 4 tests, not 5 — your count was right, mine was a miscount. 69 passed.

Resolving these.

bluefish-08:
`84ae66e1` brings the tool description in line with `ce1df6a4`. The SQL had stopped admitting task-less rows, but the description still promised them — "always this task's own uploads, plus the user's uploads that are not tied to a task" — so the layer the model actually reads was advertising a set the callable no longer returns.

It now says this task's own uploads and nothing else, and names the case that motivated the narrowing: files left behind by a deleted task. That case is worth stating rather than leaving as an implication, since `purge_task_rows` detaches `UploadedFile` rows instead of deleting them (`task_deletion.py:21-27` — the relationship has no cascade, so the unit of work nulls `task_id` before the task row goes). Every deleted task therefore produces task-less uploads permanently, which is exactly what the old permissive leg was letting back in.

Verified against a real listing that had regressed this way: a task with no uploads of its own was being handed 29 registered files, all leftovers of deleted tasks, with paths still reading `web_task_14/15/16`. After the narrowing it gets none.

Three assertions, each mutation-checked: dropping the deleted-task clause, reverting the path wording, or restoring the old permissive claim each fails the suite.

PR #1353, opened 2026-08-14
feat: let the Get Started video card link out via the home extension point

yiboyasss:
## Summary
- `HomeGetStartedDestinationOverrides` gains a `video` key, resolved through the same `resolveHomeGetStartedDestination` helper already used for `docs`/`guides`/`whatsNew`.
- The Get Started "Video Tutorial" card's `link` now comes from `resolveHomeGetStartedDestination(homeGetStartedDestinationOverrides.video, defaultHomeGetStartedDestinations.video)` instead of the hardcoded `null`.
- The OSS default for `video` is `null`, so with no override configured the card behaves exactly as before (inline autoplaying video, no link wrapper, no hover/focus affordances).
- When a replacement `home-page-extension` module sets `video` to a URL, the card becomes a link (same `target="_blank"` treatment as the other cards) while still autoplaying its inline tutorial video — the link wraps the existing video, it doesn't replace it.

## Why
Downstream distributions (e.g. the SaaS build) want their own "Video Tutorial" card to point at their own hosted how-to-videos page, without forking `page.tsx`. The existing extension point already supports this pattern for docs/guides/whatsNew; this extends it to video for consistency.

## Test plan
- [x] `npx vitest run src/app/page.test.tsx` (111 passed) — includes a new test asserting the video card links out with the inline video intact, and stays inert for null/blank/undefined overrides
- [x] `tsc --noEmit` — no errors
- [x] `eslint` on changed files — no new warnings

### DIFF
--- a/frontend/src/app/page.test.tsx
+++ b/frontend/src/app/page.test.tsx
@@ -365,6 +365,7 @@ describe("Home", () => {
   }
 
   beforeEach(() => {
+    delete homeGetStartedDestinationOverridesMock.video
     delete homeGetStartedDestinationOverridesMock.docs
     delete homeGetStartedDestinationOverridesMock.guides
     delete homeGetStartedDestinationOverridesMock.whatsNew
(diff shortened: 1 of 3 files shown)


rogercloud (approved):
## Summary

The Get Started "Video Tutorial" card hardcoded its link to `null`, so downstream distributions that replace the home-page extension module could not point the card at their own hosted videos page. This PR adds a `video` key to the existing `HomeGetStartedDestinationOverrides` contract (`frontend/src/lib/page-extension-contracts.ts`), resolves it through the shared `resolveHomeGetStartedDestination` helper in `frontend/src/app/page.tsx`, and keeps the OSS default `video: null` so default builds keep the inline autoplaying video with no link wrapper. Tests cover the linked-card behavior, the null/blank inert fallback, and the four-key contract (~41 insertions across three files, CI green).

## Approach

Sound. The change extends the existing extension-point abstraction exactly as designed: one more key in the same contract, one resolver, one interaction owner (the card render), and the exhaustive `Record` plus source-lock keeps key coverage enforced. The video card now reads `link` from the resolver like its three siblings while the inline `<video>` element stays untouched, so default builds are behaviorally identical to before.

**D1 — three-state semantics collapse to two for `video`.** For `docs`/`guides`/`whatsNew`, `null` means "explicitly disable" and is distinct from the canonical default; for `video`, the canonical default is itself `null`, so `undefined` (default) and `null` (explicit disable) are indistinguishable — both render the inert card. This collapse is deliberate, documented in the JSDoc added by this PR, and unobservable in output, so no change is requested.

**D2 — destination-only scope.** The extension point controls the link destination only; the video source stays hardcoded to `/videos/Tutorial.mp4`. A `video: null` override still autoplays the inline video — the card is inert but the media remains. Distributions wanting to replace or suppress the media itself stay outside this extension point, which is exactly the stated intent and is recorded in the JSDoc.

## Findings

### Minor

- **`frontend/src/app/page.tsx:165` — type widening drops the non-null guarantee for the canonical defaults.** Widening `defaultHomeGetStartedDestinations` to `Record<keyof HomeGetStartedDestinationOverrides, string | null>` removes the compile-time guarantee that the `docs`/`guides`/`whatsNew` defaults are non-empty strings: a future `docs: null` would typecheck and silently inert that card. Runtime coverage exists (`page.test.tsx:444-469` guards the defaults) and `page.test.tsx:802` source-locks the current annotation, but that lock must be updated in lockstep. Suggested fix (resolver signature stays `string | null`):
  ```ts
  const defaultHomeGetStartedDestinations: { video: null } & Record<
    Exclude<keyof HomeGetStartedDestinationOverrides, "video">,
    string
  > = {
    video: null,
    docs: "https://docs.xagent.co/api-reference/introduction",
    guides: "https://docs.xagent.co/models/overview",
    whatsNew: "https://docs.xagent.co/release-notes",
  }
  ```

## Simplification

- `shrink:` — `frontend/src/app/page.test.tsx:518`: the `as Record<string, unknown>` cast on `homeGetStartedDestinationOverridesMock.video = invalid` is redundant — `invalid: string | null` assigns directly to `video?: string | null`, and the sibling assignment at line 505 does so without a cast. The cast is erased at runtime; plain assignment compiles identically. net: -1 line possible

Blocking: no — APPROVE. No blocking issues; the findings above are optional polish (one type-precision tightening, one redundant cast).

PR #1354, opened 2026-08-14
feat(interactions): add the interaction lifecycle service's answer-side entry point

AlexLiu190625:
## Summary

Adds `respond()` to the task interaction lifecycle service: the entry point that answers one active native interaction row and stages the command that resumes the task's execution with that answer, in one all-or-nothing transaction.

This is the core of a four-part series. The resume-coordinator seam ships separately (it reads only the active-row criteria that already exist on main — a coverage run proved it never executes this function). Two follow-up changes, staged on top of this one, restore precision that this change deliberately trades for size: fine-grained classification of a missed answer fence, and reconciliation after a lost commit acknowledgment.

## What's here

- The answer fence: one UPDATE that flips the active row to `answered` only if it is still active, still owned by the caller's principal, and the task is still waiting — ownership is enforced at the write, not only in a prior read.
- `RespondEnvelope` validation with type-before-value checks (a list-typed kind or a boolean protocol version is rejected as invalid input, not crashed on or coerced).
- The outcome vocabulary as `Literal` types: each outcome's reason field is typed by the exact set of reasons this change can produce, and a meta-test derives the required test cells from the type — "this change only produces these outcomes" is checked by the type system, not claimed in prose.
- Idempotent replay: a repeat of the same answer under the same content-derived key returns the original receipt instead of double-staging.
- Conservative failure handling (see below), with three tests proving the conservative paths leave zero residue: no orphaned row state, no staged command, no task-state change.
- The zero-production-caller gate watches both `create` and `respond`; neither has a production caller, and the gate is the regression guard for that fact. The change that wires a caller is the change that takes the corresponding name out of the gated set.

## Deliberately conservative in this change

Two failure paths return `OutcomeUnknown` here rather than a precise classification:

- **Fence miss**: when the fence UPDATE matches no row, this change reports the outcome as unknown instead of rereading the row to distinguish "already answered" from "superseded run" and friends. The reread-and-classify block arrives in a follow-up; a test here proves the miss path rolls back cleanly.
- **Commit-ack loss and staging conflict**: when the final commit raises, or staging the resume command hits a concurrent duplicate, this change rolls back and reports unknown instead of reconciling against the durable state. The reconciliation (a fresh-session verification of what actually landed) arrives in the other follow-up; tests here prove both paths leave no residue.

Unknown is conservative in both cases and zero-residue in all of them. Retrying under the same idempotency key converges for an ambiguous commit (the replay check finds the staged command) and for a rolled-back staging race (it simply runs again). A fence miss does not converge — the same predicate misses the same row every time — so the fence-miss path logs the reread row's state, and classifying that miss is the fence classification change's job.

## Why this ships at this size

The series was split four ways, and this change is the part that cannot shrink further without moving safety evidence out of the change that needs it:

- The fence, the envelope validation, the outcome types, and the vocabulary checks live in one module and reference each other directly; any cut separates a definition from its only callers. (Three names are imported rather than defined here — the kind vocabulary and command-id normalizer from the staging and transport modules, and the protocol-version constant from the models package — so "one module" describes where the answer-side logic lives, not a closed import graph.)
- The shared test fixtures, the vocabulary meta-test, the repo-level lock-strength guard, and the gate change all have to land with the first change that contains `respond()`.
- The tests that prove this change is safe to merge — per-cell mutation-verified, including real two-connection PostgreSQL races for the locking claims — stay here by policy. What could move out without weakening that proof has moved out: the classification and reconciliation follow-ups, and this change also drops (not defers — drops, with the rationale in the docstrings) an optimistic-concurrency parameter pair that had no caller and no settled contract.

### DIFF
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -78,6 +78,7 @@ the scanned package tree).
 from __future__ import annotations
 
 import json
+import logging
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Literal, cast
@@ -122,6 +123,8 @@ from .task_lease_service import TASK_RUN_ID_TRACE_FIELD
 
 if TYPE_CHECKING:
     from sqlalchemy.orm import Session
+
+logger = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
 # Principal and the shared public-chat ownership predicate
@@ -1616,9 +1619,6 @@ def _retire_respond_session_best_effort(db: "Session") -> None:
     helper is named for.
     """
 
-    import logging
-
-    logger = logging.getLogger(__name__)
     try:
         db.close()
         return
@@ -1868,9 +1868,7 @@ def respond(
     documented above as unreachable and is in this same category.
     """
 
-    if not isinstance(envelope.kind, str):
-        return RespondValidationRejected(reason="unknown_kind")
-    if envelope.kind not in _KIND_VOCABULARY:
+    if not isinstance(envelope.kind, str) or envelope.kind not in _KIND_VOCABULARY:
         return RespondValidationRejected(reason="unknown_kind")
     if (
         not isinstance(envelope.protocol_version, int)
@@ -2051,9 +2049,7 @@ def respond(
             # under the same idempotency key resolves the ambiguity by
             # itself. The classification this build does not compute is
             # exactly the set of fields logged here.
-            import logging
-
-            logging.getLogger(__name__).warning(
+            logger.warning(
                 "answer fence matched zero rows for interaction %s on task "
                 "%s; reread status=%s active_slot=%s terminal_reason=%s "
                 "run_id=%s responder_identity=%s",
@@ -2147,9 +2143,7 @@ def respond(
             # against the durable graph (that reconciliation is not
             # delivered here): it retires its session and reports the
             # ambiguity, leaving the caller to re-check.
-            import logging
-
-            logging.getLogger(__name__).warning(
+            logger.warning(
                 "commit failed while answering interaction %s on task %s; "
                 "the write may or may not be durable -- a retry under the "
                 "same idempotency key resolves which",
@@ -2180,9 +2174,7 @@ def respond(
             # purpose: this wraps one post-commit best-effort notification
             # and nothing else -- no statement above the commit is inside
             # it.
-            import logging
-
-            logging.getLogger(__name__).warning(
+            logger.warning(
                 "failed to notify the task command dispatcher after "
                 "committing the answer for interaction %s on task %s; "
                 "the dispatcher's idle poll will still pick the command up",
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -1357,14 +1357,27 @@ def _answer_fence_task_predicate(principal: InteractionPrincipal) -> list[Any]:
     row the fence statement joins in via ``Task.id == task_id`` (see
     ``_answer_fence_stmt``), not a second, independently-scoped read.
 
-    The ownership terms are redundant on PostgreSQL -- step 2 of
-    ``respond()`` already holds this row's ``FOR NO KEY UPDATE`` lock -- and
-    load-bearing on SQLite, where the dialect drops every locking clause
-    silently and nothing serializes a caller until this statement, the
-    transaction's first write (see ``respond()``'s own docstring for the
-    two backends' different serialization points). Enforcing ownership at
-    the write point rather than only in ``respond()``'s step 3 is what
-    makes the two backends agree on what a successful answer proves.
+    The ``Task.user_id`` term requires the owner in person, on both
+    backends: step 3's authorization admits an admin acting on another
+    user's task, and the guest ownership predicate never reads
+    ``principal.user_id`` (see ``task_is_owned_by_public_principal``), so
+    for both of those callers this term is the only ownership constraint
+    in the path and the fence misses. Answering on another user's behalf
+    is not delivered in this build; whether an admin may do so is a policy
+    decision left to the change that wires the first production caller. A
+    ``principal.user_id`` of ``None`` compiles to ``tasks.user_id IS
+    NULL`` and matches nothing (the column is ``nullable=False``), so such
+    a caller fails closed here rather than matching another user's row.
+
+    Against a *concurrent* ownership change the term is redundant on
+    PostgreSQL -- step 2 of ``respond()`` already holds this row's
+    ``FOR NO KEY UPDATE`` lock -- and load-bearing on SQLite, where the
+    dialect drops every locking clause silently and nothing serializes a
+    caller until this statement, the transaction's first write (see
+    ``respond()``'s own docstring for the two backends' different
+    serialization points). Enforcing ownership at the write point rather
+    than only in ``respond()``'s step 3 is what makes the two backends
+    agree on what a successful answer proves.
 
     ``Task.status`` is compared through ``TaskStatusPredicate.eq``, never a
     raw string literal beside the column -- this module's own convention
@@ -1696,9 +1709,10 @@ def respond(
        from under this transaction's own row lock, then reports
        ``OutcomeUnknown`` without further classifying why the fence missed
        (a fine-grained classification -- already-answered replay/conflict,
-       three terminated reasons, wrong task state, foreign run, or --
-       reachable only on SQLite, see ``_answer_fence_task_predicate`` -- an
-       ownership change -- is not delivered in this build); ``rowcount > 1``
+       three terminated reasons, wrong task state, foreign run, or an
+       ownership miss, reachable on both backends (see
+       ``_answer_fence_task_predicate``) -- is not delivered in this build);
+       ``rowcount > 1``
        is a schema invariant violation (``uq_task_interaction_active_slot``)
        and raises.
     7. The Task CAS via ``apply_task_control_transition``, called with no
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -1545,8 +1545,26 @@ def _respond_receipt(
     table's audit-authoritative record of who answered (see ``respond()``'s
     own docstring for why ``responder_user_id`` cannot be used for the same
     purpose).
+
+    The only caller is ``respond()``'s idempotent-replay branch, whose
+    precondition -- a staged RESUME command this service itself committed
+    in the same transaction as the answer fence UPDATE -- implies an
+    answered row, and the paired CHECK constraints
+    (``ck_task_interaction_requests_responded_at_pairs_status`` and
+    ``ck_task_interaction_requests_responder_pairs_responded_at``) make an
+    answered row with a NULL ``responded_at`` or ``responder_identity``
+    impossible. That reasoning spans two modules, so the guard below turns
+    it into a loud failure instead of trusting it silently: a receipt must
+    never carry a coerced ``'None'`` identity or a ``None`` timestamp.
     """
 
+    if interaction.responded_at is None or interaction.responder_identity is None:
+        raise RuntimeError(
+            f"interaction {interaction.id} on task {interaction.task_id} "
+            "matched a staged RESUME command but carries no answer; "
+            "ck_task_interaction_requests_responder_pairs_responded_at "
+            "makes this impossible"
+        )
     return InteractionResponseReceipt(
         interaction_id=int(interaction.id),
         task_id=int(interaction.task_id),
@@ -1688,9 +1706,18 @@ def respond(
     3. Pure Python authorization against the row step 2 loaded: a
        ``"user"`` principal must own the task or be an admin; a
        ``"guest"`` principal must satisfy the shared
-       ``task_is_owned_by_public_principal`` predicate. Runs before the
-       idempotency pre-read (step 5) so an unauthorized caller can never use
-       a guessed idempotency key to read back someone else's receipt.
+       ``task_is_owned_by_public_principal`` predicate. A principal whose
+       ``kind`` is neither ``"user"`` nor ``"guest"`` is always
+       unauthorized -- there is no third branch that defaults to allow. A
+       malformed guest principal that populates zero or more than one
+       entity-binding field makes the ownership predicate raise
+       ``ValueError``; this function catches only that one exception type
+       from that one call and treats it as unauthorized -- the same
+       fail-closed-on-a-malformed-caller behavior the predicate itself
+       documents and ``create()`` applies to the same two cases. Runs
+       before the idempotency pre-read (step 5) so an unauthorized caller
+       can never use a guessed idempotency key to read back someone else's
+       receipt.
     4. Load the interaction row by ``(id, task_id)`` -- absent,
        ``Unavailable(interaction_missing)``. Present but its own ``kind`` /
        ``protocol_version`` columns disagree with ``envelope``'s,
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -1873,7 +1873,7 @@ def respond(
     if not isinstance(envelope.values, dict):
         return RespondValidationRejected(reason="invalid_values")
     try:
-        json.dumps(envelope.values, allow_nan=False)
+        json.dumps(envelope.values, allow_nan=False, sort_keys=True)
     except (TypeError, ValueError):
         # The same probe ``_render_request_payload`` runs on the question
         # side (see its own docstring), applied here for the two failure
@@ -1889,6 +1889,13 @@ def respond(
         # case into the ``ValueError`` caught here. Structural, like every
         # other check in step 1 -- it asks whether these values can be
         # stored at all, not whether they answer this particular question.
+        # ``sort_keys=True`` matches ``_canonical_payload``
+        # (``task_command_transport.py``), which sorts keys recursively: a
+        # dict mixing int and str keys passes an unsorted dump (int keys
+        # are silently coerced to strings -- the same silent-corruption
+        # class as the nan case) and then raises ``TypeError`` inside the
+        # replay comparison. The probe has to be a strict superset of every
+        # serializer downstream of step 1.
         return RespondValidationRejected(reason="invalid_values")
 
     from ..models.database import get_session_local
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -1822,7 +1822,10 @@ def respond(
        retires its session and reports ``OutcomeUnknown`` unconditionally,
        leaving the caller to re-check.
     10. After a successful commit, outside any transaction:
-        ``notify_task_command_dispatcher()``.
+        ``notify_task_command_dispatcher()``. Best-effort: a raise here
+        would turn a committed answer into a reported failure, so it is
+        caught, logged as a warning, and the dispatcher's idle poll
+        delivers the command instead.
 
     What this function lets escape, deliberately, and what it does not.
     The eight ``RespondOutcome`` variants cover every outcome this build
@@ -1833,9 +1836,10 @@ def respond(
 
     - Database-level failures outside the two units caught above --
       a deadlock, a lost connection, a pool checkout timeout -- raised by
-      any statement from step 2 onward. The two catches are narrow on
+      any statement from step 2 onward. The three catches are narrow on
       purpose: step 8's is ``IntegrityError`` only, step 9's covers the
-      commit only.
+      commit only, and step 10's wraps one post-commit best-effort
+      notification whose failure is logged and swallowed.
     - ``TaskCommandOwnerStateError`` and ``TaskCommandTaskMissing`` from
       ``stage_task_command``. Neither is an ``IntegrityError`` subclass, so
       step 8's catch does not see them, and both mean a precondition this
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -2028,13 +2028,15 @@ def respond(
                 )
             # The reread above already has the row in hand, so recording
             # what it found costs no extra statement. Without this line a
-            # fence miss is the one path out of this function that leaves
-            # nothing at all behind: the caller is told ``OutcomeUnknown``,
-            # and a retry against a terminated, superseded, or
-            # foreign-owned row produces the same miss and the same
-            # ``OutcomeUnknown`` every time, so no amount of retrying ever
-            # reveals which of them it was. The classification this build
-            # does not compute is exactly the set of fields logged here.
+            # fence miss is the one exit a retry cannot clarify on its
+            # own: the caller is told ``OutcomeUnknown``, and a retry
+            # against a terminated, superseded, or foreign-owned row
+            # produces the same miss and the same ``OutcomeUnknown`` every
+            # time, so no amount of retrying ever reveals which of them it
+            # was -- unlike the commit-exception door below, where a retry
+            # under the same idempotency key resolves the ambiguity by
+            # itself. The classification this build does not compute is
+            # exactly the set of fields logged here.
             import logging
 
             logging.getLogger(__name__).warning(
@@ -2128,6 +2130,16 @@ def respond(
             # against the durable graph (that reconciliation is not
             # delivered here): it retires its session and reports the
             # ambiguity, leaving the caller to re-check.
+            import logging
+
+            logging.getLogger(__name__).warning(
+                "commit failed while answering interaction %s on task %s; "
+                "the write may or may not be durable -- a retry under the "
+                "same idempotency key resolves which",
+                interaction_id,
+                task_id,
+                exc_info=True,
+            )
             session_retired = True
             _retire_respond_session_best_effort(db)
             return RespondOutcomeUnknown()
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -59,9 +59,10 @@ either); and the response-conflict counter
 Not delivered here, and named so a reader does not go looking for them in
 this module: the compatibility seam that routes the existing resume
 coordinator (``websocket.py``'s ``_handle_resume_task_unserialized``)
-through ``_active_native_row_criteria()`` ships as its own change in this
-same series and is not part of this one -- nothing in ``websocket.py``
-references this module today. ``create()``'s own call body -- the write
+through ``_active_native_row_criteria()`` shipped as its own change in this
+same series and has already merged -- ``websocket.py`` now imports
+``_active_native_row_criteria`` from this module -- but is not part of this
+change. ``create()``'s own call body -- the write
 that actually stages a row -- is not delivered here either; its own
 zero-production-caller gate
 (``tests/web/services/test_task_interaction_service_create_gate.py``)
@@ -483,13 +484,12 @@ def public_chat_identity_matches(task: "Task", principal: InteractionPrincipal) 
 
 
 # ---------------------------------------------------------------------------
-# RespondOutcome: the answer-side discriminated union. Types only in this
-# delivery -- respond() itself, and every branch that could actually
-# produce most of these variants, lands with a later change (see the module
-# docstring). Defining the full union now, rather than growing it
-# incrementally alongside respond()'s call body, is what lets the reason
-# vocabulary be pinned once and the counting below be a real guard instead
-# of a moving target.
+# RespondOutcome: the answer-side discriminated union. respond() itself,
+# and the call body that produces every one of these variants, are both
+# delivered in this module (see the module docstring). Defining the full
+# union up front, rather than growing it incrementally alongside that call
+# body, is what lets the reason vocabulary be pinned once and the counting
+# below be a real guard instead of a moving target.
 # ---------------------------------------------------------------------------
 
 
@@ -1526,7 +1526,7 @@ def _respond_command_payload(
     *, interaction_id: int, principal: InteractionPrincipal, values: Any
 ) -> dict[str, Any]:
     """The RESUME command payload ``respond()`` stages in step 8, and the
-    payload it re-derives at steps 5 and 6 to look an existing command up by
+    payload it re-derives at steps 5 and 8 to look an existing command up by
     the same shape.
 
     Carries ``responder_identity`` alongside the answer ``values`` so that
@@ -1563,7 +1563,15 @@ def _respond_receipt(
     a caller's ``principal`` -- that column, not ``principal``, is this
     table's audit-authoritative record of who answered (see ``respond()``'s
     own docstring for why ``responder_user_id`` cannot be used for the same
-    purpose).
+    purpose). This is not the only receipt this module builds: the accept
+    path builds ``RespondAccepted``'s receipt from plain locals captured in
+    the same transaction before the commit below, including
+    ``principal.identity_string()`` for ``responder_identity`` rather than
+    a read of this row's column. The two are not competing sources of
+    truth -- the answer fence this function's own precondition depends on
+    is the statement that wrote the column from that same
+    ``principal.identity_string()`` value, so the column and the local are
+    equal by the fence write's own semantics, not by coincidence.
 
     The only caller is ``respond()``'s idempotent-replay branch, whose
     precondition -- a staged RESUME command this service itself committed
@@ -1679,12 +1687,18 @@ def respond(
     the module docstring's audit-identity paragraph for why
     `responder_identity` and not this column is this table's audit source
     of truth. `actor_user_id` (written on the staged command, step 8)
-    records whose identity the resulting RESUME command executes as, which
-    is always the task's owning user for both principal kinds -- a guest's
-    turn already runs as that owner, matching the existing
-    `public_chat_access.py` precedent of dispatching guest-originated work
-    under the owner's identity. Do not "fix" this into agreement; they are
-    answers to different questions.
+    records whose identity the resulting RESUME command executes as. For a
+    guest principal, and for a non-admin `"user"` principal, that is always
+    the task's owning user: a guest's turn already runs as the entity owner
+    it is chatting through, matching the existing `public_chat_access.py`
+    precedent of dispatching guest-originated work under the owner's
+    identity, and step 3's non-admin branch requires `task.user_id ==
+    principal.user_id` to even reach here. The admin branch carries no such
+    requirement -- `principal.is_admin` authorizes on its own -- so an
+    admin's `actor_user_id` is the admin's own `principal.user_id` (which
+    can be absent, or belong to someone other than the task's owner), not
+    the task owner's. Do not "fix" this into agreement; they are answers to
+    different questions.
 
     Of the two audit-relevant columns this function fills on the interaction
     row, `responder_identity` is the one a reader can trust to stay
@@ -1730,7 +1744,7 @@ def respond(
        A dict whose contents cannot be rendered as JSON -- a ``datetime``,
        a ``set``, ``bytes``, or a ``nan``/infinite float -- is rejected
        here too, by the same ``json.dumps(..., allow_nan=False)`` probe
-       the question side runs (see ``_render_request_payload``). Without
+       the question side runs (see ``build_v1_request_payload``). Without
        it the first two would surface as a ``StatementError`` raised
        inside step 6's fence UPDATE, outside the ``RespondOutcome``
        contract entirely, and the third would be stored silently as a
@@ -1875,7 +1889,7 @@ def respond(
     try:
         json.dumps(envelope.values, allow_nan=False, sort_keys=True)
     except (TypeError, ValueError):
-        # The same probe ``_render_request_payload`` runs on the question
+        # The same probe ``build_v1_request_payload`` runs on the question
         # side (see its own docstring), applied here for the two failure
         # modes this side has. A ``values`` dict holding a ``datetime``,
         # ``set``, ``bytes``, or any other object the JSON encoder does not
@@ -2068,11 +2082,14 @@ def respond(
         # on ``ir``/``task`` the instant ``db.commit()`` returns -- reading
         # ``ir.run_id`` or ``task.state_version`` afterward would silently
         # re-issue a SELECT against a session this function is about to
-        # decide whether to keep or retire. The fence UPDATE and the CAS
-        # above are Core statements, so ``ir``/``task``'s in-memory
-        # attributes were never updated by them either; these locals are
-        # the values this call itself just wrote, not a read of what those
-        # statements left behind.
+        # decide whether to keep or retire. The fence UPDATE above is a
+        # Core statement, so ``ir``'s in-memory attributes are not synced
+        # by it; ``task``'s, by contrast, already are --
+        # ``apply_task_control_transition`` re-reads it with its own
+        # ``session.refresh(task)`` right after its atomic UPDATE. Either
+        # way these locals still have to be captured now: ``expire_on_commit``
+        # is what invalidates them the instant ``db.commit()`` returns, not
+        # whether this transaction has kept them current up to this point.
         answered_run_id = str(ir.run_id)
         answered_responder_identity = principal.identity_string()
         answered_responded_at = now
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -48,27 +48,30 @@ value object and the shared public-chat ownership predicate extracted from
 discriminated unions; the ``create()`` typed seam (validates and returns,
 does not stage a row); ``get()``/``list_active()``; the three-tier
 compatibility materialization view; the answer fence and its active-row
-predicate; ``respond()``'s own call body (validation, authorization,
-anchor resolution, the idempotency pre-read, the answer fence, the task
+predicate; ``respond()``'s own call body (validation, authorization, the
+idempotency pre-read, anchor resolution, the answer fence, the task
 control-state transition, staging the resume command, and committing --
 this delivery reports an unresolved fence miss or a failed commit as
 ``RespondOutcomeUnknown`` rather than further classifying or reconciling
-either; the response-conflict counter
-(``COUNTER_LIFECYCLE_RESPONSE_CONFLICT``); and the compatibility seam that
-routes the existing resume coordinator through this service
-(``websocket.py``'s ``_handle_resume_task_unserialized``). ``create()``'s
-own call body -- the write that actually stages a row -- is still not
-delivered here; its own zero-production-caller gate
+either); and the response-conflict counter
+(``COUNTER_LIFECYCLE_RESPONSE_CONFLICT``).
+
+Not delivered here, and named so a reader does not go looking for them in
+this module: the compatibility seam that routes the existing resume
+coordinator (``websocket.py``'s ``_handle_resume_task_unserialized``)
+through ``_active_native_row_criteria()`` ships as its own change in this
+same series and is not part of this one -- nothing in ``websocket.py``
+references this module today. ``create()``'s own call body -- the write
+that actually stages a row -- is not delivered here either; its own
+zero-production-caller gate
 (``tests/web/services/test_task_interaction_service_create_gate.py``)
-now watches ``create()`` alone. ``respond()`` left that gate not because
-it gained a caller -- the compatibility seam only reads
-``_active_native_row_criteria()`` and never calls it, and zero production
-code does today -- but because gating and callers are independent facts;
-the gate's own docstring states this and the seam's role. The gate gives
-that boundary a regression guard against import bindings, not an absolute
-one -- its docstring lists what it cannot see (dynamic access,
-alias/re-export chains, filename-stem exclusion, and code outside the
-scanned package tree).
+watches ``create()`` alone. ``respond()`` is not under that gate, and that
+is a statement about the gate, not about callers: zero production code
+calls ``respond()`` today, and the gate simply does not watch it. The gate
+gives ``create()``'s boundary a regression guard against import bindings,
+not an absolute one -- its docstring lists what it cannot see (dynamic
+access, alias/re-export chains, filename-stem exclusion, and code outside
+the scanned package tree).
 """
 
 from __future__ import annotations
@@ -619,12 +622,21 @@ RespondOutcome = (
 
 # ---------------------------------------------------------------------------
 # CreateOutcome: the create() seam's own discriminated union. Same family,
-# same style as RespondOutcome, but a separate set of classes -- the two
-# unions are not reused between each other even where a reason string
-# happens to be spelled the same way (e.g. "not_task_principal" appears in
-# both vocabularies below because both seams reuse the shared ownership
-# predicate's verdict, not because the two outcome types share a base
-# class).
+# same style as RespondOutcome, but two different mechanisms, and the
+# asymmetry is intentional rather than unfinished cleanup. Respond's
+# outcomes declare each reason as a ``Literal`` on the dataclass field, so
+# the type is the vocabulary. Create's still carry ``reason: str`` beside
+# the runtime dictionaries below, which record something a ``Literal``
+# cannot: which reasons are *producible* in this delivery as opposed to
+# merely declared, a distinction that exists only because create()'s call
+# body is not delivered yet. The two vocabularies overlap in nine strings
+# and are deliberately not shared -- ``"not_task_principal"`` appears in
+# both because both seams reuse the shared ownership predicate's verdict,
+# not because the two outcome types share a base class, and a change to
+# one side's word list must never silently move the other's. When
+# create()'s call body lands and the producible/declared distinction
+# disappears with it, that side converts to the Literal mechanism and this
+# note goes away.
 # ---------------------------------------------------------------------------
 
 
@@ -1389,7 +1401,15 @@ def _answer_fence_task_predicate(principal: InteractionPrincipal) -> list[Any]:
     ``respond()``'s own docstring for the two backends' different
     serialization points). Enforcing ownership at the write point rather
     than only in ``respond()``'s step 3 is what makes the two backends
-    agree on what a successful answer proves.
+    agree on the terms this statement actually carries -- which is not
+    every term step 3 evaluated. Three are re-asserted here: the task is
+    still ``WAITING_FOR_USER``, ``Task.user_id`` is still the principal's,
+    and, for a guest, the ``guest_id`` in ``agent_config`` still matches.
+    The rest of the guest conjunction -- ``auth_mode``, the entity binding,
+    and the channel binding -- is evaluated once in Python at step 3 and
+    not re-checked in SQL. A successful answer therefore proves the three
+    terms above held at write time and that the other three held at read
+    time, not that all six held at write time.
 
     ``Task.status`` is compared through ``TaskStatusPredicate.eq``, never a
     raw string literal beside the column -- this module's own convention
@@ -1641,16 +1661,28 @@ def respond(
     hold.
 
     `_handle_resume_task_unserialized` re-issues the same
-    `RESUME_REQUESTED` transition when it applies
-    the command this call staged. `apply_task_control_transition` has no
-    legality table, so that second transition succeeds and bumps
-    `state_version` again. Post-commit verification must therefore be
-    monotone in `state_version` and must not compare `control_state`. One
-    concrete consequence, recorded here because a future reader who treats
-    `state_version` as an operation counter will be surprised by it:
-    answering one interaction bumps `state_version` twice -- once in step 7
-    below, once when the coordinator re-applies the transition -- so a
-    consumer that expects "one answer, one version bump" will observe +2.
+    `RESUME_REQUESTED` transition when it applies the command this call
+    staged (`websocket.py`, reached from `_execute_durable_task_command`).
+    `apply_task_control_transition` has no legality table, so that second
+    transition succeeds and bumps `state_version` again. Post-commit
+    verification must therefore be monotone in `state_version` and must not
+    compare `control_state`. One concrete consequence, recorded here
+    because a future reader who treats `state_version` as an operation
+    counter will be surprised by it: answering one interaction bumps
+    `state_version` twice -- once in step 7 below, once when the
+    coordinator re-applies the transition -- so a consumer that expects
+    "one answer, one version bump" will observe +2.
+
+    That second bump is not reachable today: nothing in production calls
+    `respond()`, so no RESUME command carrying this function's payload is
+    ever staged or executed. It becomes reachable with the change that
+    gives `respond()` its first production caller, and deciding which of
+    the two writers owns the transition belongs to that change, not this
+    one. Making the transition single-writer here is not an option this
+    change can take on its own: `apply_task_control_transition` is shared
+    by eight direct call sites across five modules plus five more through
+    `transition_task_control_state_sync`, and adding a legality table to it
+    is a change to the task state machine, not to this service.
 
     `responder_user_id` and `actor_user_id` answer two different questions
     and are allowed to disagree. `responder_user_id` (written on the
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -65,11 +65,11 @@ references this module today. ``create()``'s own call body -- the write
 that actually stages a row -- is not delivered here either; its own
 zero-production-caller gate
 (``tests/web/services/test_task_interaction_service_create_gate.py``)
-watches ``create()`` alone. ``respond()`` is not under that gate, and that
-is a statement about the gate, not about callers: zero production code
-calls ``respond()`` today, and the gate simply does not watch it. The gate
-gives ``create()``'s boundary a regression guard against import bindings,
-not an absolute one -- its docstring lists what it cannot see (dynamic
+watches both ``create()`` and ``respond()``: zero production code calls
+either name today, and the change that wires a caller is the change that
+takes the corresponding name out of the gated set. The gate gives
+``create()``'s boundary a regression guard against import bindings, not
+an absolute one -- its docstring lists what it cannot see (dynamic
 access, alias/re-export chains, filename-stem exclusion, and code outside
 the scanned package tree).
 """
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -10,9 +10,7 @@ both live on the ask side, and both require the caller to already hold the
 transaction they run inside. This module's ``respond()`` is the opposite
 shape on every one of those points: it owns its own session end to end
 (opens it, commits or rolls it back, retires it) and never nests inside a
-caller's transaction -- the one savepoint it opens
-(``db.begin_nested()``, around staging its own resume command) is its
-own, not a caller's. ``create()``, by contrast, takes a caller-owned
+caller's transaction. ``create()``, by contrast, takes a caller-owned
 session and only reads through it, never opening or committing one of its
 own. Putting ``respond()`` in the same file as the staging primitive would
 make that staging module's merge-reason docstring false. What this module
@@ -52,8 +50,10 @@ does not stage a row); ``get()``/``list_active()``; the three-tier
 compatibility materialization view; the answer fence and its active-row
 predicate; ``respond()``'s own call body (validation, authorization,
 anchor resolution, the idempotency pre-read, the answer fence, the task
-control-state transition, staging the resume command, and committing or
-reconciling an ambiguous acknowledgment); the response-conflict counter
+control-state transition, staging the resume command, and committing --
+this delivery reports an unresolved fence miss or a failed commit as
+``RespondOutcomeUnknown`` rather than further classifying or reconciling
+either; the response-conflict counter
 (``COUNTER_LIFECYCLE_RESPONSE_CONFLICT``); and the compatibility seam that
 routes the existing resume coordinator through this service
 (``websocket.py``'s ``_handle_resume_task_unserialized``). ``create()``'s
@@ -74,7 +74,6 @@ scanned package tree).
 from __future__ import annotations
 
 import json
-import time
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Literal, cast
@@ -103,12 +102,9 @@ from .ops_signals import (
     register_degradation,
 )
 from .task_command_transport import (
-    TaskCommandConflictKind,
     TaskCommandKind,
-    _canonical_payload,
     _matches_existing,
     _normalize_command_id,
-    classify_task_command_conflict,
     notify_task_command_dispatcher,
     stage_task_command,
 )
@@ -542,15 +538,8 @@ RespondUnauthorizedReason = Literal["not_task_principal"]
 RespondUnavailableReason = Literal[
     "task_missing", "interaction_missing", "checkpoint_unavailable"
 ]
-RespondConflictReason = Literal["already_answered", "idempotency_key_reused"]
-RespondStaleReason = Literal[
-    "expired",
-    "run_superseded",
-    "answered_via_chat",
-    "run_ended",
-    "foreign_run",
-    "anchor_dangling",
-]
+RespondConflictReason = Literal["idempotency_key_reused"]
+RespondStaleReason = Literal["anchor_dangling"]
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,619 @@ def materialize_compatibility_view(
...
+    compiles (``->>`` on PostgreSQL, ``json_extract`` on SQLite) -- both are
+    exercised by ``test_task_interaction_service_postgresql.py`` and this
+    module's SQLite unit tests, deliberately, rather than assumed
+    equivalent from one compiled form.
+    """
+
+    terms: list[Any] = [
+        TaskStatusPredicate.eq(TaskStatus.WAITING_FOR_USER),
+        Task.user_id == principal.user_id,
+    ]
```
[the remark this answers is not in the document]
AlexLiu190625: The unconditional `Task.user_id == principal.user_id` term is intentional: for a guest principal it is the only ownership constraint in the whole path, because the Python-side ownership predicate deliberately does not read `principal.user_id` (see `task_is_owned_by_public_principal`'s docstring), and a test pins that behavior (`test_respond_rejects_a_guest_whose_bindings_match_but_principal_user_id_does_not`).

On the null concern: `tasks.user_id` is `nullable=False`, so a `None` principal id compiles to `tasks.user_id IS NULL` and matches zero rows — the fence fails closed rather than matching unintended rows. Adding a `principal.user_id is not None` guard would drop the ownership term exactly in the case it is meant to cover, weakening the fence.

The admin observation is accurate as far as the code goes: an admin acting on another user's task passes the Python authorization step and then misses the fence, reporting this build's catch-all unknown outcome. That is fail-closed with no partial writes, and every fence miss collapses onto that outcome by design in this build. There is also no production construction of an admin principal today and no production caller of this function, so whether an admin may answer on someone else's behalf is a policy question for the change that wires the first caller.

Documented in the follow-up commit now on this branch: the fence predicate's docstring now states the term requires the owner in person on both backends and that answering on another user's behalf is out of scope here, and the step-6 line that described an ownership miss as reachable only on SQLite is corrected. The predicate itself stays as is.

tests/web/services/test_task_interaction_service.py
```diff
@@ -1204,3 +1333,1124 @@ def test_module_issues_zero_sa_text_calls() -> None:
...
+    owner_id, task_id = _waiting_task(
+        _respond_db,
+        agent_config={
+            "auth_mode": "widget",
+            "guest_id": "guest-1",
+            "widget_workforce_id": 10,
+        },
+    )
+    interaction_id = _active_row_ready_for_respond(_respond_db, task_id=task_id)
+    wrong_user_id = make_user(_respond_db())
```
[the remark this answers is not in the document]
AlexLiu190625: Applied. The setup session is now wrapped in try/finally and closed, matching the pattern every other session in this file already uses. Worth noting the scope of the problem: the shared helper commits and then refreshes, so the abandoned session held one pooled connection in an open read transaction until garbage collection — a hygiene issue rather than a live failure, since the engine fixture is function-scoped and a full instrumented run showed zero checked-out connections at every teardown. Fixed for consistency with the file's own convention.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -267,34 +299,537 @@ def test_active_row_predicate_three_way_tiering(db_session) -> None:
...
+
+
+def test_concurrent_ownership_change_is_blocked_until_respond_commits(
+    _respond_pg, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    import time
+
+    owner_id, task_id = _pg_waiting_task(_respond_pg)
+    interaction_id = _pg_active_row(_respond_pg, task_id=task_id)
+    intruder_id = make_user(_respond_pg())
```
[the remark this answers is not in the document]
AlexLiu190625: Applied. Same shape as the sibling comment on the SQLite file: the setup session is now closed in a try/finally, consistent with the nine other call sites in this file that already do so.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -267,34 +299,537 @@ def test_active_row_predicate_three_way_tiering(db_session) -> None:
...
+    import time
+
+    import psycopg2.extras
+
+    user_id, task_id = _pg_waiting_task(_respond_pg)
+    interaction_id = _pg_active_row(_respond_pg, task_id=task_id)
+    principal = _pg_owning_principal(user_id)
+    replacement_trace_event_id = _make_trace_event(_respond_pg(), task_id=task_id)
+
+    engine = _respond_pg().get_bind()
```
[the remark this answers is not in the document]
AlexLiu190625: Partially applied.

The trace-event setup session is now closed in a try/finally — that one did hold a pooled connection (the helper commits then refreshes), and closing it matches the file's convention.

The `db.connection()` suggestion is declined, on measured grounds. `Session.get_bind()` returns the Engine and checks out no connection at all: on SQLAlchemy 2.0.47, `pool.checkedout()` stays at 0 after `get_bind()` and rises to 1 after `Session.connection()` — so the suggested change adds a pool checkout rather than removing one. It also cannot work here: this call site exists solely to reach `engine.raw_connection()`, which opens the second, genuinely independent DBAPI connection this concurrency test is built on. `sqlalchemy.engine.Connection` has no `raw_connection` attribute, so the suggestion raises `AttributeError` as written; and reaching the DBAPI handle through the session instead would hand the worker thread the session's own connection, collapsing the two-connection lock contention this deadlock regression exists to pin — the test would still pass while proving nothing.

Reusing a single session across both lines is declined for the same reason: the two lines need different things (one writes a row, one only needs the bind), and merging them reintroduces a checkout where none is required. Instead, the test now requests the module's existing `engine` fixture directly — the same Engine object the session factory is bound to — and the throwaway-session `get_bind()` line is gone.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -267,34 +299,537 @@ def test_active_row_predicate_three_way_tiering(db_session) -> None:
...
+# ---------------------------------------------------------------------------
+
+
+def test_respond_vs_concurrent_purge_does_not_deadlock(_respond_pg) -> None:
+    import time
+
+    user_id, task_id = _pg_waiting_task(_respond_pg)
+    interaction_id = _pg_active_row(_respond_pg, task_id=task_id)
+    principal = _pg_owning_principal(user_id)
+    engine = _respond_pg().get_bind()
```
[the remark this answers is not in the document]
AlexLiu190625: There was no session leak at this line: `Session.get_bind()` returns the Engine without checking out a connection (measured — `pool.checkedout()` stays at 0), so the throwaway session held nothing. The `db.connection()` substitution is declined for the reasons detailed on the sibling thread above: it is the variant that adds a pool checkout, `Connection` has no `raw_connection` attribute, and routing the purge thread through the session's own connection would collapse the two-connection contention this deadlock regression exists to exercise. The line is fixed the other way instead: the test now takes the module's `engine` fixture directly and the throwaway session is gone.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,633 @@ def materialize_compatibility_view(
...
+    purpose).
+    """
+
+    return InteractionResponseReceipt(
+        interaction_id=int(interaction.id),
+        task_id=int(interaction.task_id),
+        run_id=str(interaction.run_id),
+        status=str(interaction.status),
+        responded_at=cast("datetime", interaction.responded_at),
+        responder_identity=str(interaction.responder_identity),
```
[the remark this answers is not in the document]
AlexLiu190625: Thanks — the nullability is real, but this call site is not reachable with a NULL.

`_respond_receipt` has exactly one caller: the idempotent-replay branch. Reaching it requires a committed `task_execution_commands` row whose canonical payload is byte-equal to the three-key shape `{"interaction_id", "responder_identity", "values"}`. The only writer producing that shape is `respond()` itself, which stages the command in the same transaction as the answer fence UPDATE, only after the fence matched one row; every failure path in between (fence miss, staging `IntegrityError`, commit failure) leaves neither the command nor the answer durable. The other RESUME producer, the legacy WebSocket resume handler, always carries a `type` key in its payload and can never canonicalize to the same shape. Three CHECK constraints then close it at the schema level: `(status = 'answered') = (responded_at IS NOT NULL)`, `(responded_at IS NULL) = (responder_identity IS NULL)`, and `responder_identity <> ''` — an answered row cannot carry a NULL or empty identity, and the table was created with all three, so there is no pre-constraint data either.

The `""` default specifically is the one value I don't want here: the non-empty CHECK exists precisely to keep `""` out of this column (it satisfies `IS NOT NULL` while naming nobody), so writing `""` into the receipt would reproduce on the read side exactly what the schema forbids on the write side — and `""` is falsy, so a consumer testing `if receipt.responder_identity` would take the "no responder" branch on a receipt whose `status` says `"answered"`. That fails more quietly than `'None'` would. The suggestion also covers only one of two paired symptoms: if the identity were NULL, `responded_at` would be NULL too by the biconditional, and the `cast(...)` on the line above is a runtime no-op.

What I do agree with is that "never NULL here" rested on an invariant spanning two modules with nothing pinning it. The follow-up commit now on this branch makes it explicit instead of silent: the receipt builder raises if an interaction that matched a staged RESUME command carries no answer, covering `responded_at` as well, in the same style this function already uses for its other impossible states. A test pins the guard (verified by removing the guard, which turns the test red).

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,633 @@ def materialize_compatibility_view(
...
+            authorized = principal.is_admin or (
+                principal.user_id is not None and task.user_id == principal.user_id
+            )
+        elif principal.kind == "guest":
+            try:
+                authorized = task_is_owned_by_public_principal(task, principal)
+            except ValueError:
+                authorized = False
+        else:
+            authorized = False
```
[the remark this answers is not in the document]
AlexLiu190625: Keeping both branches as they are: this block is a verbatim copy of the one already on `main` in `create()`, and that shape is a documented contract rather than an incidental default.

`create()`'s docstring states it explicitly: a principal whose `kind` is neither `"user"` nor `"guest"` is always unauthorized, with no third branch that defaults to allow; and a malformed principal that populates zero or more than one entity-binding field makes the ownership predicate raise `ValueError`, which that function catches — only that exception, only from that call — and treats as unauthorized. Three tests on `main` pin exactly these behaviors, two of them explicitly asserting that no exception escapes.

On the `ValueError`: the predicate's own docstring defines its raise as "fail closed on a malformed caller, not fail open on a guess". Catching it and denying honors that contract rather than hiding it. The information is not lost globally either — the predicate's production callers deliberately don't catch it; they guard the malformed inputs out before constructing the principal.

On `else: raise`: `principal.kind` is a server-side literal at every construction site, so neither branch is reachable in production today, which makes this a style question rather than a correctness one — and on an authorization path the default should be one more denial, not one fewer. `else: authorized = False` guarantees any value outside the allowlist is refused, with no dependency on how a caller's error handling maps an unexpected exception. The decisive point is symmetry: these are the two authorization seams of one module over one shared predicate, and changing only this one would leave the same malformed input producing a denial through one entry point and a server error through the other — a distinguishable, probeable difference on an authorization path. If these cases should be surfaced rather than silent, the right change is observability (a log or counter) at both seams in one follow-up, keeping the denial semantics intact; that is out of scope here.

One thing this round did surface: `respond()`'s step-3 docstring described only the two happy branches and said nothing about the unknown-kind fallback or the `ValueError` catch, unlike `create()`'s. The follow-up commit brings the two into line so the documented contract matches the code at both seams.

rogercloud (changes requested):
## Re-review: `feat(interactions): add the interaction lifecycle service's answer-side entry point`

### Summary

This PR adds `respond()` — the answer-side entry point of the interaction lifecycle service — which atomically (a) flips the unique active native-interaction row to `answered` only while it is still active, still owned by the caller, and its task is still waiting; (b) stages the RESUME command; (c) returns a typed `RespondOutcome`/receipt with content-derived idempotent replay and conservative zero-residue ambiguity handling. It is the core of a four-part series; the resume-coordinator seam ships separately, and two follow-ups target recovery precision (fence-miss classification, commit-ack reconciliation). There are zero production callers of `respond()`. The design verdict is **acceptable-with-reservations**: the direction is right, the conservative failure policy is sound, and the four-part split is coherent — but the structural issues below block merge.

### Update summary

Since the last substantive review (~`0430d1bb`) there is exactly one commit: `b5d37796` "refactor(interactions): assert the replay receipt's answered-row precondition" — it adds an answered-row precondition guard to `_respond_receipt` (raises `RuntimeError` when `responded_at`/`responder_identity` are `None`, responding to P6), pins it with a test, and aligns `respond()`'s step-3 docstring with `create()`'s contract (responding to P7). All 5 blocking findings from this pass are new; nothing previously raised was left unfixed without a documented waiver.

### Prior findings checklist

- **P1 — admin fence predicate (thread 1397): FIXED.** The author's point-by-point explanation verified: guest term is intentional, `user_id` is `nullable=False` so a `None` binding matches zero rows and fails closed, and the admin case fails closed with no production construction. Doc correction landed in `d9a928d6`.
- **P2 — session leak at test line 1769: FIXED** (`try/finally` close landed in `0430d1bb`).
- **P3 — session leak at PG test line 489: FIXED** (same commit; no unclosed session remains in the file).
- **P4 — session leak at PG test line 660 + `db.connection()` suggestion: FIXED.** The close landed; the `db.connection()` suggestion was correctly rejected (`get_bind` does not check out a connection, `Connection` has no `raw_connection` so it would `AttributeError`, and it would collapse the two-connection contention the test exercises). The engine-fixture direct fetch landed instead.
- **P5 — PG test line 775 session + connection suggestion: FIXED.** That line had no leak (`get_bind` does not check out); the engine-fixture replacement landed and the throwaway session was deleted.
- **P6 — `str(None)` at line 1574: FIXED.** The author's unreachability argument verified point-by-point (replay branch is the sole caller; canonical payload has a three-key shape; legacy RESUME payloads carry a `type` key and never match; the CHECK constraint seals the shape; `""` is banned by the nonempty CHECK and is falsy anyway). `b5d37796` additionally hardened `_respond_receipt` with a guard that raises when the answered-row precondition fails, pinned by a test that goes red when the guard is removed.
- **P7 — unknown-kind silent unauthorized + `ValueError` catch at line 1822: NOT FIXED but WAIVED.** The code behavior is unchanged (the author explicitly declined to change it), and that is acceptable: the block is verbatim identical to `create()`'s — a documented contract on main; three pin tests on main assert no exception escapes; the predicate docstring defines the intent as "fail closed on malformed caller"; and both authorization seams share one predicate and one contract, so changing only one side would create a detectable divergence. The step-3 docstring alignment in `b5d37796` landed. Waived — documented here so the disposition is explicit.

### New findings this pass (by severity)

#### Major (blocking)

- **[F1] `src/xagent/web/services/task_interaction_service.py:1957` — staging result is consumed incompletely; a lost race surfaces as a false `RespondAccepted`.** `respond()` reads only `staged.staged_db_id` and ignores `created`/`payload_matches`. If a concurrent non-respond writer commits the same `(task_id, command_id)` between the idempotent pre-read and staging, `stage_task_command` does not raise `IntegrityError` — it returns the existing row — and `respond()` still commits the fence + CAS and returns `RespondAccepted` carrying another writer's command id, with the correct RESUME never staged. That is exactly the scenario the function's own non-raced path would report as `RespondConflict(idempotency_key_reused)`, silently downgraded to a false Accepted, breaking its own raced-duplicate → `OutcomeUnknown` promise. Repo convention is clear: `enqueue_task_command` commits only when `staged.created`, and every production caller checks `payload_matches` (`api/a2a.py:1394`, `websocket.py:4990/7420/7670`). **Suggestion:** after staging, require `staged.created and staged.payload_matches`; if either is `False`, `db.rollback()` and return `RespondOutcomeUnknown()`.

- **[F2] `src/xagent/web/services/task_interaction_service.py:1832` — anchor resolution precedes the idempotent pre-read; delayed retries after phantom commit failures mis-classify as `Stale`.** An answered row's checkpoint anchor is not protected by the retention pruner (it guards ACTIVE rows only, `trace_handlers.py:963-969`, and `ON DELETE SET NULL` is legal on answered rows). A retry that arrives after the pruner cleared the anchor hits step 4.5 and returns `Stale(anchor_dangling)` instead of `Replayed` — falsifying the PR text's claim that "retry after a phantom failure is caught by the replay check" (that only converges for immediate retries while the anchor is still alive). **Suggestion:** move the idempotent pre-read before anchor resolution (or make the replay branch independent of anchor liveness); at minimum document the precondition.

- **[F3] `src/xagent/web/services/task_interaction_service.py:1917` — `WAITING_FOR_USER → RESUME_REQUESTED` has two writers and no legality table.** `respond()` applies the transition here, and the coordinator re-applies it when executing the staged command (`websocket.py:7810-7814`); `apply_task_control_transition` has no legality check, so one answer double-bumps `state_version`. The docstring at 1631-1641 asks consumers to "keep monotonic" instead of fixing the state machine. Separately, `respond()` deliberately does not flip `Task.status`, so between the answer commit and coordinator execution the task reads `status=WAITING_FOR_USER` + `control_state=RESUME_REQUESTED` with no active row — and the status-gated path at `websocket.py:7784` still treats the task as resumable. **Suggestion:** single-writer the transition (respond does not transition; the coordinator transitions once during execution), or add a legality check to `apply_task_control_transition` so repeat application is a no-op.

- **[F5] `src/xagent/web/services/task_interaction_service.py:1791` — answer payload validation is `isinstance(dict)` only; malformed answers are stored and staged.** No per-field validation against the `InteractionArg` definition is performed (the docstring at 1699-1702 states it is not implemented and not in either named follow-up), so a malformed-but-dict answer is stored and staged into the RESUME payload for the execution side to consume — while `create()` validates via `parse_v1_request_payload` (pydantic) plus a JSON-serializability probe. Worse, a dict containing non-JSON-serializable objects (`datetime`/`set`/`bytes`) raises an uncaught `TypeError`/`StatementError` during fence-bind serialization (verified against SQLAlchemy 2.0.51), escaping the 8-variant `RespondOutcome` contract and contradicting the docstring's "stored as submitted". **Suggestion:** after row load, validate `values` against the `InteractionArg` definition (presence/type/option membership), reusing `create()`'s shape; at minimum, name the closing gap in a follow-up change.

- **[F6] `src/xagent/web/services/task_interaction_service.py:59` — the module docstring claims a compatibility seam that does not exist.** The docstring says the module "now delivers ... the compatibility seam that routes the existing resume coordinator through this service", and the gate docstring (`tests/web/services/test_task_interaction_service_create_gate.py:7-12`) claims "That change has landed" — but the seam exists nowhere: not in this PR, not in HEAD, not in the merge-base (`websocket.py` has no `task_interaction_service`/`_active_native_row_criteria` references and this PR does not touch `websocket.py`). The actual legacy resume path retires the active row via `close_legacy_resume_interaction_sync` and continues — it neither reads the criteria nor rejects. The zero-production-caller gate narrowed to exclude `respond` is premised on this nonexistent seam, and the base-version gate's retirement condition ("routes the existing resume coordinator through this module's compatibility seam") never occurred. **Suggestion:** rewrite both docstrings to state the facts (seam ships separately; `respond` has zero production callers), or deliver the seam before/with the docstring.

#### Minor (non-blocking)

- **[F4] `src/xagent/web/services/task_interaction_service.py:1813` — step-3 authorization has a structural dead branch; typed rejection deferred.** The two-layer construction is contradictory: step 3 admits admin-for-others, but the fence's `Task.user_id == principal.user_id` term never matches, making the admin branch dead code whose only observable effect is a mislabeled `OutcomeUnknown`; the guest binding matches yet the `user_id` mismatch case is the same (pinned by tests). Discussed in threads 3781826034/3781913641; the author's point-by-point explanation was verified (fail-closed, no production admin construction, docstring records the deferral) and `d9a928d6` landed the doc fix. The remaining gap — a step-3 typed rejection (`RespondUnauthorized`) — is deferred as a "policy issue for when wiring changes". **Suggestion:** land it together with that wiring change; do not block this PR on it separately.

- **[F8] `src/xagent/web/services/task_interaction_service.py:870` — `RespondOutcomeUnknown` conflates 8 conditions; the "Unknown is safe because the caller retries" claim over-reaches.** Retries converge only for same-key already-answered/commit-ack scenarios; retries against terminated/superseded/ownership-miss rows produce the same fence miss and a permanent `Unknown` with no logs. **Suggestion:** soften the convergence claim to name the converging scenarios, and log the re-read row state on the fence-miss path.

- **[F9] `src/xagent/web/services/task_interaction_service.py:1606` — the typed-outcome contract is incomplete.** The documented `RuntimeError` invariant and `StaleTaskRunError` are documented, but `DBAPIError` outside the two catch units (deadlock/connection loss), JSON-serialization `TypeError`, and the notify `RuntimeError` race all escape raw. **Suggestion:** document the intentional escape surface in the signature docstring, and decide per unit whether operational errors degrade to `OutcomeUnknown`.

- **[F10] `src/xagent/web/services/task_interaction_service.py:1398` — the fence guest branch re-checks only `guest_id` + `user_id`; the step-3 conjunction is not re-asserted in SQL.** The `entity_binding`/`auth_mode`/`channel_id` terms evaluated at step 3 are absent from the SQL re-check; unreachable on PG (row lock), while on SQLite without a lock clause a concurrent `agent_config` change yields either a busy error or no re-evaluation — making the docstring claim that "both backends agree on what a successful answer proves" too strong. The fail-closed direction is intact. **Suggestion:** soften the predicate docstring, or add the `entity_binding`/`auth_mode` terms to the guest branch.

- **[F11] `src/xagent/web/services/task_interaction_service.py:1971` — `notify_task_command_dispatcher()` runs unguarded after a successful commit.** The `loop.is_closed()` TOCTOU can raise `RuntimeError` and make `respond()` escape with an exception after the answer was already committed; a retry then hits the replay branch, returns a receipt, but never re-notifies. The idle polling fallback (`DISPATCHER_IDLE_SECONDS=0.5`) is documented as self-healing. **Suggestion:** wrap the notify in a best-effort `try/except`.

- **[F12] `tests/web/services/test_task_interaction_service.py:37` — the matrix meta-test overstates coverage and its arithmetic is stale.** The matrix claims A1 and A2 both produce `(Unauthorized, not_task_principal)`, but no respond-layer test covers A2 (guest rejected at step 3); the guest `ValueError → unauthorized` and unknown-kind branches also lack respond-layer coverage. The enumeration says 18 cells vs the claimed 16, and the numbers are stale ("20 pairs" is actually 14; the "24-cell/five not_task_principal" description is outdated). The mapping meta-test cannot detect the A2 gap from the pair set (the module docstring admits as much). **Suggestion:** add the A2 test, fix the arithmetic, refresh the numbers.

- **[F13] `tests/web/services/test_task_interaction_service_postgresql.py:541` — the `racer_wait_s >= 0.8` assertion depends on wall-clock luck.** It relies only on a 0.1s stagger plus the holder's injected `sleep(1.0)`; CI scheduling delays can flake in both directions (holder late to acquire the lock, or racer delayed). **Suggestion:** use `threading.Event` signaling — holder sets after acquiring the lock, racer waits on the signal before issuing its UPDATE.

- **[F14] `src/xagent/web/services/task_interaction_service.py:641` — two outcome-vocabulary mechanisms coexist.** `Respond` uses a `Literal`-per-outcome union while `Create` still uses `reason: str` plus a runtime dict; the 9 overlapping strings are kept apart by comment disclaimers. Note from the verifier: the author's dual-delivery rationale is documented and internally consistent, `create()` is the unfinished seam, and the dict carries delivery-scope facts the type system cannot express — frameable at merge as "intentional asymmetry worth acknowledging in a module comment".

- **[F16] `src/xagent/web/services/task_interaction_service.py:1582` — `_retire_respond_session_best_effort` is the third copy of the same-shaped best-effort session-retirement idiom.** `task_orchestrator.py:1221` and `websocket.py:4685` have the identical shape; only the log text, logger acquisition, and `task_id` parameter differ, and the docstring argues a naming difference, not a behavioral one. The repo already has precedent for cross-module import of such private helpers (`tasks.py:89`). **Suggestion:** extract a shared helper (context label + optional `task_id`) used by all three sites.

- **[F17] `tests/web/services/test_task_interaction_service_create_gate.py:286` — the gate stopped watching `respond` in the very PR that makes it callable.** The new test pins "production call to respond is not flagged", and the PR text claims "the gate is the regression guard for that fact" — which contradicts the code: the gate no longer watches `respond`, and restoring `respond` to the gated set produces zero violations (verified, still green). **Suggestion:** restore `respond` to the gated set, flip the test into a positive control, and align the PR text.

#### Nit (optional)

- **[F15] `src/xagent/web/services/task_interaction_service.py:106` — cross-module private imports (`_KIND_VOCABULARY`/`_normalize_command_id`/`_matches_existing`).** 2 of 3 are a base-era pattern, repo precedent is broad (e.g. `lancedb.py:18`), the module docstring records the decision, and a rename would fail loudly with `ImportError`. List as an optional nit or omit; no action required.

- `delete:` `_rowcount()` (L1341) — single caller at line 1882; inline `int(getattr(fence_result, "rowcount", 0) or 0)` and delete its direct unit tests. (~-15 net lines)
- `shrink:` `_retire_respond_session_best_effort` (L1582) — same-shape copy (see F16); import the existing helper instead. (~-12 net lines)
- `delete:` `_waiting_task` `run_id`/`channel_id`/`agent_id` params (L1360) — 18 call sites never override; `run_id` is hardcoded `"run-a"` in the body. (~-12 net lines)
- `delete:` `_active_row_ready_for_respond` `run_id`/`idempotency_key` params and the dead branch (L1391) — 14 call sites never pass them. (~-10 net lines)
- `delete:` `_answered_row_with_valid_anchor` `idempotency_key` param (L1427) — 2 call sites rely on the default. (~-6 net lines)
- `delete:` `_stage_matching_command` `status` param (L1464) — 4 call sites never override; `"completed"` is hardcoded in the body. (~-4 net lines)
- `shrink:` `_owning_principal_for_fence` + `_owning_user_principal` (L1206) — semantically identical, formatting differs; plus a third `_owning_principal` at 328-335; merge into one. (~-25 net lines)
- `delete:` `_pg_waiting_task` `run_id`/`state_version` params (L366) — 5 call sites never pass; keep the assignment lines hardcoded. (~-6 net lines)
- `delete:` `_pg_active_row` `run_id` param (L387) — 5 call sites never pass. (~-4 net lines)
- `shrink:` unanswered-fixture `run_id`/`status` fields (L1935-1936, updated lens) — `_respond_receipt` never reads them (the guard raises first); delete both lines. (~-2 net lines)

Total estimated ~-96 net lines with no behavior change.

**Blocking: yes — recommended event: REQUEST_CHANGES.**

Blocking findings (all `[new]` this pass):

- `src/xagent/web/services/task_interaction_service.py:1957` — [major] staging result consumed partially (`created`/`payload_matches` ignored); a concurrent same-key writer turns a would-be `RespondConflict` into a false `RespondAccepted` with the correct RESUME never staged. [new]
- `src/xagent/web/services/task_interaction_service.py:1832` — [major] anchor resolution precedes the idempotent pre-read; delayed retries after phantom commit failures return `Stale(anchor_dangling)` instead of `Replayed`, falsifying the PR's replay-safety claim. [new]
- `src/xagent/web/services/task_interaction_service.py:1917` — [major] `WAITING_FOR_USER → RESUME_REQUESTED` double-written by `respond()` and the coordinator with no legality table (double `state_version` bump); interim task state is misread as resumable by the status-gated path. [new]
- `src/xagent/web/services/task_interaction_service.py:1791` — [major] answer payload validated as dict-only; malformed and non-JSON-serializable answers are stored/staged or escape as uncaught exceptions, violating the typed-outcome contract. [new]
- `src/xagent/web/services/task_interaction_service.py:59` — [major] module and gate docstrings claim a landed compatibility seam that exists in no commit of this PR or its base; the gate's narrowing narrative rests on a change that never shipped. [new]

Recommended path: address F1-F3 and F5 with the concrete suggestions above (all small, localized changes), rewrite the F6 docstrings to state current facts (or land the seam), then re-request review. The minor findings (F4, F8-F14, F16-F17) and the simplification list can be folded into the same pass or follow-ups; F4/F15 explicitly do not need to block.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,660 @@ def materialize_compatibility_view(
...
+            # means (a fine-grained classification via
+            # ``classify_task_command_conflict``, distinguishing a genuine
+            # replay from a real conflict, is not delivered here): the
+            # whole transaction rolls back, undoing this call's own fence
+            # UPDATE and CAS along with it, and this call reports the
+            # ambiguity rather than guessing which case it was.
+            db.rollback()
+            return RespondOutcomeUnknown()
+
+        command_db_id = staged.staged_db_id
```
rogercloud, line 2136: respond() consumes only staged.staged_db_id and ignores staged.created/payload_matches: if a concurrent non-respond writer commits the same (task_id, command_id) between the idempotent pre-read and staging, stage_task_command returns the existing row (no IntegrityError), and respond() still commits the fence + CAS and returns RespondAccepted with the other writer's command id — the scenario the non-raced path would report as RespondConflict(idempotency_key_reused) becomes a false Accepted and the correct RESUME is never staged. Every production caller checks payload_matches (e.g. api/a2a.py:1394, websocket.py:4990). Suggest requiring staged.created and staged.payload_matches after staging and, if either is False, db.rollback() + return RespondOutcomeUnknown().

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,660 @@ def materialize_compatibility_view(
...
+        if not authorized:
+            return RespondUnauthorized(reason="not_task_principal")
+
+        ir = get(db, task_id=task_id, interaction_id=interaction_id)
+        if ir is None:
+            return RespondUnavailable(reason="interaction_missing")
+        if ir.kind != envelope.kind or ir.protocol_version != envelope.protocol_version:
+            return RespondValidationRejected(reason="kind_version_mismatch")
+
+        unresolved = _resolve_read_direction_anchor(db, ir)
```
rogercloud: Anchor resolution (step 4.5) runs before the idempotent pre-read (step 5), but an answered row's checkpoint anchor is not protected by the retention pruner (ACTIVE rows only, trace_handlers.py:963-969; ON DELETE SET NULL is legal on answered rows). A delayed retry after a phantom commit failure hits step 4.5 and returns Stale(anchor_dangling) instead of Replayed, so the PR's "retry after a phantom failure is caught by the replay check" only holds for immediate retries while the anchor survives. Suggest moving the idempotent pre-read ahead of anchor resolution (or making the replay branch independent of anchor liveness); at minimum document the precondition.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,660 @@ def materialize_compatibility_view(
...
+            return RespondOutcomeUnknown()
+
+        # No expected_run_id/expected_state_version to pass: this function
+        # takes no caller-supplied optimistic-concurrency token, so
+        # apply_task_control_transition's own staleness checks never fire.
+        # StaleTaskRunError from its atomic UPDATE matching zero rows is
+        # provably unreachable here -- step 2's row lock is still held --
+        # and is deliberately left uncaught (see this function's own
+        # docstring, step 7).
+        apply_task_control_transition(task, TaskControlState.RESUME_REQUESTED)
```
rogercloud, line 2073: WAITING_FOR_USER → RESUME_REQUESTED is applied here and re-applied by the coordinator when it executes the staged command (websocket.py:7810-7814), and apply_task_control_transition has no legality table — one answer double-bumps state_version. Also, because respond() never flips Task.status, between answer-commit and coordinator execution the task reads status=WAITING_FOR_USER + RESUME_REQUESTED with no active row, and the status-gated path at websocket.py:7784 still treats it as resumable. Suggest single-writer: let respond() not transition and the coordinator transition exactly once during execution, or add a legality check making repeat application a no-op.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,660 @@ def materialize_compatibility_view(
...
+        or envelope.protocol_version != INTERACTION_PROTOCOL_VERSION
+    ):
+        return RespondValidationRejected(reason="unknown_protocol_version")
+    if not isinstance(envelope.idempotency_key, str):
+        return RespondValidationRejected(reason="malformed_idempotency_key")
+    try:
+        normalized_key = _normalize_command_id(envelope.idempotency_key)
+    except ValueError:
+        return RespondValidationRejected(reason="malformed_idempotency_key")
+    if not isinstance(envelope.values, dict):
```
rogercloud, line 1885: isinstance(envelope.values, dict) is the only answer validation: no per-field validation against the InteractionArg definition (the docstring at 1699-1702 admits it and it is in neither named follow-up), so malformed-but-dict answers are stored and staged into the RESUME payload — while create() validates via parse_v1_request_payload (pydantic) plus a JSON-serializability probe. Dicts with non-JSON-serializable values (datetime/set/bytes) also raise an uncaught TypeError/StatementError during fence-bind serialization (verified on SQLAlchemy 2.0.51), escaping the 8-variant RespondOutcome contract. Suggest validating values against the InteractionArg definition after row load (presence/type/option membership), reusing create()'s shape; at minimum, name the closing gap in a follow-up.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -44,55 +42,81 @@
...
+compatibility materialization view; the answer fence and its active-row
+predicate; ``respond()``'s own call body (validation, authorization,
+anchor resolution, the idempotency pre-read, the answer fence, the task
+control-state transition, staging the resume command, and committing --
+this delivery reports an unresolved fence miss or a failed commit as
+``RespondOutcomeUnknown`` rather than further classifying or reconciling
+either; the response-conflict counter
+(``COUNTER_LIFECYCLE_RESPONSE_CONFLICT``); and the compatibility seam that
+routes the existing resume coordinator through this service
+(``websocket.py``'s ``_handle_resume_task_unserialized``). ``create()``'s
```
rogercloud: This docstring claims the module "now delivers ... the compatibility seam that routes the existing resume coordinator through this service", and the gate docstring (tests/web/services/test_task_interaction_service_create_gate.py:7-12) says "That change has landed" — but the seam exists in no commit of this PR, HEAD, or the merge-base (websocket.py has no task_interaction_service/_active_native_row_criteria reference and this PR does not touch it). The legacy resume path retires the active row via close_legacy_resume_interaction_sync and continues. Suggest rewriting both docstrings to state the facts (seam ships separately; respond has zero production callers), or delivering the seam before/with the docstring.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1329,660 @@ def materialize_compatibility_view(
...
+            # acknowledgment could have been lost after the server applied
+            # it -- but this build does not attempt to reconcile that
+            # against the durable graph (that reconciliation is not
+            # delivered here): it retires its session and reports the
+            # ambiguity, leaving the caller to re-check.
+            session_retired = True
+            _retire_respond_session_best_effort(db)
+            return RespondOutcomeUnknown()
+
+        notify_task_command_dispatcher()
```
rogercloud: notify_task_command_dispatcher() runs unguarded after a successful commit; the loop.is_closed() TOCTOU can raise RuntimeError and make respond() escape with an exception after the answer is committed — a retry then hits the replay branch, returns a receipt, but never re-notifies (idle polling at DISPATCHER_IDLE_SECONDS=0.5 is the documented fallback). Suggest wrapping the notify in a best-effort try/except so a post-commit notify failure cannot surface as a failed answer.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -267,34 +299,542 @@ def test_active_row_predicate_three_way_tiering(db_session) -> None:
...
+
+    t1 = threading.Thread(target=holder)
+    t2 = threading.Thread(target=racer)
+    t1.start()
+    t2.start()
+    t1.join()
+    t2.join()
+
+    assert isinstance(results["holder_outcome"], svc.RespondAccepted), results
+    assert results["racer_wait_s"] >= 0.8, results
```
rogercloud, line 549: racer_wait_s >= 0.8 rests on a 0.1s wall-clock stagger plus the holder's injected sleep(1.0); under CI scheduling delays it can flake in both directions (holder late to take the lock, or racer delayed past its UPDATE). Suggest threading.Event signaling: the holder sets the event after acquiring the lock and the racer waits on it before issuing its UPDATE, making the interleaving deterministic.

tests/web/services/test_task_interaction_service_create_gate.py
```diff
@@ -291,24 +283,42 @@ def handler(**kwargs):
     assert _production_uses(source) == set()
 
 
+def test_a_production_call_to_respond_is_not_flagged() -> None:
```
rogercloud: This test pins that a production call to respond() is not flagged — but respond() was removed from GATED_NAMES in this very PR (the one that makes it a real callable), and restoring it to the gated set produces zero violations (verified). The PR text's "the gate is the regression guard for that fact" thus contradicts the code. Suggest restoring respond() to the gated set, flipping this test into a positive control, and aligning the PR text.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. The staging result is now required to be a row this call itself created carrying this call's own payload: `if not (staged.created and staged.payload_matches)` rolls the whole transaction back — fence UPDATE and CAS included — and reports the unresolved outcome, exactly like the `IntegrityError` branch beside it. The two doors for the same race now get the same conservative answer.

One deliberate choice, recorded in the code and pinned by a test: a `created=False` hit whose payload *matches* is also folded onto the unresolved outcome rather than reported as a replay. Replay recognition belongs to the pre-read; a hit that only becomes visible after it is a race, not a replay. The test is parametrized over both doors (`payload-mismatch` / `payload-matches`), each asserting zero residue, and mutation-checked: narrowing the guard to `payload_matches` alone turns exactly the matching-payload case red.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. The idempotency pre-read now runs before anchor resolution — still after authorization, so a guessed key cannot read back another principal's receipt. Replay recognition asks about this call's key and the row it already wrote; it no longer depends on a live anchor.

A new test pins the exact scenario you described: an answered row with a staged matching command whose `resume_trace_event_id` has been nulled (the `ON DELETE SET NULL` outcome after pruning) now returns the original receipt instead of `Stale(anchor_dangling)`. Mutation-checked by moving the block back — the test goes red with exactly the stale outcome.

One note on the regression side: the "fresh call with a dangling anchor still reports stale" guard needs no new test and cannot be built the way I first tried — `ck_task_interaction_requests_active_anchor` forbids an active row with a null anchor at the schema level, which also means the pruner can never null an *active* row's anchor (the SET NULL would violate the CHECK). The existing wrong-run-partition test already covers the reachable stale path after the move.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Taking the documentation half of this and deliberately not the mechanism half in this change.

Confirmed and now recorded in the docstring: the coordinator re-applies the same transition when it executes the staged command, `apply_task_control_transition` has no legality table, so one answer bumps `state_version` twice. The docstring now states the +2 plainly and that consumers must be monotone in `state_version`.

Why not the mechanism here: the second bump is not reachable today — nothing in production calls `respond()`, so no RESUME command carrying its payload is ever staged or executed. It becomes reachable with the change that wires the first caller, and choosing the single writer (respond stops transitioning, or the coordinator's application becomes idempotent) is that change's design decision. Adding a legality table to `apply_task_control_transition` is not something this change can carry: that helper is shared by eight direct call sites across five modules plus five more through `transition_task_control_state_sync` — it would put the whole task state machine under this PR.

On the interim-state half: `websocket.py`'s status gate must admit `WAITING_FOR_USER`, or the staged RESUME command itself could never execute; and a concurrent second resume in that window is blocked by `reserve_resume`, not by the status gate. So the window exists but is not misread as resumable by an unguarded path. Leaving this thread open for your read on the deferral.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed the escape half here; the per-field half is now tracked in #1368.

Step 1 now probes `json.dumps(envelope.values, allow_nan=False)` — the same probe the question side runs — and rejects as `invalid_values`. That covers the `datetime`/`set`/`bytes` case you found (reproduced locally: `StatementError` from inside the fence UPDATE, and `StatementError` is not an `IntegrityError` subclass so nothing caught it) plus one quieter case the reproduction surfaced: a `nan`/infinite float did not raise at all — the default encoder stored a bare non-JSON token silently. Four parametrized tests pin the probe; deleting it turns all four red.

Per-field validation against the question's own argument definitions is a schema decision (answer contract across the interaction types, needs the stored `request_payload` which is only available after the row load), so it is written up as #1368 rather than squeezed in here. The docstring already stated that gap; the signature docstring now also records exactly which operational failures still escape and why folding them into the unresolved outcome would hide a dead database behind an ambiguous answer.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. Both docstrings now state the facts: the compatibility seam ships as its own change in this same series and is not part of this one — nothing in `websocket.py` references this module today — and the gate docstring no longer claims the routing change landed. The gate itself also changed in response to the sibling thread on the gated set: it watches both `create` and `respond` again, so the "zero production callers" statement is once more a guarded fact rather than an unguarded one.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. The notify is now best-effort: it is the only statement that runs after the answer is durable, so a raise there turned a committed answer into a reported failure. The catch wraps that one call and nothing else — the commit stays outside it in its own narrow handler — and logs a warning naming the interaction and task, with the dispatcher's idle poll as the documented delivery fallback. A test pins it: with the notify patched to raise `RuntimeError("Event loop is closed")`, the call still returns the accepted outcome and the row is durably answered; removing the guard turns the test red with the raw `RuntimeError` escaping.

tests/web/services/test_task_interaction_service_postgresql.py
AlexLiu190625: Fixed with the `threading.Event` you suggested, placed on the existing monkeypatch point: the holder's wrapped fence-statement builder sets the event — that code runs after step 2's row lock is taken and before the fence write, which is the exact moment the racer's UPDATE must start blocking from — and the racer waits on the event (30s timeout with a message) instead of a 0.1s stagger. The real lock contention the test exercises is unchanged, and the `>= 0.8` assertion now measures from a deterministic start. Verified against a real PostgreSQL server: the file's 11 tests pass.

tests/web/services/test_task_interaction_service_create_gate.py
AlexLiu190625: Fixed as suggested. `respond` is back in the gated set, the production scan watches both names (and stays green — your zero-violations check reproduced), and the negative test flipped into a positive control asserting a hypothetical production call to `respond` *is* detected, same shape as the `create` control. The PR text's gate sentence is updated to match, and the gate file's remaining single-name sentences (the resolution rule, one bare-import docstring) were brought in line in the same pass. The change that wires a caller is the change that takes the corresponding name out of the gated set.

AlexLiu190625:
Dispositions for the review-body findings without their own threads, all pushed in this batch (the branch head as of this comment):

- **Unknown conflation / retry-convergence claim**: the fence-miss path now logs the reread row's full state (status, slot, terminal reason, run, responder) — the reread already had the row in hand, so it costs no extra statement — and `RespondOutcomeUnknown`'s docstring plus the PR text now separate the retry-converging reasons (ambiguous commit, rolled-back staging race) from the one that never converges (fence miss). Classifying the miss stays with the fence-classification follow-up.
- **Escape surface**: documented in the signature docstring — which exception families still propagate (database-level failures outside the narrow catches, the two non-`IntegrityError` staging exceptions, the two structural invariants) and why they are not folded into the unresolved outcome: a caller that cannot tell "the database is down" from "your answer was ambiguous" retries the first one forever. Every escaping path leaves no partial write (the session retirement in `finally` rolls back).
- **Fence SQL vs step-3 terms**: docstring narrowed to the actual guarantee — three terms re-asserted at write time (task still waiting, `Task.user_id`, guest `guest_id`), three evaluated at read time only. The SQL itself is unchanged: moving the remaining terms into the fence is scope this change should not take on, and the described SQLite scenario surfaces as a busy/snapshot error under the WAL journaling the test engine runs with, not as a silent non-re-evaluation.
- **Matrix arithmetic + missing authorization tests**: the four missing step-3 rejection tests landed (guest entity-binding mismatch, zero and two populated directions, unknown principal kind — each mirroring its question-side twin, each mutation-checked), and the matrix was recounted after this batch's own additions: 27 cells, the same 14 distinct pairs, with the cell-id gaps annotated.
- **Two vocabulary mechanisms**: kept, with one note replacing the scattered disclaimers — the runtime dictionaries record which reasons are *producible* as opposed to declared, a distinction that exists only while `create()`'s call body is undelivered; that side converts to the `Literal` mechanism in the change that fills the body.
- **Session-retirement helper**: declined — the control flow matches the other two sites, but this module has no module-level logger, so importing the existing helper would re-attribute its warnings to another module's logger; not a pure move. (The import-precedent module referenced in the finding is the one under the v1 API package, `api/v1/tasks.py`.)
- **Cross-module private imports**: no action, per the finding's own note.
- **Simplification list**: applied except the helper merge above — the `_rowcount` inline, the six fixture-parameter removals, the fixture-field trim (keeping `id`/`task_id`, which the receipt guard's error message reads), and the three-way principal-helper merge (21 call sites). Two corrections to the list as written: the `run_id`/`status` values were default parameters rather than hardcoded in the bodies, and the unanswered-fixture item points at the unit test file, not the service module.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1353,821 @@ def materialize_compatibility_view(
...
+    if not isinstance(envelope.idempotency_key, str):
+        return RespondValidationRejected(reason="malformed_idempotency_key")
+    try:
+        normalized_key = _normalize_command_id(envelope.idempotency_key)
+    except ValueError:
+        return RespondValidationRejected(reason="malformed_idempotency_key")
+    if not isinstance(envelope.values, dict):
+        return RespondValidationRejected(reason="invalid_values")
+    try:
+        json.dumps(envelope.values, allow_nan=False)
```
rogercloud: NN1 [minor] — The step-1 probe here mirrors the fence bind's serializer except for `sort_keys=True`, which `_canonical_payload` applies recursively (src/xagent/web/services/task_command_transport.py:209-210). A `values` dict with mixed int/str keys passes this probe (`json.dumps` coerces int keys to strings) and then raises an uncaught `TypeError` on the replay/conflict comparison path — a crash that escapes the typed-outcome contract. It is not reachable from JSON-over-HTTP (string keys only), so it needs a programmatic caller, and the rollback stays clean, but the contract violation is real (reproduced with a one-liner). Suggestion: add `sort_keys=True` to the probe so the same dict is rejected here in step 1 with `RespondValidationRejected(reason="invalid_values")` instead of crashing later.

tests/web/services/test_task_interaction_service.py
```diff
@@ -1204,3 +1349,1443 @@ def test_module_issues_zero_sa_text_calls() -> None:
...
+    with pytest.raises(RuntimeError, match="carries no answer"):
+        svc._respond_receipt(
+            interaction=unanswered,  # type: ignore[arg-type]
+            task=task,  # type: ignore[arg-type]
+            command_db_id=1,
+            idempotency_key="key-1",
+        )
+
+
+def test_respond_reports_outcome_unknown_when_the_fence_misses_and_leaves_no_residue(
```
rogercloud, line 2156: NN2 [minor] — This test proves the fence-miss → `RespondOutcomeUnknown` → zero-residue path end-to-end for the already-answered trigger, and the guest authorization tests only reach the predicate level: no behavioral test drives (a) a guest principal end-to-end to `RespondAccepted`, (b) a fence miss via a task not in `WAITING_FOR_USER`, or (c) a fence miss via a `run_id` mismatch. The fence's `WHERE` re-asserts the lifecycle, task-waiting, and ownership terms at write time, so a regression in either uncovered term would go uncaught. Suggestion: three small tests — one guest accept asserting the receipt fields, and one miss per uncovered term asserting `RespondOutcomeUnknown` plus `_asserts_no_side_effects`.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1353,821 @@ def materialize_compatibility_view(
...
+            logging.getLogger(__name__).warning(
+                "failed to notify the task command dispatcher after "
+                "committing the answer for interaction %s on task %s; "
+                "the dispatcher's idle poll will still pick the command up",
+                interaction_id,
+                task_id,
+                exc_info=True,
+            )
+        return RespondAccepted(
+            receipt=InteractionResponseReceipt(
```
rogercloud, line 2186: NN3 [minor] — The accept path builds `InteractionResponseReceipt` inline here, duplicating `_respond_receipt`'s ten fields (src/xagent/web/services/task_interaction_service.py:1552-1598) with different sourcing: `status="answered"` is hardcoded and `responder_identity` is principal-derived, while `_respond_receipt` reads both from the row. Values are identical today, but the two constructions can drift. The accept path cannot call `_respond_receipt` as-is (it reads ORM attributes that would silently re-issue SQL after commit under `expire_on_commit`), so a shared builder fed by the ten captured scalars would remove the duplication without the lazy-reload hazard. Suggestion: extract a scalar-fed builder (taking status, responder_identity, and the other captured locals) used by both paths.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1353,821 @@ def materialize_compatibility_view(
...
+
+        # Capture every receipt value this transaction can still commit as
+        # plain Python locals now, before the commit below. SQLAlchemy's
+        # default ``expire_on_commit=True`` invalidates every ORM attribute
+        # on ``ir``/``task`` the instant ``db.commit()`` returns -- reading
+        # ``ir.run_id`` or ``task.state_version`` afterward would silently
+        # re-issue a SELECT against a session this function is about to
+        # decide whether to keep or retire. The fence UPDATE and the CAS
+        # above are Core statements, so ``ir``/``task``'s in-memory
+        # attributes were never updated by them either; these locals are
```
rogercloud: NN4 [minor] — Two documentation inaccuracies. (a) This comment claims the fence UPDATE and the CAS are Core statements so `ir`/`task`'s in-memory attributes "were never updated by them either" — but `apply_task_control_transition` ends with `session.refresh(task)` (src/xagent/web/services/task_execution_controller.py:161), so the captured `committed_state_version`/`committed_control_state` are post-CAS row reads, not "the values this call itself just wrote". (b) `_respond_command_payload`'s docstring (src/xagent/web/services/task_interaction_service.py:1527-1528) says the payload is re-derived "at steps 5 and 6", but the lookups happen at steps 5 and 8 — step 6 is the fence. Both are fixable in a sentence; the refresh also means the `expire_on_commit` rationale applies to `ir`'s attributes but not `task`'s.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1314,3 +1353,821 @@ def materialize_compatibility_view(
...
+            # ``Replayed`` on purpose -- step 5 is where a replay is
+            # recognized, and a hit that only appears after it is a race,
+            # not a replay.
+            db.rollback()
+            return RespondOutcomeUnknown()
+
+        command_db_id = staged.staged_db_id
+        try:
+            db.commit()
+        except Exception:
```
rogercloud, line 2139: NN7 [minor] — This catch retires the session and returns `RespondOutcomeUnknown` without any log record, and the two staging-race doors (IntegrityError and `created=False`) are equally silent. A commit-time DB outage therefore leaves no trace distinguishing "ack lost after a durable write" from "write never left" — inconsistent with the fence-miss branch, which logs the reread row state by policy (34cb1e45), and with the escape-surface rationale in the signature docstring. Suggestion: mirror that discipline — `logging.getLogger(__name__).warning(..., exc_info=True)` naming the interaction and task ids before retiring, so an operator can at least see the ambiguity was surfaced.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -492,24 +519,50 @@ class InteractionResponseReceipt:
     task_control_state: str
```
rogercloud, line 522: NN8 [minor] — The `RespondOutcome` section header (src/xagent/web/services/task_interaction_service.py:486-488) still says respond() itself "lands with a later change (see the module docstring)" — pre-existing text now directly contradicted by this PR's delivery, and missed by 1f8d8e16's docstring rewrite, which updated the CreateOutcome sibling but not this header. Suggestion: rewrite the parenthetical so the header states that respond() is delivered here, rather than telling future readers the opposite of the code.

rogercloud (approved):
## Summary

`respond()` is the answer-side entry point of the task-interaction lifecycle service: it validates the `RespondEnvelope` type-before-value, authorizes the principal, flips the still-active interaction row to `answered` through a compare-and-swap fence whose `WHERE` re-asserts row lifecycle, task-waiting state, and ownership at write time, transitions the task control state to `RESUME_REQUESTED`, stages the RESUME command under a normalized idempotency key, commits, and best-effort notifies the dispatcher. Every outcome is one of eight typed `RespondOutcome` variants; every conservative failure path returns `RespondOutcomeUnknown` with zero residue, and a same-key replay returns the original receipt. The function has zero production callers and is gate-watched. Blocking: no — recommended event: APPROVE

## Update summary

The 11 commits since b5d37796 are a focused fix-and-test pass over the previously reviewed design: the raced-command-row door now rolls back to `RespondOutcomeUnknown` (ec4fd1a4); replay recognition moved before anchor resolution (cd0835fa); answer-side failures were kept inside the typed-outcome contract via a JSON probe, a guarded notify, and an escape-surface docstring (365146bb); fence-miss row-state logging landed (34cb1e45); delivered-scope docstrings were corrected (1f8d8e16); the caller gate watches both entry points again and the wall-clock race test was replaced with event signaling (3a064e13); four authorization-branch tests were added (a8061238); test-helper parameters were dropped and the owning-principal helpers merged (21d16a75, 0ad7c868); a gate docstring was fixed (2c94e707); and the matrix cell accounting was updated for this change's own new cells (9d8ea6d1).

## Prior findings checklist

- F1 [major] FIXED — the staging race now rolls back through both doors (`created=False` and `IntegrityError`); no raced shape can still produce a `RespondAccepted` naming another writer's row; pinned by two zero-residue tests.
- F2 [major] FIXED — the idempotency pre-read now runs before anchor resolution; the pruned-anchor replay test covers an answered row whose anchor was pruned.
- F3 [major] PARTIAL (residual minor) — the docstring now records the +2 bump and the monotone-consumer contract, but the mechanism is unchanged and the author's `reserve_resume` claim does not cover the pre-execution window; latent only given zero production callers.
- F4 [minor] WAIVED with residual nit — the admin branch is still unpinned by tests and the `actor_user_id` docstring claim is conditionally false for an admitted admin; policy deferred to the wiring change.
- F5 [major] FIXED — the step-1 JSON probe closes the serialization escape (NaN included); per-field validation is tracked in #1368.
- F6 [major] FIXED — phantom-seam docstrings corrected (module docstring, gate docstring, zero websocket references).
- F8 [minor] FIXED — the fence-miss branch logs the reread row state (status/active_slot/terminal_reason/run_id/responder_identity) and is pinned by a test.
- F9 [minor] FIXED — the escape surface is enumerated with per-unit rationale; serialization and notify degrade to typed outcomes.
- F10 [minor] FIXED — the guest TOCTOU docstring now states exactly which terms are re-asserted at write time vs read-time-only.
- F11 [minor] FIXED — notify is guarded (warning log + documented idle-poll fallback) and pinned by a test.
- F12 [minor] PARTIAL (residual minor) — matrix arithmetic fixed (27 cells summing to 27, 14 pairs) and the A2 cell is covered; the meta-test blind spot is not documented, latent.
- F13 [minor] FIXED — the wall-clock race window was replaced with `threading.Event` signaling and a deterministic start.
- F14 [minor] FIXED — the intentional-asymmetry note landed on the CreateOutcome section.
- F15 [nit] WAIVED — unchanged, acceptable.
- F16 [minor] NOT FIXED (residual minor) — the third same-shaped retire-session copy remains at src/xagent/web/services/task_interaction_service.py:1601-1620; recorded non-blocking.
- F17 [minor] FIXED — the gate watches both entry points again (`GATED_NAMES = frozenset({"create", "respond"})`) and the positive control test passes.

## New findings this pass (all minor, non-blocking)

- NN1 — a `values` dict with mixed int/str keys passes the step-1 probe (no `sort_keys=True`) and crashes `_canonical_payload` with an uncaught `TypeError` on the replay/conflict path; add `sort_keys=True` to the probe (task_interaction_service.py:1876).
- NN2 — no behavioral test drives a guest principal end-to-end to `RespondAccepted`, a fence miss via a task not in `WAITING_FOR_USER`, or a fence miss via a `run_id` mismatch; the fence-miss → `RespondOutcomeUnknown` → zero-residue path is proven end-to-end only for the already-answered trigger.
- NN3 — the accept path's receipt construction duplicates `_respond_receipt`'s ten fields with divergent sourcing (hardcoded `status="answered"` + principal-derived identity vs row reads); a shared scalar-fed builder would remove the drift risk (task_interaction_service.py:2158-2169 vs 1552-1598).
- NN4 — the receipt-capture comment is wrong about task attributes (`apply_task_control_transition` ends with `session.refresh(task)`, task_execution_controller.py:161) and `_respond_command_payload`'s docstring cites steps 5/6 where the lookups happen at steps 5/8 (task_interaction_service.py:2064, 1527-1528).
- NN5 — the PR body's one-module coupling claim overstates: `_KIND_VOCABULARY` is imported from task_interaction_staging (defined at :186) and has callers in two modules; `INTERACTION_PROTOCOL_VERSION` and `_normalize_command_id` also live elsewhere.
- NN7 — the commit catch (and both staging-race doors) log nothing, so a commit-time DB outage leaves no trace distinguishing "ack lost after a durable write" from "write never left"; mirror the fence-miss logging discipline (task_interaction_service.py:2117).
- NN8 — the `RespondOutcome` section header still says respond() "lands with a later change", now directly contradicted by this PR's delivery and missed by 1f8d8e16's rewrite of the CreateOutcome sibling (task_interaction_service.py:486-488).
- NN6 dropped — the replay-receipt wording is already covered by the "(built from the current row state)" docstring clause.

- `shrink:` merge the two adjacent kind checks at task_interaction_service.py:1857-1859 into `if not isinstance(envelope.kind, str) or envelope.kind not in _KIND_VOCABULARY:` — create() at :943 already does exactly this and the short-circuit `or` preserves type-before-value ordering.
- `shrink:` hoist the three function-local `import logging` (task_interaction_service.py:1611/2031/2147) to the module header — 49 of 51 sibling services import at module level; `from ..models.database import get_session_local` at :1894 must stay function-local, it is a deliberate test seam both fixtures patch.
- `delete:` drop `_waiting_task`'s `state_version: int = 5` parameter (test_task_interaction_service.py:1378) — none of the 27 call sites overrides it (the only explicit pass, at :2356, repeats the default).
- `shrink:` replace the hand-rolled save/restore of `notify_task_command_dispatcher` (test_task_interaction_service.py:2360-2377) with the `monkeypatch.setattr` fixture the sibling test at :2401 already uses — ~12 lines.
- `net: -18 lines possible` — 9 of the 10 prior simplification items landed via 21d16a75/0ad7c868; only the retire-session extraction was not applied (same item as F16).

Residuals, all non-blocking:
- F3 [prior, minor] — latent double-writer in the pre-execution window; the wiring change must resolve single-writer ownership and not rely on `reserve_resume`.
- F12 [prior, minor] — the meta-test blind spot should be documented in the docstring.
- F16 [prior, minor] — the third same-shaped retire-session copy remains; extraction still pending.
- F4 [prior, minor] — admin-branch test gap plus the `actor_user_id` docstring overstatement.
- The seven new findings (NN1-NN5, NN7, NN8) are all minor [new] — none blocks approval.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed — and reproduction showed it was slightly worse than described: the first call with a mixed-key dict did not merely pass the probe, it stored the coerced form (`{1: 'a'}` persisted as `{'1': 'a'}`) with no error, the same silent-corruption class the probe's nan case exists to catch; the crash only surfaced on the second call. The probe now passes `sort_keys=True`, making it a strict superset of every serializer downstream of step 1, and the mixed-key dict is rejected up front as `invalid_values`. A new parametrized case pins it; removing `sort_keys` turns exactly that case red (the outcome degrades to an acceptance).

tests/web/services/test_task_interaction_service.py
AlexLiu190625: Agreed on all three gaps, and recording them for the change that wires the first production caller rather than adding them here. Reasoning on priority: the guest-accept path's risk direction is functional (a guest wrongly refused), not a security direction — the existing guest rejection tests already pin the over-admission direction at both the predicate and the fence level. The two uncovered fence-miss triggers can only assert the same collapsed unknown outcome this build maps every miss onto, so their regression value rises substantially once the fence classification change lands; landing them together with it (or with the wiring change's end-to-end pass) buys the same coverage with assertions that can actually distinguish outcomes. Leaving this open as the record.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Agreed the duplication is real and the scalar-fed builder is the right shape — you're also right about why the accept path cannot call `_respond_receipt` as-is. Deferring the extraction rather than doing it here: the module has zero production callers, so the drift risk is contained while the two constructions are pinned by the same receipt-field tests, and this PR is at the end of an approved review cycle. What did land now is the documentation half: `_respond_receipt`'s docstring claimed the identity is "never reconstructed from a caller's principal", which the accept path contradicts by construction; it now states the exception and why the two sources cannot disagree (the fence UPDATE writes that column from the same `identity_string()` the accept path captures). Leaving this open as the record for the extraction.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Both fixed. The receipt-capture comment now credits the actual mechanism — `apply_task_control_transition` ends with a `session.refresh(task)`, so the captured values are post-CAS row reads; what makes the pre-commit capture necessary is `expire_on_commit` invalidating every ORM attribute after commit, not the statements being Core. And `_respond_command_payload`'s docstring now says steps 5 and 8. While in there, the same pass corrected two comments that referenced a serialization probe by a name that exists nowhere in the tree (the question-side probe lives in `build_v1_request_payload`).

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Taken for the commit catch, declined for the two staging-race doors.

The commit catch was the real gap: it swallowed an arbitrary, unclassified exception whole — no log, no `exc_info`, nothing an operator could find after a commit-time outage. It now logs a warning with the exception before rolling back, and a test pins the log line.

The two staging-race doors stay silent on a different rationale than the fence miss. The fence-miss logging landed because a retry there never resolves the ambiguity — the same predicate misses the same row every time, so the log line is the only place the distinction exists. Both race doors are the opposite case: a retry under the same key goes through the step-5 pre-read and comes back as a replay or a conflict, so the ambiguity is self-resolving, and under load a raced duplicate is an expected event rather than an incident — logging each one is noise with no diagnostic value the retry doesn't already provide. The docstring sentence that called the fence miss "the one path that leaves nothing behind" was stale either way and now states the distinction precisely.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. The section header now states that `respond()` is delivered in this module, and the same pass caught the two sibling sentences that had gone stale the other way: the module docstring and the gate docstring both claimed `websocket.py` holds no reference to this module — true when written, false since the resume-seam change merged — and both now state the merged fact (the seam reads `_active_native_row_criteria`; `respond()` still has no production caller, which the gate keeps guarding).

AlexLiu190625:
Dispositions for the body-level items in this round, all pushed (head 0c976a9df): the PR text's one-module coupling claim now names the three imported symbols instead of overstating; the four simplification items landed (merged kind checks, module-level logger with the patched database import kept function-local as noted, the redundant test-helper parameter dropped, and the hand-rolled notify save/restore replaced with monkeypatch); the mapping meta-test's blind spot is documented in its docstring. Of the recorded residuals: the retire-session extraction and the admin-branch test gap stay recorded for the wiring change, and the single-writer decision for the control-state transition is registered as an acceptance condition on that same change.

PR #1355, opened 2026-08-14
build(sandbox): lock runtime dependencies with uv

bsbds:
## Summary

- keep the sandbox as a separate, non-root image with a dedicated `sandbox` dependency group
- add `fsspec` and resolve the complete sandbox Python environment through `uv.lock`
- export the locked group in a build-only stage and install it without copying lock metadata into the runtime image
- fail image builds when any supported sandbox package cannot be imported
- remove the hand-maintained `docker/sandbox-requirements.txt` and document the build and release flow

## Motivation

The sandbox image previously installed an independent requirements file, so packages present in the backend lockfile were not guaranteed to exist in the sandbox. This caused `fsspec` to be missing even though it was present in `uv.lock`.

## Verification

- `pytest -q tests/test_docker_workflow_versions.py` — 19 passed
- `pre-commit run --all-files` — all hooks passed
- built `docker/Dockerfile.sandbox` locally
- confirmed the runtime runs as UID 1100 and imports all supported packages, including `fsspec 2026.2.0`
- confirmed `pyproject.toml` and `uv.lock` are absent from the runtime image

## Deployment impact

A new `xprobe/xagent-sandbox` image must be published before deployments can use these dependencies. Existing deployments remain unchanged until their sandbox image tag is updated.

## Rollout

1. Merge this PR.
2. Publish `xprobe/xagent-sandbox:<new-tag>` from the merged revision using `sandbox-publish.yml`.
3. Verify both image architectures and run `python -c "import fsspec; print(fsspec.__version__)"` against the published image.
4. Update the sandbox image pin in `xagent-saas`.
5. Deploy through development, stage, and production.

Existing `xagent-saas` deployments remain on sandbox `0.6.4` until the downstream pin changes. Idle sandboxes reconcile to the new image when reused; active sandboxes are not forcibly replaced.

## Rollback

Restore the previous sandbox image pin in `xagent-saas` and redeploy. Use a maintenance drain only when every existing sandbox must change immediately.


rogercloud (changes requested):
## Re-review of PR #1355 — build(sandbox): lock runtime dependencies with uv

**PR summary.** This PR replaces the hand-maintained `docker/sandbox-requirements.txt` with a dedicated `[dependency-groups].sandbox` group in `pyproject.toml` and `uv.lock` (adding `fsspec`), and reworks `docker/Dockerfile.sandbox` into a multi-stage build: a build-only stage exports the locked group via `uv export --locked --only-group sandbox --no-emit-project` (keeping `pyproject.toml`/`uv.lock` out of the runtime image), the runtime stage installs the pinned set with `uv pip install --system`, and an import smoke test fails the build if any supported package is missing. The hand-maintained requirements file is deleted, and the build/release flow is documented and covered by tests. The approach is the canonical uv multi-stage pattern and fixes the root cause (backend lockfile and sandbox image drifting apart). **Blocking: yes — recommended event: REQUEST_CHANGES**

**Update summary.** There are no commits since the prior automated assistant review (gemini-code-assist, state COMMENTED, submitted 2026-08-14T07:48:34Z) — commit `37c66828` is still HEAD and the diff is unchanged. The prior review contained a single body-only note (no inline comments) about a `--no-cache`/cache-mount redundancy in the sandbox Dockerfile, which is still present at HEAD (see P1 below). This re-review covers the full diff at the same commit.

**Approach verdict.** Sound. The build-only export stage keeps lock metadata out of the runtime image, installs exact locked versions, mirrors `Dockerfile.backend` conventions, and preserves the non-root UID 1100 setup; the import smoke check enforces group coverage at build time. The findings below are minor with one major exception (M1), which is a runtime regression introduced by the image change.

**Prior findings checklist**

| # | Prior finding (gemini-code-assist, body-only) | Status |
|---|---|---|
| P1 | `--no-cache` combined with `--mount=type=cache,target=/root/.cache/uv` at `docker/Dockerfile.sandbox:52-53` — the cache mount is never read or written | **NOT FIXED** (identical code at HEAD; same issue as M3/L52) — minor, does not block |

No other prior findings; no prior inline comments.

**Severity findings**

**M1 — MAJOR — runtime `uv>=0.8.0` requirement now triggers a pip install of uv in every fresh sandbox container** (`docker/Dockerfile.sandbox:28` ↔ `src/xagent/core/tools/adapters/vibe/sandboxed_tool/sandboxed_mcp_tool_helper.py:36`)

The image now ships only the standalone `uv`/`uvx` binaries (`COPY --from=uv /uv /uvx /usr/local/bin/`), which pip does not see as an installed distribution. However, `_MCP_SANDBOX_EXTRA_PACKAGES` in `sandboxed_mcp_tool_helper.py:36` still declares `"uv>=0.8.0"`, and at runtime `SandboxDependencyManager.ensure_requirements` runs `pip install -r /tmp/requirements.txt` in the sandbox (`sandboxed_tool_wrapper.py:121-125`). Under the old image uv was pip-installed, so this was a no-op; under the new image pip finds nothing installed and downloads the uv wheel from PyPI (15–40 MB, unpinned) on the first MCP-tool load. Without PyPI egress, `ensure_requirements` raises `RuntimeError` and the sandboxed MCP server's tools fail to load (recorded as `MCPServerLoadFailure`, `mcp_adapter.py:1473-1484`). The runtime-installed copy also lands in `~/.local`, which `ensure_user_bin_in_path` (`runner_utils.py:9-14`) prepends to PATH, shadowing the pinned image binary — and the download is unpinned, contradicting this PR's locked-deps goal.

Suggested fix: drop `"uv>=0.8.0"` from `_MCP_SANDBOX_EXTRA_PACKAGES` — the image now guarantees the binaries, and already-published old images carry the `uvx` console scripts installed via pip. Alternative: add `uv` to the sandbox group, but then the binary `COPY` at `Dockerfile.sandbox:28` is redundant.

**M2 — MINOR — `uv export --locked` downloads a managed CPython in the build stage** (`docker/Dockerfile.sandbox:17-18`)

The `sandbox-requirements` stage is `node:22-slim` (no Python). `uv export` performs interpreter discovery before the lock check and, with the default `PythonDownloads=Automatic`, downloads a managed CPython (~30–50 MB from github.com) on cold builds. That is a hidden network dependency with a non-deterministic interpreter version, and the build fails when github.com downloads are blocked — contrast `Dockerfile.backend:24` (`ENV UV_PYTHON_DOWNLOADS=0`). Fix options: use `--frozen` (skips discovery; the lock-staleness check is preserved by the `uv lock --check` pre-commit job that ci.yml runs on every non-draft PR), or install python3.11 in the stage. Note the exported markers are evaluated against the downloaded interpreter (e.g. 3.14) while the runtime installs under apt python3.11 — a latent marker-evaluation wrinkle.

**M3 — MINOR — dead cache mount** (`docker/Dockerfile.sandbox:52-53`; same as P1, NOT FIXED)

`--mount=type=cache,target=/root/.cache/uv` combined with `uv pip install --no-cache` means the cache is never read or written. Drop `--no-cache` (mirroring `Dockerfile.backend:64`) or drop the mount.

**M4 — MINOR — package set maintained in three places with no cross-validation** (`pyproject.toml:166-177`, `docker/Dockerfile.sandbox:56`, `tests/test_docker_workflow_versions.py:226-236`)

Adding a package updates pyproject.toml, uv.lock and the test's exact-list assertion, but nothing forces the Dockerfile import line to cover it — a new package would ship without import-smoke coverage while all tests still pass. Today's coverage is 9/9, so this is drift risk only. Fix: extend the Dockerfile test to assert that each group package name appears in the import line, or derive the import list from the exported requirements.

**M7 — MINOR — no PR-time CI exercises the sandbox uv pipeline** (`.github/workflows/sandbox-publish.yml`)

`sandbox-publish.yml` triggers only on tag push and `workflow_dispatch`, and the new tests assert command strings only. This matches the repo-wide convention (no image is built on PRs), but the sandbox image is the least-exercised — it is not even in `nightly-build.yml`. Cheap improvement: a CI job running `uv export --locked --only-group sandbox --no-emit-project` and asserting the output contains the 9 group packages.

**M8 — MINOR — test name under-describes scope** (`tests/test_docker_workflow_versions.py:218-219`)

`test_docker_readme_documents_lockfile_requirement` now also asserts the sandbox README content. Rename to `test_docker_readme_documents_lockfile_and_sandbox_requirements`; do not split it — a second test would duplicate the README read.

**M9 — MINOR — docs still describe a two-image publish flow** (`docker/README.md:363` new Sandbox section vs. 'Docker Files' list ~316-326, 'Publishing Images' overview ~395-404, 'First Time Setup' ~441-456, 'Docker Hub Repositories' ~458-462)

The new Sandbox section documents the image well, but the rest of the README covers only backend+frontend, and `publish.sh` builds only backend+frontend — so operators following the overview will miss that a third image must be published for the new dependency set. Fix: add a sandbox row to those sections.

**M10 — MINOR — design observation, no action required** (body-only)

The sandbox group is a hand-curated whitelist, so the failure class this PR fixes ("backend lock has X, sandbox lacks X") still applies to packages outside the whitelist — e.g. user code doing `s3://` via fsspec needs `s3fs`, which is not in the group. This boundary is intentional and documented, so it is acceptable-with-reservation rather than a defect.

**Simplification opportunities**

- `L52: delete: the --mount=type=cache,target=/root/.cache/uv mount at docker/Dockerfile.sandbox:52 is inert because uv pip install --no-cache (line 53) never reads or writes it — drop the mount (net: -1 line possible; same issue as M3/P1).`

Net: -1 line possible.

**Blocking status & recommended decision**

Blocking: **yes** — recommended event: **REQUEST_CHANGES**.

Blocking findings:
- **M1 [new]** (MAJOR): runtime regression on the new sandbox image — unpinned per-container uv download, sandboxed MCP tool-load failure without PyPI egress, and PATH shadowing of the pinned binary.

Non-blocking: M2, M3, M4, M7, M8, M9, M10 (all minor); prior finding P1 remains NOT FIXED but is minor and does not block.

docker/Dockerfile.sandbox
```diff
@@ -1,8 +1,32 @@
...
+
+############################
+# Sandbox runtime
+############################
+FROM node:22-slim AS sandbox
 
 LABEL maintainer="xagent"
 LABEL description="Full sandbox image with all tools"
 
+COPY --from=uv /uv /uvx /usr/local/bin/
```
rogercloud, line 30: **MAJOR** — this COPY ships only the standalone `uv`/`uvx` binaries, which pip does not see as an installed distribution. `src/xagent/core/tools/adapters/vibe/sandboxed_tool/sandboxed_mcp_tool_helper.py:36` still declares `"uv>=0.8.0"` in `_MCP_SANDBOX_EXTRA_PACKAGES`, and `SandboxDependencyManager.ensure_requirements` runs `pip install -r /tmp/requirements.txt` in the sandbox (`sandboxed_tool_wrapper.py:121-125`). Every fresh container will now download the uv wheel from PyPI on first MCP-tool load (15–40 MB, unpinned — contradicting this PR's locked-deps goal), and without PyPI egress `ensure_requirements` raises `RuntimeError`, so the sandboxed MCP server's tools fail to load (`MCPServerLoadFailure`, `mcp_adapter.py:1473-1484`). The pip-installed copy in `~/.local` also shadows the pinned binary via `ensure_user_bin_in_path` (`runner_utils.py:9-14`). Suggested fix: drop `"uv>=0.8.0"` from `_MCP_SANDBOX_EXTRA_PACKAGES` — the image now guarantees the binaries, and already-published old images carry the `uvx` console scripts via pip. Alternative: add `uv` to the sandbox group, but then this COPY is redundant.

docker/Dockerfile.sandbox
```diff
@@ -1,8 +1,32 @@
...
+############################
+FROM node:22-slim AS sandbox-requirements
+
+COPY --from=uv /uv /uvx /usr/local/bin/
+WORKDIR /build
+COPY pyproject.toml uv.lock ./
+
+# Export only the dedicated sandbox dependency group. The lock metadata stays
+# in this build stage instead of becoming part of the runtime image.
+RUN uv export --quiet --locked --only-group sandbox --no-emit-project \
```
rogercloud, line 19: **MINOR** — this stage has no Python, so `uv export --locked` performs interpreter discovery before the lock check and, with the default `PythonDownloads=Automatic`, downloads a managed CPython (~30–50 MB from github.com) on cold builds: hidden network dependency, non-deterministic interpreter version, build failure when github.com downloads are blocked (contrast `Dockerfile.backend:24` `ENV UV_PYTHON_DOWNLOADS=0`). Use `--frozen` instead (skips discovery; the lock-staleness check is preserved by the `uv lock --check` pre-commit job ci.yml runs on every non-draft PR), or install python3.11 in the stage. Note the export markers are evaluated against the downloaded interpreter (e.g. 3.14) while the runtime installs under apt python3.11.

docker/Dockerfile.sandbox
```diff
@@ -21,13 +45,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 
-COPY docker/sandbox-requirements.txt /tmp/sandbox-requirements.txt
+COPY --from=sandbox-requirements /sandbox-requirements.txt /tmp/sandbox-requirements.txt
 
-# Install tools dependencies
-RUN pip install --no-cache-dir --break-system-packages -r /tmp/sandbox-requirements.txt \
-    && rm /tmp/sandbox-requirements.txt
+# Install the exact versions exported from uv.lock. The import check fails the
+# build if a supported runtime package is missing from the resulting image.
+RUN --mount=type=cache,target=/root/.cache/uv \
```
rogercloud, line 54: **MINOR** — the cache mount is dead weight: `uv pip install --no-cache` (line 53) never reads or writes `/root/.cache/uv`. This was raised by the prior automated review as P1 and the code is unchanged at HEAD, so status: NOT FIXED. Drop `--no-cache` (mirroring `Dockerfile.backend:64`) or drop the mount.

docker/Dockerfile.sandbox
```diff
@@ -21,13 +45,19 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+    uv pip install --system --break-system-packages --no-cache \
+        -r /tmp/sandbox-requirements.txt \
+    && rm /tmp/sandbox-requirements.txt \
+    && python -c "import cloudpickle, fsspec, matplotlib, mcp; import numpy, openpyxl, pandas, pydantic, pydantic_settings"
```
rogercloud, line 59: **MINOR** — this import smoke line is maintained manually, and nothing forces it to cover new additions to the `[dependency-groups].sandbox` group (`pyproject.toml:166-177`). The test at `tests/test_docker_workflow_versions.py:226-236` asserts the exact group list but not that this import line covers each package, so a new group package would ship without import-smoke coverage while all tests still pass (today 9/9 — drift risk only). Fix: assert in the test that each group package name appears in this import line, or derive the import list from the exported requirements.

tests/test_docker_workflow_versions.py
```diff
@@ -214,4 +215,45 @@ def test_docker_readme_documents_lockfile_requirement() -> None:
...
+        "mcp>=1.12.4",
+        "pandas>=1.3.0",
+        "numpy>=1.21.0",
+        "matplotlib>=3.5.0",
+        "openpyxl>=3.1.0",
+        "fsspec>=2024.0.0",
+    ]
+
+
+def test_sandbox_dockerfile_installs_locked_group_and_smoke_tests_imports() -> None:
```
rogercloud, line 305: **MINOR** — this test (and the exact-list test above) assert command strings only; nothing in PR-time CI builds the sandbox image or runs the uv export. `sandbox-publish.yml` triggers only on tag push / `workflow_dispatch`, and the sandbox image is not in `nightly-build.yml`. This matches the repo convention (no image builds on PRs), but a cheap guard would be a CI job running `uv export --locked --only-group sandbox --no-emit-project` and asserting the output contains the 9 group packages.

tests/test_docker_workflow_versions.py
```diff
@@ -214,4 +215,45 @@ def test_docker_readme_documents_lockfile_requirement() -> None:
 
     assert "`uv.lock` during the Docker build" in readme
     assert "uv sync --locked" in readme
+    assert "`[dependency-groups].sandbox`" in readme
```
rogercloud, line 251: **MINOR** — `test_docker_readme_documents_lockfile_requirement` now also asserts the sandbox README content, so the name under-describes the scope. Rename to `test_docker_readme_documents_lockfile_and_sandbox_requirements`; keep it as one test — splitting would duplicate the README read.

docker/README.md
```diff
@@ -360,6 +360,28 @@ docker buildx build \
   --push .
 ```
 
+### Sandbox
```
rogercloud, line 364: **MINOR** — the new Sandbox section is good, but the rest of the README still documents a two-image flow: 'Docker Files' (~316-326), 'Publishing Images' (~395-404), 'First Time Setup' (~441-456) and 'Docker Hub Repositories' (~458-462) cover only backend+frontend, and `publish.sh` builds only backend+frontend. Operators following the overview will miss that a third image must be published for the new dependency set. Add a sandbox row to those sections.

docker/Dockerfile.sandbox
bsbds: We are not making a separate change for this thread because it reports the same `--no-cache` and uv cache-mount conflict as the earlier canonical finding, so its supplied classification is duplicate ([canonical cache thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3782003973)).

The canonical resolution removes `--no-cache` while retaining the BuildKit cache mount, so no second implementation is needed ([resolution commit](https://github.com/xorbitsai/xagent/commit/b68d45601ab185db96bd046bb21ccf086c921b28)).

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/sandbox-locked-dependencies`
- **Head:** `e4aeca196196fea409c05249d5e8acb0c3f885b3`
- **Destination:** `bsbds/xagent:fix/sandbox-locked-dependencies`
- **Result:** success
- **PR description:** not requested

### Resolved

1. **Thread:** [`PRRT_kwDORQ4Kr86ZM6dn`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3782003973)
   **Local outcome:** resolved
   **Change:** `b68d45601ab185db96bd046bb21ccf086c921b28`
   **GitHub:** resolved

2. **Thread:** [`PRRT_kwDORQ4Kr86ZPguA`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783001795)
   **Local outcome:** resolved
   **Change:** `85a23f772641e554f332685bd75bef85765df281`
   **GitHub:** resolved

3. **Thread:** [`PRRT_kwDORQ4Kr86ZPg4S`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783002682)
   **Local outcome:** resolved
   **Change:** `b68d45601ab185db96bd046bb21ccf086c921b28`
   **GitHub:** resolved

4. **Thread:** [`PRRT_kwDORQ4Kr86ZPhMP`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783004406)
   **Local outcome:** resolved
   **Change:** `b68d45601ab185db96bd046bb21ccf086c921b28`
   **GitHub:** resolved

5. **Thread:** [`PRRT_kwDORQ4Kr86ZPhV7`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783005271)
   **Local outcome:** resolved
   **Change:** `b68d45601ab185db96bd046bb21ccf086c921b28`
   **GitHub:** resolved

6. **Thread:** [`PRRT_kwDORQ4Kr86ZPhda`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783005986)
   **Local outcome:** resolved
   **Change:** `b68d45601ab185db96bd046bb21ccf086c921b28`
   **GitHub:** resolved

7. **Thread:** [`PRRT_kwDORQ4Kr86ZPhlm`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783006778)
   **Local outcome:** resolved
   **Change:** `e4aeca196196fea409c05249d5e8acb0c3f885b3`
   **GitHub:** resolved

### Rejected

1. **Thread:** [`PRRT_kwDORQ4Kr86ZPhA9`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783003462)
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** This reports the same `--no-cache` and uv cache-mount conflict as the earlier canonical finding, and the canonical resolution removes `--no-cache` while retaining the cache mount. A second implementation is not required.
   **Sources:** [canonical cache thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3782003973), [resolution commit](https://github.com/xorbitsai/xagent/commit/b68d45601ab185db96bd046bb21ccf086c921b28)
   **Reply:** [posted](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3788905391)
   **GitHub:** left unresolved

2. **Review body:** [`PRR_kwDORQ4Kr88AAAABJidkWg`](https://github.com/xorbitsai/xagent/pull/1355#pullrequestreview-4935083098)
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** The body repeats the cache finding represented by the canonical inline thread; that thread was implemented and resolved.
   **Sources:** [canonical cache thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3782003973), [resolution commit](https://github.com/xorbitsai/xagent/commit/b68d45601ab185db96bd046bb21ccf086c921b28)
   **Reply:** disposition recorded in this aggregate comment
   **GitHub:** non-thread review disposition recorded

3. **Review body:** [`PRR_kwDORQ4Kr88AAAABJjschg`](https://github.com/xorbitsai/xagent/pull/1355#pullrequestreview-4936375430)
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** The aggregate review body repeats its individual inline outcomes. Each accepted inline outcome was implemented and resolved, and the repeated cache thread received its supplied duplicate disposition.
   **Sources:** [runtime uv thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783001795), [export-stage thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783002682), [cache thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783003462), [smoke-coverage thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783004406), [functional-export thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783005271), [test-name thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783005986), [publishing-docs thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783006778)
   **Reply:** disposition recorded in this aggregate comment
   **GitHub:** non-thread review disposition recorded

### Unresolved

None.

### Totals

- **Resolved:** 7
- **Rejected:** 3
- **Unresolved:** 0

tests/test_docker_workflow_versions.py
```diff
@@ -1,9 +1,39 @@
...
+import subprocess
+import tomllib
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parents[1]
 
+# Distribution names and import names are separate interfaces. Keep the mapping
+# explicit so packages such as pydantic-settings are checked without guessing
+# their import name from punctuation.
+SANDBOX_DISTRIBUTION_IMPORTS = {
```
rogercloud, line 18: Minor: with b68d4560's dict-driven smoke enforcement in place, this is the remaining duplication — `SANDBOX_DIRECT_REQUIREMENTS` (lines 25-35) is a verbatim mirror of `[dependency-groups].sandbox` in pyproject.toml whose only consumers are the self-consistency asserts, and this dict just mirrors its keys. Adding one sandbox dependency touches four files/five places (pyproject.toml, uv.lock, the Dockerfile.sandbox:58 smoke line, and both dicts). Consider a single `{dist: (spec, import_name)}` dict with the import name derived via `dist.replace("-", "_")` (verified correct for all 9 entries), keeping one source of truth for both consumers.

src/xagent/core/tools/adapters/vibe/sandboxed_tool/sandboxed_mcp_tool_helper.py
```diff
@@ -33,7 +33,7 @@
 _MCP_SANDBOX_COMMANDS = {"npx", "uvx"}
 _MCP_SANDBOX_TIMEOUT_SECONDS = 60
 
-_MCP_SANDBOX_EXTRA_PACKAGES = ["mcp>=1.12.4", "uv>=0.8.0"]
+_MCP_SANDBOX_EXTRA_PACKAGES = ["mcp>=1.12.4"]
```
rogercloud, line 36: Minor: this spec (and `SANDBOX_BASE_DEPENDENCIES` at sandboxed_tool_wrapper.py:38-42, e.g. `pydantic>=2.0.0` vs the group's `pydantic>=2.11.7`) is a third, untracked place where sandbox runtime specs live beside the `[dependency-groups].sandbox` group. Note the mechanism: the sandbox container runs pip as root (docker_sandbox.py:1117-1118), so a future floor raise would install into system site-packages, replacing the pinned version from uv.lock — the runtime channel can bypass the lock. Inert today (the pinned image satisfies these floors), but nothing ties these specs to the group: the integration test asserts against this list itself, which is a tautology. Consider deriving the list from the group or asserting compatibility.

docker/Dockerfile.sandbox
```diff
@@ -1,8 +1,34 @@
...
+
+############################
+# Sandbox runtime
+############################
+FROM node:22-slim AS sandbox
 
 LABEL maintainer="xagent"
 LABEL description="Full sandbox image with all tools"
 
+COPY --from=uv /uv /uvx /usr/local/bin/
```
rogercloud, line 30: Minor: this COPY is now load-bearing — after 85a23f77 dropped `uv>=0.8.0` from `_MCP_SANDBOX_EXTRA_PACKAGES`, these binaries are the image's only source of uv/uvx — yet no test asserts them for the sandbox (the backend has `test_backend_runtime_keeps_uv_binaries` at tests/test_docker_workflow_versions.py:203-206), and the smoke check at line 58 imports only Python packages. A refactor to plain-pip installs could silently drop uv/uvx while the suite stays green. Consider mirroring the backend assert plus a `uvx --version` step in the smoke check.

tests/test_docker_workflow_versions.py
```diff
@@ -209,9 +239,101 @@ def test_docker_workflows_pass_package_version_to_backend_build() -> None:
...
+    imported_modules = {
+        alias.name
+        for node in ast.walk(ast.parse(smoke_command.group(1)))
+        if isinstance(node, ast.Import)
+        for alias in node.names
+    }
+    assert imported_modules == set(SANDBOX_DISTRIBUTION_IMPORTS.values())
+
+
+def test_uv_export_locked_sandbox_group_contains_direct_distributions() -> None:
```
rogercloud, line 371: Minor: this subprocess test requires `uv` on PATH with no guard — developers without uv get a raw FileNotFoundError instead of a skip. CI also installs uv unpinned (ci.yml:255 `python3 -m pip install --user uv`) while the image pins `UV_VERSION=0.11.10`, so the export semantics being tested can drift from the image's. Consider skipping when `shutil.which("uv")` is None, and pinning the tested uv (e.g. `uv==0.11.10`) or asserting the version.

src/xagent/core/tools/adapters/vibe/sandboxed_tool/sandboxed_mcp_tool_helper.py
rogercloud, line 36: Minor (residual of the M1 fix): removing `uv>=0.8.0` here couples uvx availability to the sandbox image itself. Custom `SANDBOX_IMAGE` overrides (config.py:2053, env-overridable) not derived from the published image used to work — the backend injected uv at first MCP use — and now fail inside mcp_runner with a raw "uvx: command not found" (PATH-only resolution). Both built-in paths are safe (the new image ships binaries; the pinned 0.6.5 tag has pip-installed uvx). Consider documenting that custom sandbox images must ship uv/uvx, or a targeted early error when a uvx command is requested and uvx is absent.

pyproject.toml
```diff
@@ -163,6 +163,21 @@ all = [
...
+sandbox = [
+  "pydantic>=2.11.7",
+  "pydantic-settings",
+  "cloudpickle>=3.0.0",
+  "mcp>=1.12.4",
+  "pandas>=1.3.0",
+  "numpy>=1.21.0",
+  "matplotlib>=3.5.0",
+  "openpyxl>=3.1.0",
+  "fsspec>=2024.0.0",
```
rogercloud, line 178: Nit: `fsspec>=2024.0.0` — fsspec is CalVer and no such release exists (2023.12.2 → 2024.2.0). Satisfiable (2026.2.0 resolves), so purely cosmetic; the same odd bound pre-exists at line 77 and the group mirrors it.

docker/README.md
```diff
@@ -360,6 +361,28 @@ docker buildx build \
...
+
+The sandbox remains a separate image for untrusted code execution. Its Python
+packages come only from the dedicated `[dependency-groups].sandbox` group in
+`pyproject.toml`. Update `uv.lock` whenever that group changes.
```
rogercloud: Nit: "Update `uv.lock` whenever that group changes" doesn't say how (the backend section documents `uv sync --locked`) — consider "Run `uv lock`". Related: the sandbox compose overlays still pin `xprobe/xagent-sandbox:0.6.5` (the image without fsspec; docker/docker-compose.sandbox.docker.yml:7 and docker/docker-compose.sandbox.boxlite.yml:6), and the "update deployments to reference the new sandbox image tag" note at lines 383-384 names no files — pointing at the compose pins would make the release step actionable.

tests/test_docker_workflow_versions.py
```diff
@@ -209,9 +239,101 @@ def test_docker_workflows_pass_package_version_to_backend_build() -> None:
...
+    dockerfile = read_repo_file("docker/Dockerfile.sandbox")
+    requirements_stage = dockerfile.split(
+        "FROM python:3.11-slim AS sandbox-requirements\n", maxsplit=1
+    )[1].split("FROM node:22-slim AS sandbox\n", maxsplit=1)[0]
+
+    assert "ENV UV_PYTHON_DOWNLOADS=0" in requirements_stage
+    assert "uv export --quiet --locked --only-group sandbox --no-emit-project" in (
+        requirements_stage
+    )
+    assert "--frozen" not in requirements_stage
```
rogercloud: Simplification (yagni): these three absence asserts (`--frozen`, `uv python install`, `apt-get install` not in the requirements stage) guard spellings the positive asserts in this test (lines 285-288) and `test_sandbox_dockerfile_installs_locked_group_and_smoke_tests_imports` already pin; dropping them loses no stage-contract coverage. Weakest leg: `apt-get` has no positive counterpart, but apt in an export-only stage is pointless anyway.

rogercloud (approved):
## Follow-up review: build(sandbox): lock runtime dependencies with uv

### Summary

The PR removes the hand-maintained `docker/sandbox-requirements.txt` and derives the sandbox image's Python environment from the single committed `uv.lock` via a dedicated `[dependency-groups].sandbox` group, exported in a build-only stage and installed pinned into the runtime stage. The image build now fails fast when any supported import is missing (fsspec was previously absent from the sandbox), and the Docker README documents the sandbox build/publish flow. All checks pass.

### Update since last review

Three commits landed after the last review (37c66828): 85a23f77 drops the per-fresh-sandbox `uv>=0.8.0` reinstall for MCP tools (the image ships pinned uv/uvx binaries) and pins the resulting requirement list in a test; b68d4560 hardens the locked-dependency checks (export stage moved to `python:3.11-slim` with `ENV UV_PYTHON_DOWNLOADS=0`, `--no-cache` removed, mirror dicts plus AST-parsed smoke-import equality and a real `uv export --locked` subprocess test); e4aeca19 documents sandbox image publishing in `docker/README.md`. All prior findings are addressed; the residual half of M7 is carried forward as V6 below.

### Prior findings checklist

- [x] **P1/M3** (minor, dead `--no-cache` + cache mount) — **FIXED**: `--no-cache` removed in b68d4560; the cache mount is retained (docker/Dockerfile.sandbox:54-58).
- [x] **M1** (major, runtime `uv>=0.8.0` pip install per fresh sandbox) — **FIXED**: uv dropped from `_MCP_SANDBOX_EXTRA_PACKAGES` (now `["mcp>=1.12.4"]`); the new integration test pins the list without uv; the image ships pinned binaries via `COPY --from=uv`.
- [x] **M2** (minor, `uv export` downloads a managed CPython in the build stage) — **FIXED**: export stage is now `python:3.11-slim` with `ENV UV_PYTHON_DOWNLOADS=0`; `--locked` retained.
- [x] **M4** (minor, package set in three places with no cross-validation) — **FIXED**: chained validation now covers every path — pyproject group == `SANDBOX_DIRECT_REQUIREMENTS` (tests/test_docker_workflow_versions.py:255-257, 310), dict keys == dict keys (:309), Dockerfile smoke imports == `SANDBOX_DISTRIBUTION_IMPORTS.values()` via AST set equality (:318), and the `uv export` test asserts each distribution is pinned (:338-339).
- [x] **M7** (minor, no PR-time CI exercises the sandbox uv pipeline) — **FIXED in substance**: a real `uv export --locked --only-group sandbox --no-emit-project` now runs in PR CI as a pytest subprocess test. The residual half (no PR-time `docker build` of the sandbox image) is carried as **V6**.
- [x] **M8** (minor, test name under-describes scope) — **FIXED**: renamed to `test_docker_readme_documents_backend_and_sandbox_lockfile_requirements`.
- [x] **M9** (minor, docs describe a two-image publish flow) — **FIXED**: the sandbox image is now covered in the Docker Files list, Publishing Images, First Time Setup, Docker Hub Repositories, and Automatic Publishing sections, and `publish.sh`'s scope is stated explicitly.
- [x] **M10** (minor, design observation) — **WAIVED**: the curated whitelist boundary remains intentional and documented in the README.

### New findings

**Minors**

- **V1** (minor) — tests/test_docker_workflow_versions.py:14-35: two parallel dicts. `SANDBOX_DIRECT_REQUIREMENTS` is a verbatim mirror of `[dependency-groups].sandbox` whose only consumers are self-consistency asserts, and `SANDBOX_DISTRIBUTION_IMPORTS` mirrors its keys; adding one sandbox dependency touches four files/five places (pyproject.toml, uv.lock, the Dockerfile.sandbox:58 smoke line, both dicts). Acknowledging b68d4560's dict-driven smoke enforcement, this is the *remaining* duplication — not a re-raise of the fixed drift issue. Consider a single `{dist: (spec, import_name)}` dict with import names derived via `dist.replace("-", "_")` (verified correct for all 9 entries).
- **V2** (minor) — sandboxed_tool_wrapper.py:38-42 plus sandboxed_mcp_tool_helper.py:36: runtime dependency-manager specs are a third untracked spec location beside the group — `pydantic>=2.0.0` vs the group's `pydantic>=2.11.7` floor, and `mcp>=1.12.4` — with no test tying them to the group (the integration test imports the wrapper dict as its own truth, a tautology). Corrected mechanism: the sandbox container runs pip as **root** (docker_sandbox.py:1117-1118, "Run as root to match the file access behavior of Boxlite"), so a future floor raise would install into **system site-packages, replacing the pinned version** — not unprivileged `~/.local` shadowing. Inert today because the pinned image satisfies the floors, but the runtime channel can bypass uv.lock.
- **V3** (minor) — docker/Dockerfile.sandbox:30: `COPY --from=uv /uv /uvx /usr/local/bin/` is now load-bearing (sole uvx source after the `uv>=0.8.0` removal) but no test asserts it for the sandbox (the backend has `test_backend_runtime_keeps_uv_binaries` at tests/test_docker_workflow_versions.py:203-206), and the smoke check at :58 imports only Python packages. Consider mirroring the backend assert plus `uvx --version` in the smoke check.
- **V4** (minor) — tests/test_docker_workflow_versions.py:321-339: the `uv export` subprocess test has no `shutil.which("uv")` guard (local devs without uv get a FileNotFoundError instead of a skip), and CI installs uv unpinned (ci.yml:255) while the image pins `UV_VERSION=0.11.10` — the tested export semantics can drift from the image's.
- **V5** (minor) — sandboxed_mcp_tool_helper.py:36: removing the runtime uv injection couples uvx availability to the sandbox image. Custom `SANDBOX_IMAGE` overrides (config.py:2053, env-overridable) not derived from the published image previously worked (uv was injected at first MCP use) and now fail inside mcp_runner with a raw "uvx: command not found" (PATH-only resolution). Both built-in paths are safe (new image ships binaries; pinned 0.6.5 tag has pip-installed uvx). This is a new residual of the M1 fix. Consider documenting that custom sandbox images must ship uv/uvx, or a targeted early error when a uvx command is requested and uvx is absent.
- **V6** (minor) — the sandbox image is never built in PR-time CI (sandbox-publish.yml triggers on tag push / workflow_dispatch only; nightly and docker-publish build backend + frontend only). PR-time coverage is string tests plus the uv export subprocess, so a Dockerfile regression invisible to regexes (RUN syntax error, smoke typo, missing arch wheel) stays green until a tag push publishes. This is the carry-forward of M7's residual half — the uv-export half is now fixed via pytest (author resolved that thread); the docker-build half remains open and is tracked here in the body, since the author already resolved the M7 thread. The repo-wide convention of no PR-time image builds is noted; minor is right.

**Nits**

- **V7** (nit) — pyproject.toml:178: `fsspec>=2024.0.0` is a CalVer bound with no such release (2023.12.2 → 2024.2.0). Satisfiable, purely cosmetic; the same odd bound pre-exists at :77 and the group mirrors it.
- **V8** (nit) — docker/README.md:368: "Update `uv.lock` whenever that group changes" omits the command (the backend section documents `uv sync --locked`); the sandbox compose overlays still pin `xprobe/xagent-sandbox:0.6.5` (the image without fsspec; docker/docker-compose.sandbox.docker.yml:7, docker/docker-compose.sandbox.boxlite.yml:6), and the "update deployments" note at :383-384 names no files. Suggest "Run `uv lock`" plus pointing at the compose pins.
- **V9** (nit) — PR description: "15 passed" is stale; the file now defines 19 tests (13 base + 6 new). PR-description claim only.
- **V10** (nit) — floating base tags (`python:3.11-slim`, `node:22-slim`) sit beside the pinned `UV_VERSION=0.11.10`; this matches the repo-wide pattern (Dockerfile.backend floats too). Note only.

- L14 (yagni): `SANDBOX_DISTRIBUTION_IMPORTS = {d: d.replace("-", "_") for d in SANDBOX_DIRECT_REQUIREMENTS}` replaces the 11-line literal; drop the now-tautological L309 keys-equality assert. (-14)
- L289 (yagni): drop the three absence asserts (`--frozen` / `uv python install` / `apt-get install` not in the requirements stage) — the positive asserts already pin those spellings; no stage-contract coverage is lost. (Weakest leg: `apt-get` has no positive counterpart, but apt in an export-only stage is pointless anyway.) (-3)
- Dropped during verification: the ast.walk shrink (the AST set-equality at :318 is behavior-preserving and substring checks are NOT equivalent — `pydantic` ⊂ `pydantic_settings` collisions would pass) and the export-stage base image (`uv export --locked` discovers an interpreter unconditionally; the python base is load-bearing for deterministic marker evaluation).
- **Net: ≈ -17 lines.**

No confirmed critical or major findings: all prior findings are FIXED or WAIVED, and the new findings are minor/nit with no correctness or security impact on the built image. The minor items are maintainability/observability suggestions for follow-up, not merge gates.

tests/test_docker_workflow_versions.py
bsbds: We are not making this change because deriving Python import names from distribution names is not generally safe: this repository declares `python-docx` but imports it as `docx` ([dependency declaration](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/pyproject.toml#L118), [corresponding import](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/src/xagent/core/tools/core/workspace_file_tool.py#L130)).

The explicit distribution-to-import mapping is intentional coverage from the previously accepted smoke-import finding, and the current test preserves that contract directly ([prior coverage thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783004406), [current explicit mapping](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/tests/test_docker_workflow_versions.py#L11-L35)). Consolidating the two test dictionaries would not remove the independent pyproject, lockfile, and Docker smoke surfaces, so the remaining benefit does not justify another change.

tests/test_docker_workflow_versions.py
bsbds: We are not making this change because the functional test intentionally requires the project-native `uv` executable instead of skipping when that prerequisite is absent ([functional export test](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/tests/test_docker_workflow_versions.py#L366-L386), [documented uv-based development setup](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/AGENTS.md#L218-L229)).

The image build independently pins and exercises uv 0.11.10, while this test verifies that the current repository lock can be exported; forcing CI, development dependencies, and the lockfile to share the image bootstrap version would broaden the change without fixing a demonstrated failure ([Docker uv pin](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/docker/Dockerfile.sandbox#L2-L4), [approved evaluated-head review reporting passing checks](https://github.com/xorbitsai/xagent/pull/1355#pullrequestreview-4943441764)).

pyproject.toml
bsbds: We are not making this cosmetic change because `fsspec>=2024.0.0` is already the repository-established constraint and the sandbox group intentionally mirrors it ([existing project bound](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/pyproject.toml#L77), [sandbox-group bound](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/pyproject.toml#L178)).

The constraint resolves successfully to fsspec 2026.2.0 and is recorded in the committed lock, while the image export enforces lock synchronization ([resolved lock entry](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/uv.lock#L1909-L1912), [sandbox lock metadata](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/uv.lock#L8146), [locked export](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/docker/Dockerfile.sandbox#L19-L20)). Changing the lower-bound spelling would add lockfile churn without changing resolved behavior.

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/sandbox-locked-dependencies`
- **Head:** `ad45954339a67d784a418db2f5f4e6c013431067`
- **Destination:** `bsbds/xagent:fix/sandbox-locked-dependencies`
- **Result:** success
- **PR description:** updated and confirmed

### Resolved

1. **Review body:** [`PRR_kwDORQ4Kr88AAAABJqbvZA`](https://github.com/xorbitsai/xagent/pull/1355#pullrequestreview-4943441764)
   **Local outcome:** resolved
   **Change:** PR description updated from `15 passed` to `19 passed`
   **GitHub:** description update confirmed

2. **Thread:** [`PRRT_kwDORQ4Kr86Zfcnv`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019137)
   **Local outcome:** resolved
   **Change:** `b96d2c7142a2b1f44397308b46022106241bdd94`
   **GitHub:** resolved

3. **Thread:** [`PRRT_kwDORQ4Kr86Zfcnx`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019139)
   **Local outcome:** resolved
   **Change:** `b96d2c7142a2b1f44397308b46022106241bdd94`
   **GitHub:** resolved

4. **Thread:** [`PRRT_kwDORQ4Kr86Zfcn1`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019143)
   **Local outcome:** resolved
   **Change:** `ad45954339a67d784a418db2f5f4e6c013431067`
   **GitHub:** resolved

5. **Thread:** [`PRRT_kwDORQ4Kr86Zfcn4`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019146)
   **Local outcome:** resolved
   **Change:** `ad45954339a67d784a418db2f5f4e6c013431067`
   **GitHub:** resolved

6. **Thread:** [`PRRT_kwDORQ4Kr86Zfcn5`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019148)
   **Local outcome:** resolved
   **Change:** `b96d2c7142a2b1f44397308b46022106241bdd94`
   **GitHub:** resolved

### Rejected

1. **Thread:** [`PRRT_kwDORQ4Kr86Zfcnu`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019135)
   **Local outcome:** rejected
   **Reason:** Generic distribution-to-import derivation is unsafe, and the explicit mapping implements the previously accepted smoke-import contract. Consolidating the test dictionaries would not eliminate the independent production surfaces.
   **Sources:** [dependency declaration](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/pyproject.toml#L118), [corresponding import](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/src/xagent/core/tools/core/workspace_file_tool.py#L130), [prior coverage thread](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3783004406), [current explicit mapping](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/tests/test_docker_workflow_versions.py#L11-L35)
   **Reply:** [posted](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789913946)
   **GitHub:** left unresolved

2. **Thread:** [`PRRT_kwDORQ4Kr86Zfcnz`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019141)
   **Local outcome:** rejected
   **Reason:** The functional test intentionally requires project-native uv. The image independently pins and exercises uv 0.11.10; aligning CI, development dependencies, and the lockfile would broaden scope without fixing a demonstrated failure, while skipping would hide a missing prerequisite.
   **Sources:** [functional export test](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/tests/test_docker_workflow_versions.py#L366-L386), [development setup](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/AGENTS.md#L218-L229), [Docker uv pin](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/docker/Dockerfile.sandbox#L2-L4), [approved evaluated-head review](https://github.com/xorbitsai/xagent/pull/1355#pullrequestreview-4943441764)
   **Reply:** [posted](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789913971)
   **GitHub:** left unresolved

3. **Thread:** [`PRRT_kwDORQ4Kr86Zfcn3`](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789019145)
   **Local outcome:** rejected
   **Reason:** The valid, satisfiable bound intentionally matches the established project constraint and resolves to fsspec 2026.2.0. Changing its spelling would create lockfile churn without changing behavior.
   **Sources:** [existing project bound](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/pyproject.toml#L77), [sandbox-group bound](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/pyproject.toml#L178), [resolved lock entry](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/uv.lock#L1909-L1912), [sandbox lock metadata](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/uv.lock#L8146), [locked export](https://github.com/xorbitsai/xagent/blob/ad45954339a67d784a418db2f5f4e6c013431067/docker/Dockerfile.sandbox#L19-L20)
   **Reply:** [posted](https://github.com/xorbitsai/xagent/pull/1355#discussion_r3789913991)
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 6
- **Rejected:** 3
- **Unresolved:** 0

PR #1356, opened 2026-08-14
enh: humanize agent trace event labels for end users

yiboyasss:
## Summary
- Replace developer-facing trace labels ("LLM Call Start", "Tool Call Completed", "Execute Tool: web_search", ...) with plain, non-technical status text in both `en.ts` and `zh.ts`.
- Add a `traceEventRenderer.toolNames` friendly-name map for common built-in tools (web_search, python_executor, browser_use, file_tool, excel, create_agent, ...); tools not in the map fall back to a prettified version of the raw name (`some_tool_name` -> `Some Tool Name`) instead of the bare identifier.
- Wire the new tool-name resolution into both the collapsible trace detail (`TraceEventRenderer.tsx`) and the live status line shown above it (`ChatMessage.tsx`), so a running task now reads e.g. "Searching the web" instead of "Tool Call Start".
- Surface the model's own progress narration (from #1357's `send_message` work) directly in that same status line instead of a generic action label — a live smoke test showed the narration was otherwise only visible three levels deep inside the collapsed trace detail, defeating its own purpose. It keeps showing through a later, un-narrated tool call rather than getting bumped back to a generic label.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `trace-event-processing.test.ts` (pure reducer unit tests) passes unchanged
- [x] `ChatMessage.test.tsx` — added coverage for narration surfacing (shown, persists through a later un-narrated tool call, and a pending question is correctly excluded from being treated as narration)
- [x] `eslint` shows no new violations vs. the pre-existing baseline
- [x] Verified new i18n strings resolve correctly for both locales via the real translation resolver, including the unmapped-tool fallback path


frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -256,6 +256,30 @@ const getWaitingQuestionFromEvents = (events: TraceEvent[]): string | null => {
...
+// The i18n toolNames table (i18n/locales/*.ts -> traceEventRenderer.toolNames)
+// holds friendly phrasing ("Searching the web") for the common built-in
+// tools; anything not listed there degrades to the prettified raw name
+// instead of surfacing the identifier verbatim.
+export function getFriendlyToolName(toolName: string, tDynamic?: TranslateDynamic): string {
+  if (!toolName) return toolName;
+  const fallback = prettifyToolName(toolName);
+  if (!tDynamic) return fallback;
+  return tDynamic(`traceEventRenderer.toolNames.${toolName}`, fallback);
+}
```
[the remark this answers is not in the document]
yiboyasss: Fixed — `getFriendlyToolName` and the `tool_execution_start` extraction point in `TraceEventRenderer.tsx` now validate `tool_name` is actually a string before using it (it comes straight off an external WS payload; the `string` type is a compile-time annotation, not a runtime guarantee), instead of assuming the shape and letting a malformed payload throw inside `.replace()`. Added regression tests in `trace-event-processing.test.ts` covering non-string `tool_name` values (number/object/null) at both the reducer level and directly on `getFriendlyToolName`.

I scoped this to the code this PR introduces rather than a repo-wide boundary-normalization pass, since that would be a much larger change than this PR's actual scope (friendly trace labels).

frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -256,6 +256,33 @@ const getWaitingQuestionFromEvents = (events: TraceEvent[]): string | null => {
...
+
+// The i18n toolNames table (i18n/locales/*.ts -> traceEventRenderer.toolNames)
+// holds friendly phrasing ("Searching the web") for the common built-in
+// tools; anything not listed there degrades to the prettified raw name
+// instead of surfacing the identifier verbatim.
+export function getFriendlyToolName(toolName: string, tDynamic?: TranslateDynamic): string {
+  // toolName ultimately traces back to an external WS/API payload; the string
+  // type is a compile-time annotation only; not a runtime guarantee. Guard
+  // here too since this helper is exported and may be reused elsewhere.
+  if (typeof toolName !== 'string' || !toolName) return toolName;
```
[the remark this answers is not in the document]
yiboyasss: Fixed — returns `''` now instead of the raw non-string value, matching the declared return type.

frontend/src/components/chat/trace-event-processing.test.ts
```diff
@@ -234,3 +247,16 @@ describe("processTraceEvents tool_call_id attribution", () => {
...
+  it("prettifies an unmapped snake_case tool name", () => {
+    expect(getFriendlyToolName("some_future_tool")).toBe("Some Future Tool")
+  })
+
+  it("returns malformed (non-string) input as-is instead of throwing", () => {
+    expect(() => getFriendlyToolName(42 as never)).not.toThrow()
+    expect(() => getFriendlyToolName({ unexpected: "object" } as never)).not.toThrow()
+    expect(() => getFriendlyToolName(null as never)).not.toThrow()
+    expect(getFriendlyToolName(42 as never)).toBe(42)
+  })
```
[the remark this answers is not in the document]
yiboyasss: Updated the test to assert `''` per the fixed implementation.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -348,6 +348,17 @@ export function ChatMessage({
   const getEventTitle = (e: TraceEvent | undefined) => {
     if (!e) return "";
     const type = e.event_type || "";
+    // Tool events carry a specific tool name — prefer "Searching the web" over
+    // the generic "Working on it" fallback whenever we know which tool it is.
+    if (type === "tool_execution_start" || type === "tool_execution_end") {
+      const rawToolName =
+        (typeof e.data?.tool_name === "string" && e.data.tool_name) ||
+        (typeof e.data?.response?.tool_name === "string" && e.data.response.tool_name) ||
+        "";
```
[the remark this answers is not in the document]
yiboyasss: Fixed — normalized `e.data`/`e.data.response` once at the top instead of chaining optional access per field, and swapped the priority order to check `response.tool_name` before the top-level `tool_name`, matching `TraceEventRenderer.tsx`'s extraction order for the same event shape.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -348,6 +348,23 @@ export function ChatMessage({
...
+      const eventData = e.data || {};
+      const response = eventData.response || {};
+      const rawToolName =
+        (typeof response.tool_name === "string" && response.tool_name) ||
+        (typeof eventData.tool_name === "string" && eventData.tool_name) ||
+        "";
+      if (rawToolName) {
+        return getFriendlyToolName(rawToolName, tDynamic);
+      }
+    }
```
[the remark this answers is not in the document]
yiboyasss: Good catch on the UX bug — fixed. Restricted the tool-name override to `tool_execution_start` only; `tool_execution_end` now falls through to the generic "Done" label instead of continuing to name the finished tool as if it were still active.

Also added regression tests for both states in `ChatMessage.test.tsx`. That required switching the module mock to `vi.importActual` for `getFriendlyToolName` — the module was previously mocked out entirely, which meant this code path (and any crash in it) wasn't actually exercised by the existing tests.

On the payload normalization: this is effectively already what landed in the previous round (`const eventData = e.data || {}; const response = eventData.response || {};`, checked before any nested field access) — kept as-is.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +344,53 @@ export function ChatMessage({
   const hasTraceEvents = Array.isArray(traceEvents) && traceEvents.length > 0;
   const shouldShowProcess = !!showProcessView && hasTraceEvents;
 
+  // Mirrors TraceEventRenderer.tsx's isAgentProgressEvent — kept in sync by
+  // hand since the two files declare separate TraceEvent shapes.
+  const isProgressNarrationEvent = (e: TraceEvent): boolean => {
+    if (e.event_type === "agent_progress") return true;
+    if (e.event_type !== "agent_message") return false;
+    return e.data?.expect_response !== true && e.data?.message_type !== "question";
+  };
```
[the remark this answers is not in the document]
yiboyasss: Good catch — this is a real crash risk, not just style: traceEvents comes off an external WS/API payload, and a null/malformed entry anywhere in the array (not just the last one) would have thrown inside isProgressNarrationEvent's property access. Fixed by sanitizing traceEvents once at the top (filtering to well-formed objects) and reusing that sanitized list in every helper that reads from it, including the pre-existing trace-error scan that had the identical gap. Added a regression test mirroring TraceEventRenderer.test.tsx's malformed-entry coverage (null/number/string/bad-shaped entries mixed with a valid one).

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +344,62 @@ export function ChatMessage({
...
+
+  // The model's own narration ("I'll look into pricing now...") is more
+  // useful here than a generic action label, and it should keep showing
+  // even after a later, un-narrated tool call — the narration is still
+  // accurate until the model says something newer.
+  const latestProgressNarration = (): string => {
```
rogercloud, line 379: F2 [major]: latestProgressNarration() (lines 369-377) scans the whole sanitizedTraceEvents array backwards and returns the LAST narration anywhere in it, with no expiry — nothing resets it (not dag_step_start, tool_execution_end, or LLM-phase boundaries). Within one long running turn the status line keeps an early narration ("Checking pricing pages first") while the trace below advances to the current step — a visible mismatch lasting minutes. The PR description only asked to survive one later un-narrated tool call; the implementation is unbounded across step boundaries. Suggest an expiry condition (a dag_step_start boundary, or the corresponding tool_execution_end, clears/degrades the narration).

frontend/src/components/chat/TraceEventRenderer.test.tsx
```diff
@@ -147,7 +147,7 @@ describe("TraceEventRenderer", () => {
 
     fireEvent.click(
       screen.getByRole("button", {
-        name: /traceEventRenderer.executeTool:generate_image/,
+        name: /traceEventRenderer.executeTool:Generate Image/,
```
rogercloud, line 150: F3 [major]: this assertion (and its siblings, e.g. lines 441, 817) exercises only the prettify fallback, not the toolNames map: the tests mock useI18n WITHOUT tDynamic (mock at TraceEventRenderer.test.tsx:18-27), so getFriendlyToolName short-circuits to prettifyToolName — here asserting "Generate Image" while the real en.ts:3425 value for generate_image is "Creating an image". ChatMessage.test.tsx mocks tDynamic as (key) => key (lines 22-27), so its assertions (e.g. ChatMessage.test.tsx:463) only check the KEY string traceEventRenderer.toolNames.web_search. The 56-key toolNames map — the PR's core user-visible deliverable — thus has zero effective test coverage; emptying the whole map would leave the suite green, and the PR body's claim "Verified new i18n strings resolve correctly for both locales via the real translation resolver" corresponds to no committed test. (en/zh key parity is separately enforced by translations.test.ts:66-69; this finding is only about value coverage.) Suggest at least one test resolving a mapped and an unmapped name through the real resolver (en and zh), or direct assertions on getFriendlyToolName(name, realTDynamic).

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +344,62 @@ export function ChatMessage({
...
+  // that each entry is actually an object here, once, rather than repeating
+  // a null/type guard in every helper below that reads from the list.
+  const sanitizedTraceEvents: TraceEvent[] = Array.isArray(traceEvents)
+    ? traceEvents.filter(
+        (event): event is TraceEvent =>
+          !!event && typeof event === "object" && !Array.isArray(event),
+      )
+    : [];
+
+  // Mirrors TraceEventRenderer.tsx's isAgentProgressEvent — kept in sync by
```
rogercloud: F4 [major]: isProgressNarrationEvent (lines 357-363) hand-duplicates isAgentProgressEvent (TraceEventRenderer.tsx:286-293) under a "kept in sync by hand" comment, and the two copies have ALREADY diverged: line 373 uses `event.data?.message ?? event.data?.content` while TraceEventRenderer.tsx:496 uses `||`. A `{message: "", content: "text"}` agent_progress renders "text" in the trace detail but is skipped by the status line. ChatMessage already imports { TraceEventRenderer, getFriendlyToolName } from "./TraceEventRenderer" — export the predicate (and unify the ?? / || semantics) to remove the drift hazard.

frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -256,6 +256,33 @@ const getWaitingQuestionFromEvents = (events: TraceEvent[]): string | null => {
...
+// tools; anything not listed there degrades to the prettified raw name
+// instead of surfacing the identifier verbatim.
+export function getFriendlyToolName(toolName: string, tDynamic?: TranslateDynamic): string {
+  // toolName ultimately traces back to an external WS/API payload; the string
+  // type is a compile-time annotation only; not a runtime guarantee. Guard
+  // here too since this helper is exported and may be reused elsewhere.
+  if (typeof toolName !== 'string' || !toolName) return '';
+  const fallback = prettifyToolName(toolName);
+  if (!tDynamic) return fallback;
+  return tDynamic(`traceEventRenderer.toolNames.${toolName}`, fallback);
```
rogercloud, line 289: F6 [minor]: for unmapped tools — custom/user-created tools are by definition unmapped — this tDynamic call goes through the missing-key path, which fires console.warn("Missing translation key") (i18n-context.tsx:87-93) once per unique unmapped tool per session. Every dev session gets warnings for a designed first-class path. Suggest a silent lookup for this call (check the map directly or an onMissing opt-out).

frontend/src/i18n/locales/en.ts
```diff
@@ -3351,15 +3351,15 @@ Build when you need.`,
     showProcess: "Show process",
     hideProcess: "Hide process",
     unknownModel: "Unknown Model",
-    callLLM: "Call LLM: {model}",
-    llmResponse: "LLM Response",
+    callLLM: "Thinking",
```
rogercloud, line 3353: F7 [minor]: callLLM was rewritten to "Thinking" with no {model} placeholder, but TraceEventRenderer.tsx:459 still passes `{ model: eventData.model_name || t('traceEventRenderer.unknownModel') }` — the interpolation is discarded, and unknownModel ("Unknown Model", en.ts:3353) now has exactly one dead consumer: the expanded-detail path uses eventData.model_name with no fallback, so "Unknown Model" never renders anywhere. Suggest dropping the dead interpolation (keep data.model) or restoring the placeholder.

frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -606,7 +639,7 @@ export function processTraceEvents(
         step.actions.push({
           id: eventId,
           type: 'tool',
-          title: t('traceEventRenderer.executeTool', { tool: toolName }),
+          title: t('traceEventRenderer.executeTool', { tool: toolDisplayName }),
```
rogercloud, line 668: F9 [minor]: the start branch (this line) renders the friendly tool name, but the tool_execution_end branch (TraceEventRenderer.tsx:708-715, pre-existing) sets only status='completed' and never updates the title — completed tools in the trace detail (and in conversation-logs history) keep progressive titles like "Searching the web". Commit f1e8565f fixed the same issue on the status line (ChatMessage no longer names the tool at end), but the trace detail wasn't treated the same; the two surfaces are semantically inconsistent. Either accept as deliberate copy style or give completed actions a completed-form label.

frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -256,6 +256,33 @@ const getWaitingQuestionFromEvents = (events: TraceEvent[]): string | null => {
   return null;
 };
 
+// Raw tool identifiers (e.g. "python_executor") are internal names, not
+// user-facing copy. Turn any unmapped one into readable words as a safety net
+// so new/uncommon tools never fall back to a bare snake_case string.
+function prettifyToolName(name: string): string {
```
rogercloud, line 262: F11 [minor]: prettifyToolName only splits on [_-]: camelCase `webSearch` → "WebSearch" (glued), `tool.v2` → "Tool.v2", and all-separator names → '' — which then renders an EMPTY executeTool title (`{tool}` template, en.ts:3358) at TraceEventRenderer.tsx:642 and :1454. MCP/API tool names are camelCase-plausible (mcp_adapter preserves camelCase). Suggest camelCase splitting and/or falling back to a non-empty label when prettify yields ''.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +344,62 @@ export function ChatMessage({
   const hasTraceEvents = Array.isArray(traceEvents) && traceEvents.length > 0;
   const shouldShowProcess = !!showProcessView && hasTraceEvents;
 
+  // traceEvents comes straight off an external WS/API payload, so validate
+  // that each entry is actually an object here, once, rather than repeating
+  // a null/type guard in every helper below that reads from the list.
+  const sanitizedTraceEvents: TraceEvent[] = Array.isArray(traceEvents)
```
rogercloud, line 357: F12 [minor]: two sanitizers with different semantics now feed the same payload. This new filter (lines 350-355) checks only object-ness, while TraceEventRenderer.tsx:95-138 deep-normalizes (deletes non-string event_type, coerces non-object data to {}). Commit 95eeb2e4 claims "sanitize once at the boundary", but the renderer re-sanitizes downstream and the status line uses the weaker copy: an `event_type: 17` entry survives and renders a garbage status label (tDynamic falls back to the raw non-string value '17'). Extends the prior P6 thread: object-structure half fixed, field-type normalization (what the renderer's own sanitizer does) omitted. No crash — cosmetic garbage only.

frontend/src/components/pages/conversation-logs.test.tsx
```diff
@@ -253,7 +253,7 @@ describe("ConversationLogsPage", () => {
     // A completed task's trace is collapsed behind a toggle; expand it first.
     fireEvent.click(await screen.findByRole("button", { name: /showProcess/ }))
     expect(
-      await screen.findByRole("button", { name: /web_search/ })
+      await screen.findByRole("button", { name: /Web Search/ })
```
rogercloud, line 256: F13 [minor]: the prettify fallback is English-only — zh users with unmapped/custom tools see English Title-Case inside a Chinese UI (this zh-locale test now asserts "Web Search" at line 256, though that specific assertion comes from the mock's missing tDynamic). Arbitrary-name transliteration isn't feasible and the PR is still an improvement over raw snake_case — flag as an accepted-but-unstated decision, not a bug.

frontend/src/i18n/locales/en.ts
```diff
@@ -3064,44 +3064,44 @@ Build when you need.`,
...
+          dag_step_end: "Finished this step",
+          dag_step_failed: "This step ran into a problem",
+          llm_call_start: "Thinking",
+          llm_call_end: "Finished thinking",
+          llm_call_info: "Still thinking",
+          llm_call_result: "Got a response",
+          llm_call_failed: "Had trouble with that",
+          tool_execution_start: "Working on it",
+          tool_execution_end: "Done",
+          tool_execution_failed: "That didn't work",
```
rogercloud, line 3081: F14 [minor]: this PR rewrites two parallel namespaces for the same events with divergent phrasings: the status line reads agent.logs.event.actions.* (en.ts:3066-3105 — here tool_execution_failed = "That didn't work") while the trace detail reads traceEventRenderer.* (en.ts:3347+, executionFailed = "Ran into a problem"). Also agent.logs.event.messages.* remains developer-toned ("Start executing task (Iteration 1)", "Context compact start") and still renders into the user-visible chat stream via app-context-chat.tsx — the humanization boundary (labels yes, messages no) is never stated. Copy-consistency/UX-tone only.

frontend/src/components/chat/ChatMessage.test.tsx
```diff
@@ -442,4 +451,132 @@ describe("ChatMessage Session file capability", () => {
...
+      />,
+    )
+
+    expect(
+      screen.getByText(/Looking into LangChain and CrewAI pricing now\./),
+    ).toBeInTheDocument()
+    expect(screen.queryByText(/agent\.logs\.event\.actions/)).not.toBeInTheDocument()
+  })
+
+  it("keeps showing the last narration even after a newer, un-narrated tool call", () => {
```
rogercloud, line 538: F15 [minor]: narration state-machine coverage gaps around this test: (1) no test for a LATER narration replacing an earlier one — the last-wins core of the backward scan at ChatMessage.tsx:369-377, so a scan-direction regression would pass; (2) no test for the message_type === 'question' exclusion alone (only expect_response is tested at ChatMessage.test.tsx:558-574; deleting the disjunct at ChatMessage.tsx:362 would not fail the suite); (3) the "does not throw" malformed test (ChatMessage.test.tsx:484-509) never reaches getEventTitle because its fixture's narration hits first and `||` short-circuits at ChatMessage.tsx:412 — getEventTitle's guards are unexercised by it.

frontend/src/components/chat/trace-event-processing.test.ts
```diff
@@ -234,3 +247,16 @@ describe("processTraceEvents tool_call_id attribution", () => {
...
 })
+
+describe("getFriendlyToolName", () => {
+  it("prettifies an unmapped snake_case tool name", () => {
+    expect(getFriendlyToolName("some_future_tool")).toBe("Some Future Tool")
+  })
+
+  it("returns empty string for malformed (non-string) input instead of throwing", () => {
+    expect(() => getFriendlyToolName(42 as never)).not.toThrow()
+    expect(() => getFriendlyToolName({ unexpected: "object" } as never)).not.toThrow()
```
rogercloud, line 267: P3 [minor, prior partial]: toBe('') is asserted only for the number input (42, line 260); the object (line 258) and null (line 259) inputs assert only not.toThrow(). A regression returning non-empty garbage for object/null would pass the suite. Change the two remaining not.toThrow() to toBe('').

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -356,11 +408,13 @@ export function ChatMessage({
...
-  const latestTitle = getEventTitle(
-    Array.isArray(traceEvents) && traceEvents.length > 0
-      ? traceEvents[traceEvents.length - 1]
-      : undefined
-  );
+  const latestTitle =
+    latestProgressNarration() ||
+    getEventTitle(
+      sanitizedTraceEvents.length > 0
+        ? sanitizedTraceEvents[sanitizedTraceEvents.length - 1]
```
rogercloud: U2 [simplification]: `sanitizedTraceEvents.length > 0 ? sanitizedTraceEvents[sanitizedTraceEvents.length - 1] : undefined` (lines 414-417) → `sanitizedTraceEvents.at(-1)`. The lib target already supports esnext (.at(-1) precedent in files-disabled-presentation.ts:85,617).

rogercloud (changes requested):
## Summary

This PR humanizes the agent trace UI for end users: developer-facing status labels ("LLM Call Start", "Tool Call Completed", "Execute Tool: web_search") are replaced with plain copy in en.ts/zh.ts, a 56-key-per-locale `traceEventRenderer.toolNames` friendly-name map with a prettified-raw-name fallback (`some_tool_name` → `Some Tool Name`) is added and wired into both the trace detail (`processTraceEvents`, new optional `tDynamic` param) and the live status line (`ChatMessage.getEventTitle`), and the model's own progress narration (agent_progress / non-question agent_message) now surfaces in the status line, persisting through later un-narrated tool calls. 8 frontend files, +495/−129, 6 commits, CI fully green. The approach is sound and the rewrite is a clear improvement, but three major issues remain: narration has no expiry, the toolNames map has zero effective test coverage, and the narration predicate is hand-duplicated with the renderer's copy and has already diverged.

## Update since last review

The new commit 95eeb2e4 ("fix: sanitize traceEvents once at the boundary in ChatMessage") adds a `sanitizedTraceEvents` object-ness filter in ChatMessage.tsx (lines 350-355) that is now consumed by every traceEvents reader in the component, including the pre-existing trace-error scan, and adds a malformed-entry regression test in ChatMessage.test.tsx (lines 484-509). The commit is 44 insertions / 8 deletions across 2 files. The new sanitizer is weaker than the renderer's existing one — object-structure filtering only, no field-type normalization — which is tracked as new finding F12.

## Prior findings status

| Finding | Status | Notes |
|---|---|---|
| P1 — boundary normalization of malformed tool_name | FIXED | Guard at TraceEventRenderer.tsx:280, reducer-level typeof check :611-614, regression tests trace-event-processing.test.ts:71-81 + :256-261; commit 95eeb2e4 added the ChatMessage boundary sanitizer on top |
| P2 — getFriendlyToolName returns '' for non-string input | FIXED | TraceEventRenderer.tsx:280 |
| P3 — malformed-input test asserts toBe('') for number input only | PARTIAL | trace-event-processing.test.ts:256-261; object and null inputs assert only not.toThrow() — inline comment posted |
| P4 — response.tool_name before data.tool_name, e.data/e.data.response normalized | FIXED | ChatMessage.tsx:389-399 |
| P5 — tool_execution_end no longer names the active tool | FIXED | override restricted to tool_execution_start (ChatMessage.tsx:388); tests :457-468/:470-486 use vi.importActual so the real getFriendlyToolName runs |
| P6 — centralized sanitizer for traceEvents | FIXED (residual) | sanitizedTraceEvents at ChatMessage.tsx:350-355 consumed by narration scan, getEventTitle, trace-error scan; regression test :484-509. Residual: filter-only (object-ness), field types not normalized — see F12 |

## Findings

### Major

**F2 — [major] Narration priority has no expiry — ChatMessage.tsx:369-377 (inline at :369).** `latestProgressNarration()` scans the whole `sanitizedTraceEvents` array backwards and returns the last narration; nothing resets it — not `dag_step_start`, not `tool_execution_end`, not LLM-phase boundaries. Within one long running turn the status line keeps an early narration ("Checking pricing pages first") while the trace below advances to the current step — a visible mismatch lasting minutes. The PR description only asked to survive one later un-narrated tool call; the implementation is unbounded across step boundaries. Suggest an expiry condition (a `dag_step_start` boundary, or the corresponding `tool_execution_end`, clears/degrades the narration).

**F3 — [major] The toolNames map (the PR's core user-visible deliverable) has zero effective test coverage — TraceEventRenderer.test.tsx:150 (inline), also ChatMessage.test.tsx:22-27/:463.** ChatMessage.test.tsx mocks `tDynamic: (key) => key`, so its assertions check only the KEY string `traceEventRenderer.toolNames.web_search`; TraceEventRenderer.test.tsx and conversation-logs.test.tsx mock `useI18n` without `tDynamic`, so `getFriendlyToolName` short-circuits to the prettify fallback — e.g. TraceEventRenderer.test.tsx:150 asserts "Generate Image" while real en.ts:3425 says "Creating an image". Emptying the whole map would leave the suite green, and the PR body's test-plan claim ("Verified new i18n strings resolve correctly for both locales via the real translation resolver") corresponds to NO committed test. (The en/zh key-parity concern is already enforced by translations.test.ts:66-69 `assertTranslationTreeParity`; this finding is only about value coverage.) Suggest at least one test resolving mapped + unmapped names through the real resolver (en and zh), or direct assertions on `getFriendlyToolName(name, realTDynamic)`.

**F4 — [major] Hand-duplicated narration predicate has already diverged — ChatMessage.tsx:357-363 (inline at :357).** `isProgressNarrationEvent` duplicates `isAgentProgressEvent` (TraceEventRenderer.tsx:286-293) under a "kept in sync by hand" comment, and the copies already differ: ChatMessage.tsx:373 uses `event.data?.message ?? event.data?.content` while TraceEventRenderer.tsx:496 uses `||`. A `{message: "", content: "text"}` agent_progress renders "text" in the trace detail but is skipped by the status line. ChatMessage already imports `{ TraceEventRenderer, getFriendlyToolName }` from "./TraceEventRenderer" — export the predicate (and unify the ?? / || semantics) to remove the drift hazard.

### Minor

**F1 — [minor] No-step_id fallback path still concatenates the RAW tool name — frontend/src/contexts/app-context-chat.tsx:3390,3425,3460 (body only; file not in the PR diff).** The inline-message fallback renders `{t('agent.logs.event.actions.tool_execution_start')}: {toolName}`, so if it fires the UI shows "Working on it: web_search". The path is latent (no live backend emitter produces step_id-less tool events) and the raw concatenation pre-exists, but the PR humanized the same event types on the main path and left this fallback behind — route `toolName` through `getFriendlyToolName` there too.

**F6 — [minor] Designed fallback path triggers dev-mode missing-key warnings — TraceEventRenderer.tsx:283 (inline).** For unmapped tools (custom/user-created tools are by definition unmapped), the `tDynamic` call goes through the missing-key path, firing `console.warn("Missing translation key")` (i18n-context.tsx:87-93) once per unique unmapped tool per session. Suggest a silent lookup for this call (check the map directly or an onMissing opt-out).

**F7 — [minor] Dead `{model}` interpolation on callLLM — en.ts:3354 (inline); TraceEventRenderer.tsx:459.** The callLLM copy is now "Thinking" with no `{model}` placeholder, but TraceEventRenderer.tsx:459 still passes `{ model: eventData.model_name || t('traceEventRenderer.unknownModel') }`; the interpolation is discarded and `unknownModel` has exactly one dead consumer. The expanded-detail path uses `eventData.model_name` with no fallback, so "Unknown Model" never renders anywhere. Drop the dead interpolation (keep data.model) or restore the placeholder.

**F9 — [minor] tool_execution_end never updates the trace-detail title — TraceEventRenderer.tsx:708-715 (inline re-anchored at :642).** The end branch sets only `status='completed'` and never updates the title, so completed tools in the trace detail (and in conversation-logs history) keep progressive titles like "Searching the web". Commit f1e8565f fixed the same issue on the status line (ChatMessage no longer names the tool at end) but the trace detail wasn't treated the same — the two surfaces are semantically inconsistent. Either accept as deliberate copy style or give completed actions a completed-form label.

**F11 — [minor] prettifyToolName mishandles camelCase and all-separator names — TraceEventRenderer.tsx:262-270 (inline at :262).** It only splits on `[_-]`: camelCase `webSearch` → "WebSearch" (glued), `tool.v2` → "Tool.v2", and all-separator names → `''`, which then renders an EMPTY executeTool title (`{tool}` template, en.ts:3358) at :642/:1454. MCP/API tool names are camelCase-plausible (mcp_adapter preserves camelCase). Suggest camelCase splitting and/or falling back to a non-empty label when prettify yields `''`.

**F12 — [minor] Two sanitizers with different semantics feed the same payload — ChatMessage.tsx:350-355 (inline at :350).** The new filter checks only object-ness while TraceEventRenderer.tsx:95-138 deep-normalizes (deletes non-string event_type, coerces non-object data to `{}`). Commit 95eeb2e4 claims "sanitize once at the boundary", but the renderer re-sanitizes downstream and the status line uses the weaker copy: an `event_type: 17` entry survives and renders a garbage status label (tDynamic falls back to the raw non-string value '17'). This extends the prior P6 thread: object-structure half fixed, field-type normalization omitted. No crash — cosmetic garbage only.

**F13 — [minor] Prettify fallback is English-only — conversation-logs.test.tsx:256 (inline).** zh users with unmapped/custom tools see English Title-Case inside a Chinese UI (the zh-locale test now asserts "Web Search" at :256, though that specific assertion comes from the mock's missing tDynamic). Arbitrary-name transliteration isn't feasible and the PR is still an improvement over raw snake_case — flag as an accepted-but-unstated decision, not a bug.

**F14 — [minor] Dual parallel namespaces with divergent phrasings for the same events — en.ts:3081 (inline); also en.ts:3066-3105 vs 3347+.** The status line reads `agent.logs.event.actions.*` (here `tool_execution_failed` = "That didn't work") while the trace detail reads `traceEventRenderer.*` (`executionFailed` = "Ran into a problem"). Also `agent.logs.event.messages.*` remains developer-toned ("Start executing task (Iteration 1)", "Context compact start") and still renders into the user-visible chat stream via app-context-chat.tsx — the humanization boundary (labels yes, messages no) is never stated. Copy-consistency/UX-tone only.

**F15 — [minor] Narration state-machine test gaps — ChatMessage.test.tsx:538 (inline).** No test for a LATER narration replacing an earlier one (the last-wins core of the backward scan at ChatMessage.tsx:369-377 — a scan-direction regression would pass); no test for the `message_type === 'question'` exclusion alone (only `expect_response` is tested; deleting the disjunct at ChatMessage.tsx:362 would not fail the suite); and the "does not throw" malformed test (:484-509) never reaches getEventTitle because its fixture's narration hits first and `||` short-circuits at ChatMessage.tsx:412, so getEventTitle's guards are unexercised by it.

**P3 — [minor, prior partial] — trace-event-processing.test.ts:256-261 (inline at :258).** `toBe('')` is asserted only for the number input (42); object and null inputs assert only `not.toThrow()`. A regression returning non-empty garbage for object/null would pass. Change the two remaining `not.toThrow()` to `toBe('')`.

- L415: stdlib — `sanitizedTraceEvents.length > 0 ? sanitizedTraceEvents[sanitizedTraceEvents.length - 1] : undefined` (ChatMessage.tsx:414-417) → `sanitizedTraceEvents.at(-1)`. The lib target already supports esnext (`.at(-1)` precedent in files-disabled-presentation.ts:85,617).
- net: 1 stdlib simplification; no dead-flexibility, reinvention, or speculative abstraction found elsewhere — otherwise lean.

Blocking issues:
- ChatMessage.tsx:369-377 — [major] F2 — narration priority never expires; stale narration can persist for minutes across step boundaries. [new]
- TraceEventRenderer.test.tsx:150 / ChatMessage.test.tsx:22-27 — [major] F3 — the 56-key toolNames map has zero effective test coverage; the PR body's locale-resolution test claim is uncommitted. [new]
- ChatMessage.tsx:357-363 — [major] F4 — duplicated narration predicate already diverged from TraceEventRenderer.tsx:286-293 (?? vs ||). [new]

Recommended decision: **REQUEST_CHANGES** (blocking: yes). All three majors are small, well-scoped fixes: add an expiry condition for narration, commit one real-resolver coverage test for toolNames (en+zh, mapped+unmapped), and export/share the narration predicate. The minor findings (F1, F6, F7, F9, F11-F15, P3) can be addressed in the same pass.

frontend/src/components/chat/TraceEventRenderer.tsx
yiboyasss: Acknowledged, left as-is for now — fixing this cleanly needs a silent-lookup mode threaded through TranslateDynamic (or a direct map check bypassing the missing-key path), which is more plumbing than this PR's scope justifies for a dev-console-only warning. Happy to pick it up as a follow-up if it's actually noisy in practice.

frontend/src/components/chat/TraceEventRenderer.tsx
yiboyasss: Acknowledged, left as-is — this is a real inconsistency but it's a copy-style question (should a completed trace-detail entry keep its progressive title or switch to a completed-form one?) rather than a correctness bug, and I'd rather get your take on the intended behavior before changing the trace-detail rendering path.

frontend/src/components/pages/conversation-logs.test.tsx
yiboyasss: Agreed this is an accepted decision rather than a bug, as you noted — leaving as-is. Arbitrary tool-name transliteration isn't practical, and English Title-Case is still strictly better than raw snake_case for zh users.

frontend/src/i18n/locales/en.ts
yiboyasss: Agreed the humanization boundary (labels rewritten, messages.* left developer-toned) isn't stated anywhere — left as-is for this PR since unifying the two namespaces' phrasing is a larger copy pass than this PR's scope, but noted for a follow-up.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +350,71 @@ export function ChatMessage({
...
+  const latestProgressNarration = (): string => {
+    let toolStartsSinceNarration = 0;
+    for (let i = sanitizedTraceEvents.length - 1; i >= 0; i--) {
+      const event = sanitizedTraceEvents[i];
+      if (isAgentProgressEvent(event)) {
+        const message = getProgressNarrationText(event);
+        if (!message) continue;
+        return toolStartsSinceNarration <= 1 ? message : "";
+      }
+      if (event.event_type === "tool_execution_start") {
```
rogercloud, line 388: [minor] Narration-expiry counter design gaps (N2 + N4 merged) — ChatMessage.tsx:387-388.

The `toolStartsSinceNarration` counter increments only on `tool_execution_start`, which makes the expiry heuristic both under- and over-inclusive:

- Under-inclusive: after narration → tool_execution_start → tool_execution_failed the counter never advances, so the pre-failure narration keeps rendering through the failure aftermath — visible during mid-run retry windows (terminal failures replace the status line).
- Over-inclusive: two parallel tool_execution_start events expire a narration that describes both in-flight tools (concurrentSteps is special-cased elsewhere exactly because parallel tools are common), and delegated child steps' tool starts expire the manager's narration.
- Never counts: legacy `tool_using` events, so narration never expires for clients/backends emitting that type.

Worst case is a stale or prematurely-cleared status label, so minor — but the "survives one un-narrated tool call" contract would be better served by counting step boundaries (e.g. dag_step_start) or pairing tool_execution_failed/tool_using into the count.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +350,71 @@ export function ChatMessage({
...
+    // here would read as still in progress, so tool_execution_end falls
+    // through to the generic "Done" label below instead.
+    if (type === "tool_execution_start") {
+      // Normalize the (possibly missing) nested payload once, then read off
+      // it, rather than optional-chaining each field individually. Checks
+      // response.tool_name first to match the extraction order used for the
+      // same event shape in TraceEventRenderer.tsx.
+      const eventData = e.data || {};
+      const response = eventData.response || {};
+      const rawToolName =
```
rogercloud: [minor] Duplicated tool-name extraction with divergent coercion — ChatMessage.tsx:410-413 vs TraceEventRenderer.tsx:626-627.

This re-implements the response.tool_name → data.tool_name extraction TraceEventRenderer already does, but with different semantics: per-field typeof-guarded fall-through here (a non-string response.tool_name falls through to data.tool_name) vs a `||`-candidate chain there (a truthy non-string response.tool_name wins, then gets dropped by the typeof guard at :628). On `{data:{response:{tool_name:42},tool_name:'web_search'}}` the status line shows "Searching the web" while the trace detail shows the generic label — the exact cross-surface divergence the PR's shared getProgressNarrationText was created for. Low plausibility (no backend emits response.tool_name in tool_execution_start); suggest exporting a shared getRawToolName(event) from TraceEventRenderer.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -356,11 +423,8 @@ export function ChatMessage({
     return action || t("traceEventRenderer.taskExecution");
   };
 
-  const latestTitle = getEventTitle(
-    Array.isArray(traceEvents) && traceEvents.length > 0
-      ? traceEvents[traceEvents.length - 1]
-      : undefined
-  );
+  const latestTitle =
```
rogercloud: [minor] Narration not gated on resolved process status — ChatMessage.tsx:426-427.

`latestTitle = latestProgressNarration() || getEventTitle(...)` ignores process status: a completed turn with empty content keeps the mid-turn narration as its terminal label instead of "Done" (common.statusDone) — live and in read-only share logs (showProcessView=true on both). Pre-PR this slot showed the last event's "Done" label, so the completed path changed behavior cosmetically. Suggest gating the narration on a non-completed resolvedProcessStatus (the stopped-without-answer path already has its own handling).

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -344,10 +350,71 @@ export function ChatMessage({
   const hasTraceEvents = Array.isArray(traceEvents) && traceEvents.length > 0;
   const shouldShowProcess = !!showProcessView && hasTraceEvents;
 
+  // traceEvents comes straight off an external WS/API payload, so validate
+  // that each entry is actually an object here, once, rather than repeating
+  // a null/type guard in every helper below that reads from the list.
+  const sanitizedTraceEvents: TraceEvent[] = Array.isArray(traceEvents)
```
rogercloud, line 357: [minor] "Sanitize once at the boundary" overclaims — ChatMessage.tsx:356-363.

`sanitizedTraceEvents` is an object-ness filter only; the component still passes the RAW `traceEvents` to `<TraceEventRenderer events={traceEvents}>` (ChatMessage.tsx:516-517), which re-sanitizes with its stronger normalizer (TraceEventRenderer.tsx:95-138: data coercion, step_id/timestamp/field-type deletion) at two points, and the status line reads the weaker copy. No user-visible divergence is reachable today (every ChatMessage reader is individually type-guarded), so this is maintainability-only — but commit 95eeb2e4's title ("sanitize traceEvents once at the boundary") describes a single sanitizer that doesn't exist. Either pass sanitizedTraceEvents to the renderer or soften the claim.

frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -819,10 +867,10 @@ export function processTraceEvents(
 }
 
 function useProcessedSteps(events: TraceEvent[], taskStatus?: string): ProcessedStep[] {
-  const { t } = useI18n();
+  const { t, tDynamic } = useI18n();
   return useMemo(
-    () => processTraceEvents(events, t, taskStatus),
-    [events, taskStatus, t],
+    () => processTraceEvents(events, t, taskStatus, tDynamic),
```
rogercloud, line 883: [minor] tDynamic → processTraceEvents → getFriendlyToolName wiring unpinned — TraceEventRenderer.tsx:872.

Every component test mocks useI18n without tDynamic, so the prettify-fallback branch runs in all of them, and the real-resolver tests call getFriendlyToolName directly — deleting the tDynamic argument at :872 would leave the suite green while silently dropping the curated phrases from the trace detail. Failure mode is cosmetic (prettified name instead of curated phrase), but this wiring is the PR's core user-visible deliverable; one real-resolver test asserting a mapped tool renders its curated phrase would pin it.

frontend/src/components/chat/TraceEventRenderer.tsx
```diff
@@ -265,13 +298,23 @@ const isAgentProgressEvent = (event: TraceEvent): boolean => (
   )
 );
 
+// Shared with ChatMessage.tsx's status line so both surfaces agree on what
+// counts as narration text for the same event — a hand-duplicated copy
+// there previously used ?? instead of ||, silently disagreeing with this
+// file on events shaped like {message: "", content: "..."}.
+export function getProgressNarrationText(event: TraceEvent): string {
+  const raw = event.data?.message || event.data?.content;
```
rogercloud, line 321: [minor] `||` fallthrough in getProgressNarrationText unpinned — TraceEventRenderer.tsx:306.

The `message || content` (not `??`) fallthrough — the exact semantic this shared function exists to guarantee after the F4 divergence — has no test. A regression back to `??` would silently re-break `{message:'', content:'text'}` events. The current backend cannot produce that shape for narration-relevant event types, so coverage-only; a single assertion with an empty-string message would pin it.

frontend/src/components/chat/trace-event-processing.test.ts
```diff
@@ -234,3 +249,42 @@ describe("processTraceEvents tool_call_id attribution", () => {
...
+// The mocked useI18n() above (t: key => key) means every other test in this
+// file exercises processTraceEvents' own logic, not what the toolNames map
+// actually resolves to — deleting the whole map would leave them green. Use
+// the real resolver here so a regression in en.ts/zh.ts's toolNames entries,
+// or in getFriendlyToolName's lookup path, actually fails a test.
+describe("getFriendlyToolName against the real translation trees", () => {
+  const realTDynamic = (locale: Locale): TranslateDynamic => (key, fallback) =>
+    resolveDynamicTranslation(locale, key, fallback)
+
+  it("resolves a mapped tool name to its curated phrase in both locales", () => {
```
rogercloud, line 284: [minor] Locale parity enforced for ONE key only — trace-event-processing.test.ts:277-279.

The real-resolver tests pin only web_search; the other 55 toolNames entries are unpinned against zh, and the dynamic-key lookup bypasses TranslationKey typing — a future en-only key would silently degrade zh users to the English prettified name. Sets are in sync today (56/56). A cheap exhaustive test iterating Object.keys of en's toolNames and asserting each resolves in zh would close the gap.

frontend/src/components/chat/ChatMessage.tsx
```diff
@@ -356,11 +423,8 @@ export function ChatMessage({
     return action || t("traceEventRenderer.taskExecution");
   };
 
-  const latestTitle = getEventTitle(
-    Array.isArray(traceEvents) && traceEvents.length > 0
-      ? traceEvents[traceEvents.length - 1]
-      : undefined
-  );
+  const latestTitle =
+    latestProgressNarration() || getEventTitle(sanitizedTraceEvents.at(-1));
```
rogercloud: [minor] Unmapped event types get no prettifier — ChatMessage.tsx:421 (getEventTitle fallback).

`tDynamic(key, action || type)` renders the raw snake_case event_type for unmapped types, and this is live today: the backend emits react_step_start/react_step_end (ws_trace_handlers.py:178,184) while translations only define step_start_react/step_end_react, so those real events render developer labels in the status line. Tool names got the prettify fallback but event types didn't; consider routing the event-type fallback through prettifyToolName (or adding the react_* keys) for consistency with the PR's humanization goal.

rogercloud (approved):
## Summary

This PR humanizes xagent's agent trace UI for end users. The chat status line and trace detail previously leaked developer vocabulary ("LLM Call Start", "Tool Call Completed", "Execute Tool: web_search"), and the model's progress narration from #1357 was buried in the collapsed trace detail. The PR relabels the copy in both locales (en.ts/zh.ts), maps built-in tool ids to curated friendly phrases through a 56-key-per-locale `traceEventRenderer.toolNames` map with a prettified-raw-name fallback (`some_tool_name` → "Some Tool Name"), and surfaces the model's narration in the status line, keeping it through one later un-narrated tool call. 9 files changed, +657/−139, CI green. The raw-vs-display name split is sound, the cutover of removed keys is clean, and the real-translation-tree test strategy is unusually good — what remains is minor.

## Update since last review

One new commit since the previous review: 74b78d8a ("fix: address rogercloud's PR review on progress narration and toolNames"). It adds the tool-start-count narration expiry (with three pinned tests), exports `isAgentProgressEvent` and shares `getProgressNarrationText` between ChatMessage and TraceEventRenderer (the hand-duplicated predicate is gone), adds real-translation-tree tests for `getFriendlyToolName`, fixes prettifyToolName's camelCase/all-separator handling, removes the dead `unknownModel` interpolation, blanks the garbage `event_type: 17` label, and applies the `.at(-1)` simplification. +190/−38 over 7 files.

## Prior findings status

| Finding | Status | Notes |
|---|---|---|
| F2 — narration priority has no expiry (major) | FIXED (residual → minor) | Tool-start-count expiry added and test-pinned (ChatMessage.tsx:379-393); residual: tool-less phases (dag/LLM boundaries) don't expire — folded into N2/N4 |
| F3 — toolNames map has zero effective coverage (major) | FIXED | Real-resolver tests against the actual translation trees (trace-event-processing.test.ts:273-290); emptying the map would now fail the suite |
| F4 — duplicated narration predicate (major) | FIXED | `isAgentProgressEvent` exported and `getProgressNarrationText` shared (TraceEventRenderer.tsx:296-308); zero duplication remains |
| F1 — app-context-chat.tsx raw-name fallback (minor) | NOT FIXED | Pre-existing latent path, file untouched; the humanization boundary stops at ChatMessage/TraceEventRenderer |
| F6 — dev-mode missing-key warnings (minor) | WAIVED | Acknowledged + deferred (comment 3784364447) |
| F7 — dead `{model}` interpolation (minor) | FIXED | `unknownModel` removed from both locales, zero refs; callLLM call simplified |
| F9 — tool_execution_end trace-detail title (minor) | WAIVED | Acknowledged; copy-style question (comment 3784364669) |
| F11 — prettifyToolName camelCase/all-separator (minor) | FIXED | camelCase split + all-separator raw fallback; nit: new branches unpinned by tests |
| F12 — two sanitizers, divergent semantics (minor) | FIXED | `event_type: 17` garbage-label case closed (blanking added); structural divergence remains — folded into N5 |
| F13 — English prettify fallback (minor) | WAIVED | Accepted decision, now pinned by the zh fallback test (comment 3784364953) |
| F14 — dual parallel namespaces (minor) | WAIVED | Acknowledged + deferred to follow-up (comment 3784384311) |
| F15 — narration state-machine test gaps (minor) | FIXED | All three gaps pinned (last-wins, message_type-only exclusion, malformed tail reaches getEventTitle) |
| P3 — malformed-input assertions (minor) | FIXED | `toBe('')` now asserted for number/object/null |
| U2 — `.at(-1)` (simplification) | FIXED | Applied (ChatMessage.tsx:427) |

## New findings (all minor)

**N1 — duplicated tool-name extraction, divergent coercion — ChatMessage.tsx:410 (inline).** The status line hand-duplicates the response.tool_name/data.tool_name extraction from TraceEventRenderer.tsx:626-627 with different semantics: per-field typeof-guarded fall-through here vs a `||`-candidate chain there. On `{data:{response:{tool_name:42},tool_name:'web_search'}}` the status line shows "Searching the web" while the trace detail shows the generic label. Low plausibility (no backend emits response.tool_name in tool_execution_start); suggest exporting a shared `getRawToolName(event)`.

**N2+N4 — narration-expiry counter design gaps — ChatMessage.tsx:387 (inline, merged).** The expiry counter only increments on tool_execution_start: failures never advance it (pre-failure narration outlives the failure aftermath in mid-run retry windows), parallel tool starts over-expire a narration that describes both in-flight tools, delegated child-step starts cross-expire the manager's narration, and legacy tool_using events never count at all.

**N3 — narration not gated on process status — ChatMessage.tsx:426 (inline).** A completed turn with empty content keeps the mid-turn narration as its terminal label instead of "Done" — live and in read-only share logs (showProcessView=true on both). Pre-PR this slot showed the last event's "Done" label.

**N5 — "sanitize once at the boundary" overclaims — ChatMessage.tsx:356 (inline).** ChatMessage passes raw traceEvents to TraceEventRenderer, which re-sanitizes with its stronger normalizer at two points; the status line reads the weaker filter-only copy. No user-visible divergence reachable today — maintainability-only, but commit 95eeb2e4's title describes a boundary that doesn't exist.

**N6 — tDynamic wiring unpinned — TraceEventRenderer.tsx:872 (inline).** Every component test mocks useI18n without tDynamic (fallback branch) and the real-resolver tests call getFriendlyToolName directly; deleting the tDynamic argument would leave the suite green while silently dropping the curated phrases. Cosmetic failure mode only.

**N7 — `||` fallthrough unpinned — TraceEventRenderer.tsx:306 (inline).** The `message || content` (not `??`) fallthrough that getProgressNarrationText exists to guarantee has no test; a regression back to `??` would silently re-break `{message:'', content:'text'}` events. Coverage-only — the current backend cannot produce that shape.

**N8 — locale parity enforced for one key only — trace-event-processing.test.ts:277 (inline).** Only web_search is pinned across locales; the other 55 toolNames entries are unpinned and the dynamic-key lookup bypasses TranslationKey typing. Sets are in sync today (56/56); an exhaustive en→zh parity test would close it.

**N9 — unmapped event types get no prettifier — ChatMessage.tsx:421 (inline).** The getEventTitle fallback renders raw snake_case for unmapped types — live today, since the backend emits react_step_start/react_step_end (ws_trace_handlers.py:178,184) while translations only define step_start_react/step_end_react. Consider routing the event-type fallback through prettifyToolName.

**N10 — O(n) narration scan per render — ChatMessage.tsx:426 (body note only).** sanitizedTraceEvents reallocates and latestProgressNarration() backward-scans on every render — O(n) per streaming event. Negligible at current sizes (sub-millisecond cumulative); useMemo would help little since the parent rebuilds event arrays per event. Note-level only.

## Simplification

Lean already — both simplification-lens candidates were dropped in verification (N11, N12); no over-engineering survives in this diff.

All three prior majors are resolved: F2's residual degraded to minor (merged into N2/N4), F3 is fixed with real-resolver tests, F4 is fixed with the shared predicate. All new findings (N1–N10) are minor — maintainability, coverage, and copy-consistency — and none block merge. No critical or major finding remains.

Blocking: no — recommended event: **APPROVE**.

frontend/src/components/chat/ChatMessage.tsx
yiboyasss: Acknowledged, left as-is for now — these are genuine gaps but each needs a product call on the intended behavior (should a failed tool's retry count toward expiry? should a parallel narration describing two in-flight tools survive both? should a delegated child's steps share the manager's expiry clock or have their own?) rather than a driveby heuristic tweak. Happy to take this on as a focused follow-up once there's a clear answer for each.

frontend/src/components/chat/ChatMessage.tsx
yiboyasss: Fair — the commit title overclaimed. There's no shared sanitization boundary; ChatMessage's filter is local and narrower than TraceEventRenderer's. Leaving the current (weaker but sufficient for the concrete garbage-title case it closed) filter as-is rather than chasing a wording fix after the fact, but noted for whenever the two get unified.

frontend/src/components/chat/TraceEventRenderer.tsx
yiboyasss: Left as-is — the direct getRawToolName/getProgressNarrationText/getFriendlyToolName unit tests added this round (against the real translation trees, not a mock) cover the semantics this was pointing at without needing to wire tDynamic through a full component render, which felt like more test-harness plumbing than the marginal coverage gain justified.

frontend/src/components/chat/ChatMessage.tsx
yiboyasss: Confirmed and it runs deeper than the visible symptom — the backend actually emits react_step_start/react_step_end (ws_trace_handlers.py:178,184), and app-context-chat.tsx (outside this PR's diff) has two branches comparing against the reversed step_start_react/step_end_react that, as far as I can tell, have never matched. Fixing just the i18n key wouldn't address that dead-looking logic, so I've filed this as its own follow-up rather than a driveby fix bundled into this PR.

PR #1363, opened 2026-08-14
fix(agent): require final deliverable file links

bsbds:
## Summary

- define one shared final-deliverable file-reference instruction
- apply it to ReAct, DAG, and Auto final-answer prompts and tool schemas
- require exact trusted `markdown_link` values while excluding intermediate and internal files
- add prompt-construction regression coverage for every execution mode

## Verification

- `pytest -q tests/core/test_file_ref.py tests/core/agent/test_react.py tests/core/agent/test_dag.py tests/core/agent/test_auto.py`
- `pre-commit run --all-files`

## Scope

This is a prompt-level mitigation. It does not add frontend fallback rendering, structured deliverable metadata, or runtime link insertion.


src/xagent/core/file_ref.py
```diff
@@ -5,7 +5,26 @@
...
 from urllib.parse import quote, unquote, urlsplit
 
-FILE_REF_OUTPUT_INSTRUCTIONS = """## FILE REFERENCE OUTPUTS
+FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS = """## FINAL DELIVERABLE FILE REFERENCES
+If a successful tool result or trusted public FileRef produced a file that
+satisfies the user's requested final deliverable, the final answer MUST include
+the exact markdown_link returned for that file.
+
+If a generated final deliverable exists but no trusted file UUID or markdown_link
+remains in the current context, call get_workspace_output_files once before finalizing.
```
rogercloud: **major — Unactionable recovery instruction.** This line tells the model to "call get_workspace_output_files once before finalizing", but at 3 of the 6 sites that embed this block the tool is not in the schema: the ReAct force-final-answer prompt (src/xagent/core/agent/pattern/react/react.py:828; the schema exposes only final_answer and the prompt says "Do not call any other tool"), the Auto routing decision (src/xagent/core/agent/pattern/auto/auto.py:1313; "Call exactly one routing tool" auto.py:1294, answer must be in the same tool call), and the DAG completion assessment (src/xagent/core/agent/pattern/dag/dag.py:1528; "Call the assessment tool exactly once" dag.py:1519). Attempting the call at these sites is a protocol error (react.py:519 unavailable_tool_call; auto.py:1222, MAX_DECISION_PARSE_ATTEMPTS=2), and "do the lookup earlier" is unavailable because the instruction first appears on the finalizing turn itself. Suggestion: parameterize the block like grounding_rule — e.g. FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS(can_lookup: bool) or a builder function; forced sites pass can_lookup=False and omit the lookup paragraph; keep it only where the tool schema is present (e.g. the has_tools schema description, react.py:1702). (Design note 8 — builder abstraction — is the same fix.)

src/xagent/core/file_ref.py
```diff
@@ -15,7 +34,9 @@
 - Internal FileRefs (`internal: true`) are runtime context only. Never render or expose a `file:` link for them.
 - Never invent, guess, or construct a file_id or `file:` link. Use only a trusted FileRef supplied for an existing input/attachment, or the exact FileRef or markdown_link returned by a successful tool result. A link mentioned only in prior assistant prose is not provenance.
 - When the user requests a new file or file-based artifact, it is not delivered until a successful tool result returns its registered FileRef or markdown_link.
-- Do not call final_answer claiming that a file was created or delivered unless that result exists."""
+- Do not call final_answer claiming that a file was created or delivered unless that result exists.
+
+{FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS}"""
```
rogercloud: **major — Out-of-scope global embedding + 2-3x duplication.** Embedding the block here (into FILE_REF_OUTPUT_INSTRUCTIONS, which line 51 embeds into FILE_REF_MODEL_INSTRUCTIONS) puts it on every LLM call: it rides ExecutionContext._system_context (execution.py:463-464) on every ReAct turn, Auto decision, and DAG step execution, plus the production uploaded-files prompts (websocket.py:5471, file_turn.py:365). Per final-answer call the block is duplicated — ReAct forced 3x (system context + react.py:828 + schema react.py:1702), Auto decision 3x (system context + auto.py:1313 + schema auto.py:1451), DAG assessment 2x — and the final-deliverable exclusivity sentence ("Include only user-requested final deliverables. Do not include intermediate...") now also reaches intermediate step prompts whose contract is to report step results, a direct conflict. This exceeds the PR's declared scope (final-answer prompts and tool schemas only). Suggestion: restore the plain string (ending at line 37, "unless that result exists."), keep only the six explicit insertion sites, and remove the corresponding `assert instruction in FILE_REF_OUTPUT_INSTRUCTIONS` pin in tests/core/test_file_ref.py.

src/xagent/core/file_ref.py
```diff
@@ -5,7 +5,26 @@
...
+satisfies the user's requested final deliverable, the final answer MUST include
+the exact markdown_link returned for that file.
+
+If a generated final deliverable exists but no trusted file UUID or markdown_link
+remains in the current context, call get_workspace_output_files once before finalizing.
+Match the exact deliverable path and filename, then use only a
+returned non-null file_id or markdown_link. If the lookup has no registered ID
+or link, do not repeat the lookup, invent a link, or claim delivery.
+
+Copy the markdown_link verbatim and preserve its original filename and extension.
```
rogercloud: **minor — Lookup wording gap.** The lookup fallback returns no markdown_link: get_workspace_output_files (workspace.py:1871-1892, workspace_file_tool.py:673-688) returns only file_id (possibly None) plus filename/extension. "Copy the markdown_link verbatim" reads as if the lookup returned one, so a model may under-claim delivery despite a valid returned id. (The requirement is still satisfiable: the pre-existing render rule at file_ref.py:29, "Files: [filename](file:file_id)", produces exactly the string build_file_ref emits as markdown_link, file_ref.py:137; line 16 marks the returned file_id as trusted.) Suggestion: add one clarifying sentence to the lookup paragraph, e.g. "render [filename](file:file_id) from the returned file_id — that rendered link is the exact markdown_link form."

src/xagent/core/file_ref.py
```diff
@@ -5,7 +5,26 @@
 from typing import Any
 from urllib.parse import quote, unquote, urlsplit
 
-FILE_REF_OUTPUT_INSTRUCTIONS = """## FILE REFERENCE OUTPUTS
+FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS = """## FINAL DELIVERABLE FILE REFERENCES
+If a successful tool result or trusted public FileRef produced a file that
+satisfies the user's requested final deliverable, the final answer MUST include
+the exact markdown_link returned for that file.
```
rogercloud: **minor — Conflict with computer-use screenshot guidance.** This mandate (include the exact markdown_link returned for that file) conflicts with the screenshot instructions at computer.py:262-263 and 471-476, which tell the model to include the returned inline_markdown verbatim in final_answer instead of file_ref.markdown_link — the screenshot result returns both fields. Both texts share the same model context on screenshot turns, so the model faces contradictory render instructions: duplicate references or dropped inline rendering (delivery integrity unaffected — same file id). Suggestion: exempt tool-prescribed renderings, e.g. "unless the tool result prescribes a different rendering for that artifact (e.g. inline_markdown for screenshots)."

src/xagent/core/file_ref.py
```diff
@@ -5,7 +5,26 @@
 from typing import Any
 from urllib.parse import quote, unquote, urlsplit
 
-FILE_REF_OUTPUT_INSTRUCTIONS = """## FILE REFERENCE OUTPUTS
+FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS = """## FINAL DELIVERABLE FILE REFERENCES
+If a successful tool result or trusted public FileRef produced a file that
```
rogercloud: **minor — Vocabulary drift.** "trusted public FileRef" matches no system vocabulary: the FileRef distinction is `internal: bool` (file_ref.py:122, 220), "public" elsewhere means public context (artifacts.py:196), and the parallel phrase "trusted file UUID" (line 13) is also wrong — the lookup returns file_id, an opaque string (workspace.py:288/616 show ids are not guaranteed UUIDs). Mismatched vocabulary weakens prompt compliance. Suggestion: use "non-internal FileRef" and "file_id".

src/xagent/core/file_ref.py
```diff
@@ -5,7 +5,26 @@
 from typing import Any
 from urllib.parse import quote, unquote, urlsplit
 
-FILE_REF_OUTPUT_INSTRUCTIONS = """## FILE REFERENCE OUTPUTS
+FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS = """## FINAL DELIVERABLE FILE REFERENCES
```
rogercloud: **minor — Heading interpolated into plain descriptions + wording redundancy.** The "## FINAL DELIVERABLE FILE REFERENCES" heading is baked into the shared constant, so it lands mid-sentence inside JSON-schema description fields (react.py:1701-1703, auto.py:1450-1452, dag.py:1560-1562) — an invalid ATX heading that reads as literal text — and at file_ref.py:39 it becomes a sibling ## section of its container (also ##). Separately, "Never invent, guess, shorten, or rewrite a file_id" (line 20) overlaps the pre-existing "Never invent, guess, or construct a file_id or `file:` link" (line 35) with complementary coverage. Cosmetic. Suggestion: strip the heading from the shared constant and add it at the interpolation sites, or accept ## as a pattern-level marker and drop the mid-sentence interpolation.

tests/core/agent/test_react.py
```diff
@@ -1351,6 +1351,7 @@ def test_react_grounding_rule_present_in_both_answer_paths() -> None:
         assert "illustrative placeholders" in prompt
     assert "use an appropriate tool to verify" in tool_prompt
     assert "use an appropriate tool" not in forced_prompt
+    assert "## FINAL DELIVERABLE FILE REFERENCES" in forced_prompt
```
rogercloud: **minor — Test cannot fail on the regression it guards.** This assertion checks the heading in forced_prompt, but the heading also arrives via the system context (FILE_REF_MODEL_INSTRUCTIONS → ExecutionContext._system_context, execution.py:463-464), so removing the react.py:828 insertion entirely would still leave this test green. The same shape repeats across the pattern tests: all assert only the heading marker, none assert a distinctive body phrase, and none assert count==1 or absence (the DAG test at test_dag.py:394 and the Auto test at test_auto.py:1095 are specific — good). Suggestion: assert a distinctive body phrase that exists only in the explicit insertion, e.g. "call get_workspace_output_files once before finalizing", on the forced path; optionally also assert it appears exactly once.

src/xagent/core/agent/pattern/auto/auto.py
```diff
@@ -1308,7 +1309,8 @@ def _decision_prompt(
             "when action is final_answer, you must include a complete non-empty "
             "answer field in the same tool call. Put action before answer in the "
             "tool arguments. "
-            f"When writing that answer field: {grounding_rule(can_call_tools=False)} "
+            f"When writing that answer field: {grounding_rule(can_call_tools=False)}\n\n"
+            f"{FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS}\n\n"
```
rogercloud: **P1 follow-up — NOT FIXED.** The block still sits between the grounding_rule line (auto.py:1312) and the "If the answer would need such unsupported specifics..." fallback (auto.py:1314-1316), interrupting the anaphoric flow flagged in the prior review. Honest reassessment: the concern was technically overstated — the antecedent of "such unsupported specifics" is internal to grounding_rule itself (grounding.py:26-35), and the inserted block introduces no competing referent, so the anaphora remains unambiguous. This is a readability improvement, not a broken reference, and is no longer blocking-grade. If you adopt it: move the block after the conditional fallback sentence (lines 1314-1316). If not, the concern can be dropped — your call.

rogercloud (changes requested):
**Follow-up review — PR #1363: fix/harden-final-deliverable-links** (head `4112557a`; prior review was against `50eeb918`)

## Summary

Final answers across the ReAct, DAG, and Auto patterns can omit deliverable links, rewrite file IDs, claim undelivered files, or leak intermediate/internal files. This PR adds one shared prompt-level instruction block requiring the exact trusted `markdown_link` for user-requested deliverables, with a `get_workspace_output_files` lookup fallback that fails closed, applied to all three patterns' final-answer prompts and tool schemas. Runtime enforcement is explicitly out of scope (prompt-level mitigation).

## Update since prior review

Since the prior review, the branch added one commit (`4112557a`) that extends the shared instruction block with the `get_workspace_output_files` lookup paragraph (+6 lines, src/xagent/core/file_ref.py:13-17) and 4 matching test assertions (tests/core/test_file_ref.py:27-30). The prior inline suggestion on src/xagent/core/agent/pattern/auto/auto.py (block placement) was not addressed — auto.py is unchanged. CI is green (Pytest fast web/integration/deepdoc, migrations, pre-commit, frontend-build — all SUCCESS).

## Prior-findings checklist

- **P1 — block placed between `grounding_rule` and the fallback (auto.py:1313) — NOT FIXED.** auto.py is byte-identical since the prior review. Honest reassessment: the concern was technically overstated — the antecedent of "such unsupported specifics" is internal to `grounding_rule` itself (grounding.py:26-35), and the inserted block introduces no competing referent, so the anaphora remains unambiguous. It is a readability improvement, not a broken reference, and is no longer blocking-grade. Either adopt the reordering (block after the fallback sentence, auto.py:1314-1316) or drop the concern — author's call.

## Findings

### Major

**1. Unactionable recovery instruction** — src/xagent/core/file_ref.py:13-14
The instruction "call get_workspace_output_files once before finalizing" is impossible at 3 of the 6 insertion sites: the ReAct force-final-answer prompt (react.py:828; the schema exposes only `final_answer` and the prompt says "Do not call any other tool"), the Auto routing decision (auto.py:1313; only routing tools are available — "Call exactly one routing tool" auto.py:1294 — and the answer must be in the same tool call), and the DAG completion assessment (dag.py:1528; "Call the assessment tool exactly once" dag.py:1519). `get_workspace_output_files` exists only in the normal tool list; attempting it at these sites is a protocol error (react.py:519 `unavailable_tool_call`; auto.py:1222, MAX_DECISION_PARSE_ATTEMPTS=2). "Do the lookup earlier" is not available either: the instruction first appears on the finalizing turn itself.
Suggestion: parameterize the block like `grounding_rule` — e.g. `FINAL_DELIVERABLE_FILE_REFERENCE_INSTRUCTIONS(can_lookup: bool)` or a builder function — with forced sites passing `can_lookup=False` (omit the lookup paragraph); keep the lookup paragraph only where the tool schema is present (e.g. the has_tools schema description, react.py:1702). See also design note 8.

**2. Out-of-scope global embedding + 2-3x duplication** — src/xagent/core/file_ref.py:39
Line 39 embeds the block into `FILE_REF_OUTPUT_INSTRUCTIONS`, which line 51 embeds into `FILE_REF_MODEL_INSTRUCTIONS`, which rides `ExecutionContext._system_context` (execution.py:463-464) on every LLM call — every ReAct turn, Auto decision, and DAG step execution — plus the production uploaded-files prompts (websocket.py:5471, file_turn.py:365). Per final-answer call the block is duplicated: ReAct forced 3x (system context + react.py:828 + schema react.py:1702), Auto decision 3x (system context + auto.py:1313 + schema auto.py:1451), DAG assessment 2x. It also injects final-deliverable exclusivity ("Include only user-requested final deliverables. Do not include intermediate...") into intermediate step prompts whose contract is to report step results — a direct conflict. This exceeds the PR's own stated scope ("apply it to ReAct, DAG, and Auto final-answer prompts and tool schemas").
Suggestion: drop the f-string embedding at line 39 (restore the plain string ending at line 37, "unless that result exists."), keep the six explicit insertion sites, and remove the corresponding `assert instruction in FILE_REF_OUTPUT_INSTRUCTIONS` pin in tests/core/test_file_ref.py.

### Minor

**3. Lookup wording gap** — src/xagent/core/file_ref.py:19
The lookup result (workspace.py:1871-1892, workspace_file_tool.py:673-688) has no `markdown_link` field — only `file_id` (possibly None) plus filename/extension. The block remains satisfiable: the pre-existing rendering rule (file_ref.py:29 "Files: [filename](file:file_id)") produces exactly the string `build_file_ref` emits as `markdown_link` (file_ref.py:137); the never-construct rule (file_ref.py:35) targets untrusted provenance, and line 16 marks the returned `file_id` as trusted. Residual risk: "Copy the markdown_link verbatim" reads as if the lookup returned one, so a model may under-claim delivery despite a valid returned id.
Suggestion: add one clarifying sentence to the lookup paragraph, e.g. "render [filename](file:file_id) from the returned file_id — that rendered link is the exact markdown_link form."

**4. Conflict with computer-use screenshot guidance** — src/xagent/core/file_ref.py:10-11
The mandate to include the exact `markdown_link` conflicts with the computer-use screenshot instructions (computer.py:262-263, 471-476), which tell the model to include the returned `inline_markdown` verbatim in `final_answer` instead of `file_ref.markdown_link`; the screenshot result returns both. Both texts share the same model context on screenshot turns → duplicate references or dropped inline rendering (delivery integrity unaffected: same file id).
Suggestion: exempt tool-prescribed renderings, e.g. "unless the tool result prescribes a different rendering for that artifact (e.g. inline_markdown for screenshots)."

**5. Vocabulary drift** — src/xagent/core/file_ref.py:9, 13
"trusted public FileRef" and "trusted file UUID" match no system vocabulary: the FileRef distinction is `internal: bool` (file_ref.py:122, 220); "public" elsewhere means public context (artifacts.py:196); and the lookup returns `file_id`, an opaque string — workspace.py:288/616 show ids are not guaranteed UUIDs. Mismatched vocabulary weakens prompt compliance.
Suggestion: use "non-internal FileRef" and "file_id".

**6. Heading interpolated into plain descriptions + wording redundancy** — src/xagent/core/file_ref.py:8, 20
"## FINAL DELIVERABLE FILE REFERENCES" is baked into the shared constant, so it lands mid-sentence inside JSON-schema description fields (react.py:1701-1703, auto.py:1450-1452, dag.py:1560-1562) — an invalid ATX heading that reads as literal text — and at file_ref.py:39 it becomes a sibling `##` section of its container (also `##`). Separately, "Never invent, guess, shorten, or rewrite a file_id" (line 20) overlaps the pre-existing "Never invent, guess, or construct a file_id or `file:` link" (line 35) with complementary coverage. Cosmetic.
Suggestion: strip the heading from the shared constant and add it at the interpolation sites, or accept `##` as a pattern-level marker and drop the mid-sentence interpolation.

**7. Test weaknesses** — tests/core/agent/test_react.py:1354
(a) The new assertion checks the heading in `forced_prompt`, but the heading also arrives via the system context (`FILE_REF_MODEL_INSTRUCTIONS` through execution.py:464), so removing the react.py:828 insertion entirely would still leave the test green — it would not fail on the regression it claims to guard. (b) All pattern-level assertions check only the heading marker, not a distinctive body phrase. (c) No count==1 / absence assertions. (The DAG test, test_dag.py:394, and the Auto test, test_auto.py:1095 with messages[-1], ARE specific.)
Suggestion: assert a distinctive body phrase for the forced path, e.g. "call get_workspace_output_files once before finalizing"; optionally assert it appears exactly once.

### Design note

**8. Abstraction shape** — folded into finding 1's fix. The house pattern for final-answer rules is parameterized builders (`grounding_rule(can_call_tools=...)`, grounding.py:26-35; `final_answer_language_rule(subject=...)`); the new static constant embeds capability-dependent content (the lookup advice), which is the root cause of finding 1. Treat this as one fix: a builder function with a `can_lookup` parameter.

- **L16: delete:** single-use test constant `FINAL_DELIVERABLE_INSTRUCTION_MARKER` (tests/core/test_file_ref.py:16, used once at line 22; react/dag/auto tests inline the literal). Inline the literal string. → net: -3 lines possible.

Blocking: yes — two confirmed major findings (M1 unactionable recovery instruction, M3 out-of-scope global embedding).

- M1 — src/xagent/core/file_ref.py:14 — major — "call get_workspace_output_files once before finalizing" is unactionable at 3 of 6 insertion sites (ReAct force-final, Auto routing, DAG assessment) where the tool is absent from the schema [new]
- M3 — src/xagent/core/file_ref.py:39 — major — global embedding rides every LLM call via `FILE_REF_MODEL_INSTRUCTIONS`/`ExecutionContext._system_context` plus production prompts; 2-3x duplication per final answer and a conflict with intermediate-step prompts; exceeds declared scope [new]

Recommended event: REQUEST_CHANGES.

tests/core/agent/test_react.py
bsbds: We are not making a separate change for this comment because its regression outcome is implemented by the accepted global-context/duplication fix: the ReAct coverage now asserts one explicit final-deliverable block and absence from the normal work prompt, while the pattern schemas assert distinctive body text without an inline ATX heading ([global-context fix and regression coverage](https://github.com/xorbitsai/xagent/commit/b7822ca529ada214f594560b39a9390b28b9ef24); [schema-format coverage](https://github.com/xorbitsai/xagent/commit/f12d468869fd8a7524dc8f3c50a348e6c833064e)).

Treating this as another independent implementation outcome would duplicate the same accepted defect and tests ([related global-embedding review thread](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796533)).

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/harden-final-deliverable-links`
- **Head:** `5ae89bf512cb7e977cb9b4c142e9a524e193ddbb`
- **Destination:** `bsbds/xagent:fix/harden-final-deliverable-links`
- **Result:** success
- **PR description:** not requested

### Resolved

1. **Thread:** `PRRT_kwDORQ4Kr86ZRm-F` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796520
   **Local outcome:** resolved
   **Change:** `5711618af6bbec385e80277fbbabf0bf109eb37d`
   **GitHub:** resolved

2. **Thread:** `PRRT_kwDORQ4Kr86ZRm-O` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796533
   **Local outcome:** resolved
   **Change:** `b7822ca529ada214f594560b39a9390b28b9ef24`
   **GitHub:** resolved

3. **Thread:** `PRRT_kwDORQ4Kr86ZRm-V` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796540
   **Local outcome:** resolved
   **Change:** `e674a0fb3c5c02ce1ab390db46e30a04c6bc88c8`
   **GitHub:** resolved

4. **Thread:** `PRRT_kwDORQ4Kr86ZRm-a` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796549
   **Local outcome:** resolved
   **Change:** `3c63caefa1b72e2bc760e9dbc74283a0e004738f`
   **GitHub:** resolved

5. **Thread:** `PRRT_kwDORQ4Kr86ZRm-g` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796556
   **Local outcome:** resolved
   **Change:** `e7c5e7d502ae487f2bbee7701a1333a45fdd4079`
   **GitHub:** resolved

6. **Thread:** `PRRT_kwDORQ4Kr86ZRm-o` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796565
   **Local outcome:** resolved
   **Change:** `f12d468869fd8a7524dc8f3c50a348e6c833064e`
   **GitHub:** resolved

7. **Thread:** `PRRT_kwDORQ4Kr86ZRm-5` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796585
   **Local outcome:** resolved
   **Change:** `5ae89bf512cb7e977cb9b4c142e9a524e193ddbb`
   **GitHub:** resolved

### Rejected

1. **Thread:** `PRRT_kwDORQ4Kr86ZRm-w` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796575
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** The regression-test outcome is implemented by the accepted global-context/duplication resolution, so a separate implementation would duplicate the same fix and coverage.
   **Sources:** [global-context fix and regression coverage](https://github.com/xorbitsai/xagent/commit/b7822ca529ada214f594560b39a9390b28b9ef24), [schema-format coverage](https://github.com/xorbitsai/xagent/commit/f12d468869fd8a7524dc8f3c50a348e6c833064e), [related accepted review thread](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796533)
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3785191530
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 7
- **Rejected:** 1
- **Unresolved:** 0

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1697,6 +1699,7 @@ def _builtin_tool_schemas(self) -> list[dict[str, Any]]:
                                 "description": (
                                     "Complete user-facing answer. It must be "
                                     "non-empty and must match response_language. "
+                                    f"{final_deliverable_file_reference_instructions(can_lookup=True, include_heading=False)} "
```
rogercloud: **Major — forced-final-answer schema contradicts the forced prompt (the accepted M1 fix is partial).**

This shared `final_answer` schema description embeds `final_deliverable_file_reference_instructions(can_lookup=True, ...)`, which mandates "call get_workspace_output_files once before finalizing" (src/xagent/core/file_ref.py:9). But this same schema object is the only tool on forced-final turns (react.py:435-439; forced retries react.py:900-902), where the prompt says "Do not call any other tool" and deliberately passes `can_lookup=False` (react.py:822-829). So on forced turns the schema instructs a tool call that the active tool list does not expose and the prompt forbids — the exact unactionable-instruction defect that 5711618a was meant to fix, now leaking through the schema route.

Failure is real: `_response_requires_tool_protocol_retry` rejects any non-`final_answer` call on forced turns (react.py:1053-1054); a provider-level `unavailable_tool_call` violation (deepseek_tool_protocol.py:318-321) triggers the recovery path (react.py:518-561, 1104-1113), discarding the response and costing an extra LLM round trip while consuming a repair attempt — a second violation fails the run. Production-reachable via `single_call` (`ReActPattern(max_iterations=2, finalize_after_tool_result=True)`, execution_adapter.py:310-318; `force_final_answer_next` set after every successful tool result at react.py:2507).

The builder's own docstring states that omitting lookup advice "prevents the prompt from requiring a tool call that the active schema rejects" (file_ref.py:24-27) — this site violates exactly that design goal. Test gap: tests/core/agent/test_react.py:1354-1363 pins the forced prompt only; :3105-3107 pins this schema on the normal path; nothing pins the forced-turn schema content.

Suggestion: make the lookup sentence capability-conditional for the forced turn — e.g. a `can_lookup=False` schema variant for forced turns (the forced path already selects a dedicated schema list at react.py:435-439, so it can select a dedicated schema description too), or move the lookup sentence out of the shared `final_answer` schema into the `has_tools` path only. Add a test asserting the forced-turn schema does not contain "get_workspace_output_files".

rogercloud: **Minor — lookup advice is static while the tool is conditionally absent.**

The `can_lookup=True` lookup text is unconditional, but `get_workspace_output_files` can be absent from the active schema: `get_file_tools_enabled() == False` drops the workspace tools (src/xagent/core/tools/workspace_file_tool.py:284-285), and DAG steps restrict tools via `step.tool_names` (dag.py:1009) while builtin schemas are always appended (react.py:1842). On such turns this description contradicts the same-request rule "Never call a tool name that is not in this list" (react.py:867-868). Failure is absorbed by the recovery path, so this is minor; wiring the advice to actual tool availability (e.g. emitting the lookup sentence only when `get_workspace_output_files` is in the active tool list) would remove the contradiction.

src/xagent/core/file_ref.py
```diff
@@ -5,6 +5,45 @@
 from typing import Any
 from urllib.parse import quote, unquote, urlsplit
 
+_FINAL_DELIVERABLE_FILE_LOOKUP_INSTRUCTIONS = """If a generated final deliverable exists but no trusted file_id or markdown_link
+remains in the current context, call get_workspace_output_files once before finalizing.
+Match the exact deliverable path and filename, then use only a
+returned non-null file_id or markdown_link. From a returned file_id, render [filename](file:file_id)
+with the exact returned filename; this is the canonical markdown_link form.
```
rogercloud: **Minor — reconstruction advice vs the global "never construct" rule; duplicated claim rules (consistency nit, not a reversal).**

The new lookup sentence tells the model to "render [filename](file:file_id)" from a returned `file_id` and calls this "the canonical markdown_link form", but the pre-existing global rule in the same prompt context says "Never invent, guess, or construct a file_id or `file:` link... or the exact FileRef or markdown_link returned by a successful tool result" (file_ref.py:55). The "exact... returned" wording literally excludes a link the model reconstructs from a lookup `file_id` — which is the only path, since `get_workspace_output_files` returns no `markdown_link` (workspace.py:1881-1892). Separately, the new "Do not claim that a file was delivered unless its exact selected rendering appears in the final answer" (file_ref.py:39-40) duplicates the pre-existing "Do not call final_answer claiming that a file was created or delivered unless that result exists" (file_ref.py:57).

This derives from the accepted m3 fix (e674a0fb), so I'm not asking to reverse it — just noting the two sources of truth will drift. Consider marking the lookup-returned id as trusted provenance in the never-construct rule, and cross-referencing the two claim rules.

rogercloud: **Minor — reconstruction form vs the percent-encoded canonical form (theoretical).**

This sentence calls the literal `[filename](file:file_id)` reconstruction "the canonical markdown_link form", but the canonical encoder percent-encodes: `build_file_id_ref` applies `quote(normalized, safe='')` (file_ref.py:~121), pinned by `test_build_file_id_ref_uses_canonical_form` (tests/core/test_file_ref.py:87-91, "legacy id" → "file:legacy%20id"). In practice every minting site produces `str(uuid4())` or `internal-{uuid4()}` ids for which `quote()` is the identity, so divergence requires a non-UUID id (e.g. a custom client via websocket.py:1898-1902 or 4944-4946). Theoretical, hence minor — but the "canonical" claim is slightly overstated; e.g. "this matches the markdown_link form" would be accurate.

src/xagent/core/file_ref.py
```diff
@@ -5,6 +5,45 @@
...
+    rejects. Schema descriptions also omit the Markdown heading because they
+    embed these rules inline after existing prose.
+    """
+    lookup_instructions = (
+        _FINAL_DELIVERABLE_FILE_LOOKUP_INSTRUCTIONS if can_lookup else ""
+    )
+    heading = "## FINAL DELIVERABLE FILE REFERENCES\n" if include_heading else ""
+    return f"""{heading}If a successful tool result or trusted non-internal FileRef produced a file that
+satisfies the user's requested final deliverable, the final answer MUST include
+the exact markdown_link returned for that file, unless the successful tool result
```
rogercloud, line 43: **Minor — image deliverables: "exact markdown_link" mandate vs the global inline-image rule.**

This mandate ("the final answer MUST include the exact markdown_link returned for that file, unless the successful tool result prescribes a different exact user-facing rendering") coexists with the global rule "Images intended for inline display: filename" (file_ref.py:50), which rides `ExecutionContext._system_context` on every LLM call (execution.py:464) — while `build_file_ref`'s `markdown_link` has no `!` prefix (file_ref.py:155). Two divergent directives for the same image deliverables.

`generate_image`/`edit_image` are mitigated (the artifacts route returns `filename`, artifacts.py:120-121), but `browser_screenshot` (adapters/vibe/browser_use.py:717-721) and `download_web_asset` (download_web_asset.py:119-122) return only `markdown_link`/`file_ref`/`preview_url`/`download_url` — no artifacts and no prescribed rendering, so the model must choose between the two rules. The m4 exemption covered computer-use screenshots only. Consider extending the exemption clause to image-producing tools that carry no prescribed rendering, or have those tools return `inline_markdown`/artifacts like the image-generation path.

src/xagent/core/agent/pattern/auto/auto.py
```diff
@@ -1311,7 +1312,8 @@ def _decision_prompt(
             f"When writing that answer field: {grounding_rule(can_call_tools=False)} "
             "If the answer would need such unsupported specifics, set "
             "existing_context_sufficient=false and choose react so the agent can "
-            "verify them with tools. "
+            "verify them with tools.\n\n"
+            f"{final_deliverable_file_reference_instructions(can_lookup=False)}\n\n"
```
rogercloud, line 1316: **Minor — block interposed between "verify them with tools." and "You must also classify".**

The P1 placement fix (5ae89bf5) moved the deliverable block after the grounding/fallback pair, but it now sits between "verify them with tools." (line 1315) and "You must also classify whether..." (line 1317) — the base had those two sentences adjacent, and the "also" now refers across the interposed block. This is a side effect of the accepted fix rather than a new problem, and it's LLM-parseable (the ordering is pinned deliberately by tests/core/agent/test_auto.py:1096-1098), so I'm flagging it as a readability nit only: if the ordering test allows it, consider moving the block after "You must also classify" or joining it to the answer-field sentence above.

rogercloud (changes requested):
**Follow-up review — PR #1363: fix/harden-final-deliverable-links** (head `5ae89bf5`; prior review was against `4112557a`)

## Summary

Final answers across the ReAct, DAG, and Auto patterns can omit deliverable links, rewrite file IDs, claim undelivered files, or leak intermediate/internal files. This PR adds one shared prompt-level instruction block requiring the exact trusted `markdown_link` for user-requested deliverables, with a `get_workspace_output_files` lookup fallback that fails closed, applied to all three patterns' final-answer prompts and tool schemas via a parameterized builder. Runtime enforcement is explicitly out of scope (prompt-level mitigation).

## Update since prior review

Seven commits landed since the prior review (`4112557a`). `5711618a` scopes the deliverable-lookup advice to capable turns via a new `can_lookup` parameter on the builder `final_deliverable_file_reference_instructions` (six sites); `b7822ca5` removes the block from the global `FILE_REF_OUTPUT_INSTRUCTIONS` context; `e674a0fb` clarifies how a recovered deliverable link is rendered; `3c63caef` exempts tool-prescribed renderings such as `inline_markdown` for screenshots; `e7c5e7d5` canonicalizes the file-reference vocabulary; `f12d4688` makes the `##` heading optional so schema sites omit it; `5ae89bf5` keeps the Auto grounding guidance adjacent to its fallback sentence. The parameterization itself is correct at all six sites, but it left one route open: the forced-final-answer turn still receives the `can_lookup=True` lookup text through the shared `final_answer` schema (new major finding N1 below).

## Prior-findings checklist

- **P1 — auto block placement — FIXED.** `5ae89bf5` moves the deliverable block after the grounding/fallback pair; `grounding_rule` and the fallback sentence are adjacent (auto.py:1312-1315).
- **M1 — unactionable recovery instruction — PARTIAL (major).** All six sites are parameterized correctly (`can_lookup=False` at the forced/Auto/DAG sites, `can_lookup=True` only at react.py:1702), but the forced-final-answer turn still receives the `can_lookup=True` lookup text via the shared `final_answer` schema — the exact unactionable-instruction defect this fix targeted. See new finding 1.
- **M2 — global embedding — FIXED.** `b7822ca5` removes the embedding; `FILE_REF_OUTPUT_INSTRUCTIONS` is a plain string; exactly six explicit sites remain; tests flipped accordingly (tests/core/test_file_ref.py:52-53).
- **m3 — lookup wording — FIXED.** `e674a0fb` adds the render-from-`file_id` sentence (file_ref.py:11-12); fails-closed retained.
- **m4 — screenshot conflict — FIXED.** Exemption clause present (file_ref.py:35-36).
- **m5 — vocabulary — FIXED.** Canonical terms used repo-wide; old terms gone.
- **m6 — heading — FIXED.** `include_heading` parameter; schema sites pass `False`; `##` reachable only at prompt sites.
- **m7 — test cannot fail — FIXED.** The author's rejection rationale is verified: the `count == 1` assertion (tests/core/agent/test_react.py:1355) now independently guards the react.py:828 insertion (the global-context leak was removed), and schema-site tests pin distinctive body text.
- **D8 — builder shape — FIXED.** Keyword-only builder matching the `grounding_rule` house pattern.
- **S1 — single-use test constant — FIXED/moot.** The constant is now used 4×; inlining would regress.

## New findings

### Major

**1. Forced-final-answer schema contradicts the forced prompt — the accepted M1 fix is partial** — src/xagent/core/agent/pattern/react/react.py:1702

The shared `final_answer` tool schema embeds `final_deliverable_file_reference_instructions(can_lookup=True, include_heading=False)` in its `answer` description, which mandates "call get_workspace_output_files once before finalizing" (file_ref.py:9). But this same schema object is the only tool on forced-final turns (react.py:435-439; forced retries react.py:900-902), where the prompt says "Do not call any other tool" and deliberately passes `can_lookup=False` (react.py:822-829). The result is a per-request contradiction: the schema instructs a tool call that the active tool list does not expose and the prompt forbids.

The failure is real: `_response_requires_tool_protocol_retry` rejects any non-`final_answer` call on forced turns (react.py:1053-1054); a provider-level `unavailable_tool_call` violation (deepseek_tool_protocol.py:318-321) triggers the recovery path (react.py:518-561, 1104-1113), discarding the response and costing an extra LLM round trip while consuming a repair attempt — a second violation fails the run. Production-reachable: the `single_call` pattern runs `ReActPattern(max_iterations=2, finalize_after_tool_result=True)` (execution_adapter.py:310-318), which sets `force_final_answer_next` after every successful tool result (react.py:2507) with `tool_choice='required'`.

The builder's own docstring states that omitting lookup advice "prevents the prompt from requiring a tool call that the active schema rejects" (file_ref.py:24-27) — the schema site violates exactly that design goal. Tests pin the forced prompt (tests/core/agent/test_react.py:1354-1363) and the schema on the normal path (tests/core/agent/test_react.py:3105-3107), but nothing pins the forced-turn schema content, which is why this route survived the M1 fix.

Suggestion: make the lookup sentence capability-conditional for the forced turn — e.g. a separate `can_lookup=False` schema variant for forced turns (the forced path already selects a dedicated schema list at react.py:435-439, so it can select a dedicated schema description too), or move the lookup sentence out of the shared `final_answer` schema into the `has_tools` path only. Add a test asserting the forced-turn schema does not contain "get_workspace_output_files".

### Minor

**2. Lookup advice is static while the tool is conditionally absent** — react.py:1702
The `can_lookup=True` text is unconditional, but `get_workspace_output_files` can be absent from the active schema: `get_file_tools_enabled() == False` drops the workspace tools (workspace_file_tool.py:284-285), and DAG steps restrict tools via `step.tool_names` (dag.py:1009) while builtin schemas are always appended (react.py:1842). On such turns the schema contradicts the same-request rule "Never call a tool name that is not in this list" (react.py:867-868). Failure is absorbed by the recovery path, so minor; wiring the advice to actual tool availability would remove the contradiction.

**3. Image deliverables: "exact markdown_link" mandate vs the global inline-image rule** — file_ref.py:35
The new mandate "the final answer MUST include the exact markdown_link returned for that file, unless the successful tool result prescribes a different exact user-facing rendering" coexists with the global rule "Images intended for inline display: filename" (file_ref.py:50), injected into every request via `ExecutionContext._system_context` (execution.py:464), while `build_file_ref`'s `markdown_link` has no `!` prefix (file_ref.py:155). `generate_image`/`edit_image` are mitigated (the artifacts route returns `filename`, artifacts.py:120-121), but `browser_screenshot` (adapters/vibe/browser_use.py:717-721) and `download_web_asset` (download_web_asset.py:119-122) return only `markdown_link`/`file_ref`/`preview_url`/`download_url` — no artifacts, no prescribed rendering. Minor; consider extending the exemption to image-producing tools without a prescribed rendering.

**4. Render advice vs the global "never construct" rule; duplicated claim rules** — file_ref.py:12
The new sentence "From a returned file_id, render [filename](file:file_id)... this is the canonical markdown_link form" (file_ref.py:11-12) sits beside the pre-existing "Never invent, guess, or construct a file_id or `file:` link... or the exact FileRef or markdown_link returned by a successful tool result" (file_ref.py:55) — the "exact... returned" wording literally excludes a link the model reconstructs from a lookup `file_id`, which is the only path since lookups return no `markdown_link` (workspace.py:1881-1892). Separately, "Do not claim that a file was delivered unless its exact selected rendering appears in the final answer" (file_ref.py:39-40) duplicates "Do not call final_answer claiming that a file was created or delivered unless that result exists" (file_ref.py:57). Consistency nit deriving from the accepted m3 fix — not a reversal; consider cross-referencing so the two sources cannot drift.

**5. Reconstruction form vs the percent-encoded canonical form (theoretical)** — file_ref.py:12
The new text calls the literal `[filename](file:file_id)` reconstruction "the canonical markdown_link form", but the canonical encoder percent-encodes (`build_file_id_ref` → `quote(normalized, safe='')`, file_ref.py:~121), pinned by `test_build_file_id_ref_uses_canonical_form` (tests/core/test_file_ref.py:87-91, "legacy id" → "file:legacy%20id"). In practice every minting site produces `str(uuid4())` or `internal-{uuid4()}` ids for which `quote()` is the identity, so divergence requires a non-UUID id (e.g. a custom client via websocket.py:1898-1902). Theoretical; the "canonical" claim is slightly overstated.

**6. Auto: block interposed between "verify them with tools." and "You must also classify"** — auto.py:1316
The P1 placement fix (5ae89bf5) moved the deliverable block after the grounding/fallback pair, but it now sits between "verify them with tools." (auto.py:1315) and "You must also classify whether..." (auto.py:1317) — a side effect of the accepted fix that breaks the adjacency the base had at that junction, so the "also" now refers across the interposed block. LLM-parseable and deliberate (tests/core/agent/test_auto.py:1096-1098 pins the ordering); readability nit.

None — both simplification lenses returned "Lean already", and the prior review's single simplification item (S1, inline the single-use test constant) is moot because the constant is now used 4×.

Blocking: yes — M1 PARTIAL [prior] + N1 [new].

- src/xagent/core/agent/pattern/react/react.py:1702 — major — forced-final-answer turn receives `can_lookup=True` lookup advice via the shared `final_answer` schema while the prompt forbids other tools and the schema exposes only `final_answer`; the accepted M1 fix (5711618a) left this route open [new]
- src/xagent/core/file_ref.py:13-14 — major — unactionable recovery instruction persists on the forced-final path [prior, PARTIAL]

Recommended event: REQUEST_CHANGES (author bsbds — not qinxuye).

src/xagent/core/agent/pattern/react/react.py
bsbds: We are not making a separate change for this comment because it requests the same capability invariant as the accepted forced-schema finding: lookup advice must appear only when `get_workspace_output_files` is actually present. The shared fix now derives `can_lookup_output_files` from the active external tool schemas, so normal turns without that tool and forced turns both omit the lookup action, while turns that actually expose the tool retain it ([active-tool capability fix and tests](https://github.com/xorbitsai/xagent/commit/78375eb146ffe17df558bb1a4c295fa9931b8064)).

A second implementation would duplicate the same resolved outcome ([related accepted review thread](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564654)).

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/harden-final-deliverable-links`
- **Head:** `d9e140485879c8a58575a22b817ae6aef93bdc79`
- **Destination:** `bsbds/xagent:fix/harden-final-deliverable-links`
- **Result:** success
- **PR description:** not requested

### Resolved

1. **Thread:** `PRRT_kwDORQ4Kr86ZeRTw` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564654
   **Local outcome:** resolved
   **Change:** `78375eb146ffe17df558bb1a4c295fa9931b8064` with follow-up test clarification `d9e140485879c8a58575a22b817ae6aef93bdc79`
   **GitHub:** resolved

2. **Thread:** `PRRT_kwDORQ4Kr86ZeRUE` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564685
   **Local outcome:** resolved
   **Change:** `bbae957621cda31d747f9ecc81a0ac552786fb6a`
   **GitHub:** resolved

3. **Thread:** `PRRT_kwDORQ4Kr86ZeRUQ` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564700
   **Local outcome:** resolved
   **Change:** `c04af1fcf69a33b194076b323a9ab62c32ff7a4d`
   **GitHub:** resolved

4. **Thread:** `PRRT_kwDORQ4Kr86ZeRUU` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564709
   **Local outcome:** resolved
   **Change:** `0f9164b6b69e5db1a901ccafb8f57cf8f089495c`
   **GitHub:** resolved

5. **Thread:** `PRRT_kwDORQ4Kr86ZeRUY` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564718
   **Local outcome:** resolved
   **Change:** `aa4c9e675b38aae1faa62055816a6318cc9e6a4f`
   **GitHub:** resolved

### Rejected

1. **Thread:** `PRRT_kwDORQ4Kr86ZeRT-` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564674
   **Local outcome:** rejected
   **Classification:** duplicate
   **Reason:** The active-tool capability derivation for the accepted forced-schema finding is the same required outcome and covers normal turns where workspace lookup is absent.
   **Sources:** [active-tool capability fix and tests](https://github.com/xorbitsai/xagent/commit/78375eb146ffe17df558bb1a4c295fa9931b8064), [related accepted review thread](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564654)
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788595430
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 5
- **Rejected:** 1
- **Unresolved:** 0

src/xagent/core/file_ref.py
```diff
@@ -5,6 +5,45 @@
...
+    Final-only protocol turns cannot call workspace tools. Omitting lookup advice
+    there prevents the prompt from requiring a tool call that the active schema
+    rejects. Schema descriptions also omit the Markdown heading because they
+    embed these rules inline after existing prose.
+    """
+    lookup_instructions = (
+        _FINAL_DELIVERABLE_FILE_LOOKUP_INSTRUCTIONS if can_lookup else ""
+    )
+    heading = "## FINAL DELIVERABLE FILE REFERENCES\n" if include_heading else ""
+    return f"""{heading}If a successful tool result or trusted non-internal FileRef produced a file that
```
rogercloud: **Minor (C1)** — In the `can_lookup=False` variant (forced/assessment/routing turns) this states "the final answer MUST include the exact markdown_link returned for that file" with no fallback clause for the state "deliverable exists but its link is no longer in context and no lookup tool is available". The anti-fabrication half is covered by the global never-construct rule plus `grounding_rule(can_call_tools=False)` at those sites, but the model must infer the omit-and-say-so resolution instead of being told it — the `can_lookup=True` variant explicitly instructs this scenario, while this variant silently drops it. Suggestion: add a short clause, e.g. "If the link is not available in context, omit the link and say so rather than reconstructing it."

src/xagent/core/file_ref.py
```diff
@@ -5,6 +5,45 @@
 from typing import Any
 from urllib.parse import quote, unquote, urlsplit
 
+_FINAL_DELIVERABLE_FILE_LOOKUP_INSTRUCTIONS = """If a generated final deliverable exists but no trusted file_id or markdown_link
+remains in the current context, call get_workspace_output_files once before finalizing.
+Match the exact deliverable path and filename, then use only a
```
rogercloud: **Minor (C2)** — `get_workspace_output_files` never returns `markdown_link`: verified the full chain (workspace_file_tool.py:673-686 → workspace.py:1811-1831 → `_get_file_info` 1875-1892; keys are file_id/file_path/relative_path/location/size/modified_time/filename/extension/is_readable/is_writable; no markdown_link anywhere in workspace.py or workspace_file_tool.py). The dead alternative "or markdown_link" invites the model to scan for a field that cannot exist. Suggest dropping "or markdown_link" or adding a parenthetical like "(lookup results carry only file_id)".

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1828,7 +1837,16 @@ def _tool_schemas_with_builtin_controls(
             for tool in tools
             if self._tool_name(tool) not in control_tool_names
         ]
-        return [*external_tools, *self._builtin_tool_schemas()]
+        can_lookup_output_files = any(
+            schema.get("function", {}).get("name") == "get_workspace_output_files"
```
rogercloud: **Minor (C3)** — The tool name `get_workspace_output_files` is hardcoded at three uncoupled sites: this detection, the prompt text in file_ref.py:9, and the registration in adapters/vibe/workspace_file_tool.py:237-238 — with no shared constant. A registration-only rename would silently disable the lookup instruction with no error, and the positive test (independent FakeWorkspaceOutputTool) would stay green. Suggest a shared constant for the tool name referenced by detection/prompt/registration.

src/xagent/core/file_ref.py
```diff
@@ -14,6 +53,7 @@
 File delivery integrity:
 - Internal FileRefs (`internal: true`) are runtime context only. Never render or expose a `file:` link for them.
 - Never invent, guess, or construct a file_id or `file:` link. Use only a trusted FileRef supplied for an existing input/attachment, or the exact FileRef or markdown_link returned by a successful tool result. A link mentioned only in prior assistant prose is not provenance.
+- An exact non-null file_id and filename returned together by get_workspace_output_files are trusted provenance for rendering a file link under the final-deliverable rules.
```
rogercloud: **Minor (C4)** — This provenance bullet rides every LLM call via `FILE_REF_MODEL_INSTRUCTIONS` (execution.py:464, context_builder.py:134/620, websocket.py:5471, file_turn.py:365), while the "## FINAL DELIVERABLE FILE REFERENCES" block is deliberately absent from all those contexts — a mild regression against the PR's own design rule that the global constant must not carry final-deliverable content. The sentence is self-contained enough to be harmless; suggest dropping the qualifier: "…are trusted provenance for rendering a file link."

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -823,7 +824,8 @@ def _messages_for_llm(
                 "tool-call markup as plain text. Set outcome=completed only when "
                 "every requested action or verification succeeded; otherwise set "
                 "outcome=partial or outcome=blocked and say what remains. "
-                f"{grounding_rule(can_call_tools=False)} "
+                f"{grounding_rule(can_call_tools=False)}\n\n"
+                f"{final_deliverable_file_reference_instructions(can_lookup=False)}\n\n"
```
rogercloud, line 831: **Minor (C5)** — This deliverable instruction is only in the forced-final-answer prompt and the `final_answer` schema description (react.py:1708), NOT in the `has_tools` system prompt (react.py:831-869, unchanged from base — note `grounding_rule` at ~855 IS prompt-level there). `has_tools` turns accept plain-text zero-tool-call responses as final answers (react.py:715, 1478-1495), so a plain-text final answer on the main production path never sees the MUST rule. Production-reachable where providers relax `tool_choice=required` to `auto` (zhipu.py:197-198, router.py:155-163, dashscope.py:152-160). Suggest adding a short version of the instruction to the `has_tools` system prompt.

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1697,6 +1705,7 @@ def _builtin_tool_schemas(self) -> list[dict[str, Any]]:
                                 "description": (
                                     "Complete user-facing answer. It must be "
                                     "non-empty and must match response_language. "
+                                    f"{final_deliverable_file_reference_instructions(can_lookup=can_lookup_output_files, include_heading=False)} "
```
rogercloud, line 1714: **Nit (C6)** — Forced/assessment/routing turns present the deliverable rules twice: the full heading version in the system instruction (react.py:828, auto.py:1316, dag.py:1528) plus this inline version in the answer-field description (~430 chars read twice per turn). Informational only; the schema copy is required for normal turns, so removing it would need a second schema variant.

src/xagent/core/file_ref.py
rogercloud: **Minor (C7, prose)** — "If a successful tool result or trusted non-internal FileRef produced a file" attributes production to the FileRef; per the module's own definitions, tool results produce files and return FileRefs — the FileRef is a reference, not a producer. Suggest: "If a successful tool result produced a file (or a trusted non-internal FileRef references one) …" (Second sub-point "Preserve every returned file_id exactly" was checked and is consistent with adjacent wording.)

src/xagent/core/agent/pattern/react/react.py
```diff
@@ -1110,8 +1112,12 @@ def _requires_full_tool_set_recovery(
             for tool_call in normalized.get("tool_calls") or []
         )
 
-    def _final_answer_tool_schema(self) -> dict[str, Any]:
-        for schema in self._builtin_tool_schemas():
+    def _final_answer_tool_schema(
```
rogercloud: **Simplify (S1, delete)** — The `can_lookup_output_files: bool = False` parameter is never passed `True` at either call site (react.py:436, 901 — both forced paths where `False` is the only correct value). Drop the parameter and call `_builtin_tool_schemas()` bare; keep the parameter on `_builtin_tool_schemas` itself (the real capability path passes `True` through it). net: -1 parameter.

tests/core/test_file_ref.py
```diff
@@ -11,6 +13,55 @@
...
+    )
+    inline_instruction = file_ref.final_deliverable_file_reference_instructions(
+        can_lookup=False,
+        include_heading=False,
+    )
+
+    for instruction in (lookup_instruction, forced_instruction, inline_instruction):
+        assert "exact markdown_link" in instruction
+        assert "trusted non-internal FileRef" in instruction
+        assert "trusted public FileRef" not in instruction
```
rogercloud, line 38: **Simplify (S2, shrink)** — The five negative assertions at lines 34-35, 43-44, 56 pin historical phrasings absent from the implementation ("trusted public FileRef", "file UUID", "Never invent, guess, shorten", "canonical markdown_link form", "Do not claim that a file was delivered"); 3 of 5 are strictly redundant with positive assertions and none fires on a behavioral break. Drop the five negatives; keep the positives. net: -5 assertions.

rogercloud (approved):
## Summary

This PR makes final-deliverable file links mandatory and trustworthy at the moment the answer is written. Agent runs producing file-based deliverables previously could omit the link from the final answer or emit reconstructed/hallucinated, intermediate, or internal runtime-only file references, because the pre-existing global `FILE_REF_OUTPUT_INSTRUCTIONS` made nothing mandatory at answer-writing time and offered no recovery when compaction or context loss dropped the link. The fix introduces one shared pure string-builder, `final_deliverable_file_reference_instructions(*, can_lookup, include_heading)` in `src/xagent/core/file_ref.py`, and applies it to all six final-answer sites (ReAct forced prompt + final_answer schema, Auto decision prompt + schema, DAG assessment prompt + schema), with workspace-lookup recovery scoped to turns whose active tool schema actually exposes `get_workspace_output_files`.

## Update since prior review (5ae89bf5 → d9e14048)

Six commits (+72/−13 across 6 files) since the last review, addressing the remaining findings:

- `78375eb1` — scope deliverable lookup to active tools via the `can_lookup_output_files` parameter (fixes N1, N2); adds a new positive test and a forced-schema negative assertion
- `bbae9576` — align recovered-link provenance rules: provenance bullet added to the global block; duplicated "Do not claim that a file was delivered" sentence removed (N4)
- `c04af1fc` — qualify recovered-link wording: "canonical markdown_link form" → "matches the markdown_link form" (N5)
- `0f9164b6` — preserve inline image deliverable rendering via an extended exemption clause (N3)
- `aa4c9e67` — remove auto-routing anaphora: "You must also classify" → "You must classify" (N6)
- `d9e14048` — weaken the forced-prompt test assertion that could not pass against the current code (prior V8; the weakening is necessary and harmless — the bare tool name legitimately enters the forced prompt via the provenance bullet riding `ExecutionContext._system_context`)

## Prior findings — all FIXED

- **P1** (auto block placement) — FIXED
- **M1** (unactionable recovery instruction) — FIXED via 78375eb1: both `_final_answer_tool_schema` call sites are forced paths with default `False`; the normal path derives `can_lookup` from `external_tools`; tests pin both directions and fail on the old behavior
- **M2** (global embedding) — FIXED with no regression: only a one-line provenance bullet was added to the global constant; no heading/MUST/exclusivity
- **m3** (lookup wording) — FIXED
- **m4** (screenshot conflict) — FIXED
- **m5** (vocabulary) — FIXED
- **m6** (heading) — FIXED
- **m7** (test cannot fail) — FIXED (d9e14048 weakening verified necessary; `count == 1` + streams-schema assertions independently guard)
- **D8** (builder shape) — FIXED
- **N1** (forced-turn schema contradiction) — FIXED: default `False` at react.py:1115-1116; forced call sites bare at 436/901; schema conditional at 1708; tests pin; production path in execution_adapter.py:310-318 safe
- **N2** (static lookup advice) — FIXED: `external_tools`-derived; DAG/auto/forced turns never use builtin `can_lookup=True`
- **N3** (image deliverables vs inline-image rule) — FIXED via the inline-image exemption clause (covers browser_screenshot/download_web_asset, which lack `inline_markdown`)
- **N4** (render advice vs never-construct + duplicated claim rules) — FIXED via the provenance bullet; duplicated sentence removed and pinned by test
- **N5** (canonical-form overclaim) — FIXED via c04af1fc wording
- **N6** (auto anaphora) — FIXED via aa4c9e67, pinned both directions
- **S1** (single-use test constant) — moot: now used 4×

## Round 0 approach verdict

Sound. A single shared pure string-builder applied at all six final-answer sites, with lookup recovery gated on the active tool schema, is the right shape: it removes the duplication that let the sites drift apart and guarantees the MUST rule reaches every answer-writing location, without forcing tool calls that the active schema would reject.

## New findings (current head)

### Minor

- **C1 — no-fallback bind in the `can_lookup=False` variant** (file_ref.py:33-37): forced/assessment/routing turns state "the final answer MUST include the exact markdown_link returned for that file" but give no clause for "deliverable exists but its link is no longer in context and no lookup tool is available". The anti-fabrication half is covered by the global never-construct rule plus `grounding_rule(can_call_tools=False)` at those sites, so the model must infer the omit-and-say-so resolution instead of being told it — the `can_lookup=True` variant spells this out explicitly, this variant silently drops it.
- **C2 — dead 'or markdown_link'** (file_ref.py:10-11): `get_workspace_output_files` never returns `markdown_link` (verified chain: workspace_file_tool.py:673-686 → workspace.py:1811-1831 → `_get_file_info` 1875-1892; keys are file_id/file_path/relative_path/location/size/modified_time/filename/extension/is_readable/is_writable). The dead alternative misleads the model into scanning for a field that cannot exist.
- **C3 — hardcoded tool-name literal at three uncoupled sites** (react.py:1841, file_ref.py:9, adapters/vibe/workspace_file_tool.py:237-238): no shared constant. A registration-only rename of `get_workspace_output_files` would silently disable the lookup instruction with no error, and the positive test (independent fake tool) would stay green.
- **C4 — dangling cross-reference** (file_ref.py:56): the provenance bullet "…trusted provenance for rendering a file link under the final-deliverable rules" rides every LLM call via `FILE_REF_MODEL_INSTRUCTIONS` (execution.py:464, context_builder.py:134/620, websocket.py:5471, file_turn.py:365), while the "## FINAL DELIVERABLE FILE REFERENCES" block is deliberately absent from all those contexts — a mild regression against the PR's own design rule that the global constant must not carry final-deliverable content. The sentence is self-contained enough to be minor; dropping the qualifier would restore the boundary.
- **C5 — plain-text final answer bypasses the MUST rule** (react.py:827-828 vs 831-869): `has_tools` turns accept plain-text zero-tool-call responses as final answers (react.py:715, 1478-1495), and the new deliverable instruction exists only in the `final_answer` schema description (react.py:1708), not in the `has_tools` system prompt (unchanged from base, while `grounding_rule` there IS prompt-level). A plain-text final answer on the main production path never sees the MUST rule. Production-reachable where providers relax `tool_choice=required` to `auto` (zhipu.py:197-198, router.py:155-163, dashscope.py:152-160).
- **C6 (nit) — per-turn duplication** (react.py:1708, auto.py:1451, dag.py:1561): forced/assessment/routing turns present the rules twice — the full heading version in the system instruction plus this inline version in the answer-field description (~430 chars read twice per turn). Informational; the schema copy is required for normal turns, so removing it would need a second schema variant.
- **C7 — prose: FileRef as producer** (file_ref.py:33): "If a successful tool result or trusted non-internal FileRef produced a file" attributes production to the FileRef; per the module's own definitions, tool results produce files and return FileRefs — the FileRef is a reference, not a producer.

### Design note (body only)

- **C8 — unconventional home for a shared prompt fragment**: `file_ref.py` is not the established home for shared final-answer prompt fragments (`grounding.py`, `language.py` are), but this is defensible — the module already hosts `FILE_REF_OUTPUT_INSTRUCTIONS` prompt fragments, there is no import cycle, and all patterns already import from it. Awareness only; no change requested.

- `L1115: delete` — `_final_answer_tool_schema`'s `can_lookup_output_files: bool = False` parameter is never passed `True` at either call site (react.py:436, 901 — both forced paths where `False` is the only correct value). Drop the parameter and call `_builtin_tool_schemas()` bare; keep the parameter on `_builtin_tool_schemas` itself (the real capability path passes `True` through it).
- `L34: shrink` — tests/core/test_file_ref.py:34-35, 43-44, 56: five negative assertions pin historical phrasings absent from the implementation ("trusted public FileRef", "file UUID", "Never invent, guess, shorten", "canonical markdown_link form", "Do not claim that a file was delivered"); 3 of 5 are strictly redundant with positive assertions and none fires on a behavioral break. Keep the positives.
- `net:` -1 dead parameter; -5 assertions

Blocking: no — no blocking issues. All prior findings are fixed (P1, M1, M2, m3-m7, D8, N1-N6 FIXED; prior S1 moot), CI is green, and the 287 tests in the 4 changed test files pass. The remaining items are 7 minor/nit findings, 1 design-level awareness note, and 2 optional simplifications, none of which block merge. Recommended event: **APPROVE**.

src/xagent/core/agent/pattern/react/react.py
bsbds: We are leaving both copies in place. The answer-schema copy is still required for normal structured `final_answer` responses, while the normal system-prompt copy closes the reviewed plain-text final-response bypass ([plain-text coverage change and tests](https://github.com/xorbitsai/xagent/commit/e3524b112a32b1b863e363c529c46ffd5f7b543b); [related accepted C5 thread](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741721)).

Removing either copy would reopen one of those answer paths, and this comment explicitly marked the duplication informational only, so there is no separate safe change to apply.

tests/core/test_file_ref.py
bsbds: We are retaining these negative assertions because they guard the absence of wording that earlier review rounds found contradictory or drifted: the old FileRef/UUID vocabulary ([vocabulary finding](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796556)), the duplicated delivery-claim rule ([consistency finding](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564685)), and the overstated canonical-form claim ([encoding finding](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564700)). The current focused test keeps those absence contracts together with their positive replacements ([current test source](https://github.com/xorbitsai/xagent/blob/a529259763d331b65b447966dd2a801de4d1c853/tests/core/test_file_ref.py#L34-L63)).

Deleting all five would reduce regression coverage without changing runtime behavior, so we are not applying the requested wholesale removal.

bsbds:
## Finalization Summary

### Push

- **Branch:** `fix/harden-final-deliverable-links`
- **Head:** `a529259763d331b65b447966dd2a801de4d1c853`
- **Destination:** `bsbds/xagent:fix/harden-final-deliverable-links`
- **Result:** success
- **PR description:** not requested

### Resolved

1. **Thread:** `PRRT_kwDORQ4Kr86ZeuAk` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741715
   **Local outcome:** resolved
   **Change:** `0e327c89ae898f687b3bb3bb71919c4109eddd96`
   **GitHub:** resolved

2. **Thread:** `PRRT_kwDORQ4Kr86ZeuAl` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741716
   **Local outcome:** resolved
   **Change:** `319de4bef922c2ff3dc6da5e15ed76df77c1e044`
   **GitHub:** resolved

3. **Thread:** `PRRT_kwDORQ4Kr86ZeuAm` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741717
   **Local outcome:** resolved
   **Change:** `99b1c9a80df244c9bb9332daecf6906e65e7122a`
   **GitHub:** resolved

4. **Thread:** `PRRT_kwDORQ4Kr86ZeuAn` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741720
   **Local outcome:** resolved
   **Change:** `9eb9256943a63764a9af759ffb303f6a614e5084`
   **GitHub:** resolved

5. **Thread:** `PRRT_kwDORQ4Kr86ZeuAo` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741721
   **Local outcome:** resolved
   **Change:** `e3524b112a32b1b863e363c529c46ffd5f7b543b`
   **GitHub:** resolved

6. **Thread:** `PRRT_kwDORQ4Kr86ZeuAq` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741723
   **Local outcome:** resolved
   **Change:** `be61a1bbb93e503e2afb7d66e9d63924dc2b0159`
   **GitHub:** resolved

7. **Thread:** `PRRT_kwDORQ4Kr86ZeuAr` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741725
   **Local outcome:** resolved
   **Change:** `a529259763d331b65b447966dd2a801de4d1c853`
   **GitHub:** resolved

### Rejected

1. **Thread:** `PRRT_kwDORQ4Kr86ZeuAp` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741722
   **Local outcome:** rejected
   **Reason:** The schema copy covers structured final answers and the system copy covers forced/plain-text paths; removing either reopens a verified answer-path gap. The reviewer marked this informational only.
   **Sources:** [plain-text coverage change and tests](https://github.com/xorbitsai/xagent/commit/e3524b112a32b1b863e363c529c46ffd5f7b543b), [related accepted C5 thread](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741721)
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788765628
   **GitHub:** left unresolved

2. **Thread:** `PRRT_kwDORQ4Kr86ZeuAt` — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788741727
   **Local outcome:** rejected
   **Reason:** The negative assertions guard absence of previously reviewed contradictory, drifted, duplicated, and overstated wording and have no runtime cost.
   **Sources:** [vocabulary finding](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3783796556), [consistency finding](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564685), [encoding finding](https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788564700), [current test source](https://github.com/xorbitsai/xagent/blob/a529259763d331b65b447966dd2a801de4d1c853/tests/core/test_file_ref.py#L34-L63)
   **Reply:** posted — https://github.com/xorbitsai/xagent/pull/1363#discussion_r3788765637
   **GitHub:** left unresolved

### Unresolved

None.

### Totals

- **Resolved:** 7
- **Rejected:** 2
- **Unresolved:** 0

PR #1371, opened 2026-08-14
Resolve an agent's declared knowledge bases against its own team

AlexLiu190625:
**Builds on #1349.** That PR resolves the knowledge-base team layer on the
governing agent's team, adds the typed scope error, and carries the governing
agent's creator and stored declaration to the search entry point without reading
them. This PR is the rule that reads them. With #1349 merged, this diff is the rule
plus three small pieces that only make sense once the rule exists: the typed
scope error's HTTP exit in the task-creation endpoint (defence in depth, mirroring
the connector error's exit beside it — no request path resolves the team layer
today), a behavioural pin on the live-ORM derivation branch, and docstring wording
aligned with what the production call sites actually bind.

### What this changes

With the seam in place, every value the rule needs is already at
`_search_knowledge_base_impl` and unused. This PR is the rule.

A team-governed run now gives **every name in the agent's stored `knowledge_bases`
declaration its own verdict**:

| Situation | Verdict |
|---|---|
| The governing team owns the name | The team's copy, resolved on the team's storage tenant |
| The team does not own it, and the runner **is** the agent's creator | The creator's own collection, exactly as with no governing team |
| The team does not own it, the runner is not the creator, and the creator does hold it | Nothing searchable, plus a sentence saying the knowledge base was never shared with the team |
| The team does not own it and nobody holds it | Nothing searchable, plus the existing "does not exist" wording |
| The creator lookup raised | Nothing searchable, plus wording that asserts no absence, because nothing established one |

Without this, a name the team never received quietly resolved onto whatever
same-named personal collection the runner happened to own. The visible union seeds
the runner's own collections, so a rule keyed on that union would never see the very
name it exists to reject — which is why the search set and both report sets are all
built from the verdicts, never from an intersection with what is visible.

### The gate

The rule applies only when three things hold together: a governing team, an
installed team-keyed hook, and a non-empty declaration.

- **No installed hook** means the deployment's team layer still resolves the old,
  runner-keyed way. Turning the rule on there would produce a half-migrated state
  nothing asked for, so the rule stays off and behaviour is unchanged byte for byte.
- **An empty declaration** has nothing to authorise against, so the run falls
  through to the existing "search everything visible" branch. The governing team's
  knowledge bases are still in that visible set — the team layer resolves
  independently of what was declared — so such a run searches the runner's own
  material *and* the governing team's, as it always has.

This gate is intentionally not the same predicate as the team layer's own gate: the
team layer's job is "resolve onto the governing team", which an empty declaration
does not change.

### Reporting

The terminal summary is derived from the verdicts, one report set per verdict
class, so no name is described twice and none is described as both absent and
held. When nothing in scope is the creator's own, both inherited terminal
sentences render byte for byte on their own call style; when at least one
failed name is the creator's unshared collection, the summary lists only the
names actually classified missing, then the sharing-gap sentences, then the
admitted set. The creator-existence probe excludes names on the creator's
tenant that a team the creator belongs to owns, so another team's collection
hosted there is not reported as the creator's.

The "Available"/"Allowed" clauses are derived from the names the verdict loop
admitted, captured **before** the model's explicit request narrows the search. Two
things follow:

- A fully narrowed-away request still tells the model what it could have asked for.
- The clause can never name the governing team's collections the agent did not
  declare, nor the runner's own same-named personal copies — precisely the
  collections the rule refuses to search.

The re-based "not allowed" gate takes the agent's stored declaration as its
authority rather than `tool_args.allowed_collections`, since that field is
model-authored and a model-supplied value can win over the agent's configured one.

### The creator probe

Classifying a name as the creator's own unshared collection needs one lookup of the
creator's collection names, memoised at most once per search call. A lookup that
fails degrades to "not held" for the search decision and switches the report to
wording that asserts no absence.

The probe is bound to the **stored declaration**, not to anything on `tool_args`, so
a model-supplied name can never reach it and turn its two distinguishable outcomes
into an enumeration oracle over the creator's private collection names. A remaining,
costlier existence-only side channel is documented in the probe's docstring: the
stored declaration is editable by any same-team member, so a member can add a name
and read the outcome off the summary, at the cost of an agent edit per name and
revealing only existence, never contents.

### DIFF
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -250,15 +250,21 @@ async def _list_knowledge_bases_impl(
             ``knowledge_tools.py`` builds the list tool only when the
             agent's allowed collections are ``None`` -- a non-empty list
             builds the search tool alone, and an empty list builds no
-            knowledge tool at all -- and both chat call sites derive that
-            value and the declaration from the same stored field, so on this
-            path there is no declared name to classify as "unshared" versus
-            "missing". See ``declared_knowledge_bases`` below.
+            knowledge tool at all -- and both chat call sites derive the
+            allowed collections and the declaration from the same stored
+            field, so in production a list tool is built with no declared
+            name to classify as "unshared" versus "missing". A test may
+            still hand this function a declaration alongside ``None``
+            allowed collections; that combination is constructed to isolate
+            forwarding, not something a call site produces. See
+            ``declared_knowledge_bases`` below.
         declared_knowledge_bases: Accepted for signature symmetry and never
             read here, for the same reason. This function reports "here is
             what is available" -- an unresolved declared name simply does
             not appear in the list, with no separate message field, and
-            never triggers the creator-existence probe.
+            never triggers the creator-existence probe. That holds whatever
+            is passed, so an artificially constructed declaration changes
+            nothing about the answer.
 
     Returns:
         ListKnowledgeBasesResult containing knowledge base information
@@ -522,6 +528,11 @@ async def _search_knowledge_base_impl(
             ``tool_args.allowed_collections`` and never
             ``tool_args.collections``, both of which are model-authored on
             every path (a declared schema field the model can overwrite).
+            Both chat call sites derive this and the tool's allowed
+            collections from that one stored field, so in production the
+            two are equal by construction; a test that pairs a declaration
+            with different allowed collections is isolating forwarding, and
+            this function's behaviour is defined for that pairing too.
 
     Returns:
         KnowledgeSearchResult with formatted search results
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -537,7 +537,6 @@ class _CreatorCollectionProbe:
                 # restore exactly the misclassification this call exists to
                 # remove.
                 team_hosted = await _team_names_hosted_on(self._agent_creator_user_id)
-                self._names = {c.name for c in creator_result.collections} - team_hosted
             except Exception:
                 logger.warning(
                     "Failed to resolve creator's personal collections while "
@@ -546,6 +545,28 @@ class _CreatorCollectionProbe:
                 )
                 self._names = set()
                 self.lookup_failed = True
+            else:
+                if creator_result.status == "success":
+                    self._names = {
+                        c.name for c in creator_result.collections
+                    } - team_hosted
+                else:
+                    # The listing reports an infrastructure failure by
+                    # returning status="error" with an empty collection list
+                    # instead of raising (RAG_tools/management/collections).
+                    # Read as a plain answer, that empty list says "the
+                    # creator holds nothing", and the caller would assert
+                    # the certain "does not exist" wording off a lookup that
+                    # never completed. Same treatment as a raised failure:
+                    # nothing is established, so the report must hedge.
+                    logger.warning(
+                        "Creator collection listing reported status=%s while "
+                        "classifying an unshared knowledge base: %s",
+                        creator_result.status,
+                        creator_result.message,
+                    )
+                    self._names = set()
+                    self.lookup_failed = True
         return name in self._names
 
 
@@ -619,9 +640,16 @@ async def _search_knowledge_base_impl(
         # Team-governed run WITH a declaration, on a deployment that has
         # actually installed the team-keyed hook: the declaration is the
         # authority for both what may be searched and what gets reported.
-        # Gated on this three-way conjunction, not merely on
-        # governing_team_id being set, for two independent reasons:
+        # Gated on this four-way conjunction, not merely on
+        # governing_team_id being set, for three independent reasons:
         #
+        # - an unauthenticated call (``user_id`` is None) never had a team
+        #   layer resolved for it at all: ``_list_visible_collections``
+        #   returns before any team resolution when the runner id is absent,
+        #   so the visible set it produced is the plain listing. Running the
+        #   rule on top of that would classify every declared name through a
+        #   creator probe made on nobody's behalf. Both chat call sites carry
+        #   a real runner id today; this conjunct keeps it that way;
         # - an empty declaration (the case where agent_config carried no
         #   agent at all) has nothing to authorise or verdict-derive against,
         #   so the declared-name rule does not apply and the search falls
@@ -655,15 +683,17 @@ async def _search_knowledge_base_impl(
         # This gate and the team layer's own gate in
         # ``_list_visible_collections`` are not the same predicate: they
         # share ``governing_team_id is not None and
-        # team_knowledge_base_hook_installed()``, but this one adds a third
-        # conjunct (``declared_names_for_this_run``) that the team layer's
-        # gate does not read. That is intentional, not a mismatch to close --
+        # team_knowledge_base_hook_installed()``, but this one adds two more
+        # conjuncts (``user_id is not None`` and
+        # ``declared_names_for_this_run``) that the team layer's gate does
+        # not read. That is intentional, not a mismatch to close --
         # the team layer's job is only "resolve onto the governing team",
         # which an empty declaration does not change; this gate's job is
         # "does the declared-name rule apply at all", which an empty
         # declaration answers no to.
         declared_name_rule_applies = (
-            governing_team_id is not None
+            user_id is not None
+            and governing_team_id is not None
             and team_knowledge_base_hook_installed()
             and bool(declared_names_for_this_run)
         )
@@ -817,37 +847,76 @@ async def _search_knowledge_base_impl(
                 ]
 
                 if unresolvable_names:
-                    # The creator-collection lookup raised during this call,
-                    # so every unresolved declared name is unclassified, not
-                    # established absent. Reported as one undifferentiated
-                    # outcome for all of them -- a lookup failure is not
+                    # The creator-collection lookup did not complete during
+                    # this call, so every unresolved declared name is
+                    # unclassified, not established absent. Reported as one
+                    # undifferentiated outcome -- a lookup failure is not
                     # per-name, so this sentence still says nothing about
-                    # which names the creator holds. The other two report
-                    # sets are necessarily empty here: the first failure
-                    # caches the creator's name set as empty and marks the
-                    # probe failed, so every later unresolved name takes
-                    # this same classification, and a name reaches the
-                    # missing set only on a run whose runner is the creator
-                    # -- which never calls the probe at all.
-                    summary = _TERMINAL_UNRESOLVABLE_MESSAGE.format(
-                        names=", ".join(unresolvable_names)
-                    )
+                    # which names the creator holds -- and scoped like the
+                    # other two report sets, so an explicit request is
+                    # answered about the names it asked for.
+                    #
+                    # The narrowed list cannot come out empty: with an
+                    # explicit request, reaching this branch means every
+                    # requested name was unresolved, because a requested
+                    # name the verdict loop admitted would have survived the
+                    # narrowing and left the search set non-empty; without
+                    # one, the scope is the whole declaration.
+                    #
+                    # The other two report sets are necessarily empty here:
+                    # the first failure caches the creator's name set as
+                    # empty and marks the probe failed, so every later
+                    # unresolved name takes this same classification, and a
+                    # name reaches the missing set only on a run whose
+                    # runner is the creator -- which never calls the probe
+                    # at all.
+                    #
+                    # The admitted clause renders on the same condition as
+                    # the mixed branch below, so a request narrowed away to
+                    # nothing still says what could have been asked for.
+                    failed_unresolvable_names = [
+                        name for name in unresolvable_names if name in reported_scope
+                    ]
+                    unresolvable_segments = [
+                        _TERMINAL_UNRESOLVABLE_MESSAGE.format(
+                            names=", ".join(failed_unresolvable_names)
+                        )
+                    ]
+                    if admitted_names:
+                        unresolvable_segments.append(
+                            "Available collections: "
+                            f"{', '.join(sorted(admitted_names))}."
+                        )
+                    summary = " ".join(unresolvable_segments)
                 elif not failed_unshared_names:
                     # Nothing in scope is the creator's own collection, so
                     # the two inherited terminal sentences are true as
                     # written and each renders byte for byte on its own
-                    # call style.
+                    # call style. The availability clause they carry is
+                    # dropped when the verdict loop admitted nothing, the
+                    # same as the mixed branch below: its source is the
+                    # admitted set, which this rule can empty where the
+                    # visible union it replaced never could, and a label
+                    # with an empty list after it tells the model nothing.
                     if explicit_request is not None:
                         summary = (
-                            f"Error: The following collections do not exist: {', '.join(sorted(explicit_request))}. "
-                            f"Available collections: {', '.join(sorted(admitted_names))}"
+                            "Error: The following collections do not exist: "
+                            f"{', '.join(sorted(explicit_request))}."
                         )
+                        if admitted_names:
+                            summary += (
+                                " Available collections: "
+                                f"{', '.join(sorted(admitted_names))}"
+                            )
                     else:
                         summary = (
-                            f"Error: None of the allowed collections exist. "
-                            f"Allowed: {', '.join(sorted(authorised))}. "
-                            f"Available: {', '.join(sorted(admitted_names))}"
+                            "Error: None of the allowed collections exist. "
+                            f"Allowed: {', '.join(sorted(authorised))}."
                         )
+                        if admitted_names:
+                            summary += (
+                                f" Available: {', '.join(sorted(admitted_names))}"
+                            )
                 else:
                     # At least one failed name is the creator's own
                     # collection, which does exist. Neither inherited
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -164,7 +164,7 @@ class KnowledgeSearchArgs(BaseModel):
     query: str = Field(description="The search query or question")
     collections: List[str] = Field(
         default=[],
-        description="Specific knowledge base collection names to search. Empty list uses allowed_collections if set, otherwise searches all collections.",
+        description="Specific knowledge base collection names to search. Empty list searches every knowledge base this agent is configured with: allowed_collections when set, all collections when not, and for an agent owned by a team, that agent's own stored knowledge base list.",
     )
     search_type: str = Field(
         default="hybrid",
@@ -183,7 +183,7 @@ class KnowledgeSearchArgs(BaseModel):
     )
     allowed_collections: Optional[List[str]] = Field(
         default=None,
-        description="Optional list of allowed collection names. Used as default when collections is empty.",
+        description="Optional list of allowed collection names. Used as default when collections is empty. Ignored for an agent owned by a team, whose own stored knowledge base list is the authority.",
     )
 
 
@@ -695,6 +695,11 @@ async def _search_knowledge_base_impl(
         if tool_args.allowed_collections:
             logger.info(f"   - Allowed collections: {tool_args.allowed_collections}")
 
+        # Collected by whichever branch below resolves the search set; a
+        # partial failure warns without stopping the rest of the agent's
+        # knowledge bases from being searched normally.
+        partial_failure_notes = []
+
         if declared_name_rule_applies:
             resolved_by_name = {c.name: c for c in collections_result.collections}
             authorised = set(declared_names_for_this_run)
@@ -872,8 +877,7 @@ async def _search_knowledge_base_impl(
                     summary = "Error: " + " ".join(segments)
                 return KnowledgeSearchResult(results=[], summary=summary)
 
-            collections_to_iterate: List[CollectionInfo] = to_search
-            partial_failure_notes = []
+            collections_to_iterate = to_search
             if unshared_names:
                 partial_failure_notes.append(_render_unshared_names(unshared_names))
             if missing_names or unresolvable_names:
@@ -912,7 +916,6 @@ async def _search_knowledge_base_impl(
             collections_to_iterate = [
                 c for c in collections_result.collections if c.name in collections_set
             ]
-            partial_failure_notes = []
             logger.info(f"Searching specific collections: {sorted(collections_set)}")
         elif tool_args.allowed_collections is not None:
             allowed_set = set(tool_args.allowed_collections)
@@ -935,11 +938,9 @@ async def _search_knowledge_base_impl(
             collections_to_iterate = [
                 c for c in collections_result.collections if c.name in valid_collections
             ]
-            partial_failure_notes = []
             logger.info(f"Searching allowed collections: {sorted(valid_collections)}")
         else:
             collections_to_iterate = collections_result.collections
-            partial_failure_notes = []
             logger.info("Searching all collections")
 
         # Build base search config (per-collection overrides happen below)
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -36,9 +36,10 @@ _PARTIAL_FAILURE_NOTE = (
     "bases were searched normally."
 )
 # The terminal counterpart of the note above, for the one case the certainty
-# wording would be a falsehood: the creator-collection lookup raised, so
-# nothing established that these names are absent -- only that they could not
-# be classified. Reuses the note's hedge; nothing was searched, so it does not
+# wording would be a falsehood: the creator-collection lookup did not
+# complete -- it raised, or returned a result whose status reports failure --
+# so nothing established that these names are absent, only that they could
+# not be classified. Reuses the note's hedge; nothing was searched, so it does not
 # borrow the note's second sentence.
 _TERMINAL_UNRESOLVABLE_MESSAGE = "Error: {names} could not be resolved for this agent."
 
@@ -532,7 +533,8 @@ class _CreatorCollectionProbe:
     def __init__(self, agent_creator_user_id: Optional[int]) -> None:
         self._agent_creator_user_id = agent_creator_user_id
         self._names: Optional[set] = None
-        # Whether the one lookup this probe makes raised. Read by the caller
+        # Whether the one lookup this probe makes failed -- raised, or
+        # returned a result whose status reports failure. Read by the caller
         # to tell "the creator does not hold this name" apart from "nobody
         # asked" -- the failure degrades to "not held" for the search
         # decision (unchanged), but the report must not claim an absence
@@ -821,7 +823,8 @@ async def _search_knowledge_base_impl(
                         unshared_names.append(name)
                     elif creator_probe.lookup_failed:
                         # Not "missing": the lookup that would have decided
-                        # raised. The search outcome is the same either way
+                        # did not complete (raised, or reported failure
+                        # through its status). The search outcome is the same either way
                         # (nothing resolves), only the report differs.
                         unresolvable_names.append(name)
                     else:
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -435,13 +435,59 @@ def _resolve_declared_name(
     return _UNRESOLVED
 
 
+async def _team_names_hosted_on(storage_user_id: int) -> set:
+    """Names on ``storage_user_id``'s tenant that a team owns, not the user.
+
+    A team's knowledge base lives in some member's storage namespace, so a
+    raw ``list_collections`` for that member returns the team's rows
+    alongside their own, and nothing on the row itself separates them:
+    ``CollectionInfo.ownership`` is ``"personal"`` on every row
+    ``list_collections`` produces -- the one place it becomes ``"team"`` is
+    the overlay in ``_list_visible_collections``, which reads the very hook
+    this function reads.
+
+    Asked through the runner-keyed visibility hook, called with the
+    tenant's own id, because it is the only seam that answers "which
+    knowledge bases does a team own" for a team other than the governing
+    one. The answer only ever narrows what may be called this user's own,
+    and is never rendered: a name it removes is reported as absent instead
+    of as the creator's, so the report can say less about the creator than
+    before, never more.
+
+    Bounded by what the seam can answer: it reports the teams this user is
+    a member of *now*, so a team's row hosted on the tenant of someone who
+    has since left that team still counts as personal. Closing that would
+    need a "which team owns this name" question the hook contract does not
+    have, and answering it would widen what a run can learn.
+    """
+    from ....web.services.db_runtime import run_db_io_cancellation_safe
+    from ....web.services.knowledge_base_team_scope import (
+        has_knowledge_base_visibility_hook,
+        visible_team_knowledge_bases,
+    )
+
+    if not has_knowledge_base_visibility_hook():
+        # Standalone xagent, and any deployment that installed only the
+        # team-keyed hook: no team owns anything here, so every row on the
+        # tenant is the user's own.
+        return set()
+    refs = await run_db_io_cancellation_safe(
+        lambda: visible_team_knowledge_bases(None, int(storage_user_id))
+    )
+    # Only the rows physically hosted on this tenant. A team row stored
+    # elsewhere never appeared in this tenant's listing, and excluding its
+    # name would drop a same-named collection this user really does own.
+    return {ref.name for ref in refs if ref.storage_user_id == storage_user_id}
+
+
 class _CreatorCollectionProbe:
     """Memoised lookup of the agent creator's own collection names.
 
-    At most one ``list_collections`` call for the creator's id per search
-    call, regardless of how many declared names need classifying -- the
-    lookup result is cached after the first call, including a failure
-    (cached as "nothing"). A failure degrades to "not held" and never
+    At most one creator lookup per search call -- one collection listing
+    and one team-ownership question, both memoised together -- regardless
+    of how many declared names need classifying. The lookup result is
+    cached after the first call, including a failure (cached as
+    "nothing"). A failure degrades to "not held" and never
     raises: a probe that cannot complete must not block the search, and
     must not be distinguishable from "the creator does not have this
     collection" (see the module docstring's identity-disclosure limits).
@@ -484,7 +530,14 @@ class _CreatorCollectionProbe:
                 creator_result = await list_collections(
                     user_id=self._agent_creator_user_id, is_admin=False
                 )
-                self._names = {c.name for c in creator_result.collections}
+                # Inside the same try on purpose: if the team lookup fails,
+                # which of these rows are the creator's own is unknown, and
+                # the caller must hedge rather than assert the creator
+                # holds them. Falling back to the unfiltered listing would
+                # restore exactly the misclassification this call exists to
+                # remove.
+                team_hosted = await _team_names_hosted_on(self._agent_creator_user_id)
+                self._names = {c.name for c in creator_result.collections} - team_hosted
             except Exception:
                 logger.warning(
                     "Failed to resolve creator's personal collections while "
@@ -660,7 +713,14 @@ async def _search_knowledge_base_impl(
                     return KnowledgeSearchResult(
                         results=[],
                         summary=f"Error: The following collections are not allowed: {', '.join(sorted(disallowed))}. "
-                        f"Allowed collections: {', '.join(sorted(authorised & available_names))}",
+                        # The clause lists the declaration itself, not its
+                        # intersection with the visible union: that union
+                        # seeds the runner's own same-named personal
+                        # collections, which the verdict loop below then
+                        # refuses to search -- advertising them here as
+                        # allowed and rejecting them one step later is the
+                        # same contradiction the terminal branch avoids.
+                        f"Allowed collections: {', '.join(sorted(authorised))}",
                     )
                 explicit_request = requested_set
                 logger.info(f"Searching specific collections: {sorted(requested_set)}")
@@ -727,68 +787,89 @@ async def _search_knowledge_base_impl(
                 # which would fall through to the generic search-miss text
                 # below instead of the two outcomes this branch exists to
                 # report.
-                # Both branches list "Available" from the verdict-admitted
-                # names, never from ``authorised & available_names``. That
-                # intersection reports two sets of names it must not: the
-                # governing team's collections the agent never declared
-                # (``_list_visible_collections`` merges the team's full
-                # catalogue in before any declared-name filter runs), and --
-                # on this branch specifically -- the runner's own same-named
-                # personal copies, which seed the union and are precisely the
-                # collections the rule refuses to search. On the no-explicit
-                # branch nothing was admitted, so the clause renders empty;
-                # on the explicit branch the admitted set is the one taken
-                # before the narrowing above, so a run whose requested names
-                # all failed still tells the model what it may ask for.
+                # The summary is derived from the verdicts, one report set
+                # per verdict class, so no name is described twice and none
+                # is described as both absent and held. "Available" is
+                # always the verdict-admitted set, never
+                # ``authorised & available_names``: that intersection
+                # reports two sets of names it must not -- the governing
+                # team's collections the agent never declared, and the
+                # runner's own same-named personal copies, which seed the
+                # visible union and are precisely what the rule refuses to
+                # search. The admitted set is the one taken before the
+                # narrowing above, so a run whose requested names all
+                # failed still tells the model what it may ask for.
+                reported_scope = (
+                    explicit_request
+                    if explicit_request is not None
+                    else set(declared_names_for_this_run)
+                )
+                failed_unshared_names = [
+                    name for name in unshared_names if name in reported_scope
+                ]
+                failed_missing_names = [
+                    name for name in missing_names if name in reported_scope
+                ]
+
                 if unresolvable_names:
                     # The creator-collection lookup raised during this call,
                     # so every unresolved declared name is unclassified, not
                     # established absent. Reported as one undifferentiated
                     # outcome for all of them -- a lookup failure is not
                     # per-name, so this sentence still says nothing about
-                    # which names the creator holds. ``unshared_names`` is
-                    # necessarily empty here: the first failure caches the
-                    # creator's name set as empty, so no later name can be
-                    # classified as held.
+                    # which names the creator holds. The other two report
+                    # sets are necessarily empty here: the first failure
+                    # caches the creator's name set as empty and marks the
+                    # probe failed, so every later unresolved name takes
+                    # this same classification, and a name reaches the
+                    # missing set only on a run whose runner is the creator
+                    # -- which never calls the probe at all.
                     summary = _TERMINAL_UNRESOLVABLE_MESSAGE.format(
                         names=", ".join(unresolvable_names)
                     )
-                    failed_unshared_names: List[str] = []
-                elif explicit_request is not None:
-                    summary = (
-                        f"Error: The following collections do not exist: {', '.join(sorted(explicit_request))}. "
-                        f"Available collections: {', '.join(sorted(admitted_names))}"
-                    )
-                    failed_unshared_names = [
-                        name for name in unshared_names if name in explicit_request
-                    ]
+                elif not failed_unshared_names:
+                    # Nothing in scope is the creator's own collection, so
+                    # the two inherited terminal sentences are true as
+                    # written and each renders byte for byte on its own
+                    # call style.
+                    if explicit_request is not None:
+                        summary = (
+                            f"Error: The following collections do not exist: {', '.join(sorted(explicit_request))}. "
+                            f"Available collections: {', '.join(sorted(admitted_names))}"
+                        )
+                    else:
+                        summary = (
+                            f"Error: None of the allowed collections exist. "
+                            f"Allowed: {', '.join(sorted(authorised))}. "
+                            f"Available: {', '.join(sorted(admitted_names))}"
+                        )
                 else:
-                    summary = (
-                        f"Error: None of the allowed collections exist. "
-                        f"Allowed: {', '.join(sorted(authorised))}. "
-                        f"Available: {', '.join(sorted(admitted_names))}"
-                    )
-                    failed_unshared_names = unshared_names
-                # A total failure keeps the existing terminal wording
-                # unchanged (above) -- this branch does not decide whether
-                # a failed name happened to be the creator's own unshared
-                # collection, it only decides that nothing is searchable.
-                # Whichever of the failed names the verdict loop already
-                # classified as the creator's own unshared collection still
-                # gets that message appended, with the "Error:" prefix and
-                # the empty result set both preserved: this is the terminal
-                # path, not the partial-failure warnings channel, so it
-                # does not fold through collection_warnings.
-                if failed_unshared_names:
-                    # A bare space here would leave the appended sentence
-                    # reading like one more entry in the comma-separated
-                    # "Available: ..." list the base text ends with -- both
-                    # to a human reader and to the model consuming this
-                    # summary. The period closes that list before the new
-                    # sentence starts.
-                    summary = (
-                        summary + ". " + _render_unshared_names(failed_unshared_names)
-                    )
+                    # At least one failed name is the creator's own
+                    # collection, which does exist. Neither inherited
+                    # sentence can carry that: one is a statement about
+                    # every allowed collection, the other enumerates the
+                    # whole request, and both would assert the absence of a
+                    # name this same summary goes on to say the creator
+                    # holds. The absence sentence therefore enumerates only
+                    # the names actually classified missing and is dropped
+                    # when there are none; the availability clause is
+                    # dropped when nothing was admitted rather than
+                    # rendering an empty list. Same shape on both call
+                    # styles: each report set lists only its own names, and
+                    # a name appears in exactly one of them.
+                    segments: List[str] = []
+                    if failed_missing_names:
+                        segments.append(
+                            "The following collections do not exist: "
+                            f"{', '.join(sorted(failed_missing_names))}."
+                        )
+                    segments.append(_render_unshared_names(failed_unshared_names))
+                    if admitted_names:
+                        segments.append(
+                            "Available collections: "
+                            f"{', '.join(sorted(admitted_names))}."
+                        )
+                    summary = "Error: " + " ".join(segments)
                 return KnowledgeSearchResult(results=[], summary=summary)
 
             collections_to_iterate: List[CollectionInfo] = to_search
--- a/src/xagent/core/tools/core/document_search.py
+++ b/src/xagent/core/tools/core/document_search.py
@@ -8,7 +8,7 @@ from pydantic import BaseModel, Field
 
 from ....config import get_kb_search_timeout_seconds
 from .knowledge_base_scope import KnowledgeBaseScopeError
-from .RAG_tools.core.schemas import ListCollectionsResult
+from .RAG_tools.core.schemas import CollectionInfo, ListCollectionsResult
 from .RAG_tools.management.collections import list_collections
 from .RAG_tools.pipelines.document_search import run_document_search
 
@@ -22,6 +22,25 @@ logger = logging.getLogger(__name__)
 # filter silently stops matching - the readonly notice reappears in summaries,
 # which the readonly tests in test_document_search_collection_concurrency catch.
 _READONLY_WARNING_PREFIX = "READONLY_MODE:"
+
+# The fixed, unparaphrasable strings a team-governed search reports through.
+# See the module's callers for when each one is used; the wording itself is
+# a reviewed user-facing contract and must not be edited casually.
+_UNSHARED_CREATOR_KB_MESSAGE = (
+    "{name} is a personal knowledge base belonging to this agent's creator and has not "
+    "been shared with the team. Ask the agent's owner to share it with the team to make "
+    "it available here."
+)
+_PARTIAL_FAILURE_NOTE = (
+    "Note: {names} could not be resolved for this agent. The agent's other knowledge "
+    "bases were searched normally."
+)
+# The terminal counterpart of the note above, for the one case the certainty
+# wording would be a falsehood: the creator-collection lookup raised, so
+# nothing established that these names are absent -- only that they could not
+# be classified. Reuses the note's hedge; nothing was searched, so it does not
+# borrow the note's second sentence.
+_TERMINAL_UNRESOLVABLE_MESSAGE = "Error: {names} could not be resolved for this agent."
 
 if TYPE_CHECKING:
     from .RAG_tools.kb import KBToolCompatibilityFacade
@@ -233,12 +252,13 @@ async def _list_knowledge_bases_impl(
             builds the search tool alone, and an empty list builds no
             knowledge tool at all -- and both chat call sites derive that
             value and the declaration from the same stored field, so on this
-            path there is no declared name to classify at all. See
-            ``declared_knowledge_bases`` below.
+            path there is no declared name to classify as "unshared" versus
+            "missing". See ``declared_knowledge_bases`` below.
         declared_knowledge_bases: Accepted for signature symmetry and never
             read here, for the same reason. This function reports "here is
             what is available" -- an unresolved declared name simply does
-            not appear in the list, with no separate message field.
+            not appear in the list, with no separate message field, and
+            never triggers the creator-existence probe.
 
     Returns:
         ListKnowledgeBasesResult containing knowledge base information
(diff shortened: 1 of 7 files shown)


src/xagent/core/tools/core/document_search.py
```diff
@@ -287,11 +629,169 @@ async def _search_knowledge_base_impl(
...
+                    # necessarily empty here: the first failure caches the
+                    # creator's name set as empty, so no later name can be
+                    # classified as held.
+                    summary = _TERMINAL_UNRESOLVABLE_MESSAGE.format(
+                        names=", ".join(unresolvable_names)
+                    )
+                    failed_unshared_names: List[str] = []
+                elif explicit_request is not None:
+                    summary = (
+                        f"Error: The following collections do not exist: {', '.join(sorted(explicit_request))}. "
```
rogercloud: On an explicit-request total failure, this branch renders "Error: The following collections do not exist: <all requested names>", and then at L777 appends the sharing-gap sentence for every name the verdict loop classified as the creator's own unshared collection — the same name is asserted both non-existent ("do not exist") and existing ("is a personal knowledge base belonging to this agent's creator") in one summary. The default-style branch (L754-757, "None of the allowed collections exist") has the same defect. The PR's own test docstring (test_document_search_team_declared_names.py:547-549) admits "does not exist" is "false of this one" for a creator-held name, and the test pins the contradictory output. Fix: exclude unshared names from the "do not exist" enumeration (e.g. report `explicit_request - set(failed_unshared_names)`), or reword that sentence to "could not be searched" for them, and update the pinning test.

src/xagent/core/tools/core/document_search.py
```diff
@@ -236,40 +340,278 @@ async def search_knowledge_base(
...
+        # decision (unchanged), but the report must not claim an absence
+        # that was never established.
+        self.lookup_failed = False
+
+    async def holds(self, name: str) -> bool:
+        if self._agent_creator_user_id is None:
+            return False
+        if self._names is None:
+            try:
+                creator_result = await list_collections(
```
rogercloud, line 549: `_CreatorCollectionProbe.holds` classifies via raw `list_collections(user_id=creator)` with no ownership filter — the result is the creator's whole storage tenant, while `ownership == "team"` only materializes in `_list_visible_collections`' overlay, which this probe bypasses. Production stores team collections on a member's tenant (`storage_user_id` is a member id), so a non-governing team's KB physically stored on the creator's tenant, declared by the governing team's agent, is classified "held by the creator" — the summary then falsely asserts "a personal knowledge base belonging to this agent's creator" and aims its remediation at the creator for a KB the creator's team does not own. Legitimate configuration, not an attacker construction; the search outcome is unchanged (name skipped). Fix: filter `creator_result.collections` to `ownership != "team"` before building `_names` (or document why that is impossible). The PR's own `test_no_cross_team_kb_borrowing` pins only the negative direction — add the positive case.

src/xagent/core/tools/core/document_search.py
```diff
@@ -287,11 +629,169 @@ async def _search_knowledge_base_impl(
...
+                # tool_args.allowed_collections, because that field is
+                # model-authored on every path and setdefault lets a
+                # model-supplied value win outright over the agent's own
+                # configured one.
+                disallowed = requested_set - authorised
+                if disallowed:
+                    return KnowledgeSearchResult(
+                        results=[],
+                        summary=f"Error: The following collections are not allowed: {', '.join(sorted(disallowed))}. "
+                        f"Allowed collections: {', '.join(sorted(authorised & available_names))}",
```
rogercloud: The gate's "Allowed collections:" clause renders `authorised & available_names` — the exact intersection the comment at L717-725 condemns for the terminal branches ("reports two sets of names it must not: the governing team's collections the agent never declared ... and the runner's own same-named personal copies"). `available_names` is the pre-verdict visible union (it seeds the runner's own collections), and the gate returns before the verdict loop, so a non-creator runner holding a same-named personal copy of a declared name sees it listed "Allowed" and is then refused by the verdict loop. The rejection decision itself is correct and this is an error-path message only, but the clause is misleading. Fix: render the clause from the declaration (`authorised`) or the admitted set, and pin the clause content in a test (only the "not allowed" substring is asserted today).

src/xagent/core/tools/core/document_search.py
```diff
@@ -287,11 +629,169 @@ async def _search_knowledge_base_impl(
         if tool_args.allowed_collections:
             logger.info(f"   - Allowed collections: {tool_args.allowed_collections}")
```
rogercloud, line 759: On team-governed runs with a non-empty declaration, `allowed_collections` is only logged here and never read — the pre-PR narrowing branch is unreachable on those runs. The model-facing schema (L184) still promises "Used as default when collections is empty", which is false on governed runs; model-authored narrowing is silently dropped. Bounded by the declaration, this is not a security widening (the model can still narrow via explicit `collections`), and the behavior is intentional and test-pinned — the residual is the stale model-facing contract. Fix: update the schema description (and the `collections` field's "Empty list uses allowed_collections" wording at L165) to state that on team-governed runs the agent's stored declaration is the authority. (Anchored here, the governed branch's log-only read, because the schema line itself is outside this diff's hunks.)

tests/core/tools/core/test_document_search_team_declared_names.py
```diff
@@ -0,0 +1,1137 @@
...
+        monkeypatch,
+        runner_id=MEMBER,
+        declared=["handbook"],
+        collections=["victim"],
+        allowed_collections=["victim"],
+    )
+
+    assert spy.calls_for(CREATOR) == 0
+    assert result.results == []
+    assert "Error:" in result.summary and "not allowed" in result.summary
```
rogercloud, line 1113: This test exercises the declared-name "not allowed" gate (declared=["handbook"], collections=["victim"]) but asserts only the "Error:"/"not allowed" substrings — never the "Allowed collections:" clause content, and nothing covers a team-owned undeclared name in an explicit request. Several load-bearing criteria are untested: (1) the gate clause content — reverting document_search.py:650 to the visible union would leak undeclared team KB names with no failing test; (2) the partial-path probe-failure wording (both creator-listing-failure tests at L1011-1081 install `{TEAM: []}`, so everything fails terminally); (3) `is_admin=True` governed runs — the `_search` helper hardcodes `is_admin=False` at L148; (4) a team-owned undeclared name in explicit collections — a gate change whitelisting team-owned names would pass this suite and leak exactly the information the PR hides; (5) `governing_team_id=None` + installed hook + non-empty declaration — the "No governing team" section header (L953-962) has no test beneath it. Add pins for the clause content and at least the admin-run and team-owned-undeclared cases.

rogercloud (changes requested):
## Summary

This PR makes a team-governed agent's `knowledge_bases` declaration resolve against its own governing team: every declared name now gets an explicit verdict — the team's copy (team-owned), the creator's own collection (runner is the creator), nothing searchable plus a sharing-gap sentence (creator holds but did not share), "does not exist" (nobody holds), or no-absence wording (creator lookup failed). The search set is built from those verdicts rather than from the visible union (which would seed the runner's own same-named personal copies back in), the "not allowed" gate is re-based on the stored declaration instead of the model-authored `allowed_collections`, and deployments without the team-keyed hook keep the legacy behavior byte-for-byte. The rule is pinned by a new 1,137-line test file. Blocking: yes — recommended event: REQUEST_CHANGES.

## Approach verdict

The design is sound. Key properties verified: the rule is keyed on per-name verdicts, never on the complement of the visible union; the gate requires a governing team, an installed team-keyed hook and a non-empty declaration, so unhooked deployments are untouched; the creator probe is memoised once, bound to the stored declaration, and degrades to "not held" with no-absence wording. Two reservations survived verification as design notes rather than defects: (1) on a team-governed run with an **empty declaration**, the run fail-opens to the legacy gate/visibility entirely — intended and pinned by `test_team_governed_empty_declaration_keeps_legacy_not_allowed_gate`, but worth an explicit acknowledgment in the module docstring; (2) the `name not in declared_names` guard (`document_search.py:414`) is optional defense-in-depth that never fires on the sole caller — acceptable to keep, honestly documented, and also listed under simplifications below.

## Findings (severity order)

### MAJOR F1 — Terminal summary self-contradiction (`document_search.py:746`, `:754`, `:769-778`)

On an explicit-request total failure, the terminal branch renders `Error: The following collections do not exist: <all requested names>` (`document_search.py:746`) and then appends the sharing-gap sentence for every name the verdict loop classified as the creator's own unshared collection (`:777`) — the same name is asserted both non-existent and existing in one message. The default-style variant (`:754` "None of the allowed collections exist") has the same defect. The PR's own test docstring (`test_document_search_team_declared_names.py:547-549`) admits "does not exist" is "false of this one" for a creator-held name, and the test pins the contradictory output. **Fix:** exclude unshared names from the "do not exist" enumeration (report `explicit_request - set(failed_unshared_names)` there), or reword that sentence to "could not be searched" for them; update the pinning test with the fix.

### MAJOR F2 — Creator probe lacks an ownership filter (`document_search.py:471-479`)

`_CreatorCollectionProbe.holds` classifies via raw `list_collections(user_id=creator)` with no ownership filter — the result is the creator's whole storage tenant, while `ownership == "team"` only materializes in `_list_visible_collections`' overlay, which the probe bypasses. Production stores team collections on a member's tenant (`storage_user_id` is a member id), so a **non-governing** team's collection physically stored on the creator's tenant, declared by the governing team's agent, is classified "held by the creator" → the summary falsely asserts "a personal knowledge base belonging to this agent's creator" and aims its remediation at the creator for a KB the creator's team does not own. This is a legitimate configuration, not an attacker construction (the same-team-editable declaration is acknowledged in the probe docstring); the search outcome is unchanged (name skipped), so it is not critical. **Fix:** filter `creator_result.collections` to `ownership != "team"` before building `_names` (or document why that is impossible). The PR's own `test_no_cross_team_kb_borrowing` pins only the negative direction; add the positive case.

### MINOR F3 — Gate "Allowed collections:" clause uses the intersection the comment condemns (`document_search.py:650`)

The re-based gate's clause renders `authorised & available_names` — the exact intersection the PR's own comment at `document_search.py:717-725` condemns for the terminal branches. `available_names` is the pre-verdict visible union (it seeds the runner's own collections), and the gate returns before the verdict loop, so a non-creator runner holding a same-named personal copy of a declared name sees it listed "Allowed" and is then refused by the verdict loop. Informational error-path message only; the rejection decision itself is correct. **Fix:** render the clause from the declaration (`authorised`) or the admitted set, and add a test pinning the clause content (only the "not allowed" substring is asserted today).

### MINOR F4 — Stale model-facing schema contract on governed runs (`document_search.py:184`, `:630`)

On team-governed runs with a non-empty declaration, `tool_args.allowed_collections` is only logged (`document_search.py:629-630`), never read — the pre-PR narrowing branch is unreachable on those runs. The model-facing schema still promises "Used as default when collections is empty" (`:184`), which is false on governed runs; model-authored narrowing is silently dropped. Bounded by the declaration, this is not a security widening — the model can still narrow via explicit `collections` — and the behavior is intentional, security-motivated and test-pinned. **Fix:** update the schema description (and the `collections` field's "Empty list uses allowed_collections" wording at `:165`) to state that on team-governed runs the agent's stored declaration is the authority.

### MINOR F6 — Test gaps (`test_document_search_team_declared_names.py`)

Five load-bearing criteria are untested: (1) the governed gate's "Allowed collections:" clause content — reverting `document_search.py:650` to the visible union would leak undeclared team KB names with no failing test (only the "not allowed" substring is asserted at `:682` and `:858`); (2) the partial-path probe-failure wording ("could not be resolved" Note alongside still-searched team KBs) — both creator-listing-failure tests (`:1011-1081`) install `{TEAM: []}` so everything fails terminally; (3) `is_admin=True` governed runs — the `_search` helper hardcodes `is_admin=False` (`:148`); (4) a team-owned undeclared name in explicit collections — a gate change whitelisting team-owned names would pass the suite and leak exactly the information this PR hides; (5) `governing_team_id=None` + installed hook + non-empty declaration — the "No governing team" section header (`:953-962`) has no test beneath it.

- **L414 yagni:** the `name not in declared_names` guard never fires — the sole caller iterates `declared_names_for_this_run` and passes `authorised = set(declared_names_for_this_run)` (`:634`), so membership holds by construction (the docstring admits it). Drop the guard and the `declared_names` parameter (`:383`) with its speculative `set[str] | frozenset[str]` union. Honestly documented defense-in-depth for the enumeration oracle — optional; the author may reasonably decline.
- **L1046 shrink:** `test_creator_listing_failure_hedges_on_the_explicit_call_style_too` duplicates `test_creator_listing_failure_reports_unresolved_not_absent` (`:1011`) — the 10-line `_broken_list_collections` fixture is copied verbatim and the bodies differ only in `collections=["handbook"]` plus a strict 3-assertion subset. Merge into one `@pytest.mark.parametrize("explicit", [False, True])` test (the file's own convention) with the union of assertions — verified lossless.
- **L782 shrink:** `partial_failure_notes = []` is repeated identically in all four branches (`:782`, `:821`, `:844`, `:848`); hoist one initialization above the if/elif chain.
- **L358 shrink:** `_Unresolved` is over-built for one module singleton — `__slots__ = ()` (`:367`) is dead and `__repr__` (`:369`) is a `# pragma: no cover` debugging aid. `_UNRESOLVED = object()` with `verdict is _UNRESOLVED` covers the same case in ~9 fewer lines.
- **L743 shrink:** `failed_unshared_names: List[str] = []` in the unresolvable branch is provably always empty — a failed lookup caches `_names = set()`, so no later name can land in `unshared_names` — and the shared `if failed_unshared_names:` append (`:769`) never fires for it. Early-return the `_TERMINAL_UNRESOLVABLE_MESSAGE` summary from that branch and delete the forced-empty variable.
- **L781 shrink:** `collections_to_iterate: List[CollectionInfo] = to_search` re-annotates an alias whose type is already declared at `:655`; drop the `: List[CollectionInfo]` annotation. Keep the alias itself — it is the name bridge into the pre-existing shared tail.

**net: ~30 lines possible** (drop the annotation, hoist one init, `object()` sentinel, merged test, early-return restructure; the L414 guard is optional).

**Blocking: yes — REQUEST_CHANGES.**

Blocking issues:
- F1 — `document_search.py:746` — major — terminal summary self-contradiction [new]
- F2 — `document_search.py:476-479` — major — creator probe lacks an ownership filter [new]

CI: all checks green except Pytest Fast (web), still in progress at pre-flight time.

rogercloud (changes requested):
## Summary

This PR makes a team-governed agent's `knowledge_bases` declaration resolve against its own governing team: every declared name now gets an explicit verdict — the team's copy (team-owned), the creator's own collection (runner is the creator), nothing searchable plus a sharing-gap sentence (creator holds but did not share), "does not exist" (nobody holds), or no-absence wording (creator lookup failed). The search set is built from those verdicts rather than from the visible union (which would seed the runner's own same-named personal copies back in), the "not allowed" gate is re-based on the stored declaration instead of the model-authored `allowed_collections`, and deployments without the team-keyed hook keep the legacy behavior byte-for-byte. The rule is pinned by a new 1,137-line test file. Blocking: yes — recommended event: REQUEST_CHANGES.

## Approach verdict

The design is sound. Key properties verified: the rule is keyed on per-name verdicts, never on the complement of the visible union; the gate requires a governing team, an installed team-keyed hook and a non-empty declaration, so unhooked deployments are untouched; the creator probe is memoised once, bound to the stored declaration, and degrades to "not held" with no-absence wording. Two reservations survived verification as design notes rather than defects: (1) on a team-governed run with an **empty declaration**, the run fail-opens to the legacy gate/visibility entirely — intended and pinned by `test_team_governed_empty_declaration_keeps_legacy_not_allowed_gate`, but worth an explicit acknowledgment in the module docstring; (2) the `name not in declared_names` guard (`document_search.py:414`) is optional defense-in-depth that never fires on the sole caller — acceptable to keep, honestly documented, and also listed under simplifications below.

## Findings (severity order)

### MAJOR F1 — Terminal summary self-contradiction (`document_search.py:746`, `:754`, `:769-778`)

On an explicit-request total failure, the terminal branch renders `Error: The following collections do not exist: <all requested names>` (`document_search.py:746`) and then appends the sharing-gap sentence for every name the verdict loop classified as the creator's own unshared collection (`:777`) — the same name is asserted both non-existent and existing in one message. The default-style variant (`:754` "None of the allowed collections exist") has the same defect. The PR's own test docstring (`test_document_search_team_declared_names.py:547-549`) admits "does not exist" is "false of this one" for a creator-held name, and the test pins the contradictory output. **Fix:** exclude unshared names from the "do not exist" enumeration (report `explicit_request - set(failed_unshared_names)` there), or reword that sentence to "could not be searched" for them; update the pinning test with the fix.

### MAJOR F2 — Creator probe lacks an ownership filter (`document_search.py:471-479`)

`_CreatorCollectionProbe.holds` classifies via raw `list_collections(user_id=creator)` with no ownership filter — the result is the creator's whole storage tenant, while `ownership == "team"` only materializes in `_list_visible_collections`' overlay, which the probe bypasses. Production stores team collections on a member's tenant (`storage_user_id` is a member id), so a **non-governing** team's collection physically stored on the creator's tenant, declared by the governing team's agent, is classified "held by the creator" → the summary falsely asserts "a personal knowledge base belonging to this agent's creator" and aims its remediation at the creator for a KB the creator's team does not own. This is a legitimate configuration, not an attacker construction (the same-team-editable declaration is acknowledged in the probe docstring); the search outcome is unchanged (name skipped), so it is not critical. **Fix:** filter `creator_result.collections` to `ownership != "team"` before building `_names` (or document why that is impossible). The PR's own `test_no_cross_team_kb_borrowing` pins only the negative direction; add the positive case.

### MINOR F3 — Gate "Allowed collections:" clause uses the intersection the comment condemns (`document_search.py:650`)

The re-based gate's clause renders `authorised & available_names` — the exact intersection the PR's own comment at `document_search.py:717-725` condemns for the terminal branches. `available_names` is the pre-verdict visible union (it seeds the runner's own collections), and the gate returns before the verdict loop, so a non-creator runner holding a same-named personal copy of a declared name sees it listed "Allowed" and is then refused by the verdict loop. Informational error-path message only; the rejection decision itself is correct. **Fix:** render the clause from the declaration (`authorised`) or the admitted set, and add a test pinning the clause content (only the "not allowed" substring is asserted today).

### MINOR F4 — Stale model-facing schema contract on governed runs (`document_search.py:184`, `:630`)

On team-governed runs with a non-empty declaration, `tool_args.allowed_collections` is only logged (`document_search.py:629-630`), never read — the pre-PR narrowing branch is unreachable on those runs. The model-facing schema still promises "Used as default when collections is empty" (`:184`), which is false on governed runs; model-authored narrowing is silently dropped. Bounded by the declaration, this is not a security widening — the model can still narrow via explicit `collections` — and the behavior is intentional, security-motivated and test-pinned. **Fix:** update the schema description (and the `collections` field's "Empty list uses allowed_collections" wording at `:165`) to state that on team-governed runs the agent's stored declaration is the authority.

### MINOR F6 — Test gaps (`test_document_search_team_declared_names.py`)

Five load-bearing criteria are untested: (1) the governed gate's "Allowed collections:" clause content — reverting `document_search.py:650` to the visible union would leak undeclared team KB names with no failing test (only the "not allowed" substring is asserted at `:682` and `:858`); (2) the partial-path probe-failure wording ("could not be resolved" Note alongside still-searched team KBs) — both creator-listing-failure tests (`:1011-1081`) install `{TEAM: []}` so everything fails terminally; (3) `is_admin=True` governed runs — the `_search` helper hardcodes `is_admin=False` (`:148`); (4) a team-owned undeclared name in explicit collections — a gate change whitelisting team-owned names would pass the suite and leak exactly the information this PR hides; (5) `governing_team_id=None` + installed hook + non-empty declaration — the "No governing team" section header (`:953-962`) has no test beneath it.

- **L414 yagni:** the `name not in declared_names` guard never fires — the sole caller iterates `declared_names_for_this_run` and passes `authorised = set(declared_names_for_this_run)` (`:634`), so membership holds by construction (the docstring admits it). Drop the guard and the `declared_names` parameter (`:383`) with its speculative `set[str] | frozenset[str]` union. Honestly documented defense-in-depth for the enumeration oracle — optional; the author may reasonably decline.
- **L1046 shrink:** `test_creator_listing_failure_hedges_on_the_explicit_call_style_too` duplicates `test_creator_listing_failure_reports_unresolved_not_absent` (`:1011`) — the 10-line `_broken_list_collections` fixture is copied verbatim and the bodies differ only in `collections=["handbook"]` plus a strict 3-assertion subset. Merge into one `@pytest.mark.parametrize("explicit", [False, True])` test (the file's own convention) with the union of assertions — verified lossless.
- **L782 shrink:** `partial_failure_notes = []` is repeated identically in all four branches (`:782`, `:821`, `:844`, `:848`); hoist one initialization above the if/elif chain.
- **L358 shrink:** `_Unresolved` is over-built for one module singleton — `__slots__ = ()` (`:367`) is dead and `__repr__` (`:369`) is a `# pragma: no cover` debugging aid. `_UNRESOLVED = object()` with `verdict is _UNRESOLVED` covers the same case in ~9 fewer lines.
- **L743 shrink:** `failed_unshared_names: List[str] = []` in the unresolvable branch is provably always empty — a failed lookup caches `_names = set()`, so no later name can land in `unshared_names` — and the shared `if failed_unshared_names:` append (`:769`) never fires for it. Early-return the `_TERMINAL_UNRESOLVABLE_MESSAGE` summary from that branch and delete the forced-empty variable.
- **L781 shrink:** `collections_to_iterate: List[CollectionInfo] = to_search` re-annotates an alias whose type is already declared at `:655`; drop the `: List[CollectionInfo]` annotation. Keep the alias itself — it is the name bridge into the pre-existing shared tail.

**net: ~30 lines possible** (drop the annotation, hoist one init, `object()` sentinel, merged test, early-return restructure; the L414 guard is optional).

**Blocking: yes — REQUEST_CHANGES.**

Blocking issues:
- F1 — `document_search.py:746` — major — terminal summary self-contradiction [new]
- F2 — `document_search.py:476-479` — major — creator probe lacks an ownership filter [new]

CI: all checks green except Pytest Fast (web), still in progress at pre-flight time.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Agreed, and fixed in the "give each declared knowledge base one verdict in the
summary" commit — not by trimming names out of the old sentences, but by
deriving the terminal summary from the verdicts so no name is described twice.
Three shapes now:

- nothing in scope is the creator's own: both inherited sentences render byte
  for byte, on their own call style, exactly as before;
- at least one failed name is the creator's own unshared collection: the
  summary lists only the names actually classified missing under "do not
  exist" (and drops that clause when there are none), then one sharing-gap
  sentence per creator-held name, then the admitted set when there is one;
- the creator lookup raised: the hedged sentence, unchanged.

The default call style's blanket "None of the allowed collections exist" has
no name slot, so it cannot be made true when one of the names is the
creator's; that case now uses the same name-listing sentence the explicit
style already uses. No new wording — the three strings are unchanged, only how
they compose. Pinned by a mixed-verdict test on both call styles, a
single-unshared test that now asserts the absence wording is absent, and an
admitted-set test for the fully narrowed-away request.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: The observation is right; the suggested filter would not have caught it. Every
row `list_collections` produces has `ownership == "personal"` — the only place
that field becomes `"team"` is the overlay in `_list_visible_collections`, so
filtering the creator's raw listing on it removes nothing (verified: zero rows
dropped on the exact case you describe).

Fixed in the "give each declared knowledge base one verdict" commit by changing
the data source instead: names on the creator's tenant that a team the creator
currently belongs to owns are excluded from what the probe may call the
creator's own, asked through the runner-keyed visibility hook. That answer only
ever moves a name from "held, unshared" to "absent" — a run learns less, never
more — and it lives inside the probe's existing try, so a failed lookup hedges
rather than falling back to the unfiltered listing. One bound worth stating: a
team row hosted on the tenant of someone who has since left that team is out
of the seam's reach and still classifies as personal; the hook contract has no
"which team owns this name" question, and adding one would widen what a run
can learn. Pinned by a test that hosts another team's collection on the
creator's tenant, and a second case that keeps the creator's same-named
personal collection classified as theirs when the team's row lives elsewhere.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed in the same commit: the clause now lists the agent's declaration itself.
One correction to the framing, for the record: `authorised` is the declaration,
so the intersection was always a subset of it and never surfaced an undeclared
team collection — the actual defect was the other direction, dropping a
declared name that happens to be invisible right now, which told the model it
may not ask for something the agent is configured with. Pinned by a test whose
declaration includes a currently-unresolvable name and asserts the clause is
exactly the declaration, plus a second one asserting a team-owned but
undeclared name is still refused when the model names it explicitly.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Both descriptions reworded in the "correct the search tool's collection-default
contract" commit. `collections` now says an empty list searches whatever the
agent is configured with — `allowed_collections` when set, everything
otherwise, and for a team-owned agent that agent's own stored list — and
`allowed_collections` says it is ignored for a team-owned agent, whose stored
declaration is the authority.

tests/core/tools/core/test_document_search_team_declared_names.py
AlexLiu190625: Taken as follows. Added: an admin running a team-owned agent (still resolves
against the governing team, still refuses the creator's unshared collection
despite seeing it platform-wide, and names nothing undeclared), and a
team-owned name the agent never declared arriving as an explicit request
(refused, clause lists the declaration only). The "not allowed" clause now has
its own assertion through the declaration test above. The empty section
headers left behind when this suite was split are removed — six of them, on
inspection. The partial-failure-plus-probe-failure wording is the one gap left
open here: covering it needs a team row hosted off the creator's tenant so the
overlay's own listing does not hit the same broken stub, and it is a coverage
gap rather than a behaviour gap (the path was walked by hand and behaves), so
it is not blocking this change.

AlexLiu190625:
All five inline points are answered in place. Two things about the branch
itself, then the simplification list.

**The diff has moved since your review.** You reviewed this while it was still
stacked on the seam branch, so the diff showed both halves; the seam has since
merged and this branch is rebased onto main, so the diff is now the rule alone.
None of your five points landed on the seam half — every anchor is in the
rule's own code, and GitHub re-anchored the inline threads onto the current
lines. Separately, one commit here — "Map the knowledge-base scope error to its
HTTP status and pin the live-ORM derivation" — was pushed after your review and
has not been seen by anyone yet: it adds the typed scope error's HTTP exit in
the task-creation endpoint (defence in depth beside the connector error's exit;
no request path resolves the team layer today), a behavioural pin on the
live-ORM derivation branch, and docstring wording aligned with what the call
sites actually bind. Worth a look with the rest.

**Simplification list**, in order:

1. The `name not in declared_names` guard — keeping it, as you allowed. It is
   the control that stops a model-supplied name from ever reaching the
   creator-existence probe; on the one caller today it is redundant by
   construction, and it is redundant on every path except the one it exists
   for, which is exactly why it carries a comment saying so. The second caller
   the design anticipated is deferred, not abandoned.
2. The two probe-failure tests are merged into one parametrized test with the
   union of assertions.
3. `partial_failure_notes` is initialised once above the branch chain.
4. `_Unresolved` stays a class. Replacing it with a bare `object()` sentinel
   fails this repository's own mypy check: `is` against an `object()` does not
   narrow the remaining branch, so appending the verdict to
   `List[CollectionInfo]` is rejected as `arg-type`. The class-based sentinel
   is what makes the `isinstance` narrowing work.
5. The always-empty variable in the probe-failure branch no longer exists —
   the verdict-derived rendering above removed it.
6. The alias keeps its name and loses the redundant annotation.

Net on production code from the list is a few lines; the larger saving is in
the merged test.

Changes in this round are three commits: the verdict-derived terminal summary
plus the probe's narrowed data source and the declaration-listing clause with
their pins; the schema wording plus the admin and undeclared-name tests and the
header cleanup; and one extra pin that makes the tenant filter in the probe's
exclusion observable.

src/xagent/core/tools/core/document_search.py
```diff
@@ -337,6 +363,201 @@ async def search_knowledge_base(
...
+        # decision (unchanged), but the report must not claim an absence
+        # that was never established.
+        self.lookup_failed = False
+
+    async def holds(self, name: str) -> bool:
+        if self._agent_creator_user_id is None:
+            return False
+        if self._names is None:
+            try:
+                creator_result = await list_collections(
```
rogercloud, line 549: Minor: the creator probe never checks `ListCollectionsResult.status` — `list_collections` degrades to `status='error'` with an empty list instead of raising (collections.py:998-1006), so during an outage the probe treats the creator as holding nothing, `lookup_failed` stays `False`, and the summary asserts the certain "does not exist" wording, violating this probe's own hedge contract (docstring at 489-493). Fix: check `status` before building `_names` and add a `status='error'` test.

src/xagent/core/tools/core/document_search.py
```diff
@@ -402,11 +695,201 @@ async def _search_knowledge_base_impl(
...
+                    else set(declared_names_for_this_run)
+                )
+                failed_unshared_names = [
+                    name for name in unshared_names if name in reported_scope
+                ]
+                failed_missing_names = [
+                    name for name in missing_names if name in reported_scope
+                ]
+
+                if unresolvable_names:
```
rogercloud, line 883: Minor: this terminal unresolvable branch renders every unresolvable name, while the sibling report sets are scoped by `reported_scope` (812-817): with `explicit_request={C}` and A, B, C all unresolvable, the summary reports A, B and C even though only C was requested — the PR description promises "only the names actually classified missing". Fix: filter through `reported_scope` here (and, for uniformity, in the partial-path note at 883-886).

src/xagent/core/tools/core/document_search.py
```diff
@@ -402,11 +695,201 @@ async def _search_knowledge_base_impl(
...
+                    # outcome for all of them -- a lookup failure is not
+                    # per-name, so this sentence still says nothing about
+                    # which names the creator holds. The other two report
+                    # sets are necessarily empty here: the first failure
+                    # caches the creator's name set as empty and marks the
+                    # probe failed, so every later unresolved name takes
+                    # this same classification, and a name reaches the
+                    # missing set only on a run whose runner is the creator
+                    # -- which never calls the probe at all.
+                    summary = _TERMINAL_UNRESOLVABLE_MESSAGE.format(
```
rogercloud: Minor: this branch drops the admitted clause even when names were admitted before the narrowing — contradicting the guard comment's promise (~798-806) and the PR description that a fully narrowed-away request "still tells the model what it could have asked for". Reachable: declared=[team-owned A, B], explicit=[B], creator lookup raises → A never appears in the summary. Fix: append the admitted clause here, with a test.

src/xagent/core/tools/core/document_search.py
```diff
@@ -337,6 +363,201 @@ async def search_knowledge_base(
...
+    need a "which team owns this name" question the hook contract does not
+    have, and answering it would widen what a run can learn.
+    """
+    from ....web.services.db_runtime import run_db_io_cancellation_safe
+    from ....web.services.knowledge_base_team_scope import (
+        has_knowledge_base_visibility_hook,
+        visible_team_knowledge_bases,
+    )
+
+    if not has_knowledge_base_visibility_hook():
```
rogercloud, line 470: Minor: "no team owns anything here" is false for deployments that installed only the team-keyed hook — `set_knowledge_base_team_hooks` takes independent kwargs, and the rule gate keys on the team-keyed hook only. In that config a team-owned row hosted on the creator's tenant classifies as the creator's personal collection and the sharing-gap sentence misattributes another team's KB, against the "never more" invariant at 453-455. Fix: word the comment as deployment-dependent.

src/xagent/core/tools/core/document_search.py
```diff
@@ -384,7 +603,81 @@ async def _search_knowledge_base_impl(
...
+        # ``_list_visible_collections`` are not the same predicate: they
+        # share ``governing_team_id is not None and
+        # team_knowledge_base_hook_installed()``, but this one adds a third
+        # conjunct (``declared_names_for_this_run``) that the team layer's
+        # gate does not read. That is intentional, not a mismatch to close --
+        # the team layer's job is only "resolve onto the governing team",
+        # which an empty declaration does not change; this gate's job is
+        # "does the declared-name rule apply at all", which an empty
+        # declaration answers no to.
+        declared_name_rule_applies = (
```
rogercloud, line 727: Minor (latent): the gate lacks a `user_id is not None` conjunct — with `user_id=None` plus a governing team, an installed hook and a non-empty declaration, the rule would activate without the team layer (whose overlay is pre-empted by the early return at ~84), misclassifying every declared name via an unauthenticated probe. No production path reaches this today. One-line fix: add the conjunct.

src/xagent/core/tools/core/document_search.py
```diff
@@ -402,11 +695,201 @@ async def _search_knowledge_base_impl(
...
+                    if explicit_request is not None:
+                        summary = (
+                            f"Error: The following collections do not exist: {', '.join(sorted(explicit_request))}. "
+                            f"Available collections: {', '.join(sorted(admitted_names))}"
+                        )
+                    else:
+                        summary = (
+                            f"Error: None of the allowed collections exist. "
+                            f"Allowed: {', '.join(sorted(authorised))}. "
+                            f"Available: {', '.join(sorted(admitted_names))}"
```
rogercloud: Minor: with an empty admitted set, this branch renders a dangling "Available: " (and the explicit-style sibling at 842-843 renders "Available collections: "), while the mixed branch deliberately drops the clause when nothing was admitted (pinned by the admitted-set test). Fix: drop the clause when `admitted_names` is empty in the uniform branches too, or pin the dangling rendering if intentional.

src/xagent/core/tools/core/document_search.py
```diff
@@ -164,7 +183,7 @@ class KnowledgeSearchArgs(BaseModel):
     )
     allowed_collections: Optional[List[str]] = Field(
         default=None,
-        description="Optional list of allowed collection names. Used as default when collections is empty.",
+        description="Optional list of allowed collection names. Used as default when collections is empty. Ignored for an agent owned by a team, whose own stored knowledge base list is the authority.",
```
rogercloud: Minor: "Ignored for an agent owned by a team" overstates the contract — `allowed_collections` is ignored only when the three-way gate holds (governing team + team-keyed hook + non-empty declaration); an empty-declaration team run still honours it (pinned by the empty-declaration tests). Model-facing wording that flips results on a pinned run should be precise. Fix: "Ignored when the agent is governed by a team and has a non-empty stored knowledge base list".

src/xagent/core/tools/core/knowledge_base_scope.py
```diff
@@ -31,12 +31,13 @@ class KnowledgeBaseScopeError(RuntimeError):
...
-    ``code``, ``status_code``, ``details``, or ``safe_message`` off this
-    class -- both run-path handlers re-raise it unchanged rather than
-    unpacking it -- but a future HTTP-facing catch site should find the
-    same attribute name on either error, not two conventions for the same
-    shape.
+    build the HTTP response. This error is unpacked the same way, at the
+    same place: the ``create_task`` handler in ``web/api/chat.py`` answers
+    ``status_code`` with ``safe_message`` as the detail, directly alongside
+    its ``ConnectorRuntimeError`` arm, and logs ``code``/``details`` rather
+    than returning them. Every other handler -- the two run-path ones and
```
rogercloud: Minor: "Every other handler ... re-raises it unchanged so those attributes survive to that single HTTP-facing site" is false two ways — the probe (document_search.py:541-548) swallows the exception rather than re-raising, and the two run-path re-raises (document_search.py:306-307, 1149-1150) surface in run orchestration, never `create_task`. The pre-PR docstring was accurate; this rewrite introduced the claim. Fix: describe what each call site actually does.

src/xagent/core/tools/core/document_search.py
```diff
@@ -337,6 +363,201 @@ async def search_knowledge_base(
...
+    """Memoised lookup of the agent creator's own collection names.
+
+    At most one creator lookup per search call -- one collection listing
+    and one team-ownership question, both memoised together -- regardless
+    of how many declared names need classifying. The lookup result is
+    cached after the first call, including a failure (cached as
+    "nothing"). A failure degrades to "not held" and never
+    raises: a probe that cannot complete must not block the search, and
+    must not be distinguishable from "the creator does not have this
+    collection" (see the module docstring's identity-disclosure limits).
```
rogercloud: Minor: this reference to "the module docstring's identity-disclosure limits" dangles — the module docstring is a single line and no such section exists anywhere (grep-verified); the limits are actually stated right here in the probe class docstring. Fix: point the reference at this docstring or drop it.

src/xagent/core/tools/core/document_search.py
```diff
@@ -337,6 +363,201 @@ async def search_knowledge_base(
...
+
+    A failure is still not distinguishable per name -- it hits every name
+    this run classifies alike -- so the enumeration oracle stays closed;
+    what the caller does with ``lookup_failed`` is choose a report that does
+    not assert an absence, not reveal one.
+
+    This still leaves a costlier existence-only side channel open to
```
rogercloud: Minor: this side channel is described unconditionally, but the reference deployment installs the KB validator (`set_agent_team_hooks`), which blocks the edit at save time (agent_store.py:461-469 → `UnsharedKnowledgeBasesError`). The channel exists only in deployments without that validator. Fix: qualify the description as deployment-dependent.

src/xagent/core/tools/core/document_search.py
```diff
@@ -145,7 +164,7 @@ class KnowledgeSearchArgs(BaseModel):
     query: str = Field(description="The search query or question")
     collections: List[str] = Field(
         default=[],
-        description="Specific knowledge base collection names to search. Empty list uses allowed_collections if set, otherwise searches all collections.",
+        description="Specific knowledge base collection names to search. Empty list searches every knowledge base this agent is configured with: allowed_collections when set, all collections when not, and for an agent owned by a team, that agent's own stored knowledge base list.",
```
rogercloud: Nit: "Empty list searches every knowledge base this agent is configured with" over-promises — the empty-declaration team-governed fallback searches the visible union, including undeclared team knowledge bases (pinned by the empty-declaration tests). Fix: note the fallback in the wording.

src/xagent/core/tools/core/document_search.py
```diff
@@ -337,6 +363,201 @@ async def search_knowledge_base(
...
+                )
+                # Inside the same try on purpose: if the team lookup fails,
+                # which of these rows are the creator's own is unknown, and
+                # the caller must hedge rather than assert the creator
+                # holds them. Falling back to the unfiltered listing would
+                # restore exactly the misclassification this call exists to
+                # remove.
+                team_hosted = await _team_names_hosted_on(self._agent_creator_user_id)
+                self._names = {c.name for c in creator_result.collections} - team_hosted
+            except Exception:
```
rogercloud, line 559: Minor (latent): this broad `except Exception` would swallow `KnowledgeBaseScopeError` if a future typed path ever feeds it — defeating the typed error channel #1349 introduced. Fix: add `except KnowledgeBaseScopeError: raise` before it (matching the codebase's own convention for typed scope errors) or document the intentional carve-out.

src/xagent/web/api/chat.py
```diff
@@ -4211,6 +4212,30 @@ def _get_default_internal_model_ids() -> Dict[str, Optional[str]]:
         raise HTTPException(
             status_code=exc.status_code, detail=exc.safe_message
         ) from exc
+    except KnowledgeBaseScopeError as exc:
```
rogercloud, line 4215: YAGNI: this arm has no producer — nothing in this endpoint resolves the team knowledge-base layer today, and unlike the connector arm it mirrors (whose producer is live: `prepare_connector_runtime_selection_snapshot` → `ConnectorRuntimeError` on this exact path), no code path can raise `KnowledgeBaseScopeError` here; the 97-line test injects the raise artificially. Consider dropping the arm and its test until a real in-request producer exists, or keep it only with an explicit defence-in-depth note acknowledging the asymmetry.

rogercloud (approved):
## Summary

This is the follow-up review of the force-pushed rewrite of PR #1371 (old head `2803b52e` → new head `81283b9c`): the verdict-derived terminal summary and probe data-source change (702e177f), the schema reword (c2d751da), the tenant-filter pin (81283b9c), and the previously unseen commit 678c6013 (HTTP scope-error exit in `create_task`, live-ORM derivation pin, docstring alignment). All five prior findings were re-verified against the update surface: F1–F4 are fixed, F6 is partially addressed with the author's own admission of the remaining coverage gap, and the 13 new confirmed findings are all minor or nit — no blocking issues. CI is green (14 SUCCESS, 8 SKIPPED). (The prior round's two `CHANGES_REQUESTED` review bodies are an identical duplicate posting artifact and are counted once.)

**Blocking: no — recommended event: APPROVE**

## Round 0 approach verdict

**Acceptable-with-reservations.** The direction is correct and the root-cause fix is coherent: the rule keys on per-name verdicts; the search set and both report sets are built from verdicts, never from intersections with the visible union; the gate leaves unhooked deployments byte-for-byte unchanged; the creator probe is memoised, bound to the stored declaration, and hedges on failure. The reservations:

- **R0-2 — informational:** the third core→web local import (`document_search.py:466-467`) deepens core→web coupling, but it is seam-mediated, matches the codebase convention, and inverting it would be the wrong trade; accepted as a design note.
- **R0-5 — informational:** `_list_knowledge_bases_impl` still reports the visible union; unreachable for team-governed agents in production (the rule owns governed runs) and documented in code; accepted.
- **R0-4 — nit:** the rule gate's conjunct composition (`document_search.py:665-669`) duplicates the team layer's own gate conjuncts (~`:110`) without a meta-invariant test; hook detection itself is single-sourced and complementary behavioural pins cover drift, and the divergence is intentional and documented. Non-blocking.
- **R0-6 — minor (latent):** the probe's broad `except Exception` (`document_search.py:541-548`) — detailed in the findings below.
- **R0-1 — dropped:** the five defensive dead-code sites were verified factual, but the guards are sanctioned defence-in-depth (documented and pinned elsewhere); no action.

## Prior findings verification

- **F1 — MAJOR terminal summary self-contradiction — FIXED.** The summary is now verdict-derived with disjoint report sets; the two inherited terminal sentences render byte-for-byte on their own call styles (verified string-for-string) and the composition is pinned by three tests.
- **F2 — MAJOR creator probe ownership filter — FIXED.** Team-hosted rows on the creator's tenant are excluded through the runner-keyed visibility hook inside the probe's `try`; the bound (rows of a team the creator has since left still classify as personal) is documented in the hook contract; the exclusion is pinned by two tests.
- **F3 — MINOR gate clause intersection — FIXED.** The clause now lists the declaration itself; two pin tests assert the exact clause content.
- **F4 — MINOR stale schema contract — FIXED.** Both descriptions were reworded in c2d751da; the residual wording overstatement is tracked as new finding A6 below.
- **F6 — MINOR test gaps — PARTIAL.** Four of the five criteria are now pinned (clause content, admin run, team-owned-undeclared explicit request, explicit-style probe failure via the merged parametrized test). The residual — the partial-path probe-failure wording (probe raises while team knowledge bases are still searched) — remains unpinned, with the author's own admission:

  > "The partial-failure-plus-probe-failure wording is the one gap left open here: covering it needs a team row hosted off the creator's tenant so the overlay's own listing does not hit the same broken stub, and it is a coverage gap rather than a behaviour gap (the path was walked by hand and behaves), so it is not blocking this change."

  Severity stays minor; non-blocking.
- **Simplification items (6) — RESOLVED.** Items 2 (merged parametrized probe-failure test), 3 (single `partial_failure_notes` init), 5 (forced-empty var removed) and 6 (redundant annotation dropped) are done; items 1 (guard) and 4 (`_Unresolved` class) were kept with technically verified justifications — the guard blocks model-supplied names from the probe, and `_Unresolved` is needed for mypy narrowing via `isinstance` — both sanctioned.

## New findings (verified)

### A1 — MINOR (adjusted from a MAJOR candidate) — creator probe never checks `ListCollectionsResult.status` (`document_search.py:530-540`)

The probe reads only `creator_result.collections`; `list_collections` degrades to `status='error'` plus an empty list instead of raising (`collections.py:998-1006`). In an infra outage the probe therefore treats the creator as holding nothing, `lookup_failed` stays `False`, and the summary asserts the certain "does not exist" wording — violating the probe's own hedge contract (`document_search.py:489-493`). Adjusted to minor because it needs an infrastructure failure to trigger, the pre-existing list path already produces the same false assertion in a full outage, and there is no security/oracle impact — the search decision is unchanged. Fix: check `status` before building `_names`, and add a `status='error'` test.

### A2 — MINOR — unresolvable terminal branch ignores `reported_scope` (`document_search.py:819-834`)

The terminal unresolvable branch renders **all** unresolvable names, while the sibling report sets are scoped by `reported_scope` (`:812-817`): with `explicit_request={C}` and A, B, C all unresolvable, the summary reports A, B and C even though only C was requested. The PR description promises "only the names actually classified missing". The partial-path note (`:883-886`) is unfiltered too (uniform design there). Fix: filter through `reported_scope` in the terminal branch (and, for uniformity, the note).

### B3 — MINOR — unresolvable branch drops the admitted clause (`document_search.py:832`)

This branch drops the admitted clause even when names were admitted pre-narrowing — contradicting the guard comment's promise (~`:798-806`) and the PR description that a fully narrowed-away request "still tells the model what it could have asked for". Reachable: declared = [team-owned A, B], explicit = [B], creator lookup raises → A never appears in the summary. Untested. Fix: append the admitted clause here, with a test.

### A3 — MINOR — no-hook comment is false for team-keyed-only deployments (`document_search.py:469-473`)

"no team owns anything here" is wrong when only the team-keyed hook is installed: `set_knowledge_base_team_hooks` takes independent kwargs, and the rule gate keys on the team-keyed hook only. In that configuration a team-owned row hosted on the creator's tenant is classified as the creator's personal collection, so the sharing-gap sentence misattributes another team's knowledge base — against the "never more" invariant (`:453-455`). Search decisions are unaffected; this is a residual of F2's fix for partial-hook configs. Fix: word the comment as deployment-dependent.

### A4 — MINOR (latent) — rule gate lacks a `user_id` conjunct (`document_search.py:665-669`)

With `user_id=None` plus a governing team, an installed team-keyed hook and a non-empty declaration, the rule would activate without the team layer — the overlay is pre-empted by the early return at `:84` — misclassifying every declared name via an unauthenticated probe. No production path reaches this today. One-line fix: add `user_id is not None` to the gate.

### A5 — MINOR — empty admitted set renders dangling availability clauses (`document_search.py:840-850` vs `:871-876`)

With an empty admitted set, the two uniform branches render "Available: " (`:849`) / "Available collections: " (`:843`), while the mixed branch deliberately drops the clause when nothing was admitted (pinned by the admitted-set test). Fix: drop the clause when `admitted_names` is empty in the uniform branches too (or pin the dangling rendering if intentional).

### A6 — MINOR — schema description overstates when `allowed_collections` is ignored (`document_search.py:186`)

"Ignored for an agent owned by a team" is true only when the three-way gate holds (governing team + team-keyed hook + non-empty declaration); an empty-declaration team run still honours the field (pinned by the empty-declaration tests). This is model-facing wording, and the field flips results on a pinned run. Fix: reword to "Ignored when the agent is governed by a team and has a non-empty stored knowledge base list".

### A7 — MINOR — scope-error docstring claims every other handler re-raises (`knowledge_base_scope.py:38-40`)

The rewritten docstring says "Every other handler ... re-raises it unchanged so those attributes survive to that single HTTP-facing site" — false two ways: the probe (`document_search.py:541-548`) **swallows** the exception rather than re-raising, and the two run-path re-raises (`document_search.py:306-307`, `:1149-1150`) surface in run orchestration, never `create_task`. The pre-PR docstring was accurate; this rewrite introduced the claim. Fix: describe what each call site actually does.

### B1 — MINOR — dangling docstring reference (`document_search.py:493`)

"(see the module docstring's identity-disclosure limits)" points at a section that does not exist — the module docstring is a single line (grep-verified). The limits are actually stated in the probe class docstring itself. Fix: point the reference at this docstring or drop it.

### B5 — MINOR — side channel presented unconditionally (`document_search.py:500-512`)

The same-team-member side channel is real only in deployments without the KB validator: the reference deployment installs `set_agent_team_hooks`, which blocks the edit at save time (`agent_store.py:461-469` → `UnsharedKnowledgeBasesError`). Fix: qualify the description as deployment-dependent.

### R0-6 — MINOR (latent) — broad `except Exception` in the probe (`document_search.py:541-548`)

The probe's blanket `except Exception` would swallow `KnowledgeBaseScopeError` if a future typed path ever feeds it — defeating the typed error channel that #1349 introduced. Fix: add `except KnowledgeBaseScopeError: raise` before it (matching the codebase's own convention) or document the intentional carve-out.

### B7 — NIT — `collections` description over-promises (`document_search.py:167`)

"Empty list searches every knowledge base this agent is configured with" is false for the empty-declaration team-governed fallback, which searches the visible union including undeclared team knowledge bases (pinned by the empty-declaration tests). Defensive corner, but the wording over-promises. Fix: note the fallback.

- `yagni: S1` — drop the dead `KnowledgeBaseScopeError` arm in `create_task` (`chat.py:4215-4238`) and its 97-line test until a real in-request producer exists: unlike the connector arm it mirrors — whose producer is live (`prepare_connector_runtime_selection_snapshot` → `ConnectorRuntimeError` on this exact path) — nothing in this endpoint can raise it, and the test injects the raise artificially. (Author's defence-in-depth intent noted; the asymmetry with the live connector arm is the deciding evidence.) `net: -120`
- `yagni: UPD-1` — `reported_scope`'s else-arm and both `failed_*` comprehensions (`document_search.py:807-817`) are no-ops when `explicit_request is None` (membership by construction); filter only in the explicit branch. `net: ~0`
- `shrink: UPD-2` — `test_chat_knowledge_base_scope_http_exit.py:53-76`: the two tests duplicate the same body (identical `_raise` closure, monkeypatch, `create_task` call) and differ only in the final assertions; merge (~20 of 97 lines saved). `net: -20`
- `S2` (guard at `document_search.py:422-433`) — dropped: prior-round suggestion, author's decline technically verified and sanctioned. `net: 0`

Overall `net: ≈ -140 lines`.

**Blocking issues:** none.

All confirmed findings are minor or nit; the only prior-finding residual (F6) is an author-admitted coverage gap at minor severity. The rule's security model — per-name verdicts, the three-way gate, the probe's declaration-bound memoised lookup with failure hedging — verified intact at the new head. Recommended event: **APPROVE** (author: AlexLiu190625).

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed. The probe now treats a listing whose status is not "success" exactly as it treats one that raised: it logs the reported status, caches an empty name set, and sets lookup_failed, so the terminal path hedges instead of asserting an absence. The judgement is `== "success"` rather than `!= "error"`, so any future status value also lands on the hedging side. Pinned by a new test that returns status="error" with an empty collection list and asserts the hedged wording, plus the absence of both certainty sentences. The wording contract for the hedge now reads "raised, or returned a result whose status reports failure".

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed for the terminal branch: it filters through reported_scope, so an explicit request is answered about the names it asked for. New test: three unresolvable declared names, one requested, and the two unrequested names asserted absent from the summary. The narrowed list cannot come out empty — reaching this branch with an explicit request means every requested name was unresolved, since an admitted one would have survived the narrowing and left the search set non-empty.

The partial-path note is left as it is. Its wording and scope are unchanged by this change, and narrowing it is a separate decision from how the terminal summary is composed.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed. The branch now appends "Available collections: {admitted}." on the same condition as the mixed branch — dropped when nothing was admitted — so the promise in the guard comment holds on this branch too. Pinned by a new test built the way you describe: a team-owned collection hosted off the creator's tenant (so the visible listing survives the broken creator listing), an explicit request naming only the other name, and the creator lookup raising. Before the change that test sees no availability clause at all.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed. The comment now says what the early return actually means — the question cannot be asked, so every row on the tenant is treated as the user's own — and names the team-keyed-only configuration as the case where that is not exact: teams exist there, and a team row hosted on the creator's tenant is classified as personal, so the sharing-gap sentence can name another team's knowledge base. Search decisions are unaffected either way. The ask stops at the return rather than falling back to the team-keyed hook, which answers only for the governing team and so cannot answer this question.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed: `user_id is not None` added as the first conjunct, with a comment bullet for why. Pinned by a test that asserts both halves — the rule stays off, and the creator listing is never requested — since the sharp consequence is the probe running on nobody's behalf, not the wording.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed in both uniform branches: the clause is dropped when the admitted set is empty, as the mixed branch already did. The rest of both sentences is untouched, "Allowed: …" included, and a non-empty admitted set renders exactly the bytes it renders today.

Worth recording why this is not a change to inherited behaviour: before this feature the clause's content was the visible union, which the empty-visible-set early return keeps non-empty, so no run could render it empty. This feature replaced that source with the verdict-admitted set, which it can empty — both through the early return it skips and through a declaration that resolves entirely to the runner's own same-named copies. Pinned by a new byte-exact test on both call styles; the existing generic-error test now asserts the clause is absent instead of splitting on it.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed, with your wording: "Ignored when the agent is owned by a team and its stored knowledge base list is non-empty, which is then the authority." The empty-declaration team run that honours the field stays pinned by the existing tests.

src/xagent/core/tools/core/knowledge_base_scope.py
AlexLiu190625: Fixed. The docstring now describes each site as it behaves: the task-creation handler unpacks the status and the safe message; the run-path and tool-build handlers re-raise unchanged, each to its own caller, rather than to that site; and the creator-collection probe is called out as the one deliberate exception, which swallows the type because its answer chooses how a report is worded and never what is searched or by whom. The test module's docstring carried the same claim and is corrected with it.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed — the dangling reference is gone. The limits are stated in the two paragraphs directly below, which is where they actually live, and the sentence now points there.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed. The paragraph now opens on the condition it needs — a deployment that does not install the save-time knowledge-base validator — and closes by naming what the reference deployment does instead: an update whose declaration names a knowledge base the team does not own is rejected at save time, so the edit this channel depends on never lands. With no validator installed that call returns an empty list and the edit is accepted, which is the configuration the paragraph describes.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Fixed: "Empty list searches the agent's default set: its own stored knowledge base list when the agent is owned by a team and that list is non-empty, otherwise allowed_collections when set, and every collection visible to this run when not." That covers the empty-declaration fallback without promising the search is confined to declared names.

src/xagent/core/tools/core/document_search.py
AlexLiu190625: Kept blanket, with the carve-out documented at the handler.

Checked first: nothing inside that try can produce the typed error today. The listing reports failure through its status rather than raising, and the runner-keyed hook is called directly, without the wrapper that mints the type — the only site that raises it.

And if a typed producer is ever added there, swallowing stays the right answer for this one site. The probe's answer chooses how a failure is worded and never what is searched or by whom, and its documented contract is to degrade to "not held" rather than abort a search that can still succeed for the agent's other knowledge bases. Re-raising would convert a hedged summary into a failed request, which is a change to that contract rather than a tightening of this handler. The comment now states all of it, including that every other handler of the type re-raises and this one must not.

src/xagent/web/api/chat.py
AlexLiu190625: Taking the second option in your comment — keeping the arm with an explicit note acknowledging the asymmetry.

The comment now records the asymmetry as fact: the connector arm's producer is live on this path, this one has none. And why the arm still earns its lines: what the two arms share is the failure-path contract rather than a producer. Both errors carry their own status and a caller-safe message, and the handler below turns anything it does not name into a 500 built from str(exc), so the first in-request producer — the run-path resolution moving earlier, or a save-time validation added here — would answer 500 with a raw exception string, silently. The test's injected raise is what pins that classification, which is why it stays; its two cases, which differed only in their final assertions, are now one test.

AlexLiu190625:
All 13 findings addressed; two of the three simplification items taken.

Terminal summaries. The unresolvable branch now scopes its name list the way its sibling report sets do and appends the admitted-collection clause on the same condition as the mixed branch (the unresolvable branch). The two uniform branches drop that clause when nothing was admitted instead of rendering an empty list after the label (the empty-admitted case); with a non-empty admitted set both sentences are byte-for-byte what they were. Five new tests pin the four new outcomes and the gate change; one existing test that split on the clause now asserts its absence.

Probe. A listing that reports an error status is treated as a failed lookup, so the summary hedges instead of asserting an absence no lookup established (the probe's status check). The blanket except keeps its scope with the carve-out and its reason written at the handler: nothing inside that try produces the typed error today, and the probe's answer feeds a report rather than a search or an authorization decision, so swallowing there must not abort a search that can still succeed (the probe's blanket except).

Gate. The declared-name rule now requires an authenticated runner (the authenticated-runner conjunct), pinned by a test that asserts the creator listing is never requested on such a call.

Wording. The scope error's docstring, the probe's no-hook comment, the side-channel paragraph, the two model-facing field descriptions and the dangling docstring reference are corrected to what the code does (the seven wording points).

Simplifications. The two scope-exit tests are merged (the two exit tests). The task-creation arm is kept with the defence-in-depth note your comment offers as the alternative (the task-creation arm). The reported_scope else-arm and the two comprehensions are left alone (the reported_scope no-ops): they are no-ops as you say, but the terminal changes above already rewrite this block, and the filtering is what makes the three report sets read as one rule rather than three special cases.

The residual on the partial path plus probe failure is unchanged in scope from the last round.

AlexLiu190625:
One follow-up commit after the round above, from an independent re-check of the
branch: three comments still described the creator probe's failure only as
"the lookup raised", while the probe now also treats a listing whose status
reports failure as a failed lookup — each names both forms now, so the wording
matches the branch that acts on them. The two tests that host a team-owned
collection off the creator's tenant share one setup helper instead of twenty
duplicated lines. No behaviour change; still 90 tests across the touched files.

PR #1373, opened 2026-08-14
fix(steps): close planning step on dag_execute_end failure

AlexLiu190625:
## Why

When planning fails, the DAG pattern emits `dag_execute_end` with
`data['status'] == 'failed'`, but the public step projector dropped the whole
`dag_execute_*` family — so the planning step stayed `running` forever on
`GET /v1/chat/tasks/{task_id}/steps`.

## What

The projector now consumes two events from that family. `dag_execute_end`
closes the open planning step as `failed` only when `data['status']` is the
literal `"failed"`; `interrupted`, `waiting_for_user`, `completed`, a missing
key, or `None` all leave it running — the rule stays: a step's status changes
only on an explicit terminal signal, and no signal leaves it running.
`dag_execute_start` clears a stale open-planning key from a prior
`DAGPattern.run()` round, so a later round's failure cannot reach back and
close an unrelated stranded step. The close attaches no extra data —
`dag_execute_end`'s `data['result']` carries the full DAG output and must not
reach the public surface, so the step's public `data` stays exactly
`{"phase": "planning"}`.

A plan-generation exception that escapes `DAGPattern.run()` before
`dag_execute_end` is emitted still leaves the planning step running: that path
emits no end event, matching the documented rule that a step without an end
event stays `running`.

## Verification

A parametrized test over six `data['status']` cases (only `failed` closes;
`interrupted`, `waiting_for_user`, `completed`, a missing key and `None` all
leave the step running); an orphan-end test pinning that a failure with no open
planning step is a no-op; a stale-key test proving a prior round's open step is
not closed by a later round's failure; a leak gate asserting the closed step's
public `data` keys equal exactly `{"phase"}` against an event carrying a full
`result`; and a planning-failure sequence added to the projector/batch
equivalence set, pinning the incremental and batch paths to the same output.
Full `tests/web/api/v1` suite green (364 passed).

### DIFF
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -86,19 +86,28 @@ Pairing rule:
         entry. ``dag_execute_end`` instead reaches into whichever
         ``dag_execution`` planning key is currently open and closes it
         as ``"failed"`` -- but only when ``data['status']`` is the
-        literal string ``"failed"``. ``"interrupted"``,
-        ``"waiting_for_user"``, ``"completed"``, a missing key, and
-        ``None`` all leave the planning step running instead, the
-        opposite fallback direction from ``dag_step_*`` above (which
-        cannot be reused here for exactly that reason). A plan-
+        literal string ``"failed"``. This branch only has a key left
+        to close while plan generation itself is still open: the
+        ``dag_execution`` ``executing``-phase event already clears
+        the key the moment plan generation finishes, so an
+        ``"interrupted"`` or ``"waiting_for_user"`` round reported
+        from that point on (during step execution) can't reach back
+        here. Only a round that is interrupted (or reports
+        ``"waiting_for_user"``, ``"completed"``, a missing key, or
+        ``None``) *during plan generation* leaves the planning step
+        running, the opposite fallback direction from ``dag_step_*``
+        above (which cannot be reused here for exactly that reason). A plan-
         generation exception that escapes ``DAGPattern.run()``
         before it ever calls ``on_pattern_end`` emits no
         ``dag_execute_end`` at all, so the planning step is left
         running indefinitely too -- out of scope for this module.
         ``dag_execute_start`` clears a stale open planning key left by
-        a prior ``DAGPattern.run()`` invocation that never reached a terminal
-        event, so a later round's failure can't reach back and close
-        an unrelated stranded step.
+        a prior round that ended with any non-``"failed"`` status
+        (typically ``"interrupted"`` -- that round did reach a
+        terminal ``dag_execute_end``, it just wasn't the literal
+        ``"failed"`` this module closes on), or whose terminal event
+        was itself dropped, so a later round's failure can't reach
+        back and close an unrelated stranded step.
 
     Orphan ends (end with no matching start) are dropped -- they
     represent malformed data and the SDK contract is "every step has
@@ -148,8 +157,11 @@ class PublicStepProjector:
     """Incremental folding state machine: trace events -> public steps.
 
     Holds all the folding state a fold needs -- the pending
-    (start-seen, end-not-yet-seen) pairing table and the ``dag_plan_*``
-    counter/open-key used to disambiguate replan. The batch driver
+    (start-seen, end-not-yet-seen) pairing table, the ``dag_plan_*``
+    counter/open-key used to disambiguate replan, and the
+    ``dag_execution`` counter/open-key (``_dag_execution_counter`` /
+    ``_open_dag_execution_key``) that tracks the same planning phase
+    from its second, independent event source. The batch driver
     (``map_trace_events_to_public_steps``) keeps none of its own: it
     builds one instance and reads back the result. Holding this state
     on an instance is what lets a caller feed a live event stream one
@@ -484,8 +496,11 @@ class PublicStepProjector:
         # the module docstring's "Pairing rule" section for the full
         # rationale. =====
         if event_type == "dag_execute_start":
-            # Drop a stale open key from a prior DAGPattern.run() round that
-            # never reached a terminal event; the stale pending step
+            # Drop a stale open key left by a prior round that ended with
+            # any non-"failed" status (typically "interrupted" -- that
+            # round did reach a terminal dag_execute_end, it just wasn't
+            # the literal "failed" this module closes on), or whose
+            # terminal event was itself dropped. The stale pending step
             # itself is left as-is (still "running").
             self._open_dag_execution_key = None
             return []
--- a/src/xagent/web/api/v1/_step_mapping.py
+++ b/src/xagent/web/api/v1/_step_mapping.py
@@ -19,10 +19,11 @@ Background:
                               meaningfully different from a flat tool)
       - ``message``          (one user or assistant message)
 
-    Everything else (llm_call_*, memory_*, dag_execute_*, react_task_*,
-    react_step_*, visualization_update, task_completion, trace_error,
+    Everything else (llm_call_*, memory_*, react_task_*, react_step_*,
+    visualization_update, task_completion, trace_error,
     action_*_compact) is **intentionally not exposed** so the SDK
-    surface can evolve without breaking clients.
+    surface can evolve without breaking clients. ``dag_execute_*`` is
+    a partial exception -- see the "Pairing rule" section below.
 
 Pure-function design:
 
@@ -80,6 +81,24 @@ Pairing rule:
         when both appear in the same event stream.
       - ``skill_select_*`` events pair on ``task_id`` (single skill
         selection phase per task).
+      - ``dag_execute_*`` events carry the whole DAGPattern's lifecycle,
+        not a single step's, so they never open their own pending
+        entry. ``dag_execute_end`` instead reaches into whichever
+        ``dag_execution`` planning key is currently open and closes it
+        as ``"failed"`` -- but only when ``data['status']`` is the
+        literal string ``"failed"``. ``"interrupted"``,
+        ``"waiting_for_user"``, ``"completed"``, a missing key, and
+        ``None`` all leave the planning step running instead, the
+        opposite fallback direction from ``dag_step_*`` above (which
+        cannot be reused here for exactly that reason). A plan-
+        generation exception that escapes ``DAGPattern.run()``
+        before it ever calls ``on_pattern_end`` emits no
+        ``dag_execute_end`` at all, so the planning step is left
+        running indefinitely too -- out of scope for this module.
+        ``dag_execute_start`` clears a stale open planning key left by
+        a prior ``DAGPattern.run()`` invocation that never reached a terminal
+        event, so a later round's failure can't reach back and close
+        an unrelated stranded step.
 
     Orphan ends (end with no matching start) are dropped -- they
     represent malformed data and the SDK contract is "every step has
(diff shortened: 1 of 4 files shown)


src/xagent/web/api/v1/_step_mapping.py
```diff
@@ -460,10 +479,42 @@ def feed(self, event: Any) -> List[Dict[str, Any]]:
...
+            self._open_dag_execution_key = None
+            return []
+        if event_type == "dag_execute_end":
+            if self._open_dag_execution_key is None:
+                return []  # No open planning step to close.
+            # Deliberately not _terminal_status_from_event: that
+            # helper's fallback direction (anything but "failed"
+            # becomes "completed") is the opposite of what this branch
+            # needs -- only the literal "failed" may close the step.
+            if _data_get(event, "status") != "failed":
```
rogercloud, line 514: **F1 (minor): non-`failed` ends keep the open key.** The docstring and inline comments attribute stale keys to runs that "never reached a terminal event", but the real dominant source is an interrupted round: `LLMCallInterrupted` during plan generation returns an interrupted result (dag.py:498-505), `on_pattern_end` emits `dag_execute_end(status='interrupted')` (dag.py:426, runtime.py:695-700) with the key still open, and this branch deliberately preserves it. Clearing the key on non-`failed` ends is behavior-equivalent today and safer if the tracer drops the next round's `start`+planning events (a later `end(failed)` could otherwise close the previous round's stranded step). Suggestion: clear the key on non-`failed` ends too.

rogercloud, line 514: **F3 (minor): success-side permanent `running`.** The only `completed` closer of the planning step is the `executing` event (465-478; emitted after successful generate+validate, dag.py:1789-1799), and tracer emission is best-effort (exceptions silently swallowed, runtime.py:1335-1366). If `executing` is lost while the pattern succeeds, `dag_execute_end(status='completed')` arrives with the key still open and this branch no-ops — same symptom as the fixed failure side. Closing on `completed` is safe (a completed end can only mean `executing` was lost; do not extend to `interrupted`/`waiting_for_user`, since `_interrupt_if_requested` (dag.py:1802-1827) can return interrupted before the plan completes). Suggestion: close on `completed` and name this residual gap in the docstring.

tests/web/api/v1/test_steps_mapping.py
```diff
@@ -1511,3 +1524,113 @@ def test_dag_plan_and_dag_execution_coexist_without_id_collision():
...
+
+
+# ===== dag_execute_end: planning failure closes an open planning step =====
+
+
+@pytest.mark.parametrize(
+    "status",
+    ["failed", "interrupted", "waiting_for_user", "completed", "__missing__", None],
+)
+def test_dag_execute_end_only_failed_closes_open_planning_step(status):
```
rogercloud, line 1536: **F2 (minor): production same-round failure sequence untested.** The parameterized test feeds only `[planning, end]`; the stale-key test (1608-1636) omits its own round's planning (negative pin only); the equivalence case (1101-1113) has no `dag_execute_start`. Fix correctness depends on the producer ordering invariant — `on_pattern_start` (dag.py:404) firing before the planning emission inside `_run` (dag.py:1758-1765) — which is unpinned: if the ordering reverses or the start-clear moves, this bug regresses silently with a green suite. Suggestion: add same-round (`start` → planning → `end(failed)`) and two-round sequence cases.

src/xagent/web/schemas/v1.py
```diff
@@ -687,7 +687,13 @@ class PublicStep(BaseModel):
             "state can still contain running steps: interrupted or "
             "cancelled steps never get an end event, so terminal "
             "state is judged from the task's own status, not from "
-            "the absence of running steps."
+            "the absence of running steps. A planning ('thinking' with "
+            "data.phase='planning') step follows the same terminal-"
+            "event rule: it is marked failed only when the DAG "
+            "pattern's own run reports a literal status='failed'; an "
+            "interrupted or waiting_for_user outcome, or a plan-"
```
rogercloud: **F4 (minor): description overstates the running semantics.** This text says an interrupted or `waiting_for_user` outcome "leaves it running instead", but `executing` fires at the end of `_generate_plan` (dag.py:1789-1799), before any step execution (`_execute_ready_steps`, dag.py:808-883) — so by the time a `waiting_for_user`/`interrupted` result arrives, the planning step is already `completed` and the key is `None`, making the new branch a no-op. A `running` step actually occurs only on a plan-generation-time interrupt (`LLMCallInterrupted`, before `executing` fires) or a non-terminal escape. Suggestion: correct the wording.

rogercloud (approved):
## Summary

This PR fixes a permanently-`running` planning step in the public steps projection (`GET /v1/chat/tasks/{task_id}/steps`). Previously the projector dropped the whole `dag_execute_*` event family, so a planning step opened by `dag_execution` (phase=planning/replanning) was closed only by the `executing` phase event and stayed `running` forever when plan generation failed. The change makes `dag_execute_end` close the open `dag_execution` planning step when `data['status']` is the literal `"failed"` (without leaking the DAG's full `result` into public `data`), and makes `dag_execute_start` clear a stale open key from a prior round; docs (`Pairing rule` docstring, `tasks.py`/`schemas/v1.py` descriptions) and projector tests (equivalence case, parameterized status matrix, anti-leak gate, stale-key clear) cover the new surface across 4 files (+195/-10).

**Blocking: no — recommended event: APPROVE**

## Verdict (Round 0)

Design verdict: **acceptable-with-reservations**. The approach is sound and producer-verified: `PlanValidationError`, generation-time generic exceptions, and `RequiredToolCallError` all funnel through `_fail()` producing a top-level `status='failed'` result, so the closed branch is reachable; the non-`failed` fallback direction is safe because when a `completed` end arrives the planning step has already been closed by `executing`. The projector is constructed per request (batch endpoint is its only consumer), so multi-task interleaving is not a risk. Reservations are all minor and itemized below: two non-failed end states can strand the step with the same user-visible symptom (F1, F3), a producer ordering invariant the fix depends on is unpinned (F2), and one schema description overstates the failure semantics (F4).

Design-level findings (not blocking, in body rather than inline):

- **F5 (minor, design, out of scope — disclosed by the PR).** A plan-generation exception that escapes `DAGPattern.run()` before it ever calls `on_pattern_end` (the generic `except` only calls `on_pattern_error` → `trace_error`; dag.py:428-430) emits no `dag_execute_end` at all, leaving the planning step `running` indefinitely — the same user-visible symptom this PR fixes. The PR discloses this honestly (module docstring, `tasks.py:1178-1193`, schema description), so it is a scope choice, not an omission. Suggested follow-up: mirror the `_fail` + `on_pattern_end` pattern of dag.py:419-427 inside the generic exception path (then re-raise), or have `on_pattern_error` emit a DAG-specific end event. Pre-existing; does not block.
- **F6 (minor, design, theoretical).** `dag_execute_end` closes only the `dag_execution` family; the `dag_plan_*` family (`_step_mapping.py:277-307`) has a single `dag_plan_end` completed-only close path and no failure signal at all (`ws_trace_handlers.py:28-38` has no failed action combination). The coexist test (1490-1526) proves both families can hold two planning steps open simultaneously, so a failed mixed-stream run would strand the `plan:` step permanently — the legacy manifestation of the bug being fixed. Mitigation: `trace_dag_plan_start/end` currently have zero callers (no live producer), so a docstring note on the asymmetry or a pin is sufficient.

## Line-level findings (all minor)

- **F1 (minor, `src/xagent/web/api/v1/_step_mapping.py:499`)** — Non-`failed` ends keep the open key. The module docstring and inline comments attribute stale keys to runs that "never reached a terminal event", but the real dominant source is an interrupted round: `LLMCallInterrupted` during plan generation returns an interrupted result (dag.py:498-505), `on_pattern_end` emits `dag_execute_end(status='interrupted')` (dag.py:426, runtime.py:695-700) with the key still open, and the new branch deliberately leaves it open. Clearing the key on non-`failed` ends is behavior-equivalent today and strictly safer if the tracer drops the next round's `start`+planning events (prevents a later `end(failed)` from closing the previous round's stranded step). Suggestion: clear the key on non-`failed` ends as well. (inline)
- **F3 (minor, `src/xagent/web/api/v1/_step_mapping.py:499`)** — Success-side permanent `running`. The only `completed` closer of the planning step is the `executing` event (465-478; emitted after successful generate+validate, dag.py:1789-1799), and tracer emission is best-effort (exceptions silently swallowed, runtime.py:1335-1366). If `executing` is lost while the pattern succeeds, `dag_execute_end(status='completed')` arrives with the key still open and the new branch no-ops — same symptom as the fixed failure side. Closing on `completed` is safe (a completed end can only mean `executing` was lost; do not extend to `interrupted`/`waiting_for_user`, since `_interrupt_if_requested` (dag.py:1802-1827) can return interrupted before the plan completes). Suggestion: close on `completed` and name this residual gap in the docstring. (inline)
- **F2 (minor, `tests/web/api/v1/test_steps_mapping.py:1536`)** — Production same-round failure sequence untested. The parameterized test (1536-1564) feeds only `[planning, end]`; the stale-key test (1608-1636) omits its own round's planning (negative pin only); the equivalence case (1101-1113) has no `dag_execute_start`. Fix correctness depends on the producer ordering invariant — `on_pattern_start` (dag.py:404) firing before the planning emission inside `_run` (dag.py:1758-1765) — which is unpinned: if the ordering reverses or the start-clear moves, the bug regresses silently with a green suite. Suggestion: add same-round and two-round sequence cases. (inline)
- **F4 (minor, `src/xagent/web/schemas/v1.py:694`)** — Description overstates the running semantics. The new `PublicStep.status` text says an interrupted or `waiting_for_user` outcome "leaves it running instead", but `executing` fires at the end of `_generate_plan` (dag.py:1789-1799), before any step execution (`_execute_ready_steps`, dag.py:808-883) — so by the time a `waiting_for_user`/`interrupted` result arrives the planning step is already `completed` and the key is `None`, making the new branch a no-op. A `running` step actually occurs only on a plan-generation-time interrupt (`LLMCallInterrupted`, before `executing` fires) or a non-terminal escape. Suggestion: correct the wording. (inline)
- **F7 (minor, `src/xagent/web/api/v1/_step_mapping.py:150-157`)** — Class docstring state enumeration is stale: it lists only the pending table and the `dag_plan_*` counter/open key, omitting `_dag_execution_counter`/`_open_dag_execution_key` (208-209, pre-existing omission) and the new cross-family reads/writes in the `dag_execute_start/end` branch (486-512) — this PR widens the gap even though the module docstring's Pairing-rule section documents the new state. Doc-only nit. (body)

**Blocking: no.** All 7 findings are minor (4 inline, 3 body); none affect the core correctness of the literal-`failed` path this PR targets, and the same-symptom residual gaps (F1, F3) are pre-existing degradation surfaces rather than regressions introduced here. Design-level items (F5, F6) are disclosed or theoretical. CI is fully green (unit, slow, e2e, pre-commit, frontend build). Recommended event: **APPROVE**.

src/xagent/web/api/v1/_step_mapping.py
AlexLiu190625: Confirmed on both halves. The docstring wording was wrong -- an interrupted round does reach a terminal event: the interrupt path returns an interrupted result and `on_pattern_end` emits `dag_execute_end` with that status, leaving the key open. Corrected in the module docstring and the inline comment.

On the behavioral suggestion: I reproduced the risk with a probe -- after an interrupted round, dropping a later round's start and planning events lets that round's failure close the previous round's stranded step. Clearing the key on non-failed ends removes that, and I could not construct a sequence where clearing loses a legitimate close. Landing it with the ordering tests rather than widening an approved PR: tracked in #1379, decided jointly with the completed-close suggestion below since both answer "an event was dropped" and they point in opposite directions.

AlexLiu190625: The producer analysis is right: trace emission is best-effort, the executing signal is the only completed-closer of the planning step, and losing it leaves the step running forever. I verified the loss case, and also that with no loss this branch could never fire -- the end event always arrives with the key already cleared.

Holding it rather than folding it in: it widens this PR's contract (only the literal "failed" closes the step) to a second status, and it takes the opposite stance from the key-clearing suggestion on the same dropped-event question -- that one says never resolve a step from an inference about a missing event, this one infers a completed close. The public contract already tells clients a terminal task may contain running steps, so leaving the step open is not a contract violation. The direction is settled in #1379 (key-clearing adopted, inference-close declined).

One note for whoever revisits: this must stay an explicit completed check -- reusing `_terminal_status_from_event` would map interrupted and waiting_for_user to completed as well, exactly the fallback direction this branch must not have.

tests/web/api/v1/test_steps_mapping.py
AlexLiu190625: Agreed, the gap is real. Verified the production order works today -- start, then the planning emission, then a failed end closes the step as failed, and the two-round sequence keeps round one running while failing round two -- so this is a missing regression pin, not a bug. Nothing currently guards the ordering invariant (`on_pattern_start` firing before the planning emission inside plan generation). Both cases are tracked in #1379 and land together with the key-clearing change.

src/xagent/web/schemas/v1.py
AlexLiu190625: Correct, and it is user-facing, so it is fixed in this PR. Verified with the projector directly: a sequence that reaches the executing signal and later ends waiting-for-user yields the planning step as completed, not running -- the executing signal fires at the end of plan generation, and waiting-for-user can only arise during step execution, by which point the key is already cleared. Interrupted is only half true for the same reason: only a plan-generation-time interrupt leaves the step running. The field description is reworded accordingly, and the module docstring carried the same phrasing and is corrected with it.

AlexLiu190625:
Addressed in the commit just pushed; per-thread details on each inline thread. The class docstring's state list now covers the execution-family counter and open key as well.

The two design-level items go to follow-ups: the escape-path gap is documented in four places and its fix lives outside this projector -- #1380 records both fix shapes and the cancellation case neither covers; the dag_plan family's missing failure signal turned out to be unreachable -- the family has no producers at all -- recorded in #1381.

Issue #1368, opened 2026-08-14
Interaction answers are not validated against the question's argument definitions


AlexLiu190625:
## Current behavior

`task_interaction_service.respond()` validates an answer's `values` structurally only:

- `values` must be a `dict` (rejected as `invalid_values` otherwise), and
- the dict must be renderable as strict JSON — `json.dumps(values, allow_nan=False)` — so `datetime`/`set`/`bytes` members and non-finite floats are rejected as `invalid_values` instead of failing at bind time.

Nothing checks the answer against the question it answers. The interaction row's `request_payload` carries the question's argument definitions (the `interactions` list: field names, types, options), and an answer that ignores them — unknown field names, a string where a number was asked for, an option value outside the offered set, missing required fields — is accepted, stored as `response_payload`, and staged into the RESUME command payload for the execution side to consume.

The question side already validates its own payload shape via `parse_v1_request_payload` (pydantic) plus a JSON-serializability probe, so the asymmetry is answer-side only. `respond()`'s docstring states the gap explicitly ("Validating an answer against the interaction's own argument definitions is not implemented here").

## Why it is not a quick patch

Per-field validation needs the stored `request_payload`, which is only available after the interaction row load (validation today runs before any session is opened), and it needs an answer contract defined across the interaction argument types (`text_input`, `select`, and the other variants in `ask_user_tool`). That is a schema decision, not just a check: e.g. whether unknown extra fields are rejected or ignored, and whether a `select` answer must match an offered option by value or by label.

## Expected

A validation step, after the row load, that checks `values` against the row's own argument definitions — presence, type, and option membership — returning the existing `ValidationRejected(invalid_values)` outcome (or a more specific reason if the vocabulary grows one), with per-argument-type tests. Until then the execution side must treat `response_payload` as unvalidated input.

PR #1375, opened 2026-08-15
fix(interactions): correct the clarification payload key and converge the SQLite protocol-version constraint

### LINKED ISSUE #1290: fix(migrations): converge the SQLite CHECK on tasks.interaction_protocol_version before the first writer
## Summary

`20260810_add_task_interaction_protocol_version`'s `upgrade()` adds the `tasks.interaction_protocol_version` column on SQLite but never the `ck_tasks_interaction_protocol_version` CHECK, while `Base.metadata.create_all()` adds both. Two SQLite deployments of the same version can therefore enforce different invariants on this column depending on install history.

## Current behavior

- Fresh install (stamp head, then `create_all`): column **and** CHECK.
- Database carried through the revision chain: column only.
- One `downgrade()` + `upgrade()` cycle on a fresh-install database: the `batch_alter_table` rebuild drops the CHECK together with the column, and `upgrade()` adds only the column back. The CHECK never returns, and an insert the database previously rejected is then accepted. Pinned today by `test_sqlite_create_all_downgrade_upgrade_round_trip_loses_the_check`.
- PostgreSQL is unaffected: both paths carry the CHECK.
- Harmless as of today only because the column has zero writers.

## Why it was deferred

Offline (`--sql`) generation cannot run `batch_alter_table`, and a plain CHECK-add is unsupported on SQLite, so `upgrade()` has no offline path to the constraint. Online, `batch_alter_table` could add it exactly the way `downgrade()` already removes it.

## Acceptance criteria

- [ ] Before the first writer of `tasks.interaction_protocol_version` lands, either teach `upgrade()`'s online SQLite branch to add the CHECK via `batch_alter_table`, or record why writing to a column whose constraint is not uniformly enforced is acceptable.
- [ ] `test_sqlite_check_asymmetry_is_expected` and `test_sqlite_create_all_downgrade_upgrade_round_trip_loses_the_check` assert the converged shape instead of re-pinning the asymmetry.
- [ ] Offline `--sql` upgrade rendering still works on both dialects.

AlexLiu190625:
A set of independent defect fixes and deferred-coverage closures in the interaction request path. They share no code and can be reviewed in any order; each commit stands alone.

## What is fixed

**The clarification payload could not be read back.** The builder wrote the waiting turn's free text under a `question` key; the reader validates against `AskUserQuestionArgs`, whose required field is `message`. Validation failed on every payload the builder produced, and the read surface swallowed that failure and fell back to the legacy transcript with no log, no signal and no counter. The builder now aligns to the reader — not the reverse, since that model is also the `ask_user_question` tool's argument schema. A round-trip test now crosses both modules through a real staged row; there was none before, which is why this stayed invisible.

**The SQLite protocol-version CHECK depended on install history.** A fresh install (create_all) carried `ck_tasks_interaction_protocol_version`; a database that walked the revision chain did not. `upgrade()` now adds it on SQLite by rebuilding the table through `batch_alter_table`, mirroring what `downgrade()` already does to remove it. Offline `--sql` generation still cannot converge — batch mode has no `--sql` rendering on either dialect — and that remaining asymmetry is asserted rather than left uncovered. Closes #1290.

**Three degradation log lines were at error level.** They sit on paths that degrade and let the round proceed. They become warnings. A fourth error-level line in the same module stays, because it is only reachable when a caller mixes two runs' results together; a comment now says so.

## What is documented or guarded

- The two anchor resolvers classify legacy checkpoint types oppositely, each for its own reason. Both docstrings now name the other and state why reconciling them is a separate decision.
- An idempotent replay is matched on identity alone and compares no content. That contract, and the key-derivation obligation it rests on, are now written down, with guards on the marker format and the key's input set.
- The post-conflict re-check's dependency on READ COMMITTED becomes two assertions — the server's effective isolation level on a real PostgreSQL session, and the absence of any `isolation_level` on the engine — instead of a comment.
- Five coverage gaps recorded when the anchor resolver landed: a connectivity test from resolver output into the staging primitive, a structural equivalence check between the two copies of the six row-validity conditions, an `__all__` completeness assertion, a docstring on the two outcomes with no counter, and one shared source-scan helper replacing two verbatim copies.

## Two refactors, no behavior change

**The lease fence predicates are now shared.** Three booleans decide whether a settlement still belongs to the run that produced it; the clarification resolver spelled all three inline and the interaction handoff spelled one of them again. They become three predicates beside the lease type they read. Only the boolean is shared — each caller keeps its own failure consequence, and no message text changes. The two WebSocket finalizers deliberately keep their own form (the ownership condition compiled into a locking SELECT's WHERE clause); that the two forms agree — same fields, same comparison, and neither finalizer comparing attempts — is now asserted structurally instead of by review.

**The repository-wide guards moved to their own home.** Three source-tree guards (the single-row chat-message delete ban, the message-type monotonicity sweep, and the chat-history-service layering rule) lived in the suite for one service helper; they now live in `tests/architecture/`, with the shared parsing fixture and every positive and negative control, and the fast test matrix runs the new directory. The move commit is byte-for-byte pure: no guard's body, docstring, assertion message or fixture table changes, and the collected test ids before and after the move are identical. Review that commit with `git show -M -C`; its reviewable surface is the new module docstring plus import lines.

The CI path lists for the migration workflow also gain the interaction suite's four unlisted direct dependencies, with a note that the lists are hand-maintained direct dependencies, not a transitive closure.

## Verification

- Full suite on SQLite for every touched surface; the three zero-caller static gates are untouched and still green.
- Migration integration on both backends: upgrade, idempotence, incremental, downgrade.
- PostgreSQL-only interaction suites against a disposable database.
- Red/green mutation for every behavioural change and every new guard, including: reverting the payload key turns the round-trip red; weakening the shared ownership predicate turns both its behavioural cell and the structural equivalence cell red; deleting the SQLite constraint branch turns the parity cells red; changing the turn-marker encoding turns the format pin red.


tests/web/services/test_interaction_staging.py
```diff
@@ -3123,3 +3126,69 @@ def test_handoff_never_writes_any_task_column(tmp_path: Path) -> None:
...
+    return [
+        node
+        for node in ast.walk(tree)
+        if isinstance(node, ast.Call)
+        and isinstance(node.func, ast.Name)
+        and node.func.id == "create_engine"
+    ]
+
+
+def test_configure_db_sets_no_isolation_level_on_either_engine() -> None:
```
rogercloud: Minor (test blind spot): this scan covers only `configure_db`'s own AST. The PostgreSQL branch's keywords are spread in via `**get_db_pool_kwargs()` (`database.py:194`, expanded at `config.py:2386-2406`), so an `isolation_level` added there would pass both this static scan and the `current_setting` probe at `test_interaction_staging_postgresql.py:673` — the exact regression this test exists to catch. Suggest extending the scan to `get_db_pool_kwargs` (or building a real engine and reading `current_setting`); at minimum, disclose the blind spot in the docstring.

tests/alembic/test_20260810_add_task_interaction_protocol_version.py
```diff
@@ -179,7 +179,7 @@ def test_offline_branch_does_not_reflect(dialect_name, operation) -> None:
 # ---- T-M-1g: online SQLite upgrade/downgrade ----
 
 
-def test_online_sqlite_upgrade_adds_column_without_check_and_downgrade_removes_it() -> (
+def test_online_sqlite_upgrade_adds_column_and_check_and_downgrade_removes_both() -> (
```
rogercloud, line 182: Minor (test blind spot): the online SQLite test rebuilds only the single-column `_legacy_tasks_table` (:189), and the parity suite's narrow inventory (`test_task_interaction_protocol_version_parity.py:113-130`) compares only columns plus `ck_tasks_*` CHECKs. The real `tasks` table carries 4 FKs and 6 named indexes; a `batch_alter_table` rebuild silently dropping any of them would leave the whole suite green. Suggest diffing `get_foreign_keys` + `get_indexes` across upgrade, or at least rebuilding the real table shape here.

src/xagent/web/services/ops_signals.py
```diff
@@ -80,6 +80,18 @@
 # task_clarification_draft.py registers this when a waiting run's result
 # carries no clarification draft to publish, and clears it the next time a
 # draft is successfully resolved to a publishable payload.
+#
+# What "stays lit" costs, operationally: the clearing point is the
+# publishable path only, so on a deployment where publication is blocked
```
rogercloud: Minor (docs): the latch-cost premise contradicts the code. Register (`task_clarification_draft.py:522`) and clear (`:566`) both live inside `resolve_publishable_clarification`, with no rollout/source gating inside the function — gating happens at the callers — so "on a deployment where publication is blocked nothing ever clears it" doesn't hold as written. Rewrite the comment to match the actual register/clear locations.

tests/web/services/test_task_clarification_draft.py
```diff
@@ -513,6 +563,80 @@ def test_payload_round_trip_diverges_from_the_legacy_reader_for_an_empty_form_as
         engine.dispose()
 
 
+def test_built_payload_round_trips_through_a_staged_row_and_the_v1_reader(
```
rogercloud, line 570: Minor (test blind spot): this round-trip pins a shape with type/field/label all present. `tool_waiting` sources normalized through `_normalize_ask_user_interactions` (`react.py:140-183`) only guarantee `field`, so a payload missing `type`/`label` still hits the same silent ValidationError → legacy fallback that this PR's central fix addresses — reachable through a second shape. Suggest a `tool_waiting`-shaped cell or a docstring note.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -293,22 +298,35 @@ def build_clarification_payload(draft: ClarificationDraft) -> dict[str, Any]:
     protocol version and giving :func:`parse_clarification_payload` an
     explicit branch for the old shape -- never a silent redefinition of what
```
rogercloud, line 299: Minor (docs): this promise — "never a silent redefinition of what v1 means" — is contradicted by the very rename in this commit: `question`→`message` renames a key without bumping the protocol version. Harmless today because the column has zero writers, but the promise text (:292-300) should record the exception; the naming paragraph (:302-311) documents the rationale but doesn't claim it.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -293,22 +298,35 @@ def build_clarification_payload(draft: ClarificationDraft) -> dict[str, Any]:
...
     """
 
     original_question = draft.message
     cleaned_question = _strip_control_characters(original_question)
     stripped_characters = len(original_question) - len(cleaned_question)
 
     if not cleaned_question:
-        logger.error(
+        logger.warning(
             "clarification question was empty after removing control characters",
```
rogercloud, line 330: Minor (test blind spot): of the three error→warning downgrades in this module, only the missing-draft path has a level pin (`test_missing_draft_logs_at_warning_not_error`). The empty-after-filter warning here and the oversize-after-truncation warning at :363 could regress to `error` with the suite still green. Suggest caplog level pins for both.

tests/alembic/test_20260810_add_task_interaction_protocol_version.py
```diff
@@ -443,15 +437,16 @@ def test_sqlite_create_all_downgrade_upgrade_round_trip_loses_the_check() -> Non
...
         ).scalar()
-        assert CONSTRAINT_NAME not in ddl
+        assert CONSTRAINT_NAME in ddl
 
     tasks_after = sa.Table(TABLE, sa.MetaData(), autoload_with=engine)
-    with engine.begin() as connection:
-        connection.execute(
-            sa.insert(tasks_after).values(
-                interaction_protocol_version=99, user_id=1, title="probe"
+    with pytest.raises(sa.exc.IntegrityError):
```
rogercloud: Minor (test blind spot): the post-round-trip probe only asserts `raises(IntegrityError)` and never attributes the failure to the CHECK by name (finding verified against lines 432-440). Today the CHECK is the only plausible thrower (FKs off, NOT NULL satisfied), but if a future rebuild breaks server defaults or another constraint, this test stays green for the wrong reason. Suggest matching `'CHECK constraint failed: ck_tasks_interaction_protocol_version'` — the repo already has an `_assert_constraint_named_in_error` precedent.

src/xagent/web/services/task_lease_service.py
```diff
@@ -91,6 +91,72 @@ class TaskLease:
...
+
+    Both fields are compared, never one: a successor runner can reuse a
+    ``run_id`` and a single runner can move between runs, so either field
+    alone leaves a real ownership change looking like a match.
+    """
+
+    return bool(task.runner_id == lease.runner_id and task.run_id == lease.run_id)
+
+
+def task_row_matches_lease_attempt(task: Task, lease: TaskLease) -> bool:
```
rogercloud, line 139: Minor (naming): `task_row_matches_lease_attempt` fails open — returns True when `lease.attempt_id is None` — so the name "matches" overstates what it proves. Both callers' docstrings actively disclaim the matches reading, and sibling `lease_is_fenced` fails closed on None (the opposite direction). Suggest a rename such as `task_row_does_not_contradict_lease_attempt`, or a tri-state return.

tests/web/services/test_interaction_fence_equivalence.py
```diff
@@ -0,0 +1,295 @@
...
+
+def _parse_function(func: Any) -> ast.FunctionDef:
+    source = textwrap.dedent(inspect.getsource(func))
+    tree = ast.parse(source)
+    (node,) = tree.body
+    assert isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
+    return node
+
+
+def _unfenced_lease_guard(func_node: ast.AST) -> ast.If:
```
rogercloud, line 62: Minor (test design): this AST coupling pins non-contractual details — the local name `task_lease` here, the parameter name `task_id` in `_finalizer_has_primary_key_filter`, and cell 3's ban on any `attempt_id` identifier (including unrelated same-named variables). Renaming a local or extracting a helper turns the test red with no failure localization. An ownership-condition factory is structurally feasible (the SQL conditions could be extracted as SQLAlchemy expressions embedded in `.filter()`, leaving the lock position unchanged), though the pure-Python row-comparison predicates themselves can't be called by the finalizers; at minimum compare attribute triples only.

tests/web/services/test_task_interaction_anchor.py
```diff
@@ -556,6 +562,136 @@ def test_matching_non_empty_execution_identity_resolves_anchor(
...
+    return candidates[0].values
+
+
+def _normalized_dump(node: ast.expr) -> str:
+    rewritten = _NormalizeSelfTaskId().visit(node)
+    ast.fix_missing_locations(rewritten)
+    return ast.dump(rewritten, annotate_fields=False)
+
+
+def test_row_validity_conditions_match_trace_handlers_structurally() -> None:
```
rogercloud, line 672: Minor (test design): whole-tree `ast.dump` equality on two six-way disjunctions, with only `self.task_id`→`task.id` normalized. Renaming a variable, extracting a helper, or reordering commutative operands turns this red, and the failure gives no diff localization. Suggest narrowing to the contractual part (condition content + order) or improving the failure signal.

tests/web/services/test_supersede_static_guards.py
```diff
@@ -1,890 +1,38 @@
...
-    """The sweep itself, against a file shaped like the one the deleted
-    substring prefilter used to skip: it imports the constants and the
-    literal ``"question_superseded"`` appears nowhere in its text. Run in
-    memory over a synthetic triple -- nothing is written into ``src/``.
-    """
-    source = """
-from xagent.web.services.chat_history_service import (
-    QUESTION_MESSAGE_TYPE,
-    SUPERSEDED_MESSAGE_TYPE,
+from tests.architecture.test_architecture_guards import (
```
rogercloud: Minor (architecture): post-move, this feature suite imports private helpers (`_string_constant_bindings`/`_string_values`) from `tests.architecture.test_architecture_guards`, against that module's own docstring that feature facts belong to the feature. Pre-move these helpers lived in the feature file itself. Suggest extracting a shared helper module (pattern: `interaction_static_scan_shared.py`).

src/xagent/migrations/versions/20260810_add_task_interaction_protocol_version.py
```diff
@@ -112,22 +111,54 @@ def upgrade() -> None:
...
+    # test watches it: batch_alter_table rebuilds the table and adds the
+    # constraint by name either way, so a rebuild that redundantly re-adds
+    # an already-present CHECK is not something a test can observe from
+    # outside. Removing this guard would still leave the migration correct,
+    # it would just rebuild the table once for nothing.
     if CONSTRAINT_NAME not in checks:
-        op.create_check_constraint(
-            CONSTRAINT_NAME, TABLE, CONSTRAINT_CONDITION, schema=schema
-        )
+        with op.batch_alter_table(TABLE, schema=schema) as batch_op:
```
rogercloud, line 170: Minor (ops/docs): the SQLite online branch forces a full-table `batch_alter_table` rebuild — O(N) copy, exclusive write — for every existing deployment walking the revision chain, while this column has zero writers. The comment here covers correctness but not scale/cost, unlike the PostgreSQL branch, which explicitly discusses locks on large deployments. The base deferred convergence to the first writer (#1290); this PR reverses that without a cost note, and SQLite is the default backend (`config.py:2335-2343`). Suggest an operational note.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -293,22 +298,35 @@ def build_clarification_payload(draft: ClarificationDraft) -> dict[str, Any]:
...
+    ``ask_user_question`` tool's own argument schema, so it is the fixed
+    side of this pairing: this builder aligns to it, never the reverse. A
+    payload keyed ``question`` fails that validation, and the reader
+    swallows the failure and falls back to the legacy transcript with no
+    log, no signal and no counter -- a shape that is silent by
+    construction, which is why the round-trip test below crosses both
+    modules instead of checking this function against itself.
     """
 
     original_question = draft.message
```
rogercloud, line 324: Minor (naming): the rename is half-done — the persisted key and flag variable are now `message`, but the question vocabulary survives in `_QUESTION_MAX_BYTES` (:134), `_QUESTION_TRUNCATION_SUFFIX` (:138), these locals, and `NotApplicable('empty_question')` (:191). No repo-wide readers of the old keys exist (no functional impact); complete the rename or add a comment explaining the residue.

src/xagent/web/services/task_clarification_draft.py
```diff
@@ -293,22 +298,35 @@ def build_clarification_payload(draft: ClarificationDraft) -> dict[str, Any]:
     protocol version and giving :func:`parse_clarification_payload` an
     explicit branch for the old shape -- never a silent redefinition of what
     v1 means.
+
+    The free-text field is called ``message``, not ``question``, because
+    the persisted row's reader is ``parse_v1_request_payload``
+    (``task_interaction_service.py``), which validates against
+    ``AskUserQuestionArgs`` (``core/tools/adapters/vibe/ask_user_tool.py``)
+    and whose required field is ``message``. That model is also the
```
rogercloud: Minor (docs): "whose required field is `message`" — `AskUserQuestionArgs` has both `message` and `interactions` required (neither has a default). Suggest "required fields are `message` and `interactions`".

tests/web/services/test_interaction_fence_equivalence.py
```diff
@@ -0,0 +1,295 @@
...
+            ):
+                candidates.append(node)
+                break
+    assert len(candidates) == 1, (
+        f"expected exactly one ownership .filter(...) call, found {len(candidates)}"
+    )
+    return candidates[0]
+
+
+def _compare_triple(node: ast.Compare) -> tuple[str, str, str]:
```
rogercloud: Minor (test blind spot): `_compare_triple` records `(left field, Eq, right field)` without the right side's root `Name`. If a finalizer filter drifted to `Task.runner_id == <some other object>.runner_id` — e.g. a self-comparison that is always true — cell 2's set assertions would stay green. Suggest asserting the right side's root Name id.

tests/web/services/test_interaction_handoff_surface.py
```diff
@@ -286,6 +287,56 @@ def test_th6_all_names_resolve_and_are_public() -> None:
...
+# T-H-7: __all__ -- completeness, not just soundness. T-H-6 above proves
+# every name in __all__ resolves and is public; this proves the converse,
+# that every public name this module defines at module level is in
+# __all__. Neither direction implies the other.
+# ---------------------------------------------------------------------------
+
+_NOT_PUBLIC_API = frozenset({"logger"})
+
+
+def _module_level_public_definitions(module: Any) -> set[str]:
```
rogercloud, line 300: Minor (test blind spot): this walks only `tree.body` top level and does not descend into `if`/`try` blocks, unlike the architecture guards' `_module_level_statements`, which explicitly descends. A future `__all__`-relevant definition inside a conditional block would escape the completeness net. Not a live bug today; reuse the descending logic or disclose the limitation.

src/xagent/web/services/task_lease_service.py
```diff
@@ -91,6 +91,72 @@ class TaskLease:
...
+# "no row came back" is what tells them ownership changed, and the row lock
+# they take is scoped by that same condition. Rewriting them to load a row
+# and then call the predicate below would move the lock and change what a
+# late result means, so they keep their own form. That the two forms agree
+# is asserted statically, in
+# tests/web/services/test_interaction_fence_equivalence.py.
+# ---------------------------------------------------------------------------
+
+
+def lease_is_fenced(lease: TaskLease) -> bool:
```
rogercloud, line 117: Simplification: `lease_is_fenced` (:117) and `task_row_matches_lease_owner` (:128) each have exactly one production caller (`task_clarification_draft.py:546/549`); the finalizers don't use them, so their main remaining purpose is being test anchors. Inlining both back into the resolver restores the base shape (De Morgan-equivalent, no behavior change) and deletes ~25 lines plus 2 imports; fence-equivalence cells 1/2 would then resolve the resolver's inlined guards (~10 mechanical lines). net: -25 lines possible.

rogercloud (approved):
## Summary

This PR bundles independent fixes and deferred-coverage closures on the interaction request path: the clarification payload builder now emits `message` (the key the `ask_user_question` reader requires) instead of `question`, so staged payloads validate instead of being silently discarded back to the legacy transcript; the SQLite online `upgrade()` now converges the `ck_tasks_interaction_protocol_version` CHECK via `batch_alter_table` (closing #1290), with the remaining offline `--sql` asymmetry asserted rather than pinned as a bug; three degradation logs are downgraded from error to warning; the lease fence predicates are shared with no behavior change; repo-level static guards move to `tests/architecture/`; and five coverage gaps (cross-module round-trip, structural equivalence, `__all__` completeness, docstrings, a shared source-scan helper) are closed. All 14 commits are independently reviewable, CI is green, and the three acceptance criteria of #1290 are met.

## Design verdict

Sound. The core mechanisms were verified against the code and hold up: the migration rebuild (batch_alter_table + FK-off runner + idempotent guard), the builder-to-reader key alignment, the predicate equivalence (De Morgan, step by step), and the guard semantics have no substantive defect, so there is no critical or major finding. The move commit is byte-for-byte clean and the CI path lists are complete.

One initial finding (F18, NULL-semantics documentation) was dropped during verification: SQLAlchemy compiles `== None` to `IS NULL`, which is NULL-safe and agrees with Python `None == None` in every cell; the "SQL never matches NULL" caveat only applies to hand-written raw SQL, which this codebase's ORM expressions never emit.

All remaining findings are minor and grouped below.

## Findings (all minor)

### Observability

- **F1 — silent fallback stays silent.** `src/xagent/web/services/task_interaction_service.py:1307-1310` — `materialize_compatibility_view` swallows `_PydanticValidationError` and falls back to the legacy transcript with no log, no signal, no counter. This PR fixed the cause (the `question`/`message` key mismatch) but not the silence itself; the same module's `_resolve_read_direction_anchor` registers a degradation on its paths, and this one does not. The except branch predates the PR, but it sits squarely in the region this PR exists to fix. Suggest registering a counter or a `logger.warning`/`debug` in the except branch.

### Test blind spots

- **F2 — isolation-level scan misses the kwargs indirection.** `tests/web/services/test_interaction_staging.py:3151` — the static scan covers only `configure_db`'s own AST; the PostgreSQL branch's keywords are spread in via `**get_db_pool_kwargs()` (`src/xagent/web/models/database.py:194`, expanded at `src/xagent/web/config.py:2386-2406`), so an `isolation_level` added there would pass both this scan and the `current_setting` probe (`tests/web/services/test_interaction_staging_postgresql.py:673`). Extend the scan to `get_db_pool_kwargs` (or build a real engine and read `current_setting`); at minimum, disclose the blind spot in the docstring.

- **F3 — SQLite batch rebuild has no FK/index retention assertions.** `tests/alembic/test_20260810_add_task_interaction_protocol_version.py:182` and `tests/migrations/test_task_interaction_protocol_version_parity.py:113-130` — the online test rebuilds only the single-column `_legacy_tasks_table`, and the parity suite's narrow inventory compares only columns plus `ck_tasks_*` CHECKs. The real `tasks` table carries 4 FKs and 6 named indexes; a rebuild silently dropping any of them would leave the suite green. Diff `get_foreign_keys` + `get_indexes` across upgrade, or at least rebuild the real table shape in the online test.

- **F5 — round-trip pins only one interaction shape.** `tests/web/services/test_task_clarification_draft.py:566` — the cross-module round-trip covers a shape where type/field/label are all present. `tool_waiting` sources normalized through `_normalize_ask_user_interactions` (`src/xagent/core/agent/pattern/react/react.py:140-183`) only guarantee `field`, so payloads missing `type`/`label` still hit the same silent ValidationError → legacy fallback this PR's central fix addresses — reachable through a second shape. Add a `tool_waiting`-shaped cell or a docstring note.

- **F7 — only the missing-draft warning has a level pin.** `src/xagent/web/services/task_clarification_draft.py:321,362` — `test_missing_draft_logs_at_warning_not_error` pins the missing-draft path only; the empty-after-filter and oversize-after-truncation paths could regress to `error` with the suite still green. Pin both with caplog.

- **F8 — IntegrityError probe not attributed to the CHECK.** `tests/alembic/test_20260810_add_task_interaction_protocol_version.py:432-443` — the post-round-trip probe only asserts `raises(IntegrityError)`. Today the CHECK is the only plausible thrower (FKs off, NOT NULL satisfied), but if a future rebuild breaks server defaults or another constraint, the test stays green for the wrong reason. Match `'CHECK constraint failed: ck_tasks_interaction_protocol_version'` — the repo already has an `_assert_constraint_named_in_error` precedent.

- **F16 — the "exactly" promise is unguarded on the marker key inputs.** `tests/core/agent/test_clarification_draft.py:408` — `test_idempotency_key_input_set_is_exactly_the_turn_marker_components` pins `message`/`message_type`/`interactions` but not `source` and `origin_execution_id`, which are real `ClarificationDraft` fields the `_marker_draft` helper hardcodes constant. Folding either into the key would silently change replay identity with the suite green. Add the two assertions or narrow the docstring.

- **F17 — triple comparison doesn't pin the right-side root.** `tests/web/services/test_interaction_fence_equivalence.py:149-157` — `_compare_triple` records `(left field, Eq, right field)` without the right side's root `Name`; a finalizer filter drifting to `Task.runner_id == <some other object>.runner_id` (e.g. a self-comparison that is always true) keeps cell 2's set assertions green. Assert the right side's root Name id.

- **F19 — public-definitions scan doesn't descend into if/try.** `tests/web/services/test_interaction_handoff_surface.py:300-311` — `_module_level_public_definitions` walks only `tree.body` top level, unlike the architecture guards' `_module_level_statements`, which explicitly descends. A future module-level definition inside a conditional block would escape the `__all__` completeness net. Not a live bug today; reuse the descending logic or disclose the limitation.

### Naming & docs

- **F4 — latch-cost comment contradicts the code.** `src/xagent/web/services/ops_signals.py:84-92` — the comment claims a blocked deployment "nothing ever clears it", but register (`src/xagent/web/services/task_clarification_draft.py:522`) and clear (`:566`) both live inside `resolve_publishable_clarification`, and there is no rollout/source gating inside the function (gating is caller-side), so the latch premise doesn't hold as written. Rewrite the comment to match the real register/clear locations.

- **F6 — the rename contradicts the version-evolution promise.** `src/xagent/web/services/task_clarification_draft.py:292-300` (promise at :299) — the `question`→`message` key rename happens without bumping the protocol version, against the promise "never a silent redefinition of what v1 means". Harmless today because there are zero writers, but the promise text should record the exception; the naming paragraph (:302-311) documents the rationale without claiming it.

- **F9 — "matches" is fail-open.** `src/xagent/web/services/task_lease_service.py:139-157` — `task_row_matches_lease_attempt` returns True when `lease.attempt_id is None`; "matches" overstates what it proves. Both callers' docstrings actively disclaim the matches reading, and sibling `lease_is_fenced` fails closed on None — the opposite direction. Rename (e.g. `task_row_does_not_contradict_lease_attempt`) or return tri-state.

- **F10 — AST coupling pins non-contractual details.** `tests/web/services/test_interaction_fence_equivalence.py` (all-new, 295 lines) — `_unfenced_lease_guard` asserts the local name `task_lease`, `_finalizer_has_primary_key_filter` pins the parameter name `task_id`, and cell 3 bans any `attempt_id` identifier, including unrelated same-named variables. Renaming a local or extracting a helper turns the test red with no failure localization. An ownership-condition factory is structurally feasible (the SQL conditions could be extracted as SQLAlchemy expressions embedded in `.filter()`, leaving the lock position unchanged), though the predicates themselves — pure-Python row comparisons — can't be called by the finalizers; at minimum compare attribute triples only.

- **F11 — whole-tree dump equality with no failure localization.** `tests/web/services/test_task_interaction_anchor.py:672` — two six-way disjunctions are compared via full `ast.dump` trees with only `self.task_id`→`task.id` normalized; renaming a variable, extracting a helper, or reordering commutative operands turns red with no diff localization. Narrow to the contractual part (condition content + order) or improve the failure signal.

- **F12 — feature suite imports private architecture helpers.** `tests/web/services/test_supersede_static_guards.py:29-31` — after the move, the feature suite imports `_string_constant_bindings`/`_string_values` from `tests/architecture/test_architecture_guards.py`, against that module's own docstring that feature facts belong to the feature; pre-move they lived in the feature file. Extract a shared helper module (pattern: `tests/web/services/interaction_static_scan_shared.py`).

- **F13 — SQLite rebuild cost is not documented.** `src/xagent/migrations/versions/20260810_add_task_interaction_protocol_version.py:160-162` — the SQLite online branch forces a full-table `batch_alter_table` rebuild (O(N) copy, exclusive write) for every existing deployment walking the revision chain, while the column has zero writers; the comment covers correctness but not scale/cost, unlike the PostgreSQL branch, which explicitly discusses lock behavior on large deployments. The base explicitly deferred convergence to the first writer (#1290); this PR reverses that without a cost note, and SQLite is the default backend (`src/xagent/web/config.py:2335-2343`). Add an operational note.

- **F14 — the rename is half-done.** `src/xagent/web/services/task_clarification_draft.py:316-318` (also :134, :138, :191) — the persisted key and flag variable were renamed to `message`, but `_QUESTION_MAX_BYTES`, `_QUESTION_TRUNCATION_SUFFIX`, `original_question`/`cleaned_question`/`stripped_characters`, and `NotApplicable('empty_question')` keep the question vocabulary. There are no old-key readers anywhere (no functional impact); complete the rename or add a comment.

- **F15 — docstring understates the required fields.** `src/xagent/web/services/task_clarification_draft.py:306` — "whose required field is `message`": `AskUserQuestionArgs` has both `message` and `interactions` required (neither has a default). Say "required fields are `message` and `interactions`".

- L117: delete `lease_is_fenced` + `task_row_matches_lease_owner`. Each has exactly one production caller (`task_clarification_draft.py:546/549`); the finalizers don't use them; their main remaining purpose is being test anchors. Inline both back into the resolver (De Morgan-equivalent to base) and point fence-equivalence cells 1/2 at the resolver's inlined guards (~10 mechanical lines). ~-25 lines, -2 imports.
- L297: shrink `_NOT_PUBLIC_API = frozenset({"logger"})` to `{"logger"}` — it's used once, in a single set difference.

net: -25 lines possible

## Blocking status

Blocking: no. The design is sound, all mechanisms are verified, and every remaining finding is minor (18 confirmed, 1 dropped during verification). Recommended decision: **APPROVE**.

tests/web/services/test_interaction_staging.py
AlexLiu190625: Fixed, both halves. The static scan now also walks `get_db_pool_kwargs`'s own source and asserts no `isolation_level` key there (mutation-checked: injecting one into its return dict turns the scan red), and both docstrings drop the joint-completeness claim in favor of stating each half's real boundary — the scan cannot see through the indirection it now covers explicitly, and the PostgreSQL probe's engine never goes through `configure_db`, so it proves the server default, not this codebase's restraint.

tests/alembic/test_20260810_add_task_interaction_protocol_version.py
AlexLiu190625: Verified against the real table shape rather than the single-column fixture: after a full downgrade→upgrade round trip on a `create_all`-built database, `get_foreign_keys` reports 4 before and 4 after, `get_indexes` 7 before and 7 after (7 named indexes today, for the record). So this is a genuine blind spot but not a live defect. Recording the preservation assertion as a follow-up rather than adding it here — the diff-across-upgrade cell is worth ~25 lines and stands alone. Leaving open as the record.

src/xagent/web/services/ops_signals.py
AlexLiu190625: Fixed. The comment now describes the actual lifecycle: register and clear both live inside `resolve_publishable_clarification` (register on the missing-draft branch, clear on the next successful resolve), no rollout or source gating inside the function, and — since that function has no production caller today — the latch scenario is stated as hypothetical rather than as a standing operational cost.

tests/web/services/test_task_clarification_draft.py
AlexLiu190625: Fixed, and the reproduction confirmed this is the real remaining path to the same silent endpoint: a payload whose interactions carry only `field` (all `_normalize_ask_user_interactions` guarantees) fails the same validation and fell back with nothing logged. Two cells landed: a second round-trip through a real staged row for a fuller shape (select with options), and a cell pinning the incomplete-shape path — the parse failure plus the fallback's new warning log (the fallback branch in the read surface now logs task id and a validation summary before returning the legacy view; the degradation-signal design stays with the read-surface change, and per-field validation remains tracked in #1368).

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed. The promise paragraph now records the exception in place: this rename happened under the zero-writers premise — the table holds no rows, so there was no old shape to keep readable — and any rename after a writer exists requires the version bump the promise describes.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed. Both remaining downgrades now have caplog level pins (empty-after-filter and oversize-after-truncation), each mutation-checked by flipping the site back to `logger.error` and watching exactly that pin go red.

tests/alembic/test_20260810_add_task_interaction_protocol_version.py
AlexLiu190625: Fixed. Both probes (the pre-round-trip one and this one) now pass `match=` naming `ck_tasks_interaction_protocol_version`, checked against the real SQLite error text; pointing the pattern at a wrong constraint name fails the match, so the attribution is live, not decorative.

src/xagent/web/services/task_lease_service.py
AlexLiu190625: Agreed on the reading — the name promises more than the fail-open check proves, and both call sites already disclaim it. Declining the rename in this change on cost grounds: it touches three source files plus tests for a purely nominal gain, and the docstring carries the None-sentinel semantics in full. Recording it for the next change that touches this module. Leaving open as the record.

tests/web/services/test_interaction_fence_equivalence.py
AlexLiu190625: Partially agreed, keeping the current shape. The pinned local and parameter names are real, but each assertion carries a named f-string message pointing at the function and condition it checked, so a rename failure localizes to the exact cell rather than a bare mismatch. The "compare attribute triples only" suggestion is what cell 2 already does — and this round strengthened those triples to record the right side's root object as well (see the sibling thread). The condition factory would indeed keep the lock position, but as you note the Python predicates cannot serve the finalizers, so the factory buys structure without deleting the duplication it targets; not taking it here. Leaving open for your read.

tests/web/services/test_task_interaction_anchor.py
AlexLiu190625: Keeping this one as is — the description doesn't match the test's mechanics. It is not a whole-tree dump comparison: the two sides are lists of per-condition normalized dumps compared element-wise, so a mismatch localizes to the failing condition's index and pytest prints both dumps for that element. Order sensitivity is deliberate and contractual — the module's own docstring requires the ownership condition to run before the partition condition — so a reorder turning the test red is the test doing its job. The suggested narrowing ("condition content + order") is the current behavior.

tests/web/services/test_supersede_static_guards.py
AlexLiu190625: Fixed with one adjustment to the suggested shape. The helpers moved to a shared module (`tests/architecture/string_constant_resolution.py`) with public names, imported by both suites. They could not move verbatim: both closed over a module-global constants table in the guards file, which would have kept a hidden dependency pointing back into the file they were extracted from. They now take the known-constants table as an explicit parameter — the guards file passes its two-entry table, the feature suite passes the single entry it needs. Test counts on both sides are unchanged before/after the move.

src/xagent/migrations/versions/20260810_add_task_interaction_protocol_version.py
AlexLiu190625: Fixed. The SQLite branch now carries the operational note, symmetric with the PostgreSQL branch's lock discussion: the rebuild is a full-table copy under an exclusive write, its cost scales with the table, and the reason it is acceptable to pay on the revision chain today is the same zero-writers fact the base version cited for deferring — with no writer, converging deployments rebuild a small table now rather than a large one later.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Recording rather than completing the rename here, with one hard boundary worth naming: `"empty_question"` is a member of the `NotApplicableReason` literal — a consumed contract vocabulary, not internal naming — so renaming it is a semantics change that needs its consumers enumerated, which is out of scope for this collection. The remaining private names (`_QUESTION_MAX_BYTES`, the truncation suffix, the locals) are rename-safe but pure churn; the naming paragraph the commit added already explains why the free-text field is called `message`, and the residue is now tracked. Leaving open as the record.

src/xagent/web/services/task_clarification_draft.py
AlexLiu190625: Fixed — "required fields are `message` and `interactions`", matching the model (neither has a default).

tests/web/services/test_interaction_fence_equivalence.py
AlexLiu190625: Fixed. `_compare_triple` now records the right side's root name, the known legal difference (`task_lease` on the finalizer side vs `lease` on the predicate side) joined the normalization table with its reason, and the mutation you described — pointing the predicate's right side at the wrong object — was run for real and turns cell 2 red with the mismatched root visible in the failure.

tests/web/services/test_interaction_handoff_surface.py
AlexLiu190625: Confirmed and recorded rather than changed: the scanned module currently has zero module-level `if`/`try` blocks, so the non-descending walk has nothing to miss today (measured, not assumed). The completeness net's docstring gains nothing from borrowing the descending walker until a conditional definition exists to find; recording the limitation instead, with the architecture guards' `_module_level_statements` named as the shape to adopt when it does. Leaving open as the record.

src/xagent/web/services/task_lease_service.py
AlexLiu190625: Declining the inline-back. The count is off by one: `task_row_matches_lease_attempt` has two production callers (the resolver and the interaction handoff's re-check), so the extraction can't collapse back to the base shape without leaving a one-predicate module behind — deleting two of three siblings while their family docstring explains the trio would trade 25 lines for an asymmetry that invites the next reader to inline the third. The predicates also just gained a behavioral pin on the run half of the ownership comparison, which the fence-equivalence cells alone don't carry. Leaving open for your read.

AlexLiu190625:
Round closed at head d548adbae. Fixed in three commits: the four comments this change introduced or dated, the guard gaps this round's own tests left open (log-level pins, constraint attribution, the equivalence triples' right-side root, the key's remaining input-surface cells, the isolation scan's real boundary plus its extension through the pool-kwargs indirection, and one shared constant-resolution helper replacing the cross-package private import), and the second unreadable-payload path — the read-surface fallback now logs before returning the legacy view, with the degradation-signal design left to the read-surface change and per-field validation tracked in #1368. Declined with reasons on their threads: the predicate inline-back (the third sibling has two callers), the anchor-condition comparison rewrite (it already compares per-condition, order is contractual), and the fail-open rename (recorded). Recorded for follow-up: the FK/index preservation cell (measured intact today: 4 FKs, 7 indexes through a full round trip) and the descending completeness walk (scanned module has zero conditional module-level definitions today).

PR #1376, opened 2026-08-15
feat(interactions): classify why the answer fence missed

AlexLiu190625:
## Summary

Classifies why the answer fence missed. The merged answer-side entry point (#1354) collapses every zero-row fence UPDATE onto `RespondOutcomeUnknown` by design, with the reread row state visible only in a log line; this change delivers the classification that build deferred: the reread now resolves to a typed outcome — `Conflict(already_answered)`, `Unauthorized(not_task_principal)` for an ownership miss, and the stale family (`run_ended`, `run_superseded`, `expired`, `answered_via_chat`, `foreign_run`) — instead of the catch-all. A same-key command whose payload does not match resolves to `Conflict(idempotency_key_reused)`, the same verdict the idempotency pre-read gives that situation, so the pre-read and the fence-miss classification agree.

## What's here

- The reread-and-classify block in the fence-miss branch, ordered so the most specific verdicts resolve first, with two structural invariants that raise rather than classify (an unrecognized terminal reason or row status is a schema fact, not an outcome).
- One deliberate ordering pin: when a run has ended **and** the row hangs on the old run — the most common overlap in practice, since status and run id move together — the verdict is `run_ended`, because it describes the task's own present state and does not disclose run history to a caller that was merely late. A test pins the guard order; swapping the two checks turns exactly that test red.
- The fence-miss log line stays: the classification hands the caller a verdict, while the log carries row fields (`active_slot`, `terminal_reason`, `run_id`, `responder_identity`) that no outcome variant reports — operator information, not a duplicate of the outcome.
- The conflict counter now increments on a confirmed `already_answered` — the previous build deliberately did not count an unclassified miss as a conflict, and the test that pinned "counter untouched" is translated to pin the confirmed-conflict semantics instead.
- Matrix and vocabulary arithmetic rebuilt over the merged build's cells; the outcome vocabulary meta-test derives the new pairs from the `Literal` types as before.

## Verification

Full SQLite suite for the touched surfaces (122 in the service file after this round's two race tests), PostgreSQL-only suite against a disposable database, and red/green mutation for every restored branch: reverting the ownership classification to the catch-all turns the ownership tests red, deleting the fence-miss log turns its pin red, swapping the ordered guards turns the ordering pin red while everything else stays green, deleting the reread's replay recognition turns the matching-payload race test red, and reverting the mismatched-payload verdict to `already_answered` turns the mismatched race test red.

### DIFF
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -2327,20 +2327,38 @@ def respond(
                     )
                     .first()
                 )
-                if fresh_command is not None and _matches_existing(
-                    fresh_command,
-                    actor_user_id=actor_user_id,
-                    kind=TaskCommandKind.RESUME,
-                    payload=command_payload,
-                ):
-                    return RespondReplayed(
-                        receipt=_respond_receipt(
-                            interaction=reread,
-                            task=task,
-                            command_db_id=int(fresh_command.id),
-                            idempotency_key=normalized_key,
+                if fresh_command is not None:
+                    # A command row under this exact idempotency key exists
+                    # now, even though step 5's pre-read (run before the
+                    # fence attempt) found none: the row that answered this
+                    # interaction was staged concurrently, in the same
+                    # window this call's own fence attempt lost. Whether
+                    # that command is this call's own request replaying, or
+                    # a different payload racing under the same reused key,
+                    # is exactly the question step 5's own two-way split
+                    # answers, so this reread applies the same test: a
+                    # payload match is a replay, a mismatch is
+                    # ``idempotency_key_reused`` (see step 5 above).
+                    if _matches_existing(
+                        fresh_command,
+                        actor_user_id=actor_user_id,
+                        kind=TaskCommandKind.RESUME,
+                        payload=command_payload,
+                    ):
+                        return RespondReplayed(
+                            receipt=_respond_receipt(
+                                interaction=reread,
+                                task=task,
+                                command_db_id=int(fresh_command.id),
+                                idempotency_key=normalized_key,
+                            )
                         )
-                    )
+                    increment_counter(COUNTER_LIFECYCLE_RESPONSE_CONFLICT)
+                    return RespondConflict(reason="idempotency_key_reused")
+                # No command row exists under this key at all: some other,
+                # unrelated write answered this row (not one racing this
+                # call's own idempotency key), so there is nothing to
+                # replay or compare payloads against.
                 increment_counter(COUNTER_LIFECYCLE_RESPONSE_CONFLICT)
                 return RespondConflict(reason="already_answered")
             if reread.status == "terminated":
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -2015,11 +2015,12 @@ def respond(
        requires the owner in person, so an admin acting on another user's
        task and a guest whose ``principal.user_id`` is not the owner's
        both reach it on PostgreSQL too, alongside SQLite's own concurrent
-       ownership change (see that predicate's own docstring). The reread
-       row's state is logged before the classification runs, because no
-       outcome variant carries those columns. ``rowcount > 1`` is a schema
-       invariant violation (``uq_task_interaction_active_slot``) and
-       raises.
+       ownership change and, for a guest, SQLite's own concurrent write to
+       ``agent_config["guest_id"]`` between step 2's read and the fence
+       (see that predicate's own docstring). The reread row's state is
+       logged before the classification runs, because no outcome variant
+       carries those columns. ``rowcount > 1`` is a schema invariant
+       violation (``uq_task_interaction_active_slot``) and raises.
     7. The Task CAS via ``apply_task_control_transition``, called with no
        ``expected_run_id`` / ``expected_state_version`` -- this function
        takes no caller-supplied optimistic-concurrency token, so neither of
@@ -2279,7 +2280,14 @@ def respond(
             # changed since step 4's read); without expiring first, the
             # ORM would hand this query's result back through the same
             # already-loaded, now-stale Python object instead of the fresh
-            # row this reread exists to see.
+            # row this reread exists to see. This has to expire the whole
+            # session, not just ``ir`` -- the classification below also
+            # reads ``task.status`` and ``task.run_id`` (the "active"
+            # branch further down), and those need refreshing too: without
+            # it, ``task.status`` would still read back step 2's own
+            # WAITING_FOR_USER value even after some other writer moved the
+            # task on, and a genuine ``run_ended`` miss would misclassify
+            # as an ownership miss (``not_task_principal``) instead.
             db.expire_all()
             reread = get(db, task_id=task_id, interaction_id=interaction_id)
             if reread is None:
@@ -2291,7 +2299,14 @@ def respond(
             # what it found costs no extra statement. The classification
             # below tells the *caller* which miss this was; this line tells
             # an *operator* what the row actually looked like, and the two
-            # are not the same information. No ``RespondOutcome`` variant
+            # are not the same information. The reason it hands back
+            # describes the row and task as this reread found them, not
+            # necessarily the exact condition that made the fence UPDATE's
+            # own WHERE clause fail to match: neither backend holds a lock
+            # across the gap between that UPDATE and this SELECT, so in
+            # principle the row could change again in between and the
+            # label would name whatever this reread actually saw, not the
+            # original miss. No ``RespondOutcome`` variant
             # carries ``active_slot``, ``terminal_reason``, ``run_id`` or
             # ``responder_identity``: ``Stale(run_superseded)`` names the
             # reason without naming the run, and ``Conflict
@@ -2362,13 +2377,22 @@ def respond(
                 increment_counter(COUNTER_LIFECYCLE_RESPONSE_CONFLICT)
                 return RespondConflict(reason="already_answered")
             if reread.status == "terminated":
-                terminal_reason_map: dict[str, RespondStaleReason] = {
+                # No ``or ""`` fallback needed on the lookup key --
+                # ``ck_task_interaction_requests_terminal_pairs_status`` is
+                # a biconditional (``status = 'terminated'`` iff
+                # ``terminal_reason IS NOT NULL``), so a row that reaches
+                # this branch is guaranteed by the database itself to carry
+                # a non-``None`` ``terminal_reason``. The ``cast`` below is
+                # only for mypy's benefit (the mapped column's static type
+                # is ``str | None``); it asserts that guarantee, it does
+                # not substitute a runtime value the way ``or ""`` did.
+                terminal_stale_reasons: dict[str, RespondStaleReason] = {
                     "deadline_elapsed": "expired",
                     "run_superseded": "run_superseded",
                     "answered_via_legacy_resume": "answered_via_chat",
                 }
-                mapped_reason = terminal_reason_map.get(
-                    str(reread.terminal_reason or "")
+                mapped_reason = terminal_stale_reasons.get(
+                    cast(str, reread.terminal_reason)
                 )
                 if mapped_reason is None:
                     raise RuntimeError(
@@ -2385,16 +2409,23 @@ def respond(
                 # left that can have failed is
                 # ``_answer_fence_task_predicate``'s ownership conjunction
                 # -- which requires ``principal.user_id`` to be the task's
-                # owner on both backends. Three callers land here: an admin
+                # owner on both backends. Four callers land here: an admin
                 # answering someone else's task, a guest whose bindings
                 # match but whose ``principal.user_id`` is not the owner's
                 # (both of which pass step 3's Python authorization and
                 # fail only at the write point), and, on SQLite alone,
-                # ownership changing between step 2's read and this
-                # statement. See ``_answer_fence_task_predicate``'s own
-                # docstring for which of its six terms are re-asserted in
-                # SQL and which are checked once in Python.
+                # ``Task.user_id`` changing between step 2's read and this
+                # statement, or, also SQLite-only, a guest's
+                # ``agent_config["guest_id"]`` changing in that same window.
+                # See ``_answer_fence_task_predicate``'s own docstring for
+                # which of its six terms are re-asserted in SQL and which
+                # are checked once in Python.
                 if task.status != TaskStatus.WAITING_FOR_USER:
+                    # Covers every other ``TaskStatus`` member: PENDING,
+                    # RUNNING, PAUSED, COMPLETED, FAILED -- five non-waiting
+                    # states this one comparison treats alike, without
+                    # distinguishing which of them the task actually moved
+                    # to.
                     return RespondStale(reason="run_ended")
                 if task.run_id != reread.run_id:
                     return RespondStale(reason="foreign_run")
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -51,13 +51,14 @@ discriminated unions; the ``create()`` typed seam (validates and returns,
 does not stage a row); ``get()``/``list_active()``; the three-tier
 compatibility materialization view; the answer fence and its active-row
 predicate; ``respond()``'s own call body (validation, authorization, the
-idempotency pre-read, anchor resolution, the answer fence, the task
-control-state transition, staging the resume command, and committing or
-reconciling an ambiguous acknowledgment -- this delivery classifies a
-staging conflict and reconciles an ambiguous commit against the durable
-graph, and still reports an unresolved fence miss as
-``RespondOutcomeUnknown`` rather than classifying it, which is a separate
-follow-up build); and the response-conflict counter
+idempotency pre-read, anchor resolution, the answer fence and its
+zero-rowcount classification, the task control-state transition, staging
+the resume command, and committing or reconciling an ambiguous
+acknowledgment -- this delivery classifies why the fence missed,
+classifies a staging conflict, and reconciles an ambiguous commit against
+the durable graph, so ``RespondOutcomeUnknown`` is left reporting only a
+commit whose acknowledgment was lost and whose reconciliation could not
+confirm what landed); and the response-conflict counter
 (``COUNTER_LIFECYCLE_RESPONSE_CONFLICT``).
 
 Not delivered here, and named so a reader does not go looking for them in
@@ -554,8 +555,15 @@ RespondUnauthorizedReason = Literal["not_task_principal"]
 RespondUnavailableReason = Literal[
     "task_missing", "interaction_missing", "checkpoint_unavailable"
 ]
-RespondConflictReason = Literal["idempotency_key_reused"]
-RespondStaleReason = Literal["anchor_dangling"]
+RespondConflictReason = Literal["already_answered", "idempotency_key_reused"]
+RespondStaleReason = Literal[
+    "expired",
+    "run_superseded",
+    "answered_via_chat",
+    "run_ended",
+    "foreign_run",
+    "anchor_dangling",
+]
 
 
 @dataclass(frozen=True)
@@ -596,27 +604,31 @@ class RespondStale:
 @dataclass(frozen=True)
 class RespondOutcomeUnknown:
     """A call this function could not resolve to one of the specific
-    outcomes above, for either of two reasons this build does not further
-    distinguish: the answer fence's UPDATE matched zero rows and this
-    build does not classify why (step 6 -- a future delivery may reread
-    and classify that miss the way the fence classification build does),
-    or committing raised and the durable-graph reconciliation this build
-    runs afterward could not confirm the write landed on every one of its
-    three attempts (step 9). Step 8's staging race is no longer one of
-    them: both of its doors are classified into ``Replayed`` or
-    ``Conflict`` now. Not an exception this service lets escape -- a
-    stable, typed result a caller can act on (e.g. surface "we could not
-    confirm this went through" rather than crash).
+    outcomes above, for the one reason that still reaches it: committing
+    raised, and the durable-graph reconciliation this build runs
+    afterward could not confirm the write landed on every one of its
+    three attempts (step 9) -- the acknowledgment could have been lost
+    after the server applied the write, and reading the graph back did
+    not settle it either way, so this build reports the ambiguity rather
+    than guessing. Neither of the other two races arrives here any more.
+    The answer fence's UPDATE matching zero rows (step 6) is *not* one of
+    them: this build rereads the row and classifies every such miss into
+    a specific outcome. Step 8's staging race is not one either: both of
+    its doors are classified into ``Replayed`` or ``Conflict`` now. Not an
+    exception this service lets escape -- a stable, typed result a caller
+    can act on (e.g. surface "we could not confirm this went through"
+    rather than crash).
 
-    Retrying under the same idempotency key resolves the second reason and
-    not the first, and a caller should not be told otherwise. A retry
+    Retrying under the same idempotency key resolves every reason that
+    still reaches this outcome, and that is now the whole set. A retry
     after an ambiguous commit finds the command this call staged and
-    returns ``Replayed``. A retry after a fence miss re-evaluates the same
-    predicate against the same row and misses again: a terminated,
-    superseded, foreign-run, or foreign-owned row stays that way, so the
-    second ``OutcomeUnknown`` is as final as the first. Step 6 logs the
-    reread row's state for that reason -- until the fence classification
-    build lands, that log line is the only place the distinction exists.
+    returns ``Replayed``. What a retry could never have clarified on its
+    own is a fence miss -- a terminated, superseded, foreign-run, or
+    foreign-owned row stays that way and misses again -- but that case no
+    longer lands on this outcome: step 6 hands the caller the specific
+    reason directly, and its own warning log records the reread row state
+    (``active_slot``, ``terminal_reason``, ``run_id``,
+    ``responder_identity``) that no outcome variant carries.
     """
 
 
@@ -1609,15 +1621,21 @@ def _respond_receipt(
     ``principal.identity_string()`` value, so the column and the local are
     equal by the fence write's own semantics, not by coincidence.
 
-    Two callers, each with its own precondition for why the row it hands in
-    is already answered. ``respond()``'s idempotent-replay pre-read branch
-    (step 5) calls this on a row it found by this call's own idempotency
-    key -- a staged RESUME command this service itself committed in some
-    earlier transaction alongside the answer fence UPDATE, which implies an
-    answered row. ``_verify_respond_durable_graph`` calls this only after
-    its own three checks already confirmed the row it is holding matches:
-    ``status == "answered"``, the answering identity, and the canonical
-    submitted payload. Either way, the paired CHECK constraints
+    Three callers, each with its own precondition for why the row it hands
+    in is already answered. ``respond()``'s idempotent-replay pre-read
+    branch (step 5) calls this on a row it found by this call's own
+    idempotency key -- a staged RESUME command this service itself
+    committed in some earlier transaction alongside the answer fence
+    UPDATE, which implies an answered row. ``respond()``'s fence-miss
+    classification (step 6) calls it on the row its own reread just read
+    back as ``status == "answered"``, paired with a command row found
+    under this call's own idempotency key whose payload matches -- the
+    same two facts the pre-read branch stands on, established one
+    statement earlier instead of before the fence attempt.
+    ``_verify_respond_durable_graph`` calls this only after its own three
+    checks already confirmed the row it is holding matches: ``status ==
+    "answered"``, the answering identity, and the canonical submitted
+    payload. In all three cases, the paired CHECK constraints
     (``ck_task_interaction_requests_responded_at_pairs_status`` and
     ``ck_task_interaction_requests_responder_pairs_responded_at``) make an
     answered row with a NULL ``responded_at`` or ``responder_identity``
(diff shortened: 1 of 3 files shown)


src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
...
+            if reread.status == "answered":
+                fresh_command = (
+                    db.query(TaskExecutionCommand)
+                    .filter(
+                        TaskExecutionCommand.task_id == task_id,
+                        TaskExecutionCommand.command_id == normalized_key,
+                    )
+                    .first()
+                )
+                if fresh_command is not None and _matches_existing(
```
rogercloud: **M1 (minor)** — This `Replayed` return is the one branch of the answered-reread path no test exercises: it is reachable only through a same-key, same-actor race between step 5's pre-read (task_interaction_service.py:1999-2015) and the fence, and deleting it keeps the suite green (R1/R2 pin the pair at step 5, and the AST meta-test at test_task_interaction_service.py:2929+ only checks pair coverage, not branch coverage). Suggest a SQLite test that answers the row and stages the matching command mid-flight using a racing-fence monkeypatch — the pattern already exists at test_task_interaction_service.py:2520-2561 (`_racing_fence_stmt_that_changes_ownership`); generalize it to answer the row + stage the command instead of changing ownership.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
...
+                    return RespondReplayed(
+                        receipt=_respond_receipt(
+                            interaction=reread,
+                            task=task,
+                            command_db_id=int(fresh_command.id),
+                            idempotency_key=normalized_key,
+                        )
+                    )
+                increment_counter(COUNTER_LIFECYCLE_RESPONSE_CONFLICT)
+                return RespondConflict(reason="already_answered")
```
rogercloud, line 2378: **M2 (minor)** — For an answered reread row with a command for this key that *mismatches*, this returns `Conflict(already_answered)`, while step 5's twin (task_interaction_service.py:2014-2015) returns `Conflict(idempotency_key_reused)` for the same logical situation — the reason a caller gets depends on race timing even though the code has the distinguishing information in hand. Consider checking `fresh_command` mismatch → `idempotency_key_reused` before falling through to `already_answered` (the counter should increment either way).

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
...
+                return RespondStale(reason=mapped_reason)
+            if reread.status == "active":
+                # Same order as the fence's own WHERE clause narrows: the
+                # task-level terms first, the ownership term last. Reaching
+                # the final return means the row is still live and the task
+                # is still waiting on this very run, so the only fence term
+                # left that can have failed is
+                # ``_answer_fence_task_predicate``'s ownership conjunction
+                # -- which requires ``principal.user_id`` to be the task's
+                # owner on both backends. Three callers land here: an admin
```
rogercloud: **M3 (minor)** — The comment enumerates 'three callers' that land on this `Unauthorized` fallthrough, but there is a fourth: the guest `agent_config.guest_id` SQL term (task_interaction_service.py:1425-1426) can fail on its own — a concurrent `agent_config` rewrite between step 3's read and the fence on SQLite's unlocked window — and lands here identically. The label is still correct; the comment just overclaims exhaustiveness. Name the guest-term case (or say the list is illustrative).

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2025,12 +2046,7 @@ def respond(
...
             # ORM would hand this query's result back through the same
             # already-loaded, now-stale Python object instead of the fresh
-            # row this reread exists to see. This build does not classify
-            # why the fence missed -- a fine-grained classification against
-            # the reread row's status is not delivered here -- it only
-            # confirms the row did not disappear out from under this
-            # transaction's own row lock, then reports the miss as
-            # ``OutcomeUnknown`` (see that outcome's own docstring).
+            # row this reread exists to see.
             db.expire_all()
```
rogercloud, line 2291: **M4 (minor)** — This comment explains `expire_all()` only in terms of the interaction-row reread, but the classification's `task.status`/`task.run_id` guards (task_interaction_service.py:2146-2149) depend on the same expiry: on SQLite the ORM would otherwise hand back the stale task object loaded at step 2, silently misclassifying `run_ended`/`foreign_run` as `Unauthorized(not_task_principal)`. A future 'optimization' to `db.expire(ir)` alone would therefore break the classification without touching this comment (the comment block above spans lines 2042-2049). Add one sentence naming the task-refresh dependency.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2039,16 +2055,24 @@ def respond(
...
-            # own: the caller is told ``OutcomeUnknown``, and a retry
-            # against a terminated, superseded, or foreign-owned row
-            # produces the same miss and the same ``OutcomeUnknown`` every
-            # time, so no amount of retrying ever reveals which of them it
-            # was -- unlike the commit-exception door below, where a retry
-            # under the same idempotency key resolves the ambiguity by
-            # itself. The classification this build does not compute is
-            # exactly the set of fields logged here.
+            # what it found costs no extra statement. The classification
+            # below tells the *caller* which miss this was; this line tells
```
rogercloud, line 2300: **M5 (minor)** — 'which miss this was' overstates what the reread can tell: the classification labels the *reread* row state, not necessarily the fence-time cause. On PostgreSQL a reclaim/legacy-close writer can commit a terminal transition between the 0-rowcount fence and this reread — e.g. an admin ownership miss rereads as terminated and comes back `Stale(expired)` instead of `Unauthorized`. Every label is a true statement about the current row; the contract is just never stated. Add one sentence: labels describe the reread row, and the fence cause is inferred, not observed.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -553,12 +654,14 @@ def test_two_concurrent_respond_calls_serialize_on_the_tasks_row(_respond_pg) ->
...
-    (see ``RespondOutcomeUnknown``'s own docstring) -- a more detailed
-    build's loser would land on ``Stale`` or ``Conflict``, this build's
-    lands on ``OutcomeUnknown`` -- so the loser is accepted here as any
-    non-``Accepted`` outcome; the invariant this test exists to pin is
-    "never both win", not which label the loser gets."""
+    commits. This build classifies that miss, so the loser now gets a
+    specific outcome rather than the conservative sibling's
+    ``OutcomeUnknown`` -- but the assertion below deliberately stays wide,
+    accepting any ``Stale`` or ``Conflict``. Which of them the loser lands
+    on depends on how far the winner's transaction had progressed when the
```
rogercloud: **M6 (nit)** — The concurrency model in this docstring is wrong: the loser blocks at step 2's tasks-row lock and only resumes after the winner's transaction has committed atomically, so it deterministically lands on `Conflict(already_answered)` (different keys — its step-5 pre-read finds no command for its own key and the reread sees the answered row). The label never depends on 'how far the winner's transaction had progressed' (docstring lines 654-664). The wide assertion is harmless; consider stating the deterministic outcome instead.

tests/web/services/test_task_interaction_service.py
```diff
@@ -2306,6 +2363,216 @@ def test_respond_reports_conflict_when_a_guest_and_the_owner_share_one_key(
...
+        _respond_db, task_id=task_id, interaction_id=interaction_id
+    ):
+        outcome = svc.respond(
+            interaction_id=interaction_id,
+            task_id=task_id,
+            principal=_owning_principal(user_id),
+            envelope=_respond_envelope(idempotency_key="never-seen-before-terminal"),
+        )
+
+        # Written out as three literal constructions rather than the one
```
rogercloud, line 2581: **M7 (nit)** — The stated reason for this workaround is false: the AST meta-test (`test_every_vocabulary_pair_is_produced_by_at_least_one_cell_test`, test_task_interaction_service.py:2929+) asserts produced ⊇ vocabulary, so folding to `svc.RespondStale(reason=expected_reason)` makes it fail LOUDLY — three pairs would vanish from the scanned set and the test would go red, not 'silently drop three Stale pairs … while every test still passed'. Fold the three-literal if/elif (lines 2401-2406) and delete this comment; if the fold is instead kept out for scan-visibility, say that honestly — it is a deliberate narrowing of the meta-test's scan, not a silent pass.

tests/web/services/test_task_interaction_service.py
```diff
@@ -1817,7 +1867,7 @@ def test_respond_rejects_a_guest_whose_bindings_match_but_principal_user_id_does
             envelope=_respond_envelope(),
         )
 
-        assert isinstance(outcome, svc.RespondOutcomeUnknown)
+        assert outcome == svc.RespondUnauthorized(reason="not_task_principal")
```
rogercloud, line 1887: **M8 (minor)** — This assertion is now byte-identical to the step-3 refusal asserted at test_task_interaction_service.py:1813 and 1909, so the test no longer pins that the guest reached the fence (in the base build, `isinstance(OutcomeUnknown)` was producible only at step 6). A future hardening of `task_is_owned_by_public_principal` that reads `principal.user_id` would flip this call to a step-3 refusal and the test would stay green, silently losing the fence-reach coverage. The A8 TOCTOU test (test_task_interaction_service.py:2546+) covers the fallthrough for a user principal, not the guest cell — consider a guest-specific racing-fence test, or at least a comment asserting the fence-reach intent here.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -11,19 +11,16 @@
   backends agree. (The TaskStatus-bind-parameter structural guard this
   bullet used to also cover needs no database at all and lives in the
   sibling SQLite-backed file instead.)
-- The three answered-row CHECK constraints. (This build does not classify
-  a rowcount=0 fence miss -- see ``RespondOutcomeUnknown``'s own
-  docstring -- so the six non-active-row classification cells a more
-  detailed build would need proven against a real PostgreSQL server are
-  not delivered here either.)
+- ``respond()``'s six non-active-row rowcount=0 classifications and the
```
rogercloud: **M9 (minor)** — Two gaps around this bullet: (a) two of the 'six non-active-row rowcount=0 classifications' are ACTIVE-row classifications — the `run_ended` and `foreign_run` cases in `test_answer_fence_rowcount_across_six_non_active_states` (lines 498-521) leave the row `active` and move the task — so 'non-active-row' is wrong for a third of the list; (b) the seventh branch, `Unauthorized(not_task_principal)`, which the PR's own step-6 docstring (task_interaction_service.py:2136-2137) says is reachable on PostgreSQL, has zero end-to-end `respond()` tests against PG (no `Unauthorized` hit anywhere in this file). Consider adding the PG guest/admin fence-miss cell to the six-state test.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
...
+                # owner on both backends. Three callers land here: an admin
+                # answering someone else's task, a guest whose bindings
+                # match but whose ``principal.user_id`` is not the owner's
+                # (both of which pass step 3's Python authorization and
+                # fail only at the write point), and, on SQLite alone,
+                # ownership changing between step 2's read and this
+                # statement. See ``_answer_fence_task_predicate``'s own
+                # docstring for which of its six terms are re-asserted in
+                # SQL and which are checked once in Python.
+                if task.status != TaskStatus.WAITING_FOR_USER:
```
rogercloud, line 2423: **M10 (nit)** — The reason literal `run_ended` overclaims: this guard covers all five non-WAITING statuses, including PENDING (never started) and PAUSED, and RUNNING (in progress — the suite itself pins that at test_task_interaction_service.py:2427-2435, where a RUNNING task produces `run_ended`). Only COMPLETED/FAILED correspond to an ended run. The docstrings consistently say 'the task is no longer waiting' — consider a literal (or mapping) that names that condition rather than asserting the run ended.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
...
+            if reread.status == "terminated":
+                terminal_reason_map: dict[str, RespondStaleReason] = {
+                    "deadline_elapsed": "expired",
+                    "run_superseded": "run_superseded",
+                    "answered_via_legacy_resume": "answered_via_chat",
+                }
+                mapped_reason = terminal_reason_map.get(
+                    str(reread.terminal_reason or "")
+                )
+                if mapped_reason is None:
```
rogercloud, line 2397: **M11 (minor)** — Both new `RuntimeError` invariants (this one for `terminal_reason`, and the status guard at task_interaction_service.py:2151-2155) are zero-covered. They are unreachable on honest data — the DB CHECK constraints close both vocabularies — but they are cheaply testable through the existing monkeypatch-svc.get pattern (test_task_interaction_service.py:1154-1158 and 2050-2054): a fabricated reread row would hit each raise. The author's own meta-test enforces full pair coverage, making these the sole uncovered branches against that standard; two small tests would close the gap.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
                 reread.run_id,
                 reread.responder_identity,
             )
-            return RespondOutcomeUnknown()
+            if reread.status == "answered":
+                fresh_command = (
```
rogercloud, line 2337: **M12 (minor)** — This ~18-line replay lookup (2089-2110: same query, same `_matches_existing`, same `_respond_receipt`) is a near-copy of step 5's pre-read block at task_interaction_service.py:1991-2015, differing only in the row source (`reread` vs `ir`) and the fallback reason. Extract a shared helper that returns the matching command (or None) and call it from both places — the copy is exactly where M2's reason drift (idempotency_key_reused vs already_answered) crept in.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2061,7 +2085,74 @@ def respond(
...
+                            interaction=reread,
+                            task=task,
+                            command_db_id=int(fresh_command.id),
+                            idempotency_key=normalized_key,
+                        )
+                    )
+                increment_counter(COUNTER_LIFECYCLE_RESPONSE_CONFLICT)
+                return RespondConflict(reason="already_answered")
+            if reread.status == "terminated":
+                terminal_reason_map: dict[str, RespondStaleReason] = {
```
rogercloud: **S1 (simplification)** — `delete:` the `terminal_reason_map` dict — three direct `if reread.terminal_reason == ...` comparisons plus the trailing `raise RuntimeError` are ~4 lines shorter and drop the `str(... or '')` None-coercion, which is dead defensiveness (the pairing CHECK makes `terminal_reason` NOT NULL for terminated rows).

tests/web/services/test_task_interaction_service.py
```diff
@@ -1470,6 +1484,40 @@ def _answered_row_with_valid_anchor(
         db.close()
 
 
+def _terminated_row_with_valid_anchor(
```
rogercloud, line 1504: **S2 (simplification)** — `shrink:` this helper is ~85% duplicate of `_answered_row_with_valid_anchor` (test_task_interaction_service.py:1455-1484) — same trace event, same row shape, same live anchor, differing only in the close-out fields (`status`, `active_slot`, `terminal_reason`, `terminated_at`, key string). Merge into one status-parametrized close-out helper (~34 lines vs 71 today). The distinct idempotency-key strings serve no purpose: the service never matches the stored key — replay detection runs on `TaskExecutionCommand.command_id`.

rogercloud (approved):
## PR #1376 — feat(interactions): restore respond()'s fine-grained answer fence classification

**Summary:** The merged #1354 collapsed every zero-row answer-fence UPDATE in `respond()` onto `RespondOutcomeUnknown`. This PR replaces that step-6 catch-all with a reread-and-classify block: an answered row returns `Replayed` (when a matching command for the key exists) or `Conflict(already_answered)` plus the response-conflict counter, a terminated row maps the three DB-known `terminal_reason`s to `Stale(expired / run_superseded / answered_via_chat)` and raises `RuntimeError` on anything else, an active row lands on `Stale(run_ended)`, `Stale(foreign_run)`, or `Unauthorized(not_task_principal)` (ownership term), and an unrecognized status raises. The test matrix grew from 27 cells/14 pairs to 34 cells/20 pairs with an AST-scanning meta-test enforcing pair coverage, and the PostgreSQL suite now exercises the rowcount=0 classifications end to end through real `respond()` calls. CI is all green.

**Approach verdict (Round 0): sound.** The reread-and-classify design is the right shape: the reread row state is the only durable evidence a fence miss leaves, the classification order mirrors the fence's own WHERE-clause narrowing (task terms first, ownership last), the two `RuntimeError` invariants are appropriately loud rather than folded into a normal-looking `OutcomeUnknown`, and keeping the operator log line is right since no outcome variant carries the row columns. The design-level concerns are comment-contract issues, not behavior issues, and are raised inline: M3 (the 'three callers' enumeration overclaims — a fourth caller exists), M4 (the `expire_all()` comment understates the classification's dependency on the task refresh), M5 (labels classify the reread state, not the fence-time cause — the contract is never stated), and M10 (the `run_ended` literal overclaims). None affects delivered behavior.

## Findings — minor

- **M1 — `task_interaction_service.py:2088-2112` (inline at 2097).** The answered-branch `Replayed` return is reachable only via a same-key, same-actor race between step 5's pre-read and the fence, and has ZERO test coverage — deleting it keeps the suite green (R1/R2 cover the pair at step 5; the AST meta-test only checks pair coverage). Suggest a SQLite test using a racing-fence monkeypatch (pattern exists at `test_task_interaction_service.py:2520-2561`) that answers the row and stages the matching command mid-flight.
- **M2 — `task_interaction_service.py:2111-2112` (inline).** When the reread row is answered AND a command for this key exists but mismatches, the code returns `Conflict(already_answered)`; step 5's twin (`:2014-2015`) returns `Conflict(idempotency_key_reused)` for the same condition. Same logical situation, different reason depending on race timing, even though the code has the distinguishing info. Fix: check `fresh_command` mismatch → `idempotency_key_reused` before falling to `already_answered`.
- **M3 — `task_interaction_service.py:2130-2145` (inline at 2137).** The 'only fence term left … ownership conjunction' comment enumerates three callers but omits a fourth: the guest `agent_config.guest_id` SQL term (`:1425-1426`) can fail alone (SQLite concurrent config rewrite between step 3 and the fence) and lands on the same `Unauthorized` fallthrough. Label still correct; comment overclaims exhaustiveness.
- **M4 — `task_interaction_service.py:2042-2050` (inline at 2050).** The `expire_all()` comment explains the expiry only in terms of the interaction-row reread; the classification's `task.status`/`task.run_id` guards (`:2146-2149`) depend on it equally (on SQLite, without the post-write refresh the ORM serves the stale task object, silently misclassifying `run_ended`/`foreign_run` as `Unauthorized`). A future targeted `db.expire(ir)` 'optimization' would break the classification without touching this comment. Add one sentence naming the task-refresh dependency.
- **M5 — `task_interaction_service.py:2056-2074` (inline at 2059).** Labels classify the REREAD row state, not necessarily the fence-time cause: on PostgreSQL a reclaim/legacy-close writer can commit a terminal transition between a 0-rowcount fence and the reread (e.g. an admin ownership miss rereads as terminated → `Stale(expired)` instead of `Unauthorized`). Every label is a true statement about the current row; the contract is never stated and comments say 'which miss this was'. Add one sentence stating the contract.
- **M8 — `test_task_interaction_service.py:1870` (inline).** The guest-bindings test's assertion changed from `isinstance(OutcomeUnknown)` (producible only at step 6 in the base) to `== RespondUnauthorized(not_task_principal)` — byte-identical to step 3's refusal, so the test no longer pins that the guest reached the fence; a future `task_is_owned_by_public_principal` hardening (reading `principal.user_id`) would silently flip this test to a step-3 refusal and stay green. The A8 TOCTOU test covers the fallthrough for a user principal, not the guest cell.
- **M9 — `test_task_interaction_service_postgresql.py:14-17, 419-521` (inline at 14).** (a) The module docstring says 'six non-active-row rowcount=0 classifications', but two of the six cases (`run_ended`, `foreign_run`) are ACTIVE-row classifications. (b) The seventh branch — `Unauthorized(not_task_principal)`, which the PR's own step-6 docstring says is reachable on PG — has zero end-to-end `respond()` tests against PostgreSQL (no `Unauthorized` hit anywhere in the PG file).
- **M11 — `task_interaction_service.py:2121-2127, 2151-2155` (inline at 2122).** The two new `RuntimeError` invariants (unrecognized `terminal_reason` / `status`) are zero-covered. They are unreachable on honest data (DB CHECKs close both vocabularies) but cheaply testable via the existing monkeypatch-svc.get pattern (`test_task_interaction_service.py:1154-1158, 2050-2054`) — a fabricated reread row would hit each raise. The author's own meta-test enforces full pair coverage, making these the sole uncovered branches against their own standard.
- **M12 — `task_interaction_service.py:2089-2110` vs `1991-2015` (inline at 2089).** The answered branch's replay lookup is an ~18-line near-copy of step 5's pre-read block (same query, same `_matches_existing`, same `_respond_receipt`; differing only in row source and fallback reason). Extract a shared helper returning the matching command or None — the copy is exactly where M2's reason drift crept in.

## Findings — nit

- **M6 — `test_task_interaction_service_postgresql.py:654-664` (inline at 661).** The serialize-test docstring claims the loser's label 'depends on how far the winner's transaction had progressed' — false: the loser blocks at step 2's tasks-row lock until the winner's atomic commit, then deterministically lands on `Conflict(already_answered)` (different keys). The wide assertion is harmless; the rationale is a false concurrency model.
- **M7 — `test_task_interaction_service.py:2394-2406` (inline at 2394).** The three-literal if/elif workaround's comment claims folding to `svc.RespondStale(reason=expected_reason)` 'would silently drop three Stale pairs … while every test still passed' — false: the AST meta-test (`:2929+`) asserts produced ⊇ expected, so the fold makes it fail LOUDLY. The workaround is unnecessary and the comment's stated reason is wrong; fold and delete the misleading comment (or fix the comment).
- **M10 — `task_interaction_service.py:2146-2147` (inline).** The reason literal `run_ended` overclaims — the guard covers ALL five non-WAITING states including PENDING (never started), RUNNING (in progress — the suite itself pins this at `test_task_interaction_service.py:2427-2435`), and PAUSED; only COMPLETED/FAILED have ended runs. Docstrings consistently use 'the task is no longer waiting'; consider a literal that names that.
- **M13 — PR body only.** The description lists the stale family as (run_ended, run_superseded, question_expired, question_withdrawn, foreign_run) — the delivered vocabulary (`task_interaction_service.py:549-556`) is (expired, run_superseded, answered_via_chat, run_ended, foreign_run, anchor_dangling); 'question_withdrawn' exists nowhere in src, and 'answered_via_chat' is delivered but absent from the list.
- **M14 — `test_task_interaction_service.py:171` (body-only, pre-existing).** A comment above the meta-test still says it 'confirms the 14 (outcome, reason) pairs'; the vocabulary is now 20. The PR updated the module docstring and the meta-test docstring but not this adjacent comment. Not in the diff.
- **M15 — `task_interaction_service.py:2017-2021` (body-only, pre-existing, informational).** A fresh-key call on an ANSWERED row whose resume anchor was pruned is turned away at step 5.5 with `Stale(anchor_dangling)` before the classification can see it's answered. Behavior byte-identical to base; the PR's 'classifies every fence miss' claim is technically scoped to fence misses. Optional follow-up.

- `L2114: delete: terminal_reason_map dict with three one-off entries plus the str(... or '') None-coercion. Replace with three direct if reread.terminal_reason == ... comparisons and the trailing raise (~4 lines shorter; the coercion is dead defensiveness since the pairing CHECK makes terminal_reason NOT NULL for terminated rows).`
- `L1487: shrink: _terminated_row_with_valid_anchor is ~85% duplicate of _answered_row_with_valid_anchor (~1448-1484) — same trace event, same row shape, same anchor, differing only in the close-out fields. Merge into one status-parametrized close-out helper (~34 lines vs 71 today; the distinct idempotency-key strings serve no purpose since the service never matches the stored key — replay detection runs on TaskExecutionCommand.command_id).`
- `net: -41 lines possible`

Blocking: no — no blocking issues; all findings are minor or nit-level and none affects delivered behavior. CI is all green. Recommended event: **APPROVE**.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. Two race tests now exercise this branch through exactly the pattern you pointed at — the racing-fence monkeypatch, generalized to answer the row and stage a same-key command mid-flight between the pre-read and the fence: one with a matching payload pins `Replayed` naming the racer's command row, one with a mismatched payload pins the conflict verdict. Deleting the replay recognition turns the first red (confirmed: before this round the whole suite stayed green with the block removed).

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. A same-key command whose payload does not match now resolves to `Conflict(idempotency_key_reused)` here as well, so the verdict no longer depends on which side of the fence the race landed; `already_answered` is reserved for the case where the row is answered and no command exists under this key at all. This also brings the fence-miss branch in line with the staging-side classification in the sibling change, so all three points that can see a same-key mismatch now say the same thing. The mismatched-payload race test pins it — reverting the verdict turns exactly that test red.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed — and there were two copies of that enumeration, not one: the same three-item list also lives in `respond()`'s main docstring. Both now name the fourth arrival: the guest `guest_id` term failing on its own after a concurrent `agent_config` rewrite in the unlocked SQLite window. Reproduced before writing it (guest id changed at fence time with the owner unchanged lands on exactly this fallthrough).

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed, and worth stating how real this was: narrowing to `db.expire(ir)` reclassifies a run-ended miss as `Unauthorized(not_task_principal)` — "you're too late" reported as "you're not allowed" — and the full suite stayed green under that mutation. The comment now names the task-row refresh dependency alongside the interaction-row one; the two race tests added this round give the classification's task-side reads their first behavioural pin.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed — the classification's contract now states it in one sentence: every label describes the row and task as reread after the miss, and the fence-time cause is inferred from that state, not observed. Your PostgreSQL example (a terminal writer committing between the zero-rowcount fence and the reread) is exactly the case that sentence exists for.

tests/web/services/test_task_interaction_service_postgresql.py
AlexLiu190625: Fixed. The docstring now traces the deterministic outcome instead of hedging: the loser blocks on the tasks-row lock until the winner's transaction has committed whole, its own pre-read finds no command under its key, and the reread sees the answered row — so it lands on `Conflict(already_answered)` regardless of interleaving. The assertion is left wide since the deterministic claim is a statement about the code path, not something the test needs to narrow to prove.

tests/web/services/test_task_interaction_service.py
AlexLiu190625: You were right about the mechanism and I have to decline the fold — the two halves of the comment contradict each other, and the first half wins on measurement: folding the three literals to `svc.RespondStale(reason=expected_reason)` makes the AST meta-test fail loudly, naming exactly the three pairs it can no longer see. So the branches stay, and the comment now says the true thing: the literals exist to keep the three cells visible to the meta-test's scan, and folding them would fail that test — a deliberate choice about the scan's reach, not a silent pass.

tests/web/services/test_task_interaction_service.py
AlexLiu190625: Agreed on the loss of distinguishing power, and recording it rather than adding the guest-specific racing-fence cell here — this round already adds two race tests against roughly seventy lines of classification logic, and a third for the guest arrival would push the test-to-code ratio well past what the branch justifies on its own. The docstring on this test states the fence-reach intent explicitly so the coverage gap is disclosed, and the guest racing cell is recorded for the change that wires the first production caller. Leaving open as the record.

tests/web/services/test_task_interaction_service_postgresql.py
AlexLiu190625: (a) fixed — in both places: the module docstring and the test's own name, which carried the same "non-active" claim (`test_answer_fence_rowcount_across_six_non_active_states` is now `..._six_fence_miss_states`; two of the six leave the row active and move the task). (b) recorded rather than added: the PostgreSQL guest/admin fence-miss cell is real coverage, but it belongs with the guest racing cell from the sibling thread — the same wiring change should land both together. Leaving open as the record for (b).

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Kept the literal, documented the coverage: the comment now states that `run_ended` covers all five non-waiting statuses (PENDING, RUNNING, PAUSED, COMPLETED, FAILED), with the running-task test as the pinned example. Renaming the reason is a vocabulary change on a `Literal` the meta-test derives cells from — not something to do in passing on an approved change; the docstrings consistently phrase the condition as "the task is no longer waiting", and the literal is now annotated to match that reading.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Agreed they are zero-covered and cheaply testable, and recording rather than adding them here for the same reason as the guest racing cell: the two invariants guard vocabularies the CHECK constraints already close, so the tests would pin a fabricated-row shape the database cannot produce — worth having under the meta-test's standard, but this round's test additions are already at the edge of what the change's own logic size supports. Recorded alongside the other deferred cells. Leaving open as the record.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Agreed the copy is where the drift crept in — the drift itself is fixed this round (sibling thread). Deferring the helper extraction: the sibling change that classifies the staging-side race touches the same lookup shape from a third direction, and extracting once after both land is one helper serving three sites rather than two extractions in sequence. Leaving open as the record.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Done in substance: the `str(... or "")` coercion is gone — it was dead defensiveness given the pairing CHECK, and it now reads the column through a static-only `cast` (mypy still needs the type asserted even though the constraint makes the value non-null). The mapping stays as a mapping rather than three comparisons because the trailing `RuntimeError` for an unrecognized reason is the loud-failure invariant the meta-test tracks; the net line count is the same either way.

tests/web/services/test_task_interaction_service.py
AlexLiu190625: Not merging these two — the estimate doesn't hold up. The two helpers take different parameter sets, so a status-parametrized merge has to grow optional arguments for the close-out fields, and the result is not the 34-line version described; the readability cost of one helper with a mode switch outweighs the duplication in two short setup functions. Your point about the idempotency-key strings serving no matching purpose is correct (replay runs on the command id), and that's recorded. Leaving open as the record.

PR #1377, opened 2026-08-15
feat(interactions): reconcile the durable graph after an ambiguous commit

AlexLiu190625:
## Summary

Reconciles an ambiguous commit against the durable graph, and classifies staging conflicts precisely. The merged answer-side entry point (#1354) reports two situations as `RespondOutcomeUnknown` by design: a commit that raised after the answer may or may not have become durable, and a staged-command collision with a concurrent writer. This change delivers the precision that build deferred.

## What's here

- `_verify_respond_durable_graph`: after a failed commit acknowledgment, a fresh session reads what actually landed. If the interaction row is answered under this call's identity and the staged command carries this call's payload, the answer was durable and the caller receives `Accepted` with the real receipt; a graph that landed differently (or not at all) keeps the unresolved outcome. Each verification attempt logs on failure; three exhausted attempts fall back to exactly what the merged build returns without this layer.
- Both staging-race doors classified the same way: the unique-constraint `IntegrityError` and the returned pre-existing row are two signals for one race, and both now resolve through the same comparison — a matching payload is a replay (the winning row carries this call's own actor, kind and canonical payload, so the RESUME about to execute is this answer), a mismatched one is a `Conflict`. The replay verdict additionally asserts the fence UPDATE and the control-state transition commit — without them the executing command would answer a row that never recorded an answer.
- The post-commit dispatcher notify is best-effort on both exits (the normal accept and the reconciled accept), through one shared helper carrying the merged build's guarantee: a durable answer is never reported as a failure because a wakeup raised.
- The commit-failure warning log stays, unconditionally, before reconciliation begins: even when reconciliation upgrades the result to `Accepted`, the lost acknowledgment itself is something an operator should be able to find.
- A database-level `IntegrityError` with no duplicate row behind it re-raises rather than folding into a typed outcome, and a test pins that escape.

## Verification

Full SQLite suite for the touched surfaces (116 in the service file, 186 with the command-transport suite), PostgreSQL-only suite against a disposable database, and red/green mutation for every restored branch: collapsing the raced-row door back to the catch-all turns both race tests red, ignoring the reconciliation verdict turns the landed-graph tests red, and removing the identity condition from the graph check turns the foreign-identity test red.

### DIFF
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -2400,7 +2400,45 @@ def respond(
             # that staged that row owes it (see ``stage_task_command``'s
             # caller obligation (a)), and the dispatcher's idle poll is the
             # documented fallback either way.
-            db.commit()
+            try:
+                db.commit()
+            except Exception:
+                # Same ambiguity as step 9's commit below, on the other
+                # door of this function's single race funnel: the fence
+                # UPDATE and CAS this call issued may have landed durably
+                # even though this call never saw the acknowledgment. The
+                # winner's command row is not this transaction's own write
+                # and is not rolled back by this exception, so a bare
+                # re-raise would turn an already-committed answer into a
+                # crash loop on retry -- a retry under the same
+                # idempotency key would land on the replay branch above
+                # and hit this same commit again. Reconcile against a
+                # fresh session instead, exactly as step 9 does.
+                logger.warning(
+                    "commit failed while answering interaction %s on task %s; "
+                    "the write may or may not be durable -- reconciling against "
+                    "the durable graph",
+                    interaction_id,
+                    task_id,
+                    exc_info=True,
+                )
+                session_retired = True
+                _retire_respond_session_best_effort(db)
+                receipt = _verify_respond_durable_graph(
+                    task_id=task_id,
+                    interaction_id=interaction_id,
+                    expected_run_id=run_id_for_verification,
+                    expected_state_version_after=expected_state_version_after,
+                    principal=principal,
+                    canonical_submitted_values=canonical_submitted_values,
+                    command_id=normalized_key,
+                    command_kind=TaskCommandKind.RESUME,
+                    command_payload=command_payload,
+                    actor_user_id=actor_user_id,
+                )
+                if receipt is not None:
+                    return RespondReplayed(receipt=receipt)
+                return RespondOutcomeUnknown()
             return RespondReplayed(
                 receipt=InteractionResponseReceipt(
                     interaction_id=interaction_id,
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -1609,10 +1609,15 @@ def _respond_receipt(
     ``principal.identity_string()`` value, so the column and the local are
     equal by the fence write's own semantics, not by coincidence.
 
-    The only caller is ``respond()``'s idempotent-replay branch, whose
-    precondition -- a staged RESUME command this service itself committed
-    in the same transaction as the answer fence UPDATE -- implies an
-    answered row, and the paired CHECK constraints
+    Two callers, each with its own precondition for why the row it hands in
+    is already answered. ``respond()``'s idempotent-replay pre-read branch
+    (step 5) calls this on a row it found by this call's own idempotency
+    key -- a staged RESUME command this service itself committed in some
+    earlier transaction alongside the answer fence UPDATE, which implies an
+    answered row. ``_verify_respond_durable_graph`` calls this only after
+    its own three checks already confirmed the row it is holding matches:
+    ``status == "answered"``, the answering identity, and the canonical
+    submitted payload. Either way, the paired CHECK constraints
     (``ck_task_interaction_requests_responded_at_pairs_status`` and
     ``ck_task_interaction_requests_responder_pairs_responded_at``) make an
     answered row with a NULL ``responded_at`` or ``responder_identity``
@@ -2356,7 +2361,15 @@ def respond(
                 # ambiguous" about a database-level failure. Re-raised.
                 db.rollback()
                 raise
-            assert classification.raced is not None
+            if classification.raced is None:
+                raise RuntimeError(
+                    f"classify_task_command_conflict reported RACED_DUPLICATE "
+                    f"for interaction {interaction_id} on task {task_id} "
+                    "with no raced projection attached; "
+                    "TaskCommandConflictClassification's own constructor "
+                    "always pairs RACED_DUPLICATE with a raced projection, "
+                    "making this impossible"
+                )
             raced_command_db_id = classification.raced.command_db_id
             raced_payload_matches = classification.raced.payload_matches
         else:
--- a/src/xagent/web/services/task_interaction_service.py
+++ b/src/xagent/web/services/task_interaction_service.py
@@ -10,7 +10,9 @@ both live on the ask side, and both require the caller to already hold the
 transaction they run inside. This module's ``respond()`` is the opposite
 shape on every one of those points: it owns its own session end to end
 (opens it, commits or rolls it back, retires it) and never nests inside a
-caller's transaction. ``create()``, by contrast, takes a caller-owned
+caller's transaction -- the one savepoint it opens
+(``db.begin_nested()``, around staging its own resume command) is its
+own, not a caller's. ``create()``, by contrast, takes a caller-owned
 session and only reads through it, never opening or committing one of its
 own. Putting ``respond()`` in the same file as the staging primitive would
 make that staging module's merge-reason docstring false. What this module
@@ -50,10 +52,12 @@ does not stage a row); ``get()``/``list_active()``; the three-tier
 compatibility materialization view; the answer fence and its active-row
 predicate; ``respond()``'s own call body (validation, authorization, the
 idempotency pre-read, anchor resolution, the answer fence, the task
-control-state transition, staging the resume command, and committing --
-this delivery reports an unresolved fence miss or a failed commit as
-``RespondOutcomeUnknown`` rather than further classifying or reconciling
-either); and the response-conflict counter
+control-state transition, staging the resume command, and committing or
+reconciling an ambiguous acknowledgment -- this delivery classifies a
+staging conflict and reconciles an ambiguous commit against the durable
+graph, and still reports an unresolved fence miss as
+``RespondOutcomeUnknown`` rather than classifying it, which is a separate
+follow-up build); and the response-conflict counter
 (``COUNTER_LIFECYCLE_RESPONSE_CONFLICT``).
 
 Not delivered here, and named so a reader does not go looking for them in
@@ -79,6 +83,7 @@ from __future__ import annotations
 
 import json
 import logging
+import time
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Literal, cast
@@ -107,9 +112,12 @@ from .ops_signals import (
     register_degradation,
 )
 from .task_command_transport import (
+    TaskCommandConflictKind,
     TaskCommandKind,
+    _canonical_payload,
     _matches_existing,
     _normalize_command_id,
+    classify_task_command_conflict,
     notify_task_command_dispatcher,
     stage_task_command,
 )
@@ -592,19 +600,18 @@ class RespondOutcomeUnknown:
     distinguish: the answer fence's UPDATE matched zero rows and this
     build does not classify why (step 6 -- a future delivery may reread
     and classify that miss the way the fence classification build does),
-    or committing raised and this build does not attempt to reconcile the
-    durable graph afterward (steps 8/9) -- the acknowledgment could have
-    been lost after the server applied the write, or a racing writer could
-    have taken the idempotency key first; either way this build reports
-    the ambiguity rather than guessing. Not an exception this service lets
-    escape -- a stable, typed result a caller can act on (e.g. surface "we
-    could not confirm this went through" rather than crash).
+    or committing raised and the durable-graph reconciliation this build
+    runs afterward could not confirm the write landed on every one of its
+    three attempts (step 9). Step 8's staging race is no longer one of
+    them: both of its doors are classified into ``Replayed`` or
+    ``Conflict`` now. Not an exception this service lets escape -- a
+    stable, typed result a caller can act on (e.g. surface "we could not
+    confirm this went through" rather than crash).
 
     Retrying under the same idempotency key resolves the second reason and
     not the first, and a caller should not be told otherwise. A retry
     after an ambiguous commit finds the command this call staged and
-    returns ``Replayed``, and a retry after a rolled-back staging race
-    simply runs again. A retry after a fence miss re-evaluates the same
+    returns ``Replayed``. A retry after a fence miss re-evaluates the same
     predicate against the same row and misses again: a terminated,
     superseded, foreign-run, or foreign-owned row stays that way, so the
     second ``OutcomeUnknown`` is as final as the first. Step 6 logs the
@@ -1635,6 +1642,10 @@ def _respond_receipt(
     )
 
 
+_RESPOND_DURABLE_GRAPH_ATTEMPTS = 3
+_RESPOND_DURABLE_GRAPH_RETRY_SLEEP_SECONDS = 0.01
+
+
 def _retire_respond_session_best_effort(db: "Session") -> None:
     """Release an owned Session without letting a close/invalidate failure
     replace the transaction error that is already in flight. Same shape as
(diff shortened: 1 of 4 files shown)


src/xagent/web/services/task_interaction_service.py
```diff
@@ -2102,86 +2302,137 @@ def respond(
...
+            # canonical payload, so the RESUME that executes is this
+            # answer. Commit the fence UPDATE and the CAS -- they are what
+            # makes the interaction row agree with the command that will
+            # run -- and report ``Replayed`` naming the winner's row rather
+            # than ``Accepted``, because the command row is not this
+            # call's. No dispatcher notification is sent here: the writer
+            # that staged that row owes it (see ``stage_task_command``'s
+            # caller obligation (a)), and the dispatcher's idle poll is the
+            # documented fallback either way.
+            db.commit()
```
rogercloud: This `db.commit()` is the only graph-committing statement in respond() not covered by the lost-ack reconciliation: the accept-path commit (line 2395) wraps failure with `_verify_respond_durable_graph`, but this Replayed-branch commit raises raw on a possibly-durable answer — the exact ambiguity class this PR's reconciliation exists to prevent. If the commit truly aborts, the interaction stays active while the winner's command row is durable; a retry hits step 5's pre-read, returns Replayed again, and `_respond_receipt`'s guard (1598-1604) raises RuntimeError — a permanent retry crash loop. (Nuance: the crash-loop class pre-existed in the base — same guard, same pre-read — and respond() has no production caller yet, so this PR narrows the class rather than introducing it.) Suggest routing this commit through the same try/except + reconcile shape; the verifier's three conditions already accept the winner's row.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1609,6 +1616,10 @@ def _respond_receipt(
     )
 
 
+_RESPOND_DURABLE_GRAPH_ATTEMPTS = 3
```
rogercloud, line 1650: Doc nit: the `_respond_receipt` docstring (lines 1586-1596) claims "The only caller is respond()'s idempotent-replay branch" with a same-transaction-staging precondition, but this PR adds a second caller, `_verify_respond_durable_graph` (line 1771), whose precondition is different (fresh-session read filtered by `status == answered`). The guard reasoning still holds — the paired CHECK constraints make an answered row with NULL `responded_at`/`responder_identity` impossible — but the "only caller"/precondition claim is now false. tests/web/services/test_task_interaction_service.py:2147-2149 repeats the stale claim.

tests/web/services/test_task_interaction_service.py
```diff
@@ -2613,7 +2862,70 @@ def _racing_stage_task_command(*args: Any, **kwargs: Any) -> StagedTaskCommand:
...
+    user_id, task_id = _waiting_task(_respond_db)
+    interaction_id = _active_row_ready_for_respond(_respond_db, task_id=task_id)
+    principal = _owning_principal(user_id)
+    before = _graph_snapshot(
+        _respond_db, task_id=task_id, interaction_id=interaction_id
+    )
+
+    def _racing_stage_task_command(*args: Any, **kwargs: Any) -> StagedTaskCommand:
+        return StagedTaskCommand(
+            staged_db_id=4242,
```
rogercloud, line 2904: This test monkeypatches `stage_task_command` to return `staged_db_id=4242` with no command row actually in the DB, then asserts the receipt names 4242 and that the fence/CAS committed. It would pass a regression that commits the fence/CAS while no winner row exists — the exact hazard the code comments (task_interaction_service.py:2367-2376) say the commit prevents — and the receipt assertion is circular (the expected id comes from the same monkeypatch). The PG race tests (test_task_interaction_service_postgresql.py:944-1115) anchor the receipt to a real row and assert the command table, so the gap is SQLite-only. Trivial fix: assert `after['commands_count'] == 1` or query the row with id 4242.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2102,86 +2302,137 @@ def respond(
...
+            savepoint.rollback()
+            classification = classify_task_command_conflict(
+                db,
+                task_id=task_id,
+                command_id=normalized_key,
+                actor_user_id=actor_user_id,
+                kind=TaskCommandKind.RESUME,
+                payload=command_payload,
+            )
+            if classification.kind == TaskCommandConflictKind.TASK_MISSING:
```
rogercloud, line 2351: This TASK_MISSING branch is provably unreachable through respond(): the task row has been held FOR KEY SHARE since step 2, blocking a concurrent DELETE, and the fence's unconditional `Task.user_id == principal.user_id` term means only the owner reaches step 8, so the actor FK cannot fail. The test docstring (test_task_interaction_service.py:2709-2749) acknowledges it is untested. enqueue_task_command keeps TASK_MISSING live, so the classification itself is fine — only this mapping branch is dead. Suggest deleting the branch (TASK_MISSING then falls into the existing '!= RACED_DUPLICATE' re-raise) and fixing the step-8 docstring mapping claim at lines 2010-2011.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -1630,6 +1641,164 @@ def _retire_respond_session_best_effort(db: "Session") -> None:
...
+    command payload the third check below compares); (3) exactly one
+    command row exists for this idempotency key and it matches what this
+    call staged. Returns the receipt on success, ``None`` if every attempt
+    fails to confirm all three.
+    """
+
+    from ..models.database import get_session_local
+
+    SessionLocal = get_session_local()
+    for attempt in range(_RESPOND_DURABLE_GRAPH_ATTEMPTS):
```
rogercloud, line 1754: This is the third copy of the fresh-session retry-loop plumbing: range(3), a SessionLocal per attempt, sleep(0.01), retire-in-finally, per-attempt warning log — the same shape as task_orchestrator.py:1145-1218 and websocket.py:4730-4802 (both pre-date this PR; the docstring at 1696-1701 admits the copy). A parameterized helper (attempts, sleep, retire, log_label, check: Callable[[Session], T]) would collapse ~20 lines per site with zero behavior change.

src/xagent/web/services/task_interaction_service.py
```diff
@@ -2102,86 +2302,137 @@ def respond(
...
+                # ``UNRELATED``: no duplicate row exists and the task is
+                # still there, so some other constraint on the row this
+                # call tried to insert failed -- most plausibly the
+                # ``actor_user_id`` foreign key. That is not a race this
+                # function has an answer for, and folding it into a typed
+                # outcome would tell the caller "your answer was
+                # ambiguous" about a database-level failure. Re-raised.
+                db.rollback()
+                raise
+            assert classification.raced is not None
```
rogercloud: Bare, message-less assert for a cross-module invariant, where this module's documented convention is an explicit RuntimeError-with-message (cf. 1599, 2209, 2229; respond()'s docstring 2054-2059). Under python -O the assert is stripped and line 2334 raises AttributeError instead. The invariant holds by construction today; the failure-mode difference only matters if classify's contract changes, but the convention here is loud-and-typed.

tests/web/services/test_task_interaction_service_postgresql.py
```diff
@@ -846,3 +860,256 @@ def respond_call() -> None:
...
+                assert release_event.wait(timeout=10)
+            real_add(instance)
+
+        db.add = paused_add
+        return db
+
+    return factory
+
+
+def test_respond_races_a_committed_duplicate_command_and_replays_on_matching_payload(
```
rogercloud, line 944: These two race tests (944-1021 and 1023-1115) duplicate ~26-29 lines of byte-identical scaffolding — session-factory monkeypatch, events, thread, racer, release/join — differing only in the payload variable and the verification sections. A parametrization over the payload-match axis would remove the duplication (the verification sections genuinely differ, so this is a hedged nit). Also worth noting: the replay branch's fence/CAS commit assertion exists only in the SQLite door-two test; door-one coverage is transitive through the shared decision block (task_interaction_service.py:2359-2390) — acceptable, worth keeping if the doors diverge.

rogercloud (approved):
## PR summary

This single-commit PR completes the precision deferred from merged #1354 in the respond() path: after a failed commit acknowledgment, `_verify_respond_durable_graph` re-reads the durable graph in a fresh session and returns `Accepted` with a real receipt only when the interaction row is answered under this identity with the canonical payload, exactly one command row matches actor/kind/payload, and `task.state_version >= expected` (>= because the coordinator re-issues RESUME_REQUESTED; `control_state` intentionally not compared); otherwise it falls back to `OutcomeUnknown`. Both staging-race doors (IntegrityError and `created=False`) now resolve through the single `_matches_existing` verdict: match -> `Replayed` with the fence+CAS commit, mismatch -> `Conflict`, `UNRELATED` re-raised, `TASK_MISSING` -> `Unavailable`; post-commit dispatcher notify is best-effort on both accept exits. respond() still has no production caller (pinned by the create-gate test), and fence-miss classification remains explicitly deferred.

## Approach (Round 0)

**Sound.** Converging both staging-race doors on one `_matches_existing` verdict removes the classification split, and the fresh-session reconciliation keeps the lost-ack ambiguity class contained. Two design-level concerns, both minor and folded into the findings below: the replay branch's commit (line 2377) sits outside the lost-ack reconciliation umbrella the accept path's commit gets (finding 1), and the `_respond_receipt` docstring's "only caller" claim is now stale (finding 2).

## Findings (all minor)

1. **[minor, error-handling] src/xagent/web/services/task_interaction_service.py:2377** — The Replayed branch's `db.commit()` is unguarded, the only graph-committing statement not covered by reconciliation or a typed outcome. A raised exception here leaves the interaction active while the winner's command row is durable — the exact ambiguity class this PR's reconciliation exists to prevent; a retry then re-reads the pre-read, returns Replayed again, and `_respond_receipt`'s guard (1598-1604) raises RuntimeError: a permanent retry crash loop. Nuance: the crash-loop class pre-existed in the base (same guard, same pre-read — the PR narrows it), and respond() has no production caller yet.
2. **[minor, docs] task_interaction_service.py:1586-1596** — `_respond_receipt`'s docstring claims "The only caller is respond()'s idempotent-replay branch" with a same-transaction-staging precondition; this PR adds a second caller, `_verify_respond_durable_graph` (line 1771), with a different precondition (fresh-session read filtered by `status == answered`). Guard reasoning still holds (paired CHECK constraints), but the claim is now false; tests/web/services/test_task_interaction_service.py:2147-2149 repeats it.
3. **[minor, test-quality] tests/web/services/test_task_interaction_service.py:2886-2927** — The SQLite replay test monkeypatches `stage_task_command` to return `staged_db_id=4242` with no command row in the DB, then asserts the receipt names 4242 and the fence/CAS committed. It would pass a regression that commits the fence/CAS while no winner row exists — the exact hazard the code comments (2367-2376) say the commit prevents — and the receipt assertion is circular (expected id from the same monkeypatch). The PG tests anchor the receipt to a real row and assert the command table, so the gap is SQLite-only.
4. **[minor, test-quality] task_interaction_service.py:2320-2322** — The TASK_MISSING mapping branch is provably unreachable through respond(): the task row is held FOR KEY SHARE since step 2 (blocking concurrent DELETE), and the fence's unconditional `Task.user_id == principal.user_id` term means only the owner reaches step 8, so the actor FK cannot fail; the test docstring (sqlite:2709-2749) acknowledges it is untested. enqueue keeps TASK_MISSING live, so the classification itself is fine — only respond()'s mapping branch is dead. Simplification: delete the branch and fix the docstring mapping claim at 2010-2011.
5. **[minor, code-quality] task_interaction_service.py:1705-1799** — `_verify_respond_durable_graph` is a third copy of the fresh-session retry-loop plumbing (task_orchestrator.py:1145-1218 and websocket.py:4730-4802 pre-date the PR): range(3), SessionLocal per attempt, sleep(0.01), retire-in-finally, per-attempt warning log; the code's own docstrings admit the copies.
6. **[minor, code-quality] task_interaction_service.py:2333** — Bare, message-less `assert classification.raced is not None` for a cross-module invariant, where this module's documented convention is an explicit RuntimeError-with-message (cf. 1599, 2209, 2229; respond() docstring 2054-2059). Under python -O the assert is stripped and line 2334 raises AttributeError. The invariant holds by construction today; the failure-mode difference matters only if classify's contract changes.
7. **[minor, test-quality] tests/web/services/test_task_interaction_service_postgresql.py:944-1115** — The two PG race tests duplicate ~26-29 lines of byte-identical scaffolding (setup/events/monkeypatch/call/thread/racer/before/release/join), differing only in the payload variable and verification sections; a parametrization over the payload-match axis could remove it (verification sections genuinely differ — hedged nit). Also: the replay-branch fence/CAS commit assertion exists only in the SQLite door-two test; door-one coverage is transitive through the shared decision block (task_interaction_service.py:2359-2390) — acceptable, worth keeping if the doors diverge.

- `delete:` respond()'s TASK_MISSING mapping branch at 2320-2322 — unreachable (task row held FOR KEY SHARE since step 2 blocks concurrent DELETE; the fence's unconditional `Task.user_id == principal.user_id` term means only the owner reaches step 8, so the actor FK cannot fail); enqueue keeps the kind live, so let it fall into the existing '!= RACED_DUPLICATE' re-raise + fix the step-8 docstring mapping claim at 2010-2011.
- `shrink:` _verify_respond_durable_graph (1683-1799) — third copy of the fresh-session retry-loop plumbing (range(3), SessionLocal per attempt, sleep(0.01), retire-in-finally, per-attempt warning) after task_orchestrator.py:1145-1218 and websocket.py:4730-4802; parameterize into a shared helper (attempts, sleep, retire, log_label, check: Callable[[Session], T]) for ~20 lines saved per site, zero behavior change.

net: -5 lines possible (3 branch lines deleted + 2 docstring lines; the retry-loop helper is net-neutral in this file)

**Blocking: no** — all 7 confirmed findings are minor (two code nits, one doc nit, three test-quality gaps, one simplification). Recommended event: **APPROVE**.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. The replay branch's commit now runs through the same try/except-and-reconcile shape as the accept path: on a raised commit, `_verify_respond_durable_graph` reads what landed in a fresh session; a graph carrying this call's identity and payload returns `RespondReplayed` with the recovered receipt (not `RespondAccepted` — that is the one difference from the accept path and the reason it could not be shared verbatim), and a graph that did not land returns the unresolved outcome. Both halves of the crash loop you described were reproduced before the fix — the raw raise, and the retry hitting the receipt guard's `RuntimeError` on an active row beside a durable winner command — and two tests now pin them: a lost acknowledgment on a durable graph resolves to `Replayed` naming the winner's row, and a real failure with exhausted verification resolves to unresolved with zero residue. Removing the guard turns the first red with the original exception escaping.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed. Both docstrings now state the two callers and their distinct preconditions: the pre-read replay branch hands the builder an answered row from the same transaction's idempotency hit, and the reconciliation path hands it a row read fresh under `status == answered`. The guard's justification is unchanged — the paired CHECK constraints rule out an answered row with a null timestamp or identity from either source.

tests/web/services/test_task_interaction_service.py
AlexLiu190625: The observation is right and the fix landed on the PostgreSQL side rather than here — with one correction to the suggested assertion. `assert after['commands_count'] == 1` fails as written: `stage_task_command` is patched out entirely in this test, so no command row exists (measured: 0). The real gap was the one your sibling comment on the PG file points at from the other direction: the only test with a genuine winner row never asserted the fence and CAS committed. That test now does — row `answered`, `responded_at` set, `state_version` advanced — and it is the first real pin on this branch's commit: swapping that commit for a rollback turned the entire PG suite green before this round and turns exactly that test red now. This SQLite test keeps its role as the shape check for the receipt fields; the circular id is disclosed in its docstring.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Agreed the branch is unreachable through `respond()` for exactly the reasons you give (the FOR KEY SHARE from step 2 blocks the delete; the fence's owner term keeps the actor FK satisfiable), and deleting three lines was measured to leave the SQLite suite green. Keeping it anyway: with the branch removed, a `TASK_MISSING` classification would fall through to the raw re-raise — trading a typed `Unavailable(task_missing)` for a 500 on a path whose unreachability rests on lock semantics that a future caller could change. The test docstring already discloses the branch is unreached, so there is no misleading residue; the cost of the three lines is a graceful outcome on a path we believe closed. Leaving open for your read.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Agreed on the shape and the count, and declining the extraction in this change: the other two copies live in the task orchestrator and the WebSocket coordinator, both pre-existing production paths, and pulling a shared retry helper through them is a refactor of code this change otherwise does not touch. The right home for that is a change of its own, with all three call sites and their per-attempt logging labels in scope. Leaving open as the record.

src/xagent/web/services/task_interaction_service.py
AlexLiu190625: Fixed — the invariant now raises `RuntimeError` with a message like the module's other three, so the failure mode no longer depends on the interpreter's optimization flag.

tests/web/services/test_task_interaction_service_postgresql.py
AlexLiu190625: Keeping the two tests separate — as you note, the verification sections genuinely differ, and after this round they differ more: the matching-payload test now carries the fence/CAS commit assertions (row `answered`, `responded_at` set, `state_version` advanced), which is what closes the transitive-only coverage you flagged in the second half. That is worth stating plainly: before this round the replay branch's commit was pinned only through the SQLite test with a synthetic winner row; swapping the commit for a rollback left every PostgreSQL test green. Now the door-one race pins it directly, and the doors are no longer covered only transitively.

PR #1383, opened 2026-08-15
fix(mcp): forward structuredContent and read result fields by wire name

### LINKED ISSUE #1378: fix(mcp): forward structuredContent and outputSchema per MCP spec
## Summary

`MCPToolAdapter` (the generic MCP tool integration) discards `CallToolResult.structured_content` and never reads `Tool.output_schema` when building tool definitions. Both fields are defined by the MCP specification itself (not a custom extension) — `output_schema` on `Tool` and `structured_content` on `CallToolResult`, versioned since the 2025-06-18 revision (per the `mcp` Python SDK's own type docstrings, which cite the spec revisions directly). Any MCP server that declares an `outputSchema` and relies on `structuredContent` carrying information not present in the human-readable `content` text is silently broken end-to-end: the LLM never sees the structured result, only the free-text summary.

## Root cause

`src/xagent/core/tools/adapters/vibe/mcp_adapter.py`, `MCPToolAdapter._execute_mcp_call`:

```python
result = await session.call_tool(
    self.mcp_tool.name,
    dict(tool_args),
    meta=dict(tool_meta) or None,
)

content = []
if result.content:
    for content_item in result.content:
        ...

return {
    "content": content,
    "is_error": result.isError if hasattr(result, "isError") else False,
}
```

`result.structured_content` is read nowhere in this method (confirmed via `git log -p --follow` on this file back to the commit that first introduced MCP support — the field has never been referenced). The returned dict only carries `content`/`is_error`, so nothing downstream (including the LLM's own context) ever sees it.

Separately, tool-definition construction for MCP tools never reads `Tool.output_schema` either (also zero matches in this file), so the LLM isn't even told a structured return shape exists in the first place — this is a full gap, not a partial one.

## Observed impact

Reproduced against a real MCP server (a coding-agent dispatcher) whose long-poll status tool returns both:
- `content`: a human-readable text summary (sometimes just a conversational status update, e.g. "I'll wait for the background agents to complete — no need to poll further.")
- `structuredContent`: `{"status": "completed", "handle": {...}, ...}` — the actual machine-readable state

Because `structuredContent` never reaches the model, the LLM has no way to distinguish a terminal, permanently-cached result from an in-progress one by content text alone. In one real run, the agent polled the same terminal `run_id` 50+ times over several minutes (each poll correctly returning the frozen result instantly) before giving up — the tool description had already been updated to explain the terminal-caching behavior, but that guidance is moot if the field it refers to is never delivered to the model.

## Suggested fix

- In `_execute_mcp_call`, include `result.structured_content` in the returned dict when present (e.g. serialize it into the LLM-visible payload, similar to how `websocket.py`'s builder-chat path already reconstructs a JSON block for a different, unrelated feature — that pattern isn't reusable as-is but is a precedent for "serialize structured data into something the LLM can read").
- When building the tool's exposed schema/description from the underlying `mcp.types.Tool`, also surface `output_schema` so the model knows a structured return shape exists at all, not just at the point where a call result happens to include one.

## Test plan (for whoever picks this up)

- A test MCP tool whose `CallToolResult` has `structured_content` set but a `content` that doesn't restate it distinctly — assert the structured value is present somewhere in what `_execute_mcp_call` returns.
- A tool with `output_schema` declared — assert the resulting tool definition surfaces it (schema or description).

rogercloud:
## Summary

`MCPToolAdapter._execute_mcp_call` discarded `CallToolResult.structuredContent` entirely, so the LLM never saw an MCP tool's machine-readable result — only the human-readable `content` text. This silently breaks any MCP server that puts meaningful state only in `structuredContent` (both are official MCP spec fields, not custom extensions).

Reproduced with a real MCP server (a long-poll status tool): its `structuredContent` carried the actual terminal state (`{"status": "completed", ...}`), while `content` held a vague conversational status update. Because `structuredContent` never reached the model, the agent polled the same already-finished run 50+ times.

## Fix

- `_execute_mcp_call` now reads `structured_content`/`is_error` via `result.model_dump(by_alias=True, exclude={"content"})` instead of a `getattr`/`hasattr` chain keyed to one SDK version's Python attribute names. The MCP wire/JSON field names (`structuredContent`, `isError`) are spec-fixed and stable across mcp SDK 1.x and 2.x, even though the SDK renamed its Python attributes between those versions (verified against both a real 1.19.0 and a real 2.0.0 install).
- The adapter's declared return model (`MCPToolResult`) now includes a `structured_content` field, matching what `run_json_async` actually returns.
- `_mcp_return_value_as_string` gained a docstring clarifying it is not the path ReAct's observation text is built from for MCP tool calls (that's `ExecutionContext._format_tool_result`), to avoid a future reader assuming otherwise.

## Out of scope (filed as follow-ups, not fixed here)

- Surfacing the MCP tool's declared `outputSchema` to the LLM would require wiring it into `react.py`'s shared tool-schema construction (used by every tool of every agent) — too broad a blast radius for this bug fix, which is about not discarding a result value that's already returned.
- Hardening `inputSchema` access elsewhere in the adapter (and in `cua_driver.py`) against the same mcp SDK 2.x attribute-naming change is a separate "support mcp 2.x" migration, not part of this fix.
- Servers using the official Python SDK's convenience path (which mirrors dict-shaped structured results into `content` as serialized JSON for backward compatibility) will now see that payload appear twice in the observation text; no truncation/dedup is added here.

## Test plan

- [x] New tests simulate mcp SDK 2.x's field naming (a pydantic double using `alias_generator=to_camel`) to guard against silently regressing to the original bug if the SDK dependency ever resolves to 2.x.
- [x] A seam test runs a fake MCP call through `_execute_mcp_call` and then the real `ExecutionContext.add_tool_result`, asserting the structured data actually reaches the LLM-visible string (not just that the adapter's return dict contains the key).
- [x] Existing fakes for `call_tool`'s return value migrated from loosely-typed `SimpleNamespace` to real `mcp.types.CallToolResult`/`TextContent` instances.
- [x] `tests/core/tools/adapters/vibe/test_mcp_adapter.py` and `tests/core/agent/test_context.py` pass; full `tests/core/tools`/`tests/core/agent` regression sweep passes (excluding pre-existing, unrelated `test_pptx_tool.py` failures present on `main` as well).
- [x] `pre-commit run --all-files` passes (ruff, ruff-format, mypy, isort, codespell, alembic-check; `npm lint`/`npm type-check` fail only due to missing local frontend toolchain, unrelated to this change — no frontend files touched).

Closes #1378.

### DIFF
--- a/src/xagent/core/tools/adapters/vibe/mcp_adapter.py
+++ b/src/xagent/core/tools/adapters/vibe/mcp_adapter.py
@@ -6,6 +6,7 @@ enabling MCP tools to be used in DAG plan-execute patterns and other agent workf
 
 import asyncio
 import inspect
+import json
 import logging
 import os
 import re
@@ -305,20 +306,37 @@ def _mcp_access_denied_result(user_id: Optional[str], tool_name: str) -> dict[st
 
 
 def _mcp_return_value_as_string(value: Any) -> str:
+    """Renders an MCP result dict for the ``AbstractBaseTool.return_value_as_string``
+    interface, used only by the tool wrappers (e.g. output-filter/sandboxed
+    wrappers). This is NOT the path the LLM's observation text is built from
+    for MCP tool calls in ReAct -- that is
+    ``ExecutionContext._format_tool_result`` (execution.py), which
+    stringifies the whole result dict returned by ``_execute_mcp_call``
+    directly. Keep this renderer lossless anyway, since any future caller
+    of ``return_value_as_string`` should see the same fields.
+    """
     try:
         if isinstance(value, dict):
+            texts = []
             content = value.get("content", [])
             if isinstance(content, list) and content:
-                texts = []
                 for item in content:
                     if isinstance(item, dict) and "text" in item:
                         texts.append(item["text"])
                     else:
                         texts.append(str(item))
-                return "\n".join(texts)
-            if content:
-                return str(content)
-            return "No content returned"
+            elif content:
+                texts.append(str(content))
+            else:
+                texts.append("No content returned")
+
+            structured_content = value.get("structured_content")
+            if structured_content is not None:
+                texts.append(
+                    "Structured result: " + json.dumps(structured_content, default=str)
+                )
+
+            return "\n".join(texts)
         return str(value)
     except Exception as e:
         logger.warning(f"Failed to convert return value to string: {e}")
@@ -535,6 +553,13 @@ class MCPToolAdapter(AbstractBaseTool):
         class MCPToolResult(BaseModel):
             content: List[Dict[str, Any]] = Field(
                 default_factory=list, description="Tool execution result content"
+            )
+            structured_content: Any = Field(
+                default=None,
+                description=(
+                    "Structured JSON result of the tool call, if the server "
+                    "returned one."
+                ),
             )
             is_error: bool = Field(
                 default=False,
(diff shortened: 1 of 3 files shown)


qinxuye (approved):
LGTM

